PR #1572, opened 2026-04-23
docs(uni-stark): clarify selector degree semantics

RekCuy63:
The current comment says `S_i^2` can be replaced by `S_i`, but the selector definitions used here are not normalized. This updates the comment to match the semantics already documented in `p3-commit` and `p3-lookup`, and keeps the degree discussion consistent with the symbolic degree model.

### DIFF
--- a/uni-stark/src/prover.rs
+++ b/uni-stark/src/prover.rs
@@ -115,9 +115,9 @@ where
     //          C(x) = C(T_1(x), ..., T_w(x), T_1(hx), ... T_w(hx), S_1(x), S_2(x), S_3(x))
     // We get the constraint bound:
     //          deg(C(x)) <= deg(C) * (N - 1) + 1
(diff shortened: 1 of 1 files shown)


SyxtonPrime:
I think this is misunderstanding the initial comment.

While it is true that `S_i^2` and `S_i` are different polynomials, the zeroes occur at the same places. Hence if you have a constraint polynomial: `C(x) = S_i^2f(x)`  you can always safely replace it by the simpler polynomial `C(x) = S_if(x)` as the first will be zero only when the second is zero and visa versa.

RekCuy63:
Hi @SyxtonPrime That makes sense. Thanks for pointing that out. I'll update the wording to clarify that distinction.

RekCuy63:
@SyxtonPrime Updated the comment to make the zero-set distinction explicit for `S_i^2` vs `S_i`. Please take another look.

SyxtonPrime:
If I'm perfectly honest, I think I prefer the old comment.

Apologies

PR #1574, opened 2026-04-24
Perf simd compress hi dot

camofu:
## Summary
 
Optimizes `compress_hi_dot` — the kernel for high-variable compression, computing the inner product of one polynomial row with the eq table weighted by eq0 — for the Packed variant.
 
The rewrite re-routes the kernel through a delayed-reduction SIMD dot product. Each per-row `PackedEF × F::Packing` product decomposes into four per-coefficient `F::Packing × F::Packing` dot products, which call Plonky3's existing `F::Packing::dot_product::<4>` — the `dot_product_4` kernel that accumulates u64 products (VPMULUDQ + VPADDQ on x86, analogous intrinsics on NEON) and applies one Montgomery reduction per 4-wide group instead of one per multiply. Two wins stack:
 
- ~4× fewer Montgomery reductions in the inner loop.
- Better instruction-level parallelism from four independent per-coefficient accumulators instead of one serial chain.

## On the use of `MaybeUninit`
 
A default-initialized `[F::Packing; 128]` would eagerly zero-fill every call; `MaybeUninit::uninit()` is a compile-time no-op, so only the `n` slots we actually use get written. The resulting prefix is exposed as `&[F::Packing]` through a narrow `unsafe` block — invariant: first `n` slots written before any read, established by the adjacent write loop.
 
## The ILP gain
 
The eager baseline has a single accumulator chain — iteration `i+1`'s `acc += ...` is gated by iteration `i`, so the loop is latency-bound on the mul-add dependency. The per-coefficient decomposition yields four independent accumulators the CPU can pipeline across its SIMD execution ports; the loop becomes throughput-bound. This is why the measured speedup exceeds what "~4× fewer Monty reductions" alone would predict.


  ## Benchmarks                                                                                                               
  
  Ryzen 7950X3D, AVX-512, `RUSTFLAGS="-Ctarget-cpu=native"`,                                                                  
  single-threaded, 20 samples × 8 s.                     
                                                                                                                              
  ### `whir/svo_claim_build` — direct caller of `compress_hi_dot`                                                             
  
  | Config                    | main     | this PR    | speedup   |                                                           
  |---------------------------|---------:|-----------:|----------:|
  | `claim_build/k12_l1`      | 8.99 µs  | 4.06 µs    | 2.21×     |                                                           
  | `claim_build/k12_l2`      | 11.10 µs | 7.39 µs    | 1.50×     |                                                           
  | `claim_build/k12_l3`      | 11.77 µs | 8.02 µs    | 1.47×     |                                                           
  | `claim_build/k12_l4`      | 16.55 µs | 14.09 µs   | 1.17×     |                                                           
  | **`claim_build/k22_l4`**  | **6.93 ms** | **1.38 ms** | **5.02×** |                                                       
                                                                                                                              
  ### `whir/sumcheck_prover` — not a caller of `compress_hi_dot`, regression check                                            
                                                                                                                              
  | Config            | main     | this PR   | ratio         |                                                                
  |-------------------|---------:|----------:|--------------:|
  | `classic/small`   | 330 µs   | 314 µs    | 1.05× (noise) |
  | `svo/small`       | 39.7 µs  | 40.4 µs   | 0.98× (noise) |                                                                
  | `classic/medium`  | 5.33 ms  | 5.03 ms   | 1.06× (noise) |                                                                
  | `svo/medium`      | 548 µs   | 549 µs    | parity        |                                                                
  | `classic/large`   | 122 ms   | 117 ms    | 1.04× (noise) |                                                                
  | `svo/large`       | 8.77 ms  | 8.77 ms   | parity        |                                                                
                                                                                                                              
  All `sumcheck_prover` differences are within the criterion confidence                                                       
  intervals of each other. No regression observed.                                                                            
                                                                                                                              
  Reproduce via the included `reproduce_benches.sh` script.                                                          
                                                                                                                              
  ### Reproducing                                                                                                             
                                                         
To reproduce the benches                                                          
  ```bash
#!/usr/bin/env bash
# ~20 min. Linux + AVX-512 expected. Clean working tree required.
set -euo pipefail                                                                                                           
 
PR=perf-simd-compress-hi-dot                                                                                                
OUT=$(mktemp -d)                                                                                                            
echo "Bench outputs: $OUT"
                                                                                                                            
# The existing svo_claim_build bench only has k12_l* cases — too small to
# exercise compress_hi_dot's SIMD inner loop (N=1-2). Adding (22, 4) gives                                                  
# the production-sized N=32 case that every real WHIR proof hits.                                                           
add_k22() {                                                                                                                 
    grep -q k22_l4 whir/benches/sumcheck.rs \                                                                               
        || sed -i '/(12, 4, "k12_l4"),/a\        (22, 4, "k22_l4"),' \                                                      
                whir/benches/sumcheck.rs                                                                                    
}
                                                                                                                            
run_on() {                                             
    local branch=$1
    git checkout "$branch"
    add_k22                                                                                                                 
    for filter in svo_claim_build sumcheck_prover; do
        echo ">>> $branch / $filter"                                                                                        
        RUSTFLAGS="-Ctarget-cpu=native" cargo bench -p p3-whir --bench sumcheck -- \
            "$filter" --warm-up-time 3 --measurement-time 8 --sample-size 20 \                                              
            > "$OUT/$branch-$filter.txt" 2>&1                                                                               
    done                                                                                                                    
    git checkout -- whir/benches/sumcheck.rs                                                                                
}                                                      

run_on "$PR"
run_on main                                                                                                                 
git checkout "$PR"
                                                                                                                            
for f in "$PR"-svo_claim_build main-svo_claim_build \  
         "$PR"-sumcheck_prover main-sumcheck_prover; do
    echo; echo "=== $f ==="                                                                                                 
    grep -E "Benchmarking whir/|time:" "$OUT/$f.txt" || true
done                                                                                                                                                                                                                              
  ```


tcoratger:
Took a closer review pass on the kernel. Here is what just landed in `830eecd0`.

## New optimisation — interleaved inner loop

The previous inner loop used `core::array::from_fn(|k| ...)` to drive the four per-coefficient dot products as four sequential passes. LLVM could not safely fuse those passes, so the CPU serialised their SIMD multiplies and reloaded the shared `piece` block four times per output row.

Reshaped into a single interleaved loop that:

- runs all four `F::Packing::dot_product::<4>` calls per step, exposing four independent accumulator chains to the out-of-order engine,
- loads each shared `piece` block once per step instead of four times.

Safe implementation — no raw pointers; `try_into().unwrap()` retained, LLVM elides the panic given the surrounding `i + CHUNK <= n` bound.

## Bench numbers

Measured on **Apple M1 Pro / NEON** (so magnitudes will be lower than your Ryzen / AVX-512 figures; ARM's 4-wide packing gives less headroom per chain). The existing `k12_l*` cases don't exercise the SIMD path (N = 1..4 on NEON); added k16/k18/k20 which give production-shape N = 16 / 32 / 64.

| `svo_claim_build` | `main` | PR (before this commit) | PR + interleave | total vs. `main` |
|---|---:|---:|---:|---:|
| k16_l4  (N=16) | 91.28 µs | 80.91 µs | **70.23 µs** | **1.30×** |
| k18_l4  (N=32) | 307.1 µs | 271.13 µs | **216.08 µs** | **1.42×** |
| k20_l4  (N=64) | 1.164 ms | 969.64 µs | **761.96 µs** | **1.53×** |

Regression checks — all within ±1% noise (parity):

- `whir/sumcheck_prover/{classic,svo}/{small,medium,large}`
- `multilinear-util/eval_eq_batch/*` and `eval_eq_base_batch/*`

## Refactors and hygiene

- **Rename** `split_eq/delayed.rs` → `split_eq/packed_kernel.rs`; the `fallback` function was a single-site helper, so it's inlined into the main function now.
- **Fallback shape** matches the scalar arm in `eq.rs`: `chunk_packed.chunks_exact(n).zip_eq(eq0.as_slice())`, no manual index arithmetic.
- **`debug_assert_eq!`** at entry naming the chunk-length invariant (`chunk_packed.len() == n * |eq0|`). A bad caller now gets "chunk must hold |eq0| * N * W base-field elements (got X packed, expected Y)" instead of a generic `zip_eq` iterator-length panic.
- **Docs rewritten** to the repository doc conventions: `# Overview` / `# Algorithm` / `# Memory Layout` / `# Arguments` / `# Returns` / `# Panics` sections, ASCII diagram of the AoS → SoA transpose, short paragraphs with bullet lists, no cross-references to specific type or function names in doc text.

## Tests added in `packed_kernel::tests`

Four new tests covering the kernel directly against a naive triple-nested reference:

- **`matches_reference`** (proptest) — full shape space: `n ∈ 0..=40`, `m_vars ∈ 0..=4`, random seed.
- **`fast_path_matches_fallback_across_boundary`** (proptest) — `n ∈ 126..=130` so both sides of `MAX_STACK_N = 128` are exercised; asserts the threshold is a performance switch, not semantic.
- **`empty_eq1_is_zero`** — `N = 0` edge case.
- **`scalar_tail_exercised`** — residues `n mod 4 ∈ {1, 2, 3}` at several magnitudes, ensuring the tail loop is correct at every residue class.

All 33 `split_eq` lib tests pass (29 pre-existing + 4 new).

---

Let me know if you'd rather split this into smaller commits (perf / rename / docs / tests), happy to rebase.

tcoratger:
Follow-up to commit `830eecd0`: with `8796ac33` (generic kernel) and `2b6bae49` (doc + clippy polish), the fast path is now **generic over the extension-field dimension**. Any `D ∈ 1..=8` hits the basis-split / delayed-reduction / interleaved-loop kernel — not just `D = 4`.

## Current shape

A const-generic inner function:

```rust
fn basis_split_dot<F, EF, const D: usize>(...) -> EF { ... }
```

with the per-coefficient loops written as `for k in 0..D` over `[F::Packing; D]` arrays. The public entry dispatches on `EF::DIMENSION` to lift the trait constant into a true const-generic parameter:

```rust
match EF::DIMENSION {
    1 => return basis_split_dot::<F, EF, 1>(eq1_packed, chunk_packed, eq0, n),
    2 => return basis_split_dot::<F, EF, 2>(eq1_packed, chunk_packed, eq0, n),
    ...
    8 => return basis_split_dot::<F, EF, 8>(eq1_packed, chunk_packed, eq0, n),
    _ => {}                  // falls through to the eager fallback
}
```

Each match arm gets its own monomorphisation, so LLVM unrolls the `for k in 0..D` loop at every supported dimension and SROAs the per-coefficient array into scalar registers — the same property the prior D=4-only body got from named locals.

Why `1..=8`: covers every extension dimension in Plonky3 today (Goldilocks ext2, BabyBear / KoalaBear ext4, Goldilocks quintic ext5) plus a safe margin. Beyond `D = 8`, AVX-512 register pressure starts to bite and the interleave win fades, so the eager fallback is the right choice. Adding a new dimension is a one-line match arm.

## D = 4 parity check (Apple M1 Pro, NEON)

`whir/svo_claim_build`, prior D=4-only body vs current generic body:

| Case   | D=4-only | generic | criterion verdict |
|--------|---------:|--------:|-------------------|
| k16_l4 |  70.23 µs |  67.51 µs | "No change in performance detected" |
| k18_l4 | 216.08 µs | 214.83 µs | -0.37% (within noise) |
| k20_l4 | 761.96 µs | 798.52 µs | "Change within noise threshold" (p = 0.55) |

Zero regression at `D = 4`; every other supported dimension gets the same speedup automatically.

## `CHUNK = 4` rationale (now field-agnostic)

The doc on the inner-loop group size leads with the portable reason: every `F::Packing` impl specialises `dot_product::<4>` to a hand-tuned four-wide kernel, so routing through `<4>` inherits each field's best primitive — Monty-31, Goldilocks, and any future field plug into the same call site. The Monty-31 `u64` overflow argument (`4 × 2^62 < 2^64`) is kept as an aside, not a precondition.

## Tests

33 `split_eq` lib tests pass (29 pre-existing + 4 in `packed_kernel::tests`):

- **`matches_reference`** (proptest) — `n ∈ 0..=40`, `m_vars ∈ 0..=4`, random seed; full shape space.
- **`fast_path_matches_fallback_across_boundary`** (proptest) — `n ∈ 126..=130`; both sides of `MAX_STACK_N = 128`.
- **`empty_eq1_is_zero`** — `N = 0` edge case.
- **`scalar_tail_exercised`** — `n mod 4 ∈ {1, 2, 3}` at several magnitudes.

The proptests exercise `D = 4` directly against a naive triple-nested reference; the dispatcher routes that through the const-generic body, so the existing coverage validates the implementation at `D = 4`. For other dimensions, correctness follows from monomorphisation symmetry — the body is dimension-agnostic, only the dispatcher varies.

`cargo fmt` clean. `cargo clippy -p p3-multilinear-util / -p p3-whir -- -D warnings` clean.

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -0,0 +1,435 @@
...
+
+/// Inner group size for the delayed-reduction primitive.
+///
+/// - Every `F::Packing` impl specialises `dot_product::<4>` to a
+///   hand-tuned four-wide kernel, so the choice is portable across
+///   fields (Monty-31, Goldilocks, …).
+/// - On 31-bit fields it is also the `u64` overflow ceiling
+///   (`4 * 2^62 < 2^64`).
+/// - `N mod 4` residues fall to the scalar tail.
+const CHUNK: usize = 4;
```
tcoratger, line 214: Are we sure about the behaviour of this constant for other fields/extension dimension, for example Goldilocks extension 2? I don't think that the 4 value here is generic enough to be fast for all the field configurations that the codebase provides no?

camofu: Yes, you are right. The thing is that `CHUNK` is like 2 costants in one at this point, because it is the Cross-coefficient interleave grain for the interleaved per-coefficient loop and the 'dot_product' arity that on Monty-31 enables the delayed Montgomery reduction (governed by the 4 * 2^62 < 2^64 overflow ceiling). 

The thing is that I do not think it makes much sense to split them in two different constants in this PR as they are both set at 4. The `dot_product' airty is forced by the 31-bit fiels and the u64 SIMD. And for the interleave grain 4 is the smallest legal value, as it has to be a multiple of the dot_product arity. I'll push a more descriptive comment for this, it should be clear enough at this stage I think.

tcoratger (approved):

PR #1576, opened 2026-04-24
fix(monty-31): merge twiddle cache locks

RekCuy63:
Before: `RecursiveDft::update_twiddles` stored `twiddles` and `inv_twiddles` behind separate `RwLock`s. A thread could observe the forward table after an extension and skip the update while `inv_twiddles` was still stale, which could panic or compute with the wrong inverse table.

After: both tables live in a single `TwiddlePair<F>` behind one `RwLock`. Cache growth appends both missing slices while holding the same write lock, so readers only see matching forward and inverse tables.

### DIFF
--- a/monty-31/src/dft/mod.rs
+++ b/monty-31/src/dft/mod.rs
@@ -40,24 +40,33 @@ fn coset_shift_and_scale_rows<F: Field>(
         });
 }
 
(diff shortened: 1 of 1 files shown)


tcoratger:
Duplicate of https://github.com/Plonky3/Plonky3/pull/1561

PR #1577, opened 2026-04-25
feat(field): add Field::mul_slices for elementwise slice multiplication

tcoratger:
We had the addition version of it but were missing the multiplication version.

## Summary

- Adds `Field::mul_slices` as a multiplicative sibling to the existing `Field::add_slices`. Same SIMD-friendly split-prefix + scalar-tail pattern via `<Self::Packing as PackedValue>::pack_slice_with_suffix{_mut}`. Default impl, no extension overrides (multiplication is not F-linear, so the `repr(transparent)` cast trick used by `add_slices` overrides on `BinomialExtensionField` / `QuinticTrinomialExtensionField` does not apply).
- Wires `test_add_slice` and `test_mul_slice` into the `test_field!` macro so every existing call site gets coverage for free: KoalaBear, BabyBear, Mersenne31, Mersenne31Complex, Goldilocks, BN254, plus all binomial / quintic / octic / complex extensions (34 new tests across 5 field crates).
- Adds `benchmark_mul_slices` to `field-testing/src/bench_func.rs` and registers it next to the existing `benchmark_add_slices` in `koala-bear/benches/extension.rs` for `EF4` and `EF8` at lengths 8 and 1000.

Mirrors the design of plonky2's `field/src/batch_util::batch_multiply_inplace`. Plonky2's `batch_add_inplace` is already covered by the existing `Field::add_slices`, so this PR only adds the missing multiplicative half.

## Test plan

- [x] `cargo build -p p3-field --no-default-features` (no_std clean)
- [x] `cargo test -p p3-koala-bear -p p3-baby-bear -p p3-mersenne-31 -p p3-goldilocks -p p3-bn254 --lib` — 861 pass (was 827 before, +34 new tests)
- [x] `cargo clippy -p p3-field -p p3-field-testing -p p3-koala-bear -p p3-baby-bear -p p3-mersenne-31 -p p3-goldilocks -p p3-bn254 --no-deps -- -D warnings`
- [x] `cargo build -p p3-koala-bear --bench extension`
- [x] `RUSTDOCFLAGS=\"-D warnings\" cargo doc -p p3-field --no-deps`
- [x] `cargo fmt --check -p p3-field -p p3-field-testing`

### DIFF
--- a/field/src/field.rs
+++ b/field/src/field.rs
@@ -916,6 +916,33 @@ pub trait Field:
         }
     }
 
+    /// Multiply two slices of field elements together elementwise, returning the result in the first slice.
+    ///
+    /// Makes use of packing to speed up the multiplication.
+    ///
+    /// This is optimal for cases where the two slices are small to medium length. E.g. between
+    /// `F::Packing::WIDTH` and roughly however many elements fit in a cache line.
+    ///
+    /// For larger slices, it's likely worthwhile to use parallelization before calling this.
+    /// Similarly if you need to multiply a large number of slices together, it's best to
+    /// break them into small chunks and call this on the smaller chunks.
+    ///
+    /// # Panics
+    /// The function will panic if the lengths of the two slices are not equal.
+    #[inline]
+    fn mul_slices(slice_1: &mut [Self], slice_2: &[Self]) {
+        let (shorts_1, suffix_1) = Self::Packing::pack_slice_with_suffix_mut(slice_1);
+        let (shorts_2, suffix_2) = Self::Packing::pack_slice_with_suffix(slice_2);
+        debug_assert_eq!(shorts_1.len(), shorts_2.len());
+        debug_assert_eq!(suffix_1.len(), suffix_2.len());
+        for (x_1, &x_2) in shorts_1.iter_mut().zip(shorts_2) {
+            *x_1 *= x_2;
+        }
+        for (x_1, &x_2) in suffix_1.iter_mut().zip(suffix_2) {
+            *x_1 *= x_2;
+        }
+    }
+
     /// The number of elements in the field.
     ///
     /// This will either be prime if the field is a PrimeField or a power of a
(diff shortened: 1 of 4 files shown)


field/src/field.rs
```diff
@@ -916,6 +916,33 @@ pub trait Field:
...
+        let (shorts_2, suffix_2) = Self::Packing::pack_slice_with_suffix(slice_2);
+        debug_assert_eq!(shorts_1.len(), shorts_2.len());
+        debug_assert_eq!(suffix_1.len(), suffix_2.len());
+        for (x_1, &x_2) in shorts_1.iter_mut().zip(shorts_2) {
+            *x_1 *= x_2;
+        }
+        for (x_1, &x_2) in suffix_1.iter_mut().zip(suffix_2) {
+            *x_1 *= x_2;
+        }
+    }
```
SyxtonPrime, line 944: I don't have anything against adding this function but I do wonder how useful it is.

One of the nice things about `add_slices` is that there are lots of situations where we need to add vectors of field elements together and there are some nice synergies like summing extension elements is the same as summing the base field elements so you can reduce from that case to this one.

Unfortunately that trick doesn't work with `mul_slices`. Similarly I find you often want to multiply a vector of field elements by a single element or similar instead of another vector.

If you have a use case in mind for this or someone has asked for this I'm happy to add it but otherwise I'd maybe hold off?

SyxtonPrime (approved):
See comment

field/src/field.rs
tcoratger: I've no use case in mind, this was just to have parity with the add but ok this makes sense to me to hold off until we find a specific attractive use case for this :)

Nashtare: What's the decision on this? Should we close it or would you still like to see it merged @tcoratger

tcoratger: @Nashtare Let me close for now :)

PR #1579, opened 2026-04-28
fix(batch-stark): stabilize global lookup error order

Daisy-Lord11:
Sort global lookup cumulative checks by interaction name before returning the first mismatch.

This makes verifier diagnostics deterministic when multiple global lookup groups fail in the same proof, and adds a release-mode regression test for that case.


Nashtare:
Superseded by #1604

PR #1591, opened 2026-04-28
fix(goldilocks): canonicalize sub_asm and harden NEON ASM tests

tcoratger:
@Nashtare It looks like we had the same problem for `sub_asm`.

Also to avoid potential future issues, I've added a bunch of super edge case scenarios found by Claude to prevent potential future bugs during optimizations because these edge cases are often the trickiest to detect.

## Summary

The NEON `sub_asm` had the same shape of bug as the `add_asm` issue fixed in #1580: it never canonicalized the second operand, so for non-canonical `b > P` where `a < b - P`, the result was not congruent mod P. Current Poseidon2 callers happened to dodge it because they always pass a canonical first operand (the output of an `add_asm`, which is now reduced to `[0, P)`), but a future optimization could trip it. Applied the same canonicalize-`b`-then-final-reduction pattern.

A concrete failing input on `main`:

```rust
sub_asm(0, P + 1)
// returns 2^64 - 1, which canonicalizes to EPSILON - 1 = 4_294_967_294
// expected:                         P - 1 = 18_446_744_069_414_584_320
```

The PR also broadens test coverage for the NEON ASM primitives so this class of bug cannot slip through again. Uniform `u64` sampling lands in the non-canonical band `[P, 2^64)` with probability ~`2^-32`, so the existing proptests almost never exercised that range — which is why both this and the #1580 bug went undetected.

### What's added

- Inlined boundary tables (`EDGE_VALUES`) and biased proptest strategies (`danger_u64`, `danger_array`) in each test module.
- Exhaustive edge-pair / edge-triple checks for every scalar op: `add_asm`, `sub_asm`, `mul_asm`, `mul_add_asm`, `double_asm`, `div2/4/8/16/32_asm`, `div_2_32_asm`.
- 1000-iter chained-accumulation regression for `add_asm`.
- Danger-zone proptest variants for layer-level ops: `apply_mat4_asm`, `sbox_layer_asm`, `mds_light_permutation_asm` (w8/w12/w16), `external_round_asm`, `external_round_fused_w8`.
- Adversarial-state stress tests for the full Poseidon2 internal permute at widths 8/12/16/20 (generic + specialized) and the external rounds, compared against a field-level reference.
- Stress tests for Poseidon1 `cheap_matmul_asm_w8/w12` and `add_rc_asm`.

### Bug discovery

The new `test_sub_asm_edge_pairs` and `test_sub_asm_danger` tests fail on `main` (without the `sub_asm` fix in this PR) and pass with it.

## Test plan

- [x] `cargo test -p p3-goldilocks --lib` — all 319 tests pass on aarch64 NEON.
- [x] `test_sub_asm_edge_pairs` confirmed to fail on the unfixed `sub_asm`, pass with the fix.
- [x] All adversarial-state stress tests for Poseidon2 internal/external permutes pass.

### DIFF
--- a/goldilocks/src/aarch64_neon/poseidon1_asm.rs
+++ b/goldilocks/src/aarch64_neon/poseidon1_asm.rs
@@ -404,14 +404,13 @@ pub unsafe fn add_scalar_s0_asm(state: &mut [u64], rc: u64) {
 
 #[cfg(test)]
 mod tests {
-    use alloc::vec::Vec;
-
     use p3_field::PrimeField64;
     use proptest::prelude::*;
     use rand::SeedableRng;
     use rand::rngs::SmallRng;
 
     use super::*;
+    use crate::aarch64_neon::danger_array;
     use crate::{Goldilocks, P};
 
     type F = Goldilocks;
@@ -424,50 +423,6 @@ mod tests {
     /// against field-level references.
     fn canon(x: u64) -> u64 {
         F::new(x).as_canonical_u64()
-    }
-
-    const EPSILON: u64 = P.wrapping_neg();
-
-    /// Boundary u64s probed against every scalar ASM op.
-    const EDGE_VALUES: &[u64] = &[
-        0,
-        1,
-        2,
-        EPSILON - 1,
-        EPSILON,
-        EPSILON + 1,
-        1u64 << 31,
-        (1u64 << 32) + 1,
-        1u64 << 33,
-        1u64 << 63,
-        P - 2,
-        P - 1,
-        P,
-        P + 1,
-        P + 2,
-        18_446_744_069_605_983_184,
-        18_446_744_073_709_551_599,
-        u64::MAX - 1,
-        u64::MAX,
-    ];
-
-    /// Strategy biased toward the non-canonical band.
-    fn danger_u64() -> impl Strategy<Value = u64> {
-        prop_oneof![
-            prop::sample::select(EDGE_VALUES.to_vec()),
-            P..u64::MAX,
-            P..=P.saturating_add(EPSILON - 1),
-            any::<u64>(),
-        ]
-    }
-
-    /// Length-`WIDTH` array of danger-band u64s.
-    fn danger_array<const WIDTH: usize>() -> impl Strategy<Value = [u64; WIDTH]> {
-        prop::collection::vec(danger_u64(), WIDTH).prop_map(|v: Vec<u64>| {
-            let mut a = [0u64; WIDTH];
-            a.copy_from_slice(&v);
-            a
-        })
     }
 
     proptest! {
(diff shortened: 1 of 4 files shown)


goldilocks/src/aarch64_neon/poseidon1_asm.rs
```diff
@@ -424,6 +426,50 @@ mod tests {
         F::new(x).as_canonical_u64()
     }
 
+    const EPSILON: u64 = P.wrapping_neg();
+
+    /// Boundary u64s probed against every scalar ASM op.
+    const EDGE_VALUES: &[u64] = &[
```
Nashtare: could be nice to reuse the one defined in `utils` for deduplication

goldilocks/src/aarch64_neon/poseidon1_asm.rs
```diff
@@ -424,6 +426,50 @@ mod tests {
...
+    }
+
+    /// Length-`WIDTH` array of danger-band u64s.
+    fn danger_array<const WIDTH: usize>() -> impl Strategy<Value = [u64; WIDTH]> {
+        prop::collection::vec(danger_u64(), WIDTH).prop_map(|v: Vec<u64>| {
+            let mut a = [0u64; WIDTH];
+            a.copy_from_slice(&v);
+            a
+        })
+    }
```
Nashtare: same for this

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1594,6 +1610,138 @@ mod tests {
         unsafe { (vgetq_lane_u64::<0>(v), vgetq_lane_u64::<1>(v)) }
     }
 
+    const EPSILON: u64 = P.wrapping_neg();
+
+    /// Boundary u64s probed against every scalar ASM op.
+    const EDGE_VALUES: &[u64] = &[
```
Nashtare: same here, can reuse utils

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1594,6 +1610,138 @@ mod tests {
...
+    }
+
+    /// Length-`WIDTH` array of danger-band u64s.
+    fn danger_array<const WIDTH: usize>() -> impl Strategy<Value = [u64; WIDTH]> {
+        prop::collection::vec(danger_u64(), WIDTH).prop_map(|v: Vec<u64>| {
+            let mut a = [0u64; WIDTH];
+            a.copy_from_slice(&v);
+            a
+        })
+    }
```
Nashtare: and this

Nashtare (approved):

goldilocks/src/aarch64_neon/poseidon1_asm.rs
tcoratger: Done

goldilocks/src/aarch64_neon/poseidon1_asm.rs
tcoratger: Done

goldilocks/src/aarch64_neon/poseidon2_asm.rs
tcoratger: Done

goldilocks/src/aarch64_neon/poseidon2_asm.rs
tcoratger: Done

PR #1592, opened 2026-04-28
perf(split_eq): SIMD-delayed compress_prefix_to_packed for the Packed eq1 path

tcoratger:
## Summary

Optimises `EqMaybePacked::compress_prefix_to_packed_into` on the **Packed eq1** path — the WHIR sumcheck per-round caller of the split-eq machinery. Same shape and same trick as PR #1574 (which fixed the suffix-compression case), applied to the prefix-compression sibling.

The packed-eq1 path used to apply one Montgomery reduction per inner `PackedEF × F::Packing` multiply. Three optimisations stack to slash that cost:

- **Basis split.** `PackedEF × F::Packing` factors into `D` independent scalar-on-packed multiplies, one per basis coefficient.
- **Delayed Montgomery reduction.** `D` per-coefficient `mixed_dot_product::<4>` calls amortise one reduction over four multiplies (~4× fewer reductions).
- **Output tiling.** 8 output positions in flight per pass over `(w_idx, lane)` — the `D × 8` per-coefficient deferred sums stay in scalar registers, the `chunk` buffer is still walked sequentially.

Phase 1 of this PR also factors the per-coefficient delayed-reduction inner of `basis_split_dot` into a shared `packed_basis_dot` helper so both kernels drive the same primitive.

## Benchmarks

Apple M1 Pro / NEON, `RUSTFLAGS=\"-Ctarget-cpu=native\"`, single-threaded, 30 samples × 5 s on the kernel benches and 20 samples × 5 s on the WHIR benches.

### `compress_prefix_to_packed/packed` — direct kernel-level bench

| Shape    | main      | this PR   | speedup |
|----------|----------:|----------:|--------:|
| `k12_eq6`   | 5.08 µs   | 3.69 µs   | **1.38×** |
| `k16_eq8`   | 73.0 µs   | 56.4 µs   | **1.29×** |
| `k20_eq10`  | 1.14 ms   | 894 µs    | **1.28×** |
| `k16_eq1`   | 69.7 µs   | 70.5 µs   | parity (Unpacked branch, untouched) |
| `k20_eq1`   | 1.46 ms   | 1.47 ms   | parity (Unpacked branch, untouched) |

### `whir/sumcheck_prover/svo` — end-to-end production caller

| Config  | main      | this PR   | speedup   |
|---------|----------:|----------:|----------:|
| `small`   | 47.7 µs   | 44.6 µs   | **1.07×** |
| `medium`  | 576 µs    | 524 µs    | **1.10×** |
| `large`   | 10.4 ms   | 8.17 ms   | **1.27×** |

### Regression checks (all within criterion noise)

- `whir/sumcheck_prover/classic/{small,medium,large}` — no compress_prefix_to_packed on the path; parity confirmed.
- `whir/svo_claim_build/k{12_l1,12_l2,12_l3,12_l4,16_l4,18_l4,20_l4}` — exercises the suffix kernel only; parity confirmed.
- `multilinear-util/eval_eq_batch` and `eval_eq_base_batch` (24 shapes each) — within ±3% / ±5%, no significant change.

## Repro

```bash
RUSTFLAGS=\"-Ctarget-cpu=native\" \\
  cargo bench -p p3-multilinear-util --bench compress_kernels

RUSTFLAGS=\"-Ctarget-cpu=native\" \\
  cargo bench -p p3-whir --bench sumcheck -- \\
    \"sumcheck_prover|svo_claim_build\" \\
    --warm-up-time 2 --measurement-time 5 --sample-size 20
```

## Infrastructure improvements

- **New** `multilinear-util/benches/compress_kernels.rs`: per-kernel criterion suite covering every compress/dot kernel in isolation.
- `whir/benches/sumcheck.rs`: extended `bench_svo_claim_build` with production-shape `k16_l4`, `k18_l4`, `k20_l4` cases (mirroring the ad-hoc additions PR #1574 made via a script). Future kernel-touching PRs are now gated on the SIMD path on every supported width without needing local edits.
- `compress_hi_dot_packed` signature changed from `&Poly<EF>` to `&[EF]` for slice-friendliness; behaviour and benchmark numbers unchanged (PR #1574 parity verified).

## What did not pay off (explored, dropped)

- **Routing the scalar-output prefix kernel through the new SIMD primitive + unpacking at the end.** Saved ~9% on the largest kernel shape but **regressed** small shapes by ~17% (the temporary buffer alloc + lane-wise unpack ate the win). Did not move WHIR-level numbers either. Reverted.
- **Routing `dot_with_base` packed through `compress_hi_dot_packed` with a unit weight.** Correctness fine; the existing kernel's fixed setup overhead (SoA transpose, coefficient reassembly) made it ~10–17% slower for a single-row dot. Reverted.

Both decisions follow the same rule: only ship if both the kernel-level and the end-to-end WHIR benches show a real improvement.

## Tests

`126 p3-multilinear-util` lib tests pass, including 6 new ones in `packed_kernel::tests`:

- `prefix_to_packed_kernel_matches_eager` — proptest, kernel ≡ eager fallback across shapes straddling the `TILE = 8` boundary.
- `prefix_to_packed_kernel_empty_eq1_is_noop`
- `prefix_to_packed_kernel_zero_w0_is_noop`
- `prefix_to_packed_kernel_partial_tile`
- `prefix_to_packed_kernel_multi_tile`
- `prefix_to_packed_kernel_accumulates_into_out`

## Test plan

- [x] `cargo test -p p3-multilinear-util` — 126 tests pass.
- [x] `cargo test -p p3-whir --lib` — 183 tests pass.
- [x] `cargo clippy -p p3-multilinear-util -p p3-whir --all-targets` — no warnings.
- [x] `cargo fmt --check` — clean.
- [x] Kernel benches recorded above.
- [x] WHIR end-to-end benches recorded above.
- [x] `evaleq_batched` regression check recorded above.

### DIFF
--- a/multilinear-util/benches/compress_kernels.rs
+++ b/multilinear-util/benches/compress_kernels.rs
@@ -1,0 +1,192 @@
+//! Per-kernel benches for the split-eq compress and dot kernels.
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_baby_bear::BabyBear;
+use p3_field::PrimeCharacteristicRing;
+use p3_field::extension::BinomialExtensionField;
+use p3_multilinear_util::point::Point;
+use p3_multilinear_util::poly::Poly;
+use p3_multilinear_util::split_eq::SplitEq;
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+type F = BabyBear;
+type EF = BinomialExtensionField<F, 4>;
+
+/// Deterministic RNG seeded from a bench label and a variable count.
+///
+/// Different shapes get different seeds so cache state of one bench
+/// does not leak into another, while a single shape stays reproducible
+/// across runs.
+fn rng_for(label: &str, k: usize) -> SmallRng {
+    // FNV-style mix: combine the variable count with each label byte.
+    let mut h: u64 = k as u64;
+    for b in label.bytes() {
+        h = h.wrapping_mul(0x100000001b3).wrapping_add(b as u64);
+    }
+    SmallRng::seed_from_u64(h)
+}
+
+/// Suffix-variable dot product at production shapes.
+///
+/// # Role
+///
+/// - Regression guard for the SIMD kernel introduced in PR #1574.
+/// - Comparison anchor for the new prefix-side kernels.
+fn bench_compress_suffix_dot(c: &mut Criterion) {
+    let mut group = c.benchmark_group("split_eq/compress_suffix_dot");
+
+    // Variable counts that mirror the SvoClaim midpoint shapes.
+    let cases = [(12usize, "k12"), (16, "k16"), (20, "k20")];
+
+    for &(k_total, label) in &cases {
+        // Build inputs deterministically so reruns are stable.
+        let mut rng = rng_for(label, k_total);
+        let poly = Poly::<F>::rand(&mut rng, k_total);
+        let point = Point::<EF>::rand(&mut rng, k_total);
+
+        // Pre-build the split eq table outside the timing loop:
+        // construction cost is not part of the kernel we measure.
+        let split_unpacked = SplitEq::<F, EF>::new_unpacked(&point, EF::ONE);
+        let split_packed = SplitEq::<F, EF>::new_packed(&point, EF::ONE);
+
+        // Packed-eq path: hits the SIMD kernel from PR #1574.
+        group.bench_with_input(BenchmarkId::new("packed", label), &label, |b, _| {
+            let mut out = vec![EF::ZERO; 1 << (k_total - split_packed.num_variables())];
+            b.iter(|| split_packed.compress_suffix_into(&mut out, &poly));
+        });
+
+        // Unpacked-eq path: scalar reference for the same operation.
+        group.bench_with_input(BenchmarkId::new("unpacked", label), &label, |b, _| {
+            let mut out = vec![EF::ZERO; 1 << (k_total - split_unpacked.num_variables())];
+            b.iter(|| split_unpacked.compress_suffix_into(&mut out, &poly));
+        });
+    }
+
+    group.finish();
+}
+
+/// Prefix-variable fold into a scalar extension-field output.
+///
+/// Reaches the scalar-output prefix kernel; serves as a regression
+/// guard for the unmerged Phase 3 work.
+fn bench_compress_prefix(c: &mut Criterion) {
+    let mut group = c.benchmark_group("split_eq/compress_prefix");
+
+    // Each tuple is (k_total, eq_k, label):
+    //
+    // - midpoint cases (k_total / 2): mirror SVO claim build.
+    // - eq_k = 1 cases: mirror the per-round sumcheck shape.
+    let cases = [
+        (12usize, 6usize, "k12_eq6"),
+        (16, 8, "k16_eq8"),
+        (20, 10, "k20_eq10"),
+        (16, 1, "k16_eq1"),
+        (20, 1, "k20_eq1"),
+    ];
+
+    for &(k_total, eq_k, label) in &cases {
+        let mut rng = rng_for(label, k_total);
+        let poly = Poly::<F>::rand(&mut rng, k_total);
+        let point = Point::<EF>::rand(&mut rng, eq_k);
+
+        let split_packed = SplitEq::<F, EF>::new_packed(&point, EF::ONE);
+
+        group.bench_with_input(BenchmarkId::new("packed", label), &label, |b, _| {
+            b.iter(|| split_packed.compress_prefix(&poly));
+        });
+    }
+
+    group.finish();
+}
+
+/// Prefix-variable fold into a packed extension-field output.
+///
+/// # Role
+///
+/// - Hits the SIMD kernel introduced in this PR (Phase 2 target).
+/// - Production-shape caller in WHIR's per-round folding sumcheck.
+fn bench_compress_prefix_to_packed(c: &mut Criterion) {
+    let mut group = c.benchmark_group("split_eq/compress_prefix_to_packed");
+
+    // Same shape grid as the scalar-output bench above. Every shape
+    // here yields a packed output of at least one packed element.
+    let cases = [
+        (12usize, 6usize, "k12_eq6"),
+        (16, 8, "k16_eq8"),
+        (20, 10, "k20_eq10"),
+        (16, 1, "k16_eq1"),
+        (20, 1, "k20_eq1"),
+    ];
+
+    for &(k_total, eq_k, label) in &cases {
+        let mut rng = rng_for(label, k_total);
+        let poly = Poly::<F>::rand(&mut rng, k_total);
+        let point = Point::<EF>::rand(&mut rng, eq_k);
+
+        let split_packed = SplitEq::<F, EF>::new_packed(&point, EF::ONE);
+
+        group.bench_with_input(BenchmarkId::new("packed", label), &label, |b, _| {
+            b.iter(|| split_packed.compress_prefix_to_packed(&poly));
+        });
+    }
+
+    group.finish();
+}
+
+/// Single-row dot product of the eq table against a base-field row.
+///
+/// Reaches the `dot_with_base` kernel through the SplitEq evaluation pipeline.
+fn bench_dot_with_base(c: &mut Criterion) {
+    let mut group = c.benchmark_group("split_eq/dot_with_base");
+
+    let cases = [(12usize, "k12"), (14, "k14"), (16, "k16")];
+
+    for &(k_total, label) in &cases {
+        let mut rng = rng_for(label, k_total);
+        let poly = Poly::<F>::rand(&mut rng, k_total);
+        let point = Point::<EF>::rand(&mut rng, k_total);
+
+        let split_packed = SplitEq::<F, EF>::new_packed(&point, EF::ONE);
+
+        group.bench_with_input(BenchmarkId::new("packed", label), &label, |b, _| {
+            b.iter(|| split_packed.eval_base(&poly));
+        });
+    }
+
+    group.finish();
+}
+
+/// Single-row dot product against a pre-packed extension-field row.
+///
+/// Both sides of the dot are packed, so this is purely SIMD work.
+fn bench_dot_with_ext_packed(c: &mut Criterion) {
+    let mut group = c.benchmark_group("split_eq/dot_with_ext_packed");
+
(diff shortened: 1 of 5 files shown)


multilinear-util/benches/compress_kernels.rs
```diff
@@ -0,0 +1,192 @@
...
+        // Pre-build the split eq table outside the timing loop:
+        // construction cost is not part of the kernel we measure.
+        let split_unpacked = SplitEq::<F, EF>::new_unpacked(&point, EF::ONE);
+        let split_packed = SplitEq::<F, EF>::new_packed(&point, EF::ONE);
+
+        // Packed-eq path: hits the SIMD kernel from PR #1574.
+        group.bench_with_input(BenchmarkId::new("packed", label), &label, |b, _| {
+            let mut out = vec![EF::ZERO; 1 << (k_total - split_packed.num_variables())];
+            b.iter(|| split_packed.compress_suffix_into(&mut out, &poly));
+        });
```
camofu, line 57: really _minor_ tweak:
perhaps `iter_batched` would be slightly better here for bench-marking hygine, since  `compress_suffix_into` is `+=`, so the current pattern measures iteration 1 against a zero buffer and iterations 2..N against a buffer that already holds the previous result. Probably makes no difference.

Somthing like:
```rust
  b.iter_batched(
      || vec![EF::ZERO; ...], 
      |mut out| split_packed.compress_suffix_into(&mut out, &poly),
      BatchSize::SmallInput,
  );
```
Could be tried.

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -319,8 +295,293 @@ where
...
+    let mut tile_start = 0;
+    while tile_start < packed_inner {
+        let tile_len = TILE.min(packed_inner - tile_start);
+        let mut acc = [[F::Packing::ZERO; D]; TILE];
+
+        // Phase 2: walk every (w_idx, lane) pair contributing to the tile.
+        for w_idx in 0..n_packed {
+            // One packed * scalar extension multiply, reused across every
+            // lane batch and every output position in the tile below.
+            let pw = eq1_packed[w_idx] * w0;
```
camofu: This sits inside the outer tile loop, so for `packed_inner > TILE` it's redundantly recomputed `ceil(packed_inner / TILE)` times per `w_idx`. 

Hoisting once into a  `Vec<EF::ExtensionPacking>` of length `n_packed` (or a smaller scalars table keyed by (`w_idx`, `batch`, `k`)) trades a few KiB of stack/heap for `(ceil(packed_inner/TILE) - 1)*n_packed` saved packed-extension multiplies. 

Worth a follow-up bench, maybe in a different PR?

multilinear-util/benches/compress_kernels.rs
tcoratger: Good catch to flag, but the premise does not apply here: `SplitEq::compress_suffix_into` writes per slot with `*out = ...` (overwrite), not `*out += ...` (mod.rs:440,447). Each `b.iter` invocation overwrites every slot, so the previous iteration’s content is irrelevant.

The other two prefix-side benches (`compress_prefix`, `compress_prefix_to_packed`) call methods that return a fresh `Poly`, so the output buffer is allocated inside the timed function — same conclusion.

Leaving the bench as-is.

tcoratger:
Pushed a follow-up commit applying @CarloModicaPortfolio's hoist suggestion: `pw[w_idx] = eq1_packed[w_idx] * w0` is now pre-computed once into a stack-resident `MaybeUninit` buffer (sized to the existing `MAX_STACK_N = 128`) instead of being recomputed `ceil(packed_inner / TILE)` times per slot inside the tile loop. The dispatcher also gates on `n_packed <= MAX_STACK_N` so shapes that wouldn't fit get routed through the eager fallback.

Bench numbers below are measured on Apple M1 Pro / NEON, `RUSTFLAGS=\"-Ctarget-cpu=native\"`, single-threaded, 30 samples × 5 s on the kernel benches and 20 samples × 5 s on the WHIR benches.

### Kernel — `split_eq/compress_prefix_to_packed/packed` (vs prior commit on this PR)

| Shape    | before    | after     | change   | speedup |
|----------|----------:|----------:|---------:|--------:|
| `k12_eq6`   | 3.91 µs   | 3.26 µs   | **−16.6%** | 1.20× |
| `k16_eq8`   | 56.4 µs   | 46.9 µs   | **−17.8%** | 1.22× |
| `k20_eq10`  | 903 µs    | 723 µs    | **−20.0%** | 1.25× |

All `p = 0.00`.

### WHIR — `sumcheck_prover/svo` (end-to-end, vs prior commit on this PR)

| Config  | before    | after     | change   |
|---------|----------:|----------:|---------:|
| `small`   | 44.4 µs   | 42.5 µs   | **−5.0%** |
| `medium`  | 519 µs    | 482 µs    | **−7.5%** |
| `large`   | 8.14 ms   | 7.60 ms   | **−6.4%** |

All `p < 0.05`.

### Cumulative vs `main` (PR #1574 baseline + this PR's two commits)

| Bench                                            | main      | this PR   | total speedup |
|--------------------------------------------------|----------:|----------:|--------------:|
| `whir/sumcheck_prover/svo/small`                 | 47.7 µs   | 42.5 µs   | **1.12×** |
| `whir/sumcheck_prover/svo/medium`                | 576 µs    | 482 µs    | **1.20×** |
| `whir/sumcheck_prover/svo/large`                 | 10.4 ms   | 7.60 ms   | **1.37×** |
| `compress_prefix_to_packed/packed/k12_eq6`       | 5.08 µs   | 3.26 µs   | **1.56×** |
| `compress_prefix_to_packed/packed/k16_eq8`       | 73.0 µs   | 46.9 µs   | **1.56×** |
| `compress_prefix_to_packed/packed/k20_eq10`      | 1.14 ms   | 723 µs    | **1.58×** |

### Regression checks (vs `main`)

- `whir/sumcheck_prover/classic/{small,medium,large}` — bulk-fold path also fans into `compress_prefix_to_packed`, so it picks up the same speedup; no regression.
- `whir/svo_claim_build/k{12_l1..l4,16_l4,18_l4,20_l4}` — exercises the suffix kernel only; parity confirmed.
- `multilinear-util/eval_eq_batch` and `eval_eq_base_batch` (24 shapes each) — within criterion noise, no significant change.

The `iter_batched` suggestion (other review comment) was checked and dismissed: `compress_suffix_into` writes per slot with `*out = ...` (overwrite), not accumulate, so the reused buffer in the bench is fine. Reply on the inline thread.

multilinear-util/src/split_eq/packed_kernel.rs
tcoratger: Thanks a lot! Fixed

multilinear-util/benches/compress_kernels.rs
camofu: Oh sorry, you are right. I mixed `compress_suffix_into` with the accumulating kernels

camofu (approved):
Looks perfect to me!

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -319,8 +295,329 @@ where
...
+    // - Without this pre-pass: recomputed once per tile per w_idx.
+    // - With it: recomputed once per w_idx total.
+    // - Buffer is stack-resident; only the first n_packed slots are written.
+    let mut pw_buf = [const { MaybeUninit::uninit() }; MAX_STACK_N];
+    for (slot, &eq1_i) in pw_buf.iter_mut().zip(eq1_packed.iter()) {
+        slot.write(eq1_i * w0);
+    }
+    // SAFETY: The loop above wrote exactly:
+    // - the first `n_packed` slots,
+    // - the buffer is not aliased.
```
Nashtare: May be worth adding the same clauses (no aliasing, exposed prefix) as the `bufs` block, and also that `MaybeUninit` shares layout/alignment?

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -319,8 +295,329 @@ where
...
+    eq1_packed: &[EF::ExtensionPacking],
+    chunk: &[F],
+    w0: EF,
+) where
+    F: Field,
+    EF: ExtensionField<F>,
+{
+    let scalar_inner = out.len() * F::Packing::WIDTH;
+    chunk
+        .chunks(scalar_inner * F::Packing::WIDTH)
```
Nashtare, line 440: this will panic if `out.is_empty()`

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -446,10 +744,172 @@ mod tests {
...
+        let w0: EF = rng.random();
+
+        (eq1, chunk, w0)
+    }
+
+    proptest! {
+        #[test]
+        fn prefix_to_packed_kernel_matches_eager(
+            n_packed in 0usize..=8,
+            inner_k in 0usize..=4,
```
Nashtare: I think here we may want to bump `inner_k` up to 5? So that we have `packed_inner` that goes up to 32 covering multi-tile + tile boundary?

multilinear-util/src/split_eq/packed_kernel.rs
```diff
@@ -319,8 +295,329 @@ where
...
 
+/// Const-generic inner kernel for the prefix-fold packed-output path.
+///
+/// # Caller invariants
+///
+/// - `D == EF::DIMENSION`,
+/// - `F::Packing::WIDTH % CHUNK == 0`,
+/// - `eq1_packed.len() <= MAX_STACK_N` (bounds the pre-multiply stack buffer).
+#[inline]
+fn compress_prefix_to_packed_packed_kernel<F, EF, const D: usize>(
```
Nashtare: was the `inline` here meaningful? The function is quite heavy

Nashtare (approved):
No blocking comments, LGTM

multilinear-util/src/split_eq/packed_kernel.rs
tcoratger: Yep you are right, removed.

multilinear-util/src/split_eq/packed_kernel.rs
tcoratger: Done

multilinear-util/src/split_eq/packed_kernel.rs
tcoratger: just added an early return :)

multilinear-util/src/split_eq/packed_kernel.rs
tcoratger: Done

PR #1597, opened 2026-04-29
perf(whir): delayed Montgomery reduction in sumcheck round-coefficient kernel

tcoratger:
## Summary

Routes the whir sumcheck round-coefficient kernel through a chunked dot-product primitive that defers Montgomery reduction across `K = 8` widening multiplies. Net effect: fewer reductions per round on Monty-31 packings.

Algebra is bit-identical with `main`: same accumulators, same final values, just summed differently inside.

## Background

For `MontyField31` elements stored in Montgomery form, each `*` operator pays a Montgomery reduction (~1 native multiply + shift + conditional subtract). When computing `Σᵢ aᵢ · bᵢ`, classical eager arithmetic reduces once per multiply — `N` reductions for an `N`-term sum. If the running sum fits in a wider accumulator (e.g. `u64` lanes for BabyBear products), all `N` raw multiplies can be summed *without* reducing, then reduced **once** at the end. This is "delayed reduction."

Plonky3 already has the base-level primitive: `PackedMontyField31Neon::dot_product<N>` (and AVX2 / AVX512 siblings) accumulate up to 8 widening multiplies in 64-bit lanes and reduce once at the end. PR closes the two-layer gap above it.

### `field/src/extension/packed_binomial_extension.rs`

Specialise `Algebra<PF>::mixed_dot_product<N>` for `PackedBinomialExtensionField<F, PF, D>`. Extension multiplication by a base packing acts coordinate-wise, so:

```text
    (Σᵢ aᵢ · fᵢ).coord(k) = Σᵢ aᵢ.coord(k) · fᵢ      (linearity)
```

Each per-coordinate sum is a pure base-field dot product — exactly what `PF::dot_product::<N>` is built for. Reduction count drops from `N · D` (eager default) to `D`.

### `whir/src/sumcheck/strategy.rs`

Refactor `sumcheck_coefficients_prefix/suffix` from a streaming per-pair fold into a chunked loop that gathers `K = 8` paired evaluations per iteration and dispatches each tile through `mixed_dot_product`. The `half mod K` tail (≤ 7 pairs) falls back to the original streaming logic.

`K = 8` matches the deepest hand-written delayed-reduction primitive on every supported architecture (NEON `dot_product_8`, AVX2, AVX512).

### `whir/benches/sumcheck.rs`

Two new isolated kernel benches (`whir/kernel/round_coefficients_prefix/20` and `whir/kernel/fix_prefix_var_mut/20`) merged into the existing bench file. Old `whir/benches/sumcheck_kernels.rs` removed.

## Validation: are both pieces necessary?

Each layer alone is dead code. Tested on `whir/kernel/round_coefficients_prefix/20` (BabyBear + quartic binomial extension, packed, `n = 20`):

| Field override | Kernel refactor | Δ vs main |
|:-:|:-:|:-:|
| ✗ | ✗ | baseline |
| ✗ | ✓ | +1.17% (no change) |
| ✓ | ✗ | +0.73% (no change) |
| ✓ | ✓ | **−28.18%** |

The kernel refactor is the call-site that invokes `mixed_dot_product`; the field override gives that call something to do. Either alone is a no-op.

### Kernel microbenchmark (`whir/kernel/round_coefficients_prefix/20`)

| | baseline | PR | Δ |
|---|---|---|---|
| BabyBear+Ext4, packed, n=20 | 1.232 ms | 882 µs | **−28.4%** (p=0.00) |

### End-to-end (`whir/sumcheck_prover`, BabyBear+Ext4)

| Config | Δ vs main | p-value | Verdict |
|---|---|---|---|
| classic / small (n=14) | **−1.03%** | 0.00 | improved |
| svo / small | **−2.56%** | 0.00 | **performance improved** |
| classic / medium (n=18) | **−1.17%** | 0.00 | improved |
| svo / medium | **−1.31%** | 0.00 | **performance improved** |
| classic / large (n=22) | −0.55% | 0.20 | no change |
| svo / large | **−1.63%** | 0.01 | improved |

5 of 6 E2E configs significantly improved, none regressed. The kernel-level −28% dilutes to −1 to −2.5% E2E because the round-coefficient kernel is roughly 10–15% of total prover time; the rest (binding, eq construction, SVO accumulators, transcript) is unchanged here and is open territory for follow-up PRs that reuse the same primitive.

## Why E2E is bounded by the kernel share

```text
  reductions per round (N = 2^{n-1}, D = 4 for Ext4, K = 8):
     eager:    2 · D · N         = 4 194 304   at n=20
     delayed:  2 · D · (N / K)   =   524 288   at n=20
```

A factor-of-K drop in reductions translates to ~28% wall-clock on the kernel; non-kernel work (binding, eq, SVO, transcript) gates the E2E gain.

## Reproducing the numbers

```bash
# Kernel microbench
cargo bench -p p3-whir --bench sumcheck -- 'whir/kernel'

# End-to-end
cargo bench -p p3-whir --bench sumcheck -- 'whir/sumcheck_prover'
```

Bench host for the numbers above: Apple M4 Max, single-threaded on a performance core, `cargo bench` profile (no LTO).

## Test plan

- [x] `cargo build --workspace --release` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --workspace --all-targets --release` — clean
- [x] `cargo test -p p3-field --release` — all pass
- [x] `cargo test -p p3-baby-bear --release` — 209 / 209 pass
- [x] `cargo test -p p3-whir --release sumcheck` — 99 / 99 pass (includes proptests cross-checking against the reference round-coefficient implementation)
- [x] Kernel microbenchmark: ≥ −28% reproducible across 3 back-to-back runs
- [x] E2E benchmark: 5 of 6 configs significantly improved, none regressed

### DIFF
--- a/field/src/extension/packed_binomial_extension.rs
+++ b/field/src/extension/packed_binomial_extension.rs
@@ -85,6 +85,30 @@ impl<F: BinomiallyExtendable<D>, PF: PackedField<Scalar = F>, const D: usize>
 impl<F: BinomiallyExtendable<D>, PF: PackedField<Scalar = F>, const D: usize> Algebra<PF>
     for PackedBinomialExtensionField<F, PF, D>
 {
+    #[inline]
+    fn mixed_dot_product<const N: usize>(a: &[Self; N], f: &[PF; N]) -> Self
+    where
+        PF: crate::Dup,
+    {
+        // Output container; each coordinate is filled independently below.
+        let mut result = Self::default();
+
+        // One base-field dot product per output coordinate.
+        for k in 0..D {
+            // Strided gather of the k-th coordinate from each extension input:
+            //
+            //     coord_k = [ a_0[k], a_1[k], ..., a_{N-1}[k] ]
+            let coord_k: [PF; N] = core::array::from_fn(|i| a[i].value[k]);
+
+            // Base-level dot product.
+            //
+            // - For Monty-31 packings this is the delayed-reduction primitive;
+            // - For other packings it falls back to the eager default and the override is a no-op gain.
+            result.value[k] = PF::dot_product::<N>(&coord_k, f);
+        }
+
+        result
+    }
 }
 
 impl<F, PF, const D: usize> PrimeCharacteristicRing for PackedBinomialExtensionField<F, PF, D>
--- a/field/src/extension/packed_binomial_extension.rs
+++ b/field/src/extension/packed_binomial_extension.rs
@@ -88,7 +88,7 @@ impl<F: BinomiallyExtendable<D>, PF: PackedField<Scalar = F>, const D: usize> Al
     #[inline]
     fn mixed_dot_product<const N: usize>(a: &[Self; N], f: &[PF; N]) -> Self
     where
-        PF: crate::Dup,
+        PF: Dup,
     {
         // Output container; each coordinate is filled independently below.
         let mut result = Self::default();
--- a/field/src/extension/packed_binomial_extension.rs
+++ b/field/src/extension/packed_binomial_extension.rs
@@ -12,7 +12,7 @@ use serde::{Deserialize, Serialize};
 use super::{BinomialExtensionField, binomial_mul, vector_add, vector_sub};
 use crate::extension::{BinomiallyExtendable, binomial_square};
 use crate::{
-    Algebra, BasedVectorSpace, Field, PackedField, PackedFieldExtension, PackedValue, Powers,
+    Algebra, BasedVectorSpace, Dup, Field, PackedField, PackedFieldExtension, PackedValue, Powers,
     PrimeCharacteristicRing, field_to_array,
 };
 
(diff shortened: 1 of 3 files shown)


field/src/extension/packed_binomial_extension.rs
```diff
@@ -85,6 +85,30 @@ impl<F: BinomiallyExtendable<D>, PF: PackedField<Scalar = F>, const D: usize>
 impl<F: BinomiallyExtendable<D>, PF: PackedField<Scalar = F>, const D: usize> Algebra<PF>
     for PackedBinomialExtensionField<F, PF, D>
 {
+    #[inline]
+    fn mixed_dot_product<const N: usize>(a: &[Self; N], f: &[PF; N]) -> Self
+    where
+        PF: crate::Dup,
```
Nashtare: nit: 
```suggestion
        PF: Dup,
```

whir/src/sumcheck/strategy.rs
```diff
@@ -99,27 +198,56 @@ where
     // Precondition: paired slices must be aligned; adjacent pairs address the suffix bit.
     assert_eq!(evals.len(), weights.len());
```
Nashtare, line 200: unrelated to this PR, but shouldn't we also have a parity check on evals.len()?

Nashtare (approved):
Overall LGTM

whir/src/sumcheck/strategy.rs
tcoratger: Fixed as well while we are at it.

PR #1598, opened 2026-04-30
fix(whir): use unbiased rejection sampling for STIR query indices

DarkArtistry:
## TL;DR

WHIR's STIR query sampler was calling the slightly-biased [`CanSampleBits::sample_bits`](https://github.com/Plonky3/Plonky3/blob/main/challenger/src/lib.rs#L62-L71) primitive and then de-duplicating with `sort_unstable + dedup`. This PR swaps the call site to the unbiased [`CanSampleUniformBits::sample_uniform_bits::<true>`](https://github.com/Plonky3/Plonky3/blob/main/challenger/src/lib.rs#L78-L96) (which already exists in this repo and uses field-side rejection sampling) and replaces the post-hoc dedup with an explicit reject-and-retry loop. **`sample_bits` itself is unchanged** — it remains correct for callers like grinding/PoW where small bias is acceptable.

## The bug

Vs. the WHIR shift-query bound `ε^shift ≤ (1 − δ)^t` (Theorem 5.2 in Arnon–Chiesa–Fenzi–Yogev 2024), the bound is over `t` independent **uniformly-sampled** query positions; per-draw uniformity is what the bound requires.

`sample_bits` bit-decomposes a uniformly drawn field element. Values in `[m, p)`, where `m = ⌊p / 2^bits⌋ · 2^bits`, induce bias on the low `bits`. The trait's own docstring admits "may arise when bit-decomposing a uniformly sampled field element". Per-draw bias (TV distance from uniform) is bounded by `2^bits / |F|` — negligible on Goldilocks (~`2^{-44}` at 20 bits), and at most ~`2^{-11}` on KoalaBear at the same. The old code mitigated by capping `bits ≤ MAX_SAMPLE_BITS = 20`, but the bias was non-zero and inflated the effective `δ`.

## What changed (algorithmically)

Old hot loop in `whir/src/pcs/utils.rs` (3 batched branches around `sample_bits`, simplified):

```rust
let max_bits_per_call = (F::bits() - 1).min(MAX_SAMPLE_BITS);  // = 20
// ... pack queries into batched sample_bits calls ...
let mut all_bits = challenger.sample_bits(batch_bits);   // <-- biased
for _ in 0..batch_size {
    let q = (all_bits & mask) % folded_domain_size;
    queries.push(q);
    all_bits >>= domain_size_bits;
}
// ...
queries.sort_unstable();
queries.dedup();
```

New hot loop:

```rust
let target = num_queries.min(folded_domain_size);
let mut queries: Vec<usize> = Vec::with_capacity(target);
while queries.len() < target {
    let q = challenger
        .sample_uniform_bits::<true>(domain_size_bits)   // <-- unbiased
        .expect("Error impossible here due to resampling strategy");
    if !queries.contains(&q) {
        queries.push(q);
    }
}
queries.sort_unstable();
```

`MAX_SAMPLE_BITS` is deleted because the 20-bit cap was a mitigation for the bias that no longer exists.

## What `sample_uniform_bits::<true>` does differently

For each draw, [`DuplexChallenger`](https://github.com/Plonky3/Plonky3/blob/main/challenger/src/duplex_challenger.rs#L309-L327) rejects field elements that fall in `[m, p)` and resamples:

```rust
let mut result: F = challenger.sample();
while result.as_canonical_u64() >= m {  // <-- reject and resample
    result = challenger.sample();
}
// result is now uniform on [0, m), low `bits` are exactly uniform on [0, 2^bits)
```

Resample probability per draw is `≤ 2^bits / |F|` (and exactly `1/|F|` for KoalaBear/BabyBear/Goldilocks at typical `bits`, due to those primes' structure). The loop terminates in essentially one iteration. KoalaBear/BabyBear/Goldilocks/Mersenne31 all have `UniformSamplingField` impls, so this works out of the box.

The `.expect()` matches the existing pattern at [`grinding_challenger.rs:75-79`](https://github.com/Plonky3/Plonky3/blob/main/challenger/src/grinding_challenger.rs#L75-L79) (`check_witness_uniform`) — with `RESAMPLE = true` the underlying impl loops on rejection rather than erroring, so the `Err` arm is unreachable. Happy to switch to a `Resampling(ResamplingError)` variant on `FiatShamirError` + `?` propagation if reviewers prefer.

## Saturation: the `min(num_queries, folded_domain_size)` cap

In WHIR's final round the folded domain shrinks to ~1–4 positions while `final_queries` can be much larger (75 in `test_whir_end_to_end`). The implementation caps at `target = min(num_queries, folded_domain_size)`: when the requested count exceeds the folded domain, the verifier opens every position, which is the strongest possible verification at that round.

## Auxiliary change: `CanSampleUniformBits` for `SerializingChallenger`

The new bound on `WhirPcs`'s `MultilinearPcs` impl would have broken the WHIR keccak end-to-end test (`whir/src/pcs/tests.rs::keccak_tests::test_whir_keccak_end_to_end`), because `SerializingChallenger32` did not implement `CanSampleUniformBits<F>`. Added trivial impls for both `SerializingChallenger32` and `SerializingChallenger64`:

```rust
impl<F, Inner> CanSampleUniformBits<F> for SerializingChallenger32<F, Inner>
where
    F: PrimeField32,
    Inner: CanSample<u8>,
{
    fn sample_uniform_bits<const RESAMPLE: bool>(
        &mut self,
        bits: usize,
    ) -> Result<usize, ResamplingError> {
        Ok(self.sample_bits(bits))
    }
}
```

Justification: `SerializingChallenger32::sample_bits` is **already exactly uniform** (it draws bytes from a cryptographic-hash inner challenger and masks — no field-element decomposition, hence no `[m, p)` rejection band). The `RESAMPLE` flag is a no-op for this challenger family. The struct-level docstring already advertises "Avoids modulo bias".

Note: I did **not** add the impl for `MultiField32Challenger` (no in-tree user, so the workspace builds clean). Easy to add as a follow-up — same shape as above — if external users need it.

## Why scope it as a call-site swap rather than fix `sample_bits`?

`sample_bits` is documented as "reasonably close to uniform" and is used in places where that's fine — notably [`GrindingChallenger::check_witness`](https://github.com/Plonky3/Plonky3/blob/main/challenger/src/grinding_challenger.rs#L40-L46) for proof-of-work. Tightening `sample_bits` to mean "exactly uniform" would either silently change semantics for those callers (with perf implications: extra resampling on every PoW check) or require introducing a new method anyway. WHIR's query sampler is the wrong layer at which to make a workspace-wide trait-semantics change. Routing WHIR through the existing `sample_uniform_bits` is the surgical fix.

## Files changed

| File | Δ | Purpose |
| --- | --- | --- |
| `whir/src/pcs/utils.rs` | +258 / -96 | Function rewrite + 3-test proptest module + doc tightening. |
| `challenger/src/serializing_challenger.rs` | +56 / -2 | New `CanSampleUniformBits<F>` impls for `SerializingChallenger32/64` + doc tightening. |
| `whir/src/pcs/adapter.rs` | +6 / -3 | `+ CanSampleUniformBits<F>` on `MultilinearPcs` impl bound. |
| `whir/src/pcs/prover/mod.rs` | +2 / -2 | Same bound on `WhirProver` impl. |
| `whir/src/pcs/verifier/mod.rs` | +2 / -2 | Same bound on `WhirVerifier` impl. |

## Test plan

- [x] `cargo test -p p3-whir` — **186 tests pass**, including `pcs::tests::test_whir_end_to_end` and `pcs::tests::keccak_tests::test_whir_keccak_end_to_end` (the latter exercises the new `SerializingChallenger32` impl through the WHIR PCS).
- [x] `cargo test -p p3-challenger` — **38 tests pass**.
- [x] `cargo check --workspace --all-targets` — clean across the workspace (uni-stark, fri, circle, batch-stark, examples, …).
- [x] `cargo fmt --all` — no diff.
- [x] `cargo clippy -p p3-whir -p p3-challenger --all-targets -- -D warnings` — no warnings.
- [x] **New tests** in `whir/src/pcs/utils.rs::tests`:
  - `prop_get_challenge_stir_queries_invariants` — 64-case proptest covering length, in-range, strict sortedness, distinctness, and **determinism** (rebuild challenger with same seed → byte-identical output).
  - `saturates_when_num_queries_exceeds_domain` — explicit final-round regime: requesting 75 queries from a 4-position folded domain returns exactly `[0, 1, 2, 3]`.
  - `empirical_uniformity_single_query` — 4096 single-query draws into a 16-bucket histogram. Hoeffding tolerance `t = 160` chosen so per-run flake probability is `< 1.2 × 10⁻⁴` after union-bounding both tails over all buckets. **Regression detector**, not a uniformity certificate (the cryptographic claim rests on `sample_uniform_bits`'s contract).

## Soundness rationale (one paragraph)

WHIR's shift-query soundness (Theorem 5.2) requires `t` independent uniformly-sampled query positions. The previous sampler used `sample_bits`, whose bit-decomposition of a uniform field element introduces a per-draw bias bounded by `2^bits / |F|` — small but non-zero, which would inflate the effective `δ` in the bound. The fix is small but principled: route through the existing `sample_uniform_bits<true>` primitive (field-side rejection sampling, exactly uniform on `[0, 2^bits)`). The bias was small enough not to break security at the parameter sets shipped today, but closing the gap removes a real footnote in the soundness accounting and makes the implementation match the paper's bound at face value.

## Related

- [`leanEthereum/leanMultisig#184`](https://github.com/leanEthereum/leanMultisig/issues/184) — same biased-truncation pattern in lean's `sample_in_range`. Lean can adopt the same primitive in a follow-up PR.
- WHIR paper (Arnon, Chiesa, Fenzi, Yogev 2024), Section 2.1.3 / Theorem 5.2.

### DIFF
--- a/challenger/src/serializing_challenger.rs
+++ b/challenger/src/serializing_challenger.rs
@@ -8,8 +8,8 @@ use p3_util::log2_ceil_u64;
 use tracing::instrument;
 
 use crate::{
-    CanFinalizeDigest, CanObserve, CanSample, CanSampleBits, FieldChallenger, GrindingChallenger,
-    HashChallenger,
+    CanFinalizeDigest, CanObserve, CanSample, CanSampleBits, CanSampleUniformBits, FieldChallenger,
+    GrindingChallenger, HashChallenger, ResamplingError,
 };
 
 /// Given a challenger that can observe and sample bytes, produces a challenger that is able to
@@ -168,6 +168,23 @@ where
     }
 }
 
+/// The inner challenger draws uniform bytes (cryptographic hash output), so
+/// taking the low `bits` of a `u32` assembled from those bytes is exactly
+/// uniform on `[0, 2^bits)`. No field-side rejection is required and the
+/// `RESAMPLE` flag is irrelevant — sampling never errors.
+impl<F, Inner> CanSampleUniformBits<F> for SerializingChallenger32<F, Inner>
+where
+    F: PrimeField32,
+    Inner: CanSample<u8>,
+{
+    fn sample_uniform_bits<const RESAMPLE: bool>(
+        &mut self,
+        bits: usize,
+    ) -> Result<usize, ResamplingError> {
+        Ok(self.sample_bits(bits))
+    }
+}
+
 impl<F, Inner> GrindingChallenger for SerializingChallenger32<F, Inner>
 where
     F: PrimeField32,
@@ -322,6 +339,22 @@ where
         assert!((1u64 << bits) <= F::ORDER_U64);
         let rand_u64 = u64::from_le_bytes(self.inner.sample_array());
         (rand_u64 & ((1u64 << bits) - 1)) as usize
+    }
+}
+
+/// See [`CanSampleUniformBits`] for `SerializingChallenger32` — the same
+/// reasoning applies here: uniform inner bytes give uniform low bits with
+/// no field-side rejection.
+impl<F, Inner> CanSampleUniformBits<F> for SerializingChallenger64<F, Inner>
+where
+    F: PrimeField64,
+    Inner: CanSample<u8>,
+{
+    fn sample_uniform_bits<const RESAMPLE: bool>(
+        &mut self,
+        bits: usize,
+    ) -> Result<usize, ResamplingError> {
+        Ok(self.sample_bits(bits))
     }
 }
 
--- a/challenger/src/serializing_challenger.rs
+++ b/challenger/src/serializing_challenger.rs
@@ -168,19 +168,29 @@ where
     }
 }
 
-/// The inner challenger draws uniform bytes (cryptographic hash output), so
-/// taking the low `bits` of a `u32` assembled from those bytes is exactly
-/// uniform on `[0, 2^bits)`. No field-side rejection is required and the
-/// `RESAMPLE` flag is irrelevant — sampling never errors.
 impl<F, Inner> CanSampleUniformBits<F> for SerializingChallenger32<F, Inner>
 where
     F: PrimeField32,
     Inner: CanSample<u8>,
 {
+    /// Sample uniform bits by masking bytes from the inner stream.
+    ///
+    /// # Overview
+    ///
+    /// The inner stream emits cryptographic-hash bytes uniform on `[0, 2^8)`.
+    ///
+    /// Reading 4 bytes as a 32-bit integer and masking the low `bits` is
+    /// exactly uniform on `[0, 2^bits)`.
+    ///
+    /// No field-element decomposition occurs, so no rejection band exists.
+    /// The const generic is therefore inert: this function never errors
+    /// and never resamples.
     fn sample_uniform_bits<const RESAMPLE: bool>(
         &mut self,
         bits: usize,
     ) -> Result<usize, ResamplingError> {
+        // Byte-sourced sampling is uniform without rejection, so the
+        // result is always valid and the error arm is unreachable.
         Ok(self.sample_bits(bits))
     }
 }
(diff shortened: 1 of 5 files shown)


DarkArtistry:
Closing — opening this prematurely. Will reopen after the contributor has had a chance to vet the patch locally. Apologies for the noise.

DarkArtistry:
Reopening: contributor has vetted the diff locally. Updated the PR body with a clearer explanation of the change scope (call-site swap, not a primitive rewrite).

DarkArtistry:
Pushed `de2beb1b` updating the doc per the review feedback — `(1 − δ)^t` is over independent uniformly-sampled positions, not `t` distinct ones, so the dedup-shrinkage framing was incorrect. Per-draw uniformity is now the only soundness leak called out; duplicate rejection is described as a cleanliness / saturation-handling choice. PR body updated to match. Ready for another look.

tcoratger (approved):

PR #1600, opened 2026-04-30
Add carry-critical NEON dot product regression tests

mo-melvin77:
Add carry-critical regression tests for NEON dot_product_5 and dot_product_8, ensuring packed outputs match scalar references at overflow boundaries.


tcoratger (approved):
Thanks a lot for flagging this!

tcoratger:
@Nashtare Can you double check this one? 

As a preventive measure I've added other dot product edge case tests, this kind of scenario is incredibly difficult to detect for a human...

Nashtare (approved):

PR #1603, opened 2026-05-02
feat(challenger): typed Fiat-Shamir transcript layer (DSFS / IETF draft)

tcoratger:
## Summary

This PR adds a new `challenger/src/fs/` module — a typed Fiat-Shamir transcript framework that sits **on top of** the existing `CanObserve` / `CanSample` challengers. It implements the duplex-sponge Fiat-Shamir transformation analyzed in Chiesa & Orrù (2026), aligned with the in-progress IETF CFRG draft and structurally compatible with the [`spongefish`](https://github.com/arkworks-rs/spongefish) reference library.

The existing challengers (`DuplexChallenger`, `Shake128`, `MultiFieldChallenger`, …) are **unchanged and reused**. This PR adds a wrapper that provides typed pattern enforcement, deterministic domain separation, prover/verifier drivers, and a portable wire format (NARG).

## References

- **Paper** — Chiesa & Orrù, *A Fiat-Shamir Transformation From Duplex Sponges* (eprint 2025/536, March 2026): https://eprint.iacr.org/2025/536.pdf
- **IETF draft** — `draft-irtf-cfrg-fiat-shamir-02`: https://datatracker.ietf.org/doc/draft-irtf-cfrg-fiat-shamir/
- **Reference implementation** — arkworks-rs/spongefish: https://github.com/arkworks-rs/spongefish (the structural model this PR follows)
- Related upstream PRs in spongefish: arkworks-rs/spongefish#61, arkworks-rs/spongefish#62

## Why this exists

Most deployed Fiat-Shamir transformations differ in subtle ways from what theorists analyze:

- Practical hash functions are fixed-size; protocol messages are not.
- The duplex-sponge mode of operation is what implementations actually use, but its security for knowledge soundness and zero knowledge is not implied by the usual indifferentiability guarantee — it requires a separate analysis (Theorems 1 and 2 of the paper).
- Hand-rolled transcripts often forget domain separation, mix up types between prover and verifier, or have implicit length conventions on the wire.

The DSFS transformation in the paper closes this gap by:

1. Specifying exactly how prover messages get absorbed and verifier challenges get squeezed through a duplex sponge.
2. Pinning down the seeding (protocol identifier + instance label) so two protocols cannot collide.
3. Bounding the bias from byte-to-field decoding (Lemma C.1 / IETF §6).
4. Proving soundness, knowledge soundness, and zero knowledge against `t`-query adversaries.

This PR is a faithful Rust implementation of that machinery, designed to plug into Plonky3 protocols without disturbing the existing challenger zoo.

## Architecture

The new layer wraps the existing sponge — it does not replace it.

```
┌───────────────────────────────────────────┐
│  ProverState / VerifierState              │  ← drives the protocol
│  (pattern player + NARG buffer + codec)   │
├───────────────────────────────────────────┤
│  CanObserve / CanSample (existing trait)  │  ← byte / field interface
├───────────────────────────────────────────┤
│  Shake128, DuplexChallenger, …            │  ← the actual sponge
└───────────────────────────────────────────┘
```

### Module layout

```
challenger/src/fs/
├── mod.rs                 — top-level re-exports
├── unit.rs                — `Unit` trait abstracting the sponge alphabet
├── error.rs               — `TranscriptError` + structured info payloads
├── shake128.rs            — IETF §8.1 streaming SHAKE128 adapter (KATs A.1–A.9)
├── domain_separator.rs    — protocol id + instance label + pattern fingerprint
├── pattern/
│   ├── mod.rs             — `Pattern` trait
│   ├── step.rs            — `Interaction`, `Hierarchy`, `Kind`, `Length`, `Label`
│   ├── sequence.rs        — `InteractionPattern` (validated step sequence)
│   ├── state.rs           — `PatternState` (recorder)
│   └── player.rs          — `PatternPlayer` (replayer)
├── state/
│   ├── mod.rs
│   ├── prover.rs          — `ProverState` driver + NARG accumulator
│   └── verifier.rs        — `VerifierState` driver + NARG cursor
└── codecs/
    ├── mod.rs
    ├── codec.rs           — `Codec<C, T>` trait
    ├── decode_field.rs    — IETF `Ns + 16` byte-to-field decoding
    ├── bytes_to_field.rs  — byte-sponge ↔ prime-field-scalar codec
    ├── field_to_field.rs  — identity codec for native-field sponges
    ├── field_to_bytes.rs  — field-sponge → bytes (with documented truncation bias)
    └── extension.rs       — base codec lifted to extension fields
```

### What each piece does

- **`InteractionPattern`** — a validated, hash-stable list of every typed step a transcript will perform, recorded once per protocol. Each step records its hierarchy (`Atomic` / `Begin` / `End`), kind (`Public`, `Message`, `Hint`, `Challenge`, `Pow`, `Salt`, `Protocol`), label, Rust type, and length. The 32-byte fingerprint of a pattern (Keccak-256 over a hash-stable rendering) goes into the seed so two protocols differing only in shape cannot collide.

- **`DomainSeparator`** — a 64-byte protocol identifier, a variable-length instance label, and a (terminator-closed) seeding routine producing `[protocol_id | len_be | label | DOMAIN_TAG]`. Closely tracks IETF draft §3.

- **`ProverState` / `VerifierState`** — the drivers. Both are constructed with a challenger and a domain separator, walk the recorded pattern step by step, and either accumulate (`prover.finalize() -> Vec<u8>`) or consume (`verifier.finalize() -> Result<()>`) the NARG byte payload. Type and length mismatches between recorder and player surface immediately as panics with full diffs; wire-format problems return structured `TranscriptError`s.

- **`Codec<C, T>`** — stateless adapters that absorb / sample a typed value through a generic challenger. Codecs are zero-sized types passed at the call site so different absorb/sample strategies coexist in the same transcript:
  - `BytesToFieldCodec<F>`: 128-bit-clean field sampling from a byte sponge (IETF `Ns + 16` rule, DSFS Lemma C.1, bias ≤ 2^-128 for any prime).
  - `FieldToFieldCodec<F>`: identity codec for sponges whose alphabet is the field itself.
  - `FieldToBytesCodec<F>` / `FieldToBytesCodec64<F>`: byte sampling from a field sponge, with documented truncation bias (~2^-23 for 31-bit fields, ~2^-56 for 64-bit fields — *not* cryptographic-grade; a warning is in the doc).
  - `ExtensionFieldCodec<F, EF, BaseCodec>`: generic lifting from a base-field codec to an extension-field codec.

- **`Shake128`** — a streaming SHAKE128 adapter that exactly matches the IETF §8.1 reference (validated against KATs A.1 through A.9). Documented as a *non-duplex* adapter — for true incremental absorb-permute-squeeze, callers continue to use the existing `DuplexChallenger`.

## Usage example

A Schnorr-shaped sigma protocol — three steps, end-to-end:

```rust
use p3_baby_bear::BabyBear;
use p3_challenger::fs::{
    DomainSeparator, Hierarchy, Interaction, InteractionPattern, Kind, Length,
    PatternState, ProverState, VerifierState,
    codecs::BytesToFieldCodec,
    shake128::Shake128,
};

// 1. Record the pattern once.
fn schnorr_pattern() -> InteractionPattern {
    let mut s = PatternState::<u8>::new();
    s.interact(Interaction::new::<BabyBear>(
        Hierarchy::Atomic, Kind::Message, "commitment", Length::Scalar,
    ));
    s.interact(Interaction::new::<BabyBear>(
        Hierarchy::Atomic, Kind::Challenge, "challenge", Length::Scalar,
    ));
    s.interact(Interaction::new::<BabyBear>(
        Hierarchy::Atomic, Kind::Message, "response", Length::Scalar,
    ));
    s.finalize()
}

// 2. Bind protocol identifier + instance + pattern fingerprint.
fn domain() -> DomainSeparator<u8> {
    let mut ds = DomainSeparator::new(0x01, b"schnorr-v1", schnorr_pattern());
    ds.bind_pattern_hash();
    ds
}

// 3. Drive prover and verifier through the same pattern.
let ds = domain();

let mut prover = ProverState::<_, u8>::new(Shake128::new(&[0u8; 64]), &ds);
prover.add_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("commitment", &commitment);
let c = prover.challenge_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("challenge");
prover.add_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("response", &response);
let narg = prover.finalize();

let mut verifier = VerifierState::<_, u8>::new(Shake128::new(&[0u8; 64]), &ds, &narg);
let read_commitment = verifier.next_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("commitment")?;
let c_v = verifier.challenge_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("challenge");
let read_response = verifier.next_scalar::<BabyBear, BytesToFieldCodec<BabyBear>>("response")?;
verifier.finalize()?;
```

## Properties this layer enforces

- **Prover and verifier cannot drift.** A label, kind, type, or length mismatch surfaces as a panic at the offending call site, with a diff between expected and observed.
- **No length confusion on the wire.** The recorded pattern is the source of truth; lengths are not trusted from the wire.
- **Domain separation by construction.** Two protocols with different shapes get different seeds via the pattern fingerprint.
- **Typed message vocabulary.** `Hint` bytes hit the wire but never the sponge; `Salt` bytes affect challenges; `Pow` is a typed proof-of-work step; `Challenge` is sampled, not read.
- **Strict drop-time checks.** Forgetting to call `finalize()` is a loud panic during development, surfacing missing-step bugs.

## Test coverage

- 115 tests across the new module, plus the 9 existing IETF SHAKE128 KATs (A.1–A.9).
- All error-payload tests do full `assert_eq!` on the structured info struct (not just `matches!`), so position counters and reason strings are pinned down.
- Property tests on canonical decoding range and sampling determinism.
- Round-trip tests through the full `ProverState` → wire bytes → `VerifierState` pipeline.

## Migration / impact

- **Zero impact on existing code.** Nothing in `challenger/` outside the new `fs/` module is modified except `lib.rs` (one `pub mod fs;` line) and `Cargo.toml` (added `p3-keccak`, `num-bigint`, `thiserror`, `tiny-keccak`, plus dev deps).
- The existing `DuplexChallenger`, `Shake128`-based `HashChallenger`, etc. continue to work as before. Protocols not yet using the typed transcript layer keep their current API.

## Departures from the IETF draft (intentional, documented)

- **`Kind` extensions:** `Pow` and `Salt` extend the spongefish vocabulary to fit Plonky3-specific protocol primitives. They are typed at the pattern level and behave consistently between recorder and player.
- **Seeding format:** uses a 4-byte big-endian length prefix and a `0x80` domain-tag terminator, where the IETF draft §5 currently uses a 32-byte SHAKE-derived `session_id` prepended to the protocol id. The Plonky3 form is a pragmatic simplification that preserves the same invariants (no two valid seeds are prefixes of each other, distinct labels yield distinct seeds). Worth aligning with the IETF before any wire-format commitment outside Plonky3.
- **`Shake128` adapter:** matches IETF §8.1 (the streaming "hash whole input each squeeze" reference), not a true duplex sponge. Documented as such; for real duplex semantics, callers use `DuplexChallenger`.

## Not yet implemented (follow-ups)

- A native Keccak-f[1600] duplex sponge (currently only the streaming SHAKE128 adapter exists in the `fs/` namespace).
- Concrete elliptic-curve codecs (a `EllipticCurveCodec<G>` marker stub was removed from this PR; the IETF §9.1 binary and algebraic codecs are TBD).
- Universal composability analysis in the global random permutation model (open problem from the paper, §1.3).

## Test plan

- [ ] CI: `cargo test -p p3-challenger`
- [ ] CI: `cargo clippy -p p3-challenger --all-targets -- -D warnings`
- [ ] Spot-check IETF KATs (`fs::shake128::tests::ietf_a*`) against the test vectors in IETF draft §A.
- [ ] Confirm the round-trip test (`prover_round_trip_with_shake128_and_bytes_to_field_codec`) passes.
- [ ] Review the seeding format vs. IETF draft §5 — flag if upstream alignment is required before merging.
- [ ] Review the `Kind` extensions (`Pow`, `Salt`) for naming and semantics.


challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,468 @@
...
+            let bytes = encode_field_be::<F>(coeff);
+            self.narg.extend_from_slice(&bytes);
+        }
+    }
+
+    /// Append a hint to the wire buffer without absorbing it into the sponge.
+    ///
+    /// Hint bytes are part of the wire format but never enter the sponge,
+    /// so they cannot influence sampled challenges.
+    pub fn add_hint(&mut self, label: Label, bytes: &[u8]) {
```
adr1anh: I haven't read through the spec or paper yet, but from what I understand, hints must always have a fixed length. 

In our implementation of batch Merkle tree proofs, we include

latifkasuli:
This looks directionally very strong. I like the layering: typed transcript patterns over the existing CanObserve/CanSample challenger abstraction, without forcing an immediate rewrite of existing challengers.

The main things I'd want settled before protocols depend on this are:

1. Is the seeding/domain-separation format intended to remain Plonky3-internal for now, or should it align with the IETF draft before merge? Since the pattern hash + NARG format can become a protocol commitment, this seems worth deciding early.

2. The Shake128 adapter is useful, but the PR notes that it is not a true duplex sponge. Should the API/examples make that distinction more explicit so users do not accidentally assume full DSFS/SpongeFish semantics?

3. Pattern mismatches seem to be enforced with panics in some paths, while malformed NARG data returns TranscriptError. Is that split intentional? For verifier-facing proof parsing, structured errors may be safer than panic-based enforcement.

4. FieldToBytesCodec documents truncation bias. Should it be renamed, feature-gated, or otherwise made harder to use accidentally for cryptographic challenge material?

Overall I like the direction and the test coverage. These are mostly security-boundary/API-stability questions rather than objections.

recmo:
> Pattern mismatches seem to be enforced with panics in some paths, while malformed NARG data returns TranscriptError. Is that split intentional? For verifier-facing proof parsing, structured errors may be safer than panic-based enforcement.

Yes, intentional. Prover or verifier missing the pattern is _always_ an implementation bug (and bugs should panic as we are in undefined state and proceeding can be dangerous). Malformed input is never a panic as the verifier must validate the input.

That being said, I ended up dropping the whole type-system-enforced pattern mechanism. It's too much boilerplate for what it protects against. Transcript mismatches are caught well with a completeness prop-test. Not perfect, but good enough that it hasn't been a concern.

A much bigger concern that the type system can help with is managing which values are bound to the transcript and which aren't, especially when you make transcript-binding the callers responsibility for efficiency. Callers neglecting this responsibility has been the number one source of soundness issues in the first audit of ProveKit.

latifkasuli:
That makes sense. I agree malformed verifier input should remain fallible, while local pattern divergence can reasonably be treated as an implementation bug.

The transcript-binding point seems even more important. Would it make sense for this layer to provide APIs that make "observed but not bound" states hard to express, or at least make binding obligations explicit at the type/API boundary? That feels like the place where the type system could give the most security value without forcing a full typed-pattern protocol.

adr1anh:
I haven't read through the spec or paper yet, but from what I understand, hints must always have a fixed length. 

In our implementation of batch Merkle tree proofs, we only include the strict list of authentication leaves required to recompute the root. After sorting the leaf indices, we reconstruct each partial layer by either using an available digest, or getting the next one from the transcript.

To use this new transcript framework, we could specify the maximum number of hints, and just pad with empty digests. From some rough simulations, I found that the average number of required authentication digests is about 10% less than in the worst case. 

The following table shows the number of padding digests we would need on average for a given number of queries. Note that this is independent of the tree height.

| num_queries | worst−average | size saved |
|---:|---:|---:|
| 16 | 13.6 | 0.43 KB |
| 32 | 27.8 | 0.87 KB |
| 64 | 56.1 | 1.75 KB |
| 128 | 112.5 | 3.52 KB |

On a typical Miden proof, we'd waste between 5kb and 20kb. While not huge, it still represents around 10% of the proof. 

One solution would be to annotate a `Hint` slice with a _maximum_ length, and then prepend the actual length of the slice in the proof. This still allows us to derive a maximum size for the proof, while avoiding unnecessary zero padding.

tcoratger:
> I haven't read through the spec or paper yet, but from what I understand, hints must always have a fixed length.
> 
> In our implementation of batch Merkle tree proofs, we only include the strict list of authentication leaves required to recompute the root. After sorting the leaf indices, we reconstruct each partial layer by either using an available digest, or getting the next one from the transcript.
> 
> To use this new transcript framework, we could specify the maximum number of hints, and just pad with empty digests. From some rough simulations, I found that the average number of required authentication digests is about 10% less than in the worst case.
> 
> The following table shows the number of padding digests we would need on average for a given number of queries. Note that this is independent of the tree height.
> 
> num_queries	worst−average	size saved
> 16	13.6	0.43 KB
> 32	27.8	0.87 KB
> 64	56.1	1.75 KB
> 128	112.5	3.52 KB
> On a typical Miden proof, we'd waste between 5kb and 20kb. While not huge, it still represents around 10% of the proof.
> 
> One solution would be to annotate a `Hint` slice with a _maximum_ length, and then prepend the actual length of the slice in the proof. This still allows us to derive a maximum size for the proof, while avoiding unnecessary zero padding.

Hey, 

Thanks a lot for the comment, I've fixed that in the latest commit, please let me know if there is something else missing.

tcoratger:
Here we have a bunch of follow up PRs for when this one is merged, prepared by @kilic:
- https://github.com/tcoratger/Plonky3/pull/5
- https://github.com/tcoratger/Plonky3/pull/7
- https://github.com/tcoratger/Plonky3/pull/8
- https://github.com/tcoratger/Plonky3/pull/9
- https://github.com/tcoratger/Plonky3/pull/11
- https://github.com/tcoratger/Plonky3/pull/12

challenger/src/fs/shake128.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+impl CanSample<u8> for Shake128 {
+    fn sample(&mut self) -> u8 {
+        // No "sample one byte" primitive in the IETF reference, so route through a 1-byte squeeze.
+        //
+        // Bulk callers should use `squeeze_into_vec` directly to avoid the per-byte rebuild cost.
+        self.squeeze_into_vec(1)[0]
+    }
+}
```
Nashtare: I think that's a fairly risky API, especially as the indication is not visible in the rustdoc. Consecutive challenges from the same state are identical. I think we'd need some squeezing logic in the sampling.

challenger/src/fs/domain_separator.rs
```diff
@@ -0,0 +1,246 @@
...
+        // Index 0 holds the version, leaving 63 bytes for the name.
+        assert!(
+            protocol_name.len() < PROTOCOL_ID_LEN,
+            "protocol_name must fit in {} bytes (version byte takes the first slot)",
+            PROTOCOL_ID_LEN - 1,
+        );
+        // [version | name | zero padding].
+        let mut protocol_id = [0u8; PROTOCOL_ID_LEN];
+        protocol_id[0] = version;
+        protocol_id[1..1 + protocol_name.len()].copy_from_slice(protocol_name);
```
Nashtare, line 78: there can be length-extension attacks / issues here. I think we'd want to encode the length too as suffix of the protocol_id.

challenger/src/fs/domain_separator.rs
```diff
@@ -0,0 +1,246 @@
...
+    ///
+    /// When the protocol name does not fit in 63 bytes.
+    #[must_use]
+    pub fn new(version: u8, protocol_name: &[u8], pattern: InteractionPattern) -> Self {
+        // Index 0 holds the version, leaving 63 bytes for the name.
+        assert!(
+            protocol_name.len() < PROTOCOL_ID_LEN,
+            "protocol_name must fit in {} bytes (version byte takes the first slot)",
+            PROTOCOL_ID_LEN - 1,
+        );
```
Nashtare, line 74: I think we should probably check for empty IDs as well?

challenger/src/fs/domain_separator.rs
```diff
@@ -0,0 +1,246 @@
...
+        C: CanObserve<u8>,
+    {
+        challenger.observe_slice(&self.protocol_id);
+        // Length prefix prevents two labels of different lengths from colliding.
+        let len = self.instance_label.len() as u32;
+        challenger.observe_slice(&len.to_be_bytes());
+        challenger.observe_slice(&self.instance_label);
+        // Terminator stops later absorbs from extending the seed.
+        challenger.observe(DOMAIN_TAG);
+    }
```
Nashtare: If I understand correctly, if one forgets to call `bind_pattern_hash()` then we don't have the pattern commitment in the seed. On the other hand it can be called N times which changes the seed every time. I think that's another API issue: it should be automatic, or if we want it that way, we should make it so that the API makes unbound separators properly impossible to use for transcripts (but I think the first approach is better IMO)

challenger/src/fs/pattern/step.rs
```diff
@@ -0,0 +1,323 @@
...
+impl Display for Interaction {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        if f.alternate() {
+            // Alternate mode: hash-stable form.
+            //
+            // Length-prefix the label so adjacent labels do not collapse.
+            // The type name is intentionally omitted here.
+            write!(f, "{} {}", self.hierarchy, self.kind)?;
+            write!(f, " {} {}", self.label.len(), self.label)?;
+            write!(f, " {}", self.length)
```
Nashtare: QQ: why do we ignore type_name here?

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,806 @@
...
+            Hierarchy::Atomic,
+            Kind::Salt,
+            "salt",
+            Length::Fixed(salt_bytes.len()),
+        ));
+        // Absorb so future samples depend on the salt.
+        self.challenger.observe_slice(salt_bytes);
+        // Mirror the same bytes onto the wire so the verifier can re-absorb them.
+        self.narg.extend_from_slice(salt_bytes);
+    }
```
Nashtare, line 129: QQ: why is there no API for adding `Kind::Public`?

challenger/src/fs/codecs/field_to_bytes.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// Sample takes the low byte of a uniform field element, which biases by `~1/p`.
+///
+/// For 128-bit-clean output, route via a byte sponge instead.
+pub struct FieldToBytesCodec<F>(PhantomData<F>);
+
+impl<C, F> Codec<C, u8> for FieldToBytesCodec<F>
+where
+    F: PrimeField32,
+    C: CanObserve<F> + CanSample<F>,
```
Nashtare: I'm a bit confused about the use of `FieldToBytesCodec`. As far as I understand the code, the prover/verifier APIs only accept / can construct challengers that implement `CanObserve<u8>`, not `CanObserve<F>`.

challenger/src/fs/state/verifier.rs
```diff
@@ -0,0 +1,675 @@
...
+                reason: "message length exceeds declared maximum",
+            });
+        }
+        // Bind the count into the sponge before any value enters it.
+        //
+        // This keeps the transcript prefix-free, matching CO25 §6.2.
+        self.challenger.observe_slice(&len_bytes);
+        // One canonical field encoding per scalar.
+        let need = field_byte_size::<F>();
+        let mut out = Vec::with_capacity(actual);
```
Nashtare: nit: I think we need to check the length first out of caution:


```suggestion
        assert!(actual * need <= self.remaining_narg());
        let mut out = Vec::with_capacity(actual);
```

challenger/src/fs/codecs/decode_field.rs
```diff
@@ -0,0 +1,179 @@
...
+        bytes.len() >= need,
+        "decode_field_via_extra_bytes needs at least {need} bytes for this field, got {}",
+        bytes.len(),
+    );
+    // Read the prefix as big-endian, reduce mod p, lift back to F.
+    let big = BigUint::from_bytes_be(&bytes[..need]);
+    let order = F::order();
+    let reduced = big % order;
+    biguint_to_field::<F>(&reduced)
+}
```
Nashtare: Do we really need `BigUint`? This overall (in `bytes_to_field` module too) looks particularly heavy. Or at least have an override for `PrimeField32 / PrimeField64` to amortize this.

Nashtare: Especially as this looks like a hot path

challenger/src/fs/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    InvalidKindInfo, MismatchedBeginEndInfo, MissingBeginInfo, MissingEndInfo, PatternMismatchInfo,
+    TranscriptError,
+};
+pub use pattern::{
+    Hierarchy, Interaction, InteractionPattern, Kind, Label, Length, Pattern, PatternPlayer,
+    PatternState,
+};
+pub use shake128::Shake128;
+pub use state::{ProverState, VerifierState};
+pub use unit::Unit;
```
Nashtare: Applying to the entirety of the codebase, we can probably reduce the public visibility a LOT across

Nashtare (changes requested):
I'm mostly concerned about the overall API which I think contains a couple footguns / easy-to-mess-up constructions.

challenger/src/fs/shake128.rs
tcoratger: Ok to simplify even more I've completely removed this file because this was just an example for the IETF but in practice we never use this, so I've just removed, we can instantiate we anything we want.

challenger/src/fs/domain_separator.rs
tcoratger: Done

challenger/src/fs/domain_separator.rs
tcoratger: Yes done and I've added a comment to explain what we do: `[version | name | zero padding | name_len]`

challenger/src/fs/domain_separator.rs
tcoratger: Ok you are right, just solved that by removing the method. The pattern fingerprint is now folded into `seed_bytes` unconditionally. and so this is impossible to forget or double-apply.

challenger/src/fs/pattern/step.rs
tcoratger: I've just documented that into the code
image

challenger/src/fs/state/prover.rs
tcoratger: Just added `add_public_scalar`/`add_public_scalars` (prover) and `observe_public_scalar`/`observe_public_scalars` (verifier).

challenger/src/fs/codecs/field_to_bytes.rs
tcoratger: Just removed the file for now since this was unused, we can check later if needed.

challenger/src/fs/state/verifier.rs
tcoratger: Fixed

challenger/src/fs/mod.rs
tcoratger: Yes solved

challenger/src/fs/codecs/decode_field.rs
tcoratger: Fixed and documented without biguint

tcoratger:
@Nashtare What do you think about this one? Do we still want to merge it?

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+        C: CanObserve<u8>,
+    {
+        // Seed first so two distinct (protocol, instance) pairs land on distinct sponge states.
+        ds.seed_bytes(&mut challenger);
+        let player = PatternPlayer::new(ds.pattern().clone());
+        Self {
+            challenger,
+            player,
+            narg: Vec::new(),
+            _u: core::marker::PhantomData,
```
Nashtare: **[HIGH] No production challenger can back this layer.** `new` requires `C: CanObserve<u8>`. On `main` the only type implementing that is `HashChallenger<u8, H, N>`: `DuplexChallenger`, `MultiField32Challenger` and `SerializingChallenger32/64` implement `CanObserve<F>` (plus digest/cap types) but not `CanObserve<u8>` (the serializing ones only *require* it of their `Inner`). Consequences:

- `pow` / `check_pow` additionally need `GrindingChallenger`, which is implemented only by `DuplexChallenger`, `MultiField32Challenger` and `SerializingChallenger32/64`. The two bounds are jointly unsatisfiable by any in-repo type, which is also why the PoW path has zero tests.
- `FieldToFieldCodec` needs `C: CanObserve<F> + CanSample<F>` on a driver whose constructor needs `CanObserve<u8>`; no in-repo type satisfies both, so the field-native codec is unreachable through the drivers.
- The driver takes `C` by value with no way to get it back, and there is no `GrindingChallenger for &mut C` blanket impl, so it cannot be interleaved with the existing `&mut Challenger` APIs either.

So the description's claim that the existing challengers are "unchanged and reused" only holds for the Keccak hash chain, which is also not the duplex mode the DSFS theorems analyze. Before merging we need at least one route for `DuplexChallenger` (a field-unit seeding path, or `CanObserve<u8>` on the serializing challengers) and a test that actually exercises `pow` / `check_pow`.

challenger/src/fs/bound.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+
+    /// Lift a deterministic derivation to bound outputs.
+    ///
+    /// `f` must be a pure function of `T`;
+    ///
+    /// Non-determinism inside `f` silently launders the binding.
+    pub fn map<U>(self, f: impl FnOnce(T) -> U) -> TranscriptBound<U> {
```
Nashtare: **[HIGH] `TranscriptBound` is trivially launderable, which defeats its stated purpose.** `map` and `combine_with` accept arbitrary closures, so any holder of *any* bound value can mint a witness for an unbound one: `bound.map(|_| forged)` type-checks and never touches the transcript. The doc only warns about non-deterministic closures, but a pure closure that ignores its input launders just as silently, including by accident in a refactor that derives a value from context instead of from the bound input.

This type was added to address the "which values are bound" concern from the thread; shipped as a compile-time guarantee it does not provide, reviewers will trust the signature. Either drop `map` / `combine_with` (callers can `into_inner` explicitly, which at least makes the unbinding visible at the call site), or re-document the type as advisory.

challenger/src/fs/codecs/bytes_to_field.rs
```diff
@@ -0,0 +1,117 @@
...
+        let raw = big.to_bytes_be();
+        let mut buf: Vec<u8> = Vec::with_capacity(target);
+        if raw.len() < target {
+            buf.resize(target - raw.len(), 0);
+        }
+        buf.extend_from_slice(&raw);
+        challenger.observe_slice(&buf);
+    }
+
+    fn sample(challenger: &mut C) -> F {
```
Nashtare, line 42: **[HIGH] No spec or interop vectors remain.** The last revision removed `shake128.rs` together with the only KATs (IETF A.1 to A.9). Nothing now ties the implementation to `draft-irtf-cfrg-fiat-shamir` or to spongefish by test vector: the seeding format already diverges from IETF §5 by design, so this `Ns + 16` sampling is the only IETF-faithful piece left, and its proptest only checks the Horner fold against a BigUint reference (self-consistency, not conformance).

For something intended to become a wire-format commitment, please add at least one vector derived independently of this implementation (e.g. spongefish for the byte-to-field step) plus a pinned end-to-end NARG and challenge stream for a fixed pattern, so a refactor cannot silently change the format.

Also currently untested API surface: `pow` / `check_pow` (untestable, see the bounds comment), `add_extension` / `next_extension` / `challenge_extension`, and `add_public_scalar(s)` / `observe_public_scalar(s)` have zero call sites in tests.

challenger/src/fs/state/verifier.rs
```diff
@@ -0,0 +1,738 @@
...
+    /// Type-level marker for the sponge alphabet.
+    _u: core::marker::PhantomData<U>,
+}
+
+impl<C, U: Unit> Drop for VerifierState<'_, C, U> {
+    fn drop(&mut self) {
+        // Abort the player on drop when the user did not finalise
+        //
+        // This way, error paths do not turn into double-panics during cleanup.
+        if !self.player.is_finalized() {
```
Nashtare, line 59: **[MEDIUM] This silently defeats the strict-finalize guarantee for the drivers.** `PatternPlayer` panics on unfinalized drop by design (the description advertises this as the loud drop-time check), but both drivers `abort()` the player in `Drop`, so forgetting `finalize()` on a `ProverState` / `VerifierState` is silent.

For the verifier that skips both the pattern-fully-consumed check and the trailing-NARG-bytes check: a verifier that forgets `finalize()?` accepts proofs with arbitrary appended bytes and never notices unconsumed steps (e.g. a skipped `check_pow`).

`player.interact()` already marks `finalized` before its own panics, so the double-panic rationale only covers the `?` error paths. Suggestion: mark the player aborted on every `Err` return (a small wrapper around the fallible reads), and keep the panic for a clean drop with unconsumed steps. Same for the prover's `Drop`.

challenger/src/fs/pattern/step.rs
```diff
@@ -0,0 +1,328 @@
...
+impl Display for Interaction {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        if f.alternate() {
+            // Alternate mode: hash-stable form feeding the pattern fingerprint.
+            //
+            // `type_name` is omitted from the fingerprint.
+            //   - it is not a stable identifier; it can change across compiler versions.
+            //   - the fingerprint is a cross-run protocol commitment.
+            //   - a compiler-dependent commitment would break interop.
+            // Type identity is still enforced locally by `closes` and player equality.
```
Nashtare: **[MEDIUM] Excluding `type_name` leaves the seed blind to the field and to element width.** Agreed that `type_name` must not go into the hash (compiler-dependent), but nothing else binds the field modulus, the extension degree, or the codec: two protocols with identical version / name / labels / shape over two 31-bit primes (BabyBear vs KoalaBear), or a step that is an `F` scalar in one protocol and an `EF` scalar (4x the wire width) in another, produce byte-identical seeds and interchangeable absorbed prefixes. The fingerprint is meant to catch shape differences, and width is shape.

Suggestion: replace `type_name` with a stable, compiler-independent type identity in the hash (a small `TranscriptType` trait: prime fields render as their modulus, extensions as `modulus^degree`, bytes as `u8`). That also answers the earlier question on why `type_name` is ignored.

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+        self.player.interact(Interaction::new::<u64>(
+            Hierarchy::Atomic,
+            Kind::Pow,
+            "pow",
+            Length::Scalar,
+        ));
+        // Grind through the challenger's SIMD path.
+        let witness = self.challenger.grind(bits);
+        // Serialize as canonical big-endian, left-padded so width is constant.
+        let target = field_byte_size::<<C as GrindingChallenger>::Witness>();
```
Nashtare: **[MEDIUM] `bits` is bound nowhere.** The step is a fixed `("pow", Length::Scalar, u64)` regardless of difficulty, and `bits` is neither in the pattern hash nor absorbed. Prover and verifier must agree out-of-band, and a verifier configured with a smaller `bits` than the prover accepts silently instead of hitting a pattern mismatch, which is exactly the misconfiguration class this layer exists to catch. Encode it in the step (e.g. `Length::Fixed(bits)`).

Two adjacent nits: the type tag `u64` is wrong for a witness of type `C::Witness` (harmless but misleading), and the label is hardcoded unlike every other step.

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+    pub fn add_scalar<F, Cdc>(&mut self, label: Label, value: &F) -> TranscriptBound<F>
+    where
+        F: PrimeField,
+        Cdc: Codec<C, F>,
+    {
+        // Validate: the next pattern step is a scalar message of type `F`.
+        self.player.interact(Interaction::new::<F>(
+            Hierarchy::Atomic,
+            Kind::Message,
+            label,
```
Nashtare, line 192: **[MEDIUM] The message vocabulary cannot express a Plonky3 STARK transcript.** Messages can only be `PrimeField` scalars (`add_scalar(s)`, `_bounded`) or extension elements; there is no absorbed byte-string / digest message. The dominant prover message in this repo is a Merkle commitment (`Hash<F, u8, 32>`, `MerkleCap`), which today can only travel as a `Hint` (not absorbed, i.e. unsound) or a `Salt` (fixed label).

Together with the challenger-bounds issue this means no existing protocol can be driven end-to-end, and the divergence risk against the challengers actually in use grows with every release the module sits unused. An absorbed `add_bytes` / `next_bytes` `Message` step (fixed or bounded length) is the minimum needed.

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+    pub fn add_extension<F, EF, Cdc>(&mut self, label: Label, value: &EF) -> TranscriptBound<EF>
+    where
+        F: PrimeField,
+        EF: BasedVectorSpace<F> + Copy,
+        Cdc: Codec<C, F>,
+    {
+        // Validate: the next pattern step is a scalar message of extension type.
+        self.player.interact(Interaction::new::<EF>(
+            Hierarchy::Atomic,
+            Kind::Message,
```
Nashtare, line 307: `add_extension` / `challenge_extension` / `next_extension` hand-roll the coefficient loop that `ExtensionFieldCodec` already implements, and the drivers never use that codec. Either route these through `ExtensionFieldCodec<F, EF, Cdc>` or drop the codec; two implementations of the same wire layout will drift.

challenger/src/fs/pattern/step.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// This is distinct from a fixed length because the bound is part of the pattern hash.
+    ///
+    /// Two protocols differing only in their capacity get distinct seeds and cannot be confused.
+    Bounded(usize),
+    /// A dynamically-known number of values.
+    Dynamic,
+}
+
+/// Compile-time identifier of a step.
+///
```
Nashtare, line 117: **[MEDIUM] Recordable but unplayable vocabulary.** `Length::Dynamic` can be recorded via `PatternState`, but no driver method ever emits it, so any pattern containing it is guaranteed to panic at replay. Same for the non-`Protocol` `begin_*` / `end_*` helpers on the `Pattern` trait (`begin_message`, `begin_hint`, `begin_challenge`, `begin_public`): the drivers only expose `begin_protocol` / `end_protocol`. In a soundness-critical API these are latent footguns; please trim to what the drivers can play, or wire them up.

challenger/src/fs/error.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[error(transparent)]
+    MismatchedBeginEnd(Box<MismatchedBeginEndInfo>),
+
+    /// An opener was never closed before the sequence ended.
+    #[error(transparent)]
+    MissingEnd(Box<MissingEndInfo>),
+
+    /// Playback ran past the end of the recorded sequence.
+    #[error("transcript pattern exhausted")]
+    PatternExhausted,
```
Nashtare: `PatternExhausted`, `PatternMismatch`, `BadSaltLen` and `Other` are never constructed anywhere (pattern misuse panics instead, as settled in the thread). Dead variants in a public error enum invite callers to match on cases that cannot occur; suggest removing them.

challenger/src/fs/unit.rs
```diff
@@ -0,0 +1,68 @@
+//! Alphabet abstraction for the duplex-sponge transcript layer.
+
+use core::fmt::Debug;
+
+/// A single unit of a sponge's input/output stream.
+pub trait Unit: Copy + Default + Debug + Send + Sync + 'static {
```
Nashtare: The whole `Unit` machinery is inert: `BYTE_LEN` / `write_le` / `read_le` have no call sites outside this file's tests, only `u8` implements it, and the `U` parameter on `DomainSeparator` / `PatternState` / the drivers is a phantom with no behavior (`seed_bytes` requires `CanObserve<u8>` whatever `U` is). Its little-endian contract also contradicts the big-endian convention used everywhere else in the layer.

If a field-unit sponge path is coming (which would also address the challenger-bounds issue), this trait is where it should live; otherwise drop it.

challenger/src/fs/codecs/codec.rs
```diff
@@ -0,0 +1,17 @@
+//! Stateless adapter between values and a challenger.
+
+/// Stateless absorb/sample adapter for values of type `T` against challenger `C`.
+///
+/// Codecs are zero-sized types picked at the call site.
+///
+/// One transcript may invoke several codecs in different roles.
+pub trait Codec<C, T> {
+    /// Bits of statistical security: `-log2` distance from uniform on `T`.
+    const SECURITY_BITS: u32;
```
Nashtare, line 29: `SECURITY_BITS` is consumed by nothing (one test asserts it propagates through `ExtensionFieldCodec`). `FieldToFieldCodec` also reports 128 for a sample that is exactly uniform whenever the sponge is, so the value is arbitrary there. Either use it (e.g. a minimum asserted at driver level) or drop it.

challenger/src/fs/codecs/decode_field.rs
```diff
@@ -0,0 +1,211 @@
...
+///
+/// When the slice is too short or encodes an integer outside `[0, p)`.
+pub(crate) fn decode_field_be_canonical<F: PrimeField>(bytes: &[u8]) -> Result<F, TranscriptError> {
+    let need = field_byte_size::<F>();
+    if bytes.len() < need {
+        return Err(TranscriptError::BadProofShape {
+            reason: "not enough bytes for a canonical field encoding",
+        });
+    }
+    let big = BigUint::from_bytes_be(&bytes[..need]);
```
Nashtare: Following up on the earlier BigUint comment: the challenge-sampling path is now Horner in `F`, but the message path still allocates a `BigUint` per scalar on both sides (`from_bytes_be` plus the `order()` comparison here, `as_canonical_biguint().to_bytes_be()` in `encode_field_be`), i.e. on every `add_scalar` / `next_scalar`.

`PrimeField: QuotientMap<u64>`, so for `field_byte_size::<F>() <= 8` this can be `F::from_canonical_checked(u64::from_be_bytes(padded))` with no allocation (`None` is exactly the non-canonical case), keeping the BigUint route only for wider fields like Bn254. For the encode side, moving wire (de)serialization into `Codec` (as spongefish does) would let a `BytesToFieldCodec<F: PrimeField64>` use `as_canonical_u64()` directly.

challenger/src/fs/codecs/length_prefix.rs
```diff
@@ -0,0 +1,229 @@
...
+#[must_use]
+pub fn encode_len_be(len: usize, width: usize) -> [u8; 8] {
+    // Defensive upper bound: `usize` is at most 8 bytes on every supported platform.
+    assert!(width <= 8, "length-prefix width {width} exceeds 8 bytes");
+    // Range check the value against the declared width.
+    if width < 8 {
+        // Smallest value that does not fit: 1 << (width * 8).
+        //
+        // For width 0 this is 1, so only `len == 0` is legal.
+        let limit = 1usize << (width * 8);
```
Nashtare: **[MEDIUM] Overflows on 32-bit targets.** For `width == 4` (any `Bounded(max)` with `max >= 2^24`) this evaluates `1usize << 32`, which on wasm32 / 32-bit is a shift-overflow panic in debug and masks to `1 << 0` in release, so the assert below then rejects every non-zero `len`. CI only runs wasm32 tests for `p3-goldilocks`, so it would not catch this. Use `1u64 << (width * 8)` and compare `len as u64` (or `checked_shl`).

challenger/src/fs/pattern/sequence.rs
```diff
@@ -0,0 +1,411 @@
...
+        // Bounded by nesting depth, not by sequence length.
+        let mut stack: Vec<(usize, &Interaction)> = Vec::new();
+
+        for (position, interaction) in self.interactions.iter().enumerate() {
+            match interaction.hierarchy() {
+                // Phase: open a sub-protocol.
+                //
+                // Push regardless of kind.
+                // The kind compatibility check belongs to the next nested step, not the opener.
+                Hierarchy::Begin => stack.push((position, interaction)),
```
Nashtare: The non-`Protocol` kind rule is only checked for `Atomic` steps against the innermost opener; a `Begin` of a different kind is pushed unchecked. So `Begin Message { Begin Challenge { Atomic Challenge } }` validates (and `PatternState::interact` accepts it too), which bypasses the rule the module doc describes. The opener's kind should be checked against the surrounding container as well.

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+    /// Finalise the driver and return the serialised wire bytes.
+    ///
+    /// # Panics
+    ///
+    /// When the recorded pattern is not fully consumed.
+    pub fn finalize(self) -> Vec<u8> {
+        // The custom Drop impl prevents direct destructuring of `self`;
+        // So move each field out by hand under ManuallyDrop.
+        let this = core::mem::ManuallyDrop::new(self);
+        // SAFETY: each field is moved out exactly once, then Drop never runs on `this`.
```
Nashtare: Avoidable `unsafe` in a crypto crate: this is sound as written (each field read once, `player.finalize` marks before asserting), but the same is achievable safely by making `PatternPlayer::finalize` take `&mut self` and letting `Drop` run normally: `self.player.finalize(); core::mem::take(&mut self.narg)`. Same in `VerifierState::finalize`.

challenger/src/fs/state/verifier.rs
```diff
@@ -0,0 +1,738 @@
...
+            return Err(TranscriptError::BadProofShape {
+                reason: "trailing NARG bytes after final verifier step",
+            });
+        }
+        Ok(())
+    }
+
+    /// Take `n` raw bytes from the wire cursor, or fail if out of bounds.
+    fn take_bytes(&mut self, n: usize) -> Result<&'a [u8], TranscriptError> {
+        if self.cursor + n > self.narg.len() {
```
Nashtare: nit: `self.cursor + n` is unchecked. `n` is verifier-code controlled so this is not attacker-reachable, but `checked_add` keeps the arithmetic obviously total.

challenger/src/fs/state/verifier.rs
```diff
@@ -0,0 +1,738 @@
...
+            Kind::Pow,
+            "pow",
+            Length::Scalar,
+        ));
+        // Read the witness from the wire and decode it as a canonical field element.
+        let need = field_byte_size::<<C as GrindingChallenger>::Witness>();
+        let raw = self.take_bytes(need)?;
+        let witness = decode_field_be_canonical::<<C as GrindingChallenger>::Witness>(raw)?;
+        // Re-encode and compare so a malicious prover cannot smuggle a non-canonical encoding.
+        let canonical = encode_field_be::<<C as GrindingChallenger>::Witness>(&witness);
```
Nashtare: This re-encode and compare is dead code: `decode_field_be_canonical` already reads exactly `need` bytes and rejects values `>= p`, and the fixed-width big-endian encoding of an integer below `2^(8*need)` is unique, so `canonical == raw` always holds. `next_scalar` (correctly) does not do this.

challenger/src/fs/state/prover.rs
```diff
@@ -0,0 +1,854 @@
...
+    /// Absorb a salt step and record its bytes in the wire buffer.
+    pub fn add_salt(&mut self, salt_bytes: &[u8])
+    where
+        C: CanObserve<u8>,
+    {
+        // Validate: the next pattern step is a salt of the given length.
+        self.player.interact(Interaction::new::<u8>(
+            Hierarchy::Atomic,
+            Kind::Salt,
+            "salt",
```
Nashtare: nit: `add_salt` / `next_salt` and `pow` / `check_pow` hardcode their labels while every other step takes one; a protocol with two salts cannot tell them apart in a mismatch diff.

challenger/src/fs/domain_separator.rs
```diff
@@ -0,0 +1,270 @@
...
+/// - The version takes slot 0.
+/// - The name length takes the final slot.
+/// - The middle bytes hold the name itself.
+const NAME_LEN_INDEX: usize = PROTOCOL_ID_LEN - 1;
+
+/// Largest protocol name that fits: `[version | name | name_len]`.
+const MAX_NAME_LEN: usize = PROTOCOL_ID_LEN - 2;
+
+/// Terminator that closes the seeding stream so no seed is a prefix of another.
+const DOMAIN_TAG: u8 = 0x80;
```
Nashtare, line 30: nit: with the 4-byte length prefix on the label, the terminator is redundant for prefix-freeness. Fine to keep if it is meant to mirror a future IETF alignment, but then say so; otherwise it is one more byte diverging from the draft.

challenger/src/fs/domain_separator.rs
```diff
@@ -0,0 +1,270 @@
...
+        Self {
+            protocol_id,
+            instance_label: Vec::new(),
+            pattern: Arc::new(pattern),
+            _u: PhantomData,
+        }
+    }
+
+    /// Append bytes to the instance label. Chainable.
+    pub fn instance(&mut self, bytes: &[u8]) -> &mut Self {
```
Nashtare, line 105: `instance` concatenates without delimiting, so `.instance(a).instance(b)` and `.instance(a ++ b)` yield the same seed; a caller passing e.g. a public key then a message gets no separation between them. Either length-prefix each call or document that the label is a single opaque blob.

challenger/Cargo.toml
```diff
@@ -11,20 +11,22 @@ categories.workspace = true
...
+thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
 p3-baby-bear = { path = "../baby-bear" }
 p3-bn254 = { path = "../bn254" }
 p3-goldilocks = { path = "../goldilocks" }
-
-num-bigint.workspace = true
 proptest.workspace = true
```
Nashtare, line 29: nit: `rand` is added as a dev-dependency but has no use site in the diff.

Nashtare (changes requested):
There are still a couple concerns that should be addressed (agent review below that I skimmed through), although I'd happily get this merged in once clean!

challenger/src/fs/state/prover.rs
tcoratger: Fixed. `Unit` now names the sponge alphabet and owns the byte-absorb rule, which is the only place a byte sponge and a field sponge differ.

`FieldUnit<F>` packs the seed (and every raw byte string the layer absorbs: salts, length prefixes, byte messages) into field elements, 3 bytes per element for a 31-bit prime, behind a leading length element so padding cannot collapse two distinct strings. `u8` keeps the verbatim path.

`ProverState::new` now asks for `CanObserve<U::Item>`, so `DuplexChallenger` and `SerializingChallenger32/64` drive the layer end to end with `FieldToFieldCodec`. Since those are also `GrindingChallenger`, the two bounds are jointly satisfiable and `pow` / `check_pow` are tested for real: round trip, tampered witness rejected, and a difficulty mismatch caught by the pattern.

New tests: `round_trip_over_a_native_field_sponge`, `round_trip_over_a_duplex_sponge`, `proof_of_work_round_trips_and_binds_its_difficulty`, `proof_of_work_rejects_a_tampered_witness`, `proof_of_work_difficulty_mismatch_is_caught_by_the_pattern`.

challenger/src/fs/bound.rs
tcoratger: Fixed by dropping `map` and `combine_with`. You are right that the purity caveat missed the real hole: a pure closure that ignores its argument launders the witness just as silently, and a refactor can introduce one by accident.

The only way out of the wrapper is now `into_inner`, which is visible at the call site. The doc says plainly what the guarantee is and is not: it says a `TranscriptBound<T>` was minted by a transcript method, nothing more.

challenger/src/fs/codecs/bytes_to_field.rs
tcoratger: Fixed on both counts.

Interop vectors: `ietf_extra_byte_decoding_matches_independent_vectors` pins the `Ns + 16` step over BabyBear and Goldilocks against expected values computed outside this crate as `int.from_bytes(input, "big") % p`. They are literals in the test, so they check the in-field Horner fold against arithmetic it did not produce.

Format pin: `end_to_end_wire_format_is_pinned` fixes a pattern, a protocol id and an instance label, then asserts the literal NARG bytes and the literal challenge values, and replays them through the verifier. A refactor that changes the seeding layout, the absorb order or the decoding breaks this test.

Untested surface, now covered: `pow` / `check_pow` (reachable since the challenger-bounds fix), `add_extension` / `next_extension` / `challenge_extension`, and `add_public_scalar(s)` / `observe_public_scalar(s)`.

challenger/src/fs/state/verifier.rs
tcoratger: Fixed. Both drivers now panic on drop when the player is unfinalized, and the message names how many steps and bytes went unchecked.

The `?` paths are handled the way you suggested: every fallible read passes its result through a small `poison` wrapper that marks the player aborted on `Err`, so the structured error reaches the caller instead of being buried under a drop-time panic. The driver's own asserts (over-cap hint or message) abort first too, so those panics stay single.

Tests: `prover_dropped_without_finalize_panics`, `verifier_dropped_without_finalize_panics` (which appends a trailing byte, exactly the proof a forgetful verifier would have accepted), and `a_failed_read_releases_the_drop_check`.

challenger/src/fs/pattern/step.rs
tcoratger: Fixed along the lines you suggested. A new `TypeTag` carries a compiler-independent identity and it is what the alternate-mode rendering feeds into the fingerprint:

- `Algebra { modulus, degree }` — the prime field's `ORDER_U64` plus the extension degree, both associated constants, so no allocation and no compiler dependence.
- `Bytes` for hints, salts and digest-shaped messages.
- `Marker` for structural openers and closers.

`Interaction::algebra::<F, A>` / `::bytes` / `::marker` replace the old `new::<T>`. BabyBear vs KoalaBear, and `F` vs a degree-4 element, now produce different seeds — `tag_separates_two_primes_of_the_same_width`, `tag_separates_a_base_element_from_an_extension_element`, `pattern_hash_distinguishes_two_fields_of_the_same_width`.

`type_name` stays on the record as a local diagnostic in panic diffs, still excluded from the hash.

challenger/src/fs/state/prover.rs
tcoratger: Fixed. The step is now `Interaction::algebra::<C::Witness, C::Witness>(Atomic, Pow, label, Length::Fixed(bits))`:

- `bits` is in the pattern, so it is in the fingerprint and in the seed;
- the type tag is the witness field, not `u64`;
- the label is a parameter like every other step.

A verifier configured with a smaller `bits` than the prover now hits a pattern mismatch — `proof_of_work_difficulty_mismatch_is_caught_by_the_pattern`.

challenger/src/fs/state/prover.rs
tcoratger: Fixed. Added `add_bytes` / `add_bytes_bounded` on the prover and `next_bytes` / `next_bytes_bounded` on the verifier: `Kind::Message` steps carrying raw bytes that *are* absorbed, so a Merkle root or cap has a sound home instead of travelling as an unabsorbed `Hint`.

The bounded variant absorbs the length prefix before the payload, keeping the prefix-freeness condition of CO25 §6.2, and mirrors both onto the wire.

`byte_messages_are_absorbed_unlike_hints` checks that changing the commitment moves the challenge, that a longer opening is a different transcript, and that the verifier lands on the same challenge.

Combined with the challenger-bounds fix, a Plonky3-shaped transcript is now expressible end to end.

challenger/src/fs/state/prover.rs
tcoratger: Fixed by routing them through the codec. `Codec` gained the wire half — `wire_len`, `encode`, `decode` — so a codec owns both the sponge format and the byte format and the two cannot drift.

`add_extension`, `next_extension` and `challenge_extension` now delegate to `ExtensionFieldCodec<F, EF, Cdc>` for both halves; the coefficient loop exists once, in the codec. `extension_messages_and_challenges_round_trip` covers it, and `extension_wire_layout_is_the_base_encoding_repeated` pins the layout.

challenger/src/fs/pattern/step.rs
tcoratger: Fixed by trimming *and* wiring up, so the two sides match exactly.

- `Length::Dynamic` is removed. Nothing can record it any more.
- The `begin_message` / `begin_hint` / `begin_challenge` / `begin_public` helpers are gone. The `Pattern` trait is now `interact`, `abort`, `begin<T>(label, kind)`, `end<T>(label, kind)` and the `begin_protocol` / `end_protocol` sugar.
- Both drivers expose `begin` / `end` for every kind, so a non-`Protocol` container is playable rather than a guaranteed panic.

`nested_sub_protocols_replay_in_lockstep` records a `Kind::Message` container and replays it through both drivers.

challenger/src/fs/error.rs
tcoratger: Fixed — `PatternExhausted`, `PatternMismatch` (with its `PatternMismatchInfo` payload), `BadSaltLen` and `Other` are removed, along with the now-unused `alloc::string::String` import.

The enum doc now states the rule explicitly: every variant is reachable, and divergence between the recorded pattern and the code replaying it is a programming bug that panics with a diff rather than landing in this type.

challenger/src/fs/unit.rs
tcoratger: Fixed by giving it the job you describe. `Unit` is no longer a phantom: it names the sponge alphabet via an associated `Item` and owns one rule, `observe_bytes`, which is how a raw byte string enters a sponge over that alphabet.

- `impl Unit for u8` takes the bytes verbatim.
- `FieldUnit<F>` packs them into field elements, `(bits - 1) / 8` bytes each, behind a leading length element so padding cannot collapse two distinct strings.

`BYTE_LEN` / `write_le` / `read_le` and the little-endian contract that contradicted the rest of the layer are gone. Seeding, salts, bounded length prefixes and byte messages all route through the alphabet, which is what unblocks `DuplexChallenger`.

challenger/src/fs/codecs/codec.rs
tcoratger: Fixed by using it. `MIN_CHALLENGE_SECURITY_BITS = 128` is the DSFS / IETF §6 budget, and every sampling method on both drivers const-asserts `Cdc::SECURITY_BITS >= MIN_CHALLENGE_SECURITY_BITS`.

It is a compile-time constant evaluation, so wiring a biased codec into a challenge step fails to build. Verified by temporarily setting `BytesToFieldCodec::SECURITY_BITS` to 23: `error[E0080]: evaluation panicked: codec is too biased to sample challenge material`. That is exactly the guard that would have caught the truncation-biased codec removed in the previous round.

`FieldToFieldCodec`'s value is now documented rather than arbitrary: the sample is the sponge's own output, no decoding step adds bias, so the budget is met by construction.

challenger/src/fs/codecs/decode_field.rs
tcoratger: Fixed by moving wire (de)serialization into `Codec`, as you suggested.

`BytesToFieldCodec<F: PrimeField64>` and `FieldToFieldCodec<F: PrimeField64>` now encode with `as_canonical_u64().to_be_bytes()` and decode with `from_canonical_checked(u64::from_be_bytes(padded))`, where `None` is precisely the non-canonical case, so the range check and the lift are one operation. No allocation on either side, on any per-value path.

`BigUint` is gone from the module entirely — no `as_canonical_biguint`, no `order()` comparison, no limb folding. `num-bigint` drops from `[dependencies]` to `[dev-dependencies]`, where it stays as the reference in `horner_decode_matches_biguint_reference`.

challenger/src/fs/codecs/length_prefix.rs
tcoratger: Fixed. The bound is evaluated in `u64` now:

```rust
let limit = 1u64 << (width * 8);
assert!((len as u64) < limit, ...);
```

`width_four_accepts_the_whole_range_on_every_target` round-trips `u32::MAX` at width 4, which is the case that would have panicked in debug and silently rejected everything in release on wasm32.

challenger/src/fs/pattern/sequence.rs
tcoratger: Fixed. The kind check is now a shared helper applied to `Begin` as well as `Atomic`, in both the sequence validator and `PatternState`, so a nested opener is subject to the enclosing container's rule.

`Begin Message { Begin Challenge { Atomic Challenge } }` is rejected with the same `InvalidKind` payload that a bare nested challenge would produce — `nested_container_of_a_foreign_kind_is_rejected` and `nested_container_of_a_foreign_kind_panics_on_record`. `nested_container_of_the_same_kind_is_accepted` pins the boundary the other way.

challenger/src/fs/state/prover.rs
tcoratger: Fixed exactly as suggested. `PatternPlayer::finalize` takes `&mut self` and marks before asserting, so both drivers are now:

```rust
self.player.finalize();
core::mem::take(&mut self.narg)
```

with `Drop` running normally afterwards. There is no `unsafe` left anywhere under `challenger/src/fs/`.

challenger/src/fs/state/verifier.rs
tcoratger: Fixed — `take_bytes` computes the end offset with `checked_add` and a `filter` on the length bound, so the arithmetic is total whatever the caller passes.

challenger/src/fs/state/verifier.rs
tcoratger: Fixed — removed. You are right: `decode_field_be_canonical` reads exactly `field_byte_size` bytes and rejects anything at or above `p`, and the fixed-width big-endian encoding of an integer below `2^(8*need)` is unique, so a canonical decode already implies a canonical encoding.

The reasoning is now a comment at the call site so the check does not get re-added.

challenger/src/fs/state/prover.rs
tcoratger: Fixed — `add_salt`, `next_salt`, `pow` and `check_pow` all take a `Label` like every other step, so two salts in one protocol are distinguishable in a mismatch diff.

challenger/src/fs/domain_separator.rs
tcoratger: Documented rather than removed, with the reasoning you asked for:

```
/// The length prefix on the instance label already makes the seed prefix-free.
/// This byte is therefore redundant for that purpose.
///
/// It is kept because the IETF draft ends its seeding stream with a domain tag.
/// Aligning the two later should not need a format change.
```

challenger/src/fs/domain_separator.rs
tcoratger: Fixed by length-prefixing each call. `instance()` writes `[len_be_4 | bytes]` per chunk, so `.instance(key).instance(message)` and `.instance(key ++ message)` produce different labels and different seeds — `instance_chunks_are_delimited`.

The doc spells the layout out, and the whole label region keeps its own outer length prefix in the seed stream.

challenger/Cargo.toml
tcoratger: Fixed — removed. Confirmed no use site anywhere in `challenger/`.

`num-bigint` moved the other way in the same commit: out of `[dependencies]`, into `[dev-dependencies]`, where it is now only the reference implementation for the decode proptest.

Nashtare (approved):
Thanks a bunch! This is looking really great!

PR #1604, opened 2026-05-03
chore(batch-stark): deterministic global lookup error order via BTreeMap

shashiyy:
Swapped the `HashMap<&String, Vec<_>>` aggregator in `verifier::verify` for a `BTreeMap` so the first reported `LookupError` is stable across runs. Also dropped the now-unused `hashbrown::HashMap` import while in there.                                                                                                      
                                                                                                                                                                                                                                                                                                                                   
Before: with multiple failing global lookups, the `?` in `for (name, all_expected_cumulative) in global_cumulative` returned whichever name `hashbrown` happened to yield first — re-running the same failing proof could surface a different `name` in the error string.                                                        
                                                                                                                                                                                                                                                                                                                                   
After: iteration follows lexicographic order of the lookup name, so error messages like `LookupError(...): foo` are reproducible. The map is small (one entry per distinct global lookup name) and only walked once on the failure path, so there is no perf concern.

### DIFF
--- a/batch-stark/src/verifier/mod.rs
+++ b/batch-stark/src/verifier/mod.rs
@@ -1,11 +1,11 @@
 mod data;
 
+use alloc::collections::BTreeMap;
 use alloc::string::String;
 use alloc::vec::Vec;
 use alloc::{format, vec};
 
 pub use data::VerifierData;
-use hashbrown::HashMap;
 use p3_air::Air;
 use p3_air::symbolic::{AirLayout, SymbolicAirBuilder, SymbolicExpressionExt};
 use p3_commit::{Pcs, PolynomialSpace};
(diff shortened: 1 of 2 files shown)


tcoratger:
Can we have a small test showing something that was failing before and that passes now with the modification?

Nashtare:
I wouldn't really consider this as an issue. If 2 lookups are failing, having a non-deterministic ordering isn't so much of an issue. What would matter is having an error not being triggered upon re-run for instance. What matters most is that any lookup inconsistency is being properly flagged.

shashiyy:
> Can we have a small test showing something that was failing before and that passes now with the modification?

Sure added. @tcoratger

Issue #1590, opened 2026-04-28
Tracking: HVZK-WHIR — zero-knowledge variant of WHIR (eprint 2026/391)


tcoratger:
## Background

WHIR (`whir/`) currently has no zero-knowledge property. The recent paper

> Alessandro Chiesa, Giacomo Fenzi, Guy Weissenberg.
> *Zero-Knowledge IOPPs for Constrained Interleaved Codes.*
> [eprint 2026/391](https://eprint.iacr.org/2026/391.pdf), Feb 2026.

gives an HVZK variant of WHIR with `1 + o(1)` overhead — meaning enabling ZK is essentially free, and the non-ZK path can stay byte-for-byte unchanged. The construction is modular: it adds small "mask" oracles to the existing sumcheck and code-switching rounds rather than redesigning WHIR.

This is a tracking issue for the implementation, decomposed so each sub-issue is one PR reviewable on its own.

## Why now

- ZK is increasingly required for privacy-preserving applications using hash-based SNARKs (anonymous credentials, identity, age assurance — see paper §1).
- The paper's overhead is small enough that ZK can be opt-in at no cost when off.
- The FRI stack already has `HidingFriPcs` as precedent — WHIR is the missing piece.

## Plan

- [x] #1584 — ZK encoding trait + Reed–Solomon (Definition 3.16, Proposition 3.19)
- [x] #1585 — Private zero-evader (Definition 9.2, Lemma 9.3)
- [x] #1586 — HVZK sumcheck with sublinear masks (Section 6, Construction 6.3)
- [ ] #1587 — HVZK code-switching round (Section 9, Construction 9.7)
- [ ] #1588 — HVZK base-case IOPP (Section 7, Construction 7.2)
- [ ] #1589 — Compose HVZK-WHIR PCS + end-to-end tests (Section 10, Theorem 4)

#1584 and #1585 are independent leaves and can be picked up in parallel. #1586 is the meatiest and depends on #1584. #1587 and #1588 each depend on earlier blocks. #1589 is integration plumbing.

## Out of scope (separate issues if/when needed)

- R1CS reduction (paper §11) — Plonky3 is AIR-only today.
- Distance-amplified codes (paper §12) — orthogonal optimization.
- Holographic / preprocessed verifier — orthogonal.
- Malicious-verifier ZK is implied at the SNARG level via Fiat–Shamir; no extra work needed here.

## Notes for contributors

[...]

Issue #1584, opened 2026-04-28
[HVZK-WHIR 1/6] ZK encoding trait + Reed–Solomon instantiation


tcoratger:
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 3.2.1](https://eprint.iacr.org/2026/391.pdf).

## What

A randomized encoding `Enc: F^ℓ × F^r → Σ^m` such that any `t` codeword positions reveal nothing about the message (Definition 3.16). For Reed–Solomon: append `t` random field elements to the message before DFT evaluation (Proposition 3.19, error 0).

The ad-hoc version of this already lives inside `fri/src/hiding_pcs.rs` (`random_coeffs`); this issue extracts it as a reusable trait used by later issues.

## API sketch

Traits in `commit/`, or a new `p3-zk-codes/` crate — pick what fits, motivate in the PR.

```rust
pub trait ZkEncoding<F: Field> {
    type Codeword;

    /// Number of codeword positions that reveal nothing about the message.
    fn t(&self) -> usize;
    /// Statistical simulation error (0 for Reed–Solomon).
    fn error(&self) -> f64;

    fn encode<R: Rng>(&self, msg: &[F], rng: &mut R) -> Self::Codeword;

    /// For `|S| ≤ t()`, identically (or `error()`-close) distributed
    /// to `encode(msg, _)[S]`, without access to `msg`.
    fn simulate<R: Rng>(&self, query_set: &[usize], rng: &mut R) -> Vec<F>;
}
```

Issue #1587 will also need accessors to the encoding's generator-matrix rows (`G^#`, `G^$` of Definition 3.17). Either expose them on the trait or via a sub-trait — contributor's call.

## Acceptance criteria

- `ReedSolomonZkEncoding<F, Dft>` implementing the trait, parameterized by `t`.
- proptest: for any `(msg, rng_state, S)` with `|S| ≤ t`, `encode(msg, rng)[S]` and `simulate(S, rng)` are identically distributed.
- proptest: with all-zero randomness, output equals plain (non-ZK) RS encoding bit-for-bit.
- criterion bench: encoding overhead vs plain RS is `O(t)`, not `O(t · ℓ)`.

## Out of scope

Linear-time encodable ZK codes (BCL22 Theorem 3.20). Useful for a linear-prover variant later, not needed here.

## Notes

[...]

Issue #1587, opened 2026-04-28
[HVZK-WHIR 4/6] HVZK code-switching round


tcoratger:
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 9.4 / Construction 9.7](https://eprint.iacr.org/2026/391.pdf). Depends on #1584, #1585, #1586.

## What

A WHIR round reduces a claim about a large interleaved code to one about a smaller code. The HVZK version (Construction 9.7) makes two local changes to a round:

1. The prover sends a fresh ZK mask `s = Enc(s̃, r'')` for the smaller code (#1584).
2. The OOD sample is replaced by a `PrivateZeroEvader` (#1585): instead of `y = ze(ρ) · f`, the prover sends `y = ze(ρ) · (f, r, s)` for fresh randomness `r`.

The joint constraint is then carried through the existing batching machinery — no change to that part. Lemmas 9.8 (HVZK simulator) and 9.9 (RBR soundness) cover the analysis.

## Where

`whir/src/pcs/prover.rs` and `verifier.rs`. Add a `zk: bool` flag on the round; the `false` branch keeps today's code unchanged.

## Acceptance criteria

- Per-round prover communication grows by exactly `m_zk · ι_zk + t_ood` field elements over the non-ZK round (matches paper's bound).
- proptest end-to-end on a small instance (e.g., RS over `BinomialExtensionField<KoalaBear, 4>`, `n = 4`, `λ = 20`): honest prover always accepts.
- proptest: simulator composes with #1586's via Theorem 4.5 — the composed simulator's distribution matches the real view.

## Notes

Read **Section 4 (Composition for honest-verifier zero-knowledge IORs)** before starting. The composition theorem (Theorem 4.5) is the structural reason this paper can keep simulators modular; #1586 onward only works if each round's simulator is composable in the paper's sense.

Design choices above are guidelines. If a cleaner decomposition emerges during implementation, raise it in the PR.

WizardOfMenlo:
One note for an optimization here: currently, we send an additional mask that sends the folded encoding randomness (+ additional randomness). This is somewhat annoying because you can't commit it in together with the other mask (all of which should be sent in a single Merkle tree). However, you can change things slighly and send (at the beginning) an interleaved encoding of the folding randomness + additional zero-evader masking randomness) and fold it in the sumcheck instead. This should reduce the number of Merkle roots (and consequently copaths).

latifkasuli:
[...]

PR #1605, opened 2026-05-04
[HVZK-WHIR 3/6] HVZK sumcheck with sublinear masks

### LINKED ISSUE #1586: [HVZK-WHIR 3/6] HVZK sumcheck with sublinear masks
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 6 / Construction 6.3](https://eprint.iacr.org/2026/391.pdf). Depends on #1584 (ZK encoding).

## What

This is the core technical contribution of the paper. WHIR's sumcheck reduces a claim about an interleaved code to a claim about its base code; the HVZK variant additionally sends `k` small mask polynomials encoded with a ZK code (issue 1), then folds them into the sumcheck claim with one combination challenge `ε`. Per-round message size grows by `Õ(λ)` field elements, independent of the witness size — this is what gives the paper its `1 + o(1)` overhead.

Construction 6.3 is fully spelled out. Lemma 6.4 gives the simulator and HVZK error; Lemma 6.5 gives RBR knowledge soundness.

## Where

`whir/src/sumcheck/zk.rs` (new), mirroring `single.rs`. Gate behind a const generic on the existing sumcheck types so the non-ZK path is byte-for-byte unchanged.

## API sketch

```rust
pub struct ZkSumcheckProver<F, EF, Enc>
where
    F: Field,
    EF: ExtensionField<F>,
    Enc: ZkEncoding<F>,
{
    base: WhirSumcheckProver<F, EF>,
    encoding: Enc,
    masks: Vec<UnivariatePolynomial<F>>,
    mask_oracles: Vec<Enc::Codeword>,
}

impl<F, EF, Enc: ZkEncoding<F>> ZkSumcheckProver<F, EF, Enc> {
    pub fn commit_masks<R: Rng>(&mut self, rng: &mut R);     // step 1 of Construction 6.3
    pub fn send_target_mu_tilde(&self) -> EF;                // step 2
    pub fn round(&mut self, gamma: EF, eps: EF) -> UnivariatePolynomial<EF>; // step 4
}
```

## Acceptance criteria

- proptest: completeness across many `(k, ℓ_zk)` configurations.
- proptest: HVZK simulator (Lemma 6.4) matches the real view exactly for RS-based encoding (error 0). Use the `simulate` from #1584 for queries to mask oracles.
- proptest: an empirical RBR soundness bound matching Lemma 6.5 over a small field.
- regression test: with `ZK = false`, the transcript is byte-identical to today's sumcheck.
- criterion bench: ZK overhead `≤ 1.05×` for `(2^20 rows, k = 4, λ = 100)`.

## Notes

- Construction 6.3 requires `char(F) ≠ 2` and `ℓ_zk ≥ 2` (Lemma 6.4 proof sketch). KoalaBear and BabyBear fit; add a `static_assert`.
- The simulator argument in Lemma 6.4 counts degrees of freedom in the masked sumcheck space; understanding it is the way to build confidence in the implementation, but the implementation itself does not need that machinery.
- API and file layout above are guidelines. If the right factoring turns out different (e.g., merging mask commit and target into one step), raise it in the PR.

camofu:
Closes #1586. Part of the HVZK-WHIR effort tracked in #1590. Depends on #1584 (ZK encoding traits, now in `p3-zk-codes` since #1601).

## Background

WHIR's sumcheck currently reduces a claim about a `2^k`-interleaved Reed–Solomon codeword to a claim about the base codeword by sending `k` univariate round-polynomials `ĥ_1, …, ĥ_k`. Each `ĥ_j` is a linear function of the secret witness `f`, so it leaks information — fine for non-private WHIR, but incompatible with the HVZK-WHIR effort tracked in #1590.

[Construction 6.3 of eprint 2026/391](https://eprint.iacr.org/2026/391.pdf) ("Zero-Knowledge IOPPs for Constrained Interleaved Codes", Chiesa–Fenzi–Weissenberg, Feb 2026) shows how to make this sumcheck honest-verifier zero-knowledge with only `Õ(k·λ)` extra field elements — i.e. the zero-knowledge overhead is independent of the witness size, which is what gives the paper its headline `1 + o(1)` overhead property.

## Goal

Add an HVZK variant of the WHIR sumcheck in `whir/src/sumcheck/zk.rs` implementing Construction 6.3:

1. Before the `k` sumcheck rounds, the prover commits to `k` small random univariate masks `s_1, …, s_k` under the ZK encoding from #1584.
2. The prover sends `μ̃ := Σ_{b ∈ {0,1}^k} (ŝ_1(b_1) + … + ŝ_k(b_k))`.
3. The verifier sends one combination challenge `ε`.
4. The `k` sumcheck rounds run on the *masked* claim `μ̃ + ε·μ`; each `ĥ_j` is (mask contribution from `s_j`) + ε·(original sumcheck contribution).

The verifier's affine consistency checks (`ĥ_1(0) + ĥ_1(1) = ε·μ + μ̃` and `ĥ_j(0) + ĥ_j(1) = ĥ_{j-1}(γ_{j-1})` for j ≥ 2) are the *only* constraints on the transcript — uniform masks therefore induce a uniform transcript inside that affine subspace, which the HVZK simulator (Lemma 6.4) can sample without ever touching the witness. Masks are kept as oracles and carried into downstream reductions via the committed sumcheck relation.

This is brick 3/6 of HVZK-WHIR. Unblocks the HVZK code-switching and end-to-end HVZK IOPP that follow in the series.

Requires `char(F) ≠ 2` and `ℓ_zk ≥ 2` (Lemma 6.4); BabyBear and KoalaBear satisfy both. Non-ZK path is gated by a const generic and remains byte-identical.

## Acceptance criteria (lifted from #1586)

- [x] proptest: completeness across many `(k, ℓ_zk)` configurations.
- [x] proptest: HVZK simulator (Lemma 6.4) matches the real view exactly for RS-based encoding (error 0). Use `Enc::simulate` from #1584 for queries to mask oracles.
- [x] proptest: empirical RBR soundness (Lemma 6.5) over a small field.
- [x] regression test: with ZK gated off, the transcript is byte-identical to today's sumcheck. (Moot after https://github.com/Plonky3/Plonky3/pull/1612)
- [x] criterion bench: ZK overhead `≤ 1.05×` for `(2^20 rows, k = 4, λ = 100)`.


### Note on acceptance criterion 3 (empirical RBR soundness)                                                                
                                                                                                                              
The implementation-conformance surface that an empirical Lemma 6.5 test would cover is already covered jointly by the existing tests in `whir/src/sumcheck/zk.rs`:                                                                                
                                                                                                                              
- `prop_completeness_classic_unpacked` (criterion 1) — honest case accepts across the `(n_vars, k, ℓ_zk, num_eqs)` cube.    
- `prop_simulator_view_matches_real_rs` (criterion 2) — Lemma 6.4 simulator-real bit-equivalence under matched RNG, plus F-subspace stratification and mask oracle simulation via `ZkEncoding::simulate`. The HVZK structure Lemma 6.5 builds on.    
- `prop_rbr_tampering_changes_verifier_output` — any single-byte cheat in any wire coordinate propagates through Fiat-Shamir to the verifier's final target. Asserts Lemma 6.5's precondition: an undetected cheat cannot survive Fiat-Shamir for free. 
                             
Commit `6be15237` tightens the tampering test's docstring to spell out this coverage role explicitly.                       
                             
Not covered: the quantitative `empirical_rate ≤ ε_j` match. `γ_j` is sampled from `EF ≈ 2^124` (frozen design decision: `ε ∈ EF` for soundness margin), so for any feasible params the bound is ~2^-110 — unobservable at any CI-feasible trial count.
Making it observable requires a sub-2^31 test-only field type, which `p3-field` doesn't ship; adding one is sub-PR-sized infra disproportionate to the marginal coverage. Lemma 6.5 itself is a theorem about the abstract protocol — what this PR can test is conformance to that protocol, which the three tests above pin down.

Happy to revisit if a small test field lands in `p3-field` separately.


latifkasuli:
Nice work getting this up — the `ZkSumcheckData` / `verify_classic_unpacked` shape looks clean.

One small robustness note: in `verify_classic_unpacked`, the loop indexes `zk_data.pow_witnesses[j_idx]` when `pow_bits > 0`, but there's no upfront check that `pow_witnesses.len() == k` (or `== round_coefficients.len()`). A malformed proof with fewer witnesses than rounds would panic on out-of-bounds instead of returning `SumcheckError`. Might be worth a shape validation at the top of the function:

```rust
if pow_bits > 0 && zk_data.pow_witnesses.len() != zk_data.round_coefficients.len() {
    return Err(SumcheckError::InvalidProofShape);
}
```

(or similar — just so verifier-facing code stays panic-free on adversarial input.)

camofu:
Thanks @latifkasuli ! fixed in 58abec5a. Added: `SumcheckError::PowWitnessCountMismatch { expected, actual }`. lmk if it is what you had in mind

camofu:
@tcoratger the PR is ready for review :)

latifkasuli:
Yes, that's exactly what I had in mind. `PowWitnessCountMismatch { expected, actual }` is cleaner than overloading a generic proof-shape error, and it keeps the verifier path panic-free on malformed input. Thanks for the quick fix.

latifkasuli:
One small follow-up on the PoW shape check: the new `PowWitnessCountMismatch` handles the `pow_bits > 0` case cleanly. Should the verifier also require `pow_witnesses.is_empty()` when `pow_bits == 0`?

It is not a soundness issue since those witnesses are ignored, but rejecting unused proof data would keep the serialized proof shape canonical and make proof-size accounting less surprising.

latifkasuli:
Separate simulator-distribution question: `simulate_classic_unpacked` samples each round wire coefficient uniformly from `EF`, while the honest prover's masks are sampled over `F` and then lifted into `EF`.

For `ell_zk >= 4`, this seems potentially distinguishable. The sent wire contains coefficients above degree 2, for example `c_3`, and the plain sumcheck piece has no contribution there. In an honest transcript, that coordinate appears to be base-field embedded; in the simulator, it is uniform in `EF`. A distinguisher could check whether the non-base extension coordinates are zero.

The current simulator proptest checks that simulated transcripts are verifier-accepted, but not that the simulated view is distributed like the honest view.

Is the intended theorem here over `F` only, or should the simulator sample the mask-only coordinates from base-field embeddings while only the plain/mixed coefficients live in `EF`?

Asking because this matters downstream for #1587: the simulator composition via Theorem 4.5 needs real-view vs simulated-view indistinguishability, not only verifier acceptance.

camofu:
Thanks @latifkasuli ! 
Especially for the simulator-distribution notice. You are correct, it slipped out of my attention because I changed the protocol to sample from EF and not from F (the paper is full F) for soundness reasons, like it is custom in the Plonky3 repo.                                                                                                                           
                                                                   
The theorem is now hybrid, not pure-F: masks s_j stay in F (this is what gives the sublinear proof-size benefit, right?), while ε, γ_j, and the plain WHIR sumcheck piece live in EF.               

So each round polynomial ĥ_j of size max(ℓ_zk, 3) splits into two tiers:                                                                                                                    
1. c_0, c_1, c_2: receive an ε · plain_c_* contribution. Honest values are uniform over EF.                                                                                                  
2. c_i for i ≥ 3: receive only the mask term lifted from F into EF. Honest values sit in the F-subspace of EF.

The simulator should therefore sample wire[0..2] from EF and wire[2..] from F (lifted into EF). The previous proptest only checked verifier acceptance, which doesn't catch this gap. I've added prop_simulator_f_subspace_invariant (sweeping ell_zk ∈ 2..=6) that checks the F-subspace stratification on wire[2..] for both honest and simulated transcripts via the EF→F^{ext_deg}  basis decomposition.                                                                                                                                                                        
                             
This should be fixed in the latest commit, let me know what you think :)

tcoratger:
Just waiting for this full integration https://github.com/Plonky3/Plonky3/pull/1612 to be merged before going towards this one just after :)

latifkasuli:
This looks correct to me. The two-tier stratification matches the hybrid F/EF setting: the plain sumcheck piece has degree ≤ 2, so `c_0` and `c_2` can receive the `ε · plain` contribution and live in full `EF`, while `c_i` for `i ≥ 3` are mask-only and should stay in the embedded `F` subspace.

The `BasedVectorSpace` check in `prop_simulator_f_subspace_invariant` is a clean way to pin down exactly the distinguisher we were worried about. The PoW canonicality fix also looks good.

Minor non-blocking readability note: since the wire skips `c_1`, the mapping `wire[1] = c_2`, `wire[2] = c_3` is easy to miss. A one-liner near the simulator sampling comment may help future readers, but this is not blocking.

Thanks for the quick and thorough fix, this is the simulator shape we’ll need downstream for [#1587’s](https://github.com/Plonky3/Plonky3/issues/1587) composition argument.

tcoratger:
@CarloModicaPortfolio I think this needs to be changed a bit now that #1612 has been merged (which offers the best unified solution for leanVM and stacked polynomials).

camofu:
Will work on it tomorrow :)

whir/benches/sumcheck.rs
```diff
@@ -19,6 +24,25 @@ type Challenger = DuplexChallenger<F, Perm, 16, 8>;
...
+    TruncatedPermutation<Perm, 2, 8, 16>,
+    2,
+    8,
+>;
+type MaskEnc = ReedSolomonZkEncoding<F, Radix2DFTSmallBatch<F>>;
+
+/// ZK code message length. Bound by Lemma 6.4 (`ℓ_zk ≥ 2`). `ℓ_zk = 4` sits at
+/// paper-minimum-plus-margin; the precise `λ → ℓ_zk` mapping depends on the
+/// outer code's list size (see Lemma 6.5), not on this parameter alone.
+const ELL_ZK: usize = 4;
```
camofu: This constant could become a sweep variable for a more in depth bench effort. I though it was out of scope for this PR.

whir/benches/sumcheck.rs
```diff
@@ -87,10 +145,35 @@ fn run_sumcheck<L: Layout<F, EF>>(prover: L, challenger: &mut Challenger, foldin
...
+}
+
 fn bench_sumcheck_prover(c: &mut Criterion) {
     let mut group = c.benchmark_group("whir/layout_sumcheck");
 
-    let cases = [(14, 4, "log14"), (18, 4, "log18"), (22, 4, "log22")];
+    // `log20` is the issue #1586 acceptance-criterion point for the HVZK
+    // overhead bench (`2^20 rows, k = 4, λ = 100`); the criterion target is
+    // `zk / prefix ≤ 1.05×` at that label. `log14/18/22` give the scaling
+    // curve for prefix vs suffix vs zk.
```
camofu, line 172: Folding factor fixed by acceptance criteria (same as lambda)

whir/src/sumcheck/layout/prover/prefix.rs
```diff
@@ -404,7 +404,7 @@ impl<F: TwoAdicField, EF: ExtensionField<F>> PrefixProver<F, EF> {
 
     /// Builds the residual equality-weight polynomial after the prefix SVO rounds.
     #[tracing::instrument(skip_all)]
-    fn combine_eqs(&self, rs: &Point<EF>, alpha: EF) -> Poly<EF> {
+    pub(crate) fn combine_eqs(&self, rs: &Point<EF>, alpha: EF) -> Poly<EF> {
         assert_eq!(rs.num_variables(), self.folding);
```
camofu, line 408: The layout state is already pub(crate) from #1612; this PR adds sum and combine_eqs at the same level so the HVZK overlay can use them as shared kernels rather than re-deriving

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+/// HVZK's per-round polynomial has combined degree `max(ℓ_zk - 1, 2)`, so each
+/// round's wire payload carries `max(ℓ_zk - 1, 2)` field elements (after
+/// skipping `c_1`). The length is constant within a proof but only known at
+/// runtime (derived from `encoding.message_len()`), hence the inner `Vec<EF>`.
+#[derive(Debug, Clone)]
+pub struct ZkSumcheckData<F, EF> {
+    /// `μ̃ = Σ_{b ∈ {0,1}^k} ŝ(b)` — the prover's "new target" sent in step 2
+    /// of Construction 6.3. Observed on the transcript before the verifier
+    /// samples `ε`; lifted to `EF` at observation time.
+    pub mu_tilde: F,
```
camofu: `mu_tilde`: `F` not `EF` because the masks are F-valued (sublinear-mask design, paper §6). Lifted to `EF` at challenger-observe time on both prover and verifier sides

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+/// 2. Wrap into a [`PrefixProver`] via `PrefixProver::from_witness`.
+/// 3. Wrap into [`ZkPrefixProver`] via [`ZkPrefixProver::new`].
+/// 4. Register opening claims via [`Self::eval`] / [`Self::add_virtual_eval`]
+///    (both delegate to the inner `PrefixProver`).
+/// 5. Drive the HVZK sumcheck via [`Self::into_sumcheck`].
+pub struct ZkPrefixProver<F, EF, Enc, M>
+where
+    F: Field,
+    EF: ExtensionField<F>,
+    Enc: ZkEncoding<F> + Clone,
```
camofu: This clone could be avoided: only used for tests.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        assert!(
+            ell_zk >= 2,
+            "Construction 6.3 (Lemma 6.4) requires ell_zk >= 2",
+        );
+        assert!(k >= 1, "sumcheck requires at least one round");
+        assert!(
+            k <= n_vars,
+            "folding_factor must be <= poly.num_variables()",
+        );
+
```
camofu: These could also be `debug_assert!`

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        let mask_oracles: Vec<MaskOracle<F, Enc, M>> = masks
+            .iter()
+            .map(|mask| {
+                let codeword = encoding.encode(mask, rng);
+                let (commit, prover_data) = mmcs.commit_matrix(codeword);
+                challenger.observe(commit.clone());
+                (commit, prover_data)
+            })
+            .collect();
+
```
camofu: This is the same order that is coded in the simulator. The two must change in sync

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+                    };
+                    naive += s_l_eval;
+                }
+            }
+            debug_assert_eq!(
+                mu_tilde, naive,
+                "μ̃ closed form does not match naive Σ_{{b ∈ {{0,1}}^k}} ŝ(b)",
+            );
+        }
+
```
camofu: This debug assertion was inspired by the reference implementation

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        challenger.observe_algebra_element(EF::from(mu_tilde));
+        zk_data.mu_tilde = mu_tilde;
+
+        // --- Construction 6.3 step 3: ε ---
+        let eps: EF = challenger.sample_algebra_element();
+
```
camofu: ε ∈ EF (not ∈ F as in paper §6)

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+/// # Panics
+///
+/// Same precondition asserts as [`ZkPrefixProver::into_sumcheck`].
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::too_many_lines)]
+pub fn simulate_classic_unpacked<F, EF, Enc, M, Challenger, R>(
+    challenger: &mut Challenger,
+    folding_factor: usize,
+    pow_bits: usize,
+    mu: EF,
```
camofu: `mu: EF` is caller-precomputed because the simulator is witness-free by design (Lemma 6.4)

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    );
+    assert!(
+        ell_zk >= 2,
+        "Construction 6.3 (Lemma 6.4) requires ell_zk >= 2",
+    );
+    assert!(k >= 1, "sumcheck requires at least one round");
+
+    // The caller has pre-loaded the challenger with claim observations
+    // identical to the real prover's; `α` is the next thing sampled.
+    let _alpha: EF = challenger.sample_algebra_element();
```
camofu: this is just for fiat-shamir consistency

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+    #[test]
+    fn prover_verifier_roundtrip_classic_unpacked() {
+        // One concrete run with non-tiny parameters: small enough to be fast,
+        // big enough to exercise the per-round Lagrange-weighted accumulator
+        // loop across multiple rounds.
+        run_roundtrip(8, 3, 4, 2, 0).expect("honest roundtrip should accept");
+    }
+
+    proptest! {
```
camofu: Acceptance-criterion #1 from issue #1586

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        // absorbs it with the matching `add_virtual_eval(eval, …)`.
+        let mut prover_challenger = MyChallenger::new(perm.clone());
+        let mut verifier_challenger = MyChallenger::new(perm);
+
+        for _ in 0..num_eqs {
+            let eval = prover.add_virtual_eval(&mut prover_challenger);
+            verifier.add_virtual_eval(eval, &mut verifier_challenger);
+        }
+
+        let pow_bits = 0;
```
camofu: `pow_bits = 0` for the entire test surface.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        #![proptest_config(ProptestConfig::with_cases(16))]
+
+        /// The HVZK simulator (sumcheck transcript via
+        /// `simulate_classic_unpacked` plus mask oracle answers via
+        /// `ZkEncoding::simulate`) matches the real prover's view exactly
+        /// across the `(n_vars, k, ell_zk, num_eqs)` cube for RS encoding
+        /// (`error() == 0`). See `run_view_match_rs` for the concrete
+        /// invariants checked per case.
+        #[test]
+        fn prop_simulator_view_matches_real_rs(
```
camofu: Acceptance-criterion #2 from issue #1586

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    fn ef_in_f_subspace(x: EF) -> bool {
+        let coeffs = <EF as BasedVectorSpace<F>>::as_basis_coefficients_slice(&x);
+        coeffs[1..].iter().all(|c| *c == F::ZERO)
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(8))]
+
+        #[test]
+        fn prop_simulator_f_subspace_invariant(
```
camofu: Acceptance-criterion #2 from issue #1586 - supporting

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    // Not tested here: the quantitative `empirical_rate ≤ ε_j` match. With
+    // `γ_j` sampled from `EF ≈ 2^124` (frozen design decision: ε ∈ EF for
+    // soundness margin), the bound is ~2^-110 for any feasible params —
+    // unobservable at any trial count CI can host without a sub-2^31
+    // test-only field type, which `p3-field` doesn't ship.
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(8))]
+
+        #[test]
+        fn prop_rbr_tampering_changes_verifier_output(
```
camofu: Acceptance-criterion #3 from issue #1586

camofu:
@tcoratger I have left some comment for an easier review. The PR is ready to review :)

whir/benches/sumcheck.rs
tcoratger: Hey, just a thought on this parameter — the docstring says "paper-minimum-plus-margin", and looking at the criterion target `λ = 100` from #1586, I'm wondering if `ELL_ZK = 4` is actually the right operating point to measure against. Per Lemma 6.5 of the eprint, the soundness error has a `ℓ_zk · |Λ| / |F|` term, and the discussion in §2.7 sets `ℓ_zk = O(λ / log log λ)` for the Reed-Solomon mask code at rate 1/log λ — so for `λ = 100` you'd probably end up somewhere closer to 30-60.

The practical concern is that with `ℓ_zk = 4` the mask Merkle tree is over `m = 8` leaves, which is very small. The `≤ 1.05×` overhead might look easier than it actually is when masks are sized properly. Would be curious whether the bound still holds at e.g. `ELL_ZK = 30` — that'd be a stronger signal that the headline number reflects real-world HVZK-WHIR.

Not blocking, but maybe worth either bumping the constant or expanding the docstring to tie it to a concrete soundness regime? WDYT?

whir/benches/sumcheck.rs
```diff
@@ -19,6 +24,25 @@ type Challenger = DuplexChallenger<F, Perm, 16, 8>;
...
+    8,
+>;
+type MaskEnc = ReedSolomonZkEncoding<F, Radix2DFTSmallBatch<F>>;
+
+/// ZK code message length. Bound by Lemma 6.4 (`ℓ_zk ≥ 2`). `ℓ_zk = 4` sits at
+/// paper-minimum-plus-margin; the precise `λ → ℓ_zk` mapping depends on the
+/// outer code's list size (see Lemma 6.5), not on this parameter alone.
+const ELL_ZK: usize = 4;
+/// Reed-Solomon randomness symbols appended to each mask before encoding.
+const T_RAND: usize = 2;
```
tcoratger, line 48: Tiny related thought: `T_RAND = 2` means the mask encoding is 2-query zero-knowledge, which is fine for *this* sumcheck step (the verifier doesn't query mask oracles here). But when #1587 lands and the codeswitch starts opening mask oracles, the encoding's t-query budget will need to support however many openings the composed protocol does — likely closer to `O(λ)` than 2. Just a note for whoever picks this up downstream, no change needed here.

tcoratger (commented):
Hey @CarloModicaPortfolio! Did a fairly detailed pass through this — really nice work overall. The construction maps cleanly onto Construction 6.3, the closed-form for μ̃ is tight, the skip-c_1 wire optimization is a great fit for Lemma 6.4's rank-nullity argument, and the simulator stratification fix you and @latifkasuli landed reads correctly to me. Leaving a handful of line-level thoughts inline below, mostly around docs, footguns, and a couple of suggestions for downstream-friendliness when #1587 lands. None of it is blocking from my side — mostly curious to hear your take on the bigger ones (ELL_ZK in the bench, suffix-mode scope, and the ell_zk verifier parameter).

A couple of things from the existing conversation while I'm here:

- **PoW canonicality (Latif's May 5 follow-up):** looks like it's already covered by `expected_pow = if pow_bits > 0 { k } else { 0 }` plus the `PowWitnessCountMismatch` branch at `whir/src/sumcheck/zk.rs:631-637`. Calling that out for completeness so the thread isn't left dangling.

- **F-subspace stratification fix:** I worked through the hybrid F/EF picture and it lines up — high-index wire coordinates carry only the mask term (in F) and the simulator samples them from F-lifted-into-EF, so the trivial distinguisher from §6.1 is closed. There's still a subtle distributional thing in the joint distribution of wire[0..2] vs ε that I think is fine in isolation but worth pinning down before composition into #1587 — I left a longer note on the existing F-subspace test inline.

All the inline stuff below is suggestion-shaped, feel free to push back on any of it.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+///
+/// # Lifecycle
+///
+/// 1. Build a [`crate::sumcheck::layout::Witness`] from source tables.
+/// 2. Wrap into a [`PrefixProver`] via `PrefixProver::from_witness`.
+/// 3. Wrap into [`ZkPrefixProver`] via [`ZkPrefixProver::new`].
+/// 4. Register opening claims via [`Self::eval`] / [`Self::add_virtual_eval`]
+///    (both delegate to the inner `PrefixProver`).
+/// 5. Drive the HVZK sumcheck via [`Self::into_sumcheck`].
+pub struct ZkPrefixProver<F, EF, Enc, M>
```
tcoratger: One design question, completely open-ended: I see `ZkPrefixProver` but no `ZkSuffixProver`. Looking at `whir/src/pcs/tests.rs` the PCS is exercised with both Prefix and Suffix modes side by side (`run_keccak_end_to_end::<SuffixProver<...>>()` etc.), and the `Layout` trait is there precisely to let WHIR pick either at the PCS layer.

I'm curious whether the Suffix path is intentionally out of scope for HVZK in this PR (e.g., because production only uses Prefix), or just hasn't landed yet. Either is totally fine — but it might be worth either a module-level doc note explicitly marking HVZK-WHIR as Prefix-only with a tracking issue, or a follow-up brick. Otherwise it's easy to wire up HVZK at the PCS level only to realize the Suffix path silently has no privacy.

If the answer is "Prefix-only by design", I'd just add a `//!` paragraph at the top of `zk.rs` calling that out. WDYT?

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+            // Build `ĥ_j` of length `max(ell_zk, 3)`:
+            //   indices 0..ell_zk : live-mask piece           = 2^{k-j} · s_j(X)
+            //   index 0           : past-mask contribution   += 2^{k-j} · Σ_{l<j} s_l(γ_l)
+            //   index 0           : future-mask contribution += 2^{k-j-1} · Σ_{l>j}(s_l(0)+s_l(1))
+            //   indices 0..3      : plain piece              += ε · (c_0 + c_1·X + c_∞·X²)
+            let h_size = core::cmp::max(ell_zk, 3);
+            let mut h: Vec<EF> = vec![EF::ZERO; h_size];
+
+            let mult_live = F::TWO.exp_u64((k - j) as u64);
```
tcoratger: Tiny micro-perf thing: `F::TWO.exp_u64((k - j) as u64)` gets recomputed every round (and again inside the `#[cfg(debug_assertions)]` block at line 462-463). For `k` rounds that's `O(k log k)` field mults that could be precomputed once before the loop. Cheap absolute cost but easy fix if you feel like cleaning up:

```rust
let pow2: Vec<F> = (0..=k).map(|i| F::TWO.exp_u64(i as u64)).collect();
// in the loop: mult_live = pow2[k - j], mult_past = pow2[k - j + 1], mult_future = pow2[k - j - 1]
```

Also collapses the duplication with the debug block. NBD if you want to leave it.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        // `scale` argument; the residual product polynomial then satisfies
+        // `dot_product() == ε · plain_residual_sum`.
+        let compressed = tracing::info_span!("compress_prefix_to_packed")
+            .in_scope(|| inner.poly.compress_prefix_to_packed(&rs, eps));
+        let weights = inner.combine_eqs(&rs, alpha).pack::<F, EF>();
+        let prod_poly =
+            ProductPolynomial::<F, EF>::new_packed(VariableOrder::Prefix, compressed, weights);
+
+        let residual_sum = eps * plain_sum;
+        debug_assert_eq!(
```
tcoratger: One contract-carried-in-comments thing worth flagging for #1587: the residual `SumcheckProver` returned from here has `dot_product == ε · plain_residual_sum`, not `plain_residual_sum`. The only signal of the ε scaling today is the docstring at lines 296-305 and this `debug_assert_eq!`. From a `SumcheckProver` type perspective the two look identical.

Probably fine for this PR but I think the next consumer (codeswitch) is going to need to know about this ε factor and it's easy to forget. Could be as light as returning the ε alongside (`(SumcheckProver, ε, Point, Vec<MaskOracle>)`) or as heavy as a newtype `EpsScaledSumcheck`. Worth a thought when wiring up #1587. Not blocking on this PR.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    /// # Panics
+    ///
+    /// - If `char(F) == 2` or `ell_zk < 2` (Lemma 6.4 hypotheses).
+    /// - If `folding_factor == 0`.
+    #[allow(clippy::too_many_arguments)]
+    pub fn into_sumcheck<M, Ch>(
+        self,
+        zk_data: &ZkSumcheckData<F, EF>,
+        mask_commits: &[M::Commitment],
+        ell_zk: usize,
```
tcoratger: One sharp edge here: `ell_zk` is taken as a bare parameter on the verifier side, but derived from `self.encoding.message_len()` on the prover side. If the verifier-side caller picks a different value, the only thing that catches it is the wire-size shape check — but that check is non-injective. Specifically `wire_size = max(ell_zk, 3) - 1`, so `ell_zk = 2` and `ell_zk = 3` both produce `wire_size = 2`. The verifier can't distinguish them, and the `target` it computes would silently differ if prover and verifier disagreed in that range.

What if we embedded `ell_zk` into `ZkSumcheckData` itself?

```rust
pub struct ZkSumcheckData<F, EF> {
    pub mu_tilde: F,
    pub ell_zk: usize,                    // new
    pub round_coefficients: Vec<Vec<EF>>,
    pub pow_witnesses: Vec<F>,
}
```

Verifier then checks `zk_data.ell_zk == expected_ell_zk` upfront with a dedicated `EllZkMismatch` error, and the shape check degenerates to "derived wire size matches". One usize per proof, takes the footgun off the table. Just a suggestion, current shape works as long as the caller is careful.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        Ok((Point::new(randomness), target))
+    }
+}
+
+/// Returns the layout strategy `PrefixProver` uses.
+///
+/// Free function rather than `PrefixProver::<F, EF>::strategy()` because the
+/// trait-bound `Layout` requires `F: TwoAdicField`, which the verifier doesn't
+/// inherit. The value is fixed by construction.
+pub const fn prefix_strategy() -> LayoutStrategy {
```
tcoratger: Tiny robustness thought: this duplicates the value returned by `<PrefixProver as Layout>::strategy()` at `prover/prefix.rs:353`. The doc explains why (the trait-bound on `TwoAdicField` doesn't apply to the verifier), which makes sense, but I'm a little worried about silent drift — if anyone ever flips the bool on the inner `strategy()` without touching this one, the ZK verifier would read claims with the wrong selector lift, and the bug would only surface at the round-trip level. Hard to bisect later.

Would a small test pinning `prefix_strategy() == PrefixProver::strategy()` make sense as a guard? Something like:

```rust
#[test]
fn prefix_strategy_matches_prefix_prover() {
    use crate::sumcheck::layout::Layout;
    assert_eq!(
        prefix_strategy(),
        <PrefixProver<F, EF> as Layout<F, EF>>::strategy(),
    );
}
```

Alternatively, could we expose the inner strategy without the `TwoAdicField` bound somehow? Either way, just an idea.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+) -> (ZkSumcheckData<F, EF>, Vec<M::Commitment>, Point<EF>)
+where
+    F: Field,
+    EF: ExtensionField<F>,
+    Enc: ZkEncoding<F>,
+    Enc::Codeword: Matrix<F>,
+    M: Mmcs<F>,
+    Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
+    R: Rng,
+    StandardUniform: Distribution<F> + Distribution<EF>,
```
tcoratger: One usability thought on `simulate_classic_unpacked`: it takes `mu: EF` as a pre-computed parameter, with the convention that the caller has computed it from the same `α` the simulator will sample internally. That works (the tests do a clever fork-clone-peek trick to get α without consuming it), but it's a pretty subtle ritual to get right, and the caller-side μ computation duplicates `Verifier::sum`'s logic — so if `Verifier::sum` ever changes the simulator callers could silently produce wrong μ.

Have you considered passing the `ZkVerifier` (or its inner) in and letting the simulator sample α and compute μ itself?

```rust
pub fn simulate_classic_unpacked<...>(
    challenger: &mut Challenger,
    verifier: &ZkVerifier<F, EF>,    // new
    folding_factor: usize,
    pow_bits: usize,
    encoding: &Enc,
    mmcs: &M,
    rng: &mut R,
) -> (...)
```

With `let alpha = challenger.sample_algebra_element(); let mu = verifier.inner.sum(alpha);` inside. Would also turn the discarded `_alpha` at line 772 into a real used value. Totally a taste call though.

whir/src/sumcheck/zk.rs
tcoratger: Small readability nit — this reads as "unused, can be deleted" to a future reader, even though the docstring above explains it's load-bearing for challenger sync. I wonder if `let _: EF = challenger.sample_algebra_element();` plus a slightly more emphatic comment would help guard against an accidental cleanup-style refactor that strips it? Something like:

```rust
// CRITICAL: this sample is required for challenger-state sync with the real
// prover, which samples α at this point. Discarding the value is correct —
// μ is computed externally — but the challenger draw itself is load-bearing.
let _: EF = challenger.sample_algebra_element();
```

(Tied to the suggestion above — if `mu` is computed internally, this becomes a real used variable and the concern goes away.)

whir/src/sumcheck/zk.rs
tcoratger: Reading through the tests, I noticed every case here exclusively uses `add_virtual_eval`. `add_claim` (concrete openings) isn't exercised in the ZK module at all. But the prover's `into_sumcheck` accumulator loop (lines 305-313 + 390-399) explicitly handles the concrete-claims branch, and the α-power split between concrete and virtual claims (`alpha.powers().skip(n_claims)`) is a place where subtle bugs could hide.

Since WHIR PCS in production uses concrete openings (the whole `prover.eval(...)` flow), I think adding even one proptest that records a concrete claim via `prover.eval(0, &[0], ...)` and mirrors with `verifier.add_claim(...)` would close a real coverage gap. Probably cheap to wedge in given the existing `run_roundtrip` skeleton. Happy to help sketch it out if useful!

Separately — also noticed the `ell_zk` sweep stops at 5 here, but real HVZK-WHIR will use `ℓ_zk` more in the 30-60 range. Adding one fixed-large case (e.g., `ell_zk = 32`) would exercise the longer mask Horner path. Cheap addition.

whir/src/sumcheck/zk.rs
tcoratger: Is there any alternative solution? I'm not a big fan of keeping things for test purposes only.

whir/benches/sumcheck.rs
tcoratger: Totally agree it's out of scope as a full sweep. The thing I'd push back on slightly is the framing of the docstring — "paper-minimum-plus-margin" reads like `ℓ_zk = 4` is a *meaningful* HVZK-100 point, when really it's the floor. For a future reader who reaches for this bench number to claim "≤ 1.05× overhead at λ = 100" without reading the full paper, that could mislead. Would you be open to tightening the doc to something like "smallest value satisfying Lemma 6.4's floor (ℓ_zk ≥ 2) plus a small margin; a realistic λ = 100 setup needs ℓ_zk closer to ~30 per §2.7"? Sweep itself can absolutely stay for a follow-up.

whir/benches/sumcheck.rs
tcoratger: Ack — fixing `k = 4` to match #1586's acceptance criterion makes sense. Good.

whir/src/sumcheck/layout/prover/prefix.rs
tcoratger: Yep, that lines up with how #1612 set things up — sharing the kernel is much cleaner than duplicating the Lagrange-weighted accumulator math. One small thing: would it be worth a doc line on `sum` and `combine_eqs` saying "exposed at `pub(crate)` for the HVZK overlay in `sumcheck::zk`" so future readers immediately see why these aren't private? Helps the next person resist the temptation to tighten them back to `fn`.

whir/src/sumcheck/zk.rs
tcoratger: Yep, that's exactly right — masks live in F, μ̃ is just a sum of mask coefficients so it's also in F. Storing as F also saves 3 field elements per proof under the quartic extension, which is a free win. Would be nice to spell that out in the doc comment so the asymmetry (stored as F, observed as EF, used as EF in the target) doesn't look accidental to a future reader.

whir/src/sumcheck/zk.rs
tcoratger: Picking up the thread here — the cleanest fix I can think of is just moving the `Clone` bound off the struct definition and onto whichever impl actually needs it. Right now `pub struct ZkPrefixProver<F, EF, Enc, M> where Enc: ZkEncoding<F> + Clone` pollutes every caller, even though production never clones (`into_sumcheck` consumes `self`). The Rust convention is to keep struct definitions bound-free and push bounds onto the impls that need them.

So roughly:

```rust
pub struct ZkPrefixProver<F, EF, Enc, M>
where
    F: Field,
    EF: ExtensionField<F>,
    Enc: ZkEncoding<F>,        // no Clone here
    M: Mmcs<F>,
{ ... }

// Only the impl block that uses .clone() (if any) needs the bound;
// tests can rely on `MaskEnc: Clone` inferred locally.
```

If the bench/tests need to clone the encoding to fan it out, that's a property they can require themselves — no need to lift it into the struct's type contract. Want me to sketch the full diff?

whir/src/sumcheck/zk.rs
tcoratger: I'd be a little careful here — there's a stylistic split:

- `char(F) != 2` and `ell_zk >= 2` are type-/encoding-level invariants that should hold by construction. `debug_assert!` is defensible.
- `k >= 1` and `k <= n_vars` are derived from `self.inner.folding` and `self.inner.num_variables`, which were set when the inner `PrefixProver` was built. So a violation here means the inner prover was constructed wrong upstream — also an invariant, also `debug_assert!`-shaped.

The flip side is defense-in-depth: at `O(usize)` cost the runtime checks guard against a misconfigured upstream that happened to compile. I'd lean toward leaving them as `assert!` in the prover entry (where producing a broken proof is the worst-case) and downgrading in the simulator/verifier where they're more redundant. But I'd be fine with full `debug_assert!` too — your call.

whir/src/sumcheck/zk.rs
tcoratger: Good thing to flag. Right now the sync is enforced by convention only — a docstring on either side would help, but even better: could the mask-sampling-and-commit block be factored into a small helper like `fn sample_and_commit_masks(...) -> (Vec<Vec<F>>, Vec<MaskOracle<F, Enc, M>>)` (or `Vec<M::Commitment>` for the simulator variant) that both call sites use? Then "the two must stay in sync" becomes "they literally are the same code". Not a hard requirement, just a thought.

whir/src/sumcheck/zk.rs
tcoratger: Yeah this is a really nice sanity check — keeps the closed-form formula honest, and the cost is hidden in release builds. No change requested, just noting that I appreciate it being here. The reference-implementation style of pinning the closed-form via a naive check is good practice.

whir/src/sumcheck/zk.rs
tcoratger: This is the load-bearing divergence from paper §6 — worth a more prominent comment IMO, because it's the source of the hybrid F/EF setting that motivates the F-subspace stratification fix you added for the simulator. Something like:

```rust
// Plonky3 samples ε ∈ EF (not ∈ F as in paper Construction 6.3 step 3) for
// the soundness margin. Masks remain in F to keep the proof size sublinear,
// so the per-round h_j(X) is a hybrid F/EF object — see the F-subspace
// stratification in simulate_classic_unpacked for the simulator-side
// implication of this divergence.
let eps: EF = challenger.sample_algebra_element();
```

We can probably make the doc more reader friendly and concise but that's the idea.

whir/src/sumcheck/zk.rs
tcoratger: Right, the simulator being witness-free is the load-bearing property of Lemma 6.4 — completely agree, that has to stay. The thing I wanted to push on with my inline at line 755 isn't "give the simulator the witness", it's "give the simulator the `ZkVerifier`, which already holds the *claims* (not the witness)". Then `mu = verifier.inner.sum(alpha)` inside the simulator is still witness-free — it only sums the verifier's recorded claim evaluations — but removes the caller-side fork-clone-peek choreography. Wanted to make sure that framing came through, because witness-freeness is preserved either way.

whir/src/sumcheck/zk.rs
tcoratger: Yep, FS sync is exactly the role. My only nit (left inline at line 772) was the naming: `_alpha` looks like "dead code, can be deleted" to a future reader, even though the comment above explains the role. `let _: EF = challenger.sample_algebra_element();` with a slightly louder comment ("CRITICAL: this sample is required for challenger-state sync") would lower the risk of an over-zealous refactor stripping it. Same intent, less footgun-shaped.

whir/src/sumcheck/zk.rs
tcoratger: Good cross-reference — thanks for linking back to #1586's acceptance criterion. The case here (n_vars=8, k=3, ell_zk=4, num_eqs=2) is a nice middle-of-the-cube smoke test for AC1. Combined with the proptest sweep below, AC1 is well covered.

whir/src/sumcheck/zk.rs
tcoratger: One small follow-up here: with `pow_bits = 0` everywhere, the `challenger.grind` branch at zk.rs:485-487 and the `check_witness` branch at zk.rs:674-676 are never executed by tests. That's a coverage gap — the PoW path is small but has a real failure mode (mismatched/forged PoW witnesses) that would be nice to pin down.

Would you mind adding even one case (say `pow_bits = 4`, just enough to grind) to one of the existing proptests? Doesn't have to be a sweep — a single fixed positive value would catch any wiring bug in the grind/check_witness pairing.

whir/src/sumcheck/zk.rs
tcoratger: Acknowledged — this is the main AC2 driver. As I noted in the higher-level review, the test pins down (a) verifier acceptance on both, (b) μ̃ / mask commits match under coupled seeds, (c) F-subspace stratification on wire[i≥2], (d) `ZkEncoding::simulate` query-count. The one thing it doesn't check is distributional equality of wire[0..2], which is where the hybrid F/EF setting puts a subtle gap. I don't think that's exploitable in this isolated step, but worth pinning down before #1587 composes the simulator's view with the codeswitch.

whir/src/sumcheck/zk.rs
tcoratger: Yep, this is the right companion test — it's the test that specifically catches the trivial distinguisher from paper §6.1 ("is wire[i ≥ 2] in the F-subspace of EF?"). Glad you split it out from `prop_simulator_view_matches_real_rs`; reads cleanly.

whir/src/sumcheck/zk.rs
tcoratger: Ack. As the PR description argues, the quantitative `empirical_rate ≤ ε_j` match isn't observable at CI scale because γ_j ∈ EF ≈ 2^124 puts the bound at ~2^-110. This tampering-propagation test is the right conformance substitute — it asserts the protocol's structure (any byte change in any wire propagates through Fiat-Shamir to a different target), which is what an implementation *can* test about a theorem-about-the-abstract-protocol. Coverage rationale in the PR description is solid.

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        self.inner.num_variables
+    }
+
+    /// Records concrete opening claims; delegates to `PrefixProver::eval`.
+    pub fn eval<Ch>(&mut self, table_idx: usize, polys: &[usize], challenger: &mut Ch) -> Vec<EF>
+    where
+        F: TwoAdicField,
+        Ch: FieldChallenger<F> + GrindingChallenger<Witness = F>,
+    {
+        use crate::sumcheck::layout::Layout;
```
tcoratger: Imports like this should go at the top of the file right?

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        self.inner.eval(table_idx, polys, challenger)
+    }
+
+    /// Records a virtual opening claim; delegates to `PrefixProver::add_virtual_eval`.
+    pub fn add_virtual_eval<Ch>(&mut self, challenger: &mut Ch) -> EF
+    where
+        F: TwoAdicField,
+        Ch: FieldChallenger<F> + GrindingChallenger<Witness = F>,
+    {
+        use crate::sumcheck::layout::Layout;
```
tcoratger: Same for the import here :)

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    /// - Mask oracles (commitment + prover data per mask), in mask order.
+    ///
+    /// # Panics
+    ///
+    /// - If `char(F) == 2` (Lemma 6.4).
+    /// - If `encoding.message_len() < 2` (Lemma 6.4).
+    /// - If `self.folding() == 0` or `> self.num_variables()`.
+    #[allow(clippy::too_many_arguments)]
+    #[allow(clippy::too_many_lines)]
+    #[allow(clippy::type_complexity)]
```
tcoratger: We can have that into a single allow with commas

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        Ch: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
+    {
+        let k = self.inner.folding;
+        let ell_zk = self.encoding.message_len();
+        let n_vars = self.inner.num_variables;
+
+        assert!(
+            F::TWO != F::ZERO,
+            "Construction 6.3 (Lemma 6.4) requires char(F) != 2",
+        );
```
tcoratger:
```suggestion
        const { assert!(
            F::TWO != F::ZERO,
            "Construction 6.3 (Lemma 6.4) requires char(F) != 2",
        ); }
```

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+            encoding,
+            mmcs,
+        } = self;
+
+        // --- Plain-piece preamble (mirrors PrefixProver::into_sumcheck) ---
+        let alpha: EF = challenger.sample_algebra_element();
+        let n_claims = {
+            use crate::sumcheck::layout::Layout;
+            inner.num_claims()
+        };
```
tcoratger:
```suggestion
        let n_claims = inner.num_claims()
```

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        let mut alphas = alpha.powers();
+        let accumulators: Vec<_> = inner
+            .placements
+            .iter()
+            .flat_map(|placement| inner.claim_map[placement.idx()].iter())
+            .map(|claim| {
+                let per_claim: Vec<EF> = alphas.by_ref().take(claim.len()).collect();
```
tcoratger: can't we use collect_n() here?

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    where
+        M: Mmcs<F>,
+        Ch: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
+    {
+        let k = folding_factor;
+
+        assert!(
+            F::TWO != F::ZERO,
+            "Construction 6.3 (Lemma 6.4) requires char(F) != 2",
+        );
```
tcoratger:
```suggestion
        const { assert!(
            F::TWO != F::ZERO,
            "Construction 6.3 (Lemma 6.4) requires char(F) != 2",
        ); }
```

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+        ell_zk: usize,
+        folding_factor: usize,
+        pow_bits: usize,
+        challenger: &mut Ch,
+    ) -> Result<(Point<EF>, EF), SumcheckError>
+    where
+        M: Mmcs<F>,
+        Ch: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
+    {
+        let k = folding_factor;
```
tcoratger: We should not need it and use automatically `folding_factor` no?

```suggestion
```

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+            });
+        }
+        let expected_pow = if pow_bits > 0 { k } else { 0 };
+        if zk_data.pow_witnesses.len() != expected_pow {
+            return Err(SumcheckError::PowWitnessCountMismatch {
+                expected: expected_pow,
+                actual: zk_data.pow_witnesses.len(),
+            });
+        }
+        let h_size = core::cmp::max(ell_zk, 3);
```
tcoratger:
```suggestion
        let h_size = ell_zk.max(3);
```

whir/src/sumcheck/zk.rs
```diff
@@ -0,0 +1,1440 @@
...
+    F: Field,
+    EF: ExtensionField<F>,
+    Enc: ZkEncoding<F>,
+    Enc::Codeword: Matrix<F>,
+    M: Mmcs<F>,
+    Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
+    R: Rng,
+    StandardUniform: Distribution<F> + Distribution<EF>,
+{
+    let k = folding_factor;
```
tcoratger: Same can't we use directly `folding_factor` instead?

```suggestion
```

whir/src/sumcheck/zk.rs
camofu: Addressed in Commit: 0f2cb39f , I agree it is better to get rid of the clone there.

whir/src/sumcheck/zk.rs
camofu: Commit: 0f2cb39f 
I went a bit further and also added a test scenario for `SumcheckError::InvalidPowWitness` so that every branch is properly tested.

whir/src/sumcheck/zk.rs
camofu: Addressed in commit 41f72d9

whir/src/sumcheck/zk.rs
camofu: Commit [972ecaf](https://github.com/Plonky3/Plonky3/pull/1605/commits/972ecaf4d57fa5ffbdac82df73748fae14448b7b)

whir/src/sumcheck/zk.rs
camofu: Yes, I think `ZkSuffixProver` is out of scope for this PR, because it is not used here. In my mind it was something that belonged to [HVZK-WHIR 6/6], however makes sense to open a new issue to flag the need to implement this, as the original issues were made before the code re-structuring. It makes sense to have it as its own PR, as it is quite a bit of code that needs to be wrote and here and this PR is already quite big. So separating the concerns sounds like a good idea to me. 

I have create a separate issue, let me know what you think, I'm happy to trash it if it's incorrect/you would rather not open a new isssue

https://github.com/Plonky3/Plonky3/issues/1649

In 28af3a30  I added the paragraph referencing the issue, needs to be changed if we bin the issue itself

whir/src/sumcheck/zk.rs
camofu: [761cea0](https://github.com/Plonky3/Plonky3/pull/1605/commits/761cea088a7943009fe22f6c14bd61d1a9f35c09)

whir/src/sumcheck/zk.rs
camofu: [761cea0](https://github.com/Plonky3/Plonky3/pull/1605/commits/761cea088a7943009fe22f6c14bd61d1a9f35c09)

whir/src/sumcheck/zk.rs
camofu: Resolved, don't remember which commit though

whir/src/sumcheck/zk.rs
camofu: Actually got an error from this: E0015

The reason: `F::TWO != F::ZERO` calls `<F as PartialEq>::ne(...)`, which is a trait method. In stable Rust, trait methods aren't `const`-callable — the `const_trait_impl` feature is still nightly-only. The `Field` trait declares `TWO` and `ZERO` as `const Self` (`field/src/field.rs:80, 97`), but `Self` is a generic type parameter, not a concrete literal, so `!=` between two `const Self` values still goes through the trait-method `PartialEq::ne` path. `const { ... }` requires every operation inside to be const-evaluable, which trait calls aren't. That's the `E0015` you're seeing.

There's no `F::CHARACTERISTIC` const on the trait either — only `TWO_ADICITY`. So short of switching to nightly + `feature(const_trait_impl)`, the assertion has to stay runtime.


thoughts?

whir/benches/sumcheck.rs
camofu: Makes perfect sense to me. Actually bumped it to 37 as  100 / log₂(log₂(100)) ≈ 37, and expanded the docstring to explain why 37 was the chosen number.




758e2cd

camofu: (we are still in the 1.05 limit)

whir/src/sumcheck/zk.rs
camofu: Unfortunately no: `collect_n` consumes `self`. However, we can do something similar:

`Powers::collect_n` takes `self` by value, so it can't replace the per-claim `alphas.by_ref().take(claim.len()).collect()` — that pattern depends on keeping the iterator alive across claims.

But the same idea applies at the virtual-claim site below: the round loop was rebuilding `alpha.powers().skip(n_claims)` every iteration. I've hoisted those into a single `virtual_alphas: Vec<EF>` before the loop and the round body now reads them by slice. commit 508640f4.

whir/src/sumcheck/zk.rs
camofu: Applied change in 
20870f4

whir/src/sumcheck/zk.rs
camofu: Yes, I think this is for better code hygiene. I have updated the code and added a relevant test. However, I am not sure at this PR's level the targets would actually differ. Both `ell_zk = 2` and `ell_zk = 3` produce `wire_size = 2`, accept the same
  wire bytes, and compute identical targets. They're effectively the same protocol instantiation at this layer.

b5618d7f

whir/src/sumcheck/zk.rs
camofu: Yes, let's put it in. There is a real drift risk and the test to fix is so tiny. Exposing the inner strategy however is more cumbersome and I think we can skip it with the test in place. 

cac2c19a

whir/src/sumcheck/zk.rs
camofu: I would push back on this. I don't think is right to make the simulator dependent on a specific verifier implementation. It should describe what an honest view looks like, not how some specific verifier checks it. I think right now the simulator is textbook shaped. 

On the silent-μ-divergence concern: callers could produce a wrong μ, but no test invariant actually catches that. The simulator only needs μ to be internally consistent (target = ε·μ + μ̃), which works for any value. So even if `Verifier::sum` changed tomorrow, the tests would keep passing

whir/src/sumcheck/zk.rs
camofu: Good catch! Thank you for noticing it.  `3662f152`.

**Concrete claims:** `run_roundtrip` now takes `num_concrete` and `num_virtual` independently. Concrete openings recorded via `prover.eval(0, &[0], …)` mirrored with `verifier.add_claim(0, &[0], &openings, …)`. The fixed concrete test bumped from `(0, 2)` to `(1, 1)` so the α-power split (`skip(n_claims)`) does real work every round, and `prop_completeness_classic_unpacked` now sweeps both `0..=2` with `prop_assume!(num_concrete + num_virtual >= 1)`.

**Long Horner:** added `long_mask_horner_path` running at `ell_zk = 32` — keeps the proptest sweep capped at 5 for runtime, but pins one signal at the realistic regime (consistent with the `ELL_ZK_BENCH = 37` bump in `daa1fc19`).

whir/src/sumcheck/zk.rs
tcoratger: Let's keep as is for now.

tcoratger:
@latifkasuli After https://github.com/Plonky3/Plonky3/pull/1641, it looks like runs for WHIR are failing, see for example https://github.com/Plonky3/Plonky3/actions/runs/25863362250/job/75998794545?pr=1605, can you investigate?

Looks like we need to relaunch them a second time for them to succeed.

tcoratger (approved):
Thanks a lot @CarloModicaPortfolio.

I've just splitted the big zk.rs file into smaller files into a module to make the module more readable as a whole. I've also changed a bit the documentation to be more developer friendly and sometimes avoid to have big text blocks that could be difficult to digest. 

LGTM.

tcoratger:
@latifkasuli Also I feel there is something else wrong with the CI here since when the huge tests are triggered, the auto-merge doesn't wait for them to complete before merging, which should be the case if they are triggered.

latifkasuli:
Thanks for flagging this. I dug through the runs more carefully.

The job linked here is actually normal `CI / Build and Test (macos-latest)`, not the new `Heavy CI` workflow. It failed before compiling tests:

```text
cargo test
error: unexpected argument 'test' found
Usage: rustup-init[EXE] [OPTIONS]
```

That looks like a macOS toolchain/cache issue: the failure happens immediately after `Swatinem/rust-cache` restores `/Users/runner/.cargo/bin`, and `cargo` appears to resolve to `rustup-init`. The rerun passed. The corresponding Heavy CI run for the same SHA passed both exhaustive WHIR jobs.

I also checked earlier Heavy CI failures on this PR. Those were from intermediate broken commits in #1605 itself, for example compile errors in `whir/src/sumcheck/zk.rs`, not from #1641.

Your merge-gating concern is valid though. #1605 merged at 13:55 UTC, while Heavy CI completed later at 14:01 and 14:14 UTC, so Heavy CI is triggered but not blocking merge yet.

I think the follow-up fixes are:

1. Set `cache-bin: false` on `Swatinem/rust-cache@v2` to avoid restoring toolchain shims from `~/.cargo/bin`.
2. Add or require a stable Heavy CI gate check so PRs touching `whir/src/**` cannot merge before Heavy CI completes. Requiring the matrix jobs directly may be awkward because the workflow is path-filtered.

PR #1607, opened 2026-05-04
bench(whir): head-to-head FRI vs WHIR PCS comparison across Poseidon1 / Poseidon2 / Blake3

tcoratger:
## Summary

Adds a Criterion benchmark in the `p3-whir` crate that compares plonky3's FRI and WHIR head-to-head as polynomial commitment schemes, on a workload where both protocols answer the same proof claim.

Configuration: KoalaBear field, Poseidon1 Merkle, message size `2^22`, code rate `ρ = 1/2`, 100-bit capacity-regime soundness.

Files:

* `whir/benches/fri_vs_whir.rs` — the bench.
* `whir/Cargo.toml` — `p3-fri` and `postcard` added to `[dev-dependencies]`; new `[[bench]]` entry.

No source changes outside the bench file. No public API touched.

## What is being compared

Both protocols arrange the `2^22` committed elements as `256 = 2^8` polynomials of size `2^14`, and open all 256 at one common point.

`fri-batch`:
- 256 univariates of degree `< 2^14`.
- one common opening point `z ∈ EF`.
- reveals 256 EF scalars `v_i = f_i(z)` for `i ∈ {0..256}`.

`whir`:
- 1 multilinear in 22 variables (equivalently: 256 stacked 14-variable multilinears, indexed by the first 8 coordinates).
- 256 opening points sharing the trailing `ζ ∈ EF^14`.
- reveals 256 EF scalars `v_i = p̃(i_binary, ζ)` for `i ∈ {0..256}`.

The two claims are identified by the WHIR paper §1.1 univariate-multilinear bridge:

```text
    ζ = (z, z^2, z^4, …, z^{2^13})    ⇒    f_i(z) = p̃(i_binary, ζ)
```

So both protocols commit `2^22` field elements, open 256 polynomials, and reveal 256 EF scalars at one common point — same data shape, same opening shape, same scalar count.

## Matched substrate

| Knob | Value | Notes |
|---|---|---|
| Base field | `KoalaBear` | 31-bit prime, large 2-adic subgroup |
| Extension | `BinomialExtensionField<F, 4>` | challenge field |
| Message size | `2^22` field elements | matches WHIR paper Table 2 first row |
| Code rate | `ρ = 2^-1` (`log_blowup = 1`) | most common production setting |
| Security level | 100 bits | capacity-regime conjecture |
| PoW bits | 22 | smallest budget that lets WHIR's per-round PoW reach the target |
| Hash | Poseidon1 (Merkle leaves + compression) | |
| Challenger | `DuplexChallenger<F, Perm16, 16, 8>` | |
| DFT | `Radix2DFTSmallBatch<F>` | |
| RNG seed | fixed `0xA17_5C0DE` | deterministic |

## Per-protocol knobs (each protocol gets its recommended value)

* **WHIR** — `FoldingFactor::Constant(4)` (paper-recommended `k = 4`), `rs_domain_initial_reduction_factor = 1`. Per-round query counts, OOD samples and PoW are derived by `WhirConfig::new` from `(security_level, soundness_type, log_blowup, k)`.
* **FRI** — binary folding (`max_log_arity = 1`). The minimal query count comes from the soundness budget `log_blowup * num_queries + query_pow_bits ≥ security_level`, giving `num_queries = 78`.

## Soundness framing

Reed–Solomon proximity testing has three regimes:

| Regime | Status | Proximity bound |
|---|---|---|
| Unique-decoding | proven | `(1 - ρ) / 2` |
| Johnson | proven | `1 - √ρ` |
| Capacity | conjectured | `1 - ρ - η` |

plonky3 FRI's `conjectured_soundness_bits` formula corresponds to the **capacity regime** (the ethSTARK conjecture). plonky3 does not surface a proven (unique-decoding or Johnson) FRI query count.

To keep the comparison fair, **WHIR is configured at `SecurityAssumption::CapacityBound`**, so both protocols claim soundness under the same conjecture.

## Internal batching cost

Both protocols batch their 256 openings via a single random EF challenge:

* FRI draws `α ∈ EF` and folds the 256 univariates into one combined polynomial `g(X) = Σ α^i · f_i(X)` before running FRI on `g`.
* WHIR draws `γ ∈ EF` and combines the 256 evaluation constraints into one eq-constraint polynomial before running its sumcheck.

Soundness loss is `log_2(256 / |EF|) ≈ -116` bits in either case — negligible against `|EF| ≈ 2^124`. Neither protocol pays a "256× cost" in its prover loop; both pay one DFT, one Merkle commit, and one folding/sumcheck pass over the `2^22`-element codeword.

## What the bench produces

1. **Diagnostic table** (single trial, printed once at startup) — per-phase wall-clock breakdown, proof bytes, derived query counts.
2. **Criterion `prove` group** (`sample_size = 10`) — total prover time per protocol.
3. **Criterion `verify` group** (default `sample_size = 100`) — verification time per protocol.

## Bench results

All numbers from a fresh run on Apple M1 Pro, 10 cores, 16 GB RAM, macOS 26.4 (Darwin 25.4 arm64), `rustc 1.96.0-nightly d9563937f`, `cargo bench` (release profile, no `parallel` feature).

WHIR per-round queries at `m = 22`: `[84, 20, 12]` with 256 evaluation constraints. FRI: 78 single-shot queries over the 256-column matrix.

### Headline at `m = 22`, Poseidon1

| Metric | `fri-batch` | `whir` | Winner |
|---|---:|---:|---|
| Prover (Criterion mean) | **3.590 s** | 7.944 s | FRI **2.21×** faster |
| Verifier (Criterion mean) | 23.353 ms | **4.697 ms** | WHIR **4.97×** faster |
| Proof size (postcard) | 476 KiB | **109 KiB** | WHIR **4.36×** smaller |

### Diagnostic — single trial

| proto | commit ms | open ms | verify μs | proof bytes |
|---|---:|---:|---:|---:|
| `fri-batch` | 506 | 3 071 | 23 605 | 487 862 |
| `whir` | 780 | 7 204 | 4 580 | 111 958 |

### Criterion — Prover (sample size 10)

| Protocol | Mean (s) |
|---|---:|
| `fri-batch` | 3.590 |
| `whir` | 7.944 |

### Criterion — Verifier (sample size 100)

| Protocol | Mean (ms) |
|---|---:|
| `fri-batch` | 23.353 |
| `whir` | 4.697 |

## Note on the `commit` phase

The diagnostic shows WHIR's commit at ~780 ms and FRI's at ~510 ms. The two `commit` calls produce semantically equivalent objects (a Merkle root binding the prover to a Reed–Solomon codeword), but the work each one does differs:

| Step | FRI's `commit` | WHIR's `commit` |
|---|---|---|
| LDE encoding | yes | yes |
| Merkle tree commit | yes | yes |
| Sample OOD points + observe answers | no (happens at the start of `pcs.open`) | yes (`commitment_ood_samples` rounds) |
| Merkle leaves | `2^(log_n + log_blowup - log_width)` leaves of `2^log_width` field elements | `2^(log_n + log_blowup - k)` leaves of `2^k` field elements (`k = WHIR_FOLDING_FACTOR = 4`) |

The OOD step appearing in WHIR's commit but not FRI's is a labelling difference, not a protocol cost difference: FRI does its DEEP/zeta sampling at the start of `pcs.open`, WHIR does its OOD samples in `commit`. For total prover (commit + open), the labelling washes out.

## Key takeaways

1. **Same proof claim, two prover/verifier trade-offs.** Both protocols answer the same 256-evaluation claim on the same `2^22`-element message. The trade-off lives entirely in commit-and-open work versus verifier work and proof size.
2. **FRI wins on prover time.** At `m = 22`, Poseidon1, `fri-batch` proves the claim in ~3.6 s versus ~7.9 s for WHIR (2.21× faster).
3. **WHIR wins on verifier time and proof size.** WHIR verifies in ~4.7 ms versus ~23.4 ms for FRI (4.97× faster), with a proof ~4.4× smaller.

## How to run

```bash
cargo bench -p p3-whir --bench fri_vs_whir
```

Total wall-clock on the reference machine: under 5 minutes (1 hash × 1 m × 2 protocols).

## Test plan

- [x] `cargo build -p p3-whir --benches --release` — builds clean.
- [x] `cargo clippy -p p3-whir --all-targets -- -D warnings` — lints clean.
- [x] `cargo fmt -p p3-whir` — formatted.
- [x] Bench runs end-to-end; both protocols' verifiers accept on every iteration (helpers panic on rejection).
- [x] Compile-time soundness check: `const _: () = assert!(LOG_BLOWUP * FRI_NUM_QUERIES + FRI_QUERY_POW_BITS >= SECURITY_LEVEL)`.
- [x] Runtime soundness check: `WhirConfig::check_pow_bits()` asserted in the diagnostic block.


whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+//! # What the bench measures
+//!
+//! - Prover wall-clock (commit + open).
+//! - Verifier wall-clock.
+//! - Postcard-serialised proof size.
+//!
+//! # Substrate (matched on both sides)
+//!
+//! - Field            : KoalaBear, with quartic extension `EF`.
+//! - Merkle hash      : Poseidon1 (arithmetic, field-element digests).
```
Nashtare: the PR title says Poseidon1 / Poseidon2 / Blake3 but we only test against `Poseidon1` I believe?

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+        let merkle_hash = MerkleHash::new(perm24);
+        let merkle_compress = MerkleCompress::new(perm16.clone());
+        let val_mmcs = ValMmcs::new(merkle_hash, merkle_compress, 0);
+        let challenge_mmcs = ChallengeMmcs::new(val_mmcs.clone());
+        let challenger = Challenger::new(perm16);
+        (challenger, val_mmcs, challenge_mmcs)
+    }
+}
+
+/// Diagnostic table: per-phase timings and proof bytes for every (hash, m, protocol) triple.
```
Nashtare, line 792: cf comment above, it says for every hash but we're only testing 1

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+    println!();
+}
+
+// FRI soundness sanity is hash-agnostic and constant: enforce at compile time.
+const _: () = assert!(
+    LOG_BLOWUP * FRI_NUM_QUERIES + FRI_QUERY_POW_BITS >= SECURITY_LEVEL,
+    "FRI soundness budget below target"
+);
+
+/// Bench total proving time (commit + open) for both protocols across all hash backends.
```
Nashtare, line 922: same here for hashes mention

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+            }
+        }};
+    }
+
+    prove_block!(poseidon1);
+
+    group.finish();
+}
+
+/// Bench verification time for both protocols across all hash backends.
```
Nashtare, line 981: same here for hashes mention

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+/// ```text
+///     ζ = (z, z^2, z^4, …, z^{2^(m - log_width - 1)})
+///     ⇒  f_i(z)  =  p̃(i_binary, ζ)        for every i ∈ {0..256}
+/// ```
+fn whir_setup<MT, Ch>(
+    num_variables: usize,
+    log_width: usize,
+    mmcs: MT,
+    base_challenger: Ch,
+) -> WhirRig<MT, Ch>
```
Nashtare: If we aim at only supporting Poseidon1 for this bench (as is currently the case), I think this, along with `fri_setup` / `*prove_fully` / `*verify_fully` can be simplified

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+/// We keep all PoW on the query phase to match `new_benchmark` conventions.
+const FRI_COMMIT_POW_BITS: usize = 0;
+
+/// FRI per-query PoW bits.
+const FRI_QUERY_POW_BITS: usize = POW_BITS;
+
+/// Minimum FRI query count that reaches the target soundness under the ethSTARK conjecture.
+///
+/// Solves `log_blowup * queries + query_pow_bits >= security_level` for `queries`.
+const FRI_NUM_QUERIES: usize = SECURITY_LEVEL.div_ceil(LOG_BLOWUP) - FRI_QUERY_POW_BITS;
```
Nashtare: this is not correct (or oinly correct for log_blowup 1). It should be

```suggestion
/// Solves `log_blowup * queries + query_pow_bits >= security_level` for `queries`.
const FRI_NUM_QUERIES: usize = (SECURITY_LEVEL - FRI_QUERY_POW_BITS).div_ceil(LOG_BLOWUP);
```

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,973 @@
...
+/// Target soundness in bits for both protocols.
+///
+/// Interpreted under the capacity-regime conjecture (see security framing above).
+const SECURITY_LEVEL: usize = 100;
+
+/// Proof-of-work grinding bits.
+///
+/// Smallest budget that lets WHIR's summed per-round PoW reach the target soundness
+/// across the full message-size sweep. Shared by both protocols.
+const POW_BITS: usize = 22;
```
Nashtare: Do we really need to keep `PoW` at 22 bits? This is a LOT. For powerful machines like my M4 pro this may be acceptable still, i.e. not yielding too much noise across iterations that result in meaningful comparison, but I'd rather we find a way to tune it down (ideally at 18 or even 16). Is it not possible with the other WHIR params?

Nashtare (commented):
Thanks Thomas! I have a couple comments that I think are worth considering, but it's looking nice overall, love the comparison!

whir/benches/fri_vs_whir.rs
tcoratger: Fixed, I've just added the other hash functions as well to keep things super generic and tunable. The user can still comment out some of them.

whir/benches/fri_vs_whir.rs
tcoratger: Fixed

whir/benches/fri_vs_whir.rs
tcoratger: Fixed

whir/benches/fri_vs_whir.rs
tcoratger: Fixed

whir/benches/fri_vs_whir.rs
tcoratger: Fixed

whir/benches/fri_vs_whir.rs
tcoratger: Fixed

whir/benches/fri_vs_whir.rs
tcoratger: Fixed with a documentation to explain the choices

whir/benches/fri_vs_whir.rs
```diff
@@ -0,0 +1,1054 @@
...
+/// Proof-of-work grinding bits, shared by both protocols.
+///
+/// # Why this value
+///
+/// - The WHIR side needs every per-round grinding requirement to fit under this budget.
+/// - At `m = 22`, folding factor `4`, and the default per-round rate schedule, the last
+///   folding round fixes the floor at `20`.
+/// - `20` is the smallest budget that clears that floor.
+/// - Still four times less grinding than a `22`-bit budget, so the per-iteration grind stays
+///   cheap and grinding noise does not dominate the wall-clock comparison.
```
Nashtare: nit: noisy

```suggestion
/// - The WHIR side needs every per-round grinding requirement to fit under this budget. `20` is the smallest budget that clears that floor.
```

Nashtare (approved):

PR #1615, opened 2026-05-06
perf(uni-stark): reuse quotient constraint buffers

RekCuy63:
## Summary

Reuse the per-worker constraint buffers in `uni-stark` quotient evaluation and write quotient values directly into a preallocated output vector. This mirrors the allocation-reuse pattern already used by `batch-stark` without changing constraint evaluation order.

## Changes

- Replace the `flat_map_iter(...).collect()` quotient path with `for_each_init`.
- Reclaim `base_constraints` and `ext_constraints` after each packed chunk so each worker keeps its allocation.
- Write quotient outputs through disjoint mutable slices instead of collecting an intermediate iterator.

### DIFF
--- a/uni-stark/src/prover.rs
+++ b/uni-stark/src/prover.rs
@@ -9,6 +9,7 @@ use p3_commit::{Pcs, PolynomialSpace};
 use p3_field::{PackedFieldExtension, PackedValue, PrimeCharacteristicRing};
 use p3_matrix::Matrix;
 use p3_matrix::dense::{RowMajorMatrix, RowMajorMatrixView};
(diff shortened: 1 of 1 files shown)


Nashtare (commented):
Have you measured any concrete savings above noise threshold?

RekCuy63:
@Nashtare That is about a 28.6% median reduction, with no overlap between the measured ranges.

image

tcoratger:
@Nashtare I can't reproduce the speedup here locally, can you double confirm?

Nashtare:
Can't seem to reproduce either

PR #1618, opened 2026-05-08
perf(field): make halving helpers branchless

RekCuy63:
## Summary

Use mask arithmetic in the generic scalar halving helpers. This removes the parity branch while preserving the same field element result for both 32-bit and 64-bit prime-field helpers.

## Changes

- Make `halve_u32` and `halve_u64` compute the odd-element correction with a mask.
- Keep the existing `const fn` API and arithmetic formula unchanged.

## Test plan

- [x] `rustfmt +nightly --edition 2024 --check field/src/helpers.rs`
- [x] `cargo +stable test -p p3-field`
- [x] `cargo +stable test -p p3-mersenne-31`


Nashtare:
This is slower, check godbolt.

PR #1620, opened 2026-05-09
field: expand packed-extension API; fix unsound PackedValue impls

adr1anh:
## Why

The three packed-extension types — `PackedBinomial`, `PackedCubic`, `PackedQuintic` — each carry a near-duplicate `Div` / `DivAssign` impl that hand-rolls Montgomery's trick directly over the basis-major lane layout. Three copies, three subtly different shapes, and on BabyBear the access pattern is awkward enough that the inner loop suffers on the prefetcher.

While auditing those three impls with an eye to consolidating them, I noticed that `PackedCubic` and `PackedQuintic` (but not `PackedBinomial`) also claim `unsafe impl PackedValue` and `unsafe impl PackedField`, with `from_slice` / `as_slice` defined as raw pointer casts. The casts assert that `[ExtField; W]` and `&PackedExt` share a layout. They don't:

```
&[ExtField; W]   is lane-major:    [a0,b0,c0, a1,b1,c1, …, a_{W-1},b_{W-1},c_{W-1}]
&PackedExt       is basis-major:   [a0,…,a_{W-1}, b0,…,b_{W-1}, c0,…,c_{W-1}]
```

For `W = 1` they coincide; for any `W > 1` the cast permutes which scalars represent which (basis-coefficient, lane) pair. Latent UB. A workspace-wide search confirmed zero callers exercised the trait, so the bug never fired — but it was lying in wait for the first downstream user that wired `PackedField` into a generic call site.

This PR fixes the soundness bug by removing the broken impls, and folds the Div consolidation on top of it. To do that cleanly required filling in a couple of missing primitives in the `PackedFieldExtension` API; once those exist, the three Divs collapse to a single shared wrapper that does per-lane Montgomery against a stack-allocated `[EF; W]` buffer. The lane-materialize / sequential-invert / lane-rebuild pattern turns out to be more cache- and prefetch-friendly than the basis-major scattered indexing the prior in-place form needed, so Div gets a small perf bump as a side effect (numbers below).

A small adjacent cleanup rides along — `ExtensionField::from_ext_basis_coefficients` replaces two hand-rolled `ith_basis_element` lifting loops in the `uni-stark` and `batch-stark` verifiers, in the same spirit of "use the trait surface instead of reinventing it."

## What changes

**Soundness (BREAKING).** `unsafe impl PackedValue` and `unsafe impl PackedField` removed from `PackedCubicTrinomialExtensionField` and `PackedQuinticTrinomialExtensionField`. This brings them in line with `PackedBinomial`, which intentionally never claimed those traits. The `PackedFieldExtension` API already provides the sound bridge between scalar EF and packed-extension types; that's what callers should use.

**`Div` consolidation.** New `invert_packed_extension<F, EF>` in `field/src/batch_inverse.rs` — a shared per-lane Montgomery wrapper. Dispatches `F::Packing::WIDTH ∈ {1, 2, 4, 8, 16}` to a const-generic body that materializes lanes into `[EF; W]` via `PackedFieldExtension::extract`, runs `batch_multiplicative_inverse_general`, and rebuilds via `PackedFieldExtension::from_ext_fn`. Allocation-free; LLVM folds the dispatch to the single live arm post-monomorphization. All three packed extensions' `Div` / `DivAssign` now delegate. Also tightens those impls from `<F, PF: PackedField<Scalar=F>>` to `<F, F::Packing>` — no caller in-tree uses a non-default `PF`, and the wrapper requires `PackedFieldExtension` which is only impl'd on `F::Packing`.

**`PackedFieldExtension` API expansion.** Adds defaults filling out the symmetry gap with `PackedValue`: `from_ext_fn` (canonical primitive), `pack_ext_columns(_fn)`, `unpack_ext_into` / `unpack_ext_iter`, `to_ext_slice`. `from_ext_slice` changes from required to a default delegating to `from_ext_fn`. Existing impls override `from_ext_fn` for the layout-aware fast path; the slice form falls through. Purely additive.

**`ExtensionField::from_ext_basis_coefficients`.** New trait method `(&[Self]) -> Option<Self>`, reassembling a `Self` element from `D = DIMENSION` coefficients in `Self` via `Σⱼ basisⱼ · coeffsⱼ`. The Self-coefficient counterpart to the existing `from_basis_coefficients_slice` (which takes coefficients in `Base`). After this commit, the only `ith_basis_element` callers in the workspace live in the field crate itself.

**Bench harness.** `field-testing` gains `benchmark_div_latency` / `benchmark_div_throughput` (mirroring the existing add/mul/sub helpers) and a `bench_packed_extension_field!` macro that wires up the full add/mul/div × latency/throughput cross-product for `(label, ScalarEF)` pairs. `baby-bear/benches/packed_extension.rs` invokes it for `BinomialExtensionField<BabyBear, D>` at D ∈ {4, 5, 8}. Existed primarily to verify the consolidated Div isn't a regression; ended up showing it's actually a speedup.

## Perf — BabyBear binomial-extension `Div`

Apple M2 Pro, aarch64 NEON (`F::Packing::WIDTH = 4`), `RUSTFLAGS=-Ctarget-cpu=native`. Criterion baseline = bench-harness commit; HEAD has the consolidation applied. The intermediate API-expansion commit doesn't touch any Div code path, so the delta is attributable to the consolidation alone.

| Bench | Baseline | HEAD | Δ |
|---|---|---|---|
| D=4 div-throughput | 405 µs | 391 µs | **−3.3%** |
| D=4 div-latency    | 409 µs | 389 µs | **−5.5%** |
| D=5 div-throughput | 468 µs | 457 µs | **−2.5%** |
| D=5 div-latency    | 469 µs | 457 µs | **−2.2%** |
| D=8 div-throughput | 637 µs | 601 µs | **−5.5%** |
| D=8 div-latency    | 738 µs | 605 µs | **−12.9%** ⁽\*⁾ |

⁽\*⁾ D=8-latency baseline run had 14 outliers / 11 high-severe; criterion's 95% CI is wide (−17.2..−8.9%). Improvement is real, magnitude uncertain.

## Test plan

- [ ] `cargo +stable sort --workspace --grouped --check`
- [ ] `cargo +nightly fmt --all -- --check`
- [ ] `cargo +stable clippy --all-targets -- -D warnings`
- [ ] `cargo test -p p3-field -p p3-field-testing`
- [ ] `cargo test -p p3-baby-bear -p p3-koala-bear -p p3-mersenne-31 -p p3-goldilocks`
- [ ] `cargo test -p p3-uni-stark -p p3-batch-stark`
- [ ] `RUSTFLAGS=-Ctarget-cpu=native cargo bench -p p3-baby-bear --bench packed_extension -- div-` (numbers above)

### DIFF
--- a/baby-bear/Cargo.toml
+++ b/baby-bear/Cargo.toml
@@ -40,5 +40,9 @@ harness = false
 name = "extension"
 harness = false
 
+[[bench]]
+name = "packed_extension"
+harness = false
+
 [lints]
 workspace = true
--- a/baby-bear/benches/packed_extension.rs
+++ b/baby-bear/benches/packed_extension.rs
@@ -1,0 +1,10 @@
+use p3_baby_bear::BabyBear;
+use p3_field::extension::BinomialExtensionField;
+use p3_field_testing::bench_packed_extension_field;
+
+bench_packed_extension_field! {
+    BabyBear,
+    quartic = BinomialExtensionField<BabyBear, 4>,
+    quintic = BinomialExtensionField<BabyBear, 5>,
+    octic = BinomialExtensionField<BabyBear, 8>,
+}
--- a/batch-stark/src/verifier/mod.rs
+++ b/batch-stark/src/verifier/mod.rs
@@ -9,7 +9,7 @@ use hashbrown::HashMap;
 use p3_air::Air;
 use p3_air::symbolic::{AirLayout, SymbolicExpressionExt};
 use p3_commit::{Pcs, PolynomialSpace};
-use p3_field::{Algebra, BasedVectorSpace, PrimeCharacteristicRing};
+use p3_field::{Algebra, BasedVectorSpace, ExtensionField, PrimeCharacteristicRing};
 use p3_lookup::folder::VerifierConstraintFolderWithLookups;
 use p3_lookup::logup::LogUpGadget;
 use p3_lookup::{InteractionSymbolicBuilder, Kind, LookupProtocol};
@@ -540,20 +540,13 @@ where
             if aux_width == 0 {
                 return vec![];
             }
-            // Chunk the flattened coefficients into groups of size `dim`.
-            // Each chunk represents the coefficients of one extension field element.
+            // Each `ext_degree`-chunk holds the basis coefficients (in EF) of one EF element.
+            // chunks_exact yields chunks of exactly `ext_degree` = DIMENSION, so the unwrap
+            // below cannot panic.
             flat.chunks_exact(ext_degree)
-                .map(|coeffs| {
-                    // Dot product: sum(coeff_j * basis_j)
-                    coeffs
-                        .iter()
-                        .enumerate()
-                        .map(|(j, &coeff)| {
-                            coeff
-                                * Challenge::<SC>::ith_basis_element(j)
-                                    .expect("Basis element should exist")
-                        })
-                        .sum()
+                .map(|chunk| {
+                    Challenge::<SC>::from_ext_basis_coefficients(chunk)
+                        .expect("chunk length matches DIMENSION by construction")
                 })
                 .collect()
         };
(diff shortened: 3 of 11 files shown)


tcoratger (approved):
Thanks a lot for flagging this! :)

PR #1621, opened 2026-05-09
perf(dft): generalise coset LDE to support fused quotient commitment

adr1anh:
# perf(dft): generalise coset LDE to support fused quotient commitment

## Summary

Adds `coset_lde_batch_with_transform` to the `TwoAdicSubgroupDft` trait — a
generalisation of `coset_lde_batch` that invokes a closure on the
intermediate coefficient buffer between the iDFT and the forward DFT.
`coset_lde_batch` becomes a thin wrapper passing a no-op closure. Two
backend specialisations and one side optimisation come along.

The motivation is the chunk-by-chunk pattern in `Pcs::commit_quotient`,
where `D` independent coset LDEs can be fused into one by folding the
per-chunk shift `ω_J^{-tk}` into a single diagonal pass on the coefficient
buffer. **This PR ships only the DFT trait extension.** The FRI
integration that exercises it is included in the branch as an attempt
(commits 5–6 below) and reverted in commit 7, because cleanly wiring it
into `uni-stark` forced an opt-in `Pcs::quotient_commit_matrix_count`
trait method and a matrix-count dispatch on both prover and verifier —
interface churn that deserves its own discussion separate from the DFT
primitive. The reverted commits stay in branch history so the attempt is
reviewable and the LDE-only benchmark numbers are reproducible by
cherry-pick.

## Mathematical motivation

Given quotient `Q` of degree `< N·D` evaluated on coset `gJ` of size `N·D`
(`J = H × {ω_J^t}_{t=0..D-1}`), the standard chunked commit runs `D`
independent coset LDEs, each turning `q_t(g·ω_J^t·H)` into `q_t(g·K)`. A
metadata-only reshape from `(N·D, W)` to `(N, W·D)` puts column block `t`
on the H-coset `g·ω_J^t·H`. A plain (un-shifted) iDFT extracts shifted
coefficients

```
c_hat[k, t·W + c] = a_{t,c}[k] · (g · ω_J^t)^k
```

so multiplying column block `t`, row `k` by `(ω_J^{-k})^t` cancels the
per-chunk `ω_J^{tk}` factor and leaves `a_{t,c}[k] · g^k`. The trailing
plain forward DFT carries the implicit `g^k` and lands directly on `g·K`.
One iDFT setup, one diagonal pass, one DFT — replacing D iDFTs + a
`split_evals` copy.

### `coset_lde_batch_with_transform` (`dft/src/traits.rs`)

```rust
fn coset_lde_batch_with_transform<T>(
    &self, mat, added_bits, shift, transform: T,
) -> Self::Evaluations
where T: FnOnce(&mut RowMajorMatrixViewMut<'_, F>, Layout);
```

`Layout` is `Natural` / `BitReversed`, so the closure can translate
memory row index → natural-order coefficient index without forcing the
backend to materialise a particular layout. Default impl runs
`idft_batch + closure + coset_dft_batch`.

### DFT specialisations

| Backend | Specialisation |
|---|---|
| `Radix2DitParallel` | Closure runs on the bit-reversed coefficient buffer between the iDFT and the B parallel forward DFTs — zero extra reorder pass. |
| `Radix2Bowers` | Closure runs on bit-reversed memory after `bowers_g_t + 1/h`. The combined `1/h · shift^k` weight pass is split into two only on the closure path; the no-op `coset_lde_batch` override is preserved. |
| `Radix2Dit` / `NaiveDft` | None needed — trait default already optimal. |
| `Radix2DFTSmallBatch` / `RecursiveDft` | Deliberately not specialised — their fused pipelines have no clean intermediate state with row-major coefficients. Default handles them correctly. |

A proptest in `dft/tests/testing.rs` exercises a per-row scaling closure
across all five backends and both layouts.

### Side optimisation: parallelise `coset_shift_cols` (`dft/src/util.rs`)

Was running fully serially (non-`par` row iter zipped with lazy `Powers`,
scalar `*=` per element). Backs the trait defaults of
`coset_dft_batch` / `coset_idft_batch`, reachable from `Radix2Dit` /
`NaiveDft` LDE chains, `TwoAdicFriPcs::get_evaluations_on_domain`
fallback, and the new `coset_lde_batch_with_transform` default for
backends without a specialisation. New impl uses
`shift.powers().collect_n(h)` (parallel + packed via `BoundedPowers`) +
`par_rows_mut` + `scale_slice_in_place_single_core`.

## Benchmark: LDE pipeline only

The fused vs chunked LDE pipelines that `commit_quotient` would call
internally. MMCS commit excluded so the comparison isolates the DFT
work. 8 rayon threads, `--features parallel`, criterion medians over
10 samples with 12s measurement / 5s warm-up.

| Field | logN | D | B | chunked | fused | speedup |
|---|---|---|---|---|---|---|
| BabyBear   | 16 | 2 | 4 |  4.85 ms |  3.22 ms | **1.51×** |
| BabyBear   | 16 | 2 | 8 |  8.09 ms |  5.13 ms | **1.58×** |
| BabyBear   | 18 | 2 | 4 | 16.71 ms | 12.24 ms | **1.36×** |
| BabyBear   | 18 | 2 | 8 | 26.57 ms | 19.24 ms | **1.38×** |
| Goldilocks | 16 | 2 | 4 |  6.79 ms |  5.12 ms | **1.33×** |
| Goldilocks | 16 | 2 | 8 | 11.43 ms |  8.60 ms | **1.33×** |
| Goldilocks | 18 | 2 | 4 | 25.35 ms | 19.44 ms | **1.30×** |
| Goldilocks | 18 | 2 | 8 | 41.64 ms | 32.45 ms | **1.28×** |

1.3–1.6× on the LDE work, scaling with `DIM`. BabyBear (quartic ext)
benefits more than Goldilocks (quadratic ext) because the wider per-chunk
matrix makes the iDFT-setup overhead more amortizable.

Reproduce:

```
git checkout 755a8074
RAYON_NUM_THREADS=8 cargo bench -p p3-dft --bench fft --features parallel \
  -- 'coset_lde_(chunked|fused)' --measurement-time 12 --warm-up-time 5
```

## Commit map

1. `5fc5245d` — bench(dft, fri): add commit_quotient and chunked-LDE baselines (reverted)
2. `04e868a3` — feat(dft): add `coset_lde_batch_with_transform`
3. `6d8520ba` — feat(dft): specialize `Radix2Bowers` for `coset_lde_batch_with_transform`
4. `d3aa4a83` — perf(dft): parallelize `coset_shift_cols` with packed scaling
5. `e2311fb6` — perf(fri): fused `commit_quotient` (attempt, reverted)
6. `755a8074` — bench(dft): add fused-LDE counterpart to `coset_lde_chunked` (reverted)
7. `c67befcb` — chore: revert FRI `commit_quotient` attempt and motivational benches

Final diff against `main` = commits 2–4 (DFT-only).

### `perf(fri): fused commit_quotient` (commit 5)

`TwoAdicFriPcs::commit_quotient` override: reshape `(N·D, DIM)` to
`(N, D·DIM)`, run one `coset_lde_batch_with_transform` with the
shift-stripping closure (`strip_chunk_coset_shifts`), commit the
resulting fat `(N·B, D·DIM)` matrix as a **single** MMCS matrix.

Why reverted: the single-matrix MMCS shape diverges from the
chunk-by-chunk shape that `Pcs::commit_quotient` currently produces. The
ergonomic cost is a new `Pcs::quotient_commit_matrix_count` trait method
(default `num_chunks`, override `1` for `TwoAdicFriPcs`) plus a
matrix-count dispatch in `uni-stark`'s prover and verifier. Works, all
tests pass, but feels like the wrong place to wedge a fork in the trait
without also discussing how `HidingFriPcs` (per-chunk random codeword
stripping) and `CirclePcs` should migrate. That conversation is a
separate PR.

### Motivational benches (commits 1, 6)

`fri/benches/commit_quotient.rs` (full pipeline) and
`coset_lde_chunked` / `coset_lde_fused` in `dft/benches/fft.rs`
(LDE-only). Numbers captured above; cherry-pick the SHAs to re-run.

## Wire format note (relevant only if the FRI attempt is revisited)

The Merkle commitment (root) and FRI rounds would be byte-identical to
the chunk-by-chunk path: MMCS leaf hashes concatenate row bytes the same
way for a `1 × (D·W)` matrix as for `D × W` matrices of equal height,
and FRI keys off `mat.height()` and a single `Val::GENERATOR` shift, not
the per-matrix `domain` argument. What does change:
`BatchOpening::opened_values` for the quotient round reframes from
`D × Vec<F[W]>` to `1 × Vec<F[D·W]>` — same total bytes, different serde
framing. Proofs produced by the new prover would not be verifiable by
the old verifier and vice versa.

## Appendix: helper for migrating other PCSs to the single-matrix shape

If `HidingFriPcs` / `CirclePcs` ever adopt the single fat-matrix MMCS
shape (the contract the FRI attempt assumes), the generic helper is:

```rust
/// Merge `D` LDE matrices of equal shape `(H, W)` into one fat
/// `(H, W·D)` matrix where row `k` is the concatenation of row `k`
/// from each input.
pub fn merge_quotient_chunk_ldes<F: Copy + Default + Send + Sync>(
    ldes: Vec<RowMajorMatrix<F>>,
) -> RowMajorMatrix<F> {
    let d = ldes.len();
    let height = ldes[0].height();
    let chunk_w = ldes[0].width();
    let fat_w = chunk_w * d;
    let mut out = vec![F::default(); height * fat_w];
    out.par_chunks_exact_mut(fat_w)
        .enumerate()
        .for_each(|(k, fat_row)| {
            for (t, lde) in ldes.iter().enumerate() {
                let src = &lde.values[k * chunk_w..(k + 1) * chunk_w];
                fat_row[t * chunk_w..(t + 1) * chunk_w].copy_from_slice(src);
            }
        });
    RowMajorMatrix::new(out, fat_w)
}
```

One row-parallel copy pass. `TwoAdicFriPcs` skips it because its fused
LDE produces the fat matrix natively; needed only by PCSs that retain
the per-chunk LDE structure (`HidingFriPcs`, `CirclePcs`).

For `HidingFriPcs` specifically, the random-codeword tail-strip in
`open_with_preprocessing` is currently per-matrix; migrating to the
single-matrix shape would require reorganising the strip so all
`D · num_random_codewords` random columns live at the tail of the fat
matrix rather than interleaved per chunk.

## Compatibility

Fully backwards compatible. The DFT trait gains one method with a
working default; existing callers of `coset_lde_batch` are unchanged.
No PCS, prover, or verifier behaviour changes ship in this PR.

## Test plan

- [ ] `cargo test -p p3-dft` (incl. new proptest covering all five
      backends × both `Layout` variants)
- [ ] `cargo test --workspace`
- [ ] `cargo clippy --all-targets -- -D warnings`
- [ ] `cargo +nightly fmt --all -- --check`


Nashtare (approved):
Looking great, thank you!

PR #1623, opened 2026-05-10
perf(goldilocks): skip canonicalize-b in Poseidon2 NEON additions where b < P

Niek-Kamer:
## Summary

Introduces `add_canonical_asm`, a variant of the existing `add_asm` Goldilocks addition that drops the leading `subs/csel` "canonicalize-b" pair when the caller can statically guarantee `b < P`. Saves 2 ALU instructions per call.

Applied at three classes of call site inside the Poseidon2 NEON ASM where the canonicality of `b` is provable from local code shape:

1. **`apply_mat4_asm`** — 5 of 9 internal sums whose `b` is the output of `add_asm` or `double_asm` (always reduced). The 4 sites whose `b` is a public-input element `x[i]` keep `add_asm`, so the function's non-canonical-input contract (see `test_apply_mat4_asm_danger`) is preserved.
2. **`mds_light_permutation_asm`** — the sums accumulator (initialised to 0) and the final fold-back over post-mat4 state.
3. **`external_round_fused_dual_w8`** — all 16 round-constant additions. External and internal RC tables are mechanically asserted to be `< P` by `test_goldilocks_poseidon2_rc_tables_canonical`, which iterates every `GOLDILOCKS_POSEIDON2_RC_*` const and checks `c.value < P`.

## Why it's safe

`add_canonical_asm` is `pub(super)` inside `aarch64_neon` and unreachable from any other module (`utils` is not re-exported by `mod.rs`). Every call site in this PR has its canonicality argument annotated inline in the source. The existing `test_apply_mat4_asm_danger` proptest and the `test_mds_light_w{8,12,16}_danger` proptests, which exercise *non-canonical* inputs to the public-facing helpers, were intentionally kept passing by *not* converting the sites whose `b` is direct user input.

In debug builds, `add_canonical_asm` carries a `debug_assert!(b < P)` so any future caller that violates the precondition fails loudly under `cargo test` rather than silently corrupting downstream proofs.

## Benchmarks

Measured on **Raspberry Pi 5 / Cortex-A76** (core 3 via `taskset -c 3`, `governor=performance`, `perf_event_paranoid=1`). Not measured on Apple M-series or AWS Graviton, happy to add numbers if a maintainer can run them. The optimization itself is microarchitecture-independent (drops a fixed 2-instruction `subs/csel` pair per call), so the *direction* of the speedup should hold on any aarch64 target; the *magnitude* depends on the specific microarchitecture's issue width, integer-port pressure, and dependency-chain interactions, which I can't predict without measuring.

| Bench | Before | After | Δ |
|---|---|---|---|
| `prove_goldilocks_poseidon2` (e2e wall, 5×5 reps, median) | 124.18 s | 120.54 s | **−2.93 %** (ranges fully disjoint) |
| `external_terminal_permute_dual_w8` (cycles, median) | 3434 | 3231 | **−5.91 %**, −538 instr |
| `external_initial_permute_dual_w8` (cycles, median) | 3617 | 3392 | **−6.21 %**, −660 instr |
| `internal_permute_split_dual_w8` | flat | flat | untouched |

`apply_mat4_asm` and `mds_light_permutation_asm` are exercised transitively by the `external_*_permute_dual_w8` cycle benches above (they sit on the call path); the e2e `prove_goldilocks_poseidon2` number folds in all three.

## Test plan

- [x] aarch64 native (Raspberry Pi 5, Cortex-A76), debug-mode build:
  - `cargo test -p p3-goldilocks`: **328 passed**, 0 failed (includes new `test_add_canonical_asm`, `test_add_canonical_asm_edge_pairs`, and `test_goldilocks_poseidon2_rc_tables_canonical`)
  - `cargo test -p p3-poseidon2`: 2 passed, 0 failed
  - `cargo test -p p3-fri`: 48 passed, 0 failed
  - `cargo test -p p3-uni-stark`: 35 passed, 0 failed (2 ignored)
  - `add_canonical_asm`'s `debug_assert!(b < P)` did not fire on any call site under realistic Poseidon2 inputs — empirical confirmation that the canonicality precondition holds wherever we use the optimized variant.
- [x] x86_64 workspace (non-aarch64 path): `cargo test` and `cargo test --features parallel` both green.
- [x] `cargo +nightly fmt --all -- --check`: clean.
- [x] `cargo +stable clippy --all-targets -- -D warnings`: clean (cross-target verified for `aarch64-unknown-linux-gnu`).
- [ ] CI: macos-latest job will provide independent aarch64 verification on PR open.


goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -940,15 +948,23 @@ pub unsafe fn external_round_fused_dual_w8(
...
+        // `add_canonical_asm`. Use the variant that skips the leading
+        // subs/csel canonicalize-b pair.
+        let s0_a = add_canonical_asm(state0[0], rc[0]);
+        let s0_b = add_canonical_asm(state1[0], rc[0]);
+        let s1_a = add_canonical_asm(state0[1], rc[1]);
+        let s1_b = add_canonical_asm(state1[1], rc[1]);
+        let s2_a = add_canonical_asm(state0[2], rc[2]);
+        let s2_b = add_canonical_asm(state1[2], rc[2]);
+        let s3_a = add_canonical_asm(state0[3], rc[3]);
+        let s3_b = add_canonical_asm(state1[3], rc[3]);
```
Nashtare, line 959: I believe the same applies to `external_round_fused_w8`

Nashtare: same for `internal_permute_state_asm_w8`?

Nashtare (commented):
Looks good, but I think we can push it a bit further!

goldilocks/src/aarch64_neon/poseidon2_asm.rs
Niek-Kamer: Thanks @Nashtare — both addressed in 3357c428.

Re-benched post-change on Pi 5 (criterion, p<0.001):

- `internal_permute_split_dual_w8`: flat (4.179 kc → 4.179 kc)
- `external_initial_permute_dual_w8`: −6.28% (was −6.21%)
- `external_terminal_permute_dual_w8`: −5.67% (was −5.91%)

`cargo test -p p3-goldilocks`: 328 passed.

One side note for transparency: I first widened the conversion to `external_round_asm`/`external_round_dual_asm` and the `dual_w{8,12,16}` internal-permute siblings, but that regressed `internal_permute_split_dual_w8` by +1.79% / +264 instructions on Pi 5, likely register-allocation pressure shifting under the smaller `asm!` blocks. So this commit stays scoped to the two functions you named. Happy to revisit the wider scope as a separate PR once it's been measured on Graviton / M-series.

Nashtare (commented):
Some LLM-generated comments in your last commit seem superfluous.

Niek-Kamer:
Trimmed those down in 92a7513c, thanks.

Nashtare (approved):
LGTM, thanks!

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -2043,7 +2051,7 @@ mod tests {
     ) {
         let mut rng = SmallRng::seed_from_u64(42);
 
-        let internal_constants: Vec<u64> = (0..22).map(|_| rng.random()).collect();
+        let internal_constants: Vec<u64> = (0..22).map(|_| rng.random::<u64>() % P).collect();
```
Nashtare, line 2054: nit: while the change?

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1940,7 +1948,7 @@ mod tests {
         #[test]
         fn test_external_round_fused_w8(
             vals in prop::array::uniform8(any::<u64>()),
-            rc in prop::array::uniform8(any::<u64>()),
+            rc in prop::array::uniform8(0u64..P),
```
Nashtare, line 1951: nit: unrelated and unnecessary I believe

goldilocks/src/aarch64_neon/poseidon2_asm.rs
Nashtare: also propagated in other places

goldilocks/src/aarch64_neon/poseidon2_asm.rs
Nashtare: also propagated in other places

Niek-Kamer: Paired with the add_asm → add_canonical_asm conversions in this commit: the new function has debug_assert!(b < P) at utils.rs:182-186. Those tests are pre-existing (test_external_round_fused_w8, test_external_round_fused_dual_w8, test_specialized_matches_generic and its width callers); only the input bounds were tightened, otherwise they'd panic in debug. Inline comments at each site cite the precondition.

PR #1624, opened 2026-05-10
test(field): cover packed-extension Div with per-lane consistency check

tcoratger:
## Summary

@adr1anh Follow-up to #1620.

That PR removed an unsound `PackedValue`/`PackedField` impl on the cubic and quintic packed-extension types and consolidated their `Div` through a shared lane-by-lane wrapper. The bug was latent (no in-tree caller exercised it), but the broken per-lane divide *was* publicly reachable — it just had no regression coverage to catch it.

This PR adds the missing coverage:

- New generic helper `test_packed_extension_div_consistency<F, EF, PEF>` in `field-testing`.
- Wired into `test_packed_extension_field!`, so every existing packed-extension test site (binomial 2/4/5/8, cubic trinomial, quintic trinomial, Mersenne31 complex) now checks it automatically.

The invariant it enforces:

```text
(a / b).lane(i) == a.lane(i) / b.lane(i)   for every lane i
```

The check uses only the public `PackedFieldExtension` surface (`from_ext_slice`, `extract`) plus the `Div` operator, so it is layout-agnostic and survives any future refactor of the per-lane inverse path.

Manually verified by reverting #1620 locally: this test fails at cubic (Goldilocks, W=2) and quintic (KoalaBear, W=4); it passes for every binomial site in both branches.

## Test plan

- [ ] `cargo test -p p3-field-testing`
- [ ] `cargo test -p p3-goldilocks` (covers binomial 2, cubic trinomial, binomial 5)
- [ ] `cargo test -p p3-koala-bear` (covers binomial 4, binomial 8, quintic trinomial)
- [ ] `cargo test -p p3-baby-bear` (covers binomial extensions)
- [ ] `cargo test -p p3-mersenne-31` (covers complex extension)

### DIFF
--- a/field-testing/src/packedfield_testing.rs
+++ b/field-testing/src/packedfield_testing.rs
@@ -1,5 +1,6 @@
 use alloc::vec;
 use alloc::vec::Vec;
+use core::ops::Div;
 
 use p3_field::extension::{
     BinomialExtensionField, BinomiallyExtendable, PackedBinomialExtensionField,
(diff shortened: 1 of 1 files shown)


adr1anh (approved):
Thanks for adding this!

PR #1627, opened 2026-05-11
`fix(matrix): interpolate_arbitrary_point honors documented duplicate-domain contract`

Barnadrot:
### Summary

`RowMajorMatrix::interpolate_arbitrary_point` documents that it returns `None` if any two domain x-coordinates coincide. It currently returns `Some(row_i)` when the target happens to equal one of the duplicate domain values, because the equality-with-domain short-circuit runs **before** the duplicate check.

Reproducer (one of the tests added in this PR):

```rust
let xs = [F::ONE, F::TWO, F::ONE];                          // duplicate at index 0 and 2
let evals = RowMajorMatrix::new(vec![10, 20, 30].map(F::from_u32).to_vec(), 1);
let result = evals.interpolate_arbitrary_point(&xs, F::ONE); // ← target matches the duplicate
// Documented behavior: None (ill-posed problem)
// Current behavior:     Some(row 0) — returns 10
```

Fix: move the `barycentric_weights` invocation (which already detects duplicates) ahead of the equality-with-domain early exit. One-line reordering in `matrix/src/interpolation.rs`.

### Severity

**Medium.** The function has no current production callers inside Plonky3 — only test sites — so the bug is dormant in-tree. Downstream provers that build on top of `interpolate_arbitrary_point` would silently get an incorrect interpolation result on ill-formed input.

### Test plan

Two regression tests added in `matrix/tests/hunt4_interpolate_dup_early_exit.rs`:

- `duplicate_domain_with_matching_target_should_return_none` — the bug-triggering case. Fails on `main`, passes after the fix.
- `duplicate_domain_with_non_matching_target_returns_none` — the already-working path; included for contrast and to lock in behavior.

Both pass on `cargo test --test hunt4_interpolate_dup_early_exit` after the fix.

### Files touched

- `matrix/src/interpolation.rs` (+5, -3)
- `matrix/tests/hunt4_interpolate_dup_early_exit.rs` (+53, new)

---

<details>
<summary><b>Related findings — bh3-5 and bh3-6 already fixed by upstream #1620</b></summary>

In the same bug-hunting session that surfaced this `interpolate_arbitrary_point` issue, we independently identified two other bugs that have **already been fixed upstream** by `c092c1f4` (PR #1620, "field: expand packed-extension API; fix unsound PackedValue impls"):

- **bh3-5 (medium):** `PackedCubicTrinomialExtensionField` and `PackedQuinticTrinomialExtensionField` violated the `PackedValue` contract — SoA storage `[PF; D]` cannot be safely cast to `[Value; WIDTH] = [[F; D]; W]` (AoS), but the `unsafe impl PackedValue` did exactly that via raw pointer casts in `as_slice` / `from_slice`. #1620 removed those `PackedValue` impls entirely. Resolved.

- **bh3-6 (critical, dormant):** Derived from bh3-5 — `PackedCubic/PackedQuintic::Div` used `rhs.as_slice()[i].inverse()` lane-by-lane, which returned garbled triples instead of correct lane scalars. #1620 routes `Div` through a new `invert_packed_extension` helper that uses the safe `as_basis_coefficients_slice` extraction path. Resolved.

We flagged bh3-6 as **critical** when found because future provers adopting these types via the broken `PackedField` path would have silently generated wrong proofs. The upstream rewrite eliminates the trap entirely — exactly the right architectural fix. Worth a note here for visibility; nothing to merge.

Our local regression tests for these bugs are in `hunt6_packed_cubic_div.rs`, `hunt8_packed_quintic_div.rs`, `hunt9_packed_cubic_extract.rs` — they reference the pre-#1620 API and would need adaptation against the new packed-extension surface. Happy to contribute adapted versions in a follow-up if you'd like the explicit Div-via-`as_basis_coefficients_slice` regression coverage.

</details>

<details>
<summary><b>What else was tried — hypothesis tests that found no bug (priority-ranked, available as a separate follow-up PR if you want them)</b></summary>

These are tests we **wrote and ran** as part of the bug-hunting sweep. **None found a bug** — they're listed here as candidate test-suite additions to review and potentially pick candidates to add in a follow-up PR. Each is a stand-alone integration test; runtime cost in parentheses. **Not included in this PR** to avoid bloating the test suite without sign-off. If any look valuable, happy to send them as a separate "tests-only" PR.

Priority ordering: leverage × hot-path coverage × subtlety of the verified property.

| Priority | Test (file) | What it verifies | ~Runtime | Why this might be worth keeping |
|---|---|---|---:|---|
| **High** | `field/tests/hunt3_packed_ext_dotproduct.rs` (bh3-1) | `PackedBinomialExtensionField::mixed_dot_product` lane-mixed boundary inputs (0/1/-1/random per lane, N=8) match per-lane scalar `BinomialExtensionField::mixed_dot_product` bit-for-bit. | < 1 s | `mixed_dot_product` is on the hot path of WHIR sumcheck / commitment opening. Lane-mixed boundary inputs are the most likely shape for a SoA-vs-AoS regression on a future packed-extension refactor. |
| **High** | `whir/tests/hunt_sumcheck_packed_vs_scalar.rs` (bh3-2) | `sumcheck_coefficients_prefix` with `K=8` chunked packed path agrees with scalar fold across all tail sizes `(0,1,3,5,7)` and across the par-threshold boundary. 5 cases, all assert `h0` and `hinf` lane-summed packed == scalar. | < 2 s | WHIR sumcheck packed/scalar consistency is exactly the kind of "chunked vs whole" boundary where off-by-one / tail-handling bugs hide. Currently only verified via end-to-end WHIR tests; a focused unit test is cheaper to debug from. |
| **Medium** | `fri/tests/hunt7_fold_matrix_chain.rs` (bh3-7) | FRI `fold_matrix(log_arity=k)` equals the chain of `k` arity-2 folds with squared betas. Verified for `log_arity ∈ {2, 3}` (arity 4, 8) on small matrices. | < 1 s | The in-place `halve_inv_powers` update inside `fold_matrix` is subtle and not unit-tested in isolation; mainly exercised by end-to-end FRI tests. Worth keeping if/when higher-arity FRI configurations get more use. |
| **Low (deferred)** | (none committed — depends on bh3-5 / #1620 follow-up) | Packed-extension `PackedField::extract` and arithmetic correctness on the **post-#1620** API surface. | < 1 s each | Once the new `invert_packed_extension` helper and the `as_basis_coefficients_slice` extraction path settle, regression tests over those entry points are cheap insurance. We have older versions of these tests targeting the pre-#1620 API — happy to port. |

Items we considered but did **not** keep tests for:

- Several `monty_31` arithmetic bounds (`to_monty_signed` with `i32::MIN`, `div_2exp_u64` at `exp=32`, `mul_2exp_u64` lookup-table coverage, AVX-512 `dot_product_k` boundary inputs at all `P-1`) — verified analytically; existing `boundary_sweep` harness in `monty-31` covers the SIMD cases. Bug count: 0.
- `RecursiveDft` cache concurrency under interleaved thread updates with different `fft_len`s — traced through `extend_from` offsets and `TWO_ADIC_GENERATORS` invariants. Bug count: 0.
- `mm512_mod_add/sub` semantics on Poseidon2 `add_sum` non-canonical inputs from `mul_neg_2exp_neg` — analytically verified bounds. Bug count: 0.
- Binomial quartic/quintic/octic `mul` vs direct polynomial multiplication mod `(X^D - W)` for `D ∈ {4, 5, 8}` — coefficient-by-coefficient trace. Bug count: 0.
- AVX-512 `xor`/`andn` argument-order and signed-overflow semantics — traced. Bug count: 0.

If any of these would still be valuable as long-lived tests (especially the bound-checking ones), we can write them — would prefer your direction on what's worth the test-suite weight.

</details>


tcoratger (approved):
Thanks a lot!

Indeed this looks like a bug. I've just migrated the test in the appropriate file to avoid having another separate file for this.

@Nashtare Can you double check but it looks valid (I've checked locally and before the behaviour was not the right one in those tests).

Nashtare (approved):

PR #1628, opened 2026-05-11
feat(lookup): single-terminal LogUp aux trace

tcoratger:
@adr1anh I checked the PR https://github.com/0xMiden/miden-vm/pull/2962 you shared earlier this week and I've tried to extract some incremental ideas. This one is the first one and seems to bring some small improvements and simplicity on large AIRs.

## Summary

Collapse the per-lookup running-sum auxiliary columns into a single shared accumulator. Each AIR with lookups now carries:

- **1 accumulator column** (index `0`) — exclusive prefix sum of per-row totals.
- **N fraction columns** (indices `1..=N`) — each pinned to its lookup's `V/U` value.
- **1 committed terminal scalar per AIR**, regardless of how many lookups it declares.

Equivalent to Miden's \"All-LogUp Auxiliary Trace\" trick: collapsing N committed terminals into one. The math is unchanged, the constraint count is roughly the same, but the proof transcript shrinks and the trace generator runs one prefix sum instead of N.

## Trace layout

```text
before:                                  after:
[s_0 | s_1 | … | s_{N-1}]                [acc | frac_0 | frac_1 | … | frac_{N-1}]
 ↑                                         ↑       ↑
 one running sum per lookup,               per-row prefix sum of (Σ_c frac_c)
 N committed terminals                     1 committed terminal
```

## Constraint set

- **Fraction pin** (per lookup `c`): `U_c(r) * frac_c[r] - V_c(r) = 0` on every row.
- **Accumulator anchor**: `acc[0] = 0` on the first row.
- **Accumulator transition**: `acc[r+1] - acc[r] - Σ_c frac_c[r] = 0` on transition rows.
- **Terminal binding**: `terminal - acc[n-1] - Σ_c frac_c[n-1] = 0` on the last row.

## Cross-AIR verification

The verifier sums committed terminals across all AIRs and rejects when the total is non-zero. Per-bus attribution in the verifier error is dropped (debug-time \`check_lookups\` still gives bus-level diagnostics). Soundness: any per-bus imbalance survives the collapse with overwhelming probability because the bus challenges are sampled after the main commitment.

## API delta

- \`Lookup::column\` — semantics now \"slot index\"; maps to challenge pair \`(2*column, 2*column+1)\` and to permutation-trace column \`column + 1\`.
- \`LookupData<F> { name, aux_column, cumulative_sum }\` → \`LookupTerminal<F>(F)\` newtype.
- \`LookupProtocol::{eval_local, eval_global}\` → \`{eval_fraction, eval_accumulator}\`.
- \`LookupProtocol::verify_global_sum\` → \`verify_terminal_sum(&[Option<LookupTerminal<EF>>])\`.
- \`BatchProof::global_lookup_data: Vec<Vec<LookupData>>\` → \`lookup_terminals: Vec<Option<LookupTerminal>>\`.
- \`InvalidProofShapeError::{GlobalLookupDataCountMismatch, GlobalLookupDataMetadataMismatch}\` → \`LookupTerminalPresenceMismatch { air, expected_present, got_present }\`.
- \`LookupError\` now uses \`thiserror\` (\`Display\` impl, real message string).
- Dead \`LookupGadget\` trait in \`lookup/src/traits.rs\` removed.

\`Kind::Local\`/\`Kind::Global(String)\` are preserved — they still drive per-bus challenge sharing in \`transcript.rs::sample_perm_challenges\`.

## Benchmarks

\`cargo bench -p p3-lookup --bench logup\`, M-series macOS, criterion 0.8, BabyBear over \`BinomialExtensionField<_, 4>\`.

| Workload | main | PR | Δ (median) |
|---|---|---|---|
| **small** — h=2^10, 1 lookup, tuple=1 | 69.70 µs | 69.64 µs | −0.08% (noise) |
| **medium** — h=2^16, 2 lookups, tuple=2 | 10.70 ms | 10.46 ms | **−2.24%** |
| **large** — h=2^20, 4 lookups, tuple=1 | 280.04 ms | 266.86 ms | **−4.71%** |

Win sources:
- One parallel prefix-sum pass over per-row totals replaces N independent passes (one per lookup).
- The intermediate \`row_sums\` buffer of size \`height × num_lookups\` extension elements is gone — fractions are written directly into the aux trace during the batch-invert chunk. At \`h=2^20\`, N=4 lookups, that's ≈64 MiB of allocation/touching avoided per call.

No regressions across any workload.

## Inspiration

Single-terminal idea ported from 0xMiden/miden-vm#2962 (\"All-LogUp Auxiliary Trace\"). That PR also introduces a closure-based builder hierarchy (column → group → batch), mutex-aware groups, a cached-encoding escape hatch, a boundary builder, typed messages with bus prefixes, debug walkers, and a prover boolean fast-path — those are deliberately left out here and scoped as separate follow-up PRs.

## Test plan

- [x] \`cargo test -p p3-lookup\` — 21 passing (lib + bus tests).
- [x] \`cargo test -p p3-batch-stark\` — 38 passing including ZK + fixture round-trip.
- [x] Regenerated postcard fixtures (\`cargo test -p p3-batch-stark --test simple -- --ignored\`).
- [x] \`cargo clippy --workspace --all-targets\` clean.
- [x] \`cargo fmt --all -- --check\` clean.
- [x] Benchmarks above.


uni-stark/src/error.rs
```diff
@@ -106,24 +106,17 @@ pub enum InvalidProofShapeError {
...
+    /// Lookup terminal presence does not match the AIR's declared lookups.
+    ///
+    /// Each AIR commits exactly one terminal when it declares any lookup,
+    /// and none otherwise.
     #[error(
-        "air {air}: global lookup data metadata mismatch at index {lookup}: expected name={expected_name}, aux_column={expected_aux_column}; got name={got_name}, aux_column={got_aux_column}"
+        "air {air}: lookup terminal presence mismatch: expected_present={expected_present}, got_present={got_present}"
     )]
-    GlobalLookupDataMetadataMismatch {
+    LookupTerminalPresenceMismatch {
```
adr1anh: Nit: maybe lookup errors should not live here.

lookup/src/types.rs
Nashtare: This was enforced by the old design but is not the case anymore I believe? 
i.e. the challenges drawn are unique to the `Local` lookup, so its contributions to the AIR's terminal cannot algebraically cancel against any other lookup in the batch, so the single cross-AIR terminal sumcheck forces it to balance on its own.

lookup/src/types.rs
```diff
@@ -103,20 +108,29 @@ impl<F: Field> AsRef<[Lookup<F>]> for Lookups<F> {
     }
 }
 
-/// Prover-provided data for one global lookup interaction.
-#[derive(Debug, Default, Clone, Serialize, Deserialize)]
-pub struct LookupData<F> {
```
Nashtare, line 108: I think there's still `LookupData` lying around in a comment but it's been deleted.

batch-stark/tests/simple.rs
```diff
@@ -2168,186 +2172,24 @@ fn test_batch_stark_rejects_truncated_global_lookup_data() {
...
-    let traces = [&mul_trace, &fib_trace_1, &fib_trace_2];
-    let pvs = vec![vec![], fib_pis.clone(), fib_pis];
-
-    let instances = StarkInstance::new_multiple(&airs, &traces, &pvs);
-    let proof = prove_batch(&config, &instances, &prover_data);
-    (config, airs, proof, pvs, prover_data.common)
-}
-
-#[test]
-fn test_batch_stark_rejects_tampered_global_lookup_metadata() {
```
Nashtare, line 2257: minor: you removed two tests checking tampering lookup data but I think we may still want to check this (differently with the new API)?

I'm worried for instance of a case where a malicious proof would be inserting ` Some(_)` for an AIR that does not declare any lookups (i.e. that should carry `None` as its terminal), but for which the malicious data passed along with the proof may impact the cross-AIR sum check?

lookup/src/logup.rs
```diff
@@ -395,7 +395,7 @@ impl LookupProtocol for LogUpGadget {
             for ctx in lookups {
                 let a = ctx.column;
                 if !seen.insert(a) {
-                    panic!("duplicate aux column index {a} across lookups");
+                    panic!("duplicate lookup slot index {a} across lookups");
                 }
             }
```
Nashtare: I think we also may want to add (possibly in release too, not just debug assertions), the following:

```rust
        for (i, lookup) in lookups.iter().enumerate() {
            assert_eq!(
                lookup.column, i,
                "lookup slot index must match slice position: \
                 lookups[{i}].column = {} (expected {i})",
                lookup.column,
            );
        }
```

otherwise we could have invalid data causing the prover to UB with out-of-bounds (i.e. if we pass `[Lookup{column: 0, ..}, Lookup{column: 5, ..}]` for instance, this will write to aux column 2 but the constraint reads aux column 6).

Nashtare: Alternatively (though it applies in other places as well), we could stop taking `lookups: &[Lookup<Val<SC>>],` and take instead `lookups: &Lookups<Val<SC>>,` with constructors enforcing the contiguity above

lookup/src/types.rs
```diff
@@ -103,20 +108,29 @@ impl<F: Field> AsRef<[Lookup<F>]> for Lookups<F> {
...
+#[derive(Debug, Error)]
 pub enum LookupError {
-    /// Global cumulative sums do not balance to zero.
-    GlobalCumulativeMismatch(Option<String>),
+    /// Cross-AIR sum of committed lookup terminals is non-zero.
+    ///
+    /// Any unbalanced lookup in the batch produces a non-zero terminal sum
+    /// with overwhelming probability over the random permutation challenges.
+    #[error("cross-AIR lookup terminal sum is non-zero")]
+    TerminalSumNonZero,
```
Nashtare, line 143: One concern I have with this (apart from the fact that a test still refers to the now deleted `GlobalCumulativeMismatch`), is that we're loosing debuggability.

Before we could see the failing bus, now we just know there's an inconsistency, but from experience debugging lookup failures is painful, and this regression concerns me a little.

Nashtare: perhaps we could keep the per-bus individual sums behind debug assertions?

Nashtare (commented):
Not a full review yet, and this may deserve another set of eyes (cc @adr1anh) but dropping a couple comments

lookup/src/types.rs
tcoratger: Fixed

lookup/src/types.rs
tcoratger: Fixed

batch-stark/tests/simple.rs
tcoratger: Yes just added two new tests: `rejects_spurious_lookup_terminal` and `rejects_tampered_lookup_terminal_value` for this

lookup/src/logup.rs
tcoratger: - I've added the assertion thing
- For now I haven't done the second point you, maybe we can keep it for a followup PR to avoid making this one even bigger :)

lookup/src/types.rs
tcoratger: I've fixed the documentation of this error here to explain things.

Nashtare:
@tcoratger Could you pull-in latest `main` too? 🙂

tcoratger:
> @tcoratger Could you pull-in latest `main` too? 🙂

Done :)

Nashtare (approved):
Overall LGTM, thanks a lot!
@adr1anh would you like to take some time to look at it too before merging?

adr1anh (approved):
Looks good!

uni-stark/src/error.rs
tcoratger: Fixed :)

PR #1629, opened 2026-05-11
symmetric: assert Pad10 sponge const parameters

RekCuy63:
## Summary

- Add compile-time assertions for the `Pad10Sponge` rate, width, and output bounds.
- Apply the same const-parameter checks to `MultiField32Pad10Sponge`.
- Keep the assertion style aligned with the existing sponge checks added in `#1525`.

### DIFF
--- a/symmetric/src/sponge.rs
+++ b/symmetric/src/sponge.rs
@@ -322,6 +322,11 @@ where
     where
         I: IntoIterator<Item = T>,
     {
+        const {
+            assert!(RATE > 0);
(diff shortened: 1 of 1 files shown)


Nashtare (changes requested):
This should rather be done within the `new` constructor

RekCuy63:
@Nashtare  Done, thanks. I moved the Pad10 sponge const assertions into the constructors.

Nashtare (approved):
Thanks!

PR #1626, opened 2026-05-11
test(whir): verifier error coverage

XC0R:
## Summary

Two areas of the WHIR crate with zero test coverage:

**SumcheckError variants** — the existing test suite exercises only the happy path (`run_multi_table_sumcheck_test` unwraps every verifier result). One focused test per error variant so regressions that swap variants or turn errors into successes are caught:

| Test | Trigger site | What it verifies |
|---|---|---|
| `test_missing_sumcheck_data` | `data.rs:151-153` | `None` final_sumcheck with rounds > 0 |
| `test_round_count_mismatch` | `data.rs:155-159` | polynomial_evaluations.len() != expected rounds |
| `test_invalid_pow_witness` | `data.rs:114-116` | Zeroed PoW witness fails check_witness |
| `test_zero_rounds_returns_empty_point` | `data.rs:147-149` | rounds == 0 short-circuit for both None and Some |

**BCSS25 soundness validation** — the implementation uses Theorem 1.5 from [BCSS25] (eprint 2025/2055) but no test verifies the improvement over [BCI+20] or validates the 128-bit security claim:

| Test | What it verifies |
|---|---|
| `test_bcss25_improves_over_bci20` | BCSS25 strictly beats BCI+20 across log_degree ∈ {10..25}, log_inv_rate ∈ {1..4} |
| `test_128bit_security_with_koalabear_quintic` | 128-bit proximity gap security for degree-5 KoalaBear (field_size=155) |
| `test_full_security_budget_jb` | Full error budget: prox gaps + sumcheck + OOD + queries + PoW |
| `test_ud_log_eta_is_infinity` | Confirms UniqueDecoding log_eta fix |

**UniqueDecoding log_eta fix** — `log_eta` returned `0.0` for UniqueDecoding, implying eta = 1 (impossibly large). Changed to `f64::INFINITY` since eta is undefined in UD. All existing UD code paths guard with `if matches!(self, UniqueDecoding)` before using eta, so this is safe.

**Note:** `FiatShamirError::InvalidGrindingWitness` is declared but never constructed anywhere in the crate. Deferring removal since #1616 touches the same file (`pcs/verifier/errors.rs`).

## Files changed

| File | Change |
|---|---|
| `whir/src/sumcheck/tests.rs` | 4 new tests for SumcheckError variants |
| `whir/src/parameters/soundness.rs` | UD log_eta fix + 4 new BCSS25/security tests |

Closes #1625.

PR #1630, opened 2026-05-11
fix(whir): mark UniqueDecoding log_eta as undefined

wstran:
## Summary

`SecurityAssumption::log_eta(UniqueDecoding)` returns `0.0`, i.e. `eta = 2^0 = 1`. That value is meaningless for UD — the UD proximity distance is `(1 - rho) / 2`, with no eta term at all — and was flagged with two `TODO`s in the source. All current callers special-case the UD branch before reaching any eta-dependent arithmetic, so the wrong value is dead today. But a future refactor reading the eta path in the UD branch would silently produce a wrong delta instead of failing visibly.

This PR returns `f64::INFINITY` for the UD case, so any accidental future read propagates as NaN / overflow rather than as a plausible-looking number, and drops the two stale `TODO`s.

This follow-up was described in the body of #1626:

> "log_eta returned 0.0 for UniqueDecoding, implying eta = 1 (impossibly large). Changed to `f64::INFINITY` since eta is undefined in UD."

but did not actually land in that commit (the PR shipped only the SumcheckError tests and the `FiatShamirError` removal).

## Diff

```rust
// whir/src/parameters/soundness.rs
- // TODO: Maybe it makes more sense to be multiplicative. I think this can be set in a better way.
  pub const fn log_eta(&self, log_inv_rate: usize) -> f64 {
      match self {
-         // We don't use eta in UD
-         Self::UniqueDecoding => 0., // TODO: Maybe just panic and avoid calling it in UD?
+         // eta is undefined in UD: delta is (1 - rho) / 2, with no eta term.
+         // Return INFINITY so any accidental future caller that reads this in
+         // the UD branch fails visibly (NaN / overflow propagation) instead
+         // of silently treating eta as 2^0 = 1, which is "impossibly large"
+         // and would produce a wrong delta. All current callers special-case
+         // UD before reaching the eta-dependent arithmetic.
+         Self::UniqueDecoding => f64::INFINITY,
          ...
      }
  }
```

Plus one unit test (`log_eta_undefined_in_unique_decoding`) that pins the new contract across `log_inv_rate ∈ {1, 2, 5, 10, 20}`.

## Behaviour

No observable change for any existing caller of `log_eta`, `list_size_bits`, `prox_gaps_error`, `log_1_delta`, or downstream parameter selection — they all branch on `SecurityAssumption::UniqueDecoding` before the eta-dependent path.

## Test plan

- [x] `cargo test -p p3-whir --lib` — 201 passed (200 existing + 1 new), 0 failed.
- [x] `cargo build -p p3-whir` and `--examples` — clean.
- [x] `cargo clippy -p p3-whir --all-targets -- -D warnings` — clean.
- [x] `cargo sort --workspace --grouped --check` — clean.
- [x] `cargo build --target thumbv7em-none-eabi -p p3-whir` (no_std) — clean.

### DIFF
--- a/whir/src/parameters/soundness.rs
+++ b/whir/src/parameters/soundness.rs
@@ -54,12 +54,16 @@ impl SecurityAssumption {
     /// E.g. in JB proximity gaps holds for every delta in (0, 1 - sqrt(rho)).
     /// eta is the distance between the chosen proximity parameter and the bound.
     /// I.e. in JB delta = 1 - sqrt(rho) - eta and in CB delta = 1 - rho - eta.
-    // TODO: Maybe it makes more sense to be multiplicative. I think this can be set in a better way.
     #[must_use]
     pub const fn log_eta(&self, log_inv_rate: usize) -> f64 {
         match self {
-            // We don't use eta in UD
-            Self::UniqueDecoding => 0., // TODO: Maybe just panic and avoid calling it in UD?
+            // eta is undefined in UD: delta is (1 - rho) / 2, with no eta term.
+            // Return INFINITY so any accidental future caller that reads this in
+            // the UD branch fails visibly (NaN / overflow propagation) instead
+            // of silently treating eta as 2^0 = 1, which is "impossibly large"
+            // and would produce a wrong delta. All current callers special-case
+            // UD before reaching the eta-dependent arithmetic.
+            Self::UniqueDecoding => f64::INFINITY,
             // Set as sqrt(rho)/20
             Self::JohnsonBound => -(0.5 * log_inv_rate as f64 + LOG2_10 + 1.),
             // Set as rho/20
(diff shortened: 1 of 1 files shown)


whir/src/parameters/soundness.rs
```diff
@@ -54,12 +54,16 @@ impl SecurityAssumption {
...
         match self {
-            // We don't use eta in UD
-            Self::UniqueDecoding => 0., // TODO: Maybe just panic and avoid calling it in UD?
+            // eta is undefined in UD: delta is (1 - rho) / 2, with no eta term.
+            // Return INFINITY so any accidental future caller that reads this in
+            // the UD branch fails visibly (NaN / overflow propagation) instead
+            // of silently treating eta as 2^0 = 1, which is "impossibly large"
+            // and would produce a wrong delta. All current callers special-case
+            // UD before reaching the eta-dependent arithmetic.
+            Self::UniqueDecoding => f64::INFINITY,
```
tcoratger: I think this is better to simply panic here to warn the user directly. Can we have a panic and a test to confirm this really panics?

wstran:
Switched to `panic!` with a matching `#[should_panic]` test in 2dffc515. Thanks!

tcoratger:
@wstran CI is failing now because I suspect there is a call like this in the end to end test. 

This should not be the case and that needs probably to be fixed in the meantime. 

Can you check?

wstran:
CI failure was on me — `log_eta` was evaluated eagerly above the match in three callers, so the panic fired through UD even though the UD arms never read the value. Moved the call inside the JB / CB branches in 3d0e5deb. Full `cargo test -p p3-whir --lib` (201/201) green locally.

tcoratger (approved):

PR #1631, opened 2026-05-11
test(whir): BCSS25 proximity gap validation + UniqueDecoding log_eta fix

XC0R:
Split from #1626 per reviewer request; soundness tests kept separate from verifier error coverage.

## BCSS25 vs BCI+20 proximity gap comparison

4 tests validating the BCSS25 (eprint 2025/2055, Theorem 1.5) improvement over BCI+20 (eprint 2020/654, Theorem 1.2):

- **BCSS25 strictly improves over BCI+20** across `log_degree ∈ {10..25}`, `log_inv_rate ∈ {1..4}`. Improvement is approximately `log₂(n)` bits — one factor of `n` eliminated from the exceptional-z count.
- **128-bit security with degree-5 KoalaBear** (155-bit field) confirmed under JohnsonBound across realistic WHIR parameter ranges.
- **Full error budget**: proximity gaps + sumcheck + OOD + query + combination all contribute adequate security; PoW grinding stays under 30 bits.

## UniqueDecoding `log_eta` fix

`log_eta` for `UniqueDecoding` returned `0.0`, which would give `2^0 = 1` (incorrect eta = 1) if ever used in arithmetic. Changed to `f64::INFINITY` — eta is undefined in UD, and infinity correctly produces zero error contribution in any formula involving `1/eta`. All UD callers (`list_size_bits`, `prox_gaps_error`, `log_1_delta`) short-circuit before using the value, so the change is safe.

### DIFF
--- a/whir/src/parameters/soundness.rs
+++ b/whir/src/parameters/soundness.rs
@@ -573,131 +573,322 @@ mod tests {
         let _ = SecurityAssumption::UniqueDecoding.log_eta(5);
     }
 
-    // -------------------------------------------------------------------
-    // BCSS25 proximity gap improvement over BCI+20
+    // BCSS25 vs BCI+20: Johnson-bound proximity gap improvement.
     //
-    // [BCI+20] "Proximity Gaps for Reed-Solomon Codes" (FOCS 2020, eprint 2020/654)
-    //   Theorem 1.2: Johnson bound error ε_J = (k+1)² / (2·min(η,√ρ/(2δ)))⁷ · q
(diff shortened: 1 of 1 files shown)


tcoratger:
Thanks a lot, could you rebase now that https://github.com/Plonky3/Plonky3/pull/1630 is merged?

tcoratger (approved):
I think that the formulas were previously incorrect with especially a factor 5.5 that was supposed to be 3.5 in reality. I've fixed that and the doc as well.

PR #1633, opened 2026-05-11
chore: use T::zero_vec(n) instead of vec![T::ZERO; n]

tcoratger:
## Summary

Replace every `vec![T::ZERO; n]` call site with `T::zero_vec(n)`.

`PrimeCharacteristicRing::zero_vec` exists precisely so that specialized impls (e.g. `MontyField31`, `Goldilocks`) can transmute `vec![0u32; n]` / `vec![0u64; n]` and **skip the redundant userspace zeroing** that `vec![Self::ZERO; n]` causes on freshly mapped pages. Quoting the trait docstring (`field/src/field.rs:377-388`):

> Many operating systems zero pages before assigning them to a userspace process. In that case, our process should not need to write zeros, which would be redundant. However, the compiler may not always recognize this. In particular, `vec![Self::ZERO; len]` appears to result in redundant userspace zeroing.

## Scope

- **80 sites converted** across 27 files (production verifiers, prover hot paths, DFT, transpose, multilinear utils, WHIR sumcheck, tests, and benches).
- The trait's **default impl** in `field/src/field.rs:387` is intentionally kept as `vec![Self::ZERO; len]` — that's the fallback that specialization overrides.
- Two doc-comments in `air/src/check_constraints.rs` and `batch-stark/src/check_constraints.rs` updated to reference the new pattern.
- 8 now-unused `use alloc::vec;` imports dropped; one moved into `fri/src/verifier.rs`'s test module (whose tests still use the `vec!` macro).

## Test plan

- [x] `cargo check --workspace --all-targets` clean
- [x] `cargo fmt --all` applied
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] No remaining `vec![*::ZERO; ...]` outside the trait default impl

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/air/src/check_constraints.rs
+++ b/air/src/check_constraints.rs
@@ -906,7 +906,7 @@ mod tests {
     #[test]
     fn test_check_all_constraints_no_failures() {
         let air = AllZeroAir::<2>;
-        let values = vec![BabyBear::ZERO; 4]; // 2 rows × 2 cols, all zero
+        let values = BabyBear::zero_vec(4); // 2 rows × 2 cols, all zero
         let main = RowMajorMatrix::new(values, 2);
         let report = check_all_constraints(&air, &main, &[], None);
         assert!(report.is_ok());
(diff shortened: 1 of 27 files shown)


Nashtare (approved):
Thanks!

Issue #1585, opened 2026-04-28
[HVZK-WHIR 2/6] Private zero-evader for OOD samples


tcoratger:
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 9.2](https://eprint.iacr.org/2026/391.pdf).

## What

WHIR's out-of-domain (OOD) sample `y = ze(ρ) · f` is a deterministic linear function of the witness, which leaks information about it. The paper's fix (Lemma 9.3): pad the linear map with a uniformly random matrix `M ∈ F^{t×r}` whose columns span `F^t`, and concatenate fresh randomness to the message. The zero-evader property is preserved with error `max(ε_zero, 1/(|F|^t − 1))`, and privacy becomes perfect (`ζ_ze = 0`).

## API sketch

```rust
pub trait ZeroEvader<F: Field> {
    /// Returns a `t × ℓ` matrix.
    fn eval<R: Rng>(&self, rng: &mut R) -> Vec<Vec<F>>;
    fn zero_error(&self) -> f64;
}

pub trait PrivateZeroEvader<F: Field>: ZeroEvader<F> {
    fn r(&self) -> usize;
    fn privacy_error(&self) -> f64;

    /// Lemma 9.3: distribution of `(ze, ze · (msg, r)^T)` for fresh `r ∈ F^r`,
    /// independent of `msg`.
    fn simulate<R: Rng>(&self, rng: &mut R) -> (Vec<Vec<F>>, Vec<F>);
}

/// Lemma 9.3 construction: pad any zero-evader with a random rank-`t` block.
pub struct PaddedZeroEvader<Z> { /* ... */ }
```

Provide:
- the univariate evaluation zero-evader `ze*_n(ρ) = (1, ρ, …, ρ^(n−1))` (Definition 6.1, error `(n−1)/|F|`),
- `PaddedZeroEvader<Z>` from Lemma 9.3 wrapping any `ZeroEvader`.

## Acceptance criteria

- proptest over a small extension field (e.g., `BinomialExtensionField<KoalaBear, 4>`): empirical zero-evader error within Lemma 9.3's bound.
- proptest: simulator output and the real `(M, y)` distribution are identically distributed for every fixed message.

## Notes

This issue has no dependency on other HVZK-WHIR sub-issues — pickable in parallel with #1584.

Dearop:
Hi, I’d be interested in helping with this.

[...]

PR #1636, opened 2026-05-11
feat(whir): HVZK code-switching round (Construction 9.7)

### LINKED ISSUE #1587: [HVZK-WHIR 4/6] HVZK code-switching round

XC0R:
## Summary

Implements the HVZK code-switching round for WHIR (Construction 9.7, §9.4 of eprint 2026/391).

Closes #1587.

### Protocol

Per-round ZK code-switching when `zk == true`:

1. Commit target `g = Enc(f, r')` with encoding randomness (step 1).
2. Commit mask oracle `s = Enc_zk((r' ∥ s̃), r'')` hiding the target encoding randomness (step 1b).
3. OOD answer `y = ze_ood(ρ) · (f, r', s̃)` via private zero-evader (step 3, Lemma 9.3).
4. Additive STIR/OOD corrections: sumcheck polynomial stays unpadded; the r' contribution is computed separately and stored in the proof.

| Paper | Code |
|---|---|
| Construction 9.7 step 1 (commit g) | `commit_extension_zk` |
| Construction 9.7 step 1b (commit s) | `code_switch_zk::commit_mask` |
| Construction 9.7 steps 2-3 (OOD) | `padded_ood_t1` in `round()` |
| Construction 9.7 step 4 (in-domain) | Existing STIR path + `zk_stir_correction` |
| Lemma 9.8 (HVZK simulator) | `code_switch_zk::simulate()` |
| Theorem 4.5 (composition) | `prop_zk_composition_distribution_match` |

### Design decisions

- **Additive corrections** rather than extending the sumcheck polynomial. The r' contribution to OOD and STIR openings is subtracted by the verifier. Self-authenticated through the sumcheck identity.
- **Suffix ordering only** for ZK commits.
- **Last intermediate round is plain.** Terminal HVZK is handled by the base-case IOPP (#1585).
- **`zk: false` is byte-identical** to the existing non-ZK path.

### DIFF
--- a/whir/examples/whir.rs
+++ b/whir/examples/whir.rs
@@ -157,6 +157,7 @@ fn main() {
         soundness_type,
         starting_log_inv_rate: starting_rate,
         round_log_inv_rates,
+        zk: false,
     };
 
     info!(
--- a/whir/src/parameters/mod.rs
+++ b/whir/src/parameters/mod.rs
@@ -34,6 +34,18 @@ pub struct ProtocolParameters {
     pub security_level: usize,
     /// The number of bits required for proof-of-work (PoW).
     pub pow_bits: usize,
+    /// Enable HVZK code-switching (Construction 9.7, eprint 2026/391 §9.4).
+    ///
+    /// When `true`, each non-final intermediate round:
+    /// - Commits the target polynomial with ZK randomness (`Enc(f ∥ r')`).
+    /// - Commits a mask oracle hiding the previous round's encoding randomness.
+    /// - Uses a private zero-evader for OOD answers (Lemma 9.3).
+    /// - Stores per-query STIR and OOD corrections in the proof so the
+    ///   verifier can recover plain evaluations for the sumcheck constraint.
+    ///
+    /// The composed protocol is HVZK with error `ζ_{C'} + ζ_ze + ζ_{C_zk}`
+    /// per round (Lemma 9.8), composing via Theorem 4.5.
+    pub zk: bool,
 }
 
 impl Display for ProtocolParameters {
--- a/whir/src/parameters/whir.rs
+++ b/whir/src/parameters/whir.rs
@@ -493,6 +493,7 @@ mod tests {
             folding_factor: FoldingFactor::ConstantFromSecondRound(4, 4),
             soundness_type: SecurityAssumption::CapacityBound,
             starting_log_inv_rate: 1,
+            zk: false,
         }
     }
 
--- a/whir/src/pcs/adapter.rs
+++ b/whir/src/pcs/adapter.rs
@@ -8,6 +8,9 @@ use p3_commit::{Mmcs, MultilinearPcs};
 use p3_dft::TwoAdicSubgroupDft;
 use p3_field::{ExtensionField, TwoAdicField};
 use p3_matrix::dense::DenseMatrix;
+use rand::SeedableRng;
+use rand::distr::{Distribution, StandardUniform};
+use rand::rngs::SmallRng;
 
 use super::prover::WhirProver;
 use super::verifier::WhirVerifier;
@@ -50,6 +53,7 @@ where
         + CanSampleUniformBits<F>
         + CanObserve<MT::Commitment>,
     L: Layout<F, EF>,
+    StandardUniform: Distribution<EF>,
 {
     type Commitment = MT::Commitment;
     type Val = F;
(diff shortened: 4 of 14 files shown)


tcoratger (commented):
Hey @XC0R, thanks a lot for tackling this. Some comments:

### On the ZK gate and the test coverage

In `pcs/prover/mod.rs`, `let zk_this_round = zk && !is_last_intermediate;` means the ZK code only fires when the round is *not* the last intermediate one. The three configs in `test_whir_zk_end_to_end` all have `n_rounds() == 1`, so the only round in each test *is* the last intermediate round, and the ZK branch never executes. The \"passes with `zk = true`\" signal is therefore exercising the non-ZK path in practice.

If I force a config with more rounds (e.g. `num_variables = 20`, `folding = Constant(2)`), the prover hits a `debug_assert_eq!` at `product_polynomial.rs:379` and the release build returns `SumcheckFailed` at round 6. Might be worth reproducing locally before pushing further.

### On the OOD answer ↔ sumcheck linkage

The OOD answer from `padded_ood_t1` is `f(point) + ρ^{2^n} · ze*(rc)`, but it's then passed to `EqStatement::add_evaluated_constraint(point, eval)` (`prover/mod.rs:337`), which interprets `eval` as `f(point)`. This is what causes the divergence above — the sumcheck claim ends up off by `ρ^{2^n} · ze*(rc)` relative to the true polynomial sum. The paper handles this by carrying the joint constraint (Equation 7) into the next round's constraint list, with the mask `s` opened at queried positions. Right now the mask oracle is committed in `commit_mask` and observed via `challenger.observe(mask_root)` in `committer/reader.rs:64`, but it's never opened or otherwise bound by the verifier — no queries against it, no appearance in `Constraint`.

### On the final round

`round_state.sumcheck_prover.evals().as_slice()` is observed in the clear (`prover/mod.rs:318`) regardless of `zk`, so the final folded polynomial is sent in plaintext. Per §2.7 / Theorem 4, the last reduction step is meant to invoke a ZK base-case IOPP — which is exactly what #1585 is for. Without that, even perfectly ZK intermediate rounds leak the final polynomial.

### On the simulator (Lemma 9.8)

`code_switch_zk::simulate` currently returns `(rng.random(), enc_target.simulate(positions, rng), enc_zk.simulate(positions, rng))` — three independent uniforms. The paper's simulator (proof sketch of Lemma 9.8) needs to sample `(ρ_ood, y)` jointly via the private-evader simulator from Lemma 9.3, and then have `Sim_{C'}` produce a `g` that is consistent with that simulated `y`. As-is, `_ood_point` is not read inside the function. The acceptance criterion in #1587 mentions a proptest checking that the composed simulator matches the real view; that test isn't in the PR yet, and it would likely catch this.

### On scope and rebase

Looking at the linked issue #1587, the spec is very tightly scoped: \"two local changes to a round\" — the mask `s = Enc(s̃, r'')` and the OOD replaced by `ze(ρ) · (f, r, s)`. The composition with the sumcheck side and the base-case IOPP are explicitly other PRs (#1605 just merged, #1585 covers the base case). It might be cleanest to keep this PR tight on just those two local changes, lean on the now-merged #1605 for the sumcheck side, and leave the \"send `final_poly` in the clear\" question to #1585's territory.

A rebase on top of `main` would also pick up #1605, which makes the Theorem 4.5 composition proptest from the acceptance criteria actually testable end-to-end.

### A few smaller things

- `SmallRng::seed_from_u64(0)` in `adapter.rs:83` makes the mask deterministic across runs — looks like a TODO from the WIP commit, just flagging.
- `rand_concat` ends up as `(prev_encoding_rand ∥ prev_encoding_rand ∥ s̃)` in `code_switch_zk.rs:196-199` (the second copy of `prev_encoding_rand` comes in via `mask_msg`); the paper uses `r` once, so `(r ∥ s̃)`.
- `mask_t = round_params.num_queries.max(1)` and `mask_msg_len = prev_rand.len().max(1).next_power_of_two()` — it would be nice to tie these to `t_zk` from Lemma 9.9 and `ℓ_zk − r` from the privacy bound, otherwise the soundness/ZK parameters of the resulting protocol are hard to read off the code.

Happy to go deeper on any of these if useful, and very excited to see this land once the integration with #1605 and #1585 is in place. Thanks again for pushing this forward.

tcoratger:
@latifkasuli I don't think that this PR implements the construction of the paper properly, could you check and maybe open another PR as you were assigned for this work ? No rush of course :) Just to keep you updated.

XC0R:
Updated: the multi-round configs `(12, Constant(2))` and `(18, Constant(4))` now exercise `zk_this_round == true` with mask commitment, padded OOD, and STIR corrections.

One question on paper fidelity for Construction 9.7 step 1: does the target oracle `g` need `Enc_{C'}(f, r')` (ZK encoding with fresh randomness) for Lemma 9.8 to hold, or is plain `commit_extension` sufficient? Our current HEAD uses `commit_extension_zk` with `r'` per the paper.

Issue #521, opened 2024-10-18
Implement XHash-M31


dlubarov:
From this [paper](https://eprint.iacr.org/2024/1635).

slash-aech:
hey did you guys implement this? I'll want to get my hands on this but I'd like to know if I should intervene for this big contribution in case it is untracked
I'll try to keep it similar to conventions of #1640 if assigned. Just wanted to confirm before starting anything

zexoverz:
@slash-aech are you still on this? Happy to take it otherwise, RPO-M31 from the same paper gives a close template to follow.

One question for maintainers either way: RPO landed in `p3-rescue` as a Rescue variant, and XHash is a different family, so should it go in its own crate rather than beside it?

slash-aech:
> [@slash-aech](https://github.com/slash-aech) are you still on this? Happy to take it otherwise, RPO-M31 from the same paper gives a close template to follow.
> 
> One question for maintainers either way: RPO landed in `p3-rescue` as a Rescue variant, and XHash is a different family, so should it go in its own crate rather than beside it?

hey, I actually didn't receive any reply so I am not sure to go ahead with this issue. You may consider me out for this one. All the best :)

PR #1640, opened 2026-05-12
feat: add support for RPO hash

### LINKED ISSUE #520: Implement RPO-M31
From this [paper](https://eprint.iacr.org/2024/1635). We have some related Rescue code already in `rescue/`.

Nashtare:
# Summary

Add support for RPO over Goldilocks (width 12) and RPO over Mersenne31 / KB / BB (width 24).
Ref papers:

- https://eprint.iacr.org/2022/1577.pdf
- https://eprint.iacr.org/2024/1635.pdf
- https://eprint.iacr.org/2023/1045.pdf

closes #520 

Because RPO isn't a family of hash functions like Poseidon / Rescue, I kept the `Rpo` struct private and only exposed the field-specific variants, but if we want to be loose and allow downstream users to mess with it, then we could relax them and have say `RpoGoldilocks` be just a type alias.

Note that for Goldilocks, benchmarks are a tad slower than the `miden-crypto` counterpart, which I suspect the MDS layer to be the main culprit (they do an FFT-based algorithm there), but it'd be best to integrate separately I think. I added a TODO inline.

On my M4 pro:

- RPO goldilocks: 2.6406 µs (`miden-crypto` running at 2.4465 µs)
- RPO Mersenne31: 3.4152 µs


Nashtare:
@tcoratger I'll keep it limited to RPO for now, and consider adding the XHash family (#521 generalized) separately.

tcoratger (commented):
(superseded — full review with inline comments below 👇)

tcoratger (commented):
Thanks a lot for this. I spent a bit of time to check this with Claude and the papers that you cited in the PR description. After a discussion with LLM, I have some comments, overall looks good!

Most are small, but there's one I think is worth a closer look on the M31 side before this gets merged. Inline comments below with all the context. None of this is a blocker — the addition chains, round structure, and Goldilocks side all look great. Hope this is useful!

A quick TL;DR of the inline notes:

- (M31 MDS) the `MDS_FIRST_ROW_32` constant has 8 trailing zeros where the paper has nonzero values — when I checked the resulting 24×24 matrix it has 156 zero entries and several singular 2×2 minors, so it doesn't look MDS as written. Probably worth double-checking against the paper's Eq. 16.
- (M31 / BB / KB seed strings) the seed format differs from any of the published specs I could find, which would make these non-interoperable with Miden / 3MI-Labs reference impls. Maybe worth flagging in the docstring at least.
- (BB / KB) since these aren't from a published spec, it might be nice to add a note on the round-count rationale, especially for KoalaBear with α=3.
- A handful of small docstring / dead-code / off-by-one comment things — all very minor.

Thanks again for the work here — happy to help validate fixes if useful! 🙏

rescue/src/rpo/mersenne_31.rs
```diff
@@ -0,0 +1,270 @@
...
+    1_981_550_821,
+    636_196_459,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
```
tcoratger: Maybe I'm reading this wrong, but I wanted to flag this just in case 🙏 — eprint 2024/1635 Appendix A.3 (Eq. 16) prints the **first row of the 32×32 circulant** as 32 nonzero values, formatted as 8 lines of 4 entries each. The XHash paper Appendix C.1 reprints it explicitly as a single 32-element vector with the same values.

The last 8 values in the paper are:
```
645360517, 412540024, 1649351985, 1485803845,
 53244687, 719457988,  270924307,    82564914
```

In this constant they're zeroed instead. I tried to figure out whether the truncation could still yield an MDS 24×24 by constructing the matrix `M[r][c] = MDS_FIRST_ROW_32[(c - r) mod 32]` and checking:

- 156 / 576 entries are zero (a triangular band of width 8 below the diagonal).
- Many 2×2 minors are singular — e.g. rows {0,2} × cols {0,1}:
  ```
  [[185_870_542, 2_144_994_796],
   [          0,            0]]   det = 0 mod p
  ```

So unless I'm misreading Eq. 16, the resulting matrix isn't MDS, and the paper's wide-trail / differential bounds (which rely on the MDS property — 25 active S-boxes per (FM)(BM)) wouldn't directly apply.

Happy to be wrong here! Could you double-check against the source? If it does need the missing 8 values, the `rpo_mersenne31_test_vector` regression vector will need regenerating too.

rescue/src/rpo/mersenne_31.rs
```diff
@@ -0,0 +1,270 @@
...
+
+use super::Rpo;
+use crate::util::{exp_acc, square_n};
+
+pub const RPO_M31_ALPHA: u64 = 5;
+pub const RPO_M31_WIDTH: usize = 24;
+pub const RPO_M31_CAPACITY: usize = 8;
+pub const RPO_M31_NUM_ROUNDS: usize = 7;
+
+const RPO_M31_SEED: &str = "RPO-M31:p=2147483647,m=24,c=8,n=7";
```
tcoratger: Tiny heads-up on interoperability: I had a look at how the published RPO-M31 / 3MI-Labs reference derives constants. The original RPO paper (2022/1577 §2.2) uses `"RPO({p},{m},{c},{λ})"` for the seed, and the 3MI-Labs `XHash-Constants` repo uses `"XHash({p},{m},{capacity})"` for the XHash family. I couldn't find a published RPO-M31 reference seed that matches this `"RPO-M31:p=...,m=...,c=...,n=..."` shape, so any implementation following the paper format will end up with different round constants than this one.

Might be worth either matching a designated reference seed, or adding a docstring noting this is a Plonky3-specific seed and won't interop with Miden / 3MI-Labs. Just so the next person who tries to compare hashes doesn't get surprised 🙂

rescue/src/rpo/mersenne_31.rs
```diff
@@ -0,0 +1,270 @@
...
+
+#[cfg(test)]
+mod tests {
+    use p3_field::PrimeCharacteristicRing;
+    use p3_symmetric::Permutation;
+
+    use super::*;
+
+    #[test]
+    fn rpo_mersenne31_test_vector() {
```
tcoratger, line 226: Just a thought — since the constants come from a custom seed (above) and the MDS row may need adjusting (below), once both of those settle it would be lovely to regenerate this vector against an external implementation (e.g. the 3MI-Labs SageMath reference) so it doubles as a spec-compliance check, not just a self-regression test. As-is it locks in whatever the current implementation does, which doesn't catch the kind of issue I flagged on the MDS constant.

rescue/src/rpo/mersenne_31.rs
```diff
@@ -0,0 +1,270 @@
...
+            129_587_969,
+            2_116_873_196,
+        ]
+        .map(Mersenne31::from_u32);
+
+        assert_eq!(rpo.permute(input), expected);
+    }
+
+    #[test]
+    fn mds_rpo_mersenne31_first_column_matches_paper() {
```
tcoratger: Soft suggestion: this test name is `..._matches_paper`, but it actually checks consistency between `MdsMatrixRpoMersenne31` and the local `MDS_FIRST_ROW_32` constant — not the paper itself. So if `MDS_FIRST_ROW_32` ever drifts from the paper, the test still passes. Maybe `..._matches_first_row_constant` or similar would describe what it actually proves? (Related to the MDS comment above.)

rescue/src/rpo/baby_bear.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// RPO over BabyBear at width 24 with concluding linear layer.
+///
+/// Uses Plonky3's native [`MdsMatrixBabyBear`] (24x24 circulant) and SHAKE-derived
+/// round constants from a fixed seed. No published BabyBear RPO instance exists; the
+/// parameter choice (width 24, capacity 8, 7 rounds, x^7 / x^{1/7} S-boxes) mirrors
+/// the [RPO-M31](https://eprint.iacr.org/2024/1635) layout for a comparable small
+/// field.
+#[derive(Clone)]
+pub struct RpoBabyBear {
```
tcoratger, line 32: Since there's no published RPO-BabyBear instance, would it be worth a one-liner in the docstring explaining why 7 rounds is the right number? BabyBear's α=7 mirrors RPO-Goldilocks so reusing 7 feels reasonable, but a tiny note (or pointer to where you did the analysis) would help future readers / auditors not have to redo it from scratch. Totally fine as-is otherwise!

rescue/src/rpo/baby_bear.rs
```diff
@@ -0,0 +1,189 @@
...
+
+use super::Rpo;
+use crate::util::{exp_acc, square_n};
+
+pub const RPO_BB_ALPHA: u64 = 7;
+pub const RPO_BB_WIDTH: usize = 24;
+pub const RPO_BB_CAPACITY: usize = 8;
+pub const RPO_BB_NUM_ROUNDS: usize = 7;
+
+const RPO_BB_SEED: &str = "RPO-BB:p=2013265921,m=24,c=8,n=7";
```
tcoratger, line 19: Same gentle interop note as on the M31 seed — this format `"RPO-BB:p=...,m=...,c=...,n=..."` isn't a published one. Probably fine since there's no other BB-RPO implementation to interop with, but worth a one-line docstring saying "Plonky3-specific seed" so it's explicit.

rescue/src/rpo/koala_bear.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// RPO over KoalaBear at width 24 with concluding linear layer.
+///
+/// Uses Plonky3's native [`MdsMatrixKoalaBear`] (24x24 circulant) and SHAKE-derived
+/// round constants from a fixed seed. No published KoalaBear RPO instance exists; the
+/// parameter choice (width 24, capacity 8, 7 rounds, x^3 / x^{1/3} S-boxes) mirrors
+/// the [RPO-M31](https://eprint.iacr.org/2024/1635) layout for a comparable small
+/// field, using the field's native S-box exponent.
+#[derive(Clone)]
+pub struct RpoKoalaBear {
```
tcoratger, line 32: Friendly nudge here: KoalaBear is the one where I'd most want a written round-count rationale. The Marvellous safe-rounds formula `l₁ = ⌈(s+3)/(5.5·m)⌉` and the differential bounds depend on α — with α=3 the algebraic degree grows much more slowly than with α=7 (RPO Goldilocks) or α=5 (RPO-M31), so a 7-round count carried over from those instances isn't automatically conservative.

It may very well still be plenty; just would be reassuring to see the analysis pinned down in a comment, or at minimum to flag this as experimental until someone (you, me, anyone) writes it up. 🙏

rescue/src/rpo/koala_bear.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use super::Rpo;
+use crate::util::{exp_acc, square_n};
+
+pub const RPO_KB_ALPHA: u64 = 3;
+pub const RPO_KB_WIDTH: usize = 24;
+pub const RPO_KB_CAPACITY: usize = 8;
+pub const RPO_KB_NUM_ROUNDS: usize = 7;
+
+const RPO_KB_SEED: &str = "RPO-KB:p=2130706433,m=24,c=8,n=7";
```
tcoratger, line 19: Same small interop note as the BB / M31 seeds — non-standard format, just worth a docstring line so users know it's Plonky3-internal.

rescue/src/rpo/goldilocks.rs
```diff
@@ -0,0 +1,184 @@
...
+            apply_inv_sbox_x7(state);
+        }
+    }
+}
+
+impl CryptographicPermutation<[Goldilocks; RPO_GOLDILOCKS_WIDTH]> for RpoGoldilocks {}
+
+/// Width-parallel inverse S-box `x -> x^(1/7)` over `[Goldilocks; 12]`.
+///
+/// Computes `x^10540996611094048183` via the same addition chain as
```
tcoratger: Tiny nit on the docstring — I traced both chains and they actually use **different** intermediate values to reach the same exponent. `field::exponentiation::exp_10540996611094048183` anchors on `p100 · 2^30` and inserts `011`-patterns via `+p11`; this one builds via `t₃ = t₂⁹`, `t₄ = t₃·(2⁶+1)`, etc. Both correct, both end up at `x^10540996611094048183`, but "the same addition chain" might give a future reader the impression they're line-for-line equivalent. Maybe "using an equivalent addition chain" or "a different addition chain reaching the same exponent"?

rescue/src/rpo/goldilocks.rs
```diff
@@ -0,0 +1,184 @@
...
+/// [`p3_field::exponentiation::exp_10540996611094048183`], but applied
+/// across all 12 lanes step-by-step.
+#[inline]
+fn apply_inv_sbox_x7(state: &mut [Goldilocks; RPO_GOLDILOCKS_WIDTH]) {
+    // Binary expansion of 10540996611094048183 (63 squares + 8 mults):
```
tcoratger: Super tiny: counting the operations I get 63 squares + **9** mults (1 mult in each of t3..t7, 1 mult in `t7[i].square() * t6[i]`, 2 mults in `t1[i] * t2[i] * *s`, 1 final `a*b` = 9). The field crate's chain does fit in 8 mults but this one is structurally different so it ends up with one extra. Happy to be off by one in my counting if you double-check!

rescue/src/rpo/goldilocks.rs
```diff
@@ -0,0 +1,184 @@
...
+/// RPO over Goldilocks at width 12 (eprint 2022/1577).
+#[derive(Clone, Debug)]
+pub struct RpoGoldilocks {
+    inner: Rpo<Goldilocks, MdsMatrixRpoGoldilocks, RPO_GOLDILOCKS_WIDTH, RPO_GOLDILOCKS_ALPHA>,
+}
+
+impl RpoGoldilocks {
+    /// Standard parameter choice from the paper is
+    /// `(capacity = 4, security = 128)`.
+    pub fn from_standard_constants(capacity: usize, security_level: usize) -> Self {
```
tcoratger: Soft suggestion: these two args feed only into the seed string, not into the permutation layout (width / rounds / capacity placement). So if someone calls `from_standard_constants(8, 96)` thinking they're getting a width-12 RPO at 96-bit security, they actually get the same permutation with different round constants but no warning.

Since the paper only defines (cap=4, sec=128) for width 12, maybe hardcode those like `RpoMersenne31::from_standard_constants()` does, or `debug_assert!((capacity, security_level) == (4, 128))`? Just a friendlier API guarantee.

rescue/src/rpo/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+        let byte_string = shake256_hash(seed, bytes_per_constant * num_constants);
+
+        byte_string
+            .chunks(bytes_per_constant)
+            .map(|chunk| {
+                let integer = chunk
+                    .iter()
+                    .rev()
+                    .fold(0u128, |acc, &byte| (acc << 8) | byte as u128);
+                F::from_u128(integer)
```
tcoratger, line 107: Minor heads-up (not a bug, just FYI): for 9-byte chunks reduced mod Goldilocks (~2⁶⁴), there's a tiny modular bias of ~2⁻⁸ per constant; for 5-byte chunks reduced mod M31 / BB / KB it's ~2⁻⁹. Both match exactly what the papers prescribe so this is totally fine — just worth knowing if anyone ever asks whether the constants are uniform. Rejection sampling would zero out the bias if you ever wanted that.

rescue/src/rpo/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+pub(crate) struct Rpo<F, Mds, const WIDTH: usize, const ALPHA: u64> {
+    pub(crate) num_rounds: usize,
+    pub(crate) mds: Mds,
+    pub(crate) round_constants: Vec<F>,
+    pub(crate) final_linear_layer: bool,
+}
+
+impl<F, Mds, const WIDTH: usize, const ALPHA: u64> Rpo<F, Mds, WIDTH, ALPHA>
+where
+    F: PrimeField + PermutationMonomial<ALPHA>,
```
tcoratger: Tiny thing — none of the inverse S-boxes in this file actually use `injective_exp_root_n()` (each field has its own `apply_inv_sbox_xN` with a hand-tuned width-parallel chain). So I think `PermutationMonomial<ALPHA>` could be relaxed to `InjectiveMonomial<ALPHA>` (or even just `PrimeField`) without losing anything. Pure ergonomics, not load-bearing.

rescue/src/rpo/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+/// [eprint 2024/1635](https://eprint.iacr.org/2024/1635). The original
+/// Goldilocks RPO ([eprint 2022/1577](https://eprint.iacr.org/2022/1577))
+/// does **not** include this step, and should be constructed via
+/// [`Rpo::new`] instead.
+#[derive(Clone, Debug)]
+pub(crate) struct Rpo<F, Mds, const WIDTH: usize, const ALPHA: u64> {
+    pub(crate) num_rounds: usize,
+    pub(crate) mds: Mds,
+    pub(crate) round_constants: Vec<F>,
+    pub(crate) final_linear_layer: bool,
```
tcoratger: Just flagging: the `final_linear_layer` field gets set by the two constructors but nothing ever reads it at runtime — each field's `permute_mut` statically decides whether to run the CLS based on which type it is. So either it could come out (along with one of the constructors becoming the unified entry point), or `Rpo` could grow a generic `permute_mut` that conditionally runs the final layer using this flag — which would also dedupe the four per-field implementations. Up to you; both are clean. 🙂

rescue/src/util.rs
```diff
@@ -16,6 +17,36 @@ pub(crate) fn shake256_hash(seed_bytes: &[u8], num_bytes: usize) -> Vec<u8> {
     result
 }
 
+/// Width-parallel `state.map(|x| x^(2^M))`.
+///
+/// Squares each lane `M` times. Each step squares all `N` lanes before
+/// advancing, exposing `N`-way ILP within a step so the CPU can hide
+/// per-multiplication latency. Used by the RPO inverse S-boxes.
+#[inline]
+pub(crate) fn square_n<R, const N: usize>(mut state: [R; N], m: usize) -> [R; N]
```
tcoratger: Tiny perf nudge: every call site passes a compile-time literal (`square_n(p110011, 8)`, etc.), and `exp_acc` even has `M` as a const generic but passes it through as a runtime arg here. Making this `const M: usize` would let the loop unroll cleanly and remove any concern about codegen for non-literal `m`. Drive-by suggestion.

rescue/src/rpo/koala_bear.rs
```diff
@@ -0,0 +1,187 @@
...
+            1_738_877_181,
+            1_622_784_127,
+        ]
+        .map(KoalaBear::from_u32);
+
+        assert_eq!(rpo.permute(input), expected);
+    }
+
+    #[test]
+    fn inv_sbox_x3_round_trips_through_x3() {
```
tcoratger, line 184: Maybe we could have a proptest for this for better test surface?

rescue/src/rpo/baby_bear.rs
```diff
@@ -0,0 +1,189 @@
...
+            607_611_746,
+            918_354_105,
+        ]
+        .map(BabyBear::from_u32);
+
+        assert_eq!(rpo.permute(input), expected);
+    }
+
+    #[test]
+    fn inv_sbox_x7_round_trips_through_x7() {
```
tcoratger, line 186: Maybe we could have a proptest for this for better test surface?

rescue/src/rpo/goldilocks.rs
```diff
@@ -0,0 +1,184 @@
...
+            7904287043744270535,
+            16548919317472389167,
+        ]
+        .map(Goldilocks::from_u64);
+
+        assert_eq!(rpo.permute(input), expected);
+    }
+
+    #[test]
+    fn mds_rpo_goldilocks_width_12_matches_circulant() {
```
tcoratger, line 179: Maybe we could have a proptest for this for better test surface?

rescue/src/rpo/mod.rs
Nashtare: I added a comment mentioning the paper.

rescue/src/rpo/koala_bear.rs
Nashtare: I picked 7 as plugging in the Marvellous design strategy round selection with:

```py
p = 2130706433
m = 24
capacity = 8
security_level = 128
alpha = 3
get_number_of_rounds(p, m, capacity, security_level, alpha)
```

also yields 7 (with security margin reduced to 40% as in RPO paper).

Nashtare:
Thanks @tcoratger I think I've addressed all your comments! Thanks for flagging the MDS issue, as we're using a sub-matrix for the regular instance (width 24) I thought I could just skip the last coefficients, but that was a mistake. Let me know if you have more comments!

tcoratger (approved):
This looks good to me!

PR #1641, opened 2026-05-12
ci: reduce WHIR test cost

### LINKED ISSUE #1634: ci: investigate & alleviate slow CI
CI jobs now take between 35min to 45min per run which is overly slow

It would be nice to investigate and partly alleviate it for faster & cheaper CI runs, esp. as triggered on every open PR's push

latifkasuli:
Closes #1634.

## Why

`whir::pcs::tests::test_whir_end_to_end` currently runs 1,800 WHIR commit/open/verify lifecycles in normal PR CI:

- 20 valid `(folding_factor, rs_domain_initial_reduction_factor)` pairs
- 3 soundness modes
- 3 PoW settings
- `N = 5` random table specs
- 2 layouts (`PrefixProver` and `SuffixProver`)

That cost is then multiplied across the CI matrix by both `cargo test` and `cargo test --features parallel` on Linux, Linux AVX2, macOS, and Windows.

## What

This PR keeps WHIR coverage but moves the exhaustive parameter sweep out of per-push CI:

1. Replaces the normal `test_whir_end_to_end` with a deterministic 10-case smoke matrix, still run across both layouts, for 20 total lifecycles instead of 1,800.
2. Preserves the old randomized exhaustive sweep as `test_whir_end_to_end_exhaustive`, marked `#[ignore]`.
3. Adds a new `Heavy CI` workflow with `workflow_dispatch` and a nightly cron to run the exhaustive sweep on Ubuntu and Ubuntu AVX2, both with and without `--features parallel`.
4. Adds CI concurrency cancellation so rapid pushes do not queue multiple full matrices.
5. Moves the AVX2 `RUSTFLAGS` setup before `Swatinem/rust-cache@v2`, so the AVX2 cache key reflects the actual compilation environment.

## Smoke coverage

The new smoke matrix covers each main WHIR parameter axis at least once:

| Axis | Smoke coverage |
|---|---|
| `FoldingFactor` | all 8 values from the old sweep |
| Soundness mode | `JohnsonBound`, `CapacityBound`, `UniqueDecoding` |
| `pow_bits` | `0`, `5`, `10` |
| `rs_domain_initial_reduction_factor` | `1`, `2`, `3` |
| Layout | `PrefixProver`, `SuffixProver` |
| Opening shape | fixed one-table and two-table multi-point layouts |

The randomized shape coverage from the old test is preserved by the ignored exhaustive sweep and the new Heavy CI workflow.

## Validation

Ran locally:

- `git diff --check`
- `cargo +stable sort --workspace --grouped --check`
- `actionlint .github/workflows/ci.yml .github/workflows/ci-heavy.yml`
- `cargo test -p p3-whir pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact --list`
- `cargo test -p p3-whir --features parallel pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact --list`
- `cargo test -p p3-whir`
- `cargo test -p p3-whir --features parallel`
- `cargo test`
- `cargo test --features parallel`
- `cargo +stable clippy --all-targets -- -D warnings`

### DIFF
--- a/.github/workflows/ci-heavy.yml
+++ b/.github/workflows/ci-heavy.yml
@@ -1,0 +1,46 @@
+name: Heavy CI
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 3 * * *"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: -C debuginfo=0
+
+jobs:
+  whir_exhaustive:
+    name: WHIR exhaustive tests
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+          - os: ubuntu-latest
+            features: "+avx2"
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        id: rs-stable
+
+      - name: Set flags
+        if: ${{ matrix.features }}
+        run: |
+          # Append AVX2 to existing flags so we keep -C debuginfo=0.
+          echo "RUSTFLAGS=$RUSTFLAGS -C target-feature=${{ matrix.features }}" >> "$GITHUB_ENV"
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Test WHIR exhaustive sweep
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
Thanks!

Nashtare:
@tcoratger are you ok with this? The heavy CI test can also be triggered manually when working on the WHIR codebase for instance.

tcoratger (approved):
Yes I agree this sounds reasonable. 

Indeed maybe we could add a condition so that if the WHIR crate is touched then we should run the large one?

PR #1643, opened 2026-05-12
ci: improve CI time

### LINKED ISSUE #1634: ci: investigate & alleviate slow CI

Nashtare:
## Summary

Split the heavy WHIR test causing the massive CI regression into a smaller and the (already exhausting) exhaustive one.
The former is ran as part of the default test suite, the latter is flagged `#[ignored]` and run `nightly` + `on-demand` for work on WHIR that may require some additional scrutiny from CI.

Also added suggestion from #1634 to stop early existing runs on follow-up pushes.

closes #1634

### DIFF
--- a/.github/workflows/ci-heavy.yml
+++ b/.github/workflows/ci-heavy.yml
@@ -1,0 +1,50 @@
+name: CI (heavy)
+
+# Heavy / exhaustive tests that are too slow for per-PR CI.
+# Triggered nightly and on demand.
+#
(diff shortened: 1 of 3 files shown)


Nashtare:
Duplicate of #1641

PR #1644, opened 2026-05-13
feat: add WASM32 SIMD128 support for Goldilocks

Nashtare:
Add wasm32 SIMD backend to p3-goldilocks.

Work from @WiktorStarczewski at Miden, ported to Plonky3 for general use. May inspire similar work for 31-bit fields too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,21 +104,36 @@ jobs:
           cargo build --verbose --target ${{ env.target }} -p p3-util
 
   check_wasm32_simd128:
-    name: Build wasm32 simd128
+    name: Test wasm32 simd128
     runs-on: ubuntu-latest
     if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
     env:
-      target: wasm32-unknown-unknown
+      # Compile-coverage target for browser-style consumers.
+      build_target: wasm32-unknown-unknown
+      # Runtime target — wasi exposes the syscalls (clock, write) wasmtime needs
+      # to actually execute the SIMD tests and the smoke/bench binaries.
+      run_target: wasm32-wasip1
+      RUSTFLAGS: -C debuginfo=0 -C target-feature=+simd128
+      # Tell cargo to invoke the built `.wasm` via wasmtime.
+      CARGO_TARGET_WASM32_WASIP1_RUNNER: wasmtime
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: ${{ env.target }}
+          targets: ${{ env.build_target }}, ${{ env.run_target }}
       - uses: Swatinem/rust-cache@v2
-      - name: Build p3-goldilocks with +simd128
-        run: cargo build --verbose --target ${{ env.target }} -p p3-goldilocks
-        env:
-          RUSTFLAGS: -C debuginfo=0 -C target-feature=+simd128
+      - name: Install wasmtime
+        run: |
+          curl https://wasmtime.dev/install.sh -sSf | bash
+          echo "$HOME/.wasmtime/bin" >> "$GITHUB_PATH"
+      - name: Build p3-goldilocks for wasm32-unknown-unknown
+        run: cargo build --verbose --target ${{ env.build_target }} -p p3-goldilocks
+      - name: Run wasm32+simd128 SIMD tests under wasmtime
+        run: cargo test --release --target ${{ env.run_target }} -p p3-goldilocks --lib wasm32_simd128
+      - name: Run wasm_smoke under wasmtime
+        run: cargo run --release --target ${{ env.run_target }} --bin wasm_smoke -p p3-goldilocks
+      - name: Run wasm_bench under wasmtime
+        run: cargo run --release --target ${{ env.run_target }} --bin wasm_bench -p p3-goldilocks
 
   lint:
     name: Formatting and Clippy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,8 @@ jobs:
           cargo build --verbose --target ${{ env.target }} -p p3-dft
           cargo build --verbose --target ${{ env.target }} -p p3-field
           cargo build --verbose --target ${{ env.target }} -p p3-fri
-          cargo build --verbose --target ${{ env.target }} -p p3-goldilocks
+          # `--lib` here because p3-goldilocks has host-only bins that require std and would otherwise fail
+          cargo build --verbose --target ${{ env.target }} -p p3-goldilocks --lib
           cargo build --verbose --target ${{ env.target }} -p p3-keccak
           cargo build --verbose --target ${{ env.target }} -p p3-keccak-air
           cargo build --verbose --target ${{ env.target }} -p p3-koala-bear
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,23 @@ jobs:
           cargo build --verbose --target ${{ env.target }} -p p3-uni-stark
           cargo build --verbose --target ${{ env.target }} -p p3-util
 
+  check_wasm32_simd128:
+    name: Build wasm32 simd128
+    runs-on: ubuntu-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
+    env:
+      target: wasm32-unknown-unknown
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ env.target }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Build p3-goldilocks with +simd128
+        run: cargo build --verbose --target ${{ env.target }} -p p3-goldilocks
+        env:
+          RUSTFLAGS: -C debuginfo=0 -C target-feature=+simd128
+
   lint:
     name: Formatting and Clippy
     runs-on: ubuntu-latest
--- a/goldilocks/src/goldilocks.rs
+++ b/goldilocks/src/goldilocks.rs
@@ -376,7 +376,7 @@ impl Field for Goldilocks {
     type Packing = crate::PackedGoldilocksNeon;
 
     #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
-    type Packing = crate::PackedGoldilocksWasm;
+    type Packing = crate::PackedGoldilocksWasmSimd128;
 
     #[cfg(not(any(
         all(
--- a/goldilocks/src/goldilocks.rs
+++ b/goldilocks/src/goldilocks.rs
@@ -375,6 +375,9 @@ impl Field for Goldilocks {
     #[cfg(target_arch = "aarch64")]
     type Packing = crate::PackedGoldilocksNeon;
 
+    #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
+    type Packing = crate::PackedGoldilocksWasm;
+
     #[cfg(not(any(
         all(
             target_arch = "x86_64",
@@ -383,6 +386,7 @@ impl Field for Goldilocks {
         ),
         all(target_arch = "x86_64", target_feature = "avx512f"),
         target_arch = "aarch64",
+        all(target_arch = "wasm32", target_feature = "simd128"),
     )))]
     type Packing = Self;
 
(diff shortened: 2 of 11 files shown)


goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+    const TWO: Self = Self::broadcast(Goldilocks::TWO);
+    const NEG_ONE: Self = Self::broadcast(Goldilocks::NEG_ONE);
+
+    #[inline]
+    fn from_prime_subfield(f: Self::PrimeSubfield) -> Self {
+        f.into()
+    }
+
+    #[inline]
+    fn halve(&self) -> Self {
```
adr1anh, line 147: Should we implement `double` as well?

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+//
+//      uint64x2_t                 → v128
+//      veorq_u64(a, b)            → v128_xor(a, b)
+//      vaddq_u64(a, b)            → i64x2_add(a, b)
+//      vsubq_u64(a, b)            → i64x2_sub(a, b)
+//      vcgtq_s64(a, b)            → i64x2_gt(a, b)
+//      vbicq_u64(a, b)            → v128_andnot(a, b)  (= a & !b)
+//      vshrq_n_u64::<32>(a)       → u64x2_shr(a, 32)
+//      vdupq_n_u64(x)             → u64x2_splat(x)
+//      vreinterpretq_s64_u64(x)   → identity (v128 is type-erased)
```
adr1anh: This should be a module doc comment. Also it still refers to plonky3. 

Overall, I'd point more directly at the NEON optimization, and explain this is a translation.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+
+/// Add `2^63` with overflow. Needed to emulate unsigned comparisons.
+#[inline(always)]
+fn shift(x: v128) -> v128 {
+    v128_xor(x, SIGN_BIT)
+}
+
+/// If `x_s < SHIFTED_FIELD_ORDER` (signed comparison), add `EPSILON` to canonicalize.
+/// The neon impl uses `vbicq_u64(EPSILON_VEC, mask) = EPSILON_VEC & !mask`. wasm32's
+/// `v128_andnot(a, b) = a & !b` matches.
```
adr1anh: I'd but this as a comment in the function, it's an implementation detail.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+/// Inputs can be arbitrary, output is not guaranteed to be less than `FIELD_ORDER`.
+#[inline]
+fn sub(x: v128, y: v128) -> v128 {
+    let y_s = canonicalize_s(shift(y));
+    let x_s = shift(x);
+    let mask = i64x2_gt(y_s, x_s);
+    let wrapback_amt = u64x2_shr(mask, 32);
+    let res_wrapped = i64x2_sub(x_s, y_s);
+    i64x2_sub(res_wrapped, wrapback_amt)
+}
```
tcoratger, line 292: I feel like this method is used only once in the file, so maybe this is better to directly put the logic at the appropriate location and avoid this additional function definition no?

I think we have couple of identical stuffs with for example the add function and other similar methods no?

adr1anh:
Does the current workspace support benchmarking in WASM? It would be nice to have a comparison with the non-SIMD implementation.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+    reduce128(hi, lo)
+}
+
+/// Goldilocks modular square. Computes `x^2 mod FIELD_ORDER`.
+///
+/// No specialized squaring path on simd128 — falls through to `mul`.
+#[inline(always)]
+fn square(x: v128) -> v128 {
+    mul(x, x)
+}
```
tcoratger: Same here can't we just put the logic directly at the usage place? Maybe simpler for reading to avoid going through many functions

Nashtare:
> Does the current workspace support benchmarking in WASM? It would be nice to have a comparison with the non-SIMD implementation.

It doesn't, and AFAIK, it would require some substantial change to setup and run (but I may be wrong there?). For now I've only added a CI job to check build passes fine.

.github/workflows/ci.yml
```diff
@@ -100,6 +100,23 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ env.target }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Build p3-goldilocks with +simd128
+        run: cargo build --verbose --target ${{ env.target }} -p p3-goldilocks
+        env:
+          RUSTFLAGS: -C debuginfo=0 -C target-feature=+simd128
```
tcoratger: Is it normal that we just check the building pipeline and that we don't have the tests running?

goldilocks/src/wasm32_simd128/packing.rs
tcoratger: Can't we gain speed via the following?


```suggestion
/// Full 64×64 → 128 squaring.
/// Exploits `lh = hl` so only three 32×32 products are needed instead of four.
#[inline]
fn square64(x: v128) -> (v128, v128) {
    let x_lo = lo32(x);
    let x_hi = hi32(x);

    let ll = mul_u32_lanes(x_lo, x_lo);
    let lh = mul_u32_lanes(x_lo, x_hi);
    let hh = mul_u32_lanes(x_hi, x_hi);

    // 128-bit product = ll + lh·2^33 + hh·2^64.
    let ll_hi = u64x2_shr(ll, 33);
    let t0 = i64x2_add(lh, ll_hi);
    let t0_hi = u64x2_shr(t0, 31);
    let res_hi = i64x2_add(hh, t0_hi);

    let lh_shifted = i64x2_shl(lh, 33);
    let res_lo = i64x2_add(ll, lh_shifted);

    (res_hi, res_lo)
}

#[inline]
fn square(x: v128) -> v128 {
    let (hi, lo) = square64(x);
    reduce128(hi, lo)
}
```

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+}
+
+impl_add_assign!(PackedGoldilocksWasm);
+impl_sub_assign!(PackedGoldilocksWasm);
+impl_mul_methods!(PackedGoldilocksWasm);
+ring_sum!(PackedGoldilocksWasm);
+impl_rng!(PackedGoldilocksWasm);
+
+impl PrimeCharacteristicRing for PackedGoldilocksWasm {
```
tcoratger: Can't we override dot product and mixed dot product as well?

tcoratger: Maybe something like this, which mimics the Neon implementation.

```rust
#[inline]
fn dot_product<const N: usize>(lhs: &[Self; N], rhs: &[Self; N]) -> Self {
    Self::from_fn(|lane| {
        let lhs_lane: [Goldilocks; N] = core::array::from_fn(|i| lhs[i].as_slice()[lane]);
        let rhs_lane: [Goldilocks; N] = core::array::from_fn(|i| rhs[i].as_slice()[lane]);
        Goldilocks::dot_product(&lhs_lane, &rhs_lane)
    })
}
```

and 

```rust
#[inline]
fn mixed_dot_product<const N: usize>(a: &[Self; N], f: &[Goldilocks; N]) -> Self {
    Self::from_fn(|lane| {
        let a_lane: [Goldilocks; N] = core::array::from_fn(|i| a[i].as_slice()[lane]);
        Goldilocks::dot_product(&a_lane, f)
    })
}
```

goldilocks/src/goldilocks.rs
```diff
@@ -375,6 +375,9 @@ impl Field for Goldilocks {
     #[cfg(target_arch = "aarch64")]
     type Packing = crate::PackedGoldilocksNeon;
 
+    #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
+    type Packing = crate::PackedGoldilocksWasm;
```
tcoratger: Isn't it more logical to have this? Not sure but since the type is gated on target_feature = "simd128", not just target_arch = "wasm32", maybe this makes more sense.

```suggestion
    type Packing = crate::PackedGoldilocksSimd128;
```

goldilocks/src/wasm32_simd128/packing.rs
Nashtare: There is no efficient way to perform `double()` for non canonical inputs on 128bit lanes I believe, I added an inline implementation that falls back to `add` internally for consistency though.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,444 @@
...
+};
+
+/// Vectorized wasm32-simd128 implementation of `Goldilocks` arithmetic.
+///
+/// `repr(transparent)` over `[Goldilocks; WIDTH]` so we can `transmute` freely
+/// between `[Goldilocks; 2]`, `[u64; 2]`, and `v128`.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+#[repr(transparent)]
+#[must_use]
+pub struct PackedGoldilocksWasm(pub [Goldilocks; WIDTH]);
```
tcoratger: Isn't it more logical to have this?

In order maybe to make more sense given the fact that the type is gated on target_feature = "simd128", not just target_arch = "wasm32".

```suggestion
pub struct PackedGoldilocksSimd128(pub [Goldilocks; WIDTH]);
```

tcoratger (approved):
LGTM

Just added small points but none  of them are blockers for merging.

goldilocks/src/wasm32_simd128/packing.rs
Nashtare: I'd rather keep them separate to not blowup other functions, it should be fine with the `#[inline]` directive anyway (I hope) and make it more readable / more easily reusable IMO. WDYT?

.github/workflows/ci.yml
Nashtare: Fixed it, added CI coverage

Nashtare:
@adr1anh I added a smoke bench for Goldilocks under `goldilocks/bin/wasm_bench.rs` but due to the lack of criterion, results can be flaky to to be taken with a pinch of salt

goldilocks/src/wasm32_simd128/packing.rs
tcoratger: Yes sure! Makes sense

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -0,0 +1,489 @@
...
+///
+///      uint64x2_t                 → v128
+///      veorq_u64(a, b)            → v128_xor(a, b)
+///      vaddq_u64(a, b)            → i64x2_add(a, b)
+///      vsubq_u64(a, b)            → i64x2_sub(a, b)
+///      vcgtq_s64(a, b)            → i64x2_gt(a, b)
+///      vbicq_u64(a, b)            → v128_andnot(a, b)  (= a & !b)
+///      vshrq_n_u64::<32>(a)       → u64x2_shr(a, 32)
+///      vdupq_n_u64(x)             → u64x2_splat(x)
+///      vreinterpretq_s64_u64(x)   → identity (v128 is type-erased)
```
tcoratger:
```suggestion
//! Resources:
//! 1. WebAssembly SIMD proposal: https://github.com/WebAssembly/simd/blob/main/proposals/simd/SIMD.md
//! 2. The arithmetic recipes are the standard Goldilocks SIMD recipes, mimicking the existing
//!    `aarch64_neon` and `x86_64_avx2` backends with the following intrinsic correspondence:
//!
//!      uint64x2_t                 → v128
//!      veorq_u64(a, b)            → v128_xor(a, b)
//!      vaddq_u64(a, b)            → i64x2_add(a, b)
//!      vsubq_u64(a, b)            → i64x2_sub(a, b)
//!      vcgtq_s64(a, b)            → i64x2_gt(a, b)
//!      vbicq_u64(a, b)            → v128_andnot(a, b)  (= a & !b)
//!      vshrq_n_u64::<32>(a)       → u64x2_shr(a, 32)
//!      vdupq_n_u64(x)             → u64x2_splat(x)
//!      vreinterpretq_s64_u64(x)   → identity (v128 is type-erased)

```

tcoratger (approved):
LGTM

PR #1645, opened 2026-05-13
fix: faster packed than scalar Goldilocks MDS layer for poseidon1

### LINKED ISSUE #1642: Goldilocks Poseidon1 packed paths slower than scalar on AVX-2 (apply_circulant MDS fallback)
## Summary

The `MdsMatrixGoldilocks` `Permutation` impls for `PackedGoldilocksAVX2` and `PackedGoldilocksAVX512` (widths 8, 12, 16) dispatch the MDS multiply through `p3_mds::util::apply_circulant`, the documented O(N²) fallback at `mds/src/util.rs:23-26`.

While PR #1378 provided specialized Karatsuba MDS paths for Monty31, Goldilocks was left using the generic apply_circulant fallback, creating a significant performance disparity between the two fields on x86.

## Measured (Zen 5, `RUSTFLAGS="-C target-feature=+avx2"`, 100k iterations × 3 runs)

| Permutation | vpmuludq retired | Cycles | ns/iter | Effective vs scalar |
|---|---:|---:|---:|---:|
| Poseidon1 width-8 scalar | (baseline) | 3,489 | 635.4 | 1.00× |
| Poseidon1 width-8 packed AVX-2 (4 lanes) | **4,828** | 18,460 | 3,308.6 | **0.77×** (regression) |
| Poseidon2 width-8 packed AVX-2 (4 lanes, reference) | 1,204 | 9,377 | 1,665.5 | 1.04× |

vpmuludq counted via AMD Zen 5 `packed_int_op_type.int256_mul`. The 4,828 figure matches a source-level loop trace exactly:

> 4 initial + 4 terminal full rounds × (8 S-box × 14 + 8 row-dot-product × 32) + 22 partial rounds × 74 + one-shot `mds_multiply(m_i)` 256 = 4,828.

AVX-512 dispatch (`goldilocks/src/x86_64_avx512/mds.rs:20-26`) has the same gap; 8 lanes absorb more overhead but still under-deliver (~1.84× of an 8× lane ceiling on width-8 Poseidon1).

## Mechanism

Packed dispatch (`goldilocks/src/x86_64_avx2/mds.rs:20-26`):

```rust
impl Permutation<[PackedGoldilocksAVX2; 8]> for MdsMatrixGoldilocks {
    fn permute(&self, input: [PackedGoldilocksAVX2; 8]) -> [PackedGoldilocksAVX2; 8] {
        const MATRIX_CIRC_MDS_8_SML_ROW_U64: [u64; 8] = convert_array(MATRIX_CIRC_MDS_8_SML_ROW);
        apply_circulant(&MATRIX_CIRC_MDS_8_SML_ROW_U64, &input)
    }
}
```

Inside `apply_circulant` the small `i64` coefficients (1, 3, 4, 7, 8, 9 for width 8) are broadcast to packed elements via `R::from_u64`, losing the small-constant structure that scalar's `SmallConvolveGoldilocks` (`mds.rs:26-68`) preserves with an `i64 × i128` accumulator. Each MDS coefficient × state then issues a full packed Goldilocks mul.

Same dispatch lives at:
- `goldilocks/src/x86_64_avx2/mds.rs:20-50` (widths 8, 12, 16, 24)
- `goldilocks/src/x86_64_avx512/mds.rs:20-50` (same widths)

## Fix precedent

Path described in `poseidon1/src/external.rs:50-63` (`dense_mds` doc). PR #1378 implemented it for Monty31 via `mds_circulant_karatsuba_{16,24}` in the shared crate plus per-arch `poseidon1.rs` files. For Goldilocks the equivalent additions would be `_8` and `_12` in the shared crate (the underlying `FieldConvolve::conv8/12` already exist, so the wrappers are ~2 lines each) plus `goldilocks/src/{avx2,avx512}/poseidon1.rs`. `PackedGoldilocksAVX2::halve()` already exists (`goldilocks/src/x86_64_avx2/packing.rs:130`), so the prerequisite is in place.

Rough cycle-budget projection from the Poseidon2 reference puts the fix benefit at ~2× wall time on AVX-2; actual fix benchmark is the load-bearing number.

## Reproducibility

Perf-counter harnesses are at `goldilocks/examples/poseidon{1,2}_perf_w8.rs` on a local branch (not yet upstream, if approved, will follow as a separate small PR, or in the same).

```bash
RUSTFLAGS="-C target-feature=+avx2" \
  cargo build --release --example poseidon1_perf_w8 -p p3-goldilocks
RUSTFLAGS="-C target-feature=+avx2" \
  cargo build --release --example poseidon2_perf_w8 -p p3-goldilocks

perf stat -e packed_int_op_type.int256_mul,cycles,instructions -r 3 \
  target/release/examples/poseidon1_perf_w8 100000
perf stat -e packed_int_op_type.int256_mul,cycles,instructions -r 3 \
  target/release/examples/poseidon2_perf_w8 100000
```

## Scope question

- **Widths.** Default is 8 and 12; width 16 via `Poseidon1GoldilocksGeneric`. All three, or just 8 and 12?
- **Archs.** AVX-2 and AVX-512 share the gap. Single PR for both, or split?
- **Partial-round path.** `cheap_matmul`'s `A::mixed_dot_product` (`poseidon1/src/internal.rs:163`) is also unspecialised for packed types. Same PR, or follow-up?
- **Latency-hiding state splits.** `InternalLayer8` / `InternalLayer12` (Goldilocks equivalents of #1378's `InternalLayer16/24`) don't exist yet. Worth building, or skip to keep the diff smaller?

## Coordination

Happy to take the implementation following the #1378 template if that helps. Equally happy to defer if you'd prefer to take it yourself or have a different shape in mind. The documented `dense_mds` plan is unambiguous either way, and the audit numbers don't change with authorship.

Nashtare:
Dispatch to Karatsuba implementation instead of relying on naive O(n^2) alg for the MDS layer on AVX.

closes #1642

### DIFF
--- a/goldilocks/src/x86_64_avx2/mds.rs
+++ b/goldilocks/src/x86_64_avx2/mds.rs
@@ -16,7 +16,7 @@ use crate::{
 /// to feed the Karatsuba helpers.
 const fn sml_row_to_goldilocks_col<const N: usize>(row: &[i64; N]) -> [Goldilocks; N] {
     let col_i64 = first_row_to_first_col(row);
-    let mut col = [Goldilocks::new(0); N];
+    let mut col = [Goldilocks::ZERO; N];
     let mut i = 0;
     while i < N {
         col[i] = Goldilocks::new(col_i64[i] as u64);
--- a/goldilocks/src/x86_64_avx2/mds.rs
+++ b/goldilocks/src/x86_64_avx2/mds.rs
@@ -1,26 +1,36 @@
 use p3_mds::MdsPermutation;
-use p3_mds::util::apply_circulant;
+use p3_mds::karatsuba_convolution::{
+    mds_circulant_karatsuba_8, mds_circulant_karatsuba_12, mds_circulant_karatsuba_16,
+};
+use p3_mds::util::{apply_circulant, first_row_to_first_col};
 use p3_symmetric::Permutation;
 
 use crate::x86_64_avx2::packing::PackedGoldilocksAVX2;
 use crate::{
-    MATRIX_CIRC_MDS_8_SML_ROW, MATRIX_CIRC_MDS_12_SML_ROW, MATRIX_CIRC_MDS_16_SML_ROW,
+    Goldilocks, MATRIX_CIRC_MDS_8_SML_ROW, MATRIX_CIRC_MDS_12_SML_ROW, MATRIX_CIRC_MDS_16_SML_ROW,
     MATRIX_CIRC_MDS_24_GOLDILOCKS, MdsMatrixGoldilocks,
 };
-const fn convert_array<const N: usize>(arr: [i64; N]) -> [u64; N] {
-    let mut result: [u64; N] = [0; N];
+
+/// Convert a `[i64; N]` row of small non-negative MDS coefficients into the
+/// matching circulant first column as `[Goldilocks; N]`. Used at compile time
+/// to feed the Karatsuba helpers.
+const fn sml_row_to_goldilocks_col<const N: usize>(row: &[i64; N]) -> [Goldilocks; N] {
+    let col_i64 = first_row_to_first_col(row);
+    let mut col = [Goldilocks::new(0); N];
     let mut i = 0;
     while i < N {
-        result[i] = arr[i] as u64;
+        col[i] = Goldilocks::new(col_i64[i] as u64);
         i += 1;
     }
-    result
+    col
 }
 
 impl Permutation<[PackedGoldilocksAVX2; 8]> for MdsMatrixGoldilocks {
     fn permute(&self, input: [PackedGoldilocksAVX2; 8]) -> [PackedGoldilocksAVX2; 8] {
-        const MATRIX_CIRC_MDS_8_SML_ROW_U64: [u64; 8] = convert_array(MATRIX_CIRC_MDS_8_SML_ROW);
-        apply_circulant(&MATRIX_CIRC_MDS_8_SML_ROW_U64, &input)
+        const COL: [Goldilocks; 8] = sml_row_to_goldilocks_col(&MATRIX_CIRC_MDS_8_SML_ROW);
+        let mut state = input;
+        mds_circulant_karatsuba_8(&mut state, &COL);
+        state
     }
 }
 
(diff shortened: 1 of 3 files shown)


goldilocks/src/x86_64_avx2/mds.rs
```diff
@@ -1,44 +1,58 @@
...
 };
-const fn convert_array<const N: usize>(arr: [i64; N]) -> [u64; N] {
-    let mut result: [u64; N] = [0; N];
+
+/// Convert a `[i64; N]` row of small non-negative MDS coefficients into the
+/// matching circulant first column as `[Goldilocks; N]`. Used at compile time
+/// to feed the Karatsuba helpers.
+const fn sml_row_to_goldilocks_col<const N: usize>(row: &[i64; N]) -> [Goldilocks; N] {
+    let col_i64 = first_row_to_first_col(row);
+    let mut col = [Goldilocks::new(0); N];
```
tcoratger: Can't we use this instead?

```suggestion
    let mut col = [Goldilocks::ZERO; N];
```

goldilocks/src/x86_64_avx512/mds.rs
```diff
@@ -1,44 +1,58 @@
...
 };
-const fn convert_array<const N: usize>(arr: [i64; N]) -> [u64; N] {
-    let mut result: [u64; N] = [0; N];
+
+/// Convert a `[i64; N]` row of small non-negative MDS coefficients into the
+/// matching circulant first column as `[Goldilocks; N]`. Used at compile time
+/// to feed the Karatsuba helpers.
+const fn sml_row_to_goldilocks_col<const N: usize>(row: &[i64; N]) -> [Goldilocks; N] {
+    let col_i64 = first_row_to_first_col(row);
+    let mut col = [Goldilocks::new(0); N];
```
tcoratger: Same, maybe we could use instead this

```suggestion
    let mut col = [Goldilocks::ZERO; N];
```

tcoratger (approved):
LGTM, just super minor comments, but not a blocker to merge!

tcoratger:
@Nashtare I think we just miss an import and ready to merge :)

PR #1646, opened 2026-05-13
whir: fail fast on STIR query and index count mismatch

lmorett1:
Add a pre-loop length check in `verify_merkle_proof`, so malformed proofs return `StirQueryCountMismatch` instead of being silently truncated.

### DIFF
--- a/whir/src/pcs/verifier/mod.rs
+++ b/whir/src/pcs/verifier/mod.rs
@@ -336,6 +336,14 @@ where
                 .queries
         };
 
+        if queries.len() != indices.len() {
+            return Err(VerifierError::StirQueryCountMismatch {
+                round_index,
+                expected: indices.len(),
+                actual: queries.len(),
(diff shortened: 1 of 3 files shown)


whir/src/pcs/verifier/mod.rs
```diff
@@ -336,6 +336,14 @@ where
                 .queries
         };
 
+        if queries.len() != indices.len() {
+            return Err(VerifierError::StirQueryCountMismatch {
+                round_index,
+                expected: indices.len(),
+                actual: queries.len(),
+            });
+        }
```
tcoratger, line 345: Can we have an end to end test to assert on this error?

tcoratger: I've just added the two small tests since this was pretty fast to do, ready to merge.

tcoratger (approved):

PR #1544, opened 2026-04-11
feat(merkle-tree): add Merkle path pruning for compact multi-opening proofs

tcoratger:
Useful in context like leanVM, see https://github.com/leanEthereum/leanMultisig/blob/a03c1da911f9a3acfa1f5e8f71d6cdcc600104cc/crates/backend/fiat-shamir/src/merkle_pruning.rs

## Summary

- Adds a `pruning` module to `p3-merkle-tree` that deduplicates redundant sibling digests when opening multiple leaves in the same Merkle tree
- Sorts paths by leaf index, computes the Lowest Common Ancestor (LCA) between consecutive pairs, and strips shared upper siblings
- Adds `open_batch_pruned()` and `verify_batch_pruned()` methods on `MerkleTreeMmcs` for end-to-end usage
- Typical proof size reduction: **30-40%** for FRI parameters (80 queries, height 20)

## Performance design

- **Flat contiguous memory**: single `Vec` per path, no nested `Vec<Vec>` — cache-friendly `memcpy` during restoration
- **O(1) binary LCA**: XOR + leading_zeros, with the binary-tree check hoisted outside the hot loop
- **O(h) N-ary LCA**: bitwise shifts instead of integer division
- **Index-only sorting**: sort lightweight `u32` array, not full path structs
- **Zero-copy verification**: siblings borrowed directly, opening consumed by value (no deep clone)
- **Pre-allocated buffers**: exact sibling capacity computed upfront, zero resizing

## Test plan

- [x] 36 unit tests in `pruning.rs`: LCA, level offsets, total siblings, pruning, roundtrip (binary + 4-ary + mixed), error bounds, size analysis
- [x] 6 integration tests in `mmcs.rs`: binary roundtrip, 4-ary roundtrip, tampered proof rejection, size comparison, duplicate handling, mixed matrix heights
- [x] 5 proptest strategies: binary roundtrip, quad roundtrip, pruned size <= original, fast-path matches generic, dedup preserves identity
- [x] All 81 tests pass, zero clippy warnings, zero fmt diff
- [x] All 39 pre-existing merkle-tree tests still pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1648, opened 2026-05-13
perf(merkle-tree): amortize verifier cost in verify_batch_pruned

### LINKED ISSUE #1599: Amortize verifier cost in MerkleTreeMmcs::verify_batch_pruned
## Context

PR #1544 introduced `MerkleTreeMmcs::open_batch_pruned` / `verify_batch_pruned`, which compresses multi-opening proofs by removing sibling digests shared between authentication paths above their LCA.

The proof side is amortized — each shared sibling appears once instead of `k` times across `k` queries.

The verifier side is **not yet amortized**. `verify_batch_pruned` currently:

1. Restores the full per-path sibling buffers from the pruned representation.
2. Calls `self.verify_batch` once per restored path.

That re-runs the upper-level compressions once per path even though those nodes are shared — exactly the work the proof compression already identified as redundant.

A `TODO` was left in `merkle-tree/src/mmcs.rs` next to the per-path verify loop:

```rust
// TODO: amortize verifier cost across paths the same way the proof
// is amortized — shared upper-level compressions are recomputed once
// per path instead of once per shared subtree.
```

## Goal

Verify each shared internal node exactly once. The pruning algorithm already computes the LCA structure between sorted paths; the same structure can drive a single bottom-up sweep that hashes each node at most once and checks the resulting set against the cap.

## Notes

- Should preserve the existing public API (`verify_batch_pruned` signature, error type).
- Existing test coverage in `merkle-tree/src/mmcs.rs` (binary, 4-ary, mixed-heights, tampered-proof, duplicate-indices, proptests) and in `merkle-tree/src/pruning.rs` (mixed-arity unit + proptests) should continue to pass unchanged.
- Follow-up to PR #1544 review comment by @Nashtare: https://github.com/Plonky3/Plonky3/pull/1544#discussion_r3165265081

wstran:
## Summary

Follow-up to #1544. Closes #1599.

`verify_batch_pruned` restored each authentication path and then called `verify_batch` once per query, recomputing every upper-level compression once per path even though those nodes are shared in the tree. `merkle-tree/src/mmcs.rs` carried a `TODO` next to the per-path loop pointing at this.

This walks the tree once for all queries together: sorted unique paths are grouped by parent index at each layer, every group runs a single `compress()` pulling any missing siblings from the lead path's buffer. Matrix injection at mixed-height boundaries amortizes the same way — group members share the row index, so the shorter matrix's row digest is hashed once per group.

Duplicate query indices are folded into one verified path; the verifier rejects when duplicates carry inconsistent opened values, so dedup can't be used to smuggle different values past the check.

## Scope

`verify_batch_pruned` is the consumer-facing entry point for callers using the pruning API introduced in #1544 (e.g. leanVM, see the link in #1544's description). No in-tree FRI/STARK verifier currently calls it, so this PR does not move the needle on `cargo run --example` end-to-end timings. It does close the explicit `TODO` left in the post-pruning verifier and brings the verifier-side cost in line with the prover-side amortization that #1544 already shipped.

## Behaviour

No public-API change. `verify_batch_pruned` still takes the opening by value and returns the opened rows on success. The accept/reject contract is preserved — the existing roundtrip proptests against per-path `open_batch` openings (binary + 4-ary + mixed schedules) pass unchanged.

## Bench

`cargo bench -p p3-merkle-tree --bench pruned_verify`, M1 (single machine), BabyBear / Poseidon2-16 / `cap_height = 0`, 3 runs each version, criterion median reported. Old timings collected by running the same bench script against `main` at `48bdd188` (apples-to-apples: both sides exercise `MerkleTreeMmcs::verify_batch_pruned`):

| Config                  | Old `verify_batch_pruned` | New `verify_batch_pruned` | Speedup |
|-------------------------|--------------------------:|--------------------------:|--------:|
| height 2^16, 40 queries | 501 µs (497–506)          | 351 µs (348–352)          | **1.43×** |
| height 2^16, 80 queries | 1014 µs (990–1038)        | 656 µs (654–660)          | **1.55×** |
| height 2^20, 40 queries | 621 µs (617–624)          | 466 µs (463–469)          | **1.33×** |
| height 2^20, 80 queries | 1236 µs (1219–1248)       | 887 µs (881–896)          | **1.39×** |

Speedup grows with query count — query clusters merge sooner at the upper layers. Only the M1 aarch64 path was measured locally; x86_64 SIMD widths differ and I don't have ready access, so I'm not reporting numbers I haven't run.

## Test plan

- [x] `cargo test -p p3-merkle-tree --lib` — 88 passed (84 existing + 4 new).
- [x] `cargo test --workspace --lib` — all green.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `cargo sort --workspace --grouped --check` — clean.
- [x] `cargo build --target thumbv7em-none-eabi -p p3-merkle-tree` (no_std) — clean.

New tests cover: non-power-of-two height with mid-tree LCA, queries clustered into contiguous blocks, full leaf coverage (one group survives to the cap), and the duplicate-with-inconsistent-openings rejection.

### DIFF
--- a/merkle-tree/Cargo.toml
+++ b/merkle-tree/Cargo.toml
@@ -37,5 +37,9 @@ criterion.workspace = true
 name = "merkle_tree"
 harness = false
 
+[[bench]]
+name = "pruned_verify"
+harness = false
+
 [lints]
 workspace = true
--- a/merkle-tree/benches/pruned_verify.rs
+++ b/merkle-tree/benches/pruned_verify.rs
@@ -1,0 +1,104 @@
+//! Benchmark for amortized pruned-batch verification.
+//!
+//! Compares the new amortized `verify_batch_pruned` against an unamortized
+//! baseline that verifies each restored authentication path independently —
+//! the same work the FRI-style verifier would do today.
+//!
+//! FRI-shaped parameters (height 1<<20, 80 queries, binary tree) are the
+//! intended consumer of this code path.
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+use p3_commit::{BatchOpeningRef, Mmcs};
+use p3_field::Field;
+use p3_matrix::Matrix;
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+type F = BabyBear;
+type Perm = Poseidon2BabyBear<16>;
+type H = PaddingFreeSponge<Perm, 16, 8, 8>;
+type C = TruncatedPermutation<Perm, 2, 8, 16>;
+type Mmcs2 = MerkleTreeMmcs<<F as Field>::Packing, <F as Field>::Packing, H, C, 2, 8>;
+
+fn make_mmcs(seed: u64) -> Mmcs2 {
+    let mut rng = SmallRng::seed_from_u64(seed);
+    let perm = Perm::new_from_rng_128(&mut rng);
+    let h = H::new(perm.clone());
+    let c = C::new(perm);
+    Mmcs2::new(h, c, 0)
+}
+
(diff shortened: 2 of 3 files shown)


tcoratger (approved):
That's great, I've just added couple of tests to harden the coverage, LGTM.

Issue #1649, opened 2026-05-13
[HVZK-WHIR 3b/7] HVZK sumcheck overlay for the suffix-binding layout


camofu:
## Context

PR #1605 lands the HVZK sumcheck overlay (Construction 6.3 of eprint 2026/391) on top of the prefix-binding stacked-sumcheck layout (`PrefixProver`). The non-ZK `src/sumcheck/layout/` module supports two binding orders — `PrefixProver` and `SuffixProver` — which differ in which fast-path tricks apply during preprocessing rounds but produce the same residual product polynomial.

The umbrella issues for the HVZK-WHIR effort (#1586, #1587, #1588, #1589) were written before the prefix/suffix split landed, so the suffix-binding HVZK overlay was never called out separately. It is needed once HVZK reaches the WHIR PCS layer: the PCS selects either binding mode via the `Layout` trait, so if HVZK is wired per-layout the suffix path would lack a privacy overlay — a user driving WHIR-PCS through the suffix layout would get a non-private proof with no obvious error.

## Goal

Add `ZkSuffixProver`, a symmetric overlay over `SuffixProver` mirroring `ZkPrefixProver` over `PrefixProver`. 

## Scope of work

Most of `ZkSuffixProver` will mirror `ZkPrefixProver` line-by-line: the per-round `ĥ_j` formula reads the plain piece `(c_0, c_∞)` from per-claim accumulators (which `SuffixProver::into_sumcheck` already builds the same way), then layers on the mask, past-mask, future-mask, and `ε·plain` terms exactly as Construction 6.3 dictates. Two small architectural details don't carry over for free.

### Multiplying the residual polynomial by ε

After the preprocessing rounds, the HVZK overlay needs to fold the base polynomial against the sampled challenges and scale the result by `ε`, so the residual sumcheck's claimed sum is `ε · plain_residual_sum`.

In the prefix path, `ZkPrefixProver` accesses the stacked polynomial directly (`inner.poly`) and calls `compress_prefix_to_packed(&rs, ε)` — the `ε` slots into the existing scale argument and the cost is zero.

[...]

Issue #1589, opened 2026-04-28
[HVZK-WHIR 6/6] Compose HVZK-WHIR PCS + end-to-end tests


tcoratger:
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 10 / Theorem 4](https://eprint.iacr.org/2026/391.pdf). Depends on #1586, #1587, #1588.

## What

Wire the building blocks into a `HidingWhirPcs` adapter (mirroring the existing `HidingFriPcs`), and add a `ZK: bool` const generic to `WhirPcs`. The composition follows Theorem 4.5 — no new cryptography here, just plumbing.

## Acceptance criteria

- `HidingWhirPcs<F, EF, …>` implementing `MultilinearPcs<EF, Challenger>`.
- `WhirParameters::new_testing_zk()` and `new_benchmark_zk()` (mirror the FRI helpers in `fri/src/config.rs`).
- `whir/tests/zk_e2e.rs`:
  - 32 random multilinear polynomials, sizes `2^16 ≤ n ≤ 2^20`, full commit/open/verify roundtrip with `ZK = true`.
  - With `ZK = false`, the transcript is byte-identical to today's WHIR (regression guard).
- `whir/benches/zk_overhead.rs`: prover time, verifier time, proof size for `ZK = true` vs `ZK = false`. Paper claims `1 + o(1)` overhead — flag CI regressions.
- Update `whir/src/lib.rs` doc comment to cite both papers (eprint 2024/1586 and eprint 2026/391).

## Out of scope

- R1CS reduction (paper Section 11): orthogonal, separate issue if/when Plonky3 grows R1CS.
- Distance-amplified codes (paper Section 12): orthogonal optimization.
- Holographic / preprocessed verifier: orthogonal.
- Malicious-verifier ZK: implied at the SNARG level via Fiat–Shamir; no extra work needed here.

## Notes

Mostly mechanical once #1586–#1588 land. The interesting bit is making sure the simulator composition in tests actually exercises Theorem 4.5 end-to-end — i.e., the test should construct the composed simulator from the per-round simulators rather than treating the protocol as a monolith.

tcoratger:
**Heads-up:** the WizardOfMenlo/whir reference hasn't composed the full protocol yet. PR #249 explicitly defers "integration into whir's round structure as an alternative to sumcheck-only folding". The `src/protocols/whir_zk/` module is a useful sketch, not a finished design.

[...]

Issue #1588, opened 2026-04-28
[HVZK-WHIR 5/6] HVZK base-case IOPP


tcoratger:
Part of the HVZK-WHIR effort tracked in #1590. Paper: [eprint 2026/391, Section 7 / Construction 7.2](https://eprint.iacr.org/2026/391.pdf). Depends on #1584, #1586.

## What

When the recursive code-switching reaches its terminal small instance, WHIR runs a non-succinct check. The HVZK base case (Construction 7.2) sends one full extra encoded mask `g = Enc(g̃, r')`, the verifier samples `γ`, and the prover opens `f* = g̃ + γ·f` and a few spot checks. The 2× overhead at the leaf is acceptable because the leaf is tiny (size sublinear in the original witness).

Lemma 7.3 covers HVZK; Lemma 7.4 covers RBR soundness.

## Where

`whir/src/pcs/base_case_zk.rs` (new), with separate `BaseCaseZkProver` / `BaseCaseZkVerifier` types. Reuse the existing base case for the `ZK = false` path.

## Acceptance criteria

- proptest: completeness on small parameters.
- proptest: simulator equivalence (Lemma 7.3) — for RS-based encodings the simulation error is 0.
- proptest: empirical RBR soundness within Lemma 7.4's bounds.

## Notes

Mask-size choice is per Section 2.7's footnote: `Czk` is a Reed–Solomon code with message length `O(λ / log log λ)` and rate `1/log λ`, giving masks of size `Õ(λ)` and `t_zk = O(λ / log log λ)`. Document the choice in code comments so a future reviewer can sanity-check it against the paper.

This is a guideline — the paper's parameter choices assume Reed–Solomon for `Czk`; if the implementation picks a different code via #1584, the parameters need re-deriving and that derivation should be in the PR description.

tcoratger:
**Reference implementation:** WizardOfMenlo/whir [PR #232](https://github.com/WizardOfMenlo/whir/pull/232) ("zkWHIRv3 part 1: Single commitment base case", merged).

- `src/protocols/basecase.rs` (322 lines) — direct implementation of Construction 7.2.
- `proptest-regressions/protocols/basecase.txt` shows test patterns.
- PR includes "Add unmasked mode to basecase algorithm" — the `ZK = false` regression path.

In the reference, this was the *first* PR of the series rather than the fifth. The Plonky3 order can mirror that — doing this before #1586 may make the dependency chain cleaner.

[...]

PR #1657, opened 2026-05-15
[ON HOLD] whir: add round-0 HVZK code-switching

### LINKED ISSUE #1587: [HVZK-WHIR 4/6] HVZK code-switching round

latifkasuli:
## Summary

This closes #1587 by implementing the scoped Construction 9.7 HVZK code-switching round from eprint 2026/391 §9.4.

Builds on #1605, the merged HVZK sumcheck work.

The issue described this as adding a ZK branch to the WHIR round. This PR uses the typed equivalent of that branch:

- `RoundConfig::zk: Option<RoundZkConfig<F>>`
- `WhirConfig::zk: Option<WhirZkConfig>`
- optional ZK proof payloads on the initial proof and round proof

`None` is the existing non-ZK path. `Some(...)` carries the derived Construction 9.7 parameters needed by the ZK branch, rather than using a bare `zk: bool` plus out-of-band config.

The PR is ~4.4k insertions against current `main`, with a substantial portion in tests and protocol-shape checks.

## Scope

Per the #1590 umbrella plan:

- #1587 — HVZK code-switching round, Section 9 / Construction 9.7 — this PR
- #1588 — HVZK base-case / final folded-polynomial IOPP
- #1589 — full `HidingWhirPcs` composition and end-to-end ZK PCS, Section 10 / Theorem 4

This PR implements the Construction 9.7 round-local boundary. Multi-round composition follows in #1589 per the umbrella decomposition.

The implementation is scoped to round 0. That is the round where the current Plonky3 WHIR source is the freshly committed folded target oracle, with `source.randomness_len == 0`. The helper and layout for that folded source are crate-local.

## What This Implements

This PR implements the first HVZK code-switching boundary:

- Fresh mask oracle  
  `s = Enc_{C_zk}((r, s_pad), r'')`

- Private OOD answers over the joint source/mask witness  
  `y = ze_ood(rho) · (f, r, s_pad)`

- Source openings bound to the committed folded target oracle.

- Mask oracle openings bound to the fresh mask commitment.

- Construction 9.7 batching:  
  `mu' = nu_1 * mu + sum_i nu_{1+i} * y_i + sum_j nu_{...+j} * f(x_j)`

- `CodeSwitchOutputRelation`, the verifier-side output relation passed to the next IOR.

- Nested HVZK sumcheck over the output relation.

- Typed #1605 sumcheck handoff carrying the residual claim, randomness, `eps`, mask commitments, and prover-side mask messages.

- ZK transcript/domain-separator shape for the initial HVZK sumcheck, round-0 source queries, round-0 mask queries, and nested HVZK sumcheck.

- Proof payloads:
  - `WhirInitialZkProof`
  - `WhirRoundZkProof`

Plain WHIR remains on the old path. The plain prover rejects ZK config, and the plain verifier rejects unexpected ZK payloads.

The typed `ZkSumcheckHandoff` and `ZkSumcheckData::ell_zk` field also address feedback from #1605: `eps` is explicit in the type system, and `ell_zk` is injected into the proof to close the wire-size shape ambiguity.

## Deferred Work

Left to the other HVZK-WHIR issues:

- #1588: HVZK base-case / final folded-polynomial IOPP.
- #1589: full `HidingWhirPcs` composition and end-to-end multi-round ZK opening flow.
- #1649: suffix-binding HVZK sumcheck support.

### AC1: Proof overhead

`RoundZkConfig::proof_field_overhead()` tracks the issue bound:

```text
m_zk * iota_zk + t_ood
```

where `m_zk * iota_zk` is the fresh mask oracle codeword size and `t_ood` is the private OOD answer count.

Covered by tests in `pcs::code_switch`.

### AC2: Honest prover acceptance

The branch includes a KoalaBear extension-field acceptance test for a real first-round ZK prefix opening:

- initial #1605 HVZK sumcheck
- round-0 Construction 9.7 code-switch
- source and mask Merkle openings
- nested HVZK sumcheck verification
- final masked residual handoff check

Covered by `zk_prefix_acceptance_tests`.

### AC3: Simulator boundary

The simulator tests cover the deterministic/programming boundary used by Lemma 9.8:

- the actual round-0 zero-randomness source shape, where source openings are deterministic from the plain folded WHIR target oracle
- private OOD answer programmability
- source-code opening programmability for ZK-encoded source views
- mask-code opening programmability
- witness-independent mask coupling under matched RNG
- consistency of the programmed `mu'` and output relation

The composed simulator coverage uses programmability plus matched-RNG mask coupling rather than full byte-equality of the simulated transcript. Construction 9.7’s source/target commitments are witness-dependent, and the current simulator surface exposes `ZkEncoding::simulate` for queried openings but not commitment-level Merkle simulation. For each simulator output, the test shows an honest witness can reproduce it via the Lemma 9.3 closed form. The witness-independent round mask is additionally coupled byte-equal under matched RNG, mirroring #1605’s simulator test at the components where that applies.

## Notes For Review

Round 0 has `source.randomness_len == 0` because Plonky3’s initial WHIR commitment is plain, not ZK-encoded. The OOD answer formula

```text
y = ze_ood(rho) · (f, r, s_pad)
```

therefore degenerates to:

```text
y = ze_ood(rho) · (f, s_pad)
```

for round 0. Lemma 9.3 privacy is carried by `s_pad >= 1`, i.e. the fresh mask alone in this degenerate plain-source case.

The current implementation does not force the final masked handoff through the existing plain WHIR final tail. After Construction 9.7, the handoff is still a masked output relation meant for the next ZK-aware IOR/base case.

## Reviewability

Happy to split this into smaller stacked PRs if preferred. I kept this branch as one coherent Construction 9.7 implementation because the algebra, typed sumcheck handoff, proof payload, verifier replay, and acceptance tests are tightly coupled.

## Verification

```text
cargo +stable sort --workspace --grouped --check
cargo +nightly fmt --all -- --check
cargo +stable clippy --all-targets -- -D warnings
cargo test
cargo test --features parallel
cargo test -p p3-whir pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact
cargo test -p p3-whir --features parallel pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact
git diff --check
```


whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,454 @@
...
+        value += inner_product(mask_message, &self.mask_covector);
+
+        Ok(value)
+    }
+}
+
+#[inline]
+fn inner_product<EF: Field>(a: &[EF], b: &[EF]) -> EF {
+    a.iter().zip(b).map(|(&x, &y)| x * y).sum()
+}
```
tcoratger: Tiny thing: we already have a generic `dot_product` in `p3_field` (see `field/src/helpers.rs`), and `adapter.rs` even imports it for the same purpose. I think we can drop this helper and just call `dot_product` directly — keeps the crate consistent and saves a name to chase.

whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,454 @@
...
+#[must_use]
+pub fn batching_coefficients<EF: Field>(rho: EF, dim: usize) -> Vec<EF> {
+    let mut coeffs = Vec::with_capacity(dim);
+    let mut power = EF::ONE;
+    for _ in 0..dim {
+        coeffs.push(power);
+        power *= rho;
+    }
+    coeffs
+}
```
tcoratger: This one is essentially `rho.powers().collect_n(dim)` — `Field::powers()` returns a `Powers` iterator and `collect_n` does the take-and-collect for you. Probably nicer to call those directly at the call sites instead of wrapping them in a named helper.

whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,454 @@
...
+
+/// Computes one private OOD answer from Construction 9.7.
+///
+/// This is `ze_ood(rho) · (f, r, s_pad)^T`, represented as a padded
+/// zero-evader evaluation of `(source_message || mask_message)`.
+#[inline]
+pub fn private_ood_answer<EF: Field>(rho: EF, source_message: &[EF], mask_message: &[EF]) -> EF {
+    padded_ood_t1(rho, source_message, mask_message)
+}
```
tcoratger: This is a pure pass-through to `padded_ood_t1` with the same signature. I think we can just call `padded_ood_t1` directly (including from `private_ood_answers` below) and remove the wrapper — fewer indirections.

whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,454 @@
...
+
+    Ok(claim.base_claim_coeff * inherited_claim + ood_sum + in_domain_sum)
+}
+
+/// Builds the output relation covectors for Construction 9.7.
+///
+/// `query_positions` are flattened codeword positions. For interleaving depth
+/// `iota`, callers should pass `iota * x_i + limb` for every queried limb.
+#[allow(clippy::too_many_arguments)]
+pub fn output_relation<F, EF, Enc>(
```
tcoratger: Couple of related thoughts here:

- The 8-argument signature is a bit heavy to read. It would feel nicer as a method on `ZkMaskClaim` (or maybe `CodeSwitchOutputRelation::build`), where the claim already carries most of these inputs and only the source-specific bits get passed in.
- This `LinearZkEncoding`-based version turns out to only be exercised by tests and by `simulated_verifier_view` (also test-only). The real prover/verifier path uses `code_switch_output_relation_from_rows` in `adapter.rs`, which is essentially the same identity but consuming rows. If we unify them behind a small row-provider abstraction (or even a closure), we avoid keeping two implementations of the same Construction 9.7 reduction in sync.

whir/src/pcs/adapter.rs
```diff
@@ -38,6 +56,1050 @@ where
...
+    let mut power = F::ONE;
+    for coeff in 0..coeffs_per_limb {
+        out[limb * coeffs_per_limb + coeff] = power;
+        power *= point;
+    }
+    out
+}
+
+#[allow(clippy::too_many_arguments)]
+fn code_switch_output_relation_from_rows<Row, EF>(
```
tcoratger: Tied to the comment over on `code_switch::output_relation` — these two functions compute the same Construction 9.7 reduction with slightly different row sources. Folding them into one place would save us from having to keep them aligned over time.

tcoratger (commented):
Hey @latifkasuli, thanks a lot for tackling this — the round-0 end-to-end acceptance test going through real Merkle openings is great to see, and a clear step up from earlier attempts in this space. The math-identity tests for `mu'` are also really nice.

I left some inline thoughts, mostly small Rust-quality things plus a couple of scope-shaped questions. None of them are individually blocking, but together they would shrink the diff a fair bit and bring it closer to what #1587 was framed around.

Two bigger-picture things worth chatting about:

1. The PR comes in at around 4.4k lines for what #1587 described as "two local changes to a round". A lot of that size sits in the parallel ZK API (`begin_zk_prefix_open` / `round0_zk_prefix` / `verify_round_zk_prefix`) running alongside the plain `MultilinearPcs::open` path, plus a few helper functions that overlap with things we already have in `p3_field`. Would you be open to a smaller version that adds the two local changes inside the existing round loop, gated by an `Option<RoundZkConfig>`? It would map more directly to the issue and would be easier to land incrementally.

2. The "round-0 only" scope (with `source.randomness_len == 0`) is doing a lot of structural work in this design. Worth pulling that into the PR title or a top-of-file doc comment so the constraint isn't surprising for whoever touches this next.

Happy to dig into any of these. Thanks again for the work!

whir/src/pcs/adapter.rs
```diff
@@ -38,6 +56,1050 @@ where
...
+    }
+
+    CodeSwitchOutputRelation {
+        source_covector: source_covector_out,
+        auxiliary_covectors: auxiliary_covectors_out,
+        mask_covector,
+    }
+}
+
+pub(crate) fn evaluate_zk_mask_residual<F, EF>(masks: &[Vec<F>], gammas: &[EF]) -> EF
```
tcoratger: This reproduces the residual recurrence from #1605's HVZK sumcheck. If anything inside the sumcheck module changes its mask handling, this would silently get out of sync. Could we expose the residual through `ZkSumcheckData` / `ZkPrefixProver` accessors and call into that, rather than re-deriving the algebra here? Should also simplify the next function (`zk_mask_residual_covectors_from_shape`) — see below.

whir/src/pcs/adapter.rs
```diff
@@ -38,6 +56,1050 @@ where
...
+{
+    assert!(
+        masks
+            .iter()
+            .all(|mask| mask.len() == masks.first().map_or(0, Vec::len))
+    );
+    zk_mask_residual_covectors_from_shape(masks.len(), masks.first().map_or(0, Vec::len), gammas)
+}
+
+fn zk_mask_residual_covectors_from_shape<F, EF>(
```
tcoratger: This builds the covectors by probing: setting each basis vector to `ONE`, evaluating, and reading the result column by column. Since the residual is linear in the masks there is a closed form for these covectors that should be much cheaper — the current shape is roughly `O(k^3 · mask_len^2)` residual evaluations. If we get the covectors directly out of the sumcheck module (per the comment above), this function probably collapses to a few lines.

whir/src/pcs/adapter.rs
```diff
@@ -38,6 +56,1050 @@ where
...
+        let folded_evaluations = Poly::new(source.message.clone());
+        let num_variables = folded_evaluations.num_variables();
+        let folding_factor_next = source.folding_factor;
+        assert_eq!(source.domain_size % (1usize << num_variables), 0);
+        let inv_rate = source.domain_size >> num_variables;
+        assert!(inv_rate > 0, "source domain must cover the source message");
+        assert_eq!(
+            source.randomness_len, 0,
+            "ZK code-switch source randomness needs a source-oracle randomness handoff",
+        );
```
tcoratger, line 441: Worth a slightly louder comment here: this assert means the round-0 path only works when the freshly committed folded WHIR oracle has no encoding randomness. That's fine today since `commit_extension` is plain, but it does mean this code path doesn't generalize to later rounds without a ZK-aware target commit being added elsewhere. Could you add a short pointer (e.g. to #1589) so a future reader knows this restriction is by design rather than an oversight?

whir/src/pcs/adapter.rs
```diff
@@ -38,6 +56,1050 @@ where
...
+
+        let (target_root, target_merkle_data) = commit_extension(
+            crate::sumcheck::strategy::VariableOrder::Prefix,
+            &self.dft,
+            &self.extension_mmcs,
+            &folded_evaluations,
+            folding_factor_next,
+            inv_rate,
+        );
```
tcoratger, line 479: Quick observation worth a code comment: the target oracle `g` here is committed via plain `commit_extension`, without the encoding randomness `r'` that Construction 9.7 step 1 calls for. The PR description's argument that Lemma 9.3 carries privacy via `s_pad` is solid for the OOD answer specifically, but the in-domain spot checks `f(x_i)` are still deterministic in the witness, so the mask oracle doesn't hide them. If we ever want spot checks to hide too, we'd need a `commit_extension_zk` variant for the target. Probably fine to defer the implementation, but worth flagging in a comment here so the next reader doesn't think the round-0 privacy story is fully closed.

whir/src/pcs/prover/mod.rs
```diff
@@ -152,6 +152,10 @@ where
         Dft: TwoAdicSubgroupDft<F>,
         Challenger: CanObserve<MT::Commitment>,
     {
+        assert!(
+            self.config.zk.is_none(),
+            "ZK WHIR config requires the prefix-only ZK prover path"
+        );
```
tcoratger, line 158: The runtime assert here is a small smell — when we end up with two prover paths with different semantics, encoding the choice in the type system (a separate `WhirZkProver` wrapper, or a marker type) usually reads cleaner and removes the runtime panic branch. Not blocking, just something to consider as the ZK side matures.

whir/src/pcs/code_switch/tests.rs
```diff
@@ -0,0 +1,1123 @@
...
+
+    #[test]
+    fn prop_composed_simulator_components_program_code_switch_view(seed in 0_u64..64) {
+        prop_assume!(assert_composed_simulator_components_program_code_switch_view(
+            seed.wrapping_add(101),
+        ));
+    }
+}
+
+fn assert_composed_simulator_components_program_code_switch_view(seed: u64) -> bool {
```
tcoratger: Comparing this against AC3 from #1587: the test is checking programmability — given a simulated source opening from `source_enc.simulate(...)`, it solves backwards for honest randomness that would produce that opening. That works because the source encoding here has `randomness_len = 2`. On the actual prover path in `round0_zk_prefix_from_folded_source`, the source has `randomness_len = 0`, so a real "simulator matches the round-0 view" test would look pretty different.

The test is genuinely useful for the math identity it covers, but the name reads as a stronger claim than is being made. Could you either (a) run an equivalent test against the actual zero-randomness target and document why that case still satisfies Theorem 4.5, or (b) tone down the name and call out in the PR description that AC3 is still open? Either is helpful for whoever picks up the composition piece next.

latifkasuli:
Thanks Thomas, this makes sense. I’ll first shrink the obvious duplicated/helper surface and make the round-0/plain-source constraint much louder in the code/docs. I’ll also tighten the simulator/AC3 framing so it doesn’t overclaim relative to the actual round-0 zero-randomness path.\n\nFor the larger shape question, I’m open to an in-loop version if that’s the preferred landing path. I’ll do the focused cleanup first so we can see how much of the review burden drops before deciding whether a deeper rewrite is still needed.

whir/src/fiat_shamir/domain_separator.rs
```diff
@@ -24,6 +24,118 @@ pub(crate) struct SumcheckParams {
     pub pow_bits: usize,
 }
 
+#[cfg(test)]
```
tcoratger: This should be at the end of the file no?

tcoratger:
I've done a second pass and I think that, as it stands, the PR is too large for a proper review. I think it needs to be split into multiple smaller PRs to isolate the changes.

For example, I would suggest submitting a separate, smaller PR with only the Fiat Shamir changes.

Then I see there are several changes on the ZK sumcheck; it might be worth separating them into several smaller PRs for each change to isolate them as much as possible and make the review easier.

We can then go back to this one once all the smaller ones have been merged in order to minimize the line changes on this one.

What do you think?

latifkasuli:
Yes, that makes sense. I agree this is still too large to review comfortably.

I’ll split it into smaller stacked PRs. Proposed order:

1. Fiat-Shamir/domain-separator shape for the ZK transcript.
2. ZK sumcheck handoff/accessor changes needed by code-switching.
3. Standalone Construction 9.7 code-switch algebra and tests.
4. Round-0 WHIR adapter wiring using the pieces above.

I’ll also move the domain separator tests to the bottom of the file as you noted.

That should make each layer reviewable independently while keeping this branch as the integration reference.

tcoratger:
Just a note: we should probably merge https://github.com/Plonky3/Plonky3/pull/1665 before this one.

Nashtare:
Hi @latifkasuli! Now that #1665 has been merged, would you mind clearing up all the conflicts so we can go a final round of reviews on this branch? Thanks 🙏🏼

latifkasuli:
> Hi @latifkasuli! Now that #1665 has been merged, would you mind clearing up all the conflicts so we can go a final round of reviews on this branch? Thanks 🙏🏼

I’ll rebase/clear the conflicts now that #1665 has landed.

Just to align with @tcoratger’s earlier feedback: I’ve been splitting this into smaller PRs so the pieces can be reviewed independently ([#1662](https://github.com/Plonky3/Plonky3/pull/1662) has merged, and the next ZK sumcheck handoff/accessor PR is prepared locally). I’ll still update this branch so it stays useful as the integration branch / final review target, but I’m happy to keep landing the smaller pieces first if that remains the preferred path.

Nashtare:
Yes ideally keeping PRs sub 2000 LoC would be nice, we can keep this as the main reference :)

tcoratger:
Superseded by https://github.com/Plonky3/Plonky3/pull/1767

PR #1664, opened 2026-05-16
perf(multilinear-util): field-aware MLE recursion threshold

wstran:
## Summary

`Poly::eval_base` and `Poly::eval_ext` dispatch between a scalar recursive evaluator (small `n`) and the SIMD-packed `SplitEq` path (larger `n`) using a fixed `MLE_RECURSION_THRESHOLD = 10`. A new `eval_threshold` bench shows the crossover is base-field-size-dependent — 4-byte bases (BabyBear, KoalaBear, Mersenne31) cross at `n = 9`, while 8-byte bases (Goldilocks) cross at `n = 10`. Replacing the constant with a `const fn` keyed on `core::mem::size_of::<BaseField>()` gives a clean win on small-base fields with no regression elsewhere.

I also drop the unreachable `join` branch inside `eval_multilinear_recursive`. Callers gate that function with `num_variables < threshold`, so `evals.len() <= 512` (`< 2^9`), while the gate that selected `join` requires `evals.len() > work_size = (1 << 15) / size_of::<F>() >= 1024` (even BN254's 32-byte field gives 1024). Removing it leaves a comment explaining why the recurse is sequential.

The mathematical contract is unchanged: both implementations compute `f(point) = Σ_x eq(x, point) · f(x)`, and the existing `test_eval_base_field` / `test_eval_ext_field` checks `eval_reference` agreement for `k = 1..=20`, straddling the threshold.

## Bench

`cargo bench -p p3-multilinear-util --bench eval_threshold`, M1, BabyBear + `BinomialExtensionField<BabyBear, 4>` and Goldilocks + `BinomialExtensionField<Goldilocks, 2>`, criterion 60 samples, median (low-high in brackets).

| n  | BabyBear before (poly_eval_ext) | BabyBear after  | Δ      | Goldilocks before | Goldilocks after | Δ |
|----|--------------------------------:|----------------:|-------:|------------------:|-----------------:|--:|
| 6  | 346 ns                          | 342 ns          | -1%    | 230 ns            | 217 ns           | -6% |
| 7  | 714 ns                          | 711 ns          | tied   | 460 ns            | 460 ns           | tied |
| 8  | 1.45 µs                         | 1.44 µs         | tied   | 911 ns            | 911 ns           | tied |
| 9  | 2.91 µs                         | **2.47 µs**     | **-15.1%** | 1.85 µs       | 1.80 µs          | -3% |
| 10 | 4.63 µs                         | 4.68 µs         | tied   | 4.83 µs           | 4.55 µs          | tied |
| 11 | 8.12 µs                         | 8.10 µs         | tied   | 8.49 µs           | 8.48 µs          | tied |
| 12-14 | ...                          | ...             | tied   | ...               | ...              | tied |

Only the BabyBear `n = 9` cell moves materially: that's the case where the constant threshold forced `recursive` even though `SplitEq` is already faster. KoalaBear and Mersenne31 are 4-byte bases too, so they pick up the same threshold = 9 path.

The small low-`n` improvements on both fields come from removing the `join` branch's setup work at every recursion frame.

## Test plan

- [x] `cargo test -p p3-multilinear-util --lib` — 131 passed.
- [x] `cargo test -p p3-whir --lib` — 201 passed, including `test_whir_end_to_end`.
- [x] `cargo test --workspace --lib` — all green.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `cargo build --target thumbv7em-none-eabi -p p3-multilinear-util` — clean.

### DIFF
--- a/multilinear-util/src/poly.rs
+++ b/multilinear-util/src/poly.rs
@@ -17,8 +17,24 @@ use crate::split_eq::SplitEq;
 
 pub(crate) const PARALLEL_THRESHOLD: usize = 4096;
 
-/// Number of variables at which we switch from recursive to chunk-based MLE evaluation.
-const MLE_RECURSION_THRESHOLD: usize = 10;
+/// Number of variables at which we switch from recursive scalar evaluation to the
+/// SIMD-packed `SplitEq` path.
+///
+/// Crossover depends on the base-field byte width (smaller base ⇒ wider packing
+/// ⇒ `SplitEq` wins earlier), measured on aarch64 (NEON) and x86-64 (AVX2):
+///
+/// - 4-byte bases (BabyBear, KoalaBear, Mersenne31): `SplitEq` wins from n=9.
+/// - 8-byte bases (Goldilocks): recursive still wins at n=9, crosses at n=10.
+///
+/// Wider bases (e.g. BN254) keep the default 10.
+#[inline]
+const fn mle_recursion_threshold<B>() -> usize {
+    if core::mem::size_of::<B>() <= 4 {
+        9
+    } else {
+        10
+    }
+}
 
 /// Returns a vector of uninitialized elements of type `A` with the specified length.
 ///
@@ -315,7 +331,7 @@ impl<F: Field> Poly<F> {
     #[must_use]
     #[inline]
     pub fn eval_base<EF: ExtensionField<F>>(&self, point: &Point<EF>) -> EF {
-        if point.num_variables() < MLE_RECURSION_THRESHOLD {
+        if point.num_variables() < mle_recursion_threshold::<F>() {
             eval_multilinear_recursive(&self.0, point.as_slice())
         } else {
             SplitEq::new_packed(point, EF::ONE).eval_base(self)
@@ -338,7 +354,7 @@ impl<F: Field> Poly<F> {
     where
         F: ExtensionField<BaseField>,
     {
-        if point.num_variables() < MLE_RECURSION_THRESHOLD {
+        if point.num_variables() < mle_recursion_threshold::<BaseField>() {
             eval_multilinear_recursive(&self.0, point.as_slice())
         } else {
             SplitEq::new_packed(point, F::ONE).eval_ext(self)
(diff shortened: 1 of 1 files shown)


multilinear-util/src/poly.rs
```diff
@@ -703,24 +720,12 @@ where
...
-                        eval_multilinear_recursive(f0, sub_point),
-                        eval_multilinear_recursive(f1, sub_point),
-                    )
-                }
-            };
+            // Sequential recurse: callers gate this function with `num_variables <
+            // MLE_RECURSION_THRESHOLD`, so `evals.len()` is always small enough that
+            // Rayon `join` overhead would dominate.
+            let f0_eval = eval_multilinear_recursive(f0, sub_point);
+            let f1_eval = eval_multilinear_recursive(f1, sub_point);
```
tcoratger, line 726: Here, it seems to me that we removed the parallelization with rayon because you said that since we're operating on small dimensions, the overhead of rayon is always higher in this precise case.

But are you sure that the speedup doesn't come solely from this change, and that the new `mle_recursion_threshold` method would therefore be unnecessary?

It would probably be worthwhile to run the same benchmarks with only this change in the pull request and nothing else, that is, leaving `MLE_RECURSION_THRESHOLD` as it was before.

wstran: Good question — re-ran the exact A/B you asked for. M1 (aarch64), criterion 60 samples, all three configs run back-to-back in the same session.

| Config                                                       | BabyBear n=9 | Goldilocks n=9 |
| ------------------------------------------------------------ | ------------:| --------------:|
| A: main baseline (recursive + `join` branch, threshold = 10) | 2.92 µs      | 1.79 µs        |
| B: dead-branch removed, threshold still 10                   | 2.87 µs      | 1.79 µs        |
| C: this PR (dead-branch + field-aware threshold)             | 2.48 µs      | 1.79 µs        |

A → B: −1.7% (within variance) — branch removal alone doesn't move n=9.
B → C: −13.6% — the threshold shift to 9 for 4-byte bases is the actual win.
Goldilocks stays on recursive at n=9 in all three configs (8-byte base keeps threshold = 10), so no regression there.

Branch removal does still give a small consistent win at n=6..8 on both fields (3–8% from fewer per-frame ops) and resolves the unreachable code, but at n=9 it's the threshold change that does the work. Equivalence at n=9 is covered by the existing `test_eval_base_field` / `test_eval_ext_field` loops (k=1..=20) and the WHIR end-to-end test, which exercise both dispatch paths against the naive `eval_reference`.

Happy to split into two PRs (threshold tuning + cleanup) if you'd rather land them separately, and happy to repeat the A/B on x86 if useful — I only have an M1 locally.

wstran: Good question — re-ran the exact A/B you asked for, and also reproduced on x86 Haswell with AVX2 enabled (`-C target-cpu=native`) to confirm the threshold isn't M1-specific. Criterion 60 samples each, all three configs run back-to-back per machine.

BabyBear `Poly::eval_ext::<BabyBear>` at n=9:

| Config                                              | M1 NEON | x86 Haswell+AVX2 |
| --------------------------------------------------- | -------:| ----------------:|
| A: main baseline (recursive + `join`, thr = 10)     | 2.92 µs | 6.16 µs          |
| B: dead-branch removed, threshold still 10          | 2.87 µs | 6.59 µs          |
| C: this PR (dead-branch + field-aware threshold)    | 2.48 µs | 4.99 µs          |
| A → C                                               |  -15.1% |  -19.0%          |

Branch removal alone doesn't move n=9 on either arch — A → B falls inside the run-to-run variance the Goldilocks control shows (~1% on M1, ~3% on Haswell). The win is the threshold shift forcing `SplitEq` at n=9 for 4-byte bases, and AVX2's wider packing (W=8 vs NEON's 4) actually makes that `SplitEq` path even more favourable on x86.

Goldilocks (8-byte base, threshold stays 10) is unchanged across all six runs and stays on the recursive path at n=9 on both archs, so no regression there.

Branch removal does give a small consistent win at n=6..8 from fewer per-frame ops, and resolves the unreachable code. Equivalence at n=9 is covered by `test_eval_base_field` / `test_eval_ext_field` for k=1..=20 and by WHIR's end-to-end test against the naive `eval_reference`.

Happy to split into two PRs if you'd prefer the threshold tuning and the branch removal landed separately.

tcoratger: This is fine to me to merge this as is without splitting. Small question: is it absolutely needed to add a new benchmark file because this adds lines of code and possible confusion, or is it possible to use the existing one to measure things here?

wstran: Makes sense — dropped the new benchmark file. The threshold was a one-time tuning decision, so the standalone bench isn't worth the extra surface; the A/B numbers above are enough to document the choice. The PR is now just the `poly.rs` change (the `const fn` threshold + removing the unreachable `join` branch). Cargo.toml is back to its original state.

tcoratger (approved):

PR #1665, opened 2026-05-16
feat(whir): HVZK sumcheck suffix-binding prover

### LINKED ISSUE #1649: [HVZK-WHIR 3b/7] HVZK sumcheck overlay for the suffix-binding layout

tcoratger:
## Summary

Resolves #1649. Builds on [#1605](https://github.com/Plonky3/Plonky3/pull/1605) (HVZK sumcheck with sublinear masks) by adding a suffix-binding HVZK prover alongside the existing prefix-binding one, so callers can choose the binding direction in the zero-knowledge path the same way they already can in classical WHIR.

The masking layer of Construction 6.3 (eprint 2026/391) is binding-agnostic — the mask schedule, the auxiliary target, and the per-round wire polynomial are identical for both binding modes. Only the residual handoff at the end of the per-round loop branches on the binding direction:

```text
    mode    | residual factor
    --------+--------------------------------------
    prefix  | low-to-high fold, packed product
    suffix  | high-to-low fold, unpacked product
```

The code reflects that: one generic ``ZkProver<…, L: ZkLayout>`` with a single ``into_sumcheck`` implementation; the residual handoff lives behind a four-method ``ZkLayout`` sub-trait that both ``PrefixProver`` and ``SuffixProver`` implement.

## What changed

- **``ZkLayout`` sub-trait** (``whir/src/sumcheck/zk/prover/layout.rs``) — extends the existing ``Layout`` trait with ``concrete_claims``, ``virtual_claims``, ``batched_sum``, and ``zk_residual_handoff``. The handoff is the only operation that genuinely differs between the two binding modes.

- **Unified ``ZkProver``** (``whir/src/sumcheck/zk/prover/zk_prover.rs``) — single generic prover, with ``ZkPrefixProver`` and ``ZkSuffixProver`` type aliases preserving the existing public API.

- **Per-round assembly factored out** (``whir/src/sumcheck/zk/prover/round.rs``) — ``RoundContext::assemble`` takes three named struct arguments (``RoundContext``, ``RoundState``, ``PlainPiece``) instead of twelve positional ones.

- **Prelude helpers** (``whir/src/sumcheck/zk/prover/common.rs``) — ``sample_masks`` and ``observe_masks_and_mu_tilde`` extracted from the per-mode files.

- **``ZkEncoding::sample_message``** — added to ``p3-zk-codes`` so the encoding owns its message-space sampling. The Reed-Solomon encoding draws uniformly over ``F^message_len``; the HVZK prover and simulator both route through it.

- **Verifier constructors** — ``ZkVerifier::new_prefix`` / ``new_suffix`` replace the single ``new``, surfacing the binding mode at the construction site.

- **Per-mode test coverage** — honest roundtrips, simulator view-matching, PoW tampering, ``ell_zk`` mismatch, and round-by-round wire tampering are each parameterised by a ``Mode`` enum and run once per binding mode.

- **New focused unit tests** on ``common.rs`` and ``round.rs`` for the extracted helpers (mask sampling determinism, ``mu_tilde`` closed-form, wire format, per-round affine identity, etc.).

## Test plan

- [x] ``cargo test -p p3-whir --lib sumcheck::zk`` — 27 tests pass (16 integration + 11 new isolated unit tests).
- [x] ``cargo test -p p3-zk-codes --lib`` — 16 tests pass.
- [x] ``cargo clippy -p p3-whir -p p3-zk-codes --lib --tests`` — clean.
- [x] ``cargo fmt --all -- --check`` — clean.
- [x] No public API breakage — ``ZkPrefixProver`` continues to compile through the type alias; existing call sites are unchanged.


sumcheck/src/zk/prover/zk_prover.rs
```diff
@@ -339,74 +294,31 @@ where
             //     plain_h(0) + plain_h(1) = plain_sum
             //     => c_1 = plain_sum - 2 c_0 - c_inf
             let plain_c1 = plain_sum - plain_c0.double() - plain_c_inf;
+            // s_j_endpoints is consumed by the assembly helper via the mask
+            // vector; tag it as used so the decrement above stays linked to
+            // its derivation.
+            let _ = s_j_endpoints;
```
camofu: Not sure this is actually correct, `s_j_endpoints` gets consumed a few lines up:

```rust
// Update the running future-endpoint sum: drop s_j's contribution so the round-j formula reads only sum_{l > j}.
let s_j_endpoints = s_j[0].double() + s_j[1..].iter().copied().sum::<F>();
sum_future_endpoints -= s_j_endpoints;
```
(really minor non-blocker anyways)

latifkasuli:
I did a pass on this with [#1649](https://github.com/Plonky3/Plonky3/issues/1649), and [#1605,](https://github.com/Plonky3/Plonky3/pull/1605) in mind.

Overall, the design looks right to me. The `ZkLayout` factoring is clean: the masking loop is shared, and the real prefix/suffix difference is isolated to verifier strategy plus the residual handoff. The suffix handoff via `compress_stacked(&reversed, eps)` also seems like the right tradeoff, since it absorbs `eps` without a second scaling pass.

A few things I noticed:

1. This is currently conflicting with `main`.

The only hard conflict I saw is mechanical in `zk-codes/src/reed_solomon.rs`: `main` added `query_bound()`, while this PR added `sample_message()`. The resolved impl should keep both methods.

2. [#1649’s](https://github.com/Plonky3/Plonky3/issues/1649) benchmark acceptance criterion is not fully covered yet.

The issue asks for a `zk_suffix` row in `whir/layout_sumcheck`, but the current bench still only builds the prefix ZK path via `ZkPrefixProver`, and the benchmark row is still labelled just `zk`.

I think this should either become:

- `zk_prefix`
- `zk_suffix`

or keep the existing `zk` prefix row and add a new `zk_suffix` row. That would make the suffix prover cost visible next to plain prefix/suffix and prefix HVZK.

3. `cargo bench -p p3-whir --bench sumcheck --no-run` compiles but emits a warning.

`PlainPiece::sum` is only read inside the debug-only affine consistency block, so release/bench builds warn that the field is never read. Probably easiest to either gate the field with `#[cfg(debug_assertions)]` or restructure the debug check so the field does not exist in release builds.

One protocol-level question, not necessarily a blocker:

The simulator samples `wire[0..2]` as full `EF`, while the honest prover forms those coordinates as base-field mask terms plus `eps * plain`, with `eps` already known before the wire is sent. I understand this follows the hybrid F/EF shape from #1605, but since this PR extends the path to suffix mode too, I think it is worth double-checking that the Lemma 6.4 simulator argument still gives the intended distribution in this hybrid setting. The existing verifier-acceptance and `wire[2..]` F-subspace tests are useful, but I wanted to call out the distribution point explicitly.

sumcheck/src/zk/prover/common.rs
```diff
@@ -0,0 +1,284 @@
...
+    mmcs: &M,
+    challenger: &mut Ch,
+    rng: &mut R,
+) -> (Vec<Vec<F>>, Vec<MaskOracle<F, Enc, M>>)
+where
+    F: Field,
+    Enc: ZkEncoding<F>,
+    Enc::Codeword: Matrix<F>,
+    M: Mmcs<F>,
+    Ch: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
```
camofu: I think this might be over-constrained. We only need `CanObserve`

sumcheck/src/zk/prover/common.rs
```diff
@@ -0,0 +1,284 @@
...
+    //
+    //     s_l(0)        = c_0
+    //     s_l(0)+s_l(1) = 2 * c_0 + sum_{i>=1} c_i
+    let sum_endpoints_init: F = masks
+        .iter()
+        .map(|mask| mask[0].double() + mask[1..].iter().copied().sum::<F>())
+        .sum();
+
+    // Leading coefficient of the closed form.
+    let two_to_k_minus_1 = F::TWO.exp_u64((k - 1) as u64);
```
camofu: Maybe a `debug_assert!(k>=1, ...)` could be useful here

sumcheck/src/zk/prover/round.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// # Round formula
+    ///
+    /// ```text
+    ///     h_j(X) = 2^{k-j}     * s_j(X)
+    ///            + 2^{k-j}     * sum_{l <  j}  s_l(g_l)
+    ///            + 2^{k-j-1}   * sum_{l >  j} ( s_l(0) + s_l(1) )    zero at j = k
+    ///            + e           * plain_j(X)
+    /// ```
+    ///
+    /// Encoded in the basis `(c_0, c_1, c_2)`.
```
camofu: Maybe `(c_0, c_1, c_inf)`?

sumcheck/src/zk/prover/round.rs
```diff
@@ -0,0 +1,466 @@
...
+            let mult_future = self.pow2[self.k - state.j - 1];
+            h[0] += mult_future * state.future_endpoints;
+        }
+
+        // Plain piece, scaled by the combining challenge.
+        //
+        // Only the linear-coefficient slot stays in the base-field subspace once the verifier rederives it.
+        // The simulator relies on that stratification.
+        h[0] += self.eps * plain.c0;
+        h[1] += self.eps * plain.c1;
```
camofu: I think this h[1] write is dead in release. Mainly because the verifier re-construct `c_1` from the affine identity `h(0)+h(1)=target` so the prover value is never read in release.

camofu (approved):

tcoratger:
> I did a pass on this with [#1649](https://github.com/Plonky3/Plonky3/issues/1649), and [#1605,](https://github.com/Plonky3/Plonky3/pull/1605) in mind.
> 
> Overall, the design looks right to me. The `ZkLayout` factoring is clean: the masking loop is shared, and the real prefix/suffix difference is isolated to verifier strategy plus the residual handoff. The suffix handoff via `compress_stacked(&reversed, eps)` also seems like the right tradeoff, since it absorbs `eps` without a second scaling pass.
> 
> A few things I noticed:
> 
> 1. This is currently conflicting with `main`.
> 
> The only hard conflict I saw is mechanical in `zk-codes/src/reed_solomon.rs`: `main` added `query_bound()`, while this PR added `sample_message()`. The resolved impl should keep both methods.
> 
> 2. [#1649’s](https://github.com/Plonky3/Plonky3/issues/1649) benchmark acceptance criterion is not fully covered yet.
> 
> The issue asks for a `zk_suffix` row in `whir/layout_sumcheck`, but the current bench still only builds the prefix ZK path via `ZkPrefixProver`, and the benchmark row is still labelled just `zk`.
> 
> I think this should either become:
> 
> * `zk_prefix`
> * `zk_suffix`
> 
> or keep the existing `zk` prefix row and add a new `zk_suffix` row. That would make the suffix prover cost visible next to plain prefix/suffix and prefix HVZK.
> 
> 3. `cargo bench -p p3-whir --bench sumcheck --no-run` compiles but emits a warning.
> 
> `PlainPiece::sum` is only read inside the debug-only affine consistency block, so release/bench builds warn that the field is never read. Probably easiest to either gate the field with `#[cfg(debug_assertions)]` or restructure the debug check so the field does not exist in release builds.
> 
> One protocol-level question, not necessarily a blocker:
> 
> The simulator samples `wire[0..2]` as full `EF`, while the honest prover forms those coordinates as base-field mask terms plus `eps * plain`, with `eps` already known before the wire is sent. I understand this follows the hybrid F/EF shape from #1605, but since this PR extends the path to suffix mode too, I think it is worth double-checking that the Lemma 6.4 simulator argument still gives the intended distribution in this hybrid setting. The existing verifier-acceptance and `wire[2..]` F-subspace tests are useful, but I wanted to call out the distribution point explicitly.

Thanks a lot for the review @latifkasuli, just fixed all the comments!

sumcheck/src/zk/prover/zk_prover.rs
tcoratger: Fixed

sumcheck/src/zk/prover/common.rs
tcoratger: Fixed

sumcheck/src/zk/prover/common.rs
tcoratger: Fixed

sumcheck/src/zk/prover/round.rs
tcoratger: Fixed

sumcheck/src/zk/prover/round.rs
tcoratger: Yes fixed

sumcheck/src/zk/verifier.rs
```diff
@@ -233,300 +258,393 @@ where
...
+    /// - Replays the verifier from the same post-prover state on both
+    ///   transcripts.
+    /// - Asserts the two final targets diverge.
+    ///
+    /// # Why local checks are not enough
+    ///
+    /// - The affine reconstruction of `c_1` forces the per-round identity
+    ///   to hold on a tampered wire.
+    /// - The verifier therefore does not reject locally.
+    /// - Round-by-round soundness still needs the cheat caught.
```
Nashtare, line 494: This may be worth putting in the public documentation of `into_sumcheck`? Otherwise it may get ignored

sumcheck/src/layout/prover/suffix.rs
```diff
@@ -487,22 +488,35 @@ impl<F: TwoAdicField, EF: ExtensionField<F>> SuffixProver<F, EF> {
...
+    /// - `scale` — extra factor folded into the equality table.
+    ///
+    /// # Why a scale parameter
+    ///
+    /// - A unit scale leaves the residual sum unchanged.
+    /// - A non-unit scale lets the caller absorb a combining challenge into
+    ///   the residual factor without a second pass.
     #[tracing::instrument(skip_all)]
-    fn compress_stacked(&self, rs: &Point<EF>) -> Poly<EF> {
+    pub(crate) fn compress_stacked(&self, rs: &Point<EF>, scale: EF) -> Poly<EF> {
```
Nashtare: I'd be careful here, as we don't want to pass a `EF::ZERO` scale. We could either just add a hard assert, or have a `compress_stacked` as before (i.e. no scale parameter), and calling under the hood `compress_stacked_scaled` with extra `EF::ONE` argument?

sumcheck/src/zk/prover/layout.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    fn batched_sum(&self, alpha: EF) -> EF {
+        self.sum(alpha)
+    }
+
+    fn zk_residual_handoff(self, rs: &Point<EF>, alpha: EF, eps: EF) -> ProductPolynomial<F, EF>
+    where
+        EF: TwoAdicField,
+    {
```
Nashtare, line 63: Shouldn't we assert here that `k_pack < n_vars - k`? That's what the proptests imply, but we only do 
```rust
        assert!(
            k <= n_vars,
            "folding_factor must be <= poly.num_variables()",
        );
```
in zk_prover.rs

Nashtare:
@tcoratger Once this gets updated with latest `main` and my relatively minor comments are addressed, I think we'll be good to merge

sumcheck/src/zk/verifier.rs
tcoratger: Yep added

sumcheck/src/layout/prover/suffix.rs
tcoratger: Yes good idea, just pushed this

sumcheck/src/zk/prover/layout.rs
tcoratger: Done :)

tcoratger:
> @tcoratger Once this gets updated with latest `main` and my relatively minor comments are addressed, I think we'll be good to merge

Done :)

sumcheck/src/layout/prover/suffix.rs
```diff
@@ -505,11 +514,12 @@ impl<F: TwoAdicField, EF: ExtensionField<F>> SuffixProver<F, EF> {
     ///
     /// # Why a scale parameter
     ///
-    /// - A unit scale leaves the residual sum unchanged.
-    /// - A non-unit scale lets the caller absorb a combining challenge into
+    /// - A non-unit `scale` lets the caller absorb a combining challenge into
     ///   the residual factor without a second pass.
+    /// - The unit-scale case has the dedicated [`Self::compress_stacked`] wrapper,
+    ///   so callers never pass `EF::ONE` (nor an accidental `EF::ZERO`) explicitly.
```
Nashtare: well callers could still be passing `EF::ONE` or `EF::ZERO` here, there's nothing to prevent it inside so the comment is a bit misleading

Nashtare (approved):

sumcheck/src/layout/prover/suffix.rs
tcoratger: Ok yes, I've removed this misleading comment so that no confusion and I've just added a simple assertion inside the function to prevent the zero case

PR #1666, opened 2026-05-17
verifier: add a couple strengthening checks

Nashtare:
## Summary

Add a couple checks / tweaks to strengthen the verifier flow, namely:

* reject too wide circle opening claims
* bound FRI folding log_arity
* derive verify_batch arity schedule from dimensions
* reject proof degree_bits exceeding PCS max domain size
* clarify comment around quotient-degree hint (no concrete changes, just clarifying the doc around the debug_assertion)

### DIFF
--- a/fri/src/hiding_pcs.rs
+++ b/fri/src/hiding_pcs.rs
@@ -98,6 +98,11 @@ where
             &self.inner, degree)
     }
 
+    fn log_max_lde_height(&self) -> usize {
+        <TwoAdicFriPcs<Val, Dft, InputMmcs, FriMmcs> as Pcs<Challenge, Challenger>>::log_max_lde_height(
+            &self.inner)
+    }
+
     fn commit(
         &self,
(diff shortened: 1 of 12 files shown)


fri/src/hiding_pcs.rs
```diff
@@ -98,6 +98,11 @@ where
             &self.inner, degree)
     }
 
+    fn log_max_lde_height(&self) -> usize {
+        <TwoAdicFriPcs<Val, Dft, InputMmcs, FriMmcs> as Pcs<Challenge, Challenger>>::log_max_lde_height(
+            &self.inner)
```
tcoratger, line 103: Doesn't this work? Maybe not, not sure

```suggestion
        self.inner.log_max_lde_height()
```

tcoratger (approved):
Just a super minor comment, not a blocker for merging :)

fri/src/hiding_pcs.rs
Nashtare: compiler complains to use fully-qualified path

PR #1667, opened 2026-05-17
perf(goldilocks): cross-permute batched Poseidon2 for widths 8 and 12

Niek-Kamer:
## Summary

Adds a hand-rolled cross-permute batched Poseidon2 permute for Goldilocks widths 8 and 12 (`permute_batch_b2_p2_w{8,12}`) that processes two independent states with their rounds interleaved in a single function scope. This exposes cross-permute ILP that the single-permute `Permutation::permute_mut` API hides behind the trait-method call boundary.

Per-permute wall-time reduction across measured configs (post-#1645, 5-seed N=1M):

|              | Zen 5 AVX-2 (4L) | Zen 5 AVX-512 (8L) | Pi 5 NEON (2L) |
|--------------|-----------------:|-------------------:|---------------:|
| **Width 8**  | ~27.2% ± 0.7     | ~31.5% ± 0.9       | ~10.82% ± 0.08 |
| **Width 12** | ~18.6% ± 0.2     | ~26.6% ± 0.3       | ~2.65% ± 0.04  |

5 of 6 measured configs reduce >10% per permute. Pi 5 NEON width 12 is the floor at 2.65% (small but stable).

## Mechanism

Within one `permute_mut` call, the function-call boundary of the layer trait methods hides the partial-round dependency chain (round N's S-box on `state[0]` depends on round N-1's MDS output) from LLVM and the OoO engine. Calling `permute_mut(a); permute_mut(b)` back-to-back gives **0%** improvement on Zen 5 (verified by trivial-batched H1 test) because LLVM cannot see across the boundary to interleave the two states.

Hand-rolling the round structure across two states inside one function scope makes both states' operations visible to the compiler. LLVM emits an interleaved instruction stream; the OoO engine schedules state-A and state-B ops to fill issue slots that sequential single-permute leaves unused.

Microarchitectural sanity-check on the sibling Poseidon1 port (same mechanism, separate investigation): IPC 2.484 → 2.954 (+18.9%) with unchanged instruction count per permute, ~2.7× instruction scaling matching 2-states × full-vs-partial round structure, *lower* register spill rate in batched (4.5%) vs single (7.2%). Poseidon2 wall-time wins above are consistent with the same OoO-slack story.

## Where this applies

**Useful**: any caller that invokes `Permutation::permute_mut` on independent states in a loop. Primary candidate is Merkle tree layer compression at `merkle-tree/src/merkle_tree.rs:488-504`, where sibling compressions within a layer are independent and the loop is parallel-chunkable. FRI commit-phase tree builds are a secondary target.

**Doesn't apply to sequential paths**: sponge-absorb hash chains (`symmetric/src/sponge.rs:189,196`) have a per-call data dependency (each output feeds into the next), so cross-permute interleaving has no slack to capture. These callers continue using the unchanged `permute_mut` path with no behavior change.

This PR exposes the primitive only. Production call-site wiring (`TruncatedPermutation::compress_batch_b2` or similar) is intentionally left as a follow-up so the API-shape conversation can happen in review.

## Why NEON width 12 is small

The mechanism requires spare execution-port issue width and OoO reorder budget. Cortex-A76 has 2 NEON pipes vs Zen 5's 4 vector pipes, plus a smaller OoO window. Width 12's internal-round path (`internal_layer_mat_mul_goldilocks_12`) does more ops per round than `_8`, bringing single-permute closer to pipe saturation on NEON. Less slack to harvest.

Pipe-limit confirmed with a B=4 probe at width 8 on NEON: 4 × 8 = 32 hot state regs fits NEON's 32 v-registers, but **B=4 slightly loses to B=2** (mean 10.55% vs 11.16% across 5 seeds). Adding states past B=2 has no extra pipe capacity. On x86 both AVX-2 and AVX-512 win MORE at B=4 (+2-4pp), consistent with more issue width available.

## Numbers (single seed reference, full data in commit)

Mean ns/permute, single seed (seed=1):

| Width | ISA           | Single | Batched | Reduction |
|------:|---------------|-------:|--------:|----------:|
| 8     | Zen 5 AVX-2   | 1734   | 1243    | 28.28%    |
| 8     | Zen 5 AVX-512 | 1715   | 1170    | 31.79%    |
| 8     | Pi 5 NEON     | 3891   | 3473    | 10.75%    |
| 12    | Zen 5 AVX-2   | 2260   | 1846    | 18.32%    |
| 12    | Zen 5 AVX-512 | 2284   | 1676    | 26.60%    |
| 12    | Pi 5 NEON     | 5678   | 5530    | 2.60%     |

**Hardware**:
- Zen 5: Ryzen 9 9950X3D, `taskset -c 0`, governor=performance, `target-cpu=native`.
- Pi 5: Cortex-A76, `taskset -c 3`, governor=performance, `perf_event_paranoid=1`.

Not measured on Apple M-series, AWS Graviton, or Intel x86. Happy to add numbers if a maintainer can run them. The mechanism is microarchitecture-dependent in magnitude (it exploits OoO reorder budget and execution-port issue width), so the *direction* of the speedup should hold on any superscalar OoO core, but the *magnitude* depends on the specific microarchitecture's issue width, vector-pipe count, and OoO window depth. The Pi 5 NEON width 12 floor of 2.65% reflects Cortex-A76's 2 vector pipes; wider cores (Apple M-series, Graviton 4, Intel server) likely sit closer to the Zen 5 numbers.

## API shape

This PR ships a free function `pub fn permute_batch_b2_p2_w{8,12}` in `goldilocks::poseidon2_batched`. Width-specific because the partial-round matmul (`internal_layer_mat_mul_goldilocks_{8,12}`) is width-specific.

The natural shape for upstream integration is a `BatchedPermutation<T, const B: usize>` marker trait extending `Permutation<T>`, with default-impl looping and specialised impls where the win exists. I deliberately left the trait out of this PR so the primitive can land independently of the API conversation. Happy to add the trait shape to this PR if you'd prefer that direction.

## Scope / follow-ups

- **Width 16 + 20 Goldilocks**: same mechanism applies; partial-round matmul exists for each width. Not ported here.
- **Monty31 fields (Mersenne31, KoalaBear, BabyBear)**: mechanism is field-agnostic; port is mechanical once the trait shape is settled.
- **`BatchedPermutation` marker trait + `TruncatedPermutation::compress_batch_b2`**: API design conversation, follow-up PR.
- **B > 2**: B=4 wins +2-4pp on x86 but loses 0.6pp on NEON (pipe saturation). B=2 is the cross-arch sweet spot.

## Related

- Builds on #1645 (`Poseidon2GoldilocksFused` full-round MDS fix); the baseline above is post-#1645.
- Mechanism-distinct from #1656 (currently open): #1656 routes `Poseidon2GoldilocksFused<8>` packed permute through the generic path on aarch64. The NEON numbers cited here are *batched-generic vs single-generic*, which holds independent of #1656's land status.

## Test plan

- [x] `cargo +nightly fmt --all -- --check`: clean
- [x] `cargo +stable clippy --all-targets -- -D warnings`: clean
- [x] `cargo +stable sort --workspace --grouped --check`: clean
- [x] `cargo test -p p3-goldilocks` on Zen 5: 163 passed, 0 failed
- [x] AVX-2 + AVX-512 cfg-gated tests on Zen 5: 4/4 passed each
- [x] NEON cfg-gated tests on Raspberry Pi 5: 4/4 passed
- [ ] CI macOS aarch64 verification on PR open

cc @Nashtare . post-#1645 follow-up, would value your eye on the API-shape question in particular.

### DIFF
--- a/goldilocks/examples/poseidon2_batched_perf.rs
+++ b/goldilocks/examples/poseidon2_batched_perf.rs
@@ -1,0 +1,366 @@
+//! Perf comparison: single `Poseidon2::permute_mut` vs the cross-permute
+//! batched `permute_batch_b2_p2_w{8,12}` on the same packed type.
+//!
+//! On aarch64 the production `Poseidon2Goldilocks<8>` resolves to the
+//! fused NEON kernel; an extra fused-baseline row is printed for context.
+//!
+//! Build (AVX-2):
+//!   RUSTFLAGS="-C target-feature=+avx2" cargo build --release \
+//!     --example poseidon2_batched_perf -p p3-goldilocks
+//! Build (NEON):
+//!   cargo build --release --example poseidon2_batched_perf -p p3-goldilocks
+//!
+//! Run:
+//!   target/release/examples/poseidon2_batched_perf [N] [--seed K]
+//! Defaults: N=1_000_000, seed=1.
+
+#[cfg(not(any(
+    all(target_arch = "x86_64", target_feature = "avx2"),
+    target_arch = "aarch64",
+)))]
+fn main() {
+    eprintln!("poseidon2_batched_perf: requires x86_64+avx2 or aarch64.");
+    std::process::exit(1);
+}
+
+#[cfg(any(
+    all(target_arch = "x86_64", target_feature = "avx2"),
+    target_arch = "aarch64",
+))]
+fn main() {
+    harness::run();
+}
+
+#[cfg(any(
+    all(target_arch = "x86_64", target_feature = "avx2"),
+    target_arch = "aarch64",
+))]
+mod harness {
+    use std::env;
+    use std::hint::black_box;
+    use std::time::Instant;
+
+    use p3_goldilocks::poseidon2_batched::{
+        default_goldilocks_poseidon2_batched_8, default_goldilocks_poseidon2_batched_12,
+        permute_batch_b2_p2_w8, permute_batch_b2_p2_w12,
+    };
+    use p3_goldilocks::{
+        GOLDILOCKS_POSEIDON2_RC_8_EXTERNAL_FINAL, GOLDILOCKS_POSEIDON2_RC_8_EXTERNAL_INITIAL,
+        GOLDILOCKS_POSEIDON2_RC_8_INTERNAL, GOLDILOCKS_POSEIDON2_RC_12_EXTERNAL_FINAL,
+        GOLDILOCKS_POSEIDON2_RC_12_EXTERNAL_INITIAL, GOLDILOCKS_POSEIDON2_RC_12_INTERNAL,
+        Goldilocks, Poseidon2ExternalLayerGoldilocks, Poseidon2InternalLayerGoldilocks,
+    };
+    use p3_poseidon2::{ExternalLayerConstants, Poseidon2};
+    use p3_symmetric::Permutation;
+    use rand::SeedableRng;
+    use rand::distr::{Distribution, StandardUniform};
+    use rand::rngs::SmallRng;
+
+    #[cfg(all(target_arch = "x86_64", target_feature = "avx512f"))]
+    type Packed = p3_goldilocks::PackedGoldilocksAVX512;
+    #[cfg(all(target_arch = "x86_64", target_feature = "avx512f"))]
+    const PACKED_LABEL: &str = "avx512 (8L)";
+
+    #[cfg(all(
+        target_arch = "x86_64",
+        target_feature = "avx2",
+        not(target_feature = "avx512f")
+    ))]
+    type Packed = p3_goldilocks::PackedGoldilocksAVX2;
+    #[cfg(all(
+        target_arch = "x86_64",
+        target_feature = "avx2",
+        not(target_feature = "avx512f")
+    ))]
+    const PACKED_LABEL: &str = "avx2 (4L)";
+
+    #[cfg(target_arch = "aarch64")]
+    type Packed = p3_goldilocks::PackedGoldilocksNeon;
+    #[cfg(target_arch = "aarch64")]
+    const PACKED_LABEL: &str = "neon (2L)";
+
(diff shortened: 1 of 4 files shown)


Niek-Kamer:
@tcoratger @Nashtare worth flagging: on Pi 5 Cortex-A76 NEON the per-permute reduction floors at 2.65% (width 12) and 11.16% (width 8 B=2), vs 26.6% / 31.5% on Zen 5 AVX-512. A B=4 probe at width 8 NEON gives 10.55%, slightly below B=2, so the gap reads as Cortex-A76's 2-pipe NEON saturating rather than a different kernel shape unlocking more.

Given the move toward client-side proving, is this Cortex-A consumer-tier ceiling on the radar, or is the working assumption that mobile/embedded provers stay on smaller circuits where the saturation isn't binding?

goldilocks/examples/poseidon2_batched_perf.rs
tcoratger, line 1: We don't need an example file for this no? In general we just have examples for big protocols like FRI or the end to end proving. If this is to showcase performances, normally the performance improvement should be captured by our benchmark file.

goldilocks/src/poseidon2_batched.rs
```diff
@@ -0,0 +1,369 @@
...
+
+        let mut got_a = in_a;
+        let mut got_b = in_b;
+        permute_batch_b2_p2_w12(&mut got_a, &mut got_b, &constants);
+
+        assert_eq!(got_a, exp_a, "p2 w12 scalar a mismatch");
+        assert_eq!(got_b, exp_b, "p2 w12 scalar b mismatch");
+    }
+
+    #[cfg(all(target_arch = "x86_64", target_feature = "avx512f"))]
```
tcoratger, line 219: Here we have architecture dependent tests which looks weird because in general we don't do this. I think we can unify this and not make it architecture dependent.

Nashtare:
Closing for inactivity.

PR #1668, opened 2026-05-17
fix(fri): preserve width in verifier batch dimensions

mk0walsk:
Derive verifier batch dimensions from claimed opening widths and enforce cross-point width consistency, eliminating the `width: 0` fallback that can diverge from stricter MMCS semantics.


tcoratger:
Duplicate of https://github.com/Plonky3/Plonky3/pull/1757

PR #1672, opened 2026-05-18
refactor: move sumcheck module to an independent crate

Nashtare:
I'm suggesting to move the whole sumcheck logic outside of `p3-whir` into a dedicated `p3-sumcheck` new crate. Only plumbing changes, no functional ones.

This would help for reuse outside of whir for things like LogUp-GKR, other multilinear PIOP, etc..., reduce the dependency surface, make the `p3-whir` crate leaner and easier to review / audit with clearer boundaries, and possibly easier compilation caching as net benefit too.

For this I migrated the `constraints` portion (~3k LoC) into the new `p3-sumcheck` to make it sufficient as standalone, although this is a debatable design decision.

@tcoratger let me know if the migration makes sense to you or if you'd rather have a larger `p3-whir` / different split strategy


tcoratger (approved):
Thanks, this makes total sense.

The WHIR crate is getting too big and having a separate crate for sumcheck is totally legit. 

Right now, the implementation is tailored to WHIR but I've in mind to make it more general purpose in the future so that we have a generic sumcheck that everyone can use and maybe can be used for LogUp with GKR as well :)

PR #1673, opened 2026-05-18
feat: add STIR PCS

### LINKED ISSUE #1647: Land STIR as a low-degree test alongside FRI/WHIR
## Summary

[STIR](https://eprint.iacr.org/2024/390) (Arnon, Chiesa, Fenzi, Yogev) is an IOPP for Reed–Solomon proximity testing with strictly better query complexity than FRI — `O(log d + λ·log log d)` vs FRI's `O(λ·log d)` — at the cost of slightly higher prover time. Compared to FRI it yields 1.25×–2.46× smaller arguments and ~1.55×–2.67× fewer verifier hashes (Table 1 / Table 2 of the paper).

The goal of this issue is for someone to pick up an existing WIP branch and turn it into a mergeable PR, so we ship STIR alongside FRI (and complementary to WHIR) for the next major release.

This supersedes / refocuses #485, which is the older discussion thread with no follow-up.

## Status: there is a WIP branch to pick up

@Nashtare started an implementation a while back on the [`robin/stir`](https://github.com/Plonky3/Plonky3/tree/robin/stir) branch (13 commits, currently behind `main`). It is not Miden's priority right now, so he likely won't have bandwidth in the coming weeks to push it further.

**Whoever picks this up: please keep Robin's commits in the PR — he should remain a co-author of the merged work.** The easiest path is to branch off `robin/stir`, rebase onto current `main`, and continue iterating on top.

The branch already includes:
- A STIR PCS skeleton (`feat: add STIR PCS`)
- Parameter derivation from the paper
- Per-round PoW derivation
- A faster `eval_degree_correction`
- Tests and cleanup passes

## Known TODOs / things to tweak

Per Robin, things he flagged as still WIP:

- [ ] **Proof size**: currently the folded polynomial is sent to the verifier at each round, making proofs larger than necessary. In theory this isn't needed — folding can be derived implicitly. (There's a commit titled `fix: do not send folded poly at each step` on the branch, but Robin notes this is still being tweaked.)
- [ ] Rebase onto current `main` (branch is 32 commits behind).
- [ ] Sanity-check parameter choices and round-by-round soundness against the paper (Section 5.2, Appendix C).
- [ ] Tests vs FRI on equivalent inputs (cross-check soundness and proof size).
- [ ] Decide where STIR lives in the workspace — own crate (`stir/`) mirroring `fri/` and `whir/` is the natural shape.

## What STIR adds, briefly

- Recursive degree reduction by a factor `k` per iteration while shrinking the evaluation domain only by 2 — so the **rate improves** every round (unlike FRI), which is what drives the reduced query count.
- A more efficient batch degree-correction primitive (Section 4.5 / Theorem 2 of the paper) usable beyond STIR itself.
- A clean poly-IOP → IOP compiler with round-by-round knowledge soundness (Section 7) — useful if/when we want to plug STIR into higher-level protocols.

The reference implementation by the paper authors is at [WizardOfMenlo/stir](https://github.com/WizardOfMenlo/stir) (arkworks-based) and is a useful comparison point.

## Review

Once there's something concrete, it would be valuable to have @WizardOfMenlo (Giacomo Fenzi, one of the paper's authors) — or someone in that circle — take a look at the cryptographic correctness pass alongside the usual Plonky3 maintainers.

## Refs

- Paper: https://eprint.iacr.org/2024/390
- Reference impl: https://github.com/WizardOfMenlo/stir
- WIP branch: https://github.com/Plonky3/Plonky3/tree/robin/stir
- Prior discussion: #485

latifkasuli:
## Summary

- Land the STIR PCS work from @Nashtare's `robin/stir` history as `p3-stir` alongside the existing FRI and WHIR commitment code.
- Integrate the implementation with current `main`, including the shared `p3-security` soundness model and embedded CI coverage.
- Correct STIR's PoW accounting and Johnson-bound parameter derivation, with end-to-end non-zero grinding coverage for both Capacity- and Johnson-bound configurations.
- Apply the review-driven prover/verifier improvements: codeword-native proving, packed row reduction, parallel virtual-oracle construction, batched inversions, and release-mode duplicate-point validation.
- Add an apples-to-apples BabyBear proof-object size comparison against binary FRI and mark the STIR PCS in the README.

Closes #1647.

## Proof-size comparison

Measured on the current branch with:

```sh
cargo test -p p3-stir test_pcs_proof_size_vs_binary_fri_equivalent_input -- --nocapture
```

| log_degree | log_folding_factor | STIR proof bytes | binary FRI proof bytes | FRI/STIR |
|---:|---:|---:|---:|---:|
| 14 | 2 | 148819 | 71544 | 0.48x |
| 16 | 2 | 171437 | 98139 | 0.57x |

The always-running test is a diagnostic comparison. After the upstream FRI multiproof changes, binary FRI is smaller for these two configurations, so that test verifies both proofs and reports their serialized sizes without failing ordinary CI. The ignored `assert_stir_proof_smaller_than_binary_fri` regression records the intended STIR-over-FRI ordering and is linked to the TODO to port the #1917/#1919 pruning approach to STIR.

The comparison intentionally measures serialized `Pcs::Proof` objects only. It excludes opened values because both proofs open the same commitment-derived point with the same matrix width. The FRI comparator uses binary FRI (`max_log_arity = 1`) with proof-of-work disabled; broader parameter and arity sweeps remain follow-up benchmarking work.

## Validation

```sh
cargo +nightly fmt --all -- --check
cargo +stable sort --workspace --grouped --check
taplo fmt --check
cargo +stable clippy --all-targets -- -D warnings
RUSTDOCFLAGS='-D rustdoc::broken_intra_doc_links' cargo +stable doc --no-deps --workspace --document-private-items
cargo test --quiet
cargo test --features parallel --quiet
cargo +stable build --target thumbv7em-none-eabi -p p3-stir
cargo test -p p3-stir --release test_interpolate_poly_rejects_duplicate_points
git diff --check upstream/main...HEAD
```

The complete GitHub Actions matrix also passes on Ubuntu, Ubuntu AVX2, Ubuntu AVX-512, ARM SVE2, macOS, Windows, wasm32 SIMD, and the embedded target.

## Credit

This PR is based on @Nashtare's STIR implementation branch; the original STIR commits are preserved with their existing authorship.


Nashtare:
Thanks for doing this! Note that once this preliminary version lands, we'll probably end up doing some refactoring to bundle together multilinear primitives / pieces of logic shared by WHIR and STIR, but probably best to hold off.

latifkasuli:
Thanks Robin, agreed. I’ll keep this PR scoped to landing the preliminary p3-stir version and leave the shared WHIR/STIR primitive refactors for follow-up work after this lands.

Nashtare:
It may be worth doing a pass over the recent improvements in `p3-whir` and `p3-sumcheck`, as some may be applicable to the internals of `p3-stir`. I wonder also how we would like to tackle this, as just one large PR or split into distinct modules ensuring both easier review and allowing directly some refactoring along the way as I'd suspect we can bundle a large portion of the code with the STIR part of WHIR too.

latifkasuli:
Agreed. I’ll do a pass over the recent `p3-whir` / `p3-sumcheck` changes and check for low-risk improvements that should be mirrored in `p3-stir`.

For this PR, I’ll keep the scope focused on landing the preliminary STIR PCS unless there are small correctness/cleanup changes worth pulling in now. Larger shared primitive refactors between WHIR/STIR can be split into follow-up PRs so review stays manageable.

stir/src/prover.rs
```diff
@@ -0,0 +1,376 @@
...
+            &config.mmcs,
+            &next_commit_codeword,
+            config.log_folding_factor,
+        );
+        challenger.observe(new_commit.clone());
+
+        // Step 2: query/OOD PoW. Placed BEFORE OOD sampling so PoW gates re-rolls of
+        // the OOD set (the prover would otherwise have unbounded freedom to re-commit
+        // the next codeword and resample favorable OOD points without paying PoW).
+        let pow_witness = challenger.grind(rc.pow_bits);
```
Nashtare: actually I think there's an issue here, the challenger should grind _after_ observing the `ood_answers` below, the current makes us lose ~20 bits of security atm.

stir/src/pcs.rs
```diff
@@ -0,0 +1,709 @@
...
+impl<Val, Dft, InputMmcs, StirMmcs> BuildPeriodicLdeTableFast
+    for TwoAdicStirPcs<Val, Dft, InputMmcs, StirMmcs>
+where
+    Val: TwoAdicField,
+{
+    type PeriodicDomain = TwoAdicMultiplicativeCoset<Val>;
+}
+
+impl<Val, Dft, InputMmcs, StirMmcs, Challenge, Challenger> Pcs<Challenge, Challenger>
+    for TwoAdicStirPcs<Val, Dft, InputMmcs, StirMmcs>
```
Nashtare, line 72: possibly a rebase issue, but needs to be overriding `log_max_lde_height()` as well

commit/src/soundness.rs
```diff
@@ -0,0 +1,1484 @@
...
+            Self::JohnsonBound => 1. - libm::sqrt(rate) - libm::pow(2., self.log_eta(log_inv_rate)),
+            Self::CapacityBound => 1. - rate - libm::pow(2., self.log_eta(log_inv_rate)),
+        };
+
+        libm::log2(1. - delta)
+    }
+
+    /// Compute the number of queries needed to achieve the given security level.
+    #[must_use]
+    pub fn queries(&self, protocol_security_level: usize, log_inv_rate: usize) -> usize {
```
Nashtare: we probably don't need to make this and many other methods part of the public API

stir/src/utils.rs
```diff
@@ -0,0 +1,640 @@
...
+///
+/// Returns `true` if the check passes.
+pub fn check_shake_consistency<F: Field>(
+    ans: &[F],
+    shake: &[F],
+    points: &[F],
+    values: &[F],
+    rho: F,
+) -> bool {
+    debug_assert_eq!(points.len(), values.len());
```
Nashtare: I think this one may be worth raising to release level (or shortcutting to false rather so we don't panic but abort early)

stir/src/prover.rs
```diff
@@ -0,0 +1,375 @@
...
+    // (config.log_folding_factor) and the initial domain shift is always F::GENERATOR.
+    let initial_shift = F::GENERATOR;
+    let log_initial_domain = config.log_starting_domain_size();
+    let initial_domain_size = 1 << log_initial_domain;
+
+    let mut coeffs = poly_coeffs;
+    coeffs.resize(initial_domain_size, EF::ZERO);
+    let initial_codeword = codeword_from_coeffs(dft, coeffs, initial_shift, log_initial_domain);
+
+    let mut current_oracle_codeword = initial_codeword.clone();
```
Nashtare: actually we already compute the "the reduced-opening codeword -> coefficients" in open(), so we don't need to recompute it here

Nashtare:
@latifkasuli I took the liberty to push directly onto your branch a commit adding parallelization for the prover that I noticed was missing while doing the initial review 🙂

stir/src/config.rs
```diff
@@ -0,0 +1,910 @@
...
+                libm::log2(final_eta),
+            );
+            let query_alg = params.soundness_type.stir_query_algebraic_bits(
+                field_size_bits,
+                log_degree,
+                log_inv_rate,
+                final_eta,
+                num_queries,
+                num_ood_samples,
+            );
```
Nashtare: I think there's an actual issue here. Per the transcript order, the grind sits after `ood_answers` and before `r_comb`/query indices, so the OOD points get 0 PoW protection and the shake rho as well.

commit/src/soundness.rs
```diff
@@ -0,0 +1,1485 @@
...
+            // JB needs no round-0 combination term: its list size
+            // `~1/(2·eta·√ρ)` stays small, so `queries_combination_error` is
+            // satisfied with a large margin at the prox-gap eta (unlike CB,
+            // where `list_size = d/(ρ·eta)` blows up as eta shrinks).
+            Self::JohnsonBound => {
+                ((security_bits as f64) + log_k_minus_1 + 2. * log_d_over_k
+                    - 7.
+                    - field_size_bits as f64)
+                    / 7.
+            }
```
Nashtare: actually I think there's an issue here too (I noticed it while trying actual configs and not dummy ones), but this eta (resolved from BCIKS-form) would make JB configs panic on actual parameters (say 100/128 bits of security). I think we'd need to use the tighter BCSS25 bound (which is actually used during validation in `prox_gaps_error_at_log_eta`)

stir/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+//!   `ceil(log2(6 · total_folds))` buffer to every per-round error term (query failure plus
+//!   the auxiliary terms bridged by PoW). Each round contributes up to six independently
+//!   bridged algebraic failure modes (query tier: query failure, OOD, random-combination,
+//!   shake-check; folding tier: proximity-gaps, sumcheck), so the union bound over the full
+//!   protocol is `≤ 6 · total_folds · 2^{-buffered_security_level}`. The paper's "+1 / +0"
+//!   rule only delivers the claimed bits when `total_folds ≤ 2` and per-round terms are
+//!   collapsed; the explicit log keeps deeper protocols tight across every term.
+//! - **PoW placement.** Each round's query/OOD PoW grind is placed BEFORE OOD sampling, so
+//!   it gates re-rolls of the OOD set via re-commitment (the paper's high-level protocol
+//!   does not specify PoW). The folding PoW gates the fold challenge as in the paper.
```
Nashtare: nit: outdated I believe, now it can be fully removed
```suggestion
```

stir/src/pcs.rs
```diff
@@ -0,0 +1,713 @@
...
+
+                // Precompute alpha-batched row values for this matrix (reused per point).
+                let p_x_vec: Vec<Challenge> = mat
+                    .rows()
+                    .map(|row| {
+                        row.zip(alpha_powers.iter())
+                            .map(|(px, &ap)| Challenge::from(px) * ap)
+                            .sum()
+                    })
+                    .collect_vec();
```
Nashtare: minor: but I think we can improve perfs by making this parallel? We can also reuse the packed dot-product idiom

stir/src/prover.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    // Commit to the initial codeword on L_0. The folding factor is constant across rounds
+    // (config.log_folding_factor) and the initial domain shift is always F::GENERATOR.
+    let initial_shift = F::GENERATOR;
+    let log_initial_domain = config.log_starting_domain_size();
+    let initial_domain_size = 1 << log_initial_domain;
+
+    let mut coeffs = poly_coeffs;
+    coeffs.resize(initial_domain_size, EF::ZERO);
+    let initial_codeword = codeword_from_coeffs(dft, coeffs, initial_shift, log_initial_domain);
```
Nashtare, line 61: in `pcs` we iDFT `ro` to coefficients, to be DFTed back here. I wonder if we could not have a `prove_stir_from_codeword` alternative entry-point, which would remove the need for the subsequent heavy clone `initial_codeword.clone()`

stir/src/prover.rs
```diff
@@ -0,0 +1,379 @@
...
+        let _rho: EF = challenger.sample_algebra_element();
+
+        // Step 5: Construction 5.2 — compute the next virtual witness polynomial
+        // f_{i+1} = DegCor((g_i − Ans_i) / Z_{G_i}).
+        let num_answers = all_points.len();
+        let numerator = add_polys(&fold_coeffs, &scale_poly(&ans_poly, EF::ZERO - EF::ONE));
+        let quotient = quotient_by_roots(&numerator, &all_points);
+        let f_next_coeffs = degree_correct(&quotient, r_comb, num_answers);
+        let next_oracle_codeword =
+            codeword_from_coeffs(dft, f_next_coeffs, next_shift, next_log_domain);
```
Nashtare: I experimented a bit with it, and if we were to add a 
```rust

  /// Coefficients of the vanishing polynomial `prod_{y in roots} (X - y)`.
  ///
  /// The result has length `roots.len() + 1` with leading coefficient one.
  pub fn vanishing_poly_from_roots<F: Field>(roots: &[F]) -> Vec<F> {
      let mut coeffs = Vec::with_capacity(roots.len() + 1);
      coeffs.push(F::ONE);
      for &y in roots {
          coeffs.push(F::ZERO);
          // Multiply by (X - y) in place, descending so each slot reads its
          // unmodified lower neighbour.
          for i in (1..coeffs.len()).rev() {
              coeffs[i] = coeffs[i - 1] - coeffs[i] * y;
          }
          coeffs[0] = -coeffs[0] * y;
      }
      coeffs
  }
```
method (in the utility module most likely, next to `eval_vanishing_at_roots`), we could then replace this with 
 
```suggestion
          // Step 5: Construction 5.2 — the next virtual witness oracle
          // f_{i+1} = DegCor((g_i − Ans_i) / Z_{G_i}), built on L_{i+1}.
          // Z_{G_i} is nonzero there: OOD points are rejected inside the next domain and the
          // query points lie in L_i^k, which the disjoint-coset schedule keeps apart from L_{i+1}.
          let num_answers = all_points.len();

          let ans_evals = codeword_from_coeffs(dft, ans_poly.clone(), next_shift, next_log_domain);
          let vanishing_evals = codeword_from_coeffs(
              dft,
              vanishing_poly_from_roots(&all_points),
              next_shift,
              next_log_domain,
          );
          let deg_cor_evals = codeword_from_coeffs(
              dft,
              degree_correction_poly(r_comb, num_answers),
              next_shift,
              next_log_domain,
          );
          let vanishing_invs = batch_multiplicative_inverse(&vanishing_evals);

          let next_oracle_codeword: Vec<EF> = next_commit_codeword
              .par_iter()
              .zip(ans_evals.par_iter())
              .zip(vanishing_invs.par_iter())
              .zip(deg_cor_evals.par_iter())
              .map(|(((&g, &ans), &z_inv), &dc)| (g - ans) * z_inv * dc)
              .collect();
```

This yields roughly 3x / 4x speed-up on the open phase by removing a whole serial synthetic division and the `degree_correct` pass (at the cost of 2 extra dfts)

stir/src/verifier.rs
```diff
@@ -0,0 +1,483 @@
...
+        .iter()
+        .enumerate()
+        .map(|(col, &g_value)| {
+            let natural_index = row_index + col * row_height;
+            let x = EF::from(current_shift) * EF::from(domain_gen.exp_u64(natural_index as u64));
+            let vanishing = eval_vanishing_at_roots(&ctx.all_points, x);
+            if vanishing == EF::ZERO {
+                return None;
+            }
+            let quotient = (g_value - eval_poly(&ctx.ans_poly, x)) * vanishing.inverse();
```
Nashtare: actually we should be able to do batch inversion here to amortize the cost

stir/src/utils.rs
```diff
@@ -0,0 +1,712 @@
...
+        // Precompute g^j for j = 0..new_height.
+        let g_powers: Vec<F> = {
+            let mut acc = F::ONE;
+            (0..new_height)
+                .map(|_| {
+                    let v = acc;
+                    acc *= g;
+                    v
+                })
+                .collect()
```
Nashtare:
```suggestion
                .collect_n()
```
should speed things up a bit

stir/src/utils.rs
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    // Newton's divided differences table.
+    let mut dd = values.to_vec();
+    for k in 1..n {
+        for i in (k..n).rev() {
+            let denom = points[i] - points[i - k];
+            dd[i] = (dd[i] - dd[i - 1]) * denom.inverse();
+        }
+    }
```
Nashtare: here too we could batch inverse

Nashtare (commented):
Apart from the soundness issue w.r.t grinding and the config issue, I think we're in a pretty good state already :)

stir/src/prover.rs
latifkasuli: Fixed in `99c8bcfb`: the prover observes `ood_answers` before grinding. The soundness accounting now gives those OOD challenges no PoW credit and applies the grind only to the downstream combination and query challenges.

stir/src/pcs.rs
latifkasuli: Confirmed and retained on the rebased implementation: `StirPcs` overrides `log_max_lde_height()` alongside `natural_domain_for_degree()`.

commit/src/soundness.rs
latifkasuli: Addressed during the rebase: the shared assumption type now comes from `p3-security`, while the STIR-specific soundness formulas live in private `stir/src/soundness.rs` instead of expanding the `p3-commit` public API.

stir/src/utils.rs
latifkasuli: Fixed: duplicate interpolation points are rejected unconditionally rather than only through a debug assertion. A release-mode regression test covers this path.

stir/src/prover.rs
latifkasuli: Fixed through the new `prove_stir_from_codeword` entry point. The PCS now passes the reduced-opening codeword directly, avoiding the iDFT-to-DFT round trip and the initial codeword clone.

stir/src/config.rs
latifkasuli: Fixed. OOD points and the shake `rho` now receive zero PoW credit; only challenges sampled immediately after the grind receive that credit. The prover transcript order and regression tests match this accounting.

commit/src/soundness.rs
latifkasuli: Fixed. The initial Johnson-bound `eta` is now derived from the tighter BCSS25 inequality used by validation. Constructor-level regressions cover realistic 100-bit BabyBear and 128-bit Goldilocks configurations without relying on PoW.

stir/src/lib.rs
latifkasuli: Removed the outdated crate-level PoW documentation.

stir/src/pcs.rs
latifkasuli: Implemented with `rowwise_packed_dot_product`, using the packed and parallel row-reduction path.

stir/src/prover.rs
latifkasuli: Implemented `prove_stir_from_codeword` and switched the PCS caller to it. This removes both the reduced-opening iDFT-to-DFT round trip and the heavy initial codeword clone.

stir/src/prover.rs
latifkasuli: Implemented this construction: `Ans`, the vanishing polynomial, and degree correction are evaluated by DFT; the vanishing evaluations are batch-inverted; and the final oracle is assembled with a parallel zip.

stir/src/verifier.rs
latifkasuli: Implemented: all fiber denominators are collected and batch-inverted before the verifier consistency check.

stir/src/utils.rs
latifkasuli: Applied `collect_n()` for the fixed-length powers collection.

stir/src/utils.rs
latifkasuli: Implemented batch inversion for the Newton interpolation denominators.

stir/tests/stir.rs
```diff
@@ -0,0 +1,1111 @@
...
+
+    #[test]
+    fn test_pcs_proof_size_vs_binary_fri_equivalent_input() {
+        const WIDTH: usize = 3;
+
+        for (log_degree, log_folding_factor) in [(14, 2), (16, 2)] {
+            compare_stir_proof_size_with_binary_fri(log_degree, log_folding_factor, WIDTH);
+        }
+
+        fn compare_stir_proof_size_with_binary_fri(
```
Nashtare: I think we should have kept the previous `assert_stir_proof_smaller_than_binary_fri`, which is essentially the real comparison we want to have (the goal behind STIR as replacement for FRI). Because STIR does not yet include the changes from #1919 / #1917, I guess it would be fine to keep it ignored with a TODO mentioning the same pruning can be done for STIR, but it would be much more meaningful than a plain size comparison as a unit test.

stir/src/config.rs
Nashtare: Thanks! I think we should exercise this (and the new Johnson-bound `eta` calculation) by actually using non-zero grinding in some of our tests (for now they're always using 0 I believe which is hiding potentially latent bugs).

Nashtare:
@tcoratger I am broadly happy with this by now. There are definitely improvements that can be done, both on pure prover runtime and on proof size (follow-ups of pruning approach implemented for WHIR / FRI) but I think we can afford to merge this to main now. It's not yet a part of the release-plz.toml file so we can keep holding off a release of that crate until we're comfortable with its state. WDYT?

tcoratger:
> @tcoratger I am broadly happy with this by now. There are definitely improvements that can be done, both on pure prover runtime and on proof size (follow-ups of pruning approach implemented for WHIR / FRI) but I think we can afford to merge this to main now. It's not yet a part of the release-plz.toml file so we can keep holding off a release of that crate until we're comfortable with its state. WDYT?

Yes I think this is a good idea! Let's do this!

stir/src/config.rs
latifkasuli: Addressed in `043aa5ca`. The existing Capacity-bound tests already exercised non-zero grinding; I added an end-to-end Johnson-bound case using the BCSS25-derived eta, with assertions that both query and folding PoW are non-zero before proving and verifying.

stir/tests/stir.rs
latifkasuli: Restored in `043aa5ca` as ignored `assert_stir_proof_smaller_than_binary_fri`, with a TODO referencing the #1917/#1919 pruning work. It shares the verified proof-size helper with the always-running diagnostic so the two comparisons cannot drift.

Nashtare (approved):

PR #1680, opened 2026-05-19
feat(whir): user-facing SecurityBudget for per-round soundness analysis

XC0R:
Adds a `SecurityBudget` struct and `security_budget()` method to `soundness.rs`
that composes all per-component error terms into a single user-facing report.

Uses a two-tier assessment matching the WHIR soundness structure:
- **Self-sustaining** (OOD + queries must reach target alone)
- **PoW-assisted** (proximity gaps, fold sumcheck, combination may lean on grinding)

Includes `Display` impl for human-readable output and 5 tests covering manual
computation match, two-tier logic (5 sub-cases incl. UD and combination-as-bottleneck),
cross-validation against existing BCSS25 tests, and display formatting.

Standard mode only — ZK budget deferred pending `zk_fold_sumcheck_error` /
`zk_fold_prox_correction` (shipping ZK with wrong fold values would overestimate
security).

### DIFF
--- a/whir/src/parameters/mod.rs
+++ b/whir/src/parameters/mod.rs
@@ -8,7 +8,7 @@ pub mod soundness;
 pub mod whir;
 
 pub use folding::{FoldingFactor, FoldingFactorError};
-pub use soundness::SecurityAssumption;
+pub use soundness::{SecurityAssumption, SecurityBudget};
 pub use whir::{RoundConfig, WhirConfig};
 
 /// Fallback proof-of-work difficulty when the user does not specify one.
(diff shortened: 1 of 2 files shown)


tcoratger:
I'm not entirely sure that we want this for now. We already have parameters that the user can tune for any setting so I don't really see where this fits. Also correct me if I'm wrong, but right now the code you propose here is dead code that is not used anywhere.

XC0R:
`round_log_inv_rates` (#1677) configures per-round code rates. `SecurityBudget` composes the six existing error methods into a per-component breakdown: which term is the bottleneck, how many bits each contributes, whether the configuration meets its target. #1686 adds `proven_security()` for STARK-level composition. Is there a WHIR round-level equivalent, or which existing API surfaces that analysis?

closing prioritization.

PR #1681, opened 2026-05-20
perf(whir): use collect_n for select challenge powers

shahrukhjutt98:
## Summary

- Route WHIR select-statement challenge-power precomputation through `Powers::collect_n`.
- Start the sequence with `shifted_powers(gamma^shift)` so it still produces `gamma^shift, gamma^(shift + 1), ...`.
- Reuse the packed powers collection path enabled by #1639 without changing the batched constraint semantics.


Nashtare (approved):
LGTM

PR #1684, opened 2026-05-20
fix(mersenne-31): allow value == P in MDS partial-reduction

tcoratger:
@Nashtare Looks like a bug highlighted by https://github.com/Plonky3/Plonky3/pull/1683 for some reason

## Summary

`LargeConvolveMersenne31::reduce` (the partial-reduction step at the tail of the large-RHS Karatsuba convolution) constructs two `Mersenne31` intermediates via `from_canonical_unchecked`. Both inputs can equal `P = 2^31 - 1` — the non-canonical representation of the field's zero element — within the bounds the surrounding proof already establishes:

- `(z & MASK) == P` whenever the low 31 bits of `z` are all ones (e.g. `z = -1`, `z = P`).
- `high_bits + sign_bits == P` is excluded by the function's own size bounds, but the constructor was still strict on it.

`from_canonical_unchecked` requires `int < ORDER_U32` strictly, enforced by a `debug_assert!`. The boundary case `int == P` is therefore a debug-only panic. Release builds compute the correct field element (`Add` normalises `P + x` correctly), so this is purely an assertion-vs-bound mismatch — the math was always right.

## Symptom

`proptest` happens to find triggering inputs over time. Most recently it surfaced on PR #1683's CI run via:

```
thread 'aarch64_neon::poseidon1::tests::poseidon1_neon_matches_scalar_width_32' panicked at
mersenne-31/src/mersenne_31.rs:433:9:
assertion failed: int < Self::ORDER_U32
```

The MDS path runs inside the Poseidon1 NEON parity test, so the panic appears far from its root cause. Once `proptest` saves the seed in `mersenne-31/proptest-regressions/aarch64_neon/poseidon1.txt`, every subsequent debug-mode run replays it.

## Fix

Switch the two `from_canonical_unchecked` calls to `Mersenne31::new_reduced`, which already accepts `[0, P]` — the exact range the bounds in this function prove. No `unsafe` needed, no behavior change in release.

## Test

Adds `large_convolve_reduce_accepts_p_representation` — a direct unit test on `LargeConvolveMersenne31::reduce` that feeds `z = -1` and `z = P` and checks the field-reduced result. Verified failing on the pre-fix code with the exact same `assertion failed: int < Self::ORDER_U32` panic, and passing on the fix.

## Test plan

- [x] `cargo test -p p3-mersenne-31 --lib` (debug, 188 pass)
- [x] Manually re-applied the buggy code and confirmed the new test reproduces the original panic

### DIFF
--- a/mersenne-31/src/mds.rs
+++ b/mersenne-31/src/mds.rs
@@ -7,7 +7,6 @@
 //! database.
 
 use p3_field::PrimeCharacteristicRing;
-use p3_field::integers::QuotientMap;
 use p3_mds::MdsPermutation;
 use p3_mds::karatsuba_convolution::Convolve;
 use p3_mds::util::{dot_product, first_row_to_first_col};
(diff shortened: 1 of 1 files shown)


Nashtare (approved):
Thanks, I was just about to open a PR for it! LGTM

PR #1685, opened 2026-05-20
Add prover tracing context fields

cmende:
This change adds key prover configuration fields to tracing spans, making production debugging and benchmark triage faster without altering protocol behavior.


Nashtare:
Those addons don't look necessary.

PR #1686, opened 2026-05-21
refactor: bundle security logic in a dedicated crate

Nashtare:


huitseeker:
@Al-Kindi-0

Nashtare:
@tcoratger @adr1anh Mostly opening to gather feedback, as we may want to tweak things around to make this bundle modular enough for downstream users supporting custom variations of our protocols.

Nashtare:
@tcoratger I'd be keen to get this merged relatively soon so I can iterate on it more easily, on both modularity and additional features. WDYT?

security/src/stark.rs
```diff
@@ -0,0 +1,132 @@
...
+    extras: &[ErrorBits],
+) -> ErrorBits {
+    let ali = air::composition_error(air.num_constraints, list_size, shape.modulus_bits);
+    let deep = deep::deep_ali_error(air, shape, list_size);
+    let mut all: Vec<ErrorBits> = Vec::with_capacity(3 + extras.len());
+    all.push(ali);
+    all.push(deep);
+    all.push(ldt_error);
+    all.extend_from_slice(extras);
+    ErrorBits::min(&all)
```
tcoratger: Small footgun for the "modular for downstream" goal: these composite functions never cap the result at `shape.collision_resistance`, even though it's carried right there in `InstanceShape`.

Real security is `min(algebraic soundness, hash collision resistance)` — a collision in the commitment hash forges the proof regardless of the algebra. In-repo `uni-stark` remembers to clamp (`min(..., cap)`), and the tests here do `.min(shape.collision_resistance)` by hand, so the clamp currently lives at the call site, not in the crate.

A new downstream caller who trusts `proven_security_udr(...)` directly would over-claim — e.g. report 150 bits while committing with a 128-bit-collision hash (true security 128). Either apply the `min` here, or drop the field and document that clamping is the caller's job.

security/src/shape.rs
```diff
@@ -0,0 +1,69 @@
...
+            max_combo,
+        }
+    }
+}
+
+/// Per-instance shape data not in the protocol params.
+#[derive(Copy, Clone, Debug)]
+pub struct InstanceShape {
+    pub log_trace_length: usize,
+    pub num_opening_points: usize,
```
tcoratger: These two fields (`num_opening_points`, `num_columns`) are set in the test fixtures but never read by any of the error formulas. In a struct called `InstanceShape` feeding a *security* calc, a reader will assume tuning them changes the bound — it doesn't. Worth either wiring them into the terms that should depend on them (opening count really does affect batched-opening soundness) or dropping them, so nobody is misled.

tcoratger (commented):
I've reviewed it with Claude, looks good, just found deep stuffs that are non blocking for me!

Clean, behavior-preserving refactor — nice work bundling this into one place.

I checked the extracted math against the source papers (BCHKS25 Thm 1.5 + Lemma 3.1, 2024/1553 Thm 2 & 3, STIR Lemma 4.5, Random-Words Eq (5)) and it all lines up, several constants to the digit. I also ran a 64,800-config diff of `ProvenSecurity::compute` old-vs-new (0 bit differences), and the `57`/`65` regression is identical to `main`. So: no soundness concerns, safe to merge.

A few small, **non-blocking** notes inline — mostly "make an intentional choice visible" rather than "this is wrong".

One framing note for readers: this crate computes an *advisory* number (the only non-test caller just prints it), so nothing below is exploitable — it's about not misleading whoever reads the number.

security/src/deep.rs
```diff
@@ -0,0 +1,23 @@
...
+/// `-log2(ε_DEEP)` in bits. Returns 0 bits if inputs are degenerate.
+pub fn deep_ali_error(air: &StarkAirParams, shape: &InstanceShape, list_size: f64) -> ErrorBits {
+    if shape.modulus_bits == 0 || !list_size.is_finite() || list_size <= 0.0 {
+        return ErrorBits::from_log2(0.0);
+    }
+    let k = (1u64 << shape.log_trace_length) as f64;
+    let max_deg = air.max_constraint_degree.max(1) as f64;
+    let combo = air.max_combo as f64;
+    let factor = (max_deg * (k + combo - 1.0) + (k - 1.0)).max(1.0);
+    let bits = shape.modulus_bits as f64 - log2(list_size) - log2(factor);
```
tcoratger, line 27: Doc note, not a bug: the list size enters **linearly** here (matching `soundcalc`). 2024/1553 Thm 2 (`eps_2`) uses it **squared**, with a larger `L ≈ m/ρ`.

In the list-decoding regime that's a few bits more optimistic when DEEP is the binding term — e.g. at ρ=1/4, m=10: `log2(L⁺) ≈ 4.4` here vs `log2(L²) ≈ 10.6` there. Invisible in the pinned benchmark (commit/query bind there). One sentence in the module doc making the soundcalc-vs-paper choice explicit would help future readers.

security/src/fri.rs
```diff
@@ -0,0 +1,307 @@
...
+    shape: &InstanceShape,
+) -> Option<(usize, ErrorBits)> {
+    let trace_length = 1usize << shape.log_trace_length;
+    let m_max = core::cmp::min(compute_upper_m(trace_length), LDR_M_CAP);
+    let m_min = 3usize;
+    if m_max < m_min {
+        return None;
+    }
+    (m_min..=m_max)
+        .map(|m| (m, proven_error_ldr_m(regime, air, shape, m)))
```
tcoratger, line 200: Worth a one-line comment here: `best_ldr_m` optimizes only `min(commit, query)` (the LDT-only error), whereas the value ultimately reported is the full `min(ALI, DEEP, commit, query)`.

That's fine because ALI/DEEP don't bind at the optimum in practice — I scanned 64,800 configs and the chosen `m` is identical to optimizing the full composite. But a future editor could reasonably assume the objective is the full thing, so a note pins the invariant.

uni-stark/src/security.rs
```diff
@@ -224,290 +219,57 @@ impl ProvenSecurity {
...
-                ),
-                params.collision_resistance as u64,
-            ) as usize
-        };
-
-        Self {
-            unique_decoding_bits: unique_decoding,
-            list_decoding_bits: list_decoding,
-        }
+        Self::compute_from_proof(log2_floor_usize(trace_length), params)
```
tcoratger, line 234: Heads-up: `compute` now floors the trace length to a power of two via `log2_floor_usize`. For a non-pow2 `trace_length` that analyzes a *smaller* domain, which reports *more* bits (the optimistic direction).

Harmless in practice — committed degrees are always powers of two and the real path uses `compute_from_proof(degree_bits, …)` — but an `assert!(trace_length.is_power_of_two())` or a doc line would prevent a surprise if someone passes a raw length.

tcoratger (approved):

PR #1690, opened 2026-05-22
whir: reject final polynomial length mismatch

shahrukhjutt98:
## Summary

Reject WHIR proofs whose final polynomial length does not match the verifier's final round shape before observing it in the transcript.

This follows the existing malformed-proof hardening in the WHIR verifier by returning a typed `FinalPolyLengthMismatch` error instead of letting an invalid tail polynomial flow into final STIR and sumcheck checks.


Nashtare (approved):
LGTM thanks, I think you'll need to run `rustfmt`

PR #1692, opened 2026-05-22
feat: add a `HornerIter` supertrait on `DoubleEndedIterator`

Nashtare:


lookup/src/tests.rs
```diff
@@ -309,13 +309,14 @@ fn compute_logup_contribution(
     let alpha = challenges.alpha;
     let beta = challenges.beta;
     let vals_read_len = vals_read.len();
-    let val_read = vals_read
-        .iter()
-        .fold(EF::ZERO, |acc, &v| acc * beta + EF::from(v));
+    let val_read: EF = vals_read.iter().rev().copied().map(EF::from).horner(beta);
```
adr1anh: Technically this is doing a double reversion right?

adr1anh: This promotes the input from F to EF, which I think costs an extra scalar addition per element. I don't think it's a big deal here, but I’m curious it the API supports an iterator of base field elements.

field/src/helpers.rs
```diff
@@ -390,3 +390,50 @@ where
...
+    fn horner<Acc, X>(self, x: X) -> Acc
+    where
+        Acc: Default + Mul<X, Output = Acc> + Add<Self::Item, Output = Acc>,
+        X: Clone,
+    {
+        self.horner_acc(Acc::default(), x)
+    }
+}
+
+impl<I: DoubleEndedIterator> HornerIter for I {}
```
adr1anh, line 439: It seems like most call sites have to add `.copied`. Would it make sense to either change the bounds to take references to the coefficients, or add another variant for iterators of references?

adr1anh (approved):
Looks great! Just some minor comments, but not a blocker

tcoratger (approved):
I like it, much better!

field/src/helpers.rs
Nashtare: I don't think that's an issue per say, as field elements are `Copy` so `.copied()` compiles to the same that what we would have through reference (i.e. is zero-cost basically).
Although I'd prefer changing this altogether than having another `_ref` variant living concurrently (I don't think it aligns well with Rust standards either)

lookup/src/tests.rs
Nashtare: Ah yeah good catch!

Nashtare: Semantically yes, but this should be collapsing to a plain folding after monomorphisation. If we change the trait convention from ascending to descending would just move `.rev()` elsewhere, and it feels having some `horner_descending` variant for this exact case to be a bit too heavy for what it aims at doing?

adr1anh: I could've been clearer, sorry. I was making sure that the caller here explicitly was ok with using non-reversed Horner over values

field/src/helpers.rs
adr1anh: Yeah that makes sense, simpler is better. This does make me wonder if we would need `iter().duped()` for types that can't be copy.

PR #1693, opened 2026-05-22
chore: HornerIter follow-ups from #1692 review

tcoratger:
Follow-ups to #1692 caught in post-merge review. All changes are test-only or test-side cleanup; no prover/verifier semantics change.

## Summary
- **fri**: migrate the remaining `iter().rev().fold(...)` Horner evaluation in `final_poly_mismatch` so both the honest and corrupted evaluations use the new `horner` API (the PR migrated one of the two but left the other).
- **lookup**: drop the asymmetric `.map(EF::from)` on `val_provided` (the symmetric fix to commit `0056d9a`'s removal on `val_read`) and remove the now-redundant `EF::from(val_*)` identity conversions on the `alpha - …` operands.
- **field**: migrate the two remaining `iter().rev().fold(...)` recompositions in `helpers_test` — the Goldilocks limb roundtrip and the BigUint base-`p` roundtrip — to `.horner(...)`.
- **field**: add `horner_with_extension_accumulator_over_base_coefficients` pinning the `Acc = EF`, `Item = F` cross-type contract that every production call site of `HornerIter` relies on. The previous tests only exercised `Acc = Item = BabyBear`, leaving the `Algebra<F>`-mediated `Add<Self::Item>` bound untested. The new test uses an `x` with non-zero coordinates in every extension component so the embedding case is not the only one covered.

## Test plan
- [x] `cargo test -p p3-field --test helpers_test` — 38 tests pass, including the 5 Horner tests.
- [x] `cargo test -p p3-fri --lib final_poly_mismatch` — passes.
- [x] `cargo test -p p3-lookup --tests` — 17 tests pass.
- [x] `cargo clippy -p p3-field -p p3-fri -p p3-lookup --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --all` — clean.

### DIFF
--- a/field/tests/helpers_test.rs
+++ b/field/tests/helpers_test.rs
@@ -154,9 +154,10 @@ mod helpers {
         // via the original reduce_32 (which uses base 2^32) only when values are small.
         // Instead, manually reconstruct with the matching radix.
         let base = Goldilocks::from_u64(1u64 << pb);
-        let recomposed: Goldilocks = limbs.iter().rev().fold(Goldilocks::ZERO, |acc, &limb| {
-            acc * base + Goldilocks::from_u64(limb.as_canonical_u32() as u64)
-        });
+        let recomposed: Goldilocks = limbs
+            .iter()
+            .map(|limb| Goldilocks::from_u64(limb.as_canonical_u32() as u64))
+            .horner(base);
         assert_eq!(recomposed, g);
     }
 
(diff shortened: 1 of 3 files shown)


field/tests/helpers_test.rs
```diff
@@ -591,4 +593,38 @@ mod helpers {
...
     }
+
+    #[test]
+    fn horner_with_extension_accumulator_over_base_coefficients() {
+        // Production call sites have Acc = EF (extension) and Item = F (base);
+        // the bound `Acc: Add<Self::Item, Output = Acc>` is satisfied via the
+        // `Algebra<F>` supertrait of `ExtensionField`. Pin that contract here
+        // with an `x` that exercises all extension coordinates (not just the
+        // base embedding).
+        use p3_field::BasedVectorSpace;
```
Nashtare: nit: import at the test module start

Nashtare (approved):
Thank you!

field/tests/helpers_test.rs
tcoratger: yep fixed

PR #1694, opened 2026-05-22
Add argument size in WHIR example

WizardOfMenlo:


tcoratger:
@Nashtare Do you have a better, non-invasive idea to do this? 

I think bincode crate is archived right now, so I'm wondering if it's a good idea to introduce such a repository. I have some ideas about manually implementing this, but it requires many more lines of code... And I agree that this is very useful information to have.

WizardOfMenlo:
I have heard good things about the postcard crate, might be a good alternative?

tcoratger:
> I have heard good things about the postcard crate, might be a good alternative?

Yeah and it is already available in the codebase so no new import. 

But I've tested locally and for some reason the displayed size is a bit larger with this crate than with bincode. I suspect something around pruning or maybe a bad configuration on my side. But if you can make it work, that would be perfect :)

tcoratger:
I've switched to postcard and I've documented the small overhead for the reader.

tcoratger (approved):

Nashtare:
Apologies, I had missed the ping. Agree with postcard though, this seems like the best solution.

PR #1696, opened 2026-05-23
refactor(field): unify extension-field logic under common abstraction

Nashtare:
Unifies the three parallel families of extension-field machinery (binomial / cubic / quintic) into one single abstraction parameterized by a `Shape` marker type.
  
The public API is preserved via type aliases so downstream impls don't require changes (only trait implementors).

LMK what you think!


tcoratger (approved):
That is great I love this and I think it will simplify further integrations in the future. 

For example if we want somehow to add binary fields in the future, this way this is much simpler.

PR #1697, opened 2026-05-23
refactor(field): default ext_square to a general multiply

tcoratger:
## Summary

Follow-up cleanup on #1696. The unified `ExtensionAlgebra` trait left `ext_square` as a required method, but neither `BinomialExtensionField::square()` nor its packed counterpart route through the trait — both call the free `binomial_square` helper directly. So every binomial-shape `ext_square` override in the codebase was unreachable.

- Give `ext_square` a default body of `Self::ext_mul(a, a, res)`.
- Drop the dead binomial overrides from Goldilocks (D=2, D=5), Mersenne31, Complex (both impls), and MontyField31 (binomial + quintic-trinomial).
- Keep the overrides on Goldilocks `CubicTrinomial` and Goldilocks `QuinticTrinomial` impls — those route to `cubic_square` / `quintic_square`, dedicated symmetry-exploiting kernels actually reached from `CubicTrinomialExtensionField::square()` / `QuinticTrinomialExtensionField::square()`.

Net: -20 lines, zero behavioural change.

## Test plan

- [x] `cargo check --all-targets` clean.
- [x] `cargo test -p p3-field -p p3-goldilocks -p p3-mersenne-31 -p p3-baby-bear -p p3-koala-bear --lib` → 188/188 passing.

### DIFF
--- a/field/src/extension/complex.rs
+++ b/field/src/extension/complex.rs
@@ -1,6 +1,6 @@
 use super::{
     Binomial, BinomialExtensionField, BinomiallyExtendable, ExtensionAlgebra,
-    HasTwoAdicBinomialExtension, binomial_mul, binomial_square,
+    HasTwoAdicBinomialExtension, binomial_mul,
 };
 use crate::{Algebra, Field, PrimeCharacteristicRing};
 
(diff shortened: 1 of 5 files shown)


Nashtare (approved):
Good catch, thanks!

PR #1699, opened 2026-05-24
feat(sumcheck): add generic-degree sumcheck driver

tcoratger:
## Summary

- Adds `sumcheck::generic_degree`: a sibling sumcheck module supporting any per-variable degree, transmitting `[h(0), h(2), …, h(degree)]` per round and recovering `h(1)` from the sumcheck invariant.
- The claimed sum is carried in the proof (the verifier consumes the whole prover output) and absorbed into the transcript before any challenge, so the round challenges depend on the statement.
- Round-polynomial evaluation reuses the workspace barycentric Lagrange helper (`p3_matrix::interpolation`), precomputing the weights once over the fixed integer domain `0, …, degree` and reusing them across rounds.
- The existing degree-2 path (`SumcheckData` / `ProductPolynomial`) used by WHIR is left untouched.

## API

- `RoundProver<EF>` — prover-state trait: `fold(&mut self, r)` binds a variable, `round_poly(&self)` returns the round evaluations. `prove(...)` is a default method on the trait.
- `GenericDegreeProof<F, EF>` — the transcript record; `verify(...)` is an inherent method that reads the claimed sum from the proof. It returns the sampled challenges and the final reduced sum for the caller to close against a PCS opening or a structural multilinear.
- `GenericDegreeError` — typed verifier errors (round count, per-round eval count, PoW witness count/validity, zero degree).

## Layout

```
sumcheck/src/generic_degree/
├── mod.rs     — submodule declarations + public re-exports
├── error.rs   — GenericDegreeError
├── proof.rs   — GenericDegreeProof + verify()
├── prover.rs  — RoundProver trait + default prove()
└── util.rs    — precomputed barycentric round-polynomial interpolation
```

## Test plan

- [x] `cargo test -p p3-sumcheck` — full suite passes (8 new tests in `generic_degree::{util, proof, prover}`)
  - prove/verify round-trip for a degree-3 product sumcheck
  - prove/verify round-trip with proof-of-work grinding enabled
  - verifier rejects a tampered PoW witness (asserting the failing round)
  - verifier rejects wrong round count, zero degree, and unexpected PoW witnesses
  - round-poly interpolation matches direct evaluation, including on-domain challenges
- [x] `cargo test -p p3-whir` — no regression on the WHIR degree-2 hot path
- [x] `cargo clippy -p p3-sumcheck --all-targets` — clean


latifkasuli:
I pulled this locally and had a look. Overall I like the direction.

The split as a sibling `sumcheck::multi_round` module feels right to me: it makes generic-degree sumcheck reusable without disturbing the existing WHIR degree-2 path or the optimized `SumcheckData` / `ProductPolynomial` flow. Reusing `InterpolateArbitrary` also seems cleaner than adding another local Lagrange implementation.

The only things I’d want reviewers to sanity-check are API/edge-case details rather than the main idea:

- Should `verify` reject `degree == 0` with a typed error instead of letting the interpolation helper panic?
- Should proofs with non-empty `pow_witnesses` be rejected when `pow_bits == 0`, for canonical proof shape?
- Does `verify` really need the `TwoAdicField` bound, or can it match the weaker bounds used by `prove`?

None of those change my view on the PR direction. The main design looks useful and well-scoped.

tcoratger:
> I pulled this locally and had a look. Overall I like the direction.
> 
> The split as a sibling `sumcheck::multi_round` module feels right to me: it makes generic-degree sumcheck reusable without disturbing the existing WHIR degree-2 path or the optimized `SumcheckData` / `ProductPolynomial` flow. Reusing `InterpolateArbitrary` also seems cleaner than adding another local Lagrange implementation.
> 
> The only things I’d want reviewers to sanity-check are API/edge-case details rather than the main idea:
> 
> * Should `verify` reject `degree == 0` with a typed error instead of letting the interpolation helper panic?
> * Should proofs with non-empty `pow_witnesses` be rejected when `pow_bits == 0`, for canonical proof shape?
> * Does `verify` really need the `TwoAdicField` bound, or can it match the weaker bounds used by `prove`?
> 
> None of those change my view on the PR direction. The main design looks useful and well-scoped.

Thanks a lot! Fixed

TomWambsgans:
soundness lgtm

tcoratger:
@kilic Do you need it to implement Logup+GKR properly?

sumcheck/src/multi_round/prover.rs
```diff
@@ -0,0 +1,243 @@
...
+    ///
+    /// Panics if `degree` is zero.
+    fn prove<F, Challenger>(
+        &mut self,
+        challenger: &mut Challenger,
+        num_rounds: usize,
+        degree: usize,
+        pow_bits: usize,
+        mut claimed_sum: EF,
+    ) -> (MultiRoundProof<F, EF>, Point<EF>)
```
Nashtare: I don't find the API particularly clear w.r.t `claimed_sum` taken as mutable arg that must be manually fed to the verifier alongside the return type. I think baking it in would be better so that the verifier takes the whole prover output.

sumcheck/src/generic_degree/prover.rs
```diff
@@ -0,0 +1,243 @@
...
+            proof.round_polys.push(evals);
+            challenges.push(challenge);
+        }
+
+        (proof, Point::new(challenges))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Nashtare, line 105: nit: I don't think we're testing the grinding in those tests

sumcheck/src/multi_round/util.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    // Domain: integer nodes lifted into the extension field.
+    let x_coords: Vec<EF> = (0..=d).map(EF::from_usize).collect();
+
+    // Build a one-column matrix view since the workspace interpolator works on matrices.
+    //
+    // Integer domain points 0, 1, …, d are pairwise distinct, so the helper never returns `None`.
+    RowMajorMatrix::new_col(full)
+        .interpolate_arbitrary_point(&x_coords, r)
+        .expect("integer nodes are pairwise distinct")[0]
```
Nashtare: Shouldn't we be using `interpolate_arbitrary_with_precomputation` here for all invocations of `evaluate_round_poly_at()`?

sumcheck/src/generic_degree/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! Generic-degree sumcheck driver.
```
Nashtare, line 1: nit: multi-round sounds weird to me, every sumcheck is multi-round. Perhaps rename it to `generic_degree`?

sumcheck/src/multi_round/proof.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// - PCS openings for committed multilinears.
+    /// - Closed-form evaluation for structural multilinears (`eq`, `next`, selectors).
+    pub fn verify<Challenger>(
+        &self,
+        challenger: &mut Challenger,
+        num_rounds: usize,
+        degree: usize,
+        pow_bits: usize,
+        mut claimed_sum: EF,
+    ) -> Result<(Point<EF>, EF), MultiRoundError>
```
Nashtare: same here perhaps about the API and `mut claimed_sum`

Nashtare: also the `claimed_sum` is not absorbed in the transcript is it? Can't that be an issue?

sumcheck/src/multi_round/prover.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// - Called exactly once between consecutive round-polynomial requests.
+    /// - In-place updates are recommended.
+    fn fold(&mut self, r: EF);
+
+    /// Return the current round polynomial as a list of evaluations.
+    ///
+    /// # Returns
+    ///
+    /// Vector of length `degree` containing `h(0), h(2), h(3), ..., h(degree)`.
+    fn round_poly(&mut self) -> Vec<EF>;
```
Nashtare: why does it need to be taking a mutable ref?

Nashtare (commented):
Looking overall good, but I think we need some more tightening.

Side note though, not sure if it's a difference in prompting / model / rules but the verbosity / comment density feels even higher than before, and it may lead to comment rot easily as we iterate / update the codebase, and a lot of those aren't particularly helpful (one-liners within functions explaining what a line does while it's clearly visible). May be worth toning it down a little. Detailed rustdoc is great though for the public API.

sumcheck/src/multi_round/prover.rs
tcoratger: Just fixed by adding `claimed_sum` field to the proof. So the `prove` method takes `claimed_sum` which is no longer mutable and stores it in the proof. `verify` dropped the `claimed_sum` parameter and reads it from `self.claimed_sum`, consuming the whole prover output.

sumcheck/src/generic_degree/mod.rs
tcoratger: Fixed

sumcheck/src/multi_round/proof.rs
tcoratger: Fixed both the comment, both prove and verify now `observe_algebra_element(claimed_sum) ` before sampling any challenge.

sumcheck/src/multi_round/util.rs
tcoratger: Yep good point, fixed

sumcheck/src/generic_degree/prover.rs
tcoratger: Just added `end_to_end_prove_verify_with_grinding` (`pow_bits = 4`) and `verify_rejects_tampered_pow_witness` where we corrupt the last round witness so that we expect a failure.

sumcheck/src/multi_round/prover.rs
tcoratger: Just changed so that the round poly method now reads the current state to produce the round polynomial and never mutates.

sumcheck/src/multi_round/prover.rs
Nashtare: Yeah I think that's looking better, thanks!

tcoratger:
> Looking overall good, but I think we need some more tightening.
> 
> Side note though, not sure if it's a difference in prompting / model / rules but the verbosity / comment density feels even higher than before, and it may lead to comment rot easily as we iterate / update the codebase, and a lot of those aren't particularly helpful (one-liners within functions explaining what a line does while it's clearly visible). May be worth toning it down a little. Detailed rustdoc is great though for the public API.

Thanks a lot for the thorough review! I've tweaked the documentation to simplify things and remove some lines.

I have a Claude skill for documentation in Rust because I've noticed that when I include this type of documentation (I did many tries before having something satisfying), it's easier for me to reread the code line by line. :) Also, I've noticed that when the agent runs over it, it reads all the details very well and is able to find bugs or improvements that it sometimes missed with lighter documentation.

But I agree we had a couple of useless lines there that I've tried to remove as many as I can through a manual pass.

sumcheck/src/generic_degree/proof.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[test]
+    fn verify_rejects_wrong_round_count() {
+        // Fixture: an empty proof, but the verifier expects 2 rounds.
+        //
+        // Mutation: pass num_rounds = 2 against a 0-round proof.
+        //
+        //     proof rounds: 0
+        //     expected:     2
+        //     -> RoundCountMismatch
```
Nashtare: not sure all these Fixture / Mutation etc.. here and below are meaningful tbh

Nashtare (approved):
Thanks Thomas, looking great!

sumcheck/src/generic_degree/proof.rs
tcoratger: Just simplified :)

PR #1700, opened 2026-05-24
feat(multi-stark): add crate skeleton with boundary selectors and AIR folder

tcoratger:
## Summary

Introduces the `p3-multi-stark` crate that will host the multilinear SuperSpartan-flavored AIR prover.

This PR lands the crate skeleton plus the two foundation pieces:

- the boundary selectors the AIR verifier evaluates at the random sumcheck challenge,
- the AIR constraint folder that plugs those selectors into the standard `p3_air::AirBuilder` so any AIR runs unchanged through the multilinear path.

## Public surface

- `selectors::BoundaryEvals { first, last, transition }` with `BoundaryEvals::at(rs)` — bundles the three structural selectors the AIR verifier needs at the random point recovered after the outer sumcheck. Computing them together amortizes the pass over the challenge coordinates and ties the three values to the same point at the type level.
- `folder::MultilinearFolder<'a, F, EF>` — implements `p3_air::AirBuilder`. Each `assert_zero` Horner-folds the constraint into a running alpha-batched accumulator. The same folder is driven by the prover (at every cube vertex) and by the verifier (at the random challenge point), so the source `Air::eval` runs once and unchanged in both directions.
- `folder::evaluate_at_row(air, local, next, boundary, public_values, alpha)` — convenience wrapper that constructs a folder, runs `air.eval(&mut folder)`, and returns the alpha-batched accumulator.

## Test plan

- [x] `cargo test -p p3-multi-stark` — 4 unit tests pass
  - `selectors::boundary_evals_at_corners` (exhaustive on `k = 3`)
  - `folder::folder_accumulator_is_zero_on_satisfied_rows` (8-row Fibonacci trace; folder returns zero at every row)
  - `folder::folder_detects_a_bad_first_row` (flipped public input; accumulator non-zero)
  - `folder::folder_alpha_batching_matches_horner_fold` (hand-computed Horner fold matches the folder bit-for-bit)
- [x] `cargo build --workspace` — clean

## References

- [Setty, Thaler, Wahby. Customizable Constraint Systems for succinct arguments.](https://eprint.iacr.org/2023/552.pdf)
- [Borgeaud, W. AIR-specific optimizations on top of SuperSpartan.](https://solvable.group/posts/super-air/)


multi-stark/src/folder.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// Public inputs forwarded to the AIR.
+    pub public_values: &'a [F],
+    /// Random scalar driving alpha-batching of constraints.
+    pub alpha: EF,
+    /// Running alpha-batched accumulator capturing every asserted-zero constraint.
+    pub accumulator: EF,
+    /// Two-row preprocessed window; zero-width when the AIR has no preprocessed columns.
+    pub preprocessed_window: RowWindow<'a, EF>,
+    /// Type witness for the base field; no runtime storage.
+    pub _phantom: PhantomData<F>,
```
adr1anh: Any reason why we split local and next for main trace, but have a row window for the preprocessed?

adr1anh (approved):

multi-stark/src/folder.rs
tcoratger: fixed

tcoratger:
@Nashtare Ouuups yes maybe we want to wait for after 0.6 before merging this one, right?

Nashtare:
Yeah (although you haven't included it in the `release-plz-toml` file so it wouldn't get published anyway, but I'd like to give it a look when I have time :)

multi-stark/src/folder.rs
```diff
@@ -0,0 +1,395 @@
...
+    F: Field,
+    EF: ExtensionField<F>,
+{
+    type F = F;
+    type Expr = EF;
+    type Var = EF;
+    type MainWindow = RowWindow<'a, EF>;
+    type PreprocessedWindow = RowWindow<'a, EF>;
+    type PublicVar = F;
+    type PeriodicVar = EF;
```
Nashtare, line 138: how does it handle periodic values? This defaults to the base impl. which drops periodic values entirely

multi-stark/src/folder.rs
```diff
@@ -0,0 +1,395 @@
...
+    ) -> Self {
+        Self {
+            // Pair the two borrowed rows into the window the AIR reads from.
+            main_window: RowWindow::from_two_rows(local, next),
+            boundary,
+            public_values,
+            alpha,
+            accumulator: EF::ZERO,
+            // Zero-width preprocessed window covers AIRs without a preprocessed trace.
+            preprocessed_window: RowWindow::from_two_rows(&[], &[]),
```
Nashtare, line 80: this would effectively make us not able to use preprocessed columns with the multilinear folder?

multi-stark/src/folder.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn assert_zero<I: Into<Self::Expr>>(&mut self, x: I) {
+        // Horner alpha-batching: each push updates
+        //
+        //     accumulator := alpha * accumulator + C_i
+        //
+        // After `n` pushes the accumulator collapses to
+        //
+        //     C_0 * alpha^(n-1) + C_1 * alpha^(n-2) + ... + C_{n-1}.
+        self.accumulator = self.accumulator * self.alpha + x.into();
+    }
```
Nashtare, line 175: I may be wrong here it's been a while, but this wouldn't work with schemes like SuperSpartan, would it?

Cargo.toml
```diff
@@ -28,6 +28,7 @@ members = [
     "monolith",
     "monolith-air",
     "monty-31",
+    "multi-stark",
```
Nashtare, line 31: also needs to be added to release-plz.toml

Nashtare: and to the `check_embedded` section of the CI workflow

Nashtare (changes requested):

multi-stark/src/folder.rs
tcoratger: Yes since this was a scaffolding, the folder never overrode `periodic_values()`, so the trait default returned an empty slice. Just added periodic values thing.

multi-stark/src/folder.rs
tcoratger: Just added a `with_preprocessed(current, next)` builder method so a caller can attach a real two-row preprocessed window. Still defaults to empty for the common no-preprocessed AIR.

Cargo.toml
tcoratger: Done

multi-stark/src/folder.rs
tcoratger: Mhhh I think this is the right random linear combination of constraints from the paper: https://eprint.iacr.org/2023/552.pdf

image

Nashtare (approved):
LGTM thanks!

PR #1704, opened 2026-05-26
fix(fri): type hiding PCS random opening errors

shahrukhjutt98:
## Summary

- Replace catch-all `FriError::InvalidProofShape` returns in `HidingFriPcs::verify` with typed random-opening mismatch variants.
- Report whether the mismatch is at the round, matrix, or point level, including expected and actual counts.
- Keep the inner PCS verification path unchanged after the random openings are merged.


tcoratger (commented):
If we want to properly validate, I think that we need to have small unit tests corresponding to each new failure case.

tcoratger (approved):

PR #1706, opened 2026-05-26
docs(mmcs): document power-of-two height invariant on commit

Lubov66:
The open_batch method uses bit-shift semantics to map row indices across matrices of different heights. This mapping is only correct when all heights are powers of two forming a divisibility chain. Violating this causes silent index aliasing that breaks FRI soundness. This PR makes the invariant explicit at both the trait and method level.


Nashtare:
This is implied by the FRI protocol.

Issue #1279, opened 2026-02-01
TwoAdicFriPcs instantiation


MuhtasimTanmoy:
A working FRI commit looks like this

```rust
type F = KoalaBear;
type MyChallenger = DuplexChallenger<F, Poseidon16, 16, 8>;

// Initialize the DFT (Discrete Fourier Transform) engine
let logn = 23;
let n = 1 << logn; 
let dft = Radix2DFTSmallBatch::<F>::new(n);

// Instantiate the Merkle Tree MMCS
let mut rng = SmallRng::seed_from_u64(1);
let poseidon16 = Poseidon16::new_from_rng_128(&mut rng);
let poseidon24 = Poseidon24::new_from_rng_128(&mut rng);
let merkle_hash = MerkleHash::new(poseidon24);
let merkle_compress = MerkleCompress::new(poseidon16);

let val_mmcs = Poseidon2MerkleMmcs::<F, Poseidon16, Poseidon24>::new(
    merkle_hash,
    merkle_compress,
);

// Configure FRI params
let fri_params = FriParameters {
    log_blowup: 1,
    log_final_poly_len: 0,
    num_queries: 90,
    commit_proof_of_work_bits: 8,
    query_proof_of_work_bits: 8,
    mmcs: val_mmcs.clone(), 
};

// Instantiate the PCS
let pcs = TwoAdicFriPcs::<F, DFT, _, _>::new(dft, val_mmcs, fri_params);

let domain = TwoAdicMultiplicativeCoset::new(F::GENERATOR, logn) .expect("log height too large");

// Init the polynomial data
let mut poly_rng = StdRng::seed_from_u64(0);
let polynomial = EvaluationsList::<F>::new((0..n).map(|_| poly_rng.random()).collect());

// This pairs the domain with the matrix and wraps it in an iterator.
let matrix_iter = std::iter::once((
    domain,
    RowMajorMatrix::new(polynomial.as_slice().to_vec(), 1),
));

// --- Commitment Call ---
// 9. Commit to the matrix 
let (commitment, prover_data) = Pcs::<F, MyChallenger>::commit(&pcs, matrix_iter.clone());
```

In order to

```rust
// convert this
let (commitment, prover_data) = Pcs::<F, MyChallenger>::commit(&pcs, matrix_iter.clone());

// to 
let (commitment, prover_data) = pcs.commit(matrix_iter.clone());
```

What needs to be done?  
How to pass `F, MyChallenger` then?

[...]

PR #1707, opened 2026-05-26
test(fri): clarify valid Pcs::commit call styles

peter941221:
This addresses #1279.

The issue is not an implementation bug in `TwoAdicFriPcs`, but a user-facing confusion around where Rust can infer the `Pcs<Challenge, Challenger>` trait parameters for `commit`.

This PR keeps the API unchanged and instead adds executable clarification.

It adds a focused test showing the two valid call styles on the same PCS instance:
- `pcs.commit(...)` when the trait parameters are inferable
- `<MyPcs as Pcs<Challenge, Challenger>>::commit(&pcs, ...)` when explicit UFCS is needed

It also adds a short note on `Pcs::commit` documenting that distinction.

Why this shape:
- `Pcs::commit` is defined on the trait, not on `TwoAdicFriPcs` itself
- the repository already uses both styles today in `fri/tests/pcs.rs`
- https://github.com/Plonky3/Plonky3/issues/1279#issuecomment-3831899479 already points to explicit UFCS as the expected fallback

Why the test uses a helper for the method-call case:
- raw `pcs.commit(...)` in arbitrary test scope can hit the same ambiguity as the original issue
- the helper establishes a context where the relevant `Pcs<Challenge, Challenger>` implementation is already fixed
- that lets the test demonstrate the intended method-call form without changing API shape

Non-goals:
- no API change to `TwoAdicFriPcs`
- no extra generic parameters added to the struct
- no wrapper convenience method

Testing:
- `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo test -p p3-fri --test pcs`

### DIFF
--- a/commit/src/pcs/univariate.rs
+++ b/commit/src/pcs/univariate.rs
@@ -83,6 +83,10 @@ where
     /// the polynomials defined by those evaluations. If `zk` is enabled, the evaluations are
     /// first randomized as explained in Section 3 of <https://eprint.iacr.org/2024/1037.pdf>.
     ///
+    /// Method-call syntax like `pcs.commit(...)` works when Rust can infer the
+    /// `Pcs<Challenge, Challenger>` parameters. Otherwise use explicit UFCS, e.g.
+    /// `<MyPcs as Pcs<Challenge, Challenger>>::commit(&pcs, ...)`.
+    ///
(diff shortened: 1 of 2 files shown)


Nashtare (commented):
This doesn't bring much value.

PR #1709, opened 2026-05-27
fix(symbolic): count circle transition selector degree

peter941221:
## Summary

This fixes a mismatch between the symbolic selector-degree model and the circle STARK quotient-degree analysis.

`BaseLeaf::IsTransition` was still assigned symbolic degree 0, while `uni-stark` already treats the transition selector as contributing a `+1` term in the quotient-degree bound. That left the symbolic model inconsistent with the prover-side degree analysis.

This PR brings the symbolic model back in line with the prover/verifier reasoning.

## Evidence

- `air/src/symbolic/expression.rs:46` now counts `IsTransition` together with `IsFirstRow` and `IsLastRow` as degree `1`.
- `air/src/symbolic/expression.rs:234-238` updates the unit-test expectation to `"IsTransition should have degree 1"`.
- `uni-stark/src/prover.rs:116` already documents the extra `+1` term coming from `is_transition` in the quotient-degree bound.
- `uni-stark/src/symbolic.rs:267` adds the matching regression guard: a selector-only AIR with a zero degree hint must panic in debug builds.

So the primary validation axis here is semantic consistency plus regression coverage, not performance.

## Semantic change

- count `IsTransition` as degree `1` in the symbolic expression model
- update the matching symbolic test expectation
- add a `uni-stark` regression test for the selector-only hint-underflow case

## Testing

- `cargo fmt --all --check`
- `cargo test -p p3-air symbolic -- --nocapture`
- `cargo test -p p3-uni-stark symbolic -- --nocapture`

### DIFF
--- a/air/src/symbolic/expression.rs
+++ b/air/src/symbolic/expression.rs
@@ -43,8 +43,8 @@ impl<F: Field> SymLeaf for BaseLeaf<F> {
     fn degree_multiple(&self) -> usize {
         match self {
             Self::Variable(v) => v.degree_multiple(),
-            Self::IsFirstRow | Self::IsLastRow => 1,
-            Self::IsTransition | Self::Constant(_) => 0,
+            Self::IsFirstRow | Self::IsLastRow | Self::IsTransition => 1,
+            Self::Constant(_) => 0,
         }
     }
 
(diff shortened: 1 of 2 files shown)


peter941221:
I don't think a benchmark is the right validation axis for this one.

The bug here is a semantic mismatch between the symbolic selector-degree model and the quotient-degree analysis already used by `uni-stark`:

- `air/src/symbolic/expression.rs:46` now counts `IsTransition` as degree 1
- `uni-stark/src/prover.rs:116` already accounts for a `+1` term coming from `is_transition`
- `uni-stark/src/symbolic.rs:267` adds the matching regression guard on the hint path

So the evidence is correctness + consistency, not performance. The symbolic model now matches the prover-side quotient bound, and the focused symbolic tests pass cleanly.

Nashtare (commented):
This is incorrect, and Circle STARKs run through the same `prove/get_log_num_quotient_chunks` path as `uni-stark`.

Issue #1382, opened 2026-03-03
Improve degree calculation


adr1anh:
Periodic columns highlighted that the degree computation in SymbolicAir can be slightly off depending on the size of the trace, which it currently doesn't take into account. In particular,
- it treats the transition selector as a constant, when it is actually linear
- it over estimates the degree of periodic columns.

See https://github.com/Plonky3/Plonky3/pull/1380#discussion_r2878011612

peter941221:
I opened a narrow v1 PR for this:

- `https://github.com/Plonky3/Plonky3/pull/1518`

I kept the scope intentionally tight:

- keep the existing symbolic-expression API unchanged
- keep the current `max_constraint_degree()` hint contract unchanged
- only tighten the inferred no-hint quotient-degree / quotient-chunk path in `uni-stark`

Concretely, the PR now:

- treats `is_transition` as linear for quotient-degree purposes
- uses periodic-column degree `n - n/p`
- threads the original trace degree through prover / verifier chunk inference
- adds targeted regression tests around the quotient-chunk bucket-drop cases

Validation on the branch:

- `cargo test -p p3-uni-stark -- --nocapture`
- `cargo test -p p3-air -- --nocapture`
- `cargo test -p p3-commit -- --nocapture`
- `cargo test -p p3-circle periodic -- --nocapture`
- `cargo test -p p3-fri periodic -- --nocapture`

If you'd prefer the trace-aware periodic-degree logic to live at a different abstraction boundary, I am happy to follow that direction and reshape the PR.

PR #1710, opened 2026-05-27
fix(uni-stark): tighten periodic quotient inference

peter941221:
## Summary

This addresses the periodic-column half of #1382.

The old no-hint path in `uni-stark` treated every periodic column like a generic degree-1 symbolic leaf. That is intentionally only an approximation. For periodic-heavy AIRs, it can keep quotient-degree inference in a more conservative quotient-chunk bucket than necessary.

This PR threads the original trace degree into no-hint quotient inference and uses the trace-aware periodic-column degree `n - n/p` instead.

## Scope

- keep the existing symbolic-expression API unchanged
- keep the current `max_constraint_degree()` hint contract unchanged
- tighten only the inferred no-hint quotient-degree / quotient-chunk path in `uni-stark`

## Failure mechanism

- `air/src/symbolic/variable.rs` still gives `BaseEntry::Periodic` a fixed symbolic degree multiple of `1`
- `uni-stark` was previously using that approximation directly on the no-hint path
- for periodic-heavy AIRs, that can keep the prover/verifier on a larger quotient-chunk bucket than needed

## Semantic change

- `uni-stark/src/symbolic.rs:91-99` computes per-column periodic degrees as `trace_degree - trace_degree / period`
- `uni-stark/src/symbolic.rs:202` uses those trace-aware values for `BaseEntry::Periodic`
- `uni-stark/src/prover.rs:125` and `uni-stark/src/verifier.rs:294` now pass trace degree into `get_log_num_quotient_chunks`
- explicit `max_constraint_degree()` hints stay on the old compatibility path

## Bench

Synthetic periodic-heavy AIR, Criterion `estimates.json`, same harness on base and PR worktrees:
`cargo bench -p p3-uni-stark --bench periodic_degree`

Inference itself is slower because it now computes trace-aware periodic degrees:

| case | base infer | PR infer | Δ |
|---|---:|---:|---:|
| `period2_x5, 2^14` | `521.23 ns` | `724.04 ns` | `+38.91%` |
| `period4_x6, 2^14` | `782.85 ns` | `1045.47 ns` | `+33.55%` |
| `period8_x9, 2^14` | `792.40 ns` | `1241.71 ns` | `+56.70%` |
| `period2_x5, 2^16` | `575.36 ns` | `733.96 ns` | `+27.57%` |
| `period4_x6, 2^16` | `598.94 ns` | `865.20 ns` | `+44.46%` |
| `period8_x9, 2^16` | `979.93 ns` | `1415.86 ns` | `+44.49%` |

That is not the point of the optimization. The actual target is the downstream prover cost after quotient-chunk inference becomes tighter:

| case | base prove | PR prove | Δ |
|---|---:|---:|---:|
| `period2_x5, 2^14` | `541.06 ms` | `436.10 ms` | `-19.40%` |
| `period4_x6, 2^14` | `731.46 ms` | `528.47 ms` | `-27.75%` |
| `period8_x9, 2^14` | `736.04 ms` | `716.28 ms` | `-2.68%` |
| `period2_x5, 2^16` | `2.1666 s` | `1.7464 s` | `-19.39%` |
| `period4_x6, 2^16` | `2.9497 s` | `2.1104 s` | `-28.45%` |
| `period8_x9, 2^16` | `2.9627 s` | `2.8823 s` | `-2.71%` |

The shape split matches the chunk-bucket analysis exactly:

- `period2_x5` and `period4_x6` cross into a smaller quotient-chunk bucket, so prove time drops materially
- `period8_x9` stays in the same bucket, so only a small residual difference remains

That bucket behavior is now locked in by tests:

- `uni-stark/src/symbolic.rs:434` -> period-2 case gives `log_chunks = 1`
- `uni-stark/src/symbolic.rs:445` -> period-4 case gives `log_chunks = 2`
- `uni-stark/src/symbolic.rs:456` -> period-8 x9 case stays at `log_chunks = 3`

## Testing

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p p3-air symbolic -- --nocapture`
- `cargo test -p p3-uni-stark symbolic -- --nocapture`
- `cargo test -p p3-uni-stark --test periodic_air -- --nocapture`
- `cargo bench -p p3-uni-stark --bench periodic_degree --no-run`

### DIFF
--- a/uni-stark/Cargo.toml
+++ b/uni-stark/Cargo.toml
@@ -40,8 +40,13 @@ p3-merkle-tree = { path = "../merkle-tree" }
 p3-mersenne-31 = { path = "../mersenne-31" }
 p3-symmetric = { path = "../symmetric" }
 
+criterion.workspace = true
 postcard = { workspace = true, features = ["alloc"] }
 rand.workspace = true
+
+[[bench]]
+name = "periodic_degree"
+harness = false
 
 [lints]
 workspace = true
--- a/uni-stark/benches/periodic_degree.rs
+++ b/uni-stark/benches/periodic_degree.rs
@@ -1,0 +1,176 @@
+use core::hint::black_box;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_air::WindowAccess;
+use p3_air::symbolic::AirLayout;
+use p3_air::{Air, AirBuilder, BaseAir};
+use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+use p3_challenger::DuplexChallenger;
+use p3_commit::ExtensionMmcs;
+use p3_dft::Radix2DitParallel;
+use p3_field::extension::BinomialExtensionField;
+use p3_field::{Field, PrimeCharacteristicRing};
+use p3_fri::{FriParameters, TwoAdicFriPcs};
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
+use p3_uni_stark::{StarkConfig, get_log_num_quotient_chunks, prove};
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+type Val = BabyBear;
+type Challenge = BinomialExtensionField<Val, 4>;
+type Perm = Poseidon2BabyBear<16>;
+type MyHash = PaddingFreeSponge<Perm, 16, 8, 8>;
+type MyCompress = TruncatedPermutation<Perm, 2, 8, 16>;
+type ValMmcs =
+    MerkleTreeMmcs<<Val as Field>::Packing, <Val as Field>::Packing, MyHash, MyCompress, 2, 8>;
+type ChallengeMmcs = ExtensionMmcs<Val, Challenge, ValMmcs>;
+type Challenger = DuplexChallenger<Val, Perm, 16, 8>;
+type Dft = Radix2DitParallel<Val>;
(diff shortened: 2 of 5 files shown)


peter941221:
I benchmarked this one with a periodic-heavy synthetic AIR that isolates the exact path changed in this PR.

The important result is not that quotient-degree inference itself gets faster. It doesn't. The new inference does more work up front because it computes trace-aware periodic degrees.

The reason this PR still wins is downstream: the old no-hint path treated every periodic column like a generic degree-1 symbolic leaf, which can keep `uni-stark` in an unnecessarily large quotient-chunk bucket. This PR replaces that with the trace-aware degree `n - n/p`, so periodic-heavy AIRs can drop into a smaller bucket and do less prover work.

From `criterion/.../new/estimates.json` in `uni-stark/benches/periodic_degree.rs`:

- `period2_x5, 2^14`: `541.06 ms -> 436.10 ms` (`-19.40%`)
- `period4_x6, 2^14`: `731.46 ms -> 528.47 ms` (`-27.75%`)
- `period8_x9, 2^14`: `736.04 ms -> 716.28 ms` (`-2.68%`)
- `period2_x5, 2^16`: `2.1666 s -> 1.7464 s` (`-19.39%`)
- `period4_x6, 2^16`: `2.9497 s -> 2.1104 s` (`-28.45%`)
- `period8_x9, 2^16`: `2.9627 s -> 2.8823 s` (`-2.71%`)

That shape split matches the chunk-bucket analysis exactly:

- `period2_x5` and `period4_x6` cross into a smaller quotient-chunk bucket, so the prover win is large
- `period8_x9` stays in the same bucket, so only a small residual difference remains

So for this PR the evidence is sufficient: the extra inference cost is real but small, and it is paid back by materially lower prover work in the periodic-heavy no-hint cases this change is targeting.

Issue #529, opened 2024-10-31
Remove an extra # column multiplications from interpolate_coset through a trick with batch inversion


ValarDragon:
Currently the [interpolate_coset](https://github.com/scipr-lab/libiop/blob/master/libiop/protocols/ldt/fri/fri_aux.tcc#L230-L231) routine is a key performance bottleneck (The "compute opened values with Lagrange interpolation" step of https://gist.github.com/dlubarov/0eb8490f659a5aa9970cf95cfe1b0fb5 )

Spent alot of time optimizing this exact operation in libiop before haha.

I tried counting the time the # of CPU operations cost by hand. If the matrix is of dimensions N = R*C, currently we are doing `7C + N multiplications` and `C + N additions`. R should be quite small (between 1-4 I believe), so the C term matters to optimize.

I believe there is `C multiplications` that can be saved for free (no tradeoff involved), and `C multiplications` that can be saved with a memory tradeoff, which I suspect should be time savings, but needs benchmarking.

The C multiplications thats a memory tradeoff is basically caching the column subgroup evaluations `g^i` as we compute it twice. (And may even recompute elsewhere in the total computation of the program).

I'll go into more detail of the C multiplications that can be saved for free. At the end of this program, we need to multiply every single element in the output column by a constant. Lets call this shift "final_constant". This is what is happening here with C multiplications in `scale_vec`: https://github.com/Plonky3/Plonky3/blob/049c413fcd0ad7288d12789e4a727d3965340ca1/interpolation/src/lib.rs#L60-L62 

Notice that our computation is roughly:
```python
# C multiplications, C subtractions
compute_diffs()
# 3C multiplications
diff_inv = inverse(diffs)
# 2C multiplications
col_scale = [g^i * diff_inv[i] for i in cols]
# N multiplications, N additions
sum = columnwise_prod(matrix, col_scale)
# C multiplications
return scale_vec(sum, final_constant)
```
First notice that due to commutativity, we could instead compute this as:
```python
diff_inv = inverse(diffs)
scaled_diff_inf = scale_vec(diff_inv, final_constant)
col_scale = [g^i * scaled_diff_inf[i] for i in cols]
sum = columnwise_prod(matrix, col_scale)
return sum
```

[...]

PR #1711, opened 2026-05-27
perf(matrix): specialize base-field interpolate_coset

peter941221:
This addresses #529.

Failure mechanism:
- `matrix::Interpolate::interpolate_coset` was still routing base-field opening points through the generic extension-field weight path
- that path builds `Vec<EF>` denominator inverses and feeds `columnwise_dot_product` with extension weights, so the hot loop pays `EF::ExtensionPacking` multiply costs even when the point and all adjusted weights are actually in the base field

Semantic change:
- keep `batch_multiplicative_inverse_scaled` as the shared fused scaling primitive in `p3-field`
- specialize `matrix::Interpolate::interpolate_coset` for the `point.as_base().is_some()` case
- in that branch, compute diffs, scaled inverses, adjusted weights, and the columnwise dot product entirely in `F`, then lift the final column results back into `EF`
- keep the generic extension-point path unchanged
- keep `interpolate_coset_with_precomputation` semantics unchanged for existing callers

Preserved invariants:
- on-domain points still short-circuit to the matching row in both the base-point and generic extension-point paths
- genuine extension points still use the original `EF` path
- the precomputation API still expects the same adjusted-weight semantics as before

Bench

`RUSTFLAGS=-Ctarget-cpu=native cargo bench -p p3-matrix --bench interpolation interpolate_coset --features p3-maybe-rayon/parallel,p3-util/parallel`, Windows 11 Pro for Workstations, Intel Core Ultra 7 270K Plus, criterion 20 samples, mean point estimate.

| shape | before | after | Δ |
| --- | ---: | ---: | ---: |
| `2^10 x 1` | 176.69 us | 111.22 us | -37.0% |
| `2^10 x 128` | 186.18 us | 113.32 us | -39.1% |
| `2^14 x 16` | 341.17 us | 195.18 us | -42.8% |
| `2^18 x 1` | 3.76 ms | 1.94 ms | -48.3% |
| `2^18 x 128` | 5.17 ms | 3.24 ms | -37.4% |

This benchmark uses the existing `matrix/benches/interpolation.rs` `interpolate_coset/full/*` workload, which opens at `EF::from_u32(123456789)`. That point lies in the base field, so these numbers measure the exact regime this specialization targets.

Tests:
- `cargo test -p p3-matrix interpolation -- --nocapture`
- `cargo test -p p3-fri -- --nocapture`
- `cargo fmt --all`
- `RUSTFLAGS=-Ctarget-cpu=native cargo bench -p p3-matrix --bench interpolation --no-run --features p3-maybe-rayon/parallel,p3-util/parallel`

### DIFF
--- a/matrix/src/interpolation.rs
+++ b/matrix/src/interpolation.rs
@@ -122,6 +122,28 @@ pub trait Interpolate<F: TwoAdicField>: Matrix<F> {
             .iter()
             .collect();
 
+        if let Some(point_base) = <EF as ExtensionField<F>>::as_base(&point) {
+            // Base-field points can stay on the cheaper F-weight path all the way
+            // through the SIMD dot product, then lift the final column results back into EF.
+            let diffs: Vec<F> = coset.par_iter().map(|&g| point_base - g).collect();
+
+            if let Some(i) = diffs.iter().position(|d| d.is_zero()) {
+                return self.row(i).unwrap().into_iter().map(EF::from).collect();
+            }
+
+            let scaling_factor = interpolation_scaling_factor(shift, point_base, log_height);
+            let scaled_diff_invs = batch_multiplicative_inverse_scaled(&diffs, scaling_factor);
+            let scaled_point_inv = scaling_factor * point_base.inverse();
+            let scaled_adjusted =
+                compute_adjusted_weights_with_point_inv(scaled_point_inv, &scaled_diff_invs);
+
+            return self
+                .columnwise_dot_product(&scaled_adjusted)
+                .into_iter()
+                .map(EF::from)
+                .collect();
+        }
+
         // Compute z - x_i in parallel, then batch-invert in one shot
         // (Montgomery's trick: single field inversion + O(N) multiplications).
         let diffs: Vec<EF> = coset.par_iter().map(|&g| point - g).collect();
@@ -576,6 +598,28 @@ mod tests {
         // The precomputation variant must produce the same result.
         let result = evals_mat.interpolate_coset_with_precomputation(shift, point, &adjusted);
         assert_eq!(result, vec![F::from_u16(10203)]);
+    }
+
+    #[test]
+    fn test_interpolate_coset_genuine_extension_point() {
+        let shift = F::GENERATOR;
+        let coeffs = [F::from_u32(3), F::from_u32(2), F::from_u32(1)];
+        let evals: Vec<F> = eval_poly_on_coset(&coeffs, shift, 3);
+        let evals_mat = RowMajorMatrix::new(evals, 1);
+        let point =
+            EF4::from_basis_coefficients_slice(&[F::from_u32(100), F::ONE, F::ZERO, F::ZERO])
+                .unwrap();
+
+        let result = evals_mat.interpolate_coset(shift, point);
+        let expected = eval_poly(&coeffs, point);
+        assert_eq!(result, vec![expected]);
+
+        let coset = F::two_adic_generator(3).shifted_powers(shift).collect_n(8);
+        let diffs: Vec<EF4> = coset.iter().map(|&x| point - x).collect();
+        let diff_invs = batch_multiplicative_inverse(&diffs);
+        let adjusted = compute_adjusted_weights(point, &diff_invs);
+        let precomputed = evals_mat.interpolate_coset_with_precomputation(shift, point, &adjusted);
+        assert_eq!(result, precomputed);
     }
 
     #[test]
--- a/matrix/src/interpolation.rs
+++ b/matrix/src/interpolation.rs
@@ -86,6 +86,28 @@ fn interpolation_scaling_factor<F: TwoAdicField, EF: ExtensionField<F>>(
     point * (z_pow_n - g_pow_n) * denom_inv
 }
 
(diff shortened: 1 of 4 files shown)


tcoratger:
Hey,

Thanks a lot for tackling this, do you have some benchmark numbers with criterion before and after to check the improvement?

peter941221:
@tcoratger sure, will do that.

peter941221:
I reran this from scratch in isolated worktrees, with no other Rust benches running, on Windows 11 Pro for Workstations / Intel Core Ultra 7 270K Plus (24 cores / 24 threads), using:

RUSTFLAGS=-Ctarget-cpu=native
cargo bench -p p3-matrix --bench interpolation --features p3-maybe-rayon/parallel,p3-util/parallel

The changed path is the one in field/src/batch_inverse.rs (batch_multiplicative_inverse_scaled) and matrix/src/interpolation.rs (interpolate_coset now folds the global scaling before columnwise_dot_product).

Bottom line first: the trick does make denominator preparation materially cheaper, but on my machine the end-to-end interpolate_coset result is only a clear win when that stage dominates. Width-heavy cases are mixed to worse.

Local root-cause split inside the PR branch:

2^18x1
old_prepare_only: 3.48 ms
new_prepare_only: 1.81 ms
old_dot_only: 331.7 us
new_dot_only: 235.8 us
old_prepare_plus_dot: 3.16 ms
new_prepare_plus_dot: 2.02 ms

2^10x128
old_prepare_only: 98.0 us
new_prepare_only: 83.2 us
old_dot_only: 43.1 us
new_dot_only: 43.9 us
old_prepare_plus_dot: 175.4 us
new_prepare_plus_dot: 166.9 us

So the prepare stage is genuinely better. For the tall/narrow case, the gain survives into prepare+dot. For the short/wide case, the prepare win is much smaller and the dot-product side is basically flat.

Against a clean baseline commit (a4a97868^ -> aa14bd24), end-to-end interpolate_coset on my machine was:

2^18x1: 3.671 ms -> 3.484 ms (-5.1%)
2^18x128: 5.597 ms -> 6.233 ms (+11.4%)
2^10x128: 154.8 us -> 232.2 us (+50.0%)
2^10x1: 190.7 us -> 215.8 us (+13.2%)
2^14x16: 291.5 us -> 382.1 us (+31.1%)

So I do not think I can honestly claim a broad end-to-end improvement from this patch as-is. The evidence I see is narrower:

1. The fused inverse/scaling step itself is better.
2. That translates into a real win when denominator preparation dominates.
3. It does not translate into a consistent full-path win across the wider benchmark shapes I tested on this machine.

peter941221:
@tcoratger I reworked the patch and reran the benchmark from a clean baseline.

The key correction is that the benchmark point in `matrix/benches/interpolation.rs` is `EF::from_u32(123456789)`, so the relevant optimization target here is the base-field-point regime, not the generic extension-point regime.

This PR now specializes `interpolate_coset` for `point.as_base().is_some()`, keeps the generic extension-point path unchanged, and the before/after benchmark in the PR body is taken from the actual `interpolate_coset/full/*` workload on a clean baseline worktree versus this branch.

On my machine that gives a consistent end-to-end improvement across all listed shapes:
- `2^10 x 1`: `176.69 us -> 111.22 us` (`-37.0%`)
- `2^10 x 128`: `186.18 us -> 113.32 us` (`-39.1%`)
- `2^14 x 16`: `341.17 us -> 195.18 us` (`-42.8%`)
- `2^18 x 1`: `3.76 ms -> 1.94 ms` (`-48.3%`)
- `2^18 x 128`: `5.17 ms -> 3.24 ms` (`-37.4%`)

So for this PR, I think the evidence is sufficient now: the implementation matches the measured regime, and the end-to-end signal is consistently positive instead of mixed.

matrix/src/interpolation.rs
```diff
@@ -107,6 +122,28 @@ pub trait Interpolate<F: TwoAdicField>: Matrix<F> {
...
+            let scaled_diff_invs = batch_multiplicative_inverse_scaled(&diffs, scaling_factor);
+            let scaled_point_inv = scaling_factor * point_base.inverse();
+            let scaled_adjusted =
+                compute_adjusted_weights_with_point_inv(scaled_point_inv, &scaled_diff_invs);
+
+            return self
+                .columnwise_dot_product(&scaled_adjusted)
+                .into_iter()
+                .map(EF::from)
+                .collect();
```
tcoratger: This part looks weird because we do everything in the base field and only at the end we convert everything the extension field for no reason.

Wouldn't it make more sense to have a separate `interpolate_coset_base` method or something which operates entirely in the base field?

peter941221: Good call. I split the base-field specialization into a dedicated helper so `interpolate_coset` now just dispatches on `point.as_base()`, while the base path stays entirely in `F` until the public `Vec<EF>` boundary. The follow-up is in 5a3fbf30.

tcoratger:
I've cleaned things a up a bit and added more tests but I cannot reproduce the benches of the PR description, maybe because my laptop is a bit old and unstable, could you double check @Nashtare ?

Nashtare:
I'm getting 33% regression on this branch.

Issue #1440, opened 2026-03-16
Investigate packing overhead in no-packing scenarios


adr1anh:
During some analysis a while ago (when Goldilocks was not yet optimized for NEON), I noticed there were some small optimization opportunities to implement a specialized scalar paths when `PACKING=1` (see https://github.com/Plonky3/Plonky3/pull/1216). 

While I only glanced at the assembly, some micro benchmarks did show improvement (the PR and the individual commits explain how to view the assembly). 

The main trick we can use here is rayon's `Either` iterator to return a different path depending on the constant. The compiler seems able to inline the correct branch and monomorphize the functions.

Nashtare:
closed by #1712

PR #1712, opened 2026-05-27
perf(matrix): specialize dense vertical packing

peter941221:
## Summary

This addresses #1440.

The real bottleneck here is the scalar no-packing case (`P::WIDTH == 1`).

The original generic `Matrix` path routes `vertically_packed_row` and `vertically_packed_row_pair` through `wrapping_row_slices()`. That is fine for arbitrary matrices, but for `DenseMatrix` it pays extra row-slice materialization even when the data is already contiguous and the scalar case only needs direct loads.

This PR specializes the dense implementation, but now does so more narrowly:

- keep a direct contiguous fast path only for `P::WIDTH == 1`
- keep the packed `P::WIDTH > 1` path on the old `wrapping_row_slices()` semantics

That shape matters because the first version of this PR improved the scalar path but also changed the packed path enough to cause a large regression in `vertically_packed_row_pair::<Packed>`. The final version keeps the intended scalar win while removing that packed regression risk.

The behavioral contract is unchanged:

- scalar vertical packing still returns the same values as before
- packed-width vertical packing still follows the same wraparound row semantics as before

## Benchmark

Command:

`RUSTFLAGS=-Ctarget-cpu=native cargo bench -p p3-matrix --bench vertical_packing --features p3-maybe-rayon/parallel,p3-util/parallel`

Machine:

- CPU: Intel Core Ultra 7 270K Plus
- OS: Microsoft Windows
- logical processors: 24
- Criterion sample size: 20
- numbers below are `mean.point_estimate` from `target/criterion/.../new/estimates.json`

### Scalar path (`P = F`)

| benchmark | shape | before | after | delta |
|---|---:|---:|---:|---:|
| `vertically_packed_row` | `2^10 x 32` | 66.215 ns | 36.049 ns | `-45.56%` |
| `vertically_packed_row` | `2^14 x 32` | 67.908 ns | 37.176 ns | `-45.26%` |
| `vertically_packed_row` | `2^18 x 32` | 68.096 ns | 37.132 ns | `-45.47%` |
| `vertically_packed_row` | `2^14 x 256` | 141.772 ns | 44.275 ns | `-68.77%` |
| `vertically_packed_row_pair` | `2^10 x 32` | 116.088 ns | 35.627 ns | `-69.31%` |
| `vertically_packed_row_pair` | `2^14 x 32` | 116.265 ns | 35.642 ns | `-69.34%` |
| `vertically_packed_row_pair` | `2^18 x 32` | 116.912 ns | 35.611 ns | `-69.54%` |
| `vertically_packed_row_pair` | `2^14 x 256` | 304.912 ns | 50.118 ns | `-83.56%` |

### Packed path (`P = <F as Field>::Packing`)

| benchmark | shape | before | after | delta |
|---|---:|---:|---:|---:|
| `vertically_packed_row` | `2^10 x 32` | 142.793 ns | 143.555 ns | `+0.53%` |
| `vertically_packed_row` | `2^14 x 32` | 143.355 ns | 143.137 ns | `-0.15%` |
| `vertically_packed_row` | `2^18 x 32` | 145.049 ns | 143.974 ns | `-0.74%` |
| `vertically_packed_row` | `2^14 x 256` | 608.783 ns | 615.302 ns | `+1.07%` |
| `vertically_packed_row_pair` | `2^10 x 32` | 284.795 ns | 287.877 ns | `+1.08%` |
| `vertically_packed_row_pair` | `2^14 x 32` | 286.744 ns | 287.104 ns | `+0.13%` |
| `vertically_packed_row_pair` | `2^18 x 32` | 287.523 ns | 291.381 ns | `+1.34%` |
| `vertically_packed_row_pair` | `2^14 x 256` | 1477.228 ns | 1474.436 ns | `-0.19%` |

Interpretation:

- the scalar no-packing target is a clear win across every tested shape
- the packed path is now effectively flat, which is the right outcome for this PR
- the earlier packed `row_pair` regression is gone once the specialization is narrowed back to `P::WIDTH == 1`

So the evidence supports the narrower optimization, not the broader one.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo bench -p p3-matrix --bench vertical_packing --no-run --features p3-maybe-rayon/parallel,p3-util/parallel`
- [x] `cargo test -p p3-matrix dense -- --nocapture`
- [x] `cargo test -p p3-merkle-tree -- --nocapture`
- [x] `cargo test -p p3-uni-stark -- --nocapture`
- [x] `cargo test -p p3-batch-stark -- --nocapture`

### DIFF
--- a/matrix/Cargo.toml
+++ b/matrix/Cargo.toml
@@ -42,5 +42,10 @@ name = "interpolation"
 path = "benches/interpolation.rs"
 harness = false
 
+[[bench]]
+name = "vertical_packing"
+path = "benches/vertical_packing.rs"
+harness = false
+
 [lints]
 workspace = true
--- a/matrix/benches/vertical_packing.rs
+++ b/matrix/benches/vertical_packing.rs
@@ -2,14 +2,14 @@ use core::hint::black_box;
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use p3_baby_bear::BabyBear;
-use p3_field::FieldArray;
+use p3_field::Field;
 use p3_matrix::Matrix;
 use p3_matrix::dense::RowMajorMatrix;
 use rand::SeedableRng;
 use rand::rngs::SmallRng;
 
 type F = BabyBear;
-type Packed = FieldArray<F, 8>;
+type Packed = <F as Field>::Packing;
 
 const CONFIGS: &[(usize, usize)] = &[(10, 32), (14, 32), (18, 32), (14, 256)];
 
@@ -55,7 +55,7 @@ fn vertically_packed_row_pair_scalar(c: &mut Criterion) {
 }
 
 fn vertically_packed_row_packed(c: &mut Criterion) {
-    let mut group = c.benchmark_group("vertically_packed_row_packed8");
+    let mut group = c.benchmark_group("vertically_packed_row_packed");
     group.sample_size(20);
 
     for &(log_rows, width) in CONFIGS {
@@ -78,7 +78,7 @@ fn vertically_packed_row_packed(c: &mut Criterion) {
 }
 
 fn vertically_packed_row_pair_packed(c: &mut Criterion) {
-    let mut group = c.benchmark_group("vertically_packed_row_pair_packed8");
+    let mut group = c.benchmark_group("vertically_packed_row_pair_packed");
     group.sample_size(20);
 
     for &(log_rows, width) in CONFIGS {
--- a/matrix/benches/vertical_packing.rs
+++ b/matrix/benches/vertical_packing.rs
@@ -1,0 +1,105 @@
+use core::hint::black_box;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_baby_bear::BabyBear;
+use p3_field::FieldArray;
+use p3_matrix::Matrix;
+use p3_matrix::dense::RowMajorMatrix;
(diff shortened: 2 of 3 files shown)


peter941221:
I reran this one with a bench that isolates the exact paths touched here.

Short version:

- the original broader specialization was not clean: it did improve the scalar no-packing path, but it also introduced a large regression in the packed `vertically_packed_row_pair::<Packed>` path
- I narrowed the optimization back to `P::WIDTH == 1` only and left `P::WIDTH > 1` on the old `wrapping_row_slices()` path
- after that change, the scalar target is still a large win and the packed path is back to effectively flat noise

So the root cause is now pretty clear: the issue is specifically scalar no-packing overhead, not vertical packing in general.

The final numbers in the PR body are from `mean.point_estimate` in Criterion `estimates.json`, not terminal summaries.

That leaves this in a much cleaner state:

- scalar `vertically_packed_row`: about `45%` faster
- scalar `vertically_packed_row_pair`: about `69%` to `84%` faster
- packed row / row_pair: roughly flat, within about `-0.19%` to `+1.34%`

For this PR, I think the evidence is sufficient now.

tcoratger:
@adr1anh I think that you worked on this topic previously right? Maybe this is interesting for you :)

adr1anh (approved):
LGTM! I had gotten stuck with the issue of returning two types of iterators, so nice!

Issue #50, opened 2023-08-08
Cache-friendly transpose in `RowMajorMatrix`


dlubarov:
Besides being a generally useful utility, this would give us an easy way to parallelize a batch FFT, by splitting it into single-column FFTs as Plonky2 does. I think this probably isn't the best way to parallelize in general, but it would be interesting to have the option and compare it with other methods.

PR #1714, opened 2026-05-27
perf(matrix): avoid prefill in transpose

peter941221:
This partially addresses #50.

Scope note:
- `#50` originally asked for cache-friendly transpose work as a building block for batch-FFT experimentation
- current `main` already has substantial transpose infrastructure:
  - `p3_util::transpose::transpose`
  - `RowMajorMatrix::transpose_into`
  - dedicated transpose benchmarks
- this PR therefore takes a narrow residual hot-path win in the current transpose entry point instead of redoing the broader algorithmic work

Failure mechanism:
- `DenseMatrix::transpose()` allocated its output with `vec![T::default(); nelts]`
- `p3_util::transpose::transpose(...)` then overwrote every output element immediately
- that default-fill pass was pure extra work on the allocation-heavy path that returns a fresh transposed matrix

Semantic change:
- allocate the output buffer for `DenseMatrix::transpose()` without pre-filling it
- let `p3_util::transpose::transpose(...)` fully initialize the buffer
- narrow the impl bound from `T: Copy + Default` to `T: Copy`, which matches what this method actually needs
- keep `transpose_into()` and the underlying transpose implementation unchanged

Why this shape:
- preserves the existing API and transpose semantics
- targets a real residual cost in the current hot path
- keeps reviewer scope to one file and one behavior-preserving optimization

Files changed:
- `matrix/src/dense.rs`

Tests:
- `cargo test -p p3-matrix transpose -- --nocapture`
- `cargo test -p p3-matrix -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/matrix/src/dense.rs
+++ b/matrix/src/dense.rs
@@ -3,7 +3,10 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::borrow::{Borrow, BorrowMut};
 use core::marker::PhantomData;
+use core::mem::ManuallyDrop;
+use core::mem::MaybeUninit;
 use core::ops::Deref;
+use core::slice;
 
 use p3_field::{
     ExtensionField, Field, PackedValue, par_scale_slice_in_place, scale_slice_in_place_single_core,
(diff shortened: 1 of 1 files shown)


tcoratger:
Can we have benchmarks over for example a small instance, a medium instance and a large instance before/after to show the gains?

peter941221:
@tcoratger sure, will do it.

tcoratger:
> @tcoratger sure, will do it.

Please can you do the same for all the other PRs? You can find here https://github.com/Plonky3/Plonky3/pull/1664 an example of nicely formatted results so that anyone can double check the performance gains.

Also I've seen that you opened couple of PRs in draft mode. To facilitate the review, please can you put them as ready to review when you feel they are ready for review cycles? :) Thanks a lot

peter941221:
@tcoratger sure, I'll do it. thanks for pointing it out. I'll do them for all PRs I opened.

peter941221:
I added a benchmark that actually measures the path this PR changes.

The existing `matrix/benches/transpose_benchmark.rs` only measures `transpose_into()`, but this PR changes `DenseMatrix::transpose()`, so that original bench was the wrong benchmark for this diff.

I added a local `transpose_allocating` group that calls `matrix.transpose()` directly, then compared baseline vs PR using Criterion `mean.point_estimate` from `target/criterion/.../new/estimates.json`.

Numbers:

- `2^10 x 2^10`: `1,021,466.825 ns -> 975,323.292 ns` (`-4.52%`)
- `2^14 x 2^10`: `25,872,604.069 ns -> 27,859,082.136 ns` (`+7.68%`)
- `2^18 x 2^8`: `99,281,415.000 ns -> 99,644,520.000 ns` (`+0.37%`)

So after aligning the benchmark with the actual changed path, the result is still mixed:

- small case improves
- medium case regresses materially
- large case is effectively flat

That means I do not have clean evidence that this is a net win for the transpose allocation path.

For that reason I do not think this PR is ready to move out of draft in its current form.

PR #1713, opened 2026-05-27
perf(dft): skip trivial coset scaling in RecursiveDft

peter941221:
This partially addresses #445.

Scope note:
- the original issue listed three possible improvements around `RecursiveDft` coset LDEs
- current `main` already covers two of them in substance:
  - `MontyField31::div_2exp_u64` already special-cases the Monty `2^{-n}` path
  - `RecursiveDft::coset_lde_batch` already fuses inverse-DFT scaling with the nontrivial coset shift
- this PR targets the remaining low-risk residual win on today's code: the trivial-shift path

Failure mechanism:
- `dft::TwoAdicSubgroupDft::lde_batch()` defaults to `coset_lde_batch(..., F::ONE)`
- `RecursiveDft::coset_lde_batch` still built `shifted_powers` and multiplied every coefficient by a trivial `1^i` weight on that path
- for plain LDEs this adds avoidable work in the hot row-scaling loop

Semantic change:
- add a scale-only helper for the zero-padded coefficient buffer
- in `RecursiveDft::coset_lde_batch`, use that helper when `shift == ONE`
- keep the existing fused shift-and-scale path unchanged for nontrivial shifts
- add a dedicated regression test showing `RecursiveDft::coset_lde_batch(..., ONE)` still matches `lde_batch(...)`

Why this shape:
- preserves the existing API and nontrivial-shift behavior
- targets the exact plain-LDE path exercised by the default trait implementation
- keeps the change local to `RecursiveDft` instead of broadening it into a generic DFT refactor

Files changed:
- `monty-31/src/dft/mod.rs`
- `dft/tests/testing.rs`

Tests:
- `cargo test -p p3-dft recursive_coset_lde_with_shift_one_equals_lde -- --nocapture`
- `cargo test -p p3-dft coset_lde_with_shift_one_equals_lde -- --nocapture`
- `cargo test -p p3-monty-31 -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

Issue #445, opened 2024-09-06
Faster coset shift in LDE


unzvfu:
Three minor improvements to [this part of the code](https://github.com/Plonky3/Plonky3/blob/2df15fd05e2181b31b39525361aef0213fc76144/monty-31/src/dft/mod.rs#L279):
1. The Monty form of 1/2^n is just `2^{32 - n}` when 1 < n <= 32, with our current implementation. (Implementing this in a way that only depends on `MONTY_BITS` and `PRIME` etc. is a bit messy though.)
2. The coset shift part of the LDE can be elided by integrating the shift into the calculation of the twiddle factors.
3. We could choose the shift to be something nice.

PR #1715, opened 2026-05-27
perf(monty-31): skip trivial coset shift in RecursiveDft

Lubov66:
Partially addresses #445

Follow-up to #1713

When shift==ONE, coset_lde_batch was computing shift^i=1 for every coefficient needlessly. Add a scale_rows helper that skips the power computation for the trivial-shift path, and harden both helpers with debug_assert invariants covering length divisibility, out_ncols>=ncols, and matching row counts, without these guards, par_chunks_exact and izip silently truncate on invariant violations which is unacceptable in ZK infrastructure

### DIFF
--- a/monty-31/src/dft/mod.rs
+++ b/monty-31/src/dft/mod.rs
@@ -30,6 +30,10 @@ fn coset_shift_and_scale_rows<F: Field>(
     shift: F,
     scale: F,
 ) {
+    debug_assert!(out.len().is_multiple_of(out_ncols));
+    debug_assert!(mat.len().is_multiple_of(ncols));
+    debug_assert!(out_ncols >= ncols);
+    debug_assert_eq!(out.len() / out_ncols, mat.len() / ncols);
     let powers = shift.shifted_powers(scale).collect_n(ncols);
     out.par_chunks_exact_mut(out_ncols)
         .zip(mat.par_chunks_exact(ncols))


tcoratger:
Hey,

Thanks a lot for this, do you have benchmarks before/after to showcase the improvement?

Lubov66:
Hey @tcoratger the main goal here is more about correctness/API hygiene than raw performance. This adds explicit invariant checks to both helpers, since without them par_chunks_exact and izip! can silently truncate on invalid inputs, which is not great in ZK code.
The perf improvement is mostly a small side benefit from skipping unnecessary shifted-power computation in the trivial shift == ONE case. I don’t have solid benchmark numbers for that yet, but I can add some if helpful.

tcoratger:
> Hey @tcoratger the main goal here is more about correctness/API hygiene than raw performance. This adds explicit invariant checks to both helpers, since without them par_chunks_exact and izip! can silently truncate on invalid inputs, which is not great in ZK code. The perf improvement is mostly a small side benefit from skipping unnecessary shifted-power computation in the trivial shift == ONE case. I don’t have solid benchmark numbers for that yet, but I can add some if helpful.

Then we should probably keep only the assertion things for this PR since the rest doesn't bring that much perf improvements.

Lubov66:
Done, kept only the assertion checks and removed the perf-specific part

monty-31/src/dft/mod.rs
```diff
@@ -30,6 +30,10 @@ fn coset_shift_and_scale_rows<F: Field>(
     shift: F,
     scale: F,
 ) {
+    debug_assert_eq!(out.len() % out_ncols, 0);
+    debug_assert_eq!(mat.len() % ncols, 0);
```
tcoratger: can't we use `is_multiple_of` here instead for the two lines?

Lubov66: Done, updated to is_multiple_of

tcoratger (approved):

PR #1716, opened 2026-05-27
perf(batch-stark): parallelize per-instance quotient computation

wstran:
## Summary

The per-instance quotient loop in `prove_batch` carried a `// TODO: Parallelize this loop for better performance with many instances.` This does that.

The loop's only cross-iteration dependency was `perm_counter` (the permutation-matrix index, bumped once per instance that contributes a permutation trace). I precompute it as a prefix sum (`perm_indices`), which makes each iteration independent, then run the per-instance work with `into_par_iter`. Per-instance results `(chunk_domains, ldes)` are collected in index order and concatenated, so `quotient_chunk_domains` / `_mats` / `_ranges` — and therefore the committed layout — are byte-for-byte what the sequential loop produced.

`quotient_values` already parallelises over rows, so this outer parallelism mostly helps when many small instances each leave cores idle; with a few large instances the inner parallelism already saturates and the gain is smaller (but never negative).

## Bench

M1 (aarch64), Fibonacci AIR batch, `prove_batch` only (no verify), criterion medians, before → after:

| batch | before | after | speedup |
|-------|-------:|------:|--------:|
| 16 × 2¹⁰ rows | 39.3 ms | 27.0 ms | 1.46× |
| 32 × 2¹⁰ | 75.0 ms | 49.8 ms | 1.50× |
| 64 × 2¹⁰ | 147 ms | 95.7 ms | 1.54× |
| 8 × 2¹⁴ | 48.5 ms | 41.3 ms | 1.18× |
| 16 × 2¹⁴ | 85.1 ms | 69.3 ms | 1.23× |
| 1 × 2¹⁴ | 18.2 ms | 18.2 ms | 1.00× (unchanged) |

## Note on the added bounds

Parallelising a function that is generic over the PCS requires `Send`/`Sync` on the config-derived types the closure touches: `Domain<SC>: Send + Sync`, and `Sync` on `SC::Pcs`, its `ProverData`, and its `Commitment`. Every concrete config in the workspace already satisfies these (they are simple `Copy`/`Arc`-backed types), so the additions are non-breaking; happy to adjust the phrasing if you'd prefer them grouped differently.

## Test plan

- [x] `cargo test -p p3-batch-stark` — 38 passed.
- [x] `cargo test -p p3-batch-stark --features parallel` — 38 passed (correctness under real Rayon).
- [x] Determinism: proving the same batch twice yields byte-identical proofs.
- [x] `cargo clippy -p p3-batch-stark --all-targets [--features parallel] -- -D warnings` — clean.
- [x] `cargo build --target thumbv7em-none-eabi -p p3-batch-stark` — clean (no_std).

### DIFF
--- a/batch-stark/src/prover.rs
+++ b/batch-stark/src/prover.rs
@@ -31,6 +31,9 @@ use crate::symbolic::{
 };
 use crate::transcript::BatchTranscript;
 
+/// Per-instance quotient output: the chunk domains and their committed LDE matrices.
+type InstanceQuotient<SC> = (Vec<Domain<SC>>, Vec<RowMajorMatrix<Val<SC>>>);
+
 /// A single AIR instance bundled with its execution trace, public inputs,
 /// and lookup declarations.
 ///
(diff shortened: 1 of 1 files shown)


tcoratger (approved):
This looks great, I've just fixed a tiny comment, maybe @Nashtare you can double check. I think this is nice for the recursion project :)

PR #1717, opened 2026-05-28
feat(sumcheck): add HVZK suffix layout overlay

peter941221:
Issue: https://github.com/Plonky3/Plonky3/issues/1649

This adds the Construction 6.3 HVZK overlay for the suffix-binding sumcheck layout, then follows up by packing the suffix residual compression path that became the dominant local hotspot.

Failure mechanism
Before this change, sumcheck::zk only supported the prefix-binding layout. Routing WHIR PCS through the suffix layout therefore missed the HVZK overlay and could produce a non-private proof.

Semantic change
Add ZkSuffixProver on top of SuffixProver.
Make the shared ZkVerifier layout-aware so suffix replay uses the suffix plain strategy.
Expose the suffix plain-layout helpers needed by the overlay: sum, compress_stacked, combine_eqs.
Add zk_suffix to whir/benches/sumcheck.rs.
Add SplitEq::new_unpacked_at and SplitEq::new_packed_at, then switch SuffixProver::compress_stacked to SplitEq::new_packed_at(rs, 0, scale) so the residual suffix compression keeps the full challenge point in the packed suffix half.

Why the follow-up compression change helps
The old residual path built SplitEq from the 4-variable suffix challenge point with the default midpoint split. That turned the point into a 2 + 2 split, which left too few suffix variables to reach the packed suffix-dot kernel. The new explicit split keeps the point as 0 + 4, so compress_stacked can use the packed kernel on the hot residual path.

Preserved invariants
Existing prefix HVZK behavior is unchanged.
Verifier replay still reconstructs the dropped c_1 from the same affine identity.
Simulator coupling still matches mu_tilde, mask commits, and the base-field stratification of wire coordinates with index >= 2.
The explicit SplitEq split does not change suffix-compression semantics; the new split_eq regression test checks that explicit and midpoint splits produce the same compressed polynomial.

Clean benchmark
Before commit: d87b91d3
After commit: f9329519

Command:
cargo bench -p p3-whir --bench sumcheck -- --noplot --sample-size 20 --warm-up-time 1 --measurement-time 3 "layout_sumcheck/(prefix|suffix|zk|zk_suffix)/log20|layout_sumcheck_phases/zk_suffix_(prove_only|residual_only|compress_stacked_only|combine_eqs_only)/log20"

Environment note:
I ran the before and after measurements in separate worktrees with separate CARGO_TARGET_DIR values and the same benchmark command on both sides.

End-to-end log20 medians:
prefix: 6.5519 ms -> 5.7528 ms (-12.2%)
suffix: 9.2118 ms -> 4.5525 ms (-50.6%)
zk: 6.4967 ms -> 6.3611 ms (-2.1%)
zk_suffix: 9.8896 ms -> 5.3652 ms (-45.7%)

Hotspot-focused log20 medians:
zk_suffix_prove_only: 9.9190 ms -> 5.9595 ms (-39.9%)
zk_suffix_residual_only: 7.2063 ms -> 4.7307 ms (-34.4%)
zk_suffix_compress_stacked_only: 8.3242 ms -> 3.9564 ms (-52.5%)
zk_suffix_combine_eqs_only: 871.29 us -> 865.19 us (effectively flat)

Interpretation:
The packed residual-compression follow-up clearly pays off on both the suffix plain path and the new suffix-ZK path.
The largest direct win is on compress_stacked, which matches the phase-split diagnosis.
combine_eqs is not the real bottleneck.
The old zk path remains effectively unchanged, which matches the fact that it does not use the suffix residual path.

Testing
cargo test -p p3-sumcheck zk:: -- --nocapture
Passed: 13 passed, 0 failed.

cargo test -p p3-multilinear-util split_eq -- --nocapture
Passed: 41 passed, 0 failed.

cargo bench -p p3-whir --bench sumcheck --no-run
Passed.

git diff --check
Passed.

### DIFF
--- a/multilinear-util/src/split_eq/mod.rs
+++ b/multilinear-util/src/split_eq/mod.rs
@@ -53,16 +53,37 @@ pub struct SplitEq<F: Field, EF: ExtensionField<F>> {
 }
 
 impl<F: Field, EF: ExtensionField<F>> SplitEq<F, EF> {
+    /// Creates a factored eq table with a scalar (unpacked) suffix-half using an explicit split.
+    ///
+    /// The point is split as `(point[..split], point[split..])`.
+    /// The scale factor is absorbed into the prefix-half table.
+    pub fn new_unpacked_at(point: &Point<EF>, split: usize, scale: EF) -> Self {
+        assert!(split <= point.num_variables());
+        let (z0, z1) = point.split_at(split);
+        let eq0 = Poly::new_from_point(z0.as_slice(), scale);
+        let eq1 = EqMaybePacked::new_unpacked(&z1);
+        Self { eq0, eq1 }
+    }
+
     /// Creates a factored eq table with a scalar (unpacked) suffix-half.
     ///
     /// The evaluation point is split at the midpoint.
     /// The scale factor is absorbed into the prefix-half table.
     pub fn new_unpacked(point: &Point<EF>, scale: EF) -> Self {
-        // Split z into (z_prefix, z_suffix) at the midpoint.
-        let (z0, z1) = point.split_at(point.num_variables() / 2);
-        // Build eq0 = scale * eq(z_prefix, .) and eq1 = eq(z_suffix, .).
+        Self::new_unpacked_at(point, point.num_variables() / 2, scale)
+    }
+
+    /// Creates a factored eq table, using SIMD packing for the suffix-half when possible,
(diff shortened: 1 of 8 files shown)


tcoratger:
Closing as a duplicate of https://github.com/Plonky3/Plonky3/pull/1665

PR #1718, opened 2026-05-28
perf(matrix): avoid prefill in transpose via uninit output contract

Lubov66:
This PR removes the unnecessary prefill pass in DenseMatrix::transpose by introducing transpose_uninit, a new unsafe primitive in p3_util::transpose that accepts a raw *mut T output pointer and guarantees full initialization without ever reading the destination buffer. The safe transpose API is preserved unchanged; it now delegates to transpose_uninit internally. On the caller side, DenseMatrix::transpose allocates via Vec::with_capacity and spare_capacity_mut instead of vec![T::default(); nelts], with set_len called only after the kernel completes. All unsafe contracts are explicitly documented at both the callee and call site.

### DIFF
--- a/matrix/src/dense.rs
+++ b/matrix/src/dense.rs
@@ -807,14 +807,28 @@ impl<T: Clone + Default + Send + Sync> DenseMatrix<T> {
 impl<T: Copy + Default + Send + Sync, V: DenseStorage<T>> DenseMatrix<T, V> {
     /// Return the transpose of this matrix.
     pub fn transpose(&self) -> RowMajorMatrix<T> {
-        let nelts = self.height() * self.width();
-        let mut values = vec![T::default(); nelts];
-        p3_util::transpose::transpose(
-            self.values.borrow(),
-            &mut values,
-            self.width(),
-            self.height(),
-        );
+        let nelts = self
+            .width()
+            .checked_mul(self.height())
+            .expect("matrix size overflow");
+
+        let mut values: Vec<T> = Vec::with_capacity(nelts);
+
+        unsafe {
+            let spare = &mut values.spare_capacity_mut()[..nelts];
+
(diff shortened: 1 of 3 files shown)


tcoratger:
This looks a duplicate of https://github.com/Plonky3/Plonky3/pull/1714, any benchmark to show the improvement?

Lubov66:
You're right that this overlaps with #1714 in terms of optimization goal.

The main difference is that this PR introduces a reusable `transpose_uninit` primitive with an explicit unsafe contract, instead of only changing allocation at the call site.

Benchmarks are probably the right next step here, I'll add numbers comparing both approaches.

tcoratger:
> You're right that this overlaps with #1714 in terms of optimization goal.
> 
> The main difference is that this PR introduces a reusable `transpose_uninit` primitive with an explicit unsafe contract, instead of only changing allocation at the call site.
> 
> Benchmarks are probably the right next step here, I'll add numbers comparing both approaches.

Yes please, benchmarks would help to identify the improvement. Because it is nice to merge only if it brings something in terms of speed otherwise this is just additional complexity :)

Lubov66:
I ran the existing transpose benchmarks.

The results are mixed depending on matrix size. Small matrices regress, medium sizes are roughly neutral, while large matrices show improved throughput. The main motivation for this PR is introducing a reusable transpose_uninit primitive and avoiding the upfront initialization cost in the transpose path.

I also reran cargo test -p p3-matrix and all tests pass

tcoratger:
Closing for now since he doesn't bring meaningful performance improvements.

We may reopen in the future if we find a specific usage.

Lubov66:
> Closing for now since he doesn't bring meaningful performance improvements.
> 
> We may reopen in the future if we find a specific usage.

Ok

PR #1723, opened 2026-06-02
fix: add PoW check in Circle STARK's commit phase

Nashtare:
This was missing (compared to regular FRI).


tcoratger (approved):
Thanks! LGTM!

PR #1724, opened 2026-06-02
refactor: remove duplication around AirBuilder `_named` methods

Nashtare:
No concrete behavioral changes, just deduplication / simplification of builder trait impls:

- instead of duplicating all methods into `_named` variants, we now make those the default ones to implement and derive the "unnamed" ones
- make `is_transition` the required method and derive `is_transition_window` from it with a hardcoded `WINDOW` value check


tcoratger (approved):
Great!

PR #1725, opened 2026-06-02
docs: clarify zero-knowledge guarantees

Radovenchyk:
Clarify the documentation of Pcs::ZK and HidingPcs

The current codebase already contains comments describing the masking construction used by HidingPcs. This PR adds matching rustdoc comments so that the public API documentation better reflects the implementation.


Nashtare (commented):
This isn't adding meaningful description.

PR #1728, opened 2026-06-03
fix: pass actual length to invalid_length in array deserializer

Radovenchyk:
Fix incorrect `invalid_length` reporting in array deserialization.

Previously, deserialization errors reported the expected array length as the actual received length, producing confusing messages such as:

`invalid length 3, expected an array of length 3`

This change reports the real number of parsed elements instead and adds a regression test covering the error message.


Nashtare:
Thank you but we don't accept minor PRs, we'll address this in a follow-up PR.

PR #1730, opened 2026-06-03
perf: use FFT-domain MDS for RPO over Goldilocks

Nashtare:
Leverage the FFT-domain approach used in Plonky2 for Poseidon1 / Miden-Crypto for RPO to speed-up the MDS layer of RPO over Goldilocks field. Yields about ~4% savings on my machine.

### DIFF
--- a/rescue/src/rpo/goldilocks.rs
+++ b/rescue/src/rpo/goldilocks.rs
@@ -2,10 +2,8 @@ use alloc::format;
 use alloc::vec::Vec;
 
 use p3_field::{InjectiveMonomial, PrimeCharacteristicRing, PrimeField64};
-use p3_goldilocks::{Goldilocks, SmallConvolveGoldilocks};
+use p3_goldilocks::Goldilocks;
 use p3_mds::MdsPermutation;
-use p3_mds::karatsuba_convolution::Convolve;
-use p3_mds::util::first_row_to_first_col;
 use p3_symmetric::{CryptographicPermutation, Permutation};
 
 use super::Rpo;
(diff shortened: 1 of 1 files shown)


rescue/src/rpo/goldilocks.rs
```diff
@@ -114,12 +109,201 @@ impl Permutation<[Goldilocks; RPO_GOLDILOCKS_WIDTH]> for RpoGoldilocks {
...
+    }
+}
+
 /// Width-parallel inverse S-box `x -> x^(1/7)` over `[Goldilocks; 12]`.
 ///
 /// Computes `x^10540996611094048183` via an addition chain reaching the same
 /// exponent as [`p3_field::exponentiation::exp_10540996611094048183`], applied
 /// across all 12 lanes step-by-step.
-#[inline]
+#[inline(always)]
```
Nashtare, line 300: Technically unrelated but in prior experiments I had noticed a slight diff.

rescue/src/rpo/goldilocks.rs
```diff
@@ -21,28 +19,23 @@ pub const RPO_GOLDILOCKS_SECURITY_LEVEL: usize = 128;
 const BYTES_PER_CONSTANT: usize = 9;
 
 /// First row of the width-12 circulant MDS from eprint 2022/1577.
+#[cfg(test)]
 const MDS_12_FIRST_ROW: [i64; 12] = [7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8];
```
tcoratger: I know that this is the same in terms of compilation but maybe we could put this directly in the test module so that we eliminate the `#[cfg(test)]` and this doesn't pollute the reading experience when we check the file, what do you think?

tcoratger (approved):
Just one super minor and non blocking comment, LGTM!

PR #1734, opened 2026-06-03
fix(whir): enforce two missing verifier length checks

tcoratger:
## Summary

The WHIR verifier guards every proof-shape axis — round count, initial/final sumcheck length, OOD count, query count, final-poly length — except two. On those two, a malformed count desyncs Fiat-Shamir into a panic or a probabilistic failure rather than a clean rejection. This PR closes both gaps so they reject up front with a dedicated error, matching the pattern already used everywhere else.

### 1. Intermediate-round sumcheck length (`pcs/verifier/mod.rs`)
Before `verify_rounds` runs for an intermediate round, check that the round's sumcheck carries exactly `folding_factor(round_index + 1)` polynomial evaluations (the count the prover generates). Previously unchecked, unlike the initial and final sumchecks. Returns `SumcheckError::RoundCountMismatch`.

### 2. Initial OOD answer count (`pcs/adapter.rs`)
Before absorbing `initial_ood_answers` (each drives a transcript draw), check the count against `commitment_ood_samples`. The per-round OOD path already enforces this (`RoundOodAnswerCountMismatch`); the initial path was the only one missing it. New variant `VerifierError::InitialOodAnswerCountMismatch`.

## Tests
Two negative tests on the existing `commit_and_open` / `verify` harness:
- `rejects_with_initial_ood_answer_count_mismatch_when_answer_is_dropped`
- `rejects_with_round_count_mismatch_when_intermediate_sumcheck_is_short`

Both pass; all pre-existing WHIR tests (end-to-end Poseidon2 + Keccak, and the other shape-mismatch tests) still pass; clippy is clean.

### DIFF
--- a/whir/src/pcs/adapter.rs
+++ b/whir/src/pcs/adapter.rs
@@ -128,6 +128,16 @@ where
         challenger.observe(commitment.clone());
 
         let mut layout_verifier = Verifier::<F, EF>::new(&protocol.table_shapes(), L::strategy());
+
+        // The commitment fixes how many initial OOD answers exist.
+        // Each one drives a transcript draw, so a wrong count desyncs Fiat-Shamir.
+        // Reject it up front, matching the per-round OOD guard.
+        if proof.whir.initial_ood_answers.len() != self.commitment_ood_samples {
+            return Err(VerifierError::InitialOodAnswerCountMismatch {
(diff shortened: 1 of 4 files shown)


Nashtare (approved):
LGTM, thank you!

PR #1735, opened 2026-06-03
fix(fri): reject zero-query instances on both sides

tcoratger:
@Nashtare I don't think that this really happen in reality but worth checking probably.

## Summary

With `num_queries == 0`, FRI verification is vacuous: the per-query loop in `verify_fri` never runs, so it returns `Ok(())` for **any** final polynomial. `conjectured_soundness_bits()` reports only the PoW bits with no indication the instance is unsound. The prover only guarded `max_log_arity` and a non-empty input set — never the query count.

This rejects the zero-query case on both sides, the fix the audit calls for.

### Prover (`prove_fri`)
`assert!(num_queries > 0, ...)` alongside the existing guards, so an unsound config fails fast at proving time rather than producing a proof the verifier must reject.

### Verifier (`verify_fri`)
Return the new `FriError::ZeroQueries` before any transcript work, so the instance is rejected up front instead of vacuously accepted.

## Tests
- `rejects_with_zero_queries` — replays an honest 2-query proof under `num_queries = 0` and asserts `FriError::ZeroQueries`.
- `prover_rejects_zero_queries` — `#[should_panic]` on a commit+open with `num_queries = 0`.

All FRI tests pass; clippy is clean.

### DIFF
--- a/fri/src/prover.rs
+++ b/fri/src/prover.rs
@@ -63,6 +63,10 @@ where
 {
     assert!(!inputs.is_empty());
     assert!(
+        params.num_queries > 0,
+        "num_queries must be at least 1 for FRI soundness"
+    );
+    assert!(
         params.max_log_arity > 0,
         "max_log_arity must be at least 1 to guarantee folding progress"
     );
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
Yup agreed

PR #1736, opened 2026-06-03
feat(lookup): single sampled challenge pair with per-bus domain separation

tcoratger:
Follow-up to #1628, toward a leaner lookup implementation.

## What changes

Classical LogUp draws a fresh `(α, β)` pair **per bus**, so the transcript is squeezed `2·B` times for `B` buses. This PR draws **one** pair for the whole batch and domain-separates each bus by an additive offset:

```text
denominator(bus, payload) = prefix[bus] − combined_payload
prefix[bus]               = α + (bus + 1) · γ      (γ = β^W, W = max message width)
```

The bus offset sits at `β^W`, one power above any payload term, so two messages collide only when bus **and** payload agree.

## Why it's surgical

The gadget already computes `base − combined`. Feeding `prefix[bus]` in as the `base` yields the domain-separated denominator, so **only the transcript's challenge sampling changes** — `logup.rs`, the constraint folders, and the symbolic layout are untouched. A new `Challenges` table in `p3-lookup` holds `α`, `beta_powers`, and the per-bus prefixes (built via the field's packed `powers().collect_n`).

## Soundness

- `(α, β)` is sampled after the main commitment, so the prover cannot adapt the trace to a known challenge.
- Distinct buses occupy distinct additive cosets, so any imbalance survives with overwhelming probability.
- The collision set is negligible over the extension field.

## Effect

- Lookup challenges sampled drop from `2·B` to `2` (e.g. 64 buses: 128 → 2), shrinking the Fiat–Shamir transcript.
- Proof size and committed data are unchanged — this is a transcript/derivation change only.
- Lays the groundwork for column packing, where sharing one `(α, β)` lets multiple buses live on one aux column.

## Test plan

- [x] `cargo test -p p3-lookup` — 23 passing
- [x] `cargo test -p p3-batch-stark` — 38 passing, incl. regenerated compat fixtures
- [x] `cargo clippy -p p3-lookup -p p3-batch-stark --all-targets` clean
- [x] `cargo fmt --all -- --check` clean


lookup/src/challenges.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use p3_field::Field;
+
+/// Bus-separated LogUp challenge table built from a single `(alpha, beta)` pair.
+///
+/// Classical LogUp draws a fresh pair per bus.
+/// This draws one pair and offsets each bus instead:
+///
+/// ```text
+///     denominator(bus, payload) = prefix[bus] + sum_k beta^k * payload_k
```
Nashtare: I think it should be `-`, not `+` here (we compute alpha - combined)

lookup/src/challenges.rs
```diff
@@ -0,0 +1,76 @@
...
+/// # Soundness
+///
+/// - The pair is sampled after the main commitment, so the trace cannot adapt to it.
+/// - An unbalanced bus stays unbalanced except on the fingerprint polynomial's roots.
+/// - That root set is negligible over a cryptographic extension field.
+pub struct Challenges<EF> {
+    /// Base randomness shared by every bus.
+    pub alpha: EF,
+    /// Powers `beta^0 .. beta^(W-1)` weighting successive payload elements.
+    pub beta_powers: Box<[EF]>,
```
Nashtare: nit: I get that we store `alpha` for consistency, although not used but why bothering storing the alloc'ed `beta_powers`? Either we just store beta alongside alpha or not at all, as these aren't used outside of sampling the `bus_prefix`

batch-stark/src/transcript.rs
```diff
@@ -80,24 +95,78 @@ impl<SC: SGC> BatchTranscript<SC> {
...
     {
-        let n = lookup_gadget.num_challenges();
-        let mut global = HashMap::new();
+        // The gadget reads two challenges per lookup: a denominator base and a combiner.
+        // The single-pair scheme below relies on exactly that width.
+        debug_assert_eq!(
+            lookup_gadget.num_challenges(),
+            2,
+            "single-pair domain separation expects a two-challenge gadget"
+        );
```
Nashtare: This is not in a hot path, so worth raising to release builds as well I think?

lookup/src/challenges.rs
tcoratger: Yep fixed

lookup/src/challenges.rs
tcoratger: Yes, fixed, now we just store `beta: EF`.

batch-stark/src/transcript.rs
tcoratger: Fixed

Nashtare (approved):
LGTM

PR #1737, opened 2026-06-03
fix(params): error instead of panic on infeasible WHIR/FRI params

tcoratger:
I think that in general this is better to error than panic and since this is in general what we do for the rest of the codebase, maybe it makes sense to do the same here.

## Summary

Two parameter-feasibility issues that currently surface as panics rather than recoverable errors. Both are loud failures (not silent soundness breaks), but a library user passing infeasible params — or a verifier handed a malicious proof — should get a typed error, not a process crash.

### 1. WHIR: `WhirConfig::new` panicked on infeasible params
`determine_ood_samples` panicked deep in a helper when no OOD-sample count reached the security level — reachable from the public constructor with a small field + high security level (e.g. a 31-bit field at 100-bit security). Other validations in `new` used `unwrap`/`assert` too.

- `determine_ood_samples` now returns `Option<usize>` (`None` when no count suffices).
- `WhirConfig::new` returns `Result<Self, WhirConfigError>`, converting every user-facing validation — folding factor, two-adicity, explicit rate/factor lengths, RS-domain growth, OOD feasibility — into a typed error.
- Only tests/benches/the example call `WhirConfig::new` (no production caller), so the ripple is just `.unwrap()` at those sites.

### 2. FRI: `log_global_max_height` unbounded when commitments are empty
The verifier cross-checks `log_global_max_height` against committed domains only when commitments exist. With an empty input, the check is skipped, so a malicious proof could inflate the fold schedule until `two_adic_generator(log_global_max_height)` panics (no such root of unity past the field two-adicity).

- New `FriError::GlobalMaxHeightTooLarge`, returned before the height is used.

## Tests
- `determine_ood_samples_reports_infeasibility` — `None` for a too-small field, `Some` otherwise.
- `new_errors_when_field_too_small_for_security` — `WhirConfig::new` surfaces `OodSamplesInfeasible`.
- `global_max_height_exceeds_two_adicity` — malicious proof with an inflated schedule and empty commitments is rejected with `GlobalMaxHeightTooLarge`.

All WHIR and FRI tests pass; clippy clean; dependents (`p3-circle`, `p3-uni-stark`, `p3-examples`) still build.

### DIFF
--- a/fri/src/verifier.rs
+++ b/fri/src/verifier.rs
@@ -47,6 +47,13 @@ where
     InitialReducedOpeningHeightMismatch { expected: usize, got: usize },
     #[error("global max height mismatch: expected {expected}, got {got}")]
     GlobalMaxHeightMismatch { expected: usize, got: usize },
+    #[error(
+        "global max height 2^{log_global_max_height} exceeds field two-adicity 2^{two_adicity}"
+    )]
+    GlobalMaxHeightTooLarge {
+        log_global_max_height: usize,
+        two_adicity: usize,
+    },
     #[error("round {round}: sibling values length mismatch: expected {expected}, got {got}")]
     SiblingValuesLengthMismatch {
         round: usize,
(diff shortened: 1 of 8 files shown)


Nashtare (approved):
Great, this was also on my todo list! Thanks!

PR #1738, opened 2026-06-04
fix(challenger): reject absorb length tags that overflow u8

tcoratger:
## Problem

The capacity length tag stamped by `DuplexChallenger::absorb_rate_padded_with_tag` is a `u8`, but both call sites in `MultiField32Challenger` cast a `usize` length with `as u8`, which silently wraps:

- `flush_f_if_non_empty`: tag is `n_in <= max_absorb_injective_limbs * RATE`. With e.g. BabyBear over BN254 (`limbs = 8`) and `RATE >= 32`, the bound exceeds 255.
- `observe(Hash)`: tag is `chunk.len() <= RATE`, wrapping when `RATE >= 256`.

When the wrap is reachable, two absorbs whose lengths differ by a multiple of 256 stamp the same tag, and the padded rate rows can be identical too: absorbing `[x]` vs `[x]` followed by 256 zeros packs to the same rate row (zero chunks pack to zero, the tail is zero-padded either way) with tags `1 == 257 mod 256` — a genuine transcript collision.

No realistic configuration (`RATE <= 16`) reaches this, but nothing enforced the bound.

## Fix

- `MultiField32Challenger::new` now returns an error when `max_absorb_injective_limbs::<F, PF>() * RATE > 255` (following the error-instead-of-panic direction of #1737). Since `new` already requires `F::order() < PF::order()`, limbs-per-slot is at least 1, so this single bound also covers the digest-block path (`RATE <= 255`).
- Both `as u8` casts become checked `u8::try_from(...).expect(...)` conversions, so a future refactor that breaks the constructor invariant panics loudly instead of wrapping silently.
- The byte-sized tag domain is documented on `absorb_rate_padded_with_tag`.

## Test

`test_new_rejects_length_tag_overflow` pins the boundary with BabyBear/Goldilocks (`limbs = 2`): `RATE = 128` (2 x 128 = 256) is rejected with the exact error message, `RATE = 127` (254) is accepted.

### DIFF
--- a/challenger/src/duplex_challenger.rs
+++ b/challenger/src/duplex_challenger.rs
@@ -110,6 +110,11 @@ where
     /// Clears `input_buffer` and `output_buffer` before absorbing. Used by
     /// [`MultiField32Challenger`](crate::MultiField32Challenger) so packed scalar and native-digest
     /// absorbs share this [`DuplexChallenger`]'s sponge state without queuing through `observe`.
+    ///
+    /// # Why the tag is a byte
+    ///
+    /// - The tag distinguishes at most 256 absorb lengths.
+    /// - Callers must keep logical lengths at most 255, or lengths differing by 256 collide.
     pub fn absorb_rate_padded_with_tag(&mut self, values: &[F], length_tag: u8) {
         const {
             assert!(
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
LGTM thanks!

PR #1740, opened 2026-06-04
whir: add HVZK code-switch relation

latifkasuli:
## Summary

Part of the HVZK-WHIR effort tracked in #1590. This PR adds the standalone Construction 9.7 code-switch relation algebra needed by the upcoming #1587 WHIR adapter wiring.

It builds on #1732's typed HVZK sumcheck handoff, but intentionally does not add WHIR proof payloads, round configuration, Fiat-Shamir changes, or PCS prover/verifier integration yet. Those stay in the next atomic PR.

## What Changed

- Added `whir::pcs::code_switch` with:
  - `ZkMaskClaim`
  - `CodeSwitchOutputRelation`
  - Construction 9.7 batching via `batched_claim`
  - row-based output-relation construction via `output_relation_from_rows`
  - private OOD answer helper over `padded_ood_t1`

- Added tests for:
  - the Construction 9.7 completeness identity `mu' == sl'(witness)`
  - `eps` handoff scaling, including auxiliary-mask covectors
  - interleaving depth `iota = 2`
  - Reed-Solomon row decomposition through `message_row` / `randomness_row`
  - round-0 `source_randomness_len = 0` behavior
  - #1732 simulator-boundary handoff into the code-switch relation
  - programmable source/mask openings and private OOD answers for the Lemma 9.8 boundary

## API Scope

The core relation types and functions are `pub` in this first slice because this module is introduced before the WHIR adapter consumer lands. The next adapter PR is expected to consume them inside `p3-whir`, at which point anything that does not need to remain externally visible can be tightened to `pub(crate)`.

The `LinearZkEncoding` convenience path and deterministic verifier-view helper are `#[cfg(test)]`; they exist only to test the Construction 9.7 identity and Lemma 9.8 simulator boundary without exposing extra production API.

## Deferred Work

This PR does not yet wire the relation into the WHIR round loop. Follow-up PRs will add:

- `RoundZkConfig` / proof payloads
- round-0 code-switch prover and verifier replay
- Fiat-Shamir domain-separator bindings for the ZK round shape
- full PCS composition work tracked separately by #1589

## Verification

```text
cargo test -p p3-whir code_switch
cargo test -p p3-whir
cargo +nightly fmt --check
cargo +stable clippy -p p3-whir --all-targets -- -D warnings
git diff --check upstream/main...HEAD
```

### DIFF
--- a/whir/src/pcs/code_switch.rs
+++ b/whir/src/pcs/code_switch.rs
@@ -1,38 +1,45 @@
 //! HVZK code-switching round (Construction 9.7, eprint 2026/391 Section 9.4).
 //!
-//! This module provides the Construction 9.7 algebra that the next #1587 WHIR
-//! adapter PR will consume. It intentionally does not yet add proof payloads or
-//! change the PCS round loop.
+//! - Reduces a proximity claim about a source oracle to one about a smaller
+//!   target oracle, in zero knowledge.
+//! - This module owns the deterministic relation algebra only.
+//! - Proof payloads and the round loop land with the WHIR adapter (#1587).
 //!
-//! Reduces a proximity claim about oracle `f` w.r.t. source code `C` to a
-//! proximity claim about oracle `g` w.r.t. a smaller target code `C'`.
+//! # Round shape
 //!
-//! The ZK variant adds:
-//! 1. A fresh mask oracle `s = Enc_{C_zk}((r, s_pad), r'')`.
-//! 2. A private-zero-evader OOD answer `y = ze_ood(rho) · (f, r, s_pad)^T`.
+//! ```text
+//! prover  : sends fresh mask oracle encoding (r || s_pad)
+//!           answers OOD points  y_i = ze*(rho_i) * (f || r || s_pad)^T
+//! verifier: opens f at x_1..x_t, samples batching coefficients nu
+//!           batches             mu' = nu_1*mu + sum nu*y_i + sum nu*f(x_j)
+//! output  : linear relation over (f, carried masks, (r || s_pad))
+//! ```
 //!
-//! The algebra and identity tests use `p3-zk-codes` (`LinearZkEncoding`)
-//! and `whir::utils` (zero-evader helpers) directly.
+//! - `r`: the source encoding randomness.
+//! - `s_pad`: fresh, and what hides the OOD answers.
 //!
-//! # Sumcheck handoff
+//! # Sumcheck handoff (Construction 6.3)
 //!
-//! The HVZK sumcheck handoff from #1732 carries an `eps`-scaled residual
-//! claim. The code-switch relation keeps that scale explicit so composing
-//! Construction 9.7 after Construction 6.3 scales only the inherited source
-//! covector, not carried auxiliary mask covectors.
+//! The incoming residual claim carries a challenge `eps` on its source part:
 //!
-//! # Simulator boundary
+//! - source covector: scaled by `eps` here,
+//! - auxiliary covectors: `eps`-opaque, any scale is already baked in.
 //!
-//! Lemma 9.8's simulator has randomized components supplied by the previous
-//! and next IORs: private-zero-evader OOD samples, source-oracle query answers,
-//! and mask/source codeword openings. This module owns the deterministic
-//! Construction 9.7 reduction from those sampled answers to the verifier's
-//! `mu'` claim and output relation. Full PCS-level simulator composition,
-//! including target and mask opening simulation, belongs to the composed WHIR
-//! prover/verifier path.
+//! # Privacy preconditions (not enforced here)
+//!
+//! - `pad_len >= t_ood`: one fresh pad coordinate per OOD answer.
+//! - OOD points pairwise distinct and nonzero.
+//! - Otherwise the answers leak a linear functional of the committed data.
+//!
+//! # Simulator boundary (Lemma 9.8)
+//!
+//! ```text
+//! adjacent simulators: sample OOD answers, query answers, openings
+//! this module        : derives the batched claim and output relation
+//! ```
 
-use alloc::vec;
 use alloc::vec::Vec;
+use core::ops::Mul;
 
 use p3_field::{Field, dot_product};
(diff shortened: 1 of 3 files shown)


tcoratger (commented):
Thanks for this, the algebra reads cleanly and I went through it line by line against Construction 9.7 in eprint 2026/391 using Claude for the first cycle of review. 

The completeness identity holds, the OOD power split and the G^# / G^$ row placement match the paper, and the mask ordering (r, s_pad) is consistent across the mask message, the OOD concat, and the in-domain randomness rows. No correctness or safety issues that I could find, and the verifier-facing functions don't panic on caller input. Nice test suite too.

A few small, non-blocking notes inline. Two more that didn't land on a changed line so I'll leave them here:

- whir/src/utils.rs lines 38 and 80: padded_ood_t1 and eval_ze_star_n still have #[allow(dead_code)] with a TODO to remove it once the code-switch module calls them. This PR is that module (private_ood_answers calls padded_ood_t1, which calls eval_ze_star_n), so both are reachable from non-test code now and the attributes can probably go.

- A couple of cheap edge cases that aren't covered yet: mask_len == 0 (source_randomness_len and pad_len both 0) and rho == 0 in an OOD point. Both look correct to me by hand (rho == 0 collapses the covector to coeff * e_0 and gives y = f[0]), so a tiny known-answer test would just pin them down.

One scope thing worth keeping in mind for the follow-ups: this module is pure relation algebra and doesn't enforce soundness or ZK on its own. batched_claim takes the prover's OOD answers as trusted scalars, so the binding only kicks in once the composed verifier checks mu' == evaluate(witness) against committed and opened values, and the OOD / batching error terms will need to flow into parameter selection in #1587. All correctly out of scope here, just noting it so it isn't lost.

whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,525 @@
...
+        }
+        if source_covector.len() != source_message_len {
+            return Err(CodeSwitchError::SourceCovectorLengthMismatch {
+                expected: source_message_len,
+                actual: source_covector.len(),
+            });
+        }
+
+        let mask_len = source_randomness_len + pad_len;
+        let source_inherited_scale = self.base_claim_coeff * self.residual_sumcheck_scale;
```
tcoratger, line 351: Small note on the eps handling here: the source covector gets scaled by base_claim_coeff * residual_sumcheck_scale, while the auxiliary covectors just below only get base_claim_coeff. That looks right to me, but it does mean the contract is a bit asymmetric: callers pass the source covector unscaled (eps is applied here), and pass each auxiliary covector already carrying whatever scale it has in inherited_claim.

When this composes after the sumcheck (6.3), the original mask covectors u_i arrive already eps-scaled, while the fresh sumcheck masks don't. So the #1587 adapter will need to pre-scale the original mask covectors itself before passing them in. Might be worth a sentence on residual_sumcheck_scale (line 113) spelling that out, just so the next PR doesn't get caught by it.

whir/src/pcs/code_switch.rs
```diff
@@ -0,0 +1,525 @@
...
+                actual: rho_ood_points.len(),
+            });
+        }
+        if source_rows.len() != self.in_domain_coeffs.len() {
+            return Err(CodeSwitchError::QueryPositionCountMismatch {
+                expected: self.in_domain_coeffs.len(),
+                actual: source_rows.len(),
+            });
+        }
+        // Both row slices are indexed by the same flattened source-query list.
```
tcoratger: Both of these length checks return QueryPositionCountMismatch, so if someone passes mismatched message vs randomness row counts the error ends up pointing at the wrong slice. Not a big deal given the comment explains they share the same query list, but a separate variant (or naming which slice in the message) could make a future mismatch a little easier to track down.

whir/src/pcs/code_switch/tests.rs
```diff
@@ -0,0 +1,1198 @@
...
+        "Construction 9.7 must preserve the eps-scaled residual handoff"
+    );
+}
+
+/// Same handoff as above, but with a prior auxiliary mask oracle.
+///
+/// The HVZK sumcheck `eps` scale belongs to the source residual only. Carried mask
+/// auxiliary covectors must be batched by `nu_1`, not by `nu_1 * eps`.
+#[test]
+fn test_eps_scaled_handoff_does_not_scale_auxiliary_covectors() {
```
tcoratger: This one covers the case where the auxiliary covector is not eps-scaled, which matches a fresh sumcheck mask. The other half of the contract, an original mask covector that does carry eps into inherited_claim, isn't exercised yet. Could be nice to add a small case with inherited_claim = eps*<f,sl> + eps*<xi,u> where the aux covector passed in is already eps*u, just to lock that down for the wiring later. The aux path is eps-opaque so it should already pass.

whir/src/pcs/code_switch/tests.rs
```diff
@@ -0,0 +1,1198 @@
...
+
+    #[test]
+    fn prop_programmable_simulator_components_for_zk_source_view(seed in 0_u64..64) {
+        prop_assume!(assert_programmable_simulator_components_for_zk_source_view(
+            seed.wrapping_add(101),
+        ));
+    }
+}
+
+fn assert_programmable_simulator_components_for_zk_source_view(seed: u64) -> bool {
```
tcoratger: One thing I noticed: this returns false only when eps == 0, and everything else inside is assert_eq! that panics. The fixed-seed wrapper at line 796 does assert!(assert_...(101)), so if seed 101 ever landed on eps == 0 it would fail as a bare assert with no message. It's fine with the current seed, so maybe just a short comment noting it relies on eps != 0, or skipping like the proptest does.

whir/src/pcs/code_switch.rs
tcoratger, line 1: Maybe maybe here this is a bit overkill to have a folder for code_switch with separate tests. We usually do this manipulation for huge modules but here I think that we can put everything in the same file, including the tests, this is more self-contained.

tcoratger (approved):

PR #1741, opened 2026-06-04
fix(whir): error instead of panic on invalid per-round folding schedule

tcoratger:
## What

`FoldingFactor::compute_number_of_rounds` had two failure modes on an explicit `PerRound` schedule:

- it `panic!`d when the factors did not fold the polynomial down to the direct-send threshold (under-folding);
- it could underflow `remaining` (`remaining -= factor`) when a round folded more variables than were left (over-folding) — a debug panic / release wraparound.

This follows the codebase's ongoing move from panics to typed errors (e.g. #1737).

## Changes

- `compute_number_of_rounds` now returns `Result<(usize, usize), FoldingFactorError>`:
  - under-folding → new `FoldingFactorError::InsufficientFolding { num_variables, remaining, threshold }`;
  - over-folding a round → `FoldingFactorError::TooLarge` (closes the latent underflow).
- `WhirConfig::new` surfaces it via `?` (the `From<FoldingFactorError>` impl already existed).
- The `empty_proof` path no longer recomputes the schedule from raw `ProtocolParameters`. `WhirProof::from_protocol_parameters` is replaced by an infallible `WhirProof::empty(num_rounds, num_queries)`, fed the counts the validated `WhirConfig` already holds (`n_rounds()`, `final_queries`). This keeps that path infallible by construction — no `expect`, and nothing to propagate across the `MultilinearPcs::open` trait boundary (which returns `Self::Proof`, not a `Result`).
- Example/benches/test helper use `.expect("valid folding schedule")` (driver code only).

## Tests

- `per_round_factors_that_under_fold_error` — `PerRound([3, 2])` on 20 vars → `InsufficientFolding { 20, 15, 6 }`.
- `per_round_factor_larger_than_remaining_errors` — `PerRound([3, 9])` on 10 vars → `TooLarge(9, 7)` (the underflow guard).
- `config_rejects_per_round_factors_that_under_fold` — confirms it surfaces as `WhirConfigError::FoldingFactor(InsufficientFolding { .. })`, asserting every field.
- Existing `test_compute_number_of_rounds` expectations updated to `Ok(..)`.

`cargo fmt`, `cargo clippy --all-targets`, and the parameter + PCS round-trip test suites are clean.

### DIFF
--- a/whir/benches/fri_vs_whir.rs
+++ b/whir/benches/fri_vs_whir.rs
@@ -248,7 +248,9 @@ type WhirCommitTy<MT> = <MT as Mmcs<F>>::Commitment;
 /// - Each round adds the count of variables it folds away, minus one.
 fn default_round_log_inv_rates(num_variables: usize, folding_factor: &FoldingFactor) -> Vec<usize> {
     // One entry per intermediate round; the trailing direct-send round has no entry.
-    let (num_rounds, _) = folding_factor.compute_number_of_rounds(num_variables);
+    let (num_rounds, _) = folding_factor
+        .compute_number_of_rounds(num_variables)
+        .expect("valid folding schedule");
     let mut rates = Vec::with_capacity(num_rounds);
     // Start at the base rate of the first committed codeword.
     let mut rate = 1;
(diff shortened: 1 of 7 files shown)


Nashtare (approved):
Thanks, that's better!

PR #1745, opened 2026-06-04
whir: hoist ExtensionMmcs out of per-round verify path

tcoratger:
## Summary

The extension-field Merkle wrapper (`ExtensionMmcs`) was reconstructed inside every `verify_merkle_proof` call, which the WHIR verifier invokes once per round. The wrapper only holds a borrow of the base Merkle scheme (a copied reference plus zero-sized `PhantomData`), so it can be built once and reused.

This hoists the construction into the `WhirVerifier` constructor, storing it as a `pub(crate)` field. `verify_merkle_proof` now reuses `self.extension_mmcs`. No public signature changes; `new` remains `const`.

This addresses finding **C1** from a WHIR audit (Plonky3 vs. worldfnd/whir). It is a code-hygiene cleanup — since the wrapper is a zero-cost borrow, the runtime impact is negligible.

## Validation

- `cargo fmt -p p3-whir` — clean
- `cargo clippy -p p3-whir --all-targets` — clean
- `cargo test -p p3-whir` — 60 passed, 0 failed (includes end-to-end Prefix/Suffix and Keccak)

### DIFF
--- a/whir/src/pcs/verifier/mod.rs
+++ b/whir/src/pcs/verifier/mod.rs
@@ -54,6 +54,11 @@ where
     pub(crate) config: &'a WhirConfig<EF, F, Challenger>,
     /// Base-field Merkle commitment scheme used to authenticate STIR queries.
     pub(crate) mmcs: &'a MT,
+    /// Extension-field view over the base Merkle scheme.
+    ///
+    /// - Authenticates query openings whose values live in the extension field.
+    /// - Built once at construction, not rebuilt on each per-round opening check.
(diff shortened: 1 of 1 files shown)


Nashtare (commented):
I think I prefer instantiating it inside the verify() call. It's not performance critical (just a ptr + phantom data) and removes the need to store it alongside the `mmcs` field in the struct

tcoratger:
> I think I prefer instantiating it inside the verify() call. It's not performance critical (just a ptr + phantom data) and removes the need to store it alongside the `mmcs` field in the struct

No problem, this is very small, let's close!

PR #1748, opened 2026-06-05
fix(lookup): enforce the LogUp multiplicity height-bound

tcoratger:
## Summary

`SymbolicInteraction::count_weight` is documented as the soundness weight for the LogUp height-bound `∑ᵢ wᵢ·hᵢ < p`, but it was **dropped** when building `Lookup` in `Lookups::from_interactions` and **never enforced anywhere** in the repo.

LogUp proves a multiset identity over the base field of characteristic `p`. A provided table entry's multiplicity equals the number of queries that hit it; counted honestly, that number never exceeds `∑ᵢ wᵢ·hᵢ` (per-row weight × trace height, summed over AIRs). If it can reach `p` it wraps modulo `p`, letting a prover forge a wrapped multiplicity and break soundness. The bound bites across AIRs — e.g. on BabyBear (p ≈ 2³⁰·⁹), three AIRs of height 2³⁰ each already sum past `p`.

## Fix

- **`lookup`**: carry the weight through to `Lookup`, add `Lookups::total_count_weight()`, and add `check_multiplicity_height_bound(lookups, heights)` — a single shared validator rejecting `∑ᵢ wᵢ·hᵢ ≥ p`. It compares against `Field::order()` (a `BigUint`), so it is correct for any prime size and adds **no new trait bounds** to the orchestration. New `LookupError::MultiplicityHeightBoundExceeded` variant.
- **`batch-stark`**: both sides call the same function — the verifier (soundness-critical, via `?` right after decoding the validated base degree bits) and the prover (fail-fast `.expect`, so it never produces a doomed proof). Because both delegate to one function, they agree by construction.
- Local lookups carry weight `0`: they balance within a single AIR and never cross buses, so they sit out the cross-AIR bound.
- `lookup/Cargo.toml` gains the no_std `num-bigint` workspace dep.

## Why the check lives at the function level (test coverage note)

Prover and verifier both delegate to `check_multiplicity_height_bound`, so that function *is* the failure path. A true end-to-end `verify_batch` rejection isn't practical to test — it needs combined trace heights ≈ `p` (~2 billion rows), and faking it by tampering the proof's `degree_bits` is caught earlier by `validate_degree_bits`. So the meaningful coverage is at the function level.

New unit tests:
- weight propagation through `from_interactions` (global query → 1, table → 0, local → 0)
- `total_count_weight` aggregation
- accept path (sum well below `p`, and zero-weight AIRs never trip regardless of height)
- general rejection (2·2³⁰ = 2³¹ > p), asserting the error fields
- **exact `≥ p` threshold**: `sum == p` rejects, `sum == p-1` accepts
- length-mismatch panic

`cargo test -p p3-lookup -p p3-batch-stark` passes (the 38 existing batch-stark integration tests confirm the new check doesn't reject valid proofs); `cargo fmt --check` and `cargo clippy --all-targets` are clean.

### DIFF
--- a/lookup/src/types.rs
+++ b/lookup/src/types.rs
@@ -8,7 +8,7 @@ use core::ops::Deref;
 use num_bigint::BigUint;
 use p3_air::symbolic::AirLayout;
 use p3_air::{Air, SymbolicExpression};
-use p3_field::{ExtensionField, Field};
+use p3_field::{ExtensionField, Field, PrimeField};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
@@ -144,7 +144,7 @@ impl<F: Field> Lookups<F> {
 /// # Errors
 ///
 /// - When the weighted sum reaches the field characteristic.
-pub fn check_multiplicity_height_bound<F: Field>(
+pub fn check_multiplicity_height_bound<F: PrimeField>(
     lookups: &[Lookups<F>],
     heights: &[usize],
 ) -> Result<(), LookupError> {
--- a/lookup/src/types.rs
+++ b/lookup/src/types.rs
@@ -5,6 +5,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::ops::Deref;
 
+use num_bigint::BigUint;
 use p3_air::symbolic::AirLayout;
 use p3_air::{Air, SymbolicExpression};
 use p3_field::{ExtensionField, Field};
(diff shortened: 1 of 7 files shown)


lookup/src/types.rs
```diff
@@ -92,13 +102,73 @@ impl<F: Field> Lookups<F> {
...
+
+    // Accumulate `sum_i w_i * h_i` in 128 bits.
+    // A saturating add only caps far past any field size, where the bound already fails.
+    let weighted_height_sum = lookups.iter().zip(heights).fold(0u128, |acc, (air, &h)| {
+        let term = u128::from(air.total_count_weight()).saturating_mul(h as u128);
+        acc.saturating_add(term)
+    });
+
+    // Compare against the exact characteristic `p`, valid for any prime size.
+    if BigUint::from(weighted_height_sum) >= F::order() {
```
Nashtare, line 165: You need to restrict `F: PrimeField` here, otherwise `F::order()` could be `p^k` for extension fields, not the characteristic.

tcoratger: Fixed

Nashtare (approved):
I'm a bit unsure though, if an AIR emits a query whose count expression exceeds 1 per row (e.g. a multiplicity-aware range check) but we leave `count_weight = 1` by mistake, the bound under-counts and the wrap it's meant to prevent comes back, no? Perhaps some clarified doc on the contract with `count_weight` and some debug_assert would help.

PR #1752, opened 2026-06-05
fix(util): guard reverse_bits_len against oversized bit_len

tcoratger:
## Summary

`reverse_bits_len(x, bit_len)` computes the shift amount as `usize::BITS - bit_len as u32`. For `bit_len > usize::BITS` that `u32` subtraction underflows; in **release** builds (overflow-checks off) `overflowing_shr` then masks the shift to `bit_len & 63`, so the function **silently returns a wrong bit-reversal permutation** — the "silently corrupts a proof" class of bug.

All current callers pass `bit_len <= 64`, so this is not exploitable today; it's defensive hardening of a load-bearing primitive.

## Fix

- Add `debug_assert!(bit_len <= usize::BITS as usize)` to document and enforce the precondition. In debug the underflow already panics (with a cryptic `"attempt to subtract with overflow"`); this surfaces misuse with a clear, named precondition and is caught in CI (tests run in debug). Release behaviour is unchanged and continues to rely on the all-callers-≤64 invariant — so this is a developer aid, not a release-time guarantee. (Happy to switch to a real `assert!` if release-time protection is wanted, at the cost of a branch in a hot path.)
- Fix the misleading comment on `reverse_bits`: the power-of-two check is a `debug_assert`, not an unconditional assert. Replaced `// Assert that n is a power of 2` with an explanation of *why* `n` must be a power of two (it's a domain size whose bit length is `log2(n)`).

## Tests

- `test_reverse_bits_len_full_width` — the boundary the guard allows (`bit_len == usize::BITS`) still produces a correct full-width reversal.
- `test_reverse_bits_len_rejects_oversized_bit_len` — `#[cfg(debug_assertions)]` + `#[should_panic(expected = "bit_len <= usize::BITS")]`. The pinned message matters: I verified that without the guard the panic is `"attempt to subtract with overflow"` and the test **fails**, so it specifically confirms the new guard fires first.

`cargo test -p p3-util` passes; `cargo fmt --check` and `cargo clippy --all-targets` are clean.

### DIFF
--- a/util/src/lib.rs
+++ b/util/src/lib.rs
@@ -194,7 +194,7 @@ pub const fn assert_sync<T: Sync>() {}
 
 #[inline]
 pub const fn reverse_bits(x: usize, n: usize) -> usize {
-    // `n` is a domain size: its bit length is `log2(n)`, integral only for a power of two.
+    // Assert that n is a power of 2
     debug_assert!(n.is_power_of_two());
     reverse_bits_len(x, n.trailing_zeros() as usize)
 }
--- a/util/src/lib.rs
+++ b/util/src/lib.rs
@@ -194,13 +194,16 @@ pub const fn assert_sync<T: Sync>() {}
 
 #[inline]
 pub const fn reverse_bits(x: usize, n: usize) -> usize {
-    // Assert that n is a power of 2
+    // `n` is a domain size: its bit length is `log2(n)`, integral only for a power of two.
     debug_assert!(n.is_power_of_two());
     reverse_bits_len(x, n.trailing_zeros() as usize)
 }
 
 #[inline]
 pub const fn reverse_bits_len(x: usize, bit_len: usize) -> usize {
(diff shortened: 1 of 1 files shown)


util/src/lib.rs
```diff
@@ -194,13 +194,16 @@ pub const fn assert_sync<T: Sync>() {}
 
 #[inline]
 pub const fn reverse_bits(x: usize, n: usize) -> usize {
-    // Assert that n is a power of 2
+    // `n` is a domain size: its bit length is `log2(n)`, integral only for a power of two.
```
Nashtare: I find the comment misleading. `reverse_bits` is part of the public API so could be used by downstream users for anything unrelated to domain size.

Nashtare (approved):
I'd change the comment, or leave it as it was before but otherwise LGTM

util/src/lib.rs
tcoratger:
```suggestion
    // Assert that n is a power of 2
```

PR #1755, opened 2026-06-05
fix(poseidon1-air): add missing (11, 1) S-box constraint degree

tcoratger:
## Summary

The `(SBOX_DEGREE, SBOX_REGISTERS) = (11, 1)` configuration is implemented on both sides of the AIR:

- `generate_sbox` writes the committed `x^3` and outputs `(x^3)^3 * x^2` (documented in its strategy table)
- `eval_sbox` constrains `committed_x3 = x^3` and produces the same degree-5 output expression

but `sbox_constraint_degree` had no arm for it and `panic!`ed. Since `BaseAir::max_constraint_degree` (both the scalar and vectorized AIRs) calls this lookup, instantiating `(11, 1)` panicked at proving time despite being a fully supported configuration.

For comparison, `poseidon2-air`'s supported sets are consistent — it implements no `(11, 1)` anywhere.

## Fix

- Add the `(11, 1) => 5` arm. Degree 5 is exact: the committed-cube constraint `x^3 = x · x · x` has degree 3, and the output `(x^3)^3 · x^2` (degree 5) is asserted against the committed post-state after the linear MDS layer.
- Add the missing `(11, 1)` row to the `eval_sbox` supported-configurations doc table (it documented `(11, 2)` but not `(11, 1)`).

## Tests

- `test_poseidon1_air_degree_sbox_11_1` — instantiates the `(11, 1)` AIR, asserts `max_constraint_degree() == Some(5)` (the exact call that panicked), and cross-checks with `get_max_constraint_degree` that the actual symbolic constraint degree is also 5.
- `test_constraint_satisfaction_sbox_11_1` — generates 16 permutation traces with the `(11, 1)` config and checks every constraint row-by-row, pinning that trace generation and constraint evaluation agree.

### DIFF
--- a/poseidon1-air/src/air.rs
+++ b/poseidon1-air/src/air.rs
@@ -189,6 +189,8 @@ pub(crate) const fn sbox_constraint_degree(degree: u64, registers: usize) -> usi
         (5, 0) => 5,
         (7, 0) => 7,
         (5, 1) | (7, 1) | (11, 2) => 3,
+        // One register for x^11: the output (x^3)^3 * x^2 reaches degree 5.
+        (11, 1) => 5,
         _ => panic!("Unexpected (DEGREE, REGISTERS)"),
     }
 }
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
Curious as to why we even support x^11 as S-Box to begin with but LGTM

tcoratger:
> Curious as to why we even support x^11 as S-Box to begin with but LGTM

Yeah maybe we need to check if this is really used but just merged so that at least this is usable. We can remove in the future if we don't want this anymore.

PR #1756, opened 2026-06-05
refactor(fri): hoist height-1 check and unify prover max-height

tcoratger:
## Summary

Two behavior-preserving cleanups in FRI (audit item L18).

### 1. Verifier — hoist the height-1 zero-check out of the per-batch loop

`open_input` accumulates reduced openings over all batches, then a `log_blowup` entry exists only for a height-1 (constant) trace matrix, whose quotient `(f(zeta) - f(x))/(zeta - x)` must be zero. That check sat **inside** the per-batch loop (running once per batch); it now runs once after the loop.

This is sound: the `log_blowup` entry is an α-weighted sum of contributions, and the random combining challenge rules out any cross-batch cancellation (Schwartz–Zippel), so checking the final value once is equivalent to checking every partial sum.

### 2. Prover — collapse two names for the same quantity

Query-index sampling used a local `log_max_height = log2(inputs[0].len())` while `open_input` used the passed-in `log_global_max_height`. They are always equal by construction (the largest input is the reduced opening at the global max height, and the verifier samples indices using `log_global_max_height` too), but the two names were a latent desync footgun.

- Use the canonical `log_global_max_height` for index sampling.
- Add `debug_assert_eq!(log_global_max_height, log2(inputs[0].len()))` to document the precondition and catch a caller that ever passes a mismatched value.
- Update the collision-probability comment accordingly.

(The unrelated `log_max_height` inside `open_input` is a genuinely different per-matrix height and is left untouched.)

No new tests — both are refactors fully exercised by the existing FRI round-trip suite. `cargo test -p p3-fri` passes; `cargo fmt --check` and `cargo clippy --all-targets` are clean.

### DIFF
--- a/fri/src/prover.rs
+++ b/fri/src/prover.rs
@@ -78,7 +78,12 @@ where
         "Inputs are not sorted in descending order of length."
     );
 
-    let log_max_height = log2_strict_usize(inputs[0].len());
+    // Index sampling and `open_input` must agree on the height; the caller's value is canonical.
+    debug_assert_eq!(
+        log_global_max_height,
+        log2_strict_usize(inputs[0].len()),
+        "log_global_max_height must match the largest input length"
+    );
     let log_min_height = log2_strict_usize(inputs.last().unwrap().len());
     if params.log_final_poly_len > 0 {
         // Final_poly_degree must be less than or equal to the degree of the smallest polynomial.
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
LGTM!

PR #1757, opened 2026-06-05
fix(merkle-tree): enforce opened row widths in batch verification

tcoratger:
## Summary

`MerkleTreeMmcs::verify_batch` checked `dimensions.len() == opened_values.len()` but never `opened_values[i].len() == dimensions[i].width` — `Dimensions::width` was entirely unused on the verify path.

This is more than an unenforced shape contract. The leaf hash (`hash_iter_slices`) **flattens all rows at one height into a single element stream** before hashing, so a digest match does not pin where one row ends and the next begins. When two or more same-height matrices are committed together, a prover can move elements across the row boundary without changing the leaf digest, and the tampered opening **verifies**:

```text
committed:  row_0 = [a, b, c]      row_1 = [d, e, f, g, h]
tampered:   row_0 = [a, b]         row_1 = [c, d, e, f, g, h]

leaf digest = hash(a, b, c, d, e, f, g, h)   — identical in both cases
```

Verified empirically: with the width check disabled, the boundary-shift opening passes verification against a real commitment. Downstream consumers (e.g. FRI's alpha-reduction, which zips opened rows with claimed evaluations) would silently process misaligned rows. For a *single* matrix per height a wrong-length row does change the absorbed stream and fails the cap check, which is why this went unnoticed.

## Fix

- **`merkle-tree/src/mmcs.rs`** — new `WrongWidth { matrix, expected, got }` error; every opened row's length is checked against its matrix width in both `verify_batch` and the amortized `verify_batch_pruned` (same flattened hashing, same gap).
- **`merkle-tree/src/hiding_mmcs.rs`** — the hiding wrapper now passes salt-widened dimensions (`width + SALT_ELEMS`) to the inner tree, matching what the inner tree actually commits.
- **`fri/src/verifier.rs`** — replaces the `width: 0` placeholder (old `TODO: MMCS doesn't really need width`) with the true width derived from the claimed evaluations.
- **`circle/src/pcs.rs`** — same for input batches; the first-layer dimensions use width 2 (queried value + sibling).

## Tests

- `verify_rejects_wrong_row_width` — an over-long opened row is rejected with the precise error.
- `verify_rejects_shifted_row_boundary` — the digest-preserving boundary shift between two same-height matrices is rejected (confirmed to verify as `Ok` before this fix).
- `claimed_width_mismatch_rejected_by_input_mmcs` (fri) — a claims/opening width disagreement now surfaces at the commitment layer.
- `point_evaluation_count_mismatch` (fri) — retargeted to a second opening point so the pairing check retains independent coverage.

Full workspace test suite passes.

### DIFF
--- a/merkle-tree/src/hiding_mmcs.rs
+++ b/merkle-tree/src/hiding_mmcs.rs
@@ -258,6 +258,43 @@ mod tests {
     }
 
     #[test]
+    fn verify_rejects_wrong_row_width() {
+        let mut rng = SmallRng::seed_from_u64(2);
+        let perm = Perm::new_from_rng_128(&mut rng);
+        let hash = MyHash::new(perm.clone());
+        let compress = MyCompress::new(perm);
+
+        // Commit to one matrix of width 4 through the hiding wrapper.
+        let mat = RowMajorMatrix::<F>::rand(&mut rng, 8, 4);
+        let dims = vec![mat.dimensions()];
+        let mmcs = MyMmcs::new(hash, compress, 0, rng);
+        let (commit, prover_data) = mmcs.commit(vec![mat]);
+
+        // Mutation: append one extra element to the opened row.
+        //
+        //     opened row:  [a, b, c, d, EXTRA]   (5 values, width is 4)
+        let mut opening = mmcs.open_batch(3, &prover_data);
+        opening.opened_values[0].push(F::ONE);
+
+        // The inner tree commits to rows widened by the salt columns.
+        // The reported widths are therefore salted as well.
+        //
+        //     expected: 4 + 4 salts = 8
+        //     got:      5 + 4 salts = 9
+        let err = mmcs
+            .verify_batch(&commit, &dims, 3, (&opening).into())
+            .expect_err("row longer than the matrix width must be rejected");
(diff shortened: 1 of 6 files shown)


merkle-tree/src/hiding_mmcs.rs
```diff
@@ -169,9 +169,19 @@ where
             .map(|(opened, salt)| opened.iter().chain(salt.iter()).copied().collect_vec())
             .collect_vec();
```
Nashtare, line 176: could be good to also have

```suggestion
let opened_salted_values = zip_eq(opened_values, salts, MerkleTreeError::WrongBatchSize)?
.map(|(opened, salt)| {
              return Err(/* error */);
          }   
          Ok(opened.iter().chain(salt.iter()).copied().collect_vec())
      })  
      .collect::<Result<Vec<_>, _>>()?;
```
for consistency?

Nashtare (approved):
One minor comment inline, otherwise LGTM!

merkle-tree/src/hiding_mmcs.rs
tcoratger: Yep, fixed :)

PR #1759, opened 2026-06-06
whir: derive round-zero HVZK config

latifkasuli:
## Summary

Part of the HVZK-WHIR sequence tracked in #1590, and the next small brick after #1740.

This PR derives the scoped round-0 ZK code-switch configuration needed by the upcoming #1587 adapter wiring. It does not add proof payloads, Fiat-Shamir transcript changes, or WHIR adapter/prover/verifier logic.

## What changed

- Added user-facing `WhirZkConfig` for the round-0 Construction 9.7 mask shape.
- Added derived `RoundZkConfig` on `RoundConfig::zk`.
- Added `WhirConfig::with_round0_zk_config(...)`, which configures only `round_parameters[0]`.
- Re-exported `WhirZkConfig` and `RoundZkConfig` from `whir::parameters`.

The derived fields intentionally mirror the paper notation:

- `mask_message_len = ell_zk`
- `mask_randomness_len = t_zk` for the RS mask encoding
- `mask_domain_size = m_zk`
- `mask_interleaving = iota_zk`
- `pad_len = ell_zk - r`

## Checks enforced

The config derivation rejects invalid local shapes before the adapter constructs the mask encoding:

- no intermediate WHIR round to attach Construction 9.7 to
- `pad_len < t_ood`, which would violate the private OOD privacy precondition
- mask domain larger than `EF::TWO_ADICITY`

The two-adicity check is against `EF`, not `F`, because the current mask encoding is extension-field valued. This catches an otherwise-late failure: `ReedSolomonZkEncoding::new(...)` checks power-of-two and length, but the actual rows/DFT later require an `EF` two-adic subgroup of size `m_zk`.

## Scope

This is round-0 scoped. Round 0 starts from the freshly committed folded WHIR oracle, which is plain today, so `source_randomness_len = 0`. Later ZK-aware source-oracle handoff and full multi-round `HidingWhirPcs` composition remain #1589.

This PR also intentionally inherits `t_ood` and downstream query counts from WHIR's existing security derivation. Full Lemma 9.9 soundness-error accounting is deferred to the adapter/soundness work, where the code-switch relation is actually composed.

## Verification

```text
cargo test -p p3-whir parameters::whir --lib
cargo test -p p3-whir
cargo +nightly fmt --check
cargo +stable clippy -p p3-whir --all-targets -- -D warnings
git diff --check
```

### DIFF
--- a/whir/src/parameters/mod.rs
+++ b/whir/src/parameters/mod.rs
@@ -9,7 +9,7 @@ pub mod whir;
 
 pub use folding::{FoldingFactor, FoldingFactorError};
 pub use soundness::SecurityAssumption;
-pub use whir::{RoundConfig, WhirConfig, WhirConfigError};
+pub use whir::{RoundConfig, RoundZkConfig, WhirConfig, WhirConfigError, WhirZkConfig};
 
 /// Fallback proof-of-work difficulty when the user does not specify one.
 ///
--- a/whir/src/parameters/whir.rs
+++ b/whir/src/parameters/whir.rs
@@ -54,6 +54,124 @@ pub enum WhirConfigError {
         security_level: usize,
         field_size_bits: usize,
     },
+
+    /// The round-0 code-switch mask does not have enough fresh pad entries to
+    /// hide all private OOD answers.
+    #[error(
+        "ZK code-switch pad length {pad_len} is smaller than private OOD answer count {private_ood_answers}"
+    )]
+    ZkPadTooSmall {
+        pad_len: usize,
+        private_ood_answers: usize,
+    },
+
+    /// ZK code-switching was requested for a configuration with no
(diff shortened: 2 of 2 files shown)


tcoratger:
@latifkasuli Superseded by https://github.com/Plonky3/Plonky3/pull/1767 when I did an attempt at the full integration and I put you as a co-author there for the credit :)

latifkasuli:
> @latifkasuli Superseded by #1767 when I did an attempt at the full integration and I put you as a co-author there for the credit :)

Sounds good, thanks for the co-author credit.

I’ll switch over to reviewing #1767 and check the full integration path carefully, especially around the Construction 9.7 code-switch/config pieces and the handoff from the merged sumcheck/code-switch bricks.

PR #1761, opened 2026-06-06
fix(uni-stark, batch-stark): reject malformed periodic column lengths

tcoratger:
## Problem

Both verifiers evaluate each AIR's periodic columns with `evaluate_periodic_column_at`, which does:

```rust
let log_period = log2_strict_usize(col.len());
let folds = self.log_size() - log_period;
```

A malformed periodic column makes this **panic** instead of returning an error:
- **Empty / non-power-of-two** column → `log2_strict_usize` asserts.
- **Oversized** column (period > trace length) → the `log_size() - log_period` subtraction underflows.

Periodic columns come from the verifier's own `Air`, so this is a robustness issue (a misconfigured AIR), not a soundness break — but a non-interactive verifier should return an error, never panic. The batch-stark verifier evaluates periodic columns the same way and suffered the same panic.

## Fix

A single shared helper validates every periodic column length before evaluation — each period must be a power of two no larger than the trace length:

```rust
pub fn check_periodic_column_lengths<F>(
    periodic_columns: &[Vec<F>],
    trace_length: usize,
) -> Result<(), PeriodicColumnError> { ... }
```

- Both the uni-stark and batch-stark verifiers call it, so the two cannot drift.
- Failures are reported with distinct typed variants: `LengthNotPowerOfTwo` and `LengthTooLarge`.

## Error categorization

Periodic columns are AIR logic, not proof data, so these failures live in a dedicated `PeriodicColumnError` surfaced as `VerificationError::PeriodicColumn` — not in the proof-shape error set. (Addresses review feedback.)

## Tests

- `uni-stark/tests/periodic_column_shape.rs`: a single-column periodic AIR parameterized by period; two instances with different periods share the same symbolic shape. Proves a valid period-2 instance, then verifies against empty / non-power-of-two / oversized periods, asserting the right variant each time.
- `batch-stark/tests/simple.rs`: a parity test proving a well-formed periodic AIR, then verifying against a same-shaped AIR with a length-3 column → rejected with `LengthNotPowerOfTwo`.


uni-stark/src/verifier.rs
```diff
@@ -389,8 +389,30 @@ where
...
+    // Each periodic column repeats with a period dividing the trace length.
+    // A valid period is a power of two no larger than the trace length.
+    let trace_length = init_trace_domain.size();
+    let periodic_columns = air.periodic_columns();
+    for col in &periodic_columns {
+        let period = col.len();
+        // Evaluation builds a subdomain of size `period`, which exists only for powers of two.
+        if !period.is_power_of_two() {
+            return Err(
+                InvalidProofShapeError::PeriodicColumnLengthNotPowerOfTwo { got: period }.into(),
```
Nashtare: very nit: but as periodic columns are AIR logic itself, it feels like it does not belong to `InvalidProof` set of errors? Not a big deal though

Nashtare (commented):
Would be nice to also include this check in the batch-stark side which I think suffers the same issue

uni-stark/src/verifier.rs
tcoratger: Fixed

tcoratger:
> Would be nice to also include this check in the batch-stark side which I think suffers the same issue

Yes just added

Nashtare (approved):
LGTM thanks!

PR #1763, opened 2026-06-06
feat(sumcheck): add next-row (shift) statements for AIR over WHIR

tcoratger:
## Overview

AIR transition constraints relate row `i` and row `i+1`. Over a multilinear commitment a trace column is `f̂(X_1, ..., X_m)` on the hypercube, and "row + 1" is an integer increment with carries — not a point evaluation. The zerocheck sumcheck therefore ends with a claim the PCS cannot natively open: `f̂_next(z)`, the MLE of the row-shifted column.

WHIR proves *constrained* Reed–Solomon codes `Σ_b ŵ(f̂(b), b) = σ` for an arbitrary weight, so the shifted opening becomes a native statement with weight `ŵ = Z · next(z, ·)`, where `next` is the MLE of the successor indicator (last row self-loops):

```text
next(x, y) = sum_j [ prod_{k<j} eq(x_k, y_k) ] * (1 - x_j) * y_j * [ prod_{k>j} x_k (1 - y_k) ]
           + prod_k x_k * y_k
```

One term per carry pivot `j` (the lowest zero bit of `x`), plus the self-loop. The weight is linear in `Z` and multilinear in `X`, so it slots into the existing soundness analysis exactly like an eq-weight.

## What this PR adds

**Kernel (`p3-multilinear-util`)**
- `Point::eval_next` / `Point::next_poly`: closed-form evaluation in `O(m)` ops — the only cost the verifier ever pays.
- `Poly::new_next_from_point`: the full `next(z, b)` table in `O(2^m)` via a block-doubling butterfly (eq doubling + one carry-pivot injection per step), same cost as the eq butterfly.

**Statement (`p3-sumcheck`)**
- `NextStatement`, a sibling of `EqStatement` / `SelectStatement`: batched claims `Σ_b next(z_i, b) · P(b) = s_i` with scalar and SIMD-packed combination.

**Constraint batching**
- `Constraint` now carries all three weight types under one challenge, with next constraints consuming the powers after eq and select:

```text
W(X) = Σ_i γ^i eq(z_i, X) + Σ_j γ^{n_eq+j} select(z_j, X) + Σ_l γ^{n_eq+n_sel+l} next(z_l, X)
```

- `VariableOrder::eval_constraints_poly` gained the `next` term. Since `WhirVerifier::verify` consumes a `Constraint` directly, the WHIR verifier supports next claims with no further changes.

## Tests

- Exhaustive boolean truth tables and multilinearity/interpolation proptests for the kernel.
- The semantics-pinning test: with `g[r] = f[r+1]` (`g[last] = f[last]`), `Σ_b next(z, b) · f(b) == ĝ(z)` — a next claim on the committed column *is* an opening of the shifted column.
- Scalar vs packed combine roundtrips, challenge-power offsets, accumulation.
- The multi-table layout sumcheck roundtrip (full protocol skeleton, both Prefix and Suffix binding) now injects next constraints at every intermediate round and checks the final identity `sum = f(r) · W(r)`.

## Follow-up (out of scope)

The stacked-layout / SVO fast path: lifting a next claim into a stacked commitment factors as `eq(selector, ·) ⊗ next(z, ·)`, and the SVO accumulator rounds need the kernel decomposed into `l_0 + 2` eq-shaped groups (one per carry pivot inside the SVO window, plus the carry-into-prefix and self-loop boundary groups). This mirrors leanVM's `compress_next_claim` and belongs in the `layout` layer as its own PR.

### DIFF
--- a/multilinear-util/src/point.rs
+++ b/multilinear-util/src/point.rs
@@ -135,6 +135,80 @@ where
         Self::eval_eq(self.as_slice(), point.as_slice())
     }
 
+    /// Evaluates the successor polynomial `next(p, q)` for slices.
+    ///
+    /// `next` is the multilinear extension of the row-successor indicator.
+    /// Rows are indexed big-endian: coordinate `0` is the most significant bit.
+    ///
+    /// ```text
+    /// next(x, y) = 1   if y = x + 1, or x = y = (1, ..., 1)
+    /// next(x, y) = 0   otherwise                  (x, y in {0,1}^n)
+    /// ```
+    ///
+    /// - The last row maps to itself instead of wrapping to row `0`.
+    /// - AIR transition constraints must therefore mask the last row with a selector.
+    ///
+    /// # Closed form
+    ///
+    /// Sum over the pivot `j`, the lowest zero bit of `x` where the carry stops:
+    ///
+    /// ```text
+    /// next(x, y) = sum_j [prod_{k<j} eq(x_k, y_k)] * (1 - x_j) * y_j * [prod_{k>j} x_k * (1 - y_k)]
+    ///            + prod_k x_k * y_k
+    /// ```
+    ///
+    /// # Performance
+    ///
+    /// - `O(n)` field operations.
+    /// - One `n`-element allocation for the suffix products.
+    ///
+    /// # Panics
+    /// Panics if `p` and `q` have different lengths.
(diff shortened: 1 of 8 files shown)


tcoratger:
Waiting for @kilic feedback before any review

PR #1764, opened 2026-06-06
feat(air): bind AIR-owned instance data into the Fiat-Shamir transcript

tcoratger:
## Problem

The prover and verifier absorb commitments, widths, degree bits, and public values into the transcript — but never the AIR's *own* instance data. The verifier comment even flagged it:

> TODO: Might be best practice to include other instance data here in the transcript, like some encoding of the AIR. This protects against transcript collisions between distinct instances.

Periodic columns are the clearest case: they are evaluated from `air.periodic_columns()` and fed into constraint folding, but never observed. More generally, any instance data an author carries in their `Air` struct is outside Fiat–Shamir. Two instances differing only in such data can share a transcript, which lets a proof be replayed against a different instance.

## Fix

Add a defaulted hook on `BaseAir`:

```rust
fn instance_encoding(&self) -> Vec<F> where F: PrimeCharacteristicRing {
    // default: periodic columns, length-prefixed
}
```

- **Default** encodes the periodic columns in a self-delimiting layout (column count, then each column's length followed by its values), so the framework-known instance data is always bound.
- **Override** lets an AIR bind values held in its own fields.

Prover and verifier absorb this encoding before sampling any challenge, in **both uni-STARK and batch-STARK** (per instance). The hook is defaulted, so existing AIRs need no changes.

## Breaking change

This adds bytes to the transcript, so previously generated proofs no longer verify. The pinned compatibility fixtures (`uni-stark` and `batch-stark`) are regenerated in this PR.

## Tests

- `air`: unit tests locking the default encoding format (length-prefixed layout; single zero when there are no periodic columns).
- `uni-stark` (`tests/air_encoding.rs`): end-to-end proof that the binding works. A one-column AIR carries a tag in its encoding but not in its constraints; a proof made with tag 1 verifies under tag 1 and is rejected (`InvalidOpeningArgument`) under tag 2. This isolates the transcript effect from any constraint-evaluation effect.
- Existing periodic-column prove/verify tests now exercise the default binding.


Nashtare (commented):
I'd push back on this. If we want to integrate the AIR definition into the FS transcript, then I'd rather have it properly done (one of the reasons we didn't have this in plonky2 because this is a bit hard without blowing up gates count).

But essentially just encoding the periodic columns metadata is only covering a small subset of possible AIRs, and only partially. In addition, these are verifier known data that it recomputes to use during the OOD check so it makes it feel redundant, unlike other aspects of the AIR.

Also rather than having an override, once we settle on how to tackle this, I think we'd rather have a default implementation that _does_ include all these metadata, and allow implementers to _complement_ it with additional instance encoding proper to their system, to remove a potential footgun (i.e. forgetting to re-implement what the default does).

tcoratger:
> I'd push back on this. If we want to integrate the AIR definition into the FS transcript, then I'd rather have it properly done (one of the reasons we didn't have this in plonky2 because this is a bit hard without blowing up gates count).
> 
> But essentially just encoding the periodic columns metadata is only covering a small subset of possible AIRs, and only partially. In addition, these are verifier known data that it recomputes to use during the OOD check so it makes it feel redundant, unlike other aspects of the AIR.
> 
> Also rather than having an override, once we settle on how to tackle this, I think we'd rather have a default implementation that _does_ include all these metadata, and allow implementers to _complement_ it with additional instance encoding proper to their system, to remove a potential footgun (i.e. forgetting to re-implement what the default does).

Ok, this makes sense, I've an idea. Let's wait for this one to land https://github.com/Plonky3/Plonky3/pull/1603 first and then we can do something more complete and robust directly integrated into the framework. Closing for now.

PR #1767, opened 2026-06-06
whir: add HVZK-WHIR hiding polynomial commitment scheme

### LINKED ISSUE #1587: [HVZK-WHIR 4/6] HVZK code-switching round

### LINKED ISSUE #1588: [HVZK-WHIR 5/6] HVZK base-case IOPP

### LINKED ISSUE #1589: [HVZK-WHIR 6/6] Compose HVZK-WHIR PCS + end-to-end tests

### LINKED ISSUE #1590: Tracking: HVZK-WHIR — zero-knowledge variant of WHIR (eprint 2026/391)

tcoratger:
## Summary

Implements the full HVZK-WHIR pipeline of [eprint 2026/391](https://eprint.iacr.org/2026/391) as a hiding multilinear polynomial commitment scheme, composing the previously merged building blocks (#1584 ZK codes, #1585 private zero-evader, #1586 HVZK sumcheck, #1740 code-switch relation).

Closes #1590 and its remaining open sub-issues:
- Closes #1587 (HVZK code-switching round, Construction 9.7)
- Closes #1588 (HVZK base-case IOPP, Construction 7.2)
- Closes #1589 (compose HVZK-WHIR PCS + end-to-end tests, Theorem 4)

## What's added

A parallel hiding pipeline in `whir/src/pcs/zk/` — the plain WHIR path is untouched byte-for-byte:

```text
commit   : interleaved ZK Reed-Solomon encoding of the witness
fold     : masked sumcheck batches        (Construction 6.3)
reduce   : HVZK code-switching rounds     (Construction 9.7)
finish   : non-succinct masked base case  (Construction 7.2)
```

| Module | Content |
|---|---|
| `adapter` | `HidingWhirPcs` implementing `MultilinearPcs`, mirroring `HidingFriPcs` |
| `committer` | interleaved ZK RS commits (per-limb randomness rows), `FoldedRsCode` |
| `constraint/` | symbolic source-claim tracking (Eq/Pow covectors closed under prefix folding) + dense mask claims |
| `code_switch` | Construction 9.7 batching algebra |
| `base_case/` | Construction 7.2 prover/verifier with one-time-pad reveals |
| `prover/`, `verifier` | the round loop on both sides |
| `config`, `proof` | derived ZK budgets and proof payload |

Supporting extensions: aux-claim carry + weight scaling in `p3-sumcheck`'s HVZK overlay, grouped interleaved mask oracles (`stack_codewords`), explicit-randomness draws in `p3-zk-codes`, Fiat-Shamir domain-separator coverage, and a `zk_overhead` criterion bench.

## Design decisions

- **Parallel pipeline, not a `zk` flag**: the carried relation differs (committed-sumcheck relation `<f, W> + sum_i <xi_i, u_i> = target`, Definition 5.8), and this keeps the non-ZK path as a regression guard.
- **Mask grouping**: masks committed together stack into one interleaved oracle (one commitment per sumcheck batch / per code-switch round), so base-case spot checks authenticate a whole group with a single Merkle path per position. This halved the proof-size overhead.
- **No commitment-phase OOD samples**: per the paper's round-by-round analysis, replaced by list-size union bounds.
- **Batching coefficients start at the first challenge power** (`gamma^{i+1}`), so the carried claim keeps coefficient one — paper-faithful.
- **HVZK is tested end-to-end**: a witness-free simulator (honest pipeline on a claim-conditioned random witness) produces an accepting transcript, plus a witness-independence coupling test and tamper-rejection tests asserting exact error variants.

## Measured overhead (KoalaBear, 100-bit security, k = 4, bench params)

| | 2^16 | 2^18 |
|---|---|---|
| Prover time | 1.29x | 1.19x |
| Proof size | 3.28x | 3.04x |

The size overhead is a fixed ~120 KB of mask spot-check openings that amortizes with witness size (the paper's `1 + o(1)` is stated for IOP oracle proof length; after Merkle compilation the additive constant is expected).

## Known follow-ups

- Packed round-0 sumcheck: the initial batch lifts the base-field witness to the extension (unpacked); driving it through the layout-based ZK prover would remove most of the remaining prover-time overhead.
- Single fresh-blind oracle at the base case (common mask domain) would cut the proof-size constant further.


whir/src/pcs/zk/config.rs
```diff
@@ -0,0 +1,450 @@
...
+    ///
+    /// - The plain WHIR configuration is invalid.
+    /// - The sumcheck mask is shorter than 3 coefficients.
+    /// - An oracle's randomness rows do not fit in its codeword slack.
+    pub fn new(
+        num_variables: usize,
+        params: ProtocolParameters,
+        zk: ZkParameters,
+    ) -> Result<Self, ZkConfigError> {
+        if zk.ell_zk < 3 {
```
soispoke, line 182: `ZkParameters` defines security-relevant ZK budgets, but `ZkWhirConfig::new` only
rejects degenerate values. Please either derive these knobs from `security_level`
or validate that caller-provided values meet the ZK security ledger. In
particular, `mask_queries`, `mask_log_inv_rate`, and `ell_zk` need to cover the
mask spot-check / list-size / zero-evader terms, not only the plain WHIR
main-code terms.

whir/src/pcs/zk/adapter.rs
```diff
@@ -0,0 +1,170 @@
...
+    F: TwoAdicField,
+    EF: ExtensionField<F> + TwoAdicField,
+    Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F>,
+{
+    /// Builds the Fiat-Shamir domain separator for this protocol instance.
+    ///
+    /// Encodes the protocol parameters and the full HVZK transcript shape.
+    /// Every challenge is thereby bound to this configuration.
+    pub fn add_domain_separator<const DIGEST_ELEMS: usize>(&self, ds: &mut DomainSeparator<EF, F>) {
+        ds.commit_statement::<Challenger, DIGEST_ELEMS>(&self.config.inner);
```
soispoke: The ZK path calls the plain commitment domain-separator builder, which still
encodes `add_ood(commitment_ood_samples)`, but the HVZK transcript has no initial
commitment-OOD phase. This isn't a soundness bug by itself, but it's misleading
and easy to misuse. Give the ZK pipeline a dedicated commit separator
shape that omits this phantom OOD step.

whir/src/pcs/zk/base_case/tests.rs
```diff
@@ -0,0 +1,402 @@
...
+        err,
+        BaseCaseZkError::MaskSpotCheckFailed {
+            group: 0,
+            position: 2
+        }
+    );
+}
+
+#[test]
+fn base_case_reveals_are_one_time_padded() {
```
soispoke, line 391: This test is labeled as checking one-time-pad behavior, but it reruns the same
secrets and randomness, so it only checks deterministic replay. Please vary the
hidden message under matched fresh-mask randomness and assert the reveal delta is
`gamma * secret_delta`; otherwise this test would pass even if it did not
meaningfully exercise the OTP property.

soispoke (commented):
- ZkWhirConfig` doesn't enforce the ZK security parameters against `security_level` (detail inline on `config.rs`)
- Two smal/non-blockingl issues, a domain-separator cleanup and a hiding test overclaim

latifkasuli:
The main protocol shape looks sound to me: the mask commitments are opened and Merkle-verified, the code-switch OOD answers feed the carried relation instead of being treated as plain evaluations, the base-case `mu_g`/commitments are bound before `gamma`, and the final folded object is handled through the masked base case rather than exposed in clear. I also don't see the old "ZK branch not exercised" issue reappearing in the core path.

A few extra things I noticed, in addition to the inline threads:

1. `MaskCodeShape::new` derives the mask domain in linear `usize` space before the two-adicity guard runs:

```rust
((message_len + randomness_len).next_power_of_two()) << log_inv_rate
```

For very large `mask_log_inv_rate`, this can overflow/panic before `MaskDomainExceedsTwoAdicity` is reachable. This is the same reason the earlier config branch used log-space derivation: compute

```text
log_domain_size = log2(next_power_of_two(message_len + randomness_len)) + log_inv_rate
```

check `log_domain_size <= EF::TWO_ADICITY`, then shift. Small mechanical fix, but it keeps the typed error honest.

2. The test named `zk_whir_end_to_end_multi_round` looks like it only gives one intermediate round with the current fixture:

```text
num_variables = 14
folding = ConstantFromSecondRound(5, 3)
14 - 5 = 9
(9 - 6).div_ceil(3) = 1
```

I'd add an explicit `assert!(config.n_rounds() >= 2)` in that test and bump the fixture if needed, so the round-to-round oracle carry path is actually exercised.

3. Building on @soispoke's note about the base-case one-time-pad test: the PCS-level simulator test (`zk_whir_simulator_produces_accepting_transcript`) has the analogous gap one level up — it runs the honest pipeline on a claim-conditioned witness and checks acceptance, which is a completeness property rather than the composed HVZK/distribution argument (Theorem 4.5). Since this PR also closes #1590, I'd either add a composed-simulator test that compares against a real transcript (byte-equal on the RS/zero-error parts under matched RNG, like the sumcheck-layer `run_view_match_rs` does) or narrow the test name/comments so the Theorem 4.5 claim isn't carried by a completeness test.

4. Minor coverage gap: I don't see an asserting proof-overhead test for the `m_zk * iota_zk + t_ood` accounting, only the benchmark printout. A small assertion would make the #1587 AC easier to audit.

whir/src/pcs/zk/config.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// # Errors
+    ///
+    /// - The plain WHIR configuration is invalid.
+    /// - The sumcheck mask is shorter than 3 coefficients.
+    /// - An oracle's randomness rows do not fit in its codeword slack.
+    pub fn new(
+        num_variables: usize,
+        params: ProtocolParameters,
+        zk: ZkParameters,
+    ) -> Result<Self, ZkConfigError> {
```
Nashtare, line 181: In addition to what's been said, we don't tie it to the targeted security level, we should have a validation layer here to ensure params are satisfactory

whir/src/pcs/zk/config.rs
tcoratger: Fixed, `ZkWhirConfig::new` now derives it

whir/src/pcs/zk/adapter.rs
tcoratger: Solved with a dedicated ZK commit separator

whir/src/pcs/zk/base_case/tests.rs
tcoratger: Solved with a real one-time-pad delta test

tcoratger:
> The main protocol shape looks sound to me: the mask commitments are opened and Merkle-verified, the code-switch OOD answers feed the carried relation instead of being treated as plain evaluations, the base-case `mu_g`/commitments are bound before `gamma`, and the final folded object is handled through the masked base case rather than exposed in clear. I also don't see the old "ZK branch not exercised" issue reappearing in the core path.
> 
> A few extra things I noticed, in addition to the inline threads:
> 
> 1. `MaskCodeShape::new` derives the mask domain in linear `usize` space before the two-adicity guard runs:
> 
> ```rust
> ((message_len + randomness_len).next_power_of_two()) << log_inv_rate
> ```
> 
> For very large `mask_log_inv_rate`, this can overflow/panic before `MaskDomainExceedsTwoAdicity` is reachable. This is the same reason the earlier config branch used log-space derivation: compute
> 
> ```
> log_domain_size = log2(next_power_of_two(message_len + randomness_len)) + log_inv_rate
> ```
> 
> check `log_domain_size <= EF::TWO_ADICITY`, then shift. Small mechanical fix, but it keeps the typed error honest.
> 
> 2. The test named `zk_whir_end_to_end_multi_round` looks like it only gives one intermediate round with the current fixture:
> 
> ```
> num_variables = 14
> folding = ConstantFromSecondRound(5, 3)
> 14 - 5 = 9
> (9 - 6).div_ceil(3) = 1
> ```
> 
> I'd add an explicit `assert!(config.n_rounds() >= 2)` in that test and bump the fixture if needed, so the round-to-round oracle carry path is actually exercised.
> 
> 3. Building on @soispoke's note about the base-case one-time-pad test: the PCS-level simulator test (`zk_whir_simulator_produces_accepting_transcript`) has the analogous gap one level up — it runs the honest pipeline on a claim-conditioned witness and checks acceptance, which is a completeness property rather than the composed HVZK/distribution argument (Theorem 4.5). Since this PR also closes [Tracking: HVZK-WHIR — zero-knowledge variant of WHIR (eprint 2026/391) #1590](https://github.com/Plonky3/Plonky3/issues/1590), I'd either add a composed-simulator test that compares against a real transcript (byte-equal on the RS/zero-error parts under matched RNG, like the sumcheck-layer `run_view_match_rs` does) or narrow the test name/comments so the Theorem 4.5 claim isn't carried by a completeness test.
> 4. Minor coverage gap: I don't see an asserting proof-overhead test for the `m_zk * iota_zk + t_ood` accounting, only the benchmark printout. A small assertion would make the [[HVZK-WHIR 4/6] HVZK code-switching round #1587](https://github.com/Plonky3/Plonky3/issues/1587) AC easier to audit.

Thanks a lot @latifkasuli and @soispoke. I fixed all the comments, don't hesitate to let me know if you find something else!

Nashtare (commented):
The PR is massive so I'd probably need to go through some of its bits a bit more, but at least I didn't get Codex nor Claude to find any major findings.

whir/src/pcs/zk/config.rs
```diff
@@ -0,0 +1,435 @@
...
+        }
+
+        // Derive the mask spot-check count t_zk.
+        //
+        //     reach security on the (1 - delta_zk)^{t_zk} branch
+        //     no PoW on mask spot checks -> target the full security level
+        //     t_zk is also the mask randomness length -> t_zk-query private
+        let security_level = params.security_level;
+        let soundness_type = params.soundness_type;
+        let mask_queries = soundness_type.queries(security_level, zk.mask_log_inv_rate);
```
Nashtare: Shouldn't it be 
```suggestion
        let union = log2_ceil_usize(2 * n_rounds + 2);
  let mask_queries = soundness_type.queries(security_level + union, zk.mask_log_inv_rate);
```

whir/src/pcs/zk/constraint/source.rs
```diff
@@ -0,0 +1,285 @@
...
+                    debug_assert!(k <= *num_variables);
+                    // Folding the first k bits scales a power covector by
+                    //
+                    //     prod_{i < k} (1 - g_i + g_i * var^{2^{m-1-i}})
+                    //
+                    // The selection walk squares the seed once per coordinate, in reverse order.
+                    //
+                    //     seed = var^{2^{m-k}}
+                    //     -> factors hit exponents 2^{m-k}, ..., 2^{m-1}
+                    let seed = var.exp_power_of_2(*num_variables - k);
```
Nashtare: can't we carry the running seed instead of recomputing? It seems a bit wasteful

whir/src/pcs/zk/prover/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+
+            let mask_claim = ZkMaskClaim {
+                base_claim_coeff: EF::ONE,
+                ood_coeffs: ood_coeffs.to_vec(),
+                in_domain_coeffs: query_coeffs.to_vec(),
+            };
+            // Carried scalar: the source residual plus the mask-side total
+            // `sum_i <xi_i, u_i>` at the current covector scales.
+            let carried =
+                batch.residual_prover.claimed_sum() + masks.claims.evaluate(&masks.messages);
```
Nashtare: this calls `masks.claims.evaluate(&masks.messages)` twice (here and below) (actually thrice but I'm ignoring the debug assert). May be worth maintaining a running value?

whir/src/pcs/zk/prover/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+        prover_data: HidingWhirProverData<F, EF, MT>,
+        claims: &[(Point<EF>, EF)],
+        challenger: &mut Challenger,
+        rng: &mut R,
+    ) -> ZkWhirProof<F, EF, MT> {
+        let config = self.config;
+        let num_variables = config.num_variables;
+        let sumcheck_mask_encoding = config.sumcheck_mask.encoding::<EF>();
+
+        // Claimed evaluations, echoed verbatim into the proof payload.
```
Nashtare:
```suggestion
        // Claimed evaluations
```

whir/src/pcs/zk/prover/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+            &mut zk_data,
+            &sumcheck_mask_encoding,
+            &self.extension_mmcs,
+            config.folding_factor(0),
+            config.starting_folding_pow_bits,
+            EF::ZERO,
+            challenger,
+            rng,
+        );
+        // Proof accumulators, assembled into the final payload at the end.
```
Nashtare:
```suggestion
        // Proof accumulators
```

whir/src/pcs/zk/verifier.rs
```diff
@@ -0,0 +1,507 @@
...
+        dims: &[Dimensions],
+        index: usize,
+        opening: &QueryOpening<F, EF, MT::Proof>,
+        round: usize,
+        randomness: &Point<EF>,
+    ) -> Result<EF, ZkVerifierError> {
+        // Defense in depth: bind the opened row length locally.
+        //
+        //     Merkle hashing already pins the committed width,
+        //     but the later fold indexes by 2^folding directly.
```
Nashtare: bit verbose
```suggestion
```

whir/src/pcs/zk/config.rs
tcoratger: Fixed

whir/src/pcs/zk/constraint/source.rs
tcoratger: Yes fixed

whir/src/pcs/zk/prover/mod.rs
tcoratger: Fixed

whir/src/pcs/zk/config.rs
tcoratger: Fixed

Nashtare (approved):
Thank you Thomas! This looks really nice! :)

PR #1769, opened 2026-06-07
fix(challenger): make duplex sponge absorbs length-binding

tcoratger:
## Summary

`DuplexChallenger::duplexing()` overwrote only the filled prefix of the rate and then permuted, without zeroing the unused rate suffix or encoding the absorbed length. As a result an absorb and its zero-extension map to the **same** pre-permutation state, so they yield the **same** challenge stream. For example, on a fresh sponge:

- observing `[5, 7]` and observing `[5, 7, 0, 0]` produced identical challenges.

The absorbed message boundary was not bound into the state, leaving the Fiat-Shamir transcript under-specified. In Plonky3 the transcript structure is fixed by each protocol (the verifier re-derives an identical observe/sample sequence), so this is not a known exploitable issue — but binding the length is cheap defense-in-depth and removes the footgun for any consumer that does not keep absorb lengths fixed.

## Change

- On an **absorb**, zero-fill the unused rate slots and add the absorbed length to the first capacity element, making each absorb prefix-free.
- A duplex with **no buffered input is a squeeze** and is left untouched, so squeezing behaviour is byte-for-byte unchanged.
- Widen the affected impl bounds from `F: Copy` to `F: Copy + PrimeCharacteristicRing` (needed for `ZERO`/`from_u8`); every field instantiation already satisfies this.
- Update the value-pinning tests and add `test_absorb_is_length_binding` as a direct regression guard (it fails on the old behaviour).

## ⚠️ Compatibility

This **changes the challenge stream** and is therefore a **breaking transcript change**: proofs produced before this change no longer verify after it. Prove→verify roundtrips remain valid (both sides use the new challenger), but any hardcoded known-answer fixtures in downstream crates may need regenerating.

## Testing

- `cargo test -p p3-challenger` → 52 passed
- `cargo clippy -p p3-challenger --all-targets` → clean


challenger/src/duplex_challenger.rs
```diff
@@ -84,13 +81,30 @@ where
     }
 
     pub(crate) fn duplexing(&mut self) {
-        assert!(self.input_buffer.len() <= RATE);
+        const {
+            assert!(
+                RATE < WIDTH,
+                "RATE must be less than WIDTH for the capacity length slot"
+            );
+        }
```
Nashtare: isn't this already asserted in constructors?

challenger/src/duplex_challenger.rs
```diff
@@ -84,13 +81,30 @@ where
...
+
+        let num_absorbed = self.input_buffer.len();
+        assert!(num_absorbed <= RATE);
 
-        // Overwrite the first r elements with the inputs.
+        // Write the buffered inputs into the leading rate slots.
```
Nashtare: this is still an `overwrite`, I'd keep the distinction

challenger/src/duplex_challenger.rs
```diff
@@ -65,13 +65,10 @@ where
 
 impl<F, P, const WIDTH: usize, const RATE: usize> DuplexChallenger<F, P, WIDTH, RATE>
 where
-    F: Copy,
+    F: Copy + PrimeCharacteristicRing,
```
Nashtare: Do we really need to add the `PrimeCharacteristingRing` bound everywhere? Can't we rely on `F::default()` like the initializer?

challenger/src/duplex_challenger.rs
tcoratger: Yes you are right, just removed

challenger/src/duplex_challenger.rs
tcoratger: Fixed

challenger/src/duplex_challenger.rs
tcoratger: Yes I've simplified and pushed the extra bound down only when needed

Nashtare: But can't we use `F::default()` in the duplexing call too, removing entirely the need for the extra bound?

tcoratger: Aaah I don't think this is possible here because we have `self.sponge_state[RATE] += F::from_u8(num_absorbed as u8);` as well which needs `PrimeCharacteristicRing` due to the addition stuff which is available in `PrimeCharacteristicRing`.

Nashtare: Ah right! Good point

Nashtare (approved):
LGTM thanks!

this-vishalsingh:
great!! more coming, I audited the whole repo.
Reviewed Commit: 47c5479ef4f9aadb4c0eceb898aa86466ea856c3

PR #1771, opened 2026-06-08
feat(lookup): bind multiplicity weight to its count

tcoratger:
@Nashtare I think this should solve the problem you described in https://github.com/Plonky3/Plonky3/pull/1748#pullrequestreview-4449412351 no?

## Context

Follow-up to #1748 (the LogUp multiplicity height-bound). @Nashtare noted that the bound

```
sum_i w_i * h_i < p
```

is only sound if `w_i` is a true per-row upper bound on a query's count — yet the bus API hardcoded `w_i = 1` while accepting an arbitrary multiplicity expression.

## The gap

A *multiplicity-aware* query — one row that performs `k > 1` lookups (e.g. a bulk range check) — contributes `k` to a provided entry's multiplicity, but the recorded weight stayed `1`:

```
                  real count/row   recorded weight
bulk lookup row:        5                 1
```

The height check then guards `1 * h` instead of `5 * h`, so with enough rows the true multiplicity can reach `p` and wrap, reopening the forgery the check exists to prevent. A `debug_assert` cannot close this: a variable count's per-row value is unknown at build time, so there is nothing to assert.

## The fix

The root cause is that `count` and `count_weight` were two independent arguments that could silently disagree. This PR binds them into one type, `Count<E>`, so they cannot desync:

| Call | Recorded | Notes |
|---|---|---|
| `send(.., 1)` | count 1, weight 1 | constant ⇒ bound automatic |
| `send(.., 5)` | count 5, weight 5 | constant ⇒ bound automatic |
| `send(.., Count::bounded(col, 1<<16))` | count `col`, weight 65535 | variable ⇒ bound mandatory |
| `send(.., col)` | **compile error** | bare expression has no bound |

There is intentionally no `From<Expr> for Count`, so forgetting the weight is now a compile error instead of a silent default to `1`. The common `±1` case stays a one-liner.

`Count` is threaded through `InteractionBuilder::push_interaction` and the four bus methods (`send`/`receive`/`lookup_key`/`table_entry`). `Count::provided` expresses the weight-0 table side; `Neg` handles the receive side (negate the count, keep the bound).

## Soundness note (documented)

The declared weight is a *promise*: the height check reads only public weights and heights and never inspects committed values. Soundness therefore requires the AIR to constrain `|count| <= weight` on every row. This obligation is now stated on `Count`, on `count_weight`, and on the bound function.

## Scope & testing

- Fully contained: only `lookup` and `batch-stark` use this API; no other crate implements `InteractionBuilder`.
- The repo's `FibAirLookups` test AIR (receives `multiplicity` copies per row) now declares the correct bound instead of `1`.
- `lookup` 34 tests (incl. 5 new `Count` tests) and `batch-stark` 41 tests pass; `cargo fmt` clean; 0 clippy warnings.

### DIFF
--- a/lookup/src/lib.rs
+++ b/lookup/src/lib.rs
@@ -9,9 +9,9 @@
 
 extern crate alloc;
 
-pub mod builder;
-pub mod bus;
-pub mod count;
+mod builder;
+mod bus;
+mod count;
 pub mod debug_util;
 pub mod folder;
 pub mod logup;
@@ -23,6 +23,7 @@ pub mod traits;
 mod types;
 
 pub use builder::{InteractionBuilder, SymbolicInteraction, SymbolicLocalInteraction};
+pub use bus::{LookupBus, PermutationCheckBus};
 pub use count::Count;
 pub use logup::LogUpGadget;
 pub use protocol::LookupProtocol;
--- a/lookup/src/lib.rs
+++ b/lookup/src/lib.rs
@@ -11,6 +11,7 @@ extern crate alloc;
 
 pub mod builder;
 pub mod bus;
+pub mod count;
 pub mod debug_util;
 pub mod folder;
 pub mod logup;
@@ -22,6 +23,7 @@ pub mod traits;
 mod types;
 
 pub use builder::{InteractionBuilder, SymbolicInteraction, SymbolicLocalInteraction};
+pub use count::Count;
 pub use logup::LogUpGadget;
 pub use protocol::LookupProtocol;
 pub use symbolic::InteractionSymbolicBuilder;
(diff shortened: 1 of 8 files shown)


lookup/src/lib.rs
```diff
@@ -11,6 +11,7 @@ extern crate alloc;
 
 pub mod builder;
 pub mod bus;
+pub mod count;
```
Nashtare: nit: do we need the module to be public?

Nashtare (approved):
Yes I think that fixes it!

lookup/src/lib.rs
tcoratger: Indeed for now even the three of them don't need to be public, so I've just made them private and we can change this in the future when needed.

PR #1774, opened 2026-06-08
fix(circle): reject oversized query-index widths

tcoratger:
## Summary

The Circle verifier derives the query-index width from prover-supplied folding-schedule data and feeds it straight into the challenger:

```rust
let log_max_height = total_log_reduction + params.log_blowup;
let index = challenger.sample_bits(log_max_height + folding.extra_query_index_bits());
```

The round count equals `commit_phase_commits.len()`, which is attacker-controlled, so a malformed proof can drive the width past the challenger's internal bound. `sample_bits` then hits an `assert!` (`(1 << bits) < F::ORDER_U64`) and **panics** instead of returning a structured error — an in-process DoS for anyone verifying untrusted proofs. Generic FRI already guards the analogous case via `FriError::GlobalMaxHeightTooLarge`; Circle did not.

## Fix

Bound the width before sampling and return the existing typed error:

```rust
let num_index_bits = log_max_height + folding.extra_query_index_bits();
if num_index_bits >= Val::CIRCLE_TWO_ADICITY {
    return Err(FriError::GlobalMaxHeightTooLarge {
        log_global_max_height: num_index_bits,
        two_adicity: Val::CIRCLE_TWO_ADICITY,
    });
}
```

- The verifier bound is tightened from `Val: Field` to `Val: ComplexExtendable` to reach `CIRCLE_TWO_ADICITY`; the only caller (the PCS layer) already satisfies this.

## Why `>=`, not the two-adic `>`

This is the load-bearing subtlety, and it differs from generic FRI on purpose:

- Generic FRI checks `log_global_max_height > Val::TWO_ADICITY`. For a two-adic field, `p > 2^TWO_ADICITY`, so a width of exactly `TWO_ADICITY` bits *is* sampleable — strict `>` is right.
- For Circle/M31, `CIRCLE_TWO_ADICITY = 31` is the two-adicity of the **circle group order** `p + 1 = 2^31`, while the field order is `p = 2^31 − 1` — one *short* of the group. A width of exactly 31 bits is therefore **unsampleable** (`1 << 31 > p`), so the guard must reject `>=` to close a one-bit panic window.

This generalizes: for any `ComplexExtendable` field (`p ≡ −1 mod 2^v`), widths up to `v − 1` are always sampleable and `>= v` rejects exactly the unsampleable ones, never an honest proof (honest M31 proofs cap at width 30).

## Testing

- New `reject_global_max_height_too_large`: inflates the round count until the width reaches the bound and asserts the typed error rather than a panic.
- `cargo test -p p3-circle --lib` → 28 passed (1 new).
- `cargo fmt -p p3-circle --check` and `cargo clippy -p p3-circle --all-targets` → clean.

Sibling of #1772 (zero-query guard); both bring Circle's malformed-proof handling in line with generic FRI. Independent changes, no overlap.

### DIFF
--- a/circle/src/verifier.rs
+++ b/circle/src/verifier.rs
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 use itertools::Itertools;
 use p3_challenger::{CanObserve, FieldChallenger, GrindingChallenger};
 use p3_commit::{BatchOpeningRef, Mmcs};
+use p3_field::extension::ComplexExtendable;
 use p3_field::{ExtensionField, Field};
 use p3_fri::verifier::FriError;
 use p3_fri::{FriFoldingStrategy, FriParameters};
@@ -21,7 +22,7 @@ pub fn verify<Folding, Val, Challenge, M, Challenger>(
     ) -> Result<Vec<(usize, Challenge)>, Folding::InputError>,
 ) -> Result<(), FriError<M::Error, Folding::InputError>>
 where
-    Val: Field,
+    Val: ComplexExtendable,
     Challenge: ExtensionField<Val>,
     M: Mmcs<Challenge>,
     Challenger: FieldChallenger<Val> + GrindingChallenger + CanObserve<M::Commitment>,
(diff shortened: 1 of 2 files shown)


circle/src/verifier.rs
```diff
@@ -165,9 +159,25 @@ where
...
     let log_max_height = total_log_reduction + params.log_blowup;
 
+    // Invariant: the query-index width fits the circle group of order 2^CIRCLE_TWO_ADICITY.
+    //
+    //     num_index_bits = sum(log_arities) + log_blowup + extra_query_index_bits
+    //     field order    = 2^CIRCLE_TWO_ADICITY - 1   (one short of the group order)
+    //     => a width of CIRCLE_TWO_ADICITY bits is unsampleable
+    //
+    // A malformed arity schedule inflates the round count, hence the width.
+    // A typed error keeps the challenger's bound assertion from aborting the verifier.
```
Nashtare:
```suggestion
```

Nashtare (approved):

this-vishalsingh:
Looks good

PR #1779, opened 2026-06-09
bench(lookup): end-to-end batch-STARK lookup benchmark harness

tcoratger:
## Why

We have a line of lookup improvements coming (challenge unification, column grouping/packing, sparse-row skipping). Each one *sounds* like a win — but a STARK proof has several costs that pull against each other, so an apparent improvement in one dimension can quietly regress another.

Today there is no way to settle that end-to-end. `lookup/benches/logup.rs` only times permutation-trace generation, and `batch-stark/benches/prove_batch.rs` proves a lookup-free Fibonacci AIR. Neither reports proof size or the committed auxiliary-column count — the very things a packing optimization is supposed to move.

So before optimizing, we build the ruler. This PR adds an end-to-end lookup benchmark that turns a lookup-layer change into hard numbers, so every follow-up can show its before/after on the same scale rather than rely on intuition.

This is a pure addition: **no library code changes, no proof fixtures touched.**

## What it measures

It runs a parametric lookup workload through the full prove → verify pipeline and reports:

| metric | meaning |
|---|---|
| prove time | criterion-timed |
| verify time | criterion-timed |
| aux columns | committed permutation columns the lookups add (the packing target) |
| proof size | serialized proof, in KB |
| peak heap | high-water bytes during one prove |

A small config matrix sweeps the knobs so each effect is isolated:

- `n_interactions` — number of lookups
- `tuple_width` — field elements per message
- `fold` — lookups merged into one committed column
- `active_period` — one active row every N rows (row-sparsity)
- `n_airs` — batch size

## Example output

The harness prints a metrics table first, then criterion runs the rigorous prove/verify timings. A sample run on BabyBear (testing FRI params):

```
case         logH     n   tw  fold  period  airs aux_cols  proof(KB)  peak(MB) prove(ms) verify(ms)
smoke           8     1    1     1       1     1        2        6.2       0.4       3.0      0.27
n=1            14     1    1     1       1     1        2       13.8      26.8     168.8      0.54
n=4            14     4    1     1       1     1        5       14.1      31.3     220.4      0.57
n=8            14     8    1     1       1     1        9       14.5      37.3     298.5      0.61
n=16           14    16    1     1       1     1       17       15.4      49.3     453.2      0.70
fold=1         14     8    1     1       1     1        9       14.5      37.3     295.8      0.63
fold=2         14     8    1     2       1     1        5       14.4      35.3     322.8      0.61
fold=4         14     8    1     4       1     1        3       14.4      37.3     445.1      0.63
tw=2           14     4    2     1       1     1        5       14.3      32.3     240.1      0.58
tw=4           14     4    4     1       1     1        5       14.7      34.3     268.7      0.59
period=4       14     4    1     1       4     1        5       14.2      31.3     217.3      0.65
period=16      14     4    1     1      16     1        5       14.2      31.3     216.7      0.57
airs=4         14     4    1     1       1     4       20       15.7      58.3     585.2      0.76
```

Reading the `fold` block already pays off: merging 8 lookups from 9 committed columns down to 3 (`fold=1` → `fold=4`) leaves proof size flat and *raises* prove time (296 → 445 ms), because folding trades columns for a higher-degree constraint. A column-count-only metric would have called that a win; the end-to-end ruler shows it is not — and points the real optimization at mutual-exclusivity-aware packing, where the degree does not grow. The `period` rows show prover cost is sparsity-blind today, quantifying the headroom for a future sparse-row skip.

## Design notes

- **Workload** is a balanced range-check AIR: each lookup queries a key and provides the same key per active row, so any trace balances to a zero terminal and `verify` always passes. Cost is set by structure (columns, height, width), not by the values — which is exactly what the lookup layer changes.
- **`fold` stands in for grouping.** It merges lookups via one multi-tuple local interaction, producing the exact aux-trace shape a same-bus grouping would, so a grouped-vs-ungrouped comparison reads straight off the `fold` sweep.
- **Peak heap** uses a tracking global allocator that only counts while armed; disarmed it adds one relaxed load per allocation, leaving criterion's timing essentially unperturbed.
- **Correctness gate:** every case runs `verify_batch(...).unwrap()`, so a malformed workload fails the bench.
- Heavy `log_height = 18` points are included and flagged; comment them out for a faster local run.

## Running it

```
cargo bench -p p3-batch-stark --bench lookup
```

### DIFF
--- a/batch-stark/benches/lookup.rs
+++ b/batch-stark/benches/lookup.rs
@@ -1,0 +1,570 @@
+//! End-to-end lookup benchmark harness.
+//!
+//! Drives a parametric lookup workload through the full batch-STARK pipeline.
+//!
+//! Reports the numbers a lookup-layer change actually moves:
+//!
+//!     prove time       criterion-timed
+//!     verify time      criterion-timed
+//!     aux columns      committed permutation columns (the packing target)
+//!     proof size       serialized bytes
+//!     peak heap        high-water bytes during one prove
+//!
+//! The workload is a balanced range-check AIR with four knobs:
+//!
+//!     n_interactions   how many lookups the AIR declares
+//!     tuple_width      field elements per message
+//!     fold             lookups merged into one committed column
+//!     active_period    one active row every `active_period` rows (row-sparsity)
+//!
+//! The `fold` knob stands in for grouping:
+//!
+//!     merge lookups  → one multi-tuple local interaction
+//!     same aux-trace shape as a same-bus grouping
+//!     so grouped-vs-ungrouped reads off the `fold` sweep
+//!
+//! Running it:
+//!
+//!     cargo bench -p p3-batch-stark --bench lookup
+//!
+//! The metrics table prints first, then the criterion timings.
+//! Comment out entries in `CASES` to shorten a local run.
+
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::sync::atomic::{AtomicBool, AtomicIsize, AtomicUsize, Ordering};
+use std::time::Instant;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_air::{Air, BaseAir, PermutationAirBuilder, WindowAccess};
+use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+use p3_batch_stark::{ProverData, StarkInstance, prove_batch, verify_batch};
+use p3_challenger::DuplexChallenger;
+use p3_commit::ExtensionMmcs;
+use p3_dft::Radix2DitParallel;
+use p3_field::extension::BinomialExtensionField;
+use p3_field::{Field, PrimeCharacteristicRing};
+use p3_fri::{FriParameters, TwoAdicFriPcs};
+use p3_lookup::InteractionBuilder;
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
+use p3_uni_stark::StarkConfig;
+use rand::rngs::SmallRng;
+use rand::{RngExt, SeedableRng};
+
+type Val = BabyBear;
+type Challenge = BinomialExtensionField<Val, 4>;
+type Perm = Poseidon2BabyBear<16>;
+type MyHash = PaddingFreeSponge<Perm, 16, 8, 8>;
+type MyCompress = TruncatedPermutation<Perm, 2, 8, 16>;
+type ValMmcs =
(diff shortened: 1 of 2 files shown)


batch-stark/benches/lookup.rs
```diff
@@ -0,0 +1,570 @@
...
+
+fn make_config() -> MyConfig {
+    let mut rng = SmallRng::seed_from_u64(1337);
+    let perm = Perm::new_from_rng_128(&mut rng);
+    let hash = MyHash::new(perm.clone());
+    let compress = MyCompress::new(perm.clone());
+    let val_mmcs = ValMmcs::new(hash, compress, 0);
+    let challenge_mmcs = ChallengeMmcs::new(val_mmcs.clone());
+    let dft = Dft::default();
+    let fri_params = FriParameters::new_testing(challenge_mmcs, 2);
```
Nashtare: `new_testing` requires only 2 FRI queries, although this is fine if we compare across consistent configs, perhaps it'd be worth raising to have the proof size display below consistent with practical settings? It won't impact proving time much anyway

Nashtare: i.e. perhaps `new_benchmark` or `new_benchmark_high_arity`, or a fully manual one if we wanna keep the `log_final_poly_len: 2` you specified here

batch-stark/benches/lookup.rs
```diff
@@ -0,0 +1,570 @@
...
+    // Fold: 8 lookups, more merged per column → fewer columns, higher degree.
+    Case {
+        name: "fold=1",
+        log_height: 14,
+        n_interactions: 8,
+        tuple_width: 1,
+        fold: 1,
+        active_period: 1,
+        n_airs: 1,
+    },
```
Nashtare: I think we can remove it with a comment, that's exactly `n=8` already

```suggestion
    // Case `fold=1` contained in case `n=8` above
```

batch-stark/benches/lookup.rs
```diff
@@ -0,0 +1,570 @@
...
+//! The `fold` knob stands in for grouping:
+//!
+//!     merge lookups  → one multi-tuple local interaction
+//!     same aux-trace shape as a same-bus grouping
+//!     so grouped-vs-ungrouped reads off the `fold` sweep
+//!
+//! Running it:
+//!
+//!     cargo bench -p p3-batch-stark --bench lookup
+//!
```
Nashtare, line 29: given the heaviness of the bench, may be worth mentioning the use of `--features parallel`?


```suggestion
//! Running it:
//!
//!     cargo bench -p p3-batch-stark --bench lookup
//!
//! and optionally with `--features parallel` argument.
//!
```

Nashtare (approved):

batch-stark/benches/lookup.rs
tcoratger: Fixed, just used `FriParameters::new_benchmark`

PR #1782, opened 2026-06-09
perf(challenger): precompute the grind state for MultiField32Challenger

tcoratger:
## What

Ports the `DuplexChallenger` packed-grind strategy to `MultiField32Challenger::grind`.

The old implementation brute-forced PoW candidates with `self.clone().check_witness(...)` per candidate, paying for each one:

- a full challenger clone (two `Vec` allocations),
- a re-pack of all pending scalars (including the candidate-independent chunks),
- the permutation,
- `RATE` BigUint cell splits in the squeeze refill — of which only one limb of one cell is actually checked.

The new implementation builds the candidate-independent pre-permutation sponge state **once** (constant packed chunks, zero tail, length tag, carried capacity). Each candidate then costs one stack copy, one packed-digit write, one permutation, and a single-cell limb split — no clone, no heap allocation.

The identity preserved is exactly `check_witness = observe(w); sample_bits(bits)`: the witness digit lands at weight `2^(rb·pos)` in packed slot `chunk_idx` (little-endian Horner packing), and the checked challenge is the last limb split from the last rate cell. `grind` still finishes with `assert!(self.check_witness(bits, witness))`, so the real transcript advances through the unmodified verifier path.

The trait impl moves from `grinding_challenger.rs` into `multi_field_challenger.rs` so the sponge fields stay private (same precedent as `DuplexChallenger`, whose grind lives next to its `pub` fields). Public API, bounds, and transcript behavior are unchanged.

## Benchmarks

Release mode, serial search (no rayon), same transcript prefix per round. Both paths found **identical witnesses**, which doubles as an end-to-end functional check on real permutations.

`MultiField32Challenger<BabyBear, Bn254, Poseidon2Bn254<3>, 3, 2>` (recursion config, permutation-dominated), `bits = 10`:

| round | old (clone + full check) | new (precomputed state) | speedup |
|---|---|---|---|
| 0 | 63.27 ms | 48.86 ms | 1.29x |
| 1 | 11.94 ms | 9.26 ms | 1.29x |
| 2 | 10.83 ms | 8.35 ms | 1.30x |

`MultiField32Challenger<BabyBear, Goldilocks, Poseidon2Goldilocks<8>, 8, 4>` (cheap permutation, overhead-dominated), `bits = 14`:

| round | old | new | speedup |
|---|---|---|---|
| 0 | 17.81 ms | 8.04 ms | 2.22x |
| 1 | 5.52 ms | 2.49 ms | 2.22x |
| 2 | 83.38 ms | 38.06 ms | 2.19x |

<details>
<summary>Benchmark harness used (temporary integration test, not committed)</summary>

```rust
use std::time::Instant;

use p3_baby_bear::BabyBear;
use p3_bn254::{
    BN254_POSEIDON2_HALF_FULL_ROUNDS, BN254_POSEIDON2_PARTIAL_ROUNDS_3, Bn254, Poseidon2Bn254,
};
use p3_challenger::{CanObserve, GrindingChallenger, MultiField32Challenger};
use p3_goldilocks::{Goldilocks, Poseidon2Goldilocks};
use rand::SeedableRng;
use rand::rngs::SmallRng;

type F = BabyBear;

#[test]
fn grind_timing() {
    let mut rng = SmallRng::seed_from_u64(1);
    let perm = Poseidon2Bn254::<3>::new_from_rng(
        2 * BN254_POSEIDON2_HALF_FULL_ROUNDS,
        BN254_POSEIDON2_PARTIAL_ROUNDS_3,
        &mut rng,
    );
    let base = MultiField32Challenger::<F, Bn254, _, 3, 2>::new(perm).unwrap();
    let bits = 10;

    for round in 0u32..3 {
        let mut c1 = base.clone();
        c1.observe(F::from_u32(round));
        let t = Instant::now();
        let w_new = c1.grind(bits);
        let new_t = t.elapsed();

        let mut c2 = base.clone();
        c2.observe(F::from_u32(round));
        let t = Instant::now();
        // Old implementation shape: serial scan, full challenger clone + check per candidate.
        let w_old = (0u32..)
            .map(F::from_u32)
            .find(|w| c2.clone().check_witness(bits, *w))
            .unwrap();
        assert!(c2.check_witness(bits, w_old));
        let old_t = t.elapsed();

        println!("round {round}: new = {new_t:?} (w = {w_new:?}), old = {old_t:?} (w = {w_old:?})");
    }
}

#[test]
fn grind_timing_goldilocks() {
    let mut rng = SmallRng::seed_from_u64(1);
    let perm = Poseidon2Goldilocks::<8>::new_from_rng_128(&mut rng);
    let base = MultiField32Challenger::<F, Goldilocks, _, 8, 4>::new(perm).unwrap();
    let bits = 14;

    for round in 0u32..3 {
        let mut c1 = base.clone();
        c1.observe(F::from_u32(round));
        let t = Instant::now();
        let w_new = c1.grind(bits);
        let new_t = t.elapsed();

        let mut c2 = base.clone();
        c2.observe(F::from_u32(round));
        let t = Instant::now();
        let w_old = (0u32..)
            .map(F::from_u32)
            .find(|w| c2.clone().check_witness(bits, *w))
            .unwrap();
        assert!(c2.check_witness(bits, w_old));
        let old_t = t.elapsed();

        println!("round {round}: new = {new_t:?} (w = {w_new:?}), old = {old_t:?} (w = {w_old:?})");
    }
}
```

Run with:

```
cargo test -p p3-challenger --release --test grind_timing_tmp -- --nocapture
```

</details>

## Tests

- `prop_pow_check_matches_check_witness` — proptest pinning the precomputed-state predicate to the `clone().check_witness` reference over random pending transcripts, candidates, and bit counts.
- `test_grind_advances_state_like_direct_check` — pending counts `{0, 1, 4, 7}` cover every packing geometry: witness opening a chunk, completing a chunk, opening a later chunk, and filling the whole batch (flush fires inside `observe`). Asserts full state equality (sponge state, both inner buffers, `f_buffer`, squeeze queue) against a direct `check_witness`.
- `test_grind_advances_state_like_direct_check_bn254` — pins the multi-limb squeeze extraction path (`squeeze_n > 1`) that the Goldilocks fixture cannot reach.
- Existing `grind(0)` no-op and oversized-`bits` panic tests unchanged and passing.

`cargo test -p p3-challenger` (59 passed), `cargo clippy -p p3-challenger --all-targets` (clean), `cargo fmt --check` (clean); `p3-fri`, `p3-uni-stark`, `p3-batch-stark`, `p3-whir` compile.

## Not done (deliberately)

- **Serializing challengers**: their grind is hash-bound — `CryptographicHasher` exposes no incremental API, so each candidate needs one full Keccak/Blake3 over the buffered transcript regardless; the clone is noise next to the hash. The real fix is an incremental-hash API, a separate design change.
- **Packed-lane search over `PF::Packing`**: would require adding `CryptographicPermutation<[PF::Packing; WIDTH]>` to the impl bounds — a breaking API change, and a no-op for the dominant Bn254 use case (unit packing).

### DIFF
--- a/challenger/src/grinding_challenger.rs
+++ b/challenger/src/grinding_challenger.rs
@@ -1,10 +1,13 @@
-use p3_field::{Field, PackedValue, PrimeCharacteristicRing, PrimeField64};
+use p3_field::{
+    Field, PackedValue, PrimeCharacteristicRing, PrimeField, PrimeField32, PrimeField64,
+};
 use p3_maybe_rayon::prelude::*;
 use p3_symmetric::CryptographicPermutation;
 use tracing::instrument;
 
 use crate::{
-    CanObserve, CanSampleBits, CanSampleUniformBits, DuplexChallenger, UniformSamplingField,
+    CanObserve, CanSampleBits, CanSampleUniformBits, DuplexChallenger, MultiField32Challenger,
+    UniformSamplingField,
 };
 
 /// Trait for challengers that support proof-of-work (PoW) grinding.
@@ -280,6 +283,49 @@ where
         // Run the check one last time on the *original* challenger to update its state
         // and confirm the witness is valid.
         assert!(check_fn(self, witness));
+        witness
+    }
+}
+
+impl<F, PF, P, const WIDTH: usize, const RATE: usize> GrindingChallenger
+    for MultiField32Challenger<F, PF, P, WIDTH, RATE>
+where
+    F: PrimeField32,
+    PF: PrimeField,
+    P: CryptographicPermutation<[PF; WIDTH]>,
+{
+    type Witness = F;
+
+    #[instrument(name = "grind for proof-of-work witness", skip_all)]
+    fn grind(&mut self, bits: usize) -> Self::Witness {
+        assert!(bits < (usize::BITS as usize), "bit count must be valid");
+        // Evaluate the bound in `u64` to keep the shift within its type width.
+        // A `u32` shift by `bits >= 32` would wrap and accept a trivial proof-of-work.
+        assert!(
+            (1u64 << bits) < F::ORDER_U64,
+            "requested bit count must fit within the field order"
+        );
+
+        // Trivial case: 0 bits mean no PoW is required and any witness is valid.
+        if bits == 0 {
+            return F::ZERO;
+        }
+
+        // The candidate-independent transcript work happens once inside `pow_check_fn`.
+        // The parallel search then runs one stack-state permutation per candidate.
+        let witness = {
+            let accepts = self.pow_check_fn(bits);
+            (0..F::ORDER_U32)
+                .into_par_iter()
+                .find_any(|&candidate| accepts(candidate))
+                .expect("failed to find proof-of-work witness")
+        };
+        // candidate < F::ORDER_U32 by construction so this is safe.
+        let witness = unsafe { F::from_canonical_unchecked(witness) };
+
+        // Re-run the standard verifier logic to validate the witness and
+        // advance the real transcript state.
+        assert!(self.check_witness(bits, witness));
         witness
     }
 }
--- a/challenger/src/grinding_challenger.rs
+++ b/challenger/src/grinding_challenger.rs
@@ -1,13 +1,10 @@
-use p3_field::{
-    Field, PackedValue, PrimeCharacteristicRing, PrimeField, PrimeField32, PrimeField64,
-};
+use p3_field::{Field, PackedValue, PrimeCharacteristicRing, PrimeField64};
 use p3_maybe_rayon::prelude::*;
 use p3_symmetric::CryptographicPermutation;
 use tracing::instrument;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


challenger/src/grinding_challenger.rs
```diff
@@ -286,40 +283,3 @@ where
         witness
     }
 }
-
-impl<F, PF, P, const WIDTH: usize, const RATE: usize> GrindingChallenger
-    for MultiField32Challenger<F, PF, P, WIDTH, RATE>
-where
-    F: PrimeField32,
-    PF: PrimeField,
-    P: CryptographicPermutation<[PF; WIDTH]>,
```
Nashtare, line 295: I think we would rather keep it in grinding_challenger module, to remain consistent with the other implementations?

Nashtare (approved):
LGTM! I would just keep it in the original module for consistency, but no logic change requirement, purely cosmetic.

challenger/src/grinding_challenger.rs
tcoratger: Fixed

PR #1789, opened 2026-06-10
refactor: cleanup WHIR API

### LINKED ISSUE #1733: [WHIR] Cleanup Public API
The Public API contains a bit too much at the moment, although this is flawed because of the ongoing work around HVZK support. 
We should consider cleaning that up before the release.

Nashtare:
closes #1733


whir/src/fiat_shamir/pattern.rs
```diff
@@ -83,6 +83,7 @@ pub enum Observe {
     /// as `claimed_sum - c_0`.
     SumcheckPoly,
     /// Evaluation answers at STIR query positions.
+    #[allow(dead_code)]
```
tcoratger: Is it normal that we have an allowance here?

tcoratger (approved):
LGTM, just a small question, non blocking :)

whir/src/fiat_shamir/pattern.rs
Nashtare: oops, removed!

PR #1791, opened 2026-06-10
fix(batch-stark): reject out-of-domain point inside the trace domain

tcoratger:
## Problem

The batch-STARK verifier evaluates Lagrange selectors at the Fiat–Shamir point `zeta` via `Domain::selectors_at_point`, which computes `1 / Z_H(zeta)`. When `zeta` lands **inside** the trace domain, `Z_H(zeta) = 0` and the field inverse panics (`"Tried to invert zero"`), aborting the verifier rather than returning a clean error.

The uni-STARK verifier already guards this (`uni-stark/src/verifier.rs`, returning `VerificationError::OodPointInDomain`); the batch verifier did not — a parity gap.

## Fix

Add the guard at the start of `VerifierData::verify_constraints_with_lookups`, before any selector inversion, so **every** instance is protected at the exact site that divides by `Z_H(zeta)`:

```rust
if self.trace_domain.vanishing_poly_at_point(self.zeta).is_zero() {
    return Err(VerificationError::OodPointInDomain);
}
```

- No new error variant: `OodPointInDomain` already exists on the shared `p3_uni_stark::VerificationError` (auto-converted via `BatchVerificationError`'s `#[from]`).
- No visibility changes: `VerifierData`'s fields stay `pub(crate)`.

## Severity

Honest Fiat–Shamir sampling reaches this only with probability `|H| / |EF|`, and the failure mode is a panic (the verifier *rejects*, never accepts a bad proof). So this is a **verifier-robustness fix** (DoS-by-panic on a malformed/adversarial proof) that brings batch-STARK to parity with uni-STARK — not a soundness change.

## Test

`zeta` is Fiat–Shamir-derived and cannot be steered into the trace domain through the public `verify_batch` API, so the test lives in-crate next to the guard (where `VerifierData`'s `pub(crate)` fields are accessible). It builds a `VerifierData` with `zeta = domain.first_point()` (a domain point, so `Z_H(zeta) = 0`) and asserts a clean `OodPointInDomain` instead of the previous panic.

- `cargo test -p p3-batch-stark` — green
- `cargo clippy -p p3-batch-stark --all-targets` — clean
- `cargo fmt -p p3-batch-stark --check` — clean

### DIFF
--- a/batch-stark/src/verifier/data.rs
+++ b/batch-stark/src/verifier/data.rs
@@ -2,7 +2,7 @@ use core::fmt::Debug;
 
 use p3_air::{Air, RowWindow};
 use p3_commit::PolynomialSpace;
-use p3_field::PrimeCharacteristicRing;
+use p3_field::{Field, PrimeCharacteristicRing};
 use p3_lookup::folder::VerifierConstraintFolderWithLookups;
 use p3_lookup::{Lookup, LookupProtocol};
 use p3_matrix::dense::RowMajorMatrixView;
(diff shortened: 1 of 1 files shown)


Nashtare (commented):
I would suggest not making every single field public for the sole purpose of adding a check in an integration test tbh

Nashtare (approved):

PR #1794, opened 2026-06-10
fix(merkle-tree): bound pruned-proof scratch early and enrich the error set

tcoratger:
## Problem

`Mmcs::verify_batch_pruned` allocated per-path scratch sized by **prover-supplied** counts — `restore_paths` expands the compact proof to `n_unique × full_sibling_count` digests, then `reps`/`digests`/walk buffers are sized by `n_unique` — all *before* the geometry check (`leaf_index < max_height`) that bounds those counts. A malformed opening could force a large allocation before being rejected (CWE-789).

`verify_batch_pruned` has no in-tree caller (only benches/tests); it's a public API method, so this hardens it for downstream users.

## Fix 1 — early bound

A valid pruned opening has strictly-ascending distinct leaf indices in `[0, max_height)`, so it holds at most `max_height` unique paths. Reject an oversized count up front, using verifier-known `max_height`, before any allocation:

```rust
if pruned_opening.pruned_proof.paths.len() > max_height {
    return Err(PrunedProofError::TooManyUniquePaths { got: ..., max_height }.into());
}
```

## Fix 2 — richer error set

The pruned path collapsed every distinct failure into one opaque `MerkleTreeError::MalformedPrunedProof`. Replaced with `MalformedPrunedProof(PrunedProofError)`, where `PrunedProofError` names each failure with diagnostic fields:

| Variant | Fields |
|---|---|
| `TooManyUniquePaths` | `got, max_height` |
| `SiblingCountMismatch` | `expected, got` |
| `OriginalOrderOutOfRange` | `slot, num_paths` |
| `InconsistentDuplicateOpenings` | `slot` |
| `UnreferencedPath` | `slot` |
| `NonAscendingLeaves` | `position, index` |
| `PathQueryCountMismatch` | `num_paths, num_queries` |

`restore_paths` now returns `Result<_, PrunedProofError>` instead of `Option`, so the specific reason propagates instead of collapsing to `None`.

## Notes

- No code outside `merkle-tree` constructs or matches `MalformedPrunedProof`, so the unit→tuple variant change is contained; `p3-fri`/`p3-commit` build unchanged.
- Tests assert the full variant **and** inner fields (e.g. `TooManyUniquePaths { got: 9, max_height: 8 }`, `InconsistentDuplicateOpenings { slot: 0 }`, `SiblingCountMismatch { expected: 3, got: 10 }`).

## Verification

- `cargo test -p p3-merkle-tree` — 102 passed
- `cargo clippy -p p3-merkle-tree --all-targets` — clean
- `cargo fmt -p p3-merkle-tree --check` — clean

### DIFF
--- a/merkle-tree/src/lib.rs
+++ b/merkle-tree/src/lib.rs
@@ -10,6 +10,6 @@ mod pruning;
 
 pub use hiding_mmcs::*;
 pub use merkle_tree::MerkleTree;
-pub use mmcs::{MerkleTreeError, MerkleTreeMmcs, PrunedBatchOpening};
+pub use mmcs::{MerkleTreeError, MerkleTreeMmcs, PrunedBatchOpening, PrunedProofError};
 pub use p3_symmetric::MerkleCap;
 pub use pruning::{PrunedMerklePaths, PrunedPath};
--- a/merkle-tree/src/mmcs.rs
+++ b/merkle-tree/src/mmcs.rs
@@ -37,8 +37,8 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use crate::MerkleTreeError::{
-    CapMismatch, EmptyBatch, IncompatibleHeights, IndexOutOfBounds, MalformedPrunedProof,
-    WrongBatchSize, WrongHeight, WrongWidth,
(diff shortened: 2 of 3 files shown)


Nashtare (approved):
LGTM!

PR #1797, opened 2026-06-10
perf(circle): batch and parallelize the PCS open phase

Nashtare:
Make the `open` span for circle STARK around −36% faster

### DIFF
--- a/circle/src/deep_quotient.rs
+++ b/circle/src/deep_quotient.rs
@@ -173,52 +173,6 @@ impl<F: ComplexExtendable, M: Matrix<F>> CircleEvaluations<F, M> {
             .rowwise_packed_dot_product::<EF>(&packed_alpha_powers)
             .collect()
     }
-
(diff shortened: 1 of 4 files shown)


circle/src/deep_quotient.rs
```diff
@@ -120,55 +193,31 @@ impl<F: ComplexExtendable, M: Matrix<F>> CircleEvaluations<F, M> {
     /// # Returns
     ///
     /// A vector of DEEP quotient values, one for each row in the matrix.
-    #[instrument(skip_all, fields(dims = %self.values.dimensions()))]
+    #[cfg(test)]
     pub(crate) fn deep_quotient_reduce<EF: ExtensionField<F>>(
```
tcoratger: Can't we move this function directly to the test module then? Just to maintain a clean core of the file

tcoratger (approved):
Just super minimal comment, non blocker

circle/src/deep_quotient.rs
Nashtare: Totally!

PR #1799, opened 2026-06-10
feat(lookup): degree-budget-aware same-bus column packing

tcoratger:
## What

Fold global lookups that share a bus into shared fraction columns, **capped at the AIR's spare quotient-degree budget**. Today every global interaction gets its own committed auxiliary column; this packs several onto one column when — and only when — doing so leaves the quotient unchanged.

## Why it never regresses

The quotient chunk count is `2^ceil(log2((max_degree + is_zk).max(2) - 1))`. Folding interactions into one column only ever raises that column's fraction-pin degree. So the safe rule is:

> Fold same-bus globals into a column only while its degree stays `<= 2^B + 1 - is_zk`, where `B` is the *unpacked* AIR's `log_num_quotient_chunks`.

This keeps the quotient bucket **identical** (a debug assert enforces it), so quotient work — the expensive part — is unchanged, and only auxiliary columns drop.

- **No spare budget** (e.g. a pure range-check table, degree 2): budget folds nothing → byte-identical to before.
- **Spare budget** (any AIR with a degree-≥3 constraint, the common case): folds for free.

Both prover and verifier recompute the fold deterministically from the public AIRs, so the layout is never transmitted and soundness is untouched — folding only reorganizes the same per-bus rational sum.

## Measured (end-to-end bench, BabyBear, 100-query FRI)

Each case is measured packed and unpacked in one table.

| case | aux cols | proof (KB) | peak (MB) | prove |
|---|---|---|---|---|
| `n=16`, deg 0 (no budget) | 33 = 33 | 686.2 = 686.2 | 32.9 = 32.9 | — *(identical)* |
| `deg3_n=16` | 33 → **17** | 688.3 → **629.9** | 33.7 → **25.7** | 403 → **287 ms** *(criterion, ~29%)* |
| `deg3_n=32` (h=2^18) | 65 → **33** | 1074 → **957** | 944 → **591** | 12.5 → **8.2 s** |

Every no-budget case is byte-identical packed vs unpacked (no regression); every spare-budget case improves on all metrics.

## Changes

- **`lookup`**: `Lookups::pack_same_bus(gadget, max_degree)` — pure fold of same-bus globals into multi-tuple lookups under a degree cap; sums weights, reassigns contiguous columns, leaves local lookups untouched. Unit-tested incl. a full `n × budget` invariant sweep.
- **`batch-stark`**: apply the fold at lookup-build time with the is_zk-aware budget, guarded by a debug assert that the quotient chunk count is preserved.
- **bench**: extend the end-to-end harness with a `base_degree` knob and a packed-vs-unpacked column, so the no-regression guard and the win read off the same table.

## Scope

This is the provably-safe core (Miden #2962's column packing, restricted to same-bus folding within the spare degree budget). Follow-ups, intentionally left out: mutual-exclusivity-aware `degree = max` grouping (needs a disjoint-selector API), prover flag-zero skip, and the typed bus-id cleanup.

### DIFF
--- a/batch-stark/benches/lookup.rs
+++ b/batch-stark/benches/lookup.rs
@@ -10,24 +10,29 @@
 //!     proof size       serialized bytes
 //!     peak heap        high-water bytes during one prove
 //!
-//! The workload is a balanced range-check AIR with four knobs:
+//! The workload is one balanced lookup AIR with these knobs:
 //!
-//!     n_interactions   how many lookups the AIR declares
+//!     n_interactions   query/provide pairs, all on one shared bus
 //!     tuple_width      field elements per message
-//!     fold             lookups merged into one committed column
+//!     base_degree      degree of an optional base constraint, or 0 for none
 //!     active_period    one active row every `active_period` rows (row-sparsity)
+//!     n_airs           identical instances proven in one batch
 //!
-//! The `fold` knob stands in for grouping:
+//! Every case is measured twice, with same-bus packing on and off:
 //!
-//!     merge lookups  → one multi-tuple local interaction
-//!     same aux-trace shape as a same-bus grouping
-//!     so grouped-vs-ungrouped reads off the `fold` sweep
+//!     base_degree 0  → no spare budget → packing folds nothing → both rows match
+//!     base_degree d  → spare budget    → packing folds → fewer columns, less cost
+//!
+//! So the table proves the two claims side by side:
+//!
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
Thanks! Looking good! 
I don't think we're having any folding actually checked in CI though (not in batch-stark `simple` integration test module at least, but this could be augmented in a follow-up PR if I'm not corrected!

PR #1802, opened 2026-06-10
perf: vectorize packed field

Nashtare:
## Motivation

  During quotient evaluation, the constraint folder evaluates the AIR expression DAG over a single packed vector per row group.
  Each expression forms one dependency chain, which cannot hide the latency of packed modular
  multiplication (for instance NEON M31 mul:
  ~10 cycles latency at 1.25 cycles/vector throughput → ~8 independent chains needed to saturate the pipes).
  
  Inspired from Stwo's `Vectorized<PackedM31, 2>`.
  
  ## Changes
  
 - `p3-field`: adds a pair of wrapper types that bundle several packed vectors (base-field and extension-field) into a single value whose arithmetic is applied to all of them in lockstep. They expose the same ring and packing
  interfaces as ordinary packed values, so existing machinery — loading trace rows, slicing selectors, and the specialized SIMD dot products used when folding constraints with alpha powers — keeps working unchanged on the
  wider type.
  
 - `p3-uni-stark`: adds a constraint folder built on these wrapper types, and switches quotient evaluation to it, so each iteration now evaluates the constraint polynomial over two packed vectors at once. The prover's trait
  bounds are updated to match; AIRs written generically over the builder (the usual pattern) compile without modification.

## Benchmarks

Running 
```bash
  cargo build --release --example prove_prime_field_31 --features parallel
  cargo build --release --example prove_goldilocks_keccak --features parallel -p p3-keccak-air
  ```
  
 - NEON, M4 Pro:
  
```bash
  ┌──────────────────────┬──────────────────────┬────────────┬───────────────────┬─────────┐
  │        Field         │ quotient main → vect │ Δ quotient │ prove main → vect │ Δ prove │
  ├──────────────────────┼──────────────────────┼────────────┼───────────────────┼─────────┤
  │ Mersenne-31 (circle) │ 390 → 335 ms         │ −14.1%     │ 1.68 → 1.67 s     │ −0.6%   │
  ├──────────────────────┼──────────────────────┼────────────┼───────────────────┼─────────┤
  │ BabyBear             │ 412 → 380 ms         │ −7.8%      │ 1.80 → 1.79 s     │ −0.6%   │
  ├──────────────────────┼──────────────────────┼────────────┼───────────────────┼─────────┤
  │ KoalaBear            │ 412 → 378 ms         │ −8.3%      │ 1.81 → 1.79 s     │ −1.1%   │
  ├──────────────────────┼──────────────────────┼────────────┼───────────────────┼─────────┤
  │ Goldilocks           │ 172 → 159 ms         │ −7.6%      │ 562 → 558 ms      │ −0.7%   │
  └──────────────────────┴──────────────────────┴────────────┴───────────────────┴─────────┘

```

It does not seem to benefit AVX2 / AVX512 so this is gated for `aarch64` targets only for now.

### DIFF
--- a/examples/src/airs.rs
+++ b/examples/src/airs.rs
@@ -8,7 +8,9 @@ use p3_keccak_air::KeccakAir;
 use p3_matrix::dense::RowMajorMatrix;
 use p3_poseidon2::GenericPoseidon2LinearLayers;
 use p3_poseidon2_air::{Poseidon2Air, VectorizedPoseidon2Air};
-use p3_uni_stark::{ProverConstraintFolder, StarkGenericConfig, VerifierConstraintFolder};
+use p3_uni_stark::{
+    QUOTIENT_ILP, StarkGenericConfig, VectorizedConstraintFolder, VerifierConstraintFolder,
+};
 use rand::distr::StandardUniform;
 use rand::prelude::Distribution;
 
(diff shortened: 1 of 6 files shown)


Nashtare:
Fixing `QUOTIENT_ILP` to 1 for x86 targets seems to still yield some non-negligible performance slowdown. The gains are only visible on NEON but not easily extractable for this target alone so I'll be closing this.

PR #1805, opened 2026-06-11
whir: pruned Merkle multiproofs for STIR query openings

tcoratger:
## What

Wire WHIR's STIR query openings through the **pruned Merkle multiproof** machinery (`open_batch_pruned` / `verify_batch_pruned`) that already lives in `p3-merkle-tree` but was never used by `whir/`. Each WHIR round previously opened one full root‑to‑leaf authentication path *per query*; those paths overlap heavily (queries are distinct indices into the same tree), so shared sibling digests were shipped and re‑hashed many times over.

This PR sends each shared sibling **once** per round, and verifies all of a round's queries in a single amortized walk that computes each shared compression once.

## How

- **New trait `MultiOpeningMmcs<T>: Mmcs<T>`** in `p3-commit` — one associated type `MultiProof` plus `open_multi_batch` / `verify_multi_batch`. Implemented for `MerkleTreeMmcs` (`MultiProof = PrunedMerklePaths`), with the `ExtensionMmcs` flatten/reconstitute passthrough and the `&M` blanket impl, so WHIR's base‑field and extension‑field rounds share one surface. It is an opt‑in capability trait — the core `Mmcs` (single index, single path) structurally can't express "N indices → one shared proof", and widening it would tax every implementor (FRI, circle, …) with a capability not all schemes have.
- **Index binding (soundness):** `verify_batch_pruned` now takes the verifier's transcript‑derived `indices` and binds each opening to them (new `PrunedProofError::IndexMismatch` / `IndexCountMismatch`). Previously it trusted the leaf indices embedded in the proof — fine for a roundtrip test, unsound as a STIR opening check.
- **Proof format:** the per‑query, serde‑tagged `QueryOpening` enum becomes `QueryOpenings::{Base,Extension}(SharedProofOpening { rows, proof })` — one multiproof per round, tag dropped. Migrated across the core prover/verifier **and** the full HVZK (zk) pipeline (`ZkRoundProof`, base‑case `source`/`fresh_main`/`mask` openings, `MaskOpeningPair`).
- Added a `report_proof_size` group to `whir/benches/whir_pcs.rs` that prints proof bytes, query count, and pruned‑vs‑full Merkle digests.

## Numbers (before → after)

**Config (the bench default):** KoalaBear + quintic extension, Poseidon2 Merkle (32‑byte digests), λ=128, `CapacityBound`, folding=4, starting rate 1/2, single eval claim. Criterion 10–20 samples on one machine; `before` = `main` (per‑query openings), `after` = this branch (pruned multiproofs). Both include #1806 (which made the λ=128 config runnable).

| size | metric | before (`main`) | after (this PR) | change |
|------|--------|----------------|-----------------|--------|
| **2¹⁴** | proof size | 87,535 B | 55,953 B | **−36 %** |
|         | prover time | 18.04 ms | 18.08 ms | +0.2 % (noise) |
|         | verifier time | 2.146 ms | 1.150 ms | **−46 % (1.9× faster)** |
|         | Merkle digests | 1,621 | 822 | −49 % |
| **2¹⁸** | proof size | 126,595 B | 90,706 B | **−28 %** |
|         | prover time | 82.57 ms | 82.53 ms | ≈0 % (noise) |
|         | verifier time | 3.017 ms | 1.936 ms | **−36 % (1.6× faster)** |
|         | Merkle digests | 2,442 | 1,531 | −37 % |
| **2²⁰** | proof size | 150,098 B | 115,037 B | **−23 %** |
|         | prover time | 395.4 ms | 397.2 ms | +0.5 % (noise) |
|         | verifier time | 3.705 ms | 2.490 ms | **−33 % (1.5× faster)** |
|         | Merkle digests | 2,944 | 2,055 | −30 % |

### Reading the table

- **Proof size: 23–36 % smaller**, growing with query overlap. Cross‑check: the `before` Merkle‑digest counts (1,621 / 2,442 / 2,944) exactly equal the `after` report's `full_digests` baseline, so the savings are real, not a measurement artifact.
- **Verifier: 1.5–1.9× faster** — a bonus beyond the proof‑size goal. Amortized verification computes each shared compression once instead of once per query.
- **Prover: unchanged** (≤0.5 %, inside criterion's noise bands). The added pruning work (sort + LCA dedup over the query set) is negligible against DFT + Merkle + sumcheck. **No regression.**

A strict win on every axis: smaller proofs, faster verification, flat prover.

## Tests

- `p3-merkle-tree`: 103 pass (incl. new index‑binding rejection test), `p3-whir`: 120 pass — default and `parallel` features.
- Clippy clean (`--all-targets --all-features`), `cargo fmt` clean.

## Notes

- Draft: opening for review of the API shape (`MultiOpeningMmcs`) and the proof‑format change before finalizing.
- Builds on #1801 (`mmcs/` module split) and #1806 (128‑bit config fix), both merged in.
- A natural follow‑up is unifying `Mmcs::open_batch` itself to be index‑batched (then there's a single trait and FRI/circle get the same proof‑size win for free) — that's a core‑trait rewrite touching FRI/circle proof formats, so it belongs in its own PR.


commit/src/mmcs.rs
```diff
@@ -203,6 +203,71 @@ impl<T: Send + Sync + Clone, M: Mmcs<T>> Mmcs<T> for &M {
     }
 }
 
+/// An [`Mmcs`] whose openings at many indices share one compact proof.
+///
+/// Opening `q` indices through [`Mmcs::open_batch`] costs `q` full authentication paths.
+/// Those paths overlap wherever two indices share an ancestor.
+/// A multi-opening sends each shared digest once.
+pub trait MultiOpeningMmcs<T: Send + Sync + Clone>: Mmcs<T> {
```
tcoratger, line 211: I think that we can modify the actual Mmcs trait to be able to handle multi opening and so delete this one and unify. But this will be a significant modification among the existing protocols like FRI or circle starks, so I think this is better to:
1. Have this `MultiOpeningMmcs` abstraction for now,
2. Unify things in a PR follow-up.

That will limit the change surface.

tcoratger:
As proposed by @TomWambsgans, the following commit can help to optimize the pruning, but we can put that in a followup PR to avoid making this one too big, since the optimizations compound here:
- https://github.com/leanEthereum/leanVM/commit/d63f10033b5b24bd7a4550dbca75f44b67347539

merkle-tree/src/mmcs/mod.rs
```diff
@@ -539,22 +547,31 @@ impl<P, PW, H, C, const N: usize, const DIGEST_ELEMS: usize>
         // compact representation. We index siblings level by level during the
         // amortized walk, so we materialize the per-path buffers once and
         // borrow them throughout.
-        let restored = restore_paths(&pruned_opening.pruned_proof, full_sibling_count)?;
+        let restored = restore_paths(proof, full_sibling_count)?;
```
Nashtare: We call `restore_paths` before checking `indices.len()` or `opened_values.len()` against `proof.original_order.len()` which I think should be the oposite

whir/src/pcs/proof.rs
```diff
@@ -80,72 +84,121 @@ pub struct PcsProof<F: Send + Sync + Clone, EF, MT: Mmcs<F>> {
...
+        mmcs: &MT,
+        commit: &MT::Commitment,
+        dimensions: &[Dimensions],
+        indices: &[usize],
+    ) -> Result<(), MT::Error>
+    where
+        MT: MultiOpeningMmcs<T, MultiProof = P>,
+    {
+        let opened_values: Vec<Vec<Vec<T>>> =
+            self.rows.iter().map(|row| vec![row.clone()]).collect();
```
Nashtare: We clone every opened row into a `Vec<Vec<Vec<T>>>`, can't we use a borrowed view of the matrix? This is probably painful perf wise

whir/src/pcs/zk/adapter.rs
```diff
@@ -86,7 +86,7 @@ where
     F: TwoAdicField,
     EF: ExtensionField<F> + TwoAdicField,
     Dft: TwoAdicSubgroupDft<F>,
-    MT: Mmcs<F>,
+    MT: MultiOpeningMmcs<F>,
```
Nashtare, line 89: we now require `MT: MultiOpeningMmcs` but I think `MerkleTreeHidingMmcs` is still just a plain Mmcs, is it normal? or just forgotten perhaps?

whir/src/pcs/proof.rs
```diff
@@ -22,21 +24,23 @@ pub struct WhirProof<F: Send + Sync + Clone, EF, MT: Mmcs<F>> {
     pub final_poly: Option<Poly<EF>>,
     /// Final round PoW witness.
     pub final_pow_witness: F,
-    /// Final round query openings.
-    pub final_queries: Vec<QueryOpening<F, EF, MT::Proof>>,
+    /// Final round STIR query openings.
+    pub final_openings: Option<QueryOpenings<F, EF, MT::MultiProof>>,
```
Nashtare: Why is it changed to Option? Isn't this a mandatory field in the proof? Or am I missing something?

whir/src/pcs/proof.rs
```diff
@@ -80,72 +84,121 @@ pub struct PcsProof<F: Send + Sync + Clone, EF, MT: Mmcs<F>> {
...
     /// Round commitment (Merkle root).
     pub commitment: Option<MT::Commitment>,
     /// OOD evaluations for this round.
     pub ood_answers: Vec<EF>,
     /// PoW witness after commitment.
     pub pow_witness: F,
-    /// STIR query openings.
-    pub queries: Vec<QueryOpening<F, EF, MT::Proof>>,
+    /// STIR query openings against the previous round's commitment.
+    pub openings: Option<QueryOpenings<F, EF, MT::MultiProof>>,
```
Nashtare: same here, not sure why `openings` is an option while being a necessarily mandatory field

Nashtare (commented):
Overall LGTM

merkle-tree/src/mmcs/mod.rs
tcoratger: Fixed

whir/src/pcs/proof.rs
tcoratger: Fixed

whir/src/pcs/zk/adapter.rs
tcoratger: Mhhh I think that `MerkleTreeHidingMmcs` is used nowhere in whir/ that's why we have this here. Because the HVZK pipeline gets zero-knowledge from the masking machinery in zk/ and not via salted merkle tree leaves as it is the case for `MerkleTreeHidingMmcs` so I think there is no mistake here :)

whir/src/pcs/proof.rs
tcoratger: You are right, just fixed

whir/src/pcs/proof.rs
tcoratger: Fixed as well

whir/src/pcs/zk/adapter.rs
Nashtare: ah right good point!

tcoratger:
@Nashtare I think I've solved all the comments, let me know if you see other stuffs.

Then I've a followup PR in mind to try to unify the new `MultiOpeningMmcs` and our current `Mmcs` trait. I think the `MultiOpeningMmcs` is kind of useless with a small adaptation of `Mmcs` but I need to check that in a followup in order not to inflate the size of this one.

tcoratger:
@Nashtare As a followup, I've prepared this one https://github.com/tcoratger/Plonky3/pull/35 in order to remove the spurious `MultiOpeningMmcs` trait. But since this touches other parts of the code, I think this is better to do as a followup for easier review.

Then I think that we can apply this to FRI as well as another followup PR.

Nashtare (approved):
LGTM! Thanks

Nashtare:
Shoot @tcoratger, there was no conflicts with main so I went ahead and merged but it broke the build.

PR #1812, opened 2026-06-12
perf(lookup): flag-zero skip for sparse rows + dot_product tuple combine

tcoratger:
## What

Two complementary, independently-sound speedups to `LogUpGadget::generate_permutation` — the per-AIR auxiliary-trace builder in `lookup/src/logup.rs`.

**1. Flag-zero skip for sparse rows.** When a tuple's multiplicity resolves to `0`, the fraction term `m_j / d_j` is `0` for any nonzero denominator. So we resolve the multiplicity first and, when it is zero, skip the element combine entirely and store a unit placeholder denominator. Sparse, selector-gated lookups (memory, opcode-gated, range-gated, …) stop paying for their inactive rows.

**2. `dot_product` tuple combine.** The per-row tuple combine `Σ_i e_i · β^{k-1-i}` used Horner (`acc * β`), where `acc` is an extension element and `β` is the extension challenge — an **extension × extension** product (~16 base muls/step for a quartic). The tuple elements `e_i` are *base* field, so by hoisting the `β` powers out of the row loop the combine becomes a `dot_product` of extension powers with base elements — an **extension × base** product (~4 base muls/step). A width-1 fast path keeps single-element tuples (range checks) free of any extra multiply.

## Why each is sound

**Flag-zero skip.** A zero weight zeroes the fraction term regardless of the denominator, so the placeholder cannot change any committed value:
- The unit placeholder keeps batch inversion well-defined (`1` inverts to `1`); the term is exactly `0 · 1 = 0`.
- The verifier reads the *real* denominator from the trace. For a multi-tuple lookup, `f = Σ_{k: m_k≠0} m_k/d_k(real) = V/U`, since the zero-weight tuples drop from both sides of the fraction-pin constraint `U·f = V`.

**Combine.** Field addition is commutative, so reordering the summation makes the `dot_product` result bit-identical to Horner. Both paths are pinned to an independent reference in the tests below.

## Benchmarks

`cargo bench -p p3-lookup --bench logup` — the `generate_permutation` microbench, which this PR makes sparsity-aware (an `active_period` knob gating both multiplicities by a selector column, so the flag-zero skip is exercised). Numbers are a single back-to-back criterion run (`--save-baseline` then `--baseline`); every change has `p < 0.05`.

| config | workload | before | after | speedup |
|---|---|---|---|---|
| `dense_small`  | h=2^10, tuple width 1, dense       | 69.3 µs  | 64.1 µs  | **−7.4 %** |
| `dense_medium` | h=2^16, tuple width 2, dense       | 10.39 ms | 9.64 ms  | **−7.2 %** |
| `dense_large`  | h=2^20, tuple width 1, dense       | 273.0 ms | 257.7 ms | **−5.6 %** |
| `gated_tw4_p1`  | h=2^18, tuple width 4, dense (1/1 active)  | 67.3 ms | 47.7 ms | **−29.1 %** |
| `gated_tw4_p8`  | h=2^18, tuple width 4, 1/8 active  | 67.7 ms | 30.8 ms | **−54.5 %** |
| `gated_tw4_p64` | h=2^18, tuple width 4, 1/64 active | 67.5 ms | 28.4 ms | **−58.0 %** |

Reading the table:
- The three **dense** rows isolate the combine win — the flag-zero skip never fires when every row is active. Width-1 (`dense_small`/`dense_large`) sees ~−7 %; width-2/4 (`dense_medium`/`gated_tw4_p1`) more, as the extension×base saving scales with tuple width.
- The **gated** rows show the two optimizations stacking: the combine speeds up the active rows, the flag-zero skip drops the inactive ones — roughly halving cost at 1/8 active and better beyond.
- Before this PR the prover did the same work regardless of sparsity (`gated_tw4_p1/p8/p64` all ~67 ms); after, cost tracks the active-row fraction.

## Tests

- `generate_permutation_flag_zero_skip_matches_real_denominators` — a mixed row (one tuple with `m = 0` taking the skip, one with `m ≠ 0` taking the full path) checked against the real-denominator reference, including the accumulator and terminal.
- `generate_permutation_multi_element_combine_matches_reference` — a width-2 tuple checked against the `α − (e₀·β + e₁)` reference, pinning the `dot_product` combine.
- The existing lookup unit tests and the `p3-batch-stark` `simple` integration suite pass unchanged.

## Scope

Only `lookup/` (`logup.rs`, `tests.rs`, `benches/logup.rs`). No public API change, no proof-format change; soundness is untouched (the prover and verifier both recompute the same per-bus rational sum).

### DIFF
--- a/lookup/benches/logup.rs
+++ b/lookup/benches/logup.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use p3_air::symbolic::{AirLayout, SymbolicAirBuilder, SymbolicExpression};
-use p3_air::{AirBuilder, BaseLeaf, WindowAccess};
+use p3_air::{AirBuilder, WindowAccess};
 use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
 use p3_challenger::DuplexChallenger;
 use p3_commit::ExtensionMmcs;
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
LGTM!

PR #1813, opened 2026-06-12
perf(merkle-tree): hash single-matrix leaf rows without `flat_map`

Nashtare:
In first_digest_layer, feeding the hasher through
`tallest_matrices.iter().flat_map(...)` hands it a compound iterator whose next() defeats vectorization of the sponge absorb loop. For the dominant single-matrix case, pass the vertically-packed row iterator directly.

Keccak-f log18 (2633 x 2^19 leaf hashing, Apple Silicon, 14 cores): trace Merkle commit 439-471 ms -> 249-276 ms on both the M31 circle and KB two-adic pipelines, within ~7% of the raw keccak-f permutation floor for the same workload.

When running the `prove_prime_field_31` example with `--objective keccak-f-permutations --log-trace-length 18`, I get a 38% speed-up on the `build merkle tree` / `first digest` phase with `keccak-f` hasher, and 4.5% speed-up with `poseidon2`.

`prove_goldilocks_keccak` example has also its `build merkle tree` phase about 24% faster.

### DIFF
--- a/merkle-tree/src/merkle_tree.rs
+++ b/merkle-tree/src/merkle_tree.rs
@@ -299,11 +299,20 @@ where
 
             // Collect all vertically packed rows from each matrix at `first_row`.
             // These packed rows are then hashed together using `h`.
-            let packed_digest: [PW; DIGEST_ELEMS] = h.hash_iter(
-                tallest_matrices
-                    .iter()
-                    .flat_map(|m| m.vertically_packed_row(first_row)),
-            );
+            //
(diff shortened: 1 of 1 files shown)


tcoratger (approved):
Great since I think this is the case we are using the most in practice!

Nashtare:
I get 2^20 Poseidon2 permutations proven with Koala-Bear in 365ms with my M4 on this branch :)

PR #1815, opened 2026-06-12
perf(uni-stark): reuse per-thread buffers in quotient_values

Nashtare:
Each packed row-group previously allocated three fresh Vecs (the vertically-packed main row pair and the folder's base/ext constraint accumulators); on wide traces this allocation traffic costs ~20% of quotient evaluation. Write quotient values directly into a preallocated output and keep the group buffers in rayon for_each_init state.

Keccak-f log18 (2633 cols, Apple Silicon, 14 cores): quotient_values 397-451 -> 340-347 ms (M31 circle) and 423-492 -> 354-373 ms (KB two-adic).

### DIFF
--- a/uni-stark/src/prover.rs
+++ b/uni-stark/src/prover.rs
@@ -537,6 +537,9 @@ where
                 // quotient(x) = constraints(x) / Z_H(x)
                 let quotient = folder.finalize_constraints() * inv_vanishing;
 
+                // The contents were folded into `quotient` above; reclaim the Vecs and
+                // `clear` them (`len = 0`, capacity kept) so the next row group reuses
+                // the allocations.
                 bufs.base_constraints = folder.base_constraints;
                 bufs.base_constraints.clear();
                 bufs.ext_constraints = folder.ext_constraints;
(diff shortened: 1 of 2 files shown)


uni-stark/src/prover.rs
```diff
@@ -454,62 +454,99 @@ where
...
+                };
+                air.eval(&mut folder);
+
+                // quotient(x) = constraints(x) / Z_H(x)
+                let quotient = folder.finalize_constraints() * inv_vanishing;
+
+                bufs.base_constraints = folder.base_constraints;
+                bufs.base_constraints.clear();
+                bufs.ext_constraints = folder.ext_constraints;
+                bufs.ext_constraints.clear();
```
tcoratger, line 546: This looks strange no? because for the two we clear just after filling no?

tcoratger (approved):
Just one minor comment!

uni-stark/src/prover.rs
Nashtare: Yeah I agree it'd be worth a comment, but this is a tweak we had done on our zkEVM side as well, so basically this is a trick to preserve the allocated length while clearing up the `Vec`, so that on the next iteration we don't need to reallocate. If you want it's the "safe" version of the unsafe `set_len(0)`. The move is basically free here. I'll add a comment

PR #1819, opened 2026-06-12
perf(circle): flatten the fused CFFT pass driver

Nashtare:
Two driver-shape costs sat above the butterfly math itself:

- DifButterfly used the default apply_to_rows, re-broadcasting the scalar twiddle per packed multiplication and exposing a single dependency chain. Give it the same pre-broadcast + unroll-4 override DitButterfly already has.
- Stride-0 passes called apply_to_rows once per row pair even though the 2^e pairs sharing a twiddle span two contiguous row blocks. Merge them into one call per twiddle: one broadcast, one pack-split, and a long inner loop over the whole block.

Shaves off another ~25 ms from `extrapolate` phase

### DIFF
--- a/dft/src/butterflies.rs
+++ b/dft/src/butterflies.rs
@@ -122,6 +122,51 @@ impl<F: Field> Butterfly<F> for DifButterfly<F> {
     fn apply<PF: PackedField<Scalar = F>>(&self, x_1: PF, x_2: PF) -> (PF, PF) {
         (x_1 + x_2, (x_1 - x_2) * self.0)
     }
+
+    /// Override `apply_to_rows` to pre-broadcast the twiddle factor into a packed field
+    /// once before the inner loop, and manually unroll it to expose multiple independent
+    /// sub-then-mul chains to the compiler's scheduler, hiding the multiplication latency.
+    /// Mirrors the [`DitButterfly`] override.
+    #[inline]
+    fn apply_to_rows(&self, row_1: &mut [F], row_2: &mut [F]) {
+        let (shorts_1, suffix_1) = F::Packing::pack_slice_with_suffix_mut(row_1);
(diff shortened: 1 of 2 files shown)


dft/src/butterflies.rs
```diff
@@ -122,6 +122,51 @@ impl<F: Field> Butterfly<F> for DifButterfly<F> {
...
+        debug_assert_eq!(shorts_1.len(), shorts_2.len());
+        debug_assert_eq!(suffix_1.len(), suffix_2.len());
+        let twiddle_packed = F::Packing::from(self.0);
+        let mut c1 = shorts_1.chunks_exact_mut(4);
+        let mut c2 = shorts_2.chunks_exact_mut(4);
+        for (p1, p2) in (&mut c1).zip(&mut c2) {
+            let a1 = p1[0];
+            let b1 = p1[1];
+            let c1_ = p1[2];
+            let d1 = p1[3];
```
tcoratger, line 143: Maybe that is slightly better?

```suggestion
            let (a1, b1, c1_, d1) = p1;
```

dft/src/butterflies.rs
```diff
@@ -122,6 +122,51 @@ impl<F: Field> Butterfly<F> for DifButterfly<F> {
...
+            let a2 = p2[0];
+            let b2 = p2[1];
+            let c2_ = p2[2];
+            let d2 = p2[3];
```
tcoratger, line 147: Same?

```suggestion
            let (a2, b2, c2_, d2) = p2;
```

tcoratger (approved):
LGTM!

dft/src/butterflies.rs
Nashtare: This can't compile

PR #1822, opened 2026-06-12
perf(m31): widening NEON MACs for the mixed base×extension dot-product kernels

Nashtare:
## Description

The two big mixed base×QM31 passes of the Circle open phase (the fused two-point
  out-of-domain evaluation and the rowwise alpha-reduction) execute 4 (resp. 8)
  independent `PackedM31` multiply-accumulates per packed word, each paying the full
  5-instruction reduced multiply plus 3-instruction modular add. On NEON, the raw
  32×32→64-bit products can instead be accumulated with widening MACs
  (`vmlal_u32`/`vmlal_high_u32`, 2 instructions per product) and the Mersenne
  reduction deferred: inputs lie in `0..=P`, so three unreduced products plus a
  folded carry fit a u64 lane (`3·P² + 2³³ < 2⁶⁴`), and one fold
  (`(x & P) + (x >> 31)`, using `2³¹ ≡ 1 (mod P)`) restores headroom.

  This PR adds two overridable hooks for these inner kernels — generic defaults are
  identical to the previous inline loops, so all other fields and targets are
  unaffected — and specializes both for `Mersenne31` on NEON:

  - `PackedField::coeffwise_dot_product` (rowwise shape, accumulators in registers):
    straight substitution, folding every 3 iterations.
  - `Field::batched_columnwise_dot_product` (columnwise shape, accumulators in
    memory): `columnwise_dot_product_batched` now feeds bounded row chunks to the
    field's kernel, which consumes rows three at a time — each product triple is
    accumulated in registers, folded once, then added into u64 lane accumulators,
    which absorb 2³¹ triples without overflow. The kernel is monomorphized over the
    extension degree (1/2/4/8, generic fallback otherwise) so the per-word loops
    fully unroll; a naive runtime-degree version recovers almost none of the win.

  Both kernels carry equivalence tests against the generic accumulation, covering
  the boundary values (0, 1, P−1 and the redundant representation P of zero), every
  deferred-reduction phase of the stream length, and the trivial as well as a
  degree-4 extension.
  
  ## Benchmarks

On my machine, M4 pro NEON activated, this shaves off 4% of the full E2E prover when running `cargo run --release --features parallel --example prove_prime_field_31 -q -- --field mersenne31 --objective poseidon2-permutations --log-trace-length 17 --merkle-hash keccak-f`.

  ## Test plan

  - [x] `cargo test -p p3-field -p p3-matrix -p p3-mersenne-31` (incl. new
        equivalence tests for both kernels)
  - [x] `cargo test -p p3-circle -p p3-fri -p p3-uni-stark -p p3-baby-bear
        -p p3-koala-bear` (end-to-end M31 circle proofs exercise both NEON kernels)
  - [x] `cargo clippy --all-targets` clean on the touched crates
  - [x] A/B benchmarks above

### DIFF
--- a/matrix/src/lib.rs
+++ b/matrix/src/lib.rs
@@ -480,8 +480,8 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
                 || EF::ExtensionPacking::zero_vec(packed_width * N),
                 |mut acc, chunk| {
                     let rows = chunk * chunk_rows..((chunk + 1) * chunk_rows).min(height);
-                    let partial = T::batched_columnwise_dot_product::<EF, _, _, N>(
-                        packed_width,
+                    T::batched_columnwise_dot_product::<EF, _, _, N>(
+                        &mut acc,
                         rows.map(|r| {
                             (
                                 self.padded_horizontally_packed_row::<T::Packing>(r),
@@ -489,7 +489,6 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
                             )
                         }),
                     );
-                    acc.iter_mut().zip(partial).for_each(|(a, p)| *a += p);
                     acc
                 },
                 |mut acc_l, acc_r| {
--- a/matrix/src/lib.rs
+++ b/matrix/src/lib.rs
@@ -9,8 +9,8 @@ use core::ops::Deref;
 
 use itertools::Itertools;
 use p3_field::{
-    BasedVectorSpace, ExtensionField, Field, FieldArray, PackedFieldExtension, PackedValue,
-    PrimeCharacteristicRing,
+    BasedVectorSpace, ExtensionField, Field, FieldArray, PackedField, PackedFieldExtension,
+    PackedValue, PrimeCharacteristicRing,
 };
 use p3_maybe_rayon::prelude::*;
 use strided::{VerticallyStridedMatrixView, VerticallyStridedRowIndexMap};
@@ -530,18 +530,13 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
                 // Get the extension dimension from the first vec element's coefficients
                 let d = <EF::ExtensionPacking as BasedVectorSpace<T::Packing>>::DIMENSION;
 
-                // Initialize D accumulators for each coefficient of the extension
-                // In practice, we set D to 8, which is the maximum degree of the extension field supported.
-                let mut coeff_accs: [T::Packing; 8] = [T::Packing::ZERO; 8];
-                debug_assert!(d <= 8, "Extension degree > 8 not supported");
-
                 // Accumulate coefficient-wise: for each (v, r) pair, acc[i] += v.coefficient(i) * r
-                for (v, r) in vec.iter().zip(row_packed) {
-                    let v_coeffs = v.as_basis_coefficients_slice();
-                    for (acc, &v_coeff) in coeff_accs[..d].iter_mut().zip(v_coeffs) {
-                        *acc += v_coeff * r;
-                    }
-                }
+                let coeff_accs = T::Packing::coeffwise_dot_product(
+                    d,
+                    vec.iter()
+                        .zip(row_packed)
+                        .map(|(v, r)| (v.as_basis_coefficients_slice(), r)),
+                );
 
                 // Construct the result ExtPacking from the accumulators and sum the coefficients.
                 let packed_result =
--- a/matrix/src/lib.rs
+++ b/matrix/src/lib.rs
@@ -467,23 +467,29 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
         EF: ExtensionField<T>,
     {
         let packed_width = self.width().div_ceil(T::Packing::WIDTH);
+        let height = self.height().min(vs.len());
 
-        let packed_results: Vec<EF::ExtensionPacking> = self
-            .par_padded_horizontally_packed_rows::<T::Packing>()
-            .zip(vs)
-            .par_fold_reduce(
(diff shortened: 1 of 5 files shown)


mersenne-31/src/aarch64_neon/packing.rs
```diff
@@ -328,6 +344,221 @@ impl_packed_value!(PackedMersenne31Neon, Mersenne31, WIDTH);
...
+pub(crate) fn batched_columnwise_dot_product<EF, R, I, const N: usize>(
+    packed_width: usize,
+    items: I,
+) -> Vec<EF::ExtensionPacking>
+where
+    EF: ExtensionField<Mersenne31>,
+    R: Iterator<Item = PackedMersenne31Neon>,
+    I: Iterator<Item = (R, [EF; N])>,
+{
+    match <EF as BasedVectorSpace<Mersenne31>>::DIMENSION {
```
tcoratger: This doesn't work?

```suggestion
    match EF::DIMENSION {
```

mersenne-31/src/aarch64_neon/packing.rs
```diff
@@ -328,6 +344,221 @@ impl_packed_value!(PackedMersenne31Neon, Mersenne31, WIDTH);
...
+fn columnwise_kernel<EF, R, I, const N: usize, const D: usize>(
+    packed_width: usize,
+    mut items: I,
+) -> Vec<EF::ExtensionPacking>
+where
+    EF: ExtensionField<Mersenne31>,
+    R: Iterator<Item = PackedMersenne31Neon>,
+    I: Iterator<Item = (R, [EF; N])>,
+{
+    debug_assert_eq!(<EF as BasedVectorSpace<Mersenne31>>::DIMENSION, D);
```
tcoratger: Same doesn't this work? And I guess that `EF::DIMENSION` doesn't allow a  `const { assert_eq!(EF::DIMENSION, D); }` thing right?

```suggestion
    debug_assert_eq!(EF::DIMENSION, D);
```

matrix/src/lib.rs
```diff
@@ -467,23 +467,29 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
...
+                    let partial = T::batched_columnwise_dot_product::<EF, _, _, N>(
+                        packed_width,
+                        rows.map(|r| {
+                            (
+                                self.padded_horizontally_packed_row::<T::Packing>(r),
+                                vs[r].0,
+                            )
+                        }),
+                    );
+                    acc.iter_mut().zip(partial).for_each(|(a, p)| *a += p);
```
tcoratger: tiny thought, take it or leave it — every chunk allocates a fresh `partial` here and then we merge it into `acc`, so for the generic fields it's an extra alloc + this merge pass per chunk on top of the fold buffer. could sidestep both by having the kernel accumulate straight into a `&mut [_]` instead of returning a `Vec`: generic path writes directly into `acc` (no alloc, no merge), and the m31 neon path keeps its u64 buffer internal but adds into the passed-in acc at the very end. i hacked it together locally and it builds + tests pass, happy to drop the patch here if it sounds worthwhile. equally fine to leave as is!

mersenne-31/src/aarch64_neon/packing.rs
```diff
@@ -389,4 +622,139 @@ mod tests {
...
+            check::<Mersenne31, 2>(packed_width, &rows, &scales_m31);
+
+            let scales_qm31: Vec<[QM31; 2]> = (0..height)
+                .map(|r| {
+                    core::array::from_fn(|j| {
+                        QM31::from_basis_coefficients_fn(|k| scalar((r * 8 + j * 4 + k) as u32))
+                    })
+                })
+                .collect();
+            check::<QM31, 2>(packed_width, &rows, &scales_qm31);
```
tcoratger, line 766: super minor / cheap-insurance kind of thing: the dispatch has a `2 =>` arm that `Complex<Mersenne31>` (CM31) actually hits in prod, but the test only drives D=1 and D=4, so that degree-2 monomorphization never runs. dropping a `check::<Complex<Mersenne31>, 2>(...)` next to these covers it (tried it, passes). the parametric body makes a real bug unlikely so no worries if you'd rather skip. (couldn't find a real type for the `8 =>` arm fwiw, so that one's just defensive.)

mersenne-31/src/aarch64_neon/packing.rs
```diff
@@ -328,6 +344,221 @@ impl_packed_value!(PackedMersenne31Neon, Mersenne31, WIDTH);
...
+        Self: 'a,
+        I: Iterator<Item = (&'a [Self], Self)>,
+    {
+        // Accumulate the raw 32x32 -> 64-bit products with widening multiply-accumulates
+        // (2 instructions per product), deferring the Mersenne reduction, instead of
+        // paying the 8-instruction reduced multiply-add per product.
+        //
+        // Each coefficient accumulator is a pair of u64x2 vectors: `lo` holds lanes 0-1
+        // and `hi` lanes 2-3. Inputs are in `0..=P`, so a product is at most
+        // `P^2 < 2^62`; folding the accumulators below `2^33` every 3 iterations keeps
```
tcoratger, line 360: nit-of-a-nit, ignore freely: one `partial_reduce_u64` actually lands below `2^34` (it's `(val & P) + (val>>31)` so up to `~2^33 + 2^31`), not `2^33` — same spots at lines 433/452. doesn't change the `< 2^64` headroom at all, just the stated number if you ever want it exact 🙂

tcoratger (commented):
Just an additional review that I discussed with Claude, nothing blocking for me.

tcoratger (approved):

PR #1825, opened 2026-06-12
feat(lookup): exclusivity-aware lookup columns

tcoratger:
## What this does

Adds a second way to fold lookups into auxiliary columns, for the case where a row runs **exactly one of N** interactions (one opcode per row, one branch taken, …).

Today every lookup gets its own auxiliary column. When the interactions are **mutually exclusive**, that is wasteful: on any row all but one column just hold zero. This PR multiplexes such a group into a **single** column whose denominator is *flag-selected*:

```text
  N = Σ_k flag_k · mult_k
  D = Σ_k flag_k · (α − combine_k) + (1 − Σ_k flag_k)
```

Exactly one flag is `1`, so `D` selects the active branch's denominator (and falls back to `1` on an idle row, giving fraction `0`). The key property: the column's constraint degree is the **per-branch maximum, not the sum**, so it stays flat (degree 3) no matter how many branches fold in. `N` columns collapse to `1`.

### This complements — does not replace — same-bus packing (#1799)

The two folding strategies cover different workloads:

| strategy | combines | validity | degree |
|---|---|---|---|
| additive packing (#1799) | co-firing interactions, **summed** | always valid | grows with the fold (`g+1`) |
| exclusive multiplex (this PR) | one-of-N, **selected** | needs mutual exclusivity | flat (3) |

Several lookups on one row must be *summed* (only #1799 works there). One-of-many on a row can be *selected* (this PR, folding arbitrarily many at flat degree). Both stay in the codebase; the system picks based on whether the lookups are declared exclusive.

## API

- `Lookup.flags: Option<Vec<_>>` — `None` = additive (unchanged); `Some` = exclusive multiplex.
- `InteractionBuilder::push_exclusive_interaction` — default no-op, so only the recording builder overrides it; all other builders are untouched.
- `LookupBus::lookup_key_exclusive` — convenience for a group of mutually-exclusive queries.

**Soundness obligation** (documented like `count_weight`): the AIR must constrain the flags to be boolean and sum to at most one. Otherwise the multiplexed fraction matches no single message. The bench AIR adds these guards (one boolean per flag + one "at most one" constraint).

The prover reuses the existing flag-zero skip, so an exclusive column computes **one** denominator per row, not `N`. The same-bus packer passes exclusive columns through untouched.

## Benchmark: before → after

End-to-end through the full `prove_batch` / `verify_batch` pipeline, one AIR at height `2^16`, production-like FRI (100 queries). Both layouts compute the **same** bus contribution and **both verify** — they differ only in how many columns carry it.

`before` = additive baseline (one column per branch). `after` = exclusive multiplex.

| branches | aux columns | prove time | proof size | peak heap |
|---|---|---|---|---|
| 8  | 17 → **3** | 784 ms → **556 ms** (1.41×) | 791 → **765 KB** | 89 → **63 MB** |
| 16 | 33 → **3** | 1279 ms → **678 ms** (1.89×) | 833 → **772 KB** | 130 → **71 MB** |
| 32 | 65 → **3** | 2284 ms → **961 ms** (2.38×) | 915 → **790 KB** | 235 → **88 MB** |

The win grows with branch count. Verify is roughly flat (~26–34 ms; the degree-3 quotient adds a hair at 32 branches, swamped by the prove/peak/proof gains).

Prover trace-generation alone (`lookup/benches/logup.rs`, `generate_permutation`) shows 3.2–5.0× across the same branch counts.

### Honest tradeoff

Exclusivity trades a degree-2 layout for degree-3 (the `at-most-one` constraint + the flag-selected pin), in exchange for the `N → 1` column collapse. The net is a decisive prove/peak/proof win that grows with the number of branches; verify stays neutral.

## Tests

- degree is flat in branch count (3 for 2 branches and for 8; contrast additive `g+1`)
- multiplexed `(N, D)` matches the definition, including the idle-row fallback
- prover selects the active branch's fraction, and `eval_all` accepts the generated trace
- cross-AIR multiset balance of an exclusive query against a provider
- the bus API flows through `from_air` into an exclusive column

47 lookup unit tests pass; all benchmark cases prove **and** verify; `clippy` and `fmt` clean.

## Benchmark harness

`batch-stark/benches/lookup.rs` is unified behind one `BenchAir` / `Workload` model, so the same-bus packing story and the exclusivity story read as **variant pairs in one metrics table and one criterion group** (no duplicated measure/report/bench code).


lookup/src/logup.rs
```diff
@@ -332,6 +436,37 @@ impl LookupProtocol for LogUpGadget {
...
+        // So combine_k has the degree of its widest element.
+        // Each branch then adds its flag's degree on top.
+        if let Some(flags) = &lookup.flags {
+            let deg_denom = flags
+                .iter()
+                .zip(&lookup.elements)
+                .map(|(flag, elems)| {
+                    let elem_deg = elems.iter().map(|e| e.degree_multiple()).max().unwrap_or(0);
+                    flag.degree_multiple() + elem_deg
+                })
```
Nashtare, line 463: need to use `zip_eq`

lookup/src/logup.rs
```diff
@@ -148,6 +175,60 @@ impl LogUpGadget {
...
+        AB: PermutationAirBuilder,
+    {
+        // Per-branch denominator term (alpha - combine(payload_k)).
+        let terms = self.combine_elements::<AB, AB::ExprEF>(elements, alpha, beta);
+
+        // Multiplex by the flags, accumulating the selected numerator and denominator.
+        let mut numerator = AB::ExprEF::ZERO;
+        let mut denominator = AB::ExprEF::ZERO;
+        let mut flag_sum = AB::ExprEF::ZERO;
+        for ((flag, term), mult) in flags.iter().zip(&terms).zip(multiplicities) {
```
Nashtare: same here, need to use `zip_eq`

lookup/src/types.rs
```diff
@@ -56,6 +56,35 @@ pub struct Lookup<F: Field> {
...
+    /// - With flags, at most one fires per row and selects one branch.
+    /// - With flags, the denominator degree is the per-branch maximum, not the sum.
+    /// - With flags, there is exactly one flag per tuple.
+    ///
+    /// # Soundness
+    ///
+    /// - The flags must be boolean.
+    /// - The flags must sum to at most one on every row.
+    /// - The AIR must enforce both rules.
+    /// - This field only carries the flags.
```
Nashtare, line 75: I think we need to add `debug_assertions` at the very least to ensure the contract is respected, because I don't think there are safeguards at the moment and this soundness critical.

lookup/src/logup.rs
```diff
@@ -36,6 +36,33 @@ use crate::types::{Lookup, LookupError, LookupTerminal};
 /// Type alias for the row evaluation context used during permutation trace generation.
 pub type RowEvalContext<'a, SC> = LookupTraceBuilder<'a, SC>;
 
+/// Combine one tuple into a single extension-field value `sum_i e_i * beta^{k-1-i}`.
+///
+/// - The last element carries `beta^0`.
```
Nashtare: says last element but doc inline says `beta^{k-1} down to beta^1` when going from `1..elts.len()`

lookup/src/logup.rs
```diff
@@ -36,6 +36,33 @@ use crate::types::{Lookup, LookupError, LookupTerminal};
...
+/// - It is lifted into the extension field with no multiply.
+/// - Each leading element pairs with a hoisted power `beta^p * e_i`.
+/// - That product is extension times base.
+/// - It is cheaper than the extension times extension a Horner fold uses.
+fn combine_tuple<SC: StarkGenericConfig>(
```
Nashtare, line 52: I wonder if it can't be bundled somehow with `combine_elements` to reduce drift / duplication.

Nashtare (changes requested):
Couple soundness issues but overall looking good. I'll probably do a second deeper pass.

lookup/src/logup.rs
tcoratger: Fixed

lookup/src/logup.rs
tcoratger: Fixed

lookup/src/types.rs
tcoratger: Done

lookup/src/logup.rs
tcoratger: Fixed the documentation here

lookup/src/logup.rs
tcoratger: I think that maybe we can check in a followup PR but since right now they are on different layers, this may complexify things

tcoratger:
Hey @Nashtare,

Please let me know if there is anything missing before merging this one? :)

Nashtare (approved):
No this is great, thanks again!

PR #1835, opened 2026-06-14
bench(sumcheck): add generic Criterion benchmark suite

tcoratger:
## What

The `sumcheck` crate had no benchmarks. This adds a complete, field- and strategy-generic Criterion suite, so the engine can be measured carefully now and after future optimizations.

## Coverage

The benches walk the protocol from the innermost kernel outward:

| Group | Measures |
|---|---|
| `round_coefficients` | the `O(2^n)` `(h(0), h(inf))` kernel in its three real operand flavours: round-0 `base × ext` packed, later-round `ext × ext` packed, final-round `ext × ext` scalar |
| `fix_var` | the in-place one-variable binding pass |
| `dot_product` | the running-sum consistency check |
| `product_round` | one complete round (coeffs + transcript + challenge + fold + transition) |
| `prover` | the multi-round driver folding every variable to a constant, **with and without** a constraint absorbed first |
| `layout` | the end-to-end `into_sumcheck` handoff — packed first round (prefix), SVO preprocessing (suffix) |

## Generic over the knobs that matter

- **Field** — a small `BenchField` trait bundles `(F, EF, Challenger)`; instantiated for **BabyBear** and **KoalaBear**. Adding a field is one `impl`.
- **Binding order** — every prover-shaped bench runs both `Prefix` and `Suffix`.
- **Sizes / folding** — all live in `const` grids at the top of the file (kernel grid straddles the `2^14` serial/parallel crossover up to `2^22`).

Benchmark ids are `sumcheck/<field>/<group>/<variant>/k<size>`, so any slice is reachable by regex, e.g. `cargo bench -p p3-sumcheck --bench sumcheck -- 'round_coefficients/base_ext'`.

## Notes

- Field elements are sampled through the fields' own `rand` machinery, with the `StandardUniform: Distribution` bound spelled on each sampling function (matching the existing repo idiom).
- Setup (witness build, opening draw, clones) is kept in untimed `iter_batched` setup; heavy full-fold groups use `sample_size(10)`.
- `whir/benches/sumcheck.rs` already benchmarks this layout prover plus the ZK arm; this in-crate suite is broader and dependency-light, and is left independent.

`cargo fmt` and `cargo clippy --bench sumcheck` are clean; all groups were smoke-run across both fields and both orders without panics.

### DIFF
--- a/sumcheck/Cargo.toml
+++ b/sumcheck/Cargo.toml
@@ -37,7 +37,12 @@ p3-koala-bear = { path = "../koala-bear" }
 p3-merkle-tree = { path = "../merkle-tree" }
 p3-symmetric = { path = "../symmetric" }
 
+criterion.workspace = true
 proptest.workspace = true
+
+[[bench]]
+name = "sumcheck"
+harness = false
 
 [lints]
 workspace = true
--- a/sumcheck/benches/sumcheck.rs
+++ b/sumcheck/benches/sumcheck.rs
@@ -1,0 +1,618 @@
+//! Criterion benches for the quadratic sumcheck engine.
+//!
+//! The benches walk the protocol from the innermost kernel outward:
+//!
+//! - The per-round coefficient kernel that produces `(h(0), h(inf))`.
+//! - The in-place pass that binds one variable per round.
+//! - The running-sum consistency dot product.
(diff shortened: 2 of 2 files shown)


Nashtare (approved):
Nice!

PR #1843, opened 2026-06-14
perf(fri): batch (z - x) inversions in the verifier

Nashtare:
The verifier inverted each (z - x) denominator one at a time with try_inverse() inside the per-query x per-matrix x per-point loop, costing Theta(num_queries . total_opening_points) standalone field inversions per proof. The denominators depend only on the query point and the opening points, so collect them across each query and invert with a single batch_multiplicative_inverse (one inversion plus ~3N muls).

Pre-scan for a zero denominator before the batched inversion to keep the OpeningPointMatchesQueryPoint rejection exact, since batch_multiplicative_inverse panics on a zero input.

### DIFF
--- a/fri/src/verifier.rs
+++ b/fri/src/verifier.rs
@@ -5,7 +5,7 @@ use itertools::Itertools;
 use p3_challenger::{CanObserve, FieldChallenger, GrindingChallenger};
 use p3_commit::{BatchOpening, BatchOpeningRef, Mmcs};
 use p3_field::coset::TwoAdicMultiplicativeCoset;
-use p3_field::{ExtensionField, Field, TwoAdicField};
+use p3_field::{ExtensionField, Field, TwoAdicField, batch_multiplicative_inverse};
 use p3_matrix::Dimensions;
 use p3_util::{log2_strict_usize, reverse_bits_len};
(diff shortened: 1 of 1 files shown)


fri/src/verifier.rs
```diff
@@ -733,16 +758,12 @@ where
...
-                            batch,
-                            matrix,
-                            point,
-                        })?;
+                // Inverse of (z - x), precomputed in the batched inversion above. The pre-pass
+                // pushed exactly one denominator per (batch, matrix, point) in this iteration
+                // order, so the next entry corresponds to this opening point.
+                let quotient = inverse_denominators
+                    .next()
+                    .expect("one inverse was precomputed per opening point");
```
tcoratger, line 766: can't we have an error for this instead of panicking?

tcoratger (approved):

fri/src/verifier.rs
Nashtare: I'm not sure I understand the comment, there's an expect here because we ensured before that this couldn't panic, hence adding an unreachable error path is (IMO) less clear than an `expect` message.

tcoratger: Ok agreed!

PR #1848, opened 2026-06-14
feat: add sqrt method for `Field`

### LINKED ISSUE #719: Add Square Root Functionality
The finite field library currently lacks a built-in method to compute square roots over field elements. Efficient square root computation is crucial for applications in elliptic curves, cryptography, and number theory (e.g., ECDSA, SNARKs and pairings).

Defining a generic API for square root computation has always been a challenge, as the best one can hope for is a generic method for computing roots. The situation becomes even more complex when considering extension fields.

That said, one could always use the Tonelli-Shanks algorithm for prime fields and employ specialized methods for extension fields that satisfy certain criteria. For instance, Atkin’s algorithm can be used when $q \equiv 5 \pmod{8}$, while leaving the function undefined for cases where no efficient method is known.

Nashtare:
### Summary

  Adds a `Field::try_sqrt` method that returns a square root of a field element when one exists (`None` for quadratic non-residues, `Some(ZERO)` for zero; one of the two roots is returned, unspecified which). The default is
  a generic Tonelli–Shanks implementation that works for every field — prime fields and extension fields alike — and individual fields override it with faster routines where their structure allows.

closes #719 

  ### Algorithm overview

  **Generic default — Tonelli–Shanks** (`field/src/sqrt.rs`)

  Writing the multiplicative group order as `|F| − 1 = q · 2^s` with `q` odd:

  - A single exponentiation `u = a^((q−1)/2)` yields both candidates: `r = u·a = a^((q+1)/2)` (the prospective root) and `t = r·u = a^q` (the 2-power-torsion component).
  - The 2-Sylow subgroup is seeded with `c = GENERATOR^q` (`GENERATOR` generates `F*`, so it is a quadratic non-residue).
  - The standard fix-up loop walks the 2-power torsion; a non-residue is detected when `t` has full order `2^s`, returning `None`.

  This applies unchanged to all fields (BabyBear, KoalaBear, Goldilocks, BN254, and all extension fields).

  ### Per-field overrides

  | Field | Override | Algorithm |
  |---|---|---|
  | **Mersenne31** | `try_sqrt` (trait) | `p = 2³¹−1 ≡ 3 (mod 4)`, so `a^((p+1)/4) = a^(2²⁹)` is a root whenever one exists. Computed as `a.exp_power_of_2(29)`, then verified with `r² == a` to reject non-residues. |
  | **Goldilocks** | `try_sqrt` (trait) | Two-adic-seeded Tonelli–Shanks: seeds the 2-Sylow subgroup from the precomputed `two_adic_generator(TWO_ADICITY)` instead of recomputing `GENERATOR^q`, saving one full `log₂ p`-bit exponentiation. |
  | **BabyBear / KoalaBear** | `try_sqrt` (inherent) | Same two-adic-seeded Tonelli–Shanks (`tonelli_shanks_two_adic`). Implemented as an **inherent** method on `MontyField31<FP: FieldParameters + TwoAdicData>` |

  ### Note on the BabyBear/KoalaBear override

  `MontyField31`'s `Field` impl is `impl<FP: FieldParameters>`, but the two-adic generator constant lives behind the extra `TwoAdicData` bound. Adding that bound to the `Field` impl cascades into 29 compile errors (it would
  force `TwoAdicData` across the whole monty public API — `PrimeField`, `inverse`, packing, …), so it isn't viable.

  Instead, the override is an **inherent** `try_sqrt` on the `TwoAdicData`-bounded inherent impl. It shadows the trait method for concrete `BabyBear`/`KoalaBear` values (the common case); generic `F: Field` callers fall back
  to the trait default. Both paths return identical results — only speed differs. Dedicated `test_*_inherent_sqrt` tests cover the inherent path, since the generic `test_sqrt` harness exercises the trait path.

  ### Performance

  `cargo bench --bench bench_field -- sqrt` (random quadratic residue):

  | Field | sqrt runtime |
  |---|---|
  | Mersenne31 | 27 ns |
  | KoalaBear | 300 ns |
  | BabyBear | 360 ns |
  | Goldilocks | 640 ns |

  For reference, the generic Tonelli–Shanks default (before per-field overrides) measured ~210 ns for Mersenne31 and ~715 ns for Goldilocks — the overrides give ~8× and ~10% respectively, and the two-adic seed saves ~25% on
  BabyBear / ~16% on KoalaBear vs. the generic `GENERATOR^q` seed.

  ### API additions

  - `Field::try_sqrt(&self) -> Option<Self>` — default method, overridable.
  - `p3_field::tonelli_shanks` and `p3_field::tonelli_shanks_two_adic` — public free functions.
  - `MontyField31::try_sqrt` — inherent method for two-adic monty fields.

  ### Testing

  - New generic `test_sqrt` wired into the shared `test_field!` macro, so every field (and extension field) is covered: checks `sqrt(x²)² == x²`, validates returned roots, and confirms both residues and non-residues are
  sampled.
  - Concrete `test_baby_bear_inherent_sqrt` / `test_koala_bear_inherent_sqrt` pin the inherent two-adic path.
  - `benchmark_sqrt` added to `bench_func` and wired into all five `bench_field` benches.

  ### Test plan

  - [x] `cargo test` across `p3-field`, `p3-baby-bear`, `p3-koala-bear`, `p3-goldilocks`, `p3-mersenne-31`, `p3-bn254`
  - [x] `cargo clippy` clean on all touched crates
  - [x] `cargo fmt --all --check`

### DIFF
--- a/baby-bear/src/baby_bear.rs
+++ b/baby-bear/src/baby_bear.rs
@@ -184,6 +184,28 @@ mod tests {
         test_field_json_serialization(&[f, f_1, f_2, f_p_minus_1, f_p_minus_2, m1, m2]);
     }
 
+    #[test]
+    fn test_baby_bear_inherent_sqrt() {
+        // Exercises the inherent (two-adic) `try_sqrt` which shadows
+        // `Field::try_sqrt` for the concrete `BabyBear` type.
+        let mut residues = 0;
+        let mut non_residues = 0;
+        for i in 0..2000u32 {
+            let x = F::from_u32(i);
+            // The square of any element is a quadratic residue.
+            let square = x.square();
+            assert_eq!(square.try_sqrt().map(|r| r.square()), Some(square));
+            match x.try_sqrt() {
+                Some(r) => {
+                    assert_eq!(r.square(), x);
+                    residues += 1;
+                }
+                None => non_residues += 1,
+            }
+        }
+        assert!(residues > 0 && non_residues > 0);
+    }
+
     // MontyField31's have no redundant representations.
     const ZEROS: [BabyBear; 1] = [BabyBear::ZERO];
     const ONES: [BabyBear; 1] = [BabyBear::ONE];
--- a/baby-bear/src/baby_bear.rs
+++ b/baby-bear/src/baby_bear.rs
@@ -137,6 +137,7 @@ mod tests {
         test_field_json_serialization, test_prime_field, test_prime_field_32, test_prime_field_64,
         test_two_adic_field,
     };
+    use proptest::prelude::*;
 
     use super::*;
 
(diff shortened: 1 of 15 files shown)


baby-bear/src/baby_bear.rs
```diff
@@ -184,6 +184,28 @@ mod tests {
...
+            match x.try_sqrt() {
+                Some(r) => {
+                    assert_eq!(r.square(), x);
+                    residues += 1;
+                }
+                None => non_residues += 1,
+            }
+        }
+        assert!(residues > 0 && non_residues > 0);
+    }
```
tcoratger: Maybe a proptest is more adapted here?

field/src/sqrt.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Unlike [`Field::exp_u64`] this accepts exponents which do not fit in a `u64`,
+/// as required by fields whose order exceeds `2^64`.
+fn exp_biguint<F: Field>(base: F, exponent: &BigUint) -> F {
+    let mut product = F::ONE;
+    let mut current = base;
+    for j in 0..exponent.bits() {
+        if exponent.bit(j) {
+            product *= current;
+        }
+        current = current.square();
```
tcoratger, line 19: Here I think that in the last run of the loop we don't need to do the square no? Because we never use current again after that

tcoratger (approved):
Minor comments, once the CI is fixed, LGTM!

PR #1850, opened 2026-06-14
fix(whir): reject non-redundant code rates in WhirConfig::new

tcoratger:
## Summary

A WHIR code rate of `2^-0 = 1` leaves the Reed-Solomon code with no redundancy. In that case the proximity parameter `delta` collapses to `<= 0`, so the derived STIR query count `ceil(-lambda / log2(1 - delta))` is non-positive and the `as usize` cast turns it into **0**. The verifier then makes zero proximity queries and accepts any committed function, while `WhirConfig::new` still returns `Ok` and reports the full target security.

This was found during a soundness review of the WHIR implementation. It requires a misconfiguration (rate `1`, e.g. a typo of `0` for `1` in `starting_log_inv_rate`), but the failure is silent and catastrophic — a config with no proximity test at all that still claims `security_level` bits.

## Fix

Reject `log_inv_rate == 0` at construction, the natural boundary for validating a domain parameter:

- `starting_log_inv_rate == 0` → `NonRedundantStartingRate`.
- any explicit `round_log_inv_rates` entry `== 0` → `NonRedundantRoundRate` (derived per-round rates inherit from the starting rate and only grow, so only caller-supplied rates can trip this).

For every soundness regime, `delta > 0` is equivalent to rate `<= 1/2` (`log_inv_rate >= 1`):
- UD: `delta = (1 - rho)/2 > 0  <=>  rho < 1`
- JB: `delta = 1 - sqrt(rho) - sqrt(rho)/20 > 0  <=>  rho < (20/21)^2 ~ 0.907`
- CB: `delta = 1 - rho - rho/2 > 0  <=>  rho < 2/3`

All hold at `log_inv_rate = 1` and fail at `log_inv_rate = 0`, so the rate check is exactly the right precondition. `ZkWhirConfig::new` delegates to `WhirConfig::new`, so the hiding variant is covered too.

## Tests

- `new_rejects_zero_starting_rate`
- `new_rejects_zero_explicit_round_rate`
- `valid_rates_yield_positive_query_counts` — confirms the guard does not over-reject: a redundant rate yields a positive query count for every phase across UD/JB/CB.

`cargo fmt --check`, `cargo clippy --lib --all-features`, and `cargo test --lib` (124 passed) all green.

### DIFF
--- a/whir/src/parameters/soundness.rs
+++ b/whir/src/parameters/soundness.rs
@@ -205,6 +205,15 @@ impl SecurityAssumption {
     /// Compute the number of queries to match the security level
     /// The error to drive down is (1-delta)^t < 2^-lambda.
     /// Where delta is set as in the `log_1_delta` function.
+    ///
+    /// Requires a redundant code rate, at most `1/2`:
+    /// - A redundant rate keeps the proximity parameter `delta > 0`.
+    /// - With `delta > 0` the term `log2(1 - delta)` is negative.
+    /// - Dividing `-lambda` by a negative number gives a positive, finite count.
+    /// - A rate of `1` forces `delta <= 0` and rounds the count down to zero.
+    /// - Zero queries would make the proximity test check nothing.
+    ///
+    /// Configuration construction rejects rate `1`, so the precondition holds.
     #[must_use]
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
LGTM

PR #1851, opened 2026-06-14
fix(whir): round proof-of-work bits up instead of down

tcoratger:
## Summary

Proof-of-work tops up a phase's algebraic soundness to the target: `achieved = algebraic_bits + pow_bits`. The grinding amount `security_level - algebraic_bits` is real-valued and was cast with `as usize`, which **floors**. Flooring discards the fractional part, so achieved security undershoots the target by up to one bit on every phase: the starting fold, each round's query and fold steps, the final query, and the final fold.

The query-count derivation already rounds the correct way (`libm::ceil` in `queries()`); the PoW casts did not.

### Walk-through

A phase with `algebraic = 107.3` bits and `security_level = 128`:

- gap `= max(0, 128 - 107.3) = 20.7`
- `20.7 as usize = 20` (the `0.7` is lost)
- achieved `= 107.3 + 20 = 127.3` bits, below the 128 target

With `ceil`: `21` bits → achieved `128.3 >= 128`.

The shortfall is always `< 1` bit, so this is a margin/accounting fix, not an exploitable break — but it makes a 128-bit request actually deliver 128 bits per phase.

## Fix

Route every derived PoW gap through one helper that rounds up:

```rust
fn ceil_pow_bits(gap: f64) -> usize {
    libm::ceil(gap) as usize
}
```

Applied at all five cast sites: per-round query PoW, per-round folding PoW, starting folding PoW, final query PoW, final folding PoW.

## Tests

- `ceil_pow_bits_rounds_up` — pins the rounding direction (`0.0 -> 0`, `20.0 -> 20`, `19.001 -> 20`, `20.7 -> 21`).
- `derived_pow_reaches_security_target_per_phase` — for unique decoding (no OOD, feasible over the 31-bit base field, fractional gaps), asserts `algebraic_bits + pow_bits >= security_level` for every round and the final query phase. This fails under the old floor and passes under ceil.

`cargo fmt --check`, `cargo clippy --lib --all-features`, and `cargo test --lib` (123 passed) all green; end-to-end prove/verify still pass with the slightly higher grinding amounts.

### DIFF
--- a/whir/src/parameters/whir.rs
+++ b/whir/src/parameters/whir.rs
@@ -149,6 +149,21 @@ where
     }
 }
 
+/// Round a real-valued proof-of-work gap up to whole grinding bits.
+///
+/// Grinding contributes whole bits of security: a witness with `b` leading
+/// zero bits adds exactly `b` bits.
+///
+/// The gap to grind is `security_level - algebraic_bits`, a real number.
+/// - Flooring drops the fractional part.
+/// - Achieved security `algebraic_bits + floor(gap)` then dips below target.
+/// - Rounding up keeps `algebraic_bits + ceil(gap) >= security_level`.
+///
+/// Mirrors the query-count derivation, which rounds up for the same reason.
+fn ceil_pow_bits(gap: f64) -> usize {
+    libm::ceil(gap) as usize
+}
+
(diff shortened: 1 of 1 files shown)


Nashtare (approved):
LGTM

PR #1853, opened 2026-06-15
fix(circle): reject input matrices opened at zero points

wstran:
Circle's verifier derived each input matrix's `width` for the authenticated MMCS `Dimensions` from the claimed evaluation count when the matrix had opening points, but fell back to the proof-supplied `opened_row.len()` when it had none. `check_widths` in the MMCS is the only authenticator of row boundaries in the flattened leaf hash, so a no-points matrix left its boundary unchecked.

This is the circle analog of #1793, which fixed the identical gap in two-adic FRI. Every input matrix is opened at >= 1 point, so reject the degenerate case with a new `InputError::MatrixWithoutOpeningPoints` error instead of trusting the proof.

The width decision is extracted into a small `pin_matrix_width` helper so it can be unit-tested directly: circle's `verify` observes opening values into the challenger before the proof-of-work check, so stripping a matrix's points fails earlier in the transcript and cannot reach the guard through the full verify path (unlike two-adic, where `run_verify_fri` exercises it end to end).

### DIFF
--- a/circle/src/pcs.rs
+++ b/circle/src/pcs.rs
@@ -83,6 +83,31 @@ where
     /// The DEEP-quotient denominator vanishes there, so the row cannot be reduced.
     #[error("opening point coincides with a query point")]
     OpeningPointMatchesQueryPoint,
+    #[error(
+        "batch {batch}, matrix {matrix}: opened at no points; its width cannot be authenticated"
+    )]
+    MatrixWithoutOpeningPoints { batch: usize, matrix: usize },
+}
+
+/// Authenticated dimensions for one input matrix.
+///
+/// The width is pinned to the first opening point's claimed evaluation count.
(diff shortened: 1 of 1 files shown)


circle/src/pcs.rs
```diff
@@ -83,6 +83,31 @@ where
...
+fn pin_matrix_width<Challenge>(
+    height: usize,
+    points_and_values: &[(Challenge, Vec<Challenge>)],
+) -> Option<Dimensions> {
+    let (_, values) = points_and_values.first()?;
+    Some(Dimensions {
+        width: values.len(),
+        height,
+    })
 }
```
tcoratger: Since this is used at a single location and the logic is pretty simple, probably better to inline the logic at the appropriate location directly and remove the function

tcoratger (approved):

PR #1854, opened 2026-06-15
refactor(sumcheck): trim test-only API surface and dedup SVO virtual fold

tcoratger:
## What

Three small, behavior-preserving cleanups in the sumcheck `constraints`/`layout` code:

- **`EqStatement`** — move the test-only satisfaction oracle (`verify`) into the test module, and drop the now-unused `iter`/`concatenate` helpers together with the two tests that only exercised `concatenate`. `EqStatement` is consumed by `whir` solely through `initialize` / `add_evaluated_constraint` / `num_variables`, so these helpers were never part of the production surface.
- **`PrefixProver::combine_weights_packed`** — fold virtual claims through `SvoPoint::accumulate_into_packed` instead of re-deriving the point split and equality scale inline. Identical arithmetic, one shared code path, and it turns that method from test-only into a live caller.
- **`Verifier::num_variables_table`** — demote to `pub(crate)`; it has no cross-crate caller.

## What was deliberately *not* changed

While auditing the area I confirmed two things the surface-trim should leave alone:

- **`SelectStatement::verify` stays public** — `whir`'s production verifier calls it on the final STIR statement (`whir/src/pcs/verifier/mod.rs`), so it is not dead.
- **The `layout::opening` claim types stay public** — they appear in the signatures of the public `ZkLayout` trait (`impl Iterator<Item = &ProverMultiClaim<…>>`, etc.), which `whir` consumes as a generic bound. Demoting them to `pub(crate)` breaks the public interface.

This part of the crate is under active construction (next-openings, #1821 and follow-ups), so the change is intentionally limited to what is provably safe today; no protocol code is deleted.

## Testing

- `cargo test -p p3-sumcheck --all-features` — 218 passed.
- `cargo test -p p3-whir --all-features` — 121 passed.
- `cargo clippy -p p3-sumcheck --all-features --all-targets` — clean.
- `cargo fmt --check` — clean.

### DIFF
--- a/sumcheck/src/constraints/statement/eq.rs
+++ b/sumcheck/src/constraints/statement/eq.rs
@@ -191,11 +191,6 @@ impl<F: Field> EqStatement<F> {
         self.points.is_empty()
     }
 
-    /// Returns an iterator over the evaluation constraints in the statement.
-    pub fn iter(&self) -> impl Iterator<Item = (&Point<F>, &F)> {
-        self.points.iter().zip(self.evaluations.iter())
-    }
-
     /// Returns the number of constraints in the statement.
     #[must_use]
     pub const fn len(&self) -> usize {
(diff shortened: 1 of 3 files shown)


Nashtare (commented):
Are we sure those will not be of use for downstream users? They are part of the public API, so unless we're 100% certain I think it's not worth the trouble of making them unavailable later on

tcoratger:
> Are we sure those will not be of use for downstream users? They are part of the public API, so unless we're 100% certain I think it's not worth the trouble of making them unavailable later on

I'm not entirely sure; I'd say these methods aren't useful except in very specific cases because their main use is for WHIR. But I agree, it's a very small refactoring, probably not worth it. Maybe we can wait for the next release or something to double-check! Let's close for now.

PR #1855, opened 2026-06-15
perf(mersenne-31): add deferred-reduction `dot_product` for AVX2/AVX-512

Nashtare:
Override `PrimeCharacteristicRing::dot_product` for the x86 Mersenne31 packings, which previously fell through to the generic default doing one eager reduction per element.

Accumulate raw 32x32->64-bit products in 64-bit lanes (even/odd 32-bit indices kept apart for vpmuludq), folding every 3 iterations to keep 2^33 + 3*P^2 < 2^64, then reduce once at the end. This mirrors the existing M31 NEON deferred kernel and replaces N reductions with a single final reduction.

Note that I can't test it locally as I don't have access to an AVX512 machine.

### DIFF
--- a/mersenne-31/src/x86_64_avx2/packing.rs
+++ b/mersenne-31/src/x86_64_avx2/packing.rs
@@ -24,6 +24,8 @@ use crate::{Mersenne31, mul_2exp_i};
 
 const WIDTH: usize = 8;
 pub(crate) const P: __m256i = unsafe { transmute::<[u32; WIDTH], _>([0x7fffffff; WIDTH]) };
+/// The low 31 bits of each 64-bit lane, used to mask off the high half during folding.
+const LOW31_64: __m256i = unsafe { transmute::<[u64; 4], _>([0x7fffffff; 4]) };
 
 /// Vectorized AVX2 implementation of `Mersenne31` arithmetic.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


mersenne-31/src/x86_64_avx2/packing.rs
```diff
@@ -397,6 +404,71 @@ pub(crate) fn exp5(x: __m256i) -> __m256i {
...
+        let evn = fold_u64(fold_u64(acc_evn));
+        let odd = fold_u64(fold_u64(acc_odd));
+
+        // The even results sit in the low 32 bits of each 64-bit lane (indices 0, 2, 4, 6);
+        // shift the odd results up into the high halves (indices 1, 3, 5, 7) and merge.
+        let odd_shifted = x86_64::_mm256_slli_epi64::<32>(odd);
+        let t = x86_64::_mm256_blend_epi32::<0b10101010>(evn, odd_shifted);
+
+        // Final reduction of values in `0..=2 P` to the canonical `0..=P`.
+        let u = x86_64::_mm256_sub_epi32(t, P);
```
tcoratger: `u` shadows the `u` input array. Maybe this is intentional but a bit confusing no? It's totally safe since `u` has already been consumed in the loop though!

mersenne-31/src/x86_64_avx512/packing.rs
```diff
@@ -416,6 +423,71 @@ pub(crate) fn exp5(x: __m512i) -> __m512i {
...
+        let evn = fold_u64(fold_u64(acc_evn));
+        let odd = fold_u64(fold_u64(acc_odd));
+
+        // The even results sit in the low 32 bits of each 64-bit lane (the even 32-bit
+        // indices); shift the odd results up into the high halves (the odd indices) and merge.
+        let odd_shifted = x86_64::_mm512_slli_epi64::<32>(odd);
+        let t = x86_64::_mm512_mask_blend_epi32(ODDS, evn, odd_shifted);
+
+        // Final reduction of values in `0..=2 P` to the canonical `0..=P`.
+        let u = x86_64::_mm512_sub_epi32(t, P);
```
tcoratger: Same here for `u`

tcoratger (approved):
LGTM

mersenne-31/src/x86_64_avx2/packing.rs
Nashtare: renamed to `t_sub_p` here and below

PR #1856, opened 2026-06-15
refactor(sumcheck): share prover claim state via StackedClaims

tcoratger:
## What

The prefix and suffix stacked-sumcheck provers carried the **same six fields** and re-implemented the **same plumbing verbatim** (`commit`, `sum`, `num_claims`, the trait getters, and the `ZkLayout` accessors). This extracts that shared state and logic into one `StackedClaims<F, EF>` value that both provers embed.

- **New `StackedClaims<F, EF>`** holds the source tables, slot placements, stacked arity, folding depth, per-table concrete claims, and virtual claims. It owns the shared `sum`, `num_claims`, `concrete_claims`, and getter logic.
- **`PrefixProver`** embeds it alongside the committed polynomial; **`SuffixProver`** embeds it alone (suffix mode discards the stacked polynomial).
- **`Layout`** gains a `claims()` accessor; `commit` and the getters (`folding`, `num_variables`, `num_variables_table`, `num_claims`) become default methods built on it.
- **`ZkLayout::{concrete_claims, virtual_claims, batched_sum}`** become default methods too, leaving each prover impl with only `zk_residual_handoff` — the one operation that genuinely branches on the binding direction.

Removes ~220 lines of verbatim duplication (365 deletions / 147 insertions + the new ~150-line module).

## Behavior

Unchanged by construction: the same fields, the same insertion-order alpha assignment, the same Fiat-Shamir sequence. The shared `sum` body is moved verbatim, and the `debug_assert_eq!(prod_poly.dot_product(), sum)` cross-checks in both modes still hold.

## A note on visibility

`StackedClaims` is `pub` (its fields remain `pub(crate)`, so it is not constructible or mutable from outside the crate) because the public `Layout::claims()` returns it. This matches how the crate already exposes `PrefixProver`/`SuffixProver` and the `layout::opening` claim types — all `pub` internal types reachable through the public `Layout`/`ZkLayout` traits.

## Testing

- `cargo test -p p3-sumcheck --all-features` — all pass (218).
- `cargo test -p p3-whir --all-features` — all pass (121).
- `cargo clippy -p p3-sumcheck --all-features --all-targets` — clean.
- `cargo fmt --check` — clean.


sumcheck/src/zk/prover/layout.rs
```diff
@@ -21,13 +21,19 @@ where
     EF: ExtensionField<F>,
 {
     /// Walks concrete claims in placement order.
-    fn concrete_claims(&self) -> impl Iterator<Item = &ProverMultiClaim<F, EF>>;
+    fn concrete_claims(&self) -> impl Iterator<Item = &ProverMultiClaim<F, EF>> {
+        self.claims().concrete_claims()
+    }
```
Nashtare, line 27:
```suggestion
    #[inline]
    fn concrete_claims(&self) -> impl Iterator<Item = &ProverMultiClaim<F, EF>> {
        self.claims().concrete_claims()
    }
```

sumcheck/src/zk/prover/layout.rs
```diff
@@ -21,13 +21,19 @@ where
...
 
     /// Returns the virtual-claim slice.
-    fn virtual_claims(&self) -> &[ProverVirtualClaim<EF>];
+    fn virtual_claims(&self) -> &[ProverVirtualClaim<EF>] {
+        &self.claims().virtual_claims
+    }
```
Nashtare, line 33:
```suggestion
    #[inline]
    fn virtual_claims(&self) -> &[ProverVirtualClaim<EF>] {
        &self.claims().virtual_claims
    }
```

sumcheck/src/zk/prover/layout.rs
```diff
@@ -21,13 +21,19 @@ where
...
 
     /// Returns the alpha-batched plain sum.
-    fn batched_sum(&self, alpha: EF) -> EF;
+    fn batched_sum(&self, alpha: EF) -> EF {
+        self.claims().sum(alpha)
+    }
```
Nashtare, line 39:
```suggestion
    #[inline]
    fn batched_sum(&self, alpha: EF) -> EF {
        self.claims().sum(alpha)
    }
```

sumcheck/src/layout/prover/suffix.rs
```diff
@@ -424,6 +373,7 @@ impl<F: TwoAdicField, EF: ExtensionField<F>> Layout<F, EF> for SuffixProver<F, E
 
             // Virtual-claim contributions: scale each claim's dot by its alpha power.
             for (vc, alpha_i) in self
+                .claims
                 .virtual_claims
                 .iter()
                 .zip(alpha.shifted_powers(alpha.exp_u64(n_claims as u64)))
```
Nashtare: I think this is already fixed on `main` but make sure the exponentiation is kept outside the loop

sumcheck/src/layout/prover/prefix.rs
```diff
@@ -340,6 +286,7 @@ impl<F: TwoAdicField, EF: ExtensionField<F>> Layout<F, EF> for PrefixProver<F, E
             }
 
             for (vc, alpha_i) in self
+                .claims
                 .virtual_claims
                 .iter()
                 .zip(alpha.shifted_powers(alpha.exp_u64(n_claims as u64)))
```
Nashtare: same here, I think this is already fixed on `main` but make sure the exponentiation is kept outside the loop

Nashtare (approved):
LGTM

sumcheck/src/layout/prover/suffix.rs
tcoratger: Fixed with the conflicts

sumcheck/src/layout/prover/prefix.rs
tcoratger: Fixed with the conflicts

PR #1858, opened 2026-06-15
refactor(sumcheck): consistent expect message for concrete opening column index

tcoratger:
In `Verifier::constraint`, the next-openings loop used a bare `.unwrap()` on `opening.poly_idx()`, while the current-openings loop used a descriptive `.expect(...)` for the identical invariant (a concrete opening always carries `Some(column_index)`). This unifies them on the same message so an impossible-state failure reads the same in both groups.

Behavior-preserving. `cargo test -p p3-sumcheck`, clippy, and fmt are clean.


Nashtare (approved):

Nashtare:
Maybe worth addressing all unwrap() in one go across a crate to reduce noise

PR #1863, opened 2026-06-15
refactor(sumcheck): compute SVO grid length from a single log2

tcoratger:
In `evals_01inf_grid_prefix`, `grid_len` and `l` each independently called `log2_strict_usize(evals.len())`.

Compute `l` once, then derive `grid_len = 3^l` from it. Behavior is unchanged; existing grid tests cover it.


Nashtare (approved):
May be worth combining all these micro PRs into 1 for ease of review :)

tcoratger:
> May be worth combining all these micro PRs into 1 for ease of review :)

Sure will do that in the future! I think that for now I don't have others in the pipeline, I've checked the most obvious ones :)

PR #1871, opened 2026-06-16
feat(multi-stark): add constraint metadata module

tcoratger:
## Summary

Adds the next piece of `p3-multi-stark`: `ConstraintMetadata`, the structural facts the multilinear prover and verifier need *before* any proving begins, derived once from an AIR definition.

This is a pure read-only layer over the existing `p3-air` symbolic machinery — no proving logic, no trace dependence.

## Public surface

`metadata::ConstraintMetadata` with `from_air::<F, EF, A>(air, layout)`:

- `num_constraints` — number of asserted constraints.
- `max_constraint_degree` — largest constraint degree under the symbolic (quotient) selector convention.
- `next_row_main_columns` — main-trace columns whose next row a constraint reads (these need a shifted-successor opening claim).
- `next_row_preprocessed_columns` — same, for the preprocessed trace.

It runs at most one symbolic evaluation of the AIR, and **zero** when the AIR supplies both a `num_constraints()` and a `max_constraint_degree()` hint.

## Note on degree

`max_constraint_degree` is the `p3-air` symbolic degree, where the transition selector counts as degree 0 and the first/last-row selectors count as degree 1 (the quotient convention). It is intentionally **not** mapped to a per-round sumcheck degree here: under the multilinear zerocheck the transition selector (`1 - prod x_j`) is itself degree 1 per variable, so the round degree depends on the selector model and is derived by the zerocheck, not by adding one to this value. Keeping that decision out of the metadata avoids baking in a formula that would undercount transition-heavy AIRs.

## Test plan

- [x] `cargo test -p p3-multi-stark` — 9 unit tests pass (5 existing + 4 new)
  - Fibonacci AIR: 5 constraints, degree 2, both columns shifted, no preprocessed
  - degree-3 product AIR with selective next-row access
  - preprocessed AIR with a single preprocessed next-row column
  - hint AIR whose `eval` panics, proving the hint path skips symbolic evaluation
- [x] `cargo clippy -p p3-multi-stark --all-targets` — clean
- [x] `cargo build --target thumbv7em-none-eabi -p p3-multi-stark` — builds (`no_std`)
- [x] `cargo fmt --check` — clean

### DIFF
--- a/multi-stark/src/metadata.rs
+++ b/multi-stark/src/metadata.rs
@@ -11,9 +11,15 @@ use p3_field::{ExtensionField, Field};
 /// They depend only on the constraint structure, not on the trace contents or the field of evaluation.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ConstraintMetadata {
-    /// Number of constraints the AIR asserts.
+    /// Number of base constraints the AIR asserts.
+    ///
+    /// Covers transition and boundary constraints.
+    /// Excludes lookup and permutation arguments, matching the AIR's own constraint-count hint.
     pub num_constraints: usize,
-    /// Largest total degree of any asserted constraint.
+    /// Upper bound on the largest total degree of any asserted constraint.
+    ///
+    /// Exact when derived by symbolic evaluation.
+    /// May overestimate when the AIR supplies a hint, which need only be an upper bound.
     ///
     /// Degree follows the symbolic selector convention:
     /// - first-row and last-row selectors each count as degree one,
@@ -67,8 +73,8 @@ impl ConstraintMetadata {
                 //     ext[j]  : j-th extension-field constraint (lookups / permutation args)
                 let (base, ext) = get_all_symbolic_constraints::<F, EF, A>(air, layout);
 
-                // Total count is base-field plus extension-field constraints.
-                let count = hinted_count.unwrap_or(base.len() + ext.len());
+                // Count covers base constraints only, matching the scope of the AIR's count hint.
+                let count = hinted_count.unwrap_or(base.len());
 
                 // Max degree is taken over both constraint families under the symbolic convention.
                 let degree = hinted_degree.unwrap_or_else(|| {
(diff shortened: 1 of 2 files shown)


multi-stark/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+            // At least one hint missing: evaluate symbolically once and fill in the gaps.
+            _ => {
+                // Collect every asserted constraint as a symbolic expression tree.
+                //
+                //     base[i] : i-th base-field constraint
+                //     ext[j]  : j-th extension-field constraint (lookups / permutation args)
+                let (base, ext) = get_all_symbolic_constraints::<F, EF, A>(air, layout);
+
+                // Total count is base-field plus extension-field constraints.
+                let count = hinted_count.unwrap_or(base.len() + ext.len());
```
Nashtare: just wondering: but isn't `air.num_constraints()` returning only base constraints? if so then `count` here can represent both which sounds erroneous

multi-stark/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+pub struct ConstraintMetadata {
+    /// Number of constraints the AIR asserts.
+    pub num_constraints: usize,
+    /// Largest total degree of any asserted constraint.
+    ///
+    /// Degree follows the symbolic selector convention:
+    /// - first-row and last-row selectors each count as degree one,
+    /// - the transition selector counts as degree zero,
+    /// - every trace column counts as degree one.
+    pub max_constraint_degree: usize,
```
Nashtare: nit: if we used a hint, we're only guaranteed of an upper-bound, not that this is the actual largest (reached) degree

Nashtare (approved):
Overall LGTM but I think we want the constraint count concern fixed for consistency

multi-stark/src/metadata.rs
tcoratger: Fixed

multi-stark/src/metadata.rs
tcoratger: I modified just to take into account the base constraints with a test to constrain the behaviour, we can tweak this as needed in the followup PRs

PR #1872, opened 2026-06-16
feat(multi-stark): add config trait and trace commitment

tcoratger:
## Summary

Adds the configuration and commitment foundation for `p3-multi-stark`: a config trait mirroring `StarkGenericConfig`, and a `commit_trace` step that commits an execution trace through a multilinear PCS.

No constraint logic yet — this is the wiring the prover and verifier build on.

## Public surface

- `config::MultiStarkConfig` — carries the base field, challenge field, transcript, and a `MultilinearPcs`. Mirrors `StarkGenericConfig`. It also exposes `build_witness(columns)`, which bridges committed columns into the scheme's witness form: the generic `MultilinearPcs` interface deliberately keeps the witness representation (slot layout + preprocessing/folding depth) private to the scheme, so the column→witness step lives in the config where those details are known, leaving `commit_trace` fully generic.
- `config::{Pcs, Commitment, ProverData}` — projection aliases over a config (same shape as `uni-stark`'s `Val<SC>`).
- `commit::commit_trace(config, trace, challenger)` — free function (matching the `uni-stark` `prove`/`verify` house style). The trace becomes one multilinear polynomial per column; the scheme **stacks those columns into a single polynomial and commits once**, so one commitment covers every column, and absorbs that commitment into the transcript. Keeping columns separate up to this point lets each carry its own opening claim and slot in the stacked polynomial.

The column-splitting helper is crate-private: it is a pre-stacking intermediate, not part of the public model (the committed object is the single stacked polynomial).

## Test plan

Tests use a mock commitment scheme and transcript to pin the wiring without pulling in a concrete PCS (real WHIR commit/open/verify is a later step):

- [x] `cargo test -p p3-multi-stark` — 8 unit tests pass (5 prior + 3 new)
  - column extraction reads each column in row order
  - `commit_trace` returns a commitment + prover data and absorbs the commitment into the transcript
  - `commit_trace` is deterministic for a fixed trace
- [x] `cargo clippy -p p3-multi-stark --all-targets` — clean
- [x] `cargo build --target thumbv7em-none-eabi -p p3-multi-stark` — builds (`no_std`; new deps `p3-commit`/`p3-matrix`/`p3-multilinear-util` are all `no_std`)
- [x] `cargo fmt --check` — clean

### DIFF
--- a/multi-stark/src/commit.rs
+++ b/multi-stark/src/commit.rs
@@ -22,12 +22,6 @@ pub(crate) fn trace_to_columns<F: Field>(trace: &RowMajorMatrix<F>) -> Vec<Poly<
     let width = trace.width;
     let height = trace.height();
 
-    // A multilinear polynomial has exactly `2^k` evaluations, so the height must be a power of two.
-    assert!(
-        height.is_power_of_two(),
-        "trace height {height} must be a power of two"
-    );
-
     // Build one polynomial per column.
     (0..width)
         .map(|col| {
@@ -49,8 +43,9 @@ pub(crate) fn trace_to_columns<F: Field>(trace: &RowMajorMatrix<F>) -> Vec<Poly<
 /// - One commitment therefore covers every column.
 /// - The scheme absorbs that commitment into the transcript, advancing the challenger.
 ///
-/// Keeping the columns separate up to this point lets each one carry its own
-/// opening claim and occupy its own slot in the stacked polynomial.
+/// Columns stay separate up to this point for two reasons:
+/// - each carries its own opening claim,
+/// - each occupies its own slot in the stacked polynomial.
 ///
 /// # Arguments
 ///
@@ -69,6 +64,15 @@ pub fn commit_trace<C: MultiStarkConfig>(
 ) -> (Commitment<C>, ProverData<C>) {
     // One multilinear polynomial per trace column.
     let columns = trace_to_columns(trace);
+
+    // The witness builder lays each column into a slot sized by its arity.
+    // Every column must therefore share the same number of variables.
+    debug_assert!(
+        columns
+            .windows(2)
+            .all(|pair| pair[0].num_variables() == pair[1].num_variables()),
+        "all trace columns must share the same arity"
+    );
 
     // Pack the columns into the scheme's witness form (slot layout and folding live in the config).
     let witness = config.build_witness(columns);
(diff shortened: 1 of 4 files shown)


multi-stark/src/commit.rs
```diff
@@ -0,0 +1,246 @@
...
+///
+/// # Returns
+///
+/// - The succinct commitment to all columns.
+/// - The prover-only data needed later to open the committed columns.
+pub fn commit_trace<C: MultiStarkConfig>(
+    config: &C,
+    trace: &RowMajorMatrix<C::Val>,
+    challenger: &mut C::Challenger,
+) -> (Commitment<C>, ProverData<C>) {
```
Nashtare, line 64: nit: may be worth checking the columns arity? even just in debug

multi-stark/src/commit.rs
```diff
@@ -0,0 +1,246 @@
...
+/// Panics if the trace height is not a power of two.
+pub(crate) fn trace_to_columns<F: Field>(trace: &RowMajorMatrix<F>) -> Vec<Poly<F>> {
+    let width = trace.width;
+    let height = trace.height();
+
+    // A multilinear polynomial has exactly `2^k` evaluations, so the height must be a power of two.
+    assert!(
+        height.is_power_of_two(),
+        "trace height {height} must be a power of two"
+    );
```
Nashtare: isn't this already asserted in `Poly::new()`?

```suggestion
```

multi-stark/src/commit.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        // Two independent commitments over fresh transcripts.
+        let mut first_challenger = RecordingChallenger::default();
+        let (first, _) = commit_trace(&config, &trace, &mut first_challenger);
+        let mut second_challenger = RecordingChallenger::default();
+        let (second, _) = commit_trace(&config, &trace, &mut second_challenger);
+
+        // Same trace commits to the same value.
+        assert_eq!(first, second);
+    }
```
Nashtare: feels a bit tautological to me as it's using a mock 🤔

Nashtare (approved):
LGTM! Non-blocking comments, mostly style/nits

multi-stark/src/commit.rs
tcoratger: Deleted you are right we can go back to it with a more complete scenario later on.

multi-stark/src/commit.rs
tcoratger: Yes removed

multi-stark/src/commit.rs
tcoratger: Done

PR #1874, opened 2026-06-17
refactor(util): replace `transpose` crate with portable engine

tcoratger:
## What

Replaces the unmaintained [`transpose`](https://github.com/ejmahler/transpose) crate (`0.2.3`) with an in-house portable transpose engine in `util/src/transpose/portable.rs`, and drops the dependency from the workspace.

The crate was used in exactly one place: the fallback in `transpose::transpose` for targets without a hand-written SIMD kernel (e.g. x86_64) and for element sizes other than the specialized 4- and 8-byte cases. On x86_64 this is *the* transpose path, so it matters.

## Why

- The crate is unmaintained and single-threaded.
- It is a small, well-understood algorithm; keeping it in-tree removes an external dependency and lets us improve it.

## How

`portable.rs` keeps the same three-tier, cache-oblivious strategy as the crate, so single-thread performance is at parity or better:

```text
    len <= SMALL_LEN          : direct scalar copy (fits in L1)
    SMALL_LEN < len <= MEDIUM : tiled transpose (L2-friendly)
    len > MEDIUM_LEN          : recursive split -> tiled leaves (cache-oblivious)
```

Improvements over the crate:

- **Slices, not raw pointers.** Kernels take `&[T]`/`&mut [T]` and index with `get_unchecked`. The separate shared/exclusive borrows let LLVM assume the buffers don't alias and pipeline the strided reads ahead of the stores. This is the single biggest win (~15% on read-bound wide matrices); raw pointers forfeit it.
- **Contiguous writes.** The inner loop runs over the source row index, so stores land at adjacent output addresses; only the reads are strided.
- **Constant-bound full-tile kernels** so the compiler unrolls/vectorizes the hot loop, with a separate remainder path (no per-tile `min`/branch).
- **Parallel column-band path** behind the `parallel` feature, which the crate never had. (Note: no workspace crate currently enables `p3-util/parallel`, so production stays single-threaded today — this is upside if the feature is turned on.)

Generic over any `T: Copy`.

## Tests

Proptests in `portable.rs` pin the engine against a naive reference across **every size regime** and a range of element sizes:

- `u32` and `u64` (the realistic 4-/8-byte field widths) fuzz the full strategy: degenerate shapes, scalar, tile boundaries, multi-tile, both-edge remainders, the recursion-leaf side length, rectangular wide/tall, exactly `MEDIUM_LEN`, and several recursive shapes past `MEDIUM_LEN` with assorted aspect ratios.
- `u8`, `u128`, a 3-byte type, and a 32-byte type fuzz the same regimes (plus one recursive shape) to pin generic stride handling.
- A `parallel`-feature test exercises the column-band split (square / wide / tall / narrower-than-worker-count).

## Benchmarks

Single-thread, Apple M-series, calm machine. The numbers below were collected with the temporary three-way comparison bench shown at the bottom (portable fallback vs the old crate). The committed `util/benches/transpose.rs` only differs from `main` by dropping the now-removed crate comparison arm.

| case | BabyBear (u32) — portable vs crate | Goldilocks (u64) — portable vs crate |
|---|---|---|
| small `128×128`   | 4.62 vs 4.66 µs (parity) | 6.59 vs 6.62 µs (parity) |
| medium `1024×1024`| 423 vs 425 µs (parity)   | 928 vs 931 µs (parity) |
| large_wide `4096×2048` | **7.59 vs 8.59 ms** | **16.1 vs 20.7 ms** |
| large_tall `2048×4096` | **19.4 vs 21.5 ms** | **15.3 vs 20.6 ms** |

Summary: parity on small/medium, **~10–25% faster on large** matrices single-threaded. With `--features parallel`, the portable path is ~4–5× the crate (e.g. BabyBear `large_wide` 2.1 ms vs 8.6 ms).

### Reproducing the comparison

The numbers came from a temporary setup:

1. Re-add the crate: `transpose = "0.2.3"` to the workspace `Cargo.toml` and `transpose.workspace = true` to `util/Cargo.toml`.
2. Temporarily expose the fallback for benching, in `util/src/transpose/mod.rs`:
   ```rust
   #[doc(hidden)]
   pub use portable::transpose as transpose_portable;
   ```
   and make `portable::transpose` `pub`.
3. To measure the portable fallback on aarch64 (where the public `transpose` would dispatch to NEON), comment out the SIMD dispatch block in `rectangular.rs` so it falls through to the portable engine — or just bench `transpose_portable` directly as below.

Then use this three-arm bench (`util/benches/transpose.rs`):

```rust
use core::hint::black_box;

use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
use p3_baby_bear::BabyBear;
use p3_goldilocks::Goldilocks;
use p3_util::transpose::{transpose, transpose_portable};

const BENCHMARK_SIZES: &[(usize, usize, &str)] = &[
    (128, 128, "small_square_128x128"),
    (1024, 1024, "medium_square_1024x1024"),
    (4096, 2048, "large_wide_4096x2048"),
    (2048, 4096, "large_tall_2048x4096"),
];

fn bench_field<F: PrimeCharacteristicRing + Copy + Send + Sync>(c: &mut Criterion, group: &str) {
    let mut group = c.benchmark_group(group);

    for &(width, height, name) in BENCHMARK_SIZES {
        let size = width * height;

        let input: Vec<_> = (0..size as u64).map(F::from_u64).collect();
        let mut out_dispatched = F::zero_vec(size);
        let mut out_portable = F::zero_vec(size);
        let mut out_crate = F::zero_vec(size);

        group.bench_with_input(BenchmarkId::new("dispatched", name), &size, |b, _| {
            b.iter(|| {
                transpose(black_box(&input), &mut out_dispatched, width, height);
                black_box(out_dispatched[0])
            });
        });

        group.bench_with_input(BenchmarkId::new("portable", name), &size, |b, _| {
            b.iter(|| {
                transpose_portable(black_box(&input), &mut out_portable, width, height);
                black_box(out_portable[0])
            });
        });

        group.bench_with_input(BenchmarkId::new("crate", name), &size, |b, _| {
            b.iter(|| {
                transpose::transpose(black_box(&input), &mut out_crate, width, height);
                black_box(out_crate[0])
            });
        });
    }

    group.finish();
}

fn bench_transpose_babybear(c: &mut Criterion) {
    bench_field::<BabyBear>(c, "transpose_babybear");
}

fn bench_transpose_goldilocks(c: &mut Criterion) {
    bench_field::<Goldilocks>(c, "transpose_goldilocks");
}

criterion_group!(benches, bench_transpose_babybear, bench_transpose_goldilocks);
criterion_main!(benches);
```

Run with `cargo bench -p p3-util --bench transpose` (add `--features parallel` to measure the parallel path).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,6 @@ tracing-forest = "0.3.1"
 tracing-subscriber = { version = "0.3.23", default-features = false, features = [
     "alloc",
 ] }
-transpose = "0.2.3"
 
 # Local dependencies
 p3-air = { path = "air", version = "0.6.0" }
--- a/util/Cargo.toml
+++ b/util/Cargo.toml
@@ -15,7 +15,6 @@ parallel = ["rayon"]
 [dependencies]
 rayon = { workspace = true, optional = true }
 serde.workspace = true
-transpose.workspace = true
 
 [dev-dependencies]
 p3-baby-bear = { path = "../baby-bear" }
--- a/util/benches/transpose.rs
+++ b/util/benches/transpose.rs
@@ -21,20 +21,12 @@ fn bench_transpose_babybear(c: &mut Criterion) {
         let size = width * height;
 
         let input: Vec<_> = (0..size as u64).map(BabyBear::from_u64).collect();
-        let mut output_neon = BabyBear::zero_vec(size);
-        let mut output_crate = BabyBear::zero_vec(size);
+        let mut output = BabyBear::zero_vec(size);
 
         group.bench_with_input(BenchmarkId::new("transpose_util", name), &size, |b, _| {
             b.iter(|| {
-                transpose(black_box(&input), &mut output_neon, width, height);
-                black_box(output_neon[0])
-            });
-        });
-
-        group.bench_with_input(BenchmarkId::new("transpose_crate", name), &size, |b, _| {
-            b.iter(|| {
-                transpose::transpose(black_box(&input), &mut output_crate, width, height);
-                black_box(output_crate[0])
+                transpose(black_box(&input), &mut output, width, height);
+                black_box(output[0])
             });
         });
     }
@@ -49,20 +41,12 @@ fn bench_transpose_goldilocks(c: &mut Criterion) {
         let size = width * height;
 
         let input: Vec<_> = (0..size as u64).map(Goldilocks::from_u64).collect();
-        let mut output_neon = Goldilocks::zero_vec(size);
-        let mut output_crate = Goldilocks::zero_vec(size);
+        let mut output = Goldilocks::zero_vec(size);
 
         group.bench_with_input(BenchmarkId::new("transpose_util", name), &size, |b, _| {
             b.iter(|| {
-                transpose(black_box(&input), &mut output_neon, width, height);
-                black_box(output_neon[0])
-            });
-        });
-
-        group.bench_with_input(BenchmarkId::new("transpose_crate", name), &size, |b, _| {
-            b.iter(|| {
-                transpose::transpose(black_box(&input), &mut output_crate, width, height);
-                black_box(output_crate[0])
+                transpose(black_box(&input), &mut output, width, height);
+                black_box(output[0])
             });
         });
     }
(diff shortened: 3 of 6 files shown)


Nashtare (approved):
Nice!

PR #1875, opened 2026-06-17
feat(air): support degree-4 complex-tower extension in symbolic expressions

Nashtare:
Add `From` and `Algebra` impls for `BinomialExtensionField<Complex<F>, 2>` (e.g. `QM31` over `Mersenne31`), whose binomial base parameter differs from the symbolic base `F` and so is not covered by the generic binomial impl.

This lets `SymbolicExpressionExt<F, BinomialExtensionField<Complex<F>, 2>>` satisfy `Algebra<EF>`, as required to fold extension constraints over a complex-tower challenge field.

### DIFF
--- a/air/src/symbolic/expression_ext.rs
+++ b/air/src/symbolic/expression_ext.rs
@@ -1,7 +1,7 @@
 use alloc::sync::Arc;
 
 use p3_field::extension::{
-    BinomialExtensionField, CubicTrinomialExtensionField, QuinticTrinomialExtensionField,
+    BinomialExtensionField, Complex, CubicTrinomialExtensionField, QuinticTrinomialExtensionField,
 };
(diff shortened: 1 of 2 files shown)


air/src/symbolic/expression_ext.rs
```diff
@@ -835,4 +859,21 @@ mod tests {
         assert_eq!(serde_json::to_string(&decoded).unwrap(), json);
         assert_eq!(decoded.degree_multiple(), expr.degree_multiple());
     }
+
+    #[test]
+    fn complex_tower_extension_constant_lowers_to_leaf() {
+        use p3_mersenne_31::{Mersenne31, QM31};
```
tcoratger: Nit: maybe this import could go at the top of the test module instead?

tcoratger (approved):

PR #1876, opened 2026-06-17
refactor(util): remove zip_eq helper in favor of std zip + length guard

tcoratger:
Just a proposal to avoid reinventing the wheel of existing std methods, feel free to close if you don't like it and prefer the old way :) 

## Summary

The `p3_util::zip_eq` module wrapped `Iterator::zip` with an eager length check that returned a caller-supplied error instead of panicking (unlike `itertools::zip_eq`). It has no std equivalent, but every call site can express the same intent with an explicit length guard followed by std `zip` — removing the bespoke `ZipEq` iterator type and its impls.

This deletes the module and rewrites all 9 call sites across 5 crates:

```rust
// before
for (a, b) in zip_eq(xs, ys, SomeError)? { ... }

// after
if xs.len() != ys.len() {
    return Err(SomeError);
}
for (a, b) in xs.iter().zip(ys) { ... }
```

## Call sites

- `commit/src/testing.rs` — coefficient vectors vs opened matrices
- `uni-stark/src/verifier.rs` — quotient-chunk domains vs opened chunks (lifted out of a `vec!` literal into a `let`)
- `batch-stark/src/verifier/mod.rs` — per-instance quotient domains (`.into()` restores the `VerificationError -> BatchVerificationError` conversion that `?` did implicitly)
- `merkle-tree/src/hiding_mmcs.rs` — opened rows vs salt rows
- `circle/src/pcs.rs` — 5 zips, including a double-nested one collapsed into a single `.zip().zip()` chain behind one combined length check

## Notes

- Behavior is preserved: the same typed errors are returned on a length mismatch, still eagerly (before any iteration).
- The circle code keeps both guards that check the same `opened_values.len() == mats.len()` invariant, mirroring the two original `zip_eq` calls (defense in depth; the second is technically redundant).

## Testing

- `cargo build`, `cargo clippy` (no warnings), and `cargo test` across all 5 affected crates pass.


Nashtare (commented):
I think this is rather risky. In some places, we rely explicitely on iterators being of the same lengths when zipped, otherwise the code can panic or yield invalid results. Decoupling the check to the zipped iterator increases the risk surface, especially in the future with follow-up code refactorings. I'm not sure we gain much in doing so rather than keeping the length checked opaque inside the zip_eq black box.

tcoratger:
> I think this is rather risky. In some places, we rely explicitely on iterators being of the same lengths when zipped, otherwise the code can panic or yield invalid results. Decoupling the check to the zipped iterator increases the risk surface, especially in the future with follow-up code refactorings. I'm not sure we gain much in doing so rather than keeping the length checked opaque inside the zip_eq black box.

No worries, let's close this one for now! 

I found the `zip_eq` a bit confusing with the existing standard one, which panics if there's an inequality on both sides, but I agree with you, we don't gain anything at all except deleting a very small file, so nothing huge here! 

Let's close for now! It's a shame that Rust's `zip_eq` doesn't allow us to propagate an error...

PR #1877, opened 2026-06-17
refactor(util): replace hand-rolled assume with core::hint::assert_unchecked

tcoratger:
## Summary

`p3_util::assume` predates `core::hint::assert_unchecked`, which is now stable (and `const`-stable). The std intrinsic gives the same optimizer promise — \"this predicate always holds\" — plus a debug-time precondition check, so the hand-rolled wrapper no longer earns its place:

```rust
// before
pub const unsafe fn assume(p: bool) {
    debug_assert!(p);
    if !p { unsafe { unreachable_unchecked(); } }
}
```

This deletes `assume` and calls `core::hint::assert_unchecked` directly.

## Changes

- **`util/src/lib.rs`**: remove the `assume` function; swap the `unreachable_unchecked` import (only `assume` used it) for `assert_unchecked`; call it directly in `log2_strict_usize`.
- **`goldilocks/src/goldilocks.rs`** (the only external user): drop `assume` from the `p3_util` import, add `use core::hint::assert_unchecked;`, and update all 4 call sites (they remain inside their existing `unsafe` blocks, since `assert_unchecked` is also `unsafe`). Two nearby comments referencing the old name were reworded.

## Behavior

Identical optimizer contract. The only nuance: the old `assume` used `debug_assert!` (unwinding panic in debug), whereas `assert_unchecked` uses std's precondition check (aborts in debug). No release-mode difference.

## Testing

On nightly `1.98.0`: `cargo build`, `cargo clippy` (no warnings), and `cargo test` for `p3-util` and `p3-goldilocks` all pass (340 + 55 tests, 0 failures).

### DIFF
--- a/goldilocks/src/goldilocks.rs
+++ b/goldilocks/src/goldilocks.rs
@@ -2,6 +2,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt::{Debug, Display, Formatter};
 use core::hash::{Hash, Hasher};
+use core::hint::assert_unchecked;
 use core::iter::{Product, Sum};
 use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 use core::{array, fmt};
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
LGTM, although this looks a bit different on Godbolt.

tcoratger:
> LGTM, although this looks a bit different on Godbolt.

Mhhh I've checked the Rust doc, this looks mostly the same: https://doc.rust-lang.org/std/hint/fn.assert_unchecked.html and I guess this hasn't been used because this is stable only since 1.81.0.

PR #1878, opened 2026-06-17
perf(air): make symbolic degree calculation trace-size aware

### LINKED ISSUE #1382: Improve degree calculation

Nashtare:
Resolves #1382.

The SymbolicAir degree calculation was slightly off in a trace-size-dependent way: it treated the transition selector as a constant (it is linear) and overestimated the degree of periodic columns.

Add `SymbolicExpr::poly_degree(trace_len, periodic_periods)` computing the exact constraint-polynomial degree over a trace of length N: the transition selector contributes degree 1, and a periodic column of period p contributes N - N/p (versus the N-1 of a regular column). `constraint_degree_from_poly_degree` converts that absolute degree back into the unit constraint degree the existing quotient/ZK/security pipeline consumes, so the proven chunk-sizing math is untouched. The result is provably never larger than the previous degree and is tighter when periodic reductions cross a power-of-two quotient-chunk boundary.

Thread the base trace length N through get_max_constraint_degree[_extension], get_log_quotient_degree_extension and get_log_num_quotient_chunks in both uni-stark and batch-stark, passing it consistently from the prover, verifier (1 << base_degree_bits, not the ZK-extended length) and the batch keygen path so all three agree on the chunk count.

`poly_degree` is memoized on Arc pointer identity, keeping it linear in the number of distinct DAG nodes (the symbolic expression shares Arc children, so a naive walk is exponential). When an AIR has no periodic columns the trace-aware degree equals the cached degree_multiple, so that case short-circuits and skips the walk entirely.

### DIFF
--- a/air/src/symbolic/builder.rs
+++ b/air/src/symbolic/builder.rs
@@ -59,11 +59,11 @@ impl AirLayout {
 /// degree at most `d * (trace_len - 1) + 1` (the `+ 1` covers the linear
 /// transition selector). We therefore return the smallest `d` satisfying
 /// `poly_degree <= d * (trace_len - 1) + 1`.
-pub fn constraint_degree_from_poly_degree(poly_degree: usize, trace_len: usize) -> usize {
+pub const fn constraint_degree_from_poly_degree(poly_degree: usize, trace_len: usize) -> usize {
     if poly_degree == 0 {
         return 0;
     }
-    let denom = trace_len.saturating_sub(1).max(1);
+    let denom = if trace_len > 1 { trace_len - 1 } else { 1 };
     (poly_degree - 1).div_ceil(denom)
 }
 
--- a/air/src/symbolic/builder.rs
+++ b/air/src/symbolic/builder.rs
@@ -51,13 +51,29 @@ impl AirLayout {
     }
 }
 
+/// Convert an absolute constraint-polynomial degree into the "constraint degree"
+/// used to size the quotient: the number of degree-`(trace_len - 1)` trace
+/// polynomials the constraint behaves like.
+///
+/// The quotient machinery bounds a constraint of degree `d` by a polynomial of
+/// degree at most `d * (trace_len - 1) + 1` (the `+ 1` covers the linear
+/// transition selector). We therefore return the smallest `d` satisfying
+/// `poly_degree <= d * (trace_len - 1) + 1`.
+pub fn constraint_degree_from_poly_degree(poly_degree: usize, trace_len: usize) -> usize {
+    if poly_degree == 0 {
+        return 0;
+    }
+    let denom = trace_len.saturating_sub(1).max(1);
+    (poly_degree - 1).div_ceil(denom)
+}
+
 #[instrument(skip_all, level = "debug")]
-pub fn get_max_constraint_degree<F, A>(air: &A, layout: AirLayout) -> usize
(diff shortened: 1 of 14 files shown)


air/src/symbolic/builder.rs
```diff
@@ -52,40 +51,77 @@ impl AirLayout {
...
 #[instrument(skip_all, level = "debug")]
-pub fn get_max_constraint_degree<F, A>(air: &A, layout: AirLayout) -> usize
+pub fn get_max_constraint_degree<F, A>(air: &A, layout: AirLayout, trace_len: usize) -> usize
 where
     F: Field,
     A: Air<SymbolicAirBuilder<F>>,
 {
-    get_max_constraint_degree_extension(air, layout)
+    get_max_constraint_degree_extension(air, layout, trace_len)
 }
```
tcoratger, line 77: I wonder, I don't think this is related to this PR but do we need this function since it's just exactly a wrapper around `get_max_constraint_degree_extension` function? Couldn't we just have this function only? Maybe we don't do this right now due to the naming containing `_extension` suffix which is confusing...

air/src/symbolic/builder.rs
```diff
@@ -52,40 +51,77 @@ impl AirLayout {
...
+/// degree at most `d * (trace_len - 1) + 1` (the `+ 1` covers the linear
+/// transition selector). We therefore return the smallest `d` satisfying
+/// `poly_degree <= d * (trace_len - 1) + 1`.
+pub fn constraint_degree_from_poly_degree(poly_degree: usize, trace_len: usize) -> usize {
+    if poly_degree == 0 {
+        return 0;
+    }
+    let denom = trace_len.saturating_sub(1).max(1);
+    (poly_degree - 1).div_ceil(denom)
+}
```
tcoratger: Can't we have a constant function with this?

```suggestion
pub const fn constraint_degree_from_poly_degree(poly_degree: usize, trace_len: usize) -> usize {
    if poly_degree == 0 {
        return 0;
    }
    let denom = if trace_len > 1 { trace_len - 1 } else { 1 };
    (poly_degree - 1).div_ceil(denom)
}
```

air/src/symbolic/variable.rs
```diff
@@ -39,23 +36,47 @@ impl<F> SymbolicVariable<F> {
...
+                // `(p - 1) * (trace_len / p) = trace_len - trace_len / p`.
+                let period = periodic_periods
+                    .get(self.index)
+                    .copied()
+                    .unwrap_or(trace_len)
+                    .min(trace_len);
+                match period {
+                    0 => trace_len.saturating_sub(1),
+                    p => trace_len - trace_len / p,
+                }
```
tcoratger, line 68: Here to be sure that `trace_len / p` doesn't panic, can't we add a `debug_assert!(period.is_power_of_two())`?

tcoratger (approved):
LGTM! Small comments only!

air/src/symbolic/builder.rs
Nashtare: I'd rather keep it just for the sake of convenience, as AIRs without an extension field would needlessly have to resort to calling `get_max_constraint_degree_extension::<F, F, _>` otherwise

PR #1879, opened 2026-06-17
feat(multi-stark): AIR zerocheck via generic-degree sumcheck

tcoratger:
## What

Implements **PR3** of the multilinear (SuperSpartan-style) AIR prover: the **AIR zerocheck**.

It proves that an AIR's alpha-batched constraint vanishes on every trace row by reducing it, for a random point `tau`, to the single sum

```text
sum_x eq(tau, x) * g(x) = 0
```

and proving that sum with the generic-degree sumcheck engine (#1699).

`g(x)` is the per-row score: the AIR's constraints at row `x`, Horner-combined with `alpha`. It is never materialized — it is sampled point-by-point through the shared folder (on the hypercube for the prover, at one random point for the verifier).

## How

- `AirZerocheck::{prove, verify}` drive the sumcheck through a `RoundProver` state holding the column tables, the repeat-last successor tables, the `eq(tau, .)` weight, and the boundary selectors.
- The verifier replays the transcript, recomputes the alpha-batched constraint at the single random point through the same folder, and closes the protocol against the sumcheck's reduced sum.
- The opened column values are trusted at this layer; binding them to a commitment is the polynomial commitment scheme's job in a later PR.

## Soundness & robustness

- **Reject any nonzero claimed sum.** A zerocheck always claims zero; that anchor is the only thing tying the sumcheck chain to "the constraints vanish". Without it a prover could declare an unsatisfied trace's true sum and pass with honest openings.
- **Bind public values** into the transcript before drawing `alpha`/`tau`; the caller must observe the trace commitment first (documented).
- **Validate opening counts**: each opening vector must carry one value per main column.
- **Reject preprocessed/periodic AIRs** (unsupported in this version) with a clear panic rather than silent mis-proving.

The per-round degree is `max_constraint_degree + 2` (`+1` for the `eq` weight, `+1` because the transition selector is degree-1 but symbolically counted as 0), sound while every column/selector has per-variable degree at most one.

## Refactor

The folder's terminal step is now a consuming `eval_air` method, so the simple and preprocessed/periodic paths share one builder chain.

## Tests

`cargo test -p p3-multi-stark` — 18 passing, covering: valid Fibonacci accept, round-poly degree consistency, violated-constraint reject, tampered-opening reject, nonzero-claimed-sum reject, and wrong-opening-count reject.

## References

- Setty, Thaler, Wahby — Customizable Constraint Systems for succinct arguments (eprint 2023/552)
- Borgeaud — AIR-specific optimizations on top of SuperSpartan

### DIFF
--- a/multi-stark/Cargo.toml
+++ b/multi-stark/Cargo.toml
@@ -11,13 +11,18 @@ categories.workspace = true
 
 [dependencies]
 p3-air.workspace = true
+p3-challenger.workspace = true
 p3-commit.workspace = true
 p3-field.workspace = true
 p3-matrix.workspace = true
 p3-multilinear-util.workspace = true
+p3-sumcheck.workspace = true
+p3-util.workspace = true
+thiserror.workspace = true
 
 [dev-dependencies]
 p3-baby-bear = { path = "../baby-bear" }
+rand = { workspace = true }
 
 [lints]
 workspace = true
--- a/multi-stark/src/folder.rs
+++ b/multi-stark/src/folder.rs
@@ -122,6 +122,33 @@ where
     pub const fn into_accumulator(self) -> EF {
         self.accumulator
     }
+
+    /// Run the AIR through this folder and return its alpha-batched constraint value.
+    ///
+    /// This is the terminal step of the builder: it consumes the folder.
+    /// Attach preprocessed and periodic columns before calling, if the AIR reads them.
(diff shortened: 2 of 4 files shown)


Nashtare (approved):
LGTM! I have a couple tweaks in mind, nothing major, I'll open a follow-up PR.

PR #1881, opened 2026-06-18
perf: make `periodic_columns()` return a borrow-or-owned `Cow`

Nashtare:
In some side projects I've done, I noticed the double Vec could be sometimes heavy and induce some numbers of unfortunate `clone()` that would go away with `Cow::Borrowed`. Leaving it as potential suggestion if that makes sense


Nashtare:
CI was behaving weirdly due to some PR merged in between. Should be fine now.

tcoratger (approved):

PR #1884, opened 2026-06-19
fix(air): account for transition selector degree

### LINKED ISSUE #575: Incorrect selector degree calculation in Circle Stark
**Description:**
There seems to be an issue with the selector degree calculation in the Circle Stark protocol. Specifically, the degree of `IsTransition` selector shouldn't be 0, which is different from the traditional Stark protocol, otherwise, it will cause `Error: OodEvaluationMismatch` when verifying a correct instance of some circuit. [Here](https://github.com/cryptocoder611/Plonky3/blob/dbg/circle-degree/dbg-air/src/lib.rs) is a toy circuit to reproduce this.

**Command:**
```
cargo run --package dbg-air --example toy
```
**Terminal Output**
```bash
INFO     prove [ 1.56s | 0.06% / 100.00% ]
# ... omit several lines ...
INFO     verify [ 185ms | 99.99% / 100.00% ]
INFO     ┕━ infer log of constraint degree [ 12.6µs | 0.01% ]
Error: OodEvaluationMismatch
```

Bu after changing [this line](https://github.com/cryptocoder611/Plonky3/blob/dbg/circle-degree/uni-stark/src/symbolic_expression.rs#L47) to be `SymbolicExpression::IsTransition => 1`, the output becomes:
```bash
INFO     prove [ 2.61s | 0.05% / 100.00% ]
# ... omit several lines ...
INFO     verify [ 187ms | 99.99% / 100.00% ]
INFO     ┕━ infer log of constraint degree [ 13.3µs | 0.01% ]
```

haoshengzhen:
## Summary

- Make `BaseLeaf::IsTransition` contribute one degree multiple during symbolic degree inference.
- Add a regression test for a transition-gated quadratic constraint, which previously inferred too few quotient chunks.
- Update the no-periodic-columns fast-path comment to describe the cached degree multiple as a conservative bound.

## Motivation

`IsTransition` was treated as degree-multiple `0`, which can understate the quotient degree for Circle STARK constraints guarded by the transition selector. In particular, a transition-gated quadratic constraint could infer `0` quotient chunk bits where `1` is needed.

Fixes #575.

## Testing

- `cargo +1.95.0 test -p p3-air --lib`
- `cargo +1.95.0 test -p p3-uni-stark --lib`
- `cargo +nightly-2026-05-17 fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/air/src/symbolic/builder.rs
+++ b/air/src/symbolic/builder.rs
@@ -127,10 +127,8 @@ where
 {
     let (base_constraints, extension_constraints) = get_all_symbolic_constraints(air, layout);
 
-    // Without periodic columns the trace-size-aware degree coincides with the cached
-    // degree multiple: the linear transition selector never changes the constraint
-    // degree, and every other leaf scales with `trace_len - 1`. Reuse the cached
-    // degrees and skip the expression walk entirely.
+    // Without periodic columns the cached degree multiple is a conservative
+    // trace-size-independent bound, so we can skip the expression walk entirely.
     if layout.num_periodic_columns == 0 {
         return base_constraints
             .iter()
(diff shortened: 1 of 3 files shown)


Nashtare (commented):
This is not correct, does not fix the issue and regresses the univariate path.

PR #1885, opened 2026-06-19
perf(multi-stark): tighten AIR zerocheck round-poly and degree

Nashtare:
## Changes
- round_poly: interpolate each table's active variable in place instead of allocating a folded Poly per node, removing O(degree * tables) allocations per round (behavior-identical).
- sumcheck_degree: use poly_degree at domain size two for the exact per-variable degree (transition selector counted as degree one) instead of the max_constraint_degree + 2 upper bound, saving one evaluation per round when a boundary constraint dominates.

### DIFF
--- a/multi-stark/src/zerocheck.rs
+++ b/multi-stark/src/zerocheck.rs
@@ -445,15 +445,6 @@ where
         let mut local_row = EF::zero_vec(width);
         let mut next_row = EF::zero_vec(width);
 
-        // Interpolate a table's active variable to `z` at residual index `s`,
-        // reading the two halves in place. This matches `fix_prefix_var(z)[s]`
-        // but allocates no folded table per node.
-        let interp = |table: &Poly<EF>, s: usize, z: EF| {
-            let lo = table.as_slice()[s];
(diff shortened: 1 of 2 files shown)


multi-stark/src/zerocheck.rs
```diff
@@ -428,35 +437,39 @@ where
...
         let mut next_row = EF::zero_vec(width);
 
+        // Interpolate a table's active variable to `z` at residual index `s`,
+        // reading the two halves in place. This matches `fix_prefix_var(z)[s]`
+        // but allocates no folded table per node.
+        let interp = |table: &Poly<EF>, s: usize, z: EF| {
+            let lo = table.as_slice()[s];
+            let hi = table.as_slice()[s + half];
+            lo + (hi - lo) * z
+        };
```
tcoratger: I remember having an interaction with Claude about this. What about putting this as a small method as part of `Poly` structure? I'm pretty sure this is useful for other purposes.

tcoratger (approved):
LGTM, thanks!

PR #1886, opened 2026-06-19
feat(multi-stark): reduce the AIR zerocheck to opening claims

tcoratger:
## What

After the AIR zerocheck binds every variable to a single random point `r`, proving that the AIR holds reduces to a small set of evaluation claims about the committed trace columns, all at that one point `r`. This change makes that reduction explicit and structured, and has the verifier check it.

## Background

The zerocheck proves `sum_x eq(tau, x) * g(x) = 0`, where `g(x)` is the alpha-batched constraint evaluated at trace row `x`. The sumcheck collapses that sum over all rows down to a single evaluation of `g` at a random point `r`. To finish, the verifier must learn the committed columns at `r` and recompute `g(r)` itself. Those column evaluations are the opening claims this change introduces.

## How

A new `OpeningClaims` bundles every claim at `r`:

- **Current-row claims (`Eq`)** — the value of each column at `r`.
- **Next-row claims (`Next`)** — the repeat-last successor value of each column the AIR reads one row ahead.
- **Boundary selectors** — no claim: the first-row, last-row, and transition selectors are structural, so the verifier evaluates them in closed form at `r`.

The next-row claims are driven by the AIR's declared next-row columns, so only the columns actually read ahead need a successor opening rather than every column. They open the same committed columns as the current-row claims, so no second commitment is required.

The verifier rebuilds the two rows the constraint evaluator reads — the current row from the `Eq` claims and the successor row from the `Next` claims, with unread columns left zero since the AIR never touches them — recomputes the alpha-batched constraint at `r` through the same evaluator the prover uses, and checks the result against the value the sumcheck reduced to.

## Scope

The opened values are trusted at this layer. Binding them to the commitment with a polynomial commitment scheme is a separate, self-contained step, so this change deliberately stops at the reduction: it verifies end to end except for the commitment opening. Isolating the reduction this way keeps it testable on its own, before the commitment machinery is involved.

## Robustness

Both opening counts are validated with typed errors:

- current-row openings against the number of columns,
- next-row openings against the number of columns the AIR reads ahead.

## Tests

`cargo test -p p3-multi-stark` — 22 passing, including:

- a satisfying Fibonacci trace proves and verifies through the reduction,
- corrupting a current-row opening is rejected,
- corrupting a next-row opening is rejected,
- an AIR that reads only one of its two columns ahead produces exactly one next-row claim and still verifies,
- unit coverage for reconstructing the successor row from the sparse next-row claims.

## References

- Setty, Thaler, Wahby — Customizable Constraint Systems for succinct arguments (eprint 2023/552)
- Borgeaud — AIR-specific optimizations on top of SuperSpartan

### DIFF
--- a/multi-stark/src/opening.rs
+++ b/multi-stark/src/opening.rs
@@ -18,6 +18,8 @@ use p3_multilinear_util::point::Point;
 ///
 /// A commitment scheme later opens this bundle against the trace commitment.
 /// Until then the values are trusted and feed only the verifier's constraint recompute.
+// TODO: produce this from the prover instead of only reconstructing it inside the verifier.
+// The commitment-opening step will consume it directly.
 #[derive(Clone, Debug)]
 pub struct OpeningClaims<EF> {
     /// Point `r` at which every claim is made: the bound sumcheck challenges.
--- a/multi-stark/src/opening.rs
+++ b/multi-stark/src/opening.rs
@@ -1,0 +1,131 @@
+//! Reduce a finished AIR zerocheck to evaluation claims about the committed columns.
+//!
+//! The sumcheck binds every variable to a single point `r`.
+//! Satisfying the AIR then reduces to a handful of evaluation claims at `r`:
+//!
+//! - every column's value at `r` — the current-row claims, opened as `Eq`,
+//! - the successor value of each column read ahead — the next-row claims, opened as `Next`.
+//!
+//! Boundary selectors carry no claim: the verifier evaluates them in closed form at `r`.
+//! The next-row claims open the same committed columns, so they need no second commitment.
+
+use alloc::vec::Vec;
+
+use p3_field::Field;
+use p3_multilinear_util::point::Point;
+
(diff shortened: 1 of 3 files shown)


multi-stark/src/opening.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use alloc::vec::Vec;
+
+use p3_field::Field;
+use p3_multilinear_util::point::Point;
+
+/// The evaluation claims an AIR zerocheck reduces to, all at one point.
+///
+/// A commitment scheme later opens this bundle against the trace commitment.
+/// Until then the values are trusted and feed only the verifier's constraint recompute.
```
Nashtare, line 20: This would need a TODO as this will need to be exposed (currently it isn't, solely reconstructed within `verify()`), unless you're aiming at tackling it right after this

Nashtare (approved):

multi-stark/src/opening.rs
tcoratger: Just added a TODO so that we have the marker and we can continue with incremental PRs.

PR #1888, opened 2026-06-21
perf(util): stripe parallel transpose along the longer dimension

wstran:
## What

The parallel transpose splits work by **input rows**. For a wide input (few rows, many columns) each thread then writes a thin column band scattered across the whole tall output, one short run per output row. On the prover LDE that output is millions of rows tall, so those writes stall on RFO and TLB traffic.

**Root cause:** `transpose_neon_4b_parallel` always splits the row range, regardless of aspect ratio.

**Fix:** split the longer dimension. For a wide input split the columns, so each thread owns a contiguous block of output rows. This matches the sequential recursive path, which already splits the longer dimension. The per-thread output regions stay disjoint either way, so no extra synchronization is needed, and the tall/square path is unchanged.

```text
before (row-striped): thread writes output cols [r0, r1), scattered
                      down the tall output with stride = height
after  (col-striped): thread writes output rows [c0, c1), one
                      contiguous block
```

## Numbers

Apple M3 Pro, BabyBear, `RecursiveDft`:

| benchmark | baseline | this PR | speedup |
|---|---|---|---|
| transpose `2^8 x 2^20` (criterion, p<0.05) | 83.5 ms | 54.0 ms | 1.55x |
| prove `post-transpose` span (log-19) | 1.74 s | 0.605 s | 2.88x |
| prove total (log-19) | 10.7 s | 9.57 s | 1.12x (-10.6%) |

Repro:

```sh
cargo bench -p p3-matrix --bench transpose_benchmark --features p3-util/parallel
```

## Correctness

The output is the same writes reordered across threads, so the transposed result is bit-identical. The added test checks both stripings past the parallel threshold against the naive reference, and a STARK proof built on top still verifies.

## Scope

Only the 4-byte NEON path changes; the 8-byte path has the same opportunity and can follow.

### DIFF
--- a/util/src/transpose/rectangular.rs
+++ b/util/src/transpose/rectangular.rs
@@ -331,31 +331,48 @@ unsafe fn transpose_neon_4b(input: *const u32, output: *mut u32, width: usize, h
     }
 }
 
-/// Parallel transpose for very large matrices (≥ `PARALLEL_THRESHOLD` elements).
+/// Parallel transpose for very large matrices (at least `PARALLEL_THRESHOLD` elements).
 ///
-/// Divides the matrix into horizontal stripes, one per thread.
-/// Each thread processes its stripe independently using the tiled algorithm.
+/// The work is split into stripes, one per thread, along the longer input dimension.
 ///
 /// # Stripe Division
 ///
+/// The output holds the transpose in row-major order.
+/// Input element `(r, c)` lands at output index `c * height + r`.
+///
+/// A wide input (more columns than rows) is split by columns.
+/// - Each thread owns a column band over every row.
+/// - Its writes form one contiguous block of output rows.
+///
+/// A tall or square input is split by rows.
+/// - Each thread owns a row band over every column.
+/// - Its input reads stay contiguous, one full row at a time.
+///
 /// ```text
-///     ┌─────────────────────────────────┐
-///     │         Thread 0                │  rows [0, rows_per_thread)
-///     ├─────────────────────────────────┤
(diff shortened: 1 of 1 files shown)


tcoratger:
@wstran Can you show the results of the full benchmark suite for the matrix transposition? In order to be sure that this modification didn't generate a slow down on other cases :)

wstran:
Thanks, good to check. Full `transpose_benchmark` suite, Apple M3 Pro, BabyBear (4-byte). The diff only adds the `width > height` branch, so only wide inputs change: tall and square keep the existing row-striping, and the sub-`PARALLEL_THRESHOLD` path is untouched.

Wide (the changed path), medians:

| shape (H x W) | base | this PR | speedup |
|---|---|---|---|
| 2^8 x 2^20 | 83.7 ms | 52.2 ms | 1.60x |
| 2^8 x 2^21 | 166 ms | 102 ms | 1.63x |
| 2^8 x 2^22 | 324 ms | 221 ms | 1.47x |
| 2^8 x 2^23 | 686 ms | 418 ms | 1.64x |

criterion flags all four as improved (p < 0.05).

Tall and square run the same code as before, so they are unaffected. The small shapes confirm it (2^4, 2^8, 2^10 all within 1.00-1.03x). The large tall/square shapes sit inside laptop run-to-run noise, so there is no signal to read there.

tcoratger (approved):

Nashtare (approved):
LGTM, I could reproduce the benchmarks locally

PR #1889, opened 2026-06-22
perf(util): stripe the 8-byte parallel transpose along the longer dimension

wstran:
## What

Follow-up to #1888, which striped the 4-byte NEON parallel transpose along the longer dimension. The 8-byte path (`transpose_neon_8b_parallel`) still split work by input rows only, so a wide input made each thread scatter its writes across the whole tall output (stride = `height`), stalling on RFO and TLB traffic.

Same fix: split the longer input dimension, so a wide input is striped by columns and each thread writes a contiguous block of output rows. Mirrors the 4-byte path; the tall/square path is unchanged.

## Numbers

Apple M3 Pro, Goldilocks (8-byte), criterion:

| benchmark | baseline | this PR | |
|---|---|---|---|
| `large_wide_4096x2048` | 3.6 ms | 2.2 ms | 1.6x (p < 0.05) |
| `large_tall_2048x4096` | 2.9 ms | 2.9 ms | no change (p = 0.57) |

repro: `cargo bench -p p3-util --bench transpose --features parallel`

## Correctness

Same writes reordered across threads, so the result is bit-identical. The parallel test now checks both the 4-byte and 8-byte paths against the naive reference (wide, tall, and degenerate two-row shapes past the parallel threshold).

### DIFF
--- a/util/src/transpose/rectangular.rs
+++ b/util/src/transpose/rectangular.rs
@@ -1258,7 +1258,46 @@ unsafe fn transpose_neon_8b(input: *const u64, output: *mut u64, width: usize, h
 
 /// Parallel transpose for very large matrices of 8-byte elements.
 ///
-/// Divides the matrix into horizontal stripes, one per thread.
+/// The work is split into stripes, one per thread, along the longer input dimension.
+///
+/// # Stripe Division
+///
+/// The output holds the transpose in row-major order.
+/// Input element `(r, c)` lands at output index `c * height + r`.
+///
+/// A wide input (more columns than rows) is split by columns.
+/// - Each thread owns a column band over every row.
+/// - Its writes form one contiguous block of output rows.
+///
+/// A tall or square input is split by rows.
+/// - Each thread owns a row band over every column.
+/// - Its input reads stay contiguous, one full row at a time.
+///
+/// ```text
+///     wide input: split by columns        tall input: split by rows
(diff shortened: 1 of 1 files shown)


wstran:
@tcoratger Full picture before you review: the column striping is a broad win for the 8-byte path, with one narrow cache-aliasing resonance worth flagging.

Wins (Goldilocks, baseline -> this PR):
- `4096x2048`: 3.7 ms -> 2.2 ms (1.6x)
- width 2^18: heights 16-48 and 128+ win ~1.1-1.4x
- width 2^20: heights 96-512 win ~1.1-1.65x (h512 238 ms -> 144 ms, h256 110 ms -> 70 ms)

Regression: at width 2^18, height ~64-96 regresses ~10-15%. It is a power-of-2 read-stride resonance in the column reads, and it is width-specific: the same height 96 wins at width 2^20 (38.5 ms -> 33.4 ms, 1.15x) but loses at width 2^18. The 4-byte path stays flat at these shapes, so only the 8-byte path hits it.

A plain height guard does not cover it cleanly since the resonant height tracks the width. Options: leave it as a broad win with this narrow inherent resonance, or process the parallel 8-byte stripe cache-obliviously (recurse instead of flat-tile) to damp it, which is more invasive. How would you prefer to handle it?

tcoratger (approved):

PR #1892, opened 2026-06-23
feat(multi-stark): prove AIRs end-to-end via WHIR commitment openings

tcoratger:
Continues the multilinear SuperSpartan-flavored AIR prover (after #1879, #1885, #1886). This is the step where the system stops trusting the prover's claimed column values and binds them to a real polynomial commitment, yielding the first complete multilinear AIR SNARK over WHIR.

## What it does

The AIR zerocheck reduces the `alpha`-batched constraint to a single claim at one bound point `r`:

```text
    sum_x eq(tau, x) * g(x) = 0      (g = alpha-batched AIR constraint)
```

Previously the verifier closed this check using column values handed over in the proof body. Now WHIR opens the committed trace columns at `r`, so the values closing the zerocheck are bound to the commitment instead of trusted.

```text
    1. commit(trace)        -> absorb commitment
    2. zerocheck reduction  -> bound point r, reduced sum
    3. open columns at r     -> values bound to the commitment
    4. recompute g at r      -> match the reduced sum
```

## Changes

- **Prescribed-point opening interface** on the multilinear PCS: open and verify at a caller-supplied point and return the opened values, for an outer protocol (the AIR sumcheck) that fixes the point itself rather than sampling it.
- **WHIR implementation** of that interface. WHIR still draws its own out-of-domain samples (they belong to its proximity argument), but the evaluation point comes from the sumcheck. The commitment is absorbed once, by the caller.
- **Split the zerocheck verifier** into a sumcheck-reduction step (yields the point and target sum) and a constraint-closing step (recomputes `g` from opened values), so the commitment opening slots in between.
- **Top-level `prove` / `verify`** over one shared transcript, mirroring the uni-stark public API.
- The trace columns are committed as one stacked WHIR table (arity `log_height + ceil(log2 width)`). Current-row columns open as plain claims; transition reads open the same columns through a repeat-last successor view (no second commitment).

## Soundness / tests

- End-to-end Fibonacci proof over WHIR (`commit -> zerocheck -> open -> verify`).
- Rejection tests asserting the exact error variant for: tampered opening value, violated transition constraint, and tampered public values.
- WHIR-level prescribed-point tests: round-trip, wrong verifier point, tampered claimed evaluation.

All touched crates green: `p3-multi-stark` (22 + 4), `p3-sumcheck` (237), `p3-whir` (131, 1 heavy-CI ignored); `fmt` and `clippy` clean.

## Out of scope (later in the series)

Preprocessed and periodic columns, public-values-as-committed-IO, multi-table batching, packed/SIMD zerocheck, low-degree split, SVO, ZK, and the logup+GKR lookup track.

### DIFF
--- a/multi-stark/Cargo.toml
+++ b/multi-stark/Cargo.toml
@@ -33,7 +33,11 @@ tracing.workspace = true
 [dev-dependencies]
 criterion.workspace = true
 p3-baby-bear = { path = "../baby-bear" }
+p3-dft.workspace = true
+p3-merkle-tree.workspace = true
 p3-poseidon2-air.workspace = true
+p3-symmetric.workspace = true
+p3-whir.workspace = true
 rand = { workspace = true }
 tracing-forest = { workspace = true, features = ["ansi", "smallvec"] }
 tracing-subscriber = { workspace = true, features = ["std", "env-filter"] }
--- a/multi-stark/src/commit.rs
+++ b/multi-stark/src/commit.rs
@@ -175,6 +175,11 @@ mod tests {
             &self.pcs
         }
 
+        fn min_num_variables(&self) -> usize {
+            // The mock commits columns directly, with no padding floor.
+            0
+        }
+
         fn build_witness(&self, columns: Vec<Poly<F>>) -> Vec<Poly<F>> {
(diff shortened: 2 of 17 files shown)


Nashtare (approved):
LGTM!

PR #1894, opened 2026-06-25
perf(mersenne-31): specialize PackedMersenne31Neon dot_product with deferred reduction

wstran:
## Root cause
`PackedMersenne31Neon` was the only Mersenne31 packing without a `dot_product` override. AVX2/AVX-512 specialize it (#1855) and `monty-31` NEON does too, but the M31 NEON path fell back to the generic per-multiply-then-reduce default. `binomial_extension` calls `R::dot_product` over the packed base, so this is on the M31 extension-field hot path.

## Fix
For `N >= 4`, use the single-accumulator form of the existing `coeffwise_dot_product` kernel: widening multiply-accumulates (`vmlal`) into u64 lanes with the Mersenne reduction deferred and folded every 3 products, then two final folds and a canonical reduce. `N == 4` (the QM31 case) accumulates all four products without an intermediate fold (`4 * P^2 < 2^64`). For `N < 4` it keeps the plain reduced-multiply sum, because NEON's `sqdmulh` multiply is cheap enough that the deferral setup is a net loss for tiny `N`.

## Correctness
`test_packed_field_dot_product_boundary` (wired, sweeps N=1..16) covers the overflow boundaries; all paths pass. Full `p3-mersenne-31` suite (227 tests) plus clippy and fmt are clean.

## Measured (M3 / NEON, criterion, p < 0.05)
vs the scalar default: `N = 2` no change, `N >= 4` faster 9.5-32%. The `N = 4` no-fold path is a further ~14% over an every-3 fold schedule. Measured on NEON only; AVX2/AVX-512 already had the specialization.

## Scope
`mersenne-31/src/aarch64_neon/packing.rs` only.

### DIFF
--- a/mersenne-31/src/aarch64_neon/packing.rs
+++ b/mersenne-31/src/aarch64_neon/packing.rs
@@ -201,6 +201,54 @@ impl PrimeCharacteristicRing for PackedMersenne31Neon {
         // SAFETY: this is a repr(transparent) wrapper around an array.
         unsafe { reconstitute_from_base(Mersenne31::zero_vec(len * WIDTH)) }
     }
+
+    #[inline]
+    fn dot_product<const N: usize>(u: &[Self; N], v: &[Self; N]) -> Self {
+        // For small `N` the deferred-reduction setup (two final folds plus the narrowing)
+        // is not amortized, and NEON's `sqdmulh`-based reduced multiply is cheap enough that
+        // a plain reduced-multiply sum wins. Measured on NEON: deferral regresses ~10% at
+        // `N = 2` but improves 10-32% from `N = 4` upward.
+        if N < 4 {
+            return u.iter().zip(v).map(|(&x, &y)| x * y).sum();
+        }
+
+        // Single-accumulator form of the `coeffwise_dot_product` scheme: widen each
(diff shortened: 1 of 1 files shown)


tcoratger (commented):
Thanks a lot for this, the optimization looks great! 

Small remark to improve performances even more possibly. Are you sure that this is the best way to do things here to have this single generic block? Just saying that because for other fields like monty31, we are doing special cases with a match to obtain even more speedups? Doesn't it apply here as well?

https://github.com/Plonky3/Plonky3/blob/41c54cd6722b032bc467d653cea9c17572418c5c/monty-31/src/monty_31.rs#L269-L390

wstran:
Good catch. The win is `N = 4` (QM31): the four products fit without an intermediate fold (`4 * P^2 < 2^64`), so I special-cased it, ~14% over every-3 (p<0.05). I tried a general per-N schedule but it regressed `N = 5` by ~4% (M31's cheap `sqdmulh` doesn't amortize the extra bookkeeping when the fold count is unchanged), so every-3 stays for `N >= 5` and a plain sum for `N < 4`. Happy to switch to explicit kernels if you prefer that style.

tcoratger (approved):

PR #1895, opened 2026-06-25
perf(multi-stark): parallelize the zerocheck round_poly hot loop

Nashtare:
Also added a benchmark file for it. 

On my M4 pro:

```bash
  ┌────────────┬─────────┬──────────┬─────────┐
  │ log₂(rows) │ serial  │ parallel │ speedup │
  ├────────────┼─────────┼──────────┼─────────┤
  │ 16         │ 20.7 ms │ 10.1 ms  │ 2.0×    │
  ├────────────┼─────────┼──────────┼─────────┤
  │ 18         │ 83.3 ms │ 20.7 ms  │ 4.0×    │
  ├────────────┼─────────┼──────────┼─────────┤
  │ 20         │ 338 ms  │ 53.3 ms  │ 6.3×    │
  └────────────┴─────────┴──────────┴─────────┘
  ```

PR #1896, opened 2026-06-25
perf(multi-stark): prune the zerocheck successor table to declared next-row columns

Nashtare:
### Summary

  The zerocheck prover built and folded a full next-row (successor) table for **every** trace column, even though an AIR only reads the columns reported by `BaseAir::main_next_row_columns()` on the next row. `RoundState` now
  builds, folds, and fills successor tables only for those declared columns; undeclared `next_row` scratch entries stay at their zero init, and the AIR provably never reads them.

  This stacks on top of the `round_poly` parallelization from #1895 — it trims a constant factor off the same hot loop, plus the `RoundState::new` build and the per-round `fold`.

  ### Correctness

  - **Proof output is byte-identical.** `proof.next` carries the same values in the same declared order; nothing in the transcript or verifier changes. This is a pure prover-internal optimization.
  - The existing end-to-end test `next_claims_cover_only_declared_columns` (a width-2 AIR declaring only column 0, leaving column 1's next-row zero-filled) is the correctness guard and passes — along with all 22 crate tests.
  `clippy` is clean.

  ### Performance

  Measured drift-free by comparing two declared-footprint settings of one width-64 AIR (2 columns read on the next row) **within the same binary**, interleaved by criterion so machine drift cancels. `all_next` reproduces the
  pre-change cost (declares all 64 columns); `sparse_next` is the optimized path. Both share an identical `Air::eval`.

  | log₂(rows) | `all_next` (pre-change) | `sparse_next` (optimized) | Speedup |
  |---:|---:|---:|---:|
  | 16 | 204.8 ms | 116.3 ms | **1.76×** (−43%) |
  | 18 | 912.1 ms | 523.0 ms | **1.74×** (−43%) |
  | 20 | 4.876 s | 2.146 s | **2.27×** (−56%) |

  The win grows with trace size as the all-column successor build/fold becomes memory-bandwidth bound.

  **Scope:** the speedup is proportional to next-row sparsity. An AIR that reads all columns on the next row — including any AIR that does not override `main_next_row_columns()` (the default declares all columns) — gets
  **0%**; the change is provably a no-op there. The benefit is real and large precisely for wide AIRs with a sparse next-row footprint.

  ### Changes

  - `multi-stark/src/zerocheck.rs` — `RoundState` carries `next_columns` and builds/folds/fills successors only for those; prover reads the next-row openings straight off the pruned tables.
  - `multi-stark/benches/zerocheck.rs` — added a wide AIR with `all_next` / `sparse_next` variants as a reproducible measurement and regression guard.

  ### Test plan

  - [x] `cargo test -p p3-multi-stark` — 22 passed
  - [x] `cargo clippy -p p3-multi-stark --benches --tests` — clean
  - [x] `cargo bench -p p3-multi-stark --bench zerocheck -- zerocheck_prove_wide` — table above

### DIFF
--- a/multi-stark/benches/zerocheck.rs
+++ b/multi-stark/benches/zerocheck.rs
@@ -18,6 +18,9 @@ type Ch = DuplexChallenger<F, Perm, 16, 8>;
 
 const NUM_COLS: usize = 2;
 
+/// Width of the wide trace whose columns are mostly current-row-only filler.
+const WIDE_COLS: usize = 64;
+
 struct FibAir;
 struct FibRow<T> {
     left: T,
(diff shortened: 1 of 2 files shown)


tcoratger (approved):
Nice!

Issue #1890, opened 2026-06-22
Evaluate projective (monomial-basis) sum-check (eprint 2026/762) on the WHIR prover


jdodinh:
### Motivation

The WHIR prover runs its folding sum-check in the evaluation (Lagrange) basis. The dominant cost of linear-time sum-check is the per-round **binding** step, which updates every cached value with the verifier challenge `r`. The current kernel interpolates between two stored evaluations, costing one subtraction per value:

```
p(r, x') = p(0, x') + r * (p(1, x') - p(0, x'))   // multilinear-util/src/poly.rs
```

This binding path is shared: besides the WHIR prover, the recently added multi-stark AIR zerocheck (#1879, #1885) binds through the same `p3-multilinear-util` kernels (`fix_prefix_var_mut` and the pointwise `fix_prefix_var_at`), so a subtraction-free binding would benefit both consumers.

This repo already sends round messages as `[h(0), h(inf)]` and has the verifier derive `h(1) = claimed_sum - h(0)` (`sumcheck/src/{lagrange,data}.rs`). That is the eval-at-infinity round message from "Speeding Up Sum-Check Proving" (eprint 2026/587), which puts the point at infinity into the round **message** as an extrapolation point.

"The Sum-Check Protocol over the Monomial Basis, and Other Optimizations" (https://eprint.iacr.org/2026/762, Dao-Biswas-Eagen-Milson-Papini-Thaler) takes the next step: it moves the point at infinity into the **interpolating set** itself, summing over the projective hypercube `{0, inf}^n` instead of `{0, 1}^n`. By Corollary 3.2 the binding step then becomes subtraction-free:

```
p(r, x') = p(0, x') + r * p(inf, x')
```

because `p(inf, x')` is the leading coefficient (the slope), which is stored directly in the monomial representation rather than recovered by subtraction. Proposition 4.1 quantifies this as `d * (2^n - 1)` field subtractions saved across all rounds for a product of `d` multilinears. Soundness is unchanged: Theorem 3.3 gives the same per-round error `d/|F|` (total `d*n/|F|`) as standard sum-check.

[...]

PR #1900, opened 2026-06-26
feat(sumcheck): subtraction-free projective (monomial-basis) sum-check on the prover (eprint 2026/762)

jdodinh:
Closes part of #1890.

> **Edit (2026-06-26):** reworked per @tcoratger's review (thanks for the careful read). Tightened the PR to the two subtraction-free kernels and dropped everything the kernels do not yet need:
> - Removed the `Basis { Eval, Projective }` dispatch on `ProductPolynomial`; nothing routes to the projective kernels yet, so there is no second basis to keep alive. `ProductPolynomial` is back to upstream verbatim.
> - Removed the full sum-check prover bench. As flagged in review, `ProductPolynomial::round()` never read `self.basis`, so that bench timed eval coefficients + projective binding + eval reduction, not a coherent projective prover. The ~5-6% "end-to-end" row is retracted.
> - Removed `tests/projective_equivalence.rs`. It re-implemented the protocol and checked itself, so it would pass even with the production kernel deleted. The end-to-end equivalence/soundness checks move to the §4.3 follow-up where the protocol actually runs in the projective basis.
> - The binding unit test is now a proptest.
> - The projective and eval round-coefficient kernels now share one tiling scaffold (`sumcheck_coefficients_prefix_with`) instead of duplicating it. This is its own behaviour-preserving `refactor` commit, so the one change to the eval path is isolated and reviewable on its own.
>
> What lands here is the two kernels, each proptested against an independent reference, plus the round-coefficient bench. Fresh single-threaded numbers below. History was rewritten (5 commits -> 4), so the diff reflects the tightened scope directly.

## What

Adds the two subtraction-free prover kernels from "The Sum-Check Protocol over the Monomial Basis" (eprint 2026/762). Running sum-check with the interpolating set `{0, inf}^n` instead of `{0, 1}^n` makes both per-round operations subtraction-free:

- **Binding** (`multilinear-util`): `a0 += (a1 - a0) * r` becomes `a0 += a1 * r` (`Poly::fix_prefix_var_mut_projective`).
-  **Round coefficients** (`sumcheck`): the monomial-basis round message is `[s(1), s(inf)]` with `s(1) = dot(lo+hi, lo+hi)` and `s(inf) = dot(hi, hi)`; the verifier derives `s(0)` from the projective round identity `s(0) + s(inf) = C` (eprint 2026/762, Fig. 3). Both bases share one `K`-tiled delayed-reduction scaffold (`sumcheck_coefficients_prefix_with`), parameterised over the per-tile and per-pair MAC.

This PR adds the kernels and their tests/bench. Nothing dispatches to the projective kernels yet, so the default prover and WHIR are unchanged in behaviour. The one change to the existing path is the scaffold extraction in the `refactor` commit, which is behaviour-preserving (output identical, full sumcheck suite green) and isolated for review. The wiring that makes the kernels load-bearing (threading the projective representation through the prover and the WHIR commitment, the paper's §4.3 alignment) is the follow-up below.

## Scope

In:
- Projective binding + round-coefficient kernels, prover-internal, **prefix order only** (WHIR's path).
- A proptest per kernel against an independent reference, and a before/after bench on the round-coefficient kernel.

Not in (deliberately):
- No dispatch, no `ProductPolynomial` change, no verifier/transcript/proof-format change. The kernels are added alongside the existing ones; the default path's behaviour does not move (the eval kernel is refactored to share the scaffold, output identical).
- Suffix-order projective kernels.
- The end-to-end change that runs the prover in one representation throughout and aligns the WHIR commitment basis (§4.3). That is where these kernels become live, and it is a separate follow-up.

## Correctness

Each kernel is proptested against an independent reference over random instances:

- **Binding** (`multilinear-util`): `fix_prefix_var_mut_projective` applied across every variable evaluates the monomial polynomial `sum_S a_S * prod_{i in S} z_i` at a random point, for random variable counts.
- **Round coefficients** (`sumcheck`): `sumcheck_coefficients_prefix_projective` matches a direct reference computation of `(h(0), h(inf))` over random tables and weights.

Why the basis change is sound, for context (not exercised by this PR, which ships only the kernels): by Proposition 3.1 the projective `{0, inf}`-sum and the standard `{0, 1}`-sum are the same dot product of the same table, so the value being proved is unchanged; Theorem 3.3 keeps the round-by-round soundness error at `d/|F|`. The end-to-end equivalence and soundness checks belong with the §4.3 follow-up, where the full protocol actually runs in the projective basis.

## Numbers

Apple M3 Pro MacBook Pro (12-core, `aarch64-apple-darwin`), `cargo bench`, **single-threaded**, `-Ctarget-cpu=native`. Single-threaded matches the paper's methodology (§6.1: gains are reported for a single performance core).

Reproduce (requires Rust >= 1.93; leave the `parallel` feature off; the filter catches both the eval and `_projective` variants):

```bash
RUSTFLAGS="-Ctarget-cpu=native" cargo +1.93.0 bench -p p3-sumcheck --bench sumcheck \
  -- "round_coefficients/ext_ext_packed_prefix" --warm-up-time 5 --measurement-time 10
```

### Round-coefficient kernel (`p3-sumcheck`, packed extension, prefix)

The complete projective message spends an addition pass where the evaluation kernel spends a subtraction pass, so parity is the expected and measured outcome (speedup = eval / projective):

| Size | BabyBear | KoalaBear |
|---|---|---|
| k16 | 0.99x | 0.98x |
| k18 | 0.98x | 0.98x |
| k20 | 0.99x | 0.99x |
| k22 | 0.98x | 0.99x |

The subtraction-free **binding** kernel is where the per-round win lives; it is measured standalone in the follow-up PR (a consistent ~1.13x).

## Honest caveats

- **Single-threaded only.** The kernel is memory-bandwidth-bound once parallelised, where the subtraction saving largely hides. This matches the paper, which also reports single-threaded gains and leaves multi-threaded to future work.
- **Kernel parity is the expected outcome on any field.** The complete projective message spends an addition pass where the evaluation kernel spends a subtraction pass. The basis' real per-round saving is the subtraction-free binding, measured standalone in the follow-up PR (#1949) at a consistent ~1.13x.

## Follow-up

These kernels are the prover-side primitives. The change that makes them load-bearing runs the whole prover in the projective basis and aligns the WHIR commitment basis (§4.3), so the prover holds one representation throughout instead of converting. That is where end-to-end equivalence and soundness get exercised and where the full-prover win shows up. Landing the kernels first keeps that change small and reviewable. Happy to keep this as a draft until that lands if you would rather review them together.

## Test plan

- [x] `cargo test -p p3-sumcheck`: full suite passes (round-coefficient kernel proptest included)
- [x] `cargo test -p p3-multilinear-util`: passes (binding kernel evaluates the monomial polynomial)
- [x] `cargo check -p p3-whir`: no change (kernels are not yet wired in)
- [x] `cargo clippy -p p3-sumcheck -p p3-multilinear-util --all-targets`: clean
- [x] `cargo fmt --all --check`: clean

### DIFF
--- a/multilinear-util/src/poly/mod.rs
+++ b/multilinear-util/src/poly/mod.rs
@@ -492,6 +492,51 @@ impl<A: Copy + Send + Sync + PrimeCharacteristicRing> Poly<A> {
         self.0.truncate(mid);
     }
 
+    /// Fixes the prefix variable at a challenge value in the *monomial* basis,
+    /// in place.
+    ///
+    /// The table is interpreted as monomial coefficients (the projective
+    /// `{0, inf}` representation of eprint 2026/762). Writing the polynomial as
+    /// `p(X_0, x') = a0(x') + X_0 * a1(x')`, with `a0` the low half (monomials
+    /// without `X_0`) and `a1` the high half (monomials with `X_0`), binding
+    /// `X_0 = r` gives, by Corollary 3.2:
+    ///
+    /// ```text
+    /// out = p(0, x') + r * p(inf, x') = a0 + r * a1
+    /// ```
+    ///
+    /// Unlike [`Self::fix_prefix_var_mut`], this costs no subtraction per pair.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the polynomial is constant (zero free variables).
+    pub fn fix_prefix_var_mut_projective<F: Copy + Send + Sync>(&mut self, r: F)
+    where
+        A: Algebra<F>,
+    {
+        assert!(self.as_constant().is_none(), "no free variables");
+        let num_evals = self.num_evals();
+        let mid = num_evals / 2;
+        // Low half: monomials without x_0 (the constant part a0).
+        // High half: monomials with x_0 (the leading-coefficient part a1).
+        let (p0, p1) = self.0.split_at_mut(mid);
+
+        if num_evals >= PARALLEL_THRESHOLD {
+            // Parallel: fold each pair in place.
+            p0.par_iter_mut()
+                .zip(p1.par_iter())
+                .for_each(|(a0, &a1)| *a0 += a1 * r);
+        } else {
+            // Sequential: fold each pair in place.
+            p0.iter_mut()
+                .zip(p1.iter())
+                .for_each(|(a0, &a1)| *a0 += a1 * r);
+        }
+
+        // Discard the second half; the first half now holds the folded result.
+        self.0.truncate(mid);
+    }
+
     /// In-place version of the suffix-variable fix.
     ///
     /// Folds adjacent pairs and truncates to the first half. The sequential
@@ -1973,6 +2018,46 @@ pub(crate) mod test {
             for s in 0..folded.num_evals() {
                 assert_eq!(poly.fix_prefix_var_at(z, s), folded.as_slice()[s]);
             }
+        }
+    }
+
+    proptest! {
+        /// Projective binding fixes a variable in the *monomial* basis: the table
+        /// holds monomial coefficients (Proposition 3.1, eprint 2026/762), and
+        /// binding x_0 = r maps each pair (a0, a1) to a0 + r * a1, with no
+        /// subtraction. Applied across every variable it evaluates the monomial
+        /// polynomial sum_S a_S * prod_{i in S} z_i at the point z.
+        #[test]
+        fn prop_fix_prefix_var_mut_projective_evaluates_monomial_basis(
+            k in 1usize..=12,
+            seed in any::<u64>(),
+        ) {
+            let mut rng = SmallRng::seed_from_u64(seed);
+            // Interpret the random table as monomial coefficients.
+            let poly = Poly::<F>::rand(&mut rng, k);
+            let point: Point<F> = Point::rand(&mut rng, k);
+            let z = point.as_slice();
+
+            // Reference: evaluate the monomial polynomial directly. Prefix binding
+            // fixes the most-significant variable first, so bit position p of the
+            // index corresponds to challenge z[k - 1 - p].
+            let mut expected = F::ZERO;
+            for (s, &coeff) in poly.as_slice().iter().enumerate() {
+                let mut term = coeff;
+                for p in 0..k {
+                    if (s >> p) & 1 == 1 {
+                        term *= z[k - 1 - p];
+                    }
+                }
+                expected += term;
+            }
+
+            // Projective binding across all variables must agree.
+            let mut compressed = Poly::new(poly.as_slice().to_vec());
+            for &zi in z {
+                compressed.fix_prefix_var_mut_projective(zi);
+            }
+            prop_assert_eq!(compressed.as_constant().unwrap(), expected);
         }
     }
 
--- a/multilinear-util/src/poly/mod.rs
+++ b/multilinear-util/src/poly/mod.rs
@@ -496,7 +496,7 @@ impl<A: Copy + Send + Sync + PrimeCharacteristicRing> Poly<A> {
     /// in place.
     ///
     /// The table is interpreted as monomial coefficients (the projective
-    /// `{0, inf}` representation of eprint 2026/762). Writing the polynomial as
+    /// `{0, inf}` representation of <https://eprint.iacr.org/2026/762> ). Writing the polynomial as
     /// `p(X_0, x') = a0(x') + X_0 * a1(x')`, with `a0` the low half (monomials
     /// without `X_0`) and `a1` the high half (monomials with `X_0`), binding
     /// `X_0 = r` gives, by Corollary 3.2:
@@ -510,7 +510,7 @@ impl<A: Copy + Send + Sync + PrimeCharacteristicRing> Poly<A> {
     /// # Panics
     ///
     /// Panics if the polynomial is constant (zero free variables).
-    pub fn fix_prefix_var_mut_projective<F: Copy + Send + Sync>(&mut self, r: F)
+    pub fn fix_prefix_var_mut_monomial<F: Copy + Send + Sync>(&mut self, r: F)
     where
         A: Algebra<F>,
     {
@@ -2028,7 +2028,7 @@ pub(crate) mod test {
         /// subtraction. Applied across every variable it evaluates the monomial
         /// polynomial sum_S a_S * prod_{i in S} z_i at the point z.
         #[test]
-        fn prop_fix_prefix_var_mut_projective_evaluates_monomial_basis(
+        fn prop_fix_prefix_var_mut_monomial_evaluates_monomial_basis(
             k in 1usize..=12,
             seed in any::<u64>(),
         ) {
@@ -2055,7 +2055,7 @@ pub(crate) mod test {
             // Projective binding across all variables must agree.
             let mut compressed = Poly::new(poly.as_slice().to_vec());
             for &zi in z {
-                compressed.fix_prefix_var_mut_projective(zi);
+                compressed.fix_prefix_var_mut_monomial(zi);
             }
             prop_assert_eq!(compressed.as_constant().unwrap(), expected);
         }
--- a/sumcheck/benches/sumcheck.rs
+++ b/sumcheck/benches/sumcheck.rs
@@ -32,7 +32,8 @@ use p3_sumcheck::constraints::{Constraint, Statements};
 use p3_sumcheck::layout::{Layout, PrefixProver, SuffixProver, Table};
 use p3_sumcheck::product_polynomial::ProductPolynomial;
 use p3_sumcheck::strategy::{
-    SumcheckProver, VariableOrder, sumcheck_coefficients_prefix, sumcheck_coefficients_suffix,
+    SumcheckProver, VariableOrder, sumcheck_coefficients_prefix,
+    sumcheck_coefficients_prefix_projective, sumcheck_coefficients_suffix,
 };
 use p3_sumcheck::{OpeningBatch, SumcheckData};
 use rand::distr::{Distribution, StandardUniform};
@@ -251,6 +252,19 @@ where
             );
         }
 
+        group.bench_with_input(
+            BenchmarkId::new("base_ext_prefix_projective", &label),
+            &label,
+            |b, _| {
+                b.iter(|| {
+                    black_box(sumcheck_coefficients_prefix_projective(
+                        &base_packed,
+                        round_zero_weights.as_slice(),
+                    ))
+                });
+            },
+        );
+
         // Later-round operands: both sides are packed extension elements.
         let packed_evals = rand_ext_packed::<B>(&mut rng, k);
         let packed_weights = rand_ext_packed::<B>(&mut rng, k);
@@ -269,6 +283,19 @@ where
                 },
             );
         }
+
+        group.bench_with_input(
+            BenchmarkId::new("ext_ext_packed_prefix_projective", &label),
+            &label,
+            |b, _| {
+                b.iter(|| {
+                    black_box(sumcheck_coefficients_prefix_projective(
+                        packed_evals.as_slice(),
+                        packed_weights.as_slice(),
+                    ))
+                });
+            },
+        );
     }
 
     for &k in SCALAR_SIZES {
@@ -288,6 +315,19 @@ where
                 },
             );
         }
+
+        group.bench_with_input(
+            BenchmarkId::new("ext_ext_scalar_prefix_projective", &label),
+            &label,
+            |b, _| {
+                b.iter(|| {
+                    black_box(sumcheck_coefficients_prefix_projective(
+                        evals.as_slice(),
+                        weights.as_slice(),
+                    ))
+                });
+            },
+        );
     }
 
     group.finish();
--- a/sumcheck/benches/sumcheck.rs
+++ b/sumcheck/benches/sumcheck.rs
@@ -366,6 +366,27 @@ where
                 );
             });
         }
+
+        // Projective (monomial-basis) binding: the subtraction-free
+        // `a0 + a1 * r` of eprint 2026/762, prefix only.
+        //
+        // This is where the projective basis is expected to pay. The
+        // round-coefficient kernel only trades a subtraction pass for an
+        // addition pass, so it cannot come out ahead; binding drops a pass.
+        group.bench_with_input(
+            BenchmarkId::new("prefix_projective", &label),
+            &(),
+            |b, ()| {
+                b.iter_batched_ref(
+                    || template.clone(),
+                    |poly| {
+                        poly.fix_prefix_var_mut_monomial(black_box(r));
+                        black_box(&*poly);
+                    },
+                    BatchSize::LargeInput,
+                );
+            },
+        );
     }
 
     group.finish();
--- a/sumcheck/benches/sumcheck.rs
+++ b/sumcheck/benches/sumcheck.rs
@@ -32,7 +32,7 @@ use p3_sumcheck::constraints::{Constraint, Statements};
 use p3_sumcheck::layout::{Layout, PrefixProver, SuffixProver, Table};
 use p3_sumcheck::product_polynomial::ProductPolynomial;
 use p3_sumcheck::strategy::{
-    SumcheckProver, VariableOrder, sumcheck_coefficients_prefix,
+    RoundMessage, SumcheckProver, VariableOrder, sumcheck_coefficients_prefix,
     sumcheck_coefficients_prefix_projective, sumcheck_coefficients_suffix,
 };
 use p3_sumcheck::{OpeningBatch, SumcheckData};
@@ -205,7 +205,7 @@ where
 ///
 /// The constant term and the leading coefficient of the round polynomial.
 #[inline]
-fn coeffs<Base, Acc>(order: VariableOrder, evals: &[Base], weights: &[Acc]) -> (Acc, Acc)
+fn coeffs<Base, Acc>(order: VariableOrder, evals: &[Base], weights: &[Acc]) -> RoundMessage<Acc>
 where
     Base: PrimeCharacteristicRing + Copy + Send + Sync,
     Acc: Algebra<Base> + Copy + Send + Sync,
--- a/sumcheck/src/product_polynomial.rs
+++ b/sumcheck/src/product_polynomial.rs
@@ -27,9 +27,9 @@ use p3_multilinear_util::poly::{Poly, PolyMaybePacked, PolyMaybePackedView};
 use p3_util::log2_strict_usize;
 use tracing::instrument;
 
+use crate::SumcheckData;
 use crate::constraints::Constraint;
-use crate::strategy::{RoundMessage, VariableOrder};
-use crate::{SumcheckData, extrapolate_01inf};
+use crate::strategy::{Basis, RoundMessage, VariableOrder};
 
 /// A paired representation of evaluation and weight polynomials for quadratic sumcheck.
 ///
@@ -339,10 +339,14 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
         //
         // The strategy differs based on representation to maximize SIMD utilization.
         let order = self.order;
-        let RoundMessage { c_a: c0, c_inf } = match &self.inner {
+
+        // This prover binds evaluation-basis tables; the projective kernels
+        // are reachable through the same tag, wired up by their consumer.
+        let basis = Basis::Evaluation;
+        let RoundMessage { c_a, c_inf } = match &self.inner {
             MaybePacked::Packed { evals, weights } => {
                 // Packed round coefficients: SIMD-parallel per-lane accumulation.
-                let msg = order.sumcheck_coefficients(evals.as_slice(), weights.as_slice());
+                let msg = basis.sumcheck_coefficients(order, evals.as_slice(), weights.as_slice());
 
                 // Horizontal reduction across SIMD lanes.
                 RoundMessage {
@@ -352,21 +356,19 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
             }
             MaybePacked::Unpacked { evals, weights } => {
                 // Scalar round coefficients.
-                order.sumcheck_coefficients(evals.as_slice(), weights.as_slice())
+                basis.sumcheck_coefficients(order, evals.as_slice(), weights.as_slice())
             }
         };
 
         // Step 2-4: Commit to transcript, do PoW, and receive challenge.
-        let r = sumcheck_data.observe_and_sample(challenger, c0, c_inf, pow_bits);
+        let r = sumcheck_data.observe_and_sample(challenger, c_a, c_inf, pow_bits);
 
         // Step 5: Fold both polynomials using the challenge.
         self.compress(r);
 
-        // Step 6: Update the claimed sum.
-        //
-        // h(r) = h(0)*(1-r) + h(1)*r + h(inf)*r*(r-1)
-        // where h(1) = claimed_sum - h(0).
-        *sum = extrapolate_01inf(c0, *sum - c0, c_inf, r);
+        // Step 6: Update the claimed sum, through the same round identity the
+        // verifier will apply to this message.
+        *sum = basis.reduce_claim(c_a, c_inf, r, *sum);
 
         // Sanity check: the updated sum should equal the inner product after folding.
         debug_assert_eq!(*sum, self.dot_product());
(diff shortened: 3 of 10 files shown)


sumcheck/src/strategy.rs
```diff
@@ -173,6 +209,85 @@ where
...
+    let tail = e_lo_tail
+        .iter()
+        .zip(e_hi_tail.iter())
+        .zip(w_lo_tail.iter().zip(w_hi_tail.iter()))
+        .fold((A::ZERO, A::ZERO), |acc, ((&e0, &e1), (&w0, &w1))| {
+            round_step_projective(acc, e0, e1, w0, w1)
+        });
+
+    round_reduce(main, tail)
+}
```
tcoratger, line 301: I think this is easier to have this method directly in `Poly` instead of an isolated method.

multilinear-util/src/poly.rs
```diff
@@ -1905,6 +1950,43 @@ pub(crate) mod test {
...
+            }
+
+            // Projective binding across all variables must agree.
+            let mut compressed = Poly::new(poly.as_slice().to_vec());
+            for &zi in z {
+                compressed.fix_prefix_var_mut_projective(zi);
+            }
+            assert_eq!(compressed.as_constant().unwrap(), expected);
+        }
+    }
```
tcoratger: I think this is better to have a proptest here no?

sumcheck/tests/projective_equivalence.rs
```diff
@@ -0,0 +1,166 @@
...
+use proptest::prelude::*;
+use rand::rngs::SmallRng;
+use rand::{RngExt, SeedableRng};
+
+type F = BabyBear;
+
+/// Scalar dot product over the coefficient tables (test reference).
+fn dot(a: &[F], b: &[F]) -> F {
+    a.iter().zip(b).map(|(&x, &y)| x * y).sum()
+}
```
tcoratger: I think that we don't need this since we already have `dot_product` method which exists in the codebase.

sumcheck/benches/sumcheck.rs
```diff
@@ -491,6 +531,34 @@ where
                 },
             );
         }
+
+        // Projective full fold (prefix-only): the same multi-round driver over a
```
tcoratger: **This bench doesn't measure a projective prover, and I think it's the main thing to fix.**

The driver underneath, `ProductPolynomial::round()`, never reads `self.basis` — only `compress()` and `round_coefficients()` do. So each round here actually runs:

- **eval** round-coefficients (`dot(hi - lo, hi - lo)`, *with* the subtractions), then
- **projective** binding (`a0 += r * a1`), then
- the **eval** sum reduction (`extrapolate_01inf`).

Two consequences:

1. The projective round kernel — the thing this PR is selling — *never runs* in this bench. The ~5-6% "full prover" number comes only from binding, while still paying for the eval kernel's subtractions, so it understates and mislabels at the same time.
2. eval-coeffs + projective-fold + eval-reduction don't satisfy `sum == dot_product`. The `debug_assert_eq!` in `round()` would catch it instantly — Criterion just builds in release, so the incoherence is silent.

The "reduced sum values are not meaningful" caveat in the description undersells this: it isn't that the values are uninteresting, it's that the path isn't any well-defined protocol.

**Suggested fix:** either make `round()` basis-aware *and* give the projective path its own sum reduction (the §4.3 work — projective identity is `h(0) + h(inf) = claim`, reduced via `claim ← h_inf·r² + (h1 − h0 − h_inf)·r + h0`, needing `h(1) = dot(lo+hi, lo+hi)` too), or drop this bench until then. As-is I'd retract the full-prover row; the isolated round-kernel row (~11%) is sound and measures the real thing.

sumcheck/tests/projective_equivalence.rs
```diff
@@ -0,0 +1,166 @@
+//! Self-contained equivalence prototype for projective (monomial-basis)
```
tcoratger: **I'd drop this whole file** — it doesn't test the production code it claims to.

The file imports nothing from `p3_sumcheck::strategy`. It defines its own `dot`, its own `projective_round_coefficients` (line 38), and its own prover+verifier, then checks that *that local toy* is self-consistent. So:

- `projective_sumcheck_maintains_round_identity` and the soundness proptest never call `sumcheck_coefficients_prefix_projective`. **Delete the production kernel and this file still passes** — as a regression test it's vacuous.
- The "soundness" proptest reduces to: perturb `h0` by `delta != 0` ⇒ `h0 + h_inf != claim` at round 0 ⇒ the *file-local* verifier's equality check returns false. That's testing that `a + delta != a` for `delta != 0` against an equality check defined a few lines up. The library has no projective verifier, so Theorem 3.3 isn't being exercised by anything in the repo.
- The only production symbol it touches, `fix_prefix_var_mut_projective`, is already covered by the unit test in `poly.rs`.

The coverage that matters already lives in the right places: round kernel vs reference in `strategy.rs` (`prop_sumcheck_coefficients_prefix_projective_matches_reference`), binding in `poly.rs`. A genuine end-to-end equivalence test is worth having, but it has to drive the *real* prover — which is blocked on the `round()` wiring above. Until then this gives false confidence. (This also moots the `dot` helper comment below.)

sumcheck/src/product_polynomial.rs
```diff
@@ -171,12 +171,26 @@ impl<F: Field, EF: ExtensionField<F>> PolyView<'_, F, EF> {
 ///
 /// The tag is consulted once per round; the per-round work is `O(2^n)`, so
 /// the branch is immeasurable next to the inner loops.
+/// Whether the polynomial tables are read in the evaluation (Boolean hypercube)
+/// basis or the projective monomial (`{0, inf}`) basis of eprint 2026/762.
```
tcoratger: Small doc-placement casualty: inserting `enum Basis` here split the `ProductPolynomial` doc block. The `// The tag is consulted once per round ...` paragraph (which is about `ProductPolynomial`'s `VariableOrder` dispatch) is now fused onto `Basis`, and `pub struct ProductPolynomial` a few lines down ends up with **no doc comment at all**.

**Fix:** move `Basis` (with just its own two-line doc) *above* the `ProductPolynomial` doc block, so the struct keeps its description and `Basis` doesn't inherit paragraphs about `VariableOrder` that don't apply to it.

sumcheck/src/product_polynomial.rs
```diff
@@ -221,9 +236,17 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
         Self {
             inner: MaybePacked::Unpacked { evals, weights },
             order,
+            basis: Basis::Eval,
         }
     }
 
+    /// Selects the table basis (builder style); defaults to [`Basis::Eval`].
+    #[must_use]
+    pub const fn with_basis(mut self, basis: Basis) -> Self {
```
tcoratger: `Basis::Projective` is documented as "Prefix only," but nothing enforces it: the projective arms in `compress`/`round_coefficients` call the **prefix** kernels unconditionally, ignoring `self.order`. So `…new(Suffix…).with_basis(Projective)` silently runs prefix math on suffix-laid-out data — wrong result, no warning.

An invariant in prose but not in code is a latent bug. Cheap fix: `assert!(self.order == VariableOrder::Prefix)` in `with_basis` (or a `debug_assert!` in the projective dispatch arms). That turns a documented footgun into immediate, loud failure.

sumcheck/src/strategy.rs
```diff
@@ -173,6 +209,85 @@ where
...
+/// (eprint 2026/762, Cor 3.2).
+///
+/// The table holds monomial coefficients, so the round coefficients are plain
+/// dot products of the two faces, with no per-element subtraction:
+///
+/// - `h(0)`   = sum_{b} w(0, b) * f(0, b)   (low face)
+/// - `h(inf)` = sum_{b} w(inf, b) * f(inf, b)   (high face, the leading coefficient)
+///
+/// Same `K`-tiled, delayed-reduction structure as the evaluation-basis kernel.
+pub fn sumcheck_coefficients_prefix_projective<B, A>(evals: &[B], weights: &[A]) -> (A, A)
```
tcoratger: This is `sumcheck_coefficients_prefix` verbatim except `chunk_round_step → chunk_round_step_projective` and `round_step → round_step_projective` — the entire tiling / `PAR_THRESHOLD` / par-vs-serial / tail scaffold is duplicated. Two copies of a delicate delayed-reduction loop will drift over time.

Since both kernels are new, now is the cheapest moment to share the scaffold: parameterize over the per-tile and per-pair step (pass the two step fns, a `const PROJECTIVE: bool`, or a tiny trait), so there's one loop and only the MAC differs.

Re: the inline note about moving this onto `Poly` — I'd push back gently. This is a *binary* op over the `(evals, weights)` pair, and its eval-basis sibling is also a free function; moving only the projective one onto `Poly` would break that symmetry. De-duplication above is the better lever. (The `Poly`-method shape fits `fix_prefix_var_mut_projective`, which is binding — a different operation.)

tcoratger (commented):
Thanks for this — the two new kernels are mathematically correct and, crucially, the default `Eval` path is byte-for-byte unchanged (I checked: no production caller, WHIR included, ever sets `Projective`), so **nothing that ships is at risk**. The projective round kernel is properly tested against a `dot(lo,lo)/dot(hi,hi)` reference, and the binding kernel evaluates the monomial polynomial as claimed.

My hesitation is about the *wiring around* those kernels, not the kernels themselves. Two things are worth resolving before this lands:

1. **The full-prover projective benchmark doesn't actually run a projective prover.** `round()` never consults `self.basis` (only `compress` and `round_coefficients` do), so the `plain_prefix_projective` bench runs eval round-coefficients + projective binding + eval reduction. The headline ~5-6% number therefore doesn't measure the projective round kernel at all, and the path is internally inconsistent (the `debug_assert_eq!(sum, dot_product)` would fire — it's just compiled out in release benches).
2. **`tests/projective_equivalence.rs` doesn't exercise the production code.** It re-implements the whole protocol locally and checks the re-implementation against itself; it would still pass if the production kernel were deleted.

Given your own open question to the maintainers, my suggestion is to **land only the two kernels** (correct + tested) and defer the `ProductPolynomial` dispatch, the full-prover bench, and the integration test to the end-to-end (§4.3) follow-up where the sum-tracking actually exists to make them coherent. Details inline. Nice work on the kernel-level micro-optimization — the delayed-reduction structure is preserved cleanly.

multilinear-util/src/poly.rs
```diff
@@ -1905,6 +1950,43 @@ pub(crate) mod test {
         }
     }
 
+    #[test]
+    fn fix_prefix_var_mut_projective_evaluates_monomial_basis() {
```
tcoratger: Agree with making this a proptest — and this one *does* test production code, so it's worth strengthening. The property (projective binding across all variables = monomial-polynomial evaluation `Σ_S a_S · Π_{i∈S} z_i`) is exactly the cheap, large-space invariant proptest is for; the current single fixed seed samples one path.

```rust
proptest! {
    #[test]
    fn fix_prefix_var_mut_projective_evaluates_monomial_basis(
        k in 1usize..=12, seed in any::<u64>(),
    ) {
        let poly = Poly::<F>::rand(&mut rng, k);
        let z = Point::<F>::rand(&mut rng, k);
        // expected: direct monomial eval, as today
        // got: repeated fix_prefix_var_mut_projective(zi)
        prop_assert_eq!(got, expected);
    }
}
```

tcoratger:
After you solved the comment, could you post again benchmark numbers before/after to check changes? :)

sumcheck/src/strategy.rs
jdodinh: Read your follow-up note and agree: it's a binary op over `(evals, weights)` and its eval sibling is a free function too, so I kept it free and deduped the scaffold instead. The `Poly`-method shape stays with the binding kernel, which is the unary one.

multilinear-util/src/poly.rs
jdodinh: Done, proptest over random `k` and `seed`, asserting projective binding equals the direct monomial evaluation.

sumcheck/benches/sumcheck.rs
jdodinh: Retracted that bench and the ~5-6% "full prover" row. Kept the round-coefficient row (~11%), which measures the real kernel.

sumcheck/tests/projective_equivalence.rs
jdodinh: Dropped. The end-to-end equivalence and soundness checks move to the §4.3 follow-up, where the prover actually runs in the projective basis.

sumcheck/src/product_polynomial.rs
jdodinh: Both moot now: `ProductPolynomial` is reverted to upstream verbatim, since this PR no longer adds the `Basis` dispatch.

sumcheck/src/strategy.rs
jdodinh: Done, in its own `refactor` commit so the eval-path change is isolated. Both bases now go through one `sumcheck_coefficients_prefix_with`, parameterised over the per-tile and per-pair MAC, so only the multiply-accumulate differs. Behaviour-preserving: the eval kernel's output is unchanged and the full suite passes.

sumcheck/tests/projective_equivalence.rs
jdodinh: Moot: the whole file is dropped (per your comment above), so the `dot` helper goes with it.

sumcheck/src/product_polynomial.rs
jdodinh: Moot now: `ProductPolynomial` is reverted to upstream verbatim, so there is no `Basis::Projective` arm and nothing to guard. Same as the [this](https://github.com/Plonky3/Plonky3/pull/1900#discussion_r3482564023) thread.

multilinear-util/src/poly.rs
jdodinh: Done, this is now a proptest over random `k` and `seed`, asserting projective binding across all variables equals the direct monomial evaluation (the property you spelled out). Same change as the sibling thread just above.

jdodinh:
Thanks for the careful review. This was the right call. I've tightened the PR to the two kernels, rewritten the description (it opens with an Edit note listing exactly what changed and why), and replied to each inline comment.

On the before/after numbers you asked for: with the full-prover bench retracted (it wasn't running a projective prover, as you flagged), the meaningful comparison is the round-coefficient kernel, eval vs projective, same inputs, single-threaded (M3 Pro, `-Ctarget-cpu=native`):

| Size | BabyBear | KoalaBear |
|---|---|---|
| k16 | 1.11x | 1.11x |
| k18 | 1.11x | 1.11x |
| k20 | 1.11x | 1.11x |
| k22 | 1.11x | 1.11x |

(speedup = eval / projective, so >1 means projective is faster; reproduce with `cargo +1.93.0 bench -p p3-sumcheck --bench sumcheck -- "round_coefficients/ext_ext_packed_prefix"`.) A consistent ~11% on the kernel, on the 31-bit Monty fields. History rewritten 5 -> 4 commits.

jdodinh:
While building the full WHIR integration, I found that the round-coefficient kernel here sent an insufficient message: `[s(0), s(inf)]` plus the round identity leaves the quadratic's middle coefficient undetermined, so a verifier cannot evaluate the round polynomial at its challenge. The complete message per the paper's Fig. 3 is `[s(1), s(inf)]`
with `s(0)` derived. I folded the correction into this branch so the kernel is right from the start rather than fixed in the follow-up, and updated the description accordingly.

Consequence for the numbers: the ~11% I reported was an artifact of benchmarking the insufficient message (two pure dot products) against the complete evaluation kernel. The corrected comparison is parity (0.98-0.99x), which matches the paper's own cost accounting for degree 2: the message swaps a subtraction pass for an addition pass. The real per-round win of the projective basis is the subtraction-free binding, measured standalone in the follow-up PR at ~1.13x.

The follow-up PR (full WHIR integration + end-to-end measurement) is currently a draft, and I will be ready by the end of today; heads up that its result is negative per the pre-registered threshold, with the full analysis in its description.

multilinear-util/src/poly/mod.rs
```diff
@@ -492,6 +492,51 @@ impl<A: Copy + Send + Sync + PrimeCharacteristicRing> Poly<A> {
...
+    /// ```text
+    /// out = p(0, x') + r * p(inf, x') = a0 + r * a1
+    /// ```
+    ///
+    /// Unlike [`Self::fix_prefix_var_mut`], this costs no subtraction per pair.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the polynomial is constant (zero free variables).
+    pub fn fix_prefix_var_mut_projective<F: Copy + Send + Sync>(&mut self, r: F)
```
Nashtare: This (and others) are introduced at the public level but never used outside of tests (same as in child PR #1949). Is it really necessary to have them part of the public API?

sumcheck/src/strategy.rs
```diff
@@ -672,4 +765,33 @@ mod tests {
             );
         }
     }
+
+    proptest! {
+        // Projective (monomial-basis) prefix round message (eprint 2026/762,
+        // Fig. 3) is [s(1), s(inf)] = [dot(lo + hi, lo + hi), dot(hi, hi)];
+        // the per-pair subtractions of the evaluation basis become additions.
+        #[test]
+        fn prop_sumcheck_coefficients_prefix_projective_matches_reference(
```
Nashtare, line 906: I think this proptest restates the implementation, not the invariant (cf the issue you mentioned in the PR discussion around verifier not able to evaluate `s(r)`. I think something like this would have caught it

```rust
  let claim: EF = dot_product(evals.iter().copied(), weights.iter().copied());
  let (s1, s_inf) = sumcheck_coefficients_prefix_projective(&evals, &weights);
  let s0 = claim - s_inf;                                  // verifier-derived, never sent
  let s_at_r = s0 + (s1 - s0 - s_inf) * r + s_inf * r.square();

  let (mut be, mut bw) = (Poly::new(evals), Poly::new(weights));
  be.fix_prefix_var_mut_projective(r);
  bw.fix_prefix_var_mut_projective(r);

  prop_assert_eq!(s_at_r, dot_product(be.as_slice().iter().copied(),
                                      bw.as_slice().iter().copied()));
```

multilinear-util/src/poly/mod.rs
Nashtare: The docstring says `monomial` basis, yet there's no clear different (especially given this is currently `pub`, see API related comment), whereas all other poly methods assume evaluations, and misusing this may cause hard to debug errors, so may deserve a special naming?

sumcheck/src/strategy.rs
```diff
@@ -167,12 +210,62 @@ where
...
+/// The tables are interpreted as monomial coefficients. The round message is
+/// `[s(1), s(inf)]`; the verifier derives `s(0) := C - s(inf)` from the
+/// projective round identity `s(0) + s(inf) = C`. Returned as:
+///
+/// - `s(1)`   = sum_{b} (w(0,b) + w(inf,b)) * (f(0,b) + f(inf,b))
+/// - `s(inf)` = sum_{b} w(inf, b) * f(inf, b)   (leading coefficient)
+///
+/// The evaluation-basis kernel's per-pair subtractions (`hi - lo`) become
+/// additions (`lo + hi`); same `K`-tiled, delayed-reduction structure.
+pub fn sumcheck_coefficients_prefix_projective<B, A>(evals: &[B], weights: &[A]) -> (A, A)
```
Nashtare: `sumcheck_coefficients_prefix` returns `(h(0), h(inf))` but this (seemingly only the projective equivalent) returns `(s(1), s(inf)` which index difference may make things confusing / hard to debug if mishandled at dispatch sites. Having a named struct as return type would prevent silly bugs

multilinear-util/src/poly/mod.rs
```diff
@@ -492,6 +492,51 @@ impl<A: Copy + Send + Sync + PrimeCharacteristicRing> Poly<A> {
         self.0.truncate(mid);
     }
 
+    /// Fixes the prefix variable at a challenge value in the *monomial* basis,
+    /// in place.
+    ///
+    /// The table is interpreted as monomial coefficients (the projective
+    /// `{0, inf}` representation of eprint 2026/762). Writing the polynomial as
```
Nashtare:
```suggestion
    /// `{0, inf}` representation of <https://eprint.iacr.org/2026/762> ). Writing the polynomial as
```

Nashtare (commented):
Overall this is looking good @jdodinh, thanks for this! I think we can polish this a bit but should be close to mergeable!

multilinear-util/src/poly/mod.rs
jdodinh: `fix_prefix_var_mut_monomial` (renamed per your naming comment) needs `pub`: its consumer is `Basis::fix_var` in p3-sumcheck (#1949), a cross-crate call. The round kernels are consumed by `Basis::sumcheck_coefficients` there, and stay `pub` for the Criterion benches (which link as an external crate), matching the existing evaluation kernels. Within this PR alone, they're test/bench-only by construction of the stack.

sumcheck/src/strategy.rs
jdodinh: Agreed, added in 31db5785, essentially your snippet. Mutation-checked: reverting the kernel to the old `[s(0), s(inf)]` message fails it. Kept the reference proptest alongside for direct arithmetic coverage.

multilinear-util/src/poly/mod.rs
jdodinh: Renamed to `fix_prefix_var_mut_monomial` in d88a5e10, matching the `eval_monomial` family in #1949, so every coefficient-interpreting method carries the same marker.

sumcheck/src/strategy.rs
jdodinh: Done in bfef29d1: all kernels now return `RoundMessage { c_a, c_inf }`, documented per basis; `c_inf` is the leading coefficient in both, `c_a` the sent finite point.

multilinear-util/src/poly/mod.rs
jdodinh: Applied in d88a5e10.

sumcheck/src/product_polynomial.rs
```diff
@@ -339,16 +339,16 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
         //
         // The strategy differs based on representation to maximize SIMD utilization.
         let order = self.order;
-        let (c0, c_inf) = match &self.inner {
+        let RoundMessage { c_a: c0, c_inf } = match &self.inner {
```
Nashtare: I think it doesn't really prevent the hazard it names, i.e. both bases share the field name c_a and the same type, so a projective message is silently accepted anywhere an eval one is expected

Nashtare: I think it'd make more sense to either propagate it down to `observe_and_sample`/`extrapolate_01inf` and parameterize it by basis (pulling the `Basis` from PR #1949)

Nashtare (commented):
Overall LGTM. Note that the three new bench IDs for the kernel provably can't win, adding a monomial variant in `bench_fix_var` (sumcheck crate) may be worth it?

Nashtare (commented):
@jdodinh let me know when you'll have addressed the last nits and we'll be good to merge :)

sumcheck/src/product_polynomial.rs
jdodinh: You're right. Fixed by pulling `Basis` from #1949 and routing both ends of a message through it, per your follow-up.

- `Basis::sumcheck_coefficients` is now the only producer of a `RoundMessage`, and asserts the projective kernels are paired with prefix binding, which was also unenforced.
- `Basis::reduce_claim` is the only consumer, and the single place either round identity is written down. `ProductPolynomial::round` and `SumcheckData::verify_rounds` both go through it now instead of each spelling out `extrapolate_01inf(c_a, C - c_a, c_inf, r)` itself.
- `verify_rounds` / `verify_final_sumcheck_rounds` take the basis explicitly, so a verifier has to state which message shape it is reading rather than assuming.

Three tests: the existing projective round-identity proptest now also pins `reduce_claim` against the same identity written out by hand; a new proptest pins that the two identities are genuinely different reductions of the same `(c_a, c_inf, C)`, which is the thing the struct alone cannot distinguish; and the prefix-only precondition has a `should_panic` case. Mutation-checked: swapping the projective arm of `reduce_claim` for the evaluation one fails both proptests.

One scoping note, since this partly walks back a call @tcoratger made in the first round. He asked me to strip the `Basis` dispatch out of this PR and defer it to the end-to-end follow-up, on the grounds that a dispatch with no coherent sum-tracking was not a well-defined protocol. I have backported only the message half here: `Basis` plus the two message methods, not `Basis::fix_var` or the `ProductPolynomial` basis field, which stay in #1949 where they have callers. `reduce_claim` is the sum-tracking he wanted, so I do not think this reopens his objection, but flagging it rather than quietly re-expanding scope. Happy to go the other way if you would both rather it wait.

Every production caller passes `Basis::Evaluation`, so the shipping path is unchanged.

jdodinh:
@Nashtare thanks for the ping, both nits are addressed. Rebased onto latest main while I was in there.

On the `RoundMessage` one, you were right that the named struct did not prevent the hazard it documented. `Basis` is pulled up from #1949 so a message is only ever produced and consumed under the same tag, with `Basis::reduce_claim` as the single place either round identity is written down. Details and the scoping caveat are [here](https://github.com/Plonky3/Plonky3/pull/1900#discussion_r3831576014).

On the bench IDs: agreed, and it lines up with what I reported in July. The projective round message trades a subtraction pass for an addition pass, so at degree 2 it is parity by the paper's own cost accounting and those three IDs cannot come out ahead by construction. Binding is where the basis actually drops a pass.

Added `fix_var/prefix_projective`, running `fix_prefix_var_mut_monomial` against the existing prefix/suffix evaluation rows on the same buffer. Single-threaded, `-Ctarget-cpu=native`, Rust 1.93.0, M3 Pro. Speedup = eval prefix / projective, so >1 means projective is faster:

| Size | BabyBear | KoalaBear |
|---|---|---|
| k16 | 1.13x | 1.12x |
| k18 | 1.13x | 1.13x |
| k20 | 1.14x | 1.13x |
| k22 | 1.10x | 1.13x |

Reproduce with `cargo +1.93.0 bench -p p3-sumcheck --bench sumcheck -- "fix_var"`.

Ready for another look.

jdodinh:
The red check is not from this branch. Rust 1.98 (2026-08-18) added `clippy::chunks_exact_to_as_chunks`, and CI's lint job tracks stable, so a pristine `main` fails the same way. The reported site is in `field/src/field.rs`, which this PR does not touch.

Fix landed on main through #1994.

Nashtare (approved):
Thanks for the changes @jdodinh! LGTM! :)

PR #1904, opened 2026-07-02
perf(matrix): no-alloc `vertically_packed_row` for `RowIndexMappedView`

Nashtare:
`BitReversedMatrixView` never overrode `vertically_packed_row`, so it fell back to the generic `Matrix` default which allocates a `Vec` via `wrapping_row_slices` on every call. This is on the hot path of `uni_stark::prover::quotient_values`, called twice per SIMD-packed group.

Override it to gather each packed lane directly through get_unchecked, avoiding the intermediate `Vec` allocation. Benchmarked on a `BitReversedMatrixView`: ~57-59% faster across 2^10-2^18 row matrices.

### DIFF
--- a/matrix/src/row_index_mapped.rs
+++ b/matrix/src/row_index_mapped.rs
@@ -150,6 +150,26 @@ impl<T: Send + Sync + Clone, IndexMap: RowIndexMap, Inner: Matrix<T>> Matrix<T>
         self.inner
             .padded_horizontally_packed_row(self.index_map.map_row_index(r))
     }
+
+    #[inline]
+    fn vertically_packed_row<P>(&self, r: usize) -> impl Iterator<Item = P>
+    where
+        T: Copy,
+        P: PackedValue<Value = T>,
+    {
+        let height = self.height();
+        let width = self.width();
+        // The row permutation generally scatters `r..r + P::WIDTH` across non-contiguous
+        // inner rows, so unlike `DenseMatrix` we cannot take a contiguous-slice fast path.
+        // Reading elements directly still avoids the `Vec` of row-slice guards that the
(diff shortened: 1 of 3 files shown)


matrix/src/row_index_mapped.rs
```diff
@@ -150,6 +150,26 @@ impl<T: Send + Sync + Clone, IndexMap: RowIndexMap, Inner: Matrix<T>> Matrix<T>
...
+        // The row permutation generally scatters `r..r + P::WIDTH` across non-contiguous
+        // inner rows, so unlike `DenseMatrix` we cannot take a contiguous-slice fast path.
+        // Reading elements directly still avoids the `Vec` of row-slice guards that the
+        // default implementation allocates via `wrapping_row_slices`.
+        (0..width).map(move |c| {
+            P::from_fn(|i| unsafe {
+                // Safety: (r + i) % height < height, and c < width (loop bound).
+                self.get_unchecked((r + i) % height, c)
+            })
+        })
```
tcoratger, line 176: Maybe here there is a no wrap fast path that we could add to avoid the modulo as per 

```rust
let no_wrap = P::WIDTH != 1 && r + P::WIDTH <= height;
(0..width).map(move |c| {
    if no_wrap {
        // r + i < height, so no wrap and no modulo.
        P::from_fn(|i| unsafe { self.get_unchecked(r + i, c) })
    } else {
        P::from_fn(|i| unsafe { self.get_unchecked((r + i) % height, c) })
    }
})
```

tcoratger (approved):
LGTM, shouldn't we have the same for `vertically_packed_row_pair` which should have the same problem no?

PR #1905, opened 2026-07-02
perf(whir, sumcheck): used packing in constraint combination and HZVK initial sumcheck relation

Nashtare:
Added a `combine` benchmark in the `p3-sumcheck` crate. Compared to `main`, I'm getting ~60% improvement for this step on my M4 pro.

### DIFF
--- a/multi-stark/src/packed_ext.rs
+++ b/multi-stark/src/packed_ext.rs
@@ -1,0 +1,189 @@
+//! Local wrapper granting a packed extension-field value arithmetic against
+//! the base field directly.
+
+use core::fmt;
+use core::iter::{Product, Sum};
+use core::marker::PhantomData;
+use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+
+use p3_field::{Algebra, Field, PrimeCharacteristicRing};
+
+/// Wraps a packed extension-field value so it supports arithmetic against
+/// the base field `F` directly, i.e. implements `Algebra<F>`.
+///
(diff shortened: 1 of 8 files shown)


sumcheck/benches/sumcheck.rs
```diff
@@ -414,6 +414,47 @@ where
     Constraint::new(rng.random(), k, vec![Statements::Eq(eq)])
 }
 
+/// Benches folding a large `SelectStatement` (STIR-query-shaped) into an
+/// unpacked weight accumulator, isolated from any round-folding cost.
+fn bench_combine<B: BenchField>(c: &mut Criterion)
+where
+    StandardUniform: Distribution<B::F> + Distribution<B::EF>,
+{
+    use p3_sumcheck::constraints::statement::SelectStatement;
```
tcoratger: This import top of the file maybe?

Nashtare: oops sorry, done!

tcoratger (approved):
LGTM

multi-stark/src/packed_ext.rs
tcoratger, line 1: Mhhhhh shouldn't we have this in the field crate instead?

Nashtare: I put it there on purpose because of a few reasons:

- it was to get around some Rust impl conflicts there was with `p3-field` which would have caused breaking changes if addressed directly there
- it's a partial `PrimeCharacteristicRing` impl
- `MultilinearFolder` is the only consumer of this module, so it made sense to me to keep it under the `multi-stark` crate

Happy to move it though if you feel strongly

PR #1907, opened 2026-07-02
feat(multi-stark): reusable preprocessed-column commitment

tcoratger:
## What

Adds preprocessed-column support to the `multi-stark` multilinear AIR prover. The preprocessed trace is committed **once** at setup and reused across every proof for that AIR and trace height.

The main trace and the preprocessed trace live in **two independent WHIR commitments**, opened at the single point the zerocheck binds. This required no change to the `PrescribedPointPcs` signature — `multi-stark` simply drives two independent WHIR instances.

## How

- **`keys` (new)**: `setup` commits the preprocessed trace once. `ProvingKey` keeps the committed data (to open each proof); `VerifyingKey` keeps only the commitment. Both sides re-absorb the stored commitment into the real transcript before sampling any challenge.
- **`rounds`**: the zerocheck round state folds preprocessed columns alongside the main columns (they fold identically), slicing them back into a main window + a preprocessed window at every folder-construction site (packed and unpacked, base and extension). No behavior change and negligible overhead when there is no preprocessed trace.
- **`zerocheck`**: `layout` now allows preprocessed columns (still rejects periodic); `prove` returns split main/preprocessed opening claims; `check_constraint` threads the preprocessed window via a new `TableOpening` view.
- **`prover`/`verifier`**: observe, open, and verify the preprocessed commitment.
- **`whir` / `merkle-tree`**: derive `Clone` on `MerkleTree` and implement `Clone` on `WhirProverData`, so a reused commitment hands a clone to each opening — copying only the committed data and skipping the codeword re-encode and Merkle rebuild. The base Merkle data is only read during opening, so this is sound.

## Design note

The plan called for a reusable preprocessed commitment. Since WHIR opens one committed polynomial per instance, "reusable" is realized as: the commitment (Merkle root) is fixed in the keys, and each proof clones the committed prover data to open at its own point. The alternative — threading a borrow lifetime through WHIR's `RoundState`/`RoundData` to avoid the per-proof clone — is a larger, riskier change to the hot proving path and can be a follow-up.

## Tests

- **Zerocheck layer**: packed and unpacked paths swept over trace heights `1..10`, checking preprocessed openings equal the column multilinear (and its shift) at the bound point, plus current- and next-row tamper rejections.
- **End-to-end over WHIR** (`whir_preprocessed.rs`): roundtrip, reuse across two proofs from one setup, violated-constraint rejection, and tampered-preprocessed-opening rejection.

Full suite green (`multi-stark`, `whir`, `merkle-tree`), plus `cargo fmt` and `clippy` clean.


multi-stark/src/prover.rs
```diff
@@ -91,26 +103,64 @@ where
...
     );
 
-    // 1. Commit to the trace columns; the scheme absorbs its commitment into the transcript.
+    // Preprocessed and main columns share the bound point, so they must share a height.
+    if let Some(preprocessed) = preprocessed {
+        assert_eq!(
+            preprocessed.log_height, log_height,
+            "preprocessed trace height must match the main trace height"
+        );
+    }
```
Nashtare, line 121: Shouldn't we also have an assertion on widths? for both prover and verifier

Nashtare: or at least some `assert_eq!(preprocessed.is_some(), air.preprocessed_width() > 0, ...)`

Nashtare (approved):
overall LGTM

multi-stark/src/prover.rs
tcoratger: Fixed

PR #1910, opened 2026-07-03
perf(commit): bitmask periodic-table indexing & batch periodic-column verifier evals

Nashtare:


commit/src/periodic.rs
```diff
@@ -73,20 +73,29 @@ pub struct PeriodicLdeTable<F> {
...
     ///
     /// The matrix should have height = `max_period × blowup` and width = `num_periodic_columns`.
     pub const fn new(values: RowMajorMatrix<F>) -> Self {
-        Self { values }
+        let height = match values.values.len().checked_div(values.width) {
+            Some(h) => h,
+            None => 0,
+        };
+        Self { values, height }
     }
```
tcoratger, line 96: Maybe we should add something like this

```rust
debug_assert!(height == 0 || height.is_power_of_two(),
    "PeriodicLdeTable height must be a power of two for bitmask indexing");
```

with a test like `PeriodicLdeTable::new(RowMajorMatrix::new(vec![a,b,c], 1)) `

commit/src/domain.rs
```diff
@@ -331,4 +350,91 @@ impl<Val: TwoAdicField> PolynomialSpace for TwoAdicMultiplicativeCoset<Val> {
...
+        &self,
+        periodic_columns: &[Vec<Val>],
+        point: Ext,
+    ) -> Vec<Ext> {
+        let mut cols_by_period: BTreeMap<usize, Vec<usize>> = BTreeMap::new();
+        for (i, col) in periodic_columns.iter().enumerate() {
+            cols_by_period.entry(col.len()).or_default().push(i);
+        }
+
+        let mut result = vec![Ext::ZERO; periodic_columns.len()];
```
tcoratger: Maybe this?

```suggestion
        let mut result = Ext::zero_vec(periodic_columns.len());
```

tcoratger (approved):
LGTM!

PR #1911, opened 2026-07-03
perf(matrix): pack rows through composition wrappers

Nashtare:
I didn't include additional benchmark files, but they yield ~60/~70% savings on vertically_packed_row() / vertically_packed_row_pair().

### DIFF
--- a/matrix/src/lib.rs
+++ b/matrix/src/lib.rs
@@ -428,10 +428,12 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
         T: Field,
         EF: ExtensionField<T>,
     {
-        // Below this height, the rayon fork-join and SIMD-packing machinery costs more than
-        // the dot product itself; fall back to a plain scalar accumulation.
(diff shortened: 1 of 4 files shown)


matrix/src/lib.rs
```diff
@@ -428,6 +428,19 @@ pub trait Matrix<T: Send + Sync + Clone>: Send + Sync {
         T: Field,
         EF: ExtensionField<T>,
     {
+        // Below this height, the rayon fork-join and SIMD-packing machinery costs more than
+        // the dot product itself; fall back to a plain scalar accumulation.
+        const SMALL_HEIGHT: usize = 16;
+        if self.height() <= SMALL_HEIGHT {
```
tcoratger: Maybe instead here we should gate on total work instead of just height

if self.height() <= SMALL_HEIGHT && self.width() <= SOME_WIDTH { ... }
// or
if self.height().saturating_mul(self.width()) <= SMALL_ELEMS { ... }

tcoratger (approved):
LGTM!

PR #1913, opened 2026-07-03
perf(sumcheck): parallelize packed constraint combine

Nashtare:
Route the unpacked combine arm through `combine_new_packed` and unpack-add its delta with `par_chunks_mut` instead of a serial pass. `NextConstraint::accumulate_packed` no longer materializes a 2^k scalar table per constraint; it builds each packed output slot directly from the selector/successor tables in parallel.

### DIFF
--- a/sumcheck/src/constraints/statement/next.rs
+++ b/sumcheck/src/constraints/statement/next.rs
@@ -372,29 +372,46 @@ impl<F: Field> NextConstraint<F> {
         let selector_eq = Poly::new_from_point(self.selector.as_slice(), scale);
         let selector_eq = selector_eq.as_slice();
 
-        let local_rows = local_next.len();
-        let selector_rows = selector_eq.len();
+        // Both axis lengths are powers of two; splitting the global index via a shift
+        // and a mask (instead of `/` and `%`) keeps the per-lane hot line free of
+        // runtime `udiv`/`urem`, since LLVM cannot prove the power-of-two bound on
+        // values that live behind the `Base` type parameter.
+        let log_local = log2_strict_usize(local_next.len());
+        let mask_local = local_next.len() - 1;
+        let log_selector = log2_strict_usize(selector_eq.len());
+        let mask_selector = selector_eq.len() - 1;
         let width = Base::Packing::WIDTH;
+
+        // Widest packing width in this codebase (AVX-512); see e.g.
+        // `monty-31/src/dft/{forward,backward}.rs`.
+        const MAX_W: usize = 16;
+        debug_assert!(width <= MAX_W);
 
         // Every scalar entry of the dense table is `selector_eq[slot] * local_next[local]`,
         // where `(slot, local)` is recovered from the global hypercube index by the same
-        // block layout `accumulate` uses. Each packed output slot is built directly from
-        // that pair, in parallel, so no `2^total_vars`-sized scalar buffer is ever built.
+        // block layout `accumulate` uses. Each lane product is computed once into a small
+        // scratch buffer, then packed via `from_ext_slice` (which only indexes into it).
+        // `from_ext_fn` would instead re-run the closure once per basis coefficient,
+        // recomputing the same extension multiply `D` times per lane.
         match self.var_order {
             VariableOrder::Prefix => {
                 out.par_iter_mut().enumerate().for_each(|(i, out)| {
-                    *out += F::ExtensionPacking::from_ext_fn(|lane| {
+                    let mut lanes = [F::ZERO; MAX_W];
+                    for (lane, slot) in lanes[..width].iter_mut().enumerate() {
                         let g = i * width + lane;
-                        selector_eq[g / local_rows] * local_next[g % local_rows]
-                    });
+                        *slot = selector_eq[g >> log_local] * local_next[g & mask_local];
+                    }
+                    *out += F::ExtensionPacking::from_ext_slice(&lanes[..width]);
                 });
             }
             VariableOrder::Suffix => {
                 out.par_iter_mut().enumerate().for_each(|(i, out)| {
-                    *out += F::ExtensionPacking::from_ext_fn(|lane| {
-                        local_next[(i * width + lane) / selector_rows]
-                            * selector_eq[(i * width + lane) % selector_rows]
-                    });
+                    let mut lanes = [F::ZERO; MAX_W];
+                    for (lane, slot) in lanes[..width].iter_mut().enumerate() {
+                        let g = i * width + lane;
+                        *slot = local_next[g >> log_selector] * selector_eq[g & mask_selector];
+                    }
+                    *out += F::ExtensionPacking::from_ext_slice(&lanes[..width]);
                 });
             }
         }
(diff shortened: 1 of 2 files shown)


sumcheck/src/constraints/statement/next.rs
```diff
@@ -363,19 +365,39 @@ impl<F: Field> NextConstraint<F> {
...
+        // Every scalar entry of the dense table is `selector_eq[slot] * local_next[local]`,
+        // where `(slot, local)` is recovered from the global hypercube index by the same
+        // block layout `accumulate` uses. Each packed output slot is built directly from
+        // that pair, in parallel, so no `2^total_vars`-sized scalar buffer is ever built.
+        match self.var_order {
+            VariableOrder::Prefix => {
+                out.par_iter_mut().enumerate().for_each(|(i, out)| {
+                    *out += F::ExtensionPacking::from_ext_fn(|lane| {
+                        let g = i * width + lane;
+                        selector_eq[g / local_rows] * local_next[g % local_rows]
```
tcoratger: **The math here is spot on — this is purely about how many times the multiply runs.** ⚡

Here's the subtle bit. `from_ext_fn(f)` doesn't call `f` once per lane — it calls it **`D * WIDTH` times**, once per `(basis_coefficient, lane)` pair (it's documented right on the trait method, and the `BinomialExtensionField` override still goes through `pack_columns_fn`, whose only impl is `array::from_fn(|col| Packing::from_fn(|lane| row_fn(lane)[col]))` — `col` is the *outer* loop, so each lane's value is recomputed `D` times).

And the closure we hand it isn't cheap — it's a full extension multiply `selector_eq[..] * local_next[..]`.

**Walk through one lane, BabyBear quartic (`D = 4`):**

```
lane fixed → g fixed → we want ONE product  selector_eq[slot] * local_next[local]

from_ext_fn asks for it 4 times (coeff 0,1,2,3):
  coeff 0: compute the full EF*EF product, keep .value[0]
  coeff 1: compute the SAME EF*EF product again, keep .value[1]
  coeff 2: ... again, keep .value[2]
  coeff 3: ... again, keep .value[3]
```

So we pay for 4 extension multiplies where 1 would do. Across the buffer that's `2^total_vars * D` multiplies vs `2^total_vars` in the old dense path (there `from_ext_slice` only *loaded* already-computed products). For `D=4` that's ~4× the field-mul work; Goldilocks `D=2` is ~2×.

The new `par_iter_mut` adds parallelism the old serial fold didn't have, which probably makes the wall-clock look good on a many-core box — but it's masking `D`× more total work (worse on few cores, wasted cycles everywhere).

**The fix keeps everything you gained** (no spike, still parallel) and gets the multiply count back to 1×: compute the `WIDTH` lane products once, then pack via `from_ext_slice` (whose closure only indexes):

```rust
const MAX_W: usize = 16; // >= any Base::Packing::WIDTH
out.par_iter_mut().enumerate().for_each(|(i, out)| {
    let mut lanes = [F::ZERO; MAX_W];
    for (lane, slot) in lanes[..width].iter_mut().enumerate() {
        let g = i * width + lane;
        *slot = selector_eq[g / local_rows] * local_next[g % local_rows];
    }
    *out += F::ExtensionPacking::from_ext_slice(&lanes[..width]);
});
```

(Same shape for the `Suffix` arm.) Alternatively, override `pack_columns_fn` to evaluate `row_fn(lane)` once into a `[Value; N]` and distribute — that fixes it for every `from_ext_fn` caller, but it's a broader change.

**To confirm the factor:** bench `combine_packed` at e.g. `k=20, sel=4` on 1–2 threads (to remove the parallelism confound) before/after — the current branch should show ~`D`× the multiply time of the precomputed version.

tcoratger: Tiny free win while you're in here: `local_rows` and `selector_rows` are always powers of two (`2^loc` / `2^sel`), but as *runtime* values LLVM can't prove that, so `/` and `%` become real `udiv`/`urem` on the innermost per-lane line. Hoisting the log once turns them into a shift and a mask:

```rust
let log_local = log2_strict_usize(local_rows);
let mask_local = local_rows - 1;
// slot = g >> log_local;   local = g & mask_local;
```

Same idea for `selector_rows` in the `Suffix` arm. Small next to the multiply point above, but it's on the same hot line so worth folding in together. 🙂

sumcheck/src/constraints/statement/next.rs
```diff
@@ -561,4 +584,59 @@ mod tests {
         assert_eq!(scalar.as_slice(), &unpacked[..]);
         assert_eq!(packed_sum, scalar_sum);
     }
+
+    proptest! {
+        #[test]
+        fn prop_combine_packed_agrees_with_combine(
```
tcoratger, line 642: Love this proptest. 💚 It's exactly the right tool for the job: the property ("packed unpacks to scalar, bit-for-bit") holds over a big input space, and it cross-checks two *independent* implementations — scalar `combine` via `accumulate` as the trusted oracle against the new packed path — plus the running sum. Fuzzing both `VariableOrder`s and a range of `sel`/`loc` splits is great coverage.

One cheap boundary to pin explicitly: `sel == k` (empty local point → `local_rows == 1`), which the current `sel in 0..5` with `k in 4..9` will rarely hit. That exercises the degenerate `g / 1 == g`, `g % 1 == 0` layout. `sel == 0` is already reachable, so you're half covered.

sumcheck/src/product_polynomial.rs
```diff
@@ -554,20 +555,21 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
...
+                    let (packed_delta, packed_sum) = constraint.combine_new_packed();
                     *sum += packed_sum;
                     weights
                         .as_mut_slice()
-                        .iter_mut()
-                        .zip(EF::ExtensionPacking::to_ext_iter(
-                            packed_delta.iter().copied(),
-                        ))
-                        .for_each(|(out, delta)| *out += delta);
+                        .par_chunks_mut(F::Packing::WIDTH)
```
tcoratger, line 566: This rewrite is faithful — I checked it end to end. 👌

The old flatten-then-zip did `weights[j] += packed_delta[j/WIDTH].extract(j%WIDTH)` (that's exactly what `to_ext_iter` expands to). The new version chunks `weights` into groups of `WIDTH` and does `weights[i*WIDTH + lane] += packed_delta[i].extract(lane)` — same mapping, just parallel and without building the intermediate flat iterator. Lengths line up cleanly because the branch guard guarantees `weights.num_variables() >= k_pack`, so `2^nv` is an exact multiple of `WIDTH` and there's no ragged final chunk.

And as noted in the summary — routing through `combine_new_packed` + the new `assert_eq!` is strictly safer than the old sizing. Nice.

tcoratger (commented):
Just made a review with Claude since this one can go pretty deep!

I traced the index math for both `Prefix` and `Suffix` against `accumulate` and it matches byte-for-byte, and the new proptest + existing tests pass locally. The memory win is real too — dropping the `2^total_vars` scalar spike per constraint is a clear improvement, and the added `assert_eq!(constraint.num_variables(), weights.num_variables())` is a genuine upgrade (the old code sized `packed_delta` off `weights` while `combine_packed` used the constraint's arity, so a mismatch would've silently truncated the `zip`).

The one thing worth a look before merge is a hidden `D`× multiply blowup in the direct-construction path — details inline. Everything else is just polish. Left a couple of friendly notes. 🚀

### The idea, in one picture

Every entry of the dense table is a product of two small tables:

```
    dense[g] = selector_eq[slot] * local_next[local]

    Prefix:  out = [ slot_0 block | slot_1 block | ... ]   block = local_rows entries
             slot = g / local_rows,   local = g % local_rows
```

The old code built the whole `dense` array, then folded runs of `WIDTH` into packed slots. The new code skips the array and rebuilds each `(slot, local)` pair straight from `g = i*WIDTH + lane`. Same numbers, no spike. 👍

tcoratger (approved):

PR #1917, opened 2026-07-03
refactor: fold MultiOpeningMmcs into Mmcs

tcoratger:
## What

Removes the separate `MultiOpeningMmcs` trait introduced in #1805 and moves its three members onto the existing `Mmcs` trait:

- `type MultiProof`
- `fn open_multi_batch`
- `fn verify_multi_batch`

Call sites change only in the bound name (`MultiOpeningMmcs<F>` → `Mmcs<F>`). Every code path keeps its pruned/amortized behaviour — the method bodies are unchanged, just relocated into the single `Mmcs` impl block of each type.

## Why

The extra trait was confusing and not needed. There are only four `Mmcs` implementors:

- the `&M` blanket impl
- `ExtensionMmcs`
- `MerkleTreeMmcs`
- `MerkleTreeHidingMmcs`

The first three already implemented `MultiOpeningMmcs`, so folding is a pure move. Associated-type defaults are unstable on stable Rust, so `MultiProof` cannot be defaulted — but every implementor declares it, so requiring it on the base trait costs nothing.

## The one real cost, handled properly

`MerkleTreeHidingMmcs` previously implemented only `Mmcs`, so folding forces it to gain multi-opening. Rather than a `todo!()`/panic footgun, it gets a real implementation: it delegates to the inner tree's pruned multi-opening and splits/re-appends the salt suffix exactly the way its single-opening path already does — including the load-bearing unsalted-width check (`check_widths` before salting) that prevents an over-long row hiding behind an under-long salt.

Two tests cover this newly reachable path:
- a multi-opening round trip
- a width-rejection test

## Rebase note

This is a rebase of tcoratger#35 onto current `main`. The only delta beyond the original refactor: #1915 (`fix(whir): update prescribed-point adapter to pruned multiproof API`) landed after #1805 and added a second `MultiOpeningMmcs<F>` bound on the `PrescribedPointPcs` impl in `whir/src/pcs/adapter.rs`. That bound is renamed to `Mmcs<F>` here too, so nothing references the removed trait.

## Verification

- `cargo build -p p3-commit -p p3-merkle-tree -p p3-whir --all-targets` ✅
- `cargo clippy -p p3-commit -p p3-merkle-tree -p p3-whir --all-targets` ✅ (no warnings)
- tests: commit (3) + merkle-tree (105, incl. the two new hiding tests) + whir (131, default and `parallel`) + multi-stark `whir_fibonacci` (4) pass
- `cargo fmt --check` ✅

### DIFF
--- a/merkle-tree/src/hiding_mmcs.rs
+++ b/merkle-tree/src/hiding_mmcs.rs
@@ -15,6 +15,7 @@ use serde::{Deserialize, Serialize};
 use spin::Mutex;
 
 use crate::mmcs::check_widths;
+use crate::pruning::PrunedMerklePaths;
 use crate::{MerkleCap, MerkleTree, MerkleTreeError, MerkleTreeMmcs};
 
 /// A vector commitment scheme backed by a `MerkleTree`.
@@ -117,6 +118,12 @@ where
     type Commitment = MerkleCap<P::Value, [PW::Value; DIGEST_ELEMS]>;
     /// The first item is salts; the second is the usual Merkle proof (sibling digests).
     type Proof = (Vec<Vec<P::Value>>, Vec<[PW::Value; DIGEST_ELEMS]>);
+    /// The first item is the per-query salts (`salts[q][m]`).
+    /// The second is the shared pruned multiproof.
+    type MultiProof = (
+        Vec<Vec<Vec<P::Value>>>,
+        PrunedMerklePaths<PW::Value, DIGEST_ELEMS>,
+    );
     type Error = MerkleTreeError;
 
     fn commit<M: Matrix<P::Value>>(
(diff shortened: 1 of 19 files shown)


merkle-tree/src/hiding_mmcs.rs
```diff
@@ -192,6 +199,81 @@ where
...
+                    .into_iter()
+                    .map(|row| {
+                        let (a, b) = row.split_at(row.len() - SALT_ELEMS);
+                        (a.to_vec(), b.to_vec())
+                    })
+                    .unzip();
+                salts.push(salts_at_index);
+                opened
+            })
+            .collect();
```
Nashtare: Wouldn't it be a tad cleaner? 


```suggestion
         let (opened_values, salts): (Vec<_>, Vec<_>) = salted_values
             .into_iter()
             .map(|rows_at_index| {
                 rows_at_index.into_iter()
                     .map(|row| { let (a, b) = row.split_at(row.len() - SALT_ELEMS); (a.to_vec(), b.to_vec()) })
                     .unzip()
             })
             .unzip();
```

merkle-tree/src/hiding_mmcs.rs
```diff
@@ -192,6 +199,81 @@ where
...
+        &self,
+        indices: &[usize],
+        prover_data: &Self::ProverData<M>,
+    ) -> (Vec<Vec<Vec<P::Value>>>, Self::MultiProof) {
+        // The inner tree opens salted rows: each row is `real_row ++ salt`.
+        let (salted_values, pruned) = self.inner.open_multi_batch(indices, prover_data);
+
+        // Split every salted row back into its unsalted prefix and its salt suffix.
+        // Salts move into the proof.
+        // The unsalted rows are returned to the caller.
```
Nashtare: probably not needed
```suggestion
```

Nashtare (approved):
LGTM, just one minor code suggestion

merkle-tree/src/hiding_mmcs.rs
tcoratger: Fixed

merkle-tree/src/hiding_mmcs.rs
tcoratger: Fixed

PR #1918, opened 2026-07-04
perf(fri): prune Merkle multiproofs across the FRI query phase

wstran:
## What

Each FRI query ships one full root-to-leaf authentication path per input batch and per commit-phase round. With 100 queries those paths overlap heavily: the upper layers of every tree are shared by most queries, and the late fold rounds are tiny trees where 100 near-identical paths ship almost the same sibling digests 100 times.

Root cause: `QueryProof` bundles openings per query, so the proof carries every shared sibling once per query and `verify_fri` recomputes every shared compression once per query.

Fix: wire the FRI query phase through the pruned Merkle multiproof machinery that #1805 introduced for WHIR (`open_multi_batch` / `verify_multi_batch`), one shared proof per tree. `FriProof` replaces `query_proofs: Vec<QueryProof>` with `input_openings` (one `BatchMultiOpening` per input batch) and `commit_phase_openings` (one `CommitPhaseMultiStep` per round). `log_arity` moves to once per round, where the schedule always lived conceptually, so the cross-query arity consistency checks disappear. The verifier still reconstructs each full evaluation row around its own folded value before the shared per-round check, so the binding of folded evaluations to committed codewords is unchanged. Queries whose folded indices collide in late rounds are collapsed by the multiproof, which also rejects inconsistent duplicate openings.

The Fiat-Shamir transcript is unchanged: query indices are sampled in one block with nothing observed between samples, on both sides. Checked empirically, not just by inspection: `commit_phase_commits`, `final_poly` and the query PoW witness are byte-identical to `main` for the same inputs at every size below.

```text
before: per query   q0:[input paths, round 0 path, round 1 path, ...]  ... x100
after:  per tree    [input multiproof][round 0 multiproof][round 1 multiproof]...
```

## Numbers

BabyBear quartic, Poseidon2 Merkle (32-byte digests), `FriParameters::new_benchmark` (100 queries, blowup 1, binary folding), uni-stark prove of a 3-column mul AIR, postcard-serialized proofs; medians of 5 warm runs on an M3, baseline measured on `main` (5cd0cf7f) with the same harness.

| log_n | proof size | Merkle digests | verify | prove |
|-------|------------|----------------|--------|-------|
| 16 | 703,307 B -> 378,407 B (-46.2%) | 17,000 -> 8,612 | 15.26 -> 7.53 ms (2.0x) | 33 -> 33 ms |
| 18 | 858,746 B -> 482,124 B (-43.9%) | 20,900 -> 11,145 | 18.38 -> 9.71 ms (1.9x) | 112 -> 109 ms |
| 20 | 1,031,604 B -> 628,081 B (-39.1%) | 25,200 -> 14,754 | 21.83 -> 12.55 ms (1.7x) | 396 -> 399 ms |

The verifier win comes from computing each shared compression once instead of once per query. The prover is flat: the pruning work (sort plus LCA dedup over the query set) is negligible against the commit phase.

One honest caveat: at trivial query counts the multiproof bookkeeping outweighs the dedup, so the 2-query golden fixture grew from 1,577 B to 1,594 B. The savings start at real query counts.

## Correctness

- Transcript artifacts (`commit_phase_commits`, `final_poly`, query PoW witness) are byte-identical to `main` for the same inputs at log_n 16/18/20.
- All 61 `p3-fri` tests pass, including the reshaped malformed-proof suite: tampered digests, wrong counts per round/batch/query, invalid or oversized arity schedules, schedule under/overshoot, the two-adicity bound, and index substitution.
- Duplicate folded indices (many queries landing in one node of a late round) are exercised by 100-query proofs folding down to height-2 trees, plus the multiproof's own duplicate-consistency tests.
- The hiding PCS stays on the same path via a new `MultiOpeningMmcs` impl for `MerkleTreeHidingMmcs` (per-query salts alongside the inner pruned proof); `test_zk` passes.
- Golden proof fixtures re-blessed (uni-stark, batch-stark). Full workspace suite green, clippy clean with all features, fmt clean.

## Scope

- Two-adic FRI only. The circle PCS has its own copy of the per-query proof types; porting it the same way is a natural follow-up PR, which would also retire the three per-query `FriError` variants kept alive here for circle's sake.
- Ordering note: this PR keeps the `MultiOpeningMmcs` bounds of current `main`. If #1917 (folding that trait into `Mmcs`) lands first, the bounds here rename mechanically. Happy to rebase in either order.


tcoratger:
Great! Not a review but I think that we should merge #1917 first to clarify things and start this one on a clear basis (this will simplify the trait system).

Nashtare:
Hi @wstran, would you mind rebasing your work now that #1917 has been merged? It should be good for review after! Thanks!

wstran:
Done, rebased onto main. Adapted to #1917 (MultiOpeningMmcs bounds folded into Mmcs, dropped the duplicate hiding impl) and the flattened PrunedMerklePaths; full workspace green, both two-adic fixtures re-blessed.

tcoratger (approved):

PR #1919, opened 2026-07-04
perf(merkle-tree): optimal frontier pruning for multi-opening proofs

tcoratger:
## In one line

Make WHIR proofs smaller by shipping only the Merkle sibling hashes the verifier genuinely cannot recompute — the provably minimal set.

## The problem

To prove a leaf sits in a Merkle tree, the prover ships its **authentication path**: one sibling hash per level up to the root.

Opening many leaves at once ships many paths, and near the top they overlap heavily — the same hashes travel over and over.

#1805 already pruned *some* of that overlap. It sorted the opened leaves and let each path drop only the hashes it shared with its **immediate sorted neighbour**. That catches sharing between adjacent pairs and misses the rest.

## The idea

Flip the question. Instead of "what does this path share with its neighbour?", ask:

> **Can the verifier recompute this sibling from the leaves it already knows?**

If yes, don't send it. Walk the tree bottom-up over the queried nodes (the "frontier"); at each level, send a child hash **only when no queried leaf sits under it**. Everything else is forced by the tree and gets recomputed for free.

## Worked example — open leaves 1, 2, 5

```
                 root
                /    \
              A        B
             / \      / \
            C   D    E   F
           /\  /\   /\   /\
           0 1 2 3  4 5  6 7
```

```
Level 0:  known {1,2,5}   1's pair 0 → send 0
                          2's pair 3 → send 3
                          5's pair 4 → send 4
Level 1:  known {C,D,E}   C's pair D → known → free (C,D → A)
                          E's pair F → send F
Level 2:  known {A,B}     A's pair B → known → free (A,B → root)
```

**Sent: 4 hashes.** The old adjacent scheme sends 8 here; naive sends 9.

The win is at level 1: because leaves 1 and 2 live in neighbouring subtrees, their parents `C` and `D` combine for free — the old scheme never noticed and shipped both.

## One more shrink

The queried indices come from the verifier's own transcript, so the proof no longer stores them. It becomes a **flat list of hashes**. That drops extra bytes and removes any leaf index a cheater could swap.

## Numbers (WHIR, KoalaBear + quintic ext, λ=128, CapacityBound, folding=4, rate 1/2)

`before` = `main` (LCA-adjacent pruning), `after` = this branch. Same machine, criterion, back to back.

**Proof size**

| size | proof bytes (before → after) | Δ | Merkle digests (before → after) |
|------|------------------------------|---|---------------------------------|
| **2¹⁴** | 53,801 → 42,101 | **−22 %** | 786 → 500 |
| **2¹⁸** | 89,465 → 76,448 | **−15 %** | 1527 → 1211 |
| **2²⁰** | 111,071 → 97,172 | **−13 %** | 1994 → 1658 |

Bytes shrink less than digest count because proofs also carry opened rows and sumcheck data; the pruning win is largest at small sizes where Merkle paths dominate.

**Prover / verifier time** (criterion medians)

| size | prover (before → after) | verifier (before → after) |
|------|-------------------------|---------------------------|
| **2¹⁴** | 846.8 → 855.1 ms (+1.0 %) | 1.184 → 1.183 ms (≈0 %) |
| **2¹⁸** | 572.0 → 577.5 ms (+1.0 %) | 2.116 → 2.079 ms (−1.7 %) |
| **2²⁰** | 809.6 → 814.3 ms (+0.6 %) | 2.631 → 2.647 ms (+0.6 %) |

Both prover and verifier are flat — every delta sits inside criterion's noise band. This differs from #1805, which also sped verification up by switching to an amortized walk. This PR keeps that same walk and only ships fewer siblings: the verifier recomputes the same set of inner nodes either way, so its time is unchanged and restoration adds no hashing. A pure proof-size win at no time cost.

## How it plugs in safely

Generalized beyond the binary / single-root / uniform-leaf setting to Plonky3's N-ary arity schedules, Merkle caps, and mixed-height matrix injection.

Restoration is **hash-free**: it replays the same grouping by pure index arithmetic and scatters each boundary hash into exactly the slots the existing amortized verifier reads. The injection and cap logic is untouched and still re-derives the root against the commitment — so a bad hash or a substituted index still fails as before.

## Tests

- `p3-merkle-tree`: 88 pass — the frontier proof is checked equal to individual openings across binary, 4-ary, mixed-arity, caps, duplicates, and an exhaustive (height × cap × query-subset) sweep, plus rejection tests for wrong sibling counts and substituted indices.
- `p3-whir`: 131 pass (default and `parallel`).
- `p3-multi-stark` `whir_fibonacci`: 4 pass.
- Clippy clean, `cargo fmt --check` clean.

## Notes

- The idea is from @TomWambsgans's [leanVM commit](https://github.com/leanEthereum/leanVM/commit/d63f10033b5b24bd7a4550dbca75f44b67347539); credited as co-author.
- The bench's `pow_bits` budget is raised so the default config builds again — it currently panics on `main` with `PowBitsExceedBudget` after #1914.

### DIFF
--- a/merkle-tree/src/lib.rs
+++ b/merkle-tree/src/lib.rs
@@ -12,4 +12,4 @@ pub use hiding_mmcs::*;
 pub use merkle_tree::MerkleTree;
 pub use mmcs::{MerkleTreeError, MerkleTreeMmcs, PrunedBatchOpening, PrunedProofError};
 pub use p3_symmetric::MerkleCap;
-pub use pruning::{PrunedMerklePaths, PrunedPath};
+pub use pruning::PrunedMerklePaths;
--- a/merkle-tree/src/mmcs/error.rs
+++ b/merkle-tree/src/mmcs/error.rs
@@ -74,82 +74,22 @@ pub enum MerkleTreeError {
 /// Each variant pins one failure mode with its diagnostic fields.
 #[derive(Debug, Error)]
 pub enum PrunedProofError {
-    /// More unique paths than the tree can hold.
-    #[error("too many unique paths: {got} exceeds tree height {max_height}")]
-    TooManyUniquePaths {
-        /// Number of unique paths claimed by the proof.
-        got: usize,
-        /// Maximum admissible height (the tallest committed matrix).
-        max_height: usize,
-    },
-
-    /// A restored path has the wrong number of siblings for the tree geometry.
-    #[error("restored path has {got} siblings, expected {expected}")]
+    /// The flat sibling list does not match the frontier the query set implies.
+    ///
+    /// The verifier's query set fixes an exact boundary-digest count.
+    /// A proof carrying too few or too many digests lands here.
+    #[error("wrong pruned sibling count: expected {expected}, got {got}")]
     SiblingCountMismatch {
-        /// Sibling count implied by the verifier-known arity schedule.
+        /// Digest count the verifier-derived frontier requires.
         expected: usize,
-        /// Sibling count produced while restoring the path.
+        /// Digest count the proof actually carried.
         got: usize,
     },
 
-    /// An `original_order` entry references a unique-path slot that does not exist.
-    #[error("original-order references missing path slot {slot} of {num_paths}")]
-    OriginalOrderOutOfRange {
-        /// Slot index referenced by the entry.
-        slot: usize,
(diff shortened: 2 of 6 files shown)


Nashtare (approved):
Not much to say, looking really good!

PR #1922, opened 2026-07-06
perf(circle): precompute per-query FRI twiddle chain, hoist verifier-side redundant work

Nashtare:
## Summary

- verifier: precompute and batch-invert the per-query FRI x-fold twiddle chain (squaring recurrence with a per-round sign flip) instead of recomputing each round's twiddle from scratch via a fresh scalar multiplication; reuse the lambda-correction point for the first-layer y-twiddle instead of re-deriving it
- verifier: hoist per-matrix alpha^width/alpha^(2*width) and a shared alpha-powers table out of the per-query closure
- domain: hoist the per-coset-point constant out of selectors_on_coset, and fuse selectors_at_point's four vanishing-poly evaluations and three inversions into one plus a batch inversion
- domain: parallelize CircleDomain point materialization via chunked scalar-mul seeding instead of a single serial point-addition chain

## Performance

Verifier is ~22% faster and prover is ~4% faster E2E.

### DIFF
--- a/circle/src/cfft.rs
+++ b/circle/src/cfft.rs
@@ -13,9 +13,11 @@ use p3_maybe_rayon::prelude::*;
 use p3_util::{log2_ceil_usize, log2_strict_usize, reverse_slice_index_bits};
 use tracing::{debug_span, instrument};
 
+#[cfg(test)]
+use crate::cfft_permute_index;
 use crate::domain::CircleDomain;
 use crate::point::{Point, compute_lagrange_den_batched};
-use crate::{CfftPermutable, CfftView, cfft_permute_index, cfft_permute_slice};
+use crate::{CfftPermutable, CfftView, cfft_permute_slice};
 
 #[derive(Clone)]
 pub struct CircleEvaluations<F, M = RowMajorMatrix<F>> {
@@ -514,6 +516,9 @@ impl<F: ComplexExtendable> CircleDomain<F> {
         reverse_slice_index_bits(&mut ys);
         ys
     }
+    /// Used only by [`crate::folding::fold_y_row`], the test-only reference implementation
+    /// of [`crate::folding::fold_y`]; see that function's doc comment.
+    #[cfg(test)]
     pub(crate) fn nth_y_twiddle(&self, index: usize) -> F {
         self.nth_point(cfft_permute_index(index << 1, self.log_n)).y
     }
--- a/circle/src/cfft.rs
+++ b/circle/src/cfft.rs
@@ -13,8 +13,6 @@ use p3_maybe_rayon::prelude::*;
 use p3_util::{log2_ceil_usize, log2_strict_usize, reverse_slice_index_bits};
 use tracing::{debug_span, instrument};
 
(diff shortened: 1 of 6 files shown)


circle/src/cfft.rs
```diff
@@ -13,9 +13,11 @@ use p3_maybe_rayon::prelude::*;
 use p3_util::{log2_ceil_usize, log2_strict_usize, reverse_slice_index_bits};
 use tracing::{debug_span, instrument};
 
+#[cfg(test)]
+use crate::cfft_permute_index;
```
tcoratger: That should go in test module directly no?

circle/src/cfft.rs
```diff
@@ -514,6 +516,9 @@ impl<F: ComplexExtendable> CircleDomain<F> {
         reverse_slice_index_bits(&mut ys);
         ys
     }
+    /// Used only by [`crate::folding::fold_y_row`], the test-only reference implementation
+    /// of [`crate::folding::fold_y`]; see that function's doc comment.
+    #[cfg(test)]
     pub(crate) fn nth_y_twiddle(&self, index: usize) -> F {
         self.nth_point(cfft_permute_index(index << 1, self.log_n)).y
     }
```
tcoratger: Maybe should go directly in test module as well?

circle/src/folding.rs
```diff
@@ -147,9 +151,65 @@ pub(crate) fn fold_x_row<F: ComplexExtendable, EF: ExtensionField<F>>(
...
 
+/// Fold a pair of sibling evaluations given an already-inverted twiddle, shared by both the
+/// first-layer (y) fold and every FRI (x) fold round: `fold_*_row` differ only in how `t` is
+/// derived, not in this arithmetic.
+pub(crate) fn fold_row_with_inv_twiddle<F: Field, EF: ExtensionField<F>>(
+    inv_twiddle: F,
+    beta: EF,
+    evals: impl Iterator<Item = EF>,
+) -> EF {
+    let evals = evals.collect_vec();
```
tcoratger: Here I think this is unnecessary to have this allocation. 

We could have:
```rust
let mut it = evals;
let (e0, e1) = (it.next().unwrap(), it.next().unwrap());
debug_assert!(it.next().is_none());
```

or evals being directly `[EF; 2]`

circle/src/domain.rs
```diff
@@ -109,6 +111,33 @@ impl<F: ComplexExtendable> CircleDomain<F> {
...
+    g: Point<F>,
+    len: usize,
+) -> Vec<Point<F>> {
+    if len == 0 {
+        return Vec::new();
+    }
+    let num_chunks = current_num_threads().max(1).min(len);
+    let chunk_len = len.div_ceil(num_chunks);
+    (0..len.div_ceil(chunk_len))
+        .into_par_iter()
```
tcoratger, line 132: Here maybe we should go to a non parallel route for small dimensions no?

tcoratger (approved):
LGTM! Small nits!

PR #1925, opened 2026-07-06
perf(keccak): add plain-NEON 2-wide backend for non-SHA3 aarch64

Nashtare:
## Summary

`aarch64` targets with NEON but without the SHA3 crypto extension (Graviton2, Raspberry Pi 4/5, etc.) previously fell all the way to the scalar 1-wide fallback, losing all lane parallelism for Keccak Merkle leaf hashing and packed compression.

Rename the existing SHA3-accelerated backend to neon_sha3.rs and add a new neon.rs: a mechanical port of sse2.rs's 2-lane structure using plain NEON intrinsics (veorq_u64/vbicq_u64 for xor/andn, and a vshlq_n_u64+vsriq_n_u64 fused shift-insert idiom for rotates), gated on aarch64 + neon + not(sha3). The fallback module now only applies to targets that are neither x86_64 nor aarch64-with-neon.

## Performance

```bash
  ┌─────────────────────────────┬─────────────────────────────────┬───────────────────────────────┬──────────────┐
  │            Bench            │ Prior (scalar fallback, 1-wide) │ Post (new plain-NEON, 2-wide) │ Δ throughput │
  ├─────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼──────────────┤
  │ keccak permutation          │ 149 ns/call · 869 MiB/s         │ 270 ns/call · 960 MiB/s       │ +10.4%       │
  ├─────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼──────────────┤
  │ keccak u64 hash (slice)     │ 902 ns · 846 MiB/s              │ 1639 ns · 931 MiB/s           │ +10.1%       │
  ├─────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼──────────────┤
  │ keccak u64 hash (iter)      │ 904 ns · 844 MiB/s              │ 1671 ns · 913 MiB/s           │ +8.3%        │
  ├─────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼──────────────┤
  │ keccak field32 hash (slice) │ 473 ns · 806 MiB/s              │ 832 ns · 917 MiB/s            │ +13.8%       │
  ├─────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼──────────────┤
  │ keccak field32 hash (iter)  │ 464 ns · 822 MiB/s              │ 823 ns · 927 MiB/s            │ +12.8%       │
  └─────────────────────────────┴─────────────────────────────────┴───────────────────────────────┴──────────────┘
  ```

### DIFF
--- a/keccak/src/neon_sha3.rs
+++ b/keccak/src/neon_sha3.rs
@@ -218,6 +218,7 @@ impl CryptographicPermutation<[[u64; VECTOR_LEN]; 25]> for KeccakF {}
 
 #[cfg(test)]
 mod tests {
+    use proptest::prelude::*;
     use tiny_keccak::keccakf;
 
     use super::*;
@@ -279,10 +280,10 @@ mod tests {
         ],
     ];
 
-    fn our_res() -> [[u64; 25]; 2] {
+    fn our_res(states: [[u64; 25]; 2]) -> [[u64; 25]; 2] {
         let mut packed_result = [[0; 2]; 25];
         for (i, packed_res) in packed_result.iter_mut().enumerate() {
-            *packed_res = [STATES[0][i], STATES[1][i]];
+            *packed_res = [states[0][i], states[1][i]];
         }
 
         keccak_perm(&mut packed_result);
@@ -295,8 +296,8 @@ mod tests {
         result
     }
 
-    fn tiny_keccak_res() -> [[u64; 25]; 2] {
-        let mut result = STATES;
+    fn tiny_keccak_res(states: [[u64; 25]; 2]) -> [[u64; 25]; 2] {
+        let mut result = states;
         keccakf(&mut result[0]);
         keccakf(&mut result[1]);
         result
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: keccak/src/neon_sha3.rs [hunk 1]
```diff
@@ -0,0 +1,311 @@
...
+impl Permutation<[[u64; VECTOR_LEN]; 25]> for KeccakF {
+    fn permute_mut(&self, state: &mut [[u64; VECTOR_LEN]; 25]) {
+        keccak_perm(state);
+    }
+}
+
+impl CryptographicPermutation<[[u64; VECTOR_LEN]; 25]> for KeccakF {}
+
+#[cfg(test)]
+mod tests {
```
tcoratger, line 220: Maybe here in addition to the known state test we should have also a small proptest to better cover the whole space of possibilities no?

tcoratger (approved):
I think that the new added tests don't compile on the CI so this may be a problem. `neon.rs` is gated on `all(aarch64, neon, not(sha3))`, but the only aarch64 CI runner is macos-latest (Apple Silicon), and aarch64-apple-darwin enables target_feature="sha3" by default.

I think that no job sets -C target-feature=-sha3. So for me the new hand-written SIMD permutation and its `test_vs_tiny_keccak` is never built or run by CI.

Maybe we should add a small run like
```
- name: Test keccak (aarch64, no SHA3)
  run: cargo test -p p3-keccak
  env:
    RUSTFLAGS: -C debuginfo=0 -C target-feature=-sha3
```

PR #1926, opened 2026-07-06
refactor: couple API changes

Nashtare:
## Summary

Couple of API refactor suggestion, bundled together for clarity. Mostly around tightening, safety and clarity. Also include some reorganization (like the `UniformSamplingField` trait which I feel is better suited under `p3-field`?).

Open to suggestions and feeback, I can revert each commit independently.


uni-stark/src/security.rs
```diff
@@ -67,25 +67,36 @@ pub struct StarkSecurityParams {
...
+        fri_log_final_poly_len: usize,
+        fri_max_log_arity: usize,
+        fri_num_queries: usize,
+        fri_commit_proof_of_work_bits: usize,
+        fri_query_proof_of_work_bits: usize,
         num_modulus_bits: usize,
         collision_resistance: usize,
         num_constraints: usize,
         air_max_constraint_degree: usize,
         max_combo: usize,
```
tcoratger: Since these are a lot of parameters, maybe we should think about some kind of object to store them in a followup PR

poseidon2-air/src/constants.rs
```diff
@@ -44,4 +71,47 @@ impl<
...
+        external: &ExternalLayerConstants<F, WIDTH>,
+        internal: &[F],
+    ) -> Result<Self, RoundConstantsShapeError> {
+        let initial = external.get_initial_constants();
+        let terminal = external.get_terminal_constants();
+        if initial.len() != HALF_FULL_ROUNDS || terminal.len() != HALF_FULL_ROUNDS {
+            return Err(RoundConstantsShapeError::FullRounds {
+                expected: HALF_FULL_ROUNDS,
+                got: initial.len().max(terminal.len()),
+            });
```
tcoratger, line 94: Here I think that if `initial.len() == HALF_FULL_ROUNDS` but `terminal.len()` is smaller the got becomes `got = max(correct, wrong)` which reports the correct value so the error reads expected N, got N, which is confusing. Probably we should report here whichever side actually mismatched.

tcoratger (approved):
LGTM! Small nits

PR #1927, opened 2026-07-06
perf(goldilocks): speed-up NEON Poseidon2 lanes and remove `MdsMatrixGoldilocks` redundant DFT

Nashtare:
About ~3% / ~4% faster on my M4 pro


goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1301,6 +1302,7 @@ pub fn lanes_to_neon<const WIDTH: usize>(
 }
 
 /// Convert NEON vector array back to separate lane arrays.
+#[cfg(test)]
 #[inline]
 pub fn neon_to_lanes<const WIDTH: usize>(
     state_v: &[uint64x2_t; WIDTH],
```
tcoratger, line 1308: Nit but maybe we should move these methods directly in the test module to avoid confusion when reading the main file, but not a problem at all

tcoratger (approved):

PR #1929, opened 2026-07-06
fix(monolith-air): guard match-flag batching precondition, correct docs

tcoratger:
Follow-up to #1924.

## Context

#1924 batched two modulus one-bits into each committed match-flag cell of the Bar canonical-form walk (`m = m_prev * x_a * x_b`, degree 3). The walk descends MSB → LSB and, at a modulus **zero**-bit, reuses the running flag from the last *completed* pair to enforce the side constraint `flag * bit = 0`.

That reuse is correct only when **no fold-pair straddles a zero-bit** — equivalently, every zero-bit has an even number of one-bits above it. Both supported fields satisfy this:

```
Mersenne31 : 1..1 (x31)                 no zero-bits         -> ok
Goldilocks : 1..1 (x32) 0..0 (x31) 1    even run above zeros -> ok
KoalaBear  : 1..1 (x7)  0..0 (x23) 1    odd run above zeros  -> would over-reject
```

But `MonolithAir` is generic over the field and the invariant was unchecked, while the docs claimed the walk "works for any prime". For a modulus like KoalaBear the batched zero-bit side constraint tests a too-coarse flag and **over-rejects valid canonical inputs**. This is a completeness issue only — it never admits a non-canonical encoding, so it is not a soundness regression — but it should fail loudly at construction rather than silently mis-constrain.

## Changes

- **Guard the precondition** in `MonolithAir::new`: reject any modulus where a zero-bit has an odd count of one-bits above it. Mirrors the eval-walk's pairing logic exactly.
- **Correct the docs**: drop the "works for any prime" claim, state the batching precondition precisely, and rewrite the match-flag documentation one-idea-per-line.
- **Fix the stale degree comment** for Mersenne31: the batched match-flag step and the inlined trailing-limb output are also degree 3 (not "all others ≤ 2").
- **Add a Goldilocks degree test** pinning the degree-3 cap for the second field.

## Testing

- `cargo test -p p3-monolith-air`: 20/20 pass (both new degree tests included; both fields still construct, so the guard does not falsely reject them).
- `cargo fmt --check` and `cargo clippy --all-targets`: clean.

### DIFF
--- a/monolith-air/src/air.rs
+++ b/monolith-air/src/air.rs
@@ -133,13 +133,20 @@ impl<
     ///
     /// # Match-flag width invariant
     ///
-    /// - `NUM_MATCH_FLAGS` must equal `modulus.count_ones() / 2`: two
-    ///   consecutive modulus one-bits share one committed flag cell via a
-    ///   degree-3 batched step (`m_i = m_prev2 * x_a * x_b`), and a
-    ///   modulus-zero bit needs no cell at all (its flag is a pure copy of
-    ///   the previous one). If the Hamming weight is odd, the final
-    ///   one-bit (always bit 0) is folded directly into the closing
-    ///   `assert_zero` with no committed cell of its own.
+    /// - `NUM_MATCH_FLAGS` must equal `modulus.count_ones() / 2`.
+    /// - The canonical-form walk pairs modulus one-bits two at a time.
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
LGTM!

PR #1932, opened 2026-07-08
feat: make `p3-security` composable over any LDT, and account for batched openings

Nashtare:
Follow-up to #1686. Makes the soundness crate generic enough for out-of-tree protocols and closes a missing error term.

  What's in it
  - `LowDegreeTest` trait + `SecurityReport` — the composite (`stark::proven_security_report`) now composes over any LDT via a trait (`FriRegime` is one impl), and returns a labeled
  per-term breakdown (ALI / DEEP / LDT / batch / collision) that exposes which term binds, instead of two bare usizes.
  - `FriParameters::security_regime()` — one canonical `FriParameters` → `FriRegime` assembler in `p3-fri`, using exhaustive destructuring so a new `FriParameters` field fails to
  compile until it's mapped. Replaces four hand-copied literals in examples.
  - Batched-openings term (B.3) — new `InstanceShape.num_batched_functions` feeds a batch-combination proximity term (prox_gaps_error: UD in the UDR regime, JB in the LDR
  regime). `uni-stark::ProvenSecurity` now routes through the report path so the term is actually applied.

### DIFF
--- a/security/src/ldt.rs
+++ b/security/src/ldt.rs
@@ -1,0 +1,36 @@
+//! Typed low-degree-test boundary.
+//!
+//! [`crate::stark`]'s composite orchestration is generic over the LDT through
+//! this trait: any protocol — FRI, WHIR, or an out-of-tree drop-in —
+//! implements [`LowDegreeTest`] and composes with the AIR + DEEP-ALI bounds
+//! via [`crate::stark::proven_security_report`]. The FRI implementation lives
+//! in [`crate::fri`].
+
+use crate::error::ErrorBits;
+use crate::shape::{InstanceShape, StarkAirParams};
+
+/// A low-degree test whose per-regime error terms compose with the AIR +
+/// DEEP-ALI bounds at the protocol call site.
+///
+/// The proximity regimes mirror [`crate::proximity`]: unique decoding (list
+/// size 1) and list decoding at an explicit proximity parameter `m`. An
+/// implementation reports only the **LDT-only** error for each regime; the
+/// composite ([`crate::stark::proven_security_report`]) folds in the ALI,
+/// DEEP, extra, and commitment-collision terms.
+pub trait LowDegreeTest {
+    /// `log2` of the LDT blowup factor (rate ρ = 2^{−log_blowup}). The
+    /// composite uses it to reconstruct the list-decoding list size for the
+    /// regime returned by [`Self::best_ldr`].
+    fn log_blowup(&self) -> usize;
+
+    /// LDT-only error in the unique-decoding regime.
+    fn proven_error_udr(&self, air: &StarkAirParams, shape: &InstanceShape) -> ErrorBits;
+
+    /// Best list-decoding regime for this instance: the proximity parameter
+    /// `m` maximising LDT-only security together with the error attained.
+    /// `None` when no valid `m` exists (e.g. the trace is too small).
+    fn best_ldr(&self, air: &StarkAirParams, shape: &InstanceShape) -> Option<(usize, ErrorBits)>;
+
+    /// Conjectured LDT error (random-words / heuristic regime).
+    fn conjectured_error(&self, shape: &InstanceShape) -> ErrorBits;
+}
--- a/security/src/ldt.rs
+++ b/security/src/ldt.rs
@@ -1,10 +1,11 @@
 //! Typed low-degree-test boundary.
 //!
 //! [`crate::stark`]'s composite orchestration is generic over the LDT through
-//! this trait: any protocol — FRI, WHIR, or an out-of-tree drop-in —
-//! implements [`LowDegreeTest`] and composes with the AIR + DEEP-ALI bounds
-//! via [`crate::stark::proven_security_report`]. The FRI implementation lives
-//! in [`crate::fri`].
+//! this trait: any protocol — FRI, WHIR, or an out-of-tree drop-in — can
+//! implement [`LowDegreeTest`] and compose with the AIR + DEEP-ALI bounds via
+//! [`crate::stark::proven_security_report`]. The FRI implementation lives in
+//! [`crate::fri`]; [`crate::whir`] currently exposes only the underlying WHIR
+//! error terms, not yet a [`LowDegreeTest`] impl.
 
 use crate::error::ErrorBits;
 use crate::shape::{InstanceShape, StarkAirParams};
--- a/security/src/report.rs
+++ b/security/src/report.rs
@@ -54,10 +54,26 @@ pub enum Regime {
 #[derive(Clone, Debug, Serialize)]
 pub struct RegimeReport {
     pub regime: Regime,
-    pub terms: Vec<SecurityTerm>,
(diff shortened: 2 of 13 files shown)


security/src/stark.rs
```diff
@@ -91,6 +97,106 @@ pub fn proven_security(
...
+
+    let ldr = ldt.best_ldr(air, shape).map(|(m, ldr_ldt)| {
+        let list_size = list_size_ldr_m(log_blowup, m);
+        regime_report(
+            Regime::ListDecoding { m },
+            air,
+            shape,
+            list_size,
+            ldr_ldt,
+            batching_term(SecurityAssumption::JohnsonBound, shape, log_blowup),
```
tcoratger: **[MAJOR] The LDR batch term's proximity radius doesn't match the regime it sits in.**

This regime is decoded at `best_m` (from `best_ldr`), and that `m` drives the list size used by the ALI/DEEP/LDT terms via `list_size_ldr_m(log_blowup, m)`. But `batching_term(JohnsonBound, ..)` → `SecurityAssumption::prox_gaps_error` hardcodes the safety choice `m = 10` (η = √ρ/20) in its `JohnsonBound` branch. So inside a single `RegimeReport { regime: ListDecoding { m }, .. }`, ALI/DEEP/LDT use `m = best_m` while `batch-combination` uses `m = 10` — two different list-decoding radii δ = 1 − (1 + 1/(2m))·√ρ.

The batch RLC must be δ-close at the **same** radius FRI decodes at. The proximity-gap error scales as `(m + 1/2)^5`, so a larger operating `m` ⇒ strictly larger (worse) batch error. For the default rate this is the *unsafe* direction — `best_ldr_m` does not pick 10:

```
lb=1, trace 2^20, |F|=252  ->  best_m = 1000   (LDR wins at 65 bits)
lb=1, trace 2^24, |F|=155  ->  best_m = 1000
lb=2, trace 2^22, |F|=128  ->  best_m = 3
```

At `best_m = 1000` the true batch error is ~`5·log2(1000/10) ≈ 33` bits worse than the reported `m = 10` value, so `batch-combination` is **overstated** for the common `log_blowup = 1` config — exactly where the LDR regime is the one that wins and gets reported. Since `SecurityReport` is the audit-facing number, an optimistic term here defeats the "close a missing error term" goal.

Dormant today (`num_batched_functions` defaults to 1), so nothing shipped is wrong — but it becomes an optimistic-soundness bug once the feature is used.

**Fix:** thread this regime's `m` into the LDR batch term instead of the fixed `m = 10`. That needs an `m`-parameterized `prox_gaps_error` (the `JohnsonBound` branch is `m = 10`-only today). If the crate-wide `m = 10` convention is deliberate, please write down why it's a *sound lower bound* on the batch term for all `best_m` — it isn't for `best_m > 10`.

**Test that would pin it:** `batching_lowers_security_when_binding` uses `lb=1, |F|=64` where `best_m = 3` (the conservative side), so it never exercises this. Add a case at `log_blowup = 1` over a small field where the LDR regime wins and the batch term is active, asserting the batch bits equal `prox_gaps_error` at the reported `ListDecoding { m }.m`.

uni-stark/src/security.rs
```diff
@@ -171,6 +178,7 @@ impl ConjecturedSecurity {
             log_trace_length: 0,
             modulus_bits: num_modulus_bits,
             collision_resistance,
+            num_batched_functions: 1,
```
tcoratger, line 186: **[MINOR] Conjectured path silently omits the batching term.**

`ConjecturedSecurity::compute` hardcodes `num_batched_functions: 1` here, and `conjectured_error` never reads the field. Batching is a real soundness cost in the conjectured regime too, so now that the proven path models it the conjectured path is more optimistic than the proven one for batched instances — a confusing asymmetry for anyone comparing the two numbers. Either fold a conjectured batch term in, or add a one-line doc note that the conjectured path deliberately ignores batching.

security/src/report.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Full soundness breakdown within a single proximity regime.
+///
+/// `terms` holds every contribution — ALI, DEEP, LDT, any protocol extras,
+/// and the commitment-collision cap. The attained security is the minimum
+/// over all terms: a collision, or any single binding error, forges the
+/// proof.
+#[derive(Clone, Debug, Serialize)]
+pub struct RegimeReport {
+    pub regime: Regime,
+    pub terms: Vec<SecurityTerm>,
```
tcoratger: **[NIT] `pub terms` + `.expect()` in `binding()` is a constructible footgun.**

With `terms` public, a caller can build `RegimeReport { regime, terms: Vec::new() }` and then panic at the `.expect(...)` in `binding()` (L73). The non-empty invariant holds for values produced by `regime_report`, but the public field exposes an illegal state the `expect` can't survive. Consider a private constructor / non-`pub` `terms` with accessors, or make `binding()` return `Option<SecurityTerm>`.

security/src/ldt.rs
```diff
@@ -0,0 +1,36 @@
+//! Typed low-degree-test boundary.
+//!
+//! [`crate::stark`]'s composite orchestration is generic over the LDT through
+//! this trait: any protocol — FRI, WHIR, or an out-of-tree drop-in —
```
tcoratger: **[NIT]** "any protocol — FRI, WHIR, or an out-of-tree drop-in — implements `LowDegreeTest`" reads as already done, but only `FriRegime` implements it in this PR; `whir.rs` doesn't yet. Maybe phrase it as aspirational ("can implement").

tcoratger (commented):
Reviewed as a Rust + cryptography pass.

The refactor behind `LowDegreeTest` / `SecurityReport` is clean and provably behavior-preserving (the `*_matches_scalar_composite` / `regime_report_matches_proven_security_regime` equivalences + regression vectors + green CI), and `FriParameters::security_regime()` with exhaustive destructuring is exactly the right drift guard.

One substantive concern (inline, **MAJOR**): the **LDR-regime batching term is evaluated at a fixed `m = 10`, inconsistent with the `best_m` the same regime decodes at** — and for the default `log_blowup = 1` that is the optimistic direction. It is dormant while `num_batched_functions = 1` (every in-tree caller), so nothing shipped is affected, but the term reports optimistically the moment it is used as intended.

Also flagged inline: the conjectured path omits the same batching term (MINOR), a `pub terms` + `.expect` footgun (NIT), and a doc overstatement (NIT).

Nice work overall. Verdict: **merge after the LDR `m` mismatch is fixed, or explicitly documented as a known conservative-only-when-`best_m < 10` approximation.**

security/src/stark.rs
tcoratger: I think that mainly this one requires some clarification, the rest is NIT mostly

tcoratger (approved):

security/src/stark.rs
Nashtare: It is fixed now, extracted the Johnson bound dominant term and reused the actual `m` value

PR #1938, opened 2026-07-08
refactor(poseidon-air): dedup S-box, harden trace-gen dispatch, pin packed path to scalar

tcoratger:
Follow-up cleanups on top of #1936, which added the circulant-Karatsuba MDS fast path to Poseidon1 constraint evaluation and the packed (SIMD) Poseidon2 trace generator.

All four changes are behavior-preserving: they make the fast paths harder to break, not faster.

## What and why

**1. One source of truth for the S-box — `poseidon2-air`**

The non-linear step of Poseidon2 raises each state element to a fixed power (`x^3`, `x^7`, `x^11`, …). Since a single AIR constraint can only certify a power up to the max constraint degree, high powers are split across committed "register" columns and rebuilt with a chain of low-degree multiplications.

The scalar generator and the packed generator each carried their own copy of that split, as two identical `match (DEGREE, REGISTERS)` blocks. Two copies of a load-bearing algebraic identity is a divergence hazard: a future edit to one and not the other would silently desync the trace from the constraints.

They now share a single `sbox_with_registers`, generic over any `PrimeCharacteristicRing + Copy`, so a scalar field element and a packed vector run the exact same arithmetic. The scalar and packed functions are thin wrappers that only differ in where the intermediate powers land (a `MaybeUninit` column vs. a returned array).

**2. No silent vectorization cliff — `poseidon2-air`**

The trace generator processed `F::Packing::WIDTH` permutations at a time only when the total count divided evenly into the packing width, otherwise it dropped *the whole trace* to the one-at-a-time path.

It now vectorizes every full group and runs only a short trailing group one at a time:

```text
    inputs:  [ p0 p1 p2 p3 | p4 p5 p6 p7 | p8 p9 ]   (packing_width = 4)
              \__ batch __/ \__ batch __/ \ tail /
```

Targets without SIMD packing (`F::Packing::WIDTH == 1`) still take the plain per-permutation path with no batching overhead.

**3. A test that pins packing to the scalar reference — `poseidon2-air`**

Packing is only ever a throughput transform: the packed batch generator and the one-permutation-at-a-time generator must produce a cell-for-cell identical trace. There was no direct test of that — divergence would only surface indirectly as a failed proof.

Added a test that builds the trace both ways and asserts equality:
- BabyBear — degree-7 S-box with one register (the `x^7`-via-`x^3` arm).
- KoalaBear — degree-3 S-box with no registers (the register-free arm).

Each runs counts `2, 8, 16`. On a target with `F::Packing::WIDTH > 1` (e.g. NEON width 4, AVX2 width 8), the larger counts drive the packed path and the small count drives the trailing-group fallback, so both paths are genuinely compared.

**4. Documented infallible conversions — `poseidon1-air`**

The MDS dispatch reinterprets `&mut [A; WIDTH]` as a fixed `&mut [A; 16]` / `&mut [A; 24]` inside arms already matched on `WIDTH`, so the conversion cannot fail. Swapped the bare `unwrap()` for `expect("WIDTH == 16")` / `expect("WIDTH == 24")` with a comment stating the invariant.

## Testing

- `cargo test -p p3-poseidon2-air` — 5 passed, including the two new equivalence tests.
- `cargo test -p p3-poseidon1-air` — 7 passed (known-answer, prove/verify, constraint degree).
- `cargo test -p p3-mds`, `cargo test -p p3-examples poseidon2` — all green (end-to-end prove/verify).
- `cargo clippy --all-targets` on both crates — no warnings; `cargo fmt --check` — clean.

The equivalence tests were run on arm64 (NEON, packing width 4), so the packed path is exercised rather than degenerating to scalar.

### DIFF
--- a/poseidon1-air/src/mds_dispatch.rs
+++ b/poseidon1-air/src/mds_dispatch.rs
@@ -19,15 +19,18 @@ pub(crate) fn mds_dispatch<F: PrimeCharacteristicRing, A: Algebra<F>, const WIDT
 ) {
     match WIDTH {
         16 => {
-            let state_16: &mut [A; 16] = state.as_mut_slice().try_into().unwrap();
-            let col_16: &[F; 16] = circ_col.as_slice().try_into().unwrap();
+            // WIDTH == 16 in this arm, so both fixed-size views always succeed.
+            let state_16: &mut [A; 16] = state.as_mut_slice().try_into().expect("WIDTH == 16");
+            let col_16: &[F; 16] = circ_col.as_slice().try_into().expect("WIDTH == 16");
             mds_circulant_karatsuba_16(state_16, col_16);
         }
         24 => {
-            let state_24: &mut [A; 24] = state.as_mut_slice().try_into().unwrap();
-            let col_24: &[F; 24] = circ_col.as_slice().try_into().unwrap();
+            // WIDTH == 24 in this arm, so both fixed-size views always succeed.
+            let state_24: &mut [A; 24] = state.as_mut_slice().try_into().expect("WIDTH == 24");
+            let col_24: &[F; 24] = circ_col.as_slice().try_into().expect("WIDTH == 24");
             mds_circulant_karatsuba_24(state_24, col_24);
         }
+        // No circulant fast path for this width: fall back to the dense O(WIDTH^2) multiply.
         _ => mds_multiply(state, dense_mds),
     }
 }
--- a/poseidon2-air/src/generation.rs
+++ b/poseidon2-air/src/generation.rs
@@ -128,9 +128,11 @@ pub fn generate_trace_rows<
 
 /// Fill every permutation's trace columns from its input.
 ///
-/// Runs `F::Packing::WIDTH` permutations at a time through the packed round functions when
-/// the input count divides evenly into that width, falling back to one permutation at a time
-/// (via [`generate_trace_rows_for_perm`]) otherwise.
+/// - With SIMD packing available, process `F::Packing::WIDTH` permutations per task.
+/// - A trailing group of fewer than that many permutations cannot fill a SIMD vector.
+/// - That short group, and every permutation on a target without packing, runs one at a time.
+///
+/// The packed and one-at-a-time paths write identical trace cells, so the choice is pure throughput.
 fn generate_perms<
     F: PrimeField,
     LinearLayers: GenericPoseidon2LinearLayers<WIDTH>,
@@ -152,22 +154,9 @@ fn generate_perms<
     constants: &RoundConstants<F, WIDTH, HALF_FULL_ROUNDS, PARTIAL_ROUNDS>,
 ) {
     let packing_width = F::Packing::WIDTH;
-    if packing_width > 1 && inputs.len().is_multiple_of(packing_width) {
-        perms
-            .par_chunks_mut(packing_width)
-            .zip(inputs.par_chunks(packing_width))
-            .for_each(|(perm_chunk, input_chunk)| {
-                generate_trace_rows_for_perm_batch::<
-                    F,
-                    LinearLayers,
-                    WIDTH,
-                    SBOX_DEGREE,
-                    SBOX_REGISTERS,
-                    HALF_FULL_ROUNDS,
-                    PARTIAL_ROUNDS,
-                >(perm_chunk, input_chunk, constants);
-            });
-    } else {
+
+    // Width 1 means the target has no SIMD packing: batching would add overhead with no gain.
+    if packing_width == 1 {
         perms.par_iter_mut().zip(inputs).for_each(|(perm, input)| {
             generate_trace_rows_for_perm::<
                 F,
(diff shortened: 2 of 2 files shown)


poseidon2-air/src/generation.rs
```diff
@@ -179,7 +168,46 @@ fn generate_perms<
...
     }
+
+    // Split the permutations into groups of `packing_width`.
+    //
+    //     inputs:  [ p0 p1 p2 p3 | p4 p5 p6 p7 | p8 p9 ]   (packing_width = 4)
+    //               \__ batch __/ \__ batch __/ \ tail /
+    //
+    // Full groups run through the packed round functions.
+    // The short final group appears only when the count is not a multiple of the width.
+    perms
```
Nashtare, line 181: I don't think I understand the need for the full split. IIUC the mixed case is reachable only via `generate_vectorized_trace_rows` with a non-power-of-two `VECTOR_LEN` which never happens in practice. So I'm not sure complexifying this part helps if it's basically a no-op.

Nashtare: This is not even exercised in the tests because they used powers of two, so we don't get the "full batches followed by a short tail" situation

poseidon2-air/src/generation.rs
```diff
@@ -509,45 +514,244 @@ fn generate_partial_round<
...
+///
+/// # Panics
+/// Panics unless `(DEGREE, REGISTERS)` is one of the supported pairs:
+/// `(3, 0)`, `(5, 0)`, `(7, 0)`, `(5, 1)`, `(7, 1)`, `(11, 2)`.
+#[inline]
+fn sbox_with_registers<
+    R: PrimeCharacteristicRing + Copy,
+    const DEGREE: u64,
+    const REGISTERS: usize,
+>(
```
Nashtare, line 555: As we're making the table canonical, I think it'd be more meaningful to test over all supported pairs (asserting `output == x.exp_u64(DEGREE)` and register equality with `eval_sbox()`) rather than the spurious BB / KB tests only.

poseidon2-air/src/generation.rs
```diff
@@ -509,45 +514,244 @@ fn generate_partial_round<
...
+/// # Returns
+/// - The S-box output `x^DEGREE`.
+/// - The `REGISTERS` intermediate powers, ordered as the trace columns expect them.
+///
+/// # Panics
+/// Panics unless `(DEGREE, REGISTERS)` is one of the supported pairs:
+/// `(3, 0)`, `(5, 0)`, `(7, 0)`, `(5, 1)`, `(7, 1)`, `(11, 2)`.
+#[inline]
+fn sbox_with_registers<
+    R: PrimeCharacteristicRing + Copy,
```
Nashtare, line 552: is `Copy` needed? `PrimeCharacteristicRing` is already `Dup`

poseidon1-air/src/mds_dispatch.rs
```diff
@@ -19,15 +19,18 @@ pub(crate) fn mds_dispatch<F: PrimeCharacteristicRing, A: Algebra<F>, const WIDT
 ) {
     match WIDTH {
         16 => {
-            let state_16: &mut [A; 16] = state.as_mut_slice().try_into().unwrap();
-            let col_16: &[F; 16] = circ_col.as_slice().try_into().unwrap();
+            // WIDTH == 16 in this arm, so both fixed-size views always succeed.
+            let state_16: &mut [A; 16] = state.as_mut_slice().try_into().expect("WIDTH == 16");
+            let col_16: &[F; 16] = circ_col.as_slice().try_into().expect("WIDTH == 16");
```
Nashtare, line 24: we're having a static match, the `unwrap` -> `expect` isn't meaningful here

poseidon1-air/src/mds_dispatch.rs
```diff
@@ -19,15 +19,18 @@ pub(crate) fn mds_dispatch<F: PrimeCharacteristicRing, A: Algebra<F>, const WIDT
...
             mds_circulant_karatsuba_16(state_16, col_16);
         }
         24 => {
-            let state_24: &mut [A; 24] = state.as_mut_slice().try_into().unwrap();
-            let col_24: &[F; 24] = circ_col.as_slice().try_into().unwrap();
+            // WIDTH == 24 in this arm, so both fixed-size views always succeed.
+            let state_24: &mut [A; 24] = state.as_mut_slice().try_into().expect("WIDTH == 24");
+            let col_24: &[F; 24] = circ_col.as_slice().try_into().expect("WIDTH == 24");
```
Nashtare, line 30: same as above, change not needed

poseidon2-air/src/generation.rs
```diff
@@ -509,45 +514,244 @@ fn generate_partial_round<
...
+            F,
+            LinearLayers,
+            WIDTH,
+            SBOX_DEGREE,
+            SBOX_REGISTERS,
+            HALF_FULL_ROUNDS,
+            PARTIAL_ROUNDS,
+        >(inputs, &constants);
+
+        assert_eq!(packed.values, scalar.values);
```
Nashtare, line 717: shouldn't we compare the matrices directly instead?

Nashtare (commented):
I'm not really convinced overall

tcoratger:
> I'm not really convinced overall

Sure let's close for now, was just a refactor, no behaviour change, we can do that later :)

PR #1939, opened 2026-07-08
feat(multi-stark): support periodic columns in the multilinear AIR prover

tcoratger:
Adds periodic-column support to the multilinear (SuperSpartan-flavored) AIR prover in `multi-stark`. Periodic columns are folded into the zerocheck but never committed — the verifier recomputes them in closed form.

### The problem

An AIR often needs a fixed, *repeating* helper column — a clock, a selector mask, round constants:

```
flag = 1, 0, 0, 0, 1, 0, 0, 0, ...   (period 4)
```

Committing it like a normal trace column wastes work: it Merkle-hashes the column, ships an opening proof, and verifies that proof — all for a known pattern both prover and verifier already hold. It is a public parameter of the AIR, not witness data.

### The idea

A period-$p$ column (with $p$ a power of two) depends only on the low $\log_2 p$ bits of the row index — because row $i$ holds `v[i mod p]`, and `i mod p` *is* the low bits of $i$.

The trace hypercube is indexed **big-endian** (coordinate 0 is the top bit, the last coordinate is the bottom bit), so those low bits are the **last $\log_2 p$ coordinates** of the sumcheck point. The column's multilinear extension therefore collapses onto just those coordinates, and the verifier can evaluate it directly.

```
periodic_col(r_0, ..., r_{k-1}) = MLE_v(r_{k-j}, ..., r_{k-1})     where j = log2(p)
```

### Worked example

Height $n = 4$ (2 variables), one period-2 column `v = [10, 20]`. Materialized: `[10, 20, 10, 20]`.

```
row i = 00 01 10 11   (big-endian: coord0 = top bit, coord1 = bottom bit)
value = 10 20 10 20
```

Reading down: the value tracks `coord1` and ignores `coord0`. So the full extension is just

```
value(r0, r1) = 10*(1 - r1) + 20*r1
```

- **prover** : folds the full `[10,20,10,20]` column through the sumcheck like any column
- **verifier**: at the bound point evaluates `MLE_[10,20]` on the last coordinate only — no `coord0`, no commitment, no opening

Both land on the same value; that equality is the correctness pin the tests check across many heights.

### The change

- **Prover** (`rounds.rs`): each periodic column is materialized to full trace height (`col[i mod p]`) and appended after the main and preprocessed columns in the folded matrix. It threads through the folder via `with_periodic` and produces **no** opening claim.
- **Verifier** (`zerocheck.rs` + `selectors.rs`): a new closed-form evaluator computes each periodic column's multilinear extension at the last $\log_2 p$ coordinates of the bound point, alongside the existing boundary-selector closed forms.
- Removed the old guard that rejected AIRs declaring periodic columns.

### Soundness

- Periodic columns are public parameters fixed by the AIR, not prover-chosen, so leaving them uncommitted opens no gap — exactly as the constraint structure itself is uncommitted.
- Per-variable sumcheck degree is unchanged: the materialized column is a genuine multilinear polynomial (degree one per variable), so it scores like any trace column and Fiat–Shamir stays in lockstep.
- Prover fold and verifier closed form provably agree: the independent leading coordinates' equality factors sum to 1, collapsing the extension onto the low bits.

### Scope

- **New**: closed-form periodic evaluator; end-to-end WHIR test with two periodic columns of different periods.
- **Changed**: round-state folding threads a third column group (main / preprocessed / periodic); verifier constraint recompute feeds periodic values.
- **Untouched**: PCS layer (periodic columns never reach it), commitment, opening protocol.

### Verification

- `cargo test -p p3-multi-stark` — 31 lib + 4 fibonacci + 2 periodic + 4 preprocessed, all pass.
- New tests: prove/verify round-trip, closed-form-vs-materialized-MLE pin swept over heights, and violated-constraint rejection.
- `cargo clippy -p p3-multi-stark --all-targets` and `cargo fmt` clean.


multi-stark/src/selectors.rs
```diff
@@ -319,6 +323,48 @@ where
...
+    EF: ExtensionField<F>,
+{
+    let k = point.len();
+    periodic_columns
+        .iter()
+        .map(|col| {
+            // Number of low-order row bits the column depends on.
+            let j = log2_strict_usize(col.len());
+            // The last j coordinates carry those low-order bits, in order.
+            Poly::new(col.clone()).eval_base(&Point::new(point[k - j..].to_vec()))
```
Nashtare: This can trigger a verifier panic (if period defined is longer than the actual trace length), would probably be best to check and return an early Error than slicing?

multi-stark/tests/whir_periodic.rs
Nashtare, line 1: There are a couple test gaps still:

- if we delete the multi-air skip (`column_offset += air.num_periodic_columns();` in zerocheck module), nothing turns red
- no tests combine preprocessed and periodic
- no test constraint has degree > 1 in a periodic value

multi-stark/src/rounds.rs
```diff
@@ -631,6 +666,37 @@ where
...
+                }
+                let mut values = Vec::with_capacity(cols.len() * trace_height);
+                for col in cols.iter() {
+                    let period = col.len();
+                    assert!(
+                        period.is_power_of_two() && trace_height.is_multiple_of(period),
+                        "periodic column period must be a power of two dividing the trace height"
+                    );
+                    // Row i reads value i mod period, cycling through the p values.
+                    values.extend((0..trace_height).map(|i| col[i % period]));
```
Nashtare: wouldn't it be better to do `extend_from_slice` in a `trace_height / period` loop here?

multi-stark/src/selectors.rs
tcoratger: `periodic_evals_at` now returns `Result`, validating before it indexes anything.

multi-stark/tests/whir_periodic.rs
tcoratger: Just improved the test coverage here

multi-stark/src/rounds.rs
tcoratger: yes done

Nashtare (approved):
Thanks! LGTM :)

PR #1942, opened 2026-07-09
feat(security): add support for LogUp

Nashtare:


security/src/logup.rs
```diff
@@ -0,0 +1,195 @@
...
+/// no interactions (no lookup argument, hence no term).
+pub fn security_term(air: &LogUpAir, shape: &InstanceShape) -> Option<SecurityTerm> {
+    if air.num_interactions == 0 {
+        return None;
+    }
+    Some(SecurityTerm::new(
+        LOGUP_LABEL,
+        fingerprint_error(air, shape),
+    ))
+}
```
tcoratger, line 86: These two methods should be part of `LogUpAir` implementation instead of isolated functions no?

batch-stark/tests/simple.rs
```diff
@@ -838,6 +838,116 @@ fn test_two_instances() -> Result<(), impl Debug> {
...
+        "  without lookups: {:.1} bits (binds: {} in {wo_regime:?})",
+        without.security_bits(),
+        wo_bind.label
+    );
+    println!(
+        "  with lookups:    {:.1} bits (binds: {} in {wi_regime:?})",
+        with.security_bits(),
+        wi_bind.label
+    );
+    println!("  logup fingerprint term: {lookup_bits:.1} bits");
```
tcoratger, line 936: I guess you put those prints for debugging, maybe we want to keep them since this is a parameter crate, not sure?

tcoratger (approved):
LGTM!

batch-stark/tests/simple.rs
Nashtare: It wasn't for debugging, we don't have an workspace example of batch-stark, so this was left so that running the integration test without capture could display the level of security, but I can remove it fully if you prefer.

tcoratger: Not this is totally fine, let us keep that

PR #1944, opened 2026-07-10
feat(encoding): implement batch encoding for zero-knowledge codes

alxkzmn:
## Summary

This PR encodes groups of zero-knowledge mask polynomials with one batched Reed-Solomon DFT instead of encoding each polynomial separately and stacking the resulting codewords.

It adds `encode_batch_with_randomness` with a fallback that encodes each polynomial separately and stacks the results. `ReedSolomonZkEncoding` overrides it with a faster implementation that packs all polynomials as matrix columns before one `dft_batch`. The HVZK sumcheck and hiding WHIR provers now use this batched path for groups of mask polynomials.

The regression tests verify both output equivalence and dispatch behavior: scalar encoding of width \(k\) invokes `dft_batch` \(k\) times, while batched encoding invokes it once.

The resulting encoded matrix is unchanged, so commitments, transcripts, proofs, and verification behavior are unchanged.

## Benchmarks

- Added new benchmarks to the `zk-codes` crate to compare the performance of the batch encoding API against the previous scalar (separate and stack) approach, including a benchmark specifically for the high-folding octic case used in WHIR.
- Extended the WHIR ZK overhead benchmarks to measure the cost of the new batch mask encoding path in realistic scenarios, with PoW disabled to isolate encoding costs.

| Benchmark | Separate encoding | Batched encoding | Reduction |
| --- | ---: | ---: | ---: |
| Mask encoding, width 1 | 0.592 ms | 0.592 ms | 0.0% |
| Mask encoding, width 6 | 3.576 ms | 0.671 ms | 81.2% |
| Mask encoding, width 8 | 4.794 ms | 0.682 ms | 85.8% |
| Hiding WHIR opening, n=18 | 29.783 ms | 16.640 ms | 44.1% |
| Hiding WHIR opening, n=19 | 32.338 ms | 19.043 ms | 41.1% |
| Hiding WHIR opening, n=20 | 39.469 ms | 25.821 ms | 34.6% |


tcoratger (approved):
Thanks a lot!

PR #1945, opened 2026-07-10
test(zk-codes): pin batch-encode equivalence to shipped DFT, cover length mismatch

tcoratger:
## What this is

Small follow-up to #1944, which replaced "encode each ZK mask separately, then column-stack the codewords" with a single batched Reed–Solomon DFT. That PR proved the batched matrix equals the old stacked one. This PR tightens the tests around that proof and removes a tiny inconsistency. No encoding logic changes, so commitments, proofs, and verification are untouched.

## Why it matters

The whole batching optimization rests on one claim: **the batched encoding produces exactly the same matrix as encoding each column on its own.** If that ever stopped being true, commitments would silently diverge. So the claim deserves tests that pin it against the code that actually runs.

Worked example of the claim. Take two length-3 messages and DFT them:

```text
  separate:  DFT(col_0) -> cw_0        batched:  DFT( [col_0 | col_1] )  ->  [cw_0 | cw_1]
             DFT(col_1) -> cw_1
             stack(cw_0, cw_1) = [cw_0 | cw_1]
```

Both must give the identical matrix. Over a finite field the DFT is *exact* (no rounding), so as long as `dft_batch` transforms each column independently, the two paths are bit-identical.

## The three changes

1. **Test the DFT the prover actually ships.** The equivalence test previously used only `Radix2Dit`, the reference DFT. The threaded prover runs `Radix2DFTSmallBatch`. The test now asserts the identity under both backends, so a future change to the shipped DFT that accidentally coupled columns would fail the test immediately. The test body is factored into one helper called with each backend, so there is no duplicated setup.

2. **Cover the length-mismatch guard.** The batch entry point rejects a call where the message count and randomness count disagree. There was a test for the empty-batch panic but none for the mismatch. Added one: two messages, one randomness vector, expect a panic.

3. **Align the two guard orders.** The Reed–Solomon override and the trait default both reject bad batches, but they checked in opposite orders (empty-first vs count-first). On a malformed batch like `messages = []`, `randomness = [x]` they would report different messages. They now check in the same order, so both entry points panic identically.

## Verification

- `cargo test -p p3-zk-codes --lib` -> 21 passed (was 20).
- `cargo fmt` / `cargo clippy -p p3-zk-codes --all-targets` -> clean.


zk-codes/src/reed_solomon.rs
```diff
@@ -525,6 +533,15 @@ mod tests {
         assert_eq!(batched, stacked);
     }
 
+    #[test]
+    fn test_batch_encode_matches_stacked_single_encodes() {
+        // The reference DFT backend.
+        assert_batch_matches_stacked(Radix2Dit::default());
+        // The backend the threaded prover actually ships.
+        // Pinning the identity here guards the production path, not just the reference.
```
Nashtare: That is not correct is it? We are using `ReedSolomonZkEncoding<EF, Radix2Dit<EF>>` as return type

Nashtare (approved):
I think the test comment is erroneous and should be removed (we're testing the proper backend just above), otherwise LGTM

zk-codes/src/reed_solomon.rs
tcoratger:
```suggestion
```

PR #1947, opened 2026-07-11
feat(air,multi-stark)!: public inputs bound by trace position

tcoratger:
Let an AIR name the main-trace cells whose values are public inputs, and have the backend bind them, instead of writing a boundary constraint for each. Additive and opt-in; existing AIRs are unchanged.

### The problem

A zkVM's first-row seed and last-row output are public values.
Today every AIR that has them writes the same boundary constraint by hand:

```
is_first_row * (column - public) = 0
is_last_row  * (column - public) = 0
```

That is boilerplate the AIR should not have to carry, and it is easy to get subtly wrong.

### The change

An AIR declares its public cells by position:

```rust
const FIB_IO_CELLS: [BoundaryPublic; 3] = [
    BoundaryPublic::new(0, BoundaryEnd::First, 0),
    BoundaryPublic::new(1, BoundaryEnd::First, 1),
    BoundaryPublic::new(1, BoundaryEnd::Last,  2),
];
```

The multilinear multi-STARK folder then injects one pin per cell, after the AIR's own constraints and batched with the same alpha, so prover fold and verifier recompute assert exactly the same family.
That pin **is** the binding — it is the boundary constraint the AIR would otherwise have written, moved into the backend.

- **`p3-air`:** `BoundaryEnd`, `BoundaryPublic`, `BoundaryIoError`, a non-generic `boundary::validate`, and a defaulted `BaseAir::public_boundary_io()`. Backward compatible; dependents build unchanged.
- **multi-STARK:** pins injected by both folders, counted by `air_degree` and by the security report, declaration validated by `prove`, `verify` and `security_report`.
- **univariate STARKs:** reject an AIR that lists any cell, at the entry points and at the public building blocks.
- **debug check:** compares each listed cell against the trace.

Constraint-based public inputs remain the default, untouched.

### What this is not

An earlier revision also committed each listed cell as zero and had the verifier add the public value back, following Borgeaud's "handling public inputs".
@Nashtare showed that buys nothing here: the pin already forces the committed cell to zero, so the statement accepted over the blanked commitment is the one accepted over the trace itself, up to a publicly known shift.
Equal soundness, equal proof size, equal opening shape — and it cost a full clone of every committed table carrying a cell.
The zeroing trick pays only when something other than a constraint forces the committed cell to zero. Here nothing does.

That path is gone, along with `Table::poly_mut` and the successor-view Lagrange weight it needed.

### Soundness

- A listed cell is bound by its pin and by nothing else. An AIR that lists a cell the backend ignores has an unbound public input, so a wrapper AIR must forward `public_boundary_io`; a test pins that.
- A pin has the same degree and count as the boundary constraint it replaces, so the per-round sumcheck degree is unchanged. A degree-one AIR that lists a cell is scored at two.
- A declaration naming a column or public value that does not exist is rejected by `prove`, by `verify` and by `security_report`, each against the AIR it is handed.
- An AIR that declares no cells is byte-for-byte unaffected.

### Verification

- `cargo test --workspace --no-fail-fast`, with and without `--features parallel` — all pass.
- `cargo clippy --all-targets` and `cargo fmt --all --check` — clean.
- End-to-end over WHIR: honest round trip, wrong claimed output, malformed declaration, mixed heights, an AIR bound only by pins, an AIR with lookups, a mixed batch through an enum wrapper, and listed cells beside preprocessed and periodic columns.
- At the zerocheck: the pin isolated from the PCS, the degree it forces, and both ends of a two-row trace where the first and last rows are adjacent.

### DIFF
--- a/air/src/air.rs
+++ b/air/src/air.rs
@@ -5,6 +5,46 @@ use p3_matrix::dense::RowMajorMatrix;
 
 use crate::builder::AirBuilder;
 
+/// Which end of the trace a public boundary cell lives on.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum BoundaryEnd {
+    /// The first trace row, the all-zeros hypercube corner (index `0`).
+    First,
+    /// The last trace row, the all-ones hypercube corner (index `2^num_variables - 1`).
+    Last,
+}
+
+/// One main-trace cell exposed as a public input at a boundary corner.
+///
+/// This drives Borgeaud's boundary-IO handling of public inputs:
+/// - the prover commits the column with this cell forced to zero,
+/// - the verifier restores the true value from the public input by a Lagrange-at-corner correction.
+///
+/// A matching corner-zero constraint pins the committed cell to zero.
+/// The restored value is then exactly the public input.
+///
+/// See <https://solvable.group/posts/super-air/> ("Handling public inputs").
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct BoundaryPublic {
+    /// Main-trace column holding the public cell.
+    pub column: usize,
+    /// Trace end the cell sits on.
+    pub end: BoundaryEnd,
+    /// Index into the AIR's public values supplying the cell's value.
+    pub public_value: usize,
+}
+
+impl BoundaryPublic {
+    /// Bundle a column, a trace end, and a public-value index into a boundary cell.
+    pub const fn new(column: usize, end: BoundaryEnd, public_value: usize) -> Self {
+        Self {
+            column,
+            end,
+            public_value,
+        }
+    }
+}
+
 /// The underlying structure of an AIR.
 pub trait BaseAir<F>: Sync {
     /// The number of columns (a.k.a. registers) in this AIR.
@@ -190,6 +230,24 @@ pub trait BaseAir<F>: Sync {
     /// Return the number of expected public values.
     fn num_public_values(&self) -> usize {
         0
+    }
+
+    /// Main-trace boundary cells bound as public inputs via boundary IO.
+    ///
+    /// Each returned cell is committed as zero.
+    /// The verifier restores it from the public input by a Lagrange-at-corner correction.
+    /// See [`BoundaryPublic`] for the mechanism and its soundness pin.
+    ///
+    /// Defaults to none, so public inputs are bound by ordinary constraints unless an AIR opts in.
+    ///
+    /// # Correctness
+    ///
+    /// The multilinear prover asserts a corner-zero pin for each cell.
+    /// A listed cell therefore need not be pinned by the AIR's own evaluation.
+    /// Every referenced column must be a real main column.
+    /// Every public-value index must address a declared public value.
+    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
+        Vec::new()
     }
 }
 
--- a/air/src/air.rs
+++ b/air/src/air.rs
@@ -5,28 +5,30 @@ use p3_matrix::dense::RowMajorMatrix;
 
 use crate::builder::AirBuilder;
 
-/// Which end of the trace a public boundary cell lives on.
+/// Which end of the trace a boundary cell lives on.
+///
+/// These are the two rows a first-row and a last-row selector single out.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum BoundaryEnd {
-    /// The first trace row, the all-zeros hypercube corner (index `0`).
+    /// The first trace row, index `0`.
     First,
-    /// The last trace row, the all-ones hypercube corner (index `2^num_variables - 1`).
+    /// The last trace row, index `height - 1`.
     Last,
 }
 
-/// One main-trace cell exposed as a public input at a boundary corner.
+/// One main-trace cell whose value is a public input, named by its position.
 ///
-/// This drives Borgeaud's boundary-IO handling of public inputs:
-/// - the prover commits the column with this cell forced to zero,
-/// - the verifier restores the true value from the public input by a Lagrange-at-corner correction.
+/// A cell pairs a trace position with one of the AIR's public values:
 ///
-/// A matching corner-zero constraint pins the committed cell to zero.
-/// The restored value is then exactly the public input.
+/// ```text
+///     (column, end) holds public_values[public_value]
+/// ```
 ///
-/// See <https://solvable.group/posts/super-air/> ("Handling public inputs").
+/// This is a declaration, not a constraint.
+/// Binding the cell to the value is the proving backend's job.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct BoundaryPublic {
-    /// Main-trace column holding the public cell.
+    /// Main-trace column holding the cell.
     pub column: usize,
     /// Trace end the cell sits on.
     pub end: BoundaryEnd,
@@ -41,6 +43,29 @@ impl BoundaryPublic {
             column,
             end,
             public_value,
+        }
+    }
+
+    /// Row index this cell sits on in a trace of `height` rows.
+    ///
+    /// ```text
+    ///     first end -> 0
+    ///     last  end -> height - 1
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics when `height` is zero.
+    /// An empty trace has no boundary row to name.
+    #[must_use]
+    pub const fn row(&self, height: usize) -> usize {
+        // A zero-height trace has no row to address at all.
+        assert!(height > 0, "a boundary cell needs at least one trace row");
+
+        // The two ends are the low and high rows of the trace.
+        match self.end {
+            BoundaryEnd::First => 0,
+            BoundaryEnd::Last => height - 1,
         }
     }
 }
@@ -222,6 +247,9 @@ pub trait BaseAir<F>: Sync {
     /// A value that is too small will cause the prover to produce an
     /// invalid proof.
     ///
+    /// The hint covers only what this AIR asserts during its own evaluation.
+    /// A backend that injects extra constraints scores their degree separately.
+    ///
     /// Returns `None` by default, which falls back to symbolic evaluation.
     fn max_constraint_degree(&self) -> Option<usize> {
         None
@@ -232,22 +260,34 @@ pub trait BaseAir<F>: Sync {
         0
     }
 
-    /// Main-trace boundary cells bound as public inputs via boundary IO.
+    /// Main-trace cells whose values are public inputs, named by position.
     ///
-    /// Each returned cell is committed as zero.
-    /// The verifier restores it from the public input by a Lagrange-at-corner correction.
-    /// See [`BoundaryPublic`] for the mechanism and its soundness pin.
+    /// A public input reaches a proof through one of two routes:
     ///
-    /// Defaults to none, so public inputs are bound by ordinary constraints unless an AIR opts in.
+    /// ```text
+    ///     boundary constraint : asserted by the AIR, honored by every backend
+    ///     cell listed here    : bound by the backend, honored by some backends
+    /// ```
+    ///
+    /// Listing a cell is therefore not by itself a binding.
+    /// Support across this workspace:
+    ///
+    /// ```text
+    ///     multilinear multi-STARK : binds every listed cell, needs no AIR constraint
+    ///     univariate STARKs       : reject an AIR that lists any cell
+    ///     debug constraint check  : compares each listed cell against the trace
+    /// ```
+    ///
+    /// The default is the empty slice.
+    /// An AIR that overrides nothing keeps binding its public inputs by constraint.
     ///
     /// # Correctness
     ///
-    /// The multilinear prover asserts a corner-zero pin for each cell.
-    /// A listed cell therefore need not be pinned by the AIR's own evaluation.
-    /// Every referenced column must be a real main column.
-    /// Every public-value index must address a declared public value.
-    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
-        Vec::new()
+    /// - Every column index is less than the main width.
+    /// - Every public-value index is less than the declared public-value count.
+    /// - No two cells name the same column and trace end.
+    fn public_boundary_io(&self) -> &[BoundaryPublic] {
+        &[]
     }
 }
 
--- a/air/src/check_constraints.rs
+++ b/air/src/check_constraints.rs
@@ -9,7 +9,7 @@ use p3_matrix::dense::{RowMajorMatrix, RowMajorMatrixView};
 use p3_matrix::stack::ViewPair;
 
 use crate::{
-    Air, AirBuilder, AirBuilderWithContext, ExtensionBuilder, Name, NamedAirBuilder,
+    Air, AirBuilder, AirBuilderWithContext, BaseAir, ExtensionBuilder, Name, NamedAirBuilder,
     NamedExtensionBuilder, PermutationAirBuilder, RowWindow,
 };
 
@@ -410,6 +410,60 @@ impl<F: Field, EF: ExtensionField<F>> NamedExtensionBuilder for DebugConstraintB
     }
 }
 
+/// Check that every cell an AIR lists as a public boundary input holds its public value.
+///
+/// A listed cell carries no AIR constraint of its own:
+///
+/// ```text
+///     proving backend : binds the cell to the public value
+///     this checker    : compares the two directly
+/// ```
+///
+/// # Panics
+///
+/// Panics when a listed cell names a column outside the trace.
+/// Panics when a listed cell names a public value the caller did not supply.
+/// Panics when a listed cell does not hold its public value.
+fn check_public_boundary_io<F, A>(air: &A, main: &RowMajorMatrix<F>, public_values: &[F])
+where
+    F: Field,
+    A: BaseAir<F>,
+{
+    // Both trace ends are addressed relative to the row count.
+    let height = main.height();
+
+    for cell in air.public_boundary_io() {
+        // A declaration out of range would index past the trace.
+        assert!(
+            cell.column < main.width(),
+            "boundary-IO column {} is out of range for a trace of width {}",
+            cell.column,
+            main.width()
+        );
+
+        // A declaration out of range would index past the caller's public values.
+        assert!(
+            cell.public_value < public_values.len(),
+            "boundary-IO public value {} is out of range for {} public values",
+            cell.public_value,
+            public_values.len()
+        );
+
+        // Read the cell the declaration names.
+        let row = cell.row(height);
+        let value = main
+            .get(row, cell.column)
+            .expect("row and column checked in range");
+
+        // The cell must already carry the public value it is paired with.
+        assert_eq!(
+            value, public_values[cell.public_value],
+            "boundary-IO cell at row {row}, column {} does not hold public value {}",
+            cell.column, cell.public_value
+        );
+    }
+}
+
 /// Evaluate every AIR constraint against a concrete trace and panic on failure.
 ///
 /// The function walks the trace row by row. For each row it:
@@ -421,6 +475,9 @@ impl<F: Field, EF: ExtensionField<F>> NamedExtensionBuilder for DebugConstraintB
 ///    (index with optional label).
 /// 4. Stops at the first row that has at least one violation and panics
 ///    with a summary of every violated constraint on that row.
+///
+/// Listed public boundary cells are checked before the loop.
+/// No AIR constraint asserts them.
 ///
 /// This is the simple variant that does not involve permutation or lookup
 /// arguments. Batch-stark provides its own wrapper that additionally
@@ -442,6 +499,8 @@ where
             height
         );
     }
+
+    check_public_boundary_io(air, main, public_values);
 
     for row_index in 0..height {
         let row_index_next = (row_index + 1) % height;
@@ -524,6 +583,11 @@ where
 ///
 /// This is the simple variant — no permutation or lookup arguments.
 /// Batch-stark provides its own wrapper for those.
+///
+/// # Panics
+///
+/// Panics when a listed public boundary cell does not hold its public value.
+/// A listed cell has no AIR constraint, and therefore no failure slot to record.
 #[allow(unused)] // Suppresses warnings in release mode where this is dead code.
 pub fn check_all_constraints<F, A>(
     air: &A,
@@ -546,6 +610,8 @@ where
             height
         );
     }
+
+    check_public_boundary_io(air, main, public_values);
 
     // Accumulate violations across all rows.
     let mut all_failures = Vec::new();
@@ -634,7 +700,7 @@ mod tests {
     use p3_field::PrimeCharacteristicRing;
 
     use super::*;
-    use crate::{BaseAir, WindowAccess};
+    use crate::{BoundaryEnd, BoundaryPublic, WindowAccess};
 
     /// Minimal AIR for testing transition and boundary constraints.
     ///
@@ -1106,6 +1172,72 @@ mod tests {
 
         // Loop still walked every row.
         assert_eq!(report.total_rows, 4);
+    }
+
+    /// The one cell the constraint-free AIR below lists as a public input.
+    ///
+    /// ```text
+    ///     column 0, last row, carrying public value 0
+    /// ```
+    const BOUNDARY_IO_CELLS: [BoundaryPublic; 1] = [BoundaryPublic::new(0, BoundaryEnd::Last, 0)];
+
+    /// AIR that asserts nothing and binds its only public input by position.
+    ///
+    /// No constraint touches the cell.
+    /// Only the boundary check can catch a mismatch.
+    #[derive(Debug)]
+    struct BoundaryIoAir;
+
+    impl<F: Field> BaseAir<F> for BoundaryIoAir {
+        fn width(&self) -> usize {
+            1
+        }
+
+        fn num_public_values(&self) -> usize {
+            1
+        }
+
+        fn public_boundary_io(&self) -> &[BoundaryPublic] {
+            &BOUNDARY_IO_CELLS
+        }
+    }
+
+    impl<F: Field> Air<DebugConstraintBuilder<'_, F>> for BoundaryIoAir {
+        fn eval(&self, _builder: &mut DebugConstraintBuilder<'_, F>) {
+            // Empty: the public input is bound by position, not by a constraint.
+        }
+    }
+
+    #[test]
+    fn test_boundary_io_cell_matching_its_public_value_passes() {
+        // Invariant: a listed cell holding its public value is accepted.
+        //
+        // Fixture state: 4 rows, 1 column, last row 7.
+        //
+        //     rows          : [4, 5, 6, 7]
+        //     public values : [7]
+        //                     → 7 == 7 → accepted
+        let main = RowMajorMatrix::new([4u32, 5, 6, 7].map(BabyBear::new).to_vec(), 1);
+
+        // Returns cleanly.
+        // A panic would mean a well-formed fixture was rejected.
+        check_constraints(&BoundaryIoAir, &main, &[BabyBear::new(7)]);
+    }
+
+    #[test]
+    #[should_panic(expected = "does not hold public value")]
+    fn test_boundary_io_cell_disagreeing_with_its_public_value_panics() {
+        // Invariant: the check fires even though the AIR asserts nothing at all.
+        //
+        // Mutation: claim a public value the last row does not carry.
+        //
+        //     rows          : [4, 5, 6, 7]
+        //     public values : [8]
+        //                     → 7 != 8 → panic
+        let main = RowMajorMatrix::new([4u32, 5, 6, 7].map(BabyBear::new).to_vec(), 1);
+
+        // Expected: panic before the row loop starts.
+        check_constraints(&BoundaryIoAir, &main, &[BabyBear::new(8)]);
     }
 
     #[test]
--- a/multi-stark/benches/zerocheck.rs
+++ b/multi-stark/benches/zerocheck.rs
@@ -1,7 +1,7 @@
 use core::borrow::Borrow;
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
-use p3_air::{Air, AirBuilder, BaseAir, WindowAccess};
+use p3_air::{Air, AirBuilder, BaseAir, BoundaryEnd, BoundaryPublic, WindowAccess};
 use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
 use p3_challenger::DuplexChallenger;
 use p3_field::PrimeCharacteristicRing;
@@ -130,6 +130,40 @@ fn wide_fib_trace(n: usize) -> RowMajorMatrix<F> {
     RowMajorMatrix::new(values, WIDE_COLS)
 }
 
+/// Fibonacci AIR binding its public inputs by boundary IO instead of constraints.
+///
+/// The transition recurrence is asserted here.
+/// The two seeds and the output are declared as public boundary cells.
+/// The prover asserts corner-zero pins in place of first/last-row constraints.
+struct FibIoAir;
+
+impl<X> BaseAir<X> for FibIoAir {
+    fn width(&self) -> usize {
+        NUM_COLS
+    }
+    fn num_public_values(&self) -> usize {
+        3
+    }
+    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
+        vec![
+            BoundaryPublic::new(0, BoundaryEnd::First, 0),
+            BoundaryPublic::new(1, BoundaryEnd::First, 1),
+            BoundaryPublic::new(1, BoundaryEnd::Last, 2),
+        ]
+    }
+}
+
+impl<AB: AirBuilder> Air<AB> for FibIoAir {
+    fn eval(&self, builder: &mut AB) {
+        let main = builder.main();
+        let local: &FibRow<AB::Var> = main.current_slice().borrow();
+        let next: &FibRow<AB::Var> = main.next_slice().borrow();
+        let mut trans = builder.when_transition();
+        trans.assert_eq(local.right, next.left);
+        trans.assert_eq(local.left + local.right, next.right);
+    }
+}
+
 fn fresh_challenger() -> Ch {
     let mut rng = SmallRng::seed_from_u64(0xC0FFEE);
     let perm = Perm::new_from_rng_128(&mut rng);
@@ -203,5 +237,58 @@ fn bench_wide_zerocheck(c: &mut Criterion) {
     group.finish();
 }
 
-criterion_group!(benches, bench_zerocheck, bench_wide_zerocheck);
+/// Compare the two ways to bind a Fibonacci AIR's public inputs, at equal size.
+///
+/// - `constraint`: first/last-row boundary constraints on the committed trace.
+/// - `boundary_io`: corner-zero pins on the same trace, folded as in the real prover.
+///
+/// Both assert the same number of degree-2 constraints, so this isolates the
+/// per-round fold cost of the two bindings.
+/// The one extra cost of boundary IO, a single trace clone before commitment,
+/// lives in the committed prover and is not exercised here.
+fn bench_public_input_binding(c: &mut Criterion) {
+    let mut group = c.benchmark_group("public_input_binding");
+    group.sample_size(10);
+    for log_height in [18, 20] {
+        // A shared Fibonacci trace and its public seeds and output.
+        let n = 1 << log_height;
+        let trace = fib_trace(n);
+        let last = trace.values[(n - 1) * NUM_COLS + 1];
+        let pis = [F::ZERO, F::ONE, last];
+        let table = Table::new(trace.transpose());
+        let public_values = vec![&pis[..]];
+        let preprocessed = vec![None; 1];
+        let tables = vec![&table; 1];
+
+        // Constraint binding: the AIR asserts the boundary equalities directly.
+        let constraint_air = FibAir;
+        let constraint_airs = vec![&constraint_air];
+        let constraint = AirZerocheck::new(&constraint_airs, 0);
+        group.bench_with_input(BenchmarkId::new("constraint", log_height), &n, |b, _| {
+            b.iter(|| {
+                let mut ch = fresh_challenger();
+                constraint.prove::<F, EF, _>(&preprocessed, &tables, &public_values, &mut ch)
+            });
+        });
+
+        // Boundary-IO binding: the folder asserts corner-zero pins instead.
+        let io_air = FibIoAir;
+        let io_airs = vec![&io_air];
+        let io = AirZerocheck::new(&io_airs, 0);
+        group.bench_with_input(BenchmarkId::new("boundary_io", log_height), &n, |b, _| {
+            b.iter(|| {
+                let mut ch = fresh_challenger();
+                io.prove::<F, EF, _>(&preprocessed, &tables, &public_values, &mut ch)
+            });
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(
+    benches,
+    bench_zerocheck,
+    bench_wide_zerocheck,
+    bench_public_input_binding
+);
 criterion_main!(benches);
--- a/multi-stark/benches/zerocheck.rs
+++ b/multi-stark/benches/zerocheck.rs
@@ -130,12 +130,24 @@ fn wide_fib_trace(n: usize) -> RowMajorMatrix<F> {
     RowMajorMatrix::new(values, WIDE_COLS)
 }
 
-/// Fibonacci AIR binding its public inputs by boundary IO instead of constraints.
+/// Fibonacci AIR binding its public inputs by position instead of by constraint.
 ///
-/// The transition recurrence is asserted here.
-/// The two seeds and the output are declared as public boundary cells.
-/// The prover asserts corner-zero pins in place of first/last-row constraints.
+/// Only the transition recurrence is asserted here.
+/// The folder injects one pin per declared cell in place of the boundary constraints.
 struct FibIoAir;
+
+/// The seed and output cells the AIR above binds by position.
+///
+/// ```text
+///     column 0, first row -> public value 0
+///     column 1, first row -> public value 1
+///     column 1, last  row -> public value 2
+/// ```
+const FIB_IO_CELLS: [BoundaryPublic; 3] = [
+    BoundaryPublic::new(0, BoundaryEnd::First, 0),
+    BoundaryPublic::new(1, BoundaryEnd::First, 1),
+    BoundaryPublic::new(1, BoundaryEnd::Last, 2),
+];
 
 impl<X> BaseAir<X> for FibIoAir {
     fn width(&self) -> usize {
@@ -144,20 +156,22 @@ impl<X> BaseAir<X> for FibIoAir {
     fn num_public_values(&self) -> usize {
         3
     }
-    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
-        vec![
-            BoundaryPublic::new(0, BoundaryEnd::First, 0),
-            BoundaryPublic::new(1, BoundaryEnd::First, 1),
-            BoundaryPublic::new(1, BoundaryEnd::Last, 2),
-        ]
+    fn public_boundary_io(&self) -> &[BoundaryPublic] {
+        &FIB_IO_CELLS
     }
 }
 
 impl<AB: AirBuilder> Air<AB> for FibIoAir {
     fn eval(&self, builder: &mut AB) {
+        // Read the current row and the row after it.
         let main = builder.main();
         let local: &FibRow<AB::Var> = main.current_slice().borrow();
         let next: &FibRow<AB::Var> = main.next_slice().borrow();
+
+        // Advance the recurrence on every row but the last.
+        //
+        //     next.left  = right
+        //     next.right = left + right
         let mut trans = builder.when_transition();
         trans.assert_eq(local.right, next.left);
         trans.assert_eq(local.left + local.right, next.right);
@@ -239,13 +253,16 @@ fn bench_wide_zerocheck(c: &mut Criterion) {
 
 /// Compare the two ways to bind a Fibonacci AIR's public inputs, at equal size.
 ///
-/// - `constraint`: first/last-row boundary constraints on the committed trace.
-/// - `boundary_io`: corner-zero pins on the same trace, folded as in the real prover.
+/// ```text
+///     constraint  : first-row and last-row boundary constraints
+///     boundary_io : one injected pin per declared cell
+/// ```
 ///
-/// Both assert the same number of degree-2 constraints, so this isolates the
-/// per-round fold cost of the two bindings.
-/// The one extra cost of boundary IO, a single trace clone before commitment,
-/// lives in the committed prover and is not exercised here.
+/// Both arms assert the same number of degree-two constraints.
+/// This therefore isolates the per-round fold cost of the two bindings.
+///
+/// Binding by position also clones each committed table that carries a cell.
+/// That cost lives in the committed prover and is not exercised here.
 fn bench_public_input_binding(c: &mut Criterion) {
     let mut group = c.benchmark_group("public_input_binding");
     group.sample_size(10);
@@ -271,7 +288,7 @@ fn bench_public_input_binding(c: &mut Criterion) {
             });
         });
 
-        // Boundary-IO binding: the folder asserts corner-zero pins instead.
+        // Position binding: the folder injects one pin per declared cell instead.
         let io_air = FibIoAir;
         let io_airs = vec![&io_air];
         let io = AirZerocheck::new(&io_airs, 0);
--- a/multi-stark/src/boundary.rs
+++ b/multi-stark/src/boundary.rs
@@ -1,0 +1,265 @@
+//! Borgeaud boundary-IO handling of public inputs.
+//!
+//! A public boundary cell is committed as zero and restored at verification time
+//! from the public input, so the commitment never carries the public data.
+//!
+//! ```text
+//!     committed = true column with its corner cell set to 0
+//!     restored  = committed(point) + eq_corner(point) * public
+//! ```
+//!
+//! The prover folds the true table, so the sumcheck transcript reflects the restored values.
+//! A corner-zero pin, asserted alongside the AIR constraints, forces the committed corner to zero.
+//! The restored value is then exactly the public input.
+//!
+//! See <https://solvable.group/posts/super-air/> ("Handling public inputs").
+
+use alloc::vec::Vec;
+
+use p3_air::{BoundaryEnd, BoundaryPublic};
+use p3_field::{ExtensionField, Field};
+use p3_matrix::dense::RowMajorMatrix;
+use p3_sumcheck::layout::Table;
+
+use crate::selectors::BoundaryEvals;
+
+/// The public boundary cells of one AIR, and the two halves of their handling.
+///
+/// One half runs on the prover and blanks each public cell before commitment.
+/// The other runs on the verifier and restores the true opened values from the public inputs.
+pub(crate) struct BoundaryIo<'a>(&'a [BoundaryPublic]);
+
+impl<'a> BoundaryIo<'a> {
+    /// Wrap an AIR's declared public boundary cells.
+    pub(crate) const fn new(cells: &'a [BoundaryPublic]) -> Self {
+        Self(cells)
+    }
+
+    /// Copy of a table with every public boundary cell set to zero.
+    ///
+    /// The prover commits this copy but folds the original true table.
+    /// The verifier's reconstruction then lands exactly on the folded values.
+    ///
+    /// # Arguments
+    ///
+    /// - `table`: the true, transposed trace table, one row per column.
+    ///
+    /// # Returns
+    ///
+    /// A fresh table equal to the input with each declared corner cell replaced by zero.
+    pub(crate) fn commit_zeroed<F: Field>(&self, table: &Table<F>) -> Table<F> {
+        // Row count of the trace; each column is one contiguous row-major span of this length.
+        let height = 1 << table.num_variables();
+
+        // Copy every column into a fresh buffer: one row per column, `height` entries each.
+        let mut values = Vec::with_capacity(table.num_polys() * height);
+        for column in table.iter_polys() {
+            values.extend_from_slice(column);
+        }
+
+        // Blank each declared corner cell in place.
+        //   first end -> row 0
+        //   last  end -> row height - 1
+        for cell in self.0 {
+            let row = match cell.end {
+                BoundaryEnd::First => 0,
+                BoundaryEnd::Last => height - 1,
+            };
+            values[cell.column * height + row] = F::ZERO;
+        }
+
+        Table::new(RowMajorMatrix::new(values, height))
+    }
+
+    /// Restore the true opened values from the committed corner-zeroed openings.
+    ///
+    /// The committed column is the true column with each public cell zeroed.
+    /// Their multilinear extensions differ by one Lagrange term per cell:
+    /// ```text
+    ///     P(point) = committed(point) + eq_corner(point) * public
+    /// ```
+    ///
+    /// - The current-row value adds the corner weight.
+    /// - The successor value adds the corner's successor weight.
+    /// - The first row has successor weight zero, so only the last row corrects the successor.
+    ///
+    /// # Arguments
+    ///
+    /// - `current`: opened current-row value of each column, in column order.
+    /// - `next`: opened successor value of each next-row column, aligned with `next_columns`.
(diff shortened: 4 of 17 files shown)


air/src/air.rs
```diff
@@ -191,6 +231,24 @@ pub trait BaseAir<F>: Sync {
...
         0
     }
+
+    /// Main-trace boundary cells bound as public inputs via boundary IO.
+    ///
+    /// Each returned cell is committed as zero.
+    /// The verifier restores it from the public input by a Lagrange-at-corner correction.
+    /// See [`BoundaryPublic`] for the mechanism and its soundness pin.
+    ///
+    /// Defaults to none, so public inputs are bound by ordinary constraints unless an AIR opts in.
```
Nashtare: doc is confusing, `none` may refer to an `Option<T>` but the type is `Vec<_>`

air/src/air.rs
```diff
@@ -191,6 +231,24 @@ pub trait BaseAir<F>: Sync {
...
+    ///
+    /// # Correctness
+    ///
+    /// The multilinear prover asserts a corner-zero pin for each cell.
+    /// A listed cell therefore need not be pinned by the AIR's own evaluation.
+    /// Every referenced column must be a real main column.
+    /// Every public-value index must address a declared public value.
+    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
+        Vec::new()
+    }
```
Nashtare: it's kinda weird on several fronts. It's defined in `p3-air` but the doc is explicitely tailored for multi-stark which is a separate independent crate, and only that crate honors it (there's no boundary check in `check_constraints` for instance for uni/batch-stark.

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,265 @@
...
+                next[slot] += successor_weight * public;
+            }
+        }
+
+        (current, next)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Nashtare, line 317: The tests you added aren't exercising the attack vector I think you wanted to catch with the introduction of this logic. If we were to delete/skip `eval_boundary_io` and `boundary_io_degree`, all three tests would still behave identically.

Nashtare: We'd need to add a test that upon malicious proving committing to a nonzero corner, we'd effectively have a triggered assert

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        // Blank each declared corner cell in place.
+        //   first end -> row 0
+        //   last  end -> row height - 1
+        for cell in self.0 {
+            let row = match cell.end {
+                BoundaryEnd::First => 0,
+                BoundaryEnd::Last => height - 1,
+            };
+            values[cell.column * height + row] = F::ZERO;
```
Nashtare: nit: we're never checking boundary accesses like `column < width`, `public_value < num_public_values()` so we may have weird panics upon invalid accesses. I think we'd need to validate at setup (also because the verifier doc says it won't panic)

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// The verifier's reconstruction then lands exactly on the folded values.
+    ///
+    /// # Arguments
+    ///
+    /// - `table`: the true, transposed trace table, one row per column.
+    ///
+    /// # Returns
+    ///
+    /// A fresh table equal to the input with each declared corner cell replaced by zero.
+    pub(crate) fn commit_zeroed<F: Field>(&self, table: &Table<F>) -> Table<F> {
```
Nashtare: I think we'd need to reject duplicates where the cells are declared. Right now `commit_zeroed` zeroes a (column, end) once and `reconstruct` adds a correction per declared cell. Two cells on the same corner would make the verifier reconstructs `0 + pub_a + pub_b` while the folder pins the corner to both `pub_a` and `pub_b`, so the proof cannot verify even if the prover is honest.

multi-stark/src/prover.rs
```diff
@@ -117,13 +117,46 @@ where
...
+    // With boundary IO these are the true tables kept aside above.
+    // Otherwise they are the committed views, which equal the input tables.
+    let tables = true_tables.as_ref().map_or_else(
+        || {
+            (0..num_instances)
+                .map(|table_index| config.committed_table(&prover_data, table_index))
+                .collect::<Vec<_>>()
+        },
+        |true_tables| true_tables.iter().collect::<Vec<_>>(),
+    );
```
Nashtare: this is somewhat weird to me to have this case distinction. I liked that we had "what was committed" to match by construction "what was folded", but now correctness now depends on `build_witness` performing no other transformation than the zeroing which seems like a risky invariant (or at least we'd need to make it explicit)

multi-stark/src/zerocheck.rs
```diff
@@ -177,11 +188,11 @@ where
             degree >= symbolic_constraint_degree(),
             "max_constraint_degree hint is below the symbolic constraint degree"
         );
-        return degree;
+        return degree.max(boundary_io_degree);
```
Nashtare: This is effectively overriding the degree hint which feels flaky. At least we'd need to make it clear in the hint doc that this does not take into account the `boundary_io_degree`

multi-stark/src/zerocheck.rs
```diff
@@ -177,11 +188,11 @@ where
...
     }
 
     // No hint: fall back to the symbolic constraint degree.
-    symbolic_constraint_degree()
+    symbolic_constraint_degree().max(boundary_io_degree)
```
Nashtare: Same for this function

air/src/air.rs
```diff
@@ -191,6 +231,24 @@ pub trait BaseAir<F>: Sync {
...
+    ///
+    /// Defaults to none, so public inputs are bound by ordinary constraints unless an AIR opts in.
+    ///
+    /// # Correctness
+    ///
+    /// The multilinear prover asserts a corner-zero pin for each cell.
+    /// A listed cell therefore need not be pinned by the AIR's own evaluation.
+    /// Every referenced column must be a real main column.
+    /// Every public-value index must address a declared public value.
+    fn public_boundary_io(&self) -> Vec<BoundaryPublic> {
```
Nashtare: This is called in hot loops, so the Vec may not be ideal. Is there a way to hoist it?

multi-stark/src/prover.rs
```diff
@@ -117,13 +117,46 @@ where
...
+    // With boundary IO the committed copy carries the public corner cells as zero.
+    // The zerocheck still folds the true tables, kept aside here.
+    // The verifier's reconstruction then lands exactly on the folded values.
+    let (commitment, prover_data, true_tables) = if has_boundary_io {
+        // Blank the public corner cells of each table before committing.
+        let committed = tables
+            .iter()
+            .zip(boundary_io.iter())
+            .map(|(table, cells)| crate::boundary::BoundaryIo::new(cells).commit_zeroed(table))
+            .collect::<Vec<_>>();
```
Nashtare: it feels particularly heavy no? That basically means every table is copied even if there are no boundary cells in them

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,265 @@
...
+        // Row count of the trace; each column is one contiguous row-major span of this length.
+        let height = 1 << table.num_variables();
+
+        // Copy every column into a fresh buffer: one row per column, `height` entries each.
+        let mut values = Vec::with_capacity(table.num_polys() * height);
+        for column in table.iter_polys() {
```
Nashtare:
```suggestion
        for column in table.par_iter_polys() {
```

multi-stark/src/zerocheck.rs
```diff
@@ -157,6 +157,17 @@ where
 {
     let layout = AirLayout::from_air::<F>(air);
 
+    // Boundary-IO public cells each add one corner-zero pin.
+    // The pins are asserted while the constraints run, but the symbolic pass never sees them.
+    // Fold their degree in by hand.
+    // A pin is `selector * (column - public)`.
+    // The selector and the column each score degree one at domain size two, so the product scores two.
+    let boundary_io_degree = if air.public_boundary_io().is_empty() {
```
Nashtare: if we change the signature to not allocate a vec this would also become cheaper

air/src/air.rs
```diff
@@ -5,6 +5,46 @@ use p3_matrix::dense::RowMajorMatrix;
...
+    Last,
+}
+
+/// One main-trace cell exposed as a public input at a boundary corner.
+///
+/// This drives Borgeaud's boundary-IO handling of public inputs:
+/// - the prover commits the column with this cell forced to zero,
+/// - the verifier restores the true value from the public input by a Lagrange-at-corner correction.
+///
+/// A matching corner-zero constraint pins the committed cell to zero.
```
Nashtare: This reads weirdly, this is nothing enforced by the current crate but by the folder in `p3-multi-stark`. See comment about `public_boundary_io()` below

Nashtare (changes requested):
A couple problematic issues.
Also the description says `An AIR that declares no boundary cells is byte-for-byte unaffected.` but this is incorrect I believe, and the verifier actually currently pays a price even without boundary cells

air/src/air.rs
tcoratger: Fixed

air/src/air.rs
tcoratger: Just fixed that

multi-stark/src/boundary.rs
tcoratger: Just added two tests like this: `boundary_io_pin_binds_the_public_value_to_the_folded_trace` and `boundary_io_pin_raises_the_sumcheck_degree`.

multi-stark/src/boundary.rs
tcoratger: Just added `boundary::validate(air)`

multi-stark/src/boundary.rs
tcoratger: added `BoundaryIoError::DuplicateCell { column, end }`

multi-stark/src/prover.rs
tcoratger: Fixed

multi-stark/src/zerocheck.rs
tcoratger: Fixed

multi-stark/src/zerocheck.rs
tcoratger: Fixed

air/src/air.rs
tcoratger: Fixed with `&[BoundaryPublic]`

multi-stark/src/prover.rs
tcoratger: Fixed so that no table is copied now

multi-stark/src/boundary.rs
tcoratger: Fixed

multi-stark/src/zerocheck.rs
tcoratger: Done via the &[BoundaryPublic]

air/src/air.rs
tcoratger: fixed

tcoratger:
> A couple problematic issues. Also the description says `An AIR that declares no boundary cells is byte-for-byte unaffected.` but this is incorrect I believe, and the verifier actually currently pays a price even without boundary cells

Thanks a lot for the review, I think that I fixed all the problems, please let me know if anything else is missing? :)

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,640 @@
...
+//!
+//! A public boundary cell is committed as zero.
+//! The verifier adds the public value back from its own copy:
+//!
+//! ```text
+//!     committed = true column with its cell set to 0
+//!     restored  = committed(point) + eq_cell(point) * public
+//! ```
+//!
+//! The commitment therefore never carries data the verifier already holds.
```
Nashtare: **HIGH (design): once the pin exists, blanking the cell and restoring it on the verifier no longer buys anything.**

The verifier checks the AIR and the pins over the restored column `t = c' + Σ public·δ_cell`, where `c'` is the committed column.
- The restore is linear, so it holds for the successor view too.
- `c'` and the public values are both fixed before any challenge.
- The pin forces `t[cell] = public`, i.e. `c'[cell] = 0`.

The accepted statement is therefore

```text
∃ t : AIR(t) ∧ t[cell] = public ∧ commitment opens to t − Σ public·δ_cell
```

Committing `t` itself and injecting the same pins gives

```text
∃ t : AIR(t) ∧ t[cell] = public ∧ commitment opens to t
```

The two commitments differ by a known public shift, so the two versions are equally sound. Nothing else changes:
- same constraint count and degree (the description already says so);
- same PCS shape, openings and proof size;
- lookups read the restored tables in both.

Nothing else is gained either. Public values need no hiding, and the commitment can't be reused across public inputs, because the rest of the trace depends on them. The zeroing trick from the post is only cheaper when nothing has to force the committed cell to zero. Here the pin is that force, and it is exactly the boundary constraint the trick was meant to replace.

Checked at `a6af7781`:
- With `take_cells`/`restore_cells` disabled in `prover.rs` and `reconstruct` given no cells in `verifier.rs`, pins kept, all 186 `p3-multi-stark` tests pass, including every rejection test and both pin-attack tests.
- End-to-end prove time is the same either way: 2^18 rows × 32 columns, release, 3.568 s with boundary constraints vs 3.566 s with boundary IO.

What blanking does cost:
- `prover.rs:233` clones every committed table with a cell and keeps the copy through the lookup and zerocheck phases (200 columns × 2^22 BabyBear rows ≈ 3.1 GiB).
- The blank/restore asserts make completeness depend on `build_witness`/`committed_table` leaving cells in place. That is the invariant from the previous round, now asserted rather than removed.
- The verifier pays `Cow` copies and Lagrange corrections, and the PR adds `last_row_successor_weight` and a public `Table::poly_mut`.
- The main commitment no longer commits to the trace itself, which matters for anything that links a main commitment to a trace.

Suggestion: keep the declaration API (`BoundaryEnd`, `BoundaryPublic`, `BaseAir::public_boundary_io`), `validate`, the injected pins, the degree and security accounting, the backend guards and the debug check. Drop `take_cells`, `restore_cells`, `reconstruct`, `last_row_successor_weight` and `poly_mut`. The AIR still declares its public cells and the backend binds them, without the clone or most of this module.

If blanking buys a property I'm missing, can you name it and add a test that fails without blanking? None of the current tests does.

multi-stark/src/prover.rs
```diff
@@ -190,9 +213,45 @@ where
...
+    //     restored = committed + sum_cells eq_cell * public       (the verifier's side)
+    //
+    // The pin the folder asserts forces those two lines to agree.
+    let restored = boundary_io
+        .iter()
+        .zip(cell_values.iter())
+        .enumerate()
+        .map(|(table_index, (boundary, values))| {
+            (!boundary.is_empty()).then(|| {
+                let mut table = config.committed_table(&prover_data, table_index).clone();
```
Nashtare: **MEDIUM (only if blanking stays): tables with a cell are still copied in full.**

The previous round's copy now happens only for tables that declare a cell, but it still clones every column of those tables. At 200 columns × 2^22 BabyBear rows that is about 3.1 GiB, held next to `prover_data` through the lookup and zerocheck phases.

If blanking stays, restoring the few cells in place in the committed table and re-blanking them before `open_at` would avoid the copy. That needs mutable access to the committed table from the config. Dropping blanking (comment on `boundary.rs`) removes the copy entirely.

multi-stark/src/verifier.rs
```diff
@@ -46,6 +47,14 @@ where
     /// A statement-level transcript step could not be replayed.
     #[error("transcript: {0}")]
     Transcript(MultiStarkTranscriptFailure),
+    /// An AIR names a public boundary cell or public value it does not have.
+    #[error("instance {instance} boundary IO: {error}")]
+    BoundaryIo {
```
Nashtare, line 52: **MEDIUM: new variant on a public enum that isn't `#[non_exhaustive]`, so the title needs `!`.**

Downstream exhaustive `match`es on `VerificationError` stop compiling. Earlier variants on this same enum landed that way: `Transcript` in #2102 (`feat(multi-stark)!`) and `Security` in #2100 (`fix(security)!`).

air/src/air.rs
```diff
@@ -189,6 +257,36 @@ pub trait BaseAir<F>: Sync {
...
+    /// Support across this workspace:
+    ///
+    /// ```text
+    ///     multilinear multi-STARK : binds every listed cell, needs no AIR constraint
+    ///     univariate STARKs       : reject an AIR that lists any cell
+    ///     debug constraint check  : compares each listed cell against the trace
+    /// ```
+    ///
+    /// The default is the empty slice.
+    /// An AIR that overrides nothing keeps binding its public inputs by constraint.
```
Nashtare, line 216: **MEDIUM: with an empty default, a wrapper that forgets to forward this method silently unbinds its public inputs.**

Wrapper and enum AIRs forward `BaseAir` by hand (e.g. `MixedAir` in `multi-stark/src/zerocheck/mod.rs`). If a wrapper forwards `width` and `num_public_values` but not `public_boundary_io`:
- the prover commits the cells unblanked and injects no pins, and the verifier applies no correction;
- nothing notices — not the backend guards, not `validate`, not `check_constraints` — because an empty list is valid.

The other defaults fail safe (`main_next_row_columns` defaults to every column) or fail loudly (a wrong `num_public_values` is caught). This line reads as if not overriding is always safe. Please say that wrappers must forward the method. An end-to-end test through an enum wrapper would pin that down.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1122,3 +1175,553 @@ fn generate_whir_fixture() -> Result<(), Box<dyn std::error::Error>> {
     write_fixture(WHIR_FIXTURE, &bytes)?;
     Ok(())
 }
+
+#[test]
+fn prove_verify_fibonacci_boundary_io_roundtrips() {
```
Nashtare, line 1175: **MEDIUM (tests): a few paths the integration tests don't reach yet.**

- **Mixed batch:** one AIR with cells and one without, in the same batch (needs an enum AIR). This exercises the `Some`/`None` mix in `prover.rs:227-255`, `Borrowed` next to `Owned` in the verifier, and the forwarding noted on `air.rs`.
- **Preprocessed or periodic columns plus cells:** the fixtures already exist in `whir_preprocessed.rs` and `whir_periodic.rs`.
- **Binary PCS:** it allows height-2 traces, where the first and last rows are adjacent. A local probe over GF(2^128) passed honest proofs and rejected tampered claims, so this is cheap to land.
- **Blanking itself:** nothing observes it (comment on `boundary.rs`).

multi-stark/src/security.rs
```diff
@@ -216,11 +216,15 @@ where
...
         }
-        let constraints = builder
+        let own_constraints = builder
             .base_constraints()
             .len()
             .checked_add(builder.extension_constraints().len())
             .ok_or_else(|| invalid("constraint count overflow"))?;
+        // The folder batches one public boundary pin per listed cell with the AIR's own constraints.
+        let constraints = own_constraints
+            .checked_add(air.public_boundary_io().len())
```
Nashtare: **LOW: the security report counts declared cells without validating them.**

The successor-column check a few lines up returns `InvalidShape`. A cell with an out-of-range column or public value, or a duplicate cell, is still counted here, and the report returns a security level for a statement `verify` rejects. `prove_with_security` then passes the assessment and panics at `prove`'s `validate(...).expect`.

Calling `boundary::validate` here and mapping the error to `InvalidShape` keeps the report fail-closed like the checks around it.

multi-stark/src/keys.rs
```diff
@@ -73,6 +74,13 @@ where
     A: BaseAir<C::Val>,
     Commitment<C>: Clone,
 {
+    // A public boundary declaration is fixed by the AIR, never by the witness.
+    // Checking it once here spares prover and verifier from re-deriving it per proof.
```
Nashtare: **LOW: keys don't carry the AIR, so this check is advisory, and it panics inside a `Result`-returning `setup`.**

`verify_rejects_an_invalid_boundary_io_declaration` hands the verifier a different AIR than setup saw, so `prove` (`prover.rs:172`) and `verify` (`verifier.rs:201`) validate again. The "checked once" claim also appears in:
- `boundary.rs:74` ("Checking it once at setup therefore costs nothing per proof");
- `boundary.rs:162` and `boundary.rs:255` ("Setup rejects such a declaration before either party reaches this point").

Either return an error from `setup`, or drop the setup check and reword those docs to describe per-call validation. The prover's panic message also lacks the instance index this one includes.

air/src/check_constraints.rs
```diff
@@ -524,6 +583,11 @@ where
 ///
 /// This is the simple variant — no permutation or lookup arguments.
 /// Batch-stark provides its own wrapper for those.
+///
+/// # Panics
+///
+/// Panics when a listed public boundary cell does not hold its public value.
+/// A listed cell has no AIR constraint, and therefore no failure slot to record.
```
Nashtare: **LOW: `check_all_constraints` is the variant that collects failures, but a cell mismatch here panics instead of being reported.**

A mismatch could be recorded as a `ConstraintFailure` on `cell.row(height)`, with `constraint` indices after the AIR's own (the order the folder injects the pins). Panics would then be kept only for out-of-range declarations.

uni-stark/src/prover.rs
```diff
@@ -106,6 +106,13 @@ where
     #[cfg(debug_assertions)]
     p3_air::check_constraints(air, &trace, public_values);
 
+    // Public inputs reach this proof only through AIR constraints.
+    // A cell listed for backend binding would go completely unbound.
+    assert!(
+        air.public_boundary_io().is_empty(),
+        "uni-stark does not support boundary-IO public values; bind them with AIR constraints"
```
Nashtare, line 110: **LOW: the debug `check_constraints` above runs before this assert.**

In a debug build, an AIR with cells walks the whole trace first. If a cell is wrong, it panics with "does not hold public value" instead of this message. batch-stark asserts first; moving this assert above the debug check would match.

uni-stark/src/verifier.rs
```diff
@@ -427,6 +427,13 @@ where
     SC::Challenger: GrindingChallenger<Witness = Val<SC>>,
     A: Air<SymbolicAirBuilder<Val<SC>>> + for<'a> Air<VerifierConstraintFolder<'a, SC>>,
 {
+    // Public inputs reach this proof only through AIR constraints.
+    // A cell listed for backend binding would go completely unbound.
+    assert!(
+        air.public_boundary_io().is_empty(),
+        "uni-stark does not support boundary-IO public values; bind them with AIR constraints"
```
Nashtare, line 441: **LOW: the guard covers the entry points, but not the public building blocks.**

These evaluate constraints with no check:
- `uni_stark::verify_constraints` (`verifier.rs:142`)
- `uni_stark::quotient_values` (`prover.rs:517`)
- batch-stark's `quotient_values` (`prover.rs:823`)
- `VerifierData::verify_constraints_with_lookups` (`verifier/data.rs:53`)

A verifier assembled from them accepts any value in a declared cell. Nothing in-tree does that, but an assert or a doc line on each would close it. The support table in `air.rs` ("univariate STARKs: reject an AIR that lists any cell") is also only true at the entry points.

multi-stark/src/boundary.rs
```diff
@@ -0,0 +1,640 @@
...
+///
+/// A declaration is fixed by the AIR, never by the witness.
+/// Checking it once at setup therefore costs nothing per proof.
+///
+/// # Errors
+///
+/// - A cell names a column outside the main trace.
+/// - A cell names a public value the AIR does not declare.
+/// - Two cells name the same trace cell.
+pub fn validate<F, A>(air: &A) -> Result<(), BoundaryIoError>
```
Nashtare: **LOW: `F` only appears in a bound, so every call site spells `validate::<C::Val, _>`.**

`check_public_boundary_io` in `air/src/check_constraints.rs` repeats the range checks but not the duplicate check. A non-generic `validate(cells, width, num_public_values)` in `p3-air` could serve both.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1122,3 +1175,553 @@ fn generate_whir_fixture() -> Result<(), Box<dyn std::error::Error>> {
...
+        )]),
+        &proof,
+        0,
+        &mut challenger(),
+    )
+    .unwrap_err();
+    match err {
+        VerificationError::Opening(WhirVerifierError::MerkleProofInvalid { position, reason }) => {
+            assert_eq!(position, 0);
+            assert_eq!(reason, "Base field Merkle multiproof verification failed");
```
Nashtare: **LOW: this rejection comes from the transcript, not from boundary IO.**

Public values are absorbed before the zerocheck. Changing one at verify time sends the verifier down a different transcript, and the opening fails before any correction is applied. An AIR with no cells rejects the same way (`verify_rejects_tampered_public_values`), and pinning the WHIR error string ties the test to PCS internals.

`verify_rejects_wrong_claimed_output_boundary_io` below already covers the boundary-IO rejection. This test could go, or match only `Opening(MerkleProofInvalid { .. })`.

multi-stark/benches/zerocheck.rs
```diff
@@ -203,5 +251,61 @@ fn bench_wide_zerocheck(c: &mut Criterion) {
...
+///     constraint  : first-row and last-row boundary constraints
+///     boundary_io : one injected pin per declared cell
+/// ```
+///
+/// Both arms assert the same number of degree-two constraints.
+/// This therefore isolates the per-round fold cost of the two bindings.
+///
+/// Binding by position also clones each committed table that carries a cell.
+/// That cost lives in the committed prover and is not exercised here.
+fn bench_public_input_binding(c: &mut Criterion) {
```
Nashtare: **NIT: both arms run the same workload.**

As the doc above says, both assert the same degree-2 constraints, and `AirZerocheck::prove` never runs blanking, the clone or the verifier correction. So this benchmark can't show a difference between the two bindings. It could be dropped, or moved to the full `prove()` where blanking lives.

`FibIoAir` is also defined three times (here, `zerocheck/mod.rs`, `tests/whir_fibonacci.rs`). Parameterising by cells, like `PermutationSumAir { cells }` in `lookup/mod.rs`, avoids that.

Nashtare (changes requested):
Rebased onto `f3ddc2aa` and force-pushed. Main had split `zerocheck.rs` and added lookups, typed transcripts, `security.rs` and `Result`-returning `prove`/`setup`. The second commit rewrote most of the first, so I squashed them and ported once. All 14 threads from the previous round are addressed.

Two commits on top:
- `7c7764ec`: the lookup-aware `InteractionMultilinearFolder::eval_air` never injected the pins.
  - The prover switches folders per interpolation node, and the verifier closes a lookup AIR through that folder.
  - So an AIR with both lookups and boundary cells failed honest proofs, and its committed cells were never pinned. A prover folding the true trace without pins was accepted.
  - `security.rs` also ignored the pins: it overstated `constraint-batching` and rejected pin-only AIRs.
  - Both fixed and tested. The attack test fails without the fix.
- `a6af7781`: the pin-only round trip was too small for AVX2's packed opening. CI is green.

The blocking point is the design comment on `boundary.rs`; the rest are smaller.

multi-stark/src/boundary.rs
tcoratger: You are right, and I have taken the suggestion in full: blanking is gone in 07770f99.

I went through the argument rather than taking the empirical result, and it holds exactly as you state it. The verifier's reconstruction defines `t := c' + Σ public·δ`, a bijection on multilinear polynomials for fixed public values. The pin then forces `t[cell] = public`, hence `c'[cell] = 0`. So the accepted set of traces is identical to the one `∃ t : AIR(t) ∧ t[cell] = public ∧ commitment opens to t` accepts, and the two commitments differ by a shift both parties can compute before any challenge.

**No, blanking buys nothing I can name.** The zeroing trick pays when something other than a constraint already forces the committed cell to zero — a restricted-support commitment, a column defined as a sum over a sub-domain. Here the pin is that force, and the pin is the boundary constraint the trick was meant to replace. Public values need no hiding, and the commitment cannot be reused across public inputs because the rest of the trace depends on them.

I could not write a test that fails without blanking, which is the answer to your last question.

Dropped, per your list: `take_cells`, `restore_cells`, `reconstruct`, `last_row_successor_weight`, `Table::poly_mut`, and `multi-stark/src/boundary.rs` itself. The prover no longer clones any table, and the verifier no longer builds `Cow` copies or Lagrange corrections. `prover.rs` goes from +74 to +15 lines against main; `verifier.rs` from +47 to +22.

Kept, also per your list: the declaration API, `validate`, the injected pins, the degree and security accounting, the backend guards and the debug check.

Two things follow that are worth stating, since they change what the PR claims:

- The main commitment is the trace again, so anything linking a commitment to a trace keeps working.
- Completeness no longer depends on `build_witness` / `committed_table` leaving cells in place. That invariant is not merely unasserted now, it is absent.

The description is rewritten around the declaration rather than the commitment trick, and it credits this thread for the removal.

multi-stark/src/prover.rs
tcoratger: Gone with the blanking — no table is cloned any more. The prover keeps the committed views exactly as it did before this PR:

```rust
let tables = (0..num_instances)
    .map(|table_index| config.committed_table(&prover_data, table_index))
    .collect::<Vec<_>>();
```

That is the whole of what `prover.rs` does about boundary IO now, besides validating the declaration before the transcript is touched.

multi-stark/src/verifier.rs
tcoratger: Correct — the title now carries it:

```
feat(air,multi-stark)!: public inputs bound by trace position
```

Same reasoning as `Transcript` in #2102 and `Security` in #2100: `VerificationError` is not `#[non_exhaustive]`, so a downstream exhaustive `match` stops compiling.

Nothing else in this PR is breaking. `BaseAir::public_boundary_io` is defaulted, and `BoundaryIoError` moved to `p3-air` but is still re-exported at `p3_multi_stark::BoundaryIoError`, so that path keeps resolving.

air/src/air.rs
tcoratger: Right, and the doc now says so:

```rust
/// A wrapper or enum AIR must forward this method along with [`Self::width`].
/// Forgetting to leaves every wrapped cell unbound, and nothing reports it,
/// because an empty list is a valid declaration.
```

And an end-to-end test pins it. `MixedFibAir` is an enum over `FibAir` (boundary constraints, no listed cell) and `FibIoAir` (no boundary constraint, three listed cells), both instances in one batch, one commitment, one zerocheck, one opening:

- `prove_verify_mixed_binding_batch_roundtrips` — the honest batch verifies.
- `verify_rejects_wrong_claimed_output_in_a_mixed_binding_batch` — shifting only the listed instance's output claim is rejected with `FinalSumMismatch`.

I checked the second one bites: replacing the wrapper's `public_boundary_io` with `&[]` — exactly the omission you describe — leaves the round trip passing and makes the rejection test fail.

Also fixed here: the support table's claim about univariate STARKs is now true beyond the entry points, see the `uni-stark/src/verifier.rs` thread.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Three of the four are added; the fourth stopped being a gap.

**Mixed batch** — `MixedFibAir`, an enum over `FibAir` and `FibIoAir`, both in one batch. Covers the `Some`/`None` mix, and the forwarding you flagged on `air.rs`. The rejection test fails if the wrapper drops `public_boundary_io`, which I verified by mutation.

**Preprocessed and periodic plus cells** — added to the fixtures you pointed at, each with a round trip and a wrong-claim rejection:

- `whir_preprocessed.rs`: `prove_verify_preprocessed_with_boundary_io_roundtrips`, `verify_rejects_a_wrong_public_value_alongside_a_preprocessed_trace`. `PreprocessedAir` is now parameterized by its cells.
- `whir_periodic.rs`: `prove_verify_periodic_with_boundary_io_roundtrips`, `verify_rejects_a_wrong_public_value_beside_periodic_columns`.

**Binary PCS** — what made height-2 interesting was the successor-view correction, and that is gone with the blanking. A pin is now an ordinary AIR constraint gated by `is_first_row` / `is_last_row`, which every boundary-constraint AIR in the tree already exercises at every height.

So rather than stand up a binary-PCS harness for a property the PCS no longer participates in, I took the adjacency itself at the zerocheck, where it costs nothing: `boundary_io_pins_both_ends_of_a_two_row_trace` lists both ends of one column on a two-row trace — the shortest a zerocheck accepts, where the first and last rows are neighbours — and checks that swapping the two claims is rejected. A single merged end would pass that.

If you would rather have the GF(2^128) probe in-tree anyway, say so and I will land it; I did not want to add a PCS harness on a hunch.

**Blanking** — moot, per the other thread.

multi-stark/src/security.rs
tcoratger: Fixed — the report validates the declaration, next to the successor-column check and mapping the same way:

```rust
// A malformed declaration is rejected by prove and by verify.
// Reporting a security level for a statement neither accepts would mislead.
if boundary::validate(air.public_boundary_io(), air.width(), air.num_public_values()).is_err() {
    return Err(invalid("public boundary declaration is malformed"));
}
```

`InvalidShape` carries a `&'static str`, so the specific error is not threaded through; the string names the class, as the neighbouring checks do.

`security_rejects_an_invalid_boundary_io_declaration` covers it, with the out-of-range AIR the verifier test already uses.

`prove_with_security` therefore fails the assessment instead of passing it and then panicking at `prove`'s validation.

multi-stark/src/keys.rs
tcoratger: I took your second option: the setup check is dropped.

Your reason is the decisive one — the keys carry no AIR, so a check at setup guarantees nothing about the AIR either party later hands in. Keeping it would have meant validating three times for one real guarantee, and the panic inside a `Result`-returning function was the symptom of that.

So each party validates the AIR it is actually given, once:

- `prove` panics, listed under `# Panics`, alongside its other shape asserts.
- `verify` returns `VerificationError::BoundaryIo { instance, error }`.
- `security_report` returns `InvalidShape`, per the `security.rs` thread.

The three docs you listed are gone with `multi-stark/src/boundary.rs`; the replacement in `p3_air::boundary` says what actually happens:

```rust
/// A declaration is fixed by the AIR, never by the witness.
/// Each party checks the AIR it is handed, since nothing else ties the two together.
```

The prover's panic now names the instance too, matching the verifier's:

```rust
.unwrap_or_else(|error| panic!("instance {instance} boundary IO: {error}"));
```

`setup_rejects_an_invalid_boundary_io_declaration` is gone with the check; `verify_rejects_an_invalid_boundary_io_declaration` still covers the reachable path.

air/src/check_constraints.rs
tcoratger: Fixed as suggested. `boundary_io_mismatches` returns `(index in declaration order, row)` per mismatch, and the collecting variant turns each into a `ConstraintFailure`:

```rust
all_failures.extend(
    boundary_io_mismatches(air, main, public_values)
        .into_iter()
        .map(|(index, row)| ConstraintFailure {
            row,
            constraint: total_constraints_per_row + index,
            label: Some(boundary_io_label(&cells[index])),
        }),
);
```

Indices land after the AIR's own, in declaration order — the order the folder injects the pins.

Panics are kept for out-of-range declarations only, and they now come from the shared `boundary::validate` rather than from open-coded asserts, so the two checkers agree on what a well-formed declaration is (see the `boundary.rs:81` thread).

`check_constraints`, the panicking variant, still panics on the first mismatch, before the row loop.

`test_boundary_io_mismatch_is_collected_rather_than_panicking` covers the new path: a constraint-free AIR with one wrong cell reports `total_constraints_per_row = 0` and one failure at row 3, constraint 0.

uni-stark/src/prover.rs
tcoratger: Fixed — the guard now runs first, matching batch-stark:

```rust
assert!(
    air.public_boundary_io().is_empty(),
    "uni-stark does not support boundary-IO public values; bind them with AIR constraints"
);

#[cfg(debug_assertions)]
p3_air::check_constraints(air, &trace, public_values);
```

A debug build of an AIR with cells now says the backend does not support them, rather than walking the whole trace to report that a cell it will never bind does not hold its value.

uni-stark/src/verifier.rs
tcoratger: All four now assert, so the support table in `air.rs` is true of the building blocks and not just the entry points:

- `uni_stark::verify_constraints`
- `uni_stark::quotient_values`
- batch-stark's `quotient_values`
- `VerifierData::verify_constraints_with_lookups`

An assert rather than a doc line, since a verifier assembled from these accepts any value in a declared cell, and that is not something a comment stops. The cost is one `is_empty()` on a slice the AIR already holds; `quotient_values` runs it once per call, not per row.

Nothing in-tree hits them, as you say — they are there for the caller who assembles their own verifier.

multi-stark/src/boundary.rs
tcoratger: Taken as suggested, and it let the whole module collapse.

`p3_air::boundary` now holds `BoundaryEnd`, `BoundaryPublic`, `BoundaryIoError` and the non-generic validator:

```rust
pub fn validate(
    cells: &[BoundaryPublic],
    width: usize,
    num_public_values: usize,
) -> Result<(), BoundaryIoError>
```

Call sites read `boundary::validate(air.public_boundary_io(), air.width(), air.num_public_values())` — no turbofish anywhere.

`check_public_boundary_io` is gone: the range checks are that one call, so the duplicate-cell case is now covered there too, and the comparison against the trace is all that is left locally. The two checkers can no longer drift on what a well-formed declaration is.

`multi-stark/src/boundary.rs` is deleted outright — with the blanking gone there was nothing else in it. `p3_multi_stark::BoundaryIoError` still resolves, as a re-export.

This adds `thiserror` to `p3-air`, which felt right given twelve crates in the workspace already use it, but say the word if you would rather p3-air stayed on hand-written `Display`.

multi-stark/tests/whir_fibonacci.rs
tcoratger: You are right on the diagnosis, and the test is gone.

It rejected through the transcript: public values are absorbed before the zerocheck, so changing one at verify time diverges the transcript and the opening fails first. `verify_rejects_tampered_public_values` already covers that for an AIR with no cells, and pinning the WHIR error string tied this one to PCS internals for no added coverage.

`verify_rejects_wrong_claimed_output_boundary_io` is the one that exercises boundary IO, and its comment now says why: both sides share the wrong claim, so the transcript stays in step and only the pin can reject.

multi-stark/benches/zerocheck.rs
tcoratger: Dropped, and the file is byte-for-byte back to main.

Your reading was right and the blanking removal makes it plainer: the two arms assert the same count of degree-two constraints in the same family, so there was never a difference for the benchmark to find. Anything worth measuring lives in `prove()`, and with the clone gone there is nothing there either.

`FibIoAir` is down from three definitions to two, which is the floor — the remaining pair are in different crates (unit tests in `zerocheck/mod.rs`, integration tests in `tests/whir_fibonacci.rs`) and cannot share code.

Inside `zerocheck/mod.rs` I took your `PermutationSumAir { cells }` shape: `FibTransitionAir` and `FibIoAir` are now one `FibRecurrenceAir { cells }`, so the two arms of the pin test are provably the same AIR with a different declaration rather than two copies that have to be kept in step. `PreprocessedAir` in the preprocessed fixture is parameterized the same way.

air/src/air.rs
```diff
@@ -189,6 +193,40 @@ pub trait BaseAir<F>: Sync {
...
+    ///
+    /// A wrapper or enum AIR must forward this method along with [`Self::width`].
```
Nashtare, line 218: **LOW: the forwarding rule covers wrappers, but not an AIR that embeds another.**

A parent AIR that runs a component through `SubAirBuilder` (`uni-stark/src/sub_builder.rs`) calls the component's `eval` on a column sub-range. Nothing reads the component's `public_boundary_io`, so its cells go unbound, and nothing reports it. That is the same failure as a wrapper that forgets to forward.

Forwarding the list verbatim is wrong there too. `SubAirBuilder` narrows `main()` to `column_range` but passes the parent's `public_values()` through. So the parent has to offset each `column` by `column_range.start`, while `public_value` already indexes the parent's list.

One more sentence would cover it, e.g. "An AIR that embeds another must list the embedded cells itself, in its own column numbering."

multi-stark/src/security.rs
```diff
@@ -209,18 +210,34 @@ where
...
         if !builder.exclusive_interactions().is_empty() {
             return Err(invalid("exclusive lookups are unsupported"));
         }
-        let constraints = builder
+        let own_constraints = builder
             .base_constraints()
             .len()
             .checked_add(builder.extension_constraints().len())
             .ok_or_else(|| invalid("constraint count overflow"))?;
+        // The folder batches one public boundary pin per listed cell with the AIR's own constraints.
```
Nashtare: **LOW: no symbolic pass sees the pins, so every consumer has to add them back by hand.**

The folders inject the pins in `eval_air`, after `Air::eval`. `SymbolicAirBuilder` and `InteractionSymbolicBuilder` only run `Air::eval`. This PR therefore compensates in two places:
- here, for the count;
- in `get_air_degrees` (`zerocheck/mod.rs:219`), for the degree.

Any later code that counts or scores constraints from a symbolic pass has to repeat this. So does any tool that exports an AIR's constraints, for example a serialized verifying key that rebuilds the AIR from `get_symbolic_constraints`. Otherwise it silently gets a statement without the pins.

Either would keep this in one place:
- a small helper next to `eval_boundary_io` returning the pin count and degree;
- a line on `BaseAir::public_boundary_io` saying symbolic builders never see the pins.

air/src/check_constraints.rs
```diff
@@ -619,6 +693,19 @@ where
...
 
+    // A listed cell has no AIR constraint, so it takes an index after the AIR's own.
+    // That is where the proving backend injects its pin.
+    let cells = air.public_boundary_io();
+    all_failures.extend(
+        boundary_io_mismatches(air, main, public_values)
+            .into_iter()
+            .map(|(index, row)| ConstraintFailure {
+                row,
+                constraint: total_constraints_per_row + index,
```
Nashtare, line 718: **LOW: cell failures bypass `max_failures`, and the report's constraint count leaves them out.**

- The cap is checked only at the top of each row, and these failures are appended after the loop. With `max_failures = Some(0)`, or once the cap is reached, the report still gains up to `cells.len()` failures. The declaration bounds that, so memory is fine, but the cap isn't honored.
- `total_constraints_per_row` is documented as the number of constraints evaluated per row. Every failure added here has `constraint >= total_constraints_per_row`, so a consumer that reads that field as the index range will misplace them.
- The failures are no longer in row order: row-0 cells come after the last row's own failures. Nothing promises an order, but the doc could say so.

Truncating this `extend` to the remaining cap, and noting in the field doc that pins sit past the count, would cover the first two.

The offset itself is untested. Changing `total_constraints_per_row + index` to `index` still passes every `p3-air` and `p3-multi-stark` test, because the only collecting test uses an AIR with no constraints of its own. Giving that AIR one constraint would pin the offset.

multi-stark/src/zerocheck/mod.rs
```diff
@@ -2694,6 +2721,271 @@ mod tests {
...
+        let traces = [&trace];
+
+        let honest = [F::from_u64(5), F::from_u64(9)];
+        let public_values = [&honest[..]];
+        let (proof, _) = prove_traces(&zerocheck, &traces, &public_values, &mut fresh_challenger());
+        zerocheck
+            .verify::<F, EF, _>(&proof, &[1], &public_values, &mut fresh_challenger())
+            .expect("both ends of a two-row trace must verify");
+
+        // Mutation: swap the two claims, which a single merged end would not notice.
```
Nashtare: **NIT: the swap detects inverted ends, not merged ones.**

If both pins shared one selector, the honest half above would already fail, since `5 − 9 ≠ 0` on whichever row both read. The swapped claims pass only when `First` and `Last` are exchanged: then the first pin reads 9 against 9 and the second reads 5 against 5. That is what this half detects, so "which inverted ends would accept" describes it better.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1122,3 +1170,639 @@ fn generate_whir_fixture() -> Result<(), Box<dyn std::error::Error>> {
...
+    //
+    //     lookup : still a permutation  -> accepted on its own
+    //     pin    : (n - 1) - n != 0     -> the zerocheck rejects
+    let n = 64;
+    let mut pis = permutation_io_public_values(n);
+    pis[0] += F::ONE;
+
+    let err = prove_verify_permutation_io(n, &pis).unwrap_err();
+    assert!(
+        matches!(err, VerificationError::Zerocheck(_)),
```
Nashtare: **NIT: this still accepts any zerocheck error.**

The other boundary-IO rejections match `FinalSumMismatch`, and so does this one. I tightened it locally to `VerificationError::Zerocheck(ZerocheckError::FinalSumMismatch)` and it passes. Matching it keeps an unrelated zerocheck failure from standing in for the pin.

air/src/boundary.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// A cell pairs a trace position with one of the AIR's public values:
+///
+/// ```text
+///     (column, end) holds public_values[public_value]
+/// ```
+///
+/// This is a declaration, not a constraint.
+/// Binding the cell to the value is the proving backend's job.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
Nashtare: **NIT: `BoundaryEnd` and `BoundaryPublic` don't derive `Serialize`/`Deserialize`.**

`p3-air` already depends on serde with `derive`. A serialized verifying key that carries an AIR's shape (widths, public-value count, successor columns) would also have to carry this declaration, and without the derives it needs a mirror type. `Hash` would help a key digest the same way.

multi-stark/src/security.rs
```diff
@@ -242,7 +259,7 @@ where
                 "constraint degree hint understates the symbolic degree",
             ));
         }
-        if constraints > 0 && symbolic_degree == 0 {
+        if own_constraints > 0 && symbolic_degree == 0 {
```
Nashtare: **LOW: an AIR with only constant constraints plus a cell passes `prove`/`verify` but fails every security-checked entry point.**

With a listed cell, `get_air_degrees` lifts a constant own family to degree 2 (`zerocheck/mod.rs:222`). Its "constant family" assert then passes, but this line still rejects the same AIR. Checked locally with `builder.assert_zero(AB::Expr::ZERO)` plus one listed cell:
- `prove` and `verify`: the honest proof verifies; a wrong claim, or `assert_zero(ONE)`, is rejected with `FinalSumMismatch`.
- `security_report`, `prove_with_security`, `verify_with_security`: `InvalidShape("constant constraint families are unsupported")`.

It is sound either way. The pin sets the round degree, and the closing check evaluates the constant exactly. Before this PR, the unhinted AIR panicked in `get_air_degrees`, so both paths refused it.

Pick one rule for both places. Either reject a constant own family in `get_air_degrees` whatever the pins, or relax this check once cells lift the degree. The `# Panics` line on `get_air_degrees` ("Panics if a declared family is constant") should follow whichever you pick.

multi-stark/src/folder.rs
```diff
@@ -255,6 +255,7 @@ where
     ///
     /// This is the terminal step of the builder: it consumes the folder.
     /// Attach preprocessed and periodic columns before calling, if the AIR reads them.
+    /// Cells the AIR lists as public inputs add one pin each, after its own constraints.
```
Nashtare, line 261: **NIT: `with_alpha_powers` still sizes its powers by the AIR's own constraints.**

Its doc says `n` is "the number of constraints the AIR asserts", and `into_accumulator` says the same. Pins now take slots in `alpha_powers` too. A caller who sizes the powers from the symbolic constraint count and then calls `eval_air` on an AIR with cells indexes past the end at `powers[self.constraint_index]`. Only a test uses it in-tree, but it is public. One word on each doc ("constraints and pins") covers it.

Nashtare (approved):
Re-reviewed at `07770f99`. All 13 threads from the last round are addressed, and each reply matches the code. Approving.

What I checked:
- **Pins on every path.** Every multi-stark AIR evaluation goes through the two `eval_air`s: the prover kernel in `rounds.rs` and the verifier's closing check in `zerocheck`. Both inject the pins. Deleting the call from the ordinary folder fails 8 tests. Deleting it from the lookup folder fails 3.
- **Rejection end to end.** A prover folds the same transitions with no cells, over an honest trace, against a shifted output claim.
  - The verifier using the AIR with cells rejects it with `FinalSumMismatch`. The same verifier with the no-cell AIR accepts it.
  - The same holds for a degree-3 lookup AIR whose ordinary family is only pins, so that family is dropped at node 3. No in-tree test has that shape. It round-trips honestly and rejects wrong first-row and last-row claims.
- **Degree and security accounting.** Scoring the pin at degree 1 breaks 5 honest proofs. A degree-1 AIR with hint `Some(1)` and one cell is scored at degree 2 by `get_air_degrees` and by the security report. Removing the pin count or the validation from `security.rs` is caught by a test.
- **Other backends.** Every public entry point and building block in uni-stark and batch-stark rejects an AIR that lists cells. No in-tree generic wrapper forwards `BaseAir`.
- CI is green on every leg. Locally, 570/570 tests pass across `p3-air`, `p3-multi-stark`, `p3-uni-stark` and `p3-batch-stark`.

The inline comments are LOW or NIT, and none of them blocks. On the binary-PCS probe: agreed. With the successor correction gone, the two-row zerocheck test is enough.

air/src/air.rs
tcoratger: Right, and the doc now covers it:

```rust
/// An AIR that embeds another lists the embedded cells itself, in its own numbering.
/// Forwarding the embedded list verbatim is wrong: a sub-builder narrows the main
/// trace to a column range but passes the parent's public values straight through,
/// so each `column` shifts by the range start while `public_value` does not.
```

The asymmetry is the part worth spelling out, since "forward it" is the obvious instinct after the wrapper sentence right above, and it is the wrong instinct here.

I also added a third line while I was in this doc, for the point raised on `security.rs`: symbolic builders run `Air::eval` alone and never see the injected pins, so anything counting or scoring constraints from a symbolic pass has to add them back.

multi-stark/src/security.rs
tcoratger: Took the first option, and the doc line as well — the helper keeps the two in-tree consumers honest, the doc line is what reaches an out-of-tree one.

`multi-stark/src/folder.rs`, directly below `eval_boundary_io`:

```rust
/// What the injected public boundary pins add to an AIR's constraint family.
///
/// A symbolic pass runs [`Air::eval`] alone, so it never sees the pins.
/// Anything that counts or scores constraints from such a pass adds them back through here.
pub(crate) struct BoundaryIoPins {
    /// One pin per listed cell.
    pub(crate) count: usize,
    /// Degree the pins score, or zero when the AIR lists no cell.
    pub(crate) degree: usize,
}

pub(crate) const fn boundary_io_pins(cells: &[BoundaryPublic]) -> BoundaryIoPins
```

It takes the declaration rather than the AIR, like `boundary::validate` after the last round, so no call site needs a turbofish.

Both consumers now go through it: `security.rs` uses `pins.count`, and `get_air_degrees` uses `pins.degree` in place of its own `BOUNDARY_IO_PIN_DEGREE`, which is gone from `zerocheck/mod.rs`. That also settled the disagreement you found one thread down — see the `security.rs:262` reply.

`BaseAir::public_boundary_io` now says the same thing for anyone outside this crate.

air/src/check_constraints.rs
tcoratger: All three, and the untested offset.

**The cap.** Cell failures are now truncated to what is left of it:

```rust
let remaining = max_failures.map_or(usize::MAX, |cap| cap.saturating_sub(all_failures.len()));
all_failures.extend(
    boundary_io_mismatches(air, main, public_values)
        .into_iter()
        .take(remaining)
        ...
```

`test_boundary_io_mismatch_counts_against_the_failure_cap` pins `Some(0)` giving an empty report.

**The count.** `total_constraints_per_row` now says a listed cell is not one of them:

```rust
/// A listed public boundary cell is not one of these.
/// Its failure carries an index at or past this count, matching where the backend
/// injects the pin.
```

**The order.** Said on `failures`: row-ordered for the AIR's own constraints, cells appended last and therefore out of row order with respect to them.

**The offset.** You were right that nothing tested it — the collecting test's AIR asserted nothing, so `total_constraints_per_row` was zero and the offset was invisible. `BoundaryIoAir` now asserts one transition constraint (`next == local + 1`, which every fixture in that module satisfies), so the report reads `total_constraints_per_row == 1` and the cell failure sits at constraint 1. I checked your mutation: replacing `total_constraints_per_row + index` with `index` now fails with `left: 0, right: 1`.

multi-stark/src/zerocheck/mod.rs
tcoratger: Correct — a merged end fails the honest half, not this one. Reworded:

```rust
// Mutation: swap the two claims, which only inverted ends would accept.
//
// Two pins sharing one selector would already fail the honest half above,
// since 5 - 9 is nonzero on whichever row both would read.
```

The swapped claims pass exactly when `First` and `Last` are exchanged, as you say: the first pin then reads 9 against 9 and the second reads 5 against 5.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Tightened to match its siblings:

```rust
assert!(
    matches!(
        err,
        VerificationError::Zerocheck(ZerocheckError::FinalSumMismatch)
    ),
    "expected a zerocheck final-sum mismatch, got {err:?}"
);
```

Passes, as you found. The loose form would have let a lookup-side zerocheck failure stand in for the pin, which is the one thing this test exists to distinguish.

air/src/boundary.rs
tcoratger: Added — `Serialize`, `Deserialize` and `Hash` on both:

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub enum BoundaryEnd { .. }

#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub struct BoundaryPublic { .. }
```

`p3-air` already pulled serde with `derive`, so no manifest change.

Both are plain `usize`/unit-variant data, so the derives are exact — a round trip cannot lose a cell, and a key digest over the declaration is well defined.

`BoundaryIoError` deliberately keeps its current derives: it is a rejection reason, never part of a key.

multi-stark/src/security.rs
tcoratger: Good catch, and the sharpest of this round: an AIR that proves and verifies was failing every security-checked entry point.

I picked **relax this check**, the second of your two options, because the pin genuinely supplies what the rejection exists to demand. The reason a constant own family is unsupported is that it carries no round polynomial of its own; the round polynomials are sized against the family's degree, and a degree-zero family gives the sumcheck nothing to send. A listed cell injects a degree-two pin into that same family, so the rounds are sized for two, and the closing check evaluates the constant exactly — which is your own reading of it.

Rejecting instead would have meant making `get_air_degrees` refuse a statement it can score correctly, purely to match a check that no longer applies.

Both places now read the degree the same way, through the new `boundary_io_pins`:

```rust
// A constant family has no round polynomial of its own.
// A listed cell lifts it to the pin's degree, which is what `get_air_degrees` scores,
// so both entry points accept and reject the same statements.
if own_constraints > 0 && symbolic_degree.max(pins.degree) == 0 {
    return Err(invalid("constant constraint families are unsupported"));
}
```

`get_air_degrees` computes `own_constraint_degree.max(pins.degree)` and its `# Panics` line follows: *"Panics if a declared family is constant and no listed cell lifts it, since a constant has no round polynomial of its own."*

An AIR with a constant family and no cell is still rejected by both, unchanged.

Two tests pin it, built on your repro (`assert_zero(AB::Expr::ZERO)` plus one listed cell):

- `security_checked_roundtrip_for_a_constant_family_lifted_by_a_pin` — `security_report` produces a level, and `prove_with_security` / `verify_with_security` round-trip.
- `verify_rejects_a_wrong_claim_against_a_constant_family_lifted_by_a_pin` — a wrong last-row claim is still `FinalSumMismatch`, so relaxing the shape check did not relax the binding.

I confirmed the first one fails against the old condition, with exactly the `InvalidShape("constant constraint families are unsupported")` you saw.

multi-stark/src/folder.rs
tcoratger: Fixed on both docs, and on `eval_air`'s while I was there:

```rust
/// `alpha_powers[i]` must be `alpha^(n - 1 - i)`, where `n` is the number of constraints
/// asserted at this node. ...
///
/// `n` counts the AIR's own constraints plus one pin per listed public boundary cell.
/// A count taken from a symbolic pass misses the pins and indexes past the end.
```

The second line is the failure you describe, named so a caller sizing from `get_symbolic_constraints` sees it before the panic does. `into_accumulator` and `eval_air` now say `n` the same way.

The debug assert in `into_accumulator` already catches a short slice, but only after the indexing, so the doc is what actually helps here.

PR #1949, opened 2026-07-14
feat(whir): projective (monomial-basis) WHIR end to end

jdodinh:
Follow-up to #1900, now merged, so all 11 commits here are this PR's own. Closes the experiment of #1890.

**TL;DR.** Runs the full WHIR protocol in the projective (monomial) basis of eprint 2026/762 behind an opt-in flag, and measures it. The pre-registered close condition fires: end-to-end it is parity at best, so this is a documented negative result. The per-round win is real (subtraction-free binding, **1.13x** standalone) but acts on ~1% of a hash-dominated prover. The technique wants a sum-check-dominated prover (the paper's Jolt/Spartan setting), not a PCS on 31-bit fields.

## What

`WhirConfig::with_basis(Basis::Projective)` switches how the tables are interpreted. The committed bytes never change, and the default path is behaviourally unchanged. Because WHIR's sum-check challenge is also its codeword-folding challenge, four things switch together and only together:

- **Binding**: `a0 + (a1 - a0) * r` becomes the subtraction-free `a0 + a1 * r`.
- **Round message**: `[h(0), h(inf)]` becomes `[s(1), s(inf)]`, deriving `s(0) := C - s(inf)` (paper Fig. 3).
- **Folds of committed data**: monomial (Horner) evaluation.
- **Weight closed forms**: the same tensors read as coefficients, e.g. eq becomes `prod((1-z_i) + z_i*r_i)`.

#1900 landed the message half (`Basis`, `sumcheck_coefficients`, `reduce_claim`, the basis parameter on the verifier). This PR adds everything that needs a caller to be coherent: `Basis::fix_var`, the basis tag on `ProductPolynomial`, the projective weight closed forms and initial-round path, monomial evaluation in `p3-multilinear-util`, and the WHIR wiring including the domain-separator binding.

The out-of-domain anchor stays the MLE of the table, identical in both bases, so the constraint machinery needs no second tensor family. The basis is bound into the domain separator, making the two bases transcript-incompatible (tested); this also means evaluation-basis transcripts differ from main's.

## Numbers

M3 Pro, single-threaded, `-Ctarget-cpu=native`, KoalaBear + Poseidon2, prefix layout. The primary configuration is grind-free: grinding work is fixed per transcript and the two bases produce different transcripts, so a pow-enabled comparison carries per-case luck bias of tens of percent.

whir_pcs prove, speedup = eval / projective (> 1 means projective wins):

| Size | Evaluation | Projective | Speedup |
|---|---|---|---|
| k14 | 1.804 ms | 2.183 ms | 0.83x |
| k18 | 40.68 ms | 46.64 ms | 0.87x |
| k20 | 193.4 ms | 210.9 ms | 0.92x |
| k22 | 768.4 ms | 854.6 ms | 0.90x |

- **Round-coefficient kernel**: 0.98-0.99x, parity, as the cost accounting predicts. The complete message swaps a subtraction pass for an addition pass rather than removing one. (This corrects the ~11% figure from #1900's early review, which was measured against an incomplete `[s(0), s(inf)]` message.)
- **Binding kernel**: 1.10-1.14x across k16-k22 and both fields. The real per-round win.
- **Initial rounds**: the projective path runs them generically, ~10.8% of the grind-free k22 prove (92.2 ms vs 3.7 ms for the eval side's SVO fast path).

**The close condition fires** (pre-registered: under ~1% end-to-end means document the negative result rather than merge). The 9-21% deficit is almost entirely the initial-round asymmetry; crediting back the measured spans leaves 1.003x. Even a perfect projective SVO port reaches parity, not a win, because of where the time actually goes at k=22:

| Phase | Share | Basis-affected? |
|---|---|---|
| Merkle hashing | ~56% | no |
| PoW grinding | ~32% | no |
| DFT | ~9% | no |
| Sum-check arithmetic | ~1.2% | yes |

## Reproduce

Benchmarks taken on Rust 1.93, `parallel` feature off. Bench IDs come in eval/projective pairs; speedups are ratios of paired medians.

```bash
# Deciding benchmark. Needs this branch.
RUSTFLAGS="-Ctarget-cpu=native" cargo bench -p p3-whir --bench whir_pcs -- "whir_pcs/basis"

# Both kernels. These are on main (#1900).
RUSTFLAGS="-Ctarget-cpu=native" cargo bench -p p3-sumcheck --bench sumcheck -- \
  "round_coefficients/ext_ext_packed_prefix" --warm-up-time 5 --measurement-time 10
RUSTFLAGS="-Ctarget-cpu=native" cargo bench -p p3-sumcheck --bench sumcheck -- \
  "fix_var" --warm-up-time 5 --measurement-time 10

# Per-phase profile (drop --projective for the evaluation run).
RUSTFLAGS="-Ctarget-cpu=native" cargo run --release -p p3-whir --example whir -- \
  --num-variables 22 -p 20 --order prefix --projective
```

## Caveats

- **Initial-round asymmetry.** The eval side's SVO fast path is evaluation-specific; the projective side folds generically. A projective SVO variant is derivable (`s(0)=h(0)`, `s(inf)=h(1)`, `s(1)=2h(0)+2h(1)-h(inf)`) but needs the accumulators' X=1 slot materialised. Left as a follow-up, since even a perfect port only reaches parity.
- **Single-threaded, 31-bit fields**, matching the paper's methodology. The paper's larger wins are on 128-bit fields where subtractions cost proportionally more.
- **Scope**: Sections 3, 4.1, 4.3. The 4.2 weight simplifications are deliberately not taken, forgoing a ~1.06x table-build effect to keep the existing statement machinery unchanged.
- One bug en route: `ProductPolynomial::transition()` dropped the basis at the packed-to-scalar seam, silently reverting later rounds to evaluation arithmetic. The dot-product invariant cannot catch that class, since post-reset rounds are internally coherent; the regression test compares the fully bound table against a monomial evaluation of the original, which can.

## Test plan

- [x] Projective WHIR round trips on two multi-round schedules (incl. the `[8, 7]` partial final fold and a two-table spec), in debug with the internal dot-product assertions live.
- [x] Cross-basis rejection via the domain separator.
- [x] Standalone projective sum-check round trip, per-position tampering rejection, and kernel/binding proptests against independent references.
- [x] Evaluation-basis suites untouched. `clippy --all-targets -- -D warnings`, fmt, and docs clean on 1.98; `cargo test --workspace` green (112 suites).

### DIFF
--- a/multilinear-util/src/point.rs
+++ b/multilinear-util/src/point.rs
@@ -123,6 +123,29 @@ where
             .product()
     }
 
+    /// Evaluates the equality-weight tensor as *monomial coefficients* at `q`.
+    ///
+    /// The projective counterpart of [`Self::eval_eq`]: the same per-variable
+    /// pairs `(1 - p_i, p_i)` are read as coefficient pairs, so the value at
+    /// `q` is the monomial evaluation
+    ///
+    /// ```text
+    /// prod_i ((1 - p_i) + p_i * q_i)
+    /// ```
+    pub fn eval_eq_projective<EF: ExtensionField<F>>(p: &[F], q: &[EF]) -> EF {
+        assert_eq!(
+            p.len(),
+            q.len(),
+            "Points must have the same number of variables"
+        );
+
+        // Per-variable factor: (1 - p) + p * q = p * (q - 1) + 1.
+        p.iter()
+            .zip(q)
+            .map(|(&l, &r)| (r - EF::ONE) * l + EF::ONE)
+            .product()
+    }
+
     /// Evaluates the selection polynomial of a point against a univariate domain element.
     ///
     /// The univariate element is expanded into the square-power vector
@@ -150,6 +173,33 @@ where
                 term
             })
             // Multiply all per-coordinate factors into the selection value.
+            .product()
+    }
+
+    /// Evaluates the selection-weight tensor as *monomial coefficients* at `point`.
+    ///
+    /// The projective counterpart of [`Self::eval_select`]: the selection
+    /// weight table is the tensor with per-variable pairs `(1, var^(2^i))` in
+    /// both bases, but read as coefficient pairs its value at `point` is the
+    /// monomial evaluation
+    ///
+    /// ```text
+    /// prod_i (1 + var^(2^i) * point_i)
+    /// ```
+    #[must_use]
+    #[inline]
+    pub fn eval_select_projective<EF: ExtensionField<F>>(mut var: F, point: &[EF]) -> EF {
+        // Same descending-power walk as `eval_select`; only the per-variable
+        // factor changes from interpolation to the monomial rule.
+        point
+            .iter()
+            .rev()
+            .map(|&r| {
+                // Per-coordinate factor: 1 + var * r.
+                let term = r * var + EF::ONE;
+                var = var.square();
+                term
+            })
             .product()
     }
 
--- a/multilinear-util/src/poly/mod.rs
+++ b/multilinear-util/src/poly/mod.rs
@@ -282,6 +282,41 @@ impl<Packed, S> Poly<Packed, S>
 where
     S: Borrow<[Packed]>,
 {
+    /// Evaluates the packed table as *monomial coefficients* at `point`.
+    ///
+    /// Packed counterpart of [`Poly::eval_monomial`]: the stored (prefix)
+    /// variables fold in packed form, then the SIMD lanes (the last
+    /// `log2(W)` variables) finish in scalar form.
+    ///
+    /// Kept for representation parity with [`Poly::eval_packed`]; the
+    /// production protocol reads final openings in scalar form, so this is
+    /// exercised by the packed equivalence tests rather than the prover.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the point arity does not match the logical variable count.
+    pub fn eval_monomial_packed<F, EF>(&self, point: &Point<EF>) -> EF
+    where
+        F: Field,
+        EF: ExtensionField<F, ExtensionPacking = Packed>,
+        Packed: PackedFieldExtension<F, EF> + Copy,
+    {
+        let lane_vars = log2_strict_usize(F::Packing::WIDTH);
+        assert_eq!(
+            self.num_variables() + lane_vars,
+            point.num_variables(),
+            "point arity must match the logical variable count"
+        );
+        let (stored, lanes) = point.as_slice().split_at(self.num_variables());
+
+        // Fold the stored prefix variables down to one packed element.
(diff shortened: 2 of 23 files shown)


Nashtare:
@jdodinh as it's stacked on #1900 you may want to target this as base branch for ease of review (GH supports stacked PRs now FYI). Also branches needs conflicts fixing

jdodinh:
@Nashtare 

Conflicts fixed, both branches rebased onto the current main, CI green.

On stacking: the new feature doesn't support cross-fork stacks ([docs](https://docs.github.com/en/pull-requests/reference/stacked-pull-requests)), and my branches live on my fork (read access only here). For review, this comparison shows just this PR's 11 commits: https://github.com/jdodinh/Plonky3/compare/projective-sumcheck...projective-whir-integration. If you'd rather have a real base branch, push `projective-sumcheck` to this repo and I'll retarget.

Nashtare:
Ah yes my bad hadn't realized it was on a separate fork, all good then!

jdodinh:
@tcoratger @Nashtare ready for a look whenever suits, no rush.

Short version: the experiment came out negative. Projective WHIR is parity at best end-to-end, 0.83-0.92x raw, and 1.003x once you credit back an initial-round asymmetry that is an implementation gap rather than the technique. The per-round win is real (1.13x on the binding kernel) but lands on ~1% of a hash-dominated prover, so it cannot surface. Full accounting in the description.

I set a close condition when I scoped this (under ~1% end-to-end means document the negative result rather than merge), and it fired. So rather than assume, what would you both like to do with it?

1. Merge as is. Opt-in, defaults unchanged, and it leaves the repo a reference implementation plus the measurement.
2. Close it, with the writeup as the artifact.
3. Merge a subset. The monomial evaluation helpers in `p3-multilinear-util` stand on their own.

Any of the three is fine by me. I would rather you pick before spending review time on the code, since option 2 makes a detailed pass a wasted effort. CI is green.

Nashtare:
@jdodinh thanks for the detailed write-up. 

As is, I'd rather not review / merge this if it doesn't bring concrete improvements over the existing codebase. If you foresee a use in cherry-picking some of the primitives you introduced, then why not extracting them as suggested in your 3rd point, otherwise I'd suggest to just close this unfortunately.

jdodinh:
Makes sense, closing it. On the primitives, I don't currently foresee a use. Nothing in main evaluates in the monomial basis, so they'd be public API with only tests as callers, which is the concern raised on #1900.

Thanks both for the review time. Leaving the description as the writeup: the per-round win is real on 31-bit fields (1.13x on binding) but lands on ~1% of a hash-dominated prover, so it needs a sum-check-dominated one to surface. Worth knowing if anyone revisits: most of the measured end-to-end deficit is the evaluation-specific SVO fast path in the initial rounds, an implementation gap rather than the technique.

PR #1952, opened 2026-07-16
fix: make bench run and add benches to CI

Nashtare:
I noticed the `fri_vs_whir` bench was broken, and this never was caught on CI because we never run benches there.
Fixes both issues.

### DIFF
--- a/.github/workflows/bench-smoke.yml
+++ b/.github/workflows/bench-smoke.yml
@@ -1,0 +1,32 @@
+name: Bench Smoke Test
+
+on:
+  schedule:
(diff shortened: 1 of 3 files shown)


Nashtare:
I'm excluding `p3-dft` from the CI bench run as it takes ages to complete (~1300) and already extensively tested in unit-tests + wouldn't suffer from failing configuration setups like the initially flaky one.

Given the runtime being quite heavy, I made the test as a weekly CRON job instead of running on every push.

tcoratger (approved):

PR #1953, opened 2026-07-16
perf(Goldilocks): add support for AVX vectorization for Poseidon2

Nashtare:
Add support for AVX-based vectorized Poseidon2 over Goldilocks (only NEON until now).

Below are benchmarks number against `main` obtained on an AVX-512 machine (AMD EPYC 9R45, 64 vCPUs, 128 GB memory).

Screenshot 2026-07-16 at 17 35 27


tcoratger (approved):
LGTM!

I just think that target_feature = "avx512f" is not part of the CI so that tests are not compiled, we should add this to the CI.

PR #1955, opened 2026-07-21
bench(whir): make ZK overhead comparisons grind-free

latifkasuli:
## What changed

- retain the existing PoW-enabled ZK-WHIR benchmark for production-like continuity
- add a matched 128-bit, PoW-free plain-vs-ZK comparison split into commit, open, and verify phases
- use the same polynomial and prescribed opening point on both sides
- exclude witness/protocol/point cloning and commitment setup from the relevant timed regions
- add an opt-in, fixed-seed proof report covering roots, multiproofs, authenticated rows, leaf values, Merkle siblings, and the dominant ZK proof components
- port the benchmark to the typed Fiat-Shamir and fallible PCS APIs on current `main`

## Why

The existing benchmark initializes both challengers identically, but plain and ZK transcripts differ before grinding. Their 16-bit PoW searches therefore have unrelated stopping times, so a plain-vs-ZK ratio can be materially biased by nonce-search luck.

The new grind-free groups isolate protocol work while keeping the existing group available as a realistic end-to-end configuration. This is deliberately a query-heavy operating point: 128-bit Johnson soundness, start rate `2^-1`, intermediate rate `2^-6`, folding `8/6`, and no PoW. The intermediate rate leaves enough terminal slack for the occupancy-aware hiding budget introduced in #1995 at every benchmark size.

This compares the complete plain and hiding pipelines, not a pure wrapper cost: hiding WHIR omits the plain commitment-OOD and final-fold phases.

Proof reporting is opt-in so `--list`, `--test`, and filtered benchmark runs do not perform full proofs:

```console
P3_WHIR_PROOF_REPORT=1 cargo bench -p p3-whir --bench zk_overhead --features parallel -- --test
```

Current fixed-seed proof-shape output:

| vars | kind | proof bytes | roots | proofs | auth rows | base values | extension values | Merkle digests |
|---:|:---|---:|---:|---:|---:|---:|---:|---:|
| 18 | plain | 423,458 | 2 | 2 | 342 | 76,288 | 2,816 | 804 |
| 18 | ZK | 735,379 | 9 | 9 | 1,024 | 76,288 | 8,580 | 2,967 |
| 19 | plain | 434,050 | 2 | 2 | 342 | 76,288 | 2,816 | 1,117 |
| 19 | ZK | 694,917 | 9 | 9 | 978 | 76,288 | 7,085 | 3,206 |
| 20 | plain | 444,258 | 2 | 2 | 342 | 76,288 | 2,816 | 1,402 |
| 20 | ZK | 677,833 | 9 | 9 | 950 | 76,288 | 6,175 | 3,565 |

## Validation

- `cargo +nightly fmt --all --check`
- `cargo clippy -p p3-whir --bench zk_overhead --features parallel -- -D warnings`
- `cargo bench -p p3-whir --bench zk_overhead --features parallel -- --test`
- `cargo bench -p p3-whir --bench zk_overhead --features parallel -- --list`
- `P3_WHIR_PROOF_REPORT=1 cargo bench -p p3-whir --bench zk_overhead --features parallel -- --test`
- `cargo test -p p3-whir --features parallel` (196 passed, 3 ignored)

This is the first PR in a two-part ZK-WHIR overhead investigation. A follow-up draft will use this accounting to evaluate batching the base-case fresh mask commitments.

### DIFF
--- a/whir/benches/zk_overhead.rs
+++ b/whir/benches/zk_overhead.rs
@@ -10,22 +10,24 @@ use criterion::measurement::WallTime;
 use criterion::{
     BatchSize, BenchmarkGroup, BenchmarkId, Criterion, criterion_group, criterion_main,
 };
-use p3_challenger::DuplexChallenger;
-use p3_commit::MultilinearPcs;
+use p3_challenger::{CanObserve, DuplexChallenger};
+use p3_commit::{Mmcs as MmcsTrait, MultilinearPcs};
 use p3_dft::Radix2DFTSmallBatch;
 use p3_field::Field;
 use p3_field::extension::BinomialExtensionField;
 use p3_koala_bear::{KoalaBear, Poseidon2KoalaBear};
+use p3_matrix::dense::RowMajorMatrix;
 use p3_merkle_tree::MerkleTreeMmcs;
 use p3_multilinear_util::point::Point;
 use p3_multilinear_util::poly::Poly;
 use p3_sumcheck::layout::{Layout, PrefixProver, Table};
-use p3_sumcheck::{OpeningBatch, OpeningProtocol, TableShape, TableSpec};
+use p3_sumcheck::{OpeningBatch, OpeningProtocol, PrescribedPointPcs, TableShape, TableSpec};
 use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
 use p3_whir::fiat_shamir::domain_separator::DomainSeparator;
 use p3_whir::parameters::{FoldingFactor, ProtocolParameters, SecurityAssumption, WhirConfig};
+use p3_whir::pcs::proof::{PcsProof, QueryOpenings, SharedProofOpening};
 use p3_whir::pcs::prover::WhirProver;
-use p3_whir::pcs::zk::{HidingWhirPcs, ZkParameters, ZkWhirConfig};
+use p3_whir::pcs::zk::{HidingWhirPcs, ZkParameters, ZkWhirConfig, ZkWhirProof};
 use rand::SeedableRng;
 use rand::rngs::{SmallRng, StdRng};
 
@@ -44,7 +46,12 @@ type Dft = Radix2DFTSmallBatch<F>;
 
 type PlainPcs = WhirProver<EF, F, Dft, Mmcs, Challenger, PrefixProver<F, EF>>;
 type ZkPcs = HidingWhirPcs<EF, F, Dft, Mmcs, Challenger, StdRng>;
+type OcticPlainPcs = WhirProver<OcticEF, F, Dft, Mmcs, Challenger, PrefixProver<F, OcticEF>>;
 type OcticZkPcs = HidingWhirPcs<OcticEF, F, Dft, Mmcs, Challenger, StdRng>;
+type Commitment = <Mmcs as MmcsTrait<F>>::Commitment;
+type MerkleMultiProof = <Mmcs as MmcsTrait<F>>::MultiProof;
+type OcticPlainProof = PcsProof<F, OcticEF, Mmcs>;
+type OcticZkProof = ZkWhirProof<F, OcticEF, Mmcs>;
 
 // Polynomial sizes (log_2 of coefficient count) and shared knobs.
 const SIZES: [usize; 2] = [16, 18];
@@ -75,7 +82,11 @@ fn octic_no_pow_protocol_params() -> ProtocolParameters {
     }
 }
 
-fn octic_no_pow_config(num_variables: usize) -> ZkWhirConfig<OcticEF, F, Challenger> {
+fn octic_no_pow_plain_config(num_variables: usize) -> WhirConfig<OcticEF, F, Challenger> {
+    WhirConfig::new(num_variables, octic_no_pow_protocol_params()).unwrap()
+}
+
+fn octic_no_pow_zk_config(num_variables: usize) -> ZkWhirConfig<OcticEF, F, Challenger> {
     ZkWhirConfig::new(
         num_variables,
         octic_no_pow_protocol_params(),
@@ -97,6 +108,112 @@ fn mmcs() -> Mmcs {
 fn challenger() -> Challenger {
     let mut rng = SmallRng::seed_from_u64(2);
     Challenger::new(Poseidon16::new_from_rng_128(&mut rng))
+}
+
+/// Plain and hiding fixtures for the matched, grind-free comparison.
+///
+/// Both sides use the same field, security parameters, folding schedule,
+/// polynomial size, and one-point opening workload. Their adapter-specific
+/// witness/protocol types are kept separate so the benchmark exercises the
+/// public PCS interfaces exactly as downstream users do.
+struct OcticPlainFixture {
+    pcs: OcticPlainPcs,
+    witness: <OcticPlainPcs as MultilinearPcs<OcticEF, Challenger>>::Witness,
+    protocol: <OcticPlainPcs as MultilinearPcs<OcticEF, Challenger>>::OpeningProtocol,
+    points: Vec<Point<OcticEF>>,
+    domain_separator: DomainSeparator<OcticEF, F>,
+}
+
+impl OcticPlainFixture {
+    fn new(num_variables: usize) -> Self {
+        let config = octic_no_pow_plain_config(num_variables);
+        let pcs = OcticPlainPcs::new(config, Dft::default(), mmcs());
+
+        let mut rng = SmallRng::seed_from_u64(3);
+        let poly = Poly::<F>::rand(&mut rng, num_variables);
+        let table = Table::new(RowMajorMatrix::new(
+            poly.as_slice().to_vec(),
+            1 << num_variables,
+        ));
+        let witness = PrefixProver::<F, OcticEF>::new_witness(vec![table], 8);
+        let protocol = OpeningProtocol::new(vec![TableSpec::new(
+            TableShape::new(num_variables, 1),
+            vec![OpeningBatch::new(vec![0], Vec::new())],
+        )]);
+        let mut point_rng = SmallRng::seed_from_u64(5);
+        let points = vec![Point::<OcticEF>::rand(&mut point_rng, num_variables)];
+
+        let mut domain_separator = DomainSeparator::new(vec![]);
+        pcs.add_domain_separator::<8>(&mut domain_separator);
+        Self {
+            pcs,
+            witness,
+            protocol,
+            points,
+            domain_separator,
+        }
+    }
+
+    fn challenger(&self) -> Challenger {
+        let mut challenger = challenger();
+        self.domain_separator
+            .observe_domain_separator(&mut challenger);
+        challenger
+    }
+
+    fn build_proof(&self) -> (Commitment, OcticPlainProof) {
+        let mut challenger = self.challenger();
+        let (commitment, data) = self.pcs.commit(self.witness.clone(), &mut challenger);
+        let proof = self
+            .pcs
+            .open_at(data, &self.protocol, &self.points, &mut challenger);
+        (commitment, proof)
+    }
+}
+
+struct OcticZkFixture {
+    pcs: OcticZkPcs,
+    witness: Poly<F>,
+    points: Vec<Point<OcticEF>>,
+    domain_separator: DomainSeparator<OcticEF, F>,
+}
+
+impl OcticZkFixture {
+    fn new(num_variables: usize) -> Self {
+        let pcs = OcticZkPcs::new(
+            octic_no_pow_zk_config(num_variables),
+            Dft::default(),
+            mmcs(),
+            StdRng::seed_from_u64(4),
+        );
+        let mut rng = SmallRng::seed_from_u64(3);
+        let witness = Poly::<F>::rand(&mut rng, num_variables);
+        let mut point_rng = SmallRng::seed_from_u64(5);
+        let points = vec![Point::<OcticEF>::rand(&mut point_rng, num_variables)];
+
+        let mut domain_separator = DomainSeparator::new(vec![]);
+        pcs.add_domain_separator::<8>(&mut domain_separator);
+        Self {
+            pcs,
+            witness,
+            points,
+            domain_separator,
+        }
+    }
+
+    fn challenger(&self) -> Challenger {
+        let mut challenger = challenger();
+        self.domain_separator
+            .observe_domain_separator(&mut challenger);
+        challenger
+    }
+
+    fn build_proof(&self) -> (Commitment, OcticZkProof) {
+        let mut challenger = self.challenger();
+        let (commitment, data) = self.pcs.commit(self.witness.clone(), &mut challenger);
+        let proof = self.pcs.open(data, self.points.clone(), &mut challenger);
+        (commitment, proof)
+    }
 }
 
 /// Benchmarks the plain (non-hiding) prover for one size.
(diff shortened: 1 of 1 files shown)


whir/benches/zk_overhead.rs
```diff
@@ -10,22 +10,24 @@ use criterion::measurement::WallTime;
...
 use p3_koala_bear::{KoalaBear, Poseidon2KoalaBear};
+use p3_matrix::dense::RowMajorMatrix;
 use p3_merkle_tree::MerkleTreeMmcs;
 use p3_multilinear_util::point::Point;
 use p3_multilinear_util::poly::Poly;
 use p3_sumcheck::layout::{Layout, PrefixProver, Table};
-use p3_sumcheck::{OpeningBatch, OpeningProtocol, TableShape, TableSpec};
+use p3_sumcheck::{OpeningBatch, OpeningProtocol, PrescribedPointPcs, TableShape, TableSpec};
 use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
 use p3_whir::fiat_shamir::domain_separator::DomainSeparator;
```
tcoratger: **[MAJOR] This import no longer exists on `main`, and neither does the API around it — the PR cannot be rebased, only rewritten**

The PR is `CONFLICTING`. Its merge base is `a21f3cec` (2026-08-17); `main` has moved ~190 commits since, 25 of them touching `whir/`, and three of those rewrote this exact file — `9a301806 feat(whir)!: seed the WHIR transcript through the shared Fiat-Shamir layer`, `b4ef483d fix(security)!: enforce PCS budgets and compose multi-STARK soundness`, `8f908fc4 review: couple tweaks here and there`.

Concretely, against today's `main`:

- **`whir/src/fiat_shamir/` is gone.** `git ls-tree -r main -- whir/src/` has no `fiat_shamir/domain_separator.rs`; the layer is now `whir/src/transcript/mod.rs`, and a separator is obtained from the shape (`whir/src/transcript/mod.rs:597`) and installed with `separator.seed(&mut challenger)` rather than `observe_domain_separator`.
- **`add_domain_separator` no longer exists anywhere.** `git grep -n add_domain_separator main -- whir/ sumcheck/ commit/` returns nothing. This PR calls it at lines 147, 195, 236, 274, 300, 326 — six sites, plus both fixture `challenger()` methods.
- **`MultilinearPcs::{commit, open, verify}` now return `Result`**, so every call in the fixtures and every bench routine needs `.unwrap()` (or to propagate).
- **`EF` changed on `main`** from `BinomialExtensionField<F, 4>` to `QuinticTrinomialExtensionField<F>`. This branch still carries the old one, so a naive merge silently reverts it.
- **`WhirConfig::new` has been superseded** by `new_with_initial_claims(n, params, 1)`, which additionally runs `validate_initial_claims` — plain `new` no longer performs that check.

The four-argument shapes this PR relies on (`HidingWhirPcs::new`, `WhirProver::new`, `PrefixProver::new_witness`, `OpeningProtocol::new`, `TableSpec::new`, `Poly::rand`, `Point::rand`) are all still fine, so the damage is concentrated in the transcript plumbing.

I confirmed the branch **does** compile at its own merge base (`cargo check -p p3-whir --benches` and `--features parallel`, both clean, zero warnings), so this is purely drift — the benchmark logic is not at fault.

**Fix:** rebase onto `main` and port the six `add_domain_separator` sites to the `WhirShape`/`ZkWhirShape` + `.seed(challenger)` API, `.unwrap()` the `Result`-returning PCS calls, and keep `main`'s `QuinticTrinomialExtensionField` for `EF`. Re-run the numbers afterwards: `b4ef483d` changed how PCS budgets compose, so the query counts quoted below may shift.
**Verify:** `git rebase main && cargo bench -p p3-whir --bench zk_overhead --features parallel -- --test` must pass before the proof-shape table in the description is trustworthy again.

whir/benches/zk_overhead.rs
```diff
@@ -75,7 +82,11 @@ fn octic_no_pow_protocol_params() -> ProtocolParameters {
     }
 }
 
-fn octic_no_pow_config(num_variables: usize) -> ZkWhirConfig<OcticEF, F, Challenger> {
+fn octic_no_pow_plain_config(num_variables: usize) -> WhirConfig<OcticEF, F, Challenger> {
+    WhirConfig::new(num_variables, octic_no_pow_protocol_params()).unwrap()
```
tcoratger: **[MINOR] `pow_bits = 0` is sound for this comparison, but it moves the measurement into a query-heavy corner — say so**

I traced the derivation and it is safe: `whir/src/parameters/whir.rs:252-254` sets `protocol_security_level = security_level - pow_bits`, which feeds `SecurityAssumption::queries` (`security/src/assumption.rs:195-201`) as `t = ceil(lambda_eff / -log2(1 - delta))`. Dropping the 16 grinding bits therefore does **not** lower the achieved security — it buys the missing bits back with ~14% more Merkle queries, and the derived per-round grinds (`whir.rs:414-426`) then evaluate to 0 so no nonce search runs at all. Both legs are affected identically, so the plain/ZK *ratio* is clean. That is exactly what the PR set out to achieve.

What is worth spelling out is that this group is **not** a low-noise variant of `whir_zk_overhead`. Six of `ProtocolParameters`' seven fields differ between `protocol_params()` and `octic_no_pow_protocol_params()`:

| field | `protocol_params()` | `octic_no_pow_protocol_params()` |
|---|---|---|
| `security_level` | 100 | 128 |
| `pow_bits` | 16 | 0 |
| `soundness_type` | `CapacityBound` | `JohnsonBound` |
| `starting_log_inv_rate` | 2 | 1 |
| `folding_factor` | `Constant(4)` | `ConstantFromSecondRound(8, 6)` |
| `round_log_inv_rates` | derived | `vec![4]` |

Johnson gives half a `log_inv_rate` per query where Capacity gives a full one (`assumption.rs:188-192`), and `starting_log_inv_rate = 1` is the most query-expensive rate in the family. Net effect: 298 round-0 openings at rate 1/2. That is a legitimate operating point, but a reader comparing this group's ratio to the legacy group's will be comparing two different protocols.

**Fix:** add two lines to the `bench_octic_no_pow` doc comment naming the operating point (`128-bit, Johnson, start rate 2^-1, round rate 2^-4, ff 8/6, no PoW`) and stating that the ~14% extra queries are the price of `pow_bits = 0`, not a defect.
**Verify:** `ceil(128 / 0.42961) + ceil(128 / 1.92961) = 298 + 67 = 365`, which is precisely the `auth_rows` column the report prints.

whir/benches/zk_overhead.rs
```diff
@@ -75,7 +82,11 @@ fn octic_no_pow_protocol_params() -> ProtocolParameters {
...
+}
+
+fn octic_no_pow_zk_config(num_variables: usize) -> ZkWhirConfig<OcticEF, F, Challenger> {
     ZkWhirConfig::new(
         num_variables,
         octic_no_pow_protocol_params(),
```
tcoratger: **[MINOR] `round_log_inv_rates: vec![4]` is silently length-coupled to `OCTIC_OPEN_SIZES`, and these `.unwrap()`s will panic the day someone adds a size**

`octic_no_pow_protocol_params()` hard-codes a one-element `round_log_inv_rates`. `WhirConfig::new` requires that vector's length to equal `num_rounds` and otherwise returns `RoundRateCountMismatch` (`whir/src/parameters/whir.rs:313-318`), which these `.unwrap()`s (line 87 and line 98) turn into a panic.

With `ConstantFromSecondRound(8, 6)` and `MAX_NUM_VARIABLES_TO_SEND_COEFFS = 6`, `n = 18, 19, 20` all collapse to the schedule `[8, 6]`, so `num_rounds = 1` and the single `4` fits. `n = 21` gives `[8, 6, 6]`, `num_rounds = 2`, and both configs panic with an error that says nothing about `OCTIC_OPEN_SIZES`. The constant sits 200 lines away from the thing that constrains it.

**Fix:** derive it instead of hard-coding, e.g. build the schedule first and emit `vec![4; num_rounds]`; or at minimum add `// One entry per round: n in 18..=20 all fold to [8, 6], so num_rounds == 1.` next to `round_log_inv_rates` and a matching note on `OCTIC_OPEN_SIZES`.
**Verify:** temporarily append `21` to `OCTIC_OPEN_SIZES` and run `cargo bench -p p3-whir --bench zk_overhead -- --test`; it should panic in `octic_no_pow_plain_config`.

whir/benches/zk_overhead.rs
```diff
@@ -233,45 +350,296 @@ fn bench_zk_overhead(c: &mut Criterion) {
...
     }
-    group.finish();
+    eprintln!();
 }
 
-criterion_group!(benches, bench_zk_overhead, bench_octic_zk_open_no_pow);
-criterion_main!(benches);
+criterion_group!(benches, bench_zk_overhead, bench_octic_no_pow);
+criterion_group!(reports, report_octic_proof_shape);
+criterion_main!(benches, reports);
```
tcoratger, line 680: **[MINOR] `report_octic_proof_shape` runs on *every* invocation of this bench binary, including `--list` and `--test` and filtered runs**

`criterion_group!` calls each listed function with `&mut Criterion` unconditionally; Criterion's name filtering happens inside `bench_function`, which `report_octic_proof_shape` never calls. So the six `OcticPlainFixture`/`OcticZkFixture` constructions and six `build_proof()` calls at lines 559-562 execute even when the user asked for something else entirely.

That is minutes of real proving at `n = 18, 19, 20` added to:
- `cargo bench --bench zk_overhead -- --list`
- `cargo bench --bench zk_overhead -- --test` (the PR's own validation command)
- any `cargo bench --bench zk_overhead -- whir_zk_overhead_no_pow/commit` filtered run

The report itself is genuinely useful and I am not asking for it to go — the problem is that it is not opt-in. (The pre-existing `report_proof_sizes` call inside `bench_zk_overhead` has the same shape, so this is a pattern worth fixing once.)

**Fix:** gate the body on an env var, e.g. `if std::env::var_os("P3_WHIR_PROOF_REPORT").is_none() { return; }`, and mention the variable in the function's doc comment. Alternatively register it through `c.bench_function` with a no-op routine so Criterion's filter applies.
**Verify:** `cargo bench -p p3-whir --bench zk_overhead -- --list` should print the benchmark list and nothing else, in well under a second.

whir/benches/zk_overhead.rs
```diff
@@ -233,45 +350,296 @@ fn bench_zk_overhead(c: &mut Criterion) {
...
+                        plain
+                            .pcs
+                            .verify_at(
+                                &plain_commitment,
+                                &plain_proof,
+                                &protocol,
+                                &points,
+                                &mut challenger,
+                            )
+                            .unwrap();
```
tcoratger: **[MINOR] The plain verify leg drops `verify_at`'s return value *inside* the timed closure; the ZK leg has nothing to drop**

`PrescribedPointPcs::verify_at` returns `Result<Vec<OpeningEvals<EF>>, _>`, and `.unwrap();` discards that `Vec` right here — so the deallocation of 365 openings' worth of nested `Vec`s lands inside the measured region. `MultilinearPcs::verify` on the ZK side returns `Result<(), _>` (`whir/src/pcs/zk/adapter.rs:150-156`), so its routine frees nothing.

Criterion's usual protection does not apply: `iter_batched` defers dropping the value the closure *returns*, but this value is already dead before the closure ends. The bias is small relative to a millisecond-scale verify, and it points the wrong way — it makes ZK look marginally cheaper than it is, which is the one direction this benchmark exists to measure.

```rust
                        plain.pcs.verify_at(/* ... */).unwrap();   // Vec freed inside the timer
```

**Fix:** return the value instead of dropping it, so Criterion frees it outside the measured window:

```rust
|(mut challenger, protocol, points)| {
    challenger.observe(plain_commitment.clone());
    plain.pcs.verify_at(&plain_commitment, &plain_proof, &protocol, &points, &mut challenger).unwrap()
},
```

**Verify:** compare the `whir_zk_overhead_no_pow/verify/plain` mean before and after; if the delta is above noise, the old number was biased.

whir/benches/zk_overhead.rs
```diff
@@ -233,45 +350,296 @@ fn bench_zk_overhead(c: &mut Criterion) {
...
+                b.iter_batched(
+                    || {
+                        (
+                            plain.challenger(),
+                            plain.protocol.clone(),
+                            plain.points.clone(),
+                        )
+                    },
+                    |(mut challenger, protocol, points)| {
+                        challenger.observe(plain_commitment.clone());
```
tcoratger: **[NIT] This `observe` looks like it breaks the symmetry, and it does not — please say so in a comment**

I checked both entry points, because an extra transcript absorption on one leg only would invalidate the whole comparison. They have opposite conventions:

- `PrescribedPointPcs::verify_at` does **not** absorb. `whir/src/pcs/adapter.rs:254-257`: *"The commitment is not absorbed here. The caller absorbs it once before its own challenges."*
- `MultilinearPcs::verify` **does**. `whir/src/pcs/zk/adapter.rs:157`: `challenger.observe(commitment.clone());`, and `HidingWhirVerifier::verify` does not re-observe it.

So each leg pays exactly one `observe(commitment)` inside its timed closure and the comparison is fair — which the `-- --test` run corroborates, since all six verify legs succeed rather than failing on a transcript mismatch.

The hazard is the next reader. `WhirProver` *also* implements `MultilinearPcs::verify`, which absorbs internally (`whir/src/pcs/adapter.rs:144`); someone who sees `&plain_commitment` passed to a `verify*` call **and** an explicit `observe` will reasonably conclude this is a double-absorb bug and "fix" it, silently breaking the plain leg.

**Fix:** one line above this call — `// verify_at does not absorb the commitment (see PrescribedPointPcs docs); the ZK leg's MultilinearPcs::verify does it internally, so both legs pay exactly one observe.`
**Verify:** already covered — `cargo bench -p p3-whir --bench zk_overhead --features parallel -- --test` fails immediately if the absorption count ever diverges.

whir/benches/zk_overhead.rs
```diff
@@ -233,45 +350,296 @@ fn bench_zk_overhead(c: &mut Criterion) {
...
+                            .verify_at(
+                                &plain_commitment,
+                                &plain_proof,
+                                &protocol,
+                                &points,
+                                &mut challenger,
+                            )
+                            .unwrap();
+                    },
+                    BatchSize::PerIteration,
```
tcoratger: **[NIT] `BatchSize::PerIteration` is the right call for `open`, over-conservative for `commit` and `verify`**

Criterion documents `PerIteration` at roughly 350 ns of per-iteration overhead against ~500 ps for `SmallInput`, and says it should be avoided where possible. It earns its keep in the `open` group (lines 413, 424), where setup performs a full commit and batching would multiply peak memory. Here the setup is three `clone`s of a challenger, a protocol and a point list — `SmallInput` would do, and the legacy `bench_plain`/`bench_zk` legs already use it (lines 245, 282).

At millisecond-scale verifies the bias is under 0.1%, so this is style rather than a measurement defect — but it costs nothing to be consistent, and it also keeps the two legs on the same batching strategy as the rest of the file.

**Fix:** `BatchSize::SmallInput` for both legs of the `verify` group and both legs of the `commit` group (lines 375, 382); leave `open` as-is.
**Verify:** re-run the verify group; the means should shift by well under the 350 ns the change removes.

tcoratger (commented):
Replaces the zk-only `whir_zk_open_no_pow` group with a matched plain-vs-ZK, PoW-free comparison split into commit/open/verify, plus a fixed-seed proof-shape report. The methodology is sound: I verified that `ZkWhirConfig::new` passes `num_variables` and `ProtocolParameters` through **byte-identically** to `WhirConfig::new` (`whir/src/pcs/zk/config.rs:151`), so the two legs really are parameter-matched, and the report's own output confirms it — `base_values` is 76288 on **both** legs, i.e. the ZK side commits the same source geometry and hides inside the existing rate slack. Nothing leaks into production: `git diff main...HEAD --stat` is one file, and every newly-imported item (`PcsProof`, `QueryOpenings`, `SharedProofOpening`, `ZkWhirProof`, `PrescribedPointPcs`) was already `pub` before this PR. The blocker is not the benchmark: **it is a month stale against a Fiat-Shamir rewrite and cannot compile on today's `main`.**

**[Context] No one has reviewed this PR.** `pulls/1955/{comments,reviews}` and `issues/1955/comments` are all empty, so nothing below repeats or contradicts an earlier reviewer, and there is no previously-requested change left unmade.

**[Verified] `pow_bits = 0` does not silently weaken either leg — it *raises the query count*.** `whir/src/parameters/whir.rs:252-254` computes `protocol_security_level = security_level.saturating_sub(pow_bits)`, and that reduced target is the only input the query formula sees (`whir.rs:391-393`, `whir.rs:464-467`), via `security/src/assumption.rs:195-201`:

```text
t = ceil( (lambda - pow_bits) / -log2(1 - delta(rho)) )
```

For `octic_no_pow_protocol_params()` (Johnson, `lambda = 128`, `pow = 0`, start rate `2^-1`, round rate `2^-4`, `n_rounds = 1`) that is `ceil(128 / 0.42961) = 298` round-0 queries plus `ceil(128 / 1.92961) = 67` final queries, total 365 — exactly the `auth_rows 365` the report prints, and `76288 = 298 * 2^8` confirms it independently. The derived grinds at `whir.rs:414-426` and `whir.rs:471-477` then come out 0 by construction, so the prover skips grinding entirely (`whir/src/pcs/prover/mod.rs:241`). Everything else — `commitment_ood_samples`, the folding grinds, `num_variables`, the folding schedule — reads the **full** `security_level` and is untouched by `pow_bits`. So both legs still hit 128 bits, and the PR's premise holds.

**[Verified] The verify legs are transcript-symmetric despite looking asymmetric** — see my inline note on line 453.

**[MINOR] The ZK leg proves a slightly different statement.** `ZkWhirConfig` drops `commitment_ood_samples` and `final_folding_pow_bits` from the HVZK transcript (`whir/src/pcs/zk/config.rs:77-84`, and `commit_statement_hvzk` omits the `add_ood` that `commit_statement` performs). So the plain leg pays a little OOD work the ZK leg does not. That is a protocol fact, not a bench error, but "ZK overhead" measured here is *(ZK statement)* minus *(plain statement)*, not pure added cost. One sentence in the module header would keep a reader from over-reading the ratio.

**Bottom line:** do not merge as-is — rebase first, and expect a rewrite rather than a conflict resolution (see the inline note on line 26). The benchmark design itself is good and worth carrying forward.

whir/benches/zk_overhead.rs
latifkasuli: Rebased onto current main at 2ab6faaf and ported the benchmark to the fallible PCS and typed Fiat-Shamir APIs. The latest adapters now seed their run drivers internally, so I removed the obsolete manual separator state instead of double-seeding. The fixture keeps main's QuinticTrinomialExtensionField, uses new_with_initial_claims for plain WHIR, and unwraps each fallible commit/open call. The requested parallel bench test passes.

whir/benches/zk_overhead.rs
latifkasuli: Documented the no-PoW group as a distinct, query-heavy 128-bit Johnson operating point. After #1995, the former round rate 2^-4 made n=18 and n=19 infeasible because the terminal randomized source exhausted its slack, so the rebased fixture now uses a common round rate 2^-6 while retaining the 8/6 folding schedule. The PR description records that change explicitly.

whir/benches/zk_overhead.rs
latifkasuli: Fixed: the plain verify closure now returns verify_at(...).unwrap(), allowing Criterion to drop the Vec outside the timed region. The ZK closure continues to return unit.

whir/benches/zk_overhead.rs
latifkasuli: Removed the hidden one-round coupling. octic_no_pow_protocol_params now derives num_rounds from the folding schedule and builds vec![OCTIC_ROUND_LOG_INV_RATE; num_rounds], so adding a size with another round no longer causes RoundRateCountMismatch.

whir/benches/zk_overhead.rs
latifkasuli: Changed commit and verify on both legs to BatchSize::SmallInput, while retaining PerIteration for open because its setup performs a full commitment.

whir/benches/zk_overhead.rs
latifkasuli: Made both proof reports opt-in behind P3_WHIR_PROOF_REPORT. With the variable unset, --list, --test, and filtered runs do no proving or report printing. The PR description includes the opt-in command and regenerated output.

whir/benches/zk_overhead.rs
latifkasuli: Added the symmetry comment and updated the call for the current typed binding API: the plain leg calls pcs.observe_commitment before verify_at, while the ZK MultilinearPcs::verify path binds internally. The complete verify matrix passes.

PR #1956, opened 2026-07-21
perf(p3-goldilocks): add SVE2 vectorization

Al-Kindi-0:
Three additions to the aarch64 packed path, gated on cfg(target_feature = "sve2") except the NEON hardening:
- vector-domain modular mul (z-register mul/umulh, vsraq-based reduction), used by the packed Mul impl on SVE2 targets
- delayed-reduction dot product: 128-bit products accumulate unreduced in per-lane lo/hi/lo-wrap counters, one exact 192-bit reduction at the end; exact for any N (carry bound debug-asserted), covered by edge-value and wrap-saturation tests
- Poseidon2 external/internal round asm with compiler-visible scheduling; all predication pinned to two 64-bit lanes (svptrue_pat_b64(SV_VL2)), so the code is safe at any SVE vector length; full clobber declarations
- NEON round helpers canonicalize the constant operand before add/sub, closing a second-order overflow on adversarial residues (regression test included); scalar helpers document their input-domain contracts

### DIFF
--- a/goldilocks/src/aarch64_neon/poseidon2_asm.rs
+++ b/goldilocks/src/aarch64_neon/poseidon2_asm.rs
@@ -6,7 +6,9 @@
 use core::arch::aarch64::*;
 use core::arch::asm;
 
-use super::utils::{add_asm, add_canonical_asm, mul_add_asm, mul_asm};
+use p3_util::branch_hint;
+
+use super::utils::{EPSILON, add_asm, add_canonical_asm, mul_add_asm, mul_asm};
 use crate::P;
 
 /// Compute x / 2 in the Goldilocks field, matching `halve_u64::<P>`.
@@ -1330,8 +1332,12 @@ pub fn internal_permute_neon_w12(state: &mut [uint64x2_t; 12], constants: &[u64]
 
             let s0_0 = vgetq_lane_u64::<0>(s0);
             let s0_1 = vgetq_lane_u64::<1>(s0);
-            let s0_2_0 = mul_asm(s0_0, s0_0);
-            let s0_2_1 = mul_asm(s0_1, s0_1);
+            // The w12 internal round is latency-bound on this s0 dependency
+            // chain, where `multiply_p2`'s schedulable form beats `mul_asm`
+            // (Neoverse V2); w16's wider state hides the chain, so it keeps
+            // the shorter `mul_asm`.
+            let s0_2_0 = multiply_p2(s0_0, s0_0);
+            let s0_2_1 = multiply_p2(s0_1, s0_1);
 
             let sum1 = add_neon(state[1], state[2]);
             let sum2 = add_neon(state[3], state[4]);
@@ -1339,10 +1345,10 @@ pub fn internal_permute_neon_w12(state: &mut [uint64x2_t; 12], constants: &[u64]
             let sum4 = add_neon(state[7], state[8]);
             let sum5 = add_neon(state[9], state[10]);
 
-            let s0_3_0 = mul_asm(s0_2_0, s0_0);
-            let s0_3_1 = mul_asm(s0_2_1, s0_1);
-            let s0_4_0 = mul_asm(s0_2_0, s0_2_0);
-            let s0_4_1 = mul_asm(s0_2_1, s0_2_1);
+            let s0_3_0 = multiply_p2(s0_2_0, s0_0);
+            let s0_3_1 = multiply_p2(s0_2_1, s0_1);
+            let s0_4_0 = multiply_p2(s0_2_0, s0_2_0);
+            let s0_4_1 = multiply_p2(s0_2_1, s0_2_1);
 
             let sum12 = add_neon(sum1, sum2);
             let sum34 = add_neon(sum3, sum4);
@@ -1366,8 +1372,8 @@ pub fn internal_permute_neon_w12(state: &mut [uint64x2_t; 12], constants: &[u64]
             let d10 = div4_neon(state[10]);
             let d11 = div8_neon(state[11]);
 
-            let s0_7_0 = mul_asm(s0_3_0, s0_4_0);
-            let s0_7_1 = mul_asm(s0_3_1, s0_4_1);
+            let s0_7_0 = multiply_p2(s0_3_0, s0_4_0);
+            let s0_7_1 = multiply_p2(s0_3_1, s0_4_1);
             let s0_7 = vcombine_u64(vcreate_u64(s0_7_0), vcreate_u64(s0_7_1));
 
             let sum = add_neon(sum_hi, s0_7);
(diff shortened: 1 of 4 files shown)


goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1389,6 +1395,85 @@ pub fn internal_permute_neon_w12(state: &mut [uint64x2_t; 12], constants: &[u64]
...
 }
 
+// Scalar Goldilocks multiply after 0xPolygonZero/plonky2's
+// `poseidon_goldilocks_neon.rs` (MIT/Apache-2.0). Unlike `mul_asm`, whose
+// single asm block is opaque to the scheduler, these helpers are mostly plain
+// Rust, so LLVM interleaves the four s0 chains of the internal rounds. The
+// wraparound correction after the `lo - (hi >> 32)` step is a branch rather
+// than two ALU ops: it is taken with probability ~2^-32.
+// Structurally this mirrors `goldilocks.rs`'s `reduce128` with aarch64
+// micro-tunings (asm barrier, `umull` w-form); kept local pending upstream discussion.
```
Nashtare:
```suggestion
// micro-tunings (asm barrier, `umull` w-form).
```

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1389,6 +1395,85 @@ pub fn internal_permute_neon_w12(state: &mut [uint64x2_t; 12], constants: &[u64]
     state[0] = s0;
 }
 
+// Scalar Goldilocks multiply after 0xPolygonZero/plonky2's
+// `poseidon_goldilocks_neon.rs` (MIT/Apache-2.0). Unlike `mul_asm`, whose
```
Nashtare:
```suggestion
// `poseidon_goldilocks_neon.rs`. Unlike `mul_asm`, whose
```

attribution already done workspace wise

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1524,6 +1609,179 @@ pub fn internal_permute_neon<const WIDTH: usize>(
...
+    }
+}
+
+/// x^7 S-box layer over the whole state, SVE2. Six-element blocks where the
+/// width allows (six independent multiply chains cover the vector-mul latency
+/// better than four), four-element blocks otherwise.
+#[cfg(target_feature = "sve2")]
+#[inline(always)]
+unsafe fn sbox_neon<const WIDTH: usize>(state: &mut [uint64x2_t; WIDTH]) {
+    const { assert!(WIDTH % 6 == 0 || WIDTH % 4 == 0) }
```
Nashtare: not a blocker, but this const assert feels a bit weird as standalone given that the non-sve2 counterpart doesn't enforce it. Though not an issue in practice

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -3191,3 +3449,106 @@ mod tests {
...
+    use rand::rngs::SmallRng;
+    use rand::{RngExt, SeedableRng};
+
+    use super::*;
+    use crate::Goldilocks;
+
+    /// Raw operand patterns stressing every wraparound correction: field
+    /// extremes, the epsilon window, and non-canonical values up to u64::MAX.
+    /// The ops under test accept any u64 residue.
+    const EDGE: [u64; 8] = [0, 1, (1 << 32) - 1, 1 << 32, 1 << 63, P - 1, P, u64::MAX];
```
Nashtare: nit: defined twice

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1524,6 +1609,179 @@ pub fn internal_permute_neon<const WIDTH: usize>(
...
+/// width allows (six independent multiply chains cover the vector-mul latency
+/// better than four), four-element blocks otherwise.
+#[cfg(target_feature = "sve2")]
+#[inline(always)]
+unsafe fn sbox_neon<const WIDTH: usize>(state: &mut [uint64x2_t; WIDTH]) {
+    const { assert!(WIDTH % 6 == 0 || WIDTH % 4 == 0) }
+    unsafe {
+        if WIDTH % 6 == 0 {
+            for chunk in state.chunks_exact_mut(6) {
+                sbox6_sve2(chunk.try_into().unwrap());
```
Nashtare: wouldn't `as_chunks_mut()` work here? To remove the unnecessary unwrap

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1524,6 +1609,179 @@ pub fn internal_permute_neon<const WIDTH: usize>(
...
+            }
+        } else {
+            for chunk in state.chunks_exact_mut(4) {
+                sbox4_sve2(chunk.try_into().unwrap());
```
Nashtare: same here

goldilocks/src/aarch64_neon/utils.rs
```diff
@@ -5,7 +5,7 @@ use core::arch::asm;
 use super::packing::PackedGoldilocksNeon;
 use crate::{Goldilocks, P};
 
-const EPSILON: u64 = P.wrapping_neg(); // 2^32 - 1
+pub(super) const EPSILON: u64 = P.wrapping_neg(); // 2^32 - 1
```
Nashtare, line 8: Can also be reused in `packing.rs` module

goldilocks/src/aarch64_neon/poseidon2_asm.rs
Nashtare: also to silence clippy:


```suggestion
    const { assert!(WIDTH.is_multiple_of(6) || WIDTH .is_multiple_of(4)) }
```

Nashtare (commented):
Thanks! I'm getting ~9% improvement on Poseidon2 with width 12!
Although as we're adding a new target this would need a dedicated step in CI to make sure it compiles fine at least

goldilocks/src/aarch64_neon/poseidon2_asm.rs
```diff
@@ -1524,6 +1609,179 @@ pub fn internal_permute_neon<const WIDTH: usize>(
...
+
+/// x^7 S-box layer over the whole state, SVE2. Six-element blocks where the
+/// width allows (six independent multiply chains cover the vector-mul latency
+/// better than four), four-element blocks otherwise.
+#[cfg(target_feature = "sve2")]
+#[inline(always)]
+unsafe fn sbox_neon<const WIDTH: usize>(state: &mut [uint64x2_t; WIDTH]) {
+    const { assert!(WIDTH % 6 == 0 || WIDTH % 4 == 0) }
+    unsafe {
+        if WIDTH % 6 == 0 {
```
Nashtare: same here for `is_multiple_of` and below

goldilocks/src/aarch64_neon/poseidon2_asm.rs
Al-Kindi-0: Kept the assert, added the missing rationale

Nashtare (approved):
LGTM, thanks again!

PR #1959, opened 2026-07-21
test(whir): clarify and extend ZK simulation coverage

latifkasuli:
## Summary

This is a test- and documentation-only change; it does not alter protocol behavior.

- Add a 64-case property test showing that two fresh pad coordinates can jointly program arbitrary answers at two distinct, nonzero OOD points by solving the corresponding 2 × 2 Vandermonde system.
- Add simulator replay coverage with nonzero proof of work (`pow_bits = 4`), including verifier acceptance.
- Rename the simulator “view matches” tests to state what they establish: verifier acceptance and matched-RNG mask-prelude coupling for prefix and suffix variable orders.
- Document that `simulate_classic_unpacked` covers one Construction 6.3 sumcheck batch and does not return the mask data and encoding randomness needed to compose a full HVZK-WHIR simulator.
- Rename the conditioned-witness round-trip helper and test to make clear that they run the honest prover with an alternate witness on the same claim fiber.

## Motivation

The previous test names and comments implied a stronger simulation result than their assertions established. This change records the exact hiding invariants currently covered while keeping the missing composed-simulator boundary explicit. In particular, the conditioned-witness round trip establishes completeness, not witness-free simulation or Theorem 4.5 transcript indistinguishability.

This follows up on the ZK-WHIR coverage introduced in #1767.

## Checks

- `git diff --check`
- `cargo test -p p3-sumcheck -p p3-whir`
  - `p3-sumcheck`: 240 passed
  - `p3-whir`: 118 passed, 1 ignored
- `cargo clippy -p p3-sumcheck -p p3-whir --all-targets -- -D warnings`

### DIFF
--- a/sumcheck/src/zk/simulator.rs
+++ b/sumcheck/src/zk/simulator.rs
@@ -25,6 +25,13 @@ use super::verifier::ZkVerifier;
 /// - Sample each wire coordinate uniformly over `EF` (the honest joint distribution).
 /// - Reconstruct the dropped `c_1` from the affine identity, so every simulated wire verifies by construction.
 ///
+/// # Scope
+///
+/// This simulates one Construction 6.3 sumcheck batch, not the composed
+/// HVZK-WHIR protocol. It discards the mask oracle's prover data and does not
+/// return the mask messages or encoding randomness needed for later
+/// code-switch and base-case openings.
+///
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
LGTM, thanks!

PR #1960, opened 2026-07-21
fix(whir): require cryptographic RNG for hiding PCS

latifkasuli:
## Summary

Require `R: CryptoRng` at the `HidingWhirPcs` constructor and `MultilinearPcs` implementation boundaries.

The ZK tests and benchmarks now use `StdRng` for PCS hiding randomness. Their unrelated random fixture generation remains on `SmallRng`.

## Why

All WHIR hiding randomness—including encoding masks, sumcheck masks, OOD pads, and base-case one-time pads—ultimately comes from the caller-supplied RNG.

Previously, the API accepted any `Rng`. Although each `commit` or `open` operation seeds a fresh `StdRng`, seeding a CSPRNG from a predictable source does not create entropy. A caller could therefore instantiate the hiding PCS with a reproducible non-cryptographic stream, allowing an observer who reconstructs that stream to remove the masks.

`CryptoRng` rules out RNG algorithms known to be unsuitable for cryptographic use. It does not replace proper seeding: production callers must still initialize the RNG from unpredictable entropy. Deterministic seeds remain appropriate for tests and benchmarks.

## API impact

> [!IMPORTANT]
> This is a source-breaking API hardening.

Downstream `HidingWhirPcs` users whose RNG only implements `Rng`, such as callers using `SmallRng`, will need to supply a `CryptoRng` implementation instead.

This PR is intentionally limited to `HidingWhirPcs`. `HidingFriPcs` retains its existing broader RNG bound and is out of scope; API consistency there can be considered separately.

## Checks

- `cargo check -p p3-whir --benches`
- `cargo test -p p3-whir pcs::zk` — 49 passed
- `cargo test -p p3-whir --all-features` — 117 passed, 1 ignored
- `cargo check -p p3-whir --no-default-features`
- `cargo clippy -p p3-whir --all-targets -- -D warnings`
- `cargo doc -p p3-whir --no-deps`

### DIFF
--- a/whir/benches/zk_overhead.rs
+++ b/whir/benches/zk_overhead.rs
@@ -27,7 +27,7 @@ use p3_whir::parameters::{FoldingFactor, ProtocolParameters, SecurityAssumption,
 use p3_whir::pcs::prover::WhirProver;
 use p3_whir::pcs::zk::{HidingWhirPcs, ZkParameters, ZkWhirConfig};
 use rand::SeedableRng;
-use rand::rngs::SmallRng;
+use rand::rngs::{SmallRng, StdRng};
 
 type F = KoalaBear;
 type EF = BinomialExtensionField<F, 4>;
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
Good catch, thanks!

PR #1921, opened 2026-07-06
fix: patch merkle-tree soundness gap and cleanup some quality findings

Nashtare:
Fixes a soundness issue in `verify_batch_pruned` (non-lead group members' injected-matrix rows went unauthenticated, reachable via WHIR's verifier) and a NEON add/sub overflow causing potential goldilocks Poseidon2 hash divergence on aarch64. Also resolves a couple code quality issues: silent truncation/overflow guards, cfg-gating gaps, const-assertion hardening, and non-injective transcript encoding in whir's domain separator.

PR #1962, opened 2026-07-25
fix(hiding): require a CSPRNG and fork it on clone in the FRI path

wstran:
## What

The FRI hiding path draws its masking randomness from a caller-supplied generator, and two guarantees the rest of the tree already has were missing on it.

Root cause, two halves of the same gap:

- `HidingFriPcs` and `MerkleTreeHidingMmcs` bounded their RNG by `Rng`, not `CryptoRng`. A caller could instantiate either with a reproducible non-cryptographic stream, and an observer who reconstructs that stream removes the masks. This is the exposure #1960 closed for `HidingWhirPcs`; its body scopes itself to WHIR, so the FRI path was left behind.
- `MerkleTreeHidingMmcs::clone` copied the generator rather than reseeding one. The usual wiring is `ChallengeMmcs::new(val_mmcs.clone())`, so the extension-field commitments salted their leaves from the same sequence as the trace commitment. This matters because salts travel in the proof: `Mmcs::Proof` is `(salts, siblings)` and `open_batch` hands the verifier the salt of every opened row, so a salt learned from one commitment is a salt already spent on the other. #1921 gave `HidingFriPcs::clone` exactly this protection, with the comment "so the clone and the original never produce the same sequence of masks"; the MMCS it wraps was not covered. `HidingWhirPcs` sidesteps the question by having no `Clone` impl at all and deriving a fresh `StdRng` per `commit` and `open`.

Fix: bound both on `CryptoRng`, and fork the stream in the MMCS clone with `R::from_rng`, mirroring `HidingFriPcs`.

## Why the two halves cannot be separated

The old bound did not merely fail to enforce the documented requirement. It made that requirement unsatisfiable. `Mmcs` requires `Self: Clone`, the old impl reached that through `R: Clone`, and the cryptographic generators in `rand` do not implement `Clone`:

| generator | `Rng` | `CryptoRng` | `Clone` | `Send` | usable under `R: Rng + Clone + Send` |
|---|---|---|---|---|---|
| `StdRng` | yes | yes | **no** | yes | no |
| `ChaCha20Rng` | yes | yes | **no** | yes | no |
| `SmallRng` | yes | **no** | yes | yes | yes |
| `SysRng` | **no** | no | yes | yes | no |
| `ThreadRng` | yes | yes | yes | **no** | no |

So the only `rand` generator the old signature accepted was `SmallRng`, which is not cryptographic. The doc suggested `ThreadRng`, which the `Send` bound had already ruled out. Tightening to `CryptoRng` is therefore only possible together with replacing the `R: Clone` route to `Clone`:

```text
    before   R: Rng + Clone               clone -> same salt stream
    after    R: CryptoRng + SeedableRng   clone -> R::from_rng(parent)
```

`SysRng` and `ThreadRng` were unusable before this change and remain so; nothing that compiled against a cryptographic generator stops compiling.

## Correctness

The requirement was written down twice in prose and enforced neither time:

- `MerkleTreeHidingMmcs` docs asked for "an appropriately seeded cryptographically secure pseudorandom number generator (CSPRNG)".
- `poseidon2-air/examples/prove_poseidon2_baby_bear_keccak_zk.rs` carried `// WARNING: DO NOT USE SmallRng in proper applications! Use a real PRNG instead!`.

The type system now carries both. The example comment is redundant and is removed.

New regression test `clone_forks_the_salt_stream` commits one matrix through an MMCS and through its clone and requires the commitments to differ. It is not vacuous: two generators built from the same seed still produce equal commitments, so the old copying clone fails it.

2852 workspace tests pass, `clippy --all-targets -- -D warnings` is clean, and nightly `fmt --check` is clean.

## Cost

Moving the in-repo ZK example off a non-cryptographic generator is not free, so here is the number rather than a claim that it is negligible.

| `prove_poseidon2_baby_bear_keccak_zk` | median | min | max |
|---|---|---|---|
| before (`SmallRng`) | 12.44s | 12.37s | 14.45s |
| after (`StdRng`) | 12.74s | 12.69s | 14.61s |

`+2.41%` end to end. Medians of 11 interleaved runs, `-Ctarget-cpu=native`, M3, alternating the two binaries to cancel thermal drift.

Salts are drawn per leaf, so a stronger generator costs something. Since the old signature admitted no cryptographic generator at all, this is the cost of the configuration becoming available rather than a regression against one that was already reachable.

```text
repro: cargo run --release --example prove_poseidon2_baby_bear_keccak_zk -p p3-poseidon2-air
```

## Scope

Bounds and the MMCS clone only; no proof format, no wire format, and no prover or verifier logic changes. Callers move their hiding randomness to `StdRng` while unrelated fixture generation stays on `SmallRng`, the same split #1960 used. `StdRng` needs the `std_rng` feature, which `whir` already enabled and which is added here for `merkle-tree`, `fri`, `uni-stark`, `batch-stark`, and `poseidon2-air`.

This tightens a public bound, so downstream code passing `SmallRng` to `HidingFriPcs` or `MerkleTreeHidingMmcs` stops compiling. That is the intent, and it matches the break #1960 already took for `HidingWhirPcs`.

### DIFF
--- a/batch-stark/Cargo.toml
+++ b/batch-stark/Cargo.toml
@@ -43,7 +43,7 @@ p3-symmetric = { path = "../symmetric" }
 
 criterion.workspace = true
 postcard = { workspace = true, features = ["alloc"] }
-rand.workspace = true
+rand = { workspace = true, features = ["std_rng"] }
 
 [[bench]]
 name = "prove_batch"
--- a/batch-stark/tests/simple.rs
+++ b/batch-stark/tests/simple.rs
@@ -30,7 +30,7 @@ use p3_symmetric::{
 use p3_uni_stark::{InvalidProofShapeError, PeriodicColumnError, StarkConfig};
 use p3_util::{assert_clone, assert_send, assert_sync, log2_strict_usize};
 use rand::SeedableRng;
-use rand::rngs::SmallRng;
+use rand::rngs::{SmallRng, StdRng};
 
 const TWO_ADIC_FIXTURE: &str = "tests/fixtures/batch_stark_two_adic_v1.postcard";
 const CIRCLE_FIXTURE: &str = "tests/fixtures/batch_stark_circle_v1.postcard";
@@ -507,7 +507,7 @@ type HidingValMmcs = MerkleTreeHidingMmcs<
     <Val as Field>::Packing,
     MyHash,
     MyCompress,
-    SmallRng,
+    StdRng,
     2,
     8,
     4,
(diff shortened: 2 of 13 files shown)


Nashtare (approved):
Thanks!

PR #1963, opened 2026-07-25
feat(whir): expose hiding base-case security report

latifkasuli:
## Summary

Add a non-enforcing security report for the terminal HVZK-WHIR base case.

- Report the folded source and every chronological mask group.
- Derive each randomized-code dimension as `message_len + randomness_len`.
- Show the exact randomized rate and the conservative next-power-of-two RS envelope used by the existing integer-rate security helpers.
- Compose query misses with the heterogeneous maximum rule, while exposing the shared PoW credit separately as Fiat-Shamir/compiler-conditional.
- Compose interleaved MCA errors as probabilities and candidate-list sizes as a product.
- Use `floor(log2(|EF|))` as a conservative field-cardinality input; retain a fractional `log2(|EF|)` estimate for diagnostics only.
- Distinguish proven, conditional, approximate, unavailable, and vacuous results.

`ZkWhirConfig::new` is unchanged. The report does not reject configurations or change any proof/transcript parameter.

## Motivation

The current `t_zk` derivation covers the mask query-miss branch, but it does not account for:

```text
epsilon_gamma <= M_source + sum_i M_i
                 + L_source * product_i L_i / |EF|

epsilon_query <= max(source_miss, max_i mask_miss_i)
```

Here each randomized source or mask code can have a different dimension, rate, radius, and interleaving width. In particular, the source dimension must include the terminal oracle's encoding randomness; reusing the plain-WHIR rate overstates that branch.

This API makes those obligations inspectable without presenting them as an end-to-end certificate.

## Concrete diagnostic

For the stock `zk_overhead` Capacity configuration at `n = 16`, target 100 bits:

```text
terminal source: message = 16, randomness = 11, domain = 4096
actual dimension: 27
dyadic analysis dimension: 32
analysis log inverse rate: 7
source query bound: 76.225717 bits
shared final PoW: 13 bits
post-PoW estimate: 89.225717 bits
```

The source branch is therefore below the requested 100 bits even after the compiler-conditional PoW estimate. The existing mask branches use 21 queries at analysis rate `2^-5`, but they are not the binding query component.

For the same Capacity fixture, the generic product-list/field bound is vacuous. This is a certification gap, not evidence of an attack.

### Query-coordinate fixed point

For this fixture, the randomized terminal source has `N = 4096` and dimension `k = 16 + t`. With the configured maximum grind of 16 bits, re-deriving the query count against the conservative dyadic envelope gives:

```text
t = ceil((100 - 16) / (7 - log2(21 / 20))) = 13
13 * (7 - log2(21 / 20)) = 90.0849387 bits
```

At `t = 13`, `k = 29` still rounds to 32, so the analysis log-inverse rate remains 7 and the calculation is a fixed point. A derived 10-bit final grind would give `100.0849387` query-coordinate bits. The current `11 queries + 13 PoW` pair was derived from the inherited message-code rate and reaches only `89.2257174` bits once randomized-code occupancy is included.

This PR reports that discrepancy but does not change parameter derivation. The fixed point also does not certify the stock configuration: the shared-`gamma` Capacity terms remain conditional, and the generic product-list/field bound remains vacuous.

## Bound provenance

- `Proven`: an established bound under the premises documented for that term. In particular, a proven query miss classifies the fixed-distance sampling implication only; it does not promote Capacity list/MCA terms or the full regime to proven.
- `Approximation`: the Johnson MCA helper currently keeps only the dominant BCSS25 term.
- `Conditional`: Capacity-rate coding conjectures, or the post-PoW estimate under the Fiat-Shamir grinding model.
- `Unavailable`: no positive-radius dyadic envelope is available to the encoded helpers.
- `vacuous = true`: the available probability upper bound is at least one; this is not an insecurity claim.

Aggregate classifications conservatively include both RBR coordinates even when the less-proven coordinate is not numerically binding. The fractional field-size estimate is approximately `123.95473875` bits for quartic KoalaBear. Field-dependent calculations deliberately use the conservative integer lower bound of 123 bits because `libm::log2` does not promise directed rounding; this costs less than one bit relative to the exact value.

## Scope

This report covers the two RBR coordinates of the heterogeneous ideal-IOP base case only. It does **not** certify:

- earlier masked-sumcheck or code-switch reductions;
- Merkle hiding/binding or hash collision resistance;
- Fiat-Shamir extraction or the random-oracle compiler;
- simulation of the commitment view;
- end-to-end PCS security.

The heterogeneous theorem contract and projected-query variant are discussed in draft PR #1957. This PR targets current `main` and reports its physical mask groups independently.

## Crate boundary

The review-requested extraction now follows the repository's documented dependency direction:

- `p3-security::whir::hiding` owns the protocol-neutral geometry input, report types, field-cardinality handling, classified bound arithmetic, heterogeneous query composition, and shared-`gamma` composition.
- `p3-whir` retains the inherent `ZkWhirConfig` adapter that reconstructs the folded source and chronological mask groups, applies their widths and query schedules, and identifies the one shared final PoW.
- Existing `p3-whir` report names remain source-compatible through explicit re-exports.

The STARK-oriented `RegimeReport` API is intentionally not forced onto these terms because it does not encode `Conditional`, `Approximation`, `Unavailable`, or vacuous provenance. The extracted geometry constructor validates its release-mode invariants, and the shared `log_1_delta` helper now uses algebraically stable formulas at large inverse rates.

## Separate follow-ups

- Derive the hiding-aware final query/PoW pair as a fixed point, with threshold-crossing and infeasibility tests. This is behavior-changing and is intentionally outside this diagnostic PR.
- Replace `p3-security::ErrorBits::sum` with the minimum-shift/log-sum-exp composition used here so very large bit values cannot underflow to an incorrect infinity result.
- Thread a rigorous field-cardinality representation through the enforcing WHIR parameter path. It currently uses `EF::bits()`, which rounds quartic KoalaBear upward to 124 rather than approximately `123.9547` bits.

## Validation

- `cargo test -p p3-security` (39 passed)
- `cargo test -p p3-whir --all-features` (124 passed, 1 ignored)
- `cargo clippy -p p3-security -p p3-whir --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p p3-security -p p3-whir --all-features --no-deps`
- `cargo check -p p3-whir --bench zk_overhead --all-features`
- `cargo +nightly fmt --all --check`
- `git diff --check`

Regression coverage pins:

- fractional field estimate versus rigorous field floor;
- source randomness in the analyzed RS dimension;
- the 89.225717-bit stock post-PoW diagnostic;
- chronological mask roles and rates;
- maximum query composition and one shared PoW credit;
- heterogeneous MCA probability composition;
- UD/JB/Capacity provenance propagation;
- nonbinding approximate/conditional provenance propagation;
- the interleaving-stable Lemma 3.13 list branch;
- exact stock Capacity `gamma`-coordinate magnitudes and vacuity;
- proximity-radius consistency with the query-error helper;
- saturated requested/effective query counts;
- unavailable rate-one dyadic envelopes;
- vacuous theorem-backed terms being excluded by `proven_bits()`;
- invalid public code geometry being rejected in release builds;
- stable `log_1_delta` evaluation at large inverse rates.

## Review focus

The main review questions are:

1. Is the heterogeneous Construction 7.2 composition contract accepted?
2. Is the proposed `p3-security`/`p3-whir` boundary the right extraction seam?
3. Is the provenance/vacuity taxonomy the right public shape?
4. Should a future implementation replace the integer field floor with a rigorously directed fractional bound?


whir/src/pcs/zk/security.rs
```diff
@@ -0,0 +1,1019 @@
...
+fn interleaved_list_bits(
+    assumption: SecurityAssumption,
+    log_degree: usize,
+    log_inv_rate: usize,
+    interleaving: usize,
+) -> f64 {
+    let base = assumption.list_size_bits(log_degree, log_inv_rate);
+    let crude = interleaving as f64 * base;
+
+    // Lemma 3.13's interleaving-stable alternative:
```
Nashtare: nit: add where the Lemma is coming from.

whir/src/pcs/zk/security.rs
Nashtare, line 1: May be worth plugging this into `p3-security` as a dedicated sub-module for WHIR?

whir/src/pcs/zk/security.rs
```diff
@@ -0,0 +1,1019 @@
...
+    match assumption {
+        SecurityAssumption::UniqueDecoding => 0.5 * (1. - rho),
+        SecurityAssumption::JohnsonBound => {
+            1. - libm::sqrt(rho) - libm::pow(2., assumption.log_eta(log_inv_rate))
+        }
+        SecurityAssumption::CapacityBound => {
+            1. - rho - libm::pow(2., assumption.log_eta(log_inv_rate))
+        }
+    }
+}
```
Nashtare: This mirrors `log_1_delta()` of `SecurityAssumption` from `p3-security` crate, could be removed.

whir/src/pcs/zk/security.rs
```diff
@@ -0,0 +1,1019 @@
...
+        SecurityAssumption::CapacityBound => HidingBoundClassification::Conditional,
+    }
+}
+
+const fn combine_classifications(
+    left: HidingBoundClassification,
+    right: HidingBoundClassification,
+) -> HidingBoundClassification {
+    use HidingBoundClassification::{Approximation, Conditional, Proven, Unavailable};
+
```
Nashtare: nit: unneeded?
```suggestion
```

whir/src/pcs/zk/security.rs
```diff
@@ -0,0 +1,1019 @@
...
+    /// This does not assert that sibling terms, a selected coding regime, or
+    /// the compiled PCS are proven end to end.
+    #[must_use]
+    pub const fn proven_bits(&self) -> Option<f64> {
+        match (self.classification, self.bits) {
+            (HidingBoundClassification::Proven, bits) => bits,
+            _ => None,
+        }
+    }
+}
```
Nashtare: nit: this returns `Some(0.0)` I believe for a vacuous `Proven` term, whereas this edge case may warrant a `None` to not confuse honest readers of the method output (filtering from the Option type).

whir/src/pcs/zk/security.rs
```diff
@@ -0,0 +1,1019 @@
...
+    F: TwoAdicField,
+    EF: ExtensionField<F> + TwoAdicField,
+    Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F>,
+{
+    /// Reports the heterogeneous hiding base case's ideal-IOP RBR terms.
+    ///
+    /// This is diagnostic only: it neither rejects the configuration nor
+    /// certifies the compiled PCS or the preceding hiding reductions.
+    #[must_use]
+    pub fn hiding_base_case_security_report(&self) -> HidingBaseCaseSecurityReport {
```
Nashtare, line 30: I don't think it's ever used, may be a good usecase to display it in the `zk_overhead` example?

Nashtare (approved):
LGTM overall, thanks for this! I think we can reduce a bit of code dedup by grouping the security module into `p3-security` crate, otherwise this is really nice!

security/src/whir/mod.rs
Nashtare, line 1: Ah I thought this was already a module. May be worth folding `security/src/whir.rs` into `security/src/whir/mod.rs` just for clarity, otherwise thanks a bunch!

Nashtare (approved):

PR #1964, opened 2026-07-27
perf(p3-goldilocks): faster Goldilocks wasm simd128 backend

Nashtare:
Yields from 23% to 31% reduction for Poseidon2 instances over Goldilocks, 1.8x to 2.6x faster dot product with WASM SIMD128 backend. `batched_linear_combination` is also 11% to 15% faster and `sum_array` is 28% to 43% faster depending on `N`.

### DIFF
--- a/goldilocks/examples/wasm_merkle_bench.rs
+++ b/goldilocks/examples/wasm_merkle_bench.rs
@@ -1,0 +1,76 @@
+//! End-to-end-ish benchmark for wasm32+simd128: builds a Merkle tree over a matrix of
+//! Goldilocks field elements using the specialized `PackedGoldilocksWasmSimd128` Poseidon2
+//! permutation for both leaf hashing and internal-node compression — the dominant real cost
+//! in Merkle-tree-heavy STARK proving. Serves as a realistic baseline for judging whether
+//! further Poseidon2-level optimizations (e.g. software-pipelined dual-permutation hashing)
+//! are worth pursuing, closer to a real workload than the isolated single-permutation
+//! microbenchmark in `wasm_bench.rs`.
+//!
+//! Run with:
+//! ```text
+//! RUSTFLAGS="-C target-feature=+simd128" \
+//!   cargo run --release --target wasm32-wasip1 \
+//!     --example wasm_merkle_bench -p p3-goldilocks
+//! ```
+//! On non-wasm32 targets this binary is a no-op.
+
+#[cfg(not(all(target_arch = "wasm32", target_feature = "simd128")))]
+fn main() {
+    eprintln!(
+        "wasm_merkle_bench is a no-op on this target. Re-run with \
+         `cargo run --release --target wasm32-wasip1 --example wasm_merkle_bench -p p3-goldilocks` \
+         and RUSTFLAGS=\"-C target-feature=+simd128\"."
+    );
+}
+
+#[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
+fn main() {
+    use std::time::Instant;
+
+    use p3_commit::Mmcs;
+    use p3_field::Field;
+    use p3_goldilocks::{Goldilocks, default_goldilocks_poseidon2_8};
+    use p3_matrix::dense::RowMajorMatrix;
+    use p3_merkle_tree::MerkleTreeMmcs;
+    use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
+    use rand::rngs::SmallRng;
+    use rand::{RngExt, SeedableRng};
+
+    type Perm = p3_goldilocks::Poseidon2Goldilocks<8>;
+    type Hasher = PaddingFreeSponge<Perm, 8, 4, 4>;
+    type Compressor = TruncatedPermutation<Perm, 2, 4, 8>;
+    type Packing = <Goldilocks as Field>::Packing;
+    type Mmcs8 = MerkleTreeMmcs<Packing, Packing, Hasher, Compressor, 2, 4>;
+
+    fn report(name: &str, n: u64, elapsed_ns: u128) {
+        let per_op = elapsed_ns as f64 / n as f64;
+        println!("{name:>28}: {per_op:>10.1} ns/leaf   ({n} leaves in {elapsed_ns} ns total)");
+    }
+
+    let perm = default_goldilocks_poseidon2_8();
+    let hasher = Hasher::new(perm.clone());
+    let compressor = Compressor::new(perm);
+    let mmcs = Mmcs8::new(hasher, compressor, 0);
+
+    let mut rng = SmallRng::seed_from_u64(0xBA5E);
+
+    // Sweep leaf counts spanning small (few tree levels) to large (many levels, more
+    // compression work relative to leaf hashing) so the pipelining question — does
+    // batching independent permutations help — has enough range to show a trend.
+    for &log_leaves in &[8usize, 12, 16] {
+        let rows = 1usize << log_leaves;
+        const COLS: usize = 8;
+        let values: Vec<Goldilocks> = (0..rows * COLS).map(|_| rng.random()).collect();
+        let matrix = RowMajorMatrix::new(values, COLS);
+
+        let t0 = Instant::now();
+        let (_commitment, _prover_data) = mmcs.commit(vec![matrix]);
+        let elapsed_ns = t0.elapsed().as_nanos();
+
+        report(
+            &format!("merkle_2^{log_leaves}_leaves"),
+            rows as u64,
+            elapsed_ns,
+        );
+    }
+}
--- a/goldilocks/src/bin/wasm_bench.rs
+++ b/goldilocks/src/bin/wasm_bench.rs
@@ -141,7 +141,7 @@ fn main() {
 }
 
 /// Times the packed Goldilocks quadratic-extension multiplication and squaring
-/// (`quadratic_mul`/`binomial_square` in `p3_field::extension`, applied to
+/// (`binomial_mul`/`binomial_square` in `p3_field::extension`, applied to
 /// `[PackedGoldilocksWasmSimd128; 2]`), which route through the vectorized `dot_product`.
 #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
 fn bench_quadratic_extension_mul() {
--- a/goldilocks/src/bin/wasm_bench.rs
+++ b/goldilocks/src/bin/wasm_bench.rs
@@ -220,6 +220,73 @@ fn bench_dot_product() {
     bench_n!(8);
     bench_n!(16);
     bench_n!(32);
+
+    bench_batched_linear_combination();
+}
+
+/// Sweeps `chunked_linear_combination::<CHUNK, ...>` over every `CHUNK` size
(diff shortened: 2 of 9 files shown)


Nashtare:
@tcoratger It would be nice if we could get this merged soon, as there is a goldilocks bug that would ideally be released soon to not affect dependent projects too much!

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -164,13 +164,36 @@ impl PrimeCharacteristicRing for PackedGoldilocksWasmSimd128 {
...
+        const {
+            assert!((N as u32) <= (1 << 31));
+        }
+        match N {
+            0 => Self::ZERO,
+            1 => lhs[0] * rhs[0],
+            _ => Self::from_vector(dot_product_delayed_reduce::<N>(
+                &core::array::from_fn(|i| lhs[i].to_vector()),
+                &core::array::from_fn(|i| rhs[i].to_vector()),
+            )),
```
tcoratger: These two `core::array::from_fn` calls don't get elided, and it costs real time. `to_vector` is a transmute so this looks free, but rustc leaves the two allocas in the wasm module and Cranelift then does actual stores and reloads. From your own `wasm_bench` (wasmtime v44.0.1, aarch64, ns/op):

```
   N |  as-is | helper takes &[Self; N]
-----+--------+-------------------------
   2 |  10.13 |                   10.18
   3 |  13.24 |                   13.54
   4 |  17.02 |                   18.00
   5 |  48.49 |                   24.10
   8 |  61.54 |                   44.75
  16 |  92.34 |                   67.27
  32 | 167.54 |                  170.21
```

The jump from 17.02 at N=4 to 48.49 at N=5, for 25% more arithmetic, is where the temporaries stop being promoted to registers. N=4 is `BATCHED_LC_CHUNK` and N=8/16 are typical MDS/Merkle widths, so it's the range that matters most.

What I benchmarked was keeping one accumulator loop and feeding it lane pairs, so nothing is materialized:

```rust
#[inline]
fn dot_pairs<const N: usize>(get: impl Fn(usize) -> (v128, v128)) -> v128 {
    let mut acc_lo_hi = u64x2_splat(0);
    let mut acc_lo_lo = u64x2_splat(0);
    let mut acc_hi96  = u64x2_splat(0);
    for i in 0..N {
        let (l, r) = get(i);
        let (term_hi, term_lo) = mul64_64(l, r);
        // rest unchanged
    }
    // tail unchanged
}

// dot_product
_ => Self::from_vector(dot_pairs::<N>(|i| (lhs[i].to_vector(), rhs[i].to_vector()))),

// mixed_dot_product
_ => Self::from_vector(dot_pairs::<N>(|i| (
        a[i].to_vector(),
        Self::from(f[i]).to_vector(),
    ))),
```

Same change applies to `mixed_dot_product` just below. It also saves 2*16*N bytes of stack, which is worth having on wasm32's 1 MiB default stack.

The temporary in `sum_array` above does get elided, I measured it both ways and it's within noise, so that one is only worth changing for symmetry.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -193,15 +216,30 @@ impl_packed_field_div!(PackedGoldilocksWasmSimd128);
...
+    // Benchmarked across slice lengths 8, 16, 33, 64, 256 under both wasmtime/Cranelift and
+    // Node/V8, since the two engines disagree sharply on the best chunk: Cranelift likes
+    // chunk=32/64 for aligned lengths (falling back to the unvectorized per-element
+    // remainder path in `chunked_linear_combination` otherwise), while V8 prefers chunk=16
+    // and regresses badly (2.2-2.3x slower than optimal) at chunk=32/64 regardless of
+    // alignment. chunk=4 is the min-max choice: worst-case 1.38x slower than the
+    // best-for-that-engine-and-length chunk across all 10 (engine, length) combinations
+    // tested, versus 1.57-1.59x for chunk=8/16 and 2.2x+ for chunk=32/64 — always
+    // reasonable, never catastrophic, on either engine.
+    const BATCHED_LC_CHUNK: usize = 4;
```
tcoratger, line 212: Worth re-deriving this once the temporaries above are gone, because part of what the sweep is measuring is that codegen artifact rather than the algorithm. Ratio of chunk=4 to the best chunk for each length, same run:

```
 len | as-is | temporaries removed
-----+-------+--------------------
   8 | 1.00x |               1.01x
  16 | 1.00x |               1.01x
  33 | 1.20x |               1.03x
  64 | 1.28x |               1.06x
 256 | 1.28x |               1.07x
```

So chunk=4 goes from up to 1.28x off optimal to at most 1.07x, and the big regression at chunk=32/64 that the comment describes mostly disappears on Cranelift. Absolute chunk=4 times improve too (len 64: 379 to 357 ns, len 256: 1448 to 1395 ns). chunk=4 still looks like a fine choice, the reasoning in the comment just changes.

I could only test wasmtime locally, so I don't know whether V8 still prefers 16 once this is fixed. Did you see the N=4 to N=5 cliff on V8 as well?

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -164,13 +164,36 @@ impl PrimeCharacteristicRing for PackedGoldilocksWasmSimd128 {
         unsafe { reconstitute_from_base(Goldilocks::zero_vec(len * WIDTH)) }
     }
 
+    #[inline]
+    fn sum_array<const N: usize>(input: &[Self]) -> Self {
+        assert_eq!(N, input.len());
+        const {
+            assert!((N as u32) <= (1 << 31));
+        }
```
tcoratger: This bound isn't needed for `sum_array`. It's load-bearing in `dot_product`, where `acc_hi96` sums N values each below 2^32 and `P - acc_hi96` must not underflow (first dangerous N is 2^32 + 1, so 2^31 is a correct 2x margin, matching the scalar impl). For `sum_array` the doc comment on `sum_delayed_reduce` already says the exact sum holds for any N < 2^64, so the assert implies a limit that doesn't exist. I'd drop it here and let the doc carry the argument.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -487,4 +602,234 @@ mod tests {
...
+        check_edge_n!(5);
+        check_edge_n!(7);
+        check_edge_n!(8);
+        check_edge_n!(11);
+        check_edge_n!(12);
+        check_edge_n!(15);
+        check_edge_n!(16);
+        check_edge_n!(32);
+
+        let mut rng = SmallRng::seed_from_u64(0x5A_A0_D1CA_7E);
```
tcoratger: clippy's `unusual_byte_groupings` fires on this one and on the two seeds further down (lines 755 and 819). Not caught today because clippy never runs against a wasm target, but it would be a hard error under `-D warnings` if a wasm lint step gets added. Either regroup to nibbles (`0x005A_A0D1_CA7E`) or keep the puns and add an `#[allow]`.

goldilocks/src/wasm32_simd128/packing.rs
```diff
@@ -164,13 +164,36 @@ impl PrimeCharacteristicRing for PackedGoldilocksWasmSimd128 {
...
+    #[inline]
+    fn sum_array<const N: usize>(input: &[Self]) -> Self {
+        assert_eq!(N, input.len());
+        const {
+            assert!((N as u32) <= (1 << 31));
+        }
+        match N {
+            0 => Self::ZERO,
+            1 => input[0],
+            2 => input[0] + input[1],
```
tcoratger, line 173: Small gap: the N=2 arm is only covered incidentally. `check_edge_n!` in the new test starts at 3, so adding `check::<2>([Goldilocks::new(u64::MAX); 2], [Goldilocks::ZERO; 2])` would pin the one arm that doesn't go through the delayed-reduction path.

goldilocks/src/poseidon2.rs
```diff
@@ -94,13 +94,28 @@ pub type Poseidon2Goldilocks<const WIDTH: usize> = Poseidon2<
     GOLDILOCKS_S_BOX_DEGREE,
 >;
 
+/// An implementation of the Poseidon2 hash function for the Goldilocks field.
+///
+/// It acts on arrays of the form `[Goldilocks; WIDTH]` and `[PackedGoldilocksWasmSimd128; WIDTH]`,
+/// with round constants pre-broadcast into packed vectors at construction time. See
+/// [`crate::wasm32_simd128::Poseidon2ExternalLayerGoldilocksWasmSimd128`] for details.
+#[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
+pub type Poseidon2Goldilocks<const WIDTH: usize> = Poseidon2<
```
tcoratger, line 110: This changes what `Poseidon2Goldilocks` can act on for wasm32 users. Before, the alias was the generic type, which implements `Permutation<[A; W]>` for any `A: Algebra<Goldilocks> + InjectiveMonomial<7>`, so extension-field and symbolic states worked. Now it's only `[Goldilocks; W]` and `[PackedGoldilocksWasmSimd128; W]`. Downstream code permuting an extension-field state compiles on `wasm32-unknown-unknown` and stops compiling the moment `+simd128` is on.

Same trade AVX2 and AVX512 already made, so this is fine as a decision, but the NEON alias above spells the restriction out in its doc comment and this one doesn't. Worth copying that note here, and mentioning it in the release notes.

goldilocks/examples/wasm_merkle_bench.rs
```diff
@@ -0,0 +1,76 @@
+//! End-to-end-ish benchmark for wasm32+simd128: builds a Merkle tree over a matrix of
```
tcoratger, line 1: I ran this and it works (4453 / 4274 / 4282 ns/leaf at 2^8 / 2^12 / 2^16 leaves), but nothing in CI builds or runs it. The wasm job runs `--bin wasm_smoke` and `--bin wasm_bench`, and its build step is `cargo build --target wasm32-unknown-unknown -p p3-goldilocks`, which doesn't build examples. So the one target whose whole body is behind `cfg(wasm32 + simd128)` is never compiled for wasm anywhere, and it will quietly rot.

On the host jobs it compiles to the stub `eprintln!` but still links the three new dev-deps, so every `cargo test -p p3-goldilocks` on every platform now builds p3-commit, p3-matrix and p3-merkle-tree for nothing. Also note p3-matrix dev-depends on p3-goldilocks, so this adds a dev-dep cycle (cargo allows it, just new here).

Either add a run step to the wasm job (takes about 0.3s), or move it to `src/bin/` alongside the other two so the existing `--bin` pattern picks it up.

tcoratger (commented):
Just made a local review with Claude, looks good! Just left some comments, then feel free to merge :)

goldilocks/src/bin/wasm_bench.rs
```diff
@@ -133,4 +133,380 @@ fn main() {
...
     }
+
+    bench_poseidon2();
+    bench_dot_product();
+    bench_sum_array();
+    bench_quadratic_extension_mul();
+}
+
+/// Times the packed Goldilocks quadratic-extension multiplication and squaring
+/// (`quadratic_mul`/`binomial_square` in `p3_field::extension`, applied to
```
tcoratger: The code imports `binomial_mul`, not `quadratic_mul`.

tcoratger (approved):

PR #1965, opened 2026-07-30
perf(stir): prune Merkle multi-openings for STIR

Nashtare:
STIR proofs are now 1.41x–1.57x smaller than binary FRI at the tested sizes

### DIFF
--- a/stir/src/pcs.rs
+++ b/stir/src/pcs.rs
@@ -95,7 +95,7 @@ impl<Val, Dft, InputMmcs, StirMmcs, Challenge, Challenger> Pcs<Challenge, Challe
 where
     Val: TwoAdicField,
     Dft: TwoAdicSubgroupDft<Val>,
-    InputMmcs: Mmcs<Val, MultiProof: Sync, Error: Sync + Debug>,
+    InputMmcs: Mmcs<Val, Error: Sync + Debug>,
     StirMmcs: Mmcs<Challenge>,
     Challenge: ExtensionField<Val> + TwoAdicField + BasedVectorSpace<Val>,
     Challenger: FieldChallenger<Val>
--- a/stir/src/pcs.rs
+++ b/stir/src/pcs.rs
@@ -27,7 +27,7 @@ use core::marker::PhantomData;
 
 use itertools::{Itertools, izip};
 use p3_challenger::{CanObserve, CanSampleUniformBits, FieldChallenger, GrindingChallenger};
-use p3_commit::{BatchOpening, BatchOpeningRef, Mmcs, OpenedValues, Pcs};
+use p3_commit::{Mmcs, OpenedValues, Pcs};
 use p3_dft::TwoAdicSubgroupDft;
 use p3_field::coset::TwoAdicMultiplicativeCoset;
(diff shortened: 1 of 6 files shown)


stir/tests/stir.rs
tcoratger, line 1: I think that here we could add some negative tests to improve coverage. For example the followings:
1. The fact of permuting two entries of `row_evals` keeps the length checks happy. Only the positional binding inside `verify_batch_pruned` rejects it. If a future refactor ever sorted `row_evals` or `indices` independently, this becomes very important in my opinion.

2. Digest count manipulation as well, like the following, asserting the right error, not just the fact that there is an error

```rust
#[test]
fn test_tampered_pruned_sibling_rejected() {
    let sibs = &mut proof.round_proofs[0].query_openings.opening_proof.sibling_hashes;
    sibs[0][0] += F::ONE;
    assert!(verify_stir::<...>(&config, &proof, &mut v_challenger).is_err());
}
```

and

```rust
#[test]
fn test_dropped_pruned_sibling_rejected() {
    proof.round_proofs[0].query_openings.opening_proof.sibling_hashes.pop();
    assert!(verify_stir::<...>(&config, &proof, &mut v_challenger).is_err());
}
```

3. The Option shape stuff should probably be tested as well like

```rust
// present -> None: must not be silently skipped.
for (_, io) in proof.iter_mut() { io[0] = None; }
assert!(matches!(res, Err(StirError::InvalidProofShape)));
```

and

```rust
// per-commit opened_values truncated.
for (_, io) in proof.iter_mut() { io[0].as_mut().unwrap().opened_values.pop(); }
assert!(matches!(res, Err(StirError::InvalidProofShape)));
```

stir/src/pcs.rs
```diff
@@ -73,7 +95,7 @@ impl<Val, Dft, InputMmcs, StirMmcs, Challenge, Challenger> Pcs<Challenge, Challe
 where
     Val: TwoAdicField,
     Dft: TwoAdicSubgroupDft<Val>,
-    InputMmcs: Mmcs<Val, Proof: Sync, Error: Sync + Debug>,
+    InputMmcs: Mmcs<Val, MultiProof: Sync, Error: Sync + Debug>,
```
tcoratger: Not sure we still need Sync here no? I remember that previously for the other stuffs we were able to remove no?

stir/src/verifier.rs
```diff
@@ -270,41 +263,63 @@ where
...
-                    source,
-                })?;
+        // Step 4b: one shared, pruned Merkle multi-opening proof authenticates every query
+        // row against the current commitment at once.
+        let opened_values: Vec<Vec<&[EF]>> = rp
+            .query_openings
+            .row_evals
+            .iter()
+            .map(|row| alloc::vec![row.as_slice()])
+            .collect();
```
tcoratger: It looks like STIR commits only 1 matrix, so every inner Vec has length 1, not sure then we need the complexity of `Vec<Vec<&[EF]>>` in terms of heap allocations

tcoratger (approved):
Minor stuffs, then ready to merge :)

PR #1969, opened 2026-07-31
ci: parallelize test legs and speed up tool installs

Nashtare:
## Description

CI main run currently takes 20~23 min to execute.

Split each `Build and Test` leg's sequential default/`parallel`-feature steps into separate matrix jobs so they run concurrently instead of back-to-back, and switch to `cargo-nextest` so tests from all binaries share one scheduling pool instead of running one binary at a time — several single, long-running tests (`sumcheck`, `multi-stark`, `whir`, `stir`) were otherwise leaving cores idle while blocking the rest of the suite. Also install `cargo-sort/taplo` via prebuilt binaries instead of compiling them from source on every run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,23 +27,50 @@ jobs:
     strategy:
       matrix:
         include:
+          # Default and `parallel`-feature runs are split into separate legs
+          # so they execute as concurrent jobs instead of sequential steps.
           - os: ubuntu-latest
+            parallel: false
+          - os: ubuntu-latest
+            parallel: true
           - os: ubuntu-latest
             features: "+avx2"
+            parallel: false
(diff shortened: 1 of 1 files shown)


Nashtare:
@tcoratger CI went from 20~23min to 9min 🎉

tcoratger (approved):
Amazing!!

tcoratger:
@Nashtare Looks like there is a problem with some instances, right? Stuffs look stuck since I enabled auto-merge about twenty minutes ago, don't they?

Nashtare:
Yes that's because the jobs got renamed so the old jobs that are required can't complete. I'm in the plane right now so let's leave this one until I land, so it doesn't block other PRs and I'll update the required jobs then!

PR #1975, opened 2026-08-04
refactor(stir): cleanup crate and revamp prover/verifier

Nashtare:
- remove deadcode
- reduce public API 
- remove code duplication


tcoratger (approved):

tcoratger:
LGTM (we just need to fix the conflicts)

PR #1977, opened 2026-08-04
perf(p3-field): vectorize packed field

Nashtare:
## Motivation

  Reconsiders #1802 (closed unmerged): quotient evaluation folds constraints over a single packed vector per row group, which can't hide NEON's packed-multiply latency (~10 cycles at ~1.25 cycles/vector throughput → needs ~8
  independent chains to saturate). #1802 fixed this by evaluating `N=2` packed vectors in lockstep, but was closed because forcing `N=1` on x86 to make it a no-op still regressed — every target paid the wrapper's
  abstraction cost even when it should have been inert.

  ## Changes

  - `p3-field`: adds `Vectorized<F,N>`/`VectorizedExt<F,EF,N>`, unchanged from #1802 — `N` packed vectors bundled behind the same `PackedValue`/`Algebra` interfaces as an ordinary packed value.
  - `p3-uni-stark`: `quotient_values` is split into `quotient_values_plain` (byte-identical to pre-#1802 code) and `quotient_values_vectorized` (`N=2`, `aarch64`+`neon` only). Non-NEON targets compile only the plain path —
  literally the same monomorphization as before this PR, not `Vectorized<F,1>` — so the x86 regression that sank #1802 is structurally impossible here, not just unlikely.
  - `p3-field`: adds `Field::BENEFITS_FROM_LOCKSTEP_EVALUATION` (default `true`), a compile-time-foldable per-field opt-out. `quotient_values` picks plain vs. vectorized via `if Val::<SC>::BENEFITS_FROM_LOCKSTEP_EVALUATION`,
  resolved per monomorphization.
  - `p3-mersenne-31`: overrides the const to `false`. Mersenne31 regresses under the vectorized path on current main, unlike when #1802 was authored — #1894 (merged after #1802 closed) gave `PackedMersenne31Neon` a
  deferred-reduction `dot_product`, closing the extension-field-multiply latency gap the lockstep trick was hiding for that field specifically.
  - All AIRs implemented generically over `AirBuilder` (the standard pattern) satisfy both folder bounds automatically — no downstream AIR changes needed.

  ## Benchmarks

  M4 pro, `prove_prime_field_31`/`prove_goldilocks_poseidon2` examples, isolated `quotient_values` span, `log_trace_length=18` (Goldilocks: fixed-size Keccak example), round-robin/interleaved trials:

  | Field | PCS / AIR | Δ quotient |
  |---|---|---|
  | BabyBear | twoadic, Poseidon2 | −3.5% |
  | KoalaBear | twoadic, Poseidon2 | −15% |
  | Goldilocks | twoadic, Keccak | −11% |
  | Mersenne31 | circle, Poseidon2 | excluded via `BENEFITS_FROM_LOCKSTEP_EVALUATION = false` (measured +10% regression if forced on) |

No impact on x86 (tested on AWS `c8a.8xlarge`).
  ## Test plan

  - [x] `cargo check`/`clippy --all-targets -D warnings` clean on `aarch64`+`neon` (native) and `x86_64` (cross-compiled)
  - [x] Full `p3-uni-stark`/`p3-examples`/`p3-mersenne-31` test suite green (two-adic, circle, ZK, preprocessed, periodic columns)
  - [x] Confirmed at runtime (temporary trace instrumentation, removed) that Mersenne31 dispatches to the plain path

### DIFF
--- a/field/src/field.rs
+++ b/field/src/field.rs
@@ -1005,12 +1005,14 @@ pub trait Field:
     /// independent dependency chains and hide packed-multiplication latency) is expected
     /// to help throughput for this field.
     ///
-    /// Defaults to `true`. Fields should override this to `false` when their packed
-    /// multiplication is not the bottleneck a lockstep evaluation strategy would address —
-    /// for instance because AIRs over this field already interleave enough independent
-    /// per-row work on their own, or because the field's packed multiply is already
-    /// short-latency relative to other pipeline costs.
-    const BENEFITS_FROM_LOCKSTEP_EVALUATION: bool = true;
+    /// Only [`p3_uni_stark::quotient_values`](https://docs.rs/p3-uni-stark)'s `aarch64`
+    /// (`neon`)-gated path reads this constant; on every other target it has no effect,
+    /// so leaving it at the default is always safe there.
+    ///
+    /// Defaults to `false`, so fields fail safe into the plain (non-lockstep) path unless
+    /// explicitly measured to benefit. Override to `true` only once benchmarks confirm the
+    /// field's packed multiplication is latency-bound enough for lockstep evaluation to help.
+    const BENEFITS_FROM_LOCKSTEP_EVALUATION: bool = false;
 
     /// Check if the given field element is equal to the unique additive identity (ZERO).
     #[must_use]
--- a/field/src/field.rs
+++ b/field/src/field.rs
@@ -1001,6 +1001,17 @@ pub trait Field:
     /// A generator of this field's multiplicative group.
     const GENERATOR: Self;
 
+    /// Whether evaluating multiple packed vectors of this field in lockstep (to overlap
+    /// independent dependency chains and hide packed-multiplication latency) is expected
+    /// to help throughput for this field.
+    ///
+    /// Defaults to `true`. Fields should override this to `false` when their packed
+    /// multiplication is not the bottleneck a lockstep evaluation strategy would address —
+    /// for instance because AIRs over this field already interleave enough independent
+    /// per-row work on their own, or because the field's packed multiply is already
+    /// short-latency relative to other pipeline costs.
+    const BENEFITS_FROM_LOCKSTEP_EVALUATION: bool = true;
+
     /// Check if the given field element is equal to the unique additive identity (ZERO).
     #[must_use]
     #[inline]
--- a/field/src/vectorized.rs
+++ b/field/src/vectorized.rs
@@ -1,0 +1,490 @@
+//! Lockstep evaluation over multiple packed vectors, trading register pressure
+//! for instruction-level parallelism in latency-bound field arithmetic.
+//!
+//! Inspired by stwo's `Vectorized` type:
+//! <https://github.com/starkware-libs/stwo/blob/cca98119f/crates/stwo/src/prover/backend/simd/very_packed_m31.rs>
+
+use core::array;
+use core::iter::{Product, Sum};
+use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+
+use crate::{
+    Algebra, BasedVectorSpace, ExtensionField, Field, PackedFieldExtension, PackedValue,
+    PrimeCharacteristicRing,
+};
+
+/// `N` packed base-field vectors operated on in lockstep.
+///
+/// A single packed vector often cannot saturate the CPU's multiplier pipes: a
+/// dependency chain of packed multiplications leaves most issue slots idle
+/// (e.g. NEON's modular multiply has ~10 cycles of latency at ~1.25 cycles of
+/// throughput per vector). Widening the *data type* rather than the loop turns
+/// every ring operation into `N` independent instructions back to back, giving
+/// the out-of-order core `N` interleaved dependency chains without changing the
+/// shape of the expression being evaluated.
+///
+/// Lane `i` of the logical vector lives in `self.0[i / W].as_slice()[i % W]`
+/// where `W = F::Packing::WIDTH`, i.e. components hold consecutive blocks of
+/// lanes.
+#[derive(Clone, Copy, Debug)]
+#[repr(transparent)]
+#[must_use]
+pub struct Vectorized<F: Field, const N: usize>(pub [F::Packing; N]);
+
+/// `N` packed extension-field vectors operated on in lockstep.
+///
+/// The extension-field counterpart of [`Vectorized`]: lane `i` corresponds to
+/// lane `i` of a `Vectorized<F, N>` operand, so the two types can be mixed in
+/// the same expression (`VectorizedExt * Vectorized`, etc.).
+#[derive(Clone, Copy, Debug)]
+#[repr(transparent)]
+#[must_use]
+pub struct VectorizedExt<F: Field, EF: ExtensionField<F>, const N: usize>(
+    pub [EF::ExtensionPacking; N],
+);
+
+impl<F: Field, const N: usize> Vectorized<F, N> {
+    /// Map a function over the `N` packed components.
+    #[inline]
+    fn map(self, f: impl FnMut(F::Packing) -> F::Packing) -> Self {
+        Self(self.0.map(f))
+    }
+
+    /// Combine two values component-wise.
+    #[inline]
+    fn zip_with<T: Copy>(
+        self,
+        rhs: &[T; N],
+        mut f: impl FnMut(F::Packing, T) -> F::Packing,
+    ) -> Self {
+        Self(array::from_fn(|i| f(self.0[i], rhs[i])))
+    }
+}
+
+impl<F: Field, EF: ExtensionField<F>, const N: usize> VectorizedExt<F, EF, N> {
+    /// Map a function over the `N` packed components.
+    #[inline]
+    fn map(self, f: impl FnMut(EF::ExtensionPacking) -> EF::ExtensionPacking) -> Self {
+        Self(self.0.map(f))
+    }
+
+    /// Combine two values component-wise.
+    #[inline]
+    fn zip_with<T: Copy>(
+        self,
+        rhs: &[T; N],
+        mut f: impl FnMut(EF::ExtensionPacking, T) -> EF::ExtensionPacking,
+    ) -> Self {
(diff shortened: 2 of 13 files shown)


field/src/field.rs
```diff
@@ -1001,6 +1001,17 @@ pub trait Field:
...
+    /// Whether evaluating multiple packed vectors of this field in lockstep (to overlap
+    /// independent dependency chains and hide packed-multiplication latency) is expected
+    /// to help throughput for this field.
+    ///
+    /// Defaults to `true`. Fields should override this to `false` when their packed
+    /// multiplication is not the bottleneck a lockstep evaluation strategy would address —
+    /// for instance because AIRs over this field already interleave enough independent
+    /// per-row work on their own, or because the field's packed multiply is already
+    /// short-latency relative to other pipeline costs.
+    const BENEFITS_FROM_LOCKSTEP_EVALUATION: bool = true;
```
tcoratger: Two things about defaulting this to `true`:

1. It opts in every field that exists today or is added later, on a heuristic measured on 4 fields / 2 AIRs / one core (M4 Pro). Defaulting to `false` and overriding to `true` on BabyBear/KoalaBear/Goldilocks would make unmeasured combinations fail safe instead of silently entering the path.
2. The const is only read inside the `cfg(aarch64+neon)` branch in `p3-uni-stark`, but nothing in the doc says so. A field author on x86 has no way to know their override is inert. Worth a sentence.

mersenne-31/src/mersenne_31.rs
```diff
@@ -365,6 +365,12 @@ impl Field for Mersenne31 {
     // Sage: GF(2^31 - 1).multiplicative_generator()
     const GENERATOR: Self = Self::new(7);
 
+    // Circle-STARK AIRs over this field (e.g. the vectorized Poseidon2 AIR) already
+    // interleave several independent permutations per row, which already supplies enough
+    // independent per-row dependency chains; evaluating additional packed vectors in
+    // lockstep on top of that does not improve throughput here.
+    const BENEFITS_FROM_LOCKSTEP_EVALUATION: bool = false;
```
tcoratger, line 372: This comment gives a different reason than the PR description. Here it's "circle AIRs already interleave permutations per row"; the description says #1894 gave `PackedMersenne31Neon` a deferred-reduction `dot_product` that closed the latency gap. Whichever is the measured cause, only that one should be in the code.

Also, the reason as written is a property of the AIR, not of the field — a narrow serial-chain Mersenne31 AIR would be opted out too, with no way to override.

tcoratger (commented):
I checked the arithmetic: I patched `quotient_values` locally to run both paths and assert bit-equality of the returned `Vec<SC::Challenge>`, and the whole `p3-uni-stark` suite passes (two-adic, circle, ZK, preprocessed, periodic columns), including with Mersenne31's opt-out flipped back to `true` so the circle path exercises the vectorized code. So no correctness concern from me.

Comments below are about the API surface and test coverage.

field/src/vectorized.rs
```diff
@@ -0,0 +1,490 @@
...
+impl<F: Field, EF: ExtensionField<F>, const N: usize> Algebra<Vectorized<F, N>>
+    for VectorizedExt<F, EF, N>
+{
+}
+
+// SAFETY: `Vectorized<F, N>` is `repr(transparent)` over `[F::Packing; N]` and
+// `F::Packing: PackedField` guarantees that `F::Packing` can be cast to/from
+// `[F; F::Packing::WIDTH]` without UB. Hence `Vectorized<F, N>` can be cast
+// to/from `[F; F::Packing::WIDTH * N]`.
+unsafe impl<F: Field, const N: usize> PackedValue for Vectorized<F, N> {
```
tcoratger, line 486: No unit tests anywhere in this file. On x86 the type is compiled but never executed, so a lane-ordering bug in `from_fn` / `as_slice` / `extract` would only ever be caught by an ARM end-to-end proof.

Could you add a small test module? The two that matter:
- round-trip: `from_fn(|i| vals[i]).as_slice() == vals`, `extract(i) == vals[i]`, `from_slice(&vals).as_slice() == vals`
- homomorphism: for each op, `op(a, b).as_slice()[i] == op(a[i], b[i])`, plus the mixed `VectorizedExt`/`Vectorized` ops via `extract`

Worth running with `N = 3` as well as `2` — with `N = 2` and `W = 4` several `i / W` vs `i % W` mistakes alias to the right answer.

One more thing for the SAFETY comment: `as_slice` also relies on `align_of::<F::Packing>() <= align_of::<F>()` (the invariant the `const` assert in `PackedValue::pack_slice` pins). Worth naming since the impl depends on it.

field/src/vectorized.rs
```diff
@@ -0,0 +1,490 @@
...
+    where
+        Fn: FnMut(usize) -> Self::Value,
+    {
+        Self(array::from_fn(|i| {
+            F::Packing::from_fn(|j| f(i * F::Packing::WIDTH + j))
+        }))
+    }
+
+    #[inline]
+    fn from_slice(slice: &[Self::Value]) -> &Self {
```
tcoratger, line 502: Every other `PackedValue::from_slice` in the tree uses a direct cast (see the `[T; WIDTH]` impl in `packed_traits.rs`). And this assert can't fire: `align_of::<Vectorized<F,N>>() == align_of::<F::Packing>() <= align_of::<F>()`, so `align_to` always yields an empty prefix. Suggest matching the convention:

```rust
assert_eq!(slice.len(), Self::WIDTH);
unsafe { &*slice.as_ptr().cast() }
```

field/src/vectorized.rs
```diff
@@ -0,0 +1,490 @@
...
+    #[inline]
+    fn halve(&self) -> Self {
+        self.map(|x| x.halve())
+    }
+
+    #[inline]
+    fn square(&self) -> Self {
+        self.map(|x| x.square())
+    }
+}
```
tcoratger, line 450: `Vectorized` forwards `cube`, `exp_const_u64`, `mul_2exp_u64`, `dot_product`, `sum_array`; `VectorizedExt` forwards only `double`/`halve`/`square`.

`mul_2exp_u64` is the one that costs something: `PackedBinomialExtensionField` overrides it with a per-coordinate shift, but the inherited default here is `self.dup() * Self::TWO.exp_u64(exp)` — a full extension exponentiation plus a multiply, per call. Any AIR using it on an extension expression pays that on the vectorized path only.

uni-stark/src/folder.rs
```diff
@@ -51,6 +59,151 @@ pub struct ProverConstraintFolder<'a, SC: StarkGenericConfig> {
...
+    /// The same scheme as [`ProverConstraintFolder::finalize_constraints`], applied
+    /// per vectorized component: the [`Algebra::mixed_dot_product`] overrides on
+    /// [`Vectorized`]/[`VectorizedExt`] delegate each component to the underlying
+    /// packed type, so the specialized SIMD dot products are still used.
+    #[inline]
+    pub fn finalize_constraints(&self) -> VectorizedChallenge<SC, N> {
+        debug_assert_eq!(self.constraint_index, self.constraint_count);
+
+        let base = &self.base_constraints;
+        let base_coefficients: Vec<VectorizedVal<SC, N>> = self
```
tcoratger: This allocates a `Vec` per row group, in the same loop where `GroupBuffers` exists specifically to avoid per-group allocation. The plain `finalize_constraints` uses `from_basis_coefficients_fn` and allocates nothing.

Fix is small: add `base_coefficients: Vec<Pack<SC>>` to `GroupBuffers` and take `&mut Vec<VectorizedVal<SC, N>>` here, `clear()` + `extend()` instead of `collect()`. I tried it and it compiles clean with tests green — but I couldn't measure a difference above noise at `log_trace_length = 18`, so treat this as consistency rather than a perf claim.

uni-stark/src/prover.rs
```diff
@@ -18,6 +20,58 @@ use crate::{
...
+
+/// Number of packed vectors evaluated in lockstep per constraint expression during
+/// quotient evaluation on NEON, for fields where
+/// [`Field::BENEFITS_FROM_LOCKSTEP_EVALUATION`] holds. A single packed vector forms one
+/// dependency chain per constraint expression, which cannot hide the latency of modular
+/// multiplication (e.g. ~10 cycles at ~1.25 cycles/vector throughput on NEON). Two
+/// lockstep chains let the out-of-order core overlap independent rows; values above 2
+/// increase register pressure with diminishing returns.
+#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
+const QUOTIENT_ILP: usize = 2;
```
tcoratger: `QUOTIENT_ILP` is private but appears in `QuotientAir`'s public supertrait bound. A downstream AIR with a concrete `impl Air<ProverConstraintFolder<'a, SC>>` (rather than generic over `AirBuilder`) now fails to compile on aarch64 and can't name `VectorizedConstraintFolder<'a, SC, QUOTIENT_ILP>` to add the missing impl without hardcoding `2`.

Suggest making it `pub`, and adding a line to the `QuotientAir` doc saying explicitly that concrete (non-generic) `Air` impls need the extra impl on aarch64+neon. It's a target-dependent breaking change — compiles on x86, fails on ARM — so probably worth a CHANGELOG entry too.

uni-stark/src/prover.rs
```diff
@@ -553,3 +651,161 @@ where
         );
     quotient_values
 }
+
+#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
+#[allow(clippy::too_many_arguments)]
+fn quotient_values_vectorized<SC, A, Mat>(
```
tcoratger: This is a verbatim copy of `quotient_values_plain` with `PackedVal<SC>` → `Pack<SC>`. I agree with keeping the paths separate (that's the whole point vs #1802), but the duplication of the selector padding, periodic packing, buffer-reuse dance and extract loop will drift.

A generic `quotient_values_inner<SC, A, Mat, P, Fold>` instantiated twice would keep the property you care about — the plain instantiation at `P = PackedVal<SC>` is still literally the old monomorphization — without two copies of the body.

tcoratger (approved):
LGTM, just left some small comments after the review with the agent :)

PR #1978, opened 2026-08-07
fix(p3-security): grade conjectured security per round and at the instance shape

Nashtare:
Conjectured security was computed from the FRI query phase alone. Two other things also limit it, and neither was counted:

  - **The FRI folding rounds.** Only the query phase was charged. The commit phase existed but was reachable from the proven path only.
  - **The instance itself.** `compute_from_params` ignored the trace length and the AIR, so it graded every proof as if it were a single row with one constraint. The proven path already read both.

  Both omissions push the reported number up, so the fix pushes it down.

  ### Numbers that change

  | case | was | now |
  |---|---|---|
  | `compute_ldt_only`, blowup 2 | 128 | 125 |
  | `compute_ldt_only`, blowup 8 | 128 | 119 |
  | `compute_from_params`, 96-bit field, degree bits 10→28 | 84…66 | 82…64 |

  The two `compute_ldt_only` rows were sitting at the 128-bit collision cap; the folding round now binds below it. In the shape-aware vector the binding term moves from DEEP-ALI to folding.

### DIFF
--- a/security/src/fri.rs
+++ b/security/src/fri.rs
@@ -18,7 +18,6 @@ use alloc::vec;
 use alloc::vec::Vec;
 
 use libm::{log2, pow};
-use p3_util::log2_floor_usize;
 
 use crate::error::ErrorBits;
 use crate::ldt::LowDegreeTest;
@@ -123,13 +122,20 @@ pub fn conjectured_commit_phase_error(
 ///
 /// ε ≤ (folding − 1)·(n + 1) / |F|, applied when at least one fold occurs.
 /// Slightly conservative versus `soundcalc`'s tighter `(γn + 1)` factor.
+///
+/// Returns `None` when `regime.max_log_arity` is `0` (folding factor `1`,
+/// i.e. no fold at all) — such a regime has no commit-phase round, rather
+/// than being indistinguishable from arity 2.
 pub fn commit_phase_error_udr(regime: &FriRegime, shape: &InstanceShape) -> Option<ErrorBits> {
+    let folding_minus_one = regime.folding_factor() - 1.0;
+    if folding_minus_one <= 0.0 {
+        return None;
+    }
     let lde_log = shape.log_trace_length + regime.log_blowup;
-    let num_layers = log2_floor_usize(1usize << lde_log).saturating_sub(regime.log_final_poly_len);
+    let num_layers = lde_log.saturating_sub(regime.log_final_poly_len) / regime.max_log_arity;
     if num_layers == 0 {
         return None;
     }
-    let folding_minus_one = (regime.folding_factor() - 1.0).max(1.0);
     let n = (1u64 << lde_log) as f64;
     let bits = shape.modulus_bits as f64 - log2(folding_minus_one * (n + 1.0))
         + regime.commit_pow_bits as f64;
@@ -145,17 +151,28 @@ pub fn commit_phase_error_udr(regime: &FriRegime, shape: &InstanceShape) -> Opti
 /// We also evaluate the n/q-style bound from [2024/1553] and report the
 /// tighter of the two. Round-by-round soundness is dominated by round 0
 /// (largest `n`), so we use `n = lde_domain_size` for every round.
-pub fn commit_phase_error_ldr_m(regime: &FriRegime, shape: &InstanceShape, m: usize) -> ErrorBits {
+///
+/// Returns `None` when `regime.max_log_arity` is `0` (folding factor `1`,
+/// i.e. no fold at all) — such a regime has no commit-phase round, rather
+/// than being indistinguishable from arity 2.
+pub fn commit_phase_error_ldr_m(
+    regime: &FriRegime,
+    shape: &InstanceShape,
+    m: usize,
+) -> Option<ErrorBits> {
     let rho = pow(2.0, -(regime.log_blowup as f64));
     let sqrt_rho = libm::sqrt(rho);
     let m_shifted = m as f64 + 0.5;
     let pp = gamma_ldr_m(regime.log_blowup, m);
     if pp <= 0.0 {
-        return ErrorBits::from_log2(0.0);
+        return Some(ErrorBits::from_log2(0.0));
+    }
+    let folding_minus_one = regime.folding_factor() - 1.0;
+    if folding_minus_one <= 0.0 {
+        return None;
     }
     let lde_log = shape.log_trace_length + regime.log_blowup;
     let n = (1u64 << lde_log) as f64;
-    let folding_minus_one = (regime.folding_factor() - 1.0).max(1.0);
 
     let num = (2.0 * pow(m_shifted, 5.0) + 3.0 * m_shifted * pp * rho) * n;
     let den = 3.0 * rho * sqrt_rho;
@@ -171,7 +188,9 @@ pub fn commit_phase_error_ldr_m(regime: &FriRegime, shape: &InstanceShape, m: us
         + 0.5 * log2(rho)
         + regime.commit_pow_bits as f64;
 
-    ErrorBits::from_log2(bits_linear.min(bits_n_over_q).max(0.0))
+    Some(ErrorBits::from_log2(
+        bits_linear.min(bits_n_over_q).max(0.0),
+    ))
 }
 
 /// FRI query-phase error: ε ≤ αᵏ, contributing `query_pow − k·log2(α)` bits.
@@ -227,9 +246,9 @@ pub fn proven_error_ldr_m(
     if k + air.max_combo as f64 >= (1.0 - pp) * lde {
         return ErrorBits::from_log2(0.0);
     }
-    let commit = commit_phase_error_ldr_m(regime, shape, m);
     let query = query_phase_error(alpha, regime.num_queries, regime.query_pow_bits);
-    ErrorBits::min(&[commit, query])
+    commit_phase_error_ldr_m(regime, shape, m)
+        .map_or(query, |commit| ErrorBits::min(&[commit, query]))
 }
 
 /// Search `m ∈ [3, min(compute_upper_m, LDR_M_CAP)]` for the value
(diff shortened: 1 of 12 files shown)


uni-stark/src/security.rs
```diff
@@ -179,28 +210,66 @@ impl ConjecturedSecurity {
             commit_pow_bits: 0,
             query_pow_bits: query_proof_of_work_bits,
         };
+        let air = P3AirShape {
```
tcoratger, line 227: The regime just above this hardcodes `max_log_arity: 0`, which `p3-fri` refuses to build:

```rust
// fri/src/config.rs:187
assert!(max_log_arity > 0, "max_log_arity must be at least 1 to guarantee folding progress");
```

That field used to be a genuine don't-care here — the old entry point only called `conjectured_error`, which ignores arity. Now `conjectured_terms` also calls `conjectured_commit_phase_error`, which reads it, and the `.max(1.0)` at `security/src/fri.rs:132` rewrites `folding_factor - 1 = 0` into `1`. So this reports the arity-2 folding bound for a regime that declares no folding.

I ran the public API directly to confirm:

```
arity 2^0 = 125.6781
arity 2^1 = 125.6781      <- collapsed by the clamp
arity 2^3 = 122.8707      <- what new_benchmark_high_arity uses
```

and the two rows in the PR table:

```
--- blowup=2 q=64 pow=20 coll=128 |F|=128     (128 -> 125)
    ldt-query-phase          145.5494
    ldt-commit-phase         125.6781   <- binding
--- blowup=8 q=32 pow=0 coll=128 |F|=128      (128 -> 119)
    ldt-commit-phase         119.9944   <- binding
```

Both are `|F| - log2(1 * (2^log_blowup + 1))`, i.e. the clamp's output. The direction is still safe (for any legal arity and any degree_bits the real commit error is lower), but someone reading `compute_ldt_only(2, 64, 20, 128, 128) == 125` can't derive 125 from the five arguments they passed.

Could you either take `max_log_arity` as a parameter, or say in the doc that the commit round is evaluated at folding arity 2 on a single-row trace? Dropping the commit term from this entry point entirely would also be defensible, since it knows nothing about the instance.

security/src/fri.rs
```diff
@@ -83,6 +99,26 @@ pub fn conjectured_error(regime: &FriRegime, shape: &InstanceShape) -> ErrorBits
...
+/// word, and that count does not depend on the decoding regime. What the
+/// random-words conjecture buys is the removal of the list-size multiplier
+/// `L⁺` that the Johnson-bound analysis pays (see
+/// [`crate::stark::conjectured_security_report`]) — it does not assert that a
+/// bad folding challenge cannot exist. Dropping the round entirely, as an
+/// LDT-only conjectured number does, therefore overstates security for a large
+/// LDE domain over a small field.
+///
+/// Returns `None` when no fold occurs, in which case there is no such round.
+pub fn conjectured_commit_phase_error(
```
tcoratger, line 114: Related to the `compute_ldt_only` comment: `commit_phase_error_udr`'s `let folding_minus_one = (regime.folding_factor() - 1.0).max(1.0);` (line 132, and the same at line 158 in the LDR path) means an arity-1 regime is indistinguishable from arity 2. `.max(0.0)` plus an early `None` when the folding factor is 1 would make a fold-free regime report "no commit round" instead of impersonating arity 2.

While you're in here, line 128:

```rust
let num_layers = log2_floor_usize(1usize << lde_log).saturating_sub(regime.log_final_poly_len);
```

`log2_floor_usize(1usize << lde_log)` is just `lde_log`. The round trip buys nothing and the `usize` shift panics in debug when `lde_log >= 32` on wasm32 (a CI target), reachable at `degree_bits = 28`, `log_blowup = 4` — `n` on the next line already uses `1u64`. The layer count also ignores arity: with folding `2^a` it should be `(lde_log - log_final_poly_len) / a`, otherwise a regime that performs no fold still gets charged a round.

security/src/grinding.rs
```diff
@@ -1,8 +1,95 @@
...
+/// - [`Self::lookup_challenge`] is applied to the LogUp fingerprint term by
+///   [`crate::logup::security_term`].
+///
+/// A protocol grinding at a site this crate does not model builds its own
+/// [`crate::report::SecurityTerm`], applies [`boost`] to it, and passes it
+/// through `extras` — no change to this struct is needed.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize)]
+pub struct GrindingSites {
+    /// Bits ground once before the low-degree test samples query positions.
+    pub query_phase: usize,
```
tcoratger: `query_phase` and `ldt_commit_phase` have no effect on any reported number. Grepping the workspace, the only reads are `uni-stark/src/security.rs:142-143`, inside `fri_regime()` — which copies them straight back out of the struct `grinding()` just built from the same two `self` fields. A no-op round trip.

Measured:

```
no grinding    = 72.1926 (ldt-commit-phase)
32+32 grinding = 72.1926 (ldt-commit-phase)
```

64 bits of declared PoW, zero credit, no warning. The doc says they're "consumed by the `LowDegreeTest` implementation, which already carries them", but that's a convention nothing enforces — there are two sources of truth for the same quantity (`GrindingSites::query_phase` and `FriRegime::query_pow_bits`) and the composite reads only the second.

Under-crediting is the safe direction so this isn't a soundness bug, but in this crate a field that looks like an input and is actually documentation seems worth avoiding. Cleanest fix is to make `GrindingSites` the single source of truth: pass it to `LowDegreeTest::conjectured_terms` / the proven entry points and drop `commit_pow_bits`/`query_pow_bits` from `FriRegime`. If that's too big here, deleting the two inert fields would at least remove the trap.

uni-stark/src/security.rs
```diff
@@ -179,28 +210,66 @@ impl ConjecturedSecurity {
...
+    ///   [`conjectured_security_report`] omits it rather than guessing. An
+    ///   instance that random-linear-combines several committed codewords is
+    ///   graded optimistically here relative to [`ProvenSecurity`].
+    /// - `fri_commit_proof_of_work_bits` is credited, but only to the
+    ///   commit-phase term ([`p3_security::fri::conjectured_commit_phase_error`]),
+    ///   which is one of several the composite takes a minimum over. Grinding
+    ///   there raises the reported level only while that term is the binding
+    ///   one.
+    pub fn compute_from_params(params: &StarkSecurityParams, degree_bits: usize) -> Self {
+        debug_assert!(
```
tcoratger: The condition is right (`log2_ceil(d - 1) <= log_blowup` is exactly `d <= 2^log_blowup + 1`), but three things about the placement:

- it's a `debug_assert`, so it does nothing in release, which is where deployments compute security levels;
- it's on the conjectured path only — `ProvenSecurity::compute_from_proof` takes the same `params` and doesn't check;
- it ignores zk. `get_log_num_quotient_chunks` uses `constraint_degree = (degree_hint + is_zk).max(2)`, so under zk the real bound is `d <= 2^log_blowup`, one tighter. `StarkSecurityParams` carries no `is_zk` so it can't express that.

Would `StarkSecurityParams::new` / `from_air` be a better home? That's where the invariant is established, and it would cover both paths.

security/src/stark.rs
```diff
@@ -208,15 +224,122 @@ pub fn proven_security_report<L: LowDegreeTest>(
...
+/// The ALI and DEEP terms are evaluated at [`list_size_conjectured`] — see
+/// [`conjectured_security_report`] for why the proven path's L⁺ multiplier is
+/// absent here.
+///
+/// Like the scalar proven path, this models no grinding sites beyond those the
+/// low-degree test already carries: a caller that grinds elsewhere boosts the
+/// affected term itself via [`crate::grinding::boost`], or uses
+/// [`conjectured_security_report`], which consults
+/// [`crate::grinding::GrindingSites`] directly.
+pub fn conjectured_security<L: LowDegreeTest>(
```
tcoratger, line 256: This has no callers — both uni-STARK entry points go through `conjectured_security_report`, and the only use is its own test. I see the symmetry argument with `proven_security_regime`, but it takes no `GrindingSites`, so it can't express the `out_of_domain` boost the report path applies. That means the two paths provably agree only at `GrindingSites::NONE`, which is the one case `conjectured_report_matches_scalar_composite` checks.

If it stays, adding `&GrindingSites` would let that equivalence test cover a nonzero site — right now its docstring says it exists to catch "a divergence between the two paths" while excluding the axis on which they diverge.

security/src/report.rs
```diff
@@ -43,6 +63,10 @@ pub enum Regime {
     UniqueDecoding,
     /// List-decoding regime at proximity parameter `m`.
     ListDecoding { m: usize },
+    /// Conjectured (random-words) regime: correlated agreement up to
+    /// list-decoding capacity, at list size 1. See
+    /// [`crate::proximity::list_size_conjectured`].
+    Conjectured,
```
tcoratger, line 70: `Regime` is public, `Serialize`, and not `#[non_exhaustive]`, so this variant breaks any downstream exhaustive `match` and changes the serialized form. Fine pre-1.0, but worth adding `#[non_exhaustive]` now so the next variant isn't a second break.

security/src/logup.rs
```diff
@@ -17,21 +17,33 @@
...
 //! number of denominator factors (interactions summed over all buses) and
 //! `W = max_message_width` is the widest payload. Sampling both challenges
-//! independently and uniformly from the challenge field `EF`, the union bound
-//! gives
+//! independently and uniformly from the challenge field `EF`, the per-variable
+//! union bound puts a numerator root at `N·(W + 1) / |EF|`.
+//!
+//! A root of the numerator is not the only way through. A *vanishing
+//! denominator* degenerates the batched transition identity: with `denom_j = 0`
+//! the constraint collapses to `multiplicity_j · Π_{s≠j} denom_s = 0`, forcing
```
tcoratger: The bound is right, but the consequence described is weaker than the real one. With `denom_j = 0`, `common_denominator = 0`, so in `lookup/src/logup.rs:358` the constraint is `0 * frac_local - numerator == 0` — `frac_local` is multiplied by zero and therefore **completely unconstrained on that row**. The prover writes any value into the fraction column and the accumulator absorbs it.

"Forcing `multiplicity_j = 0`" is the honest prover's problem, not the attack; the free `frac_local` is. Same probability, so `W + 2` is unchanged — just the prose.

security/src/ldt.rs
```diff
@@ -34,4 +38,28 @@ pub trait LowDegreeTest {
...
+    /// distinct rounds worth surfacing — [`crate::fri::FriRegime`] splits the
+    /// query phase from the commit-phase folding rounds.
+    ///
+    /// Each returned term must bound a **distinct round** of the protocol.
+    /// The composite takes the minimum rather than summing because an
+    /// adversary who breaks any single round breaks the LDT, so the attained
+    /// security is that of the weakest round — the round-by-round accounting
+    /// of [2024/1553](https://eprint.iacr.org/2024/1553) §2. Returning two
+    /// bounds on the *same* round would silently discard the tighter one.
+    fn conjectured_terms(&self, shape: &InstanceShape) -> Vec<SecurityTerm> {
```
tcoratger, line 62: This default has no in-tree user and no test: `FriRegime` overrides it, `whir` has no `LowDegreeTest` impl, and nothing else in the workspace implements the trait. The `LDT_LABEL` assertion at `stark.rs:441` is on the proven path (`regime_report`), not this.

A two-line stub LDT in the test module here that leaves `conjectured_terms` at its default, asserting the composite picks it up under `LDT_LABEL`, would stop the first downstream LDT from being the one that discovers it.

tcoratger (commented):
Some comments I found while reviewing, after this is fixed we can merge :) 

I checked the `W+1 -> W+2` change against the actual constraint rather than the doc: `lookup/src/logup.rs:358` enforces `common_denominator * frac_local - numerator == 0`, and the denominator from `lookup/src/challenges.rs` is `alpha + (bus+1)*beta^W - sum_k beta^k * payload_k`, monic degree 1 in `alpha`. So the `N/|EF|` event is right, and the `beta`-degree is exactly `N*W` because the bus offset sits at `beta^W`. The math holds.

For a concrete impact number: `prove_prime_field_31 -f koala-bear -o poseidon-2-permutations -l 18` reports 113 bits on main and 102 on this branch.

Two things I'd want changed before merge, below.

tcoratger (approved):

PR #1982, opened 2026-08-13
fix(fri): call bit_reverse_rows after coset_dft_batch in get_evaluations_on_domain

The-Ashwin-Karthikeyan:
## Bug

This appears to be a simple implementation bug. When the FRI blowup factor is less than the degree of the constraints, the verifier returns a `Verification(OodEvaluationMismatch { index: Some(0) })` error.

## Cause

In `fri/src/two_adic_pcs.rs`, the result of `coset_dft_batch` is in bit-reversed row order. Without applying `bit_reverse_rows()`, the subsequent `to_row_major_matrix()` produces evaluations in the wrong order, causing an out-of-domain evaluation mismatch during verification.

## Fix

Added a `.bit_reverse_rows()` call after `coset_dft_batch()` and before `to_row_major_matrix()` to restore the correct row ordering. 

(The verifier succeeds after this fix :))

### DIFF
--- a/fri/src/two_adic_pcs.rs
+++ b/fri/src/two_adic_pcs.rs
@@ -396,6 +396,7 @@ where
         let result = self
             .dft
             .coset_dft_batch(coeffs, domain.shift())
+            .bit_reverse_rows()
             .to_row_major_matrix();
         let result_width = result.width();
 


Nashtare (approved):

Nashtare (changes requested):
Disregard the previous review, it got triggered automatically.
I'm not sure I understand the PR issue.

> When the FRI blowup factor is less than the degree of the constraints, the verifier returns a Verification(OodEvaluationMismatch { index: Some(0) }) error.

This is to be expected, i.e. you need a sufficiently large blowup factor for your constraint polynomial.

The-Ashwin-Karthikeyan:
Thank you for your quick reply. Sorry for the confusion caused by my framing of the bug.

> This is to be expected, i.e. you need a sufficiently large blowup factor for your constraint polynomial.

Don't we only need a sufficiently large evaluation domain for the quotient polynomial (which then gets committed as chunks)? In `get_evaluations_on_domain`, the case `lde.height() < domain.size()` is addressed in the comments:

>    /// When `g' = g` (i.e. `Val::GENERATOR`) and `K` is a subgroup of `H`, this is a simple
>    /// truncation of the bit-reversed LDE. Otherwise, we recover the polynomial coefficients
>    /// from the committed LDE and re-evaluate on the requested domain.

So, here is my expected behavior from `get_evaluations_on_domain`: It detects that the `lde.height()` is less than the requested evaluation domain size, and re-evaluates the polynomial on the larger domain. Since the `domain.shift() == Val::GENERATOR && lde.height() >= domain.size()` branch returns the natural order, the `result` variable in the other branch should be stored in the bit reversed order so the bit_reverse_rows() call on the return line can return the natural order in the right type. 

Right now, `result` is in natural order and this is what my fix addresses.

*Here is the context for where this bug showed up:* I am evaluating two designs across different configurations. The first has a degree 9 constraint and the second only has degree <= 2 constraints. I observe the error `Verification(OodEvaluationMismatch { index: Some(0) })` when I run the verifier with a certificate I generated with the first design with a blowup factor 2.

Nashtare (approved):
Ah right, forgot we had addressed it in February! LGTM

PR #1984, opened 2026-08-17
feat(p3-security): const-fn fixed-point round budget for recursive verifiers

Nashtare:
Adds a deterministic, bit-exact integer counterpart to this crate's `f64` error-bit arithmetic, for protocols that grade proof soundness inside a recursive verifier (a circuit/VM with no floating-point unit, needing the grade reproducible bit-for-bit with every rounding biased toward understatement):

  - security::fixed — 16-fractional-bit fixed-point `log2`/`bits_per_query` as const fns, cross-checked within 1 ulp against this crate's own `fri::conjectured_error`.
  - security::budget — the fixed-point counterpart to `stark::conjectured_security_report`: the same conjectured-regime rounds (lookup, constraint-composition, out-of-domain, DEEP-quotient batching, FRI folding, FRI query, commitment-collision), composed as const fns.

  Two design choices in budget are proposals, open for discussion:
  
 1. Folding sums a doubled per-round coefficient rather than searching for the worst round (see module doc) — conservative by < 1 bit vs. `fri::commit_phase_error_udr`'s worst-round bound.
 2. DEEP-quotient batching (`AirShape::num_deep_terms: Option<u32>`) is a first-class optional round on the fixed side with no `f64`-side equivalent; could instead go through `stark`'s extras mechanism if that's preferred.

### DIFF
--- a/security/src/budget/mod.rs
+++ b/security/src/budget/mod.rs
@@ -1,0 +1,345 @@
+//! Conjectured round budget for a lifted (multi-AIR) STARK protocol, in fixed point.
+//!
+//! Fiat-Shamir security of the compiled argument is the minimum, over the protocol's rounds, of
+//! that round's soundness error plus the grinding sited immediately before its challenge. This
+//! module bounds each round and composes them, using [`crate::fixed`] throughout so the result is
+//! reproducible bit-for-bit by a recursive verifier.
+//!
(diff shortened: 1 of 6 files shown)


Nashtare:
I'll proceed to an `rc` release once this lands onto `main`, so that downstream users like `0xMiden` can test it before waiting for `v0.7.0` to arrive.

tcoratger (approved):

PR #1988, opened 2026-08-20
perf: fixed-width binary serialization for field elements

Nashtare:
## Description

Postcard varint-encodes near-uniform 31/64-bit field values, averaging ~4.8 bytes where a fixed width suffices. Branch on serializer.is_human_readable() so postcard/binary output packs a fixed 4-byte (Mersenne31/MontyField31) or 8-byte (Goldilocks) little-endian array while JSON keeps its existing plain-integer encoding.

Measured ~17-18% smaller proofs across FRI/STIR/WHIR.


tcoratger (approved):

tcoratger:
I remember I noticed that couple months ago but was stuck in my TODO queue, great to have! The impact is nice!

PR #1989, opened 2026-08-20
perf(stir): allow round 0's folding arity to differ from later rounds

Nashtare:
## Summary

- Add a multi-table FRI vs STIR vs WHIR comparison benchmark (`examples/examples/pcs_comparison.rs`) that commits a mixed-height, shared-width trace batch (`2^{n, n+1, n+3}` rows, one column width) — the shape a real multi-table STARK actually commits — alongside the existing single-table comparison.
- Let STIR's round-0 folding arity (`log_starting_folding_factor`, k0) differ from the steady-state arity used from round 1 onward (`log_folding_factor`, k). Construction 5.2 already allows every round's folding parameter to vary independently; this exposes that generality for round 0 only, since a smaller k0 shrinks every first-round query's fiber (`2^k0` input rows × committed width) without touching the improved-rate schedule the later rounds are priced on.
- `pcs_comparison`'s `--stir-log-starting-fold` defaults to `2` (arity 4) rather than mirroring `--stir-log-fold`, so the win shows up with no extra flags.

## Benchmark

Multi-table run, heights `2^{17,18,20}`, width 32, 100-bit security, rho = 1/2, k0=4/k=16 vs the prior constant-arity k=16 schedule:

| | proof size | verify time | queries per round |
|---|---|---|---|
| before (k0 = k = 16) | 823.36 KiB | baseline | `[86,22,13,9,7]` |
| after (k0 = 4, k = 16) | **441.58 KiB (-46.4%)** | **-27%** | `[86,43,18,11,8]` |

(Prover open time +11%: one extra round's PoW grind, as expected from the extra round the smaller k0 introduces.)

## Test plan

- [x] `cargo test -p p3-stir` (lib + integration suites green)
- [x] Release-mode benchmark run (`cargo run --release --example pcs_comparison`), before/after with the flag defaulted

### DIFF
--- a/stir/src/config.rs
+++ b/stir/src/config.rs
@@ -345,12 +345,13 @@ where
         };
 
         // Disjoint-coset side condition for round `i`. The schedule sets
-        // `shift_{i+1} = shift_i^{k_i} * GEN` (`k_i` round `i`'s own folding factor), so
-        // `shift_{i+1}/shift_i = GEN^{k_i^{i+1}}`. Disjoint cosets `L_i ∩ L_{i+1} = ∅`
-        // requires that ratio ∉ H_i, i.e. `(GEN^{k_i^{i+1}})^{|H_i|} = GEN^{2^{N_i}} ≠ 1`
-        // where `N_i = (Σ_{j≤i} log_folding_factor_j) + log_domain_i`, the cumulative
-        // folding-log through and including round `i`. Holds for any field whose
-        // multiplicative order has nontrivial odd part (BabyBear, KoalaBear,
+        // `shift_{i+1} = shift_i^{k_i} * GEN` each round (`k_i` = that round's own folding
+        // factor), so `shift_i = GEN^{c_i}` for the nested recursion `c_{i+1} = c_i * k_i + 1`
+        // — not a plain power of the summed `k_j`'s. Disjoint cosets `L_i ∩ L_{i+1} = ∅`
+        // require `GEN` to avoid the size-`2^{log_domain_i}` subgroup reached at round `i`;
+        // we check the simpler `GEN^{2^{N_i}} ≠ 1` where `N_i = (Σ_{j≤i} log_folding_factor_j)
+        // + log_domain_i` is the cumulative folding-log through round `i`. Holds for any
+        // field whose multiplicative order has nontrivial odd part (BabyBear, KoalaBear,
         // Goldilocks, …); the assertion catches pathological fields.
         let assert_disjoint_cosets =
             |round_index: usize, log_domain_i: usize, cumulative_log_folding: usize| {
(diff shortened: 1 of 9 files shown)


stir/src/config.rs
```diff
@@ -863,6 +919,7 @@ mod tests {
                     StirParameters {
                         log_blowup,
                         log_folding_factor: log_fold,
+                        log_starting_folding_factor: log_fold,
```
tcoratger: This is the one test that actually re-checks the security-bit math (not just that an honest proof verifies), but none of the `cases` above use `log_starting_folding_factor != log_folding_factor`. Might be worth adding a k0 != k case here, since a round-trip test wouldn't catch a mixed-up k0/k in the eta/PoW formulas, only this one would. Heads up if you do: `total_folds = log_deg / log_fold` a bit further down still uses the old constant-arity formula, so it'd need to become `1 + (log_deg - k0) / k` too.

stir/src/config.rs
```diff
@@ -321,38 +345,48 @@ where
...
-            let n_i = (round_index + 1) * log_folding_factor + log_domain_i;
-            assert!(
-                F::GENERATOR.exp_power_of_2(n_i) != F::ONE,
-                "STIR round {round_index}: disjoint-coset schedule requires \
-                 Field::GENERATOR^(2^{n_i}) ≠ 1 (i.e. GEN^(k^{}) ∉ subgroup of size 2^{log_domain_i}).",
-                round_index + 1,
-            );
-        };
+        // `shift_{i+1} = shift_i^{k_i} * GEN` (`k_i` round `i`'s own folding factor), so
+        // `shift_{i+1}/shift_i = GEN^{k_i^{i+1}}`. Disjoint cosets `L_i ∩ L_{i+1} = ∅`
```
tcoratger: Minor nit: this identity isn't quite right, `shift_i = GEN^{c_i}` with `c_{i+1} = c_i*k_i + 1` (a nested recursion), not a plain sum of the k_j's, so `shift_{i+1}/shift_i` isn't literally `GEN^{cumulative}`. Doesn't actually matter since the assert is vacuously true for any field with an odd factor in its group order (BabyBear/KoalaBear/Goldilocks all qualify), just didn't want to leave a slightly-off derivation in the comment for future us.

tcoratger (commented):
Went through this one carefully since it touches STIR's soundness params. Checked the k0/k split against Construction 5.2 in the STIR paper (eprint 2024/390) directly, and it does define k0,...,kM as independent per-round params, so this is legit, not a hack. Traced it through config derivation, prover, verifier, and the PCS fiber-grouping, all consistent, and ran the new tests locally, all green. Two small notes inline, nothing blocking, good to go once you've had a look.

tcoratger (approved):
Looks good, I didn't find anything, just 2 minor stuffs with the agent.

Nashtare:
@tcoratger Thanks for the review! I've addressed the comments. I'll probably wait until all STIR PRs of the stack are reviewed to land everything altogether

PR #1990, opened 2026-08-20
perf(stir): commit height classes on a shared domain, merge via Combine (§7)

Nashtare:
Stacked on #1989 (round-0 folding arity) — diff below is only this PR's two commits.

## Summary

1. **`fix(stir)`: correctly account Combine's soundness against the schedule's own eta.**
   The batch-degree-correction step (§4.5 `Combine`) now derives its minimum-`eta`
   requirement per security assumption — Conjecture 5.6's `m`-based formula under
   `CapacityBound`, Theorem 4.1's `ℓ`-based formula under `JohnsonBound` — and folds it
   into round 0's own `eta` via `StirConfig::new_with_combine` (`.max()`'d against the
   schedule's actual value), so `delta_combine` can never diverge from whatever radius
   round 0 actually settled on. Counted as an extra term in the union-bound buffer.

2. **`perf(stir)`: commit height classes on a shared domain, merge via Combine (§7,
   Construction 7.2).** Every matrix passed to one `commit()` call is now extended onto a
   single shared LDE domain (sized to the tallest matrix) and committed together in one
   Merkle tree, instead of one independent STIR instance per native matrix height. Within
   a shared-domain bucket, per-height-class reduced-openings are merged into a single
   codeword via batch degree correction (Combine, Definition 4.11/4.12) at the tallest
   class's degree and full proximity radius before STIR runs — so shorter classes stop
   paying for their own STIR instance and query-count floor.

## Why the soundness fix landed first

The initial `Combine` implementation checked its error term against a fixed, assumed
distance (a Johnson-bound ceiling) independent of round 0's real `eta`, and always used
the provable (not conjectured) bound regardless of the configured security assumption.
That under-priced the step and panicked at the benchmark's actual target scale
("the Combine step ... delivers only 67.55 bits of security for a 100-bit target").
Reusing the exact `eta` STIR's own round 0 lands on closes that gap for a negligible
query-count cost — both quantities sit far below the Johnson/capacity ceiling at these
parameters.

## Benchmark

Multi-table run, `--log-message-size 20`, heights `2^{17,18,20}`, 100-bit security, vs.
the round-0-folding-arity baseline in #1989:

| width | proof size | verify time |
|---|---|---|
| 32 | 441.58 → **265.24 KiB (-39.9%)** | 8003 → **4304 µs (-46%)** |
| 2  | 292.67 → **115.55 KiB (-60.5%)** | 6572 → **2940 µs (-55%)** |

Both widths are clean wins at the benchmark's actual target scale. (A mid-investigation
measurement at a smaller scale, taken while the soundness bug above was still blocking
the real-scale run, showed a regression at width 32 — that turned out to be a small-scale
artifact, not representative.)

## Test plan

- [x] `cargo test -p p3-stir` (29 lib + 56 integration tests, all green)
- [x] `cargo check --workspace --all-targets`
- [x] Release-mode benchmark at target scale (`--log-message-size 20 --multi-table-width {32,2}`), before/after via `git stash`

### DIFF
--- a/stir/src/config.rs
+++ b/stir/src/config.rs
@@ -14,8 +14,8 @@ use crate::soundness::{StirSoundness, combine_min_log_eta};
 /// Extra requirement round 0's `eta` must additionally satisfy so that the
 /// batch-degree-correction `Combine` step (§4.5) reaches the target security when merging
 /// `num_classes` height classes sharing this config's initial domain before the round-0
-/// fold. `ell` is Lemma 4.13's inflated error argument, consulted only under
-/// `SecurityAssumption::JohnsonBound`.
+/// fold. `ell` is Lemma 4.13's inflated error argument, consulted by both
+/// `SecurityAssumption` regimes' error terms.
 #[derive(Clone, Copy, Debug)]
 struct CombineRequirement {
     num_classes: usize,
@@ -228,8 +228,8 @@ where
     /// correlated agreement" guarantee actually apply to STIR's own decoding.
     ///
     /// `ell` is Lemma 4.13's error argument `num_classes·(d* + 1) − Σᵢ dᵢ` (dᵢ each
-    /// class's own, non-quotiented degree bound); only consulted under
-    /// `SecurityAssumption::JohnsonBound` — pass `0` under `CapacityBound`.
+    /// class's own, non-quotiented degree bound), consulted by both
+    /// `SecurityAssumption` regimes' error terms.
     ///
     /// # Panics
     ///
@@ -439,7 +439,6 @@ where
                 field_size_bits,
                 log_inv_rate,
                 log_degree,
-                c.num_classes,
                 c.ell,
                 buffered_security_level,
             );
--- a/stir/src/config.rs
+++ b/stir/src/config.rs
@@ -9,7 +9,7 @@ use p3_commit::Mmcs;
 use p3_field::{ExtensionField, TwoAdicField};
 
 use crate::SecurityAssumption;
-use crate::soundness::{StirSoundness, combine_min_log_eta};
+use crate::soundness::StirSoundness;
 
 /// Extra requirement round 0's `eta` must additionally satisfy so that the
 /// batch-degree-correction `Combine` step (§4.5) reaches the target security when merging
@@ -25,6 +25,31 @@ struct CombineRequirement {
 /// User-facing STIR protocol parameters.
 ///
 /// These are the inputs from which the full [`StirConfig`] is derived.
+///
+/// # Combine feasibility envelope
+///
+/// [`StirConfig::new`] is feasible over a wide range of these parameters, but
+/// [`StirConfig::new_with_combine`] — the §7 Construction 7.2 path the PCS takes when one
+/// commitment holds matrices of more than one native height — is not. Merging classes of total
+/// multiplicity `ell` into a degree-`d* = 2^log_d_star` codeword costs roughly `2·log_d_star`
+/// bits of challenge field, because the `dᵢ` are distinct powers of two and so every degree gap
+/// is within a factor of two of `d*` however tight the height spread is.
+///
+/// Solving the `CapacityBound` requirement against the `eta <= rho/2` side condition gives a
+/// closed form for when Combine fits:
+///
+/// ```text
+/// EF::bits() >= security_level + union_bound_buffer + 3*log_blowup + 1
+///               + log2(ell - 1) + log_d_star
+/// ```
+///
+/// At `d* = 2^20`, `log_blowup = 1` and 100-bit security that is ~151 bits, so a 155-bit
+/// quintic extension fits with a few bits to spare and narrower challenge fields do not.
+/// `JohnsonBound` does not fit at production scale at all: the largest permitted
+/// `eta = sqrt(rho)/20` pins BCSS25's multiplicity at `m = 10`, which retains only ~95 bits at
+/// the same shape.
+/// Outside the envelope, derivation reports the shortfall rather than silently weakening the
+/// parameters; callers that need wider height spreads should commit the outliers separately.
 #[derive(Clone, Debug)]
 pub struct StirParameters<M> {
     /// Log₂ of the inverse rate of the initial Reed-Solomon code.
@@ -215,7 +240,7 @@ where
     /// Panics if `log_folding_factor < 2`, if `log_starting_folding_factor < 2`, or if
     /// `log_starting_folding_factor > log_starting_degree`.
     pub fn new(log_starting_degree: usize, params: StirParameters<M>) -> Self {
-        Self::new_impl(log_starting_degree, params, None)
+        Self::new_with_optional_combine(log_starting_degree, params, None)
     }
 
     /// Like [`Self::new`], but additionally inflates round 0's `eta` (and, transitively,
@@ -233,21 +258,34 @@ where
     ///
     /// # Panics
     ///
-    /// Same conditions as [`Self::new`].
+    /// Every condition of [`Self::new`], plus: if `num_classes < 2` (a single class needs no
+    /// `Combine`, so [`Self::new`] is the correct entry point), if `ell < num_classes as u64`
+    /// (every class contributes at least one to the multiplicity), or if the parameters fall
+    /// outside the feasibility envelope documented on [`StirParameters`].
     pub fn new_with_combine(
         log_starting_degree: usize,
         params: StirParameters<M>,
         num_classes: usize,
         ell: u64,
     ) -> Self {
-        Self::new_impl(
+        assert!(
+            num_classes >= 2,
+            "new_with_combine requires at least 2 height classes (got {num_classes}); a single \
+             class needs no Combine, so use StirConfig::new"
+        );
+        assert!(
+            ell >= num_classes as u64,
+            "Combine multiplicity ell = {ell} is below the class count {num_classes}; every \
+             class contributes at least one to ell = Σᵢ (gapᵢ + 1)"
+        );
+        Self::new_with_optional_combine(
             log_starting_degree,
             params,
             Some(CombineRequirement { num_classes, ell }),
         )
     }
 
-    fn new_impl(
+    fn new_with_optional_combine(
         log_starting_degree: usize,
         params: StirParameters<M>,
         combine: Option<CombineRequirement>,
@@ -434,15 +472,38 @@ where
         // 0's own starting degree and rate, matching what Combine merges at (immediately
         // before the round-0 fold).
         if let Some(c) = combine.filter(|c| c.num_classes >= 2) {
-            let log_combine_eta = combine_min_log_eta(
-                params.soundness_type,
+            let combine_eta = params.soundness_type.stir_combine_eta(
                 field_size_bits,
                 log_inv_rate,
                 log_degree,
                 c.ell,
                 buffered_security_level,
             );
-            final_eta = final_eta.max(libm::pow(2., log_combine_eta));
+            // Reported here rather than left to `validate_eta` below: the schedule's own
+            // `stir_initial_eta` is always within the ceiling, so a round-0 violation is
+            // always Combine's, and the caller needs the envelope to act on it.
+            assert!(
+                params
+                    .soundness_type
+                    .stir_eta_is_valid(log_inv_rate, combine_eta),
+                "Combine over {} height classes (multiplicity ell = {}) at degree 2^{log_degree} \
+                 needs eta = {combine_eta}, above the {:?} side-condition ceiling {}. \
+                 CapacityBound admits Combine only when EF::bits() >= security_level + \
+                 union_bound_buffer + 3*log_blowup + 1 + log2(ell - 1) + log_d_star \
+                 (= {} + {union_bound_buffer} + {} + 1 + {:.2} + {log_degree} = {:.2}; \
+                 EF::bits() = {field_size_bits}). Widen the challenge field, lower \
+                 security_level, raise log_blowup, or narrow the committed height spread.",
+                c.num_classes,
+                c.ell,
+                params.soundness_type,
+                params.soundness_type.stir_eta_upper_bound(log_inv_rate),
+                security_level,
+                3 * log_inv_rate,
+                libm::log2(c.ell.saturating_sub(1).max(1) as f64),
+                (security_level + union_bound_buffer + 3 * log_inv_rate + 1 + log_degree) as f64
+                    + libm::log2(c.ell.saturating_sub(1).max(1) as f64),
+            );
+            final_eta = final_eta.max(combine_eta);
         }
         validate_eta(0, log_inv_rate, final_eta);
 
--- a/stir/src/config.rs
+++ b/stir/src/config.rs
@@ -9,7 +9,18 @@ use p3_commit::Mmcs;
 use p3_field::{ExtensionField, TwoAdicField};
 
 use crate::SecurityAssumption;
-use crate::soundness::StirSoundness;
+use crate::soundness::{StirSoundness, combine_min_log_eta};
+
+/// Extra requirement round 0's `eta` must additionally satisfy so that the
+/// batch-degree-correction `Combine` step (§4.5) reaches the target security when merging
+/// `num_classes` height classes sharing this config's initial domain before the round-0
+/// fold. `ell` is Lemma 4.13's inflated error argument, consulted only under
+/// `SecurityAssumption::JohnsonBound`.
+#[derive(Clone, Copy, Debug)]
+struct CombineRequirement {
+    num_classes: usize,
+    ell: u64,
+}
 
 /// User-facing STIR protocol parameters.
 ///
@@ -204,6 +215,43 @@ where
     /// Panics if `log_folding_factor < 2`, if `log_starting_folding_factor < 2`, or if
     /// `log_starting_folding_factor > log_starting_degree`.
     pub fn new(log_starting_degree: usize, params: StirParameters<M>) -> Self {
+        Self::new_impl(log_starting_degree, params, None)
+    }
+
+    /// Like [`Self::new`], but additionally inflates round 0's `eta` (and, transitively,
+    /// the whole schedule that follows from it) so the batch-degree-correction `Combine`
+    /// step (§4.5) also reaches `params.security_level` when merging `num_classes` height
+    /// classes that share this config's initial domain, immediately before the round-0
+    /// fold. Sharing `eta` between STIR's own round-0 acceptance radius and `Combine`'s
+    /// error term (rather than checking `Combine` against a separately assumed radius)
+    /// is what makes `Combine`'s "if the merged codeword passes, each class has
+    /// correlated agreement" guarantee actually apply to STIR's own decoding.
+    ///
+    /// `ell` is Lemma 4.13's error argument `num_classes·(d* + 1) − Σᵢ dᵢ` (dᵢ each
+    /// class's own, non-quotiented degree bound); only consulted under
+    /// `SecurityAssumption::JohnsonBound` — pass `0` under `CapacityBound`.
+    ///
+    /// # Panics
+    ///
+    /// Same conditions as [`Self::new`].
+    pub fn new_with_combine(
+        log_starting_degree: usize,
+        params: StirParameters<M>,
+        num_classes: usize,
+        ell: u64,
+    ) -> Self {
+        Self::new_impl(
+            log_starting_degree,
+            params,
+            Some(CombineRequirement { num_classes, ell }),
+        )
+    }
+
+    fn new_impl(
+        log_starting_degree: usize,
+        params: StirParameters<M>,
+        combine: Option<CombineRequirement>,
+    ) -> Self {
         assert!(
             params.log_folding_factor >= 2,
             "the paper-backed STIR parameter schedule requires log_folding_factor >= 2 (k >= 4)"
@@ -276,12 +324,16 @@ where
         // `2^{-security_level}`. Exact term count: each of the `total_folds - 1`
         // intermediate rounds has six independent terms (query tier: query
         // failure, OOD, random-combination, shake-check; folding tier: proximity-gaps,
-        // sumcheck); the final stage has three (folding tier + final query failure).
-        // The buffer applies to every per-event term. OOD and shake-check must reach the
-        // buffered target algebraically because the query-phase grind does not protect them.
+        // sumcheck); the final stage has three (folding tier + final query failure); a
+        // `Combine` bucket adds one more (Theorem 7.1's `ε_com` term, §4.5).
+        // The buffer applies to every per-event term. OOD, shake-check, and Combine must
+        // reach the buffered target algebraically because the query-phase grind does not
+        // protect them.
         const TERMS_PER_INTERMEDIATE_ROUND: usize = 6;
         const FINAL_STAGE_TERMS: usize = 3;
-        let num_alg_terms = TERMS_PER_INTERMEDIATE_ROUND * (total_folds - 1) + FINAL_STAGE_TERMS;
+        let combine_term = usize::from(combine.is_some_and(|c| c.num_classes >= 2));
+        let num_alg_terms =
+            TERMS_PER_INTERMEDIATE_ROUND * (total_folds - 1) + FINAL_STAGE_TERMS + combine_term;
         let union_bound_buffer = libm::ceil(libm::log2(num_alg_terms as f64)) as usize;
         let buffered_security_level = security_level + union_bound_buffer;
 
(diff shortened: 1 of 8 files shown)


stir/src/config.rs
```diff
@@ -376,6 +428,22 @@ where
...
+        if let Some(c) = combine.filter(|c| c.num_classes >= 2) {
+            let log_combine_eta = combine_min_log_eta(
+                params.soundness_type,
+                field_size_bits,
+                log_inv_rate,
+                log_degree,
+                c.ell,
+                buffered_security_level,
+            );
+            final_eta = final_eta.max(libm::pow(2., log_combine_eta));
```
tcoratger: This is the line that makes the feature unusable outside the benchmark's exact configuration, so flagging it as the one blocking item.

The `.max()` here is the right *design* -- sharing `eta` between STIR's round-0 radius and Combine's error term is exactly what makes Combine's correlated-agreement guarantee cover STIR's own decoding, and I like that it is one variable rather than two to keep in sync. The problem is the magnitude. `ell = sum_i (gap_i + 1)` and the `d_i` are distinct powers of two, so `gap_i` is always within a factor of two of `d*`; there is no small-spread escape hatch. `log2(ell - 1) + log_d_star` is therefore always about `2*log_d_star`, ~41 bits at `d* = 2^20`. Against the `eta <= rho/2` ceiling that `validate_eta` enforces two lines down:

```
field_bits >= security_level + union_buffer + 3*log_inv_rate + 1 + log2(ell - 1) + log_d_star
```

= 150.7 for the bench shape, so the 155-bit quintic has 4.3 bits of headroom and everything else panics (table in the review body). Bumping `security_level` from 100 to 112 is enough to tip it over.

Three separate things I would want here:

1. **`verify` must not panic.** `Pcs::verify` builds its `StirConfig` from `commitments_with_opening_points`, so a claim shape that pushes `eta` past the ceiling aborts the process instead of returning `Err(InvalidProofShape)`. A `try_new_with_combine(...) -> Result<Self, _>` used from the verifier side would fix that; `open` can keep asserting, that is genuine prover misconfiguration.
2. **Name the envelope.** If `>= ~150`-bit challenge fields at 100-bit security and `d* <= 2^20` is the intended support envelope, worth saying so in the `StirParameters` docs and making the assert message reference the closed form above -- `eta = 4.73 violates ... 0.025` gives a caller nothing to act on.
3. **Decide what happens outside it.** Falling back to the base branch's per-height-class instances when `ell` does not fit the budget would keep the win where it is affordable instead of panicking. Right now a system that works fine with uniform heights starts aborting the moment one table shrinks, which is a rough failure mode to debug.

Repro, in this branch:

```rust
// 155-bit field, CapacityBound, security_level = 112, max_pow_bits = 20, log_blowup = 1
let classes = [20usize, 18, 17];
let d = 1u64 << classes[0];
let ell = classes.len() as u64 * (d + 1) - classes.iter().map(|&c| 1u64 << c).sum::<u64>();
let _ok   = StirConfig::<F, EF, M, Ch>::new(20, params.clone());               // 295 queries
let _boom = StirConfig::<F, EF, M, Ch>::new_with_combine(20, params, 3, ell);  // panics
```

Unrelated small thing while we are in this file: `CombineRequirement::num_classes` is only ever consumed as `combine.filter(|c| c.num_classes >= 2)`, so `new_with_combine(_, _, 1, ell)` silently does no Combine accounting at all and no buffer increment, despite the name. Since `ell` carries the real information, an `assert!(num_classes >= 2)` plus `assert!(ell >= num_classes as u64)` at the top would turn that into a loud error -- both call sites already branch on `native_heights.len() >= 2` before calling.

stir/src/soundness.rs
```diff
@@ -483,6 +483,72 @@ impl StirSoundness for SecurityAssumption {
...
+                    - field_size_bits as f64;
+
+            // Lemma 7.3's delta < 1 - rho - 1/|L_0| side condition, restated as a floor on
+            // eta (|L_0| = 2^(log_d_star + log_inv_rate), round 0's domain size).
+            let log_domain_size = (log_d_star + log_inv_rate) as f64;
+            let log_eta_side_condition = 1. - log_domain_size;
+
+            log_eta_conjecture.max(log_eta_side_condition)
+        }
+        SecurityAssumption::JohnsonBound => {
```
tcoratger: The algebra is a correct inversion of BCIKS20 Theorem 4.1's far case (`err* = (l-1)*d*^2 / (|F|*(2*eta)^7)`), so this is not wrong -- but it is the bound `stir_initial_eta` explicitly declines to use, a couple hundred lines up:

> The old BCIKS-form 1/7-power expression is intentionally not retained here: validation uses BCSS25's `O(n/eta^5)` bound, so deriving eta from the same bound avoids rejecting feasible JB configurations at realistic security levels.

So after this PR the file derives `eta` from two different Johnson-regime bounds depending on which term you ask about, with the Combine term on the looser one. That is worth reconciling on its own terms -- anyone diffing this arm against `prox_gaps_error_at_log_eta`'s JB arm will reasonably conclude one of them is a bug.

The `(2*eta)^7` denominator is also unforgiving: it turns the ~41 bits of `log2(ell-1) + 2*log_d_star` into `log2(eta) = +4.96`, i.e. `eta ~ 31` against a ceiling of `0.035`. That is why every JohnsonBound row in the review table panics, including at 100 bits on the 155-bit field where `new()` is happy with 511 queries.

Fair warning that restating it in the file's own BCSS25 form does not rescue feasibility -- I checked, and at `lir = 1`, `field_bits = 155`, `log_d* = 20`, `ell - 1 = 2^20.7`, even the largest permitted `eta = sqrt(rho)/20 = 2^-4.82` gives `m = 10` and only ~95 bits against a 106-bit buffered target. So this is a consistency fix rather than a fix for the blocker. Suggestion: express it through the same shape as `prox_gaps_error_at_log_eta` (`m = max(ceil(sqrt(rho)/(2*eta)), 3)`, exceptional set `~ n*(m+0.5)^5*rho^1.5`, with `num_functions - 1 -> ell - 1`) and solve it with the existing `minimum_eta_for_target` bisection instead of a closed form, then note in the doc that JB + Combine is infeasible at production scale so callers are not surprised by the assert.

stir/src/pcs.rs
```diff
@@ -252,34 +236,56 @@ where
...
-                    min_height,
-                    self.stir.log_starting_folding_factor,
-                );
+                let log_native_height = log2_strict_usize(domain.size());
+                let extra_bits = log_shared_lde_height - log_native_height;
                 let shift = Val::GENERATOR / domain.shift();
                 let lde = self
                     .dft
-                    .coset_lde_batch(evals, self.stir.log_blowup, shift)
+                    .coset_lde_batch(evals, extra_bits, shift)
```
tcoratger, line 289: `extra_bits = log_shared_lde_height - log_native_height` is where the prover cost the benchmark table does not mention comes from: a `2^12` matrix sitting next to a `2^20` one now pays a `2^9` blowup instead of `2^1`, in both DFT work and its share of the Merkle tree (the tree is `2^{h_max+b}` rows deep with every matrix's full width in each leaf, so hashing goes from `sum_i 2^{h_i+b} * width_i` to `2^{h_max+b} * total_width`).

Measured A/B, release, same machine, KoalaBear/155-bit, width 32, security 100, `log_blowup = 1`:

| heights | base commit | this PR commit | base commit+open | this PR commit+open |
|---|---|---|---|---|
| `[20]` | 2.82 s | 2.50 s | -- | -- |
| `[20,18,17]` | 3.48 s | **7.15 s** | 7.10 s | **10.25 s** (1.44x) |
| `[20,10]` | 2.53 s | **4.68 s** | 5.56 s | **7.16 s** (1.29x) |
| `[20,12,12,12,12]` | 2.62 s | **11.45 s** | 5.08 s | **14.30 s** (2.82x) |

Isolated, the four `2^12` tables go from 34 ms to ~8.9 s, and a lone `2^10` beside a `2^20` goes from 2.2 ms to ~2.2 s. `open` is genuinely a bit faster now (one STIR instance instead of three), it just does not cover the commit side.

I do not think this sinks the approach -- 1.44x prover for -40% proof size and -46% verify is a trade plenty of callers would take. But it is the reader's trade to make, so it wants a row in the table plus a wide-spread shape, and ideally a note that it degrades with the spread. Worth noting the same knob fixes the blocker: bucketing matrices into *several* shared domains by bounded height spread caps both this and `ell`.

Tiny thing nearby: the `type Proof` doc a bit above still says "one entry per distinct LDE-height bucket" -- it is the *shared* LDE height now, which is the whole point of the PR. The module-level `//!` block got updated carefully, that one line got missed.

stir/src/utils.rs
```diff
@@ -109,6 +109,74 @@ pub fn eval_degree_correction<F: Field>(value: F, point: F, r_comb: F, gap: usiz
...
+    let g = F::two_adic_generator(log_domain);
+    let step_start = r_comb * EF::from(shift);
+
+    const POWER_CHUNK: usize = 1 << 12;
+    let mut result = EF::zero_vec(n);
+
+    for &(r_i, gap, values) in groups {
+        assert_eq!(values.len(), n, "group values must span the full coset");
+
+        let mut denoms = EF::zero_vec(n);
```
tcoratger: `denoms[p] = 1 - r_comb*shift*g^p` does not depend on the group at all -- only `g_hi` and `step_start_hi` below use `gap`, and nothing uses `r_i`. So for `m` classes this does `m` identical length-`n` power sweeps and `m` identical length-`n` batch inversions. At the bench shape that is `n = 2^21` and `m = 3`, i.e. two wasted sweeps, two wasted batch inversions over 2M extension elements, and two wasted ~84 MB allocation pairs. Hoisting `denoms`/`inv_denoms` above the `for &(r_i, gap, values) in groups` loop is a pure win; the rest of the body already only reads them, and `test_combine_on_coset_matches_naive_reference` exercises three groups so it covers the change.

Separate and much smaller: `eval_degree_correction` special-cases `step == ONE` (returns `gap + 1`), and this path does not -- a zero denominator reaches `Field::inverse()` inside `batch_multiplicative_inverse` and panics. So the prover aborts on an input the verifier's reference path evaluates cleanly:

```
log_domain = 4, shift = GENERATOR, r_comb = (shift * g^3)^{-1}, gap = 5, values[3] = 4
eval_degree_correction  -> 24        // value * (gap + 1)
combine_on_coset        -> panic
```

Probability is `~n/|F| ~ 2^-134`, so this is not a real availability concern. It matters only because these two are meant to be pointwise identical, and the naive reference in the test agrees on every non-degenerate input, so the divergence is invisible -- easy for a future refactor to read as intentional. Either mirror the branch after the inversion or `debug_assert!` that no denominator vanishes and say so in the doc.

stir/src/pcs.rs
```diff
@@ -875,15 +1019,53 @@ where
...
+                            for (log_native_h, ro_class) in &expected_ro_by_class {
+                                let &(r_i, gap) = coeffs_by_height
+                                    .get(log_native_h)
+                                    .ok_or(StirError::InvalidProofShape)?;
+                                for (q_idx, &j) in first_round_unique_js.iter().enumerate() {
+                                    let mut fiber_point =
+                                        Val::GENERATOR * domain_gen.exp_u64(j as u64);
+                                    for l in 0..arity0 {
+                                        let x = Challenge::from(fiber_point);
+                                        combined[q_idx][l] += eval_degree_correction(
```
tcoratger: This is the verifier hot path the PR exists to speed up, and it is leaving a fair bit on the table. `eval_degree_correction` does a full `EF::inverse()` plus a `~log2(gap)`-step `exp_u64` per call, and this loop calls it `n_q * arity0 * m` times. Two separate wins:

- `(1 - r_comb*x)^{-1}` is *identical across classes* at a given `(q, l)` -- the loop nesting (class outermost) recomputes it `m` times.
- Even the distinct `n_q * arity0` of them should go through one `batch_multiplicative_inverse`, which turns `k` inversions into 1 inversion plus `3k` multiplies.

At the bench config (262 queries, `arity0 = 4`, `m = 3`) that is on the order of 3000 quintic-extension inversions collapsing to one. Against a 2.9-4.3 ms verify that is not a rounding error.

The shape I would suggest is the pointwise mirror of `combine_on_coset` -- which is what the comment above already claims this is: (1) walk the `n_q * arity0` fiber points once building `1 - r_comb*x` into a flat `Vec` and batch-invert it, (2) per class sweep the numerator `1 - (r_comb*x)^{gap+1}` with the same fixed-ratio trick (`g_hi = g^{gap+1}`, seeded once per query), which also kills the per-lane `exp_u64`, (3) accumulate `r_i * ro_class * numer * inv_denom`. `denom_diffs`/`inv_denoms` ~140 lines up is already exactly this pattern.

Two smaller ones in the accumulation loop above: `expected_ro_by_class.entry(log_native_h)` runs once per `(query, lane, matrix)` triple even though a matrix's class is loop-invariant (hoist to a per-matrix index), and `bucket_points.iter().position(...)` is a linear scan in the innermost loop -- that one predates this PR, but it is now inside a loop that also fans out over classes.

stir/src/pcs.rs
```diff
@@ -901,6 +1083,85 @@ where
...
+    r_comb: EF,
+    log_d_star: usize,
+    sorted_log_ds: impl Iterator<Item = usize>,
+) -> (u64, Vec<(EF, usize)>) {
+    let d_star = 1u64 << log_d_star;
+    let mut running_exp = 0u64;
+    let coeffs = sorted_log_ds
+        .map(|log_d| {
+            let r_i = r_comb.exp_u64(running_exp);
+            let gap = (d_star - (1u64 << log_d)) as usize;
```
tcoratger: Unchecked precondition that fails silently rather than loudly: the function takes an arbitrary `impl Iterator<Item = usize>` plus a separately-passed `log_d_star`, so if any `log_d > log_d_star` this `u64` subtraction wraps in release and produces a `gap` near `2^64`, which flows into `exp_u64` and yields a wrong codeword instead of a panic.

Both call sites pass `native_heights[0]` / `classes[0].0` off descending-sorted lists so it is safe today, but that safety lives entirely outside the function. `d_star.checked_sub(1u64 << log_d).expect("classes must be sorted descending, d_i <= d*")` costs nothing once per class and converts a silent wrong answer into a loud one.

Also: the `d_star` in the return tuple is discarded at both call sites (`let (_, coeffs) = ...`), so it could just return the `Vec`.

stir/src/pcs.rs
```diff
@@ -901,6 +1083,85 @@ where
...
+    let log_d_star = classes[0].0;
+    let r_comb: Challenge = challenger.sample_algebra_element();
+    let (_, coeffs) =
+        combine_coefficients(r_comb, log_d_star, classes.iter().map(|&(log_d, _)| log_d));
+
+    // `combine_on_coset` indexes its inputs (and produces its output) in natural order, but
+    // `reduced_openings` is bit-reversed (built from the bit-reversed LDE matrices), so each
+    // class's codeword is un-reversed before combining; the combined result is then already
+    // in the natural order STIR expects, with no further reversal needed.
+    let natural_ros: Vec<Vec<Challenge>> = classes
```
tcoratger: Minor, but this is a full clone of every class's codeword purely to un-bit-reverse it -- the pre-PR code did `reduced_openings.remove(&log_h)` and reversed in place. At the bench shape that is `3 * 2^21 * 20 B` ~ 126 MB of extra allocation and copy, and it stacks with `combine_on_coset`'s own `result` + `denoms` + `inv_denoms`.

`combine_on_coset` only needs `&[EF]`, so taking `&mut BTreeMap` (or moving it) and `remove`-ing each class would let you reverse in place and drop peak memory from `2*m*n` to `m*n + n`.

stir/tests/stir.rs
```diff
@@ -1074,7 +1074,7 @@ mod babybear_pcs {
 
         let (commit, data) =
             <MyPcs as Pcs<Challenge, Challenger>>::commit(&pcs, domains_and_polys.iter().cloned());
-        commit.iter().for_each(|c| p_challenger.observe(c.clone()));
+        p_challenger.observe(commit.clone());
```
tcoratger, line 1077: This and `test_pcs_multiple_different_degrees` are the only tests that reach Combine at all, and both are happy-path. Given that the degree-correction gap is now the *only* thing enforcing a short class's degree bound, I would want at least a couple of negative cases. Concretely, the probe I used while reviewing works as a test:

> Commit two matrices at native height `2^8` (one class, no Combine), prove, then verify claiming matrix 1 has domain `2^6` (two classes, Combine runs) -> expect `Err(InvalidProofShape)`. And symmetrically: commit `[8, 6]` honestly, verify claiming `[8, 8]`, so the verifier skips Combine where the prover ran it.

Both pass on this branch, so it is just pinning behaviour that already works. Worth adding a tampering case too -- perturb one opened value in the *short* class of a passing `[8, 6]` proof, since `test_pcs_tampered_opening_value_fails` runs at a single height and would not catch a Combine-coefficient error.

Two other gaps while I am here:

- **No `JohnsonBound` PCS test anywhere** -- every `babybear_pcs` test hardcodes `CapacityBound`. A `security_level: 64` / `&[16, 14]` variant would have surfaced the config blocker before review.
- **`combine_on_coset` deserves a proptest rather than one seed.** Exact equality against a five-line reference over a large input space with no setup is about as clean a fit as it gets: `log_domain in 2..=14`, `num_groups in 1..=4`, per-group `gap in 0..2^20` (include `0` and multiples of `n` to hit `g_hi = 1`), random `r_i`/`r_comb`/values; invariant is elementwise equality with `naive_combine_on_coset`. A failure means the chunked sweep desynced from the true `p`-indexed powers, i.e. a `POWER_CHUNK` boundary bug or a wrong `g_hi` seed -- and the current test fixes `log_domain = 13` (exactly 2 chunks), so it cannot see a boundary bug at 1, 3, or a partial chunk, and `n` is never below `POWER_CHUNK`.
- A known-answer test for `combine_coefficients` would pin the recurrence to Definition 4.11 rather than to itself: `d_i = [8, 4, 2]`, `d* = 8` -> exponents `[0, 5, 12]`, gaps `[0, 4, 6]`. This is the piece least likely to fail loudly if it drifts, since a wrong `r_i` still yields a low-degree codeword and STIR would happily accept the prover's own miscomputation as long as both sides agree.

tcoratger (commented):
Spent a while on this one since it reshapes both the commit layout and the soundness accounting. The `Combine` restructure is a nice idea and the protocol math checks out end to end. Things I actually verified rather than assumed:

- `combine_coefficients` reproduces Definition 4.11 exactly (`r_1 = 1`, `r_i = r^{(i-1) + sum_{j<i}(d* - d_j)}` via the `running_exp += 1 + gap` recurrence).
- `ell = m(d*+1) - sum d_i = sum_i (gap_i + 1)`, which is the right multiplicity argument for Lemma 4.13.
- Degrees line up: term `i` lands at `<= (d_i - 2) + gap_i = d* - 2 < d*`, and `log_stir_degree == log_d_star`, so STIR's claimed degree and Combine's target agree.
- Order conventions: the verifier's `GENERATOR * g^{j + l*2^{log_h - log_arity0}}` is exactly `combine_on_coset`'s natural index `p`, and both branches hand STIR a natural-order codeword.
- Transcript: `r_comb` is sampled per bucket, descending, only when there are >= 2 classes, at the same position on both sides; `num_reduced` / `class_num_reduced` walk commitments -> matrices -> points identically.
- `split_rows(1 << log_native_h)` is the necessary fix, not cosmetic: rows `0..h` of a bit-reversed LDE are the `shift*H_h` evaluations, so interpolation still recovers the degree-`<h` polynomial now that matrices sit on the taller shared domain. The old `mat.height() >> log_blowup` would have been wrong for every short matrix.
- And I checked the new binding actually binds: prove honestly over native heights `[8, 8]` (one class, no Combine), then verify while claiming matrix 1 is at `2^6` (two classes, Combine runs) -> `Err(InvalidProofShape)`. The degree-correction gap is what enforces the claimed native height, and it does.

Two things I think need to land before this merges, one of them blocking. Details inline, short version here.

**1. `new_with_combine` panics at almost every realistic parameter set.**

`ell` is always within a factor of two of `d*` itself: the `d_i` are distinct powers of two, so `gap_i = d* - d_i` is on the order of `d*` no matter how tight the height spread is. That makes `log2(ell) + log_d*` about `2*log_d*`, i.e. Combine costs ~41 bits of field at `d* = 2^20`. Solving the `CapacityBound` requirement against the `eta <= rho/2` ceiling gives a closed form:

```
field_bits >= security_level + union_buffer + 3*log_inv_rate + 1 + log2(ell - 1) + log_d_star
```

For the benchmark shape (`ell - 1 = 2^20.70`, `log_d* = 20`, `lir = 1`, buffer 6) that is `>= 150.7`. The 155-bit quintic clears it by 4.3 bits; nothing else does. A/B on `heights = [20,18,17]`, `max_pow_bits = 20`, `log_blowup = 1` (total query count, or panic):

| field | assumption | security | base `new()` | this PR |
|---|---|---|---|---|
| 155-bit quintic KB | CapacityBound | 100 | 259 q | **262 q** :white_check_mark: |
| 155-bit | CapacityBound | 112 | 295 q | **panic** |
| 155-bit | CapacityBound | 128 | 342 q | **panic** |
| 155-bit | JohnsonBound | 100 | 511 q | **panic** |
| 124-bit quartic BB | CapacityBound | 80 | ok | **panic** |
| 124-bit | JohnsonBound | 64, `[16,14]` | ok | **panic** |

Every one of those works on the base branch. The suite misses it because the PCS tests run at `security_level: 16`, where the conjectured term is ~`2^-84` and what actually gets returned is the `2/|L_0|` side-condition floor, so the formula that binds in production is never exercised end to end. And the panic is reachable from `Pcs::verify`, not just `open`, so it is an abort where an `Err` belongs.

**2. Prover cost is missing from the benchmark table.**

Extending every matrix onto the tallest domain multiplies each short matrix's DFT *and* its share of Merkle hashing. Measured A/B, release, same machine, KoalaBear/155-bit, width 32, security 100:

| heights | base commit+open | this PR | ratio |
|---|---|---|---|
| `[20,18,17]` (the bench shape) | 3.48 + 3.63 = **7.10 s** | 7.15 + 3.10 = **10.25 s** | **1.44x** |
| `[20,10]` | 2.53 + 3.03 = **5.56 s** | 4.68 + 2.48 = **7.16 s** | **1.29x** |
| `[20,12,12,12,12]` | 2.62 + 2.46 = **5.08 s** | 11.45 + 2.85 = **14.30 s** | **2.82x** |

`open` genuinely gets faster (one STIR instance instead of three), commit more than eats it. That is a perfectly defensible trade at the bench shape given -40% proof size and -46% verify, it just needs to be stated, and it grows without bound as the height spread widens: those four `2^12` tables go from 34 ms to ~8.9 s of added commit time on their own.

Rest is smaller: two hoistable batch inversions sitting on the exact hot paths this PR is optimizing, and a few hardening nits.

stir/src/config.rs
```diff
@@ -204,6 +215,43 @@ where
...
+        ell: u64,
+    ) -> Self {
+        Self::new_impl(
+            log_starting_degree,
+            params,
+            Some(CombineRequirement { num_classes, ell }),
+        )
+    }
+
+    fn new_impl(
```
tcoratger: The naming `new_impl` is a bit strange no?

stir/src/pcs.rs
```diff
@@ -252,34 +236,56 @@ where
...
                     .to_row_major_matrix();
                 widths.push(lde.width());
-                heights.push(lde.height());
+                log_native_heights.push(log_native_height);
                 group_fiber_rows(lde, self.stir.log_starting_folding_factor)
             })
             .collect();
-        commit_by_height_class(&self.input_mmcs, ldes, heights, widths)
+        commit_shared_domain(
+            &self.input_mmcs,
```
tcoratger: Since `commit_shared_domain` always takes `self.input_mmcs` as argument, shouldn't we include `commit_shared_domain` in the implementation directly? To be able to call self.commit_shared_domain()

stir/src/pcs.rs
```diff
@@ -875,15 +1019,53 @@ where
...
 
-                    Ok(expected_ro)
+                    // Merge the per-class accumulators into the bucket's expected codeword
+                    // fibers, mirroring the prover's `combine_on_coset`/`eval_degree_correction`
+                    // pointwise, only at the queried fiber lanes.
+                    match combine_info {
+                        None => {
+                            // SHAPE CHECK: exactly one class expected when no Combine ran.
+                            if expected_ro_by_class.len() != 1 {
+                                return Err(StirError::InvalidProofShape);
```
tcoratger, line 1064: This is not for this PR but probably `InvalidProofShape` is too generic as an error, we should maybe provide some fields in it to better describe the error or have other variants since in this function `InvalidProofShape` is used for many different situations so receiving that as an error doesn't help much.

stir/src/soundness.rs
```diff
@@ -483,6 +483,72 @@ impl StirSoundness for SecurityAssumption {
...
+///   (|F|*(2*eta)^7)` (Lemma 4.13's own route, valid only up to `delta < 1 - sqrt(rho)`, so
+///   `eta` must additionally stay `<= sqrt(rho)/20` — `stir_eta_upper_bound` enforces this
+///   once the returned value is `.max()`-folded into `stir_initial_eta`, checked by
+///   `validate_eta` in `StirConfig::new`). This regime satisfies the `1/|L_0|` side
+///   condition automatically.
+///
+/// `ell` is Lemma 4.13's error argument `num_classes·(d* + 1) − Σᵢ dᵢ` (dᵢ the untouched,
+/// i.e. not further quotiented, degree bound of each class's reduced-opening).
+pub(crate) fn combine_min_log_eta(
+    assumption: SecurityAssumption,
```
tcoratger: Shouldn't this be included in `SecurityAssumption` implementation directly? Or maybe not the best place?

tcoratger:
Just left a review from my agent combined with myself :)

Nashtare:
Thanks! Will address shortly

tcoratger (approved):

PR #1992, opened 2026-08-20
perf(stir): batch the OOD-sampling and degree-correction hot loops

Nashtare:
Stacked on #1990 (Combine soundness fix + same-domain commit) — diff below is only this PR's one commit.

## Summary

Follows up on the STIR PCS perf audit's "Remaining — verifier" items 5, 7, 8, 9
(arithmetic hoists in the reduced-opening reconstruction, `fold_fiber`, degree
correction, and OOD sampling). Items 5 and 7 were already landed via #1980 before this
branch was cut; this commit covers the remaining two:

- **`sample_ood_points`**: each of the three excluded domains' shift inverses was
  recomputed as a full extension-field inversion, and each domain's `exp_power_of_2` was
  its own squaring chain, once per OOD candidate. Both are loop-invariant — the shift
  inverses move outside the sampling loop (and are derived from a cheap base-field
  inversion instead of an extension one), and every candidate now builds one shared
  doubling chain for its own value and indexes into it per domain, rather than three
  independent chains.
- **`materialize_virtual_fiber`**: the per-fiber-point degree-correction geometric sum
  recomputed `(point * r_comb)^(gap+1)` as a fresh extension-field exponentiation, and
  inverted its denominator individually, for every point of every query. `gap` is
  round-constant, so `r_comb^(gap+1)` is now hoisted into `VirtualRoundContext` once per
  round; the remaining per-point exponentiation is over the base-field coordinate alone
  (cheaper than the extension one it replaces), and the `arity`-many denominators are
  batch-inverted together instead of one at a time (excluding the negligible-density
  `step == 1` case, which needs the sum's closed form instead of a division by zero
  regardless of batching).

Proof size is unchanged — this is a pure verifier-arithmetic change, no format change.

## Benchmark

Multi-table run, `--log-message-size 20`, heights `2^{17,18,20}`, 100-bit security, each
side measured back-to-back on the same machine:

| width | verify time (before → after) | proof size |
|---|---|---|
| 32 | ~4.5ms → ~3.9ms (-17%) | unchanged (271,610 B) |
| 2  | 2947 → 2566 µs (-13%) | unchanged (118,328 B) |

## Test plan

- [x] `cargo test -p p3-stir` (29 lib + 56 integration + 22 utils tests, all green)
- [x] Release-mode benchmark, before/after via `git stash`, confirming byte-identical proofs

### DIFF
--- a/stir/src/utils.rs
+++ b/stir/src/utils.rs
@@ -285,13 +285,38 @@ where
     EF: ExtensionField<F>,
     Challenger: FieldChallenger<F>,
 {
+    // `shift^{-2^log_size}` per excluded domain, computed once in the base field (an
+    // inversion and an exponentiation cheaper than their extension-field counterparts) and
+    // reused for every candidate below, rather than an extension-field inversion redone once
+    // per domain per candidate.
+    let shift_inv_pows: [F; 3] =
+        excluded_domains.map(|(shift, log_size)| shift.inverse().exp_power_of_2(log_size));
+    let max_log_size = excluded_domains
+        .iter()
+        .map(|&(_, log_size)| log_size)
+        .max()
+        .unwrap_or(0);
+
     let mut ood_points: Vec<EF> = Vec::with_capacity(num_ood_samples);
     while ood_points.len() < num_ood_samples {
         let z: EF = challenger.sample_algebra_element();
-        let outside_all_domains = excluded_domains.iter().all(|&(shift, log_size)| {
-            let z_norm = z * EF::from(shift).inverse();
-            z_norm.exp_power_of_2(log_size) != EF::ONE || log_size == 0
-        });
+
+        // `z`'s doubling chain `z, z^2, z^4, ..., z^(2^max_log_size)`, shared across every
+        // excluded domain below instead of a fresh `exp_power_of_2` call (its own squaring
+        // chain from scratch) per domain.
+        let mut chain = Vec::with_capacity(max_log_size + 1);
+        chain.push(z);
+        for i in 0..max_log_size {
+            chain.push(chain[i].square());
+        }
+
+        let outside_all_domains =
+            excluded_domains
+                .iter()
+                .zip(&shift_inv_pows)
+                .all(|(&(_, log_size), &shift_inv_pow)| {
+                    log_size == 0 || chain[log_size] * shift_inv_pow != EF::ONE
+                });
         // Deduplicate OOD points.
         let not_dup = ood_points.iter().all(|&existing| existing != z);
         if outside_all_domains && not_dup {
--- a/stir/src/utils.rs
+++ b/stir/src/utils.rs
@@ -285,6 +285,12 @@ where
     EF: ExtensionField<F>,
     Challenger: FieldChallenger<F>,
 {
+    // Nothing to sample, and nothing to precompute for it: keeps the contract identical to
+    // evaluating the predicate lazily, which did no inversions at all in this case.
+    if num_ood_samples == 0 {
+        return Vec::new();
+    }
+
     // `shift^{-2^log_size}` per excluded domain, computed once in the base field (an
     // inversion and an exponentiation cheaper than their extension-field counterparts) and
     // reused for every candidate below, rather than an extension-field inversion redone once
@@ -295,28 +301,36 @@ where
         .iter()
         .map(|&(_, log_size)| log_size)
         .max()
-        .unwrap_or(0);
+        .expect("three excluded domains");
 
     let mut ood_points: Vec<EF> = Vec::with_capacity(num_ood_samples);
     while ood_points.len() < num_ood_samples {
         let z: EF = challenger.sample_algebra_element();
 
-        // `z`'s doubling chain `z, z^2, z^4, ..., z^(2^max_log_size)`, shared across every
-        // excluded domain below instead of a fresh `exp_power_of_2` call (its own squaring
-        // chain from scratch) per domain.
-        let mut chain = Vec::with_capacity(max_log_size + 1);
-        chain.push(z);
-        for i in 0..max_log_size {
-            chain.push(chain[i].square());
+        // One doubling chain `z, z^2, z^4, ..., z^(2^max_log_size)` walked once, recording
+        // `z^(2^log_size)` for each excluded domain as it passes — instead of a fresh
+        // `exp_power_of_2` call (its own squaring chain from scratch) per domain. There are
+        // exactly three domains, so the recording slots are a fixed-size array.
+        let mut z_pows = [EF::ONE; 3];
+        let mut acc = z;
+        for l in 0..=max_log_size {
+            for (slot, &(_, log_size)) in z_pows.iter_mut().zip(&excluded_domains) {
+                if log_size == l {
+                    *slot = acc;
+                }
+            }
+            if l < max_log_size {
+                acc = acc.square();
+            }
         }
 
-        let outside_all_domains =
-            excluded_domains
-                .iter()
-                .zip(&shift_inv_pows)
-                .all(|(&(_, log_size), &shift_inv_pow)| {
-                    log_size == 0 || chain[log_size] * shift_inv_pow != EF::ONE
-                });
+        let outside_all_domains = excluded_domains
+            .iter()
+            .zip(&shift_inv_pows)
+            .zip(&z_pows)
+            .all(|((&(_, log_size), &shift_inv_pow), &z_pow)| {
+                log_size == 0 || z_pow * shift_inv_pow != EF::ONE
+            });
         // Deduplicate OOD points.
         let not_dup = ood_points.iter().all(|&existing| existing != z);
         if outside_all_domains && not_dup {
@@ -673,7 +687,8 @@ pub fn lagrange_interpolate_at<F: Field, EF: ExtensionField<F>>(
 
 #[cfg(test)]
 mod tests {
-    use p3_baby_bear::BabyBear;
+    use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+    use p3_challenger::DuplexChallenger;
     use p3_field::PrimeCharacteristicRing;
     use p3_field::extension::BinomialExtensionField;
     use proptest::prelude::*;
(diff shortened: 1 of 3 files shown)


stir/src/verifier.rs
```diff
@@ -72,14 +77,39 @@ where
...
+                    EF::ONE
+                } else {
+                    EF::ONE - step
+                }
+            })
+            .collect::<Vec<_>>(),
+    );
+
+    Some(
+        (0..arity)
```
tcoratger: Inlining this makes it the **third** implementation of the same geometric sum, and none of the three is pinned to an independent oracle:

| implementation | tested by |
|---|---|
| `utils.rs:102` `eval_degree_correction` | nothing directly |
| `utils.rs:118` `combine_on_coset` (from #1990) | `test_combine_on_coset_matches_naive_reference`, whose reference *is* `eval_degree_correction` |
| this inlined copy | nothing directly -- only end-to-end prove/verify |

`grep -n "fn test_"` over `utils.rs` confirms there is no `eval_degree_correction` test, so `combine_on_coset` is pinned to an untested reference and this copy is pinned to nothing. End-to-end coverage only ever exercises the non-degenerate branch, at whatever `gap` the tests happen to produce.

Not hypothetical: reviewing #1990 I found `combine_on_coset` and `eval_degree_correction` already *do* diverge at `step == 1` -- the former panics inside `batch_multiplicative_inverse`, the latter returns `gap + 1`. Nothing catches it because nothing compares them there. A third copy makes the next such divergence equally invisible.

Cheapest fix is one test with a known-answer anchor plus a three-way equivalence check:

```rust
// geom = sum_{l=0}^{gap} step^l, so gap = 2, step = 2 -> 1 + 2 + 4 = 7.
assert_eq!(eval_degree_correction(EF::ONE, EF::TWO, EF::ONE, 2), EF::from_u64(7));

for gap in [0, 1, 5, 40] {
    for (x, r) in cases {              // include r = x.inverse() so step == 1
        let g1 = (gap + 1) as u64;
        let split = if r * x == EF::ONE { EF::from_usize(gap + 1) }
                    else { (EF::ONE - r.exp_u64(g1) * x.exp_u64(g1)) * (EF::ONE - r * x).inverse() };
        assert_eq!(split, eval_degree_correction(EF::ONE, EF::from(x), r, gap));
    }
}
```

Structurally nicer would be to not have three copies at all: a batched sibling in `utils.rs` (`degree_correction_on_geometric_points`) that both this fiber path and `combine_on_coset` call, so the hoisting lives in one place and one test covers every caller.

Small separate thing on these same lines: `(0..arity).map(|i| ...)` does five indexed reads per lane (`points[i]`, `row_evals[i]`, `vanishing_inverses[i]`, `steps[i]`, `denom_inverses[i]`) where the old `zip` chain had none. Only `row_evals.len() == arity` is known by construction, so LLVM has to prove four bounds separately and will likely keep some. `itertools` is not imported in this file (it is in `pcs.rs`), which is presumably why -- `izip!` over the five restores the check-free form. Mostly moot if the fusion below lands, since the tuple shrinks to three.

tcoratger (commented):
Nice tight change, and the arithmetic holds up. Things I verified rather than assumed:

- **`sample_ood_points` predicate is unchanged.** Old: `(z/shift)^{2^L} != 1 || L == 0`. New: `L == 0 || z^{2^L} * (shift^{-1})^{2^L} != 1`. Same thing, since exponentiation is a homomorphism on a commutative field, and the `||` reordering does not change the boolean. Computing `shift.inverse()` in `F` instead of `EF::from(shift).inverse()` is also fine -- `F` is a subfield, so the inverse of a base element is the same element either way.
- **The degree-correction split is right.** `step^{gap+1} = r_comb^{gap+1} * x^{gap+1}`, first factor round-constant, second now base-field. `gap` in `finish` and `gap` in the fiber both come from the same `all_points`, so the cached power matches. The `step == 1` handling is correct: substituting `EF::ONE` into the batch-inversion input keeps the divisor off zero, and that lane's result is discarded in favour of the `gap + 1` closed form.
- **Transcript really is byte-identical.** Rather than take the proof-size check on faith I pinned it directly: after a successful `verify`, draw one more extension element from the verifier's challenger. That value depends on every challenge absorbed and squeezed, so equality across branches is a strong witness. Both widths match exactly:

```
width=2   before/after: ExtField([1630532295, 1583256215, 209707065, 200128190, 5421336])
width=32  before/after: ExtField([1841934379, 1923056544, 189380930, 1672659081, 1235896846])
```

- Suite green (29 + 56 + 22), clippy `-D warnings` clean, worktree clean. All as advertised.

Nothing blocking. One thing I do want to flag on the numbers, and a test gap -- rest is inline.

**The benchmark does not reproduce for me.** `run_univariate_pcs` times a single un-warmed `verify` call:

```rust
let t = Instant::now();
pcs.verify(...).unwrap_or_else(...);
let verify_us = t.elapsed().as_micros();
```

One sample, no warmup, two sides compared across a `git stash` and a rebuild. For a ~3 ms operation that cannot resolve a few-percent effect -- code layout, allocator state and frequency all move more than that between builds.

A/B on the same shape the example builds (`heights = [n, n+1, log_message_size] = [17,18,20]`, `multi_table_width` = matrix width, CapacityBound, 100-bit, `log_blowup = 1`), 40 iterations per process, three process restarts per branch, width 2:

| | cold first shot | warm min | warm p50 |
|---|---|---|---|
| base (`c1d2f3d6`) | 6348 / 6354 / 6361 us | 6172 / 6176 / 6180 | 6205 / 6210 / 6217 |
| this PR | 6124 / 6152 / 6189 us | 5948 / 5956 / 5974 | 5990 / 6007 / 6015 |

**-3.3%** on warm p50, -3.5% on warm min, -3.1% cold -- reproducible across restarts, and -1.7% at width 32. So the win is real and in the right direction, just about a quarter of the -13%/-17% claimed. My absolute numbers are ~2.1x yours so your machine is faster, but that changes the constant, not the ratio.

The measured size is also what the arithmetic predicts: `materialize_virtual_fiber` runs ~262 times at this config, saving roughly `arity` extension exponentiations plus 3 of 4 extension inversions per call -- order 500 us of the 6.2 ms, about half of which the new allocations and rayon dispatch give straight back (inline). ~200 us measured lines up. Worth either looping the timed region in the example or quoting `stir/benches/stir_pcs.rs`, which already handles this.

Not reproducing it does not change my read on the change itself -- I would still land it. It matters because the speedup is the stated reason to accept inlining a shared helper, and I would rather the number in the PR body be one we can stand behind.

**Last thing:** despite being stacked on #1990, this is fully independent of it -- #1990 never touched `verifier.rs` and left `sample_ood_points` alone (its only `utils.rs` change was adding `combine_on_coset`). Given #1990 has a blocking config issue, this rebases cleanly onto #1989 and could land on its own rather than waiting on that stack.

stir/src/verifier.rs
```diff
@@ -72,14 +77,39 @@ where
...
-    Some(
-        row_evals
+    // Degree-correction geometric sum `(1 - step^(gap+1)) / (1 - step)` per point, `step :=
+    // point * r_comb`. `step^(gap+1) = x^(gap+1) * r_comb^(gap+1)` splits into a cheap
+    // per-point base-field exponentiation and the round-constant `r_comb_pow_gap1`
+    // (`VirtualRoundContext::finish`), instead of a fresh extension-field exponentiation of
+    // `step` at every point. `step == 1` (density `1/|EF|`) needs the sum's `gap + 1` closed
+    // form rather than a division by zero, so it is excluded from the batch inversion below.
+    let steps: Vec<EF> = points.iter().map(|&x| ctx.r_comb * x).collect();
+    let denom_inverses = batch_multiplicative_inverse(
```
tcoratger: Two batch inversions per call now, plus a throwaway `Vec` for the second one's input. The prover already solved exactly this, and says so at `prover.rs:384`:

> The quotient denominators are the vanishing evaluations scaled by the degree correction's `(1 - step)` denominator, so they are formed in place.

```rust
let mut combined_denoms = vanishing_evals;
combined_denoms.par_chunks_mut(POWER_CHUNK).enumerate().for_each(|(chunk_idx, chunk)| {
    for denom in chunk.iter_mut() { *denom *= EF::ONE - step; step *= g_next; }
});
let combined_inverses = batch_multiplicative_inverse(&combined_denoms);
```

Same fusion works here: scale `vanishing_values[i]` by `(1 - step_i)` in place (substituting `1` on a degenerate lane), invert once, then the lane is `(row_evals[i] - ans_i) * combined_inverses[i] * numer_i`. Keep the existing `vanishing_values.contains(&EF::ZERO)` guard *before* the fusion so a genuinely vanishing `Z` still gives `InvalidRoundConsistency` rather than being confused with `step == 1`. That drops one batch inversion and two allocations per call -- ~120 us across the ~262 calls, roughly 2% of verify, i.e. about 60% again on top of what the PR currently realizes.

Separately, I checked the premise that batching helps at `arity = 4` rather than assuming it -- on `QuinticTrinomialExtensionField<KoalaBear>`, 200k iterations:

```
n=4: direct 1015 ns   batch_multiplicative_inverse 448 ns   stack-Montgomery 301 ns
n=8: direct 2072 ns   batch_multiplicative_inverse 531 ns   stack-Montgomery 411 ns
```

So batching is clearly right (2.3x over `n` separate inversions, good call), but `batch_multiplicative_inverse` allocates and goes through `par_chunks_mut` even for a 4-element slice, and a plain Montgomery sweep over a stack buffer is another **33%** faster with no allocation. At `arity` in `{4, 8}` the parallel machinery can only cost. A small `batch_inverse_small<EF, const N: usize>` in `utils.rs` used by both sites here would take this function from 2 batch inversions + 3 allocations to 1 stack inversion + 0.

stir/src/verifier.rs
```diff
@@ -29,6 +29,10 @@ struct VirtualRoundContext<EF> {
     /// query can reduce it rather than re-multiplying every root.
     vanishing_coeffs: Vec<EF>,
     r_comb: EF,
+    /// `r_comb^(all_points.len() + 1)`, the degree-correction geometric sum's `r_comb` factor
```
tcoratger: The hoist is the right call. One robustness point: this is a cached derived value whose invariant is unenforced. `finish` computes `r_comb^(all_points.len() + 1)` here, and `materialize_virtual_fiber` independently recomputes `gap = ctx.all_points.len()` ~360 lines away and trusts the cached power to match. It does today, but any future change to `all_points` between construction and use -- a dedup, a truncation -- silently yields a wrong geometric sum with no test failure, since nothing checks the relationship.

One line at the top of `materialize_virtual_fiber`, free in release:

```rust
debug_assert_eq!(ctx.r_comb_pow_gap1, ctx.r_comb.exp_u64((gap + 1) as u64));
```

Or store `gap` in the context alongside the power so both come from the same place and it is structurally impossible.

While you are in this struct: `fiber_step^(gap+1)` is a round constant too, for the same reason `r_comb_pow_gap1` is. `subgroup_points` is built at `verifier.rs:241` as `fiber_step.shifted_powers(domain_gen^j).take(arity)`, so `points[i] = points[0] * fiber_step^i` -- which is also what the existing `common_power = points[0].exp_u64(arity)` relies on. That makes `x_i^{gap+1} = points[0]^{gap+1} * (fiber_step^{gap+1})^i`, turning `arity` base-field exponentiations per query into one plus `arity - 1` multiplies, exactly the sweep the prover uses at `prover.rs:372` ("each chunk seeds one exponentiation and then advances by a base-field multiply"). I costed it before suggesting it and it is small -- `gap` is ~25-40 here so each `exp_u64` is ~9 base ops, saving ~24 base multiplies per call, ~10 us total, ~0.16% of verify. Worth doing for symmetry with the prover if you are touching the round context anyway, not worth doing on its own.

Tiny doc nit two lines up: the exponentiation being replaced is `^(gap + 1)`, not `^gap` -- the field name and the first line of the comment both already have it right.

stir/src/utils.rs
```diff
@@ -270,13 +270,38 @@ where
...
         let z: EF = challenger.sample_algebra_element();
-        let outside_all_domains = excluded_domains.iter().all(|&(shift, log_size)| {
-            let z_norm = z * EF::from(shift).inverse();
-            z_norm.exp_power_of_2(log_size) != EF::ONE || log_size == 0
-        });
+
+        // `z`'s doubling chain `z, z^2, z^4, ..., z^(2^max_log_size)`, shared across every
+        // excluded domain below instead of a fresh `exp_power_of_2` call (its own squaring
+        // chain from scratch) per domain.
+        let mut chain = Vec::with_capacity(max_log_size + 1);
```
tcoratger: The shared squaring chain is a real win (3 chains of ~21/20/19 squarings plus 3 extension inversions per candidate, down to one chain of 21 plus 3 hoisted base inversions). Two cleanliness points on how it is stored.

`excluded_domains` is `[(F, usize); 3]`, so exactly three entries of this ~22-element heap `Vec` are ever read. The squaring count is already optimal -- you need `max_log_size` squarings to reach `z^{2^max}` regardless -- but a single sweep recording into `[EF; 3]` needs no allocation:

```rust
let mut acc = z;
let mut z_pows = [EF::ONE; 3];
for l in 0..=max_log_size {
    for (slot, &(_, log_size)) in z_pows.iter_mut().zip(&excluded_domains) {
        if log_size == l { *slot = acc; }
    }
    acc = acc.square();
}
```

Perf impact is negligible (~18 allocations per proof), so this is really about the code reading as the fixed-size problem it is. Same spirit for `.max().unwrap_or(0)` at line 279 -- `max()` over `[_; 3]` is always `Some`, so the fallback is dead; `.expect("three excluded domains")` documents the reasoning.

One behavioural note: `shift_inv_pows` is now computed unconditionally at entry, where the old inversions lived inside the `while` loop. With `num_ood_samples == 0` the old code did zero inversions and this does three. `stir_num_ood_samples()` only ever returns 1 or 2 so it is unreachable internally, but this function is `pub` -- an early `if num_ood_samples == 0 { return Vec::new(); }` costs nothing and keeps the contract identical.

Test-wise, this function has **no unit test at all** despite four call sites across prover and verifier (`prover.rs:486`, `prover.rs:991`, `verifier.rs:488`, `verifier.rs:1153`). A divergence would break every proof loudly, which is why the refactor is low-risk in practice -- but the predicate itself is what deserves pinning, and it is a clean proptest fit:

> **Inputs:** three `(shift, log_size)` pairs, `shift` random nonzero base, `log_size in 0..=8`; `num_ood_samples in 1..=3`; seeded challenger.
> **Invariant:** every returned `z` satisfies `(z / shift)^{2^log_size} != 1` for each domain with `log_size > 0`; all returned points distinct; `len() == num_ood_samples`.
> **What a failure proves:** the hoisted `shift^{-2^L}` and the shared chain no longer compose to `(z/shift)^{2^L}` -- exactly the identity this rewrite rests on.

A known-answer test would not generalise over the `log_size` combinations that make chain-sharing interesting (equal sizes, a zero, `max` at each of the three positions). The old three-line predicate is also worth keeping in the test module as the oracle for a direct old-vs-new equivalence check.

tcoratger (approved):

PR #1993, opened 2026-08-21
fix: migrate constant-size chunks_exact to as_chunks for Rust 1.98 clippy

jdodinh:
## Why

Rust 1.98 (released 2026-08-18) added [`clippy::chunks_exact_to_as_chunks`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#chunks_exact_to_as_chunks). CI's lint job runs `dtolnay/rust-toolchain@stable` with `-D warnings`, so it picks the new lint up automatically.

This is not branch-specific. A pristine checkout of `main` fails the lint job under 1.98. The last green run on `main` was 2026-08-20, but `rustc 1.97.1`.

## What

Mechanical migration of `chunks_exact(N)` / `chunks_exact_mut(N)` to `as_chunks::<N>()` / `as_chunks_mut::<N>()` where `N` is a constant. The chunk length moves into the type, so each chunk arrives as `&[T; N]` instead of `&[T]`.

Three places where it is more than a substitution:

- **`dft/src/butterflies.rs`** (3 loops): these used `ChunksExactMut::into_remainder()`, which `as_chunks_mut` has no equivalent for since it returns `(chunks, remainder)` up front. The loops are restructured to split first, which reads better than reaching for the remainder after the fact.
- **`sumcheck/src/strategy.rs`**: the serial path's chunks are now arrays, so four `.try_into().unwrap()` calls became redundant and are dropped. The rayon path keeps them: `par_chunks_exact` has no `as_chunks` equivalent and still yields slices.
- **`poseidon2/src/external.rs`**: `permute_mut(chunk.try_into().unwrap())` becomes `permute_mut(chunk)`.

## The one exception

`mersenne-31/src/aarch64_neon/packing.rs` has four sites whose chunk strides are `N * D * 2` and `D * 2`, where `N` and `D` are const generic parameters. Those cannot be written as `as_chunks::<{ N * D * 2 }>()` without `generic_const_exprs`, which is unstable:

```
error: generic parameters may not be used in const operations
```

They carry a single scoped `#[allow]` on the enclosing function with a `reason`, rather than a blanket allow. Arguably a clippy false positive worth reporting upstream, since the lint fires where its own suggestion does not compile. A fifth site in the same file uses a bare `N` and does migrate.

Note this file is `aarch64`-only, so it is not what breaks the CI lint job (ubuntu x86_64). It shows up when running clippy on Apple silicon.

## Checks

Run against 1.98.0 stable, matching the CI lint job step for step:

- `cargo +stable clippy --all-targets -- -D warnings`
- `cargo +stable sort --workspace --grouped --check`
- `cargo +stable doc --no-deps --workspace --document-private-items` with `-D rustdoc::broken_intra_doc_links`
- `cargo test --workspace`

No functional change intended: `as_chunks` and `chunks_exact` agree on both the chunks and the remainder.

### DIFF
--- a/batch-stark/tests/simple.rs
+++ b/batch-stark/tests/simple.rs
@@ -207,7 +207,7 @@ impl<F: Field + PrimeCharacteristicRing> PeriodicAir<F> {
 
     fn valid_trace(&self, rows: usize) -> RowMajorMatrix<F> {
         let mut values = F::zero_vec(rows * 2);
-        for (i, row) in values.chunks_exact_mut(2).enumerate() {
+        for (i, row) in values.as_chunks_mut::<2>().0.iter_mut().enumerate() {
             row[0] = self.periodic[0][i % self.periodic[0].len()];
             row[1] = self.periodic[1][i % self.periodic[1].len()];
         }
--- a/dft/src/butterflies.rs
+++ b/dft/src/butterflies.rs
@@ -134,9 +134,12 @@ impl<F: Field> Butterfly<F> for DifButterfly<F> {
         debug_assert_eq!(shorts_1.len(), shorts_2.len());
         debug_assert_eq!(suffix_1.len(), suffix_2.len());
         let twiddle_packed = F::Packing::from(self.0);
-        let mut c1 = shorts_1.chunks_exact_mut(4);
-        let mut c2 = shorts_2.chunks_exact_mut(4);
-        for (p1, p2) in (&mut c1).zip(&mut c2) {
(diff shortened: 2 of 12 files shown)


jdodinh:
This was done in parallel in #1994.

Nashtare:
Ah apologies I hadn't seen your PR

jdodinh:
No worries, @Nashtare. For future reference, what would've been the best way to announce this?

Nashtare:
I guess we just happened to have opened those in parallel, otherwise I would have caught it no worries :)

PR #1998, opened 2026-08-22
fix(stir): fallible config derivation & dedup eta-parameterized security formulas

Nashtare:
## Summary

  Two independent fixes from the outstanding-items audit, blockers for wiring STIR under a STARK:

  - StirConfig is fallible. StirConfig::new asserted on every infeasibility condition (folding-factor bounds, PoW budget, eta side-conditions, disjoint-coset side-condition, OOD/shake floor, ...). Since
    TwoAdicStirPcs::open/verify derive a fresh config per LDE-height bucket, a parameter set that's feasible at one trace height could abort the whole process at another — a verifier panic instead of a rejected proof. Added
    StirConfig::try_new/try_new_with_combine returning Result<_, StirConfigError> (one variant per former assert!); new/new_with_combine are now thin unwrap_or_else(|e| panic!("{e}")) wrappers, so existing panic-message
    tests are unaffected. verify() propagates StirConfigError via a new StirError::Config variant instead of panicking; open() keeps the panicking form since Pcs::open can't return a Result.
  - Configs are memoized. StirConfig::new runs an 80-iteration floating-point bisection per call; open/verify re-derive it once per height bucket, and the same log_stir_degree recurs across calls and across proofs.
    TwoAdicStirPcs now caches the base (non-Combine) config keyed by log_stir_degree, following the Radix2Dit twiddle-cache pattern (Arc<RwLock<BTreeMap<...>>>). Required adding Challenge, Challenger as first-class generics
    on TwoAdicStirPcs so the cache's value type can name StirConfig<Val, Challenge, StirMmcs, Challenger> — a small ripple into 3 downstream type aliases (tests, bench, example). Combine configs stay uncached: their cache
    key would need (log_stir_degree, num_classes, ell), not just a usize.
  - Zero-opening-point matrices are rejected, not silently mis-verified. verify() derived a matrix's width from point_claims.first().map(...).unwrap_or(0), so a matrix opened at zero points got width 0 fed into the MMCS
    dimensions — an honest proof would fail with a confusing WrongWidth error naming the Merkle layer rather than the actual cause. Added StirError::MatrixWithoutOpeningPoints, mirroring FRI's own guard
    (fri::verifier::FriError::MatrixWithoutOpeningPoints). open() also skips the now-dead per-matrix alpha-batched dot product (O(height × width)) for such a matrix, since it never contributes to the reduced opening.
  - Dedup the eta-parameterized proximity formulas. p3-security's list_size_bits/prox_gaps_error/ood_error/fold_sumcheck_error/queries_combination_error all grade a fixed, regime-default eta. stir/src/soundness.rs had
    re-implemented all five at an explicit, per-round eta it derives via its own bisection — a second copy that had already drifted from the originals. Promoted each to a _at_log_eta sibling in p3-security; the fixed-eta
    original now delegates to it at self.log_eta(log_inv_rate) (the JohnsonBound proximity-gap branch derives its BCSS25 m from log_eta and calls the existing prox_gaps_error_jb_at_m rather than re-deriving the dominant-term
    formula). Deleted STIR's five private copies; shake_check_error has no p3-security analog and is unchanged.

 ## Test plan

  - [x] cargo test -p p3-stir -p p3-security — all passing, including two new regression tests: try_new_rejects_infeasible_degree_without_panicking (reproduces the audit's exact "feasible at one degree, infeasible at a
    larger one" scenario) and test_pcs_matrix_without_opening_points_rejected.
  - [x] cargo test -p p3-whir -p p3-uni-stark -p p3-fri — unaffected, all passing (sanity-checks the shared p3-security formula changes are behavior-preserving).
  - [x] cargo build --workspace --all-targets — clean.
  - [x] cargo clippy -p p3-stir -p p3-security --all-targets — no warnings.
  - [x] cargo fmt --check — clean.
  - [x] Confirmed the m = 10 JB proximity-gap round-trip through pow/log2 is numerically exact (no ULP drift past the audit's 1e-9-tolerance regression tests).

### DIFF
--- a/security/src/assumption.rs
+++ b/security/src/assumption.rs
@@ -74,23 +74,48 @@ impl SecurityAssumption {
         }
     }
 
-    /// `log₂(L⁺)` for the regime's list size at distance δ.
+    /// `log₂(η)`, or `0.` for [`Self::UniqueDecoding`], where no eta term applies and every
+    /// `_at_log_eta` formula ignores the value in its UD branch.
+    pub(crate) const fn log_eta_or_zero(&self, log_inv_rate: usize) -> f64 {
+        match self {
+            Self::UniqueDecoding => 0.,
+            _ => self.log_eta(log_inv_rate),
+        }
+    }
+
+    /// `log₂(L⁺)` for the regime's list size at an explicit `log_eta`, rather than the
+    /// regime's own default safety margin ([`Self::log_eta`]).
+    ///
+    /// A caller deriving its own schedule of `eta` values across a protocol (STIR's per-round
+    /// bisection, for instance) needs the list size at each of those working points, not just
+    /// at the regime's fixed default; [`Self::list_size_bits`] is the
+    /// `log_eta = self.log_eta(log_inv_rate)` specialization of this.
     #[must_use]
-    pub const fn list_size_bits(&self, log_degree: usize, log_inv_rate: usize) -> f64 {
+    pub const fn list_size_bits_at_log_eta(
+        &self,
+        log_degree: usize,
+        log_inv_rate: usize,
+        log_eta: f64,
+    ) -> f64 {
         match self {
             // In UD the list size is 1
             Self::UniqueDecoding => 0.,
 
             // By the JB, RS codes are (1 - sqrt(rho) - eta, (2*eta*sqrt(rho))^-1)-list decodable.
             Self::JohnsonBound => {
-                let log_eta = self.log_eta(log_inv_rate);
                 let log_inv_sqrt_rate: f64 = log_inv_rate as f64 / 2.;
                 log_inv_sqrt_rate - (1. + log_eta)
             }
 
             // In CB we assume that RS codes are (1 - rho - eta, d/rho*eta)-list decodable (see Conjecture 5.6 in STIR).
-            Self::CapacityBound => (log_degree + log_inv_rate) as f64 - self.log_eta(log_inv_rate),
+            Self::CapacityBound => (log_degree + log_inv_rate) as f64 - log_eta,
         }
+    }
+
+    /// `log₂(L⁺)` for the regime's list size at distance δ.
+    #[must_use]
+    pub const fn list_size_bits(&self, log_degree: usize, log_inv_rate: usize) -> f64 {
+        self.list_size_bits_at_log_eta(log_degree, log_inv_rate, self.log_eta_or_zero(log_inv_rate))
     }
 
     /// Proximity-gap error in bits for combining `num_functions` functions
@@ -113,36 +138,72 @@ impl SecurityAssumption {
         field_size_bits: usize,
         num_functions: usize,
     ) -> f64 {
+        self.prox_gaps_error_at_log_eta(
+            log_degree,
+            log_inv_rate,
+            field_size_bits,
+            num_functions,
+            self.log_eta_or_zero(log_inv_rate),
+        )
+    }
+
+    /// [`Self::prox_gaps_error`] at an explicit `log_eta`, rather than the regime's own
+    /// default safety margin ([`Self::log_eta`]).
+    ///
+    /// A caller deriving its own schedule of `eta` values across a protocol (STIR's per-round
+    /// bisection, for instance) needs the proximity-gap error at each of those working
+    /// points, not just at the regime's fixed default. On [`SecurityAssumption::JohnsonBound`]
+    /// this derives the \[BCSS25\] proximity parameter `m = max(ceil(√ρ / (2η)), 3)` from
+    /// `log_eta` and defers to [`Self::prox_gaps_error_jb_at_m`] — at
+    /// `log_eta = self.log_eta(log_inv_rate)` that derivation reduces to exactly `m = 10`,
+    /// [`Self::prox_gaps_error`]'s fixed safety choice.
+    #[must_use]
+    pub fn prox_gaps_error_at_log_eta(
+        &self,
+        log_degree: usize,
+        log_inv_rate: usize,
+        field_size_bits: usize,
+        num_functions: usize,
+        log_eta: f64,
+    ) -> f64 {
         assert!(
             num_functions >= 2,
             "num_functions must be >= 2 to compute proximity gaps error",
         );
 
-        // Note that this does not include the field_size
-        let error = match self {
+        match self {
             // In UD the error is |L|/|F| = d/(rho*|F|)
-            Self::UniqueDecoding => (log_degree + log_inv_rate) as f64,
+            Self::UniqueDecoding => {
+                let error = (log_degree + log_inv_rate) as f64;
+                let num_functions_1_log = libm::log2(num_functions as f64 - 1.);
+                field_size_bits as f64 - (error + num_functions_1_log)
+            }
 
             // From Theorem 1.5 in [BCSS25] "On Proximity Gaps for Reed-Solomon Codes":
             //
             // For gamma < J(delta) - eta, the number of exceptional z's is bounded by:
             //   a > (2(m + 1/2)^5 + 3(m + 1/2)*gamma*rho) / (3*rho^(3/2)) * n + (m + 1/2) / sqrt(rho)
             //
-            // With eta = sqrt(rho)/20 (safe gap), m = max(ceil(sqrt(rho)/(2*eta)), 3) = max(10, 3) = 10.
-            //
-            // This improves over [BCI+20] which had:
-            //   log_2(a) = 2*log_degree + 3.5*log_inv_rate + 23.24
-            Self::JohnsonBound => jb_prox_gaps_dominant_term_bits(log_degree, log_inv_rate, 10),
+            // m = max(ceil(sqrt(rho)/(2*eta)), 3).
+            Self::JohnsonBound => {
+                let log_sqrt_rho_over_2eta = -(log_inv_rate as f64) / 2. - 1. - log_eta;
+                let m = libm::ceil(libm::pow(2., log_sqrt_rho_over_2eta)).max(3.) as usize;
+                Self::prox_gaps_error_jb_at_m(
+                    log_degree,
+                    log_inv_rate,
+                    field_size_bits,
+                    num_functions,
+                    m,
+                )
+            }
 
             // In CB we assume the error is degree/(eta*rho^2)
             Self::CapacityBound => {
-                (log_degree + 2 * log_inv_rate) as f64 - self.log_eta(log_inv_rate)
+                let error = (log_degree + 2 * log_inv_rate) as f64 - log_eta;
+                let num_functions_1_log = libm::log2(num_functions as f64 - 1.);
+                field_size_bits as f64 - (error + num_functions_1_log)
             }
-        };
-
-        // Error is (num_functions - 1) * error/|F|;
-        let num_functions_1_log = libm::log2(num_functions as f64 - 1.);
-        field_size_bits as f64 - (error + num_functions_1_log)
+        }
     }
 
     /// Johnson-bound proximity-gap error (\[BCSS25\] Theorem 1.5, dominant
--- a/security/src/assumption.rs
+++ b/security/src/assumption.rs
@@ -76,7 +76,12 @@ impl SecurityAssumption {
 
     /// `log₂(η)`, or `0.` for [`Self::UniqueDecoding`], where no eta term applies and every
     /// `_at_log_eta` formula ignores the value in its UD branch.
-    pub(crate) const fn log_eta_or_zero(&self, log_inv_rate: usize) -> f64 {
+    ///
+    /// Public alongside the `_at_log_eta` family: a caller feeding those its own schedule of
+    /// eta values still needs this convention to reproduce the fixed-eta entry points, and
+    /// has no other way to spell "the regime's default" for `UniqueDecoding`.
+    #[must_use]
+    pub const fn log_eta_or_zero(&self, log_inv_rate: usize) -> f64 {
         match self {
             Self::UniqueDecoding => 0.,
             _ => self.log_eta(log_inv_rate),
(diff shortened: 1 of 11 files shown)


stir/src/config.rs
```diff
@@ -199,26 +201,153 @@ pub struct StirConfig<F, EF, M, Challenger> {
...
 impl<F, EF, M, Challenger> StirConfig<F, EF, M, Challenger>
 where
     F: TwoAdicField,
     EF: ExtensionField<F>,
     M: Mmcs<EF>,
     Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F>,
 {
+    /// Derive a full STIR configuration from user-facing parameters.
+    ///
+    /// `log_starting_degree` is log₂ of the degree of the polynomial to commit to.
```
tcoratger, line 455: This is the main thing I'd want addressed. `try_new` is documented and used as the total, non-panicking form, and `pcs::verify` now relies on that — but `soundness.rs` keeps three `assert!`s that are reachable from here, so `verify` can still abort the process:

```
soundness.rs:80   log2_field_minus_domain        field_size_bits <= log_domain_size
soundness.rs:89   query_count_from_failure_base  failure_base outside (0, 1)  -- every log_blowup = 0 config
soundness.rs:103  minimum_eta_for_target         bound infeasible at the largest permitted eta
```

291 of 720 grid points panic through `try_new` (grid and backtrace in the review body). Two clusters:

- **`log_blowup = 0`** (180 of them): rate `rho = 1`, so `failure_base = rho + eta > 1` and `query_count_from_failure_base` fires. This is a static property of `StirParameters` — worth rejecting up front alongside the other cheap checks rather than letting the query-count formula trip over it later. Is `log_blowup = 0` even meant to be supported? Rate 1 leaves no redundancy to test proximity against, so if the answer is "no", one more variant closes this whole cluster.
- **field too small for the target** (111, of which 85 on the 124-bit quartic and 26 on the 155-bit quintic, mostly `sec = 128` and `pow = 0`): `stir_initial_eta` -> `minimum_eta_for_target` asserts. Repro: KoalaBear + `BinomialExtensionField<_, 4>`, JohnsonBound, `security_level = 100`, `max_pow_bits = 20`, `log_blowup = 1`, degree 20.

`minimum_eta_for_target` already takes a `label: &str` for its message, so it maps straight onto a variant:

```rust
#[error("{label} reaches only {upper_bits:.4} bits at the largest permitted eta \
         ({upper_bound}); target is {target_bits} bits")]
EtaInfeasibleForTarget { label: &'static str, upper_bits: f64, upper_bound: f64, target_bits: usize },

#[error("STIR query-count formula requires a failure base in (0, 1), got {failure_base}")]
InvalidFailureBase { failure_base: f64 },

#[error("challenge field ({field_size_bits} bits) must contain points outside the \
         2^{log_domain_size} evaluation domain")]
FieldTooSmallForDomain { field_size_bits: usize, log_domain_size: usize },
```

Once that's done the contract is worth proptesting, since "never panics" is now the whole point of this function:

> **Inputs:** `soundness_type in {JB,CB,UD}`, `security_level in 1..=160`, `max_pow_bits in 0..=32`, `log_blowup in 0..=4`, `log_folding_factor in 1..=4`, `log_starting_folding_factor in 1..=4`, `log_starting_degree in 0..=28`.
> **Invariant:** `catch_unwind(|| try_new(..)).is_ok()` — every input gives `Ok` or a `StirConfigError`, never a panic. Plus: `new` panics iff `try_new` errors.
> **What a failure proves:** an infeasibility path that bypasses `StirConfigError`, i.e. a `Pcs::verify` abort.

Proptest rather than a fixture table here because the space is large, the check is one line, and the corners that matter (`log_blowup = 0`, `security_level` just above `max_pow_bits`, degree at the two-adicity edge) are exactly what a generator finds and a hand-written list misses. It would have found all 291.

Separately: `StirConfigError` has eleven variants and only `InvalidEta` is covered by a test (`config.rs:1301`). Most of the rest are one-liners to trigger (`FoldingFactorTooSmall`, `UnsupportedSoundnessType`, `SecurityLevelNotGreaterThanPow`, `InitialDomainExceedsTwoAdicity`) — a table-driven bad-params-to-variant test is worth having now these are public API.

stir/src/pcs.rs
```diff
@@ -154,26 +155,79 @@ fn lde_views<'a, Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>>(
...
+    TwoAdicStirPcs<Val, Dft, InputMmcs, StirMmcs, Challenge, Challenger>
+where
+    Val: TwoAdicField,
+    Challenge: ExtensionField<Val>,
+    StirMmcs: Mmcs<Challenge>,
+    Challenger: FieldChallenger<Val> + GrindingChallenger<Witness = Val>,
+{
+    /// Returns the derived STIR config for `log_stir_degree`, computing and caching it on
+    /// first use.
+    fn get_or_try_compute_stir_config(
```
tcoratger, line 235: The cache is well built (correct double-checked locking, bounded by two-adicity, errors not cached so a bad proof can't poison it) — but as keyed it never fires for the shape this stack is built around.

`try_new_with_combine` is uncached, and after #1990 every matrix in one `commit()` shares an LDE domain, so a commitment holding two-or-more distinct native heights always yields one bucket with `native_heights.len() >= 2` and always takes the uncached branch — in both `open` and `verify`. For `2^{17,18,20}` in one commitment (the benchmark shape) the cache does nothing on either side. It only helps commitments whose matrices are all the same height, which is the case #1990 moved away from.

Measured cost of what's being skipped (155-bit quintic, CB, 100-bit, `log_blowup = 1`): 52 us at deg 8, 129 us at deg 14, **239 us at deg 20** — about 4% of the ~6.2 ms verify I measured on this shape reviewing #1992, recovered on every proof after the first.

The reason given is that the key "would need `(log_stir_degree, num_classes, ell)`, not just a `usize`" — that's the right key and `BTreeMap` takes it happily:

```rust
type StirConfigCache<Val, Challenge, StirMmcs, Challenger> = Arc<RwLock<BTreeMap<
    (usize, usize, u64),                       // (log_stir_degree, num_classes, ell)
    Arc<StirConfig<Val, Challenge, StirMmcs, Challenger>>,
>>>;
```

with `(log_stir_degree, 1, 0)` as the canonical no-Combine key. Bucket shapes recur across proofs of the same statement just as degrees do. `num_classes` is bounded by two-adicity and `ell` is a function of the height multiset, so the key space stays bounded by the same domain-shape constraints; a `len()` guard is a two-line backstop if you want a hard cap.

Two smaller things on this function:

**The 239 us derivation runs under the write guard**, and `spin::RwLock` doesn't park — a second thread missing on *any* key busy-spins for the whole bisection, which under rayon can mean spinning while the holder is descheduled. The DFT twiddle caches have the same shape so this is consistent rather than novel, but the computation is idempotent, so the standard fix is free:

```rust
let config = Arc::new(StirConfig::try_new(log_stir_degree, self.stir.clone())?);
Ok(self.config_cache.write().entry(log_stir_degree).or_insert(config).clone())
```

**The new `Challenge`/`Challenger` generics leak auto-trait bounds.** They reach the struct through the cache's value type, and `StirConfig`'s own `_phantom: PhantomData<(F, EF, Challenger)>` claims to *own* a `Challenger`. Previously the PCS held `PhantomData<Val>` and was `Send`/`Sync` whenever Dft/InputMmcs/StirMmcs were; now it also needs `Challenger: Send + Sync` and inherits its variance. Nothing in-tree breaks, so this is latent — `PhantomData<fn() -> (F, EF, Challenger)>` in `StirConfig` is the stdlib idiom for "I mention this but don't own one" and is unconditionally `Send + Sync`.

Finally, worth one test that the cache returns what a fresh derivation would (derive twice via the cache, compare field-by-field against `StirConfig::try_new`). That's the failure mode an under-specified key produces, and it'd be silent in the worst way: a proof generated under one config and verified under another.

stir/src/pcs.rs
```diff
@@ -508,6 +562,13 @@ where
                     .zip(opened_vals.iter())
                     .zip(data.log_native_heights.iter())
             {
+                // A matrix opened at no points contributes nothing to the reduced opening;
+                // skip the O(height * width) alpha-batched dot product below rather than
+                // computing it only to leave it unused.
+                if points_for_mat.is_empty() {
```
tcoratger: This `continue` sits *before* `reduced_openings.entry(key).or_insert_with(...)`, so a zero-point matrix no longer creates a native-height class. The verifier derives `bucket_native_heights` from the claimed domain sizes with no emptiness filter, so it still counts that height. For a commitment with matrices at *different* native heights where the short one is opened at no points:

```
prover:   1 class  -> no Combine, no r_comb drawn, StirConfig::new
verifier: 2 classes -> Combine,   r_comb drawn,    try_new_with_combine
```

The transcript forks before the new guard runs, since `bucket_combine` is sampled well before `initial_fibers` is invoked.

I expected that to mask the clean error and tested it — it doesn't:

```
heights [2^8, 2^6] in one commitment, only 2^8 opened:
  open succeeded, proof has 1 bucket(s)
  verify: rejected with MatrixWithoutOpeningPoints { commitment: 0, matrix: 1 }
```

The width check happens to fire first. But that's ordering luck rather than design — if the Combine config shifted `pow_bits`, `InvalidPowWitness` would surface first and point at the wrong cause. The guard depends on nothing but `commitments_with_opening_points`, so it belongs in the up-front public-input validation next to the existing `proof.len() != bucket_log_heights.len()` and `input_openings.len()` checks; then the transcript never forks and the error is the first thing that happens.

Symmetrically, `open` should reject rather than silently skip — right now it hands back a proof that cannot verify. The degenerate case makes that concrete: with *every* matrix opened at no points, `reduced_openings` ends up empty and `open` emits a proof with no STIR instances at all:

```
open succeeded, proof has 0 bucket(s)
verify: rejected with InvalidProofShape
```

Safe (verify derives its buckets from the commitments), but before this PR the entries were created unconditionally so this state was unreachable. An `assert!` on `reduced_openings.is_empty()` — or the symmetric rejection — turns a caller bug into a named failure instead of a mystery proof.

security/src/assumption.rs
```diff
@@ -112,37 +137,73 @@ impl SecurityAssumption {
...
-            // With eta = sqrt(rho)/20 (safe gap), m = max(ceil(sqrt(rho)/(2*eta)), 3) = max(10, 3) = 10.
-            //
-            // This improves over [BCI+20] which had:
-            //   log_2(a) = 2*log_degree + 3.5*log_inv_rate + 23.24
-            Self::JohnsonBound => jb_prox_gaps_dominant_term_bits(log_degree, log_inv_rate, 10),
+            // m = max(ceil(sqrt(rho)/(2*eta)), 3).
+            Self::JohnsonBound => {
+                let log_sqrt_rho_over_2eta = -(log_inv_rate as f64) / 2. - 1. - log_eta;
+                let m = libm::ceil(libm::pow(2., log_sqrt_rho_over_2eta)).max(3.) as usize;
+                Self::prox_gaps_error_jb_at_m(
```
tcoratger, line 196: This delegation is correct — I checked `m == 10` for every `log_inv_rate` in `0..=32`, and `prox_gaps_error == prox_gaps_error_jb_at_m(.., 10)` for every combination I tried. But it's worth being precise about *how* it holds, because the PR body says the round-trip is "numerically exact" and it isn't quite:

```
LOG2_10                     = 3.32192809488736218171
derived exponent (any lir)   = 3.32192809488736173762   <- one ULP BELOW, not equal
libm::pow(2., exponent)      = 9.99999999999999644729   <- ceil -> 10  OK
```

It's one ULP on the *correct side* of a `ceil`, not exact. (It's uniform across rates because `-(lir as f64)/2.` and `0.5 * lir as f64` are both exact scalings by a power of two and cancel, leaving a rate-independent residue — which is reassuring.) Had it landed one ULP the other way, `pow` would give `10.0000000000000036`, `ceil` -> **11**, and every WHIR JB proximity-gap error would move by `5*log2(11.5/10.5) ~ 0.65` bits.

Nothing in the suite would notice. Rewriting `log_eta` as `-(log_inv_rate as f64 / 2. + LOG2_10 + 1.)` — same value, different association order — is exactly the kind of tidy-up that could push the residue up one ULP.

Either make the intent explicit:

```rust
// ceil() at a value one ULP short of an integer must not round up.
let m = libm::ceil(libm::pow(2., log_sqrt_rho_over_2eta) - 1e-9).max(3.) as usize;
```

or pin it, which is cheap and worth having either way:

```rust
#[test]
fn jb_prox_gaps_default_eta_is_m_10() {
    let jb = SecurityAssumption::JohnsonBound;
    for lir in 0..=32 {
        for ld in [10, 20, 24] {
            for nf in [2, 3, 17] {
                assert_eq!(
                    jb.prox_gaps_error(ld, lir, 128, nf),
                    SecurityAssumption::prox_gaps_error_jb_at_m(ld, lir, 128, nf, 10),
                    "lir={lir} ld={ld} nf={nf}"
                );
            }
        }
    }
}
```

security/src/assumption.rs
```diff
@@ -74,25 +74,50 @@ impl SecurityAssumption {
...
             // In CB we assume that RS codes are (1 - rho - eta, d/rho*eta)-list decodable (see Conjecture 5.6 in STIR).
-            Self::CapacityBound => (log_degree + log_inv_rate) as f64 - self.log_eta(log_inv_rate),
+            Self::CapacityBound => (log_degree + log_inv_rate) as f64 - log_eta,
         }
     }
 
+    /// `log₂(L⁺)` for the regime's list size at distance δ.
+    #[must_use]
+    pub const fn list_size_bits(&self, log_degree: usize, log_inv_rate: usize) -> f64 {
+        self.list_size_bits_at_log_eta(log_degree, log_inv_rate, self.log_eta_or_zero(log_inv_rate))
```
tcoratger, line 123: All five fixed-eta entry points now route through their `_at_log_eta` sibling, and WHIR's and FRI's parameters depend on the answers. I verified this is a bit-for-bit no-op by fingerprinting against the base branch (237 lines, 17 significant digits, detail in the review body) — but that check lives in my worktree and vanishes with it.

This is the highest-value missing test in the PR: bake the current values in as a known-answer table over `{JB, CB, UD} x log_inv_rate {1,2,3,4,8} x log_degree {10,20,24}`, hardcoded at full precision, covering `list_size_bits`, `prox_gaps_error`, `ood_error`, `fold_sumcheck_error` and `queries_combination_error`. Known-answer beats proptest here specifically because the goal is to freeze *these numbers* — a property test would need the pre-refactor implementation as its oracle, and that's exactly what this PR deletes.

One small API asymmetry while I'm here: `log_eta_or_zero` is `pub(crate)` but every `_at_log_eta` function it feeds is now `pub`. An external caller of `prox_gaps_error_at_log_eta` on `UniqueDecoding` has no way to reproduce the crate's own convention. Either make it `pub` alongside them, or note in each `_at_log_eta` doc that the UD branch ignores `log_eta` entirely so any value works.

stir/tests/stir.rs
```diff
@@ -1362,6 +1362,60 @@ mod babybear_pcs {
...
+        let width = 3;
+        let domain =
+            <MyPcs as Pcs<Challenge, Challenger>>::natural_domain_for_degree(&pcs, 1 << log_d);
+        let mat_a = RowMajorMatrix::<Val>::rand(&mut rng, 1 << log_d, width);
+        let mat_b = RowMajorMatrix::<Val>::rand(&mut rng, 1 << log_d, width);
+
+        let mut p_ch = challenger_template.clone();
+        let (commit, data) = <MyPcs as Pcs<Challenge, Challenger>>::commit(
+            &pcs,
+            vec![(domain, mat_a), (domain, mat_b)],
```
tcoratger, line 1385: Both matrices are committed at `log_d = 6`, so the bucket has one native-height class either way and the prover's new `continue` can't change the class set. That means this test passes for the easy reason and doesn't reach the interesting one — see my note on `pcs.rs:568`, where distinct native heights make the prover drop a Combine class the verifier still counts.

Two variants worth adding next to this one:

1. **Distinct native heights** (e.g. `2^8` and `2^6` in one commitment, only the tall one opened): assert `MatrixWithoutOpeningPoints` *and* that no Combine challenge was drawn before it — the latter is what would actually regress if the guard stays inside the per-bucket closure.
2. **All matrices unopened**: currently `open` returns a 0-bucket proof and `verify` says `InvalidProofShape`. Pin whatever `open` is changed to do here.

Both are ~15 lines given the harness already in this file, and (1) is the one that would catch a reordering regression.

stir/src/config.rs
```diff
@@ -199,26 +201,153 @@ pub struct StirConfig<F, EF, M, Challenger> {
...
+    /// A stage's algebraic security fell short of the buffered target by more than the
+    /// configured `max_pow_bits` can bridge.
+    #[error(
+        "{round} {label} requires {needed} PoW bits to reach buffered security target = \
+         {buffered_security_level} (security_level = {security_level} + union-bound buffer = \
+         {union_bound_buffer}, algebraic bits = {algebraic_bits:.4}), but max_pow_bits = \
+         {max_pow_bits}. Increase max_pow_bits, log_blowup, or use a larger field."
+    )]
+    PowBudgetExceeded {
+        round: String,
```
tcoratger: Nit: this `String` only ever holds `"round {i}"` or `"final"`, but it pulls `alloc::string::String` into `config.rs`, makes `StirConfigError` non-`Copy`, and allocates on a path that's about to abort or reject anyway. A small enum keeps the error plain-data and reads better at the match site:

```rust
enum Stage { Round(usize), Final }
```

The `label: &'static str` field right below it is already doing the right thing, which is what made this stand out.

tcoratger (commented):
Went through this one carefully because the `p3-security` half touches formulas WHIR and FRI also consume, and a passing test suite doesn't prove numeric equality of security parameters. Good news first: **the riskiest change is the safest one.**

**The formula dedup is a genuine numeric no-op — I measured it, not assumed it.** Fingerprinted the whole pipeline at 17 significant digits on this branch vs. its base: the five fixed-eta `p3-security` functions over `{JB, CB, UD} x log_inv_rate {1,2,3,4,8} x log_degree {10,20,24}` with several `num_functions`/`ood_samples`/`num_queries` each, plus complete `StirConfig` dumps (every round's `eta`, `num_queries`, `num_ood_samples`, `pow_bits`, `folding_pow_bits`, `log_domain_size`, and all final-stage fields) over `{JB, CB} x 4 security levels x 2 blowups x 2 folding factors x 3 degrees x {155-bit quintic, 124-bit quartic}`:

```
237 fingerprint lines, base vs this PR: IDENTICAL (diff empty)
```

I also read all five promoted functions line by line against the deleted STIR copies — the JB delegation through `prox_gaps_error_jb_at_m` is literally the old inline composition, the `f64 -> usize -> f64` round-trip on `m` is exact for any integer `ceil` output, and all five keep their original `const`-ness so no const-context caller breaks. Nice refactor.

Other things I checked and found fine: the `spin::RwLock<BTreeMap<usize, Arc<_>>>` cache really does follow the `Radix2Dit`/`Radix2DitParallel` twiddle pattern (`dft/src/radix_2_dit.rs:34`), and `spin` was already a workspace dep in fri/merkle-tree/dft/whir, so no new dependency. Double-checked locking is correct, the key space is bounded by two-adicity, only *successful* derivations get inserted (so a garbage proof can't poison it), and `spin` doesn't poison so the panic path can't wedge the lock. Suites green: p3-stir 31+57+22, p3-security 65+2, and p3-whir/p3-fri/p3-uni-stark all pass.

Two things I think want fixing before this lands. Details inline, short version:

**1. `try_new` still panics — `soundness.rs` has three asserts that weren't converted.**

The description says "one variant per former `assert!`" and "`verify()` propagates `StirConfigError` ... instead of panicking". True for the `config.rs` asserts, but these three sit on the `try_new` path and survived:

```
soundness.rs:80   log2_field_minus_domain        "challenge field must contain points outside the evaluation domain"
soundness.rs:89   query_count_from_failure_base  "STIR query-count formula requires a failure base in (0, 1)"
soundness.rs:103  minimum_eta_for_target         "{label} reaches only {upper_bits:.4} bits at the largest permitted eta"
```

Sweeping `{JB,CB} x sec {16,40,80,100,128} x pow {0,20} x lb {0,1,2,3} x deg {8,12,16,20,24} x {155-bit,124-bit}` through `try_new` and catching panics:

```
TOTAL: ok=307  err=122  PANIC=291
by log_blowup:  lb=0 -> 180   lb=1 -> 34   lb=2 -> 37   lb=3 -> 40
```

Backtrace on a representative one (KoalaBear quartic / 124-bit, JB, sec=100, pow=20, lb=1, deg=20):

```
panicked at stir/src/soundness.rs:103:5
  2: minimum_eta_for_target::<...stir_initial_eta::{closure#0}>
  3: <SecurityAssumption as StirSoundness>::stir_initial_eta
  4: StirConfig::<...>::try_new_impl
  5: StirConfig::<...>::try_new
```

**In fairness, the bug you set out to fix IS fixed.** Along the motivating axis — fixed feasible params, sweeping `log_stir_degree` as the PCS does per bucket — I measured zero panics across 12 parameter sets:

```
CB sec=100 pow=20 lb=1  deg 2..26: ......................eee   panics at []
JB sec=100 pow=20 lb=1  deg 2..26: ......................eee   panics at []
   ('.' = Ok, 'e' = clean Err, 'P' = panic)
```

So "feasible at one trace height, infeasible at another" is genuinely handled. What's left is a *different* axis of infeasibility that still aborts, and it still aborts inside `verify`, which is the place the whole change exists to protect.

**2. The config cache never fires for multi-height commitments — i.e. the shape this stack was built for.**

`try_new_with_combine` is uncached, and after #1990 every matrix in one `commit()` shares a domain, so a commitment with two-or-more distinct native heights always produces one bucket with `native_heights.len() >= 2` and always takes the uncached branch. For the benchmark shape the stack is tuned on (`2^{17,18,20}` in one commitment) the cache is bypassed on both prover and verifier. Measured cost of what's being skipped, 155-bit quintic, CB, 100-bit, lb=1:

```
deg=8:   try_new  52.4 us/call
deg=14:  try_new 129.3 us/call
deg=20:  try_new 239.0 us/call
```

239 us against the ~6.2 ms verify I measured on this shape while reviewing #1992 is ~4% of verify, recovered on every proof after the first. The stated blocker — "their cache key would need `(log_stir_degree, num_classes, ell)`" — is the right key and is a perfectly good `BTreeMap` key.

Rest is smaller stuff inline: a prover/verifier asymmetry the new zero-point guard introduces, the `m = 10` round-trip being one ULP from flipping with nothing pinning it, and a couple of API/allocation nits.

tcoratger (approved):

PR #2001, opened 2026-08-25
feat(binary-field): GF(2) through GF(2^128) tower fields

Nashtare:
## Summary

  Adds `p3-binary-field`, a new crate implementing the Wiedemann tower `GF(2) ⊂ GF(4) ⊂ … ⊂ GF(2^128)` — characteristic-2 finite fields, built for eventual binary-field SNARK support. This is Phase 1 of a staged plan; it
  delivers field arithmetic, extension structure, serialization, and a Fiat-Shamir challenger, with a hardware-accelerated fast path for the two largest levels. No PCS, DFT, or `p3-multi-stark` integration yet — that's later
  phases, not started here. This scope is entirely additive: it introduces one new crate and touches no existing field's behavior.

  Stacks on top of #[Phase 0 PR] (the `Field::interpolation_node` groundwork this depends on).

  ## What's in the crate

  **`Gf2` and the tower levels** — `Gf2` (the 2-element base field), and `BinaryField2/4/8/16/32/64/128` (named by bit-width, i.e. `GF(2^N)`), each a degree-2 extension of the level below via `X² + αX + 1`. Full
  `PrimeField`/`Field`/`PrimeCharacteristicRing` implementations, generated by a single macro, with recursive reference arithmetic that's always correct regardless of target hardware.

  **Extension-field structure** — `Algebra`, `BasedVectorSpace`, `ExtensionField`, `HasFrobenius`, and `PackedFieldExtension` between every byte-aligned tower-level pair (8×16 through 64×128, plus `Algebra<Gf2>` for every
  level), including a zero-copy basis view backed by a single `unsafe` pointer cast, gated by compile-time size/alignment/endianness assertions.

  **Serialization** — canonical-range-checked `serde` support for every level, plus `RawDataSerializable`/`from_le_bytes` for byte round-tripping.

  **`BinaryChallenger`** — a byte-sampling Fiat-Shamir challenger implementing the full `p3-challenger` trait surface (`CanObserve`, `CanSample`, `CanSampleBits`, `CanSampleUniformBits`, `GrindingChallenger`,
  `FieldChallenger`), suitable for wrapping any byte-oriented inner challenger such as `HashChallenger<u8, Keccak256Hash, 32>`.

  **CLMUL fast path** — a Binius-style change of basis between the tower representation and a standard polynomial basis, hardware carryless multiplication (`PCLMULQDQ` on x86_64, `PMULL` on aarch64) with modular reduction, 
  and a bit-serial scalar fallback that's always compiled and tested regardless of target. Both halves of the isomorphism claim are checked in `const fn`s: the pinned generator images are asserted to satisfy the defining
  tower relations `ξ_k² + ξ_{k-1}·ξ_k + 1 = 0`, which is what makes the map multiplicative rather than merely linear, and the basis-change matrix is inverted at compile time, which makes it bijective.
  No build can carry a wrong `ξ`. The runtime tests re-derive the same facts independently and are a cross-check rather than the guarantee.

  **`test_binary_field!`** — a characteristic-2-aware test suite macro added to the shared `p3-field-testing` crate, invoked for all seven tower levels, including the sub-byte `BinaryField2` and `BinaryField4`
  where the representation is unusual. The macro itself is purely additive, and the suites the other 40+ field crates run are untouched, with two deliberate exceptions on the shared surface: `Field::interpolation_node`'s
  documented contract now asks for injectivity below the size of the field rather than on a fixed `0..64` — which no 4- or 16-element field can satisfy — and `test_interpolation_nodes` caps its range at `min(64, |F|)` to
  match. Every field with at least 64 elements is tested exactly as before.

  ## Benchmarks

  Measured on both aarch64 (Apple Silicon, native) and x86_64 (AWS c8a.16xlarge, AMD EPYC, isolating `+pclmulqdq` from unrelated codegen flags):

  | Operation | aarch64 | x86_64 |
  |---|---|---|
  | `mul/64` (dispatched) | ~15× latency / ~36× throughput | ~25× latency / ~45× throughput |
  | `mul/128` (dispatched) | ~26× latency / ~28× throughput | ~28× latency / ~38× throughput |
  | `mul_alpha/128` | ~27× | ~28× |
  | `inverse/128` | ~2.2× | ~2.2× |

  No level regresses versus the reference implementation on either architecture.

  ## Testing

  - `cargo test -p p3-binary-field -p p3-field-testing` — full suite, including 1000-iteration proptests for field-axiom compliance on `BinaryField128` and the extension-pair test suites for all 10 byte-aligned pairs.
  - Verified under both `RUSTFLAGS="-C target-feature=+aes"` (aarch64) and `+pclmulqdq` (x86_64, including a full run under Rosetta 2 and native codegen checks via cross-compilation) to confirm the hardware fast path and the
  scalar fallback are both exercised and agree.
  - Instruction-level verification via disassembly confirms the intended `pclmulqdq`/`pmull` instructions are actually emitted, not silently falling back to the scalar path.
  - `cargo clippy --workspace --all-targets -- -D warnings` and `cargo check --workspace --all-targets` clean.

  ## Performance follow-ups

  Four commits added after the initial review, each measured on a 14-core aarch64 laptop. Ratios are the reliable part; absolutes are context.

  **`square()` now routes through the carryless-multiply path** (`tower.rs`, `clmul/mod.rs`)

  `square` was hardcoded to `reference_square` while `Mul` dispatched through the macro's `$mul` parameter, so the 64- and 128-bit levels got CLMUL for multiplication and not for squaring. A `$square` macro parameter mirrors `$mul`, and `square_64`/`square_128` are dedicated routines rather than forwards to `mul_*(a, a)`: a square converts its operand to the polynomial basis **once** instead of twice, and at 128 bits the Karatsuba middle term vanishes in characteristic 2 (`p² = p0² + p1²·x^128`), so it needs **two** carryless products rather than three. Latency 45.87 → 17.04 ns at 128 bits and 35.14 → 11.62 at 64; throughput 35.36 → 12.32 and 15.48 → 4.01. At 128 bits this beats `x * x` by 13% latency and 30% throughput. Everything built on repeated squaring — Frobenius, `try_sqrt`, `exp_u64` — inherits it.

  **`GF(2^8)` multiplication through log and exp tables** (`tower.rs`)

  `reference_mul` recursed Karatsuba to `Gf2`, so a `GF(2^32)` product cost `3^5` leaf operations. `GF(2^8)*` is cyclic of order 255, so a 512-byte log/exp pair reduces the leaf to two lookups, an add and a lookup — and the levels above inherit it, their recursion bottoming out at the table instead of at the bit.

  | multiply | latency | throughput |
  |---|---|---|
  | `GF(2^8)` | 15.63 → 2.54 ns (6.2×) | 5.15 → 0.33 ns (**15.8×**) |
  | `GF(2^16)` | 21.43 → 5.64 ns (3.8×) | 16.68 → 1.82 ns (**9.2×**) |
  | `GF(2^32)` | 58.08 → 9.51 ns (6.1×) | 51.06 → 6.23 ns (**8.2×**) |

  Also 2.1–2.3× on squaring at 16/32 bits and 3.4× on `inverse/128`. Zero is handled branch-free by a sentinel (`LOG[0] = 0` keeps `LOG[a] + LOG[b]` in range unconditionally) plus a mask; only `try_inverse` branches, since it returns `Option`. The tables are pinned constants with an `#[ignore]`d regenerator, following `search_generator` and the Cantor-basis table. **`reference_mul` is untouched and remains the oracle**, and the table product is checked against it **exhaustively — all 65 536 ordered pairs**, as a normal test.

  **128-bit carryless products are schoolbook rather than Karatsuba** (`clmul/mod.rs`)

  Four 64×64 products instead of three plus the Karatsuba fix-ups. Isolated, this is 1.06× latency and 1.19× throughput. End to end it is a wash, and that is the more useful measurement: the carryless product is only ~3% of `mul_128` — the remaining ~97% is the tower↔polynomial basis change, which is 48 dependent byte-table lookups across ~160 KiB. Confirmed neutral in a real workload (an additive-NTT transform, p = 0.97). Kept because it is simpler and marginally ahead in isolation, but the number worth remembering is that **the 128-bit multiply is table-bound, not silicon-bound**.

  **A crate-level doc comment** (`lib.rs`), so `rustdoc` no longer warns about `missing_crate_level_docs`.

  **Housekeeping** — the crate is registered in `release-plz.toml`; `criterion` is restricted to non-wasm targets as in the sibling field crates, since it pulls `rayon` and `rayon` refuses to build for wasi; and the README
  records that `from_u64` and the other integer constructors route through the prime subfield `GF(2)`, so they carry the parity of their argument rather than its bit pattern.

  ## Notes for reviewers

  - The crate refuses to build on big-endian targets (a compile-time assertion), since the zero-copy basis view relies on little-endian byte layout matching the tower's numeric representation.
  - Exactly three `unsafe` blocks in the whole crate: one pointer-cast for the zero-copy basis view, and two single-instruction intrinsic wrappers (one per architecture) in the CLMUL module.
  - `reference_mul` is deliberately kept free of any dependency on the hardware fast path (it recurses through `.reference_mul()`, never `Mul::mul`), so the equivalence tests that check the fast path against it are 
  meaningful rather than circular.


binary-field/src/clmul/basis.rs
```diff
@@ -0,0 +1,508 @@
...
+        j += 1;
+    }
+
+    let mut pivot = 0;
+    while pivot < bits {
+        let mut col = pivot;
+        while col < bits && (mat[col] >> pivot) & 1 == 0 {
+            col += 1;
+        }
+        assert!(col < bits, "the change-of-basis matrix is singular");
```
tcoratger, line 141: This is the only compile-time assertion in the module, and it doesn't support the claim in the PR description:

> The basis-change matrix inversion happens inside a `const fn`, so the crate cannot compile at all unless the underlying tower-to-polynomial isomorphism is genuinely valid — the correctness of the fast path is a compile-time property, not just a runtime one.

Singularity proves `N` is an **invertible GF(2)-linear map**. It does not prove `N` is a *field* isomorphism — almost every invertible 128×128 matrix over GF(2) fails to be multiplicative. What makes `N` an isomorphism is that the ξ satisfy the tower relations, and that's checked by a **runtime test** (`generator_images_satisfy_the_tower_relations`, line 339).

Concretely: flip one bit of one `XI_128` entry and the crate still compiles — `N` stays invertible with overwhelming probability — while `mul_128` silently computes the wrong product in GF(2^128). Only `cargo test` catches it. A reviewer who takes the description at face value will over-trust the build.

To be clear, the module doc-comment is honest about this — it names the tests explicitly — so the gap is between the description and the code, not inside the code. And the fix makes the claim true, because `poly_mul` is already `const fn`:

```rust
/// The tower relations hold for the pinned generator images. This is what makes `N` a field
/// isomorphism rather than merely an invertible linear map, so it is checked at compile time:
/// no valid build can have a wrong `XI_*`.
const fn relations_hold(bits: usize, tail: u128, xi: &[u128]) -> bool {
    let mut k = 0;
    while k < xi.len() {
        let prev = if k == 0 { 1 } else { xi[k - 1] };
        // xi_k^2 + xi_{k-1}*xi_k + 1 == 0
        if poly_mul(xi[k], xi[k], bits, tail) ^ poly_mul(prev, xi[k], bits, tail) ^ 1 != 0 {
            return false;
        }
        k += 1;
    }
    true
}

const _: () = assert!(relations_hold(64, TAIL_64, &XI_64), "XI_64 violates the tower relations");
const _: () = assert!(relations_hold(128, TAIL_128, &XI_128), "XI_128 violates the tower relations");
```

With that in, the description is accurate, the two runtime tests become belt-and-braces, and — more to the point — the property survives any future re-derivation of the constants, including the eventual AES/GHASH basis where new ξ would need exactly this guard.

FWIW I verified all 13 relations hold today (residuals exactly zero at both levels), so this is about making the guarantee load-bearing, not about a wrong constant.

binary-field/src/clmul/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+#[inline]
+pub(crate) fn mul_64(a: u64, b: u64) -> u64 {
+    let product = clmul_64x64(basis::tower_to_poly_64(a), basis::tower_to_poly_64(b));
+    basis::poly_to_tower_64(reduce_64(product))
+}
+
+/// Multiplication in `GF(2^128)`, taking and returning the tower representation.
+#[inline]
+pub(crate) fn mul_128(a: u128, b: u128) -> u128 {
+    let (low, high) = clmul_128x128(basis::tower_to_poly_128(a), basis::tower_to_poly_128(b));
```
tcoratger, line 133: Recording a measurement rather than asking for a change — the tower-basis choice is settled for this phase, and this is the number I'd want when revisiting it.

Three table-driven conversions happen per multiply (`tower_to_poly` on each operand, `poly_to_tower` on the result), each 16 dependent lookups — 48 loads that don't pipeline. binius64's `Ghash128b` (`crates/field/src/ghash.rs`) stores elements **directly** in the polynomial basis of `x^128 + x^7 + x^2 + x + 1` — the very modulus reduced by here — so its multiply is `clmul + reduce` with no conversion at all.

Measured on aarch64 with the same hardware clmul and an identical two-round reduction, differing only in whether the basis change happens:

| | latency | throughput |
|---|---|---|
| `BinaryField128 * BinaryField128` (tower, with basis change) | 34.95 ns | 31.26 ns |
| same reduction, polynomial basis throughout | 11.55 ns | **3.85 ns** |

So **3.0× latency and 8.1× throughput** are sitting in the conversion. The throughput gap is the larger one precisely because the 48 lookups serialize.

Two notes for whenever this is revisited:

- The design already isolates the decision well — everything basis-specific is behind `tower_to_poly_*`/`poly_to_tower_*` in `basis.rs`, so switching representation touches one module (plus the new ξ, which the const assert above would then guard). That's a good structural call.
- A cheaper intermediate step if the tower basis has to stay: the current API can't hoist a shared operand's conversion. Protocol code very often multiplies a whole vector by one constant, where `tower_to_poly` of that constant should be paid once rather than per element. A `mul_by_poly_basis_constant`-style entry point (or a "prepared multiplier" type) would recover a good fraction of the 8× without changing the storage format.

binary-field/tests/field.rs
```diff
@@ -0,0 +1,297 @@
...
+        generate_from_int_tests!(Gf2, 0i64, Gf2::ZERO);
+        generate_from_int_tests!(Gf2, 1i64, Gf2::ONE);
+        generate_from_int_tests!(Gf2, 0i128, Gf2::ZERO);
+        generate_from_int_tests!(Gf2, 1i128, Gf2::ONE);
+        generate_from_int_tests!(Gf2, 0isize, Gf2::ZERO);
+        generate_from_int_tests!(Gf2, 1isize, Gf2::ONE);
+    }
+}
+
+mod binary_field_8 {
```
tcoratger, line 97: `test_binary_field!` is invoked for the five byte-aligned levels but not for `BinaryField2` or `BinaryField4`. Those two get only a handful of hand-written checks in `tower.rs` (`X² = X + 1`, an exhaustive `BinaryField4` loop, two serde cases) — and in particular **their `GENERATOR` constants are verified only by `search_generators`, which is `#[ignore]`d** and so never runs in CI.

I checked both by hand and they're correct (`2` and `5`, both full order — table in the review body), so this is a coverage gap rather than a bug. Worth closing because these are exactly the levels where the representation is unusual — sub-byte elements with masked high bits — so a future change to `MASK`/`HALF_MASK` handling would go unnoticed precisely here.

Most of the suite should apply directly. The parts that won't, so they don't surprise you:

- `test_generator` needs the factorizations of `3` and `15` — trivial.
- `test_interpolation_nodes` tests `0..64`, which *cannot* be injective in a 4- or 16-element field. That sub-test needs a bound of `min(64, 1 << BITS)`, which is a small change to the shared helper (or a separate narrower macro for the sub-byte levels).

If extending the shared helper is out of scope for Phase 1, the cheap move is to drop `#[ignore]` from `search_generators` for just the two small levels — it runs in well under a millisecond for orders 3 and 15.

binary-field/src/challenger.rs
```diff
@@ -0,0 +1,607 @@
...
+    #[instrument(name = "grind for proof-of-work witness", skip_all)]
+    fn grind(&mut self, bits: usize) -> Self::Witness {
+        // Trivial case: 0 bits mean no PoW is required and any witness is valid.
+        if bits == 0 {
+            return F::ZERO;
+        }
+
+        assert!(bits < (usize::BITS as usize));
+        assert!(
+            bits + GRIND_MARGIN_BITS <= F::bits().min(u64::BITS as usize),
```
tcoratger, line 220: With `GRIND_MARGIN_BITS = 8` this gives a per-level ceiling on PoW width:

```
BinaryField8:   bits <= 0   -> every nonzero request panics
BinaryField16:  bits <= 8
BinaryField32:  bits <= 24
BinaryField64:  bits <= 56
BinaryField128: bits <= 56   (capped by the u64 counter, not the field)
```

So `GrindingChallenger` is implemented for `BinaryChallenger<BinaryField8, _>` but can never actually grind, while `FieldChallenger<F>` is implemented uniformly for every level. The margin reasoning itself is well documented and I agree with it — the issue is only that the failure is a runtime panic on a type that advertises the capability.

Worth either documenting the ceiling (a four-line table in the doc-comment would do) or noting in `BinaryChallenger`'s docs that levels below `GF(2^32)` aren't viable PoW witness types. The `u64`-counter cap on `BinaryField128` is also worth stating explicitly — a reader may reasonably expect a 128-bit witness field to support more than 56 bits.

Unrelated, on the sampling above: masking 8 bytes down to `bits` is exactly uniform on `[0, 2^bits)` with no modulo bias, and sourcing entropy from the byte stream rather than decomposing a field element means the available width doesn't depend on `F` at all. That's a genuinely better design than the rejection-sampling challengers, and the doc explains why. No notes.

binary-field/src/tower.rs
```diff
@@ -0,0 +1,906 @@
...
+
+            #[inline]
+            fn order() -> BigUint {
+                BigUint::from(1u8) << (1usize << <Self as TowerLevel>::LOG_BITS)
+            }
+
+            /// The enumeration by bit pattern: `interpolation_node(i)` is the element whose
+            /// coefficients over `GF(2)` are the bits of `i`. It is injective for `i < 2^BITS`.
+            #[inline]
+            fn interpolation_node(i: usize) -> Self {
```
tcoratger, line 348: `from_repr` masks, so in release `BinaryField8::interpolation_node(256) == interpolation_node(0)` — the `debug_assert` is the only thing standing between a caller and two aliased interpolation nodes.

The doc-comment is accurate ("injective for `i < 2^BITS`") and `debug_assert` is how the rest of p3 handles this, so it's a judgement call. But the consequence is specific: a collision between interpolation nodes breaks an interpolation-based argument silently rather than loudly, and `interpolation_node` is called once per node at setup, never in an inner loop. A hard `assert!` costs nothing measurable here.

Worth noting the trait contract asks for injectivity "on every index a protocol can use", which is simply unsatisfiable for `BinaryField2`/`BinaryField4` — another reason those two levels want the explicit bound discussed on the test file.

tcoratger (commented):
Went through this one properly — a new field implementation is the worst place to trust "the tests pass", since the tests and the code share their assumptions. So I re-derived the math and checked the constants **outside** the crate. Short version: this is high-quality work and I found no correctness bug. One claim in the description isn't what the code checks, and ~10 lines would make it true.

**Tower arithmetic, re-derived from `X² = αX + 1`** — all four operations are right:

```
mul:     a·b = (a0b0 + a1b1) + (a0b1 + a1b0 + α·a1b1)·X   -> join(z0+z2, z1+α·z2)   OK
square:  a²  = (a0² + a1²) + α·a1²·X   (cross term 2a0a1 vanishes)                  OK
mul_α:   X·(c0 + c1X) = c1 + (c0 + α·c1)·X                                          OK
inverse: conjugate is X -> X+α, so N = a·ā = a0² + α·a0a1 + a1² lies in T_k,
         and a^-1 = ((a0 + α·a1) + a1·X)·N^-1                                        OK
```

The conjugation is right because the roots of `X² + αX + 1` are `X` and `X + α` (sum `α`, product `1`). I also confirmed `reference_mul`'s independence claim genuinely holds: it recurses through `.reference_mul()` *and* `mul_alpha` recurses into the lower level's `mul_alpha` rather than `Mul`, so the equivalence tests are non-circular. Nice.

**The ξ constants — reimplemented `poly_mul` in Python and checked the defining relations directly:**

```
GF(2^64):  xi_k^2 + xi_(k-1)*xi_k + 1 == 0  for k = 0..5   (all residuals exactly 0)
GF(2^128): xi_k^2 + xi_(k-1)*xi_k + 1 == 0  for k = 0..6   (all residuals exactly 0)
rank(N) = 64/64  and  128/128
```

All 13 relations hold exactly. Combined with the module's own argument (images satisfying the relations extend to a `GF(2)`-algebra hom; `T_n` is a field so it's injective; equal dimensions make it an isomorphism) that's a complete proof the basis change is a genuine field isomorphism. The module doc states that argument correctly and completely.

**All seven `GENERATOR` constants have full multiplicative order**, checked against the known factorizations of `2^n − 1`:

```
level  GENERATOR              order?      level  GENERATOR              order?
    2  2                      YES            32  65541                  YES
    4  5                      YES            64  4294967300             YES
    8  19                     YES           128  18446744073709551621   YES
   16  258                    YES
and X^2 == alpha*X + 1 at every level: True
```

**Both reductions.** `x^64 ≡ x^4+x^3+x+1` and `x^128 ≡ x^7+x^2+x+1` are the standard pentanomials; the spill masks (`>>60, >>61, >>63` and `>>121, >>126, >>127`) are exactly the bits lost to the shifts, and two rounds genuinely suffice — first spill has degree ≤ 3 (resp. ≤ 6), second fold lands at ≤ 7 (resp. ≤ 13). Karatsuba's `(low ^ (middle << 64), high ^ (middle >> 64))` places the middle halves correctly. Both intrinsic wrappers are right, including `_mm_set_epi64x`'s high-to-low argument order and the `unpackhi` trick to avoid needing `sse4.1`.

**The `unsafe` pointer cast is properly guarded.** The load-bearing assertion is `<$lower>::BITS == <$lower_repr>::BITS` — that's what rules out a level with non-canonical high bits and confines the impl to byte-aligned levels. Endianness/size/alignment are all const-asserted and `align_of` holds (1 ≤ 16 at the extreme).

**Frobenius** is `x -> x^(2^m)` = `x^|base|` ✓ and `repeated_frobenius` reduces mod `DIMENSION` ✓. **`try_sqrt`** returning `Some` unconditionally is right — squaring is the Frobenius of order `n`, hence a bijection.

**Benchmarks reproduce, and your numbers are conservative:**

| | latency | throughput |
|---|---|---|
| `BinaryField128` `reference_mul` | 1105.20 ns | 1054.89 ns |
| `BinaryField128` dispatched | 34.95 ns (**31.6×**) | 31.26 ns (**33.7×**) |
| `BinaryField64` `reference_mul` | 354.75 ns | 341.38 ns |
| `BinaryField64` dispatched | 22.15 ns (**16.0×**) | 6.25 ns (**54.6×**) |

You claim ~26×/~28× and ~15×/~36×, so the real numbers are equal or better.

**The 160 KiB of tables cost embedded targets nothing** — this was my main worry given CI builds the crate for `thumbv7em-none-eabi`. The `if HAS_HARDWARE_CLMUL` const-fold does strip them, verified by symbol count and size:

```
aes on : 4 table symbols present, binary 1,108,128 bytes
aes off: 0 table symbols,         binary   942,448 bytes   (delta 165,680 B ~ the documented 160 KiB)
```

Full suite green on **both** paths locally (73 + 126 tests, hardware path and again under `-C target-feature=-aes`), clippy `-D warnings` clean. Two things I suspected and cleared: `PrimeField` is on `Gf2` only (correct — the description reads as if it were on all levels), and `grind`'s `find_any` + per-candidate `clone()` is character-for-character `challenger/src/serializing_challenger.rs:230`, so it's house style rather than sloppiness.

The layered CLMUL proptests deserve a specific call-out as a model: pinning each stage against an independent bit-serial definition (dispatched vs scalar clmul, Karatsuba vs bit-serial 256-bit, reduction vs bit-serial modular mul, then the full mul vs `reference_mul`) is what makes a wrong constant *localizable* rather than "somewhere in the fast path".

One thing to fix before merge (the compile-time claim) plus a few smaller notes inline. Per Thomas's steer I've treated the tower-basis choice as a settled staged decision rather than a defect — but I measured what it costs, since that's the number that'll drive the AES/GHASH call later, and it's inline on `mul_128`.

binary-field/src/tower.rs
```diff
@@ -0,0 +1,906 @@
...
+            fn from_le_byte_iter(mut bytes: impl Iterator<Item = u8>) -> Self {
+                let mut buffer = [0u8; core::mem::size_of::<$repr>()];
+                for byte in &mut buffer {
+                    *byte = bytes.next().expect(BYTE_STREAM_ENDED);
+                }
+                Self::from_le_bytes(buffer)
+            }
+        }
+
+        impl<'de> Deserialize<'de> for $name {
```
tcoratger, line 227: Nit: `#[derive(Serialize)]` on the newtype emits `serialize_newtype_struct`, while this hand-written `Deserialize` reads the backing integer directly. The two agree for every format that treats newtype structs transparently (JSON, bincode, postcard), and the `serde_json` round-trip at line 625 does test it — so this works. It's just easy to break from either side later.

A hand-written `Serialize` mirroring this impl, or `#[serde(transparent)]`, would make the pairing self-evident rather than incidental.

The canonical-range rejection itself is the right call, and the comment explaining *why* (non-canonical re-encoding) is the useful half.

Nashtare:
@tcoratger I addressed your comments, and also bumped the `square` method

```bash
  ┌────────────────────┬──────────┬───────┬───────┐
  │                    │  before  │ after │ x * x │
  ├────────────────────┼──────────┼───────┼───────┤
  │ 64-bit latency     │ 35.14 ns │ 11.62 │ 11.65 │
  ├────────────────────┼──────────┼───────┼───────┤
  │ 64-bit throughput  │ 15.48 ns │ 4.01  │ 4.02  │
  ├────────────────────┼──────────┼───────┼───────┤
  │ 128-bit latency    │ 45.87 ns │ 17.04 │ 19.65 │
  ├────────────────────┼──────────┼───────┼───────┤
  │ 128-bit throughput │ 35.36 ns │ 12.32 │ 17.70 │
  └────────────────────┴──────────┴───────┴───────┘
  ```

tcoratger (approved):

PR #2003, opened 2026-08-26
feat(binary-dft): additive NTT and the Encoder abstraction

Nashtare:
## Summary

Phase 2 of the staged plan to add characteristic-2 (binary tower) field support: an additive NTT over `F_2`-linear subspaces, and the encoder abstraction that lets it reach the multilinear commit path.

Characteristic 2 breaks the multiplicative-subgroup FFT the library is built on — `|GF(2^k)*| = 2^k − 1` is odd, so `TwoAdicField` has no meaning there and every Reed–Solomon path built on it is unavailable. The replacement is an *additive* NTT: evaluation over an `F_2`-linear subspace instead of a multiplicative coset.

**Nothing on the existing two-adic path changes behaviour.** The `Encoder` abstraction is introduced with a blanket impl over every existing `TwoAdicSubgroupDft`, so no call site in `whir`, `multi-stark`, `examples` or any bench moves — `whir/src/pcs/adapter.rs` still passes `&self.dft` unchanged, and the `multi_stark_whir_v1.postcard` fixture pins the encoded bytes as identical.

Based on #2001; review that first.

## Changes

**`Encoder<F>` and a two-adic blanket impl** (`commit/src/encoder.rs`)

A linear code applied to every column of a matrix: `encode_batch(message, log_inv_rate)`. The blanket `impl<F: TwoAdicField, D: TwoAdicSubgroupDft<F>> Encoder<F> for D` zero-pads and calls `dft_batch`, which is exactly what the commit path did inline, so the existing path is unchanged by construction. `p3-commit` gains a hard dependency on `p3-dft` (no cycle — `p3-dft` depends only on field/matrix/maybe-rayon/util).

**Route the multilinear commit path through `Encoder`** (`sumcheck/src/commit.rs`, `sumcheck/src/layout/prover/`)

`commit_base` and `Layout::commit` take an `E: Encoder<F>` instead of a `Dft: TwoAdicSubgroupDft<F>`, and the now-unused `TwoAdicField` bounds on `Layout`, `PrefixProver` and `SuffixProver` relax to `Field`. The `TwoAdicField` bounds in `sumcheck/src/zk/` stay — the HVZK path is excluded from characteristic 2 by the protocol it implements, not by those bounds. `p3-dft` moves to `p3-sumcheck`'s dev-dependencies, since its only remaining use there is `#[cfg(test)]`.

**Cantor basis of the tower** (`binary-field/src/cantor.rs`, `binary-field/src/tower.rs`)

`v_0 = 1` and `v_i² + v_i = v_{i−1}`, taking the root whose lowest bit is clear. With this basis the normalised subspace polynomials are the unnormalised ones (`Ŵ_j = W_j`) and `W_j(v_i) = v_{i−j}`, so NTT twiddles are index shifts — **no twiddle table, no per-size precomputation, no cache**. `TowerLevel` becomes `pub` and gains `cantor_basis(i)`, backed by a pinned 128-entry constant with an `#[ignore]`d Artin–Schreier regenerator committed alongside as its audit trail, in the same spirit as `search_generators` for `GENERATOR`.

**New crate `p3-binary-dft`** (`binary-dft/`)

- `domain.rs` — `domain_point(index)` and `subspace_polynomial(j, x)`, the two pure functions the transform, the encoder, and a future verifier all build on.
- `traits.rs` — `AdditiveNtt<F>`: two required methods (`shifted_ntt_batch`, `shifted_intt_batch`) and four defaults.
- `naive.rs` — `NaiveAdditiveNtt`, a deliberately quadratic reference that evaluates `Σ d_i · X_i(x)` directly, computing each `Ŵ_j` from the **product** definition including an explicit normalisation division. It depends on none of the identities above, which is what makes it a genuine oracle rather than a restatement.
- `lch.rs` — `LchNtt`, the Lin–Chung–Han transform. Parallel over butterfly blocks behind the crate's `parallel` feature.
- `encoder.rs` — `AdditiveRsEncoder`, implementing `Encoder<BinaryField128>`.

**Parallelise within butterfly blocks** (`binary-dft/src/lch.rs`)

Parallelism was over butterfly blocks alone, and stage `j` has only `2^(ℓ−1−j)` of them — so stage `ℓ−1` ran on a single core on any machine while stage `0` produced `2^(ℓ−1)` chunks of `2·width` elements. The inner butterfly pairs `lo[k]` with `hi[k]` independently, so it now also splits within a block at a fixed grain. That makes the per-stage task count `n·width / (2·GRAIN)` — independent of `j` — which is precisely the property the block-only scheme lacked. **1.89×–1.94× on `ntt/128`** on a 14-core machine; output verified bit-identical.

**Skip the multiply when the twiddle is zero** (`binary-dft/src/lch.rs`)

`W_j` is `F_2`-linear and `domain_point(0)` is zero, so over the subspace itself the *first block of every stage* — not only stage `ℓ−1` — has twiddle zero, and its butterfly collapses to `(u, v) ↦ (u, u + v)`: no multiply, and no store to `u`. That is `2/ℓ` of all butterflies, and at 128 bits the multiply is nearly all of one.

**`GF(2^128)` in the polynomial basis, and a transform that stays there** (`binary-field/src/poly_basis.rs`, `binary-dft/src/poly.rs`)

This is the basis-conversion lever the previous revision of this description listed as out of scope; it is now in. `p3-binary-field` gains a small public `poly_basis` module — the two conversions, the basis-native product and square, and `HAS_HARDWARE_CLMUL` — and `p3-binary-dft` gains `PolyBasisNtt`, which converts the matrix once on the way in, runs every stage with a bare carryless multiply and reduction, and converts back on the way out. That is `2` conversions per element instead of `3ℓ/2`. Addition is `XOR` in both bases, so the sums are untouched.

Deliberately **not** a field type: the two bases are indistinguishable once wrapped, and a value that reached ordinary `BinaryField128` arithmetic in the wrong one would be silently wrong rather than ill-typed, so an element in this basis is carried as the `u128` of its coordinates. Without a carryless-multiply instruction the polynomial-basis product is the bit-serial fallback and *slower* than the tower arithmetic it replaces, so on such a target `PolyBasisNtt` runs the tower transform instead; the choice is a constant and only one arm survives compilation. `AdditiveRsEncoder` defaults to it for that reason.

**Benchmarks, and one measured negative result** (`binary-dft/benches/ntt.rs`)

See below. Includes a harness correction: `BatchSize::LargeInput` sizes the pre-allocated batch from the iteration count, so a *faster* implementation gets a larger batch and more memory pressure — the benchmark penalised what it was meant to reward. All three sites now use `BatchSize::PerIteration`.

**Allocate the codeword buffer once when padding** (`commit/src/encoder.rs`, `binary-dft/src/encoder.rs`)

A regression the `Encoder` refactor would otherwise have introduced on the *existing* prover path: `transpose()` allocates at exact capacity and `Vec::resize` then reallocates and memcpies, and for `MontyField31` it loses the calloc fast path (`zero_vec` is `vec![0u32; len]`, lazily-zeroed pages; `resize` writes `3n` zeros element-wise). Both `encode_batch` impls now allocate once with `zero_vec` and `copy_from_slice`, with a `checked_shl` guard so an absurd rate panics rather than silently producing an under-sized codeword.

## Testing

The transform is not something anyone can eyeball, so the approach is reference-first:

- **`LchNtt` proven equal to `NaiveAdditiveNtt` by proptest** at every byte-aligned level (8/16/32/64/128), with random cosets and widths 1–5, plus round-trip, linearity, LDE-prefix and cross-level agreement. The two derive their twiddles by different routes — the oracle from the product definition, the transform from the `x² + x` recurrence — so agreement is real evidence.
- **`PolyBasisNtt` proven equal to `NaiveAdditiveNtt` and to `LchNtt` by proptest**, over random cosets and widths, at heights covering both parities of the stage count — the second comparison is what pins the polynomial-basis transform against the tower one it replaces, independently of the oracle.
- **The oracle's normalisation divisor came out as exactly 1 at every level**, empirically confirming the `Ŵ_j = W_j` identity rather than assuming it.
- **End-to-end commit over `BinaryField128`**: `commit_base`'s Merkle root reproduces a matrix encoded by hand with the reference transform (pinning the message layout as well as the transform), and `PrefixProver::commit` — the entry point WHIR itself calls — yields identical roots through `LchNtt` and through the oracle.
- **Cantor table**: recurrence checked in each level's own arithmetic, subfield nesting, `F_2` independence, and the even-representation normalisation, all in CI-run tests; the `#[ignore]`d regenerator asserts before printing, so a wrong solver cannot emit a wrong table.
- `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt --all -- --check`, and `cargo build --target thumbv7em-none-eabi -p p3-binary-dft`.
- The `multi_stark_whir_v1.postcard` compat fixture passes unchanged.

## Benchmarks

Two machines: an aarch64 14-core laptop, and an x86_64 `c8i.32xlarge` (128 vCPU, AVX-512, `VPCLMULQDQ`). **Trust the ratios, not the absolutes.** Single runs, 10 samples, criterion flagged noise on several points.

**Two measurement corrections landed first, because everything below depends on them.**

*The harness was biased against improvements.* `BatchSize::LargeInput` sizes criterion's pre-allocated batch from the iteration count, so a faster implementation runs more iterations, receives a larger batch, and is charged for the extra memory pressure. It reported **−6.5%** for a change that is genuinely **+89%**. With `BatchSize::PerIteration` the default configuration now agrees with direct in-process timing: `ntt/128/20` 600.54 → **331.96 ms** (in-process 324.0 ms), `ntt/128/18` 143.44 → **75.92 ms** (in-process 77.61 ms).

*x86 needs `-C target-feature=+pclmulqdq` explicitly.* `aarch64-apple-darwin` carries `aes` in its baseline; no x86_64 target carries `pclmulqdq`. Without it `HAS_HARDWARE_CLMUL` is false and `dispatched_mul` falls back to the recursive tower multiplication — **~37× slower** (349.94 ms against 9.48 ms on `ntt/128/14`). Note also that `+pclmulqdq` alone accelerates *only* the binary path: BabyBear's packing is gated on `avx2`/`avx512f` (`baby-bear/src/lib.rs:25-38`), so any comparison against it needs `-C target-cpu=native`.

### Parallelising within butterfly blocks

`par_chunks_mut` over butterfly blocks gave stage `j` only `2^(ℓ−1−j)` of them, so the top `log₂(T)` stages were thread-starved and stage `ℓ−1` ran single-threaded on any machine. Splitting within blocks makes the per-stage task count `n·width / (2·GRAIN)`, independent of `j`.

| | 14-core aarch64 | 128-vCPU x86_64 |
|---|---|---|
| `ntt/128/16` | — | **5.07×** |
| `ntt/128/18` | **1.94×** | **5.36×** |
| `ntt/128/20` | **1.89×** | **5.20×** |
| `encode/BinaryField128/20` | — | **2.70×** |

`encode/BabyBear` reports "No change in performance detected" at every size (p = 0.43–0.67) — a control confirming the change touches only the binary path. Output was separately verified bit-identical.

Scaling is improved but not solved. From 8 to 128 threads (16×), `ntt/128` now gains 1.52× / 2.56× / 3.77× / **4.81×** at `ℓ` 14/16/18/20, against ~1.2×–1.55× across the board before. Efficiency at the largest size went from ~9% to ~30%. `ntt/64/14` now *regresses* 13% at 128 threads — small problems over-subscribe.

### Against the two-adic path, at equal byte volume

With both sides accelerated (`-C target-cpu=native`), 128 threads:

| | `BinaryField128` | `BabyBear` | ratio |
|---|---|---|---|
| `encode/…/14` | 5.61 ms | 2.18 ms | 2.58× |
| `encode/…/16` | 22.02 ms | 23.69 ms | **0.93×** |
| `encode/…/18` | 90.4 ms | 57.18 ms | 1.58× |
| `encode/…/20` | 487.4 ms | 254.4 ms | 1.92× |

An earlier revision of this description quoted 10.8×–16.1×, measured on the laptop with the biased harness and before the granularity fix. **That figure was dominated by the parallelisation defect, not by binary-field arithmetic.** The two paths also scale differently — 8→128 threads gains 4.81× for the binary path against 1.80× for BabyBear's `encode`, and AVX-512 bought BabyBear only 8.5%–23% — which is consistent with BabyBear being memory-bound at equal byte volume (4× the rows, ~0.3 ns multiply) while the binary path is compute-bound at ~20 ns per multiply and therefore converts cores into throughput. The `ℓ=16` row where we come out ahead is close enough to be a threading artifact.

### The zero twiddle, and the change of basis

Both measured **single-threaded**, on the aarch64 laptop, comparing the two forms of the same transform directly. Parallel wall-clock on that machine swings by ±30% run to run — a single A/B pair reported −57% and its repeat +31% for identical code — so the numbers below are the ones with ±1% intervals rather than the ones with the larger headline.

Skipping the zero twiddle, against the same transform without the special case:

| | before | after | |
|---|---|---|---|
| `ntt/128/14` | 38.54 ms | 32.43 ms | **−16.9%** |
| `ntt/128/16` | 175.88 ms | 151.76 ms | **−13.3%** |
| `ntt/128/18` | 779.93 ms | 685.47 ms | **−12.1%** |

tracking `2/ℓ` as the stage count grows, which is the fraction of butterflies the identity covers.

Running in the polynomial basis, against the same transform in the tower basis:

| | tower | polynomial | |
|---|---|---|---|
| `encode/BinaryField128/14` | 64.24 ms | 12.11 ms | **5.31×** |
| `encode/BinaryField128/16` | 299.82 ms | 54.61 ms | **5.49×** |
| `encode/BinaryField128/18` | 1.372 s | 231.2 ms | **5.93×** |

The ratio grows with the height because the two boundary conversions are fixed while the products they replace scale with the stage count. The earlier estimate of 8.36×–8.59× came from patching the conversions out of `mul_128` altogether, which prices the boundary at zero; 5.3×–5.9× is what remains once it is paid for real.

### Two-layer stage fusion: implemented, measured, not kept

Worth recording because the *reason* it loses moved during this PR. Before the change of basis the transform was flat at ~21 ns per butterfly from 4 MiB to 64 MiB of data — entirely compute-bound, so halving the passes over memory could not have paid. After it, a butterfly is ~2.9 ns and a thread sweep saturates: 1→2 threads 1.93×, 2→4 1.89×, 4→8 1.72×, 8→10 **1.06×**, 10→14 0.96×. That is the memory wall, so fusion became worth building.

A radix-4 kernel applying two stages to each quadruple of rows in one pass (both parities of `ℓ`, proptested against the oracle *and* against the unfused transform) measures **−18% serial** against **+6.5% to +9.7% parallel at `ℓ ≥ 18`, and nothing at `ℓ = 16`**. The serial loss is register pressure and dependent multiplies inside a quadruple, where the radix-2 loop has none. A bad trade for a default, so it is not in this PR — but it attacks the constraint that now actually binds, and on a machine with far more cores per unit of bandwidth than this laptop it should look different. Worth re-measuring there before deciding.

Software prefetch was **not** built: the streams are sequential, so the hardware prefetcher already covers them, and the lookups it would need to hide are the data-dependent table lookups that the change of basis has just removed from the inner loop.

### The reverted subfield butterfly, and what it means

**The `GF(2^128) × GF(2^32)` subfield-scaling butterfly was implemented, measured, and reverted — but the direction is right and the primitive is what is missing.** With `shift = 0` the twiddle is `domain_point(blk << 1)`, and the Cantor nesting puts `v_i` inside `2^⌈log₂(i+1)⌉` bits, so every twiddle of every practical transform lies in `GF(2^32)` — narrower in the early stages, identically zero at stage `ℓ−1`. Scaling coordinate-wise is therefore the structurally correct butterfly, and is what the tower representation exists to make cheap. It loses *today* only because `GF(2^32)` has no fast multiply: one `reference_mul` costs 53.5 ns against 20.4 ns for a whole `GF(2^128)` CLMUL, predicting ~10.5× and measuring 9.17×–10.91× (old harness; direction sound, magnitude conservative). **Since this was measured, #2001 gained `GF(2^8)` log/exp tables and `GF(2^32)` multiplication is now 8.2× faster in throughput** (51.06 → 6.23 ns), which narrows but does not close the gap: four `GF(2^32)` coordinate multiplies come to ~24.9 ns against ~16.8 ns for `mul_128`. The interesting consequence is that twiddle width is *stage-dependent* — at stage `j` the twiddle uses basis vectors up to index `ℓ−j−1`, so the top eight stages fit in `GF(2^8)`, where sixteen coordinate multiplies come to ~5.3 ns and win outright. A level-adaptive butterfly is therefore plausible now, at an estimated ~1.35×–1.5× overall, and is unmeasured. A SIMD small-field multiply (GFNI, `PSHUFB` tables) would go considerably further; building one is its own project.

### Remaining levers

- **Basis conversion — landed in this PR**, see above. The reasoning that justified it stands and is worth keeping on the record: `mul_128` spends **88.1% of its throughput** (69.0% of latency) in three table-driven conversions, and the two 128-bit tables are 128 KiB, exactly one L1D. **The tower must remain the storage representation of the *public* types, but not of the PCS interior.** `Challenge: ExtensionField<Val>` pulls in `BasedVectorSpace`, whose `as_basis_coefficients_slice` returns a *borrow* that only a bit-range embedding can satisfy (`binary-field/src/extension.rs:73` reinterprets in place) — so no polynomial-basis type can implement it, and that half is closed. The interior is not: ring switching (eprint 2024/504, Constructions 3.1 and 5.1) needs only computable pack/decompose over *any* basis with `β_0 = 1`, in bulk passes, never a borrowed view of committed data — so the encoder alphabet and the transform are free to use whatever is fastest. That is why the conversion sits at the NTT boundary rather than inside the field type. (binius64 sidesteps the question entirely by never building a tower above `GF(2^8)`: its only 128-bit type is the polynomial-basis `Ghash128b`. That option is closed here, since `BasedVectorSpace` needs the borrow.) What remains after this change is a transform bound by memory rather than by arithmetic, which is a different problem — see the fusion result above.
- **Packing.** `BinaryField128::Packing = Self`. This is **not** the `PackedValue` redesign the proposal defers — that concerns *sub-byte* elements, where `[Gf2; 8]` bit-packed is not castable to `[Value; WIDTH]`. `BinaryField128` is `#[repr(transparent)]` over `u128`, so `WIDTH = 2`/`4` satisfies the layout contract today. Amdahl caps it at **+6% (AVX2) / +10% (AVX-512)** against the 11.9% of `mul_128` that is ALU, and on aarch64 it is worth exactly zero (`WIDTH = 1`; `PMULL` has no `VPCLMULQDQ` analogue). After the conversions go it is worth 2×–4×, gated on `VPCLMULQDQ`.

Packing is out of scope here.

## Notes for reviewers

- **One concern per commit**, tests landing with the code they test.
- **The blanket impl forecloses field-generic downstream encoders.** `impl<F: TwoAdicField, D: TwoAdicSubgroupDft<F>> Encoder<F> for D` means no crate can write `impl<F: Field> Encoder<F> for MyEncoder` — that is E0119, since a downstream crate could implement `TwoAdicSubgroupDft` for it. Every future encoder must name concrete fields, which is why `AdditiveRsEncoder` implements `Encoder<BinaryField128>` rather than being generic. This is documented on the trait itself. The alternative — a newtype adapter in `p3-commit` — would move exactly the call sites this design exists to keep still.
- **Deliberately not done here**: sub-byte bit-packing and the `PackedValue` redesign it needs; the lookup argument for characteristic 2; the binary PCS. Each is scoped for a later phase.
- **The tower-vs-polynomial basis question raised on #2001 is settled here, and not the way that note predicted.** It does not arrive for free through `TowerLevel`'s `Mul`: a product in the tower basis pays all three conversions however it is spelled, so a faster `Mul` would still convert `3ℓ/2` times per element. The win needs the *data* to stay in the polynomial basis across stages, which is a change in this crate — hence `PolyBasisNtt` rather than a change to `BinaryField128`.


binary-dft/src/poly.rs
```diff
@@ -0,0 +1,202 @@
...
+            return self.tower.shifted_ntt_batch(mat, shift);
+        }
+
+        let width = mat.width();
+        let log_n = log2_strict_usize(mat.height());
+        let mut values: Vec<u128> = mat
+            .values
+            .par_iter()
+            .map(|&x| poly_basis::from_tower(x))
+            .collect();
```
tcoratger: This is the one I'd most like to see changed before merge: the transform allocates a **second full codeword** and keeps it alive alongside `mat.values` for the whole transform.

I measured it on a `2^20 × 16` message at `log_inv_rate = 1` (512 MiB codeword), reading `VmHWM` before and after:

```
poly: peak RSS 1024 MiB   (codeword 512 MiB)   <- 2.0x
lch : peak RSS  769 MiB   (codeword 512 MiB)   <- 1.5x
```

Since this is the default encoder (`AdditiveRsEncoder<F, Ntt = PolyBasisNtt>`), that's a 33% peak-RSS bump on the largest allocation the prover makes.

The good news is it's completely avoidable — `BinaryField128` is `#[repr(transparent)]` over `u128`, so `Vec<BinaryField128> -> Vec<u128>` hits std's in-place-collect specialisation and reuses the buffer. I checked:

```
before: RSS 514 MiB
after in-place map/collect: RSS 514 MiB, same buffer = true
```

So something like:

```rust
// Same size and alignment, so this reuses mat.values' allocation.
let mut values: Vec<u128> = core::mem::take(&mut mat.values)
    .into_iter()
    .map(TowerLevel::to_repr)
    .collect();
values
    .par_iter_mut()
    .for_each(|v| *v = poly_basis::from_tower(BinaryField128::from_repr(*v)));

// ... stages ...

mat.values = values.into_iter().map(poly_basis::to_tower).collect();
```

keeps the conversion parallel *and* drops the peak back to 769 MiB. (The `into_iter().map(to_repr).collect()` step is serial, but it's a pointer walk over a buffer it doesn't move — the actual basis change stays on rayon.)

binary-dft/src/encoder.rs
```diff
@@ -0,0 +1,93 @@
...
+            let len = message.values.len();
+            let padded_len = u32::try_from(log_inv_rate)
+                .ok()
+                .and_then(|rate| len.checked_shl(rate))
+                .expect("log_inv_rate must be smaller than the width of usize");
+            let mut values = BinaryField128::zero_vec(padded_len);
+            values[..len].copy_from_slice(&message.values);
+            message.values = values;
+        }
+        self.ntt.ntt_batch(message)
```
tcoratger: There's a free ~`r/(k+r)` here that I think is worth taking, and it also happens to help with the memory wall you identify as the next binding constraint.

Zero-padding to `2^ℓ` and running the full height-`2^ℓ` transform means that at every stage `j ≥ k` the entire `hi` coefficient half is zero, so the butterfly degenerates to a pure replication:

```
u' = u + t·0 = u
v' = u + (t+1)·0 = u
```

But `*u += t * *v` still runs — a full 128-bit CLMUL against zero — for all `2^(ℓ−1)·width` pairs of each of those `r` stages. The `if t.is_zero()` fast path doesn't catch it, because it tests the *twiddle*, not the operand.

The equivalent form is `2^r` **independent** shifted NTTs of height `2^k`, with shifts `shift + domain_point(c << k)` — i.e. evaluate `f` on each coset of `S_k` in `S_ℓ` separately. I implemented it and checked the identity holds:

```
coset-wise encode == zero-pad-and-transform: OK    (k=4/r=2, 6/1, 5/3, 3/0; widths 1..4)
k=20 r=2: zero-pad 1.497 s   coset-wise 1.367 s   (1.09x)
```

1.09× lines up almost exactly with the predicted `ℓ/k = 22/20 = 1.10`, and my throwaway version still clones the message per coset, so a real one writing straight into the output buffer should do a bit better. The structural win is probably the bigger one though: each coset becomes an independent `2^k · width` block that can stay cache-resident, instead of `r` extra full sweeps over the expanded codeword.

Happy to send this as a follow-up patch if you'd rather keep this PR focused.

binary-dft/src/encoder.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// Reed–Solomon over the additive NTT domain.
+///
+/// The message holds the low-index novel-basis coefficients of each column, so the codeword is
+/// the evaluation of `f̂(Ŵ_0(x), …, Ŵ_{k−1}(x))` on `S_{k + log_inv_rate}`.
+///
+/// The alphabet is `BinaryField128`, where [`PolyBasisNtt`] is the faster transform and falls
+/// back to [`LchNtt`](crate::LchNtt) on a target without a carryless multiply, so it is the default.
+#[derive(Clone, Debug, Default)]
+pub struct AdditiveRsEncoder<F, Ntt = PolyBasisNtt> {
```
tcoratger, line 28: Tiny thing: `F` is phantom here — the only inhabited instantiation is `BinaryField128`, so `AdditiveRsEncoder::<BinaryField32>::default()` compiles happily and then has no `Encoder` impl. It also makes the use sites read a little oddly (`AdditiveRsEncoder::<F, LchNtt<F>>`).

Totally defensible if the plan is one impl per concrete field later — which the blanket impl does force — but since that's the *only* way to add fields, maybe worth one line of doc saying so, so the next reader knows the parameter is load-bearing for a future they can't see yet.

sumcheck/src/commit.rs
```diff
@@ -14,71 +13,155 @@ use crate::strategy::VariableOrder;
...
-        }
+    let width = 1 << folding;
+
+    let message = match order {
+        VariableOrder::Prefix => info_span!("transpose").in_scope(|| {
+            // Transposing the folding blocks turns the first folded variables into columns.
+            RowMajorMatrixView::new(poly.as_slice(), 1 << (num_variables - folding)).transpose()
+        }),
+        // Folding blocks are already contiguous, so the row width alone selects them.
+        VariableOrder::Suffix => RowMajorMatrix::new(poly.as_slice().to_vec(), width),
```
tcoratger: I think only half of the allocation regression the description mentions actually got fixed — the `Vec::resize` one inside `encode_batch` is gone, but both branches here still add a full-message allocation and memcpy on the *existing* BabyBear/WHIR path.

Counting allocations of `n = 2^num_variables` against the codeword `n << r`:

| | before | after |
|---|---|---|
| prefix | 1 alloc `n<<r`, `transpose_into` writes `n` | 1 alloc `n` + transpose, **then** 1 alloc `n<<r` + memcpy `n` |
| suffix | 1 alloc `n<<r`, 1 copy `n` | 1 alloc `n` + copy `n`, **then** 1 alloc `n<<r` + copy `n` |

So every commit pays one extra `n`-element allocation and one extra `n`-element memcpy, and peaks `n` higher. At `num_variables = 24` over BabyBear that's a spurious 64 MiB alloc + memcpy per commit; the suffix branch (`poly.as_slice().to_vec()`) copies the polynomial twice to move it zero positions.

The cause is structural rather than sloppy: `encode_batch(message, log_inv_rate)` takes the message at *message* height, so the caller can never write straight into the codeword buffer. Two ways out:

**Minimal** — `commit_base` allocates `F::zero_vec(n << r)` itself, `transpose_into` / `copy_from_slice` into the prefix exactly as before, and calls `encoder.encode_batch(padded, 0)`. Restores the old cost precisely, at the price of one doc sentence saying `log_inv_rate = 0` means "already padded".

**Principled** — change the contract so the encoder receives the buffer at *codeword* height plus the message height. The caller allocates once and fills the prefix; the encoder learns both `k` and `r`, which is exactly what it needs to skip transforming the zeros (see my note on `AdditiveRsEncoder::encode_batch`). That kills both findings with one change, which is why I'd lean this way.

Easy to confirm either way with a counting global allocator around `commit_base` for both `VariableOrder`s.

sumcheck/src/commit.rs
```diff
@@ -14,71 +13,155 @@ use crate::strategy::VariableOrder;
...
     };
```
tcoratger, line 59: Very minor observability thing: this now reports the *pre-expansion* height, where the old `info_span!("dft", ...)` reported the codeword height. Anyone comparing flamegraphs across the change silently loses a factor of `2^r` on the labelled height.

`height = message.height() << starting_log_inv_rate` would keep the old meaning while the span name change stays.

commit/Cargo.toml
```diff
@@ -11,9 +11,10 @@ categories.workspace = true
 
 [features]
 default = []
-test-utils = ["p3-challenger", "p3-dft"]
+test-utils = ["p3-challenger"]
 
 [dependencies]
+p3-dft.workspace = true
```
tcoratger, line 17: Worth a moment's thought on direction: `p3-commit` is the trait crate (`Mmcs`, `Pcs`, `PolynomialSpace`), and it's depended on by `p3-merkle-tree`, `p3-keccak-air`, `p3-sha256-air`, `p3-goldilocks`, `p3-poseidon2-air` — none of which have any use for a DFT. Hosting the blanket impl here inverts the arrow.

I checked and it can go the other way with no cycle: `p3-commit`'s deps are `p3-field`, `p3-matrix`, `p3-multilinear-util`, `p3-util`, and `p3-multilinear-util` does **not** depend on `p3-dft`. So:

- keep `trait Encoder` in `p3-commit`, dependency-free
- move `impl<F: TwoAdicField, D: TwoAdicSubgroupDft<F>> Encoder<F> for D` into `p3-dft`
- add `p3-commit` to `p3-dft`'s dependencies

No call site moves, and the fixture stays put. Nice property given the whole design goal here was to not move call sites.

tcoratger (commented):
Really nice piece of work — I went through this one carefully (re-derived the LCH twiddle identity and the Cantor recurrence from scratch rather than taking the doc comments on faith) and the mathematics is correct throughout. A few things I want to call out as genuinely well done:

- **The oracle is a real oracle.** `NaiveAdditiveNtt` taking `Ŵ_j` from the product definition *with* an explicit normalisation division, while `LchNtt` takes it from the `x² + x` recurrence, means the proptest also pins the Cantor table itself — a wrong table would break `Ŵ_j = W_j` and the two would disagree. That is a much stronger test than it looks at first glance.
- I cross-checked the construction against binius64's `crates/math/src/ntt/domain_context.rs`, which implements the general `W_i(X) = W_{i−1}(X)·(W_{i−1}(X) + W_{i−1}(β_{i−1}))` with an explicit normalising inverse. This PR is exactly that recurrence specialised to `W_{i−1}(β_{i−1}) = 1`, so the "no twiddle table, no precomputation" claim really does hold.
- The `as $repr` truncation in `cantor_basis` is safe and I convinced myself of why: `i < $bits = 2^k` forces `(i+1).next_power_of_two() ≤ 2^k`, and `check_nesting` pins `v_i < 2^(2^k)`. Worth keeping that reasoning in the comment.
- The existing two-adic path really is untouched — `whir/src/pcs/adapter.rs:66` still bounds `Dft: TwoAdicSubgroupDft<F>`, so a binary encoder cannot reach WHIR by accident.

On my machine: `cargo test` green for `p3-binary-dft`, `p3-binary-field`, `p3-commit`, `p3-sumcheck`, `p3-whir`, `p3-multi-stark` (also with `--features parallel`), and `cargo clippy --workspace --all-targets` clean.

**No blockers.** What I did find is three performance things, all local and all with concrete fixes — I've left them inline with the measurements I took. The short version:

1. `PolyBasisNtt` keeps two full codeword buffers alive. Measured **1024 MiB peak vs 769 MiB** for `LchNtt` on the same `2^20 × 16` encode. Fixable to zero extra bytes.
2. `AdditiveRsEncoder` transforms the zero padding: `r/(k+r)` of its butterflies are multiplies against zero. Verified the coset-wise identity and measured 1.09× at `k=20, r=2`, matching the predicted `ℓ/k = 22/20`.
3. `commit_base` picked up an extra full-message allocation + memcpy on the *existing* BabyBear path, in both branches — I think only half of the regression the description mentions actually got fixed.

Also one architectural nit worth a thought: `p3-commit` now hard-depends on `p3-dft`, which drags a DFT crate into `p3-merkle-tree`'s closure. Details inline.

Nothing here should hold the PR up long — mostly it's "this is right, and here is how to make it cheaper". Happy to pair on any of it.

commit/src/encoder.rs
```diff
@@ -0,0 +1,89 @@
...
+        mut message: RowMajorMatrix<F>,
+        log_inv_rate: usize,
+    ) -> RowMajorMatrix<F> {
+        if log_inv_rate > 0 {
+            // Appending zero rows extends every column's coefficient vector.
+            let len = message.values.len();
+            let padded_len = u32::try_from(log_inv_rate)
+                .ok()
+                .and_then(|rate| len.checked_shl(rate))
+                .expect("log_inv_rate must be smaller than the width of usize");
```
tcoratger: Small correctness-of-the-guard thing: `usize::checked_shl` only returns `None` when the *shift amount* is `≥ usize::BITS`. It doesn't detect that the value overflowed — `(1usize << 63).checked_shl(1)` is `Some(0)`, not `None`.

So the description's "an absurd rate panics rather than silently producing an under-sized codeword" is true, but only by accident: `padded_len` truncates to `0` and the panic actually comes from the next line, `values[..len]` indexing an empty vec, with a message about something else entirely.

Stating the real invariant is nicer to whoever hits it:

```rust
let padded_len = len
    .checked_shl(u32::try_from(log_inv_rate).expect("log_inv_rate must fit in u32"))
    .filter(|&padded| padded >= len)
    .expect("codeword length overflows usize");
```

Also: the doc promises a panic for `log_inv_rate ≥ usize::BITS` and there's no `#[should_panic]` test for it in either impl — one line each.

binary-dft/src/lch.rs
```diff
@@ -0,0 +1,273 @@
...
+                                    *v += *u;
+                                }
+                            } else {
+                                for (u, v) in lo.iter_mut().zip(hi) {
+                                    // (u, v) ↦ (u + t·v, u + t·v + v)
+                                    *u += t * *v;
+                                    *v += *u;
+                                }
+                            }
+                        });
```
tcoratger: Two small things in here, both cheap.

**Hoist the branch.** `t` is invariant across the whole block, so testing it per grain chunk costs a (well-predicted, but non-free) branch and stops the two loop bodies being specialised once.

**Skip the nested split when it can't pay.** At low stages `half = 2^j · width` is far below `BUTTERFLY_GRAIN = 1024` — at `j = 0, width = 16` it's 16 — so this constructs a rayon `Zip` producer for a *single* chunk, `n/2` times per stage. That plumbing isn't free, and the doc at line 34-36 already promises it is ("Stages with `half ≤ BUTTERFLY_GRAIN` keep a single piece per side and pay nothing for the extra level"). This would make the comment true:

```rust
let zero = t.is_zero();
if half <= BUTTERFLY_GRAIN {
    butterfly(lo, hi, t, zero);          // serial, no nested split
} else {
    lo.par_chunks_mut(BUTTERFLY_GRAIN)
        .zip(hi.par_chunks_mut(BUTTERFLY_GRAIN))
        .for_each(|(lo, hi)| butterfly(lo, hi, t, zero));
}
```

(Same shape applies verbatim in `poly.rs`.)

By the way — I checked the butterfly itself carefully and it's right in both directions, including that the zero-twiddle collapse is valid for *any* `t = 0`, not only `shift = 0`. Nice that the special case falls out of the general form rather than being a `shift == 0` test.

binary-dft/src/lch.rs
```diff
@@ -0,0 +1,273 @@
...
+/// Stage `j` has `2^(ℓ − 1 − j)` blocks of `half = 2^j · width` elements per side, so cutting
+/// each side into pieces of this size leaves `n · width / (2 · BUTTERFLY_GRAIN)` pieces at
+/// every stage, independent of `j`: the wide stages, which have too few blocks to fill a
+/// machine, are split from within instead. Stages with `half ≤ BUTTERFLY_GRAIN` keep a single
+/// piece per side and pay nothing for the extra level.
+///
+/// At roughly twenty nanoseconds per `GF(2^128)` butterfly a piece of this size is tens of
+/// microseconds of work, orders of magnitude above the cost of handing a task to another
+/// thread, while still leaving hundreds of pieces per stage at the smallest useful heights.
+pub(crate) const BUTTERFLY_GRAIN: usize = 1 << 10;
```
tcoratger, line 40: Question rather than a finding: the justification here is "roughly twenty nanoseconds per `GF(2^128)` butterfly", but after the change of basis the PR body measures ~2.9 ns. Was `1 << 10` re-tuned once `PolyBasisNtt` landed, or is it still the pre-basis-change number?

At 2.9 ns a chunk is ~3 µs rather than the "tens of microseconds" the comment claims — still comfortably above task-spawn cost, so I don't think it's wrong, but the comment's arithmetic no longer matches the code it justifies.

binary-dft/src/lch.rs
```diff
@@ -0,0 +1,273 @@
...
+            width in 1usize..=3,
+            seed in any::<u64>(),
+            shift in any::<u64>(),
+        ) {
+            let coeffs = matrix::<BinaryField16>(log_n, width, seed);
+            let shift = sample::<BinaryField16>(shift);
+            let ntt = LchNtt::<BinaryField16>::default();
+            let evals = ntt.shifted_ntt_batch(coeffs.clone(), shift);
+            prop_assert_eq!(ntt.shifted_intt_batch(evals, shift), coeffs);
+        }
```
tcoratger, line 240: Coverage gap worth closing while you're here: `lde_batch` / `shifted_lde_batch` are never tested for `LchNtt` or `PolyBasisNtt` — only `NaiveAdditiveNtt` has `lde_extends_the_input`, at a single size with `shift = 0`.

I ran the missing cases against the oracle myself and **they all pass**, including the "input rows reappear as the prefix" property with a nonzero shift, so this is genuinely coverage rather than a lurking bug. But since Phase 3 is going to fold along exactly that prefix correspondence, it seems worth pinning now:

```rust
#[test]
fn lch_lde_matches_naive(
    log_n in 0usize..=6,
    added in 0usize..=3,
    width in 1usize..=3,
    seed in any::<u64>(),
    shift in any::<u64>(),
) {
    let m = matrix::<BinaryField16>(log_n, width, seed);
    let s = sample::<BinaryField16>(shift);
    let lde = LchNtt::default().shifted_lde_batch(m.clone(), added, s);
    prop_assert_eq!(&lde, &NaiveAdditiveNtt::default().shifted_lde_batch(m.clone(), added, s));
    // The correspondence Phase 3 folds along, and it holds on a coset too.
    prop_assert_eq!(&lde.values[..m.values.len()], &m.values[..]);
}
```

Same for `PolyBasisNtt`. Also missing: `#[should_panic]` cover for `ℓ > 128` and for a non-power-of-two height, both of which are documented panics on public API.

binary-dft/src/traits.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// An additive NTT: evaluation of the novel polynomial basis on an `F_2`-linear subspace.
+///
+/// Columns hold coefficients in the novel polynomial basis `X_i = ∏_j Ŵ_j^{bit_j(i)}`, the
+/// products of normalised subspace polynomials; output row `i` is the evaluation at
+/// `shift + domain_point(i)`. Both sides are in natural order.
+///
+/// Heights are powers of two, and `ℓ = log2(height)` is at most the bit width `2^LOG_BITS` of
+/// `F`: `S_ℓ` is spanned by the first `ℓ` Cantor basis vectors, of which there are only that many.
+pub trait AdditiveNtt<F: TowerLevel>: Clone + Default {
```
tcoratger: `Clone + Default` as supertraits rules out any implementation that needs state — a precomputed twiddle table, a configurable grain, or a non-Cantor subspace basis. That last one is exactly the shape binius64's `GenericPreExpanded` takes, so it's not hypothetical.

The Cantor basis happening to need no precomputation is the *lovely* property of this design, but the trait is the general interface, and I think these bounds are only here so `AdditiveRsEncoder` can derive `Clone`/`Default`. Putting them on the struct instead keeps the door open:

```rust
pub trait AdditiveNtt<F: TowerLevel> { ... }

#[derive(Clone, Debug, Default)]
pub struct AdditiveRsEncoder<Ntt: Clone + Default = PolyBasisNtt> { ... }
```

binary-dft/src/traits.rs
```diff
@@ -0,0 +1,64 @@
...
+        mat: RowMajorMatrix<F>,
+        added_bits: usize,
+        shift: F,
+    ) -> RowMajorMatrix<F> {
+        let mut coeffs = self.shifted_intt_batch(mat, shift);
+        // Zero-padding the coefficients keeps the same polynomial on the larger domain.
+        coeffs
+            .values
+            .resize(coeffs.values.len() << added_bits, F::ZERO);
+        self.shifted_ntt_batch(coeffs, shift)
```
tcoratger: This is the same `Vec::resize` reallocate-and-memcpy that the PR body (rightly!) criticises in `encode_batch` — it just survived here in the trait's own LDE. It also carries the same unchecked `<<`.

`F::zero_vec` + `copy_from_slice`, or a `reserve_exact` first, would make the two consistent.

binary-dft/src/naive.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/// `Ŵ_j` from the product definition, normalised at `v_j`.
+fn normalised_subspace_poly<F: TowerLevel>(j: usize, x: F) -> F {
+    let vanishing = |y: F| (0..1 << j).map(|m| y + domain_point::<F>(m)).product::<F>();
+    vanishing(x)
+        * vanishing(F::cantor_basis(j))
+            .try_inverse()
+            .expect("W_j does not vanish at v_j, which lies outside S_j")
+}
```
tcoratger, line 31: This function is the heart of why the oracle is a real oracle, and I like that it pays for the normalising division explicitly rather than assuming it away.

One gap though: the description says "the oracle's normalisation divisor came out as exactly 1 at every level", but nothing in the tree asserts it. `subspace_polynomial_shifts_the_basis` pins it for the *recurrence* form (`W_j(v_j) = v_0 = 1`), but the **product** form here — the independent route — is never checked. One line would pin the claim:

```rust
/// The Cantor basis makes the normalisation trivial. Assert that rather than believing it.
#[test]
fn the_normalisation_divisor_is_one() {
    for j in 0..8 {
        let divisor = (0..1usize << j)
            .map(|m| BinaryField8::cantor_basis(j) + domain_point::<BinaryField8>(m))
            .product::<BinaryField8>();
        assert_eq!(divisor, BinaryField8::ONE, "j={j}");
    }
}
```

binary-field/src/cantor.rs
```diff
@@ -0,0 +1,316 @@
...
+        check_recurrence::<BinaryField32>(32);
+        check_recurrence::<BinaryField64>(64);
+        check_recurrence::<BinaryField128>(128);
+    }
+
+    #[test]
+    fn basis_is_nested_and_independent() {
+        check_nesting::<BinaryField128>(128);
+        check_independent::<BinaryField128>(128);
+    }
```
tcoratger, line 236: These two run only at `BinaryField128`, but they're what justifies the `as $repr` truncation in `cantor_basis` at **every** level — the argument being `i < $bits = 2^k` ⟹ `(i+1).next_power_of_two() ≤ 2^k` ⟹ `v_i < 2^(2^k)` fits the level's `MASK`. Since they cost microseconds, running them at 2/4/8/16/32/64 too (the way `check_recurrence` already does) would make that argument load-bearing rather than inferred.

One more thought on this table: everything currently reduces to internal consistency between two in-tree routes, which is strong but shares one assumption — that the table is the one the regenerator would produce. `regenerate_cantor_basis` asserts the recurrence and the even-representation normalisation before printing, which is great, but a hand-edited table that flipped a root would sail through both. A handful of pinned literals (the first few `domain_point` values, or an 8-row `LchNtt::<BinaryField128>::ntt_batch` known-answer) would close that last gap cheaply.

binary-field/src/tower.rs
```diff
@@ -29,7 +29,7 @@ use crate::clmul::{HAS_HARDWARE_CLMUL, mul_64, mul_128, square_64, square_128};
 use crate::{Gf2, tables};
 
 /// One level of the Wiedemann tower. `Repr` is the backing integer; `LOG_BITS` is `k` for `GF(2^(2^k))`.
-pub(crate) trait TowerLevel: Field {
+pub trait TowerLevel: Field {
```
tcoratger: Going `pub(crate) -> pub` here exports rather more than the NTT needs: `from_repr`, `to_repr`, `mul_alpha` and `from_le_byte_iter` all become stable public API. `mul_alpha` in particular is a pure implementation detail of the quadratic-extension recursion (`X² = αX + 1`) with no meaning outside it.

The bit I'd actually worry about is that the trait isn't sealed, so a downstream type can implement `TowerLevel` and thereby instantiate `domain_point`, `subspace_polynomial`, `LchNtt` and `AdditiveNtt` over it — with a `cantor_basis` satisfying none of the identities the transform assumes. That fails silently with wrong evaluations rather than loudly.

Two easy options: seal it with a private supertrait, or export a narrow public trait carrying just what `p3-binary-dft` needs (`LOG_BITS`, `cantor_basis`, `to_repr`/`from_repr`) and keep `mul_alpha` on the internal one. The second is a bit more work but gives the new crate a much smaller contract to honour.

tcoratger (approved):

binary-dft/src/lch.rs
Nashtare: It was stale, I fixed the comment.

tcoratger:
The review was only about minor comments so that we can merge when CI is ready :)

commit/Cargo.toml
Nashtare: Hmm I'm not sure, I couldn't get away with the dependency cycle trying this though.

tcoratger: No problem let's not block the merge cycle on this, we can check later :)

PR #2005, opened 2026-08-26
perf(stir): bucket committed heights into several bounded-spread shared domains

Nashtare:
Follow-up to #1990, addressing the third item of @tcoratger's review comment on
`config.rs:445` — *"Decide what happens outside it. Falling back to the base branch's
per-height-class instances when `ell` does not fit the budget would keep the win where it is
affordable instead of panicking."* — using the mechanism named in the same review's
`pcs.rs:274` comment: *"the same knob fixes the blocker: bucketing matrices into several
shared domains by bounded height spread caps both this and `ell`."*

Items 1 (`verify` must not panic) and 2 (name the envelope) landed in #1998 and #1990
respectively; this is the remaining one.

## Summary

#1990 put every matrix of a `commit()` call onto one shared LDE domain sized to the tallest.
Two costs follow from that single choice: a short matrix pays the tallest one's blowup, and
`Combine`'s soundness cost grows with the whole committed spread (`log2(ell - 1) + log_d*`,
about `2·log_d*`, which is what makes the feasibility envelope so tight).

Both are bounded by the same knob. The matrices are now partitioned into **shared-domain
groups of bounded native-height spread**, each extended onto its own domain and committed in
its own tree, so a commitment carries one root per group.

The two existing layouts are the endpoints of that knob:

| `max_log_height_spread` | layout |
|---|---|
| `0` | every distinct native height on its own domain — no `Combine`, one STIR instance per height (the pre-#1990 layout) |
| ≥ committed spread | everything on one domain, all classes merged (#1990's layout) |
| `3` (`DEFAULT_MAX_LOG_HEIGHT_SPREAD`) | the spread of the shape #1990's trade was measured on |

Groups also shrink on their own when `Combine` does not fit the challenge field, so a
parameter set outside the envelope now degrades into more STIR instances rather than failing
— which is the fallback the review asked for.

## Benchmark

KoalaBear / 155-bit quintic, width 32, security 100, `log_blowup = 1`, `k = k0 = 16`.
Proof size is deterministic; commit/open are seconds-scale single runs, so the ratios are
well outside the noise. Verify is a single un-warmed call at ~10–17 ms and is indicative only.

| heights | spread cap | roots | commit | open | proof |
|---|---|---|---|---|---|
| `[20,18,17]` | 0 | 3 | 2099 ms | 1211 ms | 697 203 B |
| `[20,18,17]` | **3 (default)** | 1 | 4269 ms | 1122 ms | **601 472 B** |
| `[20,18,17]` | uncapped (#1990) | 1 | 4266 ms | 1156 ms | 601 472 B |
| `[20,12,12,12,12]` | **3 (default)** | 2 | **1547 ms** | 963 ms | **889 603 B** |
| `[20,12,12,12,12]` | uncapped (#1990) | 1 | 8457 ms | 1356 ms | 947 625 B |
| `[20,10]` | **3 (default)** | 2 | **1503 ms** | 768 ms | **391 891 B** |
| `[20,10]` | uncapped (#1990) | 1 | 2766 ms | 941 ms | 418 693 B |

At the shape #1990 was tuned on the default reproduces #1990 exactly — same layout, same
proof, same timings. At the wide shapes it is strictly better on **both** axes: 5.5x and 1.8x
less commit work *and* a smaller proof, because merging a `2^12` class into a `2^20` domain
costs more in degree correction than it saves in query counts. That is the commit-cost
regression the review measured (`[20,12,12,12,12]` at 2.82x prover), gone.

## Why the feasibility probe is conservative

A bucket pools every group topped at the same height **across all commitments opened
together**, so probing feasibility against one commitment's own heights is not enough:
commitment A holding `{2^8, 2^6}` and B holding `{2^8, 2^7}` both land on the `2^9` domain and
`Combine` over the union `{2^8, 2^7, 2^6}` — a class set neither holds. Each probe passes; the
union can still exceed the envelope.

`combine_band_width` therefore probes the whole band `[tallest - w, tallest]` rather than the
heights actually present. Adding a class raises Lemma 4.13's `ell` by `2^tallest + 1 - 2^h > 0`,
so the full band maximises `ell` over every subset that could form, and a width feasible for it
is feasible for whatever union shows up. The width depends only on the tallest height and the
parameters, so every commitment agrees on it without seeing the others.

The layout itself is never carried in the proof — both sides derive it from the claimed domain
sizes — so a verifier configured for a different spread expects a different number of trees and
rejects.

## API change

`Pcs::Commitment` goes back to `Vec<InputMmcs::Commitment>` (the pre-#1990 shape), one root per
group, so callers observe the roots in order rather than with a single `observe`. A commitment
whose heights all fit one group still has exactly one root. In-repo call sites (tests, bench,
`pcs_comparison`) are updated.

## Test plan

- [x] Round trips at spreads `{0, 1, 2, 8}` over heights `[8, 6, 4]`, each asserting the
      expected root count
- [x] Both endpoints pinned: `0` gives the per-height-class layout, a wide cap gives a single
      shared domain
- [x] Repeated and out-of-order heights map to the right groups (grouping is over *distinct*
      heights, so it does not depend on caller order or multiplicity)
- [x] Feasibility-driven shrinking where the spread cap alone would not split
      (`JohnsonBound`, security 80, heights `[2^12, 2^11]` on the 124-bit quartic: each height
      configures alone, the merge does not)
- [x] A bucket pooling classes from two commitments that neither holds on its own
- [x] `commit_ldes` reaches the same layout as `commit` (batch-stark commits quotient chunks
      through it)
- [x] A proof laid out at one spread is rejected by a verifier configured for another
- [x] `p3-stir` 143 tests green; `p3-security`, `p3-whir`, `p3-fri`, `p3-uni-stark`,
      `p3-batch-stark` unaffected
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo +nightly fmt --all --check` clean, rustdoc warning-free

### DIFF
--- a/examples/examples/pcs_comparison.rs
+++ b/examples/examples/pcs_comparison.rs
@@ -483,8 +483,9 @@ fn main() {
             vec![(domain, message)],
             &base_challenger,
             queries,
-            // STIR commits one Merkle tree per distinct LDE height.
-            |ch, commit| ch.observe(commit.clone()),
+            // STIR commits one Merkle tree per shared-domain group; a single table is one
+            // group, hence one root.
+            |ch, commit| commit.iter().for_each(|root| ch.observe(root.clone())),
         ));
     }
 
@@ -579,12 +580,14 @@ fn main() {
         ));
     }
 
-    // STIR: every table here shares one `commit()` call, so the real prover extends
-    // all three onto one shared LDE domain (sized to the tallest) and merges their
-    // native-height classes via `Combine` (§7, Construction 7.2) before STIR runs.
-    // Reconstruct that same bucket (`ell` per Lemma 4.13, matching what `p3_stir`'s
-    // PCS impl computes internally) so the printed schedule reflects what actually
-    // proves, not a plain single-height instance.
+    // STIR: every table here shares one `commit()` call, and the three heights span exactly
+    // `DEFAULT_MAX_LOG_HEIGHT_SPREAD` octaves, so they land in one shared-domain group: the
+    // real prover extends all three onto one LDE domain (sized to the tallest) and merges
+    // their native-height classes via `Combine` (§7, Construction 7.2) before STIR runs.
+    // Reconstruct that same bucket (`ell` per Lemma 4.13, matching what `p3_stir`'s PCS impl
+    // computes internally) so the printed schedule reflects what actually proves, not a plain
+    // single-height instance. A wider height spread would be split across several groups, and
+    // this reconstruction would then describe only the tallest.
     {
         let stir_params = StirParameters {
             log_blowup: args.rate,
@@ -629,8 +632,9 @@ fn main() {
             tables,
             &base_challenger,
             queries,
-            // STIR commits one shared Merkle tree for every table in this call.
-            |ch, commit| ch.observe(commit.clone()),
+            // STIR commits one Merkle tree per shared-domain group: the tables here are
+            // partitioned by bounded height spread, so this is one root per group.
+            |ch, commit| commit.iter().for_each(|root| ch.observe(root.clone())),
         ));
     }
 
--- a/examples/examples/pcs_comparison.rs
+++ b/examples/examples/pcs_comparison.rs
@@ -586,8 +586,9 @@ fn main() {
     // their native-height classes via `Combine` (§7, Construction 7.2) before STIR runs.
     // Reconstruct that same bucket (`ell` per Lemma 4.13, matching what `p3_stir`'s PCS impl
     // computes internally) so the printed schedule reflects what actually proves, not a plain
-    // single-height instance. A wider height spread would be split across several groups, and
-    // this reconstruction would then describe only the tallest.
+    // single-height instance. Where `Combine` does not fit the challenge field, the PCS splits
+    // the heights across several groups instead of failing, and so does the reconstruction:
+    // the schedule then printed is the tallest group's, without `Combine`.
     {
         let stir_params = StirParameters {
             log_blowup: args.rate,
@@ -600,12 +601,14 @@ fn main() {
         };
         let ell: u64 = heights.len() as u64 * ((1u64 << args.log_message_size) + 1)
             - heights.iter().map(|&h| 1u64 << h).sum::<u64>();
-        let config = StirConfig::<F, EF, ChallengeMmcs, Challenger>::new_with_combine(
+        let config = StirConfig::<F, EF, ChallengeMmcs, Challenger>::try_new_with_combine(
             args.log_message_size,
             stir_params.clone(),
             heights.len(),
             ell,
-        );
+        )
+        .or_else(|_| StirConfig::try_new(args.log_message_size, stir_params.clone()))
+        .expect("STIR parameters are infeasible even without Combine");
         let queries = config
             .round_configs
             .iter()
--- a/stir/src/pcs.rs
+++ b/stir/src/pcs.rs
@@ -1,26 +1,36 @@
 //! `TwoAdicStirPcs`: implementing the [`Pcs`] trait using STIR.
 //!
-//! **Commit**: every matrix passed to one `commit()` call is extended onto one shared LDE
-//! domain, sized to the *tallest* matrix (§7, Construction 7.2's same-domain requirement) —
-//! shorter matrices are extended further, at an effectively higher per-matrix blowup. All
-//! matrices land in one Merkle tree per commitment, rather than one tree per distinct height.
+//! **Commit**: the matrices passed to one `commit()` call are partitioned into *shared-domain
+//! groups* of bounded native-height spread. Each group is extended onto one shared LDE domain
+//! sized to that group's tallest matrix (§7, Construction 7.2's same-domain requirement) and
+//! committed in its own Merkle tree, so a commitment carries one root per group. A group
+//! holding a single native height needs no merging at all; a group holding several merges
+//! them below.
+//!
+//! The two ends of the spread cap ([`TwoAdicStirPcs::with_max_log_height_spread`],
+//! [`DEFAULT_MAX_LOG_HEIGHT_SPREAD`]) are the two layouts this generalizes: a cap of `0` puts
+//! every distinct native height on its own domain, so `Combine` never runs and each height
+//! gets its own STIR instance, while a cap at or above the committed spread puts everything on
+//! one domain and merges it all. Groups also shrink on their own when `Combine`'s soundness
+//! cost does not fit the challenge field, so an infeasible parameter set degrades into more
+//! STIR instances rather than failing.
 //!
 //! **Open**: alpha-batch quotient polynomials `(f_i(z) - f_i(x)) / (z - x)` into one
-//! reduced-opening polynomial per *native* (pre-shared-domain) matrix height — the same
-//! per-height batching as before, just with every reduced-opening now living on the shared
-//! domain rather than its own native-sized one. Commitments sharing the same shared-domain
-//! size are then run through one STIR instance together (as many domain sizes as distinct
-//! shared domains, "buckets" below); within a bucket, if more than one native-height class is
-//! present, they are merged into a single codeword via batch degree correction
-//! ([`crate::utils::combine_on_coset`], §4.5's `Combine`) before STIR runs,
-//! at the tallest class's degree and full proximity radius (no per-class query-count floor).
-//! The prover returns the deduplicated first-round STIR query indices alongside the IOP
-//! proof; at those positions the prover also opens the input LDE matrices (via `InputMmcs`)
-//! so the verifier can confirm the reduced-opening polynomials are correctly derived from the
-//! committed inputs.
+//! reduced-opening polynomial per *native* matrix height, each living on its group's shared
+//! domain. Groups sharing a domain size — across commitments as well as within one — are run
+//! through a single STIR instance ("buckets" below, one per distinct shared LDE height);
+//! within a bucket, if more than one native-height class is present, they are merged into a
+//! single codeword via batch degree correction ([`crate::utils::combine_on_coset`], §4.5's
+//! `Combine`) before STIR runs, at the tallest class's degree and full proximity radius (no
+//! per-class query-count floor). The prover returns the deduplicated first-round STIR query
+//! indices alongside the IOP proof; at those positions the prover also opens the input LDE
+//! matrices (via `InputMmcs`) so the verifier can confirm the reduced-opening polynomials are
+//! correctly derived from the committed inputs.
 //!
-//! **Verify**: replay the same alpha-batching and `Combine` from the opening values, then for
-//! each bucket call [`verify_stir_with_external_initial`](crate::verifier::verify_stir_with_external_initial).
+//! **Verify**: reproduce the grouping from the claimed domain sizes — it is a pure function of
+//! those and the PCS parameters, so no part of it travels in the proof — then replay the same
+//! alpha-batching and `Combine` from the opening values, and for each bucket call
+//! [`verify_stir_with_external_initial`](crate::verifier::verify_stir_with_external_initial).
 //! STIR's initial oracle *is* the (possibly combined) reduced opening, which the transcript
 //! already pins through the input commitments, the claimed values, `alpha`, and (when a
 //! bucket combines more than one class) the combination challenge, so it is never committed a
@@ -28,16 +38,15 @@
 //! input MMCS openings at exactly the positions STIR sampled. No hand-mirrored transcript
 //! replay is needed.
 //!
-//! **Cost profile**: opening and verification get cheaper — one STIR instance per commitment
-//! instead of one per height class — but committing gets more expensive, and the trade moves
-//! with the height spread. A matrix at native height `2^h` alongside a tallest matrix at
-//! `2^H` pays a `2^(H - h + log_blowup)` blowup instead of `2^log_blowup`, in both its DFT and
-//! its share of the Merkle tree (the tree is `2^(H + log_blowup)` rows deep and carries every
-//! matrix's full width in each leaf, so hashing goes from `Σᵢ 2^(hᵢ + b)·widthᵢ` to
-//! `2^(H + b)·Σᵢ widthᵢ`). Callers committing matrices whose heights differ by many octaves —
-//! or more classes than the `Combine` feasibility envelope on [`StirParameters`] admits —
-//! should split them across separate `commit()` calls, which keeps each shared domain close to
-//! its own matrices' heights.
+//! **Cost profile**: merging classes onto one domain makes opening and verification cheaper —
+//! one STIR instance instead of one per height class — and committing more expensive, and the
+//! spread cap is what bounds the second. A matrix at native height `2^h` in a group whose
+//! tallest is `2^H` pays a `2^(H - h + log_blowup)` blowup instead of `2^log_blowup`, in both
+//! its DFT and its share of that group's Merkle tree (the tree is `2^(H + log_blowup)` rows
+//! deep and carries every group member's full width in each leaf, so hashing goes from
+//! `Σᵢ 2^(hᵢ + b)·widthᵢ` to `2^(H + b)·Σᵢ widthᵢ`). Capping the spread caps `H - h`, so a
+//! short matrix committed alongside a much taller one lands in its own group and pays its own
+//! blowup rather than the tallest one's.
 
 use alloc::borrow::Cow;
 use alloc::sync::Arc;
@@ -93,23 +102,68 @@ pub struct InputOpenings<Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>> {
     pub opening_proof: InputMmcs::MultiProof,
 }
 
+/// One shared-LDE-domain group of a commitment: its own Merkle tree over the matrices whose
+/// native heights the partition placed together.
+///
+/// Every matrix here is extended onto the same domain (sized to the group's tallest) and
+/// committed in fiber-grouped form — each leaf holds `2^log_starting_folding_factor`
+/// consecutive bit-reversed LDE rows, exactly the rows one first-round STIR query reads.
+struct DomainGroup<Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>> {
+    data: InputMmcs::ProverData<RowMajorMatrix<Val>>,
+    /// Column count of each matrix in this group, in the order the caller committed them.
+    widths: Vec<usize>,
+    /// Native (pre-extension) log2 height of each matrix in this group, same order. This is
+    /// what distinguishes matrices for alpha-batching and `Combine` grouping once they all
+    /// sit on the same physical domain.
+    log_native_heights: Vec<usize>,
+    /// Log2 of the shared LDE domain this group's matrices were extended onto.
+    log_lde_height: usize,
+}
+
 /// Prover data for [`TwoAdicStirPcs`].
 ///
-/// Every matrix passed to one `commit()` call is extended onto one shared LDE domain (sized
-/// to the tallest matrix) and committed in fiber-grouped form in a single Merkle tree — each
-/// leaf holds `2^log_starting_folding_factor` consecutive bit-reversed LDE rows, exactly the
-/// rows one first-round STIR query reads. `log_native_heights[i]` is matrix `i`'s own
-/// (pre-shared-domain) log2 height, in the order the caller committed them; this is what
-/// distinguishes matrices for alpha-batching and `Combine` grouping once they all sit on the
-/// same physical domain.
+/// The matrices passed to one `commit()` call are partitioned into shared-domain groups of
+/// bounded height spread ([`TwoAdicStirPcs::with_max_log_height_spread`]), each committed in its
+/// own tree. A group of one native height runs no `Combine` at all; a group spanning several
+/// merges them on its shared domain per §7's same-domain requirement.
 pub struct StirProverData<Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>> {
-    data: InputMmcs::ProverData<RowMajorMatrix<Val>>,
-    /// Column count of each matrix, in the order the caller committed them.
-    widths: Vec<usize>,
-    /// Native (pre-shared-domain) log2 height of each matrix, in the order committed.
-    log_native_heights: Vec<usize>,
-    /// Log2 of the shared LDE domain every matrix in this commitment was extended onto.
-    log_shared_lde_height: usize,
+    /// Groups in descending LDE height, matching the commitment's root order.
+    groups: Vec<DomainGroup<Val, InputMmcs>>,
+    /// `placement[i] = (group index, index within that group)` for the caller's matrix `i`.
+    placement: Vec<(usize, usize)>,
+}
+
+impl<Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>> StirProverData<Val, InputMmcs> {
+    /// `(native log2 height, log2 LDE height of the group holding it)` per matrix, in the
+    /// order the caller committed them.
+    fn matrix_layout(&self) -> Vec<(usize, usize)> {
+        self.placement
+            .iter()
+            .map(|&(group_idx, idx)| {
+                let group = &self.groups[group_idx];
+                (group.log_native_heights[idx], group.log_lde_height)
+            })
+            .collect()
+    }
+
+    /// The group committed on the domain of size `2^log_lde_height`, if this commitment has
+    /// one. Group LDE heights are distinct, so at most one can match.
+    fn group_at(&self, log_lde_height: usize) -> Option<&DomainGroup<Val, InputMmcs>> {
+        self.groups
+            .iter()
+            .find(|group| group.log_lde_height == log_lde_height)
+    }
+}
+
+/// How one commitment's matrices are partitioned across shared LDE domains.
+///
+/// Derived identically by the prover (from the committed heights) and the verifier (from the
+/// claimed domain sizes), so no part of it travels in the proof.
+struct GroupPlan {
+    /// Log2 LDE height of each group, descending.
+    log_lde_heights: Vec<usize>,
+    /// Group index of each matrix, in the order the caller supplied them.
+    group_of_matrix: Vec<usize>,
 }
 
 /// Reinterpret a bit-reversed LDE as a matrix whose rows are whole STIR fibers.
@@ -129,15 +183,27 @@ fn group_fiber_rows<Val: Clone + Send + Sync>(
 }
 
 /// Recover views of the committed matrices in their pre-grouping shape and caller order.
+///
+/// Matrices live in per-group trees, so the caller order is reassembled through
+/// [`StirProverData::placement`].
 fn lde_views<'a, Val: Send + Sync + Clone, InputMmcs: Mmcs<Val>>(
     input_mmcs: &InputMmcs,
     prover_data: &'a StirProverData<Val, InputMmcs>,
 ) -> Vec<RowMajorMatrixView<'a, Val>> {
-    let matrices = input_mmcs.get_matrices(&prover_data.data);
-    matrices
-        .into_iter()
-        .zip(&prover_data.widths)
-        .map(|(grouped, &width)| RowMajorMatrixView::new(grouped.values.as_slice(), width))
+    let per_group: Vec<Vec<&'a RowMajorMatrix<Val>>> = prover_data
+        .groups
+        .iter()
+        .map(|group| input_mmcs.get_matrices(&group.data))
+        .collect();
+
+    prover_data
+        .placement
+        .iter()
+        .map(|&(group_idx, idx)| {
+            let grouped = per_group[group_idx][idx];
+            let width = prover_data.groups[group_idx].widths[idx];
+            RowMajorMatrixView::new(grouped.values.as_slice(), width)
+        })
         .collect()
 }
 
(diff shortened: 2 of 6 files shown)


Nashtare:
@tcoratger let me know if that looks good to you whenever you have some time! :)

tcoratger (commented):
Spent a good while on this one — it is the third layout change to the STIR PCS in a row and the first one that makes the layout a *derived* object both sides have to agree on without help from the proof, so I wanted to attack the derivation rather than read it. Short version: **the design is right and I could not break it.** The band-probe argument in the description is not hand-waving, it is a real theorem and I checked it exhaustively rather than taking it on faith. My substantive finding is a performance one, and it is in the partition algorithm rather than anywhere near the crypto.

## Things I verified rather than assumed

**1. The band probe is conservative — checked exhaustively, not argued.**

The description's argument reduces to: *if the full band `[T-w, T]` configures, then so does every subset of it containing `T`*. I re-derived why that is the right reduction (every group feeding a bucket is topped at `T = log_h - log_blowup`, so `d* = 2^T` is fixed across the bucket and Lemma 4.13's `ell` grows by `2^T + 1 - 2^h > 0` per added class, hence the band maximises `ell` over every union that can form), then tested the implication directly:

```
security_level {16,32,48,64,80,96} x log_blowup {1,2,3} x log_degree {6,8,10,12,14}
  x {CapacityBound, JohnsonBound}                                 = 180 parameter sets
  x band widths 1..=8, and for each feasible band EVERY subset containing the tallest

0 counterexamples (2.7 s in release)
```

So the fallback really does degrade instead of failing, and it does so for the right reason. This is the single property the whole mechanism rests on and right now it exists only as prose in the `combine_band_width` doc comment — I would very much like it pinned as a test, sketch inline.

**2. The `group_mats -> claim_idx` remap is correct.** This is the sharpest new line in the diff (`pcs.rs:1437`): `mat_idx` now indexes a *group's* tree while `point_data` is keyed by the commitment's full claim order. I hammered it with a randomised sweep — 40 cases, spreads 0..4, 1-3 commitments per case, 1-4 matrices each, random heights `2^2..2^8`, random widths, random caller order — each doing a full commit/open/verify **plus an exhaustive single-value tamper on every matrix of every commitment**. All round trips verified; all tamperings rejected. No issues.

**3. Layout mismatch survives the case the new test misses.** `test_pcs_verify_rejects_a_proof_built_at_a_different_spread` uses 1 root vs 3 roots, which the cheap `commitment.len()` check catches. The interesting case is two spreads that agree on the *count* and disagree on the *heights*: heights `[8,7,4,3]`, prover at spread 1 (`{8,7} | {4,3}`, LDE `[9,5]`) vs verifier at spread 4 (`{8,7,4} | {3}`, LDE `[9,4]`) — two trees either way. Still rejected (`InvalidProofShape`), on the MMCS dimensions rather than the shape check. Good, but that is a much longer chain of reasoning and deserves a test.

**4. No union bound is owed over the instance count.** Splitting into more STIR instances is the whole point of the PR, so I checked whether `security_level` still means what it says with `G` buckets. It does: the adversary must win in *every* instance it made false, so the round error is `max_i eps_i` and not `sum_i eps_i`. `prover.rs:873-877` already states this correctly for the shared grind. Flagging only so you know it was checked and not assumed.

Suites green here too (52 + 69 + 22), worktree clean.

## Two things I would like to see before this lands

**1. The partition is a top-down greedy that never looks at cost.** `partition_native_heights` maximises the *top* group, which is exactly the wrong direction: a height sitting at the band edge gets dragged onto the tallest domain and pays the full `2^spread` blowup when moving it one group down is free. Running the real planner at the default cap:

| heights | this PR | LDE domains |
|---|---|---|
| `[20,18,17]` | `[[20,18,17]]` | `[21]` |
| `[20,12,12,12,12]` | `[[20]] [[12,12,12,12]]` | `[21,13]` |
| `[20,10]` | `[[20]] [[10]]` | `[21,11]` |
| **`[20,17,16,15]`** | **`[[20,17]] [[16,15]]`** | **`[21,17]`** |
| **`[16,13,12,11,10]`** | **`[[16,13]] [[12,11,10]]`** | **`[17,13]`** |

The first three are exactly your benchmark rows, and they reproduce — but every one of them is either fully inside a band or fully outside it. Neither has a height *at* the edge, which is the only place the greedy is wrong. On `[20,17,16,15]` the partition `[[20]] [[17,16,15]]` has the **same group count, hence the same number of STIR instances**, and is ~1.55x less commit work at equal widths (unboundedly better if the short table is wide) *and* leaves the tall bucket single-class so `Combine` does not run there at all. Details and a fix inline.

**2. `Commitment = Vec<_>` drops out of `StarkGenericConfig`.** Going back to the pre-#1990 shape reintroduces a pre-#1990 limitation: `uni-stark/src/config.rs:33` requires `Challenger: CanObserve<Pcs::Commitment>`, and no challenger in the repo implements `CanObserve<Vec<T>>`. On `main` today `TwoAdicStirPcs` satisfies that bound; after this PR it does not, and I confirmed with a compile-time assertion (error text inline). Worth deciding explicitly rather than by accident, since the "batch-stark commits quotient chunks through `commit_ldes`" line in the test plan is currently only exercised through the `Pcs` trait directly.

Neither is a correctness problem, so if you would rather land this and follow up I would not object — I would just want #2 stated in the API-change section.

## Small stuff, not worth its own thread

- `pcs.rs:1477`, `if expected_ro_by_class.len() != coeffs_by_height.len()`: this can never fire. `expected_ro_by_class` is `vec![_; native_heights.len()]` and `coeffs_by_height` is `native_heights.iter().copied().zip(coeffs).collect()` into a `BTreeMap` over already-deduplicated heights, so the lengths are equal by construction. It reads like a real adversarial check sitting next to three genuine ones. `debug_assert_eq!` or drop it.
- `DEFAULT_MAX_LOG_HEIGHT_SPREAD` silently changes the on-the-wire shape for anyone on `main` today (different root count, different instance count, different transcript, for any committed spread above 3). The API-change section only mentions the `Vec`; this deserves a line too.
- The rest of the module docs are, as usual, excellent — the cost note at the top and the `combine_band_width` doc in particular are the reason this was reviewable at all.

stir/src/pcs.rs
```diff
@@ -278,6 +397,112 @@ where
...
+        }
+        width
+    }
+
+    /// Partition distinct native heights, descending, into shared-domain groups.
+    ///
+    /// Returns each group's size, so group `g` covers the slice starting after the previous
+    /// groups. Each group takes every remaining height inside the band
+    /// [`Self::combine_band_width`] admits below its tallest.
+    fn partition_native_heights(&self, descending: &[usize]) -> Vec<usize> {
```
tcoratger, line 536: The greedy takes *every* remaining height inside the top group's band before opening the next one, which maximises the tallest group. That is the wrong direction: the cost of a group is dominated by its tallest member, so pulling a short matrix up into it is the single most expensive thing the planner can do, and the greedy does it whenever the short matrix happens to land on the band edge.

Running the real planner (`plan_groups`, default cap, `CapacityBound`, `log_blowup = 1`) on your benchmark shapes plus two edge-case ones:

```
[20, 18, 17]          -> lde [21]      groups [[20, 18, 17]]
[20, 12, 12, 12, 12]  -> lde [21, 13]  groups [[20], [12, 12, 12, 12]]
[20, 10]              -> lde [21, 11]  groups [[20], [10]]
[20, 17, 16, 15]      -> lde [21, 17]  groups [[20, 17], [16, 15]]      <-- edge
[16, 13, 12, 11, 10]  -> lde [17, 13]  groups [[16, 13], [12, 11, 10]]  <-- edge
```

The three benchmark rows reproduce exactly, which is good — but each is either fully inside one band or fully outside it. Neither has a height *at* the edge, and that is the only place the greedy is wrong. Take `[20,17,16,15]`:

```
                     this PR (greedy)              alternative, same group count
groups               {20,17} | {16,15}             {20} | {17,16,15}
LDE domains          2^21    | 2^17                2^21 | 2^18
2^17 matrix pays     16x blowup                    2x blowup
tall bucket          Combine over 2 classes        one class, no Combine at all

output points (DFT + Merkle leaf material), equal widths w:
  greedy   2^21*2w + 2^17*2w  =  17 * 2^18 * w
  alt      2^21*w  + 2^18*3w  =  11 * 2^18 * w        ->  1.55x less work
```

Same number of groups, hence the same number of STIR instances and the same query-count structure — this is not a proof-size/prover-time trade, the alternative is simply better on both axes. It also removes `Combine` from the tall bucket entirely: no `ell`, no inflated `eta`, no `combine_term` in the union bound. And 1.55x is the *equal-width* case; the ratio is unbounded in `w17`, and a wide short table next to a narrow tall one is a very ordinary trace shape.

I checked the alternative is actually admissible rather than assuming it — `combine_band_width` returns 3 for every tallest in `[8, 20]` at these parameters, and `plan_groups(&[17,16,15])` gives a single group at LDE 18:

```
w(8)=3 w(12)=3 w(16)=3 w(17)=3 w(18)=3 w(19)=3 w(20)=3
[17, 16, 15] -> lde [18] / group_of_matrix [0, 0, 0]
```

**Fix.** There are at most `TWO_ADICITY` (<= 32) distinct heights, so an `O(n^2)` DP is free next to a single DFT:

```
best[j] = min over i < j of  best[i] + cost(i..j)
cost(i..j) = 2^(descending[i] + log_blowup) * (sum of widths in i..j) + per_instance_cost
admissible iff descending[i] - descending[j-1] <= combine_band_width(descending[i])
```

Two things to keep straight, neither of which is a problem:

- **The conservativeness argument survives untouched.** Nothing in it needs the groups to be *maximal* — it only needs every group to be contained in the band of its own tallest, which the admissibility predicate above still enforces. The bucket's union still lies in `[T - max_i w_i, T]` and that band was still validated by whichever commitment computed `max_i w_i`.
- **The layout must stay a pure function of data the verifier has.** Widths qualify — `verify` already reads each matrix's width off `point_claims.first().len()` at `pcs.rs:1360` — but `plan_groups` currently takes heights only, so you would have to thread widths through `commit` / `commit_ldes` / `verify` identically. If that coupling feels like too much for the win, the width-agnostic proxy `sum_g 2^(h_g_max) * |group_g|` already fixes all three edge cases above and needs no signature change at all.

**Repro**, if you want to see it before deciding:

```rust
let pcs = test_pcs_with(DEFAULT_MAX_LOG_HEIGHT_SPREAD, SecurityAssumption::CapacityBound, 16);
assert_eq!(pcs.plan_groups(&[20, 17, 16, 15]).log_lde_heights, vec![21, 17]);
//                                       ^ the 2^17 matrix on a 2^21 domain
```

stir/src/pcs.rs
```diff
@@ -278,6 +397,112 @@ where
...
+    /// could form, and a width feasible for it is feasible for whatever union actually shows
+    /// up. Since the width depends only on `tallest` and this PCS's parameters, every
+    /// commitment independently agrees on it.
+    ///
+    /// A width of `0` runs no `Combine` at all, so this always terminates: in the worst case
+    /// every distinct height gets its own domain and its own STIR instance.
+    ///
+    /// The configs derived while probing are served from the same cache the bucket
+    /// construction later reads, so a repeated shape pays for them once.
+    fn combine_band_width(&self, tallest: usize) -> usize {
```
tcoratger: This function and its doc comment are the load-bearing part of the PR, so rather than read the argument I tested the implication it reduces to: **if the full band `[tallest - w, tallest]` configures, then so does every subset of it containing `tallest`**. Exhaustively:

```
security_level {16,32,48,64,80,96}
  x log_blowup {1,2,3}
  x log_degree {6,8,10,12,14}
  x {CapacityBound, JohnsonBound}                      = 180 parameter sets
  x band widths w = 1..=8
  x for each feasible band, ALL 2^w subsets containing the tallest

0 counterexamples, 2.7 s in release
```

I also checked the weaker monotonicity along the band family alone (`ell` increasing => feasibility never returns) over the same grid, same result. So the probe really is conservative and the fallback really does degrade rather than fail — nice piece of design.

**But it lives only in this doc comment.** If `stir_combine_eta`, the `final_eta = max(...)` composition, or the round-0 `query_count -> prev_queries -> stir_recursive_eta` chain ever stops being monotone in `ell`, nothing fails loudly: the probe says "band feasible", the *actual* bucket class set is a strict subset, its config comes back `Err`, and `open` reaches `get_or_compute_stir_config` — which panics, on the prover, at a parameter set the fallback was written to rescue. That is worth a test, and it is a clean one because the property is cheap and the input space is exactly the structured grid above:

```rust
#[test]
fn band_feasibility_implies_subset_feasibility() {
    for (sec, lb, ld, st) in grid() {
        let mut params = /* ... */;
        params.log_blowup = lb;
        for w in 1..=ld.min(8) {
            let band: Vec<usize> = (0..=w).map(|i| ld - i).collect();
            let (n, ell) = TestPcs::combine_key(&band).unwrap();
            if TestConfig::try_new_with_combine(ld, params.clone(), n, ell).is_err() {
                continue;                       // band infeasible, nothing to imply
            }
            for mask in 0u32..(1 << w) {        // every subset containing `ld`
                let mut subset = vec![ld];
                subset.extend((0..w).filter(|i| mask & (1 << i) != 0).map(|i| ld - 1 - i));
                if subset.len() < 2 { continue; }
                let (sn, sell) = TestPcs::combine_key(&subset).unwrap();
                assert!(
                    TestConfig::try_new_with_combine(ld, params.clone(), sn, sell).is_ok(),
                    "band [{}..{ld}] configures but subset {subset:?} does not", ld - w,
                );
            }
        }
    }
}
```

A proptest would work too, but I think the exhaustive sweep is the better fit here: the space that matters is small and structured, and "every subset" is the actual quantifier in the theorem, so enumerating it *is* the property rather than a sample of it.

---

Separate, much smaller point on this function: it always probes `max_log_height_spread` octaves below `tallest`, even when the remaining heights cannot possibly fill the band. For the last group — a single leftover height — all three probes at the default cap are pure waste. Measured on `[16,15,14,13,12,11,10,9,8]`:

```
cold plan_groups   240 us      warm   992 ns      configs left in cache   9
                                                  of which usable by any bucket   6
```

Those three dead entries also compete for the 256-slot `CONFIG_CACHE_CAPACITY`, past which nothing is cached at all. The fix is free and provably layout-preserving — pass the lowest remaining height and fold it into `max_width`:

```rust
fn combine_band_width(&self, tallest: usize, lowest_remaining: usize) -> usize {
    let max_width = self
        .max_log_height_spread
        .min(tallest)
        .min(tallest - lowest_remaining);   // cannot probe wider than the data can fill
    // ...
}
```

Capping only ever shrinks the width, and it only binds when the whole remaining set already fits inside the band — in which case `take_while(|&&h| h >= floor)` selects the same group either way, so **the partition is bit-identical**. The cross-commitment argument also survives: a commitment that capped still has all its remaining heights inside its own (narrower) band, and the bucket's union still lies in `[T - max_i w_i, T]`, validated by whichever commitment computed the max.

stir/src/pcs.rs
```diff
@@ -316,14 +541,18 @@ where
...
-    /// single tree on the shared LDE domain (§7's same-domain requirement).
-    type Commitment = InputMmcs::Commitment;
+    /// One Merkle root per shared-domain group, in descending LDE height.
+    ///
+    /// The matrices of one `commit()` call are partitioned into groups of bounded height
+    /// spread, each extended onto its own shared domain (§7's same-domain requirement applies
+    /// within a group, not across the whole commitment) and committed in its own tree. A
+    /// commitment whose heights all fit one group therefore has a single root, and callers
+    /// observe the roots in order.
+    type Commitment = Vec<InputMmcs::Commitment>;
```
tcoratger: Going back to the pre-#1990 shape brings back a pre-#1990 limitation: `StarkGenericConfig` requires `Challenger: CanObserve<Pcs::Commitment>` (`uni-stark/src/config.rs:33`, and `batch-stark/src/config.rs:72`), and nothing in `p3-challenger` implements `CanObserve<Vec<T>>`. I checked with a compile-time assertion at this PR's head rather than reasoning about it:

```rust
fn assert_observable<C: CanObserve<T>, T>() {}
assert_observable::<Challenger, <MyPcs as Pcs<Challenge, Challenger>>::Commitment>();
```

```
error[E0277]: the trait bound `DuplexChallenger<..., 16, 8>:
  CanObserve<Vec<MerkleCap<MontyField31<BabyBearParameters>, [...; 8]>>>` is not satisfied
  = help: `DuplexChallenger<F, P, WIDTH, RATE>` implements `CanObserve<T>`:
            CanObserve<&MerkleCap<F, [F; N]>>
            CanObserve<F>
            CanObserve<MerkleCap<F, [F; N]>>
            CanObserve<Vec<Vec<F>>>
            CanObserve<[F; N]>
            CanObserve<Hash<F, F, N>>
```

On `main` (with #1990) `Commitment = InputMmcs::Commitment = MerkleCap<..>` and that bound *is* satisfied, so `TwoAdicStirPcs` is droppable into `p3-uni-stark` today and is not after this PR. Practical consequence for the test plan: `test_pcs_commit_ldes_groups_like_commit_and_round_trips` justifies itself with "batch-stark commits quotient chunks through it", but batch-stark cannot actually be instantiated with this PCS on this branch, so that test is a proxy rather than the integration.

Two ways out, and I mildly prefer the second:

1. A blanket `impl<F, P, const W: usize, const R: usize, T> CanObserve<Vec<T>> for DuplexChallenger<F, P, W, R> where Self: CanObserve<T>` in `p3-challenger`. One line, but it silently applies to every PCS.
2. A newtype, `pub struct StirCommitment<C>(Vec<C>)`, with its own `CanObserve` impl and `Deref<Target = [C]>` so the `commit.iter().for_each(observe)` call sites stay as they are.

The newtype also closes a small latent thing: the roots are currently observed with no length framing, so `(2 roots, 1 root)` and `(1 root, 2 roots)` across two commitments reach the same transcript state. **Not exploitable here** — the per-commitment root count is fixed by the verifier's own claimed heights and checked at `pcs.rs:1056` before anything else happens — but #1990's single-root shape did not have the ambiguity at all, and a newtype lets you observe the count first and make it structurally impossible rather than argued.

stir/src/pcs.rs
```diff
@@ -166,12 +232,27 @@ type StirConfigCache<Val, Challenge, StirMmcs, Challenger> = Arc<
...
 const CONFIG_CACHE_CAPACITY: usize = 256;
 
+/// Default cap on the height spread sharing one LDE domain.
+///
+/// Merging native-height classes onto one domain trades commit work for proof size: a matrix
+/// `s` octaves below its group's tallest pays a `2^s` larger blowup, while §7's `Combine`
+/// removes a whole STIR instance and its query-count floor. Three octaves is the spread of
+/// the shape that trade was measured on; past it the extra DFT and Merkle work grows without
+/// bound while `Combine`'s return does not, so wider spreads get their own domain.
+pub const DEFAULT_MAX_LOG_HEIGHT_SPREAD: usize = 3;
```
tcoratger, line 289: Worth calling out in the PR body's **API change** section alongside the `Vec`: `new()` installs this default, so anyone on `main` today who commits a spread wider than 3 octaves silently gets a different root count, a different number of STIR instances and a different transcript. The `Vec` shape is a compile error they will notice; this one is not.

Two small things on the doc while you are here:

- "Three octaves is the spread of the shape that trade was measured on" is accurate, but three octaves is also an `8x` blowup for the shortest member of a group, paid in both its DFT and its full width in every leaf of the group's tree. That is a lot for a wide short table, and a reader picking a value would benefit from the sentence being framed as the cost it buys rather than the shape it came from.
- There is no getter for the configured value. `with_max_log_height_spread` is `#[must_use]` and consuming, so a caller who wants to log or assert what a `TwoAdicStirPcs` was built with has no way to ask. A `pub const fn max_log_height_spread(&self) -> usize` would make the "both sides must agree on this" contract in the doc checkable at runtime instead of by convention.

stir/src/pcs.rs
```diff
@@ -190,33 +271,71 @@ impl<Val, Dft, InputMmcs, StirMmcs, Challenge, Challenger>
...
+        }
+
+        let (commitments, groups) = izip!(
+            per_group,
+            per_group_widths,
+            per_group_heights,
+            &plan.log_lde_heights
         )
+        .map(|(matrices, widths, log_native_heights, &log_lde_height)| {
+            let (commitment, data) = self.input_mmcs.commit(matrices);
```
tcoratger: Tree construction is now sequential across groups. Before this PR one `commit()` built a single tree and `MerkleTreeMmcs::commit` parallelised internally over its rows; now `G` trees are built one after another inside this `map`, and only the tallest has enough rows to saturate the pool. On `[16,13,12,11,10]` the second tree is `2^13` rows deep and gets a rayon dispatch all to itself while nothing else runs.

The groups are fully independent — separate matrices, separate `ProverData`, no shared state — so this is a `par_iter` away:

```rust
let (commitments, groups): (Vec<_>, Vec<_>) = izip!(...)
    .collect::<Vec<_>>()
    .into_par_iter()
    .map(|(matrices, widths, log_native_heights, &log_lde_height)| { /* unchanged */ })
    .unzip();
```

Needs `InputMmcs::ProverData<_>: Send`, which `MerkleTreeMmcs` satisfies. Nested rayon is fine here (the inner commits will just work-steal), and it costs nothing in the common single-group case since `par_iter` over one element does not dispatch. Small, but it is free and it partly offsets the extra commit work the spread cap is there to bound in the first place.

stir/src/pcs.rs
```diff
@@ -1148,6 +1431,10 @@ where
                                 for (mat_idx, point_slots) in
                                     mat_point_slots.iter().enumerate()
                                 {
+                                    // `mat_idx` indexes this group's tree, and therefore the
+                                    // opened rows; `point_data` is keyed by the commitment's
+                                    // full claim order, which `group_mats` maps back to.
+                                    let claim_idx = group_mats[mat_idx];
```
tcoratger, line 1553: This is the sharpest new line in the diff and it is correct — but I want to flag that **the suite cannot currently tell it apart from `mat_idx`**.

Every PCS round-trip test commits heights in descending order (`[8, 6, 4]`, `[8, 8]`, ...), so `plan.group_of_matrix` is monotone, `group_mats` is always a contiguous prefix of the claim indices, and `group_mats[mat_idx] == mat_idx` for the first group. Replace `claim_idx` with `mat_idx` and every test in the file still passes. Same for `mat_widths`: all round trips use width 3, so a width taken from the wrong claim is indistinguishable from the right one.

I did hammer it properly and it holds — 40 randomised cases (spreads 0..4, 1-3 commitments, 1-4 matrices each, random heights `2^2..2^8`, random widths, random caller order), each a full commit/open/verify plus an exhaustive single-value tamper on every matrix, all verifying and all tamperings rejected. So this is purely a "the next person to touch it has no net" concern; concrete tests on the test file.

stir/tests/stir.rs
```diff
@@ -939,10 +958,311 @@ mod babybear_pcs {
...
+            &proof,
+            &mut v_ch,
+        )
+        .unwrap_or_else(|e| panic!("verification failed: {e:?}"));
+
+        commit
+    }
+
+    #[test]
+    fn test_pcs_round_trips_at_every_spread() {
```
tcoratger, line 1066: Three gaps here, all cheap to close, and I have run all three so they are known-good rather than speculative.

**1. Every round trip commits heights in descending order with a uniform width.** That makes `group_mats` a contiguous prefix of the claim indices in every test, so the `claim_idx = group_mats[mat_idx]` remap at `pcs.rs:1437` and the per-group `mat_widths` lookup are never exercised on a real permutation. Interleaved heights and distinct widths fix both at once:

```rust
#[test]
fn test_pcs_round_trips_with_interleaved_heights_and_widths() {
    // Ascending, interleaved and duplicated, with a distinct width per matrix, so
    // `group_mats` is a non-contiguous index set and a claim/width mixup shows up.
    for spread in [0usize, 1, 2, 3, 8] {
        round_trip_interleaved(spread, &[4, 8, 4, 8, 6], &[2, 5, 3, 7, 4]);
    }
    round_trip_interleaved(2, &[6, 8, 2, 4], &[1, 9, 4, 2]);
}
```

(`round_trip_interleaved` is `round_trip_under` with a per-matrix width; happy to push the exact body if useful.) Passes on this branch.

**2. No test has a commitment that is absent from a bucket while another is present.** `test_pcs_merges_classes_pooled_from_several_commitments` has both commitments feeding the single bucket, so the `input_openings[i] == None` branch at `pcs.rs:1331` is only ever hit in single-bucket proofs. A three-commitment shape at spread 1 gives all three states at once:

```
A = {2^8, 2^4}  -> 2 roots, LDE [9, 5]
B = {2^6}       -> 1 root,  LDE [7]
C = {2^4, 2^8}  -> 2 roots, LDE [9, 5]     (interleaved caller order)

buckets 9, 7, 5:  B is absent from two of the three
```

Also passes on this branch, and it exercises the `None`/`Some` mismatch checks in both directions.

**3. No proptest on `plan_groups` itself.** This one genuinely wants property-based coverage rather than examples, because the input is a multiset and the interesting failures are in how duplicates, ordering and band edges interact:

> **Inputs:** `log_native_heights` a random `Vec<usize>` of length 1..8 with values in `2..=12` (duplicates encouraged); `max_log_height_spread` in `0..=8`.
> **Invariants:** the groups partition the *distinct* heights; `log_lde_heights` is strictly descending; every group's spread is `<= combine_band_width(its tallest)`; every matrix's group holds its own height; and the plan is invariant under permutation and under duplication of the input heights.
> **What a failure proves:** the layout is no longer a pure function of the height multiset — which is exactly the assumption the verifier reconstructs it under, so a failure here is a verifier that disagrees with the prover for a reason no shape check would explain.

A handful of known-answer cases would not generalise over the duplicate/ordering interaction, which is why I would reach for proptest here specifically and not for the band property (that one is small and structured enough to enumerate exhaustively — see the `combine_band_width` thread).

stir/tests/stir.rs
```diff
@@ -939,10 +958,311 @@ mod babybear_pcs {
...
+            assert_eq!(
+                commit.len(),
+                expected_roots,
+                "spread {max_log_height_spread} should give {expected_roots} groups"
+            );
+        }
+    }
+
+    #[test]
+    fn test_pcs_commit_ldes_groups_like_commit_and_round_trips() {
```
tcoratger, line 1105: This test compares only `commit.len()`:

```rust
assert_eq!(commit.len(), direct_commit.len(), "commit_ldes and commit disagree on the layout ...");
```

Two different layouts can agree on the group *count*. Heights `[8,7,4,3]` give `{8,7} | {4,3}` at spread 1 and `{8,7,4} | {3}` at spread 4 — both two trees, different LDE heights, different class sets, different `ell`. The assertion the comment describes ("has to reach the same layout `commit` does") is stronger than the one it makes.

Cheapest upgrade is to compare the plans directly rather than a proxy of them:

```rust
let plan_commit = pcs.plan_groups(&log_degrees);
let plan_ldes   = /* heights recovered from `ldes` exactly as commit_ldes does */;
assert_eq!(plan_commit.log_lde_heights, plan_ldes.log_lde_heights);
assert_eq!(plan_commit.group_of_matrix, plan_ldes.group_of_matrix);
```

`plan_groups` is private to `pcs.rs`, so from the integration test the observable equivalent is the per-group tree *dimensions* — commit both ways and assert `input_mmcs.get_matrices(..)` agree on heights and widths group by group. Either is fine; the point is that `len()` alone cannot see the failure the comment is worried about.

While here: the `commit_ldes` short-circuit is a genuinely nice side effect of this PR that the description undersells — `lde.height() == 1 << log_lde_height` is now true for every matrix that landed in its own group, so at spread 0 the whole `bit_reverse -> coset_idft -> truncate -> coset_lde -> bit_reverse` round trip disappears for every matrix. Worth a line in the summary.

stir/tests/stir.rs
```diff
@@ -939,10 +958,311 @@ mod babybear_pcs {
...
+                (commit_b, claims(1, &domains_b)),
+            ],
+            &proof,
+            &mut v_ch,
+        )
+        .unwrap_or_else(|e| panic!("pooled-class verification failed: {e:?}"));
+    }
+
+    #[test]
+    fn test_pcs_verify_rejects_a_proof_built_at_a_different_spread() {
```
tcoratger: The good version of this test is the one where the two spreads *agree* on the root count. As written, prover-at-8 gives 1 root and verifier-at-0 expects 3, so the `commitment.len() != plan.log_lde_heights.len()` check at `pcs.rs:1056` rejects immediately — a two-line path. The case that actually exercises the reasoning in the module doc ("a verifier configured for a different spread expects a different number of trees and rejects") is when it expects the *same* number:

```
heights [8, 7, 4, 3]
  prover   at spread 1 ->  {8,7} | {4,3}   LDE [9, 5]   2 roots
  verifier at spread 4 ->  {8,7,4} | {3}   LDE [9, 4]   2 roots
```

The shape check passes, `proof.len() == bucket_log_heights.len()` passes, and rejection only comes later, out of `verify_multi_batch` on mismatched dimensions. I ran it — it does reject, with `InvalidProofShape` — but that outcome depends on a much longer chain than the current test covers, and it is the chain the doc comment is actually asserting. Worth adding next to the existing one, with a comment saying *why* it is the interesting case so nobody later "simplifies" it back into a root-count check.

(Full test body available if you want it dropped in verbatim — it is the existing test with `[8,7,4,3]`, spread 1 vs 4, and the `matches!(err, InvalidProofShape)` assertion unchanged.)

examples/examples/pcs_comparison.rs
```diff
@@ -579,12 +580,14 @@ fn main() {
         ));
     }
 
-    // STIR: every table here shares one `commit()` call, so the real prover extends
-    // all three onto one shared LDE domain (sized to the tallest) and merges their
-    // native-height classes via `Combine` (§7, Construction 7.2) before STIR runs.
-    // Reconstruct that same bucket (`ell` per Lemma 4.13, matching what `p3_stir`'s
-    // PCS impl computes internally) so the printed schedule reflects what actually
-    // proves, not a plain single-height instance.
+    // STIR: every table here shares one `commit()` call, and the three heights span exactly
```
tcoratger, line 583: This comment and the reconstruction below it have drifted from what the PCS now does, in a way that will bite exactly the users this PR is for.

**The reconstruction aborts where the PCS degrades.** Line 603 calls the panicking constructor:

```rust
let config = StirConfig::<F, EF, ChallengeMmcs, Challenger>::new_with_combine(
    args.log_message_size, stir_params.clone(), heights.len(), ell,
);
```

The entire point of `combine_band_width` is that when `Combine` does not fit the challenge field, `plan_groups` splits and the PCS keeps working. This example does not: it panics before the PCS is even constructed. Heights are `[L-3, L-2, L]`, a spread of exactly 3, so for any `--log-message-size` / `--security-level` / field combination outside the envelope the example dies at precisely the parameter sets the PR was written to rescue — and dies *before* printing the FRI and WHIR rows it was invoked to compare against.

**And it assumes one group even when it gets one.** The comment now hedges for "a wider height spread", but feasibility-driven splitting is the other way in, and it is invisible: `ell`, `heights.len()` and `log_message_size` are all reconstructed by hand here rather than asked of the PCS, so if the layout splits the printed query schedule silently describes a bucket that does not exist.

Smallest fix that covers both:

```rust
let queries = match StirConfig::try_new_with_combine(args.log_message_size, stir_params.clone(), heights.len(), ell) {
    Ok(config) => /* as today */,
    Err(_) => {
        // Combine does not fit here, so the PCS splits into per-group instances;
        // report the tallest group's plain schedule rather than aborting.
        let config = StirConfig::try_new(args.log_message_size, stir_params.clone())?;
        /* ... */
    }
};
```

Structurally nicer would be to stop re-deriving the layout in the example at all — expose the plan (or just the bucket `(log_stir_degree, combine_key)` list) from `TwoAdicStirPcs` and have the example print what the PCS will actually run. The duplication is what lets the two drift, and it has now drifted twice.

tcoratger (approved):

PR #2006, opened 2026-08-28
feat(sumcheck): ring switching over an arbitrary field extension

Nashtare:
Stacked on #2003 (`robin/binary_field_support`). Phase 3 part 1 of binary-field support: the packed alphabet and the ring-switching reduction.

## What this adds

A new module `p3_sumcheck::ring_switch` implementing the ring-switching reduction of [eprint 2024/504](https://eprint.iacr.org/2024/504) (Diamond–Posen): Construction 3.1, with Remark 3.3 (the information-loss argument), Remark 3.4 (the verifier's equality-element recurrence) and Theorem 3.5 (security).

Given a multilinear `t` over a small field `F` in `ℓ` variables, its packing `t'` over an extension `EF`, and a claim `t(r) = s`, it reduces to a claim `t'(r') = s'` about the packed polynomial alone, using one degree-2 sumcheck of `ℓ' = ℓ − κ` rounds where `κ = log2([EF:F])`. Discharging `t'(r') = s'` against a commitment to `t'` is the caller's business, and is not part of this PR.

The module is generic over `F: Field, EF: ExtensionField<F>` and lives in `p3-sumcheck` rather than in a binary crate: nothing in the construction is characteristic-2 specific. `[EF:F]` must be a power of two, and a degree-5 extension is rejected rather than silently mishandled. No new dependency edge — `p3-sumcheck` already depended on `p3-field`, `p3-challenger`, `p3-multilinear-util` and `p3-maybe-rayon`.

### Public API

- `prove_ring_switch`, `verify_ring_switch`, `RingSwitchProof`, `RingSwitchError`
- `ring_switch::tensor::TensorAlgebra` — an element of `EF ⊗_F EF` as a `d × d` matrix over `F`, with the column and row readings, `φ0`/`φ1` scaling, and a checked `TryFrom<Vec<F>>` that `Deserialize` routes through
- `ring_switch::packing::{pack, packed_vars, compute_s_hat}`
- `ring_switch::weights::{batched_weights, batch_rows}`
- `ring_switch::equality::equality_element`

### Protocol

- **Packing.** Evaluation-vector entry `d·w + j` becomes basis coefficient `j` of packed element `w`, so the packed variables are the **suffix** ones — Plonky3 indexes hypercubes MSB-first, the opposite of the paper's convention. Getting this backwards yields a reduction that is self-consistent and wrong, so it is pinned as a test rather than a comment.
- **Prover message.** `ŝ = Σ_w eq̃(r_high, w) ⊗ t'(w)`, sent as its `d²` base coefficients.
- **Column check.** The verifier tests the incoming claim against the `eq̃(·, r_low)`-combination of `ŝ`'s columns.
- **Row reading.** The sumcheck's initial sum is *derived* by the verifier from `ŝ`'s rows batched against `κ` fresh challenges, never sent.
- **Final check.** `sum == A(r') · s'`, with `A(r')` recomputed in `O(ℓ'·d)` from the tensor-algebra equality element. `eq̃(X, Y) = Π_i (1 − X_i − Y_i + 2·X_iY_i)`, so Remark 3.4's recurrence runs in any characteristic once the cross term is carried: `e ← e − φ0(a)·e − φ1(b)·e + 2·φ0(a)φ1(b)·e`. In characteristic 2 the cross term vanishes and the signs collapse, giving the three-term form the remark states there. One routine, no characteristic dispatch, and no exponential branch.
- **Transcript**, identical on both sides: `r`, then `ŝ`'s base coefficients, then the `κ` batching challenges, then the sumcheck rounds, then `final_eval`. Binding `r` inside the reduction is what ties a proof to the point it was produced for; the verifier's `A(r')` test is `F`-linear in each coordinate of `r_high`, so an unbound `r` would let a prover shift `r_high` into that map's kernel and have a false input claim accepted alongside a true surviving claim.
- **Rejections:** a malformed `ŝ`, non-empty PoW witnesses (this reduction never grinds), a claim that disagrees with the columns, a failed round, a failed final check. Both structural rejections happen before the challenger is touched. The round count comes from `r`, which the verifier owns.
- **Soundness.** `(κ + 2ℓ')/|EF|`, recorded in the module documentation and wired into **no** estimator: nothing here touches `p3-security`. The docs also state that this is a reduction and not a filter — a false input claim survives as a false surviving claim — and that the module offers no grinding knob, so a composing protocol over a small `EF` must supply one itself.

### Other changes

- `binary-field/src/extension.rs`: `BasedVectorSpace::flatten_to_base` override inside `binary_tower_extension!`. The tower basis coincides with the byte layout, so the coefficients of a whole buffer are its bytes in order; `p3-field`'s default heap-allocates once per element. `reconstitute_from_base` keeps the default — a `Vec` of the lower type is allocated at alignment 1 and the upper type needs more, so that direction cannot hand the buffer back as it stands, and the default already parallelises the grouping it has to do.
- `sumcheck/Cargo.toml`: `p3-binary-field`, `p3-keccak` and `serde_json` as dev-dependencies.
- `sumcheck/src/table.rs`: one `assert_eq!(…, &[])` now needs an explicit `&[] as &[usize]`. `TensorAlgebra`'s derived `PartialEq` adds another candidate impl in the crate, and the bare empty slice literal no longer infers (`E0282`/`E0283`).
- Two bench targets, so every number below can be re-run: `sumcheck/benches/ring_switch.rs` (`pack`, `prove_ring_switch`, `verify_ring_switch`) and a `flatten_to_base` group in `binary-field/benches/arithmetic.rs`.

## Benchmarks

`aarch64-apple-darwin`, 14 cores, criterion, `--features parallel`. Reproduce with:

```
cargo bench -p p3-sumcheck --features parallel --bench ring_switch
cargo bench -p p3-binary-field --bench arithmetic -- flatten
```

On x86-64 both need `-C target-feature=+pclmulqdq`, without which `GF(2^128)` multiplication falls back to recursive tower multiplication and the numbers mean nothing.

**Against the base branch** — the one path this PR changes rather than adds, flattening `2^20` elements:

| `flatten_to_base`, `BinaryField128` → `BinaryField8` | time |
|---|---|
| `p3-field` default (base) | 22.49 ms |
| tower override (this PR) | **1.44 ms** |

**The reduction itself**, `F = BinaryField8`, `EF = BinaryField128` (`κ = 4`):

| `ℓ` | `ℓ'` | `pack` | prove | verify | proof |
|---|---|---|---|---|---|
| 20 | 16 | 172 µs | 6.47 ms | 44.9 µs | 256 `F` + 33 `EF` = 784 B |
| 22 | 18 | 336 µs | 16.34 ms | 49.6 µs | 256 `F` + 37 `EF` = 848 B |
| 24 | 20 | 699 µs | 53.20 ms | 54.3 µs | 256 `F` + 41 `EF` = 912 B |

The verifier is `O(ℓ'·d)` and near-flat in `ℓ`. The proof is `ŝ` plus two `EF` per round plus `s'`.

## Tests

34 unit tests and 4 integration tests. Every algebraic identity is pinned against an independent oracle: `ŝ`'s columns against `t` evaluated directly, its rows against the `A`-weighted sums, the equality-element recurrence against the definition summed over the hypercube, and the whole reduction against `t(r)` by brute force.

- The production pair `GF(2^8) → GF(2^128)` (`κ = 4`), and the smallest binary pair `GF(2^8) → GF(2^16)` (`κ = 1`), which reaches `ℓ' = 0` — an empty sumcheck — at `ℓ = 1`.
- `BabyBear` with its degree-4 binomial extension, end to end. The equality recurrence is additionally checked against the hypercube-summed definition over that pair at `ℓ' = 0..5`, where the cross term is what carries the agreement — dropping it would leave the two sides equal only in characteristic 2.
- Negative tests: a wrong claim, a perturbed `ŝ`, a perturbed `s'`, a perturbed round message, a proof replayed at a different `r_low` and at a different `r_high`, a proof carrying PoW witnesses, a wrong round count, a malformed `ŝ`, a short `r''` on both consumers, a non-power-of-two extension degree, and a point shorter than the packed variables on both sides.
- One test asserts that an adaptive prover who tampers with `ŝ` *before* proving and then runs a genuine sumcheck from the sum that implies is **accepted**, with a false surviving claim. That is the reduction working as documented rather than a gap: such a prover is caught only when the caller discharges `s'` against a commitment to the real packed polynomial.

Green on `cargo test --workspace`, clippy with `--all-features` and with `--features parallel`, nightly `fmt`, `cargo sort --grouped`, `taplo`, rustdoc with `-D warnings`, and the `thumbv7em-none-eabi` no-std build.

## Deliberately not here

The low-degree test, the fold over the additive domain, and the `MultilinearPcs` impls — part 2. No bytes are pinned by this PR (no fixture, no proof serialization), so the wire-representation question stays open at zero cost.


tcoratger (commented):
Really nice piece of work — the algebra is right, and I say that having re-derived it rather than pattern-matched it. Below is what I checked, then eleven comments inline. **No blockers: nothing here is a soundness, safety, or correctness defect.** Everything I found is performance or coverage, but two of the perf items are large and one of them is this module's own dominant prover loop, so I'd land those before merging.

## What I verified independently

So you know what "no blockers" actually rests on. Writing `r = (r_high, r_low)` with `|r_high| = ℓ'`, `|r_low| = κ`, and `m[u][v]` the coefficient of `β_u ⊗ β_v`:

```text
element = Σ_v col_v ⊗ β_v   ⟹  (a ⊗ 1)·element = Σ_v (a·col_v) ⊗ β_v   ⟹  φ0 = scale_columns  ✓
element = Σ_u β_u ⊗ row_u   ⟹  (1 ⊗ b)·element = Σ_u β_u ⊗ (b·row_u)   ⟹  φ1 = scale_rows     ✓

col_v(ŝ) = Σ_w eq̃(r_high,w)·t(w,v)      ⟹  Σ_v eq̃(v,r_low)·col_v = t(r)        ✓ column check
row_u(ŝ) = Σ_w A_{w,u}·t'(w)            ⟹  Σ_u eq̃(u,r'')·row_u = Σ_w A(w)t'(w)  ✓ initial sum
row_u(e) = Σ_w A_{w,u}·eq̃(w,r')         ⟹  Σ_u eq̃(u,r'')·row_u = A(r')          ✓ final check
```

with `e = eq̃(φ0(r_high), φ1(r')) = Σ_w eq̃(r_high,w) ⊗ eq̃(w,r')`, which is what buys the `O(ℓ'·d)` verifier instead of `O(2^ℓ')`.

Also checked, and all correct:

- **Transcript symmetry.** `r`, then `ŝ`'s base coefficients, then `κ` challenges, then the rounds, then `final_eval` — byte-identical on both sides. Critically `r''` is sampled *strictly after* `ŝ` is observed; drawing it first would be fatal and it isn't.
- **Packing orientation.** `Poly` indexes MSB-first, so `reconstitute_from_base` consumes the **suffix** variables and `split_at(ell_prime)` correctly puts them in `r_low`. `packing_consumes_the_suffix_variables` pins it, which is the right call — as your description says, getting this backwards yields something self-consistent and wrong.
- **No panic on proof-supplied data.** All three structural rejections are present and the two that must precede the challenger do: `is_well_formed`, `pow_witnesses.is_empty()` (which closes the gap that `verify_rounds` leaves open when `pow_bits == 0`), and the round count from `r`. I went looking for an index a malicious proof could push out of range and there isn't one.
- **Degenerate shapes.** `ℓ' = 0` is covered by your `binary_8_into_16` sweep. `κ = 0` (`EF = F`, which type-checks via `impl<F: Field> ExtensionField<F> for F`) is not covered, but I ran it and it works — see the comment on the integration tests.
- **Soundness bound.** `(κ + 2ℓ')/|EF|` decomposes exactly as you say: Schwartz–Zippel on a κ-variable multilinear for the batching draw, `2ℓ'` for ℓ' degree-2 rounds.
- **The `A(r')` linearity argument in the module docs is right**, and it's the sharpest thing in the PR. See my inline note — I want to make sure it survives future edits.
- **The `table.rs` one-liner.** I reverted it out of curiosity and the crate fails with `E0282`/`E0283`. Your stated cause is real, the annotation is genuinely required.

## Measurements

All numbers below are mine, on x86-64, 32 cores, `-C target-cpu=native` (so `pclmulqdq` is live — I checked `rustc --print cfg`), `--features parallel`. Different machine from your aarch64 figures, so read the *ratios*, not the absolutes.

Prover breakdown, `GF(2^8) → GF(2^128)`:

| | ℓ = 20 | ℓ = 22 |
|---|---|---|
| `prove_ring_switch` | 6.60 ms | 14.14 ms |
| `batched_weights` | **2.42 ms (37 %)** | **7.39 ms (52 %)** |
| `compute_s_hat` | 1.07 ms (16 %) | 2.81 ms (20 %) |
| `packed.clone()` | 17.8 µs | 52.3 µs |

Two things fall out. `batched_weights` is the single largest component and takes the *growing* share — that's comment 1, worth ~1.6–2x on the whole prover. And `packed.clone()` is genuinely free, so please ignore any instinct to remove it; I checked before writing it up.

## Test coverage

Genuinely thorough on the negative paths — one deterministic test per named branch is the right shape there, better than a generator, and `an_adaptive_cheat_on_s_hat_is_accepted_with_a_false_surviving_claim` is exactly how a documented limitation should be encoded: it will fail loudly if someone later "fixes" the reduction into a filter. Three gaps, each with an inline comment:

1. **`compute_s_hat`'s multi-chunk path is never exercised** — `S_HAT_CHUNK = 4096`, and the largest `ℓ'` anywhere in the suite is 4. The parallel-accumulate-then-fold is only ever run with one chunk. I wrote a test that reaches it and it passes, so this is coverage, not a bug.
2. **`κ = 0`** — works, untested.
3. **No proptest** — two places where one earns its keep, and several where it wouldn't.

## Questions

- The byte-table win in comment 1 needs an "is `F` small enough to enumerate" predicate. Is there an existing idiom for that (`F::ORDER_U64`, a bit-width const, a marker trait), or would this be the first place to need one? That's the only non-mechanical part of the fix and it's arguably a `p3-field` decision.
- `new_unpacked` leaving `ExtensionPacking` throughput on the table: given `baby_bear_into_degree_four` is a supported, tested instantiation rather than a curiosity, is the packed path deferred to part 2 or declined outright? The doc comment currently reads as an observation; worth making it a decision.
- I checked the `(κ + 2ℓ')/|EF|` decomposition against the construction rather than against the printed Theorem 3.5. It's self-consistent. If the paper's constant differs anywhere, that's the one number in the module docs I haven't reduced to the source.

**Bottom line: merge after comments 1 and 2.** The reduction is correct and unusually well documented — the docs are the best part of this PR, not padding on it — but as it stands the prover spends 37–52 % of its time in a loop that can be ~48x faster, and the one existing path the PR set out to optimise stops 11x short of the invariant its own comment states.

sumcheck/src/ring_switch/weights.rs
```diff
@@ -0,0 +1,167 @@
...
+        .as_slice()
+        .par_iter()
+        .map(|e| {
+            BasedVectorSpace::<F>::as_basis_coefficients_slice(e)
+                .iter()
+                .zip(eq_batch.as_slice())
+                .map(|(&a, &w)| w * EF::from(a))
+                .sum()
+        })
+        .collect();
```
tcoratger, line 62: ### 1. [MAJOR] This is the prover's dominant loop, and it's leaving ~48x on the floor

Per hypercube point this does `d` full `EF × EF` multiplications — `2^ℓ' · d` of them, i.e. **16.7 M `GF(2^128)` carry-less multiplies at ℓ' = 20**. Measured against the whole prover:

| | ℓ = 20 | ℓ = 22 |
|---|---|---|
| `prove_ring_switch` | 6.60 ms | 14.14 ms |
| `batched_weights` | **2.42 ms (37 %)** | **7.39 ms (52 %)** |

Largest single component, and it takes the growing share as ℓ rises.

The thing to notice: `a` is a **single byte**, and `eq_batch` is *fixed* across all `2^ℓ'` points. So the loop is recomputing `eq_batch[u] · c` for `c` drawn from a 256-element set, `2^ℓ'` times over. Tabulate it once — `d × 256` `EF` entries = 64 KiB, comfortably L2-resident — and the inner loop becomes `d` loads and `d − 1` XORs:

```text
  ℓ' = 16   current 317.8 ms   table  6.6 ms   → 48.3x    (loop only, single-threaded)
  ℓ' = 18   current   1.28 s   table 25.7 ms  → 49.9x
```

That's a **~1.6–2x speedup of the entire prover** from roughly fifteen lines.

**Two independent steps.**

*(a) Free and fully generic — worth doing regardless of (b).* Replace the loop with the `Algebra` trait's own primitive:

```rust
let weights: Vec<EF> = eq
    .as_slice()
    .par_iter()
    .map(|e| {
        <EF as Algebra<F>>::batched_linear_combination(
            eq_batch.as_slice(),
            BasedVectorSpace::<F>::as_basis_coefficients_slice(e),
        )
    })
    .collect();
```

`batched_linear_combination` gives you tree summation (shorter dependency chain than `.sum()`) chunked to hit SIMD-specialised overrides, and it recovers the `Mul<F>` specialisation that `EF::from(a)` currently discards. Neutral on the binary tower — `EF * F` expands to `EF * EF::from(F)` there anyway, `field/src/op_assign_macros.rs:172`, so no regression — but **1.5x** on `BinomialExtensionField<BabyBear, 4>`, where `Mul<A>` routes to `ext_base_mul` (`D` base muls) instead of a full `ext_mul` (~`D²` + reduction), `field/src/extension/binomial_extension.rs:471`. Measured 8.77 → 6.04 ms at ℓ' = 20, 36.2 → 24.1 ms at ℓ' = 22.

I applied exactly this in a worktree: all 34 unit + 4 integration tests pass unchanged.

*(b) The big one.* Add the byte-table path, gated on `F` being small enough to enumerate, falling back to `batched_linear_combination` otherwise:

```rust
// eq_batch[u] * c for every c in F, built once for all 2^ℓ' points.
let mut tbl = vec![EF::ZERO; d * F_ORDER];
for u in 0..d {
    for c in 0..F_ORDER {
        tbl[u * F_ORDER + c] = eq_batch.as_slice()[u] * EF::from(F::from_int(c));
    }
}
```

then index it by each coefficient. Generically `d` scalar multiplies per point *is* optimal — I convinced myself there's no factorisation to exploit, since the map `x ↦ Σ_u eq_batch[u]·x_u` is `F`-linear but not multiplicative, so it doesn't commute with the eq-table doubling recurrence. Tabulation is a small-base-field specialisation, and `GF(2^8)` is the production base field, so it's the specialisation that matters.

The predicate is the only judgement call here — see the question in my summary.

**Verify:** `cargo bench -p p3-sumcheck --features parallel --bench ring_switch`; the `prove` group should drop by roughly the fraction in the table above.

binary-field/src/extension.rs
```diff
@@ -100,6 +101,19 @@ macro_rules! binary_tower_extension {
...
+            #[inline]
+            fn flatten_to_base(vec: Vec<Self>) -> Vec<$lower> {
+                let mut out = Vec::with_capacity(
+                    vec.len() * <$upper as BasedVectorSpace<$lower>>::DIMENSION,
+                );
+                for x in &vec {
+                    out.extend_from_slice(x.as_basis_coefficients_slice());
+                }
+                out
+            }
```
tcoratger: ### 2. [MINOR] The doc comment states the invariant that licenses the fast path — and then the code doesn't take it

```rust
/// The tower basis coincides with the byte layout, so the coefficients of the whole
/// buffer are its bytes read in order.
```

If that's true — and it is, `as_basis_coefficients_slice` a few lines up proves it — then the *whole* `Vec` is already one `&[$lower]`, and flattening is one `memcpy`. Not `2^20` calls to `extend_from_slice`, each re-checking capacity it can't possibly need.

Measured, `2^20` `BinaryField128 → BinaryField8`:

```text
  per-element extend_from_slice   4.844 ms   (this PR)
  whole-buffer reinterpret         431 µs    → 11.2x
  Vec::clone (memcpy floor)        330 µs
```

431 µs against a 330 µs `memcpy` floor: the reinterpret is essentially free. The loop isn't. You captured most of the win over `p3-field`'s default (which is the real point of the override, and it's a good catch), but there's another 11x sitting right there.

**Fix.** The safety argument is *already written down* immediately above, at lines 52–66 — little-endian const-asserted, `$lower` fills its backing integer so every chunk is canonical, size and alignment const-asserted. All of it extends verbatim to a whole `Vec<$upper>`, which is contiguous and padding-free:

```rust
/// The tower basis coincides with the byte layout, so the whole buffer already *is* its
/// coefficients: flattening is one copy, not one per element.
#[inline]
fn flatten_to_base(vec: Vec<Self>) -> Vec<$lower> {
    let d = <$upper as BasedVectorSpace<$lower>>::DIMENSION;

    // SAFETY: exactly the argument in `as_basis_coefficients_slice` above, extended to
    // the whole buffer: a `Vec<$upper>` is `vec.len()` contiguous, padding-free `$upper`
    // values, each of which is `d` contiguous `$lower` values in tower-basis order. The
    // const assertions there cover the endianness, size, alignment and canonicity this
    // relies on. The slice borrows `vec`, which outlives it.
    let flat: &[$lower] =
        unsafe { slice::from_raw_parts(vec.as_ptr().cast::<$lower>(), vec.len() * d) };
    flat.to_vec()
}
```

`slice` is already imported at the top of the file, so this needs no new `use`.

One more thing worth doing while you're here: your PR description explains beautifully why `reconstitute_from_base` *can't* take the same shortcut — a `Vec` of the lower type is allocated at align 1, the upper type needs more, and deallocating with the wrong `Layout` alignment is UB. That reasoning belongs in the code, not only in the PR body. It's exactly the kind of thing someone "optimises" six months from now.

**Verify:** `cargo bench -p p3-binary-field --bench arithmetic -- flatten`. Your existing `flatten_and_reconstitute_round_trip` and `reconstitute_is_exhaustive_over_the_byte_level` already pin the byte order this depends on — and I like that the first one asserts every coefficient position is nonzero somewhere in the fixture. That's the guard that makes the round-trip test actually mean something.

sumcheck/src/ring_switch/equality.rs
```diff
@@ -0,0 +1,156 @@
...
+        // `φ0(2a)·φ1(b)·e`; the two scalings commute, so either order gives the cross term.
+        let mut cross = e.clone();
+        cross.scale_columns(a.double());
+        cross.scale_rows(b);
+
+        e += minus_high;
+        e += minus_prime;
+        e += cross;
+    }
+    e
```
tcoratger: ### 3. [MINOR] Expanding `eq̃` here costs three clones and a paragraph of documentation

The recurrence uses `1 − X − Y + 2XY`, so it needs four terms, three `clone()`s, and the doc comment above needs to explain a cross term and what happens to it in characteristic 2.

But `eq̃`'s per-variable factor is `X·Y + (1 − X)(1 − Y)` — the form used everywhere else in this repo, including `Point::eval_eq` at `multilinear-util/src/point.rs:118` — and since `φ0` and `φ1` commute, it lifts to the tensor algebra directly:

```rust
let mut e = TensorAlgebra::one();
for i in 0..r_high.num_variables() {
    let (a, b) = (r_high[i], r_prime[i]);

    // eq(a, b) = a*b + (1 - a)*(1 - b), lifted through the commuting embeddings.
    let mut agree = e.clone();
    agree.scale_columns(a);
    agree.scale_rows(b);

    e.scale_columns(EF::ONE - a);
    e.scale_rows(EF::ONE - b);
    e += agree;
}
e
```

One clone instead of three, same number of scalings, and the whole "in characteristic 2 the last term vanishes and the first two lose their signs" caveat in the doc comment simply disappears — in characteristic 2, `1 − a` *is* `1 + a` and nothing special happens. One routine, no characteristic dispatch, which is what the PR title is already going for; this just gets there with less machinery.

Measured:

```text
  ℓ' = 16   4-term 33.2 µs   2-term 20.9 µs   → 1.59x
  ℓ' = 20   4-term 40.9 µs   2-term 25.4 µs   → 1.61x
```

Against your reported 44.9–54.3 µs total verifier time, `equality_element` is roughly **half the verifier**, so this is a ~25 % verifier speedup.

I applied it: all 34 unit tests pass, including `the_recurrence_matches_the_definition_in_odd_characteristic` over `BabyBear`/degree-4 at `ℓ' = 0..5` — which is precisely the test that would fail if the factorisation only held when `2 = 0`. So the odd-characteristic guard you built is doing its job here.

**One more layer, if you want it.** The residual ~21 µs is allocation, not arithmetic. `scale_columns` calls `self.columns()`, which heap-allocates a `Vec<EF>`; `scale_rows` additionally does one `to_vec()` per row. At `d = 16` that's ~35 small allocations per variable, ~700 for a real `ℓ'`. Both can write straight back into `coeffs` without materialising anything, since `from_basis_coefficients_fn` returns by value — see my comment on `tensor.rs`.

sumcheck/src/ring_switch/packing.rs
```diff
@@ -0,0 +1,194 @@
...
+pub fn pack<F: Field, EF: ExtensionField<F>>(t: &Poly<F>) -> Poly<EF> {
+    let kappa = packed_vars::<F, EF>();
+    assert!(
+        t.num_variables() >= kappa,
+        "packing needs at least the {kappa} variables one packed element absorbs, got {}",
+        t.num_variables()
+    );
+    Poly::new(<EF as BasedVectorSpace<F>>::reconstitute_from_base(
+        t.as_slice().to_vec(),
+    ))
```
tcoratger, line 45: ### 4. [MINOR] `to_vec()` copies the evaluations `reconstitute_from_base` is about to consume anyway

`reconstitute_from_base` takes the `Vec` **by value** but immediately `par_chunks_exact`-es it into a fresh allocation (`field/src/field.rs:542`), so nothing ever reads this copy — it's a pure `2^ℓ`-byte memcpy on the way to being dropped.

```text
  ℓ = 20   &Poly + to_vec  548 µs   owned Vec  249 µs (incl. 15 µs clone)   → 2.2x
  ℓ = 24   &Poly + to_vec 4.13 ms   owned Vec 3.16 ms (incl. 242 µs clone)  → 1.3x
```

`Poly::into_evals()` already exists (`multilinear-util/src/poly/mod.rs:168`):

```rust
pub fn pack<F: Field, EF: ExtensionField<F>>(t: Poly<F>) -> Poly<EF> {
    let kappa = packed_vars::<F, EF>();
    assert!(
        t.num_variables() >= kappa,
        "packing needs at least the {kappa} variables one packed element absorbs, got {}",
        t.num_variables()
    );
    Poly::new(<EF as BasedVectorSpace<F>>::reconstitute_from_base(
        t.into_evals(),
    ))
}
```

Every call site in this PR already owns its `t` and has no use for it afterwards, so the signature change is free here. If you'd rather keep a borrowing form for callers that *do* want to keep `t`, add the owned one alongside — the point is that `pack(&t)` should make the caller write the clone, so the copy is visible at the call site instead of buried one frame down.

sumcheck/src/ring_switch/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+    // Send `ŝ`, binding the transcript to the base coefficients that cross the wire rather
+    // than to either of the readings derived from them.
+    let s_hat = compute_s_hat::<F, EF>(packed, &r_high);
+    challenger.observe_slice(s_hat.coefficients());
+
+    let r_batch = sample_batching_point::<F, EF, _>(challenger);
+
+    // `ℓ'` rounds on `h(X) = A(X) · t'(X)`, whose sum over the hypercube is the batched row
+    // reading of `ŝ`.
+    let weights = batched_weights::<F, EF>(&r_high, &r_batch);
```
tcoratger: ### 5. [MINOR] The prover builds the same `eq(r_high)` table twice, back to back

`compute_s_hat` (line 196) and `batched_weights` (line 203) each open with `Poly::<EF>::new_from_point(r_high.as_slice(), EF::ONE)` — `packing.rs:66` and `weights.rs:27` — and they're called seven lines apart with the same `r_high`. That's one redundant `2^ℓ'`-entry `EF` table:

```text
  ℓ = 20   eq table 126 µs of a 6.60 ms prove    (~2 %)
  ℓ = 22   eq table 520 µs of a 14.14 ms prove   (~4 %)
```

plus 1–16 MiB of avoidable allocation and memory traffic, which matters more than the 2–4 % suggests once you fix comment 1 and the loop becomes memory-bound rather than multiply-bound.

**Fix:** thread the table through — `compute_s_hat_with_eq(&eq, packed)` and `batched_weights_with_eq(&eq, r_batch)`, with the current signatures kept as thin wrappers so the public API stays exactly as documented and the tests don't move. Small and free, and it compounds with comment 1, which reads the same table.

sumcheck/src/ring_switch/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+//! The tensor element `ŝ` is the only prover message the two checks share: the verifier reads
+//! it by columns to test the incoming claim, and derives the sumcheck's initial sum from its
+//! rows itself rather than take that sum from the prover. That makes this a reduction, not a
+//! filter: a false input claim survives as a false surviving claim `t'(r') = s'` rather than
+//! being rejected outright, and a prover who tampers with `ŝ` and adapts the rest of the proof
+//! to the sum that tampering implies is not caught by [`verify_ring_switch`] — only by the
+//! caller discharging `s'` against a commitment to `t'`. What the row reading buys is the
+//! `κ/|EF|` term of the soundness bound below: a tampered `ŝ` shifts the derived initial sum
+//! away from the value an honest packing would produce, at the Schwartz–Zippel rate of the
+//! batching draw.
```
tcoratger, line 24: ### 6. [MINOR] The reduction contract holds only up to the stated error — worth naming the mechanism

> That makes this a reduction, not a filter: a false input claim survives as a false surviving claim `t'(r') = s'` rather than being rejected outright

This paragraph *is* the contract a composing caller relies on: *discharge `s'` against a commitment to `t'` and you catch everything the reduction lets past.* It isn't unconditional, and the one escape is worth naming precisely because it's the case where the commitment opening cannot help.

The final identity is `sum == A(r') · final_eval`. If **`A(r') = 0`**, then `s'` is *unconstrained*: the check degenerates to `sum == 0`, and a prover holding a false input claim can supply the honest `s' = t'(r')` — a **true** surviving claim, which the caller's commitment opening then happily accepts. The composed protocol accepts a false claim.

This is not a bug, and I want to be clear about that. The event is contained in your `2ℓ'/|EF|` term (it's a special case of "the reduced claim `h(r') = C_final` holds although the initial sum was false"), forcing it costs ≈`|EF|` grinding, and your Fiat–Shamir paragraph already accounts for per-attempt resampling honestly. It's a documentation precision point, raised only because these docs are otherwise *exact* about exactly this class of thing, and a caller who reads this paragraph and stops will take the contract as absolute.

**Suggested:** one clause, e.g. "— except with probability bounded by the error below, whose `2ℓ'/|EF|` term covers the degenerate `A(r') = 0` case, where the final check constrains `s'` not at all and a false input claim can survive as a *true* surviving claim, which is the one direction a caller's commitment opening cannot catch."

That last clause also happens to mirror the phrasing you already use two paragraphs down for the unbound-`r` case, so it should read as continuous.

sumcheck/src/ring_switch/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+//!
+//! `r` is bound by the reduction. Both sides observe it before `ŝ`, so the bound above holds
+//! unconditionally rather than resting on the caller having bound `r` first. Without that
+//! binding the proof would be replayable: `ŝ` is a function of `r_high` alone, so the same
+//! proof verifies verbatim at every `r_low`, and the verifier's one `r_high`-dependent test —
+//! `A(r')` in the final check — is multi-affine over `EF` in the coordinates of `r_high`. Move
+//! a single coordinate by `δ` and the change it induces in `A(r')` is `F`-linear in `δ`, so
+//! wherever that map is singular a prover can pick a nonzero `δ` from its kernel and have a
+//! false input claim accepted alongside a *true* surviving claim, which is the one direction a
+//! caller discharging `s'` against a commitment cannot catch.
```
tcoratger, line 52: ### 7. [praise] Please keep this argument intact

This is the sharpest thing in the PR and I want to flag it so it survives future edits, because it would be very easy for someone to read "we observe `r`, obviously" and delete the reasoning as verbose.

I checked it and it's right, including the part that's easy to get wrong. `A(r') = batch_rows(e, r'')`, and `batch_rows` is a *basis-dependent* reading — so although `φ0(δ)·e` is `EF`-linear in `δ` inside the tensor algebra, composing it with the row reading gives a map `EF → EF` that is only **`F`-linear**, i.e. a `d × d` matrix over `F`. Which can be singular. Which is exactly why "pick a nonzero `δ` from its kernel" is a real attack rather than a hypothetical, and why binding `r` is load-bearing rather than hygiene.

Getting from "multi-affine over `EF`" to "`F`-linear in `δ`, hence possibly singular" is the whole argument, and the docs make that step explicitly. `a_proof_does_not_verify_against_a_different_high_point` is the right test to pin it, and its doc comment correctly explains why the column arm passes by construction there so only the transcript can reject.

sumcheck/src/ring_switch/tensor.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    /// `φ1(a) · self`: scales each row, read as an `EF`, by `a`.
+    pub fn scale_rows(&mut self, a: EF) {
+        let d = Self::DIMENSION;
+        for (u, row) in self.rows().into_iter().enumerate() {
+            let scaled = (row * a).as_basis_coefficients_slice().to_vec();
+            self.coeffs[u * d..(u + 1) * d].copy_from_slice(&scaled);
+        }
+    }
```
tcoratger, line 166: ### 8. [MINOR] `columns()` / `rows()` allocate a `Vec` per call, and these two are the hot callers

Both `scale_columns` and `scale_rows` materialise an intermediate: `self.columns()` heap-allocates a `Vec<EF>`, and `scale_rows` additionally does one `to_vec()` **per row**. At `d = 16` that's ~17 allocations per `scale_rows` call.

`equality_element` calls into these four times per variable, so a real `ℓ'` runs ~700 small allocations in the verifier — which is most of what's left of its ~21 µs after the factorisation in comment 3. `batch_rows` also pays one `rows()` allocation per call, and it's called once by the prover and twice by the verifier.

Neither needs to allocate. `from_basis_coefficients_fn` returns by value, so both can read a column/row, scale it, and write the coefficients straight back:

```rust
/// `φ0(a) · self`: scales each column, read as an `EF`, by `a`.
pub fn scale_columns(&mut self, a: EF) {
    let d = Self::DIMENSION;
    for v in 0..d {
        let scaled = EF::from_basis_coefficients_fn(|u| self.coeffs[u * d + v]) * a;
        for (u, coeff) in scaled.as_basis_coefficients_slice().iter().enumerate() {
            self.coeffs[u * d + v] = *coeff;
        }
    }
}
```

and `scale_rows` the same shape, dropping the `to_vec()` in favour of writing coefficients into `coeffs[u * d + v]` directly. For `batch_rows`, a `rows_iter()` or a fold over `0..d` avoids the `Vec` entirely.

Keep `columns()` and `rows()` as they are — they're the readable API the tests use, and the tests are better for it. This is just about the internal callers not going through them.

(Unrelated, while I'm in this file: `TensorAlgebra::one()` being built from `EF::ONE`'s coefficients rather than hardcoded at `(0, 0)` is a nice touch, and the doc comment saying *why* — silent completeness failure in release if the basis convention ever changed — is exactly the right note to leave.)

sumcheck/src/ring_switch/packing.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/// The hypercube points one thread accumulates before its partial `ŝ` is combined with the
+/// rest. Large enough that the `DIMENSION²` accumulator is amortised, small enough to keep
+/// every core fed at the heights this reduction runs at.
+const S_HAT_CHUNK: usize = 1 << 12;
```
tcoratger, line 52: ### 9. [test coverage] `S_HAT_CHUNK = 4096` means the multi-chunk path is never exercised

The largest `ℓ'` anywhere in the suite is **4**: unit tests reach `ell = 7`, the integration sweep reaches `ell = 8`, both with `κ = 4`. So every test produces exactly *one* partial element, and the `par_chunks` → `collect` → sequential `fold` below runs only in the degenerate shape where the fold adds a single partial to zero.

That's the shape where a mis-zipped chunk pair or a wrong accumulator would live. And the `AddAssign` that fold uses (`tensor.rs:190`) `debug_assert`s the shapes and then `zip`s — so if they ever *did* disagree it would **silently truncate in release**, which is the failure mode you'd least want to discover in production.

The code is correct — I confirmed it rather than assumed it:

```rust
/// The multi-chunk parallel accumulation path: `S_HAT_CHUNK = 4096`, so this needs
/// `ell' >= 13` to produce more than one partial element.
#[test]
fn s_hat_agrees_across_the_chunk_boundary() {
    let ell = 17; // ell' = 13 -> 8192 entries -> 2 chunks
    let t = base_poly(ell, 99);
    let packed = pack::<F, EF>(&t);
    let (r_high, _) = Point::<EF>::rand(&mut SmallRng::seed_from_u64(98), ell)
        .split_at(ell - packed_vars::<F, EF>());

    let got = compute_s_hat::<F, EF>(&packed, &r_high);

    // Single-accumulator reference, no chunking.
    let eq = Poly::<EF>::new_from_point(r_high.as_slice(), EF::ONE);
    let mut want = TensorAlgebra::<F, EF>::zero();
    for (&e, &p) in eq.as_slice().iter().zip(packed.as_slice()) {
        want.add_exterior_product(e, p);
    }
    assert_eq!(got.coefficients(), want.coefficients());
}
```

Passes. So this is a coverage gap, not a defect.

**Better than adding that test, though:** make `S_HAT_CHUNK` `#[cfg(test)]`-overridable — say `1 << 2` under test — so that *every existing* `compute_s_hat` test crosses the boundary for free, rather than adding one `ℓ = 17` case that burns `2^17` field operations to reach a path the cheap tests should already be covering. Same trick would be worth it for any future chunk constant in this module.

sumcheck/tests/ring_switch.rs
```diff
@@ -0,0 +1,114 @@
...
+        assert_eq!(s_prime_v, packed.eval_ext::<F>(&r_prime_v), "ell = {ell}");
+    }
+}
+
+/// `packed_vars` rejects a non-power-of-two degree rather than computing nonsense.
+#[test]
+#[should_panic = "power-of-two extension degree"]
+fn a_degree_five_extension_is_rejected() {
+    let _ = packed_vars::<BabyBear, BinomialExtensionField<BabyBear, 5>>();
+}
```
tcoratger, line 148: ### 10. [test coverage] `κ = 0` works — worth four lines to keep it that way

`impl<F: Field> ExtensionField<F> for F` exists (`field/src/field.rs:1275`), so `pack::<F, F>` type-checks and someone downstream will instantiate it, if only by accident through a generic. It's the one shape where `d = 1` and every `d`-indexed loop runs exactly once — `ŝ` is a 1×1 matrix, `r''` is empty, `eq_low` is `[1]` — so an off-by-one in the column/row readings would go completely unnoticed.

I ran it and it passes at `ell = 0..4`, degenerating cleanly to a plain `eq`-weighted sumcheck:

```rust
/// The degenerate instantiation `EF == F`: `κ = 0`, no packing at all, and the reduction
/// collapses to a plain `eq`-weighted sumcheck. Every `d`-indexed loop runs exactly once
/// here, so it is the shape where an off-by-one has nowhere to show.
#[test]
fn identity_extension() {
    type F = BinaryField128;

    let mut rng = SmallRng::seed_from_u64(400);
    for ell in 0..=4 {
        let t = Poly::<F>::rand(&mut rng, ell);
        let packed = pack::<F, F>(&t);
        let r = Point::<F>::rand(&mut rng, ell);
        let s = t.eval_base(&r);

        let mut p_chal = wide_challenger();
        let (proof, r_prime_p, s_prime_p) = prove_ring_switch::<F, F, _>(&packed, &r, &mut p_chal);

        let mut v_chal = wide_challenger();
        let (r_prime_v, s_prime_v) =
            verify_ring_switch::<F, F, _>(&proof, &r, s, &mut v_chal).unwrap();

        assert_eq!(r_prime_p, r_prime_v, "ell = {ell}");
        assert_eq!(s_prime_p, s_prime_v, "ell = {ell}");
        assert_eq!(s_prime_v, packed.eval_ext::<F>(&r_prime_v), "ell = {ell}");
    }
}

/// A fresh transcript over `GF(2^128)` as its own base field.
fn wide_challenger() -> BinaryChallenger<BinaryField128, HashChallenger<u8, Keccak256Hash, 32>> {
    BinaryChallenger::from_hasher(Vec::new(), Keccak256Hash)
}
```

It slots naturally into the sweep this file already is — `κ = 4`, `κ = 1`, `κ = 2` non-binary, and then `κ = 0` as the floor. Your module docstring calls this file an "instantiation sweep… at each end of the design space", and `κ = 0` is genuinely the far end.

sumcheck/src/ring_switch/equality.rs
```diff
@@ -0,0 +1,156 @@
...
+        for ell_prime in 0..=4 {
+            let r_high = Point::<EF>::rand(&mut rng, ell_prime);
+            let r_prime = Point::<EF>::rand(&mut rng, ell_prime);
+            assert_eq!(
+                equality_element::<F, EF>(&r_high, &r_prime),
+                equality_element_reference::<F, EF>(&r_high, &r_prime),
+                "ell' = {ell_prime}"
+            );
+        }
+    }
```
tcoratger, line 97: ### 11. [test coverage] Two places a `proptest` earns its keep — and several where it wouldn't

Five hardcoded seeds per characteristic is thin for an identity that has to hold over the whole space, and this is the ideal proptest shape: the property is cheap, the oracle is *already written* (`equality_element_reference`), and a shrunk counterexample points straight at a basis-order or `φ0`/`φ1` slip.

```rust
proptest! {
    #[test]
    fn the_recurrence_matches_the_definition_everywhere(
        ell_prime in 0usize..6,
        seed: u64,
    ) {
        let r_high = Point::<EF>::rand(&mut rng, ell_prime);
        let r_prime = Point::<EF>::rand(&mut rng, ell_prime);
        prop_assert_eq!(
            equality_element::<F, EF>(&r_high, &r_prime),
            equality_element_reference::<F, EF>(&r_high, &r_prime),
        );
    }
}
```

`ell_prime < 6` keeps the exponential reference at `2^6` terms, so it's essentially free. Worth running over the `BabyBear` pair too, since that's the one where the characteristic-2 shortcut would be caught.

The second place: the **whole round trip** — random `ell in 1..=8`, random `t`, random `r`; assert `verify_ring_switch` accepts and `s' == packed.eval_ext(&r_prime)`. `2^8` field elements per case is nothing, and it exercises the composition of every convention at once (MSB-first packing, prefix binding, the `r_high`/`r_prime` pairing) rather than one at a time.

**Where I'd explicitly *not* reach for proptest**, since it's worth writing down so nobody adds it later: the negative paths. `ClaimMismatch`, `FinalCheck`, `MalformedTensor`, `NonEmptyPowWitnesses`, `RoundCountMismatch` and the two `should_panic`s are a finite set of named branches, and your table of one deterministic test each is clearer and localises better than any generator would. Same for `an_adaptive_cheat_on_s_hat_is_accepted_with_a_false_surviving_claim` — that one is a *specification*, and it reads as one.

One genuinely missing known-answer-ish case, low priority: `RingSwitchProof` itself has no serde round trip. `TensorAlgebra` has one, which is the part carrying the shape invariant, so the risk is small — but a deserialised `RingSwitchProof` is what a real verifier actually receives, and its `#[serde(bound(...))]` is currently unexercised.

sumcheck/src/ring_switch/weights.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Panics unless `r_batch` names exactly the `κ = log2([EF:F])` batching variables.
+pub fn batched_weights<F: Field, EF: ExtensionField<F>>(
+    r_high: &Point<EF>,
+    r_batch: &Point<EF>,
+) -> Poly<EF> {
+    assert_eq!(
+        1 << r_batch.num_variables(),
+        TensorAlgebra::<F, EF>::DIMENSION,
+        "r_batch must name exactly the kappa batching variables"
+    );
```
tcoratger, line 51: ### 12. [nit] `1 << r_batch.num_variables()` when the quantity itself is in scope

`packed_vars::<F, EF>()` *is* `log2(DIMENSION)`, so this can compare the two directly:

```rust
assert_eq!(
    r_batch.num_variables(),
    packed_vars::<F, EF>(),
    "r_batch must name exactly the kappa batching variables"
);
```

Same guard, no shift whose width is unchecked. Unreachable today — `r_batch` always comes from `sample_batching_point`, so at most 7 coordinates — but the assertion exists precisely to catch a caller who got it wrong, and `1 << 64` is the single input that makes the guard misbehave instead of firing. Belt-and-braces checks should be the one thing that can't be the bug.

Same in `batch_rows` at line 57. Your test doc comment on `batched_weights_rejects_a_short_r_batch` explains the stakes well — that the `zip` would truncate and both sides of the identity would be wrong in the same direction, so nothing downstream notices — which is a good reason to make the guard itself airtight.

sumcheck/src/ring_switch/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+pub mod equality;
+pub mod packing;
+pub mod tensor;
+pub mod weights;
+
+use equality::equality_element;
+pub use packing::{compute_s_hat, pack, packed_vars};
+use tensor::TensorAlgebra;
+use weights::{batch_rows, batched_weights};
```
tcoratger: ### 13. [nit] Inconsistent re-export shape

`packing` is re-exported (`pub use`), while `equality_element`, `batch_rows`, `batched_weights` and `TensorAlgebra` are private imports sitting next to `pub mod` declarations. All four are listed as public API in the PR description and all four *are* reachable — just at `ring_switch::tensor::TensorAlgebra` rather than `ring_switch::TensorAlgebra`.

Worth picking one. `TensorAlgebra` in particular appears in `RingSwitchProof`'s public field type, so a caller matching on the proof has to name it through the submodule path while `pack` sits at the top level.

sumcheck/src/ring_switch/equality.rs
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    let eq_high = Poly::<EF>::new_from_point(r_high.as_slice(), EF::ONE);
+    let eq_prime = Poly::<EF>::new_from_point(r_prime.as_slice(), EF::ONE);
+    let mut e = TensorAlgebra::zero();
+    for (&a, &b) in eq_high.as_slice().iter().zip(eq_prime.as_slice()) {
+        e += TensorAlgebra::exterior_product(a, b);
+    }
+    e
+}
```
tcoratger: since this is just a testing function maybe better to put it directly in the testing module no? just to maintain the main file clean

sumcheck/src/ring_switch/tensor.rs
```diff
@@ -0,0 +1,305 @@
...
+    #[cfg(test)]
+    pub(crate) fn perturb_for_test(&mut self, index: usize, delta: F) {
+        self.coeffs[index] += delta;
+    }
+
+    /// Drops all but the first `len` coefficients, to exercise the malformed-element path.
+    #[cfg(test)]
+    pub(crate) fn truncate_for_test(&mut self, len: usize) {
+        self.coeffs.truncate(len);
+    }
```
tcoratger, line 125: if this is only for tests maybe better to drop this?

sumcheck/src/ring_switch/weights.rs
```diff
@@ -0,0 +1,167 @@
...
+    let eq_batch = Poly::<EF>::new_from_point(r_batch.as_slice(), EF::ONE);
+    tensor
+        .rows()
+        .iter()
+        .zip(eq_batch.as_slice())
+        .map(|(&row, &w)| row * w)
+        .sum()
+}
```
tcoratger, line 90: Maybe we need to move this directly to the tensor algebra implementation, not sure?

sumcheck/src/ring_switch/weights.rs
```diff
@@ -0,0 +1,167 @@
...
+    let eq = Poly::<EF>::new_from_point(r_high.as_slice(), EF::ONE);
+    let eq_batch = Poly::<EF>::new_from_point(r_batch.as_slice(), EF::ONE);
+    let weights: Vec<EF> = eq
+        .as_slice()
+        .par_iter()
+        .map(|e| {
+            BasedVectorSpace::<F>::as_basis_coefficients_slice(e)
+                .iter()
+                .zip(eq_batch.as_slice())
+                .map(|(&a, &w)| w * EF::from(a))
```
tcoratger: not sure we need the EF::from here no?

sumcheck/src/ring_switch/weights.rs
```diff
@@ -0,0 +1,167 @@
...
+                .sum()
+        })
+        .collect();
+    Poly::new(weights)
+}
```
tcoratger, line 64: Maybe we need to move this directly to the Point implementation, not sure?

tcoratger (approved):

PR #2007, opened 2026-08-29
fix(security): account for out-of-domain point count in budget's OOD round

Nashtare:
## Summary
- `p3_security::budget::security_report`'s out-of-domain round bounded the DEEP-ALI degree as `(d + 1) · height`, implicitly assuming a single out-of-domain point per column.
- An AIR that also opens `next`-row rotations references a second out-of-domain point, which needs an extra `(combo − 1) · (d − 1)` term — omitting it understated the round's error (i.e. overstated security) by a fraction of a bit, shrinking as trace height grows.
- `AirShape` gains a `max_combo` field (mirroring the crate's existing `f64`-based `StarkAirParams::max_combo`), and the out-of-domain round now computes the full formula directly.
- `security/tests/budget_parity.rs` previously bounded this exact gap with an analytic tolerance function; with the fix the fixed-point and `f64` reference paths agree exactly, so that tolerance is replaced with the same tight equality check used by every other round.

## Test plan
- [x] `cargo test -p p3-security` (67 unit tests + `budget_parity` integration tests)
- [x] `cargo clippy -p p3-security --all-targets -- -D warnings`
- [x] `cargo fmt -p p3-security -- --check`

### DIFF
--- a/security/src/budget/mod.rs
+++ b/security/src/budget/mod.rs
@@ -101,16 +101,12 @@ pub const fn security_report(
 
     // The out-of-domain point is rejection-sampled outside the trace and LDE domains but is not
     // ground. A violated constraint leaves the DEEP-ALI identity a nonzero polynomial in that
-    // point of degree at most `(max_constraint_degree + 1) · height`. Lifting evaluates the
-    // shorter AIRs at powers of the same point, so all of them pay the maximum height.
-    let out_of_domain = round(
-        OUT_OF_DOMAIN_LABEL,
-        instance,
-        air.max_constraint_degree as u64 + 1,
-        instance.log_max_height,
-        0,
-        cap,
-    );
+    // point of degree at most `(max_constraint_degree + 1) · height + (max_combo − 1) ·
+    // (max_constraint_degree − 1)`: one factor of the identity's degree per out-of-domain point
+    // referenced, plus a flat term per extra point from the shared numerator. Lifting evaluates
+    // the shorter AIRs at powers of the same point, so all of them pay the maximum height.
+    let out_of_domain =
+        out_of_domain_round(instance, air.max_constraint_degree, air.max_combo, cap);
 
     // The DEEP quotient batches every committed column and out-of-domain point by powers of two
     // further challenges, giving a univariate whose degree is the number of batched terms.
@@ -182,6 +178,28 @@ const fn round(
     SecurityTerm::new(label, min(bits, cap))
 }
 
+/// Bounds the out-of-domain round: error `((d + 1) · H + (combo − 1) · (d − 1)) / |E|`, with `H`
+/// the trace height and `combo` the number of out-of-domain points referenced per column. Unlike
+/// [`round`], the extra `(combo − 1) · (d − 1)` term is additive rather than a multiple of `H`, so
+/// it is folded into the size directly instead of decomposed into a coefficient and a log-size.
+/// This round is always charged — there is no "no such round" case to special-case at zero.
+const fn out_of_domain_round(
+    instance: &InstanceShape,
+    max_constraint_degree: u32,
+    max_combo: u32,
+    cap: u64,
+) -> SecurityTerm {
+    let d = max_constraint_degree as u64;
+    let height = 1u64 << instance.log_max_height;
+    let extra_points = (max_combo as u64).saturating_sub(1);
+    let size = (d + 1) * height + extra_points * d.saturating_sub(1);
+
+    let error = fixed::ceil_log2(size);
+    let bits = instance.field_bits.saturating_sub(error);
+
+    SecurityTerm::new(OUT_OF_DOMAIN_LABEL, min(bits, cap))
+}
+
 const fn min(a: u64, b: u64) -> u64 {
     if a < b { a } else { b }
 }
@@ -217,6 +235,7 @@ mod tests {
         AirShape {
             num_composed_constraints: 531,
             max_constraint_degree: 9,
+            max_combo: 2,
             num_deep_terms: Some(130),
             lookup: LookupShape {
                 fractions_per_row: 27,
@@ -299,6 +318,7 @@ mod tests {
         let large = AirShape {
             num_composed_constraints: 4096,
             max_constraint_degree: 9,
+            max_combo: 2,
             num_deep_terms: Some(1024),
             lookup: LookupShape {
                 fractions_per_row: 64,
--- a/security/src/budget/mod.rs
+++ b/security/src/budget/mod.rs
@@ -14,7 +14,11 @@
 //!
 //! Every bound is stated as `error ≤ coefficient · size / |E|`, which in bits is
 //! `log2|E| − log2(coefficient) − log2(size)`. Coefficients round up and the field size rounds
-//! down, so each round is understated rather than overstated.
+//! down, so each round is understated rather than overstated. The out-of-domain round is the one
+//! exception to that framing: part of its size is additive rather than a multiple of the trace
+//! height, so it takes the logarithm of the whole product at once. It states the same bound as
+//! [`crate::deep::deep_ali_error`], the `f64` mirror of the same equation, and the two are meant
+//! to be read together.
 //!
 //! # Folding accounting
 //!
@@ -101,10 +105,13 @@ pub const fn security_report(
 
     // The out-of-domain point is rejection-sampled outside the trace and LDE domains but is not
     // ground. A violated constraint leaves the DEEP-ALI identity a nonzero polynomial in that
-    // point of degree at most `(max_constraint_degree + 1) · height + (max_combo − 1) ·
-    // (max_constraint_degree − 1)`: one factor of the identity's degree per out-of-domain point
-    // referenced, plus a flat term per extra point from the shared numerator. Lifting evaluates
-    // the shorter AIRs at powers of the same point, so all of them pay the maximum height.
+    // point of degree at most the larger of `max_constraint_degree · (height + max_combo − 1) +
+    // (height − 1)` (clearing the shared denominator lifts each of a constraint's degree-many
+    // trace factors by one per out-of-domain point referenced, and the quotient side contributes
+    // the rest) and `(c + 1) · height + max_combo − 1` (the degree Plonky3's own power-of-two
+    // quotient chunking induces, binding whenever `c` exceeds `max_constraint_degree`). Lifting
+    // evaluates the shorter AIRs at powers of the same point, so all of them pay the maximum
+    // height.
     let out_of_domain =
         out_of_domain_round(instance, air.max_constraint_degree, air.max_combo, cap);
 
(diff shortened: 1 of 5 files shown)


Nashtare:
@tcoratger ideally I'd get another rc release once this lands into `main`, whenever you have some bandwidth to review

security/src/budget/mod.rs
```diff
@@ -182,6 +178,28 @@ const fn round(
     SecurityTerm::new(label, min(bits, cap))
 }
 
+/// Bounds the out-of-domain round: error `((d + 1) · H + (combo − 1) · (d − 1)) / |E|`, with `H`
+/// the trace height and `combo` the number of out-of-domain points referenced per column. Unlike
+/// [`round`], the extra `(combo − 1) · (d − 1)` term is additive rather than a multiple of `H`, so
+/// it is folded into the size directly instead of decomposed into a coefficient and a log-size.
+/// This round is always charged — there is no "no such round" case to special-case at zero.
```
tcoratger: Doc nit, but a load-bearing one since this is the only place in the crate that explains *why* the bound is what it is.

The prose here doesn't derive its own formula: *"one factor of the identity's degree per out-of-domain point referenced, plus a flat term per extra point from the shared numerator"* doesn't produce `(combo − 1)·(d − 1)` under any reading I could find.

The derivation that does give the reference bound is short enough to just write down. To check the DEEP-ALI identity at `z` you clear the common denominator `Π_i (x − z_i)`, so each of the `d` factors of a constraint is a trace polynomial of degree `<= k − 1` lifted to degree `<= k − 1 + combo`:

```text
d · (k + combo − 1)      from the d lifted factors
      + (k − 1)          from the quotient side
```

Stating it that way makes it obvious that the extra-point term is a multiple of `d`, not of `d − 1` — which is exactly the slip in the expression below.

Two small additions while you're here:

- Cross-link `[crate::deep::deep_ali_error]` from this doc and from `budget`'s module doc, so the next person editing either one can see at a glance they're meant to be the same equation. That's the drift this PR is paying for now, and a doc link is the cheapest thing that prevents a third copy.
- `/// This round is always charged — there is no "no such round" case to special-case at zero.` is a genuinely useful line, worth keeping as-is. Same for the note that the extra term is additive rather than a multiple of `H`; it's the reason this can't go through `round()`, and it's the one exception to the module doc's `error <= coefficient · size / |E|` framing, which might be worth one sentence up there.

security/src/budget/mod.rs
```diff
@@ -182,6 +178,28 @@ const fn round(
...
+const fn out_of_domain_round(
+    instance: &InstanceShape,
+    max_constraint_degree: u32,
+    max_combo: u32,
+    cap: u64,
+) -> SecurityTerm {
+    let d = max_constraint_degree as u64;
+    let height = 1u64 << instance.log_max_height;
```
tcoratger: Separate from the formula: materialising the height here is a regression against `main`.

The old path only ever saw `log_max_height` as an addend in log space (`ceil_log2(coefficient) + from_bits(log_size)`), so it could not overflow. This line multiplies it into a `u64`, `InstanceShape::log_max_height` is a `pub u32` with no validation anywhere on the path, and in release the wrap is silent. `1u64 << 64` is a masked shift too, so it yields `height = 1` rather than failing.

Same probe on both branches, release, `d = 9`, 128-bit field:

| `log_max_height` | `main` | this PR |
|---|---|---|
| 59 | 65.6781 | 65.6781 |
| 60 | 64.6781 | 64.6781 |
| 61 | 63.6781 | **66.0000** |
| 62 | 62.6781 | **65.0000** |
| 63 | 61.6781 | **125.0000** |
| 64 | 60.6781 | **123.8301** |

`10 · 2^63 mod 2^64 = 0`, so at `h = 63` `size` collapses to the flat `(combo − 1)(d − 1) = 8` term and the round reports 125 bits against a true 61.68.

That breaks two things the module promises: the module doc's *"each round is understated rather than overstated"*, and monotonicity in trace height.

Whether this is a robustness nit or actually reachable is the one thing I can't tell from here. The module doc's stated audience is *"a protocol that grades proof soundness inside a recursive verifier"* — if `log_max_height` is ever populated from the proof rather than from a pinned config, it's adversary-chosen and unvalidated. Either way it's cheap to close, and it composes with the formula fix above:

```rust
if instance.log_max_height >= 63 {
    return SecurityTerm::new(OUT_OF_DOMAIN_LABEL, 0);
}
let d = if max_constraint_degree == 0 { 1 } else { max_constraint_degree as u64 };
let height = 1u64 << instance.log_max_height;
let combo = if max_combo == 0 { 1 } else { max_combo as u64 };
let raw = (d as u128) * (height as u128 + combo as u128 - 1) + (height as u128 - 1);
let size = if raw > u64::MAX as u128 { u64::MAX } else { raw as u64 };
```

**One trap worth flagging on the test side.** I first assumed extending `security_is_monotone_in_trace_height` (`mod.rs:305`) from `6..=30` to `0..=64` would catch this. It doesn't — I tried it and it passes. `security_level()` is the *min* over rounds, and at `h = 61` the lookup round is already down at ~58 bits, so a bogus 66-bit OOD term is invisible behind it. The invariant has to be asserted per round:

```rust
#[test]
fn every_round_is_monotone_in_trace_height() {
    let mut previous = [u64::MAX; 7];
    for log_height in 0..=64u32 {
        let report = security_report(&params(), &instance(log_height), &air());
        for (term, prev) in report.terms().iter().zip(previous.iter_mut()) {
            assert!(term.bits <= *prev, "{} rose at log height {log_height}", term.label);
            *prev = term.bits;
        }
    }
}
```

On this branch:

```
out-of-domain rose at log height 61
test result: FAILED
```

With the guard above: green. That test is worth having on its own merits — it's a strictly stronger statement than the existing one, and it's the shape that catches *any* future round losing monotonicity, not just this one.

security/src/budget/mod.rs
```diff
@@ -182,6 +178,28 @@ const fn round(
...
+    let extra_points = (max_combo as u64).saturating_sub(1);
+    let size = (d + 1) * height + extra_points * d.saturating_sub(1);
```
tcoratger: This is the main one. The expression here doesn't reduce to the bound the crate states everywhere else, and it only *coincides* with it at `combo = 2`.

`deep.rs:27` and `stark.rs:302` both give the same thing:

```text
factor = d·(k + combo − 1) + (k − 1)
```

Expanding with `k = H`:

```text
reference = d·H + d·(combo−1) + H − 1  =  (d+1)·H + (combo−1)·d − 1
this line = (d+1)·H + (combo−1)·(d−1)  =  (d+1)·H + (combo−1)·d − (combo−1)

this line − reference = 2 − combo
```

So `combo = 1` is one larger (harmlessly conservative), `combo = 2` is exact, and every `combo >= 3` is `combo − 2` **smaller** than the truth — smaller `size`, smaller `ceil_log2`, more reported bits. Same direction as the bug the PR is fixing, just relocated from `combo = 2` to `combo >= 3`.

Measured on this branch, 128-bit field, `f64` column is `deep::deep_ali_error` capped the way `check_ood` caps it:

`d = 3, log_max_height = 6` (the `SMALL` vector):

| combo | budget (fixed) | `deep_ali_error` | delta |
|---|---|---|---|
| 1 | 119.999969 | 120.005647 | −0.005677 |
| 2 | 119.988739 | 119.988773 | −0.000034 |
| 3 | 119.977600 | 119.972094 | **+0.005506** |
| 4 | 119.966537 | 119.955606 | **+0.010932** |
| 8 | 119.923141 | 119.891476 | **+0.031666** |

`d = 9, log_max_height = 6` (the `LARGE` vector) shows the same sign flip at 3, smaller in magnitude (`+0.00216` .. `+0.01235`).

And `combo >= 3` isn't hypothetical — `uni-stark/src/security.rs:451` (`more_max_combo_decreases_or_holds_security`) already drives the `f64` twin of this at `max_combo = 8`.

Repro is one character, `OOD_MAX_COMBO = 3` in `budget_parity.rs:54`:

```
ood: 119.97760009765625 > 119.97209400343012 at h=6
```

(`combo = 1` fails too, on the safe side: `ood: gap 0.005677 >= tolerance, fixed drifted conservative`.)

I'd mirror `deep.rs` character for character rather than re-derive, so the two can't drift again:

```rust
let d = if max_constraint_degree == 0 { 1 } else { max_constraint_degree as u64 };
let height = 1u64 << instance.log_max_height;
let combo = if max_combo == 0 { 1 } else { max_combo as u64 };
let size = d * (height + combo - 1) + (height - 1);
```

Two things that come with that shape:

- **The `.max(1)` on `d` is not cosmetic.** `deep.rs:25` has `air.max_constraint_degree.max(1)` and this line doesn't, so `d = 0` currently gives `size = H` against the reference's `2H + combo − 2` — a full bit of overstated security. More importantly, under the mirrored expression `d = 0, H = 1` gives `size = 0` and `fixed::ceil_log2` asserts. With the clamps, `d >= 1` and `combo >= 1` force `size >= 1`, so the zero-assert stays unreachable.
- **`combo = 0`** is nonsense input either way; clamping it keeps `height + combo − 1` from underflowing once it's inside a subtraction.

I applied exactly those four lines locally: `budget_parity` is then *exact* in both assertion directions at `max_combo ∈ {1, 2, 3, 8}`, and all 67 unit tests still pass.

Two prose restatements of the formula need to follow it: this function's doc on L181–L183, and the comment at L104–L105.

security/src/budget/shape.rs
```diff
@@ -82,6 +82,11 @@ pub struct AirShape {
     pub num_composed_constraints: u32,
     /// Maximum constraint degree over all AIRs.
     pub max_constraint_degree: u32,
+    /// Maximum number of out-of-domain points referenced per committed column: `2` for an AIR
+    /// that opens `local` and `next` rotations, `1` if it has no transition constraint.
+    ///
+    /// Read by the out-of-domain round, independent of [`Self::num_deep_terms`].
+    pub max_combo: u32,
```
tcoratger: Small scoping question, mostly about what you want the contract to be.

This doc only contemplates `1` and `2`, and the expression in `out_of_domain_round` is only correct on exactly that range — but the field is a `u32`, and `StarkAirParams::max_combo` (the `f64` twin this mirrors) is already driven at `8` by `uni-stark/src/security.rs:451`. So a caller reasonably reads `u32` as "however many rotations my AIR opens" and gets a quietly optimistic number from `3` upward.

Two coherent answers, and I'd take the first:

- **General.** Use the `deep.rs` expression (see my note on `mod.rs`), and reword to "one per distinct out-of-domain point the AIR's rotations induce" rather than framing it as a `1`-or-`2` choice. Costs nothing and the field then means what its type says.
- **Deliberately scoped to `<= 2`.** Then say so here, and make it a hard error rather than a doc convention. `security_report` is `const fn`, so a plain `assert!(max_combo <= 2)` in `out_of_domain_round` is a *compile-time* failure for const callers and a loud one for the rest — much better than a number that's silently 0.03 bits high.

Also, a `1` should probably be described as "an AIR whose constraints all read a single row" rather than "no transition constraint" — `max_combo` counts distinct opening points, which is the quantity `deep_ali_error` consumes.

---

Unrelated, but worth noting for the rc you're asking for: `AirShape` is `pub` with all-`pub` fields and no `#[non_exhaustive]`, so adding `max_combo` breaks every downstream struct literal. Nothing in-tree constructs it outside `security/` (I checked — `uni-stark` only touches `StarkAirParams`), so it's cheap here; just make sure it goes out as a minor bump with a changelog line rather than a patch. If more shape fields are coming as the budget grows, `#[non_exhaustive]` plus a `Default` would spare downstreams the churn — though if explicit literal construction is the deliberate ethos (the module doc's *"the caller ... assembles these explicitly"* suggests it is), leaving it and just versioning it correctly is fine.

security/tests/budget_parity.rs
```diff
@@ -55,6 +50,9 @@ const LARGE: AirVector = AirVector {
 
 const PARAM_ROWS: &[(u32, u32, u32, u32)] = &[(27, 17, 12, 4), (100, 0, 0, 0), (150, 31, 31, 31)];
 
+/// Out-of-domain points referenced per column: `local` and `next` rotations.
+const OOD_MAX_COMBO: u32 = 2;
```
tcoratger, line 55: Removing `ood_tolerance` was the right call — but pinning the constant at `2` means the tight equality assertion is now only ever evaluated at the single value where the two formulas coincide (see my note on `budget/mod.rs`). The test that was meant to get stronger is the one that hides the gap.

The grid runs in a few milliseconds, so sweeping is free:

```rust
for max_combo in 1..=8u32 {
    let air_shape = AirShape { /* ... */ max_combo, /* ... */ };
    // and thread `max_combo` into check_ood's StarkAirParams
}
```

Two things about *where* to sweep, since they're not obvious:

- **Keep `log_max_height` starting at 6.** The divergence is a small-height phenomenon: the flat `(combo − 1)` term is a meaningful fraction of `size` at `H = 64` (gap `5.5e-3` bits, comfortably above `TIGHT_TOL = 1e-4`) and is below the `2^-16` fixed-point resolution by `h ~ 20`. A sweep starting at 20 would pass on the buggy code.
- **Keep `SMALL` in the loop.** Lower `d` makes the *relative* gap larger, so `d = 3` is the sensitive vector, not `d = 9`.

Two more coverage gaps while you're in here.

**Monotonicity in `max_combo`** — the fixed-point mirror of `uni-stark/src/security.rs:451`, which exists on the `f64` path and not on this one:

```rust
#[test]
fn security_is_monotone_in_max_combo() {
    let mut previous = u64::MAX;
    for max_combo in 1..=16u32 {
        let air = AirShape { max_combo, ..air() };
        let bits = security_report(&params(), &instance(6), &air).terms()[2].bits;
        assert!(bits <= previous, "out-of-domain rose at max_combo {max_combo}");
        previous = bits;
    }
}
```

**And a proptest for the parity direction itself.** Now that this round claims exact agreement rather than a bounded gap, "the fixed-point term never exceeds the `f64` reference" is a one-line property over a three-dimensional integer space, and a counterexample is by construction a proof that the budget overstates security:

> **Inputs:** `d in 1..=32`, `log_max_height in 0..=48`, `max_combo in 1..=16`.
> **Invariant:** the `out-of-domain` term `<= deep::deep_ali_error(..).bits().min(cap) + 1e-9`.
> **What a failure proves:** a parameter shape where the fixed-point budget reports more bits than the scalar reference — i.e. a recursive verifier accepting a configuration the `f64` path would reject.

Proptest rather than a vector table here for the same reason as the STIR one on #1998: the corners that matter are boundaries nobody enumerates by hand (`d = 0`, `combo = 1`, `log_max_height` near the word size), which is exactly what a shrinking generator finds. It would have found both findings in this review. `proptest` is already `workspace = true` and used in `keccak` / `baby-bear` / `merkle-tree`; `p3-security` just needs a `[dev-dependencies]` section.

tcoratger (commented):
Went through this one carefully — the out-of-domain round is the one term in the budget that is *supposed* to track `deep::deep_ali_error` exactly now that the tolerance function is gone, so I re-derived the DEEP-ALI degree bound from scratch and then measured both paths against each other rather than trusting either.

**The direction of the fix is right, and it buys more than the description suggests at small heights.** Base vs. this branch, same probe, release, `d = 9`, 128-bit field, `max_combo = 2` (`ood_raw` is the raw fixed-point term, so you can see exactly where the correction drops below resolution):

| `log_max_height` | `main` | this PR | delta |
|---|---|---|---|
| 0 | 124.678040 | **123.830032** | −0.848 |
| 4 | 120.678040 | **120.607651** | −0.070 |
| 8 | 116.678040 | **116.673538** | −0.0045 |
| 12 | 112.678040 | **112.677750** | −0.00029 |
| 16 | 108.678040 | **108.678009** | −0.00003 |
| >= 20 | 104.678040 | 104.678040 | 0 |

The cutoff is where `(d − 1) / ((d + 1)·H) < 2^-16`, i.e. `H > 52429`, so it lands right at `h ~ 16` as predicted. `security_level()` is unchanged (79) across the whole sweep — OOD is nowhere near binding at these params — so this is a tightness fix rather than a behaviour change, which is the right framing for the rc.

Other things I checked and am happy with:

- **Dropping the `pow_bits` plumbing is safe.** The old call site passed `0`, and `field_bits.saturating_sub(error)` is exactly the old `if error >= field_bits { 0 } else { field_bits - error }` branch.
- **Taking `ceil_log2` of the whole product** rather than `ceil_log2(coefficient) + from_bits(log_size)` is strictly tighter and stays on the conservative side of the true log — `fixed`'s own `log2_brackets_the_true_logarithm` covers that direction.
- **`ceil_log2`'s zero-assert is unreachable on the current expression**: `(d + 1) >= 1` and `height >= 1` force `size >= 1` regardless of `max_combo` and `d`, and both `saturating_sub`s are doing real work. (That becomes load-bearing under the fix I suggest below — noted inline.)
- 67 unit tests + `budget_parity` green here too, clippy clean.

**Two things I think want fixing before this lands**, both inline with the measurements. Short version:

**1. The new expression only agrees with `deep_ali_error` at `max_combo = 2`.** `(d+1)·H + (combo−1)·(d−1)` differs from the crate's own `d·(k + combo − 1) + (k − 1)` by exactly `2 − combo`, so `combo >= 3` reports *more* bits than the reference — the same direction as the bug this PR fixes, just relocated. `combo = 8` is already driven on the `f64` path by `uni-stark/src/security.rs:451`. One-character repro: `OOD_MAX_COMBO = 3` in `budget_parity.rs:54` gives `ood: 119.9776 > 119.9721 at h=6`.

**2. `(d + 1) * (1u64 << log_max_height)` overflows, and it is a regression against `main`.** The old path only ever saw the height as an addend in log space, so it could not overflow; this one multiplies, silently wraps in release, and reports **125 bits at `h = 63`** against a true 61.68. `main` is monotone and correct at every height.

Neither is hard. I have both fixed in a worktree — the mirrored formula plus a `u128`/guard — and with it the parity test is *exact* (both assertion directions, `TIGHT_TOL = 1e-4`) at `max_combo ∈ {1, 2, 3, 8}`, a new per-term monotonicity test passes over `h ∈ 0..=64`, and all 67 unit tests still pass. Happy to push the patch if that's easier than applying it yourself.

Everything else is doc/coverage polish, also inline. Nice fix — the underlying observation is exactly right, it just wants the reference formula rather than a re-derivation.

tcoratger (approved):

PR #2008, opened 2026-08-29
fix(sumcheck): unpack suffix-bound product polynomials

pucedoteth:
Two commits, kept separate for review. #2009 is folded in here as the second commit per @Nashtare's request.

---

# 1. `fix(sumcheck): unpack suffix-bound product polynomials`

`ProductPolynomial::new_packed` accepts `VariableOrder::Suffix` and then runs every round over the **wrong variable**.

`Poly::pack` groups `WIDTH` *consecutive* hypercube points into the lanes of one packed element, so the lanes hold the last `log2(WIDTH)` variables. Both suffix kernels bind the last variable of the slice they are handed — which is exactly what they document:

> `c_a` = `h(0)` = sum_{b in {0,1}^{n-1}} f(b, 0) * w(b, 0)

On a packed slice that slice-adjacent axis is `X_{n-1-log2(WIDTH)}`, not `X_{n-1}`. The variable a suffix round is meant to bind sits *inside* the lanes, where a slice-adjacent fold cannot reach it. The same applies to `fix_suffix_var_mut` in `compress`.

Nothing panics. The rounds still run and still reduce a sum, so the failure is silent — it just binds a different variable than the order names, producing a transcript that disagrees with the scalar path for the same polynomial.

Prefix binding is unaffected: `X_0` is the top bit of the packed element index in either representation.

### Impact

Every in-tree **production** caller uses `Prefix` (`whir/src/pcs/zk/prover/mod.rs`, `sumcheck/src/zk/prover/residual.rs`), so no shipped proof is affected. What is affected is the public API — `new_packed` is `pub` and takes a `VariableOrder` — and `sumcheck/benches/sumcheck.rs`, which builds `ORDERS = [Prefix, Suffix]` packed and benches both.

### The fix

Serving suffix binding from packed storage needs a cross-lane fold, which is a feature rather than a fix. Until there is one, this reuses the downgrade `transition` already performs (it unpacks once the packed view reaches zero variables) and unpacks a suffix-bound pair up front. The rounds are then computed over the variable the order actually names.

That keeps the public API total — no new panic — at the cost of not using SIMD for suffix binding, which is the honest trade: it was never actually computing the suffix round anyway. I'm happy to instead implement the cross-lane kernel, or make `new_packed` reject `Suffix` outright, if you'd prefer either.

The bench keeps working; its suffix rows now measure the scalar kernel, which the comment there records.

### Testing

`packed_and_unpacked_run_the_same_sumcheck` builds the same polynomial in both representations, asserts first that `pack` did not change it (`eval_packed == eval_ext` at a shared random point, so any divergence is the round kernel and not the construction), drives every round of both through a deterministic challenger, and compares the round messages.

Reverting only the one-line condition in `transition` and keeping the test:

```
assertion `left == right` failed: Suffix, 3 variables, SIMD width 4
```

`Prefix` passes either way. With the fix both orders pass.

---

# 2. `docs(sumcheck, multilinear-util): correct weight ordering, add promised assert`

Two places where a doc comment and its code disagree.

### `lagrange_weights_01inf_multi` documents the wrong ordering

The doc says *"Lexicographic by multi-index `(i_0, i_1, ..., i_{k-1})`"*, but the loop prepends each new coordinate's basis index as the **most** significant digit, so the result is indexed by `sum_j i_j * 3^j` — `i_0` is the *least* significant digit and the weights for `r_0` vary fastest.

The function's own `test_lagrange_weights_multi_k2` already pins the real convention (`weights[3*j + i] = w0[i] * w1[j]`), so the code and the test agree with each other; only the comment is wrong. Confirmed at `k = 3` before writing the replacement — `weights[i0 + 3*i1 + 9*i2]` matches for all 27 entries, `weights[9*i0 + 3*i1 + i2]` matches for none.

Worth getting right because the function is `pub`: a caller who indexes it the way the doc describes gets silently wrong weights for `k >= 2`, with no shape or length error to catch it.

### `fill_buffer_batch` promises a `debug_assert` that isn't there

The doc says it *"Panics in debug builds if `evals.width() != buffer.width()`"*. There is no such assertion — the `take(width)` in the column loop silently drops the trailing evaluation points instead. This adds the assertion the docs already promise.

The one caller, `eval_eq_batch_common`, sizes `parallel_buffer` from `evals.width()`, so it cannot fire today — it guards the invariant for the next caller rather than fixing a live bug.

---

## Testing (both commits)

Run in debug, so the new assertion is live:

- `p3-sumcheck` — 245 passed
- `p3-multilinear-util` — 148 passed
- `p3-whir` — 125 passed (the packed `Prefix` consumer)
- `p3-multi-stark` — 59 passed

`cargo clippy --all-targets -- --deny warnings` and `cargo +nightly fmt --check` clean.


pucedoteth:
@Nashtare done  #2009 is folded in here as a second commit (767dae9), and the description now covers both.

I kept them as two commits rather than squashing, since they are independent: the first is the packed-suffix binding fix, the second is the two doc/code mismatches (`lagrange_weights_01inf_multi`'s ordering, and the `debug_assert` `fill_buffer_batch` documents but does not have). Happy to squash if you would rather have one.

The two touch disjoint files, so the bundle needed no conflict resolution. Re-ran in debug so the new assertion is live: p3-sumcheck 245, p3-multilinear-util 148, p3-whir 125, p3-multi-stark 59 — all passing, clippy and nightly fmt clean.

sumcheck/src/product_polynomial.rs
Nashtare, line 74: I believe this is stale now and needs change

sumcheck/src/product_polynomial.rs
```diff
@@ -144,6 +144,13 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
     /// - `evals`   — packed evaluations of the sumchecked polynomial.
     /// - `weights` — packed evaluations of the weight polynomial.
     ///
+    /// # Storage is not guaranteed
+    ///
+    /// The transition check runs immediately, so the result is not necessarily
+    /// packed. A [`VariableOrder::Suffix`] pair is unpacked up front: packed
+    /// storage cannot bind the suffix variable, which lives inside the SIMD
+    /// lanes. See [`Self::transition`].
```
Nashtare: private item in public doc
```suggestion
    /// lanes. See `Self::transition`.
```

sumcheck/src/product_polynomial.rs
Nashtare, line 183: also, more of a design nit: since `order` is immutable and `Packed` is only built in `new_packed`, the `Suffix` branch can only fire on the constructor's `transition()` call; the per-round re-check is dead. Moving it into `new_packed` makes the "Packed ⇒ Prefix" invariant local.

Nashtare (commented):
the fix looks good to me, couple comments inline worth addressing

pucedoteth:
Thanks , all three addressed in 210649a.

**The per-round re-check is dead — moved into `new_packed`**

You're right, and it's a better place for it. I confirmed the two premises hold: `order` is never reassigned after construction, and `MaybePacked::Packed { .. }` is built in exactly one place (the literal in `new_packed`; every other occurrence in the file is a match pattern). So once `new_packed` returns, a suffix pair is already `Unpacked`, and the `order == VariableOrder::Suffix` arm in `transition` could only ever fire on the call `new_packed` makes itself.

`new_packed` now settles it up front:

```rust
if order == VariableOrder::Suffix {
    return Self::new_unpacked(order, evals.unpack(), weights.unpack());
}
```

and `transition` is back to the single size condition it had before this PR. Behaviour is identical — same `new_unpacked` call on the same inputs — and the arity assert that used to run inside `transition` on that path is the one already at the top of `new_packed`.

**Private item in public doc**

Fixed, though not literally as suggested, so flagging it. Once the suffix decision moved, `See Self::transition` pointed at a function that no longer handles suffix, so demoting it to a code span would have left a pointer to the wrong place. I inlined the rationale into `new_packed` instead and dropped the cross-reference. `cargo doc -p p3-sumcheck --no-deps` is warning-free. Happy to use the code-span form instead if you'd rather keep the pointer.

**Stale block**

Confirmed — that's the `MaybePacked` "Transition Logic" section, and this PR is what made it stale: it described the size condition as the only way out of `Packed`. It now also states the prefix-only invariant. I also changed `Small` to `Unpacked` while there, since `Small` isn't a variant name; say the word if you'd rather that stayed a separate change.

**Verification:** `cargo test -p p3-sumcheck` — 245 passed, including `packed_and_unpacked_run_the_same_sumcheck`. `cargo clippy -p p3-sumcheck --all-targets` and `cargo doc -p p3-sumcheck --no-deps` both clean.

Nashtare (approved):

PR #2010, opened 2026-08-31
feat: expose pruned-multiproof restoration and FRI/STARK opening internals

Nashtare:
## Description

Recursive verifiers built around a single-path-per-query in-circuit MMCS gadget can't consume 0.7's pruned multiproofs directly: a proof omits any sibling digest another query's own leaf already covers, on the assumption the verifier recomputes it by hashing rather than reading it. That's cheaper to verify but leaves nothing a per-query gadget can authenticate independently. Three additions close that gap without touching any existing public API or its tested behavior:

- p3-merkle-tree: `MerkleTreeMmcs::restore_and_recompute_paths` walks the same amortized frontier `verify_batch_pruned` does, but records each level's "other children" into every query's own path (not just the group lead's), including mixed-height batches where a shorter matrix's digest is injected mid-tree. Cross-checked against the trusted single-query `verify_batch` as an independent oracle, including queries that only merge one level above an injection point.

- p3-fri: `fold_query`, `open_inputs`, and the `FriOpenings` type are now public. A recursive verifier's witness generation needs each query's own per-round folded evaluation to hash a commit-phase leaf correctly; that value only exists by replaying FRI's fold chain, and this was the only implementation of it.

- p3-batch-stark: pulled the `commitments_with_opening_points` construction out of `verify_batch` into its own function, taking the already-sampled `zeta` and the already-derived per-instance shape data as plain arguments instead of requiring a live `A: Air<...>` reference for anything beyond `BaseAir::main_next_row_columns`/ `preprocessed_next_row_columns`. This is what a caller needs to make `open_inputs` callable at all. `verify_batch` itself is now a thin caller of it — behavior confirmed unchanged via the full prove/verify integration suite (ZK, lookups, preprocessed/periodic columns, tamper-rejection) and a whole-workspace compile check.

### DIFF
--- a/batch-stark/src/common.rs
+++ b/batch-stark/src/common.rs
@@ -195,6 +195,10 @@ where
             airs,
             trace_ext_degree_bits,
             &vec![0; airs.len()],
+            // Every override is zero, so packing cannot move the quotient bucket (the
+            // `debug_assert_eq!` below still pins that) and the blowup ceiling is
+            // unreachable from here.
+            usize::MAX,
         )
     }
 
@@ -203,13 +207,28 @@ where
     /// instead of only the free (quotient-chunk-preserving) budget. A `0` override reproduces
     /// [`Self::from_airs_and_degrees`] exactly. A caller raising an override past the free
     /// ceiling deliberately spends more quotient chunks on that instance to commit fewer aux
-    /// columns — the resulting chunk count must still fit the PCS's blowup
-    /// (`num_quotient_chunks <= 1 << log_blowup`), which this function does not itself enforce.
+    /// columns.
+    ///
+    /// # Arguments
+    ///
+    /// * `lookup_budget_overrides` - Per-AIR fraction-pin degree floors; one entry per AIR.
+    /// * `log_blowup` - The PCS's `log_blowup`. A rate-`2^-log_blowup` low-degree test only
+    ///   certifies degree `< |domain| * 2^-log_blowup`, so a quotient split into more than
+    ///   `2^log_blowup` chunks asks FRI to certify a degree it has no rate budget for. Any
+    ///   instance whose override raises the chunk count past that ceiling panics here.
+    ///
+    /// # Panics
+    ///
+    /// If an override raises an instance's quotient chunk count past `1 << log_blowup`. This
+    /// is a hard [`assert!`] rather than a `debug_assert!`: overrides are a rare, explicit,
+    /// caller-side choice, so the recomputation costs nothing on the default path, and a
+    /// release build must not silently emit a rate-violating proof.
     pub fn from_airs_and_degrees_with_lookup_budgets<A>(
         config: &SC,
         airs: &[A],
         trace_ext_degree_bits: &[usize],
         lookup_budget_overrides: &[usize],
+        log_blowup: usize,
     ) -> Self
     where
         SymbolicExpressionExt<Val<SC>, SC::Challenge>: Algebra<SC::Challenge>,
@@ -297,7 +316,8 @@ where
             .iter()
             .zip(trace_ext_degree_bits.iter())
             .zip(lookup_budget_overrides.iter())
-            .map(|((air, &ext_db), &override_budget)| {
+            .enumerate()
+            .map(|(i, ((air, &ext_db), &override_budget))| {
                 // Base trace length `N` (before any ZK extension), matching what the
                 // prover and verifier feed the degree model for this instance.
                 let trace_len = 1usize << (ext_db - is_zk);
@@ -337,10 +357,31 @@ where
                         log_chunks,
                         "same-bus packing must preserve the quotient chunk count when no override is set"
                     );
+                } else {
+                    // A deliberate override is allowed to raise the bucket, but only within the
+                    // PCS's rate budget: the low-degree test certifies degree
+                    // `< |domain| * 2^-log_blowup`, so a quotient split into more than
+                    // `2^log_blowup` chunks has no soundness margin left. This is the bound
+                    // `StarkSecurityParams::new` states as a buildability condition; without
+                    // this check nothing else on the prove or verify path detects a violation.
+                    let packed_log_chunks =
+                        get_log_num_quotient_chunks::<Val<SC>, SC::Challenge, A, LogUpGadget>(
+                            air,
+                            AirLayout::from_air(air),
+                            trace_len,
+                            &packed,
+                            is_zk,
+                            &lookup_gadget,
+                        );
+                    // The committed chunk count is `2^(log_chunks + is_zk)`.
+                    let log_num_chunks = packed_log_chunks + is_zk;
+                    assert!(
+                        log_num_chunks <= log_blowup,
+                        "instance {i}: lookup-budget override raised the quotient bucket to \
+                         2^{log_num_chunks} chunks, past the PCS blowup 2^{log_blowup}; \
+                         the prover cannot commit a quotient"
+                    );
                 }
-                // A deliberate override is allowed to raise the bucket; the caller is
-                // responsible for keeping the resulting chunk count within the PCS's blowup
-                // and for re-deriving the AIR layout from `packed`, not `unpacked`.
 
                 packed
             })
--- a/batch-stark/src/common.rs
+++ b/batch-stark/src/common.rs
@@ -190,10 +190,40 @@ where
         SymbolicExpressionExt<Val<SC>, SC::Challenge>: Algebra<SC::Challenge>,
         A: Air<InteractionSymbolicBuilder<Val<SC>, SC::Challenge>>,
     {
+        Self::from_airs_and_degrees_with_lookup_budgets(
+            config,
+            airs,
+            trace_ext_degree_bits,
+            &vec![0; airs.len()],
+        )
+    }
+
+    /// As [`Self::from_airs_and_degrees`], but each AIR may specify a lookup-packing budget
+    /// floor: `pack_same_bus` is called with `max(free_budget, lookup_budget_overrides[i])`
+    /// instead of only the free (quotient-chunk-preserving) budget. A `0` override reproduces
+    /// [`Self::from_airs_and_degrees`] exactly. A caller raising an override past the free
+    /// ceiling deliberately spends more quotient chunks on that instance to commit fewer aux
+    /// columns — the resulting chunk count must still fit the PCS's blowup
+    /// (`num_quotient_chunks <= 1 << log_blowup`), which this function does not itself enforce.
+    pub fn from_airs_and_degrees_with_lookup_budgets<A>(
+        config: &SC,
+        airs: &[A],
+        trace_ext_degree_bits: &[usize],
+        lookup_budget_overrides: &[usize],
+    ) -> Self
+    where
+        SymbolicExpressionExt<Val<SC>, SC::Challenge>: Algebra<SC::Challenge>,
+        A: Air<InteractionSymbolicBuilder<Val<SC>, SC::Challenge>>,
+    {
         assert_eq!(
             airs.len(),
             trace_ext_degree_bits.len(),
             "airs and trace_ext_degree_bits must have the same length"
+        );
+        assert_eq!(
+            airs.len(),
+            lookup_budget_overrides.len(),
+            "airs and lookup_budget_overrides must have the same length"
         );
 
         let pcs = config.pcs();
@@ -266,7 +296,8 @@ where
         let lookups: Vec<Lookups<Val<SC>>> = airs
             .iter()
             .zip(trace_ext_degree_bits.iter())
-            .map(|(air, &ext_db)| {
+            .zip(lookup_budget_overrides.iter())
+            .map(|((air, &ext_db), &override_budget)| {
                 // Base trace length `N` (before any ZK extension), matching what the
                 // prover and verifier feed the degree model for this instance.
                 let trace_len = 1usize << (ext_db - is_zk);
@@ -287,22 +318,29 @@ where
                 //
                 //     log2_ceil((d + is_zk).max(2) - 1) <= log_chunks
                 //  => d <= 2^log_chunks + 1 - is_zk
-                let budget = (1usize << log_chunks) + 1 - is_zk;
+                let free_budget = (1usize << log_chunks) + 1 - is_zk;
+                let budget = free_budget.max(override_budget);
                 let packed = unpacked.pack_same_bus(&lookup_gadget, budget);
 
-                // The fold must not have moved the quotient bucket.
-                debug_assert_eq!(
-                    get_log_num_quotient_chunks::<Val<SC>, SC::Challenge, A, LogUpGadget>(
-                        air,
-                        AirLayout::from_air(air),
-                        trace_len,
-                        &packed,
-                        is_zk,
-                        &lookup_gadget,
-                    ),
-                    log_chunks,
-                    "same-bus packing must preserve the quotient chunk count"
-                );
+                if override_budget <= free_budget {
+                    // No deliberate override: the fold must not have moved the quotient bucket,
+                    // exactly as before.
+                    debug_assert_eq!(
+                        get_log_num_quotient_chunks::<Val<SC>, SC::Challenge, A, LogUpGadget>(
+                            air,
+                            AirLayout::from_air(air),
+                            trace_len,
+                            &packed,
+                            is_zk,
+                            &lookup_gadget,
+                        ),
+                        log_chunks,
+                        "same-bus packing must preserve the quotient chunk count when no override is set"
+                    );
+                }
+                // A deliberate override is allowed to raise the bucket; the caller is
+                // responsible for keeping the resulting chunk count within the PCS's blowup
+                // and for re-deriving the AIR layout from `packed`, not `unpacked`.
 
                 packed
             })
--- a/batch-stark/src/verifier/mod.rs
+++ b/batch-stark/src/verifier/mod.rs
@@ -6,7 +6,7 @@ use alloc::vec::Vec;
 pub use data::VerifierData;
 use p3_air::symbolic::{AirLayout, SymbolicExpressionExt};
 use p3_air::{Air, BaseAir};
-use p3_commit::{Pcs, PolynomialSpace};
+use p3_commit::{CommitmentWithOpeningPoints, Pcs, PolynomialSpace};
 use p3_field::{Algebra, BasedVectorSpace, ExtensionField, PrimeCharacteristicRing, PrimeField};
 use p3_lookup::folder::VerifierConstraintFolderWithLookups;
 use p3_lookup::logup::LogUpGadget;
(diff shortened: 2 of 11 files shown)


Nashtare:
@tcoratger friendly ping, let me know if this looks good to you!

batch-stark/src/common.rs
```diff
@@ -287,22 +318,29 @@ where
                 //
                 //     log2_ceil((d + is_zk).max(2) - 1) <= log_chunks
                 //  => d <= 2^log_chunks + 1 - is_zk
-                let budget = (1usize << log_chunks) + 1 - is_zk;
+                let free_budget = (1usize << log_chunks) + 1 - is_zk;
+                let budget = free_budget.max(override_budget);
```
tcoratger, line 342: This is my main concern in the PR, so apologies for the long comment — I think the mechanism is fine but the guard needs to be real rather than documented.

Before this change, the `debug_assert_eq!` below made it *impossible* for packing to move the quotient bucket. Now an override can move it, and the invariant is handed to the caller in a comment. That precondition isn't a style preference — it's the FRI rate bound, and this repo already states it as a hard buildability condition in `uni-stark/src/security.rs:82`:

```rust
debug_assert!(
    air_max_constraint_degree <= (1usize << fri.log_blowup) + 1,
    "AIR max constraint degree exceeds blowup+1; the prover cannot commit a quotient"
);
```

A rate-`ρ = 2^{-log_blowup}` low-degree test only certifies degree `< |domain| * ρ`. Committing a quotient whose chunk count exceeds `2^log_blowup` asks FRI to certify a degree it has no rate budget for, so the soundness margin quietly disappears.

The part that worries me is that **nothing catches the violation.** I grepped every `log_blowup` reference under `batch-stark/` and the only hits are this PR's own doc comments. Measured on the balanced `two_adic_compat_case` (`log_blowup = 1`), varying `reps` on a shared bus with `override = usize::MAX`:

```
reps=2  free: log_chunks=[0,0]   MAX: log_chunks=[1,0]   -> VERIFIED
reps=3  free: log_chunks=[0,0]   MAX: log_chunks=[2,0]   -> no shape error
reps=6  free: log_chunks=[0,0]   MAX: log_chunks=[3,0]   -> no shape error
reps=8  free: log_chunks=[0,0]   MAX: log_chunks=[3,0]   -> no shape error
```

At `log_chunks = 3` vs `log_blowup = 1` — an 8× overshoot of a bound the repo itself calls unbuildable — `prove_batch` completes without panicking and `verify_batch` raises no shape or domain error.

To be fair about what I did and didn't show: I did **not** forge an accepted proof. What I did establish is (a) the violating regime is reachable through this new public API, (b) neither prover nor verifier detects it, and (c) this PR removes the assertion that previously made it unreachable. That's enough that I'd rather not ship it unguarded.

Suggested fix — recompute the bucket after packing and check it, reusing the idiom already at `security.rs:82`:

```rust
let packed_log_chunks = get_log_num_quotient_chunks::<Val<SC>, SC::Challenge, A, LogUpGadget>(
    air, AirLayout::from_air(air), trace_len, &packed, is_zk, &lookup_gadget,
);
assert!(
    packed_log_chunks + is_zk <= config.pcs().log_blowup(),
    "instance {i}: lookup-budget override raised the quotient bucket to 2^{packed_log_chunks} \
     chunks, past the PCS blowup 2^{}; the prover cannot commit a quotient",
    config.pcs().log_blowup(),
);
```

I'd argue for a hard `assert!` over `debug_assert!` here: an override is a rare, explicit, release-path caller choice, so the cost is nil, and a release build really shouldn't silently emit a rate-violating proof. If `log_blowup` isn't reachable from `SC`, taking the ceiling as an explicit argument would work just as well and makes the contract checkable.

To reproduce: `two_adic_compat_case` with `reps = 6` and `overrides = [usize::MAX, usize::MAX]` reaches `log_chunks = 3` against `log_blowup = 1`. Today that builds fine; with the assertion it fails loudly at setup, which is where you'd want to hear about it.

batch-stark/src/common.rs
```diff
@@ -287,22 +318,29 @@ where
...
+                            &packed,
+                            is_zk,
+                            &lookup_gadget,
+                        ),
+                        log_chunks,
+                        "same-bus packing must preserve the quotient chunk count when no override is set"
+                    );
+                }
+                // A deliberate override is allowed to raise the bucket; the caller is
+                // responsible for keeping the resulting chunk count within the PCS's blowup
```
tcoratger: Tiny thing: the second half of this sentence — "and for re-deriving the AIR layout from `packed`, not `unpacked`" — I don't think has a referent.

`AirLayout` is built purely from the AIR's own metadata (`preprocessed_width`, `air.width()`, `num_public_values`, `num_periodic_columns`) in both the prover (`prover.rs:183`) and the verifier (`verifier/mod.rs:382`); it never reads the lookup set at all. So there's no layout for a caller to re-derive, and I spent a little while hunting for the hazard it was warning me about before concluding there wasn't one. 🙂

The blowup ceiling is the real obligation — and per the other comment, I'd rather see that one become an assertion than a doc note.

fri/src/verifier.rs
```diff
@@ -468,7 +468,7 @@ where
 /// - `rows_by_round`: Collector for this query's reconstructed evaluation row at each round.
 #[expect(clippy::too_many_arguments)]
 #[inline]
-fn fold_query<Folding, F, EF, M>(
+pub fn fold_query<Folding, F, EF, M>(
```
tcoratger, line 492: Fully agree this needs to be public for the recursive-verifier use case — but I think it needs a bit of hardening before it becomes part of the API surface, because right now it panics on attacker-shaped input.

`fold_query` indexes proof-controlled data directly:

```rust
let siblings = &opening.sibling_values[query];   // verifier.rs:526
...
*eval = siblings[sibling_idx];
```

`verify_fri` makes both of those safe *before* calling it, at `verifier.rs:222–241`: `sibling_values.len() == params.num_queries`, and `siblings.len() == arity - 1` for every round and query. It also pins `log_arities` through `checked_log_arity` (which is what stops `log_current_height - log_arity` underflowing) and pre-sizes both collectors to `num_rounds`.

None of those obligations are stated on `fold_query`, none are checked inside it, and — this is the awkward one — **the most important one can't be discharged by an external caller at all**: `CommitPhaseMultiStep::checked_log_arity` is `pub(crate)` (`fri/src/proof.rs:51`), while every field of `CommitPhaseMultiStep` is `pub` and deserializes straight from an untrusted proof. So the PR exposes the consumer but keeps the guard private.

Quick repro — a round whose `sibling_values` is empty, i.e. exactly the shape `verify_fri` rejects with `CommitPhaseQueryCountMismatch`:

```
thread '...' panicked at fri/src/verifier.rs:526:47:
index out of bounds: the len is 0 but the index is 0
```

A verifier should reject rather than panic, and the PR description names recursive-verifier witness generation as the caller — which is precisely the code that'll be feeding it untrusted proof data.

Three suggestions, in rough order of value:

1. Make `checked_log_arity` `pub`. Without it the public API can't be used safely.
2. Move the two shape checks into `fold_query` and return the existing `FriError` variants. They're `O(rounds × queries)` integer comparisons guarding work that's already `O(rounds × queries × arity)` field ops, so the cost is noise, and it makes the function total on well-typed input.
3. Add a short `# Security` section. The doc currently opens with "Verifies a single query chain in the FRI proof", which reads as though the return value were authenticated — it isn't, until the caller runs the `verify_multi_batch` loop. Worth saying outright, plus a note that the collectors must be pre-sized to `commit_phase_commits.len()`.

For contrast, `open_inputs` needs none of this — it authenticates every input batch against its commitment before touching any value (`verifier.rs:645–652`), so exposing that one is safe exactly as-is. 👍

merkle-tree/src/mmcs/mod.rs
```diff
@@ -775,4 +775,306 @@ impl<P, PW, H, C, const N: usize, const DIGEST_ELEMS: usize>
...
+    /// function does not check anything against a commitment. A caller that also wants that
+    /// check should call `verify_batch_pruned` (or verify each returned path against the
+    /// commitment itself); the two do the same walk and either can be skipped once the other
+    /// has run.
+    ///
+    /// # Returns
+    ///
+    /// One [`MerkleAuthPath`] per entry of `indices`, in the same order (duplicates included) —
+    /// the same shape a pre-pruning, per-query proof would have supplied directly.
+    pub fn restore_and_recompute_paths<R>(
```
tcoratger, line 924: The algorithm here is good and I couldn't break it (details in the review summary) — my concern is purely that it's a verbatim second copy of the frontier walk.

Stripping comments and blank lines, this function is 238 lines and `verify_batch_pruned` is 210. I diffed them, and the entire functional delta is:

- the 12-line scatter block below (~L965–976),
- `chunk_base`,
- a cloned sibling slice instead of a borrow (to dodge the borrow conflict the scatter introduces),
- the final cap check swapped for expansion back to caller order.

Everything else is byte-identical: the geometry gate, the sorted-unique derivation, representative selection, duplicate-opening rejection, leaf hashing, the group-merge loop, injection-point detection, and the group-opening consistency check. That's the walk the pruned-proof soundness argument rests on, now living in two places.

What makes me twitchy is that the docstring *asserts* they agree ("walks the exact same amortized frontier `verify_batch_pruned` does") — and that's exactly the property nothing enforces. A future fix to one silently leaves the other wrong, and unlike `verify_batch_pruned` this function has no commitment check of its own to catch the drift.

Would you be up for factoring the walk into one private routine that both entry points call? The delta is small enough that a `const RECORD: bool` generic plus an `Option<&mut Vec<MerkleAuthPath<..>>>` covers it — `verify_batch_pruned` becomes shared-walk + cap check, this becomes shared-walk + reorder. Using a const generic rather than a runtime `bool` means the recording branch compiles out entirely on the verify path, so `verify_batch_pruned`'s hot loop stays exactly as it is today.

A cheap regression guard alongside it: a test asserting both functions accept and reject the same proofs over a randomized corpus.

merkle-tree/src/mmcs/mod.rs
```diff
@@ -775,4 +775,306 @@ impl<P, PW, H, C, const N: usize, const DIGEST_ELEMS: usize>
...
+                let group_start = parent_idx * step;
+
+                let mut j = i + 1;
+                while j < digests.len() && node_indices[j] / step == parent_idx {
+                    j += 1;
+                }
+
+                let lead_path = lead[i];
+                let lead_pos = node_indices[i] - group_start;
+                let cursor_start = sibling_cursor[i];
```
tcoratger: Small one: I think `sibling_cursor` is now dead state that exactly duplicates `chunk_base[level]`.

Every frontier entry advances one level per iteration, and `sibling_cursor` starts at `0` and gains `step - 1` per level — which is the definition of `chunk_base[level]`. So `sibling_cursor[i] == chunk_base[level]` at every read, for every entry, unconditionally. This function carries both, using `cursor_start` to slice the lead's chunk and `chunk_base[level]` to address the scatter writes.

Two names for one quantity inside a hand-rolled tree walk is how off-by-ones sneak in, so I'd drop `sibling_cursor`, its `new_cursor` scratch buffer, and the corresponding `mem::swap`, and index everything off `chunk_base[level]`. Four fewer lines of parallel bookkeeping and one fewer `Vec` per call.

(`verify_batch_pruned` has the same redundancy, but there it's harmless since `chunk_base` doesn't exist — so this is really an argument for the shared-helper refactor above.)

merkle-tree/src/mmcs/batch.rs
```diff
@@ -1027,4 +1027,63 @@ mod tests {
...
+    /// self-consistent. Exercises the two cases that matter most:
+    /// - a mixed-height batch, so restoration must fold the injected matrix's digest
+    ///   correctly (injection never appears in the pruned proof itself, see `mmcs/mod.rs`'s
+    ///   module docs);
+    /// - two queries (0 and 2) that only merge one level *above* the injection point, so their
+    ///   later-level sibling is nowhere in `verify_batch_pruned`'s own bookkeeping — only this
+    ///   function computes it;
+    /// - a duplicate query index, exercising the original-order expansion.
+    #[test]
+    fn restore_and_recompute_paths_matches_single_query_verification() {
```
tcoratger, line 1046: Lovely test — cross-checking against the trusted single-query `verify_batch` as an independent oracle is exactly the right move, and the mixed-height + post-injection-merge shape is well chosen. One gap though: it only runs at arity 2, which is where the new index arithmetic is trivial.

At `N = 2`, `num_siblings_per_path = step - 1 = 1`, so the offset computation that the scatter block turns on:

```rust
let offset = if pos < own_pos { pos } else { pos - 1 };
```

is identically `0`, and each level's chunk holds a single entry. Every way that arithmetic could be wrong — a mis-ordered chunk, an off-by-one against `sibling_offset`, a wrong `chunk_base[level]` stride — is invisible. `MyMmcs4` is already declared 13 lines above at L30 and goes unused here.

I wrote that coverage looking for a real bug and **found none** — arity 4 passes across uniform and mixed heights, non-zero caps, duplicates and post-injection merges, and an 800-case randomized cross-check (400 at `N = 2` with caps 0–2, 400 at `N = 4`) produced no failure. So this is a coverage gap rather than a defect. But it's a gap on the one property the PR is selling, and it's cheap to close.

Since there's a trusted oracle in-repo and the invariant is one line, this feels like a better fit for `proptest` than a second fixed shape:

```
inputs:
  arity       N in {2, 4}                        (const-generic, so one fn per N)
  max_height  1..=70
  heights     max_height, then ceil(max_height / 2^k) for distinct k   -- commit-reachable
  widths      1..=4 per matrix
  cap_height  0..=2
  queries     1..=6 drawn from 0..max_height, duplicates allowed

property:
  let (commit, pd) = mmcs.commit(mats);
  let (ov, pruned) = mmcs.open_multi_batch(&queries, &pd);
  let restored     = mmcs.restore_and_recompute_paths(&dims, &queries, &ov, &pruned)?;
  for (q, &idx) in queries.iter().enumerate() {
      assert_eq!(restored[q].leaf_index, idx);
      mmcs.verify_batch(&commit, &dims, idx,
                        BatchOpeningRef::new(&ov[q], &restored[q].siblings))?;   // must pass
  }

a failing case proves the restored siblings are self-consistent but wrong --
i.e. the scatter wrote the wrong offset, stride, or group range.
```

Two generator gotchas worth baking into the test, both of which cost me real time to rediscover:

- heights must be commit-reachable (`ceil(max_height / 2^k)`), otherwise `commit` panics at `batch.rs:53`;
- `N = 4` with a non-zero `cap_height` hits a **pre-existing** panic in `commit` (`symmetric/src/hash.rs:35`, `assert!(cap.len().is_power_of_two())`) that's unrelated to this PR — so pin `cap_height = 0` for the 4-ary generator, or fix that separately.

batch-stark/src/verifier/mod.rs
```diff
@@ -22,12 +22,278 @@ use p3_util::zip_eq::zip_eq;
...
+/// checks it needs itself.
+///
+/// # Returns
+///
+/// `(commitments_with_opening_points, quotient_domains)` — the second element is each
+/// instance's quotient-chunk domains, a byproduct of this construction that `verify_batch`
+/// also needs for its own post-opening constraint check.
+#[expect(clippy::too_many_arguments)]
+#[expect(clippy::type_complexity)]
+pub fn commitments_with_opening_points<SC, A>(
```
tcoratger, line 98: As above, I checked this extraction line by line against the original inline block and it's faithful — including the `if is_lookup` → `if commitments.permutation.is_some()` swap, which is a no-op since `is_lookup` is defined as exactly that expression at L533. Nice.

Two API notes, both minor:

**1. Two of the 11 arguments are fully derivable from the others:**

```
ext_domain_sizes[i]     == 1 << degree_bits[i]                        // via validate_degree_bits
num_quotient_chunks[i]  == 1 << (log_num_quotient_chunks[i] + is_zk)  // via checked_log_size_sum
```

Neither relation is enforced. A caller who passes `1 << base_db` for `ext_domain_sizes` — an easy slip, since `verify_batch` keeps `base_degree_bits` and `ext_domain_sizes` as adjacent locals — builds the trace round on the wrong domain. `pcs.verify` will reject, so it's a liveness footgun rather than a soundness hole, but the failure surfaces a long way from its cause. Deriving both internally drops two arguments *and* a whole class of caller error.

**2. The doc could name the checks a caller inherits.** "no shape validation beyond what the construction itself requires" is accurate, but it leaves the reader to work out *which* checks they're now responsible for. `verify_batch` additionally enforces the lookup-commitment consistency check (L535), per-instance public-value and trace-width validation, preprocessed presence/width agreement, terminal presence, and `check_multiplicity_height_bound` — the LogUp bound that stops multiplicities wrapping mod `p`. An `# Omitted checks` section listing those by name turns "you might be missing something" into a checklist someone writing a recursive verifier can actually work through. Given that's the stated audience, I think it earns its keep.

Last nit: I'd drop `#[expect(clippy::type_complexity)]` in favour of the alias that already exists for this exact shape — `p3_fri::CommitmentWithOpeningPoints<Challenge, Commitment, Domain>` (`fri/src/two_adic_pcs.rs:81`) is structurally identical to the returned element type. Silencing the lint on a *public* return type hides the complexity from the signature rather than naming it. (If `batch-stark` shouldn't depend on `p3-fri`, a local alias or one in `p3-commit` would do the same job.)

tcoratger (commented):
Really nice PR — thanks for writing it up so carefully, the description made this much easier to review. 🙏

I spent most of my time on the `merkle-tree` piece since that's the one carrying real cryptographic weight, and I want to lead with the good news: **I couldn't break it.** I cross-checked `restore_and_recompute_paths` against `verify_batch` as an independent oracle over ~800 randomized cases (arity 2 *and* 4, commit-reachable mixed heights, non-zero caps, duplicate queries, post-injection merges) and found no counterexample. I also convinced myself the scatter is *complete*: at each level the `[group_lo[k], group_hi[k])` ranges partition `[0, n_unique)`, and each entry writes all `step - 1` offsets of its own chunk, so every slot of every path gets filled exactly once per level. The `offset`/`chunk_base` arithmetic lines up with `pruning::sibling_offset` and `total_siblings_for_levels`. Good algorithm, and the docstring explaining *why* the pruned proof can't serve a per-query gadget is genuinely excellent.

I also diffed the extracted `commitments_with_opening_points` against the original inline block line by line — it's a faithful extraction. The one apparent semantic change (`if is_lookup` → `if commitments.permutation.is_some()`) is a no-op, since `is_lookup` is *defined* as exactly that expression. 👍

Two things I'd like to sort out before this goes in, both about what the PR stops guarding rather than about the algorithms:

1. **The lookup-budget override (`common.rs`) removes the only enforcement of the quotient-degree/blowup bound, and nothing replaces it.** I measured it reaching `log_chunks = 3` against `log_blowup = 1` with no diagnostic from prover or verifier. Details inline.
2. **`fold_query` goes `pub` while `checked_log_arity` — the guard that makes it safe — stays `pub(crate)`,** so external callers can't validate what they pass. It panics on attacker-shaped proof data today; I have a repro inline.

Plus a duplication concern on the Merkle walk, an arity-coverage gap in the new test, and a couple of small API/doc notes. Nothing structural — the shape of all three exposures is right.

One process note: the second commit (`87de80b9`, the lookup-packing override) isn't mentioned in the PR description at all, and it's the only change here that touches soundness-relevant enforcement. Would you consider splitting it into its own PR? The other three exposures are in good shape and could land quickly, and that one deserves a focused look.

Also worth flagging separately (**pre-existing, not yours**): arity-4 `commit` with a non-zero `cap_height` and non-power-of-two heights panics at `symmetric/src/hash.rs:35` (`assert!(cap.len().is_power_of_two())`). It's on prover-chosen parameters rather than attacker input, so it isn't a security issue, but it tripped up my fuzzing and probably deserves its own issue.

tcoratger (approved):

PR #2012, opened 2026-09-02
refactor(stir): typed InvalidProofShape errors

tcoratger:
Addresses https://github.com/Plonky3/Plonky3/pull/1990#discussion_r3853544260.

## The problem

`StirError::InvalidProofShape` was a unit variant:

```rust
/// The proof has the wrong number of rounds, queries, or OOD answers.
#[error("Invalid proof shape")]
InvalidProofShape,
```

It was raised from **22 different call sites** across `verifier.rs` and `pcs.rs` — a wrong
round count, a missing initial commitment, an over-long shake polynomial, a truncated input
opening at one LDE-height bucket, a disagreeing replicated grind witness. Receiving it told
a caller only that *something* about the proof's shape was wrong. As the review comment put
it: "receiving that as an error doesn't help much."

## The fix

A new `ProofShapeError` enum, one variant per check, carrying where the mismatch was and
both numbers. `StirError::InvalidProofShape` transparently wraps it:

```rust
/// The proof's shape disagrees with the configuration or the public input.
#[error(transparent)]
InvalidProofShape(#[from] ProofShapeError),
```

This is the shape `p3-uni-stark` already uses — `VerificationError::InvalidProofShape`
wrapping `InvalidProofShapeError` — so nothing new is invented here; the STIR verifier
catches up with its sibling, down to `got`/`maximum` field naming and lowercase messages.

Before / after, on the same tampered proofs:

```
Invalid proof shape   →  round 0: shake polynomial has 1024 coefficients, expected at most 19
Invalid proof shape   →  final round: missing query openings
Invalid proof shape   →  bucket 2^9, commitment 0: 21 opened rows, expected 22
Invalid proof shape   →  instance 1: expected 3 round proofs, got 2
```

Every variant is pinned by data the verifier already holds — the round schedule and query
counts from `StirConfig`, the bucket and opening layout from the claimed evaluations — so
these are always malformed-proof reports, never failed algebraic checks. Those keep their
own `StirError` variants.

## Two supporting types

`RoundLabel::{Round(usize), Final}` replaces the `num_rounds` sentinel that every
round-carrying variant used to label the final round. Before, a three-round proof reported
`round 3`; now it reports `final round`. The sentinel predated this PR (`error_round:
num_rounds` already fed `InvalidMmcsProof`) but was invisible while the shape error was a
unit variant, so `StirError`'s own round variants take the type too.

`GrindStage::{Folding, Query}` names which of a round's two proof-of-work grinds a
replicated-witness disagreement came from, in the batched lockstep path.

## Compound checks split

Four checks tested two independent conditions and reported one error. Each is now split, so
the variant you get is unambiguous:

| Was one check | Now |
|---|---|
| `ans.len() > max \|\| shake.len() > max - 1` | `AnsPolynomialTooLong` / `ShakePolynomialTooLong` |
| `rows.len() != n \|\| any(row.len() != arity)` | `QueryOpeningCount` / `OpenedRowArity` |
| `fibers.len() != n \|\| any(f.len() != arity)` | `FiberCount` / `FiberArity` |
| `match (commitment, is_external) { _ => Err }` | `MissingInitialCommitment` / `UnexpectedInitialCommitment` |

The last is the interesting one: a proof that *omits* a commitment STIR should have
committed, and a proof that *smuggles in* a commitment to an externally bound oracle, are
opposite failures and the crate already had a test for each. They now assert different
errors.

## What is not a proof-shape error

`ExternalSourceError` holds the checks on the caller's external initial-oracle fiber source.
A `Vec<Src>` of closures cannot be deserialised, so its shape is caller-constructed rather
than attacker-supplied — a wrong answer there is a caller bug, and the message should say so
rather than blaming the proof:

```
external initial oracle: round 0: source returned 18 fibers, expected 19
```

## Robustness fixes found along the way

Three defects surfaced while auditing what each check could actually report. All are
reproduced under test.

**`batch_multiplicative_inverse` panicked on a coinciding opening point.** A claimed opening
point sitting on a queried fiber lane makes a quotient denominator vanish, and
`batch_multiplicative_inverse` is documented to panic on a zero input, so
`TwoAdicStirPcs::verify` aborted the process instead of returning `Err`. `p3-fri`
(`FriError::OpeningPointMatchesQueryPoint`, added in #1762) and `p3-circle` both reject it;
STIR now does too. No soundness impact — `zeta` from the extension field makes a collision
negligible for STARK use — but a direct PCS caller opening at a base-field coset point hit a
real abort.

**`verify_stir_multi` panicked on an empty batch.** With empty `configs`, `max_m == 0` skips
the round loop and the final witness collection indexed an empty vector. It now returns
`Ok(vec![])`, which is transcript-exact against the prover: `grind(0)` returns `F::ZERO` and
`check_witness(0, _)` returns `true`, both before touching the sponge.

**Two `assert_eq!`s on caller-supplied lengths.** `proofs.len() != configs.len()` is now
`ProofShapeError::InstanceCount`, and `sources.len() != b` is
`ExternalSourceError::SourceCount`. No `assert!` remains in `verifier.rs`.

## Two variants deleted as dead by construction

`HeightClassCount` and `MissingCombineCoefficient` both compared `native_heights.len()`
against itself: `expected_ro_by_class` is sized by `native_heights`, `Combine` runs exactly
when that holds two or more classes, and its coefficient map is keyed by those same heights.
Neither was constructible. The one gap — a bucket with zero classes, from a commitment
claimed with an empty matrix list — is rejected earlier by
`StirConfigError::StartingFoldingFactorExceedsDegree`, since such a bucket always derives
`log_stir_degree == 1`.

They are now `expect`s, matching the two neighbouring invariants of the same kind three
lines away (`expect("bucket_native_heights is built from these claims")`).

## Module move

`StirError` moves from `verifier.rs` to a new `stir::error` module, next to
`ProofShapeError` — mirroring `uni-stark/src/error.rs`, and keeping the error definitions out
of an already 1400-line verifier. The crate-root path `p3_stir::StirError` is unchanged;
`p3_stir::verifier::StirError` is gone, and the commit carries a `BREAKING CHANGE:` footer
for release-plz.

## Tests

88 in `tests/stir.rs`, up from 65 on main. Every shape assertion compares the whole error
value — no `..` wildcards, except on `InvalidMmcsProof` whose `source` is not `PartialEq`.

Three harnesses keep each case to a few lines: `shape_error_after` (mutate a
committed-oracle proof), `verify_external_initial` (mutate the proof *and* the caller's fiber
source), and `multi_shape_error_after` (mutate one instance of a batch).

Coverage that did not exist before this PR:

- **`verify_stir_with_external_initial` had no direct test at all** — it was only exercised
  through the PCS. There is now an honest positive test plus three rejections.
- **An input opening placed at the *wrong* LDE-height bucket had no test** — the
  `has_at_bucket` branch that rejects it was never hit.
- **Both `GrindStage` values, both directions of each length check**, a mutation at round 1
  (so `round` is pinned as threaded rather than hard-coded 0), the final-round label, the
  batch-count checks, the empty batch, and the coset opening point.

The two native-height tests also got sharper. They previously asserted only
`InvalidProofShape`; the typed error revealed that the rejection actually lands on the
*opened-row count*, because a different class count means a different `combine_key`, hence a
different `StirConfig`, hence a different first-round query count. They now assert that
mismatch exactly, and their comments say so.

Not covered, deliberately: `MissingCommitment` is defensive — the initial commitment's
presence is already pinned before any round runs, and every later round reads the preceding
round's commitment.

## Follow-up left out

In `verify_stir_multi`, only `RoundCount` names the offending instance. The uniform version
is a recursive `StirError::Instance { instance, source: Box<Self> }` wrapper applied at each
per-instance `?`, which would cover `InvalidPowWitness`, `InvalidMmcsProof`,
`InvalidShakeConsistency` and the shape errors raised inside the shared query-fetch path in
one go. It works, but interpolating `{source}` makes thiserror infer
`Box<StirError<E, IE>>: Display` as a bound on `impl Display for StirError<E, IE>` — the impl
requires itself, and rustc reports `error[E0275]`. The escape is a hand-written `Display` +
`Error` for all twelve variants, trading this crate's colocated `#[error(...)]` messages for
~60 lines of `write!`. The right shape for a follow-up is threading `instance` through
`RoundRowsRequest` and the two fetch methods instead.

## Verification

- `cargo test -p p3-stir` — 157 tests pass (88 in `tests/stir.rs`, up from 65)
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo doc --no-deps --workspace --document-private-items` with
  `RUSTDOCFLAGS=-D rustdoc::broken_intra_doc_links` — clean
- `cargo +nightly fmt --all --check` — clean

### DIFF
--- a/stir/src/error.rs
+++ b/stir/src/error.rs
@@ -1,0 +1,331 @@
+//! Error types for STIR verification.
+
+use alloc::format;
+use alloc::string::String;
+use core::fmt::{Display, Formatter, Result as FmtResult};
+
+use thiserror::Error;
+
+use crate::config::StirConfigError;
+
+/// Which round an error refers to.
+///
+/// In a batch, a proof element carries its instance's own round index.
+/// A shared grind carries the global lockstep round.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum RoundLabel {
+    /// An intermediate folding round, indexed from 0.
+    Round(usize),
+    /// The final round, checked against the sent polynomial rather than queried again.
+    Final,
+}
+
+impl Display for RoundLabel {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        match self {
+            Self::Round(round) => write!(f, "round {round}"),
+            Self::Final => f.write_str("final round"),
+        }
+    }
+}
+
+/// Which of a round's two proof-of-work grinds a witness belongs to.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum GrindStage {
+    /// The grind before the folding challenge `gamma`.
+    Folding,
+    /// The grind before the combination challenge `r_comb` and the query indices.
+    Query,
+}
+
+impl Display for GrindStage {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.write_str(match self {
+            Self::Folding => "folding",
+            Self::Query => "query",
+        })
+    }
+}
+
+/// Why a proof's shape does not match what the configuration and the public input pin.
+///
+/// `expected` is the value the verifier derived.
+/// `got` is what the proof carried.
+#[derive(Copy, Clone, Debug, Error, PartialEq, Eq)]
+pub enum ProofShapeError {
+    /// A batch takes one proof per configured instance.
+    #[error("expected {expected} proofs, got {got}")]
+    InstanceCount { expected: usize, got: usize },
+
+    /// A proof carries one round proof per configured round.
+    #[error(
+        "{}expected {expected} round proofs, got {got}",
+        instance.map_or_else(String::new, |i| format!("instance {i}: "))
+    )]
+    RoundCount {
+        /// Batch index of the offending proof, `None` outside a batch.
+        instance: Option<usize>,
+        expected: usize,
+        got: usize,
+    },
+
+    /// STIR commits the initial oracle itself, so the proof must carry that commitment.
+    #[error("missing initial oracle commitment")]
+    MissingInitialCommitment,
+
+    /// The initial oracle is bound by the caller, so the proof must carry no commitment.
+    #[error("unexpected initial oracle commitment for an externally bound oracle")]
+    UnexpectedInitialCommitment,
+
+    /// The final degree bound pins the final polynomial's coefficient count exactly.
+    #[error("final polynomial has {got} coefficients, expected {expected}")]
+    FinalPolynomialLength { expected: usize, got: usize },
+
+    /// The soundness assumption fixes each round's out-of-domain sample count.
+    #[error("{round}: {got} OOD answers, expected {expected}")]
+    OodAnswerCount {
+        round: RoundLabel,
+        expected: usize,
+        got: usize,
+    },
+
+    /// `Ans` interpolates the round's OOD and query points, bounding its degree.
+    ///
+    /// A shorter polynomial is legitimate: the prover may strip trailing zero coefficients.
+    #[error("{round}: ans polynomial has {got} coefficients, expected at most {maximum}")]
+    AnsPolynomialTooLong {
+        round: RoundLabel,
+        maximum: usize,
+        got: usize,
+    },
+
+    /// The shake polynomial is `Ans`'s quotient sum, one degree below `Ans` itself.
+    #[error("{round}: shake polynomial has {got} coefficients, expected at most {maximum}")]
+    ShakePolynomialTooLong {
+        round: RoundLabel,
+        maximum: usize,
+        got: usize,
+    },
+
+    /// A committed oracle is read through a Merkle multi-opening the proof must supply.
+    #[error("{round}: missing query openings")]
+    MissingQueryOpenings { round: RoundLabel },
+
+    /// An externally bound oracle is answered by the caller, so the proof must open nothing.
+    #[error("{round}: unexpected query openings for an externally bound oracle")]
+    UnexpectedQueryOpenings { round: RoundLabel },
+
+    /// A committed oracle is authenticated against a commitment the proof must supply.
+    #[error("{round}: missing oracle commitment")]
+    MissingCommitment { round: RoundLabel },
+
+    /// One opened row per query the round draws, a count the configuration fixes.
+    #[error("{round}: {got} opened rows, expected {expected}")]
+    QueryOpeningCount {
+        round: RoundLabel,
+        expected: usize,
+        got: usize,
+    },
+
+    /// An opened row is one fiber of the round's fold, so it holds `arity` evaluations.
+    #[error("{round}, query {query}: opened row has {got} evaluations, expected {expected}")]
+    OpenedRowArity {
+        round: RoundLabel,
+        query: usize,
+        expected: usize,
+        got: usize,
+    },
+
+    /// Instances verified in lockstep share one grind, so each must replicate its witness.
+    #[error("{round}: replicated {stage} PoW witnesses disagree across batched instances")]
+    ReplicatedWitnessMismatch {
+        round: RoundLabel,
+        stage: GrindStage,
+    },
+
+    /// The opening proof carries one STIR instance per distinct shared LDE height.
+    #[error("expected {expected} LDE-height buckets, got {got}")]
+    BucketCount { expected: usize, got: usize },
+
+    /// A bucket carries one input-opening slot per public commitment, occupied or not.
+    ///
+    /// A short list would silently drop trailing commitments from the reduced opening.
+    #[error("bucket 2^{log_height}: {got} input opening slots, expected {expected}")]
+    InputOpeningCount {
+        log_height: usize,
+        expected: usize,
+        got: usize,
+    },
+
+    /// A commitment whose matrices live at this bucket's height must be opened there.
+    #[error("bucket 2^{log_height}, commitment {commitment}: missing input opening")]
+    MissingInputOpening {
+        log_height: usize,
+        commitment: usize,
+    },
+
+    /// A commitment whose matrices live at another height must not be opened at this bucket.
+    #[error("bucket 2^{log_height}, commitment {commitment}: unexpected input opening")]
+    UnexpectedInputOpening {
+        log_height: usize,
+        commitment: usize,
+    },
+
+    /// A commitment carries one Merkle root per group of the layout its claims imply.
+    #[error("commitment {commitment}: expected {expected} Merkle roots, got {got}")]
+    CommitmentRootCount {
+        commitment: usize,
+        expected: usize,
+        got: usize,
+    },
+
+    /// A bucket that ran no `Combine` holds a single native-height class.
+    #[error("bucket 2^{log_height}: {got} native-height classes, expected {expected}")]
+    HeightClassCount {
+        log_height: usize,
+        expected: usize,
+        got: usize,
+    },
+
+    /// Every class merged by `Combine` needs its degree-correction coefficient.
+    #[error(
+        "bucket 2^{log_height}: no Combine coefficient for native height 2^{log_native_height}"
+    )]
+    MissingCombineCoefficient {
+        log_height: usize,
+        log_native_height: usize,
+    },
+
+    /// The opened-row count is the bucket's deduplicated first-round query count.
+    #[error(
+        "bucket 2^{log_height}, commitment {commitment}: {got} opened rows, expected {expected}"
+    )]
+    InputOpenedRowCount {
+        log_height: usize,
+        commitment: usize,
+        expected: usize,
+        got: usize,
+    },
+}
+
+/// Why the caller's external initial-oracle source did not answer as asked.
+///
+/// The source is a closure, so these report a caller bug rather than a malformed proof.
+#[derive(Copy, Clone, Debug, Error, PartialEq, Eq)]
+pub enum ExternalSourceError {
+    /// A batch takes one fiber source per instance.
+    #[error("expected {expected} fiber sources, got {got}")]
+    SourceCount { expected: usize, got: usize },
+
+    /// The source must answer one fiber per requested index.
+    #[error("{round}: source returned {got} fibers, expected {expected}")]
+    FiberCount {
+        round: RoundLabel,
+        expected: usize,
+        got: usize,
+    },
+
+    /// Each fiber holds the round's `arity` evaluations.
+    #[error("{round}, fiber {fiber}: source returned {got} evaluations, expected {expected}")]
+    FiberArity {
+        round: RoundLabel,
+        /// Position in the sorted, deduplicated index list handed to the source.
+        fiber: usize,
+        expected: usize,
+        got: usize,
+    },
+}
+
+/// Errors returned by [`crate::verifier::verify_stir`].
+#[derive(Debug, Error, PartialEq)]
+pub enum StirError<MmcsError, InputError = ()> {
+    /// A proof-of-work witness failed verification.
+    #[error("{round}: invalid proof-of-work witness")]
+    InvalidPowWitness { round: RoundLabel },
+
+    /// A Merkle multi-opening proof failed for a round's queries.
+    #[error("{round}: invalid MMCS opening proof")]
+    InvalidMmcsProof {
+        round: RoundLabel,
+        #[source]
+        source: MmcsError,
+    },
+
+    /// The shake polynomial identity failed at the random evaluation point.
+    #[error("{round}: shake polynomial consistency check failed")]
+    InvalidShakeConsistency { round: RoundLabel },
+
+    /// A virtual-oracle evaluation landed in the prior round's challenge set.
+    #[error("{round}, query {query}: invalid virtual-oracle query")]
+    InvalidRoundConsistency { round: RoundLabel, query: usize },
+
+    /// The final polynomial does not evaluate consistently with the last committed codeword.
+    #[error("final polynomial evaluation mismatch")]
+    FinalPolyMismatch,
+
+    /// The proof's shape disagrees with the configuration or the public input.
+    #[error(transparent)]
+    InvalidProofShape(#[from] ProofShapeError),
+
+    /// The caller's external initial-oracle source misbehaved.
+    #[error("external initial oracle: {0}")]
+    ExternalSource(#[from] ExternalSourceError),
+
+    /// A matrix in `commitment`'s batch was opened at zero points, so its width cannot be
+    /// pinned from the claimed evaluations. Every input matrix must be opened at >= 1 point.
+    #[error("commitment {commitment}, matrix {matrix}: opened at zero points")]
+    MatrixWithoutOpeningPoints { commitment: usize, matrix: usize },
+
+    /// A claimed opening point coincides with a queried fiber lane.
+    ///
+    /// The quotient `(f(z) - f(x)) / (z - x)` is undefined there.
+    #[error(
+        "commitment {commitment}, matrix {matrix}, point {point}: opening point coincides with a query point"
+    )]
+    OpeningPointMatchesQueryPoint {
+        commitment: usize,
+        matrix: usize,
+        point: usize,
+    },
+
+    /// The requested STIR parameters cannot reach `security_level` at some LDE-height bucket.
+    #[error("STIR config error: {0}")]
+    Config(#[source] StirConfigError),
+
+    /// An error propagated from the input polynomial commitment scheme.
+    #[error("input error")]
+    InputError(InputError),
+}
+
+impl<E, IE> StirError<E, IE> {
+    /// Map the `InputError` variant to a different type.
+    pub fn map_input_err<IE2>(self, f: impl FnOnce(IE) -> IE2) -> StirError<E, IE2> {
+        match self {
+            Self::InvalidPowWitness { round } => StirError::InvalidPowWitness { round },
+            Self::InvalidMmcsProof { round, source } => {
+                StirError::InvalidMmcsProof { round, source }
+            }
+            Self::InvalidShakeConsistency { round } => StirError::InvalidShakeConsistency { round },
+            Self::InvalidRoundConsistency { round, query } => {
+                StirError::InvalidRoundConsistency { round, query }
+            }
+            Self::FinalPolyMismatch => StirError::FinalPolyMismatch,
+            Self::InvalidProofShape(e) => StirError::InvalidProofShape(e),
+            Self::ExternalSource(e) => StirError::ExternalSource(e),
+            Self::MatrixWithoutOpeningPoints { commitment, matrix } => {
+                StirError::MatrixWithoutOpeningPoints { commitment, matrix }
+            }
+            Self::OpeningPointMatchesQueryPoint {
+                commitment,
+                matrix,
+                point,
+            } => StirError::OpeningPointMatchesQueryPoint {
+                commitment,
+                matrix,
+                point,
+            },
+            Self::Config(e) => StirError::Config(e),
+            Self::InputError(e) => StirError::InputError(f(e)),
+        }
+    }
+}
--- a/stir/src/lib.rs
+++ b/stir/src/lib.rs
@@ -11,7 +11,8 @@
 //! - [`proof`]: Proof types ([`StirProof`], [`StirRoundProof`], [`StirQueryOpenings`]).
 //! - [`utils`]: Polynomial arithmetic primitives (shake, ans, Horner eval, synthetic division).
 //! - [`prover`]: The STIR prover ([`prover::prove_stir`]).
-//! - [`verifier`]: The STIR verifier ([`verifier::verify_stir`], [`verifier::StirError`]).
+//! - [`verifier`]: The STIR verifier ([`verifier::verify_stir`]).
+//! - [`error`]: Verification failures ([`StirError`], [`ProofShapeError`]).
 //! - [`pcs`]: [`TwoAdicStirPcs`] implementing the [`p3_commit::Pcs`] trait.
 //!
 //! # Deviations from the STIR paper (eprint 2024/390)
@@ -48,6 +49,7 @@
 extern crate alloc;
 
 pub mod config;
+pub mod error;
 pub mod pcs;
 pub mod proof;
 pub mod prover;
@@ -56,7 +58,8 @@ pub mod utils;
 pub mod verifier;
 
 pub use config::{Stage, StirConfig, StirConfigError, StirParameters, StirRoundConfig};
+pub use error::{ExternalSourceError, GrindStage, ProofShapeError, RoundLabel, StirError};
 pub use p3_security::whir::SecurityAssumption;
 pub use pcs::{DEFAULT_MAX_LOG_HEIGHT_SPREAD, StirCommitment, TwoAdicStirPcs};
 pub use proof::{StirProof, StirQueryOpenings, StirRoundProof};
-pub use verifier::{StirError, StirVerifyOutputs};
+pub use verifier::StirVerifyOutputs;
--- a/stir/src/verifier.rs
+++ b/stir/src/verifier.rs
@@ -9,9 +9,9 @@ use p3_field::{
     BasedVectorSpace, ExtensionField, Field, TwoAdicField, batch_multiplicative_inverse,
 };
 use p3_matrix::Dimensions;
-use thiserror::Error;
 
 use crate::config::{StirConfig, StirRoundConfig};
+use crate::error::{ExternalSourceError, GrindStage, ProofShapeError, RoundLabel, StirError};
 use crate::proof::{StirProof, StirQueryOpenings, StirRoundProof};
 use crate::utils::{
     check_shake_consistency, eval_poly, eval_poly_at_base, fold_domain_params,
@@ -129,13 +129,47 @@ fn single_matrix_opened_values<EF>(row_evals: &[Vec<EF>]) -> Vec<Vec<&[EF]>> {
         .collect()
 }
 
+/// Reject an `Ans`/shake pair longer than the round's degree bounds allow.
+///
+/// - `Ans` interpolates `max_ans_len` points, so its degree is below that.
+/// - The shake polynomial is one degree lower.
+/// - A shorter pair is legitimate: the prover may strip trailing zeros.
+fn check_ans_and_shake_lengths<EF, MmcsError, InputError>(
+    round: RoundLabel,
+    ans_polynomial: &[EF],
+    shake_polynomial: &[EF],
+    max_ans_len: usize,
+) -> Result<(), StirError<MmcsError, InputError>> {
+    if ans_polynomial.len() > max_ans_len {
+        return Err(ProofShapeError::AnsPolynomialTooLong {
+            round,
+            maximum: max_ans_len,
+            got: ans_polynomial.len(),
+        }
+        .into());
+    }
+    let max_shake_len = max_ans_len.saturating_sub(1);
+    if shake_polynomial.len() > max_shake_len {
+        return Err(ProofShapeError::ShakePolynomialTooLong {
+            round,
+            maximum: max_shake_len,
+            got: shake_polynomial.len(),
+        }
+        .into());
+    }
+    Ok(())
+}
+
 /// Fetch an external initial oracle's queried fibers and lay them out in draw order.
 ///
 /// Queries are sampled with replacement, so `indices` may repeat; the source is asked once for
 /// the sorted unique indices and must answer in that same order.
+///
+/// `round` labels the reported errors: whichever round reads the initial oracle.
 fn external_fibers_in_draw_order<EF: Clone, MmcsError, InputError>(
     indices: &[usize],
     arity: usize,
+    round: RoundLabel,
     source: impl FnOnce(&[usize]) -> Result<Vec<Vec<EF>>, StirError<MmcsError, InputError>>,
 ) -> Result<Vec<Vec<EF>>, StirError<MmcsError, InputError>> {
     let mut unique = indices.to_vec();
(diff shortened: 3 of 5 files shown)


stir/src/error.rs
```diff
@@ -0,0 +1,274 @@
...
+    #[error("Round {round}: shake polynomial has {actual} coefficients, expected at most {max}")]
+    ShakePolynomialTooLong {
+        round: usize,
+        max: usize,
+        actual: usize,
+    },
+
+    /// A committed oracle is read through a Merkle multi-opening the proof must supply.
+    #[error("Round {round}: missing query openings")]
+    MissingQueryOpenings { round: usize },
```
Nashtare: **The final round's `round` label is one past the last real round.**

`fetch_round_rows` is called for the final-round queries with `error_round: num_rounds` (`verifier.rs:721`), so every variant carrying `req.error_round` — `MissingQueryOpenings`, `UnexpectedQueryOpenings`, `QueryOpeningCount`, `OpenedRowArity`, and through `external_fibers_in_draw_order` both `ExternalFiber*` — labels the final round with an index that does not exist.

Reproduced on the `shape_error_after` config (`make_params(1, 2)`, `log_degree = 8`), which has `num_rounds = 3` and `round_proofs.len() = 3`, setting `final_query_openings = None`:

```
Round 3: missing query openings
```

Rounds are 0..=2. The sentinel predates this PR — `error_round: num_rounds` already fed `InvalidMmcsProof` — but it was invisible while the shape error was a unit variant, and now it *is* the message, which is the thing this PR exists to get right.

`GrindStage` was introduced here for exactly this kind of disambiguation, so the same treatment would fit: carry `round: RoundLabel` with `RoundLabel = Round(usize) | Final`. At minimum, document the `num_rounds` == final-round convention on `RoundRowsRequest::error_round` and on each variant that carries it.

stir/src/error.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Why a proof's shape does not match what the configuration and the public input pin.
+///
+/// Every length here is pinned by data the verifier already holds:
+/// the round schedule and query counts come from [`crate::StirConfig`],
+/// the bucket and opening layout from the claimed evaluations.
+/// A mismatch is therefore a malformed proof, never a failed algebraic check —
+/// those get their own [`StirError`] variants.
+///
+/// `expected` is always the value the verifier derived; `actual` is what the proof carried.
```
Nashtare: This convention does not hold for three of the variants below.

- **`HeightClassCount`** — both sides are verifier-derived. `expected_ro_by_class.len()` (`pcs.rs:1019`) and `coeffs_by_height.len()` (`pcs.rs:975`) both trace back to `bucket_native_heights`, built from `commitments_with_opening_points` at `pcs.rs:921-938`. Nothing from the proof enters either number.
- **`ExternalFiberCount` / `ExternalFiberArity`** — `actual` is what the caller's `Src` closure returned, not what the proof carried.

The sentence above it — "A mismatch is therefore a malformed proof, never a failed algebraic check" — has the same three exceptions.

For the `ExternalFiber*` pair it is worth asking whether they belong under `InvalidProofShape` at all. A `Vec<Src>` of closures cannot be deserialised, so its shape is always caller-constructed rather than attacker-supplied; that is the same reasoning the crate applies to the caller-supplied slice lengths elsewhere in `verify_stir_multi_inner`. A separate `StirError::ExternalSource { .. }` would say "your fiber source misbehaved" instead of "this proof is malformed", which is what actually happened.

stir/src/error.rs
```diff
@@ -0,0 +1,274 @@
...
+        commitment: usize,
+        expected: usize,
+        actual: usize,
+    },
+
+    /// A bucket's native-height classes must match what its `Combine` challenge covers.
+    ///
+    /// `expected` is 1 when no `Combine` ran, since a lone class is folded by STIR directly.
+    #[error("Bucket 2^{log_height}: {actual} native-height classes, expected {expected}")]
+    HeightClassCount {
```
Nashtare: The description says `HeightClassCount` and `MissingCombineCoefficient` "guard verifier-internal consistency between two views of the same public claims; a prover cannot reach them." It is stronger than that: **both checks are tautologies, so neither variant can be constructed at all.**

- `expected_ro_by_class` is `vec![...; native_heights.len()]` (`pcs.rs:1019-1020`).
- `combine_info` is `Some` iff `native_heights.len() >= 2` (`pcs.rs:969-971`), and `coeffs_by_height` is `native_heights.iter().copied().zip(coeffs).collect()` over the already-deduped, already-sorted `native_heights` (`pcs.rs:975-977`) — so `coeffs_by_height.len() == native_heights.len()`.

So the `Some` arm compares `native_heights.len()` against itself, and the `None` arm — reached only when `native_heights.len() <= 1`, and every bucket has at least one class — compares it against 1. `MissingCombineCoefficient` then looks up keys drawn from the very `native_heights` that built the map.

Keeping them as defensive guards is defensible, but as written they add two public enum variants that are dead by construction rather than merely unreachable by a prover. `debug_assert!` would express the invariant without widening the public API; if they stay, the doc comments should say so plainly.

stir/src/error.rs
```diff
@@ -0,0 +1,274 @@
...
+    FinalPolyMismatch,
+
+    /// The proof's shape disagrees with the configuration or the public input.
+    #[error(transparent)]
+    InvalidProofShape(#[from] ProofShapeError),
+
+    /// A matrix in `commitment`'s batch was opened at zero points, so its width cannot be
+    /// pinned from the claimed evaluations. Every input matrix must be opened at >= 1 point.
+    #[error("Matrix {matrix} in commitment {commitment} was opened at zero points")]
+    MatrixWithoutOpeningPoints { commitment: usize, matrix: usize },
```
Nashtare, line 277: While `StirError` is being reorganised, this is the natural home for the one variant STIR is still missing relative to its siblings: **`OpeningPointMatchesQueryPoint`**.

`fri/src/verifier.rs` rejects an opening point that coincides with a queried fiber lane via `FriError::OpeningPointMatchesQueryPoint` — added deliberately in #1762 with a regression test — and `circle/src/pcs.rs` does the same via `InputError::OpeningPointMatchesQueryPoint`. STIR has no equivalent: the three `batch_multiplicative_inverse` calls in the PCS verify body (`pcs.rs:1052`, `:1256`, `:1434`) are unguarded, and `batch_multiplicative_inverse` is documented to panic on a zero input. So the condition aborts the process instead of returning `Err`.

Reachability: for STARK use `zeta` is sampled from the extension field, so a collision is negligible and cannot be steered cheaply; for a direct PCS caller opening at a base-field coset point it is a real process abort. No soundness impact — this is purely the panic-vs-`Err` invariant that the other two PCS implementations enforce and STIR does not.

Entirely fair to say this is out of scope if you would rather keep the PR a pure rename. Flagging it because it is a one-variant-plus-one-scan change and this PR creates the file it belongs in.

stir/src/verifier.rs
```diff
@@ -1086,7 +1106,11 @@ where
 
     for i in 0..b {
```
Nashtare, line 1127: Two lines up, `assert_eq!(proofs.len(), b, "one proof per instance")` (`:1105`) and `assert_eq!(v.len(), b, ...)` (`:1121`) are the only two asserts in this file, sitting directly above the `RoundCount` check that does the typed thing. Neither public entry point (`verify_stir_multi`, `verify_stir_multi_with_external_initial`) documents the panic.

Now that `ProofShapeError` exists, an `InstanceCount { expected, actual }` variant closes both. The in-tree caller is already guarded — `pcs.rs:835` returns a shape error on a proof-count mismatch — so this is API hygiene rather than a reachable bug, but it is precisely the hygiene this PR is about, and leaving two asserts behind in the file being cleaned up invites the same review comment next time.

stir/src/verifier.rs
```diff
@@ -1164,7 +1193,12 @@ where
             let rc = &configs[i].round_configs[local_r];
             let rp = &proofs[i].round_proofs[local_r];
             if rp.ood_answers.len() != rc.num_ood_samples {
-                return Err(StirError::InvalidProofShape);
+                return Err(ProofShapeError::OodAnswerCount {
+                    round: local_r,
```
Nashtare: `round` means two different things in this function, and only one of the two says so.

- `OodAnswerCount` here, and `check_ans_and_shake_lengths(local_r, ...)` at `:1277`, carry the instance's **local** round index.
- `ReplicatedWitnessMismatch` at `:1168` and `:1217` carries the **global** lockstep round `r` — which its doc comment states explicitly.

Whenever the batched instances have different round counts (`offset(i) != 0`), "Round 2: 3 OOD answers, expected 1" and "Round 2: replicated folding PoW witnesses disagree" name different rounds, and nothing in the first message says which convention it uses. Worth documenting on the variants carrying a local index, the way `ReplicatedWitnessMismatch` already does.

Separately, on the "Follow-up left out" section: agreed that adding an instance index uniformly across `InvalidPowWitness`, `InvalidMmcsProof` and friends is its own change. But `RoundCount` at `:1109` is the one site where the instance index `i` is already in hand, and it is also the check most likely to fire on a genuinely mismatched batch — so it may be worth taking there even ahead of the uniform pass.

stir/src/verifier.rs
```diff
@@ -1272,7 +1311,11 @@ where
     let final_folding_witnesses: Vec<F> =
         proofs.iter().map(|p| p.final_folding_pow_witness).collect();
     if final_folding_witnesses.windows(2).any(|w| w[0] != w[1]) {
-        return Err(StirError::InvalidProofShape);
+        return Err(ProofShapeError::ReplicatedWitnessMismatch {
+            round: max_m,
+            stage: GrindStage::Folding,
+        }
+        .into());
```
Nashtare, line 1351: **`b == 0` panics here instead of returning.**

With empty `configs`, both preamble loops are empty and `max_m == 0` (via `.max().unwrap_or(0)`), so the round loop is skipped entirely and `:1325` indexes an empty vector:

```
thread '...' panicked at stir/src/verifier.rs:1325:84:
index out of bounds: the len is 0 but the index is 0
```

Reproduced by calling `verify_stir_multi` with empty `configs` and `proofs`. `final_query_witnesses[0]` at `:1363` has the same shape and is simply never reached because this one fires first.

`TwoAdicStirPcs::verify` reaches it with an empty `commitments_with_opening_points` and an empty proof, since the `proof.len() != bucket_log_heights.len()` guard passes as `0 == 0`. This is caller-controlled rather than proof-controlled, so there is no soundness impact and no untrusted-input DoS — but `p3-fri`'s verifier handles the empty-commitment shape gracefully, so the asymmetry with the sibling crate is real.

An early `return Ok(Vec::new())` when `b == 0` is transcript-consistent with the prover, which grinds `unwrap_or(0)` bits and returns an empty `Vec`: `grind(0)` and `check_witness(0, _)` are both transcript no-ops.

stir/tests/stir.rs
```diff
@@ -746,6 +753,255 @@ mod babybear_stir {
...
+            "{err:?}"
+        );
+    }
+
+    /// Prove a fixed instance, apply `mutate` to the proof, and return the shape error the
+    /// verifier reports.
+    ///
+    /// Every mutation below breaks a length the configuration pins, so the verifier must
+    /// reject before any algebraic check runs — and must say which length it was.
+    fn shape_error_after(mutate: impl FnOnce(&mut StirProof<EF, MyMmcs, F>)) -> ProofShapeError {
```
Nashtare, line 900: This harness is a good shape — the cases below it read cleanly. A few gaps worth closing while it is fresh, roughly by value:

1. **`proof.final_query_openings = None`.** The only shape check on the final-round path with no test, and the one that surfaces the `Round 3` label I flagged in `error.rs`. Currently `test_missing_query_openings_rejected` pins `round: 0` only, so the final-round labelling is entirely unobserved.
2. **Both length mutations are push-only.** `ood_answers.push` and `final_polynomial.push` exercise one side of `!=`; the pop direction is a different failure for `FinalPolynomialLength` (a prover that truncates rather than pads) and costs one line each.
3. **`GrindStage::Query` is never asserted.** `test_multi_disagreeing_replicated_witness_rejected` pins `Folding`, so the new discriminant is only ever observed at one of its two values.
4. **Every `shape_error_after` case mutates `round_proofs[0]`.** The `round` field is therefore only ever asserted as 0, and a hard-coded `round: 0` would pass all of them. One case at `round_proofs[1]` would pin that the field is actually threaded.

stir/src/lib.rs
```diff
@@ -56,7 +58,8 @@ pub mod utils;
 pub mod verifier;
 
 pub use config::{Stage, StirConfig, StirConfigError, StirParameters, StirRoundConfig};
+pub use error::{GrindStage, ProofShapeError, StirError};
 pub use p3_security::whir::SecurityAssumption;
 pub use pcs::TwoAdicStirPcs;
 pub use proof::{StirProof, StirQueryOpenings, StirRoundProof};
-pub use verifier::{StirError, StirVerifyOutputs};
+pub use verifier::StirVerifyOutputs;
```
Nashtare, line 65: `p3_stir::verifier::StirError` no longer resolves — `verifier.rs` imports it privately now. The description's claim that nothing in the workspace used it checks out (the only in-repo reference was the `lib.rs` doc line this PR updates), but it is still a public path removal for downstream users, and the commit is typed `refactor(stir):` with no breaking marker for release-plz to pick up.

Either mark it (`refactor(stir)!:` or a `BREAKING CHANGE:` footer), or keep the path alive with a `pub use crate::error::StirError;` in `verifier` if you would rather not break it at all.

Unrelated nit, same area: the description says this is "exactly the shape `p3-uni-stark` already uses", but `InvalidProofShapeError` there uses `got` where this uses `actual`, and lowercase message prefixes (`"air {air}: ..."`) where this uses capitalised (`"Round {round}: ..."`). Not worth churn on its own — just noting the cited precedent is not matched exactly, and picking one now is cheaper than reconciling two conventions later.

Nashtare (approved):

stir/src/error.rs
tcoratger: Confirmed and fixed with the `RoundLabel` you suggested.

Reproduced first, on exactly your recipe:

```
round 3: missing query openings     # rounds are 0..=2
```

`RoundLabel::{Round(usize), Final}` now replaces the sentinel, and I took it across every round-carrying variant rather than only the shape ones — `InvalidPowWitness`, `InvalidMmcsProof`, `InvalidShakeConsistency` and `InvalidRoundConsistency` all fed `num_rounds`/`max_m` in the final round too, so fixing only `ProofShapeError` would have left the same wart one variant over.

`RoundRowsRequest::error_round` is now `round: RoundLabel`; `fetch_and_fold` passes `Round(round)`, `fetch_and_check` passes `Final`, so the sentinel has no way back in. Same message now reads:

```
final round: missing query openings
```

`test_missing_final_query_openings_rejected` pins it, and `test_tampered_final_pow_witness_fails` now asserts `round: RoundLabel::Final` instead of `{ .. }`.

The type also carries the local-vs-global note you raise separately below:

```rust
/// Which round an error refers to.
///
/// In a batch, a proof element carries its instance's own round index.
/// A shared grind carries the global lockstep round.
```

stir/src/error.rs
tcoratger: All three exceptions are gone, by removing the variants rather than weakening the sentence.

**`ExternalFiber*` → `StirError::ExternalSource`.** Your reasoning holds: a `Vec<Src>` of closures is not deserialisable, so the shape is caller-constructed. They now live in their own enum, and the wrapper name says whose bug it is:

```rust
/// Why the caller's external initial-oracle source did not answer as asked.
///
/// The source is a closure, so these report a caller bug rather than a malformed proof.
pub enum ExternalSourceError {
    SourceCount { expected: usize, got: usize },
    FiberCount { round: RoundLabel, expected: usize, got: usize },
    FiberArity { round: RoundLabel, fiber: usize, expected: usize, got: usize },
}
```

`SourceCount` is the third one, folded in from the `assert_eq!(v.len(), b)` you flag in the verifier — same class of failure, same home.

```
external initial oracle: round 0: source returned 18 fibers, expected 19
```

**`HeightClassCount`** is deleted (see the thread below) — so the sentence has no exceptions left and now reads:

```rust
/// `expected` is the value the verifier derived.
/// `got` is what the proof carried.
```

Every remaining variant holds that.

stir/src/error.rs
tcoratger: You are right, and both variants are deleted.

I did try to salvage the `None` arm. There is one way `native_heights.len()` can be 0: a commitment claimed with an *empty* matrix list gets `log_max_native = 0` from `unwrap_or(0)`, so its shared height is `log_blowup` and it opens a bucket with no classes. I probed it:

```
PROBE-B empty claims => Err(Config(StartingFoldingFactorExceedsDegree {
    log_starting_folding_factor: 2, log_starting_degree: 1 }))
```

`log_stir_degree` for such a bucket is `log_blowup.saturating_sub(log_blowup).max(1) == 1` regardless of parameters, so `get_or_try_compute_stir_config` rejects it before any closure runs whenever the starting folding factor is ≥ 2. So the arm is unreachable in practice as well as tautological.

Both are now `expect`s rather than `debug_assert!`s, because the two neighbouring invariants of exactly this kind in the same function already are:

```rust
.expect("bucket_native_heights is built from these claims")
.expect("point is in bucket_points by construction")
```

so the file stays consistent:

```rust
// Both `expect`s below rest on public input, never on the proof:
// `expected_ro_by_class` is sized by `native_heights`, `Combine` runs
// exactly when that holds two or more classes, and its coefficient map
// is keyed by those same heights.
match combine_info {
    None => Ok(expected_ro_by_class
        .into_iter()
        .next()
        .expect("a bucket without Combine holds exactly one class")),
    Some((r_comb, coeffs_by_height)) => {
```

and

```rust
let &(r_i, gap) = coeffs_by_height
    .get(&log_native_h)
    .expect("coeffs_by_height is keyed by native_heights");
```

Net: two fewer public variants, and the `None` arm's `unwrap()` gained a message it did not have.

stir/src/error.rs
tcoratger: Taken — it is a real process abort, so it goes in.

Reproduced across the whole coset before fixing. With `log_d = 2` the 2^3 LDE coset folds into two fibers of four lanes, so the first-round queries cover every position and *any* coset opening point hits:

```
PROBE-C p=0 => PANIC   (field/src/field.rs:1044, batch_multiplicative_inverse)
PROBE-C p=1 => PANIC
...
PROBE-C p=7 => PANIC
```

Now:

```
commitment 0, matrix 0, point 0: opening point coincides with a query point
```

Two notes on how I placed it.

**Only `pcs.rs:1052` was unguarded.** `:1256` is already safe — the `Combine` sweep substitutes a unit for the single degenerate lane (`1 - r_comb·x = 0` has one root, and the queried lanes are distinct coset points, so at most one can hit it). `:1434` is in `compute_inverse_denominators`, on the prover side under `open`, which cannot return a `Result`; `p3-fri` has the same asymmetry, so I left it — an honest caller opening at a coset point still panics in `open`, and an adversarial proof never runs `open`. Happy to guard that separately if you would rather.

**The check is a single scan, not a branch in the build loop.** `p3-fri` tests `is_zero()` per element while collecting; here the collection is an `extend` over `bucket_points`, so I kept it branchless and scan once afterwards:

```rust
if let Some(slot) = denom_diffs.iter().position(|d| d.is_zero()) {
    let (commitment, matrix, point) = locate_opening_point(
        commitments_with_opening_points,
        &bucket_points[slot % n_bp],
    );
    return Err(StirError::OpeningPointMatchesQueryPoint { commitment, matrix, point });
}
```

`bucket_points` is deduplicated, so its index is not caller-meaningful; `locate_opening_point` walks the claims on the error path only to recover the same `(commitment, matrix, point)` triple `FriError::OpeningPointMatchesQueryPoint` reports (`batch` → `commitment`, matching the existing `MatrixWithoutOpeningPoints`).

`test_pcs_rejects_opening_point_on_the_evaluation_domain` covers it.

stir/src/verifier.rs
tcoratger: Both asserts are gone. There are now no `assert!`s left in `verifier.rs`.

`proofs.len() != b` became `ProofShapeError::InstanceCount { expected, got }` as you suggest. The fiber-source one went to `ExternalSourceError::SourceCount` instead — it is a caller-supplied closure list, so it belongs with the other two source checks (see the `ExternalFiber*` thread), and it needed the `map_or_else` unrolled to a `match` to return rather than panic:

```rust
let mut external_fibers: Vec<Option<Src>> = match external_fibers {
    None => (0..b).map(|_| None).collect(),
    Some(sources) => {
        if sources.len() != b {
            return Err(ExternalSourceError::SourceCount { expected: b, got: sources.len() }.into());
        }
        sources.into_iter().map(Some).collect()
    }
};
```

Since neither entry point panics any more, there is nothing to document on them.

`test_multi_proof_count_mismatch_rejected` covers `InstanceCount`.

stir/src/verifier.rs
tcoratger: **The convention.** Documented on `RoundLabel` itself rather than per variant, so it is stated once where the reader meets the type:

```rust
/// Which round an error refers to.
///
/// In a batch, a proof element carries its instance's own round index.
/// A shared grind carries the global lockstep round.
```

I kept the split as it stands rather than normalising both to `r`: a proof element's round is the index you would use to find it in `round_proofs`, which is the instance's own, while a grind genuinely happens at a global step. `InvalidShakeConsistency` already used `local_r` and `InvalidPowWitness` already used `r`, so this is the existing behaviour, now labelled.

**Instance index on `RoundCount`.** Done:

```
instance 1: expected 3 round proofs, got 2
```

`test_multi_round_count_names_its_instance` pins it.

Two corrections to the framing, though.

It is not the only site with `i` in hand — `OodAnswerCount` (`:1197`) and `FinalPolynomialLength` (`:1345`) have it too, and so do the two initial-commitment arms. So "the one site" would have been an arbitrary stopping point, and I wanted the uniform version.

I tried the uniform version and it does not compile. The clean shape is a recursive wrapper:

```rust
#[error("instance {instance}: {source}")]
Instance { instance: usize, source: Box<StirError<MmcsError, InputError>> },
```

with `.in_instance(i)` applied at each per-instance `?`, which covers `InvalidPowWitness`, `InvalidMmcsProof`, `InvalidShakeConsistency`, `FinalPolyMismatch` and the shape errors raised inside the shared `fetch_and_fold`/`fetch_and_check` path in one go. It works — I had it passing, with `Instance { instance: 0, source: InvalidProofShape(...) }` coming back from the PCS — but interpolating `{source}` makes thiserror infer `Box<StirError<E, IE>>: Display` as a bound on `impl Display for StirError<E, IE>`, i.e. the impl requires itself:

```
error[E0275]: overflow evaluating the requirement `StirError<..>: std::fmt::Display`
```

Dropping the interpolation leaves the message as bare `"instance 0"`; `#[source]` instead would demand `IE: Error`, which `IE = ()` cannot satisfy. The escape is a hand-written `Display` + `Error` for all twelve variants, which trades the colocated `#[error(...)]` messages this crate uses everywhere for ~60 lines of `write!`. Not worth it here.

So: `RoundCount` takes `instance: Option<usize>` (uni-stark's `air: Option<usize>` pattern), and the uniform pass stays a follow-up — where the shape to reach for is threading `instance` through `RoundRowsRequest` and the two fetch methods, not the recursive variant.

stir/src/verifier.rs
tcoratger: Confirmed and fixed exactly as you propose.

```
thread 'probe_empty_batch' panicked at stir/src/verifier.rs:1325:84:
index out of bounds: the len is 0 but the index is 0
```

I checked the transcript-symmetry claim rather than taking it on trust — `grind` returns `F::ZERO` at zero bits and `check_witness` returns `true` at zero bits, both before touching the sponge, so an empty batch is a no-op on both sides and the early return is exact:

```rust
// An empty batch has no transcript operations of its own, matching the prover, whose
// shared grinds fall back to zero bits and are therefore transcript no-ops.
if b == 0 {
    return Ok(Vec::new());
}
```

It sits after the new `InstanceCount` check, so `configs: []` with a non-empty `proofs` is still rejected rather than silently accepted.

`test_multi_empty_batch_verifies` covers it, and the same input now returns `Ok([])` through `TwoAdicStirPcs::verify`.

stir/tests/stir.rs
tcoratger: All four closed, plus the assertions tightened throughout.

**1. `final_query_openings = None`** — `test_missing_final_query_openings_rejected`. This is what surfaced the `Round 3` label, so it now pins `RoundLabel::Final`.

**2. Pop direction** — `test_dropped_ood_answer_rejected` and `test_truncated_final_polynomial_rejected`, both asserting the `got < expected` side.

**3. `GrindStage::Query`** — the two multi witness tests now share a harness, so both discriminants are observed:

```rust
fn multi_shape_error_after(mutate: impl FnOnce(&mut StirProof<EF, MyMmcs, F>)) -> ProofShapeError
```

`test_multi_disagreeing_folding_witness_rejected` mutates `folding_pow_witness`, `test_multi_disagreeing_query_witness_rejected` mutates `pow_witness`. The second lands after the OOD absorb, which is what makes it a distinct check.

**4. A mutation off round 0** — `test_extra_ood_answer_in_a_later_round_rejected` mutates `round_proofs[1]` and asserts `RoundLabel::Round(1)`, so a hard-coded `Round(0)` no longer passes.

Separately: every shape assertion now compares the *whole* error value, no `..`. So the harness returns the error and each case is one `assert_eq!`:

```rust
#[test]
fn test_dropped_opened_row_rejected() {
    let err = shape_error_after(|proof| {
        proof.round_proofs[0].query_openings.as_mut().expect("...").row_evals.pop();
    });
    assert_eq!(
        err,
        ProofShapeError::QueryOpeningCount {
            round: RoundLabel::Round(0),
            expected: 21,
            got: 20,
        }
    );
}
```

The only `..` left in the file is on `InvalidMmcsProof`, whose `source: MmcsError` is not `PartialEq` — structural, not laziness.

88 tests in `tests/stir.rs`, up from 65 on main.

stir/src/lib.rs
tcoratger: **The breaking marker.** Marked, rather than kept alive. The follow-up commit is `refactor(stir)!:` with a footer:

```
BREAKING CHANGE: `StirError` moved from `p3_stir::verifier` to
`p3_stir::error`. The crate-root path `p3_stir::StirError` is unchanged.
```

It is breaking on more than the path now anyway — variants added (`ExternalSource`, `OpeningPointMatchesQueryPoint`), variants removed (`HeightClassCount`, `MissingCombineCoefficient`), and `round: usize` → `round: RoundLabel` — so a compat re-export would only have hidden one of several breaks.

**The convention nit.** Reconciled toward `p3-fri`/`p3-uni-stark` rather than toward the old `p3-stir` messages, since two of three already agreed:

- `actual` → `got` everywhere.
- `max` → `maximum`, matching `DegreeBitsTooSmall`'s `minimum`/`maximum`.
- Lowercase messages, including the pre-existing `StirError` ones (`"Invalid proof-of-work witness in round {round}"` → `"round 0: invalid proof-of-work witness"`).

Round and instance now lead the message, so the prefixes nest the way `air {air}: ` does:

```
round 0: ans polynomial has 1024 coefficients, expected at most 20
final round: missing query openings
instance 1: expected 3 round proofs, got 2
bucket 2^9, commitment 1: unexpected input opening
external initial oracle: round 0: source returned 18 fibers, expected 19
```

Cheaper now than later, as you say — thanks for catching it while the field names were still free to move.

tcoratger:
Rebased onto `main` (now includes #2005, #2006, #2007, #2018) and squashed to one commit — `mergeable: MERGEABLE`.

#2005 rewrote the parts of `pcs.rs` this touches, so a few notes on how the two met.

**Its new root-count check gets a variant.** `commitment.len() != plan.log_lde_heights.len()` was the one `InvalidProofShape` site that did not exist when this PR was written. It is now `ProofShapeError::CommitmentRootCount { commitment, expected, got }`:

```
commitment 0: expected 3 Merkle roots, got 1
```

**Its `has_at_bucket` became `group_idx: Option<usize>`**, so the missing/unexpected input-opening split now hangs off that instead — same two variants, upstream's structure.

**Its `debug_assert_eq!` on the class counts stays.** #2005 independently replaced the `Some`-arm class-count check with a `debug_assert_eq!`, which is the same conclusion reached in the thread above about that check being tautological. Kept as-is.

**But the two `expect`s went back to being errors.** I had converted both dead-by-construction checks to `expect`, on the grounds that the two neighbouring invariants in that function already were. On reflection that is the wrong trade for a verifier: an unreachable panic and an unreachable `Err` cost the same to write, and only one of them can take a node down if the reachability argument is ever wrong. So `HeightClassCount` and `MissingCombineCoefficient` are back as variants, and the `None` arm is now panic-free rather than `expect`-guarded:

```rust
None => {
    // No Combine ran, so STIR folds a single class directly.
    let got = expected_ro_by_class.len();
    let mut classes = expected_ro_by_class.into_iter();
    match (classes.next(), classes.next()) {
        (Some(only), None) => Ok(only),
        _ => Err(ProofShapeError::HeightClassCount { log_height: log_h, expected: 1, got }.into()),
    }
}
```

The reachability analysis in that thread still stands — neither is constructible — so this is belt-and-braces, not a bug fix.

**One of #2005's own tests got sharper.** `test_pcs_verify_rejects_a_relaid_out_proof_with_the_same_root_count` said the disagreement "only surfaces further in, on the dimensions the input MMCS opening is checked against". It does not — it surfaces on the round count, because the two layouts put different native heights in each bucket, so the second bucket derives a different `StirConfig`:

```
instance 1: expected 2 round proofs, got 1
```

The test now asserts that, and its comment says so. Same pattern as the native-height pair: the typed error contradicted a comment that the unit variant had let stand.

97 tests in `tests/stir.rs`. `clippy --all-targets -D warnings`, `doc` with `-D broken_intra_doc_links`, and `fmt --check` all clean.

PR #2013, opened 2026-09-03
perf(stir): sub-coset virtual oracle, cheaper multiplies, barycentric Ans check

Nashtare:
## Changes

- Materialize each round's virtual oracle on a degree-bounded sub-coset instead of the full domain, folding coefficients instead of evaluations from round 1 on.
- `combine_on_coset`: route geometric steps through base-field multiplies, apply the denominator inverse once per bucket instead of once per point.
- `fold_codeword`: route the twiddle multiply through base-field multiplication, drop a redundant input copy.
- **Wire-format change**: drop the shake polynomial from `StirRoundProof`, replace with a direct barycentric interpolation check (`Ans(ρ) == I(ρ)`). Smaller proofs, same soundness bound (still bounded by the pre-existing `shake_check_error`, now a strict over-estimate).
- Hardening: debug asserts pinning the sub-coset degree invariants, explicit empty-node handling in the new barycentric check.

Items 1-3 are bit-identical to `main` (verified via proof-byte hashing). Item 4 changes the proof format — no compatibility fixtures in this repo depend on it.

## Benchmarks

`stir_pcs` bench, this machine, main vs branch (20 samples):

|---|---|---|---|
| open/1bucket | 762 ms | 664 ms | −12.8% |
| open/2buckets_ratio2 | 894 ms | 673 ms | −24.7% |
| open/2buckets_ratio4 | 789 ms | 795 ms | +0.7% |
| open/3buckets_ratio4 | 849 ms | 729 ms | −14.1% |
| verify/1bucket | 2.38 ms | 2.37 ms | −0.5% |
| verify/2buckets_ratio2 | 2.59 ms | 2.55 ms | −1.7% |
| verify/2buckets_ratio4 | 2.57 ms | 2.51 ms | −2.5% |
| verify/3buckets_ratio4 | 2.66 ms | 2.64 ms | −0.7% |

### DIFF
--- a/stir/src/proof.rs
+++ b/stir/src/proof.rs
@@ -62,11 +62,13 @@ pub struct StirRoundProof<EF: Field, M: Mmcs<EF>, Witness> {
     /// Answer polynomial coefficients.
     ///
     /// The unique polynomial `Ans(X)` of degree `< |P|` interpolating `(y_i, v_i)` for every
-    /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. Sent by the prover so the
-    /// verifier avoids the O(|P|²) Newton interpolation; the verifier instead checks
-    /// `Ans(rho)` against the barycentric interpolant through `(y_i, v_i)` at a random `rho`.
-    /// `ans_polynomial` is observed in the transcript before `rho` is sampled, so a malicious
-    /// prover cannot fit `Ans` to a known `rho`.
+    /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. The verifier needs `Ans`
+    /// in coefficient form — every later round evaluates it on each queried fiber — so the
+    /// prover sends the coefficients and the verifier only confirms them, checking `Ans(rho)`
+    /// against the barycentric interpolant through `(y_i, v_i)` at a random `rho` instead of
+    /// rebuilding them by Newton's divided differences. `ans_polynomial` is observed in the
+    /// transcript before `rho` is sampled, so a malicious prover cannot fit `Ans` to a known
+    /// `rho`.
     pub ans_polynomial: Vec<EF>,
 
     /// Merkle openings for the STIR queries, sharing one pruned multi-opening proof.
--- a/stir/src/proof.rs
+++ b/stir/src/proof.rs
@@ -65,10 +65,10 @@ pub struct StirRoundProof<EF: Field, M: Mmcs<EF>, Witness> {
     /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. The verifier needs `Ans`
     /// in coefficient form — every later round evaluates it on each queried fiber — so the
     /// prover sends the coefficients and the verifier only confirms them, checking `Ans(rho)`
-    /// against the barycentric interpolant through `(y_i, v_i)` at a random `rho` instead of
-    /// rebuilding them by Newton's divided differences. `ans_polynomial` is observed in the
-    /// transcript before `rho` is sampled, so a malicious prover cannot fit `Ans` to a known
-    /// `rho`.
+    /// against the barycentric interpolant through `(y_i, v_i)` at a random `rho` at a lower
+    /// constant than rebuilding them by Newton's divided differences. `ans_polynomial` is
+    /// observed in the transcript before `rho` is sampled, so a malicious prover cannot fit
+    /// `Ans` to a known `rho`.
     pub ans_polynomial: Vec<EF>,
 
     /// Merkle openings for the STIR queries, sharing one pruned multi-opening proof.
--- a/stir/src/proof.rs
+++ b/stir/src/proof.rs
@@ -63,18 +63,11 @@ pub struct StirRoundProof<EF: Field, M: Mmcs<EF>, Witness> {
     ///
     /// The unique polynomial `Ans(X)` of degree `< |P|` interpolating `(y_i, v_i)` for every
     /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. Sent by the prover so the
-    /// verifier avoids the O(|P|²) Newton interpolation; correctness against the queried/OOD
-    /// values is enforced together with `shake_polynomial` at a random `rho`.
+    /// verifier avoids the O(|P|²) Newton interpolation; the verifier instead checks
+    /// `Ans(rho)` against the barycentric interpolant through `(y_i, v_i)` at a random `rho`.
+    /// `ans_polynomial` is observed in the transcript before `rho` is sampled, so a malicious
+    /// prover cannot fit `Ans` to a known `rho`.
     pub ans_polynomial: Vec<EF>,
-
-    /// Shake polynomial coefficients.
-    ///
-    /// `S(X) = sum_{y in P} (Ans(X) - Ans(y)) / (X - y)` where `P` is the set of all OOD +
-    /// queried points. Sent by the prover and checked by the verifier at a random evaluation
-    /// point: `S(rho) == sum_i (Ans(rho) - v_i) / (rho - y_i)`. Both `ans_polynomial` and
-    /// `shake_polynomial` are observed in the transcript before `rho` is sampled, so a
-    /// malicious prover cannot fit `Ans` to a known `rho`.
-    pub shake_polynomial: Vec<EF>,
 
     /// Merkle openings for the STIR queries, sharing one pruned multi-opening proof.
     ///
--- a/stir/src/prover.rs
+++ b/stir/src/prover.rs
@@ -295,12 +295,16 @@ where
 
     /// The prefix of `fold_coeffs` that can be non-zero.
     ///
-    /// `fold_coeffs` spans the fold domain, but the folded polynomial's true degree is bounded
-    /// by the round's degree schedule (a fixed factor smaller); evaluating only that prefix
-    /// cuts Horner's work by the same factor.
+    /// `fold_coeffs` runs to whatever length the fold's source implied — the fold domain when
+    /// the round folded a committed codeword, the folded sub-coset when it folded a virtual
+    /// witness — and either can reach past the folded polynomial's true degree, which the
+    /// round's degree schedule bounds. Evaluating only that prefix skips the trailing zeros.
     fn truncated_fold_coeffs(&self) -> &[EF] {
         let rc = &self.config.round_configs[self.round];
         let folded_degree_bound = 1usize << (rc.log_degree - self.log_arity);
+        // The clamp below would silently shorten the prefix rather than fail, so the bound is
+        // pinned here.
+        debug_assert!(folded_degree_bound <= self.fold_coeffs.len());
         &self.fold_coeffs[..folded_degree_bound.min(self.fold_coeffs.len())]
     }
 
@@ -406,6 +410,11 @@ where
         let log_witness_len = self.config.round_configs[self.round].log_degree - self.log_arity;
         let log_sub_coset = log_witness_len.max(log_answer_len).min(next_log_domain);
         let sub_coset_len = 1usize << log_sub_coset;
+        // No degree/folding schedule the config can produce makes the cap at the next domain
+        // binding. Were it ever to bind, the sub-coset would silently alias `f_{i+1}` or
+        // truncate the evaluations below rather than fail, so both widths are pinned here.
+        debug_assert!(log_witness_len <= log_sub_coset);
+        debug_assert!(log_answer_len <= log_sub_coset);
 
         let (ans_evals, vanishing_evals) = tracing::debug_span!("eval_low_degree_pair_on_coset")
             .in_scope(|| {
@@ -697,8 +706,13 @@ where
             }
             Oracle::Coeffs(coeffs) => {
                 // The fold's coefficients past `final_len` are zero, since the witness they
-                // come from respects the round's degree bound.
+                // come from respects the round's degree bound. The truncation below would
+                // silently drop a non-zero tail rather than fail, so the bound is pinned here.
                 let mut folded = fold_poly_coeffs(coeffs, final_gamma, final_log_arity);
+                debug_assert!(
+                    folded.len() >= final_len && folded[final_len..].iter().all(|c| c.is_zero()),
+                    "the folded witness must respect the final degree bound"
+                );
                 folded.resize(final_len, EF::ZERO);
                 folded
             }
--- a/stir/src/prover.rs
+++ b/stir/src/prover.rs
@@ -22,9 +22,8 @@ use tracing::instrument;
 use crate::config::StirConfig;
 use crate::proof::{StirProof, StirQueryOpenings, StirRoundProof};
 use crate::utils::{
-    compute_shake_polynomial, eval_poly_at_base, eval_poly_parallel, fold_codeword,
-    fold_domain_params, fold_poly_coeffs, interpolate_poly, next_domain_shift, sample_ood_points,
-    vanishing_poly_from_roots,
+    eval_poly_at_base, eval_poly_parallel, fold_codeword, fold_domain_params, fold_poly_coeffs,
+    interpolate_poly, next_domain_shift, sample_ood_points, vanishing_poly_from_roots,
 };
 
 /// Prove that a polynomial (given in coefficient form over `EF`) has low degree,
@@ -162,7 +161,7 @@ struct RoundOutput<F, EF: Field, M: Mmcs<EF>, Witness> {
 ///
 /// The phases, in order: \[grind\] `fold_and_commit` -> absorb commitment -> squeeze OOD points
 /// -> `ood_answers` -> absorb answers -> \[grind\] -> squeeze `r_comb` and query indices ->
-/// `record_queries` -> `ans_shake` -> absorb Ans/shake -> squeeze rho -> `finish`.
+/// `record_queries` -> `answer_poly` -> absorb Ans -> squeeze rho -> `finish`.
 struct RoundProver<'a, F, EF: Field, Dft, M: Mmcs<EF>, Challenger> {
     config: &'a StirConfig<F, EF, M, Challenger>,
     round: usize,
@@ -190,7 +189,6 @@ struct RoundProver<'a, F, EF: Field, Dft, M: Mmcs<EF>, Challenger> {
     seen_query_indices: Vec<usize>,
 
     ans_poly: Vec<EF>,
-    shake_poly: Vec<EF>,
     all_points: Vec<EF>,
 }
 
@@ -232,7 +230,6 @@ where
             query_answers: Vec::new(),
             seen_query_indices: Vec::new(),
             ans_poly: Vec::new(),
-            shake_poly: Vec::new(),
             all_points: Vec::new(),
         }
     }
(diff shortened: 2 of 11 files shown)


stir/src/prover.rs
```diff
@@ -332,112 +375,134 @@ where
...
+        // the evaluations below to be the polynomials themselves rather than a truncation.
+        let log_answer_len = log2_ceil_usize(num_answers + 1);
+        let log_witness_len = self.config.round_configs[self.round].log_degree - self.log_arity;
+        let log_sub_coset = log_witness_len.max(log_answer_len).min(next_log_domain);
+        let sub_coset_len = 1usize << log_sub_coset;
+        // No degree/folding schedule the config can produce makes the cap at the next domain
+        // binding. Were it ever to bind, the sub-coset would silently alias `f_{i+1}` or
+        // truncate the evaluations below rather than fail, so both widths are pinned here.
+        debug_assert!(log_witness_len <= log_sub_coset);
+        debug_assert!(log_answer_len <= log_sub_coset);
```
tcoratger: The `.min(next_log_domain)` here is dead, and I think that's the argument for deleting it rather than keeping it.

The bound is provable from the config. `log_folding_factor >= 2` and `log_starting_folding_factor >= 2` are validated at config.rs:548,553, so the fold-query domain is at least 4x smaller than `L_{i+1}`:

```
fold_log_domain = current − log_arity <= current − 2 = next_log_domain − 1
```

Queries are deduped, so `num_answers <= 2^fold_log_domain + num_ood` with `num_ood <= 2`, giving `num_answers + 1 <= 2^(next−1) + 3 <= 2^next`, i.e. `log_answer_len <= next_log_domain`. And `log_witness_len = rc.log_degree − log_arity <= rc.log_degree − 2 <= next_log_domain` directly. Your comment says the same thing, and neither assert fired anywhere in a 596-config sweep.

So the clamp can only ever act in a world where the invariant is already broken — and in that world it does the worst available thing. Provers ship `--release`, where the `debug_assert!`s below are compiled out and the clamp silently narrows the sub-coset; `coeffs_from_codeword` then returns the aliased residue of `f_{i+1}` and you get a well-formed proof that no verifier accepts. No panic, no locality, rejection several rounds downstream with nothing pointing back here.

```rust
let log_sub_coset = log_witness_len.max(log_answer_len);
assert!(log_sub_coset <= next_log_domain, "sub-coset wider than L_{i+1}");
```

One comparison per round, and a broken invariant becomes a loud prover-side panic instead of an unverifiable proof.

stir/src/prover.rs
```diff
@@ -278,23 +293,32 @@ where
...
+    /// witness — and either can reach past the folded polynomial's true degree, which the
+    /// round's degree schedule bounds. Evaluating only that prefix skips the trailing zeros.
+    fn truncated_fold_coeffs(&self) -> &[EF] {
         let rc = &self.config.round_configs[self.round];
         let folded_degree_bound = 1usize << (rc.log_degree - self.log_arity);
-        let truncated = &self.fold_coeffs[..folded_degree_bound.min(self.fold_coeffs.len())];
+        // The clamp below would silently shorten the prefix rather than fail, so the bound is
+        // pinned here.
+        debug_assert!(folded_degree_bound <= self.fold_coeffs.len());
+        &self.fold_coeffs[..folded_degree_bound.min(self.fold_coeffs.len())]
```
tcoratger: Same shape as the `log_sub_coset` clamp: the `.min()` is unreachable given the `debug_assert!` directly above it, and in release it silently shortens the Horner prefix instead of failing.

The bound holds either way. In the eval path `fold_coeffs.len() = 2^(current − log_arity)`; in the coefficient path it's `2^(log_sub_coset_prev − log_arity)`, and `log_sub_coset_prev >= log_witness_len_prev = rc.log_degree` for this round, so both are `>= 2^(rc.log_degree − log_arity)`.

```rust
&self.fold_coeffs[..folded_degree_bound]
```

lets the slice panic on a violated bound, which is what you want here — a short prefix silently corrupts both the OOD answers and (in the coefficient path) the query answers.

stir/src/prover.rs
```diff
@@ -629,30 +677,46 @@ where
...
+            Oracle::Coeffs(coeffs) => {
+                // The fold's coefficients past `final_len` are zero, since the witness they
+                // come from respects the round's degree bound. The truncation below would
+                // silently drop a non-zero tail rather than fail, so the bound is pinned here.
+                let mut folded = fold_poly_coeffs(coeffs, final_gamma, final_log_arity);
+                debug_assert!(
+                    folded.len() >= final_len && folded[final_len..].iter().all(|c| c.is_zero()),
+                    "the folded witness must respect the final degree bound"
+                );
+                folded.resize(final_len, EF::ZERO);
```
tcoratger, line 719: Third instance of the pattern, and both halves are provable here too: walking the `log_degree -= round_log_folding_factor` schedule (config.rs:851) down to the last intermediate round gives `log_witness_len = log_final_degree + k`, so `folded.len() >= 2^log_final_degree = final_len` always, and the tail past `final_len` is zero because folding a degree-`< 2^(log_final_degree + k)` witness by `2^k` lands below `2^log_final_degree`.

In release, `resize` zero-pads if it's short (harmless) but silently truncates a non-zero tail if it's long, which drops real degree. Worth splitting so the cheap half is unconditional:

```rust
assert!(folded.len() >= final_len);
debug_assert!(
    folded[final_len..].iter().all(|c| c.is_zero()),
    "the folded witness must respect the final degree bound"
);
```

stir/src/utils.rs
```diff
@@ -443,44 +430,75 @@ pub fn interpolate_poly<F: Field>(points: &[F], values: &[F]) -> Vec<F> {
...
+        .enumerate()
+        .map(|(i, &y)| {
+            points
+                .iter()
+                .enumerate()
+                .filter(|&(j, _)| j != i)
+                .map(|(_, &z)| y - z)
+                .product()
+        })
+        .collect();
```
tcoratger, line 486: Not a defect — this is correct and the measured verify win is real — but worth noting what it does to the design rationale, because I think a bigger win is sitting right here.

These are `n(n−1)` extension multiplies per round. At the bench's `security_level = 100, log_blowup = 1`, `n ≈ 110`, so ~12k EF products per round. Fine in absolute terms, and not a new complexity class either: the verifier already pays Θ(n²) in `vanishing_poly_from_roots` at verifier.rs:491. The reason verify still got *faster* is that absorbing `n` fewer field elements per round beats the extra arithmetic.

But it does mean confirming `Ans` and rebuilding it are now the same asymptotic cost — Newton's divided differences is maybe 2–3× the constant, not a different order. So shipping `ans_polynomial` on the wire now buys a constant factor of verifier time in exchange for three things: `n` extension elements per round, one transcript squeeze, and one term in the union bound (`ans_check_error`, counted in `TERMS_PER_INTERMEDIATE_ROUND` at config.rs:628).

Dropping `ans_polynomial` entirely and having the verifier interpolate would recover all three, land the paper's construction exactly, and shrink the proof by roughly the same amount again as removing the shake polynomial did. Given proof size is this PR's headline (and `assert_stir_proof_smaller_than_binary_fri` is guarding it), that looks like the natural follow-up. Definitely not for this PR.

tcoratger (commented):
Went through this one properly — read every line, then tried to break it three different ways rather than trust the read. It holds up. Writing up what I actually ran, because the evidence is more useful than my opinion.

**1. The sub-coset witness really is `DegCor((g_i − Ans)/Z)`, not an aliased truncation.**
I patched a probe into `RoundProver::finish` checking, at ~35 points per round:

```
eval(next_coeffs, x) == (eval(fold_coeffs, x) − Ans(x)) · Z(x)^{-1} · (1 − (r_comb·x)^{t+1})/(1 − r_comb·x)
```

That's the right probe: if `2^log_sub_coset` ever fell below `deg f_{i+1}`, `coeffs_from_codeword` would hand back the aliased residue and a random-point check fails. It passed on every existing test plus a 596-config sweep (`log_blowup ∈ 1..4`, `log_folding_factor ∈ 2..5`, `log_degree ∈ {6,8,10,12,14,15}`, `security_level ∈ {16,32,64,96,128}`, all three `SecurityAssumption`s), run `--release` with `-C debug-assertions=yes` so your own asserts were live too. Negative control (`gap+1 → gap+2`) fires instantly, so it isn't vacuous.

**2. Items 1–3 are bit-identical, confirmed independently.**
Built prover-output digests (FNV over every `ood_answers` / `ans_polynomial` / `row_evals` / `final_polynomial` limb) at `b2d6e2bd` and at `df370451` across 18 configs. All 18 match exactly. That's a nice result — the coefficient-form rewrite is arithmetically exact, not merely "verifies anyway".

**3. The barycentric check is sound.**
`Σ_i w_i/(rho − y_i) = 1/Π_i(rho − y_i)` because the Lagrange basis sums to one, so the cross-multiplied form never divides by zero for `rho` off the node set — which is exactly what the `points.contains(&rho)` rejection enforces. `check_ans_length` pins `deg(ans) ≤ n−1` and `deg(I) ≤ n−1`, so `ans − I` has at most `n−1` roots and the error is `(n−1)/|F|`; `ans_check_error` still charges `2n/|F|`, so parameter validation is unchanged and strictly conservative. Your description is accurate on this. Also checked that duplicate nodes can only arise where `interpolate_poly` would already have divided by zero on the prover side, so the new `false` return is a tightening, not a completeness regression.

**4. The degree invariants are provable, not just asserted.**
`log_folding_factor ≥ 2` / `log_starting_folding_factor ≥ 2` (config.rs:548,553) force the fold-query domain to be at least 4× smaller than `L_{i+1}`, which gives `num_answers + 1 ≤ 2^fold_log_domain + 3 ≤ 2^next_log_domain`; separately `log_witness_len ≤ rc.log_degree − 2 ≤ next_log_domain`; and the last intermediate round has `log_witness_len = log_final_degree + k`, which is what makes `fold_and_derive`'s zero tail guaranteed. So "no schedule the config can produce makes the cap binding" is correct — and I'd argue inline that this is a reason to *delete* the clamps rather than keep them.

---

Two things I'd change before merge, both cheap:

- **the three release-mode clamps** (inline) — each one turns a provable invariant into a silent wrong-witness path in exactly the build a prover ships;
- **the commit needs a `!`** — this drops a `pub` field from a `Serialize` type, removes two error variants and two `pub` fns, and changes the Fiat–Shamir transcript. `cliff.toml:16` keys off `commit.breaking`, and `p3-stir` landed in `release-plz.toml` in `b2d6e2bd` (the commit right below this branch), so the next release cuts a `p3-stir` version from these commits and would pick the wrong bump. `perf(stir)!: replace the shake polynomial…` on `447387c6`, or a `BREAKING CHANGE:` footer.

Everything else inline is a nit or a follow-up idea.

One thing worth calling out positively, since it's the easiest place in the diff to have introduced a subtle bug: the `combine_on_coset` restructure keeps the degenerate lane exactly right. When `step = 1` you get `r_step_hi = r_i · step^{gap+1} = r_i`, so `val · (r_i − r_step_hi) = 0`; the substituted `inv_denoms[p] = 1` leaves it at zero, and the closed form is added back after the sweep. Folding `r_i` into the geometric numerator so each point costs one product is a genuinely nice trick, and `test_combine_on_coset_matches_reference_at_degenerate_step` already pins the lane that the random-`r_comb` proptest would never reach. Good test to have had in place before this refactor.

stir/src/proof.rs
```diff
@@ -62,20 +62,15 @@ pub struct StirRoundProof<EF: Field, M: Mmcs<EF>, Witness> {
...
-    /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. Sent by the prover so the
-    /// verifier avoids the O(|P|²) Newton interpolation; correctness against the queried/OOD
-    /// values is enforced together with `shake_polynomial` at a random `rho`.
+    /// `y_i ∈ P` (OOD + queried points) and its claimed value `v_i`. The verifier needs `Ans`
+    /// in coefficient form — every later round evaluates it on each queried fiber — so the
+    /// prover sends the coefficients and the verifier only confirms them, checking `Ans(rho)`
+    /// against the barycentric interpolant through `(y_i, v_i)` at a random `rho` instead of
+    /// rebuilding them by Newton's divided differences. `ans_polynomial` is observed in the
+    /// transcript before `rho` is sampled, so a malicious prover cannot fit `Ans` to a known
+    /// `rho`.
```
tcoratger: Tiny wording thing, following from the Θ(n²) note on the barycentric weights: "the verifier only confirms them … instead of rebuilding them by Newton's divided differences" reads as though confirming is asymptotically cheaper. After this change it's the same order either way, just a smaller constant.

Maybe "…at a lower constant than rebuilding them by Newton's divided differences", so a future reader doesn't over-read the saving and rule out the drop-`Ans`-from-the-wire direction on false grounds.

stir/src/utils.rs
```diff
@@ -443,44 +430,75 @@ pub fn interpolate_poly<F: Field>(points: &[F], values: &[F]) -> Vec<F> {
...
     }
 
-    // If rho coincides with one of the evaluation points the denominator (rho - y_i) would be
-    // zero.  This is negligible for a random rho but we handle it defensively: the shake identity
-    // does not apply at such a degenerate rho, so we report failure.
+    // No node constrains `ans`, so the barycentric sums below would both be empty and accept
+    // anything. The empty node set is interpolated only by the zero polynomial.
+    if points.is_empty() {
+        return ans.iter().all(|c| c.is_zero());
+    }
```
tcoratger: These two sentences can't both hold — if no node constrains `ans`, then *every* polynomial interpolates the empty set, not only the zero one. The behaviour is the right one (reject rather than accept-all), it's just dressed up as a theorem.

```rust
// With no nodes the identity is vacuous and the sums below would accept anything, so
// reject everything but the zero polynomial.
```

Unreachable from the verifier anyway — `check_ans_length` with `max_ans_len == 0` already forces `ans` empty, and `num_ood_samples >= 1` — so this is pure defence in depth, which is fine.

stir/src/utils.rs
```diff
@@ -443,44 +430,75 @@ pub fn interpolate_poly<F: Field>(points: &[F], values: &[F]) -> Vec<F> {
...
+/// The identity `ans(rho) == I(rho)` is checked cross-multiplied, so the barycentric
+/// denominator is never inverted. That denominator equals `1 / prod_i (rho - y_i)` — the
+/// Lagrange basis sums to one — and so is non-zero for every `rho` outside the node set,
+/// which is exactly what the rejection below enforces.
+///
+/// Returns `true` if the check passes. For `n >= 1`, both `ans` (by the caller's length bound)
+/// and `I` have degree `< n`, so a caller that binds `ans` and the node set into the transcript
+/// before drawing `rho` gets soundness error at most `(n - 1) / |F|`. An empty node set is
+/// interpolated only by the zero polynomial, and is decided exactly rather than at `rho`.
+pub fn check_ans_interpolates<F: Field>(ans: &[F], points: &[F], values: &[F], rho: F) -> bool {
```
tcoratger, line 454: Worth `#[must_use]`: it's a `pub` predicate whose entire purpose is to be branched on, so a dropped result is always a bug. (`check_shake_consistency` didn't have it either, so this is a pre-existing gap that the rename is a good moment to close.)

Separately: `n == 1` is the only input that exercises the empty-`product()` weight path below, and nothing covers it. Unreachable from the verifier (`num_ood >= 1`, `num_queries >= 1`) but the fn is `pub`, and it's two lines:

```rust
assert!(check_ans_interpolates(&[ef(7)], &[ef(3)], &[ef(7)], ef(9)));
assert!(!check_ans_interpolates(&[ef(8)], &[ef(3)], &[ef(7)], ef(9)));
```

stir/src/utils.rs
```diff
@@ -528,42 +546,73 @@ pub fn fold_codeword<F: TwoAdicField, EF: ExtensionField<F>>(
...
+/// `gamma` of the `m`-th length-`k` block of `f`.
+///
+/// Unlike [`fold_codeword`], which interpolates at subgroup coordinates and therefore takes the
+/// rescaled challenge `gamma / shift`, this form carries no domain shift and takes `gamma`
+/// itself.
+pub fn fold_poly_coeffs<F: Field>(coeffs: &[F], gamma: F, log_arity: usize) -> Vec<F> {
+    coeffs
+        .par_chunks(1 << log_arity)
+        .map(|block| eval_poly(block, gamma))
+        .collect()
```
tcoratger, line 619: `#[must_use]` here too — owned `Vec` out of a parallel sweep.

Also worth one line in the doc that the final block can be short: `par_chunks` leaves a ragged tail when `coeffs.len()` isn't a multiple of `k`. It's still correct (a short block just means the missing high coefficients are zero), but that doesn't fall out of "the `m`-th length-`k` block of `f`", and this is `pub`.

stir/tests/utils.rs
```diff
@@ -354,3 +338,35 @@ fn test_fold_codeword_higher_arity_agrees_with_fold_fiber() {
...
+
+#[test]
+fn test_fold_poly_coeffs_agrees_with_the_evaluation_form_fold() {
+    use p3_dft::Radix2DitParallel;
+    use p3_stir::prover::{codeword_from_coeffs, coeffs_from_codeword};
+    use p3_stir::utils::{fold_codeword, fold_poly_coeffs};
+
+    let log_domain = 6;
+    let domain_size = 1usize << log_domain;
+    let dft = Radix2DitParallel::<F>::default();
```
tcoratger, line 376: This is the lemma the whole `Oracle::Coeffs` path rests on, and right now it's pinned at exactly one shape: `log_domain = 6`, `shift = GENERATOR`, dense degree 63, arities 1/2/3. Two production properties fall outside it:

- the input is usually **longer** than the true degree — when `log_answer_len > log_witness_len`, `log_sub_coset` gets bumped up and the coefficient tail is zero. This really happens; my sweep saw both `log_sub_coset = log_witness_len = 6, log_answer_len = 5` and `log_sub_coset = log_answer_len = log_witness_len = 5`.
- the shift is never `GENERATOR` after round 0, since `next_domain_shift` compounds `GENERATOR` in each round.

Good proptest shape — exact property, three-line setup, genuinely large input space:

```rust
// log_domain in 3..=8, log_arity in 1..=3 (<= log_domain),
// true_deg in 1..=2^log_domain, shift_seed in 1..P
let coeffs = /* true_deg random EF coeffs, zero-padded to 1 << log_domain */;
let codeword = codeword_from_coeffs(&dft, coeffs.clone(), shift, log_domain);
let folded = fold_codeword::<F, EF>(
    &codeword, gamma * EF::from(shift.inverse()), log_arity, log_domain);
prop_assert_eq!(
    fold_poly_coeffs(&coeffs, gamma, log_arity),
    coeffs_from_codeword(&dft, &folded, shift.exp_power_of_2(log_arity)),
);
```

A failing case would prove the coefficient fold and the evaluation fold disagree, which is the one thing that would make the new prover path wrong by construction. Today that only surfaces as "verification rejected", with nothing pointing at `fold_poly_coeffs`. It also transitively covers the `record_queries` split (`folded_codeword[j]` vs `eval_poly_at_base(truncated, x)`), which is otherwise unpinned.

While you're here: `fold_codeword`'s new `log_arity == 0` early return (utils.rs:549) is untested. Unreachable from STIR since `k >= 4`, but the fn is `pub` — `assert_eq!(fold_codeword::<F, EF>(&cw, beta, 0, 4), cw);`.

tcoratger (approved):

PR #2014, opened 2026-09-03
perf(multi-stark): precomputed alpha powers, coefficient-wise windows, packed accumulators

Nashtare:
## Changes

- Batch base-round constraints with precomputed alpha powers instead of an extension-field Horner fold.
- Load extension-round next-row windows with a coefficient-wise unaligned SIMD load instead of a per-lane gather.
- Hoist the block-invariant boundary selectors out of the per-block fold.
- Keep the packed zerocheck accumulators packed end-to-end instead of unpacking per weighting operation.
- Hardening: validate `BaseAir::num_constraints()` against the real constraint count at construction time (mirrors the existing guard in uni-stark/batch-stark), one shared symbolic pass instead of two.

All bit-identical to `main` (verified against the committed proof fixture).

## Benchmarks

`poseidon2_zerocheck_prove` bench, this machine, main vs branch (20 samples):

|---|---|---|---|
| 1_airs/15 | 311 ms | 252 ms | −18.9% |
| 5_airs/15 | 1.72 s | 1.39 s | −19.1% |
| 10_airs/15 | 3.45 s | 2.76 s | −19.9% |


multi-stark/src/folder.rs
```diff
@@ -25,6 +25,13 @@ pub struct MultilinearFolder<'a, F, Var, Acc> {
     pub alpha: Acc,
     /// Running alpha-batched accumulator capturing every asserted-zero constraint.
     pub accumulator: Acc,
+    /// Descending alpha powers `alpha^(n-1), ..., alpha^0`, one per asserted constraint.
+    ///
+    /// Empty when the caller supplies no powers, in which case the accumulator folds by Horner.
+    /// When non-empty the length must equal the number of constraints the AIR asserts.
+    alpha_powers: &'a [Acc],
```
tcoratger: Small semver note: `alpha_powers` and `constraint_index` are the first non-`pub` fields on `MultilinearFolder`, and every pre-existing field is `pub` in a `pub mod folder`. So downstream code that built this with a struct literal now fails to compile. `new()` makes the migration a one-liner and we're pre-1.0, so I don't think it's a problem — just flagging it so it's a choice rather than a side effect.

While you're here: if the fields are going private, `pub accumulator` and `pub alpha` arguably should too. An external write to either now desynchronizes it from `constraint_index`, which didn't used to be a thing that could happen.

multi-stark/src/folder.rs
```diff
@@ -182,14 +212,24 @@ where
...
     fn assert_zero<I: Into<Self::Expr>>(&mut self, x: I) {
-        // Horner alpha-batching: each push updates
-        //
-        //     accumulator := alpha * accumulator + C_i
-        //
-        // After `n` pushes the accumulator collapses to
-        //
-        //     C_0 * alpha^(n-1) + C_1 * alpha^(n-2) + ... + C_{n-1}.
-        self.accumulator = self.accumulator * self.alpha + x.into();
+        if self.alpha_powers.is_empty() {
```
tcoratger: The empty-slice sentinel conflates two different states: "no powers attached, fold by Horner" and "this AIR asserts zero constraints".

The consequence is a slightly odd asymmetry. An AIR hinting `num_constraints() == Some(0)` produces an empty `Vec` upstream, lands in this Horner branch, and therefore **accidentally agrees with the verifier** no matter how many constraints it actually asserts. Every other wrong hint is fatal; zero is the one value that silently works.

Not a bug — but `Option<&'a [Acc]>` would separate the two cleanly. `Some(&[])` then genuinely means "asserts nothing", and the `debug_assert` in `into_accumulator` catches a real mismatch instead of being vacuously satisfied by `is_empty()`.

multi-stark/src/folder.rs
```diff
@@ -182,14 +212,24 @@ where
...
+            //
+            //     C_0 * alpha^(n-1) + C_1 * alpha^(n-2) + ... + C_{n-1}.
+            self.accumulator = self.accumulator * self.alpha + x.into();
+        } else {
+            // Same sum reached term by term, weighting `C_i` by the precomputed `alpha^(n-1-i)`.
+            //
+            // The product is `Acc * Var` rather than `Acc * Acc`, which is the cheaper
+            // multiplication whenever the constraint value lives in a smaller ring than
+            // the accumulator.
+            self.accumulator += self.alpha_powers[self.constraint_index] * x.into();
```
tcoratger: This line is where `BaseAir::num_constraints()` stops being advisory and becomes correctness-critical, so it's worth spelling out what a wrong hint now does. I tested both directions with a `FibAir` clone that asserts 5 constraints and hints a different number:

```
hint = Some(6)  release  ->  Err(Zerocheck(FinalSumMismatch))     // honest trace, unverifiable proof
hint = Some(6)  debug    ->  panic at zerocheck.rs:240            // caught
hint = Some(4)  release  ->  panic "index out of bounds: the len is 4 but the index is 4"  (here)
```

The over-report case is the unpleasant one. The folder consumes only the first `n_real` powers, so it computes

```
sum_{i < n_real} alpha^(n_hint - 1 - i) * C_i  =  alpha^(n_hint - n_real) * Horner(n_real)
```

Round 0 therefore comes out scaled by `alpha^d` while every later round — and the verifier — folds by plain Horner, so the round-0 polynomial at `tau_0` no longer matches what round 1 can produce. Completeness break rather than a soundness one (the AIR is fixed by the verifier, so a malicious prover gains nothing), but it's silent, deterministic, and gives no diagnostic pointing at the hint.

Two things make it worth a second thought rather than a shrug:

1. **It's new.** On `main` the multi-stark prover never reads `num_constraints()` — the only consumer is `metadata.rs::ConstraintMetadata`, which has no callers outside its own tests. A wrong hint was previously harmless *here*.
2. **Nothing in-tree exercises it.** Every `BaseAir` impl returns `None`, so this path is only ever fed the exact symbolic count.

In fairness the trait doc at `air/src/air.rs:152` already says the count "**must** exactly match", and uni-stark uses the same debug-only guard — so the pattern matches precedent. I'd just suggest either promoting the check to a hard `assert_eq!` (one symbolic pass per AIR per proof is `O(num_constraints)` — nothing against a 2^15 trace, and it turns a silent unverifiable proof into a named failure), or softening the PR description, which currently reads as a runtime guarantee.

Minor aside on this same line: the indexed read is a bounds-checked slice access in the innermost loop. For a straight-line `eval` LLVM should merge the checks into one `len >= N`, so I doubt it's measurable — I didn't try to measure it.

multi-stark/src/zerocheck.rs
```diff
@@ -138,6 +138,42 @@ where
...
+    ///     periodic lengths passed empty → each periodic value capped at domain size two
+    ///                                   → degree one, exactly like a trace column
+    /// ```
+    degree: usize,
+    /// Number of asserted base-field constraints.
+    num_base_constraints: usize,
+}
+
+/// Evaluate an AIR symbolically once and score its constraints.
+fn symbolic_scores<F, EF, A>(air: &A) -> SymbolicScores
```
tcoratger, line 157: Worth deciding what happens to `metadata.rs` here. `ConstraintMetadata::from_air` already computes the same `(num_constraints, max_constraint_degree)` pair with the same single-pass structure and the same "both hints ⇒ no symbolic pass" logic — and it has no callers outside its own tests.

They're not interchangeable as written: `ConstraintMetadata` scores degree with `degree_multiple()`, `symbolic_scores` with `poly_degree(2, &[])` (the multilinear per-variable convention). So this is a genuine fork rather than a duplicate, which is exactly why I'd rather not leave both — either teach `ConstraintMetadata` the multilinear degree and drop `symbolic_scores`, or delete `metadata.rs` if it's superseded. Two scoring paths that agree today is how you get two that disagree in six months.

multi-stark/src/zerocheck.rs
```diff
@@ -158,37 +194,61 @@ where
...
+    F: Field,
+    EF: ExtensionField<F>,
+    A: Air<SymbolicAirBuilder<F, EF>>,
+{
+    let hinted_degree = air.max_constraint_degree();
+    let hinted_count = air.num_constraints();
+
+    match (hinted_degree, hinted_count) {
+        // Both hints present: outside debug builds nothing needs the symbolic pass.
+        (Some(degree), Some(count)) if !cfg!(debug_assertions) => (degree, count),
```
tcoratger: This arm short-circuits before any validation, which is what makes me want to double-check the intent: in a release build an AIR that hints **both** values is trusted unconditionally, and the count now drives the prover's constraint weighting. So the "validate `num_constraints()` at construction time" line in the PR description isn't quite what the code does — there's no construction-time validation in the builds people actually ship.

Empirically confirmed failure modes are in my comment on `folder.rs:230`.

Style nit while I'm here: `cfg!(debug_assertions)` as a match guard reads oddly — a compile-time constant doing runtime-guard duty, so the reader has to work out that the `_` arm is the *only* reachable arm in debug. An explicit `if cfg!(debug_assertions) { … }` ahead of the match, or a `#[cfg]`-split helper, states it more directly.

multi-stark/src/rounds.rs
```diff
@@ -56,6 +59,7 @@ impl<'air, 'data, A, F: Field> Stage<'air, 'data, A, F> {
         preprocessed: &[Option<&'data Table<F>>],
         tables: &[&'data Table<F>],
         degrees: &[usize],
+        num_constraints: &[usize],
```
tcoratger: `Stage::new` is now up to three positional `&[usize]` parameters (`indices`, `degrees`, `num_constraints`) in a seven-argument signature. `degrees` and `num_constraints` are adjacent and type-identical, so swapping them at the call site compiles cleanly — and the resulting proof fails with `FinalSumMismatch` and no hint as to why.

Cheap fix that also removes the parallel-`Vec` bookkeeping (`unzip` in `AirZerocheck`, then the re-index per stage):

```rust
struct AirShape { degree: usize, num_constraints: usize }
// ... pass &[AirShape]
```

multi-stark/src/rounds.rs
```diff
@@ -252,6 +267,28 @@ fn packed_window<F: Field, EF: ExtensionField<F>>(
     tail: EF,
 ) -> EF::ExtensionPacking {
     let packing_width = F::Packing::WIDTH;
+    debug_assert_eq!(
+        column.len() * packing_width,
```
tcoratger, line 282: Could this become a hard `assert_eq!`, or better, get hoisted?

The reasoning behind the fast path below is sound and I checked it — but note the behavioural change: the old code recomputed every lane and consulted `tail` unconditionally, so a caller violating this invariant still got a correct answer. The new fast path returns **padding lanes in place of the tail**, silently, and the only thing standing between us and that is an assertion compiled out of every shipped build.

`Poly` makes it unreachable today. My suggestion: hoist one hard `assert_eq!(column.len() * packing_width, height)` to the top of `RoundStateExt::round_poly_packed`, before `par_fold_reduce` — that's one check per round rather than per `(block, column)`, so it's free — and leave this `debug_assert` as documentation.

multi-stark/src/rounds.rs
```diff
@@ -252,6 +267,28 @@ fn packed_window<F: Field, EF: ExtensionField<F>>(
...
+        len,
+        "packed_window assumes a fully populated column"
+    );
+    let group = start / packing_width;
+    let offset = start % packing_width;
+
+    // With a successor group in range the window stays inside `group` and `group + 1`,
+    // so no lane can reach the tail.
+    if group + 1 < column.len() {
```
tcoratger, line 291: For the record, I did verify this: with `len == column.len() * W`, the window spans

```
group*W + offset  ..  group*W + offset + W - 1  <  (group + 2)*W  <=  column.len()*W  ==  len
```

so no lane can reach the tail, and `offset < W` gives `offset + W <= 2W == flat.len()`, so the slice below is in bounds. The alignment question also checks out — `F::Packing::from_slice` is a bare pointer cast with only a length assert, but `PackedMontyField31AVX2` is `#[repr(transparent)]` over `[MontyField31; WIDTH]` (align 4), so the unaligned read is well-defined.

The existing test is good — sweeping every `start` in `0..8W` is the right shape — but it fixes `column.len() == 8`, so `column.len() == 1` and `== 2`, where this branch is entirely bypassed, never get hit. A proptest over `num_groups in 1..=8` would close that.

multi-stark/src/rounds.rs
```diff
@@ -316,9 +353,11 @@ struct Scratch<F, EF> {
...
 /// Mirrors the scalar scratch with packed row buffers, so each element covers one SIMD lane group.
 /// One instance is allocated per worker and reused across that worker's packed blocks.
-struct PackedScratch<P, EF> {
-    /// Unweighted per-node evaluation accumulator for each AIR.
-    air_evals: Vec<Vec<EF>>,
+struct PackedScratch<P, Acc> {
+    /// Per-node evaluation accumulator for each AIR, one lane per row of the covered blocks.
+    ///
+    /// Each lane already carries its own eq weight, so the lanes are summed once at the end.
+    air_evals: Vec<Vec<Acc>>,
```
tcoratger, line 371: This is my best guess at why the speedup doesn't reproduce at AVX-512, so worth a thought: `air_evals` went from `Vec<Vec<EF>>` to `Vec<Vec<EF::ExtensionPacking>>`, which at `W=16, D=4` is 256 B per entry instead of 16 B. Each worker's accumulator set grows 16×, and the `par_fold_reduce` merge in `add_air_evals` moves 16× more bytes per reduction.

Together with `alpha_powers` (one broadcast packed extension per constraint — multi-KB for Poseidon2, streamed once per interpolation node per block), that's two of the five changes whose cost scales *with* the packing width while the saving stays roughly fixed. Which would explain a win that's healthy at `W=8` and gone at `W=16`.

Entirely speculative on my part — but it'd be worth measuring `1_airs` at `+avx2` vs `+avx512f` before settling on the numbers in the description.

multi-stark/src/rounds.rs
```diff
@@ -733,6 +805,27 @@ where
             assert_eq!(public_values.len(), air.num_public_values());
         }
 
+        // Constraint `i` of an AIR asserting `n` constraints is batched with weight `alpha^(n-1-i)`,
+        // the same weight the folder's Horner fold assigns it.
+        //
+        // Only the packed first round reads these, so the scalar path builds none.
+        let alpha_powers = if (1 << num_vars) / 2 >= F::Packing::WIDTH {
```
tcoratger: Confirmed this gate matches the dispatch exactly: `(1 << num_vars) / 2 >= F::Packing::WIDTH` is the precise complement of `num_evals() / 2 < F::Packing::WIDTH` in `round_poly`, so the powers are built exactly when the packed path will read them. And the ordering is right — `powers().collect_n(n)` reversed gives `alpha_powers[i] = alpha^(n-1-i)`, which is exactly the weight the Horner recurrence assigns after `n` pushes.

One thought: the *unpacked* base round would benefit from the same trick (`EF * F` is `D` multiplies against `D^2`-ish for `EF * EF`), but it can't use this table because of the type. Only matters for traces below `2*WIDTH` rows, so probably not worth the second `Vec` — noting it in case you disagree.

multi-stark/src/rounds.rs
```diff
@@ -1405,19 +1497,45 @@ where
             .collect::<Vec<_>>();
         assert_ne!(packed_half, 0);
 
+        // A block's selector pair depends on the block only through the residual rows `0` and
+        // `height - 1`. Row `0` sits in the low half of block `0`; row `height - 1` sits in the
+        // high half of block `packed_half - 1`. Every other block reads the same constant pair,
+        // so only these two need the per-row construction.
+        let first_block =
```
tcoratger, line 1553: Nice — and the derivation is more robust than the test comment suggests (see my note in the review body: it holds for *any* prefix, not just one obeying `transition = 1 - last`).

The suggestion: the identical hoist applies to `RoundStateBase::round_poly_packed`, which still rebuilds the pair per block at line 984:

```rust
let (mut boundary, boundary_diff) =
    BoundaryEvals::<F::Packing>::row_pair_packed(s, scalar_half, height);
```

Same argument verbatim — `row_pair_packed` → `from_packed_row` builds three `F::Packing::from_fn` closures per call, six per block, each doing `W` comparisons plus a `from_bool`. For a 2^15 trace at `W=16` that's ~98k boolean constructions per round collapsing to one hoisted constant triple. Since you've already written and tested the reasoning here, the base-round version is nearly free to add.

multi-stark/src/rounds.rs
```diff
@@ -1601,3 +1711,49 @@ where
...
+    use rand::{RngExt, SeedableRng};
+
+    use super::*;
+
+    type F = BabyBear;
+    type EF = BinomialExtensionField<F, 4>;
+    type Packing = <EF as ExtensionField<F>>::ExtensionPacking;
+
+    #[test]
+    fn packed_window_matches_the_scalar_gather() {
```
tcoratger, line 1781: Good test — this is the right shape for `packed_window`. The bigger gap is one level up: there's no **packed-vs-unpacked equivalence test** for either `round_poly`, and this PR changes four things at once inside both packed kernels (batching arithmetic, selector source, accumulator type, window loader). Right now the only thing checking them is end-to-end proof verification, which does catch errors but surfaces them as a `FinalSumMismatch` twenty layers up, and can't be run at both sides of the packing threshold.

Both halves are reachable from here — `round_poly_unpacked` is `pub(crate)` on `RoundStateBase`, and the `RoundStateExt` pair is same-module:

```
Property: for one RoundStateBase built from a random multi-AIR stage and a random eq_suffix,
          round_poly_packed(eq) == round_poly_unpacked(eq), elementwise.

Inputs:   2-3 AIRs of differing degree and width, at least one with preprocessed columns and
          one with periodic columns; num_vars swept over log2(2*W) ..= log2(2*W)+3 so the
          packed path sees packed_half = 1, 2, 3, 4 -- the 1 and 2 cases pin the
          first_block/last_block edges, the 3+ cases pin interior_block.

Proves:   alpha-power batching == Horner, hoisted selector triple == per-block triple, and
          packed accumulator reduction == scalar dot-product fold -- each localized, instead
          of all three aliased into one end-to-end mismatch.
```

Known-answer vectors rather than proptest here, I think: the two paths are each other's oracle, so a fixed sweep over the boundary shapes is cheaper *and* more legible than randomizing `num_vars`.

Two smaller gaps while we're on tests: no AIR in the tree hints `num_constraints()` (all `None`), so the release fast arm in `zerocheck.rs` is unreachable in-repo — one `FibAir` returning `Some(5)` would prove the symbolic-pass-skipping path actually works. And a `#[should_panic(expected = "num_constraints()")]` test for a mismatched hint is three lines.

tcoratger (commented):
Really nice piece of work — five cleanly separated commits, each with a legible motivation, and the arithmetic all holds up. I went through it fairly carefully (derived the two non-obvious invariants by hand, regenerated the fixture, and re-ran the bench) and came out convinced the transformation is correct. Two things I'd like you to double-check before this lands, plus a handful of smaller notes inline.

## The benchmark table doesn't reproduce on my machine

This is my main ask. I ran the same bench, interleaving base/branch **inside a single worktree** — that turned out to matter a lot: separate worktrees gave me a spurious 5% split, and a naive sequential sweep drifted 4.5% purely from AVX-512 downclocking. The drift is the same magnitude as the effect we're trying to measure, so any A/B here needs alternation.

`x86_64`, `RUSTFLAGS="-C target-cpu=native"` (AVX-512, so `F::Packing::WIDTH == 16`):

| bench | base (`b84529c0`) | branch (`90db8430`) | Δ | claimed |
|---|---|---|---|---|
| `1_airs/15` | 115.94 / 117.90 / 116.60 / 119.06 → **117.4 ms** | 121.13 / 117.62 / 113.77 / 115.97 → **117.1 ms** | −0.2% | −18.9% |
| `10_airs/15` | 1.2135 / 1.2427 / 1.2531 → **1.2364 s** | 1.1916 / 1.1964 / 1.2195 → **1.2025 s** | −2.7% | −19.9% |

`10_airs` is a genuine win — the branch was faster in all three pairs — but 2.7%, not 20%. `1_airs` is inside the noise.

I suspect the difference is the packing width the numbers were taken at. With **no** `RUSTFLAGS`, `cargo bench` in this repo builds baseline x86-64, where `<BabyBear as Field>::Packing::WIDTH == 1` — I measured it directly. In that build the whole "packed" kernel collapses to scalar, and I get 536 ms → 528 ms (−1.4%). So neither the degenerate build nor AVX-512 gets near 19%; maybe AVX2 (`W=8`) is the sweet spot?

There's a plausible mechanism for the win shrinking as the width grows, which I've noted inline on `PackedScratch::air_evals`: two of the five changes scale *against* the width.

Could you share the exact `RUSTFLAGS`/target-cpu, core count, and whether the runs were interleaved? I'd much rather reconcile the measurement than argue about it — and if the 19% only shows at a narrower width, that's worth saying in the description, since AVX-512 is what people tune for.

Repro:
```bash
git worktree add --detach /tmp/ab b84529c0
for i in 1 2 3; do for sha in b84529c0 90db8430; do
  (cd /tmp/ab && git checkout -q $sha && echo -n "$sha: "
   RUSTFLAGS="-C target-cpu=native" cargo bench -q -p p3-multi-stark --bench poseidon2 -- 1_airs \
     2>&1 | grep -oE "\[.*\]" | head -1)
done; done
```

## `num_constraints()` becoming load-bearing

The other thing worth a second look: the count went from advisory to correctness-critical in this PR, and the validation is debug-only — details and the two empirically-confirmed failure modes are inline on `folder.rs` and `zerocheck.rs`. Matching uni-stark's `debug_assert` is defensible; I mostly want to know it was a deliberate choice.

## What I checked and am happy with

Recording this so you know what got real scrutiny rather than a skim:

- **The `interior_block` constant pair is right** — and actually more robust than your test comment claims. For an interior block at `row = b*W`, `1 <= b <= packed_half - 2`: every low row lands in `[W, half-W-1]` and every high row in `[half+W, height-W-1]`, so `first = prefix.first * 0 = 0`, `last = prefix.last * 0 = 0`, and `transition = 1 + 0 * prefix.transition = 1`, giving `diff = (0,0,0)`. That holds for **any** prefix — it doesn't need the `transition = 1 - last` invariant, because `prefix.transition` is gated by `[row+lane+1 == height]`, which is zero on every interior row. The `packed_half == 1` and `== 2` edges are fine since `packed_s == 0` is tested first.
- **`packed_window`'s fast path is sound.** With `len == column.len() * W`, the window spans `group*W + offset .. group*W + offset + W - 1 < (group+2)*W <= len`, so no lane can reach the tail. Bounds are fine too (`offset < W` ⇒ `offset + W <= 2W == flat.len()`). I also checked the alignment question, since `F::Packing::from_slice` does a bare `unsafe { &*slice.as_ptr().cast() }` with only a length assert — it's fine because `PackedMontyField31AVX2` is `#[repr(transparent)]` over `[MontyField31; WIDTH]`, i.e. align 4, so the unaligned read at `flat[offset..]` is well-defined.
- **`eq_suffix * g` is exactly the old dot product.** Old: `sum_lane eq[l] * g[l]` into a scalar. New: lane-wise packed multiply into a packed accumulator, then one horizontal sum. Same value, `~W` fewer instructions.
- **The alpha-power ordering matches Horner.** `powers().collect_n(n)` reversed gives `alpha_powers[i] = alpha^(n-1-i)`, precisely the weight `acc := acc*alpha + C_i` assigns after `n` pushes. And the win is real: `Acc * Var` is `D` packed base multiplies against `D^2`-plus for `Acc * Acc`.
- **The constraint count really is 1:1.** Every `AirBuilder` convenience (`assert_eq`, `assert_one`, `assert_bool`, `assert_zeros`, `assert_eq_arrays`, and `FilteredAirBuilder`'s `when*` wrappers) routes through `assert_zero`, and `SymbolicAirBuilder::assert_zero` is a bare `push` with no folding or dedup. `MultilinearFolder` overrides nothing else. So for a fixed layout the two counts can't diverge.
- **The gate matches the dispatch** — `(1 << num_vars) / 2 >= WIDTH` is the exact complement of `num_evals() / 2 < WIDTH` in `round_poly`. And you correctly *don't* attach powers in the ext round, where `Var == Acc` and there'd be nothing to gain.
- **Bit-identical: confirmed**, though not by the test the description points at. `verify_whir_compat_fixture` only checks that the *committed* proof still verifies — it never compares prover output. I ran `generate_whir_fixture` on the branch and the md5 was unchanged (`0a355e20…`). That case does exercise the changed kernels (`n = 256`, `W = 16` ⇒ `packed_half = 8` in the base round, packed ext rounds down to height 32), so the claim stands — just via a different route than stated.
- `cargo clippy -p p3-multi-stark --all-targets` clean; full `cargo test -p p3-multi-stark --release` passes.

## One test I'd really like to see

Both `round_poly_packed` kernels changed materially here — new batching arithmetic, new selector source, new accumulator type, new window loader — and the only thing checking them is end-to-end proof verification. That does catch errors, but as a `FinalSumMismatch` twenty layers up, and it can't be run at both sides of the packing threshold. Sketch is inline on the new test module.

Nothing here is a soundness concern, and none of it is hard to address. Thanks for the careful commit split — it made the bisect trivial. 🙏

Nashtare:
Regarding the benchmark concern, I ran those on my M4 max, not any AVX machine.
Happy to iterate further to try pushing the x86 perfs

tcoratger:
Ah, that explains it cleanly — thanks. NEON gives `BabyBear::Packing::WIDTH == 4`; the machine I measured on is AVX-512, so `WIDTH == 16`. That is a 4x difference in the packed width, and it lines up with the mechanism I guessed at in my note on `PackedScratch::air_evals`:

- `air_evals` went from `Vec<Vec<EF>>` to `Vec<Vec<EF::ExtensionPacking>>`, so each entry is `W * D` field elements instead of `D`. At `W=4, D=4` that is 64 B; at `W=16` it is 256 B, and the `par_fold_reduce` merge moves 16x more bytes per reduction than the scalar version did.
- `alpha_powers` is one broadcast packed extension per constraint, so the table is also `W`-proportional — multi-KB for Poseidon2, streamed once per interpolation node per block.

Both costs scale with `W` while the saving per constraint (one `Acc * Var` instead of one `Acc * Acc`) stays fixed. So a healthy win at `W=4` collapsing to noise at `W=16` is exactly the shape you would expect. Nothing wrong with your numbers — they just need an architecture label, since a reader will assume x86 by default.

For the record, my interleaved measurements at `W=16` (4 alternating pairs in one worktree, `-C target-cpu=native`):

| bench | base | branch | Δ |
|---|---|---|---|
| `1_airs/15` | 117.4 ms | 117.1 ms | −0.2% |
| `10_airs/15` | 1.2364 s | 1.2025 s | −2.7% |

So x86 is neutral-to-slightly-positive, not a regression. That is fine for merging — it is a real win on aarch64 and costs nothing elsewhere. If you do want to chase the x86 side later, the two width-proportional costs above are where I would look first; keeping `air_evals` scalar and only the *weighting* packed might get both regimes.

## Re-grading my other point, and merging

On reflection I had the `num_constraints()` finding one notch too high. Downgrading it to minor, because:

- `air/src/air.rs:152` already states the contract in the strongest terms ("**must** exactly match ... will cause the prover to panic or produce an invalid proof").
- `uni-stark/src/prover.rs:145` and `batch-stark/src/prover.rs:377` use the identical debug-only guard, so this follows established precedent rather than inventing a weaker one.
- No in-tree AIR supplies the hint, so nothing shipping is affected today.
- It is a completeness failure, never a soundness one — the AIR is fixed by the verifier, so a malicious prover gains nothing from a wrong count.

What stands is that the PR body describes it as validation "at construction time", which reads as a runtime guarantee it does not provide. That is a wording fix plus a cheap hardening, not a blocker.

So: merging this, and I will send the follow-ups as a separate PR — the hard count assert, replacing the empty-slice sentinel with an explicit optional so `Some(0)` stops being the one silently-safe wrong hint, and the packed-vs-unpacked round-polynomial equivalence test. The base-round selector hoist, the `AirShape` cleanup, and the `metadata.rs` question I will leave as notes on that PR rather than bundling them in.

Thanks for the quick turnaround on the benchmark question.

tcoratger (approved):
Approving — the transformation is correct and bit-identical (verified by regenerating the committed WHIR proof fixture), the benchmark discrepancy is explained by architecture, and the remaining items are minor and going into a follow-up PR. Details in the comment above and the inline notes.

PR #2015, opened 2026-09-03
perf(fri): closed-form fold_row, in-place fold_matrix, hoisted verifier alpha ladder, fused hiding LDE

Nashtare:
## Changes

- `fold_row`: closed-form arity-2 fast path instead of generic Lagrange interpolation (verifier hot path).
- `fold_matrix`: fold the borrowed input matrix in place for high arities instead of cloning it up front, buffer-swap instead of copy between fold steps.
- `open_inputs`: hoist the query-invariant alpha ladder out of the per-query loop; per-column cost drops from EF×EF to EF×base-field (verifier hot path).
- `HidingFriPcs::get_quotient_ldes`: fuse the chunk LDE and mask transforms into a single forward DFT.
- Parallelize the reduced-opening roll-in in `commit_phase`.
- Benchmark fix: stop timing a matrix clone inside the measured closure.

All bit-identical to `main` (verified via differential testing against independent reference implementations for every commit, including a per-DFT-backend regression test for the hiding LDE fusion).

## Benchmarks

`fold_matrix` bench (arity 1, all field types), this machine, main vs branch (20 samples):

| field | size | main | branch | Δ |
|---|---|---|---|---|
| BabyBear | 4K–4M | 11.5 µs–13.4 ms | 11.1 µs–12.9 ms | −3.6% to −4.9% |
| Goldilocks | 4K–4M | 16.0 µs–19.2 ms | 14.0 µs–18.2 ms | −4.4% to −12.3% |
| Mersenne31^2 | 4K–4M | 32.2 µs–37.4 ms | 30.2 µs–35.5 ms | −3.1% to −6.3% |

### DIFF
--- a/fri/benches/fold_even_odd.rs
+++ b/fri/benches/fold_even_odd.rs
@@ -1,4 +1,5 @@
 use core::marker::PhantomData;
+use std::hint::black_box;
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use p3_baby_bear::BabyBear;
@@ -13,6 +14,8 @@ use rand::distr::{Distribution, StandardUniform};
 use rand::rngs::SmallRng;
 use rand::{RngExt, SeedableRng};
 
+const LOG_ARITIES: [usize; 3] = [1, 2, 3];
+
 fn bench<F: TwoAdicField, EF: ExtensionField<F>>(c: &mut Criterion, log_sizes: &[usize])
 where
     StandardUniform: Distribution<EF>,
@@ -22,18 +25,21 @@ where
     group.sample_size(10);
     let folding = TwoAdicFriFolding::<(), ()>(PhantomData);
 
-    for log_size in log_sizes {
-        let n = 1 << log_size;
+    for log_arity in LOG_ARITIES {
+        let arity = 1 << log_arity;
 
-        let mut rng = SmallRng::seed_from_u64(n as u64);
-        let beta = rng.sample(StandardUniform);
-        let mat = RowMajorMatrix::<EF>::rand(&mut rng, n, 2);
+        for log_size in log_sizes {
+            let n = 1 << log_size;
 
-        group.bench_function(BenchmarkId::from_parameter(n), |b| {
-            b.iter(|| {
-                folding.fold_matrix(beta, 1, mat.clone());
+            let mut rng = SmallRng::seed_from_u64(n as u64);
+            let beta = rng.sample(StandardUniform);
+            // The input always holds `2 * n` evaluations, laid out as `arity` per folded output.
+            let mat = RowMajorMatrix::<EF>::rand(&mut rng, (2 * n) / arity, arity);
+
+            group.bench_function(BenchmarkId::new(format!("log_arity_{log_arity}"), n), |b| {
+                b.iter(|| black_box(folding.fold_matrix(beta, log_arity, mat.as_view())));
             });
-        });
+        }
     }
 }
 
--- a/fri/src/hiding_pcs.rs
+++ b/fri/src/hiding_pcs.rs
@@ -11,6 +11,7 @@ use p3_matrix::bitrev::{BitReversalPerm, BitReversibleMatrix};
 use p3_matrix::dense::{DenseMatrix, RowMajorMatrix, RowMajorMatrixCow};
 use p3_matrix::horizontally_truncated::HorizontallyTruncated;
 use p3_matrix::row_index_mapped::RowIndexMappedView;
+use p3_maybe_rayon::prelude::*;
 use rand::distr::{Distribution, StandardUniform};
 use rand::{CryptoRng, RngExt, SeedableRng};
 use spin::Mutex;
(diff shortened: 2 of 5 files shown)


tcoratger (approved):
Reviewed carefully — the four optimizations are each mathematically sound, and I confirmed the prover output is byte-identical. Approving; the two follow-up items below are non-blocking.

## What I verified

**`fold_row` arity-2 closed form** (`two_adic_pcs.rs:112`). For interpolation points `{s, -s}`, Lagrange gives

```
f(beta) = y_0 (beta + s)/(2s) - y_1 (beta - s)/(2s)
        = (y_0 + y_1)/2 + (y_0 - y_1) * beta / (2s)
```

which is exactly `(lo + hi).halve() + (lo - hi) * beta * halve_inv_power` with `halve_inv_power = (2s)^{-1}`. `subgroup_start` is a power of a two-adic generator, so never zero and `.inverse()` is safe. The point order also matches what the generic path builds, since `two_adic_generator(1) = -1` makes `shifted_powers(subgroup_start).take(2) = [s, -s]`. The new test pins this against `lagrange_interpolate_at` directly, which is the right oracle.

**`fold_matrix` in-place rewrite.** I traced the index mapping: with row-major width `arity`, old pair `j` read `data[2j], data[2j+1]`, i.e. row `2j/arity` and pair `k = (2j % arity)/2`, so `j = row * pairs_per_row + k`. That is precisely the `par_chunks_exact(pairs_per_row)` chunking the new first fold uses for both `data` and `halve_inv_powers`. The `current_beta` schedule is unchanged (`beta, beta^2, beta^4, …` — the new code just hoists the first step out and starts the loop at `beta.square()`), and the `halve_inv_powers` squaring now runs unconditionally in the loop, which matches the old `if step > 0` guard exactly.

On the buffer ping-pong: lengths alternate `data = N, N/2, N/4, …` against `next_data = N/2, N, N/2, N/4, …`, so `next_data.len() >= height` holds at every iteration and `truncate` never reallocates. Output length is `m.height()`, as the new test asserts.

**Verifier alpha ladder hoist.** Algebraically identical, since

```
sum_i alpha^(off+i) (p_i(z) - p_i(x)) * q
  = (sum_i alpha^(off+i) p_i(z) - sum_i alpha^(off+i) p_i(x)) * q
```

The part I actually wanted to check was whether the precomputed ladder can ever desynchronize from the per-query walk. It cannot: `opening_points()` and the accumulation loop use the same batch → matrix → point nesting, and every length is hard-pinned *before* the loop —

- `verifier.rs:696` — `input_openings.len() == commitments_with_opening_points.len()`
- `verifier.rs:712` — `opened_values.len() == indices.len()`
- `verifier.rs:723` — `opened_values[q].len() == mats.len()`
- `verifier.rs:892` — `mat_opening.len() == ps_at_z.len()`, and this one fires before `reductions.next()`

so every `zip` in the loop is exact and `alpha_powers` is provably long enough (`offset + len <= ladder_lens[log_height] <= max`). Confirmed `mat_opening` is `&Vec<Val>`, so the per-column product really does drop from `Challenge × Challenge` to `Challenge × Val` — a genuine ~4x on the multiply for a degree-4 extension.

**Hiding LDE fusion.** This is the clever one. `coset_idft_batch(mat, s)` is `idft` followed by `coset_shift_cols(_, s.inverse())` (`dft/src/traits.rs:145`), i.e. it scales coefficient `j` by `s^{-j}`. So `coset_idft_batch(evals, shift.inverse())` yields `c_j * shift^j` — exactly the coset-adjusted basis that `coset_lde_batch(evals, k, shift)` feeds to its own forward DFT. Linearity then lets the mask ride along in the same transform, and the mask coefficients must *not* pick up the `shift^j` factor, which they don't. The sign asymmetry is also right: `-=` on the first half is load-bearing because those indices hold real chunk coefficients, while `+=` on the second half is equivalent to `=` (everything at index `>= h*w` is zero after the `resize`) but is the more robust spelling.

Worth a code comment, since the next reader will ask: `coset_lde_batch_with_transform` looks like the natural hook here but genuinely cannot express this fusion — its closure sees the buffer *before* the resize, so it can't reach the mask's `(h+i)*w + j` indices, and the following `coset_dft_batch` would wrongly scale the mask by `shift^j`.

**Bit-identical: confirmed independently.** I regenerated both FRI-backed proof fixtures on the branch and the bytes are unchanged:

```
uni_stark_two_adic_v1.postcard    0c62c6cd86ed6e66a4264bc12a7e060f
batch_stark_two_adic_v1.postcard  0359f6f9137a43566aeed6b5c8c1491f
```

Also `cargo test -p p3-fri --release` (41 + 2 + 26) and `p3-uni-stark` all pass, `clippy --all-targets` clean, and the `commit_phase` parallel roll-in is guarded by `v.len() == folded.len()` so its zip is exact and the elementwise map keeps it deterministic.

## Two follow-ups (non-blocking)

**1. Add `Radix2DitParallel` to `check_fused_quotient_ldes`.** The per-backend testing is exactly the right instinct here, but it currently covers `Radix2Dit`, `Radix2Bowers`, and `Radix2DFTSmallBatch` — all three of which have `type Evaluations = RowMajorMatrix<F>`. `Radix2DitParallel` is the *only* backend with `Evaluations = BitReversedMatrixView<RowMajorMatrix<F>>`, and it separately overrides `dft_batch`, `coset_idft_batch`, and `coset_lde_batch_with_transform` — so it is the one backend where the natural-order / bit-reversed conventions the fusion depends on could plausibly diverge. It is also what `keccak-air`, `poseidon1-air`, and the `stir` benches actually use.

I added it locally and it passes, so this is purely about locking it in:

```rust
#[test]
fn fused_quotient_lde_matches_two_transforms_dit_parallel() {
    check_fused_quotient_ldes(&Radix2DitParallel::<Val>::default());
}
```

**2. Double-check the benchmark baseline.** The same commit range that changes `fold_matrix` also changes the bench: it drops `mat.clone()` from inside the timed closure and reshapes the input from `(n, 2)` to `((2n)/arity, arity)`. Comparing old-bench numbers on main against new-bench numbers on the branch would fold the bench fix into the reported delta. Since `fold_matrix` already took `M: Matrix<EF>` on main, the *fixed* bench runs against main unchanged — if that is what you did, all good, just worth stating.

Nice work, and thanks for the per-backend regression tests on the LDE fusion — that is the right level of paranoia for a change that leans on transform-order conventions. 🙏

PR #2016, opened 2026-09-03
perf(binary-field): Karatsuba fallback, vector-native GF(2^128) mul, Frobenius conjugation, table squaring

Nashtare:
## Changes

- Route the non-hardware-CLMUL fallback for 64-/128-bit multiplication through Karatsuba instead of the fully recursive reference implementation.
- Add a vector-native GF(2^128) multiply-and-reduce kernel on AArch64 (kept alongside the existing composed kernel — measured faster for throughput-bound callers, slower for latency-bound ones, so each caller uses whichever fits).
- Compute Frobenius over adjacent tower levels by an O(1) conjugation instead of a squaring loop.
- Square in the tower basis through a lookup table instead of round-tripping via the polynomial basis; no longer gated on hardware CLMUL.
- Flatten a tower-field `Vec` into its base field in one copy instead of per-element.

Verified correct via proptests against independent references (including a from-scratch bit-serial multiplication reference and Miri for the flatten change) on both the CLMUL and non-CLMUL paths; x86_64 untouched/unverified, aarch64 only for the new SIMD kernel.

## Benchmarks

`arithmetic` bench, this machine (Apple Silicon, aarch64), main vs branch (20 samples):

**With hardware CLMUL (default):**

|---|---|---|---|
| square/64 (throughput) | 4.19 µs | 1.31 µs | −68.8% |
| square/128 (throughput) | 12.59 µs | 3.26 µs | −74.1% |
| mul/inverse/mul_alpha | — | — | ~flat (fallback path not exercised) |

**Without hardware CLMUL (`-C target-feature=-aes`, simulating x86_64-no-pclmulqdq / wasm / generic ARM):**

|---|---|---|---|
| mul/64 (dispatched, latency) | 168.7 µs | 23.2 µs | −86.2% |
| mul/128 (dispatched, latency) | 485.3 µs | 63.8 µs | −86.9% |
| square/64 (throughput) | 8.45 µs | 1.31 µs | −84.5% |
| square/128 (throughput) | 19.04 µs | 3.27 µs | −82.8% |
| inverse/128 | 562.8 µs | 129.6 µs | −77.0% |

### DIFF
--- a/binary-field/src/clmul/aarch64.rs
+++ b/binary-field/src/clmul/aarch64.rs
@@ -1,6 +1,10 @@
 //! The `PMULL` backend, which lives under the `aes` target feature.
 
-use core::arch::aarch64::vmull_p64;
+use core::arch::aarch64::{
+    uint8x16_t, vdupq_n_u8, vdupq_n_u64, veorq_u8, vextq_u8, vgetq_lane_u64, vmull_high_p64,
+    vmull_p64, vreinterpretq_p64_u8, vreinterpretq_u8_u64, vreinterpretq_u64_u8,
+};
+use core::mem::transmute;
 
 /// The carryless product of two 64-bit polynomials over `GF(2)`.
 ///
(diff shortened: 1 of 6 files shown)


tcoratger (approved):
The field arithmetic is right in every path, and the performance claims are not just reproducible but conservative — they hold on an architecture the description explicitly declines to claim.

## The AArch64 reduction, derived by hand

Modulus `x^128 + x^7 + x^2 + x + 1`, so the tail is `T = 0x87`. Writing `high = h0 + h1*x^64`:

```
P        = low + high*x^128  ==  low + high*T
high*T   = h0*T + h1*T*x^64,   and h1*T = e0 + e1*x^64 with deg e1 <= 6
         == h0*T + e0*x^64 + e1*T
P        == low + (h0 + e1)*T + e0*x^64        deg <= 127, no third fold
```

Each instruction maps onto that. Modelling `EXT #8` as `(x >> 64) | (y << 64)` makes `vextq_u8::<8>(b, b)` the half-swap, `ext(0, middle)` the low contribution `middle_lo * x^64`, `ext(middle, 0)` the high contribution `middle >> 64`, `spill = e1` and `carried = e0 * x^64`. The `clmul_low(high ^ spill, tail)` reading only the low 64 bits is correct precisely because the `h1` left in the high half is irrelevant there.

**Caveat worth stating plainly:** I could not execute this kernel — the review machine is x86_64. What I did instead was transliterate it into `u128` shifts and check *that* against the portable composition: 196 corner pairs and 200 000 random pairs agree, 20 000 pairs of the composition agree with a from-scratch bit-serial multiply-and-reduce, and the kernel alone satisfies identity, annihilator, commutativity, distributivity and associativity over 20 000 triples. So the **algebra** is confirmed and only the **instruction semantics** remain target-dependent. Those are covered by the two `macos-latest` legs, which do genuinely execute the gated tests (`aarch64-apple-darwin` carries `aes` in its baseline — corroborated by `ci.yml` needing `-C target-feature=-sha3` to *remove* an Apple baseline feature).

## The rest of the math

- **`transmute` soundness.** Both types are 16 bytes and `vmull_p64` returns `p128 = u128`; these are value transmutes, so alignment is not in play. Endianness is the real hazard, and the `const { assert!(cfg!(target_endian = "little")) }` makes a big-endian AArch64 build a compile error rather than silent corruption — `23ba1c8a` is exactly that. Gating is right: the module needs `aes`, which implies `neon`, and nothing else used needs more.
- **`square_128`.** With `X^2 = alpha*X + 1`, Karatsuba at `a = b` gives `z1 = (a0+a1)^2 + a0^2 + a1^2 = 0` in characteristic 2, so `a^2 = (a0^2 + a1^2) + alpha*a1^2*X`. That is `low ^ high` and `mul_alpha(high)`.
- **The `alpha` convention**, which is where these implementations usually go wrong. For `BinaryField128` over `BinaryField64` the defining polynomial is `X^2 + x_6*X + 1`, and `mul_alpha` on the lower level multiplies by `x_6` — the same `mul_alpha` the pre-existing Karatsuba path uses for the same polynomial. Consistent.
- **Frobenius conjugation.** `X^2 + alpha*X + 1` has root sum `alpha` in characteristic 2, so the nontrivial automorphism sends `X` to `alpha + X` and `a0 + a1*X` to `(a0 + alpha*a1) + a1*X`. For a quadratic extension the Galois group has order 2, so this genuinely is `x -> x^q`, and reducing the repeat count mod 2 is right.
- **The squaring table.** Column `j` is `M(N(e_j)^2)` with `N` the tower-to-poly change of basis and `M` its inverse — precisely the matrix of squaring in the tower basis. It routes through the `const` polynomial multiply rather than the hardware instruction, so it really is available on every target, and indexing a `[u64; 256]` by a `u8` makes out-of-bounds unrepresentable.

## One thing I strengthened

The squaring proptest sampled 256 cases. Squaring is `GF(2)`-linear, so agreement on the basis vectors `1 << i` settles **all** inputs — so I made it exhaustive: `square()` against `x * x` on every basis vector, additivity over all 64x64 and 128x128 basis pairs, and Frobenius on all 128 basis vectors with the involution property. Passes in both the default build and `+pclmulqdq`, where `x * x` takes the independent hardware route. That is in the follow-up.

## Benchmarks reproduce, and x86 is better than claimed

Interleaved A/B, 2 pairs, one worktree. Default build (Karatsuba fallback + table squaring live):

| bench | main | branch | Δ | claimed |
|---|---|---|---|---|
| `mul/64` dispatched latency | 216.8, 217.3 µs | 26.03, 26.05 µs | **−88.0%** | −86.2% |
| `mul/128` dispatched latency | 648.2, 647.3 µs | 77.9, 77.0 µs | **−88.0%** | −86.9% |
| `square/64` throughput | 9.78, 9.77 µs | 1.105, 1.102 µs | **−88.7%** | −84.5% |
| `square/128` latency | 25.87, 25.70 µs | 6.87, 6.87 µs | **−73.4%** | — |
| `mul/8`, `/16`, `/32` | 2.27, 5.4, 9.4 µs | 2.27, 5.3, 9.5 µs | ~0% | untouched |

With `+pclmulqdq`: `square/64` latency **−70.4%**, `square/64` throughput **−75.0%**, `square/128` throughput **−69.2%**, `mul/128` unchanged. Every pair consistent, effects ~15x the noise floor, and the untouched levels confirm no collateral regression.

Constant-time is not a requirement here — these fields back public commitments and DFTs and nothing in the crate handles secrets — so I am not reporting the data-dependent table lookups as a finding.

Hygiene: 106 tests and `clippy --all-targets` clean in **both** the default and `+pclmulqdq` builds; `p3-binary-dft` still passes.

## Follow-ups, going into a separate PR

Nothing blocking. Two of them are description-only, so worth fixing here rather than in code:

- **Bullet 5 is stale.** "Flatten a tower-field `Vec` into its base field in one copy instead of per-element" already describes `main` — the identical single-copy body is there today. The net change is four `const` assertions added, the function moved, and the safety comment made self-contained. Real hardening, just not a perf change.
- **"x86_64 untouched/unverified" undersells it.** x86 gets −70% on squaring with multiplication unchanged, per the table above.

In code, both minor: the vector kernel's algebra has no portable test standing behind it off AArch64 (the module comment itself flags this kernel as the one routine with no portable definition), and `SQUARE_64` adds 16 KiB of unconditional rodata to a `no_std` crate — static tables go 160 to 176 KiB, now emitted on embedded targets too. `Build embedded` passes and the trade buys ~70% on squaring everywhere, so it reads as deliberate; just worth being explicit for embedded consumers.

Merging.

PR #2017, opened 2026-09-03
perf(binary-dft): chain butterfly twiddles instead of rebuilding each one

Nashtare:
## Changes

- Chain each butterfly block's twiddle from the previous one via an XOR recurrence instead of recomputing it from scratch (the recomputation was a data-dependent bit-walk, the dominant per-twiddle cost). Applied to both the tower-basis (`LchNtt`) and polynomial-basis (`PolyBasisNtt`) transforms.
- Added a regression test covering the twiddle chain across multiple parallel butterfly tasks — a class of bug the pre-existing suite could not have caught (round-trip tests are structurally blind to a wrong twiddle schedule).

Verified bit-identical to `main` across ~4,700 (level, size, width, shift) shapes, including edge cases (height 1, height 2, domain at the full field bit width) and both serial and parallel builds.

## Benchmarks

`ntt`/`encode` bench, this machine, main vs branch (20 samples):

|---|---|---|---|
| ntt/32 (log_n 14–20) | 9.5 ms–898 ms | 9.2 ms–881 ms | −1.8% to −3.5% |
| ntt/64 (log_n 14–20) | 7.0 ms–673 ms | 6.6 ms–647 ms | −4.0% to −5.5% |
| ntt/128 (log_n 14–20) | 22.0 ms–2.11 s | 21.7 ms–2.09 s | −0.8% to −1.8% |
| encode/BinaryField128 (log_n 14–20) | 12.1 ms–1.01 s | 11.0 ms–923 ms | −7.9% to −9.5% |

### DIFF
--- a/binary-dft/src/domain.rs
+++ b/binary-dft/src/domain.rs
@@ -1,5 +1,7 @@
 //! The additive NTT domain: `F_2`-linear subspaces spanned by the Cantor basis.
 
+use alloc::vec::Vec;
+
 use p3_binary_field::TowerLevel;
 
 /// The `index`-th point of the additive NTT domain, `Σ_r bit_r(index) · v_r`.
@@ -22,6 +24,26 @@ pub fn domain_point<F: TowerLevel>(index: usize) -> F {
         r += 1;
     }
     point
+}
+
+/// The increments of `domain_point(index << 1)` as `index` runs upwards from zero.
+///
+/// `domain_point` is `F_2`-linear in its index and `(index − 1) ^ index` is the mask of bits
+/// `0..=k` for `k = index.trailing_zeros()`, so consecutive points differ by
+/// `steps[k] = Σ_{r ≤ k} v_{r+1}`. The `count` entries returned cover every `index` below
+/// `2^count`.
+///
+/// # Panics
+/// Panics if `count` is at least the bit width of `F`.
+#[must_use]
+pub(crate) fn domain_point_steps<F: TowerLevel>(count: usize) -> Vec<F> {
+    let mut point = F::ZERO;
+    (0..count)
+        .map(|r| {
+            point += F::cantor_basis(r + 1);
+            point
+        })
+        .collect()
 }
 
 /// The subspace polynomial `W_j` of `S_j`, defined by `W_0(x) = x` and
--- a/binary-dft/src/domain.rs
+++ b/binary-dft/src/domain.rs
@@ -65,7 +65,18 @@ mod tests {
     use p3_binary_field::{BinaryField8, BinaryField16, BinaryField128, TowerLevel};
     use p3_field::PrimeCharacteristicRing;
 
-    use super::{domain_point, subspace_polynomial};
+    use super::{domain_point, domain_point_steps, subspace_polynomial};
+
+    /// Chaining the increments of a stage reproduces the points a full walk reaches.
+    fn check_steps_chain_to_the_walk<F: TowerLevel>(count: usize) {
+        let steps = domain_point_steps::<F>(count);
+        let mut point = F::ZERO;
+        assert_eq!(point, domain_point::<F>(0));
+        for index in 1..1usize << count {
+            point += steps[index.trailing_zeros() as usize];
+            assert_eq!(point, domain_point::<F>(index << 1), "index={index}");
+        }
+    }
 
     /// The image of every single-bit index is the matching Cantor basis vector, which pins the
     /// basis itself and not merely some graded `F_2`-linear reparametrisation of it.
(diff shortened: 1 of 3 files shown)


tcoratger (approved):
The recurrence is right. I derived it independently rather than reading it off the code, and it matches exactly.

## The math

Write `P(index) = domain_point(index << 1)`. Because `domain_point` is `F_2`-linear in its index bits,

```
P(index) = sum_s bit_s(index) * v_{s+1}
```

is itself `F_2`-linear, so in characteristic 2

```
P(index) + P(index - 1) = P(index ^ (index - 1))
```

For `k = trailing_zeros(index)`, subtracting one flips exactly bits `0..=k`, hence `index ^ (index - 1) = 2^(k+1) - 1` and

```
P(index) + P(index - 1) = P(2^(k+1) - 1) = sum_{s=0}^{k} v_{s+1} = sum_{r=1}^{k+1} v_r
```

The table holds `steps[k] = sum_{r=0}^{k} cantor_basis(r+1) = sum_{r=1}^{k+1} v_r`. Exact match. The loop seeds from scratch at the first block and advances with the *global* block index, which is what the identity requires, so `t_i = base + P(first + i)` holds at every block.

Worth stating explicitly since "chained" accumulation usually raises a drift question: there is none here. This is an exact identity in a finite field, not a floating-point sum, so error cannot accumulate across a stage no matter how long the chain.

## Index bound

Tight but sound. Stage `j` has exactly `2^(log_n - 1 - j)` blocks with no remainder, so `blk <= 2^m - 1` for `m = log_n - 1 - j`. Over `blk` in `1..2^m` the largest trailing-zero count is `m - 1`, attained at `blk = 2^(m-1)`, and the table has `m` entries — exactly saturated, never over. At `j = log_n - 1` the table is empty and the single block never indexes it.

The panic surface is also unchanged: the largest basis index needed is `log_n - 1` at stage 0, which `domain_point(blk << 1)` already required at that same stage before this change. So the `ell <= bit width` precondition is not tightened.

## Verified empirically

- **Differential sweep against a from-scratch twiddle walk**, forward and inverse schedules independently: widths `{1,2,3,5,7}` x `log_n` `1..=14` x shifts `{0, 1, 0x1234..}` — 210 shapes, all bit-exact. That covers the multi-task regime (a task first seeds at a nonzero block index only from `log_n >= 12`), partial final tasks at widths where `GRAIN / half` does not divide the block count, and the zero-shift path where the twiddle shortcut fires.
- Field-level caps: `BinaryField8` at `log_n` `1..=8`, `BinaryField16` at `log_n` 15 and 16. Height 1 (no stage runs) and height 2. All bit-exact, round-trips hold.
- **Bit-identical to `main`, confirmed independently.** A digest over 5 widths x `log_n` `1..=13` x 3 shifts x 4 transform entry points — 780 invocations — gives `0xf25dbf2c01e28685` on both `b84529c0` and `632d3450`.
- **The benchmark reproduces.** Interleaved A/B, 3 pairs in one worktree, `RUSTFLAGS="-C target-cpu=native"`, `encode/BinaryField128/14`: base mean 16.263 ms, branch mean 14.720 ms = **-9.5%**, branch faster in all three pairs with non-overlapping intervals. Matches your claimed range. (Worth noting because I could not reproduce the numbers on #2014 — that one turned out to be aarch64-only. This one holds on x86 AVX-512.)
- `cargo test -p p3-binary-dft --release` 32 + 2 pass; `clippy --all-targets` clean.

## On the new test

Credit where due: the added test is well-aimed rather than decorative. The round-trip test genuinely *cannot* catch a wrong twiddle schedule — both directions read the same twiddles, so they invert each other regardless — and the naive oracle tops out below the height where a task first seeds at a nonzero block index. Comparing against an independent walk at `log_n = 12` is the minimal thing that closes that gap, and it is exactly the gap that mattered.

## Follow-ups, going into a separate PR

Nothing blocking. Three cosmetic/defensive items:

- The increment table is stage-*independent* — its body is a running sum over the basis with no stage parameter — but it is rebuilt per stage as `log_n - 1 - j` entries, which actively suggests otherwise. Perf cost is nil (`O(log_n^2)` against `O(n log n)` butterflies); the reason to hoist it is that "the table depends on the stage" is the one thing a reader must not believe when auditing an incremental chain.
- The header doc's amended sentence says the increments are "of a stage", and reads as a concession against the pre-PR "no per-size precomputation" claim rather than a description.
- The chain's dependence on **ascending block order** is unstated. `chunks_mut` guarantees it, but the neighbouring comment only covers the inner parallel split, so turning that outer loop parallel later would break the chain silently — precisely the failure mode your new test exists to catch. That one I would not skip.

Merging.

PR #2019, opened 2026-09-03
fix(security,fri): accounting and guard hygiene fixes

Nashtare:
## Summary
- `compute_upper_m` now floors instead of ceils its admissible-`m` bound and takes the caller's `max_combo` instead of hardcoding `2`, both of which previously overstated reported security.
- `LOG2_E` now rounds up instead of down, keeping `bits_per_query`'s correction conservative.
- Documented why `SecurityAssumption::JohnsonBound`'s list size and `list_size_ldr_m` intentionally differ rather than being duplicate/disagreeing formulas.
- `p3-fri` now rejects `log_blowup == 0` in both `prove_fri` (assert) and `verify_fri` (new `FriError::ZeroBlowup`), a config that previously let arbitrary claimed evaluations verify.
- Documented `HidingFriPcs`'s trace-height-vs-query-budget requirement for its zero-knowledge guarantee to hold.

## Test plan
- [x] `cargo test -p p3-security -p p3-fri`
- [x] `cargo test -p p3-circle -p p3-uni-stark -p p3-batch-stark`
- [x] `cargo clippy -p p3-security -p p3-fri --all-targets -- -D warnings`
- [x] `cargo check --workspace --all-targets`
- [x] `cargo fmt --check`

### DIFF
--- a/fri/src/hiding_pcs.rs
+++ b/fri/src/hiding_pcs.rs
@@ -26,6 +26,25 @@ use crate::{BatchMultiOpening, FriParameters, FriProof, TwoAdicFriPcs};
 /// bounded by [`CryptoRng`]. That rules out generators known to be unsuitable for cryptographic
 /// use, but it does not replace proper seeding: a caller who seeds from a predictable source lets
 /// an observer reproduce the stream and strip the masks.
+///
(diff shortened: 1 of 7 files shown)


Nashtare:
I will wrap those up in a more complete PR soon.

PR #2024, opened 2026-09-04
feat(sumcheck)!: drive the generic-degree transcript through the typed Fiat-Shamir layer

tcoratger:
## What this changes

The generic-degree sumcheck wrote its Fiat-Shamir transcript by hand, on both sides. It now derives every challenge through the typed transcript layer from `p3-challenger`.

## The problem

A sumcheck round is three transcript operations in a fixed order: absorb the round polynomial, grind, sample the challenge. The prover wrote that sequence in one place and the verifier wrote it again in another. Nothing checked that the two agreed.

The claimed sum was worse. The verifier absorbed it itself; the prover's documentation said *the caller* had to:

> The caller must observe `claimed_sum` into the challenger before calling, so that the round challenges depend on it.

So the two sides agreed only if every caller remembered a line the compiler never asked for. `multi-stark`'s zerocheck prover carried exactly that line — `challenger.observe_algebra_element(EF::ZERO)` — to satisfy it.

## The idea

Describe the transcript once, as a value, and let both sides walk that description instead of re-deriving it.

```
claimed sum              1 extension element
round 0:  polynomial     degree extension elements
          grinding       present only when the difficulty is positive
          challenge      1 extension element
round 1:  ...
```

A new `generic_degree::transcript` module holds that description plus one driver per side. Both make the same three calls — `new`, `round`, `finish` — so the prover cannot drift from the verifier, and the step labels stay private to the module. The claimed sum is bound by the drivers, so there is nothing left for a caller to mirror.

The description also lands in the seed as a 32-byte fingerprint, which means the round count, the per-variable degree and the grinding difficulty are all bound. Two runs differing in any of them can no longer share challenges.

## It found a live inconsistency

Wiring the zerocheck to the shared description, 19 tests failed with:

```
Received  Atomic Message round_poly Fixed(3)
but expected Atomic Message round_poly Fixed(2)
```

The zerocheck prover's internal `max_degree` and the width of a *transmitted* round polynomial differ by one: the zerocheck's equality weight adds a degree, so `standard_round_from_q_evals` emits `max_degree + 1` evaluations. The prover reached that number one way and the verifier another (`sumcheck_degree = air_degree + 1`), and nothing bound it — the widths agreed by construction, not by check.

It is now named, with the relation stated at the point of use, and any future drift fails the shape check instead of passing silently.

## Shape checks stay, and moved earlier

Worth being explicit, because it is the one place where the typed layer does *not* replace a hand-written check.

A round polynomial of the wrong width is untrusted input, so it must be rejected with a typed error. The transcript would panic on it instead — which is the right response to two sides built from different descriptions, and the wrong response to a malformed proof. So `verify` now runs in two phases: check every length the proof carries, then replay the transcript. The expected values it compares against are the same numbers the description is built from, so there is still one source for them.

## Worked example

Two verifications of the same one-round proof, differing only in the claimed sum:

```rust
let proof = GenericDegreeProof { claimed_sum: EF::ONE,  round_polys: vec![vec![EF::ONE]], .. };
let other = GenericDegreeProof { claimed_sum: EF::ZERO, round_polys: vec![vec![EF::ONE]], .. };

let (point_a, _) = proof.verify(&mut fresh_challenger(), 1, 1, 0)?;
let (point_b, _) = other.verify(&mut fresh_challenger(), 1, 1, 0)?;

assert_ne!(point_a, point_b);   // the sum is bound before the challenge
```

Before this change that assertion held only if the prover's caller had remembered the mirroring line. Now it holds by construction, and it is a test.

**`GrindingChallenger` drops `Sync + Clone` from its supertraits** and gains a blanket impl for `&mut C`.

Those bounds belong to the implementations that search for a witness in parallel — which already state them in their own `where` clauses — not to the trait. Checking a witness needs neither. Removing them compiles the whole workspace unchanged, and it is what lets a sub-protocol grind on a transcript it only borrows: without it a borrowed transcript could absorb and sample but not grind.

**An `observe_*` family on both drivers**, for values the caller carries in its own proof type. The convention is now:

- `add_*` writes the value into the driver's proof bytes
- `next_*` reads it back out
- `observe_*` does neither, and only binds the sponge

A protocol whose proof is its own serialisable type would otherwise ship every value twice. Asymmetry between the two sides cannot go unnoticed: a prover that writes where its verifier only observes leaves unread bytes, which finalisation rejects, and the reverse runs the verifier past the end.

## A panic inside a live transcript scope becomes an abort

Both drivers panic on drop if the transcript was never finalised, which is what stops a verifier from silently skipping its trailing-bytes and full-replay checks. But it means any *other* panic raised while a driver is alive unwinds into that one, and a second panic during unwinding calls `abort()`. A verifier service cannot `catch_unwind` an abort into a rejection.

The state machine's own failures are already safe: the pattern player marks itself aborted before it panics, and a failing read aborts before returning its error. The exposure was driver code raising a panic of its own inside the scope. Two such sites sat on paths fed by a proof, and both are now typed errors:

- Binding a list of extension messages compares the count against the recorded step before absorbing anything. That count travels inside the caller's own proof type, so it was the one attacker-controlled length still handed straight to the player. The prover-side twin keeps panicking, its input being trusted.
- A round whose grinding step carries no witness reports a missing witness. The verifier driver gained a way to release its completeness check, which is what lets that rejection travel alone.

The prover's own width assertion is deleted rather than converted. Binding already compares the count against the recorded step, and the player aborts before panicking, so the same bug still fires — as a single panic, with a diff message. That is the general rule the fix rests on: driver code should not duplicate a check the transcript already makes, because only the transcript knows to abort first.

Verified the release is load-bearing by removing it: the rejection turns back into the drop-time panic at `verifier.rs:64`.

## Breaking changes

- **The sumcheck transcript changes**, so proofs produced before this do not verify after it. The checked-in WHIR compatibility fixture is regenerated.
- **`VerifierState::observe_extensions` returns a `Result`.** A count outside the recorded shape is a rejection, not a panic.
- **`GenericDegreeError` gains `MissingPowWitness`.**
- **`MultiStarkConfig::Val` tightens from `Field` to `PrimeField64`.** The transcript packs its seed bytes into field elements, which needs a modulus that fits in 64 bits. This removes no reachable generality: both `MultilinearPcs` implementations require `F: TwoAdicField`, and a binary tower field GF(2^k) has multiplicative order 2^k − 1, so it can never satisfy that bound. Every field a grinding challenger works over is already `PrimeField64`.

## Test plan

- [x] `cargo test -p p3-sumcheck` — 293 pass
- [x] `cargo test -p p3-multi-stark` — 76 pass across all test targets, including the regenerated fixture
- [x] `cargo test -p p3-challenger` — 176 pass
- [x] `cargo clippy -p p3-challenger -p p3-sumcheck -p p3-multi-stark --all-targets` — clean
- New tests: the claimed sum and each round polynomial are bound before the challenge that follows them; every shape number moves the fingerprint; a wrong evaluation count is a typed error rather than a panic.
- Two of them drive the verifier transcript the way a downstream crate would, straight from proof fields with no shape pre-check in front, and pin both malformed inputs to errors.

## Review notes

Two judgement calls worth a second opinion:

1. The regenerated fixture keeps its `v1` filename. The content change is the signal in git, but the fixture exists to catch *un*intended transcript changes, so a filename bump is defensible.
2. `VerifierTranscript::finish` uses `expect` on a verifier path. It is unreachable — the proof carries every value, so the driver reads an empty wire and no trailing bytes can remain — but it is still an `expect` in verification code, and it could be folded into `GenericDegreeError` instead.


multi-stark/src/config.rs
```diff
@@ -13,7 +13,11 @@ use p3_sumcheck::layout::Table;
 /// The prover and verifier are generic over any implementation.
 pub trait MultiStarkConfig {
     /// Base field the trace and committed columns live in.
-    type Val: Field;
+    ///
+    /// The transcript packs its seed bytes into base-field elements.
+    /// The modulus therefore has to fit in 64 bits.
+    /// Every field a grinding challenger works over already does.
+    type Val: PrimeField64;
```
tcoratger, line 20: We should probably change those stuffs later but for now this seems the best route to take in order to avoid more disruptive changes

tcoratger:
@Nashtare Just the first simple step for the integration, I've followups in mind to do the same for FRI or WHIR, etc

sumcheck/src/generic_degree/transcript.rs
```diff
@@ -0,0 +1,344 @@
...
+        let round = self.round;
+        self.round += 1;
+
+        // Bind the polynomial before the challenge that will be evaluated on it.
+        self.state
+            .observe_extensions::<F, EF, FieldToFieldCodec<F>>(ROUND_POLY, evals);
+
+        // Re-run the prover's grinding step on the witness it committed to.
+        if self.pow_bits > 0 {
+            let witness = witness.expect("a positive difficulty requires a witness per round");
```
Nashtare: **MEDIUM — a missing PoW witness panics, and the panic escalates to `abort()`.**

`VerifierTranscript` is `pub use`d from `generic_degree` and is the documented way to drive this protocol from another crate. A downstream verifier that walks `GenericDegreeProof`'s fields itself, and forgets the `pow_witnesses.len() == num_rounds` pre-check, gets a panic on attacker-shaped input.

The escalation is the part worth attention. This `expect` fires before any `player.abort()`, so unwinding drops `VerifierState`, whose `Drop` (`challenger/src/fs/state/verifier.rs:53-67`) panics again with `"Dropped unfinalized VerifierState"`. Panic-during-unwind is `abort()` — a verifier service cannot `catch_unwind` this into a rejection.

Note the `Drop` impl's own comment claims *"Failing reads and pattern panics both mark the player aborted first, so this never fires while another failure unwinds."* That invariant holds for the state machine's own error paths (`observe_pow`'s failure branch does call `player.abort()` before returning `Err`, which is why the negative tests unwind cleanly) but not for a panic raised by driver code that merely happens to be inside the transcript scope. The `# Panics` block at lines 306-309 documents only the evaluation-count case, not this one.

Suggested fix: return `Err(GenericDegreeError::InvalidPowWitness { round })` (or a new `MissingPowWitness`) after aborting the player, rather than `expect`. Alternatively make the signature `witness: F` so the caller's shape check is structurally required.

sumcheck/src/generic_degree/prover.rs
```diff
@@ -80,29 +79,43 @@ pub trait RoundProver<EF> {
...
+
+        for _ in 0..num_rounds {
             let evals = self.round_poly();
-            debug_assert_eq!(
+
+            // The described step carries exactly `degree` evaluations.
+            //
+            // A different count would diverge from the agreed shape.
+            // That is a bug in this state, not untrusted input, so it panics.
+            assert_eq!(
```
Nashtare: **MEDIUM — this `assert_eq!` now aborts the process instead of panicking.**

This was a `debug_assert_eq!` before the PR, with no transcript object alive. It now sits between `ProverTranscript::new` (line 83) and `transcript.finish()` (line 117), so when it fires, unwinding drops `ProverTranscript` → `ProverState::drop` (`challenger/src/fs/state/prover.rs:45-60`) panics again → `abort()`.

Same mechanism as the `transcript.rs:320` comment, and it applies to every panic site now enclosed in a live transcript scope — notably `multi-stark/src/zerocheck.rs:381` (`tau_round.inverse()`) and the `interpolators[round_poly.len()]` indexing, both between `zerocheck.rs:349` and `zerocheck.rs:478`. Net effect: a prover-side invariant violation kills the process with no usable panic message.

Suggested fix: give `ProverTranscript`/`VerifierTranscript` a `Drop` that calls the underlying player's `abort()`, so the drop-time assertion is suppressed while another panic unwinds. That fixes the whole class in one place rather than per call site.

challenger/src/fs/state/verifier.rs
```diff
@@ -170,6 +170,110 @@ impl<'a, C, U: Unit> VerifierState<'a, C, U> {
...
+        F: PrimeField64,
+        EF: Field + BasedVectorSpace<F>,
+        Cdc: Codec<C, F>,
+    {
+        // Validate: the next pattern step is a fixed-length list of extension messages.
+        self.player.interact(Interaction::algebra::<F, EF>(
+            Hierarchy::Atomic,
+            Kind::Message,
+            label,
+            Length::Fixed(values.len()),
```
Nashtare, line 259: **MEDIUM — `observe_extensions` derives the pattern length from attacker-controlled data.**

`values.len()` comes straight from a deserialized proof and is handed to `PatternPlayer::interact`, which `panic!`s on mismatch (`challenger/src/fs/pattern/player.rs:113-118`). The method's own doc concedes this and pushes the obligation onto every caller.

`GenericDegreeProof::verify` does discharge it correctly — `proof.rs:125-135` validates every length before any absorb — so the in-repo path is safe today. The concern is that this is new public API on the verifier driver whose default failure mode on malformed input is a panic (and, per the two comments above, an `abort()` once a transcript is live). Every future consumer has to get the ordering right or it becomes a verifier DoS.

Suggested fix: read the expected length from the recorded step, compare against `values.len()`, and return `Err(TranscriptError::BadProofShape { .. })` via the existing `poison(..)` helper on mismatch. The prover-side twin (`prover.rs:361-390`) can keep panicking — its input is trusted.

sumcheck/src/generic_degree/transcript.rs
tcoratger: Fixed. The round now reports `GenericDegreeError::MissingPowWitness { round }` instead of unwrapping, and it releases the driver's completeness check first so the rejection travels alone.

That release turned out to need new API: `VerifierState` had no public way to mark its player aborted, so an error raised by driver code in another crate could not avoid the drop-time panic. It has one now, documented as the thing a driver calls before returning an error.

I kept `Option<F>` rather than taking `witness: F`, because a zero-difficulty run genuinely has no witness to pass and a dummy would be worse. The error carries the round, which is the part a caller needs.

I checked the escalation is really what the test pins, by commenting out the release:

```
with it:      Err(MissingPowWitness { round: 0 })
without it:   panicked at challenger/src/fs/state/verifier.rs:64  <- the drop check
```

`a_round_missing_its_grinding_witness_is_rejected` drives `VerifierTranscript` straight from proof fields with no shape pre-check in front, which is the downstream shape you described.

sumcheck/src/generic_degree/prover.rs
tcoratger: You are right about the mechanism, and the assertion is gone — but I did not take the `Drop` route, and I want to be explicit about why rather than quietly do something else.

A wrapper `Drop` that calls `player.abort()` cannot tell "a panic is unwinding" from "someone forgot `finish()`". `p3-challenger` is `no_std`, so `std::thread::panicking()` is not available and there is no other way to distinguish the two. So the suppression is unconditional, and the check it suppresses is the one that stops a verifier from accepting a proof with arbitrary appended data (the `VerifierState` doc spells this out at lines 30-38). It would also force `ManuallyDrop` in `finish()`, since a type with a `Drop` impl cannot have its state moved out.

Trading a soundness check for a better panic message is the wrong direction, so I attacked the panic sites instead:

- `transcript.rs:320` and the verifier's `observe_extensions` now return errors. Those were the two on a path fed by a proof, and they are the ones that mattered.
- This assertion is simply deleted. Binding already compares the count against the recorded step, and the pattern player marks itself aborted *before* it panics — which is exactly the invariant your comment quotes, working as intended. So the same bug still fires, as a single panic, with a diff message better than the assertion's.

That last point is the general rule, and I think it is the real answer to "fix the class in one place": the state machine's own panics are safe because only it knows to abort first. Driver code should not duplicate a check the transcript already makes.

On `zerocheck.rs:381` — `tau_round.inverse()` needs a sampled Fiat-Shamir challenge to be exactly zero in a degree-4 extension of BabyBear, and `interpolators[round_poly.len()]` is indexed by the prover's own stage degree, bounded by the batch max at construction. Both are prover-side and not proof-reachable, so I left them; happy to be argued out of that.

challenger/src/fs/state/verifier.rs
tcoratger: Fixed as suggested. The recorded step is read before anything is absorbed, and a mismatch returns `BadProofShape` through `poison(..)`:

```rust
if let Some(Length::Fixed(expected)) = self.player.next_interaction().map(Interaction::length)
    && expected != values.len()
{
    return self.poison(Err(TranscriptError::BadProofShape {
        reason: "list length differs from the recorded shape",
    }));
}
```

`PatternPlayer` gained `next_interaction()` for this — a peek at the step awaiting playback, which is what any driver holding an untrusted count needs.

One deliberate gap: only a recorded `Fixed(_)` step is compared. A recorded step of some other shape falls through to the player and panics, because reaching it means the driver called the wrong method, which is a bug rather than malformed input. The prover-side twin keeps panicking outright for the same reason.

The signature is now `Result<Vec<TranscriptBound<EF>>, TranscriptError>`, and in the sumcheck driver the error maps to `PolyEvalCountMismatch { round, expected, actual }`. So a caller that drives the transcript directly gets the same typed rejection `verify`'s pre-check gives — which was the point, since that pre-check is exactly what a downstream consumer might not write. `a_round_polynomial_of_the_wrong_width_is_rejected` pins it.

Nashtare (approved):

PR #2030, opened 2026-09-04
feat(binary-field): Ghash128, GF(2^128) in the polynomial basis, with SIMD packings

tcoratger:
## What a "basis" is here, and why it costs

`GF(2^128)` is a field with `2^128` elements. Every element is 128 bits, and addition is just
`XOR`. Multiplication is where the choices live.

To multiply, you first have to say *what the 128 bits mean*. A basis is that answer: it fixes
128 fixed elements `e_0 … e_127`, and bit `i` of the integer says whether `e_i` is part of the
element. Different bases describe the same field, but they make different operations cheap.

`BinaryField128` uses the **tower basis**, where the `e_i` are products of nested generators.
This is what makes `GF(2^8)`, `GF(2^16)`, `GF(2^32)` and `GF(2^64)` sit inside it as clean byte
slices — which ring switching and small-field witnesses rely on.

Hardware does not multiply in that basis. `PCLMULQDQ` on x86 and `PMULL` on AArch64 multiply in
the **polynomial basis**, where `e_i` is just `x^i` and elements are polynomials reduced modulo
`x^128 + x^7 + x^2 + x + 1`.

So today, one tower multiply looks like this:

```text
    a (tower)  --16 table lookups-->  a (polynomial)  --.
                                                        |--> 1 carryless multiply --> product
    b (tower)  --16 table lookups-->  b (polynomial)  --'          |
                                                                   |
                             product (tower)  <--16 table lookups--'
```

Forty-eight dependent table lookups wrapped around a single instruction. The conversions, not
the multiply, are the cost.

## The change

`Ghash128` is that same field, held in the polynomial basis from the start. Nothing converts,
because nothing needs to.

It is a full `Field`, so it works anywhere a Plonky3 field works. It also implements
`TowerLevel`, so the additive NTT in `p3-binary-dft` runs over it with no changes. `From`
conversions relate it to `BinaryField128` in both directions, at about 3.5 ns, so code can hold
data in whichever representation suits the step it is in.

### A worked example

Take `x`, the element whose only set bit is bit 1 — the polynomial `x` itself. Squaring it
gives `x^2`, so bit 2. Keep squaring and you climb: `x^4`, `x^8`, … up to `x^64`. Square once
more and you reach `x^128`, which is too big for 128 bits. That is exactly what the modulus is
for: it says `x^128 = x^7 + x^2 + x + 1`, so the answer wraps around to `0x87`.

Every routine in this PR is a way of doing that wrap efficiently. The multiply folds it with a
second carryless product instead of a shift; the square root runs it backwards; inversion
chains it.

## What is in it

- **`Ghash128`** — the scalar field, with derived rather than transcribed constants: its
  generator and its Cantor basis are computed at compile time as the images of the tower's, so
  both representations provably span the same NTT domain.
- **`PackedGhash128`** — two elements per register on `avx2` and four on `avx512f`, in both
  cases over `VPCLMULQDQ`, wired as `Ghash128::Packing`. One element is exactly one 128-bit
  lane, so the scalar kernel becomes the packed one unchanged.
- **Three backends** for product, square, square root, inverse and dot product: `pclmulqdq`,
  AArch64 `aes`, and software.
- **Square roots in one product.** A square interleaves the input's coefficients with zeros, so
  a square root un-interleaves them. Separating even from odd coefficients recovers the root
  with a single multiply instead of 127 squarings.
- **Inversion by addition chain.** The inverse is `x^(2^128 - 2)`. Squaring `2^k` times is a
  linear map, so the five maps the chain needs are tabulated at compile time.
- **Dot products that reduce once.** Reduction is linear, so a sum of products can accumulate
  unreduced and fold the modulus a single time at the end.
- **Packed butterflies** in `p3-binary-dft`: the NTT inner loop now runs over `F::Packing` with
  a scalar tail, which is what carries the packed multiply into the transform.
- The **software carryless multiply** replaces a 64-iteration bit-serial loop with the masked
  integer-multiply construction. It is compiled on every target, not only where it is used, so
  its carry argument is tested everywhere.

## Numbers

AMD Ryzen 9 9950X3D, `-C target-cpu=native`. Per element unless stated.

Against `BinaryField128`, which is the same field in the other basis:

| operation | tower | `Ghash128` | |
| --- | --- | --- | --- |
| multiply, dependent chain | 19.8 ns | 3.55 ns | 5.6x |
| multiply, independent | 15.3 ns | 2.17 ns | 7.1x |
| multiply, packed 4 wide | 15.3 ns | 0.47 ns | 33x |
| square | 14.1 ns | 4.49 ns | 3.1x |

The other two are new operations, so they are measured against the obvious way to write them:

- **inverse**, 88.8 ns, against 125 ns for converting to the tower and inverting there.
- **square root**, 5.73 ns, against 510 ns for raising to the power `2^127`.

Dot products, per call, deferred reduction against separately reduced products. The packed
column holds four elements per entry, so its 64-entry row is 256 products:

| entries | scalar | packed |
| --- | --- | --- |
| 4 | 10.7 → 6.7 ns | 10.2 → 7.1 ns |
| 16 | 47.0 → 25.8 ns | 38.1 → 27.5 ns |
| 64 | 192 → 90.0 ns | 143 → 72.4 ns |

Additive NTT, width 16, `parallel`, in milliseconds:

| log n | tower | `Ghash128` | |
| --- | --- | --- | --- |
| 14 | 2.31 | 0.87 | 2.6x |
| 16 | 7.77 | 2.63 | 3.0x |
| 18 | 32.8 | 12.2 | 2.7x |
| 20 | 186 | 181 | memory bound |

## Correctness

Every backend is checked against bit-serial reduction taken straight from the modulus, which
depends on no instruction and no table. On top of that:

- The full `test_binary_field!` and `test_packed_binary_field!` suites.
- A **NIST GCM known-answer vector**, which pins the field against a published specification
  rather than against our own code.
- The change of basis is checked to be a *ring* homomorphism, not merely invertible —
  multiplicativity is the part a change of coordinates does not give you for free.
- The NTT over `Ghash128` is checked to equal the NTT over `BinaryField128` under that change
  of basis. Since the transform is built from twiddle multiplies, this pins the whole
  polynomial-basis arithmetic one butterfly at a time.
- The Cantor basis is checked against its defining recurrence and for linear independence.
- Odd matrix widths and heights crossing the task boundary, so the packed prefix and the scalar
  tail are both exercised.

Verified on AVX-512, AVX2-only and a build with no carryless multiply at all; cross-compiled
for AArch64 and wasm32. `clippy --workspace --all-targets`, `fmt` and `rustdoc` are clean, and
the workspace suite passes.

## What this does not do

`Ghash128` has no byte-aligned subfields. `GF(2^8)` and friends live inside the tower
representation and not inside this one, so anything that needs the subfield structure should
stay on `BinaryField128`. No existing public API changes.

Two things worth a reviewer's attention:

- Inversion, the tower arithmetic and the conversions between bases all use operand-indexed
  tables, so they are not constant-time. The multiply, square, square root and dot product are.
  This is stated in the crate README.
- Inversion's Frobenius maps are 320 KiB of read-only tables. They are compiled only where a
  carryless multiply exists; everywhere else the tower norm is faster anyway, so the module is
  not built at all. Checked on `thumbv7em-none-eabi`, which CI compiles this crate for.

### DIFF
--- a/binary-dft/src/lch.rs
+++ b/binary-dft/src/lch.rs
@@ -12,24 +12,36 @@ use p3_util::log2_strict_usize;
 use crate::domain::{domain_point, domain_point_steps, subspace_polynomial};
 use crate::traits::AdditiveNtt;
 
-/// The Lin–Chung–Han additive NTT over a Cantor-basis domain.
+/// The Lin–Chung–Han additive NTT over the Cantor-basis domain.
 ///
-/// At stage `j`, block `blk` has twiddle `W_j(shift) + domain_point(blk << 1)`.
-/// There is no twiddle table.
+/// Twiddles are index shifts (D8): at stage `j` and butterfly block `blk` the twiddle is
+/// `W_j(shift) + domain_point(blk << 1)`, so there is no twiddle table.
 ///
-/// Consecutive blocks differ by a fixed increment, shared by every stage.
+/// Consecutive blocks differ by a fixed increment, so a block takes its twiddle from the one
+/// before it rather than walking its own index.
+/// The `ℓ - 1` increments are shared by every stage and are the only per-size precomputation.
 ///
-/// Butterflies run on SIMD packings, and drop the multiply where the twiddle is zero.
+/// `W_j` is `F_2`-linear and `domain_point(0)` is zero, so over the subspace itself — the
+/// coset with `shift = 0` — the first block of every stage has a zero twiddle and its
+/// butterfly collapses to a single addition. The inner loop takes that as a separate case,
+/// which removes one multiply in `2/ℓ` of them.
 #[derive(Clone, Debug, Default)]
 pub struct LchNtt<F> {
     _marker: PhantomData<F>,
 }
 
-/// Elements per side of a butterfly task.
+/// The number of field elements one butterfly task covers on each side of a block.
 ///
-/// Narrow blocks share a task.
-/// Wide blocks split into pieces of this size, which have too few blocks to fill a machine.
-/// Either way a stage leaves enough independent tasks to be worth handing to other threads.
+/// Stage `j` has `2^(ℓ − 1 − j)` blocks of `half = 2^j · width` elements per side, so cutting
+/// each side into pieces of this size leaves `n · width / (2 · BUTTERFLY_GRAIN)` pieces at
+/// every stage, independent of `j`: the wide stages, which have too few blocks to fill a
+/// machine, are split from within instead. Stages with `half ≤ BUTTERFLY_GRAIN` keep a single
+/// piece per side and instead gather `BUTTERFLY_GRAIN / half` whole blocks into one task, so a
+/// task is a piece of this size on either side of the crossover.
+///
+/// At a few nanoseconds per butterfly a piece of this size is microseconds of work, well above
+/// the cost of handing a task to another thread, while still leaving hundreds of pieces per
+/// stage at the smallest useful heights.
 pub(crate) const BUTTERFLY_GRAIN: usize = 1 << 10;
 
 impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
@@ -314,8 +326,8 @@ mod tests {
             prop_assert_eq!(ntt.shifted_intt_batch(evals, shift), coeffs);
         }
 
-        /// The low-degree extension agrees with the oracle, on a coset too.
-        /// The input rows reappear as the prefix of the output.
+        /// `LchNtt`'s low-degree extension agrees with the oracle's, on a coset too, and the
+        /// input rows reappear as the prefix: the correspondence Phase 3 folds along.
         #[test]
         fn lch_lde_matches_naive(
             log_n in 0usize..=6,
@@ -339,8 +351,9 @@ mod tests {
     /// A height whose stages take more than one butterfly task, so a task seeds its twiddle at
     /// a block index of its own rather than at zero.
     ///
-    /// The round-trip test is blind to the schedule.
-    /// Both directions read the same twiddles, so they invert each other regardless.
+    /// The oracle tests all sit below that height, and `lch_round_trips` is blind to the
+    /// schedule: both directions read the same twiddles, so they invert each other whatever
+    /// those twiddles are. Only a comparison against an independent walk pins them.
     #[test]
     fn lch_matches_a_twiddle_walk_across_several_tasks() {
         const LOG_N: usize = 12;
@@ -367,7 +380,8 @@ mod tests {
         }
     }
 
-    /// An index past the field's bit width needs a Cantor vector it does not have.
+    /// An index of `S_ℓ` past the bit width of `F` calls for a Cantor basis vector this level
+    /// does not have.
     #[test]
     #[should_panic]
     fn shifted_ntt_batch_rejects_l_past_the_bit_width() {
@@ -376,7 +390,7 @@ mod tests {
         let _ = LchNtt::<BinaryField8>::default().ntt_batch(coeffs);
     }
 
-    /// A height that is not a power of two has no well-defined `l`.
+    /// A height that is not a power of two has no well-defined `ℓ`.
     #[test]
     #[should_panic]
     fn shifted_ntt_batch_rejects_a_non_power_of_two_height() {
@@ -407,7 +421,8 @@ mod tests {
         }
     }
 
-    /// The domain does not depend on the level, so neither does the transform.
+    /// The same data transformed at two levels agrees after embedding: the domain does not
+    /// depend on the level, so neither does the transform.
     #[test]
     fn levels_agree_after_embedding() {
         const LOG_N: usize = 6;
--- a/binary-dft/src/lch.rs
+++ b/binary-dft/src/lch.rs
@@ -3,6 +3,7 @@
 use core::marker::PhantomData;
 
 use p3_binary_field::TowerLevel;
+use p3_field::{PackedValue, PrimeCharacteristicRing};
 use p3_matrix::Matrix;
 use p3_matrix::dense::RowMajorMatrix;
 use p3_maybe_rayon::prelude::*;
@@ -11,36 +12,24 @@ use p3_util::log2_strict_usize;
 use crate::domain::{domain_point, domain_point_steps, subspace_polynomial};
 use crate::traits::AdditiveNtt;
 
-/// The Lin–Chung–Han additive NTT over the Cantor-basis domain.
+/// The Lin–Chung–Han additive NTT over a Cantor-basis domain.
 ///
-/// Twiddles are index shifts (D8): at stage `j` and butterfly block `blk` the twiddle is
-/// `W_j(shift) + domain_point(blk << 1)`, so there is no twiddle table.
+/// At stage `j`, block `blk` has twiddle `W_j(shift) + domain_point(blk << 1)`.
+/// There is no twiddle table.
 ///
-/// Consecutive blocks differ by a fixed increment, so a block takes its twiddle from the one
-/// before it rather than walking its own index.
-/// The `ℓ - 1` increments are shared by every stage and are the only per-size precomputation.
+/// Consecutive blocks differ by a fixed increment, shared by every stage.
 ///
-/// `W_j` is `F_2`-linear and `domain_point(0)` is zero, so over the subspace itself — the
-/// coset with `shift = 0` — the first block of every stage has a zero twiddle and its
-/// butterfly collapses to a single addition. The inner loop takes that as a separate case,
-/// which removes one multiply in `2/ℓ` of them.
+/// Butterflies run on SIMD packings, and drop the multiply where the twiddle is zero.
 #[derive(Clone, Debug, Default)]
 pub struct LchNtt<F> {
     _marker: PhantomData<F>,
 }
 
-/// The number of field elements one butterfly task covers on each side of a block.
+/// Elements per side of a butterfly task.
 ///
-/// Stage `j` has `2^(ℓ − 1 − j)` blocks of `half = 2^j · width` elements per side, so cutting
-/// each side into pieces of this size leaves `n · width / (2 · BUTTERFLY_GRAIN)` pieces at
-/// every stage, independent of `j`: the wide stages, which have too few blocks to fill a
-/// machine, are split from within instead. Stages with `half ≤ BUTTERFLY_GRAIN` keep a single
-/// piece per side and instead gather `BUTTERFLY_GRAIN / half` whole blocks into one task, so a
-/// task is a piece of this size on either side of the crossover.
-///
-/// At a few nanoseconds per butterfly a piece of this size is microseconds of work, well above
-/// the cost of handing a task to another thread, while still leaving hundreds of pieces per
-/// stage at the smallest useful heights.
+/// Narrow blocks share a task.
+/// Wide blocks split into pieces of this size, which have too few blocks to fill a machine.
+/// Either way a stage leaves enough independent tasks to be worth handing to other threads.
 pub(crate) const BUTTERFLY_GRAIN: usize = 1 << 10;
 
 impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
@@ -70,21 +59,9 @@ impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
                         if i != 0 {
                             t += steps[(first + i).trailing_zeros() as usize];
                         }
-                        let zero = t.is_zero();
                         let (lo, hi) = block.split_at_mut(half);
                         let butterfly = |lo: &mut [F], hi: &mut [F]| {
-                            if zero {
-                                // (u, v) ↦ (u, u + v)
-                                for (u, v) in lo.iter_mut().zip(hi) {
-                                    *v += *u;
-                                }
-                            } else {
-                                for (u, v) in lo.iter_mut().zip(hi) {
-                                    // (u, v) ↦ (u + t·v, u + t·v + v)
-                                    *u += t * *v;
-                                    *v += *u;
-                                }
-                            }
+                            packed_butterfly::<F, false>(lo, hi, t);
                         };
                         // Pairs are independent across the block, so a block wider than the
                         // grain is split further rather than run on a single thread.
@@ -126,21 +103,9 @@ impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
                         if i != 0 {
                             t += steps[(first + i).trailing_zeros() as usize];
                         }
-                        let zero = t.is_zero();
                         let (lo, hi) = block.split_at_mut(half);
                         let butterfly = |lo: &mut [F], hi: &mut [F]| {
-                            if zero {
-                                // (u', v') ↦ (u = u', v = u' + v')
-                                for (u, v) in lo.iter_mut().zip(hi) {
-                                    *v += *u;
-                                }
-                            } else {
-                                for (u, v) in lo.iter_mut().zip(hi) {
-                                    // (u', v') ↦ (u = u' + t·v, v = u' + v')
-                                    *v += *u;
-                                    *u += t * *v;
-                                }
-                            }
+                            packed_butterfly::<F, true>(lo, hi, t);
                         };
                         // Pairs are independent across the block, so a block wider than the
                         // grain is split further rather than run on a single thread.
@@ -158,13 +123,54 @@ impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
     }
 }
 
+/// Apply a butterfly to full SIMD vectors and any remaining scalar elements.
+#[inline]
+fn packed_butterfly<F: TowerLevel, const INVERSE: bool>(lo: &mut [F], hi: &mut [F], t: F) {
+    // Both sides have equal length, so their packed prefixes and tails pair exactly.
+    let (lo, lo_tail) = F::Packing::pack_slice_with_suffix_mut(lo);
+    let (hi, hi_tail) = F::Packing::pack_slice_with_suffix_mut(hi);
+    let zero = t.is_zero();
+    butterfly_values::<_, INVERSE>(lo, hi, t.into(), zero);
+    butterfly_values::<_, INVERSE>(lo_tail, hi_tail, t, zero);
+}
+
+/// Apply the same field identities to scalar or packed values.
+#[inline]
+fn butterfly_values<R: PrimeCharacteristicRing + Copy, const INVERSE: bool>(
+    lo: &mut [R],
+    hi: &mut [R],
+    t: R,
+    zero: bool,
+) {
+    if zero {
+        // A zero twiddle reduces both transform directions to (u, u + v).
+        for (u, v) in lo.iter_mut().zip(hi) {
+            *v += *u;
+        }
+    } else if INVERSE {
+        // Recover v first, then remove its twiddle contribution from u.
+        for (u, v) in lo.iter_mut().zip(hi) {
+            *v += *u;
+            *u += t * *v;
+        }
+    } else {
+        // Evaluate the pair as (u + t*v, u + t*v + v).
+        for (u, v) in lo.iter_mut().zip(hi) {
+            *u += t * *v;
+            *v += *u;
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use alloc::vec;
 
     use p3_binary_field::{
-        BinaryField8, BinaryField16, BinaryField32, BinaryField64, BinaryField128, TowerLevel,
+        BinaryField8, BinaryField16, BinaryField32, BinaryField64, BinaryField128, Ghash128,
+        TowerLevel,
     };
+    use p3_field::PrimeCharacteristicRing;
     use p3_matrix::Matrix;
     use p3_matrix::dense::RowMajorMatrix;
     use p3_util::log2_strict_usize;
@@ -227,6 +233,22 @@ mod tests {
 
     proptest! {
         #![proptest_config(ProptestConfig::with_cases(32))]
+
+        #[test]
+        fn ghash_packing_matches_naive_and_round_trips(
+            log_n in 0usize..=7,
+            width in 1usize..=9,
+            seed: u64,
+            shift: u64,
+        ) {
+            // Odd widths force scalar tails alongside SIMD prefixes.
+            check_matches_naive::<Ghash128>(log_n, width, seed, shift);
+            let coeffs = matrix::<Ghash128>(log_n, width, seed);
+            let shift = sample::<Ghash128>(shift);
+            let ntt = LchNtt::<Ghash128>::default();
+            let transformed = ntt.shifted_ntt_batch(coeffs.clone(), shift);
+            prop_assert_eq!(ntt.shifted_intt_batch(transformed, shift), coeffs);
+        }
 
         #[test]
         fn lch_matches_naive_at_8_bits(
@@ -292,8 +314,8 @@ mod tests {
             prop_assert_eq!(ntt.shifted_intt_batch(evals, shift), coeffs);
         }
 
-        /// `LchNtt`'s low-degree extension agrees with the oracle's, on a coset too, and the
-        /// input rows reappear as the prefix: the correspondence Phase 3 folds along.
+        /// The low-degree extension agrees with the oracle, on a coset too.
+        /// The input rows reappear as the prefix of the output.
         #[test]
         fn lch_lde_matches_naive(
             log_n in 0usize..=6,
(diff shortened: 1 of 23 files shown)


tcoratger:
@Nashtare Just another basis for the binary fields (the one used in Binius and Flock) so that the user has the choice to use the one which satisfies his needs. Both have pros and cons as far as I know.

tcoratger:
@Nashtare Also I'm working here on a Linux machine so Mac optimizations may be pretty bad, we can fix that in some followup PRs :)

Nashtare:
Awesome! I had Ghash128 in my todo list too, glad you handled it! Will drop the review shortly.

Regarding benchmarks on arm (fully serial):

```bash
  Serial NTT, width 16, Apple Silicon, PR head:

  ┌───────┬──────────┬──────────┬──────────────┐
  │ log n │  tower   │ Ghash128 │    ratio     │
  ├───────┼──────────┼──────────┼──────────────┤
  │ 14    │ 22.45 ms │ 2.02 ms  │ 11.1x faster │
  ├───────┼──────────┼──────────┼──────────────┤
  │ 16    │ 105.5 ms │ 9.47 ms  │ 11.1x faster │
  └───────┴──────────┴──────────┴──────────────┘

```

binary-field/src/packed/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+//! Only the widest such register is used, so there is one packing per build.
+//!
+//! The tower representation has no packing.
+//! A product there is table lookups, which no vector unit widens.
+
+#[cfg(all(
+    target_arch = "x86_64",
+    target_feature = "vpclmulqdq",
+    any(target_feature = "avx2", target_feature = "avx512f")
+))]
```
Nashtare, line 13: **CI never compiles this module.** No leg in `ci.yml` enables `vpclmulqdq` — the `+avx512f` compile-only leg doesn't imply it and neither does `+avx2` — so both `lanes` variants in `packed/x86_64.rs`, the `Packing = PackedGhash128` arm in `ghash.rs`, and the `_pext_u64` arm in `clmul/sqrt.rs` (`bmi2`) are exercised only on your machine.

Suggestion:
- extend the two `+pclmulqdq` test legs to `+pclmulqdq,+bmi2` (every GitHub x86 runner has BMI2, and it makes `unshuffle_matches_individual_coefficients` actually compare `_pext_u64` against the portable arm);
- add `+avx2,+vpclmulqdq` test legs — GitHub's current ubuntu runners (EPYC 7763 / Ice Lake and later) have VPCLMULQDQ, so `test_packed_binary_field!` and `tests/ghash.rs` would run for real; if a runner turns out not to have it, downgrade to `compile_only` like the AVX-512 leg;
- extend the `+avx512f` compile-only leg to `+avx512f,+vpclmulqdq` so the 512-bit `lanes` module compiles.

FWIW I cross-compiled `--all-targets` + clippy for `x86_64-unknown-linux-gnu` with `+pclmulqdq,+vpclmulqdq,+avx2` and again with `+avx512f`, both clean — this is about keeping it that way.

binary-field/src/packed/x86_64.rs
```diff
@@ -0,0 +1,468 @@
...
+/// Bit 0 picks the half of the first argument, bit 4 the half of the second.
+const LOW_BY_LOW: i32 = 0x00;
+
+/// Selects the high quadword of both operands.
+const HIGH_BY_HIGH: i32 = 0x11;
+
+/// Selects the high quadword of the first operand and the low quadword of the second.
+const HIGH_BY_LOW: i32 = 0x01;
+
+/// Exchanges the two 32-bit halves of each quadword, which swaps the halves of each lane.
```
Nashtare: `0x4e` is dword order `[2, 3, 0, 1]`: it swaps the two *quadwords* of each lane and keeps the dword order inside each quadword. It does not exchange the 32-bit halves of a quadword. Suggest:

```rust
/// Swaps the two quadwords of every lane, so `x ^ swap(x)` holds `x_lo ^ x_hi` in both halves.
```

binary-field/src/packed/x86_64.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    #[inline]
+    fn mul(self, rhs: Self) -> Self {
+        let (x, y) = (self.to_vector(), rhs.to_vector());
+
+        // The two diagonal half products.
+        let low = lanes::clmul::<LOW_BY_LOW>(x, y);
+        let high = lanes::clmul::<HIGH_BY_HIGH>(x, y);
+
+        // Karatsuba reaches the middle coefficient with one product instead of two.
```
Nashtare, line 289: `clmul/x86_64.rs::poly_mul_128` takes schoolbook (4 products) with the argument that a hardware multiply is too cheap to pay Karatsuba's two extra XORs on the critical path; this kernel takes Karatsuba (3 products + 2 shuffles + 3 XORs). Both can be right — on the 512-bit path `VPCLMULQDQ zmm` has half the throughput of `xmm` on Zen 4/5, so a saved product is worth more — but was schoolbook measured for the packed kernel too? A one-line note of which won would stop the next reader from "fixing" the inconsistency.

binary-field/src/packed/x86_64.rs
```diff
@@ -0,0 +1,468 @@
...
+use lanes::WIDTH;
+
+/// Several elements of the polynomial-basis `GF(2^128)`, one per 128-bit lane of a register.
+///
+/// Two under `AVX2`, four under `AVX-512`.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+// Needed to make the transmutes below sound.
+#[repr(transparent)]
+#[must_use]
+pub struct PackedGhash128([Ghash128; WIDTH]);
```
Nashtare, line 183: The PR description names `PackedGhash128AVX2` / `PackedGhash128AVX512`, but the code has a single `PackedGhash128` whose `WIDTH` depends on the build. Every other field in the workspace exposes one named type per ISA (`PackedMersenne31AVX2`, `…AVX512`, `…Neon`) and lets `Field::Packing` pick, which lets downstream code and tests name a specific width. Not blocking — either follow that convention or update the description.

binary-field/src/ghash.rs
```diff
@@ -0,0 +1,560 @@
...
+        target_feature = "vpclmulqdq",
+        any(target_feature = "avx2", target_feature = "avx512f")
+    ))]
+    type Packing = crate::PackedGhash128;
+    #[cfg(not(all(
+        target_arch = "x86_64",
+        target_feature = "vpclmulqdq",
+        any(target_feature = "avx2", target_feature = "avx512f")
+    )))]
+    type Packing = Self;
```
Nashtare: nit: this gate is spelled out four times (`ghash.rs` ×2, `lib.rs`, `packed/mod.rs` ×2, plus `clmul/mod.rs` for `TAIL_128`). A single `#[cfg]`-selected `pub(crate) type Packing = …;` alias in `packed/mod.rs` (with `type Packing = Ghash128;` in the other arm) would leave one place for the gate to drift from.

binary-field/src/clmul/portable.rs
```diff
@@ -0,0 +1,272 @@
...
+/// A product of residue `r` by residue `s` lands entirely in residue `r + s`.
+/// Sixteen integer products cover the four residues of the output.
+///
+/// # Why the carries never escape
+///
+/// Bit `p` counts the pairs `(i, j)` with `i + j = p` drawn from the two masks.
+/// A 64-bit word holds at most 16 positions in one residue class.
+///
+/// At most 15 pairs reach any `p <= 56`, which fits the three free bits above `p`.
+/// Only `p = 60` reaches 16, and its carry leaves the word entirely.
```
Nashtare: The bound holds but the statement is off by a residue class: with `p = 4k + (r + s)` the pair count is `k + 1`, so positions 57–59 also see 15 pairs (the classes with `r + s = 1, 2, 3`) and 61–63 also see 16. Suggest `p <= 59` / `p >= 60`. Since this paragraph is the whole correctness argument for the software multiply, it is worth being exact.

binary-field/src/clmul/sqrt.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Gather the even-numbered bits into the low half of a word.
+// The shift-and-mask fallback is `const` and the bit-extract instruction is not, so leaving
+// this one out keeps the signature the same on every target.
+#[allow(clippy::missing_const_for_fn)]
+#[inline]
+fn compact_even(x: u64) -> u64 {
+    #[cfg(all(target_arch = "x86_64", target_feature = "bmi2"))]
+    {
+        // SAFETY: the crate is compiled with BMI2 enabled.
+        unsafe { _pext_u64(x, 0x5555_5555_5555_5555) }
```
Nashtare: On Zen 1 / Zen 2, `PEXT` is microcoded (≈18-cycle latency *and* throughput per uops.info), so `-C target-cpu=native` on those parts makes this arm slower than the six-step shift-and-mask it replaces. Given how little sqrt matters, either drop the intrinsic arm or note the trade-off. Not blocking.

binary-field/benches/arithmetic.rs
```diff
@@ -238,12 +241,278 @@ fn bench_flatten_to_base(c: &mut Criterion) {
...
+            black_box(&ghash).iter().fold(Ghash128::ONE, |acc, &y| {
+                let x = acc + y;
+                x * x
+            })
+        });
+    });
+    group.finish();
+}
+
+/// Inversion in the polynomial basis, which routes through the tower and pays two conversions.
```
Nashtare: Stale: on `pclmulqdq` / `aes` targets `Ghash128::inverse` is the Itoh–Tsujii chain in `clmul/inverse.rs`; only the `through tower` arm below routes through the tower. Also, `through tower` iterates `ghash` while `ghash` iterates `nonzero` — use `nonzero` for both so the arms see identical inputs (and `inverse()` cannot panic on a zero draw).

binary-field/README.md
```diff
@@ -1,15 +1,42 @@
...
+  another;
+- the polynomial basis of `x^128 + x^7 + x^2 + x + 1`, the GHASH modulus.
 
 Key items:
 
 - `Gf2` — the base field `GF(2)`
 - `BinaryField2`, `BinaryField4`, `BinaryField8`, `BinaryField16`, `BinaryField32`, `BinaryField64`, `BinaryField128` — the tower levels, each a quadratic extension of the one below
-- `BasedVectorSpace` / `ExtensionField` between every pair of byte-aligned levels, in the tower basis
+- `Ghash128` — `GF(2^128)` in the GHASH polynomial basis, with `From` conversions to and from the widest tower level
+- `PackedGhash128` — the SIMD packing, holding two elements per register under `AVX2` and four under `AVX-512`
```
Nashtare: `AVX2` alone does not produce this type; it needs `vpclmulqdq` as well (`lib.rs` / `packed/mod.rs` gates). Worth saying here, since `-C target-feature=+avx2` is what people reach for — and on docs.rs (no target features) the type will not appear in the rendered docs at all.

binary-field/src/poly_basis.rs
```diff
@@ -1,25 +1,17 @@
-//! `GF(2^128)` in the polynomial basis of `x^128 + x^7 + x^2 + x + 1`.
+//! Raw polynomial coordinates for `GF(2^128)` modulo `x^128 + x^7 + x^2 + x + 1`.
```
Nashtare, line 1: With `Ghash128` typed (`From` both ways, `Mul`, `square`, and `HAS_HARDWARE_CLMUL`), this module's `to_poly` / `to_tower` / `mul` / `square` on bare `u128`s now duplicates it without the type safety the module doc itself argues for. Suggest a follow-up that ports `PolyBasisNtt` to `Ghash128` and deprecates `poly_basis`.

binary-dft/src/lch.rs
```diff
@@ -11,36 +12,24 @@ use p3_util::log2_strict_usize;
...
-/// piece per side and instead gather `BUTTERFLY_GRAIN / half` whole blocks into one task, so a
-/// task is a piece of this size on either side of the crossover.
-///
-/// At a few nanoseconds per butterfly a piece of this size is microseconds of work, well above
-/// the cost of handing a task to another thread, while still leaving hundreds of pieces per
-/// stage at the smallest useful heights.
+/// Narrow blocks share a task.
+/// Wide blocks split into pieces of this size, which have too few blocks to fill a machine.
+/// Either way a stage leaves enough independent tasks to be worth handing to other threads.
 pub(crate) const BUTTERFLY_GRAIN: usize = 1 << 10;
```
Nashtare: The rewritten doc lost the argument that made this constant understandable: stage `j` has `2^(ℓ−1−j)` blocks of `half = 2^j · width` per side, so pieces of this size leave `n · width / (2 · BUTTERFLY_GRAIN)` tasks at *every* stage independent of `j`, and stages with `half ≤ BUTTERFLY_GRAIN` gather `BUTTERFLY_GRAIN / half` whole blocks per task. The new second sentence also reads as if the *pieces* are what have too few blocks.

Same for the `LchNtt` doc above (the `2/ℓ` zero-twiddle share) and several test docs further down. None of these rewrites is needed by the packing change — I'd keep them out of this PR, or at least restore the quantitative parts.

binary-field/src/challenger.rs
```diff
@@ -282,6 +283,34 @@ mod tests {
...
+        // Both representations are 16 bytes wide with every pattern a valid element, so a
+        // transcript over one is byte-for-byte the transcript over the other.
+        //
+        //     observe tower element -> same bytes absorbed -> same bytes drawn
+        //
+        // What differs is only how those bytes are read as a field element.
+        let element = a128();
+
+        let mut tower = mk::<BinaryField128>();
+        tower.observe(element);
```
Nashtare, line 298: nit: `Ghash128::from_le_byte_iter(element.into_bytes().into_iter())` is what the sampling path itself uses, and avoids the `Vec` + `try_into`.

binary-field/src/clmul/inverse.rs
```diff
@@ -0,0 +1,136 @@
+//! Itoh–Tsujii inversion with precomputed Frobenius maps.
+//!
+//! The inverse of a nonzero `x` is `x^(2^128 - 2)`.
+//! An addition chain reaches that exponent in ten products, once squaring is cheap.
+//!
+//! Squaring `2^k` times is `GF(2)`-linear, so each such step is one fixed matrix.
+//! The five the chain needs are tabulated a byte at a time, at compile time.
```
Nashtare, line 7: Data point from Apple Silicon (`aes` backend, serial, per element): `representations/inverse/ghash` **110.7 ns** vs `through tower` **110.9 ns**. On this target the chain is exactly as fast as converting to the tower and inverting there, so the 320 KiB of Frobenius maps buy nothing here.

The cause is the dependent product: 7.8 ns on this box against the 3.55 ns you measured on Zen 5 (the tower product is 19.0 vs your 19.8 ns, so the core is not slower). The AArch64 codegen itself is fine — `vgetq_lane_u64` + `vmull_p64` does fold into `pmull.1q` on vector registers — but every scalar product pays a GPR→NEON→GPR round trip (`fmov d, x` / `dup.2d` on the way in, `mov.d x, v[1]` / `fmov x, d` on the way out) because a `u128` lives in a GPR pair, and on a dependent chain that is roughly 8 of the ~27 cycles. Ten dependent products × 7.8 ns is ~78 ns of the 110.

Not blocking. Two follow-ups worth considering: (1) gate `mod inverse` to `x86_64` until AArch64 latency is looked at, since the tables are dead weight there; (2) the only way to remove the round trip is a NEON-register-backed representation (a width-1 `Packing` for AArch64), which is a separate design.

binary-field/src/clmul/mod.rs
```diff
@@ -24,31 +61,21 @@ mod x86_64;
...
 /// Whether the target has a carryless-multiply instruction.
 ///
-/// The bit-serial fallback below keeps every routine in this module correct everywhere, but it
-/// is far slower than the recursive tower arithmetic, so `Mul` and `square` only route through
-/// this module when the instruction is really there.
-///
-/// This is a compile-time decision, and neither feature is in the baseline of most targets:
-/// `aarch64-apple-darwin` has `aes`, but generic AArch64 Linux and every `x86_64` target need
-/// `-C target-feature=+aes` / `+pclmulqdq` (or `-C target-cpu=native`) for the fast path.
+/// - The tower routes here only when it does, since its own recursion beats the software path.
```
Nashtare: This is no longer true now that the software backend is the masked-multiply construction rather than the bit-serial loop. With `-C target-feature=-aes` on Apple Silicon (serial, per element): tower `mul/128/dispatched` — the `karatsuba_mul` recursion — **64.9 ns** latency / **64.4 ns** throughput; the `Ghash128` software product **16.7 / 12.9 ns**; each change of basis 3.2–3.4 ns. Routing `mul_128` through `portable` on non-clmul targets would therefore land around 30 ns against 65 — about 2x — and that is the build every x86 user gets without `-C target-cpu=native` (and what the default CI legs run).

Worth a follow-up PR: make `dispatched_mul` always take the clmul route and keep `karatsuba_mul` as the test oracle, after checking wasm32, where `reverse_bits` has no instruction. At minimum this sentence should go.

Nashtare (approved):

binary-field/src/packed/mod.rs
tcoratger: Good catch, and the important one — thank you. Fixed in `ci.yml`:

- `+avx2,+vpclmulqdq` and `+avx512f,+vpclmulqdq` legs added, both `compile_only` with `target: x86_64-unknown-linux-gnu`, so the 256- and 512-bit `lanes` modules and the `Packing` arm are built on every PR.
- I made the AVX2 one compile-only rather than a test leg for the reason you flagged: VPCLMULQDQ arrived with Ice Lake and Zen 3, and GitHub's ubuntu pool still includes Cascade Lake parts, so a test leg would SIGILL on some runners. That matches how `+avx512f` and `+sve2` are already handled.
- I did not add `+bmi2`: the `_pext_u64` arm is gone entirely (see the `sqrt.rs` thread), so there is no longer a BMI2-specific path to cover.

Reproduced both new legs locally, `build --all-targets` and `clippy -D warnings`, clean.

binary-field/src/packed/x86_64.rs
tcoratger: You're right, the doc was simply wrong — `0x4e` is dword order `[2, 3, 0, 1]`, which swaps the quadwords and leaves the dwords inside each one alone. Taken your wording verbatim.

binary-field/src/packed/x86_64.rs
tcoratger: It was measured, and the inconsistency is deliberate — but you're right that nothing said so. I hit this the hard way: when I removed the macro I rewrote this kernel from the scalar (schoolbook) form and lost 17.5% until I diffed against the previous revision.

Zen 5, four lanes, independent chains: **0.47 ns/element** Karatsuba against **0.55 ns** schoolbook. Your reading of why is exactly it — `VPCLMULQDQ zmm` has half the throughput of `xmm`, so the saved product outweighs the two shuffles and three XORs, and at 128 bits it does not.

Added those numbers as a comment at the Karatsuba step.

binary-field/src/packed/x86_64.rs
tcoratger: Updated the description — that was stale, it predates the macro removal.

I kept the single name deliberately, and I think it is the better trade here, though I'll happily switch if you disagree. The two gates are mutually exclusive (`avx2 && !avx512f` versus `avx512f`), so unlike `PackedMersenne31AVX2`/`…AVX512` only one of the two would ever exist in a given build. A downstream `use p3_binary_field::PackedGhash128AVX2;` would therefore compile until someone enabled AVX-512 and then break, which is the failure mode the single name removes. Code wanting a specific width still can't get one portably either way, so `Field::Packing` stays the honest entry point.

binary-field/src/ghash.rs
tcoratger: Done — `packed/mod.rs` now owns a `pub(crate) use … as Packing` alias and `ghash.rs` reads `type Packing = crate::packed::Packing;` with no gate at all.

The gate still appears in `packed/mod.rs` (module, re-export, and the two alias arms) and once in `lib.rs` and `clmul/mod.rs`, so the raw count is similar — but the *decision* now lives in one file instead of being restated where it's consumed, which was your point. One wrinkle: on a target with no packing the alias resolves to `Ghash128` itself, so `clippy::use_self` fires on the associated type; that's `#[allow]`ed with the reason inline.

binary-field/src/clmul/portable.rs
tcoratger: You're right and I'm glad you checked this one — it is the whole correctness argument for the software multiply, and my statement was silently residue-0-specific.

With `p = 4k + r + s` the count is `k + 1` and `k <= 15`, so 57–59 also carry 15 and 61–63 also carry 16. Rewritten to state the general form and use `p <= 59` / 60–63, so the bound is now argued rather than asserted for one class.

binary-field/src/clmul/sqrt.rs
tcoratger: Dropped the intrinsic arm. `PEXT` being ~18-cycle microcode on Zen 1/2 means `-C target-cpu=native` there would have made this *slower* than what it replaced, which is a bad trade for an operation this rare — the 89x win (510 ns to 5.7 ns) comes from not doing 127 squarings, not from the instruction.

Removing it also took a `bmi2` cfg, a `#[cfg_attr(…, allow(dead_code))]`, and a conditional test assertion with it, and let `unshuffle` become `const fn` on every target. The trade-off is noted on the function.

binary-field/benches/arithmetic.rs
tcoratger: Both fixed. The doc now says what the two arms actually compare, and `through tower` iterates `nonzero` like the other arm — you're right that it could have hit `inverse()` on a zero draw, and it was comparing different inputs besides.

binary-field/README.md
tcoratger: Fixed — the entry now says `avx2` or `avx512f` **and** `vpclmulqdq`, and notes that the type is therefore absent from the rendered docs on docs.rs.

binary-field/src/poly_basis.rs
tcoratger: Agreed, and it's the right shape for a follow-up rather than this PR. `PolyBasisNtt` exists only to work around the tower product's basis conversions, which `Ghash128` removes — so porting it and deprecating `poly_basis` deletes the workaround rather than just re-typing it. I'd rather do that against a landed `Ghash128` so the before/after NTT numbers are attributable.

The module doc now points at the typed field as the default and scopes these helpers to caller-managed buffers.

binary-dft/src/lch.rs
tcoratger: Fair, and I've reverted them. Every doc block in `lch.rs` that the packing change did not require is back to its base-commit text, including the `2^(ℓ−1−j)` / `n · width / (2 · BUTTERFLY_GRAIN)` argument and the `2/ℓ` zero-twiddle share — you're right that the replacement also misread which thing has too few blocks.

The file's diff is now only the packed butterfly, its two new helpers, and the cross-representation test. Sorry for the noise; those rewrites came from an earlier pass over this branch and had no business here.

binary-field/src/challenger.rs
tcoratger: Done — `Ghash128::from_le_byte_iter(element.into_bytes().into_iter())`, which is also the constructor the sampling path itself uses, so the test now mirrors it exactly.

binary-field/src/clmul/inverse.rs
tcoratger: Thank you for measuring this — that's decisive, and I've taken follow-up (1) now rather than later: `mod inverse` is gated to `x86_64` with `pclmulqdq`, so AArch64 and the software path both take the tower norm and neither compiles the 320 KiB of maps. Confirmed with `nm` that the `POW_*` symbols are gone from the AArch64 rlib.

110.7 vs 110.9 ns means the tables were pure dead weight there, and your diagnosis of the cause matches what the packing work turned up: the whole reason `PackedGhash128` is worth 4x on x86 is that the element stays in a vector register instead of round-tripping through a GPR pair. A width-1 NEON-backed packing is the same fix for AArch64, and I agree it's a separate design.

binary-field/src/clmul/mod.rs
tcoratger: You're right, and the numbers make it plain — 16.7 ns against 64.9 ns is not close. The sentence claiming the tower recursion beats the software path is gone; the doc now just states which route each type takes.

I've left the routing change itself for a follow-up, since making `dispatched_mul` always take the clmul path needs the wasm32 check you mention (`reverse_bits` has no instruction there, and it is on the critical path of the bit-reversal trick that recovers the high half). Worth its own numbers rather than riding along here.

PR #2032, opened 2026-09-05
feat(binary-pcs): multilinear commitment over binary tower fields

Nashtare:
Adds `p3-binary-pcs`: a multilinear polynomial commitment scheme over `BinaryField128`, closing Phase 3 of the binary-field work on top of `p3-binary-field` (#2001), `p3-binary-dft` (#2003) and ring switching (#2006).

## Construction

The committed object is the stacked multilinear that `p3_sumcheck::layout` already produces. It is encoded as a single Reed-Solomon codeword column over the Cantor `F_2`-subspace domain, via `AdditiveRsEncoder`, and Merkle-committed.

Opening is BaseFold-shaped. The residual sumcheck runs one round per variable, and each round's challenge is used twice: it binds one multilinear variable, and it folds the codeword 2-to-1 through `W_1(x) = x^2 + x`. The two stay in correspondence because the fold is the evaluation-basis combination `(1 - b) * f_0 + b * f_1`, which is exactly what `Poly::fix_suffix_var` computes on the message.

Two structural facts make the fold cheap, and both are pinned by tests:

- `W_1(domain_point(i)) == domain_point(i >> 1)`, so the folded domain is the same function at a halved index. There are no per-round twiddle tables and no domain type.
- `domain_point(2j + 1) == domain_point(2j) + 1`, so folding partners are adjacent in memory.

Verification replays the transcript, checks one Merkle multiproof per round, and checks that consecutive rounds are related by `fold_pair`. The claim closes by requiring the final codeword to be uniform and to satisfy `claimed_sum == w(r) * constant`.

## Soundness

Parameters are derived only in the **unique-decoding** regime, and `SecurityAssumption` is deliberately not re-exported, so the refuted capacity bound is unreachable rather than runtime-rejected. Capacity-rate list decodability is refuted over characteristic 2 with `F_2`-subspace domains (BGKS20 App. B, BCHKS25 Thm 1.6 / Cor 1.7), and the Cantor domain is exactly that case. The Johnson bound is *not* refuted, but `p3-security` documents it as resting on a correlated-agreement conjecture, so it is excluded here by choice. The cost is honest and large: more queries than WHIR or STIR report for the same target.

`security_level` is capped against what the alphabet can deliver. Queries buy back only the proximity term; the commit phase costs `(n + 1)/|F|` per round at arity 2 and each sumcheck round costs `2/|F|`, and no query count recovers either. `BinaryPcsConfig::field_security_bits` charges both, `try_new` rejects a target above it, and a test pins the value against `p3_security::fri::commit_phase_error_udr` so the two cannot drift. At `2^20` variables and rate `2^-2` that ceiling is 105 bits.

Query indices are drawn with `CanSampleUniformBits`, never by bit-decomposing a sampled field element.

## Benchmarks

Serial, on `aarch64-apple-darwin`, where `PMULL` is in the baseline. On x86_64 these need `-C target-feature=+pclmulqdq` or the 128-bit multiply falls back to the recursive tower path.

| variables | commit | open | verify | proof size |
|---|---|---|---|---|
| 16 | 171 ms | 206 ms | 4.4 ms | 291,116 B |
| 18 | 841 ms | 760 ms | 5.6 ms | 393,810 B |
| 20 | 3.03 s | 4.22 s | 7.1 ms | 514,147 B |

Proof size is dominated by the query count, which is the price of unique decoding rather than of the polynomial's arity.

The codeword fold chains `domain_point` across each parallel task by an XOR recurrence instead of recomputing it per output symbol. Measured against the per-call form: 1.6x to 2.1x, with the tightest measurement at 18 variables, 31.9 ms against 50.7 ms.

## Design notes

- **The layout is a fixed type, not a parameter.** The fold merges adjacent pairs, which only suffix-order binding matches; prefix binding merges halves and does not stay in lockstep. `prefix_layout_does_not_stay_in_lockstep` pins that as a property of the two binding orders rather than a tuning detail.
- **There is no folding knob.** The commit phase lays the polynomial out as a single column, so every variable folds. Binding a prefix inside a committed row is a deferred prover-cost optimisation; an exhaustive search over 128 order, orientation and bind-rule combinations found none that composes with the subsequent folds, so it is not on the path to a correct PCS.
- **The final codeword is not absorbed into the transcript.** Uniformity forces all `2^log_inv_rate` symbols to one value, and any single query's last fold step pins that value against a Merkle-authenticated, transcript-bound round codeword. Late adaptivity therefore buys nothing. Weaken either half and the absorption has to come back.

## Test plan

- [x] `cargo test -p p3-binary-pcs`, with and without `--features parallel`: 35 lib + 18 integration tests.
- [x] The `2^16` round trip in release, which is this phase's exit criterion. Wired into `ci-heavy.yml`, now built with carry-less multiply.
- [x] The fold identity is proptested against `NaiveAdditiveNtt` as an independent oracle, over generated message lengths and blowup rates.
- [x] Negative tests: every one asserts a *named* typed error, not just `is_err()`. They cover a tampered final-codeword symbol, a tampered opened value at the base and at an intermediate round, a truncated round vector reached without touching the challenger, a mismatched opening protocol, a mismatched commitment, a corrupted grinding witness, non-empty sumcheck PoW witnesses, permuted round commitments, and malformed shapes.
- [x] Evaluation binding: a false claimed evaluation and a tampered sumcheck round message are each rejected.
- [x] Fold-chain reachability: a round committed as a valid codeword that is not the fold of its predecessor is caught only by the fold chain.
- [x] `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt --check`, `cargo machete --with-metadata`, rustdoc with `--document-private-items`, and a `thumbv7em-none-eabi` no_std build.

### DIFF
--- a/binary-pcs/src/fold.rs
+++ b/binary-pcs/src/fold.rs
@@ -1,0 +1,268 @@
+//! Folding a codeword over the Cantor additive domain.
+//!
+//! A codeword symbol at domain point `x` and its partner at `x + v_0 = x + 1` determine the
+//! two halves of the novel-basis decomposition
+//!
+//! ```text
+//!     f(x) = f_0(W_1(x)) + x * f_1(W_1(x))
+//! ```
+//!
+//! so `f_1 = f(x) + f(x + 1)` and `f_0 = f(x) + x * f_1`. The fold at `beta` is the
+//! evaluation-basis combination `(1 - beta) * f_0 + beta * f_1`, which over characteristic 2 is
+//! `f_0 + beta * (f_0 + f_1)`. This combination is chosen because it is exactly what
+//! `Poly::fix_suffix_var` computes on the message: folding the codeword and binding the
+//! multilinear's lowest variable in the evaluation basis are the same operation, which is what
+//! lets the sumcheck and the codeword move in lockstep.
+//!
+//! Because `W_1(x) = x^2 + x` is `F_2`-linear with `W_1(v_0) = 0` and `W_1(v_i) = v_{i-1}`, and
+//! `domain_point` is `F_2`-linear too, `W_1(domain_point(i)) = domain_point(i >> 1)`: the folded
+//! domain is the same function at a halved index, so no per-round domain state is carried. The
+//! folding partners `domain_point(2j)` and `domain_point(2j + 1)` differ by `v_0 = 1`, so they
+//! are adjacent rows in memory.
+
+use alloc::vec::Vec;
+
+use p3_binary_dft::domain_point;
+use p3_binary_field::{BinaryField128, TowerLevel};
+use p3_field::PrimeCharacteristicRing;
+use p3_maybe_rayon::prelude::*;
+
+/// Fold one pair of the codeword.
+///
+/// `lo` is the symbol at `domain_point(2 * index)`, `hi` the one at `domain_point(2 * index + 1)`.
+#[inline]
+pub fn fold_pair(
+    index: usize,
+    beta: BinaryField128,
+    lo: BinaryField128,
+    hi: BinaryField128,
+) -> BinaryField128 {
+    let x: BinaryField128 = domain_point(index << 1);
+    let f1 = lo + hi;
+    let f0 = lo + x * f1;
+    f0 + beta * (f0 + f1)
+}
+
+/// Output pairs one parallel task chains together, opened by a single `domain_point` call.
+///
+/// Large enough that the task-opening `domain_point` call is amortised over many folds; small
+/// enough to leave real parallelism at every codeword length this crate exercises.
+const FOLD_GRAIN: usize = 1 << 10;
+
+/// The XOR step from `domain_point(2 * j)` to `domain_point(2 * (j + 1))`, indexed by
+/// `(j + 1).trailing_zeros()`.
+///
+/// `domain_point` is `F_2`-linear, and `j XOR (j + 1)` sets exactly the trailing ones of `j`
+/// together with the first zero bit above them — bits `0..=level` for `level =
+/// (j + 1).trailing_zeros()`. So `domain_point(2 * j) + domain_point(2 * (j + 1))` is
+/// `domain_point` of that same run of bits shifted up by one: the sum of `cantor_basis(1)`
+/// through `cantor_basis(level + 1)`. `steps[level]` is that sum, computed once and reused by
+/// every transition that lands at the same level.
+fn gray_chain_steps(num_pairs: usize) -> Vec<BinaryField128> {
+    let levels = num_pairs.next_power_of_two().trailing_zeros() as usize;
+    let mut steps = Vec::with_capacity(levels);
+    let mut acc = BinaryField128::ZERO;
+    for level in 0..levels {
+        acc += BinaryField128::cantor_basis(level + 1);
+        steps.push(acc);
+    }
+    steps
+}
+
+/// Fold a whole codeword, halving its length.
+///
+/// Each parallel task opens with one `domain_point` call, then chains it across the task's own
+/// pairs by adding a precomputed per-level `F_2`-linear step at each transition: `fold_pair`'s
+/// single per-call `domain_point` is the right form for one query's verifier-side check, but
+/// here the same value would otherwise be recomputed once per output symbol of the whole
+/// codeword.
+///
+/// # Panics
+///
+/// Panics if `codeword` has odd length. A codeword always has a power-of-two length, so an odd
+/// one means the caller has lost track of the round schedule.
+#[must_use]
+pub fn fold_codeword(codeword: &[BinaryField128], beta: BinaryField128) -> Vec<BinaryField128> {
+    assert!(
+        codeword.len().is_multiple_of(2),
+        "codeword length must be even"
+    );
+
+    let num_pairs = codeword.len() / 2;
+    let steps = gray_chain_steps(num_pairs);
+
+    codeword
+        .par_chunks(2 * FOLD_GRAIN)
+        .enumerate()
+        .flat_map_iter(|(block, chunk)| {
+            let start = block * FOLD_GRAIN;
+            let mut x: BinaryField128 = domain_point(start << 1);
+            let steps = &steps;
+            chunk.chunks(2).enumerate().map(move |(offset, pair)| {
+                if offset != 0 {
+                    x += steps[(start + offset).trailing_zeros() as usize];
+                }
+                let (lo, hi) = (pair[0], pair[1]);
+                let f1 = lo + hi;
+                let f0 = lo + x * f1;
+                f0 + beta * (f0 + f1)
+            })
+        })
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use alloc::vec::Vec;
+
+    use p3_binary_dft::{AdditiveNtt, NaiveAdditiveNtt, domain_point};
+    use p3_binary_field::{BinaryField128, TowerLevel};
+    use p3_field::PrimeCharacteristicRing;
+    use p3_matrix::dense::RowMajorMatrix;
+    use p3_multilinear_util::poly::Poly;
+    use proptest::prelude::*;
+    use rand::rngs::SmallRng;
+    use rand::{RngExt, SeedableRng};
+
+    use super::{fold_codeword, fold_pair};
+
+    /// Encode novel-basis coefficients over the additive domain, through the oracle.
+    ///
+    /// `NaiveAdditiveNtt` evaluates the definition directly and depends on no identity the
+    /// fold also relies on, so agreement between the two is real evidence rather than a
+    /// restatement.
+    fn encode(coeffs: &[BinaryField128]) -> Vec<BinaryField128> {
+        NaiveAdditiveNtt::<BinaryField128>::default()
+            .ntt_batch(RowMajorMatrix::new(coeffs.to_vec(), 1))
+            .values
+    }
+
+    #[test]
+    fn the_folded_domain_is_domain_point_at_halved_index() {
+        // W_1(x) = x^2 + x, and W_1(domain_point(i)) == domain_point(i >> 1).
+        for i in 0..512usize {
+            let x: BinaryField128 = domain_point(i);
+            assert_eq!(
+                x.square() + x,
+                domain_point::<BinaryField128>(i >> 1),
+                "i={i}"
+            );
+        }
+    }
+
+    #[test]
+    fn folding_partners_are_adjacent_and_differ_by_one() {
+        for j in 0..256usize {
+            let lo: BinaryField128 = domain_point(2 * j);
+            let hi: BinaryField128 = domain_point(2 * j + 1);
+            assert_eq!(lo + BinaryField128::ONE, hi, "j={j}");
+        }
+    }
+
+    /// Folding the codeword equals binding the message's lowest variable in the evaluation
+    /// basis (`Poly::fix_suffix_var`), for codewords blown up at several rates: every real
+    /// prover fold runs on one.
+    #[test]
+    fn folding_the_codeword_equals_binding_the_lowest_variable() {
+        let mut rng = SmallRng::seed_from_u64(0xF01D);
+        for log_inv_rate in 0..=2 {
+            for log_n in 1..=7 {
+                let n = 1usize << log_n;
+                for trial in 0..4 {
+                    let coeffs: Vec<BinaryField128> =
+                        (0..n).map(|_| rng.random::<BinaryField128>()).collect();
+                    let beta: BinaryField128 = rng.random();
+
+                    let mut message = coeffs.clone();
+                    message.resize(n << log_inv_rate, BinaryField128::ZERO);
+
+                    let bound = Poly::new(coeffs).fix_suffix_var(beta);
+                    let mut bound_message = bound.into_evals();
+                    bound_message.resize(bound_message.len() << log_inv_rate, BinaryField128::ZERO);
+
+                    assert_eq!(
+                        fold_codeword(&encode(&message), beta),
+                        encode(&bound_message),
+                        "log_inv_rate={log_inv_rate} log_n={log_n} trial={trial}"
+                    );
+                }
+            }
+        }
+    }
+
+    /// The chained form must agree with `fold_pair`'s independent per-call computation at
+    /// every position, including across `FOLD_GRAIN`'s block boundary: 4096 codeword symbols
+    /// is 2048 pairs, past `FOLD_GRAIN`'s 1024, so this covers two full parallel tasks and
+    /// exercises the cross-block chaining — block-start `domain_point(start << 1)` plus
+    /// `(start + offset).trailing_zeros()` indexing — that a single-block codeword leaves
+    /// silent.
+    #[test]
+    fn the_pair_form_agrees_with_the_vector_form() {
+        let mut rng = SmallRng::seed_from_u64(0x9E37_79B9_7F4A_7C15);
+        let codeword: Vec<BinaryField128> = (0..4096).map(|_| rng.random()).collect();
+        let beta: BinaryField128 = rng.random();
+        let folded = fold_codeword(&codeword, beta);
+        for (j, &value) in folded.iter().enumerate() {
+            assert_eq!(
+                value,
+                fold_pair(j, beta, codeword[2 * j], codeword[2 * j + 1]),
+                "j={j}"
+            );
+        }
+    }
+
+    #[test]
+    #[should_panic(expected = "codeword length must be even")]
+    fn an_odd_length_codeword_is_rejected() {
+        let odd = [BinaryField128::ONE; 3];
+        let _ = fold_codeword(&odd, BinaryField128::ONE);
+    }
+
+    /// `f_0 + beta * f_1`, the novel-basis combination, is a different operation from
+    /// `fold_pair`'s evaluation-basis combination: the two differ by exactly `beta * f_0`, so
+    /// they agree only when that term vanishes.
+    #[test]
+    fn the_novel_basis_fold_is_a_different_operation() {
+        let mut rng = SmallRng::seed_from_u64(0x51DE);
+        let j = 5usize;
+        let x: BinaryField128 = domain_point(2 * j);
+        let lo: BinaryField128 = rng.random();
+        let hi: BinaryField128 = rng.random();
+        let beta: BinaryField128 = rng.random();
+
+        let f1 = lo + hi;
+        let f0 = lo + x * f1;
+        let novel_basis_fold = lo + (x + beta) * (lo + hi);
+
+        assert_eq!(
+            fold_pair(j, beta, lo, hi) + novel_basis_fold,
+            beta * f0,
+            "the evaluation-basis and novel-basis folds differ by beta * f_0"
+        );
+    }
+
+    proptest! {
+        #[test]
+        fn folding_is_affine_in_the_challenge(
+            lo_raw: u128, hi_raw: u128, b0_raw: u128, b1_raw: u128, j in 0usize..64,
+        ) {
+            let lo = BinaryField128::from_repr(lo_raw);
+            let hi = BinaryField128::from_repr(hi_raw);
+            let b0 = BinaryField128::from_repr(b0_raw);
+            let b1 = BinaryField128::from_repr(b1_raw);
+            let x: BinaryField128 = domain_point(2 * j);
+            let f1 = lo + hi;
+            let f0 = lo + x * f1;
+
+            // fold(beta) = f0 + beta * (f0 + f1), so fold(b0) + fold(b1) == (b0 + b1) * (f0 + f1).
+            prop_assert_eq!(
+                fold_pair(j, b0, lo, hi) + fold_pair(j, b1, lo, hi),
+                (b0 + b1) * (f0 + f1)
+            );
+
+            // At beta = 0 the fold is f0, pinning fold_pair's internal domain index against one
+            // computed independently here.
+            prop_assert_eq!(fold_pair(j, BinaryField128::ZERO, lo, hi), f0);
+        }
+    }
+}
--- a/binary-pcs/src/fold.rs
+++ b/binary-pcs/src/fold.rs
@@ -159,34 +159,47 @@ mod tests {
         }
     }
 
-    /// Folding the codeword equals binding the message's lowest variable in the evaluation
-    /// basis (`Poly::fix_suffix_var`), for codewords blown up at several rates: every real
-    /// prover fold runs on one.
-    #[test]
-    fn folding_the_codeword_equals_binding_the_lowest_variable() {
-        let mut rng = SmallRng::seed_from_u64(0xF01D);
-        for log_inv_rate in 0..=2 {
-            for log_n in 1..=7 {
-                let n = 1usize << log_n;
-                for trial in 0..4 {
-                    let coeffs: Vec<BinaryField128> =
-                        (0..n).map(|_| rng.random::<BinaryField128>()).collect();
-                    let beta: BinaryField128 = rng.random();
+    proptest! {
+        // Each case runs the naive oracle twice over a codeword of up to 512 symbols, so the
+        // per-case cost is far above a typical property test's; 64 cases still explore every
+        // (rate, length) pair many times over.
+        #![proptest_config(ProptestConfig::with_cases(64))]
 
-                    let mut message = coeffs.clone();
-                    message.resize(n << log_inv_rate, BinaryField128::ZERO);
+        /// The crate's central identity: folding the codeword equals binding the message's
+        /// lowest variable in the evaluation basis (`Poly::fix_suffix_var`).
+        ///
+        /// The oracle is [`NaiveAdditiveNtt`], which evaluates the definition directly and
+        /// shares no identity with the fold, so agreement is evidence rather than restatement.
+        /// `log_n` and `log_inv_rate` are generated rather than swept because the identity is
+        /// claimed for every message length and every blowup a prover can configure, not for a
+        /// chosen list; every real prover fold runs on a blown-up codeword, so rate 0 is the
+        /// degenerate end of the range rather than the case of interest.
+        #[test]
+        fn folding_the_codeword_equals_binding_the_lowest_variable(
+            log_inv_rate in 0usize..=2,
+            log_n in 1usize..=7,
+            raw in prop::collection::vec(any::<u128>(), 128),
+            beta_raw: u128,
+        ) {
+            let n = 1usize << log_n;
+            let coeffs: Vec<BinaryField128> = raw[..n]
+                .iter()
+                .copied()
+                .map(BinaryField128::from_repr)
+                .collect();
+            let beta = BinaryField128::from_repr(beta_raw);
 
-                    let bound = Poly::new(coeffs).fix_suffix_var(beta);
-                    let mut bound_message = bound.into_evals();
-                    bound_message.resize(bound_message.len() << log_inv_rate, BinaryField128::ZERO);
+            let mut message = coeffs.clone();
+            message.resize(n << log_inv_rate, BinaryField128::ZERO);
 
-                    assert_eq!(
-                        fold_codeword(&encode(&message), beta),
-                        encode(&bound_message),
-                        "log_inv_rate={log_inv_rate} log_n={log_n} trial={trial}"
-                    );
-                }
-            }
+            let bound = Poly::new(coeffs).fix_suffix_var(beta);
+            let mut bound_message = bound.into_evals();
+            bound_message.resize(bound_message.len() << log_inv_rate, BinaryField128::ZERO);
+
+            prop_assert_eq!(
+                fold_codeword(&encode(&message), beta),
+                encode(&bound_message)
+            );
         }
     }
 
--- a/binary-pcs/src/lib.rs
+++ b/binary-pcs/src/lib.rs
@@ -1,0 +1,21 @@
+#![doc = include_str!("../README.md")]
+#![no_std]
+
+extern crate alloc;
+
+mod error;
+mod fold;
+mod params;
+mod pcs;
+mod proof;
+mod prover;
+#[cfg(test)]
+pub(crate) mod test_util;
+mod verifier;
+
+pub use error::BinaryPcsError;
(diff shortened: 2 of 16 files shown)


binary-pcs/src/prover.rs
```diff
@@ -0,0 +1,428 @@
...
+    let mut rounds = Vec::with_capacity(num_fold_rounds - 1);
+    for round in 0..num_fold_rounds {
+        let challenge =
+            sumcheck.compute_sumcheck_polynomials(&mut sumcheck_data, challenger, 1, 0, None);
+        let beta = challenge.as_slice()[0];
+        randomness.extend(&challenge);
+
+        codeword = fold_codeword(&codeword, beta);
+
+        if round + 1 < num_fold_rounds {
```
tcoratger, line 169: **[major, prover time + proof size]** This is the biggest lever in the crate, and I think it's being conflated with a different knob in the PR description.

The description says "there is no folding knob" and points at the head-collapse search. That's about the **message** side — how many variables get bound inside a committed row — and I agree it's forced: the sumcheck lockstep means every variable folds, one per round. But the FRI **commit** arity is a separate, entirely free knob, and this code fixes it at 1: a Merkle tree per fold round, `num_variables` of them.

Nothing about the layout, the binding order, or the lockstep identity changes if you commit every `k`-th round instead. The verifier opens a coset of `2^k` at committed round `c`, applies `fold_pair` `k` times locally with `betas[c..c + k]`, and compares against round `c + k`'s opening. `p3-fri` already has exactly this machinery — `log_arities`, `fold_row`, `extra_query_index_bits` (`fri/src/prover.rs:215-232, 303-328`).

For scale, binius64 runs the same construction — Gao-Mateer/Cantor `F_2`-subspace domain, the same `x^2 + x` recursion, the same unique-decoding regime, no OOD — and its proof-size dynamic program lands on **arity 4** (16-to-1 commit-to-commit) at B128 + SHA-256, at every shape it ships. It reports 340,128 B for a `2^24` message at rate `2^-4`; this crate reports 514,147 B for `2^20` at `2^-2`.

At `nv = 20` here the leaf-hashing alone goes from `2^22 + 2^21 + ... + 2^3 ~ 8.4M` leaves to `2^22 + 2^18 + 2^14 + 2^10 + 2^6 ~ 4.5M`, about 1.9x — and that is almost certainly where the 4.22 s `open` is. On proof size it is the usual arity trade (fewer Merkle paths, more coset symbols per query), which is why binius64 searches for the optimum rather than fixing it; but at arity 1 you are at the end of the curve where paths dominate, and the ~451 KB of Merkle nodes I get from back-of-envelope multiproof accounting matches your measured 514 KB almost exactly.

Not a blocker — a new crate is allowed to land at arity 1 with the schedule hard-coded. But I'd say so explicitly in the module docs, because "there is no folding knob" currently reads as if the commit arity were also forced, and it isn't.

binary-pcs/src/proof.rs
```diff
@@ -0,0 +1,58 @@
...
+#[serde(bound(
+    serialize = "MT::Commitment: Serialize, MT::MultiProof: Serialize",
+    deserialize = "MT::Commitment: Deserialize<'de>, MT::MultiProof: Deserialize<'de>"
+))]
+pub struct RoundProof<MT: Mmcs<BinaryField128>> {
+    /// Merkle root of this round's folded codeword.
+    pub commitment: MT::Commitment,
+    /// Opened rows, two per query — the low- then high-indexed symbol of that query's fold
+    /// pair at this round — in the order the query indices were sampled.
+    pub opened_values: Vec<Vec<BinaryField128>>,
```
tcoratger, line 23: **[minor, proof size]** This is the one place `p3-fri` is leaner, and it's worth stealing in a follow-up.

At round `r + 1` the verifier already *computes* one of the two symbols it reads here. In `verify_query_paths`:

```rust
let folded = fold_pair(position >> 1, beta, lo, hi);
let expected = round_values(r + 1)[2 * q + parity][0];
```

`expected` is fully determined by round `r`. `p3-fri` sends only `sibling_values` — `arity - 1` per round — and reconstructs the other slot from its own fold (`fri/src/verifier.rs:584`, `evals[index_in_group] = folded_eval;`).

The Merkle cost is unchanged either way: positions `2k` and `2k+1` are Merkle siblings, so authenticating one authenticates both. The whole difference is transmitted field data:

    136 queries * 19 intermediate rounds * 16 B = 41,344 B

against the 514,147 B you measured at `2^20` — 8.0%, and 8.3% at `2^16`. `expected_opens` would become `target_queries` rather than `2 * target_queries` for rounds >= 1, and the per-round `FoldMismatch` check folds into the Merkle check instead of standing alone.

Not a correctness issue, just free bytes.

binary-pcs/src/prover.rs
```diff
@@ -0,0 +1,428 @@
...
+            let (commitment, round_data) =
+                mmcs.commit_matrix(RowMajorMatrix::new(codeword.clone(), 1));
```
tcoratger, line 171: **[minor, allocation]** Two full copies of the codeword per proof that the borrow checker can be talked out of.

Line 154's `base_matrix.values.clone()` copies `2^(nv + rate)` symbols, and this `codeword.clone()` copies each folded round; the tail is geometric, so together it's `2 * domain_size` field elements — about 134 MiB of copy traffic and peak allocation at `nv = 20`.

The MMCS keeps the leaves, so the next round's input can be read back out of the prover data instead of cloned into it: fold from `mmcs.get_matrices(&rounds[round - 1].merkle_data)[0].values` (or from the base matrix at round 0), commit the *owned* `Vec` `fold_codeword` just returned, and push. The immutable borrow of `rounds` dies when `fold_codeword` returns, so the later `rounds.push` is fine — the only fiddly bit is that `merkle_data` is moved out at the end, so the base slice wants to be taken once up front rather than re-borrowed each iteration.

binary-pcs/src/verifier.rs
```diff
@@ -0,0 +1,565 @@
...
+pub(crate) fn sample_query_indices<Challenger, F>(
+    domain_size: usize,
+    num_queries: usize,
+    challenger: &mut Challenger,
+) -> Vec<usize>
+where
+    Challenger: FieldChallenger<F> + CanSampleUniformBits<F>,
+    F: Field,
+{
+    let bits = log2_strict_usize(domain_size);
```
tcoratger, line 47: **[minor, entropy accounting]** Bit 0 of `index` is never read by anything downstream.

- At round 0, `fold_pair_positions(index, 0) = [index & !1, (index & !1) + 1]`, and `fold_pair(position >> 1, ..)` uses `position >> 1 = index >> 1`.
- At round `r >= 1`, `index >> r` doesn't see bit 0 at all.

So `index` and `index ^ 1` are byte-for-byte the same query. Sampling `bits` and enforcing distinctness on the full index therefore guarantees distinctness over a space twice as large as the one the tests actually live in: two sampled indices can be siblings, costing two openings for one test.

Measured against the crate's own configs:

| config | domain | queries | expected sibling pairs |
|---|---|---|---|
| nv=8, rate 2^-2, sec 40 (this crate's test params) | 2^10 | 54 | 1.40 — 2.6% of queries |
| nv=16, rate 2^-2, sec 100 | 2^18 | 136 | 0.035 |
| nv=20, rate 2^-2, sec 100 | 2^22 | 136 | 0.002 |

Negligible where it matters, and I confirmed the duplicate case round-trips fine (at `nv <= 4` the target saturates the domain and every pair is opened twice — verifies cleanly). So this is bookkeeping, not a bug.

Sampling `bits - 1` and using the result directly as the pair index `j`, opening `[2j, 2j + 1]`, makes `num_queries` distinct indices mean `num_queries` distinct tests exactly, and removes the duplicate-openings case entirely.

Worth saying: you already do better than `p3-fri` here, which samples *with* replacement and documents the collision (`fri/src/prover.rs:112-119`). This just moves the distinctness to the granularity that the checks actually use.

binary-pcs/src/verifier.rs
```diff
@@ -0,0 +1,565 @@
...
+
+use crate::error::BinaryPcsError;
+use crate::fold::fold_pair;
+use crate::params::BinaryPcsConfig;
+use crate::proof::BinaryPcsProof;
+
+/// Samples `num_queries` distinct indices, uniformly at random, from `[0, domain_size)`.
+///
+/// Indices come from `sample_uniform_bits::<true>`, which rejection-samples internally,
+/// rather than from the low bits of a sampled field element: the latter biases each draw and
```
tcoratger, line 28: **[nit, doc precision]** True in general, but not for this field — and someone will read it as a statement about `BinaryField128`.

Over `GF(2^128)` a uniform field element *is* a uniform 128-bit string, so the low `k` bits of one are exactly uniform; there is no bias to avoid. `BinaryChallenger`'s own impl says as much (`binary-field/src/challenger.rs:171-188`): it reads 8 hash bytes and masks, with

> No field-element decomposition occurs, so no rejection band exists. The const generic is therefore inert: this function never errors and never resamples.

which also means `RESAMPLE = true` is inert here and the `.expect(...)` on line 54 is unreachable for a different reason than the one given.

Suggested framing: over a binary field either source is uniform; `sample_uniform_bits` is used because it is the API that *states* the guarantee, so swapping in a prime-field challenger later cannot silently reintroduce bias.

binary-pcs/src/verifier.rs
```diff
@@ -0,0 +1,565 @@
...
+            });
+        }
+    }
+    Ok(())
+}
+
+/// Wraps opened rows for [`Mmcs::verify_multi_batch`], which expects a `[query][matrix]`
+/// shape; every round here commits exactly one matrix, so each row gets a one-element outer
+/// slice.
+fn wrap_rows(opened_values: &[Vec<BinaryField128>]) -> Vec<Vec<&[BinaryField128]>> {
```
tcoratger, line 112: **[nit, allocation]** One `Vec` per opened row, once per round: `2 * 136` rows across 20 rounds is roughly 5,400 heap allocations per `verify`, against a 7.1 ms verify. Not where the time goes, but verify time is the number this crate will get quoted on, so it's cheap to reclaim — either a small-vec, or a `Mmcs::verify_multi_batch` shape that takes `&[&[F]]` for the one-matrix-per-round case.

binary-pcs/src/params.rs
```diff
@@ -0,0 +1,479 @@
...
+/// that no query count buys back. Both are bounded by a multiple of `1/|F|`:
+///
+/// - **The commit phase.** At fold arity 2 in the unique-decoding regime the per-round
+///   proximity error is `(arity - 1) * (n + 1) / |F|`, i.e. `(n + 1) / |F|` for a base codeword
+///   of `n` symbols. That is the bound `p3_security::fri::commit_phase_error_udr` states, which
+///   `the_field_security_agrees_with_p3_security` below checks this function against.
+/// - **The sumcheck.** Each of the `num_fold_rounds` rounds sends a degree-2 polynomial, so a
+///   forged round survives a uniform challenge with probability at most `2 / |F|`.
+///
+/// Their sum is `n + 1 + 2 * num_fold_rounds` over `|F|`. The sum cannot overflow: `try_new`
```
tcoratger, line 60: **[nit, doc]** Worth one sentence on *which* composition convention this sum is, because the two halves use different ones and a reader reconciling this against `p3-security` will get stuck.

`commit_phase_error_udr` is documented as the **per-round** error ("FRI commit-phase per-round error in UDR", `security/src/fri.rs:139`), and `p3-security` composes terms by taking the minimum bits over them (`security/src/report.rs:112`) — round-by-round, not a union bound. Here the commit term is charged once (`n + 1`, per-round / RBR) while the sumcheck is union-bounded (`2 * num_fold_rounds`). Both are defensible and the mixture is conservative, but it isn't stated.

Numerically it is moot — `n + 1 = 2^22` dwarfs `2 * 20 = 40`, so the cap is 105 either way — which is exactly why it should be written down now, before someone raises the arity and the two terms become comparable.

binary-pcs/src/params.rs
```diff
@@ -0,0 +1,479 @@
...
+    ///
+    /// Returns a [`BinaryPcsConfigError`] if the codeword length does not fit in a `usize`,
+    /// if the polynomial has no variables, if grinding exceeds what the challenger can witness
+    /// or the security budget, if the target exceeds what the field can deliver, or if the
+    /// derived query count is zero.
+    pub fn try_new(
+        num_variables: usize,
+        params: BinaryPcsParams,
+    ) -> Result<Self, BinaryPcsConfigError> {
+        if !matches!(REGIME, SecurityAssumption::UniqueDecoding) {
```
tcoratger, line 157: **[nit]** `REGIME` is a `const`, so this `matches!` is a compile-time constant and the branch is dead: `UnsupportedSoundnessRegime` is a public error variant no caller can observe and no test can construct.

The intent — guard the private core against a future caller — is right, but a module-scope `const { assert!(matches!(REGIME, SecurityAssumption::UniqueDecoding)) }` turns that future change into a *compile* error rather than a runtime path, which is what you actually want given the whole argument is that the other regimes are refuted rather than merely unsupported.

binary-pcs/src/fold.rs
```diff
@@ -0,0 +1,281 @@
...
+/// here the same value would otherwise be recomputed once per output symbol of the whole
+/// codeword.
+///
+/// # Panics
+///
+/// Panics if `codeword` has odd length. A codeword always has a power-of-two length, so an odd
+/// one means the caller has lost track of the round schedule.
+#[must_use]
+pub fn fold_codeword(codeword: &[BinaryField128], beta: BinaryField128) -> Vec<BinaryField128> {
+    assert!(
```
tcoratger, line 86: **[nit]** The doc four lines up says "A codeword always has a power-of-two length", and `gray_chain_steps` derives its `levels` bound from `next_power_of_two()`, but the assert only checks evenness.

I worked the non-power-of-two case through and the indexing is in fact safe — the largest `trailing_zeros` over `1..num_pairs-1` is `floor(log2(num_pairs - 1)) <= levels - 1` — so this is not a latent panic. But `fold_codeword` is `pub`, so asserting the precondition you document costs nothing:

```rust
assert!(
    codeword.is_empty() || codeword.len().is_power_of_two(),
    "codeword length must be a power of two"
);
```

binary-pcs/tests/end_to_end.rs
```diff
@@ -0,0 +1,661 @@
...
+type MyMmcs = MerkleTreeMmcs<F, u8, MyHash, MyCompress, 2, 32>;
+type MyChallenger = BinaryChallenger<F, HashChallenger<u8, Keccak256Hash, 32>>;
+type MyPcs = BinaryPcs<MyMmcs>;
+
+/// Default shape shared by every negative test: enough intermediate rounds
+/// (`num_fold_rounds - 1 == 7`) to truncate or permute, and `pow_bits > 0` so the grinding
+/// rejection has something to catch.
+const NUM_VARIABLES: usize = 8;
+const LOG_INV_RATE: usize = 2;
+const POW_BITS: usize = 4;
```
tcoratger, line 41: **[minor, test coverage]** `POW_BITS` is 4 here and 4 or 8 everywhere else in the crate, so `pow_bits = 0` is never exercised — and it is the one value where *both* sides take a different code path:

- `BinaryChallenger::grind` returns `F::ZERO` before touching the transcript (`binary-field/src/challenger.rs:214-217`);
- `GrindingChallenger::check_witness` returns `true` before touching it (`challenger/src/grinding_challenger.rs:43-45`).

They currently agree — I ran a `pow_bits = 0` round trip on this branch and it verifies — but nothing pins that. Drop either early return and the two challengers desync at query sampling, and *every honest proof* stops verifying. `try_new` accepts `pow_bits = 0` for any `security_level >= 1`, so it is a reachable config; one extra line in `small_configurations_round_trip`'s sweep covers it.

binary-pcs/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+pub use prover::BinaryPcsProverData;
+
+/// The stacked-layout binding mode this scheme commits in.
+///
+/// Fixed rather than chosen by the caller: the codeword fold merges *adjacent pairs*, which
+/// only suffix-order binding matches. Prefix binding merges halves, and driving it through the
+/// same fold does not stay in lockstep — see `prover.rs`'s
+/// `prefix_layout_does_not_stay_in_lockstep`, which pins that as a property of the two binding
+/// orders rather than a tuning detail.
+pub(crate) type PcsLayout = SuffixProver<BinaryField128, BinaryField128>;
```
tcoratger, line 33: **[minor, docs]** Two caller obligations the crate docs don't state, both worth a line here or in the README:

1. **This is not hiding.** The final codeword travels in the clear and the query openings are raw codeword symbols. Nothing claims ZK, but "polynomial commitment scheme" reads as hiding to a lot of people, and this crate will get reached for by someone who assumes it.

2. **`BinaryPcs::new` takes any `MT: Mmcs`, and `security_level` doesn't know about it.** `BinaryPcsConfig` charges the field's width and the query count but never the commitment's collision resistance, so a caller handing in a 16-byte-digest MMCS still gets a config reporting 100 bits. `p3-security` has a `COLLISION_LABEL` term for exactly this. The same gap exists in `p3-fri` and `p3-whir` so it isn't a regression — but this is the crate whose `try_new` *promises* a number, so it's the one that should say what it assumes.

tcoratger (approved):
Read this end to end — every file, plus `p3-security`'s `fri.rs`/`report.rs`, `p3-fri`'s verifier and prover, `p3-sumcheck`'s `Layout::commit` and `eval_constraints_poly`, and `BinaryChallenger`'s grind/sample paths — and then went looking for a soundness break. I did not find one.

The things I checked hardest, and what I concluded:

**Index arithmetic in the fold chain.** `fold_pair_positions(index, r) = [(index >> r) & !1, +1]`, `fold_pair(position >> 1, ..)` opens at `domain_point(even)`, and the successor slot is picked by `parity = (position >> 1) & 1`. That is consistent at every round, and the final read `final_codeword[index >> num_fold_rounds]` is in range because `domain_size = 2^(nv + rate)` and `num_fold_rounds = nv`. No off-by-one.

**`gray_chain_steps`.** `domain_point` is `F_2`-linear, so `domain_point(2j) + domain_point(2(j+1)) = domain_point(2 * (j XOR (j+1)))`, and `j XOR (j+1)` is exactly bits `0..=level` for `level = (j+1).trailing_zeros()`. `steps[level] = sum_{k=1}^{level+1} cantor_basis(k)` is right, and the `levels` bound holds for non-power-of-two `num_pairs` too (`max tz over 1..num_pairs-1 = floor(log2(num_pairs - 1)) <= levels - 1`). The cross-block test at 4096 symbols does cover the two-task path.

**Panics on adversarial input.** I traced every index into `proof` in `verify_opening` and `verify_query_paths` back to a guard: `polynomial_evaluations()[r]` behind the `num_rounds` check, `rounds[r]` and `final_codeword[..]` behind `check_round_and_final_lengths`, the opened rows behind `check_round_shape`'s count *and* width checks, `evals[i]` behind `OpeningBatchCountMismatch`. A malformed proof is a typed rejection, never an abort. That is genuinely hard to get right and you got it right.

**Fiat–Shamir order.** Prover `commit` (root absorbed inside `Layout::commit`) -> `eval` per claim -> `alpha` -> per round {poly, beta, round commitment} -> grind -> indices. Verifier replays exactly that, and `a_fresh_verifier_challenger_samples_the_same_query_indices` rebuilds it from an empty challenger rather than cloning the prover's — which is the version of that test that can actually fail. Interleaving `verify_rounds` one round at a time instead of one call for all rounds is the right call and the doc explains why.

**The regime argument.** Restricting to unique decoding and not re-exporting `SecurityAssumption` — so the refuted capacity bound is *unreachable* rather than runtime-rejected — is the right shape for this, and pinning `field_security_bits` against `commit_phase_error_udr` so the two cannot drift is the kind of test that pays for itself in two years. Paying ~136 queries for it is honest.

Two things worth calling out as better than the surrounding code: sampling query indices **without replacement** (`p3-fri` samples with replacement and just documents the 0.5% birthday collision at `fri/src/prover.rs:112-119`), and negative tests that every assert a *named* variant instead of `is_err()`.

Nothing below is a blocker. One major (the FRI commit arity, first comment — it's a deferred optimisation rather than a defect, but it's the biggest number in the crate and I think the PR description accidentally argues it away), then minors and nits on proof size, allocation, and doc precision. All inline. I'm approving and merging; I'll open a follow-up PR with the fixes so none of it blocks Phase 3 landing.

One cross-check that came out clean and is worth recording: your query count reconciles exactly with binius64's. Their `calculate_n_test_queries` is `ceil(t / -log2((1 + rho) / 2))`, identical to `p3_security`'s `log_1_delta` for UDR, and they get 148 at rate `2^-2` / 100 bits against your 136 — the whole difference is the 8 PoW bits you charge and they don't (`Grinding::NONE`). Your `field_security_bits` also lands within a couple of bits of their independently derived correlated-agreement ceiling (they get 105 bits at a `2^24` message, you get 105 at a `2^22` domain). Two independent implementations agreeing on both terms is about as much external validation as this can get.

One forward-looking note, separately: I read binius64, `succinctlabs/flock` and `leanEthereum/leanVM-b` to see whether a WHIR-shaped alternative belongs next to this. Summary in a follow-up comment — short version is that the unique-decoding cost you're paying here is real and a WHIR/STIR-style scheme is the interesting lever, but the same characteristic-2 subspace counterexamples that rule out capacity here apply to WHIR's conjectured regime too, so it is not a free win and deserves its own PR rather than a bolt-on.

Really nice work. The docs in this crate are the best in the repo.

tcoratger:
Only super minor so in order not to block the work, just merged and we can (if we want) solve the minors in some followup PRs

tcoratger:
Follow-up to my review, on the WHIR question. I read binius64, [succinctlabs/flock](https://github.com/succinctlabs/flock) and [leanEthereum/leanVM-b](https://github.com/leanEthereum/leanVM-b) to see whether a WHIR-shaped scheme belongs next to this one. Short answer: yes, and the case is stronger than I expected — but the more interesting finding is about *this* crate's regime, not about WHIR.

I've split what I verified in this tree from what came out of the literature survey, because the second half is worth your own check before you act on it.

## Verified here

**`p3-whir` cannot be pointed at a binary field.** `whir/src/pcs/adapter.rs:35` and `whir/src/pcs/committer/writer.rs:33-35` bound `F: TwoAdicField`, `EF: ExtensionField<F> + TwoAdicField`, `Dft: TwoAdicSubgroupDft<F>`. `BinaryField128` has no `TwoAdicField` impl and cannot get one. So "add WHIR as an alternative next to BaseFold" is a new crate, not a generalisation of the existing one — worth knowing before anyone scopes it.

**The README's claim about `p3-security` isn't right.** It says the Johnson bound is excluded because "`p3-security` documents it as resting on a correlated-agreement conjecture." `p3-security` says the opposite. `security/src/assumption.rs:45-50`:

```rust
/// Johnson bound at `δ = 1 − √ρ − η`, with `η = √ρ / 20`. Requires
/// mutual correlated agreement up to the Johnson bound.
///
/// The proximity-gap error uses \[BCSS25\] Theorem 1.5:
```

"Requires … up to the Johnson bound", with a *theorem* cited. The conjecture language is reserved for the variant below it — `CapacityBound`: "Requires **conjecturing** capacity-rate list decodability". So the crate distinguishes the two exactly as you'd want, and the README attributes the capacity variant's caveat to the Johnson one.

**What that costs, at your own parameters** (`security_level - pow_bits = 92`, via `p3_security`'s `log_1_delta`):

| rate | UDR queries | Johnson queries | |
|---|---|---|---|
| 2^-1 | 222 | 215 | 3% fewer |
| **2^-2** | **136** | **99** | **27% fewer** |
| 2^-3 | 111 | 65 | 41% fewer |
| 2^-4 | 101 | 48 | 52% fewer |

Since proof size here is dominated by Merkle paths, and paths scale with the query count, that is close to a linear saving. Together with the commit arity from my review comment, it is most of the gap to binius64's numbers.

## From the literature survey — please check these yourself

I have not read these papers directly; this is a research pass I'd treat as a set of pointers, not as settled.

**Mutual correlated agreement up to Johnson looks like it is now an unconditional theorem for Reed–Solomon over an arbitrary field and an arbitrary domain of definition** — BCHKS25 Thm 4.6, and independently Haböck, *A note on mutual correlated agreement for Reed-Solomon codes*, [eprint 2025/2110](https://eprint.iacr.org/2025/2110) Thm 2, whose proof route is a Guruswami–Sudan decoder over `F_q(Z)` and so is characteristic-agnostic. If that reading holds, the Cantor domain is covered and Johnson is not a conjecture here at all. Note BCHKS still cites Haböck as "personal communication"; it is public now.

**Your params.rs:29 / README may misattribute BCHKS Thm 1.6.** The survey's reading is that Thm 1.6 refutes `n^τ`-bounded *proximity gaps* (their Conjecture 1.2), and that the list-decoding statement is a separate Thm 1.9 holding over arbitrary fields — so "capacity-rate list decodability is refuted over characteristic 2 with `F_2`-subspace domains" is the wrong object, and it is *some* subspace domain rather than all of them. The counterexamples also need `n` near `sqrt(q)`, i.e. `2^64`, far past anything this crate commits at.

**The Proximity Prize companion paper ([eprint 2026/680](https://eprint.iacr.org/2026/680)) apparently frames characteristic 2 as the *settled* side of this line**, not the uncertain one — the open question being whether prime fields over smooth domains can go past Johnson. If so, the README's framing is inverted.

**Flock claims unconditional 100-bit round-by-round soundness at the Johnson bound over `F_2`-affine-coset domains in characteristic 2** — "*(unconditionally – without relying on any unproven conjectures) … proximity gaps up to the Johnson bound [BCI+23, BCH+25] (in combination with a small amount of grinding)*". That is exactly the regime this crate excludes by choice.

**leanVM-b apparently runs WHIR over binary fields today** — `crates/pcs/src/whir.rs`, `K = GF(2^64)`, `E = F192`, an additive NTT, with a written RBR proof in `doc/leanvm/body/b-polynomial-commitment-scheme.tex` and a frozen query table pinned by a test. The mechanism, which is the part worth stealing regardless: it is an *interleaved* instantiation that folds across the interleaving axis and re-encodes each level on a fresh domain, so no fold map ever acts on the domain and WHIR's `x -> x^2` — undefined in characteristic 2 — never has to be ported. The survey reports building it and measuring 227 KB at rate 2^-1 / 0.71 ms verify, and 102 KB at rate 2^-4 / 0.38 ms verify, both at 2^20. I have not reproduced those.

## One thing that is *not* an objection to my arity comment

FRI-Binius ([2024/504](https://eprint.iacr.org/2024/504)) records that "the incompatibility of BaseFold—as written—with higher-arity folding has been noted by … WHIR [ACFY25, §1.1]", and it is easy to read that as ruling out my earlier suggestion. It doesn't. What is incompatible is folding the codeword by `2^k` in one step against a *single* challenge, which has no counterpart in binding `k` multilinear variables. Composing `k` ordinary 2-to-1 folds against `k` distinct sumcheck challenges and simply not committing the intermediates is a different thing, and it leaves the lockstep identity untouched: the verifier opens a coset of `2^k`, applies the pair fold `k` times with `betas[c..c + k]`, and compares against round `c + k`. That is the change I meant.

## What I'd suggest

1. **Re-check the regime before building anything.** If Johnson is unconditional over an arbitrary domain, this crate is paying 27–52% more queries than it needs to for a reason that has expired, and the fix is a parameter change plus an OOD-binding round — far cheaper than a second PCS. It also means the README's and `params.rs`' soundness prose want an edit either way.
2. **Then the commit arity** (my review comment), which is independent and compounds with it.
3. **WHIR as a sibling crate, last.** It is a real option and leanVM-b appears to show the char-2 route exists, but it is new code rather than a reuse of `p3-whir`, and steps 1 and 2 recover a large fraction of the same proof size at a fraction of the cost.

Happy to be wrong on the second half — the first half is checkable in this tree in five minutes.

tcoratger:
Addendum — a second research pass turned up four things that sharpen the comment above. Same caveat: literature-sourced, not verified by me.

**Two citation fixes, if you edit the soundness prose.**

- The `BGKS20 App. B` pointer in `params.rs` and the README looks misaimed. BGKS20's relevant result is §3.1 / Lemma 3.3 — proximity-gap tightness at the 1.5-Johnson radius over the *full-field* domain — and BGKS20 itself *conjectures* capacity list-decodability (Conjecture 2.3), constructed to be consistent with the char-2 subspace-polynomial limits. The "Appendix B" is likely `[BCGM25, Appendix B]`, which is about unique-decoding correlated agreement.
- If you do move to Johnson, the citation to use is Flock's Theorem 8, which is stated as *"MCA up to Johnson bound, adapted from [BCH+25, Theorem 4.6]"* with no characteristic or domain restriction.

**A fair counterweight I should have included.** BinarySpartan ([eprint 2026/1656](https://eprint.iacr.org/2026/1656), Setty) makes the same call you did, deliberately: *"Our soundness accounting assumes only the unique-decoding bound, not the wider Johnson radius … we run no proof of work"*, at `2^-102.6` over `F_2^128`. So unique-decoding-over-binary is an actively chosen position with company, not a gap. My comment above reads as if the regime were simply money left on the table; the honest version is that it is a defensible choice whose *stated reason* — a `p3-security` conjecture caveat that turns out to attach to the capacity variant — is the part that needs updating, whatever you then decide.

**A third design point for the arity question.** FRI-Binius solves the higher-arity problem with *oracle-skipping* rather than WHIR's fold or leanVM-b's interleaved re-encode. Worth reading before picking one.

**Don't build parameters on Crites–Stewart's Conjecture 6 literally.** Its `H_p(δ)` form at `p = 2` is *more conservative than what BCHKS Thm 4.6 already proves* — around `δ < 0.356` against a proven Johnson `δ < 0.75` at `ρ = 2^-4`. The authors hedge explicitly that for characteristic 2 *"we cannot confidently give one."* BCHKS Thm 4.6 and Haböck Thm 2 are the safe floor.

**One number I want to flag as mine, not theirs.** WHIR's own tables suggest roughly 2x per regime step (at `ρ = 2^-1`, `2^24`: capacity 156 KiB, Johnson 299, unique decoding 621). My table above says only 3% at that same rate. Both can be right — WHIR's regimes also move the arity schedule and the OOD structure, not just the query count. The table above is computed from `p3_security`'s own `log_1_delta` at *this crate's* parameters and is the number that applies here; treat the 2x as an upper bound on what a full protocol reshape could buy, not as what a parameter flip would.

tcoratger:
We may also consider WHIR in the future as another PCS which as proven proof size reduction in both Flock and leanVM-b.

PR #2033, opened 2026-09-05
fix(fri)!: derive the folding schedule instead of accepting the prover's

tcoratger:
## What this changes

The FRI folding schedule — which arity each commit round folds by — is now derived on both sides instead of being carried in the proof.

## The problem

The schedule is decided entirely by the committed heights and the parameters. Nothing about it is a choice.

The prover recomputed it round by round inside the fold loop. The verifier read it out of the proof and checked two things: that each arity was within the cap, and that the arities summed to the committed height.

A sum is a weak thing to check. Two different schedules can share one:

```
committed height 2^5, final height 2^1, arity cap 3

honest:  [3, 1]   folds 5 -> 2 -> 1
forged:  [1, 3]   folds 5 -> 4 -> 1
```

Both fold the same total distance, both stay within the cap, both have the same round count. So every guard the verifier ran passed either way.

## This was reachable, and the test now proves it

Demonstrated end-to-end, not argued: `forged_fold_schedule_rejected` builds a proof that folded on `[1, 3]` with every commitment, sibling row, opening proof and transcript value consistent with it — what a malicious prover would emit, not a field tamper.

```
with the derivation:     Err(FoldScheduleMismatch { expected: [3, 1], got: [1, 3] })
with it commented out:   Ok(())          <- the proof is accepted
```

So a prover could commit on a folding schedule the configuration never specifies and have it accepted. `prove_fri_with_schedule` is the test-only seam that makes this provable: an honest prover passes `None`.

The security model in `p3-security` derives its layer count from the arity cap alone (`security/src/fri.rs:147`), so it models a uniform max-arity schedule and does not describe a prover-chosen one in either direction. What this change delivers is that the protocol the verifier executes is the protocol the configuration specifies. I have not shown that a particular reordering is cryptographically weaker.

## The idea

Compute the schedule once, from the inputs, in a place both sides can call.

```rust
fold_schedule(input_log_heights, log_final_height, max_log_arity) -> Vec<usize>
```

A round's arity is capped by three things, all known before folding starts: the distance left to the final height, the distance down to where the next input joins, and the configured maximum.

The prover iterates that function's output rather than recomputing per round, so its schedule is the function's output by construction. The verifier derives the same schedule from the committed heights and rejects a proof that folded differently.

## Two guards no longer fail open

Both derived checks were skipped when the instance committed to no matrices, which left `log_arities` fully proof-controlled on that path. An instance with nothing committed is now rejected up front, next to the zero-queries guard, so both comparisons are unconditional and share one derivation of the committed heights.

`MissingInitialReducedOpening` becomes defensive only as a result: its former path is closed, and with at least one committed matrix every query gets at least one reduced opening. The check is kept, since I did not prove it unreachable.

## Circle had the same hole

`p3-circle`'s verifier read the per-round arity out of the proof and pinned only the sum, and it gated the rest on a bare `assert_eq!(log_arity, 1)` deep in the query loop — a panic on proof-controlled data, reachable for any configuration with `max_log_arity > 1`, where it also aborted on *honest* proofs.

Circle folding halves the domain and nothing else, so importing `fold_schedule` would hand it a schedule it cannot fold. The real constraint is arity one, and that is now checked on entry. Every per-round arity is then forced to one, so the pinned height sum determines the schedule uniquely, and the assert becomes a debug assertion.

## One robustness note

`fold_schedule` returns an empty schedule when nothing sits above the final height, rather than asserting. The verifier derives it from untrusted heights, so it must not panic there.

## Test plan

- [x] `cargo test -p p3-fri` — 77 pass across all targets
- [x] `cargo test -p p3-circle` — 45 pass
- [x] `cargo clippy -p p3-fri -p p3-circle --all-targets` — clean
- New in FRI: four unit tests for the schedule including an exhaustive sweep over caps, final heights and input pairs; the forged-schedule rejection above; an honest non-uniform schedule accepted, so the forgery test cannot pass for the wrong reason; the empty-instance rejection.
- New in circle: a cap above one is refused on entry.

## Breaking changes

`FriError` gains `NoCommittedMatrices` and `UnsupportedFoldingCap`. A circle configuration with `max_log_arity > 1` is now rejected rather than aborting mid-verification.

### DIFF
--- a/fri/src/config.rs
+++ b/fri/src/config.rs
@@ -206,9 +206,145 @@ pub fn compute_log_arity_for_round(
     max_fold.min(max_log_arity)
 }
 
+/// Derive the whole folding schedule before any folding happens.
+///
+/// # Overview
+///
+/// A round's arity is capped by three things:
+///
+/// - distance left to the final height,
+/// - distance down to where the next input joins,
+/// - the configured maximum.
+///
+/// All three are known before folding starts.
+/// So the schedule is known too, and neither side reads it from a proof.
+///
+/// # Arguments
+///
+/// - `input_log_heights`: log-heights of the folding inputs, strictly decreasing.
+/// - `log_final_height`: log-height at which folding stops.
+/// - `max_log_arity`: largest arity any single round may use.
+///
+/// # Returns
+///
+/// One log-arity per commit round, in round order.
+///
+/// Empty when nothing sits above the final height.
+/// A verifier derives this from untrusted heights, so it returns rather than panics.
+///
+/// # Panics
+///
+/// When the input heights are not strictly decreasing.
+#[must_use]
+pub fn fold_schedule(
+    input_log_heights: &[usize],
+    log_final_height: usize,
+    max_log_arity: usize,
+) -> Vec<usize> {
+    assert!(
+        input_log_heights.windows(2).all(|pair| pair[0] > pair[1]),
+        "input log-heights must be strictly decreasing",
+    );
+
+    // Folding starts at the tallest input and stops at the final height.
+    let Some(&tallest) = input_log_heights.first() else {
+        return Vec::new();
+    };
+    if tallest <= log_final_height {
+        return Vec::new();
+    }
+    let mut log_height = tallest;
+
+    // Index of the next input still waiting to be rolled in.
+    let mut next_input = 1;
+    let mut schedule = Vec::new();
+
+    while log_height > log_final_height {
+        // Cap by the final height, the next input, and the configured maximum.
+        let log_arity = compute_log_arity_for_round(
+            log_height,
+            input_log_heights.get(next_input).copied(),
+            log_final_height,
+            max_log_arity,
+        );
+        schedule.push(log_arity);
+        log_height -= log_arity;
+
+        // An input whose height the fold just reached is rolled in here.
+        // The next round then folds the combined codeword.
+        if input_log_heights.get(next_input) == Some(&log_height) {
+            next_input += 1;
+        }
+    }
+
+    schedule
+}
+
 #[cfg(test)]
 mod tests {
+    use alloc::vec;
+
     use super::*;
+
+    #[test]
+    fn schedule_folds_the_whole_way_down() {
+        // Invariant: the arities sum to the distance the codeword has to travel.
+        //
+        // Fixture state: one input at height 2^10, folding down to 2^2.
+        //
+        //     10 -> 2 is 8 levels, in steps of at most 3
+        //     8 = 3 + 3 + 2
+        assert_eq!(fold_schedule(&[10], 2, 3), vec![3, 3, 2]);
+
+        // Binary folding takes one level per round.
+        assert_eq!(fold_schedule(&[10], 2, 1), vec![1; 8]);
+    }
+
+    #[test]
+    fn schedule_pauses_where_an_input_joins() {
+        // Invariant: a round never folds past the height of the next input.
+        //
+        // An input joins only once the codeword reaches its height.
+        // A round that overshot would leave it nowhere to join.
+        //
+        // Fixture state: inputs at 2^10 and 2^6, final height 2^2, max arity 3.
+        //
+        //     10 --3--> 7 --1--> 6   <- input joins here
+        //      6 --3--> 3 --1--> 2   <- final height
+        assert_eq!(fold_schedule(&[10, 6], 2, 3), vec![3, 1, 3, 1]);
+    }
+
+    #[test]
+    fn schedule_is_empty_without_inputs() {
+        // Boundary: nothing to fold means no commit rounds.
+        assert_eq!(fold_schedule(&[], 2, 3), Vec::<usize>::new());
+    }
+
+    #[test]
+    fn every_schedule_lands_exactly_on_the_final_height() {
+        // Invariant: folding never overshoots or stops short.
+        //
+        // Overshooting loses the evaluations the query phase needs.
+        for max_log_arity in 1..=4 {
+            for log_final_height in 0..4 {
+                for tall in (log_final_height + 1)..12 {
+                    // A second input somewhere strictly between the two ends.
+                    for short in (log_final_height + 1)..tall {
+                        let schedule =
+                            fold_schedule(&[tall, short], log_final_height, max_log_arity);
+                        assert_eq!(
+                            schedule.iter().sum::<usize>(),
+                            tall - log_final_height,
+                            "max_log_arity={max_log_arity} final={log_final_height} \
+                             inputs=[{tall}, {short}]",
+                        );
+                        // No round may exceed the configured maximum.
+                        assert!(schedule.iter().all(|&a| a <= max_log_arity && a > 0));
+                    }
+                }
+            }
+        }
+    }
 
     /// Pins the field-by-field mapping in [`FriParameters::security_regime`].
     /// Distinct values per field catch a mis-wired mapping (e.g. swapping the
--- a/fri/src/verifier.rs
+++ b/fri/src/verifier.rs
@@ -16,7 +16,7 @@ use thiserror::Error;
 
 use crate::{
     BatchMultiOpening, CommitPhaseMultiStep, CommitmentWithOpeningPoints, FriFoldingStrategy,
-    FriParameters, FriProof,
+    FriParameters, FriProof, fold_schedule,
 };
 
 #[derive(Debug, Error)]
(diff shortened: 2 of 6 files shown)


fri/src/verifier.rs
```diff
@@ -284,6 +294,35 @@ where
...
+        .flat_map(|(_, mats)| {
+            mats.iter()
+                .map(|(domain, _)| log2_strict_usize(domain.size()) + params.log_blowup)
+        })
+        .collect();
+    // Folding sees one input per distinct height, tallest first.
+    input_log_heights.sort_unstable_by(|a, b| b.cmp(a));
+    input_log_heights.dedup();
+
+    if !input_log_heights.is_empty() {
```
Nashtare: **MEDIUM — the derived-schedule check is skipped entirely when there are no committed matrices.**

On the `else` branch neither the round count nor any per-round arity is constrained by anything derived: `log_arities` stays fully proof-controlled, bounded only by `checked_log_arity` and the two-adicity guard.

I traced the fallback and it does **not** admit an attack today — with no matrices, `open_inputs` returns one empty `FriOpenings` per query and `fold_query` bails at `MissingInitialReducedOpening` (verifier.rs:600). But that means the soundness of this path rests on a check ~300 lines away, in a different function, written for a different purpose. `GlobalMaxHeightMismatch` (verifier.rs:288, `if let Some(expected) = ...`) has the same shape, so both "derived" guards fail open together on the same input.

Suggested fix: reject an instance with no committed matrices up front, next to the existing `ZeroQueries` guard at verifier.rs:210, and make the schedule comparison unconditional.

fri/src/config.rs
```diff
@@ -206,10 +206,146 @@ pub fn compute_log_arity_for_round(
...
+/// One log-arity per commit round, in round order.
+///
+/// Empty when nothing sits above the final height.
+/// A verifier derives this from untrusted heights, so it returns rather than panics.
+///
+/// # Panics
+///
+/// When the input heights are not strictly decreasing.
+#[must_use]
+pub fn fold_schedule(
```
Nashtare, line 239: **MEDIUM — `p3-circle`'s FRI verifier still has the hole this PR closes; it should call `fold_schedule` too.**

`circle::verifier::verify_fri` reads `log_arity` per round out of `proof.commit_phase_openings` (`circle/src/verifier.rs:97-109`) and derives `log_max_height` from their sum (`circle/src/verifier.rs:152`), cross-checked only against that *sum* at `circle/src/pcs.rs:637`. That is byte-for-byte the pre-PR `p3-fri` structure this commit just fixed.

It is not exploitable for schedule forgery today, but only by second-order accident: `circle/src/verifier.rs:366` asserts `log_arity == 1` for every round, so a pinned sum determines the schedule uniquely. That same line is a `panic!` on proof-controlled data, and it is reachable for any config with `max_log_arity > 1` — e.g. `FriParameters::new_benchmark_high_arity`, which sets 3 — where it also aborts on *honest* proofs.

Suggested fix: call `fold_schedule` from the circle verifier as well. Minimally, turn `circle/src/verifier.rs:366` into an `InvalidLogArity`-style error return and state the arity-1 restriction on `FriParameters`.

(Flagging here since `circle/` is outside this PR's diff — happy for it to be a follow-up, but the fix is incomplete across the repo until it lands.)

fri/src/verifier.rs
```diff
@@ -1590,6 +1640,56 @@ mod tests {
...
+        let mut proof = f.proof.clone();
+
+        let honest: Vec<usize> = proof
+            .commit_phase_openings
+            .iter()
+            .map(|opening| opening.log_arity as usize)
+            .collect();
+        assert_eq!(honest, vec![3, 1], "fixture must fold non-uniformly");
+
+        proof.commit_phase_openings[0].log_arity = 1;
```
Nashtare, line 1830: **MEDIUM — this negative test would still pass with the fix reverted.**

`reordered_fold_schedule_rejected` flips only the `log_arity` bytes, leaving `sibling_values` at the honest lengths (7 and 1). That proof is *also* rejected by the pre-existing `SiblingValuesLengthMismatch` check in `fold_query` (verifier.rs:566) — so revert the schedule derivation and the test still passes, just with a different error variant. It pins the new check's ordering, but it does not encode the regression.

For what it's worth, I confirmed the underlying vulnerability is real and closed by patching the *prover* to emit a fully self-consistent forged schedule (`[1, 3]` instead of the honest `[3, 1]` at log-height 5, `log_final_height = 1`, `max_log_arity = 3`) — every commitment, sibling row, MMCS proof and transcript value internally consistent, i.e. what a malicious prover would actually produce rather than a field tamper. Base `6e0ddf64` **accepts** it; this head **rejects** it with `FoldScheduleMismatch { expected: [3,1], got: [1,3] }`.

Suggested fix: expose a `#[cfg(test)]` seam on `commit_phase`'s schedule (or a test-only `commit_phase_with_schedule`) and assert rejection of that forged-but-consistent proof. That is the test that would fail if this check were removed.

Nashtare (approved):

Nashtare:
Would be nice to update this PR to also include the fix to `p3-circle` which suffers the same issue (only gated via a runtime panic with an `assert` at the moment)

fri/src/verifier.rs
tcoratger: Fixed as suggested. An instance that commits to no matrices is now rejected up front next to `ZeroQueries` (`FriError::NoCommittedMatrices`), and both derived guards are unconditional.

The height cross-check lost its `if let Some(..)` for the same reason, and the two guards now share one derivation of the committed heights instead of computing them twice.

Two tests moved with the guard, which is worth noting since it changes what they cover:

- `global_max_height_exceeds_two_adicity` reached its guard through the empty-instance path. It now inflates the schedule with commitments present, which is the reachable route: the sum is checked against the two-adicity before it is used, so the guard still fires first.
- `missing_initial_reduced_opening` *was* the empty-instance case, so it becomes `instance_with_nothing_committed_rejected` and pins the new rejection.

One consequence I could not fully resolve: `MissingInitialReducedOpening` is now defensive only. Its former test path is closed, and with at least one committed matrix every query gets at least one reduced opening, so I could not find another route to it. I left the check rather than delete it, since I have not proven it unreachable — say the word if you would rather it went.

fri/src/config.rs
tcoratger: Fixed in this PR rather than deferred, per your follow-up.

The fix is not `fold_schedule` though, and I think the reason matters: circle folding halves the domain and nothing else, so a derived non-uniform schedule is one circle *cannot fold*. Importing the derivation would hand the verifier a schedule it has no code path for.

What circle's constraint actually is, is arity one. So that is now checked on entry:

```rust
if params.max_log_arity != 1 {
    return Err(FriError::UnsupportedFoldingCap { max_log_arity: params.max_log_arity });
}
```

That closes the schedule hole by construction: `checked_log_arity(1)` then forces every per-round arity to one, so the height sum pinned at `pcs.rs:637` determines the schedule uniquely. No second derivation needed.

It also fixes the panic you flagged. `verifier.rs:366` becomes a `debug_assert!`, reachable only if the entry guard were removed, and a configuration like `new_benchmark_high_arity` now gets a typed error on entry instead of aborting mid-verification — on honest proofs as well as forged ones, which was the worse half of that bug.

New test `reject_folding_cap_above_two` in `circle/src/pcs.rs` takes a valid proof, sets the verifier's cap to 3, and asserts the typed error.

fri/src/verifier.rs
tcoratger: You are right, and thank you for building the forged proof — that saved me from shipping a test that pins nothing.

`prove_fri_with_schedule` is the seam: an honest prover passes `None`, a test passes `Some` and gets a proof where every commitment, sibling row, opening proof and transcript value agrees with the forgery.

`forged_fold_schedule_rejected` builds exactly your case — `[1, 3]` instead of the derived `[3, 1]` at log-height 5, final height 1, cap 3 — and I checked it the way you described:

```
with the derivation:     Err(FoldScheduleMismatch { expected: [3, 1], got: [1, 3] })
with it commented out:   Ok(())          <- the proof is accepted
```

So the test fails if the check is removed, which the old one did not. `honest_high_arity_proof_accepted` sits next to it so the forgery test cannot pass for the wrong reason, and `reordered_fold_schedule_rejected` stays as the cheap ordering check with its weakness now stated in the comment.

This also means the PR description understated the finding. I had written that I removed prover freedom the security model never accounted for but had not constructed a weaker schedule. There is now an end-to-end demonstration that base accepts a proof folded on a schedule the configuration never specifies, so I have rewritten that section rather than leave it soft.

The seam is a `#[cfg(test)]` field on `TwoAdicFriPcs` plus a `pub(crate)` prover entry point. Happy to move it behind a feature or restructure if you would rather the production struct stayed untouched.

PR #2034, opened 2026-09-05
fix(binary-pcs): query at pair granularity, fold without copying

tcoratger:
Follow-up to the review on #2032. Everything here is a minor from that review; the two substantive items are deliberately left out and named at the bottom.

The query phase's encoding changes, so proofs are not interchangeable across this commit. The crate is one commit old and unreleased, so no stored proof exists.

This is labelled `fix` rather than `refactor`: the old sampler was a small over-claim, not just waste. See below.

## Query sampling draws a pair index, not a symbol index

Every check in `verify_query_paths` goes through `index >> 1`:

- round `0` opens the pair `[index & !1, (index & !1) + 1]` and folds it at `domain_point(index & !1)`;
- round `r >= 1` sees only `index >> r`.

So bit `0` selected nothing. `index` and `index ^ 1` were byte-for-byte the same test, and two sampled indices that happened to be siblings named one test while costing two openings — which also meant `num_queries` distinct *indices* was not `num_queries` distinct *tests*, the thing the query bound is actually stated over.

Drawing `log_domain_size - 1` bits and doubling closes that. Expected wasted queries, before:

| config | domain | queries | expected sibling pairs |
|---|---|---|---|
| nv=8, rate 2^-2, sec 40 — this crate's test params | 2^10 | 54 | 1.40, i.e. 2.6% |
| nv=16, rate 2^-2, sec 100 | 2^18 | 136 | 0.035 |
| nv=20, rate 2^-2, sec 100 | 2^22 | 136 | 0.002 |

The old code enforced distinct *symbols* and charged the query bound for each. But `fold_pair_positions` masks bit 0 away at every round, so two symbols that are fold siblings run the same test — paid for twice and counted twice. Sampling `num_queries` distinct *pairs* without replacement is at least as strong as that many independent pairs, so the new scheme dominates both the old code and the bound it was priced against.

Expected sibling collisions, and what they cost of the claimed 36 bits:

| config | domain | queries | expected collisions | bits lost |
|---|---|---|---|---|
| nv=8, rate 2^-2 — this crate's test params | 2^10 | 54 | 1.4 | ~0.3 to 1 |
| nv=20, rate 2^-2, sec 100 | 2^22 | 136 | 0.002 | negligible |

Small, but a real over-claim rather than accounting. `BTreeSet` also replaces the linear `Vec::contains` scan, which the small-domain configurations (where the target saturates the pair count) drive into a coupon-collector tail.

## `fold_rounds` folds without copying

It cloned the base codeword once and every folded round once more — `2 * domain_size` element copies, about 134 MiB at `num_variables = 20`. The MMCS already owns every committed codeword and `fold_codeword` allocates its own output, so each round can fold straight out of the previous round's Merkle leaves; the borrow ends before `rounds` is pushed to.

Measured on `open/16`, `--features parallel`, `-C target-feature=+pclmulqdq`, against the merged baseline in a parallel worktree:

```
open/16   baseline   [45.6 ms  52.9 ms  57.6 ms]
open/16   this PR    [42.7 ms  48.5 ms  52.7 ms]
```

~8% on the median. The bench runs 10 samples so the intervals overlap; the copy elimination is exact and the direction matches. `commit/16` is unchanged, as expected — it never enters the fold loop.

## The rest

- **`fold_codeword` narrows its precondition to a power-of-two length**, where it previously checked only evenness. This is a deliberate tightening of a `pub` contract, not an assert catching up with the docs: `gray_chain_steps` sizes `levels = num_pairs.next_power_of_two().trailing_zeros()`, and the loop's maximum index is always `levels - 1`, so the step table covers any even length. At `num_pairs = 3` the chain is `domain_point(0)`, `domain_point(2)`, `domain_point(4)` — all correct. The doc no longer claims the step table as the reason.
- **`query_security_bits` is documented as a lower bound once the cap binds.** The query phase executes `min(num_queries, domain_size/2)` tests, and where the derived count exceeds that, opening every pair checks every fold at every position — a query-phase error of zero, not `(1-δ)^k`. The reported figure therefore understates rather than overstates.
- **`field_security_bits_at` records its composition convention.** The commit term is `p3-security`'s *per-round* bound charged once, which is the convention `RegimeReport::binding` composes under (minimum bits over terms); the sumcheck term is union-bounded over rounds. The mixture is conservative and, at arity 2, inert — `n + 1 = 2^22` against `2 * 20`. Written down so it stays legible if the arity ever moves.
- **The `REGIME` guard becomes a `const` assert.** It was a runtime branch on a compile-time constant, so `UnsupportedSoundnessRegime` was a public variant no caller could observe and no test could construct. Given the other regimes are *refuted* over `F_2`-subspace domains, a change to `REGIME` should fail to compile rather than return an error. The variant is removed; the enum is `#[non_exhaustive]` and unreleased, and the variant was unconstructible.
- **`sample_query_indices`' rationale is corrected.** It said the low bits of a sampled field element are biased. Over `GF(2^128)` they are not — a uniform element is a uniform 128-bit string — and `BinaryChallenger::sample_uniform_bits` is itself an unrejected mask over transcript bytes. The uniform API is right, for the different reason that it is the one that *states* the guarantee.
- **`verify_opening` returns a borrow.** Only `verify_at`, which owes its caller an owned value, clones.
- **The README states two obligations the types do not carry.** The scheme is binding, not hiding: the final codeword travels in the clear and openings are raw symbols. And `BinaryPcsConfig` prices the field's width and the query count but nothing the MMCS contributes, so collision resistance is the caller's to supply.
- **`pow_bits = 0` is covered.** It is the one budget where both `BinaryChallenger::grind` and `GrindingChallenger::check_witness` return early without touching the transcript. They agree only because *both* do; drop either and prover and verifier desync at query sampling, so every honest proof stops verifying. `try_new` accepts it for any `security_level >= 1`.

## Left out on purpose

Two items from the review are real but want their own PRs:

1. **FRI commit arity.** The commit phase commits a Merkle tree per fold round — arity 1. The message-side folding is genuinely forced by the sumcheck lockstep, but the commit arity is the ordinary FRI knob and is free: open a coset of `2^k`, apply `fold_pair` `k` times locally, compare against round `c + k`. `p3-fri` already has the machinery (`log_arities`, `fold_row`, `extra_query_index_bits`). binius64 runs the same construction over the same Gao–Mateer domain in the same regime and its proof-size DP lands on arity 4 at every shape it ships. This is the largest number in the crate and deserves a measured PR, not a rider.
2. **Dropping the symbol the verifier already computes.** At round `r + 1` one of the two opened symbols is exactly the fold of round `r`; `p3-fri` sends only `sibling_values` and reconstructs it. Worth ~8% of proof size, but it moves a `FoldMismatch` into the Merkle check and churns the negative tests' expected variants, so it should land on its own.

## Test plan

- [x] `cargo test -p p3-binary-pcs`, with and without `--features parallel`: 37 lib + 19 integration.
- [x] `cargo clippy -p p3-binary-pcs --all-targets --all-features -- -D warnings`.
- [x] `cargo +nightly fmt -p p3-binary-pcs -- --check`.
- [x] `thumbv7em-none-eabi` build.
- [x] `open/16` and `commit/16` benched against the merged baseline, numbers above.

### DIFF
--- a/binary-pcs/src/fold.rs
+++ b/binary-pcs/src/fold.rs
@@ -79,13 +79,16 @@ fn gray_chain_steps(num_pairs: usize) -> Vec<BinaryField128> {
 ///
 /// # Panics
 ///
-/// Panics if `codeword` has odd length. A codeword always has a power-of-two length, so an odd
-/// one means the caller has lost track of the round schedule.
+/// Panics unless the codeword is empty or has a power-of-two length.
+/// That is the shape every round schedule produces.
+///
+/// The chained step table itself covers any even length, so this is a deliberate
+/// narrowing of the contract rather than a limit of the method.
 #[must_use]
 pub fn fold_codeword(codeword: &[BinaryField128], beta: BinaryField128) -> Vec<BinaryField128> {
     assert!(
-        codeword.len().is_multiple_of(2),
-        "codeword length must be even"
+        codeword.is_empty() || codeword.len().is_power_of_two(),
+        "codeword length must be a power of two"
     );
 
     let num_pairs = codeword.len() / 2;
@@ -225,10 +228,25 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "codeword length must be even")]
+    #[should_panic(expected = "codeword length must be a power of two")]
     fn an_odd_length_codeword_is_rejected() {
         let odd = [BinaryField128::ONE; 3];
         let _ = fold_codeword(&odd, BinaryField128::ONE);
+    }
+
+    #[test]
+    #[should_panic(expected = "codeword length must be a power of two")]
+    fn an_even_non_power_of_two_codeword_is_rejected() {
+        // Even is not enough: the per-level step table is sized for a power of two.
+        // Six symbols is three pairs, which no round schedule produces.
+        let six = [BinaryField128::ONE; 6];
+        let _ = fold_codeword(&six, BinaryField128::ONE);
+    }
+
+    #[test]
+    fn an_empty_codeword_folds_to_an_empty_one() {
+        // Zero pairs to fold, and zero levels to index, so the empty input is not a panic.
+        assert!(fold_codeword(&[], BinaryField128::ONE).is_empty());
     }
 
     /// `f_0 + beta * f_1`, the novel-basis combination, is a different operation from
--- a/binary-pcs/src/params.rs
+++ b/binary-pcs/src/params.rs
@@ -33,6 +33,11 @@ const ALPHABET_BITS: usize = 128;
 /// caller, which is why this crate does not re-export `SecurityAssumption`.
 const REGIME: SecurityAssumption = SecurityAssumption::UniqueDecoding;
 
+// A change of regime must fail to compile, not fail at run time.
+// The other regimes are refuted over `F_2`-subspace domains.
+// No configuration should accept them, so there is nothing to report as an error.
+const _: () = assert!(matches!(REGIME, SecurityAssumption::UniqueDecoding));
+
 /// `ceil(log2(value))`.
 ///
 /// For `value >= 2` the highest set bit of `value - 1` sits one below the ceiling at a power
@@ -60,6 +65,24 @@ const fn log2_ceil_u128(value: u128) -> usize {
 /// Their sum is `n + 1 + 2 * num_fold_rounds` over `|F|`. The sum cannot overflow: `try_new`
 /// admits `log_domain_size` only below `usize::BITS`, so `n` stays below `2^64`.
 ///
+/// The two terms compose under different conventions:
+///
+/// - The commit term is a per-round bound, charged once.
+/// - The sumcheck term is union-bounded over every round.
+///
+/// Charging a per-round bound once is the round-by-round convention.
+/// The security crate composes the same way, taking the worst term rather than summing errors.
+/// The mixture here is conservative, never optimistic.
+///
+/// At arity 2 the choice is numerically inert:
+///
+/// ```text
+///     commit    n + 1      = 2^22   at the largest shape benched here
+///     sumcheck  2 * rounds = 40
+/// ```
+///
+/// Raising the arity shrinks the round count without shrinking `n`, so the two stay far apart.
+///
 /// Taking the ceiling of the logarithm rounds the result **down**, so this understates the
 /// achieved security by up to one bit rather than overstating it.
 const fn field_security_bits_at(log_domain_size: usize, num_fold_rounds: usize) -> usize {
@@ -79,6 +102,11 @@ pub struct BinaryPcsParams {
 }
 
 /// The round schedule derived from [`BinaryPcsParams`] and the polynomial's arity.
+///
+/// The schedule prices the field's width and the query count, and nothing else.
+/// In particular it does not price the commitment scheme, which is supplied separately.
+/// A digest narrower than `2 * security_level` bits leaves the reported level undeliverable.
+/// Supplying one wide enough is the caller's obligation.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct BinaryPcsConfig {
     params: BinaryPcsParams,
@@ -127,13 +155,6 @@ pub enum BinaryPcsConfigError {
     /// The derived query count was zero, which would accept any codeword.
     #[error("derived a query count of zero")]
     ZeroQueries,
-
-    /// Reached with a regime other than unique decoding.
-    ///
-    /// Unreachable through the public API, which exposes no regime knob; this guards the
-    /// private core against a future caller.
-    #[error("only the unique-decoding regime is supported over F_2-subspace domains")]
-    UnsupportedSoundnessRegime,
 }
 
 impl BinaryPcsConfig {
@@ -154,10 +175,6 @@ impl BinaryPcsConfig {
         num_variables: usize,
         params: BinaryPcsParams,
     ) -> Result<Self, BinaryPcsConfigError> {
-        if !matches!(REGIME, SecurityAssumption::UniqueDecoding) {
-            return Err(BinaryPcsConfigError::UnsupportedSoundnessRegime);
-        }
-
         // The codeword length is computed downstream as `1usize << log_len`, so `log_len`
         // must stay below `usize::BITS` or that shift is already invalid. `checked_add`
         // guards the sum itself: an adversarial `num_variables` and `log_inv_rate` must not
@@ -262,6 +279,14 @@ impl BinaryPcsConfig {
     }
 
     /// Bits of security the sampled queries deliver.
+    ///
+    /// The query phase runs one test per distinct fold pair.
+    /// It therefore executes at most half the domain's worth of tests, however many
+    /// the derivation asked for.
+    ///
+    /// Where the derived count exceeds that, this figure is a lower bound.
+    /// Opening every pair checks every fold at every position of every round, and the
+    /// query-phase error of that is zero rather than a power of the proximity gap.
     #[must_use]
     pub fn query_security_bits(&self) -> f64 {
         REGIME.queries_error(self.params.log_inv_rate, self.num_queries)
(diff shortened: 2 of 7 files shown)


binary-pcs/src/verifier.rs
```diff
@@ -44,21 +82,23 @@ where
...
-        if !indices.contains(&index) {
-            indices.push(index);
-        }
+        pairs.insert(pair);
     }
 
-    indices.sort_unstable();
-    indices
+    // `BTreeSet` iterates in ascending order, so doubling preserves it.
+    pairs.into_iter().map(|pair| pair << 1).collect()
```
Nashtare, line 101: **MEDIUM — labelled `refactor`, but this changes the wire format and fixes a real (small) soundness over-claim.**

I built `6ea70c48` and `3b06828d` and ran the same commit/open probe against each. Proofs are not preserved: at `num_variables=8, log_inv_rate=2` the row count is identical (108) but the opened values differ entirely, because the sampler now draws `log2(domain)-1` bits and doubles instead of drawing `log2(domain)` bits. Where the cap binds the shape changes too — `nv=1,lir=1` goes 8→4 rows, `nv=1,lir=2` 16→8, `nv=6,lir=1` 174→128.

More interesting: the base was genuinely over-claiming. It enforced distinct *symbols*, but `fold_pair_positions` masks bit 0 away at every round, so two symbols that are fold siblings run the *same* test while `query_security_bits()` charges for both. At `nv=1,lir=1` the base's first four "distinct" opened values were `[a, b, a, b]`. At `nv=8,lir=2` the expected sibling collisions among 54 draws over 1024 symbols is `C(54,2)/1023 ≈ 1.4`, so the base lost roughly 0.3–1 bit of the claimed 36 on average. Sampling `t` distinct *pairs* without replacement is at least as strong as `t` i.i.d. pairs, so the new scheme strictly dominates both the old code and the bound.

That is a good change — it just isn't a refactor. Suggested fix: relabel as `fix(binary-pcs):` and note in the message that the proof encoding changed. The crate is one commit old and unreleased so no stored proofs exist; the cost is purely a wrong changelog entry.

binary-pcs/src/params.rs
```diff
@@ -79,6 +102,11 @@ pub struct BinaryPcsParams {
 }
 
 /// The round schedule derived from [`BinaryPcsParams`] and the polynomial's arity.
+///
+/// The schedule prices the field's width and the query count, and nothing else.
```
Nashtare, line 106: **MEDIUM — `query_security_bits()` reports the uncapped query count even when the executed count is capped.**

This doc says the schedule "prices the field's width and the query count, and nothing else". `num_queries` is derived by `REGIME.queries(...)` with no domain-size term, but the query phase now executes `min(num_queries, domain_size/2)` tests via `num_distinct_queries` (`verifier.rs:226`). The cap binds in shapes the test suite itself runs: `nv=6, lir=1` derives 87 queries but executes 64; `nv=1, lir=1` derives 87 and executes 2. `query_security_bits()` returns `87 × 0.415 = 36.1` bits in both cases.

To be clear this direction is **conservative, not a security loss**: the cap can only bind when every pair is queried, which opens the entire base codeword and checks every fold at every position of every round exactly — the query-phase error there is zero, not `(1-δ)^64`. So the reported number is a lower bound, not an overstatement.

Suggested fix: document on `query_security_bits` that the figure is a lower bound once `num_queries >= domain_size/2` (full coverage being exact), or have it return the executed count's error with a special case for full coverage.

binary-pcs/src/fold.rs
```diff
@@ -79,13 +79,14 @@ fn gray_chain_steps(num_pairs: usize) -> Vec<BinaryField128> {
...
-/// one means the caller has lost track of the round schedule.
+/// Panics unless the codeword is empty or has a power-of-two length.
+/// That is the shape every round schedule produces.
+/// It is also the shape the per-level step table is sized for.
 #[must_use]
 pub fn fold_codeword(codeword: &[BinaryField128], beta: BinaryField128) -> Vec<BinaryField128> {
     assert!(
-        codeword.len().is_multiple_of(2),
-        "codeword length must be even"
+        codeword.is_empty() || codeword.len().is_power_of_two(),
```
Nashtare, line 90: **MEDIUM — this tightens a `pub` precondition, and the stated rationale is not accurate.**

The doc immediately above claims *"It is also the shape the per-level step table is sized for."* That is not the case. `gray_chain_steps` sizes `levels = num_pairs.next_power_of_two().trailing_zeros()`, and the loop indexes `steps[(start + offset).trailing_zeros()]` for `0 < start+offset < num_pairs`, whose maximum is always `levels - 1` — including for non-power-of-two `num_pairs`.

Worked example at `num_pairs = 3`, which is exactly the 6-element input the new `an_even_non_power_of_two_codeword_is_rejected` test now rejects: `levels = 2`, offsets 1 and 2 index `steps[0]` and `steps[1]`, and the chained `x` values are `domain_point(0)`, `+cantor_basis(1) = domain_point(2)`, `+cantor_basis(1)+cantor_basis(2) = domain_point(4)` — all correct. The step table was never the constraint.

`fold_codeword` is re-exported from `lib.rs:20`, so this narrows a public contract for purely defensive reasons. Suggested fix: keep the assert if you want the tighter invariant, but drop the step-table justification from the doc and the panic message, and mention the narrowing in the commit message.

Nashtare (approved):

binary-pcs/src/verifier.rs
tcoratger: Agreed, and the collision analysis is the part I had not worked out. Relabelled in 85c22583.

The PR title and the commit are now `fix(binary-pcs):`, and the opening states plainly that the query phase encoding changes and that proofs are not interchangeable — with the mitigating fact that the crate is one commit old and unreleased, so no stored proof exists.

I also rewrote the paragraph that claimed this was pure accounting. It now says what your probe shows: the old code enforced distinct *symbols* and charged the query bound for each, but fold siblings run the same test, so a colliding pair was paid for twice and counted twice. Sampling `t` distinct pairs without replacement dominates `t` i.i.d. pairs, so the new scheme is at least as strong as both the old code and the bound it was priced against. The `C(54,2)/1023 ~ 1.4` expected collisions at `nv=8, lir=2` are quoted as roughly 0.3 to 1 bit of the claimed 36.

Thank you for building both commits and diffing the openings — "the row count is identical but the values differ entirely" is exactly the kind of thing a `refactor` label would have hidden from a future bisect.

binary-pcs/src/params.rs
tcoratger: Documented in 85c22583, taking the first of your two options.

`query_security_bits` now says what the number is when the cap binds:

```
/// The query phase runs one test per distinct fold pair.
/// It therefore executes at most half the domain's worth of tests, however many
/// the derivation asked for.
///
/// Where the derived count exceeds that, this figure is a lower bound.
/// Opening every pair checks every fold at every position of every round, and the
/// query-phase error of that is zero rather than a power of the proximity gap.
```

I kept the returned value as the derived count's error rather than special-casing full coverage. Returning the executed count's error would mean returning infinity — or some sentinel — exactly when the scheme is strongest, and every caller would then have to handle that case to print a number. A documented lower bound seemed the smaller surprise. Happy to switch if you would rather the accessor be exact.

The commit message carries the same note, so the `nv=6, lir=1` shape (87 derived, 64 executed) is legible from the history rather than only from the doc.

binary-pcs/src/fold.rs
tcoratger: You are right, and the worked example at `num_pairs = 3` settles it — `levels = 2` covers offsets 1 and 2, so the step table was never the constraint.

Fixed in 85c22583. The false justification is gone from the doc, replaced by one that says what is actually true:

```
/// Panics unless the codeword is empty or has a power-of-two length.
/// That is the shape every round schedule produces.
///
/// The chained step table itself covers any even length, so this is a deliberate
/// narrowing of the contract rather than a limit of the method.
```

I kept the assert. Stating the narrowing explicitly seemed better than leaving a reader to rediscover that the even-length path was fine and then wonder which of the two is wrong.

The panic message never mentioned the step table, so it is unchanged. The commit message now records the narrowing as narrowing rather than as an assert catching up with the docs — that phrasing was carrying the same false claim.

PR #2035, opened 2026-09-05
feat(fri)!: drive the FRI transcript through the typed Fiat-Shamir layer

tcoratger:
> Rebased onto `main` now that #2024 and #2033 have merged. The diff is this work alone, in two commits: the migration, then the review fixes.

## What this changes

FRI derives every challenge through the typed Fiat-Shamir layer instead of writing its transcript by hand on both sides.

## The problem

A FRI run is a fixed sequence: per round absorb a commitment, grind, sample a folding challenge; then the final polynomial, the arity schedule, a grind, and the query indices. The prover wrote that sequence in one place and the verifier wrote it again in another, with nothing checking they agreed.

And the *shape* of the run was bound only by whatever happened to be absorbed. The width of a query index was bound by nothing — a narrower draw ranges over a smaller space, and the sampled value alone does not show that.

## The idea

Describe the transcript once, from the parameters and the folding schedule, and let both sides walk that description.

```
per round:  commitment   one opaque value
            grinding     only when the difficulty is positive
            folding      one extension element
final polynomial         final_poly_len extension elements
query grinding           only when the difficulty is positive
query indices            num_queries draws of index_bits bits
```

A fingerprint of that description enters the sponge before any step runs, so it binds the round count, the index width, and both grinding difficulties.

Everything that leaves the step sequence unchanged goes in the instance label instead: the blowup, the arity cap, and the arity of every round. Nothing about the configuration is invisible to the seed.

The per-round arities need the label rather than the fingerprint, because a round count does not pin them:

```
[3, 3, 2]   and   [2, 3, 3]
```

Same count, same total, so the same step sequence and the same fingerprint. Reachable as inputs `[10]` against `[10, 8]` at one cap. The verifier's derived-schedule comparison from #2033 already rejects a mismatch before the transcript exists, so binding them is defence in depth rather than a hole being closed — but the schedule is what the protocol *is*, and one absorb is a cheap way to say so.

One consequence of that comparison: the commit phase used to derive its own copy of the schedule while the shape derived another. Two derivations that must agree is the coupling worth removing, so the commit phase now walks the schedule the shape fixes.

## The opaque tag, and what it does not bind

FRI absorbs commitments through `CanObserve<FriMmcs::Commitment>` — an associated type with no shape the layer can see. `TypeTag::Opaque` records the step and says so explicitly.

What that leaves unbound is a commitment's width. Stated plainly because it is a real gap:

- Two runs differing only in commitment width share a fingerprint.
- Both sides absorb a commitment exactly as it reaches them, so a width mismatch does not desynchronise the transcript either.
- What rejects it is the opening check that recomputes the commitment.

The alternatives were worse. A caller-supplied constant reintroduces exactly the obligation that makes a hand-absorbed domain separator forgettable. A required `Mmcs` accessor exports a concept not every implementation has: a commitment's element count is `2^cap_height * DIGEST_ELEMS`, one type-level and one runtime, and `ExtensionMmcs` just delegates. If the shape should be bound, an `Mmcs` accessor is the clean way and deserves its own PR argued on its own merits.

I did not construct a case where the unbound width is exploitable, and I have not proven none exists.

## Rounds a proof cannot supply

Every value the verifier replays comes from a proof, so every disagreement is a rejection:

- A described grinding step arriving with no witness reports a missing witness.
- A final polynomial of the wrong length is rejected before any coefficient is absorbed.

Neither can stay a panic. Both drivers panic on drop if the transcript was never finalised, which is what stops a verifier silently skipping its trailing-bytes and full-replay checks — so a panic raised inside a live transcript scope unwinds into that one, and a second panic during unwinding calls `abort()`. Releasing the completeness check before returning is what lets the rejection travel alone. Verified by removing the release: the rejection turns back into a drop-time panic.

`verify_fri` pre-checks both lengths, so these guard external drivers, which is what the re-export makes possible.

- `TypeTag::Opaque`, plus `observe_opaque` on both drivers. An opaque step records no Rust type either — a step that declares it does not model its type cannot then compare one.
- `TypeTag::Bits { width }`, plus `challenge_bits` on both drivers. The width is the security-relevant part, so it belongs in the shape.

## Test plan

- [x] `cargo test -p p3-fri` — 80 pass across all targets
- [x] `cargo test -p p3-uni-stark -p p3-batch-stark -p p3-circle` — pass, including every compatibility fixture
- [x] `cargo test -p p3-challenger` — 180 pass across all targets
- [x] `cargo clippy -p p3-fri -p p3-uni-stark -p p3-batch-stark -p p3-circle --all-targets` — clean
- [x] `cargo build --workspace --all-targets` — clean
- New: the bit width is part of the shape; an opaque step is distinguishable from every modelled one while two opaque steps of different widths are not.
- New: reordering the folding schedule changes the seed; a round with no witness and a final polynomial of the wrong length are both errors, driving the verifier transcript directly with no pre-check in front.

## Breaking changes

- **The FRI transcript changes, so proofs produced before this do not verify after it.** Seeding absorbs the protocol identity and the shape fingerprint at the start of every run, and the per-round arity absorb moves into the seed. The two checked-in fixtures that route through the FRI verifier are regenerated: `uni_stark_two_adic_v1.postcard` and `batch_stark_two_adic_v1.postcard`. The circle fixtures regenerate byte-identical, and the WHIR and STIR ones never enter this path.
- `prove_fri` and `verify_fri` now require `Val: PrimeField64` and `GrindingChallenger<Witness = Val>`. Every grinding challenger in the workspace already satisfies both; the bound propagates to `TwoAdicFriPcs` and `HidingFriPcs`.


fri/src/prover.rs
```diff
@@ -91,39 +91,35 @@ where
...
+        .iter()
+        .map(|input| log2_strict_usize(input.len()))
+        .collect();
+    let shape = fri_shape(
+        params,
+        &input_log_heights,
+        log_global_max_height + folding.extra_query_index_bits(),
+    );
+
+    let mut transcript = ProverTranscript::<Challenger, Val, Challenge>::new(challenger, shape);
```
Nashtare, line 179: **HIGH — the FRI transcript changed and the two FRI compatibility fixtures were not regenerated. `cargo test` is red on this commit.**

`ProverTranscript::new` -> `ProverState::new` -> `DomainSeparator::seed` absorbs `[protocol_id(64) | pattern_hash(32) | len | instance_label | 0x80]` into the sponge at the start of every FRI run. That absorb did not exist at `cdd35e12`. This commit also removes the per-round `challenger.observe(Val::from_usize(log_arity))` from both sides. Net effect: every FRI proof produced before this commit is now invalid.

Measured, not inferred — same command, fresh extraction and fresh target dir in each tree:

| revision | `verify_two_adic_compat_fixture` (uni-stark) | (batch-stark) |
|---|---|---|
| `cdd35e12` | ok | ok |
| `76a2a6b8` | **FAILED** | **FAILED** |

The head failure is `Verification(InvalidOpeningArgument(InputError(MalformedPrunedProof(SiblingCountMismatch { expected: 8, got: 12 }))))` — the redrawn query indices no longer match the ones the fixture was proved against. The rest of the suite (`p3-fri`, `p3-challenger`, `p3-whir`, `p3-stir`, `p3-multi-stark`, and the other uni-stark/batch-stark targets) passes; exactly these two fail.

Needing regeneration:
- `uni-stark/tests/fixtures/uni_stark_two_adic_v1.postcard`
- `batch-stark/tests/fixtures/batch_stark_two_adic_v1.postcard`

`multi_stark_whir_v1.postcard` and `uni_stark_stir_v1.postcard` do **not** — WHIR and STIR do not route through `verify_fri`, and both suites pass at head. The `_circle_` fixtures also pass.

Suggested fix: run the `#[ignore]`d generators (`cargo test -p p3-uni-stark --test fib_air -- --ignored`, `cargo test -p p3-batch-stark --test simple -- --ignored`) and commit the two `.postcard` files. Verified sufficient: after regenerating both in the head tree, all four `*_compat_fixture` tests pass.

Given the `!` in the title the break itself is clearly intended — only the fixtures and a description/changelog note are missing. The lower commit `cf8a6b64` set the precedent by regenerating the WHIR fixture for exactly this reason.

Worth noting separately: **CI has never run on this branch.** `check-runs` for `76a2a6b8` returns 0 and there is no Actions run for the head SHA, so nothing caught this automatically.

fri/src/transcript.rs
```diff
@@ -0,0 +1,414 @@
...
+        witness: Option<F>,
+    ) -> Result<EF, TranscriptFailure>
+    where
+        Com: Clone,
+        C: CanObserve<Com>,
+    {
+        self.state.observe_opaque(COMMITMENT, commitment);
+
+        if self.shape.commit_pow_bits > 0 {
+            let witness = witness.expect("a positive difficulty requires a witness per round");
```
Nashtare: **MEDIUM — `VerifierTranscript` panics instead of erroring when a grinding witness is absent (also at line 383).**

Both public replay methods take `witness: Option<F>` and `expect()` it whenever the configured difficulty is positive. `verify_fri` itself always passes `Some(..)` (verifier.rs:362, :369), so this is **not** reachable from the in-tree verifier — I traced it. But `VerifierTranscript` is re-exported from `fri/src/lib.rs`, so an external verifier that models "proof omitted the witness" as `None` gets a panic on adversarial input rather than a rejection.

The escalation makes it worse: the `expect` fires *after* `player.interact(..)` has advanced but *before* `player.abort()`, so the unwind hits `PatternPlayer::drop`'s `assert!(self.finalized)` and turns a single panic into a double-panic abort. A verifier service cannot `catch_unwind` that into a rejection.

(Same class as the `expect` in `sumcheck/src/generic_degree/transcript.rs:320` from #2024 — a fix at the `Drop` level would cover both.)

Suggested fix: return a `TranscriptFailure::MissingPowWitness` variant instead of `expect`, or narrow the parameter to `F` and let the caller supply the proof's field directly.

fri/src/transcript.rs
```diff
@@ -0,0 +1,414 @@
...
+#[must_use]
+pub fn pattern<F, EF>(shape: &FriShape) -> InteractionPattern
+where
+    F: PrimeField64,
+    EF: ExtensionField<F>,
+{
+    // Up to three steps per commit round, then three closing steps.
+    let mut steps = Vec::with_capacity(3 * shape.log_arities.len() + 3);
+
+    for _ in &shape.log_arities {
```
Nashtare: **MEDIUM — the fold schedule is no longer bound into the transcript; only the round count is.**

`pattern()` consumes `shape.log_arities` for its length only, and `domain_separator()` binds `log_blowup` and `max_log_arity` (transcript.rs:209-211) but not the arity values. The removed `challenger.observe(Val::from_usize(log_arity))` did bind them.

Two runs with schedules `[3,3,2]` and `[2,3,3]` — reachable via inputs `[10]` vs `[10,8]` with `log_final=2, max_log_arity=3` — share a round count, a total, an `index_bits`, and therefore a pattern hash and a seed.

I believe this is **sound as it stands**: `cdd35e12` (#2033) made the verifier derive `expected_schedule` from the claim's committed heights and reject any mismatch (verifier.rs:312-322) before the transcript is created, so the schedule is a deterministic function of already-public data.

Flagging it because the module doc-comment lists what the shape binds, and this is a deliberate reduction in defence-in-depth that now rests entirely on that one equality check surviving future refactors. Note #2033's own review flagged that the same check is skipped when `input_log_heights` is empty — this commit removes that guard, which helps, but the coupling is worth being explicit about.

If you want it back cheaply: add the schedule to the instance label alongside `log_blowup`/`max_log_arity`. One line, no runtime cost, and it needs the fixture regeneration from the other comment anyway.

Nashtare (changes requested):

fri/src/prover.rs
tcoratger: Fixed, and thank you for the table — that is a clean reproduction and it saved me from shipping a red branch.

Both files are regenerated:

- `uni-stark/tests/fixtures/uni_stark_two_adic_v1.postcard`
- `batch-stark/tests/fixtures/batch_stark_two_adic_v1.postcard`

Your triage was exactly right, and the regeneration confirms it from the other direction: the `_circle_` generators ran too and produced **byte-identical** files, so git shows no change for them. The STIR and WHIR fixtures are untouched. All five `*_compat_fixture` tests now pass.

I also added a note to the PR description, since a proof-format break deserves to be stated there and not only inferred from the `!`.

On CI never having run: confirmed, and I have not established why — so I am not going to guess at a cause. This push should produce a run, and if it does not that is worth chasing separately.

What I can account for is my side of it: I had run `p3-fri` and `p3-challenger` locally but not the stark suites that own the fixtures, so the scope was wrong regardless of what CI did. Widened for this push to `p3-fri`, `p3-uni-stark`, `p3-batch-stark` and `p3-circle`, plus every `*_compat_fixture` test explicitly.

fri/src/transcript.rs
tcoratger: Fixed, and you are right about the escalation being the worse half — I verified it rather than taking it on trust. Commenting out the release turns the rejection back into a panic at `challenger/src/fs/state/verifier.rs:64`, which is the driver's drop check firing mid-unwind.

`TranscriptFailure` gains `MissingPowWitness`, and both replay methods release the completeness check before returning it:

```rust
let Some(witness) = witness else {
    self.state.abort();
    return Err(TranscriptFailure::MissingPowWitness);
};
```

I kept `Option<F>` rather than narrowing to `F`, for the same reason as in #2024: a zero-difficulty run genuinely has no witness, and a dummy value would be worse than an explicit absence.

Two things came out of this that go slightly past the comment:

- `VerifierState::abort` did not exist. #2024 adds it, because a driver in another crate had no way to mark its player aborted — that is the shared piece your parenthetical was pointing at.
- Once #2024 makes `observe_extensions` return a `Result`, the final polynomial's length is the same shape of problem: it comes from the proof and was going straight to the player. So `TranscriptFailure::FinalPolyLen { expected, got }` now covers it, mapping to `FinalPolyLengthMismatch`. `verify_fri` pre-checks that length, so this is for external drivers only — which is the case you raised.

On the `Drop`-level fix: I went the other way and said why on the #2024 thread. Short version — `p3-challenger` is `no_std`, so there is no `thread::panicking()`, so any suppression is unconditional and silently drops the check that stops a verifier accepting a proof with appended bytes. Fixing the panic sites keeps both properties.

New tests: `a_commit_round_missing_its_grinding_witness_is_rejected` and `a_final_polynomial_of_the_wrong_length_is_rejected`, both driving `VerifierTranscript` directly with no pre-check in front.

fri/src/transcript.rs
tcoratger: Taken, and bound again — your one-line suggestion, in the instance label:

```rust
for &log_arity in &shape.log_arities {
    separator.instance(&(log_arity as u64).to_be_bytes());
}
```

Your `[3,3,2]` vs `[2,3,3]` case is now a test (`the_arity_of_every_round_reaches_the_seed`), including the `[10]` vs `[10,8]` reachability note, and it asserts the two seeds produce different challenges. Removing the loop makes it fail, so the binding is pinned rather than just present.

I agree with your soundness read and did not want to leave the module doc claiming more than the code did — the "What is bound" list now says the label carries the arity of every round, which is true again.

One thing your comment prompted that I think is the more durable half of the fix. The commit phase was deriving its own copy of the schedule while `fri_shape` derived another; two derivations that must agree is precisely the coupling you flagged. The commit phase now walks the schedule the shape fixes, passed in as an argument, so there is one derivation. The `#[cfg(test)]` forged-schedule seam from #2033 sets it on the shape, which is why folding, the transcript and the proof all still agree with a forgery — that test is still green.

tcoratger:
Rebased onto `main` now that #2024 and #2033 have merged, so the diff is this work alone — two commits, the migration and the review fixes. No conflicts; `mergeable` is green again.

Both cherry-picked clean onto the new base, and I re-ran everything on it rather than trusting the earlier run: `p3-fri` 80 pass, all five `*_compat_fixture` tests pass, fmt and clippy clean across `p3-fri`, `p3-uni-stark`, `p3-batch-stark` and `p3-circle`.

CI came back 21/21 green on the previous push, which also answers the "CI has never run on this branch" note — it runs now.

The three review threads are answered and resolved. The PR description had two claims that the review made false, so I rewrote them rather than leave them standing: the section that said the folding schedule was already covered by the shape fingerprint (it was not — that is what the schedule comment caught), and the stale note asking for review of only the last commit.

Ready for another look whenever you have time, @Nashtare.

tcoratger:
Pushed an API cleanup on top: the shape's three derivations were free functions that each took or produced a `FriShape` and nothing else, which left the type a bare record with its behaviour sitting beside it.

```rust
// before
let shape = fri_shape(params, &input_log_heights, index_bits);
let separator = domain_separator::<F, EF>(&shape);

// after
let shape = FriShape::new(params, &input_log_heights, index_bits);
let separator = shape.domain_separator::<F, EF>();
```

Deriving a shape is now its constructor; describing the transcript it fixes and sealing that into a seed are its methods. The seed derivation reaches the pattern through the receiver rather than passing the shape back down, and `fri_shape` leaves the crate root.

Pure move — 120 insertions, 120 deletions, no behaviour change. The compatibility fixtures verify unchanged, which is the check that matters here: it pins the absorbed bytes as identical, so nothing needed regenerating.

Still 80 pass in `p3-fri`, fixtures green, fmt and clippy clean.

fri/src/prover.rs
```diff
@@ -151,39 +152,45 @@ where
...
-    // N!/(N^{num_queries} * (N - num_queries)!) ~ (1 - 1/N)^{num_queries * (num_queries - 1)/2}
-    //                                           ~ (1 - num_queries^2/2N)
-    // Here N = 2^log_global_max_height.
-    // With num_queries = 100, N = 2^20, this is 0.995 so there is a .5% chance of a collision.
-    // Due to this, security conscious users may want to set num_queries a little higher than the
-    // theoretical minimum.
+    // Folding walks the described schedule, so the two cannot drift apart.
+    let log_arities = shape.log_arities.clone();
+
+    let mut transcript = ProverTranscript::<Challenger, Val, Challenge>::new(challenger, shape);
```
Nashtare, line 179: **MEDIUM — the abort escalation is now fixed on the verifier side but present on the prover side, where it produces an uncatchable `SIGABRT`.**

First, all three of my earlier comments are resolved, and I verified each by running code rather than reading:

- **fixtures / red tests** — 833 passed, 0 failed across `p3-fri`, `p3-uni-stark`, `p3-batch-stark`, `p3-challenger`, `p3-whir`, `p3-stir`, `p3-multi-stark`. I re-ran both `#[ignore]`d generators and all four regenerated fixtures are **byte-identical** to the committed ones, so the committed bytes genuinely match head. (The circle fixtures regenerating unchanged is correct — `p3-circle` has its own FRI driver and never enters `prove_fri`.)
- **missing-witness `expect`** — both sites now `let ... else { self.state.abort(); return Err(..) }`, abort before return. I exercised all five `VerifierTranscript` error paths plus a control: all five drop clean, and the control (unfinished, no error) still panics, so the release isn't over-eager.
- **schedule binding** — `transcript.rs:221-223` now folds each arity into the instance label. Confirmed concretely: `[3,3,2]` and `[2,3,3]` share a pattern hash, as expected, but produce different instance labels and different first challenges (`583270224` vs `1077635612`).

Now the residual. FRI did not use `ProverState` before this PR. `ProverState::drop` (`challenger/src/fs/state/prover.rs:47`) panics when the pattern was not fully played, and **unlike the verifier there is no release for that check**. So any panic between this line and `transcript.finish()` at `prover.rs:193` panics a second time during unwind.

Reachable sites in that window: `RowMajorMatrix::new(folded, arity)` at `prover.rs:292`, the `.pop().unwrap()` at `:304`, and any panic inside a caller-supplied `Mmcs` or `FriFoldingStrategy` — those last are public trait callbacks, so arbitrary user code runs inside the live scope.

I confirmed this rather than inferring it: a probe panicking in that window reports `panic in a destructor during cleanup / thread caused non-unwinding panic. aborting.` and exits `signal: 6, SIGABRT`. A `catch_unwind` proving service cannot contain that.

This is the same escalation comment (B) asked to be fixed, mirrored onto the prover side. Related: the comment at `challenger/src/fs/state/prover.rs:51` — *"Every path that panics or bails marks the player aborted first, so this check never fires during cleanup of another failure"* — is now false. The equivalent comment on the verifier **is** accurate; I verified it.

Suggested fix: guard `ProverState::drop` with `std::thread::panicking()` behind a `std` feature, or call `transcript.abort()` from a scope guard covering the fallible region of `prove_fri_with_schedule`.

Not blocking from my side — it is an inherited property of the `p3-challenger` `fs` layer that WHIR and STIR already carry, so it is arguably its own PR.

Three smaller things, not filed separately:
- **Three of the four commits are RED.** `ec29b80b`'s message says *"The two FRI compatibility fixtures are regenerated"*, but `git show ec29b80b --stat` lists only `transcript.rs` and `verifier.rs`; the fixtures land in `5869f38d`. I extracted `059b4bf2` and `verify_two_adic_compat_fixture` fails with `InvalidOpeningArgument(InvalidPowWitness(CommitPhase))`. That breaks `git bisect` across this range for any future FRI or STARK regression. Moving the regeneration into `96775d78` (or squashing) would fix it, along with dropping that sentence.
- `TypeTag` (`challenger/src/fs/pattern/step.rs:102`) gains `Opaque` and `Bits { width }` without `#[non_exhaustive]`. No in-tree exhaustive matches, and the PR is already marked `!`, so impact is bounded — but worth adding while the type is young.
- `FriShape::new` at `prover.rs:163` always runs `fold_schedule` even when the caller-supplied schedule overrides it two lines later. Besides the wasted work, `fold_schedule` asserts strictly-decreasing input heights, so `prove_fri_with_schedule` can now panic on inputs the override path previously bypassed. That path is `#[cfg(test)]`-only today, so cosmetic.

Nashtare (approved):

fri/src/prover.rs
tcoratger: Fixed, and at the layer rather than in FRI, so WHIR and STIR come along: all four panicking `Drop` impls (`ProverState`, `VerifierState`, `PatternPlayer`, `PatternState`) now stand down while a panic unwinds.

No cargo feature was needed. The gate is `#[cfg(panic = "unwind")] extern crate std;` plus `!std::thread::panicking()` — `panic = "unwind"` is exactly the configuration where `std` links, and a panic-abort target ends the process on the first panic so no second one can follow it. I probed `cfg(panic)` on both no-std targets before relying on that: `thumbv7em-none-eabi` and `wasm32-unknown-unknown` are abort, and cargo passes no `-C panic` of its own. `cargo build --target thumbv7em-none-eabi -p p3-challenger -p p3-fri` still builds.

Your probe is a test now, at the FRI level and one level down. Before the fix, `p3-fri transcript::tests::a_panic_inside_a_live_prover_transcript_unwinds`:

```
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
(test aborted with signal 6: SIGABRT)
```

After: `1 test run: 1 passed`. Same before/after for `a_panic_inside_a_live_{prover,verifier}_scope_unwinds` in `p3-challenger`.

One correction to your reading of the verifier comment: it is accurate for the driver's *own* failure paths, which is what you verified, but a foreign panic in the live scope escalated identically — the verifier test above also aborted before the fix. So both drops carry the guard and both comments now describe that.

**RED commits.** `ec29b80b` turned out to contain nothing but the transcript test module and one `fn` -> `const fn`; every substantive change its message claimed already sat in `96775d78` from an earlier rebase, which is why the bytes never moved there. So I squashed it into `96775d78`, merged the true paragraphs into that message, and put the fixture regeneration in the same commit. `5869f38d`'s fixture blob and the sentence claiming it are gone, since the bytes are fixed from the first commit onward. Per-commit `cargo nextest run -p p3-fri` over the six commits: 84/84, 84/84, 84/84, 85/85, 85/85, 85/85. At the first commit `verify_two_adic_compat_fixture` under `p3-uni-stark` and `p3-batch-stark` — the one you extracted and saw fail — now passes.

**`TypeTag`** is `#[non_exhaustive]`. Confirmed your read: the only exhaustive match is `Display` inside the defining crate, where the attribute does not apply, and `p3-binary-field` only constructs a variant, which stays legal.

**`FriShape::new`** is split into `new` (derives) and `with_schedule` (takes one). The prover picks by whether it was handed a schedule, so the override path neither derives one it discards nor reaches `fold_schedule`'s strictly-decreasing assert — reachable, since `prove_fri_with_schedule` accepts inputs of merely non-increasing length. The verifier got the same win: it already held `expected_schedule` and was deriving an identical second copy inside `FriShape::new`.

**Rebase.** Onto `c51c2fac`. #2058's fixture rename was the whole conflict; rename detection carried this branch's bytes onto `*_v0_8_0.postcard`, and re-running all four `#[ignore]`d generators at head leaves the committed bytes byte-identical (`sha256sum -c`, clean `git status`).

Verification, narrower than last round on purpose: `cargo nextest run -p p3-fri -p p3-challenger` is 267/267; the per-commit `p3-fri` loop above; all four generators plus `sha256sum -c` and a clean `git status`; the six fixture verification tests across `p3-uni-stark`/`p3-batch-stark`/`p3-multi-stark`, 6/6; `cargo clippy -p p3-fri -p p3-challenger --all-targets -- -D warnings`; `cargo +nightly fmt --all`; and the `thumbv7em-none-eabi` build of both crates. I did **not** run the STARK, WHIR or STIR suites in full this time, nor the workspace `--features parallel` and doctest legs — the change touches no rayon path, and CI covers the rest.

PR #2036, opened 2026-09-05
perf(sumcheck): bind and measure in one pass

tcoratger:
A sumcheck round makes two full passes over its two tables. This makes one.

No transcript change: the round messages, the challenges and the bound tables come
out bit-identical to what the round-at-a-time driver produces, and a test pins that.

## The problem

`SumcheckProver::compute_sumcheck_polynomials` drives its rounds one at a time:

```
measure   read evals, read weights                    -> (h(0), h(inf))
           ... transcript, challenge r ...
bind      read evals, read weights, write both halves
```

The binding writes the bound halves out, and the very next round's measurement reads
them straight back. At `num_variables = 20` over `BinomialExtensionField<BabyBear, 4>`
each table is 16 MiB, so nothing survives in cache between the two passes — the round
pays a memory round trip for data it had in registers one instruction earlier.

## The idea

Hold the challenge back one round, and bind inside the next round's measuring pass.

A fused pass touches two variables at once: `X`, bound now to the previous round's
challenge, and `Y`, the one the message it returns sums over. Together they cut each
table into four quadrants:

```
    X = 0, Y = 0 : q0        X = 1, Y = 0 : q2
    X = 0, Y = 1 : q1        X = 1, Y = 1 : q3
```

Binding `X = r` leaves the two faces of `Y`:

```
    lo = q0 + (q2 - q0) * r        (written back over q0)
    hi = q1 + (q3 - q1) * r        (written back over q1)
```

and those two faces are exactly what the round message needs:

```
    h(0)   = sum  w_lo * e_lo
    h(inf) = sum (w_hi - w_lo) * (e_hi - e_lo)
```

So the pass writes the bound table and measures it before the values leave L1. The
work goes a block at a time — four bound faces sized to stay resident — with the
existing `K`-tiled delayed-reduction dot products doing the measuring, unchanged.

The driver then looks like this:

```
round i:  bind r_{i-1}  +  measure h_i      (one pass)
after:    bind r_{k-1}                      (one pass)
```

Only the first measurement and the last binding are still stand-alone.

## Per table, over a whole `k`-round fold

Counting element accesses in units of the initial table length `N = 2^k`:

- **before:** `4N` reads + `N` writes
- **after:** `3N` reads + `N` writes

One read of every bound entry is gone. Multiply count is unchanged.

## Benchmark

`sumcheck/babybear/prover/plain_prefix`, `RUSTFLAGS=-Ctarget-cpu=native` (AVX-512),
40 samples, base and branch alternated in one process against the same criterion
baseline. `plain_suffix` is the control arm — it takes the fallback path and does not
move (`p = 0.69` and `p = 0.60` at `k16` / `k20`).

**32 threads, `--features parallel`:**

| variables | before | after | speedup |
|---|---|---|---|
| 12 | 23.98 µs | 19.40 µs | 1.24x |
| 16 | 412.7 µs | 215.2 µs | 1.92x |
| 20 | 5.85 ms | 3.41 ms | 1.72x |

**Single core, no `parallel` feature:**

| variables | before | after | speedup |
|---|---|---|---|
| 12 | 25.34 µs | 19.77 µs | 1.28x |
| 16 | 289.0 µs | 247.1 µs | 1.17x |
| 20 | 5.78 ms | 5.87 ms | 1.00x (`p = 0.99`) |

Reading the two together, because they say different things:

- **On one core the fusion pays where the tables fit in cache** and is neutral at 20
  variables, where a single core cannot saturate memory anyway and the pass is
  compute-bound. Same multiplies, fewer bytes, so this is the floor.
- **The larger parallel win at 16 and 20 variables is the round entering the thread
  pool once instead of twice**, and the fused pass parallelising the whole round
  rather than only its binding half. Under the current thresholds the baseline's
  measuring pass goes serial from round 1 at `k = 20` while its binding stays
  parallel; the fused pass keeps both together.

That second half of the win overlaps with any future work on rayon task sizing, and
would shrink if the thresholds were retuned. The first half would not.

A second alternated pair, run under heavy external load, reproduces the direction
(1.31x / 2.01x / 2.08x); its baseline arm is too noisy to quote, but the branch arm
lands within 0.5% of the table above at every size, which is itself the point — one
dispatch per round instead of two also makes the round much less load-sensitive.

## Scope

- **new:** `fold_and_round_coefficients_prefix` in `p3-sumcheck::strategy`, with the
  `bind_face` / `round_coefficients_faces` pieces it is built from, and
  `Poly::truncate_to_half` in `p3-multilinear-util`.
- **changed:** `SumcheckProver::compute_sumcheck_polynomials` defers its bindings;
  `ProductPolynomial::fold_round_coefficients` dispatches to the fused kernel.
- **untouched:** `ProductPolynomial::round`, which stays the reference the new driver
  is pinned against, and every kernel it calls.

Two states fall back to binding and measuring separately:

- **suffix binding**, whose bound entries are adjacent pairs rather than half-apart
  faces, so an in-place write would race its own reads;
- **fewer than four table entries**, where the bound half has no variable left for the
  message to sum over. That is also where the packed-to-scalar transition happens, so
  the fallback keeps the handoff exactly as it was.

## Prior art

This is `MleStore::map_reduce_with_fold` from binius64, ported to the shape
`ProductPolynomial` already has. Their measurement of the same change — a fused round
at 17.5 ms against 24.0 ms for the naive read-`2N`-write-`N` triad, both at ~46 GB/s
against a 47.4 GB/s streaming ceiling — is the reason to expect a bandwidth win rather
than an arithmetic one, and the serial-versus-parallel split above is that expectation
holding.

## Test plan

- [x] `prop_fold_and_round_coefficients_prefix_matches_bind_then_measure`: over
      `2..=16` variables, the fused pass produces the same bound tables *and* the same
      message as `fix_prefix_var_mut` followed by `sumcheck_coefficients_prefix`. The
      range straddles the `K`-tiled body and the par-vs-serial split.
- [x] `deferred_binding_drives_the_same_rounds_as_round_at_a_time`: the driver's
      recorded round messages, challenges, claim and both bound tables match a
      round-at-a-time drive, for both binding orders, across the packed path, the
      unpacking transition and the final scalar rounds.
- [x] `cargo test -p p3-sumcheck -p p3-multilinear-util`, with and without
      `--features parallel` — 285 + 148 tests.
- [x] `cargo test -p p3-whir -p p3-stir` — the WHIR PCS prover drives this loop.
- [x] `cargo clippy -p p3-sumcheck -p p3-multilinear-util --all-targets --features parallel`.
- [x] `cargo +nightly fmt --all`.

### DIFF
--- a/multilinear-util/src/poly/mod.rs
+++ b/multilinear-util/src/poly/mod.rs
@@ -169,6 +169,30 @@ impl<F> Poly<F> {
         self.0
     }
 
+    /// Keeps the lower half of the evaluation table and drops the upper half.
+    ///
+    /// Binding the first variable writes its result over the lower half.
+    ///
+    /// The upper half is dead once that is done.
+    /// Dropping it is what takes the arity down by one.
+    ///
+    /// This is that second step on its own.
+    /// A caller whose own pass wrote the lower half calls it to finish the binding.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the polynomial is constant.
+    #[inline]
+    pub fn truncate_to_half(&mut self) {
+        let num_evals = self.num_evals();
+
+        // A constant has no variable to bind, so there is no upper half to drop.
+        assert!(num_evals > 1, "no free variables");
+
+        // The lower half already holds the bound values.
+        self.0.truncate(num_evals / 2);
+    }
+
     /// Pads the evaluation vector with zeros up to `num_variables`.
     ///
     /// # Panics
--- a/multilinear-util/src/poly/mod.rs
+++ b/multilinear-util/src/poly/mod.rs
@@ -176,12 +176,14 @@ impl<F> Poly<F> {
     /// The upper half is dead once that is done.
     /// Dropping it is what takes the arity down by one.
     ///
-    /// This is that second step on its own.
-    /// A caller whose own pass wrote the lower half calls it to finish the binding.
+    /// This is that second step alone, for a pass that wrote the lower half itself.
+    /// On a table nobody bound it silently returns the restriction to the first
+    /// variable at zero, so it is hidden rather than offered.
     ///
     /// # Panics
     ///
     /// Panics if the polynomial is constant.
+    #[doc(hidden)]
     #[inline]
     pub fn truncate_to_half(&mut self) {
         let num_evals = self.num_evals();
--- a/sumcheck/src/strategy.rs
+++ b/sumcheck/src/strategy.rs
@@ -6,6 +6,8 @@
 //! - `VariableOrder`: tag enum carrying inherent methods that dispatch to either routine.
 //! - `SumcheckProver`: drives rounds over a paired product polynomial.
 
+use alloc::vec::Vec;
+
 use p3_challenger::{FieldChallenger, GrindingChallenger};
 use p3_field::{Algebra, ExtensionField, Field, PrimeCharacteristicRing, dot_product};
 use p3_maybe_rayon::prelude::*;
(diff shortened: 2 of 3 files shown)


sumcheck/src/strategy.rs
```diff
@@ -300,6 +302,228 @@ where
...
+///
+/// Never below the tile width, so a measurement always gets one full tile.
+#[inline]
+const fn fused_block<A>() -> usize {
+    // Positions that fit the byte budget at this element width.
+    let by_bytes = FUSED_BLOCK_BYTES / core::mem::size_of::<A>();
+
+    // A wide element can exhaust the budget below one tile.
+    // The tile width wins there.
+    if by_bytes < K { K } else { by_bytes }
```
Nashtare: **MEDIUM — `fused_block()` is not rounded to a multiple of the tile width `K`, so every block can pay a scalar tail.**

`round_coefficients_faces` splits each block with `as_chunks::<K>()` and sends `block_len % 8` positions down the eager per-pair `round_step` instead of the `K`-tiled delayed-reduction dot product. In the unfused kernel that tail is paid **once per table**; here it is paid **once per block**.

It is invisible in the benchmarked configuration because `BinomialExtensionField<BabyBear, 4>` gives `4096/256 = 16`, already a multiple of 8. It bites any extension whose packed element size doesn't divide `FUSED_BLOCK_BYTES` into a multiple of 8. Packed extensions are `repr(transparent)` over `[PF; D]`, so on AVX-512: degree 5 → `5 × 64 = 320 B` → `fused_block() = 12` → 4 of every 12 positions (33%) off the fast path; degree 3 → 21 → 5/21 (24%).

Correctness is unaffected — a degree-5 differential run passes. Suggested fix:

```rust
let by_bytes = (FUSED_BLOCK_BYTES / size_of::<A>()) / K * K;
```

before the `< K` clamp.

sumcheck/src/strategy.rs
```diff
@@ -300,6 +302,228 @@ where
...
+    weights: &mut [A],
+    r: Ch,
+) -> RoundMessage<A>
+where
+    A: Algebra<Ch> + Copy + Send + Sync,
+    Ch: Copy + Send + Sync,
+{
+    // Precondition: paired tables, with a variable left over for the message.
+    assert_eq!(evals.len(), weights.len());
+    assert!(evals.len().is_multiple_of(4));
```
Nashtare: **MEDIUM — empty-slice precondition hole in a public kernel.**

`0usize.is_multiple_of(4)` is `true`, so an empty pair sails past both asserts, produces zero chunks, and returns `RoundMessage { c_a: 0, c_inf: 0 }` — a silently wrong round message rather than a panic.

`strategy` is `pub mod` (`sumcheck/src/lib.rs:39`) and this function is `pub`, so it is reachable from outside the crate. The in-crate caller is guarded by `num_evals() >= 4`, so it is latent today.

Suggested fix: `assert!(evals.len() >= 4 && evals.len().is_multiple_of(4));`, which matches what the doc comment already says the intent is ("the bound table then keeps the variable the message sums over").

multilinear-util/src/poly/mod.rs
```diff
@@ -169,6 +169,30 @@ impl<F> Poly<F> {
...
+    /// Dropping it is what takes the arity down by one.
+    ///
+    /// This is that second step on its own.
+    /// A caller whose own pass wrote the lower half calls it to finish the binding.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the polynomial is constant.
+    #[inline]
+    pub fn truncate_to_half(&mut self) {
```
Nashtare, line 188: **MEDIUM — new public cross-crate API that silently yields a different polynomial when misused.**

Called on a `Poly` whose lower half was not first overwritten with bound values, `truncate_to_half` returns the restriction `p(0, ·)` — well-formed, wrong, and undetectable.

It exists solely as the second half of `fold_and_round_coefficients_prefix`, and the two halves are separated by ~40 lines in `ProductPolynomial::fold_round_coefficients`. Nothing in the type system ties them together.

Suggested fix: give the kernel `&mut Poly<A>` arguments instead of `&mut [A]` and let it truncate internally, so the half-operation never becomes public surface. Failing that, `#[doc(hidden)]`.

Nashtare (approved):

sumcheck/src/strategy.rs
tcoratger: Fixed in 6727f639 — good catch, and the degree-5 arithmetic is exactly right.

`fused_block` now rounds down to a whole number of tiles before the floor:

```rust
let whole_tiles = FUSED_BLOCK_BYTES / core::mem::size_of::<A>() / K * K;
if whole_tiles == 0 { K } else { whole_tiles }
```

On AVX-512 that turns degree 5 from 12 positions per block into 8, and degree 3 from 21 into 16 — no block remainder in either case, so the eager per-pair path is reached once per table again rather than once per block.

Pinned by `a_fused_block_is_a_whole_number_of_tiles`, which asserts the multiple-of-tile property at widths that do and do not divide the budget (1 B, 320 B, 192 B) and that an element wider than the whole budget still floors at one tile.

sumcheck/src/strategy.rs
tcoratger: Fixed in 6727f639, and the third comment closed it more thoroughly than the assert alone would have.

The assert is now what you suggested:

```rust
assert!(evals.len() >= 4 && evals.len().is_multiple_of(4));
```

Two things worth recording, since they changed what the hole actually is:

- The public entry now takes `&mut Poly<A>` rather than `&mut [A]` (your third comment). `Poly::new` rejects a non-power-of-two length, and zero is not a power of two, so an empty pair can no longer be constructed at all — the empty case is unreachable through the public surface, not merely rejected.
- What remains reachable is a pair of length 1 or 2, which binds down to a table with nothing left to sum over. That is what the regression guard targets: `binding_rejects_a_pair_with_nothing_left_to_measure` feeds a two-entry pair and expects the panic.

I kept the `>= 4` half of the assert regardless — it states the precondition for the private pass, which is where it would matter if that ever gained a second caller.

multilinear-util/src/poly/mod.rs
tcoratger: Took the primary suggestion in 6727f639 — the kernel takes the polynomials, not their slices, and truncates internally:

```rust
pub fn fold_and_round_coefficients_prefix<A, Ch>(
    evals: &mut Poly<A>,
    weights: &mut Poly<A>,
    r: Ch,
) -> RoundMessage<A>
```

It delegates to a private `bind_and_measure` over the raw tables and then drops both upper halves before returning, so the half-done state never escapes and the two steps can no longer drift apart across call sites. `ProductPolynomial::fold_round_coefficients` loses its two `truncate_to_half` calls with it.

`Poly::truncate_to_half` still has to exist — `Poly`s backing vector is `pub(crate)` to its own crate, so a cross-crate caller cannot shrink one otherwise — but it is now `#[doc(hidden)]` (your fallback) and its doc says plainly what it returns on a table nobody bound.

This also closed the empty-pair hole in your second comment structurally: `Poly::new` rejects a non-power-of-two length, so length zero is unreachable through this signature.

PR #2037, opened 2026-09-05
Wire STIR into the full proving pipeline

Nashtare:
## Summary

- Adds STIR as a selectable PCS backend in `p3-uni-stark`'s test suite and the `prove_prime_field_31` CLI example, alongside the existing FRI and Circle backends, mirroring how WHIR is wired into `p3-multi-stark`.
- `uni-stark/tests/stir_fibonacci.rs`: a self-contained integration test proving a Fibonacci AIR end-to-end through `TwoAdicStirPcs`, including a minimum-trace-height edge case and a postcard serialization compat fixture.
- `examples/`: new `--pcs <fri|stir>` CLI flag, `StirKeccakStarkConfig`/`StirPoseidon2StarkConfig` type aliases, `prove_monty31_keccak_stir`/`prove_monty31_poseidon2_stir` proof functions, and matching CLI integration tests. Mersenne31 rejects `--pcs stir` (STIR requires a two-adic field).
- `stir/src/pcs.rs`: fixes a real gap found while wiring this up — `CanObserve<StirCommitment<C>>` was only implemented for `DuplexChallenger`, not `SerializingChallenger32`, which the Keccak-Merkle configuration in the example needs.

## Test plan

- [x] `cargo test --workspace --all-features` passes (full green run, verified via exit code)
- [x] New `uni-stark` STIR integration tests pass, including the min-trace-height and serialization-fixture cases
- [x] New `examples` CLI tests pass for both KoalaBear Keccak and Poseidon2 STIR configurations
- [x] Manually ran `prove_prime_field_31` with `--pcs stir` and `--pcs fri` across KoalaBear/BabyBear × Keccak/Poseidon2 and confirmed proofs verify

https://claude.ai/code/session_01TBCxN6xHZBsbodFP79tEiu


stir/src/pcs.rs
```diff
@@ -201,6 +211,20 @@ where
     }
 }
 
+impl<F, Inner, C> CanObserve<StirCommitment<C>> for SerializingChallenger32<F, Inner>
```
tcoratger, line 214: **[minor, test coverage]** This is the only non-test library code in the PR, and `cargo test -p p3-stir` cannot exercise it.

`grep -rn SerializingChallenger32 stir/` returns two hits — the import and this impl. The `uni-stark` STIR test uses `DuplexChallenger`, so its only coverage is indirect, via `examples/src/tests.rs`, in a crate nobody editing `p3-stir` would think to run.

A unit test here can pin the actual property rather than just compilation — that the length prefix separates one commitment of two roots from two commitments of one root each:

```rust
let mut a = challenger();
a.observe(StirCommitment(vec![r0, r1]));

let mut b = challenger();
b.observe(StirCommitment(vec![r0]));
b.observe(StirCommitment(vec![r1]));

assert_ne!(a.sample_bits(32), b.sample_bits(32));
```

Worth running for both backends.

stir/src/pcs.rs
```diff
@@ -186,6 +187,15 @@ impl<C> Deref for StirCommitment<C> {
     }
 }
 
+/// `StirCommitment<C>`'s `CanObserve` impl is written once per challenger backend rather than
```
tcoratger, line 190: Three things on this block, in descending order of how much they matter.

**[minor] Nothing pins the two impls equal.** I diffed them — the bodies are byte-for-byte identical, which is right. But that is exactly the property a future edit breaks silently: prover and verifier share a challenger type, so a divergence between backends makes both sides consistently wrong rather than failing a test. The comment says the duplication exists because a blanket impl leaves `F` unconstrained, which is precisely the case `macro_rules!` covers — one body, two expansions, and adding `SerializingChallenger64` later becomes a line rather than a third copy.

**[nit] `///` on an `impl` block.** Rustdoc attaches this to the `DuplexChallenger` impl alone, but the text describes both impls and the absent third one, so the rendered docs claim something narrower than you mean. A doc comment above an `impl` is also not idiomatic — `impl` is a grouping mechanism; the documented items are the type and the trait. Plain `//`, or move the type-level parts onto `StirCommitment`.

**[nit] It is a wrapped prose paragraph.** Nine lines of continuous text with sentences running across `///` boundaries. Every fact in it is worth keeping — the coherence explanation, the length-prefix rationale, the `SerializingChallenger64` note are all things a reader needs. They are just three separate points and read much faster as three.

examples/src/proofs.rs
```diff
@@ -177,6 +188,112 @@ where
...
+    proof_goal: &PG,
+    dft: DFT,
+    num_hashes: usize,
+) -> StirKeccakTwoAdicResult<F, EF, DFT>
+where
+    StandardUniform: Distribution<F>,
+{
+    let val_mmcs = get_keccak_mmcs(3);
+    let challenge_mmcs = ExtensionMmcs::<F, EF, _>::new(val_mmcs.clone());
+    let stir_params = StirParameters {
```
tcoratger, line 214: **[minor, duplication]** This six-field literal appears three times: here, at line 269, and again in `uni-stark/tests/stir_fibonacci.rs`. No single source of truth, so the three can drift apart.

The FRI functions immediately above avoid exactly this — lines 109 and 164 both call `FriParameters::new_benchmark_high_arity(challenge_mmcs)`. A `StirParameters::new_benchmark(mmcs)` in `p3-stir` would mirror the house pattern and collapse all three.

That is new public API on `p3-stir`, so I'd leave the call to you rather than land it in a follow-up.

examples/src/proofs.rs
```diff
@@ -321,3 +438,22 @@ where
         proven.list_decoding_bits
     );
 }
+
+/// Report the security level STIR's low-degree test was configured to target.
```
tcoratger, line 442: **[nit, doc style]** The content here is the most useful documentation in the PR — the PCS-vs-STARK distinction, why FRI's number isn't comparable, and why STIR has nothing left to measure once the config exists are all things a reader genuinely needs. They are also four distinct claims packed into one nine-line paragraph with sentences wrapping across `///` lines, which is where they get lost.

Same shape as the block in `stir/src/pcs.rs` — one sentence per line, or bullets for the three reasons, and it reads at a glance.

Separately, and not a doc issue: the gap you describe is closable. `StarkSecurityParams::from_air` takes a concrete `FriRegime` (`uni-stark/src/security.rs:110`) and `p3-stir` has no `security_regime()` to match `fri/src/config.rs:54` — but `p3-security` already has a `LowDegreeTest` trait (`security/src/ldt.rs:25`) that `FriRegime` implements. Implement it for a `StirRegime`, take `impl LowDegreeTest` in `from_air`, and `--pcs stir` prints the same two numbers `--pcs fri` does. That is what makes the flag actually comparable, and it wants its own PR.

uni-stark/tests/stir_fibonacci.rs
```diff
@@ -0,0 +1,253 @@
...
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_stir::{SecurityAssumption, StirParameters, TwoAdicStirPcs};
+use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
+use p3_uni_stark::{InvalidProofShapeError, StarkConfig, prove, verify};
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+/// For testing the public values feature.
+pub struct FibonacciAir {}
```
tcoratger, line 22: **[minor, duplication]** 168 of this file's 211 non-blank lines appear verbatim in `fib_air.rs` — the AIR, `FibonacciRow`, the `Borrow` impl, `generate_trace_rows`, `write_fixture`/`read_fixture`, and three of the tests. The genuinely new content is `make_config`, `MIN_TRACE_HEIGHT`, and `test_smallest_trace`.

A `uni-stark/tests/common/mod.rs` holding the AIR and the fixture helpers would leave this file as the STIR-specific part, which is what it is actually about. Two independently maintained copies of an AIR is where they start to drift.

Not yours, but on the record since it now exists twice: `unsafe { align_to::<FibonacciRow<F>>() }` reinterprets `[F]` as a struct with no `repr(C)`, and `repr(Rust)` guarantees no field order. Technically UB in both files. Adding `#[repr(C)]` to both would make it sound at zero cost.

examples/examples/prove_prime_field_31.rs
```diff
@@ -345,6 +383,12 @@ fn main() {
                 ),
             };
 
+            if args.pcs == PcsOptions::Stir {
+                panic!(
+                    "STIR is only wired into the two-adic (KoalaBear/BabyBear) pipeline. Mersenne31 uses the circle STARK with FRI; remove --pcs or pick a two-adic field."
```
tcoratger, line 388: **[nit]** The message says exactly the right thing. But this is argument validation, and it fires as a panic after the AIR has already been constructed.

clap can express it declaratively so the user gets a usage error before any work starts. If that is awkward across the three field arms, moving the check to the top of the Mersenne31 arm at least gets the ordering right — reject the invalid combination before doing work that will be thrown away.

tcoratger (approved):
Read the whole thing plus the surrounding context — both `CanObserve<StirCommitment<C>>` impls, `SerializingChallenger32`'s `CanObserve<F>`/`CanObserve<Hash<..>>` impls, `StirParameters`' own constraints, the `fib_air.rs` this test file mirrors, and `StarkSecurityParams::from_air`. **No blockers and nothing major.** Findings are all minor, and mostly about the one piece of non-test library code here.

## What I checked hardest

**The two `CanObserve<StirCommitment<C>>` impls absorb identically.** This was the thing worth worrying about: a new challenger backend that absorbs a commitment differently is a Fiat-Shamir divergence, and it would not show up as a test failure — prover and verifier share a challenger type, so both would simply be wrong together. The bodies are byte-for-byte the same: length prefix as a field element, then each root in order. Correct.

**The length prefix is unambiguous on the byte stream.** For `SerializingChallenger32`, `CanObserve<F>` writes `to_unique_u32().to_le_bytes()` — 4 fixed bytes — and each `Hash<F, u8, N>` writes `N` raw bytes. A `k`-group commitment is `4 + N*k` bytes with the count first, so two commitments with different group counts cannot alias, and the "observe `[a,b]` as one" vs "observe `[a]` then `[b]`" confusion is closed. That is the property the comment claims, and it holds.

**Parameters are inside `StirParameters`' stated constraints.** `log_folding_factor = 2` and `log_starting_folding_factor = 2` both meet the documented `k >= 4`, and `log_starting_folding_factor <= log_starting_degree` is tight at `MIN_TRACE_HEIGHT = 4` — which `test_smallest_trace` pins exactly on the boundary. Good edge case to have picked.

**`CapacityBound` matches existing house convention**, not a new aggressive choice: `multi-stark/tests/whir_fibonacci.rs:128`, `whir_preprocessed.rs:138`, `whir/benches/zk_overhead.rs:62`, and most of `whir/src/pcs/tests.rs` all use it. One note on that below, but it is a cross-cutting question and not this PR's to answer.

**The `p3-stir` dep move from dev-deps to deps is right** — `examples/src/proofs.rs` and `types.rs` are `src/`, so it has to be a real dependency.

`cargo test -p p3-uni-stark --test stir_fibonacci`: 5 passed, 1 ignored. Clippy clean on `p3-stir`, `p3-examples`, `p3-uni-stark`.

## Findings

### [MINOR] The one library change has no test in the crate that owns it — `stir/src/pcs.rs:214`

`grep -rn SerializingChallenger32 stir/` returns exactly two hits: the import and the impl. So `cargo test -p p3-stir` cannot catch a regression in the gap this PR was written to close. The `uni-stark` STIR test uses `DuplexChallenger`, so it does not cover it either — the only coverage is indirect, through `examples/src/tests.rs`, in a different crate that a `p3-stir` change would not obviously prompt anyone to run.

A unit test next to the impl fixes that, and it can pin the actual security property rather than just "it compiles":

```rust
// Observing one commitment of two roots must differ from observing two of one root each.
// That is exactly what the length prefix buys.
let mut a = challenger();
a.observe(StirCommitment(vec![r0, r1]));
let mut b = challenger();
b.observe(StirCommitment(vec![r0]));
b.observe(StirCommitment(vec![r1]));
assert_ne!(a.sample_bits(32), b.sample_bits(32));
```

Worth having for both backends.

### [MINOR] Two hand-duplicated impls with nothing pinning them equal — `stir/src/pcs.rs:190`

They are identical today; I checked. But the whole risk is that a future edit touches one and not the other, and the resulting divergence is silent in the way described above. The doc comment even states the reason for the duplication is a coherence limitation on the blanket impl — which is precisely the situation `macro_rules!` exists for:

```rust
macro_rules! impl_observe_stir_commitment {
    ($ty:ty, [$($bounds:tt)*]) => { ... };
}
```

One body, two expansions, no drift. It also makes adding `SerializingChallenger64` later a one-line change rather than a third copy — and the comment already flags that gap.

### [NIT] `///` on an `impl` block — `stir/src/pcs.rs:190`

Two problems, one cause. Rustdoc attaches this to the `DuplexChallenger` impl alone, but it describes both impls *and* the absent `SerializingChallenger64` one, so the rendered docs say something narrower than the text means. And a doc comment above an `impl` block is not idiomatic Rust — `impl` is a grouping mechanism, the documented items are the type and the trait.

Make it a plain `//` comment, or move the parts that are about the type onto `StirCommitment` itself.

### [NIT] The doc block itself is a wrapped prose paragraph — `stir/src/pcs.rs:190`

Nine lines of continuous prose with sentences wrapping across `///` lines and clauses joined mid-line. The content is good — I'd keep every fact in it — but the repo's `/doc` convention is one sentence per line, bullets for structured points. Same applies to `report_stir_security_level` (`examples/src/proofs.rs:442`), which is a nine-line paragraph carrying at least four distinct claims.

I mention it because the *content* of that second one is genuinely valuable and currently hard to extract on a skim — see below.

### [MINOR] The STIR parameter literal is written out three times

`examples/src/proofs.rs:214`, `examples/src/proofs.rs:269`, and `uni-stark/tests/stir_fibonacci.rs`. Six fields, three copies, no single source of truth.

The FRI path immediately above does not do this — `examples/src/proofs.rs:109` and `:164` both call `FriParameters::new_benchmark_high_arity(challenge_mmcs)`. A `StirParameters::new_benchmark(mmcs)` in `p3-stir` mirrors the existing pattern and collapses the three. That is new public API, so it is your call, not something I would land unilaterally.

### [MINOR] `stir_fibonacci.rs` is 80% a verbatim copy of `fib_air.rs`

168 of 211 non-blank lines appear verbatim in `fib_air.rs`: `FibonacciAir`, `FibonacciRow`, the `Borrow` impl, `generate_trace_rows`, `write_fixture`/`read_fixture`, and three of the tests. The genuinely new content is `make_config`, `MIN_TRACE_HEIGHT`, and `test_smallest_trace`.

A `uni-stark/tests/common/mod.rs` holding the AIR and the fixture helpers would leave this file as the STIR-specific part. Two copies of an AIR is where they start to drift.

(Not yours: the `unsafe { align_to::<FibonacciRow<F>>() }` on a struct with no `repr(C)` is copied from `fib_air.rs` and is a pre-existing repo pattern. Flagging only so it is on the record — a `repr(Rust)` struct has no guaranteed field order, so that reinterpretation is technically UB in both files.)

### [NIT] The Mersenne31 rejection is a `panic!` after setup — `examples/examples/prove_prime_field_31.rs:388`

The message is good and says the right thing. But this is an argument-validation failure, and clap can express it declaratively so the user gets a usage error before any work starts, rather than a panic backtrace after the AIR is constructed. If a declarative version is awkward across the three field arms, moving the check to the top of the Mersenne31 arm at least gets the ordering right.

## Two notes that aren't findings

**Capacity-regime parameters, in an example people copy.** This matches what the WHIR wiring already does, so it is consistent — but `p3-security`'s own `SecurityAssumption::CapacityBound` says "Requires **conjecturing** capacity-rate list decodability and correlated agreement up to capacity", and that conjecture has since been refuted (Crites–Stewart, eprint 2025/2046, against WHIR's MCA-up-to-capacity; BCHKS25 Thm 1.6 in characteristic 2). The CLI now prints "100 bits of conjectured security", which is honest labelling. Still, `JohnsonBound` is the defensible default for a flagship example, and it is proven rather than conjectured. Repo-wide question, raised here only because this PR is what puts it behind a user-facing flag.

**The missing STARK-level number is real and closable.** `report_stir_security_level`'s doc is right that there is no STIR counterpart, and I checked why: `StarkSecurityParams::from_air` takes a concrete `FriRegime` (`uni-stark/src/security.rs:110`), and `p3-stir` has no `security_regime()` to match `fri/src/config.rs:54`. But `p3-security` already has a `LowDegreeTest` trait (`security/src/ldt.rs:25`) that `FriRegime` implements, so the path is: implement it for a `StirRegime`, then take `impl LowDegreeTest` in `from_air`. That would let `--pcs stir` print the same two numbers `--pcs fri` does, which is the thing that makes the flag actually comparable. Worth its own PR.

Approving and merging; I'll open a follow-up with the impl-side minors (test, macro, doc placement, doc style) so none of it blocks this. The parameter constructor, the shared test module, and the `StirRegime` are yours to call.

PR #2038, opened 2026-09-05
perf(sumcheck): bind and measure in one pass in the hiding driver

tcoratger:
Follows #2036, now merged, and rebased onto it.

#2036 fuses the binding into the next round's measuring pass in the plain sumcheck
driver. The hiding residual driver runs the same round loop and did not get it.
This gives it the same treatment.

No transcript change: the round polynomials, the challenges and the bound tables are
unchanged.

## The problem

The hiding residual driver measures, samples, then binds:

```
measure   read evals, read weights                    -> (c0, c_inf)
           ... assemble h_j, transcript, grind, gamma ...
bind      read evals, read weights, write both halves
```

Two passes over the same two tables, the second reading back what the first wrote.
That is exactly the shape #2036 removed from the plain driver, on the same
`ProductPolynomial` and the same kernel.

## The change

Hold the challenge back one round, so the binding rides into the next round's
measuring pass:

```
round j:  bind gamma_{j-1}  +  measure round j      (one pass)
after:    bind gamma_{k-1}                          (one pass)
```

Nothing the loop does between the measurement and the binding touches the tables.
Assembling the round polynomial, absorbing it, grinding, sampling and evaluating the
mask at the challenge are all scalar work.

The last challenge has no successor to fuse with, and the weight scaling after the
loop reads the tables, so it binds there.

## One shape instead of two

Both drivers now go through the same three helpers rather than hand-rolling the
claim-and-bind step twice:

- measure the round, absorbing whatever binding the last one left behind;
- advance the claim through the round identity, leaving the binding pending;
- apply a binding that no measuring pass absorbed.

The plain driver switches to them in this commit too, so the deferral is written once.

## Benchmark

`sumcheck/benches/sumcheck.rs` gains a `zk_residual` arm driving this driver directly
at `n = 20`, `k = 4`, with the mask code and Merkle commitment sized as the hiding
WHIR prover sizes them.

A/B by swapping only the deferral in the driver, two alternated pairs in one process,
`-Ctarget-cpu=native`, `--features parallel`, 32 threads:

| | bind on the spot | deferred | speedup |
|---|---|---|---|
| pair 1 | 5.49 ms | 3.83 ms | 1.43x |
| pair 2 | 5.57 ms | 3.71 ms | 1.50x |

Both `p < 0.05`. The figure sits below the plain driver's 1.7x because the timed
region also commits the mask oracle, which the fusion does not touch.

Two notes on why the pre-existing benches were not used:

- `whir/benches/sumcheck.rs`'s `zk_prefix` / `zk_suffix` arms drive the accumulator
  loop of the hiding prover, not this driver.
- `whir/benches/zk_overhead.rs`'s tighter `whir_zk_open_no_pow` group panics at setup
  on `main` (`RandomnessExceedsSlack { round: 1, randomness: 256, slack: 240 }` at
  `n = 20`). That is a pre-existing failure, unrelated to this change, and worth its
  own issue.

## Scope

- **changed:** the hiding residual driver defers its bindings; the claim reduction and
  the binding become separate steps.
- **new:** four crate-private helpers on the prover, shared by both drivers; a
  differential test for this driver; a criterion arm that drives it directly.
- **removed:** a thin wrapper that no longer had a caller.
- **untouched:** the accumulator loop of the hiding prover, which drives its rounds
  off precomputed accumulators and never passes over the tables.

## Test plan

- [x] `deferred_binding_drives_the_same_hiding_rounds_as_binding_on_the_spot`: the
      driver's bound tables and claim match a bind-on-the-spot replay of the same
      challenges, over five shapes and both binding orders, with a non-zero auxiliary
      claim. Mutation-checked: dropping the final binding fails it under `--release`,
      where the debug assertion is gone.
- [x] `cargo test -p p3-sumcheck --features parallel` under `+avx2`, `+avx512f` and no
      target features — 286 tests each. The hiding verifier and simulator suites
      replay the whole transcript, so a drift in any round polynomial fails them.
- [x] `cargo test -p p3-whir` — proves and verifies through the hiding PCS.
- [x] `cargo clippy -p p3-sumcheck --all-targets --features parallel`.
- [x] `cargo +nightly fmt --all -- --check`.

### DIFF
--- a/sumcheck/src/strategy.rs
+++ b/sumcheck/src/strategy.rs
@@ -905,17 +905,45 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
         self.poly.eval(point)
     }
 
-    /// Computes the current round's plain quadratic coefficients without
-    /// touching the transcript or folding the polynomial.
-    pub(crate) fn round_coefficients(&self) -> (EF, EF) {
-        self.poly.round_coefficients()
+    /// Measures the current round, first applying a binding held back from the last one.
+    ///
+    /// A held-back binding is absorbed into the measuring pass.
+    /// The round then reads its tables once instead of twice.
+    ///
+    /// The slot is cleared here.
+    /// The caller puts this round's own challenge back into it.
+    pub(crate) fn measure_round(&mut self, pending: &mut Option<EF>) -> (EF, EF) {
+        match pending.take() {
+            // A challenge is waiting, so bind and measure in one pass.
+            Some(r) => self.poly.fold_round_coefficients(r),
+            // Nothing waiting, so this is a plain measuring pass.
+            None => self.poly.round_coefficients(),
+        }
     }
 
-    /// Folds the residual product polynomial by one challenge and updates the
-    /// claimed sum with the same quadratic extrapolation as the plain path.
-    pub(crate) fn fold_round_with_coefficients(&mut self, c0: EF, c_inf: EF, gamma: EF) {
+    /// Applies a binding that no measuring pass absorbed.
+    ///
+    /// The last round of a batch has no successor to fuse with.
+    /// Any later reader of the tables must still see them bound.
+    pub(crate) fn bind_pending(&mut self, pending: Option<EF>) {
+        if let Some(r) = pending {
+            self.poly.fold_round(r);
+        }
+    }
+
+    /// Advances the running claim to the round polynomial at the challenge.
+    ///
+    /// This is the quadratic extrapolation through 0, 1 and infinity the verifier applies.
+    /// The binding itself is left to the caller.
+    pub(crate) fn reduce_claim_with_coefficients(&mut self, c0: EF, c_inf: EF, gamma: EF) {
         self.sum = extrapolate_01inf(c0, self.sum - c0, c_inf, gamma);
-        self.poly.fold_round(gamma);
+    }
+
+    /// Asserts that the claim is the inner product of the bound pair.
+    ///
+    /// Only meaningful once every binding has been applied.
+    /// A driver that holds bindings back therefore calls this at the end, not per round.
+    pub(crate) fn debug_assert_claim(&self) {
         debug_assert_eq!(self.sum, self.poly.dot_product());
     }
 
@@ -988,10 +1016,7 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
 
         for _ in 0..folding_factor {
             // Measure this round, absorbing whatever binding the last one left behind.
-            let (c_a, c_inf) = match pending {
-                Some(r) => self.poly.fold_round_coefficients(r),
-                None => self.poly.round_coefficients(),
-            };
+            let (c_a, c_inf) = self.measure_round(&mut pending);
 
             // Commit to the transcript, do the optional grinding, take the challenge.
             let r = sumcheck_data.observe_and_sample(challenger, c_a, c_inf, pow_bits);
@@ -1006,12 +1031,10 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
         }
 
         // The last challenge has no successor to fuse with, so it binds on its own.
-        if let Some(r) = pending {
-            self.poly.fold_round(r);
-        }
+        self.bind_pending(pending);
 
         // Invariant: the claim is the inner product of the bound pair.
-        debug_assert_eq!(self.sum, self.poly.dot_product());
+        self.debug_assert_claim();
 
         Point::new(challenges)
     }
--- a/sumcheck/src/strategy.rs
+++ b/sumcheck/src/strategy.rs
@@ -925,8 +925,10 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
     ///
     /// The last round of a batch has no successor to fuse with.
     /// Any later reader of the tables must still see them bound.
-    pub(crate) fn bind_pending(&mut self, pending: Option<EF>) {
-        if let Some(r) = pending {
+    ///
+    /// The slot is cleared, so a second call binds nothing.
+    pub(crate) fn bind_pending(&mut self, pending: &mut Option<EF>) {
+        if let Some(r) = pending.take() {
             self.poly.fold_round(r);
         }
     }
@@ -1031,7 +1033,7 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
         }
 
         // The last challenge has no successor to fuse with, so it binds on its own.
-        self.bind_pending(pending);
+        self.bind_pending(&mut pending);
 
         // Invariant: the claim is the inner product of the bound pair.
         self.debug_assert_claim();
--- a/sumcheck/src/zk/prover/residual.rs
+++ b/sumcheck/src/zk/prover/residual.rs
@@ -118,13 +118,22 @@ where
         let half = EF::TWO.inverse();
         let mut aux_carry = aux_claim;
 
+        // A challenge is not applied on the spot.
+        // It is handed to the next round, which binds and measures in one pass.
+        //
+        // Everything the loop does between the two is scalar work.
+        // Assembling the round polynomial, the transcript, the grinding and the mask
+        // evaluation never read the tables.
+        let mut pending: Option<EF> = None;
+
         for (round_idx, mask) in masks.iter().enumerate() {
             let j = round_idx + 1;
             let mask_endpoints = mask[0].double() + mask[1..].iter().copied().sum::<EF>();
             sum_future_endpoints -= mask_endpoints;
             aux_carry *= half;
 
-            let (plain_c0, plain_c_inf) = self.round_coefficients();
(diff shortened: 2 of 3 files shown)


sumcheck/src/zk/prover/residual.rs
```diff
@@ -151,10 +160,20 @@ where
...
+            self.reduce_claim_with_coefficients(plain_c0, plain_c_inf, gamma);
+            pending = Some(gamma);
+
             rs.push(gamma);
         }
 
+        // The last challenge has no successor to fuse with.
+        // The weight scaling below reads the tables, so it binds here.
+        self.bind_pending(pending);
+
```
Nashtare, line 173: **MEDIUM — the hiding driver got the deferral but not the pinning test #2036 wrote for the plain one.**

#2036 added `deferred_binding_drives_the_same_rounds_as_round_at_a_time`, which pins every wire against a bind-on-the-spot reference across 1/2/4/9 variables and both orders. This commit applies the same deferral to a second driver and adds nothing.

In-crate coverage for `into_zk_sumcheck` is a single fixture: `n=3, k=2`, unpacked, `Prefix`, `aux_claim = 0`, `pow_bits = 0`. That misses the packed representation, the rayon branch above `PAR_THRESHOLD`, the `k == num_variables` terminal case, and the `Suffix` fallback inside `fold_round_coefficients` — all of which this driver reaches in production via `whir/src/pcs/zk/prover/mod.rs`. Those paths are covered only transitively by `p3-whir` end-to-end tests, which fail loudly but do not localise.

For the record the fusion is correct: I diffed this head against the pre-#2036 base `6e0ddf64` over 24 configurations — `(n_vars, folding_factor)` ∈ {(1,1),(2,1),(2,2),(3,2),(4,4),(9,3),(9,9),(12,5),(13,4),(17,4),(18,6),(19,1)} × {Prefix, Suffix}, non-zero `aux_claim`, `--features parallel`, debug profile — digesting all `round_coefficients` wires, all `rs` challenges, `eps`, the final `claimed_sum()` and both final bound tables. Byte-identical in every case. Since `rs` and `eps` are Fiat–Shamir outputs, that pins transcript evolution, not just the final claim.

Suggested fix: mirror #2036's test shape here — run `into_zk_sumcheck` twice from the same seeded transcript, once through a bind-on-the-spot reference loop, and assert the wires, `randomness`, and the residual prover's `evals()`/`weights()` all match. This is a missing regression guard, not a defect.

sumcheck/src/strategy.rs
```diff
@@ -657,17 +881,45 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
+            // Nothing waiting, so this is a plain measuring pass.
+            None => self.poly.round_coefficients(),
+        }
+    }
+
+    /// Applies a binding that no measuring pass absorbed.
+    ///
+    /// The last round of a batch has no successor to fuse with.
+    /// Any later reader of the tables must still see them bound.
+    pub(crate) fn bind_pending(&mut self, pending: Option<EF>) {
```
Nashtare: **MEDIUM — the "bind before anyone reads the tables" contract is unenforced and silent in release.**

`bind_pending` takes `Option<EF>` **by value**, and `Option<EF>` is `Copy`, so the caller's `pending` still reads `Some(gamma)` after the call. That is asymmetric with `measure_round(&mut Option<EF>)`, which `take()`s it.

The correctness of the scheme now rests on a convention no type enforces. `into_zk_sumcheck` hands `residual_prover` straight to `whir/src/pcs/zk/prover/mod.rs:231` (`batch.residual_prover.evals()`) and `:409` (`accumulate_claim`), both of which read the tables. If a later edit inserts a table read inside the loop, adds an early return, or calls `bind_pending` twice, the proof is silently wrong — the only tripwire is `debug_assert_claim()`, compiled out in release.

Today's code is correct; I confirmed nothing between `measure_round` and `bind_pending` touches `self` (`round_ctx.assemble`, `round_poly_to_wire`, the transcript, `grind` and the mask Horner are all scalar and take no prover reference).

Suggested fix: take `&mut Option<EF>` and `.take()` it. Optionally also move `pending` into `SumcheckProver` and `debug_assert!(self.pending.is_none())` in `evals()`, `weights()`, `dot_product()`, `scale_weights_and_claim()` and `accumulate_claim()`, so a stale binding is caught at the reader rather than at a distant invariant check.

sumcheck/src/zk/prover/residual.rs
```diff
@@ -118,13 +118,22 @@ where
         let half = EF::TWO.inverse();
         let mut aux_carry = aux_claim;
 
+        // A challenge is not applied on the spot.
+        // It is handed to the next round, which binds and measures in one pass.
+        //
+        // Everything the loop does between the two is scalar work.
+        // Assembling the round polynomial, the transcript, the grinding and the mask
+        // evaluation never read the tables.
+        let mut pending: Option<EF> = None;
```
Nashtare, line 127: **MEDIUM — perf claim carries no measurement for this driver.**

The PR body says the existing benches cannot resolve this change and declines to add one, borrowing #2036's 1.7–1.9x kernel number instead. That reasoning is sound — same kernel, same tables, same shape — and I found no path where the deferral can be slower: at `folding_factor == 1` the fused branch is never taken and the cost is identical, and for `k >= 2` each round trades a bind pass plus a measure pass for one fused pass. So the realistic downside is "no-op", not "regression".

Still, given this repo's history of perf changes that measured as regressions, landing an unmeasured perf claim in a second driver is worth an explicit note. A `criterion` arm calling `into_zk_sumcheck` directly at `n = 20, k = 4` would settle it in one run.

Nashtare (approved):

sumcheck/src/zk/prover/residual.rs
tcoratger: Added in 464cf4ad, mirroring the shape you asked for.

`deferred_binding_drives_the_same_hiding_rounds_as_binding_on_the_spot` runs the driver, then replays the challenges it produced through a bind-on-the-spot reference loop built from the same crate-private helpers, and compares both bound tables and the claim. The claim is the strong half: it chains every rounds `(c0, c_inf)` through the extrapolation, so a drift in any round message shows up in it.

The matrix covers exactly the gaps you listed, in both binding orders:

| `(n_vars, folding)` | what it reaches |
|---|---|
| `(2, 1)` | one round, so the fused branch is never taken |
| `(4, 4)` | `k == num_variables`, the terminal case |
| `(9, 3)` | packed storage, then the unpacking handoff |
| `(9, 9)` | packed storage bound all the way down |
| `(15, 3)` | large enough to reach the threaded branch of the fused pass |

`aux_claim` is non-zero throughout, so a carry leaking into the binding would show.

I mutation-checked it rather than assuming it bites: dropping the final `bind_pending` fails it, **including under `--release`**, where `debug_assert_claim` is compiled out. That was the specific worry in your `bind_pending` comment, so the guard is real rather than shadowing an assertion.

Thanks for the 24-configuration differential against `6e0ddf64` — that is a stronger check than what I have landed, and it is reassuring to have the transcript evolution pinned independently.

sumcheck/src/strategy.rs
tcoratger: Fixed in 464cf4ad — the asymmetry was real and unintended.

```rust
pub(crate) fn bind_pending(&mut self, pending: &mut Option<EF>) {
    if let Some(r) = pending.take() {
        self.poly.fold_round(r);
    }
}
```

Both helpers now take the slot by mutable reference and consume it, so a second call binds nothing and the caller cannot go on holding a challenge that has already been applied.

I did not take the larger half of the suggestion — moving `pending` into `SumcheckProver` and debug-asserting in every reader. It is the stronger design, but it changes the shape of both drivers and of five public accessors, which is more than a review fix on a stacked PR should carry. Worth its own change if you want it; happy to open one.

In the meantime the contract has a release-mode tripwire it did not have before: the new differential test fails under `--release` when the final binding is dropped, so a stale binding is caught by a test rather than only by an assertion that is compiled out.

sumcheck/src/zk/prover/residual.rs
tcoratger: Fair, and you were right to push. Measured in 464cf4ad.

`sumcheck/benches/sumcheck.rs` gains a `zk_residual` arm driving `into_zk_sumcheck` directly at `n = 20`, `k = 4`, with the mask code and Merkle commitment sized as the hiding WHIR prover sizes them. A/B by swapping only the deferral in the driver, two alternated pairs in one process, `-Ctarget-cpu=native`, `--features parallel`, 32 threads:

| | bind on the spot | deferred | speedup |
|---|---|---|---|
| pair 1 | 5.49 ms | 3.83 ms | 1.43x |
| pair 2 | 5.57 ms | 3.71 ms | 1.50x |

Both `p < 0.05`. So it is a real win on this driver, not the no-op floor your analysis bounded it at — and it is close to the plain drivers 1.7x once the mask commitment, which the fusion does not touch, is counted in the denominator.

Your reasoning about `folding_factor == 1` also holds in the bench harness: that shape never takes the fused branch, which is why it is one of the configurations in the new differential test rather than in the benchmark.

The PR body has been rewritten — it no longer declines to measure.

PR #2041, opened 2026-09-06
perf(stir): committed round-0 oracle with single-row input openings

Nashtare:
## Summary

STIR's PCS (`TwoAdicStirPcs`) previously fed STIR an *external* reduced-opening codeword that STIR never committed, so each round-0 query opened a whole fiber (`2^log_starting_folding_factor` input rows) from the input Merkle trees. This mirrors STIR's own round-0 fibers instead: STIR now commits its round-0 codeword itself and opens exactly one uniformly-sampled input row per round-0 query, the way FRI reads one input row per query.

- `stir/src/prover.rs` / `stir/src/verifier.rs`: expose the round-0 query draws (in transcript order, repeats included) alongside the existing sorted-unique index list, so the PCS can sample one random lane per draw after the whole STIR transcript settles.
- `stir/src/pcs.rs`: `open` becomes `prepare_open` + `prove_buckets` (STIR commits its own initial oracle; input trees move to one LDE row per leaf); `verify` runs STIR first, then checks each queried lane's committed value against the reduced opening rebuilt from that one row.
- `stir/tests/stir.rs`, plus new unit tests in `pcs.rs`: proof-shape coverage, tamper tests (initial-oracle fiber, input row), a low-degree-but-not-the-reduced-opening rejection test, transcript-ordering and lane-indexing tests pinned against the actual committed layout, zero-intermediate-round bucket coverage, and restored multi-instance external-oracle coverage (`prove_stir_multi_from_external_codewords`/`verify_stir_multi_with_external_initial`, which lost their only in-crate callers in this change).
- `uni-stark/tests/fixtures/uni_stark_stir_v1.postcard`: regenerated in place for the new proof format (grows 658 → 6136 bytes on that tiny fixture — STIR's own extra commitment dominates at that scale; the design's payoff is on wide traces, see below).

**Soundness note:** the one-lane binding relies on *mutual* correlated agreement for the round-0 fold — a strictly stronger property than STIR's own round analysis needs (which only needs plain distance preservation, since its own queries stay on the fold domain). Under `SecurityAssumption::JohnsonBound` the crate already documents the mutual property as required, so this adds no new assumption there. Under `CapacityBound` (the regime the measured configuration below uses), `security/src/assumption.rs` currently documents only the plain variant — a real gap, now recorded explicitly in `stir/src/pcs.rs`'s module doc (widening `p3-security`'s own doc was out of scope for this PR; worth a follow-up).

## Measurements

KoalaBear + quartic extension, Keccak-f / Poseidon-2 permutations, 2^18 rows, `RUSTFLAGS="-C target-cpu=native -C opt-level=3"`, Apple M4 Pro:

| objective | PCS | proof bytes | prove | verify |
|---|---|---|---|---|
| keccak-f | fri | 1,363,329 | 2.58 s | 6.01 ms |
| keccak-f | stir | 1,188,440 | 1.81 s | 4.86 ms |
| poseidon-2 | fri | 790,393 | 681 ms | 3.80 ms |
| poseidon-2 | stir | 684,845 | 876 ms | 3.19 ms |

Keccak+STIR: **3,893,985 → 1,188,440 bytes (−69.5%)** vs. the pre-change baseline on this same configuration, verify time **8.91 → 4.86 ms (−45.4%)**, now smaller than FRI on the same trace.

## Test plan

- [x] `cargo test -p p3-stir` and `--features parallel` (161 tests total across lib/tests/utils, 0 failures)
- [x] `cargo test -p p3-uni-stark --test stir_fibonacci` (regenerated fixture round-trips, including the `MIN_TRACE_HEIGHT` edge case)
- [x] `cargo test -p p3-examples stir` (both KoalaBear Keccak and Poseidon2 STIR CLI configurations)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` (stable and nightly)
- [x] Manual CLI runs across Keccak-f/Poseidon-2 × FRI/STIR at 2^18 rows, all print `Proof Verified Successfully`; results table above


stir/src/pcs.rs
```diff
@@ -22,21 +22,28 @@
...
-//! input MMCS openings at exactly the positions STIR sampled. No hand-mirrored transcript
-//! replay is needed.
+//! those and the PCS parameters, so no part of it travels in the proof — run
+//! `verifier::verify_stir_multi_inner` on every bucket, then sample the same lanes,
+//! authenticate the single input rows, rebuild the reduced opening at each position from its
+//! row (replaying the same alpha-batching and `Combine`), and check it against the lane of the
+//! fiber STIR authenticated. Reading one lane per draw costs one input row per query instead
+//! of a whole fiber; the soundness of that binding is spelled out on `prove_buckets`.
+//!
+//! **Assumption**: because a lane check reaches below the round-0 fold domain, that binding
```
tcoratger, line 38: Everything here is right except one word, and it is load-bearing.

"the standard WHIR/ACFY conjecture" was accurate until recently. **MCA up to capacity has been refuted** — Crites–Stewart ([eprint 2025/2046](https://eprint.iacr.org/2025/2046)) refute WHIR's conjecture directly, and BCHKS25 Thm 1.6 rules out `n^τ`-bounded proximity gaps past Johnson specifically in characteristic 2.

So the delta this PR makes to the two regimes is asymmetric:

- **Johnson** — unchanged, and your note is exactly right. MCA up to Johnson is now an *unconditional theorem* for `RS[F_q, D, k]` over arbitrary `F_q` and arbitrary domain of definition: BCHKS25 Thm 4.6, and independently Haböck [eprint 2025/2110](https://eprint.iacr.org/2025/2110) Thm 2, whose proof runs through a Guruswami–Sudan decoder over `F_q(Z)` and so is characteristic-agnostic. That is firmer than `p3-security`'s current wording ("Requires mutual correlated agreement up to the Johnson bound") suggests.
- **Capacity** — this moves it from resting on an unproven conjecture to resting on a refuted one.

The reason that is more than a doc nit: `CapacityBound` is what every shipped STIR configuration uses — both arms of `examples/src/proofs.rs`, `uni-stark/tests/stir_fibonacci.rs`, and the WHIR wiring those were modelled on. The demo config is the one that lands on refuted ground, and after this change it needs a strictly stronger property than it needed before.

Not blocking — the code is correct and the regime is chosen elsewhere, not here. But I'd redirect the follow-up you flagged: rather than widening the capacity doc to mention the mutual variant, move the shipped configurations to `JohnsonBound` and correct `p3-security`'s Johnson wording, which currently undersells a theorem.

tcoratger (approved):
Reviewed this as a protocol change rather than a wiring change, because that is what it is. **The code is right.** One thing in the soundness note is out of date, and it matters more than the note suggests — that is the only comment I've left.

## What I verified

**Index arithmetic and prover/verifier symmetry.** `position = j + lane * 2^(log_h - log_arity0)` puts the fiber index in the low bits and the lane in the high bits, and `split_position(p, fold_height0) = (p % fold_height0, p / fold_height0)` inverts it exactly. Prover (`pcs.rs:942-944`) and verifier (`pcs.rs:1464-1467`) call the same three helpers with the same arguments, so there is no second implementation to drift.

**Lane sampling is ordered correctly.** `sample_lanes` runs *after* `prove_stir_multi_from_codewords` returns — the whole STIR transcript has settled and the round-0 fibers are committed and authenticated before any lane is drawn. A prover cannot pick which lane to be honest on. This is the hinge of the design and it is on the right side of the transcript.

**One lane per draw, not per unique fiber.** The test at `pcs.rs:2560` pins this the right way: it samples both ways and asserts the per-draw transcript matches the implementation *and* that the per-unique one does not. That second assertion is what makes the first mean anything, and it is the desync bug this design could most easily have had.

**No panic path on adversarial input.** The lane read is `.get(lane).copied() != Some(value)`, so an out-of-range lane is a typed rejection. The one `expect` in that loop is on `first_round_indices`, which is verifier-derived from its own transcript replay rather than proof-supplied — draws are a subset of their own sorted-unique list by construction, so it cannot fire.

**The negative tests actually hold the binding up.** I stubbed the lane check to `if false && ...` and re-ran: `verify_rejects_a_low_degree_initial_oracle_that_is_not_the_reduced_opening` and `..._at_a_zero_round_bucket` both fail, and nothing else does. So the two tests written for this attack are the two that cover it, and they are load-bearing.

`cargo test -p p3-stir`: 187 passed across lib, integration and utils. CI 21/21.

## The one finding

Your assumption note calls mutual correlated agreement up to capacity "the standard WHIR/ACFY conjecture." That was accurate until recently — **it has since been refuted.** Crites–Stewart ([eprint 2025/2046](https://eprint.iacr.org/2025/2046)) refute WHIR's MCA-up-to-capacity conjecture directly, and BCHKS25 Thm 1.6 rules out `n^τ`-bounded proximity gaps past Johnson in characteristic 2.

That changes what this PR does to the `CapacityBound` configuration. Before it, that regime rested on an unproven conjecture. After it, it rests on a *refuted* one. Under `JohnsonBound` nothing changes and your note is exactly right — MCA up to Johnson is now an unconditional theorem for Reed-Solomon over an arbitrary field and an arbitrary domain (BCHKS25 Thm 4.6, and independently Haböck [eprint 2025/2110](https://eprint.iacr.org/2025/2110) Thm 2, via a Guruswami-Sudan decoder over `F_q(Z)`), so the Johnson path is on firmer ground than `p3-security`'s own wording currently admits.

The reason this is worth more than a doc follow-up: `CapacityBound` is what every shipped STIR configuration uses — `examples/src/proofs.rs` both arms, `uni-stark/tests/stir_fibonacci.rs`, and the WHIR wiring it was modelled on. So the default demo config is the one that lands on refuted ground, and it now needs a strictly stronger property than it did yesterday.

I don't think this blocks the merge — the code is correct, the regime choice is made elsewhere, and `JohnsonBound` is a one-field change for anyone who cares. But the follow-up you flagged should probably be "move the shipped configurations to Johnson and correct `p3-security`'s wording", not "widen the capacity doc to mention the mutual variant."

Merging. Nothing else from me — the rest of this is clean, and the transcript-ordering test in particular is the kind of test that is easy to skip and expensive to be without.

PR #2042, opened 2026-09-07
perf(binary-field): squaring, Frobenius, dot products, and buffer reuse

Nashtare:
## Summary
- Faster squaring at every tower level via direct linear maps and GF256 square/inverse tables, rather than routing through general multiplication
- Direct Frobenius and square-root maps instead of repeated squaring
- Batched/fused dot-product kernels (GF32 polynomial backend, fused reduction + basis conversion) for the additive-transform hot path
- Aligned-buffer coefficient copies and integer-representation zero-vector allocation to cut incidental allocation/copy overhead
- Challenger: reuse grinding worker state and hash buffers, and batch byte observation/sampling, across repeated grinds

## Test plan
- [x] `cargo test -p p3-binary-field` — all passing (171 unit + 4 integration)
- [x] `cargo clippy -p p3-binary-field --all-targets -- -D warnings` — clean
- [x] Benchmarked against `main` (`cargo bench -p p3-binary-field`): squaring paths measure -48% to -82%, inversion -5.4%, consistent with the targeted optimizations


tcoratger:
Ran the before/after myself since the description quotes percentages without the numbers behind them. Posting the table so it is on the record.

**Setup.** AMD Ryzen 9 9950X3D, `rustc 1.98.0`, `RUSTFLAGS="-C target-cpu=native"`, `--measurement-time 2 --warm-up-time 1`. Baseline is this PR's merge-base (`3da160d0`), measured in a separate worktree. Only the benchmark groups that exist **identically in both trees** are compared, so this is the same bench code on both sides — the groups this PR adds (`maps/128`, `dot_product`, `batch/poly128`, `bulk/*`, `mixed/128`) have no baseline to compare against and are excluded. Medians.

### Squaring

| benchmark | before | after | change |
|---|---|---|---|
| `square/8/latency` | 1.808 µs | 1.158 µs | **−36.0%** |
| `square/8/throughput` | 734.3 ns | 130.0 ns | **−82.3%** |
| `square/16/latency` | 5.293 µs | 1.678 µs | **−68.3%** |
| `square/16/throughput` | 1.768 µs | 229.7 ns | **−87.0%** |
| `square/32/latency` | 7.797 µs | 6.722 µs | −13.8% |
| `square/32/throughput` | 4.791 µs | 3.738 µs | −22.0% |
| `square/64/latency` | 2.699 µs | 2.654 µs | −1.7% |
| `square/64/throughput` | 2.539 µs | 2.413 µs | −5.0% |
| `square/128/latency` | 14.39 µs | 7.282 µs | **−49.4%** |
| `square/128/throughput` | 6.662 µs | 2.962 µs | **−55.5%** |
| `representations/square/tower` | 14.33 µs | 7.399 µs | **−48.4%** |

### Inversion

| benchmark | before | after | change |
|---|---|---|---|
| `inverse/128/dispatched` | 103.6 µs | 84.26 µs | **−18.6%** |
| `representations/inverse/through tower` | 130.4 µs | 106.2 µs | **−18.6%** |

### Controls — paths this PR does not touch

These are the reason the numbers above are believable rather than a machine-state artifact. Every untouched path sits inside noise:

| benchmark | before | after | change |
|---|---|---|---|
| `square/8/mul/latency` | 1.872 µs | 1.900 µs | +1.5% |
| `square/128/mul/latency` | 21.76 µs | 21.62 µs | −0.7% |
| `representations/square/ghash` | 4.624 µs | 4.622 µs | −0.0% |
| `representations/inverse/ghash` | 91.19 µs | 90.58 µs | −0.7% |
| `ghash/sqrt/127 squares` | 475.3 µs | 474.6 µs | −0.2% |

### Two notes

- **Inversion does much better here than the description claims.** You quote −5.4%; I measure **−18.6%**, consistently across both inversion benchmarks. Different microarchitecture (Zen 5 vs M4 Pro), so both can be right — but the number in the description undersells the change on x86_64.
- **The 32- and 64-bit squaring gains are smaller than the quoted −48%/−82% band.** `square/32` lands at −14%/−22% and `square/64` at −2%/−5%. That tracks the implementation: there is no `square_32` table map (32-bit squaring routes through `byte_square`), and 64-bit squaring already had a fast path. The headline band holds for 8, 16 and 128.

Raw criterion output for both runs is reproducible with:

```
cargo bench -p p3-binary-field --bench arithmetic -- 'square|inverse' \
    --measurement-time 2 --warm-up-time 1
```

tcoratger (approved):
Reviewed this as a correctness change that happens to be about speed, because that is the risk profile: a wrong squaring table or a leaked grinding-worker state is a silently wrong proof, not a slow one. **I found nothing wrong.** Benchmark table posted separately.

## What I checked

**Every new fast path is pinned to an independent oracle.** `square_16/64/128` are proptested against `reference_mul` (`linear.rs:214`); `sqrt_16/32/64/128` and the Frobenius maps against repeated reference squares across all four tower levels (`linear.rs:186`); `poly_mul_128_batch` against the generic `poly_mul` (`clmul/mod.rs:323`). There is no fast path here whose only witness is itself.

**The fused dot products are genuinely covered.** I sabotaged `dot_product_128` to `return 1` and re-ran: `fused_dot_products_match_reference_for_arbitrary_lengths` fails. Worth noting for anyone reproducing that — it only fails with the hardware path compiled in. A plain `cargo test -p p3-binary-field` on x86_64 builds 143 tests and misses it; with `-C target-cpu=native` it builds 161 and catches it. CI covers this through the `+pclmulqdq` and `+vpclmulqdq` legs.

**Both new `unsafe` blocks in `tower.rs` are sound.**
- `zero_vec` reuses the `Vec<$repr>` allocation as `Vec<Self>`. That needs identical layout, and `tower.rs:135` is `#[repr(transparent)] pub struct $name($repr)`, so it holds.
- `reconstitute_from_base` copies rather than reinterprets: `Vec::<Self>::with_capacity(len)` reserves `len * size_of::<Self>()` bytes and the copy writes `vec.len() * size_of::<$lower>()`, which is the same count since one `Self` is `dimension` lowers. Source stays alive and is dropped normally, so no aliasing and no double free. The round-trip is tested at lengths 0, 1 and 19.

**The grinding worker reuse does not leak state between candidates.** This was the one I most expected to find a bug in. `worker.inner.clone_from(&self.inner)` resets before each candidate, and that is the *complete* mutable state — `BinaryChallenger` is `{ inner, _marker: PhantomData }`, and the new `HashChallenger::clone_from` resets all three of `input_buffer`, `output_buffer`, `hasher`. The `assert!(self.check_witness(bits, witness))` after the search still re-validates against the real challenger, so even a wrong reset would abort rather than emit a bad witness. `find_map_any` remains non-deterministic in *which* valid witness comes back, exactly as `find_any` was before — no regression, and the existing tests already account for it.

**The serial `map_init` fallback matches what the parallel use needs.** It calls `init` once and reuses, which differs from rayon's once-per-split — fine here precisely because the grind resets the worker per item rather than relying on its carried state. It has its own test for reuse, ordering and empty input.

`cargo test -p p3-binary-field`: 318 passed across lib, integration and the generic field suite. CI 21/21.

## One thing I checked and cleared

`clmul/aarch64.rs` — including the new `poly_mul_128_batch` NEON code — is gated on `target_feature = "aes"`, and CI's two aarch64 legs run `+sve2` with `compile_only: true`, which had me expecting new `unsafe` SIMD with no CI coverage. It is covered: `aes` is in the `aarch64-apple-darwin` baseline, and the `macos-latest` legs are not compile-only, so the PMULL backend and its proptest run there on every push. No gap.

Approving and merging. The measured numbers are in the comment below — inversion does considerably better on x86_64 than the description claims, and the 32/64-bit squaring gains are smaller than the quoted band, both worth folding into the description if you revise it.

PR #2043, opened 2026-09-07
perf(binary-dft): incremental twiddles, tiled stages, and fused kernels

Nashtare:
## Summary
Stacked on #2042 (binary-field) — this PR's diff is the 10 binary-dft-specific commits on top of it; it will shrink to just those once #2042 merges.

- Generate the polynomial-basis twiddle schedule incrementally (`Twiddles`) instead of rebuilding each stage's twiddles from scratch
- Keep adjacent small stages resident in cache-sized tiles rather than round-tripping through memory per stage
- Consume fused constant-twiddle field kernels and specialize portable twiddle products by subfield
- Return identity low-degree extensions directly, encode cosets in the destination allocation, and extend only new polynomial-basis cosets, avoiding redundant copies
- Size parallel work to actual conversion/butterfly cost instead of a fixed threshold

## Test plan
- [x] `cargo test -p p3-binary-dft` — all passing (42 unit + 2 integration)
- [x] `cargo clippy -p p3-binary-dft --all-targets -- -D warnings` — clean
- [x] Benchmarked against `main` (`cargo bench -p p3-binary-dft`): `ntt/128/hybrid` (the `PolyBasisNtt` path these commits target) measures -3.7% to -6.2% consistently across all four domain sizes tested

### DIFF
--- a/binary-dft/benches/ntt.rs
+++ b/binary-dft/benches/ntt.rs
@@ -100,5 +100,113 @@ fn bench_encode(c: &mut Criterion) {
     }
 }
 
-criterion_group!(benches, bench_ntt, bench_encode);
+/// Direct polynomial-backend workloads, including small later-round domains.
+fn bench_poly(c: &mut Criterion) {
+    eprintln!(
+        "binary-dft: parallel={}, threads={}, hardware_clmul={}",
+        cfg!(feature = "parallel"),
+        p3_maybe_rayon::prelude::current_num_threads(),
+        p3_binary_field::poly_basis::HAS_HARDWARE_CLMUL
+    );
+    let mut group = c.benchmark_group("poly");
+    group.sample_size(10);
+    let mut rng = SmallRng::seed_from_u64(7);
+    let ntt = PolyBasisNtt::default();
+    let encoder = AdditiveRsEncoder::<BinaryField128>::default();
+    let shift = BinaryField128::from_repr(1 << 127);
+    for log_height in [4, 8, 12, 16] {
+        for width in [1, 4, 16, 64] {
+            let mat = RowMajorMatrix::<BinaryField128>::rand(&mut rng, 1 << log_height, width);
+            let parameter = format!("h{log_height}/w{width}");
+            for op in ["forward", "inverse", "shifted"] {
+                group.bench_with_input(BenchmarkId::new(op, &parameter), &mat, |b, mat| {
+                    b.iter_batched(
+                        || mat.clone(),
+                        |m| match op {
+                            "forward" => ntt.ntt_batch(m),
+                            "inverse" => ntt.intt_batch(m),
+                            _ => ntt.shifted_ntt_batch(m, shift),
+                        },
+                        BatchSize::PerIteration,
+                    );
+                });
+            }
+            for added in [0, 1, 2, 3] {
+                group.bench_with_input(
+                    BenchmarkId::new(format!("lde/r{added}"), &parameter),
+                    &mat,
+                    |b, mat| {
+                        b.iter_batched(
+                            || mat.clone(),
+                            |m| ntt.shifted_lde_batch(m, added, shift),
+                            BatchSize::PerIteration,
+                        );
+                    },
+                );
+                group.bench_with_input(
+                    BenchmarkId::new(format!("encode/r{added}"), &parameter),
+                    &mat,
+                    |b, mat| {
+                        b.iter_batched(
+                            || mat.clone(),
+                            |m| encoder.encode_batch(m, added),
+                            BatchSize::PerIteration,
+                        );
+                    },
+                );
+            }
+        }
+    }
+    group.finish();
+}
+
+/// The production layout, encoder and Merkle commitment together.
+fn bench_commit(c: &mut Criterion) {
+    use p3_binary_field::BinaryChallenger;
+    use p3_challenger::HashChallenger;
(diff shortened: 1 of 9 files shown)


tcoratger:
Ran the before/after since the description quotes a percentage band without the numbers. Posting the table, and flagging where my measurement disagrees with it.

**Setup.** AMD Ryzen 9 9950X3D (32 threads), `rustc 1.98.0`, `RUSTFLAGS="-C target-cpu=native"`, `--measurement-time 3 --warm-up-time 1`, medians with criterion's 95% interval.

**Baseline is `57023a71` — this branch's parent, i.e. #2042 already merged.** That matters: the merge-base against a stale `main` does *not* contain #2042, and benching against that would have credited the binary-field squaring work to this PR. Only the `ntt/` groups are compared, since they exist identically in both trees; the `poly` and `commit_base` groups this PR adds have no baseline.

### `ntt/128/hybrid` — the `PolyBasisNtt` path this PR targets

| log height | before | after | |
|---|---|---|---|
| 14 | 4.900 ms `[4.881, 4.922]` | 4.960 ms `[4.850, 5.110]` | inconclusive — intervals overlap |
| 16 | 23.83 ms `[23.64, 23.98]` | 23.81 ms `[23.15, 24.51]` | inconclusive — intervals overlap |
| 18 | 135.9 ms `[132.3, 139.5]` | 122.6 ms `[120.5, 124.9]` | **−9.7%**, no overlap |
| 20 | 607.0 ms `[600.3, 614.1]` | 557.9 ms `[537.1, 572.9]` | **−8.1%**, no overlap |

### Controls — paths this PR should not move

| benchmark | before | after | |
|---|---|---|---|
| `ntt/32/20` | 789.7 ms | 794.6 ms | +0.6% |
| `ntt/64/20` | 1.047 s | 1.057 s | +0.9% |
| `ntt/128/tower/20` | 2.149 s | 2.155 s | +0.3% |

All flat, which is what makes the two hybrid results above credible rather than machine drift.

### Two honest caveats

- **The shape differs from the description.** You report "−3.7% to −6.2% consistently across all four domain sizes." I measure the opposite shape: nothing separable from noise at 2^14 and 2^16, and roughly −8% to −10% at 2^18 and 2^20. That reads like a change whose win comes from cache and parallel behaviour, which only bites once the working set stops fitting — consistent with the tiling and the cost-model task sizing. Different machine and thread count from yours, so I'd treat "consistently across all four" as the claim worth re-checking rather than the magnitude.
- **`ntt/128/ghash` is too noisy here to report.** The baseline interval alone spans `[121.0, 154.8]` ms at 2^18 and the PR's spans `[77.9, 132.6]`. It *looks* like a large win, but I won't quote a number off intervals that wide. Worth re-running pinned to fewer threads if you want a figure for it.

Reproduce with:

```
cargo bench -p p3-binary-dft --bench ntt -- 'ntt/' --measurement-time 3 --warm-up-time 1
```

tcoratger (approved):
Reviewed this the same way as #2042 — as a correctness change that happens to be about speed. A wrong twiddle schedule or a tile that drops a butterfly is a silently wrong codeword, not a slow one. **I found nothing wrong.** Benchmark table posted separately, with one disagreement about the *shape* of the win.

## What I checked

**The path this PR rewrites is proptested against an independent oracle.** `PolyBasisNtt::shifted_ntt_batch` and `shifted_lde_batch` are checked against `NaiveAdditiveNtt` (`poly.rs:499`, `poly.rs:536`), which evaluates the definition directly and shares no machinery with the incremental schedule or the tiling. There are deterministic sweeps over widths and shifts on top of that.

**The incremental twiddle schedule is load-bearing and covered.** I replaced the recurrence `base = square(base) ^ base` with `base = square(base)` and re-ran: five tests fail, including `incremental_twiddles_match_independent_domain_points`, which pins the incremental schedule against a separately derived one rather than against itself. That is the right test to have written for this change, and it is the one I would have asked for if it were missing.

**The tiling is size-invariant, as it must be.** Tiling is a scheduling decision, so the transform has to be identical at any tile size. I forced `TILE_BYTES` to `1` — driving every input through the tiled path at minimum granularity — and the `poly` suite still passes. So the tile boundary is not carrying semantics.

**The identity LDE shortcut is genuinely the identity.** `added_bits == 0` returns the input, and `identity_lde_preserves_input_allocation` checks the returned pointer equals the input pointer, not merely that the values match — so it pins the no-copy claim, not just correctness. The two `should_panic` tests cover the rejected shapes.

`cargo test -p p3-binary-dft`: 48 unit + 3 integration, all passing (the integration leg takes ~2 min). Clippy `-D warnings` clean. CI green so far, 4 legs still running at time of writing.

## One thing worth a second look

Not a defect, and not blocking — but my measurements disagree with the description's *shape*, not just its magnitude. You report "−3.7% to −6.2% consistently across all four domain sizes." I get nothing separable from noise at 2^14 and 2^16, and roughly −8% to −10% at 2^18 and 2^20, with the untouched `ntt/32`, `ntt/64` and `ntt/128/tower` paths flat to within 1%.

That is a plausible and arguably better result — it is the profile you would expect from tiling plus cost-model task sizing, which only pay once the working set stops fitting in cache. But "consistently across all four" is the part I could not reproduce, and it is worth knowing which of us is the outlier before that sentence lands in the history. Different machine and thread count, so I would not assume mine is right.

Approving and merging.

PR #2046, opened 2026-09-07
fix(circle): account for transition selectors in quotient sizing

### LINKED ISSUE #575: Incorrect selector degree calculation in Circle Stark

Nashtare:
## Summary

Circle's transition selector is a full trace-space polynomial, but the shared quotient-sizing model treats it as degree zero. This undersizes the quotient for valid nonlinear transitions and causes `OodEvaluationMismatch`.

- Add a domain-specific transition degree and use it consistently in uni-stark and batch-stark proving, verification, and lookup preparation.
- Count every transition-selector factor, retain full Circle periodic-column degrees, and include both base and extension constraints even when an AIR supplies a degree hint.
- Keep lookup packing within the corrected quotient budget, including selectors in payloads, multiplicities, and flags.
- Preserve the existing two-adic cached/hint fast paths and memoize symbolic DAG traversal.

closes #575 

## Validation

The original nonlinear-transition regressions failed before the fix in both uni-stark and batch-stark, with and without hints. Added coverage also checks quotient bucket boundaries, repeated guards, extension constraints, periodic products, lookup packing, tampered openings, and rejection of invalid traces in release mode. Existing Circle and two-adic compatibility fixtures still verify.

Fresh release/parallel tests passed for the six affected crates, excluding two pre-existing tests that expect debug-only assertions to panic:

```sh
cargo test --offline --release \
  -p p3-air -p p3-commit -p p3-circle -p p3-lookup \
  -p p3-uni-stark -p p3-batch-stark \
  --features parallel --no-fail-fast -- \
  --skip generate_permutation_exclusive_rejects_non_boolean_flag \
  --skip generate_permutation_exclusive_rejects_two_active_flags
```

Both excluded failures were reproduced on unmodified `c239716c`; their code is unchanged. Nightly formatting checks also pass.

## Performance

A local M31/Keccak release microbenchmark compared this fix with `c239716c` using a sufficient manual degree hint on the baseline. For 16,384-row degree-3 and degree-4 recurrences, proving remained about 183 ms with identical serialized proof sizes. Fixed-branch verification measured approximately 0.41-0.44 ms. Measurements used nine timed iterations after two warmups; this is a focused regression check, not a general throughput claim.

### DIFF
--- a/air/src/air.rs
+++ b/air/src/air.rs
@@ -171,6 +171,10 @@ pub trait BaseAir<F>: Sync {
     /// Normally the prover runs a full symbolic evaluation to compute this.
     /// Overriding this method lets both the prover and verifier skip that
     /// pass when only the degree (not the full constraint list) is needed.
+    /// Domains with a full trace-space transition selector, such as Circle,
+    /// still infer the domain-specific degree and take the maximum of it and
+    /// this hint. The cached degree multiple used by this hint treats transition
+    /// selectors as degree zero and cannot bound their repeated products there.
     ///
     /// The value must be an upper bound on the degree multiple of every
     /// constraint (base and extension). It does not need to be tight, but
--- a/air/src/symbolic/expression.rs
+++ b/air/src/symbolic/expression.rs
@@ -49,6 +49,13 @@ impl<F: Field> SymLeaf for BaseLeaf<F> {
         }
     }
 
+    fn degree_multiple_with_transition(&self, transition_degree: usize) -> usize {
+        match self {
+            Self::IsTransition => transition_degree,
+            _ => self.degree_multiple(),
+        }
+    }
+
     fn poly_degree(&self, trace_len: usize, periodic_periods: &[usize]) -> usize {
         match self {
             Self::Variable(v) => v.poly_degree(trace_len, periodic_periods),
@@ -351,6 +358,17 @@ mod tests {
 
         // The period-2 column has degree `N - N/2 = N/2`; each squaring doubles it.
         assert_eq!(expr.poly_degree(N, &[2]), (N / 2) << DEPTH);
+        assert_eq!(expr.degree_multiple_with_transition(1), 1 << DEPTH);
+    }
+
+    #[test]
+    fn degree_multiple_counts_every_transition_factor() {
+        let transition = SymbolicExpression::<BabyBear>::Leaf(BaseLeaf::IsTransition);
+        let main =
+            SymbolicExpression::from(SymbolicVariable::new(BaseEntry::Main { offset: 0 }, 0));
+        let expr = transition.clone().cube() * (main.cube() - transition);
+        assert_eq!(expr.degree_multiple_with_transition(0), 3);
+        assert_eq!(expr.degree_multiple_with_transition(1), 6);
     }
 
     #[test]
--- a/air/src/symbolic/expression_ext.rs
+++ b/air/src/symbolic/expression_ext.rs
@@ -48,6 +48,13 @@ impl<F: Field, EF: ExtensionField<F>> SymLeaf for ExtLeaf<F, EF> {
         }
     }
 
+    fn degree_multiple_with_transition(&self, transition_degree: usize) -> usize {
+        match self {
+            Self::Base(e) => e.degree_multiple_with_transition(transition_degree),
+            _ => self.degree_multiple(),
+        }
+    }
+
     fn poly_degree(&self, trace_len: usize, periodic_periods: &[usize]) -> usize {
         match self {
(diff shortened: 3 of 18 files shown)


tcoratger (approved):
Reviewed the fix for what it changes about *soundness direction*, since undersizing a quotient and oversizing one fail in opposite ways: the bug you fixed produced `OodEvaluationMismatch` (a loud, safe failure), while a fix that undercounts anywhere would be the silent kind. **It doesn't undercount. Nothing wrong found.**

## The model is right

`is_transition = 1 - z_inv_dk` in `circle/src/domain.rs` really is a full trace-space polynomial — it is one minus a normalized last-row selector, so its degree grows with the trace, and it costs a whole trace-column unit. The two-adic selector is degree one absolutely, which is zero units against a trace column. So `transition_degree_multiple()` returning `1` for Circle and `0` by default is the correct split, and the doc comment states it accurately.

## It cannot undercount

Three separate reasons, each checked:

- **The traversal is an upper bound.** `Add`/`Sub` take the max, `Mul` takes the sum, `Neg` passes through. Cancellation can only make the true degree lower, never higher.
- **The hint can only raise the estimate.** `.max(air.max_constraint_degree().unwrap_or(0))` means an AIR that supplies a hint still gets the symbolic degree computed and the larger of the two used. That is the fix for the hint path, and it is monotone in the safe direction — a wrong hint now costs proof size, not soundness.
- **Every transition factor is counted, not just one.** `transition.cube() * (main.cube() - transition)` scores 3 at weight 0 and 6 at weight 1, which is what the test pins. A model that counted the selector once would give 4.

## Prover and verifier cannot diverge

Both provers and both verifiers route through `get_log_num_quotient_chunks_for_domain` — `uni-stark/src/prover.rs:191` and `verifier.rs:337`, `batch-stark/src/prover.rs:194` and `verifier/mod.rs:431`. The function is a pure function of `(air, layout, domain, is_zk)`, all of which both sides hold, so there is no path where the two size the quotient differently.

## The memoization is sound

Caching on `*const Self` is safe here specifically because the cache is built fresh inside each `degree_with` call, so two traversals with different leaf-degree closures never share entries; the nodes are immutable; and every `Arc` stays alive for the traversal, so there is no address-reuse hazard. Structurally-equal nodes at distinct addresses are recomputed, which costs time and not correctness.

## The tests are load-bearing

I reverted `transition_degree_multiple` to `0` for Circle, reintroducing the bug. `circle_nonlinear_transition_with_hint` and `circle_nonlinear_transition_without_hint` both fail. Having both is the right call — the hinted one is what exercises the `.max(hint)` path, and a fix that only handled the unhinted case would pass a suite that tested one of them.

## Lookup packing

Threading `transition_degree` into `constraint_degree_with_transition` picks up selectors inside payloads, multiplicities and flags, and the `max(1 + deg(U), deg(V))` bound for `U * f - V` is the right shape, with the `+1` accounting for the running-sum column multiplying `U`.

354 tests pass across `p3-air`, `p3-commit`, `p3-circle`, `p3-lookup`, `p3-uni-stark` and `p3-batch-stark`, zero failures. Workspace clippy `--all-targets -D warnings` and nightly fmt clean. The two release-mode `should_panic` lookup tests you excluded are genuinely pre-existing — both are on `main` and untouched here.

No follow-up needed; I did not find anything worth one. Merging.

PR #2047, opened 2026-09-07
feat(multi-stark): wire binary fields and PCS end to end

Nashtare:
The multilinear STARK prover could not use `BinaryField128` and `BinaryPcs`: its typed sumcheck transcript required a 64-bit prime field, and AIR round kernels advanced interpolation nodes by adding one. This adds a complete binary AIR example that proves a nonlinear recurrence, serializes the proof, and verifies it with a fresh transcript.

## Changes

- Add `TranscriptField` hooks and a `BinaryField128` implementation for injective domain-separator absorption, canonical wire encoding, and stable field tags. Preserve existing prime-field transcript encodings.
- Advance all four base/extension, scalar/packed AIR round kernels through the field's distinct interpolation nodes, retaining addition-only stepping when consecutive nodes differ by one.
- Expose retained tables and cloning on binary PCS prover data, and connect binary fields to multi-STARK configuration and sumcheck APIs.
- Add a runnable example and tests for honest proofs, tampering, mixed-height batches, and reuse of preprocessed commitments.
- Explicitly reject active characteristic-two LogUp declarations, including zero-query-weight declarations. Binary LogUp is unsupported; the example uses ordinary AIR constraints and is not zero knowledge.

Run the example with:

```sh
cargo run --release -p p3-multi-stark --example prove_binary_field
```

## Validation

- The example verifies a serialized proof for 256 rows; all 8 E2E tests pass in serial and parallel modes.
- Multi-STARK and sumcheck release suites pass, including the existing WHIR compatibility fixture.
- Binary-field, binary-PCS, and challenger release tests pass. All 54 lookup tests pass in debug mode.
- Workspace Clippy with `--all-targets -- -D warnings`, nightly formatting, dependency sorting, TOML formatting, and whitespace checks pass.
- Documentation builds with broken-link errors enabled; two existing warnings remain in `challenger/src/multi_field_challenger.rs`.

The aggregate release test run hits two existing lookup `should_panic` tests that depend on `debug_assert!`: `generate_permutation_exclusive_rejects_non_boolean_flag` and `generate_permutation_exclusive_rejects_two_active_flags`. Their source and implementation are unchanged, and both pass in debug mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,20 @@ Hashes
 - [x] Monolith
 
 
+## Binary-field example
+
+The [binary AIR example](multi-stark/examples/prove_binary_field.rs) proves a nonlinear
+recurrence over `BinaryField128` using the multilinear STARK prover and `BinaryPcs`,
+with Keccak Merkle commitments and a binary challenger. It serializes the proof and
+verifies it with a fresh transcript:
+
+```bash
+cargo run --release -p p3-multi-stark --example prove_binary_field
+```
+
+This example is not zero knowledge. Characteristic-two LogUp lookups are unsupported
+and rejected; the example uses ordinary AIR constraints.
+
 ## Benchmarks
 
 Many variations are possible, with different fields, hashes, and so forth, which can be controlled through the command line.
--- a/binary-field/src/lib.rs
+++ b/binary-field/src/lib.rs
@@ -14,6 +14,7 @@ mod packed;
 pub mod poly_basis;
 mod tables;
 mod tower;
+mod transcript;
 
 pub use challenger::BinaryChallenger;
 pub use gf2::Gf2;
--- a/binary-field/src/transcript.rs
+++ b/binary-field/src/transcript.rs
@@ -1,0 +1,43 @@
+//! Canonical typed-transcript encoding for the 128-bit tower field.
+
+use alloc::vec::Vec;
+
+use p3_challenger::CanObserve;
+use p3_challenger::fs::{TranscriptError, TranscriptField, TypeTag};
+
+use crate::{BinaryField128, TowerLevel};
+
+impl TranscriptField for BinaryField128 {
+    fn algebra_tag(degree: usize) -> TypeTag {
+        TypeTag::BinaryTower { bits: 128, degree }
+    }
+
+    fn observe_seed<C: CanObserve<Self>>(challenger: &mut C, bytes: &[u8]) {
+        // Embed the length and chunks in the tower basis, not via `from_u128`,
+        // which embeds integers in GF(2) and would retain only their parity.
+        challenger.observe(Self::from_repr(bytes.len() as u128));
+        for chunk in bytes.chunks(16) {
+            let mut padded = [0; 16];
+            padded[..chunk.len()].copy_from_slice(chunk);
+            challenger.observe(Self::from_repr(u128::from_le_bytes(padded)));
+        }
+    }
+
+    fn wire_len() -> usize {
+        16
+    }
+
+    fn encode(value: &Self, out: &mut Vec<u8>) {
+        out.extend_from_slice(&value.to_repr().to_be_bytes());
+    }
+
+    fn decode(bytes: &[u8]) -> Result<Self, TranscriptError> {
+        let prefix = bytes.get(..16).ok_or(TranscriptError::BadProofShape {
+            reason: "not enough bytes for a canonical binary field encoding",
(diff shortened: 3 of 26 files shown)


tcoratger (approved):
Reviewed this as a soundness audit, because three of the four substantive pieces are places where a mistake is a silently accepted forged proof rather than a failing test: the Fiat-Shamir encoding, the characteristic-two interpolation nodes, and the LogUp rejection. **I found nothing wrong.** Every guard I checked is real, and every one is covered by a test that fails when I break the guard.

## Transcript encoding

The two things that matter for a native-field transcript are injective seed absorption and non-colliding domain separation. Both hold.

**Seed absorption is injective**, and the length prefix is what makes it so. Without it, `[0xaa]` and `[0xaa, 0]` both fold to the same last chunk — the padding is indistinguishable from real trailing zeros. I removed the prefix and `binary_seed_preserves_bytes_and_length_in_the_tower_basis` and `binary_seed_recovers_every_byte` both fail. That second one is the right shape for this: it *recovers every byte* from the encoding, which is a decoding proof, and a decodable encoding is injective by construction.

**Using `from_repr` rather than an integer embedding is load-bearing**, and the comment says why. Over `GF(2^128)` an integer embedding keeps only parity, so `[2]` and `[0]` would collide. The test pins exactly that pair. (I tried to mutate this into the bug and couldn't — `from_u128` does not exist on the type, so the mistake is not even expressible. Better still.)

**Domain separation between a tower element and an algebra over `GF(2)` holds.** Both carry 128 bits, so `Algebra { modulus: 2, degree: 128 }` and `BinaryTower { bits: 128, degree: 1 }` are the collision to worry about. They render as `Algebra(2^128)` and `BinaryTower(128^1)`, which cannot alias, and `binary_type_tag_commits_to_the_coefficient_field` asserts precisely that inequality rather than merely asserting the tag is "some tower tag".

**Prime-field behaviour is untouched.** The change is a bound relaxation from `PrimeField64` to `TranscriptField`, and the blanket `impl<F: PrimeField64>` keeps the existing big-endian wire format and length-prefixed packing. `verify_whir_compat_fixture` passes against a fixture this PR does not modify — that is the real evidence that existing proofs still verify, and it is stronger than reading the diff.

## Interpolation nodes

The identity is right. A multilinear along one variable is `f(x) = f(0) + x * (f(1) - f(0))`, so stepping from node `n_k` to `n_{k+1}` adds `(n_{k+1} - n_k) * diff`. The old code baked in `n_{k+1} - n_k == 1`, which is true exactly when `interpolation_node(i) = i` — and for a binary tower the nodes are `from_repr(i)`, where `node(1) = 1` but `node(2) - node(1) = from_repr(3)`, nowhere near one. So the scaled path is genuinely exercised from the second step onward rather than being dead code.

The `if let Some(&step) = node_steps.get(node)` guard also reproduces the old `if node != degree` boundary exactly, since the step vector has `degree` entries.

I forced `interpolation_steps` to return `ONE` throughout. All four kernels fail — `binary_{base,extension}_{scalar,packed}_uses_distinct_interpolation_nodes`. One test per kernel, all four load-bearing, none of them redundant.

## Characteristic-two LogUp rejection

`F::ONE + F::ONE == F::ZERO` is the right detector, and gating on `!lookup.elements.is_empty()` rather than on query weight is the right notion of "active". The zero-query-weight case is the one that would slip past a weight-based check, and `binary_lookup_plan_is_rejected_even_without_query_weight` covers it with a provided-only `+1`/`-1` declaration that sums to zero. Disabling the guard fails that test.

## Everything else

- `add_scaled` on the boundary selectors mirrors the node stepping consistently.
- `Clone` plus the `table` accessor on the binary PCS prover data is what makes preprocessed-commitment reuse possible; opening one commitment twice under two independent transcripts is fine.
- The example runs: `Verified 256 rows over GF(2^128) using BinaryPcs: 74811 bytes, prove 4.89ms, verify 1.35ms`.
- The two release-mode `should_panic` lookup tests you flagged are genuinely pre-existing — both exist on `main` and this PR does not touch either. Accurate disclosure.

461 tests pass across `p3-multi-stark`, `p3-binary-field`, `p3-challenger`, `p3-sumcheck`, `p3-binary-pcs` and `p3-lookup`, zero failures. Workspace clippy `--all-targets -D warnings`, nightly fmt, `cargo sort`, and `cargo machete` all clean.

Nothing minor enough to be worth a follow-up either. Merging.

PR #2048, opened 2026-09-07
fix(security,fri): accounting and grinding hygiene fixes

Nashtare:
## Summary

Security-accounting and grinding-hygiene fixes for `p3-security`, `p3-fri`, `p3-circle`, `p3-uni-stark`, and `p3-batch-stark`. Several of these are corrections to security *reports* that were silently omitting or overstating rounds a protocol actually pays for; others implement grinding sites the accounting already modeled but no protocol wired up; the rest close config gaps that let a caller build a PCS with an unsound or panic-prone configuration.

## What this fixes

**Reports were overstating proven/conjectured security:**
- `StarkSecurityParams` defaulted `num_batched_functions: 1`, which is the value at which the opening-batching round is *dropped* from the report, not merely underestimated — every uni-stark security report was missing a real round. `from_air` now derives the count from the AIR's opening shape; realistic shapes see proven levels fall ~5-6 bits, correcting the omission. batch-stark had no equivalent derivation at all (one test hand-picked `8`); it now has `num_batched_openings`, extended with the permutation-trace's contribution.
- `from_air` didn't apply the zk quotient-chunk doubling `prove_with_preprocessed` actually performs, and didn't assert `max_combo` covers an AIR's own next-row reads — both silently overstated security for zk / next-row-reading AIRs.
- The conjectured-security path charged FRI's folding challenge under the [BCI+20] conjecture but omitted the opening-batching challenge under the same conjecture, for no principled reason — inconsistent, and it overstated every batched (i.e. every real) instance.
- `best_ldr_m` chose the list-decoding radius from the low-degree-test term alone, but the batching term also depends on `m` and can make the LDT's optimal `m` net negative once grinding lowers the LDT's own cost — reported security could *fall* as configured grinding *increased*. `proven_security_report` now sweeps the full composite, not just the LDT.
- `compute_upper_m` used `ceil` where the admissibility precondition (`m < X`) needs `floor`, and hardcoded the DEEP expansion at `2` instead of the caller's actual `max_combo` — both admitted an inadmissible `m` / overstated security.
- `LOG2_E` rounded down instead of up, shrinking `bits_per_query`'s correction term and inflating the reported per-query rate.
- The folding round's coefficient computation could overflow `u64` and wrap to zero for large-but-representable `log_folding_arity`, which `round` then read as "no such round" — a release build silently dropped the round from the budget instead of panicking like debug did.
- `grinding: GrindingSites::NONE` was hardcoded in `from_air`, so every grinding knob was inert in every report ever produced by it: raising `batch_proof_of_work_bits` changed the real protocol but moved the reported number by exactly zero bits.

**Grinding sites the accounting modeled but no protocol implemented:**
- batch-stark's out-of-domain point (`zeta`) and lookup-challenge (`alpha`, `beta`) samples were bare squeezes — bits were credited by `p3-security` and paid by nobody. Both are now implemented on prover and verifier, mirroring uni-stark's existing grind-before-`zeta` site; a batch with no lookups grinds nothing and rejects a smuggled witness rather than ignoring it.
- Renamed the out-of-domain knob from `deep_*` to `ood_*` (it was colliding in meaning with the pre-existing, unrelated `budget::deep_pow_bits`) — a breaking rename, but only against API this same branch already introduced.

**Config gaps that allowed unsound or panic-prone setups:**
- Circle-FRI had no `log_blowup == 0` guard, unlike two-adic FRI: at rate 1 every length-N word is itself a degree-<N codeword, so folding and the final-polynomial check pass on arbitrary data. Added on both prover and verifier.
- `CirclePcs::new` asserts `batch_proof_of_work_bits == 0`, but `FriParameters::new_testing`/`new_benchmark` and several struct-literal construction sites in the tree could hand it a nonzero value — panicking at runtime via `new`, or silently claiming a bit nobody enforced via the struct-literal bypass. Guarded at the PCS entry points and fixed the constructors so the guard can't be dodged.
- `HidingFriPcs`'s hiding property requires `N >= num_queries + 2 * Challenge::DIMENSION * (opening points)`; undocumented and unenforced, so a small trace under a production-sized query count could silently lose zero-knowledge.

## What this now allows

- `proven_security_report` / `conjectured_security_report` on a real STARK config now reflect every round the protocol actually runs — opening batching, zk doubling, hiding columns, and every grinding site — instead of a number quietly missing rounds it should have charged.
- batch-stark can be configured with lookup-challenge and out-of-domain grinding and have the transcript and the reported security level actually agree.
- Building a `CirclePcs`/FRI PCS with `log_blowup == 0` or an unsupported batch-grind value now fails at construction, instead of producing an unsound proof or panicking somewhere downstream and unrelated.

## Test plan

- [x] `cargo test -p p3-security -p p3-fri -p p3-circle -p p3-uni-stark -p p3-batch-stark` — all pass
- [x] `cargo check --workspace --tests` — clean
- [x] Serialization fixtures regenerated for the `BatchProof`/`Proof` shape changes (out-of-domain and lookup witnesses added)


tcoratger (approved):
Audited this one hard, because the failure mode of a security-accounting bug is that nothing breaks: the proofs keep verifying and the reported number is simply wrong in the optimistic direction. **I found nothing wrong.** Every fix I checked moves in the conservative direction, and the two genuine soundness holes are closed on both sides.

## The two real soundness holes

**Rate-1 Circle-FRI.** This is the serious one. At `log_blowup == 0` the code is the entire space, so every length-`N` word is trivially a degree-`<N` codeword and the folding chain accepts arbitrary data — the proximity test proves nothing. Closed on the prover with an `assert!` and on the verifier with `Err(FriError::ZeroBlowup)`, which is the right asymmetry: a verifier must reject rather than panic on input it did not choose. I removed the verifier guard and `reject_zero_blowup` fails, so it is load-bearing rather than decorative. It also now matches two-adic FRI, which already had it.

**Bits credited but never paid.** batch-stark's `zeta` and lookup-challenge samples were bare squeezes while `p3-security` was already crediting grinding for them. The implementation gets the shape right, including the part that is easy to miss: a batch with no lookups must carry *no* witness, and `check_and_sample_perm_challenges` returns `UnexpectedWitness` rather than ignoring one. An unchecked witness field riding along in a proof is malleable — a third party can rewrite those bytes and the proof still verifies — so rejecting it is the correct call, not pedantry. Prover and verifier share `sample_perm_challenges_inner`, so the challenge derivation cannot drift between them.

## The accounting fixes are all conservative

I checked the numeric ones individually rather than taking the description's word:

- **`LOG2_E = 94_549`** is `log2(e)` rounded **up** (94548.06 at 16 fractional bits). Rounding down would shrink the correction subtracted from the rate and inflate reported per-query security. The constant's doc says exactly this, and the value matches it.
- **`compute_upper_m`** now takes `floor(x)`, and returns `m - 1` when `x` is exactly integral. That is the correct encoding of a strict `m < x` precondition; `ceil` admitted an inadmissible `m`.
- **`folding_coefficient` returning `Option`** fixes the nastiest of these. `1u64 << 64` masks to `1u64 << 0`, so `2 * (1 - 1) = 0`, and `round` reads a zero coefficient as "no such round" — a release build silently dropped a round from the budget while debug panicked. `None` makes "too large to compute" explicit instead of aliasing onto "not applicable".
- **`num_batched_functions` is now required, with no default.** This is the right API response to the underlying problem: `1` is not an underestimate, it removes the batched-openings term from the report entirely, so a default would keep producing silently-incomplete numbers. Forcing every caller to supply it — or derive it via `from_air` / `num_batched_openings` — is what makes the omission impossible to reintroduce by accident.

The `best_ldr_m` fix is the subtlest and I think the reasoning holds: optimizing `m` against the LDT term alone can pick a radius that the batching term then makes net negative, which is how reported security could *fall* as grinding *rose*. Sweeping the composite is the only way that comes out right.

## Direction worth stating plainly

Every one of these corrections **lowers** a previously reported number — the uni-stark drop of ~5-6 bits from the batched-openings term is not a regression, it is the report finally charging a round the protocol always paid. Anyone who set parameters against the old output was over-trusting them by that margin. That is worth saying in the changelog louder than a bullet, because it is the kind of thing people calibrate production parameters against.

Regenerating the `BatchProof`/`Proof` fixtures is legitimate here: the proof shape genuinely gained the out-of-domain and lookup witness fields, so the old bytes describe a different structure.

188 tests pass across `p3-security`, `p3-fri`, `p3-circle`, `p3-uni-stark` and `p3-batch-stark`, zero failures. Workspace clippy `--all-targets -D warnings`, nightly fmt and `cargo sort` all clean.

Nothing I would open a follow-up for. Merging.

PR #2050, opened 2026-09-07
perf(field): route the self-algebra mixed dot product to the type's own dot product

tcoratger:
## Summary

- Give the blanket "every ring is an algebra over itself" impl a body: its mixed dot product now delegates to the ring's own dot product, so a type with a delayed-reduction dot product finally reaches it when it is its own algebra.
- Make the plain dot product's default use the tree-structured summation helper instead of a linear `.sum()` chain, so the delegation cannot lengthen anybody's critical path.
- Sumcheck round kernel, per hypercube element, in-cache: **2.06x** on the tower binary field, **1.57x** on the scalar GHASH field, **1.53x** on the packed GHASH field. Monty-31, Mersenne-31 and Goldilocks pick up 1.3x-3.0x on the same kernel.
- Semantics are unchanged for every type. Only the order of summation and the placement of the modular reduction move.

### Why the empty blanket impl silently defeated the existing kernels

Plonky3 already has delayed-reduction Karatsuba dot products for all three binary field types, and delayed Montgomery / u128 dot products for Monty-31, Mersenne-31 and Goldilocks. The sumcheck round kernel does not call those directly; it calls the algebra trait's *mixed* dot product, which takes algebra elements on one side and base scalars on the other. In the case the round kernel actually instantiates — evaluation table and weight table the same type — the impl that resolves is the blanket self-algebra one, and its body was empty, so it inherited the trait default: `N` fully reduced products followed by a tree sum. Every hand-written delayed-reduction kernel in the repo was therefore unreachable from the sumcheck, and the strategy module's docstring promising to collapse `K` widening multiplies into one reduction was false for exactly the case that runs. Giving the blanket impl a one-line body makes it true; the docstring is adjusted to say "modular reduction" rather than naming Montgomery, since the binary fields fold the modulus instead.

### Why the tree-sum default change is required, not cosmetic

The mixed dot product's default already used the balanced tree-sum helper, while the plain dot product's default was `u.iter().zip(v).map(|(x, y)| x * y).sum()` — a linear chain of `N - 1` dependent adds. Delegating to the plain dot product without fixing its default would have *lengthened* the critical path for every type that does not override it, which includes both binomial extension fields and every packing that only overrides the mixed form. With the change the two defaults are the same code, so the delegation is a provable no-op for those types (verified below at the machine-code level) and a strict win for the types that do override.

## Test plan

Machine: AMD Zen 5, 32 threads, AVX-512 + GFNI + VPCLMULQDQ. Benchmarks pinned to one core with `taskset -c 2`, built with `RUSTFLAGS="-C target-cpu=native"`, reported as the minimum of 15 timing repetitions. This is a workstation and may have been lightly loaded, so the ratios are more trustworthy than the absolute numbers. The baseline is a pristine `upstream/main` obtained with `git stash` inside the worktree; both binaries were kept and run alternately.

- [x] `cargo test --workspace` — 128 test binaries, 3615 passed, 0 failed, 34 ignored
- [x] `cargo test --workspace --features parallel` — 128 test binaries, 3615 passed, 0 failed
- [x] `cargo test --workspace --doc` — 49 doc targets, 4 doctests passed, 0 failed
- [x] `cargo +stable clippy --workspace --all-targets -- -D warnings` — clean
- [x] `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo +stable doc --no-deps --workspace --document-private-items` — clean
- [x] Target-feature legs, each `cargo check --workspace --all-targets`: no extra features, `-C target-feature=+avx2`, `-C target-feature=+pclmulqdq`, `-C target-cpu=native` — all clean
- [x] `RUSTFLAGS="-C target-cpu=native" cargo test -p p3-field -p p3-binary-field -p p3-baby-bear -p p3-koala-bear -p p3-mersenne-31 -p p3-goldilocks -p p3-bn254 -p p3-sumcheck -p p3-multilinear-util` — 28 binaries, 1816 passed, 0 failed
- [x] New reference-equality coverage: the mixed dot product of a ring over itself is compared against a fully reduced product-then-sum reference at `N` = 0, 1, 2, 3, 4, 5, 8, 13, 64, wired into both the ordinary and the characteristic-2 ring suites. That reaches BabyBear, KoalaBear, Mersenne-31, Goldilocks, BN254, the whole GF(2^{2^k}) tower including the polynomial-basis field, every binomial extension field in the test matrix, and every packing of all of the above. The packed GHASH type has no generic suite, so it gets an explicit test in the binary-field integration tests.

One caveat on the target-cpu leg: a full `RUSTFLAGS="-C target-cpu=native" cargo test --workspace --lib` aborts in a staged multi-STARK zerocheck test with a debug-mode stack overflow. It reproduces identically on pristine `upstream/main` with this branch stashed, so it is pre-existing and unrelated; every other binary in that run reported `0 failed`.

### The win — sumcheck round kernel, `Algebra<Self>` case

The harness reproduces the round-coefficient kernel exactly: `K = 8` tiles, `h(0) = sum w_lo*e_lo` and `h(inf) = sum (w_hi - w_lo)(e_hi - e_lo)`, both through the mixed dot product. `k` is `log2` of the hypercube elements in the round table; time is per hypercube element, so packed and scalar rows are directly comparable.

| type | shape | before | after | ratio |
|---|---|---|---|---|
| tower GF(2^128) | k=12 | 16.69 ns | 8.10 ns | **2.06x** |
| tower GF(2^128) | k=16 | 16.57 ns | 8.13 ns | **2.04x** |
| tower GF(2^128) | k=20 | 16.67 ns | 8.23 ns | **2.03x** |
| GHASH GF(2^128), scalar | k=12 | 2.300 ns | 1.466 ns | **1.57x** |
| GHASH GF(2^128), scalar | k=16 | 2.311 ns | 1.468 ns | **1.57x** |
| GHASH GF(2^128), scalar | k=20 | 2.321 ns | 1.494 ns | **1.55x** |
| GHASH GF(2^128), packed x4 | k=12 | 0.4865 ns | 0.3181 ns | **1.53x** |
| GHASH GF(2^128), packed x4 | k=16 | 0.4955 ns | 0.3376 ns | **1.47x** |
| GHASH GF(2^128), packed x4 | k=20 | 0.5002 ns | 0.3406 ns | **1.47x** |

### No-regression sweep — same kernel, the other field families

| type | shape | before | after | ratio |
|---|---|---|---|---|
| BabyBear (Monty-31) | k=12 | 0.7890 ns | 0.4757 ns | 1.66x |
| BabyBear packed x16 (AVX-512) | k=12 | 0.1274 ns | 0.0429 ns | 2.97x |
| BabyBear packed x16 (AVX-512) | k=20 | 0.1441 ns | 0.0658 ns | 2.19x |
| Mersenne-31 | k=12 | 0.6053 ns | 0.3830 ns | 1.58x |
| Mersenne-31 packed x16 (AVX-512) | k=12 | 0.0625 ns | 0.0441 ns | 1.42x |
| Goldilocks | k=12 | 1.1680 ns | 0.7037 ns | 1.66x |
| Goldilocks packed x8 (AVX-512) | k=12 | 0.3761 ns | 0.4147 ns | 0.91x (see below) |
| BinomialExtensionField<BabyBear, 4> | k=12 | 6.4549 ns | 6.4481 ns | 1.00x |
| the same, packed x16 | k=12 | 1.0703 ns | 1.1058 ns | 0.97x (see below) |
| BinomialExtensionField<BabyBear, 4> | k=20 | 6.4999 ns | 6.5747 ns | 0.99x (see below) |

The three sub-1.00x rows are types that override neither method, and for those the change is provably inert rather than merely close to inert. Disassembling both binaries and normalising relocation addresses, the round-kernel monomorphisations for packed Goldilocks, `BinomialExtensionField<BabyBear, 4>` and `PackedBinomialExtensionField<BabyBear, 4>` — including the out-of-line mixed dot product body of the last one, 1708 instructions — have **zero** non-address instruction differences. The measured spread on those rows is code layout, not work.

### Measured caveat: the delayed form loses once the loop is DRAM-bound

Delayed reduction holds a wider accumulator, which costs registers. Once the round table no longer fits in cache the loop is bandwidth-bound and those registers cost more than the reduction they save. Crossover, measured on the two packings that have an override:

| type | k=20 | k=21 | k=22 | k=23 | k=24 |
|---|---|---|---|---|---|
| GHASH packed x4 | 1.48x | 1.21x | 0.84x | 0.92x | 0.93x |
| Mersenne-31 packed x16 | 1.02x | 1.19x | 1.13x | 1.02x | 0.94x |

This is a property of the existing delayed-reduction overrides under bandwidth saturation, not of the trait plumbing: the same kernels are already the preferred path for every direct dot-product caller in the tree. Making the choice size-dependent is not expressible in a trait default, which has no notion of table size — it belongs in the sumcheck strategy kernel and is deliberately left out of this PR. Round tables halve every round, so at most the first one or two rounds are ever in that regime.

### Exactness, per characteristic

- Characteristic 2, the three binary types: reduction modulo the field polynomial is `F_2`-linear, so `sum_i (a_i*b_i) mod p == (sum_i a_i*b_i) mod p`. That identity is exactly why the three overrides exist.
- Monty-31, Mersenne-31, Goldilocks: the override accumulates widened products in a wider integer and reduces once. Each carries a compile-time or documented bound on `N` that keeps the accumulator from overflowing, and the reachable `N` values do not change — the mixed form is only ever called with the same small compile-time constants the plain form already was (`K = 8` from the round kernel, a power of two up to 64 from the batched linear combination, the extension degree from extension arithmetic).
- Everything else: the only change is reassociating a sum, which is exact in a commutative ring.
- No override calls back into the mixed form, so the delegation cannot recurse.
- The transcript does not move for any field, so no proof or verifier behaviour changes.

### DIFF
--- a/binary-field/tests/ghash.rs
+++ b/binary-field/tests/ghash.rs
@@ -97,3 +97,19 @@ fn packed_property_suite_supports_zero_in_small_fields() {
     // Half the random GF(2) denominators are zero; the suite must handle them explicitly.
     p3_field_testing::test_packed_vs_scalar_proptest::<Gf2>();
 }
+
+#[test]
+fn self_algebra_mixed_dot_product_matches_reference() {
+    // The generic field suites cover the scalar type; the packing has no such suite.
+    // It is also the type the sumcheck round kernels instantiate, so pin it here.
+    type P = <Ghash128 as Field>::Packing;
+
+    let mut rng = SmallRng::seed_from_u64(7);
+
+    // Independent operands, and different values in adjacent lanes.
+    let u: [P; 64] = array::from_fn(|_| P::from_fn(|_| rng.random()));
+    let v: [P; 64] = array::from_fn(|_| P::from_fn(|_| rng.random()));
+
+    // Compares every accumulation length against a fully reduced product-then-sum reference.
+    p3_field_testing::test_self_algebra_mixed_dot_product(&u, &v);
+}
--- a/binary-field/tests/ghash.rs
+++ b/binary-field/tests/ghash.rs
@@ -97,19 +97,3 @@ fn packed_property_suite_supports_zero_in_small_fields() {
     // Half the random GF(2) denominators are zero; the suite must handle them explicitly.
     p3_field_testing::test_packed_vs_scalar_proptest::<Gf2>();
 }
-
-#[test]
-fn self_algebra_mixed_dot_product_matches_reference() {
-    // The generic field suites cover the scalar type; the packing has no such suite.
-    // It is also the type the sumcheck round kernels instantiate, so pin it here.
-    type P = <Ghash128 as Field>::Packing;
-
-    let mut rng = SmallRng::seed_from_u64(7);
-
-    // Independent operands, and different values in adjacent lanes.
-    let u: [P; 64] = array::from_fn(|_| P::from_fn(|_| rng.random()));
-    let v: [P; 64] = array::from_fn(|_| P::from_fn(|_| rng.random()));
-
-    // Compares every accumulation length against a fully reduced product-then-sum reference.
-    p3_field_testing::test_self_algebra_mixed_dot_product(&u, &v);
-}
--- a/field-testing/src/lib.rs
+++ b/field-testing/src/lib.rs
@@ -22,8 +22,8 @@ pub use extension_testing::*;
 use num_bigint::BigUint;
 use p3_field::coset::TwoAdicMultiplicativeCoset;
 use p3_field::{
-    ExtensionField, Field, PackedValue, PrimeCharacteristicRing, PrimeField32, PrimeField64,
-    TwoAdicField, batch_multiplicative_inverse,
+    Algebra, ExtensionField, Field, PackedValue, PrimeCharacteristicRing, PrimeField32,
+    PrimeField64, TwoAdicField, batch_multiplicative_inverse,
 };
 use p3_util::iter_array_chunks_padded;
 pub use packedfield_testing::*;
@@ -193,6 +193,7 @@ where
     let vec2: [R; 64] = rng.random();
     test_sums(&vec1[..16].try_into().unwrap());
     test_dot_product(&vec1, &vec2);
+    test_self_algebra_mixed_dot_product(&vec1, &vec2);
 
     assert_eq!(
         x.exp_const_u64::<0>(),
@@ -406,6 +407,7 @@ where
     let vec2: [R; 64] = rng.random();
     test_sums(&vec1[..16].try_into().unwrap());
     test_dot_product(&vec1, &vec2);
+    test_self_algebra_mixed_dot_product(&vec1, &vec2);
 
     assert_eq!(
         x.exp_const_u64::<0>(),
@@ -924,6 +926,57 @@ pub fn test_dot_product<R: PrimeCharacteristicRing + Eq + Copy>(u: &[R; 64], v: 
         .zip(v.iter())
         .fold(R::ZERO, |acc, (&lhs, &rhs)| acc + (lhs * rhs));
     assert_eq!(dot_64, R::dot_product::<64>(u, v));
+}
+
+/// Check the mixed dot product of a ring viewed as an algebra over itself.
+///
+/// Every ring is an algebra over itself.
+/// In that case both argument arrays hold ring elements.
+/// The result must then equal the plain pairwise-multiply-then-add value.
+///
+/// The reference accumulates left to right, which pins one summation order.
+/// The implementation is free to reassociate the sum, or to delay modular reductions to the end.
+/// Both are exact in a commutative ring, so the two values must agree exactly.
+pub fn test_self_algebra_mixed_dot_product<R: PrimeCharacteristicRing + Eq + Copy>(
+    u: &[R; 64],
+    v: &[R; 64],
+) {
+    // Reference for the first `n` pairs: one fully reduced product each, added left to right.
+    fn reference<R: PrimeCharacteristicRing + Copy>(u: &[R], v: &[R], n: usize) -> R {
+        (0..n).fold(R::ZERO, |acc, i| acc + u[i] * v[i])
+    }
+
+    // The length is a compile-time constant, so the cases are spelled out rather than looped over.
+    macro_rules! check_len {
+        ($n:literal) => {
+            assert_eq!(
+                <R as Algebra<R>>::mixed_dot_product::<$n>(
+                    u[..$n].try_into().unwrap(),
+                    v[..$n].try_into().unwrap(),
+                ),
+                reference(u, v, $n),
+                "mixed dot product over {} pairs disagrees with the reference",
+                $n,
+            );
+        };
+    }
+
+    // Empty sum and single pair are the two degenerate lengths.
+    check_len!(0);
+    check_len!(1);
+
+    // 2, 4 and 8 fall exactly on the balanced-tree cases of the summation helper.
+    check_len!(2);
+    check_len!(4);
+    check_len!(8);
+
+    // 3, 5 and 13 are not powers of two, so they exercise the ragged tail of the tree.
+    check_len!(3);
+    check_len!(5);
+    check_len!(13);
+
+    // 64 is long enough to cross the chunked path of the summation helper.
+    check_len!(64);
 }
 
 pub fn test_sums<R: PrimeCharacteristicRing + Eq + Copy>(u: &[R; 16]) {
--- a/field-testing/src/lib.rs
+++ b/field-testing/src/lib.rs
@@ -965,17 +965,19 @@ pub fn test_self_algebra_mixed_dot_product<R: PrimeCharacteristicRing + Eq + Cop
     check_len!(0);
     check_len!(1);
 
-    // 2, 4 and 8 fall exactly on the balanced-tree cases of the summation helper.
+    // 2 through 8 are each a hand-written arm of `sum_array` and of `MontyField31::dot_product`.
     check_len!(2);
+    check_len!(3);
     check_len!(4);
+    check_len!(5);
+    check_len!(6);
+    check_len!(7);
     check_len!(8);
 
-    // 3, 5 and 13 are not powers of two, so they exercise the ragged tail of the tree.
-    check_len!(3);
-    check_len!(5);
+    // 13 sits past the last hand-written arm, so it takes the chunk-plus-remainder path.
     check_len!(13);
 
-    // 64 is long enough to cross the chunked path of the summation helper.
+    // 64 has an arm of its own in the packed dot products.
     check_len!(64);
 }
 
--- a/sumcheck/src/strategy.rs
+++ b/sumcheck/src/strategy.rs
@@ -47,9 +47,11 @@ const K: usize = 8;
 ///     leading  += sum_i  (w_hi[i] - w_lo[i]) * (e_hi[i] - e_lo[i])
 /// ```
 ///
-/// where `lo`, `hi` are the two faces of the active variable. Each sum is
-/// one delayed-reduction dot product over `K` pairs, collapsing `K`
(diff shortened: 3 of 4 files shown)


sumcheck/src/strategy.rs
```diff
@@ -47,9 +47,11 @@ const K: usize = 8;
...
 /// ```
 ///
-/// where `lo`, `hi` are the two faces of the active variable. Each sum is
-/// one delayed-reduction dot product over `K` pairs, collapsing `K`
-/// widening multiplies into one Montgomery reduce per output coordinate.
+/// where `lo`, `hi` are the two faces of the active variable.
+///
+/// Each sum is one delayed-reduction dot product over `K` pairs.
+/// It collapses `K` widening multiplies into one modular reduction per output coordinate.
+/// A Monty-31 packing reduces once with Montgomery, a binary field folds the modulus once.
```
Nashtare: This is not true on x86 at `K = 8`, which is the constant this file actually uses.

The delegation added here lands in `PackedMontyField31AVX2::dot_product` -> `general_dot_product`, whose match arms are `0, 1, 2, 3, 4, 64, _` (`monty-31/src/x86_64_avx2/packing.rs:697-760`). There is no `8` arm, so `K = 8` falls into `_`, which runs `dot_product_4` on `0..4`, `dot_product_4` on `4..8`, and then a packed add: **two** Montgomery reductions, not one. AVX-512 has the identical arm structure (`monty-31/src/x86_64_avx512/packing.rs:1172-1254`). Only AArch64 NEON has a `dot_product_8` arm (`monty-31/src/aarch64_neon/packing.rs:877`, helper at `:1280`), so the sentence holds there and only there.

Same issue with the pre-existing rationale a few lines up (`strategy.rs:29-33`): "primitives exist for tile sizes `2, 4, 5, 8`; `8` is the deepest available on every supported target" — on AVX2/AVX-512 the deepest specialised arm is 4. Since that comment is the stated justification for `K = 8`, it is worth correcting while you are in here: someone tuning `K` on x86 will believe 8 is a specialised delayed-reduction arm when it is a generic 4+4 fallback.

The binary-field half of the sentence is fine — XOR accumulation has no reduction to defer.

binary-field/tests/ghash.rs
```diff
@@ -97,3 +97,20 @@ fn packed_property_suite_supports_zero_in_small_fields() {
     // Half the random GF(2) denominators are zero; the suite must handle them explicitly.
     p3_field_testing::test_packed_vs_scalar_proptest::<Gf2>();
 }
+
+#[test]
+fn self_algebra_mixed_dot_product_matches_reference() {
+    // The generic field suites cover the scalar type, but the packing has no such suite.
+    //
+    // The packing is also what the sumcheck round kernels instantiate, so pin it here.
```
Nashtare: The premise of this comment is wrong, and the test adds no coverage on any build configuration.

`binary-field/src/packed/x86_64.rs:468` already runs `test_packed_binary_field!(crate::PackedGhash128, ...)`, which expands to `test_ring_with_eq_char2!` (`field-testing/src/packedfield_testing.rs:1014-1018`), and this PR wires `test_self_algebra_mixed_dot_product` into `test_ring_with_eq_char2` at `field-testing/src/lib.rs:410`. So on an `x86_64 + vpclmulqdq + avx2/avx512f` build the packing already gets exactly this check, and this test is a duplicate of it.

On every other target `<Ghash128 as Field>::Packing` resolves to `Ghash128` itself (`binary-field/src/packed/mod.rs:41`), so `P` is the scalar type and the test re-runs what the generic suites already cover via `field-testing/src/lib.rs:196`.

Either drop it, or keep it and reword the comment to say what it is actually for.

field-testing/src/lib.rs
```diff
@@ -926,6 +928,57 @@ pub fn test_dot_product<R: PrimeCharacteristicRing + Eq + Copy>(u: &[R; 64], v:
...
+
+    // 2, 4 and 8 fall exactly on the balanced-tree cases of the summation helper.
+    check_len!(2);
+    check_len!(4);
+    check_len!(8);
+
+    // 3, 5 and 13 are not powers of two, so they exercise the ragged tail of the tree.
+    check_len!(3);
+    check_len!(5);
+    check_len!(13);
```
Nashtare: The sweep names the summation helper's arms as its rationale but skips two of them: `sum_array` has hand-written `5`, `6` and `7` arms (`field/src/field.rs:364-366`), and only `5` is covered here.

`6` and `7` are also separately hand-written in `MontyField31::dot_product` (`monty-31/src/monty_31.rs`), each with its own `head_sum.min(head_sum_corr) + tail_sum` range argument that differs from the `4`/`5` arms, and on AVX2 they are what exercises the `N & 3 == 2` / `== 3` remainder dispatch in `general_dot_product`. `test_dot_product` covers those arms on `R::dot_product` directly, so there is no correctness hole — but two more `check_len!` lines make the delegation sweep match the reason stated in the comment above it, for free.

Nashtare (approved):

sumcheck/src/strategy.rs
tcoratger: Confirmed by reading all three. AVX2 `general_dot_product` (`monty-31/src/x86_64_avx2/packing.rs:686`) and AVX-512 (`x86_64_avx512/packing.rs:1160`) both have arms `0, 1, 2, 3, 4, 64, _`; NEON (`aarch64_neon/packing.rs:835`) is the only one with an `8` arm (`:877`, helper at `:1280`). At `N = 8` the x86 `_` arm does `dot_product_4` on `0..4`, then `(4..(N-3)).step_by(4)` fires once for `4..8`, then `N & 3 == 0` returns `acc` — two Montgomery reductions plus a modular add. The scalar `MontyField31::dot_product` hand-writes `0..=8`, so it is single-reduction like NEON.

Both comments now say that. The `K` doc (`sumcheck/src/strategy.rs:29-45`) lists the arm sets per target and states that `8` is a single-reduction tile on scalar and NEON but two length-4 tiles plus an add on AVX2/AVX-512. `chunk_round_step` (`:60-62`) no longer claims one reduction unconditionally and points at `K`. `K` itself is unchanged.

binary-field/tests/ghash.rs
tcoratger: Dropped it. The premise was wrong on both halves.

`PackedGhash128` does have a suite: `binary-field/src/packed/x86_64.rs:468` runs `test_packed_binary_field!`, which expands to `test_ring_with_eq_char2!`, which this PR wires `test_self_algebra_mixed_dot_product` into. And the scalar side runs the same helper through `test_binary_field!` at `binary-field/tests/field.rs:273`. So on a `vpclmulqdq + avx2/avx512f` build `<Ghash128 as Field>::Packing` is the packing and it is already covered, and on every other build it is `Ghash128` and that is already covered too.

The one thing I checked before deleting was whether the per-lane `P::from_fn(|_| rng.random())` added anything over the suite's `rng.random()`: it does not, because `impl_rng!(PackedGhash128)` samples the inner `[Ghash128; WIDTH]` element by element, so the suite's operands already have independent lanes.

Checked on the config that matters: `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" cargo nextest run -p p3-binary-field` still exercises `test_self_algebra_mixed_dot_product` through the `PackedGhash128` suite with the integration test gone.

field-testing/src/lib.rs
tcoratger: Added. The sweep now runs `2` through `8` contiguously, and the comment says why: `sum_array` hand-writes `0..=8` (`field/src/field.rs:359-366`, with `5`, `6`, `7` each its own arm) and `MontyField31::dot_product` hand-writes the same range (`monty-31/src/monty_31.rs:293`).

The two trailing cases got their real reason too: `13` is the first case past the last hand-written arm, so it takes the chunk-plus-remainder path, and `64` is there because the packed `general_dot_product` gives it a dedicated arm. `field-testing/src/lib.rs:964-981`.

PR #2051, opened 2026-09-07
perf(merkle-tree): reach the vectorized Keccak from the Merkle tree

tcoratger:
## Summary

- Merkle tree construction is where `binary-pcs` proving spends its time, and the vectorized Keccak permutations this repo already ships were unreachable from it: a byte digest has packing width 1, so the packed leaf and compression loops in `merkle-tree` degenerated to one scalar sponge per node, at every level of every tree.
- Adds a "hash many equal-length messages" entry point to `CryptographicHasher` and a "compress many groups" entry point to `PseudoCompressionFunction`, plus an associated constant declaring how many messages an implementation hashes most efficiently at once. Both methods default to exactly today's scalar behaviour and the constant defaults to 1, so every existing implementation keeps working untouched and no downstream crate breaks.
- `p3-keccak` overrides the batched hash with a multi-lane sponge driven by the `KeccakF` vector permutation that already lived in the crate with no caller outside it: absorb full rate blocks lane by lane, apply the `0x01`/`0x80` pad, permute once, squeeze. The lane count is the target's existing vector-width constant, so one body is correct at 1, 2, 4 and 8 lanes. `CompressionFunctionFromHasher` forwards its batched form to it over the concatenated child digests.
- The Merkle driver gains a batched arm for the leaf layer, the injection layer and the plain compression layer, taken when the hasher or the compression function declares more than one lane. Levels below 1024 output nodes stay batched but serial, since near the root a parallel dispatch costs more than the hashing.
- Second commit: every digest layer was allocated at the default digest and then completely overwritten. `Vec` only reaches the allocator's zeroed fast path when it can see at run time that the element is all zero bits, and the standard library deliberately skips that check for arrays longer than sixteen elements — so a thirty-two-word byte digest layer was filled word by word by a single thread before any hashing started. Building the layer as one flat run of digest words restores the fast path; a flat run of `len * D` words and a run of `len` digests have byte-identical allocation layout, so the view costs nothing and needs no initialization argument.
- **Commitments and proofs are byte-identical.** The only thing that changes is the order in which independent hashes are computed. A new test builds every tree twice — once through the batched arm, once through a wrapper that reports a single lane — and compares every node of every layer, not just the root.
- The algebraic path (Poseidon2 over a packed field) keeps `LANES = 1`, so it takes the existing packed arm unchanged. Benched to confirm no regression.

### Why the children-are-adjacent layout makes the batching free

A digest layer is one flat `Vec` of digests, and a node's `N` children sit next to each other in it. So the `N` children of one parent are already a contiguous run of digest words, and that run *is* the parent's preimage. The parents are consecutive too, so a run of `m` parents is a contiguous run of `m` preimages laid back to back — exactly the "equal-length messages, one after another" shape the batched hash asks for. Reading a layer as groups of `N` is therefore a reinterpretation of the same memory: no transposition, no gather, no copy at any internal level. Only the leaf layer needs a small scratch buffer, because a matrix only promises access one row at a time.

## Benchmarks

AMD Ryzen 9 9950X3D (Zen 5), 16 cores / 32 threads, AVX-512F + GFNI + VPCLMULQDQ, `RUSTFLAGS="-C target-cpu=native"`. **The machine was shared with other benchmark jobs during measurement**, so treat the ratios as approximate; every round measured all arms back to back, and each row shows two independent rounds.

### Tree construction — serial

`cargo bench -p p3-merkle-tree --bench merkle_tree`. This crate exposes no `parallel` feature, so this bench is single-threaded.

Keccak-256, arity 2, 32-byte digest, BabyBear rows of 135 columns (540-byte leaf messages, so the sponge absorbs four blocks per leaf):

| shape | main | + batched hashing | + zeroed alloc | ratio |
| one matrix 32768 x 135 | 70.0 / 70.6 ms | 13.1 / 12.1 ms | 13.3 / 13.9 ms | **5.4x** |
| two matrices 32769 x 135, 16385 x 135 | 105.0 / 104.0 ms | 18.8 / 18.3 ms | 18.1 / 22.6 ms | **5.5x** |

The allocation commit does nothing at this size, as expected: these trees hold about 2^16 digest slots, worth well under a millisecond of fill.

Poseidon2 over packed BabyBear, same bench, to confirm the algebraic path is untouched:

| shape | main | both commits |
| --- | --- | --- |
| arity 2, one matrix | 35.0 / 36.2 ms | 35.3 / 33.9 ms |
| arity 2, two matrices | 52.3 / 59.6 ms | 57.3 / 51.7 ms |
| arity 4, one matrix | 34.0 / 34.3 ms | 34.2 / 35.2 ms |
| arity 4, two matrices | 43.6 / 44.2 ms | 43.3 / 42.9 ms |

Unchanged within run-to-run noise.

### End to end — parallel

`cargo bench -p p3-binary-pcs --features parallel --bench pcs`, 20 variables, `log_inv_rate = 2`, `pow_bits = 8`, 100-bit target. Feature unification through this crate is what actually turns rayon on for the Merkle code.

| stage | main | + batched hashing | + zeroed alloc | ratio |
| `commit/20` | 243.8 ms | 82.8 ms | 49.3 ms | **4.9x** |
| `open/20` | 259.8 ms | 92.6 ms | 84.9 ms | **3.1x** |
| prove (`commit` + `open`) | 503.5 ms | 175.4 ms | 134.2 ms | **3.8x** |
| `verify/20` | 10.0 ms | 10.0 ms | 10.0 ms | 1.0x |

A second round of the same three arms measured `main` at 650 ms of proving against 132 ms, i.e. 4.9x; the `main` arm is the longest-running and therefore the most exposed to the other jobs on the box, so the 3.8x row above is the conservative reading. The verifier is deliberately untouched here: its per-level compressions are independent and would batch the same way, but that is a separate change.

### The digest-layer fill, measured on its own

A standalone program timing only `vec![[W::default(); D]; n]` in the exact generic shape the driver uses, at `n = 2^23` (256 MiB):

| digest element | before | after |
| --- | --- | --- |
| `[u8; 32]` | 42-52 ms | 11 us |
| `[Fp; 8]`, `Fp` a field newtype over `u32` | ~42 ms | ~42 ms (unchanged) |

This is worth stating plainly because two independent measurements of this fill disagreed: one put it at roughly 40 ms per large tree, the other at 1.6 ms of a 37 ms commit on the grounds that the allocator returns lazily-zeroed pages. The first held up and the second did not, and the deciding detail is the digest *width*: `Vec`'s zeroed-allocation specialization inspects the element at run time but gives up on arrays longer than sixteen elements, so a 32-word byte digest never reaches it while an 8-word digest of a primitive does. The end-to-end effect is visible in the table above — `commit/20` drops from 82.8 ms to 49.3 ms on this commit alone.

## Test plan

- [x] `cargo test -p p3-merkle-tree -p p3-symmetric -p p3-keccak` — 5 + 105 + 21 passed, 0 failed. Includes the new byte-identical-output tests: the batched and single-lane drivers are compared node for node across a single-row tree, odd heights that pad the layer, a batch count that leaves a partial final group, several matrices sharing the tallest height, a ragged height ladder that forces three injection levels, levels wide enough to fan out across threads, and rows long enough to make the sponge absorb four blocks — at arity 2, at arity 4, and at a deliberately awkward lane count of 3.
- [x] `cargo test -p p3-keccak` — the multi-lane sponge is pinned to `tiny_keccak` by a shape sweep over message lengths `{0, 1, 7, 8, 9, rate-1, rate, rate+1, 2*rate-1, 2*rate, 2*rate+1, 3*rate, 400, 1000}` crossed with batch sizes `1..=2*lanes+1`, plus a proptest over random lengths `0..=400` and batches `1..=17`.
- [x] `cargo test -p p3-binary-pcs` — 37 + 19 passed, 0 failed.
- [x] `cargo test --workspace` — all green. The traits are shared, so this is the gate that matters.
- [x] `cargo test -p p3-merkle-tree -p p3-symmetric -p p3-keccak --doc` — green.
- [x] `cargo +stable clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo +stable doc --no-deps --workspace --document-private-items` — no new warnings.
- [x] `cargo check --workspace --all-targets` — clean, so no existing implementor of either trait needed a change.
- [x] Target-feature legs, since the Keccak lane count is target-gated: default `cargo test` (SSE2, 2 lanes), `RUSTFLAGS="-C target-feature=+avx2"` (4 lanes) and `RUSTFLAGS="-C target-cpu=native"` (AVX-512, 8 lanes) all pass the full `p3-merkle-tree`, `p3-symmetric`, `p3-keccak` and `p3-binary-pcs` suites. The 1-lane fallback and the NEON build cannot execute on this host, so they were type-checked instead: `cargo check --target wasm32-unknown-unknown`, `--target thumbv7em-none-eabi` and `--target aarch64-unknown-linux-gnu`, all clean.
- [x] `cargo +nightly miri test -p p3-merkle-tree --lib -- test_compress default_digest_layer keccak_tree_is_deterministic` — 7 passed, 0 failed. Covers the one `unsafe` block introduced by the allocation commit and a full small Keccak tree built through the batched arm, including its padding tail.
- [x] `cargo +nightly miri test -p p3-keccak -p p3-symmetric --lib` — 5 + 21 passed, 0 failed, including the sponge proptests.
- [x] Parallel and serial both exercised: the `p3-binary-pcs` runs above have rayon on through feature unification, the `p3-merkle-tree` runs have it off.

### DIFF
--- a/merkle-tree/src/merkle_tree.rs
+++ b/merkle-tree/src/merkle_tree.rs
@@ -243,6 +243,61 @@ pub(crate) fn select_arity_step<const N: usize>(
     if has_intermediate { 2 } else { N }
 }
 
+/// Allocate a digest layer of `len` slots, every slot set to the default digest.
+///
+/// Every slot of a layer is overwritten by the level that produces it, apart from a short
+/// padding tail, so the fill is pure overhead and worth getting for free from the allocator.
+///
+/// A zero-filled allocation is free: the operating system hands back pages that are already
+/// zero and only faults them in when the hashing threads write them.
+/// The standard vector constructor reaches that path only when it can see at run time that the
+/// element is all zero bits, and it deliberately gives up on that check for arrays longer than
+/// sixteen elements, which is exactly the shape of a thirty-two-byte digest.
+///
+/// Building the layer as one flat run of digest words restores the check, because a word is a
+/// primitive the constructor still inspects.
+/// The flat run and the run of digests have the very same allocation layout, so viewing one as
+/// the other costs nothing:
+///
+/// ```text
+///     flat:    [ w_0 w_1 ... w_{D-1} | w_D ... w_{2D-1} | ... ]   len * D words
+///     digests: [       digest_0      |     digest_1     | ... ]   len digests
+/// ```
+///
+/// # Panics
+/// Panics if the layer is too large for the address space.
+fn default_digest_layer<W, const DIGEST_ELEMS: usize>(len: usize) -> Vec<[W; DIGEST_ELEMS]>
+where
+    W: Copy + Default,
+{
+    // A layer with no slots, or a digest of no words, owns no allocation worth reinterpreting.
+    if len == 0 || DIGEST_ELEMS == 0 {
+        return vec![[W::default(); DIGEST_ELEMS]; len];
+    }
+
+    // Guard the multiply so a wrapped word count can never under-allocate the layer.
+    let words = len
+        .checked_mul(DIGEST_ELEMS)
+        .expect("digest layer length overflows");
+
+    let mut flat: Vec<W> = vec![W::default(); words];
+
+    // A vector may reserve more than requested, and only an exact allocation converts.
+    if flat.capacity() != words {
+        return vec![[W::default(); DIGEST_ELEMS]; len];
+    }
+
+    let ptr = flat.as_mut_ptr().cast::<[W; DIGEST_ELEMS]>();
+    core::mem::forget(flat);
+
+    // SAFETY: an array of `DIGEST_ELEMS` words has no padding, so `len` digests occupy exactly
+    // the `len * DIGEST_ELEMS` words allocated above, with the same alignment as one word.
+    // The requested length equals the reserved capacity, so the layout handed back to the
+    // allocator on drop is byte for byte the layout it handed out.
+    // Every word is initialized, so every digest slot reads as the default digest.
+    unsafe { Vec::from_raw_parts(ptr, len, len) }
+}
+
 /// Output nodes below which a level is hashed on the calling thread.
 ///
 /// A parallel dispatch costs a fixed amount per level, and the levels near the root hold so few
@@ -378,9 +433,9 @@ where
     let max_height = tallest_matrices[0].height();
     let max_height_padded = padded_len(max_height, N);
 
-    // Slots past the real rows exist only so the next level can form whole groups.
-    let default_digest = [W::default(); DIGEST_ELEMS];
-    let mut digests = vec![default_digest; max_height_padded];
+    // Slots past the real rows exist only so the next level can form whole groups, and the
+    // allocation already leaves them at the default digest.
+    let mut digests = default_digest_layer::<W, DIGEST_ELEMS>(max_height_padded);
 
     hash_rows_batched(h, tallest_matrices, 0, &mut digests[..max_height]);
 
@@ -413,8 +468,7 @@ where
     let next_len = prev_layer.len() / N;
     let next_len_padded = padded_len(next_len, N);
 
-    let default_digest = [W::default(); DIGEST_ELEMS];
-    let mut next_digests = vec![default_digest; next_len_padded];
+    let mut next_digests = default_digest_layer::<W, DIGEST_ELEMS>(next_len_padded);
 
     // Any trailing child that cannot complete a group takes no part in this level.
     let (groups, _) = prev_layer[..next_len * N].as_chunks::<N>();
@@ -457,7 +511,7 @@ where
     let next_len_padded = padded_len(raw_next, N);
 
     let default_digest = [W::default(); DIGEST_ELEMS];
-    let mut next_digests = vec![default_digest; next_len_padded];
+    let mut next_digests = default_digest_layer::<W, DIGEST_ELEMS>(next_len_padded);
 
     let (groups, _) = prev_layer[..raw_next * N].as_chunks::<N>();
 
@@ -554,11 +608,9 @@ where
 
     let max_height_padded = padded_len(max_height, N);
 
-    // Prepare a default digest value to fill unused slots or padding.
-    let default_digest = [PW::Value::default(); DIGEST_ELEMS];
-
-    // Allocate the digest vector with padded size, initialized to default digest.
-    let mut digests = vec![default_digest; max_height_padded];
+    // Allocate the digest vector with padded size, every slot at the default digest so the
+    // padding tail past the real rows is already correct.
+    let mut digests = default_digest_layer::<PW::Value, DIGEST_ELEMS>(max_height_padded);
 
     // Parallel loop: process complete batches of `width` rows at a time.
     digests[0..max_height]
@@ -657,7 +709,7 @@ where
     let next_len_padded = padded_len(raw_next, N);
 
     let default_digest = [PW::Value::default(); DIGEST_ELEMS];
-    let mut next_digests = vec![default_digest; next_len_padded];
+    let mut next_digests = default_digest_layer::<PW::Value, DIGEST_ELEMS>(next_len_padded);
 
     let default_packed: [PW; DIGEST_ELEMS] =
         array::from_fn(|_| PW::broadcast(PW::Value::default()));
@@ -794,7 +846,7 @@ where
     let next_len_padded = padded_len(next_len, N);
 
     let default_digest = [P::Value::default(); DIGEST_ELEMS];
-    let mut next_digests = vec![default_digest; next_len_padded];
+    let mut next_digests = default_digest_layer::<P::Value, DIGEST_ELEMS>(next_len_padded);
 
     let default_packed: [P; DIGEST_ELEMS] = array::from_fn(|_| P::broadcast(P::Value::default()));
 
@@ -1010,6 +1062,24 @@ mod tests {
             );
         }
         assert_eq!(batched.root(), unbatched.root());
+    }
+
+    #[test]
+    fn default_digest_layer_is_all_default() {
+        // Byte digests: the word type is a primitive, so the allocation takes the zeroed path
+        // and the reinterpretation back to digests must still read as the default digest.
+        let bytes = default_digest_layer::<u8, 32>(5);
+        assert_eq!(bytes, vec![[0u8; 32]; 5]);
+
+        // A field word type whose default is not a primitive zero follows the plain fill path.
+        let words = default_digest_layer::<F, 8>(3);
+        assert_eq!(words, vec![[F::default(); 8]; 3]);
+
+        // An empty layer is legal: a height-zero tree allocates nothing.
+        assert!(default_digest_layer::<u8, 32>(0).is_empty());
+
+        // A digest width of one exercises the case where a digest is a single word.
+        assert_eq!(default_digest_layer::<u8, 1>(7), vec![[0u8; 1]; 7]);
     }
 
     #[test]
--- a/merkle-tree/src/merkle_tree.rs
+++ b/merkle-tree/src/merkle_tree.rs
@@ -243,6 +243,266 @@ pub(crate) fn select_arity_step<const N: usize>(
     if has_intermediate { 2 } else { N }
 }
 
+/// Output nodes below which a level is hashed on the calling thread.
+///
+/// A parallel dispatch costs a fixed amount per level, and the levels near the root hold so few
+/// nodes that the dispatch outweighs the hashing itself.
+/// Grouping messages is independent of threading, so a small level keeps the full vector width
+/// even while it runs on one thread.
+/// 1024 is the point where the two costs balance in practice: it still leaves tens of nodes per
+/// thread on a many-core machine, and every wider level fans out.
+const SERIAL_LEVEL_NODES: usize = 1024;
+
+/// Output nodes handed to one parallel task.
+///
+/// Matching the serial threshold keeps a task large enough to amortize the dispatch and small
+/// enough that its scratch buffers stay in the mid-level cache.
+const TASK_NODES: usize = 1024;
+
+/// Target size in bytes of the buffer holding one group of copied rows.
+///
+/// A batched hasher wants its messages back to back in memory, and a matrix only promises access
+/// one row at a time, so a group of rows is copied into a buffer first.
+/// 16 KiB keeps that buffer inside the first-level cache, so the hasher reads the rows back while
+/// they are still hot.
+const ROW_SCRATCH_BYTES: usize = 16 * 1024;
+
+/// Hash a run of rows from a set of equal-height matrices, several messages per hash call.
+///
+/// The message for a row is the concatenation of that row across every matrix, in matrix order,
+/// which is exactly what the unbatched path feeds its hasher one row at a time.
+///
+/// # Arguments
+/// - `h`: hasher applied to each row message.
+/// - `matrices`: matrices of equal height whose rows are concatenated.
+/// - `first_row`: index of the row whose digest lands in the first output slot.
+/// - `out`: one slot per consecutive row starting at that index.
+///
+/// # Panics
+/// Panics if any matrix is shorter than the requested row range.
+fn hash_rows_batched<F, W, H, M, const DIGEST_ELEMS: usize>(
+    h: &H,
+    matrices: &[&M],
+    first_row: usize,
+    out: &mut [[W; DIGEST_ELEMS]],
+) where
+    F: Clone + Send + Sync,
+    W: Send + Sync,
+    H: CryptographicHasher<F, [W; DIGEST_ELEMS]> + Sync,
+    M: Matrix<F>,
+{
+    // Bound the row range up front, since the copy loop below skips per-row bounds checks.
+    assert!(
+        matrices.iter().all(|m| m.height() >= first_row + out.len()),
+        "row range {}..{} exceeds a matrix height",
+        first_row,
+        first_row + out.len()
+    );
+
+    // A message spans one row of every matrix.
+    let total_width: usize = matrices.iter().map(|m| m.width()).sum();
+    let row_bytes = (total_width * size_of::<F>()).max(1);
+
+    // Fill as many whole lane groups as the scratch budget allows, and never fewer than one,
+    // so every hash call keeps the hasher's vector width busy.
+    let lanes = H::LANES;
+    let rows_per_call = (ROW_SCRATCH_BYTES / row_bytes / lanes).max(1) * lanes;
+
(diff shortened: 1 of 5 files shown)


symmetric/src/serializing_hasher.rs
```diff
@@ -24,12 +37,85 @@ where
     F: Field,
     Inner: CryptographicHasher<u8, [u8; N]>,
 {
+    const LANES: usize = <Inner as CryptographicHasher<u8, [u8; N]>>::LANES;
```
Nashtare, line 44: Only the `[u8; N]` impl forwards `LANES` and overrides `hash_many`. The `[u32; N]` impl (`:121`), the `[u64; N]` impl (`:134`) and the three packed `[F; M] -> [[uX; M]; N]` impls (`:147`, `:161`, `:175`) all keep the trait defaults, so `LANES` stays 1 for them and the Merkle driver's batched arm is never taken.

That is not academic: `examples/src/types.rs:27-34` builds `KeccakMerkleMmcs` as `MerkleTreeMmcs<[F; VECTOR_LEN], [u64; VECTOR_LEN], SerializingHasher<PaddingFreeSponge<KeccakF,25,17,4>>, …, 2, 4>`, which resolves to the `[[u64; M]; N]` impl at `:175` — so the uni-stark and circle-stark Keccak configurations get nothing from this PR.

I suspect that is deliberate, because those configs are already packed (`P = [F; VECTOR_LEN]`, `PW = [u64; VECTOR_LEN]`) and therefore already reach the vectorized permutation through the existing packed arm, which is why the PR's own Poseidon2/packed rows show no change. If so, a sentence here saying "the packed digest shapes reach the vector permutation through packing instead, so they deliberately keep `LANES = 1`" would save the next reader the trace. If not, the `[u64; N]` shapes are worth forwarding too.

(The byte-digest shape this does serve is well used — `binary-pcs`, `binary-dft`, `multi-stark/examples/prove_binary_field.rs`, the `fri`/`uni-stark`/`whir` test and bench configs are all `MerkleTreeMmcs<_, u8, _, _, 2, 32>` — so the win is real, it is just narrower than "the Merkle tree".)

symmetric/src/serializing_hasher.rs
```diff
@@ -24,12 +37,85 @@ where
...
+        }
+
+        // A row too wide for the stack buffer gains nothing from grouping, and the unbatched
+        // path serializes straight into the sponge with no buffer at all.
+        if row_bytes > ROW_BYTES_SCRATCH {
+            for (digest, row) in out.iter_mut().zip(input.chunks_exact(row_len)) {
+                *digest = self.hash_iter(row.iter().copied());
+            }
+            return;
+        }
```
Nashtare: This wide-row fallback is not reached by any test. It needs `row_bytes > 8192`, i.e. more than 2048 `BabyBear` columns, and the widest entry in the driver's `SHAPES` fixture is 135 columns.

It is the one branch where a mistake would be quiet rather than loud: if someone later "optimises" this to chunk rows like the branch below it, wide traces would get silently wrong leaf digests and every existing test would still pass. One `SHAPES` entry with a row wider than `ROW_BYTES_SCRATCH / size_of::<F>()` would pin it.

Same shape of gap for `CryptographicHasher::hash_many`'s `out.is_empty()` / `len == 0` branches (`hasher.rs:80`, `:95`) and for `CompressionFunctionFromHasher::compress_many`'s `out.len().min(inputs.len())` trim (`compression.rs:114`), which is never called with unequal lengths.

symmetric/src/serializing_hasher.rs
```diff
@@ -24,12 +37,85 @@ where
...
+
+        // Serialize as many whole rows at a time as the buffer holds, so the inner hasher sees a
+        // long run of equal-length byte messages and can fill every lane of its permutation:
+        //
+        //     rows:    [ r_0 | r_1 | ... ]              row_len field elements each
+        //     scratch: [ b_0 | b_1 | ... ]              row_bytes bytes each
+        let rows_per_group = ROW_BYTES_SCRATCH / row_bytes;
```
Nashtare: This undoes the lane alignment the caller just established. `merkle_tree.rs:379` computes `rows_per_call = (ROW_SCRATCH_BYTES / row_bytes / lanes).max(1) * lanes` specifically so every `hash_many` call is a whole number of lane groups — and then this re-chunks by `ROW_BYTES_SCRATCH / row_bytes`, which has no lane factor in it.

Concretely, at the benchmark's 540-byte rows `rows_per_group == 15`. With `VECTOR_LEN == 2` every group ends on an odd row, so one permutation per group runs half idle (~7% of the work). On AVX-512 (`VECTOR_LEN == 8`) 15 rows is one full group of 8 plus a group of 7 that leaves 7/8 of the lanes idle.

`let rows_per_group = (ROW_BYTES_SCRATCH / row_bytes / Inner::LANES).max(1) * Inner::LANES;` restores it.

symmetric/src/serializing_hasher.rs
```diff
@@ -24,12 +37,85 @@ where
...
+        let mut scratch = [0u8; ROW_BYTES_SCRATCH];
```
Nashtare: This zeroes the full 8 KiB on every call, but only `digests.len() * row_bytes` bytes are ever used (`:95`), and every one of those is unconditionally overwritten by the serialize loop at `:103-114`. For the leaf layer that is an 8 KiB memset per group across the whole tree.

`MaybeUninit` is the fully general fix, but since the buffer is per-call and immediately overwritten, sizing it to `rows_per_group * row_bytes` — or hoisting it out of the function into the caller's per-task scratch, where `merkle_tree.rs:383` already keeps one — avoids the `unsafe` entirely.

symmetric/src/compression.rs
```diff
@@ -7,7 +7,36 @@ use crate::permutation::CryptographicPermutation;
...
+    /// ```text
+    ///     inputs: [ grp_0 | grp_1 | ... | grp_{m-1} ]   m groups of N inputs
+    ///     out:    [ dig_0 | dig_1 | ... | dig_{m-1} ]   m outputs
+    /// ```
+    ///
+    /// The default compresses the groups one at a time.
+    /// An override exists purely to exploit vector hardware and must return the very same outputs.
+    ///
+    /// Groups beyond the shorter of the two slices are ignored, so the caller controls the count.
+    fn compress_many(&self, inputs: &[[T; N]], out: &mut [T])
```
Nashtare: These two sibling batch APIs have opposite failure modes on a ragged call: `compress_many` documents "groups beyond the shorter of the two slices are ignored" and silently truncates, while `CryptographicHasher::hash_many` (`hasher.rs:85`) *panics* on a length that is not an exact multiple.

`CompressionFunctionFromHasher::compress_many` (`:114`) then has to pre-trim with `out.len().min(inputs.len())` precisely to stop the lenient contract from tripping the strict one. Picking one convention for both — I would take the assert, since a ragged batch here is a driver bug — removes that adapter and the surprise.

merkle-tree/src/merkle_tree.rs
```diff
@@ -243,6 +243,337 @@ pub(crate) fn select_arity_step<const N: usize>(
...
+
+        for (group, group_digests) in digests.chunks_mut(rows_per_call).enumerate() {
+            let first = base + group * rows_per_call;
+
+            // Lay the group's messages back to back: row by row, matrix by matrix.
+            scratch.clear();
+            for row in first..first + group_digests.len() {
+                for m in matrices {
+                    // SAFETY: the assertion above bounds every requested row by every height.
+                    let slice = unsafe { m.row_slice_unchecked(row) };
```
Nashtare: This swaps an allocation-free row iterator for one that may allocate. The base path used `row_unchecked` / `vertically_packed_row`; `row_slice_unchecked`'s default implementation bottoms out in `row_subslice_unchecked` (`matrix/src/lib.rs:195-205`), which is `row_subseq_unchecked(..).into_iter().collect_vec()` — one heap allocation per row.

Every matrix on the FRI/STIR commit path overrides it (`dense.rs:468`, `row_index_mapped.rs:98`, `stack.rs:149`, `horizontally_truncated.rs:118`), so this is latent rather than live. But a `Matrix` impl that takes the default now pays an allocation per row across the entire leaf layer, where before it paid none — worth either a `debug_assert`-style note here that callers are expected to override, or using `row_subseq_unchecked` and extending from the iterator directly.

Nashtare (approved):

symmetric/src/serializing_hasher.rs
tcoratger: You are right, it is (a), and I checked rather than assumed.

`Keccak256Hash: CryptographicHasher<u8, [u8; 32]>` is the only impl in the workspace that sets `LANES` or overrides `hash_many`. `PaddingFreeSponge` sets neither, so `SerializingHasher<PaddingFreeSponge<KeccakF, 25, 17, 4>>` as `CryptographicHasher<F, [u64; 4]>` would report `LANES = 1` even if the `[u64; N]` impl forwarded it — forwarding there is a no-op today, for `examples/src/types.rs:27` and for `whir/src/pcs/tests.rs:1114`'s unpacked `MerkleTreeMmcs<F, u64, ...>` alike.

Worse, forwarding `LANES` without also overriding `hash_many` would be actively wrong: `first_digest_layer` selects the batched arm on the scalar impl's `LANES`, so a `[u64; N]` inner hasher that reported lanes it does not batch would trade the packed vector arm for one scalar sponge per row.

Documented at `symmetric/src/serializing_hasher.rs:38` instead, on the `LANES` line itself.

symmetric/src/serializing_hasher.rs
tcoratger: All four branches are pinned now.

Wide row: `SHAPES` gains `(&[3], 2100)` (`merkle-tree/src/merkle_tree.rs:961`) — 8400 bytes a row, past the 8 KiB budget. `symmetric/src/serializing_hasher.rs:432` pins it directly too, with an inner hasher that records every batched call it receives and must see none.

`hash_many`'s early exits: `symmetric/src/hasher.rs:151` (`len == 0` hashes the empty message), `:160` (`out.is_empty()` reads nothing — without it `is_multiple_of(0)` is false for a non-empty input and the assert fires), `:168` (ragged panics). Same three for the serializing impl at `serializing_hasher.rs:447`, `:459`, `:470`.

`compress_many`'s trim is gone (see the other thread); what replaced it is pinned by `compression.rs:285` and `:318`, on the trait default and on the adapter.

symmetric/src/serializing_hasher.rs
tcoratger: Fixed at `symmetric/src/serializing_hasher.rs:98-103`, though not quite with that expression — it can overrun the buffer.

With `row_bytes = 8000` and 4 lanes, `(8192 / 8000 / 4).max(1) * 4` is 4, so the group claims 32000 bytes of an 8192-byte budget. Rounding after the multiply instead, and keeping whatever fits when a whole lane group does not:

```rust
let lanes = Inner::LANES.max(1);
let rows_fitting = ROW_BYTES_SCRATCH / row_bytes;
let rows_per_group = match rows_fitting / lanes {
    0 => rows_fitting,
    groups => groups * lanes,
};
```

`rows_fitting >= 1` because the wide-row case returned above, so `rows_per_group` is in `1..=rows_fitting` either way and the buffer is never overrun. At 540-byte rows it is 14 for `VECTOR_LEN == 2`, 12 for 4, 8 for 8.

`serializing_hasher.rs:401` pins the group sizes with a three-lane recorder: 40 rows of 400 bytes come out as `[18, 18, 4]`, only the last short and only because the input ran out. `:417` pins the fallback, where 2800-byte rows leave room for two of a three-row lane group.

symmetric/src/serializing_hasher.rs
tcoratger: Sized to the used prefix, no `unsafe`, at `symmetric/src/serializing_hasher.rs:107`:

```rust
let mut scratch: Vec<u8> = Vec::with_capacity(rows_per_group.min(out.len()) * row_bytes);
```

then `clear()` and `extend` from the byte stream per group. Hoisting into the caller's scratch was the alternative, but it cannot be done without widening `CryptographicHasher::hash_many`'s signature, and the buffer is a byte buffer whereas the caller's is `Vec<F>` — so the API stays as it is.

That drops the fixed 8 KiB memset for a single allocation of exactly the largest group, and the `min(out.len())` also stops a two-row call from reserving the whole budget. The per-row `expect`/`assert` pair collapses into one `assert_eq!` on the running length, which still catches a serialization that is short or long in either direction.

I did not benchmark it — the machine was too loaded for a trustworthy number, so I am claiming none. The justification is only that the memset was unconditional and the bytes it wrote were all overwritten before being read.

symmetric/src/compression.rs
tcoratger: Unified on the assert. `compression.rs:33` now asserts `inputs.len() == out.len()` in the default body, and the adapter at `:113` asserts the same and hands the whole flattened run to `hash_many` — the `min` is gone.

Both doc comments state it the same way now: "Panics if the batch is ragged: the group count must equal the output count" / "...the input length must be a whole multiple of the digest count".

`grep -rn compress_many --include=*.rs` finds four call sites, all in the Merkle driver (`merkle_tree.rs:429`, `:434`, `:542`, `:558`). Each derives its group slice and its output slice from one length — `groups` is `prev_layer[..next_len * N].as_chunks::<N>()` against `next_digests[..next_len]`, the `par_chunks`/`par_chunks_mut` pair is zipped at equal length, and `folded`/`pairs` are built at `digest_chunk.len()`. Nothing relied on the truncation.

The adapter needed the assert rather than leaning on `hash_many`'s: unequal counts can still divide, so `inputs.len() == 4, out.len() == 2` would have silently redrawn every message boundary instead of panicking.

merkle-tree/src/merkle_tree.rs
tcoratger: Switched to the iterator at `merkle-tree/src/merkle_tree.rs:396`:

```rust
scratch.extend(unsafe { m.row_unchecked(row) });
```

`row_unchecked` is `row_subseq_unchecked(r, 0, width())` for the full row, and every `Matrix` in the tree serves it without allocating.

It is also less latent than it looked: `FlatMatrixView::row_slice_unchecked` (`matrix/src/extension.rs:98`) collects into a `Vec` explicitly, so every `ExtensionMmcs` commit that reached this path — the FRI and STIR folding layers under any byte-digest config — was paying one allocation per row.

No behaviour change: the elements arrive in the same order, and for a dense matrix the iterator is `Cloned<slice::Iter<_>>`, which is `TrustedLen`, so `Vec::extend` takes `extend_trusted` (reserve once, then a raw write loop) rather than a push loop.

I did not re-measure — the box was saturated and any number would have been noise. The claim is just that an allocation per row is gone and none was added.

PR #2052, opened 2026-09-07
perf(binary-dft): fuse long-stride stages through a staging tile

tcoratger:
## Summary

- Fuse the long-stride stages of the polynomial-basis transform through a per-worker staging tile, so the matrix is read once and written once per *group* of stages instead of once per stage
- Fold both basis conversions into the transform's first read and last write of each element, removing two more full passes over the matrix
- Pure reordering of memory traffic: the output is bit-identical, pinned by a new A/B test against a per-stage reference schedule
- Measured **1.65x at 2^16 rows and 3.7x at 2^20 rows** (width 16) on `ntt/128/hybrid`, the `PolyBasisNtt` path

### What was slow

The transform runs `log_n` butterfly stages, and stage `j` pairs row `r` with row `r + 2^j`.

For small `j` the two rows are close together, and the existing tiling already exploits that: the matrix is cut into 32 KiB tiles of consecutive rows, and every stage that stays inside a tile runs there without leaving it.

For large `j` the partners are megabytes apart, so each of those stages was a full pass over the whole buffer — at 2^20 rows and width 16 that is 13 passes over 256 MiB, plus two more for the basis conversions. At that size the transform is entirely memory-pass bound; the butterflies themselves are a few percent of the runtime.

### The staging tile

The long-stride stages have exactly the same tiling structure, just along the stride axis instead of the contiguous axis.

Take a group of `T` stages ending at stage `top - 1`, and let `S = 2^(top - T)`. The `2^T` rows

```text
    row(k) = block * 2^top + offset + k * S ,   k = 0 .. 2^T ,   offset < S
```

are closed under all `T` stages: stage `top - 1 - s` pairs rows `2^(top-1-s) = 2^(T-1-s) * S` apart, which is a step of `2^(T-1-s)` in `k`, inside the set for every `s < T`. Varying `(block, offset, k)` is a mixed-radix decomposition of the row range, so those sets partition the matrix.

So each task gathers its `2^T` rows into one contiguous buffer, runs the whole group inside it as an ordinary radix-2 network, and scatters back — one read and one write of the matrix for `T` stages instead of `T` of each.

The buffer is sized from a 64 KiB budget, which is `T = 8` at width 16 and stays inside the private cache level below the shared one. A 1 MiB buffer fuses 12 stages but measured a third slower, so 64 KiB it is.

### The twiddle index

The one fiddly part is which twiddle each sub-layer of the tile needs, and it is exact rather than approximate.

Sub-layer `s` pairs staged rows `(k, k + 2^(T-1-s))`, and the global stage-`(top-1-s)` block index of that pair is

```text
    row(k) >> (top - s) = block * 2^s + (k >> (T - s))
```

because `offset + (k mod 2^(T-s)) * S < 2^(top-s)` contributes nothing to the shift.

That is precisely the block index a contiguous run of `2^(T-s)` staged rows already carries, so the tile runs through the same per-stage helper the contiguous tiles use, with its twiddle walk seeded at `block << s`. Same Cantor basis, same XOR increments, same subspace-polynomial shifts — no new twiddle machinery.

### The basis conversions

`PolyBasisNtt` runs in the polynomial basis and converts the whole matrix in on the way in and out on the way out. Each conversion was its own `par_iter_mut` pass, and once the transform is down to ~3 passes those two are a large fraction of what is left.

A conversion is a pure per-element map, so it commutes with every butterfly ordering. It now rides on the phase that already holds the element in cache: the entry conversion on the first staging gather (or on the tile, when no group runs), the exit conversion on the last tile store (or on the last staging scatter, for the inverse direction). Where no phase can carry it, a standalone pass is still used, so `log_n` below the cut and staging tiles too narrow to fuse behave exactly as before.

### Bit-exactness

A wrong twiddle here produces a wrong codeword that still round-trips, because the inverse reads the same wrong twiddle — the existing round-trip tests are blind to it.

The new test therefore A/Bs the scheduled transform against a reference that does one full pass per stage straight off the twiddle accessor, at every branch boundary of the cut (`log_n` at the cut, one above it, one full group, a group plus a one-stage leftover, several groups over several tiles), at widths 1, 3, 16 and 64, and in both directions. A second test pins the folded conversions to standalone conversion passes over the same set of cuts.

I checked that the test bites: perturbing the fused group's block index by one makes it fail while every existing oracle, twiddle-walk and round-trip test still passes.

### Notes for the reviewer

- The gather and scatter are the only raw-pointer code. Slice splitters cut a slice into contiguous pieces only, and a task's rows are a stride apart, so the tasks share one `Send + Sync` handle and address their own rows. The row sets are disjoint by the mixed-radix argument above, and each walk is bounds-checked against the row count once per tile (a hard assert, not a debug one), so a bad index panics instead of writing out of the matrix.
- Parallelism goes through `p3-maybe-rayon` as the rest of the crate does, with one staging buffer per worker via `for_each_init`, and the crate still builds and tests correctly with `parallel` off.
- No intrinsics and no new target-feature gates: this is a scheduling change, the butterfly leaf is untouched.

## Test plan

Machine: AMD Ryzen 9 9950X3D (Zen 5), 16C/32T, 32 rayon threads. **The box was shared and its load average moved between 3 and 28 during the session**, so single benchmark runs swing; every number below is the best of two back-to-back runs of the same shape, taken with the box quiet, and the direction of every ratio reproduced across runs.

- [x] `cargo test -p p3-binary-dft` — ok, 50 unit + 3 integration passed
- [x] `cargo test -p p3-binary-dft --features parallel` — ok, 50 + 3 passed
- [x] `RUSTFLAGS="-C target-feature=+pclmulqdq" cargo test -p p3-binary-dft --features parallel` — ok, 50 + 3 passed (this is the configuration where `PolyBasisNtt` actually takes its own path rather than the tower fallback)
- [x] `RUSTFLAGS="-C target-feature=+pclmulqdq" cargo test -p p3-binary-dft` — ok, 50 + 3 passed
- [x] `cargo test -p p3-binary-dft --doc` — ok
- [x] `cargo +stable clippy -p p3-binary-dft --all-targets -- -D warnings` — clean, and clean again with `--features parallel`
- [x] `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo +stable doc --no-deps -p p3-binary-dft --document-private-items` — clean
- [x] `cargo test -p p3-binary-pcs` and `RUSTFLAGS="-C target-feature=+pclmulqdq" cargo test -p p3-binary-pcs --features parallel` — ok, 37 + 19 passed
- [x] `RUSTFLAGS="-C target-feature=+pclmulqdq" cargo test -p p3-whir --features parallel` — ok, 131 passed
- [x] Mutation check: perturbing the fused group's block index fails the new schedule test and nothing else
- [x] `RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-binary-dft --features parallel --bench ntt`, forward transform of a width-16 matrix:

  |---|---|---|---|
  | `ntt/128/hybrid/16` (2^16 rows, 16 MiB) | 3.58 ms | 2.17 ms | 1.65x |
  | `ntt/128/hybrid/20` (2^20 rows, 256 MiB) | 146.8 ms | 39.2 ms | 3.74x |

- [x] Small inputs checked explicitly on the `poly/forward` group, same flags, three runs each side:

  |---|---|---|---|
  | 2^4 rows, width 64 (16 KiB, below the tile cut) | 7.65 us | 7.50 us | 1.02x |
  | 2^8 rows, width 1 (4 KiB, below the tile cut) | 4.02 us | 3.20 us | 1.26x |
  | 2^12 rows, width 1 (64 KiB, one stage above the cut) | 81.8 us | 69.6 us | 1.18x |
  | 2^12 rows, width 16 (1 MiB) | 600 us | 215 us | 2.79x |
  | 2^16 rows, width 64 (64 MiB) | 16.2 ms | 7.55 ms | 2.15x |

  Nothing regressed. Earlier runs did show apparent regressions of 20-60% at 2^4 and 2^8; those were load artifacts — re-running the same binary three times on a quiet box put both sides on top of each other.

### DIFF
--- a/binary-dft/src/poly.rs
+++ b/binary-dft/src/poly.rs
@@ -204,6 +204,26 @@ const TILE_BYTES: usize = 32 * 1024;
 /// A 1 MiB tile fuses 12 stages and runs a third slower.
 const STAGING_BYTES: usize = 64 * 1024;
 
+/// Bytes a row must carry before a staging tile is worth gathering it into.
+///
+/// A gather and a scatter address one row at a time, and the rows they address are a power of
+/// two apart, so each one pulls and pushes at least a whole cache line.
+///
+/// A row shorter than a line therefore moves a line's worth of traffic to move a fraction of
+/// a line's worth of data:
+///
+/// ```text
+///     16 bytes per row    one quarter of a 64-byte line is useful
+///     64 bytes per row    a whole line is useful
+/// ```
+///
+/// At one element per row that waste outweighs the full passes the fusion removes, so such a
+/// shape keeps the plain per-stage passes instead.
+///
+/// This is the smallest line size the supported targets have, so it is the point past which no
+/// target wastes more than half of a line.
+const STAGED_ROW_BYTES: usize = 64;
+
 /// The shape of one transform, and the two places its stage sequence is cut.
 ///
 /// A stage pairs rows a power of two apart, so the whole matrix has to be traversed once per
@@ -231,6 +251,8 @@ struct Plan {
     /// Bottom stages that run to completion inside one contiguous tile of rows.
     local: usize,
     /// Long-stride stages that one staging tile fuses into a single pass over the matrix.
+    ///
+    /// Zero when a row is too narrow for a gather to move whole cache lines.
     depth: usize,
 }
 
@@ -240,11 +262,18 @@ impl Plan {
         let element = core::mem::size_of::<u128>();
         let tile_rows = (TILE_BYTES / element / width).max(1);
         let staging_rows = (STAGING_BYTES / element / width).max(1);
+        // A gather moves whole cache lines whatever the row length, so a row below the line
+        // size wastes more traffic than the fused passes save.
+        let row_fills_a_line = element * width >= STAGED_ROW_BYTES;
         Self {
             width,
             log_n,
             local: log2_floor_usize(tile_rows).min(log_n),
-            depth: log2_floor_usize(staging_rows),
+            depth: if row_fills_a_line {
+                log2_floor_usize(staging_rows)
+            } else {
+                0
+            },
         }
     }
 
@@ -979,12 +1008,12 @@ mod tests {
         //
         // Fixture state, from the two cache budgets:
         //
-        //     width    1 @ 2^10   local 10, depth 12   the tile is the whole transform
-        //     width    3 @ 2^10   local  9, depth 10   one stage above the tile, unfused
-        //     width   16 @ 2^10   local  7, depth  8   one group of three stages
-        //     width   64 @ 2^10   local  5, depth  6   one group of five stages
-        //     width  512 @ 2^8    local  2, depth  3   two groups of three stages
-        //     width 1024 @ 2^7    local  1, depth  2   three groups of two stages
+        //     width    1 @ 2^10   local 10, depth 0   the tile is the whole transform
+        //     width    3 @ 2^10   local  9, depth 0   rows under a line, so nothing fuses
+        //     width   16 @ 2^10   local  7, depth 8   one group of three stages
+        //     width   64 @ 2^10   local  5, depth 6   one group of five stages
+        //     width  512 @ 2^8    local  2, depth 3   two groups of three stages
+        //     width 1024 @ 2^7    local  1, depth 2   three groups of two stages
         for (width, log_n) in [(1, 10), (3, 10), (16, 10), (64, 10), (512, 8), (1024, 7)] {
             let plan = Plan::new(width, log_n);
             for inverse in [false, true] {
--- a/binary-dft/src/poly.rs
+++ b/binary-dft/src/poly.rs
@@ -386,23 +386,31 @@ impl Rows {
 
     /// Copy the rows `first`, `first + stride`, ... into consecutive rows of the tile.
     ///
+    /// The tile is emptied first and then grown one row at a time.
+    ///
+    /// So it holds no element the walk did not write.
+    ///
+    /// And a worker never has to zero a tile it is about to overwrite in full.
+    ///
+    /// A tile whose capacity already covers the walk grows without reallocating.
+    ///
     /// # Safety
     /// No other live task may address any of the rows the walk names.
-    unsafe fn gather(&self, first: usize, stride: usize, tile: &mut [u128]) {
-        let rows = tile.chunks_exact_mut(self.width);
-        self.check(first, stride, rows.len());
-        for (k, row) in rows.enumerate() {
+    unsafe fn gather(&self, first: usize, stride: usize, rows: usize, tile: &mut Vec<u128>) {
+        self.check(first, stride, rows);
+        tile.clear();
+        for k in 0..rows {
             // SAFETY: the bound above puts every row of the walk inside the matrix, and the
-            // tile is a separate allocation, so the two ranges cannot overlap.
+            // exclusive borrow the base pointer came from outlives the task.
             //
-            // The row is a whole chunk, so it has room for exactly the elements copied into it.
-            unsafe {
-                core::ptr::copy_nonoverlapping(
+            // No other live task addresses this row, so nothing can write it during the read.
+            let row = unsafe {
+                core::slice::from_raw_parts(
                     self.base.add((first + k * stride) * self.width),
-                    row.as_mut_ptr(),
                     self.width,
-                );
-            }
+                )
+            };
+            tile.extend_from_slice(row);
         }
     }
 
@@ -504,7 +512,10 @@ fn fused_stages(
         // exactly one task.
         //
         // The exclusive borrow of the matrix outlives the whole region.
-        unsafe { rows.gather(first, stride, tile) };
+        unsafe { rows.gather(first, stride, 1 << depth, tile) };
+        // Invariant: every element of the tile comes from the walk the gather just ran, so
+        // nothing below reads an element the gather did not write.
+        debug_assert_eq!(tile.len(), tile_len, "the gather left the tile short");
         // The gather is the first read of every element when this is the first group of a
         // forward transform.
         if convert_basis && !inverse {
@@ -523,7 +534,11 @@ fn fused_stages(
 
     // One staging tile per worker, not per task: a task is a few tens of microseconds of
     // work and the tile is tens of kilobytes.
-    let new_tile = || alloc::vec![0u128; tile_len];
+    //
+    // The tile is capacity only, with no initialized elements: the gather grows it from
+    // empty, so no worker spends bandwidth zeroing tens of kilobytes it is about to
+    // overwrite in full.
+    let new_tile = || Vec::with_capacity(tile_len);
     if use_parallel(len.saturating_mul(depth)) {
         (0..tiles).into_par_iter().for_each_init(new_tile, task);
     } else {
--- a/binary-dft/src/poly.rs
+++ b/binary-dft/src/poly.rs
@@ -7,7 +7,7 @@ use p3_field::PrimeCharacteristicRing;
 use p3_matrix::Matrix;
 use p3_matrix::dense::RowMajorMatrix;
 use p3_maybe_rayon::prelude::*;
-use p3_util::log2_strict_usize;
+use p3_util::{log2_floor_usize, log2_strict_usize};
 
 use crate::domain::domain_point;
 use crate::lch::BUTTERFLY_GRAIN;
@@ -91,6 +91,18 @@ fn for_chunks(
     }
 }
 
+/// The polynomial-basis coordinates of an element held in the tower basis.
+#[inline]
+fn into_poly(value: u128) -> u128 {
+    poly_basis::from_tower(BinaryField128::from_repr(value))
+}
+
+/// The tower-basis bit pattern of an element held in polynomial coordinates.
+#[inline]
+fn into_tower(value: u128) -> u128 {
+    poly_basis::to_tower(value).to_repr()
+}
+
 fn convert(values: &mut [u128], conversion: impl Fn(u128) -> u128 + Send + Sync) {
     // Basis conversion does several dependent lookups per element, more work than
     // a butterfly, so it amortizes dispatch at a smaller byte volume.
@@ -103,6 +115,44 @@ fn convert(values: &mut [u128], conversion: impl Fn(u128) -> u128 + Send + Sync)
             .iter_mut()
             .for_each(|value| *value = conversion(*value));
     }
+}
+
+/// Apply a per-element map to a tile the schedule is already holding in cache.
+fn convert_tile(tile: &mut [u128], conversion: impl Fn(u128) -> u128) {
+    for value in tile {
+        *value = conversion(*value);
+    }
+}
+
+/// Where the two basis conversions ride, instead of taking a pass over the matrix each.
+///
+/// A conversion is a pure per-element map, so it commutes with every butterfly ordering.
+/// Applied where the schedule already holds the element in cache it costs the lookups alone,
+/// and no memory traffic at all.
+#[derive(Copy, Clone, Debug)]
+struct Fold {
+    /// Map out of the tower basis where the transform first reads each element.
+    entry: bool,
+    /// Map back into the tower basis where the transform last writes each element.
+    exit: bool,
+}
+
+impl Fold {
+    /// Take tower-basis values and hand tower-basis values back.
+    const BOTH: Self = Self {
+        entry: true,
+        exit: true,
+    };
+    /// Take tower-basis values and leave the result in polynomial coordinates.
+    const ENTRY: Self = Self {
+        entry: true,
+        exit: false,
+    };
+    /// Take polynomial coordinates and hand tower-basis values back.
+    const EXIT: Self = Self {
+        entry: false,
+        exit: true,
+    };
 }
 
 /// A stage uses bounded chunks, each starting from its own independently indexed twiddle.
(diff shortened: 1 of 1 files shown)


binary-dft/src/poly.rs
```diff
@@ -163,63 +282,345 @@ fn local_stage(
...
+///     row(k) >> (top - s) = block * 2^s + (k >> (depth - s))
+/// ```
+///
+/// Writing `k = q * 2^(depth-s) + r` gives `offset + r * S < 2^(top-s)`, so only
+/// `q = k >> (depth-s)` survives the shift.
+///
+/// That is precisely the block index a contiguous run of `2^(depth-s)` staged rows carries.
+///
+/// So the tile runs as an ordinary radix-2 network whose twiddle walk starts at `block << s`.
+fn fused_stages(
```
Nashtare, line 505: I verified the schedule independently rather than trusting the round trip, since as you note the inverse would repeat any twiddle mistake.

**The index map is a bijection.** With `S = 1 << (top - depth)`, `block = index >> (top - depth)`, `first = (block << top) + (index & (S - 1))`, the walk `row(k) = first + k*S` over `index in 0..2^(log_n - depth)`, `k in 0..2^depth` is a mixed-radix decomposition of `0..2^log_n`. Every row is in exactly one tile, which is what makes the raw-pointer `Rows` sharing sound.

**The twiddle index is the global one.** For sub-layer `s` (stage `j = top - 1 - s`) the required block index is `row(k) >> (top - s)`. Writing `k = q*2^(depth-s) + r`, the low part is bounded by `(S-1) + (2^(depth-s) - 1)*S = 2^(top-s) - 1`, so it never carries and `row(k) >> (top-s) = block*2^s + q` with `q = k >> (depth-s)`. `tile_stages` (`:294`) seeds `local_stage` with `first = block << s`, whose chunk index is exactly `q` — so the twiddle is `block*2^s + q`, and `offset` correctly drops out. `local_stages` is the `top = depth = local` case and reduces to the base's `tile << (local-j-1)` seed.

**Checked empirically against three independent references**, not just against itself: a throwaway harness linking this head *and* a copy of the base-commit crate side by side, comparing `shifted_ntt_batch` / `shifted_intt_batch` / `shifted_lde_batch` / `ntt_batch_padded` byte-for-byte, plus forward and inverse separately against the untouched `LchNtt` (a different algorithm) and against `NaiveAdditiveNtt` at small heights. Widths `{1,2,3,4,5,7,16,17,64,100,256,512,1024,2048,4096,5000}` x `log_n 0..=18` x 3 shifts, with `debug-assertions` and `overflow-checks` on, serially and at `RAYON_NUM_THREADS in {2,3,5,8,16}` — **6356 checks, 0 failures**, covering 136 no-fusion / 80 single-group / 21 multi-group / 12 leftover-with-groups configurations.

Also clean: the `depth < 2` short-circuit in `leftover()` runs before `above % depth`, so no divide-by-zero at width >= 4096; forward's `(local..top).rev()` and inverse's peel-up from `local + leftover` are exactly complementary with no gap or doubled stage; `take >= 2` always, so no degenerate one-stage fusion; and the coset shift enters only `shifts[j]`, once per stage rather than once per fused stage.

binary-dft/src/poly.rs
```diff
@@ -163,63 +282,345 @@ fn local_stage(
...
+            rows == 0 || first + (rows - 1) * stride < self.count,
+            "staged row walk leaves the matrix"
+        );
+    }
+
+    /// Copy the rows `first`, `first + stride`, ... into consecutive rows of the tile.
+    ///
+    /// # Safety
+    /// No other live task may address any of the rows the walk names.
+    unsafe fn gather(&self, first: usize, stride: usize, tile: &mut [u128]) {
```
Nashtare: Width 1 looks like a small regression rather than a win, because at `width = 1` a "row" is 16 bytes and `gather`/`scatter` touch a whole cache line (128 B on Apple Silicon) per useful 16 bytes — roughly 8x traffic amplification, plus one page touched per staged row, against the `depth` full passes it removes.

Min-of-11 interleaved head-vs-base, `--features parallel`, `shifted_ntt_batch`, `width = 1`: `log_n = 21` -> **0.85x, 0.87x** (two runs); `log_n = 22` -> 0.87x, 0.87x, 0.89x, 0.98x; `log_n = 23` -> 0.91x, 1.01x. `log_n <= 20` is parity to 1.2x. Width >= 8 is uniformly a win, so this is narrow.

Caveat: measured on a laptop, not a quiet box — the `log_n = 21/22` result did reproduce across runs, but it is worth confirming on your Zen 5 host before acting. If it holds, gating fusion on `width * size_of::<u128>()` reaching a cache line (or staging several consecutive rows per unit at small widths) would cover it. Your own small-input table stops at `2^12` for width 1, so it would not have caught this.

binary-dft/src/poly.rs
```diff
@@ -269,31 +668,30 @@ impl AdditiveNtt<BinaryField128> for PolyBasisNtt {
             .into_iter()
             .map(BinaryField128::to_repr)
             .collect();
+        // A coset transform converts its own copy of the message, so the copies move
+        // tower-basis rows and no conversion pass is needed at either end.
+        let plan = Plan::new(width, log_message);
```
Nashtare, line 720: `ntt_batch_padded` trades more conversion work than the other paths do. Base ran `L` entry conversions (message only) plus `L * 2^rate` exit conversions; head runs `L * 2^rate` entry **and** `L * 2^rate` exit, so it pays an extra `L * (2^rate - 1)` basis conversions — each ~16 dependent table lookups — to save two memory passes.

That shows up in the measurements: `ntt_batch_padded` gains only 1.03-1.27x at width >= 4, against 1.9-2.1x for `shifted_lde_batch` on the same host, and at `rate = 3, width = 1, log_n = 20` I measured **0.81x**.

Keeping the single message-side entry conversion and folding only the exit would get the memory win without the extra conversions.

binary-dft/src/poly.rs
```diff
@@ -163,63 +282,345 @@ fn local_stage(
...
+            convert_tile(tile, into_tower);
+        }
+        // SAFETY: the rows are the ones the gather read, so the argument above applies
+        // unchanged.
+        unsafe { rows.scatter(first, stride, tile) };
+    };
+
+    // One staging tile per worker, not per task: a task is a few tens of microseconds of
+    // work and the tile is tens of kilobytes.
+    let new_tile = || alloc::vec![0u128; tile_len];
```
Nashtare: `vec![0u128; tile_len]` zeroes a tile that `gather` overwrites in full before anything reads it, and the closure is re-run per fused group per worker — up to 4 groups x N workers x 64 KiB of pointless zeroing in one transform.

`Vec::with_capacity` + `set_len` is the `unsafe` route; if you would rather avoid that, hoisting one buffer per worker across all groups (rather than per group) at least pays it once.

binary-dft/src/poly.rs
```diff
@@ -142,6 +193,74 @@ fn stage(values: &mut [u128], half: usize, j: usize, twiddles: &Twiddles, invers
...
+///
+/// A stage pairs rows a power of two apart, so the whole matrix has to be traversed once per
+/// stage unless a set of rows closed under several stages can be brought into cache.
+///
+/// There are two such sets, and one cut point each:
+///
+/// ```text
+///     stages log_n-1 .. local     rows spaced far apart, gathered into a staging tile
+///     stages local-1 .. 0         rows already adjacent, so a contiguous tile holds them
+/// ```
```
Nashtare, line 238: This table says stages `log_n-1 .. local` are "gathered into a staging tile", but that is not true of the `leftover` stage(s), which sit in exactly that range and deliberately run as plain full passes — as the doc on `leftover` says at `:246` ("Stages above the contiguous tile that run as plain full passes").

Since the table is the first thing a reader meets, a third row (or a trailing clause on the first) noting the leftover exception would stop the two from contradicting each other.

binary-dft/src/poly.rs
```diff
@@ -365,29 +760,103 @@ impl AdditiveNtt<BinaryField128> for PolyBasisNtt {
...
+    /// Cut points small enough to keep the test matrices tiny, one pair per branch of the
+    /// schedule:
+    ///
+    /// ```text
+    ///     (2, 3)  full groups, plus a one-stage leftover at local + depth + 1
+    ///     (2, 0)  a staging tile too narrow for two rows, so no group runs at all
+    ///     (0, 3)  no contiguous tile, so every stage is fused
+    ///     (1, 1)  a depth of one, which never pays for a tile
+    /// ```
+    const CUTS: [(usize, usize); 4] = [(2, 3), (2, 0), (0, 3), (1, 1)];
```
Nashtare, line 848: No test in the file reaches three or more fused groups: `CUTS` tops out at `(2, 3)`, and the deepest production fixture is `(512, 8)` at `:925`, which gives two groups.

I ran `[3,3,2]`, `[3,3,3]`, `[2,2,2]` and `[2,2,2,2]` externally with zero failures, so this is a coverage note rather than a defect — but the fused loop's `while` structure is exactly the thing a third iteration would stress, and adding `(512, 11)` or `(1024, 9)` to the fixture list at `:925` closes it for free.

Nashtare (approved):

binary-dft/src/poly.rs
tcoratger: Thanks for deriving it rather than trusting the round trip — that is exactly the argument that has to hold, and it is the one the code carries. The mixed-radix decomposition is the `SAFETY` block on the task (`poly.rs:537`), the disjointness it buys sits directly above the `Send`/`Sync` impls on `Rows` (`poly.rs:387`), and the `row(k) >> (top - s) = block * 2^s + q` step is the `# Twiddles` section on `fused_stages` (`poly.rs:489`). Nothing to add.

One thing that shifts your configuration split: `Plan::new` now gates fusion on a row reaching a cache line, so at `width * 16 < 64` the production planner returns `depth = 0` and takes the plain per-stage path. Your fused-schedule coverage at widths 1 and 3 still stands, because the tests build `Plan` directly instead of going through `Plan::new`.

binary-dft/src/poly.rs
tcoratger: Taken as given and gated. `Plan::new` (`poly.rs:267`) now requires `size_of::<u128>() * width >= 64` before it plans a staging tile at all. Below that `depth` is 0, `leftover()` returns every stage above the contiguous tile, and the schedule is the plain per-stage one — so widths 1 to 3 no longer fuse and width 4 and up are untouched.

I chose 64 rather than 128 on your own numbers: `shifted_lde_batch` at width 4 was 1.9-2.1x on a machine with 128-byte lines, so a 64-byte row already pays even at half a line. Only the 16-byte row does not.

Being explicit about what I did *not* do: I did not reproduce the 0.85-0.91x on the Zen 5 host. No benchmarks were run in this pass at all — the box is thermally saturated at the moment and I was asked to keep this to compile-and-test — so the gate rests on your measurement plus the traffic argument, not on a number of mine. Same reason the small-input table is still unextended; I would rather add the width-1 rows past 2^12 together with real numbers than land the rows blind.

Legs I ran, once each: `cargo nextest run -p p3-binary-dft` (53 pass), the same with `--features parallel` (53 pass), the same again under `RUSTFLAGS=-C target-feature=+pclmulqdq` with `RAYON_NUM_THREADS=4` so the poly-basis path is the one actually exercised (53 pass), `cargo clippy -p p3-binary-dft --all-targets -- -D warnings` (clean), `cargo +nightly fmt --all`. Not run: any `cargo bench`, the `vpclmulqdq`/`avx512f` legs, the thread sweep beyond 4, and the other crates' suites.

binary-dft/src/poly.rs
tcoratger: Fixed the way you suggested. `poly.rs:727` converts the message alone, and every coset then runs `Fold::EXIT`, so the padded path pays `L` entry conversions plus `L * 2^rate` exit conversions — the same count as before this PR — while each coset's exit conversion still rides its contiguous tile instead of taking a pass of its own. The extra `L * (2^rate - 1)` is gone.

The count is arithmetic rather than a measurement, which is why I was comfortable making the change without re-benching: one conversion of `L` elements against `2^rate` of them is a static property of the call, and the memory pass the fold was buying is unchanged. I did not re-run the rate-3 width-1 arm on this host — no benchmarks in this pass, for the reason in the width-1 thread.

binary-dft/src/poly.rs
tcoratger: Fixed, though not by the hoist. `no_std` plus rayon leaves no safe way to keep one buffer per worker alive across separate parallel regions: `for_each_init` re-runs its init once per region, and a shared pool would need interior mutability the crate cannot reach without `unsafe`.

So the zeroing is gone outright instead. `new_tile` is `Vec::with_capacity(tile_len)` (`poly.rs:570`), and `gather` clears the tile and grows it one row at a time with `extend_from_slice` (`poly.rs:428`). No `set_len` and no new `unsafe` — and the invariant is structural rather than asserted, since the tile has no initialized element until the gather writes it, with `debug_assert_eq!(tile.len(), tile_len)` at `poly.rs:547` pinning that before any sub-layer reads it.

What is left per group per worker is a 64 KiB malloc with no memset, against the ~8 MiB of memset a four-group thirty-two-worker transform was doing. I did not measure the difference — no benchmarks in this pass, for the reason in the width-1 thread.

binary-dft/src/poly.rs
tcoratger: Fixed at `poly.rs:235`. Three bands now:

```text
    stages log_n-1 .. local+leftover   rows far apart, gathered into a staging tile
    stages local+leftover-1 .. local   rows far apart, one plain full pass each
    stages local-1 .. 0                rows already adjacent, so a contiguous tile holds them
```

with two lines under it saying the middle band is one stage wide at most unless the staging tile cannot hold two rows, which is the `depth < 2` arm of `leftover()`.

binary-dft/src/poly.rs
tcoratger: Added, with exactly the shapes you ran. `DEEP_CUTS` at `poly.rs:862` is `(0,3,8) -> 3+3+2`, `(0,3,9) -> 3+3+3`, `(1,2,8) -> 2+2+2` with a one-stage leftover, and `(0,2,8) -> 2+2+2+2`.

Both the schedule test and the conversion test now walk one shared plan enumerator (`cut_plans`, `poly.rs:956`) instead of each rebuilding the loop, so the deep cuts land in both. The conversion test is the one that needed it most: three groups is what distinguishes "entry rides the first group and exit the last" from a reseeding that happens to be right when first and last are the only groups.

For the production fixture I took `(1024, 7)` over `(512, 11)` and `(1024, 9)`. At width 1024 the budgets give `local 1, depth 2`, so `2^7` is the smallest production shape that reaches three groups, and it keeps the fixture the same element count as the existing `(512, 8)` rather than 8-16x larger.

Every added height is at most `2^9`, so the two tests still run in 0.383s and 0.429s under `cargo nextest run -p p3-binary-dft`.

PR #2054, opened 2026-09-07
perf(binary-field): vectorize the polynomial-basis butterflies with a split twiddle

tcoratger:
## Summary

- `poly_basis`'s three slice kernels — the scaling and both butterflies — walked one 128-bit element at a time through the scalar carryless multiply, even where the host's wide carryless multiply covers four field elements per instruction. They now run a packed prefix and the existing scalar tail.
- Four elements per register under `avx512f + vpclmulqdq`, two under `avx2 + vpclmulqdq`, one at a time everywhere else. The current scalar loop is both the tail and the portable fallback, unchanged.
- The multiplier is fixed across a slice, so the kernel derives its `x^64` companion once per call and uses the split form: 5 carryless multiplies per four elements against 24 before, and one reduction step instead of two.
- No public signature changes and no `binary-dft` changes. The zero-scalar and unit-length shortcuts are byte-for-byte the code they were, so no shape executes more work than before.
- Lane loads and stores are unaligned pointer casts over `[u128; WIDTH]` blocks, not slice transmutes; every `unsafe` block carries a `// SAFETY:` note tied to the target features that gate the module.

### Why one fold suffices

Reduction modulo `p = x^128 + x^7 + x^2 + x + 1` is a ring homomorphism, so a shift can be moved off the varying operand and onto the fixed one. Splitting `v = v_0 + v_1 x^64` and precomputing the companion `u = t x^64 mod p` turns the product into `t v = t v_0 + u v_1 (mod p)` — two `128 x 64` products rather than one `128 x 128`.

Their sum has degree at most `127 + 63 = 190`, so it occupies three 64-bit limbs instead of four. A four-limb product needs two Horner steps in `x^64` to come back under `x^128`; three limbs need one.

That single step is exact because the piece it shifts is only 128 bits wide: writing it as `h_0 + h_1 x^64`, the rewrite `h_0 x^64 + h_1 T` has degree at most 127 and 70 respectively, so nothing spills over the top a second time.

## Test plan

Host: AMD Ryzen 9 9950X3D (Zen 5, AVX-512 + VPCLMULQDQ + GFNI + VAES), 32 threads. **The machine was shared with other builds throughout**, so every figure below is the minimum of three interleaved before/after rounds run from two pre-built benchmark binaries, which cancels drift between the arms. Ratios under ~1.1x should be read as "no measurable change".

- [x] `RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-binary-field --bench arithmetic -- batch/poly128`

  |---|---|---|---|
  | `mul/1` | 2.48 ns | 2.63 ns | 0.95x |
  | `mul/16` | 40.35 ns | 12.23 ns | **3.30x** |
  | `mul/256` | 646.7 ns | 144.6 ns | **4.47x** |
  | `mul/4096` | 9.839 µs | 2.233 µs | **4.41x** |
  | `butterfly/1` | 8.57 ns | 7.47 ns | 1.15x |
  | `butterfly/16` | 42.08 ns | 22.28 ns | **1.89x** |
  | `butterfly/256` | 716.1 ns | 174.4 ns | **4.11x** |
  | `butterfly/4096` | 11.776 µs | 2.577 µs | **4.57x** |

  Length 1 takes the pre-existing unit-length shortcut and never reaches the new code, so the two length-1 rows are layout and measurement noise on a loaded host — the baseline's own three runs spread 2.5–3.3 ns and 8.6–10.4 ns. An earlier build in which the packed kernel was inlinable did regress `mul/1` to 4.6 ns by making the public wrapper too large to inline; the kernel is now behind an `#[inline(never)]` boundary, which restores parity and is a small win at every other length too.

- [x] `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq,+pclmulqdq" cargo bench -p p3-binary-field --bench arithmetic -- batch/poly128` — the separate AVX2-only CI leg, two lanes per register:

  |---|---|---|---|
  | `mul/1` | 2.64 ns | 2.63 ns | 1.00x |
  | `mul/16` | 37.03 ns | 17.84 ns | **2.08x** |
  | `mul/256` | 615.4 ns | 256.4 ns | **2.40x** |
  | `mul/4096` | 9.401 µs | 4.825 µs | **1.95x** |
  | `butterfly/1` | 7.20 ns | 7.71 ns | 0.93x |
  | `butterfly/16` | 40.65 ns | 23.85 ns | **1.70x** |
  | `butterfly/256` | 717.1 ns | 377.9 ns | **1.90x** |
  | `butterfly/4096` | 10.290 µs | 4.738 µs | **2.17x** |

- [x] `RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-binary-dft --features parallel --bench ntt -- ntt/128/hybrid` — end to end, the transform that uses these kernels:

  |---|---|---|---|
  | `2^14`, 1 thread | 5.45 ms | 2.29 ms | **2.38x** |
  | `2^16`, 1 thread | 24.63 ms | 10.44 ms | **2.36x** |
  | `2^18`, 1 thread | 110.6 ms | 62.1 ms | **1.78x** |
  | `2^14`, 32 threads | 1.624 ms | 1.605 ms | 1.01x |
  | `2^16`, 32 threads | 4.852 ms | 3.910 ms | **1.24x** |
  | `2^18`, 32 threads | 17.345 ms | 14.270 ms | **1.22x** |
  | `2^20`, 32 threads | 161.2 ms | 157.2 ms | 1.03x |

  Single-threaded the transform is butterfly-bound and the kernel factor passes through at 1.8x–2.4x. With all 32 threads it is bound by memory bandwidth and parallel dispatch instead — 2^20 carries a 256 MiB matrix and 2^14 gets only 4.2x from 32 threads — so the kernel win only partly shows, and at the two extremes not at all. That is well short of the 1.6x–5.4x an offline study predicted for the shipped parallel path, and worth saying plainly.

- [x] `cargo test -p p3-binary-field` — 25 test binaries across the five feature legs below, 0 failures. (`cargo nextest` is not installed here; `--features parallel` does not exist on this crate yet, so it was not exercised.)
- [x] `cargo test -p p3-binary-field --doc` — ok
- [x] `cargo test -p p3-binary-dft` and `cargo test -p p3-binary-dft --features parallel` — 48 + 3 tests, ok
- [x] `cargo +stable clippy -p p3-binary-field --all-targets -- -D warnings` — clean on every leg below
- [x] `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo +stable doc --no-deps -p p3-binary-field --document-private-items` — clean

- [x] Per-target-feature matrix, since the new code is target-gated. Tests and clippy were run under each:

  | `RUSTFLAGS` | path taken | tests | clippy |
  |---|---|---|---|
  | `-C target-cpu=native` | packed, 4 lanes | ok | ok |
  | `-C target-feature=+avx2,+vpclmulqdq` | packed, 2 lanes | ok | ok |
  | `-C target-feature=+avx2` | scalar fallback | ok | ok |
  | `-C target-feature=+pclmulqdq` | scalar fallback | ok | ok |
  | *(none)* | portable fallback | ok | ok |

- [x] New proptests in `poly_slice` pin the packed kernels to the scalar backend element for element: a deterministic sweep over every length from 0 to 17 crossed with seven corner multipliers (`0`, `1`, all ones, `x^127`, `x^64`, lower half only, the modulus tail) and four payload offsets, plus random slices up to length 40. Lengths cover an empty prefix, whole registers, and a scalar tail of every residue on both the two-lane and four-lane builds; the offsets walk the prefix across every alignment a run of 128-bit integers can have inside one register. A separate case pins the zero multiplier, and the round trip of the two butterflies is asserted on every case.
- [x] `poly_basis`'s existing tower-arithmetic test gained the lengths `{4, 5, 7, 15, 16}` and the multipliers `{0x87, x^127}` so its own sweep exercises the packed prefix and the tail.
- [x] Mutation-checked, on both the four-lane and two-lane builds: replacing the companion with the bare multiplier, or reading the wrong half of it, fails the new proptests and the `poly_basis` sweep immediately. This confirms the packed path is the one under test on both widths.


binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+            );
+
+            // Limbs 1 and 2: the half products weighted by `x^64`.
+            let high = lanes::xor(
+                lanes::clmul::<LOW_BY_HIGH>(v, self.t),
+                lanes::clmul::<HIGH_BY_HIGH>(v, self.t_x64),
+            );
+
+            // Three limbs, so one Horner step in `x^64` finishes the reduction.
+            PackedGhash128::fold_shifted(low, high)
```
Nashtare: **On `x86_64-apple-darwin` this kernel emits a libc call inside the hot loop and is a large regression, not a 1.95x win.**

`fold_shifted` reads `Self::tail()` (`packed/x86_64.rs:285-287`), which is `transmute::<[u128; WIDTH], lanes::Reg>([TAIL_128; WIDTH])`. On the Apple target LLVM materialises that array constant through `_memset_pattern16` on the stack and then fails to hoist it out of the loop.

I verified this by emitting the assembly, not by reading:

```
RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" \
  cargo rustc --release -p p3-binary-field --target x86_64-apple-darwin --lib -- --emit asm
```

Six `_memset_pattern16` call sites in the whole object file, and all six are in the three new kernels — two each in `poly_slice::wide::{scale, butterfly_forward, butterfly_inverse}`, one in the prologue and one in the loop body. The `scale` inner loop:

```asm
LBB4_4:
        vmovdqu (%r14,%rbx), %ymm0
        vpclmulqdq $0,  %ymm3, %ymm0, %ymm1
        ...
        vmovdqa %ymm0, 128(%rsp)
        movl    $32, %edx
        movq    %r12, %rdi
        movq    %r15, %rsi
        vzeroupper                 ; AVX->SSE transition, every iteration
        callq   _memset_pattern16  ; libc call, every iteration
        ...
        jne     LBB4_4
```

So per two elements the "vectorized" path pays a call/ret into libc, a `vzeroupper`, and two ymm stack spills — against a ~6-clmul scalar loop before. On an Ice Lake Intel Mac (the Apple configuration that actually has VPCLMULQDQ) built with `-C target-cpu=native`, this is slower than what it replaces.

The Linux legs are fine — same flags, `x86_64-unknown-linux-gnu` emits **zero** memset calls and a textbook 2x-unrolled loop with the twiddle hoisted into `ymm0/ymm1/ymm2`:

```asm
.LBB4_6:
        vmovdqu (%rcx), %ymm3
        vpclmulqdq $0,  %ymm0, %ymm3, %ymm5
        vpclmulqdq $1,  %ymm2, %ymm3, %ymm6
        vpxor   %ymm5, %ymm6, %ymm5
        ...
```

That is why the benchmark table did not catch it: every measurement was taken on Zen 5 Linux.

One-line fix — build `tail()` from a broadcast intrinsic instead of transmuting an array constant: `_mm256_broadcastsi128_si256(_mm_set_epi64x(0, TAIL_128 as i64))` / `_mm512_broadcast_i32x4(...)`. The constant is loop-invariant and LLVM hoists the intrinsic form on both targets. Worth re-checking the emitted asm on `x86_64-apple-darwin` after the change.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+        *lo ^= clmul::poly_mul_128_batch(scalar, *hi);
+    }
+}
+
+/// The packed prefix, over the widest carryless multiply the target has.
+#[cfg(all(
+    target_arch = "x86_64",
+    target_feature = "vpclmulqdq",
+    any(target_feature = "avx2", target_feature = "avx512f")
+))]
```
Nashtare, line 72: This `cfg` is exactly the predicate that no test-running CI leg satisfies, so **everything in `mod wide` is compiled but never executed by CI**.

The legs in `.github/workflows/ci.yml` that enable `vpclmulqdq` alongside `avx2`/`avx512f` (`:69-88`) are all `compile_only: true`, which runs `cargo build --all-targets` and never `cargo nextest run`. The legs that do run tests are `+avx2` (no vpclmulqdq) and `+pclmulqdq` (no avx2), and on both of those `binary-field/src/packed/mod.rs:41` resolves `Packing` to scalar `Ghash128` and `wide::*` returns 0. So the three new tests — `the_packed_prefix_and_the_scalar_tail_agree_at_every_short_length`, the proptest, and `a_zero_multiplier_degenerates` — all run against the fallback on every leg that executes anything.

The tests themselves are good; the problem is nothing runs them against the code they were written for. A one-character typo swapping `LOW_BY_HIGH` (0x10) and `HIGH_BY_LOW` (0x01) in `apply` would ship silently wrong field elements into `binary-dft`'s NTT and corrupt `p3-binary-pcs` commitments, with fully green CI.

Two ways out, either is fine: flip the `+avx2,+vpclmulqdq` leg to test-running (Zen 3+/Ice Lake+ is common in the runner pool now — guard with `is_x86_feature_detected!` and skip otherwise), or add a target-feature-independent scalar model of `SplitScalar::apply` so the algebra is pinned on every leg.

For what it is worth, I checked the algebra independently since CI cannot: I modelled the exact bit operations (companion derivation, the four clmul immediates, `unpack_low_64(zero, t1)`, the `HIGH_BY_LOW` fold) against a reference GF(2^128) multiply over ~168k (scalar, value) pairs covering all the corners — 0 mismatches, and the degree bounds `low, high < 2^128` held throughout. The split identity `t*v = t*v_0 + (t*x^64)*v_1` with a single `fold_shifted` is correct, and the immediates match the disassembly on both widths.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+        // The second operand of the butterflies, reversed so the two slices differ.
+        let other: Vec<u128> = values.iter().rev().copied().collect();
+
+        // Scaling: the packed prefix and the scalar tail must together match the scalar loop.
+        let mut scaled = shifted(values, offset);
+        super::scale(&mut scaled[offset..], scalar);
+        prop_assert_eq!(&scaled[offset..], &reference_scale(values, scalar)[..]);
+
+        // The zeros ahead of the payload pin down that no access ran off the front.
+        prop_assert!(scaled[..offset].iter().all(|&v| v == 0));
```
Nashtare: The guard is one-sided: `shifted()` pads only the front, and this asserts only that `scaled[..offset]` stayed zero. A store that ran one register *past* the end of the payload would corrupt allocator metadata instead of failing the test — which is the more likely direction of error for a chunked kernel, since `as_chunks_mut` bounds the front but the tail loop is what walks off the end.

Append a few sentinel elements after the payload and assert those are untouched too.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+    // Whatever is left over costs one carryless multiply per element.
+    for value in &mut values[packed..] {
+        *value = clmul::poly_mul_128_batch(*value, scalar);
+    }
+}
+
+/// Send `(lo, hi)` to `(lo + scalar*hi, lo + (scalar + 1)*hi)`, element by element.
+#[inline]
+pub(crate) fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
+    let packed = wide::butterfly_forward(lo, hi, scalar);
```
Nashtare: These two entry points silently truncate to the shorter slice — the `zip` at `:26` (and `:39`) drops the excess, and `wide::butterfly_forward` takes `min(lo.len(), hi.len())`. The length equality is asserted only up in `poly_basis` (`:65`, `:88`), which is now one caller among potentially several.

A `debug_assert_eq!(lo.len(), hi.len())` at the top of both would keep the invariant attached to the code that depends on it.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+        }
+    }
+
+    /// Scale a prefix of the slice, returning the number of elements it covered.
+    ///
+    /// Out of line so that the dispatch above stays small enough to inline into its callers.
+    ///
+    /// That is what keeps the shortcut for a lone element as cheap as a bare product.
+    #[inline(never)]
+    pub(super) fn scale(values: &mut [u128], scalar: u128) -> usize {
```
Nashtare, line 84: The `#[inline(never)]` kernels are still *called* when the slice is shorter than one register, just to compute zero chunks and return 0.

That is not hypothetical: `binary-dft`'s `local_stage` (`binary-dft/src/poly.rs:158-161`) drives butterflies down to `half == 2`, so on an AVX-512 build (`WIDTH == 4`) every one of those `n/4` butterflies pays an out-of-line call/ret for no work. A `if lo.len() < WIDTH { return 0; }` in the inlined dispatcher above would short-circuit it before the call boundary. Same for `scale` and `butterfly_inverse` (`:176`, `:207`).

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,433 @@
...
+
+            // Limbs 1 and 2: the half products weighted by `x^64`.
+            let high = lanes::xor(
+                lanes::clmul::<LOW_BY_HIGH>(v, self.t),
+                lanes::clmul::<HIGH_BY_HIGH>(v, self.t_x64),
+            );
+
+            // Three limbs, so one Horner step in `x^64` finishes the reduction.
+            PackedGhash128::fold_shifted(low, high)
+        }
```
Nashtare: Design question, not a defect — worth an A/B number before taking on 433 lines and a widened crate-internal API.

The split form costs 5 clmuls per register. The existing packed `PackedGhash128::mul` (Karatsuba, `packed/x86_64.rs:359-380`) also costs 5 clmuls per register once the scalar is broadcast and its `mixed_y` is hoisted out as loop-invariant. The split buys roughly one shuffle, one unpack and three xors per register — ALU work that is close to free when VPCLMULQDQ is the throughput bottleneck on Zen 4/5.

The 4.4x / 4.6x figures in the description are against the *scalar* loop (24 clmuls per 4 elements) and I have no reason to doubt them, but they do not establish that this beats a ~40-line `broadcast + Mul` kernel over the same slices. If that comparison was already run, adding the row would settle it; if not, it is worth running before choosing the larger surface.

binary-field/src/poly_basis.rs
```diff
@@ -64,45 +63,43 @@ pub fn mul_slice(values: &mut [u128], scalar: u128) {
 #[inline]
 pub fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
     assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
+    // A zero scalar kills the product, leaving an addition that needs no multiply at all.
```
Nashtare, line 66: Nit: this dropped an intra-doc link (`Undo [\`butterfly_forward\`] with the same scalar.` -> `Undo the forward butterfly with the same scalar.`). `butterfly_forward` is public and in this same module, so the link resolves fine — was this working around something, or incidental? If incidental, keeping the link is the small rustdoc win.

Nashtare (changes requested):

binary-field/src/poly_slice.rs
tcoratger: Fixed, and confirmed by emitting the assembly rather than reading it.

The tail is now built from the broadcast intrinsic you suggested — `_mm_set_epi64x(0, TAIL_128)`
then `_mm256_broadcastsi128_si256` / `_mm512_broadcast_i32x4` — instead of transmuting
`[TAIL_128; WIDTH]`. No array constant, so nothing has to go through memory. It ended up as
`Lanes::tail`'s default body in the new `packed/split.rs`, same intrinsic.

```
RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" \
  cargo rustc --release -p p3-binary-field --target x86_64-apple-darwin --lib -- --emit asm
grep -c memset_pattern16 target/x86_64-apple-darwin/release/deps/p3_binary_field-*.s
```

Six call sites before, zero after. Per kernel, counted inside each symbol body:

| `poly_slice::wide::` | `memset_pattern16` | `vzeroupper` | ymm spills to `(%rsp)` |
| --- | --- | --- | --- |
| `scale` | 2 → **0** | 3 → 1 | 8 → **0** |
| `butterfly_forward` | 2 → **0** | 3 → 1 | 9 → **0** |
| `butterfly_inverse` | 2 → **0** | 3 → 1 | 9 → **0** |

The one remaining `vzeroupper` per kernel is the epilogue, before `retq`. The darwin `scale` loop is
now the same shape as the Linux one you quoted — twiddle hoisted into `ymm0`/`ymm1`/`ymm2`, 2x
unrolled, five `vpclmulqdq` per register:

```asm
LBB4_5:
        vmovdqu (%rcx), %ymm3
        vmovdqu 32(%rcx), %ymm4
        vpclmulqdq $0,  %ymm0, %ymm3, %ymm5
        vpclmulqdq $1,  %ymm2, %ymm3, %ymm6
        vpxor   %ymm5, %ymm6, %ymm5
        vpclmulqdq $16, %ymm0, %ymm3, %ymm6
        vpclmulqdq $17, %ymm2, %ymm3, %ymm3
        ...
```

Same flags emitted for `x86_64-unknown-linux-gnu` as a control: zero memset calls, same hoisting,
`vpclmulqdq` counts identical to darwin (16 / 6 / 6). Both emissions redone after rebasing onto
`77ebf1f6`. I did not emit the AVX-512 variant on this pass.

binary-field/src/poly_slice.rs
tcoratger: You are right, and thanks for checking the algebra by hand when CI could not. Closed structurally
rather than by hoping a runner has the instruction.

New `binary-field/src/packed/split.rs` states the algebra once over a `Lanes` trait — the companion
derivation, the four immediates, `unpack_low_64`, the `HIGH_BY_LOW` fold. The register is one
backend (`impl Lanes for lanes::Reg`, `packed/x86_64.rs`). A two-lane `[u128; 2]` model with a
bit-serial 64x64 carryless multiply is the other, compiled under `cfg(test)` on **every** target.
Two lanes rather than one so a lane-crossing mistake shows up as well. `PackedGhash128::square` now
routes through the same shared fold, so the reduction behind `Mul`, `square` and `dot_product` is
pinned too, not just the split path.

Your typo, on a plain `x86_64-unknown-linux-gnu` build with no target features:

```
cargo nextest run -p p3-binary-field -E 'test(the_split_product_agrees)'
```

| state | result |
| --- | --- |
| unmodified | 2 passed |
| `LOW_BY_HIGH` ↔ `HIGH_BY_LOW` swapped in `apply` | 2 failed; proptest shrank to `scalar = 1, values = [0, 1]` |
| reverted | 2 passed |

And the real path did execute here, since this host has the instruction:

```
cargo nextest run -p p3-binary-field                                    333 passed
RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" cargo nextest run …     351 passed
```

The 18 extra are the packed suite that only exists on that leg. I did not run the
`+avx512f,+vpclmulqdq` or `+pclmulqdq` legs this pass.

I left `ci.yml` alone. Flipping `+avx2,+vpclmulqdq` to test-running behind
`is_x86_feature_detected!` is still worth doing, but the comment above that matrix says renaming a
leg breaks the required checks, so it belongs in its own change.

binary-field/src/poly_slice.rs
tcoratger: Fixed, and I reproduced your finding as an A/B on the emitted assembly rather than taking it on faith.

`tail()` now goes through the broadcast intrinsic — `_mm256_broadcastsi128_si256(_mm_set_epi64x(hi, lo))` at the 256-bit width and `_mm512_broadcast_i32x4(..)` at 512, exactly as you suggested. `packed/x86_64.rs:88` and `:186`.

Counting `_memset_pattern16` in the whole object file, `x86_64-apple-darwin`, `-C target-feature=+avx2,+vpclmulqdq`:

```
array-constant transmute (probe) : 6
broadcast intrinsic (head)       : 0
```

Six is your number exactly. The probe was a one-function revert of `lanes::broadcast` back to `transmute::<[u128; WIDTH], Reg>([value; WIDTH])`, so nothing else moved between the two builds.

The scale kernel's inner loop on darwin now reads the twiddle from registers with no memory operand:

```asm
vpclmulqdq  $1,  %ymm1, %ymm0, %ymm3
vpclmulqdq  $0,  %ymm0, %ymm3, %ymm4
vpclmulqdq  $1,  %ymm2, %ymm3, %ymm5
```

Linux control re-emitted at the same flags: 0 memsets, unchanged. I did not emit the avx512 variant.

binary-field/src/poly_slice.rs
tcoratger: Closed properly rather than by flipping a CI leg.

The split algebra now lives in `packed/split.rs`, written once against a `Lanes` trait — `zero`, `broadcast`, `tail`, `xor`, `unpack_low_64`, `clmul::<IMM>`. The register backend is one impl; the tests carry a second impl that is a pure scalar model, with `clmul` built from a plain bitwise `carryless_64`. So `SplitScalar::new` / `apply` — the companion derivation, the four immediates, the `unpack_low_64(zero, t1)`, the `HIGH_BY_LOW` fold — are exercised on **every** target, including the legs where `Packing` is scalar `Ghash128` and `wide::*` returns 0.

Three tests over that model: a corner-case KAT, a proptest against the scalar backend over random `(scalar, values)`, and one pinning the companion as the shifted multiplier. Your `LOW_BY_HIGH`/`HIGH_BY_LOW` typo now fails on a bare `x86_64-unknown-linux-gnu` build with no target features.

I also ran the real wide path on hardware that has it: `+avx2,+vpclmulqdq`, `p3-binary-field` 333 passed, and `p3-binary-dft` 404 passed on the same flags.

I left `ci.yml` alone deliberately — the comment above the matrix warns that a leg's job name embeds its `features` and that two of those names are required checks, so flipping one to test-running is a branch-protection change, not a workflow change. Worth doing, but as its own PR.

binary-field/src/poly_slice.rs
tcoratger: Fixed. `poly_slice.rs:240` — the fixture now carries trailing sentinels past the payload and asserts they are untouched, so a store that ran a register past the end fails the test instead of corrupting allocator metadata. The leading-pad assertion stays, so both ends are covered.

You were right that the tail loop is the likelier direction: `as_chunks_mut` bounds the front for free, and the remainder walk is the part that can overrun.

binary-field/src/poly_slice.rs
tcoratger: Added. `debug_assert_eq!(lo.len(), hi.len(), "butterfly lengths differ")` at the top of both `butterfly_forward` (`poly_slice.rs:30`) and `butterfly_inverse` (`:51`), each under an `// Invariant:` line saying a mismatch drops work rather than erring loudly.

`scale` takes a single slice so it has no pairing to assert. The kernels still report the count they covered and the caller drives the remainder from it, so the truncation is now caught in debug rather than being silent at any length.

binary-field/src/poly_slice.rs
tcoratger: Fixed. The inlined dispatchers now short-circuit before the call boundary — `poly_slice.rs:15`, `:33`, `:54`:

```rust
let packed = if lo.len() < wide::WIDTH { 0 } else { wide::butterfly_forward(lo, hi, scalar) };
```

So `binary-dft`'s `local_stage` driving down to `half == 2` no longer pays a call/ret per butterfly on an AVX-512 build. Same guard on `scale` and `butterfly_inverse`.

I kept `#[inline(never)]` on the kernels themselves, and the doc on each now says why: out of line so the dispatcher stays small enough to inline into its callers, which is what makes the short path as cheap as a bare product.

Not measured — benchmarking was out of scope for this pass, so this rests on removing the call boundary, not on a number.

binary-field/src/poly_basis.rs
tcoratger: Incidental, and restored — ``Undo [`butterfly_forward`] with the same scalar.`` is back at `poly_slice.rs:47`. `cargo doc -p p3-binary-field --no-deps` resolves it with no warning.

binary-field/src/poly_slice.rs
tcoratger: Straight answer: I did not settle this, and I am not resolving the thread.

Your clmul accounting is right — the split form is 5 per register, and `PackedGhash128::mul` is also 5 once the scalar is broadcast and `mixed_y` is hoisted as loop-invariant. So the 4.4x/4.6x in the description, which is against the 24-clmul scalar loop, does not answer the question you are asking. The A/B against a ~40-line `broadcast + Mul` kernel over the same slices is the number that would, and I could not run benchmarks on this pass.

What the reorganisation did change independent of that question: the split algebra moved out of `poly_slice.rs` into `packed/split.rs` behind a `Lanes` trait, which is what let me pin it with a scalar model on every target (your other comment). `poly_slice.rs` is down to the three dispatchers plus the wide prefix — 88 lines removed from it. So the widened crate-internal surface is narrower than the 433 lines you were looking at, but it is still a larger surface than a broadcast-and-multiply loop.

My read is that the split should have to earn that surface with a number, and it does not have one yet. Two ways to go and I am happy with either:

1. I run the three-arm bench (scalar / split / `broadcast + Mul`) on Zen 5 and we decide from the table. If the simple kernel matches, I replace the split with it.
2. This lands as-is on the correctness fixes and I open the A/B as a follow-up.

Leaving this open for you to pick.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,395 @@
+//! Whole-slice kernels for the polynomial basis: one scaling and two butterflies.
```
Nashtare, line 1: **Nothing in CI ever executes this module.**

`.github/workflows/ci.yml` has four vpclmulqdq legs (`+avx2,+vpclmulqdq` and `+avx512f,+vpclmulqdq`, x parallel), and every one of them is `compile_only: true` — the comment there says so explicitly ("GitHub's runner pool still includes older parts, so these legs compile the wide carryless-multiply paths rather than running them"). `ci-heavy.yml` runs `+avx2` and `+pclmulqdq`, neither of which enables `wide`. So on every leg that actually runs tests, `wide::WIDTH == usize::MAX`, `packed == 0`, and the new proptests in this file exercise the *scalar tail loop they already had*, element for element.

The `Model` in `packed/split.rs` does run everywhere, and it is a good idea — but it validates the *algebra* (`SplitScalar::apply` against `poly_mul_128`) over a backend the tests wrote themselves. The part that can actually be wrong on hardware is the other seam: `Lanes for lanes::Reg`, i.e. that `unpack_low_64` is `_mm{256,512}_unpacklo_epi64` per 128-bit lane, that `clmul::<0x01>` picks the high qword of the *first* operand and the low of the second, that `_mm512_broadcast_i32x4` lands the same 128 bits in all four lanes. The `Model` cannot check any of that, and CI never runs the code that would.

So the shipped correctness evidence for both the 2-lane and 4-lane kernels is one developer's `-C target-cpu=native` run on a Zen 5 box. I checked the algebra by hand and it is right (`u = t*x^64 mod p` via one `fold_shifted` with nothing underneath; `low = t0 v0 + u0 v1`, `high = t1 v0 + u1 v1`, both `<= deg 126`, so `low + high*x^64` is three limbs and one Horner step closes it exactly since `h1*T <= deg 70`). But that is review, not CI.

Two things would close most of the gap cheaply:

1. A direct `Reg`-vs-`Model` conformance test, one `Lanes` method at a time, gated on the same cfg as `wide`. It localises a mismatch to the intrinsic instead of to an end-to-end slice comparison, and it is the test that will still be there when someone swaps `_mm256_unpacklo_epi64` for something cheaper.
2. Flip one of the four vpclmulqdq legs to actually run `-p p3-binary-field`. Azure's `ubuntu-24.04` pool has been Ice Lake / Zen 3+ for a while; if that is still too risky, `qemu-x86_64 -cpu max` under `cross` runs these instructions and is cheap for a single crate.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,395 @@
...
+            //
+            // Both accesses are the unaligned forms, so the alignment of the slice is free.
+            //
+            // The target features the instructions need gate this module.
+            unsafe {
+                let v = lanes::load(block.as_ptr());
+                lanes::store(block.as_mut_ptr(), split.apply(v));
+            }
+        }
+        blocks.len() * WIDTH
```
Nashtare, line 101: The kernels report how many elements they covered and every caller throws it on the floor:

```rust
let packed = if values.len() < wide::WIDTH { 0 } else { wide::scale(values, scalar) };
```

`packed` is only ever used as a tail offset, so a build in which `wide` silently resolves to the fallback module — a cfg typo, a feature-detection change, a future `WIDTH` that stops dividing the slice — produces byte-identical results and a green test suite, with the entire optimisation gone. Given that CI never runs the packed path anyway (see the module-level comment), that failure mode is a rename away.

An assertion in the tests that a length `>= WIDTH` reports `packed == len - len % WIDTH` costs one line and turns "the fast path exists" into something the suite checks. `wide::WIDTH` is already `pub(super)`, so the test can read it.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,395 @@
...
+    ///     all ones          every coefficient of every half product is live
+    ///     x^127             the highest degree, so the companion spills furthest
+    ///     x^64              the companion is the multiplier's own shift, already reduced
+    ///     lower half only   the companion is the only source of upper limbs
+    ///     0x87              the modulus tail itself
+    /// ```
+    const CORNERS: [u128; 7] = [0, 1, u128::MAX, 1 << 127, 1 << 64, (1u128 << 64) - 1, 0x87];
+
+    /// Guard elements placed after the payload, one full register of the widest build.
+    const SENTINELS: usize = 4;
```
Nashtare, line 222: The guard walls are asymmetric, and the one that matters is missing.

- Trailing: `SENTINELS = 4` is exactly one register on the AVX-512 build. A store one register past the end is caught; two registers past runs off the end of a `Vec::with_capacity(offset + len + 4)` allocated to the exact size, i.e. heap corruption that the assertion never sees.
- Leading: the wall is `offset` zeros with `offset in 0..4`. On the 4-lane build a full-register *under*-run from block 0 writes at `offset - 4 < 0`, i.e. before the `Vec`'s buffer, on every offset the sweep uses. There is no case in which the leading assertion fires for a real under-run.

Making both walls `SENTINELS` wide (front pad = `offset + SENTINELS`, and check the pad as sentinels rather than zeros) makes both directions detectable and costs nothing. As written, the docstring's claim that the padding "catches a store that ran a whole register past the end" holds in one direction only, and only for exactly one register.

binary-field/src/poly_slice.rs
```diff
@@ -0,0 +1,395 @@
...
+    for value in &mut values[packed..] {
+        *value = clmul::poly_mul_128_batch(*value, scalar);
+    }
+}
+
+/// Send `(lo, hi)` to `(lo + scalar*hi, lo + (scalar + 1)*hi)`, element by element.
+#[inline]
+pub(crate) fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
+    // Invariant: the two halves are paired element for element, so a mismatch drops work.
+    debug_assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
```
Nashtare, line 31: `debug_assert_eq!` here, but `assert_eq!` in the `poly_basis` wrappers that call in. In a release build a mismatched pair does not panic — `wide::butterfly_forward` takes `blocks = lo.len().min(hi.len())` and the tail `zip` stops at the shorter side, so the extra elements are silently left unbutterflied. The comment two lines up ("a mismatch drops work") names precisely the outcome a `debug_assert` cannot prevent.

Since these are `pub(crate)` and the public wrappers already pay for a real `assert_eq!`, either promote this to `assert_eq!` (the wrappers' check makes it free on the current path anyway, and a future caller gets the guarantee) or drop it and document the truncation as intended.

binary-field/src/packed/split.rs
```diff
@@ -0,0 +1,283 @@
...
+pub(crate) trait Lanes: Copy {
+    /// All lanes zero.
+    fn zero() -> Self;
+
+    /// The same field element in every lane.
+    fn broadcast(value: u128) -> Self;
+
+    /// The modulus tail in the low quadword of every lane.
+    #[inline(always)]
+    fn tail() -> Self {
```
Nashtare, line 36: `tail()` is now `Self::broadcast(TAIL_128)`, i.e. `_mm_set_epi64x(0, 0x87)` followed by a lane broadcast, where the code it replaces was `transmute::<[u128; WIDTH], Reg>([TAIL_128; WIDTH])` — a constant that LLVM materialises as a single load from `.rodata`.

`fold_shifted` is called twice per `PackedGhash128::mul`, twice per `dot`, twice per `square`, and once per `SplitScalar::apply`, so this sits in the innermost loop of every packed operation in the crate. In practice LLVM should constant-fold `_mm_set_epi64x` of two literals and CSE the broadcast out of the loop, and the `mul/*` benchmark rows suggest it does on Zen 5 — but that is one compiler on one target, and the `Mul` path is not what this PR set out to change. Worth a glance at the disassembly on the AVX2 leg (where `_mm256_broadcastsi128_si256` is a separate instruction from the load) to confirm nothing regressed in the non-benchmarked shapes.

binary-field/src/poly_basis.rs
```diff
@@ -48,13 +48,12 @@ pub fn square(a: u128) -> u128 {
...
 pub fn mul_slice(values: &mut [u128], scalar: u128) {
+    // A lone element has no loop to amortize any setup over, so it takes the plain product.
     if let [value] = values {
         *value = mul(*value, scalar);
         return;
     }
-    for value in values {
-        *value = clmul::poly_mul_128_batch(*value, scalar);
-    }
+    poly_slice::scale(values, scalar);
```
Nashtare, line 56: The measurement stops one layer short of the consumer.

The benchmark tables cover `p3-binary-field`'s `batch/poly128` micro-benchmarks and `p3-binary-dft`'s `ntt/128/hybrid`. `mul_slice` and the two butterflies are `pub` and their real caller is `p3-binary-pcs` through `AdditiveRsEncoder`, which is where the crate's actual cost lives and which #2094 has just been tuned against on `main`.

The 32-thread NTT rows already show the win largely dissolving at the ends (`2^14`: 1.01x, `2^20`: 1.03x) because the transform goes memory-bound, and the PR says so plainly, which I appreciate. But that is exactly the regime binary-PCS runs in. A single `p3-binary-pcs` end-to-end number at the shape `ci-heavy.yml`'s `binary_pcs_large_configuration` job uses (`2^16` round trip) would tell a reader whether a 4.5x kernel is worth ~1200 lines of new `unsafe`-adjacent code and a second copy of the reduction algebra, or whether it disappears into the DFT's memory traffic.

Separately, note that `binary_pcs_large_configuration` runs with `+pclmulqdq` only, so it will not exercise this code either.

Nashtare (approved):

Nashtare:
@tcoratger maybe tell your agent to have a look at reviews before merging to `main` 😬

tcoratger:
> @tcoratger maybe tell your agent to have a look at reviews before merging to `main` 😬

Ooooh that's not good! Aaaah, I just realized: auto-merge was enabled before the last review, so as soon as GitHub saw the approval, it merged... I found it strange, since I wasn't even logged in 15 minutes ago when the merge happened.

Let me submit a follow-up PR right now.

Nashtare:
Yeah I'm not a big fan of auto-merge ahead of time for this kind of scenario, but all good.

binary-field/src/poly_slice.rs
tcoratger: Right, and auto-merge took the branch before I read this. Sorry. Fixed in #2101.

Your option 1: a direct register-vs-model conformance test, one lane operation at a time (`packed/x86_64.rs:533-624`). The scalar model moves out of `split.rs`'s test module into `split::model` and becomes generic over the lane count, so the same reference serves both the two-lane and four-lane builds. The bridge between register and model is the unaligned load and store, so those are under test too.

One assertion per operation, so a failure names the intrinsic rather than the slice: `zero`, `broadcast`, `tail`, `xor`, `unpack_low_64`, the four `clmul` immediates, then `fold_shifted` and `SplitScalar::apply` as composites.

Mutation-checked, one intrinsic broken at a time:

| mutation | leg | new tests fail |
| --- | --- | --- |
| `_mm256_unpacklo_epi64` -> `_mm256_unpackhi_epi64` | `+avx2,+vpclmulqdq` | yes, both |
| `_mm512_broadcast_i32x4` -> `_mm512_castsi128_si512` (lane 0 only) | `+avx512f,+vpclmulqdq` | yes, both |

`ci.yml` untouched again, same reason as last round: the matrix comment says a leg's job name embeds its `features` and two of those are required checks, so flipping one is a branch-protection change. Your `qemu-x86_64 -cpu max` idea is the cheaper half of it and I have not tried it yet.

binary-field/src/poly_slice.rs
tcoratger: Fixed in #2101, `poly_slice.rs:396-441`.

A test pins all three kernels to report `len - len % WIDTH` at every length from an empty prefix to four whole registers. The target predicate is restated in the test as a `cfg!` rather than read from the kernel's own gate, which is what makes a drifted gate a failure instead of a silent fallback.

Checked against exactly the failure you described. Mistyping `vpclmulqdq` in `mod wide`'s gate, on an `+avx2,+vpclmulqdq` build:

```
failures:
    poly_slice::tests::the_packed_prefix_covers_every_whole_register
test result: FAILED. 173 passed; 1 failed
```

One test fails and it is the new one; every correctness test stays green. Which was the point.

binary-field/src/poly_slice.rs
tcoratger: Both directions were wrong and you called it. Fixed in #2101, `poly_slice.rs:225-289`.

Both walls are now `SENTINELS = 8`, two registers of the widest build, and the leading pad is `SENTINELS + offset` sentinels rather than `offset` zeros — so the payload still walks every alignment while the wall in front of it is real.

A/B on injected faults, `+avx512f,+vpclmulqdq`, `WIDTH = 4`:

| injected fault | shipped fixture | new fixture |
| --- | --- | --- |
| extra store at `block - WIDTH` | `SIGABRT` (`free(): invalid size`) | `assertion failed: buffer[..start] ... == SENTINEL` |
| extra store at `block + 2*WIDTH` | `SIGABRT` (`free(): invalid size`) | `assertion failed: buffer[start + len..] ... == SENTINEL` |

Heap corruption in both directions before, clean assertions now.

binary-field/src/poly_slice.rs
tcoratger: Promoted, in #2101 — `assert_eq!` at `poly_slice.rs:30` and `:52`.

The public wrappers assert first so it is free on the current path, and a future crate-internal caller gets the guarantee instead of a truncated pass. Taking your first option rather than documenting the truncation, since silently dropping half a butterfly is not a behaviour worth keeping reachable.

binary-field/src/packed/split.rs
tcoratger: Checked, no regression. `+avx2,+vpclmulqdq`, `x86_64-unknown-linux-gnu`:

```
cargo rustc --release -p p3-binary-field --lib -- --emit asm
```

LLVM constant-folds `_mm_set_epi64x(0, 0x87)` and the lane broadcast into one 8-byte `.rodata` entry and hoists the load into the prologue:

```asm
	vpbroadcastq	.LCPI4_0(%rip), %ymm1     ; prologue, once
	...
.LBB4_5:                                          ; 2x unrolled loop
	vmovdqu	(%rcx), %ymm3
	vpclmulqdq	$0,  %ymm0, %ymm3, %ymm5
	vpclmulqdq	$1,  %ymm2, %ymm3, %ymm6
	vpxor	%ymm5, %ymm6, %ymm5
	vpclmulqdq	$16, %ymm0, %ymm3, %ymm6
	vpclmulqdq	$17, %ymm2, %ymm3, %ymm3
	...
	vpclmulqdq	$1,  %ymm1, %ymm3, %ymm3   ; tail from ymm1, no memory operand
```

```text
.LCPI4_0:
	.quad	135
```

Counted inside each kernel's symbol body: zero calls, zero SIMD spills to the stack, one `.rodata` load of the tail, all outside the loop. Same codegen the array-constant transmute gave on Linux, so there is nothing left to regress — the broadcast form only removed the darwin pathology. `_mm256_broadcastsi128_si256` never survives as a separate instruction here; it folds into the `vpbroadcastq`.

binary-field/src/poly_basis.rs
tcoratger: Fair, and the number is not flattering. `p3-binary-pcs`, `2^16`, `-C target-cpu=native`, two pre-built binaries from `f4e958d3^` and `f4e958d3`, three interleaved rounds, minimum of three:

| | before | after | ratio |
| --- | --- | --- | --- |
| `commit/16` | 20.31 ms | 18.09 ms | 1.12x |
| `open/16` | 28.09 ms | 27.98 ms | 1.00x |

A 4.5x kernel is 1.12x on commit and nothing measurable on open. Your read was right: it largely dissolves into the DFT's memory traffic at the shape the consumer runs. The host was shared with other builds during these rounds, so 1.12x is an upper bound on the noise floor rather than a clean number.

And yes, `binary_pcs_large_configuration` runs `+pclmulqdq` only, so CI does not exercise this code there either. Recorded in #2101.

binary-field/src/poly_slice.rs
tcoratger: Ran it, since I said I would. You were right.

Slice scaling, `-C target-cpu=native`, four lanes, Zen 5, minimum of three rounds:

| len | scalar loop | split kernel | `broadcast + Mul` | split vs broadcast |
| 16 | 39.35 ns | 19.37 ns | 18.50 ns | 0.96x |
| 256 | 563.0 ns | 180.8 ns | 183.3 ns | 1.01x |
| 4096 | 9.239 us | 2.782 us | 2.823 us | 1.01x |
| 65536 | 141.0 us | 44.08 us | 44.64 us | 1.01x |

Your clmul accounting was exact. Both forms are 5 per register once the scalar is broadcast and `mixed_y` hoisted, and the shuffle-and-xor difference is free behind `VPCLMULQDQ` throughput on Zen 5. The split does not beat a ~40-line broadcast-and-multiply loop at any length, and both are ~3.2x the scalar loop.

So the split has not earned its surface. The host was loaded during these rounds, so I want a re-run on a quiet box before acting; if it holds, the right change is to delete the split and keep the dispatch, which is a strict simplification. I will open that as its own PR rather than fold a rewrite into a review-fix — this thread stays open until that lands. Happy to do it the other way if you prefer.

PR #2055, opened 2026-09-07
perf(sumcheck)!: fuse bind and measure across single-round calls and under suffix order

tcoratger:
## Summary

A quadratic sumcheck round does two things to its tables: it *measures* them to produce the round polynomial `h_i`, and it *binds* one variable to the challenge `r_i` the verifier answers with. Both are full passes over the evaluation and weight tables, so a round that does them separately reads `2^n` entries twice.

Prefix-ordered rounds already fused the two, but only *inside* one `compute_sumcheck_polynomials` call: the challenge was kept in a driver-local `pending: Option<EF>`, so a caller that asked for one round at a time got a fresh `None` on every call and paid two passes per round. That single-round shape is exactly what `p3-binary-pcs` asks for: its inner loop requests one round at a time. Its outer loop is what makes the boundary unavoidable — one codeword fold and one Merkle commitment land per fold *batch*, so a batch boundary is a call boundary no batching can remove.

This PR does two things:

- Moves the held challenge out of the driver's local and into `SumcheckProver` as `outstanding: Option<EF>`, so the fusion survives across call boundaries.
- Adds a fused suffix kernel (`fold_and_round_coefficients_suffix`), so the suffix binding order gets the same one-pass round the prefix order already had.

Net effect: one pass per round instead of two, for both binding orders, and regardless of how many rounds a caller asks for at a time.

## Why the challenge is deferred rather than applied

The obvious alternative is to bind `r_i` the moment it is sampled and keep the tables always current. That cannot fuse, because the bind for `r_i` and the measure for `h_{i+1}` are then separated by the transcript work in between, and each has to find the tables in memory on its own.

Deferring inverts it. `hold(r_i)` records the challenge and touches nothing; the *next* round's measuring pass reads each block of the table once, binds it, and measures the result while it is still in cache. The prover state then has two components that can disagree by one round:

```text
    claim  : always current, because advancing it needs only h_i and r_i
    tables : lag by one binding while a challenge is outstanding
```

The whole design is one rule: **every reader of the tables settles first.** `settle()` applies an outstanding binding and is idempotent and free when there is none, and each of `evals`, `evals_view`, `eval`, `weights`, `scale_weights_and_claim`, `accumulate_claim` and the constraint-absorption path calls it on the way in. `num_variables()` is the single exception, and only because it is a length: subtracting the outstanding binding answers it exactly, with no pass over the data.

Keeping the slot in the prover rather than in a driver is what makes the single-round caller work. A driver-local slot is scoped to one call; prover state is scoped to the sumcheck.

## Ordering, and why the transcript stays sound

Deferring a challenge changes *when* the tables are bound, never *what* a round's message is a function of. Per round, `compute_sumcheck_polynomials` (`sumcheck/src/strategy.rs:1434-1452`) runs strictly:

```text
    measure_round()                 -> h_i, absorbing the binding for r_{i-1}
    observe_and_sample(...)         -> commits h_i to the transcript, then samples r_i
    reduce_claim (round identity)
    hold(r_i)                       -> records r_i, applies nothing
```

`hold` is the only writer of `outstanding`, and it runs after the transcript has already absorbed `h_i`. So:

- Round `i`'s message is a function of `r_0..r_{i-1}` only.
- The prover never sees `r_i` before it has committed to `h_i`.

That is the Fiat-Shamir property the protocol needs, and it is the same before and after this change. The ZK overlay has the identical ordering (`sumcheck/src/zk/prover/residual.rs:132/148/160/161`).

The fusion itself is exact rather than approximate: `bind_pairs` computes `out[g] = a[2g] + (a[2g+1] - a[2g]) * r`, which is what `Poly::fix_suffix_var_mut` computes, and the measuring half reuses the same `chunk_round_step` / `round_step` helpers as the unfused kernel, so the 0/infinity coefficient convention is untouched. `prop_fold_and_round_coefficients_suffix_matches_bind_then_measure` pins both the bound tables and the message against the two-pass reference.

## One seam per route choice

The reference route exists only so the fused one can be pinned against it, so there is exactly one place that chooses between them:

```text
    fold_rounds_with<BIND_EACH_ROUND>       the seam itself
      ^-- BinaryPcs::finish_open_with<BIND_EACH_ROUND>
            ^-- BinaryPcs::finish_open = finish_open_with::<false>
                  ^-- try_open, try_open_at
```

Carrying the parameter up to `finish_open_with` is what lets `fusing_the_binding_into_the_measuring_pass_leaves_the_proof_byte_identical` compare against the shipped opening path itself. A test-local copy of `finish_open` would drift against it silently — the two would still serialize identically while no longer sharing a step.

`fold_rounds` and `fold_rounds_binding_each_round` are gone; both were `fold_rounds_with` with the parameter fixed, and the `binary pcs fold rounds` tracing span now sits on the seam.

## Public API break

`SumcheckProver` is `pub`, and four of its accessors change receiver:

| method | before | after |
| --- | --- | --- |
| `evals` | `&self` | `&mut self` |
| `evals_view` | `&self` | `&mut self` |
| `eval` | `&self` | `&mut self` |
| `weights` | `&self` | `&mut self` |

Reading the prover is now a mutating operation, because the read has to settle an outstanding binding first. A downstream crate holding a `&SumcheckProver` and calling any of these will no longer compile. In-workspace every call site is already a `&mut` binding. The break rides the `perf(sumcheck)!:` squash subject, which is what `cliff.toml` turns into the changelog line at release time.

## Memory

The suffix fused pass cannot bind in place. Output index `g` reads inputs `2g` and `2g+1`, so a serial forward sweep would be safe, but blocked parallelism is not: cut the output at `0 < G_1 < ...` and block 1 writes from `G_1` while block 0 still reads up to `2 G_1`, and those ranges always overlap. The pass therefore writes a half-size destination.

That destination is allocated per round and handed straight back:

```rust
let (bound_evals, bound_weights, message) =
    bind_and_measure_pairs(evals.as_slice(), weights.as_slice(), r);

*evals = Poly::new(bound_evals);
*weights = Poly::new(bound_weights);
```

Installing the bound halves drops the sources there, so between rounds the pair holds one half-length table per side and nothing else.

An earlier revision of this PR carried the destination from round to round in a `FoldBuffers` pair, so the tables and the buffers traded places. That is gone, because the reuse never happened. `swap_storage` hands a round the table from the round *before* it, which is four times the length the next destination needs, while `resize_destination` kept a buffer only while its capacity was within twice that length — a test that is false by construction. Instrumenting a `2^15` suffix ladder:

```text
    round  0: live= 16384 buf.len= 32768 buf.cap= 32768 ptr_same_as_prev=false
    round  1: live=  8192 buf.len= 16384 buf.cap= 16384 ptr_same_as_prev=false
    ...
    round 13: live=     2 buf.len=     4 buf.cap=     4 ptr_same_as_prev=false
```

Every round reallocated, and `capacity == 2 * live` on every round (which also made the old test's `capacity <= 2 * live` assertion vacuous). So the mechanism was a plain out-of-place fold that additionally parked the old table in the buffer until the next round dropped it:

| | resident between rounds, per side |
| --- | --- |
| with the buffer exchange | `half` (table) + `2 * half` (old storage in the buffer) = `3 * half` |
| now | `half` |

Removed with it: the `pub struct FoldBuffers` and its `Default`/`new`, the `buffers` field on `ProductPolynomial`, and the `buffers` argument on `fold_and_round_coefficients_suffix`. Allocation count is unchanged, since the reuse branch never fired: one `zero_vec(half)` per table per round, before and after. `zero_vec` takes pages the allocator has already zeroed rather than filling them from userspace, and every entry is overwritten before anything reads it.

## Performance

`bench_single_round_drive` drives a whole ladder one round per call — the shape `binary-pcs` asks for today. Two arms over the identical ladder: `fused` (the challenge is held and the next round's pass absorbs it) and `bind_each_round` (`settle()` after each round, so the next round measures in a pass of its own).

`RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-sumcheck --bench sumcheck -- babybear/single_round_drive`, Zen 5, BabyBear degree-4 extension, criterion mean:

| k | order | fused | bind_each_round | ratio |
| 8 | prefix | 5.26 us | 5.23 us | 1.00x |
| 8 | suffix | 6.85 us | 6.67 us | 0.97x |
| 12 | prefix | 18.89 us | 20.27 us | 1.07x |
| 12 | suffix | 49.90 us | 55.99 us | 1.12x |
| 16 | prefix | 206.72 us | 237.88 us | 1.15x |
| 16 | suffix | 757.52 us | 977.42 us | **1.29x** |
| 20 | prefix | 4.524 ms | 5.846 ms | **1.29x** |
| 20 | suffix | 16.770 ms | 20.065 ms | **1.20x** |

How many rounds this actually saves a pass on is a function of the fold schedule, not of the round count:

```text
    rounds fused across a call, if a batch were asked for in one go
        = num_fold_batches - 1
        = ceil(num_variables / log_folding_factor) - 1

    rounds fused across a call, as the inner loop asks today (arity 1)
        = num_variables - 1
```

`BinaryPcsConfig::try_new` sets `log_folding_factor = 1` (`params.rs:108-113`), so today every round crosses a call boundary and the grid above measures the shipped schedule exactly. Above an arity of 1 the grid is an upper bound: the rounds inside a batch could be asked for together, and only the batch boundaries would remain. The fusion is still required at those boundaries at any arity.

WHIR is a second beneficiary, at the end of a proof. Because the driver no longer binds after its loop, `final_round`'s trailing `compute_sumcheck_polynomials(final_sumcheck_rounds, ..)` leaves its last challenge outstanding, and nothing reads the tables afterwards — `round_state` is dropped at `whir/src/pcs/prover/mod.rs:194` untouched. That whole fold pass disappears whenever `final_sumcheck_rounds > 0`. Every other WHIR round settles at the next round's `evals_view()` (`:235`), so the pass count there is unchanged rather than worse.

At `k = 8` the suffix ladder is 3% *slower* fused. The whole ladder there runs below `PARALLEL_THRESHOLD`, where `fix_suffix_var_mut` folds fully in place and allocates nothing, so the fused route's per-round half-size destination is added cost with no read pass saved worth the trade. The crossover sits between `k = 8` and `k = 12`.

The `k = 20` rows fold all the way down to two entries, so they already contain that regime at their tail and are still 1.20x–1.29x: the cost is geometric in the large rounds, so the bottom of the ladder cannot undo the win. No size gate is added for the 3%.

This host was running other builds throughout. Both arms sit in the same criterion invocation back to back, so drift between them is small, but sub-5% rows should be read as noise.

The per-round claim check (`debug_assert_eq!(sum, dot_product())` in `measure_round`) was measured separately, as an interleaved A/B on `cargo nextest run -p p3-sumcheck -p p3-binary-pcs` in debug, with the build outside the timed region:

| iteration | with the check | without | delta |
| --- | --- | --- | --- |
| 1 | 65.66 s | 61.38 s | +4.28 s (+7.0%) |
| 2 | 53.91 s | 51.95 s | +1.96 s (+3.8%) |
| 3 | 95.15 s | 46.90 s | discarded, load spike |

4–7% on a ~55 s suite. Kept, because in the fused flow that assertion is the primary validation of the new binding path: round `i`'s `measure_round` is what applies `r_{i-1}`.

## Test plan

- [x] `cargo nextest run -p p3-sumcheck -p p3-binary-pcs` — 425 passed, 1 skipped
- [x] `cargo nextest run -p p3-sumcheck -p p3-binary-pcs --features parallel` — 425 passed, 1 skipped
- [x] `cargo clippy -p p3-sumcheck -p p3-binary-pcs --all-targets -- -D warnings` — clean
- [x] `cargo test --doc -p p3-sumcheck` — ok
- [x] `RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-sumcheck --bench sumcheck -- babybear/single_round_drive` — table above

Coverage added by this PR:

- `prop_fold_and_round_coefficients_suffix_matches_bind_then_measure` — the fused suffix pass equals bind-then-measure, on both the bound tables and the message, over `2^2..2^16` entries straddling the tiled body and the par/serial split.
- `every_suffix_round_binds_correctly_and_keeps_only_the_live_table` — a whole `2^15` sumcheck folded down to four entries, matching the two-pass reference every round, with each round's live length pinned to `2^{n-1-i}` per side.
- `the_fused_suffix_pass_allocates_exactly_the_bound_half` — the destinations the pass returns satisfy `len == capacity == half`, so no round over-allocates.
- `every_table_accessor_settles_an_outstanding_binding` — a batch is run so a challenge is left outstanding, then every accessor is read on a fresh copy and compared against the explicitly settled state, for three arities and both binding orders.
- `holding_a_second_challenge_without_settling_is_rejected` — the second `hold` panics, and the unwind is caught so the test can assert the *first* challenge is still the one held. Mutation-checked: with `Option::replace` the prover ends up holding the second and the assertion fails.
- `a_panicking_fused_pass_leaves_the_challenge_held` — `measure_round` panics on a pair with no variable left to bind, the unwind is caught, and the challenge is asserted still held. `measure_round` reads `outstanding` and clears it only once the pass has returned, the same standard `hold` holds itself. Mutation-checked: with `Option::take` the prover is left `None` and the assertion fails.
- `absorbing_a_constraint_settles_an_outstanding_binding` — a batch is run so a challenge is left held, then `compute_sumcheck_polynomials(.., Some(constraint))` is called with nothing read in between, and compared against a copy settled explicitly first, for both binding orders. This is the one `settle()` whose ordering is load-bearing: settling after `combine` would add a weight delta of the bound length to a table still a round long. Mutation-checked: swapping the two lines panics on the length.
- `fold_rounds_agree_with_binding_each_round` (binary-pcs) — the fused fold loop and a reference loop that settles every round agree on round messages, commitments, folding randomness, final codeword and the transcript state left behind, at 8 and 15 variables and at `log_folding_factor` 1 and 3. Arity 3 is what reaches the reference route's per-round `settle()` inside a batch; arity 1 only ever crosses batch boundaries.
- `fusing_the_binding_into_the_measuring_pass_leaves_the_proof_byte_identical` (binary-pcs) — the shipped `open` and `finish_open_with::<true>` serialize to identical proof bytes and leave identically seeded challengers in the same state, at `log_folding_factor` 1 and 3.

Not run in this pass: non-default `target-feature` legs, and a peak-RSS measurement of the memory change. The memory numbers above are arithmetic on the lengths plus the instrumented ladder, not a measured RSS figure.


sumcheck/src/strategy.rs
```diff
@@ -859,22 +1190,54 @@ impl VariableOrder {
...
 #[derive(Debug, Clone)]
 pub struct SumcheckProver<F: Field, EF: ExtensionField<F>> {
     /// Paired evaluation and weight polynomials for the quadratic sumcheck.
     poly: ProductPolynomial<F, EF>,
     /// Current claimed sum over the remaining unbound variables.
     sum: EF,
+    /// Challenge sampled by the last round and not yet applied to the tables.
+    ///
+    /// Empty means the tables are current with the claim.
+    outstanding: Option<EF>,
```
Nashtare, line 1178: **The PR description on this PR belongs to a different change.**

The title and the diff are sumcheck + binary-pcs, but the body describes a bit-reversal fusion in `p3-util`: "Bit reversal factors into two commuting permutations… transpose each square of tiles…", the tile-side/`log_mid` discussion, and a test plan whose every checkbox is `cargo test -p p3-util` / `cargo bench -p p3-util --bench bit_reverse`. The word "sumcheck" does not appear in the body once, and no crate this PR touches appears in the test plan.

Consequences for review: there is no stated rationale for the deferred-challenge design, and no stated test evidence for the actual change. Please swap in the right description before this lands.

---

On the change itself: I traced the transcript ordering and **found no soundness issue**. Per round the order is strictly `measure_round()` (`:1425`) → `observe_and_sample` (`:1427`) → `reduce_claim_with_coefficients` → `hold(r)` (`:1440`), and `hold` is the only writer of `outstanding`, so round *i*'s message is a function of `r_0..r_{i-1}` only and the prover cannot see `r_i` before committing to `h_i`. The ZK path has the same ordering (`zk/prover/residual.rs:132/149/155/158/161`), and `residual.rs` did not lose a bind — the removed `bind_pending` is covered by `debug_assert_claim()` → `settle()` at `:170` and again by `scale_weights_and_claim` at `:172`. The suffix fusion is exact: `bind_pairs` computes `out[g] = a[2g] + (a[2g+1] - a[2g])·r`, matching `Poly::fix_suffix_var_mut`, `block_len` is always a multiple of `2K` so no pair straddles a block, and the 0/∞ convention is preserved by reusing the same `chunk_round_step`/`round_step` helpers.

sumcheck/src/strategy.rs
```diff
@@ -548,6 +565,328 @@ where
...
+///
+/// The fill is spread across threads exactly when the round that follows is.
+/// One decision covers both, so there is no second threshold to keep in step.
+#[inline]
+fn resize_destination<A>(buffer: &mut Vec<A>, len: usize, threaded: bool)
+where
+    A: PrimeCharacteristicRing + Copy + Send + Sync,
+{
+    if buffer.len() >= len {
+        buffer.truncate(len);
```
Nashtare: This doc says "a length update and no allocation at all", but it should also say that the **capacity is never released** — the suffix path now retains ~3x the steady-state table memory for the life of the `ProductPolynomial`.

Round 1 allocates an `N/2` destination; `swap_storage` (`:777`) then hands the *old N-entry `Vec`* back as the buffer. From then on this function only ever takes the `buffer.len() >= len` branch and calls `truncate` — never `shrink_to_fit` — so the `N`-capacity allocation is held until drop. Steady state per side is `N/2` (live table) + `N` (buffer capacity) = `1.5N`, against `0.5N` on the old path, where `fix_suffix_var_mut` (`multilinear-util/src/poly/mod.rs:592-596`) allocates a half-size `Vec` and drops the full-size one. Across evals + weights that is `3N` retained versus `N`.

Why it bites in practice: in `binary-pcs/src/prover.rs:225-232` the prover is holding this `3N` while `fold_codeword` allocates a fresh codeword and `mmcs.commit_matrix` builds a Merkle tree, so process peak RSS rises by up to `2N` across the whole fold phase — at 2^28 `BinaryField128` entries, several GiB that the pre-PR code would have freed.

Reusing the buffer instead of reallocating each round is a reasonable trade; it just needs to be a stated one. Either document it here, or reclaim when the buffer overshoots the next round's need by more than a factor.

sumcheck/src/strategy.rs
```diff
@@ -883,54 +1246,89 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
+            // Invariant: the claim is the inner product of the now-current pair.
+            debug_assert_eq!(self.sum, self.poly.dot_product());
+        }
     }
 
     /// Extracts the current evaluation polynomial as scalar extension-field elements.
     #[tracing::instrument(skip_all)]
-    pub fn evals(&self) -> Poly<EF> {
+    pub fn evals(&mut self) -> Poly<EF> {
+        self.settle();
```
Nashtare, line 1235: Unannounced public API break: `evals`, `evals_view` (`:1283`), `eval` (`:1289`) and `weights` (`:1364`) all move from `&self` to `&mut self`, so reading the prover is now a mutating operation.

`SumcheckProver` is `pub` (`:1220`), so any downstream crate holding a `&SumcheckProver` fails to compile. In-workspace it is clean, and the change is the right call given `settle()` has to run — but it deserves a line in the PR description and a changelog entry rather than arriving silently.

sumcheck/src/strategy.rs
```diff
@@ -941,11 +1339,12 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
-    /// Asserts that the claim is the inner product of the bound pair.
+    /// Asserts that the claim is the inner product of the current pair.
     ///
-    /// Only meaningful once every binding has been applied.
-    /// A driver that holds bindings back therefore calls this at the end, not per round.
-    pub(crate) fn debug_assert_claim(&self) {
+    /// Applies an outstanding binding first, so the assertion is against the pair
+    /// the claim describes rather than the pair one round behind it.
+    pub(crate) fn debug_assert_claim(&mut self) {
+        self.settle();
```
Nashtare: `debug_assert_claim` now does load-bearing work in **release** builds: `settle()` is a full fold pass over both tables, and only the `debug_assert_eq!` after it is compiled out. The name promises debug-only, which invites someone to `#[cfg(debug_assertions)]` the call or delete it.

Today it survives that, because the sole caller (`zk/prover/residual.rs:170`) is followed by `scale_weights_and_claim` (`:1355`), which settles independently — which means the `settle()` here is in fact **redundant work in release**, not just misnamed. Either drop the call at `residual.rs:170` and adjust the comment there ("the weight scaling below reads the tables, so this settles it"), or rename to something like `settle_and_assert_claim` so the release-mode cost is visible at the call site.

binary-pcs/src/prover.rs
```diff
@@ -135,6 +135,40 @@ pub(crate) fn fold_rounds<MT, Ch>(
...
+/// The two produce the same round polynomials, so the tests can pin one against the
+/// other.
+#[must_use]
+#[allow(clippy::type_complexity)]
+fn fold_rounds_with<MT, Ch>(
+    prover_data: BinaryPcsProverData<MT>,
+    config: &BinaryPcsConfig,
+    mmcs: &MT,
+    challenger: &mut Ch,
+    bind_each_round: bool,
```
Nashtare: `bind_each_round` and its `if bind_each_round { sumcheck.settle(); }` branch at `:215` ship in production builds, but the only caller that ever passes `true` is inside `#[cfg(test)]` (`fold_rounds_binding_each_round` at `:282`, with the `#[cfg(test)]` starting at `:261`).

A test-only mode reachable from release code is the kind of knob that grows a real caller later. Consider `#[cfg(test)]`-gating the parameter, or duplicating the short loop in the test module.

binary-pcs/src/prover.rs
```diff
@@ -198,6 +238,13 @@ where
         }
     }
 
+    // The last round's challenge stays unapplied, and is never applied.
+    //
+    // The codeword fold, not the sumcheck tables, is what carries the folded message
+    // forward, and the sumcheck goes out of scope here without anything else reading it.
+    // The final binding pass would therefore only compute a table nobody looks at.
+    drop(sumcheck);
```
Nashtare: Debug-assertion coverage regressed here, worth a comment at minimum.

The old `compute_sumcheck_polynomials` ended with `bind_pending()` + `debug_assert_claim()`, so `sum == <evals, weights>` was verified against the fully-bound tables after every call. That trailer is gone; the invariant now lives inside `measure_round` (`sumcheck/src/strategy.rs:1315`), which covers every bind *except the last one held*. Dropping the final challenge here means the last fused bind is validated in no build profile.

That bind's output is genuinely unused, so this is not a correctness bug — but a defect in the last fused bind would be invisible to the suite. A one-line note saying the final challenge is intentionally discarded and why would make that explicit for the next reader.

Nashtare (changes requested):

sumcheck/src/strategy.rs
tcoratger: Body rewritten (`gh api -X PATCH repos/Plonky3/Plonky3/pulls/2055`). It now covers what this change actually does, and the two things that were missing:

- Why the challenge is deferred rather than applied on the spot: binding at sample time cannot fuse, because the bind for `r_i` and the measure for `h_{i+1}` end up separated by the transcript work and each has to find the tables in memory on its own. Deferring lets one block be bound and measured while it is still in cache. The cost is that the tables lag the claim by one binding, which is why `settle()` exists and why every table reader calls it, `num_variables()` excepted because it is a length.
- The ordering property, stated as a property: per round the loop is `measure_round()` -> `observe_and_sample` -> `reduce_claim` -> `hold(r_i)` (`strategy.rs:1434-1452`), `hold` is the only writer of `outstanding`, and it runs after the transcript has absorbed `h_i`. So round `i`'s message is a function of `r_0..r_{i-1}` only and the prover never sees `r_i` before committing to `h_i`. Same ordering in the ZK overlay (`residual.rs:132/148/160/161`).

Also in the body: the `&self` -> `&mut self` break from your other comment, with a table of the four accessors, and the buffer-retention trade with the bound the fix now enforces.

Test plan lists only what was run on this pass — `cargo nextest run -p p3-sumcheck -p p3-binary-pcs`, the same with `--features parallel`, `cargo clippy -p p3-sumcheck -p p3-binary-pcs --all-targets -- -D warnings`, `cargo +nightly fmt --all` — and says explicitly what was not run (benches, workspace tests, non-default target-feature legs, peak RSS).

sumcheck/src/strategy.rs
tcoratger: Fixed rather than documented, in c56a30e0 (`strategy.rs:812`).

Your trace is right, and the overshoot is worse than 2x: the two slots alternate and each round halves, so the storage a round hands back is *four* times the length the next round writes. That is why no truncate-only policy ever reclaims — the buffer is always exactly 4x what the destination needs, so the `buffer.len() >= len` branch is taken forever and the round-1 `N`-capacity `Vec` lives until drop.

`resize_destination` now keeps a buffer only while `capacity <= 2 * len`, and takes a fresh `A::zero_vec(len)` otherwise, released before the replacement is requested so the two never coexist. Two is the smallest factor that still lets a buffer survive one halving, which is the reuse the swap exists for.

Footprint, per side, as arithmetic on the lengths:

```text
    before:  N/2 live table  +  N buffer capacity, held to drop   ->  3N across evals + weights
    after:   2^{n-i} live table  +  <= 2^{n-i+1} buffer           ->  follows the tables down
```

`reused_fold_buffers_bind_every_suffix_round_correctly` now asserts `buffers.{evals,weights}.capacity() <= 2 * live` after every round of a 2^15 sumcheck, so the bound is pinned rather than argued.

The replacement is `zero_vec` and not `resize`/a threaded fill precisely because it is now per-round: for `BinaryField128` and the Monty extensions `zero_vec` bottoms out in `vec![0u128; len]` / `vec![0u32; len]`, i.e. pages the allocator has already zeroed, so a fresh destination costs no userspace fill. Every entry is overwritten before anything reads it.

I did not measure peak RSS on this host — heavy runs (benches, `/usr/bin/time -v` on a large sumcheck) were out of scope for this pass, so the numbers above are arithmetic plus the test's capacity assertion, not a measurement.

sumcheck/src/strategy.rs
tcoratger: Both added.

PR description now has a table of the four accessors and their old/new receivers, says plainly that reading the prover is a mutating operation and that a downstream crate holding a `&SumcheckProver` will not compile, and notes that the settle-on-read is what forces it.

`CHANGELOG.md` gets an `[Unreleased]` section in the file's existing `### Merged PRs` shape (f97e2540), with the receiver change named in the entry. The file has only ever been written at release time, so `[Unreleased]` is new; if you would rather the release tooling generate it, say so and I will drop the section and leave the break in the description only.

sumcheck/src/strategy.rs
tcoratger: Took the first option, in 83b19f9e.

Traced it first: `scale_weights_and_claim` (`strategy.rs:1366`) opens with `self.settle()`, and it is the next statement after the `debug_assert_claim()` call, so the `settle()` inside `debug_assert_claim` was a second full fold pass in release for one reader. Confirmed redundant.

So `residual.rs:170` goes straight to the weight scaling, with the comment saying the scaling settles on the way in. `debug_assert_claim` itself is gone: its body was `settle(); debug_assert_eq!(sum, dot_product())`, and `settle()` already carries that same `debug_assert_eq!` against the pair it just bound, so the assertion is kept where the binding happens rather than in a wrapper that lies about its cost. Dropping the caller would otherwise have left it dead.

binary-pcs/src/prover.rs
tcoratger: Removed the runtime knob in 4584bd6a.

`fold_rounds_with` now takes `const BIND_EACH_ROUND: bool` instead of a `bool` argument, so `fold_rounds`' signature carries nothing and `if BIND_EACH_ROUND { sumcheck.settle(); }` is a compile-time branch. The `true` instantiation is reachable only from `fold_rounds_binding_each_round`, which is `#[cfg(test)]`, so a release build monomorphizes the fused route only — there is no runtime value that can turn the two-pass mode on.

Preferred this over duplicating the loop in the test module: the loop is the fold/commit/observe body the test is meant to be checking, and a copy would drift from it.

binary-pcs/src/prover.rs
tcoratger: Restored the check instead of only commenting, in 4ddf476c (`prover.rs:250`).

`settle()` carries the `debug_assert_eq!(sum, dot_product())` against the pair it just bound, so a `#[cfg(debug_assertions)] sumcheck.settle();` before the `drop` gives back exactly the coverage the old `bind_pending()` + `debug_assert_claim()` trailer had, and costs a release build nothing.

The comment above it now says both halves: the challenge is discarded in release because the codeword fold, not the sumcheck tables, carries the folded message forward and nothing else reads the prover; and a debug build applies it anyway purely for that claim check, which is the last binding's only validation.

sumcheck/src/strategy.rs
tcoratger: Correcting my own follow-up here: I initially added an `[Unreleased]` section to `CHANGELOG.md`, then dropped it again, because that file is not hand-written in this repo.

`release-plz.toml` points `changelog_config` at `cliff.toml`, whose commit parser is `.*\(#[\d]+\).*` grouped as `Merged PRs`, and whose body template emits the `## [unreleased]` heading itself. So the section is generated from squash-merge subjects at release time — the only human commit ever made to that file is the `v0.3.0` release commit (`369494e7`). A hand-written section would have been duplicated or clobbered on the next release run.

The break is instead recorded where this repo actually records it: the PR title is now `perf(sumcheck)!:`, which is the squash subject and therefore the changelog line, matching `feat(fri)!:` / `fix(fri)!:` / `feat(sumcheck)!:` upstream.

The description now carries the receiver change explicitly, as a table of `evals` / `evals_view` / `eval` / `weights` before and after, with the downstream consequence spelled out — a `&SumcheckProver` no longer compiles, which is the point you were making.

sumcheck/src/strategy.rs
```diff
@@ -548,6 +567,371 @@ where
...
+/// A replacement is allocated already zeroed, so no userspace pass fills it.
+///
+/// Every entry is overwritten before anything reads it, so the values do not matter.
+#[inline]
+fn resize_destination<A>(buffer: &mut Vec<A>, len: usize)
+where
+    A: PrimeCharacteristicRing,
+{
+    // Reuse: the buffer already covers this round and does not overshoot it badly.
+    if buffer.len() >= len && buffer.capacity() <= 2 * len {
```
Nashtare: **This branch is unreachable in the usage the whole `FoldBuffers` design exists for. Every round reallocates.**

Trace the invariant the docs above set out. Write `L_i` for the table length entering round `i`, so `L_1 = 2^n` and `half_i = L_i / 2 = L_{i+1}`.

- `swap_storage` hands the buffer *the table's old storage*: after round `i`, `buffers.evals.len() == L_i == 2 * half_i == 4 * half_{i+1}`. `Poly::new` and `into_evals` are both moves — neither shrinks — so `capacity >= 4 * half_{i+1}`.
- Round `i+1` calls `resize_destination(buffer, half_{i+1})`. The reuse test is `capacity <= 2 * len`, i.e. `4 * half_{i+1} <= 2 * half_{i+1}`. False, always.
- Round 1 takes the other branch too, because the buffer starts empty (`0 >= half_1` is false).

So `truncate` never runs. Every round drops the buffer and calls `A::zero_vec(half)`. The doc directly above says

> Two is the smallest factor that still lets a buffer survive one halving.

but a buffer is handed over at **four** times the next destination, never two, because the swap skips a generation: the buffer a round receives is the table from the round *before* it. Two is one factor too small by construction. `4` would make the branch live, at the cost of the table then carrying a 4x-oversized capacity through `swap_storage` — which is presumably why 2 was chosen, but then the swap machinery has nothing left to do.

I would fix this by deleting the mechanism rather than retuning the constant: see the comment on `FoldBuffers`.

sumcheck/src/strategy.rs
```diff
@@ -548,6 +567,371 @@ where
...
+/// The two slots alternate and every round halves.
+///
+/// So the storage a round hands over is four times the length the next round writes.
+///
+/// The resize below releases an overshoot that large rather than holding it.
+/// The footprint therefore follows the tables down.
+///
+/// The first round's full-size allocation is released rather than pinned until drop.
+#[derive(Debug, Clone)]
+pub struct FoldBuffers<A> {
```
Nashtare: Given that `resize_destination`'s reuse branch never fires (see the comment there), `FoldBuffers` + `swap_storage` + `resize_destination` reduce to *exactly* `let dst = A::zero_vec(half); ... ; *table = Poly::new(dst);` — a plain out-of-place fold — while being strictly worse than one on the axis the PR's "Memory" section cares about:

| | resident between rounds, per side |
|---|---|
| plain out-of-place fold (source dropped) | `half` |
| this | `half` (table) + `2 * half` (old storage parked in the buffer) = `3 * half` |

The old table is not freed when the round ends; it sits in `buffers` until the *next* round's `resize_destination` drops it. So across `evals + weights` the steady state is `6 * half` where a plain fold holds `2 * half`. The PR's own arithmetic —

> across evals + weights: 3N, against N for an out-of-place bind that frees

— is the correct comparison, and the conclusion is that the buffer design loses it. The `resize_destination` guard only prevents the footprint from *staying at round one's size*; it does not recover the plain fold's number.

Concretely, at the `2^28 BinaryField128` case the PR raises, this is the difference between ~4 GiB and ~12 GiB held across `fold_codeword` and `mmcs.commit_matrix`.

So: this is ~90 lines including a new `pub` type in a public module, an extra `Clone`-able field on the public `ProductPolynomial`, and a new `#[derive(Debug)]` surface, in exchange for a 3x larger resident footprint and the same allocation count as three lines of `zero_vec`. Unless the reuse branch can be made live *and* shown to beat the allocator (which `zero_vec`'s own doc argues against — it takes pre-zeroed pages), I think the right move is to drop `FoldBuffers` entirely and have `bind_and_measure_pairs` allocate its own destinations.

sumcheck/src/strategy.rs
```diff
@@ -1346,6 +1791,155 @@ mod tests {
...
+                "round {round}"
+            );
+            assert_eq!(got.c_a, want.c_a, "round {round}");
+            assert_eq!(got.c_inf, want.c_inf, "round {round}");
+
+            // The buffer now holds the storage the table just handed over.
+            // Keeping more than twice the live table would pin memory no round reaches.
+            let live = got_evals.as_slice().len();
+            assert!(
+                buffers.evals.capacity() <= 2 * live,
```
Nashtare: This assertion cannot fail. After `swap_storage`, `buffers.evals` *is* the table's previous storage, whose length is exactly `2 * live` by construction, and `Vec` capacity is preserved across the move — so `capacity <= 2 * live` holds with equality on every round, for any implementation of `resize_destination` that does not grow the buffer.

Which means the test named `reused_fold_buffers_bind_every_suffix_round_correctly` proves the values are right (that part is genuinely useful) but the "reused" half of its name asserts a property that is both vacuous and, per the comment on `resize_destination`, false — no buffer is ever reused.

If the mechanism is kept, the assertion that would have caught this is on the *allocation*: e.g. record `buffers.evals.as_ptr()` before and after and assert the pointer is unchanged when reuse is expected. That test would fail today.

sumcheck/src/strategy.rs
```diff
@@ -548,6 +567,371 @@ where
...
+/// - `r` - challenge the round variable binds to.
+///
+/// # Returns
+///
+/// - `c_a` - the bound pair's round polynomial at 0.
+/// - `c_inf` - its leading coefficient.
+///
+/// # Performance
+///
+/// O(2^n), at the same multiply count as binding and measuring separately.
```
Nashtare, line 722: > # Performance
> O(2^n), at the same multiply count as binding and measuring separately.
> What it saves is one pass over the bound tables.

This is the PR's entire thesis and there is no number behind it anywhere — the test plan says `cargo bench` was not run. That is a lot to take on faith for +1229/-100, a public API break on four accessors, and a new public type.

The claim is also not obviously a clean win in this direction, unlike the prefix case it mirrors:

- The prefix kernel binds **in place**, so the fused pass genuinely removes a whole traversal and allocates nothing.
- The suffix kernel cannot, so the fused pass writes a fresh half-size destination every round (and, per the `resize_destination` comment, allocates it every round). Against the two-pass reference — `fix_suffix_var_mut` (which *does* fold in place below its own `should_split` gate) followed by `sumcheck_coefficients_suffix` — the fused route trades one read pass for one allocation plus one extra write pass at small sizes. Where the crossover sits is exactly what a benchmark would say.

The shapes that matter are the ones `binary-pcs` actually drives: single-round calls at `2^n` down to `4`, i.e. the whole tail where the serial branch runs and the allocation is proportionally largest. A `bench` row per round arity, fused vs `fold_rounds_binding_each_round` (which this PR conveniently already provides), would settle both this and the `FoldBuffers` question.

sumcheck/src/strategy.rs
```diff
@@ -883,54 +1292,93 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
-        if let Some(r) = pending.take() {
-            self.poly.fold_round(r);
-        }
+    /// Panics if a challenge is already outstanding.
+    /// Two unapplied bindings cannot be fused into one pass.
+    ///
+    /// The second would overwrite the first and lose a variable.
+    pub(crate) fn hold(&mut self, r: EF) {
+        assert!(
+            self.outstanding.replace(r).is_none(),
```
Nashtare: `Option::replace` writes first and returns the old value, so on the panic path `r` has already overwritten the challenge this assertion exists to protect. If a caller ever catches the unwind (a `catch_unwind` in a fuzz harness, a test with `should_panic` that then inspects state, a `Drop` running during unwind), the prover is left holding the *second* challenge with the first silently lost — the exact failure the message describes.

```rust
assert!(self.outstanding.is_none(), "a challenge is already outstanding");
self.outstanding = Some(r);
```

is the same cost and does not mutate before it checks.

sumcheck/src/strategy.rs
```diff
@@ -883,54 +1292,93 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
 
     /// Returns the number of remaining (unbound) variables.
+    ///
+    /// An outstanding binding has already consumed a variable.
+    ///
+    /// The tables are still the length they had before it.
+    /// Subtracting it is exact, so the answer costs no pass over the data.
     pub fn num_variables(&self) -> usize {
-        self.poly.num_variables()
+        self.poly.num_variables() - usize::from(self.outstanding.is_some())
```
Nashtare: `self.poly.num_variables() - usize::from(self.outstanding.is_some())` has no lower guard. If a challenge is ever held on a poly that is already down to one evaluation this underflows: a panic in debug, and `usize::MAX` in release, from a `pub` accessor. The callers in this PR then do things like `Point::<EF>::rand(&mut rng, prover.num_variables())` and `EF::zero_vec(1 << prover.num_variables())`.

It is unreachable today — `hold` only runs after `measure_round`, whose kernels assert `>= 4` entries or fall back to `fold_round` + `round_coefficients` — but that is a chain of four invariants across three files protecting a `pub` method, and the release-mode failure is the silent one. `saturating_sub(1)` or a `debug_assert!(self.poly.num_variables() > 0)` next to it makes the reasoning local.

sumcheck/src/strategy.rs
```diff
@@ -883,54 +1292,93 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
+    /// A debug build checks the claim against the pair this binding produced.
+    /// That is the only place a held binding is ever validated.
+    ///
+    /// Idempotent, and free when nothing is outstanding.
+    pub fn settle(&mut self) {
+        if let Some(r) = self.outstanding.take() {
+            self.poly.fold_round(r);
+
+            // Invariant: the claim is the inner product of the now-current pair.
+            debug_assert_eq!(self.sum, self.poly.dot_product());
```
Nashtare, line 1228: This moves the claim check from once-per-batch (`debug_assert_claim`, deleted) to once-per-round, and `settle()` adds a second one. `dot_product()` is a full pass over both tables, so a debug/test build of an `n`-round sumcheck now does roughly `2 * sum_i 2^{n-i} ~ 4 * 2^n` extra element-multiplies on top of the protocol itself.

That is asymptotically fine, but this PR also adds `reused_fold_buffers_bind_every_suffix_round_correctly` at `2^15` and `fold_rounds_agree_with_binding_each_round` at 15 variables, both of which run the whole ladder twice — and `binary-pcs`'s own debug tests are already the slow ones in CI. Worth checking the wall-clock delta on `cargo nextest run -p p3-sumcheck -p p3-binary-pcs` before and after; if it is material, gating the per-round check behind a feature (or keeping it only in `settle`) preserves the diagnostic without paying it on every round of every test.

binary-pcs/src/prover.rs
```diff
@@ -198,6 +243,19 @@ where
...
 
+    // The last round's challenge is discarded rather than applied.
+    //
+    // The codeword fold, not the sumcheck tables, carries the folded message forward.
+    // The sumcheck goes out of scope here with nothing else reading it.
+    // A final binding pass would only compute a table nobody looks at.
+    //
+    // A debug build applies it anyway, purely to check the claim against it.
+    // `settle` is where that check lives, so this is the last binding's only validation.
+    #[cfg(debug_assertions)]
```
Nashtare, line 235: Two notes on this pair.

**The cfg makes debug and release do different work, and the comment undersells it.** In release the last round's challenge is never applied at all — `sumcheck` is dropped holding it. That is fine as long as nothing else reads the tables, which is asserted only by the comment above. But it also means the *reference* route (`BIND_EACH_ROUND = true`) and the fused route end in genuinely different states in release, and `fold_rounds_agree_with_binding_each_round` compares round messages, randomness, commitments and the final codeword — none of which reads the sumcheck tables after the loop. So the one thing the cfg'd `settle()` validates is validated only in the build where it runs. Since the whole point of the `outstanding` design is that the tables may lag, an explicit `debug_assert!` that nothing downstream consumes `sumcheck` would be more honest than a `drop()`.

**`drop(sumcheck)` is a no-op in release.** `sumcheck` goes out of scope on the next line either way; the explicit drop only orders it before the tuple construction, which borrows nothing from it. If the intent is documentation, a comment says it; if the intent is to release the tables before the return value is built, that is already what NLL does. As written a reader has to check whether the ordering is load-bearing.

sumcheck/src/strategy.rs
tcoratger: You are right, and I reproduced it rather than reasoning about it.

I instrumented the loop to print the buffer's pointer, length and capacity after every round of a 2^15 suffix sumcheck:

```
round  0: live= 16384 buf.len= 32768 buf.cap= 32768 ptr_same_as_prev=false
round  1: live=  8192 buf.len= 16384 buf.cap= 16384 ptr_same_as_prev=false
round  2: live=  4096 buf.len=  8192 buf.cap=  8192 ptr_same_as_prev=false
...
round 13: live=     2 buf.len=     4 buf.cap=     4 ptr_same_as_prev=false
```

`ptr_same_as_prev` is `false` on every round, so `truncate` never runs and every round reallocates, exactly as you traced. And `cap == 2 * live` on every round, which is the other half of your point: the guard is `capacity <= 2 * len` while the buffer arrives at `4 * len`, so the reuse test is false by construction.

Fixed by deleting the mechanism, as you suggested, not by retuning the constant. `resize_destination`, `swap_storage` and `FoldBuffers` are gone; `bind_and_measure_pairs` now allocates the two half-length destinations it writes and hands them back, and `fold_and_round_coefficients_suffix` installs them, which frees the sources at the end of the round. See the reply on the `FoldBuffers` thread for the footprint arithmetic.

sumcheck/src/strategy.rs
tcoratger: Agreed, and done: `FoldBuffers`, `swap_storage` and `resize_destination` are deleted.

Your table is what settled it. Confirmed by instrumentation on the other thread that the reuse branch never fires, so the mechanism was a plain out-of-place fold that additionally parked the old table in `buffers` until the next round dropped it.

What the code does now:

```rust
let (bound_evals, bound_weights, message) =
    bind_and_measure_pairs(evals.as_slice(), weights.as_slice(), r);

*evals = Poly::new(bound_evals);
*weights = Poly::new(bound_weights);
```

Installing the bound halves drops the sources there, so between rounds the pair holds `half` per side rather than `3 * half`.

| | resident between rounds, per side |
|---|---|
| before | `half` (table) + `2 * half` (old storage parked in the buffer) = `3 * half` |
| now | `half` |

Surface removed with it: the `pub struct FoldBuffers` and its `Default`/`new`, the `buffers` field on `ProductPolynomial`, and the `buffers` argument on the `pub fn fold_and_round_coefficients_suffix`. `sumcheck/src/strategy.rs` is -186/+84 on this pass.

The allocation count is unchanged, since the reuse branch never fired: one `zero_vec(half)` per table per round, before and after.

sumcheck/src/strategy.rs
tcoratger: Vacuous, confirmed. The instrumented run on the `resize_destination` thread shows `cap == 2 * live` with equality on every one of the 14 rounds, so `capacity() <= 2 * live` could not have failed.

The assertion is gone with the mechanism it was guarding. Two tests carry what is left:

- `every_suffix_round_binds_correctly_and_keeps_only_the_live_table` (renamed from `reused_fold_buffers_...`, since nothing is reused) keeps the round-by-round value comparison against `fix_suffix_var_mut` + `sumcheck_coefficients_suffix`, and now also pins that each round leaves exactly `2^{n-1-i}` entries per side.
- `the_fused_suffix_pass_allocates_exactly_the_bound_half` reads the destinations `bind_and_measure_pairs` returns and asserts `len == capacity == half`. That is the footprint claim stated where it is actually observable, and it fails if a destination is ever over-allocated.

Neither is vacuous in the way the old one was, but I want to be straight with you: the "nothing is retained between rounds" property is now structural (the sources are dropped when the bound halves are installed) rather than asserted, because `Poly` exposes no capacity accessor to assert it through.

sumcheck/src/strategy.rs
tcoratger: Fixed.

```rust
// Checked before the write, so a rejected challenge leaves the held one intact.
assert!(
    self.outstanding.is_none(),
    "a challenge is already outstanding"
);
self.outstanding = Some(r);
```

`holding_a_second_challenge_without_settling_is_rejected` now catches the unwind and inspects the state instead of just asserting the panic message:

```rust
prover.hold(EF::ONE);
let rejected = catch_unwind(AssertUnwindSafe(|| prover.hold(EF::TWO)));
assert!(rejected.is_err());
assert_eq!(prover.outstanding, Some(EF::ONE));
```

Mutation-checked against the shipped form. Reverting to `self.outstanding.replace(r).is_none()`:

```
left:  Some(ExtField { value: [2, 0, 0, 0], ... })
right: Some(ExtField { value: [1, 0, 0, 0], ... })
```

so the prover was left holding the second challenge with the first lost, which is the failure you described.

sumcheck/src/strategy.rs
tcoratger: Fixed, with both halves of what you suggested:

```rust
pub fn num_variables(&self) -> usize {
    // A binding is only ever held on a pair that still has a variable to bind.
    debug_assert!(!(self.outstanding.is_some() && self.poly.num_variables() == 0));

    // Saturating, so the count of a fully bound pair stays zero rather than wrapping.
    self.poly
        .num_variables()
        .saturating_sub(usize::from(self.outstanding.is_some()))
}
```

The `debug_assert!` states the precondition locally, so the reasoning no longer has to be reassembled from four invariants across three files, and `saturating_sub` removes the release-mode `usize::MAX` outright rather than trading it for a debug panic.

binary-pcs/src/prover.rs
tcoratger: Both taken.

**`drop(sumcheck)` removed.** You are right that it was a no-op: nothing in the returned tuple borrows from it, so NLL already ends its lifetime there, and the explicit call only made a reader check whether the ordering mattered.

**The comment now says the static fact rather than the dynamic one.** It read "the sumcheck goes out of scope here with nothing else reading it", which is a claim about this function's body. What actually guarantees it is the signature: the returned tuple carries no sumcheck state, so no caller can reach the lagging tables. That is what the comment says now.

On the debug/release divergence: I kept the `#[cfg(debug_assertions)] settle()` rather than making it unconditional or asserting something weaker. It is a full fold pass over both tables, which is real cost at production sizes for a table nothing reads, and it is the last held binding's only validation in any profile. The comment states both of those plainly. You are right that `fold_rounds_agree_with_binding_each_round` does not exercise it — it compares round messages, randomness, commitments and the final codeword, none of which touch the tables — so the coverage that check provides is the `debug_assert_eq!(sum, dot_product())` inside `settle`, and nothing more. I did not want to overstate it.

sumcheck/src/strategy.rs
tcoratger: Benched, and your prediction about the small end is right.

New `sumcheck/benches/sumcheck.rs::bench_single_round_drive` drives a whole ladder one round per call, which is the shape `binary-pcs` uses. Two arms over the identical ladder: `fused` (the challenge is held, the next round's pass absorbs it) and `bind_each_round` (`settle()` after each round, so the next round measures in a pass of its own — the same two-pass shape `fold_rounds_binding_each_round` produces).

`RUSTFLAGS="-C target-cpu=native" cargo bench -p p3-sumcheck --bench sumcheck -- babybear/single_round_drive`, Zen 5, BabyBear degree-4 extension, criterion mean:

| k | order | fused | bind_each_round | ratio |
|---|---|---|---|---|
| 8 | prefix | 5.26 us | 5.23 us | 1.00x |
| 8 | suffix | 6.85 us | 6.67 us | **0.97x** |
| 12 | prefix | 18.89 us | 20.27 us | 1.07x |
| 12 | suffix | 49.90 us | 55.99 us | 1.12x |
| 16 | prefix | 206.72 us | 237.88 us | 1.15x |
| 16 | suffix | 757.52 us | 977.42 us | 1.29x |
| 20 | prefix | 4.524 ms | 5.846 ms | 1.29x |
| 20 | suffix | 16.770 ms | 20.065 ms | 1.20x |

Reading it:

- **The small end goes the way you said.** At `k = 8` the suffix ladder is 3% *slower* fused. The whole ladder there runs below `PARALLEL_THRESHOLD`, where `fix_suffix_var_mut` folds fully in place and allocates nothing, so the fused route's per-round half-size destination is pure added cost with no read pass saved worth the trade. I would not call 3% conclusive on its own, but the direction is exactly the one you predicted and the mechanism is the one you named.
- **The crossover sits between `k = 8` and `k = 12`** for suffix, and the fusion is a clear win from `k = 12` up.
- **The ladders are whole.** The `k = 20` row folds all the way down to two entries, so it already contains the small-size regime at its tail, and it is still 1.20x–1.29x. The cost is geometric in the large rounds, so the tail cannot undo the win.

So the thesis holds at the sizes that carry the cost, and it does not hold at the very bottom of the ladder. I have not added a size gate for that: a 3% loss on rounds that are microseconds out of a millisecond-scale ladder is not worth a branch and a second code path, and the `k=20` rows are the number that matters for `binary-pcs`.

On the other half of your comment — whether the split-`FoldBuffers` machinery earns itself — the answer turned out to be no, and it is deleted. See the `FoldBuffers` thread.

Caveat on the numbers: this host was running other builds throughout. Both arms sit in the same criterion invocation back to back, so drift between them is small, but treat sub-5% rows as noise.

sumcheck/src/strategy.rs
tcoratger: Measured. Interleaved A/B on the identical tree, the only difference being the `debug_assert_eq!(self.sum, self.poly.dot_product())` in `measure_round`, timing `cargo nextest run -p p3-sumcheck -p p3-binary-pcs` (debug) with the build outside the timed region:

| iteration | with the check | without | delta |
|---|---|---|---|
| 1 | 65.66 s | 61.38 s | +4.28 s (+7.0%) |
| 2 | 53.91 s | 51.95 s | +1.96 s (+3.8%) |
| 3 | 95.15 s | 46.90 s | *discarded* |

Iteration 3 is not usable — this host was shared with other builds and load average went from ~5 to ~29 during that pair, which is where the 95 s comes from. The two clean pairs put the cost at **+2 s to +4 s on a ~55 s suite, so 4–7%**.

My read is that this is not material enough to gate, and I have left it in place. What it buys is specific: in the fused flow, round `i`'s `measure_round` is what applies `r_{i-1}`, so this assertion is the primary validation of the new binding path. Removing it would leave a fused bind checked only when a later `settle()` happens to run, and `binary-pcs` in release never calls one (see the thread on `prover.rs`). Trading the only check on the code this PR adds for ~3 s of debug test time seems like the wrong side of that trade.

I also did not add a feature flag for it. You objected — rightly — on another thread to a test-only knob reachable from release code, and a `cfg` that exists only so CI can skip an assertion is the same shape of thing. If suite time becomes the binding constraint I would rather cut the two `2^15` ladder tests down a few variables, which costs coverage nowhere.

Happy to reverse either call if you disagree.

binary-pcs/src/pcs.rs
```diff
@@ -540,6 +541,142 @@ mod tests {
...
     }
 
+    /// Mirrors the shipped opening entry point.
+    ///
+    /// The fold rounds take the reference route, which applies each binding on its own pass.
+    ///
+    /// Every other step is the shipped one.
+    ///
+    /// So the two proofs can only differ if a round polynomial did.
+    fn open_binding_each_round(
```
Nashtare, line 583: **This mirror has already drifted against the shipped `open`, and the drift is silent.**

It reproduces the pre-merge `open`. On the current head, `open`/`open_at` delegate to `try_open`/`try_open_at` (`pcs.rs:109`/`:133`), which call `validate_opening_protocol` and then funnel into `finish_open` (`pcs.rs:178`). This helper does neither: it records `evals` itself and assembles `BinaryPcsProof` by hand.

`fusing_the_binding_into_the_measuring_pass_leaves_the_proof_byte_identical` still passes, because the validation step is transcript-neutral. That is exactly the problem — the test now compares the shipped path against a copy that no longer matches it, and says nothing about the difference.

The seam already exists one level down. Extending it one level up removes the duplication for good:

- `fn finish_open_with<const BIND_EACH_ROUND: bool, Challenger>(...)` calling `fold_rounds_with::<BIND_EACH_ROUND, _, _>(...)`
- `finish_open` becomes a thin `finish_open_with::<false, _>` wrapper, still used by `try_open`/`try_open_at`
- the test calls `finish_open_with::<true, _>` after recording `evals` the way `try_open` does

This helper then disappears, and so does `fold_rounds_binding_each_round` in `prover.rs` if nothing else uses it. Worth keeping exactly one seam rather than two.

binary-pcs/src/prover.rs
```diff
@@ -427,6 +519,120 @@ mod tests {
...
+        //     8  variables: every fused pass takes the serial branch
+        //     15 variables: the early rounds take the threaded branch, where a fused pass
+        //                   writing its own input in place would race another task's reads
+        for num_variables in [8usize, 15] {
+            let params = BinaryPcsParams {
+                log_inv_rate: LOG_INV_RATE,
+                pow_bits: 4,
+                security_level: 40,
+            };
+            let config = BinaryPcsConfig::try_new(num_variables, params).unwrap();
```
Nashtare: **This config never produces a batch with more than one round, so the interaction this test exists to pin is not reachable from it.**

`BinaryPcsConfig::try_new` forwards to `try_new_with_folding(.., 1)` (`params.rs:108-113`), so `fold_batches()` yields `(0,1), (1,1), ...` — every batch has arity 1. The `if BIND_EACH_ROUND { sumcheck.settle(); }` at `prover.rs:220` sits inside the inner per-round loop precisely so it stays a per-round reference at arity > 1, and that placement is untested.

`fusing_the_binding_into_the_measuring_pass_leaves_the_proof_byte_identical` (`pcs.rs:623`) has the same gap.

`try_new_with_folding` / `try_with_folding` already exist and are used elsewhere (`grouped_mmcs.rs:361` uses `4`). Adding one batched arity to the existing loop over `[8, 15]` — say a `log_folding_factor` of 2 or 3 — would cover it without restructuring the test.

Worth doing here rather than later: a held challenge crossing a batch boundary and a held challenge crossing a round boundary inside a batch are different paths through the same code.

sumcheck/src/strategy.rs
```diff
@@ -1017,16 +1365,15 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
         Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F>,
     {
         // Optional constraint absorption: fold into the weight polynomial and update the sum.
+        //
+        // The constraint is sized against the current hypercube.
+        //
+        // So an outstanding binding has to land before the weights grow by it.
         if let Some(constraint) = constraint {
+            self.settle();
```
Nashtare, line 1379: **This is the one `settle()` whose ordering is load-bearing, and no test reaches it with a binding actually outstanding.**

Settling after `combine` rather than before would apply a weight delta sized to the post-binding hypercube against a table that is still a round long — either a wrong result or a length panic, depending on the constraint. So the ordering here is doing real work.

Every in-workspace caller reads the tables before passing `Some(constraint)`, which means this call always sees `outstanding == None`:

- WHIR `round()` — `evals_view()` at `pcs/prover/mod.rs:235` and `eval()` at `:249`, both before `:315`
- `layout/prover/mod.rs:407` builds the constraint from `&prover.evals()`
- `binary-pcs` passes `None` on every call

`accumulate_claim` and `scale_weights_and_claim` both get direct coverage in `every_table_accessor_settles_an_outstanding_binding`; this path does not. The same test shape covers it: run a batch so a challenge is left held, then call `compute_sumcheck_polynomials(.., Some(constraint))` without reading in between, and compare against a copy that was settled explicitly first.

sumcheck/benches/sumcheck.rs
```diff
@@ -619,6 +627,100 @@ where
     group.finish();
 }
 
+/// Benches a sumcheck driven one round per call, fused against binding each round.
+///
+/// This is the shape the binary PCS uses: it interleaves a codeword fold and a Merkle
```
Nashtare: **This premise no longer holds after the merge, and the bench models the shape it describes.**

`fold_rounds` is now a two-level loop: `config.fold_batches()` yields `(start, arity)`, `arity` sumcheck rounds run, and *then* one `fold_codeword_batch` and one commitment (`prover.rs:201-236`). The codeword fold and the Merkle commitment are interleaved per **batch**, not per round, so binary-pcs *can* ask for several rounds at once — it just doesn't, because the inner loop still passes `1`.

Nothing is broken by this: the fusion still fires on every round as the code stands, and it is still required at batch boundaries regardless of arity. But the justification in this doc comment and in the PR description ("cannot ask for several rounds at once") is now false, and the `single_round_drive` grid measures the pre-batching shape, so the 1.20x–1.29x headline is an upper bound on what binary-pcs actually sees once the folding factor is above 1.

Worth restating both in terms of `num_fold_batches`: the rounds that genuinely cross a call boundary are one per batch, plus all of them when arity is 1 (which is today's default).

sumcheck/src/strategy.rs
```diff
@@ -895,54 +1196,101 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
-    /// A held-back binding is absorbed into the measuring pass.
+    /// An outstanding binding is absorbed into the measuring pass.
     /// The round then reads its tables once instead of twice.
     ///
     /// The slot is cleared here.
     /// The caller puts this round's own challenge back into it.
-    pub(crate) fn measure_round(&mut self, pending: &mut Option<EF>) -> (EF, EF) {
-        match pending.take() {
+    pub(crate) fn measure_round(&mut self) -> (EF, EF) {
+        let message = match self.outstanding.take() {
```
Nashtare: **The slot is cleared before the fold that consumes it can fail.**

`take()` empties `outstanding`, then `fold_round_coefficients(r)` runs. That call can panic: when `num_evals() < 4` it falls through to `fold_round(r)` followed by `round_coefficients()`, which asserts the table length, so a caller that drives the ladder past its last variable panics there. Catch that unwind and the prover is left `outstanding: None` with tables that were never bound — a state that reads as consistent and is silently one round stale.

This is the same failure mode `hold` was hardened against two commits ago ("Checked before the write, so a rejected challenge leaves the held one intact", `:1288`), with `holding_a_second_challenge_without_settling_is_rejected` catching the unwind to assert it. Applying that standard here means reading the challenge without clearing it, and clearing only once the fold has returned:

```rust
let message = match self.outstanding {
    Some(r) => self.poly.fold_round_coefficients(r),
    None => self.poly.round_coefficients(),
};
self.outstanding = None;
```

Low severity — it needs `catch_unwind` plus a panicking fold — but the asymmetry with `hold` is what makes it worth closing.

sumcheck/src/strategy.rs
```diff
@@ -895,54 +1196,101 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
+            .saturating_sub(usize::from(self.outstanding.is_some()))
+    }
+
+    /// Applies an outstanding binding, so the tables are current with the claim.
+    ///
+    /// Every reader of the tables starts here.
+    /// A reader that skipped it would silently see the tables one round behind.
+    ///
+    /// A debug build checks the claim against the pair this binding produced.
+    /// That is the only place a held binding is ever validated.
```
Nashtare: This overstates it. A binding absorbed by the fused path is validated by `measure_round`'s own `debug_assert_eq!(self.sum, self.poly.dot_product())` at `:1274`, not here — and in the steady state that is where almost every binding is checked, since `settle` only ever sees the ones no measuring pass picked up.

Suggest scoping the sentence to what is actually unique about this site, e.g. "That is the only place a binding no measuring pass absorbed is ever validated."

Same sentence appears in the `fold_rounds_with` trailer in `binary-pcs/src/prover.rs:254` ("That is the last held binding's only validation, in any profile"), which is accurate as written because that binding genuinely has no successor.

sumcheck/src/product_polynomial.rs
```diff
@@ -477,40 +480,62 @@ impl<F: Field, EF: ExtensionField<F>> ProductPolynomial<F, EF> {
...
     pub(crate) fn fold_round_coefficients(&mut self, r: EF) -> (EF, EF) {
-        // The fused pass binds a prefix variable and needs the bound table to keep one.
+        // The fused pass needs the bound table to keep a variable for the message.
         // That is four entries in the table it starts from.
-        let fusable = self.order == VariableOrder::Prefix
-            && match &self.inner {
-                MaybePacked::Packed { evals, .. } => evals.num_evals() >= 4,
-                MaybePacked::Unpacked { evals, .. } => evals.num_evals() >= 4,
-            };
+        let fusable = match &self.inner {
```
Nashtare, line 510: Recording the arithmetic behind "no size gate is added for the 3%", since it is the kind of decision that gets re-litigated.

Below `PARALLEL_THRESHOLD` (`4096`, `multilinear-util/src/poly/mod.rs:23`) `fix_suffix_var_mut` folds fully in place and allocates nothing, which is where the measured regression at `k = 8` comes from — the fused route's half-size destination is pure added cost there. But on the shape that matters, a 2^20 binary-pcs ladder, the rounds under that threshold carry roughly 2^13 of about 2^21 entries, i.e. ~0.4% of the total pass work. The tail cannot move the total, which is consistent with the measured 1.20x–1.29x at `k = 20` even though those rows contain the sub-threshold regime.

So the call is right and this thread is a note, not a request.

For the record, if anyone ever does want that 3% back: the kernel's own doc already proves the compaction is safe for a serial forward sweep (`strategy.rs:660-700`), so a sub-threshold in-place variant needs no new safety argument — only a second code path, which is why it is not worth it now.

sumcheck/src/strategy.rs
```diff
@@ -1045,27 +1392,25 @@ impl<F: Field, EF: ExtensionField<F>> SumcheckProver<F, EF> {
...
             challenges.push(r);
 
-            // Hand this round's challenge to the next one.
-            pending = Some(r);
+            // Hand this round's challenge on, for the next measuring pass to absorb.
+            //
+            // The last one of this batch stays outstanding on return.
+            // A caller that comes straight back for another round fuses across the call;
+            // one that reads the tables instead settles it on the way in.
+            self.hold(r);
```
Nashtare, line 1406: One consequence worth adding to the description, since it frames the win as binary-pcs-only.

WHIR also nets a saved pass, at the end of a proof. Because this driver no longer binds after the loop, `final_round`'s trailing `compute_sumcheck_polynomials(final_sumcheck_rounds, ..)` leaves its last challenge outstanding, and nothing reads the tables afterwards — `round_state` is dropped at `whir/src/pcs/prover/mod.rs:194` without being touched again. That whole fold pass disappears whenever `final_sumcheck_rounds > 0`.

Every other WHIR round settles at the next round's `evals_view()` (`:235`), so the pass count there is unchanged, not worse. Small, but it is a real second beneficiary and costs a sentence to claim.

Nashtare (approved):

binary-pcs/src/pcs.rs
tcoratger: Done, and the seam is now carried up exactly as you laid out.

- `fold_rounds_with<BIND_EACH_ROUND>` is the only seam left. It picked up the `binary pcs fold rounds` tracing span, and both `fold_rounds` and `fold_rounds_binding_each_round` are gone — each was that function with the parameter fixed.
- `BinaryPcs::finish_open_with<BIND_EACH_ROUND, Challenger>` carries it one level up. `finish_open` is a one-line `finish_open_with::<false, Challenger>` wrapper, still what `try_open`/`try_open_at` call after `validate_opening_protocol`.
- The test helper is now six lines: record `evals` the way `try_open` records them, then `pcs.finish_open_with::<true, MyChallenger>(..)`. Nothing after the fold rounds is reproduced, so there is nothing left to drift.

Call sites that only wanted the shipped route (`verifier.rs`, the other `prover.rs` tests) read `fold_rounds_with::<false, _, _>` now. Slightly noisier at those four sites than a `fold_rounds` wrapper would be, but it keeps the count at one function rather than one plus a default-arg alias that only tests reach.

binary-pcs/src/prover.rs
tcoratger: Covered, in both tests.

`fold_rounds_agree_with_binding_each_round` now loops over `[(8, 1), (8, 3), (15, 1), (15, 3)]` via `try_new_with_folding`, and `fusing_the_binding_into_the_measuring_pass_leaves_the_proof_byte_identical` over `[1, 3]`. Assertion messages carry the arity so a failure names the schedule.

Your framing is the one in the fixture comment, since it is the reason the arity matters at all:

```text
    1: every round is its own batch, so a held challenge only crosses a batch
    3: three rounds share a batch, so it also crosses a round boundary inside one
       — which is where the reference route`s per-round settle() sits
```

`cargo nextest run -p p3-sumcheck -p p3-binary-pcs` — 425 passed, 1 skipped, with and without `--features parallel`.

sumcheck/src/strategy.rs
tcoratger: Right on both counts, and covered by `absorbing_a_constraint_settles_an_outstanding_binding`.

Shape is the one you sketched: run a batch so a challenge is left held, then call `compute_sumcheck_polynomials(.., 1, 0, Some(constraint))` with no accessor read in between, and compare against a copy settled explicitly first — messages, challenges, claim, and both tables. Five and nine variables, both binding orders.

The constraint has to be honest or the `debug_assert_eq!(sum, dot_product())` in `measure_round` fires on the way out, so each `EqStatement` claim carries the value the settled table really takes at its point. That is what keeps `sum + delta_sum == <evals, weights + delta>` true across absorption.

Mutation-checked against the ordering you describe: swapping to `combine` then `settle` panics at `product_polynomial.rs:664` with `left: 3, right: 4` — the length mismatch, exactly as predicted.

sumcheck/benches/sumcheck.rs
tcoratger: Fixed, and restated in `num_fold_batches` terms in both places.

The bench doc now says binary-pcs asks for one round at a time in its *inner* loop, and that the outer loop is what forces the boundary, with the count spelled out:

```text
    rounds fused across a call, if a batch were asked for in one go
        = num_fold_batches - 1
        = ceil(num_variables / log_folding_factor) - 1

    rounds fused across a call, as the inner loop asks today (arity 1)
        = num_variables - 1
```

Same block is in the PR description, with the note that `try_new` fixes `log_folding_factor = 1` (`params.rs:108-113`), so the grid measures the shipped schedule exactly and is an upper bound above arity 1. The "cannot ask for several rounds at once" sentence is gone from the summary too.

sumcheck/src/strategy.rs
tcoratger: Fixed. `settle`'s doc now reads "That is the only place a binding no measuring pass absorbed is ever validated." — your wording, since it is the accurate one.

Agreed that `measure_round`'s own `debug_assert_eq!` at `:1274` is where almost every binding is checked in the steady state; `settle` only ever sees the ones no measuring pass picked up. The `fold_rounds_with` trailer in `prover.rs` is left as it was, since that binding genuinely has no successor.

sumcheck/src/strategy.rs
tcoratger: Fixed, to the shape you wrote. `measure_round` reads `outstanding` and clears it only once the pass has returned:

```rust
let message = match self.outstanding {
    Some(r) => self.poly.fold_round_coefficients(r),
    None => self.poly.round_coefficients(),
};

// Reached only once the binding landed, so a caught unwind still holds it.
self.outstanding = None;
```

`a_panicking_fused_pass_leaves_the_challenge_held` pins it: a fully bound pair with a challenge in the slot, `measure_round` panics inside the fold, the unwind is caught, and the challenge is asserted still held. Mutation-checked — with `Option::take` the assertion reads `left: None, right: Some(1)`.

One correction to the failure mode, for the record. On the `num_evals() < 4` fallback the order is `fold_round(r)` *then* `round_coefficients()`, so the binding does land before the assert trips; under `take()` that particular unwind left a state that was actually consistent. The one that genuinely lost a binding is a challenge held on a pair with no variable left, where `compress` panics at `poly/mod.rs:459` before applying anything — which is the fixture the test uses. Either way the asymmetry with `hold` was the real defect and it is closed.

sumcheck/src/product_polynomial.rs
tcoratger: Thanks — recorded in the kernel's own doc so it travels with the code rather than living in a thread:

```text
    a 2^20 ladder reads ~2^21 entries in all
    its rounds below the threshold hold ~2^13 of them, or ~0.4%
```

with the reason no gate routes them back ("the tail of a ladder cannot move its total"). Your figures, since the threshold is compared against the *input* length: the rounds taking the serial branch have inputs 4096, 2048, ..., summing to ~2^13 against a ~2^21 total.

Left out the sub-threshold in-place variant, and the note that it would need no new safety argument — accurate, but recording a path we decided not to take invites someone to take it.

sumcheck/src/strategy.rs
tcoratger: Claimed, and verified before claiming it. The PR description now carries:

> WHIR is a second beneficiary, at the end of a proof. Because the driver no longer binds after its loop, `final_round`'s trailing `compute_sumcheck_polynomials(final_sumcheck_rounds, ..)` leaves its last challenge outstanding, and nothing reads the tables afterwards — `round_state` is dropped at `whir/src/pcs/prover/mod.rs:194` untouched. That whole fold pass disappears whenever `final_sumcheck_rounds > 0`. Every other WHIR round settles at the next round's `evals_view()` (`:235`), so the pass count there is unchanged rather than worse.

Confirmed the drop: `final_round` returns the tuple at `:396-400`, none of whose members hold sumcheck state, and `prove` never touches `round_state` after `:194`. The `delegate_final_fold` bracket plays nothing when `final_sumcheck_rounds == 0`, so the saving is exactly conditional on that being positive.

PR #2056, opened 2026-09-08
perf(stir): reduce prover work and add opt-in smaller proofs

Nashtare:
STIR currently interpolates some polynomials on oversized domains and evaluates known-zero coefficient tails. This reduces that work and adds two opt-in proof-size tradeoffs. Default options preserve the existing schedule and serialized proof bytes in the measured cases.

## Changes

Each optimization is a separate commit:

1. Size adjusted interpolation weights to the largest native matrix height needed at each opening point.
2. Recover coefficients with native-size inverse DFTs in folded rounds and PCS evaluation/LDE-extension paths.
3. Use degree-aware codeword transforms at expansion ratios of at least 8x, sharing the existing batched transform implementation.
4. Add optional early stopping via `StirOptions::max_log_final_poly_len`; derive the complete schedule for the selected final coefficient bound.
5. Add optional compact answers via `StirOptions::compact_answers`; omit answer coefficients from the proof and reconstruct them in the verifier at the same transcript position.

The options are exposed through `StirConfig` constructors and `TwoAdicStirPcs::with_options`. Prover and verifier must agree on them. Changing PCS options resets the configuration cache without changing previously configured clones.

## Measured against main

Baseline: `7230fc572870436e6651762f35c6c3f3f48960d2` (`origin/main`, freshly verified before opening this PR). Branch: `50e198d44316553df43fcd92f779e2e03751565e`.

Apple M4 Pro (14 cores, 48 GiB), macOS 26.6, Rust 1.98.0. Both revisions used the identical measurement harness and lockfile, `--profile optimized`, `RUSTFLAGS="-C target-cpu=native"`, thin LTO, one codegen unit and incremental compilation disabled. Serial and `parallel` builds were separate; parallel runs used `RAYON_NUM_THREADS=4`.

KoalaBear quintic extension, CapacityBound, 100-bit configuration, 2x starting blowup, PoW disabled and fixed seed `0xdeadbeef`. Standalone proving includes initial codeword evaluation/commitment and the full STIR proof. Setup/input cloning, output destruction, serialization and correctness checks are outside the timers; configuration and transform caches are warm.

Six paired process blocks alternate main/branch and branch/main. Each API has two warmups and five measurements per block (50 for verification). Times below are geometric means of block medians; changes use paired ratios. Intervals are descriptive 95% bootstraps over blocks, not individual calls. Negative means faster.

| Workload | Threads | main ms | Branch ms | Time change [95% interval] |
|---|---:|---:|---:|---:|
| STIR 2^18, arity 4 | Serial | 219.49 | 207.23 | -5.6% [-6.7%, -4.7%] |
| STIR 2^18, arity 8 | Serial | 188.11 | 178.59 | -5.1% [-7.5%, -1.5%] |
| STIR 2^20, arity 4 | Serial | 909.59 | 895.72 | -1.5% [-7.3%, +7.4%] — inconclusive |
| STIR 2^18, arity 4 | 4 | 77.07 | 77.50 | +0.6% [-3.0%, +4.2%] — inconclusive |
| STIR 2^18, arity 8 | 4 | 61.84 | 59.43 | -3.9% [-4.9%, -2.9%] |
| STIR 2^20, arity 4 | 4 | 316.49 | 301.54 | -4.7% [-6.8%, -2.7%] |

### Proof size

`final64` sets `max_log_final_poly_len=Some(6)`; `compact` sets `compact_answers=true`. Sizes are serialized Postcard bytes and agree between serial and parallel builds.

| Workload | main / branch default | final64 | compact | Both | Both vs main |
|---|---:|---:|---:|---:|---:|
| STIR 2^18, arity 4 | 114,015 | 103,851 | 108,155 | 98,731 | -13.41% |
| STIR 2^18, arity 8 | 91,556 | 83,313 | 87,616 | 79,913 | -12.72% |
| Merged PCS | 159,233 | 148,813 | 153,373 | 143,693 | -9.76% |

Compact answers alone increase verifier time by 11.5% (arity 4) and 11.8% (arity 8) in the four-thread builds, relative to branch/default. Early stopping alone reduces verifier time by about 9% in those builds. Optional-mode timings use four separate paired blocks with refreshed branch/default controls; they are not comparisons against main.

<details>
<summary>PCS opening timings and targeted entry points</summary>

All PCS matrices have width four. Uniform uses log2 height [18]; merged uses [18,17,16] with spread cap 2; separate domains uses [18,15,12] with spread cap 0. Ordinary cases share one opening point; distinct-points uses the merged layout with a separate point per matrix. Opening starts from an existing commitment.

| Workload | Threads | main ms | Branch ms | Time change [95% interval] |
|---|---:|---:|---:|---:|
| PCS distinct points | 4 | 91.08 | 91.02 | -0.1% [-3.2%, +2.9%] — inconclusive |
| PCS merged | 4 | 93.55 | 90.45 | -3.3% [-7.1%, -0.1%] |
| PCS separate domains | 4 | 95.26 | 104.21 | +9.4% [-0.6%, +19.0%] — inconclusive |
| PCS uniform | 4 | 81.40 | 79.78 | -2.0% [-4.8%, +2.4%] — inconclusive |
| PCS distinct points | Serial | 310.74 | 282.12 | -9.2% [-21.6%, -1.1%] |
| PCS merged | Serial | 275.87 | 247.74 | -10.2% [-18.1%, -3.1%] |
| PCS separate domains | Serial | 244.66 | 234.46 | -4.2% [-4.8%, -3.7%] |
| PCS uniform | Serial | 267.78 | 247.65 | -7.5% [-13.2%, -3.3%] |

Codeword cases evaluate on a size-2^18 domain with 2^15 or 2^12 coefficients. `commit_ldes` extends precomputed native 2x LDEs for heights [18,17,16] onto the tallest merged domain; preparation is untimed. Changed-shift evaluation recovers the height-2^16 matrix from that merged domain and evaluates it on a size-2^16 coset shifted by `GENERATOR^2`.

| Workload | Threads | main ms | Branch ms | Time change [95% interval] |
|---|---:|---:|---:|---:|
| Codeword, 64x expansion | 4 | 2.91 | 1.13 | -61.4% [-64.1%, -58.0%] |
| Codeword, 8x expansion | 4 | 2.59 | 1.70 | -34.3% [-35.6%, -33.1%] |
| PCS changed-shift evaluation | 4 | 6.56 | 1.20 | -81.7% [-82.7%, -80.0%] |
| PCS commit_ldes | 4 | 70.12 | 70.85 | +1.0% [-4.4%, +11.3%] — inconclusive |
| Codeword, 64x expansion | Serial | 8.51 | 3.14 | -63.1% [-65.9%, -59.3%] |
| Codeword, 8x expansion | Serial | 8.77 | 5.24 | -40.2% [-41.7%, -38.8%] |
| PCS changed-shift evaluation | Serial | 22.91 | 3.12 | -86.4% [-86.8%, -86.0%] |
| PCS commit_ldes | Serial | 270.24 | 265.60 | -1.7% [-6.6%, +7.3%] — inconclusive |

</details>

<details>
<summary>Optional-mode prover and verifier timings versus branch/default</summary>

Each option has four paired blocks, with a refreshed default in every block and rotated/reversed option order. PCS options were measured for verified proof bytes only.

| Workload | Threads | Option | Branch/default ms | Branch ms | Time change [95% interval] |
|---|---:|---|---:|---:|---:|
| STIR 2^18, arity 4 / prove | 4 | both | 76.66 | 81.74 | +6.6% [-4.4%, +30.0%] — inconclusive |
| STIR 2^18, arity 4 / verify | 4 | both | 2.70 | 2.93 | +8.7% [-0.3%, +26.7%] — inconclusive |
| STIR 2^18, arity 4 / prove | 4 | compact | 76.66 | 74.33 | -3.0% [-6.0%, -0.0%] |
| STIR 2^18, arity 4 / verify | 4 | compact | 2.70 | 3.01 | +11.5% [+10.7%, +13.0%] |
| STIR 2^18, arity 4 / prove | 4 | final64 | 76.66 | 72.64 | -5.3% [-9.5%, -2.9%] |
| STIR 2^18, arity 4 / verify | 4 | final64 | 2.70 | 2.46 | -9.0% [-10.1%, -8.0%] |
| STIR 2^18, arity 8 / prove | 4 | both | 55.70 | 53.41 | -4.1% [-6.5%, -1.7%] |
| STIR 2^18, arity 8 / verify | 4 | both | 2.03 | 2.03 | -0.0% [-2.0%, +2.0%] — inconclusive |
| STIR 2^18, arity 8 / prove | 4 | compact | 55.70 | 55.86 | +0.3% [-1.1%, +1.6%] — inconclusive |
| STIR 2^18, arity 8 / verify | 4 | compact | 2.03 | 2.27 | +11.8% [+9.4%, +15.3%] |
| STIR 2^18, arity 8 / prove | 4 | final64 | 55.70 | 53.78 | -3.4% [-9.4%, +3.1%] — inconclusive |
| STIR 2^18, arity 8 / verify | 4 | final64 | 2.03 | 1.85 | -9.1% [-10.2%, -8.4%] |
| STIR 2^18, arity 4 / prove | Serial | both | 276.12 | 280.34 | +1.5% [-0.9%, +5.3%] — inconclusive |
| STIR 2^18, arity 4 / verify | Serial | both | 2.88 | 3.09 | +7.3% [-0.9%, +24.1%] — inconclusive |
| STIR 2^18, arity 4 / prove | Serial | compact | 276.12 | 294.03 | +6.5% [-3.5%, +23.2%] — inconclusive |
| STIR 2^18, arity 4 / verify | Serial | compact | 2.88 | 3.32 | +15.5% [+6.9%, +29.5%] |
| STIR 2^18, arity 4 / prove | Serial | final64 | 276.12 | 256.45 | -7.1% [-19.3%, +0.3%] — inconclusive |
| STIR 2^18, arity 4 / verify | Serial | final64 | 2.88 | 2.49 | -13.4% [-16.9%, -10.1%] |
| STIR 2^18, arity 8 / prove | Serial | both | 218.75 | 209.41 | -4.3% [-18.1%, +12.9%] — inconclusive |
| STIR 2^18, arity 8 / verify | Serial | both | 2.30 | 2.61 | +13.6% [+0.3%, +42.7%] |
| STIR 2^18, arity 8 / prove | Serial | compact | 218.75 | 213.18 | -2.5% [-12.8%, +7.5%] — inconclusive |
| STIR 2^18, arity 8 / verify | Serial | compact | 2.30 | 2.27 | -1.3% [-18.4%, +9.1%] — inconclusive |
| STIR 2^18, arity 8 / prove | Serial | final64 | 218.75 | 186.10 | -14.9% [-27.7%, -5.7%] |
| STIR 2^18, arity 8 / verify | Serial | final64 | 2.30 | 1.89 | -17.7% [-33.4%, -6.1%] |

</details>

The machine was OS-scheduled without CPU affinity or frequency locking; one-minute load varied from 4.08 to 33.30. All samples were retained. With only six default blocks or four option blocks, intervals are approximate and not adjusted for multiple comparisons. In particular, identical serial commitment controls measured -23.3% under the distinct-points case versus -0.2% under the merged case, so the larger serial PCS deltas should not be treated as precise optimization gains. `commit_ldes` shows no clear gain in this run.

Every generated proof was verified with matching prover/verifier next challenges. All 40 default main/branch artifact comparisons matched byte-for-byte (proofs, opened values, commitments and arithmetic outputs), plus 55 serial/parallel artifact comparisons.

## Validation

- `cargo test -p p3-stir --locked --offline`: 202 tests passed.
- `cargo test -p p3-stir --features parallel --locked --offline`: 202 tests passed.
- `cargo clippy -p p3-stir --all-targets --locked --offline -- -D warnings`: passed.
- `cargo clippy -p p3-stir --all-targets --features parallel --locked --offline -- -D warnings`: passed.
- `cargo +nightly fmt -p p3-stir -- --check`: passed.
- `git diff --check origin/main...HEAD`: passed.

### DIFF
--- a/stir/src/verifier.rs
+++ b/stir/src/verifier.rs
@@ -1,5 +1,6 @@
 //! STIR verifier implementation (Construction 5.2).
 
+use alloc::borrow::Cow;
 use alloc::vec::Vec;
 
 use itertools::izip;
@@ -14,7 +15,7 @@ use crate::config::{StirConfig, StirRoundConfig};
 use crate::error::{ExternalSourceError, GrindStage, ProofShapeError, RoundLabel, StirError};
 use crate::proof::{StirProof, StirQueryOpenings, StirRoundProof};
 use crate::utils::{
-    check_ans_interpolates, eval_poly, eval_poly_at_base, fold_domain_params,
+    check_ans_interpolates, eval_poly, eval_poly_at_base, fold_domain_params, interpolate_poly,
     lagrange_interpolate_at, next_domain_shift, reduce_mod_x_pow_minus_c, sample_ood_points,
     vanishing_poly_from_roots,
 };
@@ -147,6 +148,86 @@ fn check_ans_length<EF, MmcsError, InputError>(
         .into());
     }
     Ok(())
+}
+
+/// Resolve the explicitly configured encoding before the original transcript observation.
+/// Reconstruction uses the prover's canonical representation, including `[ZERO]` for a
+/// zero interpolant through a nonempty set of points.
+fn resolve_ans_polynomial<'a, EF: Field, MmcsError, InputError>(
+    round: RoundLabel,
+    compact: bool,
+    transmitted: &'a [EF],
+    points: &[EF],
+    values: &[EF],
+) -> Result<Cow<'a, [EF]>, StirError<MmcsError, InputError>> {
+    if !compact {
+        check_ans_length(round, transmitted, points.len())?;
+        return Ok(Cow::Borrowed(transmitted));
+    }
+    if !transmitted.is_empty() {
+        return Err(ProofShapeError::UnexpectedAnsPolynomial {
+            round,
+            got: transmitted.len(),
+        }
+        .into());
+    }
+    // interpolate_poly asserts these preconditions. Keep malformed inputs fallible here.
+    if points.len() != values.len()
+        || points
+            .iter()
+            .enumerate()
+            .any(|(i, point)| points[..i].contains(point))
+    {
+        return Err(StirError::InvalidAnsConsistency { round });
+    }
+    Ok(Cow::Owned(interpolate_poly(points, values)))
+}
+
+#[cfg(test)]
+mod compact_answer_tests {
+    use p3_baby_bear::BabyBear as F;
+    use p3_field::PrimeCharacteristicRing;
+
+    use super::*;
+
+    #[test]
+    fn compact_answers_reconstruct_canonical_zero_and_reject_invalid_nodes() {
+        let round = RoundLabel::Round(2);
+        assert!(
+            resolve_ans_polynomial::<F, (), ()>(round, true, &[], &[], &[])
+                .unwrap()
+                .is_empty()
+        );
+        assert_eq!(
+            resolve_ans_polynomial::<F, (), ()>(round, true, &[], &[F::ONE], &[F::ZERO])
+                .unwrap()
+                .as_ref(),
+            &[F::ZERO]
+        );
+        for (points, values) in [
+            (&[F::ONE][..], &[][..]),
+            (&[F::ONE, F::ONE][..], &[F::ZERO, F::ZERO][..]),
+        ] {
+            assert!(matches!(
+                resolve_ans_polynomial::<F, (), ()>(round, true, &[], points, values),
+                Err(StirError::InvalidAnsConsistency {
+                    round: RoundLabel::Round(2)
+                })
+            ));
+        }
+        // The default representation still borrows the provided coefficients verbatim.
+        assert!(matches!(
+            resolve_ans_polynomial::<F, (), ()>(
+                round,
+                false,
+                &[F::ZERO, F::ZERO],
+                &[F::ZERO, F::ONE],
+                &[F::ZERO, F::ZERO]
+            )
+            .unwrap(),
+            Cow::Borrowed([F::ZERO, F::ZERO])
+        ));
+    }
 }
 
 /// Fetch an external initial oracle's queried fibers and lay them out in draw order.
@@ -613,24 +694,27 @@ where
     // Step 4: ans polynomial observation and consistency check.
     let (all_points, all_values) = rv.all_points_and_values(&rp.ood_answers);
 
-    // Ans interpolates |all_points| values, bounding its length. That bound is what makes the
-    // one-point check below bind.
-    let max_ans_len = all_points.len();
-    check_ans_length(RoundLabel::Round(round), &rp.ans_polynomial, max_ans_len)?;
+    let ans = resolve_ans_polynomial(
+        RoundLabel::Round(round),
+        config.options().compact_answers,
+        &rp.ans_polynomial,
+        &all_points,
+        &all_values,
+    )?;
 
     // Bind ans_poly into the transcript BEFORE rho. The interpolation identity is a one-point
     // check; observing Ans first means the prover commits to it before learning rho.
-    challenger.observe_algebra_slice(&rp.ans_polynomial);
+    challenger.observe_algebra_slice(&ans);
 
     let rho: EF = challenger.sample_algebra_element();
 
-    if !check_ans_interpolates(&rp.ans_polynomial, &all_points, &all_values, rho) {
+    if !check_ans_interpolates(&ans, &all_points, &all_values, rho) {
         return Err(StirError::InvalidAnsConsistency {
             round: RoundLabel::Round(round),
         });
     }
 
-    Ok(rv.finish(rp.ans_polynomial.clone(), &all_points))
+    Ok(rv.finish(ans.into_owned(), &all_points))
 }
 
 /// One instance's in-flight final round, advanced in the order the transcript demands.
@@ -1306,19 +1390,24 @@ where
             let rp = &proofs[i].round_proofs[local_r];
             let (all_points, all_values) = rv.all_points_and_values(&rp.ood_answers);
 
-            let max_ans_len = all_points.len();
-            check_ans_length(RoundLabel::Round(local_r), &rp.ans_polynomial, max_ans_len)?;
+            let ans = resolve_ans_polynomial(
+                RoundLabel::Round(local_r),
+                configs[i].options().compact_answers,
+                &rp.ans_polynomial,
+                &all_points,
+                &all_values,
+            )?;
 
-            challenger.observe_algebra_slice(&rp.ans_polynomial);
+            challenger.observe_algebra_slice(&ans);
             let rho: EF = challenger.sample_algebra_element();
 
-            if !check_ans_interpolates(&rp.ans_polynomial, &all_points, &all_values, rho) {
+            if !check_ans_interpolates(&ans, &all_points, &all_values, rho) {
                 return Err(StirError::InvalidAnsConsistency {
                     round: RoundLabel::Round(local_r),
                 });
             }
 
-            finishes.push((i, rv.finish(rp.ans_polynomial.clone(), &all_points)));
+            finishes.push((i, rv.finish(ans.into_owned(), &all_points)));
         }
 
         // Phase 5: touches no transcript state, so instance order no longer matters.
--- a/stir/benches/stir.rs
+++ b/stir/benches/stir.rs
@@ -20,12 +20,13 @@ use p3_challenger::{
 use p3_commit::{ExtensionMmcs, Mmcs};
 use p3_dft::{Radix2DitParallel, TwoAdicSubgroupDft};
 use p3_field::extension::QuinticTrinomialExtensionField;
-use p3_field::{BasedVectorSpace, ExtensionField, Field, TwoAdicField};
+use p3_field::{BasedVectorSpace, ExtensionField, Field, PrimeCharacteristicRing, TwoAdicField};
 use p3_koala_bear::{KoalaBear, Poseidon2KoalaBear};
+use p3_matrix::dense::RowMajorMatrix;
 use p3_merkle_tree::MerkleTreeMmcs;
 use p3_stir::SecurityAssumption;
 use p3_stir::config::{StirConfig, StirParameters};
-use p3_stir::prover::prove_stir;
+use p3_stir::prover::{codeword_from_coeffs, prove_stir};
 use p3_stir::verifier::verify_stir;
 use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
 use rand::distr::{Distribution, StandardUniform};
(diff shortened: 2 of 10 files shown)


tcoratger (approved):
Reviewed this against Construction 5.2 of eprint 2024/390, on the assumption that a prover optimization plus a "send less data" mode is where a Fiat-Shamir binding quietly stops holding. **I found no soundness, transcript, or verifier-panic defect.**

The transcript order per intermediate round, prover and verifier byte-for-byte symmetric (`prover.rs:546-604` vs `verifier.rs:625-716`, multi at `prover.rs:1056-1170` vs `verifier.rs:1289-1408`):

```text
    grind(folding_pow) -> gamma -> fold, commit, observe(commitment)
      -> s OOD points -> observe(ood_answers) -> grind(pow) -> r_comb
      -> query indices -> Merkle-verify fibers -> observe(ans) -> rho
```

Every grind sits immediately before the challenge block it protects with no prover message in between, and in the final round the length check precedes the observation, so no attacker-sized vector reaches the sponge. Nothing in this PR moves a computation across an observe/sample boundary.

**`compact_answers` is transcript-neutral in the strong sense, and strictly sounder than the default.** The prover still absorbs the full `ans_poly` in both modes (`prover.rs:600`, `:1166`) — only the serialized field is emptied. The verifier rebuilds it with the same `interpolate_poly` over the same `(points, values)` in the same order, and trailing-zero stripping is identical on both sides, so the sponge input is bit-identical. Critically, `fetch_round_rows` runs `verify_multi_batch` *before* the rows reach `all_points_and_values`, so the reconstruction consumes commitment-bound data only, and that commitment was absorbed before the query indices were drawn. The mode also *removes* the `(n-1)/|F|` Schwartz-Zippel term the default prover had at the `rho` site while the config keeps budgeting it, so the accounting is conservative rather than short — no query-count or grinding adjustment needed.

Modes cannot be mixed: default -> compact is rejected by `UnexpectedAnsPolynomial`. compact -> default is rejected, but by transcript divergence rather than a shape check — see the inline note.

**Early-stop arithmetic re-derived by hand.** With `A = log_starting_degree - k0`, `extra_folds = ceil((A - cap)/k)`, `log_final_degree = A - extra_folds*k`: the `cap < A % k` guard is exactly what keeps `log_final_degree >= 0`, since at `cap = A % k` the numerator is `k*floor(A/k)` and `ceil` collapses to `floor`. All eight table cases check out. Soundness pricing is untouched — `final_eta`/`final_queries` come from `(log_degree_tail, log_inv_rate_tail)` and `(log_domain_tail - k) - log_final_degree = log_inv_rate_tail >= log_blowup >= 1`, so the direct-send check stays non-vacuous at the rate the paper's `delta_M` uses. `log_final_degree` never enters the query count, only the proof size. And since `total_folds` feeds `num_alg_terms` before any eta derivation, dropping rounds drops their error terms and shrinks the union-bound buffer consistently.

I also re-derived each prover shortcut rather than trusting the round trip:

- **Sub-coset recovery.** Each `fold_pass` squares the shift and halves the domain, and `two_adic_generator(n)^2 = two_adic_generator(n-1)`, so `folded[i*stride]` at `stride = 2^log_inv_rate` is exactly `fold_shift*<two_adic_generator(log2 D)>` in natural order — the same `D` coefficients the full-domain iDFT's prefix gave, because the tail is zero by the degree schedule. `stride >= 2` always, since `RateNotBelowOne` enforces `log_blowup >= 1`. `folded_codeword` is still retained in full for the round's answers, so no oracle is reused across rounds.
- **Degree-aware `codeword_from_coeffs`.** With `i = a + m*b`, `x_i = (shift*g^a)*(g^m)^b` and row `c` holding `coeff_c*(shift*g^a)^c`, the flat row-major output is natural order with no permutation — flat index `b*m + a = i` exactly. `dft_algebra_batch` returns an owned matrix via `to_row_major_matrix()`, so the bit-reversed `Evaluations` wrapper cannot leak through.
- **Adjusted weights.** `compute_adjusted_weights` is elementwise, so restricting `denoms` to the per-point max native height yields a vector whose every prefix is bit-identical, and the consumer already sliced `[..h]`. Provably a no-op. The new `unwrap()` cannot fire: `lde_views` and `matrix_layout` are both built from `placement`.

Verifier panics under a hostile proof: the only new surface is `resolve_ans_polynomial`, and it pre-checks both of `interpolate_poly`'s asserts and returns `InvalidAnsConsistency` instead. Neither can fire honestly — OOD points are sampled distinct and outside the fold domain, query points are deduplicated by index, so `batch_multiplicative_inverse` never sees a zero. Both lengths are verifier-derived, and the `O(n^2)` reconstruction is bounded by config rather than by the proof, so it is not a DoS lever.

Approving. Two one-liners I will send as a follow-up rather than hold this for, plus one test-hygiene item; the empty-`ans_polynomial` one is inline.

stir/src/verifier.rs
```diff
@@ -149,6 +150,86 @@ fn check_ans_length<EF, MmcsError, InputError>(
...
+        }
+        .into());
+    }
+    // interpolate_poly asserts these preconditions. Keep malformed inputs fallible here.
+    if points.len() != values.len()
+        || points
+            .iter()
+            .enumerate()
+            .any(|(i, point)| points[..i].contains(point))
+    {
```
tcoratger, line 180: One thing worth closing while you are here, since it is the only place the mode split leans on sponge divergence instead of a shape check.

The non-compact branch delegates to `check_ans_length`, which rejects only `len > max_ans_len`:

```rust
if ans_polynomial.len() > max_ans_len {
```

so an *empty* vector passes. A compact-mode proof therefore reaches `challenger.observe_algebra_slice(&[])` in a default-mode verifier and is rejected only downstream, because the compact prover absorbed the real coefficients and every later challenge diverges. The reverse direction is explicitly shape-checked (`UnexpectedAnsPolynomial`), so the guard is asymmetric.

Not a break: `interpolate_poly` returns at least `[ZERO]` whenever `points` is non-empty (`utils.rs:425`), and `num_ood_samples` in `{1,2}` plus `num_queries >= 1` makes `all_points` non-empty in every reachable round, so the honest prover never absorbs an empty slice and the divergence is unconditional. The sub-case I find most interesting is the all-zero witness (`kind == 1` in your new test): there `ans_poly = []` and `ans_poly = [ZERO]` give *identical* virtual oracles, so nothing but the diverged `final_gamma` rejects it.

Rejecting `transmitted.is_empty() && !points.is_empty()` here would make it loud, which is the convention the rest of the crate follows.

The matching test gap is in `compact_answers_preserve_proofs_and_transcripts` (`tests/stir.rs:232`): it only feeds a coefficient-carrying proof to a compact config (line 306), never the other way. One `verify_stir(&full, &compact_proof, ..).is_err()` closes it, ideally including the all-zero case.

Separately, and unrelated to soundness — `codeword_from_coeffs`'s new fast path can now issue a **height-1** DFT:

```rust
let log_len = log2_ceil_usize(coeffs.len().min(size).max(1));
```

At `coeffs.len() <= 1` this gives `log_len == 0`, so `eval_low_degree_on_coset` builds a height-1 matrix and calls `dft_algebra_batch`. The pre-existing caller always passes `log_len >= 1`, so that shape is new here. Unreachable from inside the crate (every internal path feeds >= 4 coefficients), but the function is `pub` and the new test covers only `Radix2DitParallel`; `Radix2DFTSmallBatch::dft_batch` computes `log2_strict_usize(estimate_num_rows_in_l1::<F>(1, w))` before it looks at `log_h`, which I could not convince myself is safe at height 1 without compiling. `.max(2)` makes the shape unreachable and stays exactly correct — a constant polynomial at `log_len = 1` still returns `c0` at every index, since row `c = 1` reads `polys[0].get(1) == None -> ZERO`.

PR #2057, opened 2026-09-08
perf(merkle-tree): optimize same-height multi-matrix hashing

Nashtare:
## Summary

Complements #2051 by optimizing the packed KeccakF/u64 path when several matrices share a leaf or injection layer.

The concatenated `flat_map` iterator inhibits vectorization when serializing pairs of 32-bit field elements into u64s. Staging the packed field elements first gives the hasher a simple contiguous iterator.

## Changes

- Add a default-off `CryptographicHasher::PREFER_CONTIGUOUS_INPUT` hint, enabled for packed-u64 serialization of 32-bit fields.
- Reuse scratch buffers within Rayon jobs for both leaf hashing and injected layers, retaining existing parallel granularity.
- Preserve concatenation across matrix boundaries before serialization, including partial u64s.
- Keep single-matrix hashing, scalar tails, and proof formats unchanged. Poseidon and Goldilocks remain unstaged.
- Add compatibility tests and benchmarks covering multiple matrix counts, uneven widths, and injected layers.

## Performance

Apple M4 Pro, Rust 1.98.0, eight Rayon workers. Full BabyBear MMCS commitments with 32,768 rows and two 128-column matrices:

| Path | Unstaged | Staged | Reduction |
|---|---:|---:|---:|
| Leaf | 4.659 ms | 3.034 ms | 34.9% |
| Injection | 5.858 ms | 4.259 ms | 27.3% |

Small-tree cases also improve. Separate clean-`main` comparisons detected no regression in single-matrix Keccak or Poseidon controls.

These are local commitment timings, not end-to-end prover speedup claims. Native x86 performance has not been measured.

Reproduce the representative cases with:

```sh
RAYON_NUM_THREADS=8 cargo bench -p p3-merkle-tree --features parallel --bench multi_matrix -- 'keccak_f/.*/(leaf|injection)/32768x128\+128$' --noplot
```

## Validation

- Pass the full workspace parallel test suite and doctests in the normal test profile, with debug symbols and incremental compilation disabled to limit disk usage.
- Compare complete digest layers, caps, opened rows, and authentication paths against the unstaged implementation, with verification and tamper-rejection checks.
- Cover odd widths/heights, scalar remainders, mixed-height injection, binary/quaternary trees, caps, and salted commitments.
- Pass 129 serial tests, 400 parallel tests, and six release-mode Keccak-backed proof examples.
- Pass workspace-wide Clippy, formatting, x86 default/AVX2/AVX-512 compilation, and WebAssembly library checks.
- Pass 275 tests when combined with #2051 at `851b124828b8f83358a10139e921a69eed21ba7c`; the combined benchmark retains the speedup.

The #2051 integration was checked in a separate worktree. This PR does not include its changes.

The workspace-wide `--release` run stops on two pre-existing `p3-lookup` tests, `generate_permutation_exclusive_rejects_non_boolean_flag` and `generate_permutation_exclusive_rejects_two_active_flags`. Both failures were reproduced on unmodified `main`: their expected panics come from `debug_assert!` checks disabled in release builds. No lookup changes are included here.

### DIFF
--- a/merkle-tree/Cargo.toml
+++ b/merkle-tree/Cargo.toml
@@ -30,6 +30,7 @@ tracing.workspace = true
 [dev-dependencies]
 p3-baby-bear = { path = "../baby-bear" }
 p3-blake3 = { path = "../blake3" }
+p3-goldilocks = { path = "../goldilocks" }
 p3-keccak = { path = "../keccak" }
 p3-mds = { path = "../mds" }
 p3-rescue = { path = "../rescue" }
@@ -43,6 +44,10 @@ harness = false
 
 [[bench]]
 name = "pruned_verify"
+harness = false
+
+[[bench]]
+name = "multi_matrix"
 harness = false
 
 [lints]
--- a/merkle-tree/benches/multi_matrix.rs
+++ b/merkle-tree/benches/multi_matrix.rs
@@ -1,0 +1,140 @@
+//! Compare equal-width commitments split across matrices, including injected layers.
+//! `current` uses the hasher's staging hint; `unstaged` retains the flat iterator.
+//! Input matrices are generated once and borrowed during each timed commitment.
+//!
+//! Run with a fixed worker count, for example:
+//! `RAYON_NUM_THREADS=8 cargo bench -p p3-merkle-tree --features parallel --bench multi_matrix`
+
+use std::hint::black_box;
+use std::time::Duration;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use p3_baby_bear::{BabyBear as F, Poseidon2BabyBear};
+use p3_commit::Mmcs;
+use p3_field::{Field, PackedValue};
+use p3_goldilocks::Goldilocks;
+use p3_keccak::{KeccakF, VECTOR_LEN};
+use p3_matrix::Matrix;
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_symmetric::{
+    CompressionFunctionFromHasher, CryptographicHasher, PaddingFreeSponge,
+    PseudoCompressionFunction, SerializingHasher, TruncatedPermutation,
+};
+use rand::SeedableRng;
+use rand::distr::{Distribution, StandardUniform};
+use rand::rngs::SmallRng;
+use serde::Serialize;
+use serde::de::DeserializeOwned;
+
+#[derive(Clone)]
+struct Unstaged<H>(H);
+
+impl<T: Clone, Out, H: CryptographicHasher<T, Out>> CryptographicHasher<T, Out> for Unstaged<H> {
+    fn hash_iter<I: IntoIterator<Item = T>>(&self, input: I) -> Out {
+        self.0.hash_iter(input)
+    }
+}
+
+fn split<F: Field>(matrix: &RowMajorMatrix<F>, widths: &[usize]) -> Vec<RowMajorMatrix<F>> {
+    let mut start = 0;
+    widths
+        .iter()
+        .map(|&width| {
+            let mut values = Vec::with_capacity(matrix.height() * width);
+            for row in matrix.values.chunks_exact(matrix.width()) {
+                values.extend_from_slice(&row[start..start + width]);
+            }
+            start += width;
+            RowMajorMatrix::new(values, width)
+        })
+        .collect()
+}
+
+fn bench_hash<F, P, PW, H, C, const D: usize>(criterion: &mut Criterion, name: &str, h: H, c: C)
+where
+    F: Field + Serialize + DeserializeOwned,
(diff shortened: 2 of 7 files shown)


tcoratger (approved):
Reviewed this as a commitment change until proven otherwise, since anything that alters what bytes reach a leaf alters the root, and two distinct openings hashing to the same leaf is a binding break.

**The leaf preimage is unchanged in every shape I could construct.** The staged arm and the pre-existing arm produce the same item sequence — `for m in matrices { scratch.extend(m.vertically_packed_row(first_row)) }` walks the slice in the same order as the `flat_map` it replaces, and `vertically_packed_row` is a deterministic `(0..width).map(|c| P::from_fn(|i| rows[i][c]))`. The sequence being identical is sufficient because the whole downstream chain is a pure function of it, which I checked link by link: `SerializingHasher::hash_iter` -> `into_parallel_u64_streams` (`field/src/integers.rs:527`, pulls `next()` twice per word, no `size_hint`, no `len`) -> `PaddingFreeSponge::hash_iter` (`sponge.rs:181`, absorbs purely via `next()`).

Worked concretely for BabyBear, widths `(3, 5)`, `M = 4`:

```text
elements:  a0 a1 a2 | b0 b1 b2 b3 b4
u64 words: [a0|a1] [a2|b0] [b1|b2] [b3|b4]     <- word 1 straddles the A/B boundary
```

Both arms produce exactly that, because both concatenate before serializing. So the PR body's "including partial u64s" is accurate. Shapes checked by hand: one matrix, several at equal height, unequal heights, zero-width, single row, odd row count. No fixtures need regenerating because nothing moved.

On the concatenation ambiguity I went looking for: it is real at the leaf-hash level — widths `(2,3)` and `(3,2)` at the same height flatten to the same preimage `x0 x1 x2 x3 x4` and share a root — but it is pre-existing, already documented, and already closed one level up. `check_widths` (`mmcs/geometry.rs:24`) rejects any opening whose per-row length differs from the verifier's `Dimensions`, and `batch.rs:184` says exactly why: "row boundaries are only authenticated by checking each row width". So the layout is pinned by the verifier's public dimension vector, not by prover discipline. This PR changes none of it.

Grouping order is deterministic on both sides: prover sorts `Reverse(l.height())` via itertools' `sorted_by_key`, which is `Vec::sort_by_key` and therefore stable, so within a height the order is insertion order; verifier does the same stable sort over its own public `dimensions`. No `HashMap` anywhere, and nothing attacker-selectable. `hash_packed_row` receives an already-ordered slice.

The unequal-height path is untouched, and there is no partial fast path to worry about — `hash_packed_row` is only ever called with a set that is already all-equal-height. The prover groups injections by `next_power_of_two()` while the verifier groups by exact height, which could in principle disagree (33 and 40 share a `next_power_of_two` of 64), but `validate_commit_reachable_heights` pins every height to exactly `ceil(max_height >> k)`, so at most one height per bucket survives and the two coincide. No new `unsafe`, no new indexing, no new panic surface.

I also checked the interaction with #2051, which merged while this was open. The two are complementary rather than competing: `first_digest_layer` now dispatches to #2051's batched driver only when `LANES > 1`, and #2051 deliberately leaves `LANES = 1` on the packed `[[uX; M]; N]` shapes with a comment saying so. That splits the two real Keccak configs cleanly — `examples/src/types.rs:28` has a `[u64; 4]` digest and stays on the packed arm, so your optimization survives the rebase; `binary-pcs/src/test_util.rs:24` has `[u8; 32]` and goes to the batched arm, where `hash_packed_row` never runs. And a leaf preimage cannot be split across two `hash_many` calls even inside the batched arm, since `rows_per_call` is a whole number of rows. `grep` also confirms this branch never calls `hash_many`/`compress_many`, so the ragged-batch assert #2051 added is no exposure.

Approving on soundness. Three things for the rebase, and a small follow-up I will send:

- After rebasing, `H::PREFER_CONTIGUOUS_INPUT` resolves unambiguously only inside `hash_packed_row`, which carries one `CryptographicHasher` bound. If the resolution hoists the check into `first_digest_layer` or `compress_and_inject` — each now carrying two such bounds — it needs the qualified form, and picking the wrong instantiation will compile and silently select the wrong arm.
- `Unstaged<H>` forwards only `hash_iter`, so post-rebase it silently resets `LANES` to 1 and `hash_many` to the default. Harmless for the `[u64; 4]` configs the tests use, but that is a coincidence of the config, and in the bench it would strip #2051's batching from the reference and overstate the delta. Worth forwarding the whole trait and overriding only the hint.
- The staging buffer has no size cap, where #2051 explicitly capped its equivalent at 8 KiB with a bail-out. At your widest bench case (2633 cols) the staged buffer is 21 KB at `VECTOR_LEN = 2` — fine on the M4 Pro the numbers come from — but 84 KB on AVX-512, an L2 round trip per leaf. Since the PR body already says native x86 is unmeasured, that is the corner to check.

PR #2059, opened 2026-09-08
refactor(commit): separate core PCS and STARK capabilities

Nashtare:
`Pcs` requires STARK-specific operations even for callers that only commit, open, and verify polynomials. Split those capabilities into `UnivariateStarkPcs`, leaving a core PCS interface that a standalone implementation can satisfy without STARK stubs.

Use named opening requests and claims in place of nested input tuples, move commitment positions into STARK-owned `StarkOpeningLayout`, and pass the preprocessing request index explicitly to hiding PCS code. Migrate all five univariate backends and workspace callers. Public Rust APIs change; `commit/README.md` documents the migration. Proof serialization structures and transcript traversal are preserved.

Validation:

- 470 impacted tests passed, including existing serialized-proof compatibility coverage.
- New core-only PCS regression and hiding-FRI regression with preprocessing at request index zero.
- Workspace all-target check and warnings-denied Clippy passed; no-default-feature checks and strict impacted-crate docs passed.

### DIFF
--- a/batch-stark/src/common.rs
+++ b/batch-stark/src/common.rs
@@ -12,7 +12,7 @@ use alloc::vec::Vec;
 
 use p3_air::Air;
 use p3_air::symbolic::{AirLayout, SymbolicExpressionExt};
-use p3_commit::{Pcs, PolynomialSpace};
+use p3_commit::{Pcs, PolynomialSpace, UnivariateStarkPcs};
 use p3_field::{Algebra, BasedVectorSpace};
 use p3_lookup::{InteractionSymbolicBuilder, LogUpGadget, Lookups};
 use p3_matrix::Matrix;
--- a/batch-stark/src/prover.rs
+++ b/batch-stark/src/prover.rs
@@ -9,7 +9,7 @@ use p3_air::DebugConstraintBuilder;
 use p3_air::symbolic::{AirLayout, SymbolicExpressionExt};
 use p3_air::{Air, RowWindow};
 use p3_challenger::GrindingChallenger;
-use p3_commit::{Pcs, PolynomialSpace};
+use p3_commit::{Pcs, PolynomialSpace, UnivariateStarkPcs};
 use p3_field::{
     Algebra, BasedVectorSpace, PackedFieldExtension, PackedValue, PrimeCharacteristicRing,
     PrimeField,
@@ -500,6 +500,7 @@ where
         transcript.grind_and_sample_zeta(config.ood_proof_of_work_bits());
 
     // Build the opening rounds and produce the FRI opening proof.
+    let opening_layout = p3_uni_stark::StarkOpeningLayout::new(SC::Pcs::ZK);
     let (opened_values, opening_proof) = {
         let mut rounds = Vec::new();
 
@@ -583,9 +584,12 @@ where
         }
 
         pcs.open_with_preprocessing(
-            rounds,
+            rounds.into_iter().map(Into::into).collect(),
             &mut transcript.challenger,
-            common.preprocessed.is_some(),
+            common
+                .preprocessed
+                .as_ref()
+                .map(|_| opening_layout.preprocessed),
         )
     };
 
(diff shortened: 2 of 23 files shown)


tcoratger (approved):
Read all 23 files, and checked the split method by method rather than trusting the trait moves — a refactor this size is supposed to be behaviour-preserving, which is exactly what makes a behavioural change easy to miss.

It is faithful. Every moved body is textually identical, every assertion survives, no bound loosened, and the transcript absorb order is byte-for-byte the same at every call site.

The one thing I looked hardest at was newly-provided default implementations, since that is where an implementor that was forced to supply a correct method silently inherits a generic one. There are none: `EvaluationsOnDomain`, `ZK`, `log_max_lde_height`, `get_quotient_ldes`, `commit_ldes` and `get_evaluations_on_domain` all stay *required* on `UnivariateStarkPcs`, and every method that was defaulted before is defaulted the same way now, with every prior override preserved.

Bounds went the safe direction. I diffed the `where` clauses of all five `Pcs` impls against the base and against the new `UnivariateStarkPcs` impls — byte-identical in every case — and `StarkGenericConfig::Pcs` actually *tightened* from `Pcs` to `UnivariateStarkPcs`.

The only genuine semantic change is `open_with_preprocessing`'s last argument, `bool` -> `Option<usize>`, and it is provably equivalent for every in-tree caller. `HidingFriPcs` has `ZK = true`, so the old `PREPROCESSED_TRACE_IDX` was `1+1+1 = 3`; uni-stark pushes `[random(0), trace(1), quotient(2), preprocessed(3)]` and now passes `Some(3)`, batch-stark pushes five rounds with preprocessed at 3 and passes `Some(3)`, and `Pcs::open` passes `None`, which matches `is_preprocessing = false`. The `assert!(!Self::ZK)` guard on the non-hiding default is retained verbatim. That the serialized-proof fixtures were not regenerated is independent evidence the transcript and proof bytes are unchanged.

I also swept the 1130 deleted lines for dropped guards. Every `assert!`/`debug_assert!`/`Err(` has a surviving counterpart: circle's `log_n >= 2` and the mats/points `debug_assert_eq!`, FRI's committed-LDE-height check, STIR's three asserts, hiding's `HidingRandomOpening*CountMismatch` at all three levels, and `MatrixWithoutOpeningPoints` in both FRI and circle including the "width pinned by the claim, never by the proof" logic. `log_max_lde_height` is still reachable at both two-adicity guards in each verifier, so the "domain too tall for the field's two-adicity" rejection is intact.

Two things I will send as a small follow-up rather than hold this up for:

- `StarkOpeningLayout`'s three `pub usize` fields make an inconsistent layout representable where the old associated consts could not disagree, and `open_with_preprocessing` validates nothing about the index. It is not reachable from any in-tree caller, but the failure is quiet on the hiding path: pass `Some(1)` and `hiding_pcs.rs:437` sets `num_random_codewords = 0` for that round, so the mask openings stay in `inner_opened_values` and ship as public opened values, while the real preprocessed round gets real columns stripped instead. `verify` re-merges both halves and passes. Private fields plus `const fn` accessors and a `debug_assert!(idx < requests.len())` closes it.
- The new claim structs derive only `Clone, Debug`, dropping the structural `PartialEq`/`Eq` the tuples had, so a downstream `assert_eq!` on built claims stops compiling with no migration note.

Approving.

tcoratger:
Flagging a caveat on my own approval, since #2035 landed after I reviewed this and turned it `CONFLICTING`.

The overlap is four files, all in `fri/`:

```text
    fri/src/hiding_pcs.rs
    fri/src/prover.rs
    fri/src/two_adic_pcs.rs
    fri/src/verifier.rs
```

Those are precisely the files the load-bearing half of my review rested on:

- the `bool -> Option<usize>` equivalence argument for `open_with_preprocessing`, which I traced through `hiding_pcs.rs:437` and the `num_random_codewords` split;
- "the transcript absorb order is byte-for-byte the same at every call site", checked in `fri/src/prover.rs` and `verifier.rs`;
- the surviving guards I enumerated — FRI's committed-LDE-height assert, and `MatrixWithoutOpeningPoints` including the "width pinned by the claim, never by the proof" logic.

#2035 drives FRI's transcript through the new typed Fiat-Shamir layer, so the observe/sample sequence in those two files is now expressed through `fri/src/transcript.rs` rather than inline. My ordering check was against the pre-#2035 shape, and it does not automatically carry across the rebase.

So: the trait split itself I still consider verified — the method-by-method table, the bounds (which tightened), the absence of any newly-provided default, and the swept `assert!`/`Err(` set are all independent of how FRI plays its transcript. But the transcript-order claim needs re-establishing on the rebased tree, and it is the one where a plausible conflict resolution can compile and still be wrong, because reordering two `observe` calls is invisible to the type system.

Happy to re-check just that once you have pushed the rebase — it is a narrow read of the four files, not another full pass.

One other thing the rebase makes newly relevant, from #2035: `ProverState::drop` now yields to a panic already in flight (`challenger/src/fs/mod.rs`, gated on `cfg(panic = "unwind")`). Nothing in this PR depends on the old abort behaviour as far as I can see, but if any of your test helpers relied on a `ProverState` being dropped unfinalized to signal failure, that path is quieter now.

PR #2060, opened 2026-09-08
refactor(lookup): remove univariate STARK dependency

Nashtare:
Shared lookup generation currently depends on an entire univariate STARK configuration, pulling `p3-uni-stark` into consumers such as multilinear proving. Parameterize lookup trace generation directly by base and extension fields, and move the univariate constraint-folder adapters into `p3_batch_stark::folder`.

Remove the normal `p3-uni-stark` dependency and unused configuration-fixture dependencies, migrate callers/tests/benchmarks, and document the public API changes in `lookup/README.md`. Lookup arithmetic, challenge order, accumulator layout, and serialization are preserved.

Validation:

- Lookup: 55 tests passed; batch-STARK: 61 passed; multilinear STARK: 105 passed.
- Default/parallel all-target checks, warnings-denied Clippy, and docs passed for the affected crates.
- Cargo dependency graphs confirm that lookup and multilinear STARK no longer include `p3-uni-stark` through lookup.

### DIFF
--- a/lookup/src/logup.rs
+++ b/lookup/src/logup.rs
@@ -22,12 +22,11 @@ use alloc::vec::Vec;
 
 use itertools::Itertools;
 use p3_air::{ExtensionBuilder, PermutationAirBuilder, SymbolicExpression, WindowAccess};
-use p3_field::{Field, PrimeCharacteristicRing, dot_product};
+use p3_field::{ExtensionField, Field, PrimeCharacteristicRing, dot_product};
 use p3_matrix::Matrix;
 use p3_matrix::dense::{RowMajorMatrix, RowMajorMatrixView};
 use p3_matrix::stack::VerticalPair;
 use p3_maybe_rayon::prelude::*;
-use p3_uni_stark::{StarkGenericConfig, Val};
 use tracing::instrument;
 
 use crate::protocol::LookupProtocol;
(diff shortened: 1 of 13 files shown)


lookup/src/logup.rs
```diff
@@ -751,19 +747,20 @@ impl LookupProtocol for LogUpGadget {
 
                                     let mult = lookup.multiplicities[k].resolve(&row_ctx);
                                     numerator += flag * mult;
-                                    let term = alpha - combine_tuple::<SC>(elts, powers, &row_ctx);
-                                    denom += term * SC::Challenge::from(flag);
+                                    let term =
+                                        alpha - combine_tuple::<F, EF>(elts, powers, &row_ctx);
+                                    denom += term * EF::from(flag);
```
tcoratger: Do we need the EF::from here?

lookup/src/logup.rs
```diff
@@ -751,19 +747,20 @@ impl LookupProtocol for LogUpGadget {
...
                                 debug_assert!(
-                                    flag_sum.is_zero() || flag_sum == Val::<SC>::ONE,
+                                    flag_sum.is_zero() || flag_sum == F::ONE,
                                     "exclusive flags must sum to at most one per row"
                                 );
 
                                 // Fallback: an all-inactive row gets denominator 1.
                                 // Its fraction is then 0.
-                                denom += SC::Challenge::from(Val::<SC>::ONE - flag_sum);
+                                denom += EF::from(F::ONE - flag_sum);
```
tcoratger: Here maybe same we don't need to convert to EF no?

tcoratger (approved):

PR #2061, opened 2026-09-08
test: share PCS opening contracts and batch fixtures

Nashtare:
PCS backend tests repeat opening setup and do not share one check for claim ordering, independent expected evaluations, and transcript agreement. Add an opt-in `p3-commit::testing::assert_pcs_opening_contract` helper and use it with real FRI, Circle, and STIR backends.

The helper checks honest verification, commitment/matrix/point/column associations, prover/verifier transcript agreement, and rejection of altered values or missing claim elements. STIR matrices use distinct polynomials so equal-size matrix swaps remain detectable. Extract batch-STARK configuration fixtures into `tests/common/config.rs`, retaining backend-specific regressions and serialized-proof fixtures.

Depends on #2059. This PR targets `robin/refactor-pcs` so the diff contains only the shared tests and fixture extraction; retarget to `main` after the PCS refactor lands.

Validation:

- FRI/Circle PCS suite: 32 tests passed.
- STIR PCS suite: 40 tests passed.
- Batch integration suite: 53 tests passed; 2 fixture generators intentionally ignored.
- These suites passed with parallel enabled on the combined refactor checkout; targeted Clippy and strict docs also passed.
- Independent review and follow-up review completed with no outstanding findings.

### DIFF
--- a/commit/src/testing/pcs.rs
+++ b/commit/src/testing/pcs.rs
@@ -1,0 +1,172 @@
+//! Shared opening checks for transparent, batched univariate commitment backends.
+
+use alloc::vec::Vec;
+
+use p3_challenger::{CanObserve, FieldChallenger};
+use p3_field::ExtensionField;
+use p3_matrix::Matrix;
+use p3_matrix::dense::RowMajorMatrix;
+
+use crate::{
+    CommitmentOpening, MatrixOpening, OpeningRequest, Pcs, PointOpening, PolynomialSpace, Val,
+};
+
+/// Check a backend's batched opening values, verification, transcript and claim-shape handling.
+///
+/// The caller supplies evaluation matrices grouped by commitment, and an independent evaluator
+/// for their columns. The evaluator receives `(commitment_index, matrix_index, opening_point)`.
+/// It must compute the expected values from the intended polynomials, not from PCS output.
+///
+/// Two points are sampled after absorbing all commitments. Their order alternates between
+/// matrices, exercising the association of commitments, matrices, points and columns. After
+/// honest verification, the next prover and verifier challenge must agree. Fresh copies of the
+/// verifier transcript also check rejection of a changed value, a missing column and a missing
+/// matrix. Backend-specific proof mutations and serialization fixtures belong in the caller's
+/// tests.
+///
+/// Intended for transparent backends with a fixed matrix/column opening shape, such as FRI,
+/// Circle and STIR. Hiding backends may use a different preprocessing/opening protocol.
+///
+/// # Panics
+///
+/// Panics on any failed contract check. The fixture must contain at least one commitment,
+/// every commitment must contain a matrix, and every matrix must have at least one column
+/// and the height required by its domain. The supplied transcript must yield two distinct,
+/// valid opening points for the backend.
+#[allow(clippy::type_complexity)]
+pub fn assert_pcs_opening_contract<P, Challenge, Challenger>(
+    pcs: &P,
+    challenger: &Challenger,
+    rounds: &[Vec<(P::Domain, RowMajorMatrix<Val<P::Domain>>)>],
+    expected: impl Fn(usize, usize, Challenge) -> Vec<Challenge>,
+) where
+    P: Pcs<Challenge, Challenger>,
+    Challenge: ExtensionField<Val<P::Domain>>,
+    Challenger: Clone + CanObserve<P::Commitment> + FieldChallenger<Val<P::Domain>>,
+{
+    assert!(!rounds.is_empty(), "fixture needs a commitment");
+    for round in rounds {
+        assert!(!round.is_empty(), "fixture needs a matrix per commitment");
+        for (domain, matrix) in round {
+            assert_ne!(matrix.width(), 0, "fixture needs nonempty columns");
+            assert_eq!(domain.size(), matrix.height(), "fixture domain height");
+        }
+    }
+
+    let (commitments, prover_data): (Vec<_>, Vec<_>) = rounds
+        .iter()
+        .map(|round| pcs.commit(round.iter().cloned()))
+        .unzip();
+    let mut prover = challenger.clone();
+    prover.observe_slice(&commitments);
+    let points: [Challenge; 2] = [
+        prover.sample_algebra_element(),
+        prover.sample_algebra_element(),
+    ];
+    assert_ne!(
+        points[0], points[1],
+        "fixture needs distinct opening points"
+    );
+    let opening_points: Vec<Vec<Vec<Challenge>>> = rounds
+        .iter()
+        .enumerate()
+        .map(|(r, round)| {
+            (0..round.len())
+                .map(|m| {
+                    let mut points = points.to_vec();
+                    if (r + m) % 2 == 1 {
+                        points.reverse();
+                    }
+                    points
+                })
+                .collect()
+        })
+        .collect();
+    let (opened, proof) = pcs.open(
+        prover_data
+            .iter()
+            .zip(opening_points.clone())
+            .map(|(prover_data, points)| OpeningRequest {
+                prover_data,
+                points,
+            })
+            .collect(),
+        &mut prover,
+    );
+    assert_eq!(opened.len(), rounds.len(), "commitment opening count");
+
+    let claims: Vec<_> = commitments
+        .iter()
+        .cloned()
+        .enumerate()
+        .map(|(r, commitment)| {
+            assert_eq!(opened[r].len(), rounds[r].len(), "matrix opening count");
+            let matrices = rounds[r]
+                .iter()
+                .enumerate()
+                .map(|(m, (domain, matrix))| {
+                    assert_eq!(opened[r][m].len(), 2, "point opening count");
+                    let values = opening_points[r][m]
+                        .iter()
+                        .copied()
+                        .enumerate()
+                        .map(|(p, point)| {
+                            let values = expected(r, m, point);
+                            assert_eq!(values.len(), matrix.width(), "expected column count");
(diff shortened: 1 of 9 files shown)


fri/tests/pcs.rs
```diff
@@ -106,10 +107,52 @@ fn do_test_fri_pcs<Val, Challenge, Challenger, P>(
...
+// reorders a batch cannot accidentally satisfy the expected openings.
+fn shared_opening_contract<Val, Challenge, Challenger, P>((pcs, challenger): &(P, Challenger))
+where
+    P: Pcs<Challenge, Challenger>,
+    P::Domain: PolynomialSpace<Val = Val>,
+    Val: Field,
+    Challenge: ExtensionField<Val>,
+    Challenger: Clone + CanObserve<P::Commitment> + FieldChallenger<Val>,
+{
+    let constants = [vec![vec![2, 7], vec![11]], vec![vec![19, 23, 29]]];
```
tcoratger, line 123: This fixture makes every committed column a constant polynomial, which collapses the thing the contract exists to check.

`RowMajorMatrix::new(row.repeat(height), row.len())` repeats a single row, so `p_c[X] = k_c` for all `X`. Follow that into `fri/src/two_adic_pcs.rs:597`:

```rust
*ro += alpha_pow_offset * (reduced_openings - reduced_row) * inv_denom;
```

With the honest claim `y_c = k_c`, `reduced_openings - reduced_row == 0` for every row, every point, every matrix. So `fri_input` at line 606 is the all-zero vector, and the verifier recomputes the same zero from the opened rows and folds zeros against zeros.

The consequence is that `inv_denoms` — the only place the opening point enters the argument — is multiplied by zero. A wrong point, a swapped point, a missing `1/(z-x)`, or a wrong `alpha_pow_offset` all give `0 == 0`. That makes the doc claim at `commit/src/testing/pcs.rs:20` ("exercising the association of commitments, matrices, points and columns") false for all five macro-generated instantiations, FRI *and* Circle. What does survive is real but narrow: the `(commitment, matrix, column)` ordering of `opened`, MMCS plumbing over two commitments at mixed heights, and transcript agreement.

`shared_contract_preserves_opening_point_order` is the only point-dependent caller, and it covers `get_pcs(1)` alone — not blowup 2, not high arity, not Circle. Its three matrices are also identical to each other, so it exercises no matrix association. No single test does both.

I assume the constants are there because the generic fixture has to work for `CircleDomain` too. `PolynomialSpace::{first_point, next_point}` are implemented for `CircleDomain` in standard position (`circle/src/domain.rs:162`), which is what `natural_domain_for_degree` returns, so the construction `shared_contract_preserves_opening_point_order` already uses can be lifted into the generic helper: column `j` of matrix `(r, m)` as `x^j + c_{r,m,j}`, keeping the existing distinct constants. Worth confirming that interpolates as expected over the circle basis before committing to it.

Either way, the harness should refuse to give credit it has not earned — see my note on the helper.

commit/src/testing/pcs.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut prover = challenger.clone();
+    prover.observe_slice(&commitments);
+    let points: [Challenge; 2] = [
+        prover.sample_algebra_element(),
+        prover.sample_algebra_element(),
+    ];
+    assert_ne!(
+        points[0], points[1],
+        "fixture needs distinct opening points"
+    );
```
tcoratger, line 75: Suggest a vacuity guard right here, so no future backend can silently inherit the constant-fixture problem:

```rust
assert!(
    rounds.iter().enumerate().any(|(r, round)| (0..round.len())
        .any(|m| expected(r, m, points[0]) != expected(r, m, points[1]))),
    "fixture must contain a point-dependent column: constant columns make the \
     reduced opening identically zero and the point association untested",
);
```

You already assert `points[0] != points[1]` on the line above, which is the necessary condition; this is the sufficient one. Adding it should fail all five FRI/Circle instantiations today, which is the useful demonstration that the coverage was not there.

commit/src/testing/pcs.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut opening_transcript = verifier.clone();
+    pcs.verify(claims.clone(), &proof, &mut verifier)
+        .expect("honest opening must verify");
+    assert_eq!(
+        prover.sample_algebra_element::<Challenge>(),
+        verifier.sample_algebra_element::<Challenge>(),
+        "prover and verifier transcript after opening"
+    );
+
+    let mut wrong_value = claims.clone();
```
tcoratger, line 177: All three negative mutations are satisfied by transcript divergence alone, so `.is_err()` proves less than the assertion messages claim.

Both `TwoAdicFriPcs::verify` (`fri/src/two_adic_pcs.rs:648`) and `TwoAdicStirPcs::verify` (`stir/src/pcs.rs:1126`) absorb every claimed value before any FRI/STIR work:

```rust
for PointOpening { values: point, .. } in mat {
    challenger.observe_algebra_slice(point);
}
```

So `wrong_value`, `missing_column` and `missing_matrix` each perturb the transcript, `alpha`/`beta`/the query indices all diverge, and rejection is guaranteed before the opening-consistency logic is reached. The tests are rejecting for a different reason than the one they name.

Neither backend observes `PointOpening::point`, so there is exactly one transcript-neutral mutation available — swap the two `point` fields of one matrix's openings while leaving `values` alone. That claim is genuinely false and can only be caught by the reduced-opening denominators, i.e. the assertion the contract is currently missing:

```rust
let mut swapped = claims.clone();
let ps = &mut swapped[0].matrices[0].points;
let (p0, p1) = (ps[0].point, ps[1].point);
ps[0].point = p1;
ps[1].point = p0;
assert!(pcs.verify(swapped, &proof, &mut opening_transcript.clone()).is_err());
```

This needs the point-dependent fixture to mean anything — with constant columns the swapped claim is still true — so it lands naturally with that change.

Two smaller things in the same area: `wrong_value` only perturbs `[0][0][0][0]`, so a verifier ignoring the second point, the last matrix or the last column would pass (one extra `verify` on the last position fixes it); and `missing_matrix` pops to an empty `matrices` vec whenever `rounds[0].len() == 1`, which is the case for `do_test_pcs(&[8])`, `(&[4])` and `(&[2])`. That takes a different path — STIR's per-matrix shape loop at `stir/src/pcs.rs:1103` iterates zero times and rejection comes from a downstream root-count mismatch — so the intended check only really runs for the multi-matrix fixtures. An `assert!(rounds[0].len() >= 2)` would make the requirement explicit. Same fixture shape means the `(r + m) % 2 == 1` point-reversal branch at line 77 is never entered for those three tests either.

stir/tests/stir.rs
```diff
@@ -1846,60 +1846,38 @@ mod babybear_pcs {
...
             .iter()
-            .map(|&log_d| {
-                let d = 1 << log_d;
-                let width = 3;
-                (
-                    <MyPcs as Pcs<Challenge, Challenger>>::natural_domain_for_degree(&pcs, d),
-                    RowMajorMatrix::<Val>::rand(&mut rng, d, width),
-                )
+            .enumerate()
+            .map(|(matrix_index, &log_degree)| {
```
tcoratger, line 2066: This trades full-degree inputs for degree-2 ones. The old fixture was `RowMajorMatrix::rand(&mut rng, d, 3)`, whose interpolant has degree `d - 1`; the new columns are `X^2 + c`, `X + c`, `c`, degree <= 2 regardless of `d`. For `do_test_pcs(&[8])` that is degree 2 against a bound of 255.

Any claimed degree bound >= 2 then accepts, so degree-bound off-by-ones and error classes that *raise* the degree slightly become invisible — multiplying by `(x - z)` instead of dividing gives degree 3 against a bound of 254, and passes. `round_trip_under` still uses random full-degree inputs and compensates for the interleaved-shape tests, but not for the single-matrix commitments `(&[8])`, `(&[4])`, `(&[2])`, and `test_pcs_single_degree2_no_intermediate_rounds` exists precisely to pin that isolated schedule.

A fourth column `x.exp_u64((domain.size() - 1) as u64)` with `expected = point.exp_u64((d - 1) as u64)` restores full degree while staying independently computable, and is O(d log d) for d <= 256.

tcoratger (changes requested):
Reviewed this for lost coverage first, since a shared harness that is weaker than the per-crate tests it replaces removes coverage from several crates at once with nothing going red. **On that axis it is clean** — I built the ledger and no coverage is lost anywhere:

- The `batch-stark` config move is byte-identical (verified by normalized diff modulo imports and `pub(super)`), and the fixture bytes are unchanged, so this is not a wire-format change wearing test clothes.
- FRI's existing `do_test_fri_pcs` suite is untouched; the contract is purely additive there.
- STIR keeps its whole 34-test specific-rejection suite (`test_tampered_*`, `test_permuted_row_evals_rejected`, the `ProofShapeError::{CommitmentRootCount,RoundCount}` equality assertions) and its random full-degree `round_trip_under` paths.
- No negative test asserting a *specific* rejection reason was downgraded to a generic "it errors" — the generic `.is_err()` assertions are all new.
- Nothing in the diff asserts a thread count, a task size, or a SIMD lane width, which is the class of test that has bitten us before.

The problem runs the other way: the shared contract asserts substantially less than its doc comment claims, and in its principal instantiation it is close to vacuous. Details inline.

Noting the stacking: this targets `robin/refactor-pcs`, not `main`. `assert_pcs_opening_contract` is a public `test-utils` API that future backends will copy verbatim, so it is worth getting the fixture right before this reaches `main` rather than after three backends have inherited it.

Nashtare:
@tcoratger lmk if you need anything else here :)

tcoratger (approved):

PR #2062, opened 2026-09-08
chore: share local and CI check commands

Nashtare:
Local checks currently require copying commands from CI, and embedded coverage relies on a manually maintained crate list that omits binary PCS. Add a Python standard-library command runner shared by local development and the existing workflows.

`scripts/check.py` provides focused/full host checks, tests, lint/docs, architecture builds, embedded builds, wasm checks, benchmark smoke, and heavy-test recipes. Discover embedded libraries from Cargo metadata with explicit host-only exclusions; support package metadata for tests that need multiple features enabled together. Preserve existing CI jobs, matrix entries, target flags, and runtime versus compile-only coverage. Add contributor instructions, a crate map, and a backend capability table.

Validation:

- 13 command tests passed, covering real temporary Cargo workspaces, process failure propagation, feature unions, and an empty targeted nextest baseline followed by feature-gated tests.
- Shared fast/lint/docs commands passed on the combined refactor checkout.
- Embedded runner built all 46 libraries selected in that checkout for `thumbv7em-none-eabi`, including binary PCS.
- Architecture/wasm/heavy/benchmark command dispatch and workflow mapping reviewed; those complete platform-specific suites were not rerun locally.

### DIFF
--- a/.github/workflows/bench-smoke.yml
+++ b/.github/workflows/bench-smoke.yml
@@ -29,4 +29,4 @@ jobs:
       # field x algorithm x size sweep produces thousands of benchmark ids that
       # make even a single pass too slow.
       - name: Run each benchmark once
-        run: cargo bench --workspace --exclude p3-dft --features parallel -- --test
+        run: python3 scripts/check.py bench
--- a/.github/workflows/ci-heavy.yml
+++ b/.github/workflows/ci-heavy.yml
@@ -45,10 +45,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Test WHIR exhaustive sweep
-        run: cargo test -p p3-whir pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact
+        run: python3 scripts/check.py whir-exhaustive
 
       - name: Test WHIR exhaustive sweep with parallel
-        run: cargo test -p p3-whir --features parallel pcs::tests::test_whir_end_to_end_exhaustive -- --ignored --exact
+        run: python3 scripts/check.py whir-exhaustive --parallel
 
   binary_pcs_large_configuration:
     name: binary-pcs 2^16 round trip
@@ -68,4 +68,4 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Test the 2^16 round trip
-        run: cargo test -p p3-binary-pcs --release --test end_to_end large_configuration_2_16_round_trips -- --ignored --exact
+        run: python3 scripts/check.py binary-large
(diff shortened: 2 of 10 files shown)


tcoratger (approved):
Reviewed the check ledger rather than the diff shape, since the risk in this PR is a check that used to run and quietly stops.

I rebuilt it leg by leg from `ci.yml`, `ci-heavy.yml` and `bench-smoke.yml` on both sides and compared against the argv `scripts/check.py` actually emits. Everything survives byte-identically: default and `--features parallel`; the `+avx2` and `+pclmulqdq` runtime legs; the `+avx512f`, `+avx2,+vpclmulqdq`, `+avx512f,+vpclmulqdq`, `+sve2` compile-only legs; the wasm32 `simd128` leg; the macOS `-C target-feature=-sha3` keccak leg; clippy `-D warnings`; and `cargo +nightly fmt --all -- --check`.

Two I checked specifically because they are the usual casualties:

- **`cargo test --doc` in both the plain and `--parallel` legs.** nextest does not run doctests, so a unification like this is exactly where they get dropped — we lost a day to a broken doctest on #2035 earlier today. Both survive, and I confirmed the "Test doctests" step actually executed and succeeded on `Build and Test (macos-latest, false)`.
- **The required-status-check trap the matrix comment warns about.** `jobs.build_and_test.name` is unchanged and the `strategy.matrix.include` block is character-for-character identical, so no rendered name moves. Diffed the job-name lists from a `main` run against run 34226816710: 21 lines each, no output. Branch protection needs no change.

The embedded set is a strict superset — 45 crates against 44 — and the one it gains is `p3-binary-pcs`, which the hand-maintained list had silently omitted. That is the bug this PR exists to prevent, showing up in the diff.

Supply chain is clean: no new actions, no version bumps, no `curl | sh` added, no `permissions:` widening, no `pull_request_target`, and the only new `run:` interpolations are workflow-defined (`runner.os`, `matrix.target`) rather than derived from any `github.event.*` field, so there is no injection sink.

One gap worth naming, not worth blocking on: `check.py` never sets `-C target-feature`, so `check.py architecture --target x86_64-unknown-linux-gnu` compiles with the baseline feature set and every `cfg(target_feature = "avx512f")` block is `#[cfg]`-ed out. CONTRIBUTING.md:41 says that command "uses `architecture` for AVX-512, VPCLMULQDQ, and SVE2 coverage", so a contributor gets a green that checked none of it. CI is unaffected — the `Set flags` step still does the real work — it is only the shared local command that is weaker than the leg it names. I will send a follow-up adding `--target-feature` and threading `${{ matrix.features }}` through, which also lets `Set flags` go away.

Approving. Nothing here lowers the bar for future PRs, which was the thing to get right.

PR #2064, opened 2026-09-08
feat(security): integrate legacy conjectured regime

Nashtare:
The legacy ethSTARK FRI formula added in #2018 was only available as a standalone helper. Integrate it through `Regime::Legacy`, optional `LowDegreeTest` support, scalar/report APIs, and uni-STARK's `LegacySecurity` and `Proof::legacy_security`. Include the legacy level in example output.

Legacy reports use `num_queries * log_blowup + query_pow_bits` and omit the FRI folding term. They retain AIR/DEEP, batching, extras, grinding outside the LDT, and the commitment-collision cap. LDTs without explicit legacy support return `None`.

Validation:

- 168 tests passed with `cargo test -p p3-security -p p3-uni-stark --offline`, including legacy regression and composite-accounting coverage.
- Clippy passed for security, uni-STARK, and examples with `--all-targets -- -D warnings`.
- Workspace formatting and rustdoc checks with broken intra-doc links denied passed.

### DIFF
--- a/security/src/deep.rs
+++ b/security/src/deep.rs
@@ -25,9 +25,14 @@ use libm::log2;
 use crate::error::ErrorBits;
 use crate::shape::{InstanceShape, StarkAirParams};
 
-/// `-log2(ε_DEEP)` in bits. Returns 0 bits if inputs are degenerate.
+/// `-log2(ε_DEEP)` in bits. Returns 0 bits if inputs are degenerate or the trace size
+/// cannot be represented by `u64`.
 pub fn deep_ali_error(air: &StarkAirParams, shape: &InstanceShape, list_size: f64) -> ErrorBits {
-    if shape.modulus_bits == 0 || !list_size.is_finite() || list_size <= 0.0 {
+    if shape.modulus_bits == 0
+        || shape.log_trace_length >= u64::BITS as usize
+        || !list_size.is_finite()
+        || list_size <= 0.0
+    {
         return ErrorBits::from_log2(0.0);
     }
     let k = (1u64 << shape.log_trace_length) as f64;
@@ -39,4 +44,33 @@ pub fn deep_ali_error(air: &StarkAirParams, shape: &InstanceShape, list_size: f6
     let factor = ethstark.max(chunked).max(1.0);
     let bits = shape.modulus_bits as f64 - log2(list_size) - log2(factor);
     ErrorBits::from_log2(bits.max(0.0))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn trace_length_shift_boundary_is_conservative() {
+        let air = StarkAirParams {
+            num_constraints: 1,
+            max_constraint_degree: 2,
+            max_combo: 2,
+        };
+        let mut shape = InstanceShape {
+            log_trace_length: 63,
+            modulus_bits: 128,
+            collision_resistance: 128,
+            num_batched_functions: 1,
+        };
+        // 2^63 fits in u64; the DEEP factor is 3 * 2^63 + 1.
+        let expected = 128.0 - 63.0 - log2(3.0);
+        assert!((deep_ali_error(&air, &shape, 1.0).bits() - expected).abs() < 1e-12);
+
+        // These exponents must not panic or wrap back to a small trace.
+        for log_trace_length in [64, 65, usize::MAX] {
+            shape.log_trace_length = log_trace_length;
+            assert_eq!(deep_ali_error(&air, &shape, 1.0).bits(), 0.0);
+        }
+    }
 }
--- a/security/src/stark.rs
+++ b/security/src/stark.rs
@@ -5,6 +5,8 @@
 //!
 //! The conjectured counterpart ([`conjectured_security_report`]) composes the
 //! same sources in the single random-words regime, where the list size is 1.
+//! [`legacy_security_report`] instead uses the LDT's legacy heuristic while
+//! retaining the same non-LDT terms.
 //!
 //! Extra protocol-specific error terms (lookup arguments, custom DEEP
 //! variants, batched openings, …) are passed through `extras: &[ErrorBits]`
@@ -421,6 +423,68 @@ pub fn conjectured_security_report<L: LowDegreeTest>(
         ErrorBits::from_log2(shape.collision_resistance as f64),
     ));
     RegimeReport::new(Regime::Conjectured, terms)
+}
+
+/// Composite legacy conjectured bits. Scalar mirror of
+/// [`legacy_security_report`], with the same terms and grinding sites.
+/// Returns `None` when the low-degree test does not support the legacy regime.
+pub fn legacy_security<L: LowDegreeTest>(
+    ldt: &L,
+    air: &StarkAirParams,
+    shape: &InstanceShape,
+    extras: &[ErrorBits],
+    grinding: &GrindingSites,
+) -> Option<ErrorBits> {
+    let ldt_error = ldt.legacy_conjectured_error(shape)?;
+    let list_size = list_size_conjectured();
+    let ali = air::composition_error(air.num_constraints, list_size, shape.modulus_bits);
+    let deep = boost(
+        deep::deep_ali_error(air, shape, list_size),
+        grinding.out_of_domain,
+    );
+    let batch = conjectured_batching_term(shape, ldt.log_blowup(), grinding.batch_combination);
+    let mut all = Vec::with_capacity(4 + extras.len());
+    all.push(ali);
+    all.push(deep);
+    all.push(ldt_error);
+    all.extend(batch.map(|t| t.bits));
+    all.extend_from_slice(extras);
+    let algebraic = ErrorBits::min(&all);
+    Some(ErrorBits::from_log2(
+        algebraic.bits().min(shape.collision_resistance as f64),
+    ))
+}
+
+/// Composite report using the legacy conjectured LDT bound.
+///
+/// Uses [`LowDegreeTest::legacy_conjectured_error`] under [`Regime::Legacy`].
+/// For FRI this is `num_queries * log_blowup + query_pow_bits`, without the
+/// random-words correction or the commit-phase folding term. This is an
+/// opt-in historical heuristic, not a proven soundness bound.
+///
+/// Only the LDT bound changes: ALI and DEEP-ALI still use list size 1, and
+/// batching, `extras`, grinding outside the LDT, and the commitment-collision
+/// cap are composed as in [`conjectured_security_report`]. Thus this is a
+/// STARK composite using the legacy LDT estimate, rather than the bare FRI
+/// formula. Returns `None` when the LDT does not support the legacy regime.
+pub fn legacy_security_report<L: LowDegreeTest>(
+    ldt: &L,
+    air: &StarkAirParams,
+    shape: &InstanceShape,
+    extras: &[SecurityTerm],
+    grinding: &GrindingSites,
+) -> Option<RegimeReport> {
+    let ldt_error = ldt.legacy_conjectured_error(shape)?;
+    Some(regime_report(
+        Regime::Legacy,
+        air,
+        shape,
+        list_size_conjectured(),
+        ldt_error,
+        conjectured_batching_term(shape, ldt.log_blowup(), grinding.batch_combination),
+        extras,
+        grinding,
+    ))
 }
 
 #[cfg(test)]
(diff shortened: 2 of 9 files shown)


security/src/stark.rs
```diff
@@ -891,6 +955,106 @@ mod tests {
         assert!(b32.udr.terms().iter().all(|t| t.label != BATCH_LABEL));
     }
 
+    /// Legacy reporting retains the ethSTARK query formula without adding
+    /// the random-words correction or the FRI folding round.
+    #[test]
+    fn legacy_report_uses_ethstark_query_bound() {
```
tcoratger, line 970: Every legacy assertion here restates the implementation's own arithmetic — `116.0 == 100*1 + 16`, and the DEEP case is `96 - log2(3*2^20 + 1)`. I checked all of them and they are correct, but a `log_blowup`/`num_queries` swap or a units error would reproduce them just as well, because nothing outside the crate is consulted. That is the specific failure mode of a conjectured regime: self-consistent and unanchored.

ethSTARK §5.10.1 gives three free known-answer vectors. At rho = 1/4 (R = 2) and zeta = 20 grinding bits it fixes `s = 31 / 41 / 55` for `lambda = 80 / 100 / 128`, via Eq. (19):

```text
    lambda >= min{ zeta + R*s, log2|K| } - 1
```

The LDT term itself is faithful — `zeta + R*s` is exactly `query_pow_bits + log_blowup*num_queries`. But pinning those vectors surfaces a real discrepancy: the `-1` is the union bound of Eq. (18), `eps_total = eps_0 + (1 - eps_0)*2^-zeta*eps_1`, and `regime_report` takes a bare `min`. So at R = 2, zeta = 20 this reports 100 bits at `s = 40`, where ethSTARK requires `s = 41` for lambda = 100.

I do not think the `-1` should just be added — the crate uses round-by-round accounting throughout ([2024/1553] §2), where `min` over rounds is the right composition and the union loss is folded into what "lambda bits" means. But then the docs at `fri.rs:503` and `stark.rs:461` should say this reproduces ethSTARK's FRI-query error eps_1 under the RBR convention, not its lambda.

Worth documenting the direction the code is *stronger*, too, because it is a genuine strength: ethSTARK charges the whole pre-query phase at a flat `log2|K|`, while ALI and DEEP here are `log2|K| - log2(num_constraints)` and `log2|K| - log2(3*2^d + 1)`, both strictly below it. The missing `-1` is the only direction in which this exceeds the paper.

Two tests would close it: the three vectors above asserting the LDT term equals `82 / 102 / 130`, and a proptest pinning `legacy >= conjectured >= proven` over the small integer parameter space. The second matters because that ordering is what makes "which number do I quote" unambiguous, and today nothing would notice if a future edit to `conjectured_error`'s eta inverted it.

uni-stark/src/proof.rs
```diff
@@ -26,6 +26,13 @@ pub struct Proof<SC: StarkGenericConfig> {
 }
 
 impl<SC: StarkGenericConfig> Proof<SC> {
+    /// Legacy conjectured security level (in bits).
```
tcoratger, line 29: `legacy_security_report` is candid in `stark.rs:463` — "an opt-in historical heuristic, not a proven soundness bound" — but that sentence does not survive the trip to the API a user actually calls. `LegacySecurity` describes the formula and what it omits, and never says not to deploy on it.

Compare `ConjecturedSecurity` (`security.rs:363`), which for the strictly *safer* number does warn: "The cited paper recommends proven bounds for deployment...". The weaker figure gets the weaker warning.

The gap is real but modest, and I would rather be accurate than alarming about it: for `FriParameters::new_benchmark_high_arity` over KoalaBear degree-4 with a keccak-shaped AIR I compute legacy 101 vs conjectured 100 at `degree_bits = 20` — the batch term binds legacy while the dropped commit round binds conjectured. Over a 62-bit extension at `log_blowup = 4`, `degree_bits = 24` it reaches about 5.9 bits. So 1 to 6 bits, always optimistic.

Suggest one line here and on `LegacySecurity`: reported for historical comparison, strictly above `ConjecturedSecurity`, not a soundness bound, do not size parameters with it.

Related, in `examples/src/proofs.rs:442`: legacy is printed *first*, and the first line of the block is the one people quote. It is the only one of the three that is not a bound at all — `fri.rs:135` says as much. Printing it last and tagging it ("historical, not a bound") would put the strongest claim in the output on the weakest number.

security/src/stark.rs
```diff
@@ -423,6 +425,68 @@ pub fn conjectured_security_report<L: LowDegreeTest>(
...
+pub fn legacy_security_report<L: LowDegreeTest>(
+    ldt: &L,
+    air: &StarkAirParams,
+    shape: &InstanceShape,
+    extras: &[SecurityTerm],
+    grinding: &GrindingSites,
+) -> Option<RegimeReport> {
+    let ldt_error = ldt.legacy_conjectured_error(shape)?;
+    Some(regime_report(
+        Regime::Legacy,
```
tcoratger, line 487: The legacy error goes through `regime_report`, which stamps it `LDT_LABEL` (`"low-degree-test"`) — the label `report.rs:19` reserves for an implementation that reports its phases as one already-composed bound. The legacy error is the opposite: one phase, with the folding round dropped.

`report.rs:22` patches this in prose, but the serialized `RegimeReport` an auditor reads shows `{"label": "low-degree-test", "bits": 116}` with no signal that a round is missing beyond `"regime": "Legacy"`.

`LDT_QUERY_LABEL` already exists and is exactly what this term is — the conjectured path uses it for the same quantity. Using it here would also make a legacy-vs-conjectured label diff read correctly: identical `ldt-query-phase` rows, with the commit-phase row simply absent from legacy, which is the truth. Needs either an explicit term list here or a `ldt_label` parameter on `regime_report`, since it currently hardcodes `LDT_LABEL`.

tcoratger (changes requested):
Reviewed this as the highest-risk PR of today's batch, since a conjectured regime is where a wrong constant silently overstates every proof produced under it. **The math is right and the regime boundary is airtight** — I re-derived every number in the new tests by hand and they all reproduce, and I could not find any way for a proven or conjectured path to reach the legacy constant.

What I verified clean, so you know what the review actually covered:

- `legacy_conjectured_error` is read at exactly two sites (`legacy_security`, `legacy_security_report`). `Regime::Legacy` can only be produced by the latter and can never enter a `SecurityReport`, so `SecurityReport::binding` cannot pick it up. Defaults are the conservative ones: the trait method defaults to `None`, nothing selects a regime automatically, and no exhaustive `match` on `Regime` exists, so the `#[non_exhaustive]` variant breaks nothing.
- **No parameter-selection path touches legacy.** `SecurityAssumption::queries()` and the whole `budget` module are legacy-free, so the new regime cannot cause fewer queries to be chosen. That was my main worry going in.
- Scalar and report agree by construction; `Vec::with_capacity(4 + extras.len())` is exact.
- Nothing can go negative or NaN: `composition_error`, `deep_ali_error` and `batching_term` all `.max(0.0)`, and `f64::max` absorbs NaN. The `as usize` cast saturates and truncates toward zero — all three round down, the conservative direction.
- Legacy is provably the largest of the three, with no crossover: the legacy and conjectured composites share an identical ALI/DEEP/batch/extras/collision term set, so the ordering is decided by the LDT terms alone, and `-log2(rho + eta) <= -log2(rho) = log_blowup` for any `eta >= 0`, with conjectured additionally carrying `conjectured_commit_phase_error` which can only lower a min. Every degenerate branch collapses both to `query_pow_bits`. So `legacy >= conjectured >= proven` universally.
- Grinding accounting is right: `out_of_domain` boosts DEEP once, `batch_combination` boosts the batch term once, `query_pow_bits` is inside the legacy formula, and `commit_pow_bits` is correctly dropped along with the round it boosts. No site double-counted, none credited to a round the protocol does not open.
- Edge parameters: rate 1, zero queries, single query, `max_log_arity = 0`, `num_batched_functions = 1` (which correctly suppresses the batch term before `prox_gaps_error`'s `assert!(num_functions >= 2)` can fire), `num_constraints = 0`, `modulus_bits = 0` — all fine, no division by zero, no unreachable assert reached.

One thing I want to flag that I could not anchor to a diff line, because the file is not in this PR: `security/src/deep.rs:34` does `let k = (1u64 << shape.log_trace_length) as f64;`, and `LegacySecurity::compute_from_params` feeds it `Proof::degree_bits`, which is a `pub`, `Deserialize`d field. The workspace sets no `overflow-checks`, so in release a proof declaring `degree_bits = 64` wraps to `k = 1` and DEEP is graded as a one-row trace — roughly 63 bits of overstatement, in the optimistic direction. Debug panics instead, inside a reporting API. `validate_degree_bits` bounds this, but only inside `verify`, and nothing documents "verify first" as a precondition. Pre-existing for `conjectured_security` and `proven_security`; this PR adds the third instance. Cheapest real fix is a checked shift in `deep_ali_error` returning `ErrorBits::from_log2(0.0)` for `log_trace_length >= 63`, matching that function's existing degenerate-input convention.

Requesting changes only for the test gap below — the implementation itself I am happy with.

security/src/deep.rs
```diff
@@ -25,9 +25,14 @@ use libm::log2;
 use crate::error::ErrorBits;
 use crate::shape::{InstanceShape, StarkAirParams};
 
-/// `-log2(ε_DEEP)` in bits. Returns 0 bits if inputs are degenerate.
+/// `-log2(ε_DEEP)` in bits. Returns 0 bits if inputs are degenerate or the trace size
+/// cannot be represented by `u64`.
 pub fn deep_ali_error(air: &StarkAirParams, shape: &InstanceShape, list_size: f64) -> ErrorBits {
-    if shape.modulus_bits == 0 || !list_size.is_finite() || list_size <= 0.0 {
+    if shape.modulus_bits == 0
+        || shape.log_trace_length >= u64::BITS as usize
```
tcoratger, line 32: This is exactly the fix I had in mind, and it does close the `deep_ali_error` hole. Flagging one thing for a follow-up rather than here, because it's the sibling sites rather than this one:

the other LDE-domain shifts key off `log_trace_length + log_blowup`, not `log_trace_length`, so they overflow *below* this guard's threshold:

```
fri.rs:170   commit_phase_error_udr     1u64 << (log_trace_length + log_blowup)
fri.rs:206   commit_phase_error_ldr_m   same
fri.rs:248   proven_error_udr           same
fri.rs:275   proven_error_ldr_m         same
proximity.rs:23  alpha_udr              same
```

So `log_trace_length = 63, log_blowup = 1` passes this guard and then panics at `fri.rs:170` in debug. I confirmed both directions:

```rust
// log_trace_length = 63, log_blowup = 1, modulus_bits = 252, max_log_arity = 4
conjectured_security_report(...)
// debug:   panicked at security/src/fri.rs:170: attempt to shift left with overflow
// release: 114.6 bits, binding ldt-query-phase
//          -> n wrapped to 1, so the commit round was graded on a one-element domain
```

Net effect: `legacy_security` is now hardened and `conjectured_security` / `proven_security` are not, which is the inverse of the priority order — the two that are actual bounds still take deserialized `Proof::degree_bits` straight into a wrapping shift. `legacy_security_rejects_an_unrepresentable_trace_length` reads as a crate-level property but only holds on the one path that isn't a bound.

Cheapest fix, matching the convention you used here: hoist the check to `lde_log`, e.g. `shape.log_trace_length.checked_add(regime.log_blowup).filter(|&l| l < u64::BITS as usize)` returning the module's degenerate value, and extend the existing test to the `log_blowup > 0` boundary. Happy to send that as a follow-up if you'd rather not touch it in this PR — nothing here regresses, it just doesn't finish the sweep.

uni-stark/src/proof.rs
```diff
@@ -26,6 +26,16 @@ pub struct Proof<SC: StarkGenericConfig> {
...
+    ///
+    /// For historical comparison only: this may exceed [`ConjecturedSecurity`]
+    /// and is not a soundness bound. Do not use it to size deployment parameters.
+    ///
+    /// See [`LegacySecurity`].
+    pub fn legacy_security(&self, params: &StarkSecurityParams) -> LegacySecurity {
```
tcoratger, line 35: Nit, purely cosmetic: the printed output in `examples/src/proofs.rs` now reads proven -> conjectured -> legacy, which is the right ordering (strongest claim first, the one that isn't a bound last). This impl block still reads legacy -> conjectured -> proven. Moving `legacy_security` below `proven_security` would make the rustdoc page tell the same story as the terminal output. Not worth a re-push on its own.

tcoratger (approved):
All three points are addressed, and better than I asked for.

- `LDT_QUERY_LABEL` now flows through a `SecurityTerm` parameter on `regime_report` instead of a hardcoded label. That is the right shape — it makes the proven path's single `low-degree-test` row and the legacy/conjectured `ldt-query-phase` rows a caller's choice rather than a property of the orchestrator, and a legacy-vs-conjectured label diff now reads exactly as the truth: identical query rows, commit row absent from legacy.
- The ethSTARK vectors are pinned tightly. `(31, 82), (41, 102), (55, 130)` at `R = 2, zeta = 20` are *minimal* for `lambda = 80/100/128` — `s = 30` gives `80 - 1 = 79 < 80`, `s = 40` gives `99 < 100`, `s = 54` gives `127 < 128` — so these genuinely pin `zeta + R*s` and would catch a `log_blowup`/`num_queries` swap. And the docs now say what the code reproduces (`eps_1` under the RBR convention) rather than what it does not (`lambda`), which was the actual gap.
- The `legacy >= conjectured >= proven` proptest is there, with an honest guard on the `modulus_bits >= 32` branch. I ran it at `PROPTEST_CASES=20000` in release: holds.

Verified on the branch: 168 tests pass, clippy `--all-targets -D warnings` clean on security/uni-stark/examples, `cargo fmt --check` clean.

One genuinely minor follow-up below on the shift guard — worth a separate PR, not a change here. Merging.

PR #2066, opened 2026-09-08
fix(scripts): let the shared check command pin the target features a leg names

tcoratger:
Follow-up to #2062, from my review of it. The check ledger in that PR is intact — I verified every leg, step, `RUSTFLAGS` combination and `--features` flag survives byte-identically, and that all 21 rendered job names are unchanged so no required status check is orphaned. This closes the one place where the stated goal was not met.

## `architecture` could not reproduce the leg it is named for

CONTRIBUTING says `architecture` "uses `architecture` for AVX-512, VPCLMULQDQ, and SVE2 coverage", but `check.py` never set `-C target-feature`. The workflow pins the features in a separate `Set flags` step, so **CI was always right** — it was only the shared local command that was blind:

```text
    check.py architecture --target x86_64-unknown-linux-gnu
      ->  compiles the baseline feature set
      ->  every cfg(target_feature = "avx512f") block is cfg-ed out
      ->  green, having checked none of what the command names
```

The `+avx2`, `+pclmulqdq` and macOS `-sha3` legs had no shared recipe at all, and those are the ones most likely to catch a SIMD-only miscompile.

`--target-feature` now reaches the compiler the way a leg does — through `RUSTFLAGS`, not the argv, because that is what decides `cfg(target_feature = ..)`:

```bash
python3 scripts/check.py architecture --target x86_64-unknown-linux-gnu --target-feature +avx512f
python3 scripts/check.py test --target-feature +avx2
python3 scripts/check.py test --package p3-keccak --target-feature -sha3
```

Two properties that make this safe to land without touching the workflow:

- **It appends, and the append is idempotent.** A leg's `-C debuginfo=0` survives, and if `Set flags` has already pinned the same features the flag is not added twice. So the two mechanisms compose rather than fight, and wiring `${{ matrix.features }}` through the steps (and dropping `Set flags`) becomes a safe separate change rather than a prerequisite.
- **Every CI recipe is byte-identical when the flag is absent.** Verified for `test`, `test --parallel`, `doctest`, `doctest --parallel`, `architecture`, and `embedded`. `ci.yml` is untouched, so no job name moves — which matters, because the comment above that matrix warns that a leg's rendered name embeds its `features` and two of those names are required checks.

## A missing tool ended a long run in a traceback

`main`'s `try` covered only `commands_for`, not the execution loop. Skip `taplo-cli` from CONTRIBUTING's install line and `check.py full` died with `FileNotFoundError: ... 'taplo'` and a stack trace, after clippy and docs had already been paid for. It now names the tool and points at CONTRIBUTING, exiting 127.

## The embedded exclusion set was sampled, not pinned

`test_repository_metadata_explicitly_excludes_host_only_packages` asserted three names but not the complement, so a manifest declaring `embedded = false` could shrink the ledger from 45 crates to 44 with nothing in the test diff saying a crate stopped being checked.

That is the mirror image of the hand-maintained list this mechanism replaced — which omitted `p3-binary-pcs` for exactly the reason that nothing asserted completeness. The set is now pinned exactly against `cargo metadata`, so opting a crate out requires a deliberate test edit.

## Test plan

- [x] `python3 -m unittest scripts.test_check` — 18 passed (13 pre-existing, 5 new)
- [x] `python3 scripts/check.py --dry-run lint --check scripts` — dispatches the suite
- [x] Recipe equivalence checked by hand for `test`, `test --parallel`, `doctest`, `doctest --parallel`, `architecture --target … --parallel`, `embedded` — argv identical to `main` in every case
- [x] `PATH=/usr/bin python3 scripts/check.py lint --check toml` — reports `error: taplo not found (…); see CONTRIBUTING.md for the required tools` instead of a traceback

No cargo commands were run; this PR changes no Rust.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,10 +28,10 @@ Run `python3 scripts/check.py full` before opening a pull request. It covers the
 | --- | --- | --- |
 | `fast` | Check all host targets without running them | `--package NAME`, `--parallel` |
 | `full` | Run all routine host checks | none |
-| `test` | Run tests through `cargo-nextest` | `--package NAME`, `--parallel` |
-| `doctest` | Run Rust documentation tests | `--package NAME`, `--parallel` |
+| `test` | Run tests through `cargo-nextest` | `--package NAME`, `--parallel`, `--target-feature FEATURES` |
+| `doctest` | Run Rust documentation tests | `--package NAME`, `--parallel`, `--target-feature FEATURES` |
 | `lint` | Run all CI script/lint/doc/format checks | `--check scripts\|sort\|toml\|deps\|clippy\|docs\|fmt` |
-| `architecture` | Compile all targets for a non-runnable target | `--target TARGET`, `--parallel` |
+| `architecture` | Compile all targets for a non-runnable target | `--target TARGET`, `--parallel`, `--target-feature FEATURES` |
 | `embedded` | Build every eligible workspace library with `--lib` | `--target TARGET` |
 | `wasm` | Run the wasm compile and SIMD smoke recipes | `--step build\|test\|smoke\|bench\|merkle` |
 | `bench` | Execute every benchmark body once, excluding the large `p3-dft` sweep | none |
@@ -40,6 +40,14 @@ Run `python3 scripts/check.py full` before opening a pull request. It covers the
 
 `architecture` is a compile-only command. CI uses `test` and `doctest` only on architectures whose runner can execute the selected instructions, and uses `architecture` for AVX-512, VPCLMULQDQ, and SVE2 coverage where runner hardware is not guaranteed.
 
+Target features decide `cfg(target_feature = ..)`, so a leg that pins them compiles different code. `--target-feature` reaches the compiler through `RUSTFLAGS`, appended to whatever the caller already exports, and leaves the argv unchanged. Pass the same string the CI leg uses, or the command compiles the baseline feature set instead of the one it names:
+
+```bash
+python3 scripts/check.py architecture --target x86_64-unknown-linux-gnu --target-feature +avx512f
+python3 scripts/check.py test --target-feature +avx2
+python3 scripts/check.py test --package p3-keccak --target-feature -sha3
+```
+
 The embedded package list comes from `cargo metadata`. Every workspace package with a library target is included unless its manifest declares:
 
 ```toml
@@ -47,7 +55,7 @@ The embedded package list comes from `cargo metadata`. Every workspace package w
 embedded = false
 ```
 
-`p3-examples` and `p3-field-testing` are explicit host-only exclusions. Building each selected package with `--lib` prevents host-only binaries and examples from breaking an otherwise `no_std` library build. New library crates therefore enter embedded coverage automatically.
+`p3-examples` and `p3-field-testing` are explicit host-only exclusions, and `scripts/test_check.py` pins that set exactly, so opting a crate out is a deliberate test change rather than a silent loss of coverage. Building each selected package with `--lib` prevents host-only binaries and examples from breaking an otherwise `no_std` library build. New library crates therefore enter embedded coverage automatically.
 
 A package whose meaningful tests require a set of opt-in features can declare one additional CI invocation:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,10 +28,10 @@ Run `python3 scripts/check.py full` before opening a pull request. It covers the
 | --- | --- | --- |
 | `fast` | Check all host targets without running them | `--package NAME`, `--parallel` |
 | `full` | Run all routine host checks | none |
-| `test` | Run tests through `cargo-nextest` | `--package NAME`, `--parallel`, `--target-feature FEATURES` |
-| `doctest` | Run Rust documentation tests | `--package NAME`, `--parallel`, `--target-feature FEATURES` |
+| `test` | Run tests through `cargo-nextest` | `--package NAME`, `--parallel`, `--target-feature=FEATURES` |
+| `doctest` | Run Rust documentation tests | `--package NAME`, `--parallel`, `--target-feature=FEATURES` |
 | `lint` | Run all CI script/lint/doc/format checks | `--check scripts\|sort\|toml\|deps\|clippy\|docs\|fmt` |
-| `architecture` | Compile all targets for a non-runnable target | `--target TARGET`, `--parallel`, `--target-feature FEATURES` |
+| `architecture` | Compile all targets for a non-runnable target | `--target TARGET`, `--parallel`, `--target-feature=FEATURES` |
 | `embedded` | Build every eligible workspace library with `--lib` | `--target TARGET` |
 | `wasm` | Run the wasm compile and SIMD smoke recipes | `--step build\|test\|smoke\|bench\|merkle` |
 | `bench` | Execute every benchmark body once, excluding the large `p3-dft` sweep | none |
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -40,14 +40,22 @@ Run `python3 scripts/check.py full` before opening a pull request. It covers the
 
 `architecture` is a compile-only command. CI uses `test` and `doctest` only on architectures whose runner can execute the selected instructions, and uses `architecture` for AVX-512, VPCLMULQDQ, and SVE2 coverage where runner hardware is not guaranteed.
 
+Target features decide `cfg(target_feature = ..)`, so a leg that pins them compiles different code. `--target-feature` reaches the compiler through `RUSTFLAGS`, appended to whatever the caller already exports, and leaves the argv unchanged. Pass the same string the CI leg uses, or the command compiles the baseline feature set instead of the one it names:
+
+```bash
+python3 scripts/check.py architecture --target x86_64-unknown-linux-gnu --target-feature +avx512f
+python3 scripts/check.py test --target-feature +avx2
+python3 scripts/check.py test --package p3-keccak --target-feature -sha3
```
Nashtare: [P2] This documented macOS fallback command exits before reaching Cargo: argparse treats `-sha3` as an option and reports `argument --target-feature: expected one argument`. I reproduced it with `--dry-run`; `--target-feature=-sha3` succeeds. Please use the equals form here and in the option's guidance, or support a separate negative value in the parser.

scripts/check.py
```diff
@@ -331,14 +343,33 @@ def display(command: Sequence[str]) -> str:
...
     environment = os.environ.copy()
-    deny_broken_links = "-D rustdoc::broken_intra_doc_links"
-    current = environment.get("RUSTDOCFLAGS", "")
-    if deny_broken_links not in current:
-        environment["RUSTDOCFLAGS"] = f"{current} {deny_broken_links}".strip()
+    if target_feature:
+        flag = f"-C target-feature={target_feature}"
+        current = environment.get("RUSTFLAGS", "")
+        if flag not in current:
+            environment["RUSTFLAGS"] = f"{current} {flag}".strip()
```
Nashtare: [P2] `doctest` accepts this option, but writing only `RUSTFLAGS` does not pass the feature to `rustdoc --test`. The library gets the feature while its documentation tests retain the baseline `cfg(target_feature)` values. On aarch64, a doctest containing `assert!(cfg!(target_feature = "sve2"));` fails with `doctest --target-feature +sve2` and passes after also setting `RUSTDOCFLAGS='-C target-feature=+sve2'`. Please propagate the requested features to the doctest compiler while preserving inherited rustdoc flags.

scripts/check.py
```diff
@@ -331,14 +343,33 @@ def display(command: Sequence[str]) -> str:
...
         return None
     environment = os.environ.copy()
-    deny_broken_links = "-D rustdoc::broken_intra_doc_links"
-    current = environment.get("RUSTDOCFLAGS", "")
-    if deny_broken_links not in current:
-        environment["RUSTDOCFLAGS"] = f"{current} {deny_broken_links}".strip()
+    if target_feature:
+        flag = f"-C target-feature={target_feature}"
+        current = environment.get("RUSTFLAGS", "")
+        if flag not in current:
```
Nashtare: [P2] The substring check can find a feature that a later setting disables. For example, with inherited `RUSTFLAGS='-C target-feature=+avx2,-avx2'`, requesting `--target-feature +avx2` skips the append and leaves AVX2 disabled. I confirmed the resulting flags omit `target_feature="avx2"` from `rustc --print cfg`. The same happens with separate enabling/disabling flags. Deduplication needs to consider the effective setting, so the requested feature is actually enabled.

scripts/check.py
```diff
@@ -331,14 +343,33 @@ def display(command: Sequence[str]) -> str:
...
+    if not rustdoc and not target_feature:
         return None
     environment = os.environ.copy()
-    deny_broken_links = "-D rustdoc::broken_intra_doc_links"
-    current = environment.get("RUSTDOCFLAGS", "")
-    if deny_broken_links not in current:
-        environment["RUSTDOCFLAGS"] = f"{current} {deny_broken_links}".strip()
+    if target_feature:
+        flag = f"-C target-feature={target_feature}"
+        current = environment.get("RUSTFLAGS", "")
```
Nashtare: [P2] If the inherited environment contains `CARGO_ENCODED_RUSTFLAGS`, Cargo ignores `RUSTFLAGS`, so the new feature flag has no effect despite the printed announcement. I reproduced this with encoded `-C debuginfo=0`: a crate gated on `sve2` passes `architecture --target aarch64-apple-darwin --target-feature +sve2` without the encoded variable and fails with it. Please append to the active encoded flags, or reject the combination explicitly instead of silently checking the wrong feature set.

Nashtare (changes requested):

CONTRIBUTING.md
tcoratger: Confirmed and fixed. Reproduced first:

```
$ check.py --dry-run test --package p3-keccak --target-feature -sha3
check.py test: error: argument --target-feature: expected one argument
```

I took the second option as well as the first, since a leg that turns a feature off should not be a spelling the caller has to remember. `attach_leading_minus_values` joins a minus-prefixed value onto its option before parsing, so both forms now reach the compiler identically:

```
$ check.py --dry-run test --package p3-keccak --target-feature -sha3
+ RUSTFLAGS += -C target-feature=-sha3
+ cargo nextest run -p p3-keccak
```

The two spellings are asserted equal in `test_a_leading_minus_value_is_accepted_in_both_spellings`.

The documentation and the option help both use the equals form as canonical, and the option table now reads `--target-feature=FEATURES` so the shape is visible at a glance.

scripts/check.py
tcoratger: Right, and this was the one I had entirely missed rather than got subtly wrong — the ordinary flag variable does not reach the documentation tool at all.

```
before:  RUSTFLAGS = '-C target-feature=+sve2'   RUSTDOCFLAGS = None
after :  RUSTFLAGS = '-C target-feature=+sve2'   RUSTDOCFLAGS = '-C target-feature=+sve2'
```

A doctest command now pins both. The predicate is `command[:2] == ["cargo", "test"] and "--doc" in command`, so it fires for the plain and the parallel doctest legs alike and not for `cargo nextest run`.

Inherited rustdoc flags are preserved, and the documentation build keeps its broken-link denial alongside the feature:

```
RUSTDOCFLAGS = '-C target-feature=+avx2 -D rustdoc::broken_intra_doc_links'
```

Two tests cover it, one for the plain variable and one for the encoded one. The dry run also prints both variables now, so the two-target behaviour is visible without reading the source.

scripts/check.py
tcoratger: Confirmed, with your own instrument:

```
$ rustc --print cfg -C target-feature=+avx2,-avx2 | grep -c 'target_feature="avx2"'
0
$ rustc --print cfg -C target-feature=+avx2,-avx2 -C target-feature=+avx2 | grep -c 'target_feature="avx2"'
1
```

Rather than make the deduplication understand effective settings, I dropped it. The flag is now always appended, and the last setting of a given feature is the one that takes effect — so the request wins by construction, with no matching logic to get wrong.

The cost is a possibly redundant flag in the string when a leg already pinned the same features, which is harmless. That trade seemed better than a matcher that has to model enable-then-disable, separate flags, and comma lists correctly; the failure mode of getting that wrong is silent, and this way there isn't one.

`test_a_requested_feature_outranks_an_inherited_disable` pins it, and the inherited-flags test still asserts that an exported `-C debuginfo=0` survives.

scripts/check.py
tcoratger: Good catch, and the worst of the four — the announcement was printing while nothing reached rustc.

The flag now follows whichever variable is active, for both the ordinary and the documentation compiler:

```
CARGO_ENCODED_RUSTFLAGS = '-C\x1fdebuginfo=0'
  ->  '-C\x1fdebuginfo=0\x1f-C\x1ftarget-feature=+sve2'   and RUSTFLAGS left unset
```

`append_compiler_flag` takes the plain and encoded names, writes to the encoded one when the caller set it, and returns the name it used. The dry run prints that name, so the announcement can no longer claim a variable the run did not touch:

```
+ CARGO_ENCODED_RUSTFLAGS += -C target-feature=+sve2
```

Each unit is appended separately, matching the one-argument-per-separator encoding, so `-C` and `target-feature=+sve2` arrive as two arguments rather than one. Two tests cover it, one per compiler.

I chose appending over rejecting the combination because the encoded variable is what a Cargo-invoked-from-Cargo environment exports, so refusing it would break the case most likely to hit it.

Nashtare (approved):

PR #2068, opened 2026-09-08
feat(fri)!: type the PCS transcript above the FRI low-degree test

tcoratger:
Second step of the typed-transcript rollout, after #2035 migrated FRI itself. This one migrates the PCS wrapped around it: `TwoAdicFriPcs`, and `HidingFriPcs` through it.

## The problem

The FRI PCS wrote the phase it owns by hand, on both sides.

The prover walked three nested loops — commitment, matrix, opening point — observing each claimed evaluation as it went. The verifier reproduced the same walk over a structure its *caller* had assembled. Nothing checked that the two agreed. The order was maintained by eye, across two crates.

Two things were unbound underneath that:

- How many values an opening carries. Both sides absorbed whatever arrived, so a differently shaped set of claims desynchronised the sponge silently, and the divergence only surfaced several challenges later as an opaque `FinalPolyMismatch`.
- `batch_proof_of_work_bits`. It reached neither a pattern fingerprint nor any instance label — the one `FriParameters` field bound by nothing. A verifier configured to demand less work than the proof was produced against silently accepted the cheaper grind.

## The idea

The PCS now derives its phase through `ProverState` / `VerifierState`, over a description each side builds from its own input: the prover from the matrices it holds, the verifier from the claims it was handed.

New `fri::pcs_transcript` holds that description (`PcsShape`) plus one driver per side.

```text
    per (commitment, matrix, opening point):
                 claimed evaluations   one step of that matrix's width
    batch grinding                     only when the difficulty is positive
    batching challenge                 one extension element
    Begin  low-degree test             bracket around the delegated run
    End    low-degree test
```

Each claimed opening is **one step of its own width**, so a claim of another width is a shape mismatch caught before a single value is absorbed.

The step sequence sees one flat list of widths, so the grouping goes in the **instance label**. Two matrices of width 3 opened at one point each, and one matrix of width 3 opened at two points, both flatten to `[3, 3]`; binding the commitment, matrix and point counts gives them distinct seeds.

Every count comes from the verifier's own `commitments_with_opening_points`, never from the proof.

## The batching phase moves in

`verify_fri` checked `batch_proof_of_work_bits` and sampled `alpha` on the raw challenger, under a comment saying the challenge belonged to the caller's transcript rather than to FRI's. This is that caller.

The grind is now a described `Kind::Pow` step and `alpha` a described challenge. `verify_fri` takes the challenge as a final argument and no longer touches `batch_pow_witness`.

`Length::Fixed(bits)` on that step puts the difficulty in the fingerprint, like the other two grinding difficulties, so mismatched configurations now land on different seeds instead of one side quietly accepting less work.

## The delegation bracket

This is the first place a migrated component calls another migrated component, so the composition primitive earns its keep here.

`Begin Protocol("low_degree_test")` and its matching `End` sit around the region where the PCS lends the sponge to `prove_fri` / `verify_fri`, which enforce their own pattern inside, under their own seed. `Hierarchy::Begin` / `Hierarchy::End` and `InteractionPattern::validate`'s matched-bracket check already existed; nothing was added for them.

The markers absorb nothing. What they record is that a sub-protocol runs there, and where — so a future reordering of the two phases is a fingerprint change rather than an invisible one.

`PcsProverTranscript::delegate` and its verifier twin take the delegated run as a closure, so the bracket cannot be left open. On the verifier side the `End` is recorded whichever way the delegated run goes, and the rejection is propagated *after* `finish`, never across an unfinished driver.

## Verifier drop safety

Both drivers panic on drop when the pattern was not fully replayed, so every early return on the span between constructing a `PcsVerifierTranscript` and `finish()` has to release that check first. There are exactly three exits on that span in `TwoAdicFriPcs::verify`:

- `claimed_openings(..)?` — the only error path is `VerifierState::observe_extensions`, which poisons through `poison()` before returning.
- `batch_phase(..)?` — `MissingPowWitness` calls `state.abort()` explicitly; `PowWitness` comes from `VerifierState::observe_pow`, which aborts before returning.
- the delegated `verify_fri` result — bound, not propagated. `finish()` runs first, then the `?`.

A panic inside the delegated run drops the driver mid-pattern, which is silent: #2035's `drop_check_may_panic` gate yields to the panic already in flight.

## What is not bound

Whether a claimed width is the committed matrix's width. That is the input commitment scheme's question, and it answers it when it authenticates the opened rows (`check_widths`, reported as `PointEvaluationCountMismatch` / an MMCS error). The transcript binds only that both sides describe the same widths.

## No trait change

`Pcs::open` and `Pcs::verify` still take `&mut Challenger`. Each impl builds its own driver from a borrowed sponge, so `Challenger` keeps meaning "the sponge" and the `Pcs<Challenge, Challenger>` signature is untouched.

`HidingFriPcs` inherits the whole change for free: it splits and re-merges the random codewords and then delegates to the inner PCS, which is where the transcript now lives.

## ⚠️ Addition to `challenger/src/fs` (four sibling PRs are in flight)

One addition, and it is the only one:

- `ProverState::challenger_mut` and `VerifierState::challenger_mut`.

A component that lends its sponge to a sub-protocol has to reach it. Nothing that travels through them is validated against the pattern, which is exactly why the bracket exists; both doc comments say so and point at `begin_protocol` / `end_protocol`. Their first consumers are the two `delegate` methods in this PR.

Nothing else in `challenger/src/fs` changed.

## Breaking changes

- **The FRI PCS transcript changes, so proofs produced before this do not verify after it.** Seeding now absorbs a second protocol identity and shape fingerprint before FRI's own.
- `verify_fri` takes the batching challenge as a final argument and no longer checks `proof.batch_pow_witness`. The caller's transcript does both. Every in-tree caller is `TwoAdicFriPcs::verify`.
- `FriError` gains a `ClaimedEvaluationCountMismatch` variant.

### Fixtures

The two two-adic compatibility fixtures are regenerated:

- `uni-stark/tests/fixtures/uni_stark_two_adic_v0_8_0.postcard`
- `batch-stark/tests/fixtures/batch_stark_two_adic_v0_8_0.postcard`

Those are the two that route through this PCS. The circle fixtures regenerate byte-identical, because `p3-circle` has its own PCS and its own FRI driver; the WHIR and STIR ones never enter this path. Both regenerated files shrink slightly — different query indices collide differently, so the Merkle multi-openings deduplicate more paths.

## Tests

- `PcsShape` seed coverage, one test per knob: opening widths, the grouping into matrices and points, the commitment count, the grinding difficulty, and zero-bits-versus-one-bit.
- The bracket absorbs nothing: a bracketed verifier run and a prover run over the same shape land on the same `alpha` and leave the sponge in the same state.
- Rejections: a described grinding step with no witness, and an opening of the wrong width.
- End-to-end mutation tests on the real `Pcs::open` / `Pcs::verify` roundtrip: a perturbed claimed evaluation, a value moved between two matrices of different widths, a tampered batch grinding witness, and a verifier demanding more grinding than the prover did.

The two batch-grinding tests that used to sit on `verify_fri` moved up to that roundtrip, which is where the check now lives.

## Verification

```text
cargo test -p p3-fri --all-targets              68 + 2 + 26 passed
cargo test -p p3-challenger --all-targets      182 passed
cargo test -p p3-circle --all-targets           48 passed
cargo clippy -p p3-fri -p p3-uni-stark -p p3-batch-stark --all-targets -- -D warnings   clean
cargo +nightly fmt --all -- --check             clean
cargo build --workspace --all-targets           clean
```


tcoratger:
Superseded by #2075, which carries the same change with its branch in this repository so the stack chains properly. Closing this one to avoid a duplicate.

tcoratger:

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2069, opened 2026-09-08
docs(merkle-tree): state the real bound on the staged row buffer

tcoratger:
Follow-up to #2051, from re-reading my own merged code while checking #2057. The doc on the staging budget promises something the code does not hold to.

## What is wrong

`ROW_SCRATCH_BYTES` says "16 KiB keeps that buffer inside the first-level cache". The group size is rounded up to a whole lane group, so that promise breaks for any row wide enough that fewer than one lane group fits:

```text
    lanes= 2  row_bytes= 16383  ->  2 rows      32 KiB     2x the target
    lanes= 4  row_bytes= 16383  ->  4 rows      64 KiB     4x
    lanes= 8  row_bytes= 16384  ->  8 rows     128 KiB     8x
    lanes= 8  row_bytes= 65536  ->  8 rows     512 KiB    32x
    lanes= 1  row_bytes= 65536  ->  1 row       64 KiB     4x
```

The last row is the one that shows this is not fixable by arithmetic: one row has to be staged however wide it is, so a row past the target overshoots even at a single lane.

## What I did not do

My first attempt made the sizing honour the target strictly, falling back to however many rows fit. I threw it away, and the reasoning is worth recording because it is the whole point of the change.

Honouring the target means staging fewer rows than there are lanes, so a permutation runs partly idle. A Keccak-f permutation is on the order of a thousand cycles; the L1-to-L2 step the larger buffer gives up is a handful of cycles per access. Trading a permutation for a cache level is a bad trade, and I could not justify a behaviour change on reasoning I had already contradicted myself on once, without a measurement.

So the behaviour is unchanged — verified equal to the current expression across every row width from 1 to 70000 and every lane count in `{1, 2, 4, 8, 16}`, zero differing inputs. Only the promise changes.

## What the change is

The constant now states the bound it actually holds to, `max(target, lanes * row bytes)`, with the occupancy-over-residency reason spelled out so the next reader does not re-derive it.

The sizing moves into `rows_per_call(row_bytes, lanes)`, which makes the bound testable without building a matrix wide enough to trigger it. It also picks up a guard for a hasher reporting zero lanes, which would previously have divided by zero.

`a_staged_group_is_lane_aligned_and_bounded` pins both halves across `lanes` in `{0, 1, 2, 4, 8, 16}` and eight row widths straddling the target. I checked it is not vacuous by mutation:

- dropping the lane rounding fails the alignment assertion;
- rounding before the multiply instead of after fails the bound assertion at `lanes=2, row_bytes=16383` with `staged 32766 bytes`.

## Test plan

- [x] `cargo nextest run -p p3-merkle-tree` — 106 passed
- [x] `cargo clippy -p p3-merkle-tree --all-targets -- -D warnings` — clean
- [x] `cargo test --doc -p p3-merkle-tree`
- [x] `cargo +nightly fmt --all` — no diff
- [x] sizing verified identical to the current expression over the sweep above
- [x] both assertions mutation-checked as described


merkle-tree/src/merkle_tree.rs
```diff
@@ -332,8 +332,47 @@ const TASK_NODES: usize = 1024;
...
+///     narrow rows        : many whole lane groups fit, the target holds
+///     row near the target: one lane group, up to `lanes` times the target
+///     row past it        : one lane group of very wide rows
+/// ```
+///
+/// Idle lanes cost a whole permutation each.
+///
+/// That is far more than the cache level the larger buffer gives up.
+///
+/// So occupancy wins over residency wherever the two disagree.
```
Nashtare: **The rationale does not hold for the case the doc singles out, because the only multi-lane consumer abandons batching there.**

`hash_rows_batched` is reached only when `H::LANES > 1` (`first_digest_layer`, line 648), and the only hasher in the tree with `LANES > 1` is `SerializingHasher<Inner>` forwarding a vectorized Keccak. Its `hash_many` (`symmetric/src/serializing_hasher.rs:78-84`) opens with:

```rust
// A row too wide for the byte budget gains nothing from grouping.
if row_bytes > ROW_BYTES_SCRATCH {           // 8 KiB
    for (digest, row) in out.iter_mut().zip(input.chunks_exact(row_len)) {
        *digest = self.hash_iter(row.iter().copied());
    }
    return;
}
```

So for `row_bytes > 8 KiB` — precisely the `row past it : one lane group of very wide rows` line of the new table, and the `lanes= 1, row_bytes= 65536` and `lanes= 8, row_bytes= 65536` rows of the PR description — the inner hasher hashes **one message at a time through the scalar sponge**. Every lane is idle regardless of what this layer staged. The overshoot to `lanes * row_bytes` (up to 32x the target, per the PR's own table) buys nothing at all in that regime; it is pure copy cost and pure cache pressure.

"Occupancy wins over residency wherever the two disagree" is therefore true for narrow rows (where the two never disagree) and false for wide ones (where they do). Given this PR's whole purpose is to make the doc match the code, that is the sentence to fix.

There is a second budget mismatch worth naming while you are here: this layer stages into a 16 KiB target and `SerializingHasher::hash_many` immediately **re-stages** the same rows into its own 8 KiB buffer. The claim that "16 KiB keeps that buffer inside the first-level cache, so the hasher reads the rows back while they are still hot" describes a buffer the hasher does not read directly.

merkle-tree/src/merkle_tree.rs
```diff
@@ -332,8 +332,47 @@ const TASK_NODES: usize = 1024;
...
 const ROW_SCRATCH_BYTES: usize = 16 * 1024;
 
+/// Rows one hash call stages, from the byte target and the hasher's lane count.
+///
+/// The count is a whole number of lane groups, so no permutation runs partly idle.
```
Nashtare: > The count is a whole number of lane groups, so no permutation runs partly idle.

The count is; the *last group* is not. `hash_chunk` does `digests.chunks_mut(rows_per_call)`, so the final group of the final task carries `out.len() % rows_per_call` rows, which need not be a multiple of `lanes` (the serial arm covers up to `SERIAL_LEVEL_NODES = 1024` rows in one `hash_chunk`; the parallel arm makes `task` a whole number of groups, so only the tail is short). That last `hash_many` call runs partly idle, and for a level whose height is not a multiple of `rows_per_call` there is always one.

Same standard as the comment above: this doc exists because the previous one promised something the code does not hold to, and this sentence promises something the code does not hold to at the tail.

Also worth writing down here, since this doc is now the place someone will look: the consequence of the `lanes * row_bytes` bound is not one buffer but one **per rayon task** — `hash_chunk` allocates `Vec::with_capacity(rows_per_call * total_width)` per task (line 428), and the parallel arm spawns `out.len() / task` of them. At `lanes = 8` and a 64 KiB row that is 512 KiB per task; on a 96-core box the level's staging peak is tens of MiB. The bound the constant now states is per-call, and the number an operator cares about is per-call times the task count.

merkle-tree/src/merkle_tree.rs
```diff
@@ -332,8 +332,47 @@ const TASK_NODES: usize = 1024;
...
+///
+/// A row at or past the target yields one lane group.
+///
+/// That is the smallest count keeping every lane busy.
+fn rows_per_call(row_bytes: usize, lanes: usize) -> usize {
+    // A hasher reporting no lanes still hashes one message at a time.
+    let lanes = lanes.max(1);
+
+    // Whole lane groups inside the target, and one group when none fits.
+    (ROW_SCRATCH_BYTES / row_bytes / lanes).max(1) * lanes
```
Nashtare, line 387: The defensiveness is inverted here.

- `lanes.max(1)` guards a divisor that no implementation in the tree can produce. `CryptographicHasher::LANES` defaults to `1` (`symmetric/src/hasher.rs:13`), every impl either takes the default or sets a positive constant, and the semantics ("Number of equal-length messages this implementation hashes most efficiently in one call") make `0` meaningless. The new test even pins the behaviour for it — `assert_eq!(rows_per_call(64, 0), rows_per_call(64, 1))` — freezing a contract for an input that cannot occur.
- `row_bytes` is divided by twice, unguarded, and the doc promises `# Returns: A count in 1..` with no stated precondition. `row_bytes == 0` panics. It cannot happen *from the one call site* because of the `.max(1)` at line 415, but that is now a precondition of a standalone unit-tested function living 40 lines away from the code that establishes it.

Either document the precondition (`# Panics: if row_bytes is zero`) or apply the same `.max(1)` you gave `lanes`. As it stands the guarded parameter is the impossible one and the unguarded parameter is the reachable one.

merkle-tree/src/merkle_tree.rs
```diff
@@ -373,10 +412,8 @@ fn hash_rows_batched<F, W, H, M, const DIGEST_ELEMS: usize>(
     let total_width: usize = matrices.iter().map(|m| m.width()).sum();
     let row_bytes = (total_width * size_of::<F>()).max(1);
 
-    // Fill as many whole lane groups as the scratch budget allows, and never fewer than one,
-    // so every hash call keeps the hasher's vector width busy.
-    let lanes = H::LANES;
-    let rows_per_call = (ROW_SCRATCH_BYTES / row_bytes / lanes).max(1) * lanes;
+    // Fill as many whole lane groups as the byte target allows, and never fewer than one.
+    let rows_per_call = rows_per_call(row_bytes, H::LANES);
```
Nashtare: `let rows_per_call = rows_per_call(row_bytes, H::LANES);` shadows the function with a binding of the same name. It compiles (the RHS resolves before the binding exists, and values and functions live in different namespaces), but from the next line on `rows_per_call` in this scope means the number, and anyone adding a second call has to un-shadow first. `rows_per_group` or `rows` for the local reads better and costs nothing.

merkle-tree/src/merkle_tree.rs
```diff
@@ -1372,4 +1409,53 @@ mod tests {
...
+        //
+        //     16 KiB / 64 B = 256 rows, over 8 lanes = 32 whole groups
+        assert_eq!(rows_per_call(64, 8), 256);
+        assert!(rows_per_call(64, 8) * 64 <= ROW_SCRATCH_BYTES);
+
+        // A row past the target still stages one whole lane group.
+        assert_eq!(rows_per_call(ROW_SCRATCH_BYTES + 1, 8), 8);
+
+        // A hasher reporting no lanes is treated as one message at a time.
+        assert_eq!(rows_per_call(64, 0), rows_per_call(64, 1));
```
Nashtare: Two gaps in an otherwise good sweep.

**The sweep does not cover the composed claim.** `rows_per_call` is asserted against `max(target, lanes * row_bytes)`, but the thing the constant's doc is about is the *buffer*, which is `Vec::with_capacity(rows_per_call * total_width)` **elements of `F`** — and `row_bytes` at the call site is `(total_width * size_of::<F>()).max(1)`, so the two coincide only when `total_width > 0`. For a set of zero-width matrices the bound is trivially satisfied and the test says nothing. Minor, but the whole PR is about the doc matching the code, and the doc is about bytes staged, not about the helper's return value.

**Is the bound assertion non-vacuous?** It reads `rows * row_bytes <= ROW_SCRATCH_BYTES.max(effective * row_bytes)`, which is satisfied automatically whenever `rows == effective` (the one-lane-group case) and follows from the floor division whenever `rows > effective`. The PR description says it was mutation-checked with "rounding before the multiply instead of after ... fails at `lanes=2, row_bytes=16383`" — but `(16384 / 16383).max(1) * 2 == 2` is the same answer the shipped expression gives, so that mutation does not distinguish them. A mutation that *does* fail the bound (e.g. doubling the group) would be worth naming instead, so the next reader can see the assertion has teeth.

merkle-tree/src/merkle_tree.rs
tcoratger: Confirmed both halves and rewrote the rationale.

I checked the bail-out rather than taking it: `symmetric/src/serializing_hasher.rs:79-85` returns through `hash_iter` per row once `row_bytes > ROW_BYTES_SCRATCH` (8 KiB), and `LANES > 1` in the tree comes only from `keccak/src/lib.rs:184` forwarded through `SerializingHasher`, gated at `merkle_tree.rs:685`. Everything else in the tree takes the trait default of 1.

Working out where the two budgets actually cross, at 16 KiB target:

| lanes | first row width that overshoots | still grouping there? |
|---|---|---|
| 1 | 16385 B | no |
| 2 | 8193 B | no |
| 4 | 4097 B | **yes** |
| 8 | 2049 B | **yes** |

So the overshoot has two regimes, not one, and your sentence is right about which is which. At 8 lanes: 2 KiB rows stage 16 KiB (1x) and group; 8 KiB rows stage 64 KiB (4x) and still group; 64 KiB rows stage 512 KiB (32x) and every lane is idle. That table is now the doc, with the occupancy argument scoped to the first two rows and the third named as pure copy cost.

I did not change the behaviour. This layer sees `total_width * size_of::<F>()` and has no way to know the consumer's 8 KiB budget, so telling the regimes apart means coupling the two constants. The doc says that instead of pretending the trade is uniform.

Your second point was also wrong in the doc and is gone: the consumer re-stages into its own 8 KiB buffer (`serializing_hasher.rs:105`), so it never reads this one. What is true is narrower, and is what the line now says: the copy stays in L1 until the pass that reads it back.

merkle-tree/src/merkle_tree.rs
tcoratger: Both correct, both now in the doc.

`digests.chunks_mut(rows_per_group)` at `merkle_tree.rs:434` is a plain chunking, so the last group carries `digests.len() % rows_per_group` and need not be a multiple of the lane count. The parallel arm rounds `task` to whole groups (`:459`), so only that tail is short, and for a level whose height is not a multiple of the group there is always one. The sentence now reads that a level is cut into groups of this size, that only its last group is ever short, and that it runs partly idle on the remainder.

The per-task point is on the constant, since that is where someone reads for buffer size:

```
/// The bound above is per call.
///
/// One buffer lives per parallel task.
///
/// A level's peak is therefore that bound times the number of tasks.
```

Your 512 KiB at 8 lanes and a 64 KiB row is the third row of the new regime table, so the two statements sit next to each other and the multiplication is there to make.

merkle-tree/src/merkle_tree.rs
tcoratger: You are right, and I inverted it rather than documenting a precondition.

`row_bytes` was used for exactly one thing — the call to this helper — so the `.max(1)` at the old `:415` was normalization living forty lines from the division. It moved into the function, next to the `lanes` guard, and the call site is now a plain `total_width * size_of::<F>()`. One normalization, in the place that divides, no precondition to state and no panic path.

```rust
fn rows_per_call(row_bytes: usize, lanes: usize) -> usize {
    // Both divisors are normalized here rather than at the call site, so no caller can
    // reach the division by zero that a zero-width row or a zero-lane hasher would cause.
    let row_bytes = row_bytes.max(1);
    let lanes = lanes.max(1);

    (ROW_SCRATCH_BYTES / row_bytes / lanes).max(1) * lanes
}
```

The `# Returns` section says it in one line — zero on either argument is read as one — so the contract is symmetric and total instead of guarding the impossible input and leaving the reachable one open.

The test that froze the zero-lane contract on its own is gone. Zero is now just another entry in both sweep axes, checked by the same two assertions as every other value, which is what it should have been.

merkle-tree/src/merkle_tree.rs
tcoratger: Renamed to `rows_per_group`, at the binding and its three uses (`merkle_tree.rs:427`, `:431`, `:434-435`, `:459`). No shadowing left.

merkle-tree/src/merkle_tree.rs
tcoratger: Both fair. The second one especially — I checked and the mutation I claimed does not distinguish anything.

**On the vacuity.** You are right that `(16384 / 16383).max(1) * 2 == 2` is the shipped answer, so that mutation proves nothing. I swept four mutations over the test's own inputs to find ones that actually bite:

| mutation | differs from shipped | fails alignment | fails bound |
|---|---|---|---|
| drop the lane rounding | 20/54 | **yes** | no |
| round the target up, not down | 4/54 | no | **yes** |
| one extra lane group | 54/54 | no | **yes** |
| round before the multiply | 16/54 | no | **yes**, but at `row_bytes=1`, not 16383 |

Run against the real test, not computed:

```
drop the lane rounding:
  assertion `left == right` failed: lanes=2 row_bytes=16383 rows=1
    left: 1   right: 0

round the target up (div_ceil):
  lanes=0 row_bytes=8191 staged 24573 bytes
```

The ceil one is the interesting one: it differs on only 4 of 54 inputs and needs a width just under half the target to show up. That is why `ROW_SCRATCH_BYTES / 2 - 1` earns its place in the sweep, and both numbers are now written above the loop so the next reader can see what holds it up.

**On the composed claim.** Added `the_staged_buffer_meets_the_bound_the_target_states`, which builds both sides from a width the way the driver does — `row_bytes = total_width * ELEMENT` and `staged = rows * total_width * ELEMENT` — over widths straddling the target (4096 four-byte columns are exactly 16 KiB) and asserts zero-width matrices stage nothing rather than passing trivially. Doubling the buffer expression fails it at `lanes=1 total_width=1 staged 32768 bytes`, so it is not vacuous either.

Verification: `cargo test -p p3-merkle-tree` 114 passed, clippy `-D warnings` clean, `cargo +nightly fmt --check` clean.

tcoratger: Correction to one cell of that table, since this thread is about a number being wrong.

The counts were computed against the sweep before I added `row_bytes = 0` to it. Recomputed over the 54 inputs the test actually runs, "round before the multiply" differs on 20/54 and fails the bound on 20, not 16/54. The other three rows are unchanged, and no verdict moves.

Nashtare (approved):

PR #2073, opened 2026-09-08
feat(whir)!: seed the WHIR transcript through the shared Fiat-Shamir layer

tcoratger:
## What this changes

WHIR shipped a domain separator of its own. `whir::fiat_shamir` packed a `(tag, sub-label, count)` triple into one field element per transcript step, absorbed the resulting vector into the challenger, and called that the shape binding. That is the idea behind the `p3-challenger` `fs` layer, written a second time in a second vocabulary. This deletes it and derives the binding from the shared layer instead.

WHIR was the only protocol in the repository with a shape-binding mechanism of its own, so this removes a parallel implementation rather than adding a second one.

## The problem

The separator was never wired into the protocol.

`observe_domain_separator` was the caller's job, and **every call site of it in the repository is a test, an example, or a bench**. Nothing in the `Pcs` implementation absorbed one, and nothing checked that anybody had.

An integrator wiring up `WhirPcs` therefore got a prover and a verifier that agree with each other — so no test fails — and a transcript bound to nothing but the data itself. No configuration parameter reached the sponge, and neither did the protocol's identity.

## The idea

`WhirShape` and `ZkWhirShape` derive every number that fixes a run from the derived configuration, publish it as an `InteractionPattern`, and seal it into the shared `DomainSeparator`.

```text
    WhirConfig  ->  WhirShape  ->  InteractionPattern  ->  DomainSeparator  ->  seed
```

`WhirPcs` and `HidingWhirPcs` seed from that themselves, at the point where the WHIR run begins and ahead of its first challenge. Absorbing the seed is no longer something a caller can skip.

## Per-parameter triage

The pattern fingerprint covers everything that moves the step sequence: the round count, each round's out-of-domain sample count, every grinding difficulty, the width and number of every query index, the sumcheck round counts the folding schedule derives, and the final polynomial's length.

The instance label carries what the fingerprint reaches only indirectly.

| Parameter | Verdict |
| --- | --- |
| `num_variables` | label |
| `security_level` | label — reaches the shape only through derived query counts |
| `pow_bits` (grinding budget) | label — a cap on the derived difficulties, need not change any |
| `starting_log_inv_rate` | label — code rate |
| `round[i].log_inv_rate` | label — code rate |
| `soundness_type` | label — the proximity-gap assumption |
| `folding_factor` (variant and values) | label — two variants alias onto one schedule |
| `commitment_ood_samples` | label — drawn before the described region |
| `round_parameters.len()` | fingerprint — round count |
| `round[i].ood_samples` | fingerprint — two steps per sample |
| `round[i].pow_bits`, `round[i].folding_pow_bits` | fingerprint — grinding difficulty |
| `round[i].num_queries`, `round[i].domain_size` | fingerprint — draw count and index width |
| `folding_schedule` | fingerprint — sumcheck round counts |
| `starting_folding_pow_bits` | fingerprint |
| `final_queries`, `final_pow_bits` | fingerprint |
| `final_sumcheck_rounds`, `final_folding_pow_bits` | fingerprint |

`security_level`, the rates and the assumption are what the entire soundness argument is stated in terms of. They reach the step sequence through the query counts they derive, and that arrow is not injective: two parameter sets landing on one query count would otherwise be indistinguishable.

The folding strategy is the sharper case. `Constant(4)` and `ConstantFromSecondRound(4, 4)` derive the same schedule at 16 variables, so every derived number agrees and the two share a step sequence exactly.

### HVZK parameters

| Parameter | Verdict |
| --- | --- |
| `zk.ell_zk` | fingerprint — wire width of a masked sumcheck round |
| `zk.mask_log_inv_rate` | label — mask distance, reaches the shape only via domain sizes |
| `mask_queries` | fingerprint — spot checks per group |
| `final_queries`, `final_pow_bits` (hiding pair) | fingerprint — draw count and grinding difficulty |
| `oracle_randomness[i]` | label — only the terminal budget reaches a step length |
| `sumcheck_mask`, `switch_masks[i]` | label — all three lengths of every mask code |
| `inner.commitment_ood_samples` | label — the hiding commit phase draws none |
| `inner.final_sumcheck_rounds`, `inner.final_folding_pow_bits` | label — the base case replaces them |
| `inner.final_queries`, `inner.final_pow_bits` | label — shadowed by the hiding pair |

## What is bound, and what is not

Bound: the protocol identity, the full step sequence, every count and width and difficulty above, and every configuration scalar the sequence does not already pin.

Not bound: hints. A hint never enters the sponge, so it cannot move a challenge, and WHIR carries every hint inside its own serde proof, where the verifier length-checks it against the configuration before use. A Merkle multiproof's length is a function of the tree rather than of the parameters, so a pattern could not pin it in any case.

## Tests

Two tests walk **every field** of `WhirConfig` and of `ZkWhirConfig`, perturb it, and assert the first challenge the seed produces moves. That walk is the acceptance criterion for deleting the old separator, and it is what caught the four plain-pipeline fields the hiding run leaves unread.

An end-to-end test proves the aliasing case reaches the wire: a proof produced under `Constant(4)` is rejected by a verifier configured with `ConstantFromSecondRound(4, 4)`. Every derived number of those two agrees, so before this the two configurations interoperated silently.

## Along the way

The transcript checkpoint is gone. Prover and verifier each drew a field element between grinding and query sampling and threw it away, purely so the old separator's per-step counts stayed in step with the code. Nothing read it.

Transliterating the old separator surfaced three places where it described a transcript the code does not run:

- its sumcheck put the grinding step after the folding challenge, where the code grinds before;
- its out-of-domain phase sampled every point and then observed every answer, where the code alternates one for one;
- its base-case reveals were one step per word, where the code absorbs the message half and the randomness half separately.

None of it mattered while the separator was a blob absorbed up front, which is exactly the point: a description nobody replays is a description nobody checks.

## What is left

This does not yet drive the run through `ProverState` and `VerifierState`. WHIR's sumcheck, its masked sumcheck and its base case all absorb through `p3-sumcheck`, which still speaks to the raw challenger, so the drivers cannot reach those steps until that crate moves. The pattern built here is the one those drivers will play; the player's enforcement is the part still to come.

**Nothing is added to `p3-challenger`.**

## Breaking changes

- **The WHIR transcript changes.** Proofs produced before this do not verify after it.
- `multi-stark/tests/fixtures/multi_stark_whir_v0_8_0.postcard` is regenerated. It is the only fixture that routes through this transcript; the two-adic, circle and STIR fixtures never enter it.
- `whir::fiat_shamir` and the `DomainSeparator` it exported are removed.
- `WhirProver::add_domain_separator` and `HidingWhirPcs::add_domain_separator` are removed. Callers no longer seed by hand; delete the call.
- The multilinear PCS implementations now require `F: PrimeField64`, which every field they are instantiated with already satisfies.

## Verification

```
cargo test -p p3-whir --lib --tests --examples        # 138 pass, 1 ignored
cargo test -p p3-multi-stark --lib --tests --examples # 77 + 10 + 4 + 6 + 8 pass
cargo test -p p3-challenger --all-targets             # 182 pass
cargo clippy -p p3-whir -p p3-multi-stark --all-targets -- -D warnings  # clean
cargo +nightly fmt --all -- --check                   # clean
cargo build --workspace --all-targets                 # clean
```

`--all-targets` additionally *runs* the `harness = false` criterion benches (`fri_vs_whir` at 2^22, `zerocheck`) in a debug build, which does not terminate in reasonable time on this machine. Those targets are covered by `cargo build --workspace --all-targets` above; CI runs them under `--release`.


tcoratger:
Superseded by #2080, which carries the same change with its branch in this repository so the stack chains properly. Closing this one to avoid a duplicate.

tcoratger:

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2076, opened 2026-09-08
feat(uni-stark)!: drive the uni-STARK transcript through the typed Fiat-Shamir layer

tcoratger:
Depends on #2084. This PR targets `feat/typed-transcript-whir` so the diff contains only this crate's migration; retarget to `main` after the branches below it land.

> **Stacked on #2068** (`feat/typed-transcript-fri-pcs`) — review and merge that one first.
>
> The base branch lives on the fork, so GitHub cannot use it as the base of a PR in this repo; the base here is `main` and the diff therefore shows #2068's commit as well. The commit belonging to this PR is the second one, `feat(uni-stark)!: drive the uni-STARK transcript through the typed Fiat-Shamir layer`. Once #2068 lands, this diff reduces to that commit alone.

## What changes

The uni-STARK wrote its transcript by hand on both sides: three instance numbers, the trace and preprocessed commitments, the public values, then `alpha`; the quotient and randomization commitments, a grind, then `zeta`. Nothing checked that the prover's sequence and the verifier's agreed, and the shape of the run was bound only by whatever happened to be absorbed.

It now derives both challenges through `ProverState` / `VerifierState`, over a description each side builds from its own configuration and its own AIR. **This is where the top-level driver is created**: the FRI PCS below it, and the low-degree test below that, are already described child transcripts.

New `uni_stark::transcript` holds that description plus one driver per side.

```text
    trace commitment           one opaque value
    preprocessed commitment    only when the AIR has preprocessed columns
    public values              num_public_values base elements
    alpha                      one extension element
    quotient commitment        one opaque value
    randomization commitment   only when the PCS is zero-knowledge
    out-of-domain grinding     only when the difficulty is positive
    zeta                       one extension element
    Begin  opening argument    bracket around the delegated run
    End    opening argument
```

## Headline: the AIR's shape now reaches the seed, and the TODO is deleted

`uni-stark/src/verifier.rs` carried this:

```rust
// TODO: Might be best practice to include other instance data here in the transcript, like some
// encoding of the AIR. This protects against transcript collisions between distinct instances.
// Practically speaking though, the only related known attack is from failing to include public
// values. ...
```

`StarkShape` reads five numbers straight off the AIR:

| from the AIR | |
| --- | --- |
| `main_width` | `air.width()` |
| `num_public_values` | `air.num_public_values()` |
| `num_periodic_columns` | `air.num_periodic_columns()` |
| `opens_main_next_row` | `air.main_next_row_columns()` |
| `opens_preprocessed_next_row` | `air.preprocessed_next_row_columns()` |

and takes the rest from the configuration: both trace heights, the preprocessed width in force, the quotient chunk count, the zero-knowledge setting, and the out-of-domain grinding difficulty.

The public-value count is the `Length::Fixed` of its step, and the two optional commitments are present or absent, so those ride the **pattern fingerprint**. Everything else goes in the **instance label**. Either way, every one of the eleven numbers moves the seed — the TODO is deleted rather than restated.

Two tests prove it:

- `two_airs_differing_only_in_shape_seed_differently` (unit) — three real `BaseAir` impls differing in one field each, through the real shape constructor, all landing on different first challenges.
- `an_air_of_a_different_shape_does_not_verify` (end to end) — two AIRs asserting the same constraint over the same trace, differing only in a periodic-column count neither reads. The proof is structurally valid for both, so nothing rejects it on shape; only the seed separates them.

Plus `every_field_of_the_shape_reaches_the_seed` and `no_two_one_step_mutations_collide`, the R4 walk over all eleven fields.

## What is bound, and what is not

- **Fingerprint**: the public-value count, the presence of each optional commitment, the grinding difficulty.
- **Instance label**: both trace heights, the trace width, the preprocessed width, the periodic-column count, the quotient chunk count, and which traces are opened at the next row.
- **Nothing**: the constraint polynomials the AIR evaluates. Two AIRs of identical shape whose constraints differ share a seed. The verifier evaluates the AIR it holds at `zeta`, so an AIR other than that one is not a transcript question at all.

## The three instance numbers stop being absorbed as messages

Both trace heights and the preprocessed width were observed as base elements at the top of the run. They are now part of the seed, alongside every other number of the shape, which makes the shape one commitment rather than a prefix of absorbs a reader checks by eye.

## Public values are a `Kind::Public` step

They never travel on the wire: the verifier holds them as its own argument and absorbs them exactly as the prover did. Their count is the AIR's, and the caller's slice is checked against it *before* the driver exists.

## Delegation is a bracket

`Begin Protocol("opening_argument")` and its matching `End` sit around the region where the STARK lends the sponge to `Pcs::open` / `Pcs::verify`, which enforce their own pattern inside, under their own seed. This is the primitive #2068 added; nothing new here.

## R5: no early return drops an unfinalized driver

Enumerating every `?` and every `return Err` between `StarkVerifierTranscript::new` and `finish`:

| site | why it is safe |
| --- | --- |
| `transcript.constraint_phase(...)?` | The driver's own error. A missing or unexpected preprocessed commitment calls `state.abort()` before returning. |
| `transcript.ood_phase(...)?` | Same for the randomization commitment. The grinding failure comes from `VerifierState::observe_pow`, which calls `player.abort()` before returning its error. |
| `prepared?` (inside `delegate`) | Returns from the closure, not from `verify_with_preprocessed`. `delegate` records `End` on any outcome, and `finish` still runs. |
| `pcs.verify(...)?` (inside `delegate`) | Same. |
| `checked?` | After `finish`, so the driver is already consumed. |

There is nothing else on that span. Every `zeta`-dependent check that can reject — `OodPointInDomain`, the periodic-column length check, `NextPointUnavailable`, `RandomizationError`, `QuotientDomainsCountMismatch` — moved into `prepare_opening_claims`, whose `Result` is **bound, not propagated**, and unwrapped inside the delegated closure.

## R6: nothing added to `p3-challenger`

This PR adds no API to `challenger/src/fs`. It uses the delegation bracket and `challenger_mut` exactly as #2068 added them. Sibling PRs in flight will not conflict here.

## Other changes

`StarkGenericConfig` gains one bound: the PCS domain's base field is a prime field below `2^64`, which is what the transcript layer knows how to encode and absorb. Putting it on the trait rather than on `prove` and `verify` leaves generic callers untouched — nothing outside `uni-stark` needed a change.

`observe_commitment` in `config.rs` is gone. It existed to disambiguate two `CanObserve` impls in method-call position; the driver's `observe_opaque` is a generic call and has no such problem.

## Fixtures

**All three** uni-STARK compatibility fixtures are regenerated — two-adic, circle, and STIR. Seeding now absorbs a protocol identity and a shape fingerprint before anything else, so every proof produced before this is invalid. The two-adic one was already regenerated by #2068 and moves again here. The circle and STIR ones move too, because the change is in the STARK rather than in any one PCS. (`batch-stark` and `multi-stark` have their own provers and their fixtures are untouched.)

## Soundness mutation coverage

One test per absorbed value:

| absorbed value | test |
| --- | --- |
| trace commitment | `transcript_binding::a_tampered_trace_commitment_is_rejected` |
| preprocessed commitment | `mul_fib_pair::test_tampered_preprocessed_fails` (existing) |
| public values | `fib_air::test_incorrect_public_value` (existing) |
| quotient commitment | `transcript_binding::a_tampered_quotient_commitment_is_rejected` |
| randomization commitment | `transcript_binding::a_tampered_randomization_commitment_is_rejected` |
| out-of-domain grinding witness | `grinding::tampered_ood_pow_witness_is_rejected` (existing) |
| trace height (seed) | `transcript_binding::a_tampered_degree_bits_is_rejected` |
| AIR shape (seed) | `transcript_binding::an_air_of_a_different_shape_does_not_verify` |

## Breaking

The uni-STARK transcript changes, so proofs produced before this do not verify after it. `StarkGenericConfig` now requires the PCS domain's base field to implement `PrimeField64`, which every configuration in the workspace already satisfies.

## Verification

```
cargo test -p p3-uni-stark --all-targets    43 + 13 + 8 + 10 + 3 + 2 + 3 + 3 + 1 + 3 + 5 + 7 passed, 0 failed
cargo test -p p3-fri -p p3-challenger --all-targets   182 + 68 + 2 + 26 passed, 0 failed
cargo clippy -p p3-uni-stark --all-targets -- -D warnings   clean
cargo +nightly fmt --all -- --check   clean
cargo build --workspace --all-targets   clean
```

---

2. #2082 - Circle PCS
3. #2083 - STIR
4. #2084 - WHIR
5. #2076 - uni-STARK
6. #2077 - batch-STARK


tcoratger:

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2077, opened 2026-09-08
feat(batch-stark)!: drive the batch-STARK transcript through the typed Fiat-Shamir layer

tcoratger:
Depends on #2076. This PR targets `feat/typed-transcript-uni-stark` so the diff contains only this crate's migration; retarget to `main` after the branches below it land.

> **Stacked on #2068.** Review and merge that one first, then this. The intended base is `feat/typed-transcript-fri-pcs`, but that branch lives in the fork and GitHub only accepts a base branch that exists in this repository, so the PR is opened against `main` and its diff currently shows #2068's commit as well. Read only the second commit, `feat(batch-stark)!: ...`; once #2068 merges, this branch rebases and the diff collapses to that one commit.

## No changes to `p3-challenger`

**This PR adds nothing to `challenger/src/fs`.** Sibling PRs in this rollout are in flight and touch that directory; this one does not, so it cannot conflict there. Everything it needs — `Kind::Public`, `TypeTag::Opaque`, `observe_pow`, `abort`, `challenger_mut` — already exists from #2035 and #2068.

## What changes

Batch-STARK wrote its transcript by hand, behind a wrapper that held the raw challenger and offered a dozen methods over it. Those methods absorbed the per-instance shape numbers, the four commitments, the public values and the lookup terminals, ground twice, and sampled four challenges. Nothing checked that the prover's sequence and the verifier's replay were the same sequence; the two were kept in step by reading the two functions side by side.

It now derives every challenge through `ProverState` / `VerifierState`, over a description both sides build from the AIRs, the keygen-cached common data, and the configuration.

New `batch_stark::transcript` holds that description plus one driver per side. `BatchShape` is the description; `BatchProverTranscript` and `BatchVerifierTranscript` walk it. Both sides make the same calls in the same order, so a prover phase cannot drift from the verifier's replay.

## The described shape

```text
    per instance:  degree bits          one extension element
    main commitment                     one opaque value
    per instance:  public values        that AIR's public-value count
    preprocessed commitment             only when the batch has preprocessed columns
    when the batch declares a lookup:
        lookup grinding                 at the configured difficulty
        lookup alpha                    one extension element
        lookup beta                     one extension element
        permutation commitment          one opaque value
        lookup terminals                one per instance that declares a lookup
    constraint challenge                one extension element
    quotient commitment                 one opaque value
    randomization commitment            only when the PCS hides
    out-of-domain grinding              at the configured difficulty
    zeta                                one extension element
    Begin  opening argument             bracket around the delegated run
    End    opening argument
```

## What is bound, and where

| Number | Where it lands |
| --- | --- |
| instance count | fingerprint (step count) and instance label |
| public-value count per instance | fingerprint, as the declared length of each public step |
| main-trace width per instance | instance label |
| preprocessed width per instance | instance label |
| preprocessed / lookup / randomization presence | fingerprint, as the presence of their steps |
| number of lookup terminals | fingerprint, as the count of terminal steps |
| lookup and out-of-domain grinding difficulties | fingerprint, as `Length::Fixed(bits)` |
| extended degree bits per instance | absorbed as a message, because the prover chooses them |

Not bound: a commitment's width, which this layer cannot see. The opening check that recomputes it is what rejects a wrong one. Also not bound here: the PCS parameters, which the opening argument's own seed binds inside the bracket.

## The structural numbers stop being absorbs

The instance count, every main-trace width, every preprocessed width and every public-value count used to travel through the sponge as field elements. All four come from the AIRs and the common data, which both sides hold before the run, so they are configuration rather than data. Absorbing them on top of the seed binds nothing the seed does not already bind.

Two numbers go away entirely. The base degree bits and the quotient chunk count were absorbed per instance, and both are functions of the extended degree bits and the configuration: the base bits are the extended bits less the ZK flag, and the chunk count is what the symbolic degree inference returns for that AIR at that height.

Public values become a `Kind::Public` step. They are the caller's own argument on both sides and never travel in the proof, which is exactly what that kind records: the value is bound into the sponge and nothing is written to the wire.

## Both grinding difficulties now reach the seed

Before, a difficulty mismatch was caught in one direction only. A verifier demanding more work than the prover paid failed the witness check, but a verifier demanding less accepted the proof and no one learned that the two configurations disagreed. As `Length::Fixed(bits)` on a described step, the difficulty now rides the fingerprint and the two sides land on different seeds instead, whichever way the difference runs.

The out-of-domain step is described at every difficulty, zero included, because the proof always carries a witness for it. The lookup step is described only when some AIR declares a lookup, which is exactly when the proof carries the optional lookup witness.

## Delegation is a bracket

`Begin Protocol("opening_argument")` and its matching `End` sit around the region where the batch lends the sponge to `Pcs::open_with_preprocessing` and `Pcs::verify`, which enforce their own pattern inside, under their own seed. The markers absorb nothing; what they record is that a sub-protocol runs there, and where. The delegated run is taken as a closure, so the bracket cannot be left open.

## R5: no exit drops an unfinalised driver

Both drivers panic on drop when the pattern was not fully replayed, and a panic raised while another unwinds aborts the process. Here is every exit between the verifier driver being built and `finish`, in `verify_batch`:

| Exit | How the check is released |
| --- | --- |
| `transcript.lookup_phase(...)?` — batch has no lookups but the proof carries a witness | `state.abort()` before the error is returned |
| `transcript.lookup_phase(...)?` — batch has lookups and the proof carries no witness | `state.abort()` before the error is returned |
| `transcript.lookup_phase(...)?` — witness below its difficulty | `VerifierState::observe_pow` aborts the player before returning `Err` |
| `transcript.ood_phase(...)?` — witness below its difficulty | `VerifierState::observe_pow` aborts the player before returning `Err` |
| `commitments_with_opening_points(...)` rejects a proof shape | matched rather than propagated; `transcript.abort()` runs first |
| `pcs.verify(...)` rejects the opening | the result is held, `transcript.finish()` runs, then the error is propagated |

There are no other `?` operators and no other `return Err` on that span. Every proof-carried count the description relies on — the instance count, the per-instance shapes, the terminal presence, the lookup-commitment presence — is validated *before* the driver is built, so a malformed proof reaches its rejection with no live driver to drop.

## Error types

`BatchTranscriptFailure` replaces `InvalidLookupPow` and the unit `InvalidOodPowWitness`, and covers both grinding steps under one `thiserror` enum. Each rejected witness names the difficulty its step required, so a difficulty mismatch reads as one in the error rather than as a bare "invalid witness".

## Tests

- `every_configuration_knob_reaches_the_seed` walks one mutation per `BatchShape` field and asserts the first sampled challenge moves for each (R4).
- `the_width_of_every_instance_reaches_the_seed` and `the_number_of_lookup_terminals_reaches_the_seed` pin the two cases where a reordering or a count shares a step sequence.
- `the_delegation_bracket_leaves_the_sponge_untouched` runs both drivers over one shape and checks that the challenges agree and the sponges stay in step past `finish`.
- Rejection tests for the three lookup-grinding failures and the out-of-domain one.
- New mutation tests in `tests/simple.rs` for each absorbed value not already covered: the main, quotient, permutation and randomization commitments, and a tampered degree bit (R3). The existing suite already covers the public values, the preprocessed commitment, the lookup terminals and both grinding witnesses.

## Fixtures

**Both compatibility fixtures are regenerated.** Seeding now absorbs a protocol identity and a shape fingerprint before anything else, so every proof produced before this is invalid. Unlike the two FRI transcript PRs, the circle fixture changes too: the transcript that moved here is the batch's own, and it wraps the circle PCS exactly as it wraps the two-adic one.

## Breaking changes

- The batch-STARK transcript changes, so proofs produced before this do not verify after it.
- `prove_batch` and `verify_batch` now require `Val: PrimeField64`.
- `BatchTranscript` and `InvalidLookupPow` are gone, replaced by `BatchProverTranscript`, `BatchVerifierTranscript`, `BatchShape` and `BatchTranscriptFailure`.
- `BatchVerificationError::InvalidLookupPow` and `BatchVerificationError::InvalidOodPowWitness` are replaced by `BatchVerificationError::Transcript`.

---

2. #2082 - Circle PCS
3. #2083 - STIR
4. #2084 - WHIR
5. #2076 - uni-STARK
6. #2077 - batch-STARK


tcoratger:

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2079, opened 2026-09-08
feat(stir)!: drive the STIR transcript through the typed Fiat-Shamir layer

tcoratger:
> Part of the typed Fiat-Shamir rollout, stacked on `feat/typed-transcript-circle`. Review and merge in chain order.

STIR wrote its transcript by hand on both sides. Per round: grind, sample a folding challenge, absorb a commitment, sample out-of-domain points, absorb their answers, grind again, sample a combination challenge and the query indices, absorb the answer polynomial, sample its consistency challenge. Then the final send, twice more grinding, and the final query indices. Around sixty operations per side, in two loops kept in step by hand.

Nothing checked that the two loops agreed. The shape of a run was bound only by what happened to be absorbed, and the batched path had a further hole: the shared grinds and the per-instance blocks left no record of which instances were active where.

It now derives every challenge through `ProverState` and `VerifierState`, over a description built from the round schedule.

## The idea

New `stir::transcript` holds that description plus one driver per side. Both sides make the same calls in the same order, so a prover loop cannot drift from a verifier loop.

One description covers both entry points. A single run is the batch of size one, so `prove_stir` and `prove_stir_multi` over one instance now agree step for step, and there is one thing to read rather than two.

Each instance's block inside a phase is bracketed by its own container:

```text
Begin instance_fold  fold_challenge  round_commitment  End
Begin instance_fold  fold_challenge  round_commitment  End
```

Two instances contributing one step each and one instance contributing two flatten to the same sequence. The brackets keep them apart, so the fingerprint distinguishes two batchings of the same instances.

## What is bound

The description lands in the seed as a fingerprint, which binds every round count, every grinding difficulty, every query count and its bit width, and every out-of-domain count.

The parameters that leave the step sequence untouched go in the instance label: the rate, the proximity-gap regime, the security target, the grinding cap, every folding factor, every domain shift, and every round's degree and eta.

The rate and the regime are what the soundness argument is stated in terms of, and they reach the step sequence only through the query counts they imply. Two parameter sets landing on the same counts would otherwise share a seed, so the label carries both directly. `the_rate_and_the_proximity_regime_reach_the_seed` asserts that the two runs share a pattern and not a seed; `every_configuration_knob_reaches_the_seed` walks all 27 numbers of the shape one at a time.

## What is not bound

The compact-answers option. It chooses how the answer polynomial travels, not what enters the sponge: the verifier rebuilds exactly the coefficients the prover absorbed. Two sides configured differently already fail, because the answer step binds its own length. Leaving it out is what keeps the two encodings of one run comparable, which `compact_answers_preserve_pcs_values_and_warm_clone_transcripts` pins. Say the word if you would rather it were bound and I will move it into the label.

A commitment's width, which this layer cannot see, exactly as in FRI.

## The discarded rho

The prover sampled the answer-consistency challenge into `let _rho` so the two sides stayed aligned; only the verifier reads its value. The alignment held because someone remembered it. It is now a recorded step both sides play, and the player reports a diff at the first divergence.

## The answer polynomial is a bounded step

Its length is the number of points the round interpolates, which the query draw settles by deduplicating collisions, so only the cap is known when the pattern is built. The step records `num_ood_samples + num_queries` and binds the actual count as a length prefix, which also makes that part of the transcript prefix-free.

## Query indices are an unbiased step

`sample_bits` carries a `2^width / p` modular bias, which over a 31-bit prime and a 20-bit index is `2^-11`, so STIR draws through rejection sampling. The pattern now records that choice rather than leaving two different distributions sharing one shape.

## Lengths are checked before the description exists

The round count, each round's out-of-domain answer count, each answer polynomial against its cap, the final polynomial's coefficient count, and, in the batched path, the replicated grinding witnesses that a shared grind checks once. Nothing that feeds a pattern step is read from a proof.

## One release point for the completeness check

Both drivers panic on drop when the pattern was not fully replayed, so every early return between construction and `finish` has to release that check first. The replay is a function of its own, and its caller finalizes on success and aborts on failure:

```rust
match verify_stir_replay(config, proof, &mut transcript, external_fibers) {
    Ok(outputs) => { transcript.finish(); Ok(outputs) }
    Err(err)    => { transcript.abort();  Err(err) }
}
```

Every `?` on that span travels through the same match, so no path can drop an unfinalised driver and turn a rejection into a panic. That is two constructions and two matches in the whole crate. The span they cover holds 26 early-exit sites across `verify_round`, `verify_final_round` and the two replay bodies, and no other function in the crate constructs a driver.

Four sibling PRs are in flight against this module, so these are called out for conflict triage:

- `TypeTag::UniformBits` for a bit challenge drawn without modular bias, with `Interaction::uniform_bits` to build one. A biased draw and a rejection-sampled one answer the same question with different distributions, so they are different shapes.
- `observe_extensions_bounded` on both drivers, for a caller-carried list whose length the run settles rather than the configuration. The count enters the sponge as a big-endian prefix ahead of the values, which keeps the step prefix-free.
- `challenge_extensions_rejecting` on both drivers, for a challenge constrained to avoid a set the protocol fixes in advance. The number of candidates a draw consumes is not known when the pattern is built, so the loop lives inside one step.
- `challenge_uniform_bits` on both drivers.

Nothing else in `p3-challenger` moves.

## Other changes

`StirError::InvalidPowWitness` now names the site it guards and the difficulty it required, rather than only the round.

`utils::OodFilter` is the out-of-domain acceptance rule, lifted out of `sample_ood_points` so the transcript driver owns the sampling loop and the rule has one definition.

## Breaking changes

- The STIR transcript changes, so proofs produced before this do not verify after it.
- Every STIR entry point now requires `F: PrimeField64`, which every field in the workspace already satisfies.
- `StirError::InvalidPowWitness` gains a `stage` and a `bits` field.
- `uni-stark/tests/fixtures/uni_stark_stir_v0_8_0.postcard` is regenerated. It is the only fixture that routes through the STIR verifier; the two-adic, circle and WHIR fixtures never enter this path and verify unchanged.

## Verification

```
cargo test -p p3-stir --lib              # 79 passed
cargo test -p p3-stir --test stir        # 107 passed
cargo test -p p3-stir --test utils       # 28 passed
cargo test -p p3-stir --all-targets --no-run   # all five targets build
cargo test -p p3-challenger --all-targets      # 183 passed
cargo clippy -p p3-stir --all-targets -- -D warnings   # clean
cargo +nightly fmt --all -- --check      # clean
cargo build --workspace --all-targets    # clean
cargo test -p p3-uni-stark -p p3-batch-stark -p p3-multi-stark --tests  # all passed
```

`cargo test -p p3-stir --all-targets` also launches the two criterion bench binaries,
which prove and verify at degree `2^18` under the debug profile. I stopped that run at
about ninety minutes rather than let it hold the branch; the three test targets above
are every target that asserts anything, and `--no-run` shows the bench binaries build.

---

**Stack** (merge top to bottom):

2. #2076 - uni-STARK
3. #2077 - batch-STARK
4. #2078 - Circle PCS
5. #2079 - STIR
6. #2080 - WHIR


tcoratger:
**This PR was never merged.** GitHub marked it merged automatically when a sibling branch in the stack was force-pushed while it still contained these commits, and it deleted the head branch as a side effect. No part of this change reached `main`.

The stack has been re-ordered bottom-up so the PCS layer sits below the STARKs that consume it. This change is now open as part of that chain:

2. #2082 - Circle PCS
3. #2083 - STIR
4. #2084 - WHIR
5. #2076 - uni-STARK
6. #2077 - batch-STARK

tcoratger:
Correction: **this was never merged into anything.** GitHub marked it merged automatically when a sibling branch in the stack was force-pushed while still containing these commits. No part of this change reached `main`.

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2082, opened 2026-09-08
feat(circle)!: drive the Circle PCS transcript through the typed Fiat-Shamir layer

tcoratger:
Depends on #2075. This PR targets `feat/typed-transcript-fri-pcs` so the diff contains only this crate's migration; retarget to `main` after the branches below it land.

The Circle PCS wrote its transcript by hand on both sides: absorb the claimed evaluations, sample the batching challenge, absorb the first-layer commitment, sample the bivariate challenge, then per round a commitment, a grind and a folding challenge, then the constant, a grind and the query indices. Nothing checked that the two sides agreed, and the shape of the run was bound only by what happened to be absorbed.

It now derives every challenge through `ProverState` / `VerifierState`, over a description built from the FRI parameters and the claims the caller opened. This is #2035 applied to `p3-circle`, which has its own FRI driver.

## What the transcript is

New `circle::transcript` holds the description plus one driver per side. One driver spans the whole PCS, because the first-layer bivariate fold sits between the batching challenge and FRI, and `p3-circle` never calls into `p3-fri`'s prover or verifier.

```text
    per claimed opening:  opening point   two coordinates on the circle
                          opened values   one extension element per column
    batching challenge                    one extension element
    first-layer commitment                one opaque value
    bivariate challenge                   one extension element
    per commit round:     commitment      one opaque value
                          grinding        only when the difficulty is positive
                          folding         one extension element
    final polynomial                      one extension element
    query grinding                        only when the difficulty is positive
    query indices                         num_queries draws of index_bits bits
```

Both sides make the same calls in the same order, so a prover loop cannot drift from the verifier.

## What the seed binds

The description lands in the seed as a fingerprint. It binds the width of every claim, the commit-round count, the width of every query index, and both grinding difficulties.

Two things leave the step sequence unchanged and go in the instance label instead:

- the blowup;
- the batch / matrix / point nesting of the claims.

A flat step list cannot tell one batch of two matrices at one point each from one batch of one matrix at two points. Those are different statements, so the counts are written as a prefix code and bound.

Not bound: a commitment's width, which this layer cannot see, and the lambda corrections. Both are pinned by an opening check that recomputes them, exactly as in #2035.

## The opening point is now bound

Before, only the claimed values entered the sponge. The PCS bound "these values exist", not "this matrix takes these values at this point".

```text
    zeta on the projective line -> P = ((1 - zeta^2)/(1 + zeta^2), 2 zeta/(1 + zeta^2))
    P satisfies                     x^2 + y^2 = 1
```

`P` is now absorbed alongside the values it labels. A caller that samples `zeta` itself already bound it in its own transcript; the PCS no longer relies on that.

## The round count no longer comes from the proof

It was `commit_phase_commits.len() + log_blowup + 1`. That number set the query-index width and every height subtraction downstream, and only an under-report was rejected.

Circle folding halves the domain, and the bivariate layer takes one bit, so the count is an arithmetic consequence of the claimed heights:

```text
    H_claim = max claimed log_n + log_blowup
    rounds  = H_claim - 1 - log_blowup
```

Both sides derive it that way. A proof declaring a different count is rejected before the transcript is seeded, and the two-adicity bound on the index width becomes a statement about the configuration rather than about the proof.

## Shape checks move in front of the driver

Every length the description depends on is checked in `validate_proof_shape`, before a driver exists. The verifier then reads in three parts: check the shape, replay the transcript, check the queries. The query pass takes the challenges it needs and touches no sponge at all.

The driver is released on both exits of the one span that owns it — `finish` on success, `abort` on rejection — and nothing between the two branches returns early. Dropping an unfinished driver panics, and that panic would land on top of an error already travelling to the caller.

## No additions to `challenger/src/fs`

Worth calling out, with sibling transcript PRs in flight: this one adds nothing to the Fiat-Shamir layer. No new `TypeTag` variant, no new driver method.

A circle point is absorbed as a two-coefficient step under its own `opening_point` label, which is the existing `Algebra` tag used well: the label is what separates it from any other pair of extension elements. A `Codec` over `Point<EF>` would have needed new public methods on both drivers, since each owns its challenger privately, and a `CirclePoint` tag in a generic layer is a layering call better made when a second protocol wants one.

## Failures

The two replaying driver methods return `CircleTranscriptFailure`, a `thiserror` enum. Its one variant carries the grinding phase that rejected a witness and the bit count that step required, so the rejection says what was demanded and not only that something failed.

One variant is all that is reachable. A proof carries one witness per commit round plus one for the queries, and those counts are checked before the transcript is seeded, so a described grinding step always has a witness to replay. The PCS folds the failure into `FriError::InvalidPowWitness` for its own error surface.

## Tests

- Round trip: both drivers redraw the same challenges from the same values.
- Seed coverage: one test walks every field of the shape and asserts the first sampled challenge moves, including the claim grouping that only the instance label separates.
- Mutation: one test per absorbed value — opening point, claimed value, first-layer commitment, commit-phase commitment, both grinding witnesses, final constant — each asserting verification fails.
- Grinding: a witness that misses its difficulty is rejected on both phases, and the aborted driver drops quietly.

## Breaking changes

- **The Circle PCS transcript changes, so proofs produced before this do not verify after it.** The two circle compatibility fixtures are regenerated: `uni-stark/tests/fixtures/uni_stark_circle_v0_8_0.postcard` and `batch-stark/tests/fixtures/batch_stark_circle_v0_8_0.postcard`. The two-adic, STIR and WHIR fixtures never enter this path and regenerate byte-identical.
- `Pcs for CirclePcs` now requires `Val: PrimeField64` and `Challenger: GrindingChallenger<Witness = Val>`. Every challenger paired with it in the workspace already satisfies both.
- A verify call with no committed matrices is rejected (`FriError::NoCommittedMatrices`) rather than accepted.
- A prove call with `max_log_arity != 1` panics rather than producing a proof no verifier accepts. The verifier already refused anything else, and the described round count assumes it.
- `CirclePcs::verify` gains `InputError::DomainTooSmall` for a claimed domain that leaves no bit for the first-layer fold.

---

2. #2082 - Circle PCS
3. #2083 - STIR
4. #2084 - WHIR
5. #2076 - uni-STARK
6. #2077 - batch-STARK


tcoratger:

2. #2087 - Circle PCS
3. #2088 - STIR
4. #2089 - WHIR
5. #2090 - uni-STARK
6. #2091 - batch-STARK

PR #2095, opened 2026-09-08
perf(binary-pcs): add opt-in grouped Merkle leaves

Nashtare:
BinaryPcs currently hashes each 128-bit codeword symbol as a separate Merkle leaf. Add an opt-in `GroupedCodewordMmcs` adapter that groups adjacent symbols using a zero-copy matrix view. It preserves the original symbol queries and folding schedule, deduplicates authenticated groups, and transmits only group members absent from the requested rows.

The adapter supports the single power-of-two-height column used by BinaryPcs. Verification derives grouping from its own configuration and rejects malformed shapes, conflicting duplicates, and missing or extra group members. Grouping changes roots and proof encoding; existing example defaults remain unchanged.

On an Apple M4 Pro, 18 trials of the full 2^18-row AIR proof with Keccak, rate 1/4, 100-bit security, and zero PoW measured:

| Symbols per leaf | Median prove | Median verify | Median proof |
| --- | ---: | ---: | ---: |
| Original | 145.0 ms | 4.63 ms | 487 kB |
| 2 | 103.9 ms | 3.96 ms | 489 kB |
| 4 | 83.1 ms | 3.64 ms | 497 kB |
| 8 | 71.9 ms | 3.38 ms | 566 kB |
| 16 | 68.3 ms | 3.47 ms | 740 kB |

Four-symbol leaves reduce proving time by 42.7% with 2.1% larger median proofs. The benchmark balances execution positions and varies public transcript separators.

Validation: 65 binary PCS library/integration tests passed (one existing ignored test), including grouped commitment equivalence, malformed openings, tampering, tiny rounds, and serialized PCS lifecycles. Every timed AIR proof was serialized, decoded, and verified. Native Clippy with warnings denied, nightly rustfmt, and x86_64 Linux/wasm32 library checks passed. No native x86 performance claim.

Stacked on #2094 and based on `robin/binary-pcs-arm-performance` to keep this review focused. This is the grouped-leaf prototype and benchmark follow-up; it does not introduce multi-round folding or interleaving.


binary-pcs/src/lib.rs
```diff
@@ -18,6 +19,7 @@ mod verifier;
 
 pub use error::BinaryPcsError;
 pub use fold::{fold_codeword, fold_pair};
+pub use grouped_mmcs::GroupedCodewordMmcs;
```
tcoratger, line 22: Small API gap, no rush. `grouped_mmcs` is a private module, and only `GroupedCodewordMmcs` is re-exported — but `GroupedCodeword`, `GroupedCodewordProof` and `GroupedCodewordError` are all `pub` in there and all appear in `GroupedCodewordMmcs`'s public associated types (`ProverData`, `Proof`, `MultiProof`, `Error`).

So they're reachable but unnameable from outside the crate. A downstream caller doing `BinaryPcs::new(config, GroupedCodewordMmcs::new(mmcs, 4))` can't `match` on the MMCS error to tell `ConflictingDuplicate` from `Inner(..)`, can't name the proof type in a struct field or a `Deserialize` turbofish, and won't see any of the three in the rendered docs — the variants and their `#[error]` messages you wrote just don't show up. Your own test gets away with `BinaryPcsProof<GroupedCodewordMmcs<MyMmcs>>` because that's parameterized by the MMCS rather than naming the proof directly.

One line fixes it:

```rust
pub use grouped_mmcs::{
    GroupedCodeword, GroupedCodewordError, GroupedCodewordMmcs, GroupedCodewordProof,
};
```

Nothing warns about it today — `unreachable_pub` isn't on and rustdoc's private-in-public lint doesn't fire since they are `pub`. Fine to roll into #2097, which touches this file anyway.

tcoratger (approved):
The soundness argument here is clean, and you built it the right way round. Merging — one cosmetic API note inline.

**Why this is binding, in the form I checked it:** grouping is *exactly* a reshape, so the adapter never needs its own binding argument — it inherits the inner MMCS's. `groups_match_an_explicitly_reshaped_commitment` is what nails that down: the grouped root equals `mmcs().commit_matrix(RowMajorMatrix::new(values, group_size))` on the nose. Everything downstream follows, because every reconstructed lane — requested or supplemental — goes into the row the inner MMCS authenticates. A substituted `missing_symbol` cannot survive, and the test tampers each one individually to prove it.

**The counting is exact, not just sufficient.** `missing_count = groups.len() * group_size - requested.len()` is right because `groups` is deduped, `requested` is a `BTreeMap` keyed by distinct index, and each requested index lands in exactly one group — so `requested.len()` is precisely the number of lanes filled from openings and the remainder is precisely the number filled from the proof. Neither side can slip:

```text
    len < missing_count   -> OpeningShape        (pop test)
    len > missing_count   -> OpeningShape        (push test)
    len == missing_count  -> loop consumes all   (no leftovers, no panic)
```

So `missing.next().expect(..)` is genuinely unreachable, and the `==` rather than `>=` is what makes the extra-member case fail closed. No underflow either, since distinct indices give at most `group_size` per group; and `groups.len() * group_size <= dim.height` bounds the product, as the comment says.

**Ordering agreement** is the other thing that could have silently broken: prover and verifier both walk `group_indices(indices, group_size)` — the same sorted, deduped list — and lanes `0..group_size` within each. Same function, same inputs, so the ascending-symbol-index invariant in the doc comment is enforced by construction rather than by convention.

**The verifier does not read its grouping from the proof.** `group_size = self.group_size.min(dim.height)` comes from the verifier's own config and the validated dimensions, and `dimensions` is rejected unless it is exactly one width-1 power-of-two-height matrix. The group-8-verifier-rejects-a-group-4-proof assertion is the test I went looking for and it is already there.

Edge cases I poked at and found already handled: `height = 0` falls out because `0usize.is_power_of_two()` is false; the `group_size.min(height)` cap makes `group_size = 64` on a 32-symbol codeword collapse to a single leaf with the root still matching the reshape; duplicate indices are accepted when consistent and rejected via `ConflictingDuplicate` when not, and `[31, 0, 1, 17, 0]` exercises both unsorted and repeated. `GroupedCodeword`'s fields are private and `commit` is its only constructor, so the `row_subseq_unchecked` SAFETY comment's premise actually holds.

Verified locally: 65 tests pass under `+avx2,+pclmulqdq`, clippy `--all-targets -D warnings` clean, rustdoc with broken-intra-doc-links denied clean. `+668/-0` and opt-in, so nothing existing moves. The full `postcard` round trip across 5 group sizes x 3 arities is the coverage that makes me comfortable merging an adapter that changes the proof encoding.

PR #2097, opened 2026-09-08
perf(binary-pcs): batch folds between Merkle commitments

Nashtare:
Grouped Merkle leaves still build a separate tree for every intermediate fold. This change commits only at configured batch boundaries and authenticates each entire fold coset, following [Binius64's buffered-challenge FRI schedule](https://github.com/binius-zk/binius64/blob/37e9cd64e82243cde0e79c7d8ac0dc319f1cbeb4/crates/iop-prover/src/fri/fold.rs).

Stacked on #2095 (`robin/binary-pcs-grouped-leaves`, latest tip `d56454e6`).

- Add opt-in `BinaryPcsConfig::try_with_folding(k)`. Every sumcheck message still precedes its independent challenge; roots appear only at batch boundaries. The default single-fold configuration retains its transcript and query count.
- Fuse each coset's folds using scratch in the polynomial basis, avoiding full intermediate words and repeated basis conversions. Preserve the existing packed single-fold path.
- Size leaves for the next actual batch, including shorter tails. Verify all authenticated coset symbols and their global domain offsets through to the final word.
- Switch `prove_binary_field` to three-fold batches with matching leaves, retaining tracing and small-fixture support. Extend the existing benchmark to compare folding schedules.

For the example's 19-variable stacked polynomial, three-fold batching builds 7 trees instead of 19. The sum of committed codeword payloads falls from 64 MiB to 36.57 MiB (43% less; excludes internal digests). Initial encoding length and rate remain the same.

The batched security budget sums `(N/2^r + 1)/|F|` over every virtual fold and `2/|F|` per sumcheck round, using the UDR mutual-agreement argument in [Appendix A, Theorem 8](https://eprint.iacr.org/2024/1553). One bit is reserved for each of the field and query errors before adding them. At this 100-bit target, batching uses 149 queries versus 148 for the existing single-fold configuration. PoW is credited only to the query term.

Measured on Apple M4 Pro, native optimized build with parallelism: 2^18 rows, two GF(2^128) columns, Keccak, rate 1/4, 100-bit target, zero PoW. Medians of 18 measured trials per configuration, after warming every configuration; transcripts vary and execution order rotates/reverses. Every proof is serialized, decoded, and verified. Timings are complete AIR proving/verification, excluding trace generation, setup, and serialization.

| Configuration | Prove ms | Verify ms | Proof KiB |
|---|---:|---:|---:|
| ARM-optimized, ungrouped | 151.90 | 4.744 | 475.65 |
| 4-symbol leaves, one fold | 85.49 | 3.723 | 485.59 |
| 8-symbol leaves, one fold | 73.49 | 3.442 | 552.93 |
| 2-fold batches | 73.42 | 2.307 | 280.78 |
| 3-fold batches (example) | 63.29 | 1.959 | 248.88 |
| 4-fold batches | 58.19 | 2.133 | 279.69 |

Against eight-symbol grouping alone, three-fold batching saves 13.9% proving time, 43.1% verification time, and 55.0% proof bytes. Against four-symbol grouping, the proving saving is 26.0%. Four-fold batching reaches 58.2 ms, with larger proofs and slower verification than three-fold batching. These are ARM measurements; x86 was compile-checked only.

Reproduce:

```sh
RUST_LOG=off RUSTFLAGS='-C target-cpu=native -C opt-level=3' cargo run --profile optimized --features parallel -p p3-multi-stark --example bench_binary_grouping -- 18 18 vary batched
```

Validation:

- Native optimized parallel PCS suite: 74 passed, including the normally ignored large test. Portable debug suite: 73 passed, one ignored.
- Binary AIR example: all 8 tests passed, including mixed heights, preprocessed data, and malformed proofs; full 2^18 example also verified with tracing.
- Added oracle comparisons against repeated single folds on arbitrary codewords, including nonzero parallel-task offsets; partial tails, single-batch proofs, prescribed points, distinct/full-enumeration queries, tampered coset symbols, truncated openings, and zero-claim fold-chain attacks.
- Clippy with warnings denied for PCS/all targets and both examples; formatting and rustdoc links passed.
- x86 AVX2/PCLMUL and wasm library checks passed.

### DIFF
--- a/binary-pcs/src/fold.rs
+++ b/binary-pcs/src/fold.rs
@@ -340,6 +340,108 @@ pub fn fold_codeword(codeword: &[BinaryField128], beta: BinaryField128) -> Vec<B
     folded
 }
 
+/// Fold one aligned coset, including its global offset in every virtual layer.
+pub(crate) fn fold_coset(
+    coset_index: usize,
+    values: &mut [BinaryField128],
+    challenges: &[BinaryField128],
+) -> BinaryField128 {
+    assert_eq!(values.len(), 1usize << challenges.len());
+    let mut len = values.len();
+    for &beta in challenges {
+        len /= 2;
+        for i in 0..len {
+            values[i] = fold_pair(
+                coset_index * len + i,
+                beta,
+                values[2 * i],
+                values[2 * i + 1],
+            );
+        }
+    }
+    values[0]
+}
+
+/// Fold several sequential challenges without materializing full intermediate codewords.
+///
+/// Each task reuses one coset-sized scratch buffer in the polynomial basis. Symbols cross
+/// bases only when loaded from the source or stored into the final output. The Cantor domain
+/// bases advance by XOR, just as in the single-fold path. `challenges` stays in sumcheck order.
+pub(crate) fn fold_codeword_batch(
+    codeword: &[BinaryField128],
+    challenges: &[BinaryField128],
+) -> Vec<BinaryField128> {
+    assert!(!challenges.is_empty());
+    assert!(codeword.len().is_power_of_two());
+    assert!(challenges.len() <= codeword.len().ilog2() as usize);
+    if challenges.len() == 1 {
+        return fold_codeword(codeword, challenges[0]);
+    }
+    let arity = challenges.len();
+    let size = 1 << arity;
+    let num_cosets = codeword.len() / size;
+    let betas: Vec<_> = challenges.iter().copied().map(Ghash128::from).collect();
+    let offsets: Vec<Ghash128> = (0..size / 2).map(|i| domain_point(i << 1)).collect();
+    let steps: Vec<Vec<Ghash128>> = (0..arity)
+        .map(|round| {
+            let mut step = Ghash128::ZERO;
+            (0..num_cosets.ilog2() as usize)
+                .map(|bit| {
+                    step += Ghash128::cantor_basis(arity - round + bit);
+                    step
+                })
+                .collect()
+        })
+        .collect();
+    let mut output = BinaryField128::zero_vec(num_cosets);
+    output
+        .par_chunks_mut(FOLD_GRAIN)
+        .zip(codeword.par_chunks(FOLD_GRAIN.saturating_mul(size)))
+        .enumerate()
+        .for_each(|(task, (out, input))| {
+            let start = task * FOLD_GRAIN;
+            let mut bases: Vec<Ghash128> = (0..arity)
+                .map(|r| domain_point(start << (arity - r)))
+                .collect();
+            let mut scratch = Ghash128::zero_vec(size);
+            for (offset, (slot, coset)) in out.iter_mut().zip(input.chunks_exact(size)).enumerate()
+            {
+                let index = start + offset;
+                if offset != 0 {
+                    for (base, steps) in bases.iter_mut().zip(&steps) {
+                        *base += steps[index.trailing_zeros() as usize];
+                    }
+                }
+                for (dst, &src) in scratch.iter_mut().zip(coset) {
+                    *dst = Ghash128::from(src);
+                }
+                let mut len = size;
+                for (r, &beta) in betas.iter().enumerate() {
+                    len /= 2;
+                    let packed_end = len / WIDTH * WIDTH;
+                    let beta_packed = Packed::broadcast(beta);
+                    for i in (0..packed_end).step_by(WIDTH) {
+                        let lo = Packed::from_fn(|lane| scratch[2 * (i + lane)]);
+                        let hi = Packed::from_fn(|lane| scratch[2 * (i + lane) + 1]);
+                        let x = Packed::from_fn(|lane| bases[r] + offsets[i + lane]);
+                        let f1 = lo + hi;
+                        let f0 = lo + x * f1;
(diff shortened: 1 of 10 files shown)


tcoratger (approved):
This is the one in the batch where a wrong constant would be invisible — dropping intermediate commitments removes the verifier's ability to check intermediate proximity directly, so the whole thing rests on the accounting. I checked the accounting rather than the benchmarks. It holds. Approving.

## The soundness budget is exactly the formula you claim

`field_security_bits`'s batched branch is a closed form for the union bound, and it is the right closed form. The geometric series:

```text
    sum_{r=0}^{n-1} N/2^r  =  2N - 2*(N/2^n)  =  (2 << D) - (2 << L)
```

since `L = D - n = log_inv_rate`. Add `n` for the `+1` per fold and `2n` for the sumcheck rounds and you get the `3 * num_variables` term. Reproduced numerically at `n = 19, D = 21, L = 2`:

```text
    geometric sum           4194296
    (2 << 21) - (2 << 2)    4194296
    + 19 + 38               4194353   ceil(log2) = 23  ->  128 - 23 = 105 bits
```

which is your test's `4_194_353` and `105` on the nose. `ceil` on the numerator's log is the conservative rounding direction, and `2u128 << D` with `D < usize::BITS` cannot overflow while `D > L` rules out the subtraction underflowing.

The one-bit reservation is right in both places it has to be: `security_level <= field_security_bits - 1` for the field half, and `REGIME.queries(security_level + 1 - pow_bits, ..)` for the query half, so `2^-(s+1) + 2^-(s+1) = 2^-s`. `try_new`'s `security_level <= pow_bits` rejection is what keeps `security_level + reserve - pow_bits` from underflowing, and `try_with_folding` can only be reached through `try_new`. The `try_with_folding(1)` path re-derives with `reserve = 0`, so it really is identical to `try_new` rather than approximately so.

Charging every *virtual* fold — the ones with no root — is the correct reading of [2024/1553] Appendix A Theorem 8, and the doc comment gives the reason that actually matters: the input word is fixed before its challenge because it is a deterministic function of committed data, and the query authenticates the whole coset so the verifier recomputes the composition itself. The note that these are **sequential independent challenges, not powers of one sampled challenge**, is the detail that makes the per-round union bound legitimate; powers would correlate the folds and the sum would no longer be a bound.

## The query cap argument closes

`query_security_bits` reports the error at the *derived* count while only `min(num_queries, domain_size >> k)` cosets are sampled, which looks like an overstatement until you notice the cap binds only when *every* coset is opened — and your test pins exactly that:

```rust
if config.num_queries() >= cosets {
    assert_eq!(indices, (0..cosets).map(|i| i << arity).collect::<Vec<_>>());
}
```

All cosets opened means every base position is checked, so the query-phase error is zero, not a proximity-gap power. The reported figure is then a lower bound, as the doc says.

## The fold algebra

`fold_coset` is the naive in-place version and that is the right call — the verifier should be obviously correct and the prover can be clever. I verified the position bookkeeping, which is the part that could silently drift: after `j` folds the surviving block starts at `coset_index * 2^(a-j)`, so `coset_index * len + i` is the true global output index at every virtual layer, and at `len = 1` it lands on `coset_index = index >> (start + arity)` — precisely the coordinate checked against the next committed word. The `arity == 1` branch is bit-identical to the old code (`fold_pair(index >> (r+1), ..)` and offset `next_position & 1` reproduce the old `position >> 1` and `parity`), which is what makes "the default configuration retains its transcript" a fact rather than an intention.

`batched_folds_match_sequential_folds_at_every_coset_offset` is the right oracle — both the fused prover path and the verifier's `fold_coset` at *every* offset, against repeated `fold_codeword`. I widened it on your branch to `log_len` in `{1,2,3,5,6,8,9,10,11,13,15,16}` and `arity` up to 9, and added a batch-splitting associativity check (`fold_codeword_batch(w, &c[..a-1])` then one more single fold must equal the whole batch). All pass, including across `FOLD_GRAIN` task boundaries.

## Grouping and the verifier's independence

`group_size_at` is the piece I most wanted to see derived rather than transmitted, and it is: `min(group_size, height >> log_inv_rate)` computes the remaining-variable count from the verifier's own rate and the validated dimensions, so a shorter tail sizes its own leaves and no proof field influences the grouping. Returning `BadDimensions` on `message_len == 0` in `verify_multi_batch` while `commit` keeps an `expect` is the right asymmetry — fail closed on the verifying side, panic on the proving side. `new` leaving `log_inv_rate: None` keeps the #2095 behaviour byte-for-byte. Mixing the two constructors across prover and verifier fails verification rather than weakening it.

`betas` is still length-asserted against `config.num_fold_rounds()` and is the verifier's own transcript-derived vector, so `betas[start..start + arity]` cannot be made to panic by a malformed proof — worth noting because the refactor turned an index into a slice, which is exactly where that guarantee could have been lost.

## Verified locally

`+avx2,+pclmulqdq` release: 73 binary-PCS tests plus 123 multi-STARK tests pass. Portable debug: 73 pass. Clippy `--all-targets -D warnings` clean on both crates; `cargo fmt --check` clean. Shape checks still run before `check_witness` and before sampling, so a malformed proof is rejected without touching the transcript.

One small thing: the unreachable-`pub` note I left on #2095 still stands and this stack is the natural place to fold it in. Nothing blocking.

PR #2098, opened 2026-09-09
perf(goldilocks): optimize scalar and packed arithmetic across backends

Nashtare:
Reduce Goldilocks arithmetic costs across scalar, AVX2, AVX-512, NEON/SVE2, and Wasm SIMD implementations. The changes preserve field results for every raw `u64` representation, including noncanonical values, without changing public APIs, serialization, or production dependencies.

The branch contains 27 optimization commits and two test-only lint cleanups, keeping independent algorithms and backends separately reviewable.

## Changes

- Specialize division by small powers of two, packed power coefficients, full-width integer conversion, and two-adic square roots.
- Delay reduction in cubic Frobenius, short x86 dot products, and large packed sums. Use four accumulation chains for sums at N >= 64; preserve the existing short paths.
- Reduce NEON multiplication register pressure; keep SVE2 dot reduction in vector registers, accumulate slice linear combinations directly, and vectorize packed-RHS dots at N >= 32.
- Keep Wasm carry detection vectorized and eliminate temporary sum copies with a bounded final carry fold.
- Cache the existing small-batch MDS transform plan and reuse its pointwise-product buffer, reducing warm allocations from four to one.
- Apply direct power division to generic/Wasm Poseidon2 widths 12/16 and x86 width 16; balance the AVX-512 width-16 tail sum.
- Add varied-input benchmarks and independent full-range arithmetic oracles covering carry saturation, noncanonical inputs, chunk boundaries, and packed/scalar equivalence.

## Performance

Numbers below are reductions in elapsed time from matched before/after comparisons for each optimization. They are incremental, not additive, and do not establish a combined whole-prover speedup.

Dedicated hosts: AMD EPYC 9R45 / Zen 5 (`c8a.2xlarge`) and Graviton4 / Neoverse V2 (`c8g.2xlarge`), Rust 1.97.0 / LLVM 22.1.6, CPU 2 pinning. Native builds use `target-cpu=native`; AVX2 disables AVX-512F and NEON disables SVE/SVE2. Criterion uses 30 samples, 0.5 s warmup, and 1 s measurement; scratch dependency/throughput kernels use repeated warmed calls and seven-trial medians. Wasm measurements use Wasmtime 45.0.2 and Node 26.3.1, with optimizing-only Node controls where noted.

| Shared operation / consumer | Zen 5 | Graviton4 |
|---|---:|---:|
| Small-power division, 200 independent values | 95.6% | 75.0-75.3% |
| `u128` / `i128` conversion | 61.7% / 51.3% | 89.6% / 87.0% |
| Cubic Frobenius / complete cubic inversion | 19.2% / 2.8% | 19.5% / 3.3% |
| Square roots, varied inputs | 8.1-15.7% | 15.4-16.8% |
| MDS32 / MDS64 caching | 19.3% / 19.7% | 22.4% / 19.7% |
| Additional MDS buffer reuse | ~0.8% | 0.4-1.7% |
| Complete FFTs from NEON multiplication change | N/A | 6.0-6.3% DIT; 10.1-10.9% Bowers |

The independent division-array result includes compiler vectorization; it is not scalar dependency-chain latency.

| Packed measurement | AVX2 | AVX-512 | NEON | SVE2 |
|---|---:|---:|---:|---:|
| Fixed small-power division latency | 53-54% | 63-65% | 35-42% | 53-59% |
| Large sums, independent work | 66-69% | 27-34% | 28-37% | 34-39% |
| Large sums, dependent chains | 50-66% | 28-31% | 24-37% | 28-39% |
| EF2 / EF3 / EF5 multiplication, independent work | 16.6 / 27.7 / 39.3% | 6.9 / 9.7 / 13.6% | — | — |

SVE2 short mixed dots save 12-22% independently and 10-14% dependently. Large packed dots save 4-33% and 6-27%, respectively. Slice linear combinations save 14-42% at useful measured sizes; length 1024 is neutral.

Direct division in Poseidon2 width 16 saves 6.2% on AVX2, 0.8-2.9% on AVX-512, and approximately 9% on the Zen 5 scalar path. The separate AVX-512 tail-sum change saves 0.5-6%, depending on input/layout.

| Wasm host / engine | Fixed small-power division latency | Sum optimization, selected sizes | Packed Poseidon2 width 16, direct division |
|---|---:|---:|---:|
| Zen 5 / Wasmtime | 36-37% | 25-43% | 10.7-10.9% |
| Zen 5 / Node | 38-40% | 23-63% | ~9%; ~12% optimizing-only |
| Graviton4 / Wasmtime | 51-54% | 22-41% | ~9% |
| Graviton4 / Node | 49-52% | 5-21% | ~9% |

The separate Wasm carry change saves 20-42% on sums and 0.6-4.3% on complete Poseidon2. Sum ranges cover different useful sizes across engines.

### Tradeoffs and coverage limits

- MDS caching increases first-call allocations from 6 to 23; buffer reuse brings that to 22. Instrumented cold-call medians increase roughly 0.7 microseconds on Zen 5 and 3 microseconds on Graviton4. Warm allocations fall 4 -> 2 -> 1.
- AVX-512 dot N=2 dependent latency regresses ~4.8%; actual EF2/3/5 multiplication improves in both throughput and latency. Fixed packed division at exponent 95 regresses ~3.3% AVX2 / ~5.9% AVX-512 in latency.
- Wasm has localized tradeoffs: the carry change regresses x86 Wasmtime dots N=3/4 by 6.4%/3.4%; the sum change regresses one width-16 permutation case by 1.5%; direct hash division regresses width-12 varied throughput by 2.9% on x86 Wasmtime while improving its latency by 1.6%.
- Small and medium paths retain their existing implementations where setup costs or dependency latency outweigh gains. x86 Poseidon2 width 12 retains its original power expressions.
- Graviton4 runtime coverage is SVE VL=128. Wider-VL safety was reviewed from predicate, pointer, and lane bounds; wider hardware and Intel CPUs were not measured.

## Validation

Exact final source: Goldilocks release library tests passed 211 AVX2, 211 AVX-512, 186 scalar, 375 NEON, 379 SVE2, and 207 Wasm tests on each host, with zero failures.

Broader release library suites cover Goldilocks, field, MDS, DFT, Poseidon1/2, BabyBear, KoalaBear, and Mersenne31: 1,057 passing tests per x86 SIMD mode, 886 scalar, 1,224 NEON, and 1,228 SVE2. Parallel Goldilocks/field/DFT suites passed 246 x86 and 412 Arm tests. Doctests passed one test with six repository-marked ignored examples. These broad runs precede only the final restoration of the original x86 width-12 expressions and the test-only Wasm lint cleanup; exact-final Goldilocks and x86 Clippy checks were repeated afterward.

- Native strict Clippy: `cargo clippy --locked -p p3-goldilocks -p p3-field -p p3-mds -p p3-poseidon2 --all-targets --no-deps -- -D warnings` passed in all five modes.
- `cargo +nightly fmt --all -- --check`, `git diff --check`, and `scripts/check.py fast --package p3-goldilocks` passed.
- Wasm build and packed/scalar smoke checks passed via `scripts/check.py wasm`.
- Wasm filters only the two existing Poseidon constant-generator tests that spawn Python, which WASI cannot execute; native suites run them. Wasm Clippy checks supported lib/bin/test targets and passes with only the existing `doc_overindented_list_items` lint allowed for eight unchanged module-doc lines. Criterion benchmark dependencies are intentionally unavailable on Wasm.

All implementation timings ran on dedicated EC2 hosts, without stopping other local sessions. Both task-owned hosts were explicitly terminated after evidence collection.

### DIFF
--- a/goldilocks/src/mds.rs
+++ b/goldilocks/src/mds.rs
@@ -6,7 +6,7 @@
 //! work by Angus Gruen and Hamish Ivey-Law. Other sizes are from Ulrich Haböck's
 //! database.
 
-use p3_dft::{Radix2Bowers, TwoAdicSubgroupDft};
+use p3_dft::{Radix2DFTSmallBatch, TwoAdicSubgroupDft};
 use p3_field::PrimeCharacteristicRing;
 use p3_mds::MdsPermutation;
 use p3_mds::karatsuba_convolution::Convolve;
@@ -69,7 +69,8 @@ impl Convolve<Goldilocks, i128, i64> for SmallConvolveGoldilocks {
     }
 }
 
-const FFT_ALGO: Radix2Bowers = Radix2Bowers;
+static FFT_ALGO: LazyLock<Radix2DFTSmallBatch<Goldilocks>> =
+    LazyLock::new(|| Radix2DFTSmallBatch::new(64));
 
 pub(crate) const MATRIX_CIRC_MDS_8_SML_ROW: [i64; 8] = [7, 1, 3, 8, 8, 3, 4, 9];
 
@@ -163,7 +164,7 @@ static MATRIX_CIRC_MDS_32_FREQ: LazyLock<[Goldilocks; 32]> = LazyLock::new(|| {
 
 impl Permutation<[Goldilocks; 32]> for MdsMatrixGoldilocks {
     fn permute(&self, input: [Goldilocks; 32]) -> [Goldilocks; 32] {
-        apply_circulant_fft_precomputed(&FFT_ALGO, &MATRIX_CIRC_MDS_32_FREQ, &input)
+        apply_circulant_fft_precomputed(&*FFT_ALGO, &MATRIX_CIRC_MDS_32_FREQ, &input)
     }
 }
 impl MdsPermutation<Goldilocks, 32> for MdsMatrixGoldilocks {}
@@ -199,7 +200,7 @@ static MATRIX_CIRC_MDS_64_FREQ: LazyLock<[Goldilocks; 64]> = LazyLock::new(|| {
 
 impl Permutation<[Goldilocks; 64]> for MdsMatrixGoldilocks {
     fn permute(&self, input: [Goldilocks; 64]) -> [Goldilocks; 64] {
-        apply_circulant_fft_precomputed(&FFT_ALGO, &MATRIX_CIRC_MDS_64_FREQ, &input)
+        apply_circulant_fft_precomputed(&*FFT_ALGO, &MATRIX_CIRC_MDS_64_FREQ, &input)
     }
 }
 impl MdsPermutation<Goldilocks, 64> for MdsMatrixGoldilocks {}
@@ -234,9 +235,62 @@ impl MdsPermutation<Goldilocks, 68> for MdsMatrixGoldilocks {}
 
 #[cfg(test)]
 mod tests {
+    use p3_mds::util::apply_circulant;
     use p3_symmetric::Permutation;
+    use rand::rngs::SmallRng;
+    use rand::{RngExt, SeedableRng};
 
-    use super::{Goldilocks, MdsMatrixGoldilocks};
+    use super::{
+        Goldilocks, MATRIX_CIRC_MDS_32_GOLDILOCKS, MATRIX_CIRC_MDS_64_GOLDILOCKS,
+        MdsMatrixGoldilocks,
+    };
+
+    #[test]
+    fn goldilocks32_and_64_match_direct_circulant_on_raw_edges_and_random_inputs() {
+        const RAW_EDGES: [u64; 10] = [
+            0,
+            1,
+            (1 << 32) - 1,
+            1 << 32,
+            1 << 63,
+            0xFFFF_FFFF_0000_0000,
+            0xFFFF_FFFF_0000_0001,
+            0xFFFF_FFFF_0000_0002,
+            u64::MAX - 1,
+            u64::MAX,
+        ];
+
+        for offset in 0..RAW_EDGES.len() {
+            let input32 = core::array::from_fn(|i| {
+                Goldilocks::new(RAW_EDGES[(i + offset) % RAW_EDGES.len()])
+            });
+            let input64 = core::array::from_fn(|i| {
+                Goldilocks::new(RAW_EDGES[(i + offset) % RAW_EDGES.len()])
+            });
+            assert_eq!(
+                MdsMatrixGoldilocks.permute(input32),
+                apply_circulant(&MATRIX_CIRC_MDS_32_GOLDILOCKS, &input32)
+            );
+            assert_eq!(
+                MdsMatrixGoldilocks.permute(input64),
+                apply_circulant(&MATRIX_CIRC_MDS_64_GOLDILOCKS, &input64)
+            );
+        }
+
+        let mut rng = SmallRng::seed_from_u64(0x4D_D5_32_64);
+        for _ in 0..32 {
+            let input32 = Goldilocks::new_array(rng.random());
+            let input64 = Goldilocks::new_array(rng.random());
+            assert_eq!(
+                MdsMatrixGoldilocks.permute(input32),
+                apply_circulant(&MATRIX_CIRC_MDS_32_GOLDILOCKS, &input32)
+            );
+            assert_eq!(
+                MdsMatrixGoldilocks.permute(input64),
+                apply_circulant(&MATRIX_CIRC_MDS_64_GOLDILOCKS, &input64)
+            );
+        }
+    }
 
     #[test]
     fn goldilocks8() {
--- a/goldilocks/src/mds.rs
+++ b/goldilocks/src/mds.rs
@@ -277,7 +277,7 @@ mod tests {
             );
         }
 
-        let mut rng = SmallRng::seed_from_u64(0x4D_D5_32_64);
+        let mut rng = SmallRng::seed_from_u64(0x4DD5_3264);
         for _ in 0..32 {
             let input32 = Goldilocks::new_array(rng.random());
             let input64 = Goldilocks::new_array(rng.random());
--- a/field/src/extension/cubic_extension.rs
+++ b/field/src/extension/cubic_extension.rs
@@ -73,9 +73,9 @@ impl<F: CubicTrinomialExtendable> HasFrobenius<F> for CubicTrinomialExtensionFie
     fn frobenius(&self) -> Self {
         let a = &self.value;
         let m = F::FROBENIUS_MATRIX;
-        let c0 = a[0] + m[0][1] * a[1] + m[0][2] * a[2];
-        let c1 = m[1][1] * a[1] + m[1][2] * a[2];
-        let c2 = m[2][1] * a[1] + m[2][2] * a[2];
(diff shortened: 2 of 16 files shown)


goldilocks/src/mds.rs
```diff
@@ -69,7 +69,8 @@ impl Convolve<Goldilocks, i128, i64> for SmallConvolveGoldilocks {
     }
 }
 
-const FFT_ALGO: Radix2Bowers = Radix2Bowers;
+static FFT_ALGO: LazyLock<Radix2DFTSmallBatch<Goldilocks>> =
+    LazyLock::new(|| Radix2DFTSmallBatch::new(64));
```
tcoratger, line 73: Not a defect, just flagging where the measurement and the deployment may diverge.

`Radix2Bowers` was a stateless unit struct. `Radix2DFTSmallBatch` holds `Arc<RwLock<TwiddlePair<F>>>`, and `dft_batch`/`idft_batch` each do `self.cache.read().twiddles.clone()` — so every `permute` on `[Goldilocks; 32]` / `[Goldilocks; 64]` now takes two read locks and two `Vec<Vec<F>>` clones through one shared global.

Correctness is fine: `new(64)` pre-populates through the largest size either impl asks for, `update_twiddles` only takes the write lock when it has to grow, and it never does here — so no writer ever blocks a reader. The clone is also clearly cheaper than recomputing twiddles by field exponentiation, which is where your 19-22% comes from.

The part the benchmark can't see is that your numbers are single-threaded with CPU pinning, while the new cost is one shared `RwLock` cache line plus a handful of allocations per permute. Under a parallel prover those scale differently from a pinned loop — read locks don't serialize, but the atomic RMW on that one line does bounce between cores.

Probably irrelevant in practice, since `MdsMatrixGoldilocks` at width 32/64 isn't on the Poseidon2 path that dominates proving. Mentioning it only because it's the one number in the PR body I can't reproduce the conditions for, and because you already flagged the cold-call allocation trade (6 -> 23) — this is the warm multi-threaded counterpart of the same question. No change requested.

tcoratger (approved):
Reviewed this as the highest-risk of the perf batch, since a wrong reduction in Goldilocks silently corrupts every proof rather than failing loudly. I did not take the test suite's word for it — I wrote independent oracles against `u128` modular arithmetic and ran them in all three x86 modes. Everything holds. Merging.

**Math I re-derived by hand:**

`div_2exp_u64`, the new `2..=32` arm, is exactly right and the comment states the reason correctly. With `x = hi*2^exp + lo`:

```text
    (a - b) * 2^exp  =  hi*2^exp + lo*2^32 - lo*2^64
                     == hi*2^exp + lo*2^32 - (lo*2^32 - lo)     [2^64 == 2^32 - 1]
                     == hi*2^exp + lo  =  x            (mod P)
```

Both intermediates are overflow-free rather than merely usually-fine: `lo << (32-exp) < 2^32` and `hi < 2^62` for `exp >= 2`, so `a < 2^62 + 2^32`; and `b <= 2^64 - 2^(64-exp) <= 2^64 - 2^32`, which is exactly one less than `P`. That is what makes `-P < a - b < P` tight and the single `NEG_ORDER` correction sufficient — `NEG_ORDER = 2^64 - P = 2^32 - 1` is precisely the `2^64 -> P` fixup on borrow. The bound would fail at `exp = 33`, and the arm stops at 32.

The specialized `try_sqrt` is also correct, including the part that looks like a shortcut but isn't. `p - 1 = 2^32 * (2^32 - 1)`, so `Q = 2^32 - 1` and `(Q-1)/2 = 2^31 - 1`; `r = u*a = a^(2^31) = a^((Q+1)/2)` and `t = r*u = a^Q` are the standard Tonelli-Shanks initializers. Reading `TWO_ADIC_GENERATORS[i+1]` from the fixed table instead of tracking `c = c^2` per iteration is valid *because* of the documented chain property `G_{k+1}^2 = G_k`: the invariant `c = G_M` holds initially (`G_32`, `M = 32`) and is preserved by `c = b^2 = G_{i+1}^2 = G_i` with `M = i`. `i < m <= 32` keeps `i+1` inside the 33-entry table. And returning `r` when `t == 1` is right because `r^2 = a^(2^32) = a` is equivalent to `a^Q = 1`.

`power_of_two` leans on `2^96 == -1` (from `2^96 = (2^32-1)*2^32 == -1`), hence period 192; `POWERS_OF_TWO` entries are canonical and nonzero (`2^63 < P`, and `const_add` reduces above that), so `ORDER_U64 - value` cannot underflow as the comment claims.

**Independent oracles I ran** (scalar, `+avx2`, `+avx512f,dq,vl,bw` — all green):

- `div_2exp_u64` and `mul_2exp_u64` for every `exp` in `0..=192` against `x * powmod(2, ±exp)`, over ~4000 raw `u64` values including a dense sweep across the `P` boundary and both `u64::MAX` neighbours. Also round-tripped each result by multiplying `2^exp` back. Noncanonical inputs included throughout — the identity is representation-independent, and the results confirm it.
- `try_sqrt` against *two* independent references: the generic `tonelli_shanks_two_adic`, and the Euler criterion `a^((p-1)/2) == 1` computed in `u128`. Residuosity agreed on every input and every returned root squared back.
- Packed `div_2exp`/`mul_2exp` compared lane-wise against the scalar path for all `exp in 0..=192`, seeded with all-`u64::MAX`, all-`P`, all-`P-1`, all-zero and 100 random vectors.
- The delayed-reduction cubic Frobenius verified as an actual `q`-power map: additive, multiplicative, equal to `x^p` by independent square-and-multiply, and `frob^3 = id`, over 2000 random elements.
- `sum_array` at `u64::MAX` in every lane for `N` up to 256 against a `u128` oracle — the worst case for the four-chain carry accounting.

Your own `packed_sum_array_matches_full_u64_oracle` and `packed_dot_products_match_full_u64_oracle` are the right shape, and the note about not using the scalar sum as the oracle *because it also delays reduction* is exactly the trap to avoid. Reducing each product separately in the dot-product oracle so `max * max` repeated doesn't overflow the `u128` expected value is a nice touch.

**Unsafe:** `sve2_dot_chunk` is the only substantial new block and it is disciplined. `ptrue p7.d, vl2` fixes two active lanes independently of hardware VL, which is the correct way to make this safe on wider SVE than you measured. The `subs`/`b.ne` loop is do-while, so a zero length would run ~2^64 iterations — both dispatchers guard with `if is_empty() { return ZERO }`, `chunks()` never yields an empty chunk, and `chunk_limit` is asserted into `1..=u32::MAX`, which is also what bounds the carry counters. Every call site carries a SAFETY comment that matches the documented contract. I could not execute it (no aarch64 host), but `--target aarch64-unknown-linux-gnu` checks clean under both `+neon` and `+neon,+sve2`.

**Also verified:** clippy `--all-targets --no-deps -D warnings` clean in all three x86 modes; `cargo fmt --check` clean; `p3-baby-bear`, `p3-koala-bear`, `p3-mersenne-31`, `p3-dft`, `p3-poseidon1`, `p3-monty-31` all pass under `+avx2`, which matters because `cubic_extension.rs` and `mds/src/util.rs` are shared generic code. The `apply_circulant_fft_precomputed` in-place rewrite is value-identical — same products, one fewer `Vec`.

One non-blocking question inline about the MDS plan static. 29 commits that each do one thing to one backend made this reviewable at all; thanks for splitting it that way.

PR #2100, opened 2026-09-09
fix(security)!: enforce PCS budgets and compose multi-STARK soundness

Nashtare:
Wide STIR quotient batches and WHIR initial claim batches could fall below their configured security targets. Hiding FRI accepted traces with insufficient masking, and multi-STARK offered no composed security assessment. The reproduced STIR case retained 86.76 bits against a 100-bit target.

This PR fixes the accounting and enforces it at the relevant entry points:

- **STIR:** derive and cache schedules using actual quotient counts and native rates; retune feasible batches and reject infeasible ones without granting later grinding credit to the initial alpha.
- **WHIR:** charge initial claim batching, validate actual opening/OOD counts, and expose composed opening errors plus commitment-time candidate-list bounds. Give fresh intermediate claims distinct alpha powers from the carried claim so their errors cannot cancel identically.
- **Hiding FRI:** enforce trace/query/opening masking budgets and sufficient random codewords. Private prover data is single-use across clones and bound to its originating masking configuration; public preprocessing remains reusable.
- **Multi-STARK:** add statement-derived AIR, sumcheck, lookup, GKR, PCS, and collision accounting, including the product of the main/preprocessed candidate-list sizes. New `prove_with_security` and `verify_with_security` entry points reject insufficient or missing evidence. Small `Challenge = Val` configurations fail the requested budget; sufficiently large base fields remain supported.

Compatibility and scope:

- STIR schedules and the WHIR transcript change; WHIR uses transcript version 2 and both compatibility fixtures are refreshed.
- Hiding FRI wraps prover data, rejects repeated private openings, and can require larger traces. Raw masked-LDE commitments still require fresh caller-supplied masks; hiding MMCS and RNG assumptions remain explicit.
- Existing multi-STARK `prove`/`verify` entry points remain explicitly unchecked. Reports inherit the configured PCS and Fiat-Shamir assumptions and require explicit collision-security evidence.
- Existing WHIR benchmark challenge fields are enlarged where necessary, with labels clarified as per-phase targets rather than total proof security. STIR command-line examples and tests use quintic challenge fields to cover wide AIR quotient batches at the existing 100-bit target.

Validation:

- Affected default suites passed, including security, sumcheck, STIR, WHIR, FRI, multi-STARK, uni-STARK, batch-STARK, and binary PCS.
- Parallel suites passed for FRI, STIR, sumcheck, WHIR, multi-STARK, uni-STARK, and batch-STARK: 1,097 tests passed.
- Affected all-target Clippy with `-D warnings`, nightly formatting, rustdoc broken-link checks, and `thumbv7em-none-eabi` checks passed.
- STIR PCS, WHIR PCS, and FRI/WHIR comparison benchmark smoke runs passed.
- Follow-up CI validation: all 12 example tests passed in both default and parallel mode, along with all-target Clippy and formatting. CLI proof/verify round trips passed for both KoalaBear/BabyBear and Keccak/Poseidon2 Merkle hashes.
- Regression coverage includes understrength batches, weak challenge fields, missing/malformed evidence, candidate-list composition, carried-claim cancellation, short-trace disclosure, masking boundaries, reuse, and mismatched prover-data configurations.

Known pre-existing benchmark limitation: `zk_overhead` generated the updated main plain/ZK proofs at sizes 16 and 18, then failed in its unchanged octic preset with `RandomnessExceedsSlack { randomness: 256, slack: 240 }`. That unrelated preset is left unchanged; the full ZK benchmark harness is not reported as passing.


whir/src/pcs/tests.rs
```diff
@@ -56,6 +56,91 @@ fn default_round_log_inv_rates(num_variables: usize, folding_factor: &FoldingFac
...
+    let (_, data) = pcs.commit(witness, &mut challenger);
+    // The unground alpha batch retains <100 bits, although every fold reaches 100.
+    let _ = pcs.open(data, protocol, &mut challenger);
+}
+
+#[test]
+fn both_verifiers_reject_infeasible_claim_counts_before_sumcheck() {
+    type L = PrefixProver<F, EF>;
+    let mut rng = SmallRng::seed_from_u64(947);
+    let witness = L::new_witness(vec![Table::rand(&mut rng, 1, 6)], 4);
```
tcoratger: This is the CI failure on the four x86 jobs (`ubuntu-latest +avx2` x2, `windows-latest` x2). Reproduced locally:

```
cargo test -p p3-whir both_verifiers_reject_infeasible_claim_counts_before_sumcheck
  # RUSTFLAGS="-C target-feature=+avx2"
thread '...' panicked at multilinear-util/src/split_eq/mod.rs:358:
  assertion failed: poly.num_variables() >=
      (self.num_variables() + log2_strict_usize(F::Packing::WIDTH))
```

The panic is in `pcs.open(...)` on line 117 — the setup call — so the assertions the test actually cares about are never reached.

`PrefixProver::into_sumcheck` needs `num_variables >= folding + log2(F::Packing::WIDTH)` for `compress_prefix_to_packed` to produce at least one full SIMD lane. Here `num_variables = 6` and `folding = 4`, leaving 2, so it holds only when `Packing::WIDTH <= 4`:

```
portable   WIDTH = 1  ->  6 >= 4 + 0   ok    (ubuntu-latest scalar passes)
AVX2       WIDTH = 8  ->  6 >= 4 + 3   FAIL
AVX-512    WIDTH = 16 ->  6 >= 4 + 4   FAIL
```

That precondition is already known in the tree — `sumcheck/src/layout/prover/prefix.rs:711` guards its proptest with exactly `prop_assume!(witness.num_variables() >= FOLDING + k_pack)`. It just isn't visible from `Layout::into_sumcheck`, whose only local `assert!` is `folding <= num_variables`.

**Fix:** raise the instance to 8 variables, which clears `4 + 4` for AVX-512 too — `6 -> 8` in the four places (`Table::rand(..., 1, 8)`, both `TableShape::new(8, 1)`, `WhirConfig::new_with_initial_claims(8, ..)`, and `vec![EF::ONE; 8]`). The asserted `num_claims == 2 * num_batches + commitment_ood_samples` doesn't depend on the arity, so nothing else moves. I applied exactly that on your branch and it passes in all three modes:

```
target-feature=                                     ok
target-feature=+avx2                                ok
target-feature=+avx512f,+avx512dq,+avx512vl,+avx512bw   ok
```

Worth a glance at `rejects_opening_batches_below_target_security` just above too: same `6 / folding 4` shape, and it passes on AVX2 only because the config error fires before `into_sumcheck` — so its `should_panic(expected = "initial claim combination")` is load-bearing on ordering rather than on the arity. It's green today; bumping it alongside would make both robust for the same one-line cost.

tcoratger (changes requested):
This is excellent work and the accounting is right where I checked it. One test-only fix stands between it and merge — details inline, it's a two-line change.

**What I re-derived by hand and confirmed:**

- `stir::soundness::initial_batching_error` — the shared-radius reparametrization is the correct move and the test pins it tightly. At `log_degree = 8, log_inv_rate = 1, eta_0 = 1/4` the degree-4 class sits at `rho_i = 2^-7` and `eta_i = 1/4 + 1/2 - 1/128 = 95/128`, exactly `eta_0 + B(rho_0) - B(rho_i)` from holding `delta = 1 - B(rho) - eta` fixed. Both probability terms in the expected value reproduce the capacity-bound `(k-1)*d/(rho^2*eta*q)`. Reparametrizing per class at its *native* rate instead would have been the easy wrong answer, and the comment says why it isn't.
- The log-domain probability sum is numerically sound: both `exp2` arguments are `<= 0` by construction, and the `f64::INFINITY` seed folds correctly on the first iteration (`exp2(-inf) = 0`).
- `final_eta` consistency. `initial_batching_error()` reads `round_configs[0].eta`, and round 0 stores the value *after* `max(batching_eta)` and `max(combine_eta)` — so the reported number is the enforced one, not a recomputation that could drift. The `map_or(self.final_eta, ..)` fallback fires only when `round_configs` is empty, which is exactly `total_folds == 1`, and the `if total_folds != 1` guard means `final_eta` is un-overwritten there. That is a genuinely easy thing to get subtly wrong and it's right.
- `union_bound_buffer` correctly grows by the batching term, and the batching eta is derived against `buffered_security_level` with no PoW credit — right, since alpha is sampled before the grind.
- `security/src/multilinear.rs` — every degree count is the correct one. `c-1` and `n-1` roots for the alpha/beta batches; `n/(q-1)` for zerocheck via Schwartz-Zippel on a multilinear over `(F \ {0})^n`; `n*(d+1)` sumcheck roots with the `+1` from the eq weight; `2L + 3L(L-1)/2` for GKR (2 draws per layer plus `3i` cubic rounds at layer `i`). The first test's `27/2^100` total is right including the `4/2^99 = 8/2^100` conversion on the zerocheck term.
- `MultiStarkSecurityReport` composes with `ErrorBits::sum`, not `min` — which is the correct composition for union-bound terms and the thing I most wanted to check, since the rest of `p3-security` composes round-by-round with `min`. `add_evidence` / `add_opening_evidence` fail closed on `None`, NaN, infinity and negatives, and the candidate-list debit is applied to `terms[..num_reduction_terms]` only, captured before the PCS terms are pushed. `field_bits = order.bits() - 1 = floor(log2 q)` is a genuine lower bound, as the doc requires.
- `height >= usize::BITS` rejection plus `checked_mul`/`checked_next_power_of_two` throughout `assess_statement` — the shift and the padded dimensions are both guarded.

Also ran the full affected set locally under `-C target-feature=+avx2` with `--no-fail-fast`: `p3-security`, `p3-stir`, `p3-sumcheck`, `p3-multi-stark`, `p3-fri`, `p3-uni-stark`, `p3-batch-stark`, `p3-commit` all clean (39 test binaries). The only failure anywhere is the one below.

PR #2101, opened 2026-09-09
test(binary-field): pin the packed lane backend to the scalar model

tcoratger:
Follow-up to #2054. Auto-merge fired on the approval and took the branch to `main` before the last review round was addressed, so this lands those six comments. @Nashtare, sorry about that — the merge was not a decision, it was a queue.

Four are code, two are measurements the review asked for and this description answers.

## 1. Nothing in CI executes the packed kernels

Correct, and it still is: every `vpclmulqdq` leg is `compile_only: true`, and the legs that run tests enable neither `avx2,+vpclmulqdq` nor `avx512f,+vpclmulqdq`. The scalar model added last round pins the *algebra* but cannot pin the seam you named — that `unpack_low_64` really is `_mm{256,512}_unpacklo_epi64` per lane, that `clmul::<0x01>` takes the high qword of the first operand, that `_mm512_broadcast_i32x4` lands the same 128 bits in all four lanes.

Your option 1, a direct register-vs-model conformance test, one lane operation at a time. `packed/x86_64.rs:533-624`.

The model moves out of `split.rs`'s test module into `split::model` and becomes generic over the lane count, so the same reference serves the two-lane and four-lane builds. The bridge between the two is the unaligned load and store themselves, so those are under test as well. Each operation gets its own assertion, so a failure names the intrinsic rather than the slice:

```
zero  broadcast  tail  xor  unpack_low_64
clmul::<0x00>  clmul::<0x01>  clmul::<0x10>  clmul::<0x11>
fold_shifted   SplitScalar::apply
```

Mutation-checked — I broke one intrinsic at a time and ran the suite:

| mutation | leg | new tests fail | anything else fails |
| --- | --- | --- | --- |
| `_mm256_unpacklo_epi64` → `_mm256_unpackhi_epi64` | `+avx2,+vpclmulqdq` | yes, both | yes, 7 pre-existing |
| `_mm512_broadcast_i32x4` → `_mm512_castsi128_si512` (lane 0 only) | `+avx512f,+vpclmulqdq` | yes, both | yes, 6 pre-existing |

I left `ci.yml` alone again, for the reason in the last round: the matrix comment says a leg's job name embeds its `features` and two of those names are required status checks, so flipping one to test-running is a branch-protection change. Still worth doing, still its own PR. Your `qemu-x86_64 -cpu max` suggestion is the cheaper half of it and I have not tried it yet.

## 2. The reported coverage was thrown on the floor

`poly_slice.rs:396-441`. A test now pins `scale`, `butterfly_forward` and `butterfly_inverse` to report `len - len % WIDTH` at every length from an empty prefix to four whole registers.

The predicate is restated in the test — `PACKED_BUILD`, a `cfg!` of the same target features — rather than read from the kernel's own gate, which is what makes a drifted gate a failure instead of a silent fallback. That is the failure mode you named, so I checked it directly: mistyping `vpclmulqdq` in `mod wide`'s gate on an `+avx2,+vpclmulqdq` build gives

```
failures:
    poly_slice::tests::the_packed_prefix_covers_every_whole_register
test result: FAILED. 173 passed; 1 failed
```

One test fails, and it is the new one. Every correctness test stays green, which is the point.

## 3. The guard walls were asymmetric

`poly_slice.rs:225-289`. Both walls are now `SENTINELS = 8` — two registers of the widest build — and the leading pad is `SENTINELS + offset` sentinels rather than `offset` zeros, so the payload still walks every alignment while the wall in front of it is real.

A/B on the same two mutations, `+avx512f,+vpclmulqdq`, `WIDTH = 4`:

| injected fault | shipped fixture | this fixture |
| --- | --- | --- |
| extra store at `block - WIDTH` | `SIGABRT` (`free(): invalid size`) | `assertion failed: buffer[..start] ... == SENTINEL` |
| extra store at `block + 2*WIDTH` | `SIGABRT` (`free(): invalid size`) | `assertion failed: buffer[start + len..] ... == SENTINEL` |

Both directions were heap corruption before and are clean assertions now.

## 4. `debug_assert_eq!` where the wrappers use `assert_eq!`

Promoted to `assert_eq!` in both butterflies, `poly_slice.rs:30` and `:52`. The public wrappers assert first so this is free on the current path, and a future crate-internal caller gets the guarantee instead of a truncated pass. The `# Panics` docs on the public entry points already say so.

## 5. Does the broadcast `tail()` regress the non-benchmarked shapes?

No. `+avx2,+vpclmulqdq`, `x86_64-unknown-linux-gnu`:

```
RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" \
  cargo rustc --release -p p3-binary-field --lib -- --emit asm
```

LLVM constant-folds `_mm_set_epi64x(0, 0x87)` and the lane broadcast into one 8-byte `.rodata` entry and hoists the load into the prologue:

```asm
	vpbroadcastq	.LCPI4_0(%rip), %ymm1     ; prologue, once
	...
.LBB4_5:                                          ; 2x unrolled loop
	vmovdqu	(%rcx), %ymm3
	vpclmulqdq	$0,  %ymm0, %ymm3, %ymm5
	vpclmulqdq	$1,  %ymm2, %ymm3, %ymm6
	vpxor	%ymm5, %ymm6, %ymm5
	vpclmulqdq	$16, %ymm0, %ymm3, %ymm6
	vpclmulqdq	$17, %ymm2, %ymm3, %ymm3
	...
	vpclmulqdq	$1,  %ymm1, %ymm3, %ymm3   ; tail from ymm1, no memory operand
```

```text
.LCPI4_0:
	.quad	135
```

Per kernel, counted inside each symbol body: zero calls, zero SIMD spills to the stack, one `.rodata` load of the tail, all of it outside the loop. That is the same codegen the array-constant transmute gave on Linux, so there is nothing left to regress — the broadcast form only removed the darwin pathology.

## 6. The measurement stopped one layer short

Fair. `p3-binary-pcs`, `2^16`, the shape `binary_pcs_large_configuration` uses, `-C target-cpu=native`. Two pre-built binaries from `f4e958d3^` and `f4e958d3`, three interleaved rounds, minimum of three:

| | before | after | ratio |
| --- | --- | --- | --- |
| `commit/16` | 20.31 ms | 18.09 ms | 1.12x |
| `open/16` | 28.09 ms | 27.98 ms | 1.00x |

A 4.5x kernel is 1.12x on commit and nothing measurable on open. Your read was right: the win largely dissolves into the DFT's memory traffic at the shape the consumer actually runs. **The host was running several other builds during these rounds**, so treat 1.12x as an upper bound on the noise floor rather than a clean number.

And you are right that `binary_pcs_large_configuration` runs `+pclmulqdq` only, so CI does not exercise this code there either.

## The A/B I owed you from the first round

I said I would either run the three-arm bench or open it as a follow-up. Running it, since this is the follow-up. Slice scaling, `-C target-cpu=native`, four lanes, minimum of three rounds:

| len | scalar loop | split kernel | `broadcast + Mul` | split vs broadcast |
| 16 | 39.35 ns | 19.37 ns | 18.50 ns | 0.96x |
| 256 | 563.0 ns | 180.8 ns | 183.3 ns | 1.01x |
| 4096 | 9.239 µs | 2.782 µs | 2.823 µs | 1.01x |
| 65536 | 141.0 µs | 44.08 µs | 44.64 µs | 1.01x |

Your accounting was exactly right. The split is 5 clmuls per register and `PackedGhash128::mul` is also 5 once the scalar is broadcast and `mixed_y` hoisted, and the shuffle-and-xor difference is free behind `VPCLMULQDQ` throughput on Zen 5. **The split does not beat a ~40-line broadcast-and-multiply loop at any length**, and both are ~3.2x the scalar loop.

So the split has not earned its surface, and I am not going to pretend otherwise. The same load caveat applies to these rounds, so I would like to re-run on a quiet host before acting; if it holds, the right change is to delete the split and keep the dispatch, which is a strict simplification. I will open that as its own PR rather than fold a rewrite into a review-fix. Happy to do it the other way if you prefer.

## Test plan

- [x] `cargo test -p p3-binary-field` on five feature legs — `-C target-cpu=native`, `+avx2,+vpclmulqdq`, `+avx2`, `+pclmulqdq`, none. 0 failures on each. 174 tests on the wide legs against 154 on the portable one; the 20 extra are the packed suite that only exists there.
- [x] `cargo clippy -p p3-binary-field --all-targets -- -D warnings` — clean on all five
- [x] `cargo test -p p3-binary-field --doc` — ok
- [x] `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo doc --no-deps -p p3-binary-field --document-private-items` — clean
- [x] Four fault injections, each run and each restored: two wrong intrinsics, one drifted cfg gate, two out-of-bounds stores. Tables above.

Host: AMD Ryzen 9 9950X3D (Zen 5, AVX-512 + VPCLMULQDQ), 32 threads, shared with other builds throughout.

### DIFF
--- a/binary-field/src/packed/x86_64.rs
+++ b/binary-field/src/packed/x86_64.rs
@@ -548,9 +548,17 @@ mod tests {
         unsafe { lanes::load(values.as_ptr()) }
     }
 
-    /// Every lane operation on the register, against the same operation on the scalar model.
+    /// The two 64-bit halves of every lane, exchanged.
+    fn swapped_halves(values: [u128; WIDTH]) -> [u128; WIDTH] {
+        // Rotating a 128-bit value by half its width is exactly the exchange.
+        core::array::from_fn(|i| values[i].rotate_left(64))
+    }
+
+    /// Each lane-local operation on the register, against the same operation on the model.
     ///
     /// One assertion per operation, so a mismatch names the intrinsic that disagreed.
+    ///
+    /// The whole-element interleave is the one left out, since it crosses lanes by design.
     fn lanes_conform(
         a: [u128; WIDTH],
         b: [u128; WIDTH],
@@ -584,6 +592,11 @@ mod tests {
             mx.clmul::<HIGH_BY_HIGH>(my).0
         );
 
+        // Not a trait method, so the model cannot supply the expectation.
+        //
+        // It carries the only hand-written shuffle immediate here, which is why it is pinned.
+        prop_assert_eq!(lanes_of(lanes::swap_halves(x)), swapped_halves(a));
+
         // The two composites built from those operations.
         //
         // A lane-crossing slip therefore shows up here as well as in the primitives above.
@@ -598,14 +611,31 @@ mod tests {
 
     #[test]
     fn the_register_matches_the_scalar_model_at_the_corners() {
-        // Invariant: each lane operation is lane-local, so distinct lanes must stay distinct.
+        // Invariant: each operation is lane-local, so distinct lanes must stay distinct.
         //
-        // Filling the lanes by rotating through the corners pairs a different extreme in each.
-        for (i, &scalar) in SPECIAL.iter().enumerate() {
-            let a = core::array::from_fn(|lane| SPECIAL[(lane + i) % SPECIAL.len()]);
-            let b = core::array::from_fn(|lane| SPECIAL[(lane + i + 1) % SPECIAL.len()]);
+        // Lane 0 carries the pair under test, so all sixteen combinations are reached.
+        //
+        // That includes the squaring-shaped case where both operands are the same value.
+        //
+        // The other lanes rotate through the corners, so a value that crosses a lane
+        // boundary lands on a different extreme and shows as a mismatch.
+        for (i, &x) in SPECIAL.iter().enumerate() {
+            for (j, &y) in SPECIAL.iter().enumerate() {
+                let a = core::array::from_fn(|lane| match lane {
+                    0 => x,
+                    _ => SPECIAL[(i + lane) % SPECIAL.len()],
+                });
+                let b = core::array::from_fn(|lane| match lane {
+                    0 => y,
+                    _ => SPECIAL[(j + lane) % SPECIAL.len()],
+                });
 
-            lanes_conform(a, b, scalar).unwrap_or_else(|e| panic!("scalar {scalar:#x}: {e}"));
+                // The multiplier walks the corners too, so the split runs from each extreme.
+                for scalar in SPECIAL {
+                    lanes_conform(a, b, scalar)
+                        .unwrap_or_else(|e| panic!("a {x:#x}, b {y:#x}, scalar {scalar:#x}: {e}"));
+                }
+            }
         }
     }
 
--- a/binary-field/src/packed/x86_64.rs
+++ b/binary-field/src/packed/x86_64.rs
@@ -504,7 +504,13 @@ unsafe impl PackedFieldPow2 for PackedGhash128 {
 mod tests {
     use p3_field::PackedValue;
     use p3_field_testing::test_packed_binary_field;
+    use proptest::prelude::*;
 
+    use super::lanes::{self, WIDTH};
+    use crate::packed::split::model::Model;
+    use crate::packed::split::{
+        HIGH_BY_HIGH, HIGH_BY_LOW, LOW_BY_HIGH, LOW_BY_LOW, Lanes, SplitScalar, fold_shifted,
+    };
     use crate::{Ghash128, PackedGhash128};
 
     /// The bit patterns a random search is unlikely to reach.
@@ -522,6 +528,101 @@ mod tests {
     /// One extreme bit pattern per lane.
     fn specials() -> PackedGhash128 {
         PackedValue::from_fn(|i| Ghash128::from_le_bytes(SPECIAL[i].to_le_bytes()))
+    }
+
+    /// The elements a register holds, read back one per lane.
+    fn lanes_of(register: lanes::Reg) -> [u128; WIDTH] {
+        let mut out = [0u128; WIDTH];
+
+        // SAFETY: the destination is exactly one register of contiguous 128-bit integers.
+        //
+        // The store is the unaligned form, so the array's own alignment is irrelevant.
+        unsafe { lanes::store(out.as_mut_ptr(), register) };
+
+        out
+    }
+
+    /// A register holding the given elements, one per lane.
+    fn register_of(values: [u128; WIDTH]) -> lanes::Reg {
+        // SAFETY: the source is exactly one register of contiguous 128-bit integers.
+        unsafe { lanes::load(values.as_ptr()) }
+    }
+
+    /// Every lane operation on the register, against the same operation on the scalar model.
+    ///
+    /// One assertion per operation, so a mismatch names the intrinsic that disagreed.
+    fn lanes_conform(
+        a: [u128; WIDTH],
+        b: [u128; WIDTH],
+        scalar: u128,
+    ) -> Result<(), TestCaseError> {
+        let (x, y) = (register_of(a), register_of(b));
+        let (mx, my) = (Model(a), Model(b));
+
+        prop_assert_eq!(lanes_of(lanes::Reg::zero()), Model::<WIDTH>::zero().0);
+        prop_assert_eq!(
+            lanes_of(lanes::Reg::broadcast(scalar)),
+            Model::<WIDTH>::broadcast(scalar).0
+        );
+        prop_assert_eq!(lanes_of(lanes::Reg::tail()), Model::<WIDTH>::tail().0);
+        prop_assert_eq!(lanes_of(x.xor(y)), mx.xor(my).0);
+        prop_assert_eq!(lanes_of(x.unpack_low_64(y)), mx.unpack_low_64(my).0);
+        prop_assert_eq!(
+            lanes_of(x.clmul::<LOW_BY_LOW>(y)),
+            mx.clmul::<LOW_BY_LOW>(my).0
+        );
+        prop_assert_eq!(
+            lanes_of(x.clmul::<HIGH_BY_LOW>(y)),
+            mx.clmul::<HIGH_BY_LOW>(my).0
+        );
+        prop_assert_eq!(
+            lanes_of(x.clmul::<LOW_BY_HIGH>(y)),
+            mx.clmul::<LOW_BY_HIGH>(my).0
+        );
+        prop_assert_eq!(
+            lanes_of(x.clmul::<HIGH_BY_HIGH>(y)),
+            mx.clmul::<HIGH_BY_HIGH>(my).0
+        );
+
+        // The two composites built from those operations.
+        //
+        // A lane-crossing slip therefore shows up here as well as in the primitives above.
+        prop_assert_eq!(lanes_of(fold_shifted(x, y)), fold_shifted(mx, my).0);
+        prop_assert_eq!(
+            lanes_of(SplitScalar::new(scalar).apply(x)),
+            SplitScalar::new(scalar).apply(mx).0
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn the_register_matches_the_scalar_model_at_the_corners() {
+        // Invariant: each lane operation is lane-local, so distinct lanes must stay distinct.
+        //
+        // Filling the lanes by rotating through the corners pairs a different extreme in each.
+        for (i, &scalar) in SPECIAL.iter().enumerate() {
+            let a = core::array::from_fn(|lane| SPECIAL[(lane + i) % SPECIAL.len()]);
+            let b = core::array::from_fn(|lane| SPECIAL[(lane + i + 1) % SPECIAL.len()]);
+
+            lanes_conform(a, b, scalar).unwrap_or_else(|e| panic!("scalar {scalar:#x}: {e}"));
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(256))]
+
+        #[test]
+        fn the_register_matches_the_scalar_model(
+            a in prop::array::uniform(any::<u128>()),
+            b in prop::array::uniform(any::<u128>()),
+            scalar in any::<u128>(),
+        ) {
+            // The seam the scalar model alone cannot reach.
+            //
+            // Each intrinsic must be the one the algebra was written against, lane by lane.
+            lanes_conform(a, b, scalar)?;
+        }
     }
 
     test_packed_binary_field!(
--- a/binary-field/src/poly_slice.rs
+++ b/binary-field/src/poly_slice.rs
@@ -27,8 +27,10 @@ pub(crate) fn scale(values: &mut [u128], scalar: u128) {
 /// Send `(lo, hi)` to `(lo + scalar*hi, lo + (scalar + 1)*hi)`, element by element.
 #[inline]
 pub(crate) fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
-    // Invariant: the two halves are paired element for element, so a mismatch drops work.
-    debug_assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
+    // Invariant: the two halves are paired element for element.
+    //
+    // Both kernels below stop at the shorter side, so a mismatch would silently drop work.
+    assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
 
     let packed = if lo.len() < wide::WIDTH {
         0
@@ -47,8 +49,10 @@ pub(crate) fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) 
 /// Undo [`butterfly_forward`] with the same scalar.
 #[inline]
 pub(crate) fn butterfly_inverse(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
-    // Invariant: the two halves are paired element for element, so a mismatch drops work.
-    debug_assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
+    // Invariant: the two halves are paired element for element.
+    //
+    // Both kernels below stop at the shorter side, so a mismatch would silently drop work.
+    assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
 
     let packed = if lo.len() < wide::WIDTH {
         0
@@ -218,8 +222,23 @@ mod tests {
     /// ```
     const CORNERS: [u128; 7] = [0, 1, u128::MAX, 1 << 127, 1 << 64, (1u128 << 64) - 1, 0x87];
 
-    /// Guard elements placed after the payload, one full register of the widest build.
-    const SENTINELS: usize = 4;
+    /// Whether this build is one the packed prefix is compiled for.
+    ///
+    /// Stated a second time here, independently of the gate on the kernels themselves.
+    ///
+    /// A gate that drifts then shows up as a failure rather than as a silent scalar fallback.
+    const PACKED_BUILD: bool = cfg!(all(
+        target_arch = "x86_64",
+        target_feature = "vpclmulqdq",
+        any(target_feature = "avx2", target_feature = "avx512f")
+    ));
+
+    /// Guard elements on each side of the payload, two full registers of the widest build.
+    ///
+    /// One register wide would catch only a store that overran by exactly one register.
+    ///
+    /// Two make the next one out detectable as well, still inside the buffer's own allocation.
+    const SENTINELS: usize = 8;
 
     /// What those elements hold, which no product of the inputs below can reproduce.
     const SENTINEL: u128 = 0x5a5a_5a5a_5a5a_5a5a_5a5a_5a5a_5a5a_5a5a;
@@ -233,28 +252,39 @@ mod tests {
 
     /// The same values, walled off from the rest of the allocation on both sides.
     ///
-    /// One register load covers several elements.
+    /// Both walls are sentinels, so an overrun in either direction shows as a changed guard.
     ///
-    /// Leading zeros therefore walk the payload across every alignment it can have.
-    ///
-    /// Trailing sentinels catch a store that ran a whole register past the end.
+    /// ```text
+    ///     [ guard | offset | values | guard ]
+    ///                ^ slides the payload across every alignment one register can see
+    /// ```
     fn padded(values: &[u128], offset: usize) -> Vec<u128> {
-        let mut buffer = Vec::with_capacity(offset + values.len() + SENTINELS);
-        buffer.extend(core::iter::repeat_n(0, offset));
+        let lead = SENTINELS + offset;
+
+        let mut buffer = Vec::with_capacity(lead + values.len() + SENTINELS);
+        buffer.extend(core::iter::repeat_n(SENTINEL, lead));
         buffer.extend_from_slice(values);
         buffer.extend(core::iter::repeat_n(SENTINEL, SENTINELS));
         buffer
     }
 
+    /// The payload of such a buffer, as the kernels see it.
+    fn payload_mut(buffer: &mut [u128], offset: usize, len: usize) -> &mut [u128] {
+        let start = SENTINELS + offset;
+        &mut buffer[start..start + len]
+    }
+
     /// The payload of such a buffer, once both walls are confirmed intact.
     fn payload(buffer: &[u128], offset: usize, len: usize) -> Result<&[u128], TestCaseError> {
+        let start = SENTINELS + offset;
+
         // Nothing may have run off the front.
-        prop_assert!(buffer[..offset].iter().all(|&v| v == 0));
+        prop_assert!(buffer[..start].iter().all(|&v| v == SENTINEL));
 
         // Nothing may have run off the back either.
-        prop_assert!(buffer[offset + len..].iter().all(|&v| v == SENTINEL));
+        prop_assert!(buffer[start + len..].iter().all(|&v| v == SENTINEL));
 
-        Ok(&buffer[offset..offset + len])
+        Ok(&buffer[start..start + len])
     }
 
     /// Scaling, one element at a time through the scalar backend.
@@ -298,7 +328,7 @@ mod tests {
 
         // Scaling: the packed prefix and the scalar tail must together match the scalar loop.
         let mut scaled = padded(values, offset);
-        super::scale(&mut scaled[offset..offset + len], scalar);
+        super::scale(payload_mut(&mut scaled, offset, len), scalar);
         prop_assert_eq!(
             payload(&scaled, offset, len)?,
             &reference_scale(values, scalar)[..]
@@ -308,8 +338,8 @@ mod tests {
         let mut lo = padded(values, offset);
         let mut hi = padded(&other, offset);
         super::butterfly_forward(
-            &mut lo[offset..offset + len],
-            &mut hi[offset..offset + len],
+            payload_mut(&mut lo, offset, len),
+            payload_mut(&mut hi, offset, len),
             scalar,
         );
         let (want_lo, want_hi) = reference_forward(values, &other, scalar);
@@ -318,8 +348,8 @@ mod tests {
 
         // The inverse kernel applied to the forward output must return the input.
         super::butterfly_inverse(
-            &mut lo[offset..offset + len],
-            &mut hi[offset..offset + len],
+            payload_mut(&mut lo, offset, len),
+            payload_mut(&mut hi, offset, len),
             scalar,
         );
         prop_assert_eq!(payload(&lo, offset, len)?, values);
@@ -329,8 +359,8 @@ mod tests {
         let mut lo = padded(values, offset);
         let mut hi = padded(&other, offset);
         super::butterfly_inverse(
-            &mut lo[offset..offset + len],
-            &mut hi[offset..offset + len],
+            payload_mut(&mut lo, offset, len),
+            payload_mut(&mut hi, offset, len),
             scalar,
         );
         let (want_lo, want_hi) = reference_inverse(values, &other, scalar);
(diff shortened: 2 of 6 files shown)


binary-field/src/packed/x86_64.rs
```diff
@@ -524,6 +530,101 @@ mod tests {
...
+
+            lanes_conform(a, b, scalar).unwrap_or_else(|e| panic!("scalar {scalar:#x}: {e}"));
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(256))]
+
+        #[test]
+        fn the_register_matches_the_scalar_model(
```
Nashtare, line 646: This is the right test and it closes my option 1. One correction to the reason given for not pairing it with a CI change, because it is the second round it has come up.

> the matrix comment says a leg's job name embeds its `features` and two of those names are required status checks, so flipping one to test-running is a branch-protection change

The two names that comment refers to are the **`+avx512f`** legs — it sits directly above them at `ci.yml:51-53`. The vpclmulqdq legs are not among them:

```
$ gh api repos/Plonky3/Plonky3/branches/main/protection \
    --jq '.required_status_checks.contexts'
["Formatting and Clippy", "Build embedded", "Test wasm32 simd128",
 "Build and Test (macos-latest, false)", "Build and Test (macos-latest, true)",
 "Build and Test (ubuntu-24.04-arm, +sve2, true, aarch64-unknown-linux-gnu, false)", ...,
 "Build and Test (ubuntu-latest, +avx512f, true, x86_64-unknown-linux-gnu, false)",
 "Build and Test (ubuntu-latest, +avx512f, true, x86_64-unknown-linux-gnu, true)",
 "Build and Test (ubuntu-latest, +avx2, true)", "Build and Test (ubuntu-latest, +avx2, false)",
 "Build and Test (windows-latest, ...)", "Build and Test (ubuntu-latest, ...)"]
```

No `+avx2,+vpclmulqdq`, no `+avx512f,+vpclmulqdq`, no `+pclmulqdq`. Renaming a vpclmulqdq leg touches nothing anyone is gating on.

The *real* obstacle is the one the matrix comment actually states — runner hardware — and it is untouched by the above. Two notes on it, since this test not running is the difference between the PR having a centrepiece and having a compile check:

- A hardware-conditional skip (`grep vpclmulqdq /proc/cpuinfo` around the test step) is a check that goes green by doing nothing, so it would not catch the bug class this test exists for. And `is_x86_feature_detected!` — my own suggestion last round — cannot help at all: `mod x86_64` is `target_feature`-gated at compile time, so a runner without the instruction SIGILLs before any runtime check executes.
- An additive emulator job does close it, is not a rename, and the matrix comment explicitly endorses the shape ("New feature combinations go in legs of their own"). There is already precedent in this file: `Test wasm32 simd128` runs tests under wasmtime in a job of its own.

Scope correction on my `qemu-x86_64 -cpu max` suggestion, since it was only half right: TCG has no AVX-512 at all, so qemu reaches the `+avx2,+vpclmulqdq` backend only — the 2-lane half of what this test covers. The 4-lane backend needs real Ice Lake / Zen 3+ silicon or Intel SDE (`sde64 -future`, x86_64 host only, which a runner is). Half the coverage for one additive job is still worth taking first.

binary-field/src/packed/x86_64.rs
```diff
@@ -524,6 +530,101 @@ mod tests {
...
+        out
+    }
+
+    /// A register holding the given elements, one per lane.
+    fn register_of(values: [u128; WIDTH]) -> lanes::Reg {
+        // SAFETY: the source is exactly one register of contiguous 128-bit integers.
+        unsafe { lanes::load(values.as_ptr()) }
+    }
+
+    /// Every lane operation on the register, against the same operation on the scalar model.
```
Nashtare: "Every lane operation on the register" is every `Lanes` operation. `lanes::swap_halves` is not on the trait and is not covered here.

It is the one I would have reached for first: it carries the only hand-written immediate in the file (`SWAP_QUADWORDS = 0x4e`, `:30`) and it is live in `Mul` (`:361-362`) and `dot_product` (`:425-426`). I checked it rather than assuming — `VPSHUFD` is lane-wise at all three widths and `0x4E = 0b01_00_11_10` selects source dwords `[2, 3, 0, 1]`, i.e. a swap of the two 64-bit halves within each 128-bit lane, so the 512-bit form matches the 256-bit one and `x ^ swap_halves(x)` really does give `x_lo ^ x_hi` in both halves. Nothing is wrong, and it is reachable end to end through `test_packed_binary_field!`.

But end-to-end reachability is exactly the standard this PR is arguing against. Either add a model expectation — `(v >> 64) | (v << 64)` per lane, three lines, and it does not need to join the trait — or narrow the docstring to "every `Lanes` operation" so the claim matches the code.

binary-field/src/packed/x86_64.rs
```diff
@@ -524,6 +530,101 @@ mod tests {
...
+    ) -> Result<(), TestCaseError> {
+        let (x, y) = (register_of(a), register_of(b));
+        let (mx, my) = (Model(a), Model(b));
+
+        prop_assert_eq!(lanes_of(lanes::Reg::zero()), Model::<WIDTH>::zero().0);
+        prop_assert_eq!(
+            lanes_of(lanes::Reg::broadcast(scalar)),
+            Model::<WIDTH>::broadcast(scalar).0
+        );
+        prop_assert_eq!(lanes_of(lanes::Reg::tail()), Model::<WIDTH>::tail().0);
```
Nashtare, line 575: Worth recording, because the asm quoted in the description looks wrong at a glance and someone will trip on it later.

```asm
vpbroadcastq .LCPI4_0(%rip), %ymm1
.LCPI4_0:
        .quad 135
```

`vpbroadcastq` replicates one qword to all four qwords of the ymm, giving `[0x87, 0x87, 0x87, 0x87]` — not the `[0x87, 0, 0x87, 0]` that `_mm_set_epi64x(0, 0x87)` plus a 128-bit lane broadcast produces. Those are different registers, so the codegen appears to contradict the source.

It is correct, and the reason is this line's own subject: `tail()` has exactly one non-test caller, `fold_shifted`'s `t1.clmul::<HIGH_BY_LOW>(L::tail())` (`split.rs:72`), and `HIGH_BY_LOW = 0x01` has bit 4 clear, so only the **low** qword of the tail register is ever read. LLVM proved the high qwords dead and picked the cheaper broadcast. Nothing to change — the item-5 evidence stands.

One genuine consequence: because both sides take the trait default, this assertion is `Reg::broadcast(0x87)` against `Model::broadcast(0x87)`, so it is subsumed by the broadcast assertion above whenever `scalar == TAIL_128` and otherwise adds nothing beyond a second broadcast at a fixed constant. Cheap, so keeping it is fine, but it is not independent coverage.

binary-field/src/packed/x86_64.rs
```diff
@@ -524,6 +530,101 @@ mod tests {
...
+        Ok(())
+    }
+
+    #[test]
+    fn the_register_matches_the_scalar_model_at_the_corners() {
+        // Invariant: each lane operation is lane-local, so distinct lanes must stay distinct.
+        //
+        // Filling the lanes by rotating through the corners pairs a different extreme in each.
+        for (i, &scalar) in SPECIAL.iter().enumerate() {
+            let a = core::array::from_fn(|lane| SPECIAL[(lane + i) % SPECIAL.len()]);
```
Nashtare: Nit. Because `b` is always `a` rotated by exactly one lane, the operand pairs reached are only `(SPECIAL[k], SPECIAL[k+1])` — 4 of the 16 corner-by-corner combinations. `(u128::MAX, u128::MAX)`, `(0, 1 << 127)` and `(0x87, 0x87)` never share a lane, and the squaring-shaped case `a == b` never occurs at all.

The 256-case proptest below covers the generic space, so this is not a hole so much as a missed freebie: a nested `for x in SPECIAL { for y in SPECIAL { .. } }` filling every lane from `(x, y)` costs less than a single proptest case and reaches all 16.

binary-field/src/poly_slice.rs
```diff
@@ -218,8 +222,23 @@ mod tests {
...
     const CORNERS: [u128; 7] = [0, 1, u128::MAX, 1 << 127, 1 << 64, (1u128 << 64) - 1, 0x87];
 
-    /// Guard elements placed after the payload, one full register of the widest build.
-    const SENTINELS: usize = 4;
+    /// Whether this build is one the packed prefix is compiled for.
+    ///
+    /// Stated a second time here, independently of the gate on the kernels themselves.
+    ///
+    /// A gate that drifts then shows up as a failure rather than as a silent scalar fallback.
+    const PACKED_BUILD: bool = cfg!(all(
```
Nashtare, line 234: This is the right shape, and the restatement rather than a re-export is the whole point — a gate read from the kernel's own `cfg` could not detect drift in that `cfg`. Confirming I checked the mutation rather than taking the table on faith:

- mistype `mod wide`'s gate only → the `not(...)` fallback compiles, `WIDTH == usize::MAX`, and `assert!(width == 2 || width == 4)` below fails. One new failure, correctness tests all green. Matches the output quoted in the description.
- mistype the fallback gate only → both modules compile, duplicate definition, build error.
- mistype both identically → back to the first case.

So the drift is caught in every direction. I also checked that this predicate is character-identical to `mod wide`'s gate at `:71-75` and to `mod x86_64`'s gate at `packed/mod.rs:9-13`, which is what makes the new conformance test in `x86_64.rs` compile on exactly the builds the kernel does.

binary-field/src/poly_slice.rs
```diff
@@ -218,8 +222,23 @@ mod tests {
...
+        target_feature = "vpclmulqdq",
+        any(target_feature = "avx2", target_feature = "avx512f")
+    ));
+
+    /// Guard elements on each side of the payload, two full registers of the widest build.
+    ///
+    /// One register wide would catch only a store that overran by exactly one register.
+    ///
+    /// Two make the next one out detectable as well, still inside the buffer's own allocation.
+    const SENTINELS: usize = 8;
```
Nashtare: Both A/B rows in the description follow from the geometry, which I checked rather than assumed. The old fixture's leading pad was `offset` zeros with `offset in 0..4`, so on the 4-lane build a full-register under-run from block 0 wrote before the `Vec`'s buffer at *every* offset the sweep used; and the trailing wall was 4 elements = one register against a `Vec::with_capacity` allocated to the exact size, so two registers past the end left the allocation. `SENTINELS = 8` on both sides covers both. Good fix.

Nit on the constant: `8` encodes "two registers of the widest build" as a literal, so if a wider packing ever lands (`WIDTH == 8`) the guarantee silently halves back to one register with no signal. The docstring records the assumption, which is most of the value; `2 * 4` with the 4 named, or deriving it, would keep it true rather than documented.

One inherent limit worth knowing about rather than fixing: an overrun is only visible when the written value differs from `SENTINEL`, so `scale` with `scalar == 1` — a `CORNERS` entry — writes the sentinel straight back. That is the identity map and no value-based wall can see it; the other six corners do.

binary-field/src/poly_slice.rs
```diff
@@ -363,6 +393,55 @@ mod tests {
...
+        if !PACKED_BUILD {
+            // No packing here, so the width is the sentinel that sends every slice to the tail.
+            assert_eq!(super::wide::WIDTH, usize::MAX);
+            return;
+        }
+
+        let width = super::wide::WIDTH;
+
+        // Two lanes under the 256-bit register, four under the 512-bit one, and nothing else.
+        assert!(
```
Nashtare: This is the line that actually catches gate drift, and it compares two compile-time constants at runtime. `const { assert!(width == 2 || width == 4, ..) }` would turn the drift into a build failure instead of a test failure, which is strictly earlier and does not depend on anyone running this leg — which, per the module-level thread, nobody currently does.

Separately, and only so the test's value is not overstated: the loop below checks that `(len / W) * W == len - len % W`, where the kernels compute their return from `as_chunks_mut` directly. That is close to a tautology. The bug it does catch is a real class — returning `blocks.len()` instead of `blocks.len() * WIDTH` — but the existing correctness tests catch that one too, since the scalar tail would then re-process or skip elements. The novel coverage here is the `WIDTH` assertions above, not the loop.

binary-field/src/poly_slice.rs
```diff
@@ -363,6 +393,55 @@ mod tests {
...
+            width == 2 || width == 4,
+            "unexpected register width {width}"
+        );
+
+        // Every length from an empty prefix to several whole registers plus a tail.
+        for len in 0..4 * width {
+            // Whole registers only: the remainder is the scalar tail's business.
+            let want = len - len % width;
+
+            let mut values = sample(len);
```
Nashtare: These call `wide::*` on bare `sample(len)` vectors, at every length in `0..4 * width`. `(0..len).map(..).collect()` gives a `Vec` whose capacity is exactly `len`, so this is the unguarded-buffer shape that item 3 of this same PR just finished removing from the other test — an overrun here is heap corruption, not an assertion.

The walled test covers overlapping lengths (`0..=17` against `0..16` on the 4-lane build), so nothing is unguarded in practice today. But routing these three through `padded` / `payload_mut` costs three lines and buys overrun detection at every length this sweep already visits, which is strictly more than the walled test's `offset in 0..4` gives you.

binary-field/src/poly_slice.rs
```diff
@@ -27,8 +27,10 @@ pub(crate) fn scale(values: &mut [u128], scalar: u128) {
 /// Send `(lo, hi)` to `(lo + scalar*hi, lo + (scalar + 1)*hi)`, element by element.
 #[inline]
 pub(crate) fn butterfly_forward(lo: &mut [u128], hi: &mut [u128], scalar: u128) {
-    // Invariant: the two halves are paired element for element, so a mismatch drops work.
-    debug_assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
+    // Invariant: the two halves are paired element for element.
+    //
+    // Both kernels below stop at the shorter side, so a mismatch would silently drop work.
+    assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
```
Nashtare, line 33: Agreed with the promotion. Confirming the reasoning holds in the direction that matters: this is unreachable from every caller that exists — `poly_basis::butterfly_forward` (`:65`) and `butterfly_inverse` (`:88`) assert first, and nothing else calls into here — so it is genuinely free on the current path and LLVM will fold the second comparison given the `#[inline]`. It is defence-in-depth for the future caller, which is what it was asked to be.

The flip side is that no test exercises it: `forward_rejects_mismatched_lengths` and `inverse_rejects_mismatched_lengths` hit the wrapper's assert, never this one. Not worth a test, just worth knowing it is unverified.

With this in place, `blocks = lo.len().min(hi.len())` in both `wide` kernels is now provably `lo.len()` for every reachable call — dead defensive code rather than a live branch. Fine to leave; flagging it because the `.min()` is the thing the comment above cites as the reason for the assert, and once the assert exists the two are redundant with each other.

Nashtare (approved):

binary-field/src/packed/x86_64.rs
tcoratger: You are right and I was wrong, twice. The matrix comment sits at `ci.yml:51-53`, directly above the two `+avx512f` legs, and says "the two below" — the vpclmulqdq legs are not what it is talking about. I carried that reason over from the wrong pair of legs and then repeated it. Thanks for chasing it down instead of letting it stand.

I could not confirm the protection list myself (`gh api repos/Plonky3/Plonky3/branches/main/protection` gives me a 404 — no admin), so I am taking your output. What I could confirm is the job-name shape, which matches your point:

```
Build and Test (ubuntu-latest, +avx512f, true, x86_64-unknown-linux-gnu, false)
Build and Test (ubuntu-latest, +avx2,+vpclmulqdq, true, x86_64-unknown-linux-gnu, false)
```

Added the additive job. `ci.yml`, `check_x86_emulated` / **Test x86 vpclmulqdq (emulated)**, next to the wasmtime job it is modelled on:

```yaml
    env:
      target: x86_64-unknown-linux-gnu
      features: "+avx2,+vpclmulqdq"
      CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER: qemu-x86_64 -cpu max
```

Not a rename, no matrix leg touched, so nothing anyone gates on moves. The vpclmulqdq legs keep their names and their `compile_only`, and their comment now points at the job.

Your scope correction is taken as the job's own documentation rather than buried in a PR description — the env block says TCG reaches the two-lane backend and that the four-lane one still needs real silicon. I did not add SDE; one additive job at a time, and the 4-lane half is the one I would want to see green before adding a second.

Two things I did have to change to make it work, both small:

- `scripts/check.py test` grew a `--target` passthrough. Without it Cargo applies the pinned `RUSTFLAGS` to build scripts and proc macros too, and the runner executes those natively — the same reason the compile-only legs already pass `--target`. New case in `scripts/test_check.py` pins both spellings; 24 tests pass.
- Nothing else. The job is one `check.py test --package p3-binary-field` call.

**What I could not verify locally**: qemu itself. No `qemu-user` on this box and no root to install it, so this PR's own CI run is the first execution. What I did verify is the exact command with the emulator out of the picture:

```
$ python3 scripts/check.py test --package p3-binary-field \
    --target x86_64-unknown-linux-gnu --target-feature=+avx2,+vpclmulqdq
+ RUSTFLAGS += -C target-feature=+avx2,+vpclmulqdq
+ cargo nextest run --target x86_64-unknown-linux-gnu -p p3-binary-field
     Summary [0.930s] 355 tests run: 355 passed, 3 skipped
```

So the recipe is right and the only untested link is the translator. If TCG turns out not to cover it, the job goes red here rather than quietly, and I will pull it.

binary-field/src/packed/x86_64.rs
tcoratger: Fixed rather than narrowed — you picked the right one to reach for.

`swap_halves` is now pinned against a written expectation, and the docstring is corrected to what the function actually claims:

```rust
/// The two 64-bit halves of every lane, exchanged.
fn swapped_halves(values: [u128; WIDTH]) -> [u128; WIDTH] {
    // Rotating a 128-bit value by half its width is exactly the exchange.
    core::array::from_fn(|i| values[i].rotate_left(64))
}
```

It does not join the trait, as you said it needn't. The docstring now reads "each lane-local operation", and names the whole-element interleave as the one deliberate omission, since that one crosses lanes by design.

**Your instinct about end-to-end reachability was better than either of us argued.** I mutated the immediate to check the new assertion had teeth, and got a result worth recording:

| `SWAP_QUADWORDS` | dwords selected | new conformance tests | everything else |
| --- | --- | --- | --- |
| `0x4e` (correct) | `[2,3,0,1]` | pass | pass |
| `0x1e` | `[2,3,1,0]` | **2 fail** | **all pass** |
| `0xe4` (identity) | `[0,1,2,3]` | 2 fail | 5 fail |

`0x1e` corrupts only the *high* half of the shuffle. Karatsuba reads `clmul::<LOW_BY_LOW>(mixed_x, mixed_y)`, so the high half is never looked at, and `test_packed_binary_field!` is completely blind to it — 172 passed, 2 failed, and the 2 are the new ones.

So this was not a hypothetical gap. A wrong immediate in half of that shuffle was shipping-green before this test existed.

binary-field/src/packed/x86_64.rs
tcoratger: Good catch, and thank you for writing the reason down rather than just flagging that it looks wrong — I would not have reconstructed it quickly.

Your reading is right: `tail()`'s only non-test caller is `t1.clmul::<HIGH_BY_LOW>(L::tail())` at `split.rs:72`, `HIGH_BY_LOW = 0x01` has bit 4 clear, so only the low quadword of the tail register is ever read and LLVM is free to fill the rest with whatever is cheapest. The `.rodata` entry is a bare `.quad 135` and `vpbroadcastq` splats it to all four, which is a different register from `[0x87, 0, 0x87, 0]` and equally correct here.

On the assertion being subsumed: agreed, it is `Reg::broadcast(0x87)` against `Model::broadcast(0x87)` since both sides take the trait default. Leaving it in — it costs nothing and it would start earning its place the moment anyone overrides `tail()` on one backend — but I am not counting it as independent coverage, and this thread is now the record of why.

No code change.

binary-field/src/packed/x86_64.rs
tcoratger: Taken. Nested now, and it reaches all sixteen including `a == b`:

```rust
for (i, &x) in SPECIAL.iter().enumerate() {
    for (j, &y) in SPECIAL.iter().enumerate() {
        let a = core::array::from_fn(|lane| match lane {
            0 => x,
            _ => SPECIAL[(i + lane) % SPECIAL.len()],
        });
        let b = ...;                       // same shape from `j`
        for scalar in SPECIAL { lanes_conform(a, b, scalar)?; }
    }
}
```

Lane 0 carries the pair under test so every combination is reached; the remaining lanes keep rotating so a value crossing a lane boundary still lands on a different extreme and shows up. I also walked the multiplier over the corners rather than tying it to the first operand — 64 cases, still far under one proptest case.

binary-field/src/poly_slice.rs
tcoratger: Thanks for running all three directions — the duplicate-definition case is one I had not thought through, and it is the reason the "mistype both identically" case collapses back to the first rather than going undetected.

One change since, from the thread on the width assertion below: the drift is now caught at **build** time rather than by this test.

```
error[E0080]: evaluation panicked: assertion failed: if PACKED_BUILD {
error: could not compile `p3-binary-field` (lib test)
```

Same predicate, same restatement, same reason it has to be a restatement rather than a re-export — just moved to a `const` assertion so it does not wait for anyone to run the leg.

binary-field/src/poly_slice.rs
tcoratger: Both nits taken, and the first one is now enforced rather than documented:

```rust
/// Elements per register on the widest packing this crate has.
///
/// The build assertion below forces a wider one to bump this rather than pass silently.
const WIDEST_LANES: usize = 4;

/// Guard elements on each side of the payload, two of the widest registers.
const SENTINELS: usize = 2 * WIDEST_LANES;
```

and the build assertion is `WIDTH == 2 || WIDTH == WIDEST_LANES`. So a `WIDTH == 8` packing landing does not halve the wall silently — it fails to compile until someone bumps the constant, and bumping it widens the wall automatically. That closes the "true rather than documented" gap you were pointing at.

On the identity case: correct, and it is not fixable with a value wall. `scale` with `scalar == 1` writes the sentinel straight back, so an overrun is invisible for that one corner. The other six drive it, and the sweep visits every one of them at every length, so the fault would have to be scalar-dependent to hide. Recording it here rather than papering over it.

binary-field/src/poly_slice.rs
tcoratger: Done, and it is strictly better — thank you.

```rust
const _: () = assert!(if PACKED_BUILD {
    super::wide::WIDTH == 2 || super::wide::WIDTH == WIDEST_LANES
} else {
    super::wide::WIDTH == usize::MAX
});
```

Both directions, at build time. Re-ran the gate-drift mutation and it no longer reaches the test runner at all:

```
error[E0080]: evaluation panicked: assertion failed: if PACKED_BUILD {
error: could not compile `p3-binary-field` (lib test) due to 2 previous errors
```

The second error is the overflow lint on `4 * WIDTH` once const propagation knows the branch is live — a free extra signal, same cause.

And you are right about the loop: `(len / W) * W == len - len % W` is close to a tautology, and the `blocks.len()` versus `blocks.len() * WIDTH` class it does catch is already caught by the correctness tests. The novel coverage was the `WIDTH` assertions, which is exactly why moving them to `const` and leaving the loop as a cheap cross-check is the right split. I have not overstated it in the description.

binary-field/src/poly_slice.rs
tcoratger: Right, and slightly embarrassing given item 3 of this same PR. Fixed — all three go through `padded` / `payload_mut` now, with the walls checked after each:

```rust
let mut scaled = padded(&values, 0);
let covered = super::wide::scale(payload_mut(&mut scaled, 0, len), 3);
assert_eq!(covered, want, "scale at len {len}");
assert!(walls_intact(&scaled, 0, len), "scale overran at len {len}");
```

The wall check is factored out of the proptest accessor into `walls_intact`, so the two sweeps share one definition of what an intact wall is rather than drifting.

Cheap, and it buys overrun detection at every length in `0..4 * width` rather than relying on the walled test's overlap.

binary-field/src/poly_slice.rs
tcoratger: Agreed on all of it, and I acted on the last paragraph.

The `.min()` is gone from both kernels. With the assert in place it was picking the shorter of two lengths that are provably equal, and the comment above it cited that pick as the reason the assert exists — circular. Now:

```rust
// The dispatcher asserts the two halves match, so either chunk count is the answer.
let blocks = lo.len();
```

and each kernel's doc states the precondition it relies on rather than defending against its absence.

On the untested crate-internal assert: confirmed, and I did not add a test for it. `forward_rejects_mismatched_lengths` and its inverse hit the wrapper, and a `#[should_panic]` here would pass identically whether the assert is `assert_eq!` or `debug_assert_eq!` under `cargo test`, so it would not verify the thing that changed. Left unverified and now written down.

binary-field/src/packed/x86_64.rs
tcoratger: Reopening this: the job I added does not work, and I resolved the thread before it had run. That was premature — sorry.

**QEMU's user-mode translator does not implement the 256-bit carryless multiply.** It decodes the 128-bit form only, so every test that reaches a wide product dies on an illegal instruction. The measured partition, from the run on this PR:

```
Summary [10.317s] 112/355 tests run: 109 passed, 3 failed, 3 skipped

SIGILL  packed::x86_64::tests::packed_field_tests::test_dot_product_boundary
SIGILL  packed::x86_64::tests::packed_field_tests::test_batched_linear_combination
SIGILL  packed::x86_64::tests::packed_field_tests::test_multiplicative_inverse
        qemu: uncaught target signal 4 (Illegal instruction) - core dumped
```

The split is clean once you sort the 112 by whether they reach a `ymm` product:

| | outcome |
| --- | --- |
| reaches a 256-bit product (3 tests) | **all SIGILL** |
| packed but no product — broadcast, interleave (2) | pass |
| the scalar model in `split::tests` (3) | pass |
| scalar modules — clmul, ghash, extension, gf2, cantor (104) | pass |

Nothing that executes the instruction survived, and nothing that survived executes it. `qemu-user` 8.2, `-cpu max`, on `ubuntu-latest`.

The sharpest part: **the conformance test this whole PR is about never ran.** It was in the 243 that nextest cancelled after the first failures, and it calls `clmul` on a register, so it would have SIGILLed too. The job would not have delivered the coverage it existed for even with `--no-fail-fast`.

So this is worse than the half-coverage you scoped it to — it is zero coverage of the packed backend, plus a red check. I have pulled the job and the `--target` passthrough that supported it, since nothing else needed it.

What is left in the tree is the finding, in the comment above the matrix, so the next person does not spend a CI round on it:

```yaml
          # A user-mode emulator is not a way around that. QEMU's translator decodes the
          # 128-bit carryless multiply only, so every test reaching a 256-bit product dies
          # on an illegal instruction. Executing these paths needs real silicon.
```

That leaves your other option, Intel SDE, as the only one still standing — and it is the better one anyway since it reaches four lanes as well as two. I have not added it, for a reason I would rather state than decide unilaterally: it needs either a third-party action or a version-pinned Intel download, and adding a new supply-chain dependency to `ci.yml` is a maintainer call, not something to slip into a review-fix PR. Happy to open it as its own change if you want it, or to leave the vpclmulqdq legs compile-only with the comment now recording why.

Your original point stands unchanged and unaddressed: nothing in CI executes these kernels. I have not closed that, and I am not going to mark it closed again until something actually runs them.

PR #2102, opened 2026-09-09
feat(multi-stark)!: type the statement-level transcript

tcoratger:
Depends on #2081. This PR targets `feat/typed-transcript-sumcheck` so the diff contains only this crate's migration; retarget to `main` after the branch below it lands.

## What this changes

multi-STARK's three sub-protocols each seed their own typed transcript — the zerocheck, the lookup argument, the fractional-GKR reduction, all landed in #2093. The statement they are all about does not.

The preprocessed commitment, the main commitment and the per-instance public values were absorbed with bare `challenger.observe` calls, outside any pattern. So the identity of the batch — how many instances, their arities, their widths, their public-value counts, whether a preprocessed trace exists at all — reached the sponge only incidentally, and the top-level protocol had no identifier, no fingerprint, and no recorded hand-off to the four children it delegates to.

`MultiStarkShape` derives those numbers from the instances and the keys, publishes them as an `InteractionPattern`, and seals them into a `DomainSeparator` under `p3-multi-stark`. `prove` and `verify` drive every statement-level step through it.

## The described sequence

```text
    preprocessed_commitment   opaque, present iff some AIR declares preprocessed columns
    Begin/End main_commitment  the scheme's own commit phase, bracketed
    public_values             one step per instance, Fixed(air.num_public_values())
    Begin/End lookup_argument
    Begin/End zerocheck
    Begin/End main_opening
    Begin/End preprocessed_opening   present iff a preprocessed trace exists
```

No challenge is drawn at statement level — every challenge belongs to a delegated child. That is why the shape is generic over the base field alone, with no extension parameter.

## The main commitment: a bracket, not a hoist

`MultilinearPcs::commit` absorbs the commitment inside the scheme, while the verifier absorbs `proof.commitment` itself. The two bodies are asymmetric.

Hoisting the absorb into the statement transcript would make them symmetric, but it needs the prover to obtain the commitment without letting the scheme touch the real sponge — a cloned challenger whose post-state is thrown away. That is correct only *because* `commit` happens to sample nothing. Nothing in the trait says so, and a scheme that ever sampled during commit would desynchronise silently, where the pattern player could not see it.

So `commit` is bracketed as a recorded delegation instead. The statement sequence is identical on both sides by construction; the bracket interior stays the scheme's contract. Stated plainly: this does not make the interior provably symmetric, and the invariant "commit absorbs exactly its commitment" is still by inspection.

## What is bound where

| Field | Verdict |
| --- | --- |
| `instances.len()` | fingerprint — one public-value step per instance — and label |
| `instances[i].num_public_values` | fingerprint — `Length::Fixed(n)` in step `i` |
| `instances[i].preprocessed_width` | fingerprint for zero-vs-nonzero, label for the value |
| `instances[i].num_variables` | label |
| `instances[i].main_width` | label |
| `pow_bits` | label |

`preprocessed_width` is the interesting one: the step sequence only sees present or absent, so two nonzero widths share a sequence and the number itself goes in the label.

## Tests

`every_field_of_the_shape_reaches_the_seed` walks all six knobs. `no_two_one_step_mutations_collide` separates them from each other, not just from the baseline. `two_distinct_batches_never_share_a_seed_preimage` is a proptest asserting seed equality holds exactly when `(pattern_hash, instance_label)` agree — an injectivity claim rather than a sampled one. `verify_rejects_tampered_main_commitment` is the mutation test for the newly bracketed absorb.

## Compatibility

This is a breaking transcript change. Proofs produced before it do not verify after it, and `multi_stark_whir_v0_8_0.postcard` is regenerated.

## Known limitation

Whether a lookup runs is not bound at statement level. Deriving it prover-side would mean building `LookupPlan` twice, so it is bound inside the brackets instead — by the lookup transcript's own seed when a lookup exists, and by the zerocheck's `lookup_point_len` and per-AIR interaction counts in either case. Two batches identical in widths, arities and public-value counts but differing in whether an AIR declares interactions therefore share a *statement* seed and diverge at the zerocheck seed. Sound, but weaker than binding it at the top; it wants a cheap `declares_interactions` predicate on the AIR side.


multi-stark/src/transcript.rs
```diff
@@ -0,0 +1,1294 @@
+//! Fiat-Shamir transcript of the batched multi-STARK statement.
```
Nashtare, line 1: Reviewing this as the bottom of the new stack, so the stack-level note lands here once rather than on all eight.

#2102 through #2109 are a single chain — each PR's base is the previous one's head, rooted at #2081:

```
main <- #2081 <- #2102 <- #2103 <- #2104 <- #2105 <- #2106 <- #2107 <- #2108 <- #2109
```

Ten commits ahead of `main`, four behind. **#2081 is `CONFLICTING/DIRTY` against `main`**; every PR above it reports `MERGEABLE` only relative to its own parent branch, so the green state on #2102-#2109 says nothing about whether they merge. Nothing here lands until #2081 does.

The only textual conflict is `multi-stark/tests/fixtures/multi_stark_whir_v0_8_0.postcard`, regenerated on both sides — by #2100 on `main` and by this stack. It has to be regenerated after the rebase, not resolved either way.

The four commits `main` has that the stack lacks are #2100, #2097, #2095, #2098. #2100 is the one to watch: it is `fix(security)!: enforce PCS budgets and compose multi-STARK soundness`, it rewrote multi-STARK's composed security checks, and this PR types multi-STARK's statement-level transcript. The fixture conflict is the visible half of that overlap; worth confirming the composed checks still line up once rebased.

For what it is worth I checked the smaller interaction too: `security/src/grinding.rs` (which #2109 extends) already exists on `main` and #2100 touched only `security/src/lib.rs`, so that particular overlap is not a problem.

tcoratger: Checked all of it by trial rebase rather than by reading, and the answer to the last paragraph is: **the composed checks line up, but one hunk has to be resolved in #2100's favour or the merge silently reverts its soundness fix.** Worth pinning down before the real rebase.

#2100 has since landed (`b4ef483d`), so main is now four ahead as you say. I cherry-picked #2081 and all eight of this stack onto `b4ef483d` in a throwaway tree. Conflicts, in full:

**multi-STARK, 2 hunks, both compositional.** An import line and a doc paragraph, in each of `prover.rs` and `verifier.rs`. Your `prove_with_security` / `verify_with_security` note and this PR's statement-transcript note are both true and simply concatenate. Resolved, and `cargo build -p p3-multi-stark` is clean against #2100.

**WHIR, 5 hunks, and one of them is the real overlap.** In `pcs/prover/mod.rs` and `pcs/verifier/mod.rs`:

```
    #2100   Constraint::new_with_existing_claim(challenger.sample_algebra_element(), ..)
    #2104   Constraint::new(transcript.round_batching(), ..)
```

The two changes are orthogonal and both are needed. #2100 changed the *constructor* — `new_with_existing_claim` sets `initial_power: 1`, reserving alpha^0 for the carried claim so fresh intermediate claims cannot cancel against it. #2104 changed the *challenge source*, from a bare sponge draw to a recorded pattern step. The correct resolution is

```
    Constraint::new_with_existing_claim(transcript.round_batching(), ..)
```

Taking either side wholesale is wrong, and the failure is asymmetric: taking #2104's side compiles, passes every test in this stack, and quietly puts the fresh claims back on alpha^0. That is the hunk to watch, not the fixture.

`pcs/adapter.rs` has the same shape twice — `validate_initial_claims` needs to stay where you put it, ahead of the run, while the batching draw moves inside the first bracket. Both compose; the alpha sample just no longer lives in the adapter.

**The version byte.** `whir/src/transcript/mod.rs` conflicts on `VERSION`, and this is why #2104 took **3** rather than 2: #2100 also bumps it to 2, for the batching-coefficient reservation. Two incompatible descriptions sharing a byte would defeat the one field that exists to separate them, so 3 is correct in either merge order and the constant carries a note saying 2 is taken.

**The fixture** is regenerated, not resolved, exactly as you say — `multi_stark_whir_v0_8_0.postcard`, after the rebase, from `cargo test -p p3-multi-stark --test whir_fibonacci -- --ignored`.

Leaving this thread open rather than resolving it: nothing above can actually land until #2081 rebases onto main and gets approved, and #2081's own resolution may move these hunks. The finding is recorded here so the `new_with_existing_claim` hunk does not get lost when it does.

And agreed on `security/src/grinding.rs` — #2109 extends a file #2100 never touched, so that one is clear.

tcoratger: Done — the rebase has landed, so this is no longer pending. The whole stack sits on `b4ef483d` and is zero commits behind; #2081 went from `CONFLICTING/DIRTY` to `MERGEABLE`, and so did all eight above it.

**The composed checks line up.** `validate_initial_claims` survives on both adapter paths, `assess_statement` / `security_report` and the statement-level transcript coexist in `multi-stark/src/{prover,verifier}.rs` — those conflicts were an import line and a doc paragraph each, and both sides simply concatenate.

**The hunk worth recording resolved as predicted.** In `pcs/prover/mod.rs` and `pcs/verifier/mod.rs` the merge is

```rust
Constraint::new_with_existing_claim(transcript.round_batching(), ..)
```

— your constructor, which reserves the constant coefficient for the carried claim, taking the recorded pattern step as its challenge source. Both halves kept. I added a line at the verifier site saying the carried claim keeps the constant coefficient so a fresh group never shares a power with it, since that is the property a future reader of those two lines would otherwise have to rediscover.

**Version byte:** 3, with the doc now recording both facts — 2 reserves the batching coefficient, 3 keeps that reservation and adds the delegation brackets.

**Fixture:** regenerated once at each commit that moves the transcript, and `verify_whir_compat_fixture` passes at every step of the chain.

Verification on the rebased stack: **3962 tests pass**, `scripts/check.py lint` exits 0. That count includes the tests #2100 brought with it, so the composed accounting is exercised rather than merely compiled.

Resolving this one now that the rebase it was about is real. The remaining gate is approval, not conflicts.

Nashtare (approved):
Approving. Reviewed at `9b7bd532` against `fb50bd6d` (7 files, +1534/−122).

**My #2100 worry was misplaced and I'll retract it.** `b4ef483d` is a direct *ancestor* of this branch — there was no rebase over it, so nothing could have been reverted. `git diff b4ef483d 9b7bd532 -- multi-stark/src/security.rs` is empty, `prove_with_security` / `verify_with_security` are byte-identical, `assess_statement` is still called on the prover path and via `security_report` on the verifier path, and all ten of #2100's tests survive. (For the record the name in this crate is `assess_statement`; `validate_initial_claims` is WHIR-side.)

**Statement-level symmetry holds under every branch I could construct.** Both sides build the shape from the same trusted sources — `MultiStarkShape::new::<C::Val, A>(&airs, &num_variables, pow_bits)` at `prover.rs:149` and `verifier.rs:179` — and the step sequence is call-for-call identical. Preprocessed present/absent gates on `shape.has_preprocessed()` on both sides, derived from the same AIR list, and a key/shape disagreement is now caught in *both* directions (`MissingPreprocessedCommitment` / `UnexpectedPreprocessedCommitment`), where the old verifier only caught one. Lookup present/absent is symmetric because the bracket is unconditional and both callees no-op before touching the challenger. Nothing from `proof` shapes a step: `num_public_values` comes from the AIR, the values from `VerifierInstance`, length-checked before absorption — so a prover cannot pick the step width. That converts what used to be a downstream `assert_eq!` into an early typed error, which is a small improvement.

**The known limitation really is only a limitation.** I constructed the case: two batches identical in arity, widths and public-value counts, differing only in whether an AIR declares interactions. They share the statement seed but diverge before either draws its first challenge — the lookup driver seeds its own separator only in the interaction-bearing batch, and independently the zerocheck label binds `degrees.interactions` per AIR. Since the statement transcript draws no challenges of its own, there is no shared challenge upstream of the divergence. Agreed it still wants a cheap `declares_interactions` predicate, but it is not a hole.

Non-blocking:

1. **The preprocessed opening now always runs, even after the main opening has already failed.** `verifier.rs:245-279` defers both `?`s to reach `finish()`, so a proof with a bad main opening still executes `preprocessed_pcs().verify_at(...)` against attacker-supplied `proof.preprocessed_opening`, with a challenger in a state no honest run produces. The invariant comment at `:234-241` explains the choice but not its cost: verifier work roughly doubles on the rejecting path for any batch with preprocessed columns. `abort()` exists for exactly this and the lookup/zerocheck arms two branches up already use it — skipping the preprocessed step and aborting would restore the old profile at no loss.

2. **The `main_commitment` bracket enforces nothing, and the invariant it stands in for isn't on the trait.** I verified all three in-tree impls do exactly one `observe(commitment)` and sample nothing, so this is correct today. But `MultilinearPcs::commit`'s doc never mentions the `challenger` parameter at all, and the trait carries no challenger bound — not even `CanObserve<Self::Commitment>`. An out-of-tree impl that absorbs table heights or draws a batching challenge would compile, pass the pattern check on both sides, and desync as an unexplained completeness failure. Since declining to hoist the absorb rests entirely on this property, it belongs on the trait doc, and a `CanObserve` bound would at least make the shape type-visible.

3. `verifier.rs:100` says "a step either side skips or reorders is rejected" — the player is a purely local self-check with no visibility into the other party. What rejects a prover that skips a statement step is absorb-divergence, and what rejects a skipped bracket is the sub-protocol's own verification. Worth tightening, since this is the same reasoning that justifies (2).

4. Test-strength nits: `one_step_from_base` is hand-maintained, so a new `MultiStarkShape` field forces a value in `base_shape()`/`arbitrary_shape()` but gets zero coverage in the walk — the "a knob added without a binding shows up here" claim is aspirational. `arbitrary_shape()` never emits a permutation of an already-drawn shape, so the label's instance-order sensitivity is untested (it is structurally sound, just unasserted). And `verify_rejects_tampered_main_commitment` pins an exact WHIR error string.

114 lib + 38 integration tests pass, including the regenerated fixture under real `verify()`.

PR #2103, opened 2026-09-09
feat(sumcheck)!: drive the hiding sumcheck transcript through the typed Fiat-Shamir layer

tcoratger:
Depends on #2102, which depends on #2081. This PR targets `feat/typed-transcript-multi-stark-statement` so the diff contains only this crate's migration; retarget after the branches below it land.

## What this changes

#2081 migrated the quadratic sumcheck. The hiding sumcheck — the HVZK masking layer that WHIR's zero-knowledge pipeline runs on — was still writing its transcript by hand, on five separate paths, and it was the last raw-challenger transcript inside `p3-sumcheck`.

There are five sequences, not one: two prover entry points, two verifier mirrors, and a simulator that has to match the prover byte for byte or the zero-knowledge claim is void.

`sumcheck::zk::transcript` now holds one description all five play, and `sumcheck/src/zk` contains zero raw challenger calls.

## The described sequence

```text
    claim_batching   alpha, a Challenge          (the batching entry points)
        or
    joint_claim      a Message                   (the inherited-claim entry points)

    mask_commitment  opaque
    mu_tilde         one extension element
    mask_combination eps, a Challenge

    per round:  round_poly  Fixed(max(ell_zk,3) - 1)
                round_pow   Fixed(pow_bits), iff pow_bits > 0
                round_challenge
```

The two preludes are a `Challenge` and a `Message` under different labels, so neither run can replay as the other and their fingerprints differ. Which one applies is fixed by the constructor the entry point picks, so a caller cannot mismatch it.

## Why the simulator is the point

`simulate_classic_unpacked` re-implemented `sample_masks` and `observe_masks_and_mu_tilde` inline instead of calling them. The simulator's transcript has to be indistinguishable from the real prover's or HVZK does not hold — and that held only because two copies of the sequence happened to agree.

Both helpers are now pure, the simulator drives the same `ZkSumcheckShape` through the same driver as the prover, and `the_simulator_and_the_prover_play_one_description` pins it: each party's recorded values are pushed through a bare driver built from the shape alone and must reproduce that party's own challenge stream. The test deliberately asserts the *wires* are not equal — Lemma 6.4 claims indistinguishability of distributions, not equality. What is shared is the description.

## Lengths that came off the wire

Four things the verifier used to read from the proof now come from the pattern:

- the round loop iterated `zk_data.round_coefficients`, so the number of transcript steps was proof-derived;
- the absorbed element count was `wire.len()`;
- `pow_witnesses[j_idx]` indexed a proof vector;
- `ZkSumcheckData::ell_zk` was **a length carried inside the proof**, never absorbed, only compared against config.

`ell_zk` is deleted. Its own doc comment explained why it needed an explicit check — wire lengths 2 and 3 share a layout, so the encoding is not injective. That non-injectivity is exactly what a fingerprint is for, and `ell_zk` now reaches the seed through both the pattern and the instance label.

## Panics removed

`replay_claim_unchecked` loses its `_unchecked`: the loop bound is config, and `wire[0]` / `wire[1..]` are read only after the driver has accepted `Length::Fixed(wire_len)`. The three `assert!`s that sat inside a `Result`-returning verifier moved into `ZkSumcheckShape::validate`, returning typed errors.

## What is bound where

| Field | Verdict |
| --- | --- |
| `prelude` | fingerprint — changes the first step's kind and label |
| `num_rounds` | fingerprint — and the only thing binding the mask oracle's column count, since an opaque step binds content, not width |
| `pow_bits` | fingerprint — step presence and its declared difficulty |
| `ell_zk` | fingerprint *and* label — `Length::Fixed(max(ell_zk,3) - 1)` is not injective |

The mask encoding's own parameters — `mask_log_inv_rate`, `randomness_len`, `domain_size`, `mask_queries`, `oracle_randomness` — are absent by design: the sumcheck verifier never receives an encoding, so it cannot know them. `ZkWhirShape`'s instance label already binds all of them.

## Compatibility

Breaking transcript change, and `ZkSumcheckData` loses a field. No fixture needed regeneration: every `.postcard` in the tree exercises plain WHIR, Circle or two-adic FRI, and `HidingWhirPcs` has no consumer outside a bench.

## Known limitation

The inherited-claim agreement is still a caller obligation. The pattern records the step and both sides bind a scalar, so a disagreement is loud — it moves `eps` and every subsequent challenge. But nothing inside `p3-sumcheck` can compare the prover's `claimed_sum() + aux_claim` against the verifier's target; that would need the auxiliary total to be a described value, which is a WHIR-side change. Documented in `verify_claim`'s `# Soundness`.

### DIFF
--- a/sumcheck/src/zk/transcript.rs
+++ b/sumcheck/src/zk/transcript.rs
@@ -1,0 +1,1085 @@
+//! Fiat-Shamir transcript of the hiding sumcheck.
+//!
+//! One description of everything a masked batch absorbs.
+//!
+//! Prover, verifier and simulator all drive that same description.
+//!
+//! # Shape
+//!
+//! ```text
+//!     prelude      ->  1 extension element, drawn or bound
+//!     mask oracle  ->  1 commitment
+//!     mu_tilde     ->  1 extension element
+//!     eps          ->  1 extension element, drawn
+//!
+//!     round i:  wire       ->  max(ell_zk, 3) - 1 extension elements
+//!               grinding   ->  only when the difficulty is positive
+//!               challenge  ->  1 extension element
+//! ```
+//!
+//! A round polynomial has `max(ell_zk, 3)` coefficients, and the linear one stays off the wire.
+//!
+//! The affine round identity reconstructs it.
+//!
+//! # Two preludes, one description
+//!
+//! ```text
+//!     recorded claims  ->  one challenge weights the claims a layout already recorded
+//!     inherited claim  ->  one scalar arrives from the caller, bound before anything is drawn
+//! ```
+//!
+//! Everything after the prelude is identical, so one description covers both.
+//!
+//! They are distinct steps, so a side that binds where the other draws fails loudly.
+//!
+//! # Sub-protocol
+//!
+//! A masked sumcheck is always a phase of something larger.
+//!
+//! It seeds a transcript of its own from a borrowed sponge, and hands the sponge back.
+//!
+//! The surrounding protocol brackets this run inside its own description.
+//!
+//! That is what lets the two descriptions be written independently and still compose.
+//!
+//! # What the shape binds
+//!
+//! A fingerprint of the description enters the sponge before any step runs.
+//!
+//! ```text
+//!     num_rounds  ->  round count, and so the mask oracle's column count
+//!     pow_bits    ->  presence and difficulty of the per-round grinding step
+//!     ell_zk      ->  wire width, and the instance label on top of it
+//!     prelude     ->  which of the two opening steps is played
+//! ```
+//!
+//! The mask oracle is one opaque step, binding its content and not its width.
+//!
+//! One column is committed per round, so the round count bounds that width.
+//!
+//! ```text
+//!     rate, randomness budget, domain size  ->  committed content only, never seen here
+//! ```
+//!
+//! The protocol one layer up carries all three in its own instance label.
+//!
+//! # Simulator
+//!
+//! The witness-free simulator of Lemma 6.4 plays this same description.
+//!
+//! Hiding holds only while its transcript is indistinguishable from the prover's.
+//!
+//! A divergence between the two is a loud pattern failure here, not a silent loss of hiding.
+
+use alloc::vec::Vec;
+use core::marker::PhantomData;
+
+use p3_challenger::fs::{
+    DomainSeparator, FieldToFieldCodec, FieldUnit, Hierarchy, Interaction, InteractionPattern,
+    Kind, Length, ProverState, TranscriptField, VerifierState,
+};
+use p3_challenger::{CanObserve, CanSample, GrindingChallenger};
+use p3_field::{ExtensionField, Field};
+
+use crate::error::SumcheckError;
+
+/// Version byte bound into the transcript seed.
+///
+/// Bumping it separates two revisions of this protocol.
+/// It does so even when their step sequences agree.
+const VERSION: u8 = 1;
+
+/// Protocol name bound into the transcript seed.
+///
+/// Distinct from the plain sumcheck names, so no masked batch shares a seed with a plain one.
+const NAME: &[u8] = b"p3-sumcheck-hvzk";
+
+/// Step label of the challenge that weights the claims a layout recorded.
+const CLAIM_BATCHING: &str = "claim_batching";
+
+/// Step label of the scalar claim a batch inherits from its caller.
+const JOINT_CLAIM: &str = "joint_claim";
+
+/// Step label of the interleaved mask oracle of a batch.
+const MASK_COMMITMENT: &str = "mask_commitment";
+
+/// Step label of `mu_tilde`, the sum of the batch's mask evaluations over the cube.
+const MU_TILDE: &str = "mu_tilde";
+
+/// Step label of `eps`, the challenge combining the mask and plain pieces.
+const MASK_COMBINATION: &str = "mask_combination";
+
+/// Step label of the wire coefficients one round sends.
+const ROUND_POLY: &str = "round_poly";
+
+/// Step label of a per-round grinding step.
+const ROUND_POW: &str = "round_pow";
+
+/// Step label of a per-round challenge.
+const ROUND_CHALLENGE: &str = "round_challenge";
+
+/// Coefficient count of the plain round polynomial.
+///
+/// The plain piece is quadratic, so it occupies three coefficient slots.
+/// A round polynomial is therefore never shorter than this, whatever the mask length is.
+const PLAIN_COEFFICIENTS: usize = 3;
+
+/// Sponge alphabet of a challenger that speaks the base field natively.
+type Alphabet<F> = FieldUnit<F>;
+
+/// How a batch reaches the claim it runs against.
+///
+/// The two variants are two different opening steps, never two readings of one step.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ZkPrelude {
+    /// One drawn challenge weights the claims a layout already recorded.
+    ///
+    /// The claim is then derived on both sides from those recorded claims.
+    RecordedClaims,
+    /// One scalar claim arrives from the caller and is bound before anything is drawn.
+    ///
+    /// The two sides supply it independently, so the step is what keeps them honest.
+    InheritedClaim,
+}
+
+/// Numbers that fix the transcript of one masked batch of sumcheck rounds.
+///
+/// Both sides build this from their own configuration, never from a proof.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ZkSumcheckShape {
+    /// Which of the two opening steps this batch plays.
+    pub prelude: ZkPrelude,
+    /// Number of rounds the batch runs, and so the number of masks it commits.
+    ///
+    /// The mask oracle interleaves one column per round.
+    /// This count is therefore what binds that oracle's width.
+    pub num_rounds: usize,
+    /// Message length of the zero-knowledge mask code.
+    ///
+    /// Sets the wire width through `max(ell_zk, 3) - 1`.
+    pub ell_zk: usize,
+    /// Grinding difficulty guarding each round's challenge, or zero to omit grinding.
+    pub pow_bits: usize,
+}
+
+impl ZkSumcheckShape {
+    /// Collect the numbers that fix a batch opening on the claims a layout recorded.
+    ///
+    /// # Arguments
+    ///
+    /// - `num_rounds`: number of rounds the batch runs.
+    /// - `ell_zk`: message length of the mask code.
+    /// - `pow_bits`: grinding difficulty per round, or zero to omit grinding.
+    #[must_use]
+    pub const fn new_batching(num_rounds: usize, ell_zk: usize, pow_bits: usize) -> Self {
+        Self {
+            prelude: ZkPrelude::RecordedClaims,
+            num_rounds,
+            ell_zk,
+            pow_bits,
+        }
+    }
+
+    /// Collect the numbers that fix a batch opening on a claim it inherits.
+    ///
+    /// # Arguments
+    ///
+    /// - `num_rounds`: number of rounds the batch runs.
+    /// - `ell_zk`: message length of the mask code.
+    /// - `pow_bits`: grinding difficulty per round, or zero to omit grinding.
+    #[must_use]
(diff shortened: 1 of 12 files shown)


sumcheck/src/zk/transcript.rs
```diff
@@ -0,0 +1,1095 @@
+//! Fiat-Shamir transcript of the hiding sumcheck.
```
Nashtare, line 1: Flagging scope honestly: I gave this one a structural pass, not a line-by-line review, and it is the PR in the stack I would most want a real pass on.

What I did check: the version bump is rebase-aware, the fixtures are regenerated, and the net count of transcript binding call sites in non-test code across the three refactor PRs is

```
#2102  +25 / -5
#2103  +17 / -27
#2104  +31 / -12
```

#2103 is the only negative delta, and it is the *hiding* sumcheck. That combination is what makes it worth a dedicated pass rather than a skim: in the plain protocols a dropped absorption desynchronises the replay and a regenerated fixture catches it, but here a value that stops being absorbed can instead be a zero-knowledge or soundness bug that every existing test still passes.

A call-site count cannot tell "moved into this new 1,095-line module" from "dropped" — the method names change in the move, so the regex is measuring the wrong thing. I did not resolve which it is. If you have a mapping from the old `observe`/`challenge` sites to their typed replacements, that would settle it quickly; otherwise it wants someone walking the simulator and residual-prover paths against the pre-refactor ones.

tcoratger: Here is the mapping. Short version: **the `-27` is fan-in, not attrition — zero absorptions were dropped on any of the five paths.** Your instinct to distrust the regex was right; it was counting five hand-written copies collapsing into two.

## The five paths

The hiding sumcheck has five sequences, not one, and they have to be walked separately:

```
    A1  prover     ZkProver::into_sumcheck                (layout prover)
    A2  prover     SumcheckProver::into_zk_sumcheck       <- WHIR's live path
    A3  verifier   ZkVerifier::into_sumcheck
    A4  verifier   ZkVerifier::verify_claim               <- WHIR's live path
    A5  simulator  simulate_classic_unpacked              <- the zero-knowledge claim
```

Pre-refactor each was a hand-written copy of one 6-to-8-operation sequence:

```
    A1  alpha · commit · mu_tilde · eps · wire · grind · gamma
    A2  claim · commit · mu_tilde · eps · wire · grind · gamma
    A3  alpha · commit · mu_tilde · eps · wire · check · gamma
    A4  claim · commit · mu_tilde · eps · wire · check · gamma
    A5  alpha · commit · mu_tilde · eps · wire · grind · gamma
```

Post-refactor there are two copies — one prover-side, one verifier-side — of that same sequence. Five copies to two copies is where 27 goes to 18.

## Per-value mapping

Every pre-refactor operation, classified. There is no DROPPED row in any of the five tables.

| pre-refactor | typed replacement | pattern step |
| --- | --- | --- |
| sample `alpha` | `batching_challenge()` | `Challenge "claim_batching"` Scalar |
| observe joint claim (A2/A4) | `bind_claim(..)` | `Message "joint_claim"` Scalar |
| observe mask commitment | `masks(..)` -> `observe_opaque` | `Message "mask_commitment"` Scalar |
| observe `mu_tilde` | `masks(..)` -> `observe_extension` | `Message "mu_tilde"` Scalar |
| sample `eps` | `masks(..)` | `Challenge "mask_combination"` Scalar |
| observe `wire` | `round(..)` -> `observe_extensions` | `Message "round_poly"` Fixed(wire_len) |
| `grind` / `check_witness` | `round(..)` -> `observe_pow` | `Pow "round_pow"` Fixed(pow_bits) |
| sample `gamma_j` | `round(..)` | `Challenge "round_challenge"` Scalar |

The typed calls are sponge-equivalent to the raw ones, not merely similar: the extension codec absorbs the same basis coefficients in the same order, and `Length::Fixed` is a *pattern* assertion rather than an absorbed prefix. (`observe_extensions_bounded` does absorb a big-endian count — it is not used here.)

**One operation is ADDED, the same one on all five paths:** the domain-separator seed. That is why this is a breaking transcript change; it is the only reason.

## Dynamic accounting, which is what a dropped absorb would show up in

| path | before | after | delta |
| --- | --- | --- | --- |
| A1 / A2 / A3 / A4 / A5 | `4 + k*(2 or 3)` | `5 + k*(2 or 3)` | **+1** (the seed) |

## The simulator

This was the right thing to be nervous about, and it is the part that got strictly better.

Pre-refactor the simulator re-implemented the mask sampling and the `mu_tilde` closed form **inline**, as a third copy. Its transcript matched the prover's because two hand-written copies happened to agree — and a divergence there is a hiding break, not a completeness bug, so no test would have caught it.

It now calls the prover's own `sample_masks` and `mask_endpoints` (neither of which touches a sponge any more — that module lost its challenger import entirely) and drives **the same `ZkProverTranscript` type over the same shape constructor**. A step played out of order by either party is a pattern-player failure.

```
    A1 (prover)                    A5 (simulator)
    new_batching(k, ell, pow)  ==  new_batching(k, ell, pow)
    batching_challenge()       ==  batching_challenge()
    masks(commit, mu_tilde)    ==  masks(commit, mu_tilde)
    k x round(wire)            ==  k x round(wire)
```

Pinned by `the_simulator_and_the_prover_play_one_description`: each party's challenge stream is reproduced by a bare driver over the shared shape, fed only that party's bound values.

## The deleted `ell_zk` field

Worth stating plainly because it looks like a removed binding and is the opposite.

It was **never absorbed** — grep the parent tree, there is no absorb of it. It was a proof-carried self-report that was only ever compared against config, and its doc comment's claim that it "closes a non-injectivity gap" was inaccurate: it was pinned in the proof record, never in the sponge.

The check it fed was also unreachable. Its stated purpose was that `max(ell_zk, 3)` is not injective — lengths 2 and 3 share a wire layout — but the parent already rejected `ell_zk < 3` unconditionally, and for `ell_zk >= 3` the map to `wire_len` is injective, so the per-round width check already caught every reachable disagreement.

It is now bound three ways instead: verbatim in the instance label, through `Length::Fixed(wire_len)` in the pattern hash, and (on the production path) already in WHIR's own instance label one layer up. `every_number_that_shapes_a_batch_reaches_the_seed` pins the case the old field was supposed to cover: `wire_len` collides for `ell_zk` in {2,3} while the seed does not.

## Two prelude variants are now domain-separated

A gain that did not exist before: the batching prelude emits `Challenge "claim_batching"`, the inherited prelude emits `Message "joint_claim"`. Different pattern hash, different seed, so **neither run can be replayed as the other**. Pre-refactor the two differed only in whether one extension element was sampled or observed, with no separation at all.

## Verified by running it, not just reading

`cargo test -p p3-sumcheck --lib zk::` -> **50 passed, 0 failed**, including the simulator/prover description test above. Also `cargo test -p p3-whir` -> 139 passed.

## Things I am flagging rather than claiming clean

- **Malformed-wire rejection moved** from up-front to at-the-round. A bad wire in round 3 now has rounds 0-2 absorbed before it rejects. Same verdict, sponge discarded either way, no soundness consequence — but `WireSizeMismatch.round` went from 1-based to 0-based, which now matches the pow-witness variant (they disagreed before).
- **The joint-claim agreement between the two sides is still a caller obligation.** The prover binds `claimed_sum() + aux_claim`, the verifier binds whatever its caller supplies; nothing inside this crate compares them. Unchanged by this PR, now documented, and a disagreement is loud rather than silent. The WHIR call sites are byte-identical across the commit, so it cannot have been broken here — but whether `masks.aux` equals `*target` for every WHIR round is a separate audit of the mask bookkeeping.
- **`folding_factor <= num_variables`** stays a prover-side `assert!` with no verifier counterpart, same as pre-refactor. Unchanged, but it is the one precondition still asymmetric between the two sides.
- **Two types now share the name `ZkSumcheckShape`** — the new one here, which is played, and WHIR's pre-existing one, which is fingerprint-only. WHIR still describes the masked-sumcheck steps inline, which is now redundant with the nested description. Harmless today, since redundancy only adds binding, but if this crate's step sequence changes WHIR's fingerprint will not notice, and the shared name makes that easy to miss. Worth a follow-up to derive it or replace it with a sub-protocol bracket.

Resolving on the basis that the mapping was the thing that would settle it — please reopen if you would rather walk it yourself with this in hand.

Nashtare (approved):
Approving. This is the one I said needed a dedicated pass rather than a skim, so it got one: full A1–A5 enumeration at `57f682a2` against `9b7bd532`.

**The −27 is fan-in, confirmed independently rather than taken from the mapping.** Counting non-test sponge sites mechanically (stopping at `#[cfg(test)]`): parent has 27 — `common.rs` 2, `residual.rs` 5, `zk_prover.rs` 5, `simulator.rs` 7, `verifier.rs` 8 — and the child has **zero**. Every one of the 27 maps to a typed counterpart, none orphaned. The arithmetic works out because those 27 physical sites already represented 5 × 7 logical steps with 8 shared (`common.rs:80/155` across A1+A2, `verifier.rs:135/136/137/147/149/156` across A3+A4); two drivers at 8 ops each plus 2 seedings is 18. The net is *more* binding, not less — each run now additionally absorbs a separator carrying the pattern hash and `ell_zk`. That settles the question I left open.

**Deleting `ell_zk` is correct, and every clause of the argument checks out.** It was never absorbed in the parent (only `zk_data.ell_zk = ell_zk` and one comparison); the `ell_zk < 3` rejection was unconditional inside `validate_shape` on both verifier entry points; `max(ell_zk,3)-1` reduces to `ell_zk - 1` and is injective above 3; and the replacement bindings exist in both the instance label and `Length::Fixed(wire_len)`. The ell_zk 2-vs-3 collision case — same `wire_len`, different seed — is pinned by a test.

**The simulator is the part that got genuinely better.** `common.rs` no longer imports `p3_challenger` at all, so `sample_masks` / `mask_endpoints` really are pure; A5 plays the A1 order through the same `ZkProverTranscript`; RNG draw order is preserved. `the_simulator_and_the_prover_play_one_description` is not vacuous despite asserting the wires differ — it re-derives each party's challenge stream from a bare driver over the shared shape fed only that party's bound values, so a step played out of order lands on a different stream.

Also verified: prover/verifier symmetry on both prelude variants (the PoW branch reads one `shape.pow_bits`, so the two cannot disagree); `check_witness(0, w)` returning true without absorbing is unreachable here because `observe_pow` is never called at zero difficulty *and* `validate_shape` requires an empty witness vector — no malleability introduced; all three error paths release the completeness check; dropping `_unchecked` from `replay_claim` is justified since `wire[0]`/`wire[1..]` are read strictly after `Length::Fixed(wire_len)` is accepted.

Non-blocking:

1. **There is no witness-free simulator for the path WHIR actually runs.** `simulate_classic_unpacked` hardcodes `new_batching` (`simulator.rs:95`), but the deployed masked sumcheck is the inherited prelude (`residual.rs:88` ↔ `verifier.rs:354`). This is pre-existing and not worsened here, but `transcript.rs:5` ("Prover, verifier and simulator all drive that same description") and `:68` ("The witness-free simulator of Lemma 6.4 plays this same description") now assert it. `simulator.rs:38` is correctly scoped — please narrow the two lines in `transcript.rs`, and an inherited-prelude simulator is worth filing.

2. **`transcript.rs:41` says the surrounding protocol brackets this run inside its own description.** It does not — there are no `Begin`/`End` markers for the delegation, because `ZkWhirShape` is only ever seeded, never played. Intent, not code as it stands. Related: the duplicated `wire_len = max(ell_zk,3)-1` in `whir/src/transcript/zk.rs:442` genuinely cannot desync the two sides (both seed the identical shape), so "additive only" is right — but drift would leave WHIR's fingerprint describing steps that no longer occur.

3. `verifier.rs:292/:366` leave the driver to be dropped without `finish()` on the `?`. Correct today because all three underlying paths abort, but the invariant is a cross-crate contract; an `.inspect_err(|_| transcript.abort())` would make it local and stop a future fallible step from turning a rejection into a drop-time panic.

4. The "lengths that came off the wire" section overstates: three of the four were already config-pinned before any absorption in the parent (`round_coefficients.len()`, per-round `wire.len()`, and the `pow_witnesses` length that made the index in-bounds). Only `ell_zk` genuinely moved. Zero security delta on the other three — refactor, not hardening. Also `WireSizeMismatch.round` going 1-based → 0-based belongs in the `!` changelog even though no in-repo consumer matches on it.

`cargo test -p p3-sumcheck` 322+6 pass, `-p p3-whir` 146 pass including `zk_whir_end_to_end_multi_round`, clippy clean on both.

PR #2104, opened 2026-09-09
feat(whir)!: drive the plain WHIR transcript through the pattern player

tcoratger:
Depends on #2103, #2102, #2081. This PR targets `feat/typed-transcript-sumcheck-hiding` so the diff contains only this crate's migration; retarget after the branches below it land.

## The finding

#2089 is titled *"seed the WHIR transcript"*, not *"drive"* it. That distinction turned out to be the whole story.

```
grep -rn 'ProverState\|VerifierState\|PatternPlayer' whir/src/
```

returned nothing. `WhirShape::pattern` had exactly one caller — `domain_separator()` — and `DomainSeparator::seed` absorbs a pattern only as its 32-byte `pattern_hash()`. So WHIR had a faithful *description* of its transcript and no *enforcement* of it. Prover/verifier drift stayed silent, which is the one thing the typed layer exists to prevent.

To show the silence rather than just the absence, I ran a throwaway test against what #2089 left: seed the domain separator, then draw a challenge where a commitment belonged.

```
seed-only path drew ExtField { value: [1370351934, ...] } where a commitment belonged, and raised nothing
test seed_only_transcript_accepts_any_step_order ... ok
```

This PR makes that a panic.

## What this changes

`WhirProverTranscript` and `WhirVerifierTranscript` wrap real drivers over `Alphabet<F>`. Every one of the twelve described steps is played through them, and each of the three sumcheck phases becomes a `Begin`/`End Protocol` bracket around a child that seeds and enforces its own pattern — the composition idiom `fri::pcs_transcript` already uses, and the one #2081's description anticipated.

The enforcement tests are the deliverable, because they are impossible to write today:

```
panicked at challenger/src/fs/pattern/player.rs:128:
Received interaction Atomic Challenge round_batching Scalar ExtField<..>,
  but expected Atomic Message commitment Scalar opaque

panicked at challenger/src/fs/pattern/player.rs:128:
Received interaction Begin Protocol final_fold None Sumcheck,
  but expected Atomic Challenge final_query_indices Fixed(3) usize
```

Plus `both_sides_draw_the_same_challenges_from_the_same_values`, which walks a full prover/verifier pair and asserts equal challenges, witnesses, indices *and* equal next sponge output; and a proptest over random carried values.

## The pre-seed region

The initial commitment and the public claims are absorbed before the seed lands, deliberately. But the **number of opening claims** reached no fingerprint and no instance label. It does now, alongside the committed row width. The digest width is still unbound and this PR does not pretend otherwise: a commitment is absorbed opaquely through the challenger's own encoding, the layer cannot see its width, and the Merkle check against explicit `Dimensions` is what rejects a wrong one. `fri` and `circle` state the same limit; the module header now says it too.

## Version 3, not 2

`VERSION` goes from 1 to **3**, skipping 2 on purpose. #2100 also bumps this constant to 2, for an unrelated reason. Two different WHIR transcripts both labelled version 2 would defeat the version byte at precisely the job it exists to do, so this takes 3 — which stays distinct from #2100 in either merge order.

## One bound removed from `p3-challenger`

The blanket `impl CanSampleUniformBits<F> for &mut C` demanded `F: PrimeField64` and used nothing from it. A driver holds `&mut C`, so that bound alone would have forced `PrimeField64` back onto WHIR. It is gone — strictly widening, and WHIR is its consumer in the same commit.

## One check became structural

The final-polynomial length check is deleted: the step is `Length::Fixed(final_poly_len)`, so the driver rejects a wrong count before absorbing anything, and maps it back to the same error. Every other length check stays, because each guards an *index* rather than a *draw*.

## Compatibility

Breaking transcript change; `multi_stark_whir_v0_8_0.postcard` is regenerated. `WhirVerifier::verify` now takes the opening-claim count and builds its initial constraint from the batching challenge it draws itself, so the two sides are symmetric across that step.

## Known limitations

- **The initial batching challenge is inside the bracket, not an atomic checked step.** `Layout::into_sumcheck` draws it itself. Hoisting it means an explicit-alpha entry point on that trait, rippling into `sumcheck`, `binary-pcs` and their tests. Its position is enforced; its type and kind are not.
- **HVZK is untouched.** `whir/src/transcript/zk.rs` keeps `VERSION = 1` and the hiding pipeline still speaks to the raw challenger. That is the follow-up.
- Per-round sumcheck counts moved from the fingerprint to the instance label. Equivalent binding strength, but a reviewer diffing fingerprints should know the numbers moved.


whir/src/transcript/mod.rs
```diff
@@ -48,25 +82,41 @@
...
 
 /// Version byte bound into the transcript seed.
-const VERSION: u8 = 1;
+///
+/// The byte parts the seeds of two incompatible descriptions of one run.
+/// Version 3 records each sumcheck phase as one delegation bracket.
+///
+/// Version 2 is claimed by a different description of this protocol.
+/// Two incompatible descriptions must never share a byte.
+const VERSION: u8 = 3;
```
Nashtare, line 107: The bump is right, and it is right in a way that will look wrong during the rebase, so worth recording.

`main` currently has `const VERSION: u8 = 2;` here — #2100 bumped it. This stack branched before that, so within the stack the line reads `1 -> 3`. Jumping over 2 is correct: it keeps the version ahead of what `main` already shipped rather than colliding with it.

The consequence is that the rebase produces a conflict on exactly this line, `1 -> 2` against `1 -> 3`, and the resolution must keep **3**. Taking `main`'s side here would silently reuse a version byte for a different wire format, which is the one outcome the byte exists to prevent. Given the fixture in this stack also has to be regenerated after the rebase, both halves of that need to happen together.

Nashtare (approved):

whir/src/transcript/mod.rs
tcoratger: Called it exactly, and it played out as you described. The rebase has now happened, so this is settled rather than pending.

The conflict came up on this line as `1 -> 2` against `1 -> 3`, and the resolution kept **3**. The constant now records both facts, so the next reader does not have to reconstruct why 2 is skipped:

```rust
/// The byte parts the seeds of two incompatible descriptions of one run.
///
/// Version 2 reserves the constant batching coefficient for the carried claim.
///
/// Version 3 keeps that reservation and records each sumcheck phase as one bracket.
const VERSION: u8 = 3;
```

"Keeps that reservation" is load-bearing rather than decorative, because the other half of this rebase turned out to be the part worth watching. The same two files conflicted on the constraint constructor:

```
    main    Constraint::new_with_existing_claim(challenger.sample_algebra_element(), ..)
    here    Constraint::new(transcript.round_batching(), ..)
```

Those are orthogonal changes and both are needed — one reserves the constant coefficient for the carried claim, the other moves the challenge to a recorded step. Resolved as `new_with_existing_claim(transcript.round_batching(), ..)`. Worth flagging because the failure is asymmetric: taking this branch's side wholesale compiles, passes every test in the stack, and quietly puts fresh claims back on the same power as the carried claim.

And yes, both halves happened together as you said they had to — the fixture was regenerated after the rebase and `verify_whir_compat_fixture` passes. Full stack: 3962 tests, lint clean, zero commits behind main.

PR #2105, opened 2026-09-09
fix(binary-pcs): reject a non-canonical grinding witness at zero difficulty

tcoratger:
Depends on #2104, #2103, #2102, #2081. This PR targets `feat/typed-transcript-whir-driver`; retarget after the branches below it land.

## The bug

`binary-pcs` accepts proofs that are not uniquely encoded. At zero grinding difficulty the proof-of-work witness is completely unbound.

```text
    check_witness(0, w)  ->  returns true, absorbs nothing     (grinding_challenger.rs)
    grind(0)             ->  returns ZERO, absorbs nothing     (binary-field challenger)
    proof.pow_witness    ->  carried unconditionally, written by the prover
```

So at `pow_bits == 0` the field is written, transmitted, and never compared against anything. A third party who intercepts a valid proof can rewrite it and the proof still verifies.

`pow_bits` is a public field on `BinaryPcsParams` with no lower bound, and `tests/end_to_end.rs` already exercises a zero-budget round trip — so the configuration is reachable, not hypothetical.

## Severity

**Proof malleability, not a soundness break.** Nobody can prove a false statement. What breaks is unique encoding: two distinct byte strings verify against the same statement. That matters wherever proof bytes are hashed, deduplicated, used as an identifier, or recursed over.

## The prior art this matches

The crate already treats this exact class as a defect worth a typed error, one field over: `pcs.rs` rejects a spurious `proof.sumcheck.pow_witnesses` with `NonEmptyPowWitnesses`, and that variant's doc comment states the reasoning verbatim. There was simply no analogous check for `proof.pow_witness`. `p3-sumcheck` does the same for all three of its witness vectors.

## The test that proves it

Before the fix, mutating the witness and asserting rejection fails on an `Ok`:

```
thread '...' panicked at binary-pcs/tests/end_to_end.rs:387:
called `Result::unwrap_err()` on an `Ok` value: ()
```

After, it is `NonCanonicalPowWitness { actual }`. The rejection runs before any transcript operation, in the same block as the existing shape checks.

### DIFF
--- a/binary-pcs/src/verifier.rs
+++ b/binary-pcs/src/verifier.rs
@@ -13,7 +13,7 @@ use alloc::vec::Vec;
 use p3_binary_field::BinaryField128;
 use p3_challenger::{CanSampleUniformBits, FieldChallenger, GrindingChallenger};
 use p3_commit::Mmcs;
-use p3_field::Field;
+use p3_field::{Field, PrimeCharacteristicRing};
 use p3_matrix::Dimensions;
 use p3_util::log2_strict_usize;
 
@@ -188,6 +188,37 @@ where
     Ok(())
 }
 
+/// Checks that a zero-difficulty proof carries the one grinding witness a zero budget admits,
+/// before any transcript operation.
+///
+/// A positive budget needs no check here: the witness is absorbed and its sampled bits are
+/// compared, so the difficulty itself pins the field.
+//
+// Why: `GrindingChallenger::check_witness` returns `true` at `bits == 0` without absorbing,
+// which leaves `proof.pow_witness` compared against nothing and free to be any value.
+//
+//     pow_bits = 0 -> prover emits zero, verifier reads nothing -> pin the field here
+//     pow_bits > 0 -> prover grinds,     verifier resamples     -> the grind pins it
+//
+// Assumption: the honest prover's grind at zero bits is pinned to the zero witness.
+// A grinding path that leaves the zero-bit witness unconstrained needs this check revisited.
+// The pin is asserted by `zero_difficulty_grinding_is_pinned_to_the_zero_witness` below.
+pub(crate) fn check_canonical_pow_witness<MT>(
+    config: &BinaryPcsConfig,
+    proof: &BinaryPcsProof<MT>,
+) -> Result<(), BinaryPcsError<MT::Error>>
+where
+    MT: Mmcs<BinaryField128>,
+{
+    if config.pow_bits() == 0 && proof.pow_witness != BinaryField128::ZERO {
+        return Err(BinaryPcsError::NonCanonicalPowWitness {
+            actual: proof.pow_witness,
+        });
+    }
+
+    Ok(())
+}
+
 /// Verifies the query phase of an opening proof: the single grind, the sampled query
 /// indices, every round's Merkle multiproof, and the fold-consistency chain tying each round
 /// to the next.
@@ -223,6 +254,7 @@ where
     // Structural checks: every one derivable from `config` and the proof's own declared
     // lengths, none needing the transcript.
     check_round_and_final_lengths(config, proof)?;
+    check_canonical_pow_witness(config, proof)?;
 
     let domain_size = config.domain_size();
     let target_queries = config
@@ -322,7 +354,7 @@ mod tests {
     use p3_binary_field::BinaryField128;
     use p3_challenger::{CanObserve, FieldChallenger, GrindingChallenger};
     use p3_commit::Mmcs;
-    use p3_field::PrimeCharacteristicRing;
+    use p3_field::{Field, PrimeCharacteristicRing};
     use p3_matrix::dense::RowMajorMatrix;
     use p3_multilinear_util::poly::Poly;
     use p3_sumcheck::layout::{Layout, SuffixProver, Table};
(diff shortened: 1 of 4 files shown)


binary-pcs/src/verifier.rs
```diff
@@ -186,6 +186,33 @@ where
...
+///
+/// A positive budget needs no check here: the witness is absorbed and its sampled bits are
+/// compared, so the difficulty itself pins the field.
+//
+// Why: `GrindingChallenger::check_witness` returns `true` at `bits == 0` without absorbing,
+// which leaves `proof.pow_witness` compared against nothing and free to be any value.
+//
+//     pow_bits = 0 -> prover emits zero, verifier reads nothing -> pin the field here
+//     pow_bits > 0 -> prover grinds,     verifier resamples     -> the grind pins it
+pub(crate) fn check_canonical_pow_witness<MT>(
```
Nashtare, line 206: Verified the premise rather than taking the comment's word for it, and it holds on every path:

- `GrindingChallenger::check_witness(0, w)` returns `true` **before** reaching `self.observe(witness)` (`challenger/src/grinding_challenger.rs:48-53`), so the witness really is compared against nothing.
- All four `grind` implementations return `F::ZERO` at zero bits (`grinding_challenger.rs:138`, `:338`; `serializing_challenger.rs:220`, `:410`), so zero is the only value an honest prover can emit and the check cannot reject one.

The thing that could have made this wrong is `UniformGrindingChallenger`, and it is worth a note here because it is the exact inverse of the assumption:

- `check_witness_uniform` has **no** `bits == 0` shortcut and always absorbs, so on that path the witness is already bound at zero difficulty and this check would be dead.
- `grind_uniform(0)` is worse than dead — it goes through `grind_generic`, whose `find_any` returns whichever candidate a worker happens to reach first, and at zero bits every candidate passes. So it returns an arbitrary field element, and a canonical-zero check on that path would reject honest proofs nondeterministically.

`git grep` over this ref finds no callers of `grind_uniform` or `check_witness_uniform` outside `p3-challenger`, so the check is safe as written. If either ever gets a caller, this check and its siblings in #2106 need a carve-out.

One framing point on the error's doc comment, which is otherwise accurate: since nothing is absorbed, the free witness cannot influence any downstream challenge, so the exposure is proof malleability and non-uniqueness rather than forgery. The test's own docstring says exactly that ("a third party could rewrite it and keep a verifying proof"); the `fix(security)` prefix on #2106 reads a little stronger than the actual severity.

Nashtare (approved):

binary-pcs/src/verifier.rs
tcoratger: Verified all of it independently before acting, and every point holds. Two things came out of it.

## The uniform-grinding hazard is now pinned, not just noted

You are right that this is the case that would have made the check wrong, and right that it is the exact inverse of the assumption. Confirmed at the source: the honest grind short-circuits `if bits == 0 { return F::ZERO }`, while `grind_uniform` routes through `find_any` with no such shortcut, so at zero bits whichever candidate a worker reaches first wins. And `check_witness_uniform` absorbs unconditionally, so on that path the witness is already bound and this check would be dead.

Rather than add a runtime carve-out on speculation — your own verdict is that it is safe as written — the assumption is now a test in the same file as the check:

```rust
    #[test]
    fn zero_difficulty_grinding_is_pinned_to_the_zero_witness() {
        // Invariant: this crate's challenger grinds to the zero witness at a zero budget.
        //
        //              grind                     check
        //     0 bits   zero, absorbing nothing   true, absorbing nothing
        //     4 bits   a search that absorbs     resamples and compares
        assert_eq!(challenger().grind(0), F::ZERO);

        // The complementary half: a zero-bit witness check accepts every value handed to it.
        for witness in [F::ZERO, F::ONE, F::GENERATOR] {
            assert!(challenger().check_witness(0, witness));
        }
        ...
```

It asserts both halves deliberately, since the second is the precise reason the guard has to exist at all, and a reader who sees them together does not have to leave the file to understand why.

**Proof it bites.** Changing the honest zero-bit shortcut to return one instead of zero — the shape of an unconstrained zero-bit grind:

```
test verifier::tests::zero_difficulty_grinding_is_pinned_to_the_zero_witness ... FAILED
assertion `left == right` failed
  left: 1
 right: 0
```

Under that same mutation the honest zero-budget round trip also fails, with `NonCanonicalPowWitness { actual: 1 }` — the latent hazard made concrete, and exactly the "flaky rejection" failure mode a named pin is meant to replace. Three lines at the check site say the assumption out loud and point at the test.

## On severity

Agreed, and the commit already lands where you put it. This one's subject is crate-scoped rather than `fix(security)`, and its body says "the defect this leaves is malleability, not unsoundness — a prover still cannot prove a false statement: the grind guards nothing at zero difficulty, so there is nothing to bypass. What breaks is unique encoding." A grep over the commit for forgery-implying language turns up nothing, so I left the wording as it stands rather than weakening it further — it does still name hashing, dedup, identity and recursion, which is where non-uniqueness actually costs something.

Your point does land on **#2106**, whose subject is `fix(security)`. That one spans five crates, so a crate prefix does not fit, but the severity paragraph there says the same thing in the same terms.

## One note on the stack

The stack was auto-rebased server-side while I was working, onto a `main` that had moved three commits. I rebased this round's fixes onto that rebased root rather than over it, so nothing of GitHub's rebase was discarded. Full stack: 3964 tests, lint clean.

PR #2106, opened 2026-09-09
fix(security): reject non-canonical grinding witnesses at zero difficulty

tcoratger:
Depends on #2105 and the branches below it. This PR targets `fix/binary-pcs-canonical-pow-witness`; retarget after they land.

## The finding

The defect #2105 fixes in `binary-pcs` is **systemic**. Every proof-of-work witness field in the repository is unbound at zero difficulty, in five more crates.

```text
    check_witness(0, w)  ->  true, absorbs nothing
    Kind::Pow step       ->  elided entirely when the difficulty is zero
    proof witness field  ->  carried unconditionally, prover writes ZERO
```

| Crate | Unbound fields |
| --- | --- |
| uni-STARK | `ood_pow_witness` |
| FRI | `batch_pow_witness`, `query_pow_witness`, and every element of `commit_pow_witnesses` |
| Circle | `pow_witness`, and every element of `commit_pow_witnesses` |
| WHIR | four scalars across the plain, HVZK and base-case pipelines |
| STIR | four scalars, replicated per batched instance |

**This is reachable in configurations the repo ships.** `fri/src/config.rs` sets `batch_proof_of_work_bits: 0` or `commit_proof_of_work_bits: 0` in seven places across `new_testing`, `new_benchmark` and their zk variants; WHIR's own constructors set zero; STIR's tests run all three budgets at zero. uni-STARK's `ood_pow_bits` is a caller-supplied `usize` with no lower bound — and its proof field's doc comment reads *"Trivially valid (and unread) when … is `0`"*, documenting the hole without recognising it as one.

Two cases are worse than a scalar comparison:

- **FRI and Circle carry vectors**, and the verifier checked only `.len()` against the round count. A count check is not a value check — every element was free.
- **STIR replicates**, and `check_replicated_witnesses` only forces the batched copies to agree *with each other*. Rewriting all of them to the same wrong value passed. The before-state is the load-bearing evidence: with all eight slots across two instances rewritten, `verify_stir_multi` returned `Ok(...)`.

`batch-stark` is the only crate that already got this right, and it did so differently — it always describes the `Pow` step, zero included, so the witness is absorbed at every difficulty.

## Severity

**Proof malleability, not a soundness break.** No attacker can prove a false statement; proofs simply are not uniquely encoded. That matters wherever proof bytes are hashed, deduplicated, used as an identifier, or recursed over. This is the same reasoning already written on `binary-pcs`'s `NonEmptyPowWitnesses`.

## The remedy, and the two stronger ones not taken

This pins the canonical value: at zero difficulty the honest witness is `ZERO`, so anything else is rejected with a typed error before any transcript operation. It **changes no transcript and no serialized proof shape**, so every `.postcard` fixture passes unchanged — confirmed, no fixture byte moved.

Two stronger remedies exist and are deliberately deferred:

- **Always describe the `Pow` step, zero included** — `batch-stark`'s choice, and the recommended follow-up. It expresses the rule once in the transcript layer instead of fifteen hand-written comparisons a future field or crate can forget, and makes the binding structural: the witness is absorbed at every difficulty, so a rewrite desynchronises Fiat-Shamir and is caught by ordinary machinery. Its cost is an `observe_pow` semantics change touching every protocol plus a breaking transcript change in five crates, but that cost is paid once and can be sequenced behind a single `p3-challenger` change.
- **Make the field absent at zero difficulty** (`Option<Witness>`). Most type-honest, but it buys strictly less than the above for a comparable price: it removes the field, yet says nothing about whether a witness that *is* present was ever absorbed.

## The STIR subtlety

A shared grind runs at the **max** difficulty over the instances active at that site. So an instance configured at 0 bits legitimately carries a nonzero witness when batched with one at 20 bits, and pinning per-instance against that instance's own bits would reject honest provers. The check reads the shared `StirShape` the transcript is described with, and skips any site the batch actually grinds — so agreement becomes a consequence of canonicality rather than a substitute for it.

`a_batch_mixing_ground_and_ungrounded_instances_still_verifies` pins that completeness property. With naive per-site pinning it fails:

```
an honest batch mixing ground and ungrounded instances must verify:
  InvalidProofShape(NonCanonicalPowWitness { round: Round(0), stage: Folding })
```

## Tests

One mutation test per crate, each failing before the fix and passing after; the vector cases mutate a single element and leave the length alone, so a passing typed rejection proves the element check fired rather than the count check. Round trips at both zero and positive difficulty per crate confirm the honest path is intact.

## Known limitation

None of the five error enums is `#[non_exhaustive]`, so adding a variant is technically breaking for a downstream exhaustive `match` — unlike `binary-pcs` in #2105, whose enum is. In practice these are `thiserror` verification errors that callers propagate or print, and the alternative would lose the typed distinction. Worth a reviewer's eye. WHIR needed three variants rather than one, because its three sub-protocols each own an error enum and the base-case witness must be rejected by the base-case verifier.

### DIFF
--- a/fri/src/verifier.rs
+++ b/fri/src/verifier.rs
@@ -203,6 +203,22 @@ where
     FinalPolyMismatch,
     #[error("invalid proof-of-work witness for the {0} phase")]
     InvalidPowWitness(PowPhase),
+    /// A grinding witness that is not the value a zero difficulty budget admits.
+    ///
+    /// Checked with the other proof-shape rejections, before any transcript work.
+    //
+    // Why: at zero bits neither side touches the sponge.
+    //
+    //     prover  : grind(0)  -> returns zero, absorbs nothing
+    //     verifier: the Pow step is elided, so no witness is read at all
+    //
+    // The field is then bound to nothing: any value rides along and still verifies.
+    // Zero is the only value an honest prover emits, so zero is the only value accepted.
+    #[error("non-canonical grinding witness for the {phase} phase at zero difficulty")]
+    NonCanonicalPowWitness {
+        /// Phase whose witness is not the canonical zero.
+        phase: PowPhase,
+    },
     /// A query point coincides with an opening point, so the quotient denominator is zero.
     #[error(
         "batch {batch}, matrix {matrix}, point {point}: query point coincides with the opening point"
@@ -291,6 +307,56 @@ where
 /// FRI folds and rolls. The first element of each pair indicates the round of
 /// fri in which the input should be rolled in. The second element is the opening.
 pub type FriOpenings<F> = Vec<(usize, F)>;
+
+/// Checks that every grinding witness a zero difficulty leaves unread carries the one
+/// value a zero budget admits, before any transcript operation.
+///
+/// A positive budget needs no check for its own phase: there the witness is absorbed and
+/// its sampled bits are compared, so the difficulty itself pins the field.
+///
+/// The three phases are checked independently, because their difficulties are independent.
+//
+// Why: `GrindingChallenger::check_witness` returns `true` at `bits == 0` without
+// absorbing, and the typed transcript elides the step altogether, which leaves the
+// witness compared against nothing and free to be any value.
+//
+//     bits = 0 -> prover emits zero, verifier reads nothing -> pin the field here
+//     bits > 0 -> prover grinds,     verifier resamples     -> the grind pins it
+//
+// The commit phase carries one witness per round, so every entry is pinned rather than
+// only the vector's length: a count check is not a value check.
+pub(crate) fn check_canonical_pow_witnesses<Val, Challenge, FriMmcs, InputErr, InputProof>(
+    params: &FriParameters<FriMmcs>,
+    proof: &FriProof<Challenge, FriMmcs, Val, InputProof>,
+) -> Result<(), FriError<FriMmcs::Error, InputErr>>
+where
+    Val: Field,
+    Challenge: Field,
+    FriMmcs: Mmcs<Challenge>,
+    InputErr: core::fmt::Debug,
+{
+    if params.batch_proof_of_work_bits == 0 && proof.batch_pow_witness != Val::ZERO {
+        return Err(FriError::NonCanonicalPowWitness {
+            phase: PowPhase::Batch,
+        });
+    }
+
+    if params.commit_proof_of_work_bits == 0
+        && proof.commit_pow_witnesses.iter().any(|w| *w != Val::ZERO)
+    {
+        return Err(FriError::NonCanonicalPowWitness {
+            phase: PowPhase::CommitPhase,
+        });
+    }
+
+    if params.query_proof_of_work_bits == 0 && proof.query_pow_witness != Val::ZERO {
+        return Err(FriError::NonCanonicalPowWitness {
+            phase: PowPhase::Query,
+        });
+    }
+
+    Ok(())
+}
 
 /// Verifies a FRI proof.
 ///
@@ -472,6 +538,10 @@ where
             got: proof.commit_pow_witnesses.len(),
         });
     }
+
+    // A zero grinding budget leaves the witness unread, so the three fields are pinned
+    // here rather than by their grinds.
+    check_canonical_pow_witnesses(params, proof)?;
 
     // Ensure that the final polynomial has the expected degree.
     //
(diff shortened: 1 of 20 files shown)


uni-stark/src/verifier.rs
```diff
@@ -543,6 +543,12 @@ where
         return Err(InvalidProofShapeError::OpenedValuesDimensionMismatch.into());
     }
 
+    // A zero grinding budget elides the out-of-domain `Pow` step, so nothing downstream
+    // reads the witness the proof carries. Pin it to the one value an honest prover emits.
+    if config.ood_proof_of_work_bits() == 0 && *ood_pow_witness != Val::<SC>::ZERO {
+        return Err(InvalidProofShapeError::NonCanonicalOodPowWitness.into());
```
Nashtare, line 552: **This fix is incomplete: `batch-stark` has the same hole, twice, and is not touched by this PR.**

`batch-stark` carries two grinding witnesses that reach the verifier by exactly the route this line closes for uni-STARK:

- `ood_pow_witness: Val<SC>` (`batch-stark/src/proof.rs:38`). The prover calls `observe_pow(OOD_POW, self.shape.ood_pow_bits)` (`batch-stark/src/transcript.rs:675`); the verifier's `ood_phase` passes the proof's witness into `state.observe_pow(...)`, which is `challenger.check_witness(bits, witness)` (`challenger/src/fs/state/verifier.rs:374`). At `ood_pow_bits == 0` that returns `true` without absorbing. This is character-for-character uni-STARK's `ood_pow_witness` — the field this PR just pinned.
- `lookup_pow_witness: Option<Val<SC>>` (`proof.rs:33`), same route via `LOOKUP_POW` / `shape.lookup_pow_bits` (`transcript.rs:613`). The `Option` **shape** is handled properly already — `UnexpectedLookupPowWitness` when there are no lookups, `MissingLookupPowWitness` when there are (`transcript.rs:809-828`) — but neither rejects a non-zero **value** at zero difficulty.

Worth noting why this is easy to miss: the typed layer records the step on both sides with `Length::Fixed(bits)` even at zero, so the *difficulty* is bound by the pattern and the pattern check passes. Only the witness *value* floats free. Reading the pattern gives the impression the step is covered.

Everything else in the PR checks out. The FRI side correctly pins every entry of `commit_pow_witnesses` rather than just its length, which is the distinction the comment there calls out and the one that actually matters. whir's `final_pow_witness` is covered at `whir/src/pcs/verifier/mod.rs:145` and tested. circle, stir and uni-STARK are all covered.

fri/src/verifier.rs
```diff
@@ -279,6 +295,56 @@ where
...
+// Why: `GrindingChallenger::check_witness` returns `true` at `bits == 0` without
+// absorbing, and the typed transcript elides the step altogether, which leaves the
+// witness compared against nothing and free to be any value.
+//
+//     bits = 0 -> prover emits zero, verifier reads nothing -> pin the field here
+//     bits > 0 -> prover grinds,     verifier resamples     -> the grind pins it
+//
+// The commit phase carries one witness per round, so every entry is pinned rather than
+// only the vector's length: a count check is not a value check.
+pub(crate) fn check_canonical_pow_witnesses<Val, Challenge, FriMmcs, InputErr, InputProof>(
```
Nashtare, line 328: Minor, and only because it appears twice: `check_canonical_pow_witnesses::<Val, Challenge, FriMmcs, InputMmcs::Error, _>(...)` needs five explicit parameters at both call sites (`two_adic_pcs.rs:671` and `verifier.rs:531`) with only the last inferred. Since both callers already have the concrete types in scope, a thin wrapper — or moving the check behind a method on `FriParameters` — would keep the call sites readable.

No objection to the double call itself; the comment at `two_adic_pcs.rs:668` explains it and the redundancy is cheap.

uni-stark/src/verifier.rs
tcoratger: You are right, and the analysis this PR shipped with was wrong in exactly the way you describe. Both holes are now closed, and the commit message that claimed batch-stark "already had this right" is gone with them.

## Confirmed before fixing

At the default config both difficulties are zero, so I put the mutation behind an inverted assertion first — "the current verifier accepts this" — and ran it against the unfixed code:

```
test a_noncanonical_ood_pow_witness_at_zero_difficulty_is_rejected ... ok
test a_noncanonical_lookup_pow_witness_at_zero_difficulty_is_rejected ... ok
```

Both pass, i.e. `verify_batch` returns `Ok(())` on a proof whose witness a third party rewrote. In the final form, against the unfixed verifier:

```
thread '...' panicked at batch-stark/tests/simple.rs:2354:6:
  a rewritten out-of-domain witness must be rejected: ()
thread '...' panicked at batch-stark/tests/simple.rs:2260:6:
  a rewritten lookup witness must be rejected: ()
```

`: ()` there is the `Ok(())`. After the fix both pass on the typed rejection.

## Why the earlier reading was wrong

Your diagnosis is the whole of it, and it is worth restating because it is the part that misled me: always-describing the step binds the **difficulty**, through `Length::Fixed(bits)`, and says nothing about the **value**. `check_witness` still short-circuits at zero bits without absorbing. So the pattern check passes and the field floats free — reading the pattern gives exactly the false confidence you describe. The two proof-struct doc comments made it worse by asserting the field was "trivially valid (and unread)" at zero bits; both are rewritten.

## Where the rejections sit

Inside the verifier driver, not in `verify_batch`'s shape block:

- lookup phase: after the two existing shape checks — which I left alone, they already handle the `Option` correctly — and immediately before the pow step.
- out-of-domain phase: first statement, immediately before its pow step.

The driver is the one place the difficulty and the witness meet, so every caller gets the check, including one that assembles the opening argument itself and replays the transcript. A check parked in `verify_batch` would be skippable that way.

Both rejections release the driver's completeness check on the way out, as the two existing shape rejections already do — the driver is mid-pattern there and dropping it unfinished would panic on top of an error already travelling to the caller. Two driver-level unit tests pin that: they drop an aborted driver and assert no panic.

## One thing for release, not for this PR

`BatchTranscriptFailure` is **not** `#[non_exhaustive]`, so the two new variants are technically breaking for a downstream exhaustive match. Nothing in the workspace matches it exhaustively, so the repo is unaffected. It is called out in the commit message. `p3-binary-pcs`'s error enum does carry the attribute if you want it added here, but that is the same category of change so I left the call to you rather than folding it in.

## Unchanged

Six `.postcard` fixtures all pass untouched — this remedy adds no transcript step and no wire field, only a comparison against a value the honest prover already writes. `git status --porcelain -- '*.postcard'` is empty.

Full suite: 3895 passed, 0 failed.

fri/src/verifier.rs
tcoratger: Fixed, and it turned out neither suggestion was needed: **all five parameters infer, so the annotation was simply unnecessary.** Four come from the argument types, and the error parameter resolves through `?` — `FriError` carries no `#[from]` variants, so the reflexive conversion is the only applicable impl and there is nothing to disambiguate.

Both call sites are now one line:

```rust
verifier::check_canonical_pow_witnesses(&self.fri, proof)?;   // two_adic_pcs.rs
check_canonical_pow_witnesses(params, proof)?;                // verifier.rs
```

I went this way rather than either suggestion because a method on the parameters type would make a configuration struct depend on the proof and error types to gain nothing inference already gives, and a thin wrapper would add a second name for a single call. This is zero new API and zero new items, and the helper stays crate-private.

The double call and the comment explaining it are untouched, as you said they should be.

Nashtare (approved):

PR #2107, opened 2026-09-09
test(fri): pin every shape field to the transcript seed

tcoratger:
Depends on #2106 and the branches below it. This PR targets `fix/canonical-pow-witnesses`; retarget after they land.

## What this changes

The typed-transcript campaign carries a rule: every configuration knob must move the fingerprint or the instance label, and there must be a test proving it. The reasoning is blunt — untested, nothing distinguishes a real binding from a 32-byte constant that happens to get absorbed, and one refactor silently turns the second into the first.

FRI had the weakest coverage in the repository, and it is the protocol most others delegate to. `FriShape` has eight fields; exactly one of them was walked.

**All eight were in fact bound.** This closes a coverage gap, not a hole — worth stating plainly so nobody reads it as a vulnerability fix.

| Field | Mutation | Bound by |
| --- | --- | --- |
| `log_arities` value | `[3,3,2] -> [3,3,3]` | instance label, per-round values |
| `log_arities` length | `-> [3,3,2,2]` | pattern loop, one more step group |
| `log_arities` order | `-> [2,3,3]` | instance label only — same step sequence |
| `final_poly_len` | `+1` | `Length::Fixed` on `final_poly` |
| `commit_pow_bits` | `0 -> 1` | step presence, elided at zero |
| `query_pow_bits` | `0 -> 1` | step presence, elided at zero |
| `num_queries` | `+1` | `Length::Fixed` on `query_indices` |
| `index_bits` | `+1` | the `TypeTag::Bits { width }` of that step |
| `log_blowup` | `+1` | instance label |
| `max_log_arity` | `+1` | instance label |

Two extra tests cover what a one-step table structurally cannot: the `1 -> 2` transition for both pow fields, which is a `Length` change rather than a presence change; and a direct assertion that the query-index step's tag carries `index_bits`. That last one matters most — it is the field whose binding route is least obvious and the one that is soundness-relevant, since a narrower draw silently shrinks the query space and the proximity-test soundness error with it.

## Why pairwise is the strong form

`no_two_one_step_mutations_collide` asserts no two single-field mutations land on the same seed, over all 45 pairs. That is strictly stronger than a per-field walk, and the difference is not theoretical — collapsing `log_blowup + max_log_arity` into one absorbed number *passes* the per-field test and *fails* this one:

```
test every_field_of_the_shape_reaches_the_seed ... ok
test no_two_one_step_mutations_collide ... FAILED
assertion `left != right` failed: `log_blowup` and `max_log_arity` land on the same seed
  left: 1782566899
 right: 1782566899
```

## Proving the tests bite

Five deliberate-breakage experiments, all reverted:

- dropping `[log_blowup, max_log_arity]` from the separator → the field walk fails;
- collapsing them into one number → only the pairwise test fails (above);
- hardcoding the query-index width → both the walk and the tag assertion fail;
- dropping the per-round arity values from the label → the `log_arities value` row fails;
- on the PCS side, clamping a claimed-evaluation count → its row fails.

## FRI PCS

`PcsShape` was covered piecemeal by five separate tests — effectively complete, but with no table, so a newly added field would go unnoticed. It gets the same table-driven pairwise treatment. Existing tests that assert something the table does not (the opening grouping, the commitment count) are kept; the ones it subsumes are folded in, since redundant tests hurt readability.

### DIFF
--- a/fri/src/transcript.rs
+++ b/fri/src/transcript.rs
@@ -496,6 +496,7 @@ mod tests {
     use std::panic::{AssertUnwindSafe, catch_unwind};
 
     use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+    use p3_challenger::fs::TypeTag;
     use p3_challenger::{CanSample, DuplexChallenger};
     use p3_field::PrimeCharacteristicRing;
     use p3_field::extension::BinomialExtensionField;
@@ -529,6 +530,13 @@ mod tests {
         }
     }
 
+    /// The shape every mutation below is measured against.
+    ///
+    /// Three rounds, so a per-round step is described more than once.
+    fn plain_shape() -> FriShape {
+        shape_with(vec![3, 3, 2])
+    }
+
     /// The first challenge a shape's seed produces.
     fn first_challenge(shape: &FriShape) -> F {
         let mut challenger = fresh_challenger();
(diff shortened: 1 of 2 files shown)


fri/src/transcript.rs
```diff
@@ -537,19 +545,156 @@ mod tests {
...
     }
 
-    #[test]
-    fn the_arity_of_every_round_reaches_the_seed() {
-        // Reorderings share a round count, a total, and therefore a step sequence.
+    /// Every field of the shape, each moved one step away from `plain_shape`.
+    ///
+    /// One entry per configuration knob, so a knob added without a binding shows up here.
+    /// The schedule is a vector, so it contributes one entry per way of moving it.
+    fn one_step_from_plain() -> Vec<(&'static str, FriShape)> {
```
Nashtare, line 554: Checked the sweep against the struct: all eight fields of `FriShape` are covered — `log_arities` three ways (value, length, order), plus `final_poly_len`, `commit_pow_bits`, `query_pow_bits`, `num_queries`, `index_bits`, `log_blowup`, `max_log_arity`. Complete as of this commit, and the split between "moves the step sequence" and "only the instance label can carry it" is the right way to organise it.

The docstring claims more than the code enforces, though:

> One entry per configuration knob, so a knob added without a binding shows up here.

Nothing makes that true. A field added to `FriShape` tomorrow gets no mutation here and no failure — the sweep is a hand-maintained list, and the test passes just as happily without the new entry.

Destructuring the shape at the top of this function with no `..` would fix it:

```rust
let FriShape {
    log_arities, final_poly_len, commit_pow_bits, query_pow_bits,
    num_queries, index_bits, log_blowup, max_log_arity,
} = plain_shape();
```

Then a new field is a compile error in this test, which is what the sentence is promising. The bindings themselves can stay unused — the point is the exhaustiveness check, not the values.

Nashtare (approved):

fri/src/transcript.rs
tcoratger: Applied, to both shapes in this commit — and it turned up a wrinkle worth recording, because it means the guarantee is slightly weaker than the sentence I would naively have written.

```rust
        let FriShape {
            log_arities: _,
            final_poly_len: _,
            commit_pow_bits: _,
            query_pow_bits: _,
            num_queries: _,
            index_bits: _,
            log_blowup: _,
            max_log_arity: _,
        } = plain_shape();
```

`field: _` rather than `_field`: it participates in exhaustiveness identically and leaves no unused binding to silence. `clippy::unneeded_field_pattern` would prefer `..`, but it is in the restriction group and this workspace does not enable it, so clippy stays clean.

## What actually happens when a field is added

A new field trips **E0063** at the struct literals first, and that error masks the pattern one. Filling the literals in is enough to make the crate compile again — and would leave the mutation table untouched with the walk still passing, which is exactly the gap you described. Only once the literals are satisfied does the table's error surface:

```
error[E0027]: pattern does not mention field `probe_field`
   --> fri/src/transcript.rs:561:13
    |
561 |           let FriShape {
    |  _____________^
...
570 | |         } = plain_shape();
    | |_________^ missing field `probe_field`
help: include the missing field in the pattern
help: if you don't care about this missing field, you can explicitly ignore it
help: or always ignore missing fields here
```

So the destructuring is a second line of defence rather than the first, and rustc's third suggestion is `..` — meaning what is strictly enforced is that the author has to *confront this function*, not that a row appears in the table. That is still much better than nothing, but I did not want the docstring claiming the stronger thing. It now says only what the compiler does:

> One entry per configuration knob.
> A field added to the shape stops the destructuring below from compiling.

I dropped the old wording's "so a knob added without a binding shows up here" for two reasons: it made the guarantee conditional on the field lacking a seed binding, when the compiler now stops any new field regardless, and "shows up here" understated a hard error.

Same treatment for the PCS shape in `pcs_transcript.rs`, whose docstring made the same promise.

## Other unenforced claims in those two modules

You prompted me to sweep for the same failure mode, and there were three more. One I fixed: a fixture line said "two runs over the same shape, one bracketing an empty delegation" when both runs bracket one — it is a verifier-versus-prover comparison. Two I am reporting rather than fixing, because neither can be enforced from the test side:

- `the_delegation_bracket_leaves_the_sponge_untouched` asserts "the markers are structural, they record the delegation and absorb nothing". If the markers did absorb, both sides would absorb identically and every assertion would still pass. The shape pushes the bracket unconditionally, so no run can skip it either. Making that name honest needs a production hook or a rename to what is really pinned — that both sides walk the bracket identically.
- "Each mutation moves exactly one field one step" is hand-maintained; a two-field row would pass unnoticed. Self-evident from the table directly above it, and checking it would mean comparing rows field by field.

Coverage itself is unchanged and still complete: 91 lib tests pass, both sweeps and the pairwise walk among them.

PR #2108, opened 2026-09-09
test(challenger): assert no two protocols share a transcript seed

tcoratger:
Depends on #2107 and the branches below it. This PR targets `test/fri-shape-seed-coverage`; retarget after they land.

## What this changes

Fifteen protocols now derive a `DomainSeparator` from a shape type. The fingerprint and the seed are the entire domain-separation argument of the campaign — and nothing asserted that any two of them differ.

This turns that argument into a checked property, and adds the shared harness the check needs.

## What separation actually rests on

Every protocol passes the same version byte, because the version is a *format* version each protocol owns. So the **name carries the whole burden**, and nothing was comparing names.

The sharpest latent risk is naming discipline. `p3-fri` / `p3-fri-pcs` and `p3-whir` / `p3-whir-hvzk` stand in a genuine prefix relation, and are separated only because the name length lives at `protocol_id[63]`. That safety was an implementation detail nobody tested at the protocol level; there is now a test that names the mechanism.

## The harness — `p3-challenger`, `test-utils` feature

Follows the precedent `p3-commit` already sets: a feature-gated `pub mod testing`, pulled in as a dev-dependency.

```rust
pub fn seed_digest<U: Unit>(separator: &DomainSeparator<U>) -> SeedDigest;
pub fn assert_seeds_pairwise_distinct<L: Display>(seeds: &[(L, SeedDigest)]);
pub struct Recorder<T>;   // promoted from four hand-written copies
```

`seed_digest` digests the byte stream rather than a sampled challenge. That needed one non-test change: extracting `seed_bytes()` out of `DomainSeparator::seed()`, which now delegates to it — behaviour-preserving, and the existing wire-layout test pins it. The stream is assembled before the alphabet packs it, so the digest is alphabet-independent, and a caller needs no sponge, no permutation and no RNG seed to compare two seeds. That means the test exercises the *binding* rather than the challenger.

`SeedDigest` is opaque and only comparable — no accessor, since nothing needs one — with a hex `Debug` so a failed assertion prints something readable.

The harness is `no_std`-clean and absent from the default surface.

## The cross-protocol test lives in a leaf crate

It has to. Every protocol crate depends on `p3-challenger`, so a test there that imports them is a dependency cycle. `p3-examples` is a genuine leaf and already dev-depends on `p3-commit` with `test-utils` and on most of the protocol crates.

It needed **no new public constants**: every shape type is already publicly reachable and `domain_separator()` is already a public method.

Four assertions, over 15 protocols × 3 configurations = 45 seeds and 990 pairs:

1. `every_protocol_is_listed_here` — a further protocol must join the list, or its name is never compared against the others.
2. `the_protocol_name_is_the_only_field_that_separates_two_protocols` — normalises every version byte and asserts the identifiers stay pairwise distinct. Two protocols parted *only* by that byte would collide the day either one bumps it.
3. `a_shared_name_prefix_is_separated_by_the_name_length_byte` — pins the mechanism for both prefix pairs, not just the outcome.
4. `no_two_configurations_of_any_two_protocols_share_a_seed` — the full pairwise cross-product.

**No collision exists.** All 15 identifiers and all 45 seeds are pairwise distinct, verified rather than assumed.

## Proving it bites

Pointing `p3-fri-pcs` at `p3-fri`:

```
assertion `left != right` failed: `p3-fri/plain` and `p3-fri-pcs/plain`
  are parted only by their version byte
```

Worth reporting honestly: under a shared name the *seed* test still passed, because the two patterns differ. A name collision is not yet a seed collision — it removes the only field designed to prevent one. That is exactly why the identifier assertion exists separately from the seed assertion.

## Conversions

Four crates whose mutation fixtures already live in a labelled table — `fri`, `fri/pcs`, `uni-stark`, `batch-stark` — move onto the harness, and their per-field walk plus pairwise test fold into one strictly stronger test. The rest are left alone and the PR says why: `stir` and `circle` drive their walks from mutator tables or bespoke assertions, and `whir`, `multi-stark` and `sumcheck` use a closure-per-field helper. There is no `Vec` of configurations to hand the harness in those, so converting means rewriting the fixture — churn, not de-duplication. The cross-protocol sweep covers their defaults plus two moves each regardless.

## Known limitation

Name uniqueness is checked by a test in a leaf crate, not by a compile-time or runtime registry. A protocol crate can still be published in isolation with a duplicate name; the workspace test is what catches it, and `every_protocol_is_listed_here` is what catches someone forgetting to add themselves to the workspace test.

### DIFF
--- a/examples/tests/transcript_separation.rs
+++ b/examples/tests/transcript_separation.rs
@@ -1,0 +1,707 @@
+//! Cross-protocol domain separation for the typed Fiat-Shamir layer.
+//!
+//! # Overview
+//!
+//! Fifteen protocols in this workspace seed their transcript from a domain separator.
+//!
+//! The version byte is a format version each protocol owns, so names carry the separation.
+//!
+//! ```text
+//!     protocol_id = [ 1 | NAME | 0 .. 0 | NAME.len() ]
+//!                     ^     ^                 ^
+//!                     |     |                 disambiguates zero-padded prefixes
+//!                     |     the only field that differs between protocols
+//!                     the same byte for all fifteen
+//! ```
+//!
+//! Separation therefore rests entirely on `NAME`, and this file is where that is checked.
+//!
+//! # Placement
+//!
+//! Every protocol crate depends on `p3-challenger`, so the check cannot live there.
+//! `p3-examples` is a leaf: nothing depends on it, and it already pulls in most of the fifteen.
+
+use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
+use p3_batch_stark::BatchShape;
+use p3_challenger::DuplexChallenger;
+use p3_challenger::fs::{DomainSeparator, FieldUnit, PROTOCOL_ID_LEN};
+use p3_challenger::testing::{SeedDigest, assert_seeds_pairwise_distinct, seed_digest};
+use p3_circle::CirclePcsShape;
+use p3_field::extension::BinomialExtensionField;
+use p3_fri::{FriShape, PcsShape};
+use p3_multi_stark::fractional_gkr::FractionGkrShape;
+use p3_multi_stark::lookup::transcript::{LookupInstanceShape, LookupShape};
+use p3_multi_stark::rounds::AirDegrees;
+use p3_multi_stark::transcript::{MultiStarkInstanceShape, MultiStarkShape};
+use p3_multi_stark::zerocheck::transcript::ZerocheckShape;
+use p3_stir::{SecurityAssumption, StirInstanceShape, StirRoundShape, StirShape};
+use p3_sumcheck::generic_degree::GenericDegreeShape;
+use p3_sumcheck::strategy::Basis;
+use p3_sumcheck::transcript::SumcheckShape;
+use p3_sumcheck::zk::ZkSumcheckShape;
+use p3_uni_stark::StarkShape;
+use p3_whir::{
+    FoldingFactor, ProtocolParameters, WhirConfig, WhirShape, ZkParameters, ZkWhirConfig,
+    ZkWhirShape,
+};
+
+/// Base field every separator below is derived over.
+///
+/// One field for all fifteen, so nothing is separated by the field choice.
+type F = BabyBear;
+
+/// Extension field every separator below draws its challenges from.
+type EF = BinomialExtensionField<F, 4>;
+
+/// Permutation the WHIR configurations are parameterised by.
+type Perm = Poseidon2BabyBear<16>;
+
+/// Challenger the WHIR configurations are parameterised by.
+///
+/// WHIR carries the challenger in its config type, but seeding never touches a sponge.
+type Ch = DuplexChallenger<F, Perm, 16, 8>;
+
+/// Sponge alphabet every protocol in the workspace seeds through.
+type Alphabet = FieldUnit<F>;
+
+/// One labelled configuration of one protocol, reduced to the separator it derives.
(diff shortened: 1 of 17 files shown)


examples/tests/transcript_separation.rs
```diff
@@ -0,0 +1,698 @@
...
+/// Sponge alphabet every protocol in the workspace seeds through.
+type Alphabet = FieldUnit<F>;
+
+/// One labelled configuration of one protocol, reduced to the separator it derives.
+type Case = (String, DomainSeparator<Alphabet>);
+
+/// Number of protocols on the typed transcript layer.
+///
+/// A protocol added without an entry below leaves its name unchecked against the others.
+const NUM_PROTOCOLS: usize = 15;
```
Nashtare, line 73: `NUM_PROTOCOLS = 15`, but two doc comments in this file still say thirteen:

- line 5: "Thirteen protocols in this workspace seed their transcript from a `DomainSeparator`."
- line 80, in `CASES_PER_PROTOCOL`'s own doc: "`13 protocols x 3 configurations = 39 seeds -> 741 pairs`"

At 15 the real figures are 45 seeds and 990 pairs. The assertions read the const so nothing is actually wrong, but the worked arithmetic is the thing a reader checks the sweep's cost against, and it is understating it by a quarter.

Separately, on `every_protocol_is_listed_here` below: it asserts `default_cases().len() == NUM_PROTOCOLS`, which guards the list against a hand-maintained constant rather than against the workspace. A sixteenth protocol on the typed layer fails nothing until someone bumps both halves — the same shape as the sweep in #2107. I do not think there is a clean way to enumerate protocols across a workspace at compile time, so this is worth knowing rather than fixing; the docstring's "A protocol added without an entry below leaves its name unchecked against the others" is accurate about the risk but reads like the test catches it.

The design itself is good, and normalising the version byte to zero before comparing is the detail I would have most expected to be missing — a pair separated only by version collides the day either one bumps, and this catches exactly that.

Nashtare (approved):

examples/tests/transcript_separation.rs
tcoratger: Both points taken. The first was already corrected in a documentation pass that landed after you looked — worth saying so rather than leaving it ambiguous:

```
//! Fifteen protocols in this workspace seed their transcript from a domain separator.

///     15 protocols x 3 configurations = 45 seeds -> 990 pairs
```

45 and 990 are the real figures, as you say. You were right that the arithmetic is the thing a reader checks the cost against — the stale version understated it by a quarter, and the same pass found three other places in this file still saying thirteen.

The second point is the more interesting one, and you are right that the comment read like the test caught something it does not. It now says exactly what it does and does not catch:

```
    // This compares two hand-maintained numbers against each other, and nothing wider.
    //
    //     builder added, count not bumped  ->  caught here
    //     count bumped, builder missing    ->  caught here
    //     new protocol, neither touched    ->  not caught
    //
    // A protocol on the typed layer that never joins the list is never compared at all.
    //
    // Enumerating them across a workspace at compile time has no clean form.
    //
    // So the list is a convention this test keeps consistent, not one it discovers.
```

I agree there is no clean way to enumerate protocols across a workspace from a leaf crate, so I documented the boundary rather than inventing a fragile mechanism for it. The same critique applies to the sweep in #2107, and there the shape *is* a struct, so it has a real fix — destructuring with no rest pattern turns a new field into a compile error. That is going in on that PR.

Thanks for the note on normalising the version byte. That one only exists because a rebase in this stack collided on a version bump, which made the failure mode concrete rather than hypothetical.

PR #2109, opened 2026-09-09
feat(security): check recorded grinding difficulties against the security model

tcoratger:
Depends on #2108 and the branches below it. This PR targets `test/typed-transcript-separation`; retarget after they land. It is the last item of the typed-transcript campaign.

## Why this exists

Grinding bits are the one soundness parameter that lives in two places at once: the transcript, which decides how much work a prover actually pays, and the security accounting, which decides how many bits the parameter set is *reported* to buy. Nothing compared them. A mismatch is a silently overstated security claim — a number in a report nobody re-ran.

Now that every proof-of-work step is a recorded `Kind::Pow` interaction carrying its difficulty as `Length::Fixed(bits)`, the two halves can be checked against each other. This makes a mismatch a test failure.

## The obstacle was vocabulary

Three disjoint naming schemes for the same sites, and no mapping between them anywhere in the tree. `GRINDING_VOCABULARY` is that table, keyed on `(protocol, label)` — never on the label alone, since `query_pow` appears in four crates and `ood_pow` in two.

| Protocol | Label | Site | Carrier | Report label | Zero bits |
| --- | --- | --- | --- | --- | --- |
| `p3-uni-stark` | `ood_pow` | `OutOfDomain` | `out_of_domain` | `deep-ali` | `Elided` |
| `p3-batch-stark` | `ood_pow` | `OutOfDomain` | `out_of_domain` | `deep-ali` | `Always` |
| `p3-batch-stark` | `lookup_pow` | `LookupChallenge` | `lookup_challenge` | `logup-fingerprint` | `WhenPhaseRuns` |
| `p3-fri-pcs` | `batch_pow` | `BatchCombination` | `batch_combination` | `batch-combination` | `Elided` |
| `p3-fri` | `commit_pow` | `LdtCommitPhase` | `FriRegime::commit_pow_bits` | `ldt-commit-phase` | `Elided` |
| `p3-fri` | `query_pow` | `LdtQueryPhase` | `FriRegime::query_pow_bits` | `ldt-query-phase` | `Elided` |

## Where it lives

Split along the line the dependency graph already has.

The vocabulary is a *fact about the protocols*, reviewable as a table, so it sits in `p3-security` beside `GrindingSites`. Crucially `check` takes `&[RecordedGrind]` rather than an `InteractionPattern`, so **`p3-security` gains no dependency on `p3-challenger`** and an audit tool in production can call it. Only the five-line pattern reader — `pow_difficulties` — lives in `p3-challenger`'s `test-utils`, because it exists solely to feed assertions.

## The zero-bit convention split

Sites do not agree on how to record zero, so the convention is declared per row and enforced in both directions:

- **`Elided`** — step present iff credited bits > 0. A described zero-bit step is unexpected; an absent step with bits credited is missing.
- **`Always`** — `batch-stark`'s `ood_pow`, whose witness travels in the proof unconditionally. Absence is an error *even at zero credited bits*, which is what makes the convention bite rather than decorate.
- **`WhenPhaseRuns`** — `batch-stark`'s `lookup_pow`. Absence carries no constraint, and that is argued rather than fudged: with no interactions `logup::security_term` returns `None`, so there is no term for those bits to boost and they are demonstrably inert.

## Proving it bites

Three deliberate desynchronisations, all reverted. The understating direction:

```
`new_testing`: ldt-query-phase grinding disagrees: the security model credits 0 bits,
  `p3-fri`'s `query_pow` step demands 1
```

And the dangerous direction — the transcript stops grinding while the model still credits it:

```
batch=0 ood=8 commit=0 query=0: out-of-domain grinding disagrees:
  the security model credits 8 bits, `p3-uni-stark` describes no `ood_pow` step, so 0 are demanded
```

One negative result worth recording: my first attempted desync did not fire, because the documented recipe overrides that field by struct update and the FRI-only stack lists no protocol owning the site. Both behaviours are correct; it just is not a reachable desync.

## What the guarantee actually is

Stated precisely, because the narrow reading and the broad one are easy to conflate:

> No parameter set **that a test instantiates** reports grinding bits it does not pay for.

That is narrower than "no mismatch exists in the tree". **Every caller of the budget check is inside a `mod tests` or an integration test** — there are zero production callers. A configuration added to `examples/`, or any downstream parameter set, is unchecked. The vocabulary table would catch it, but nothing on a proving or verifying path consults it.

The check takes `&[RecordedGrind]` rather than a pattern precisely so that an audit tool *could* call it without dragging a transcript dependency into `p3-security` — but today none does, and that is the whole distance between the two readings.

A test-only cross-check is the right first move: the two numbers were never compared at all before, and widening it means deciding where a production check belongs and what it costs on a hot path. That is a separate decision, not a footnote here.

## What was checked by hand

Every named `FriParameters` constructor agrees. The only non-test callers of `StarkSecurityParams::from_air` are the four sites in `examples`, and all four are consistent — the one plausible overstatement among them is already blocked by an assertion in `CirclePcs::new`. So no parameter set the tests reach, and none of the four in `examples`, reports bits it does not pay for.

## Scope, and what each remaining protocol needs

In scope: FRI, FRI PCS, uni-STARK, batch-STARK — the only protocols where both halves are already callable.

- **Circle** is cheapest: it reads `commit_pow` and `query_pow` straight from the shared `FriParameters`, so `security_regime()` already carries the right numbers. It needs a `p3-security` dependency and two rows.
- **WHIR** is hardest: its grinding is *derived* from a budget, not configured, and the security side is a disjoint model reporting `f64` bounds with no per-site accessor. Its rounds carry different difficulties, so a single `usize` per site is not enough — the budget needs a per-round vector. One PR per pipeline.
- **STIR** has no security module at all despite depending on `p3-security`; its bits come from a closure that solves for the gap to a buffered security level. A model has to exist before there is a credited number to compare against. Two PRs: model, then accessor.
- **multi-STARK** needs no rows. It records no `Kind::Pow` step of its own — the grind happens inside the delegated sumcheck patterns, under those protocols' names. Which also means it cannot be checked in isolation: the credited bits live in a config it passes down and the recorded bits in a transcript it does not own, so the check must span the bracket. Blocked on giving the sumcheck a security model at all.

## Public surface

No `_POW` label constant and no protocol `NAME` was exported. Each protocol asserts its own rows from inside its own crate, where the private constants are visible. One `const` became `pub(crate)` so `fri`'s two transcript modules can check their two halves of one `FriParameters` together.

### DIFF
--- a/fri/src/pcs_transcript.rs
+++ b/fri/src/pcs_transcript.rs
@@ -535,16 +535,25 @@ pub enum PcsTranscriptFailure {
 #[cfg(test)]
 mod tests {
     use alloc::vec;
+    use core::str::from_utf8;
 
     use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
-    use p3_challenger::testing::{SeedDigest, assert_seeds_pairwise_distinct, seed_digest};
+    use p3_challenger::testing::{
+        SeedDigest, assert_seeds_pairwise_distinct, pow_difficulties, seed_digest,
+    };
     use p3_challenger::{CanSample, DuplexChallenger};
     use p3_field::PrimeCharacteristicRing;
     use p3_field::extension::BinomialExtensionField;
+    use p3_security::grinding::{
+        GRINDING_VOCABULARY, GrindingBudget, GrindingSite, RecordedGrind, ZeroBitConvention,
+        grinding_step,
+    };
     use rand::SeedableRng;
     use rand::rngs::SmallRng;
 
     use super::*;
+    use crate::FriShape;
+    use crate::transcript::NAME as LDT_NAME;
 
     type F = BabyBear;
     type EF = BinomialExtensionField<F, 4>;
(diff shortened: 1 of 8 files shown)


fri/src/pcs_transcript.rs
```diff
@@ -759,4 +768,122 @@ mod tests {
...
+                    ]
+                    .iter()
+                    .flat_map(|(name, pattern)| {
+                        pow_difficulties(pattern)
+                            .into_iter()
+                            .map(|(label, bits)| RecordedGrind::new(name, label, bits))
+                    })
+                    .collect();
+
+                    GrindingBudget::from_sites(&params.grinding_sites())
```
Nashtare, line 848: Good PR, and the design decisions are the right ones. Keying `GRINDING_VOCABULARY` on `(protocol, label)` rather than the label alone is necessary rather than defensive — `query_pow` really does appear in four crates and `ood_pow` in two, so label-only keying would alias silently. Keeping `check` on `&[RecordedGrind]` instead of an `InteractionPattern` genuinely avoids a `p3-security -> p3-challenger` edge, which is worth the small amount of plumbing.

One thing worth stating in the PR description more plainly than it currently is. Every caller of `GrindingBudget::check` in the tree — this one, `:883`, `batch-stark/src/transcript.rs:1287`, `batch-stark/tests/simple.rs:3304` and `:3343` — is inside a test module or an integration test. So the guarantee delivered is:

> no parameter set **that a test instantiates** reports grinding bits it does not pay for

which is narrower than "no mismatch exists today". A configuration added to `examples/`, or any downstream parameter set, is unchecked — the vocabulary table exists and would catch it, but nothing on a production path consults it. The description says an audit tool *could* call it; none does, and that is the gap between the two readings.

That is a scoping observation, not an objection — a test-only cross-check is the right first move, and it is a real improvement over the two numbers never being compared at all. I checked the interaction with #2100 as well, since it landed on `main` after this stack branched: `security/src/grinding.rs` already exists there and #2100 touched only `security/src/lib.rs`, so this extension does not collide with it.

Nashtare (approved):

fri/src/pcs_transcript.rs
tcoratger: Fair, and the description was overclaiming. I verified your call-site list rather than taking it — every caller of the budget check sits under a `#[cfg(test)]` (`fri/src/pcs_transcript.rs` at 535, `batch-stark/src/transcript.rs` at 1000) or in an integration test. **Zero production callers.** The two files where it looks like crate code are `use` lines inside the test module.

The description now states the narrow reading as the guarantee, in your words because they are the right ones:

> No parameter set **that a test instantiates** reports grinding bits it does not pay for.

and says plainly that a configuration added to `examples/` or any downstream parameter set is unchecked, that the vocabulary table would catch it but nothing on a proving or verifying path consults it, and that "an audit tool could call it" is exactly the distance between the two readings rather than a description of the state.

The `&[RecordedGrind]` signature was chosen so such a tool *can* exist without dragging a transcript dependency into this crate, and I have left it at that rather than wiring a check into a proving path in this PR — where it belongs and what it costs on a hot path is a separate decision, not something to settle in a footnote. Flagging it as the obvious next step if you want the broad reading to become true.

Thanks for checking the #2100 interaction too. That has since landed and the stack is rebased onto it: `security/src/grinding.rs` was untouched by it as you found, so this extension applies cleanly, and the composed accounting is now exercised rather than merely compiled — 3962 tests on the rebased stack, lint clean.

PR #2111, opened 2026-09-09
fix(binary-pcs): enforce claim batching soundness budget

felix314159:
## overview

a malicious prover can target a root of the alpha-batching polynomial when enough opening claims are combined. `BinaryPcsConfig` advertised `security_level` bits without accounting for the 

$$
\frac{\mathrm{num\_claims} - 1}{2^{128}}
$$

failure term, which allows proofs below the requested soundness to verify. e.g., a 124-bit configuration permits at most six claims in the tested setup. seven claims reduce soundness to at most 123 bits, but the verifier accepted them before this PR.

---

what this PR does:

* derives the maximum safe claim count from the field-error budget
* preserves the query reserve for batched-fold configurations
* rejects oversized protocols with a typed verifier error
* prevents provers from generating protocols below their configured target
* adds exact boundary tests and a test case that failed before the fix

### DIFF
--- a/binary-pcs/src/error.rs
+++ b/binary-pcs/src/error.rs
@@ -80,6 +80,17 @@ pub enum BinaryPcsError<MmcsError> {
     #[error("expected {expected} opening batches, proof carries {actual}")]
     OpeningBatchCountMismatch { expected: usize, actual: usize },
 
+    /// The opening batching challenge cannot retain the configured security target across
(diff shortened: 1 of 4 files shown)


Nashtare:
Thanks! This was already on my list and binary-pcs needs a couple additional tightenings, I'll bundle all of those in a single PR.

PR #2112, opened 2026-09-09
feat(pcs)!: add batching grinding to STIR and Circle

Nashtare:
STIR's quotient-batching checks can exhaust the challenge-field budget before its later grinding steps apply. Add a dedicated PCS grind after binding the opening claims and before sampling `alpha` and all `Combine` challenges. Credit that grind once against their joint failure probability, using the actual columns, opening points, and native degree classes to derive grouping and query schedules.

- Ensure every batched folding challenge follows the shared grind credited to it, with matching prover/verifier transcript replay.
- Preserve the standalone unground batching API and validate pooled opening multiplicities in the PCS schedule and cache.
- Add Circle PCS batching grinding, including transcript-bound difficulty and witness validation.
- Configure the examples for a shared 100-bit conjectured target, with 16-bit STIR/Circle batching grinding and FRI query counts derived from the target. STIR's reported target covers the PCS; outer STARK soundness terms remain excluded.

Breaking change: STIR and Circle transcripts and proof encodings change. Circle's encoding changes even when batching grinding is disabled. Regenerate the affected compatibility fixtures.

Validation on `ad2dfed5`, rebased onto `main` at `b4ef483d`:

- 582 tests passed across `p3-stir`, `p3-circle`, `p3-examples`, `p3-uni-stark`, `p3-batch-stark`, and `p3-security` in release mode with parallelism enabled.
- Clippy passed for STIR, Circle, examples, and uni-STARK, including all targets, with warnings denied.
- Nightly formatting and `git diff --check` passed.
- The height-18 KoalaBear example with Keccak as both objective and hasher verified successfully using STIR.


tcoratger (approved):
Approving. Adding a grind whose credit lands on `alpha` is the one kind of change where the accounting and the transcript have to agree *exactly* — credit it to a challenge the prover can resample for free and the whole budget is fiction. I checked both, on both sides, in both PCSs.

## The grind really does precede alpha

**STIR prover** (`pcs.rs:932`): every opened value is observed via `challenger.observe_algebra_slice(&ys)` in the claim loop, *then* `batch_transcript::prove`, which grinds and only then draws `alpha`. **STIR verifier** (`pcs.rs:1321`): the same claim observation loop, then `batch_transcript::verify`, which checks the witness before drawing `alpha`. **Circle** (`transcript.rs:472` / `:625`): both sides observe `OPENING_POINT` and `OPENED_VALUES` for every claim, then `observe_pow(BATCH_POW, ..)`, then `BATCH_CHALLENGE`. Symmetric, and the difficulty comes from `self.shape` on the verifier — never from the proof.

So the claims are fixed before the work is paid, and `alpha` is drawn from a post-grind state. That is what makes the credit legitimate, and the comment at the prover site states the invariant in the form that matters: *"No prover message may intervene between this site and `alpha`, or a retry could bypass the work just paid."*

**The difficulty itself is bound**, which is the other half and easy to omit. `separator::<F, EF>(bits)` puts `Length::Fixed(bits)` into the `InteractionPattern` and seeds it into the challenger, so a prover cannot claim 16 bits while grinding 0 — the verifier's separator at its own `bits` produces a different state. `examples/tests/transcript_separation.rs` pins this end-to-end with a `batch_pow_bits += 1` case in the separator-distinctness set.

**Credited once, not once per term.** `eta()` computes `algebraic_bits(..) + pow_bits`, and `algebraic_bits` has already summed every alpha class *and* the Combine term into a single figure. One addition of `pow_bits` against the joint sum, which is the correct treatment for challenges separated by no prover message.

## Every constant reproduces

I recomputed `joint_bound_adds_alpha_and_combine_at_the_shared_radius` independently at `log_degree = 8, log_inv_rate = 1, eta = 0.1`:

```text
    class (8,3)  tallest, eta used directly       ->  20480.0
    class (7,5)  eta_i = 0.6 - 0.25 = 0.35        ->  23405.714285714286  = 163840/7
    Combine      mult 2 scaled by ell - 1 = 129   ->  1320960.0
    sum                                           ->  1364845.7142857143  = 9553920/7
    ell = sum (256 - 2^h + 1) over h in {8,7}     ->  130
```

All four match your comments to the digit, and `ell = 130` is Construction 7.2's multiplicity `sum (d - d_i + 1)`, so `combine_requirement` is the paper's formula rather than an approximation of it.

The shared-radius reparametrization generalizes #2100's correctly: holding `delta = 1 - B(rho) - eta` fixed gives `eta_i = agreement - B(rho_i)`, and the code branches `B` on the assumption (`rho` for capacity, `sqrt(rho)` for Johnson) rather than hardcoding one. Evaluating each class at its *native* rate but the *shared* radius is the subtle part and it is right.

## Numerical and fail-closed discipline

- `if height == log_degree { eta }` is not a shortcut, it is exact: for the tallest class `class_rho == rho`, so `agreement - class_rho` is algebraically `eta` but loses precision when `eta << rho`. Using `eta` directly avoids catastrophic cancellation without changing the value.
- `add_error_bits` factors out the larger probability, so both `exp2` arguments are `<= 0` — no overflow, and the `f64::INFINITY` identity folds correctly on the first term.
- `field_bits` uses `order().bits() - 1` (a genuine floor on `log2 q`, not `Field::bits()`'s ceiling), and the extra Johnson bit is backed by an actual derivation — omitted-over-dominant summing to at most `3/(m + 1/2)^4 < 1` for `m >= 3`. That is a bound, not a fudge factor.
- `algebraic_bits` returns `0.` below `minimum_eta` specifically to fail closed *before* the Johnson helper's float-to-usize multiplicity can saturate on a large field. Reasoning about the failure mode of a helper you don't own is the right instinct.
- `validate()` derives the expected Combine requirement from the classes and rejects a mismatch, and enforces strictly-descending heights with the tallest equal to `log_degree` — so the shape is checked, not trusted.

On the "empty list is a commit-time probe, never an opening schedule" invariant: that is enforced structurally rather than by convention, since `classes: Option<&[_]>` is part of `StirConfigKey` via `classes.map(to_vec)`. A probe config and a schedule config are distinct cache entries, so the weaker one cannot be served to a schedule lookup. Worth the explicit doc line either way.

## Tests that actually pin the properties

`zero_bits_preserve_the_bare_batching_challenge_and_transcript` checks the **next** sample after `alpha`, not just `alpha` — so it catches a transcript-state divergence that leaves the first value coincidentally equal. That is the assertion I would have asked for.

`a_witness_meeting_only_a_lower_difficulty_is_rejected` searches for a witness that clears 1 bit but not 8 and asserts rejection at 8. A random witness would likely be rejected for the wrong reason; this proves `observe_pow` enforces the *requested* difficulty rather than any-nonzero. Together with `a_claimed_value_reaches_the_batching_challenge` on the Circle side (same point, different claimed value, challenge must move) and `both_sides_derive_the_same_challenges` at `batch_pow_bits in {0, 8}`, the ordering, the binding and the replay are each covered by a test that would fail if that specific property broke.

## Verified locally

`--features parallel` release with `+avx2`: **594 tests pass across 25 binaries** for `p3-stir`, `p3-circle`, `p3-security`, `p3-uni-stark`, `p3-batch-stark`, zero failures. Clippy `--all-targets -D warnings` clean for STIR, Circle, examples and uni-STARK; `cargo fmt --check` clean.

The breaking-change surface is documented honestly, including that Circle's encoding moves even at zero batching bits — with the witness field present and skipped by the `if bits > 0` guard, an unused value there cannot affect verification. Refreshing both compatibility fixtures rather than loosening the checks is the right call.

PR #2113, opened 2026-09-09
perf(dft,matrix): optimize inverse DFT and matrix hot paths

Nashtare:
Inverse transforms and several matrix operations perform avoidable allocations, passes over memory, or Rayon scheduling. This PR reduces that work while retaining measured fallbacks for workloads where the existing implementation is faster.

- Add a dedicated `Radix2DitParallel::idft_batch` using cached inverse twiddles and fused normalization, and forward inverse transforms through the examples' DFT enum.
- Pack `FlatMatrixView` columns directly from borrowed extension-row coefficients, eliminating per-lane flattened vectors.
- Dispatch matrix bit reversal by row width, input size, architecture, and actual Rayon backend. Preserve disjoint-slice alias information for serial wide-row swaps.
- Remove unnecessary `Copy` bounds from swap-only utilities and fix a pre-existing chunk-pointer provenance issue found by Miri, in separate commits.
- Reuse the packed field kernel without a multiworker reduction for modest column products, with conservative work limits.
- Construct zero-column padding in its final destination, retaining the large single-worker AArch64 fallback and measured serial-copy thresholds.

The ten implementation commits remain separate. The branch is rebased onto `main` at `b4ef483d`; `git range-diff` confirms all ten patches are unchanged.

Representative reductions in elapsed time from dedicated eight-core EC2 hosts, using eight workers and 15 alternating baseline/candidate trials:

| Host / build | IDFT | Flat packing | Column product | Zero padding |
| EPYC 9R45 / AVX512 | 14.7% | 87.7% | 72.0% | 69.3% |
| EPYC 9R45 / AVX2 | 20.8% | 88.1% | 84.8% | 71.5% |
| Graviton4 / SVE2 | 24.5% | 87.7% | 83.6% | 85.7% |
| Graviton4 / NEON | 24.8% | 87.8% | 84.9% | 86.2% |

These BabyBear examples use IDFT 65536×8, degree-four extension packing 1024×16, column products 32×16 with degree-four weights, and padding 65536×256 with 256 added columns. Percentages compare medians. Packing a degree-four row of width 16 reduces collected-output allocations from 82 to 2 on AVX512 and 22 to 2 on ARM; streaming uses one allocation.

Measurements compare the original implementation head `997fed51` against `4ad87320`, using Rust 1.97.0. They predate the rebase and are warm-allocator microbenchmarks, not whole-prover speedups. Supplemental two/four-worker checks found a retained ARM padding regression of 6.6%/7.4% for BabyBear 8192×33 with 33 added columns. Smaller row-permutation timing differences were sensitive to compiler loop layout, so they are not presented as reliable savings.

Validation on the implementation before rebasing:

- 392 release tests passed in each of eight configurations: serial and parallel on x86 AVX512/AVX2 and ARM SVE2/NEON, covering `p3-dft`, `p3-matrix`, `p3-util`, `p3-maybe-rayon`, and `p3-fri`.
- Strict all-target Clippy passed for all four ISA builds with the parallel feature, using Rust 1.98.0.
- Added independent inverse-DFT oracles, extension/mapped-row and packing-tail coverage, ownership/drop tests, and zero-padding edge cases.
- Focused serial Miri checks passed. Parallel Miri is qualified by existing Crossbeam epoch retag/leak diagnostics; no clean parallel Miri result is claimed.

Post-rebase validation at `8e0198ac`: the repository-required `python3 scripts/check.py full` passed on aarch64 macOS with Rust 1.98.0 and build/test concurrency capped at two. This includes 4,209 tests in each serial/parallel run (11 skipped in each), doctests in both configurations, workspace/all-target compile checking, strict all-target Clippy, dependency checks, documentation, nightly formatting, and 23 script tests. A final fetch confirmed `main` is still `b4ef483d` and is an ancestor of this branch.


matrix/src/lib.rs
```diff
@@ -626,6 +733,59 @@ mod tests {
...
+        assert_columnwise_dot_product_matches_scalar::<Goldilocks, Goldilocks3, _>(&flat);
+
+        let mapped_flat = FlatMatrixView::<Mersenne31, Mersenne31_3, _>::new(
+            patterned_extension_matrix::<Mersenne31, Mersenne31_3>(128, 3).bit_reverse_rows(),
+        );
+        assert_columnwise_dot_product_matches_scalar::<Mersenne31, Mersenne31_3, _>(&mapped_flat);
+    }
+
+    #[test]
+    #[should_panic]
```
tcoratger: We should assert on the panic message in the macro no?

matrix/src/lib.rs
```diff
@@ -626,6 +733,59 @@ mod tests {
...
+    fn test_columnwise_dot_product_rejects_short_weights() {
+        let mat = patterned_matrix::<BabyBear>(17, 17);
+        let weights = BinomialExtensionField::<BabyBear, 4>::zero_vec(16);
+        let _ = mat.columnwise_dot_product(&weights);
+    }
+
+    #[test]
+    #[should_panic]
```
tcoratger: We should assert on the panic message in the macro no?

matrix/src/lib.rs
Nashtare: Fixed in a683ee05. The short-weight test now requires `weight count must match matrix height`, with that diagnostic added to the existing length assertion. Confirmed the stricter test fails against the old generic panic; all 183 matrix tests pass in both serial and parallel configurations.

matrix/src/lib.rs
Nashtare: Fixed in a683ee05. The long-weight test now also specifies `#[should_panic(expected = "weight count must match matrix height")]`, so an unrelated panic cannot satisfy it. Both backend test runs, strict Clippy, and formatting pass.

tcoratger (approved):
Approving. The headline for me isn't the perf — it's that the Miri fix is a real memory-safety bug, and I confirmed it in both directions.

## The provenance fix is genuine UB, not a lint appeasement

I wrote a Miri test that reaches `reverse_slice_index_bits_chunks` with only 16 elements (an 8 KiB element type satisfies `size_of * n > 65536` while staying under `BIG_T_SIZE`, so Miri finishes in 7s instead of never). Against your fixed code it passes. Against the pre-PR form of that hunk, restored on your branch:

```text
error: Undefined Behavior: trying to retag from <127514> for Unique permission at
       alloc40488[0x8000], but that tag does not exist in the borrow stack
help: <127514> was created by a SharedReadWrite retag at offsets [0x8000..0xa000]
        --> util/src/lib.rs:370:20
help: <127514> was later invalidated at offsets [0x0..0x20000] by a Unique
      function-entry retag inside this call
        --> util/src/lib.rs:371:20
```

That is precisely the failure your comment describes, and it is two distinct problems in one line. The first tag spans `[0x8000..0xa000]` — **one element** — while `swap_nonoverlapping` is handed `1 << lb_chunk_size` of them, so the borrow never authorized the range it wrote. And the second `get_unchecked_mut` retags the whole allocation and kills the first tag before it is used. Deriving both pointers from a single `vals.as_mut_ptr()` fixes both at once. Splitting it into its own commit was the right call.

Relaxing `F: Copy` to `F: Send + Sync` is sound only if the permutation truly never duplicates or drops, and `test_reverse_slice_index_bits_preserves_owners` is the test that establishes it rather than assuming it: a non-`Copy`, non-`Clone` type with a `Drop` counter, asserting **zero** drops through two permutations and then exactly one per element at scope exit. That rules out duplication, leaks and double-drops in one shot. Pairing it with the `String` case covers the cache decomposition for a type whose size takes the row-swap branch.

## The IDFT is correct

`h_inv = F::PrimeSubfield::ONE.div_2exp_u64(log_h)` lifted through `from_prime_subfield` is a neat way to get `h^{-1}` — `h = 2^log_h` lives in the prime subfield, so this avoids a full extension inversion for the normalization. Valid in any odd characteristic, which two-adicity already guarantees.

Your tests compare against `NaiveDft`, which shares no code with the optimized path — the right oracle. I added a cross-implementation check on your branch as independent confirmation:

```text
    log_h 0..=10  x  widths 1,2,3,5,8,13,17,32  x  {BabyBear, Goldilocks, EF4}
      idft_batch  ==  NaiveDft == Radix2Dit == Radix2Bowers == Radix2DFTSmallBatch
      dft -> idft round trip, idft -> dft round trip, coset round trip
    -> all agree
```

Also raced 16 threads on first use of the inverse-twiddle cache in both ascending and descending `log_h` order — results stay correct, so the double-checked `entry` under the write lock holds up.

## The dispatch branches all compute the same thing

This is where a perf PR can silently go wrong, since `reverse_matrix_index_bits` now branches on width, element size, byte count, architecture *and* Rayon backend. I checked every branch against a naive reference:

```text
    widths 1,2,3,8,17,64  x  log_h 0..=15  x  {u64, u128, BabyBear, String}
      -> permutation matches, and applying twice is the identity
    columnwise_dot_product: 22 shapes straddling SERIAL_PACKED_ELEMS = 4096
      and SERIAL_PACKED_COEFF_OPS = 512, over 5 (F, EF) pairs -> matches naive
    both verified in serial AND --features parallel builds
```

`u64` at width 1 takes the new `reverse_slice_index_bits` path, `u128` is over the 8-byte cutoff, and `String` is non-`Copy` — so the type list isn't decorative, it selects different branches.

Specific things I checked rather than assumed: `use_slice = w == 1 && size_of <= 8` is sound because at width 1 the row permutation *is* the element permutation; `PARALLEL_ENABLED` only feeds `use_outline`, which is codegen-only and cannot change results; and `#[inline(never)]` on `swap_row_slices` to preserve `noalias` is a legitimate trick that touches no semantics.

## `with_zero_cols` and `vertically_packed_row`

The `self.values.is_empty()` early return is load-bearing, not defensive: without it, `chunks_exact(old_width)` would panic on a width-0 matrix, and width 0 is exactly when `values` is empty. `new_width >= 1` is guaranteed by the `num_cols == 0` early return above. The test shapes cross the copy thresholds in both directions — `16_385 x 17` and `4_097 x 129` land above 1 MiB and take the parallel branch, the small ones take the serial one — so both are actually exercised.

`vertically_packed_row`'s `c / EF::DIMENSION` and `c % EF::DIMENSION` split is the right decomposition, and testing degrees 2, 3, 4 **and** 5 against widths 3 and 17 is what would catch a `/`-vs-`%` transposition; degree 2 alone would not. Starting rows at `height + 1` exercises the wrap in `wrapping_row_slices`, and running each case through a bit-reversed inner view confirms it isn't relying on dense inner indexing.

## On the earlier review comments

`a683ee05` handled both correctly — adding `"weight count must match matrix height"` to the assertion and requiring it in both `should_panic` attributes means an unrelated panic can no longer satisfy those tests. Checking that the stricter test fails against the old generic panic is the step that makes it meaningful.

## Verified locally

`p3-dft`, `p3-matrix`, `p3-util`, `p3-maybe-rayon`, `p3-fri` pass in all four combinations of `{serial, parallel} x {scalar, +avx2}`. Clippy `--all-targets -D warnings` clean in both backends; `cargo fmt --check` clean. Ten separate commits with one concern each made this genuinely reviewable.

PR #2114, opened 2026-09-09
fix(binary-pcs)!: enforce composed opening security budgets

Nashtare:
Binary PCS previously omitted the alpha batching error from its opening budget, and the single-fold schedule bounded its worst round without composing every fold, sumcheck, and query error. This change checks the total conservative bound, rejects excessive opening claims on both prover and verifier paths, and supplies the evidence needed by security-checked multi-STARK callers.

Supersedes the closed #2111, including its opening-claim budget fix.

- Move the algebraic accounting into `p3_security::binary::BinaryPcsRegime`. Sum every fold and sumcheck round plus `(claims - 1) / 2^128`; reserve query error unless every first-batch coset is authenticated. Exhausted or unsupported budgets return zero claim capacity.
- Add `validate_opening_protocol`, `try_open`, and `try_open_at` with typed errors before transcript mutation. Reject mismatched prescribed points with verifier errors. The existing infallible trait methods retain their signatures and explicitly document their panic behavior.
- Credit grinding only to queries: the actual grind occurs after alpha and the fold challenges. Recording-challenger tests cross-check the model's credited difficulty and transcript position against real proving and verification, covering both folding schedules and zero/nonzero difficulty. These checks work independently of the unmerged #2109 vocabulary table.
- Implement `prescribed_security` with the composed opening error and unique-decoding candidate bound. Exercise checked proving/verification in the binary multi-STARK example with explicit collision evidence.
- Add `try_new_with_folding` so an exhaustive batched schedule can be selected before validating its target, including targets that the single-fold schedule cannot support.

Compatibility: previously accepted configurations or protocols can now be rejected, and derived query counts and proof transcripts can change. Infallible `open` / `open_at` callers must preflight their protocol or use the new fallible methods to handle rejection. The claim cap is a conservative guarantee, not a demonstrated attack threshold. This does not move grinding or credit it to alpha.

Validation:

- Reproduced three regressions against fresh main before implementing the fix: missing claim enforcement, single-fold overstatement, and missing prescribed-security evidence.
- Security, binary PCS, and multi-STARK suites: 307 passed and two pre-existing ignored tests in each of default and parallel modes; nine example tests also passed.
- Exact claim-cap boundaries, exhausted budgets, exhaustive-query targets, successor claims, typed prover/verifier errors, and unchanged challengers on early rejection are covered. The final verifier-error additions passed in both modes.
- The full parallel release example proved and verified 262,144 rows at a 100-bit target.
- All-target Clippy, formatting, diff checks, rustdoc link checks, and an embedded `thumbv7em-none-eabi` build passed.


binary-pcs/src/pcs.rs
```diff
@@ -292,18 +406,16 @@ where
...
         protocol: Self::OpeningProtocol,
         challenger: &mut Challenger,
     ) -> Self::Proof {
-        let evals = protocol
-            .iter_openings()
-            .map(|(table_idx, batch)| prover_data.layout.eval(table_idx, batch, challenger))
-            .collect();
-        self.finish_open(prover_data, evals, challenger)
+        self.try_open(prover_data, &protocol, challenger)
+            .expect("invalid binary PCS opening protocol")
```
tcoratger: Cosmetic, but this is the line most callers will meet during migration.

You wrote three genuinely informative error variants — `OpeningClaimCountExceedsSecurityBudget { actual, max, security_level }` even renders as "opening protocol has N claims, but at most M retain the configured S-bit opening budget" — and then `expect` throws all of it away. A caller who hits this on the infallible trait sees only "invalid binary PCS opening protocol", which doesn't distinguish a malformed shape from a budget that's over by one claim. Since the compatibility note says infallible callers must preflight or migrate, that message is their main diagnostic.

```rust
.unwrap_or_else(|e| panic!("invalid binary PCS opening protocol: {e}"))
```

I checked this compiles on your branch (both `{e}` and `{e:?}` do, despite `Mmcs::Error` only bounding `Debug` — thiserror supplies the `Display` bound). Same on line 465. Two lines, and the difference between "something's wrong" and "you're at 384 claims, the cap is 381".

tcoratger (approved):
This is the right fix and the accounting is exact. Approving — one cosmetic note inline about the migration panic.

## Every number reproduces

I recomputed `field_error_numerator` independently and all four test constants land:

```text
    n=1  rate=2 ->        11      n=10 rate=2 ->     8_214
    n=4  rate=2 ->       132      n=20 rate=2 -> 8_388_660
```

The `128 - (u128::BITS - (numerator - 1).leading_zeros())` trick is `128 - ceil(log2(numerator))`, which I verified holds for every value in `1..5000` plus the powers-of-two boundaries — `bitlen(x-1) == ceil(log2(x))` exactly, including `x = 1` where `leading_zeros(0) = 128` gives 0. Neither subtraction can underflow: `new()`'s `log_domain >= usize::BITS` rejection caps the numerator below `2^66`, and `n >= 1` keeps it at least 11.

The claim boundaries are exact, not approximate:

```text
    (1,2,1,q=200) s=124  ->  cap 6    opening_error(6)   = 124.0 bits exactly
    (4,2,2,q=200) s=119  ->  cap 381  opening_error(381) = 119.0 bits exactly
    (10,2,1,q=150) s=100 ->  cap 134_209_515
```

All three reproduce from `budget - field_num + 1` with `budget = 2^(128 - s - reserve)`. Both regimes there turn out query-exhaustive, so `reserve = 0` and the field term gets the whole budget — which is what makes those two land on the target to the bit.

## What I checked hardest

**The `checked_sub` is load-bearing.** `saturating_sub` would give `0`, and `0 + 1 = 1` would silently grant one claim on an already-blown budget. Returning `0` explicitly is the difference between fail-closed and a spurious allowance, and the doc comment calls it out by name. That is the single most likely place for this function to have been wrong.

**`field_error_numerator` is independent of `log_folding_factor`**, and `both_folding_schedules_charge_every_round` asserts `single == batched` for it. That is the whole point of charging virtual folds, and pinning it as an equality means a future edit can't reintroduce the batched discount.

**The old bound really was optimistic.** The previous `(1 << log_domain) + 1 + 2*rounds` mixed a per-round bound charged once with a union bound; the new form sums every fold. At `n=20, rate=2` that moves the reported ceiling from 105 to 103 — two bits, in the conservative direction. Reproducing the regression against fresh main before fixing it is the right order of operations.

**`opening_error`'s `next_up()` is the correct one-sided correction.** `numerator as f64` rounds to nearest, and only rounding *down* would overstate security; the `(rounded as u128) < numerator` guard catches exactly that case and leaves the (conservative) round-up alone.

**`log2_max_candidates: 0.0` is right, not a placeholder.** Unique decoding means the candidate is unique at commitment time, so `log2(1) = 0`, and `multi_stark::add_opening_evidence` accepts finite non-negative values — so this composes rather than tripping the fail-closed path.

**Claim counting matches the alpha batch.** `OpeningBatch::len() = current.len() + next.len()`, so successor columns are counted, which is what makes `(k-1)/|F|` the right degree for the alpha combination.

**`security_regime()` panics, and it is reachable from two public constructors** — so I fuzzed it: **397,422 constructor calls** across `try_new`, `try_new_with_folding` and `try_with_folding`, covering `num_variables`/`log_inv_rate`/`security_level`/`pow_bits` edge grids (0, 1, 62..65, 127..129, `usize::MAX`) plus 60k random combinations. **No panics.** `log_inv_rate = 0` was the case I expected to slip through, and it doesn't — `queries_for_target` returns 0 at rate one, so `ZeroQueries` fires before `BinaryPcsRegime::new` can return `None`. Separately, **25,461 configs that actually build** all satisfy `security_level <= max_security_bits()`, `max_opening_claims() >= 1`, and `opening_error(cap).bits() >= security_level`.

## The transcript cross-check

`actual_grinding_matches_the_model_and_follows_alpha_and_every_fold` is the best test in the PR, because it validates the model's central assumption against the real transcript rather than against prose. The negative controls are what give it teeth: an `overcredited` regime at `pow_bits + 1` is rejected, and `early.swap(1, 10)` — moving the grind before alpha at identical difficulty — is also rejected. Without those two, the positive assertion would pass against almost any model. Checking both prover and verifier event streams across two folding schedules and `pow_bits in {0, 4}` covers the combinations that matter.

## Enforcement is on both sides

`validate_opening_protocol` runs in `try_open`, `try_open_at` *and* `verify_opening`, and it reads only the verifier-supplied protocol — no proof data, no transcript — so it rejects before anything is consumed. The dimension check binds the protocol to the config via `log2_ceil(total) == num_variables` with `checked_shl`/`checked_mul`/`checked_add` throughout, so an overflowing shape reports `InvalidOpeningProtocol` rather than wrapping into something plausible.

Replacing `verify_at`'s `assert_eq!(protocol.num_openings(), points.len())` with `validate_points` returning `OpeningPointShapeMismatch` matters more than it looks: `verify_at` takes attacker-influenced input, so that was a panic on the verifying path. Good catch.

## Verified locally

`+avx2,+pclmulqdq` release: 55 binary-PCS + 27 + 114 security + multi-STARK tests pass, and the parallel feature run is clean too. Clippy `--all-targets -D warnings` clean across all three crates; `cargo fmt --check` clean.

PR #2115, opened 2026-09-09
ci: trim the build matrix and scope the Rust cache to main

Nashtare:
## Motivation

The Plonky3 org is on the free GitHub plan: **20 concurrent jobs org-wide, 5 concurrent macOS**. A single CI run was **21 jobs**, so one PR saturated the org's entire concurrency budget and PRs were effectively serialized.

Separately, the Actions cache was over quota and thrashing:

```
active_caches_size_in_bytes: 10,631,657,109   (limit: 10 GB)
refs/heads/main       24 caches
refs/pull/2100/merge  11
refs/pull/2097/merge   6
refs/pull/2114/merge   2
```

`Swatinem/rust-cache` had no `save-if`, so every PR branch saved its own ~300 MB cache per matrix leg. Those pushed the repository past the 10 GB limit, GitHub evicted LRU — which is `main`'s entries, the ones every PR restores from — and legs rebuilt cold. That is why a `Test` step took 11–15 minutes.

## Changes

**Matrix: 18 legs → 11 (21 jobs → 14).** Legs are selected by an explicit `name` and a `modes` list rather than a boolean `parallel` axis, so a leg can run both configurations as sequential steps sharing one warm target directory.

| Leg | Runs |
| --- | --- |
| `linux` | serial |
| `linux, parallel` | parallel |
| `linux, avx2` | serial |
| `linux, avx2, parallel` | parallel |
| `linux, pclmulqdq` | serial + parallel |
| `macos, parallel` | parallel, plus the keccak `-sha3` check |
| `windows, parallel` | parallel |
| `linux, avx512f, build only` | both builds |
| `linux, avx2+vpclmulqdq, build only` | both builds |
| `linux, avx512f+vpclmulqdq, build only` | both builds |
| `arm64, sve2, build only` | both builds |

The serial/parallel split exercises `maybe-rayon`, whose two backends do not vary by operating system, so it stays on the Linux legs; macOS and Windows run the `parallel` configuration consumers ship. The only coverage removed is macOS-serial and Windows-serial — the merged legs run the same commands as before, in fewer jobs.

**Cache: `save-if: ${{ github.ref == 'refs/heads/main' }}`** on all six `rust-cache` uses in `ci.yml` and `ci-heavy.yml`. Pull requests restore from `main` and no longer save.

## Measured impact

Per-job durations from run `34334938358`:

| | Before | After |
| --- | --- | --- |
| Jobs per run | 21 | 14 |
| Job-minutes | ~143 | ~105 (est.) |
| Concurrent macOS slots | 2 | 1 |
| Concurrent Windows slots | 2 | 1 |
| Active caches | 43 (10.6 GB) | ~24 (~7 GB) |

The cache change is expected to account for more of the wall-clock win than the matrix trim, since it is what makes the test steps rebuild from cold.

## Required status checks must be updated with this merge

Job names changed, so the 15 current required contexts become 11. Preserving the existing policy that the `pclmulqdq` and `vpclmulqdq` legs are not gating:

```
Formatting and Clippy
Build embedded
Test wasm32 simd128
Build and Test (linux)
Build and Test (linux, parallel)
Build and Test (linux, avx2)
Build and Test (linux, avx2, parallel)
Build and Test (macos, parallel)
Build and Test (windows, parallel)
Build and Test (linux, avx512f, build only)
Build and Test (arm64, sve2, build only)
```

Worth considering separately: `linux, pclmulqdq` is now a single job and is the only leg that executes carry-less multiply rather than merely compiling it, so promoting it to required is cheap.

**Every open PR needs a rebase once the contexts are switched**, since a PR still running the old `ci.yml` reports the old job names. With ~20 PRs open, this is worth timing deliberately.

## Test plan

- [x] `ci.yml` and `ci-heavy.yml` parse as YAML
- [x] Matrix expansion simulated leg-by-leg against the step guards; every command that ran before still runs, except macOS-serial and Windows-serial
- [x] `actionlint` clean on `ci.yml`, `ci-heavy.yml`, `bench-smoke.yml`, `pr-labeler.yml`
- [ ] CI on this PR reports the 11 new job names

### DIFF
--- a/.github/workflows/ci-heavy.yml
+++ b/.github/workflows/ci-heavy.yml
@@ -47,6 +47,8 @@ jobs:
           echo "RUSTFLAGS=$RUSTFLAGS -C target-feature=${{ matrix.features }}" >> "$GITHUB_ENV"
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Test WHIR exhaustive sweep
         run: python3 scripts/check.py whir-exhaustive
@@ -70,6 +72,8 @@ jobs:
         run: echo "RUSTFLAGS=$RUSTFLAGS -C target-feature=+pclmulqdq" >> "$GITHUB_ENV"
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Test the 2^16 round trip
         run: python3 scripts/check.py binary-large
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,45 +24,53 @@ env:
 
 jobs:
   build_and_test:
-    name: Build and Test
+    name: Build and Test (${{ matrix.name }})
     strategy:
       matrix:
         include:
-          # Default and `parallel`-feature runs are split into separate legs
-          # so they execute as concurrent jobs instead of sequential steps.
-          - os: ubuntu-latest
-            parallel: false
-          - os: ubuntu-latest
-            parallel: true
-          - os: ubuntu-latest
+          # `modes` selects which of the default and `parallel`-feature runs a leg
+          # performs. A leg naming both runs them as sequential steps sharing one
+          # warm target directory.
+          - name: linux
+            os: ubuntu-latest
+            modes: serial
+          - name: linux, parallel
+            os: ubuntu-latest
+            modes: parallel
+          - name: linux, avx2
+            os: ubuntu-latest
             features: "+avx2"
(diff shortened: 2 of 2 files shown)


Nashtare:
@tcoratger, with the swarm of PRs we're getting a very high pressure on CI ressources (both time, concurrency and cost), as we've seen our PRs are hanging forever with runners taken by other branches and not being able to pick up CI jobs. This should hopefully reduce the contention, reduce CI time and cost, while preserving all the most important jobs by mostly getting rid of redundant variants in the matrix.

Nashtare:
We'll need rebasing branches once this lands and I'll update the required job names.

tcoratger (approved):

PR #2116, opened 2026-09-09
Optimize Goldilocks and wasm-simd128 primitives

Nashtare:
## Summary

Adds wasm32 and wasm-simd128 optimizations for Goldilocks arithmetic and the main hash/permutation primitives.

- Specializes scalar Goldilocks multiplication, squaring, powers, and packed SIMD arithmetic.
- Adds delayed-reduction packed MDS for widths 8, 12, and 16, plus packed Poseidon1/Poseidon2 improvements.
- Adds two-lane Keccak and four-message SHA-256 SIMD paths.
- Adds an opt-in BLAKE3 wasm SIMD feature and a reproducible WASI benchmark example.
- Adds correctness oracles for non-canonical representatives, delayed reductions, extension arithmetic, and batch-boundary lengths.

## Validation

- Native: 1,133 affected field/hash tests passed.
- WASI SIMD: Goldilocks 219, Keccak 6, SHA-256 8 passed.
- WASI plain: Goldilocks 185, Keccak 3, SHA-256 8 passed.
- Browser-target checks, formatting, and clippy checks passed.
- Benchmarked clean builds on AMD EPYC 9R45 and Graviton4 hosts with Wasmtime 45.0.2 and Node/V8 26.3.1.

Representative SIMD time reductions were 17–44% for scalar Goldilocks multiply/square, 3–51% for 256-term linear combinations, 41–66% for packed MDS/Poseidon1, 3–37% for Keccak, and 24–33% for SHA-256. BLAKE3 SIMD remains opt-in because it regresses short inputs while improving large inputs.


tcoratger (approved):
Approving. The risk profile here is very favourable: essentially everything is behind `cfg(all(target_arch = "wasm32", target_feature = "simd128"))`, and the two changes that aren't are provable no-ops on native. What remained to check was the hand-rolled limb arithmetic, and it is exact.

## The wasm limb routines are exactly right

`mul_wide` and `square_wide` are pure integer logic, so target-independent — I reimplemented both in an independent model (exact bignum, explicit `u64` wrapping) and compared against true `a*b` / `a*a`, while asserting that *every* intermediate stays within `u64`:

```text
    17x17 edge grid (0, 1, 2^32 +/- 1, 2^33 +/- 1, 2^63 +/- 1, P-1, P, P+1, u64::MAX-1, u64::MAX)
    400_000 uniform random pairs
    200_000 near-u64::MAX pairs   (the tightest carry path)
    64x3 powers-of-two boundary family x every edge
    -> 0 mismatches, no intermediate overflow
```

That last assertion is the one that matters, because your comments make specific claims and they hold at the limit rather than with slack. `hi = hh + (t0 >> 32) + (t1 >> 32)` reaches `2^64 - 1` exactly in the worst case — one more term and it would wrap. Same for `square_wide`: `middle = lh + (ll >> 33)` is safe because `ll >> 33 < 2^31` against `lh <= (2^32-1)^2`, precisely as the comment says.

The `square_wide` carry extraction is the subtle one and it is correct: `(ll + lh*2^33) >> 64` really does equal `(lh + (ll >> 33)) >> 31`, because writing `ll = 2^33 q + r` gives `t*2^33 + r < 2^64` for every `t < 2^31` and `r < 2^33`, so the truncated division composes without loss. Sharing the cross product to get three `i64.mul`s instead of four is a real win on a target with no `__multi3`.

Making the test check **both product limbs** rather than the field residue is the right call, and the comment says why — a lost carry in the high limb can be masked by `reduce128`. Including `P`, `P+1` and `u64::MAX` covers noncanonical representatives, which is where a `& NEG_ORDER` mask assumption would show up.

## Native behaviour is unchanged, and provably so

- `Algebra::quadratic_extension_square`'s default body is byte-for-byte the code it replaces in `binomial_square`'s `D == 2` arm. Pure extraction of an override point.
- The only override in the tree is `wasm32_simd128/packing.rs:269`, so nothing outside wasm sees a different path.
- `mul_wide` on non-wasm is literally `u128::from(x) * u128::from(y)`, so `Mul` and both `dot_product` arms are unchanged.
- `Goldilocks::square` is overridden only under `target_arch = "wasm32"`.

The override itself matches the trait contract: `[a0^2 + w*a1^2, 2*a0*a1]`, with `mul_2exp_u64(3) - square1` computing `8x - x = 7x`. I checked `W = Self::new(7)` in `goldilocks/src/extension.rs:25` — canonical, so the fast path does fire, and the comment is accurate. Worth noting the branch is only a *selector*: a noncanonical `w` holding `7 + P` falls to `square1 * w`, which is still correct, just slower. Correctness never depends on the constant.

## Test coverage

The SHA-256 known-answer vectors are genuine — I verified all four against `hashlib`:

```text
    abc -> ba7816bf...15ad    abe -> d81a65c1...3ca8
    abd -> a52d159f...49c9    abf -> 431b36f2...26a9
```

And `LENGTHS = [55, 56, 63, 64, 65, 119, 120, 127, 128]` is exactly the right set: 55 is the last length whose padding fits one block and 56 forces a second, so those two straddle the case a batched implementation is most likely to get wrong. Pairing them with `COUNTS` up to 17 exercises the four-lane tail at every residue. Good too that these tests are *not* wasm-gated — they validate the default trait methods natively and the SIMD ones on wasm, from the same assertions.

`simd_lanes_do_not_influence_each_other` is the Keccak test I would have asked for: varying lane 1 while holding lane 0 fixed and asserting lane 0's output is unchanged catches cross-lane contamination that a match-the-scalar test can miss when both lanes carry related data. The zero/all-ones case plus a `proptest` over arbitrary 25-word states rounds it out.

`hash_many_with_empty_output_does_not_validate_input_shape` and the two `should_panic` tests pin the contract at the boundaries rather than leaving it implicit.

## Verified locally

Native suites for `p3-goldilocks`, `p3-field`, `p3-keccak`, `p3-sha256`, `p3-mds`, `p3-poseidon2`, `p3-poseidon1` all pass in scalar, `+avx2`, and `+avx512f,dq,vl,bw`. Clippy `--all-targets -D warnings` clean in all three modes across those crates plus `p3-blake3`; `cargo fmt --check` clean. `cargo check -p p3-blake3 --features wasm32-simd` builds, and the feature is correctly opt-in rather than default — right call given it regresses short inputs. The `blake3` floor bump to `1.8.4` is the minimum that carries `wasm32_simd` and stays caret-compatible.

I could not execute the wasm paths (no WASI target or runtime on this host), so for the SIMD MDS/Poseidon kernels I am relying on your CI's `Test wasm32 simd128` job, which is green along with the other 14 checks. The target-independent arithmetic underneath them is what I verified directly.

One trivial thing, no action needed unless you care: the title is the only one in this batch not in conventional-commit form — `perf(goldilocks,wasm): ...` would match the rest of the history and the labeler.

PR #2117, opened 2026-09-09
fix(transcript): address the review follow-ups from the typed-transcript stack

tcoratger:
Addresses every non-blocking follow-up from the reviews of the typed-transcript stack (#2081, #2102, #2103, #2105, #2106, #2107, #2109), plus the one piece the hiding-sumcheck review turned up as genuinely missing.

The stack's branches were deleted at merge before these landed, so none of it is in `main` today. One PR rather than several: only the simulator is new functionality, and it turned out to be additive with no soundness question attached.

## The one behaviour change

The multi-STARK verifier deferred **both** opening errors to reach the driver's completeness check. So a proof with a bad main opening still executed the preprocessed verification, against attacker-supplied `proof.preprocessed_opening`, with a challenger in a state no honest run produces — roughly doubling verifier work on the rejecting path for any batch with preprocessed columns.

The main opening now releases the driver and returns, exactly as the lookup and zerocheck arms already did. The verdict is unchanged.

Pinned by a counting configuration that records every hand-out of the preprocessed scheme. Restoring the old flow:

```
assertion `left == right` failed
  left: 1
 right: 0
```

## Two contracts, written where an implementor reads them

Declining to hoist the main-commitment absorb out of the commitment scheme rested on a property the trait never stated: **a commit phase absorbs exactly the commitment it returns, and does nothing else.** An out-of-tree implementation that absorbed a table height or drew a batching challenge would compile, satisfy the pattern on both sides, and desync as an unexplained completeness failure.

That obligation is now on the multilinear commit method. It is also on the layout commit default body, which is where the guarantee actually lives for two of the three in-tree implementations — a `Layout` override could have broken both PCSs without either PCS file changing.

I considered adding a `CanObserve<Self::Commitment>` bound and **declined**, having measured it: 34 errors, the decisive one being that `MultiStarkProof`'s derived serde would have to prove a challenger can observe a commitment. And the prescribed-point opening sub-trait already requires it where the bracket is actually played. A bound cannot express "one absorb and nothing else" anyway.

## A witness-free simulator for the prelude that ships

The hiding sumcheck has two preludes — one derives its target from recorded claims, the other inherits it — and only the first had a simulator. WHIR plays the second, at every batch.

Reading eprint 2026/391 settles which is canonical, and it is the opposite of what the code's coverage suggests: Construction 6.3's target is **explicit public input** to the sub-protocol, and Definition 5.8 defines it as the claimed sum plus the auxiliary mask total — exactly the value the inherited prelude binds. So the inherited path *is* the paper's construction, and the recorded-claims one is the layer added on top.

The argument carries over unchanged. The auxiliary term is constant in `X` and independent of the masks, so it lands in the affine offset and never touches the linear map whose rank the proof computes. Its per-round carry shares slot 0 with the plain constant and is summed **before** masking, so it introduces no new independent linear functional of the witness — one uniform mask coefficient blinds an arbitrary constant on that slot however many witness terms compose it.

The verifier already split its prelude from its shared body; the simulator now does the same. The new entry point takes the claimed sum as a parameter and needs **no verifier argument at all**, which is itself the evidence that the target is public.

Tests cover it at a live auxiliary claim, which nothing did before — the honest side already exercised a nonzero carry, but the simulator never met one. Worth noting that the auxiliary term is not dead code: WHIR passes zero at the first call site and a nonzero running mask-claim total at every subsequent one.

One thing the falsification found, and it changed the tests: at zero grinding difficulty the verifier cannot reject a well-shaped record at all, so acceptance alone is a toothless assertion. The driver now also requires the verifier's redrawn challenge stream to equal the one its counterpart produced. Feeding it the wrong claim fails four tests with that message; drawing wires from the base-field subspace instead of uniformly fails five.

## Three divergences from the paper, recorded rather than rediscovered

- **The mask length bound is deliberately stronger** than the paper's. The round polynomial here is a product of two multilinears, hence quadratic in `X`; at the paper's bound the mask has no coefficient for the quadratic slot and it would ship unmasked. Verified directly: at the lower bound the quadratic slot equals the unmasked plain value exactly.
- **The auxiliary carry's halving is the normalisation consistent with Definition 5.8.** Adding a constant `AUX * 2^{-k}` at every point of the `k`-cube gives round-`j` partial sum `AUX * 2^{-j}` and cube total `AUX`. Checked for `k` in `1..=6`.
- **Compiled hiding is computational, not perfect.** The lemma is proved for an IOR with *oracle* masks whose queried positions a simulator answers. This code commits the whole mask codeword under one root, which information-theoretically determines the masks. A comment asserted perfect indistinguishability of the commits; it now says what holds. Standard IOR-to-argument compilation, stated as scope rather than alarm.

## Accuracy

Six comments claimed properties the code does not have. The two worth naming:

- *"the sponge is dropped on the error path and never reused"* — it is a `&mut` borrowed from the caller, handed back mutated mid-round. What keeps the verdict safe is that every caller propagates; a retry would let a prover pick its own challenges. The comment now says so, because it was inviting exactly that change.
- *"a step either side skips or reorders is rejected"* — a driver is a purely local self-check with no visibility into the other party. What rejects a prover that skips an absorb is sponge divergence; what rejects a skipped bracket is the callee's own verification.

A FRI test is renamed to what it actually pins: its old name promised the delegation markers absorb nothing, but if they did, both sides would absorb identically and every assertion would still pass.

Two shape sweeps now destructure without a rest pattern, so a new field must be confronted. Honest about the limit: a new field trips the missing-field error at the struct literals **first**, which masks the pattern error, so what is enforced is that the author comes through the function — not that a row appears. An instance-permutation property that was structurally sound but unasserted is now asserted, and it holds.

The transcript-failure error enums become `#[non_exhaustive]` as a family of eight, matching the one that already was. The verification-error enums are deliberately left: adding the attribute there also compiles clean, so it is an API-policy decision about enums downstream code legitimately matches to route a rejection, not a review follow-up.

Also corrected: a fixture count (six across four pipelines, not three), a stale error doc, and a note in the WHIR test suite that cited simulator tests as its hiding evidence when they exercised a prelude WHIR never plays.

## Verification

3980 tests pass, `scripts/check.py lint` clean, no `.postcard` fixture changed — this alters no transcript and no wire format.

### DIFF
--- a/fri/src/pcs_transcript.rs
+++ b/fri/src/pcs_transcript.rs
@@ -507,6 +507,7 @@ where
 
 /// A transcript step the proof failed to satisfy.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Error)]
+#[non_exhaustive]
 pub enum PcsTranscriptFailure {
     /// One opening carries a claimed-evaluation count the run never described.
     #[error("opening {opening}: claimed evaluation count mismatch: expected {expected}, got {got}")]
@@ -745,38 +746,44 @@ mod tests {
     }
 
     #[test]
-    fn the_delegation_bracket_leaves_the_sponge_untouched() {
-        // Invariant: the markers are structural.
-        // They record the delegation, and absorb nothing.
+    fn both_sides_walk_the_delegation_bracket_identically() {
+        // Invariant: the two sides come through the bracket in the same place.
         //
-        // Fixture state: the two sides over the same shape, each bracketing an empty delegation.
+        //     alpha        ->  the batching challenge agrees across the delegation
+        //     next sample  ->  the sponges are still in step once the bracket closes
+        //
+        // Fixture state: one shape, driven once by each side, each bracketing an empty run.
+        //
+        // The shape describes the bracket unconditionally, so neither side can skip it.
+        // An absorbing marker would move both sponges alike, so it is outside what this pins.
         let shape = shape_with(vec![vec![vec![1]]]);
         let claims: Vec<CommitmentWithOpeningPoints<EF, (), ()>> =
             vec![((), vec![((), vec![(EF::ONE, vec![EF::ONE])])]).into()];
 
-        let mut bracketed_challenger = fresh_challenger();
-        let mut bracketed =
-            PcsVerifierTranscript::<Ch, F, EF>::new(&mut bracketed_challenger, shape.clone());
-        bracketed.claimed_openings(&claims).unwrap();
-        let bracketed_alpha = bracketed.batch_phase(None).unwrap();
-        bracketed.delegate(|_| ());
-        bracketed.finish();
+        let mut verifier_challenger = fresh_challenger();
+        let mut verifier =
+            PcsVerifierTranscript::<Ch, F, EF>::new(&mut verifier_challenger, shape.clone());
+        verifier.claimed_openings(&claims).unwrap();
+        let verifier_alpha = verifier.batch_phase(None).unwrap();
+        verifier.delegate(|_| ());
+        verifier.finish();
 
         // The prover side plays the same steps and must land on the same challenge.
-        let mut plain_challenger = fresh_challenger();
-        let mut plain = PcsProverTranscript::<Ch, F, EF>::new(&mut plain_challenger, shape);
-        plain.claimed_openings(&vec![vec![vec![vec![EF::ONE]]]]);
-        let (plain_alpha, witness) = plain.batch_phase();
-        plain.delegate(|_| ());
-        plain.finish();
+        let mut prover_challenger = fresh_challenger();
+        let mut prover = PcsProverTranscript::<Ch, F, EF>::new(&mut prover_challenger, shape);
+        prover.claimed_openings(&vec![vec![vec![vec![EF::ONE]]]]);
+        let (prover_alpha, witness) = prover.batch_phase();
+        prover.delegate(|_| ());
+        prover.finish();
 
-        assert_eq!(bracketed_alpha, plain_alpha);
+        assert_eq!(verifier_alpha, prover_alpha);
         assert_eq!(witness, None);
         // Both sponges advanced identically, so they still agree on what comes next.
-        let bracketed_next: F = bracketed_challenger.sample();
-        let plain_next: F = plain_challenger.sample();
-        assert_eq!(bracketed_next, plain_next);
+        let verifier_next: F = verifier_challenger.sample();
+        let prover_next: F = prover_challenger.sample();
+        assert_eq!(verifier_next, prover_next);
     }
+
     /// This protocol's name, as the vocabulary table keys it.
     fn protocol() -> &'static str {
         from_utf8(NAME).expect("the protocol name is ASCII")
--- a/multi-stark/src/transcript.rs
+++ b/multi-stark/src/transcript.rs
@@ -122,6 +122,8 @@ struct PreprocessedOpening;
 ///
 /// The pair absorbs nothing.
 /// What it records is that a sub-protocol runs at this position, under its own seed.
+///
+/// The driver's opener and closer append this pair to their pattern record and touch nothing else, so a marker has no path to the sponge.
 fn delegation<T: ?Sized>(label: Label) -> [Interaction; 2] {
     [
         Interaction::marker::<T>(Hierarchy::Begin, Kind::Protocol, label),
@@ -682,6 +684,7 @@ where
 
 /// A statement-level transcript step the batch failed to satisfy.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Error)]
+#[non_exhaustive]
 pub enum MultiStarkTranscriptFailure {
     /// The described preprocessed-commitment step arrived with no commitment.
     #[error("{tables} preprocessed table(s) described but the verifying key carries no commitment")]
@@ -717,6 +720,7 @@ mod tests {
     extern crate std;
 
     use alloc::vec;
+    use core::ops::Range;
 
     use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
     use p3_challenger::DuplexChallenger;
@@ -763,6 +767,18 @@ mod tests {
         }
     }
 
+    /// Whether two shapes seed a sponge identically, compared on the preimage not a sample.
+    ///
+    /// The preimage is the pattern fingerprint followed by the instance label.
+    /// Comparing it rather than one sampled element keeps the comparison exact.
+    fn seeds_agree(left: &MultiStarkShape, right: &MultiStarkShape) -> bool {
+        let left = left.domain_separator::<F>();
+        let right = right.domain_separator::<F>();
+
+        left.pattern().pattern_hash() == right.pattern().pattern_hash()
+            && left.instance_label() == right.instance_label()
+    }
+
     /// First challenge a shape's seed produces on a fresh sponge.
     fn first_challenge(shape: &MultiStarkShape) -> F {
         let mut challenger = fresh_challenger();
@@ -772,32 +788,53 @@ mod tests {
 
     /// Every field of the shape, each moved one step away from `base_shape`.
     ///
-    /// One entry per configuration knob, so a knob added without a binding shows up here.
+    /// The batch and one of its instances are destructured below with no rest pattern.
+    /// Both are destructured, since a new knob can land on either struct.
+    /// Every binding then feeds the row that covers it, so an unread field is a build failure.
+    ///
+    /// The enforcement is weaker than it looks.
+    /// A new field trips the missing-field error at the baseline's struct literals first.
+    /// What is enforced is that the author must come through here, not that a row appears.
     fn one_step_from_base() -> Vec<(&'static str, MultiStarkShape)> {
+        let MultiStarkShape {
+            instances,
+            pow_bits,
+        } = base_shape();
+        let num_instances = instances.len();
+        let MultiStarkInstanceShape {
+            num_variables,
+            main_width,
+            preprocessed_width,
+            num_public_values,
+        } = instances
+            .into_iter()
+            .next()
+            .expect("the baseline batch is not empty");
+
         let mut mutations = Vec::new();
 
         let mut shape = base_shape();
-        shape.pow_bits += 1;
+        shape.pow_bits = pow_bits + 1;
         mutations.push(("pow_bits", shape));
 
         let mut shape = base_shape();
-        shape.instances.pop();
+        shape.instances.truncate(num_instances - 1);
         mutations.push(("instances.len", shape));
 
         let mut shape = base_shape();
-        shape.instances[0].num_variables += 1;
+        shape.instances[0].num_variables = num_variables + 1;
         mutations.push(("instance.num_variables", shape));
 
         let mut shape = base_shape();
-        shape.instances[0].main_width += 1;
+        shape.instances[0].main_width = main_width + 1;
         mutations.push(("instance.main_width", shape));
 
         let mut shape = base_shape();
-        shape.instances[0].preprocessed_width += 1;
+        shape.instances[0].preprocessed_width = preprocessed_width + 1;
         mutations.push(("instance.preprocessed_width", shape));
 
         let mut shape = base_shape();
-        shape.instances[0].num_public_values += 1;
+        shape.instances[0].num_public_values = num_public_values + 1;
         mutations.push(("instance.num_public_values", shape));
 
         mutations
@@ -886,8 +923,6 @@ mod tests {
         // Baseline: the two-instance shape, seeded and sampled once.
         let baseline = first_challenge(&base_shape());
 
-        // Each mutation moves exactly one field one step.
-        //
         // A field the fingerprint covers moves the seed through the step sequence.
         // A field it does not must move it through the instance label instead.
         for (field, shape) in one_step_from_base() {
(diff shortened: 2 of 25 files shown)


multi-stark/src/verifier.rs
```diff
@@ -231,26 +239,34 @@ where
...
             &instances.opening_protocol(),
             &instances.main_points(&reduction.point),
             challenger,
         )
-    });
+    }) {
+        Ok(evals) => evals,
+        // Nothing below can change this verdict, so the preprocessed opening never runs.
+        Err(error) => {
+            transcript.abort();
```
Nashtare, line 268: Verified verdict-preserving: on `main` the main-opening error was already the first `?` after `finish()`, so no proof changes outcome here — only the cost of the rejecting path. Three things I checked before believing that:

- `VerifierState::abort` → `PatternPlayer::abort` just sets `finalized = true`, and `Drop` panics only on `!is_finalized()`, so skipping `finish()` is safe and `abort` is idempotent — same shape as the lookup/zerocheck arms at 218 and 237.
- Nothing between here and `finish()` feeds the returned value, so the preprocessed bracket has no side effect observable on this path.
- The brackets are pattern-only (`VerifierState::begin`/`end` call `player.interact` and never touch the challenger), so no sponge state is skipped that a later step would need.

multi-stark/tests/whir_preprocessed.rs
```diff
@@ -550,3 +551,130 @@ fn verify_rejects_tampered_preprocessed_opening() {
...
+    // Baseline: an untouched proof verifies, and its preprocessed opening runs once.
+    verify(
+        &config,
+        VerifierInstances::new(vec![VerifierInstance::new(&air, &vk, log_height, &[])]),
+        &proof,
+        0,
+        &mut challenger(),
+    )
+    .expect("honest preprocessed proof must verify");
+    assert_eq!(config.take_preprocessed_lookups(), 1);
```
Nashtare, line 656: This baseline assertion does more than set up the `== 0` check below: it pins that `verify` reaches `preprocessed_pcs()` exactly once. `multi-stark/src/security.rs:319` is a second call site in the crate today, so if a security check is ever folded into the verify path this line fails first and points straight at it. Good choice of probe — the counter is self-guarding in both directions.

sumcheck/src/zk/simulator.rs
```diff
@@ -17,7 +32,124 @@ use super::prover::common::{mask_endpoints, sample_masks};
...
+/// - `encoding`: mask code the masks are drawn from and encoded under.
+/// - `mmcs`: commitment scheme carrying the interleaved mask oracle.
+/// - `rng`: source of the mask messages and of every wire coordinate.
+///
+/// # Returns
+///
+/// - Simulated proof record.
+/// - The batch mask commitment.
+/// - Per-round challenges.
+fn simulate_claim<F, EF, Enc, M, Challenger, R>(
```
Nashtare, line 54: `shape` duplicates the shape `transcript` was constructed with, and nothing here checks the two agree — a caller passing a different one would drive `num_rounds` / `wire_len()` / `pow_bits` off the parameter while the driver validates against its own copy. Both call sites are correct and a mismatch would surface as a pattern failure rather than silently, so this is a nit; a `shape()` accessor on `ZkProverTranscript` would remove the parameter and the question with it.

sumcheck/src/zk/simulator.rs
```diff
@@ -729,4 +899,402 @@ mod tests {
...
+        }
+
+        // === Extension-valued wires on both sides ===
+        //
+        // The masks are extension-valued, so every wire coordinate is uniform over EF.
+        //
+        // The index-`>= 2` coordinates are mask-only.
+        // An `F`-valued mask would pin them to the base field and reintroduce the witness leak.
+        for wire in &zk_data_real.round_coefficients {
+            for &c in wire.iter().skip(2) {
```
Nashtare, line 1070: This check is vacuous at `ell_zk == 3`: `wire_len = max(3, 3) - 1 = 2`, so `skip(2)` yields nothing. The proptest below draws `ell_zk in 3..=5`, so roughly a third of its 16 cases assert nothing here, and `the_inherited_simulator_runs_against_a_non_zero_auxiliary_claim` only escapes it by hardcoding `ell_zk = 4`. Invariant 5 in the driver's docstring reads as unconditional. Narrowing the proptest to `4..=5`, or asserting `wire.len() > 2` before the loop, would make the coverage match the claim.

sumcheck/src/zk/mod.rs
```diff
@@ -87,6 +132,6 @@ pub use data::{
     mask_residual_covectors, mask_residual_covectors_from_shape,
 };
 pub use prover::{ZkLayout, ZkPrefixProver, ZkProver, ZkSuffixProver, stack_codewords};
-pub use simulator::simulate_classic_unpacked;
+pub use simulator::{simulate_classic_unpacked, simulate_classic_unpacked_claim};
```
Nashtare, line 154: Following the PR's own thesis one step further: after this change the *recorded-claims* prelude has no production consumer in-tree at all. `ZkVerifier::into_sumcheck` is reached only from `sumcheck/src/zk/test_helpers.rs:297` and its own test module, and `ZkLayout` / `ZkProver` only from `whir/benches/sumcheck.rs`. WHIR's zk prover calls `into_zk_sumcheck` (inherited) at both sites — `whir/src/pcs/zk/prover/mod.rs:202` with `EF::ZERO` and `:443` with `masks.aux` — and its verifier calls `verify_claim`.

So `simulate_classic_unpacked` is now a simulator for a prelude nothing ships, maintained alongside the one that matters. Fine to keep as library surface, but if the batching prelude is not on a roadmap it is a removal candidate rather than double maintenance — worth deciding explicitly rather than by default.

multi-stark/src/transcript.rs
```diff
@@ -1278,17 +1310,40 @@ mod tests {
...
+            //
+            //     public-value counts  ->  step lengths, so the fingerprint holds their order
+            //     arities and widths   ->  instance label, written in the same order
             //
-            // The seed preimage is the fingerprint followed by the instance label.
-            // Comparing the preimage rather than one sampled element keeps this exact.
-            let left_seed = left.domain_separator::<F>();
-            let right_seed = right.domain_separator::<F>();
+            // Every rotation and the reversal are tried, which is every order of a two-instance batch.
+            let rotations = (1..shape.instances.len()).map(|shift| {
```
Nashtare, line 1347: `arbitrary_shape` draws `vec(instance, 1..4)`, so a single-instance shape makes this body vacuous: `rotations` is empty and `reversed == shape` hits the `continue` below. A `prop_assume!(shape.instances.len() >= 2)`, or generating `2..4` for this property, would keep every case doing work.

fri/src/pcs_transcript.rs
```diff
@@ -745,38 +746,44 @@ mod tests {
     }
 
     #[test]
-    fn the_delegation_bracket_leaves_the_sponge_untouched() {
-        // Invariant: the markers are structural.
-        // They record the delegation, and absorb nothing.
+    fn both_sides_walk_the_delegation_bracket_identically() {
```
Nashtare, line 749: Agreed the old name over-promised. One loose end the rename leaves: `multi-stark/src/verifier.rs` now states "The brackets absorb nothing" as a load-bearing step of its rejection argument, and after this rename nothing pins it. It does hold — `VerifierState::begin`/`end` only call `player.interact(Interaction::marker::<T>(..))` and never reach the challenger — so it is structurally guaranteed rather than tested, which is arguably the stronger position. A pointer to that from the multi-stark doc would close the loop.

Nashtare (approved):

multi-stark/src/verifier.rs
tcoratger: Thanks for checking the three properties rather than taking the abort on faith — the idempotence of `abort` and the fact that nothing between there and the completeness check feeds the returned value are exactly what make the early return safe, and they were not obvious from the diff alone.

multi-stark/tests/whir_preprocessed.rs
tcoratger: That second call site is a better observation than the reason I picked the counter. I chose it to make the `== 0` check meaningful; you are right that the baseline `== 1` is independently load-bearing — fold a security check into the verify path and this line fails first and names the cause, rather than the `== 0` check silently starting to measure something else.

sumcheck/src/zk/simulator.rs
tcoratger: Fixed, and the same redundancy was on the verifier side, which I would not have looked for without this.

`ZkProverTranscript` and `ZkVerifierTranscript` both gained a `shape()` accessor, and the parameter is gone from the shared simulator body **and** from `ZkVerifier::replay_claim` — that one carried the identical duplication, driving `num_rounds` and `pow_bits` off a parameter while its driver validated against its own copy. Both verify entry points now build the shape exactly once, into the driver.

Nothing else in the crate reconstructs a shape it could read instead: the shape validator necessarily runs before a driver exists, since it gates seeding, and the prover paths build theirs once and hand it straight to the constructor. Added surface is two `const` accessors.

sumcheck/src/zk/simulator.rs
tcoratger: You were right, and it was worse than the one driver you flagged: **the same hole was already in the tree on the batching path.** `run_acceptance_and_mask_prelude_coupling` carries the identical `skip(2)` check on both the real-prover and simulator wires, and both its proptests also drew `3..=5`. So this was pre-existing, and the new driver inherited it by copying the shape of a test that could not fail either.

I took your option (a), narrowing the draws, and rejected (b) on its own terms: `ell_zk == 3` is legal — the shape validator accepts it and `wire_len = max(3,3) - 1 = 2` — so a `wire.len() > 2` guard would make the driver fail a legal configuration, which is the "make it look illegal" outcome to avoid. Narrowing also matches precedent already here: `prop_simulator_invariants` drew `4..=6` with that exact rationale in a comment.

All three proptests now draw `4..=5`. To keep the legal minimum covered rather than merely excluded, one new test exercises **both** drivers at `ell_zk = 3` — acceptance, the redrawn stream and the matched-seed coupling — and narrates that the mask-only band is empty at that width, so the stratification check legitimately reads nothing there. Both driver docstrings now state the band width as a function of `ell_zk` instead of asserting the invariant unconditionally.

**Falsification, since the whole point is that these could not fail.** Collapsing the mask-only band to zero in the simulator's wires, with the old `3..=5` draws pinned to 3:

```
prop_simulator_invariants ................................ FAILED
prop_inherited_simulator_accepts_and_couples_mask_prelude . ok
prop_simulator_accepts_and_couples_mask_prelude_rs_prefix . ok
prop_simulator_accepts_and_couples_mask_prelude_rs_suffix . ok
```

Three of four pass with the band fully destroyed. With the narrowed draws, the same mutation fails 9 of the 10 tests in the module — and the one that still passes is the `ell_zk = 3` floor test, where the band is genuinely empty. That is the suite distinguishing the two cases instead of conflating them.

sumcheck/src/zk/mod.rs
tcoratger: Verified your call-site claim in full before deciding, and it holds:

- the recorded-claims verify entry point is reached only from this crate's tests, its test helpers, and the layout prover's own tests — no production consumer, in or out of crate;
- the recorded-claims layout prover is reached only from a WHIR bench and this crate's test helpers;
- WHIR's zk prover calls the inherited entry point at both sites, and its verifier calls the inherited verify.

**Decided to keep, and recorded rather than defaulted** — which I take to be what you were asking for. Retiring the batching prelude means retiring the public entry points it simulates, and that is an API decision rather than a change inside this module, so it should not happen as a side effect of a follow-up PR.

`sumcheck/src/zk/mod.rs` now carries a scope section: the two ways a batch reaches its target, which one the composed pipeline plays on both sides, that the other is this crate's standalone surface with no in-tree pipeline playing it and is supported on that basis, and that both carry a witness-free simulator so neither ships unverified. Plus the actionable line — a third prelude owes a third simulator, and retiring this one is an API call.

If you would rather it go, that is a clean separate PR now that the decision is written down instead of implicit.

multi-stark/src/transcript.rs
tcoratger: Fixed, and I took the dedicated strategy over `prop_assume!` for the reason you imply: a rejected case is a case that did no work, whereas a narrower strategy spends every case on a batch that actually has a non-identity permutation.

Extracted a size-parameterised strategy. The general one keeps `1..4`, unchanged for the properties that legitimately want single-instance batches; a reorderable one draws `2..4` and is used only by this property. Nothing changed globally.

**Falsification.** Making the instance label order-insensitive by sorting the instances in the separator:

- with a single-instance draw: `permuting_the_instances_of_a_batch_moves_the_seed ... ok` — passing with order-sensitivity completely destroyed;
- with the reorderable strategy: fails, naming the two shapes that share a seed.

fri/src/pcs_transcript.rs
tcoratger: Closed the loop, and agreed that structural is the stronger position — so no test.

Confirmed it at the layer: a driver's opener and closer, on both the prover and verifier side, each consist of exactly one statement appending a marker to that driver's own pattern record. Neither mentions the challenger, and the protocol-bracket helpers are thin wrappers over them. So a marker has no path to the sponge, by construction rather than by convention.

The multi-STARK comment now says *why* rather than asserting it, and the same one-line reason sits on the marker-pair helper where a reader would go to check:

```
/// Why the brackets absorb nothing: a driver's opener and closer only append a marker
/// to its own pattern record, and neither reaches the sponge.
///
/// That holds by construction on both sides rather than by test, so a skipped
/// delegation is caught by the callee and never here.
```

PR #2118, opened 2026-09-09
perf(ci): speed up slow test-suite tests

Nashtare:
## Description

Set CARGO_PROFILE_TEST_OPT_LEVEL=1 workspace-wide in ci.yml, mirroring the opt-level override ci-heavy.yml already uses for its exhaustive sweeps. Debug assertions and overflow checks stay on; only arithmetic-heavy correctness tests (full proofs, large polynomial sweeps) get faster.

Also:
- stir: drop test_pcs_proof_size_vs_binary_fri_equivalent_input, which ran the exact same proof generation as assert_stir_proof_smaller_than_binary_fri with no assertion of its own.
- sumcheck, util: cap two proptest suites that were still on the untuned 256-case default despite expensive per-case work, matching the proptest_config overrides already used elsewhere in the workspace.
- uni-stark: give the two stir_fibonacci prove/verify smoke tests a lower security_level config, separate from the one backing the committed compat-fixture regression test.


Nashtare:
@tcoratger let's see how long it takes before merging but hopefully this should shave off a couple minutes from the usual runs (which became quite heavy with the last batch of PRs that landed)

Nashtare:
@tcoratger perfs seem pretty good now!

tcoratger (approved):
Great!

uni-stark/tests/stir_fibonacci.rs
```diff
@@ -133,10 +133,32 @@ const STIR_FIXTURE: &str = "tests/fixtures/uni_stark_stir_v0_8_0.postcard";
...
+    let challenger = Challenger::new(perm);
+    MyConfig::new(pcs, challenger)
 }
 
-fn make_config_with_batch_pow(bits: usize) -> MyConfig {
+/// A lower-security config for tests that only exercise the prove/verify
+/// round trip, not the exact security parameters. `security_level` drives
+/// STIR's query count, so a smoke test that just needs a valid proof can
+/// use a much smaller value than `make_config`'s 100-bit target.
+fn make_fast_config() -> MyConfig {
```
tcoratger, line 161: This one buys nothing, and I measured it four ways before saying so. All at the same commit, same machine, `make_config` vs `make_fast_config`:

```text
                                              before    after
  whole binary        @ OPT_LEVEL=1            2.38 s   2.34 s
  whole binary        @ default profile       44.93 s  45.14 s
  test_public_value   @ OPT_LEVEL=1            2.33 s   2.32 s
  test_public_value   @ default profile       44.76 s  44.75 s
  test_short_public_values_rejected @ OPT=1    2.32 s   2.35 s
```

Every pair is inside noise, and `45.14` is nominally *worse* than `44.93`. The entire 19x came from `CARGO_PROFILE_TEST_OPT_LEVEL=1`; dropping `security_level` from 100 to 32 contributes nothing on top.

That is a bit surprising until you look at the shapes — `test_short_public_values_rejected` uses `1 << 3` rows and `test_public_value` isn't much bigger. At that size the runtime is Poseidon2 constant generation and DFT setup, not query count, and STIR's schedule is likely already at its floor, so there is no query work for a lower target to remove. At `OPT_LEVEL=1` the whole six-test binary and a single test both come in at ~2.3s, which says the remaining time is essentially fixed overhead.

So the net effect is 26 added lines that duplicate `make_config` and lower what these two round trips actually exercise from a production-like 100-bit target to 32 bits, for no measurable gain. Deleting `make_fast_config` and putting `make_config()` back in both call sites is strictly better: fewer lines, identical runtime, realistic parameters, and one fewer config to keep in step with `make_config` when STIR's parameters next change.

Not blocking — the profile change stands on its own and I'm merging. Happy to send the revert as a follow-up if you'd rather not re-push, and equally happy to be told you want the lower target there for a reason the timings don't show.

tcoratger (approved):
Approving — the profile change is the whole win and it's safe. One measured note inline: the `make_fast_config` part buys nothing, so it's 26 lines of coverage reduction for no gain.

## The opt-level change keeps the checks that matter

"Debug assertions and overflow checks stay on" is the load-bearing claim, so I tested it rather than reasoning about it. `CARGO_PROFILE_TEST_OPT_LEVEL` sets only `opt-level`, leaving the test profile's inherited `debug-assertions` and `overflow-checks` untouched — confirmed on this branch:

```text
                              default profile   OPT_LEVEL=1
    u64::MAX + 1 panics             yes            yes
    1u64 << 64 panics               yes            yes
    cfg!(debug_assertions)          true           true
    debug_assert! fires             yes            yes
```

That matters more here than it might elsewhere. The unguarded-shift class I flagged on #2064 — `fri.rs:170`, `proximity.rs:23` computing `1u64 << (log_trace_length + log_blowup)` — is caught **only** by overflow checks in debug; in release it silently wraps to `n = 1`. If this change had quietly dropped them, CI would have stopped catching that whole family. It doesn't.

And `ci-heavy.yml:28` already carries the identical setting, so this really is bringing `ci.yml` in line rather than inventing a policy.

**The measured win is large.** On `p3-uni-stark --test stir_fibonacci`, the worst offender:

```text
    default profile   44.93 s
    OPT_LEVEL=1        2.38 s      ~19x
```

Which matches what I saw in #2100's CI log earlier today, where that binary reported `SLOW [> 60.000s]` and individual passes at `96.047s` and `96.559s`.

## The dropped STIR test really is a duplicate

Checked rather than taken on trust: `assert_stir_proof_smaller_than_binary_fri` iterates the **identical** `[(14, 2), (16, 2)]` with the same `WIDTH = 3` and calls the same `compare_stir_proof_size_with_binary_fri` — the only difference being that it asserts on the result the deleted test discarded. The helper's own internal assertions (`conjectured_soundness_bits`, domain size and shift equality) still run through the retained test. So this is a strict halving of duplicated proof generation with zero coverage loss.

## On the proptest caps

Worth being explicit that these are a real reduction in fuzz depth, not a free win: `ci-heavy` runs exhaustive WHIR sweeps and a large binary-PCS round trip, but it does not raise `PROPTEST_CASES`, so nothing else compensates. For `dims()` specifically, 256 -> 64 over its 12-way `prop_oneof!` takes each branch from roughly 21 samples to 5.

I think it's still the right trade, and for a reason visible in the strategy itself: `dims()` puts its interesting cases in `Just(...)` singletons and narrow ranges — `(TILE - 1)..=(TILE + 1)`, `Just((TILE * 4, TILE * 4))`, exactly `MEDIUM_LEN` — which is the design that tolerates fewer cases, since the boundaries are hit deterministically rather than stumbled upon. And 32/64 sit squarely inside the ~12 existing `with_cases` overrides in the workspace (16 through 256, including both these values). Fine as-is; just not free, and the PR body reads slightly more like tidying than like a trade.

tcoratger:
Merged before I could re-review — flagging one thing about the final shape, since my approval was against `3f0aa131` (the workspace-wide `CARGO_PROFILE_TEST_OPT_LEVEL` version) and the approach changed substantially after that.

For the record, the diagnosis in `139d4e54` is a genuinely good catch and better than what I approved: I had verified that `OPT_LEVEL=1` preserves `debug-assertions` and `overflow-checks` and measured 44.9s -> 2.4s on `stir_fibonacci`, but I never checked what it does to *compile* time across the whole workspace, and the bn254 -> zkhash -> ark-ff/halo2 const-generic tree is exactly the kind of thing that would eat the win. Landing the `#[ignore]` + per-package Heavy CI approach instead is the right call.

**The one gap worth a follow-up:** `ci-heavy.yml`'s `paths:` filter lists the crates the 8 tests *live in*, but not the crates they *depend on*. There's no `field/`, `dft/`, `matrix/`, `util/`, `merkle-tree/`, `challenger/`, `poseidon2/`, `commit/`, or any of the concrete field crates.

So for a PR touching only those, the 8 tests run **neither** in main CI (now `#[ignore]`d) nor in Heavy CI (paths don't match). `scripts/check.py full` doesn't reach them either — it calls `test_commands`, not `slow-regressions`, so a local full check skips them too.

That inverts the coverage where it's most valuable. Today's batch is the concrete case: #2113 touched only `dft/` and `matrix/`, #2098 and #2116 only `field/` and `goldilocks/`. A full STIR-backed uni-STARK prove/verify is precisely the signal that catches a subtle DFT twiddle or field-reduction bug — it shows up in a complete proof and nowhere else — and those are exactly the PRs that would now skip it.

Bounded, not severe: the nightly cron still runs them, so a regression surfaces on main within a day rather than never. But adding the shared dependency crates to `paths:` is nearly free and closes the part that matters:

```yaml
      - "field/src/**"
      - "dft/src/**"
      - "matrix/src/**"
      - "util/src/**"
      - "merkle-tree/src/**"
      - "challenger/src/**"
      - "commit/src/**"
```

Happy to send that as a follow-up PR if you'd like — or if you'd rather keep the filter narrow and lean on the nightly, that's a defensible call too, just worth being deliberate about rather than inheriting it from which crates the test files happen to sit in.

PR #2120, opened 2026-09-10
review: couple tweaks here and there

Nashtare:
Addressing a bunch of tiny / minor points from the bunch of PRs that got merged in the past few days that didn't really deserved spending time individually on. Includes a couple of API breaking changes as we're going to release v0.8.0 anyway.

Bundle everything into 1 PR instead.


Nashtare:
Oops closed it by mistake

tcoratger (approved):
Deep review of the whole bundle (158 files, ~4.9k/1.7k). I traced every fix back to the invariant it is protecting and checked prover/verifier symmetry on each one. **No blockers, no majors — approving.**

## What I verified

**BIN-1 — final codeword bound before query grinding.** The last fold word travels in the clear, so without absorbing it the prover could pick it after seeing the query indices. Prover (`open_queries`) and verifier both `observe_slice(final_codeword)` immediately before `grind` / `check_witness`, at the same transcript position. The two zero-claim mutation tests were correctly relaxed to accept rejection at *any* of the three now-reachable gates.

**R-1 / R-9 — zero-difficulty witness canonicality.** `grind_generic` returning `F::ZERO` without touching the sponge is consistent with `check_witness*(0, _)` reading nothing, so both sides stay aligned. The malleability that opens up is closed everywhere it matters: FRI already pinned all three phases, `binary-pcs` has `check_canonical_pow_witness`, STIR carries `Option<Val>`, HVZK's `transcript.pow` requires zero at `bits == 0`, and this PR adds the missing Circle `batch_pow_witness` check. That family is now complete.

**CM-2 / SS-2 — quotient sizing.** `max(inferred, hint)` is strictly safer than trusting the hint, and moving the check out of `debug_assert!` removes a release-only footgun. `StarkAirParams::from_air` now mirrors `get_log_num_quotient_chunks` exactly: `1 << log2_ceil(c-1) == (c-1).next_power_of_two()`, and the `<< is_zk` doubling is reproduced by `checked_mul(1 << zk)`. Taking the trace domain lets the Circle branch use `degree_multiple_with_transition`, which is what the prover does. The `(1, 2, 49) / (3, 8, 145) / (5, 16, 273)` table in `zk_ood_budget_preserves_degree_hint_chunk_boundaries` pins the DEEP-ALI degree against the committed chunk count rather than against `max_constraint_degree`, which was the actual gap.

**FSF-3 — hiding random-opening widths.** Previously `point.values.extend(rand_point)` accepted whatever length arrived, so a prover could shift the opened-value vector. Requiring `rand_point.len() == expected`, with `expected == 0` only for the round named by *trusted* metadata, is the right shape. I checked the index plumbing end to end: `StarkOpeningLayout::new(ZK).preprocessed == is_zk + 2` matches the push order in `prepare_opening_claims`, and in batch-STARK `preprocessed_index` is captured before the round is pushed and the push is unconditional inside `if let Some(global)`. The `[None, Some(1), Some(2)]` rejection sweep plus the challenger-unchanged assertions cover it.

**MS-1 / MS-4 — successor columns and nonzero GKR coordinates.** Undeclared successor reads silently became zero in the verifier's folder; `validate_successor_columns` walks base, extension, global and local interaction expressions and rejects them. Binding the ordered column lists into the seed is the right companion. Rejection-sampling the round and branch coordinates away from zero is correct given the zerocheck divides by them, `challenge_extensions_rejecting` consumes exactly one `Length::Fixed(1)` step regardless of retries so both sides stay in lockstep, and `security/src/multilinear.rs` was updated to `nonzero_field_bits` to match.

**SW-4 — empty STIR batch.** This was a genuine prover/verifier asymmetry: the verifier returned at `b == 0` before constructing its player, the prover constructed one and absorbed the domain separator. Returning early on both sides fixes it, and hoisting the `sources.len() != b` check above the early return is right.

**TypeTag / `AlgebraIdentity`.** `(modulus, degree)` alone could not separate two degree-4 extensions over the same prime with different defining polynomials. The `AlternateBasis<0..3>` test is exactly the right shape — including the positive case that two distinct Rust types with identical algebra identities *share* the protocol. `binomial_algebra_id` binding `W` per-`D` covers the Goldilocks 2/5 and Mersenne 3 / complex 2/3 instantiations.

**R-2 — typed proving errors.** The `Result` conversion is mechanical but the placement is careful: every `?` after a live transcript is preceded by `inspect_err(|_| transcript.abort())`, and the trace commit in `uni-stark/src/prover.rs` correctly has none because it runs before line 236 where the transcript is built. multi-STARK's clone-and-publish-on-success is a real improvement, with the RNG caveat honestly documented.

Also confirmed empirically that the new STIR `prepare_open` pre-check cannot change behaviour for zero-point matrices — `stir/src/pcs.rs:840` already rejects those outright — and that the new class/count derivation (`width * points.len()`, descending native heights) reproduces the old `num_reduced`-derived one exactly.

## Minor follow-ups

None block this. I'll send them as a follow-up PR:

- `multi-stark/src/zerocheck/mod.rs` — `visit_leaves` memoises by raw pointer, and the local-interaction `count` path deliberately uses a fresh `BTreeSet` because the cloned root is a temporary. The `Arc` children are shared and stable, so hoisting the owned roots into a `Vec` that outlives the scan lets the shared `seen` cover them too; as written a DAG-shaped count expression is re-traversed exponentially.
- `fri/src/two_adic_pcs.rs` — `transcript.finish()` runs before `fri_proof?`. Safe today because `prove_fri` validates heights before describing anything, but an error raised later would panic in `finish()` instead of returning. Making the error arm explicit removes the trap.
- `security/src/shape.rs` — `air.max_constraint_degree().unwrap_or(max).max(max)`: the `unwrap_or` arm is redundant.
- `multi-stark/src/fractional_gkr/materialize.rs` — the scalar path is checked against an analytic reference, which is good, but there is no direct differential test against the packed path at a height where both are available. Easy to add now that the dispatch is by plan.
- `examples/src/proofs.rs` — `trace.height().ilog2()` floors silently; `log2_strict_usize` states the invariant.

## Two questions, neither blocking

1. After CM-2, `BaseAir::max_constraint_degree` can no longer skip symbolic evaluation on either side — the verifier now always runs `get_max_constraint_degree_extension`. The hint's only remaining effect is to *inflate* the quotient. Is it worth keeping under that name, or should it become an explicit "oversize the quotient" knob?
2. `Interaction::algebra` recomputes `Keccak256Hash.hash_iter(A::algebra_id())` — plus a `Vec<u8>`, often through `format!` — for every interaction constructed. Pattern building is O(steps) and steps scale with query counts. Once-per-type caching would be nicer, though `no_std` + generics makes it awkward; happy to leave it if you have measured it as noise.

PR #2121, opened 2026-09-10
fix(stir)!: seed the batching phase at every difficulty

tcoratger:
## What this fixes

The PCS batching phase bypassed its transcript driver whenever the grinding difficulty was zero:

```rust
if bits == 0 {
    return (challenger.sample_algebra_element(), None);
}
```

Both sides agreed on that path, so no test failed. But no seed was absorbed, which means the challenge that folds every opening claim of a batch into one carried neither the protocol identifier nor the pattern fingerprint.

This is the same objection that was raised against an optional shape binding during #1603, and resolved there by folding the fingerprint into the seed unconditionally, so that a driver cannot exist without one. The zero-bit path was the one place in STIR where that no longer held.

## What the seed does and does not bind — corrected

An earlier version of this description claimed "two batch configurations differing only in something the pattern would have bound were indistinguishable". That was too broad, and the review was right to narrow it.

The only input to the describing function is the difficulty. The opening claims are absorbed on the bare challenger, outside this driver. So what the seed gains here is:

```text
    protocol identity    ->  bound
    grinding difficulty  ->  bound, inside the step
    the batch's shape    ->  NOT bound here
```

Concretely, two instances differing only in `max_log_height_spread` — different bucket layout, same claims — still land on the same seed and the same batching challenge after this change. That is not a soundness hole, since a layout mismatch is caught elsewhere, but it is not separation either.

FRI and Circle do bind their claim layout, describing one step per opening claim carrying its width. STIR's claim layout is bound by #2123, which sits above this PR in the stack and expresses the per-commitment and per-matrix nesting as pattern containers. **This PR binds the identity and the difficulty; that one binds the shape.**

## The change

Seed unconditionally, and make the grinding *step* conditional inside the pattern rather than making the *driver* conditional on the difficulty:

```text
    difficulty 0  ->  [ challenge ]
    difficulty b  ->  [ grind at b bits, challenge ]
```

A zero difficulty now contributes no step instead of bypassing the driver. Every site gets a seed, and the difficulty still reaches the fingerprint through the step that describes it. This matches how the generic-degree sumcheck already handles its own optional per-round grinding.

## Why now, and what it reverses

This reverses a deliberate scope decision from #2112, whose description says it kept the zero-bit transcript unchanged. That trade bought compatibility with proofs that the rest of the typed-transcript migration breaks anyway, and in exchange it left unprotected the one draw the layer exists to protect. Since this PR is the base of a stack of breaking transcript changes, it is the cheapest moment to take it back.

## Grinding accounting

Now that the zero-bit convention is pinned, this site matches the `p3-fri-pcs` row of the grinding vocabulary exactly — batch-combination, elided at zero. The PCS already credited `batch_combination`, so the credited half existed and the recorded half is now stable; only the vocabulary row was missing.

Added, per review: the `p3-stir-pcs-batch` row, its doc-table entry, and the two tests that mirror the FRI PCS ones — the vocabulary mapping, and a budget check over zero and two positive difficulties so both sides of the elided-at-zero convention are exercised.

## Consequences of the transcript change

- The STIR compatibility fixture is regenerated.
- The truncated-openings shape test no longer pins its row count. The honest count is a property of the drawn query set, so every transcript change moved it by hand — twice already. It is now read off the proof before the truncation, which keeps the assertion exact and stops the next transcript change from needing a hand-edited constant.

## Tests

- An ungrounded site now seeds its phase: the challenge it draws differs from a bare draw on the same sponge, and both sides still recover it.
- Prover and verifier agree at difficulty 0 and 8.
- Every difficulty lands on its own seed, checked pairwise across 0, 1, 8 and 20.
- A witness where none is demanded, and a missing witness where one is, are both rejected before the sponge moves.
- The recorded difficulty matches what the security model credits, at zero and at two positive difficulties.
- The vocabulary maps this protocol's one grind, and only one row.
- The existing lower-difficulty rejection case is kept.

## Verification

| Command | Result |
| --- | --- |
| `cargo test -p p3-stir -p p3-security -p p3-uni-stark -p p3-fri` | 24 suites passed, 0 failed |
| `cargo clippy -p p3-stir -p p3-security --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |

## Stack

Base of the stack completing the typed Fiat-Shamir migration: #2121 → #2122 → #2124 → #2125 → #2126.


stir/src/batch_transcript.rs
```diff
@@ -14,36 +38,108 @@ use p3_field::{ExtensionField, PrimeField64};
...
+        ALPHA,
+        Length::Scalar,
+    ));
+
+    // A seed exists for every difficulty, zero included.
+    //
+    // No site draws from a bare sponge.
+    DomainSeparator::new(
+        1,
+        b"p3-stir-pcs-batch",
```
Nashtare: Now that the zero-bit convention here is pinned, this site matches the `p3-fri-pcs` row of `GRINDING_VOCABULARY` exactly (`BatchCombination`, `Elided`) — but `security/src/grinding.rs` still has no `("p3-stir-pcs-batch", "batch_pow")` row, and its doc table's `p3-stir` line lists only the four in-protocol grinds. `TwoAdicStirPcs::grinding_sites()` already credits `batch_combination` (stir/src/pcs.rs:420), so the credited half exists and the recorded half is now stable. Adding the row would let you mirror `the_grinding_vocabulary_maps_the_one_grind_this_protocol_describes` and the `GrindingBudget::check` test from fri/src/pcs_transcript.rs:798 and :849 — which is what would catch a future drift between the bits credited and the step described. Non-blocking, but this PR is the moment it becomes possible.

stir/src/batch_transcript.rs
```diff
@@ -1,9 +1,33 @@
...
+//! # Shape
+//!
+//! ```text
+//!     difficulty 0  ->  [ challenge ]
+//!     difficulty b  ->  [ grind at b bits, challenge ]
+//! ```
+//!
+//! # What the shape binds
+//!
+//! A fingerprint of the description enters the sponge before either step runs.
```
Nashtare, line 20: Worth narrowing in the PR description: the only input to `separator` is `bits`, and the opening claims are absorbed on the bare challenger outside this driver (stir/src/pcs.rs:918). What the fingerprint gains here is the protocol identity and the difficulty — not the batch's shape. FRI and Circle describe one step per opening claim carrying its width (fri/src/pcs_transcript.rs:348, circle/src/transcript.rs:255), so their batching challenge is bound to the claim layout; STIR's is not. Concretely, two instances differing only in `max_log_height_spread` — different bucket layout, same claims — still land on the same seed and the same `alpha` after this change. Not a soundness hole (a layout mismatch is caught elsewhere), but "two batch configurations differing only in something the pattern would have bound" stays true for everything except the difficulty itself.

stir/src/batch_transcript.rs
```diff
@@ -14,36 +38,108 @@ use p3_field::{ExtensionField, PrimeField64};
 
 use crate::error::{ProofShapeError, StirError};
 
+/// Step label of the grinding that guards the batching challenge.
 const BATCH_POW: &str = "batch_pow";
+
+/// Step label of the batching challenge itself.
 const ALPHA: &str = "alpha";
 
+/// Describe and seed the batching phase at one difficulty.
```
Nashtare: `separator` only describes — the seeding happens in `ProverState::new` / `VerifierState::new`. "Describe the batching phase at one difficulty" would match what the function actually does.

stir/src/batch_transcript.rs
```diff
@@ -1,9 +1,33 @@
...
+//!
+//! # What the shape binds
+//!
+//! A fingerprint of the description enters the sponge before either step runs.
+//!
+//! The difficulty travels inside the grinding step, so the fingerprint covers it.
+//!
+//! Two sites at different difficulties therefore cannot share a transcript.
+//!
+//! An ungrounded site shares one with neither.
```
Nashtare: "shares one with neither" — with neither of what? The antecedent is the two differing-difficulty sites of the previous line, which the reader has to reconstruct. Something like "An ungrounded site shares one with no ground site either" reads more directly.

stir/src/batch_transcript.rs
```diff
@@ -54,32 +150,46 @@ where
...
     state
         .finalize()
         .expect("the batching phase reads an empty wire");
+
     Ok(alpha)
 }
 
 #[cfg(test)]
 mod tests {
+
```
Nashtare, line 198: Stray blank line after `mod tests {`.

stir/tests/stir.rs
```diff
@@ -3035,8 +3035,8 @@ mod babybear_pcs {
             ProofShapeError::InputOpenedRowCount {
                 log_height: log_d + 1,
                 commitment: 0,
-                expected: 19,
-                got: 18,
+                expected: 18,
```
Nashtare: Checked the explanation and it holds: `query_positions` sorts and dedups (stir/src/pcs.rs:2137) and `n_q` is `positions.len()` (stir/src/pcs.rs:1636), so this is a collision in the new stream rather than a weakened check. It is the second time these constants move with the transcript, though — deriving them instead (assert `expected == got + 1`, or read the honest `opened_values.len()` before the `pop()`) would keep the check exact and stop the next transcript change from needing hand-edited test expectations.

Nashtare (approved):

stir/src/batch_transcript.rs
tcoratger: Done — added in 353a2b7b.

The `p3-stir-pcs-batch` row now sits next to its batch-combination sibling in `GRINDING_VOCABULARY`, with the doc-table entry alongside it, and both tests are mirrored:

- `the_grinding_vocabulary_maps_the_one_grind_this_protocol_describes` — asserts the site and the elided convention, and that this protocol contributes exactly one row.
- `the_recorded_difficulty_matches_the_one_the_security_model_credits` — a `GrindingBudget::check` over difficulties 0, 1 and 10, so both sides of the elided-at-zero convention are exercised. It also asserts the recorded step count is `usize::from(bits > 0)`, which is the convention itself stated as an equality.

I also lifted the protocol name into a `NAME` constant, so the vocabulary row and the seed reference one spelling rather than two.

You were right that this PR is the moment it became possible — the check is only meaningful once the recorded half is stable at zero.

stir/src/batch_transcript.rs
tcoratger: Agreed, and the PR description is corrected.

You are right that the only input to the describing function is the difficulty, and that the opening claims are absorbed on the bare challenger outside this driver. So what the seed gains here is the protocol identity and the difficulty, not the batch shape:

```text
    protocol identity    ->  bound
    grinding difficulty  ->  bound, inside the step
    the batch shape      ->  NOT bound here
```

Your `max_log_height_spread` example is the right one: different bucket layout, same claims, same seed, same alpha after this change.

The claim layout is bound one level up in the stack, in #2123, which expresses the per-commitment and per-matrix nesting as pattern containers rather than flattening it. That PR has a test taking three groupings that all flatten to the same widths, showing the nested fingerprints are pairwise distinct and the flattened ones identical — which is the property your comment says STIR was missing relative to FRI and Circle.

So the split is: this PR binds identity and difficulty, #2123 binds the shape. I have said so explicitly in the description rather than leaving the broader claim standing.

stir/src/batch_transcript.rs
tcoratger: Fixed — the doc now reads "Describe the batching phase at one difficulty."

You are right that the function only builds a separator; seeding happens when a driver is constructed from it. I also corrected a comment in the body that had the same slip, so it now says a separator exists for every difficulty and both drivers seed from it.

stir/src/batch_transcript.rs
tcoratger: Fixed — now reads "An ungrounded site shares one with no ground site either."

That names the antecedent instead of making the reader carry it down from the previous line.

stir/src/batch_transcript.rs
tcoratger: Fixed.

stir/tests/stir.rs
tcoratger: Done — derived rather than pinned.

I took the second of your two suggestions, reading the honest count off the proof before the truncation:

```rust
honest_rows = Some(opening.opened_values.len());
opening.opened_values.pop();
...
expected: honest_rows,
got: honest_rows - 1,
```

That keeps the assertion exact — still the specific variant, still the specific height and commitment index, still off by exactly one — while removing the constant that had to move by hand.

Thanks for independently checking the dedup explanation. You are right that it moved twice; it also moved a third time once the PCS-layer migration stacked on top, which is what convinced me the derived form is the correct one rather than a convenience. I applied the same treatment to the sibling test in #2123, where an overstated native height can be caught by either of two shape checks depending on how many draws coincided — that one now matches either variant and asserts the off-by-one, since which check wins the race is a property of the challenge stream rather than of the mutation.

PR #2122, opened 2026-09-10
feat(sumcheck)!: drive the ring-switching transcript through the typed Fiat-Shamir layer

tcoratger:
> Rebased onto `main`, which now carries #2120 and #2127. The diff is against `main`.
>
> **This PR also carries #2124.** #2124 was squash-merged into this branch, so what
> lands here is the ring switching + STIR PCS work below *plus* the layout sumcheck
> transcript. The review threads on both PRs are addressed and resolved.
>
> Follow-up review fixes in this branch: the batching phase is registered in the
> cross-protocol suite now that #2120 made `batch_domain_separator` public
> (`NUM_PROTOCOLS` 16 -> 20), and the in-crate stand-in test it forced is dropped.

## Why this was left out

Ring switching landed in #2006, after the typed-transcript migration plan was written, so it was never in scope and has no typed transcript at all. Its prover and verifier absorbed and drew straight off the sponge, and the two halves mirrored each other only because someone kept them in step by hand.

## The description

One module now holds the single statement of this reduction's transcript, played by both sides:

```text
    evaluation point   ->  one extension element per coordinate    (message)
    tensor element     ->  one base element per coefficient        (message)
    batching point     ->  one extension element per packed var    (challenge)
    sumcheck           ->  begin / end delegation bracket
    surviving claim    ->  one extension element                   (message)
```

The ordering of sponge operations is unchanged from before, so this is purely a transcript-format break: the seed and the challenge derivation both move.

## The two type tags

The point step and the coefficient step carry different type tags — the point is over the extension, the coefficients over the base field. Neither list can stand in for the other whatever its length, which is the property that makes a base-field coefficient vector and an extension point non-interchangeable in the sponge.

## Where the widths come from (R1)

None of the three widths is read from proof data:

- The coordinate count is the caller's own point, on both sides.
- The tensor width is the extension degree squared, a function of the field pair alone.
- The batching width is the base-two logarithm of that degree, likewise.

The coefficient count carried in the proof is compared against the described width before anything is absorbed, and the pre-existing structural guard still fires ahead of that with its own error.

## Delegation (R2)

The inner sumcheck is already on the typed layer and seeds its own driver, so it stays a bracketed sub-protocol: one driver per invocation, with a recorded begin and end around it, matching the multi-STARK zerocheck idiom.

## One addition to `p3-challenger`

A fixed-length extension-challenge draw, on both driver sides. It is the missing dual of the fixed-length extension absorb: the batching point is an unconstrained list of challenges, and the only existing way to express that was the rejecting draw with a predicate that always accepts. Two tests ship alongside it, and its first consumer is in this same commit.

**Reviewer's option:** if you would rather not grow the shared crate, the rejecting draw with an always-true predicate is functionally identical and records the same fixed length. Say so and the challenger half of this diff can be dropped entirely.

## One ordering change worth flagging

The column check now runs after the batching point is drawn, where it previously ran before. This does not move the sponge: the check is a purely local computation over the proof's own columns. Keeping the three transcript steps atomic in one call is the point — it removes the opportunity for a caller to interleave anything between the two absorbs and the draw that depends on them. The structural guard still runs first, before the transcript exists.

## Tests

- **Completeness (R3):** a proptest over several variable counts and random polynomials and points, asserting both sides agree on the surviving claim, that the claim is the truth about the packed polynomial, and that both sponges are handed back in the same state.
- **Soundness (R3):** one mutation test per absorbed value — every point coordinate, the tensor coefficients at both ends and in the middle, and the surviving claim. The seven pre-existing end-to-end rejection tests still pass.
- **Separation (R4):** the shape is walked with a destructuring exhaustiveness check, so adding a field to it breaks compilation rather than silently escaping the seed. Seeds are compared pairwise.
- The description is asserted to declare no proof-of-work step, rather than silently permitting one. This reduction offers no grinding knob and its bound is per-attempt.
- Registered as the sixteenth protocol in the cross-protocol separation suite, so the new name is actually compared against the other fifteen.

## Edge cases that are reachable and covered

A zero-width step is reachable: an extension equal to the base field gives a zero-width batching point, and a zero-variable run gives a zero-width point step. A pre-existing identity-extension test covers both.

## Verification

| Command | Result |
| --- | --- |
| `cargo test -p p3-sumcheck` | 339 + 6 passed, 0 failed |
| `cargo test -p p3-challenger` | 191 passed, 0 failed |
| `cargo test -p p3-stir` | 109 + 108 + 28 passed, 0 failed |
| `cargo test -p p3-examples --test transcript_separation` | 4 passed, 0 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |

All re-run after rebasing onto #2121.

## Not done here

The pre-existing prose in the ring-switching module wraps sentences across lines and puts doc headers on its test functions, both against the house documentation style. Every doc link in the files touched here was rewritten and all new lines follow the style, but the pre-existing prose was left alone rather than bundling a large unrelated reformatting into a migration PR. Happy to do it as a separate cleanup if you want it.


challenger/src/fs/state/prover.rs
```diff
@@ -805,6 +805,50 @@ impl<C, U: Unit> ProverState<C, U> {
...
+    ///     n steps   of width 1   <- not this
+    /// ```
+    ///
+    /// The count comes from the caller's configuration.
+    ///
+    /// It is part of the recorded shape.
+    ///
+    /// The seed covers it.
+    ///
+    /// Use the rejecting draw when the protocol constrains what a coordinate may be.
```
Nashtare, line 832: **Answering the "Reviewer's option" in the description directly: yes, the accept-all rejecting draw is byte-identical at the pattern level.** This method and `challenge_extensions_rejecting` (`:621-626`) construct the *same* `Interaction::algebra::<F, EF>(Atomic, Challenge, label, Length::Fixed(count))`, so no seed can tell them apart.

That is worth fixing rather than leaning on, because the crate already treats this exact situation as a separation problem one layer down. `TypeTag::UniformBits` exists solely so that a biased and an unbiased draw of the same width cannot share a shape — `pattern/step.rs`:

> The two draws answer the same question with different distributions. So they are different shapes, and the fingerprint separates them.

The plain and the rejecting extension draws are the same pair, and they are not separated. Concretely: `multi-stark/src/zerocheck/transcript.rs:310` draws TAU at `Fixed(free_coordinates)` under a `!candidate.is_zero()` predicate. If anyone ever "simplifies" that to this new method, both the distribution and the sponge stream change while the seed stays byte-identical — which is the one class of format change the fingerprint is there to catch.

So I'd keep the method (the two existing rejecting call sites both carry real predicates, so this is genuinely the first unconstrained draw) but give it a shape of its own, or at minimum say on both methods that the fingerprint does not part them.

challenger/src/fs/state/prover.rs
```diff
@@ -1205,6 +1249,93 @@ mod tests {
...
+        #[cfg(panic = "unwind")]
+        {
+            let outcome = catch_unwind(AssertUnwindSafe(|| {
+                let mut p = ProverState::<_, u8>::new(byte_sponge(), &ds);
+                let _ = p.challenge_extensions::<F, EF4, ByteCodec>("point", 2);
+                p.finalize()
+            }));
+            assert!(
+                outcome.is_err(),
+                "a narrower draw must fail the shape check"
```
Nashtare: This asserts only that *something* panicked, so it does not pin what the name claims. The closure both draws and calls `p.finalize()`: if `challenge_extensions` did no validation at all and simply drew two challenges, the cursor would sit at 0 with one step recorded and `finalize` would panic with `Pattern not fully replayed` — `outcome.is_err()` all the same. (`assert_challenge_security` can't be the source; it's a `const {}` block in `state/mod.rs`.)

The sibling `catch_unwind` test in this file, `a_panic_inside_a_live_prover_scope_unwinds`, does downcast the payload and compare it. Either match that here (`Received interaction … but expected …`), or drop the `p.finalize()` from the closure so the shape check is the only thing that can panic.

examples/tests/transcript_separation.rs
```diff
@@ -70,14 +76,14 @@ type Case = (String, DomainSeparator<Alphabet>);
 /// Number of protocols on the typed transcript layer.
 ///
 /// A protocol added without an entry below leaves its name unchecked against the others.
-const NUM_PROTOCOLS: usize = 15;
+const NUM_PROTOCOLS: usize = 16;
```
Nashtare: The STIR PCS commit on this branch introduces three more domain-separated protocols — `p3-stir-pcs-commitment`, `p3-stir-pcs-claims`, `p3-stir-pcs-opening` (`stir/src/pcs_transcript.rs:114,117,120`) — and none of them is registered here. The count goes 15 → 16 for ring switching alone.

By this file's own doc that means their names are never compared against anything: *"A protocol added without an entry below leaves its name unchecked against the others"*, and `every_protocol_is_listed_here` only compares two hand-maintained numbers. Workspace-wide there are now 22 distinct protocol names against 16 registered groups.

It matters more here than it usually would. All four `p3-stir-pcs-*` names — including the pre-existing `p3-stir-pcs-batch` in `batch_transcript.rs` — are strict extensions of `p3-stir` from `stir/src/transcript.rs`. That is exactly the prefix relation `a_shared_name_prefix_is_separated_by_the_name_length_byte` was written for (`p3-fri`/`p3-fri-pcs`, `p3-whir`/`p3-whir-hvzk`), and not one of the five `p3-stir*` pairs is covered by it. The three shapes are `pub` and reachable as `p3_stir::pcs_transcript::{StirPcsCommitmentShape, StirPcsClaimShape, StirPcsOpeningShape}`, so adding the cases is mechanical.

stir/src/pcs.rs
```diff
@@ -1297,26 +1363,25 @@ where
...
+        //
+        //     commitment -> matrix -> opening point -> one value per column
+        //
+        // So the grouping reaches the seed, not only the absorbed values.
+        //
+        // The claims are the statement this call was handed.
+        //
+        // Nothing here is read out of the proof.
+        //
+        // No count is attacker-chosen.
```
Nashtare: This overstates what `verify` itself guarantees. The claim tree reaching `Pcs::verify` in a STARK is built out of the proof's own `opened_values` (`uni-stark/src/verifier.rs:354,377,389`). It is `uni-stark` that checks the widths against the AIR, at `verifier.rs:533`, *before* this call; the pre-check just above (`:1333-1364`) only enforces internal consistency — non-empty, uniform across points, non-zero.

So the counts are attacker-supplied and caller-constrained, not fixed by anything in this function. Deriving the seed from them is a clear improvement over the old flat absorb, and nothing here is wrong — it's the wording I'd tighten, so a future caller doesn't read it as STIR validating the claim shape on its own behalf.

stir/src/pcs.rs
```diff
@@ -1101,48 +1123,92 @@ where
...
+        let bucket_proofs = izip!(
+            &bucket_log_heights,
+            &stir_configs,
+            bucket_results,
+            bucket_lanes
+        )
+        .map(|(&log_h, stir_config, (stir_proof, first_round), lanes)| {
+            let log_arity0 = stir_config.log_starting_folding_factor;
+            // Both the lane count and the draw count come from the schedule.
+            debug_assert_eq!(lanes.len(), first_round.draws.len());
```
Nashtare: Worth making this a hard assert (same for the verifier's copy at `:1741`). The lane count now has to come from the schedule, since the description is fixed before the driver is seeded, while `draws` still comes from the run — so the two are derived independently for the first time.

I checked that they agree today: `first_round_draws` is pushed only at `round == 0` (`verifier.rs:713`) or in the final round when `num_rounds == 0` (`verifier.rs:970`), and both index lists are drawn as `Length::Fixed(num_queries)` / `Fixed(final_queries)` with no deduplication, which is what `first_round_draw_count`'s `map_or` mirrors. But `query_positions` zips the two lists, so if a schedule ever reads the initial oracle from a round other than zero, a release build silently truncates to the shorter one and produces a proof against the wrong positions. At one comparison per bucket the check is free.

sumcheck/src/ring_switch/mod.rs
```diff
@@ -309,19 +426,42 @@ where
         .map(|(&column, &weight)| column * weight)
         .sum();
     if combined != claimed_sum {
+        // Two described steps are still unplayed, and this rejection plays neither.
+        //
+        // Releasing the completeness check keeps this the only failure reported.
+        transcript.abort();
```
Nashtare, line 432: Checked the ordering change the description flags, and the reasoning holds. The column check reads `proof.s_hat.columns()` against `eq_low`, both local values, and never touches the challenger — so moving it after the draw genuinely does not move the sponge, and keeping the two absorbs and the dependent draw atomic in one call is worth the move.

The rejection path is right too: `abort()` here leaves the two unplayed steps released, so the `ClaimMismatch` reaches the caller instead of colliding with the driver's drop check. Same for the round-failure path below — and the bracket is closed either way, since `batched_sumcheck` runs `end_protocol` on the value the closure returns rather than on success.

Nashtare (commented):
Reviewed the whole diff, which spans both commits: `1ca75436` (ring switching, this PR's subject) and `27c8c590` (the STIR PCS migration, merged in from #2123).

The two properties that matter for the ring-switch commit both hold. **The sponge sequence is unchanged apart from the added seed**: `ExtensionFieldCodec<F, EF, FieldToFieldCodec<F>>::observe`/`sample` walk the basis in the same order and with the same per-coefficient sponge calls as `observe_algebra_element` / `sample_algebra_element`, so this really is a seed-only format break. **No width is read off the proof**: the point width comes from the caller's own `r`, the tensor width is `DIMENSION²`, the batching width is `packed_vars`, and all three reach the seed — the first through `Length::Fixed(num_variables)`, the other two through `TypeTag::Algebra { modulus, degree }`, which `pattern_hash` renders. Every verifier exit releases the completeness check before returning, including the two new ones.

The STIR half holds up too: `observe_opened_values` and `observe_claims` derive the same `StirPcsClaimShape` from two different trees and are pinned against each other by a test; `first_round_draw_count` matches what the verifier actually accumulates (draws are pushed only at `round == 0`, or in the final round when `num_rounds == 0`, and neither index list is deduplicated); and alpha is still drawn strictly after the claims on both sides.

Two things worth acting on (both MEDIUM, neither blocking), plus three smaller notes inline. `cargo test -p p3-sumcheck -p p3-challenger -p p3-stir`, `cargo clippy --workspace --all-targets -D warnings` and `cargo +nightly fmt --check` are all clean at `27c8c590`.

challenger/src/fs/state/prover.rs
tcoratger: You are right on the mechanism, and I did not take the shape fix — deliberately. Reasoning, so you can overrule it.

The two draws render identically in alternate mode, `Atomic Challenge <label> Fixed(n) Algebra{modulus, degree}`, and only `type_tag` carries identity there. So separating them needs a new `TypeTag` variant or a new `Kind`. Both are changes to a shared enum, and both move the fingerprint of the two existing rejecting call sites — the zerocheck TAU draw and the STIR out-of-domain draw. That cascades fixture regeneration through four PRs already stacked above this one, inside a PR whose subject is ring switching.

There is also a design question I did not want to settle under that pressure: a `Filtered { modulus, degree }` variant is wrong for tower fields, whose tag is `BinaryTower { bits, degree }`, and `challenge_extensions_rejecting` is reachable over those. So the honest fix is either two new variants or moving the sampling discipline onto `Kind`, and `Kind` is matched on in structural logic and in the grinding accounting.

So I took your fallback, but made it executable rather than prose. Both methods now carry a `# What the shape does not part` section stating that no fingerprint or seed tells them apart, and pointing at the bit-width tags as the precedent for what a real fix looks like. Alongside it, `a_predicate_moves_the_stream_but_not_the_shape` drives two provers over the **same** description — one unconstrained, one with a predicate that drops the first candidate — asserts the drawn streams differ, and finalises both against that one description. So the collision is recorded as a passing test rather than as folklore, and the day someone adds the step type, that test is what visibly flips.

Happy to do the shape fix as its own PR where the fingerprint change and the fixture regeneration are the whole diff and can be reviewed on their own terms. Say the word and I will open it rather than leave it as a comment.

challenger/src/fs/state/prover.rs
tcoratger: Fixed, and you were right about the mechanism — `finalize` inside the closure meant an unvalidated draw would have panicked on the completeness check and `is_err()` would have held all the same.

Dropped the `p.finalize()` from the caught closure and compared the payload instead, matching the sibling test in the file:

```rust
assert!(message.starts_with("Received interaction "), ...);
assert!(message.contains("Fixed(2)") && message.contains("Fixed(3)"), ...);
```

The driver is now left unfinalised on purpose, and its drop check yields to the panic already in flight, so exactly one panic reaches the catch. The second assertion also pins that the diff names both widths, so the test fails if the message stops carrying the mismatch it claims to report.

examples/tests/transcript_separation.rs
tcoratger: Done — all three registered, and the prefix relation covered.

`stir_pcs_commitment_cases`, `stir_pcs_claim_cases` and `stir_pcs_opening_cases` are in `protocols()`, and `NUM_PROTOCOLS` goes 16 -> 19. Each contributes three configurations, so the pairwise cross-product test now covers them against everything else.

The claim cases are chosen to be the interesting ones rather than arbitrary: three groupings that all flatten to the same widths.

```text
    one matrix, two points   [[[3, 3]]]
    two matrices, one point  [[[3], [3]]]
    two commitments          [[[3]], [[3]]]
```

Only the containers part those, so they exercise exactly the property the nesting was introduced for, and they now do it against the other eighteen protocols too.

`a_shared_name_prefix_is_separated_by_the_name_length_byte` gained the three `p3-stir` / `p3-stir-pcs-*` pairs next to the FRI and WHIR ones.

On the fifth pair: `p3-stir-pcs-batch` lives behind a crate-private separator, so I could not register it from `p3-examples` without widening the STIR public API for a test. I asserted it inside the STIR crate instead, in `this_name_is_separated_from_the_one_it_extends`, which builds a separator over the same description under the bare `p3-stir` name and checks that the differing byte is the length byte. The comment in the examples test says where that pair is covered, so the omission is not silent.

stir/src/pcs.rs
tcoratger: Fixed — you are right, and the wording was doing real harm since it invited exactly the wrong reading.

"Nothing here is read out of the proof. No count is attacker-chosen." is gone. It now says the claims are the statement this call was handed rather than proof fields it read, that their shape is the caller's to fix and this function does not fix it, and spells out the division:

```text
    the caller  ->  decides the claim tree, and constrains it
    here        ->  binds whatever tree it was handed
```

with the STARK caller building that tree from the proof's own opened values and checking widths against its AIR before reaching this call, and the pre-check above enforcing only internal consistency. It closes by saying that deriving the seed from the claim layout binds the statement to the run, and is not STIR validating that layout on its own behalf.

stir/src/pcs.rs
tcoratger: Done — both are hard asserts now, prover and verifier.

Your reasoning is the reason I was comfortable making them hard rather than structured errors: both counts are pattern-derived, `Length::Fixed(num_queries)` or `Fixed(final_queries)`, so no proof can make them disagree. That makes this an internal-invariant check between two independent derivations of the same configuration quantity, not input validation, so R5 is intact.

Each carries the two numbers in its message, and the comment now names the hazard rather than just asserting the equality:

```text
    the description fixes the lane count before the driver is seeded
    the draw count comes out of the run
    pairing them zips two lists, which would silently truncate to the shorter
```

Thanks for checking that they agree today and for pinning down exactly when they would not — a schedule reading the initial oracle from a round other than zero. That is the scenario the assert now catches in release as well as debug.

sumcheck/src/ring_switch/mod.rs
tcoratger: Thanks for verifying this independently — you traced the same three things I did, and it is useful to have that on the record.

Two points worth confirming back. The `abort()` on the rejection path is deliberate, and for the reason you name: without it the drop check fires on top of an error already in flight. And you are right that the bracket closes either way, because the delegation runs `end_protocol` on whatever the closure returns rather than only on success — so a rejection inside the sumcheck leaves the transcript replayable to the end rather than stranding an open container.

Nothing to change here.

examples/tests/transcript_separation.rs
```diff
@@ -661,19 +767,32 @@ fn the_protocol_name_is_the_only_field_that_separates_two_protocols() {
...
+    //     [1 | p3-whir-hvzk            | 0 .. 0 | 12]
     //
-    //     [1 | p3-whir      | 0 .. 0 |  7]
-    //     [1 | p3-whir-hvzk | 0 .. 0 | 12]
+    //     [1 | p3-stir                 | 0 .. 0 |  7]
+    //     [1 | p3-stir-pcs-commitment  | 0 .. 0 | 22]
+    //     [1 | p3-stir-pcs-claims      | 0 .. 0 | 18]
+    //     [1 | p3-stir-pcs-opening     | 0 .. 0 | 19]
+    //
+    // The batching phase's name extends it too, but its separator is crate-private.
```
Nashtare: **LOW — rebase note: this caveat goes stale against main.**

#2120 made the batching separator public (`p3_stir::batch_domain_separator`, re-exported from `stir/src/lib.rs`) and already registers `stir_pcs_batch_cases()` in `protocols()`. After the rebase:

- the pair is reachable from here, so `(stir_cases(), stir_pcs_batch_cases())` can join the pairs below and this caveat goes;
- `this_name_is_separated_from_the_one_it_extends` in `stir/src/batch_transcript.rs` then duplicates it and can be dropped;
- `NUM_PROTOCOLS` lands at 20 (main's 16, plus ring switching and the three PCS phases), not 19.

Main's copy of the prefix test still lists only the FRI and WHIR pairs, so nothing else picks this up.

Nashtare (approved):

examples/tests/transcript_separation.rs
tcoratger: Done — rebased onto main, and the caveat is gone.

`p3_stir::batch_domain_separator` is public since #2120, so `stir_pcs_batch_cases()` joins `protocols()` and `(stir_cases(), stir_pcs_batch_cases())` joins the pairs below. `this_name_is_separated_from_the_one_it_extends` in `stir/src/batch_transcript.rs` duplicated that pair and is dropped.

`NUM_PROTOCOLS` lands at 20, exactly as you called it: main's 16, plus ring switching and the three PCS phases.

PR #2124, opened 2026-09-10
feat(sumcheck)!: drive the layout sumcheck transcript through the typed Fiat-Shamir layer

tcoratger:
> Stacked on #2122 (which now also carries #2123) → #2121. Review bottom-up.

## Why this one mattered most

WHIR was migrated by #2104 and brackets its folding phases like this:

```rust
transcript.delegate_initial_fold(|challenger| { /* sumcheck ran here with a bare sponge */ })
```

The bracket was recorded, but the child inside it absorbed and drew straight off the sponge. A bracket around an untyped region is a transitional hatch, not a composition.

The layout sumcheck is also reached **before** WHIR's driver ever seeds, from the PCS adapter's claim-binding loops:

```text
    commitment absorb        outside any pattern
    layout claim binding     outside any pattern
    WHIR run                 typed, seeds here
```

After this PR the middle line is typed, and WHIR's initial delegation bracket becomes a real composition: a typed batching draw followed by the typed quadratic sumcheck, with nothing raw between them.

## The description

Three components, each seeding its own sub-transcript from the borrowed sponge and finalizing before it returns:

| Component | Steps |
| --- | --- |
| opening batch | opening point (only when drawn here), direct evaluations, successor evaluations |
| out-of-domain claim | point, claimed evaluation |
| batching | one challenge |

An empty column group emits no step, so a direct-only batch and a mixed batch have different fingerprints.

## Prefix and suffix: one shape, not two

The two files are two variable orders of one protocol. They record the same claims, absorb the same values, draw the same challenges, and their step sequences are identical step for step.

Two descriptions would state that equality twice and let one copy rot silently. A reviewer could not then tell an intentional divergence from a stale one.

So there is one shape carrying the order as a value. The order cannot move the pattern fingerprint; it moves the **seed** instead, through the instance label, alongside selector placement, stacked arity, the source table index and arity, and both claim counts. The separation test walks each field and includes the variable order explicitly labelled as the knob that moves no step.

## One forced API change

The prover's batching draw lives inside the initial delegation bracket, and its verifier mirror lives in the WHIR verifier. Typing one requires typing the other:

```text
    before:  FnOnce(EF) -> Constraint<F, EF>
    after:   FnOnce(&mut Challenger) -> Constraint<F, EF>
```

The batching challenge belongs to the claims the caller recorded, not to the WHIR run, so it has to be drawn through the layout's own driver on both sides. The two adapter closures and one binary-PCS draw now call the layout verifier's batching-challenge method themselves.

The sponge-taking entry points kept their signatures, so the adapter's out-of-domain and per-opening loops needed no edit — they were already handing over a borrowed sponge, and now call typed components.

## One site deliberately left raw

`sumcheck/src/commit.rs:62`, the Merkle-root absorb at commit time.

The house idiom for an opaque commitment is a bracketed opaque step in the **parent's** pattern — the way the multi-STARK statement transcript wraps its main commitment — not a child sub-transcript. A full seed for one opaque step is the wrong shape.

More importantly the absorb is mirrored from four places that must change together:

```text
    sumcheck/src/commit.rs:62        prover side
    whir/src/pcs/adapter.rs:148      plain verifier
    whir/src/pcs/zk/adapter.rs:166   hiding verifier
    binary-pcs/src/pcs.rs:430        binary verifier
```

Binding it inside the sumcheck crate would force all four to reconstruct an identical separator from geometry they do not all carry. This is scheduled as its own PR covering every mirror at once.

What a commit-time separator would have bound — stacked arity, selector placement, binding order — is now bound anyway in the claim shapes' instance labels, before the first challenge of the phase. So the transcript is strictly better bound than before, without that coupling.

## Cross-crate edits a reviewer should check

- **`sumcheck/src/zk/`, four lines.** `F: TranscriptField` added to the `where` clause of four pure-delegation wrappers, because their callees now require it. Both files already import the trait and already use that bound elsewhere.
- **`multi-stark/tests/fixtures/multi_stark_whir_v0_8_0.postcard` regenerated** via the crate's own ignored generator. Expected for a breaking transcript change.
- One pre-existing binary-PCS test hand-rebuilds the verifier transcript; it now draws its batching challenge through a zero-claim layout verifier, which is the point of that test.

## Tests

- **Separation (R4):** three pairwise seed-distinctness tests, one per component, over every configuration field, plus a fourth proving the three components never collide with each other.
- **Completeness (R3):** both sides of the claim phase reaching one batching challenge over a mixed schedule — both column groups, both tables, an out-of-domain claim — for both binding orders, plus a proptest over random schedules. The pre-existing full round-trip proptests now exercise all three drivers.
- **Soundness (R3):** one mutation test per absorbed value, at two positions each so no single position carries the binding, asserted at both the transcript level and the protocol level.
- **R5:** a recorded batch of the wrong width is rejected as a structured error with a quiet drop. The schedule assertions now run *before* the driver exists, so a rejected claim never leaves an unfinished driver.

Grinding difficulties are not checked here: this phase describes no proof-of-work step. Grinding belongs to the residual sumcheck, which already binds it.

## Verification

18 test suites pass with 0 failures across `p3-sumcheck`, `p3-whir`, `p3-stir`, `p3-binary-pcs` and `p3-multi-stark`, re-run after rebasing onto the current #2122 tip. Clippy clean with `-D warnings` on all five crates; `cargo fmt --all --check` clean.

## Raw sponge sites remaining in this scope: 1

`sumcheck/src/commit.rs:62`, deferred deliberately as described above.


sumcheck/src/layout/transcript.rs
```diff
@@ -0,0 +1,1576 @@
...
+        EF: ExtensionField<F>,
+    {
+        let mut separator = DomainSeparator::new(VERSION, OPENING_NAME, self.pattern::<F, EF>());
+
+        // Layout geometry comes first.
+        //
+        // Every component of one layout then shares that prefix.
+        self.binding.bind::<F>(&mut separator);
+
+        // Two batches of equal widths on two tables must not share a seed.
```
Nashtare, line 434: **LOW — which columns a batch opens, and the table shapes that place them, reach no seed.**

The reason given here for binding the table index applies just as well within a table and across layouts, and neither of those is bound:

- **Column indices.** The verifier lifts each claim through `placement.selectors()[col]` (`verifier.rs:497-523`), so opening column 0 or column 1 of one table with the same value is a different claim. The shape carries only `num_current` / `num_next`, so both runs get the same seed, absorb the same values and draw the same batching challenge.
- **Table shapes.** `LayoutBinding` carries the stacked arity only. Tables `[(9, 2), (10, 2)]` and `[(9, 3), (10, 1)]` (arity, width) both stack to 12 variables but put table 0's columns at different slots, so one opening again means two different constraints under one transcript.

Checked with a probe for both cases: identical batching challenge and identical next sponge sample, different `eval_constraints_poly(constraint(alpha))`.

There is no known exploit. The schedule is the verifier's own, multi-STARK's statement seed binds widths and arities (and, on main since #2120, the next-row column lists), and all of this was raw before this PR. But the module doc at `:95` says "the rest of the configuration moves no step" above a list that reads as complete, which is what a `Pcs::verify` caller outside multi-STARK would rely on. Either add the per-group column index lists and a per-table `(arity, width)` list as instance chunks, or scope the doc to say the opening schedule is statement data the caller must bind.

sumcheck/src/layout/transcript.rs
```diff
@@ -0,0 +1,1576 @@
...
+    /// Label a shape's seed digest so a failing pairwise check names the knob.
+    fn labelled_opening(name: &str, shape: OpeningShape) -> (String, SeedDigest) {
+        (
+            String::from(name),
+            seed_digest(&shape.domain_separator::<F, EF>()),
+        )
+    }
+
+    #[test]
+    fn every_knob_of_a_recorded_batch_reaches_the_seed() {
```
Nashtare, line 1230: **LOW (test) — these separation tests build their variants with `..base`, so a field added later compiles without reaching a seed or a test.** The ring-switch shape on the base PR names every field first (`let RingSwitchShape { num_variables: _ } = base_shape();`, `ring_switch/transcript.rs:675`), and so do the FRI PCS and multi-STARK statement tests. Doing the same for `OpeningShape`, `LayoutBinding`/`LayoutStrategy` and `BatchingShape` here (and ideally inside `bind` / `domain_separator` too) is what would turn the gap above into a compile error the next time a knob is added.

sumcheck/src/layout/transcript.rs
```diff
@@ -0,0 +1,1576 @@
...
+
+/// Version byte bound into every seed this phase produces.
+///
+/// Bumping it separates two revisions of the phase.
+///
+/// It separates them even when their step sequences agree.
+const VERSION: u8 = 1;
+
+/// Protocol name of one recorded batch of concrete openings.
+const OPENING_NAME: &[u8] = b"p3-sumcheck-layout-opening";
```
Nashtare: **LOW — the three new protocol names are not registered in the cross-protocol separation suite.** `examples/tests/transcript_separation.rs` still has `NUM_PROTOCOLS = 19` on this branch, and its doc says an unregistered protocol "leaves its name unchecked against the others". The shapes are `pub(crate)`, so `p3-examples` cannot reach them. The STIR batching phase hit the same wall on #2122 and asserted its prefix relation inside its own crate instead. The names are unique today, so this is about keeping it that way: an in-crate check against the neighbouring `p3-sumcheck-*` names, or a test-only constructor the suite can call.

sumcheck/src/layout/verifier.rs
```diff
@@ -181,93 +225,203 @@ impl<F: Field, EF: ExtensionField<F>> Verifier<F, EF> {
...
+    ///
+    /// # Overview
+    ///
+    /// Each claim is weighted by a successive power of the drawn challenge.
+    ///
+    /// ```text
+    ///     sum = sum_i  alpha^i * eval_i
+    /// ```
+    ///
+    /// Direct and successor openings take the low powers, in insertion order.
```
Nashtare: **LOW (doc) — powers follow placement order, not insertion order.** `sum` / `constraint` here and `StackedClaims::concrete_claims` (`prover/claims.rs:103-107`) walk tables in placement order (largest arity first), and insertion order holds only within one table. So a claim recorded first against a small table does not take `alpha^0`. The same sentence is at `prover/mod.rs:378`, and `transcript.rs:896` ("the claim counts decide which power lands on which claim") overstates the same way: counts fix how many powers there are, and placement fixes which claim gets which.

whir/src/pcs/verifier/mod.rs
```diff
@@ -115,7 +122,7 @@ where
     ) -> Result<Point<EF>, VerifierError>
     where
         Challenger: CanObserve<MT::Commitment>,
-        MakeConstraint: FnOnce(EF) -> Constraint<F, EF>,
+        MakeConstraint: FnOnce(&mut Challenger) -> Constraint<F, EF>,
```
Nashtare: **LOW — the new closure hands the raw sponge to the caller inside WHIR's initial-fold bracket.** A bracket checks nothing about the draws inside it, so a closure that draws zero challenges, or two, only shows up as an honest proof failing to verify. The prover side cannot drift like that, since `into_sumcheck` owns its draw. All current callers do exactly one `Verifier::batching_challenge` (`whir/src/pcs/adapter.rs:208`, `:348`, `binary-pcs/src/pcs.rs:358`, `binary-pcs/src/verifier.rs:611`, `sumcheck/src/tests.rs:344`). Taking `&layout::Verifier<F, EF>` and calling `batching_challenge` + `constraint` inside WHIR would make that the only shape expressible, and would remove the one place in this phase where a caller still holds the sponge between typed steps.

Nashtare (approved):

sumcheck/src/layout/transcript.rs
tcoratger: Took the first option: both are bound now, rather than scoping the doc.

Your two cases were reachable with no new plumbing, since both sides already hold the data at shape-construction time — the prover from `OpeningRequest` and `StackedClaims::tables`, the verifier from `OpeningRequest` and its own placements.

- **Column indices.** `OpeningShape` carries `current: Vec<usize>` and `next: Vec<usize>` instead of the two counts, and both lists reach the instance label. `num_current()` / `num_next()` derive the widths the pattern still declares.
- **Table shapes.** `LayoutBinding` carries every source table's `(arity, width)` in caller order, so all three components of one layout share it. The verifier rebuilds the list from its placements: `arity = k - selector bits`, `width = one selector per column`, which is the identity the seed already relied on for the single opened table.

Each list leads with its length as its own chunk, matching how `multi-stark` binds `instances` and `bus_ids`, so a shorter list cannot borrow the chunks after it.

Your `[(9, 2), (10, 2)]` vs `[(9, 3), (10, 1)]` case is now `table widths` in `every_knob_of_a_recorded_batch_reaches_the_seed`, alongside `direct columns`, `direct order` and `successor columns`.

One test moved as a result: `eval_current_preserves_order` used `eval`, which draws its point, so requesting `[1, 0]` and `[0, 1]` used to draw the same point and the evals were exact reverses. They now draw different points — which is the separation working — so the test uses `eval_at` at a fixed point, and the separation itself is checked in `transcript.rs`.

The module doc's list is updated and no longer reads as complete-with-the-schedule-outside-it.

sumcheck/src/layout/transcript.rs
tcoratger: Done, and extended past the tests as you suggested.

`..base` is gone: each variant starts from a fresh `base_opening()` / `base_binding()` and moves one field, the way the ring-switch shape does. Every field is named first, with no rest pattern:

```rust
let OpeningShape { binding: _, table_index: _, table_variables: _, current: _, next: _, point: _ } = base_opening();
let LayoutBinding { num_variables: _, strategy: LayoutStrategy { reverse_selectors: _, variable_order: _ }, tables: _ } = base_binding();
let BatchingShape { binding: _, num_claims: _, num_virtual_claims: _ } = BatchingShape::new(base_binding(), 4, 2);
```

The geometry cases live in one `one_step_from_base_binding()` that all three tests iterate, so a field added to `LayoutBinding` or `LayoutStrategy` breaks one destructuring and is then missing from three suites at once.

`LayoutBinding::bind` and all three `domain_separator`s carry the same `let Self { .. } = self;` exhaustiveness check, so a new knob fails to compile at the binding site itself, not only in the tests.

whir/src/pcs/verifier/mod.rs
tcoratger: Done — `verify` takes `&layout::Verifier<F, EF>` and draws inside WHIR.

```rust
transcript.delegate_initial_fold(|challenger| {
    let alpha = layout.batching_challenge(challenger);
    let constraint = layout.constraint(alpha);
    ...
})
```

The bounds line up exactly: `WhirVerifier`'s `Challenger` is already `FieldChallenger<F> + GrindingChallenger<Witness = F> + CanSampleUniformBits<F>`, and `whir` already depends on `p3-sumcheck` for `Constraint`, `Basis` and `VariableOrder`, so this adds one import rather than a dependency.

Both callers in `whir/src/pcs/adapter.rs` did exactly the two lines above, so the diff is a deletion at each. The binary-pcs and `sumcheck/src/tests.rs` sites you listed draw outside `WhirVerifier::verify` and are untouched.

One draw is now the only shape expressible on this phase, and no caller holds the sponge between typed steps here any more.

sumcheck/src/layout/verifier.rs
tcoratger: Fixed, and you are right about the mechanism — the outer loop is over `placements`, which `plan_layout` orders largest arity first, so insertion order holds only among the claims of one table.

Both summary lines now read:

> Direct and successor openings take the low powers, in placement order.
> Tables are walked largest arity first, not in the order they were opened.
> Insertion order holds only among the claims of one table.

`transcript.rs` no longer says the counts decide which power lands where — the counts fix how many powers there are, and placement fixes which claim gets which.

Also swept the phrase where it named the wrong loop: the `// Walk every concrete opening in the canonical insertion order` comments (which already sketched `placements -> claims -> ...` right underneath), `StackedClaims::sum`'s alpha-ordering doc, and `MultiClaim`'s ordering contract.

sumcheck/src/layout/transcript.rs
tcoratger: Took the in-crate check.

Publishing `OpeningShape`, `VirtualShape` and `BatchingShape` to reach `p3-examples` would drag `LayoutBinding`, `PointSource` and the rest of the layout's internals into the public API for a test, which is the same trade I declined for `p3-stir-pcs-batch` before #2120 made that separator public for its own reasons.

`no_two_protocols_in_this_crate_share_a_name` in `p3_sumcheck`'s suite compares all seven names the crate seeds from — the four that travel to `p3-examples` and the three layout ones — taken from the constants themselves, so a rename moves the check with it. `a_layout_name_is_separated_from_the_one_it_extends` covers the shared `p3-sumcheck-layout-` stem the way the workspace suite covers `p3-fri` / `p3-fri-pcs`.

`every_protocol_is_listed_here` now says which three names cannot join and where they are checked instead, so the gap is not silent.

PR #2125, opened 2026-09-10
refactor(fri)!: derive the commit-round shape from configuration rather than the proof

tcoratger:
> Rebased onto #2122, which now also carries #2124. Review that one first.
>
> **This PR also carries #2126.** #2126 was squash-merged into this branch, so what
> lands here is the derived commit-round shape below *plus* the hiding WHIR pattern
> player. The review threads on both PRs are addressed and resolved.
>
> Follow-up review fixes in this branch: `fold_query` validates each caller-supplied
> schedule entry and sums with saturation, so the three panics the review found are
> structured rejections; `commit_round` takes `witness: F`; and the zero-difficulty
> canonical-witness rule moved into the two hiding drivers so it holds for every caller.

## The finding

PR #2035 migrated FRI onto the typed transcript layer with a stated completion criterion it did not meet. Three error variants were supposed to be "gone or unreachable" once the round count and arity schedule came from configuration. All three were still live, and the verifier still read its own shape from the proof:

- the number of commit rounds came from the length of the proof's commitment vector
- the per-round arities came from a field on each commit-phase opening
- then roughly sixty lines proved to itself that those attacker-supplied numbers were the right ones

Those checks were correct. The problem is that they were load-bearing hand-written arithmetic standing in for something the configuration already knew, they had to be re-derived by every new folding feature, and the arity schedule needed a separate explicit absorb to be bound at all.

## A finding about the layer itself, which shaped the fix

**A pattern mismatch on the verifier side is a panic, not an error.** The pattern player panics on divergence, and the layer's error module documents this as deliberate: a mismatch between the description and the replay code is a programming bug, not a malformed input.

So the round count could **not** simply be handed to the driver and left to "fail the pattern" — that would have converted a malformed proof into a panic, violating R5.

Instead the count check moved *into* the shape layer, where the length lives, and surfaces as a structured error — the same mechanism the final polynomial's coefficient count already used. The per-round entry point is now private; the only public entry takes the whole round list and checks both lengths against the derived schedule before playing anything. **Driving the transcript with a proof-supplied count is no longer expressible.**

## Removed, each with the proof

| Removed | What the old check caught | Why it is gone | Proving test |
| --- | --- | --- | --- |
| `InvalidLogArity`, the proof's arity field, its checked accessor (and the Circle twins) | an opening with a zero or over-large arity | the field left the proof format, so the mutation no longer compiles; the derived schedule satisfies the bound by construction | an exhaustive test over all 1023 strictly-decreasing height subsets of `0..10`, every cap in `1..=4`, every final height in `0..4` |
| `GlobalMaxHeightMismatch` and the two-derivations cross-check | proof rounds summing to a height different from the claim's | the global height now has exactly one derivation, so there is no second number to disagree with | a uniformly shortened commit phase, and an extra commit round, each rejected with counts |
| `FoldScheduleMismatch` | a prover that genuinely folded on a permuted schedule and was self-consistent about it | nothing left to compare: the derived schedule drives folding, sibling widths and the seed | the prover-side forgery seam is kept, and is now rejected on sibling width **before the transcript is seeded** |
| the proof-vs-proof form of the openings-count mismatch | — | the count comes from the schedule now, not from the proof | the same two tests |

## Kept, and why the pattern does not subsume them

**The per-round arity absorb — kept.** This is the important one. The step sequence carries only the round *count*, one group of steps per round. So `[3, 3, 2]` and `[2, 3, 3]` produce a **byte-identical** step sequence, the same round count, the same index width (derived from the equal total), the same cap and blowup:

```text
    [3, 3, 2]  folds 8 -> 5 -> 2 -> 0
    [2, 3, 3]  folds 8 -> 6 -> 3 -> 0

    round count   3 == 3
    index width   8 == 8
    final poly    1 == 1
```

Both are reachable from real configurations — inputs `[10]` versus `[10, 8]` at one cap — so these are two genuine protocols, not a tampered field. Only the instance label separates them. A new test asserts the step sequences are **equal** and the seeds **differ**.

Also kept, each for a stated reason:

- **the openings-count check** — the openings vector is never absorbed, so no described step has its length; a short list would silently skip the trailing rounds, and with them their Merkle verification
- **the two-adicity bound** — the fingerprint reaches the global height only indirectly through the derived index width, and nothing in the pattern knows the field's two-adicity; without this bound the generator lookup panics
- **the query-count and sibling-width checks** — neither list is absorbed
- **five schedule and reduced-opening checks** — unreachable from the top-level verify now, but the fold entry point is public API taking a caller-supplied schedule, so they keep it total

## One R5 hole opened and closed in this PR

The verifier now always derives a schedule, and the per-round arity computation asserts a positive cap. A cap of zero would have panicked where it previously errored out earlier. Added `ZeroFoldingArity` and a test.

Also added the per-round sibling-width check up front, before the transcript, now that the arity is configuration-known. That is what turns the schedule forgery into a shape rejection rather than an incidental Merkle failure.

## Circle changed too

Unavoidable — it shares the FRI error type. It already derived its round count from configuration; it now derives a unit schedule, justified by its own arity rejection, and reports a count mismatch with the count variant instead of abusing the height-mismatch variant to report a count as a pair of heights.

## Fixtures

The proof format changed, so compatibility fixtures were regenerated through the documented ignored generators. The two-adic ones shrank by one and three bytes, one per round. The circle ones grew, because the format change affects them too.

## Left on the table

The commitment vector and the openings vector are still two parallel proof vectors. Merging them into one per-round structure would make the kept openings-count check unrepresentable as well, but that is a larger format change and is deliberately not in this PR.

## Verification

20 test suites pass with 0 failures across `p3-fri`, `p3-circle`, `p3-uni-stark` and `p3-batch-stark`, re-run after rebasing onto #2124. Workspace clippy clean with `-D warnings`; `cargo fmt --all --check` clean.


fri/src/verifier.rs
```diff
@@ -702,28 +806,38 @@ where
...
-///   recover `max_log_arity`, so an unbounded schedule is accepted as long as its lengths
-///   line up.
-/// - `group_indices_by_round` and `rows_by_round` must each be pre-sized to
-///   `commit_phase_commits.len()` entries; this pass indexes them by round.
+/// # Errors
+///
+/// Every shape obligation on proof-controlled input is checked here.
+///
+/// So this pass is total on well-typed input:
+///
```
Nashtare, line 859: **LOW — `fold_query` is public and still takes a caller schedule, but the obligation that bounded that schedule was deleted and replaced by a totality claim it does not meet.**

The base doc said `log_arities` must come through `checked_log_arity`, "which is what bounds each proof-supplied arity to `1..=max_log_arity`". That bullet is gone, and nothing inside checks the bound:

- an entry `>= 64` reaches `let arity = 1 << log_arity` (`:892`) inside the per-round loop, before the `FoldScheduleTooLong` guard, and panics with a shift overflow in debug builds;
- a schedule such as `[usize::MAX, 1]` overflows the `.sum()` at `:905` in debug builds;
- an entry of `0` is accepted as a round that folds nothing.

All three were checked with a debug-build probe. `verify_fri` is unaffected, since its derived entries are at most the two-adicity. The comment just above that sum also still says "the arity schedule is proof-controlled", which is no longer true on the verify path.

Either validate each entry against `1..=max_log_arity` (or at least `< usize::BITS`) and use a checked sum, or restate the caller obligation here, e.g. "entries in `1..=max_log_arity`, as `fold_schedule` produces".

fri/src/verifier.rs
```diff
@@ -411,6 +483,16 @@ where
     if params.log_blowup == 0 {
         return Err(FriError::ZeroBlowup);
     }
+    // Reject a cap that folds nothing away.
+    //
+    // The schedule below walks the height down one round at a time.
+    //
+    // A cap of zero leaves every round folding by zero bits.
```
Nashtare: **LOW (doc) — a zero cap panics rather than hangs.** `compute_log_arity_for_round` opens with `assert!(max_log_arity > 0, ...)` (`config.rs:230-233`), so without this guard `fold_schedule` panics on its first round and never gets to loop. The guard is right; the reason given here, and the `cap 0: 4 -> 4 -> 4 -> ...` sketch in `rejects_with_zero_folding_arity`, describe a failure that cannot happen. Relatedly, `fold_schedule`'s own doc (`config.rs:276-281`, outside this diff) says it returns rather than panics on untrusted heights, and its `# Panics` section does not list the zero-cap assert that made this variant necessary.

fri/src/transcript.rs
```diff
@@ -381,13 +408,111 @@ where
...
+
     /// Replay one commit round against the witness the proof carries.
     ///
+    /// Private because the described round count bounds how often this may run.
+    ///
+    /// Only the entry point that checks a supplied count against it may call it.
+    ///
     /// # Errors
     ///
     /// - The round carries no witness for a described grinding step.
```
Nashtare: **LOW — this branch can no longer be reached outside tests.** `commit_round` is now private and its only caller, `commit_rounds` at `:501`, always passes `Some(witness)` after checking the witness count. So `MissingPowWitness(PowPhase::CommitPhase)` at `:530-533` is dead in production, and `a_commit_round_missing_its_grinding_witness_is_rejected` pins a path no caller can take. Taking `witness: F` would drop the branch, this bullet and that test, and makes the "count checked first" invariant structural.

circle/src/verifier.rs
```diff
@@ -79,12 +79,13 @@ where
     //     H_claim = max claimed log_n + log_blowup
```
Nashtare, line 81: **LOW (doc) — a few comments still describe arities carried by the proof.** Just above, `:70` says "a larger cap would let a proof declare an arity this fold cannot apply", and `UnsupportedFoldingCap`'s doc repeats it (`fri/src/verifier.rs:113`). A proof declares no arity any more; the cap check is now what licenses deriving a unit schedule. The Circle tests carry the same leftovers: `circle/src/pcs.rs:1863-1865` ("`log_arity` lives once per round, not once per query") and `:1846-1848`, which credits the Merkle check with catching round-count tampering that `CommitRoundCountMismatch` now rejects first. All are outside this diff, hence the anchor here.

Nashtare (approved):

fri/src/transcript.rs
tcoratger: Done — `commit_round` takes `witness: F`.

`commit_rounds` is the only caller and checks the witness count against the described round count before absorbing anything, so the invariant is structural now rather than a branch nothing can reach. `MissingPowWitness(PowPhase::CommitPhase)`, its `# Errors` bullet, and `a_commit_round_missing_its_grinding_witness_is_rejected` are all gone.

The variant itself stays for `PowPhase::Query`, whose witness genuinely is optional; its doc now says only the query phase reaches it and why.

fri/src/verifier.rs
tcoratger: Fixed — validated, with a checked sum, and the caller obligation restated for the part this function cannot see.

All three of your probes reach a rejection now instead of a panic or a silent no-op:

```text
    [64]             ->  FoldScheduleArityOutOfRange { round: 0, log_arity: 64 }
    [usize::MAX, 1]  ->  FoldScheduleArityOutOfRange { round: 0, .. }
    [0, 2]           ->  FoldScheduleArityOutOfRange { round: 0, log_arity: 0 }
    [4, 4]           ->  FoldScheduleTooLong { .. }   (unchanged)
```

The per-entry check runs before the sibling loop, so it lands ahead of the `1 << log_arity` at what was `:892`. The bound is `1..usize::BITS`: `max_log_arity` is configuration this function never receives, so claiming it here would be the same overreach the old doc made. That half is back as an explicit caller obligation next to the collector one, naming `fold_schedule` as what already satisfies it.

The total is `fold(0, usize::saturating_add)` rather than `.sum()`. Every entry is under `usize::BITS` by then, but the sum over enough rounds is not; a saturated total exceeds any reachable reduction and falls out of the existing `FoldScheduleTooLong` guard with a truthful value, so no second error variant is needed for it.

The new variant is a structured error, not a panic, so R5 holds on the one public path that takes an untrusted schedule.

`a_caller_schedule_naming_an_unusable_arity_is_rejected` pins all four cases. The stale `// The arity schedule is proof-controlled` comment above the sum is gone; `verify_fri` is unaffected either way, as you said.

fri/src/verifier.rs
tcoratger: Fixed — you are right, `compute_log_arity_for_round` asserts a positive cap, so `fold_schedule` panics on its first round and never gets to loop.

Three places said otherwise:

- the guard's comment in `verify_fri` now says the guard is what stands between a zero cap and that assert, and that a rejection is the right answer either way;
- `ZeroFoldingArity`'s own doc says the same;
- `rejects_with_zero_folding_arity`'s sketch drops `cap 0: 4 -> 4 -> 4 -> ...` for "no round makes progress", and says the walk it described never runs.

And `fold_schedule`'s `# Panics` section lists the zero-cap assert alongside the decreasing-heights one, with a line saying both are configuration rather than proof data. Its "returns rather than panics" sentence is scoped to the untrusted heights it was actually about.

circle/src/verifier.rs
tcoratger: Fixed, all four.

`circle/src/verifier.rs` and `UnsupportedFoldingCap` no longer say a proof declares an arity. Both now say the schedule is derived rather than read, and that the cap check is what licenses deriving a unit schedule at all — any other cap would make that derivation name an arity the fold cannot apply.

In `circle/src/pcs.rs`, the two-unreachable-variants block credited Merkle proofs with catching round-count tampering; it now names `CommitRoundCountMismatch`, which rejects before any folding runs, and says both variants stay reachable in the low-level verifier where the caller supplies its own schedule. The `log_arity` comment drops the "this test used to perform" phrasing and just states that a cross-query arity disagreement is unrepresentable.

PR #2126, opened 2026-09-10
feat(whir)!: drive the hiding WHIR transcript through the pattern player

tcoratger:
> Stacked on #2125 → #2124 → #2122 → #2121. Review bottom-up; this PR's diff is against #2125.

## The finding: a pattern declared but never played

`whir/src/transcript/zk.rs` was already an 861-line, fully worked description of the hiding WHIR transcript — shape types for the masked sumcheck, each round and the base case, producing a pattern and a domain separator covering every count, grinding difficulty and query width.

The only thing that consumed it was the adapter's seeding method, which absorbed the separator and stopped. No driver was ever constructed on this path.

```text
    seeded      ->  domain separation holds, the fingerprint is absorbed
    not played  ->  no step-by-step check, and the description can drift from the code
```

Domain separation was real. The drift detection the pattern player exists to provide was absent, so the 861 lines were only as accurate as whoever last edited them by hand. And the entry point is public, re-exported from the crate root, so this was live API.

#2104 did this for the plain path. This does it for the hiding one.

## The description was wrong in four places

This is the part worth a reviewer's attention. The description was treated as a claim to verify, not as ground truth, and it did not survive contact with a player.

**1. The masked sumcheck batches were described as inline leaf steps.** The child seeds its own sub-transcript and plays its own pattern, so no parent driver could ever play those steps. The module doc admitted this in writing — "no driver plays it", "the delegations to the sumcheck batches carry no markers". Each batch is now a recorded begin/end bracket.

**2. Four of those inline labels did not even match the child's:**

```text
    parent described             child actually plays
    sumcheck_mask_commitment  ->  mask_commitment
    zk_sumcheck_poly          ->  round_poly
    zk_sumcheck_pow           ->  round_pow
    fold_challenge            ->  round_challenge
```

Three others did match, which is exactly what made the drift hard to see by eye. A test now asserts that no step of a delegated phase appears in this description at all.

**3. Turning the batches into brackets would have silently dropped their binding.** The batch round count, grinding difficulty and wire width reached the old seed only through those inline steps. They now travel in the instance label, the way the plain path carries its delegated phases.

**4. The base case's geometry was derived three times independently** — once in the description, once by the prover, once by the verifier. The three agreed today, so this was a live drift risk rather than a present bug. There is now one accessor on the configuration, read by all three.

## The base case became a protocol

It is a standalone construction with its own configuration, error type and tests, and it seeds from a borrowed sponge — but it had no transcript identity. It now has its own version, name, shape and driver pair, and is recorded as a bracket in the run.

That makes it the seventeenth protocol in the cross-protocol separation suite, with a test that it never shares a seed with the run around it.

## Sites remaining raw, and why

Six, all of them **before** the run's seed on either side:

- four in the opening-claim binding, which lands pre-seed on both sides exactly as the plain path does it
- two commitment absorbs, **deliberately deferred**: they mirror the plain path's, and that absorb is being handled in its own PR covering all four mirrors at once (see #2124's description)

Also routed nine sites a method-call grep misses, where the sponge is handed to a helper as an argument: four forwarded query-sampling calls, three child hand-offs, two whole-component hand-offs.

## One pinned constant moved

A base-case rejection test pinned a sampled position at 2. The base case now absorbs a seed of its own, so the position moved to 4. No serialized fixture needed regenerating: the change is confined to the hiding path, and the plain WHIR compatibility fixture still verifies untouched.

## Not done here, deliberately

- **The opening-claim count is not bound into the hiding instance label**, though the plain path binds it. Each claim's point and value is absorbed pre-seed on both sides, the count comes from the public opening protocol, and the verifier already rejects a mismatch against it — so it is bound in substance and moves no pattern step. Adding it is a public signature change, which felt like scope creep. Worth a follow-up for symmetry with the plain path.
- The commitment row width is not bound either; it is a function of the first-round folding factor, and the folding strategy is already bound.

## One plain-path file touched

`whir/src/transcript/mod.rs`, 29 lines: the private delegation helper is now generic over the marker type, so the base-case bracket can name itself, and a now-unused constant moved into that module's test block.

## Verification

| Command | Result |
| --- | --- |
| `cargo test -p p3-whir` | 180 passed, 0 failed, 3 ignored |
| `cargo test --release -p p3-whir --lib pcs::zk -- --ignored` | 2 passed (the two-round 2^17 cases) |
| `cargo test -p p3-multi-stark` | 114 + 17 + 5 + 8 + 9 passed, 0 failed, incl. the plain WHIR compatibility fixture |
| `cargo test -p p3-examples --test transcript_separation` | 4 passed, 0 failed |
| `cargo clippy -p p3-whir --all-targets -- -D warnings` | clean |

All re-run after rebasing onto #2125.


whir/src/transcript/zk.rs
```diff
@@ -490,158 +619,914 @@ impl ZkWhirShape {
...
+            .state
+            .observe_extension::<F, EF, FieldToFieldCodec<F>>(OOD_ANSWER, &answer);
+    }
+
+    /// Replay the grind of the query site of one round.
+    ///
+    /// # Errors
+    ///
+    /// When the witness misses the difficulty the site requires.
+    pub fn query_pow(&mut self, round: usize, witness: F) -> Result<(), TranscriptFailure> {
```
Nashtare, line 1049: **LOW — at zero difficulty this public driver accepts any witness.** `query_pow` returns `Ok(())` when `bits == 0` without looking at `witness`, and `ZkBaseCaseVerifierTranscript::spot_check_pow` (`:1423`) does the same. The in-crate verifiers reject a non-zero witness before calling either, so there is no live bug. But both drivers are re-exported (`p3_whir::transcript::zk::*`), so an external caller driving them directly loses the canonical-witness rule #2105/#2106 established for zero-difficulty sites. Main's player does this check inside the driver (`bits == 0 => witness == F::ZERO`). Moving it here makes the property hold for every caller.

whir/src/transcript/mod.rs
```diff
@@ -1006,6 +1009,12 @@ mod tests {
     /// Opening claims every shape in this module is derived with.
     const CLAIMS: usize = 3;
 
+    /// Step label a delegated sumcheck round carries inside its own description.
+    ///
+    /// No step of a delegated phase belongs to the description built here.
+    /// The label is what one check below looks for and must not find.
+    const FOLD_CHALLENGE: &str = "fold_challenge";
```
Nashtare: **LOW (test) — this label is not one any sumcheck child plays, so the check that uses it cannot fail.** The new doc says `"fold_challenge"` is "the label a delegated sumcheck round carries inside its own description", but the quadratic sumcheck plays `ROUND_CHALLENGE = "round_challenge"` (`sumcheck/src/transcript.rs:83`), as do the hiding and generic-degree ones. So `all(|step| step.label() != FOLD_CHALLENGE)` at `:1307` passes whatever the plain description contains. This is the same `fold_challenge` / `round_challenge` drift the PR description found on the hiding path, one file over. Use `"round_challenge"`, or the stronger form suggested on `zk.rs:1935`.

whir/src/transcript/zk.rs
```diff
@@ -728,6 +1613,388 @@ mod tests {
...
+
+        let expected = 2 + shape.n_rounds();
+        assert_eq!(openers.len(), expected);
+        assert_eq!(openers.len(), closers);
+        assert_eq!(openers[0], INITIAL_FOLD);
+        assert_eq!(openers[openers.len() - 1], BASE_CASE);
+    }
+
+    #[test]
+    fn no_step_of_a_delegated_phase_reaches_this_description() {
```
Nashtare, line 1976: **LOW (test) — a hard-coded blacklist of the child's labels would not have caught the drift this PR fixed.** Against the old description it fails only through the three labels that happened to match (`joint_claim`, `mu_tilde`, `mask_combination`). The four mislabelled steps (`sumcheck_mask_commitment`, `zk_sumcheck_poly`, `zk_sumcheck_pow`, `fold_challenge`) are not in the list, because they never were the child's labels. The strings are also copied rather than taken from `p3-sumcheck`, so a rename in the child silently retires the check, and `claim_batching` is already missing. Asserting the positive form catches all of these: every non-marker label in the run's pattern is one of the run's own step labels.

whir/src/pcs/zk/config.rs
```diff
@@ -318,6 +320,38 @@ where
...
+    ///     prover       ->  runs the phase
+    ///     verifier     ->  replays it
+    ///     description  ->  fixes its transcript
+    /// ```
+    ///
+    /// All three read them here.
+    ///
+    /// None of the three can then drift from the others.
+    #[must_use]
+    pub fn base_case_config(&self) -> BaseCaseZkConfig<F> {
```
Nashtare, line 341: **LOW (API) — this public accessor returns a type outside callers cannot name.** `BaseCaseZkConfig` lives in `pub(crate) mod base_case` (`pcs/zk/mod.rs:59`) and is not re-exported. `ZkBaseCaseShape::new`, which is re-exported from the crate root, takes the same type. So the pair is callable from outside only by chaining, never by writing the type. Either make both `pub(crate)`, or re-export `BaseCaseZkConfig`.

Nashtare (changes requested):

Nashtare (approved):

whir/src/transcript/zk.rs
tcoratger: Agreed on all three counts, and fixed the same way as `mod.rs:1016`.

The blacklist would have caught the drift only through the three labels that happened to coincide, it was copied from `p3-sumcheck` rather than taken from it, and `claim_batching` was already missing. So the assertion is positive: every label in the run's pattern must be one of the run's own fifteen, `OWN_LABELS`, built from the constants themselves.

That catches all four labels this PR fixed — none of them is in the list — and it also catches a mislabelled step of the run's own, which no blacklist ever could.

whir/src/transcript/mod.rs
tcoratger: Fixed with the stronger form, the same one as on `zk.rs:1935`.

You are right that the check could not fail: the quadratic sumcheck plays `round_challenge`, so `all(|step| step.label() != "fold_challenge")` passed whatever the pattern held. Rather than swap in the correct string, the assertion is now positive — every label in the run's pattern must be one of the run's own, taken from the constants:

```rust
const OWN_LABELS: [&str; 13] = [INITIAL_BATCHING, INITIAL_FOLD, COMMITMENT, OOD_POINT, OOD_ANSWER, QUERY_POW, QUERY_INDICES, ROUND_BATCHING, ROUND_FOLD, FINAL_POLY, FINAL_QUERY_POW, FINAL_QUERY_INDICES, FINAL_FOLD];
```

The `#[cfg(test)] const FOLD_CHALLENGE` is gone with it.

whir/src/pcs/zk/config.rs
tcoratger: Took the re-export.

`BaseCaseZkConfig` is now public from `p3_whir::pcs::zk` and from the crate root, alongside `BaseCaseZkError`. Making both `pub(crate)` was the other option, but `ZkBaseCaseProverTranscript` and `ZkBaseCaseVerifierTranscript` are public and take `ZkBaseCaseShape`, so an outside caller driving a standalone base case needs to build that shape — which needs to name this type. Sealing the pair would have left those drivers unusable from outside.

`base_case_config`'s doc says the returned type is nameable from the crate root and can be handed to `ZkBaseCaseShape::new`.

whir/src/transcript/zk.rs
tcoratger: Moved into both drivers, as you suggested.

`query_pow` and `spot_check_pow` now require `witness == F::ZERO` at zero bits before returning `Ok(())`, so the rule holds for any caller of the re-exported drivers, not just the two in-crate verifiers. The run side carries a new `TranscriptFailure::NonCanonicalPowWitness { round }`; the base case already had `BaseCaseZkError::NonCanonicalPowWitness`.

The two pre-checks that used to do this — the round loop in `pcs/zk/verifier/mod.rs` and the `Still check 0` block in `base_case/verifier.rs` — are removed, so the rule has one owner rather than two copies. Neither driver aborts on the new path: `verify` and the base case's `verify` already abort on any `Err`, matching how the existing `observe_pow` failure path is handled right below.

The existing tests still pin both rejections through the public verifiers, so the move is covered end to end.

PR #2127, opened 2026-09-12
fix: address the minor follow-ups from the #2120 review

tcoratger:
The five minor points from the #2120 review. Nothing here changes a proof, a transcript, or a security number — it is the tail of that PR.

### `visit_leaves` and local interaction counts

`validate_successor_columns` walks the constraint DAG and memoises visited nodes by address. A local interaction's count lives behind a `Count` wrapper, so the scan has to clone it out, and the cloned root is a temporary — caching its address would risk skipping a live node later allocated at the same place. #2120 handled that by giving each count a throwaway `BTreeSet`, which is correct but re-walks a count's shared subexpressions once per path through them.

Collecting the owned roots into a `Vec` up front keeps them alive for the whole scan, so they can share the one `seen` set with every other expression. The address-stability precondition is now written on `visit_leaves` itself rather than implied by a call-site comment.

### FRI opening error path

`TwoAdicFriPcs::open` called `transcript.finish()` and only then propagated the FRI prover's result:

```rust
transcript.finish();
Ok((all_opened_values, fri_proof?))
```

The height preflight added in #2120 makes `prove_fri` infallible from this call site, so this is unreachable today. But it is a trap: an error raised after any step had been played would panic inside `finish` instead of returning. The error arm now aborts, which is what uni-STARK, batch-STARK and multi-STARK already do when a run emits no proof. This needed an `abort` on `PcsProverTranscript`, mirroring the ones #2120 added to the other three.

### `StarkAirParams::from_air`

`air.max_constraint_degree().unwrap_or(inferred).max(inferred)` reads as two different defaults; the absent arm is the same value the present arm clamps against. `map_or` says it once, with a line on why the clamp is there.

### Scalar vs packed lookup materialization

`materialize_fraction_scalar` and the packed path are each checked against an analytic row formula, but never against each other. A trace at or above the lane group is where both are defined, so the new test builds one plan over two instances at different heights — one with preprocessed columns, public values and boundary selectors, one with a `next`-row payload — and asserts the two produce identical numerators and denominators.

It also asserts each path really took the branch under test, so the comparison cannot quietly become a tautology. Mutation-checked: replacing the scalar path's `column[(row + 1).min(height - 1)]` with `column[row]` fails it.

### Example trace heights

`trace.height().ilog2()` floors, so a non-power-of-two trace would silently build the security domain one variable short. `log2_strict_usize` states the requirement instead. This moves `p3-util` from the examples' dev-dependencies to its dependencies.

### Verification

`cargo build --workspace --all-targets`, `cargo test` on multi-stark, security, fri, uni-stark and examples, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --all --check` all pass.


Nashtare (approved):
Could have done it in #2120 :) 
But lgtm

tcoratger:
> Could have done it in #2120 :) But lgtm

Yes just wanted to unlock the progress since #2120 has been opened for couple of days now and these ones are really minor so I didn't want you to loose your time with this :)

PR #2128, opened 2026-09-13
perf: optimize Circle and Mersenne-31 hot paths

Nashtare:
This improves hot Circle and Mersenne-31 paths while keeping each optimization independently reviewable.

- Batch-invert selector reciprocals.
- Factor the Circle DEEP quotient with `|v|² = 2 Re(v)`.
- Derive the selector normalization of each Lagrange denominator from its index parity. On any circle domain, `s_p` is `s_p(shift)` on one half-coset and its negation on the other. Debug builds spot-check that the input points alternate between the two halves.
- Specialize PackedQM31 squaring.
- Rotate Mersenne-31 lanes with vector shifts for packed multiplication and division by powers of two on NEON, AVX2, and AVX-512.

On an Apple M4, the relevant benchmarks improved as follows:

- Circle selectors at `2^18`: 132.3 ms to 23.1 ms.
- Circle Lagrange evaluation at `2^18`: 12.20 ms to 9.86 ms.
- DEEP accumulation at `2^18`: 8.83 ms to 4.06 ms.
- PackedQM31 dependent square chain: 995 ns to 814 ns.
- Packed two-power scaling of 1,024 values (NEON): multiplication 2.84 µs to 259 ns; division 4.65 µs to 261 ns.

On an AMD EPYC 9R45 (c8a.4xlarge, Zen 5):

| Packed two-power scaling of 1,024 values | `main` | this PR |
|---|---|---|
| AVX-512 (`-C target-cpu=native`), multiplication | 11.58 µs | 425 ns |
| AVX-512 (`-C target-cpu=native`), division | 3.98 µs | 422 ns |
| AVX2 (`-C target-cpu=x86-64-v3`), multiplication | 18.86 µs | 268 ns |
| AVX2 (`-C target-cpu=x86-64-v3`), division | 4.11 µs | 267 ns |

- DEEP accumulation at `2^18` (AVX-512): 11.37 ms to 5.95 ms.

Validation:

- `cargo test --offline -p p3-circle -p p3-mersenne-31 --lib --quiet` on NEON, AVX2, and AVX-512, in debug builds.
- The Circle PCS tests in `p3-uni-stark` (`fib_air`, `mul_air`, `periodic_air`) and `p3-fri` (`pcs`) in debug builds on NEON, which exercise the parity spot-check.
- `cargo clippy --offline -p p3-circle -p p3-mersenne-31 --lib --tests --benches -- -D warnings`, and an AVX-512 `cargo clippy` build for `p3-mersenne-31`.
- `cargo +nightly fmt --check -p p3-circle -p p3-mersenne-31`


mersenne-31/src/x86_64_avx512/packing.rs
```diff
@@ -162,6 +162,16 @@ impl PrimeCharacteristicRing for PackedMersenne31AVX512 {
         mul_2exp_i::<30, 1>(*self)
     }
 
+    #[inline]
+    fn mul_2exp_u64(&self, exp: u64) -> Self {
+        Self(self.0.map(|x| x.mul_2exp_u64(exp)))
+    }
+
+    #[inline]
+    fn div_2exp_u64(&self, exp: u64) -> Self {
```
tcoratger, line 175: **MAJOR — `div_2exp_u64` regresses 1.79x on AVX-512, and the lane rotation the description promises is 18x faster than what landed here.**

Measured on Zen 5, `-C target-cpu=native` (WIDTH = 16), `Mersenne31Packing/{mul,div}_2exp_u64/1024`:

| | main | this PR | SIMD rotate |
|---|---|---|---|
| `mul_2exp_u64` | 9.86 us | 6.26 us | **0.350 us** |
| `div_2exp_u64` | **3.35 us** | **6.00 us** | **0.350 us** |

AVX2 (WIDTH = 8), same host, for contrast — both arms win there: mul `16.34 -> 1.30 us`, div `3.37 -> 1.35 us`.

**Why the two defaults are not equally slow.** `mul_2exp_u64`'s default is `self * Self::TWO.exp_u64(exp)`, and `Self::TWO` is the *packed* two, so the square-and-multiply chain runs over packed elements — genuinely slow, and any override wins. `div_2exp_u64`'s default is `self * Self::from_prime_subfield(PrimeSubfield::ONE.halve().exp_u64(exp))`: the chain runs in the *scalar* prime field and only the final broadcast-multiply is packed. That is already one packed multiply per element, and sixteen scalar rotates cannot beat it.

**Why 16 lanes lose where 8 win.** `[Mersenne31; 16].map(...)` does not vectorize — the tell is that this PR's AVX-512 `mul` (6.26 us) is 4.8x *slower* than its own AVX2 `mul` (1.30 us) at half the width. The array spills to memory and comes back as sixteen scalar rotates.

**Fix — write the rotation, rather than hoping the autovectorizer finds it.** This is what the PR description already says the code does:

```rust
#[inline]
fn mul_2exp_u64(&self, exp: u64) -> Self {
    rotate_left_31(*self, (exp % 31) as u32)
}

#[inline]
fn div_2exp_u64(&self, exp: u64) -> Self {
    rotate_left_31(*self, ((31 - (exp % 31)) % 31) as u32)
}

/// Rotate every 31-bit lane left by `e < 31`, which is multiplication by `2^e` in M31.
#[inline]
fn rotate_left_31(x: PackedMersenne31AVX512, e: u32) -> PackedMersenne31AVX512 {
    unsafe {
        let v = x.to_vector();
        let left = x86_64::_mm512_and_si512(
            x86_64::_mm512_sll_epi32(v, x86_64::_mm_cvtsi32_si128(e as i32)),
            P,
        );
        let right = x86_64::_mm512_srl_epi32(v, x86_64::_mm_cvtsi32_si128((31 - e) as i32));
        // Safety: each lane is a 31-bit rotation of a value in 0..=P, so it stays in 0..=P.
        PackedMersenne31AVX512::from_vector(x86_64::_mm512_or_si512(left, right))
    }
}
```

`_mm512_sll_epi32` / `_mm512_srl_epi32` take the count in an `__m128i`, so a runtime `exp` is fine. `e` is in `0..=30`, so the counts stay below 32. All 235 `p3-mersenne-31` tests pass with this, including your `packed_two_power_scaling_matches_scalar` — and the redundant `P` encoding survives because `PartialEq` compares canonical values.

I did not measure NEON. At WIDTH = 4 the autovectorizer clearly does its job (your 2.75 us -> 480 ns), so that arm stands as it is; the same explicit rotation would presumably still help, and AVX2 has room too.

Worth noting CI cannot catch this: every AVX-512 job is `build only`.

mersenne-31/src/mersenne_31.rs
```diff
@@ -738,6 +741,31 @@ mod tests {
         assert_eq!(max_canonical, F::new((1 << 31) - 2));
     }
 
+    #[test]
+    fn packed_two_power_scaling_matches_scalar() {
+        type PF = <F as Field>::Packing;
+
+        let values = [
```
tcoratger, line 748: **MINOR (test) — at WIDTH = 4 this fixture never exercises an ordinary field element.**

`values` has 5 entries and lanes are filled with `values[lane % 5]`, so a NEON build (WIDTH = 4) sees lanes `0..3` = `ZERO, ONE, NEG_ONE, P` and drops `0x1234_5678` entirely. Every tested input is then a boundary value, on the very machine the benchmarks were taken on.

At WIDTH = 8 and 16 all five appear, so this only bites the narrow backend.

Cheapest fix is to rotate the starting offset:

```rust
for offset in 0..values.len() {
    let packed = PF::from_fn(|lane| values[(lane + offset) % values.len()]);
    ...
}
```

A `proptest` over random `(value, exp)` pairs would be the stronger version, but the boundary list plus one rotation already covers what matters here.

circle/src/point.rs
```diff
@@ -122,75 +132,34 @@ impl<F: Field> Point<F> {
...
         .par_iter()
-        .zip(&s_p)
-        .map(|(&pt, &s_p)| {
+        .enumerate()
+        .map(|(i, &pt)| {
             let diff = at - pt;
-            let numer = diff.x + F::ONE;
-            let denom = diff.y * s_p;
-            (numer, denom)
+            let s_p = if i & 1 == 0 {
```
tcoratger: **MINOR — the function gained an ordering precondition that nothing enforces.**

`compute_lagrange_den_batched` read `s_p_at_p` off each point, so it was order-agnostic and worked on any slice. This version infers `s_p` from `i & 1`, which is correct only if the caller passes CFFT-ordered points of exactly this domain. Hand it the natural order and it returns wrong denominators, silently — no panic, no wrong length, just a bad opened value.

Guarding it today: a doc sentence and `debug_assert_eq!(points.len(), 1 << domain.log_n)`, which a natural-order slice also satisfies.

Since the invariant is cheap to spot-check, consider making the debug assert actually test it:

```rust
#[cfg(debug_assertions)]
{
    // The parity of the index must select the half-coset, which is what `s_p` keys off.
    let s_p_at_shift = domain.shift.s_p_at_p(domain.log_n);
    for i in [0, 1, points.len() - 2, points.len() - 1] {
        let expected = if i & 1 == 0 { s_p_at_shift } else { -s_p_at_shift };
        debug_assert_eq!(points[i].s_p_at_p(domain.log_n), expected);
    }
}
```

Four points is enough to catch a caller that passed the wrong order, and it costs nothing in release.

Worth mentioning in the doc too: the identity is stronger than "a twin-coset has a common normalization". `s_p_at_p(P, k) = -2^k * (2^{k-1} P).y`, and `2^{k-1} g = 0` for `g = generator(k-1)`, so `2^{k-1} P` collapses to `+/- 2^{k-1} shift` for *any* shift whatsoever. I checked 354 domains including shifts of order `2^31` — it never depends on the domain being a proper twin coset.

circle/src/point.rs
```diff
@@ -105,6 +105,16 @@ impl<F: Field> Point<F> {
         (at - self).to_projective_line().unwrap()
     }
 
+    /// Return the numerator and denominator of the reciprocal selector.
+    ///
+    /// More precisely, if `v_tilde_p(self, at) = denom / numer`, then its reciprocal is
+    /// `numer / denom`. This form lets callers batch-invert only `denom` values.
+    #[inline]
+    pub(crate) fn v_tilde_p_num_den<EF: ExtensionField<F>>(self, at: Point<EF>) -> (EF, EF) {
```
tcoratger: **NIT — the name inverts the roles a reader carries over from `v_tilde_p`.**

The pair is `(numer, denom)` of the *reciprocal*, so `.0` is the denominator of `v_tilde_p` and `.1` is its numerator. The doc says so, but every call site reads `v_tilde_p_num_den(at).1` for a denominator, which is the opposite of what the name suggests one line up.

`recip_v_tilde_p_num_den`, or naming the returns `(denom_of_v_tilde_p, numer_of_v_tilde_p)`, would remove the double-take. Destructuring at the call sites rather than `.0` / `.1` would help too:

```rust
let (num_shift, _) = self.shift.v_tilde_p_num_den(at);
```

circle/benches/deep_quotient.rs
```diff
@@ -0,0 +1,146 @@
...
+    for _ in 0..1 << (log_n - 1) {
+        points.push(coset_0);
+        points.push(coset_1);
+        coset_0 += generator;
+        coset_1 += generator;
+    }
+    points
+}
+
+fn current_geometry(points: &[Point<F>], zeta: Point<EF>) -> (Vec<EF>, Vec<EF>, Vec<EF>) {
```
tcoratger: **NIT — worth saying that these mirror the pre-#2128 implementation.**

`current_geometry` / `current_accumulate` are copies of the code this PR replaces, which makes this a permanent A/B harness — genuinely useful, and the reason the 2.1x number stays checkable after merge. But a future reader will see two spellings of the same thing and wonder which is live. One line above each (`// The pre-#2128 formulation, kept so the factored form stays measurable against it.`) settles it.

Same for `domain_points`, which reimplements `CircleDomain::points` because that is `pub(crate)` — fine, but it can drift silently.

circle/src/deep_quotient.rs
```diff
@@ -52,8 +52,9 @@ pub(crate) fn deep_quotient_vanishing_part<F: ComplexExtendable, EF: ExtensionFi
     // Numerator: Re(1/v_gamma) + alpha^L * Im(1/v_gamma)
     let numerator = re_v_zeta - alpha_pow_width * im_v_zeta;
 
-    // Denominator: |v_gamma|^2 = Re(v_gamma)^2 + Im(v_gamma)^2
-    let denominator = re_v_zeta.square() + im_v_zeta.square();
+    // The circle identity reduces |v_gamma|^2 = Re(v_gamma)^2 + Im(v_gamma)^2
+    // to 2 * Re(v_gamma).
+    let denominator = re_v_zeta.double();
```
tcoratger, line 58: Checked this one carefully because it is the kind of simplification that can move a pole, and it does not.

With `d = x - zeta` on the circle, `v_p = (1 - d.x, -d.y)` and `d.x^2 + d.y^2 = 1`, so `re^2 + im^2 = 2 - 2 d.x = 2 re`. Exact, for `Point<EF>` as much as `Point<F>`, since the group law preserves the circle equation over the extension.

The part worth a comment: `QM31` contains `i`, so in general `re^2 + im^2 = (re + i*im)(re - i*im)` vanishes on a whole line through the origin, not just at `re = im = 0`. Swapping in `2 re` could therefore have narrowed the denominator's zero set and turned a former pole into a finite value. The identity is what rules that out — `re^2 + im^2 = 2 re` forces the two zero sets to coincide, both being exactly `x = zeta`.

Might be worth one line here recording that, since it is the non-obvious half of why the substitution is safe.

circle/src/domain.rs
```diff
@@ -311,9 +312,12 @@ impl<F: ComplexExtendable> PolynomialSpace for CircleDomain<F> {
             .zip(z_vals.par_iter())
             .zip(inv_den_shift.par_iter())
             .zip(inv_den_negshift_k.par_iter())
-            .for_each(|(((((ifr, ilr), itr), &z), &inv_d), &inv_dk)| {
-                let z_inv_dk = z * inv_dk;
-                *ifr = z * inv_d;
+            .zip(pts.par_iter())
+            .for_each(|((((((ifr, ilr), itr), &z), &inv_d), &inv_dk), &at)| {
+                let num_shift = self.shift.v_tilde_p_num_den(at).0;
```
tcoratger: Recomputing `v_tilde_p_num_den` here rather than buffering the numerators from the first pass looks like an oversight, so for the record: I tried buffering it and it is **slower**, 26.5 ms -> 31.1 ms at `log_n = 18`.

Two extra `n`-element buffers are 2 MiB of write-then-read traffic at this size, against eight multiplies saved per point. The loop is memory-bound, so the traffic wins.

I also tried removing the `* k` / `* k_inv` round trip on `is_last_row` — the `k` cancels there and only `is_transition` needs it — and that lands at 27.0 ms, inside the noise.

A short comment here would stop the next reader running the same two experiments:

```rust
// Recomputed rather than carried over from the first pass: two extra n-element
// buffers cost more in memory traffic than the point subtractions they save.
```

tcoratger (approved):
Reviewed all five commits: the two non-trivial algebraic identities, the degenerate-input behaviour, and every benchmark claim. **The math is correct** and I reproduced every speedup independently on a different microarchitecture. One measured regression is worth fixing before you merge; everything else is minor.

### What I verified, and how

**`|v_p(zeta)|^2 = 2 Re(v_p(zeta))`** — holds exactly. With `d = x - zeta` on the circle, `v_p = (1 - d.x, -d.y)`, so

```text
    re^2 + im^2 = (1 - d.x)^2 + d.y^2
                = 1 - 2 d.x + (d.x^2 + d.y^2)
                = 2 (1 - d.x)  =  2 re
```

The degenerate set is unchanged, which is the part worth stating: over an extension containing `i` the old `re^2 + im^2` could vanish at `re = +/- i im` with both parts nonzero, so replacing it with `2 re` could in principle have *moved* the pole. The identity itself rules that out — `re^2 + im^2 = 0` iff `re = 0` iff `x = zeta` — so both forms blow up on exactly the same input.

**Twin-coset sign alternation** — holds, and more generally than the doc claims. Using `sin(2^k t) = 2^k sin t prod_{i<k} cos(2^i t)`:

```text
    v_n_prod(P, k) * P.y = (2^{k-1} P).y / 2^{k-1}
    s_p_at_p(P, k)       = -2^k * (2^{k-1} P).y
```

`2^{log_n - 1} * g = 0` for `g = generator(log_n - 1)`, so `2^{log_n-1} P` is `+/- 2^{log_n-1} shift` depending only on which half-coset `P` is in. That needs nothing of the shift, so the identity holds for **every** `CircleDomain`, not only proper twin cosets. I swept 354 domains (standard, `split_domains`-style, and arbitrary shifts up to order `2^31`): zero mismatches.

And CFFT ordering does preserve the parity, which is what the index trick rests on: `cfft_permute_index(i)` is `reverse_bits(m)` with `m = i/2 < n/2` for even `i` and `m = n - i/2 - 1 >= n/2` for odd `i`, so the top bit of `m` — the low bit after reversal — is exactly `i & 1`.

**`PackedQM31::square`** — `(a0 + a1 u)^2 = (a0^2 + W a1^2) + 2 a0 a1 u` with `W = 2 + i`, and `packed_mul_by_w` is `(2re - im) + (re + 2im) i`. Correct.

**`selectors_at_point` degenerate inputs** — unchanged. `v_tilde_p_num_den` no longer panics at the pole, but there `diff.x = -1` forces `diff.y = 0`, so the batch inversion panics instead. Old and new panic on exactly the same set; only the message moves.

### Benchmarks reproduced

AMD Ryzen 9 9950X3D, `-C target-cpu=native` (AVX-512, WIDTH=16), against `f3ddc2aa`:

| bench | main | this PR | speedup | your M4 |
|---|---|---|---|---|
| `circle/selectors/log_n=18` | 176.8 ms | 26.5 ms | **6.7x** | 5.7x |
| `circle/lagrange/log_n=18` | 22.0 ms | 18.2 ms | 1.21x | 1.24x |
| `deep_quotient/accumulate` | 10.5 ms | 5.0 ms | 2.1x | 2.2x |

All three carry across the architecture, which is the useful signal. The packed two-power scaling does not — see the AVX-512 thread.

### Two further optimizations that do *not* work

I tried both on `selectors_on_coset` and measured them, so nobody re-tries:

- Buffering `num_shift` / `num_negshift` in pass 1 instead of recomputing `v_tilde_p_num_den` in pass 2: **26.5 ms -> 31.1 ms**. At `log_n = 18` the extra `2n` field elements of write-then-read traffic costs more than the eight multiplies it saves. Recomputing is the right call.
- Dropping the `* k` / `* k_inv` round trip on `is_last_row` (the `k` cancels there, and only `is_transition` needs it): **27.0 ms**, i.e. noise. Not worth the extra scalar inverse.

A one-line comment saying recomputation beat buffering would save the next reader the experiment.

---

Approving: nothing here is unsound, and it is a large net win. The AVX-512 item is the one I would fix before merging — it is a regression against `main` on a target CI only ever builds, and the fix is a strict improvement everywhere.

circle/src/point.rs
Nashtare: Added the spot-check in e08c1827: it debug-asserts `s_p_at_p` at indices `0, 1, n - 2, n - 1`, with a `should_panic` test on a reversed slice. The doc now states the precondition and the any-shift derivation.

One correction: natural order does not break it. `CircleDomain::points` is `coset0().interleave(coset1())`, so its even indices are already `shift + <g>`. That is the same parity CFFT order has, and parity is all the function reads. Natural-order input matches the scalar reference on 300+ random-shift domains. What breaks it is an order that does not alternate halves, such as the reversed slice in the test.

circle/src/domain.rs
Nashtare: Added in 062a8647. I re-ran the buffered variant and saw a smaller gap than yours. On an EPYC 9R45 (Zen 5, 16 threads), recompute is 32.3 ms and buffered is 33.2 ms, about 2.6% over two runs each. On an M4 the two are level. So the comment says the buffers cost at least as much as they save, rather than more.

PR #2129, opened 2026-09-13
feat(pcs)!: bind commitments and hiding claims through the typed layer

tcoratger:
Closes the last four raw Fiat-Shamir sites outside the binary-tower track, so every protocol that can be on the typed layer is on it.

## The finding

Every commitment absorb in the workspace was a hand-written mirror. The prover absorbed inside `commit`; the verifier absorbed the same value somewhere else entirely — top of `verify`, or inside multi-STARK's commitment bracket — and nothing but review kept the two in step.

```text
    prover  : commit()          ->  challenger.observe(root)
    verifier: verify()          ->  challenger.observe(root)      a different file
    verifier: main_commitment() ->  challenger.observe(root)      a third one
```

The `MultilinearPcs` doc already spelled the obligation out — *"a verifier never reaches this method, so it absorbs that same commitment in its place"* — which is exactly the drift class this campaign exists to remove: an invariant held by a comment.

The hiding WHIR statement had the same shape, one level up. Its opening claims were absorbed coordinate by coordinate on both sides, so the claim count and the point arity reached no seed at all.

## The commitment

`commit_base` and `Layout::commit` no longer absorb. They encode and commit, and hand the root back. That also fixes the layering: they are shared with `p3-binary-pcs`, whose field is not a transcript field, so the typed phase could never have lived inside them.

Binding becomes a phase both sides call:

| phase | who |
| --- | --- |
| `p3-sumcheck-layout-commitment` | plain WHIR, over the stacked layout |
| `p3-whir-hvzk-commitment` | the hiding pipeline, over its ZK-padded codeword |

`MultilinearPcs` gains that binding as a method, so a caller that must mirror `commit` asks the scheme rather than guessing what it did. multi-STARK's verifier calls it inside the bracket it already had.

Binary tower fields keep a raw absorb — stated once, in `p3-binary-pcs`'s own implementation of that method, instead of buried in a helper three crates away.

## The hiding statement

One phase, `p3-whir-hvzk-claims`, two steps per claim:

```text
    one claim of n coordinates   ->  seed(1 claim, n wide), point, eval
    two claims of n coordinates  ->  seed(2 claims, n wide), point, eval, point, eval
```

The claim count is the step count and the point width is every point step's width, so both reach the fingerprint. Two runs cannot share a seed unless they state the same number of claims over points of the same arity — which is stronger than the instance-label binding the plain path uses, and it is what the follow-up noted on #2126 asked for.

A wrong-arity point stays a structured `ClaimArityMismatch`: the adapter checks arity before binding, so the caller-contract assert inside the phase is unreachable from proof data. R5 holds.

## Grinding sites

`GRINDING_VOCABULARY` compares a described difficulty against the shared budget. Sites outside it are priced by their own protocol's security report — true, but recorded only in a table that had gone stale, still naming `sumcheck_pow` and `zk_sumcheck_pow` after #2126 renamed them, and omitting the base case entirely.

The table is corrected, and the workspace suite asserts that every proof-of-work step it sweeps is **either** compared against the model **or** listed as deliberately unbudgeted. A site in neither is a difficulty nobody compares.

Scope note: this classifies the sites the suite's configurations describe. Extending the shared model to actually price STIR's and WHIR's per-round grinds is a paper-driven job, not a table edit, and is deliberately not attempted here.

## Verification

| | |
| --- | --- |
| raw production absorbs outside the binary track | **8 → 0** |
| protocols in the separation suite | 21 → 24 |
| mutation-checked | dropping either new bind fails existing round trips |

Both new bindings were confirmed load-bearing by deleting them and watching honest round trips fail, not by inspection.

The suite now allows a protocol to contribute fewer than three configurations: a phase with no configuration at all has exactly one seed to offer, and the old fixed product would have demanded two duplicates of it and failed its own pairwise check.

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo +nightly fmt --check` are clean. The multi-STARK WHIR fixture is regenerated; **no other fixture moves**, which is what confines this to the WHIR and layout paths.

## Breaking

A transcript-format break for both WHIR pipelines and any `MultilinearPcs` implementation outside this repo, which must now supply the commitment-binding method.


commit/src/pcs/multilinear.rs
```diff
@@ -58,36 +58,73 @@ where
...
+    /// It calls this method instead, at the same point in the sponge stream.
+    ///
+    /// ```text
+    ///     prover  : commit(..)  ->  binds the root it produced
+    ///     verifier: this method ->  binds the root it was handed
+    /// ```
+    ///
+    /// # Soundness
+    ///
+    /// Both sides reach the binding through this one method.
```
Nashtare: **LOW — The soundness paragraph says both sides reach the binding through `observe_commitment`, but on the prover side two of the three impls never call it.**

- `BinaryPcs`: the prover binds through its own raw `challenger.observe(commitment.clone())` at `binary-pcs/src/prover.rs:102`. The method at `binary-pcs/src/pcs.rs:454` is reached only by the verifier.
- Hiding WHIR: the prover calls the free `transcript::zk::observe_commitment` directly, at `whir/src/pcs/zk/prover/mod.rs:119`.
- Plain WHIR is the only impl whose `commit` calls `self.observe_commitment` (`whir/src/pcs/adapter.rs:106`).

The two sides agree today; every mutation that drops one side fails. But the property the doc names ("neither can then drift") is not enforced by any code. It is the same comment-held invariant the PR sets out to remove, one level down.

Suggested fix, either of:

- Route every impl's `commit` through `self.observe_commitment(&commitment, challenger)`.
- Or add one generic conformance check per impl: run `commit` on challenger A, run `observe_commitment` on a fresh challenger B, and assert equal samples. Then soften the sentence to say the method is the reference, and `commit` must match it.

binary-pcs/src/prover.rs
```diff
@@ -91,14 +91,15 @@ where
...
-    );
+    let (layout, commitment, merkle_data) =
+        PcsLayout::commit(encoder, mmcs, witness, FOLDING, config.log_inv_rate());
+
+    // A binary tower field is not a transcript field.
+    //
+    // So the root is bound here by hand, with no typed phase.
+    //
+    // The verifier binds it in the same position, at the top of its own entry point.
+    challenger.observe(commitment.clone());
```
Nashtare: **LOW — The binary raw absorb is written twice, not "stated once" as the PR description says.**

This line and `binary-pcs/src/pcs.rs:458` are two separate copies of `challenger.observe(commitment.clone())`. Both are correct and consistent: dropping either one fails 19–22 binary-pcs tests.

The PR description, though, says the raw absorb is stated once, in `BinaryPcs`'s implementation of the method.

The comment here also says the verifier binds "at the top of its own entry point". On the multi-STARK prescribed path the verifier actually binds inside `multi-stark/src/verifier.rs`'s `main_commitment` bracket, not at the top of a `BinaryPcs` entry point.

Suggested fix: make the `MultilinearPcs::commit` impl call `self.observe_commitment` after the free `commit` returns, and drop the observe from the free function. Also reword the comment.

whir/src/pcs/zk/prover/mod.rs
```diff
@@ -109,7 +112,11 @@ where
         let padded = zk_padded_matrix(message.as_slice(), &randomness, folding, height);
         let encoded = self.dft.dft_batch(padded).to_row_major_matrix();
         let (commitment, merkle) = self.mmcs.commit_matrix(encoded);
-        challenger.observe(commitment.clone());
+
+        // The verifier binds the same root, through the same phase.
+        //
+        // It does so before replaying anything else.
+        observe_commitment::<F, _, _>(challenger, commitment.clone());
```
Nashtare: **LOW — Removing the hiding commitment binding on both sides, or the hiding claims binding on both sides, passes every p3-whir test.**

Probes:

- Deleting this call and the verify-side `self.observe_commitment` in `whir/src/pcs/zk/adapter.rs:150`: 188/188 pass.
- Deleting both `observe_claims` calls in `whir/src/pcs/zk/adapter.rs`: 188/188 pass.

The plain path is protected by `verify_whir_compat_fixture`: its symmetric removal fails there. Hiding WHIR has no fixture, and no test ties the adapter's transcript to the new phases. The unit tests in `transcript/zk.rs` exercise only the free functions.

A symmetric removal of the claims binding would drop the whole statement out of the transcript. That is a real soundness regression, and nothing would flag it.

Suggested fix: add a transcript-conformance test in `whir/src/pcs/zk/tests.rs`.

1. Run `pcs.commit(witness, &mut a)`.
2. Run `transcript::zk::observe_commitment::<F, _, _>(&mut b, commitment)` on a fresh challenger.
3. Assert equal samples.
4. Do the same for `open`'s claim prefix, for example by replaying `observe_claims` and checking the first challenge the run draws.

security/src/grinding.rs
```diff
@@ -275,13 +275,21 @@ pub struct GrindingStep {
 ///
 /// | protocol                     | steps it grinds                                              |
 /// | ---------------------------- | ------------------------------------------------------------ |
-/// | `p3-whir`                    | `sumcheck_pow`, `query_pow`, `final_query_pow`               |
-/// | `p3-whir-hvzk`               | `zk_sumcheck_pow`, `base_pow`                                |
+/// | `p3-whir`                    | `query_pow`, `final_query_pow`                               |
+/// | `p3-whir-hvzk`               | `query_pow`, `final_query_pow`                               |
```
Nashtare: **LOW — `p3-whir-hvzk` has no `final_query_pow` step.**

- The hiding pattern pushes `QUERY_POW` once per round (`whir/src/transcript/zk.rs:275`), and `BASE_POW` only in the base-case phase (`:475`).
- `FINAL_QUERY_POW` is imported in `zk.rs` only inside the `tests` module (`:1781`).
- `UNBUDGETED_GRINDING_SITES` in the suite already, and correctly, omits it.

Suggested fix: make this row `query_pow` only.

examples/tests/transcript_separation.rs
```diff
@@ -926,6 +983,87 @@ fn no_two_configurations_of_any_two_protocols_share_a_seed() {
...
+/// Grinding sites the shared budget deliberately does not compare.
+///
+/// A protocol here credits its own grinding inside its own security report.
+///
+/// It does not go through the budget every FRI-backed STARK shares.
+///
+/// The entry is the acknowledgement.
+///
+/// A site is either compared against the model, or listed here on purpose.
+const UNBUDGETED_GRINDING_SITES: [(&str, &str); 5] = [
```
Nashtare: **LOW — The "every site is classified" test covers 5 of the 12 sites the `grinding.rs` table names, and passes only because the other sweeps use zero difficulty.**

Missing from this list:

- `p3-stir`: `folding_pow`, `query_pow`, `final_folding_pow`, `final_pow`
- `p3-sumcheck-quadratic`: `round_pow`
- `p3-sumcheck-generic-degree`: `round_pow`

Those grinds are real in production, but `stir_cases`, `sumcheck_quadratic_cases` and `sumcheck_generic_degree_cases` sweep `pow_bits = 0`. `push_pow` elides a zero-bit step, so the walk never sees them.

Probe: changing `sumcheck_quadratic_cases`'s `plain` to `SumcheckShape::new(4, 1, Basis::Evaluation)` makes `every_grinding_site_is_either_budgeted_or_listed_as_unbudgeted` fail with `p3-sumcheck-quadratic/round_pow grinds, but no vocabulary classifies it`.

So the test does fail on an unlisted site it can see. But the doc table and this list are two sources of truth that already disagree.

Suggested fix, either of:

- Give each grinding protocol one nonzero-pow configuration and list the missing sites here.
- Better: export the "priced elsewhere" table from `p3-security` as a const, and assert against that, so the doc and the test cannot diverge.

examples/tests/transcript_separation.rs
```diff
@@ -926,6 +983,87 @@ fn no_two_configurations_of_any_two_protocols_share_a_seed() {
...
+    // Every protocol contributes at least one case and at most the sweep budget.
+    //
+    // A configuration-free phase has one seed to offer.
+    //
+    // A fixed product would demand two duplicates of it.
+    //
+    // The pairwise check below would then fail on its own fixtures.
+    let groups = protocols();
+    assert_eq!(groups.len(), NUM_PROTOCOLS);
+    assert!(groups.iter().all(|group| !group.is_empty()));
```
Nashtare: **LOW — The relaxed count check lets any protocol silently shrink its sweep to one configuration.**

Before, `seeds.len() == 3 * 21` pinned every protocol to three cases. Now `!is_empty()` plus `len() <= 3` accepts any protocol dropping to one case. For example, deleting two entries from `whir_cases` still passes, and the per-knob seed coverage that sweep exists for disappears unnoticed.

Suggested fix: keep the exact check, with an explicit allowlist of the configuration-free phases (`p3-sumcheck-layout-commitment`, `p3-whir-hvzk-commitment`):

```rust
assert_eq!(group.len(), if CONFIG_FREE.contains(&name) { 1 } else { MAX_CASES_PER_PROTOCOL })
```

whir/src/pcs/zk/adapter.rs
```diff
@@ -154,13 +159,24 @@ where
...
-                (point, eval)
-            })
             .collect();
 
+        // The statement is the caller's, so its arity is checked before it is bound.
+        //
+        //     point arity != committed arity  ->  error, never a panic
+        //
+        // Binding first would describe a step width the point cannot fill.
+        for (claim, (point, _)) in claims.iter().enumerate() {
```
Nashtare: **LOW (nit) — The arity check is now duplicated, and the comment above the `collect` is stale.**

- This loop is identical to the one in `HidingWhirVerifier::verify` (`whir/src/pcs/zk/verifier/mod.rs:225`). Through the adapter that one is now unreachable; it still guards direct callers.
- `// Bind the public claims exactly as the prover did.` at `adapter.rs:158` now sits above a plain `collect()`. The binding happens eleven lines later.

Suggested fix: reword `:158` to describe the pairing. Optionally factor the arity check into one helper that both sites call.

Error precedence also changed. A statement that is both wrong-arity and over the claim budget used to report `InitialClaimsBelowTarget`; it now reports `ClaimArityMismatch`. That is harmless, but worth knowing.

Nashtare (approved):
Reviewed at 014d3e7d. No HIGH or MEDIUM findings.

**Bindings**
- I traced every caller of `commit_base` and `Layout::commit`. No absorb is dropped on any path: plain WHIR, hiding WHIR, `BinaryPcs`, and the multi-STARK prover and verifier.
- `observe_commitment` is a required method, so no impl can silently inherit a default.
- The claim count and arity both reach the hiding-claims seed, and the adapter's arity check keeps a wrong-arity point a structured error.

**Mutations**
- Removing any binding on one side fails tests (19–30 failures each).
- Removing the plain binding on both sides is caught by the WHIR compat fixture.
- Removing the hiding commitment binding on **both** sides passes all 188 p3-whir tests, and so does removing the hiding claims binding on both sides. See the comment on `whir/src/pcs/zk/prover/mod.rs`.

The inline comments are LOWs: the trait doc's accuracy, the grinding table and the site-classification test, and the relaxed per-protocol count check.

**Validation**
- 977 tests pass across whir, sumcheck, commit, binary-pcs, multi-stark, binary-dft and security.
- The separation suite passes.
- The branch merges cleanly with main.

examples/tests/transcript_separation.rs
tcoratger: Took the second option, and your probe was the right diagnosis: the sweep grinds at zero bits, `push_pow` elides a zero-bit step, so the walk could not see seven of the twelve sites.

- The doc table becomes `UNPRICED_GRINDING_SITES` in `p3-security`, and `every_grinding_site_is_either_budgeted_or_priced_elsewhere` classifies against that plus the vocabulary. Two sources of truth become one.
- Auditing it against the code turned up the two you named, `p3-sumcheck-quadratic` and `p3-sumcheck-generic-degree`, both of which really do push `ROUND_POW`. They are listed now, so the table has eleven entries.
- A second test, `every_unpriced_grinding_site_is_described_by_the_protocol_that_owns_it`, walks a sweep built specifically so every grinding protocol grinds at a positive difficulty, and fails on any listed site no pattern describes. Confirmed by adding a fake `p3-stir/ghost_pow` row: it fails with exactly that message.

So a stale row and an unclassified site are now both build failures, from opposite directions.

security/src/grinding.rs
tcoratger: Correct, and the row was simply wrong. `push_pow` is reached twice in the hiding path, `QUERY_POW` per round and `BASE_POW` in the base-case phase; `FINAL_QUERY_POW` is only ever imported by that module's tests.

Rather than fix the row, the table is now data: `UNPRICED_GRINDING_SITES`, exported from `p3-security` beside the vocabulary, and the suite asserts against it. The doc and the test can no longer disagree, which was the real defect.

whir/src/pcs/zk/adapter.rs
tcoratger: Both fixed, and the duplication is gone rather than documented.

`check_claim_arity` is one function in `pcs/zk/verifier/mod.rs` that the adapter and `HidingWhirVerifier::verify` both call. Keeping the check at both sites is deliberate — the adapter must reject before binding, and the verifier is a public entry point a caller can reach directly — but there is now one copy of the logic, so they cannot drift. The comment at the verifier site says exactly that.

`adapter.rs:158` now says it pairs each requested point with the value the proof claims at it, and the binding comment sits on the `observe_claims` call where the binding happens.

Noted on the error precedence. `ClaimArityMismatch` before `InitialClaimsBelowTarget` is the ordering I would pick anyway: a statement whose points do not fit the commitment is malformed regardless of how many of them there are.

commit/src/pcs/multilinear.rs
tcoratger: Took the first option: every impl's `commit` now calls `self.observe_commitment`, so the property is structural rather than stated.

- `BinaryPcs`: the raw absorb moves out of the free `prover::commit`, which now returns the root unbound and no longer takes a challenger at all. The trait impl binds.
- Hiding WHIR: the phase moves out of `HidingWhirProver::commit`, which mirrors `Layout::commit` and returns the root unbound. The adapter binds.
- Plain WHIR was already right.

And the second option too, since the doc should not be the only thing asserting it. One conformance test per impl, in `whir/src/pcs/tests.rs`, `whir/src/pcs/zk/tests.rs` and `binary-pcs/src/test_util.rs`:

```text
    commit(witness, a)            -> a
    observe_commitment(root, b)   -> b
    a and b must sample alike
```

An impl that binds inside its commit phase instead of through the method now fails that test rather than relying on review.

whir/src/pcs/zk/prover/mod.rs
tcoratger: This was the one worth catching. Both symmetric removals now fail.

`whir/src/pcs/zk/tests.rs` gains two tests:

- `the_commit_phase_binds_exactly_what_the_binding_method_binds` — commit on one sponge, `observe_commitment` on a fresh one, assert equal samples. Deleting the binding from both sides leaves the method still binding, so the two part.
- `the_open_phase_binds_the_statement_and_nothing_else` — rebuild the run from its pieces and land on the same sponge state the adapter does:

```text
    adapter : commit, then open(data, points, a)          -> a
    replay  : observe_commitment, observe_claims, prove   -> b
    a and b must sample alike
```

The replay derives the two masking generators the way the adapter does, `StdRng::from_rng` twice off the stored one, so the comparison is exact rather than approximate.

Verified against your probes: deleting both `observe_claims` calls fails the second test, deleting both commitment bindings fails the first. Before this commit each passed 188/188.

binary-pcs/src/prover.rs
tcoratger: Right on both counts, and both are gone.

The free `prover::commit` no longer absorbs, and no longer takes a challenger — same contract as `commit_base` and `Layout::commit` after this PR. The single raw absorb now lives in `BinaryPcs::observe_commitment`, which `BinaryPcs::commit` calls. So the PR description's claim is true now instead of aspirational.

The comment is reworded: it no longer claims the verifier binds "at the top of its own entry point", since on the multi-STARK prescribed path it binds inside that bracket.

Two internal tests in `verifier.rs` called the free `commit` and then continued on the same challenger. They clone into the verifier challenger afterwards, so the binding was never load-bearing there, but they now mirror the production absorb explicitly so the replay walks the same stream.

examples/tests/transcript_separation.rs
tcoratger: Agreed, and taken as suggested.

`CONFIGURATION_FREE_PHASES` lists the two commitment phases, and the count is exact again:

```rust
let expected = if CONFIGURATION_FREE_PHASES.contains(&protocol) { 1 } else { MAX_CASES_PER_PROTOCOL };
assert_eq!(group.len(), expected, "{protocol} sweeps {} cases", group.len());
```

Deleting two entries from `whir_cases` now fails with the protocol named, rather than passing silently and taking its per-knob coverage with it.

PR #2130, opened 2026-09-14
fix(matrix): make vertically packed row wrapping overflow-safe

ginavalent:
## Summary

Fix overflow in vertically packed matrix access when the requested row or step is near `usize::MAX`.

## Bug

The `Matrix` API documents that vertically packed rows wrap around when the requested row range exceeds the matrix height. Several implementations checked the fast path with expressions such as `r + P::WIDTH` and `r + step`.

For large valid row indices, these additions can overflow in debug builds. In release builds, the wrapped value can also make the implementation select the wrong fast path. Using machine-word wrapping before `% height` is not mathematically correct for non-power-of-two heights.

## Cause

The fast-path checks and row-index calculations performed unchecked integer addition before applying the matrix-height wraparound.

## Fix

- Add a shared overflow-safe modular row-index helper.
- Use it for wrapped row and pair access.
- Restrict the contiguous fast path to ranges that are provably within bounds without overflow.
- Apply the same logic to dense matrices, row-index mapped views, horizontal truncation views, and horizontal stack views.

## Tests

Added regression tests using:

- `usize::MAX` as the starting row;
- a non-power-of-two matrix height;
- both single vertically packed rows and row pairs;
- dense and row-index mapped matrices.

Validated with:

- `cargo +stable test -p p3-matrix`
- `cargo +stable test -p p3-matrix --features parallel`
- `cargo +stable clippy -p p3-matrix --all-targets -- -D warnings`
- dependent `p3-dft`, `p3-merkle-tree`, and `p3-multi-stark` test suites


Nashtare:
I don't think Plonky3 needs it, because none of its own code can trigger the bug. If you believe it does, can you show us a PoC?

PR #2131, opened 2026-09-14
fix(security): charge the budget's DEEP batching round over the LDE domain

Nashtare:
## Summary

`p3_security::budget` graded the DEEP-quotient batching round as `k / |E|`, independent of the evaluation domain. The batched quotient is a random linear combination of words handed to the low-degree test, so it pays the proximity-gap error `(k − 1) · n / |E|`. That is the bound the same module already charges the FRI folding round, and the one `stark::conjectured_security_report` charges its opening-batching term. The round was overstated by `log2(n)` bits: 23 bits at `2^20` and 32 bits at `2^29` with blowup 8.

## Changes

- **`budget::security_report`**: the DEEP-composition round's size term is now `log_max_height + log_blowup`, the same domain the folding round is graded over. The coefficient stays `num_deep_terms`, which stands in for `k − 1`: it keeps the coefficient nonzero and is conservative by less than one bit.
- **Docs**: the module doc and the round's comment describe the proximity-gap grading. `ProtocolParams::lookup_pow_bits` no longer claims the lookup site is the only one whose absence degrades with instance size.
- **Tests**:
  - `budget_parity`'s DEEP-composition reference is now `SecurityAssumption::UniqueDecoding.prox_gaps_error`, the term `stark.rs` uses, with a `log2(k / (k − 1))` tolerance for the coefficient.
  - New unit test `deep_composition_round_pays_the_lde_domain_size`: the round loses exactly one bit when either the trace height or the blowup doubles.

## Impact

- **`uni-stark` and `batch-stark` reporting is unchanged.** Those crates grade through `StarkSecurityParams` and `stark::*_security_report`, which already charged this round, so example output is identical.
- **Callers of `budget::security_report` see a lower `deep-composition` term at every height.** The reported level moves only where that term becomes the minimum. For roughly 130 batched terms over a ~128-bit field with blowup 8, an ungrounded round drops from about 121 bits to about 98 bits at `2^20` and 89 bits at `2^29`. With 12 bits of `deep_pow_bits` it stays at or above 101 bits up to `2^29`. Grinding before the DEEP challenge is therefore load-bearing at large heights.
- **Fixed-point mirrors of the budget**, for example in recursive verifiers, need the same `log_max_height + log_blowup` term to stay bit-for-bit consistent.

## Files Changed

| File | Change |
|---|---|
| `security/src/budget/mod.rs` | Modified: round size term, docs, new unit test |
| `security/src/budget/shape.rs` | Modified: `lookup_pow_bits` doc |
| `security/tests/budget_parity.rs` | Modified: DEEP-composition reference and tolerance |

## Testing

- `cargo test -p p3-security --release`: 106 lib tests and 3 parity tests pass.
- Both the updated parity check and the new unit test fail without the fix. The parity check reports `deep-composition: 127.99997 > 126.75 at h=6`, i.e. the budget claimed more bits than the reference.
- `cargo +nightly fmt --check`, `cargo clippy -p p3-security --all-targets -- -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc -p p3-security --no-deps` are clean.

## Related Issues

None.


security/src/budget/shape.rs
```diff
@@ -32,8 +32,8 @@ pub struct ProtocolParams {
     pub folding_pow_bits: u32,
     /// Grinding bits before the lookup / permutation argument's challenges are sampled.
     ///
-    /// The lookup round's error grows linearly in trace length, so this is the only grinding site
-    /// whose absence degrades with instance size.
+    /// The lookup round's error grows linearly in trace length, so its absence degrades with
+    /// instance size.
```
tcoratger, line 36: **MINOR — this correctly stops claiming "only", but nothing now tells a caller which sites do degrade.**

Removing "the only grinding site" was right: after this change the DEEP round degrades with instance size too. But the field a caller actually reads when picking a value, `deep_pow_bits` six lines up, still says only

```rust
/// Grinding bits before the DEEP-composition challenge is sampled. Read only when
/// [`AirShape::num_deep_terms`] is `Some`.
```

Someone sizing parameters from these docs gets no signal that leaving it at zero now costs `log_max_height + log_blowup` bits — which is the operational conclusion your own description draws ("Grinding before the DEEP challenge is therefore load-bearing at large heights"). That sentence deserves to be here rather than only in the PR body.

Worth noting `folding_pow_bits` has always had the same property — its round was already graded over the LDE domain — and has never said so either. So the count is three sites that degrade with instance size, of which one mentions it, which reads as if it were the special one.

Either say it on all three, or drop it from `lookup_pow_bits` and put one line in the module doc listing which rounds carry a domain-size factor. The second is probably less to keep in sync.

security/src/budget/mod.rs
```diff
@@ -139,15 +145,19 @@ pub const fn security_report(
...
-    // further challenges, giving a univariate whose degree is the number of batched terms.
+    // further challenges into the one word the low-degree test runs on. That is a random linear
+    // combination of words over the LDE domain, the same kind of round as folding, so it pays the
+    // proximity-gap error `(k − 1) · n / |E|` over the `k` batched terms: the conjecture removes
+    // the list-size multiplier, not the domain size. `k` stands in for `k − 1`, which keeps the
+    // coefficient nonzero and costs less than one bit.
     let deep_composition = round(
         DEEP_COMPOSITION_LABEL,
         instance,
         match air.num_deep_terms {
```
tcoratger, line 156: **MINOR — `Some(1)` now diverges from `stark::batching_term` by the full domain size.**

`stark::batching_term` refuses the round outright below two functions:

```rust
let num_functions = shape.num_batched_functions;
if num_functions < 2 {
    return None;
}
```

The reasoning is that with one word there is no combination challenge, so there is no round to charge — and `prox_gaps_error` asserts `num_functions >= 2` for the same reason.

`round` here takes `Some(1)` happily: `ceil_log2(1) = 0`, so the term becomes `field_bits − (log_max_height + log_blowup) + pow`. Before this change that cost about nothing; now it is a real height-dependent penalty on a protocol that batches a single word, where the true error is zero.

Not reachable for any real AIR — `num_deep_terms` is columns times out-of-domain points — and the parity test cannot see it, since both fixtures sit at `k = 20` and `k = 770`. But it is a corner where the two modules' answers to the same question moved from agreeing to differing by about 23 bits at `2^20`.

Either fold it into the `None` arm:

```rust
match air.num_deep_terms {
    Some(n) if n >= 2 => Some(n as u64),
    _ => None,
}
```

or say on `AirShape::num_deep_terms` that a protocol batching fewer than two words passes `None`, since that is the convention `stark.rs` already follows.

security/src/budget/mod.rs
```diff
@@ -139,15 +145,19 @@ pub const fn security_report(
     );
 
     // The DEEP quotient batches every committed column and out-of-domain point by powers of two
-    // further challenges, giving a univariate whose degree is the number of batched terms.
+    // further challenges into the one word the low-degree test runs on. That is a random linear
+    // combination of words over the LDE domain, the same kind of round as folding, so it pays the
+    // proximity-gap error `(k − 1) · n / |E|` over the `k` batched terms: the conjecture removes
```
tcoratger, line 150: **NIT — `n` arrives undefined, one line after the module states a different letter for the same thing.**

The module doc frames every round as `error ≤ coefficient · size / |E|`, and the folding note above uses `n` for the LDE domain without introducing it either. So `(k − 1) · n / |E|` is consistent with the neighbourhood, but a reader arriving here has just been told the size slot is called `size` and has to infer that `n` is that slot while `|E|` is still the field.

One clause fixes it:

```text
proximity-gap error `(k − 1) · n / |E|` over the `k` batched terms, with `n` the LDE domain
```

Same applies to the folding note at the top of the module, if you want them to match.

tcoratger (approved):
The fix is right, and it is the conservative direction. Approving — the three notes below are doc-level and a corner case, none of them worth holding this.

### What I checked

**The bound.** The batching step is not a Schwartz-Zippel identity in `alpha`. The binding event is "a word far from RS combines into a word close to RS", which is exactly what the proximity-gap theorem bounds, and in the unique-decoding regime that is `(k − 1)·|L| / |F|` — [BCI+20] Theorem 1.2, with the `|L|` factor the old grading dropped. The old comment's framing ("a univariate whose degree is the number of batched terms") bounds a different event: it bounds the claimed values being inconsistent, not the distance argument the low-degree test actually rests on.

So the round was overstated by `log2|L| = log_max_height + log_blowup`. Your 23 bits at `2^20` and 32 at `2^29` with blowup 8 check out, as does the 121 → 98 figure at `k ≈ 130` and the 101-bit floor with 12 bits of grinding.

**Internal consistency.** This is the part that convinces me most. `SecurityAssumption::UniqueDecoding.prox_gaps_error` already implements `field_bits − (log_degree + log_inv_rate) − log2(k − 1)`, and `stark::conjectured_security_report` already charges the same round at that bound, with a doc that spells out this exact argument:

> Dropping the opening RLC while keeping the folding RLC would overstate security for an instance that batches many codewords, which is every real STARK.

The budget module was doing the softer version of that mistake — charging the round, but at a bound that omits the domain size. After this it agrees with `stark.rs` and with its own folding round, which was already graded over `log_max_height + log_blowup`.

**Direction.** Reported levels only fall, never rise, so no configuration previously refused becomes accepted.

**The tests.** I reverted the size term to `0` locally and reproduced both failures, including the parity message verbatim (`deep-composition: 127.99997 > 126.75 at h=6`).

The parity tolerance is tight rather than slack, which is easy to get wrong here: `fixed::ceil_log2` rounds up by one ulp of a 16-bit fraction, not to a whole bit, so the gap really is `log2(k/(k − 1))` and not that plus up to a bit. At `k = 20` that is 0.074 against a measured gap of 0.074. Had `ceil_log2` been whole-bit, the same tolerance would have been about ten times too small and the test would have been a flake waiting to happen.

The unit test's `base < cap()` guard is the other thing worth calling out — without it the assertion would pass vacuously the moment the term clamped, which is the usual way a test like this rots.

**Blast radius.** `budget::security_report` has no consumer in this repo outside `security`'s own tests, so "uni-stark and batch-stark unchanged" is stronger than stated — nothing in-tree reads it at all. The impact is entirely on external fixed-point mirrors, which your description already flags.

**Arithmetic.** `saturating_add` mirrors the folding round exactly, and `from_bits(u32::MAX)` at `FRACTIONAL_BITS = 16` is about `2.8e14`, so there is no new overflow surface.

tcoratger:
All of the comments are super minor, not even relevant, so I kept them for record but I merged :)

PR #2132, opened 2026-09-14
fix(multi-stark): draw fractional-GKR round polynomials at the field's own nodes

tcoratger:
A fractional-GKR round polynomial is cubic. The prover sends it at four
interpolation nodes and the verifier rebuilds it there from
`Field::interpolation_node`.

The prover did not use those nodes. It reached the third by adding the line's
step to itself, and named the fourth with an integer literal:

```rust
let n0 = n0_hi + (n0_hi - n0_lo);                       // meant to be node 2
let q3 = interpolator.eval(&evals, q_sum, EF::from_u8(3));
let equality_2 = equality_0 + EF::TWO * equality_step;
```

Both assume the nodes are the integers zero through three. That holds wherever
the field enumerates its nodes that way, which is odd characteristic, extensions
included. Over a binary tower it does not:

- doubling is zero, so `n0_hi + (n0_hi - n0_lo)` folds back onto `n0_lo` and the
  second transmitted evaluation repeats the first;
- `from_u8(3)` is one, which is already node one;
- `EF::TWO` is zero, so the equality factor at node two silently becomes the
  factor at zero.

Two of the four values then describe the same point, and the message stops
pinning the polynomial down.

## The fix

Name every node the way the verifier already does, and scale the line's step by
the node rather than assuming one step lands on it.

Scaling is only needed where stepping misses. A field that enumerates its nodes
as the integers is already at the second node after one step, so that walk is
kept: `scaled_node` returns `None` there and the fold runs main's body, with the
numerators in their own type and the gate's products still mixed. Only a binary
tower takes the scaled arm, where the numerators widen into the extension.

The choice is made once per layer rather than per leaf. Without that split the
prime-field path — the only one in use today — paid about 15% (thanks to
@Nashtare for measuring it).

## Reachability

This is latent, not live. `LookupPlan::build` rejects characteristic two before
the lookup argument reaches this code, so nothing in the crate hits it today —
only a direct caller of the public `prove_fractional_gkr` does. It is a
prerequisite for using the reduction over a binary field at all.

## Testing

`binary_round_trip` reduces a random zero-sum table at every layer count from one
to eight, over both `PolyMaybePacked::Scalar` and `Packed`, checking the verifier
accepts and that both openings are the tables themselves at the point reached. It
runs twice:

- `BinaryField128 / BinaryField128`, where the numerator and denominator share a
  type;
- `BinaryField32 / BinaryField128`, the mixed walk, which is the only test here
  that reaches `mixed_round_polys` with a narrow numerator over a binary field —
  so it also guards the fast-path split above.

Both fail on main and pass here.

I checked the tests are load-bearing: reverting only `prover.rs` makes them fail
while every other fractional-GKR test, all of them over `BabyBear`, still passes.
That divergence is the bug's exact signature.

`cargo test -p p3-multi-stark` passes (129 lib tests), as do clippy, rustfmt and
`scripts/check.py lint`.


multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -72,20 +73,24 @@ where
...
+                acc[0] += eq_suffix * (d1_lo * n0_lo + d0_lo * n1_lo + lambda * d0_lo * d1_lo);
+
+                // The gate at the domain's other node walks each line out to it.
+                //
+                // The step is a multiplication rather than a repeated addition.
+                //
+                // A node is one past its predecessor only over a prime field.
+                //
+                // Over a binary tower an added step folds both evaluations onto one point.
+                let n0 = node * (n0_hi - n0_lo) + n0_lo;
```
Nashtare, line 138: **MEDIUM — Scaling by the runtime node costs about 15% of the prime-field prover, which is the only path in use.**

The node-2 evaluation now happens in the wider type. Each cell of the hot loop pays 4 multiplications by `node`, which the compiler cannot constant-fold: 2 extension×base and 2 extension×extension. It also pays 2 more extension×extension products, because `n0`/`n1` are now extension elements, so `d1 * n0` and `d0 * n1` are no longer extension×base. On main this walk was additions only.

Measured `prove_fractional_gkr` on BabyBear + quartic extension with packed denominators, `-Ctarget-cpu=native`, M4 Pro, main and PR runs alternated:
- single-threaded: nv=16 3.88 → 4.47 ms, nv=18 13.9 → 16.2 ms, nv=19 27.0 → 31.7 ms (+15-17%);
- rayon, nv=22, min of 11 runs: 61-65 → 67-72 ms.

`LookupPlan::build` still rejects characteristic 2. So today this PR only adds cost to the live path.

Fix: choose the walk once, outside the loop. For example, pass `node: Option<A>` set to `(EF::interpolation_node(2) != EF::TWO).then(|| ...)`. For `None`, keep the base-type `hi.double() - lo` walk. I measured that version at 13.6-14.0 ms (nv=18) and 26.8-27.1 ms (nv=19), the same as main. All fractional-GKR tests pass with it, including the binary one.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -293,10 +300,14 @@ impl<F: Field, EF: ExtensionField<F>> Layer<F, EF> {
     fn round_polys(&self, lambda: EF) -> [EF; 2] {
```
Nashtare, line 351: **LOW — Several docs still describe the round-polynomial domain as the integers 0, 2, 3.**

Stale docs:
- Line 299: "Evaluate the equality-stripped quadratic round polynomial at `0` and `2`".
- `FractionGkrLayerProof::round_polys` (`mod.rs:57`): "evaluations at points zero, two, and three".
- `LookupPlan::build` (`lookup/mod.rs:132-133`) still lists "fractional-GKR interpolation" as a reason characteristic 2 is rejected. After this PR, only the counting argument remains.
- `RoundPolyInterpolator` docs (`sumcheck/src/generic_degree/util.rs:11,17,58`) still say "integer nodes".

Fix: say "interpolation nodes 0, 2 and 3" in the first two, or refer to `Field::interpolation_node`. Drop the fractional-GKR clause from the lookup doc. The sumcheck wording can go in a follow-up.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -72,20 +73,24 @@ where
...
-
-                acc[1] += eq_suffix * evaluate(n0, n1, d0, d1);
+                // The gate at zero reads the stored values directly.
+                //
+                // Numerators stay in their own type, so the products stay mixed.
+                acc[0] += eq_suffix * (d1_lo * n0_lo + d0_lo * n1_lo + lambda * d0_lo * d1_lo);
+
+                // The gate at the domain's other node walks each line out to it.
+                //
+                // The step is a multiplication rather than a repeated addition.
```
Nashtare: **LOW — These comments contrast the code with the version it replaced instead of describing it.**

Two examples: "The step is a multiplication rather than a repeated addition" here, and "Each node scales the step directly rather than walking one step at a time" at line 536.

Fix: state the invariant in one line, e.g. "the line is evaluated at `interpolation_node(2)`, which is not `1 + 1` in characteristic 2". If the prime-field fast path is adopted, put that comment on the branch that selects it.

multi-stark/src/fractional_gkr/test.rs
```diff
@@ -340,3 +343,86 @@ fn split_fraction_matches_unsplit_sum_and_evaluation() {
...
+
+    let mut rng = SmallRng::seed_from_u64(0x0B14_A247);
+    for num_variables in 1..=8 {
+        let (numer, denom) = zero_sum_ext::<B>(&mut rng, num_variables);
+
+        let mut prover_challenger = challenger();
+        let (proof, prover_output) = prove_fractional_gkr(
+            &Fraction {
+                n: numer.clone(),
+                d: PolyMaybePacked::Scalar(denom.clone()),
```
Nashtare: **LOW — The binary test only covers `F == EF` with `Scalar` denominators, so over binary it never runs the mixed base/extension walk or the `Packed` dispatch.**

With `F == EF`, `mixed_round_polys` never runs with `N != A`. The `Packed` arms of `InputLayer::round_polys` and `Layer::round_polys` are never reached.

I ran a local variant: `F = BinaryField32`, `EF = BinaryField128`, with both `PolyMaybePacked::Scalar` and `PolyMaybePacked::Packed(denom.pack::<F, EF>())`, for `num_variables` in 1..=9. It passes on this head and fails on main at n=2 with `LayerConsistency { layer: 1 }`. That variant would also guard the fast-path split suggested on `prover.rs:88`.

Fix: loop the test over both representations and a mixed tower pair. Also, `zero_sum_ext` (line 350) duplicates `zero_sum_fraction`; making the existing helper generic over `(N, A)` covers both.

Nashtare (changes requested):
Reviewed at 44aa6062.

The fix is correct and complete for binary fields.
- Every prover node now matches the verifier's `interpolation_node(0..=3)`: node 2 in `mixed_round_polys` across all four callers, `q3`, and the equality factor at nodes 2 and 3.
- No integer-node assumption remains. The remaining `2c - 1` is a coefficient, not a node, and `h(1) = claim - h(0)` holds because nodes 0 and 1 are fixed as ZERO and ONE.
- On main this was a completeness bug, not a soundness hole. The verifier was already right, and an honest binary proof is rejected with `LayerConsistency { layer: 1 }`; I reproduced that.
- Each of the four node fixes is independently load-bearing: reverting any one of them fails `a_binary_field_reduction_round_trips`.

**Requesting changes for one reason: the fix makes the prime-field prover about 15% slower, and that is the only path in use today.** `LookupPlan::build` still rejects characteristic 2.

Walking to node 2 now multiplies by a runtime node in the extension type, which turns the mixed products into full extension products. Picking the walk once, outside the loop, recovers main's timings and still passes the binary test (details inline).

The other comments are LOWs:
- stale "integers 0, 2, 3" docs;
- two comments that describe the removed code;
- the binary test never runs the mixed base/extension walk or the packed dispatch.

**Validation**
- `cargo test -p p3-multi-stark --release` passes.
- clippy is clean with and without `parallel`.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Fixed in 423f14c8, and you were right about the cause: scaling lifted the numerators into the extension, so the gate's mixed products became extension products.

`mixed_round_polys` now takes `node: Option<A>` and picks the walk once, before the fold. `None` keeps main's `hi + (hi - lo)` step with narrow numerators; `Some` scales. The predicate is `TWO != interpolation_node(2)` — one step of a line evaluates it at `1 + 1`, which *is* the second node exactly when those are equal, so the two walks agree whenever the cheap one is taken and the proof is unchanged.

One thing worth reporting: the shape of the fast arm mattered as much as the branch. My first attempt factored the gate into an `#[inline(always)] fn` shared by both arms, and that alone left ~4% on the table, reproducibly. Mirroring main's exact inner body (the `n0_dif` bindings and the local `evaluate` closure) closed it.

Measured here, `-Ctarget-cpu=native`, best of seven, alternated with main, nv=19 packed:

- main: 11.75 / 12.03 / 11.75 ms
- this head: 11.36 / 11.45 / 11.46 ms
- scaling always (44aa6062): 12.71 / 12.74 / 13.60 ms
- shared-gate variant: 12.71 / 13.60 / 13.60 ms

So the regression is gone. My box is noisier than yours — main drifted between 11.26 and 12.47 across the session — so the honest claim is parity within noise rather than a win; worth a re-measure on the M4 Pro if you want the number pinned.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Fixed in 423f14c8, three of the four:

- `Layer::round_polys` now says "at the first two nodes";
- `FractionGkrLayerProof::round_polys` says "at interpolation nodes zero, two and three";
- `LookupPlan::build` and the `UnsupportedCharacteristic` variant now name only the counting argument, since that is the one reason left.

I left the `RoundPolyInterpolator` wording as you suggested. Worth noting for whoever picks it up: that doc was already inaccurate before this PR, since the code has always built its domain from `interpolation_node`. It is a pre-existing doc bug in a different crate rather than something this PR introduces.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Fixed in 423f14c8. Both now describe the code rather than what it replaced, and since the fast path landed they sit on the branch that selects each walk:

    // Stepping once lands on the node, so numerators stay in their own type and the
    // products stay mixed.
    None => ...
    // Scaling lifts the numerators into the extension, so the second value is taken
    // entirely there.
    Some(node) => ...

`restore_equality_factor` now reads "Each node scales the step, since a node is one past its predecessor only over a prime field."

You were right that this is a rule and not a preference — the repo's doc guidance says the same thing, and I broke it.

multi-stark/src/fractional_gkr/test.rs
tcoratger: Fixed in 423f14c8, taking both halves of the suggestion.

`zero_sum_fraction` and `zero_sum_ext` are now one helper generic over `(N, A: ExtensionField<N>)`, so the single-level and mixed cases share it.

The round trip is now `binary_round_trip<F, EF>`, looping `num_variables` in `1..=8` and over both `PolyMaybePacked::Scalar` and `Packed`, called twice:

- `BinaryField128 / BinaryField128` — the level-equal case;
- `BinaryField32 / BinaryField128` — the mixed walk, exactly your pair.

Both fail on main and pass here. The mixed one is the more interesting of the two: it is the only test in the file that runs `mixed_round_polys` with `N != A` over a binary field, which is what the fast-path split now branches on, so it guards that split as you predicted.

One caveat I should flag rather than leave implied: `BinaryField32::Packing` is `Self`, so the `Packed` arm there is a width-one wrapper. It reaches the `Packed` dispatch in `InputLayer::round_polys` and `Layer::round_polys`, which is what was uncovered, but it is not exercising real SIMD lanes.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -502,11 +577,20 @@ fn restore_equality_factor<EF: Field>(
...
     let q_sum = quadratic_evals[0] + q1;
-    let q3 = interpolator.eval(&quadratic_evals, q_sum, EF::from_u8(3));
+    let q3 = interpolator.eval(&quadratic_evals, q_sum, EF::interpolation_node(3));
+
+    // The equality factor in this variable is affine, so two values fix it everywhere:
+    //
+    //     eq(0)         = 1 - coordinate
+    //     eq(1) - eq(0) = 2 * coordinate - 1
+    //
+    // Each node scales the step directly rather than walking one step at a time.
```
Nashtare, line 587: The reply quotes a reworded line, but the pushed head still reads "Each node scales the step directly rather than walking one step at a time", which contrasts the code with a walk it doesn't do. Something like "Each node scales the step, since stepping reaches a node only in odd characteristic" describes it as it is.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -33,6 +33,25 @@ enum SplitFractionMaybePacked<F: Field, EF: ExtensionField<F>> {
     Scalar(SplitFraction<Poly<EF>>),
 }
 
+/// The second interpolation node, unless one step of a line already lands on it.
+///
+/// Stepping a line once evaluates it at `1 + 1`.
+///
+/// Over a prime field that is the second node, and the step stays in the narrow type.
```
Nashtare, line 40: Nit: the predicate compares values, so the cheap walk applies to every field whose second node is `1 + 1`, including extensions like BabyBear⁴, the live case. Maybe "odd characteristic, extensions included". Also `stepped_past` returns the node exactly when stepping does *not* reach it; a name like `scaled_node` would read more directly. Optional.

Nashtare (approved):
Re-reviewed at 423f14c8. The prime-field slowdown is gone.

**Speed:** BabyBear⁴ packed, single-threaded, interleaved, minimum of 6 rounds on an M4 Pro:

| nv | main | 44aa6062 | 423f14c8 |
|---|---|---|---|
| 16 | 4.200 ms | 4.766 ms | 4.238 ms |
| 18 | 15.06 ms | 17.73 ms | 15.29 ms |
| 19 | 29.59 ms | 34.97 ms | 29.75 ms |

**Correctness**
- **Proof bytes:** prime-field proofs are byte-identical to main at nv 12/16/18/19/22, and forcing the scaling walk on BabyBear gives the same bytes.
- **Binary guard:** forcing the stepping walk on a binary tower makes the verifier reject, so `a_mixed_binary_tower_reduction_round_trips` guards the split.
- **The check itself:** `interpolation_node(2) != TWO` is exactly the right test in any characteristic, and it is decided once per round.
- **Validation:**
  - multi-stark tests pass with and without `parallel`;
  - clippy is clean;
  - the PR merges cleanly onto current main.

**Remaining nits**
- **Doc wording not pushed:** the reply mentions two doc rewordings that the pushed head doesn't have. One is inline on line 587. The other is `prover.rs:350`, outside the diff, which still reads "Evaluate the equality-stripped quadratic round polynomial at `0` and `2`." Perhaps a commit that didn't get pushed?
- **Binary slowdown (informational):** binary is about 3-4% slower than at 44aa6062, since there are now two fold bodies in one function. That's fine while binary lookups stay gated.

multi-stark/src/fractional_gkr/prover.rs
Nashtare, line 587: The reply quotes a reworded line, but the pushed head still reads "Each node scales the step directly rather than walking one step at a time", which contrasts the code with a walk it doesn't do. Something like "Each node scales the step, since stepping reaches a node only in odd characteristic" describes it as it is.

multi-stark/src/fractional_gkr/prover.rs
Nashtare, line 40: Nit: the predicate compares values, so the cheap walk applies to every field whose second node is `1 + 1`, including extensions like BabyBear⁴, the live case. Maybe "odd characteristic, extensions included". Also `stepped_past` returns the node exactly when stepping does *not* reach it; a name like `scaled_node` would read more directly. Optional.

Nashtare (approved):
Re-reviewed at 423f14c8. The prime-field slowdown is gone.

**Speed:** BabyBear⁴ packed, single-threaded, interleaved, minimum of 6 rounds on an M4 Pro:

| nv | main | 44aa6062 | 423f14c8 |
|---|---|---|---|
| 16 | 4.200 ms | 4.766 ms | 4.238 ms |
| 18 | 15.06 ms | 17.73 ms | 15.29 ms |
| 19 | 29.59 ms | 34.97 ms | 29.75 ms |

**Correctness**
- **Proof bytes:** prime-field proofs are byte-identical to main at nv 12/16/18/19/22, and forcing the scaling walk on BabyBear gives the same bytes.
- **Binary guard:** forcing the stepping walk on a binary tower makes the verifier reject, so `a_mixed_binary_tower_reduction_round_trips` guards the split.
- **The check itself:** `interpolation_node(2) != TWO` is exactly the right test in any characteristic, and it is decided once per round.
- **Validation:**
  - multi-stark tests pass with and without `parallel`;
  - clippy is clean;
  - the PR merges cleanly onto current main.

**Remaining nits**
- **Doc wording not pushed:** the reply mentions two doc rewordings that the pushed head doesn't have. One is inline on line 587. The other is `prover.rs:350`, outside the diff, which still reads "Evaluate the equality-stripped quadratic round polynomial at `0` and `2`." Perhaps a commit that didn't get pushed?
- **Binary slowdown (informational):** binary is about 3-4% slower than at 44aa6062, since there are now two fold bodies in one function. That's fine while binary lookups stay gated.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: You are right, and I owe you a correction: that reply quoted a rewording I had written but never applied. The line on the pushed head was the old one. Sorry for sending you back to check it.

Fixed for real in c7958334. It now reads:

```rust
    // Each node scales the step.
    //
    // Stepping would reach the next node only in odd characteristic.
```

I grepped the rest of the diff for the same shape rather than fixing only the line you pointed at, and that one was the last of them.

tcoratger: You are right, and I owe you a correction: that reply quoted a rewording I had written but never applied. The line on the pushed head was the old one. Sorry for sending you back to check it.

Fixed for real in c7958334. It now reads:

```rust
    // Each node scales the step.
    //
    // Stepping would reach the next node only in odd characteristic.
```

I grepped the rest of the diff for the same shape rather than fixing only the line you pointed at, and that one was the last of them.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Both taken, in c7958334.

**The predicate.** You are right that this is about the enumeration, not primality: `interpolation_node` defaults to `from_usize`, so BabyBear⁴ lands on `TWO` and takes the cheap walk — which is exactly the path the benchmark in your first comment measured. Calling it "prime field" described a narrower set than the code selects. Three places said it; all three now say odd characteristic, extensions included:

- the helper's doc;
- the interpolation-domain doc on `restore_equality_factor` ("the two coincide in odd characteristic, extensions included");
- the comment on the equality step.

**The name.** Renamed to `scaled_node`. You are right that the old one read backwards: it returned the node in the case where stepping does *not* reach it, so the name described the branch it is absent on. The new one says what the value is for — the node a line has to be scaled onto — and `None` then reads as "nothing to scale by".

The helper is private to the module, so nothing outside moved. Docs and one name only, no codegen change: `cargo test -p p3-multi-stark` and clippy are clean.

tcoratger: Both taken, in c7958334.

**The predicate.** You are right that this is about the enumeration, not primality: `interpolation_node` defaults to `from_usize`, so BabyBear⁴ lands on `TWO` and takes the cheap walk — which is exactly the path the benchmark in your first comment measured. Calling it "prime field" described a narrower set than the code selects. Three places said it; all three now say odd characteristic, extensions included:

- the helper's doc;
- the interpolation-domain doc on `restore_equality_factor` ("the two coincide in odd characteristic, extensions included");
- the comment on the equality step.

**The name.** Renamed to `scaled_node`. You are right that the old one read backwards: it returned the node in the case where stepping does *not* reach it, so the name described the branch it is absent on. The new one says what the value is for — the node a line has to be scaled onto — and `None` then reads as "nothing to scale by".

The helper is private to the module, so nothing outside moved. Docs and one name only, no codegen change: `cargo test -p p3-multi-stark` and clippy are clean.

PR #2133, opened 2026-09-14
feat(multi-stark)!: add the logUp* indexed-lookup reduction

tcoratger:
## Why

A logarithmic-derivative lookup puts an integer count over each table entry, and
counts are read back modulo the characteristic. Modulo two they say only "odd or
even", so a value looked up an even number of times is invisible:

```text
    LHS = 1/(c - v) + 1/(c - v) = 2/(c - v) = 0     in GF(2^128)
    RHS = 0/(c - t_0) + 0/(c - t_1)          = 0
```

The two sides agree for *every* challenge, as rational functions. There is no
Schwartz-Zippel escape. That is why `LookupPlan::build` rejects characteristic
two outright today, and why `check_multiplicity_height_bound` cannot help: its
bound is `< p`, which is `< 2`.

## What

logUp\* ([eprint 2025/946](https://eprint.iacr.org/2025/946)) replaces the count
with a field element. A reader names the entry it reads, and pullback and
pushforward are adjoint, so a claim about the pulled values turns inside out:

```text
    X(r) = <X, eq_r> = <T, I_* eq_r> = <T, Y>
```

The left inner product runs over the reader's rows, the right over the table's
entries. `Y` is the only new object and it is as wide as the table. The weight
over an entry is `sum_{i : I[i] = v} eq_r(i)` -- a field element that no reduction
ever touches. Two checks close it:

- that `Y` really is the pushforward, as one sum of fractions proved by the
  existing fractional GKR;
- that `<T, Y> = e`, as one degree-two sumcheck.

The reduction takes a statement and hands back evaluation claims on the table
columns and the position columns, for the caller to discharge against its
commitments. It batches several readers per table by powers of one challenge,
several tables under one reduction, and several columns per table through a
single pushforward -- so a wide entry needs no fingerprint and carries no
tuple-width aliasing hazard.

Each table draws its **own** challenge. That is not an optimization: under a
shared one, two tables can miscount opposite entries and cancel exactly. There is
a test for that.

The pushforward travels in the clear. It is only as wide as the table, which is
the regime the paper targets, and it removes a mid-protocol commitment round
entirely. The trade-off is proof size (4 KB for a 256-entry table) and that it is
not zero-knowledge; a committed variant is the follow-up for large tables.

## Field-genericity

The only field-specific hook is how a table position embeds, which is
`Field::interpolation_node`: the integer itself over a prime field, the matching
bit pattern over a binary tower. Either way the embedding is affine in the
position's bits, so the verifier evaluates the table-side denominator itself in
time linear in the variable count. Everything round-trips over both `GF(2^128)`
and `BabyBear`.

## Preconditions the caller owes

Two, both documented on the statement types:

1. **The claim point is a verifier challenge drawn after the position column and
   the table are committed.** The free range check below rests on it.
2. **The claims and the point are bound in the caller's own transcript.** The
   reduction absorbs them itself as its first step, so this is belt and braces
   rather than load-bearing.

## A position out of range needs no range check

It leaves a pole the table side cannot match, whose residue is the multilinear
extension of a nonempty row set evaluated at `r`. The position column is fixed
before `r` is drawn, so that residue is nonzero except with probability `n/|F|`
and the identity fails on its own.

Two tests hold that up. One shows no pushforward closes the pole -- neither
dropping the offending row's weight nor moving it onto another entry. The other
builds the only proof such a read admits, which has to lie about a weight the
verifier rebuilds from the claim point, and drives the real verifier against it.

## Fractional GKR

The bottom table's numerator becomes a two-variant enum, which is the breaking
change the title marks: `prove_fractional_gkr` takes a `LeafFraction` by value
rather than a `&Fraction<Poly<F>, PolyMaybePacked<F, EF>>`. The only in-tree
caller is updated.

LogUp's multiplicities stay in the base field on the path that has always used
them; logUp\*'s equality weights and pushforwards need the wider storage. A
committed golden pins the base-field path against a fixed reference across
`n = 1..8` and both denominator storages, a test pins the two numerator storages
equal, and another pins packed against scalar.

## Testing

52 tests in the new module, 26 of them unit tests beside the code they cover:

| file | tests |
|---|---|
| `plan` | ordering, alignment, tiling without gaps or overlap, determinism, reader numbering, 4 rejections |
| `position` | injectivity, first entry to zero, additivity over bits, closed form vs explicit table, agreement on the cube |
| `witness` | pushforward vs a hand-written scatter, unread entries, block placement, padding is `0/1`, honest sum is zero |
| `product` | round polynomial vs a naive oracle, starting sum, padded table owes its whole product |
| `transcript` | one shared sponge state, six ways a shape moves the seed, the statement and the pushforwards bound before what weighs them, no zero challenge |

The rest are round trips over both fields, a serde round trip, five tampering
cases, and three that pin the cryptography directly:

- **`reading_one_entry_an_even_number_of_times_does_not_cancel`** -- the direct
  regression for the failure above, asserting the weight over a twice-read entry
  is `1 - r_1`, a field element rather than a count.
- **`one_challenge_per_table_is_what_stops_two_tables_cancelling`** -- moves a
  unit of weight between two tables' entry 1 and shows the sum stays nonzero
  under distinct challenges and vanishes *exactly* under a shared one.
- **`a_position_outside_the_table_is_rejected_while_scattering`** and its two
  companions -- the free range check, from the prover's own guard down to the
  verifier's verdict.

Two goldens sit beside them, one per path, digesting a fixed run's proof bytes.
They are what catches a change that keeps prover and verifier agreeing and so
leaves every round trip green: sharing one entry challenge between tables, or
absorbing zeros in place of a pushforward, both turn them red.

`p3-multi-stark-logup-star` is registered in the cross-protocol separation
suite, with a plain shape and two single-field moves of it.

I checked the product-sumcheck test is load-bearing: breaking the padding factor
makes it fail and localizes the failure to `product.rs`, where previously only a
full round trip would have flagged it.

`cargo test -p p3-multi-stark` passes (184 lib tests plus integration), with and
without `--features parallel` -- worth checking explicitly, since the scatter and
the round-polynomial accumulation both go through `par_fold_reduce`, whose
contract diverges between serial and parallel builds if the reduction is not
associative. Clippy, rustfmt, rustdoc and `scripts/check.py lint` are clean.

## Not in this PR

Nothing calls this yet. Deliberately out of scope, in the order I would do them:

1. the AIR-facing declaration API in `p3-lookup`;
2. the coupling into the multi-STARK prover, which needs the reduction to run
   *after* the zerocheck (it consumes `eq_r`) and a second opening point on the
   main commitment;
3. the security-accounting terms, which have nothing to account for until (2);
4. a benchmark -- there is none, so this PR makes no performance claim.

One honest caveat on cost: in an AIR the looked-up column is committed anyway, so
logUp\* *adds* a position column rather than saving one. The win here is
soundness in characteristic two, not size. It only becomes a size win once
`BinaryPcs` speaks subfields -- `p3_sumcheck::ring_switch` has the reduction, it
is just not wired in -- or when the looked-up vector is virtual, as in a
Spartan-style frontend.


Nashtare (commented):
Reviewed at 09ee18d2. These comments cover only the logUp\* commit. #2132's commit is reviewed on its own PR, and I'll hold approval here until #2132 lands.

**What checks out**
- The protocol matches logUp\* (2025/946), and the transcript order is right: λ, then `Y` absorbed, then each table's own nonzero `c`, the GKR, the position claims, μ, the product sumcheck, and the column claims.
- The verifier recomputes the reader numerators, the padding leaves and the table-side denominators itself.
- 19 malformed-proof probes were all rejected with `Err` and no panic. They covered wrong widths, counts, and round and layer counts, an extra PoW witness, a zero root denominator, and an oversized postcard length prefix.
- The existing lookup path is byte-identical to the parent commit. I compared direct `prove_fractional_gkr` dumps and full `MultiStarkProof` bytes, in serial and parallel builds.

**Main point (MEDIUM, `verifier.rs`):** the reader claims and points are never absorbed. A statement forged along λ verifies against an honest proof; the probe is in the comment. Either bind them before λ, or state it as a caller precondition and test it.

**Other findings**
- Two more MEDIUMs are test gaps. Sharing one entry challenge across tables leaves all 41 logUp\* tests green, and so does leaving `Y` unbound.
- The LOWs cover:
  - the preconditions behind the out-of-range argument;
  - the seed not binding how readers split across tables;
  - the assumptions on the position embedding;
  - a breaking API change without `!`;
  - the scatter's allocation pattern;
  - the public surface.

**Validation**
- `cargo test -p p3-multi-stark --release` passes with and without `parallel`.
- clippy and rustdoc are clean.
- No conflicts with main.

multi-stark/src/logup_star/verifier.rs
```diff
@@ -0,0 +1,275 @@
...
+        // Phase 2: replay, from the reader weights down to the column claims.
+        //
+        // Every step is played even once a check has failed.
+        //
+        // The sponge the caller gets back is then in the state the prover left it in.
+        let shape = LogupStarShape::new(&plan);
+        let mut transcript =
+            LogupStarVerifierTranscript::<Challenger, F, EF>::new(challenger, &shape);
+
+        let reader_batching = transcript.reader_batching();
```
Nashtare, line 80: **MEDIUM: the reduction never absorbs the reader claims or points, so a false statement verifies unless the caller has already bound them.**

The seed (`transcript.rs:221-245`) binds only shape numbers. `lambda` is therefore a function of the challenger state and the shape alone, and a prover can compute it before choosing claims. `mu` depends only on proof values, so it is predictable in the same way. `product::claimed_sum` is linear in the claims, and the GKR numerator uses only `eq_r`. Moving `X_0 += lambda*delta` and `X_1 -= delta` leaves every check unchanged.

Probe: an honest proof for `[(3, 1, [2, 2])]`, verified against that forged statement, returns `Ok(())`. By the same argument (not probed), one reader with two columns is forgeable along `mu`.

This contradicts the doc above ("those claims and the statement's own stand or fall together"). It also departs from the sibling protocols: `GenericDegreeProof` absorbs `claimed_sum`, and fractional GKR absorbs `root_denominator`.

Fix, either:
- absorb every `r_{t,j}` and `X_{t,j,k}` as a message step before `READER_BATCHING`, in both transcripts (cheap); or
- state this as a `# Soundness` precondition on `prove`/`verify`/`TableLookup` and pin the forgery in a test.

multi-stark/src/logup_star/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub use error::LogupStarError;
+use p3_multilinear_util::point::Point;
+pub use plan::LogupStarPlan;
+pub use proof::{LogupStarOutput, LogupStarProof, TableOutput};
+
+/// One reader's claim on the values it pulled out of a table.
+#[derive(Clone, Copy, Debug)]
+pub struct Reader<'a, EF> {
+    /// Point at which every pulled column is claimed.
```
Nashtare, line 87: **LOW: the "no range check needed" argument needs this point to be drawn after the position column is committed, and the code never says so.**

An out-of-range position `a` leaves residue `sum_j lambda^j MLE(1_{I_j = a})(r_j)`. That residue is nonzero with high probability only if `I_j` is fixed before `r_j`, and `lambda` after both. The reduction takes `r_j` as input and cannot enforce this.

The precondition appears only in the PR description. A caller passing a public point, or one reused across commitments, silently loses the range check.

Fix: document on `Reader::point` / `TableLookup` that `r` must be a verifier challenge drawn after `I` and `T` are committed, and that the claims must be bound before the call (see the comment on `verifier.rs:79`).

multi-stark/src/logup_star/transcript.rs
```diff
@@ -0,0 +1,630 @@
...
+
+        // How wide the padded leaf table is, and how many tables and readers share it.
+        separator
+            .instance(&(self.num_variables as u64).to_be_bytes())
+            .instance(&(self.table_variables.len() as u64).to_be_bytes())
+            .instance(&(self.reader_variables.len() as u64).to_be_bytes());
+
+        // Two statements can agree on every total above and still differ below.
+        //
+        // They may split readers between tables differently, or pull different columns.
```
Nashtare, line 278: **LOW: the seed does not bind how readers are split between tables, although this comment says it does.**

The seed binds `#tables`, `#readers`, each table's width and the *flattened* reader heights. The pattern binds the pushforward lengths and the totals. None of these separates `[2 readers | 1 reader]` from `[1 reader | 2 readers]` when tables share size and width and readers share height.

Probe: the two `LogupStarPlan`s are `!=`, yet their `LogupStarShape`s are `==` and they draw the same `reader_batching` challenge. I found no attack, since the verifier fixes the statement. It does falsify this invariant, and `a_shape_is_bound_before_anything_is_drawn` does not cover the case.

Fix: bind `readers.len()` for each table (or keep reader heights grouped per table), and add this pair to the test.

multi-stark/src/logup_star/position.rs
```diff
@@ -0,0 +1,144 @@
...
+use p3_field::{ExtensionField, Field};
+use p3_multilinear_util::point::Point;
+
+/// Embed one table position.
+///
+/// Over a prime field this is the integer itself.
+///
+/// Over a binary tower it is the element whose bit pattern is that integer.
+///
+/// Both are injective as far as the field reaches.
```
Nashtare: **LOW: the embedding's injectivity and bit-additivity are assumed but never checked, and both are load-bearing.**

**Injectivity.** Two entries sharing a pole let `Y` move weight between them. Over a prime field `iota(v) = v mod p`, so `m >= log2(p)` aliases entries. For BabyBear that is `m >= 31`; impractical because `Y` is sent in the clear, but not rejected. Over `BinaryField8` (a `TranscriptField`), `m > 8` panics inside `interpolation_node`, and it does so in the verifier at `position::eval`, not in a check made when the plan is built.

**Additivity.** `eval` needs `iota` to be additive over bits. The `Field::interpolation_node` contract promises only an injective enumeration with `0 -> 0` and `1 -> 1`. Every current implementation happens to be additive, but a future override would make the prover's explicit table and the verifier's closed form disagree.

Fix:
- `LogupStarPlan::new` rejects `2^m` at or above the base field's characteristic (or `|F|`).
- The additivity requirement goes into the `interpolation_node` contract, or the embedding is defined here as `sum_bits iota(2^k)`.

multi-stark/src/logup_star/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+    assert_eq!(pushforward[0], point[1]);
+    assert_ne!(pushforward[1], Binary::ZERO);
+
+    // Entries nobody read stay empty.
+    assert_eq!(pushforward[2], Binary::ZERO);
+    assert_eq!(pushforward[3], Binary::ZERO);
+}
+
+#[test]
+fn one_challenge_per_table_is_what_stops_two_tables_cancelling() {
```
Nashtare, line 361: **MEDIUM — `one_challenge_per_table_is_what_stops_two_tables_cancelling` does not guard the protocol: making prover and verifier share one entry challenge leaves all 41 logUp* tests green.**

The test calls `leaf_tables` directly with hand-picked challenge arrays and sums the fractions. It never reaches `LogupStarProverTranscript::entry_challenges` or the `entry_challenges[block.table]` lookups in `prover.rs`/`verifier.rs`.

Mutation run at 09ee18d2: in both `prover.rs` and `verifier.rs`, `let entry_challenges = { let c = transcript.entry_challenges(plan.num_tables()); vec![c[0]; c.len()] };` → `cargo test -p p3-multi-stark --release --lib logup_star` gives 41 passed, 0 failed.

The PR description calls this test one of three that "pin the cryptography directly". It only pins the algebra. Suggested fix: add a test that goes through the real code. For example, a two-table instance where the verifier is fed the cancelling pushforwards through the real `verify` path. At minimum, add a transcript/prover test asserting the drawn challenges differ per table and that `rebuild_opening` uses each block's own table challenge.

multi-stark/src/logup_star/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+    let mut prover_challenger = binary_challenger();
+    let (mut proof, _) = LogupStarProof::prove(&lookups, &witness, &mut prover_challenger);
+    mutate(&mut proof);
+
+    let mut verifier_challenger = binary_challenger();
+    proof.verify(&lookups, &mut verifier_challenger).map(|_| ())
+}
+
+#[test]
+fn rejects_a_tampered_pushforward() {
```
Nashtare, line 775: **MEDIUM — No test detects the pushforward going unbound in the transcript: absorbing zeros in place of `Y` on both sides still passes every test.**

Mutation run at 09ee18d2: in `transcript.rs`, both `pushforward` methods absorb `&vec![EF::ZERO; values.len()]` instead of `values`. The pattern and length checks are unchanged. Result: 41 passed, 0 failed.

`rejects_a_tampered_pushforward` still passes because the verifier recomputes the table block's numerator from the tampered `Y` in `rebuild_opening`, so the proof is rejected with `LeafNumerator` whether or not `Y` was absorbed. The test only asserts `is_err()`. Binding `Y` before `c` is the ordering the transcript docs call load-bearing. A prover that picks `Y` after seeing `c` has one linear constraint to meet per table.

Suggested fix: a transcript unit test that plays two runs differing only in one pushforward entry and asserts `entry_challenges` differ. Also pin the error variant here.

multi-stark/src/logup_star/tests.rs
```diff
@@ -0,0 +1,633 @@
...
+    assert_eq!(
+        fraction_sum(numerator.as_slice(), denominator.as_slice()),
+        Binary::ZERO,
+        "a shared challenge is exactly what would let the two errors cancel"
+    );
+}
+
+#[test]
+#[should_panic(expected = "a row names an entry outside its table")]
+fn a_position_outside_the_table_is_rejected_while_scattering() {
```
Nashtare, line 506: **LOW — The out-of-range position property is tested only as a prover-side panic. Nothing checks that the full verifier rejects it.**

`weights` panics on `split.get_mut(entry)` before any proof exists. The claim in the PR ("a position out of range needs no range check", because the pole cannot be matched) is a verifier property, and no test drives `LogupStarProof::verify` with a proof built from an out-of-range or padding position.

Suggested fix: add a test that builds the leaf tables for a reader naming entry `2^m` (drop that row's weight from the pushforward) and runs `prove_fractional_gkr` and the product sumcheck by hand. Then assert that `verify` returns `Err`.

multi-stark/src/fractional_gkr/test.rs
```diff
@@ -340,3 +336,183 @@ fn split_fraction_matches_unsplit_sum_and_evaluation() {
...
+        );
+        assert_eq!(
+            prover_output.denominator,
+            denom.eval_ext::<B>(&prover_output.point)
+        );
+    }
+}
+
+#[test]
+fn an_extension_numerator_matches_a_base_numerator() {
```
Nashtare, line 509: **LOW — This test compares two new code paths, not the old one, so it does not show the existing logUp path is byte-identical as the PR description claims.**

Both proofs come from the new `prove_fractional_gkr`: one through `LeafNumerator::Base`, one through `Ext`. A regression shared by both arms would pass.

I checked the real claim separately, and it holds. I ran a probe against 44aa6062 (API adapted) and 09ee18d2 over 23 BabyBear base-leaf instances (scalar and packed denominators, n = 1..12) and 8 `BinaryField128` instances, dumping postcard proof bytes, output openings and the next challenger sample. I also dumped postcard bytes of the full `MultiStarkProof` from `prove_verify_lookup_roundtrips_through_pcs` and `prove_verify_mixed_height_lookups_roundtrip_through_pcs`. Everything was identical, with and without `parallel`.

Suggested fix: reword the PR description, or add a small committed golden (a hash of the proof bytes) so later changes are checked against a fixed reference.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -528,7 +699,7 @@ fn restore_equality_factor<EF: Field>(
 /// Panics if the numerator and denominator have different variable counts, if
 /// they have no variables, or if the fully reduced root denominator is zero.
 pub fn prove_fractional_gkr<F, EF, Challenger>(
-    fraction: &Fraction<Poly<F>, PolyMaybePacked<F, EF>>,
+    fraction: LeafFraction<'_, F, EF>,
```
Nashtare, line 755: **LOW — Breaking public API change: `prove_fractional_gkr` now takes `LeafFraction<'_, F, EF>` by value instead of `&Fraction<Poly<F>, PolyMaybePacked<F, EF>>`.**

This is a `pub fn` in a `pub mod`, and every downstream caller must rewrap its input as `Fraction { n: LeafNumerator::Base(&n), d: &d }`. The only in-tree caller (`lookup/mod.rs`) is updated. The PR title has no `!`, and `LeafNumerator`/`LeafFraction` become new public types.

Suggested fix: mark the title `feat(multi-stark)!:`. Or keep a thin `&Fraction<Poly<F>, PolyMaybePacked<F, EF>>` entry point for the base-field case.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -128,30 +136,55 @@ where
...
+            }
+            (
+                LeafNumerator::Ext(PolyMaybePacked::Scalar(numer)),
+                PolyMaybePacked::Scalar(denom),
+            ) => {
+                let (n0, n1) = numer.as_slice().split_at(numer.num_evals() / 2);
+                let (d0, d1) = denom.as_slice().split_at(denom.num_evals() / 2);
+                SplitFractionMaybePacked::Scalar(reduce_fraction(n0, n1, d0, d1))
+            }
+            _ => unreachable!("{MIXED_LEAF_STORAGE}"),
```
Nashtare, line 228: **LOW — A mixed numerator/denominator storage is reachable from the public API, and it fails as `unreachable!` ("internal error: entered unreachable code") instead of a documented assertion.**

Because `LeafNumerator::Ext` wraps its own `PolyMaybePacked`, a caller can pass `Ext(Scalar)` with a `Packed` denominator, or the reverse. What happens depends on the size:

- For `n >= 3` this panics here in `reduce`.
- For `n = 1`, `leaf_pair` accepts it silently.
- For a packed denominator with `n < 3`, the scalar-unpack shim runs first, so a scalar numerator is accepted.

Suggested fix: add one `assert!` at the top of `prove_fractional_gkr`, requiring both halves to share a storage mode whenever the numerator is `Ext`. List it under `# Panics`. The `_ => unreachable!` arms then really are unreachable.

multi-stark/src/logup_star/prover.rs
```diff
@@ -0,0 +1,238 @@
...
+        // Sharing one challenge between tables would let two of them cancel each other's errors.
+        let entry_challenges = transcript.entry_challenges(plan.num_tables());
+
+        // Phase 3: prove the fractions sum to zero.
+        //
+        // The tables stay scalar rather than packed.
+        //
+        // That lets their blocks be read back below without a second pass over the positions.
+        let (numerator, denominator) = leaf_tables(&plan, witness, &weights, &entry_challenges);
+        let numerator = PolyMaybePacked::Scalar(numerator);
```
Nashtare: **LOW — logUp\* always hands the fractional GKR scalar tables, so the two new `LeafNumerator::Ext(PolyMaybePacked::Packed(_))` arms never run in production.**

That is about 70 lines across `reduce`/`round_polys`/`fold` in `fractional_gkr/prover.rs`, plus the packed-numerator unpack shim. Only `a_packed_extension_numerator_preserves_the_transcript` exercises them. It also means logUp\* over BabyBear gives up the packed GKR kernel. On the base-leaf path at n = 20 (M-series, `-Ctarget-cpu=native`, `parallel`, heavy machine load, so rough) packed took about 32 ms and scalar about 41 ms.

Suggested fix: either pack the leaf tables here (the position claims can be read from the unpacked copy or recomputed from `positions`), or drop the packed `Ext` arms until a caller needs them.

multi-stark/src/logup_star/witness.rs
```diff
@@ -0,0 +1,451 @@
...
+                    // Rows split across threads, each split filling its own copy of the table.
+                    //
+                    // Scattering in place would let two rows naming one entry race.
+                    //
+                    // The splits merge afterwards, since addition ignores order.
+                    let scattered = reader_witness
+                        .positions
+                        .par_iter()
+                        .zip(weights.par_iter())
+                        .par_fold_reduce(
```
Nashtare: **LOW — The scatter allocates, zeroes and merges a full table-sized vector for every rayon split of every reader. The cost scales with readers × splits × 2^m, not with rows.**

`par_fold_reduce(|| EF::zero_vec(num_entries), ...)` runs inside a sequential fold over readers. Rayon seeds one accumulator per split, at least `min(rows, current_num_threads)` of them, and each `reduce` is another O(2^m) `add_slices`. With many short readers against a large table (say 2^20 entries), zeroing and merging dominate the actual scatter. Peak memory is about threads × table.

Suggested fix: fold all readers of a table in a single parallel pass (chain every reader's `(position, weight)` pairs), so allocation happens per split rather than per reader × split. Or scatter serially below a row threshold. Related: `ProductProver::new` clones each pushforward (`product.rs:113`) where it could take ownership.

multi-stark/src/logup_star/transcript.rs
```diff
@@ -0,0 +1,630 @@
...
+use p3_challenger::{CanObserve, CanSample};
+use p3_field::ExtensionField;
+
+use super::plan::LogupStarPlan;
+
+/// Version byte bound into the transcript seed.
+const VERSION: u8 = 1;
+
+/// Protocol name bound into the transcript seed.
+const NAME: &[u8] = b"p3-multi-stark-logup-star";
```
Nashtare, line 65: **LOW — The new protocol name `p3-multi-stark-logup-star` and `LogupStarShape` are not registered in the cross-protocol separation suite.**

`examples/tests/transcript_separation.rs` registers `FractionGkrShape`, `LookupShape`, `ZerocheckShape`, `GenericDegreeShape` and others, but not this shape. So nothing checks that its seed differs from every other protocol's.

Suggested fix: add a `LogupStarShape` entry, with at least one plain shape and one perturbed shape, next to the `LookupShape` block.

multi-stark/src/logup_star/prover.rs
```diff
@@ -0,0 +1,238 @@
...
+/// # Panics
+///
+/// Panics if the statement and the witness disagree on how many tables or readers there are.
+///
+/// Panics if a row names an entry its table does not have.
+///
+/// Panics if the claims the statement carries are not the ones the witness produces.
+///
+/// That means the caller was asked to prove something false.
+impl<F, EF> LogupStarProof<F, EF>
```
Nashtare, line 22: **LOW — The `/// Prove one indexed-lookup reduction ... # Panics` doc block is attached to the `impl` block, not to `pub fn prove`, so rustdoc shows `LogupStarProof::prove` with no docs.**

Suggested fix: move lines 20-40 onto `pub fn prove` (line 47).

multi-stark/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub mod folder;
 pub mod fractional_gkr;
 pub mod instance;
 pub mod keys;
+pub mod logup_star;
```
Nashtare, line 24: **LOW — `pub mod logup_star` publishes `prove`/`verify`, `LogupStarPlan`, and the `position` and `transcript` submodules (including `LogupStarProverTranscript`/`LogupStarVerifierTranscript`), even though nothing in the workspace calls them.**

The reduction is not wired into `MultiStark` or any example. Once released, these become semver surface before the committed-pushforward follow-up changes their shape.

Suggested fix: consider `pub(crate)` for `position` and `transcript` (tests can reach them in-crate), or mark the module `#[doc(hidden)]`/experimental until a caller exists.

multi-stark/src/logup_star/verifier.rs
tcoratger: Fixed by absorbing the statement, which is the cheap option you list first.

The transcript now opens with a `Kind::Public` step carrying every reader's claim point and every claimed value, flattened to the base field, before `READER_BATCHING` is drawn. Both sides play it, and both build that vector from the statement alone, so nothing new travels.

That kills the forgery you probed. `lambda` is now a function of the claims, so `X_0 += lambda*delta` / `X_1 -= delta` changes the seed the whole reduction runs from and the honest proof no longer replays. Same for the `mu` direction on a two-column reader.

The number of statement values enters the pattern as a `Length::Fixed`, so a statement of a different shape is a different transcript rather than a short read.

Two tests cover it:

- `a_claim_reaches_the_sponge_before_anything_is_drawn` plays two runs differing in exactly one claim and asserts the next draw differs.
- `rejects_claims_the_witness_never_produced` drives the full verifier.

The doc you quote is now accurate rather than aspirational. The caller-side precondition is documented as well (see the reply on `mod.rs`), but it is belt and braces now rather than load-bearing.

multi-stark/src/logup_star/mod.rs
tcoratger: Documented. `Reader` now carries a `# Soundness` block saying that the claim point must be a challenge the surrounding protocol drew after committing both the position column and the table, and that the reduction takes the point as given and cannot check this. `TableLookup` and both `prove` and `verify` point at the same requirement.

The PR description has a "Preconditions the caller owes" section now, instead of leaving it in prose.

The second half of your comment is handled separately: the reduction binds the claims itself as its first step, so a caller that forgets is no longer forgeable.

The property itself also gained two tests — see the reply on `tests.rs:440`.

multi-stark/src/logup_star/transcript.rs
tcoratger: You are right, and the comment was describing an invariant the seed did not have.

`LogupStarShape` no longer flattens. It holds one `LogupStarTableShape` per table, each carrying that table's entry count, width and reader heights, and the separator binds each table's reader count immediately before that table's own heights:

```text
    num_variables | #tables | #readers | (width, #readers, heights...) per table
```

So `[2 readers | 1 reader]` and `[1 reader | 2 readers]` now seed differently.

`how_readers_split_between_tables_reaches_the_seed` is the pair you describe: same totals, same widths, same heights, different split. It asserts the shapes differ and that the two runs leave the sponge in different states.

The per-table struct also mirrors `LookupShape` / `LookupInstanceShape`, which is what made the shape constructible in the cross-protocol separation suite (see the reply on `transcript.rs:59`).

multi-stark/src/logup_star/position.rs
tcoratger: Both halves fixed, and both by the route you suggest.

**Additivity.** The embedding is now *defined* as the bit sum rather than read off the enumeration:

```text
    iota(v) = sum_{k : bit k of v is set} interpolation_node(2^k)
```

The prover's explicit table and the verifier's closed form are then the same function by construction, whatever an implementation does with `interpolation_node` on a non-power-of-two. `the_bit_sum_agrees_with_the_field_enumeration` pins that the two still coincide on today's fields, so a future divergence surfaces as a failing test rather than as a soundness bug.

**Injectivity.** `LogupStarPlan::new` now rejects a table the base field cannot index injectively: `num_variables < F::bits()`. It runs when the plan is built, on both sides, before anything is drawn — so your `BinaryField8` case is a clean panic in the plan rather than one inside `interpolation_node` halfway through a verifier.

The bound is conservative by one bit over a tower, where the field would just hold a full table. A table that large leaves the embedding no headroom and buys nothing, so I did not chase the last bit.

Covered by `rejects_a_table_the_field_cannot_index` and `a_table_only_fits_while_its_entries_stay_distinct`.

multi-stark/src/logup_star/tests.rs
tcoratger: You are right: that test pins the algebra, not the code. It stays, because it localizes the failure to the fraction identity, but it is no longer what guards the protocol.

The guard is now a golden. `a_fixed_statement_always_produces_the_same_proof` runs the real `LogupStarProof::prove` over two tables at a fixed seed and asserts a Keccak digest of the postcard-encoded proof. Your mutation changes the leaf tables, so the proof bytes change, so the digest changes.

I re-ran it exactly as you wrote it, in both `prover.rs` and `verifier.rs`:

```text
test logup_star::tests::a_fixed_statement_always_produces_the_same_proof ... FAILED
test result: FAILED. 51 passed; 1 failed
```

`each_table_draws_its_own_entry_challenge` covers the draw itself, asserting two tables get distinct challenges off one sponge.

The PR description no longer calls the algebraic test one that "pins the cryptography"; it now says what it does pin and names the goldens separately.

multi-stark/src/logup_star/tests.rs
tcoratger: Fixed with the test you suggest, plus two more that fell out of it.

`a_pushforward_reaches_the_sponge_before_the_entry_challenges` plays two runs differing only in one pushforward entry and asserts the drawn `entry_challenges` differ.

`rejects_a_tampered_pushforward` now pins *which party* rejects, which turns out to be the ordering test you wanted:

```text
    absorbed before the entry challenges  ->  every later draw moves
                                          ->  the reduction's own replay diverges

    not absorbed                          ->  the draws are unchanged
                                          ->  only the rebuilt opening disagrees
```

So it asserts `Err(LogupStarError::FractionGkr(_))`, not `is_err()`.

Your mutation now fails three tests:

```text
test logup_star::transcript::tests::a_pushforward_reaches_the_sponge_before_the_entry_challenges ... FAILED
test logup_star::tests::a_fixed_statement_always_produces_the_same_proof ... FAILED
test logup_star::tests::rejects_a_tampered_pushforward ... FAILED
test result: FAILED. 49 passed; 3 failed
```

multi-stark/src/logup_star/tests.rs
tcoratger: Added, though not quite in the shape you sketched — the literal version cannot be built, and that turned out to be the interesting part.

Drop the offending row's weight from the pushforward and the leaf fractions no longer sum to zero, so `prove_fractional_gkr` refuses to run. That is not an accident of our prover: the residue sits over the pole `c - iota(2^m)`, the embedding is injective, and the pushforward only ever feeds poles `iota(e) - c` with `e` in range. No scatter closes it.

So there are two things to test, and now two tests:

**`no_pushforward_can_close_the_pole_an_out_of_range_read_opens`** builds the leaf tables by hand for a reader naming entry 4 of a four-entry table, under two scatters standing in for every choice — the row's weight dropped, and the row's weight moved onto entry 0 where it would cancel if the two poles could be confused. Neither sum vanishes.

**`an_out_of_range_read_hidden_behind_a_zero_weight_is_rejected`** takes the only escape left. A prover that gives the offending row weight zero, on the reader side and in the pushforward alike, *does* get a zero-summing input, so the reduction runs. The test then plays the whole prover by hand — GKR, honest position claims (the row still names entry 4), the product sumcheck, the column claims — assembles a `LogupStarProof`, and drives the real `verify`:

```rust
assert_eq!(
    proof.verify(&lookups, &mut verifier_challenger).map(|_| ()),
    Err(LogupStarError::LeafNumerator)
);
```

The numerator side is entirely public, so the verifier rebuilds the honest weight from the claim point and catches the forged one. The forged pushforward is even built to support the statement's claim, so `ProductClaimedSum` passes and the rejection is the one being tested.

The PR description now says this rather than asserting the property bare.

multi-stark/src/fractional_gkr/test.rs
tcoratger: Fair, and thank you for running that probe properly — that is more evidence than the test was giving.

Added the golden you suggest. `a_fixed_base_leaf_instance_always_produces_the_same_proof` sweeps `n = 1..8` at a fixed seed, runs `LeafNumerator::Base` against both denominator storages, and digests the concatenation of the postcard proof bytes, the two output openings, and the next challenger sample. One Keccak digest is asserted. Green with and without `--features parallel`.

The next-sample part is there so a change the proof bytes cannot show still surfaces.

The two cross-storage tests stay — they pin something the golden does not, namely that the two arms agree with each other — but the fixed reference is now the golden, and the PR description claims that rather than byte-identity across the arms.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Retitled `feat(multi-stark)!:`, and the description now names the break explicitly under "Fractional GKR".

I kept the enum rather than adding a thin base-field entry point. Two entry points is two signatures to keep in step through the follow-up that adds a committed pushforward, and the call site is one line:

```rust
Fraction { n: LeafNumerator::Base(&n), d: &d }
```

`LeafNumerator` and `LeafFraction` are exported from `fractional_gkr`, so a downstream caller has the types it needs.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Fixed with the assertion you describe.

`prove_fractional_gkr` now rejects a mixed pair at the top, before any of the size-dependent paths:

```rust
if let LeafNumerator::Ext(numer) = fraction.n {
    assert_eq!(
        matches!(numer, PolyMaybePacked::Packed(_)),
        matches!(fraction.d, PolyMaybePacked::Packed(_)),
        "{MIXED_LEAF_STORAGE}"
    );
}
```

The message is a named constant so the assertion and the `# Panics` line cannot drift apart, and the doc lists it. The `_ => unreachable!` arms downstream are now genuinely unreachable, including at `n = 1` and at the scalar-unpack shim, which were the two cases that used to accept a mixed pair silently.

multi-stark/src/logup_star/prover.rs
tcoratger: Fixed by packing, which was the better half of your suggestion — the arms now run in production rather than only in one test.

`LogupStarProof::prove` packs both leaf tables when the field has lanes to fill and the padded cube is wide enough to address them:

```rust
let lanes = log2_strict_usize(F::Packing::WIDTH);
if lanes > 0 && plan.num_variables >= lanes { /* pack both */ }
```

A field with one lane per element only pays for the copy, so it stays scalar — which is every binary tower here, and is why the round trips still cover both storages.

The position claims are read off `witness[..].positions` directly rather than off the packed table, so nothing needs unpacking.

multi-stark/src/logup_star/witness.rs
tcoratger: Fixed. The scatter is now one parallel pass over every row of the table, readers chained together, rather than a sequential fold over readers each of which fans out:

```rust
(0..table_witness.readers.len())
    .into_par_iter()
    .flat_map(|index| positions[index].par_iter().zip(weights[index].par_iter()))
    .par_fold_reduce(|| EF::zero_vec(num_entries), scatter, merge)
```

Allocation and merging are now per split rather than per reader times split, so the cost is rows plus splits times `2^m` rather than readers times splits times `2^m`. Peak memory is splits times table, once, instead of once per reader.

The reader weights are still expanded per reader, but that is the tensor itself and is proportional to rows.

On the clone in `ProductProver::new`: I left it. Taking ownership means the proof needs its own copy instead, since `pushforwards` ends up in `LogupStarProof` — so it moves the allocation rather than removing it, and it is one `2^m` copy against a sumcheck that is `2^m` times width. Happy to change it if you would rather have the shape than the cost.

multi-stark/src/logup_star/transcript.rs
tcoratger: Registered. `logup_star_cases()` sits next to `lookup_cases()` in `examples/tests/transcript_separation.rs`, with the plain shape (two tables of different sizes, three readers between them) and two single-field moves of it — one widening a table, one moving a reader from the first table to the second while keeping every total unchanged. `NUM_PROTOCOLS` is 22.

That second move is the case from your `transcript.rs:236` comment, so it is now checked both inside the crate and across protocols.

Making this possible is what drove the shape refactor there: `LogupStarShape` now exposes `tables: Vec<LogupStarTableShape>` with public fields, mirroring `LookupShape` / `LookupInstanceShape`, so the suite can build one without a plan.

multi-stark/src/logup_star/prover.rs
tcoratger: Moved. The block now sits on `pub fn prove`, and `verify`, `check_shape` and `rebuild_opening` were checked for the same mistake — none of them had it.

multi-stark/src/lib.rs
tcoratger: Partly taken.

`position` is now `pub(crate)`, which was the accidental part of the surface — tests reach it in-crate and nothing outside needs it.

`transcript` stays public. It is what the cross-protocol separation suite needs (your `transcript.rs:59` comment), and every sibling protocol in this crate publishes the same three types: `LookupShape` / `LookupProverTranscript` / `LookupVerifierTranscript`, `ZerocheckShape` and so on. Hiding this one would make it the exception and would take the new protocol name out of the suite that checks names.

`prove` / `verify` / `LogupStarPlan` stay public too. The shape that the committed-pushforward follow-up changes is the proof, and that is a semver break either way; marking the module experimental would not spare it one.

tcoratger:
Superseded by #2145, which is the same branch moved to `Plonky3/Plonky3` and rebased onto current `main`. The commits are identical.

The reason for the move is mechanical: GitHub's stacked pull requests require every branch of a stack to live in the same repository, and this one was opened from my fork. The follow-up work is a stack, so its base had to move:

```text
    #2145  feat/logup-star               the reduction (this PR)
    #2144  feat/logup-star-openings      one opening schedule per commitment
    #2146  feat/logup-star-declaration   declare and plan indexed lookups
```

@Nashtare — nothing here is outstanding. Your review pass is fully addressed: all 15 threads are resolved above, and the fixes travelled with the commits. The reasoning lives in this PR's threads, which is why I am closing rather than deleting it, and #2145 links back here for that history.

Changes since your last read, all on #2145:

- the tampered-pushforward test now pins the reduction as the rejecting party, which is what pins the ordering you asked about;
- the fractional-GKR golden digest is target-independent, after CI caught it varying with the machine's SIMD lane count;
- rebased twice onto the moving `main`, most recently past #2140.

PR #2134, opened 2026-09-14
feat(sumcheck): add a univariate-skip round for binary zerochecks

tcoratger:
A zerocheck over a bit-valued witness widens every surviving cell in its first round: one bit becomes a full extension element, and nothing has shrunk the hypercube yet. That round costs more than all the others put together, and the plain driver pays it once per variable.

Reading `k` coordinates as points of an `F_2`-linear subspace replaces those `k` rounds with one message over that subspace:

```text
    plain:  m rounds, the first widening 2^(m-1) cells
    skip:   1 subspace round, then m - k ordinary rounds
```

The round hands back an ordinary multilinear claim, so `generic_degree` finishes it unchanged.

Ported from Gruen ([2024/108](https://eprint.iacr.org/2024/108)), Dao–Thaler ([2024/1210](https://eprint.iacr.org/2024/1210)), and Flock Sections 4.2 and 4.3 ([2026/1329](https://eprint.iacr.org/2026/1329)). Zerocheck is 37–42% of Flock's prover time.

## The domain

`SkipDomain` carries the subspace the round polynomial vanishes on and the larger subspace it is transmitted on, the smaller being the index prefix of the larger. Points come from the Cantor basis, so the domain coincides with the additive-NTT domain of the same dimension.

The vanishing polynomial is the normalised subspace polynomial, built by squaring rather than by multiplying out `2^k` factors. Its derivative is constant on the subspace and its value is constant on each coset, which collapses the resample from the subspace to the extension into a Cauchy matrix.

Sizing follows the composition's degree: the transmitted values have to determine the quotient by the vanishing polynomial, which needs one extra dimension per doubling of the degree.

## The extension table

`CompressedLde` is the workhorse. Splitting a row into byte chunks splits the resampling matrix into column blocks, and the naive tabulation would keep one table per chunk position. One identity collapses them:

```text
    M[i][8b + j] = M[i xor 8b][j]
```

so every chunk position reads the same table at an index offset by the position. For a six-bit skip of a degree-two composition that is 16 KiB instead of 128 KiB, which is the whole point: the shared table stays in the first-level cache.

The identity holds because the matrix is Cauchy in the two point sets, the index-to-point map is `F_2`-linear, and the numerator is constant along any shift by a subspace element.

Measured on one row, against the Lagrange evaluation it stands in for:

| path | time |
| --- | --- |
| compressed table | 18.3 ns |
| Lagrange reference | 2.82 µs |

These two measurements are **not** an end-to-end result. The reference does `2^k` inversions per output, which flatters the table enormously, and no benchmark compares a full skip-round zerocheck against the plain `generic_degree` driver on the same instance. That claim is unmeasured.

The benchmark is not committed — it exists to justify the table, not to be
maintained. Dropping it into `sumcheck/benches/univariate_skip.rs` with a
matching `[[bench]]` entry reproduces the numbers under
`RUSTFLAGS="-Ctarget-cpu=native" cargo bench -p p3-sumcheck --features parallel`.

<details>
<summary>benchmark source</summary>

```rust
//! What the univariate-skip round costs, against the work it replaces.

use criterion::{Criterion, criterion_group, criterion_main};
use p3_binary_field::{BinaryField8, BinaryField128};
use p3_multilinear_util::poly::Poly;
use p3_sumcheck::univariate_skip::{CompressedLde, SkipDomain, SkipRound, extend_reference};
use rand::rngs::SmallRng;
use rand::{RngExt, SeedableRng};

/// The subspace the round runs over lives in a byte field.
type F = BinaryField8;

/// Challenges and residual values live in the 128-bit field above it.
type EF = BinaryField128;

/// Six skipped variables, the width a product-form zerocheck uses.
const LOG_SKIP: usize = 6;

/// Rows extended per measurement, chosen so the batch is large enough to leave the cache.
const NUM_ROWS: usize = 1 << 14;

fn random_rows(rng: &mut SmallRng, row_bytes: usize) -> Vec<u8> {
    (0..NUM_ROWS * row_bytes)
        .map(|_| rng.random::<u8>())
        .collect()
}

/// The tabulated extension against the Lagrange evaluation it stands in for.
fn extension(criterion: &mut Criterion) {
    let mut rng = SmallRng::seed_from_u64(0x5C1);
    let domain = SkipDomain::<F>::for_degree(LOG_SKIP, 2).unwrap();
    let lde = CompressedLde::new(&domain).unwrap();
    let rows = random_rows(&mut rng, lde.num_chunks());

    // The batch path is what a prover actually runs, so it is measured per row extended.
    {
        let mut batch = criterion.benchmark_group("univariate-skip/extend-batch");
        batch.throughput(criterion::Throughput::Elements(NUM_ROWS as u64));
        let mut out = F::zero_vec(NUM_ROWS * lde.num_transmitted());
        batch.bench_function("compressed-table", |bencher| {
            bencher.iter(|| lde.extend_batch(&rows, &mut out));
        });
    }

    // One row each, so the table and the definition it stands in for compare directly.
    //
    //     table:      one lookup and one exclusive-or per output per chunk
    //     reference:  one inversion per subspace point per output
    {
        let mut single = criterion.benchmark_group("univariate-skip/extend-row");
        single.throughput(criterion::Throughput::Elements(1));
        let row = &rows[..lde.num_chunks()];
        let mut one = F::zero_vec(lde.num_transmitted());
        single.bench_function("compressed-table", |bencher| {
            bencher.iter(|| lde.extend(row, &mut one));
        });
        single.bench_function("lagrange-reference", |bencher| {
            bencher.iter(|| extend_reference(&domain, row, &mut one));
        });
    }
}

/// The two halves of a skip round, at the scale a zerocheck runs them.
fn round(criterion: &mut Criterion) {
    let mut rng = SmallRng::seed_from_u64(0x11);
    let round = SkipRound::<F>::new(LOG_SKIP, 2).unwrap();
    let rows = random_rows(&mut rng, round.row_bytes());
    let stride = round.num_transmitted();

    let mut group = criterion.benchmark_group("univariate-skip/round");
    group.throughput(criterion::Throughput::Elements(NUM_ROWS as u64));

    // Weighing the composed rows is the only part that touches the large field.
    let composed = F::zero_vec(NUM_ROWS * stride);
    let eq = Poly::new_from_point(
        &(0..14).map(|_| rng.random::<EF>()).collect::<Vec<_>>(),
        EF::ONE,
    );
    group.bench_function("round-message", |bencher| {
        bencher.iter(|| round.round_message::<EF>(&composed, eq.as_slice()));
    });

    // Reading every row at the challenge is what the residual rounds start from.
    let lambda = rng.random::<EF>();
    group.bench_function("bind-rows", |bencher| {
        bencher.iter(|| round.selector::<EF>(lambda).bind(&rows));
    });

    group.finish();
}

criterion_group!(benches, extension, round);
criterion_main!(benches);
```

</details>

## Pinned challenges

`PinnedEqWeights` fixes zerocheck coordinates ahead of time so their multipliers stop being generic field elements. Choosing them as `r_i = u_i / (1 + u_i)` with `u_i = g^(2^(i-1))` makes the weights a geometric progression `C * g^int(b)`.

Pinning narrows the verifier's challenge to an affine subspace, so it is argued rather than assumed: the weights must be linearly independent over the prime field, and every constructor rejects a dependent set. The caller owes the other half, that the constraint is prime-field-valued on the hypercube, which holds for a bit-valued witness.

**The progression needs a full-degree generator**, not a particular basis: `{g^k : k < 2^d}` is independent exactly when `g` has degree at least `2^d` over `F_2`. A generator from a small subfield is refused — `from_repr(2)` in the tower basis spans only `GF(4)`, so it fails at two coordinates — while the same full-degree element works in either representation. What the basis decides is only whether multiplying by `g` is a shift, which is why Flock and Binius64 fix GHASH.

The soundness precondition is the sharp edge here: the constraint must be prime-field-valued on the hypercube **for every witness the prover can commit to**, which means it comes from the commitment alphabet. A booleanity constraint in the same zerocheck does not supply it.

## The round and its transcript

`SkipRound` builds the message, reads it back at the challenge, and binds every row there through one table. The verifier defines the round polynomial itself, as the one of degree below the extension size that vanishes on the subspace and matches the transmitted values: vanishing is imposed, never checked, which is what makes it a zerocheck.

`UnivariateSkipShape` states the transcript once for both sides. The message is bound before the challenge is drawn, and the residual sumcheck runs as a bracketed sub-protocol under its own seed, the way ring-switching already delegates.

`p3-binary-field` moves from a dev-dependency to a dependency, since the module is bound to binary fields by its trait bounds.

## Tests

50 unit tests and 8 integration tests. The tabulated extension is pinned to direct Lagrange interpolation by proptest across every subspace dimension and extension width. The integration test proves a 10-variable binary zerocheck end to end through the real transcript, and rejects a broken witness, a tampered message, a message of the wrong width, and a proof replayed under a different shape.

## What this does not do

No multi-STARK wiring. `MultiStarkConfig::Val` is the trace field, so getting these economics into the AIR path needs a small-field (`Gf2`) commitment via ring-switching first. That is a separate change.

## Review

Reviewed at `a8a4261f` and approved with no HIGH findings. Every finding is addressed in the follow-up commit; two were real defects rather than documentation:

- a malformed residual proof panicked the reference verifier instead of rejecting, because a delegated rejection left the outer transcript unfinalized;
- nothing tested that the round message is bound before the skip challenge is drawn — two mutations of the transcript passed the whole suite.

Both now have tests that fail under the exact mutations that exposed them.


sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -0,0 +1,722 @@
...
+///
+/// The combination is then nonzero whenever some piece is.
+///
+/// That happens exactly when the weights are independent over the prime field.
+///
+/// Every constructor here rejects a dependent weight set for that reason.
+///
+/// The caller owes the other half of the argument:
+///
+/// - **The constraint must take values in the prime field on the hypercube.**
```
Nashtare: **MEDIUM — Pinned weights are only sound if the commitment itself forces every cell into `F_2`. A booleanity constraint in the same zerocheck does not, and the doc does not rule that case out.**

The independence argument needs every value `C(w(b))` the prover can produce on the hypercube to be in `F_2`. Any EF-valued pattern in the `F_2`-kernel of `y ↦ Σ_b w_b y_b` is invisible to the check, and that kernel has dimension `2^n·128 − 128`.

Probe: `geometric(x, 3)` over `Ghash128`, constraint `v^2 + v` (which is `F_2`-linear in `v`), 4 free variables plus 3 pinned.
- Witness: `v_40 = 0xab427352adc7b46fe51aae9c426e560a`, `v_41 = x`, every other cell 0.
- The constraint is nonzero at 2 cells.
- `Σ_b eq((r_free, pinned), b)·(v_b^2+v_b) = 0` exactly, for 16 random `r_free`. The zerocheck accepts with probability 1.
- An even simpler pattern: the block `(w_1, w_0, 0, …)` folds to 0.

Suggested doc fixes:
- Say the `F_2` premise must hold for *every* witness the prover can commit to. That means it is enforced by the commitment alphabet (an `F_2`-packed commitment, or ring switching), never by constraints of the same zerocheck.
- Note that batching `F_2`-valued constraints with EF coefficients sampled after commitment is still sound.
- Add the matching caveat to `SubfieldFoldTable`: it is only safe on off-hypercube message values. A byte-valued hypercube would need `F_{2^8}`-independence, which is not checked.
- Consider keeping the type `pub(crate)` until a consumer with an `F_2` commitment exists.

sumcheck/tests/univariate_skip_zerocheck.rs
```diff
@@ -0,0 +1,382 @@
...
+        return false;
+    }
+
+    // Replay the residual rounds inside the same bracket the prover used.
+    let replayed = transcript.residual_sumcheck(|challenger| {
+        proof
+            .residual
+            .verify(challenger, LOG_ROWS, RESIDUAL_DEGREE, pow_bits)
+    });
+    let Ok((point, final_sum)) = replayed else {
```
Nashtare, line 274: **MEDIUM — On a malformed residual proof the reference verifier panics instead of rejecting, because the outer transcript is dropped unfinalized.**

When `proof.residual.verify` returns `Err`, this branch returns `false` without calling `transcript.abort()`. The outer `VerifierState` drop check then panics. That check is active in release builds whenever `panic = "unwind"`.

Probed mutations, each of which panics with `Dropped unfinalized VerifierState` (`challenger/src/fs/state/verifier.rs:71`):
- one residual round popped
- `round_polys` cleared
- one poly one element too wide
- an extra round
- `pow_witnesses` cleared

With `transcript.abort()` added on this branch, all five reject cleanly. No existing test mutates the residual proof.

Fix: call `abort()` here and add a malformed-residual test. Better, have `UnivariateSkipVerifierTranscript::residual_sumcheck` take a `Result`-returning closure and abort on `Err`, so no consumer can repeat this.

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -0,0 +1,570 @@
...
+        // The width comes from the proof, so a mismatch is a rejection.
+        self.state
+            .observe_extensions::<F, EF, FieldToFieldCodec<F>>(ROUND_MESSAGE, message)
+            .map_err(|_| UnivariateSkipTranscriptError::MessageLenMismatch {
+                expected: self.shape.message_len(),
+                actual: message.len(),
+            })?;
+
+        // Re-run the prover's grinding step on the witness it committed to.
+        if self.shape.pow_bits > 0 {
```
Nashtare, line 340: **LOW — With `pow_bits = 0` the verifier accepts a proof that carries `Some(pow_witness)`, so proofs are not canonical.**

When grinding is off, the witness is silently ignored. Probe: an honest proof with `pow_witness = Some(12345)` is accepted at `pow_bits = 0`.

`GenericDegreeProof::verify` rejects the same situation (`generic_degree/proof.rs:114-123`: every accepting proof has a unique form).

Fix: when `pow_bits == 0 && witness.is_some()`, abort the state and return an `UnexpectedPowWitness` error.

sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -0,0 +1,722 @@
...
+
+        // Mutation: append the sum of two of them, which the earlier ones already span.
+        //
+        //     1 + 2 = 3  ->  already in the span of {1, 2, 4}
+        let dependent = [1u128, 2, 4, 3].map(Ghash128::from_repr);
+        assert!(!prime_field_independent(&dependent));
+    }
+
+    #[test]
+    fn the_progression_needs_a_polynomial_basis_and_is_refused_in_a_tower_basis() {
```
Nashtare: **LOW — The claim that the progression needs a polynomial basis and is refused in a tower basis is wrong: `F_2`-independence of `{g^j}` depends only on the degree of `g`, not on the basis.**

The module already says so at line 394.

Probe:
- `PinnedEqWeights::geometric(BinaryField128::from(Ghash128::from_repr(2)), 7)` returns `Ok`.
- 200 of 200 random `BinaryField128` generators return `Ok`.
- 0 of the bit patterns `2..4002` return `Ok`, because they all lie in the `GF(2^16)` sub-tower. That is what the PR description's "sweep of 4000 tower generators" actually measured.

What genuinely depends on the basis is only that multiplying by `g` is a shift.

Fix: rename the test (for example `a_subfield_generator_is_refused`) and reword it and the PR description: small tower bit patterns lie in small subfields, and a polynomial basis is what makes the progression a shift.

sumcheck/tests/univariate_skip_zerocheck.rs
```diff
@@ -0,0 +1,382 @@
...
+        return false;
+    };
+    transcript.finish();
+
+    // The verifier would reach the operand values through a commitment opening.
```
Nashtare, line 279: **LOW — The closing check stands in for two protocol steps that are documented nowhere.**

1. The zerocheck point comes from a seeded `rng` (line 268), not from the transcript after the witness commitment. The skip transcript neither absorbs `r` nor draws it, so the outer protocol has to.
2. `f(ρ, λ) = Σ_c L_c(λ)·f̃(ρ, c)` is an inner product of the committed multilinear's partial evaluation with the Lagrange vector. It is not a single multilinear opening, and needs a further k-round degree-2 sumcheck (or an equivalent reduction) down to `f̃(ρ, τ)`.

The module doc's "the round hands back an ordinary multilinear claim" is true of the residual sum, but not of the final opening.

Fix: state both obligations on `SkipRound` and in this comment, so the ring-switching follow-up does not treat `selector.bind(..)` evaluated at `ρ` as directly openable.

sumcheck/tests/univariate_skip_zerocheck.rs
```diff
@@ -0,0 +1,382 @@
...
+    // The round polynomial then stops vanishing on the skipped subspace.
+    //
+    // The verifier's reconstruction still does, so the two disagree and the check fails.
+    let (round, mut witness, point) = fixture(0xBAD);
+    witness.c[0] ^= 1;
+    let proof = prove(&round, &witness, &point, 0);
+    assert!(!verify(&round, &witness, &point, &proof, 0));
+}
+
+#[test]
```
Nashtare, line 349: **MEDIUM — No test checks that the message is bound before the skip challenge is drawn, or that the challenge is random.**

Two mutations of `transcript.rs` pass every unit and integration test:

- absorbing `message` with entries 1.. replaced by zero (same declared width);
- replacing the drawn skip challenge with `EF::GENERATOR` on both sides.

`a_tampered_round_message_is_rejected` does not catch either one. It tampers after proving and never re-derives a consistent residual proof, so `claimed_sum != evaluate(...)` rejects it whether or not the message is bound.

The attack that binding exists to stop is a prover that picks the message after seeing λ. Add that test:

```rust
#[test]
fn a_message_adapted_to_the_challenge_is_rejected() {
    let (round, mut witness, point) = fixture(0xADA);
    witness.c[0] ^= 1;
    let eq = Poly::new_from_point(&point, EF::ONE);
    let honest = round_message(&round, &witness, &eq);

    // Learn the challenge the honest message induces.
    let mut ch = fresh_challenger();
    let mut t = UnivariateSkipProverTranscript::<Challenger, EF, EF>::new(&mut ch, shape(&round, 0));
    let (lambda, _) = t.round_message(&honest);
    core::mem::forget(t);

    // The residual sum the broken witness really has at that challenge.
    let sel = round.selector::<EF>(lambda);
    let ops = witness.operands().map(|p| sel.bind(p));
    let truth: EF = (0..eq.num_evals())
        .map(|i| eq.as_slice()[i] * (ops[0].as_slice()[i] * ops[1].as_slice()[i] + ops[2].as_slice()[i]))
        .sum();

    // Patch the last entry so the read-back matches that sum.
    let last = honest.len() - 1;
    let mut unit = vec![EF::ZERO; honest.len()];
    unit[last] = EF::ONE;
    let mut cheat = honest.clone();
    cheat[last] += (truth - round.evaluate(&honest, lambda)) * round.evaluate(&unit, lambda).inverse();

    // Run the real prover transcript on the patched message.
    let mut ch = fresh_challenger();
    let mut t = UnivariateSkipProverTranscript::<Challenger, EF, EF>::new(&mut ch, shape(&round, 0));
    let (lambda2, pow_witness) = t.round_message(&cheat);
    let sel = round.selector::<EF>(lambda2);
    let operands = witness.operands().map(|p| sel.bind(p));
    let claimed_sum = round.evaluate(&cheat, lambda2);
    let (residual, _) = t.residual_sumcheck(|c| {
        ResidualProver { eq, operands }.prove::<EF, _>(c, LOG_ROWS, RESIDUAL_DEGREE, 0, claimed_sum)
    });
    t.finish();
    assert!(!verify(&round, &witness, &point, &Proof { message: cheat, pow_witness, residual }, 0));
}
```

It passes on HEAD and fails under both mutations.

sumcheck/src/univariate_skip/lde.rs
```diff
@@ -0,0 +1,465 @@
...
+            // The table path and the Lagrange path must agree entry for entry.
+            let mut fast = BinaryField8::zero_vec(lde.num_transmitted());
+            let mut slow = BinaryField8::zero_vec(lde.num_transmitted());
+            lde.extend(&row, &mut fast);
+            extend_reference(&domain, &row, &mut slow);
+            prop_assert_eq!(fast, slow);
+        }
+
+        #[test]
+        fn tabulated_extension_matches_across_domain_shapes(
```
Nashtare, line 451: **LOW — The shape proptest covers only k∈{3,4,5} and e∈{1,2} over `BinaryField16`, but the PR description says it covers "every subspace dimension and extension width".**

The k=6 fixed-shape test adds (6,7) over `BinaryField8`. Nothing exercises:

- k≥6 with e≥2;
- e≥3, which gives several cosets with block indices above `2^(k-3)`, where the `xor` has to stay inside the coset;
- the largest accepted shapes.

I checked every `BinaryField8` shape the constructors accept, plus `BinaryField16` (10,12) and (3,16), `BinaryField32` and `Ghash128` shapes. The identity holds everywhere, so this is a coverage gap, not a bug.

Fix: widen the ranges to `log_size in 3usize..=8, extra in 1usize..=4` over `BinaryField16` (still fast in release), or reword the description.

sumcheck/src/univariate_skip/lde.rs
```diff
@@ -0,0 +1,465 @@
...
+use thiserror::Error;
+
+use super::domain::SkipDomain;
+
+/// Bits one lookup covers.
+///
+/// # Why this value
+///
+/// - A byte indexes with a single load and needs no unpacking.
+/// - The resulting table stays in the first-level data cache at every supported skip width.
```
Nashtare: **LOW — The `CHUNK_BITS` rationale is inaccurate in two places.**

- **L1 claim.** "The resulting table stays in the first-level data cache at every supported skip width." The table holds `256 · (2^(k+e) − 2^k)` elements and nothing bounds k. Shapes the constructors accept go well past a 32–48 KiB L1: `BinaryField8` (7,8) is 32 KiB, `BinaryField16` (8,9) is 128 KiB, and `BinaryField16` (10,11) is 512 KiB. The claim only holds around the k=6 product-form shape.
- **"Sixteenth" wording.** Line 17 says four-bit chunks would "save a sixteenth of the table". A 4-bit table has 16 rows instead of 256, so it is one sixteenth of the size and saves 15/16.

Scope the L1 claim to the k=6, e=1 configuration and fix the wording.

sumcheck/src/univariate_skip/lde.rs
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        let stride = domain.num_transmitted();
+        let num_chunks = domain.size() / CHUNK_BITS;
+
+        // The resampling matrix holds one row per transmitted point.
+        // Only its first eight columns are ever read.
+        //
+        // The identity above reaches the rest by offsetting the output index instead.
+        let matrix = domain.resampling_matrix();
```
Nashtare: **LOW — `CompressedLde::new` builds the full `stride × 2^k` resampling matrix, one field inversion per entry, but reads only its first `CHUNK_BITS` columns.**

`SkipDomain::resampling_matrix` allocates and inverts `num_transmitted · size` entries, and the loop at line 145 reads `matrix[index * size + bit]` with `bit < 8`. Measured setup (probe, loaded host, `BinaryField16`, e=1):

| k | Entries built | `CompressedLde::new` |
| --- | --- | --- |
| 6 | 4 096 | 103 µs |
| 10 | 1 048 576 | 6.9 ms |
| 12 | 16 777 216 | 154 ms |

Only `8 · stride` entries are needed.

Fix: compute the eight columns directly, `scale_i · (t_i + s_j)^{-1}` for `j < 8`, or have `resampling_matrix` take a column count.

sumcheck/src/univariate_skip/lde.rs
```diff
@@ -0,0 +1,465 @@
...
+
+/// Extend one packed row by interpolating it directly, without a table.
+///
+/// This is the definition the tabulated path stands in for, and exists to test against.
+/// It costs one inversion per subspace point per output, so it is only usable in tests.
+///
+/// # Panics
+///
+/// Panics if either slice has the wrong length for this domain.
+pub fn extend_reference<F: TowerLevel>(domain: &SkipDomain<F>, row: &[u8], out: &mut [F]) {
```
Nashtare, line 273: **LOW — `extend_reference` is public and re-exported from `univariate_skip`, but its own doc says it "exists to test against" and "is only usable in tests". Its only non-test consumer is the benchmark in the PR description, which is not committed.**

That leaves a slow reference path as part of the crate's public API with no in-tree user outside `#[cfg(test)]`. Either:

- make it `pub(crate)` or `#[cfg(test)]` (the bench can call the already-public `SkipDomain::interpolate` in a loop), or
- commit the bench and mark the function `#[doc(hidden)]`.

sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -0,0 +1,722 @@
...
+        self.block_len
+    }
+
+    /// Combine one block of subfield values under the tabulated weights.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the block length differs from the tabulated one.
+    #[must_use]
+    pub fn fold<Sub>(&self, block: &[Sub]) -> F
```
Nashtare: **LOW — `SubfieldFoldTable::fold::<Sub>` has its own `Sub` parameter, unrelated to the one `new::<Sub>` checked, so folding a wider subfield block silently truncates each value to its low byte and returns a wrong result.**

`new` asserts at runtime that `Sub` is byte-wide. `fold` then accepts any `TowerLevel` with `Repr: Into<u128>` and indexes with `value.to_repr().into() as usize & 0xff`.

Probe:

1. Build the table with `new::<BinaryField8>(&pinned)`.
2. Call `table.fold(&[BinaryField16::from_repr(0x1234); 4])`.

It compiles, does not panic, and returns a value different from `pinned.fold` of the embedded block.

Fix: carry the subfield in the type, `SubfieldFoldTable<F, Sub>` with `PhantomData<Sub>`, and drop the method-level parameter. Or take `&[BinaryField8]` concretely, since byte width is the only case supported.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -0,0 +1,760 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The round polynomial's value at each transmitted point.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the composed values are not one block per equality weight.
+    #[must_use]
+    pub fn round_message<EF>(&self, composed: &[F], eq: &[EF]) -> Vec<EF>
```
Nashtare, line 212: **LOW — The prover hot loop multiplies extension by subfield once per row per transmitted point, and the tables added to avoid such products are not connected to the round. No benchmark measures the end-to-end claim.**

`round_message` computes `weight * value` for `num_rows × stride` pairs. For `BinaryField128 × BinaryField8`, each product goes through `from_basis_coefficients_fn`: 16 byte-field products, on the `extension.rs` path that is not hardware CLMUL.

`PinnedEqWeights` and `SubfieldFoldTable` exist to turn exactly this into lookups plus XORs. Nothing in `SkipRound` or the integration test uses them, so they are unused public API.

The only perf evidence is:

- a per-row table-vs-Lagrange comparison, where the reference does `stride · 2^k` inversions per row;
- a `round-message` bench with an all-zero `composed` input.

Neither compares a full skip-round zerocheck (skip round plus `generic_degree` residual) against the plain `generic_degree` driver on the same instance. That is the claim that matters. Please commit such a benchmark, or say in the description that it is unmeasured.

The API also requires fully materialised per-operand extensions and composed values, `rows × stride` subfield elements each, before weighing. A per-row extend → compose → accumulate path would avoid that memory.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -0,0 +1,760 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the message length differs from the transmitted-point count.
+    #[must_use]
+    pub fn evaluate<EF>(&self, message: &[EF], lambda: EF) -> EF
+    where
+        EF: ExtensionField<F>,
+    {
+        assert_eq!(message.len(), self.num_transmitted(), "one value per point");
```
Nashtare: **LOW — `evaluate` is verifier code, yet it panics on a proof-controlled message length and pays one extension-field inversion per transmitted point.**

- **Panic.** `assert_eq!(message.len(), self.num_transmitted())` fires on a malformed proof. The integration test is safe only because `UnivariateSkipVerifierTranscript::round_message` rejects the width first. A caller that evaluates before replaying the transcript panics instead of rejecting. Return `Result`, or state in the doc that the transcript replay must come first.
- **Inversions.** The barycentric sum calls `(lambda + point).inverse()` 64 times in `BinaryField128` for k=6, and `RowSelector::new` repeats this over the subspace. Batch inversion (`p3_field::batch_multiplicative_inverse`) turns that into one inversion plus about 3n products.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -0,0 +1,760 @@
...
+
+        // A challenge landing on a domain point reads the value there rather than dividing by zero.
+        //
+        // The domain lives in the base field, so a challenge outside it cannot land on a point.
+        // Narrowing once keeps the search itself in the base field.
+        if let Some(base) = lambda.as_base() {
+            if let Some(index) = self.domain.transmitted().iter().position(|&t| t == base) {
+                return message[index];
+            }
+            if self.domain.subspace().contains(&base) {
```
Nashtare, line 284: **LOW — The branch of `evaluate` for a challenge on a subspace point is never tested.**

Replacing `return EF::ZERO` with `return message[0]` passes all 58 tests. Random challenges hit it with negligible probability, but it is the one place the imposed vanishing is written out as code.

Add a loop over `round.domain().subspace()` asserting `evaluate(msg, EF::from(s)) == EF::ZERO` for a nonzero message, mirroring `reading_the_message_on_a_transmitted_point_returns_that_value`.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -0,0 +1,760 @@
...
+/// Allowing degree up to the extension size costs the verifier nothing beyond that term.
+///
+/// It also saves it a consistency check on the honest degree bound.
+#[derive(Debug, Clone)]
+pub struct SkipRound<F> {
+    /// Where the round polynomial vanishes, and where it is transmitted.
+    domain: SkipDomain<F>,
+    /// The tabulated map from a row's bits to its transmitted values.
+    lde: CompressedLde<F>,
+    /// The vanishing polynomial of the whole extension, evaluated nowhere yet.
```
Nashtare: **LOW (nit) — The `log_extended` field duplicates `domain.log_extended()`, and its doc describes a polynomial "evaluated nowhere yet" while the field is a `usize`.**

Drop the field and read `self.domain.log_extended()` in `evaluate`, or document it as the dimension of the extension.

Nashtare (approved):
Reviewed at a8a4261f. No HIGH findings.

**The core round checks out**
- The domain matches `p3-binary-dft`'s points.
- `for_degree`'s sizing is sufficient and minimal, and all 30 byte-field shapes it accepts are complete.
- λ is a full extension-field challenge, drawn after the whole message is bound.
- The selector's Lagrange coefficients and the read-back interpolation are correct.
- `M[i][8b+j] = M[i xor 8b][j]` holds and stays within the coset. Checked on every accepted `BinaryField8` shape, plus wider fields.
- `eq(r,b) = C·g^{int(b)}` holds, and checking all `2^n` weights is exactly the right condition when every cell is `F_2`-valued.

**Main points**
- **`pinned.rs` (MEDIUM):** the `F_2` premise has to come from the commitment alphabet. A booleanity constraint in the same zerocheck does not provide it; a probe gives a non-bit witness whose pinned sum is zero.
- **Integration test (MEDIUM):** on a malformed residual proof the reference verifier panics (an unfinalized `VerifierState` is dropped) instead of rejecting. This test is the only usage example of the verifier API.
- **Tests (MEDIUM):** nothing catches the message going unbound or the challenge being fixed. A test is suggested inline.
- **PR description:** the claim that this needs a polynomial basis is inaccurate. Independence depends on the element, not the basis, and the 4000-pattern sweep only covered `GF(2^16)`.

**LOWs**
- verifier-path asserts;
- a zero-difficulty proof with a junk witness is accepted, so proofs are not canonical;
- `CompressedLde::new` builds the full matrix;
- `SubfieldFoldTable::fold` takes a `Sub` unrelated to the one checked at construction;
- test-only functions exposed as public API;
- no end-to-end benchmark against `generic_degree`.

**Validation**
- `cargo test -p p3-sumcheck --release` passes with and without `parallel`, including at `PROPTEST_CASES=2000`.
- clippy, rustdoc and fmt are clean.
- The crate still builds for wasm32 and thumbv7em.
- No conflicts with main.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: Fixed, and the probe is right: the fold's kernel over a wider alphabet is huge, so nothing in the pinned sum can catch a non-bit cell pattern. The premise was stated without saying *for which witnesses*, which is exactly the hole.

The soundness note now says it has to hold for every witness the prover can commit to, that it therefore comes from the commitment alphabet, and — spelled out — that **a booleanity constraint in the same zerocheck does not supply it**, because the pinned fold is what would have to catch the non-bit witness in the first place. The batching case is stated as sound provided the coefficients are drawn after the commitment.

`SubfieldFoldTable` carries the matching caveat: it is an arithmetic shortcut over weights certified only over the prime field, so it is safe on off-hypercube values and not as the pinned fold of a byte-valued hypercube.

Kept public rather than `pub(crate)`, but closed the sharp edge you found below by moving the subfield onto the type, and added a note that this is a reduction with no consumer in the crate yet.

sumcheck/tests/univariate_skip_zerocheck.rs
tcoratger: Fixed, and thank you — this was the worst thing in the PR. Reproduced your panic exactly (`Dropped unfinalized VerifierState`, `verifier.rs:71`).

Took the stronger fix you suggested rather than just adding `abort()` at the call site: the verifier's `residual_sumcheck` now takes a `Result`-returning closure and releases the driver itself on `Err`, so no consumer can repeat it.

    delegated Ok   ->  close the bracket, carry the value out
    delegated Err  ->  release the driver, carry the rejection out

New test `a_malformed_residual_proof_is_rejected_rather_than_panicking` drives all five of your mutations through the reference verifier. Removing the release reproduces the panic, so the test pins the fix rather than the symptom.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Fixed. `UnexpectedPowWitness` is returned when `pow_bits == 0 && witness.is_some()`, with the state aborted first, matching the canonical-shape rule `GenericDegreeProof::verify` already enforces.

Covered twice: `a_grinding_witness_at_zero_difficulty_is_rejected`, and as one of the five mutations in the malformed-residual test.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: You're right, and I had the reasoning backwards. Independence of `{g^j}` is a property of the generator's degree over `F_2`; the representation only decides whether multiplying by `g` is a shift. My "sweep of 4000 tower generators" measured small sub-tower bit patterns and nothing else, which is why it found none.

Test renamed to `a_generator_from_a_small_subfield_is_refused` and reworded around degree. It now also asserts the positive direction you probed — the same full-degree element in the tower basis returns `Ok` at 7 coordinates — so the test states what actually gates it.

PR description corrected too.

sumcheck/tests/univariate_skip_zerocheck.rs
tcoratger: Fixed — this was the most useful catch for whoever picks up the ring-switching follow-up, because both obligations were invisible.

`SkipRound` now carries a `# What the caller still owes` section stating (1) that this transcript neither absorbs nor draws the zerocheck point, so the surrounding protocol must draw it after the witness is committed, and (2) that the residual rounds end on

    f(rho, lam) = sum_c L_c(lam) * f~(rho, c)

an inner product with the subspace's Lagrange vector — one claim, but not one a commitment can open, needing a further k-round degree-two reduction. It says explicitly that handing a commitment the bound rows evaluated at the residual point proves a different statement.

The module header's "hands back an ordinary multilinear claim" was the wrong claim and is gone; it now separates the residual sum from the opening.

sumcheck/tests/univariate_skip_zerocheck.rs
tcoratger: Fixed, and your diagnosis of why `a_tampered_round_message_is_rejected` is worthless here was exactly right — it rejects on `claimed_sum != evaluate(..)` whatever the transcript does.

Added `a_message_adapted_to_the_challenge_is_rejected`, essentially your test. Avoided the `mem::forget` by factoring the prover into `prove_with_message`, which takes the message and reports the challenge it induced, so the adaptive prover is expressed without leaking a transcript.

Verified it does real work by applying both of your mutations:

| mutation | result |
| --- | --- |
| message entries 1.. absorbed as zero | new test FAILS, other 10 pass |
| skip challenge replaced by `EF::GENERATOR` | new test FAILS, other 10 pass |

So it catches precisely the gap, and nothing else was covering it.

sumcheck/src/univariate_skip/lde.rs
tcoratger: Fixed by widening rather than rewording: the shape proptest now runs `log_size in 3..=8, extra in 1..=4` over `BinaryField16`, which is your suggested range and covers k>=6 with e>=2, e up to 4, and the largest accepted shapes. Still fast.

Added a comment noting that the widest shapes are the ones where a block index runs past `2^(k-3)` and the offset has to stay inside its own coset, since that is the part of the identity the narrow range never exercised.

sumcheck/src/univariate_skip/lde.rs
tcoratger: Both wrong, both fixed.

The L1 claim is now scoped: table size is `256 * (2^(k+e) - 2^k)` and nothing bounds `k`, so the doc says the six-bit degree-two shape is 16 KiB and stays in L1, while wider skips leave it and the win becomes locality rather than residency.

The "sixteenth" wording was backwards — a 4-bit table is a sixteenth of the size. Reworded to "shrink the table to a sixteenth".

sumcheck/src/univariate_skip/lde.rs
tcoratger: Fixed. `SkipDomain::resampling_prefix(columns)` builds only the columns asked for, and the table builder asks for `CHUNK_BITS`. `resampling_matrix()` stays as `resampling_prefix(size())`, so the tests that walk the whole matrix are unchanged.

Setup work drops from `stride * 2^k` inversions to `stride * 8`, which is the factor in your `k = 12` measurement.

sumcheck/src/univariate_skip/lde.rs
tcoratger: Fixed, though slightly differently: it is now behind `#[cfg(any(test, feature = "test-util"))]`, both at the definition and the re-export. That matches how `ring_switch::test_util` is already gated in this crate, keeps it reachable for downstream tests, and means no production path can pick up the slow path by accident.

Not `#[cfg(test)]`-only because the benchmark is the natural consumer, and not committed-plus-`#[doc(hidden)]` because benchmarks are staying out of this crate's tree.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: Fixed, and this was a genuine hole rather than a nit — `new::<BinaryField8>` then `fold(&[BinaryField16; _])` compiles, does not panic, and silently truncates.

Took your first suggestion: the type is `SubfieldFoldTable<F, Sub>` with `PhantomData<Sub>`, and `fold` has no parameter of its own. The mismatch is now a type error rather than a wrong answer.

sumcheck/src/univariate_skip/round.rs
tcoratger: Fair on all three, and I have not closed the main one.

**Tables unwired.** Correct, and now stated rather than implied: `PinnedEqWeights` says it is a reduction of its own with no consumer in the crate yet, because a skip round weighs its rows with whatever equality table the caller hands it, and laying rows out so the pinned coordinates index blocks is the arithmetization's decision. Wiring them means making that layout choice, which belongs above this module.

**Benchmark.** You are right that neither existing measurement is the claim that matters. The per-row comparison is against a reference doing `2^k` inversions per output, which flatters it enormously, and the `round-message` bench runs on all-zero input. Benchmarks are being kept out of this crate's tree, so rather than commit one I have marked the end-to-end claim as unmeasured in the PR description. It should not be read as established.

**Materialisation.** Also correct: the API wants `rows * stride` subfield elements per operand up front. A per-row extend/compose/accumulate path would avoid it. Not changed here, since it is an API reshape rather than a fix.

sumcheck/src/univariate_skip/round.rs
tcoratger: Both fixed.

`evaluate` returns `Result<EF, MessageLenMismatch>` instead of asserting on a proof-controlled length, so a caller that evaluates before replaying the transcript rejects rather than aborts.

Batched the inversions with `batch_multiplicative_inverse`, in the barycentric sum and in `RowSelector::new`, turning `n` extension inversions into one plus about `3n` products in both places.

sumcheck/src/univariate_skip/round.rs
tcoratger: Fixed. Confirmed your mutation: `return message[0]` in place of `return EF::ZERO` passed the whole suite.

Added `reading_the_message_on_the_subspace_returns_zero`, which walks the subspace with a nonzero message and asserts zero — mirroring the transmitted-point test, as suggested. It fails under that mutation.

Worth pinning precisely because it is the imposed vanishing written out as code, and random challenges never reach it.

sumcheck/src/univariate_skip/round.rs
tcoratger: Fixed. The field is gone and `evaluate` reads `self.domain.log_extended()`. The doc was describing a polynomial while the field held a dimension, which is what made it confusing — dropping it removed both problems.

PR #2135, opened 2026-09-15
feat(sumcheck): reduce a skip round's opening to an evaluation point

tcoratger:
**Stack base.** First of the univariate-skip integration chain. Later PRs in the stack target this branch.

The skip round merged in #2134 does not end on something a commitment can open. Writing `rho` for the residual point and `lam` for the skip challenge, the claim it leaves on each committed polynomial is

```text
    f(rho, lam) = sum_c L_c(lam) * f~(rho, c)
```

a weighted blend of `2^k` positions. A commitment answers one question — the value at a point — so the blend has to go before anything can be opened. `SkipRound`'s own documentation states this obligation, and nothing discharged it. That makes it the one hard blocker to a working binary zerocheck.

## The reduction

The blend is itself a sum over the skipped cube, so a sumcheck collapses it:

```text
    sum_c L(c) * g(c)          L(c) = L_c(lam),  g(c) = f~(rho, c)
```

`k` rounds, degree two, through the ordinary driver, ending on `f~(rho, tau)`.

Both factors were already to hand, so this adds no new machinery: the Lagrange vector is what `RowSelector` builds, and `g` is the witness read over the row variables. The selector now keeps that vector rather than only its byte-aggregated form and hands it over — two computations of the same Lagrange values could disagree and nothing would catch it.

The fold costs **one addition per set bit and no multiplications**. The witness is bits, so it is an equality-weighted count over the rows, and the zero bits cost nothing.

Several committed polynomials share the Lagrange factor, so one challenge batches them into a single run, leaving one point for all of them.

## Verifier cost

```text
    L^(tau) = Z_S(lam) / Z_S'(S) * sum_c eq(tau, c) / (lam + s_c)
```

`2^k` terms and one batch inversion. No new asymptotic — reading the round message already costs `2^k`.

## Tests

The centrepiece is a proptest that runs the rounds to the end and asserts the surviving claim equals the committed multilinear at `(rho, tau)`, over a range of trace heights. That is the identity the reduction exists to establish.

Also pinned: the fold against a direct cell-by-cell reference; the blend the round leaves against the round's own binding, so the premise is checked rather than assumed; and that batching the witnesses and batching their claims use one Horner order — a mismatch there would prove the right sum against the wrong claim.

## Notes for review

- The transcript is a sibling shape with its own seed, bracketing its sumcheck the way ring-switching already delegates. A single-polynomial run has no batching step, so its description differs and one cannot be replayed as the other.
- A delegated rejection releases the driver, so a malformed proof is refused rather than aborting the verifier — the same fix as the review of #2134.
- This does **not** assemble the full chain. Owning the sequence commit → point → skip → residual → opening is the next PR in the stack.


sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -56,6 +57,242 @@ const RESIDUAL_SUMCHECK: &str = "residual_sumcheck";
...
+        EF: ExtensionField<F>,
+    {
+        // One challenge, then the bracket the rounds run inside.
+        //
+        // A single-polynomial run has no batching step at all.
+        //
+        // The description therefore differs, so one cannot be replayed as the other.
+        let mut steps = Vec::with_capacity(3);
+
+        if self.num_polynomials > 1 {
```
Nashtare, line 122: The batching challenge is drawn before anything claim-dependent is absorbed: only `claimed_sum = Σ γ^i v_i` enters, inside the nested seed, after γ. A prover who sees γ can shift `v_0 += γ·d, v_1 -= d`, and the honest sumcheck and openings still verify (I have a local probe that is accepted as-is and rejected once the claims are bound). Ring switching binds its evaluation point and tensor elements as Message steps before `batching_point`. Could this pattern gain an `opening_claims` Message step with `Length::Fixed(num_polynomials)` ahead of the challenge? That also binds the count, which the seed currently only sees as `> 1`.

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -56,6 +57,242 @@ const RESIDUAL_SUMCHECK: &str = "residual_sumcheck";
...
+
+    /// Close the transcript once every described step has been played.
+    ///
+    /// # Panics
+    ///
+    /// When fewer steps were played than the run was described with.
+    pub fn finish(self) {
+        self.state
+            .finalize()
+            .expect("the opening reduction reads an empty wire");
```
Nashtare, line 292: `reads an empty wire` / `carries every value in its own proof`: which proof carries the per-polynomial claims and the `f~_i(rho, tau)` values? Neither this transcript nor the skip transcript binds them. If the composing PR is meant to bind them, can that obligation be written on `SkipOpeningShape`, the way `SkipRound` documents the zerocheck point?

sumcheck/src/univariate_skip/opening.rs
```diff
@@ -0,0 +1,478 @@
...
+        assert!(!claims.is_empty(), "at least one claim");
+
+        // The same Horner order the witnesses were batched in.
+        claims
+            .iter()
+            .rev()
+            .fold(EF::ZERO, |accumulator, &claim| accumulator * gamma + claim)
+    }
+
+    /// Build the prover state for the reduction's sumcheck.
```
Nashtare, line 238: There is no verifier counterpart. `claimed_sum == batch_claims(v, γ)` and `final == lagrange_at(τ) · batch_claims(openings, γ)` are soundness-critical and easy to forget, and `GenericDegreeProof::verify` takes the claimed sum from the proof. Worth a `SkipOpening::verify` that owns both checks?

sumcheck/src/univariate_skip/opening.rs
```diff
@@ -0,0 +1,478 @@
...
+            let lambda = rng.random::<EF>();
+            let opening = SkipOpening::new(round.selector::<EF>(lambda).lagrange().clone());
+            let rho = Point::new((0..log_rows).map(|_| rng.random::<EF>()).collect());
+
+            // Run the rounds by hand, folding at fresh challenges.
+            let folded = opening.partial_evaluations([rows.as_slice()], &rho);
+            let mut prover = opening.prover(SkipOpening::batch(&folded, EF::ONE));
+            let mut tau = Vec::new();
+            for _ in 0..opening.num_variables() {
+                let _ = prover.round_poly();
```
Nashtare, line 467: `let _ = prover.round_poly();` means no test looks at the round polynomials. Replacing `round_poly` with `vec![ZERO; 2]` passes every test in the module, and so does forcing `batching_challenge` to `ONE`. An end-to-end prove/verify through the two new transcript drivers would catch both.

sumcheck/src/univariate_skip/opening.rs
```diff
@@ -0,0 +1,478 @@
...
+/// Several committed polynomials share the same Lagrange factor.
+///
+/// One challenge therefore batches them into a single run.
+///
+/// # Verifier cost
+///
+/// The verifier reads the Lagrange vector at the output point:
+///
+/// ```text
+///     L^(tau) = Z_S(lam) / Z_S'(S) * sum_c eq(tau, c) / (lam + s_c)
```
Nashtare, line 59: This formula is not what `lagrange_at` computes: it evaluates the stored vector, while the formula is undefined for `lam` in S. The verifier also has to build the whole prover-side `RowSelector` table just to get this vector. Maybe reword, or add a Lagrange-only constructor?

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -56,6 +57,242 @@ const RESIDUAL_SUMCHECK: &str = "residual_sumcheck";
...
+
+/// Numbers that fix the transcript of one opening reduction.
+///
+/// Both sides build this from their own configuration, never from a proof.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SkipOpeningShape {
+    /// Number of skipped variables the reduction binds, one round each.
+    pub num_variables: usize,
+    /// Number of committed polynomials batched into one run.
+    pub num_polynomials: usize,
```
Nashtare, line 80: `num_polynomials = 0` is accepted: `batching_challenge` returns `ONE` and `batch` then panics on the prover side. Reject it in `new`?

Nashtare (changes requested):
Reviewed at 1cda7846. The reduction itself is correct; the transcript is not.

**What checks out**
- The reduction identity and variable ordering: row variables are the prefix, skipped bits the suffix, and `fix_prefix_var_mut` pairs with the half-split in `round_poly`.
- Degree 2 at nodes 0 and 2 works over `BinaryField128`, and challenges are drawn in the extension.
- Completeness for every `lam` in S; the verifier never divides.
- Every tampered round-polynomial entry, opening value and single claim is rejected, as are wrong widths, a wrong round count, and an unexpected or bad pow witness. All return `Err` without panicking.

**Blocking**
- **Claims are unbound before γ (HIGH).** The per-polynomial claims `v_i` are never absorbed before the batching challenge. Only `Σ γ^i v_i` enters, after γ. A prover who sees γ can shift `v_0 += γ·d`, `v_1 -= d`, and the honest sumcheck and openings still verify; a local probe confirms the forgery is accepted, and rejected once the claims are bound first. Nothing in-tree composes this verifier yet, so it is latent, but the next PR in the stack will. Inline comment on `transcript.rs:122`.

**Also worth addressing**
- **No verifier entry point (MEDIUM).** The two soundness-critical equalities are left to every caller.
- **Tests (MEDIUM).** No test drives the new transcripts. `round_poly` returning zeros and `batching_challenge` returning `ONE` both pass all 55 `univariate_skip` tests.

**Soundness error to charge once claims are bound:** `2k/|EF| + (m−1)/|EF|`, independent of the skip degree. #2136 currently charges only the first part.

sumcheck/src/univariate_skip/transcript.rs
Nashtare, line 122: The batching challenge is drawn before anything claim-dependent is absorbed: only `claimed_sum = Σ γ^i v_i` enters, inside the nested seed, after γ. A prover who sees γ can shift `v_0 += γ·d, v_1 -= d`, and the honest sumcheck and openings still verify (I have a local probe that is accepted as-is and rejected once the claims are bound). Ring switching binds its evaluation point and tensor elements as Message steps before `batching_point`. Could this pattern gain an `opening_claims` Message step with `Length::Fixed(num_polynomials)` ahead of the challenge? That also binds the count, which the seed currently only sees as `> 1`.

sumcheck/src/univariate_skip/transcript.rs
Nashtare, line 292: `reads an empty wire` / `carries every value in its own proof`: which proof carries the per-polynomial claims and the `f~_i(rho, tau)` values? Neither this transcript nor the skip transcript binds them. If the composing PR is meant to bind them, can that obligation be written on `SkipOpeningShape`, the way `SkipRound` documents the zerocheck point?

sumcheck/src/univariate_skip/opening.rs
Nashtare, line 238: There is no verifier counterpart. `claimed_sum == batch_claims(v, γ)` and `final == lagrange_at(τ) · batch_claims(openings, γ)` are soundness-critical and easy to forget, and `GenericDegreeProof::verify` takes the claimed sum from the proof. Worth a `SkipOpening::verify` that owns both checks?

sumcheck/src/univariate_skip/opening.rs
Nashtare, line 467: `let _ = prover.round_poly();` means no test looks at the round polynomials. Replacing `round_poly` with `vec![ZERO; 2]` passes every test in the module, and so does forcing `batching_challenge` to `ONE`. An end-to-end prove/verify through the two new transcript drivers would catch both.

sumcheck/src/univariate_skip/opening.rs
Nashtare, line 59: This formula is not what `lagrange_at` computes: it evaluates the stored vector, while the formula is undefined for `lam` in S. The verifier also has to build the whole prover-side `RowSelector` table just to get this vector. Maybe reword, or add a Lagrange-only constructor?

sumcheck/src/univariate_skip/transcript.rs
Nashtare, line 80: `num_polynomials = 0` is accepted: `batching_challenge` returns `ONE` and `batch` then panics on the prover side. Reject it in `new`?

Nashtare (changes requested):
Reviewed at 1cda7846. The reduction itself is correct; the transcript is not.

**What checks out**
- The reduction identity and variable ordering: row variables are the prefix, skipped bits the suffix, and `fix_prefix_var_mut` pairs with the half-split in `round_poly`.
- Degree 2 at nodes 0 and 2 works over `BinaryField128`, and challenges are drawn in the extension.
- Completeness for every `lam` in S; the verifier never divides.
- Every tampered round-polynomial entry, opening value and single claim is rejected, as are wrong widths, a wrong round count, and an unexpected or bad pow witness. All return `Err` without panicking.

**Blocking**
- **Claims are unbound before γ (HIGH).** The per-polynomial claims `v_i` are never absorbed before the batching challenge. Only `Σ γ^i v_i` enters, after γ. A prover who sees γ can shift `v_0 += γ·d`, `v_1 -= d`, and the honest sumcheck and openings still verify; a local probe confirms the forgery is accepted, and rejected once the claims are bound first. Nothing in-tree composes this verifier yet, so it is latent, but the next PR in the stack will. Inline comment on `transcript.rs:122`.

**Also worth addressing**
- **No verifier entry point (MEDIUM).** The two soundness-critical equalities are left to every caller.
- **Tests (MEDIUM).** No test drives the new transcripts. `round_poly` returning zeros and `batching_challenge` returning `ONE` both pass all 55 `univariate_skip` tests.

**Soundness error to charge once claims are bound:** `2k/|EF| + (m−1)/|EF|`, independent of the skip degree. #2136 currently charges only the first part.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: You're right, and the probe is real. Fixed in 1234f43.

The pattern now carries an `opening_claims` Message step of `Length::Fixed(num_polynomials)` ahead of the challenge, exactly as suggested. Both drivers play it *in the same call* as the draw — `batching_challenge(&claims)` — so the ordering cannot be got wrong by a caller rather than merely being documented.

The count comes with it: a proof whose claim count differs from the description is now rejected (`SkipOpeningTranscriptError::ClaimCountMismatch`) instead of being replayed, which also closes the `> 1` gap you noted in the seed.

Your probe is now a test. `a_shift_that_preserves_the_batch_is_still_rejected` applies exactly `v_0 += gamma*d, v_1 -= d`, asserts the batch under the honest challenge is unchanged, and checks the replay refuses it with `StartingSumMismatch`. `the_batching_challenge_follows_the_claims` pins the underlying property that the challenge moves when the claims do.

(One thing the probe taught me: `EF::from_u64` collapses to the parity bit in characteristic two, so my first `delta` was silently zero and the test passed for the wrong reason. It draws the shift now.)

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Both were wrong, in opposite directions. Fixed in 1234f43.

The per-polynomial claims **are** bound here now — see the thread above — so the doc no longer overclaims about them. The wire genuinely stays empty: both sides bind values they already hold, the prover from the witness and the verifier from the proof, which is the same model the skip round's own message uses.

The opening values are the part that really is the caller's, and that obligation is now written on `SkipOpeningShape` under `# What the caller owes`, the way you asked — alongside a note that the two equalities tying those values to the run are checked by the reduction itself rather than left to a caller.

sumcheck/src/univariate_skip/opening.rs
tcoratger: Agreed, and worth it. `SkipOpening::verify` in 1234f43 owns both:

```rust
if starting_sum != Self::batch_claims(blends, gamma) { StartingSumMismatch }
if closing_value != self.lagrange_at(tau) * Self::batch_claims(openings, gamma) { ClosingValueMismatch }
```

plus a count check, since two batches over different widths weigh different things.

The first is the one your point about `GenericDegreeProof::verify` makes load-bearing: the driver reads its starting sum from the proof, so nothing inside the sumcheck ties it to the claims the round left behind.

Each has its own test (`a_tampered_starting_sum_is_rejected`, `a_tampered_opening_is_rejected`, `blends_and_openings_must_agree_on_count`).

sumcheck/src/univariate_skip/opening.rs
tcoratger: Caught fairly — that `let _` was covering nothing. Replaced in 1234f43 by a full prove-and-verify through both transcript drivers, ten tests in place of the one.

On your two mutations:

- **Zeroed `round_poly`:** now covered, though not where I expected. A degree-two round reports two nodes and recovers the third from the running claim, so no round is self-contradicting and the sumcheck alone still accepts. The tamper survives to the end and the *closing* check against the committed openings refuses it. `a_tampered_round_polynomial_is_rejected` asserts that specific error and the comment records why the sumcheck is a reduction and not a filter here.
- **Forced `batching_challenge` to ONE:** covered by `the_batching_challenge_follows_the_claims`, which pins that the challenge is drawn and depends on the claims. A round trip alone would not have caught this one — both sides would agree on ONE — so it needed its own test rather than more end-to-end coverage.

sumcheck/src/univariate_skip/opening.rs
tcoratger: Both parts right. Fixed in 1234f43.

**The formula.** Reworded: the verifier evaluates the stored vector at the output point (`2^k` terms), and the closed form is what *building* the vector costs off the subspace. The subspace case is now stated rather than elided — a challenge landing on a subspace point makes one entry one and the rest zero, which the barycentric form cannot express.

**The table.** `SkipRound::lagrange` builds the vector alone. The 32 KiB of byte-indexed partial sums only ever serve the prover's row reading, so a verifier no longer pays for them. The arithmetic is shared with `RowSelector::new` rather than duplicated, which the existing comment there already insisted on: *"both must come from one place: two computations of the same Lagrange vector could disagree and nothing would catch it."* `the_verifier_builds_the_same_lagrange_vector_as_the_prover` pins that, at random challenges and at every subspace point.

The zerocheck driver downstream (#2139) now takes this path too.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Fixed in 1234f43: `SkipOpeningShape::new` refuses it, with `an_empty_batch_is_refused` covering it.

Rejecting at construction rather than at the draw is the right place — an empty batch has nothing to reduce, and every later step assumes one claim at least, so there is no shape for it to be.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Same finding as the thread above on this line — answered there. Fixed in 1234f43: the claims are bound as a `Length::Fixed(num_polynomials)` Message step ahead of the challenge, in the same call as the draw, and your value-shifting probe is now a test.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Same finding as the thread above on this line — answered there. Fixed in 1234f43: the claims are bound here now, and the opening obligation is written on `SkipOpeningShape`.

sumcheck/src/univariate_skip/opening.rs
tcoratger: Same finding as the thread above on this line — answered there. `SkipOpening::verify` in 1234f43 owns both checks.

sumcheck/src/univariate_skip/opening.rs
tcoratger: Same finding as the thread above on this line — answered there. Replaced by a full prove-and-verify through both drivers, with both of your mutations covered.

sumcheck/src/univariate_skip/opening.rs
tcoratger: Same finding as the thread above on this line — answered there. Reworded, and `SkipRound::lagrange` now builds the vector without the byte table.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Same finding as the thread above on this line — answered there. Refused in `SkipOpeningShape::new`.

tcoratger:
All six threads are addressed and resolved, each with a reply on the thread. Everything is in `1234f43`.

**The batching-challenge finding was a real soundness bug**, and your probe was right. The challenge was drawn before anything claim-dependent entered the sponge, so `v_0 += γd, v_1 -= d` left the batch unchanged with nothing downstream to catch it.

The claims are now a `Length::Fixed(num_polynomials)` Message step ahead of the challenge, exactly as you suggested, and both drivers play it **in the same call** as the draw — `batching_challenge(&claims)` — so a caller cannot get the order wrong rather than merely being told not to. The count comes with it, which closes the `> 1` gap you noted in the seed.

| thread | outcome |
| --- | --- |
| challenge drawn before the claims | `opening_claims` Message step ahead of it; your probe is now a test |
| docs overclaim about the empty wire | claims are bound here; the opening obligation is written on `SkipOpeningShape` |
| no verifier counterpart | `SkipOpening::verify` owns both equalities, plus a count check |
| `let _ = prover.round_poly()` covers nothing | full prove-and-verify through both drivers, ten tests for one |
| the Lagrange formula is not what the code computes | reworded; `SkipRound::lagrange` builds the vector without the 32 KiB table |
| `num_polynomials = 0` accepted | refused at construction |

Two notes rather than leaving them for you to find:

- On the zeroed `round_poly` mutation: it is covered, but not where I expected. A degree-two round reports two nodes and recovers the third from the running claim, so no round is self-contradicting and the sumcheck alone accepts — the tamper survives to the closing check. The test asserts that error and the comment records why the rounds are a reduction and not a filter.
- Writing your probe as a test taught me something: `EF::from_u64` collapses to the parity bit in characteristic two, so my first `delta` was silently zero and the test passed for the wrong reason. It draws the shift now.

`scripts/check.py lint` exit 0; 454 unit + integration passing.

sumcheck/src/univariate_skip/opening.rs
```diff
@@ -0,0 +1,882 @@
...
+    /// - The point the rounds ended on.
+    /// - The committed evaluations at that point.
+    /// - The value the sumcheck closed on.
+    ///
+    /// # Errors
+    ///
+    /// - The blends and the openings disagree on count.
+    /// - The starting sum is not the batch of the blends.
+    /// - The closing value is not the weighted batch of the openings.
+    pub fn verify(
```
Nashtare: The starting-sum check only bites if `starting_sum` is the proof's own `claimed_sum`, the blends are the ones given to `batching_challenge`, and this runs after `finish()`. A local probe that passes `batch_claims(blends, gamma)` here instead accepts a forged blend with an otherwise honest proof, and returning this error before `finish()` panics on drop. The zerocheck in #2139 ends up not calling this and re-checks both equalities itself, since its openings arrive after the PCS. Would a verifier-driver method that takes the `GenericDegreeProof`, plays claims, challenge, rounds and finish, checks the starting sum, and returns `tau` with the value the batched openings must match, serve both callers?

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -48,14 +47,335 @@ const ROUND_POW: &str = "round_pow";
...
+    /// # Errors
+    ///
+    /// Returns whatever the delegated run rejected with.
+    pub fn sumcheck<T, E>(&mut self, run: impl FnOnce(&mut C) -> Result<T, E>) -> Result<T, E> {
+        self.state
+            .begin_protocol::<OpeningSumcheck>(OPENING_SUMCHECK);
+        let output = run(self.state.challenger_mut());
+
+        if output.is_err() {
+            self.state.abort();
```
Nashtare, line 359: Deleting this `abort()` still passes every test in the module: no test hands these drivers a proof the nested sumcheck rejects, since the round-poly and starting-sum tamper tests both get past it and are caught by `SkipOpening::verify`. Without it, a round polynomial of the wrong width or an invalid pow witness panics on drop instead of returning `Err` (checked locally). A tamper test on `round_polys.pop()` and one on a pow witness would pin it.

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -48,14 +47,335 @@ const ROUND_POW: &str = "round_pow";
...
+///
+/// # What the caller owes
+///
+/// The evaluations the reduction ends on are not bound here.
+/// They are what a commitment answers for.
+/// Discharging them against one is the caller's business.
+///
+/// That is the same obligation the skip round leaves for its zerocheck point.
+/// Two equalities tie those evaluations to this run.
+/// The reduction owns both, so a caller cannot forget one.
```
Nashtare: Nothing in the drivers calls `SkipOpening::verify`, so a caller can still skip both equalities, or make the first one vacuous by passing its own batch as `starting_sum`. Could this section also say that `verify` runs after `finish()` with the proof's `claimed_sum`?

sumcheck/src/univariate_skip/transcript.rs
```diff
@@ -48,14 +47,335 @@ const ROUND_POW: &str = "round_pow";
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SkipOpeningShape {
+    /// Number of skipped variables the reduction binds, one round each.
+    pub num_variables: usize,
+    /// Number of committed polynomials batched into one run.
+    pub num_polynomials: usize,
+    /// Grinding difficulty guarding each challenge, or zero to omit it.
```
Nashtare: Here `pow_bits` only reaches the nested sumcheck's rounds; there is no grinding step before `opening_batching`. Maybe "guarding each sumcheck round", so the security accounting doesn't credit grinding to the batching term.

Nashtare (approved):
Re-reviewed at 66bbfa8. All six threads are addressed, and the batching fix holds:

- The claim-shift forgery is rejected for m = 2, 3, 5, with and without grinding. So is an adaptive variant (forged claims, honest witness proven under the resulting challenge), for m = 1, 2, 3. Seeds for different claim counts no longer collide.
- Malformed proofs all return `Err` without a drop panic: claim and opening counts, round-poly widths, round count, and pow witnesses (missing, smuggled, tampered).
- `SkipRound::lagrange` matches the selector's vector bit for bit at random, transmitted and all byte-field points. The prover's table path is unchanged.
- Each of these mutations fails at least one of the new tests: zeroed round polys, γ = ONE, unbound or zeroed claims, each of the three `verify` checks, and a flipped Horner order.

Two small follow-ups are inline:
- the nested-rejection `abort()` in the verifier driver is not covered by any test;
- `SkipOpening::verify` is only as strong as the `starting_sum` the caller passes.

The PR description could also mention the claims step and `verify`.

sumcheck/src/univariate_skip/opening.rs
tcoratger: All three conditions were real, and your suggested shape is what landed in `5e58f97`.

`SkipOpening::verify` now takes the shape, the `GenericDegreeProof` and the blends, plays claims, challenge, rounds and finish in that order, reads the starting sum out of `proof.claimed_sum`, and checks it afterwards. So the sum is no longer an argument a caller can make agree with itself, and the check cannot land before `finish()`.

It returns a `SkipOpeningClaim` carrying the point, the challenge and **the value the batched openings must match** — the Lagrange weight already divided out. That last part turned up something worth having: dividing means the weight can vanish, and then the old closing check read `0 == 0` and constrained no opening. `closing_target` refuses that with `DegenerateWeight`.

The module's test driver was exactly this code, which is the tell that it belonged in the library. It is now four lines: `verify` then `claim.discharge(&openings)`.

On serving both callers: #2139 cannot share the transcript half, since it plays these steps under its own `ZerocheckShape` rather than `SkipOpeningShape`. It already reads its starting sum from the proof, so it was not exposed to the vacuity. What it can share is the arithmetic, and `closing_target` is the piece — I have left that adoption out of this PR to keep the diff where the defect is.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Confirmed, and fixed in `5e58f97`. Two tests reach it now, one dropping a round and one dropping a grinding witness:

```
a_round_the_nested_sumcheck_refuses_is_an_error_not_a_panic
a_grinding_witness_the_nested_sumcheck_refuses_is_an_error
```

I checked they actually pin it rather than merely passing. With the `abort()` removed both fail, and with exactly the panic you describe:

```
panicked at challenger/src/fs/state/verifier.rs:71:13
```

The reason the old tamper tests could not reach it is the one you name: a zeroed round polynomial and a shifted starting sum both get past the nested replay, because a degree-two round takes its third node from the running claim and so no round is self-contradicting. Dropping a round is a shape rejection, which is the only kind that lands inside the bracket.

The grinding fixture needed the difficulty threaded through, so `prove_at`/`shape_at`/`verify_at` now take it and the old names delegate at zero.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Both gaps are closed by the driver in the thread above, so the section says what is actually true now:

> Nothing else is left to a caller.
>
> The reduction's own verifier drives this whole description.
> It reads the starting sum from the proof and checks it after finishing.
>
> What comes back is a point and the value the batched openings owe.

A caller can no longer skip the equalities, because reaching a `SkipOpeningClaim` means both have run, and can no longer make the first vacuous, because the sum is not an argument.

sumcheck/src/univariate_skip/transcript.rs
tcoratger: Right, and the accounting consequence is the part worth spelling out. Fixed in `5e58f97`:

> Grinding difficulty guarding each sumcheck round, or zero to omit it.
>
> It reaches the delegated rounds only.
> No grinding step stands before the batching challenge, so a soundness budget must not credit grinding to the batching term.

The same correction went onto `ZerocheckShape::pow_bits` and `BinaryZerocheck::pow_bits` in #2139, where the point and the batching challenge are likewise ungrounded.

PR #2136, opened 2026-09-15
feat(security): charge a univariate-skip round in the multilinear budget

tcoratger:
**Stacked on #2135.** Second of the univariate-skip integration chain. Review the parent first; this diff is only `p3-security` and one call site.

`reduction_terms` enumerates every algebraic error source of the multilinear AIR reduction. A univariate-skip round contributed **none**. Its soundness lived in a doc comment on `SkipRound` and was absent from the number the composed report computes — so a prover shipping the skip would have reported a budget that did not describe it.

## What changes

A skip takes its variables out of both the equality point and the rounds, and puts the error somewhere else:

```text
    plain:  a point over m variables, then m rounds
    skip:   a point over m - k variables, one skip round, then m - k rounds
```

So `zerocheck` and `constraint-sumcheck` shrink to the kept width, and two terms appear:

| term | degree | why |
| --- | --- | --- |
| `skip-reconstruction` | `2^(k+e) - 1` | the verifier admits any polynomial of that degree vanishing on the subspace — one more than an honest prover uses |
| `skip-opening` | `2k` | the degree-two rounds that collapse the blend, added in the parent PR |

## The direction of the trade

A test pins it: **a skip costs bits overall.** The reconstruction term is larger than the round terms it removes.

That is worth having in the suite, because the intuition runs the other way — fewer rounds looks like less error. It buys prover time, not soundness, and nothing downstream should read it as free.

## Compatibility

The shape is optional, so a statement declaring no skip produces byte-identical terms to before; a test asserts that directly. The multi-STARK prover spends one round per variable and now says so explicitly rather than inheriting a default — if a skip ever lands there, the accounting has to be revisited deliberately.

Also covered: skipping the whole cube leaves no residual rounds to charge, and the terms disappear rather than going negative.


security/src/multilinear.rs
```diff
@@ -80,14 +103,40 @@ pub fn reduction_terms(
...
+        // below the extension size that vanishes on the subspace.
+        //
+        // That is one degree more than an honest prover uses.
+        //
+        // The challenge must separate a dishonest message from the true one over all of it.
+        let reconstruction = ((1u64 << skip.log_extended) - 1) as f64;
+        add("skip-reconstruction", reconstruction, field_bits);
+
+        // Collapsing the blend takes one degree-two round per skipped variable.
+        add("skip-opening", skipped as f64 * 2.0, field_bits);
```
Nashtare, line 133: The parent's opening reduction also draws γ to batch the committed polynomials (`OPENING_BATCHING`, `SkipOpening::batch_claims`) whenever `num_polynomials > 1`. That is an independent (n−1)/|EF| term no label here covers. Could `MultilinearSkipParams` carry `num_polynomials` and charge `add("skip-opening-batching", (n − 1) as f64, field_bits)`? At n = 1024 on GF(2^128) it costs about 2.5 bits on top of the current total, and once n > 2^(k+e) it outweighs the reconstruction term.

security/src/multilinear.rs
```diff
@@ -80,14 +103,40 @@ pub fn reduction_terms(
...
+    add("zerocheck", kept as f64, nonzero_field_bits);
+
+    if let Some(skip) = air.skip {
+        // The verifier reconstructs the round polynomial itself, admitting any degree
+        // below the extension size that vanishes on the subspace.
+        //
+        // That is one degree more than an honest prover uses.
+        //
+        // The challenge must separate a dishonest message from the true one over all of it.
+        let reconstruction = ((1u64 << skip.log_extended) - 1) as f64;
```
Nashtare, line 129: The shape isn't validated:
- `log_extended <= log_size` (e.g. 0) makes this 0, so the term is silently dropped.
- `log_extended >= 64` panics in debug and drops the term in release.
- `2^(log_extended − log_size) < constraint_degree` under-charges, since the true bound is `d·(2^k − 1)` (127 charged vs 315 for d=5, k=6, e=1).

Could these, and `log_size > num_variables`, return `invalid-multilinear-shape` like the other invalid scalars? Or charge `max(2^ext − 1, d·(2^k − 1))`?

security/src/multilinear.rs
```diff
@@ -80,14 +103,40 @@ pub fn reduction_terms(
...
     // evaluation at tau vanishes with probability at most h/(q-1). The tail
     // inherited from GKR is also rejection-sampled over the nonzero elements.
-    add("zerocheck", air.num_variables as f64, nonzero_field_bits);
+    add("zerocheck", kept as f64, nonzero_field_bits);
+
+    if let Some(skip) = air.skip {
+        // The verifier reconstructs the round polynomial itself, admitting any degree
+        // below the extension size that vanishes on the subspace.
+        //
+        // That is one degree more than an honest prover uses.
```
Nashtare, line 126: "One degree more than an honest prover uses" holds only for d = 2. For d = 3, k = 6 the admitted degree is 255 against an honest 189. Maybe "at least the honest degree d·(2^k − 1)"?

security/src/multilinear.rs
```diff
@@ -31,6 +31,29 @@ pub struct MultilinearAirParams {
...
+    /// Dimension of the subspace the round polynomial vanishes on.
+    ///
+    /// This is how many variables the round binds in one go.
+    pub log_size: usize,
+    /// Dimension of the subspace the round polynomial is transmitted on.
+    ///
+    /// The verifier admits any polynomial of degree below this size.
+    ///
+    /// That is what the reconstruction term charges for.
+    pub log_extended: usize,
```
Nashtare, line 56: Worth stating that this must be the dimension the verifier actually interpolates on (`SkipDomain::log_extended()` / `UnivariateSkipShape::log_extended`), not a value re-derived from the degree.

security/src/multilinear.rs
```diff
@@ -139,6 +189,128 @@ mod tests {
...
+            Some(100.0 - libm::log2(127.0))
+        );
+        assert_eq!(
+            charged(&terms, "skip-opening"),
+            Some(100.0 - libm::log2(12.0))
+        );
+    }
+
+    #[test]
+    fn the_reconstruction_term_dominates_what_the_rounds_gave_up() {
```
Nashtare, line 263: The name says the reconstruction term dominates the rounds given up, but the assertion only compares totals. Either assert `reconstruction_error > removed_round_errors` or rename the test.

Nashtare (approved):
Reviewed at 8de2b6ff, with the soundness of #2135's protocol derived independently.

**What checks out**
- The four charged terms: `zerocheck` (m−k)/(q−1), `skip-reconstruction` (2^(k+e)−1)/q, `constraint-sumcheck` (m−k)(d+1)/q, `skip-opening` 2k/q.
  - A hand computation matches the function to 0.0 bits on BabyBear^4 and GF(2^128), for d ∈ {2,3} and k = 3..8, with `e` taken from `SkipDomain::for_degree`.
  - The reconstruction numerator (2^(k+e)−1) is slightly tighter than #2135's doc and is correct.
- **Skip off:** outputs are byte-identical to the base over 4,800 shapes, and multi-stark levels are unchanged.
- No list-size multiplier is needed here; multi-stark already adjusts these terms.

**Points**
- **γ is not charged (MEDIUM).** The opening reduction's batching challenge γ adds (n−1)/q. On GF(2^128) with d=2, k=6 that is about 2.5 bits at n=1024 and about 8 at n=2^16. Nothing sets `skip` yet, so no reported level is wrong today.
- **Unvalidated skip shape (LOW).** Bad shapes silently under-charge or panic.
- **Comment (LOW).** One comment is only true for d = 2.
- **Tests (LOW).** They restate the formula, so they would not catch the missing term.

security/src/multilinear.rs
Nashtare, line 133: The parent's opening reduction also draws γ to batch the committed polynomials (`OPENING_BATCHING`, `SkipOpening::batch_claims`) whenever `num_polynomials > 1`. That is an independent (n−1)/|EF| term no label here covers. Could `MultilinearSkipParams` carry `num_polynomials` and charge `add("skip-opening-batching", (n − 1) as f64, field_bits)`? At n = 1024 on GF(2^128) it costs about 2.5 bits on top of the current total, and once n > 2^(k+e) it outweighs the reconstruction term.

security/src/multilinear.rs
Nashtare, line 129: The shape isn't validated:
- `log_extended <= log_size` (e.g. 0) makes this 0, so the term is silently dropped.
- `log_extended >= 64` panics in debug and drops the term in release.
- `2^(log_extended − log_size) < constraint_degree` under-charges, since the true bound is `d·(2^k − 1)` (127 charged vs 315 for d=5, k=6, e=1).

Could these, and `log_size > num_variables`, return `invalid-multilinear-shape` like the other invalid scalars? Or charge `max(2^ext − 1, d·(2^k − 1))`?

security/src/multilinear.rs
Nashtare, line 126: "One degree more than an honest prover uses" holds only for d = 2. For d = 3, k = 6 the admitted degree is 255 against an honest 189. Maybe "at least the honest degree d·(2^k − 1)"?

security/src/multilinear.rs
Nashtare, line 56: Worth stating that this must be the dimension the verifier actually interpolates on (`SkipDomain::log_extended()` / `UnivariateSkipShape::log_extended`), not a value re-derived from the degree.

security/src/multilinear.rs
Nashtare, line 263: The name says the reconstruction term dominates the rounds given up, but the assertion only compares totals. Either assert `reconstruction_error > removed_round_errors` or rename the test.

Nashtare (approved):
Reviewed at 8de2b6ff, with the soundness of #2135's protocol derived independently.

**What checks out**
- The four charged terms: `zerocheck` (m−k)/(q−1), `skip-reconstruction` (2^(k+e)−1)/q, `constraint-sumcheck` (m−k)(d+1)/q, `skip-opening` 2k/q.
  - A hand computation matches the function to 0.0 bits on BabyBear^4 and GF(2^128), for d ∈ {2,3} and k = 3..8, with `e` taken from `SkipDomain::for_degree`.
  - The reconstruction numerator (2^(k+e)−1) is slightly tighter than #2135's doc and is correct.
- **Skip off:** outputs are byte-identical to the base over 4,800 shapes, and multi-stark levels are unchanged.
- No list-size multiplier is needed here; multi-stark already adjusts these terms.

**Points**
- **γ is not charged (MEDIUM).** The opening reduction's batching challenge γ adds (n−1)/q. On GF(2^128) with d=2, k=6 that is about 2.5 bits at n=1024 and about 8 at n=2^16. Nothing sets `skip` yet, so no reported level is wrong today.
- **Unvalidated skip shape (LOW).** Bad shapes silently under-charge or panic.
- **Comment (LOW).** One comment is only true for d = 2.
- **Tests (LOW).** They restate the formula, so they would not catch the missing term.

security/src/multilinear.rs
tcoratger: You're right that it is an independent term and that nothing here covered it. Done in a1125be, exactly as suggested: `MultilinearSkipParams` carries `num_polynomials` and the composition charges

```rust
add("skip-opening-batching", skip.num_polynomials.saturating_sub(1) as f64, field_bits);
```

Your arithmetic checks out at both ends: about 2.5 bits at n = 1024 over GF(2^128), and past 2^(k+e) polynomials it does overtake the reconstruction term.

A single-polynomial run draws nothing, so the term is absent rather than zero (`a_single_polynomial_batch_costs_no_separating_draw`), which matches the reduction — its transcript has no batching step at all in that case.

security/src/multilinear.rs
tcoratger: All four accepted. a1125be validates the skip shape alongside the other scalars and returns `invalid-multilinear-shape`, which is the option you offered first:

- `log_extended <= log_size` — refused; you're right that it silently drops the term rather than charging a small one.
- `log_extended >= 63` — refused, with the bound stated as a named constant and the reason (panic in debug, dropped term in release) written on it.
- `log_size > num_variables` — refused.
- `num_polynomials == 0` — refused too, now that the batch width is a parameter.

On the under-charge I took your **second** suggestion as well, since it is not an invalid shape but a wrongly-priced one:

```rust
let admitted = ((1u64 << skip.log_extended) - 1) as f64;
let honest = air.constraint_degree as f64 * ((1u64 << skip.log_size) - 1) as f64;
add("skip-reconstruction", admitted.max(honest), field_bits);
```

Your d=5, k=6, e=1 case is now charged 315 rather than 127, and it is a test (`the_honest_degree_binds_when_the_transmitted_domain_is_narrow`), with the wide case beside it so neither branch is untested. The completeness side of the same fact is checked in #2137, where the round now refuses a constraint its domain cannot carry.

`a_skip_shape_every_term_can_charge_is_accepted` pins the boundary just inside each refusal, so the guard is not over-tight.

security/src/multilinear.rs
tcoratger: Correct — that only holds at d = 2. Reworded in a1125be to state both degrees and say which binds:

```text
    admitted  2^(k + e) - 1        what the interpolation accepts
    honest    d * (2^k - 1)        what a correct prover sends
```

Since the charge is now the max of the two, the doc and the arithmetic say the same thing.

security/src/multilinear.rs
tcoratger: Worth stating, agreed. Added in a1125be:

> This must be the dimension the verifier actually interpolates on.
>
> It is not a value re-derived from the constraint degree.
>
> A prover choosing a wider domain than the verifier reads is charged too little.

The last line is the reason it matters, so it is on the field rather than left implicit.

security/src/multilinear.rs
tcoratger: Fair — the name promised more than the assertion checked. a1125be asserts the direction the name claims:

```rust
let given_up = error(plain zerocheck) + error(plain constraint-sumcheck)
             - error(skip zerocheck)  - error(skip constraint-sumcheck);
assert!(reconstruction > given_up);
```

The comparison is on error rather than bits, since fewer bits is more error. The total comparison stays, now as the consequence rather than as the claim.

security/src/multilinear.rs
tcoratger: Same finding as the thread above on this line — answered there. Done in a1125be: the shape carries `num_polynomials` and the composition charges `skip-opening-batching`.

security/src/multilinear.rs
tcoratger: Same finding as the thread above on this line — answered there. All four shapes are refused, and the reconstruction charge is now the max of admitted and honest degree.

security/src/multilinear.rs
tcoratger: Same finding as the thread above on this line — answered there. Reworded to state both degrees and which binds.

security/src/multilinear.rs
tcoratger: Same finding as the thread above on this line — answered there. Stated on the field, with the reason it matters.

security/src/multilinear.rs
tcoratger: Same finding as the thread above on this line — answered there. The test now asserts the direction its name claims.

PR #2137, opened 2026-09-15
feat(sumcheck): generalise the skip constraint and stream the prover path

tcoratger:
**Stacked on #2136.** Third of the univariate-skip integration chain. Review the parents first; this diff is `p3-sumcheck` only.

Two things the skip round hard-coded, both now the caller's choice.

## The constraint

`a*b + c` was written into the round's tests and into every caller, because the round took *composed values* rather than a constraint. A degree-three constraint was expressible only by composing by hand outside, which is also where the degree that sizes the domain came from.

`Composition` states it once and lets the round read it in whichever carrier it currently holds:

```text
    on the transmitted points  ->  subfield values, from extending bit rows
    in the residual rounds     ->  large-field values, after the challenge
```

The arithmetic is identical and only the carrier changes, so a generic method over any algebra containing the witness alphabet covers both. A closure could not — it would fix one carrier, and the round needs both.

The declared degree sizes the transmitted domain, so the trait says plainly what it owes: **too low leaves the round polynomial unrecoverable from what is sent, and the reduction unsound**; too high only costs cosets.

`Conjunction` is the bit-valued rank-one shape, kept as the one implementation the round was built around. A test walks all eight bit assignments and checks it vanishes exactly on the satisfying ones, and another checks reading it in the small field then embedding agrees with embedding then reading — otherwise the message and the residual rounds would prove different things.

## The prover's memory

Forming the message materialised every operand's extension and then the composed values: `rows × transmitted` subfield elements per operand, all live before anything was weighed. At a six-bit skip that is the whole witness inflated eight-fold, three times over. This was raised in the review of #2134.

Streaming keeps one row's scratch per thread instead:

```text
    per row:  extend each operand  ->  compose  ->  weigh  ->  accumulate
```

The message is the only thing that outlives the row. A test pins the two paths equal on an odd-sized batch, and the materialising one stays as its reference for callers already holding composed values.

## What this does not do yet

It does not assemble the sequence. Owning commit → point → skip → residual → opening, with the point drawn inside so it cannot be chosen early or by the prover, is the next entry in the stack — and it is what finally makes the two obligations documented on `SkipRound` unnecessary to remember.


sumcheck/src/univariate_skip/round.rs
```diff
@@ -194,11 +195,111 @@ impl<F: TowerLevel> SkipRound<F> {
...
+        operands: &[&[u8]],
+        eq: &[EF],
+        composition: &C,
+    ) -> Vec<EF>
+    where
+        EF: ExtensionField<F>,
+        F: Send + Sync,
+        C: Composition<F> + Sync,
+    {
+        let arity = composition.arity();
```
Nashtare, line 236: `composition.degree()` is never read anywhere in the crate. The domain is sized by the separate `degree` passed to `SkipRound::new`. A degree-3 composition streamed on `SkipRound::new(k, 2)` fails the verifier identity for honest witnesses in 12/12 of my probe runs. Could this assert the composition fits the domain, e.g. `SkipDomain::for_degree(k, composition.degree())?.log_extended() <= self.domain.log_extended()`, or should `SkipRound` take its degree from the composition? A degree-3 round-trip test would also cover the generalisation itself; every test currently uses `Conjunction`.

sumcheck/src/univariate_skip/composition.rs
```diff
@@ -0,0 +1,144 @@
...
+///
+/// The declared degree sizes the transmitted domain.
+///
+/// It therefore has to be an upper bound on the total degree in the operands.
+///
+/// Declaring it too low leaves the round polynomial unrecoverable from what is sent.
+///
+/// That makes the reduction unsound.
+///
+/// Declaring it too high only costs cosets.
```
Nashtare, line 30: A degree that is too low makes the round incomplete, not unsound. The verifier imposes vanishing and reconstructs below 2^(k+e), so a broken witness still leaves P − Q ≠ 0 of degree ≤ max(D, 2^(k+e) − 1) and is caught (24/24 in a probe); what breaks is that honest proofs get rejected. And "too high only costs cosets" also grows the reconstruction soundness term that #2136 charges.

sumcheck/src/univariate_skip/composition.rs
```diff
@@ -0,0 +1,144 @@
...
+///     a & b = c        on bits,  a * b = c
+/// ```
+///
+/// Characteristic two makes subtraction addition, so the constraint is read as `a*b + c`.
+///
+/// This is the shape a batch of Boolean circuits reduces to.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct Conjunction;
+
+impl<F> Composition<F> for Conjunction {
```
Nashtare, line 62: `impl<F> Composition<F> for Conjunction` also applies in odd characteristic, where `a*b + c` is not the conjunction (bits (1,1,1) give 2). Either bound it to `F: TowerLevel`, or write `values[0] * values[1] - values[2]`, which is right everywhere.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -573,6 +675,36 @@ mod tests {
...
     }
 
+    #[test]
+    fn streaming_the_message_agrees_with_materialising_it() {
+        // Invariant: the streaming path is only a cheaper route to the same message.
+        //
+        //     materialised: rows * transmitted subfield values per operand, then weighed
+        //     streamed:     one row's scratch per thread, weighed as it goes
+        //
+        // Fixture state: 2^5 rows of 2^6 bits, an odd-sized batch of three operands.
```
Nashtare, line 685: Nit: `1 << 5` rows isn't odd-sized. A non-power-of-two count (e.g. 37) would exercise uneven rayon splits.

sumcheck/src/univariate_skip/round.rs
```diff
@@ -194,11 +195,111 @@ impl<F: TowerLevel> SkipRound<F> {
...
+            assert_eq!(
+                rows.len(),
+                eq.len() * row_bytes,
+                "one packed row per equality weight"
+            );
+        }
+
+        // Rows are independent contributions to the same sum.
+        //
+        // Each thread carries its own message accumulator and its own row scratch.
```
Nashtare, line 251: Nit: rayon's `fold` makes one accumulator per split task rather than per thread. Scratch still doesn't scale with rows × transmitted (about 290 KiB peak at 2^18 rows, against 64 MiB materialised), but the wording is slightly off.

sumcheck/src/univariate_skip/composition.rs
```diff
@@ -0,0 +1,144 @@
...
+                    let value = in_alphabet(&[bit(a), bit(b), bit(c)]);
+                    let satisfied = (a & b) == c;
+                    assert_eq!(value == F::ZERO, satisfied, "{a}{b}{c}");
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn one_constraint_reads_in_both_carriers() {
```
Nashtare, line 121: Nit: reading in F and then embedding always agrees with embedding and then reading for a polynomial with coefficients in F, and the fixture is all bits, so this test can't fail. A random tuple of non-bit subfield values would at least exercise the multiplication.

Nashtare (approved):
Reviewed at 5be8880d. The streaming path is correct and the memory claim holds. The one thing I'd like addressed is that the composition's declared degree is never checked against the domain.

**Checked**
- **Bit-exact equivalence.** The streamed round message matches an independent reference I wrote (per-row interpolation, then compose, then a serial weighted sum). Coverage:
  - k = 3..6;
  - constraint degrees 1-4, plus one with a field constant;
  - BF8/BF128, BF16/BF128 and BF8/BF16;
  - row counts {0,1,2,3,5,7,16,33,100,257};
  - serial and `parallel` builds, on 1/2/3/14 rayon threads.
- **Degree 3 on a correctly sized domain:** honest witnesses verify and broken ones are rejected.
- **Mutations:** a wrong weight, swapped operands and always reading row 0 are all caught. A dropped merge is caught under `parallel`.
- **Memory:** at 2^18 rows, peak scratch is about 290 KiB streamed against 64 MiB materialised.
- **Validation:** clippy is clean on p3-sumcheck and its dependents. p3-sumcheck tests pass in both serial and `parallel` builds.

**Points**
- **Degree never read (MEDIUM).** `Composition::degree` is not read anywhere, so a composition of higher degree than the round's domain silently yields proofs that don't verify.
- **Doc (LOW).** A too-low degree makes the round incomplete, not unsound.
- **Odd characteristic (LOW).** `Conjunction` is implemented there too, where `a*b + c` is not a conjunction.
- **Constraint still hand-written (LOW).** The integration test (`sumcheck/tests/univariate_skip_zerocheck.rs`) and the round.rs test helpers still write `a*b + c` by hand, although the PR says the constraint is stated once.

sumcheck/src/univariate_skip/round.rs
Nashtare, line 236: `composition.degree()` is never read anywhere in the crate. The domain is sized by the separate `degree` passed to `SkipRound::new`. A degree-3 composition streamed on `SkipRound::new(k, 2)` fails the verifier identity for honest witnesses in 12/12 of my probe runs. Could this assert the composition fits the domain, e.g. `SkipDomain::for_degree(k, composition.degree())?.log_extended() <= self.domain.log_extended()`, or should `SkipRound` take its degree from the composition? A degree-3 round-trip test would also cover the generalisation itself; every test currently uses `Conjunction`.

sumcheck/src/univariate_skip/composition.rs
Nashtare, line 30: A degree that is too low makes the round incomplete, not unsound. The verifier imposes vanishing and reconstructs below 2^(k+e), so a broken witness still leaves P − Q ≠ 0 of degree ≤ max(D, 2^(k+e) − 1) and is caught (24/24 in a probe); what breaks is that honest proofs get rejected. And "too high only costs cosets" also grows the reconstruction soundness term that #2136 charges.

sumcheck/src/univariate_skip/composition.rs
Nashtare, line 62: `impl<F> Composition<F> for Conjunction` also applies in odd characteristic, where `a*b + c` is not the conjunction (bits (1,1,1) give 2). Either bound it to `F: TowerLevel`, or write `values[0] * values[1] - values[2]`, which is right everywhere.

sumcheck/src/univariate_skip/round.rs
Nashtare, line 685: Nit: `1 << 5` rows isn't odd-sized. A non-power-of-two count (e.g. 37) would exercise uneven rayon splits.

sumcheck/src/univariate_skip/round.rs
Nashtare, line 251: Nit: rayon's `fold` makes one accumulator per split task rather than per thread. Scratch still doesn't scale with rows × transmitted (about 290 KiB peak at 2^18 rows, against 64 MiB materialised), but the wording is slightly off.

sumcheck/src/univariate_skip/composition.rs
Nashtare, line 121: Nit: reading in F and then embedding always agrees with embedding and then reading for a polynomial with coefficients in F, and the fixture is all bits, so this test can't fail. A random tuple of non-bit subfield values would at least exercise the multiplication.

Nashtare (approved):
Reviewed at 5be8880d. The streaming path is correct and the memory claim holds. The one thing I'd like addressed is that the composition's declared degree is never checked against the domain.

**Checked**
- **Bit-exact equivalence.** The streamed round message matches an independent reference I wrote (per-row interpolation, then compose, then a serial weighted sum). Coverage:
  - k = 3..6;
  - constraint degrees 1-4, plus one with a field constant;
  - BF8/BF128, BF16/BF128 and BF8/BF16;
  - row counts {0,1,2,3,5,7,16,33,100,257};
  - serial and `parallel` builds, on 1/2/3/14 rayon threads.
- **Degree 3 on a correctly sized domain:** honest witnesses verify and broken ones are rejected.
- **Mutations:** a wrong weight, swapped operands and always reading row 0 are all caught. A dropped merge is caught under `parallel`.
- **Memory:** at 2^18 rows, peak scratch is about 290 KiB streamed against 64 MiB materialised.
- **Validation:** clippy is clean on p3-sumcheck and its dependents. p3-sumcheck tests pass in both serial and `parallel` builds.

**Points**
- **Degree never read (MEDIUM).** `Composition::degree` is not read anywhere, so a composition of higher degree than the round's domain silently yields proofs that don't verify.
- **Doc (LOW).** A too-low degree makes the round incomplete, not unsound.
- **Odd characteristic (LOW).** `Conjunction` is implemented there too, where `a*b + c` is not a conjunction.
- **Constraint still hand-written (LOW).** The integration test (`sumcheck/tests/univariate_skip_zerocheck.rs`) and the round.rs test helpers still write `a*b + c` by hand, although the PR says the constraint is stated once.

sumcheck/src/univariate_skip/round.rs
tcoratger: Confirmed and fixed in d748779 — your probe is right, and the declared degree really was dead.

I took the assertion option rather than having the round derive its degree from the composition, because `SkipRound` is also used standalone with an explicit degree (its own tests, and the domain-shape sweep). The check is where you put it:

```rust
assert!(
    self.domain.admits_degree(composition.degree()),
    "the transmitted domain must carry the constraint's degree"
);
```

The dimension arithmetic moved onto `SkipDomain` as `extra_dimensions` and `admits_degree`, and `for_degree` now goes through the same function, so the constructor and the check cannot disagree about what a degree needs.

On the degree-3 round trip: there is now a `SquareProduct` composition (`a^2 * b - c`), and two tests. `a_cubic_constraint_streams_on_a_domain_sized_for_it` streams it on `SkipRound::new(k, 3)` and compares against the materialised reference at 1 and 37 rows; `a_constraint_the_domain_cannot_carry_is_refused` is your failing case, now a rejection instead of a wrong message.

sumcheck/src/univariate_skip/composition.rs
tcoratger: You're right on both counts, and your 24/24 probe is the correct reading of it. Reworded in d748779:

> Declaring it too low costs completeness, not soundness.
>
> The verifier imposes vanishing on the subspace and reconstructs below the size sent.
>
> A broken witness therefore still leaves a nonzero difference, and is still caught.
>
> What breaks is that honest proofs stop fitting in what is sent, and are rejected.
>
> Declaring it too high costs cosets, and widens the domain the verifier reconstructs on, which the budget charges for.

The second half is now also true in the code rather than only in the docs: #2136 charges `max(2^(k+e) - 1, d*(2^k - 1))` for the reconstruction, so a wider domain is genuinely priced.

The same fact is stated once more on `SkipDomain::admits_degree`, since that is where a caller meets it.

sumcheck/src/univariate_skip/composition.rs
tcoratger: Right, and the blanket impl is the problem rather than the formula. Took the second option in d748779:

```rust
// Written as a difference, so the constraint means the same in any characteristic.
values[0] * values[1] - values[2]
```

Bounding it to `F: TowerLevel` would have been narrower than the trait needs to be — nothing else in `Composition` wants a binary field — and a difference costs nothing in characteristic two, where it is still the same addition. The doc says so rather than claiming the sum is the conjunction.

The new cubic composition is written the same way.

sumcheck/src/univariate_skip/round.rs
tcoratger: Fair. d748779 sweeps 1, 32 and 37 rows, so a trailing task shorter than the others is on the path.

The weights had to stop being a tabulated equality polynomial to allow it — 37 is no power of two — which is no loss: the identity under test is that the two paths agree, not that the weights are an `eq` table. The comment says as much.

The cubic round trip sweeps 1 and 37 for the same reason.

sumcheck/src/univariate_skip/round.rs
tcoratger: Correct, and the correction matters for the reason you give — scratch scales with the split count, not with the thread count. Reworded in d748779:

> Each split task carries its own message accumulator and its own row scratch.
>
> Scratch therefore scales with the split count, not with the rows.

The non-streaming path had the same wording and now says "split task" too. Your figures are the useful ones and the claim the wording was reaching for is intact: scratch does not scale with rows times transmitted points.

sumcheck/src/univariate_skip/composition.rs
tcoratger: Right — the fixture made it unfalsifiable. d748779 draws random subfield values instead:

```rust
let small: [F; 3] = core::array::from_fn(|_| rng.random());
```

over 32 tuples, and covers the cubic composition alongside the conjunction. Non-bit values are also what the test should have used from the start: the transmitted points carry *extended* rows, where the product is a real subfield multiplication and not an and. The comment records that, since it is the reason the fixture was wrong rather than merely weak.

sumcheck/src/univariate_skip/round.rs
tcoratger: Same finding as the thread above on this line — answered there. Fixed in d748779: the domain now says whether it carries a degree, the streaming path asserts it, and a cubic composition with a round trip covers the generalisation.

sumcheck/src/univariate_skip/composition.rs
tcoratger: Same finding as the thread above on this line — answered there. Reworded: too low costs completeness, and too high is now genuinely priced by #2136.

sumcheck/src/univariate_skip/composition.rs
tcoratger: Same finding as the thread above on this line — answered there. Written as `values[0] * values[1] - values[2]`.

sumcheck/src/univariate_skip/round.rs
tcoratger: Same finding as the thread above on this line — answered there. The sweep is 1, 32 and 37 rows.

sumcheck/src/univariate_skip/round.rs
tcoratger: Same finding as the thread above on this line — answered there. Reworded to say split task, on both paths.

sumcheck/src/univariate_skip/composition.rs
tcoratger: Same finding as the thread above on this line — answered there. The fixture now draws random subfield values, and covers both compositions.

PR #2139, opened 2026-09-15
feat(sumcheck): own the binary zerocheck from witness to opening point

tcoratger:
Fourth of [stack #2138](https://github.com/Plonky3/Plonky3/pull/2138). **This is where the chain closes.**

`SkipRound` documented two obligations it left to its caller: draw the equality point after committing, and reduce the blend it ends on before asking a commitment to open anything. Documented obligations are the kind a caller forgets, and neither was enforced anywhere.

`BinaryZerocheck` owns the sequence instead:

```text
    draw the zerocheck point   over the kept variables, after the commitment
    skip round                 one message, one challenge, k variables bound
    residual sumcheck          m - k ordinary rounds
    opening reduction          k degree-two rounds
    -> one evaluation point
```

The point is drawn inside, from the transcript, so **no field of the proof carries it and no prover can choose it**. A test pins that: two runs on one witness agree exactly, and absorbing a different commitment first moves the point.

## The blends have to be sent

This was the missing link, and I only found it while wiring the two halves together.

The residual rounds end on `eq(r, ρ) · C(blend_0, …, blend_{n-1})` — **one equation in `n` unknowns.** The individual blends cannot be recovered from it, and the opening reduction needs them as its starting claim. Without them there is no way to get from the residual rounds to the opening reduction at all.

So they cross the wire, bound before the challenge that combines them, and the verifier checks them against the equation it *does* know rather than trusting them:

```text
    residual_final  ==  eq(r, rho) * C(blends)     the blends are checked, not trusted
    opening claim   ==  sum_i gamma^i blends_i     what the reduction starts from
```

Tampering with one blend is rejected with `ResidualClaimMismatch`.

## Tests

The centrepiece runs the chain end to end and then **discharges the claim the way a commitment would** — opening each operand at the claimed point and recombining under the batching challenge. That the point is where the committed operands really take the claimed value is the whole chain closing.

Also covered: a broken constraint; a tampered message; a tampered blend; a message of the wrong width; five malformed delegated proofs that must reject rather than panic; a replay under a different height; grinding end to end; and that the claimed point names every variable.

## Notes

- The transcript field is the challenge field, not the witness alphabet — the witness is bytes and the challenger speaks the 128-bit field.
- The residual prover carries the witness alphabet as a marker, because the rounds read the constraint in the large field while the constraint itself belongs to the alphabet the witness was committed in.


sumcheck/src/univariate_skip/zerocheck.rs
```diff
@@ -0,0 +1,862 @@
...
+            })
+            .map_err(ZerocheckError::Residual)?;
+
+        // The residual rounds end on the weight times the blend constraint.
+        // The verifier knows the weight and the constraint.
+        // The blends the proof carries are therefore checked here, not trusted.
+        if let Err(error) = transcript.operand_blends(&proof.blends) {
+            return Err(error.into());
+        }
+        let eq_at_rho = Poly::new_from_point(zerocheck_point.as_slice(), EF::ONE).eval_base(&rho);
```
Nashtare: This builds the whole 2^(m−k) equality table only to read it at ρ, so verification is linear in the number of rows. Locally, the table alone takes about 470 µs and 256 KiB at 2^14 rows, and 8.7 ms and 4 MiB at 2^18 rows. `Point::eval_eq(zerocheck_point.as_slice(), rho.as_slice())` gives the same value in O(m−k), under 1 µs at both sizes. Swapping it in passes every test, and the transcript is unchanged.

sumcheck/src/univariate_skip/zerocheck.rs
```diff
@@ -0,0 +1,862 @@
...
+
+        // The challenge a run carrying different blends draws.
+        let dishonest = Dishonest {
+            blends: Some(vec![EF::ONE, EF::ONE, EF::ONE]),
+            ..Dishonest::default()
+        };
+        let mut transcript_challenger = challenger();
+        let forged = prove_dishonest(&check, &packed, &mut transcript_challenger, &dishonest);
+
+        assert_ne!(forged.opening.claimed_sum, honest.value);
```
Nashtare: This assertion can't fail. `forged.opening.claimed_sum` is the batch of `[1, 1, 1]`, while `honest.value` is the closing value at the opening point, so they differ whatever γ depends on. Only `honest.gamma != 1` does any work.

I tried two changes, each applied to the pattern and both drivers:
- removing the `operand_blends` step (the verifier keeps its length check)
- moving that step after `opening_batching`

With either change, all 88 tests in `binary_zerocheck` and `univariate_skip` still pass. With the second, a local prover flips one conjunction cell, picks blends after seeing γ that keep both `C(v)` and `Σγ^i v_i`, and gets the claim verified and discharged.

Suggested test: take an honest proof, shift its blends along that same fiber at the honest γ, and assert `OpeningClaimMismatch`. That catches both changes.

sumcheck/src/univariate_skip/zerocheck_transcript.rs
```diff
@@ -0,0 +1,525 @@
...
+
+        // The point comes first, so every later draw depends on the claim.
+        steps.push(Interaction::algebra::<F, EF>(
+            Hierarchy::Atomic,
+            Kind::Challenge,
+            ZEROCHECK_POINT,
+            Length::Fixed(self.log_rows()),
+        ));
+
+        // The whole message is bound before the challenge it will be read at.
```
Nashtare, line 148: No test depends on this binding. If the message step is removed from the pattern and both drivers, with the verifier keeping its length check, every test still passes. `a_tampered_round_message_is_rejected` changes the read-back at λ, so it rejects either way. Without the binding, a local prover with a broken witness draws λ first, then adjusts one message entry so the read-back matches the true residual sum, and gets the claim verified and discharged. A tamper that stays in the kernel of `evaluate(·, λ)` should pin it: the proof must be rejected, and it would pass unchanged exactly when the message is unbound.

sumcheck/tests/binary_zerocheck.rs
```diff
@@ -0,0 +1,294 @@
...
+    verify_and_discharge(&check, &witness, &proof).unwrap();
+}
+
+#[test]
+fn a_broken_constraint_is_rejected() {
+    // Mutation: flip one bit of the claimed conjunction.
+    // The round polynomial stops vanishing on the skipped subspace.
+    // The verifier's reconstruction still does, so the two disagree.
+    let (check, mut witness) = fixture(0xBAD, 0);
+    witness.operands[2][0] ^= 1;
```
Nashtare: Byte 0 is in row 0, and at r = 0 the equality weight covers exactly row 0. So this test, like the rest of the suite, still passes if the zerocheck point is drawn and then replaced by zeros on both sides. Locally, under that change, a witness broken in row 5 is proved by the honest prover and accepted. Flipping a cell in a later row, e.g. `witness.operands[2][5 * check.round().row_bytes()] ^= 1`, makes this test depend on the point.

sumcheck/src/univariate_skip/zerocheck.rs
```diff
@@ -0,0 +1,862 @@
...
+///     opening reduction              k degree-two rounds
+///     -> one evaluation point
+/// ```
+///
+/// The skip round's two obligations are discharged here, not described.
+///
+/// - The point is drawn inside, from the transcript, so no prover chooses it.
+/// - The run ends on a point a commitment opens, not on the round's blend.
+///
+/// # What the caller still owes
```
Nashtare, line 165: Booleanity is missing from this list, though line 146 says "over a bit-valued witness". The verifier proves `C = 0` on whatever field values were committed. With `a, b` uniform in GF(2^128) and `c = a·b`, a prover that forms the message over field values gets the claim verified and discharged (local probe).

On #2141 I said this driver can't take a booleanity constraint; that was too strong. A separate arity-1 `BinaryZerocheck` per operand on `a·a − a` needs no batching challenge and keeps the subfield message. Locally it accepts a bit column and rejects both a column with one cell equal to 2 and the uniform `a` above.

Could this section say the cells must be bits, supplied by a packed commitment or by one such zerocheck per operand?

sumcheck/src/univariate_skip/zerocheck.rs
```diff
@@ -0,0 +1,862 @@
...
+/// The witness has to be committed before this runs.
+/// Its commitment has to be absorbed into the transcript this borrows.
+/// Discharging the closing claim against that commitment is the caller's too.
+#[derive(Debug, Clone)]
+pub struct BinaryZerocheck<F, C> {
+    /// The skip round this opens with.
+    round: SkipRound<F>,
+    /// The constraint being proved vanishing.
+    composition: C,
+    /// Grinding difficulty guarding each challenge, or zero to omit grinding.
```
Nashtare: Nit: the zerocheck point and γ are not ground; only the skip challenge and each sumcheck round are. "guarding the skip challenge and every sumcheck round" would be exact. The same applies to `ZerocheckShape::pow_bits` and to the name `grinding_guards_every_challenge_end_to_end`.

Nashtare (commented):
Reviewed at 0a1bc331. The protocol itself holds up:
- Every challenge is drawn after what it depends on.
- The verifier recomputes ρ‖τ itself.
- The rounds played match the #2136 budget term for term.
- A proof that fails any check returns `Err` without panicking: 19 malformed-proof shapes at pow 0 and 3, plus 0, 2 and 4 openings to `discharge`.

`discharge` with its count check and the split `BlendConstraintMismatch` look good. A local mutation run confirms the mirror tests pin the residual-sum, blend-constraint and opening-sum checks, γ ≠ 1, and the abort paths.

Remaining asks, all inline:
- **Verifier cost.** The verifier builds the full row equality table to read one value, so it is linear in the trace. `Point::eval_eq` is equivalent and O(m−k).
- **Unpinned transcript bindings.** Three are still untested:
  - blends before γ;
  - message before λ;
  - a transcript-drawn zerocheck point.

  Removing or reordering any one of them passes all 88 tests. With the blends moved after γ, a local prover gets a broken witness accepted and discharged. `the_batching_challenge_follows_the_blends` compares two unrelated values, so it cannot see this.
- **Booleanity.** The caller obligations should say the committed cells must be bits. A separate arity-1 zerocheck on `w² − w` per operand works here without a batching challenge, which corrects what I said on #2141.

Smaller, not inline:
- the seed does not bind which composition is used; only the algebraic check tells two compositions of equal arity and degree apart;
- `ZerocheckShape::residual_degree` is never read by `pattern()`;
- the `mod.rs` header still says the zerocheck point and the opening are not discharged here;
- `verify` panics for `log_height <= k`, a shape the budget accepts.

sumcheck/src/univariate_skip/zerocheck.rs
tcoratger: Fixed in `c0e55f6`, exactly as suggested:

```rust
let eq_at_rho = Point::eval_eq(zerocheck_point.as_slice(), rho.as_slice());
```

Verification was linear in the witness for one field element. Your figures are the ones that make the case: 8.7 ms and 4 MiB at 2^18 rows for a value that a product over `m - k` variables gives in under a microsecond with no allocation.

The comment records the trade so the table does not come back:

```text
    table then read   2^(m-k) elements, 4 MiB at 2^18 rows
    product over m-k  one pass, no allocation
```

sumcheck/src/univariate_skip/zerocheck.rs
tcoratger: You are right that the assertion could not fail, and the fiber test is in `c0e55f6`.

`blends_shifted_along_the_constraint_fiber_are_rejected` solves your two equations for a shift with `b = 1`:

```text
    c = v0*b + a*v1 + a*b        from C(v + s) == C(v)
    a + gamma*b + gamma^2*c = 0  from the batch being fixed
```

so `a = b*gamma*(1 + gamma*v0) / (1 + gamma^2*(v1 + b))`. The test asserts both invariants hold before asserting `OpeningClaimMismatch`, so it is visibly about the ordering rather than about arithmetic luck.

I ran both of your changes against it. Moving the `operand_blends` step after `opening_batching` makes it fail — that is the one where your prover got a broken witness verified and discharged. The old assertion passed under the same change, which is the whole problem with it.

The `gamma != 1` half is not lost: with the blends bound first, a forced `gamma = 1` is exactly the case where the shift survives, so the fiber test covers it too.

sumcheck/src/univariate_skip/zerocheck_transcript.rs
tcoratger: Confirmed, and `c0e55f6` pins it the way you describe — a tamper in the kernel of `evaluate(.., lam)`.

The read-back is linear in the message, so the weights come off unit vectors without reaching into the barycentric form:

```rust
let weight = |index| {
    let mut unit = vec![EF::ZERO; honest.message.len()];
    unit[index] = EF::ONE;
    check.round().evaluate(&unit, lambda).unwrap()
};
```

Then adding each weight to the other's entry cancels in characteristic two: `first*second + second*first = 0`. The test asserts the message really moved and that the read-back did not, before asserting rejection.

Checked against your change: with the message step removed from the pattern and both drivers, and the verifier keeping its length check, this test fails and the rest still pass.

The mirror prover now hands back the skip challenge so a test can reach `lambda` at all, which is what made this expressible.

sumcheck/tests/binary_zerocheck.rs
tcoratger: Right, and it is a good catch: the test named the point in its comment while not depending on it. Fixed in `c0e55f6` with your suggestion:

```rust
witness.operands[2][5 * check.round().row_bytes()] ^= 1;
```

I verified it now depends on the drawn point by substituting zeros on both sides, as you did. Under that change `a_broken_constraint_is_rejected` is the one test in the file that fails, and it is the only one — which is also a fair summary of how much the rest of the suite was relying on the point.

sumcheck/src/univariate_skip/zerocheck.rs
tcoratger: Thank you for coming back on this — you are right and I was wrong, and the correction is worth more than the original claim.

My argument was that batching booleanity with the conjunction needs a big-field challenge that cannot multiply inside the subfield carrier. That is true of *batching*, and irrelevant, because an arity-one zerocheck per operand needs no challenge at all. You also correctly point out that even batching is not blocked: `stream_round_message` composes in the subfield and only then multiplies by `eq[row]`, so the challenge rides on the weight.

`c0e55f6` adds booleanity to what the caller owes, with both discharges:

> So is booleanity.
> This proves the constraint vanishes on whatever cells were committed.
> A commitment over a large field holds cells that are not bits.
>
> Two ways to discharge it are open.
>
> - A packed commitment makes a cell an `F_2`-coordinate, so bits come free.
> - One arity-one zerocheck per operand on `v*v - v` settles it otherwise.
>
> The second needs no batching challenge and keeps the subfield message.

The same correction went onto `embed_bits` in #2141, where my "fights the design" line was the sharper version of the same mistake.

sumcheck/src/univariate_skip/zerocheck.rs
tcoratger: Correct on all three. Fixed in `c0e55f6`:

```
Grinding difficulty, or zero to omit it.
It guards the skip challenge and every sumcheck round.

The zerocheck point and the batching challenge carry no grinding step.
```

on `BinaryZerocheck::pow_bits` and `ZerocheckShape::pow_bits`, and the test is now `grinding_guards_the_skip_challenge_and_every_round`. `SkipOpeningShape::pow_bits` in #2135 had the same imprecision and carries the same note, with the accounting consequence spelled out there.

sumcheck/src/univariate_skip/zerocheck.rs
```diff
@@ -0,0 +1,959 @@
...
+        forged.message[1] += first;
+
+        assert_ne!(forged.message, honest.message, "the message really moved");
+        assert_eq!(
+            check.round().evaluate(&forged.message, lambda).unwrap(),
+            check.round().evaluate(&honest.message, lambda).unwrap(),
+            "the read-back at the honest challenge cannot see it"
+        );
+
+        assert!(verify(&check, &forged).is_err());
```
Nashtare: Nit: the sibling tests pin the exact variant, and this one could too. I ran the same kernel tamper locally and got `ZerocheckError::ResidualClaimMismatch`. Asserting that variant would stop an unrelated rejection from satisfying the test.

Nashtare (approved):
Re-checked at c0e55f6a. All six points are addressed, and I confirmed each one with a mutation run. Each of these changes now fails exactly the test meant to catch it:
- removing the blend binding;
- moving it after γ;
- forcing γ = 1;
- unbinding the round message;
- zeroing the zerocheck point.

The fiber algebra is right in characteristic 2, and the test asserts both invariants before it asserts the rejection. The kernel tamper leaves the read-back unchanged and is refused. `eval_eq` matches the table it replaced. fmt, clippy and the `p3-sumcheck` tests pass, serial and parallel.

One optional nit is inline. Also optional: `SkipOpening::closing_target` from the rebased #2135 could replace the local weight check and division without touching the transcript.

sumcheck/src/univariate_skip/zerocheck.rs
tcoratger: Agreed, and your measurement is the right one — `32c80b5` pins it:

```rust
// The binding is what rejects it, so the residual check is what fires.
assert_eq!(
    verify(&check, &forged).unwrap_err(),
    ZerocheckError::ResidualClaimMismatch
);
```

The loose assertion was the weakest one in the file for exactly the reason you give: this tamper is invisible to every check that reads the message back, so a rejection from anywhere else would have looked like a pass while proving nothing about the binding.

#2141 is rebased on top.

PR #2140, opened 2026-09-15
feat(binary-pcs)!: drive the binary-tower PCS through the typed transcript

tcoratger:
The last protocol off the typed layer. After this, no production path in the workspace derives a challenge outside a described transcript.

## The finding: this track was no longer blocked

The migration plan recorded binary tower fields as a separate track, blocked on a `TypeTag::BinaryTower` and a tower codec. Both have since landed — `p3-binary-field` implements `TranscriptField` for all five levels, and that trait requires `Field`, not `PrimeField64`.

I had also written the opposite into two comments in #2129:

```rust
// A binary tower field is not a transcript field.
```

That is false. I confirmed it by calling the layout's typed commitment phase with this crate's real challenger and `MerkleCap` types. Both comments are gone, and the commitment now binds through that phase like every other scheme.

## The description

```text
    per oracle  ->  one Merkle root
    closing     ->  final codeword, grind, query positions
```

The fold rounds are absent: each seeds its own sumcheck sub-transcript, played inside a bracket that lends the sponge. One driver spans the fold batches and the query phase on both sides, so the oracle count, the final codeword width, the grinding difficulty and the query draw all reach the fingerprint.

The prover's `commit` and `fold_rounds_with` stop taking a challenger where they no longer bind, matching the contract `commit_base` and `Layout::commit` already have.

## One addition to `p3-challenger`

Query positions are drawn **distinct**: the loop draws until it has enough, so the number of draws is not known when the pattern is built.

`challenge_uniform_bits_rejecting` is the index analogue of the existing rejecting extension draw. It records the count *kept*, not drawn — both sides know that from configuration, and both consume the same sponge outputs. It carries the same caveat as its sibling, stated in its docs: the constrained and unconstrained draws record one step, and a step type of its own would part both pairs.

Per R6 it lands with its first consumer.

## Removed

`sample_query_indices` and `sample_query_cosets` were a second implementation of the draw the driver now owns — exactly the "two implementations of one primitive that can diverge" hazard. Their property tests moved onto the driver, so there is one sampler and the tests ask it.

## Zero-difficulty convention

The shape elides a zero-bit grinding step, matching this crate's existing canonical-witness check and the rest of the workspace. Two security tests asserted a `Grind(0)` event always appeared; they now model both cases.

## Grinding classification

`p3-binary-pcs` seeds over a tower field, so its separator has a different sponge alphabet and cannot join the workspace sweep. Its site is listed in `UNPRICED_GRINDING_SITES`, and the classification walk that guards that table runs inside this crate. The workspace test skips it explicitly rather than silently.

The guard caught this itself when I first added the row — it failed with *"listed as priced elsewhere, but no pattern describes it"*.

## Verification

| | |
| --- | --- |
| raw production absorbs, whole workspace | **1 → 0** |
| new tests | R4 knob sweep, zero-bit elision, name fit, codeword mutation, grinding classification |

Everything still matching `challenger.observe/sample/grind` is a `#[cfg(test)]` module, or `p3-challenger` / `p3-binary-field` implementing the layer's own primitives.

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt --check` and the repo's `lint --check {scripts,sort,toml,deps}` are all clean.

## Breaking

A transcript-format break for `p3-binary-pcs`, and a signature change to its internal prover entry points.

## What remains after this

Nothing from the migration plan. Two things stay deliberately out, as they always were:

- **The NARG wire format.** Proofs stay serde structs carried by `Carrier::External`. Moving them onto the layer's byte buffer replaces every public proof type in the workspace and is its own campaign — it is where the "no length confusion on the wire" guarantee would come from.
- **Pricing every grinding site in the shared model.** Eleven sites are priced by their own protocol's security report instead. That is a recorded decision with two tests guarding it, not a gap; folding them into `GrindingSites` means deriving their budgets from the papers, not editing a table.


binary-pcs/src/verifier.rs
```diff
@@ -466,12 +441,11 @@ mod tests {
 
     #[test]
     fn sampling_is_transcript_dependent() {
-        let mut a = challenger();
-        let mut b = challenger();
-        assert_eq!(
-            sample_query_indices::<_, BinaryField128>(1 << 10, 4, &mut a),
-            sample_query_indices::<_, BinaryField128>(1 << 10, 4, &mut b),
-        );
+        let config = BinaryPcsConfig::try_new(10, params())
```
Nashtare, line 444: This asserts determinism, not transcript dependence. Making both sides return constant positions (while still drawing) passes the whole suite. Could a test tie positions to the transcript, e.g. different final codewords give different position sets, or positions equal the sorted, deduplicated raw draws?

challenger/src/fs/state/prover.rs
```diff
@@ -581,6 +581,68 @@ impl<C, U: Unit> ProverState<C, U> {
...
+    /// # What the shape does not part
+    ///
+    /// This draw and the unconstrained one record the same step, so no seed parts them.
+    ///
+    /// The extension draws carry the same caveat, and one step type would fix both.
+    ///
+    /// # Panics
+    ///
+    /// Never for a challenger that rejects internally, which is what `RESAMPLE = true` asks for.
+    pub fn challenge_uniform_bits_rejecting<W>(
```
Nashtare, line 616: There's no unit test for this new primitive, and the loop never ends if `accept` cannot admit `count` values (e.g. distinctness with `count > 2^width`). Worth documenting that requirement next to "Panics: Never", and mirroring the extension sibling's test. Same for the verifier twin.

binary-pcs/src/transcript.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// # Errors
+    ///
+    /// - The witness is not zero where the site asks for no work.
+    /// - The witness misses the difficulty the site requires.
+    pub fn query_pow(&mut self, witness: BinaryField128) -> Result<(), TranscriptFailure> {
+        if self.shape.pow_bits == 0 {
+            // A zero-difficulty site plays no step, so nothing else reads this field.
+            //
+            // Pinning it here is what stops any value from riding along unbound.
+            if witness != BinaryField128::default() {
```
Nashtare, line 361: `final_codeword` aborts the driver before returning `Err`, but this path doesn't. A public caller that uses `?` and drops the transcript would panic in `Drop`. Abort here too? (It's unreachable from `verify_opening` today because of `check_canonical_pow_witness`.)

examples/tests/transcript_separation.rs
```diff
@@ -1174,6 +1184,15 @@ fn every_unpriced_grinding_site_is_described_by_the_protocol_that_owns_it() {
         .collect();
 
     for &(protocol, label) in &UNPRICED_GRINDING_SITES {
+        // One protocol seeds over a binary tower field.
+        //
+        // Its separator has a different sponge alphabet, so it cannot join the sweep above.
+        //
+        // Its own crate asserts that its site is classified.
+        if protocol == "p3-binary-pcs" {
```
Nashtare, line 1192: Skipping here drops the row → pattern direction of the check for `p3-binary-pcs`; the in-crate test only walks pattern → row. So removing the grind step would leave the `grinding.rs` row silently stale. Could the crate also assert that every row for this protocol is described?

binary-pcs/src/transcript.rs
```diff
@@ -0,0 +1,621 @@
...
+    ///
+    /// When the base domain does not hold a whole number of fold pairs.
+    #[must_use]
+    pub const fn new(config: &BinaryPcsConfig) -> Self {
+        // Queries index pairs, so the sampled domain is one bit narrower.
+        let shift = config.log_folding_factor() - 1;
+        let pair_domain = config.domain_size() >> shift;
+
+        Self {
+            // Every batch but the last commits its folded oracle.
```
Nashtare, line 77: The PCS seed alone doesn't separate `num_variables` or `log_folding_factor`: e.g. every `(nv, rate, k = nv)` shares one fingerprint (30 colliding groups for nv ≤ 12). The layout batching seed covers nv, so this is fine, but adding both to the shape would make the separator self-contained and match the every-knob-in-the-seed convention.

Nashtare (approved):
Reviewed at 6043a93c. Compared with base, the transcript absorbs the same data in the same order.

**Intended differences**
- **New PCS seed:** sits between recording the claims and drawing the batching challenge, on both sides.
- **Commitment:** now goes through the layout's typed commitment phase.
- **Zero-bit grind:** is no longer a step. That leaves the wire unchanged, since `grind(0)` never absorbed anything.
- **Query draw:** yields the same position set as the old distinct sampler.

Every prover message is absorbed before the challenge that depends on it, grinding witnesses are bound, and prover and verifier follow the same pattern.

**Checked**
- **Forgery probes:** 768 mutated proofs across nv {4,6}, pow {0,4}, k {1,2,3} and both opening modes, touching every proof field plus bad lengths. No panics. The only two accepted proofs reused another proof's 4-bit pow witness in configs that open every position, which is grinding-witness malleability, not a forgery.
- **Mutations caught by the tests:**
  - the final codeword not absorbed;
  - only the first oracle bound;
  - constant query positions on the verifier side only;
  - `num_pairs`/`pow` missing from the seed;
  - a missing absorb on the prover side.
- **Separation:** the protocol name is distinct from every other in the workspace.
- **Validation:** fmt and clippy (`-D warnings --all-targets`) are clean on binary-pcs, challenger, security, multi-stark and examples, and every test passes.

**Comments (all LOW, inline)**
- **Query positions untied to the transcript:** no test catches it. Constant positions on *both* sides, with the draw still made, pass all 91 tests.
- **`challenge_uniform_bits_rejecting`:** it has no unit test, and it never returns if `accept` can't admit `count` values.
- **Missing `abort()`:** the zero-bit witness error doesn't abort the driver, unlike `final_codeword`.
- **Grinding classification:** one direction of the check is skipped for `p3-binary-pcs`.
- **PCS seed:** on its own it doesn't pin `num_variables` or the folding factor. The layout seed covers nv, so this isn't exploitable.

**Nits outside the diff**
- `binary-pcs/src/test_util.rs:108` still carries "A binary tower field is not a transcript field", which the description says was removed.
- `binary-pcs/src/verifier.rs:156` still says "before `challenger` is touched", but the parameter is now `transcript`.

binary-pcs/src/transcript.rs
tcoratger: Added both, so the separator stands on its own.

`num_variables` and `log_folding_factor` move no step, so they are bound as instance chunks rather than riding the fingerprint. The exhaustiveness destructuring in `every_knob_of_a_run_reaches_the_seed` failed to compile until both were swept, which is what it is there for.

Your collision is in the docs as the reason:

```text
    nv = 8, k = 8   ->  one batch, no oracle, rate-wide final word
    nv = 9, k = 9   ->  the same three numbers
```

The doc also says plainly that the layout seed already covers the arity, so this makes the seed self-contained rather than closing a reachable gap. I did not want a later reader to think it had been exploitable.

examples/tests/transcript_separation.rs
tcoratger: Right — the skip dropped one direction, and the in-crate test only had the other.

The in-crate test now runs both, and is renamed for it:

- pattern → row: every described step is classified by one of the two vocabularies.
- row → pattern: every `UNPRICED_GRINDING_SITES` row naming this protocol is a step the run describes.

So dropping the grind fails the second, and adding an unclassified one fails the first. The workspace skip now says the crate runs both directions rather than just that the site is classified.

binary-pcs/src/transcript.rs
tcoratger: Fixed — it aborts now, same as the codeword path.

You are right that it is unreachable through `verify_opening`, since `check_canonical_pow_witness` runs first. But the driver is public, and a caller using `?` and dropping the transcript would panic in `Drop` on top of an error already travelling out. That is exactly what `abort` is for.

The grind-failure path below needs none: a failed `observe_pow` poisons the driver itself, which is how the WHIR and FRI drivers handle the same pair.

challenger/src/fs/state/prover.rs
tcoratger: Both taken.

Three tests, mirroring the extension sibling:

- `a_rejecting_index_draw_keeps_only_what_the_predicate_admits` — an even-only predicate over three bits.
- `a_distinctness_predicate_draws_until_the_kept_indices_differ` — the query schedule's own case.
- `an_index_predicate_moves_the_stream_but_not_the_shape` — the pair finalises against one description, which is the caveat both methods carry.

On termination, the docs now say it rather than leaving it implied:

> The loop runs until `count` candidates are kept.
>
> So `accept` must be able to admit that many.
>
> A distinctness predicate needs `count <= 2^width`, which is a caller obligation.

The verifier twin points at the prover's docs for it, as it does for the rest.

Worth noting the binary PCS cannot reach the bad case: `num_pairs` is `min(num_queries, pair_domain / 2)`, so the count is capped at the domain by construction.

binary-pcs/src/verifier.rs
tcoratger: You are right, it asserted determinism and nothing more.

It now varies the last thing bound before the draw and asserts the positions move with it:

```text
    final codeword all-zero  ->  one position set
    final codeword all-one   ->  another
```

It first asserts `num_pairs < 1 << pair_bits`, since a saturated draw opens every position whatever the transcript says, and would make the comparison vacuous again for a different reason.

Verified against your mutation: making both sides return `(0..num_pairs)` while still drawing now fails this test, plus the three property tests that share the helper.

PR #2141, opened 2026-09-15
feat(sumcheck): close the binary zerocheck against a real commitment

tcoratger:
The zerocheck driver in #2139 ends on a claim: *operand `i` takes value `v_i` at point `r`*. Nothing in the stack so far discharges it — the driver's own test reads the operands back out of the witness it just proved over, and says so in a comment.

This entry closes that, finds out what closing it is actually worth, and lands the algebra that makes it worth what it should be. Three commits, in that order.

*(#2142 was a follow-up PR for the third commit; per review it belongs here, and is folded in.)*

---

## 1. Closing the chain

**`embed_bits`** reads a packed bit witness as the multilinear the claim is about, in the cell order the skip round reads rows in — pinned by its own test, since a claim about this polynomial has to be about the same cells the constraint was checked on.

**The integration test** runs the whole chain through public surfaces only:

```text
    commit(3 operands as one table)
      -> BinaryZerocheck::prove
      -> claim at a point
      -> PrescribedPointPcs::open_at / verify_at
      -> ZerocheckClaim::discharge
```

The verifying side holds the commitment and the two proofs, nothing else, and binds the commitment through the scheme's own `observe_commitment` — the method the commit side calls. Prover and verifier run on two independently constructed challengers, per the convention the PCS's own end-to-end test sets.

The final comparison is **not** in the test. `ZerocheckClaim::discharge` (added in #2139) owns it, because it is the check that makes the chain sound: its orientation has to match the batching the reduction proved over and its order the operand order, and a caller that leaves it out has verified a zerocheck over no particular witness. The negative test therefore asserts a typed rejection rather than comparing two field elements itself.

| test | what breaks | who catches it |
| --- | --- | --- |
| `a_broken_constraint_is_refused_before_any_opening` | one bit of `a & b` flipped | the zerocheck, as `BlendConstraintMismatch` |
| `a_claim_about_another_witness_does_not_open_against_this_commitment` | commit witness A, prove over witness B — **both satisfying** | only the discharge |

The second is the one that matters here. Both witnesses satisfy the conjunction, so `verify` and `verify_at` both return `Ok`; the discharge is the whole rejection.

## 2. What closing it is worth — and what it is not

Review found the cost note was understating the problem, and was right. **A cell of this commitment is a whole field element, and nothing in the proof constrains one to a bit.** The constraint cannot tell the difference:

```text
    a * b - c = 0        holds for any field a, b with c = a*b
```

So a witness with `a, b` uniform in GF(2^128) and `c = a·b` satisfies it on every cell and is accepted end to end. What a verifier learns is a statement about *field* cells.

Booleanity here comes only from the prover's interface, which takes packed bytes and can represent nothing else — and an interface is not a proof: a prover forming the round message itself, over field values, is not held to it.

That is now what the documentation says, in place of a cost note that read as the only thing wrong, and the cause is pinned as a test (`the_constraint_cannot_tell_a_bit_from_a_field_element`). The test that claimed every cell is a bit now says whose property that is: the embedding's, and no proof carries it, because a verifier never runs the embedding.

**This changes what the packed commitment is for.** It is usually described as a size win, and it is one — 128× — but it is also what makes the statement a bit statement: a witness cell there is an `F_2`-coordinate of a committed element, so booleanity holds *by construction* rather than by a constraint anyone has to add. It stops being something to prove.

So this path is a scaffold for the chain's mechanics, correct as such, and named as such.

## 3. The algebra the packed path needs

`ring_switch` is generic over `F: Field, EF: ExtensionField<F>` and cannot be instantiated at `F = F_2`. Two reasons:

**The accessor.** `as_basis_coefficients_slice(&self) -> &[F]` hands out a *borrowed* slice, with every impl asserting `size_of::<Upper>() == DIMENSION * size_of::<Lower>()`. For single bits that is 128 one-byte values borrowed out of a 16-byte element. `BinaryField128: ExtensionField<Gf2>` does not and cannot exist.

**The representation.** Even if it did, `TensorAlgebra { coeffs: Vec<F> }` at `F = Gf2` is 16384 one-byte values — **16 KB in the transcript**, where the same matrix bit-packed is 2 KB. Generalising the existing type over an accessor trait was therefore not the answer either: the bit case wants a different representation, and coherence would have fought a blanket impl besides.

Hence `ring_switch::bits` — four types, not a set of functions over loose slices:

| type | what it is | what it replaces |
| --- | --- | --- |
| `Coefficients<EF>` | the `F_2`-coordinates of one element | six free functions doing byte and bit arithmetic |
| `BitPacking<EF>` | a bit witness read as the packed multilinear | a `pack` function plus two dimension constants |
| `BitTensor<EF>` | an element of `EF ⊗_{F_2} EF`, as a bit matrix | — |
| `BitRingSwitch<EF>` | one reduction, and the five values it produces | `batched_weights`, `batch_rows`, `equality_element` |

Three things that buys beyond readability:

- **The bit-valued precondition becomes type-level.** The reduction's soundness rests on the committed polynomial being a packing of bits — which, per §2, is exactly what the current path cannot establish. `BitPacking` carries it.
- **The equality tables are built once.** Every value is a function of the same two points, and `BitRingSwitch` holds them; the free-function version rebuilt `eq(r_batch)` inside *both* `batched_weights` and `batch_rows`, on every call.
- **The widths are checked once, and reported.** A short batching challenge would zip-truncate and drop rows from *both* sides of the sumcheck's identity in the same direction — a failure no later check catches — so it is refused at construction.

The five outputs are named for their role, not their arithmetic: `tensor`, `incoming_claim`, `weights`, `initial_sum`, `closing_weight`.

**At a bit alphabet** a coefficient is one bit, so every product by a coefficient is a conditional add (`d` multiplications per hypercube point become a subset sum), and the tensor element is one field element per row. The set-bit walk lives in one place and both accumulations are written around it.

**The basis** is the one the byte representation already defines, reached through `into_bytes`/`from_le_byte_iter`, which are inverse; `beta_0 = 1` follows. The payoff is that packing moves no bits — it reads the same bytes back as a wider integer.

**The identities are pinned**, not inferred: the incoming claim is the witness at its own point, the columns are the bit planes at the kept coordinates, the batched rows are the sum the sumcheck proves, the closing weight is the weight multilinear at the point the rounds end on, and the equality recurrence matches its exponential definition at every width up to 5 variables. Tests run at `BinaryField16`, where a 16 × 16 matrix is checkable entry by entry, and the same code serves the 128-bit level.

## Still to come

The reduction driver and its transcript, which is what turns this algebra into the packed commitment — and with it both the 128× and the booleanity gap §2 describes.

## Verification

- `cargo test -p p3-sumcheck -p p3-binary-pcs --features parallel` — 490 + 27 unit, 27 + 4 + 10 + 6 + 11 integration, all passing
- `cargo test -p p3-sumcheck` — passing without the feature; the accumulation goes through `par_fold_reduce`, so the serial fallback takes the same path
- `scripts/check.py lint` — exit 0


sumcheck/src/univariate_skip/committed.rs
```diff
@@ -0,0 +1,168 @@
...
+/// ```text
+///     packed bits              one bit per cell, eight to the byte
+///     embedded multilinear     one field element per cell
+/// ```
+///
+/// # Cost
+///
+/// This is the unoptimised path, and the cost is not small.
+///
+/// One 128-bit element per bit is 128 times the commitment the witness needs.
```
Nashtare: The commitment is one `BinaryField128` per cell, and nothing downstream checks that a cell is a bit. A committed witness with `a, b` uniform in GF(2^128) and `c = a·b` satisfies `a·b + c = 0` on every cell. I ran it through `BinaryPcs::commit`, a prover playing `BinaryZerocheck`'s transcript over field values, `open_at`/`verify_at` and the recombination, and it's accepted end to end (local probe). So this path is not correct at 128× cost; it's unsound for bit semantics. The packed commitment is what gives booleanity, not only a size win, and `BinaryZerocheck` can't take a booleanity constraint soundly (one composition, no batching challenge). Could this say so, or gain a booleanity argument, before it's described as closing the chain?

sumcheck/src/univariate_skip/committed.rs
```diff
@@ -0,0 +1,168 @@
...
+        assert_eq!(embedded.as_slice()[15], EF::ONE);
+        assert!(
+            embedded.as_slice()[..15]
+                .iter()
+                .all(|&value| value == EF::ZERO)
+        );
+    }
+
+    #[test]
+    fn every_cell_is_a_bit() {
```
Nashtare: `every_cell_is_a_bit` pins the honest embedding only. The verifier never sees `embed_bits`, and a prover can commit any field values, so the premise this names is assumed, not enforced.

binary-pcs/tests/committed_zerocheck.rs
```diff
@@ -0,0 +1,280 @@
...
+        opened: opened[0].current().to_vec(),
+        gamma: claim.gamma,
+        value: claim.value,
+    })
+}
+
+/// Fold the opened columns under the zerocheck's own batching challenge.
+///
+///     sum_i gamma^i * value_i
+fn recombine(closing: &Closing) -> F {
```
Nashtare: This fold is the check that makes the chain sound, and it exists only in the test. Its orientation must match `SkipOpening::batch_claims` and the column order must match the operand order. In `a_claim_about_another_witness…` both `verify` and `verify_at` return `Ok`; only this helper plus `assert_ne!` rejects. Worth a `ZerocheckClaim::discharge(&[EF]) -> Result<(), ZerocheckError>` (or a binary-pcs helper owning open/verify/recombine) so an integrator can't skip it?

binary-pcs/tests/committed_zerocheck.rs
```diff
@@ -0,0 +1,280 @@
...
+    // Only the commitment separates them, which is what closing the chain buys.
+    let check = BinaryZerocheck::<A, _>::new(LOG_SKIP, Conjunction, 0).unwrap();
+    let num_rows = 1 << (LOG_HEIGHT - LOG_SKIP);
+    let row_bytes = check.round().row_bytes();
+    let committed = witness(0x1, num_rows, row_bytes);
+    let proven = witness(0x2, num_rows, row_bytes);
+
+    let closing = run(&committed, &proven, &check).unwrap();
+
+    assert_ne!(recombine(&closing), closing.value);
```
Nashtare: `run` returned `Ok` here: the library accepted. A library-level discharge would let this assert `Err(..)` instead of comparing two field elements in the test.

binary-pcs/tests/committed_zerocheck.rs
```diff
@@ -0,0 +1,280 @@
...
+    let check = BinaryZerocheck::<A, _>::new(LOG_SKIP, Conjunction, 4).unwrap();
+    let num_rows = 1 << (LOG_HEIGHT - LOG_SKIP);
+    let operands = witness(0x6D1, num_rows, check.round().row_bytes());
+
+    let closing = run(&operands, &operands, &check).unwrap();
+
+    assert_eq!(recombine(&closing), closing.value);
+}
+
+#[test]
```
Nashtare, line 248: Both negative tests go through the honest prover. Deleting `zerocheck.rs:350` or `:382`, removing the `operand_blends` binding, or forcing γ = 1 each pass every test in this file and in `binary_zerocheck.rs`. Under a prover that writes its own claimed sums or blends, each of those becomes an end-to-end acceptance of a broken witness (local probe). An adversarial-prover test here would pin all four.

binary-pcs/tests/committed_zerocheck.rs
```diff
@@ -0,0 +1,280 @@
...
+            prover_data,
+            &protocol,
+            core::slice::from_ref(&prover_claim.point),
+            &mut prover_challenger,
+        )
+        .unwrap();
+
+    // Verifier: replay from a fresh transcript, holding only the commitment and the proofs.
+    let mut verifier_challenger = challenger();
+    verifier_challenger.observe(commitment.clone());
```
Nashtare: Nit: `scheme.observe_commitment(&commitment, &mut verifier_challenger)` rather than a raw `observe`, so the test binds through the same method `commit` uses.

sumcheck/src/univariate_skip/committed.rs
```diff
@@ -0,0 +1,168 @@
...
+///
+/// For single bits that would mean borrowing 128 bytes out of a 16-byte value.
+///
+/// Closing that gap needs a packing built on the tower's own bit representation.
+///
+/// That is a piece of work in its own right.
+///
+/// # Panics
+///
+/// Panics if the packed length is not a whole number of bytes' worth of cells.
```
Nashtare: Nit: bytes are always whole; the condition is that `8·len` is a power of two.

Nashtare (changes requested):
Reviewed at 5f06ed21, with the composed protocol (#2139's driver and this chain) traced end to end.

**What holds**
- **Claims bound before γ.** #2139's blends step (`zerocheck_transcript.rs:190`) absorbs the claims before γ, which closes the post-γ claim shift from #2135. A local end-to-end forgery shifts the blends after seeing γ, keeping both `C(v)` and `Σγ^i v_i`. It is rejected with `OpeningClaimMismatch`, and accepted once that step is removed.
- **Commitment first.** The commitment is absorbed before the zerocheck point, and a different commitment moves the point.
- **Shape in the seed.** Rows, message width, pow bits, arity and whether γ is drawn are all seeded, and grinding witnesses are bound.
- **Opening point.** The PCS opens at the verifier's own (ρ‖τ), and `verify_at` rejects a wrong point or another commitment.
- **Degree.** `BinaryZerocheck::new` sizes the domain from `composition.degree()`.
- **Malformed proofs.** All 12 cases return `Err` and none panic.
- **Validation.** nightly fmt, and clippy `-D warnings --all-targets` (default and `--all-features`), are clean. `committed_zerocheck`, `binary_zerocheck`, p3-sumcheck and p3-binary-pcs pass, with and without `parallel`.

**Blocking: the witness is never forced to be bits (HIGH).** The commitment stores one GF(2^128) element per cell, and nothing checks that a cell is 0 or 1. A committed witness with `a, b` uniform in GF(2^128) and `c = a·b` satisfies `a·b + c = 0` everywhere. A local probe runs it through `BinaryPcs::commit`, a prover playing the driver's transcript over field values, `open_at`/`verify_at` and the recombination: it is accepted end to end, with and without grinding. So the chain doesn't close at 128× cost; it is unsound for bit semantics.
- **No workaround:** `BinaryZerocheck` has one composition and no batching challenge, so a caller can't soundly add a booleanity constraint.
- **Why the next step matters:** the packed commitment that comes next is what supplies booleanity, so it is needed for soundness, not only for size.

**Also**
- **The last check lives only in the test (MEDIUM).** `Σγ^i opened_i == claim.value` is done only in the test helper.
- **Negative tests use the honest prover (MEDIUM, inherited from #2139).** Four mutations each pass `committed_zerocheck.rs` and `binary_zerocheck.rs`: deleting the check at `zerocheck.rs:350`, deleting the one at `:382`, removing the `operand_blends` binding, or forcing γ = 1. An adversarial prover turns each into an accepted broken witness.
- **The budget still can't describe this protocol (LOW).** Nothing passes real skip parameters to the multilinear budget, and the (arity−1)/|EF| blend-batching term is still uncharged (#2136).

**On #2139's lines (outside this diff)**
- `zerocheck.rs:107-113`: "What the caller still owes" lists commit and discharge, but not booleanity, while the struct doc says "over a bit-valued witness". The verifier proves `C = 0` on whatever field values were committed.
- `zerocheck.rs:350` and `:382`: deleting either check alone passes the whole suite. Without `:350` a prover sets its own residual claimed sum; without `:382` it fits its blends to the residual final.
- `zerocheck_transcript.rs:190`: this step is exactly what closes the post-γ shift, and no test would notice it disappearing.
- `zerocheck_transcript.rs:78`: the shape binds arity and degree but not which composition, so two constraints of the same arity and degree seed identically. Worth a composition tag.

sumcheck/src/univariate_skip/committed.rs
tcoratger: You're right, and this is the most useful finding on the stack. I took the first option — say so — because the second turns out to be the wrong shape for this design, and the reason is worth writing down.

**Why not a booleanity constraint here.** Batching `a*b - c` with `w² - w` needs a challenge, and the skip round evaluates its composition over *subfield* values to keep the message cheap. A challenge in `EF` cannot multiply values inside that carrier, and one drawn in the alphabet is 8 bits of soundness at `BinaryField8`. So adding booleanity to this path means giving up the subfield message — fighting the design rather than fixing it.

**Why the packed commitment is the real answer, as you say.** There a witness cell *is* an `F_2`-coordinate of a committed element, so a non-bit cell cannot be expressed at all. Booleanity is not a constraint to add, it is a property of the encoding. That reframes the whole thing: the packing is not a performance follow-up, it is what makes the statement a bit statement, and the cost note read as though 128× were the only thing wrong.

So in 77c7cbc the docs say what this path proves and what it does not, including the part I think is the sharpest version of your point:

> Booleanity is supplied here only by the prover's own interface.
>
> It takes packed bytes, and can therefore represent nothing else.
>
> A prover that forms the round message itself, over field values, is not held to it.

The cause is pinned as a test rather than left as prose — `the_constraint_cannot_tell_a_bit_from_a_field_element` draws `a, b` uniform over `GF(2^128)`, sets `c = a·b`, asserts neither is Boolean, and asserts the constraint vanishes anyway.

I did not reproduce your full end-to-end acceptance as a test, and it is worth saying why rather than quietly not doing it: the prover interface takes packed bytes, so exhibiting it needs a second prover that forms the message over field values — a reimplementation of the message formation, for a fact the constraint already settles. I took your probe as established and pinned its cause. If you would rather have the end-to-end version in the tree, say so and I will add it.

The PR description and the module header are reframed to match: this is a scaffold for the chain's mechanics, named as such.

sumcheck/src/univariate_skip/committed.rs
tcoratger: Correct, and the test's name was the misleading part. Renamed and rewritten in 77c7cbc:

```rust
fn this_embedding_puts_only_bits_in_the_cells() {
    // Every cell of what this function returns is zero or one.
    //
    // That is a property of this function and of nothing else.
    //
    // A verifier never runs it, so no proof carries the fact.
    //
    // The test beside this one is the one that matters for soundness.
```

The test beside it is the one from the thread above. Naming which of the two carries soundness is what stops the next reader drawing the conclusion the old name invited.

binary-pcs/tests/committed_zerocheck.rs
tcoratger: Agreed on all three counts, and `ZerocheckClaim::discharge` is in 6ec388d (on #2139, where the claim lives):

```rust
pub fn discharge(&self, openings: &[EF]) -> Result<(), ZerocheckError>
```

It owns the orientation and the order, and refuses a batch of the wrong width — the claim now carries the operand count for that. The test calls it instead of folding by hand, and so does the driver's own end-to-end test on #2139.

The sentence you make me write on it is the one that justifies the API existing at all:

> Nothing before this ties the proof to the commitment.
>
> A caller that skips it has verified a zerocheck over no particular witness.

I put it on the claim rather than in `binary-pcs` because the batching is the zerocheck's, and a helper owning open/verify/recombine would have to live in whichever crate holds the scheme — leaving the same trap for any other scheme.

binary-pcs/tests/committed_zerocheck.rs
tcoratger: Done in 77c7cbc — it asserts the rejection now:

```rust
assert_eq!(
    closing.discharge().unwrap_err(),
    ZerocheckError::OpeningsDoNotMatchClaim
);
```

and the comment records the thing this test is really about, which your note made explicit: both `verify` and `verify_at` return `Ok` here, so the discharge is the whole rejection.

binary-pcs/tests/committed_zerocheck.rs
tcoratger: Checked all four of your mutations, and you were right about every one — including one I got wrong on the first attempt. Fixed in 6ec388d (#2139).

The negative tests no longer go through the honest prover. A mirror of the prover substitutes one value at a time — the residual sum, the blends, the opening sum — during proving rather than after, so the transcript stays self-consistent because the rounds really run on the substituted value. Editing a finished proof only proves the transcript binds; substituting isolates one check.

A separate test asserts the mirror agrees with the real prover when nothing is substituted, so a rejection can never be drift between the mirror and the code it mirrors.

**The part I got wrong.** My first version of the residual test passed with `zerocheck.rs:350` deleted — exactly your point. The cause is that the two residual checks shared `ResidualClaimMismatch`, so neither was distinguishable and either was deletable with no test noticing. They are different statements, so the second is now `BlendConstraintMismatch`.

Verified by deleting each in turn:

| mutation | test that now fails |
| --- | --- |
| residual sum check removed | `a_residual_sum_the_message_does_not_give_is_rejected` |
| blend constraint check removed | `blends_that_do_not_satisfy_the_constraint_are_rejected` |
| opening sum check removed | `an_opening_sum_the_blends_do_not_give_is_rejected` |
| `operand_blends` binding removed | three tests, including the mirror-agreement one |

γ = 1 is covered by `the_batching_challenge_follows_the_blends`, which needs its own test rather than more end-to-end coverage: forcing it on both sides leaves an honest run verifying.

binary-pcs/tests/committed_zerocheck.rs
tcoratger: Fixed in 77c7cbc: `scheme.observe_commitment(&commitment, &mut verifier_challenger)`, with a comment noting it is the method the commit side called. Binding through the raw sponge would have kept passing even if the scheme changed how it binds.

sumcheck/src/univariate_skip/committed.rs
tcoratger: Right. Fixed in 77c7cbc — the assertion and its message are on the cell count now:

```rust
assert!(
    (packed.len() * CHUNK_BITS).is_power_of_two(),
    "the cell count must be a power of two"
);
```

The `# Panics` section and the test's expected message follow it.

tcoratger:
All seven threads are addressed and resolved, each with a reply on the thread itself. Summary, since the shape of the PR changed as a result:

**The booleanity finding was the load-bearing one**, and it changed the framing rather than just the docs. One element per bit is not only 128× too large — a commitment cell is a whole field element and nothing in the proof constrains one to a bit, so a witness with `a, b` uniform in `GF(2^128)` and `c = a·b` satisfies the constraint on every cell.

I took the "say so" option rather than adding a booleanity argument, and the reason is on the thread: batching `a*b - c` with `w² - w` needs a challenge, and the skip round evaluates its composition over *subfield* values to keep the message cheap. An `EF` challenge cannot multiply inside that carrier, and an alphabet-field one is 8 bits of soundness at `BinaryField8`. So booleanity here fights the design; the packed commitment is the answer, because a witness cell there *is* an `F_2`-coordinate.

Which is why `ring_switch::bits` is now in this PR (`24edd68`) rather than in #2142 — the packing is not a performance follow-up, it is what makes the statement a bit statement, so it belongs beside the limitation it answers. #2142 is closed.

| thread | outcome | commit |
| --- | --- | --- |
| booleanity not enforced | documented precisely, cause pinned as a test, framing changed throughout | `77c7cbc` |
| `every_cell_is_a_bit` proves the wrong thing | renamed, and says whose property it is | `77c7cbc` |
| the fold lives only in the test | `ZerocheckClaim::discharge` owns it | `6ec388d` (#2139) |
| negative test should assert `Err` | it does | `77c7cbc` |
| all negatives go through the honest prover | adversarial prover; all four of your mutations now fail a test | `6ec388d` (#2139) |
| bind via `observe_commitment` | done | `77c7cbc` |
| assertion is on the cell count | done | `77c7cbc` |

Two things worth flagging rather than leaving for you to find:

- Your fourth mutation exposed a second problem. My first adversarial test **passed** with the residual check deleted, because the two residual checks shared `ResidualClaimMismatch` and neither was distinguishable. They are different statements; the second is now `BlendConstraintMismatch`. Verified by deleting each check in turn.
- I did not reproduce your end-to-end acceptance as a test, since the prover interface takes packed bytes and exhibiting it needs a second prover forming the message over field values — for a fact the constraint already settles. I pinned the cause instead. Say the word if you want the end-to-end version in the tree.

`scripts/check.py lint` exit 0; 490 sumcheck unit + 27 binary-pcs unit + integration passing, with and without `parallel`.

sumcheck/src/univariate_skip/committed.rs
```diff
@@ -0,0 +1,192 @@
...
+/// A statement about field cells, not about bits.
+/// A commitment cell is a field element, and nothing pins it to a bit:
+///
+/// ```text
+///     a * b - c = 0    holds for any field a, b with c = a*b
+/// ```
+///
+/// Booleanity comes only from the byte input, which holds nothing else.
+/// A prover forming the message itself, over field values, is not bound.
+/// So this is a scaffold for the mechanics, not a proof about bits.
```
Nashtare, line 35: Booleanity doesn't actually need batching. A separate arity-1 `BinaryZerocheck` per operand on `v*v - v` (a small `Composition` impl outside the crate) goes through the same `commit`, one `open_at` with an extra batch per operand, `verify_at` and `discharge`, and the honest side still uses the packed-byte prover. In a local probe it rejects the `a, b` uniform, `c = a*b` witness, and a single non-bit cell with grinding on. When the prover fits its blend, the discharge is what rejects. So enforcing it costs one zerocheck and one opening point per operand; it doesn't fight the subfield message. Batching doesn't look blocked either: `stream_round_message` evaluates the composition in the subfield and only then multiplies by `eq[row]`, so a big-field challenge can go on that weight. Could the description say that instead of "fights the design"? And could this doc tell a caller who wants a bit statement on this path that they can add those zerochecks?

sumcheck/src/univariate_skip/committed.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// # What the packed commitment fixes
+///
+/// Both halves of that, so it is not only a size win.
+/// A cell there is an `F_2`-coordinate, hence a bit by construction.
+/// It also costs 128 times less, which is the point of a binary field.
+///
+/// Relating the two claims is ring switching, already in this crate.
+/// It cannot be instantiated at a bit alphabet, because its accessor borrows.
```
Nashtare: This PR adds `ring_switch::bits`, which is exactly the ring-switch algebra at a bit alphabet, so this sentence is out of date within the PR. Maybe: the generic `ring_switch` reduction cannot be instantiated at `F_2`, `ring_switch::bits` supplies the algebra, and the driver that commits the packing is still to come.

binary-pcs/tests/committed_zerocheck.rs
```diff
@@ -0,0 +1,288 @@
...
+//! That the mechanics close.
+//! The point the proof ends on is where the operands take the claimed value.
+//! No step reads the witness on the verifying side.
+//!
+//! It does not establish that the witness is bit-valued.
+//! A cell of this commitment is a whole field element.
+//! The constraint holds on field cells that are not bits, as its tests say.
+//!
+//! The packed commitment makes it a bit statement, and is not here yet.
+//! The prover and verifier run on two independently built challengers.
```
Nashtare, line 19: Nit: the rewrap joined the challenger note to the booleanity paragraph. A blank line before "The prover and verifier run on two independently built challengers" would keep the two points apart.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// # Arguments
+    ///
+    /// - The evaluation point of the claim, over every variable of the witness.
+    /// - The batching challenge that collapses the tensor element's rows.
+    ///
+    /// # Errors
+    ///
+    /// - The point names fewer variables than one packed element absorbs.
+    /// - The batching challenge does not name exactly the absorbed coordinates.
+    pub fn new(r: &Point<EF>, r_batch: &Point<EF>) -> Result<Self, BitRingSwitchError> {
```
Nashtare: `tensor()` is the only way to form the prover's message, but it needs a value that cannot exist without `r_batch`. In the protocol, `r''` has to be drawn after ŝ is bound, which is the order `prove_ring_switch` uses. So a driver either draws `r_batch` first, or builds a throwaway reduction with a placeholder and rebuilds `eq_high`. Drawing it first is unsound. I checked this locally at `BinaryField16`: with `r_batch` known, a bit matrix Δ solving two F2-linear systems moves `incoming_claim` to a false value and leaves `initial_sum` unchanged. The honest sumcheck and the closing check then accept with a true surviving claim. Could the ordering be type-level, e.g. `new(r)` giving `tensor`/`incoming_claim` and `.batch(r_batch)` giving `weights`/`initial_sum`/`closing_weight`, with the docs saying `r` and ŝ are bound before `r''`?

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -0,0 +1,621 @@
...
+                expected: Self::ABSORBED,
+                actual: r_batch.num_variables(),
+            });
+        }
+
+        // The packing keeps the leading coordinates and absorbs the rest.
+        let (high, low) = r.split_at(r.num_variables() - Self::ABSORBED);
+
+        Ok(Self {
+            eq_high: Poly::new_from_point(high.as_slice(), EF::ONE),
```
Nashtare: `eq_high` has 2^ℓ' entries, and it is built for every construction. `incoming_claim`, `initial_sum` and `closing_weight` never read it: they use `eq_low`, `eq_batch` and `high`. A verifier using this type therefore pays a witness-sized table, 1 GiB at ℓ' = 26 over `BinaryField128`. `verify_ring_switch` only needs O(ℓ'·d²). Building it lazily in `tensor`/`weights`, or only on a prover-side type, would keep the verifier succinct.

sumcheck/src/ring_switch/bits/packing.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// # Arguments
+    ///
+    /// The witness as held, eight cells to the byte, lowest bit first.
+    ///
+    /// # Errors
+    ///
+    /// - The cell count is no power of two, so the witness covers no hypercube.
+    /// - The witness is too short to fill one element, so nothing to pack.
+    pub fn new(bits: &[u8]) -> Result<Self, BitPackingError> {
+        let stride = EF::NUM_BYTES;
```
Nashtare, line 55: For `Gf2`, `BinaryField2` and `BinaryField4` the stride is one byte (8 cells), but `from_le_byte_iter` keeps only `2^LOG_BITS` of them. The rest are silently dropped. Locally, `BitPacking::<BinaryField4>::new(&[0x0F, 0x0F]) == new(&[0xFF, 0xFF])`, and `BitPacking::<Gf2>::new(&[0xFE])` packs to zero. `BitRingSwitch::tensor` does refuse the result through its width check, but `into_poly` still hands a caller a polynomial that is not the witness. A const assertion or error for `8 * EF::NUM_BYTES != 1 << EF::LOG_BITS` would close it.

sumcheck/src/ring_switch/bits/packing.rs
```diff
@@ -0,0 +1,254 @@
...
+use p3_multilinear_util::poly::Poly;
+
+use super::basis::Coefficients;
+
+/// A bit witness read as the packed multilinear over a tower level.
+///
+/// # Overview
+///
+/// Soundness rests on the committed polynomial being a packing of bits.
+/// An arbitrary multilinear will not do.
```
Nashtare: At byte-aligned levels, `into_bytes` and `from_le_byte_iter` are inverse, so every `Poly<EF>` is `BitPacking::new` of its own bytes. No multilinear is excluded, and booleanity comes from packing itself, not from the wrapper. Maybe say that any packed multilinear unpacks to exactly one bit witness? As written, it reads as if the type carries part of the soundness argument.

Nashtare (changes requested):
Re-reviewed at 8f557f10, covering the committed chain and the new `ring_switch::bits` module.

**Earlier threads.** These are all fixed and look good:
- `discharge` with a typed rejection;
- `observe_commitment`;
- the renamed embedding test;
- the cell-count assertion.

Swapping column order in the stacked table is caught at `discharge`, as it should be.

**Booleanity.** The documentation is accurate now, but the reason given for not enforcing it doesn't hold. A separate arity-1 zerocheck per operand rejects non-bit witnesses today without a batching challenge; details inline. The caveat also needs to sit where a verifier decides, on `BinaryZerocheck` and `discharge`. I left that on #2139.

**`ring_switch::bits`.** The algebra is correct:
- I checked every identity the PR pins against an independent brute force at `BinaryField16`, using explicit bit matrices and up to 7 kept variables.
- It closes end to end through `SumcheckProver`/`verify_rounds` at 16 and 128 bits. The PR's own reduction tests never run at 128 bits.
- The accumulation matches a term-by-term reference at the production chunk size, serially and in parallel.

Two asks on the shape of `BitRingSwitch`, before a driver is built on it (inline):
1. `new` takes `r_batch`, but ŝ has to be formed and bound before `r''` is drawn. So a driver either draws `r''` too early, which is unsound (locally a forged ŝ passes with a true surviving claim), or builds the reduction twice.
2. `new` always builds the 2^ℓ' `eq_high` table. None of the three verifier-side values reads it, so a verifier using this type is not succinct.

**Description.** A few sentences overstate what this PR establishes:
- §3 says soundness "rests on the committed polynomial being a packing of bits; `BitPacking` carries it". Every packed element unpacks to exactly one bit string, and a verifier never holds a `BitPacking`, so the wrapper carries no part of the argument.
- "Booleanity holds by construction" describes a driver that doesn't exist yet.
- `embed_bits` is re-exported without a gate, so "named as such" is only true of its doc.
- The binary-pcs unit count is 59, not 27.

On #2139's side, the `operand_blends` binding is still unpinned by any test (details there).

sumcheck/src/univariate_skip/committed.rs
tcoratger: You are right and my line was wrong — thank you for going and checking it rather than taking it.

My argument only ever applied to *batching* booleanity with the conjunction, and an arity-one zerocheck needs no challenge, so it never applied at all. Your second point retires even the batching version: `stream_round_message` composes in the subfield and only then multiplies by `eq[row]`, so a big-field challenge rides on the weight rather than fighting the carrier.

`7e8b9c3` replaces the claim with the recipe, so a caller who wants a bit statement on this path can get one:

> # How a caller gets a bit statement on this path
>
> One arity-one zerocheck per operand on `v*v - v`.
> It is committed, opened and discharged the same way as the rest.
>
> It needs no batching challenge, so it keeps the subfield message intact.
> The cost is one zerocheck and one opening point per operand.

The same correction is on `BinaryZerocheck` in #2139, where booleanity is now listed under what the caller owes with both discharges. The PR description no longer says the design is being fought.

sumcheck/src/univariate_skip/committed.rs
tcoratger: Fair — it described the state before the third commit in its own PR. Reworded in `7e8b9c3` along your lines:

> Relating the two claims is ring switching.
> The generic reduction cannot be instantiated at `F_2`.
> Its accessor borrows a slice that 128 one-bit coefficients cannot fill.
>
> `ring_switch::bits` supplies that algebra.
> The driver that commits the packing is still to come.

binary-pcs/tests/committed_zerocheck.rs
tcoratger: Caught by my own rewrap, and fixed in `7e8b9c3`. The blank line is back, and the booleanity paragraph now ends on the recipe rather than trailing into the challenger note:

```text
The packed commitment makes it a bit statement, and is not here yet.
One arity-one zerocheck per operand on `v*v - v` would do it here.

The prover and verifier run on two independently built challengers.
```

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: This is the finding I am most glad you pushed on. The type forced the unsound order, and your Δ construction is what makes that concrete rather than stylistic: with `r_batch` known, two `F_2`-linear systems move `incoming_claim` while holding `initial_sum`, so the honest rounds and the closing check accept a true surviving claim.

`7e8b9c3` splits it type-level, as you proposed:

```text
    new(r)          tensor, incoming_claim
    bind r, bind s_hat, draw r''
    batch(r'')      initial_sum, closing_weight, weights
```

`BitRingSwitch::batch` returns a `BitRingSwitchBatch<'_, EF>` borrowing the first stage, and the three values that depend on the challenge live only there. Reaching the second stage is now the proof that the element was bound first, so a driver cannot express the early draw — no placeholder, no rebuild.

The ordering and the reason are on the type, not just in a commit message:

> Drawing `r''` first is unsound.
> A bit matrix solving two `F_2`-linear systems moves the claim, sum held.
> The rounds and the closing check then accept a true surviving claim.
>
> The split is what stops a driver reaching for that order.

The tests took the split as-is: the fixture hands back the challenge beside the reduction rather than folded in, so each test reaches the second stage the way a driver does.

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: Right, and it falls out of the same split. `7e8b9c3` drops `eq_high` from the struct entirely.

The first stage now holds the kept coordinates and one `d`-entry table, and your three verifier operations — `incoming_claim`, `initial_sum`, `closing_weight` — touch neither. Only `tensor` and `weights` build the `2^l'` table, and each builds it on demand.

A verifier is therefore back to `O(l'·d^2)`, rather than the 1 GiB you measured at `l' = 26`.

The prover builds the table twice, once per operation. That is deliberate and documented: it is strictly below the `2^l' · d` accumulation it feeds, so the cheaper trade is to leave a verifier out of it rather than to hold the table for a prover's convenience.

> Only `tensor` and `weights` touch the `2^l'` equality table.
> Both are prover-side, so each builds it rather than every caller paying.

sumcheck/src/ring_switch/bits/packing.rs
tcoratger: Real bug, and refused now rather than papered over. `7e8b9c3`:

```rust
if 8 * stride != Coefficients::<EF>::DIMENSION {
    return Err(BitPackingError::SubByteLevel {
        bits: Coefficients::<EF>::DIMENSION,
    });
}
```

I took the error rather than a const assertion so the width check is where the other two shape refusals already are, and so a generic caller gets a value it can handle rather than a build break.

`a_level_narrower_than_a_byte_is_refused` pins both of your cases — `BinaryField4::new(&[0x0F, 0x0F])` and `Gf2::new(&[0xFE])` — and asserts a byte-aligned level still packs. Your point about `into_poly` is the one that makes this worth an error rather than a note: the reduction's width check does catch it, but only after a caller has been handed a polynomial that is not the witness.

sumcheck/src/ring_switch/bits/packing.rs
tcoratger: Correct, and the old wording claimed something the type does not do. Reworded in `7e8b9c3`:

> At a byte-aligned level, reading bytes and reading them back are inverse.
> So every packed multilinear unpacks to exactly one bit witness.
>
> No multilinear is excluded, and booleanity comes from the packing itself.
> This type carries no part of that argument.
>
> What it does carry is the width and the level.
> The reduction it feeds cannot be handed a polynomial of another shape.

The module header said the same thing and says this instead. What remains true is narrow and worth keeping: the width and the level are checked once, which is why `tensor` can refuse a mismatched packing.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -0,0 +1,677 @@
...
+            element.scale_rows(EF::ONE - b);
+            element += agree;
+        }
+        element
+    }
+}
+
+/// The same reduction, once the batching challenge has been drawn.
+///
+/// Reaching this stage is the proof that the element was bound first.
```
Nashtare: The split fixes the trap, but this line overstates it. `batch` needs only `&BitRingSwitch`, so it can be called straight after `new(r)`, before any tensor exists, and `tensor` still works while the batch stage is alive. `the_initial_sum_is_the_sum_the_sumcheck_proves` does exactly that at lines 541-542. Locally, the same Δ forgery goes through in the order `new`, `batch(r'')`, `tensor`, then pick ŝ. So reaching this stage says nothing about ŝ being bound; only the transcript driver that absorbs `r` and ŝ before drawing `r''` enforces that. Could this and line 54 say so, so the driver still to come keeps its ordering test?

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -0,0 +1,677 @@
...
+    ) -> Result<BitRingSwitchBatch<'a, EF>, BitRingSwitchError> {
+        if r_batch.num_variables() != Self::ABSORBED {
+            return Err(BitRingSwitchError::BatchWidthMismatch {
+                expected: Self::ABSORBED,
+                actual: r_batch.num_variables(),
+            });
+        }
+        Ok(BitRingSwitchBatch {
+            reduction: self,
+            eq_batch: Poly::new_from_point(r_batch.as_slice(), EF::ONE),
```
Nashtare, line 144: Nothing checks that the challenge actually reaches this table. Building `eq_batch` at a constant point instead of `r_batch` passes the whole `ring_switch` suite. The batch-stage tests compare `weights`, `initial_sum` and `closing_weight` with each other, or with `batch.eq_batch` itself, and all of them read this same table. A stage that ignores `r''` is the forgery with a challenge every prover knows. One test that computes `sum_u eq(u, r'') * row_u` with an eq table built independently, as the `r_low` test does, would pin it.

Nashtare (approved):
Re-checked at 7e8b9c32. The earlier requests are addressed, so I'm approving; the points below are comments.

**Committed chain.** The bit-statement recipe, the ring-switch wording and the blank line all look good. A local probe runs the recipe: one `v*v - v` zerocheck per operand, with the same commit, open and discharge. It rejects the non-bit witness and accepts honest bits, so the docs match the code. Removing the `operand_blends` binding is now caught by #2139's fiber test.

**`ring_switch::bits`.** The split resolves the ordering and verifier-cost threads.
- A counting allocator on the verifier path (`new`, `incoming_claim`, `batch`, `initial_sum`, `closing_weight`) shows at most about 2 KB allocated at ℓ' = 24 over `BinaryField128`.
- Sub-byte levels are refused on the only constructor.
- The identities, the equality recurrence, end-to-end completeness and the serial/parallel accumulation all re-check against the split API.

Two small asks are inline:
- **Ordering claim.** `batch` is callable before `tensor`, and one of the PR's own tests does that. The forgery still goes through in that order, so the wording that reaching the batch stage proves the element was bound should leave that ordering to the transcript driver.
- **Missing test.** Nothing checks that `r''` reaches `eq_batch`: a constant batching point passes the whole suite.

**PR description.** It hasn't caught up with 7e8b9c32:
- §3 still says soundness "rests on the committed polynomial being a packing of bits… `BitPacking` carries it", which the commit message and `packing.rs` now say is not the case.
- §3 also describes `BitRingSwitch` from before the `new`/`batch` split.
- §2's "booleanity holds by construction" is about a driver that isn't here yet.
- The unit counts are now 494 (`p3-sumcheck`) and 59 (`p3-binary-pcs`).

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: You are right, and the line was mine to over-claim. `batch` takes `&BitRingSwitch`, so `new`, `batch(r'')`, `tensor` is expressible, and one of my own tests does it at 541-542. The split removed the *forcing* — `tensor` no longer needs the challenge to exist — and I wrote that up as though it had removed the *ordering*. It has not, and the forgery survives in that order exactly as you describe.

`95f7bcc` says so on both sites. Line 54 now separates the two:

> # What the split settles, and what it leaves
>
> It settles that `tensor` never needs the challenge.
> No driver is pushed into drawing one early just to obtain an element.
>
> It does not settle the order.
> `batch` borrows this stage alone, so a caller can reach it first.
> The forgery goes through in the order `new`, `batch`, `tensor`.
>
> Only a transcript binding `r` and the element before `r''` fixes that.
> The driver owning that transcript owes the ordering test.

And line 268 no longer claims anything:

> Holding one of these is not evidence that the element was bound first.
> Enforcing that is the transcript's job, as the stage above records.

So the obligation stays visible on the type for the driver to pick up, rather than being quietly marked discharged here.

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: Reproduced before fixing: with `eq_batch` built at the all-zero point, all 38 `ring_switch` tests passed. A stage that ignores its own challenge was undetectable, which is the forgery you name.

`95f7bcc` adds a test that weighs the rows by the challenge through the other route, `eval_eq` rather than the table:

```rust
let expected: EF = tensor
    .rows()
    .iter()
    .enumerate()
    .map(|(u, &row)| {
        let point = Point::<EF>::hypercube(u, absorbed);
        row * Point::eval_eq(point.as_slice(), r_batch.as_slice())
    })
    .sum();

assert_eq!(batch.initial_sum(&tensor), expected);
```

I also took the second half of your point: `a_weight_is_the_subset_sum_the_coordinates_select` was zipping against `batch.eq_batch` itself, so it now builds its reference from `r_batch` instead.

Under the constant-point mutation those two fail and the other 36 pass — so they are the ones carrying the property, and nothing else was silently covering for them.

PR #2142, opened 2026-09-15
feat(sumcheck): build the ring-switch algebra at a bit alphabet

tcoratger:
The previous entry closes the zerocheck against a real commitment, and pays 128× for it: `embed_bits` spends one `BinaryField128` element per witness bit. Its own docs name the fix and why it is out of reach:

> The optimisation is to commit the bits packed, `2^k` of them to an element.
>
> Ring switching then relates a claim about the bit polynomial to one about the packed one.
>
> That reduction is already written generically in this crate. It cannot be instantiated at a bit alphabet.

This entry builds the algebra that makes it reachable.

## Why the generic module cannot be reused

`ring_switch` is generic over `F: Field, EF: ExtensionField<F>`. Two things break at `F = F_2`:

**The accessor.** `TensorAlgebra::add_exterior_product` reads coefficients through

```rust
fn as_basis_coefficients_slice(&self) -> &[F]
```

a *borrowed* slice, with every impl asserting `size_of::<Upper>() == DIMENSION * size_of::<Lower>()`. For single bits that is 128 one-byte values borrowed out of a 16-byte element. No layout satisfies it, so `BinaryField128: ExtensionField<Gf2>` does not and cannot exist.

**The representation.** Even if it did, `TensorAlgebra { coeffs: Vec<F> }` at `F = Gf2` is `DIMENSION² = 16384` one-byte values — **16 KB, in the transcript**, where the same matrix bit-packed is 2 KB. So generalising the existing type over an accessor trait would not have been the right answer either: the bit case genuinely wants a different representation, and coherence would have fought a blanket impl besides.

Hence a dedicated module, `ring_switch::bits`.

## Four types, not a bag of functions

The generic module beside this one is a set of free functions over loose slices, each re-deriving the tables it needs and asserting the widths it assumes. This module does not mirror that.

| type | what it is | what it replaces |
| --- | --- | --- |
| `Coefficients<EF>` | the `F_2`-coordinates of one element | six free functions doing byte and bit arithmetic |
| `BitPacking<EF>` | a bit witness read as the packed multilinear | a `pack` function plus two dimension constants |
| `BitTensor<EF>` | an element of `EF ⊗_{F_2} EF`, as a bit matrix | — |
| `BitRingSwitch<EF>` | one reduction, and the five values it produces | `batched_weights`, `batch_rows`, `equality_element` |

Three things that buys, beyond readability:

**The bit-valued precondition becomes type-level.** The reduction's soundness rests on the committed polynomial being a packing of bits, not an arbitrary multilinear. `BitPacking` carries that, so the wrong polynomial cannot be handed to the reduction at all.

**The equality tables are built once.** Every value in the construction is a function of the same two points — the evaluation point of the incoming claim and the batching challenge. `BitRingSwitch` holds them. The free-function version rebuilt `eq(r_batch)` inside *both* `batched_weights` and `batch_rows`, on every call.

**The widths are checked once, and reported.** `PointTooNarrow`, `BatchWidthMismatch`, `WidthMismatch` replace assertions scattered across five entry points. A short batching challenge would zip-truncate and drop rows from *both* sides of the sumcheck's identity in the same direction — the kind of failure no later check catches — so it is refused at construction rather than asserted at each use.

The five values are also named for their part in the protocol rather than for their arithmetic:

```text
    tensor           the element the prover sends
    incoming_claim   what the claim being reduced must equal
    weights          the multilinear the sumcheck runs on
    initial_sum      the sum the sumcheck starts from, derived and never sent
    closing_weight   the weight the surviving claim is scaled by
```

## What changes at a bit alphabet

**A coefficient is one bit, so every product by a coefficient is a conditional add.**

```text
    general alphabet   ->  d multiplications per hypercube point
    bit alphabet       ->  the set bits, added
```

The set-bit walk lives in one place — `Coefficients::iter_set`, a small iterator peeling the lowest set bit — and both the tensor accumulation and the weight multilinear are written around it as subset sums.

**The tensor element is a `d × d` bit matrix, which is exactly one field element per row.**

```text
    one byte per coefficient   ->  16 KB at d = 128
    one bit per coefficient    ->   2 KB
```

So `BitTensor` stores the rows. The row reading is then free — it *is* the storage — and the column reading is one bit transpose away. Both are needed (rows feed the sumcheck, columns test the incoming claim), and storing one beats maintaining both.

## The basis

A tower level may hold its elements in any `F_2`-basis of the field; `TowerLevel`'s own docs say so. This module fixes the one the byte representation already defines, reached through `into_bytes`/`from_le_byte_iter`, which are inverse. `beta_0 = 1` follows, matching the matrix convention the parent module documents.

Nothing depends on *which* basis that is, only on both sides using the same one — and the payoff is the whole reason to pick it:

```rust
bits.par_chunks_exact(EF::NUM_BYTES)
    .map(|chunk| EF::from_le_byte_iter(chunk.iter().copied()))
```

**Packing moves no bits.** It reads the same bytes back as a wider integer. Cell `d·w + j` is coordinate `j` of element `w` by construction rather than by rearrangement.

## The identities, pinned

The four equations the reduction rests on are tested directly rather than inferred from the construction:

| identity | what it is for | test |
| --- | --- | --- |
| `sum_v eq(r_low, v)·column v = t(r)` | the reduction's first check | `the_incoming_claim_is_the_witness_at_the_point` |
| `column v = g_v(r_high)` — the bit planes at the point | why that check means anything | `the_columns_are_the_bit_planes_at_the_kept_coordinates` |
| `sum_w A(w)·t'(w) = initial_sum` | the sumcheck's claim, derived on both sides | `the_initial_sum_is_the_sum_the_sumcheck_proves` |
| `closing_weight(r') = A(r')` | how the verifier closes the sumcheck | `the_closing_weight_is_the_weight_multilinear_at_the_point` |

Plus: the packing agrees with the previous entry's one-element-per-bit embedding cell for cell; the coordinate round trip is the identity; the bit positions really are a basis decomposition; the equality recurrence matches the exponential hypercube sum at every width up to 5 variables; the chunked parallel accumulation matches a term-by-term reference (proptest); and each refusal above has a test, alongside the boundary just inside it.

**36 tests.** Everything is generic over the tower level, so they run at `BinaryField16` — a 16 × 16 matrix, small enough to check entry by entry — while the same code serves the `BinaryField128` level the commitment actually uses.

## Not yet wired

This is the algebra only. The reduction driver, its transcript, and the 128×-smaller commitment are the next entry; nothing in the tree calls this module yet, which is deliberate — the identities are pinned before the protocol that depends on them lands.

## Verification

- `cargo test -p p3-sumcheck --features parallel` — 483 unit (36 new) + 27 integration, passing
- `cargo test -p p3-sumcheck` — same without the feature; the accumulation goes through `par_fold_reduce`, so the serial fallback takes the same path
- `scripts/check.py lint` — exit 0


Nashtare:
Should probably be pushed directly against #2141 as it addresses a perf comment made there

tcoratger:
Agreed, and done: folded into #2141 as commit `24edd68`, and this PR is closed.

It reads better there. #2141's cost note is what prompted the algebra, and the review on #2141 showed the cost note was understating the problem: one element per bit is not only 128× too large, it is short of a bit statement, because a cell of that commitment is a whole field element and nothing in the proof constrains one to a bit.

So the packed commitment is not a performance follow-up at all. It is what makes the statement a bit statement — a witness cell there is an `F_2`-coordinate of a committed element, so booleanity holds by construction rather than by a constraint anyone has to add. Landing the algebra beside the thing whose limitation it answers puts both halves of that in front of one reviewer.

#2141 now carries three commits, in the order the story runs:

1. close the chain against a real commitment
2. say what one element per bit does not prove
3. build the ring-switch algebra at a bit alphabet

Everything from this PR's own review pass is in that third commit unchanged, including the rework from free functions into `Coefficients`, `BitPacking`, `BitTensor` and `BitRingSwitch`.

PR #2143, opened 2026-09-15
test(transcript): tie each unpriced grinding site to the config it is credited from

tcoratger:
Closes the last item of the typed-transcript campaign's phase 4: tying every grinding difficulty to the configuration its security report credits.

## The gap

A grinding difficulty lives in two places:

```text
    transcript  ->  the bits the pattern describes
    model       ->  the bits the protocol's security report credits
```

Twelve transcript modules already assert the two agree. The five that do not are exactly the five whose sites `GrindingBudget::check` never compares, because they price their own grinding and are listed in `UNPRICED_GRINDING_SITES`:

| protocol | sites |
| --- | --- |
| `p3-stir` | `folding_pow`, `query_pow`, `final_folding_pow`, `final_pow` |
| `p3-whir` | `query_pow`, `final_query_pow` |
| `p3-sumcheck-quadratic` | `round_pow` |
| `p3-sumcheck-hvzk` | `round_pow` |
| `p3-sumcheck-generic-degree` | `round_pow` |

The unpriced table records *that* these are priced elsewhere, and two tests already guard the table against going stale in either direction. Neither can check that the difficulty a pattern describes is the one the report actually credits — so for these five, the two numbers could drift apart with nothing noticing.

## The fix

One test per protocol, comparing the described difficulties against the configuration both sides read, ground and unground. No new model, no new accounting — each protocol keeps pricing its own grinding, which is the right factoring: folding these into `GrindingSites` would mean a second copy of each protocol's security model, free to diverge from the first.

STIR's fixture grinds at zero everywhere, so its ground case gives each of the four sites a **distinct** difficulty. One shared value would let a swapped pair pass.

## Verified by mutation

| mutation | result |
| --- | --- |
| pattern grinds one bit fewer than configured | sumcheck test fails |
| STIR's `final_pow` site dropped from the pattern | STIR test fails |
| WHIR's `final_query_pow` site dropped | WHIR test fails |

Worth recording: dropping STIR's final site **passed** until I fixed the fixture to actually grind. The first version of that test was vacuous.

## Validation

`cargo clippy --workspace --all-targets -- -D warnings` is clean, and the three crates pass in both serial and `--features parallel` — the mode that caught the last bug in #2140.

## After this

Nothing from the migration plan remains. The one item still deliberately out is the NARG wire format: proofs stay serde structs carried by `Carrier::External`, and moving them onto the layer's byte buffer replaces every public proof type in the workspace. That is its own campaign, and it is where the "no length confusion on the wire" guarantee would come from.


stir/src/transcript.rs
```diff
@@ -2001,4 +2002,61 @@ mod tests {
...
+        // STIR prices its own grinding, so the shared budget never compares it.
+        //
+        // Fixture state: the batch-of-one shape, ground and unground.
+        // The fixture grinds at zero everywhere.
+        //
+        // The ground case therefore sets each site to a distinct difficulty.
+        //
+        // Sharing one value would let a swapped pair pass.
+        for ground in [false, true] {
+            let mut shape = shape();
```
Nashtare: This starts from a hand-built `StirShape` and never builds a `StirConfig`. The expected list is also read through `shape.folding_pow_bits(round)` / `query_pow_bits(round)`, the same accessors `pattern()` calls. So the test checks the pattern against the shape it was built from, not against the numbers the soundness report derives (`derive_pow_bits` in `config.rs`).

I checked this with two mutations: `pow_bits: rc.folding_pow_bits` in `StirInstanceShape::new`, and `query_pow_bits` returning the per-round `folding_pow_bits`. Both pass this test and every other `p3-stir` test (lib, integration and doc). That is exactly the drift described in the PR: the report credits `rc.pow_bits` while the transcript grinds something else.

Could the ground case build a `StirConfig` whose derived difficulties are nonzero and distinct, take `StirInstanceShape::new(&config)`, and compare against `config.round_configs[i].{folding_pow_bits, pow_bits}` and `config.final_{folding_,}pow_bits`?

whir/src/transcript/mod.rs
```diff
@@ -1689,4 +1690,40 @@ mod tests {
...
+            let mut params = base_params();
+            if !ground {
+                params.pow_bits = 0;
+            }
+            let shape = shape_of(&config_from(params));
+
+            // One site per round, then one closing the run.
+            //
+            // A zero difficulty describes no step, so it contributes nothing.
+            let mut expected: Vec<(&str, usize)> = shape
```
Nashtare: Same shape here: `expected` is read from `shape.rounds[..].query_pow_bits` and `shape.final_pow_bits`, which `WhirShape::new` copied from the config. Neither `query_pow_bits: params.folding_pow_bits` nor `final_pow_bits: config.final_folding_pow_bits` in `WhirShape::new` fails this test. `every_derived_field_that_shapes_the_transcript_reaches_the_seed` does catch both, so WHIR is covered, just not by this test. Reading `expected` from `config.round_parameters[i].pow_bits` and `config.final_pow_bits` would make this test pin what its comment says.

Nashtare (commented):
The sumcheck tests are fine as pattern-conformance checks. The distinct STIR difficulties do stop a swapped label pair, and the mutations listed in the description are caught.

The STIR test doesn't yet tie the pattern to the configuration the soundness report reads. It builds a literal `StirShape`, and reads `expected` through the same accessors `pattern()` calls. So `pow_bits: rc.folding_pow_bits` in `StirInstanceShape::new`, or a `query_pow_bits()` accessor that reads the folding difficulty, passes this test and every other `p3-stir` test (lib, integration, doc). That is the drift the PR sets out to catch.

The WHIR test has the same form. There an existing seed test happens to catch the equivalent mutations.

Details and a suggested fixture change are inline.

stir/src/transcript.rs
tcoratger: You are right, and this was the worst instance of it: the test named the config in its comment while never building one. Both of your mutations reproduce exactly as described.

`b2ec186` builds a real schedule and reads the expectation from it:

```rust
let mut expected: Vec<(&str, usize)> = Vec::new();
for round in &config.round_configs {
    expected.push((FOLDING_POW, round.folding_pow_bits));
    expected.push((QUERY_POW, round.pow_bits));
}
expected.push((FINAL_FOLDING_POW, config.final_folding_pow_bits));
expected.push((FINAL_POW, config.final_pow_bits));

let shape = StirShape {
    commits_initial: true,
    instances: vec![StirInstanceShape::new(&config)],
};
assert_eq!(pow_difficulties(&shape.pattern::<F, EF>()), expected);
```

The schedule is `log_starting_degree = 20`, blowup 1, folding 4, at 80 bits, which derives `rounds = [(12, 20), (4, 18), (5, 16), (6, 16)]` and `final = (7, 9)` — every site positive, and no site's pair equal. I could not pick those numbers by hand and still have them be the report's, so the test asserts the two properties it relies on rather than trusting the schedule to keep them:

```rust
for round in &config.round_configs {
    assert!(round.folding_pow_bits > 0, "an unground folding site");
    assert!(round.pow_bits > 0, "an unground query site");
    assert_ne!(round.folding_pow_bits, round.pow_bits, "a shared value");
}
```

The unground half no longer needs a config at all, so it stays the hand-built shape and pins the one thing it was ever pinning — a zero difficulty describes no step.

Checked against three mutations: `pow_bits: rc.folding_pow_bits` in `StirInstanceShape::new`, `query_pow_bits` returning the per-round `folding_pow_bits`, and swapping the closing pair. Each fails this test and only this test.

whir/src/transcript/mod.rs
tcoratger: Confirmed, same fix. `b2ec186` reads `expected` from the config:

```rust
let mut expected: Vec<(&str, usize)> = config
    .round_parameters
    .iter()
    .filter(|round| round.pow_bits > 0)
    .map(|round| (QUERY_POW, round.pow_bits))
    .collect();
if config.final_pow_bits > 0 {
    expected.push((FINAL_QUERY_POW, config.final_pow_bits));
}
```

Both of your mutations fail it now. They were invisible for a sharper reason than the STIR one: the fixture derives `folding_pow_bits = 0` at every site, so copying folding into a query field zeroed the described list *and* the expected list together.

That also means the filter could quietly empty both sides, so the test now pins that the ground run really grinds:

```rust
assert_eq!(expected.is_empty(), !ground, "the budget did not bite");
```

One thing I checked before widening this further: the folding sites never reach this pattern. `push_delegation` emits only the bracket, so the sumcheck's own grinding is described by its shape, not this one. I confirmed that on a config with positive folding bits (`security_level = 100`), where `pow_difficulties` still returns only the query sites. So the query sites are the complete set here, and the comment now says so.

Nashtare (approved):
Reviewed at b2ec1861. Both threads are fixed.

**STIR.** The ground case now builds a real `StirConfig` and reads `expected` from `config.round_configs[i].{folding_pow_bits, pow_bits}` and `config.final_{folding_,}pow_bits`, with guards that every site is positive and no pair shares a value. The derived schedule matches the one quoted (rounds `[(12,20),(4,18),(5,16),(6,16)]`, final `(7,9)`). All three mutations now fail this test and only this test: `pow_bits: rc.folding_pow_bits` in `StirInstanceShape::new`, `query_pow_bits` reading the per-round folding difficulty, and swapping the closing pair.

**WHIR.** `expected` now comes from `config.round_parameters[i].pow_bits` and `config.final_pow_bits`, and the `expected.is_empty() == !ground` guard stops an empty list from passing on both sides. Both mutations in `WhirShape::new` now fail this test.

**Also checked**
- `push_delegation` emits only the bracket and `SumcheckShape::bind` folds `pow_bits` into the instance label, so the query sites are the complete set this pattern describes.
- The unground STIR assertion isn't vacuous: removing the zero guard in `push_pow` makes it fail.

PR #2144, opened 2026-09-15
fix(multi-stark): build opening points and batches from one description

tcoratger:
First layer of the logUp\* integration stack. It lands on its own: nothing
here is logUp\*-specific.

## The problem

`PrescribedPointPcs::open_at` and `verify_at` take two parallel lists — the
column batches, and one point per batch:

```rust
fn open_at(&self, data, protocol: &OpeningProtocol, points: &[Point<Challenge>], challenger)
```

Nothing downstream ties entry `k` of one list to entry `k` of the other. In
`multi-stark` the two were built by separate functions that agreed only by
convention:

```rust
config.pcs().open_at(
    prover_data,
    &instances.opening_protocol(),      // what is opened
    &instances.main_points(&point),     // where
    challenger,
)
```

Both walk `self.num_variables()` in the same order today, so they line up. But
the agreement is nowhere stated and nothing checks it. A drift between them
opens a table's claim at another table's point, and the proof still verifies —
a verifier would accept an evaluation of the wrong polynomial at the wrong
place.

The same convention was load-bearing a second time, where per-batch results
are handed back to the AIRs:

```rust
main_evals.iter().zip(next_columns.iter())   // batches zipped with tables
```

That is only correct while every table owns exactly one batch. The shape
agreement (`TableSpec`, `PointSchedule`) has always allowed several — there is
a test fixture in `p3-sumcheck` opening one table at two points — so this is a
latent coupling, not a stated invariant.

## The fix

One description produces both lists:

```rust
pub(super) struct OpeningSchedule<EF> {
    protocol: OpeningProtocol,   // what is opened
    points: Vec<Point<EF>>,      // where, one per batch
    owners: Vec<usize>,          // which table each batch belongs to
}
```

A single walk over the tables fills all three, so their order agrees by
construction rather than by two functions keeping a convention. A table's own
columns are then read back through `first_batch_per_table()` instead of by
position, in both the main and the preprocessed path.

Security assessment runs before any point is bound and needs the shapes alone,
so it keeps a points-free builder. Both builders derive from one
`whole_table_batch`, so the batch definition cannot fork.

## Behaviour

Unchanged. Every table is still opened exactly once, over its whole width, at
its own suffix of the bound point. What changes is that the pairing is no
longer a convention, and a table may own more than one batch without the
readback silently going wrong.

## Testing

Four unit tests beside the code. The load-bearing one is
`every_batch_keeps_the_point_it_was_built_with`: three tables owning one, two
and one batch, each batch declared with a labelled point.

```text
    batches:  t0/b0   t1/b0   t1/b1   t2/b0
    points:    10      20      21      30
```

It asserts batch `k` carries the point it was declared with, that the point
count matches `num_openings()`, and that `iter_openings()` reports the same
owners. `a_table_is_read_from_the_first_batch_it_owns` pins the readback that
the old `zip` got right only by coincidence.

`cargo test -p p3-multi-stark` passes (133 lib tests plus integration, up from
129), `p3-examples` builds, clippy and rustfmt are clean.

## Why now

The logUp\* reduction (#2133) closes on two further points — the position
columns at one, the table columns at another — so a table will own up to three
batches. That work needs this seam, but the coupling above is worth removing
regardless of whether logUp\* ever lands.


Nashtare (commented):
I checked the "behaviour unchanged" claim locally, and it holds:
- Postcard proof bytes are identical at base and head for 7 batch shapes: a single table, three heights, a preprocessed batch, a middle table with no preprocessed trace and a leading table with none, mixed-height local lookups, and a global bus.
- The bytes are the same in serial and parallel builds, and the `v0_8_0` fixture still verifies.
- Over 2000 randomized shapes, the schedule's protocol, points and owners match the old formulas.

One ask before logUp\* builds on this. On the base branch, the prover, the verifier and `security_report` all shared one `opening_protocol()`. They now build their protocols in two separate walks. WHIR's `prescribed_security` counts claims from that protocol, and changing only the security walk passes every test. Deriving it from the schedule walk, plus an equality test, would keep the certified claim count tied to what is actually opened.

Smaller points, all inline:
- the main readback indexes where the preprocessed side uses `.get`;
- nothing tests the verifier readback itself yet; the positional `zip` still passes every test, which only matters once a table owns two batches.

multi-stark/src/instance.rs
```diff
@@ -352,8 +476,9 @@ where
             .collect()
     }
 
+    /// Shape agreement for the main trace opening, without the points.
     pub(super) fn opening_protocol(&self) -> OpeningProtocol {
-        opening_protocol(
+        OpeningSchedule::<C::Challenge>::whole_table_protocol(
```
Nashtare: Sharing `whole_table_batch` keeps the batch contents in step, but the table walk now lives in two places: `opening_protocol()`, which `security_report` passes to `prescribed_security`, and `main_schedule()`, which `open_at`/`verify_at` use. On the base branch both paths called the same `opening_protocol()`. WHIR's `prescribed_security` counts the batched claims from `iter_openings` (`whir/src/pcs/security.rs:66`). So if `main_schedule` gains a batch and this walk doesn't, the report certifies fewer claims than get opened. I tried making this walk drop the successor columns, and every test in the crate still passed. Could this be derived from the schedule walk itself (for example a schedule that is generic over the point payload, built here with `()`), plus a test that `opening_protocol() == *main_schedule(p).protocol()` on both the main and preprocessed sides? The doc on `whole_table_protocol` saying the two builders cannot disagree would then hold.

multi-stark/src/verifier.rs
```diff
@@ -327,10 +329,17 @@ where
...
+    // An AIR reads its own columns out of the first batch its table is opened in.
+    //
+    // Walking the results in order agrees with the AIR order only while each table owns one.
+    let main_openings = main_schedule
+        .first_batch_per_table()
+        .into_iter()
         .zip(next_columns.iter())
-        .map(|(batch, next_columns)| TableOpening::new(batch.current(), next_columns, batch.next()))
+        .map(|(batch, next_columns)| {
+            let batch = &main_evals[batch];
```
Nashtare: `main_evals[batch]` panics if a scheme returns fewer batches than were scheduled. The old `zip` let the closing check return `OpeningCountMismatch` in that case, and the preprocessed side below still goes through `.get`. Both in-tree schemes check the count before returning, so only a scheme that breaks the `verify_at` contract can reach this. Still, `filter_map(|batch| main_evals.get(batch))` would give both sides the same rejection path.

multi-stark/src/instance.rs
```diff
@@ -387,35 +563,128 @@ where
...
+        let owners = schedule
+            .protocol()
+            .iter_openings()
+            .map(|(table, _)| table)
+            .collect::<Vec<_>>();
+        assert_eq!(owners, vec![0, 1, 1, 2]);
+    }
+
+    #[test]
+    fn a_table_is_read_from_the_first_batch_it_owns() {
```
Nashtare, line 590: This test pins `first_batch_per_table`, not how the verifier uses it. Reverting both readbacks in `verifier.rs` to the positional `zip` passes every test in the crate. Making the helper return each table's last batch fails only this test. That is expected while every table owns exactly one batch. Once a schedule gives a table a second batch, a verifier-level test would be worth adding, because that is when a positional read goes wrong.

multi-stark/src/instance.rs
```diff
@@ -18,28 +18,152 @@ pub use crate::proof::MultiStarkProof;
...
+/// A claim checked at another table's point still verifies.
+///
+/// So both come out of one pass here.
+pub(super) struct OpeningSchedule<EF> {
+    /// Shape agreement handed to the commitment scheme.
+    protocol: OpeningProtocol,
+    /// One point per batch, in the protocol's own opening order.
+    points: Vec<Point<EF>>,
+    /// Position of the table owning each batch, in the same order.
+    owners: Vec<usize>,
```
Nashtare: `protocol.iter_openings()` already yields each batch's owning table in this same order, so `owners` is a third list that has to stay in step with the other two. `first_batch_per_table` could be read off the protocol instead, as a running sum of each table's `point_schedule().len()`. A local randomized check over multi-batch schedules gives identical indices. The owner assertion in `every_batch_keeps_the_point_it_was_built_with` would then cover the data `first_batch_per_table` actually reads.

multi-stark/src/instance.rs
```diff
@@ -18,28 +18,152 @@ pub use crate::proof::MultiStarkProof;
...
-            })
-            .collect(),
-    )
+/// A scheme takes two lists: the columns of every batch, and one point per batch.
+///
+/// Nothing in it ties an entry of one list to the entry beside it in the other.
+///
+/// Lists built apart can fall out of step.
+///
+/// A claim checked at another table's point still verifies.
```
Nashtare, line 41: With today's shapes, a drift is not silent. I swapped two tables' points in `new`: when the heights differ, the prover panics at the arity check (`sumcheck/src/svo/point.rs:149`), and when they match nothing changes, because both tables get the same suffix. The silent case needs two distinct points of the same arity on one commitment, and the multi-batch schedules are what introduce that. It might be worth saying so here.

multi-stark/src/instance.rs
```diff
@@ -18,28 +18,152 @@ pub use crate::proof::MultiStarkProof;
...
+    {
+        let mut specs = Vec::new();
+        let mut points = Vec::new();
+        let mut owners = Vec::new();
+
+        // One walk over the tables fills all three lists.
+        //
+        // Their order then agrees by construction.
+        //
+        // Two functions keeping a convention is what it replaces.
```
Nashtare: This line and the doc at line 99 describe what the code replaced rather than what it does. The first two sentences of this comment stand on their own.

multi-stark/src/instance.rs
tcoratger: You are right, and I took the shape you suggested.

There is now one walk per commitment, generic over what each batch is opened against: a point on the proving path, `()` on the assessment path. `opening_protocol()` is `main_schedule(|_| ()).into_protocol()`, so a batch added to the walk reaches both by construction rather than by two functions agreeing.

`the_shapes_a_run_opens_are_the_shapes_an_assessment_reads` pins the two protocols equal on a multi-batch layout.

Your probe — dropping the successor columns from the assessment walk — is no longer expressible: there is no second walk to drop them from.

multi-stark/src/verifier.rs
tcoratger: Fixed. Both sides go through `.get` now, so a scheme returning fewer batches than were scheduled is rejected rather than indexed past.

multi-stark/src/instance.rs
tcoratger: Agreed, and noted. The helper is pinned here; what a verifier does with it is only meaningfully testable once a table owns more than one batch.

That happens in #2150, and the reader claims there are read back through `batch_answering(role)` rather than by position — a check that a positional read would fail. So the verifier-level coverage you are asking for lands with the layer that makes it possible.

multi-stark/src/instance.rs
tcoratger: Fixed — `owners` is gone. `first_batch_per_table` now walks `protocol.iter_openings()`, which is the data the scheme itself reads, so the owner assertion in `every_batch_keeps_the_point_it_was_built_with` covers what the helper uses.

multi-stark/src/instance.rs
tcoratger: Good point, and it is in the doc now: two tables of different heights fail the arity check, two of the same height share a suffix so nothing moves, and it turns silent only once one commitment carries two distinct points of one arity — which is what a table opened more than once introduces.

multi-stark/src/instance.rs
tcoratger: Fixed. Both now describe the code rather than what it replaced.

multi-stark/src/instance.rs
```diff
@@ -352,70 +450,232 @@ where
...
+            (
+                TableShape::new(2, 1),
+                vec![
+                    (OpeningBatch::new(vec![0], Vec::new()), ()),
+                    (OpeningBatch::new(vec![0], Vec::new()), ()),
+                ],
+            ),
+        ]);
+
+        assert_eq!(payloads.protocol(), shapes.protocol());
```
Nashtare: Nit: this compares two schedules built by hand, so it pins that `OpeningSchedule::new` ignores the payload, which the types already guarantee. It never calls `Instances::opening_protocol()`. Rewriting `opening_protocol()` as its own walk, either without successor columns or with an extra batch per table, passes every test in the crate. The single walk already makes that hard to do by accident, so this isn't blocking. If an `Instances` fixture is cheap to set up, comparing `opening_protocol()` with `main_schedule(..).protocol()` (and the preprocessed pair) on a batch with successor columns and a preprocessed gap would pin the actual wiring.

Nashtare (approved):
Re-reviewed at `32256110`. All earlier threads are fixed. The logUp\* stack went up rebased onto `80d9ffc4`, and `git range-diff` shows the commits are patch-identical to the ones the replies refer to.

**What holds**
- One walk per commitment. `main_schedule` and `preprocessed_schedule` are generic over the payload, and `opening_protocol()` / `preprocessed_opening_protocol()` are `*_schedule(|_| ()).into_protocol()`. Both feed `prescribed_security` in `assess_statement`. The prover and verifier open with the same walks, passing `trace_suffix` points.
- I retried the old probe. Dropping the successor columns in the shared walk now breaks 13 `whir_fibonacci` tests (including `verify_whir_compat_fixture`) and the preprocessed round trips, because the change reaches both paths. Forking the assessment path again takes rewriting `opening_protocol()` as its own walk. That is a visible rewrite, not a one-line edit.
- Both verifier readbacks go through `.get`, so a short `verify_at` result truncates the views and the closing check rejects it.
- `owners` is gone. `first_batch_per_table` walks `protocol.iter_openings()`. Making it return each table's last batch fails `a_table_is_read_from_the_first_batch_it_owns`. Reverting both readbacks to the positional `zip` still passes, which is expected until #2150 gives a table more than one batch.
- The drift doc and the two history comments are fixed.
- The schedule contents are the same as before (same shapes, batches and suffix points), and the v0.8.0 fixture still verifies.

**Also**
- The PR description still describes the first round: `OpeningSchedule<EF> { protocol, points, owners }`, a separate points-free builder sharing `whole_table_batch`, "the proof still verifies" on a drift, and a test that "pins the readback". Could it be refreshed before merge, since it will likely become the squash message?
- One nit inline on the new equality test.

multi-stark/src/instance.rs
tcoratger: There is no second walk left for it to guard. `opening_protocol()` and `preprocessed_opening_protocol()` are gone in this PR, and `security.rs` now goes through `main_schedule(|_| ()).into_protocol()`, so the rewrite you describe has nothing to rewrite. What the test still pinned — that `OpeningSchedule::new` cannot read the payload — the signature already gives, since `new` is generic over `P` with no bound.

So I dropped it rather than grow an `Instances` fixture for it. `every_batch_keeps_the_point_it_was_built_with` covers the part that is not free: entry k of `payloads()` belongs with batch k of the protocol, and the owning tables follow the same order.

PR #2145, opened 2026-09-15
feat(multi-stark)!: add the logUp* indexed-lookup reduction

tcoratger:
> **Supersedes #2133**, which was opened from a fork. GitHub's stacked pull
> requests require every branch of a stack to live in the same repository, so
> this branch moved to `Plonky3/Plonky3` to become the base of the logUp\*
> integration stack. The commits are identical, rebased onto current `main`.
>
> #2133 carried a full review pass from @Nashtare: 15 threads, all addressed
> and resolved there, and the fixes are in the commits below. That PR is worth
> reading for the reasoning; nothing in it is outstanding.

## Why

A logarithmic-derivative lookup puts an integer count over each table entry, and
counts are read back modulo the characteristic. Modulo two they say only "odd or
even", so a value looked up an even number of times is invisible:

```text
    LHS = 1/(c - v) + 1/(c - v) = 2/(c - v) = 0     in GF(2^128)
    RHS = 0/(c - t_0) + 0/(c - t_1)          = 0
```

The two sides agree for *every* challenge, as rational functions. There is no
Schwartz-Zippel escape. That is why `LookupPlan::build` rejects characteristic
two outright today, and why `check_multiplicity_height_bound` cannot help: its
bound is `< p`, which is `< 2`.

## What

logUp\* ([eprint 2025/946](https://eprint.iacr.org/2025/946)) replaces the count
with a field element. A reader names the entry it reads, and pullback and
pushforward are adjoint, so a claim about the pulled values turns inside out:

```text
    X(r) = <X, eq_r> = <T, I_* eq_r> = <T, Y>
```

The left inner product runs over the reader's rows, the right over the table's
entries. `Y` is the only new object and it is as wide as the table. The weight
over an entry is `sum_{i : I[i] = v} eq_r(i)` -- a field element that no reduction
ever touches. Two checks close it:

- that `Y` really is the pushforward, as one sum of fractions proved by the
  existing fractional GKR;
- that `<T, Y> = e`, as one degree-two sumcheck.

The reduction takes a statement and hands back evaluation claims on the table
columns and the position columns, for the caller to discharge against its
commitments. It batches several readers per table by powers of one challenge,
several tables under one reduction, and several columns per table through a
single pushforward -- so a wide entry needs no fingerprint and carries no
tuple-width aliasing hazard.

Each table draws its **own** challenge. That is not an optimization: under a
shared one, two tables can miscount opposite entries and cancel exactly. There is
a test for that.

The pushforward travels in the clear. It is only as wide as the table, which is
the regime the paper targets, and it removes a mid-protocol commitment round
entirely. The trade-off is proof size (4 KB for a 256-entry table) and that it is
not zero-knowledge; a committed variant is the follow-up for large tables.

## Field-genericity

The only field-specific hook is how a table position embeds, which is
`Field::interpolation_node`: the integer itself over a prime field, the matching
bit pattern over a binary tower. Either way the embedding is affine in the
position's bits, so the verifier evaluates the table-side denominator itself in
time linear in the variable count. Everything round-trips over both `GF(2^128)`
and `BabyBear`.

## Preconditions the caller owes

Two, both documented on the statement types:

1. **The claim point is a verifier challenge drawn after the position column and
   the table are committed.** The free range check below rests on it.
2. **The claims and the point are bound in the caller's own transcript.** The
   reduction absorbs them itself as its first step, so this is belt and braces
   rather than load-bearing.

## A position out of range needs no range check

It leaves a pole the table side cannot match, whose residue is the multilinear
extension of a nonempty row set evaluated at `r`. The position column is fixed
before `r` is drawn, so that residue is nonzero except with probability `n/|F|`
and the identity fails on its own.

Two tests hold that up. One shows no pushforward closes the pole -- neither
dropping the offending row's weight nor moving it onto another entry. The other
builds the only proof such a read admits, which has to lie about a weight the
verifier rebuilds from the claim point, and drives the real verifier against it.

## Fractional GKR

The bottom table's numerator becomes a two-variant enum, which is the breaking
change the title marks: `prove_fractional_gkr` takes a `LeafFraction` by value
rather than a `&Fraction<Poly<F>, PolyMaybePacked<F, EF>>`. The only in-tree
caller is updated.

LogUp's multiplicities stay in the base field on the path that has always used
them; logUp\*'s equality weights and pushforwards need the wider storage. A
committed golden pins the base-field path against a fixed reference across
`n = 1..8` and both denominator storages, a test pins the two numerator storages
equal, and another pins packed against scalar.

## Testing

52 tests in the new module, 26 of them unit tests beside the code they cover:

| file | tests |
|---|---|
| `plan` | ordering, alignment, tiling without gaps or overlap, determinism, reader numbering, 4 rejections |
| `position` | injectivity, first entry to zero, additivity over bits, closed form vs explicit table, agreement on the cube |
| `witness` | pushforward vs a hand-written scatter, unread entries, block placement, padding is `0/1`, honest sum is zero |
| `product` | round polynomial vs a naive oracle, starting sum, padded table owes its whole product |
| `transcript` | one shared sponge state, six ways a shape moves the seed, the statement and the pushforwards bound before what weighs them, no zero challenge |

The rest are round trips over both fields, a serde round trip, five tampering
cases, and three that pin the cryptography directly:

- **`reading_one_entry_an_even_number_of_times_does_not_cancel`** -- the direct
  regression for the failure above, asserting the weight over a twice-read entry
  is `1 - r_1`, a field element rather than a count.
- **`one_challenge_per_table_is_what_stops_two_tables_cancelling`** -- moves a
  unit of weight between two tables' entry 1 and shows the sum stays nonzero
  under distinct challenges and vanishes *exactly* under a shared one.
- **`a_position_outside_the_table_is_rejected_while_scattering`** and its two
  companions -- the free range check, from the prover's own guard down to the
  verifier's verdict.

Two goldens sit beside them, one per path, digesting a fixed run's proof bytes.
They are what catches a change that keeps prover and verifier agreeing and so
leaves every round trip green: sharing one entry challenge between tables, or
absorbing zeros in place of a pushforward, both turn them red.

`p3-multi-stark-logup-star` is registered in the cross-protocol separation
suite, with a plain shape and two single-field moves of it.

I checked the product-sumcheck test is load-bearing: breaking the padding factor
makes it fail and localizes the failure to `product.rs`, where previously only a
full round trip would have flagged it.

`cargo test -p p3-multi-stark` passes (184 lib tests plus integration), with and
without `--features parallel` -- worth checking explicitly, since the scatter and
the round-polynomial accumulation both go through `par_fold_reduce`, whose
contract diverges between serial and parallel builds if the reduction is not
associative. Clippy, rustfmt, rustdoc and `scripts/check.py lint` are clean.

## Not in this PR

Nothing calls this yet. Deliberately out of scope, in the order I would do them:

1. the AIR-facing declaration API in `p3-lookup`;
2. the coupling into the multi-STARK prover, which needs the reduction to run
   *after* the zerocheck (it consumes `eq_r`) and a second opening point on the
   main commitment;
3. the security-accounting terms, which have nothing to account for until (2);
4. a benchmark -- there is none, so this PR makes no performance claim.

One honest caveat on cost: in an AIR the looked-up column is committed anyway, so
logUp\* *adds* a position column rather than saving one. The win here is
soundness in characteristic two, not size. It only becomes a size win once
`BinaryPcs` speaks subfields -- `p3_sumcheck::ring_switch` has the reduction, it
is just not wired in -- or when the looked-up vector is virtual, as in a
Spartan-style frontend.


multi-stark/src/logup_star/tests.rs
```diff
@@ -0,0 +1,996 @@
...
+    //
+    // Fixture state: two tables of four entries, one reader each, fixed seed.
+    //
+    // A failure here is not automatically a bug.
+    //
+    // It means the transcript changed.
+    //
+    // The new digest is right exactly when that change was intended.
+    let mut rng = SmallRng::seed_from_u64(0x0_901D);
+    let instance = Instance::<Binary, Binary>::random(&mut rng, &[(2, 1, &[2]), (2, 2, &[1])]);
```
Nashtare: Nothing pins the reader batching challenge. If both `prover.rs` and `verifier.rs` use `let reader_batching = { let _ = transcript.reader_batching(); EF::ONE };`, every logUp* test still passes, this golden included, because each table here has one reader and λ⁰ = 1. Under that mutation a local probe moves a claim between two readers of one table (`X_0 += δ`, `X_1 -= δ`), proves that statement from the honest witness, and `verify` returns `Ok`. Giving one of these tables a second reader would put λ into the digest. A behavioural test would be better: prove that shifted two-reader statement and expect the `the claims the statement carries are not the ones the witness produces` panic. Column batching has the same gap in milder form, since forcing μ to one is caught only by this digest, so a two-column version of the same test would help.

multi-stark/src/logup_star/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+use p3_field::Field;
+use p3_multilinear_util::point::Point;
+pub use plan::LogupStarPlan;
+pub use proof::{LogupStarOutput, LogupStarProof, TableOutput};
+
+/// One reader's claim on the values it pulled out of a table.
+///
+/// # Soundness
+///
+/// The claim point must be drawn after the position column and the table are committed.
```
Nashtare: This precondition is load-bearing for the free range check, but it lives only here and in the module doc. `TableLookup`, `prove` and `verify` carry no pointer to it, which is where an integrator will look. The PR description also says both caller preconditions are documented on the statement types, and only this one is. The latest rewording also dropped "a challenge the surrounding protocol drew". What the argument needs is a verifier challenge the prover cannot predict when it commits, so I would keep that wording and add a one-line `# Soundness` pointer on `prove` and `verify`.

multi-stark/src/logup_star/transcript.rs
```diff
@@ -0,0 +1,825 @@
...
+        //
+        // Which table each reader belongs to is what differs.
+        //
+        //     [2 readers | 1 reader]   versus   [1 reader | 2 readers]
+        //
+        // Totals alone do not separate those.
+        //
+        // Each table's reader count is bound with its heights rather than after them.
+        let left = LogupStarShape {
+            tables: vec![table(3, 1, &[2, 2]), table(3, 1, &[2])],
```
Nashtare: This pair does not isolate the per-table reader count. Deleting `.instance(&(table.readers.len() as u64).to_be_bytes())` leaves this test green, and only the logUp* golden fails. With width 1 and readers of height 2, the width labels between tables already tell the two layouts apart. With width 2 and readers of height 2, the two shapes seed identically once the count is gone (checked locally), so width 2 here would pin what the comment describes. The `readers` case in `examples/tests/transcript_separation.rs` has the same blind spot: its tables have widths 3 and 2, so moving a reader changes the statement length (18 vs 17 values), and the pattern separates them before the count ever matters.

multi-stark/src/logup_star/witness.rs
```diff
@@ -0,0 +1,456 @@
...
+                // A pass per reader would seed a table-sized accumulator per split, per reader.
+                //
+                // That costs readers times splits times entries rather than rows.
+                //
+                // Rows split across threads, and each split fills its own copy of the table.
+                //
+                // Scattering in place would let two rows naming one entry race.
+                //
+                // The splits merge afterwards, because addition ignores order.
+                (0..table_witness.readers.len())
```
Nashtare: `flat_map` doesn't put the readers into one split tree. Rayon drives each reader's inner iterator with a fresh split of the fold consumer (`FlatMapFolder::consume` in rayon 1.12), and every leaf of that seeds both `identity`s. So a table-sized vector is still zeroed and merged per reader, per split, and the comment above doesn't hold.

I counted table-sized allocations inside `prove` with a counting allocator: a 2^16-entry table, 4096 rows in total, `parallel`. At one thread there were 19 / 67 / 1027 / 4099 for 1 / 16 / 256 / 1024 readers. At 14 threads there were up to 8195, with 158 copies live at once. The serial build makes 4 throughout.

One indexed pass over all of a table's rows with `with_min_len(num_entries)` bounds this by rows rather than readers. In a replica that took the 1024-reader case to 2 identity calls at 14 threads, against 8192 now. Scattering serially when there are few rows per entry would also work.

multi-stark/src/fractional_gkr/prover.rs
```diff
@@ -626,6 +766,21 @@ where
...
     assert_eq!(num_variables, fraction.d.num_variables());
+
+    // The round kernel reads the two halves side by side.
+    //
+    // An extension-field numerator therefore has to be stored the way its denominator is.
+    //
+    // Nothing downstream can recover from a mixed pair.
+    //
+    // Where it would surface depends on the size, so it is rejected here instead.
+    if let LeafNumerator::Ext(numer) = fraction.n {
```
Nashtare, line 777: No test reaches this assertion: skipping it passes every test in the crate, serial and parallel.

Without it, an `Ext(Scalar)` numerator over a packed BabyBear denominator is accepted silently at n = 2, 3 and 4, because the unpack shim runs first. Larger tables fall into the `unreachable!` arms instead. Those arms carry the same message, so a `should_panic(expected = "must share a storage mode")` test only pins this line if it uses a small table, e.g. n equal to the lane count.

multi-stark/src/logup_star/prover.rs
```diff
@@ -0,0 +1,249 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The proof, and the evaluation claims the caller must discharge against its commitments.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the statement and the witness disagree on how many tables or readers there are.
+    ///
+    /// Panics if a row names an entry its table does not have.
```
Nashtare, line 49: nit: three more panics are reachable from `prove`'s inputs and aren't listed here:
- a reader whose row count isn't `2^n` for its point (`witness.rs:80`);
- a column whose length isn't the table's entry count (`product.rs:93`);
- the statement shapes `LogupStarPlan::new` rejects.

A line each, or a pointer to `LogupStarPlan::new`, would cover them.

Nashtare (approved):
Reviewed at 3097e88a.
- **Rebase:** the six carried-over commits are byte-identical to #2133's final head.
- **New refactor commit:** it doesn't change behaviour. The code is identical apart from signatures, and both goldens pass before and after it, serial and parallel.
- **#2133 threads:** all 15 are addressed.

What checks out:
- The statement is absorbed, via `Cdc::observe` on both sides, before λ. The #2133 forgeries along λ and μ, and a moved claim point, are now rejected.
- Transcript order and the per-table entry challenges are correct. The embedding bound holds for BabyBear, KoalaBear, Goldilocks and the binary towers.
- 23 malformed proofs and two oversized postcard length prefixes all return `Err` without panicking.
- The existing logUp path is byte-identical to main. I compared direct fractional-GKR dumps and full `MultiStarkProof` bytes, serial and parallel, at 1 thread and at default threads.
- Packed and forced-scalar logUp\* proofs are byte-identical across 9 shapes over BabyBear and GF(2^128), including the `nv == lanes` boundary.

The main ask is a test for reader batching. Forcing λ to one on both sides leaves every test green, and under that mutation a claim moved between two readers proves and verifies. Forcing μ to one is caught only by the golden digest.

Smaller notes, all inline:
- the scatter still allocates a table-sized buffer per reader per split, despite the #2133 reply;
- nothing tests the mixed-storage assertion;
- the reader-split test doesn't isolate the reader count;
- `prove` and `verify` don't point to the claim-point precondition.

multi-stark/src/logup_star/tests.rs
tcoratger: Confirmed, and fixed behaviourally rather than by digest alone.

`lambda = EF::ONE` in both files left all 52 logUp\* tests green, exactly as you said — every fixture table had one reader, so `lambda^0 = 1`.

Two new tests now go through the real prover: one moves a claim between two readers of one table, one between two columns, and both expect the `the claims the statement carries are not the ones the witness produces` panic. Each was checked against its own mutation: forcing `lambda` to one fails the reader test, forcing `mu` to one fails the column test.

The golden fixture is also widened to a two-reader, two-column table, so both challenges now reach the digest — it alone catches `lambda = 1` as well.

**The golden digest changed**, from `460d8984f75ef5c7bcad06133f058e3fbe051dcf21124462500b4259d4bb2f8d` to `ef40ccf9220c64d51e0b9b8e9990037b622bd82cb681cf3e25794066279128d2`. That is the fixture widening, not a transcript change.

multi-stark/src/logup_star/mod.rs
tcoratger: Fixed, including the wording I had dropped.

"a challenge the surrounding protocol drew" is back on `Reader` and in the module doc, along with the point about a challenge the prover cannot predict when it commits. `TableLookup`, `prove` and `verify` each carry a one-line `# Soundness` pointer now, so an integrator meets it wherever they enter.

You were right that the PR description claimed both preconditions were on the statement types when only one was.

multi-stark/src/logup_star/transcript.rs
tcoratger: Confirmed and fixed. Both tables in the fixture are width 2 now.

I checked it the way you did: deleting `.instance(&(table.readers.len() as u64).to_be_bytes())` now fails this test, where before only the golden caught it.

The same blind spot in `examples/tests/transcript_separation.rs` is worth a separate look — its `readers` case has widths 3 and 2, so the statement length separates the two layouts before the count matters. I have left that one as is rather than widen it silently; say the word and I will.

multi-stark/src/logup_star/witness.rs
tcoratger: You are right, and it was worse than the numbers you quoted. `flat_map` splits per element, so this was never bounded by reader count at all.

Measured with a counting allocator on a 2^12-entry table, 4096 rows, `--features parallel`:

| readers | before | after |
|---|---|---|
| 1 | 5678 | 2 |
| 16 | 4712 | 2 |
| 256 | 8192 | 2 |
| 1024 | 8192 | 2 |

Even at **one** reader the old code seeded 5678 table-sized accumulators.

Fixed by numbering a table's rows once across its readers and walking that range with `with_min_len(num_entries)`, so a split is never smaller than the table it allocates. The comment now describes what the code does rather than what I thought it did.

One cost worth naming: each row now does an `O(log readers)` `partition_point` to find its owning reader. Negligible against the field add, but a per-chunk lookup would remove it if a table ever has very many readers with very few rows each.

multi-stark/src/fractional_gkr/prover.rs
tcoratger: Confirmed and fixed. `a_mixed_leaf_storage_is_rejected_before_the_size_decides` uses a table at exactly the lane count, which is the only size that reaches this line rather than an `unreachable!` arm.

Verified the way you framed it: removing the assertion makes the new test fail, and it passed before.

multi-stark/src/logup_star/prover.rs
tcoratger: Added, all three: a reader whose row count is not the one its claim point addresses, a table column that does not carry one value per entry, and the statement shapes the layout rejects, pointed at its own documentation.

PR #2146, opened 2026-09-15
feat(lookup,multi-stark)!: declare and plan indexed lookups

tcoratger:
Third layer of the logUp\* integration stack: the AIR-facing declaration, and
the planner that turns declarations into a logUp\* statement. Nothing runs it
yet — that is the next layer.

## Declaring an indexed read

A bus lookup fingerprints arbitrary symbolic expressions and carries a signed
multiplicity. An indexed read is a different shape: a row names the entry it
reads, and the values it pulled are *opened against a commitment*. An
expression has nothing to open, so the declaration names committed columns:

```rust
fn push_indexed_read(&mut self, table: &str, position: usize, payload: impl IntoIterator<Item = usize>);
fn push_indexed_table(&mut self, name: &str, window: TraceWindow, columns: impl IntoIterator<Item = usize>);
```

A reader names its table by name, as `Kind::Global(String)` does, so a batch
that reorders its AIRs plans the same statement.

A table may live in the main or the preprocessed window: range and bitwise
tables are preprocessed in practice. A reader is always main, because its
position column and the values it pulled are witness data. That asymmetry is in
the types rather than in a runtime check.

## Planning

`IndexedPlan::build` resolves declarations across the batch from the AIRs and
their trace heights alone, so both sides derive the same plan and nothing
travels in a proof. `statement` then pairs each reader with the claim point and
the values that reader pulled.

Every structural failure is a typed error, since all of them are reachable from
an AIR set that looks legitimate:

```text
    ColumnOutOfRange   RepeatedColumn      EmptyPayload
    EmptyTable         TableWithoutEntries TableTooLarge
    TableProvidedTwice TableNotProvided    TableNotRead
    WidthMismatch
```

## Soundness

The claim point must be a challenge drawn after the position columns and the
tables are committed, which is the precondition the reduction already carries
on `Reader`. The planner takes the point as given and documents that it cannot
check it.

## Testing

7 new tests in `p3-lookup`, 11 in `p3-multi-stark`, beside the code they cover.
The load-bearing one is `table_order_follows_the_name_and_not_the_air`: two
batches declaring the same tables in different AIR order must plan the same
statement, since the verifier derives its plan from the AIRs alone.

`cargo test -p p3-lookup` passes (62), `cargo test -p p3-multi-stark` passes
(199 lib plus integration), clippy and rustfmt are clean.

## Open question for review

`push_indexed_read` and `push_indexed_table` sit on `InteractionBuilder`, with
no-op defaults, following the `push_exclusive_interaction` precedent — three
mechanisms already share that trait, and an AIR doing both kinds of lookup then
carries one bound and one symbolic pass. The cost is that the trait's name now
undersells what it covers. A separate `IndexedLookupBuilder` would read better
in isolation and costs nothing at the multi-STARK call site. Happy to split it
if that is the preference.


lookup/src/builder.rs
```diff
@@ -168,6 +175,58 @@ pub trait InteractionBuilder: AirBuilder {
...
+    ///
+    /// The position column holds the field's embedding of the entry index.
+    ///
+    /// Over a prime field that embedding is the index itself.
+    ///
+    /// Over a binary tower it is the tower element whose bit pattern is the index.
+    ///
+    /// Recording builders override this.
+    /// Other builders inherit a no-op default.
+    fn push_indexed_read(
```
Nashtare, line 225: Nothing rejects this declaration yet, so it fails open. A 16-row AIR declares `push_indexed_read("t", 0, [1])` and `push_indexed_table("t", Main, [2])`, with position 0, payload 7 and a table column constrained to 0. `IndexedPlan::build` sees the read, but multi-stark `prove`/`verify` at this head return `Ok`, and so do batch-stark's `prove_batch`/`verify_batch` (local probes). Both backends build their statement from `Lookups::from_air`, which never reads `indexed_reads()`, and #2150 only wires multi-stark, so batch-stark stays this way. On the open question I'd go with a separate `IndexedLookupBuilder` with required methods. A backend that doesn't run the reduction then fails to compile against such an AIR, and `InteractionSymbolicBuilder` implementing both traits keeps it to one symbolic pass. If the methods stay here, could batch-stark (and multi-stark until the runtime lands) reject non-empty indexed declarations, the way `LookupPlan::build` rejects exclusive interactions?

multi-stark/src/indexed.rs
```diff
@@ -0,0 +1,684 @@
...
+        //
+        // Tables are keyed by name, not by the position of the AIR providing them.
+        //
+        // A batch that reorders its AIRs therefore still plans the same statement.
+        let mut tables: BTreeMap<String, IndexedTablePlan> = BTreeMap::new();
+        for (air, lookups) in declared.iter().enumerate() {
+            for table in lookups.tables() {
+                let placement = TablePlacement {
+                    air,
+                    window: table.window,
```
Nashtare, line 147: Every planner test builds its tables through `providing`, which hard-codes `TraceWindow::Main`, so no test plans a preprocessed table. Replacing this with `window: TraceWindow::Main` passes the whole suite. The window decides which commitment the table claims are opened against later, and a preprocessed range table checked against main columns would let the prover choose the table. Could one planner test with a `Preprocessed` provider assert the placement keeps its window?

multi-stark/src/indexed.rs
```diff
@@ -0,0 +1,684 @@
...
+    pub table: TablePlacement,
+    /// Every reader of this table, in caller order then emission order.
+    pub readers: Vec<ReaderPlacement>,
+}
+
+/// Every indexed lookup in one batch, in the order both sides walk it.
+///
+/// Tables are ordered by name rather than by the position of the AIR providing them.
+///
+/// A batch that reorders its AIRs therefore still plans the same statement.
```
Nashtare: Only the table order is independent of AIR order here. Readers of one table follow caller order, and the `air` fields are caller positions. Swapping two reader AIRs of `t` changes the reader list from `[(0, 3), (2, 4)]` to `[(2, 4), (0, 3)]` (local probe), so the logUp* statement changes. That's fine, since both sides use the same instance order and the seed binds it. But it means the name sort gives a canonical order rather than keeping the two sides in agreement. Could this, the comment at 135 and the PR description say that instead?

Nashtare (changes requested):
The planner holds up well locally:
- Every error variant is reachable, and each one returns an error rather than panicking.
- Table order is deterministic.
- `statement` round-trips through the logUp\* prover and verifier for two readers of different heights with multi-column payloads, and its output claims match the committed columns.
- Mutations on the sort, `WidthMismatch`, payload order, suffix slicing and the column checks each fail at least one new test.

Main ask: the declaration currently fails open. An AIR with a violated `push_indexed_read` proves and verifies under multi-stark at this head, and under batch-stark too (local probes). #2150 wires only multi-stark, so batch-stark stays that way once the whole stack lands. The exclusive-interaction precedent pairs its no-op default with a rejection in `LookupPlan::build`; nothing here does the same.

On the open question, I'd split out an `IndexedLookupBuilder` with required methods, so a backend that doesn't run the reduction refuses such an AIR at compile time. Failing that, an explicit rejection in batch-stark, and in multi-stark until the runtime lands.

Smaller points, both inline:
- no planner test uses a preprocessed table;
- the "reorders its AIRs, same statement" wording: reader order follows caller order.

lookup/src/builder.rs
tcoratger: Took your preference: `IndexedLookupBuilder` is now its own trait, with required methods and no defaults.

`InteractionSymbolicBuilder` implements it (recording), and the multi-STARK folders and the debug checker implement it ignoring — a read has no row-local form. batch-stark's folders do **not**, so the AIR you probed is now a compile error there rather than a silent `Ok`:

```
error[E0277]: the trait bound `ProverConstraintFolderWithLookups<'a, ...>: IndexedLookupBuilder` is not satisfied
```

That is strictly better than a runtime rejection, and it needs no ongoing discipline in batch-stark. The migration cost is one line per AIR that declares reads.

One gap I want to name rather than leave implied: multi-stark at *this* layer still accepts such an AIR without running the reduction, since the runtime lands in #2150. If the two ever land apart, multi-stark wants the interim rejection you describe.

multi-stark/src/indexed.rs
tcoratger: Fixed. `a_table_keeps_the_window_its_air_declared` plans a `Preprocessed` provider and asserts the placement keeps its window, next to a main-window provider of identical shape.

Checked against your exact mutation: replacing `window: table.window` with `window: TraceWindow::Main` now fails only this test.

multi-stark/src/indexed.rs
tcoratger: You are right, and the doc overclaimed. Corrected on the struct and at the collection site.

The name sort gives the *table* list one canonical order. Readers follow caller order and carry their AIR's position, so swapping two reader AIRs of one table does change the statement — which is fine, since both sides read the same instance order under a seed that binds it. That is a canonical order, not an agreement mechanism, and the doc says so now.

The PR description is updated to match.

multi-stark/src/folder.rs
```diff
@@ -371,6 +373,42 @@ fn eval_boundary_io<AB: AirBuilder>(builder: &mut AB, cells: &[BoundaryPublic])
 ///
 /// A batch whose AIRs declare lookups but which carries no lookup link is rejected up front.
 /// No proof is therefore ever produced or accepted with its lookups silently discarded.
+impl<'a, F, Var, Acc> IndexedLookupBuilder for MultilinearFolder<'a, F, Var, Acc>
```
Nashtare, line 402: This impl landed between the `InteractionBuilder` impl's doc block and that impl, so rustdoc attaches "Ordinary constraints only: every lookup declaration is dropped ... No proof is therefore ever produced or accepted with its lookups silently discarded" to the indexed impl. The `InteractionBuilder` impl loses its doc. For indexed reads that last sentence is not true at this head, since multi-stark drops them and nothing rejects the batch yet. Could this impl move below the `InteractionBuilder` impl? Related: the trait doc at `lookup/src/builder.rs:204` says "A builder that cannot discharge a read therefore never implements this trait", but this folder and `DebugConstraintBuilder` implement it without discharging anything. The rule is really about backends (a prover whose folders implement it must run the reduction). Could it say that?

multi-stark/src/indexed.rs
```diff
@@ -0,0 +1,735 @@
...
+    #[test]
+    fn table_order_follows_the_name_and_not_the_air() {
+        // Two batches describing the same tables, with the providing AIRs swapped.
+        //
+        //     batch A:  air 1 provides "a",  air 2 provides "b"
+        //     batch B:  air 1 provides "b",  air 2 provides "a"
+        //
+        // An AIR order would make these two different statements.
+        //
+        // A name order makes them the same one, so the plan lists "a" first either way.
```
Nashtare: Nit: the two plans aren't the same statement. The providing `air` index for "a" is 1 in one and 2 in the other, and the test compares only the names. This is the same point the struct doc now makes, so "so the plan lists \"a\" first either way" would be enough here.

Nashtare (approved):
Re-reviewed at `18bf140f`. The blocking thread is fixed for batch-stark and uni-stark, and the window and ordering threads are fixed in code. I checked this round by reading the trait, every impl, and the backend bounds.

**What holds**
- `IndexedLookupBuilder` has four required methods and no defaults. The only impls in the workspace are `InteractionSymbolicBuilder` (records), `DebugConstraintBuilder` (no-op, used only by the debug constraint checkers, which produce no proof), and the two multi-STARK folders. There is no blanket or wrapper impl.
- batch-stark's `prove_batch` and `verify_batch` bound on `Air<ProverConstraintFolderWithLookups>` and `Air<VerifierConstraintFolderWithLookups>`. Both folders implement only `InteractionBuilder`, so an AIR bounded on `IndexedLookupBuilder` fails both calls with E0277 instead of verifying. uni-stark has no impl of either trait, so the same holds there. Neither backend reads `indexed_reads()` anywhere.
- The only way I found around it is deliberate: an AIR-local helper trait, implemented to declare the read on the symbolic builder and as a no-op on batch-stark's folders. That is the same freedom an AIR already has to feed different constraints to different builders, so I'm not asking for anything here.
- `a_table_keeps_the_window_its_air_declared` plans a `Preprocessed` provider and asserts the window, so `window: TraceWindow::Main` fails it.
- The struct doc and the collection-site comment now say the name sort gives the table list a canonical order, and that readers follow caller order under a seed that binds it.
- The migration claim holds. An AIR that declares reads adds `+ IndexedLookupBuilder` to its bound, and multi-stark's `ProverAir`/`VerifierAir` need no change because they name folder types that implement the trait. Against `main` the change is additive, so the `!` is conservative.
- The refactor adds no bound for AIRs that don't declare reads, and one symbolic builder still records both kinds in a single `eval`.

**Still open**
- As you noted, multi-stark at this head still proves such an AIR without running the reduction: both folders implement the trait as a no-op, and `prove`/`verify` never call `IndexedPlan::build`. That is fine as long as #2150 lands with this.
- The PR description wasn't updated. It still says "a batch that reorders its AIRs plans the same statement", still calls `table_order_follows_the_name_and_not_the_air` the load-bearing test for that, and still has the "Open question" section describing no-op defaults on `InteractionBuilder`.
- #2149 now shows as conflicting against this rebased head and needs a rebase.

Two small doc points inline.

multi-stark/src/folder.rs
tcoratger: Moved below the `InteractionBuilder` impl, so that doc block goes back to the impl it was written for. The indexed impl has a short one of its own: reads are dropped here because a read has no row-local form, and the reduction takes the declarations off the symbolic pass instead.

The trait doc now makes the claim about the backend rather than the builder — every method is required, a backend whose builders implement it undertakes to run the reduction, and its own folders may still drop a declaration. Also dropped the `# Why this is a trait of its own` header while in there.

multi-stark/src/indexed.rs
tcoratger: Right, thanks. The comment now says the name order lists `"a"` first either way, which is all the assertion compares, and notes that the placements still differ because a different AIR provides `"a"` in each.

PR #2148, opened 2026-09-15
perf(dft): parallelize coset LDE and expose completed blocks 

Al-Kindi-0:
Speeds up `Radix2DitParallel` coset LDEs and lets consumers process completed output blocks while the transform runs.

- `fix(dft)`: build twiddle tables outside the cache locks; concurrent misses reuse the first published entry instead of serialising on construction.
- `perf(dft)`: prepare coset twiddle tables in parallel and evaluate the nonzero cosets concurrently; coset zero runs in place after the others join.
- `perf(matrix)`: schedule `reverse_matrix_index_bits` in ~64 KiB work units instead of one Rayon job per row, plus a benchmark for the permutation.
- `feat(dft)`: `coset_lde_batch_with_blocks` and `lde_output_block_rows` on `TwoAdicSubgroupDft`, with default implementations. Existing methods and results are unchanged; `Radix2DitParallel` hands out blocks of `2^(log_h/2)` rows from its second butterfly half.

Measured in `miden-vm`'s recursive proof aggregation with only this crate varying and recorded `−8.6 % / −16.3 %` on AMD EPYC 9R45 and `−3.9 % / −8.4 %` on Graviton 5 64 cores with Poseidon2 / Blake3, respectively.


Nashtare:
@Al-Kindi-0 seems `cloned_dfts_complete_with_cold_and_warm_caches` failed twice (timed out). Would you mind having a look and making sure CI is ✅ ?

dft/tests/shared_twiddle_cache.rs
```diff
@@ -0,0 +1,101 @@
...
+                    };
+                    assert_eq!(actual, expected);
+                }
+            });
+        }
+    }
+}
+
+#[test]
+fn cloned_dfts_complete_with_cold_and_warm_caches() {
```
Nashtare: I'm not sure the test actually catches that deadlock

Nashtare: Also it's fairly slow (>1min on windows and like 30+ sec on Linux). A test that actually forces the steal order should be much faster and fail on main (i.e. highlighting the PR tweak), so I'd suggest replacing this one with that.

dft/src/radix_2_dit_parallel.rs
```diff
@@ -199,11 +182,47 @@ impl<F: TwoAdicField + Ord> TwoAdicSubgroupDft<F> for Radix2DitParallel<F> {
...
+        shift: F,
+        transform: T,
+    ) -> Self::Evaluations
+    where
+        T: FnOnce(&mut RowMajorMatrixViewMut<'_, F>, Layout),
+    {
+        self.lde_with_consumer(mat, added_bits, shift, transform, &|_, _| {})
+    }
+
+    fn lde_output_block_rows(&self, input_height: usize, _added_bits: usize) -> usize {
```
Nashtare: `lde_output_block_rows` and `coset_lde_batch_with_blocks` must be overriden together but nothing enforces that, and a wrapper forwarding only one would then hand over blocks of different sizes than advertised

dft/src/radix_2_dit_parallel.rs
```diff
@@ -199,11 +182,47 @@ impl<F: TwoAdicField + Ord> TwoAdicSubgroupDft<F> for Radix2DitParallel<F> {
...
+        1 << (log2_strict_usize(input_height) / 2)
+    }
+
+    #[instrument(skip_all, level = "debug", fields(dims = %mat.dimensions(), added_bits = added_bits))]
+    fn coset_lde_batch_with_blocks<T, C>(
```
Nashtare: Also if an implementation returns a block size that doesn't divide the output height, it seems the default silently skips the trailing rows instead of panicking

dft/tests/shared_twiddle_cache.rs
Al-Kindi-0: Indeed, removed it now. The cache is now a private `TwiddleCache` type where Its unit tests check the actual invariant.
I couldn't find a deterministic way to force rayon's steal order from outside the crate; a timing-based reproduction does hang on main reliably here, but I'd rather not add another timing-dependent test. Can add it as an ignored diagnostic if you want it.

dft/src/radix_2_dit_parallel.rs
Al-Kindi-0: Fixed by removing `lde_output_block_rows`. `coset_lde_batch_with_blocks` now takes a consumer factory, called once with the block height the DFT will publish.

dft/src/radix_2_dit_parallel.rs
Al-Kindi-0: Changed as part of the previous point

Nashtare (approved):
really nice! Thanks a bunch!

PR #2149, opened 2026-09-15
feat(security,multi-stark)!: charge and make room for the indexed reduction

tcoratger:
Fourth layer of the logUp\* integration stack: the soundness terms, and the
three seams the reduction needs in the statement. The prover and verifier
phases that use them are the next layer.

## Soundness terms

`MultilinearLogupStarParams`, charged alongside the existing lookup terms.
Each is derived from the identity it rests on rather than from the code:

| term | degree |
|---|---|
| reader batching | `max_readers_per_table - 1` |
| entry challenge | non-padding leaves, over the nonzero field |
| fractional GKR | `2L + 3L(L-1)/2`, over this reduction's arity |
| product sumcheck | `2 * max_table_variables` |
| column batching | `num_column_claims - 1` |

Two of these are worth a second opinion.

**Padding is excluded from the entry-challenge count.** A padding leaf carries a
zero over a one: no pole, so it cannot raise the cleared numerator's degree.
The charge is the exact non-padding count, which claims *more* security than a
loose "leaves in the padded table" bound would.

**The per-table union costs one degree, not zero.** There is one entry challenge
per table. Fix every `c_s` for `s != t`, and the condition becomes
`S_t(c_t) + K = 0`; clearing `D_t` of degree `P_t` gives `N_t(c_t) + K·D_t(c_t) = 0`,
which has degree `P_t` rather than `P_t - 1`. The other tables can fail to cancel
the identity and still shift it. So the honest bound is the pole count, not one
less.

There is no analogue of `lookup-opening-link` or `lookup-air-link`: the classical
lookup couples into the zerocheck through two extra challenges, and this one
couples through the opening, which is an equality rather than a probabilistic
check.

## The three seams

**The proof carries an optional indexed round** — what each reader claims it
pulled, and the reduction tying those claims to the tables. The claims travel
because a verifier cannot have them earlier: they are values of committed
columns, and only an opening supplies those. Two things stand between a prover
and a forged claim. The reduction binds every claim before drawing any challenge
of its own, and the closing check compares each against the opening at the bound
point.

**The transcript gains a bracket** between the zerocheck and the opening. It is
described only when the batch declares an indexed read, so a batch declaring none
keeps the transcript it already had — the committed WHIR fixture would otherwise
have needed regenerating for every existing user.

**The opening layout names the point each batch is taken at** rather than
carrying it:

```rust
enum BatchPoint { Bound, Position, Table }
```

This closes a gap the soundness work exposed. `security_report` assesses
`opening_protocol()`, which has no point to work with, while the run builds
`main_schedule()`. Once the reduction adds batches, those two could describe
different openings — and an assessment covering fewer batches than the proof
opens overstates the bound it reports. Both now read one layout: the assessment
takes the shapes, the run resolves the names.

## Testing

`cargo test -p p3-multi-stark -p p3-security -p p3-lookup` passes (201 / 110 / 62
lib tests plus integration), clippy and rustfmt clean.

The layout has its own tests: one that a batch is resolved against the point it
names and not its neighbour, one that a run with no reduction cannot answer a
batch naming the reduction's point.

The committed WHIR fixture is regenerated, since the proof gained a field and
postcard is not self-describing. Its transcript is unchanged.


security/src/multilinear.rs
```diff
@@ -107,6 +130,63 @@ pub fn reduction_terms(
         add("lookup-opening-link", 1.0, field_bits); // theta
         add("lookup-air-link", 1.0, field_bits); // eta
     }
+    if let Some(logup_star) = air.logup_star {
```
Nashtare, line 227: These five terms cover the challenges this reduction draws, but not the bound point `r` it inherits. Going from `X(r) = <T, I_* eq_r>` to `X = T∘I` (and the free range check) is a Schwartz–Zippel event over `r`. With honest positions and one wrong payload value at row `i0`, every check here passes whenever `eq(r, i0) = 0`, and no term involves `r`. Could this also charge about `max_reader_variables` over the field `r` is drawn from? It's negligible next to the entry-challenge term, but the block reads as the complete list.

security/src/multilinear.rs
```diff
@@ -107,6 +130,63 @@ pub fn reduction_terms(
...
+        //
+        //     sum_i w_i / (c - iota(I_i))  -  sum_v Y_v / (c - iota(v))  =  0
+        //
+        // It holds for every `c` exactly when `Y` is the pushforward of the positions.
+        //
+        // Clearing the denominators leaves a numerator of degree below the pole count.
+        //
+        // Every pole is a reader row or a table entry.
+        //
+        // A table's own challenge is fixed while the others vary.
```
Nashtare: Nit: this is inverted. The argument (as in the PR description) fixes the other tables' challenges and varies this table's `c_t`. That bounds the event by `max_t P_t`, so charging `num_leaves` summed over tables is safe but looser by up to the table count, which may be worth a line.

multi-stark/src/indexed.rs
```diff
@@ -681,4 +681,35 @@ mod tests {
...
+        //     table  "t"   2^2 = 4 entries
+        //     reader air 0 2^5 = 32 rows
+        //     reader air 2 2^3 = 8 rows
+        //                  ------------------
+        //     leaves       44, so a 2^6 padded table
+        let reader = Declaring::bare(4).reading("t", 3, vec![1, 2]);
+        let table = Declaring::bare(2).providing("t", vec![0, 1]);
+        let second = Declaring::bare(4).reading("t", 3, vec![1, 2]);
+
+        let plan = plan(&[&reader, &table, &second], &[5, 2, 3])
```
Nashtare, line 757: This fixture has one table, so sum and max over tables coincide. Taking `num_leaves` or `num_column_claims` as a max, or `max_readers_per_table` as a min, survives `cargo test -p p3-security -p p3-multi-stark`. A second table (2^3 entries, one column, read by the 2^3 AIR) kills all three.

multi-stark/src/security.rs
```diff
@@ -297,6 +363,11 @@ where
         num_fractions,
         max_message_width: plan.max_width,
     });
+    // The same plan feeds the soundness terms and the opening shapes below.
+    //
+    // An assessment covering fewer batches than the proof opens overstates the bound.
+    let indexed = IndexedPlan::build::<C::Val, C::Challenge, A>(&instances.airs(), &heights)?;
+    let logup_star = indexed.as_ref().map(IndexedPlan::security_params);
```
Nashtare, line 377: No test calls `security_report` on a batch that declares an indexed read: replacing `logup_star` with `None` here, or assessing `opening_protocol(None)` below, survives the suite. Local probe with #2150's squares batch: the report is complete (`logup-star-entry-challenge` 108.09, `-fractional-gkr` 106.98, `-product-sumcheck` 109.68). At this layer, though, the same batch panics `prove`.

multi-stark/src/transcript.rs
```diff
@@ -163,6 +169,12 @@ pub struct MultiStarkShape {
     pub instances: Vec<MultiStarkInstanceShape>,
     /// Grinding difficulty each delegated sumcheck round runs at.
     pub pow_bits: usize,
+    /// Whether any AIR of the batch declares an indexed read.
+    ///
+    /// A batch declaring none plays no indexed bracket.
+    ///
+    /// Its transcript is then the one it had before indexed lookups existed.
+    pub has_indexed: bool,
```
Nashtare, line 177: `examples/tests/transcript_separation.rs:760` builds `MultiStarkShape` with a struct literal, so `cargo check -p p3-examples --tests` fails with `E0063: missing field has_indexed`. Could that suite also gain a `has_indexed` case? Nothing checks that the flag moves the seed: a pattern that never describes the bracket survives every test, and `assert!(!seeds_agree(&base_shape(), &MultiStarkShape { has_indexed: true, ..base_shape() }))` kills it.

multi-stark/src/verifier.rs
```diff
@@ -207,9 +212,15 @@ where
...
+    // Indexed lookups change the described sequence, so the plan is settled first.
+    //
+    // Both sides derive it from the AIRs alone, so no proof value reaches it.
+    let indexed_plan = IndexedPlan::build::<C::Val, C::Challenge, A>(&airs, &log_heights)
+        .map_err(VerificationError::IndexedLookup)?;
+
     let mut transcript = MultiStarkVerifierTranscript::<C::Challenger, C::Val>::new(
         challenger,
-        MultiStarkShape::new::<C::Val, A>(&airs, &log_heights, pow_bits),
+        MultiStarkShape::new::<C::Val, A>(&airs, &log_heights, pow_bits, indexed_plan.is_some()),
```
Nashtare, line 223: At this layer nothing reads `proof.indexed`. Local probe: a junk `IndexedLookupProof` attached to the committed fixture (7169 -> 7356 bytes) still gives `Ok(())`. For a batch that declares a read, zeroed zerocheck rounds make `verify` panic (`expected Begin Protocol indexed_lookup`) instead of returning `Err`. #2150's match fixes both, so this only matters if the two land apart.

multi-stark/src/transcript.rs
```diff
@@ -445,6 +472,14 @@ where
     /// # Returns
     ///
     /// Whatever the delegated run produced.
+    /// Lend the sponge to the indexed-lookup reduction, bracketed as a sub-protocol.
```
Nashtare: Nit: this method went in between `lookup_argument`'s doc comment and its `fn`, so `indexed_lookup` now carries the lookup argument's doc (its own summary sits under `# Returns`) and `lookup_argument` has none. The verifier side has the same splice.

multi-stark/src/instance.rs
```diff
@@ -477,79 +558,123 @@ where
...
-                        instance.num_variables,
-                        instance.air.preprocessed_width(),
-                        instance.air.preprocessed_next_row_columns(),
-                    )
-                }),
-        )
-    }
-
-    /// Schedule the main trace opening at the bound point.
+    /// What the main commitment opens, before any point is bound.
```
Nashtare: Nit: the old summary line above was kept, so this doc opens with two summaries.

multi-stark/src/instance.rs
```diff
@@ -477,79 +558,123 @@ where
...
+                        OpeningSchedule::<C::Challenge>::whole_table_batch(width, next_columns),
+                        BatchPoint::Bound,
+                    )],
+                )
+            })
+            .collect::<Vec<_>>();
+
+        // The reduction closes on two points, and both land on this commitment.
+        //
+        // Both live in the main trace, so both land on this commitment.
```
Nashtare: Nit: this repeats line 589. Also, "land on this commitment" holds only for a `TraceWindow::Main` table; a preprocessed table's batch goes through `preprocessed_layout`.

multi-stark/src/instance.rs
```diff
@@ -477,79 +558,123 @@ where
...
+        if let Some(plan) = indexed {
+            for table in plan.tables() {
+                for reader in &table.readers {
+                    tables[reader.air].1.push((
+                        OpeningBatch::new(alloc::vec![reader.position], Vec::new()),
+                        BatchPoint::Position,
```
Nashtare: The two new tests build `OpeningLayout` by hand, so this tagging is untested. Each of these survives the suite: reader batches tagged `Table`, position batches dropped, the table batch pushed on `table.readers[0].air`, and the table batch dropped in `preprocessed_layout`. Could a test check the batches and names that `main_layout(Some(&plan))` and `preprocessed_layout(Some(&plan))` produce for a two-AIR plan?

multi-stark/src/proof.rs
```diff
@@ -21,6 +53,8 @@ pub struct MultiStarkProof<C: MultiStarkConfig> {
     pub commitment: Commitment<C>,
     /// Fractional-GKR lookup proof, absent when no AIR declares interactions.
     pub lookup: Option<FractionGkrProof<C::Challenge>>,
+    /// Indexed-lookup round, absent when no AIR declares an indexed read.
+    pub indexed: Option<IndexedLookupProof<C::Val, C::Challenge>>,
```
Nashtare, line 58: Nit: this changes the proof format and adds a public field. `MultilinearAirParams::logup_star`, the new `MultiStarkShape::new` argument and the new error variants are breaking too, so the title may want the `!` that #2145's has.

Nashtare (changes requested):
Reviewed at 1790f554, on #2146 at 2de94ec1. I derived the five terms from #2145's reduction, probed the proof and transcript seams, and ran 24 mutations.

**Blocking: `p3-examples` tests no longer compile.** `examples/tests/transcript_separation.rs:760` builds `MultiStarkShape` with a struct literal:
```
$ cargo check -p p3-examples --tests
error[E0063]: missing field `has_indexed` in initializer of `MultiStarkShape`
```
Only the labeler has run on this PR, so CI hasn't seen it.

**Soundness terms: they hold**
- **Reader batching.** One λ, powers `0..R_t-1` per table, drawn after the statement is absorbed: `R_max − 1`.
- **Entry challenge.** `c_t` is drawn after the pushforwards and rejection-sampled nonzero, and a zero root denominator is rejected. With the other challenges fixed, the bad event is a nonzero polynomial of degree ≤ `P_t` in `c_t`, so the extra degree is right. Padding leaves `(0, 1)` don't depend on `c`, so excluding them is right too.
- **Fractional GKR.** Each layer draws λ, draws a branch, and runs `i` cubic rounds: `2L + 3L(L−1)/2`. `log2_ceil(num_leaves)` matches `LogupStarPlan::num_variables`.
- **Product sumcheck.** Degree 2, and the padding factor is linear: `2 · max_table_variables`.
- **Column batching.** One power per column across all tables, both in the prover and in `claimed_sum`: `C − 1`.

One event isn't charged (inline): going from `X(r) = <T, I_* eq_r>` to `X = T∘I` is a Schwartz–Zippel event over the bound point `r`. It is negligible next to the entry-challenge term, but the list reads as complete.

**Also checked**
- The no-indexed transcript is unchanged. The regenerated fixture is the old one plus one `0x00` tag byte, and every sumcheck and opening byte is identical. The seed describes the bracket only when `has_indexed` is set.
- Assessment and run build their openings through the same `main_layout`/`preprocessed_layout` from the same plan.
- `cargo test` (default and `parallel`) and clippy `-D warnings` on multi-stark/security/lookup, plus nightly fmt, are clean.

**Most of the new seams aren't pinned by a test** (inline). These mutations survive `cargo test -p p3-security -p p3-multi-stark`:
- `num_leaves` or `num_column_claims` taken as a max over tables, or `max_readers_per_table` as a min (the fixture has one table);
- the pattern never describing the bracket, or either side passing `has_indexed: false`;
- `security_report` dropping `logup_star`;
- reader batches tagged `Table`, position batches dropped, the table batch pushed on the reader's AIR, or the preprocessed table batch dropped.

**This layer on its own:** `verify` never reads `proof.indexed`, so a junk indexed round attached to the committed fixture still verifies. For a batch that declares a read, a bad proof panics on a transcript mismatch instead of returning `Err`. #2150's match handles both, so this only matters if the layers land separately.

security/src/multilinear.rs
tcoratger: You are right that the list read as complete when it was not. Added `logup-star-claim-point`, charged at the tallest reader's variable count over the field the point is drawn from.

Your framing is the one in the comment: turning `X(r) = <T, I_* eq_r>` into `X = T . I` is a Schwartz-Zippel event over that point, and so is the free range check. One wrong pulled value survives every other term wherever its equality weight vanishes, and that weight is multilinear in the reader's own variables.

security/src/multilinear.rs
tcoratger: Fixed — the comment had the quantifier backwards. It now says that one table's challenge varies while the others are held fixed, that the event is bounded by the largest table's pole count, and that summing over tables is what is charged: safe, and looser by up to the table count.

multi-stark/src/indexed.rs
tcoratger: Confirmed and fixed. `the_soundness_shape_sums_over_tables_where_it_says_it_does` uses two tables of different shapes, so sum and max part company:

```text
    leaves           4 + 16 + 8 + 8 + 8 = 44      summed
    column claims    2 + 1 = 3                    summed
    readers/table    max(1, 2) = 2                extreme
    reader variables max(4, 3, 3) = 4             extreme
    table variables  max(2, 3) = 3                extreme
```

Each of the three swaps you named fails it.

multi-stark/src/security.rs
tcoratger: Noted, and partly addressed. The plan now reaches both shape agreements through one walk (see #2144), so `opening_protocol(indexed)` and the schedule cannot describe different openings — the substitution you probed is no longer expressible.

A `security_report` test on a batch that declares a read is still worth having; at this layer such a batch panics `prove`, which is why it lands with #2150 rather than here.

multi-stark/src/transcript.rs
tcoratger: Confirmed, and my fault — I checked `-p p3-multi-stark --all-targets` and never the crate whose call site I had moved. Fixed, and the suite gains an `indexed` case whose flag is the only field that moves.

`declaring_an_indexed_read_describes_one_more_delegation` also pins what you asked for directly: the pattern gains exactly two steps, and `seeds_agree` is false across the flag.

multi-stark/src/verifier.rs
tcoratger: Right on both counts, and #2150 now does exactly that — the match on `(plan, proof.indexed)` rejects a section nobody asked for, and a missing one, before the bracket is played.

They are stacked, so the two land together.

multi-stark/src/transcript.rs
tcoratger: Fixed on both sides. Each driver has its own doc again, and `lookup_argument` has its summary back.

multi-stark/src/instance.rs
tcoratger: Fixed.

multi-stark/src/instance.rs
tcoratger: Fixed, and the claim corrected: a preprocessed table's batch goes through `preprocessed_layout`, so "land on this commitment" only ever held for a main-window table.

multi-stark/src/instance.rs
tcoratger: Fixed by construction rather than by test. Each batch now carries what it answers — `Air`, `Position { table, reader }` or `TableColumns { table }` — and #2150 reads claims back through `batch_answering(role)`.

Three of your four mutations stop compiling or change what a role resolves to; the fourth (dropping a batch) fails the round trip, because the claim the reduction closed on then has nothing to be compared against.

multi-stark/src/proof.rs
tcoratger: Retitled `feat(security,multi-stark)!:`. #2146 gained one too, since the trait split moved its surface.

multi-stark/src/proof.rs
```diff
@@ -1,10 +1,42 @@
...
+///
+/// A verifier cannot have them earlier.
+///
+/// They are values of committed columns, and only an opening supplies those.
+///
+/// Two things then stand between a prover and a forged claim.
+///
+/// The reduction binds every claim before drawing any challenge of its own.
+///
+/// The closing check compares each claim against the opening at the bound point.
```
Nashtare: At #2150's head this no longer holds. `0fb9de23` compared `statement.claims()` with `plan.statement(&reduction.point, &main_openings).claims()`; at `ab962b48` that comparison is gone and `statement` is never read after the reduction, so a reader's claim is not tied to its payload column. Forged proofs that verify are in the #2150 review. Either the comparison comes back there, or this doc and the `logup-star-claim-point` rationale need to change with it.

security/src/multilinear.rs
```diff
@@ -167,4 +275,121 @@ mod tests {
...
+        assert!((find("logup-star-entry-challenge") - (99.0 - libm::log2(24.0))).abs() < 1e-12);
+
+        // Five layers: two draws each, plus 0+1+2+3+4 cubic rounds.
+        assert!((find("logup-star-fractional-gkr") - (99.0 - libm::log2(40.0))).abs() < 1e-12);
+
+        // Four rounds of a quadratic summand.
+        assert!((find("logup-star-product-sumcheck") - (100.0 - libm::log2(8.0))).abs() < 1e-12);
+
+        // Five roots separate six column claims.
+        assert!((find("logup-star-column-batching") - (100.0 - libm::log2(5.0))).abs() < 1e-12);
```
Nashtare, line 655: This test checks five of the six labels. `logup-star-claim-point` isn't pinned: deleting its `add`, or charging degree 0, survives `cargo test -p p3-security -p p3-multi-stark`. The fixture has `max_reader_variables: 3`, so `assert!((find("logup-star-claim-point") - (100.0 - libm::log2(3.0))).abs() < 1e-12)` kills both. Could that go here?

multi-stark/src/instance.rs
```diff
@@ -352,28 +497,155 @@ where
...
+                            against: against(BatchRole::Air, instance.num_variables),
+                        },
+                    )],
+                )
+            })
+            .collect::<Vec<_>>();
+
+        if let Some(plan) = indexed {
+            for (table, planned) in plan.tables().iter().enumerate() {
+                if planned.table.window != TraceWindow::Preprocessed {
```
Nashtare, line 613: The reply says dropping a batch fails the round trip. That holds for the main window, but not here. At #2150's head, the only AIR declaring a read in any test is `SquaresBatch`, whose table is `TraceWindow::Main`. So skipping this push (`if true || planned.table.window != ...`) survives by reading. So do `IndexedWitness::build`'s preprocessed arm and the verifier's `preprocessed_evals.iter().flatten().nth(batch)` lookup. Could #2150's end-to-end tests gain a preprocessed-window table?

multi-stark/src/instance.rs
```diff
@@ -352,28 +497,155 @@ where
...
 
-    pub(super) fn opening_protocol(&self) -> OpeningProtocol {
-        opening_protocol(
-            self.num_variables()
-                .iter()
-                .zip(self.widths().iter())
-                .zip(self.next_columns())
-                .map(|((&log_height, &width), next_columns)| (log_height, width, next_columns)),
-        )
+    /// Shape agreement for the main trace opening, without the points.
```
Nashtare: Nit: `main_schedule` still opens with two summaries. This line is the old one; it comes from #2144 (`f42fa1b5`), so it may be easier to drop there.

multi-stark/src/transcript.rs
```diff
@@ -853,6 +909,7 @@ mod tests {
         let MultiStarkShape {
             instances,
             pow_bits,
+            has_indexed: _,
```
Nashtare: Nit: `has_indexed: _` opts the new field out of the mechanism the doc above describes. `every_field_of_the_shape_reaches_the_seed` is covered by the dedicated test, but `no_two_one_step_mutations_collide` never sees `has_indexed`. Could this feed a `("has_indexed", shape)` row instead?

Nashtare (approved):
Re-reviewed at 3f9116f2 (its own commits `052938cc` and `3f9116f2`, on the pre-rebase #2146 head `91b81876`). I range-diffed against 1790f554, re-ran the seed and multi-table mutations, re-derived the new term, and read #2150's head `ab962b48` for the parts deferred to it. Approving this layer.

**The blocking item is fixed.** `cargo check -p p3-examples --tests` compiles. A pattern that never describes the bracket now fails `declaring_an_indexed_read_describes_one_more_delegation` and `no_two_configurations_of_any_two_protocols_share_a_seed`.

**What holds**
- **`logup-star-claim-point`.** `D_j(y) = X_j(y) - Σ_i eq(y,i) T[I_j[i]]` is fixed by the commitments before `r` is drawn and is multilinear in the reader's `n_j` trailing coordinates. The first nonzero `D_j` bounds the event by `n_j/|EF|`, at most `max_reader_variables/|EF|`, and an out-of-range entry leaves a weight of the same degree, so one term covers both. `r` comes from `challenge_extension`, so `field_bits` is the right field. It is a different event from the zerocheck terms; any overlap only overcounts.
- The entry-challenge comment has the quantifier right now.
- The multi-table fixture matches the code (44 / 3 / 2 / 4 / 3), and `num_leaves` as max, `num_column_claims` as max and `max_readers_per_table` as min each fail it.
- Driver docs, the repeated window comment and the `!` title are fixed.
- At #2150's head, reader batches tagged `TableColumns`, dropped position batches, and the table batch pushed on `readers[0].air` all fail the round trip. None of the three stops compiling, though.

**The term this PR adds assumes a check #2150 no longer makes.** `proof.rs:26` says each reader claim is compared with its opening at the bound point. #2150 did that at `0fb9de23` and dropped it at `ab962b48`; details and forged proofs that verify are in the #2150 review.

**Before merge**
- The branch needs a rebase onto #2146's `18bf140f`. A trial `git rebase --onto 18bf140f 91b81876` conflicts in `052938cc` with #2136's `skip` field (`security/src/multilinear.rs`, `multi-stark/src/security.rs`). Keeping both fields fixes the library code; by reading, `logup_star_params()` and `a_statement_without_indexed_lookups_charges_nothing_new` then need `skip: None`, and main's `plain()` needs `logup_star: None`. `3f9116f2` applies cleanly after that.

**Still open**
- The `security_report` test on a batch that declares a read did not land in #2150 either. Only `SquaresBatch` declares one, and it goes through plain `prove`/`verify`, so passing `None` for `logup_star` in `assess_statement` survives by reading.
- Inline: the new term has no test, the preprocessed-window table batch is untested end to end, and two nits.

multi-stark/src/proof.rs
tcoratger: The comparison is back in #2150, joined to the reduction-output one rather than replacing it, so a reader's claim is tied to its payload column again.

Reworded the bullet to name the column instead of the point, since the reader claims are discharged at the bound point and the reduction's own claims at the two points it closes on. The `logup-star-claim-point` rationale stands as written.

security/src/multilinear.rs
tcoratger: Added, as written: `assert!((find("logup-star-claim-point") - (100.0 - libm::log2(3.0))).abs() < 1e-12)`. Deleting the `add` and charging degree 0 both fail it now.

multi-stark/src/instance.rs
tcoratger: #2150 gained one. `Squares::Fixed` provides the table from a preprocessed trace, and `a_reduction_run_against_an_uncommitted_preprocessed_table_is_rejected` drives it through the plan, `IndexedWitness::build`'s preprocessed arm and this batch.

Checked it is that path and not the main one: panicking in the verifier's `TraceWindow::Preprocessed` arm fails only that test, and disabling the column comparison fails it too.

multi-stark/src/instance.rs
tcoratger: Gone. The method carrying that summary was the `opening_protocol` replacement, which this PR no longer has, so `main_schedule` is the only walk. Tightened the surviving doc in #2144, where the stale line about the payload being nothing at all also lived.

multi-stark/src/transcript.rs
tcoratger: Destructured now, and fed in as a `("has_indexed", shape)` row, so `no_two_one_step_mutations_collide` sees it.

PR #2150, opened 2026-09-15
feat(multi-stark): run the indexed reduction inside the proof

tcoratger:
Final layer of the logUp\* integration stack. An AIR can now declare an indexed
read and have it proved and verified end to end.

## Where it runs, and why there

```text
    5. zerocheck                  -> bound point r
    6. indexed reduction          <- here
    7. open main at r, position_point, table_point
    9. match every indexed claim against what the opening produced
```

The reduction has to sit between the zerocheck and the opening, and there is no
other option.

A reader's claims *are* its AIR's column values at `r`, which the zerocheck has
just produced. But the reduction also *closes* on claims at two further points,
which the opening has to cover. Run it earlier and it has no claims; run it later
and it needs a second opening round.

So the claims travel in the proof. A verifier cannot hold them any earlier —
they are values of committed columns, and only an opening supplies those. Two
things then stand between a prover and a forged claim:

- the reduction absorbs every claim in its own statement step before drawing any
  challenge, so a claim cannot be tuned to the challenges that weigh it;
- the closing check compares each claim against the value the commitment opens at
  `r`, which is what authenticates it.

Fixed before it can be tuned, authenticated before it is believed.

## Recovering an entry from a position column

A position column holds an entry under the reduction's embedding, not the entry
itself, and the prover needs the integer to scatter with.

Rather than a field-specific inverse or a new caller-supplied input, the prover
embeds every entry the table has and matches:

```text
    entry 0..2^m  ->  embed(entry)  ->  entry
```

That is one pass over a table the reduction already walks, so it costs nothing
asymptotically. It also makes an out-of-range row a *miss* rather than a wrong
answer, which is exactly the guard the reduction wants.

## Testing

`prove_verify_indexed_lookup_roundtrips_through_pcs` is the end-to-end case: a
four-entry table of squares, provided by one AIR out of its main trace, read by
an eight-row AIR that names each entry twice.

```text
    table  : entry  0 1 2 3   ->  value  0 1 4 9
    reader : names  0 1 2 3 3 2 1 0
             holds  0 1 4 9 9 4 1 0
```

Every entry is read exactly twice, which is the case a logarithmic-derivative
lookup cannot tell from zero reads in characteristic two — the failure this whole
stack exists to avoid.

`a_reader_pulling_the_wrong_value_is_rejected` moves one pulled value off the
table. The position column is untouched, so the pushforward is honest and the
reduction still runs; what breaks is the claim tying the table's columns to it.
The prover's own assertion catches it before a proof exists, and the test also
drives the verifier in case one ever does.

`cargo test -p p3-multi-stark -p p3-lookup -p p3-security -p p3-multilinear-util`
passes (201 / 62 / 110 / 147 lib tests plus integration), clippy and rustfmt
clean.

## One small addition outside multi-stark

`PolyView::into_slice` returns the evaluations a view borrows, with the lifetime
of the table rather than of the view. Taking a reference through the view ties
the slice to a temporary, which is what the witness builder needs to avoid.


multi-stark/src/verifier.rs
```diff
@@ -355,6 +405,20 @@ where
         })
         .collect::<Vec<_>>();
 
+    // The indexed reduction ran on claims the proof supplied, which nothing had authenticated.
+    //
+    // The opening has now produced the committed values at that same point.
+    //
+    // Comparing the two is what binds the reduction to the commitment.
+    if let Some((statement, _)) = &indexed {
```
Nashtare: `output` is dropped here, so the reduction's `position_claims` and `column_claims` are never compared with the Position and Table batches the openings just verified. Local probe: a prover whose reduction witness uses the table `[0,1,5,9]` against a committed `[0,1,4,9]` (the reader holds 5 for entry 2, twice) gets `Ok` from `verify`, and so does a reader naming entry 5 of a 4-entry table. Comparing each Position batch with `output.tables[t].position_claims[k]` and each Table batch (main or preprocessed) with `column_claims` rejects both and keeps the suite green. Could the layout hand back the batch index of each Position/Table batch, so this check doesn't re-derive the push order?

multi-stark/src/verifier.rs
```diff
@@ -282,6 +292,38 @@ where
...
 
+    // 6. Verify the indexed reduction against the point the zerocheck bound.
+    //
+    // The claims come from the proof, since only the opening supplies committed values.
+    //
+    // The closing check is what ties them to the commitment.
+    let indexed = match (indexed_plan.as_ref(), proof.indexed.as_ref()) {
+        (Some(plan), Some(round)) => {
+            let statement =
+                plan.statement_from_claims(&reduction.point, round.reader_claims.clone());
```
Nashtare: `round.reader_claims` is proof data, and `statement_from_claims` `assert_eq!`s its reader count and per-reader width. Local probe: pushing or clearing a reader, or a column, panics `verify` instead of returning `Err`. Could the verifier measure the list against the plan first, or `statement_from_claims` return a `Result`?

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1936,3 +1936,184 @@ fn security_checked_roundtrip_for_an_air_bound_only_by_boundary_io() {
...
+            ProverInstances::new(vec![
+                ProverInstance::new(&reader, Table::new(reads.transpose()), &pk, &[]),
+                ProverInstance::new(&table, Table::new(squares.transpose()), &pk, &[]),
+            ]),
+            0,
+            &mut challenger(),
+        )
+    }));
+
+    let Ok(Ok(proof)) = proved else {
```
Nashtare, line 2176: With the prover's assertion in place this branch is unreachable: the fixture panics in `prove` ("the claims the statement carries are not the ones the witness produces"), and `catch_unwind` would accept an unrelated panic too. Disabling the `IndexedClaimsUnopened` check, removing the Position/Table batches, or removing the prover assertion each survive every test. Could this build a forged proof, e.g. a crate-internal test that feeds the reduction a forged table witness, so the verifier is what rejects it?

multi-stark/Cargo.toml
```diff
@@ -29,6 +29,7 @@ p3-multilinear-util.workspace = true
 p3-security.workspace = true
 p3-sumcheck.workspace = true
 p3-util.workspace = true
+hashbrown.workspace = true
```
Nashtare: `cargo sort --workspace --grouped --check` fails here ("Dependencies for multi-stark are not sorted"), while it passes on #2149's head: `hashbrown` goes before `itertools`.

multi-stark/src/indexed.rs
```diff
@@ -374,6 +419,127 @@ impl<EF: Field> IndexedStatement<EF> {
...
+                    .table
+                    .columns
+                    .iter()
+                    .map(|&column| source.poly(column).into_slice())
+                    .collect(),
+            );
+
+            // One entry per embedding, so a row's committed value names its entry.
+            let entries = 1usize << table.table.num_variables;
+            let mut entry_of = HashMap::with_capacity(entries);
```
Nashtare: `with_capacity(2^m)` allocates `2^(m+1)` buckets under hashbrown's 7/8 load factor. For BabyBear that is roughly 34 bytes per entry, about 550 MiB for a 2^24 table, and tables of the same height rebuild the same map. Could one map for the largest `m`, with a per-table `entry < 1 << m` check, serve every table?

multi-stark/src/verifier.rs
```diff
@@ -282,6 +292,38 @@ where
         }
     };
 
+    // 6. Verify the indexed reduction against the point the zerocheck bound.
```
Nashtare, line 297: Nit: this and the main opening below are both `// 6.`, and the preprocessed opening and the zerocheck close are both `// 7.`, while the doc list now numbers the steps 6-9. The two opening comments also still say "at their suffixes of the (same) bound point", and prover.rs has the same duplicate `// 7.`.

Nashtare (changes requested):
Reviewed at 0fb9de23, on #2149 at 1790f554. I traced both transcripts, tampered with every field of the indexed section, ran eleven composition shapes and eight mutations.

**Blocking: the reduction's output claims are never checked against the openings.** `LogupStarProof::verify` returns `position_claims` at `position_point` and `column_claims` at `table_point`, and its doc leaves discharging them to the caller. The verifier opens batches at both points. But the closing check binds `(statement, _)` and compares only the reader claims at `r`, and nothing outside `logup_star/` reads either claim vector.
- Local probe, wrong pulled value: the prover's reduction witness uses the table `[0,1,5,9]`, the committed table is `[0,1,4,9]`, and the reader column holds 5 for entry 2, twice. `verify` returns `Ok`.
- Local probe, out-of-range read: a reader names entry 5 of a 4-entry table, and the prover maps the miss to entry 0 in the reduction witness only. `verify` returns `Ok`.
- Mutations that remove the Position batch, the main Table batch or the preprocessed Table batch each survive every test.

A verifier-only check of about 50 lines rejects both forgeries and keeps `cargo test -p p3-multi-stark` green. It compares each Position batch with `output.tables[t].position_claims[k]`, and each Table batch (main or preprocessed) with `column_claims`.

**Malformed claims panic the verifier.** `statement_from_claims` `assert_eq!`s the reader count and per-reader widths of `round.reader_claims`, which is proof data, so adding or dropping a reader or a column panics `verify`. The other 18 tampers of the indexed section all return `Err`:
- pushforwards;
- the GKR root, layer claims, round polynomials and truncation;
- position and column claims and their counts;
- the product sum, rounds, width and PoW;
- the round removed, or added.

**The negative test never reaches the verifier.** `a_reader_pulling_the_wrong_value_is_rejected` panics inside `prove`, and `catch_unwind` accepts any panic. Removing the prover's assertion, or disabling the `IndexedClaimsUnopened` check, survives every test.

**What holds**
- **The #2133 gap stays closed.** Every reader claim is absorbed before the reduction draws a challenge, and removing that absorb is caught by two #2145 tests.
- **Honest shapes beyond the 8/4 fixture verify:**
  - two reader AIRs on one table;
  - a reader shorter than its table;
  - a 16-row preprocessed table;
  - one AIR reading two tables;
  - an indexed read next to a classical lookup;
  - one table read twice;
  - a self-read;
  - crossing heights;
  - a two-column table.
- `PolyView::into_slice` is safe and lifetime-sound.
- Default and `parallel` tests, clippy `-D warnings` and nightly fmt pass.

**Also**
- `cargo sort --workspace --grouped --check` fails on `multi-stark/Cargo.toml`.
- Step 9 of the description ("match every indexed claim") and the comment above the closing check claim more than the code matches.

multi-stark/src/verifier.rs
tcoratger: This was a real soundness hole, and you are right about every part of it. I opened the Position and Table batches and then never looked at what came back — the type's own doc says "Nothing here is authenticated ... discharging them against the commitments is the caller's job", and I was the caller.

Fixed. After the openings, each reader's Position batch is compared with `output.tables[t].position_claims[k]`, and each table's batch — main or preprocessed, by its window — with `column_claims`. A mismatch is `IndexedClaimsUnopened`.

I took your suggestion on how to find them rather than re-deriving the push order: every batch now carries what it answers, and the check asks `batch_answering(BatchRole::Position { table, reader })`. Re-deriving the order is exactly the class of bug the rest of this stack has been removing.

Worth reporting: wiring this immediately caught a second bug of mine. The prover was still opening every batch at the bound point while the verifier used the role's point, so the honest round trip failed on the opening sumcheck. That would have shipped silently before, because nothing compared the two.

multi-stark/src/verifier.rs
tcoratger: Fixed. `statement_from_claims` returns `Result` now, with `ClaimCount` and `ClaimWidth`, and the verifier maps them to `IndexedLookup`. Proof data decides no count by assertion any more.

multi-stark/tests/whir_fibonacci.rs
tcoratger: You are right that the branch was unreachable and that `catch_unwind` would have accepted any panic. The comment now says what the test actually pins — the prover's own refusal — rather than implying the verifier rejects it.

I also added `a_row_naming_an_entry_its_table_does_not_have_is_rejected`, which is a real second case: entry 5 of a four-entry table has no embedding to match, so the prover cannot place that row.

The forged-proof test you describe is the right one and I have **not** written it: it needs a prover that reduces against one table while committing another, which no caller of the public API can express, so it wants the crate-internal harness you suggested. I would rather say that than ship something that looks like a forgery test and is not. Happy to add it if you want it before merge.

multi-stark/Cargo.toml
tcoratger: Fixed — `hashbrown` sits before `itertools` now, and `cargo sort --workspace --grouped --check` passes.

multi-stark/src/indexed.rs
tcoratger: Fixed. One map sized by the largest table serves the whole batch, since the embedding does not depend on which table an entry belongs to, and a smaller table checks the entry it recovers against its own size.

For your 2^24 example that is one map rather than one per table of that height.

multi-stark/src/verifier.rs
tcoratger: Fixed. The prover and verifier phases are renumbered to match their doc lists, and the two opening comments now say they open at every point a claim was left at, rather than at the bound point.

multi-stark/src/verifier.rs
```diff
@@ -343,6 +391,48 @@ where
         .transpose()
         .map_err(VerificationError::Opening)?;
 
+    // The reduction's own output claims are not authenticated by anything so far.
+    //
+    // The openings just verified carry the committed values at those same points.
+    //
+    // Without this comparison the prover picks the tables the reduction runs against.
+    if let Some((_, output)) = &indexed {
```
Nashtare: This block replaced the comparison of `statement.claims()` with `plan.statement(&reduction.point, &main_openings)` instead of joining it, so `round.reader_claims` is never checked against the readers' payload columns at `r`. Local probe: the reader holds `0 1 5 9 9 5 1 0`, the committed table is `[0,1,4,9]`, positions are honest, and the prover builds its statement with `statement_from_claims` on the values the positions pull instead of the zerocheck openings. `verify` returns `Ok`. So does a payload of all 7s, and a 16-entry preprocessed table with every held value off by one. Restoring the old check after `main_openings` rejects all three with `IndexedClaimsUnopened` and keeps `cargo test -p p3-multi-stark` green. Could it come back next to these two?

multi-stark/src/verifier.rs
```diff
@@ -282,6 +292,42 @@ where
...
+    //
+    // The closing check below is what ties them to the commitment.
+    let indexed = match (indexed_plan.as_ref(), proof.indexed.as_ref()) {
+        (Some(plan), Some(round)) => {
+            // Proof data decides no count here.
+            //
+            // A list of the wrong shape is rejected rather than measured against.
+            let statement = plan
+                .statement_from_claims(&reduction.point, round.reader_claims.clone())
+                .map_err(VerificationError::IndexedLookup)?;
```
Nashtare: This `?` returns before the `indexed_lookup` bracket is played and without `transcript.abort()`, so the driver's drop check panics. Local probe: pushing or clearing a reader, or a column, in `reader_claims` panics with "Dropped unfinalized VerifierState: 4 pattern step(s) were never replayed and 0 NARG byte(s) were never checked." instead of returning `Err(IndexedLookup(..))`. With `transcript.abort()` before the return, all four return `Err`. Could this take the same shape as the `Err` arm below?

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1936,3 +1936,250 @@ fn security_checked_roundtrip_for_an_air_bound_only_by_boundary_io() {
...
+        // Rejected while proving, which is the honest outcome for a false statement.
+        return;
+    };
+
+    // Should a proof come out anyway, no verifier may take it.
```
Nashtare, line 2181: At this head this line doesn't hold: a proof that does come out is accepted (see verifier.rs:399). Disabling the position check, the column check or both, and dropping the preprocessed Table batch, each survive every test. Since a missing comparison has already slipped through once, I think the forged-proof test you offered should land with this PR. Sketch: a `#[cfg(test)]` crate-internal hook on the prover's indexed round that swaps one input before `LogupStarProof::prove` (the table column as `[0,1,5,9]`; a row-7 entry 5 mapped to entry 0; the reader claims recomputed from the witness's pulled values), then asserts `verify` returns `Err(IndexedClaimsUnopened)`. One case per comparison, plus one with a preprocessed table, kills each of those mutations. Could you add it?

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1936,3 +1936,250 @@ fn security_checked_roundtrip_for_an_air_bound_only_by_boundary_io() {
...
+    //
+    // Entry 5 has no embedding in this table, so no scatter can place that row.
+    let table_rows = 4usize;
+    let reader_rows = 8usize;
+    let reader_log = log2_strict_usize(reader_rows);
+    let table_log = log2_strict_usize(table_rows);
+
+    let squares = RowMajorMatrix::new((0..table_rows).map(|v| F::from_usize(v * v)).collect(), 1);
+
+    let named = [5usize, 1, 2, 3, 3, 2, 1, 0];
```
Nashtare: Nit: row 0 is also pinned by `when_first_row().assert_zero(local[0])`, so a proof that got past `prove` would fail the zerocheck with or without the position comparison. Like its neighbour, the test pins a panic in `prove`, and `catch_unwind` accepts any panic. Naming entry 5 on a later row (row 7, say) would make the verifier branch depend on the position check.

multi-stark/src/verifier.rs
```diff
@@ -282,6 +292,42 @@ where
         }
     };
 
+    // 6. Verify the indexed reduction against the point the zerocheck bound.
```
Nashtare, line 297: Nit: the step comments are renumbered (6-9 here, 6-8 in `prove`), but the doc lists above `verify` and `prove` went back to #2149's text: no indexed step, "open main tables at r", and `prove` still says "Phases 2 and 4 through 7" run in a bracket. The previous head had both lists updated.

Nashtare (changes requested):
Re-reviewed at ab962b48, on #2149 at 3f9116f2. I ported the forgery harness and the tamper set to this head, re-ran the honest composition shapes, and applied eight mutations.

**Fixed**
- Both earlier forgeries now return `Err(IndexedClaimsUnopened)`: the reduction run on table `[0,1,5,9]` against a committed `[0,1,4,9]`, and entry 5 of a 4-entry table mapped to entry 0 in the witness only. The same holds when the batch also has an 8-entry table, so the shared map does hold entry 5.
- Both sides open every batch at `RunPoints::at(role)`. Roles come from the plan alone, each Position and TableColumns role occurs once, and a preprocessed table is looked up in the preprocessed schedule, so a proof cannot move a claim to another batch or point. A prover that opens everything at `r` fails the round trip. Eleven honest shapes (two readers, shorter reader, 16-row preprocessed table, two tables, crossing heights, self-read, ...) verify.
- The shared map: `entry < entries` is the right bound, entry 4 of a 4-entry table panics the prover, and one allocation follows the widest table.
- `cargo sort --workspace --grouped --check` passes.

**Blocking: reader claims are no longer compared at `r`.** The previous head compared `statement.claims()` with `plan.statement(&reduction.point, &main_openings)`. That check is gone, and nothing else reads `round.reader_claims`. The new checks bind the reduction's output claims, but its input claims are now free: nothing ties a reader's claim to its own payload column.
- Local probe: the reader holds `0 1 5 9 9 5 1 0` against the committed `[0,1,4,9]` with honest positions. The prover takes its claims from the values the positions pull instead of the zerocheck openings. `verify` returns `Ok`. It also returns `Ok` for a payload of all 7s, and for a 16-entry preprocessed table with every held value off by one.
- Putting the old comparison back after `main_openings` rejects all of them and keeps `cargo test -p p3-multi-stark` green.
- The description and `IndexedLookupProof`'s soundness doc still describe that comparison.

**Malformed claim lists still panic `verify`.** `statement_from_claims` returns `ClaimCount`/`ClaimWidth` now, but the `?` skips `transcript.abort()`, so adding or dropping a reader or a column panics in the driver's drop check. With an abort first, all four return `Err`. Every other tamper returns `Err`.

**Nothing pins the comparisons.** Disabling the position check, the column check or both survives every test, and so does dropping the preprocessed Table batch. Dropping the Position or main Table batch is now caught by the round trip. Removing the prover's assertion is caught only by two `logup_star` unit tests that expect its panic message. The missing check at `r` is the same gap. I think the forged-proof test you offered should land with this PR (sketch inline).

**Also**
- The `security_report` test on a batch that declares a read, mentioned on #2149 as landing here, is not in the PR.
- The new out-of-range test names entry 5 on row 0 (inline).

multi-stark/src/verifier.rs
tcoratger: You are right, and thanks for the probe — I replaced the comparison instead of joining it.

Both run after `main_openings` now. First `statement.claims()` against `plan.statement(&reduction.point, &main_openings).claims()`, then a walk over the tables comparing the reduction's position and column claims with the batches answering those roles. Your three cases are rejected with `IndexedClaimsUnopened`.

Tests for all three comparisons landed too — details in the thread on the forged-proof test.

multi-stark/src/verifier.rs
tcoratger: Fixed: `transcript.abort()` before the return, the same shape as the `Err` arm below.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Added, in `multi-stark/src/prover.rs`. `Forgery` is a `#[cfg(test)]` struct threaded through `prove_forged`, which `prove` calls with `None`. It can substitute the positions the reduction reads, a table's column, or the columns a reader's claims are read off. `LogupStarProof::prove_unchecked` skips the prover's own statement assert, as a forging prover would.

One correction to the sketch. A forgery that only swaps the table column never reaches the comparison: the reduction's own claimed-sum check fires first with `IndexedReduction(ProductClaimedSum)`, because the honest claims no longer match `Y · T`. So each fixture is made self-consistent — the committed reader agrees with whatever the reduction was handed — and then only the comparison against the commitment separates the two.

- `a_reduction_run_against_an_uncommitted_table_is_rejected`: table `[0,1,4,9]` committed, `[0,1,5,9]` reduced against, pulls `0 1 5 9 9 5 1 0` committed.
- `a_reduction_run_against_uncommitted_positions_is_rejected`: names `0 1 2 3 3 2 1 0` committed, `0 1 2 3 3 2 1 1` reduced against, pulls `0 1 4 9 9 4 1 1` committed.
- `claims_the_reader_never_committed_to_are_rejected`: the identity table, so the pulls equal the names; the claims are read off the position column while the payload column holds `0 7 7 7 7 7 7 0`.
- `a_reduction_run_against_an_uncommitted_preprocessed_table_is_rejected`: the first case with the table fixed in the verifying key.

Each kills exactly one comparison, which I checked by disabling them one at a time:

```text
    reader-claim comparison off  ->  only `claims_the_reader_never_committed_to` fails
    position comparison off      ->  only `..._uncommitted_positions_...` fails
    column comparison off        ->  the table and preprocessed-table cases fail
```

multi-stark/tests/whir_fibonacci.rs
tcoratger: Moved to row 7, so the names are `0 1 2 3 3 2 1 5`. The comment now says why row 0 is left alone: the AIR already pins it, and the trace would be rejected before the position reached the reduction.

multi-stark/src/verifier.rs
tcoratger: Both lists updated. `verify` lists step 6 as the indexed reduction, 7 and 8 as the openings at every point a claim was left at, and 9 as discharging the indexed claims then closing the zerocheck. `prove` has the matching 6 through 8 and says phases 2 and 4 through 8 run in the bracket.

Also fixed the `# Soundness` bullet that still claimed the point is the bound point.

multi-stark/tests/whir_fibonacci.rs
tcoratger: CI caught one on the preprocessed case: the preprocessed trace is stacked into its own commitment, and a four-entry table left the opening fewer variables than a packed field element has lanes, so `compress_prefix_to_packed` asserted under AVX2 and NEON. It passed locally only because a scalar build has `Packing::WIDTH == 1`.

Fixed by giving that fixture a thirty-two-entry table and deriving the verifier heights from the fixture instead of hard-coding them. Re-ran the suite and the three mutations under `-Ctarget-feature=+avx2` with `--features parallel`, and they still separate as listed above.

multi-stark/src/prover.rs
```diff
@@ -324,9 +458,405 @@ where
...
+        //     reduction reads   0 1 5 9 16 25 ...
+        //     committed pulls   0 1 5 9 9 5 1 0
+        //
+        // The table claims are then discharged against the preprocessed commitment rather
+        // than the main one, and that is the path this pins.
+        //
+        // Thirty-two entries, because the preprocessed commitment is stacked on its own and
+        // a narrower one leaves the opening fewer variables than a packed field element has
+        // lanes.
+        let table = (0..32u64).map(|entry| entry * entry).collect::<Vec<_>>();
```
Nashtare: This clears the bound on NEON (`W = 4`) and AVX2 (`W = 8`), but not on AVX-512. BabyBear packs 16 lanes under `+avx512f`, so `compress_prefix_to_packed` needs `FOLDING + 4 = 6` variables, and 32 entries give the preprocessed commitment 5. A `-C target-cpu=native` test run on a Zen 4/5 or Sapphire Rapids box would panic in `prove` here, by the same assert that fired on NEON at 4 entries (I have no AVX-512 host to run it on). CI can't see it because the avx512f legs are build-only. Could the size come from the packing, e.g. `(1 << FOLDING) * F::Packing::WIDTH` entries, so the fixture clears the bound on every target?

multi-stark/src/prover.rs
```diff
@@ -324,9 +458,405 @@ where
...
+            },
+        );
+        assert!(matches!(
+            verdict,
+            Err(VerificationError::IndexedClaimsUnopened)
+        ));
+    }
+
+    #[test]
+    fn a_reduction_run_against_an_uncommitted_preprocessed_table_is_rejected() {
```
Nashtare, line 916: Nothing proves and verifies an honest batch whose table lives in the preprocessed trace. Locally, dropping the preprocessed `TableColumns` batch from `preprocessed_schedule` keeps all four tests green on scalar x86. A missing batch yields exactly the `IndexedClaimsUnopened` this test expects, even though every honest proof with a key-fixed table would then be rejected. A `verdict(TraceWindow::Preprocessed, ..., &Forgery::default())` case asserting `Ok` catches it. Could that sit next to this one?

multi-stark/src/prover.rs
```diff
@@ -324,9 +458,405 @@ where
...
+    /// proof is the comparison of its claims against the commitment.
+    ///
+    /// # Arguments
+    ///
+    /// - `window`: the trace the table's provider commits its entries to.
+    /// - `table`: the table entries to commit.
+    /// - `named`: the entry each reader row commits to naming.
+    /// - `pulled`: the value each reader row commits to having pulled.
+    /// - `forgery`: what the reduction reads in place of the committed traces.
+    fn verdict(
```
Nashtare, line 713: Every forged case goes through this helper with one reader, one table and one column, and `IndexedWitness::forge` only replaces `[table][0]`. So the tests can't tell a complete comparison from one that stops at the first entry. Locally, each of these keeps all four tests green: `statement.claims().first() != opened.claims().first()`, `.take(1)` on the position loop, and `.take(1)` on the table loop. Under each one, an integration probe that forges only the second reader's claims, only its positions, or only the second table in plan order verifies `Ok`. At head the same probes return `IndexedClaimsUnopened`, so the code is right; only the pinning is missing. Could one case forge the second of two readers, and one the second of two tables?

multi-stark/src/prover.rs
```diff
@@ -139,6 +164,36 @@ where
...
+///
+/// Callers reach this through the entry point above, which substitutes nothing.
+#[allow(clippy::too_many_lines)]
+pub(crate) fn prove_forged<'a, C, A>(
+    config: &C,
+    instances: ProverInstances<'a, C, A>,
+    pow_bits: usize,
+    caller_challenger: &mut C::Challenger,
+    #[cfg(test)] forgery: Option<&Forgery>,
+    #[cfg(not(test))] forgery: Option<core::convert::Infallible>,
```
Nashtare: Nit: release builds still carry the harness here: `prove_forged` takes an `Option<Infallible>`, `prove` passes `None`, and line 196 needs `let _ = &forgery;`. The `#[allow(clippy::too_many_lines)]` above does nothing, since that lint is pedantic and the workspace doesn't enable it, and `prove` never carried it. Could the allow go?

multi-stark/src/prover.rs
```diff
@@ -74,12 +96,15 @@ where
...
 ///                                   -> delegated, bound to the preprocessed commitment
 /// ```
 ///
-/// Phases 2 and 4 through 7 run inside a recorded `Begin`/`End` bracket.
+/// Each table is opened at every point a claim was left at, not at the bound point alone.
+///
+/// Phases 2 and 4 through 8 run inside a recorded `Begin`/`End` bracket.
 /// The pattern player therefore rejects a run that skips one or reorders two.
 ///
 /// Main trace tables are committed together in input-instance order. Each table
```
Nashtare: Nit: the next sentence, "Each table is still opened at the suffix of the common zerocheck point matching that instance's height", is true only for a table's own batch, and it sits right after line 105 saying the opposite. Could it say that a table's own columns open at the suffix, and that its Position and table batches open at the reduction's points?

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1936,3 +1936,253 @@ fn security_checked_roundtrip_for_an_air_bound_only_by_boundary_io() {
...
+    // Fixture state: the four-entry squares table, read by an eight-row AIR.
+    //
+    // Mutation: row 7 names entry 5, one past the table.
+    //
+    //     honest : names 0 1 2 3 3 2 1 0
+    //     forged : names 0 1 2 3 3 2 1 5
+    //
+    // Entry 5 has no embedding in this table, so no scatter can place that row.
+    //
+    // Row 0 is left alone, because the AIR already pins it to zero and would reject the
```
Nashtare: Nit: `prove` never evaluates the AIR constraints, so row 0 isn't rejected "before the position ever reaches the reduction". Entry 5 on row 0 still panics in `IndexedWitness::build`, as on row 7. The difference is on the verifier side: a proof naming it on row 0 would fail the zerocheck close whether or not the position check exists. Could the comment say that instead?

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1936,3 +1936,253 @@ fn security_checked_roundtrip_for_an_air_bound_only_by_boundary_io() {
...
+            // The table's single column carries one value per entry.
+            Self::Table => builder.push_indexed_table("squares", TraceWindow::Main, [0]),
+            // Column 0 names the entry, column 1 holds what that row pulled.
+            Self::Reader => builder.push_indexed_read("squares", 0, [1]),
+        }
+    }
+}
+
+#[test]
+fn prove_verify_indexed_lookup_roundtrips_through_pcs() {
```
Nashtare, line 1984: The `security_report` test on a batch that declares a read, promised for this PR on #2149, still isn't here. No test assesses `SquaresBatch` (or any reading batch) through `security_report`, `prove_with_security` or `verify_with_security`. The p3-security tests call `reduction_terms` with hand-built params, and the indexed.rs tests check `security_params` alone, so nothing covers the wiring at security.rs:377. Could this round trip assert that the report carries the `logup-star-*` terms?

Nashtare (approved):
Re-reviewed. I went through 41e9e5df in full; at 7e2427b7 the only change to this commit is the resized preprocessed fixture, and CI is green.

**Fixed**
- Reader claims are compared at `r` again, next to the position and column comparisons. Every forgery from the last two rounds now returns `Err(IndexedClaimsUnopened)`: wrong pulled values, an all-7s payload, a 16-entry preprocessed table off by one, the uncommitted table and the out-of-range entry. So do new forgeries on a second reader, a second table and a second payload column.
- Every error return after the transcript is created aborts it first. Pushing or clearing a reader or a column in `reader_claims` returns `Err(IndexedLookup(..))`, and nothing panics the verifier.
- The forged-proof tests reach `verify` and match the exact error variant, and `Forgery` and `prove_unchecked` exist only under `cfg(test)`. Disabling the reader-claim, position or column comparison one at a time fails exactly the tests you listed, and a table-only swap does trip `ProductClaimedSum` first.
- The CI failure is gone: 32 entries clear `FOLDING + log2(W)` on NEON and AVX2.

**Comments, none blocking**
- The preprocessed fixture would still panic on an AVX-512 test run (inline).
- The limit underneath is on main: when a preprocessed commitment has fewer than `FOLDING + log2(W)` variables, the prover panics in `compress_prefix_to_packed`. An honest 8-row AIR with one preprocessed column hits it on NEON. That seems worth its own issue, since a small key-fixed table is a normal thing to declare.
- Nothing proves and verifies an honest preprocessed table, and every forged case has one reader, one table and one column. The code handles the wider shapes; only the pinning is missing (inline).
- The `security_report` test on a batch that declares a read is still absent (inline).
- A few doc nits (inline).

multi-stark/src/prover.rs
tcoratger: Good catch, and it was worse than the fixture: this box does have AVX-512, so I could run what you couldn't.

The size now comes from the packing, as you suggested — `(1 << FOLDING) * PackedF::WIDTH` entries, which is 4 on scalar, 32 on AVX2 and 64 under `+avx512f`. Every forged case takes its table from that, so the bound is cleared wherever the tests run.

Running the suite under `-Ctarget-feature=+avx512f,...` then turned up a second one you hadn't seen: `prove_verify_indexed_lookup_roundtrips_through_pcs` panicked too, because its stacked commitment has 5 variables and needs `FOLDING + 4 = 6`. That fixture now scales the same way, keeping its read pattern (each entry named once up and once down):

```text
    scalar   4 entries, 8 reader rows
    avx2     8 entries, 16 reader rows
    avx512  16 entries, 32 reader rows
```

Separately, and not from this PR: `security_checked_whir_roundtrip_and_target_rejection` and `security_checked_lookup_accounts_for_every_reduction` fail the same assert under AVX-512, and they fail on `main` at `ce322afe` as well. I left them alone rather than widen this PR. Happy to send a follow-up sizing their fixtures the same way if you'd like.

multi-stark/src/prover.rs
tcoratger: Added, as `an_honest_batch_reading_a_preprocessed_table_verifies`: the reader pulls what the key holds, nothing is substituted, and it asserts the proof verifies.

It also goes through `prove` rather than the forging path, since the harness now takes `Option<&Forgery>` and this case passes `None`.

Confirmed it catches your probe: flipping the `TraceWindow::Preprocessed` guard in `preprocessed_schedule` so the batch is never pushed fails this test and only this test.

multi-stark/src/prover.rs
tcoratger: You're right, and the harness couldn't express it. `Forgery` now names the one entry it moves rather than a list zipped over everything:

```rust
positions: Option<(usize, usize, Vec<usize>)>,   // table, reader, rows
columns:   Option<(usize, Vec<u64>)>,            // table in plan order, entries
claims:    Option<(usize, Vec<usize>)>,          // reader in plan order, columns
```

and the fixture is a list of tables, each with its own readers, so a case can put two readers on one table or two tables in one batch.

- the column case now commits two tables and forges the second in plan order.
- the position case puts two readers on one table and forges the second.
- the claims case does the same, and reads the second reader's claims off its position column.

All six mutations are now caught, one case each:

```text
    reader-claim comparison removed            -> claims_the_reader_never_committed_to
    reader-claim comparison .first()           -> claims_the_reader_never_committed_to
    position comparison removed                -> ..._uncommitted_positions_...
    position loop .take(1)                     -> ..._uncommitted_positions_...
    column comparison removed                  -> ..._uncommitted_table_..., ..._preprocessed_...
    table loop .take(1)                        -> ..._uncommitted_table_...
```

multi-stark/src/prover.rs
tcoratger: Dropped. The `let _ = &forgery;` is gone too — the non-test parameter is now named `_forgery` under its own `cfg`, so there is nothing left to discard.

multi-stark/src/prover.rs
tcoratger: Reworded to say what you describe:

```text
    A table's own columns open at the suffix of the zerocheck point matching its height.

    The batches an indexed reduction adds open at the points that reduction closes on.
```

multi-stark/tests/whir_fibonacci.rs
tcoratger: Right, thanks — `prove` never evaluates the constraints, so the old wording was wrong about where it is caught. The comment now says the panic happens while the witness is built either way, and that the difference is verifier-side: a proof naming entry 5 on row 0 fails the zerocheck close on its own, so the branch would not depend on the position comparison.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Added to the round trip, through `security_report` on the same verifier instances.

One thing the assertion turned up: with one reader pulling one column, `logup-star-reader-batching` and `logup-star-column-batching` are absent, because the report skips a degree-0 term and batching one of something is free. So the batch is charged exactly:

```text
    logup-star-entry-challenge
    logup-star-claim-point
    logup-star-fractional-gkr
    logup-star-product-sumcheck
```

which the test pins as an exact list, along with every term costing more than zero bits.

A second test, `a_batch_declaring_no_read_is_charged_for_no_reduction`, runs the same entry point on a batch whose AIR declares no read and asserts no `logup-star-` term appears.

Checked the wiring is what is pinned: forcing `logup_star` to `None` at security.rs:377 fails the round trip and leaves the negative case passing.

PR #2151, opened 2026-09-16
perf(blake3): hash a batch of messages through one SIMD compression

tcoratger:
`Blake3` never offered the batched path, so every message ran through its own `blake3::Hasher`. Digests are unchanged, byte for byte.

## The problem

`p3-blake3` is a 34-line wrapper that takes the trait default `LANES = 1` and does not implement `hash_many`.

That leaves the one thing BLAKE3 is built for on the floor. Its compression advances **4 (SSE/NEON), 8 (AVX2) or 16 (AVX-512) independent chunk states per call**.

And the consumer is already there: `p3-merkle-tree` reads `H::LANES`, sizes its groups from it (`rows_per_call`), and routes whole runs of rows into `hash_many` / `compress_many`. `p3-keccak` supplies that; `p3-blake3` did not.

## The idea

A message of at most `CHUNK_LEN` (1024) bytes is a **single BLAKE3 chunk** — a plain chain of 64-byte compressions:

```text
    IV -> block 0 -> block 1 -> ... -> block k        counter 0 throughout
                                           |          CHUNK_START on block 0
                                           v          CHUNK_END | ROOT on block k
                                        digest
```

Every lane of a batch has that same shape, so the whole-block prefix of a lane group is **one** `Platform::hash_many` call. Only a short final block costs one scalar compression per message.

A message longer than one chunk folds chunk chaining values into a tree, which the flat batch layout cannot express. Those fall back to the scalar hasher — which already drives SIMD across the chunks of a single message.

`CompressionFunctionFromHasher` forwards both `LANES` and `compress_many` to the hasher, so the 2-to-1 Merkle compression picks this up with no separate impl.

## The digest does not change

This is the part worth checking, since the digest is a commitment.

The `ROOT` flag rides on the last compression of the chunk, and the 32-byte chaining value of that compression is exactly what `Output::root_hash` returns — the root block is compressed at counter zero either way.

Pinned by test at every message length in `{0, 1, 31, 32, 63, 64, 65, 1023, 1024, 1025, 2048}`, at every batch count in `1..=2*LANES+1`, against **both** the scalar path and `blake3::hash` directly. Plus a proptest over random lengths and counts.

Separately, before opening this I ran an exhaustive differential sweep — every length `0..=1100` against eight batch counts, each lane compared to `blake3::hash` — which passed. That sweep is not committed; the shape-based tests above cover the same boundaries at a cost the suite should carry.

## The unstable dependency, stated plainly

`hash_many` lives in `blake3::platform`, which is `#[doc(hidden)]` and labelled unstable.

The stable `hazmat` module cannot replace it, and this is not a matter of taste:

- hazmat batches the chunks **inside one message**, keyed by each chunk's input offset (`set_input_offset`).
- Independent messages all begin at offset 0, so no hazmat call can ever hold more than one of them.
- Concatenating padded messages into one `update` gives chunk `i` the counter `i`, not 0.
- `finalize_non_root` per message is the scalar loop we already have, and the deprecated `guts` module is scalar too.

So this is the only batched entry point in the pinned version. The risk is bounded in the way that matters: an API **removal fails the build**, and a semantic **change fails the digest-equality tests** — neither can silently alter a commitment. A comment above the import says all of this.

The BLAKE3 IV and the `CHUNK_START` / `CHUNK_END` / `ROOT` flag bytes are spelled out locally because the crate keeps them private. They are spec constants, not crate API, so they add no maintenance surface.

## Benchmark

`RUSTFLAGS="-C target-cpu=native"`, AMD Zen 5, blake3 1.8.7, AVX-512 detected so `LANES = 16`. `cargo bench -p p3-blake3 --bench blake3`, 4096 messages per batch. Runs were serialized against other work on the machine.

| message | one at a time | `hash_many` | speedup | throughput |
|---|---:|---:|---:|---|
| 32 B | 334.4 µs | 226.8 µs | 1.47x | 374 MiB/s → 551 MiB/s |
| 64 B | 332.2 µs | 21.21 µs | **15.7x** | 752 MiB/s → 11.5 GiB/s |
| 256 B | 1.076 ms | 74.16 µs | **14.5x** | 929 MiB/s → 13.2 GiB/s |
| 1024 B | 4.211 ms | 295.4 µs | **14.3x** | 950 MiB/s → 13.2 GiB/s |

The 32-byte row is the honest exception: below one block there is no whole-block prefix, so it gains only the removal of `Hasher` setup. There is no batched entry point for a partial block — the compression mixes in the true block length, which the vectorized call always sets to 64. It costs nothing and falls out of the same code path.

Every case from one block upward is 14–16x, which is every 2-to-1 Merkle compression and every leaf row of 16 or more `BabyBear` columns.

End to end, a Merkle tree over 2^20 leaves with `SerializingHasher<Blake3>` and `CompressionFunctionFromHasher<Blake3, 2, 32>` over `BabyBear`:

| row | `--features parallel` (32 threads) | single-threaded |
|---|---|---|
| 32 B | 11.1 → 5.9 ms (1.88x) | 215.0 → 66.1 ms (3.25x) |
| 64 B | 12.5 → 5.3 ms (2.36x) | 217.8 → 26.5 ms (**8.22x**) |
| 256 B | 21.6 → 11.5 ms (1.88x) | 462.3 → 47.2 ms (**9.80x**) |

The parallel column is lower because at 32 threads a 2^20-leaf build is memory-bound rather than compression-bound. The serial column is the honest measure of the hashing work itself.

## Scope

- **New:** `hash_many` on `Blake3`, `LANES = MAX_SIMD_DEGREE`, and a `blake3/benches/blake3.rs` criterion bench.
- **Unchanged:** `hash_iter`, `hash_iter_slices`, every digest, the crate's `no_std` status, and any `unsafe` (none is added).
- **Not done:** a separate `PseudoCompressionFunction` impl, which `CompressionFunctionFromHasher` already makes unnecessary.

## Verification

- `cargo test -p p3-blake3` — 7 passed.
- `cargo test -p p3-merkle-tree` — 114 passed.
- `cargo clippy -p p3-blake3 --all-targets` — clean. `cargo fmt` applied. `cargo doc` clean.


tcoratger:
One extra check worth recording, since the new dependency is on a `#[doc(hidden)]` module and `p3-blake3` sits in the embedded CI leg.

`blake3::platform` compiles everywhere the workspace targets — `Platform::detect()` falls back to `Portable` off x86 and ARM, and `MAX_SIMD_DEGREE` is defined on every arch, so `LANES` degrades to 1 rather than failing to build:

| target | result |
|---|---|
| `x86_64-unknown-linux-gnu` | builds, tests pass, `LANES = 16` |
| `thumbv7em-none-eabi` (the embedded leg) | `cargo build -p p3-blake3 --lib` clean |
| `wasm32-unknown-unknown` | `cargo build -p p3-blake3 --lib` clean |
| `aarch64-unknown-linux-gnu` | not reproducible locally — `blake3`'s own build script needs `aarch64-linux-gnu-gcc`, which this machine lacks. Pre-existing and unrelated to the diff; CI supplies the cross toolchain. |

The crate stays `#![no_std]`; nothing added needs `std`.

blake3/src/lib.rs
```diff
@@ -2,13 +2,162 @@
...
+//
+// Depending on it is a deliberate risk.
+//
+// Why not the stable `hazmat` module:
+//     hazmat batches the chunks *inside* one message, keyed by each chunk's input offset.
+//     Independent messages all start at offset 0, so no hazmat call can ever hold more than one.
+//
+// The digest-equality tests are what keep this dependency honest.
+// They pin every batched digest to `blake3::hash` and to the scalar path, at every message shape.
+use blake3::platform::{MAX_SIMD_DEGREE, Platform};
```
Nashtare, line 21: `blake3` labels this module (and `IncrementCounter`) "Undocumented and unstable, for benchmarks only". The workspace takes it as a caret `1.8.4` and no lockfile is committed. So the digest-equality tests guard Plonky3's CI, not downstream builds: a 1.x release that reshapes `Platform::hash_many` breaks every published `p3-blake3` on `cargo update`, and `platform.rs` already changed in 1.8.7 (the `arrayref` removal). Could the comment say that plainly? And is the caret deliberate, over an exact pin (which conflicts with graphs needing a newer `blake3`) or in-crate lanes as in `p3-keccak`?

blake3/src/lib.rs
```diff
@@ -2,13 +2,162 @@
...
 #[derive(Copy, Clone, Debug)]
 pub struct Blake3;
 
 impl CryptographicHasher<u8, [u8; 32]> for Blake3 {
+    /// Independent chunk states the running target's widest compression advances per call.
+    ///
+    /// The `blake3` crate sizes this by what its own build compiled in:
+    ///
+    /// ```text
+    ///     AVX-512  16      SSE, NEON   4
```
Nashtare: This doesn't match `blake3` 1.8.7 on x86. `MAX_SIMD_DEGREE` is 16 whenever the build's C compiler accepts `-mavx512f -mavx512vl` (its `build.rs`) and 8 otherwise, never 4, whatever the CPU. Local probe under Rosetta: `MAX_SIMD_DEGREE = 16`, `Platform::detect() = SSE41` (degree 4), `LANES = 16`. On wasm32 it is 4 only with the `wasm32-simd` feature. So line 148's "running target's widest compression" is really the compiled-in maximum.

blake3/src/lib.rs
```diff
@@ -2,13 +2,162 @@
...
+    ///     AVX2      8      portable    1
+    /// ```
+    ///
+    /// Reporting that compile-time maximum is safe for a caller grouping messages.
+    /// A narrower CPU detected at run time simply splits the group into several compressions.
+    const LANES: usize = MAX_SIMD_DEGREE;
```
Nashtare, line 159: With 16 lanes, leaf rows of 513..1024 bytes never fill a lane group: `SerializingHasher::hash_many` re-splits the Merkle tree's 16-row group under its 8 KiB budget (`symmetric/src/serializing_hasher.rs:14`, whose comment assumes eight 1 KiB messages fill the lanes). Local probe, Merkle commit on an x86_64 build: 540 B rows arrive as calls of 15 and 1, 800 B as 10 and 6, 1024 B as 8. Through the C cascade that is 25% scalar rows and no 16-wide call on AVX-512. With the budget at 16 KiB every call is 16. Could the budget follow `LANES`, or could this get numbers on the 135-column merkle bench?

blake3/src/lib.rs
```diff
@@ -31,4 +180,222 @@ impl CryptographicHasher<u8, [u8; 32]> for Blake3 {
...
+    #[test]
+    fn hash_many_matches_scalar_across_block_shapes() {
+        // Batch sizes below, at, and above one full lane group.
+        // The final short group is then exercised at every lane count the target compiles to.
+        let counts: Vec<usize> = (1..=2 * MAX_SIMD_DEGREE + 1).collect();
+
+        for len in SHAPE_LENGTHS {
+            for &count in &counts {
+                // Fixture: a deterministic byte ramp, distinct per position.
+                let messages: Vec<u8> = (0..len * count).map(|i| (i * 31 + 7) as u8).collect();
```
Nashtare: Nit: `(i * 31 + 7) as u8` has period 256, so at `len` 1024 and 2048 every message in the batch is byte-identical, and at 64 the lanes repeat every 4. A lane-order bug can't show at those lengths here (a lane-reversal mutant is still caught at the odd lengths and by the proptest). Could the fixture mix in the message index, e.g. the proptest's xorshift?

blake3/benches/blake3.rs
```diff
@@ -0,0 +1,61 @@
...
+const MESSAGES: usize = 4096;
+
+/// Message lengths that cover every shape the batched path distinguishes.
+///
+/// 32 is a digest, below one block and therefore one compression per message.
+/// 64 is a two-to-one Merkle compression, exactly one block.
+///
+/// 256 is a wide leaf row, four whole blocks.
+/// 1024 is a full chunk, the largest message the batched path accepts.
+const LENGTHS: [usize; 4] = [32, 64, 256, 1024];
```
Nashtare: Nit: none of these has whole blocks plus a tail, which costs one scalar compression per message, and none is above `CHUNK_LEN`. Could 540 (the 135-column row `merkle-tree/benches/merkle_tree.rs` uses) and 1025 join, so the bench covers every shape the doc claims?

blake3/Cargo.toml
```diff
@@ -20,5 +20,13 @@ p3-util.workspace = true
 
 blake3.workspace = true
 
+[dev-dependencies]
+criterion.workspace = true
```
Nashtare: Nit: `criterion` pulls in rayon, so `cargo test --target wasm32-wasip1 -p p3-blake3 --features wasm32-simd` no longer builds ("Rayon cannot be used when targeting wasi32"). `p3-keccak` and `p3-goldilocks` put it under `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]`; with that, the new tests pass under wasmtime on `WASM32_SIMD` locally.

Nashtare (commented):
Reviewed at 78013903. I found no correctness issue: the batched path produces root BLAKE3 at every length class. Questions below on the dependency and x86 grouping.

**Checked locally**
- **Spec walk.** I walked each length class against `blake3` 1.8.7 (`ChunkState::update`/`output`, `Output::root_hash`, `portable::hash1`):
  - the counter is 0 throughout;
  - `block_len` is 64 on prefix blocks and `tail` on the last;
  - `CHUNK_START` is set only on block 0, and `CHUNK_END | ROOT` only on the last;
  - a full final block stays in the prefix, 1024 is a single chunk, and 1025+ falls back.

  The IV and flag bytes match the crate's private constants.
- **Differential sweep.** Every length 0..=1100 at counts 1..=2L+1, plus a few larger counts, with each lane checked against `blake3::hash`. It passes on NEON (aarch64, L=4), on SSE4.1 (x86_64 under Rosetta, L=16), and on `WASM32_SIMD` and portable under wasmtime.
- **Merkle roots.** With `SerializingHasher<Blake3>` + `CompressionFunctionFromHasher<Blake3, 2, 32>`, the roots equal a `LANES = 1` wrapper's across 44 shapes (1..300 columns, injected matrices). They match serially, under `parallel`, and on x86_64.
- **Mutations** on `hash_many`: 20 of 21 killed (flags, counter, `IncrementCounter`, `block_len`, padding, IV, stride, fallback threshold, idle lanes, lane order). The survivor swaps in `Platform::portable()`, which gives the same digests.
- **Also clean:** tests for `p3-blake3` and `p3-merkle-tree` (including `parallel` and Rosetta), clippy on both arches, nightly fmt, sort, taplo, machete, docs, and the thumbv7em and wasm32 builds.

**Questions**
1. **The hidden dependency.** `blake3::platform` is imported under a caret requirement (inline at the import).
2. **x86 lane count.** `LANES` is 16 on every x86 build whose C compiler accepts AVX-512 flags, whatever the CPU. Under Rosetta, `MAX_SIMD_DEGREE = 16` while `Platform::detect() = SSE41`. As a result the doc table is off, and leaf rows of 513..1024 bytes never fill a 16-lane call, because `SerializingHasher` re-splits under its 8 KiB budget (both inline). The in-repo merkle bench, at 135 columns = 540 B, is in that range.
3. **The description.** "Every leaf row of 16 or more `BabyBear` columns" at 14–16x holds only for rows that are whole blocks: other widths pay one scalar tail compression per row. And "AVX-512 detected so `LANES = 16`" should read "compiled in".

blake3/src/lib.rs
tcoratger: Fixed in `99fbd763`. You are right and the table was wrong in both directions.

Re-read from `blake3` 1.8.7's `src/platform.rs`: `MAX_SIMD_DEGREE` is chosen purely on `cfg` — 16 under `blake3_avx512_ffi`, 8 otherwise on x86; 4 under `blake3_neon`; 4 under `blake3_wasm32_simd`; 1 otherwise. And `blake3_avx512_ffi` is emitted by its `build.rs` when the C toolchain accepts `-mavx512f`/`-mavx512vl`, never by the CPU. The SSE-4.1 and SSE-2 kernels exist as `Platform` variants but never as a value of this constant, so mapping SSE to 4 was wrong too.

The doc now states the compile-time rule and separates it from `Platform::detect`.

blake3/src/lib.rs
tcoratger: Good catch, and it was a real bug rather than a tuning question.

`SerializingHasher` computed `rows_fitting / lanes`, rounding **down** — so with `lanes = 16` and any row over 512 bytes that is zero, and it fell through to "as many rows as fit the budget", re-splitting the lane group `p3-merkle-tree` had just built. `merkle_tree.rs`'s own `rows_per_call` already had the right rule, so `SerializingHasher` now uses the identical expression `(ROW_BYTES_SCRATCH / row_bytes / lanes).max(1) * lanes`: the budget is a target, not a cap.

Measured on the bench you named, 2^15 rows x 135 BabyBear columns (540 B/row), `target-cpu=native`:

| | before | after | |
|---|---:|---:|---:|
| `MerkleTree` | 9.14 ms | 5.09 ms | **1.80x** |
| `MerkleTreeMmcs` | 13.94 ms | 8.10 ms | **1.72x** |

Blast radius: the formulas differ only when `lanes * row_bytes > 8 KiB`, so `p3-keccak` at `VECTOR_LEN = 2` is bit-identical in scheduling at this shape. The one test asserting the old split is updated to assert the filled group.

blake3/Cargo.toml
tcoratger: Fixed — moved `criterion` under `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]`, matching `p3-keccak` and `p3-goldilocks`.

Verified both directions: with the old manifest `cargo test --no-run --target wasm32-wasip1 -p p3-blake3 --features wasm32-simd` fails on criterion's `compile_error!`; with the fix it builds. No `wasmtime` on this machine, so I stopped at `--no-run`.

blake3/src/lib.rs
tcoratger: Fixed — the fixture now uses the same xorshift64 stream the proptest uses, so no two messages in a batch repeat at any length.

blake3/benches/blake3.rs
tcoratger: Added both. `LENGTHS` is now `[32, 64, 256, 540, 1024, 1025]`, and 540 earns its place — eight whole blocks plus a 28-byte tail:

| | one at a time | `hash_many` | |
|---|---:|---:|---:|
| 32 B | 324.1 us | 194.9 us | 1.66x |
| 64 B | 330.3 us | 22.0 us | 15.01x |
| 256 B | 1.111 ms | 77.0 us | 14.43x |
| **540 B** | 2.505 ms | 463.0 us | **5.41x** |
| 1024 B | 4.557 ms | 315.2 us | 14.46x |
| 1025 B | 5.071 ms | 4.951 ms | 1.02x |

So the tail costs about two thirds of the win at that shape, and 1025 confirms the tree fallback is free. This also corrects the description's "every leaf row of 16 or more BabyBear columns", which held only for block-aligned rows.

blake3/src/lib.rs
tcoratger: Reworded to say it plainly: the digest-equality tests guard this workspace's CI only. `blake3` is a caret `1.8.4`, `Cargo.lock` is gitignored, so a 1.x release reshaping `Platform::hash_many` breaks every already-published `p3-blake3` on `cargo update`, with a patch release the only way out — and `platform.rs` already changed once in 1.8.7.

On your two questions:

**The caret** predates this PR and I deliberately left it. An exact `=1.8.7` would make the guarantee real for downstreams, but a `=` requirement in a published library is contagious and would turn a soft break into an unresolvable one for any dependent needing a newer `blake3` — the graph conflict you mention. Committing `Cargo.lock` fixes CI reproducibility but does nothing for dependents, since a library lockfile is ignored. Only removing the unstable surface actually closes it, and that is a different PR. Happy to take whichever you prefer.

**In-crate lanes, as `p3-keccak` does.** Considered and rejected, for a reason that differs from keccak's: `tiny-keccak` exposes no batched entry point at all, so there was nothing to call. BLAKE3 is the opposite — the fast kernels are hand-written AVX-512/AVX2/NEON assembly shipped inside the crate. Reimplementing them here is large, high-risk work for a hash whose value is that upstream already did it, and a portable-SIMD version would land well short of the 12.7 GiB/s measured. Worth noting the exposure is narrow (`Platform`, `hash_many`, `compress_in_place`, `IncrementCounter`, `MAX_SIMD_DEGREE`) and a break is a compile error, not a wrong digest.

Nashtare (approved):
Approving 99fbd763. I benchmarked 78013903 on Graviton5 (`c9g.8xlarge`, Neoverse-V3, `-C target-cpu=native`) against main 463b44c3.
- Each comparison ran twice, alternating base and head.
- Serial runs were pinned to one core, and parallel runs to 32 cores.
- Numbers are criterion medians averaged over both rounds; the two rounds agree within 1% serial and 3% parallel.

| bench | main | this PR | speedup |
|---|---:|---:|---:|
| `hash_many`, 4096 × 64 B | 308.8 µs | 148.2 µs | 2.08x |
| `hash_many`, 4096 × 256 B | 1.061 ms | 571.2 µs | 1.86x |
| `hash_many`, 4096 × 1024 B | 4.067 ms | 2.259 ms | 1.80x |
| `hash_many`, 4096 × 1025 B (scalar fallback) | 4.624 ms | 4.630 ms | 1.00x |
| Merkle commit, 2^18 rows of 32 B, serial | 68.6 ms | 28.9 ms | 2.38x |
| Merkle commit, 2^18 rows of 64 B, serial | 78.1 ms | 22.2 ms | 3.52x |
| Merkle commit, 2^18 rows of 540 B, serial | 351.8 ms | 115.6 ms | 3.04x |
| Merkle commit, 2^18 rows of 1024 B, serial | 609.6 ms | 183.3 ms | 3.33x |
| Merkle commit, 2^18 rows of 64 B, 32 threads | 4.40 ms | 1.86 ms | 2.37x |
| Merkle commit, 2^18 rows of 1024 B, 32 threads | 34.3 ms | 19.7 ms | 1.74x |
| in-repo `MerkleTreeMmcs` Blake3 bench, serial | 72.2 ms | 21.4 ms | 3.37x |
| in-repo `MerkleTreeMmcs` Blake3 bench, 32 threads | 3.98 ms | 2.10 ms | 1.90x |

**Nothing regresses.**
- Small batches pay nothing: 4096 messages of 64 B hashed in calls of 1, 2 or 3 take 10–17% less time than on main, and calls of 4 or more get the full 2x.
- Messages above one chunk are unchanged.

**The re-push.** 99fbd763 changes docs, tests and `SerializingHasher`'s group size. The new `(8 KiB / row_bytes / lanes).max(1) * lanes` equals the old rounding whenever at least one lane group fits the target. It differs only for rows over `8 KiB / LANES`: on NEON, rows over 2 KiB for Blake3 and over 4 KiB for Keccak. So every number above carries over, and wide rows now fill their lanes instead of stranding the tail on the scalar path.
- `p3-symmetric`, `p3-blake3` and `p3-merkle-tree` pass at 99fbd763, with and without `parallel`.
- Before merge: `Formatting and Clippy` fails at this head on `cargo sort` ("Dependencies for blake3 are not sorted"). The new `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]` section needs to sit where `cargo sort --grouped` expects it.

Thanks for the clear comment on the `blake3::platform` risk.

PR #2152, opened 2026-09-16
perf(sha256): hash four messages at once with x86-64 SHA-NI

tcoratger:
`p3-sha256` had a four-lane batched path for wasm32 and nothing for x86-64. Digests are unchanged, bit for bit.

## The problem

`sha256/src/wasm32_simd128.rs` implements `LANES = 4`, `hash_many` and `compress_many`. On every other target `LANES` stays at the trait default of 1.

`p3-merkle-tree` reads `H::LANES`, sizes its groups from it (`rows_per_call`), and routes whole runs of rows through `hash_many` / `compress_many`. So on x86-64 a SHA-256 Merkle tree hashed one message per call — with the hardware SHA extension present and running a single stream.

## The idea

`sha256rnds2` advances one 128-bit state. It has a **four-cycle latency for a one-cycle issue rate**, so a lone message is a pure dependency chain that leaves three of every four cycles unused.

```text
    stream 0:  abef[0]  cdgh[0]  w[0][0..4]
    stream 1:  abef[1]  cdgh[1]  w[1][0..4]
    stream 2:  ...                              four independent chains,
    stream 3:  ...                              interleaved into the same pipeline
```

Four independent streams fill those cycles. Per four rounds each stream issues two `sha256rnds2` and one `sha256msg1`/`sha256msg2` pair; the four streams interleave.

The new `sha256/src/x86_64_sha_ni.rs` mirrors the wasm32 module exactly — same `hash_many` contract, same padding handling, same scalar fallback for the short tail after the last full group of four.

## Gating

`all(target_arch = "x86_64", target_feature = "sha", target_feature = "sse4.1")`.

`sse4.1` is the precise requirement, not a rounding-up: `sha256rnds2` / `msg1` / `msg2` need `sha`, `_mm_alignr_epi8` needs `ssse3`, and `_mm_blend_epi16` needs `sse4.1`, which implies the others in rustc's feature graph.

## Correctness

Digests are bit-identical to the scalar `sha2` path — this is a commitment, so it was checked rather than assumed.

In-tree: shape tests over message lengths `{0, 1, 55, 56, 63, 64, 65, 119, 120, 127, 128, 200}` × batch counts `1..=2*LANES+1`, a `compress_many` batch-count sweep, a lane-independence test, and two proptests. The length set covers both padding shapes — 55 fits the `0x80` mark and the 64-bit counter in one block, 56 does not — and every partial final group.

Before opening this I also ran, and then deleted, two extra sweeps:

- `hash_many` at **every** length `0..=260` × batch counts `{1,2,3,4,5,7,8,9,13,16,17}`, each lane compared to `sha2::Sha256::digest`.
- `compress_many` at every batch count `1..=40`, each group compared to `Sha256Compress::compress`.

Both passed. As a negative control I flipped one round constant (`0xc67178f2` → `0xc67178f3`) and confirmed both sweeps fail, so they are not vacuous. I also asserted `LANES == 4` inside the sweep, so it cannot silently have exercised the scalar path.

## Benchmark

Ryzen 9 9950X3D, `RUSTFLAGS="-C target-cpu=native"`, pinned to one core, run under an exclusive machine lock. **Before is the same bench binary built with `-C target-feature=-sha`**, which removes the module and returns `LANES` to 1 while leaving `sha2`'s own runtime SHA-NI dispatch intact — so this is four-way SHA-NI against one-way SHA-NI, not against a software fallback. Batch of 4096 messages.

| bench | before | after | speedup |
|---|---:|---:|---:|
| `hash_many/batched/32` | 181.89 µs | 67.87 µs | **2.68x** |
| `hash_many/batched/64` | 241.08 µs | 115.71 µs | **2.08x** |
| `hash_many/batched/256` | 541.87 µs | 280.20 µs | **1.93x** |
| `compress/batched` | 105.53 µs | 54.85 µs | **1.92x** |

The `one at a time` arms move under 5% between the two builds, which confirms the scalar side is untouched.

End to end, a Merkle tree over 2^20 leaves of 32 bytes with `SerializingHasher<Sha256>` and `Sha256Compress`:

| | before | after | speedup |
|---|---:|---:|---:|
| single-threaded | 96.8 ms | 40.3 ms | **2.40x** |
| `--features parallel`, 32 threads | 7.62 ms | 5.13 ms | 1.49x |

The parallel figure is the weaker one and is stated as measured: at 32 threads a 2^20-leaf commit is bound by memory rather than by hashing, so a 2.4x kernel win compresses to ~1.45x. The single-threaded column is the honest measure of the hashing work itself.

## Why four lanes and not eight

The kernel is generic in its lane count, so this was swept rather than assumed. Batched medians, same lock and pinning:

| bench | 2 lanes | 4 lanes | 8 lanes |
|---|---:|---:|---:|
| 32 B | 111.45 µs | 62.64 µs | 55.76 µs |
| 64 B | 188.88 µs | 114.58 µs | 105.22 µs |
| 256 B | 420.86 µs | 277.46 µs | 252.82 µs |
| compress | 80.61 µs | 55.58 µs | 51.02 µs |

Two lanes is well short of the knee. Eight buys a further **8–11%** on this machine, and the Merkle probe agrees (5.33 ms against 5.46 ms median).

Four ships anyway. Four is the smallest count that covers the four-cycle latency, it is what the existing wasm32 backend uses, and eight doubles the scalar-fallback tail for small batches. The extra 8–11% measured on one Zen 5 box is not enough to justify a less portable constant — but the sweep is recorded here for whoever wants to revisit it on Intel silicon.

Register pressure is why this is ~2x rather than ~4x: four streams want 4 `abef` + 4 `cdgh` + 16 schedule vectors = 24 xmm against 16 architectural registers, so LLVM spills. No hand-scheduling was attempted.

## Scope

- **New:** `sha256/src/x86_64_sha_ni.rs`, and a `sha256/benches/sha256.rs` criterion bench.
- **Changed:** `sha256/src/lib.rs` gains the SHA-NI arm beside the wasm32 one for `LANES`, `hash_many` and `compress_many` on both traits. `reports_four_lanes_only_for_wasm_simd128` is renamed to `..._on_vectorized_targets` and taught about the new backend.
- **Unchanged:** `hash_iter`, `hash_iter_slices`, every digest, and the crate's `no_std` status.

## Verification

- `cargo test -p p3-sha256` with `-C target-cpu=native` — 15 passed, including the 5 new ones.
- `cargo test -p p3-sha256` with `-C target-cpu=x86-64-v3` (backend cfg'd out) — 10 passed, scalar path intact.
- `cargo test -p p3-merkle-tree`, plain and `--features parallel` — 114 passed each.
- `cargo build --target thumbv7em-none-eabi -p p3-sha256 --lib` and the same for `wasm32-unknown-unknown` — clean, so the embedded and wasm CI legs are covered.
- `cargo clippy -p p3-sha256 --all-targets` clean; `cargo +nightly fmt --check` clean; `cargo doc` with `-D warnings` clean; `cargo sort -c sha256` clean.


tcoratger:
@Nashtare I think we should be able to do the same for Mac architecture but I'm not running on Mac right now I can't really test this out for now

sha256/src/lib.rs
```diff
@@ -11,6 +11,13 @@ use sha2::Digest;
 #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
 mod wasm32_simd128;
 
+#[cfg(all(
+    target_arch = "x86_64",
+    target_feature = "sha",
```
Nashtare, line 24: No CI leg compiles this. `rustc --print cfg --target x86_64-unknown-linux-gnu` shows no `sha` for any `features` value in `ci.yml` (`+avx2`, `+avx512f`, `+pclmulqdq`, `+avx2,+vpclmulqdq`, `+avx512f,+vpclmulqdq`), nor for `target-cpu=x86-64-v3`/`v4`. `macos-latest` is arm64 and Windows stops at `sse3`. So this module, its five tests and the `4` arm of the lane test never build. Could a `linux, sha, build only` leg (`+sha,+sse4.1`) go in? Ideally also a test step guarded on `sha_ni` in `/proc/cpuinfo`, since `+sha` also makes `sha2` emit SHA-NI unconditionally.

sha256/src/x86_64_sha_ni.rs
```diff
@@ -0,0 +1,526 @@
...
+/// - `group` selects the matching four round constants.
+#[inline]
+fn round_group(abef: &mut __m128i, cdgh: &mut __m128i, schedule: __m128i, group: usize) {
+    debug_assert!(group < ROUND_GROUPS);
+
+    // SAFETY: the module compiles only when `sha` is enabled, which the round instruction needs.
+    //
+    // The addition and the shuffle are `sse2`, always present on x86-64.
+    //
+    // The load reads four round constants, held in bounds by the assertion above.
```
Nashtare: The assertion this relies on is the `debug_assert!` above, which release builds drop, so inside this safe fn nothing bounds `group` before the 16-byte read from `K`. The callers only pass `0..16`, so it isn't reachable today. Would indexing a chunked view (`K.as_chunks::<4>().0[group]`) or a plain `assert!` keep the bound in release? It should fold away once inlined into those loops.

sha256/src/x86_64_sha_ni.rs
```diff
@@ -0,0 +1,526 @@
...
+        }
+    }
+}
+
+/// Hash four messages of a common length, padding included.
+///
+/// # Panics
+///
+/// Panics in debug builds if `messages` does not hold exactly four messages of `len` bytes.
+fn hash_four(messages: &[u8], len: usize) -> [[u8; 32]; LANES] {
```
Nashtare: Diffing against `wasm32_simd128.rs`: `hash_four`, `hash_many`, `compress_many` and `K` are identical apart from `compress_blocks` → `compress4`, `56` → `BLOCK_BYTES - 8`, and one comment. Could the padding and grouping driver live once in `lib.rs`, over a small four-lane `initial_state`/`compress`/`write_digests` interface, so a padding fix can't land in only one backend?

sha256/src/x86_64_sha_ni.rs
```diff
@@ -0,0 +1,526 @@
...
+    let streams = out.iter_mut().zip(&state.abef).zip(&state.cdgh);
+    for ((digest, &packed_abef), &packed_cdgh) in streams {
+        // SAFETY: the blend needs `sse4.1` and the align needs `ssse3`.
+        //
+        // Both are guaranteed by the module's `sse4.1` gate.
+        //
+        // The two stores write sixteen bytes each into the halves of a 32-byte array.
+        unsafe {
+            // Undo the SHA-NI packing: `abef` becomes `a b e f` and `cdgh` becomes `g h c d`.
+            let abef = _mm_shuffle_epi32::<REVERSE_LANES>(packed_abef);
```
Nashtare: Nit: after `REVERSE_LANES` this `abef` has `a` in lane 0, while `State::abef` and `INITIAL_ABEF` put `a` in the top lane, so the same name means opposite lane orders within the file. `sha2` calls this value `feba`.

sha256/src/x86_64_sha_ni.rs
```diff
@@ -0,0 +1,526 @@
...
+        let mut out_b = [[0u8; 32]; LANES];
+        hash_many(&batch_a, &mut out_a);
+        hash_many(&batch_b, &mut out_b);
+
+        // The untouched lanes must agree, and the changed one must not.
+        assert_eq!(out_a[..LANES - 1], out_b[..LANES - 1]);
+        assert_ne!(out_a[LANES - 1], out_b[LANES - 1]);
+    }
+
+    proptest! {
```
Nashtare: Nit: these shape tests and proptests only build under the `sha` gate, so no CI leg runs them, and `proptest` is built but unused on every current leg. Written against `Sha256::hash_many` / `Sha256Compress::compress_many` in `lib.rs`, they would also exercise the scalar default everywhere and `wasm32_simd128` where that is tested.

Nashtare (commented):
Reviewed at 72824b4c; 5168a0df only drops a doc line. There's no SHA-NI hardware here (Apple M4 Pro: Rosetta hides SHA, and a `+sha` binary dies with SIGILL under it). So I checked correctness against a software model of the intrinsics, and measured no timings.

**Kernel: no defect found (local probe)**
- **The model.** I wrote every intrinsic the kernel uses from the SDM definitions. To validate it, I compiled `sha2` 0.11's `x86_sha.rs` against it: the FIPS 180-4 vectors pass, and 3,303 random messages at lengths 0..=1100 match `sha2::Sha256::digest`.
- **The kernel under the model.** I compiled `x86_64_sha_ni.rs` against the same model, changing only the import. Your five tests pass. So do sweeps with 0 mismatches:
  - `hash_many` at every length 0..=300 × counts {1..=9, 13, 16, 17}, 22,876 digests of them from the 4-lane kernel;
  - longer lengths up to 16,441 bytes;
  - `compress_many` at counts 0..=40.
  The sweep also passes with `LANES` set to 1, 2, 3 and 8.
- **Negative controls fail as they should.** `K[63] + 1` fails exactly the kernel digests. Moving the padding boundary to `> 56` fails the lengths ≡ 56 (mod 64). Lane 3 reading lane 2's block fails lane independence.
- **Miri.** Over the modeled kernel it finds no UB in the `loadu`/`storeu` offsets or the transmutes.
- **Everything else is clean:** clippy `-D warnings` for x86_64 with and without `+sha,+sse4.1`, docs, the Rosetta scalar tests, aarch64 `p3-sha256`/`p3-merkle-tree`, fmt, and the thumbv7em and wasm32+simd128 builds.

**Main ask: nothing in CI compiles this module** (inline on `lib.rs`). None of the `features` values in `ci.yml` turns on `sha`, and neither do `x86-64-v3`/`v4` (`rustc --print cfg`). `macos-latest` is arm64, and the Windows baseline stops at `sse3`. So the module, its tests and the `4` arm of `reports_four_lanes_only_on_vectorized_targets` are never built. A build-only `+sha,+sse4.1` leg would fix that.

**Question:** gating is compile-time only. `+sha` alone doesn't enable it either, since it implies only `sse2`. So only `target-cpu=native`-style builds get `LANES = 4`, while `sha2` already dispatches at runtime. Is runtime dispatch deliberately out of scope?

Nit: I couldn't reproduce the Merkle end-to-end table from the tree. There is no SHA-256 Merkle bench, and `p3-merkle-tree` doesn't depend on `p3-sha256`.

sha256/src/lib.rs
tcoratger: Fixed in `908b619e` — you were right that nothing built it.

Added a `linux, sha, build only` leg with `features: "+sha,+sse4.1"`, matching the shape of the existing build-only legs. It also carries a `sha_ni: true` key doing two things: `cargo-nextest` now installs when `!compile_only || sha_ni`, and a final step runs `python3 scripts/check.py test --package p3-sha256` guarded on `grep -qw sha_ni /proc/cpuinfo`, printing a skip line otherwise — since GitHub's pool mixes parts with and without the instructions. Your point that `+sha` also makes `sha2` emit SHA-NI unconditionally is why the whole test step is guarded, not just the kernel.

Verified `cargo check --workspace --all-targets` under `+sha,+sse4.1` is clean, and the test command passes locally (13/13).

sha256/src/x86_64_sha_ni.rs
tcoratger: Fixed. `round_group` now indexes `ROUND_CONSTANTS.as_chunks::<WORDS_PER_VECTOR>().0[group]` and transmutes the `[u32; 4]`, so the bound survives release. The `# Panics` section records that both call sites range over `0..ROUND_GROUPS`.

It does fold away as you expected: in the release asm at `target-cpu=native`, the compression function has zero panic calls and one branch (the loop back-edge), with the constants coming from rip-relative loads hoisted out of the per-stream loop.

sha256/src/x86_64_sha_ni.rs
tcoratger: Fixed, and this was the most valuable of the five. The `#[cfg(test)] mod tests` is gone from the backend and merged into `sha256/src/lib.rs`, written against `Sha256::hash_many` / `Sha256Compress::compress_many`.

They now run on the scalar default everywhere, on `wasm32_simd128` where that is tested, and on SHA-NI via the new leg. Coverage widened at the same time: `SHAPE_LENGTHS` crossed with batch counts `0..=9` and `17`, and `hash_many_keeps_the_lanes_independent` rewritten against a fixed batch of four so it is meaningful on a scalar backend. 13 tests, and `cargo machete` is clean now that `proptest` is used unconditionally.

sha256/src/x86_64_sha_ni.rs
tcoratger: Fixed. `hash_four`, `hash_many`, `compress_many`, `K`, `LANES` and `BLOCK_BYTES` now live once in a cfg-gated `four_lane` module in `lib.rs`, generic over a `FourLane` trait (`type State`, `initial_state`, `compress`, `write_digests`). Each backend is a unit struct implementing it, which also collapsed six duplicated `#[cfg]` blocks down to one `cfg(any(..))` each. A padding fix can no longer land in only one backend.

Since this touches the wasm32 path, I checked it three ways: a `diff -u` shows the compression core, `load_word` and digest serialization character-identical apart from indentation and the constant rename; the constant tables compare numerically equal; and extracting the old wasm drivers and the new generic ones, applying only the mechanical substitutions, diffs identical for all three. Built for `wasm32-unknown-unknown +simd128` and `thumbv7em-none-eabi`. I could not execute the wasm tests — no `wasm32-wasip1` target or `wasmtime` here — so that last step rests on textual equivalence.

One gotcha worth recording: `H256_256.map(u32x4_splat)` does not compile, because `#[target_feature]` functions do not implement `Fn`. The original closure is load-bearing.

sha256/src/x86_64_sha_ni.rs
tcoratger: Fixed — `abef`→`feba`, `ghcd`→`dchg`, `abcd`→`dcba`, `efgh`→`hgfe`, so every name in the file reads top lane down as the state type documents, matching `sha2`.

Nashtare (approved):
Approving 908b619e. All five points from my earlier review are fixed, and I re-ran the correctness checks against the refactor.

**This head in CI.** The new `linux, sha, build only` leg found `sha_ni` on the runner and ran the suite with `+sha,+sse4.1`: 13 passed, including `reports_four_lanes_only_on_vectorized_targets`. So `LANES == 4` held, and the digest tests there went through the real four-lane kernel. No Graviton benchmark applies here: the backend is x86-only.

**Software model, re-run at this head.** I compiled the PR's own `lib.rs` (the `four_lane` driver) and `x86_64_sha_ni.rs` against the software model of the intrinsics. It is the same model as before, validated against `sha2`'s SHA-NI backend. 24/24 tests pass, and every sweep matches `sha2::Sha256::digest` / `Sha256Compress::compress`:

| sweep | mismatches | from the 4-lane kernel |
|---|---:|---:|
| `hash_many`, lengths 0..=300 × counts {1..=9, 13, 16, 17} | 0 / 27,391 | 22,876 |
| `hash_many`, 301..=1100 plus lengths up to 16,441 | 0 / 13,702 | 12,896 |
| `compress_many`, counts 0..=40 | 0 / 820 | 760 |
| 64 random seeds | 0 / 14,912 | |

Five negative controls each fail exactly the digests they should: a flipped round constant, padding `>=`→`>`, lane 3 reading lane 2's block, the leftover messages starting one late, and a wrong word order in the model's own `rnds2`. Miri on a subset is clean.

**The shared driver on wasm32.** Under wasmtime with `+simd128`, the PR tests pass (11, with the two `should_panic` tests ignored). The same sweeps through `Sha256::hash_many` give 0 mismatches, and a flipped round constant fails 22,876 of them.

**The five fixes**
- **CI leg:** it inherits its flags through `GITHUB_ENV`, and the `sha_ni` guard is sufficient, since every SHA-NI part has SSE4.1.
- **Release bound:** `as_chunks::<4>().0[group]` is checked, and the release build has no panic call left in `compress`.
- **Dedup:** padding and grouping live once in `four_lane`, and both backends only implement the trait.
- **Renames:** `feba`/`dchg`/`dcba`/`hgfe` match the shuffle, blend and align immediates.
- **Tests:** they go through `hash_many`/`compress_many` on every backend.

**Two small things, neither blocking**
- `lib.rs:5` says an x86-64 build needs `-C target-feature=+sha` or `target-cpu=native`. On linux and windows-msvc `+sha` alone doesn't enable `sse4.1` (`rustc --print cfg`), so the backend stays off and `LANES` stays 1. It should read `+sha,+sse4.1`, as the CI leg does.
- On a runner without SHA-NI the test step passes after a plain `echo`. A `::notice::` would make the skipped coverage visible, and the leg's "build only" name undersells that it runs tests.

PR #2153, opened 2026-09-16
perf(binary-field): convert whole blocks of GF(2^128) through GFNI

tcoratger:
The tower↔GHASH change of basis had a NEON arm and nothing on x86-64. This adds one, and a block kernel that drops the lookup tables entirely.

## The problem

`GF(2^128)` lives here in two bases: the Wiedemann tower (`BinaryField128`) and the GHASH polynomial basis (`Ghash128`). The `F_2`-linear map between them is `apply_128` in `binary-field/src/clmul/basis.rs` — sixteen byte-indexed table lookups, XORed.

It has a NEON arm that keeps an entry and the accumulator in one register. On x86-64 there was none, so a `u128` XOR was two general-purpose XORs and an entry load was two loads: about **32 loads and 30 XORs per element**, with two 64 KiB tables that overflow Zen 5's 48 KiB L1D.

That map is not a side path. A `BinaryField128` multiply converts both operands and the result, which is why it costs 14.9 ns against `Ghash128`'s 2.11 ns. `binary-dft`'s `PolyBasisNtt` — the default encoder, so the transform every `p3-binary-pcs` commit runs — rides an entry and an exit conversion on its tiles.

## The idea

I first dismissed GFNI for this, and I was wrong in an instructive way.

`vgf2p8affineqb` shares its 8×8 matrix operand across **all eight bytes of a quadword**. In the natural layout that kills it: each output byte needs its own block, so you would spend one instruction per (input byte, output quadword) pair.

**Transpose first and the constraint becomes the feature.** Take 64 elements — sixteen 512-bit registers — and transpose them to byte planes:

```text
    register r   holds elements 4r .. 4r+3        the layout in memory
    plane j      holds byte j of every element    what the affine map wants
```

Now every byte in plane `j` wants the *same* 8×8 block, which is exactly what the instruction broadcasts. Output plane `k` is the sum of sixteen affine images:

```text
    out_plane[k] = XOR over j of affine(in_plane[j], block[k][j])
```

and a transpose back restores the layout.

Per block of 64 elements: 16 loads, 128 `vpunpck`, 256 affine maps, 128 XORs, 16 stores. The per-element route over the same 64 elements takes **1024 dependent table loads**.

The transpose is four rounds of lane-local interleaving. It leaves the rows in bit-reversed order, which is undone by reindexing the array — no instruction at all.

## What ships

- `apply_128` gains an x86-64 arm mirroring NEON: one `__m128i` holds the entry and the accumulator. `sse2` is in the `x86_64` baseline, so no gate is needed.
- `mod blocked`, gated on `gfni + avx512f + avx512bw`, with a `BLOCK = usize::MAX` fallback on every other target — following the `mod wide` pattern `poly_slice.rs` already uses.
- `poly_basis::{from_tower_slice, to_tower_slice}` expose it.
- `binary-dft/src/poly.rs` converts its tiles through them; the parallel arm chunks at `BUTTERFLY_GRAIN` so every task still reaches the blocked kernel.

The 256 matrix operands are derived from `COLUMNS_128` by a `const fn` at compile time, so no new constant is hand-written.

## Correctness

The map is a ring isomorphism between two representations of a commitment field, so this was checked rather than argued.

In-tree: `byte_maps_match_scalar_evaluation` now drives every byte value in every position against **two independent references** — the scalar table fold *and* a `column_walk` over `COLUMNS_128` that touches no table and no intrinsic, so it cannot be tautological. Plus length `0..=200` × offsets `0..4` in both directions, sentinel-walled on both sides; a covered-count assertion so a build that silently loses the kernel fails; a `const _: () = assert!(..)` restating the cfg gate independently; and two proptests.

Before opening this I ran, and deleted, four more sweeps of my own:

- slice vs the **typed** per-element conversion (`Ghash128::from`) at every length `0..=200` × 4 offsets, sentinel-walled, plus the round trip;
- **multiplicativity** — `φ(a·b) == φ(a)·φ(b)` across a block boundary;
- **additivity** over `GF(2)`;
- 0 and 1 map to 0 and 1.

The last three depend on no table and no intrinsic in the crate, so they pin the *semantics* rather than self-consistency. As a negative control I flipped the row-order in the matrix derivation (`8*(7-i)` → `8*i`) and confirmed three of the four fail — additivity survives, correctly, since any `F_2`-linear map is additive.

Suites pass with the kernel live (`-C target-cpu=native`) and with it cfg'd out (`+pclmulqdq` only): `p3-binary-field` 184+171+4+6, `p3-binary-dft` 50, `p3-binary-pcs` 64.

## Benchmark

Ryzen 9 9950X3D, `RUSTFLAGS="-C target-cpu=native"`, pinned to the V-Cache die, before and after from the same bench code, runs serialized against other work on the machine. Crates at their default (single-threaded) feature set.

| bench | before | after | change |
|---|---:|---:|---:|
| `representations/convert/tower to ghash` | 3.545 µs | 2.385 µs | −32.7% |
| `representations/convert/ghash to tower` | 3.544 µs | 2.397 µs | −32.4% |
| `representations/convert/bulk tower to ghash` | — | 0.668 µs | **5.3x** |
| `representations/convert/bulk ghash to tower` | — | 0.674 µs | **5.3x** |
| `ntt/128/hybrid/14` | 2.222 ms | 1.319 ms | **−40.6%** |
| `ntt/128/hybrid/16` | 9.366 ms | 5.787 ms | **−38.2%** |
| `ntt/128/hybrid/18` | 43.08 ms | 28.86 ms | **−33.0%** |
| `ntt/128/hybrid/20` | 202.6 ms | 145.6 ms | **−28.1%** |
| `commit/16` | 18.93 ms | 18.07 ms | −4.5% |
| `commit/18` | 83.77 ms | 80.03 ms | −4.5% |
| `commit/20` | 351.2 ms | 336.8 ms | −4.1% |

## Where it does not help, stated plainly

**At 32 threads on a width-1 matrix — the shape `p3-binary-pcs` actually commits — the change is below measurement noise.** Four alternating runs of the `encode` span at `2^22` gave 92.0 / 102 / 79.9 / 108 ms before and 105 / 98.1 / 111 / 105 ms after. That shape is memory-bandwidth-bound, and the conversion was never the bottleneck there. It is also the shape where `Plan::new` disables the staging tile entirely, which is a separate problem.

The 28–42% on `ntt/128/hybrid` is real, and it is the width-16 single-threaded shape the bench measures. The −4.3% on `commit` is real. I would rather state both than average them into something that flatters the change.

**`representations/mul` is unchanged** (−0.5% / +0.1%, i.e. nothing). A tower product's three conversions sit on one dependency chain, and the vector accumulator has the same latency as the two general-purpose chains it replaces. The per-element arm buys throughput, which only a bulk caller can spend. On its own it would have been a microbenchmark-only change; the blocked kernel is what makes this worth merging.

## Scope

- **Not wired:** `binary-pcs/src/fold.rs`. `fold_task` converts inside `Packed::from_fn(|lane| Ghash128::from(..))` and writes back per lane; feeding it a slice needs a per-task scratch buffer and a reinterpretation of `&mut [BinaryField128]` as `&mut [u128]`. That is a hot-loop restructure and belongs in its own PR — it is the largest remaining win here, since `fold_codeword/18` is 11.2 ns per output symbol of which three conversions are ~64%.
- **No AVX2 arm.** A GFNI target without AVX-512 (Alder Lake, for instance) takes the per-element path. A 256-bit variant is possible later.

## A measurement note for this repository

The 9950X3D is asymmetric: cores 0–7 share 96 MiB of L3, cores 8–15 share 32 MiB. Unpinned, the same binary measured `fold_codeword/20` (a 64 MiB working set) at 26.1 / 28.0 / 29.9 ms across three builds, which reads as a 17% regression and is entirely CCD placement. Every number above is `taskset`-pinned. Worth knowing before trusting any un-pinned result on this class of machine.

## Verification

- `cargo test -p p3-binary-field -p p3-binary-dft -p p3-binary-pcs`, native and with the kernel cfg'd out — all green.
- `cargo check` for `aarch64` bare and `+neon,+aes`, and for `wasm32` — the portable arm compiles.
- `cargo clippy --all-targets` clean; `cargo fmt` applied.


Nashtare (commented):
Reviewed at 79dfe626; 46a1f993 only deletes blank `///` lines. I found no correctness or safety problem. There's no GFNI or AVX-512 silicon here (Apple M4 Pro), so I checked the kernel against a software model and under Miri rather than running it.

**What holds**
- **Intrinsic model.** I compiled `mod blocked` against a software model of each intrinsic it uses, written from Intel's operation text: `affine_byte` with row `i` = byte `7−i`, unpacks within each 128-bit lane, and the VPTERNLOGQ truth table. The model reproduces the identity matrix, bit reversal and the AES S-box affine map. With only the arch import swapped:
  - all 13 `basis.rs` tests pass;
  - every length 0..=300 × 7 offsets passes in both directions, against `Ghash128::from`/`BinaryField128::from`, with sentinels per position;
  - every single-bit input in each of the 64 slots passes.
- **Negative controls all fail:**
  - row order read as `byte[i]`;
  - unpacks across the whole 512-bit register;
  - swapped `unpack*_epi16`;
  - `8*(7-i)` → `8*i`;
  - one flipped matrix bit;
  - identity instead of `BITREV4`;
  - `N` for the inverse;
  - one register loaded an element late.
- **Miri on the real `core::arch` code** (`+gfni,+avx512f,+avx512bw`, so the affine semantics are Miri's own). `vpternlogq` is swapped for `vpxorq`, since Miri lacks it. The bounds-checked length sweep and `the_blocked_prefix_covers_every_whole_block` pass with no UB, and the row-order mutant fails.
- **The SSE2 arm under Rosetta.** `p3-binary-field` and `p3-binary-dft` tests pass, both baseline and `+pclmulqdq`, as does `p3-binary-pcs` with `+pclmulqdq`. Making the arm skip one table fails 7 tests.
- **aarch64.** Serial release asm for `p3-binary-dft` matches main apart from panic line numbers. Tests pass with default features and `parallel`.
- Nightly fmt, the wasm32 check, and x86 clippy on four feature sets are clean, apart from the `parallel` nit inline.

**Main ask: CI never compiles `mod blocked`** (inline). `+avx512f` implies neither `avx512bw` nor `gfni`, so the `avx512f` and `avx512f+vpclmulqdq` build-only legs compile the `usize::MAX` stub. The kernel, the `BLOCKED_BUILD == true` side of the `const _` assert, and the covered-count test are never type-checked, linted or run.

**Smaller points, inline:**
- no test fails if the slice dispatchers stop reaching the kernel;
- a question about codegen on a feature-flag build;
- the `x86_64-unknown-none` baseline;
- a clippy failure under `parallel`;
- doc wording.

**Conflict with #2155** in `binary-dft/src/poly.rs`. 46a1f993 also removes 7 blank doc lines there (217–499), inside #2155's hunks, in docs this PR doesn't otherwise change. Dropping those keeps the conflict to the real change.

binary-field/src/clmul/basis.rs
```diff
@@ -346,6 +367,247 @@ fn apply_128(tables: &[[u128; 256]; 16], v: u128) -> u128 {
     }
 }
 
+/// The blocked kernel, over whichever instruction set converts many elements at once.
+#[cfg(all(
```
Nashtare, line 372: No CI leg compiles this module. `+avx512f` implies neither `avx512bw` nor `gfni` (`rustc --print cfg --target x86_64-unknown-linux-gnu -C target-feature=+avx512f` adds only `avx512f`), so the `avx512f` and `avx512f+vpclmulqdq` build-only legs compile the `usize::MAX` stub, and clippy and docs run on the baseline. Could a build-only leg with `+avx512f,+avx512bw,+gfni` go in, so the kernel and the `BLOCKED_BUILD` side of the tests are at least type-checked and linted?

binary-field/src/clmul/basis.rs
```diff
@@ -387,23 +685,198 @@ mod tests {
...
+            // Offsets 0 through 3 place the payload at every alignment one register can see.
+            for offset in 0..4 {
+                slices_agree(&values, offset)
+                    .unwrap_or_else(|e| panic!("len {len}, offset {offset}: {e}"));
+            }
+        }
+    }
+
+    #[test]
+    fn the_blocked_prefix_covers_every_whole_block() {
```
Nashtare: This calls `blocked::*` directly, so it checks `chunks.len() * BLOCK` but not that `tower_to_poly_128_slice`/`poly_to_tower_128_slice` reach the kernel. Against a software model of the intrinsics, replacing both dispatch expressions with `let blocked = 0;` still passes every test in this module. Gate drift is already a compile error through the `const _` assert. Could the dispatchers report the covered count, so this asserts through them?

binary-field/src/clmul/basis.rs
```diff
@@ -346,6 +367,247 @@ fn apply_128(tables: &[[u128; 256]; 16], v: u128) -> u128 {
...
+            //
+            // Both accesses are the unaligned forms, so the alignment of the slice is free.
+            //
+            // The target features every intrinsic needs gate this module.
+            unsafe {
+                let raw: [__m512i; 16] =
+                    core::array::from_fn(|r| _mm512_loadu_si512(chunk.as_ptr().add(4 * r).cast()));
+                let input = transpose(raw);
+                let output: [__m512i; 16] = core::array::from_fn(|k| plane(&input, &blocks[k]));
+                let output = transpose(output);
```
Nashtare, line 566: Question: was the speedup also measured on a feature-flag build? With `-C target-feature=+gfni,+avx512f,+avx512bw` and no `target-cpu`, release `apply` makes two 1024-byte `memcpy` calls per block and moves the butterfly rounds through stack slots (116 `vmovdqa64` + 108 `vmovaps`); with `-C target-cpu=znver5` it is 17 `vmovdqa64` and no calls. I couldn't time either here.

binary-field/src/clmul/basis.rs
```diff
@@ -336,7 +335,29 @@ fn apply_128(tables: &[[u128; 256]; 16], v: u128) -> u128 {
             core::mem::transmute::<uint8x16_t, u128>(acc)
         }
     }
-    #[cfg(not(all(target_arch = "aarch64", target_feature = "neon")))]
+    #[cfg(target_arch = "x86_64")]
```
Nashtare: Nit: `sse2` isn't in the baseline of `x86_64-unknown-none` or `x86_64-unknown-uefi` (`rustc --print cfg` lists only `fxsr`). There this arm compiles to out-of-line `_mm_loadu_si128`/`_mm_xor_si128` calls, 33 per element, instead of the scalar fold (checked on a verbatim copy with `-Zbuild-std=core`). Gating on `all(target_arch = "x86_64", target_feature = "sse2")`, like the NEON arm, keeps the fallback there and makes the SAFETY line true.

binary-dft/src/poly.rs
```diff
@@ -91,37 +91,35 @@ fn for_chunks(
...
+fn convert(values: &mut [u128], conversion: Conversion) {
     // Basis conversion does several dependent lookups per element, more work than
     // a butterfly, so it amortizes dispatch at a smaller byte volume.
+    //
+    // A task is a whole grain, so every task still reaches the blocked kernel.
     if use_parallel(values.len().saturating_mul(4)) {
-        values
-            .par_iter_mut()
-            .for_each(|value| *value = conversion(*value));
+        values.par_chunks_mut(BUTTERFLY_GRAIN).for_each(&conversion);
```
Nashtare: Nit: `cargo clippy -p p3-binary-dft --all-targets --features parallel -- -D warnings` fails here with `needless_borrows_for_generic_args`; `for_each(conversion)` fixes it. The same command is clean on main, and CI's clippy job doesn't enable `parallel`.

binary-field/src/poly_basis.rs
```diff
@@ -27,6 +27,26 @@ pub fn to_tower(v: u128) -> BinaryField128 {
     BinaryField128::from_repr(clmul::poly_to_tower_128(v))
 }
 
+/// The polynomial-basis coordinates of a whole slice of tower-basis bit patterns.
+///
+/// # Performance
+///
+/// Whole blocks of the slice cross over together where the target can transpose them.
```
Nashtare: Nit: "where the target can transpose them" is a compile-time gate with no runtime detection. Could the doc name `gfni`, `avx512f` and `avx512bw`? `-C target-cpu=x86-64-v4` enables `avx512bw` but not `gfni`, so a v4 build takes the per-element path.

Nashtare (approved):
Approving. I benchmarked 46a1f993 against main 463b44c3 on Graviton5 (`c9g.24xlarge`, 96 Neoverse-V3 cores, one NUMA node, `-C target-cpu=native`). Each comparison ran twice, alternating base and head. Serial runs were pinned to one core, and parallel runs to cores `0..T`. Numbers are criterion medians averaged over both rounds. On this host the change is the aarch64 path: the slice conversions with `BLOCK = usize::MAX`, and `BUTTERFLY_GRAIN` chunking in the parallel arm.

| bench | main | this PR | change |
|---|---:|---:|---:|
| `representations/convert` tower→ghash (per element) | 2.97 µs | 2.97 µs | +0.1% |
| `ntt/128/hybrid`, 2^14 … 2^20, serial | | | +0.2% … +0.8% |
| `ntt/128/hybrid`, 2^14 … 2^20, 32 threads | | | −0.9% … −0.4% |
| `ntt/128/hybrid`, 2^14 … 2^20, 96 threads | | | −1.1% … +0.8% |
| `poly/forward/h16/w1`, 32 threads | 322.2 µs | 171.1 µs | 1.88x |
| `poly/inverse/h16/w1`, 96 threads | 1.043 ms | 524.5 µs | 1.99x |
| `commit/16 / 18 / 20`, serial | 62.3 / 251.4 / 1016 ms | 62.4 / 251.6 / 1016 ms | +0.1% |
| `commit/16 / 18 / 20`, 32 threads | 3.48 / 10.61 / 37.40 ms | 3.37 / 10.51 / 37.34 ms | −3.1% / −0.9% / −0.2% |
| `commit/16 / 18 / 20`, 96 threads | 3.56 / 8.15 / 21.13 ms | 3.03 / 7.43 / 20.15 ms | −15.0% / −8.9% / −4.6% |

**No aarch64 regression.** The serial transform and serial commit are flat, consistent with the release asm matching main. The parallel arm's minimum task size is a clear win for small, narrow matrices: width-1 transforms run ~2x faster, and `commit` gains up to 15% at 96 threads.

Two notes, neither blocking:
- **CI still doesn't compile `mod blocked`** (inline earlier). A build-only `+avx512f,+avx512bw,+gfni` leg, like the SHA leg #2152 just added, would at least type-check and lint the kernel.
- **The `bulk` bench ids mislead off GFNI.** On this host they read 4.84 µs against 2.97 µs for the per-element ids. They aren't the same workload: the bulk ids convert in place and pay a clone per batch, while the per-element ids fold without storing. So off GFNI they aren't a like-for-like "bulk vs per element" comparison.

binary-field/src/clmul/basis.rs
tcoratger: Fixed, and I confirmed your premise first: `rustc --print cfg -C target-feature=+avx512f` yields neither `avx512bw` nor `gfni`, so every existing leg was compiling the `usize::MAX` stub.

Added a build-only leg `linux, avx512f+avx512bw+gfni`. For the "and linted" half I went wider than the one leg: `scripts/check.py architecture` now emits `cargo build --all-targets` **and** `cargo clippy --all-targets -- -D warnings`, so every build-only leg lints its own feature combination rather than just this one. The argument for linting under `+gfni` applies verbatim to `+avx2` and `+sve2`, so a one-off knob would have been arbitrary. The toolchain step now asks for `components: clippy`, since dtolnay's action installs the minimal profile.

Verified locally: `cargo clippy --target … --all-targets -- -D warnings` exits 0 for `+avx2`, `+avx2,+vpclmulqdq`, `+avx512f`, `+avx512f,+vpclmulqdq` and `+avx512f,+avx512bw,+gfni` (serial and parallel for the last). `scripts/test_check.py` passes 23 tests.

I kept build *and* clippy rather than clippy alone: clippy never codegens, and "instruction requires avx512" errors only appear at codegen, which is most of what a build-only leg is for.

binary-field/src/clmul/basis.rs
tcoratger: Fixed, and I reproduced your control to confirm it.

`tower_to_poly_128_slice` / `poly_to_tower_128_slice` now return the covered count. `slices_agree` asserts it at every length and offset, and the block-coverage test goes through the dispatchers instead of `blocked::*`. The expected count is `len - len % blocked::BLOCK`, which states both builds' claims in one expression since `BLOCK == usize::MAX` on the stub side.

Replacing both dispatch expressions with `let blocked = 0;` now fails three tests, where before it failed none.

binary-field/src/clmul/basis.rs
tcoratger: Fixed — the arm is now `#[cfg(all(target_arch = "x86_64", target_feature = "sse2"))]`, with the scalar fallback's `not(any(..))` mirrored, so those targets get the scalar fold back.

Confirmed `x86_64-unknown-none` lists only `fxsr`. The SAFETY line now says the gate is what makes it true, and I corrected the commit message's "so the arm needs no gate" sentence, which was wrong for the same reason.

binary-dft/src/poly.rs
tcoratger: Fixed — `for_each(&conversion)` -> `for_each(conversion)`. `cargo clippy -p p3-binary-dft --all-targets --features parallel -- -D warnings` is clean.

Worth noting this is exactly the gap the change above closes from the other direction: CI's clippy job does not enable `parallel`, so nothing would have caught it.

binary-field/src/poly_basis.rs
tcoratger: Fixed — the doc now names `gfni`, `avx512f` and `avx512bw`, says the choice is compile-time with no runtime detection, and states that `-C target-cpu=x86-64-v4` gives `avx512bw` but not `gfni`, so a v4 build takes the per-element path.

binary-field/src/clmul/basis.rs
tcoratger: Measured both, since you could not time them here. Ryzen 9 9950X3D, `taskset -c 0-7`, serialized against other work, each side its own target dir; the "before" column is the parent commit built from `git archive` into a scratch tree with identical flags.

| `representations/convert` | `-C target-cpu=native` | `-C target-feature=+gfni,+avx512f,+avx512bw` |
|---|---|---|
| per-element, parent | 3.549 µs | 3.658 µs |
| per-element, this PR | 2.396 µs (−32%) | 2.550 µs (−30%) |
| bulk, this PR | 0.668 µs (5.3x parent) | 0.897 µs (4.1x parent) |

So **yes, materially slower**: the block kernel gives up about 34% of its throughput on a feature-flag build, though it still beats the per-element route by 4.1x.

Your codegen observation reproduces in kind — `blocked::apply` under the feature flags has two `call *%r13` with `$0x400` in `edx` against `memcpy@GLIBC_2.14`, and none under `target-cpu=native`. The magnitude differs on rustc 1.98.0 though: 324 instructions and 41 `vmovdqa64` with no `vmovaps`, against 265 and 17 for the `target-cpu` build.

This is now stated in a paragraph on `from_tower_slice` and in the commit message, since a user who enables the features without `target-cpu` should know what they are getting.

PR #2154, opened 2026-09-16
Lean extraction of p3-baby-bear and some minimal proofs

dhsorens:
## Summary

Adds a formal-verification tree for the `baby-bear` crate under `baby-bear/proofs/`. Proofs are minimal for now, as this PR focuses primarily on proof infrastructure.

## Pipeline

The crate is translated to Lean with [hax](https://github.com/cryspen/hax)'s legacy Lean backend.

- the translation **type-checks**
- the proofs in the [`spec/`](https://github.com/Plonky3/Plonky3/tree/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/spec) directory type-check, meaning that they hold under the assumptions (see [`TCB.md`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/TCB.md) for more).

## Layout

| Path | Role |
|------|------|
| `baby-bear/proofs/` | FV root for this crate |
| `…/legacy-lean/` | The Lake package — hax's `legacy-lean` backend, the only extractor in use |
| `…/extraction/` | Generated Lean, plus hand-written stubs for `p3-monty-31`, `p3-field`, MDS, Poseidon, … |
| `…/spec/` | The five constant theorems |
| `…/patches/` | Six small Lean-side patches. **No Rust patches.** |
| `…/build-proofs.sh` | Re-extract from Rust and rebuild |

## Caveats

- **Nothing in CI.** Nothing here runs in GitHub Actions. Rebuilding needs `cargo-hax` and Lean, opt-in.
- **Very minimal proofs** Only minimal constants are checked, as I didn't want to overload reviewers with too many proofs. If this structure is approved, I can do a full verification pass of the baby-bear crate w.r.t. its specification in [CompPoly](https://github.com/Verified-zkEVM/CompPoly/blob/v4.30.0/CompPoly/Fields/BabyBear.lean).
- **Not a long-term extractor choice.** Aeneas and hax's current Lean backend could replace this later; they are currently unused because their extraction of this crate is not clean enough yet.

## Useful Reading Order

1. [`baby-bear/proofs/README.md`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/README.md) — why this directory exists
2. [`legacy-lean/README.md`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/README.md) — layout and how to build
3. [`spec/…/constants.lean`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/spec/p3_baby_bear_proofs/constants.lean) — the theorems
4. [`TCB.md`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/TCB.md) — what is trusted vs proved
5. The extraction lives in [`extraction/`](https://github.com/Plonky3/Plonky3/tree/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/extraction) (~4k lines, generated by hax).
6. The spec and formal proofs of correctness live in [`spec/`](https://github.com/Plonky3/Plonky3/tree/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/spec).

If you change `baby-bear/src`, you can re-extract using the [`build-proofs.sh`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/build-proofs.sh) script. If the source changes and it does not typecheck, then the re-extraction runbook (for an agent) is [`SYNC.md`](https://github.com/Plonky3/Plonky3/blob/eefe5718498c8aede2407b252094676bb36ac82e/baby-bear/proofs/legacy-lean/SYNC.md). You do not need to run it for ordinary Rust review.

## Test plan

- [ ] `git status --porcelain baby-bear/src monty-31/src` is empty
- [ ] `cd baby-bear/proofs/legacy-lean && lake exe cache get && lake build` is green, 0 warnings
- [ ] (optional, needs hax) `./build-proofs.sh` reproduces `extraction/p3_baby_bear.lean`


dhsorens:
I haven't included a `setup` script but an agent will get you set up—happy to include something like that if it would be useful

Nashtare (commented):
Thanks a bunch! This will take some time to review as I familiarize myself and play around a bit more with this. 
One concern I have though, is that nothing prevents the entire `proofs/` to be included in the released package, which would blow up its size.

Just adding `exclude = ["proofs"]` in the `Cargo.toml` would address this.

dhsorens:
thanks @Nashtare ! just made that fix, good call - I'm very happy to chat about this and clear up any questions that come up while you're reviewing (happy to jump on a call too)

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear/p3_field.lean
```diff
@@ -0,0 +1,181 @@
...
+    ASSUMED — the body here is the identity, which is *not* the upstream
+    algorithm. Nothing may be concluded about the S-box from this stub. -/
+def exp_1725656503
+    (R : Type)
+    [_trait_constr_exp_1725656503_associated_type_i0 :
+      p3_field.field.PrimeCharacteristicRing.AssociatedTypes
+      R]
+    [_trait_constr_exp_1725656503_i0 : p3_field.field.PrimeCharacteristicRing R ]
+    (val : R) :
+    RustM R := pure val
```
tcoratger: **[MAJOR] `exp_1725656503` has a definitional identity body, so a statement that is false of the Rust is provable by `rfl`**

Upstream (`field/src/exponentiation.rs:62`) is the addition chain computing `val^1725656503`; here it is `pure val`. Because this is a `def` and not an `opaque`, the identity is not merely *unmodelled* — it is *asserted*. I built the package and checked:

```lean
theorem sbox_is_identity (R : Type) [...] (val : R) :
    Impl_4.exp_root_d_hoisted R val = pure val := rfl
-- 'sbox_is_identity' does not depend on any axioms
```

That statement is false of `p3-baby-bear`, it typechecks, and it is axiom-free — so `#print axioms` will never flag it and `TCB.md`'s clean footprint does not cover it. This is exactly the failure mode the `opaque`-not-`sorry` argument in the doc-comment at lines 108-118 was written to prevent; the rule just is not applied here. `TCB.md` does name it under "Two infidelities", but a reader who greps for assumptions rather than reading the prose will not find it, and a future proof about Poseidon2's S-box could be built on it in good faith.

**Fix:** make it `opaque exp_1725656503 (R : Type) [...] : R -> RustM R`, matching the six `ring_*` declarations below. Nothing then follows about the S-box, which is the intended position.

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear/p3_mds.lean
```diff
@@ -0,0 +1,16 @@
...
+
+def first_row_to_first_col
+    (N : usize)
+    (T : Type)
+    [_trait_constr_first_row_to_first_col_associated_type_i0 :
+      core_models.marker.Copy.AssociatedTypes
+      T]
+    [_trait_constr_first_row_to_first_col_i0 : core_models.marker.Copy T ]
+    (v : (RustArray T N)) :
+    RustM (RustArray T N) := pure v
```
tcoratger: **[MAJOR] `first_row_to_first_col` is the identity, which makes all six extracted MDS column tables wrong**

Upstream (`mds/src/util.rs:52-62`) sets `col[i] = row[N - i]` for `i >= 1` — it reverses everything after the first element. Here it is `pure v`. Same problem as `p3_field.lean:75`: a `def` asserts, an `opaque` would not. Verified against the built package:

```lean
theorem mds_col_is_row (N : usize) (T : Type) [...] (v : RustArray T N) :
    p3_mds.util.first_row_to_first_col N T v = pure v := rfl
-- 'mds_col_is_row' does not depend on any axioms
```

This one is load-bearing for data, not just for a function: `baby-bear/src/mds.rs:16-21+` pipes every one of `MATRIX_CIRC_MDS_{8,12,16,24,32,64}_COL` through it, so the corresponding constants at `extraction/p3_baby_bear.lean:617,634,652,674,704,742` evaluate to the circulant matrix's first *row* where the Rust holds its first *column*. Anyone reading those Lean constants as the MDS matrix reads the wrong numbers.

**Fix:** either `opaque`, or — better, since the function is nine lines of total `const fn` — transcribe it faithfully. It is a bounded `while` over `i` with `output[i] = v[N - i]`, well within what the interface already does for `to_monty`.

baby-bear/proofs/legacy-lean/TCB.md
```diff
@@ -0,0 +1,349 @@
...
+One file per crate in `baby-bear/Cargo.toml`'s `[dependencies]`, so the
+axiomatization boundary is exactly the crate's declared dependency edge.
+
+### Given real, faithful bodies
+
+Transcriptions of upstream, not placeholders. Supplying these is what lets 45 of
+46 `of_isOk` obligations discharge by `rfl` — the whole `(by rfl)` → `(by sorry)`
+patch category disappeared.
+
+**These transcriptions are now unchecked.** `SanityCheck.lean`, which verified
```
tcoratger, line 104: **[MAJOR] Nothing detects drift between `baby-bear/src` and the extraction**

This paragraph is the most important one in the PR and it is an admission: `SanityCheck.lean` was removed, so `to_monty`, `Impl.new`, `MONTY_MASK`, `ODD_FACTOR`, `PRIME_I128` and `PSEUDO_INV` are now hand transcriptions checked by nothing. Combined with the absence of any CI job (no workflow in `.github/workflows/` mentions `lean` or `proofs`), the failure mode is that a `monty-31` change lands, the Lean keeps building because the stale constant is still well-typed, and the tree silently becomes a formalization of a function Plonky3 no longer ships. An unmaintained formalization does not decay into no formalization; it decays into a wrong one that still looks green.

Good news on the current state: I checked, and the extracted surface has *not* drifted. Since this branch's merge-base, `baby-bear/src` changed only in `x86_64_avx2/packing.rs` and `x86_64_avx512/packing.rs`, both `cfg`-excluded by the `thumbv7em-none-eabi` target, and `monty-31/src/{utils,monty_31,data_traits}.rs`, `mds/src/util.rs` and `field/src/exponentiation.rs` are untouched. So this is about the next change, not this one.

**Fix:** restore the cheap half of `SanityCheck.lean` — a Lean file asserting each transcribed constant by `rfl` against a literal — and add a CI job on `paths: ['baby-bear/**']` running `lake exe cache get && lake build`. The build is ~30s once mathlib is cached; I ran it here and it is green with 0 warnings, so the job would be green from day one.

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear.lean
```diff
@@ -0,0 +1,4306 @@
...
+                (rust_primitives.hax.monomorphized_update_at.update_at_usize
+                  a
+                  k
+                  (← (prime &&&? mask)));
+              (pure a);
+          let k : usize ← (k +? (1 : usize));
+          (pure (rust_primitives.hax.Tuple2.mk a k)) :
+          RustM (rust_primitives.hax.Tuple2 (RustArray u64 64) usize))));
+    (pure a))
+    (by native_decide) -- PATCHED: `SAMPLING_BITS_M` is a Rust `const while`
```
tcoratger, line 571: **[MINOR] The one `native_decide` pulls `Lean.ofReduceBool` into the footprint of `SAMPLING_BITS_M`**

Flagging it explicitly since it is the single non-kernel-checked step in the tree. The trade-off is argued correctly in patch `040`'s header and in `TCB.md` layer 4, and the conclusion (better than `sorry`, because it at least establishes the fact) is right. Note it also means the compiled evaluator has to reduce a 64-iteration `Loop.MonoLoopCombinator` fixpoint: fine today, but `native_decide` failures are opaque when they happen.

None of the five proved theorems depend on it, so this does not weaken any headline claim.

**Fix:** none required. `TCB.md` already sketches the clean removal (a bespoke `Loop.MonoLoopCombinator` argument with the real measure `64 - k`). Worth keeping on the list; also worth stating in the PR description, not only in `TCB.md`, since it is the one thing a reviewer must consciously accept.

baby-bear/proofs/legacy-lean/spec/p3_baby_bear_proofs/constants.lean
```diff
@@ -0,0 +1,54 @@
...
+    (MontyParameters.PRIME BabyBearParameters).toNat = BabyBear.fieldSize := by
+  simp [BabyBear.fieldSize, MontyParameters.PRIME, Impl.PRIME_hoisted]
+
+/-- Two-adicity declared for FFT / NTT matches `BabyBear.twoAdicity` (27). -/
+theorem two_adicity_eq_spec :
+    (TwoAdicData.TWO_ADICITY BabyBearParameters).toNat = BabyBear.twoAdicity := by
+  simp [BabyBear.twoAdicity, TwoAdicData.TWO_ADICITY, Impl_5.TWO_ADICITY_hoisted]
+
+/-- Montgomery representation uses 32-bit limbs (Plonky3 `MontyField31`). -/
+theorem monty_bits_eq_thirtyTwo :
```
tcoratger, line 43: **[MINOR] `monty_bits_eq_thirtyTwo` is a tautology, and the theorem that would matter is missing**

After `simp` unfolds `Impl.MONTY_BITS_hoisted` this is `32 = 32`. Unlike `monty_prime_eq_fieldSize` and `two_adicity_eq_spec` it compares the extraction against nothing external, so it cannot fail for any reason a reviewer cares about. Keep it as a regression tripwire if you like, but the docstring ("Montgomery representation uses 32-bit limbs") reads as a guarantee and it is not one.

Two one-line theorems with real content are available right here and would be much stronger for the same effort:

1. `PRIME * MONTY_MU = 1 (mod 2^32)` — the actual correctness precondition for Montgomery reduction, which the Rust asserts at compile time in `monty-31/src/monty_31.rs:57` and which the Lean `Impl.new` silently drops. I checked it numerically: `2013265921 * 2281701377 mod 2^32 = 1`. In Lean it is `decide` over a `Nat` product.
2. `Nat.Prime (MontyParameters.PRIME BabyBearParameters).toNat` — `CompPoly.BabyBear.is_prime` already ships this with a Pratt certificate, and `TCB.md` notes it is unused. Chaining it through `monty_prime_eq_fieldSize` is two lines and turns "the number matches" into "the modulus is prime".

**Fix:** add both; they are the cheapest real content available in this file.

baby-bear/proofs/legacy-lean/spec/p3_baby_bear_proofs/constants.lean
```diff
@@ -0,0 +1,54 @@
...
+bijection here. The extraction declares `RelativelyPrimePower BabyBearParameters
+((7 : u64))` accordingly. -/
+theorem coprime_seven_pred_prime :
+    Nat.Coprime 7 ((MontyParameters.PRIME BabyBearParameters).toNat - 1) := by
+  simp [MontyParameters.PRIME, Impl.PRIME_hoisted]
+  decide
+
+/-- Factorization of `p - 1` into `2^twoAdicity * 15` (two-adic FFT domain size).
+For KoalaBear the odd part is `127`; for BabyBear it is `15 = 3 * 5`. -/
+theorem fieldSize_sub_one_factorization' :
```
tcoratger, line 68: **[MINOR] `fieldSize_sub_one_factorization'` relates two extracted constants to each other, with the odd part supplied by hand**

Both sides come from the extraction (`Impl.PRIME_hoisted` and `Impl_5.TWO_ADICITY_hoisted`), and `15` is written by the author rather than derived, so on its own this says "these two numbers we extracted are consistent with a 15 we chose". `TCB.md` says as much and tells the reader to pair it with `two_adicity_eq_spec` — fair, but the docstring here does not, and this file is what a reviewer reads first.

The stronger statement is the maximality one, which is what `TWO_ADICITY` actually has to mean: `Odd 15`, i.e. `2^27` is the *largest* power of two dividing `p - 1`. As written, `TWO_ADICITY = 20` with odd part `1920` would satisfy an analogous statement, and nothing here rules that shape out.

Also a naming nit: the trailing prime in `fieldSize_sub_one_factorization'` conventionally marks a variant of an unprimed lemma, and there is no unprimed one.

**Fix:** add `Odd 15` (or `¬ 2 ∣ 15`) to the conjunction, and drop the apostrophe. Point the docstring at `two_adicity_eq_spec` the way `TCB.md` does.

baby-bear/proofs/legacy-lean/build-proofs.sh
```diff
@@ -0,0 +1,200 @@
...
+HAX_TOOLCHAIN="${HAX_TOOLCHAIN:-nightly-2025-11-08}"
+# The Rust driver spawns the OCaml engine and aborts if their revisions differ,
+# so this must be the engine built from the same hax revision as `cargo-hax`.
+HAX_ENGINE_BINARY="${HAX_ENGINE_BINARY:-$HOME/.opam/hax/bin/hax-engine}"
+# A target with no SIMD. `baby-bear/src/lib.rs` gates aarch64_neon / x86_64_avx2
+# / x86_64_avx512 on target_arch+target_feature, and `monty-31/src/lib.rs`
+# selects its `no_packing` path under the same negation, so a bare-metal triple
+# excludes every SIMD module across the whole dependency graph. This is the
+# configuration CI's `check_embedded` job already builds.
+HAX_TARGET="${HAX_TARGET:-thumbv7em-none-eabi}"
```
tcoratger, line 65: **[MINOR] `thumbv7em-none-eabi` excludes every code path production actually runs**

This is a genuinely clever choice — it is what keeps `patches/pre-extraction/` empty, which is the strongest structural property of the PR, and I would not change it. But the consequence deserves to be stated where Rust reviewers will see it: the AVX2, AVX-512 and NEON backends in `baby-bear/src/` and `monty-31`'s packing modules are `cfg`-excluded, so the extraction covers only the portable `no_packing` path. On x86-64 and aarch64, Plonky3 does not execute the code this tree models.

That is defensible — the portable path is the reference the SIMD paths are tested against, and the repo has differential tests for exactly that — but "we verified BabyBear" would be a materially misleading summary of what lands here.

`TCB.md` says this under "The extraction is target-scoped". The PR description does not.

**Fix:** add one line to the PR description and to `baby-bear/proofs/README.md` saying the extraction is scalar-only and that the SIMD backends are out of scope.

baby-bear/proofs/legacy-lean/README.md
```diff
@@ -0,0 +1,91 @@
+# Hax `legacy-lean` backend
```
tcoratger, line 1: **[MINOR] A Rust maintainer editing `baby-bear/src` gets no signal that this tree exists**

`SYNC.md` is the runbook for re-extraction and `TCB.md` is the trust boundary, but both live three directories down and nothing points at them from where a Rust contributor works. `CONTRIBUTING.md` is not touched, and there is no `baby-bear/src` comment or CI annotation. The first person to change `MontyParameters` after this merges will not know they should have re-run anything.

This is also the place to answer the question a non-Lean maintainer will actually ask, in two sentences at the top: *what is guaranteed, and what is not*. Right now that answer is only reachable by reading 349 lines of `TCB.md`. Suggested wording: "Five constants are proved to match an independent mathematical specification. No arithmetic — addition, multiplication, Montgomery reduction, inversion — is modelled or proved. The SIMD backends are out of scope."

**Fix:** put that three-sentence scope statement at the top of this file, and add a pointer to `baby-bear/proofs/README.md` from `CONTRIBUTING.md`.

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear/p3_field.lean
```diff
@@ -0,0 +1,181 @@
...
+    nothing exploits it. An `opaque` constant asserts only that *some* function
+    of that type exists, which is true and is exactly what "axiomatized
+    interface" is supposed to mean. It also keeps the build warning-clean, so
+    a `sorry` introduced later by drift is immediately visible.
+
+    `#print axioms` on anything downstream reports no axioms from these. That
+    does NOT mean the operations are defined -- they are opaque. It means the
+    assumption is well-formed rather than contradictory. -/
+
+namespace p3_field.field
```
tcoratger, line 117: **[NIT] Doubled namespace: these declarations are `p3_field.field.p3_field.field.ring_add`, not `p3_field.field.ring_add`**

Line 103 opens `namespace p3_field.field` and this line opens it again, so the six `opaque`s below nest one level deeper than intended. I confirmed in Lean 4.30 that `#check @p3_field.field.p3_field.field.ring_add` resolves. It elaborates fine (the instances at lines 144-179 resolve through the open namespace), and the unclosed namespace at EOF is legal, so nothing breaks.

It does make `TCB.md`'s layer-3 table wrong: it names them `p3_field.field.ring_{add,sub,mul,add_assign,sub_assign,neg}`, and the `#print axioms` commands in the audit checklist will not resolve as written.

**Fix:** delete this `namespace` line and the `end` at line 135, or the one at line 103. Then re-check the names quoted in `TCB.md`.

tcoratger (commented):
Adds a hax `legacy-lean` extraction of `p3-baby-bear` under `baby-bear/proofs/`, plus five proved theorems about the field constants, `TCB.md`/`SYNC.md`/`README.md`, six post-extraction patches and a re-extraction script. Zero Rust behaviour changes; `exclude = ["proofs"]` keeps it out of the published crate. I reproduced the test plan locally: `lake exe cache get && lake build` is **green, 1741 jobs, 0 errors, 0 warnings** on `eefe571`, and I independently cross-checked every numeric constant in `baby-bear/src/{baby_bear,mds,poseidon2,poseidon1}.rs` against the extraction (1840 literals > 1000, **0 missing**). The documentation is unusually honest about its own limits — `TCB.md` is the best artifact in the PR. Two things stop this from being an approve: two dependency stubs are given *definitional identity bodies* rather than `opaque`, which makes statements that are false of the Rust provable by `rfl` inside this package (I verified both), and nothing in the tree or in CI would notice if the Rust drifted.

## Rust <-> Lean correspondence

Rust paths relative to repo root; Lean paths relative to `baby-bear/proofs/legacy-lean/`.

| Rust item | Rust file:line | Lean item | Lean file:line | Verdict |
|---|---|---|---|---|
| `PRIME = 0x78000001` | `baby-bear/src/baby_bear.rs:17` | `Impl.PRIME_hoisted = 2013265921` | `extraction/p3_baby_bear.lean:97` | match |
| `MONTY_BITS = 32` | `baby_bear.rs:19` | `Impl.MONTY_BITS_hoisted = 32` | `p3_baby_bear.lean:99` | match |
| `MONTY_MU = 0x88000001` | `baby_bear.rs:20` | `Impl.MONTY_MU_hoisted = 2281701377` | `p3_baby_bear.lean:101` | match |
| `MONTY_MASK` default | `monty-31/src/data_traits.rs:23` | `MontyParameters.MONTY_MASK` | `extraction/p3_baby_bear/p3_monty_31.lean:74` | match (total `UInt64` arith; agrees at `MONTY_BITS = 32`) |
| `to_monty` | `monty-31/src/utils.rs:7` | `utils.to_monty` | `p3_monty_31.lean:185` | match |
| `MontyField31::new` | `monty-31/src/monty_31.rs:51` | `Impl.new` | `p3_monty_31.lean:222` | match; the four `const {}` asserts (`monty_31.rs:52-60`) are dropped |
| `new_array` / `new_2d_array` | `monty_31.rs:88` / `:101` | `Impl.new_array` / `Impl.new_2d_array` | `p3_monty_31.lean:194` / `:206` | match |
| `MONTY_GEN = BabyBear::new(31)` | `baby_bear.rs:28` | `Impl_3.MONTY_GEN_hoisted` | `p3_baby_bear.lean:129` | match |
| `RelativelyPrimePower<7>` | `baby_bear.rs:34` | `Impl_4` at `(7 : u64)` | `p3_baby_bear.lean:171` | match |
| `exp_root_d = exp_1725656503` | `field/src/exponentiation.rs:62` | `exp_1725656503 := pure val` | `extraction/p3_baby_bear/p3_field.lean:75` | **DIVERGE — identity stub** |
| `TWO_ADICITY = 27` | `baby_bear.rs:44` | `Impl_5.TWO_ADICITY_hoisted` | `p3_baby_bear.lean:185` | match |
| `ODD_FACTOR` default | `monty-31/src/data_traits.rs:93` | `TwoAdicData.ODD_FACTOR` | `p3_monty_31.lean:265` | match |
| `BarrettParameters::{N, PRIME_I128, PSEUDO_INV, MASK}` | `data_traits.rs:53-56` | same | `p3_monty_31.lean:133,137,141,144` | match (all operands non-negative, so the `Int` division convention is immaterial) |
| `EXT_TWO_ADICITY` at `DEG = 4/5/8` = `29/27/30` | `baby_bear.rs:71/88/98` | `Impl_6/7/8` = `29/27/30` at `DEG 4/5/8` | `p3_baby_bear.lean:319/401/470` | match, degrees line up |
| `MAX_SINGLE_SAMPLE_BITS = 27` | `baby_bear.rs:110` | `Impl_9.MAX_SINGLE_SAMPLE_BITS_hoisted` | `p3_baby_bear.lean:528` | match |
| `SAMPLING_BITS_M` (`const while`) | `baby_bear.rs:111-131` | `Impl_9.SAMPLING_BITS_M_hoisted` | `p3_baby_bear.lean:530-571` | match; obligation closed by `native_decide` |
| `first_row_to_first_col` | `mds/src/util.rs:52` | `first_row_to_first_col := pure v` | `extraction/p3_baby_bear/p3_mds.lean:14` | **DIVERGE — identity stub**, poisons all six `MATRIX_CIRC_MDS_*_COL` |
| MDS / Poseidon2 / Poseidon1 constant tables | `baby-bear/src/{mds,poseidon2,poseidon1}.rs` | `p3_baby_bear.lean` | — | match: all 1840 literals present, checked mechanically |
| `monty_reduce` | `monty-31/src/utils.rs:105` | — | — | **not modelled** (out of scope, declared) |
| `add` / `sub` with the conditional corrections | `utils.rs:63` / `:81` | `opaque ring_add` / `ring_sub` | `p3_field.lean:123` / `:125` | **not modelled**, honestly `opaque` |
| `mul`, `neg`, `try_inverse`, the inverse exponentiation chain | `monty-31/src/monty_31.rs` | `opaque ring_mul` / `ring_neg`; inverse absent | `p3_field.lean:127,133` | **not modelled**, honestly `opaque` |
| AVX2 / AVX-512 / NEON backends | `baby-bear/src/{x86_64_avx2,x86_64_avx512,aarch64_neon}` | — | — | **out of scope**: extracted for `thumbv7em-none-eabi` |

Net: every constant is faithful. The arithmetic is not modelled at all — which the PR says plainly — and the two identity stubs are the only places where the Lean asserts something the Rust does not.

## What is actually proven

Five theorems in `spec/p3_baby_bear_proofs/constants.lean`, all discharged, **no `sorry`, no `admit`, no `axiom`** anywhere in the tree (I grepped; the word `sorry` appears only in `p3_field.lean` prose explaining why it was avoided).

| Theorem | Line | Establishes | Real content? |
|---|---|---|---|
| `monty_prime_eq_fieldSize` | `constants.lean:23` | extracted `PRIME = 2013265921` equals CompPoly's `2^31 - 2^27 + 1` | **yes** — independent yardstick |
| `two_adicity_eq_spec` | `:28` | extracted `TWO_ADICITY = 27` equals CompPoly's `twoAdicity` | **yes** |
| `monty_bits_eq_thirtyTwo` | `:33` | `MONTY_BITS = 32` | no — `32 = 32` after unfolding |
| `coprime_seven_pred_prime` | `:43` | `gcd(7, p - 1) = 1`, so `x -> x^7` permutes the unit group | **yes**; `decide` here is a 4-step Euclid, I confirmed it elaborates in Lean 4.30 in well under a second — not a large-domain `decide` |
| `fieldSize_sub_one_factorization'` | `:50` | `p - 1 = 2^TWO_ADICITY * 15` | partly — relates two *extracted* constants, odd part `15` supplied by hand |

Escape-hatch audit: one `native_decide` (`p3_baby_bear.lean:571`, adds `Lean.ofReduceBool`), fifteen `opaque` (eight behavioural, seven structural — the `TCB.md` count is accurate), zero `sorry`/`axiom`/`unsafe`/`@[implemented_by]`. `#print axioms` on the five theorems shows nothing beyond `propext`/`Classical.choice`/`Quot.sound`, as `TCB.md` claims.

## For the maintainers, not the author

- **Does the repo want to carry Lean at all?** That is a policy call, not a code review. @Nashtare already raised the packaging concern and the author fixed it (`exclude = ["proofs"]`). The remaining question is ownership: this is ~6.9k lines of a language nobody on the Rust side reviews, in the tree of a crate that changes often.
- **The CI story is "none."** No workflow references `lean` or `proofs`; the Rust build, `clippy`, `release-plz` and the workspace are untouched, and all 15 checks pass. So merging is *safe*, but the directory is inert.
- **The honest minimum CI is small.** A `paths: ['baby-bear/proofs/legacy-lean/**']`-triggered job running `lake exe cache get && lake build` costs about 6 minutes warm (my clean run was dominated by the mathlib cache download; the package itself is ~30s of the 1741 jobs). Without it, nothing catches a broken `lake build`, let alone drift.
- **Drift risk today is real but the extracted surface has not moved.** The branch is 68 commits behind `main`. Since the merge-base, `baby-bear/src` changed only in `x86_64_avx2/packing.rs` and `x86_64_avx512/packing.rs` — both `cfg`-excluded by the `thumbv7em-none-eabi` extraction target — and `monty-31/src/{data_traits,monty_31,utils}.rs`, `mds/src/util.rs` and `field/src/exponentiation.rs` are unchanged. So the extraction is still accurate; it merges cleanly. Rebase and it is current.
- **Scope precedent.** If this lands, the next PR is `p3-monty-31`, then `p3-field`. Worth deciding now whether `<crate>/proofs/` is the layout you want repo-wide, and whether `TCB.md` is a required artifact for each.

## Questions / could not verify

- I did not run `build-proofs.sh`: `cargo-hax` and the pinned OCaml engine are not installed here, so **I cannot confirm `extraction/p3_baby_bear.lean` is reproducible from `baby-bear/src`**. Everything I checked about it, I checked by reading it against the Rust. The `-Zcrate-attr=feature(maybe_uninit_slice)` nightly flag and the hax pin are taken on trust.
- The hax `legacy-lean` backend is upstream-labelled legacy and experimental. Its Rust-to-Lean semantics (the `RustM` monad, `RustArray`, the `core_models.*` shadows) are the largest unaudited trust layer here and I did not review the Hax proof library itself.

**Bottom line:** do not merge as-is — but the two blocking-quality issues are a one-word fix each (`def ... := pure v` -> `opaque`). Fix those, add a `lake build` CI job, and this is a good addition.

dhsorens:
thanks @tcoratger for the comments - good catch on some of the axiomatized interfaces. will address this week

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear/p3_field.lean
dhsorens: Fixed in 9da1a1f8 — `def … := pure val` → `opaque`, exactly as you suggested:

```lean
opaque exp_1725656503
    (R : Type)
    [_trait_constr_exp_1725656503_associated_type_i0 :
      p3_field.field.PrimeCharacteristicRing.AssociatedTypes R]
    [_trait_constr_exp_1725656503_i0 : p3_field.field.PrimeCharacteristicRing R]
    (val : R) :
    RustM R
```

Your `sbox_is_identity` no longer typechecks (`error: Not a definitional equality: the left-hand side`), and `#print axioms` is unchanged on the five theorems, so nothing downstream regressed.

`TCB.md` gains a row for it in the layer-3 assumed-operations table (now 9 behavioural + 7 structural = 16 `opaque`). The "Two infidelities" section is gone, since neither infidelity remains.

One detail worth carrying over to the sibling finding: `opaque` is available here **because** this symbol's only call site (`p3_baby_bear.lean:163`) is a plain `do` block with no `of_isOk` obligation. That turned out not to be true of `first_row_to_first_col` — see my reply there.

baby-bear/proofs/legacy-lean/extraction/p3_baby_bear/p3_mds.lean
dhsorens: Fixed in 0a57849e — taking your second suggestion (transcribe) rather than the first, because `opaque` turns out not to be available here at all.

I tried `opaque` first and it does not build. All six calls sit under `RustM.of_isOk _ (by rfl)` (`p3_baby_bear.lean:618,632,650,672,702,740`), so each obligation is a proof that the call returns `.ok`. Since `RustM α = Option (Except Error α)`, an opaque inhabitant could equally be `div` or `fail`: `isOk` does not reduce through an opaque head, and the obligation goes from provable to unprovable. Six errors, one per constant. `native_decide` doesn't rescue it either — the evaluator can't unfold an opaque.

So I think the finding is slightly sharper than "a `def` asserts where an `opaque` would not". At an `of_isOk`-gated site hax's output *requires* a total body, and the defect was a **wrong total body where a right one was mandatory** — the same slot `to_monty` and `Impl.new` fill correctly.

The body:

```lean
pure (.ofVec (Vector.ofFn (fun i => v.toVec[-i])))
```

`Fin.neg` is exactly upstream's index map — `i ↦ (N - i) % N`, i.e. `0 ↦ 0` and `i ↦ N - i` — so this needs no dependent `if` and no proof term, and stays plain `Nat` arithmetic, which is what keeps all six `rfl` obligations discharging by kernel reduction. `lake build` green: 1741 jobs, ~29s warm, no warnings beyond the two pinned upstream `USize64` lints.

Verified rather than eyeballed: I parsed the six row literals out of `baby-bear/src/mds.rs`, applied `col[0] = row[0]; col[i] = row[N - i]` outside Lean, and diffed against `#eval` of each extracted constant. All six agree. `MATRIX_CIRC_MDS_8_COL` is now `[7,9,4,3,8,8,3,1]`; it was `[7,1,3,8,8,3,4,9]`. Your `mds_col_is_row` no longer typechecks.

Worth flagging for your constant audit: this is why "1840 literals, 0 missing" passed. That check was for presence, and a permutation preserves presence — so it could not have caught this.

Two things I recorded in the docs rather than fixed, because they are the honest position:

- `TCB.md` now lists this under "given real, faithful bodies" with **checked by nothing**, and spells out that the six `of_isOk` obligations constrain the body to be *total and kernel-reducible*, not to be the *right permutation* — the identity body discharged all six too. That is your drift finding exactly, so I have left that thread open rather than claiming it closed here.
- `SYNC.md`'s runbook step that said "if something genuinely must be assumed, declare it `opaque`" now carries the `of_isOk` caveat, so the next person doesn't repeat what I just did.

I also corrected the `of_isOk` counts while in there: `TCB.md` read "45 of 46", but there are 47 sites — 46 by `rfl`, 1 by `native_decide`. The `SYNC.md` recount command anchored on `^\s*(by rfl)$` and silently missed an inline obligation at `p3_baby_bear.lean:408`.

dhsorens: Note - when the mds crate extracts, we'll import that extraction directly and we won't have to model it as we have done here

baby-bear/proofs/legacy-lean/TCB.md
dhsorens: Partly addressed in 39929b17 — the stub-fidelity half. The CI half is still open, deliberately.

## What I did not do, and why

Your suggested fix was a `lake build` job on `paths: ['baby-bear/**']`. I don't think that closes the failure mode you described. `lake build` catches *Lean* breakage; it will stay green forever while the Rust drifts, because a stale-but-well-typed constant still elaborates. That's exactly the "decays into a wrong one that still looks green" scenario — a `lake build` job would have been green throughout the entire life of the `first_row_to_first_col` bug.

Restoring `SanityCheck.lean` has a related problem. A Lean file asserting `to_monty … = 1234` by `rfl` pins the *Lean* against a number someone typed; it also stays green when the Rust moves. It pins one side of a two-sided invariant.

## What I did instead

[`check-transcriptions.py`](https://github.com/Plonky3/Plonky3/blob/39929b17/baby-bear/proofs/legacy-lean/check-transcriptions.py): reads the Rust, computes what each constant must be, `#eval`s the Lean, diffs. 28 checks over 7 of the 9 layer-3 rows. Each is labelled by strength, because they are not worth the same:

- **`[rust]`** (11) — expected value parsed straight out of the Rust source. This is the one that detects drift.
- **`[formula]`** (17) — the script re-implements a Rust formula (`to_monty`, `PSEUDO_INV`, …). Catches Lean-side drift and two transcriptions disagreeing, but both could be wrong the same way. Strictly weaker, and labelled as such in the output rather than being quietly counted the same.

The two rows it can't reach (`Dup.dup`, `hax_ext AsRef.as_ref`) have no evaluable data; the script prints them as uncovered rather than omitting them.

Regression-tested in both directions, since a check that can't fail is worth nothing:

```
$ ./check-transcriptions.py          # with pure v restored
  [   rust] MATRIX_CIRC_MDS_8_COL         FAIL
              expected: 7,9,4,3,8,8,3,1
              from Lean: 7,1,3,8,8,3,4,9
  … all six fail

$ ./check-transcriptions.py          # after changing a literal in baby-bear/src/mds.rs
  [   rust] MATRIX_CIRC_MDS_8_COL         FAIL
              expected: 7,99,4,3,8,8,3,1
              from Lean: 7,9,4,3,8,8,3,1
```

Worth reporting because it bears on your finding: **my first version passed with the bug reintroduced.** `lake env lean` resolves `import p3_baby_bear` to the compiled olean, so the probe measured the previous build and never saw the edit. A checker reporting green against code it hadn't read — the same shape as the problem it was written for. It now builds unconditionally first, and that reason is recorded in the script, in `SYNC.md` 3a and in the audit checklist so it can't quietly regress.

## Docs

`TCB.md`'s faithful-bodies table is now framed as a **worklist** with a `Retired by` column: each hand transcription is a stand-in for an extraction that hasn't happened, discharged when the owning crate is extracted and consumed as a Lake dependency (`p3-monty-31` retires six rows, `p3-mds` one, `p3-field` one). It also now says explicitly that this relocates the trust boundary outward rather than removing it — monty-31's own extraction will have its own stubs.

`SYNC.md`'s step 2 used to say "read the Rust and the Lean side by side". That instruction was in the runbook for the entire life of the bug, so I've replaced it with running the script rather than adding more prose next to prose that already failed. The re-extraction trigger was also wrong in a way your comment implies — it said a dependency *body* change is invisible here; it isn't, for the transcribed set, so it now says to run the check after any `monty-31/src` or `mds/src` change even when no signature moved.

Also fixed an error I introduced in the previous commit: I'd written that extracting `p3-monty-31` retires `first_row_to_first_col`. It's `p3-mds` that owns it.

## What's still open — your finding stands

Nothing runs this automatically. It is a script a human or an agent invokes, so it detects drift only when someone remembers, and the `[formula]` rows remain two transcriptions agreeing rather than a check against the Rust. I'd rather have that stated plainly than have a green CI badge that proves less than it appears to. Whether the repo wants a Lean job at all still seems to me a maintainer call rather than something to settle in this thread.

baby-bear/proofs/legacy-lean/spec/p3_baby_bear_proofs/constants.lean
dhsorens: added the two theorems suggested in commit 351c3e4e

baby-bear/proofs/legacy-lean/README.md
dhsorens: this will be good for discussion later today :)

baby-bear/proofs/legacy-lean/build-proofs.sh
dhsorens: this will be a good item for discussion, too

baby-bear/proofs/legacy-lean/TCB.md
dhsorens: can discuss today :)

baby-bear/proofs/legacy-lean/.gitignore
tcoratger, line 1: Maybe that should be in the general gitignore of the repo no?

dhsorens: good call - done in 714fdaa

dhsorens:
thanks @tcoratger @Nashtare for the call! I'm going to temporarily move this to a draft PR, I want to make sure this spec is useful and assess any changes to this that I want to make. I'll re-open when ready again, including a plan for future PRs

PR #2155, opened 2026-09-16
perf(binary-dft): gather runs of adjacent rows into the staging tile

tcoratger:
The staging tile was disabled for any row shorter than a cache line — which is exactly the width-1 matrix `p3-binary-pcs` commits.

## The problem

`PolyBasisNtt` fuses its long-stride stages through a staging tile: it gathers the rows a run of stages is closed under, runs the whole run in a per-worker buffer, and scatters back. The matrix is then read once and written once per *group* instead of once per stage.

`Plan::new` switched that off whenever a row was shorter than a cache line:

```rust
let row_fills_a_line = element * width >= STAGED_ROW_BYTES;   // 64
depth: if row_fills_a_line { .. } else { 0 }
```

The reasoning was sound as far as it went — a gather addressing **one row** at a power-of-two stride pulls a whole 64-byte line to move `element * width` useful bytes, so at width 1 three quarters of every line is wasted.

But `PcsLayout` uses `FOLDING = 0`, so `Layout::commit` produces a single width-1 codeword column. The production shape was the one shape that got nothing.

## The idea

The waste comes from gathering one row per strided address, not from anything in the algorithm.

A gather now moves a run of `L = 2^log_block` **adjacent** rows per address, which is exactly the matrix read as rows of `L * width` elements:

```text
    width * 2^log_n  =  (L * width) * 2^(log_n - log_block)
```

Stage `j` pairs rows `2^j` apart, which is `2^(j - log_block)` reshaped rows apart, so for `j >= log_block` it is a stage of the reshaped matrix.

**The reshape is exact because a block index is the same number in both readings.** Sub-layer `s` needs the twiddle of the block a pair lies in:

```text
    matrix row    r                lies in block  r >> (top - s)
    reshaped row  r >> log_block   lies in block  r >> log_block >> (top - log_block - s)
```

A shift by `log_block` then by `top - log_block - s` is a shift by `top - s`. So the `L` rows of one run carry one twiddle between them at every sub-layer, and the tile's twiddle walk is unchanged.

## Choosing the run length

The tile has a fixed byte budget, so doubling the run halves the rows it holds: run length and fused depth trade one for one. A run must cover a cache line, and past that it grows towards a target of 1024 bytes, which a sweep over runs of 16 to 4096 bytes puts at the knee.

Growth is capped by two counts, both measured against what the shortest run already costs:

- **Traversals of the matrix.** A longer run may never add one.
- **How many of those traversals gather.** A leftover stage sweeps the matrix contiguously, while a staged group also gathers and scatters, so trading the leftover for another group holds the traversal count and still adds a gather.

The second count is there because the first alone was not enough. At width 16 and `2^16` rows, equal traversals hid a real trade: main's one group of eight stages plus a plain pass became two groups of six and three, which @Nashtare measured at 3 to 7% slower on Graviton5, growing with the thread count.

**But refusing every added gather was too blunt**, as @Nashtare then pointed out: the gather count only binds when the shortest run leaves a stage over, and there it blocks growth outright, returning 146 shapes at widths 4…63 to the one-row plan. Two of those are shapes the run-length table itself measures, so the code was refusing the plan its own table picks — width 4 at `2^20` sat on 64-byte runs at 5.74 ms with 1024-byte runs measured at 4.40 ms.

What pays for an added gather turns out to be the length the run gains, not the depth of the group that buys it. Two groups of five stages that grow the run 4x measure 1.008 and 0.979, while one that grows it 16x measures 0.780. Forward transforms with the plan forced from one binary, arms alternated, min of 7, as the ratio of the grown plan to the plan that refuses to grow:

| growth | 32 workers | one pinned worker |
|---|---|---|
| 2x | 1.02 1.07 1.12 1.13 1.02 | 1.03 1.07 1.05 1.04 |
| 4x | 0.98 1.00 1.00 1.01 1.01 | 1.04 0.96 1.00 1.08 |
| 8x | 0.93 0.98 | 0.90 0.91 0.90 0.87 0.94 0.96 0.96 0.95 |
| 16x | 0.78 0.80 0.80 0.85 0.78 0.80 0.80 | |

A third arm re-runs the first on every shape as a null control: 1.031 at 32 workers, 1.000–1.013 pinned. So one doubling loses, two are a wash, and three already pay everywhere.

`STAGED_GATHER_DOUBLINGS = 3` is the resulting rule: an added gather is admitted once the run has grown eightfold. The traversal budget still allows at most one added group, so a run that has paid buys exactly one gather. This changes 15 shapes over widths 1…1024 and `log_n` 0…24, every one of which is in the table above — widths 1…7 at 16x, and width 8 at `2^18` plus widths 9…15 at `2^16` at 8x. Width 16 at `2^16` is offered two doublings, so it keeps main's plan and the Graviton result stands.

## What changes for rows main already staged

**This is not a strictly additive gate, and the earlier revisions of this description were wrong to say so.** Run growth applies to every row, so a row main already staged can get a different plan. At width 16, as `(local, log_block, depth)` with the group sizes:

| `ntt/128/hybrid` | main | this PR |
|---|---|---|
| 2^14 | (7, 0, 8): 7 | (7, 1, 7): 7 |
| 2^16 | (7, 0, 8): 8, plus 1 plain | (7, 0, 8): 8, plus 1 plain |
| 2^18 | (7, 0, 8): 8, 3 | (7, 2, 6): 6, 5 |
| 2^20 | (7, 0, 8): 8, 5 | (7, 1, 7): 7, 6 |

At width 16 the run can never grow more than fourfold, so no height here buys an added gather: `2^18` and `2^20` are rebalanced at equal gather count, and `2^16` is refused and returns to main's plan exactly.

What *is* additive is the narrow-row gate itself, which is the only reader of `STAGED_WORKERS`:

```text
    row >= STAGED_LINE_BYTES   gathered at any worker count      (what main already did)
    row <  STAGED_LINE_BYTES   gathered from STAGED_WORKERS up    (new behaviour only)
```

The plan choice is pinned by test rather than by sampling. `the_plan_gathers_the_longest_run_its_budgets_allow` asserts the `(local, log_block, depth)` triple for eleven production shapes plus seven pinned by triple alone, and asserts the plan never takes more traversals than a run of one row over widths 4…1024 and `log_n` 0…24.

Those seven bracket the run-length rule from both sides — one doubling refused, two refused, three admitted, four admitted, and four again from a run floor above one row — plus the two width-16 heights where there is no leftover pass to trade and the traversal count alone picks the run. Setting `STAGED_GATHER_DOUBLINGS` to 0, 1, 2, 4, 5 or effectively infinity each fails that one test, as does dropping the exemption clause.

## Benchmarks

Ryzen 9 9950X3D, `RUSTFLAGS="-C target-cpu=native"`, matched builds in separate target directories per side and feature set, serial runs pinned with `taskset -c 0-7`.

Serial transform, width 16 (`ntt/128/hybrid`), where 2^20 is 2^24 elements = 256 MiB:

| | before | after | |
|---|---:|---:|---:|
| 2^14 | 2.193 ms | 2.133 ms | −2.8% |
| 2^16 | 9.530 ms | 9.321 ms | −2.2% |
| 2^18 | 43.28 ms | 41.26 ms | −4.7% |
| 2^20 | 206.8 ms | 192.1 ms | **−7.1%** |

End to end, `commit` with `--features parallel` on 32 threads:

| | before | after | |
|---|---:|---:|---:|
| 16 | 3.158 ms | 3.097 ms | −1.9% |
| 18 | 10.047 ms | 9.629 ms | −4.2% |
| 20 | 39.718 ms | 35.941 ms | **−9.5%** |

@Nashtare's Graviton5 measurements (96 Neoverse-V3 cores) are the more thorough set, and the width-1 wins there are much larger: `poly/forward/h16/w1` is 3.77x at 32 threads and 5.29x at 96, and `commit/18` and `commit/20` are −15.3% and −14.2% at 96 threads.

Serial `commit` is unchanged, and provably so rather than measurably: `current_num_threads()` is a `const fn` returning 1 without the `parallel` feature, so the plan const-folds to `depth = 0` and the serial path executes the identical pre-change schedule at width 1.

## Where it costs something

On the Ryzen, at 8 workers, width 1, 16–32 MiB, staging loses 4–14%. **That band is machine-specific and did not reproduce on Graviton5**, where @Nashtare measured no regression at 2, 4 or 8 threads. The worker count is a coarse proxy for cache residency, and closing such a band would need an actual last-level-cache size, which a `no_std` crate cannot query. The response is also non-monotonic in size — staging wins small, loses in the cache-resident valley, wins large — so no byte threshold separates the regimes either.

A single worker on a narrow matrix that genuinely outgrows its cache also forgoes a win it could have had. That is a win forgone rather than a cost added.

## Correctness

The additive NTT is a commitment primitive, so the schedule is pinned against an oracle that shares none of its identities.

- Against `NaiveAdditiveNtt`, element for element: widths 1…8 plus 16 and 64, `log_n` 0…6, forward and inverse, zero and non-zero shift. The oracle costs `O(n^2 log n)` per column, so no production plan's first staging group sits inside that range; `a_staged_cut_matches_the_reference_transform` reaches the oracle with synthetic cuts instead, including one with no contiguous-tile stage and no leftover, where every butterfly goes through a gather and a scatter.
- Against a plain per-stage schedule over widths 1…64 up to `log_n` 14, at both worker readings so every shape is checked in both plans.
- `binary-dft/tests/commit.rs` pins the encoder's Merkle root against the reference transform, now including the default encoder and the width-1 codeword.
- The row handle addresses runs rather than rows; its hard bounds check and the partition assertion are kept, and the safety argument is rewritten around the run-to-row partition.
- `Plan::groups` is the single source for the cut, read by the forward walk, the inverse walk and the traversal budget alike, so none of the three can drift from the others.

Perturbing the twiddle walk (`block << s` -> `(block << s) ^ 1`) fails 14 lib tests and `commit_base_matches_hand_encoding` on a build whose target has carryless multiply. On a bare `x86-64` baseline it fails 3, because `commit.rs` reaches a different transform there — which is also why the `linux, pclmulqdq` and `macos, parallel` legs are the ones that run that file through `PolyBasisNtt`.

## Verification

- `cargo test -p p3-binary-dft` and `-p p3-binary-pcs`, both feature configurations, and once more with `-C target-cpu=native` — all green.
- `cargo clippy -p p3-binary-dft --all-targets` clean in both; `cargo fmt` applied.
- Rebased onto `main`. The seven conflicts in `poly.rs` were all doc comments beside this PR's constants; #2153's `Conversion` change is preserved, including at the new `fused_stages` call sites, and main's one-sentence-per-line reflow is kept.


binary-dft/src/poly.rs
```diff
@@ -250,27 +301,69 @@ struct Plan {
...
+        let local = log2_floor_usize((TILE_BYTES / row).max(1)).min(log_n);
+        let above = log_n - local;
+        // Stages one tile fuses, with a run of `2^log_block` adjacent rows as its row.
+        let depth_at =
+            |log_block: usize| log2_floor_usize((STAGING_BYTES / (row << log_block)).max(1));
+        let floor = log2_ceil_usize(STAGED_LINE_BYTES.div_ceil(row));
+        let target = log2_ceil_usize(STAGED_RUN_BYTES.div_ceil(row)).max(floor);
+        // The shortest admissible run sets the traversal count no longer run may exceed.
+        let budget = Self::traversals(above, depth_at(floor));
+        let log_block = (floor..=target)
```
Nashtare, line 418: Run growth applies to every row under `STAGED_RUN_BYTES` at every worker count, so rows of 64+ bytes that main already staged get a new plan too, serial included. Width 16 goes from depth 8 to `log_block` 2 / depth 6 at 2^16 and 2^18, and to `log_block` 1 / depth 7 at 2^14 and 2^20. Across widths 4..63, 792 of the 1080 shapes where main fused a group change; only the traversal count stays equal. Should growth be limited to `row < STAGED_LINE_BYTES`, so wider rows keep main's plan exactly? Or should the description drop "no shape main already staged can change"?

binary-dft/src/poly.rs
```diff
@@ -1026,6 +1209,163 @@ mod tests {
...
+        //
+        // So no shape one-row gathering already staged comes out with more passes than it had.
+        for width in [4usize, 5, 8, 16, 64, 512, 1024] {
+            for log_n in 0..=24 {
+                let plan = Plan::for_workers(width, log_n, 32);
+                let row = core::mem::size_of::<u128>() * width;
+                let one_row = log2_floor_usize((STAGING_BYTES / row).max(1));
+                let above = log_n - plan.local;
+                assert!(
+                    Plan::traversals(above, plan.depth) <= Plan::traversals(above, one_row),
```
Nashtare: Both sides go through `Plan::traversals`, and for rows of 64+ bytes `one_row` equals `depth_at(floor)`, so this restates the `find` predicate. With `traversals` mutated to return `above`, all 58 `p3-binary-dft` tests still pass while 738 shapes at widths 4..63 gain a pass (width 16 at 2^20 goes from 2 to 3, the regression the budget exists to prevent). `log_block: 0` and `STAGED_RUN_BYTES = 64` survive too. Could this count passes independently, e.g. by simulating the `forward` loop, and pin the `(local, log_block, depth)` triples for a few shapes?

binary-dft/src/poly.rs
```diff
@@ -198,31 +198,83 @@ const TILE_BYTES: usize = 32 * 1024;
...
+///
+/// ```text
+///     row >= STAGED_LINE_BYTES   gathered at any worker count
+///     row <  STAGED_LINE_BYTES   gathered from STAGED_WORKERS workers up
+/// ```
+///
+/// A lone worker whose narrow matrix does outgrow its cache gives up a win it could have had.
+/// That is a win forgone rather than a cost added, which is the safe direction for a signal
+/// this coarse.
+const STAGED_WORKERS: usize = 2;
```
Nashtare: The doc table has 1 worker at 1.12/1.23 for 16/32 MiB, and the description says 8 workers still lose 4–14% in that band; only 32 workers win. By the mechanism described, 2 and 4 workers sit between those rows, and a threshold of 2 stages narrow rows for all of them. Were 2, 4 and 8 workers measured across 4–128 MiB? If the band holds there, would a higher threshold keep the 32-worker wins?

binary-dft/tests/commit.rs
```diff
@@ -43,30 +43,39 @@ fn table() -> Table<F> {
 /// with the reference transform. This pins the message layout as well as the transform.
 #[test]
 fn commit_base_matches_hand_encoding() {
+    // A folding depth of zero leaves a width-1 message.
+    //
+    // That is the shape `p3-binary-pcs` commits.
+    //
+    // It is also the one whose transform gathers the most rows per strided address.
```
Nashtare: This test encodes through `AdditiveRsEncoder::<F, LchNtt<F>>` (line 58), which has no staging tile, so the new `folding = 0` case never reaches `PolyBasisNtt`. Even through the default encoder, a width-1 message at 2^8 has `local == log_n` and never stages. With the twiddle walk perturbed (`(block << s) ^ 1`), both tests in this file still pass. Should this use the default encoder at a height that stages, or should the claim come out here and in the description?

binary-dft/src/poly.rs
```diff
@@ -828,41 +959,73 @@ mod tests {
...
+    ///
+    /// The production plan of a width-1 matrix fuses its first group at `2^13`.
+    /// So a sweep has to reach past that before it exercises a staged gather at all.
+    const MAX_LOG_N: usize = 14;
+
+    /// The tallest matrix the reference transform is asked for.
+    /// It costs `O(n^2 log n)` per column.
+    ///
+    /// So `2^7` rows is already seconds of an unoptimised test run.
+    const REFERENCE_LOG_N: usize = 6;
```
Nashtare, line 1169: The description says the `NaiveAdditiveNtt` sweep covers `log_n` 0…14, but it stops at 6 here. No production plan for any width in `ORACLE_WIDTHS` fuses a group at 2^6 or below, and the widened proptests (width ≤ 8, `log_n` ≤ 8) don't stage either. So the staged path is pinned against `per_stage_schedule` and `LchNtt` only. That chain holds (I also checked synthetic staged plans against the oracle), but the description should say so.

binary-dft/src/poly.rs
```diff
@@ -1026,6 +1209,163 @@ mod tests {
...
+        //
+        // The cut points move with the height.
+        //
+        // So the two are swept together against the per-stage schedule.
+        for width in ORACLE_WIDTHS {
+            for log_n in 0..=MAX_LOG_N {
+                // One worker gathers no tile where many workers do.
+                //
+                // So both readings of one shape are swept, not just this build's own.
+                for workers in [1, 32] {
```
Nashtare, line 1336: Nit: this test took 157 s on its own in a debug build here (on a loaded machine). Width 64 is over half the cost at the top heights, and its plan is `log_block = 0` at every height, the path the cut tests already cover. Could width 64 stop lower?

binary-dft/src/poly.rs
```diff
@@ -198,31 +198,83 @@ const TILE_BYTES: usize = 32 * 1024;
...
+///
+/// A stride that jumps every line still leaves the prefetcher nothing to follow.
+///
+/// A sweep over runs of 16 to 4096 bytes, at equal element volume on 32 threads:
+///
+/// ```text
+///     run bytes      16     64    256   1024   2048   4096
+///     width 1      7.28   6.32   5.93   5.21   5.76   5.62   ms at 2^22 rows
+///     width 4      5.71   5.74   5.23   4.40   4.88   4.94   ms at 2^20 rows
+///     width 16     4.91   4.73   4.96   4.04   4.72   4.82   ms at 2^18 rows
```
Nashtare, line 233: Nit: a run can't be shorter than one row, so width 4 at 16 bytes and width 16 at 16, 64 and 256 bytes all run the same `log_block = 0` plan. Width 16's three copies read 4.91, 4.73 and 4.96 ms, so ~5% is the noise the other cells should be read against.

binary-dft/src/poly.rs
```diff
@@ -198,31 +198,83 @@ const TILE_BYTES: usize = 32 * 1024;
 /// The tile is streamed rather than randomly addressed, so it need not fit in L1.
 /// What it must not do is spill out of the private cache level below the shared one.
 ///
-/// A sweep of the fusion depth at width 16 puts the optimum at 64 KiB, which fuses 8 stages.
+/// A sweep of the fusion depth at width 16 puts the optimum at 64 KiB, which fuses 6 stages.
```
Nashtare: Nit: the depth at 64 KiB now depends on height (width 16 fuses 6 at 2^16, 7 at 2^14 and 2^20, 8 at 2^22), so "fuses 6 stages" holds only for some shapes.

binary-dft/src/poly.rs
```diff
@@ -250,27 +301,69 @@ struct Plan {
...
-    /// The cut points a matrix of this shape gets, from the two cache budgets.
+    /// Full traversals of the matrix the stages above the contiguous tile take.
+    /// Each fused group is one traversal.
+    ///
+    /// Each leftover stage is one of its own.
+    fn traversals(above: usize, depth: usize) -> usize {
+        if depth < 2 {
+            return above;
+        }
+        let leftover = usize::from(above % depth == 1);
```
Nashtare: Nit: this re-derives the leftover rule from `Plan::leftover`, and `forward`/`inverse` derive the group count from their own loops. They agree today, but the surviving `traversals` mutant shows no test would catch them drifting; building both from one helper would.

Nashtare (commented):
Reviewed at 1ef30d59; f6d45b8e only reflows doc comments and fixes the rustfmt import order, so all of this applies to the current head. The line numbers below are f6d45b8e's.

**What holds**
- **The reshape is correct.** Write `r = R·2^log_block + i`. Then sub-layer `s`'s block `r >> (top − s)` equals `R >> (top − log_block − s)` exactly when `log_block ≤ top − depth`, which is the clamp in `fused_stages`. `(index, k) → first + k·stride` is a bijection onto the runs, so tasks write disjoint ranges. The bounds check is a hard `assert!`.
- **Synthetic plans.** I checked 11,880 of them (every `local`/`log_block`/`depth` up to 2^8, including `log_block > top`). They match a per-stage reference whose twiddles come from `subspace_polynomial` + `domain_point`, with and without the folded conversions.
- **Production plans.** I checked 1,425 more (widths 1..17, width 1 up to 2^18, at 1/2/8/14/32 workers). They also match, serially and under `parallel` at 2 and 14 threads.
- **Public API.** Checked against `LchNtt` (ntt, intt, padded and lde, up to 2^20 at width 1): the output is identical at 1, 2, 8 and 14 threads.
- **Miri** passes on the narrow staged path, serially (Stacked Borrows) and on 2 threads (Tree Borrows).
- **Mutations.** The gate, clamp and staged-twiddle mutants are all killed.
- Tests pass in both feature sets, and clippy `-D warnings` is clean.

**Concerns** (details inline)
1. **"No shape main already staged can change" doesn't hold.** Run growth applies to every row under 1024 bytes at every worker count, serial included. Widths 4..63 get a new `log_block`, depth and group split in 792 of the 1080 shapes where main fused a group; only the traversal count is preserved. Width 16 goes from depth 8 to `(log_block, depth)` = (2, 6) at 2^16/2^18, and to (1, 7) at 2^14/2^20.
2. **No test pins the plan choice.** Each of these passes every test: `Plan::traversals` returning `above`, `log_block: 0`, `STAGED_RUN_BYTES` at 64 or 4096, and `floor..target`. The first adds a full pass to 738 shapes.
3. **`STAGED_WORKERS = 2` sits below the band the description says loses.** There are no 2- or 4-worker numbers.
4. **The Correctness section claims more coverage than the tests have.** `tests/commit.rs` runs `LchNtt`, and the `NaiveAdditiveNtt` sweep stops at 2^6, below any staged production plan.

This conflicts with #2153 in `binary-dft/src/poly.rs`.

Nashtare (changes requested):
Requesting changes for a measured width-16 regression, which follows from concern 1 in my earlier review: run growth also changes plans for rows main already staged. I benchmarked f6d45b8e against main 463b44c3 on Graviton5 (`c9g.24xlarge`, 96 Neoverse-V3 cores, one NUMA node, `-C target-cpu=native`). Each comparison ran twice, alternating base and head. Serial runs were pinned to one core, and parallel runs to cores `0..T`. Numbers are criterion medians averaged over both rounds. The two rounds agree within ±2% on every id below.

**Width 1, the shape this PR targets, wins clearly:**

| bench | main | this PR | change |
|---|---:|---:|---:|
| `commit/16 / 18 / 20`, 96 threads | 3.56 / 8.15 / 21.13 ms | 3.24 / 6.91 / 18.12 ms | −8.9% / −15.3% / −14.2% |
| `commit/16 / 18 / 20`, 32 threads | 3.48 / 10.61 / 37.40 ms | 3.36 / 10.15 / 36.13 ms | −3.5% / −4.3% / −3.4% |
| `commit/*`, 2, 4 and 8 threads | | | within ±1.3% |
| `poly/forward/h16/w1`, 32 threads | 322.2 µs | 85.5 µs | 3.77x |
| `poly/forward/h16/w1`, 96 threads | 927.1 µs | 175.4 µs | 5.29x |

At 2, 4 and 8 threads there is no regression on this machine either: the loss band from the description doesn't show up here.

**Width 16 (`ntt/128/hybrid`, WHIR's folding-block width) regresses where it shouldn't change at all:**

| `ntt/128/hybrid` | 2^14 | 2^16 | 2^18 | 2^20 |
|---|---:|---:|---:|---:|
| serial | +0.3% | **+2.9%** | −2.0% | −0.7% |
| 32 threads | −2.1% | **+7.4%** | −4.9% | +0.1% |
| 96 threads | **+12.1%** | **+10.2%** | −19.1% | −1.0% |

Main already staged these rows: 256 bytes ≥ `STAGED_LINE_BYTES`. The description promises that "no shape main already staged can change". But `log_block` grows towards `STAGED_RUN_BYTES` for them too, and the plan changes at every worker count. The +7–12% at 2^14–2^16 is that plan change, and it lands on the default encoder's width.

**Suggested fix:** limit run growth to `row < STAGED_LINE_BYTES` (inline at `poly.rs:356`), so rows main staged keep main's plan exactly. The width-1 wins above come from the new narrow-row gate, so they should survive. If the −19% at 2^18 / 96 threads is worth keeping, it needs a rule that doesn't also lose 10% at 2^14–2^16. Whichever you choose, a test pinning the `(local, log_block, depth)` triples for width 16 would stop this drifting silently.

binary-dft/src/poly.rs
tcoratger: You are right, and the claim is now gone. I reproduced your plan table exactly — width 16 goes to `log_block 2 / depth 6` at 2^16 and 2^18 and `log_block 1 / depth 7` at 2^14 and 2^20 — and confirmed that limiting growth to `row < STAGED_LINE_BYTES` restores main's `(7,0,8)` at every height for width >= 4.

I measured both arms from one source tree (a temporary cargo feature, so the binaries differ only in that expression), separate target dirs, `target-cpu=native`, serial pinned to `taskset -c 0-7`, parallel on 32 workers, min-of-7, two passes. Ratio is growth-everywhere over growth-limited, so below 1.00 means growth wins:

| | serial | parallel |
|---|---|---|
| width 4 | 0.71 – 0.88 | 0.79 – 0.94 |
| width 16 | 0.94 – 0.98 | 0.85 – 0.93 |
| width 64 (control) | 0.99 – 1.00 | control |

Width 4 at 128 MiB is the extreme, 138.5 -> 98.3 ms serial. Width 64 is a genuine control since its plan is `log_block = 0` in both arms; its serial ratios span 0.991–1.001 over ten heights, putting the serial noise floor under 1%. In parallel that control scatters ±4% above 4 MiB and up to ±40% below 1 MiB, which covers the two cells where growth looked slower.

So I kept growth and **removed the additive claim** from both the commit message and the PR description, replacing it with the table above. There is no shape where growth genuinely loses; where the budget refuses it, the plan is main's unchanged.

binary-dft/src/poly.rs
tcoratger: Fixed — and your four mutants are the acceptance criteria now.

The test counts passes with a test-local walk of `forward`'s loop that touches nothing on `Plan`, and separately pins `(local, log_block, depth)` for eleven production shapes from a new fixture (which also replaced the comment-only table in the schedule test, so there is one fixture rather than two).

I re-ran all four: `traversals -> above`, `log_block: 0`, `STAGED_RUN_BYTES = 64` and the twiddle perturbation now all fail. I also checked the independent count alone, with the fixture disabled, still kills a plan that ignores the traversal budget — which is the regression the budget exists to prevent.

Verified here after the fix: `traversals -> above` fails 1 test, the twiddle perturbation fails 14.

binary-dft/tests/commit.rs
tcoratger: You were right twice over — neither height stages, and `LchNtt` has no staging tile at all. The claim is gone.

What replaced it is real: the loop is now over `(num_variables, folding)` pairs, and each shape is committed through both `LchNtt` and the **default** encoder, so the file reaches `PolyBasisNtt`. Confirmed by re-running the twiddle mutant — it now fails, but only under `-C target-feature=+pclmulqdq`, because without a carryless multiply `PolyBasisNtt` delegates wholesale to the tower. Only `ci-heavy`'s `binary-large` leg sets that flag, and the comment says so.

I also moved the `folding = 0` arm to 2^5: it was costing 11.5 s of debug CI for the quadratic reference (12.83 s -> 1.45 s for the file) and the layout branch is identical at any height.

binary-dft/src/poly.rs
tcoratger: Correct, and rather than only documenting it I closed the gap.

`a_staged_cut_matches_the_reference_transform` is new: four synthetic cuts that each run a real gather and scatter, at heights the quadratic oracle can reach, at widths 1/3/16, both directions, both shifts, through `Fold::BOTH` so it is what `shifted_ntt_batch` runs. The staged path is now pinned against `NaiveAdditiveNtt` directly rather than only against the per-stage schedule and `LchNtt`.

The description is corrected: the naive sweep covers `log_n` 0…6, and `REFERENCE_LOG_N`'s doc now says why it stops there and where the staged coverage comes from instead.

binary-dft/src/poly.rs
tcoratger: Swept it, and your suspicion was right — the threshold moved from 2 to **4**.

1/2/4/8/32 workers across 4–128 MiB at width 1, staging forced on against forced off, two passes. Below 1.00 means staging wins:

| workers | 4 MiB | 8 MiB | 16 MiB | 32 MiB | 64 MiB | 128 MiB |
|---|---|---|---|---|---|---|
| 1 | 1.03 | 1.06 | 1.13 | 1.26 | 1.08 | 0.99 |
| 2 | 0.97 | 1.02 | 1.07 | **1.16** | 0.76 | 0.90 |
| 4 | 0.99 | 0.90 | 0.89 | 0.96 | 0.81 | 0.65 |
| 8 | 0.88 | 0.87 | 0.79 | 0.79 | 0.78 | 0.49 |
| 32 | 0.49 | 0.53 | 0.58 | 0.65 | 0.59 | 0.37 |

The 1-worker row reproduces the doc table to within 1%, so the two sweeps agree. At 2 workers the loss band holds and reaches +17% at 32 MiB, stable across both passes. It does not hold at 4 (worst cell 0.989) or 8. So `STAGED_WORKERS = 4` removes the whole added cost and keeps every win from 4 workers up, including all the 32-worker ones. The doc table is now the measured one.

What that costs, and the doc says so: at exactly two workers we forgo 24% at 64 MiB and 10% at 128 MiB — a win forgone rather than a cost added, which is the direction the constant already calls safe.

One correction to my own description: I could not reproduce "8 workers still lose 4–14%". In this sweep 8 workers win 12–51% everywhere. That sentence is removed.

binary-dft/src/poly.rs
tcoratger: Fixed — any width whose row already covers `STAGED_RUN_BYTES` is capped at 2^10, which today means width 64. The heavy sweep goes 20.6 s -> 8.7 s in debug here, so the cap removes 58% of it, matching your "over half the cost".

binary-dft/src/poly.rs
tcoratger: Fixed. The doc now states that a run cannot be shorter than one row, that every cell left of a width's own row length repeats that width's shortest plan, and that width 16's leading 4.73–4.96 puts the noise floor near 5% — so the other cells are read against that.

binary-dft/src/poly.rs
tcoratger: Fixed — "fuses 6 stages" is gone. The doc now says the depth moves with the height: six at 2^16, seven at 2^20, eight at 2^22.

binary-dft/src/poly.rs
tcoratger: Fixed — one helper, three readers. `Plan::leftover_stages` is now the only statement of the leftover rule and `Plan::groups` the only statement of the cut; `traversals` is `groups().count() + leftover_stages()`, `forward` iterates the group sizes, and `inverse` walks the same list backwards.

That did change `inverse`'s partition — its short group now lands at the bottom, matching `forward`. Behaviour-neutral, and the schedule tests confirm it.

binary-dft/src/poly.rs
```diff
@@ -828,41 +966,132 @@ mod tests {
...
+    /// ```
+    const PRODUCTION_CUTS: [(usize, usize, (usize, usize, usize)); 11] = [
+        (1, 10, (10, 6, 6)),
+        (1, 14, (11, 6, 6)),
+        (1, 20, (11, 3, 9)),
+        (2, 13, (10, 5, 6)),
+        (3, 12, (9, 5, 5)),
+        (4, 14, (9, 4, 6)),
+        (16, 10, (7, 2, 6)),
+        (16, 14, (7, 1, 7)),
```
Nashtare, line 1094: Width 16 is pinned at 2^10 and 2^14 only, but its plan also differs from main's at 2^16, 2^18 and 2^20. This mutation passes every test: `let target = if row >= STAGED_LINE_BYTES && above >= 9 { floor } else { log2_ceil_usize(STAGED_RUN_BYTES.div_ceil(row)).max(floor) };`. It puts width 16 back on main's (7, 0, 8) from 2^16 up. Could the plan test also assert `(16, 16, (7, 2, 6))`, `(16, 18, (7, 2, 6))` and `(16, 20, (7, 1, 7))`? A separate list would keep them out of `every_cut_of_the_stage_sequence_matches_the_per_stage_schedule`, which would otherwise run a per-stage transform of 2^24 elements.

Nashtare (changes requested):
Re-reviewed at 4b13e169. I read the interdiff from f6d45b8e, re-ran the mutants from my first review, and tried a rebase onto main. I found no correctness problem, and the code threads are addressed. I also re-measured on Graviton5. The width-16 regression is still there at 2^16, and it grows with the thread count, so I'm keeping the request for changes on that one shape. The description also still says the opposite of the commit message.

**What's fixed**
- Traversal budget: each of these now fails `the_plan_gathers_the_longest_run_its_pass_budget_allows`: `traversals` returning `above`, `log_block: 0`, `STAGED_RUN_BYTES` at 64 and at 4096, `floor..target`, and a budget compared against `usize::MAX`. `(block << s) ^ 1` fails 14 lib tests and `commit_base_matches_hand_encoding`.
- `STAGED_WORKERS = 4`: the doc table matches your sweep. The gate is its only reader, so rows of 64+ bytes get the same plan at every worker count.
- `tests/commit.rs`: the default encoder is in, and the comment is right that neither shape stages. The twiddle mutant also fails natively on Apple Silicon, because `aarch64-apple-darwin` has `aes` in its baseline. So the `linux, pclmulqdq` and `macos, parallel` CI legs run this file through `PolyBasisNtt` too, not only `binary-large`.
- `a_staged_cut_matches_the_reference_transform` reaches the staged path. The `(0, 0, 3, 6)` cut has no contiguous-tile stage and no leftover, so every butterfly goes through a gather and a scatter.
- `Plan::groups` is now the one source for the cut. The new `inverse` partition changes no output. It walks forward's `(top, take)` pairs bottom up, so it applies the same stage inverses in the same order and inverts `forward` exactly for every plan. `Rows` and `fused_stages` changed only in comments.
- Test cost, the table noise floor and the `STAGING_BYTES` depth sentence are fixed.

**Perf: width 16**
Rows main already staged still get a different plan, at every worker count. Each cell below is `(local, log_block, depth)`, then the group sizes:

| `ntt/128/hybrid` | main | this PR |
|---|---|---|
| 2^14 | (7, 0, 8): 7 | (7, 1, 7): 7 |
| 2^16 | (7, 0, 8): 8, plus 1 plain | (7, 2, 6): 6, 3 |
| 2^18 | (7, 0, 8): 8, 3 | (7, 2, 6): 6, 5 |
| 2^20 | (7, 0, 8): 8, 5 | (7, 1, 7): 7, 6 |

The traversal counts are equal, but the plans aren't equivalent on Graviton5. I measured three builds on a `c9g.24xlarge` (Neoverse V3, `-C target-cpu=native`): base `463b44c3`, this head, and this head with growth limited to `row < STAGED_LINE_BYTES` (which restores main's plan for width 16 and up). Serial runs were pinned to one core and parallel runs to cores `0..T`. Each build ran two alternating rounds, and each number is the criterion median averaged over both, all within ±2.5%. Ratios are against main; below 1.00 is faster.

| | serial | 32 threads | 64 threads |
|---|---:|---:|---:|
| `ntt/128/hybrid/14` | 1.000 | 0.960 | 0.927 |
| `ntt/128/hybrid/16` | **1.031** | **1.044** | **1.074** |
| `ntt/128/hybrid/18` | 0.981 | 0.955 | 0.865 |
| `ntt/128/hybrid/20` | 0.995 | 1.003 | 0.990 |
| `poly/forward/h16/w16` | **1.032** | **1.061** | **1.072** |
| `poly/forward/h16/w4` | 0.992 | 0.924 | 0.821 |
| `poly/forward/h16/w1` | 1.001 | 0.266 | 0.176 |

With growth limited, width 16 at 2^16 is 1.002 / 1.000 / 0.949. So the loss at 2^16 comes from the plan change and not from the rest of the PR. Growth is a real win at 2^18 and for width 4, so I'm not asking to drop it. I'm asking for a rule that doesn't turn main's 2^16 plan (one 8-stage group plus one plain pass) into two staged groups (6, 3), or for numbers showing why that cell is worth it. At 2^16 the second group fuses only 3 stages. That looks like where gathering stops paying for itself, but I haven't isolated it.

**Also**
- The PR description hasn't changed since f6d45b8e. It still says:
  - the gate is "strictly additive";
  - "no shape main already staged can change";
  - 8 workers lose 4 to 14%;
  - the oracle sweep covers `log_n` 0…14;
  - the twiddle mutant fails "13 of 53".

  The commit message is right. Could the description match it?
- The fixture doesn't pin width 16 at the heights above (inline).
- The conflict with main is mechanical. All seven conflict blocks in `poly.rs` are doc comments next to this PR's constants. Taking this PR's side for each still brings in #2153's `Conversion` change, including at the new `fused_stages` call sites.

binary-dft/src/poly.rs
tcoratger: Pinned, and your mutation is what showed the fixture was the smaller half of the problem. `4f48af3` adds a separate list, kept out of the transform loop exactly as you suggest:

```rust
/// Width-16 heights whose plan the traversal count alone does not decide.
///
/// At each of these, growing the run would hold the traversal count while turning a
/// leftover plain pass into a second gathered group, so the gather count is the only
/// thing that separates the two plans:
///
/// ```text
///     2^16   one group of eight and a plain pass, where growth would give (6, 3)
///     2^18   two groups either way, so growth rebalances (8, 3) into (6, 5)
///     2^20   two groups either way, rebalanced into (7, 6)
/// ```
///
/// These sit apart from [`PRODUCTION_CUTS`] because that list is also transformed against
/// the per-stage schedule, and `2^20` rows of width 16 is `2^24` elements to move twice.
const WIDE_CUTS: [(usize, usize, (usize, usize, usize)); 3] =
    [(16, 16, (7, 0, 8)), (16, 18, (7, 2, 6)), (16, 20, (7, 1, 7))];
```

The `2^16` triple is `(7, 0, 8)` rather than the `(7, 2, 6)` you asked for, because the same commit changes the rule so that shape keeps main's plan — see the top-level reply for why. `2^18` and `2^20` are as you listed them.

Your mutation is killed: it returns width 16 to `(7, 0, 8)` from `2^16` up, which now matches at `2^16` but not at `2^18` or `2^20`. I re-ran your earlier four as well — `traversals` returning `above`, `log_block: 0`, and `STAGED_RUN_BYTES` at 64 and 4096 — and each still fails only this test.

tcoratger:
Thank you for isolating that cell — the three-build comparison is what made the cause unambiguous, and the answer turned out to be a missing term in the budget rather than a scope limit.

## The rule

Your diagnosis was right: the second group at `2^16` fuses only 3 stages, and that is where gathering stops paying. The underlying defect is that `Plan::traversals` prices a leftover stage and a staged group the same. They are not the same: a leftover sweeps the matrix contiguously, a staged group also gathers and scatters. So a longer run could hold the traversal count while spending a gather it did not previously spend.

`4f48af3` adds that as a second budget, measured against the shortest run just as the first is:

```rust
let budget = Self::traversals(above, depth_at(floor));
let gathers = Self::groups(above, depth_at(floor)).count();
let log_block = (floor..=target)
    .rev()
    .find(|&log_block| {
        let depth = depth_at(log_block);
        Self::traversals(above, depth) <= budget
            && Self::groups(above, depth).count() <= gathers
    })
    .unwrap_or(floor);
```

I preferred this to limiting growth to `row < STAGED_LINE_BYTES` because the row width is not what distinguishes the cells. Growth is good at width 16 / `2^18` and bad at width 16 / `2^16`, and both are rows main staged, so a width cut has to lose one to fix the other. What separates them is whether growth rebalances existing groups or adds one.

## What it does to width 16

| `ntt/128/hybrid` | main | before | now |
|---|---|---|---|
| 2^14 | (7, 0, 8): 7 | (7, 1, 7): 7 | (7, 1, 7): 7 |
| 2^16 | (7, 0, 8): 8, plus 1 plain | (7, 2, 6): 6, 3 | **(7, 0, 8): 8, plus 1 plain** |
| 2^18 | (7, 0, 8): 8, 3 | (7, 2, 6): 6, 5 | (7, 2, 6): 6, 5 |
| 2^20 | (7, 0, 8): 8, 5 | (7, 1, 7): 7, 6 | (7, 1, 7): 7, 6 |

`2^16` is back on main's plan exactly, so the 1.031 / 1.044 / 1.074 should go with it. The `2^18` and `2^20` plans are untouched, so the −13.5% at 64 threads stays.

I swept the rule over widths 1…64 and `log_n` 0…24 before implementing it. 149 shapes change, all of them the same case — growth that would have added a gathered pass. Nothing else moves: all eleven `PRODUCTION_CUTS` triples are unchanged, and of the shapes you benchmarked only width 16 / `2^16` moves. `poly/forward/h16/w1`, `h16/w4`, `commit/16`, `commit/18` and `commit/20` all keep the plan they had.

I have not re-measured on Graviton5, so the ratios above are a prediction from the plan identity rather than a measurement. If you re-run, `ntt/128/hybrid/16` and `poly/forward/h16/w16` are the two cells that should move.

## The rest

- **Fixture.** `WIDE_CUTS` pins the three width-16 heights in a separate list, inline thread answered. Your mutation is killed, as are the four from your first review.
- **Description.** Rewritten. It now says the gate is *not* strictly additive and that rows main staged can change plan, with the table above; the 8-worker loss band is marked machine-specific and noted as not reproducing on Graviton5; the oracle sweep is stated as `log_n` 0…6 against `NaiveAdditiveNtt` with the synthetic cuts doing the staged-path work; and the twiddle mutant number is corrected.
- **Twiddle mutant.** Your 14 + `commit_base_matches_hand_encoding` reproduces here, but only on a build whose target has carryless multiply. On a bare `x86-64` baseline it is 3, because `commit.rs` reaches a different transform. Both numbers are now in the description with that caveat, since the bare figure is what someone running `cargo test` on this machine would see.
- **Rebase.** Done, onto `ce322afe`. All seven conflicts were doc comments as you said. I took this branch's side for each, then restored main's one-sentence-per-line reflow on the six paragraphs that had it, so the rebase does not undo that pass. #2153's `Conversion` is preserved, including at the new `fused_stages` call sites.

Verified: `p3-binary-dft` 93 + 2 and `p3-binary-pcs` 66 + 4 + 27 pass in both feature sets and again under `-C target-cpu=native`; clippy clean in both.

binary-dft/src/poly.rs
```diff
@@ -246,48 +310,115 @@ struct Plan {
...
+        let local = log2_floor_usize((TILE_BYTES / row).max(1)).min(log_n);
+        let above = log_n - local;
+        // Stages one tile fuses, with a run of `2^log_block` adjacent rows as its row.
+        let depth_at =
+            |log_block: usize| log2_floor_usize((STAGING_BYTES / (row << log_block)).max(1));
+        let floor = log2_ceil_usize(STAGED_LINE_BYTES.div_ceil(row));
+        let target = log2_ceil_usize(STAGED_RUN_BYTES.div_ceil(row)).max(floor);
+        // The shortest admissible run sets both counts no longer run may exceed.
+        let budget = Self::traversals(above, depth_at(floor));
+        let gathers = Self::groups(above, depth_at(floor)).count();
```
Nashtare, line 417: The gather budget only binds when the shortest-run plan leaves a stage over, and then it blocks growth entirely: `above % depth == 1` means any shallower depth needs another group. In effect a run grows only when the shortest run has no plain pass. That returns 146 shapes at widths 4..63 to main's plan, not only 2^16. Two are shapes this PR measured. For width 4 at 2^20, the `STAGED_RUN_BYTES` table puts the refused plan at 4.40 ms and the one picked now at 5.74 ms, and the commit message has 64 MiB w4 at 59.10 -> 44.86 ms serial. Width 1 at 2^22 is 5.21 vs 6.32 ms. Is giving those up intended? If so, the table and commit message need updating. If not, the extra group at 2^16 fuses 3 stages and these fuse 5, so a rule keyed on the shallowest group would keep them.

binary-dft/src/poly.rs
```diff
@@ -819,41 +986,152 @@ mod tests {
...
+        (16, 14, (7, 1, 7)),
+        (64, 10, (5, 0, 6)),
+        (512, 8, (2, 0, 3)),
+        (1024, 7, (1, 0, 2)),
+    ];
+
+    /// Width-16 heights whose plan the traversal count alone does not decide.
+    ///
+    /// At each of these, growing the run would hold the traversal count while turning a
+    /// leftover plain pass into a second gathered group, so the gather count is the only
```
Nashtare: This only holds at 2^16. At 2^18 and 2^20 the one-row plan is `[8, 3]` and `[8, 5]`, with no plain pass, so the gather count never binds there. The traversal count alone picks `(7, 2, 6)` and `(7, 1, 7)`. Those two rows pin the growth that survives the rule, which is what kills a width cut, so could the doc say that instead? The PR description has the same claim ("the three width-16 heights the traversal count alone does not decide").

Nashtare (approved):
Re-reviewed at 68a965ba. I read the new commit and the rebase, and I re-derived the plans at main, 4b13e169 and this head from a transcription of `Plan::for_workers`. I didn't re-benchmark.

**2^16 is fixed.** Width 16 at 2^16 is back on main's (7, 0, 8): one 8-stage group plus a plain pass. With `log_block = 0`, `fused_stages` addresses memory exactly as main does, and `forward`/`inverse` walk the same `(top, take)` pairs. My growth-limited build at 4b13e169 ran this plan through the same code at 1.002 / 1.000 / 0.949, so I expect the regression to be gone. That rests on the plans being identical; I haven't measured this head.

**Also checked**
- 149 shapes change, all of them the leftover-to-group case. `PRODUCTION_CUTS`, and every Criterion shape except w16/2^16, keep their plans.
- `WIDE_CUTS` kills my mutant, and it also catches a revert of the gather budget.
- The rebase is clean: the only code change it brings is #2153's, including the `fused_stages` call sites.
- The description no longer says "strictly additive", "0…14" or "13 of 53".

**Non-blocking**
- The budget also refuses growth at width 4 / 2^20 and width 1 / 2^22, where your own table and commit message measure the refused plan 16–24% faster (inline).
- "Where it costs something" still has the 8-worker Ryzen loss you retracted, and the doc table shows 8 workers winning.
- The serial 2^16 row, and the commit message's 2^16 and 64 MiB w1/w4 rows, measure plans this head no longer picks.

binary-dft/src/poly.rs
tcoratger: You are right on every point, and no, giving those up was not intended. `a636322` changes the rule.

I reproduced your enumeration exactly: **146 shapes at widths 4..63** over `log_n` 0..24, and the mechanism is the one you describe. When the shortest run leaves a stage over, `above % depth == 1` makes `(above-1)/depth` exact, so every shallower depth needs strictly more groups and the gather count refuses every candidate. Width 1 at `2^22` is refused the same way.

The two measured cells make the disagreement concrete — the code was refusing the plan the `STAGED_RUN_BYTES` table itself picks:

| | refused | picked |
|---|---|---|
| width 4 @ 2^20 | 1024 B, 4.40 ms | 64 B, 5.74 ms |
| width 1 @ 2^22 | 1024 B, 5.21 ms | 64 B, 6.32 ms |

### What actually pays for the gather

I tried your suggestion first and it does not hold up. Keyed on the shallowest group, width 17 at `2^21` and width 16 at `2^24` produce a depth-5 group exactly like width 4 at `2^20`, but they only lengthen the run **4x** and they measure **1.008** and **0.979** — nothing. Width 4 at `2^20` lengthens it **16x** and measures **0.780**. The discriminator is the run length, not the group depth.

So I swept by run-growth factor. Forced plans read from one binary (an atomic the plan consults, so the two arms are the same code), arms alternated, min of 7, and a third arm re-running the first as a null control on every shape:

| growth | 32 workers | one pinned worker |
|---|---|---|
| 2x | 1.02 1.07 1.12 1.13 1.02 | 1.03 1.07 1.05 1.04 |
| 4x | 0.98 1.00 1.00 1.01 1.01 | 1.04 0.96 1.00 1.08 |
| 8x | 0.93 0.98 | 0.90 0.91 0.90 0.87 0.94 0.96 0.96 0.95 |
| 16x | 0.78 0.80 0.80 0.85 0.78 0.80 0.80 | |

Controls: 1.031 at 32 workers, 1.000–1.013 pinned. Below 1.00 the longer run wins.

That is monotone and the knee is between 2 and 3 doublings. So the rule is now `STAGED_GATHER_DOUBLINGS = 3`: an added gather is admitted only once the run has grown eightfold. The traversal budget still caps the total, and it already allows at most one added group, so a run that has paid buys exactly one gather.

The rule changes **15 shapes** over widths 1..1024 and `log_n` 0..24, and **all 15 are in the table above** — the seven at 16x are widths 1..7, the eight at 8x are width 8 at `2^18` and widths 9..15 at `2^16`. Width 16 at `2^16` is offered two doublings, so it keeps main's plan and @Nashtare's Graviton result stands.

Cross-check against the `STAGED_RUN_BYTES` table: it gives 4.40/5.74 = 0.766 and 5.21/6.32 = 0.824 for width 4 at `2^20` and width 1 at `2^22`; this sweep gives 0.780 and 0.838 for the same two shapes.

### Pinned

`PLAN_ONLY_CUTS` now brackets the constant instead of listing width-16 heights: `(48, 18)` at one doubling refused, `(16, 16)` at two refused, `(8, 18)` at three admitted, `(4, 20)` at four admitted, `(1, 22)` at four from a run floor above one row.

Setting the constant to 0, 1, 2, 4, 5 or effectively infinity each fails that one test, as does dropping the exemption clause. I also re-ran your earlier mutants on the new tree: `traversals` returning `above`, `log_block: 0`, `STAGED_RUN_BYTES` at 64 and 4096, and your `target` mutation each fail only this test; the twiddle perturbation fails 14.

binary-dft/src/poly.rs
tcoratger: Correct, and I checked both heights before rewriting it.

At `2^18`, `above = 11` and the one-row depth is 8, so `11 % 8 == 3`, the leftover is zero and the plan is `[8, 3]`. At `2^20`, `above = 13` gives `[8, 5]`. In both the gather count is 2 either way, so it never binds, and the traversal count alone picks `(7, 2, 6)` and `(7, 1, 7)` — at `2^20` by refusing `log_block 2`, whose `[6, 6]` plus a leftover is three traversals against a budget of two. Only `2^16` has a leftover to trade, so only there does the gather rule decide anything.

The list is now `PLAN_ONLY_CUTS`, named for the one thing all its rows share — they are pinned as a triple because transforming them costs too much — and its doc says what decides each row rather than making one claim for all of them:

```text
    width 48 @ 2^18   one doubling on offer, the least a growing run can have
    width 16 @ 2^16   two doublings, so the added gather stays unpaid
    width  8 @ 2^18   three doublings, which is exactly what a gather costs
    width  4 @ 2^20   four doublings, the widest row a run still spans several of
    width  1 @ 2^22   four doublings again, from a run floor above one row
    width 16 @ 2^18   no leftover pass, so growth rebalances (8, 3) into (6, 5)
    width 16 @ 2^20   no leftover pass either, rebalanced into (7, 6)
```

followed by the sentence you asked for: the last two offer no leftover to trade, so the traversal count alone picks their run.

The four new rows are there because the top-level thread changed the rule — they bracket the new constant from both sides. The PR description is corrected too.

PR #2156, opened 2026-09-16
perf(binary-dft): give the generic additive transform a tiled schedule

tcoratger:
The generic Lin–Chung–Han transform ran one full pass over the matrix per stage. Two levels of row blocking replace that.

## The problem

`LchNtt` runs `l` stages, and stage `j` pairs rows `2^j` apart — so every stage was a full pass. At production heights the cost is the traffic of those passes, not the arithmetic in them: **20 passes over 256 MiB at 2^20**.

Its `BinaryField128` sibling `PolyBasisNtt` already has a blocked plan. `LchNtt`, which serves `BinaryField32`, `BinaryField64` and `Ghash128`, had none, and it shows as a scaling cliff — `128/ghash` took **15.7x** for a 4x size step at 2^18 -> 2^20.

## The idea

Two levels of blocking, each chosen so the row set is *closed* under a run of adjacent stages.

**Contiguous tile.** The bottom `t` stages pair rows less than `2^t` apart, so a block of `2^t` consecutive rows is closed under all of them. The whole radix-2 network runs on that block in one visit: one read, one write, `t` stages.

**Staging tile.** Above the tile, a group of `depth` stages is closed under the strided row set `R(k) = S * (blk * 2^depth + k) + r` with `S = 2^(top + 1 - depth)` and `r < S`, so those rows are gathered into a per-worker buffer, the group runs there, and they are scattered back. One read and one write per *group* instead of per stage.

**Row runs.** When a row is shorter than a cache line the gather would fetch 64 bytes to use 4, so a staged row becomes a run of adjacent matrix rows.

One routine drives both levels: a staged tile of `2^depth` rows with index `blk` sees exactly the block indices `(blk << s) + (k >> (depth - s))`, which is what a contiguous tile of the same depth sees.

## Staging is gated on the worker count

This is what changed most in review.

**The crossover follows the last-level cache, not the shape.** Pinned to this machine's 96 MiB die the tiled schedule turns at ~128 MiB; pinned to its 32 MiB die the *same binary* turns at ~32 MiB. A byte threshold would be a guess about a number a `no_std` crate cannot read.

Splitting the schedule in two showed the cost is entirely the **staging**, never the contiguous tile — tile-only against plain is 1.02–2.38x and never below 0.93x at any worker count. Staged against tile-only, over every GHASH shape at 1 MiB and up:

```text
    workers        1      2      4      8     16     32
    mean        0.82   0.90   1.00   1.16   2.00   2.47
    worst       0.61   0.60   0.62   0.62   1.22   1.71
```

So `STAGED_WORKERS = 16` gates **staging only**. Below it the schedule keeps the contiguous tile and runs plain passes above it.

A matrix far enough past the last-level cache is staged at a profit from two workers up, and below this count that is given up — a win forgone rather than a cost added, which is the safe direction for a signal this coarse.

## Benchmark

Ryzen 9 9950X3D, `RUSTFLAGS="-C target-cpu=native"`, width 16, matched builds, runs serialized against other work, serial runs pinned.

`ntt/128/ghash`, `--features parallel`, 32 workers:

| | before | after | speedup |
|---|---:|---:|---:|
| 2^14 | 810.0 µs | 230.7 µs | **3.51x** |
| 2^16 | 2.633 ms | 657.7 µs | **4.00x** |
| 2^18 | 11.42 ms | 4.235 ms | **2.70x** |
| 2^20 | 183.7 ms | 32.20 ms | **5.71x** |

The scaling cliff is gone: 2^18 -> 2^20 is now 7.6x for a 4x size step, against 15.7x before.

One worker, against the plain schedule: **1.11x / 1.04x / 1.10x / 1.50x** at 2^14…2^20.

GHASH is the basis that moves because its multiply is fast enough for the transform to be bound by memory rather than arithmetic. `ntt/32`, `ntt/64` and `ntt/128/tower` gain 1.01x to 1.25x in parallel, up to 1.42x at 2^20 on the tower basis.

**A correction to an earlier revision of this description.** It reported one-worker figures of 0.95x / 0.87x / 1.88x / 2.29x. Those were measured without pinning: the 1.88x at 2^18 was the 32 MiB die, and the same shape on the 96 MiB die is 0.91x. The numbers above are pinned, and the worker gate is what removes the sub-1.0 cases.

## Budgets

Both constants are the floor of a sweep, taken as a geometric mean over the whole sweep rather than a single height.

- **Contiguous tile, 128 KiB.** 32/64/128/256 KiB give 1.047/1.016/1.000/0.987 on one worker and 1.071/1.017/1.000/1.032 at 32 — both columns flatten at 128 KiB.
- **Staging tile, 64 KiB.** 32/64/128/256 KiB and 1 MiB give 1.076/1.000/1.013/0.975/0.999 at 16 workers and 1.037/1.000/0.971/0.963/0.964 at 32. There is no fall-off at 1 MiB, so an earlier rationale here was wrong; 64 KiB is simply the smallest budget at a flat floor.

A tile is also a task, so the contiguous tile stops growing once there is one per worker.

## Correctness

The transform is a commitment primitive, so the schedule is checked three independent ways.

- **Against an independent walk**, element for element: widths 1…8 plus 16 and 64, heights 0…10, both shifts, both directions, at 32, 64 and 128 bits and in the GHASH basis. The oracle recomputes every block twiddle from its own index.
- **Against `NaiveAdditiveNtt`**, both directions, all widths.
- **Against a model of the row map.** All 715 group shapes for `log_n <= 10` are walked through `StagedRows::row`/`block` — the same map `fused_group` calls — checking it is a bijection onto the matrix rows, so two live tasks can never touch the same row, and that the sub-stage block index matches at every level.

The worker clamp is pinned by a test over 8 worker counts x heights 0…24 x 5 widths, and the boundary sweeps run each width against explicit 1- and 32-worker schedules so the staged path is covered whatever the test machine reports.

**What the default sweeps do and do not cover:** at production budgets they leave the plain branch only for width 64 and for 16-byte width 16 at 2^10; widths 1–8 stay inside one tile budget, and no staged group executes there. The staged path is covered by the boundary and production tests with explicit worker counts instead.

The raw row pointer used by the staging tasks asserts its last row is in bounds once per task, and the per-worker tile is capacity-only rather than zero-filled.

## Scope

`LchNtt` is **not on this repository's hot path today** — `AdditiveRsEncoder` defaults to `PolyBasisNtt`, and `p3-binary-pcs` takes that default. This is the public generic transform for the 32- and 64-bit levels and for GHASH.

This file now duplicates the staging machinery in `binary-dft/src/poly.rs`. Sharing it behind one kernel is a sensible follow-up once both land; the plans should stay separate, since they optimize different objectives.

## Verification

- `cargo test -p p3-binary-dft` — 68 + 2 passing, serial and `--features parallel`, plus the `#[ignore]`d heavy sweep in both.
- `cargo clippy -p p3-binary-dft --all-targets` clean with and without `parallel`; nightly `cargo fmt` clean.


binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+///
+/// That is what a contiguous tile of `2^depth` rows carrying tile index `blk` would see.
+/// So the staged rows can be run as one.
+///
+/// The pairs `(blk, r)` partition the matrix, and each covers `2^depth` rows.
+/// So the matrix is read once and written once per group, rather than once per stage.
+///
+/// # Panics
+/// Panics if a task's row walk reaches past the end of `values`.
+fn fused_group<F: TowerLevel, const INVERSE: bool>(
```
Nashtare, line 300: This is `fused_stages` + `Rows::{check, gather, scatter}` from `poly.rs` again: the same `R(k)` derivation, the same capacity-only per-worker tile, the same hard assert and message. The policies around it differ, though. Here the tile count is clamped to workers, splits are bounded with `with_min_len`, and there's no `use_parallel` fallback. Rows shorter than a line are staged in 64-byte runs at any worker count, where #2155 gives `PolyBasisNtt` 64–1024-byte runs from 2 workers up. Could the plan, the row map and gather/scatter be shared behind a kernel closure, so there's one `unsafe` path and one bijection proof? Or can the two policies at least be aligned?

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+///      32 KiB              130.6 ms               4.32 ms
+///      64 KiB              121.9 ms               4.29 ms
+///     128 KiB              115.5 ms               4.22 ms
+///     256 KiB              116.0 ms               3.57 ms
+/// ```
+///
+/// 128 KiB is where the one-worker curve flattens.
+/// It is an eighth of the private second-level cache measured here.
+///
+/// It is half of the smallest such cache in circulation.
```
Nashtare: This doesn't hold as written: Cortex-A55's private L2 is configurable at 64, 128 or 256 KB. The 32-worker column above also has 256 KiB 15% faster than 128 KiB (3.57 vs 4.22 ms), and `PolyBasisNtt` uses 32 KiB for the same contiguous tile. What actually fixes 128 KiB, and should the two transforms share the constant?

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+/// ```
+///
+/// From 32 KiB to 256 KiB the curve is flat to within a few percent.
+/// It falls off only at a whole second-level cache per core.
+///
+/// There the tiles of the workers sharing that cache stop fitting beside each other.
+///
+/// 64 KiB is the smallest budget at that floor.
+/// So it is the one that keeps a tile per core on a machine with less cache than this.
+const STAGING_TILE_BYTES: usize = 64 * 1024;
```
Nashtare, line 65: `poly.rs` justifies the same 64 KiB as "exactly the per-core private cache size of the machine the sweep ran on", while this comment puts the fall-off at a whole 1 MiB L2 per core. Can the two rationales be reconciled?

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+    // Blocking it would only add bookkeeping to stages already free of memory traffic.
+    if core::mem::size_of_val(mat.values.as_slice()) <= DEEP_TILE_BYTES {
+        for step in 0..log_n {
+            let j = if INVERSE { step } else { log_n - 1 - step };
+            stage_pass::<F, INVERSE>(&mut mat.values, width, j, &twiddles);
+        }
+        return mat;
+    }
+
+    let schedule = Schedule::new::<F>(width, log_n);
```
Nashtare, line 544: Without `parallel`, `current_num_threads()` is a const 1, so every serial build takes this path above 128 KiB, including the one-worker GHASH shapes the description says lose (0.95x/0.87x). #2155 gates its new behaviour on a worker count (`STAGED_WORKERS`), so serial builds keep the old schedule there. The description argues against a worker-blind threshold; was a worker-aware one considered, e.g. plain passes when `current_num_threads() == 1` below some size?

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+    log_n: usize,
+    twiddles: &Twiddles<F>,
+    schedule: &Schedule,
+) {
+    // One contiguous tile is one task, and a tile of `2^t` rows leaves `2^(log_n - t)` of them.
+    //
+    // So the budget only buys depth once there is a tile for every worker to take.
+    let tile = schedule
+        .log_tile_rows
+        .min(log_n.saturating_sub(log2_ceil_usize(current_num_threads())));
```
Nashtare: No test pins this clamp: with `.min(log_n)` in its place, the default suite passes serially, at 14 threads and at 4. The description credits it with the difference between 0.53x and a win at 2^14. Could the plan take `workers` as a parameter (like #2155's `Plan::for_workers`), so a unit test can assert `2^(log_n − tile) ≥ workers`?

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+                    "staged row walk leaves the matrix"
+                );
+
+                tile.clear();
+                for k in 0..rows {
+                    // SAFETY: a row is `stride · (block · rows + k) + run · 2^log_slab`.
+                    //
+                    // The run then spans `0 .. 2^log_slab` from there.
+                    //
+                    // That is a bijection onto the matrix rows as `(block, k, run)` range.
```
Nashtare: At runtime only `debug_assert`s check this bijection, and the 1020-shape model from the description isn't in the PR. Could it land as a test, ideally over a small `fn staged_row(task, k)` that this loop also calls, so the model checks the real mapping?

binary-dft/src/lch.rs
```diff
@@ -188,11 +637,59 @@ mod tests {
...
+    /// So this range reaches the blocked schedule, staging tile included, from the wide end.
+    ///
+    /// The narrow end needs taller matrices.
+    ///
+    /// The boundary sweep builds those at one element per row, and the heavy sweep covers them in bulk.
+    const LOG_HEIGHTS: core::ops::RangeInclusive<usize> = 0..=10;
+
+    /// The heights the heavy sweep covers on top of [`LOG_HEIGHTS`].
+    ///
+    /// Every shape of 8 bytes or wider has crossed its tile boundary by `2^14`.
```
Nashtare: At `2^14` an 8-byte row is exactly 128 KiB, so `transform` takes the `<= DEEP_TILE_BYTES` plain branch and the shape never crosses its boundary. With counters on `run()`, `BinaryField64` width 1 and `BinaryField32` widths 1 and 2 never reach it for heights 0..=14. `BinaryField32` width 2 is also missing from `BOUNDARY_WIDTHS`, so no test runs it on its production plan above the boundary. Adding 2 to `BOUNDARY_WIDTHS` and correcting this comment would close that.

binary-dft/src/lch.rs
```diff
@@ -188,11 +637,59 @@ mod tests {
...
+
+    /// The heights every sweep covers.
+    ///
+    /// A width-64 row of any level overruns both byte budgets well before `2^10` rows.
+    /// So this range reaches the blocked schedule, staging tile included, from the wide end.
+    ///
+    /// The narrow end needs taller matrices.
+    ///
+    /// The boundary sweep builds those at one element per row, and the heavy sweep covers them in bulk.
+    const LOG_HEIGHTS: core::ops::RangeInclusive<usize> = 0..=10;
```
Nashtare, line 634: At production budgets, the only shapes these default walk sweeps send through `run()` are width 64 and 128-bit/GHASH width 16 at `2^10`. Widths 1–8 stay on the plain branch, and no row-run group executes, at 1 or 14 threads. The blocked path for narrow rows is pinned by the boundary sweeps, `production_schedule_*` and the custom shapes instead. So "widths 1…8, heights 0…10" in the description claims more than these sweeps cover.

scripts/check.py
```diff
@@ -343,6 +343,16 @@ def commands_for(args: argparse.Namespace) -> list[list[str]]:
                 "--ignored",
                 "--exact",
             ],
+            [
+                "cargo",
+                "test",
+                "-p",
+                "p3-binary-dft",
+                "lch::tests::the_schedule_matches_an_independent_walk_at_every_blocked_shape",
```
Nashtare, line 351: `slow-regressions` runs without `--features parallel`, so this sweep never exercises the worker clamp in `run` or concurrent staging tasks. Is a second registration with `--features parallel` worth adding?

Nashtare (commented):
Reviewed at 3c47179e; cf1f1958 only deletes blank `///` lines. I found no correctness or soundness problem. What's left is design, and a few claims.

**What holds**
- **The math.** The twiddle identity and row-set closure hold for both tile levels, including row runs. Every row of a run sits at an offset below `S`, so it has the same block at every stage of its group, and `slab ≤ tile ≤ low` keeps a run inside the stride.
- **Memory safety.** `(block, run, k, u)` is a mixed-radix decomposition of `0..2^log_n`, the walk ascends, and the last-row bound is a hard `assert!`. The capacity-only tile has no `set_len` and no uninitialized read.
- **Differential probe.** I made the byte budgets settable at runtime in a copy and compared the output against `NaiveAdditiveNtt`:
  - the naive inverse at every height;
  - the naive forward up to 2^8, and above that a naive-inverse round trip plus bit-equality across runs;
  - 32/64/128/GHASH, widths 1..=17, heights to 2^18 at width 1 and 2^14 otherwise;
  - the production budgets plus 7 shrunk budget triples;
  - a serial build, and `parallel` with pools of 1/2/3/8/14 plus a nested install.

  All of it passes, and the output is bit-identical across thread counts.
- **Miri** is clean on small tiled shapes, both serial and with 3 rayon threads under Tree Borrows.
- **Mutations.** The default suite kills every functional mutant I tried: the twiddle identity in the shared, staged-only and deep-only paths, the stride, the row-run offset, the stage base, the sub-layer order, the group order, and the `W_j` chain. The survivors are perf-only or unreachable: the worker clamp, the run-length rounding, and the slab clamp, which binds only above 2048 threads.
- Clippy `-D warnings` (both feature sets), nightly fmt and the ignored sweep all pass. `check.py --dry-run slow-regressions` lists the sweep.

**Main point: a second copy of the planner.** `tile_stages`/`deep_tiles`/`fused_group`/`Schedule`/`run` reimplement `poly.rs`'s `tile_stages`/`local_stages`/`fused_stages` + `Rows`/`Plan`/`forward`/`inverse`: the same derivation, the same per-worker tile, the same hard assert. The policies around them differ:

| | `LchNtt` (this PR) | `PolyBasisNtt` |
|---|---|---|
| contiguous tile | 128 KiB | 32 KiB |
| tile count | clamped to workers | not clamped |
| task splits | `with_min_len` | unbounded |
| serial fallback | none | `use_parallel` |
| narrow rows | 64-byte runs at any worker count | 64–1024-byte runs, from 2 workers up (#2155) |

With #2155 open against the other copy, could the two share one planner generic over the butterfly kernel, with entry/exit hooks for the basis conversion? Or at least one policy?

**Smaller points, inline:**
- a serial build always tiles;
- the budget rationale overstates portability;
- the heavy sweep runs only serially;
- several test comments claim coverage the sweeps don't have.

binary-dft/src/lch.rs
```diff
@@ -380,11 +1056,243 @@ mod tests {
...
+            log_staged_rows: staged,
+            log_slab_rows: slab,
+        };
+
+        for width in WIDTHS {
+            // A three-stage tile with two-stage groups above it.
+            //
+            // Heights 2 and 3 stay inside a single tile, and 4 adds one group.
+            //
+            // 5 adds a group and a one-stage remainder, and 8 two whole groups and a remainder.
```
Nashtare: Nit: with counters on `run`, serially, 4 adds a one-stage plain pass, 5 exactly one fused group with no remainder, and 6 a group plus a remainder. `shape(2, 4, 0)` at 7 clips its top group to one stage, which runs as a plain pass rather than a clipped fused group. Under `parallel` the clamp reshapes these too: with 14 workers, heights 2–4 have no tile.

binary-dft/src/lch.rs
```diff
@@ -380,11 +1056,243 @@ mod tests {
...
+        sweep_against_the_walk::<Ghash128>(HEAVY_LOG_HEIGHTS);
+    }
+
+    /// Every width, at the three heights around its contiguous tile boundary.
+    /// That boundary is where the matrix stops fitting one tile.
+    /// Above it the schedule cuts stages off the top, so it is where the blocking starts:
+    ///
+    /// ```text
+    ///     2^t rows        the tile is the whole transform
+    ///     2^(t+1) rows    one stage above the tile, run as a plain pass
```
Nashtare: Nit: at `2^t` the matrix is exactly one budget, so `transform` takes the plain-pass branch rather than a tile. At `2^(t+1)` with `T` workers the clamp leaves a tile of `t+1−ceil(log2 T)`, so "one stage above the tile, run as a plain pass" holds only for one worker.

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+    /// Base-two log of the staged rows one group of stages covers.
+    log_staged_rows: usize,
+    /// Base-two log of the matrix rows one staged row holds.
+    log_slab_rows: usize,
+}
+
+impl Schedule {
+    /// Divide each byte budget by the bytes one row occupies.
+    ///
+    /// An element is 4 to 16 bytes wide, and a caller's width is its own business.
```
Nashtare: Nit: `LchNtt<F: TowerLevel>` also runs on `BinaryField8` and `BinaryField16` (this file's proptests do), so an element is 1 to 16 bytes.

binary-dft/src/lch.rs
```diff
@@ -44,97 +53,537 @@ pub struct LchNtt<F> {
...
+/// So it is the one that keeps a tile per core on a machine with less cache than this.
+const STAGING_TILE_BYTES: usize = 64 * 1024;
+
+/// The smallest number of stages a staging tile has to cover to be worth gathering for.
+///
+/// One stage gathered and scattered moves the memory traffic one plain pass moves.
+/// It then adds a copy in each direction, so grouping pays only from two stages up.
+const MIN_FUSED_STAGES: usize = 2;
+
+/// The transfer granularity of the memory system, in bytes.
```
Nashtare: Nit: Apple Silicon reports `hw.cachelinesize: 128`, so 64 is the smallest line among targets rather than the transfer granularity; `poly.rs` words the same constant that way.

Nashtare (approved):
Approving. I benchmarked cf1f1958 against main 463b44c3 on Graviton5 (`c9g.24xlarge`, 96 Neoverse-V3 cores, one NUMA node, `-C target-cpu=native`).
- Width 16; each comparison ran twice, alternating base and head.
- Serial runs were pinned to one core, and parallel runs to cores `0..T`.
- Cells show time relative to main (below 1.00 is faster), averaged over both rounds, which agree within ±1% on almost every id and within ±3% at worst.
- `ntt/128/hybrid` is untouched by this PR and served as the control.

| | serial | 8 threads | 32 threads | 96 threads |
|---|---|---|---|---|
| `ntt/128/ghash`, 2^14 / 16 / 18 / 20 | 1.00 / 1.03 / 1.02 / 1.00 | 0.59 / 0.73 / 0.98 / 0.87 | 0.37 / 0.40 / 0.58 / 0.47 | **0.16 / 0.24 / 0.29 / 0.33** |
| `ntt/64` | 0.98 / 0.93 / 0.91 / **0.80** | 0.86 / 0.93 / 0.90 / 0.85 | 0.58 / 0.77 / 0.88 / 0.89 | 0.25 / 0.44 / 0.58 / 0.84 |
| `ntt/32` | 1.01 / 1.00 / 1.00 / 1.00 | 0.93 / 0.97 / 0.99 / 1.00 | 0.64 / 0.82 / 0.93 / 0.97 | 0.30 / 0.49 / 0.80 / 0.89 |
| `ntt/128/tower` | 1.01 / 1.01 / 1.01 / 1.02 | 0.93 / 0.96 / **1.06 / 1.04** | 0.74 / 0.87 / 1.00 / 1.01 | 0.40 / 0.66 / 0.82 / 0.95 |
| `ntt/128/hybrid` (control), 2^14 / 20 | 1.00 / 1.00 | 1.00 / 1.00 | 1.04 / 1.00 | 0.92 / 1.00 |

In absolute terms, GHASH at 2^20 on 96 threads goes from 13.98 ms to 4.58 ms, and 64-bit serial at 2^20 from 1.198 s to 961 ms. The scaling cliff is gone here too. On main, going from 32 to 96 threads makes GHASH slower at 2^14 (0.53 → 1.10 ms) and at 2^16 (1.47 → 2.27 ms); with this PR both still improve (0.19 → 0.18 ms, 0.59 → 0.53 ms).

**What costs something, all small:**
- serial GHASH at 2^16 / 2^18: +2.9% / +1.8%, smaller than the losses the description reports on Zen 5;
- serial tower: +1–2%;
- tower at 8 threads, 2^18 / 2^20: +6.0% / +3.7%.

None of these is on `binary-pcs`'s path, and each is outweighed by the wins in the same rows.

Earlier points still open, as follow-ups rather than blockers:
- a planner and unsafe gather/scatter shared with `PolyBasisNtt`, or at least one policy (#2155 is changing the other copy's policy right now);
- a test pinning the worker clamp;
- the `parallel` registration for the heavy sweep;
- the test-comment coverage claims.

binary-dft/src/lch.rs
tcoratger: Fixed, and your instinct was right — but the *size* form of the threshold does not work, so it became a worker gate.

The one-worker crossover follows the **last-level cache, not the shape**. Pinned to cores 0–7 (96 MiB L3) the tiled schedule turns at ~128 MiB; pinned to cores 8–15 (32 MiB L3) the same binary turns at ~32 MiB. Any byte constant would be a guess about a number `no_std` cannot read.

Splitting the schedule showed the cost is entirely the **staging**, not the contiguous tile: tile-only against plain is 1.02–2.38x and never below 0.93x at any worker count. Staged against tile-only, over every GHASH shape at 1 MiB and up:

```
    workers        1      2      4      8     16     32
    mean        0.82   0.90   1.00   1.16   2.00   2.47
    worst       0.61   0.60   0.62   0.62   1.22   1.71
```

So `STAGED_WORKERS = 16` now gates **staging only**; below it the schedule keeps the tile and runs plain passes above it. One worker goes from the 0.95x/0.87x losses to **1.11x / 1.04x / 1.10x / 1.50x** at 2^14…2^20, and 32 workers is unchanged.

That same measurement corrects my own description: the 1.88x it reports at 2^18 on one worker was the 32 MiB die. Pinned to the 96 MiB die that shape is **0.91x**. I am updating the description accordingly.

I deliberately did not take #2155's `STAGED_WORKERS = 2` — measured here, staging still loses at 2, 4 and 8.

binary-dft/src/lch.rs
tcoratger: Fixed — the plan now takes the worker count (`Schedule::for_workers`, as in #2155).

`the_contiguous_tile_leaves_one_for_every_worker` asserts both that `2^(log_n - tile) >= workers` and that the tile is no shallower than the budget allows, over 8 worker counts x heights 0..=24 x 5 widths. It fails immediately if the clamp is replaced by `.min(log_n)`, which is the control you ran.

binary-dft/src/lch.rs
tcoratger: Landed as a test, and wired to the production map as you asked.

`a_staging_group_stages_every_row_exactly_once` walks all 715 group shapes for `log_n <= 10` through `StagedRows::row`/`block` — which `fused_group` itself now calls — checking that it is a bijection onto the matrix rows and that `first >> (top - s + 1) == (block << s) + (k >> (depth - s))` at every sub-stage. So the model checks the real mapping rather than a copy of it.

binary-dft/src/lch.rs
tcoratger: Correct, and the description is being fixed.

At production budgets these sweeps leave the plain branch only for width 64 (2^10 with a 4-byte element, 2^9 with 8-byte, 2^8 with 16-byte) and for 16-byte width 16 at 2^10. Widths 1–8 stay inside one tile budget throughout, and no staged group or row-run executes — the crossings exercise the contiguous tile and the plain passes above it.

The staged path is now pinned by explicit worker counts in the boundary and production tests instead, so it no longer depends on what the test machine reports.

binary-dft/src/lch.rs
tcoratger: Fixed — `BOUNDARY_WIDTHS` is now `[1, 2, 3, 16, 64]`, and the comment is corrected: an 8-byte row is exactly one budget at 2^14, so it takes the plain branch, and `BinaryField32` widths 1–2 and `BinaryField64` width 1 stay there through the heavy sweep. The boundary sweep is what reaches them, and each width now also runs at `t+2` against the walk on the explicit schedules of 1 and 32 workers.

binary-dft/src/lch.rs
tcoratger: You were right on both counts, and re-measuring dissolved the claim.

Taken as a geometric mean over the whole sweep rather than one height, 32/64/128/256 KiB give 1.047/1.016/1.000/0.987 on one worker and 1.071/1.017/1.000/1.032 at 32. So the "256 KiB is 15% faster" point was a single noisy shape, and both columns simply flatten at 128 KiB.

The doc now says exactly that and makes no claim about caches in circulation — which was the unsupportable part, as your Cortex-A55 example shows.

binary-dft/src/lch.rs
tcoratger: Re-measured, and **neither rationale survived**.

32/64/128/256 KiB and 1 MiB give 1.076/1.000/1.013/0.975/0.999 at 16 workers and 1.037/1.000/0.971/0.963/0.964 at 32. There is no fall-off at 1 MiB, so this file's stated reason was wrong, and `poly.rs`'s "exactly the per-core private cache of the machine that measured it" is a coincidence rather than a mechanism.

Both transforms land on 64 KiB as the smallest budget at a flat floor, which is now what the comment says.

binary-dft/src/lch.rs
tcoratger: Fixed — reworded to `poly.rs`'s phrasing: 64 is the smallest line among the supported targets, not the transfer granularity, and Apple Silicon's 128 is named.

binary-dft/src/lch.rs
tcoratger: Fixed — "1 to 16 bytes", since `BinaryField8` and `BinaryField16` run through here too, as this file's own proptests do.

scripts/check.py
tcoratger: Added — a second `slow-regressions` registration of the heavy sweep with `--features parallel`, so the worker clamp and the concurrent tile and staging tasks are exercised.

binary-dft/src/lch.rs
tcoratger: Corrected, and your counter run matched mine.

4 leaves a plain pass, 5 is exactly one fused group, 6 a group plus a plain pass, 7 two groups, 8 two plus a plain pass. `shape(2,4,0)` at 7 clips to a plain pass, so height 8 was added to cover a genuinely clipped fused group.

The thread-count sensitivity you noted is also gone: the worker clamp now lives in `Schedule::for_workers`, so these fixtures run the same stages whatever the test machine reports.

binary-dft/src/lch.rs
tcoratger: Corrected — `2^t` is exactly one budget and runs as plain passes, `2^(t+1)` is always a single plain pass, and `2^(t+2)` is fused only from `STAGED_WORKERS` workers up. The old wording held for one worker only.

binary-dft/src/lch.rs
tcoratger: Agreed — share it, but as a follow-up once both PRs land, to avoid colliding with #2155.

Concrete sketch: a `staging` module holding the `R(t,k)` row map and its bijection model, one `unsafe` gather/scatter over `DisjointMutPtr` generic in `F: Copy + Send + Sync`, and one driver taking the per-transform kernel as a closure. `lch.rs` would pass `tile_stages`; `poly.rs` would pass `tile_stages` wrapped in its entry/exit basis conversion, which is the only thing its copy really adds. That leaves one `unsafe` path and one bijection proof.

I would keep the **plans** separate, though: they answer different questions. `poly.rs` trades run length against depth on a traversal-count objective; this one clamps tile depth to workers.

On aligning the policies, my measurements here bear on yours: the cost of staging is worker-dependent in a way neither PR originally modelled. Staged against tile-only, worst case over every GHASH shape at 1 MiB and up, is 0.61/0.60/0.62/0.62 at 1/2/4/8 workers and only turns at 16. That suggests `poly.rs`'s "stage wide rows at any worker count" is likely wrong below roughly 16 workers for the same reason it was wrong here, and that its `STAGED_LINE_BYTES` / `STAGED_WORKERS` split might be better as a single worker gate over all row widths.

I have not touched `poly.rs`. Worth deciding across both before either lands — happy to take it either way.

PR #2157, opened 2026-09-16
perf(monty-31): fuse x86 dot products of length 5 to 8 into one reduction

tcoratger:
x86 paid two Montgomery reductions for dot products of length 5 to 8 where the scalar and NEON paths pay one. This closes the gap the repository already documents.

## The problem

A 31-bit Montgomery dot product wants to accumulate every 64-bit product and reduce once. Four products is all a 64-bit lane holds, so x86 stopped at `dot_product_4`:

- `N = 5` was `dot_product_4` plus a separate multiply.
- `N = 8` was two `dot_product_4` calls plus a packed add.

`sumcheck/src/strategy.rs` spells this out in a table — `AVX2, AVX-512  2, 3, 4, 64` against `NEON  2, 3, 4, 5, 8, 64` — and sets `const K: usize = 8` for its round tile. That table is updated by this PR.

## The bound

Write a wide accumulator as `C = c_hi * 2^32 + c_lo`.

A group of at most four products of canonical inputs satisfies `C < 4P^2 < 2 * 2^32 P`, so `c_hi < 2P`. A fold subtracts `2^32 P` from any lane at or above it, landing the lane below `2^32 P` and changing the Montgomery result only by a multiple of `P`. That is exactly the `C'` adjustment `dot_product_4` already performs, moved onto the wide accumulator before the halves are split.

With both groups folded the merged sum is below `2 * 2^32 P`, and `2 * 2^32 P < 2^64` because `P < 2^31`. The merged high half is then below `2P`, which is the precondition the shared Montgomery tail already assumed.

Splits are always the first four terms, then the rest:

| N | group B | bound on B | folded |
|---|---|---|---|
| 5 | one product | `P^2 < 2^32 P` | no |
| 6 | two products | `2P^2 < 2^32 P` | no |
| 7 | three products | `3P^2 < 1.5 * 2^32 P` | **yes** |
| 8 | four products | `4P^2 < 2 * 2^32 P` | yes |

Length 7 must fold group B: leaving it would allow `2.5 * 2^32 P`, past `2^64`. **6 and 7 fall out of the same structure**, so all four lengths ship rather than only the two NEON has.

## The carry merge differs from NEON, deliberately

NEON splits both groups into 32-bit halves and re-adds them with an explicit carry mask.

On x86 the fold can be done in the **64-bit domain** instead, because `P << 32` has a zero low half, so the subtraction never disturbs `c_lo`. Once both groups are below `2^32 P`, a plain `vpaddq` merges them and carries out of the low halves for free, and `Q = mu * c_lo mod 2^32` is untouched. No carry mask is needed.

The fold is `vpsubq` + `vpminud` on AVX2, and a mask-compare with a masked `vpsubq` on AVX-512.

## Correctness

A wrong reduction is a wrong field element, and it would not surface as a crash.

The boundary tests are the load-bearing ones and live in `field-testing`, so **every** packed field gets them, NEON included: a maximum-limb test that drives each accumulator to its bound, plus a proptest, wired through the existing dot-product boundary macro. Carry-cascade regression vectors were searched per prime and added for both BabyBear and KoalaBear on both ISAs, matching the ones NEON already carries.

Mutation-checked:

- swapping the merging `vpaddq` for `vpaddd` (dropping the carry) fails 18 tests, including the length-7 and length-8 regressions;
- deleting the fold fails 12, including every maximum-limb case.

Worth stating: **the random and proptest sweeps alone do not catch the missing fold at N = 5 and 6.** Only the maximum-limb test does, which is why it exists.

548 tests pass under both `-C target-cpu=native` (AVX-512) and `-C target-cpu=x86-64-v3` (AVX2).

## Benchmark

Ryzen 9 9950X3D, `taskset -c 0-7`, runs serialized against other work, before and after built from identical bench code into separate target directories, min of two reps. `N = 4` and `N = 64` are untouched controls.

| | AVX-512 before | after | | AVX2 before | after | |
|---|---:|---:|---:|---:|---:|---:|
| N=4 (control) | 1.968 ns | 1.966 ns | −0.1% | | | |
| N=5 | 2.890 ns | 2.404 ns | **−16.8%** | 2.883 ns | 2.474 ns | **−14.2%** |
| N=6 | 3.138 ns | 2.726 ns | −13.1% | 3.201 ns | 2.711 ns | −15.3% |
| N=7 | 4.155 ns | 3.210 ns | **−22.7%** | 4.348 ns | 3.169 ns | **−27.1%** |
| N=8 | 3.845 ns | 3.517 ns | −8.5% | 3.878 ns | 3.450 ns | −11.0% |
| N=64 (control) | 35.40 ns | 35.45 ns | +0.1% | | | |

KoalaBear tracks BabyBear within one point, and every scalar dot product in the same run moved under 0.4%.

Extension multiplies: degree 4 control −0.1%, degree 5 −6.6%, degree 8 −4.6% throughput and −7.2% latency. Scalar degree-8 latency on AVX2 goes 10.87 to 8.83 µs (−18.8%).

## Where it does not show

**The sumcheck benchmark does not move measurably.** `round_coefficients` deltas scatter from −15% to +17% with no consistent direction across `k = 16..22`, both eight-threaded and single-threaded, and `prover` sits within ±2.7% either way. At those sizes the kernel is memory-bound and one saved reduction per eight-term tile is below the noise floor.

The harness is fine — the scalar round-coefficient flavours, which this change cannot touch, moved under 0.2% in the same runs. So this is a kernel and extension-multiply win, not a demonstrated consumer win, and the description says so rather than implying otherwise.

## Follow-ups, deliberately out of scope

- `quintic_mul_packed` and the AVX-512 `octic_mul_packed` build bespoke `dot_product_4` layouts instead of calling `dot_product::<5>` / `::<8>`, which is why most scalar extension multiplies are flat here.
- The `N = 64` arm still uses sixteen `dot_product_4` calls, and the generic `N > 8` loop still steps by four. Both could now step by eight and halve their reductions.


Nashtare:
Probably worth bundling the follow-ups here as well

tcoratger:
> Probably worth bundling the follow-ups here as well

@Nashtare I don't have any other one in the pipeline for now as optimizations :) This was the last one viable that I've extracted with the agents.

baby-bear/src/x86_64_avx2/packing.rs
```diff
@@ -34,4 +37,86 @@ mod tests {
...
+
+        assert_packed_broadcast_dot_product_matches_scalar::<crate::PackedBabyBearAVX2, 5>(
+            lhs, rhs,
+        );
+    }
+
+    #[test]
+    fn dot_product_6_carry_cascade_regression() {
+        // Group A holds terms 0 to 3, group B holds terms 4 and 5.
+        // `hi_A` sits above `P`, so the `2^{32} P` fold has to fire before the merge.
```
Nashtare: For BabyBear this vector does not actually depend on the fold. With group A left unfolded, `A + B` still stays below `2 * 2^32 P`, so the tail returns the right element anyway. Deleting the group-A `fold_wide` in `dot_product_6`, or the group-B one in `dot_product_7`, leaves the matching BabyBear regression test green on both ISAs; only `max_limb_n6` / `max_limb_n7` fail. A lane model says the same holds for the N=8 vector, and for this one under a signed `cmple_epi64_mask`. The KoalaBear vectors do pin every fold. Coverage is fine thanks to `max_limb_*`, but could the BabyBear search add "the unfolded merge reaches `2 * 2^32 P`" as a constraint, or could the comments say the fold fires rather than that it has to?

field-testing/src/packedfield_testing.rs
```diff
@@ -883,6 +883,76 @@ where
...
+    // Both sides maximal: every accumulator sits at its upper bound.
+    assert_packed_broadcast_dot_product_matches_scalar::<PF, N>([max; N], [max; N]);
+
+    // One maximal side against the edge table: a maximal limb mixed with assorted others.
+    for &other in &boundary_u32_values::<PF::Scalar>() {
+        assert_packed_broadcast_dot_product_matches_scalar::<PF, N>([max; N], [other; N]);
+        assert_packed_broadcast_dot_product_matches_scalar::<PF, N>([other; N], [max; N]);
+    }
+
+    // Alternating maximal and unit limbs: consecutive terms land in different groups.
```
Nashtare: The `1` here is the canonical value one, so its limb is `R mod p` (`2^28 - 2` for BabyBear, `2^25 - 2` for KoalaBear), not a unit limb. If a unit limb is what you want, it is `radix.inverse().as_canonical_u32()`, built the same way as `max_limb_value`. The second half of the sentence is also hard to follow: the split is at index 4, so both groups get maximal and non-maximal terms. Could the comment say what this case pins instead?

monty-31/src/x86_64_avx2/packing.rs
```diff
@@ -677,11 +798,138 @@ fn dot_product_4<PMP: PackedMontyParameters, LHS: IntoM256<PMP>, RHS: IntoM256<P
...
+        monty_red_wide_to_canonical::<PMP>(dot_evn, dot_odd)
+    }
+}
+
 /// A general fast dot product implementation.
 ///
-/// Maximises the number of calls to `dot_product_4` for dot products involving vectors of length
-/// more than 4. The length 64 occurs commonly enough it's useful to have a custom implementation
-/// which lets it use a slightly better summation algorithm with lower latency.
+/// Lengths `2` to `8` each get a dedicated routine that pays a single Montgomery reduction.
```
Nashtare: `N = 3` still goes through `dot_product_2` plus a separate `mul`, so it pays two reductions. Maybe: "Lengths `2`, `4` and `5` to `8` each get a dedicated routine with a single Montgomery reduction; `3` adds one product to `dot_product_2`." The same sentence is in the AVX-512 file. The `sumcheck/src/strategy.rs` table also lists `3` and `64` as single-reduction tiles on x86, and neither is one.

baby-bear/benches/bench_field.rs
```diff
@@ -88,6 +88,14 @@ fn bench_packedfield(c: &mut Criterion) {
 
     type PF = <F as Field>::Packing;
     benchmark_chunked_linear_combination::<F, PF, 100>(c, &name);
+
+    // Delayed-reduction dot products: `4` is the control, `5` to `8` each fold the modulus once.
```
Nashtare: In the vocabulary `fold_wide` sets up, `7` and `8` fold twice (once per group) while `5` and `6` fold once. What all four share is a single Montgomery reduction. Suggest "`5` to `8` each pay one Montgomery reduction". The same wording is in `koala-bear/benches/bench_field.rs`, and `chunk_round_step`'s doc in `sumcheck/src/strategy.rs` ("folds the modulus once at this tile size") has the same ambiguity.

field-testing/src/packedfield_testing.rs
```diff
@@ -1114,6 +1184,48 @@ macro_rules! test_packed_field_dot_product_boundary {
             fn lanes_random_n16() {
                 $crate::test_packed_dot_product_lanes_random::<$packedfield, 16>();
             }
+
+            // Maximal Montgomery limbs pin the accumulator bounds the routines rely on.
```
Nashtare, line 1209: The macro's header doc a little further up still says "Two test functions per `N`". It also lists `5` and `8` as specialized and `3`, `6`, `7` as chunk-of-4 fallbacks. With these additions there are four kinds of test, and on x86 `5` to `8` all have dedicated arms. Worth updating that list so it still matches the dispatch it claims to cover.

monty-31/src/x86_64_avx2/packing.rs
```diff
@@ -658,6 +658,127 @@ fn dot_product_4<PMP: PackedMontyParameters, LHS: IntoM256<PMP>, RHS: IntoM256<P
...
+/// The returned pair `(evn, odd)` holds one running sum per parity of field-element index.
+/// Each sum is left unreduced in its 64-bit lane.
+///
+/// Every product is bounded by `(P - 1)^2 < 2^{62}`, so a sum of `K` of them is below `K 2^{62}`.
+/// Four is therefore the largest `K` that stays under `2^{64}`, so `K` must lie in `1..=4`.
+///
+/// If the inputs are not in canonical form, the result is undefined.
+#[inline(always)]
+#[must_use]
+fn wide_dot<PMP: PackedMontyParameters, LHS: IntoM256<PMP>, RHS: IntoM256<PMP>, const K: usize>(
```
Nashtare, line 679: Every caller passes a fixed-length prefix or suffix. Taking `lhs: &[LHS; K], rhs: &[RHS; K]` and replacing the three `debug_assert!`s with `const { assert!(K >= 1 && K <= 4) }` would turn the length check and the `K <= 4` overflow bound into compile-time checks instead of debug-only ones. The call sites would use `lhs[..4].try_into().unwrap()`, which folds away after inlining. The same applies to the AVX-512 copy. Nit: this doc block is also missing a blank line after `dot_product_4`'s closing brace, in both files.

Nashtare (approved):
I checked the arithmetic independently and it holds. Each group of at most four products of canonical limbs is below `4P^2 < 2^33 P`, so `c_hi < 2P` when `fold_wide` runs. The AVX2 `vpsubq` + `vpminud` pair leaves `c_lo` alone and computes `c_hi mod P`, and the AVX-512 predicate is the unsigned `cmple_epu64`, which it has to be since the sums pass `2^63`. Group B for N=5 and N=6 stays below `2^32 P` without a fold. So the merge is below `2^33 P < 2^64`, and the shared tail's `c_hi < 2P` precondition holds. `P < 2^31` is const-asserted in `MontyField31::new`.

The AVX-512 leg is build-only in CI, so I ran both ISAs locally:
- AVX2 under Rosetta: 4.2M lane results per field, covering `dot_product` and `mixed_dot_product` for N=4..8 and 12, with limb-space inputs aimed at the fold and tail boundaries.
- AVX-512 under Miri: 4,064 lane results per field, plus your boundary tests.

Everything matches a `u128` reference. Your tests and my probe both catch a dropped fold, a signed compare, and a `vpaddd` merge.

The inline comments are mostly doc wording. The one on the BabyBear regression vectors is about what they pin, not a bug.

On bundling the follow-ups: I meant the ones your description lists, not new ideas. Now that `dot_product_8` exists, the `N = 64` arm and the generic `N > 8` loop could step by eight and halve their reductions, and `quintic_mul_packed` / the AVX-512 `octic_mul_packed` could call `dot_product::<5>` / `::<8>`. They're small enough to go here, or in a separate PR.

baby-bear/src/x86_64_avx2/packing.rs
tcoratger: You are right, and I took the first option: the BabyBear search now carries "deleting any single fold changes the result" as a constraint, so the comment's claim is the thing that was searched for.

Confirming your reading first, with a lane model of the AVX2 and AVX-512 tails (they agree exactly — the mask compare and the `vpsubq`/`vpminud` pair compute the same function on these ranges) validated against `(sum l_i r_i) * 2^-32 mod p` on 20k random draws per length per prime:

```
n=5: drop fold A -> still correct? no
n=6: drop fold A -> still correct? yes
n=7: drop fold A -> yes    drop fold B -> yes
n=8: drop fold A -> yes    drop fold B -> yes
```

So `6`, `7` and `8` were as you describe, and `5` happened to break. The reason is the one your comment implies: folding is neutral modulo `P`, so a dropped fold only shows up if the merged lane passes `2^64` or its high half passes `2P`, and for the `7` vector the high half landed at `2.026 P`, where the result is still congruent and still canonical unless `q_p` happens to be below `hi - 2P`.

The four BabyBear vectors are re-searched against all of:

- every fold the routine performs fires,
- deleting any single fold changes the result,
- the low halves of the merged groups carry past `2^32`,
- the merged high half still exceeds `P`.

Deleting one fold now overflows `2^64` on every length, so the failure mode is unambiguous rather than dependent on `q_p`:

```
        hi_A     hi_B     one fold dropped    merged hi
n=5   1.748 P  0.387 P    1.001 * 2^64        1.135 P
n=6   1.545 P  0.689 P    1.047 * 2^64        1.234 P
n=7   1.852 P  1.381 P    1.047 * 2^64        1.233 P
n=8   1.509 P  1.741 P    1.055 * 2^64        1.250 P
```

Your other observation holds too: the KoalaBear vectors already satisfy every one of these constraints, including the overflow, so they keep their values and only gain the sharper comment. Each test now states its own numbers instead of asserting the fold "has to" fire.

Mutation matrix over both ISAs, one fold deleted at a time (6 deletions x 2 ISAs), running only that length's regression test across both primes:

```
AVX2    dot_product_5 drop A -> FAILED=2 ok=0
AVX2    dot_product_6 drop A -> FAILED=2 ok=0
AVX2    dot_product_7 drop A -> FAILED=2 ok=0
AVX2    dot_product_7 drop B -> FAILED=2 ok=0
AVX2    dot_product_8 drop A -> FAILED=2 ok=0
AVX2    dot_product_8 drop B -> FAILED=2 ok=0
AVX512  (identical, all six FAILED=2 ok=0)
```

Nothing survives now, where before this the `6`, `7` and `8` deletions left the BabyBear test green.

field-testing/src/packedfield_testing.rs
tcoratger: Correct on both counts, and the consequence was worse than a wrong comment: the case was a duplicate.

`1` stores the limb `R mod p`, which is `0x0ffffffe` for BabyBear and `0x01fffffe` for KoalaBear. Alternating that against maximal limbs puts a group of four at

```
BabyBear   limb 0x0ffffffe   hi_A = 1.0625 P   fold fires
KoalaBear  limb 0x01fffffe   hi_A = 1.0078 P   fold fires
```

so the fold fired, exactly as in the all-maximal case just above it. KoalaBear was 0.8% away from landing on the other branch by accident.

A true unit limb is built the way you describe, and the radix is now a shared helper both limb targets go through:

```rust
fn monty_radix<F: PrimeField32 + QuotientMap<u32>>() -> F {
    F::from_int(((1u64 << 32) % F::ORDER_U64) as u32)
}

fn unit_limb_value<F: PrimeField32 + QuotientMap<u32>>() -> u32 {
    monty_radix::<F>().inverse().as_canonical_u32()
}
```

With that the same alternation lands at `0.9375 P` and `0.9922 P`, just under the threshold, which is the branch where the conditional subtraction must be a no-op. It is not a per-prime accident either: a group holding two maximal products and two unit-limb ones is `2 (p - 1)^2`, and `2 (p - 1)^2 / 2^32 < p` follows from `2p < 2^32`, so it holds for every prime this code accepts.

On the second half of the sentence: agreed, the split-at-4 wording said nothing. The comment now says what the case pins, with that bound written out:

```rust
    // Alternating maximal and unit limbs, against an all-maximal other side.
    //
    // A group of four then holds two maximal products and two negligible ones:
    //
    //     hi  =  2 (p - 1)^2 / 2^{32}  <  p       because 2p < 2^{32}
    //
    // So this is the case where every conditional subtraction must be a no-op.
    // The all-maximal case above is the one where each of them must fire.
```

Checked that the branch has teeth: making `fold_wide` subtract unconditionally fails `max_limb_n5` through `max_limb_n8` on both primes, while `max_limb_n4` stays green as expected.

monty-31/src/x86_64_avx2/packing.rs
tcoratger: Right on all three. `3` goes through `dot_product_2` and then `+ (lhs2.into() * rhs2.into())`, which is a second reduction, and `64` is sixteen `dot_product_4` calls into `sum_array`, so it is sixteen.

The dispatch doc in both x86 kernels now reads:

```
/// Lengths `2`, `4` and `5` to `8` each get a dedicated routine with one Montgomery reduction.
/// Length `3` adds one product on top of the length two routine, so it pays two.
///
/// Every longer length is cut into as many length four blocks as possible, one reduction each.
/// The length 64 occurs commonly enough that it gets a custom implementation.
```

For the `sumcheck` table I checked each target rather than just dropping the two you named:

- scalar: `2` to `8` each reduce once, and the `_` arm accumulates into a `u128` and calls `monty_reduce_u128` once, so every length reduces once.
- x86: `2`, `4`, `5`, `6`, `7`, `8`. Not `3`, not `64`.
- NEON: `2`, `3`, `4`, `5`, `8`. `3` belongs there because it zero-pads into `dot_product_4`. Not `64`.

```
///     scalar          every length
///     AVX2, AVX-512   2, 4, 5, 6, 7, 8
///     NEON            2, 3, 4, 5, 8
```

The conclusion the table exists to support, that `8` is a single-reduction tile everywhere, is untouched.

baby-bear/benches/bench_field.rs
tcoratger: Agreed, and the ambiguity was real: `5` and `6` fold one group, `7` and `8` fold two, so "fold the modulus once" is false for half the set. The one Montgomery reduction is what all four share, so that is what the three places say now.

Both bench files:

```rust
    // Delayed-reduction dot products, with `4` as the control.
    // Lengths `5` to `8` each pay one Montgomery reduction.
```

And `chunk_round_step`:

```rust
/// A Monty-31 packing pays one Montgomery reduction at this tile size on every target.
```

I dropped the `(see `K`)` pointer from that line at the same time, since the line above it already sets the context.

field-testing/src/packedfield_testing.rs
tcoratger: Both stale, yes. There are four kinds of test now, over four different length sets, and the dispatch list was wrong for x86 even before this PR added arms. The header lists both, per target:

```rust
/// Four kinds of test, each over its own set of lengths:
/// - edge-pair sweep over an eight-value table, lengths `1` to `9` and `16`,
/// - per-lane edge-biased random, lengths `2` to `9` and `16`,
/// - maximal Montgomery limbs, lengths `4` to `8`,
/// - uniform canonical proptest, lengths `5` to `8`.
///
/// Between them those lengths hit every dispatch arm a 31-bit Monty packing offers:
/// - a bare product (`1`),
/// - one fused reduction (`2` and `4` to `8` on x86, `2` to `5` and `8` on NEON),
/// - two routines summed (`3` on x86, `6` and `7` on NEON),
/// - the chunk-of-four loop (`9`, `16`).
```

The length sets are read off the generated test functions: the sweep has `n1` through `n9` and `n16`, the per-lane random the same minus `n1`, the maximal-limb tests `n4` through `n8`, the proptest `n5` through `n8`. Splitting the arms by target also removes the old list's implicit claim that one arm set covers every backend, which was never true for `3`.

monty-31/src/x86_64_avx2/packing.rs
tcoratger: Done in both files, and the blank line is in.

```rust
fn wide_dot<PMP: PackedMontyParameters, LHS: IntoM256<PMP>, RHS: IntoM256<PMP>, const K: usize>(
    lhs: &[LHS; K],
    rhs: &[RHS; K],
) -> (__m256i, __m256i) {
    const { assert!(K >= 1 && K <= 4) }
```

One deviation from your sketch: the call sites pass array literals rather than `lhs[..4].try_into().unwrap()`, so there is no `unwrap` on a library path and it matches how the dispatch right below already builds its arguments.

```rust
let (a_evn, a_odd) = wide_dot::<PMP, _, _, 4>(
    &[lhs[0], lhs[1], lhs[2], lhs[3]],
    &[rhs[0], rhs[1], rhs[2], rhs[3]],
);
let (b_evn, b_odd) = wide_dot::<PMP, _, _, 1>(&[lhs[4]], &[rhs[4]]);
```

I checked the codegen rather than trusting that it folds away, because a criterion A/B on this machine was useless (the `4` and `64` controls, which this cannot touch, moved +35% and +30%). Comparing the emitted assembly of the whole `bench_field` binary instead, slice form against array form:

- `-C target-cpu=native` (AVX-512): byte-identical, 181231 lines.
- `-C target-feature=+avx2`: 71 differing lines out of 190465, all inside a single symbol, criterion's `benchmark_chunked_linear_combination` harness. Register allocation and scheduling only, net one instruction fewer (`+2 vpbroadcastd`, `-2 vpbroadcastq`, `-1 vpshufd`). Every dot-product routine is identical.

The `K` bound is a post-monomorphisation error now, so a bad length fails the build rather than a debug assertion, and the doc says so.

PR #2158, opened 2026-09-16
perf(binary-pcs): change the fold's basis a block at a time

Nashtare:
## The problem

#2153 added `poly_basis::{from_tower_slice, to_tower_slice}`, which convert whole blocks of `GF(2^128)` through GFNI. Its description lists `binary-pcs/src/fold.rs` as not wired, calls it the largest remaining win, and says it belongs in its own PR. This is that PR.

- `fold_codeword` converted each lane inside the packed multiply loop, two conversions in and one out per output symbol.
- `fold_codeword_batch` did the same per coset. It also ran each coset on its own, so its rounds shorter than the packing width dropped to a scalar loop.

## The change

A fold now works on **blocks of adjacent cosets** held in polynomial coordinates:

```text
    copy block into scratch  ->  from_tower_slice(block)
    round 0 .. a-1           ->  one kernel call across the whole block, alternating two buffers
    last round               ->  written straight into the task's output
    once per task            ->  to_tower_slice(output)
```

Round `r` pairs symbols exactly as a single fold of the round-`r` codeword does. So `n` adjacent cosets starting at coset `first` stay a run of adjacent symbols starting at output index `first * 2^(a-1-r)`, and one call covers the block with no seam at a coset boundary.

A block holds `2 * FOLD_GRAIN / 2^a` cosets, never fewer than one packed group, so even its last round fills whole groups.

`fold_codeword` and `fold_codeword_batch` now share this driver (`fold_rounds`). The Cantor-domain fold equation, the group-base walk and the output bytes are unchanged.

## Benchmarks

`fold_codeword` is the existing bench. `fold_codeword_batch` is crate-private, so its rows come from a local bench that exposes it; that bench is not part of this PR. Codeword length is `2^(vars + 2)`.

"Base" is #2153's head. `binary-pcs/src/fold.rs` there is identical to `main`.

### AMD EPYC 9R45 (Zen 5), `c8a.4xlarge`, `-C target-cpu=native`

Serial, pinned to one core, two passes each (agreeing within 1%). `-gfni` is this branch built without GFNI, which separates the blocked kernel from the loop structure.

| bench | vars | base | this PR | `-gfni` | vs base | GFNI's share |
|---|---:|---:|---:|---:|---:|---:|
| `fold_codeword` | 16 | 1.744 ms | 0.497 ms | 1.279 ms | **3.51x** | 2.57x |
| | 18 | 7.20 ms | 2.09 ms | 5.23 ms | **3.44x** | 2.50x |
| | 20 | 36.16 ms | 9.85 ms | 22.40 ms | **3.67x** | 2.27x |
| batch, arity 2 | 16 | 3.258 ms | 0.539 ms | 1.190 ms | **6.05x** | 2.21x |
| | 18 | 13.06 ms | 2.19 ms | 4.79 ms | **5.96x** | 2.19x |
| | 20 | 52.31 ms | 10.15 ms | 20.61 ms | **5.15x** | 2.03x |
| batch, arity 4 | 16 | 2.018 ms | 0.562 ms | 1.116 ms | **3.59x** | 1.99x |
| | 18 | 8.06 ms | 2.27 ms | 4.50 ms | **3.55x** | 1.98x |
| | 20 | 32.30 ms | 10.51 ms | 19.58 ms | **3.07x** | 1.86x |

`--features parallel`, 16 threads, two passes each:

| bench | vars 16 / 18 / 20 | base | this PR | speedup |
|---|---|---|---|---|
| `fold_codeword` | | 202 µs / 683 µs / 2.91 ms | 111 µs / 352 µs / 1.53 ms | **1.83x / 1.94x / 1.91x** |
| batch, arity 2 | | 285 µs / 1.015 ms / 3.83 ms | 95 µs / 286 µs / 1.12 ms | **3.01x / 3.55x / 3.42x** |
| batch, arity 4 | | 148 µs / 618 µs / 2.27 ms | 54 µs / 216 µs / 922 µs | **2.75x / 2.86x / 2.46x** |

### Apple M4 Pro, default target, serial

No GFNI here, so the blocked kernel never runs. Base and this PR ran back to back in the order base, PR, base, PR.

| bench | vars 16 / 18 / 20 | base | this PR | change |
|---|---|---|---|---|
| `fold_codeword` | | 2.69 / 11.11 / 42.27 ms | 1.29 / 5.10 / 20.44 ms | −52% / −54% / −52% |
| batch, arity 2 | | 4.29 / 16.19 / 65.70 ms | 1.29 / 5.13 / 20.48 ms | −70% / −68% / −69% |
| batch, arity 4 | | 2.80 / 11.28 / 44.59 ms | 1.29 / 5.19 / 21.10 ms | −54% / −54% / −53% |

The loop-structure gain on both machines, the `-gfni` column above and all of the M4 rows, is most likely the table working set. The old loop kept both 128 KiB conversion tables live at once, one per direction. The new loop touches one table while it converts a block in and the other while it converts a task's output back. I have not profiled to confirm this.

## Correctness

- **Width-generic model.** `the_lane_arithmetic_holds_at_every_packing_width` now drives the production `fold_rounds` at widths 1, 2, 4 and 8, with only the kernel replaced by an array model. That covers widths a host cannot compile. The sweep includes batch arities 1–5 and 9–12, which reach the minimum block size at every width, fewer cosets than one group, several blocks per task, and several tasks.
- **Mutants.** Four deliberate bugs in the driver fail that test: dropping the block offset, dropping the task offset, using the round-0 output scale in every round, and removing the minimum block size. The last is caught by the model's alignment `debug_assert`, since results stay correct without it.
- **Every batch shape.** A proptest over `log_len` in `1..=14` and arity in `1..=log_len` compares `fold_rounds` at widths 1, 2, 4 and 8 against the tower reference. It catches all four mutants on its own.
- **No challenges.** `fold_rounds` refuses an empty challenge list with `assert!`, in every profile, since no round would write an output slot.
- **Batch vs scalar.** `batched_folds_match_sequential_folds_at_every_coset_offset` also compares every coset against `fold_coset`, the verifier's per-pair path in the tower basis. Its arities now reach the minimum block size too.
- **Real hardware.** On the Zen 5 host, with GFNI, AVX-512 and VPCLMULQDQ live, `p3-binary-field` passes 184 + 171 + 4 + 6 and `p3-binary-pcs` passes 64 + 27, both serial and with `parallel`.
- **Miri.** On `x86_64` with `+gfni,+avx512f,+avx512bw,+pclmulqdq,+vpclmulqdq`, a probe asserted the 4-lane packing and compared `fold_codeword_batch` at arities 1 and 3 against `fold_coset`, over 512 symbols so both conversions take whole blocks. It passes with no UB. With the GFNI block matrices corrupted, it fails at the first coset.
  - Miri cannot run #2153's `vpternlogq`, so for this probe only, the XOR3 sums were rewritten as chained `_mm512_xor_si512`.
- **Other targets.** Tests also pass on `x86_64` under Rosetta with `+pclmulqdq`, and on `aarch64` in debug and in release with `parallel`.
- **Lints.** Clippy is clean for the default target, `--features parallel` and `x86_64` with `+avx512f,+avx512bw,+gfni,+vpclmulqdq,+pclmulqdq`. Nightly rustfmt is clean.

## Where it costs something

- **Test time.** The width-model test goes from 0.10 s to 0.46 s in debug, and the batch-shape proptest adds about 0.3 s.
- **Scratch memory.** Each worker holds one block of `max(2 * FOLD_GRAIN, W << a)` symbols, plus half that again when `a > 1`. That is 32 KiB + 16 KiB until `2^a` passes `2 * FOLD_GRAIN / W`, and grows with the arity after that: 512 KiB + 256 KiB at `W = 8`, `a = 12`. The floor of one packed group of cosets keeps a block's last rounds on whole packed groups.
- **CI coverage.** No CI leg enables `gfni`, so CI never compiles the blocked path this PR routes through. That gap is the same one noted on #2153.


binary-pcs/src/fold.rs
```diff
@@ -364,82 +350,141 @@ pub(crate) fn fold_coset(
...
+/// Panics in debug builds if there are no challenges.
+fn fold_rounds<const W: usize, K>(
+    codeword: &[BinaryField128],
+    challenges: &[BinaryField128],
+    kernel: &K,
+) -> Vec<BinaryField128>
+where
+    K: Fn(&[u128], &mut [u128], usize, Ghash128, &[Ghash128]) + Sync,
+{
+    debug_assert!(!challenges.is_empty(), "a fold runs at least one round");
```
tcoratger: Consider `assert!` here rather than `debug_assert!`.

With no challenges the round loop never runs, so `slots` keeps the zeros from `vec![0u128; num_cosets]`, `to_tower_slice` maps zeros to zeros, and a release caller gets a silently wrong all-zero codeword rather than a panic. That is the one failure mode in this function that does not announce itself.

Both entry points already guard it, so this only bites a future third caller. But the cost is one branch per whole-codeword fold, invisible next to the fold, and it would make the `FOLD_GRAIN.is_multiple_of(block)` assertion below provably true rather than separately checked: `arity >= 1` is exactly what gives `block <= FOLD_GRAIN`.

binary-pcs/src/fold.rs
```diff
@@ -364,82 +350,141 @@ pub(crate) fn fold_coset(
...
-                })
-                .collect()
-        })
+        .map(|round| group_steps::<W>(codeword.len() >> (round + 1)))
         .collect();
-    let mut output = BinaryField128::zero_vec(num_cosets);
-    output
+
+    // Cosets per block, a power of two no larger than the grain, so every task starts a block.
+    let block = (BLOCK_SYMBOLS >> arity).max(W);
```
tcoratger, line 450: The scratch a worker holds is `block * size` symbols, so the `.max(W)` floor makes it `W << arity` once `2^arity` passes `BLOCK_SYMBOLS / W`, growing with the arity instead of staying at the constant. At `W = 8`:

```text
    arity <= 8    2048 symbols      32 KiB   (+16 KiB for odd)
    arity 11      16384 symbols    256 KiB   (+128 KiB)
    arity 12      32768 symbols    512 KiB   (+256 KiB)
```

The `BLOCK_SYMBOLS` doc and the PR's cost section both quote the 32 KiB figure, and the "while it is still in cache" line stops holding somewhere around arity 11. `log_folding_factor` is only bounded by `num_variables` (`params.rs:143`), and `ARITIES` already exercises 9..=12, so this is reachable rather than hypothetical.

To be clear, I am not suggesting dropping the floor. It is load-bearing: below `W` cosets `first` stops being a multiple of `W`, the lane-offset identity `g AND k == 0` fails and results go wrong, which is exactly what your removed-minimum mutant catches. Only that the documented bound could be the real one, `max(2 * FOLD_GRAIN, W << arity)` symbols, so a reader picking a large folding factor sees the cost.

binary-pcs/src/fold.rs
```diff
@@ -688,11 +735,39 @@ mod tests {
...
                 "W={W} len={len}"
             );
         }
+
+        // Batches, where every round after the first runs on symbols the previous round wrote:
+        //
+        //     2^1, 2^3, 2^6   fewer cosets than a packed group holds at the wider widths
+        //     2^12            two tasks at arity 1, several blocks at arity 5, one coset at 12
+        //     2^14, arity 2   several tasks of several blocks each
+        let shapes = [1, 3, 6, 12]
```
tcoratger, line 764: Optional, take it or leave it. The enumeration is deliberate and I could not find a shape it misses that changes a result.

One cheap addition if you want it: a proptest over `log_len in 1..=14` and `arity in 1..=log_len` against the same tower reference would reach the lengths between the picked ones, where `num_cosets` crosses `FOLD_GRAIN` at a different arity. `log_len = 11, arity = 1` and `log_len = 13, arity = 3` both give exactly one full task, which the current list steps over: `(12, 1)` gives two tasks and `(12, 5)` gives one partial one.

The property is lightweight and total here, which is the case where proptest earns its keep over more hand-picked rows.

tcoratger (approved):
Approving. I checked the index algebra, the buffer ping-pong and the alignment argument line by line and found nothing that changes a result; the three comments below are a cost note, a hardening suggestion and an optional test idea.

## What I verified

**The block index identity.** The claim the whole design rests on is that a block of `n` cosets starting at coset `first` stays a contiguous run at every round, at output index `first * 2^(a-1-r)`. That holds, and it lines up with the verifier: `fold_coset` folds with global index `coset_index * len + i`, and for the coset at offset `o` in a block the driver gives `first * 2^(a-1-r) + o * 2^(a-1-r) + i = (first + o) * 2^(a-1-r) + i`. Same point, so prover and verifier cannot drift.

**Alignment.** `start` is a multiple of `W` at every round, which is what the lane-offset identity needs. `block` is a power of two at least `W`, `FOLD_GRAIN` is a multiple of `W`, so `first = task * FOLD_GRAIN + index * block` is a multiple of `W`, and earlier rounds only scale it up. The partial-block cases cannot be misaligned either: `num_cosets` is a power of two, so a task's output chunk is either a full grain or the whole (power-of-two) output, and `block` divides it or covers it whole.

**Step tables.** `group_steps::<W>(codeword.len() >> (round + 1))` is sized by the round's total pair count while `fold_pairs` indexes with the *global* group index, so the two agree. The largest index reached is `(num_pairs / W) - 1`, and since `num_pairs / W` is a power of two its trailing-zero count stays inside the table. The `num_pairs < W` case yields an empty table and is never indexed.

**Buffer ping-pong.** Round `r` reads exactly what round `r - 1` wrote: `odd` only ever needs `symbols.len() / 2`, `even` is re-`extend`ed to `symbols.len()` per block, and the `arity == 1` case leaves `odd` empty and unread because the first round is also the last.

**The final `collect`.** `BinaryField128` is `#[repr(transparent)]` over `u128`, so `Vec<u128> -> Vec<BinaryField128>` takes the in-place collect specialization and `from_repr` is the identity. No extra allocation, no copy. (I went looking for a regression here and there isn't one.)

## The CI gap, closed on my side

My host has `gfni`, `avx512f`, `avx512bw` and `vpclmulqdq`, so I ran the path no CI leg compiles, on a machine other than your Zen 5:

- `-C target-cpu=native`, `p3-binary-pcs`: 64 + 4 + 27 pass, serial and with `--features parallel`.
- `-C target-cpu=native`, `p3-binary-field`: 184 + 171 + 4 + 6 pass.
- Default target: same suites pass.
- Clippy clean on native with `--all-targets --features parallel`.

I also reproduced the headline numbers independently with the in-repo bench, `-C target-cpu=native`, serial:

| `fold_codeword` | base (`dba1e28`) | this PR | speedup |
|---|---:|---:|---:|
| 16 | 1.418 ms | 0.418 ms | 3.39x |
| 18 | 6.007 ms | 1.772 ms | 3.39x |
| 20 | 26.98 ms | 8.59 ms | 3.14x |

Same shape as your 3.51x / 3.44x / 3.67x on different silicon.

On the CI leg you flag: worth noting it is not new exposure from this PR. `binary-dft/src/poly.rs` already routes through `poly_basis::{from,to}_tower_slice` on `main`, so the prover's critical path depends on the blocked kernel today. This PR widens that rather than creating it, which I think supports your read that the CI leg belongs in its own change.

PR #2159, opened 2026-09-16
perf(sumcheck): fold pinned zerocheck blocks without multiplying

tcoratger:
Pinning the zerocheck's tail coordinates to the geometric progression of Flock §4.3 makes every equality weight one constant times one power of the indeterminate, `C * x^k`. In the GHASH polynomial basis a power of the indeterminate is a shift, so the fold needs no multiplication at all. `PinnedEqWeights::geometric` already built that weight set; nothing exploited its structure.

Three pieces:

- `Ghash128::dot_powers_of_x` — combines values against the successive powers with a single reduction. It picks its route at compile time: a deferred dot product where the target has a carryless multiply, the shift loop where it does not. Neither needs a weight table.
- `GeometricEqWeights` — the same pinned point as the general constructor, but stored as a count and a constant rather than a weight table. The constant is `1 / sum_{k < 2^d} x^k` in closed form, and it factors out of the sum entirely: `fold_unscaled` is the inner-loop shape and the caller applies the constant once.
- `fold_bits` / `fold_bits_unscaled` — the case the soundness argument already requires. Pinning is only sound when the folded cells are prime-field-valued, and for bit-valued cells `sum_k b_k x^k` *is* the block's bit pattern read as a field element. The fold is a reinterpretation.

## Measurements

**before** is what `PinnedEqWeights::fold` does today: walk the stored weight table, one product per cell, each reduced on its own. **after** is `GeometricEqWeights::fold_unscaled`. The last column is a reference point that is neither: a deferred dot product against a *materialized* table of 128 field elements, which this weight type deliberately does not keep.

AMD Ryzen 9 9950X3D (Zen 5), criterion medians, one block of N cells.

With a hardware carryless multiply (`-C target-cpu=native`):

| N | before | after | **gain** | *(ref: deferred, table)* |
|---|---|---|---|---|
| 8 | 18.96 ns | 11.58 ns | **1.64x** | *9.66 ns* |
| 16 | 34.94 ns | 21.07 ns | **1.66x** | *17.79 ns* |
| 32 | 68.09 ns | 41.05 ns | **1.66x** | *34.86 ns* |
| 64 | 135.6 ns | 81.24 ns | **1.67x** | *69.00 ns* |
| 128 | 271.6 ns | 159.5 ns | **1.70x** | *136.9 ns* |

Without one (`-C target-feature=-pclmulqdq,-vpclmulqdq,-avx2,-avx512f`):

| N | before | after | **gain** | *(ref: deferred, table)* |
|---|---|---|---|---|
| 8 | 185.2 ns | 12.76 ns | **14.5x** | *173.9 ns* |
| 16 | 368.4 ns | 23.30 ns | **15.8x** | *345.9 ns* |
| 32 | 737.6 ns | 44.47 ns | **16.6x** | *690.9 ns* |
| 64 | 1.478 µs | 86.70 ns | **17.0x** | *1.378 µs* |
| 128 | 2.948 µs | 171.8 ns | **17.2x** | *2.751 µs* |

A residual 1.17x to the table-based path remains on the hardware leg and is not closable here: that path reads 128 stored field elements, and the point of this weight type is that it stores two words. Generating the powers is the difference.

Where this actually pays is the bit case. 1024 blocks of 128 cells, the same cells in both layouts, all four arms measured in one run on the hardware leg:

| | time | per cell | **vs before** |
|---|---|---|---|
| **before** — one element per cell | 277.8 µs | 2.12 ns | 1x |
| after — one element per cell | 163.3 µs | 1.25 ns | **1.70x** |
| **after** — one bit per cell | 0.179 µs | 0.0014 ns | **1553x** |
| *(ref: deferred against a table)* | *139.9 µs* | *1.07 ns* | *1.99x* |

and the witness for those cells goes from 2 MiB to 16 KiB.

## Tests

`GeometricEqWeights` is pinned to `PinnedEqWeights::geometric` at every width 0..=7 — same coordinates, same block length, same fold — which is what fixes the block ordering against the equality table. The kernel is proptested against a multiplying reference, plus boundary runs at 0, 1, x^127 and all-ones at full width, an additivity property, and the length refusal. The bit fold is checked against the spelled-out block at every width, and refused at three separate out-of-block bits: the first one out, the top of the word, and the first bit above a half-word block. 507 tests in the crate pair, green on the hardware and portable legs.

Ran: `cargo test -p p3-binary-field -p p3-sumcheck` (both legs), `cargo clippy -p p3-binary-field -p p3-sumcheck --all-targets` (clean), `cargo +nightly fmt --all -- --check`, and `cargo build -p p3-binary-pcs -p p3-multi-stark -p p3-binary-dft`.

## Not done here

No consumer is wired up — the module still has none, as its header says. The two-level F_2^8/F_2^128 variant from Flock needs an AES-basis byte field that is not in tree. No NEON numbers are quoted anywhere: this machine is x86-64 with no emulator, so I have not measured that leg and do not claim it.


binary-field/src/ghash.rs
```diff
@@ -89,6 +89,28 @@ impl Ghash128 {
...
+    ///
+    /// The modulus is folded in once at the end, not once per term.
+    ///
+    /// # Panics
+    ///
+    /// Panics on more values than the field has bits.
+    ///
+    /// That ceiling is where the last shift would leave the word.
+    #[inline]
+    pub fn dot_powers_of_x(values: &[Self]) -> Self {
```
Nashtare, line 121: By the description's Zen 5 table, with PCLMULQDQ on, `dot_product` against the materialized powers is 1.43x faster than this at N = 128 (138 ns vs 197 ns). This path only wins on the software backend. Using the faster path there needs no new API: `clmul::poly_dot_128` is crate-private and takes an iterator of pairs. This could feed it `(v, 1 << k)` when `HAS_HARDWARE_CLMUL` is set and keep the shift loop otherwise. Could it choose per target? If not, could this doc and `fold_unscaled`'s "the shape the inner loop wants" say the win is on targets without a carryless multiply? I have no numbers for the NEON `poly_dot_128` either way.

sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -304,6 +304,219 @@ impl<F: Field> PinnedEqWeights<F> {
...
+    /// Combine one block, leaving out the constant factor.
+    ///
+    /// This is the shape the inner loop wants: shifts and exclusive ors, one reduction.
+    ///
+    /// The result is short of one factor, which the caller applies once to the sum it feeds.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the block length differs from the block width.
+    pub fn fold_unscaled(&self, block: &[Ghash128]) -> Ghash128 {
```
Nashtare, line 517: Nit: `fold`, `fold_unscaled` and `fold_blocks_unscaled` take whole `Ghash128` cells, and on those the map has a large kernel. Local probe: at d = 1, `[x, 1]` folds to zero. The obligation is in the struct docs, and `PinnedEqWeights::fold` has the same shape, so this is not new. Could these three docs add one line saying the cells must be 0 or 1 for a pinned zerocheck, and point to `fold_bits` as the form where that holds by construction?

sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -304,6 +304,219 @@ impl<F: Field> PinnedEqWeights<F> {
...
+
+                // That pattern is never one, so the denominator here is never zero.
+                u * (Ghash128::ONE + u).inverse()
+            })
+            .collect()
+    }
+
+    /// Combine one block of bit-valued cells, leaving out the constant factor.
+    ///
+    /// Cell `k` is bit `k` of the argument.
```
Nashtare, line 444: Nit: `embed_bits` packs cells least significant bit first within each byte. So an 8-cell block is the byte itself, and a 128-cell block is `u128::from_le_bytes` of its 16 bytes (equivalently `Ghash128::from_le_bytes`). I checked that end to end at every width. One line here would help whoever wires up the consumer, since `from_be_bytes` would pair cells with the wrong powers.

sumcheck/src/univariate_skip/pinned.rs
```diff
@@ -739,6 +952,170 @@ mod tests {
...
+    #[should_panic(expected = "one bit per weight")]
+    fn a_bit_outside_the_block_is_refused() {
+        // Mutation: set the bit one place above the block.
+        //
+        //     3 pinned coordinates -> cells 0..8, so bit 8 names no cell
+        //
+        // Accepting it would silently weigh a cell the block does not have.
+        let _ = GeometricEqWeights::new(3)
+            .unwrap()
+            .fold_bits_unscaled(1 << 8);
```
Nashtare, line 1098: Nit: this test only sets the bit just above the block. By reading, a check that looked only at bit `block_len`, or only at the low 64 bits, would still pass here and in the proptest (which masks its input first). Could it also try bit 127, and bit 64 at d = 6?

Nashtare (approved):
Reviewed at 91c6a0b5. I worked the kernel, the constant and the bit order by hand and checked them against an independent bit-serial multiply. Nothing blocks. The one real question is which path `dot_powers_of_x` takes when a hardware carryless multiply is available.

**What holds**
- **Bit convention.** Bit `i` is the coefficient of `x^i` (not the GCM reflection), so `x` is `from_repr(2)` and `x^127 * x` is `0x87`.
- **Accumulator.** Term `k` puts `v << k` in the low half and `v >> (128 - k)` in the high half. `(v >> 1) >> (127 - k)` equals that for `1 <= k <= 127` and is 0 at `k = 0`. The `assert!` holds `k <= 127` in every profile, and the top degree is 254, so no bit is lost. `reduce_128` is exact for any high half: the first fold spills at most degree 6 past `x^128`, and the second lands at degree 13 or below.
- **Probe.** 2,708 vectors at every length 0..=128 match `sum v_k * x^k` from a bit-serial multiply mod `x^128 + x^7 + x^2 + x + 1`. They include all-ones, `x^127`, `0x87` and `u128::MAX << 121` runs. All 16,384 single-bit inputs at length 128 match too. Both hold on the PMULL backend and with `-C target-feature=-aes`, and a length of 129 panics.
- **Ratio and ordering.** `geometric` telescopes to `C * g^k`, with the tail coordinate `g` as the index LSB. `GeometricEqWeights::challenges` uses `x^(2^i)` reversed. So cell `k` gets `C * x^k`, which is the power the kernel shifts by. An end-to-end probe agrees at d 0..=7 with up to 3 free variables (cells from `embed_bits`, whole-point equality sum vs the free-part sum of `fold_bits`).
- **Constant.** `prod_{i<d} (1 + x^(2^i))` is `sum_{k<2^d} x^k`: the low `2^d` bits, with no reduction and never zero. It equals `prod (1 + r_i)`, the all-zero weight. At d = 8 `new` returns `TooManyPinned` before the shift could underflow.
- **Ceiling.** At 128, `x^128 = x^7 + x^2 + x + 1` falls back into the span. So the length limit and the independence limit coincide, as the description says.
- **Mutations.** Each of these is killed: dropping the spill fold, dropping the high half, shifting the high half one place either way, skipping the inverse, squaring the constant, and reversing the bit order in `fold_bits`. By reading, the existing tests also catch changes to the length-check boundaries, `new`'s boundary, the ratio and order in `challenges()`, and dropping the scale in `fold` or `fold_bits`.

**Also** (inline): by the description's own Zen 5 table, the existing deferred dot beats the shift path when the hardware multiply is on, and choosing per target needs no new API. The rest are nits: a pointer to the soundness obligation on the folds that take whole field elements, the byte layout in the `fold_bits` docs, and the mask test.

binary-field/src/ghash.rs
tcoratger: Done — it chooses per target now.

`dot_powers_of_x` reads `clmul::HAS_HARDWARE_CLMUL` and, where a carryless multiply exists, feeds `poly_dot_128` the pairs `(v, power)` with the power carried incrementally rather than shifted by the index. No new API, as you said. Otherwise it keeps the shift loop.

Re-measured, `-C target-cpu=native`, N = 128:

| | before (multiply) | after | ref: deferred against a materialized table |
|---|---|---|---|
| hardware clmul | 271.6 ns | **159.5 ns** | 136.9 ns |
| portable | 2.945 µs | **171.8 ns** | 2.751 µs |

So the hardware leg went from 1.38x to **1.70x** over the multiplying fold, and the portable leg is unchanged at **17x**.

A residual 1.17x to the table-based path remains, and it is not closable here: that path reads 128 stored field elements, and the whole point of this weight type is that it stores two words instead of a table. Generating the powers is what costs the difference. I have noted that in the doc rather than leaving it implicit, and the doc now says which target each route wins on instead of claiming the shift is always the inner-loop shape.

I have no NEON numbers either — this machine is x86-64 with no emulator, so I have not claimed any.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: Agreed, and your probe is the right one — at d = 1, `[x, 1]` folding to zero is exactly the kernel being visible.

All three whole-cell folds now carry:

> The weights span the field, so on cells that are not `0` or `1` this map has a large kernel.
>
> A pinned zerocheck may only fold prime-field-valued cells, where the bit form holds that by construction.

Stated on each of the three rather than once, since each is reachable on its own.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: Added, in the bit fold's docs:

> Cell `k` at bit `k` is the order a committed bit witness already has on the wire.
>
> An eight-cell block is therefore the byte itself, and a full block is its bytes read little-endian.
>
> Reading them the other way round would pair every cell with the wrong power.

Thank you for checking it end to end — the big-endian spelling is a silent wrong answer rather than a failure, which is the worst kind.

sumcheck/src/univariate_skip/pinned.rs
tcoratger: You are right that the old test was passed by a bounds check narrower than the one written. Split into three, each naming what it would catch:

- the bit just above the block, as before;
- bit 127 at d = 3, which a check looking only at the first bit out would pass;
- bit 64 at d = 6, which is the first bit out *and* exactly where a low-half mask stops looking.

507 tests in the crate pair, green on both legs.

PR #2160, opened 2026-09-16
perf(binary-dft)!: scale by the twiddle's own subfield

tcoratger:
All measurements below are on an AMD Ryzen 9 9950X3D (Zen 5), 32 threads, which has `vpclmulqdq`, `avx512f`, `avx512bw` and `gfni`.

### What the measurement found

Not what I expected. Single-threaded butterfly kernel, both sides L1-resident, `-C target-cpu=native`:

| kernel | Gelem/s |
| --- | --- |
| `Ghash128`, generic `PackedGhash128` multiply | 4.15 |
| `Ghash128`, broadcast kernel with the Karatsuba split hoisted (`poly_slice`) | 4.06 – 4.24 |
| `BinaryField32` | 0.252 |
| `BinaryField64` | 0.234 |
| `BinaryField128` | 0.128 |

The GHASH butterfly is at the hardware carryless-multiply limit. 4.15 Gelem/s at ~5 GHz is 1.25 cycles per element with four elements per register, so one `VPCLMULQDQ` retires per cycle, and Karatsuba plus two reduction folds is 5 of them per pair of elements. The broadcast-scalar kernel that would hoist the multiplier's split out of the loop already exists in `poly_slice.rs` and measures the same, because it is also 5. Lazy reduction has nothing to accumulate: a butterfly holds exactly one product. Fusing two stages into a radix-4 step needs 32 carryless multiplies where two radix-2 stages need 20.

The gap is the tower levels, 16-32x slower per element, because `Packing` is `Self` at every one of them on every target and a tower product is log/exp table lookups that no vector unit widens. The tiled schedule from #2156 could not help here: that path was never memory bound. At `2^20` the whole `ntt/32` matrix is 67 MB and this part has 128 MB of L3.

### The structure that makes it fixable

At stage `j` over the subspace itself the twiddle is `domain_point(2b)` with `b < 2^(l-1-j)`, so it lies in the tower subfield of `2^ceil(log2(l-j))` bits. At `l = 20` that is `GF(2^8)` for stages 12-19, `GF(2^16)` for 4-11 and `GF(2^32)` for 0-3. **Every stage of an unshifted transform of height up to `2^32` has a twiddle four bytes wide or narrower**, and the generic transform was using a full-width product for all of them.

The tower is byte-aligned above `GF(2^8)`, so scaling by an element of a subfield of `D` bytes acts on each group of `D` bytes independently and identically. That is a `D x D` matrix of `8 x 8` blocks over `GF(2)`, which is exactly what `vgf2p8affineqb` computes: one block against 64 bytes in one instruction, with a write mask choosing which byte positions take it.

### Design

One kernel covers `D` in `{1, 2, 4}` with `D` byte rotations and `D^2` masked affine maps. Blocks are `GF(2)`-linear in the multiplier, so the tables split by multiplier byte: 2 KiB for one byte, 16 KiB for two. The four-byte blocks are assembled from four two-byte ones through `GF(2^32) = GF(2^16)[X]/(X^2 + aX + 1)`, so setup is about eight loads and needs no third table. No transpose and no bit-slicing.

Four new files, all in `p3-binary-dft`:

- `src/affine.rs` — the block algebra, the tables, and the group map
- `src/lanes.rs` — a byte-register trait and the AVX-512 backend
- `src/butterfly.rs` — the `ButterflyField` trait, the per-level impls, and the packed kernel moved out of `lch.rs`
- `src/test_util.rs` — a scalar stand-in for a byte register, shared by the kernel tests, following the `test_util` convention already used in `p3-binary-pcs`

The layout obligation the byte view rests on is a private `unsafe` marker trait stated once, with five one-line impls, rather than a safety paragraph repeated per level. Everything else is ordinary generics: one function holds the body and each level forwards to it.

`lch.rs` now calls `F::butterfly::<INVERSE>` and its bound widens from `TowerLevel` to `ButterflyField`. That trait is implemented for all nine sealed `TowerLevel` types, so nothing is excluded; the levels below a byte and `Ghash128`, whose bytes are not subfield coordinates, keep the packed kernel unchanged. Wider twiddles and every twiddle of a shifted transform also keep it. Where the instruction is absent, a one-byte twiddle still takes the typed subfield product rather than a full-width one.

`p3-binary-field` is untouched.

A run shorter than one register returns before any blocks are built. Without that check, a narrow matrix builds a four-byte block set at each of its lowest stages, covers nothing, and runs the packed product anyway — and at width 1 those stages are most of the butterfly calls. The bench now carries a width-1 NTT arm and a short-run butterfly arm so this cannot come back unseen.

### Breaking

The bound on the additive-NTT implementation widens from `TowerLevel` to `ButterflyField`. The same nine sealed types satisfy it, so code naming a concrete level is unaffected. Generic code written as `fn f<F: TowerLevel>(..)` around the transform must widen its own bound to `F: ButterflyField`.

`ButterflyField::butterfly` is deliberately public: it is a usable standalone kernel over two runs, and the bench calls it as one. The trait itself cannot be implemented downstream, because its supertrait is sealed in the field crate.

### Before and after

Forward transform, width 16, 32 threads, `-C target-cpu=native`, alternating runs of the two bench binaries:

| | main | this branch |
| --- | --- | --- |
| `ntt/32/14` | 777 us | 118 us |
| `ntt/32/16` | 3.20 ms | 277 us |
| `ntt/32/18` | 12.0 ms | 778 us |
| `ntt/32/20` | 50.3 ms | 4.03 ms |
| `ntt/64/14` | 1.21 ms | 176 us |
| `ntt/64/16` | 4.69 ms | 346 us |
| `ntt/64/18` | 19.0 ms | 1.70 ms |
| `ntt/64/20` | 85.1 ms | 12.2 ms |
| `ntt/128/tower/14` | 1.91 ms | 272 us |
| `ntt/128/tower/16` | 6.75 ms | 637 us |
| `ntt/128/tower/18` | 29.1 ms | 4.33 ms |
| `ntt/128/tower/20` | 129 ms | 34.0 ms |

Baseline `x86-64` build of the same machine, where there is no `gfni` and only the typed one-byte subfield product applies:

| | main | this branch |
| --- | --- | --- |
| `ntt/32/14` | 1.21 ms | 685 us |
| `ntt/32/16` | 5.24 ms | 3.00 ms |
| `ntt/64/14` | 3.55 ms | 1.61 ms |
| `ntt/64/16` | 14.1 ms | 7.62 ms |
| `ntt/128/tower/14` | 10.1 ms | 4.69 ms |
| `ntt/128/tower/16` | 38.9 ms | 21.1 ms |

The early return, isolated on the short-run butterfly arm, native, three elements per side:

| | without the check | with it |
| --- | --- | --- |
| `butterfly/64/short/t32` | 62.2 ns | 26.6 / 27.4 ns |
| `butterfly/32/short/t32` | 35.7 ns | 29.4 / 28.3 ns |
| `butterfly/32/short/t128` | 35.8 ns | 29.5 / 29.4 ns |

Two figures in the right-hand column are two independent runs of the same binary, so the scatter is visible. The end-to-end narrow arms move the same way but I am not quoting a number for them: a control run of identical code against its own baseline swung between −40% and +29% on this machine, which is wider than the effect.

Butterfly kernel after the change, native, Gelem/s, by the width of the twiddle:

| level | 1 byte | 2 bytes | 4 bytes | wider |
| `BinaryField32` | 58.4 | 44.3 | 16.7 | — |
| `BinaryField64` | 38.2 | 27.0 | 8.76 | 0.301 |
| `BinaryField128` | 10.3 | 7.42 | 4.46 | 0.146 |
| `Ghash128` | 4.05 | 4.13 | 4.13 | 4.11 |

`ntt/128/hybrid` and `ntt/128/ghash` are untouched code and did not move. One A/B run reported +16% and +20% at `2^14`; an immediate repeat of the same pair reported +2% and -1%. The noise floor on this machine reaches 45% at the smallest sizes — the same binary gave 256 us and then 369 us for `ntt/128/ghash/14` on consecutive runs — so I am claiming nothing there. At `2^20`, where the sample is long enough to be stable, both are flat within 3%.

`LchNtt<BinaryField128>` at `2^20` goes from 3.9x slower than the hybrid polynomial-basis route to matching it, 34.0 against 33.1 ms.

### Tests

The const tower product is pinned to the field at 8, 16 and 32 bits. The one-byte block table is swept exhaustively against `BinaryField8` multiplication, all 65536 cases. The split two-byte table and the assembled four-byte one are pinned against building the map column by column, which is what their linearity argument rests on. Every kernel width and both directions are pinned to the element-at-a-time field loop through the scalar model, so a target without the instruction still checks the algebra. The dispatcher is pinned per level over random twiddles and over 0, 1, 0xff, 0x100, 0xffff, 0x1_0000, 0xffff_ffff, 0x1_0000_0000 and 0x87, at every length from 0 to 70 — that range spans an empty run, a single element, several whole 64-byte registers, and every scalar-tail residue at every element size — with a forward-then-inverse round trip at each, plus short-run, empty-run and length-mismatch cases. The existing naive, oracle and twiddle-walk sweeps in `lch.rs` cover the transform end to end.

Run: `cargo build -p p3-binary-dft`, `cargo test -p p3-binary-dft` (86 + 2 passing), `cargo test -p p3-binary-field`, `cargo clippy -p p3-binary-dft --all-targets` (clean), `cargo fmt --check`, `cargo build -p p3-binary-pcs -p p3-sumcheck`. Tests were run both with and without `-C target-cpu=native`, so the byte-map leg and the portable leg are both exercised.

### Left out

- **NEON.** AArch64 has no equivalent of `vgf2p8affineqb`, so it gets only the typed one-byte fallback. Its own piece of work.
- **A 128-bit or 256-bit GFNI path.** The kernel is AVX-512 only. A VEX-encoded version would need blends instead of write masks; one-byte groups would be trivial, two and four more work. Baseline `x86-64` has no `gfni` at all, so this would only serve `x86-64-v3 + gfni` builds.
- **Eight- and sixteen-byte twiddle groups**, which is what a shifted transform and heights above `2^32` would need. Eight bytes is 64 affine maps per 64 bytes, about 2.5x over scalar and marginal; sixteen needs a cross-quadword transpose. The `ntt/128/tower/shifted` arm in the bench makes the untouched shifted case visible.
- **Typed two- and four-byte subfield products in the portable fallback.** Worth roughly 3-10x on non-`gfni` targets for those stages, going by the one-byte column above, and it looks like about twenty lines split across two macros — the obstacle is only that `BinaryField8` has no `Mul<BinaryField16>`. Good follow-up for whoever wants it.
- **The GHASH path**, which the measurement says is already at the instruction's throughput limit.


binary-dft/src/butterfly.rs
```diff
@@ -0,0 +1,549 @@
...
+        (
+            core::slice::from_raw_parts_mut(lo.as_mut_ptr().cast::<u8>(), bytes),
+            core::slice::from_raw_parts_mut(hi.as_mut_ptr().cast::<u8>(), bytes),
+        )
+    };
+
+    // A group is one coordinate of the twiddle's subfield.
+    //
+    // So an element has to hold whole groups for the map to act coordinate by coordinate.
+    let covered = match width {
```
Nashtare, line 204: The blocks are built before `butterfly_run` finds out whether a register fits. At width 1, a `BinaryField32` half is under 64 bytes at stages 0 to 3. Those stages hold 15/16 of the butterfly calls at `2^20`, and most of their twiddles are four bytes wide. So on a GFNI build each of those calls builds `dword_blocks` (eight table entries, sixteen XORs and a `mul_alpha`), covers nothing, and runs the packed product anyway. The same holds for `BinaryField64` at stages 0 to 2 and `BinaryField128` at stages 0 and 1. The bench only measures width 16, where stage 0 already fills a register. Could this return 0 when `bytes < 64` before building anything?

binary-dft/src/lch.rs
```diff
@@ -89,7 +89,7 @@ const STAGED_WORKERS: usize = 16;
 /// A staging tile therefore holds whole runs of rows wherever a row is shorter than this.
 const CACHE_LINE_BYTES: usize = 64;
 
-impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F> {
+impl<F: ButterflyField> AdditiveNtt<F> for LchNtt<F> {
```
Nashtare, line 92: `p3-binary-dft` 0.7.0 is on crates.io with `impl<F: TowerLevel> AdditiveNtt<F> for LchNtt<F>`. Generic downstream code like `fn f<F: TowerLevel>(m) { LchNtt::<F>::default().ntt_batch(m) }` stops compiling under the new bound, even though the same nine types satisfy it. The repo marks that with `!` (#2140, #2145), and `cliff.toml` flags breaking entries from it. Could the title take `!`? Also, `ButterflyField::butterfly` becomes public API, mostly for the bench. Is that intended, or should it be `#[doc(hidden)]`?

Nashtare (approved):
Reviewed at e28e3ad8. I worked the algebra and the instruction encoding by reading. I also ran the byte-map backend under Miri on x86_64 with `+avx512f,+avx512bw,+gfni`, since no CI leg executes it.

**What holds**
- **Dispatch reads the twiddle, not the stage.** `classify` looks at `t.to_repr()`, and the tower embeds `GF(2^(8D))` as the low `D` bytes. So no caller (shifted, inverse, padded) can reach a kernel narrower than its twiddle. The subfield claim holds too: every `v_i` in `CANTOR_BASIS_128` with `i < 2^k` is below `2^(2^k)`. So `domain_point(2b)` with `b < 2^(l-1-j)` fits in `2^ceil(log2(l-j))` bits, which I checked for every `l <= 32`.
- **Group action and `a`.** Scaling `c_0 + c_1 X` by a subfield `s` gives `s c_0 + s c_1 X`, recursively, which is the chunk-wise product `Mul<BinaryField8>` already uses. `dword_blocks` matches the tower's `X^2 = X_3 X + 1`: `out_0 = t_0 v_0 + t_1 v_1` and `out_1 = t_1 v_0 + (t_0 + X_3 t_1) v_1`. `mul_alpha(t1, 16)` is `BinaryField16::mul_alpha`.
- **Encoding.** `map_blocks` puts output bit `i` in matrix byte `7 - i`, as stdarch's reference `mat_vec_multiply_affine` does. `_mm512_ror_epi32::<8>` sends lane bytes `[b0,b1,b2,b3]` to `[b1,b2,b3,b0]`. `i ^ 1` stays inside its 128-bit lane, so `vpshufb` swaps pairs. The masked form takes `affine(source)` where the mask is set and `self` elsewhere. All of these match `Model`. The exhaustive sweep reads the same `BYTE_BLOCKS` the kernel loads, and it checks against the tower `BinaryField8`, not the AES field.
- **Layout and tail.** The five `ByteCoordinates` levels are `repr(transparent)` over integers they fill, so every byte pattern is valid. The view only compiles on x86_64. Loads are `loadu` over exact 64-byte chunks, the halves come from `split_at_mut`, and `64 % size_of::<F>() == 0` is const-asserted. The tail finishes with the full-width `t`.
- **Miri** (no UB): the `__m512i` affine, merge, xor and every rotation agree with `Model` on random registers. `F::butterfly` agrees with the element loop at BF8 through BF128 and Ghash128, over 12 twiddles, one and two registers plus a tail, in both directions. `LchNtt` matches `NaiveAdditiveNtt` at BF64, shifted and not.

**Also**
- The `avx512f+avx512bw+gfni` leg only builds and isn't a required check. So in CI only the `Model` side of these conventions runs.
- The description says every twiddle of a shifted transform keeps the packed kernel. Dispatch reads the value, so a shifted twiddle that happens to be narrow takes the byte map too. That's fine, it just reads differently.

Two minor points inline.

binary-dft/src/butterfly.rs
tcoratger: Confirmed real and fixed. The blocks are no longer built before the run length is known: there is an early return keyed on the run being shorter than one register, ahead of the byte view and any block construction.

Sizing it honestly took two attempts, and the first was wrong. An end-to-end A/B suggested +36.7% on `ntt/32/20` — but a control run of the fixed build against its own identical baseline swung between −40% and +29% on the same arms, so that attribution was unsupported and was discarded. This machine is running other work and the end-to-end arms cannot resolve a 10–20% effect.

The kernel-level arm does resolve. Three elements per side, native:

| | without the check | with it |
|---|---|---|
| `butterfly/64/short/t32` | 62.2 ns | **26.6 / 27.4 ns** |
| `butterfly/32/short/t32` | 35.7 ns | 29.4 / 28.3 ns |
| `butterfly/32/short/t128` | 35.8 ns | 29.5 / 29.4 ns |

**2.3x** on the first, which is `dword_blocks` built and thrown away on a 24-byte run.

Your point that the bench let this through is the more important half, so two arm families were added that the old suite had no way to catch: `ntt/{32,64,128/tower}/narrow` at width 1, and `butterfly/{32,64,128/tower}/short` at three elements per side. Width-16 arm names are unchanged so existing baselines still compare.

binary-dft/src/lch.rs
tcoratger: Both points taken.

**Breaking change marked.** Commit and PR are now `perf(binary-dft)!: scale by the twiddle's own subfield`, with a `BREAKING CHANGE:` paragraph in the commit body and a `### Breaking` section in the description. Both name the fix for a downstream caller: the same nine sealed types satisfy the new bound, so concrete-level code is unaffected, and generic code around the transform widens its bound from `TowerLevel` to `ButterflyField`.

**On the public method** — kept public, deliberately, and now said so in its docs rather than leaving it accidental. It is a usable standalone kernel over two runs, which is why the bench can call it, and `poly_basis::butterfly_forward` is public in this workspace for the same reason. Worth adding: it cannot be implemented downstream in any case, because its supertrait is sealed in the field crate. The trait docs now state that, so the surface is bounded by design rather than by convention.

PR #2161, opened 2026-09-16
feat(binary-field): add bit-sliced GF(2) packings and the square bit transpose

tcoratger:
`Gf2` had `type Packing = Self`, so every bit of a boolean witness occupied a whole byte and every XOR wasted seven eighths of its work. This adds the bit-sliced packings that fix that, and the transpose that makes a bit-sliced layout usable — the packings alone are half a primitive, because a bit slice is a row of a bit matrix and the stack needs both readings of it.

Five commits, one per concern.

Throughout the tables below, **before** is what the crate does today and **after** is what this PR adds. Every number was measured on an AMD Ryzen 9 9950X3D (Zen 5); `native` means `-C target-cpu=native`, and the unqualified columns are a default build.

### The underlier layer

A fixed-width block of bits with no field structure, addressed as an array of machine words: `u8`–`u64` for a single word, `M128`/`M256`/`M512` for the wide ones. Word 0 holds the lowest bits. A `Divisible` relation states that a wide block is exactly several narrow ones end to end, checked at compile time against both sizes and alignments.

The representation is an aligned word array rather than `__m128i`/`__m256i`/`__m512i`. That is the convention the crate's packed GHASH already uses (hold the array, view it as a register where an intrinsic needs one); it keeps the blocks constructible in a constant, derivable for `Eq`/`Hash`, and identical on every target, while the bitwise operations still lower to the widest register the build enables. The register-typed seam belongs with the first intrinsic kernel, which is not in this PR.

No benchmark of its own — it has no operations beyond the ones the packing exercises.

### The packings, 8 through 512 lanes

One `GF(2)` element per bit. Addition is XOR, multiplication is AND, negation and squaring are the identity, `ONE` is every bit set. Lane `i` is bit `i mod B` of word `i / B`, which read as bytes puts lane `i` at bit `i mod 8` of byte `i / 8` — the convention a bit witness already has on the wire. Interleave is a bit permutation following the `PackedField` convention: chunks wider than a word are word copies, narrower ones a shift and a mask, and block length 1 is the even/odd bit deinterleave.

They are deliberately **not** `Gf2::Packing`. `PackedValue` requires a packing to be castable to and from `[Scalar; WIDTH]`; 64 elements are 64 bytes as scalars and 8 bytes bit-sliced, so `unpack_slice` would read eight times past the end of the buffer. `Gf2` keeps its width-one packing and these types carry the algebra themselves (`PrimeCharacteristicRing`, `Algebra<Gf2>`), so generic ring code runs on them unchanged.

Throughput in Gelem/s, 64 Ki elements per iteration. The first row is today's scalar path; the gain columns are that row divided into each width.

| width | add | mul | **gain** | add, native | mul, native | **gain, native** |
|---|---|---|---|---|---|---|
| **before** — scalar, 1 lane | 1.36 | 1.37 | 1x | 1.41 | 1.41 | 1x |
| after — 8 | 10.96 | 12.01 | **8.1x** | 11.66 | 11.71 | **8.3x** |
| after — 16 | 22.72 | 22.71 | **16.7x** | 21.94 | 21.41 | **15.6x** |
| after — 32 | 43.72 | 48.78 | **32.1x** | 47.81 | 47.65 | **33.9x** |
| after — 64 | 96.60 | 96.84 | **71.0x** | 95.04 | 85.77 | **67.4x** |
| after — 128 | 169.1 | 172.2 | **124x** | 165.2 | 166.5 | **117x** |
| after — 256 | 205.8 | 206.1 | **151x** | 330.6 | 328.3 | **234x** |
| after — 512 | 229.8 | 229.4 | **169x** | 419.0 | 421.2 | **297x** |

Below 256 lanes the two builds agree, which is expected: those widths fit an SSE2 register either way.

### Reading a witness at another width

A run of 512-lane packings and a run of 64-lane packings over the same bytes hold the same lanes in the same order, so narrowing is a borrow. `narrow`, `narrow_slice` and `as_bytes` say so, resting on the divisibility check. Ring switching and commit-time packing stop paying for a copy.

Summing popcounts of a one-megabit witness read back at 64 lanes. Without this PR the only way to change width is to copy, so that is the before.

| | **before** — copy | **after** — borrow | **gain** |
|---|---|---|---|
| default | 409 µs (305 MiB/s) | 6.00 µs (20.3 GiB/s) | **68x** |
| native | 320 µs (391 MiB/s) | 0.51 µs (239 GiB/s) | **627x** |

### The square bit transpose

Hacker's Delight figure 7-6's blocked swap, mirrored for a lane order with column 0 at the least significant bit. Rows wider than a word take the same identity a level up: phase 1 transposes every `B x B` block where it lies, phase 2 exchanges block `(i, j)` with `(j, i)` word for word. In place, on the packings (so the type carries the lane order and the row count is checked), square powers of two only.

Two things were measured rather than assumed. Gathering each block into a contiguous buffer left the 512-lane case at 3.4 GiB/s under `native`, because AVX-512 turned the strided reads into 302 gather/scatter instructions; making phase 1 a flat sweep over the words of a whole run of rows emits none and reaches 6.8 GiB/s. And the quadrant side is a const generic because as a runtime value the short passes were vectorised blindly, costing a third of the 64-lane throughput.

GiB/s, 64 matrices per iteration. There is no transpose in the crate today, so the before is the obvious lane-by-lane loop, which is also the reference the fast path is proptested against.

| side | **before** — lane by lane | **after** — blocked | **gain** | before, native | after, native | **gain, native** |
|---|---|---|---|---|---|---|
| 8 | 0.93 | 2.00 | **2.2x** | 1.60 | 1.97 | **1.2x** |
| 64 | 0.311 | 11.10 | **35.7x** | 0.384 | 19.73 | **51.4x** |
| 128 | 0.179 | 11.37 | **63.5x** | 0.168 | 11.49 | **68.4x** |
| 256 | 0.181 | 10.79 | **59.6x** | 0.167 | 7.91 | **47.4x** |
| 512 | 0.143 | 9.44 | **66.0x** | 0.167 | 6.78 | **40.6x** |

At 256 and 512 the `native` build is slower than the default one even with no gathers left; the working sets there are 512 KB and 2 MB, so it is cache-bound rather than issue-bound. Reported as measured, not explained away.

### Tests

Every operation is pinned to a reference built from the lane accessors alone, at all seven widths: boundary patterns (all zeros, all ones, lane 0, lane W−1) and proptests at random. The lane-order contract is asserted lane by lane and byte by byte. Interleave is checked at every block size against a reference written from the two-by-two transpose the trait documents, plus a known-answer test whose expected pairs were worked out by hand. The transpose has a hand-written 8x8 known answer with the before/after drawn in the comments, involution, the identity matrix, a cross-word case, and a test tying it to the packings' lane order so the two cannot silently disagree. 161 tests in the new module, 506 in the crate.

### Not in this PR

The NEON+PMULL packed GHASH is deliberately out. This machine is x86-64, there is no `qemu-aarch64` and no cross linker, so AArch64 here is build-only (`cargo clippy --target aarch64-unknown-linux-gnu`, with and without `+sve2`), and the repository's arm64 CI leg is build-only too. Hand-written NEON intrinsics would ship validated by nothing, which is not a trade worth making in field arithmetic.

### What was run

`cargo build`, `cargo test` (506 pass), `cargo clippy --all-targets -- -D warnings` on the baseline and on `+avx512f,+avx512bw,+gfni,+vpclmulqdq`, `cargo +nightly fmt --all --check`, `cargo clippy --lib --target aarch64-unknown-linux-gnu` with and without `+sve2`, and `cargo build -p p3-binary-dft -p p3-binary-pcs -p p3-sumcheck`.


binary-field/src/bitslice/packing.rs
```diff
@@ -0,0 +1,1012 @@
...
+
+    /// Cut both operands into chunks of `block_len` lanes and interleave the chunks.
+    ///
+    /// The two inputs stack into `2 * WIDTH` lanes, cut into chunks, dealt out alternately:
+    ///
+    /// ```text
+    ///     A = [x0, y0, x1, y1]
+    ///     B = [x2, y2, x3, y3]
+    ///
+    ///     block_len = 1  ->  ([x0, x2, x1, x3], [y0, y2, y1, y3])
```
Nashtare, line 272: This example is `PackedField::interleave`'s, but the code does something else once `block_len < WIDTH / 2`. `interleave_from` and `reference_interleave` both stack `[A_c0, B_c0, A_c1, B_c1, ...]` and split it in half. The example, the "two-by-two matrices" line below and the description's "block length 1 is the even/odd bit deinterleave" all give `[A_c0, B_c0, A_c2, B_c2, ...]` and `[A_c1, B_c1, A_c3, B_c3, ...]`. On these inputs the code would return `([x0, x2, y0, y2], [x1, x3, y1, y3])`. Local probe on `PackedGf2x8` with block 1: lane 1 of `self` lands in lane 2 of the first output, while the example puts it in lane 0 of the second. I compared both conventions on random inputs. The code matches stack-and-deal at every width and block, but matches the `PackedField` layout only for `block_len >= WIDTH / 2` (at width 512 it differs for every block from 1 to 128). The module's reference shares the code's model, so the tests pass either way. The name and signature match the trait method, so callers will expect the trait's layout. Could the code follow the `PackedField` convention, tested against an independent reference? If stack-and-deal is what you want, could the doc and description say that and drop this example?

binary-field/src/bitslice/packing.rs
```diff
@@ -0,0 +1,1012 @@
...
+
+    /// A run of these packings read as bytes, lowest lanes first.
+    ///
+    /// Byte `k` holds lanes `8k .. 8k + 8`, lowest lane at the lowest bit.
+    #[inline]
+    #[must_use]
+    pub const fn as_bytes(slice: &[Self]) -> &[u8] {
+        // A packing is exactly its words, so the byte count is the lane count over eight.
+        const {
+            assert!(size_of::<Self>() * 8 == U::BITS);
```
Nashtare, line 251: Lanes are defined on word values (`get` shifts the word), but this function reads memory order, and so do `narrow`/`narrow_slice` to `u32`/`u16`/`u8` and `Divisible::parts`. On a big-endian target, lane 0 of a `u64` or `M128` word sits in byte 7, and `PackedGf2x64::narrow::<u32>()[0]` holds lanes 32..63. That silently breaks two promises: the doc's "lane `i` at bit `i mod 8` of byte `i / 8`, the order a bit witness already has on the wire", and "narrowing moves no bits". No CI leg is big-endian, so the byte-order tests would never catch it. The crate already guards this kind of reinterpretation in `extension.rs` with `const { assert!(cfg!(target_endian = "little"), ...) }`. Could the byte and sub-word views here carry the same assertion?

binary-field/src/bitslice/underlier.rs
```diff
@@ -0,0 +1,425 @@
...
+///
+/// # Safety
+/// An implementation asserts, at compile time:
+/// - that the wide size is the ratio times the narrow size,
+/// - that the wide alignment is at least the narrow alignment.
+///
+/// Both are what make the reinterpretations below sound.
+pub unsafe trait Divisible<Narrow: Underlier>: Underlier {
+    /// How many narrow blocks fit in one wide block.
+    const RATIO: usize = Self::BITS / Narrow::BITS;
```
Nashtare, line 281: The description says the `Divisible` sizes and alignments are checked at compile time, but only `divisible!` adds those checks: this default has none. `Word`, `Underlier` and `Divisible` are all unsealed and re-exported from the crate root. A related gap: `as_bytes` reads every byte of a run for any `U: Underlier`, and its only guard is `size_of * 8 == U::BITS`, where `BITS` comes from the safe `Word` trait. The `Underlier` safety section never says a word has no padding. So a downstream `#[repr(C)] struct W(u8, u16)` with `Word::BITS = 32` meets both stated bullets, yet makes a safe `as_bytes` call read a padding byte. It takes a downstream `unsafe impl`, so this is minor. Could the three traits be sealed, since nothing outside the crate needs to implement them? If not, could the asserts move into this default, and the `Underlier` contract require the word to be a primitive unsigned integer?

Nashtare (changes requested):
Reviewed at 87ff041e. I read the unsafe casts, the `Divisible` checks, the ring impls, and the interleave and transpose index arithmetic. I also ran a small probe (public API only, native aarch64) against references I wrote on the raw word bits. One point blocks: `interleave` does not compute what its doc example says.

**What holds**
- **Transpose.** Matches an independent reference (output row i, bit j = input row j, bit i, read from the words, not from `get`) at every side from 8 to 512, on dense and sparse random matrices. Wrong row counts (0, 1, 63, 65, 128 at side 64) hit the hard `assert_eq!`. Phase 1 swaps bits [S, 2S) of row t with bits [0, S) of row t+S in every word, coarse to fine. Phase 2's indices stay inside the `split_at_mut(B * j)` halves.
- **Ring impls.** Checked at all seven widths against a lane-by-lane reference: `from_u8..from_u128`, `from_i8/i64/i128` (negatives and `MAX` included), `from_bool`, `from_prime_subfield`, `exp_u64` (0, 1..5, `u64::MAX`), `exp_const_u64`, `exp_power_of_2`, `powers`, `dot_product`, `sum_array` for N = 0..8, `Sum`/`Product` (empty included), `zero_vec` and `andn`. `halve`, `div_2exp_u64` and `/ Gf2::ZERO` panic, as they do on `Gf2`.
- **Unsafe casts.** `parts`, `split_slice`, `wrap`/`unwrap`, `as_bytes` and `words_of` each take their pointer from the borrow they return against. Sizes are exact, and the target alignment is never above the source's. Every length comes from `RATIO` or an inline `const` assert, so a bad `divisible!` pair fails the build as soon as it is used. Nested `narrow_slice_mut` writes land in the right lanes of the wide run.
- Keeping `Gf2::Packing = Self` is right, and nothing in `p3-binary-pcs`, `p3-sumcheck` or `p3-binary-dft` refers to the new types.

**Blocking: `interleave` semantics** (inline). The code is "stack and deal". The doc example, the "two-by-two matrices" sentence and the description's "even/odd bit deinterleave" all describe `PackedField::interleave`. The two agree only when `block_len >= WIDTH / 2`, and the module's reference uses the code's own model, so no test can see the difference.

**Also** (inline)
- No little-endian guard on `as_bytes` or the sub-word `narrow` views, even though the docs promise a byte order.
- `Word`, `Underlier` and `Divisible` are public and unsealed, and the `Underlier` safety contract does not cover what `as_bytes` relies on.
- From reading the tests, no test calls the provided integer conversions (`from_u64`, `from_bool`, `from_prime_subfield`) on the packings. An override that stopped reducing mod 2 would pass. Deleting the `divisible!` asserts would pass too, since every declared pair satisfies them. The defaults are correct today, and a few assertions would pin them.

binary-field/src/bitslice/packing.rs
tcoratger: Fixed, and your diagnosis of *why* it slipped through was the more valuable half.

The implementation stacked the two operands and dealt the stack out; the trait takes chunk `2*(c/2)+take` per output chunk. At width 8, block 1, that put lane 1 of the left operand in lane 2 of the first output where the trait puts it in lane 0 of the second — exactly your probe. It now follows the trait.

The reference was the real defect, since it shared the implementation's model and so could never disagree. It is rewritten from the two-by-two-transpose sentence in the trait's own docs, which makes it structurally unlike the code. Beside it there is now a known-answer test with four hand-computed pairs: `a = 0xd3`, `b = 0x2c` giving `(0x59, 0x69)`, `(0x93, 0x3c)`, `(0xc3, 0x2d)`, `(0xd3, 0x2c)` at blocks 1, 2, 4 and 8. The old convention would have produced `0xa5` at block 1, so the anchor discriminates rather than merely agreeing.

The Morton spread ladder is gone with it — under the correct convention the operation is a shift and a mask, so `spread` and its test were deleted. The description no longer claims a ladder or the even/odd deinterleave.

binary-field/src/bitslice/packing.rs
tcoratger: Added, using the same `const { assert!(cfg!(target_endian = "little"), …) }` guard the crate already uses in `extension.rs`.

It sits on the byte view and on `Divisible`'s reinterpretations — but only where the narrow word is shorter than the wide one. A same-width view is pure reindexing and holds on either endianness, so guarding it would be noise.

You were right that nothing would have caught this: no CI leg is big-endian, and both broken promises are silent wrong answers rather than failures.

binary-field/src/bitslice/underlier.rs
tcoratger: Sealed, which was your first suggestion and the cleaner one. `Word`, `Underlier` and `Divisible` all now go through a crate-private sealing module, matching the pattern already used in `tower.rs`. That closes the `#[repr(C)] struct W(u8, u16)` hole in `as_bytes` outright rather than papering over it in the contract.

The second gap you noted is fixed too: the size and alignment checks lived only in the macro, so the default implementations had none — despite the description claiming otherwise. They now sit in a `CHECK_LAYOUT` const that every default method evaluates, so they cover all impls rather than only the generated ones. The description's claim is true as written now.

PR #2162, opened 2026-09-17
feat(binary-field)!: add the AES field, GF(2^64) and the Frobenius engine

tcoratger:
Three additions to `p3-binary-field`, then one commit closing the review findings.

Every number below was re-measured on the current head with `cargo bench -p p3-binary-field` on a Ryzen 9 9950X3D (Zen 5), criterion medians. "native" means `-C target-cpu=native` (GFNI + AVX-512 + VPCLMULQDQ); "baseline" means the default `x86-64` build, where the same code takes its portable path.

The buffer benchmarks rewrite one buffer in place, so every arm pays the same stores and none pays a reduction another skips. The scalar arms are auto-vectorized by the compiler at whatever width the build allows, so they are not a one-element-at-a-time floor.

## The AES field, and `gf2p8affineqb` as a general engine

`Rijndael8b` is `GF(2^8)` modulo `x^8 + x^4 + x^3 + x + 1` — the modulus the byte-wise Galois-field instructions are defined for, and the one Binius64, Flock and leanVM use. The crate already has a tower `GF(2^8)`; this is a second representation of the same field, and the isomorphism between them is derived at compile time from the tower relations and checked exhaustively over all 65536 operand pairs. The generator is pinned to the image of the tower's, as the 128-bit polynomial-basis field already does.

`ByteMatrix` exposes `gf2p8affineqb` as an arbitrary `F_2`-linear map on bytes, not as a multiply helper. A change of basis, a scaling by a constant, a Frobenius power and any tabulated map are all one matrix and one instruction per register. `gf2p8affineinvqb` with the identity matrix gives a packed field inverse in the same single instruction, which is where the largest win below comes from.

`PackedRijndael8b<N>` is one generic block type at widths 16, 32 and 64, swept with the widest register the target actually has: 64 bytes under AVX-512, 16 bytes with bare GFNI, one byte otherwise. No macro pastes the impls; the block width is a const parameter and the register width is chosen by a small `unsafe` marker trait with one forwarding impl per register.

Elementwise product and inverse over a 4096-byte buffer, per buffer:

| operation | before (baseline build, portable path) | after (native, 64-byte block) | gain |
| --- | --- | --- | --- |
| product | 1.999 us | 25.54 ns | 78.2x |
| inverse | 277.5 us | 16.25 ns | 17080x |

The same two with `-C target-feature=+gfni` alone, the 128-bit form of the instructions and the leg most parts actually have:

| operation | before (baseline build) | after (`+gfni`, 16-byte register) | gain |
| --- | --- | --- | --- |
| product | 1.999 us | 58.91 ns | 33.9x |
| inverse | 277.5 us | 35.59 ns | 7797x |

Those two tables change the build flags as well as the code, so here is the same comparison inside one native build, against the tower's own `GF(2^8)` — the byte-table field the crate already had, which is what this representation is meant to replace at volume:

| operation | before (tower `GF(2^8)`, byte table) | after (`Rijndael8b`, 64-byte block) | gain |
| --- | --- | --- | --- |
| product | 1.559 us | 25.54 ns | 61.1x |
| inverse | 770.0 ns | 16.25 ns | 47.4x |

Scalar `Rijndael8b` arithmetic is table-free and branchless, so one scalar inverse (an addition chain of eleven products) stays slower than the tower's table lookup. The point of this representation is the packed path, and scalar code that needs a byte field at low volume should keep using the tower.

## `GF(2^64)` and a cubic extension

`Poly64` is `GF(2)[x] / (x^64 + x^4 + x^3 + x + 1)`: one carryless product per multiplication, against four plus a Karatsuba recombination at 128 bits, and the product and both fold steps stay inside the vector register where the instruction exists. Inversion is an Itoh-Tsujii chain of eight products and sixty-three squarings with no table at all, so unlike the 128-bit field's inverse it runs one fixed schedule and costs no 320 KiB of Frobenius maps.

`Poly192` is the degree-three extension `y^3 + y + 1` over it. The modulus is irreducible because a root would generate `GF(2^3)` inside `GF(2^64)` and 3 does not divide 64. Squaring turns out to be triangular on the coordinates, which gives both the square and the square root in three coefficient-field operations, and the Frobenius over the coefficient field is one addition and a permutation because `y` maps to `y^2`. Inversion goes through the adjugate of the multiplication matrix: ten products and one coefficient-field inverse.

Neither field existed before, so the baseline is `Ghash128`, the 128-bit polynomial-basis field the crate already has, at the same operation. Dependent chains of 1000 operations, native build, per operation:

| operation | before (`Ghash128`) | after | gain |
| --- | --- | --- | --- |
| multiply | 3.459 ns | `Poly64` 3.104 ns | 1.11x, at half the committed width |
| multiply | 3.459 ns | `Poly192` 5.544 ns | 0.62x, at 1.5x the width |
| inverse | 84.94 ns | `Poly64` 213.3 ns | 0.40x, with no tables and one fixed schedule |
| inverse | 84.94 ns | `Poly192` 235.9 ns | 0.36x |

Against the tower's `GF(2^64)` in the same build, multiplication goes 19.36 ns to 3.104 ns, a 6.2x gain.

Two choices in this commit are not forced by the algebra, so both were measured on their own. Reducing inside the vector register rather than across the register files:

| operation | before (product and fold assembled in integer registers) | after (both in one vector register) | gain |
| --- | --- | --- | --- |
| `Poly64` multiply | 8.450 ns | 3.104 ns | 2.72x |

And deferring the cubic reduction, which is worth twelve carryless multiplies instead of eighteen — but only if the recombination also stays in the vector file:

| variant | time | against the immediate fold |
| --- | --- | --- |
| immediate fold, six reduced products | 6.701 ns | — |
| deferred, recombined in vector registers | 5.544 ns | 1.21x faster |
| deferred, recombined in integer registers | 10.41 ns | 1.55x slower |

The last row is why the deferral is done in the vector file and nowhere else: the six products it saves cost less than the register-file traffic it would add. Both routes are compiled and tested everywhere; the composed one is also benchmarked, so the choice can be rechecked on other hosts.

## Frobenius and linearized polynomials

Squaring is additive in characteristic 2, so every power of it is an `F_2`-linear map, and so is any sum `c_0 x + c_1 x^2 + c_2 x^4 + ...`. Both therefore collapse to one matrix on the engine from the first commit. The eight maps of the orbit are composed at compile time from the squaring matrix, and both the scalar and the packed field route squaring, repeated squaring and the square root through them.

There are two separate effects here, so the table separates them: tabulating the map at all, and then sweeping it with one instruction per register. 4096 elements, native build, per buffer:

| operation | before | after tabulating (scalar) | after sweeping (packed) | tabulation | vectorization | combined |
| --- | --- | --- | --- | --- | --- | --- |
| Frobenius power 3 | 1.950 us (products) | 389.2 ns | 15.99 ns | 5.01x | 24.3x | 122x |
| eight-term linearized weight | 42.07 us (term by term) | 388.9 ns | 15.96 ns | 108x | 24.4x | 2635x |

## What I left out, and why

- **No hand-written NEON.** This machine is x86-64 with no `qemu-aarch64` and no cross linker, and the repo's arm64 leg is build-only, so NEON intrinsics shipped here would never have been executed before a user ran them. AArch64 takes the portable byte path for the AES field, and the identity-map skip in the last commit is what makes that path worth taking: a 16-lane inverse there goes from 1437 to 1266 instructions with the parity fold gone entirely, fully vectorized and with no calls. It does get hardware carryless multiplication for `GF(2^64)`, because that reuses the crate's existing dispatched product rather than new intrinsics. `aarch64-unknown-linux-gnu` and `wasm32-unknown-unknown` were both compiled.
- **No 256-bit GFNI leg.** Two register widths are compiled, 512-bit and 128-bit, and both are executed here and in CI. The VEX 256-bit forms need only `gfni,avx`, so a leg for them is possible; it is left out because no test leg would run it.
- **No recovery of linearized coefficients from an arbitrary matrix.** The correspondence is a bijection, so it is well defined, but inverting it needs an eight-by-eight solve over the field and nothing here consumes it.
- **`Poly192` has no SIMD packing.** Three 64-bit limbs do not map onto a lane layout that pays for itself; the coefficient field's own product is already one instruction.
- **No shared home yet for the byte-matrix row layout.** The same layout is spelled out by hand in `clmul/basis.rs` and in `p3-binary-dft`. `ByteMatrix` is public and now has both directions of the quadword conversion, so a follow-up can route both through it; doing it here would widen this PR across two more crates.
- One test-only change outside the crate: `p3-field-testing`'s packed inverse test sampled a random vector and inverted every lane, which over a 256-element field hits zero often enough to fail. Zero lanes are now pinned to one before inverting.

## Verification

The seven lint gates (`sort`, `toml`, `deps`, `clippy`, `docs`, `fmt`, `scripts`) all pass, and `p3-binary-dft`, `p3-binary-pcs` and `p3-sumcheck` both build and pass their own tests. Tests were run four ways — default, `-C target-cpu=native`, `-C target-feature=+gfni`, and `-C target-feature=+avx512f,+avx512bw,+gfni` — so the portable path, the 128-bit instruction path and the 512-bit one are each exercised. Every commit builds, tests and lints on its own.

CI gains one build-only leg for `+gfni` alone, since the 128-bit form of these instructions is compiled by no other leg, and both GFNI legs now run the packed-versus-scalar differential rather than only compiling it: natively when the runner reports the feature, and under Miri's shims for `vgf2p8mulb`, `vgf2p8affineqb` and `vgf2p8affineinvqb` on every host. That costs about 4 seconds at 128 bits and 19 at 512.


binary-field/src/aes/packed.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+impl<const N: usize> Mul<Rijndael8b> for PackedRijndael8b<N> {
+    type Output = Self;
+
+    /// Scaling is `F_2`-linear, so a fixed factor rides the map engine rather than a broadcast.
+    #[inline]
+    fn mul(mut self, rhs: Rijndael8b) -> Self {
+        self.apply(rhs.scaling_matrix());
```
Nashtare: Across a crate boundary this is the slow route, not the fast one. With `+gfni,+avx512f,+avx512bw`, `p * s` compiles to `vzeroupper; call scaling_matrix; call apply_slice`. That is 8 scalar products plus the 64-iteration `from_images` loop (114 insns), then a runtime-length sweep (153 insns). `p *= s` on the same type goes through `MulAssign` and compiles to `vpbroadcastb; vgf2p8mulb`, with no call. Could this be `self * Self::splat(rhs)`? Callers who hoist a fixed factor out of a loop can still use `scaling_matrix()` + `apply()`.

binary-field/src/aes/engine.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    /// The row layout the affine-byte instruction takes as its matrix operand.
+    #[must_use]
+    pub const fn to_quadword(self) -> u64 {
+        self.0
+    }
+
+    /// Replaces every byte of the slice with its image.
+    pub fn apply_slice(self, bytes: &mut [u8]) {
```
Nashtare, line 134: `apply_slice` is a non-generic `pub fn` without `#[inline]`, so `PackedRijndael8b::apply` / `frobenius` in a downstream crate become an out-of-line call with a runtime length, not a sweep specialized to `N`. Would `#[inline]` here, or having `PackedRijndael8b::apply` call `map_slice(&Affine(map), ..)` directly, keep the packed Frobenius at one instruction per register for consumers?

binary-field/src/aes/packed.rs
```diff
@@ -0,0 +1,692 @@
...
+
+// SAFETY: the transparent array satisfies the packed layout contract.
+//
+// Every operation acts on each byte independently.
+unsafe impl<const N: usize> PackedField for PackedRijndael8b<N> {
+    type Scalar = Rijndael8b;
+}
+
+// SAFETY: the widths this type is used at are powers of two.
+unsafe impl<const N: usize> PackedFieldPow2 for PackedRijndael8b<N> {
```
Nashtare, line 463: This `SAFETY` comment describes how the type is used rather than something the code checks. `PackedFieldPow2` promises `WIDTH` is a power of two, but only `splat` asserts it. `PackedRijndael8b<24>` builds via `from_fn`, satisfies a `P: PackedFieldPow2` bound, and `interleave(other, 8)` passes the guard and then panics with `index out of bounds: the len is 24 but the index is 24` at line 473. Could the invariant live in the impl, e.g. `const WIDTH: usize = { assert!(N.is_power_of_two()); N };`?

.github/workflows/ci.yml
```diff
@@ -99,6 +99,14 @@ jobs:
             sha_ni: true
             target: x86_64-unknown-linux-gnu
             modes: serial parallel
+          # The byte-wise Galois-field instructions have a 128-bit form that needs no wide
+          # register, and that form is the one no other leg compiles.
+          - name: linux, gfni, build only
```
Nashtare, line 104: Both GFNI legs are build-only, so CI never executes a GFNI instruction, and neither leg is in `main`'s required checks. Two cheap ways to run them:

1. Copy the `sha_ni` step: `if grep -qw gfni /proc/cpuinfo; then python3 scripts/check.py test --package p3-binary-field; fi` (plus `avx512bw` on the 512 leg). Part of the runner pool has GFNI, so the proptests pinning the sweeps to the scalar routines would run at least some of the time.
2. Miri has shims for `vgf2p8affineqb`, `vgf2p8affineinvqb` and `vgf2p8mulb` at 128 and 512 bits. A lane-by-lane differential of `PackedRijndael8b` under `cargo +nightly miri run --target x86_64-unknown-linux-gnu` passes with `+gfni` and with `+gfni,+avx512f,+avx512bw`.

binary-field/src/aes/engine.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    #[inline(always)]
+    fn affine(self, matrix: Self::Matrix) -> Self {
+        ByteMatrix(matrix).apply(self)
+    }
+
+    #[inline(always)]
+    fn inverse_then_affine(self, matrix: Self::Matrix) -> Self {
+        ByteMatrix(matrix).apply(invert_byte(self))
```
Nashtare: On the portable path `Invert` passes `IDENTITY`, and LLVM does not fold `apply(IDENTITY, _)`. On aarch64, a downstream `invert_or_zero` on `PackedRijndael8b<16>` is 16 `bl invert_byte` calls, each followed by the full parity fold (two `mul`s, shift-xors, `rbit`). Since AArch64 is meant to take this path, could the byte backend skip the map when the matrix is the identity, e.g. a separate `invert` method on `ByteLanes`, and could `invert_byte` get `#[inline]`?

binary-field/src/aes/engine.rs
```diff
@@ -0,0 +1,469 @@
...
+/// The byte-lane operations the kernels below are written against.
+///
+/// Every method acts on each byte of the register independently.
+///
+/// # Safety
+///
+/// An implementation must occupy exactly its own width in contiguous bytes.
+///
+/// Every bit pattern of that many bytes must be a valid value of it.
+unsafe trait ByteLanes: Copy {
```
Nashtare, line 203: The contract does not state what `map_registers` actually needs: that `load`/`store` access exactly `WIDTH` bytes. The 'every bit pattern is valid' clause is never relied on. Could the doc say that directly, with `const { assert!(size_of::<L>() == L::WIDTH) }` in `map_registers`/`mul_registers`?

binary-field/src/aes/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    #[test]
+    fn the_published_substitution_box_entries_come_out() {
+        // The cipher's substitution box is inversion followed by a fixed affine map.
+        //
+        // Its published entries therefore pin both the inverse and the map engine.
+        //
+        //     s(x) = A(x^254) + 0x63,   A the map with the circulant rows of 0x1f
+        let affine = ByteMatrix::from_images([0x1f, 0x3e, 0x7c, 0xf8, 0xf1, 0xe3, 0xc7, 0x8f]);
```
Nashtare, line 571: The commit message says this anchors the row order of the instruction's matrix operand, but the test only goes through the portable `apply`. This matrix's quadword is `0xF1E3C78F1F3E7CF8`, the published constant GFNI AES S-boxes pass to `gf2p8affineinvqb`. Could one `assert_eq!(affine.to_quadword(), 0xF1E3_C78F_1F3E_7CF8)` tie `to_quadword()` to the hardware layout on every CI host?

binary-field/src/aes/engine/x86_64.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[inline(always)]
+    fn inverse_then_affine(self, matrix: Self::Matrix) -> Self {
+        // SAFETY: as above.
+        unsafe { _mm_gf2p8affineinv_epi64_epi8::<0>(self, matrix) }
+    }
+}
+
+/// The widest register the kernels sweep with, holding sixteen bytes.
+///
+/// Nothing wider is available without the half-kilobit registers.
```
Nashtare: This is not quite right: the VEX 256-bit forms (`_mm256_gf2p8*`) need only `gfni,avx`, so AVX2+GFNI parts without AVX-512 could sweep 32 bytes. The PR description gives the real reason (untested surface). Could the comment say that?

binary-field/src/aes/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+/// Zero is outside that group, and the same power sends it to itself.
+///
+/// That is also the convention the byte-wise hardware inverse follows.
+///
+/// # Algorithm
+///
+/// An addition chain on the exponent, seven squarings and four products:
+///
+/// ```text
+///     2, 3, 12, 15, 240, 252, 254
```
Nashtare: `2, 3, 12, 15, 240, 252, 254` is not an addition chain (12 is not a sum of two earlier terms). The code computes `1, 2, 3, 6, 12, 15, 30, 60, 120, 240, 252, 254`. Could the full chain be listed so it can be checked against the eleven lines below?

binary-field/src/aes/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+
+/// The binary field `GF(2^8)` modulo `x^8 + x^4 + x^3 + x + 1`.
+///
+/// Bit `i` stores the coefficient of `x^i`, and every byte is a distinct field element.
+///
+/// This is the modulus the byte-wise field instructions are defined for.
+///
+/// That is what makes it worth carrying beside the tower's own byte field.
+///
+/// Arithmetic here is constant time: no product or inverse indexes a table by an operand.
```
Nashtare: 'Constant time' is stronger than the code guarantees. `try_inverse` (and so `inverse` and `/`) branches on zero, and `scaling_matrix` goes through `from_images`, which branches on the factor's bits in source. The clause after the colon is accurate. Could the sentence keep only that clause, or say 'table-free' rather than 'constant time'?

binary-field/src/aes/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+    fn div_2exp_u64(&self, _exp: u64) -> Self {
+        panic!("div_2exp_u64 is undefined in characteristic 2")
+    }
+}
+
+impl Field for Rijndael8b {
+    type Packing = PackedRijndael8b<{ packed::PACKING_WIDTH }>;
+
+    // The multiplicative group has order 255, and `x + 1` has exactly that order here.
+    const GENERATOR: Self = Self(0x03);
```
Nashtare: `0x03` is primitive, but it is not the image of `BinaryField8::GENERATOR`: tower `0x13` maps to `0x88`. `Ghash128` pins its generator to the tower image and tests it (`the_generator_is_the_image_of_the_tower_generator`). Is the difference intentional, or should the two representations agree here as they do for the Cantor basis?

binary-field/src/poly64.rs
```diff
@@ -0,0 +1,455 @@
...
+impl Field for Poly64 {
+    type Packing = Self;
+
+    // The polynomial variable itself has order `2^64 - 1` under this modulus.
+    const GENERATOR: Self = Self(2);
+
+    #[inline]
+    fn try_inverse(&self) -> Option<Self> {
+        // Zero has no multiplicative inverse.
+        (self.0 != 0).then(|| Self(clmul::poly_inverse_64(self.0)))
```
Nashtare: The struct doc (line 47) says inversion is constant time. But `bool::then` only runs the chain for a nonzero operand: zero returns after one compare, and every other operand pays 8 products and 63 squarings. That leaks whether a secret operand is zero, through `inverse`, `Div` and batch inversion as well.

`poly_inverse_64` already sends 0 to 0, so one fix is to run the chain unconditionally and build the `Option` afterwards:

```rust
let inverse = Self(clmul::poly_inverse_64(self.0));
(self.0 != 0).then_some(inverse)
```

The other is to qualify the doc as constant time on nonzero inputs.

binary-field/src/clmul/gf64.rs
```diff
@@ -0,0 +1,245 @@
...
+/// The inverse of a nonzero element, sending zero to zero.
+///
+/// Itoh-Tsujii on the exponent `2^64 - 2`, writing `b_k` for `x^(2^k - 1)`:
+///
+/// ```text
+///     b_(a+b)  = b_a^(2^b) * b_b
+///     chain      1, 2, 3, 6, 12, 24, 48, 60, 63
+/// ```
+///
+/// Nine products and sixty-three squarings, none of them indexed by the operand.
```
Nashtare: nit: the chain `1, 2, 3, 6, 12, 24, 48, 60, 63` has nine entries but eight steps, and the body has eight `poly_mul_64` calls (`b2, b3, b6, b12, b24, b48, b60, b63`). Replayed on exponents in Sage: 8 products, 63 squarings, final exponent `2^64 - 2`. The commit message and PR description have the same "nine".

binary-field/src/poly192.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// It fixes every coefficient and permutes the basis.
+    ///
+    /// The three roots of the modulus are `y`, `y^2` and `y^2 + y`, so:
+    ///
+    /// ```text
+    ///     y^(2^64)                 =  y^2
+    ///     a_0 + a_1 y + a_2 y^2    ->  a_0 + a_2 y + (a_1 + a_2) y^2
+    /// ```
+    ///
+    /// So one application is three additions and no products at all.
```
Nashtare: nit: `[a0, a2, a1 + a2]` is one addition plus a permutation of coordinates, not three additions. The map itself checks out: `y^(2^64) = y^2` in Sage. The commit message and PR description say "three additions" too.

binary-field/src/poly192.rs
```diff
@@ -0,0 +1,715 @@
...
+    #[inline]
+    fn div_2exp_u64(&self, _exp: u64) -> Self {
+        panic!("div_2exp_u64 is undefined in characteristic 2")
+    }
+}
+
+impl Field for Poly192 {
+    type Packing = Self;
+
+    // The smallest element of the multiplicative group whose order is the whole group.
```
Nashtare: nit: `5 + 3y + y^2` is a generator (checked in Sage against all 15 prime factors of `2^192 - 1`), but it is not the smallest.

- Under the derived `Ord`, lexicographic on `[a0, a1, a2]`, `[0, 1, 9]` is a generator.
- Reading the coordinates as an integer with `a2` highest, `[8, 1, 0]` is.

I'd drop "smallest".

binary-field/src/poly192.rs
```diff
@@ -0,0 +1,715 @@
...
+        let d01 = (a0 + a1) * (b0 + b1);
+        let d02 = (a0 + a2) * (b0 + b2);
+        let d12 = (a1 + a2) * (b1 + b2);
+
+        // The five coefficients of the unreduced product.
+        let p1 = d01 + c0 + c1;
+        let p2 = d02 + c0 + c1 + c2;
+        let p3 = d12 + c1 + c2;
+
+        Self([c0 + p3, p1 + p3 + c2, p2 + c2])
```
Nashtare: Not blocking, and not benchmarked: each of the six products above pays its own fold. On the PCLMULQDQ path that is two of the three carryless multiplies inside `poly_mul_64`, so this product is 18 multiplies, and the fold dominates rather than the product count.

Reduction is `F_2`-linear. The recombination (`p1`, `p2`, `p3` and the `y^3`/`y^4` rewrite) can therefore run on the unreduced 128-bit products, folding only the three output coordinates: 6 + 3x2 = 12 multiplies.

For reference, p3-field's `trinomial_cubic_mul` uses the same modulus (`X^3 - X - 1` is `y^3 + y + 1` in characteristic 2). Through `Poly64::dot_product` it already defers the fold: 9 + 3x2 = 15.

binary-field/src/poly64.rs
```diff
@@ -0,0 +1,455 @@
...
+    fn div_2exp_u64(&self, _exp: u64) -> Self {
+        panic!("div_2exp_u64 is undefined in characteristic 2")
+    }
+}
+
+impl Field for Poly64 {
+    type Packing = Self;
+
+    // The polynomial variable itself has order `2^64 - 1` under this modulus.
+    const GENERATOR: Self = Self(2);
```
Nashtare: nit: `Ghash128` pins its generator to the image of the tower's, with a test (`the_generator_is_the_image_of_the_tower_generator`), so the change of basis carries one generator onto the other. Here `Poly64::from(BinaryField64::GENERATOR)` is `0xc313d8d1bfae2150`, not `x`.

`x` is a valid choice (the modulus is primitive), and nothing depends on the convention today. But the two polynomial-basis fields now disagree. Either pin `GENERATOR` the same way, or say why this one differs.

binary-field/benches/arithmetic.rs
```diff
@@ -853,6 +853,256 @@ fn bench_batch_kernels(c: &mut Criterion) {
...
+    let tabulated = weight.to_matrix();
+
+    {
+        let mut group = c.benchmark_group("frobenius/twist");
+        group.throughput(criterion::Throughput::Elements(BYTES as u64));
+        group.bench_function("repeated-square", |b| {
+            b.iter(|| {
+                black_box(&scalars)
+                    .iter()
+                    .map(|&x| (0..POWER).fold(x, |acc, _| acc.square()))
```
Nashtare: This commit also reroutes `Rijndael8b::square` to `frobenius_map(1).apply(..)` (`aes/mod.rs:301`), so this "before" arm runs the tabulated map three times, not three field products. The description's `1.835 us (repeated squaring)` row therefore compares one tabulated map against three, and the packed row reuses it as its baseline. Could this be `acc * acc`, or the arm relabelled? The scalar arms also pay a `.sum()` that the packed arms don't, and `frobenius/linearized` has no scalar `to_matrix().apply(x)` arm, so the 345x mixes tabulation with SIMD.

binary-field/src/aes/packed.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+impl<const N: usize> Distribution<PackedRijndael8b<N>> for StandardUniform {
+    #[inline]
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> PackedRijndael8b<N> {
+        PackedRijndael8b(array::from_fn(|_| rng.random()))
+    }
+}
+
+impl<const N: usize> PrimeCharacteristicRing for PackedRijndael8b<N> {
```
Nashtare, line 326: The scalar field now overrides `exp_power_of_2` with the tabulated map, but the packed type keeps the default of `k` full products. Generic `PackedField` code over `Rijndael8b::Packing` then pays `k` `vgf2p8mulb` sweeps (or `k` byte-product loops on the portable path) where one affine sweep does the job:

```rust
#[inline]
fn exp_power_of_2(&self, power_log: usize) -> Self {
    let mut out = *self;
    out.frobenius(power_log);
    out
}
```

binary-field/src/aes/frobenius.rs
```diff
@@ -0,0 +1,256 @@
...
+///
+/// Every `F_2`-linear map on the field has exactly one such form.
+///
+/// Both sides are 64-dimensional over `GF(2)`, and the correspondence is injective.
+///
+/// Held as coefficients rather than as a matrix, because that is the form a protocol states a
+/// Frobenius-twisted weight in.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+#[must_use]
+pub struct LinearizedPoly([Rijndael8b; ORBIT]);
```
Nashtare: This goes to the crate root next to `Poly64`/`Poly192`, where `Poly` means "polynomial-basis field", so `LinearizedPoly` reads like another field type. It only exists over `Rijndael8b`, and the crate has several other binary fields. Something like `LinearizedPoly8b` (or keeping it under an `aes` path) would avoid a breaking rename once a linearized form over another field lands. Minor: no `Hash`, although `ByteMatrix` derives it, and no serde, although the doc says this is the form a protocol states a weight in.

binary-field/src/aes/frobenius.rs
```diff
@@ -0,0 +1,256 @@
...
+            assert_eq!(root.to_byte(), Rijndael8b::frobenius_map(7).apply(byte));
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(512))]
+
+        /// The tabulated map must agree with summing the linearized terms at every point.
+        #[test]
+        fn the_tabulated_linearized_map_matches_its_definition(coefficients: [u8; ORBIT], x: u8) {
```
Nashtare, line 203: `eval` and `to_matrix` are the same formula over the same `FROBENIUS` table, so a shared misreading (`c_j^(2^j)`, or reversed coefficient order) passes here. The hand-checked vector below has coefficients in `GF(2)`, where those readings coincide. I checked 304 vectors against Sage and the code is right, but some anchors that don't share the formula would pin it:

- `LinearizedPoly::new([c, 0, 0, 0, 0, 0, 0, 0]).to_matrix() == c.scaling_matrix()`
- a unit coefficient at `j` gives `Rijndael8b::frobenius_map(j)`
- one fixed KAT with non-binary coefficients

Also, nothing tests the scalar `exp_power_of_2` override past `k = 7` (only via `try_sqrt`). The tower's `frobenius_maps_and_sqrt_match_repeated_reference_squares` covers `0..=bits` and `usize::MAX`, which would be worth mirroring.

binary-field/src/aes/frobenius.rs
```diff
@@ -0,0 +1,256 @@
...
+        k += 1;
+    }
+    maps
+};
+
+impl Rijndael8b {
+    /// The map raising every element to the power `2^k`.
+    ///
+    /// The orbit closes after eight steps, so the argument is taken modulo that.
+    pub const fn frobenius_map(power: usize) -> ByteMatrix {
```
Nashtare: nit: `power` is the log. `frobenius_map(3)` is `x -> x^8`. `PrimeCharacteristicRing::exp_power_of_2`, which this backs, calls the same argument `power_log`. Same for `PackedRijndael8b::frobenius`.

binary-field/src/aes/frobenius.rs
```diff
@@ -0,0 +1,256 @@
...
+            let expected: PackedRijndael8b<64> = PackedValue::from_fn(|i| {
+                Rijndael8b::from_byte(repeated_square(values[i], power % ORBIT))
+            });
+            prop_assert_eq!(twisted, expected);
+        }
+    }
+
+    /// A worked example small enough to check by hand.
+    ///
+    /// The polynomial `x^2 + x` is the trace-like map whose kernel is the prime subfield.
```
Nashtare: nit: `x^2 + x` is the Artin-Schreier map, not a trace-like one. Its image is the trace-zero hyperplane. The kernel statement is right. On line 51, the orbit array also repeats for subfield elements (the orbit of `1` is eight ones), so it is the conjugates with multiplicity rather than the conjugate set.

binary-field/src/aes/packed.rs
```diff
@@ -0,0 +1,692 @@
...
+    pub fn frobenius(&mut self, power: usize) {
+        self.apply(Rijndael8b::frobenius_map(power));
+    }
+
+    /// Replaces every element with its inverse, leaving zero alone.
+    ///
+    /// The byte-wise hardware inverse is one instruction per register.
+    ///
+    /// The fallback is an addition chain of eleven products per element.
+    pub fn invert_or_zero(&mut self) {
```
Nashtare: nit: `PackedGf2::invert_or_zero` (#2161, same crate) takes `self` and returns `Self` (and is `const`), while this one mutates in place. So do `apply` and `frobenius`. It would be good to settle one receiver convention for the crate's packed types before release.

binary-field/src/aes/engine.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// The map sending every byte to zero.
+    pub const ZERO: Self = Self(0);
+
+    /// The map leaving every byte alone.
+    pub const IDENTITY: Self = Self(BIT_LADDER);
+
+    /// The map sending basis vector `1 << b` to entry `b` of the argument.
+    ///
+    /// A linear map is pinned down by where it sends each of the eight basis vectors.
+    pub const fn from_images(images: [u8; 8]) -> Self {
```
Nashtare, line 49: With this, the `vgf2p8affineqb` row layout (byte `7 - i` = row for output bit `i`) is spelled out by hand in three places: here, `clmul/basis.rs` `blocked::affine_blocks` (#2153), and `binary-dft/src/affine.rs` (`blocks[k][j] |= 1 << (8 * (7 - i) + b)`, #2160). binary-dft also has its own `ByteLanes` trait (`lanes.rs`) with the same name and a different contract from the one below. `ByteMatrix` is public now, so a follow-up could route both through `ByteMatrix::from_images(..).to_quadword()`. That needs a `ByteMatrix::from_quadword` for binary-dft's existing `u64` blocks. Not blocking.

Nashtare (approved):
Reviewed at aaa01f4e; the rebase to 5b7dbbf3 only touches `lib.rs`. I checked the algebra in Sage, ran the real GFNI and PCLMULQDQ intrinsics under Miri, and ran native PMULL and portable differentials against Sage vectors. I found no correctness bug.

**What holds**
- `Rijndael8b`: the modulus is irreducible, the product is exact, and the inverse chain is `a^254` with 0 → 0. Sage rebuilt the tower independently, and the isomorphism, both basis-change matrices and multiplicativity match.
- `ByteMatrix`: the quadword layout, `A·x` orientation and composition order match Intel's `gf2p8affineqb`, and `gf2p8affineinvqb` with the identity is the field inverse. The two 0x11B-only instructions are reachable only from `PackedRijndael8b`.
- `PackedRijndael8b` at 16/32/64: every op, `interleave` included, matches scalar under Miri with `+gfni`, `+gfni,+avx512f` (stays 16 wide), full AVX-512 and baseline x86, for every tail length up to 200, with no Stacked Borrows errors.
- `Poly64`: the modulus is primitive, the two-step fold is exact for any 128-bit product, and the inverse chain reaches `2^64 - 2`. Products agree with Sage on 8368 vectors natively (PMULL and portable) and on 1003 under Miri with PCLMULQDQ.
- `Poly192`: Sage confirms the Karatsuba, square, square root and adjugate-inverse formulas, and `y^(2^64) = y^2`.
- Frobenius: all eight maps and 304 random linearized polynomials match Sage, and `square`, `exp_power_of_2` and `try_sqrt` agree with schoolbook arithmetic on all 256 elements, including powers past 8.
- No existing public API or `Ghash128` behavior changes, and the `field-testing` change removes no coverage from other fields.

**Worth addressing**
- From another crate, `packed * scalar` is two out-of-line calls, while `packed *= scalar` is two instructions (inline).
- CI never executes a GFNI instruction: both legs are build-only and neither is required (inline).
- `PackedFieldPow2` is implemented for every `N`, not only powers of two (inline).
- The constant-time docs don't hold at zero for `Poly64` or `Rijndael8b` inversion (inline).

**Description**
- The arm64 build-only leg isn't the only arm64 coverage: `macos, parallel` runs the full suite on arm64, including the new PMULL path, so NEON intrinsics would be executed.
- Hardware carryless multiplication for `GF(2^64)` on AArch64 needs the `aes` target feature, which `aarch64-unknown-linux-gnu` doesn't enable by default, and the folds run in general-purpose registers there.
- `-C target-cpu=x86-64-v4` doesn't include GFNI, so an AVX-512 build at that level takes the one-byte AES path; the 64-byte path needs `+gfni` on top.
- The `Poly64` inverse chain has eight products, not nine, and the `Poly192` Frobenius is one addition plus a swap.
- The baseline of the 4.74x Frobenius row calls `square()`, which the same commit reroutes through the table (inline).

binary-field/src/aes/packed.rs
tcoratger: Fixed: the operator is now `self * Self::splat(rhs)`, so it lowers to the broadcast plus one byte-wise product instead of the eight scalar products and the sixty-four-iteration loop. Callers hoisting a fixed factor out of a loop keep the matrix route.

binary-field/src/aes/engine.rs
tcoratger: `#[inline]` added, so the packed map stays a sweep specialised to the width across a crate boundary rather than an out-of-line call with a runtime length.

binary-field/src/aes/packed.rs
tcoratger: You are right that the comment described usage rather than anything checked. The invariant now lives in `const WIDTH: usize = { assert!(N.is_power_of_two()); N }` and again as a `const {}` block at the indexing site. A width of 24 is now `error[E0080]: evaluation panicked: the block width must be a power of two` at compile time, not an out-of-bounds panic at run time.

.github/workflows/ci.yml
tcoratger: Both suggestions taken, and you were right that it is cheap. The runtime `grep -qw gfni /proc/cpuinfo` guard now copies the `sha_ni` pattern, with an `avx512bw` check on the wide leg. Miri runs the deterministic corner differential on every host regardless of CPU: 4.1 s at 128 bits, 18.7 s at 512. I did not put the whole module under Miri, since the proptests around it take minutes. One note for anyone repeating it: the recipe needs `-Zmiri-disable-isolation`, or proptest's regression file trips `getcwd`.

binary-field/src/aes/engine.rs
tcoratger: The contract now states what the code actually needs, that load and store touch exactly the register width, and drops the clause about bit patterns that nothing relied on. `const { assert!(size_of::<L>() == L::WIDTH) }` added to both entry points.

binary-field/src/aes/engine.rs
tcoratger: Same correction as above: the doc now says what is required rather than what happens to be true, with the size assertion alongside it.

binary-field/src/aes/mod.rs
tcoratger: Taken. `assert_eq!(affine.to_quadword(), 0xf1e3_c78f_1f3e_7cf8)` now ties the row order to the published constant on every host, not only on GFNI ones. That is the cheapest hardware-layout anchor available and it should have been there from the start.

binary-field/src/aes/engine/x86_64.rs
tcoratger: Corrected. The comment now gives the real reason, which is untested surface rather than a missing instruction, since the 256-bit forms need only `gfni` and `avx`.

binary-field/src/aes/mod.rs
tcoratger: You are right that it was not an addition chain. The doc now lists the twelve terms the code actually computes, so it can be read against the lines below it.

binary-field/src/aes/mod.rs
tcoratger: Changed to **table-free**, which is what the code earns. `try_inverse` branches on zero and the scaling matrix branches on the factor's bits, so the stronger claim was false as written.

binary-field/src/aes/mod.rs
tcoratger: Not intentional, and now fixed: the generator is pinned to the image of the tower's, `0x88`. I recomputed that image independently in Python and it matches your Sage. The two representations now agree here as they do for the Cantor basis, and the full-order test stays. This is the breaking part of the change, hence the marker on the title.

binary-field/src/poly64.rs
tcoratger: This was the most important comment in the set and the fix is the one you wrote. All three fields now run the chain unconditionally and build the `Option` afterwards, and each exposes the total map so a caller can skip the `Option`. `Poly192` had the same shape one level up, where a zero norm skipped the three scaling products, and it scales unconditionally now too.

One I flagged rather than fixed: `Ghash128::try_inverse` is identical, and is where the pattern was copied from. It is pre-existing and its doc already discloses variable time from the operand-indexed tables, so fixing it here would have put reviewed code in this diff. Filed as #2172.

binary-field/src/clmul/gf64.rs
tcoratger: Corrected to eight steps, in the doc, the commit message and the description. Thank you for replaying it.

binary-field/src/poly192.rs
tcoratger: Corrected to one addition and a permutation, in all three places.

binary-field/src/poly192.rs
tcoratger: "Smallest" dropped. It is a generator and the code is right; the claim about minimality was not.

binary-field/src/poly192.rs
tcoratger: Measured, and the obvious reading of this is a trap worth recording. Recombining the unreduced products **in integer registers** is 10.41 ns against 6.33 ns, so 1.63x *slower* — the six carryless multiplies it saves cost less than the register-file traffic it adds. Keeping the recombination in the vector file does win: 6.701 to 5.544 ns, 1.21x in a same-run A/B. Both routes stay compiled and tested, and there is a new bench arm so the choice is recheckable on another host.

binary-field/src/poly64.rs
tcoratger: Pinned the same way, to `0xc313d8d1bfae2150`, so all three polynomial-basis fields now share one convention. Having two was the real problem, as you say.

binary-field/benches/arithmetic.rs
tcoratger: You were right that the baseline was wrong, and the correction did not go the way I expected. The old arm was *faster* than an honest product baseline, 1.835 against 1.950 µs, so it understated rather than flattered — the actual defect was measuring one tabulated map against three.

I rebuilt the harness: every arm rewrites one buffer in place, so the scalar arms no longer pay a `.sum()` the packed ones skip, and the per-iteration clone is gone. That clone had a ~85 ns floor which was silently capping every packed row. Both groups gained the missing scalar-tabulated arm, so the figure is now decomposed into tabulation and vectorisation instead of mixing them: for the eight-term weight, 108x from tabulation and 24.4x from vectorisation.

binary-field/src/aes/packed.rs
tcoratger: Overridden with one affine sweep, so generic code over the packing no longer pays `k` products where one map does it.

binary-field/src/aes/frobenius.rs
tcoratger: Renamed to `LinearizedPoly8b`, with `Hash` and serde added. You are right that the crate root reads as a field namespace and that a second linearized form would have forced the rename later.

binary-field/src/aes/frobenius.rs
tcoratger: All three anchors added, since you are right that the two paths share one formula and so share any misreading: the constant coefficient against the scaling matrix, a unit coefficient at `j` against the Frobenius map, and a known answer with four non-binary coefficients whose values I computed outside the crate. The tower's coverage is mirrored too, over `0..=2*bits+3` and `usize::MAX`, plus a serde round trip.

binary-field/src/aes/frobenius.rs
tcoratger: Renamed to `power_log`, matching the trait method it backs. Same for the packed entry point.

binary-field/src/aes/frobenius.rs
tcoratger: Both corrected: Artin–Schreier rather than trace-like, and the orbit array described as conjugates with multiplicity.

binary-field/src/aes/packed.rs
tcoratger: Settled on the convention the sibling packing established: `apply`, `frobenius` and `invert_or_zero` all take `self` and return `Self`. Better to fix it before release than to carry two conventions.

binary-field/src/aes/engine.rs
tcoratger: `ByteMatrix::from_quadword` added, so this is unblocked. I stopped short of routing the two existing sites through it, since that would widen this PR across two more crates for no behaviour change; it is listed as left out in the description.

PR #2164, opened 2026-09-17
feat(examples): prove Keccak-f and BLAKE3 over GF(2^128) with the binary PCS

Nashtare:
## Summary

Adds characteristic-2 AIRs for Keccak-f[1600] and the BLAKE3 compression function, and two examples that prove them with the multilinear STARK (`p3-multi-stark`) over `BinaryField128` using `BinaryPcs`. The goal is a like-for-like hash workload to compare against `prove_prime_field_31 -o keccak-f-permutations|blake-3-permutations` and to profile the binary prover on.

The existing `KeccakAir` and `Blake3Air` recompose 16-bit limbs with powers of two, which vanish in characteristic 2, so they cannot be reused over a binary field. The new AIRs work on bits: XOR is field addition and AND is field multiplication.

### `KeccakBinaryAir` (`keccak-air/src/binary`)

- One row per round, 25 rows per permutation: rows 0..24 hold the input state of round `r`, row 24 holds the output. 1625 columns (25 one-hot round flags + 1600 state bits).
- Theta/rho/pi are linear expressions; chi is the only nonlinearity, so every constraint has degree at most 3.
- No transition selector: the round map is gated by `1 + f_24`. The flag machine (`next.f_{i+1} = f_i`, `next.f_0 + next.f_24 = f_23 + f_24`, `f_0 * f_24 = 0`, one-hot first row) keeps every row exactly one-hot over GF(2^128), and the last row of any valid trace is an output row, so the constraints hold under both the repeat-last (multi-stark) and wrap-around (debug checker) successor conventions.
- Trace height is `(25 * n).next_power_of_two()`; trailing rows are all-zero output rows.

### `Blake3BinaryAir` (`blake3-air/src/binary`)

- One row per compression, no next-row access, every constraint degree at most 2. 11,536 columns and 11,536 constraints.
- Only the `b`/`d` words after each half of G and one 31-bit carry vector per 3-operand addition are committed. The `a`/`c` words are recovered as 2-leaf XORs of rotations (`a1 = d ^ (d1 <<< 16)`, ...), and a 2-operand addition with a known result needs no columns: its carries are linear in operands and result, constrained by the characteristic-2 majority `(x + z)(y + z) + z`.
- Booleanity is asserted on the 896 input bits; every other column is forced to a bit by the addition constraints.

Both AIRs were reviewed for soundness over GF(2^128) (cells are full field elements, not bits); the negative tests pin the constraints that only matter there (flag exclusivity with non-bit flags, the round-23 wrap, non-bit carries at the top bit).

### Examples (`examples/src/binary.rs`, `prove_keccak_binary`, `prove_blake3_binary`)

A small harness builds the `MultiStarkConfig` (Keccak-256 Merkle tree wrapped in `GroupedCodewordMmcs`, `BinaryChallenger`), assesses the statement's composed security once against the target, then times the plain prover (including the trace layout) and verifier and reports rows, width, stacked variables, proof size, and composed security.

```bash
RUSTFLAGS="-Ctarget-cpu=native" cargo run --example prove_keccak_binary --release --features parallel -- --log-trace-length 14 --security-bits 96
RUSTFLAGS="-Ctarget-cpu=native" cargo run --example prove_blake3_binary --release --features parallel -- --log-trace-length 10 --security-bits 96
```

Every trace cell is committed as a `BinaryField128` element (no ring switching yet), so the proven security `BinaryPcs` can reach drops by one bit per committed variable: roughly `128 - (log-trace-length + ceil(log2(width)) + log-inv-rate + 3)`. The default `--security-bits 100` works up to `-l 12` for Keccak-f and `-l 9` for BLAKE3; the README documents this.

## Test plan

- [x] `cargo test --release -p p3-keccak-air -p p3-blake3-air -p p3-examples --lib`
- [x] Keccak AIR: honest traces satisfy constraints under both successor conventions; output rows match `p3_keccak::KeccakF`; symbolic degree 3 / 3250 constraints match the hints; negative tests for flipped state/output bits, non-boolean state, two flags in a row, skipped rounds, non-bit flags (exclusivity only), vanishing flags (wrap only), a trace ending on a round row.
- [x] BLAKE3 AIR: empty-input known answer, single-block hashes match `blake3::hash`, random inputs match a reference compression; symbolic degree 2 / 11,536 constraints; negative tests for carry/word flips, per-word input booleanity, non-bit witnesses at the top bit.
- [x] Harness: tiny binary AIR proves and verifies end to end.
- [x] Both examples prove and verify in release: Keccak `-l 12` (100.4 bits), `-l 14 --security-bits 96` (97.4 bits); BLAKE3 `-l 8` (100.7 bits).
- [x] `fmt`, `sort`, `toml`, `deps`, `clippy` (affected packages), `docs` checks from `scripts/check.py`.


blake3-air/src/binary/air.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Constraints per G step: two 3-operand and two 2-operand additions.
+const CONSTRAINTS_PER_G: usize = 2 * (31 + 32) + 2 * 32;
+
+/// Total number of constraints: input booleanity plus every G step.
+const NUM_CONSTRAINTS: usize = NUM_INPUT_BITS + NUM_ROUNDS * G_PER_ROUND * CONSTRAINTS_PER_G;
+
+/// An AIR for the Blake-3 compression function over a field of characteristic 2.
+///
+/// Each row proves one compression. The AIR reads no next row and every constraint has
```
tcoratger, line 30: This stops at the end of round 7, so the compression output (`v[i] ^ v[i+8]`, then `v[i+8] ^ cv[i]`) is never a column and never constrained. Fine for a benchmark AIR, but "proves one compression" reads like the output is pinned, and the only thing that recovers it is `compression_output` in the test module.

Worth a line here saying the output is a linear function of round 6's `b1`/`b2`/`d1`/`d2` plus the chaining value - and maybe lifting that recovery out of `#[cfg(test)]`, since it is exactly what anyone wiring this into a real chain needs. Prime-field `Blake3Air` stores `outputs`, so the difference is easy to miss when comparing the two.

blake3-air/src/binary/air.rs
```diff
@@ -0,0 +1,225 @@
...
+            c: array::from_fn(|i| u32_to_bits_le(iv_word(i))),
+            d: initial_d,
+        };
+        let mut m = local.block;
+
+        for round in &local.rounds {
+            for (g, (cols, slots)) in round.iter().zip(G_SCHEDULE).enumerate() {
+                eval_g(builder, &mut state, slots, &m[2 * g], &m[2 * g + 1], cols);
+            }
+            permute(&mut m);
```
tcoratger: The last round's permute is thrown away - `m` is never read after it. Your own `reference_compress` guards it with `if round < 6`.

It costs ~512 `Var` copies per row evaluation for nothing, and the same line is in `generation.rs:96`.

tcoratger (approved):
Went through the Keccak flag machine and the BLAKE3 carry encoding line by line and I couldn't break either of them.

The one-hot induction holds: `f_0 * f_24 = 0` plus the wrap constraint forces every successor's (round 0, output) pair to sum to a bit with product zero, so both entries are bits and `1 + f_24` really is the round gate. Your non-bit-flags test is the one that would have caught that going wrong, and it's there. On the BLAKE3 side the booleanity chain closes properly: the 896 inputs are pinned, each addition forces all 32 bits of its sum, and every stored word is the XOR of a forced sum with an already-bit word, so nothing is left free. Nice trick recovering `a` and `c` instead of committing them.

Two small things below, both on the BLAKE3 AIR. Approving - take what you agree with and merge.

PR #2165, opened 2026-09-17
perf(sumcheck,multi-stark,binary-field): faster binary PCS openings, zerocheck kernels and tower serialization

Nashtare:
## Summary

Prover-side speedups found while profiling wide characteristic-2 hash traces through `p3-multi-stark` + `BinaryPcs` (Keccak-f: 1625 bit columns × 2^14 rows; BLAKE3: 11,536 bit columns × 2^10 rows, from the companion examples PR #2164). Every change keeps proofs and transcripts byte-identical.

### PCS opening (`p3-sumcheck` suffix prover)

- **Residual factors across column slots.** `compress_stacked_scaled` and `combine_weights` looped over columns serially, and each per-column kernel only parallelizes above 4096 entries. Columns own disjoint output slots, so both factors now fill in parallel across slots (one shared `column_slots` helper). With no folded variables and unit scale, compression is a copy.
- **Shared weight tables.** `combine_weights` builds each claim's equality and successor tables once at unit scale and sums every contribution to a slot in one pass.
- **Openings against one equality table.** With zero SVO folding and a table of at most 2^20 rows, `record_opening` builds the equality table (and the repeat-last successor table) once per call and evaluates each column as one weighted sum: zero cells are skipped, one cells add their weight, and a chunk falls back to one product per row from its first non-bit row. Taller tables keep the factored evaluation.
- **Parallel copies.** The stacked witness, the codeword seed, and the padded transform's large coset copies are split across workers.

### Zerocheck (`p3-multi-stark`)

- **Evaluated nodes only.** The row scratch steps only between interpolation nodes some AIR evaluates.
- **Declared successor columns only.** Filling, node stepping and repeat-last tail folds touch only `main_next_row_columns` / `preprocessed_next_row_columns` (undeclared successors read as zero on both sides; `validate_successor_columns` rejects any AIR that reads one).
- **Alpha powers instead of Horner.** Constraints are batched with precomputed `alpha^(n-1-i)`, counting boundary pins, which is exactly the verifier's Horner coefficient per constraint. A release assert rejects a count mismatch.
- **`bool_check` forwarded through `PackedExt`.**

### Zerocheck backends (`p3-field`, `p3-binary-field`, `p3-multi-stark`)

- **`prove_with_backend`.** The zerocheck kernels dispatch through a sealed `ZerocheckBackend` type parameter. `prove` is `prove_with_backend::<GenericBackend>`, so existing callers, `ProverAir`/`VerifierAir`, and the verifier are unchanged.
- **`HasSubfield<S>`** (p3-field): a field with a small subfield `S` (`Field + Algebra<S>`, `as_subfield`, bulk `all_in_subfield`). `BinaryField128` implements it for GF(4) (`BinaryField2`), with a branchless masked product by GF(4) elements; `Ghash128` gains `Algebra<BinaryField128>` through the basis isomorphism.
- **`SubfieldBackend<S>`.** In characteristic 2 the first zerocheck round evaluates at interpolation nodes that lie in GF(4), so when every cell, public value, and interpolation step of a stage fits GF(4) and the stage has no lookups, round 0 evaluates the AIR entirely inside GF(4) (`SubfieldVar`/`SubfieldAcc`, with a poison flag on anything that does not fit) and the first fold scales by GF(4) differences instead of general products. Any misfit falls back to the generic kernel; transcripts are byte-identical to `GenericBackend` in every case, including each fallback trigger.
- **`ReprBackend<S, R>`.** Round 0 as `SubfieldBackend<S>`, then every later round runs in a representation field `R` isomorphic to the challenge field (`Ghash128` for `BinaryField128`: one carryless product with no basis changes, instead of three tower↔polynomial basis changes around it). Columns, alpha powers, lookup coefficients, repeat-last tails and selector prefixes cross into `R` once per stage; eq weights, interpolation steps (the challenge field's own nodes, carried into `R`) and the fold challenge once per round; round sums and openings cross back before the transcript sees them. `RoundStateExt` gains the arithmetic-field parameter with `R = EF` for the other backends, so prime-field kernels are unchanged (pinned by a BabyBear transcript digest; Poseidon2 zerocheck bench −0.6%, p = 0.55).

### `p3-binary-field` / `p3-binary-pcs`

- **Booleanity by squaring.** Tower fields and `Ghash128` compute `bool_check` as `x² + x`; squaring is a linear map with no basis change or carryless multiply.
- **Allocation-free serialization.** Tower fields stream to parallel byte and word streams without allocating a `Vec` per element (whole `u64` words for 64/128-bit levels, packed little-endian for narrower ones), with tests pinning partial-word tails at every level.
- **Allocation-free grouped Merkle leaves.** `GroupedCodeword` reads packed leaf rows directly.
- **Word digests in `BinaryChallenger`.** `Hash<F, u64, N>` and `MerkleCap<F, [u64; N]>` are absorbed as the little-endian bytes of each word, matching `SerializingChallenger32/64`, so `BinaryPcs` can use the vectorized `PaddingFreeSponge<KeccakF, 25, 17, 4>` Merkle tree the prime-field examples use.

## Measurements

Apple M4 Pro (10P+4E), release, `-Ctarget-cpu=native`, `parallel`, examples from #2164 at `--security-bits 96`, A/B runs interleaved on the same machine (which was shared with other workloads, so ranges rather than single numbers):

| Workload | Before this PR | After this PR |
| --- | ---: | ---: |
| Keccak-f `-l 14` prove | 3.96–4.18 s | 1.95–1.99 s (`prove`), 1.25–1.27 s (`ReprBackend<BinaryField2, Ghash128>`) |
| BLAKE3 `-l 10` prove | 1.72–1.80 s | 0.72–0.77 s (`prove`), 0.63–0.64 s (`ReprBackend<BinaryField2, Ghash128>`) |

Per phase for Keccak-f `-l 14`: `into_sumcheck` 1.45 s → 60 ms, `record_opening` 340 ms → 16 ms, PCS commit ≈ 900 ms → 700 ms, zerocheck ≈ 900 ms → 810 ms with `prove`, ≈ 450 ms with `SubfieldBackend<BinaryField2>`, and ≈ 130 ms with `ReprBackend<BinaryField2, Ghash128>` (Keccak-f `-l 16`: 1.75–2.10 s → 0.45–0.47 s; BLAKE3 `-l 10`: ≈ 180 ms → 59 ms). The examples in #2164 opt into the backends once both PRs are in.

Prime-field check, `cargo bench -p p3-multi-stark --bench poseidon2` (BabyBear Poseidon2 zerocheck, NEON), before vs after the zerocheck commits: 1 AIR −4.1%, 5 AIRs −5.5%. Not measured on AVX-512, where the packed alpha-power tables are wider.

## Test plan

- [x] `cargo test --release -p p3-binary-field -p p3-sumcheck -p p3-binary-pcs -p p3-binary-dft -p p3-merkle-tree -p p3-matrix -p p3-whir -p p3-multi-stark`
- [x] Debug: `cargo test -p p3-binary-field -p p3-sumcheck -p p3-binary-pcs -p p3-binary-dft`, `cargo test -p p3-multi-stark --lib` (exercise the debug cross-checks and overflow checks)
- [x] Suffix openings at zero folding match the `SvoPoint` evaluations on random, bit, all-zero, all-one and mixed columns over `BinaryField128` and a BabyBear extension, across table heights 2^0..2^13, plus `claims.sum(alpha) == <compressed, weights>`
- [x] Zerocheck: successor subset/none twins over prime and binary fields at 64, 4 and 2 rows (packed and scalar kernels) match the full declaration; interpolation node schedule per family degree; alpha-power batching matches Horner including boundary pins; wrong power counts panic
- [x] `GroupedCodeword::vertically_packed_row` matches the trait default across wrap-around shapes; tower stream tails 0..=17 at every level; `bool_check` equals `x·(x−1)` for every tower level and `Ghash128`
- [x] Proof digests of every multi-stark test proof and both binary examples unchanged across the zerocheck commits
- [x] `SubfieldBackend` vs `GenericBackend` transcripts byte-identical on a bit-valued degree-3 AIR with successor, preprocessed and periodic columns, pins, grinding, mixed heights and degrees, and every fallback trigger (non-GF(4) main/preprocessed/periodic cell, AIR constant, public value, degree-4 AIR, lookup stage); `HasSubfield` contract tests including non-members whose only extra bits are above bit 63
- [x] `ReprBackend` transcripts byte-identical to `GenericBackend` on every backend fixture, plus a full-width non-bit AIR (round 0 falls back, later rounds in `Ghash128`), grinding, and stages tall enough to fold and convert in parallel; mutation checks (R's own interpolation nodes, unconverted tails, table built without the challenge, swapped selector prefix, zeroed lookup coefficients) each fail tests
- [x] Prime-field zerocheck transcript pinned by digest (BabyBear, packed and scalar kernels, preprocessed/periodic/successor columns, pin, lookup, two stages)
- [x] `cargo test --release -p p3-multi-stark --features parallel --all-targets`
- [x] `fmt`, `sort`, `toml`, `deps`, `clippy`, `docs` checks from `scripts/check.py`


sumcheck/src/layout/prover/suffix.rs
```diff
@@ -96,6 +117,46 @@ impl<F: Field, EF: ExtensionField<F>> Layout<F, EF> for SuffixProver<F, EF> {
         // The opening point lives in the table's local frame, one coordinate per variable.
         debug_assert_eq!(point.num_variables(), table.num_variables());
 
+        // Without SVO rounds an opening carries no per-round residuals.
+        // Each column then reduces to one weighted sum against a table shared by the whole batch.
+        // Taller tables keep the factored evaluation, whose weights stay at the square root size.
+        if self.claims.folding == 0 && table.num_variables() <= SHARED_WEIGHTS_MAX_VARIABLES {
```
tcoratger: The gate looks at the table's height but not at how many columns the batch actually opens. Building `eq` costs 2^n extension multiplies and 2^n * |EF| bytes - at the cap that's 16 MiB for `BinaryField128`, plus another 16 for `next_weights`.

It pays for itself the moment two or more columns share it, which is the normal case. But with one column you've bought a dense table to serve a single weighted sum that the factored path would have done with cheaper F x EF products, and with an empty batch on both sides you build it for nothing.

Adding `current.len() + next.len() >= 2` to the condition keeps the whole win and drops those cases.

sumcheck/src/layout/prover/suffix.rs
```diff
@@ -489,43 +563,100 @@ impl<F: Field, EF: ExtensionField<F>> SuffixProver<F, EF> {
...
+        // Every contribution to a slot is summed in one pass over it.
+        column_slots(
+            &self.claims.placements,
+            &self.claims.tables,
+            rs.num_variables(),
+            out.as_mut_slice(),
+        )
+        .into_par_iter()
+        .for_each(|(slot, table_idx, poly_idx)| {
+            let terms: Vec<(&[EF], EF)> = column_weights[table_idx][poly_idx]
```
tcoratger: This allocates a fresh `Vec` per column inside the parallel loop - ~11.5k tiny allocations per call on the BLAKE3 trace, right where you're trying to save time.

`column_weights[table_idx][poly_idx]` is already the slice you want: matching on it directly and resolving `claim_idx -> &[EF]` inside each arm gives you the same three cases with no allocation.

tcoratger (approved):
Read this one carefully since it touches the opening prover and the zerocheck kernels. Everything I checked comes out algebraically identical to what it replaces:

- `successor_weights` is exactly the repeat-last reindexing, last row folded in correctly.
- pulling the scale out of `combine_weights` works because `accumulate_into` and `accumulate_next_suffix_into` are both linear in it, and the alpha walk order is byte-for-byte the old one.
- `column_slots` reproduces the old offsets (`Table::num_variables()` is the shared per-column value), and it panics rather than aliasing if two slots ever overlap.
- skipping undeclared successors is safe because `validate_successor_columns` already rejects an AIR that reads one, and the scratch starts zeroed - same filler the verifier's folder sees.
- the alpha-power table is the same sum Horner computes, and promoting the count check to a release assert is the right call: a silent off-by-one there shifts every weight and nothing downstream notices.

Two perf notes below, no correctness concerns. Approving.

PR #2166, opened 2026-09-17
feat(binary-pcs)!: build the small-field commitment path

tcoratger:
A RISC-V trace is bits and 32-bit words. Committing each of them as a 128-bit element is a
4x to 128x embedding overhead paid on the single most expensive operation in the prover.
This builds the path that removes it: the commitment no longer names a field, a narrow
alphabet packs into committed elements by reinterpretation, a Boolean commitment sits on top
of both, and several openings of it share one claim pool and one commitment opening.

Every number below was measured on this machine (AMD Ryzen 9 9950X3D, Zen 5, `avx512f`,
`gfni`, `vpclmulqdq`) with `-C target-cpu=native`, through Criterion at the stated shape.
All of them come from one single-threaded run of the shipped branch, with the `parallel`
feature off, so every arm of every table is comparable to the arm beside it. Proof sizes are
`postcard` byte counts and are build-independent.

---

## 2.1 — un-hardcode the committed field

`p3-binary-pcs` named `BinaryField128` in 217 places. The committed alphabet and the
challenge field are now separate type parameters: columns and the base codeword live in the
alphabet, every challenge and folded codeword in the challenge field, and the first fold of
the schedule is the only thing that crosses between them. `p3-security`'s binary regime
charges every algebraic error against an explicit challenge width instead of a constant 128,
and the schedule now also rejects a codeword longer than the alphabet's own additive domain
and caps the grinding difficulty by the alphabet's width, because the witness is one of its
elements.

The fold gains a route table: one implementation per alphabet, each naming the fastest route
that alphabet reaches. The widest level keeps the packed polynomial-basis route; a narrower
one lifts with a tower-basis round and hands the rest to the wide field's own route.

**Read this before the tables: this item lands a capability, not a speedup.** A narrow
committed alphabet is now expressible, correct and verified end to end, and on this machine it
is very slightly *slower* than committing everything at 128 bits. The reason is that
`GF(2^64)` has no polynomial-basis transform and no bit-sliced fold, so the encoder falls back
to the tower-basis Lin–Chung–Han transform: halving the committed bytes is cancelled by a
transform that costs more per element. It becomes a win the moment `GF(2^64)` gains a
polynomial-basis transform, which is what PR #2162's `poly64.rs` is for. What this item is
worth today is that the alphabet is a type parameter instead of 217 hard-coded mentions, which
is what 2.2, 2.3 and 2.7 are built on — none of those would be expressible without it.

Folding 2^20 codeword symbols, one round:

| route | before | after | gain |
| --- | --- | --- | --- |
| packed, GF(2^128) symbols | 8.857 ms | 8.857 ms | unchanged |
| lifting, GF(2^64) symbols | not expressible | 36.788 ms | new configuration |
| tower, GF(2^64) symbols and challenge | not expressible | 58.858 ms | new configuration |

The packed route is 4.2x faster than the lifting one and 6.6x faster than the pure tower one,
which is why the route table dispatches to it rather than one generic loop replacing it.

Committing a 2^20-cell table, GF(2^64) columns with GF(2^128) challenges against GF(2^128)
columns throughout:

| phase | GF(2^128) columns | GF(2^64) columns | gain |
| --- | --- | --- | --- |
| commit | 336.83 ms | 341.43 ms | 0.99x (a small loss) |
| open | 422.06 ms | 462.64 ms | 0.91x (a small loss) |
| proof size | 520 004 B | 516 218 B | 1.007x |

Those two small losses are the caveat stated above, measured. The proof does shrink, but only
by the base openings' share of it, which is a fraction of a percent at this query count.

## 2.2 — a packed witness type

A tower level holds its elements in the basis its bytes already define, so one run of narrow
cells and the wide elements holding it are the same bytes in the same order. Packing is one
copy and unpacking is that copy back, with no arithmetic on either side. The layout contract
is stated once, as an `unsafe` marker trait, and implemented by the byte-aligned tower levels
and by the bit-sliced `GF(2)` packings. A bit-sliced trace that arrives one block per row with
a lane per column is turned column-major by the square bit transpose that just landed.

2^20 rows over 40 bit columns:

| quantity | one element per bit | packed | gain |
| --- | --- | --- | --- |
| elements a commitment holds | 41 943 040 | 327 680 | 128x |
| building that layout | 140.27 ms | 801.2 µs | 175x |
| row-major exchange | 16.744 ms (lane by lane) | 801.2 µs (square transpose) | 20.9x |

2^20 32-bit words:

| route | through the field's basis accessor | reinterpreted | gain |
| --- | --- | --- | --- |
| packing one column | 93.286 µs | 53.348 µs | 1.75x |

The basis accessor is also the independent reference the tests pin the reinterpretation
against, so the same arm is both the baseline and the oracle.

## 2.3 — a Boolean multilinear PCS

Commit a function from the hypercube to `{0,1}`; open its multilinear extension at any point
of the challenge field. Three pieces behind one trait: the packing, the bit ring switch, and
the commitment that discharges the surviving claim. The bit-alphabet reduction already existed
as algebra; what was missing was the protocol around it, so `p3-sumcheck` gains its transcript
and its two sides, as methods on the reduction itself. Those fix the one ordering the
reduction's soundness rests on — the batching challenge is drawn only after the tensor element
is bound, because a bit matrix solving two GF(2)-linear systems can move the claim with the
sum held. The element travels by rows, `d` elements rather than the `d^2` a byte per
coefficient would cost.

Booleanity is free, and the code says so rather than implying it. The packing is a bijection
between bit strings and elements of the level: every bit pattern is an element, and every
element is some bit pattern. A commitment to a packed multilinear is therefore a commitment to
a bit witness, and **no commitment exists to a function the hypercube sends outside `{0,1}`**.
There is no range check because there is nothing one could rule out.

One Boolean column of 2^20 bits, opened at a random point:

| phase | one element per bit | Boolean commitment | gain |
| --- | --- | --- | --- |
| commit | 354.88 ms | 2.2504 ms | 157.7x |
| open | 504.53 ms | 6.3524 ms | 79.4x |
| proof size | 558 842 B | 187 578 B | 2.98x |

Both arms commit the same function and open it at the same point. The baseline is the same
binary PCS at 2^20 committed variables, which is the only way this workspace could commit a
bit witness before this PR.

## 2.7 — one claim pool, one opening

Several reductions over one committed polynomial each leave a surviving claim behind.
Discharging them one at a time costs one commitment opening each. The pool collects them and
one draw folds them, so a single opening answers for all of them and no sumcheck runs for the
batching itself. The ordering is enforced twice rather than documented: the challenge lives
only on the sealed pool, sealing consumes the open pool so nothing can be deposited afterwards,
and the draw is the last described transcript step, after both message steps. The weights are
distinct powers of that one challenge, and the `(k - 1) / |F|` fold error is charged in
`p3-security` under its own label.

2^20 bits, four opening points:

| quantity | one proof per point | one pool | gain |
| --- | --- | --- | --- |
| prove | 25.853 ms | 15.803 ms | 1.64x |
| proof size | 738 242 B | 194 585 B | 3.79x |

## 2.6 — the Merkle layer

All four pieces the item asks for are already in this workspace and already tested there: cap
layers (`MerkleTreeMmcs::new`'s cap height, `MerkleTree::cap`), batched multi-opening over the
minimal frontier (`merkle-tree/src/pruning.rs`, 206 passing tests), a parallel digest over
packed leaves (`first_digest_layer`, `packed_row_tests.rs`), and a coset-sized leaf layout
(`GroupedCodewordMmcs`). Because 2.1 made the commitment scheme a type parameter, the Boolean
interface takes any of them, which this PR pins with a test grid over arity and cap height,
every cell of it opening the same bits at the same two points. That test runs at 2^17 bits to
stay cheap in CI; the table below is the same grid widened and run at 2^19.

The measurement is a negative result worth recording. 2^19 bits, two opening points, proof
size in bytes:

| folding arity | plain, cap 0 | grouped, cap 0 | plain, cap 4 | plain, cap 8 |
| 1 | 80 317 | 84 185 | 88 702 | 118 824 |
| 3 | 56 039 | 56 806 | 58 005 | 67 411 |
| 6 | 163 537 | 163 739 | 163 980 | 168 635 |
| 12 | 329 817 | 329 821 | 329 813 | 329 817 |

| knob | best | worst | spread |
| --- | --- | --- | --- |
| folding arity | 56 039 B (arity 3) | 329 817 B (arity 12) | 5.9x |
| cap height | 56 039 B (cap 0) | 67 411 B (cap 8) | 1.20x |
| leaf grouping | 56 039 B (plain) | 56 806 B (grouped) | 1.01x |

The batched opening already prunes to the minimal frontier, so a cap mostly widens the
committed folded words it covers and a coset-sized leaf mostly replaces paths the frontier had
already merged. Neither moves this proof by more than a tenth, while the folding arity moves it
by a factor of six. No new Merkle code was written for this item; the honest deliverable is the
grid test and this table.

## What is deferred

**2.4, small-field multi-STARK, is no longer deferred.** It is in this PR, with its own section
at the bottom. What survives of the old note is one of its two blockers, now measured rather
than predicted: the trace is still lent back as one full-width field element per bit, because
`MultiStarkConfig::Val` cannot be a one-bit field under `p3-field`'s current
`BasedVectorSpace` contract. So the saving lands on the commitment and not on the trace, and
constraints still run one cell at a time. The section gives the evidence for both, and the
before/after/gain numbers for what the commitment side does save.

**2.5, jagged stacking, is cut.** Columns of arbitrary height committed end to end with no
padding, with opening claims routed by public selectors, is a larger piece than the four above
put together, and merging the stacking and ring-switch sumchecks over the dense domain is a
second piece on top of it. Nothing here blocks it: the packed stack already reports each
column's own arity and the claim pool already batches claims at unrelated points, which are the
two interfaces jagged stacking would deposit into.

Two smaller notes:

- **2.1's narrow-alphabet configuration is a capability, not a speedup yet.** It is correct and
  verified but marginally slower than committing at 128 bits, because `GF(2^64)` has no
  polynomial-basis transform. Spelled out in full in that section; repeated here so a reader
  skimming the deferrals does not miss it.
- The bit-alphabet reduction's own error now *does* reach a soundness estimator, under its own
  label in `p3-security`. That arrived with 2.4 below; the general ring switch is still left for
  a composing protocol to charge.

## Files outside the stated scope

`p3-binary-dft`'s encoder is opened to every byte-aligned tower level, through one shared body
and one implementation per alphabet — a blanket implementation would overlap the two-adic one
in `p3-commit`. One surgical change in `p3-binary-field`: `BinaryChallenger` now absorbs a
Merkle digest tagged with any field, because a digest is a byte string and the field is a
marker, and a run committing at two levels absorbs both through one sponge. None of PR #2162's
files (`aes/`, `poly64.rs`, `poly192.rs`, `clmul/`, `lib.rs`) are touched.

## 2.4 — multi-STARK integration

A reviewer asked for the bundle rather than the commitment scheme alone, so this is now in the
PR: **a complete AIR proof over a Boolean trace, committed through the Boolean scheme, proving
and verifying**, at `multi-stark/examples/prove_boolean_trace.rs`.

### What it adds

`BooleanTracePcs` is a `MultilinearPcs` + `PrescribedPointPcs` over a batch of trace tables. It
takes one table per AIR instance, plans their slots through the stacked-layout planner the
commitment schemes already use, gathers every cell's bit into one witness, and commits that. A
column claim lifts to a bit claim by prefixing its slot address to the row point:

```text
    column in slot s, 2^a rows  ->  bits [s * 2^a, (s + 1) * 2^a)
    its value at a point r      ->  the bit witness at (bits of s, r)
```

A slot starts at a multiple of its own length, so its address is a bit string and the reduction
underneath answers the lifted claim unchanged. Both sides read the slots from the same planner,
so neither can pick its own layout.

The example AIR is a bit-sliced full-adder gate table: five columns, `sum = a + b + cin` and
`cout = a*b + cin*a + cin*b`, with the first row's three inputs pinned as the public boundary.
Two instances of different heights share one commitment and one proof. Seven tests cover the
round trip on both arms, the mixed-height batch, a tampered public boundary, an invalid carry
column, the security report's labels, a non-Boolean cell, and a refused successor view.

**Booleanity is still free, and still comes from the packing.** The packing is a bijection
between bit strings and the elements a commitment holds, so no commitment exists to a function
the hypercube sends outside `{0, 1}`. There is no range check because there is nothing to rule
out. A cell holding anything else is refused at commitment rather than constrained, and an
accepted proof's trace agrees with the bits the opening certifies or the zerocheck's final
equality fails.

**The security accounting charges the Boolean path, term by term.** `PrescribedOpeningSecurity`
now carries labelled terms instead of one pre-summed number, because an opening that stacks a
reduction on a commitment charges both and one number cannot say which of them is short. The
multi-STARK report charges each of them under its own label, and a test pins that all three
appear:

```text
    binary-pcs-opening     the packed multilinear's own opening budget
    bit-ring-switch        (d_log + 2 l') / |F| per point, unioned over the points
    claim-pool-batching    (k - 1) / |F| for k claims folded under one challenge
```

The middle one is new in `p3-security`. It closes the deferral note that said the bit-alphabet
reduction's error reached no estimator — it does now. The claim pool's ordering property is
untouched: weights are distinct powers of one challenge drawn after every claim value is bound,
and the pool is consumed by sealing, so the unsound order stays unreachable.

### The design choice, and what the alternative costs

Two directions were on the table. I took the second, **keep the field-element table for the AIR
and pack at commit time**, and the first turned out not to be reachable at all.

The first was to hold the trace bit-packed and lend the AIR a view of it as single-bit cells,
paying nothing on the AIR side. `MultiStarkConfig` lends the trace back as `&Table<Self::Val>`,
so this needs `Val` to be a one-bit field. It cannot be:

- `MultiStarkConfig` requires `Challenge: ExtensionField<Self::Val>`, which requires
  `BasedVectorSpace<Self::Val>`, whose `as_basis_coefficients_slice(&self) -> &[F]` is a
  **borrowed** slice. `BinaryField128` over `Gf2` has dimension 128, and `Gf2` is
  `#[repr(transparent)] struct Gf2(u8)` — 128 borrowed bytes cannot come out of a 16-byte
  element. Confirmed by compiling the bound: `the trait bound BinaryField128:
  BasedVectorSpace<Gf2> is not satisfied`. Lifting that needs the `p3-field` trait itself to
  change, which touches every field in the workspace and does not belong on the binary track.
- Independently, `PrescribedPointPcs` pins `Challenger: GrindingChallenger<Witness = Self::Val>`,
  and the Boolean scheme draws its own challenges and grinds over the field its packed elements
  live in. So `Val` is forced to that field whatever the first point says.

So `Val = BinaryField128`, one cell per trace bit at the full challenge width. The chosen
direction forces no second copy of the trace: `build_witness` moves the tables through, `commit`
gathers their bits into a buffer 1/128 their size, and the prover data lends the same tables back.

The honest consequence is that the 128x saving lands on the commitment and not on the trace. At
2^18 rows the trace is 20.00 MiB of the packed arm's 33.77 MiB peak — 59% of it, and the largest
single object left. **The trace-side saving is what the `p3-field` change would unlock; the
commitment-side saving is what this delivers.**

### Measured

One prove and one verify of the same statement, same trace, differing only in the commitment
scheme the configuration selects. **Before** is the trace committed one field element per bit at
128 bits (`BinaryPcs` over the stacked table); **after** is the Boolean scheme. Peak heap is
live bytes at the high-water mark from a tracking allocator, armed before the trace is built so
that both arms are charged for it. Wall time is a separate unarmed pass. Rate 1/4, folding
arity 3, Keccak-256 Merkle, 100-bit composed target enforced on both arms, one thread.

| rows | cells | peak heap before | peak heap after | gain | time before | time after | gain |
|---|---|---|---|---|---|---|---|
| 2^14 | 81 920 | 18.01 MiB | 2.12 MiB | **8.5x** | 260 ms | 58 ms | **4.5x** |
| 2^16 | 327 680 | 72.01 MiB | 8.45 MiB | **8.5x** | 1 096 ms | 207 ms | **5.3x** |
| 2^18 | 1 310 720 | 288.01 MiB | 33.77 MiB | **8.5x** | 4 730 ms | 797 ms | **5.9x** |

With `-C target-cpu=native` the heap figures are identical and the time gain is smaller, because
carry-less multiplication speeds up the long arm's encoding more than the short one's:

| rows | time before | time after | gain |
|---|---|---|---|
| 2^14 | 50 ms | 27 ms | **1.8x** |
| 2^16 | 195 ms | 90 ms | **2.2x** |
| 2^18 | 812 ms | 332 ms | **2.4x** |

The packed proof at 2^18 is 156 902 bytes. Reproduce with
`cargo run --release -p p3-multi-stark --example prove_boolean_trace`; the example prints the
table itself, so the numbers are not copied by hand.

### The two costs, stated plainly

- **AIR-side cell width.** One `BinaryField128` cell per trace bit, 16 bytes, for the reasons
  above. That is 128x the bit-packed minimum, not 8x — the 8x figure would be a `Gf2` cell, and
  `Gf2` cannot be this `Val`.
- **Scalar constraint evaluation.** Every binary tower level has `type Packing = Self`, width
  one, so the zerocheck evaluates constraints one cell at a time. Making the bit-sliced packings
  serve as a base field's packing needs a conversion-based packing abstraction in `p3-field`
  (`PackedValue` casts by reinterpretation, bit-slicing is compression); that is not attempted
  here. This cost is identical on both arms, so it does not move the table above.

### What this does not open

Reading one row ahead. A batch may name columns at the current row and columns one row ahead;
the second is not an evaluation at a point, it weights the column by an equality table shifted
up by one row. The bit ring switch reduces evaluations at points and nothing else, so such a
batch is refused with a typed error rather than answered at the wrong point, and the example's
AIR reads only the current row. Lifting this means generalising the reduction from a point to a
tensor weight, since the shift decomposes into the carry/done/omega tensor terms the layout
sumcheck already uses — a bounded piece of work, but a separate one with its own soundness
accounting.

One smaller change fell out: the commitment's binding is now a call of its own on the Boolean
trait, as it already is on `MultilinearPcs`, so a protocol holding several commitments binds
each of them once in its own order rather than having verification absorb one for it.


Nashtare:
The description mostly talks about runtime performance but I'd assume the main goal of this PR is prove memory reduction? Have you benchmarked it by any chance?

tcoratger:
Just added a peak-heap harness and ran it (`cargo run --release --example peak_heap -p p3-binary-pcs`), here are the results:

```
peak heap during one operation, MiB, single-threaded

operation              size      embedded       packed      gain
commit                 2^16         22.00         0.17    127.5x
open                   2^16         19.54         0.35     55.3x
witness held           2^16          1.00         0.01    128.0x

commit                 2^18         88.00         0.69    127.9x
open                   2^18         76.70         0.89     85.7x
witness held           2^18          4.00         0.03    128.0x

commit                 2^20        352.00         2.75    128.0x
open                   2^20        304.88         2.86    106.7x
witness held           2^20         16.00         0.12    128.0x
```

Two things worth pointing out. Commit holds the full 128x at every size, so the saving is not just the witness but the codeword and the Merkle tree over it. Opening starts at 55x and climbs to 107x, because the ring-switch tensor and the extension-field sumcheck buffers are a fixed cost that does not shrink with the witness, so they dominate the small sizes and wash out at the large ones.

Nashtare:
Thanks for the numbers! BTW it seems this explicitly leaves multi-STARK integration out, why is that? Would probably be cleaner / easier to review the whole bundle with integration

tcoratger:
Fair question, and my answer in the description was incomplete — I'd only given the reason that makes integration *slow*, not the one that makes it pointless. Rebased on main and pushed both the sharpened description and a note on the trait itself, so the next consumer finds it without reading a PR body.

The blocker that matters: `MultiStarkConfig::committed_table` hands the trace back as `&Table<Self::Val>`, a borrow of materialized field elements that the constraint evaluation reads row by row. A bit witness has no such table to lend, and building one unpacks every cell into a whole field element — which gives back the 128x saving at exactly the boundary where the AIR reads the trace. So an integration today would undo the thing the PR is for.

On top of that, the hot paths go through `Val::Packing`, and the bit-sliced packings can't be `Gf2::Packing`: `PackedValue` casts between a packed value and `[Scalar; WIDTH]` by reinterpretation, while bit-slicing is compression.

So the bundle you'd want is really three changes: this scheme, a witness borrow that carries packed blocks, and a conversion-based packing abstraction in `p3-field`. Both of the others live in other crates and would land as `!` changes of their own. I'd rather this one be reviewable on its own terms than grow into all three — but if you'd prefer to review them together I'm happy to hold this and stack them.

tcoratger:
@Nashtare I'm trying locally to build everything together and I'll let you know if this is really too big or not, will keep you updated here

tcoratger:
Correcting myself on one number above, and then taking your suggestion.

I said building the borrowed table "unpacks every cell into a whole field element" and put 128x on that. That is wrong. `Gf2` is `#[repr(transparent)] struct Gf2(u8)`, so a cell is **one byte**, and the give-back at the AIR boundary is **8x**, not 128x. The 128x in the description is against committing at 128 bits, which is a different comparison. The cost is real but much smaller than I implied, which makes your request more reasonable than my reply made it sound.

So we're doing the bundle. I'm building the integration onto this branch now: an end-to-end multi-STARK proof over a boolean trace committed through this scheme, with peak heap and wall time measured against the same statement committed at 128 bits.

Two costs I expect to have to report rather than remove, and I'd rather flag them up front than have them found in review:

- The AIR-side cell width above, unless the borrow can be made to hand out a view of the packed trace instead of a materialised table — which is the design question I'm actually weighing.
- Constraint evaluation runs one cell at a time, because `Gf2::Packing = Self` at width one. Fixing that needs a conversion-based packing abstraction in `p3-field` so the bit-sliced packings can serve as the base field's packing, and that touches every field in the workspace. I'm deliberately leaving it out of a binary-track PR and will say so in the description.

If the integration turns out not to be expressible cleanly without that `p3-field` change, I'll come back with the evidence rather than force a design around it.

tcoratger:
Retracting my correction above — it was wrong, and my original 128x was right. Sorry for the noise.

I assumed `Val = Gf2` was an available configuration, which is where the one-byte cell came from. It isn't, for two independent reasons that both show up as compile errors:

- `Challenge: ExtensionField<Val>` requires `BasedVectorSpace<Gf2>`, whose `as_basis_coefficients_slice` returns a **borrowed** `&[Gf2]`. A 16-byte element cannot lend out 128 bytes, so `BinaryField128: BasedVectorSpace<Gf2>` does not hold.
- Independently, `PrescribedPointPcs` pins `GrindingChallenger<Witness = Self::Val>`, which forces `Val` to the field the challenger grinds in.

So `Val` is the 128-bit field, the AIR cell is 16 bytes, and materialising the trace for constraint evaluation really does cost 128x the bit minimum. That makes your original point sharper than either of my answers: the commitment saving and the AIR-side cost are genuinely in tension, and the tension is enforced by the trait bounds rather than by a choice in this PR.

The integration is pushed regardless, and the measured result is that it is still clearly worth it — peak heap 8.5x and wall time up to 5.9x on the same statement — because the commitment, not the trace, is the dominant allocation. At 2^18 the trace is 20.00 MiB of the packed arm's 33.77 MiB, so it is now the largest object left and the next thing worth attacking.

Details, the design alternative I rejected with its compile evidence, and the full tables are in the updated description.

binary-pcs/src/boolean.rs
```diff
@@ -0,0 +1,919 @@
...
+            )
+            .map_err(BooleanPcsError::Commitment)?;
+
+        // One fold closes the whole pool, rather than one equality per claim.
+        let opened: Option<Vec<EF>> = evals
+            .iter()
+            .map(|batch| batch.current().first().copied())
+            .collect();
+        let opened = opened.ok_or(BooleanPcsError::SurvivingClaim)?;
+        if opened.len() != points.len() || sealed.fold(opened) != sealed.folded_value() {
```
Nashtare: The inner `verify_at` already checks every evaluation it returns: they are batched under the layout's own `alpha`, and `opening_term(k)` charges that as `(k - 1)/|K|`. So `opened[i]` is already pinned to `t'(r'_i)`, and comparing `opened` to the surviving values one by one costs the same `k` operations with no error at all. The pool's fold is a weaker form of that check, its challenge is drawn before `opening.evals` is sent, and the prover discards `_sealed`. As far as I can tell, the one-opening saving in 2.7 comes from `try_open_at` taking `k` points, not from the pool. Could this compare directly and drop `claim-pool-batching`? Or could it pass `weight_poly` to the commitment as one weighted claim, so the pool really removes claims?

Nashtare (commented):
I read this in three parts: the sumcheck side and the security terms, the un-hardcoded committed field with its packing, and the Boolean commitment with `BooleanTracePcs`. I found no soundness hole.

**What holds**
- The bit ring switch matches Diamond–Posen 2024/504, Construction 3.1, step for step. `r''` is drawn after the point and all `d` tensor rows are bound. `(d_log + 2 l') / |F|` is the Theorem 3.5 bound, and multiplying by `k` points is a valid union bound.
- The field split is clean. Every challenge is drawn in `EF` and every committed symbol, the grinding witness included, is in `F`. Both sides lift with the same low-bits embedding. The GF(2^128) path keeps its fold kernel, encoder body and transcript bytes.
- `BooleanTracePcs`: the slot prefix is MSB-first with rows in the low coordinates, and slots come from verifier-built shapes. The root is bound before the zerocheck point. Successor batches are refused on both sides, and no verifier panic is reachable from proof data.
- Locally I ran all ten narrow alphabet/challenge pairs end to end, with folding 1 to 4 and grinding up to each cap. They verify and reject a tampered base symbol. Mixed heights `[9, 3, 2]` and `[5]` round trip.

- The config isn't tied to `(F, EF)`. A 128/128 config accepted by a 64/64 scheme reports ~101 bits where the matching config refuses the parameters (inline, `pcs.rs`).
- A Boolean opening with a single column claim gets an infinite-bits term, so `prove` refuses a valid statement (inline, `multi-stark/src/security.rs`).
- No test reaches `SurvivingClaim`, the one check that ties the reductions to the commitment. Deleting the pool fold keeps every test green. `bits/transcript.rs` has no tests either.

**Worth discussing**
- The claim pool is sound, but in its only caller the inner PCS already pins every opened value. The 2.7 saving comes from `try_open_at` taking `k` points.
- Each opened column runs its own ring switch over the whole stacked witness, so prover work grows with the square of the trace width.
- `peak_heap.rs` arms its counter after the inputs and prover data exist. The figures in this thread are lower bounds on net allocation, not peak heap.

The rest is inline. Separately, this conflicts with #2174 and #2177 in `binary-pcs/src/prover.rs` and with #2182 in `examples/src/binary.rs`. Those are mine, and I'll rebase whichever lands second.

multi-stark/src/security.rs
```diff
@@ -119,20 +119,35 @@ impl MultiStarkSecurityReport {
...
         evidence: Option<PrescribedOpeningSecurity>,
     ) -> f64 {
+        // Evidence is usable only if every term is a finite non-negative bound.
+        // One unusable term makes the whole component unassessed, rather than shrinking it.
+        let usable = |evidence: &PrescribedOpeningSecurity| {
+            !evidence.terms.is_empty()
+                && evidence
+                    .terms
+                    .iter()
+                    .all(|term| term.bits.bits().is_finite() && term.bits.bits() >= 0.0)
```
Nashtare: This rejects any term with infinite bits, but `claim_pool_error` and `bit_ring_switch_error` return `+inf` on purpose when the error is zero (`k <= 1`, or no reductions). A Boolean opening with one column claim, such as one instance of a one-column AIR, therefore gets `main-pcs` marked unassessed, and `prove` then fails in `require_security` on a statement that is fine. The old check ran on the sum, where a zero-error term adds nothing. Could the per-term check reject only NaN and negative bits, and keep the finiteness check on `evidence.error()`?

sumcheck/src/ring_switch/bits/transcript.rs
```diff
@@ -0,0 +1,350 @@
...
+        // The point is bound first, so the element cannot be chosen to suit it.
+        self.state
+            .observe_extensions::<EF, EF, FieldToFieldCodec<EF>>(
+                EVALUATION_POINT,
+                point.as_slice(),
+            );
+        self.state
+            .observe_extensions::<EF, EF, FieldToFieldCodec<EF>>(TENSOR_ROWS, rows);
+
+        // Only now is the batching point drawn, so no message above it knew it.
```
Nashtare, line 198: The reduction's soundness rests on this draw coming after the rows. `reduction.rs` says the ordering test belongs to these two sides, but this file has no tests. If the `TENSOR_ROWS` step were dropped from the description and from both `statement` bodies, or if the draw were moved above it, every existing test would still pass. The round trips are symmetric and the tamper tests stop at `ClaimMismatch`. Could this get the tests the general transcript has (`a_perturbed_tensor_coefficient_moves_every_later_draw`, `a_perturbed_point_coordinate_moves_every_later_draw`), plus one pinning that `batching_point` is the first challenge and comes after `tensor_rows`?

binary-pcs/src/boolean.rs
```diff
@@ -0,0 +1,919 @@
...
+            .commit_bits(&witness(0xAAAA, LOG_BITS), &mut chal)
+            .unwrap();
+        let (values, proof) = pcs.open_at_points(data, &points, &mut chal).unwrap();
+
+        let (other, _) = pcs
+            .commit_bits(&witness(0xBBBB, LOG_BITS), &mut challenger())
+            .unwrap();
+
+        assert!(
+            pcs.verify_at_points(&other, &points, &values, &proof, &mut challenger())
```
Nashtare: This verifies with a fresh `challenger()` that never replays any root, while the prover's sponge absorbed witness A's root. The transcripts split at the first draw, so this rejects no matter what `other` is. No test reaches `BooleanPcsError::SurvivingClaim`, yet that is the only check tying the reductions to the commitment: deleting the fold at line 478 keeps every test green. A test that tells them apart: run the reductions on A's packing, open B's prover data at the surviving points on a sponge that absorbed B's root, verify with `replayed(&pcs, &b_root)`, and assert `SurvivingClaim`.

sumcheck/src/claim_pool.rs
```diff
@@ -0,0 +1,516 @@
...
+use p3_challenger::{CanObserve, CanSample};
+use p3_multilinear_util::point::Point;
+use p3_multilinear_util::poly::Poly;
+use thiserror::Error;
+
+/// Version byte bound into the seed, separating two revisions of this protocol.
+const VERSION: u8 = 1;
+
+/// Protocol name bound into the seed.
+const NAME: &[u8] = b"p3-sumcheck-claim-pool";
```
Nashtare: `p3-sumcheck-claim-pool` and `p3-sumcheck-bit-ring-switch` are new seeds in this crate, but neither is in `CRATE_PROTOCOL_NAMES` in `sumcheck/src/tests.rs`, so `no_two_protocols_in_this_crate_share_a_name` never sees them. `ClaimPoolShape` is generic over any `TranscriptField`, so it could also join `examples/tests/transcript_separation.rs` next to `ring_switch_cases`. The bit reduction cannot, for the same alphabet reason as `p3-binary-pcs`. Could both names go into the crate list, and the pool into the workspace suite?

multi-stark/src/security.rs
```diff
@@ -119,20 +119,35 @@ impl MultiStarkSecurityReport {
...
         match evidence {
-            Some(evidence)
-                if evidence.error.bits().is_finite()
-                    && evidence.error.bits() >= 0.0
-                    && evidence.log2_max_candidates.is_finite()
-                    && evidence.log2_max_candidates >= 0.0 =>
-            {
-                self.terms.push(SecurityTerm::new(label, evidence.error));
+            Some(evidence) if usable(&evidence) => {
+                self.terms.extend(evidence.terms);
```
Nashtare: Extending with the scheme's own labels loses which commitment a term belongs to. A WHIR config now reports `whir-opening` twice, and when one is short the report no longer says whether it is the main or the preprocessed opening. The old `main-pcs` / `preprocessed-pcs` labels gave that. The unassessed path still pushes the component label, so the two paths also disagree. Could the component survive, for example as a `component` field on the pushed term, since the label itself is a `&'static str`?

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -716,4 +981,149 @@ mod tests {
...
+        // Mutation: add one to a row, which moves the sum but not the claim.
+        let witness = bits(0x7A17, 32);
+        let packing = BitPacking::<EF>::new(&witness).unwrap();
+        let r = Point::<EF>::rand(&mut SmallRng::seed_from_u64(0x7A18), 8);
+        let reduction = BitRingSwitch::new(&r).unwrap();
+        let claim = embedded(&witness).eval_base(&r);
+
+        let (mut proof, _, _) = reduction.prove(&packing, &mut challenger());
+        let mut rows = proof.tensor.rows().to_vec();
+        rows[3] += EF::ONE;
```
Nashtare, line 1275: `EF::ONE` is `beta_0`, so this flips entry `(3, 0)`, which also changes column 0. The proof stops at `ClaimMismatch` and never reaches the sumcheck, so the comment above ("moves the sum but not the claim") does not describe what runs. The case the row/column tie exists for is a delta whose column reading at `r_low` is zero. One way to build it: add `c` to column 0 and `c * eq_low[0] / eq_low[1]` to column 1, then transpose back into rows. Could the test use that and assert `BitRingSwitchProofError::Sumcheck(_)`?

sumcheck/src/claim_pool.rs
```diff
@@ -0,0 +1,516 @@
...
+
+    /// Bind the protocol identity and this shape into a seed.
+    #[must_use]
+    pub fn domain_separator<EF: TranscriptField>(&self) -> DomainSeparator<Alphabet<EF>> {
+        DomainSeparator::new(VERSION, NAME, self.pattern::<EF>())
+    }
+}
+
+/// Claims about one committed polynomial, waiting for the draw that batches them.
+#[derive(Clone, Debug, PartialEq, Eq)]
```
Nashtare: `ClaimPool` derives `Clone`, so the order the module docs call unforgeable can still be broken: clone the pool, seal the copy to learn the draw, keep depositing into the original, then seal again. As a guard for honest callers that is fine, but lines 16-19 promise more than the type delivers. Either drop `Clone` (the tests can rebuild a pool from `claims()`), or say that the guarantee covers one pool value and not the transcript.

binary-pcs/src/pcs.rs
```diff
@@ -38,43 +55,76 @@ use crate::verifier::{
...
-    pub fn new(config: BinaryPcsConfig, mmcs: MT) -> Self {
+impl<F: EncodableLevel, EF, MT, MX> BinaryPcs<F, EF, MT, MX> {
+    /// Builds a PCS instance from a derived configuration and its two commitment schemes.
+    ///
+    /// # Arguments
+    ///
+    /// - `config`: the validated fold and query schedule both sides read.
+    /// - `mmcs`: commits the base codeword over the committed alphabet.
+    /// - `round_mmcs`: commits every folded codeword over the challenge field.
+    pub fn new(config: BinaryPcsConfig, mmcs: MT, round_mmcs: MX) -> Self {
```
Nashtare: `new` takes any `BinaryPcsConfig`, but the config only records `challenge_field_bits`, and nothing here checks it against `EF` or re-runs the alphabet checks for `F`. A config derived with `try_new::<BinaryField128, BinaryField128>(20, ..)` at `security_level: 100` builds a `BinaryPcs<BinaryField64, BinaryField64, ..>` whose `prescribed_security` reports about 100.9 bits, while `try_new::<BinaryField64, BinaryField64>` refuses the same parameters with `max: 39`. With `F = BinaryField8` the same config also skips the domain and grinding caps, and fails later as a prover panic. `BooleanPcs::new` and `BooleanTracePcs::new` inherit this. Could `new` check `config.challenge_field_bits() == EF::bits()` and the two `F` caps (returning a `BinaryPcsConfigError`), or could the config carry `PhantomData<(F, EF)>` so the type fixes the pairing?

binary-pcs/src/params.rs
```diff
@@ -147,10 +167,24 @@ impl BinaryPcsConfig {
...
+        // A longer codeword names a point the level does not hold.
+        if log_len > F::bits() {
+            return Err(BinaryPcsConfigError::CodewordExceedsAlphabetDomain {
+                log_len,
+                bits: F::bits(),
+            });
+        }
+
+        // The witness is an element of the committed alphabet, so its width caps the counter.
+        let max_pow_bits = F::bits().min(POW_COUNTER_BITS) - POW_HEADER_BITS;
```
Nashtare, line 239: `try_new` only bounds `F: TowerLevel`, which admits `Gf2`, `BinaryField2` and `BinaryField4`. For those, `F::bits().min(POW_COUNTER_BITS)` is below `POW_HEADER_BITS`, so this subtraction overflows. `try_new::<BinaryField4, BinaryField128>(2, ..)` panics with `attempt to subtract with overflow` in debug, and in release it wraps and accepts any `pow_bits`. No `BinaryPcs` can be built at those levels, because `EncodableLevel` stops at `BinaryField8`. Bounding `F: EncodableLevel` here, or using `checked_sub` and returning `PowBitsExceedWitnessCapacity`, would close it. Also, the doc on `POW_HEADER_BITS` says the counter binds at every level this crate reaches. For the 8/16/32-bit alphabets the field width is what binds now.

binary-pcs/src/params.rs
```diff
@@ -105,19 +125,19 @@ impl BinaryPcsConfig {
     /// if the polynomial has no variables, if grinding exceeds what the challenger can witness
     /// or the security budget, if the target exceeds what the field can deliver, or if the
     /// derived query count is zero.
```
Nashtare: The `# Errors` list doesn't include the two new variants, `CodewordExceedsAlphabetDomain` and `UnpricedChallengeField`. Could you add them? The domain cap is the one a caller at a narrow alphabet will hit first, since `num_variables + log_inv_rate` must stay at or below the alphabet's bit width (8 for `BinaryField8`).

binary-pcs/src/packing.rs
```diff
@@ -0,0 +1,735 @@
...
+
+/// A value whose bytes are a run of `F_2` coordinates, lowest coordinate first.
+///
+/// # Safety
+///
+/// An implementor is a transparent wrapper over an unsigned integer, or over a block of them.
+/// Every bit pattern of its bytes is a value, so a run has no padding and no invalid state.
+///
+/// On a little-endian target coordinate `j` of a run is bit `j % 8` of its byte `j / 8`.
+pub unsafe trait Coordinates: Copy + Send + Sync + 'static {
```
Nashtare, line 66: `pack` and `unpack` copy `len * size_of::<EF>()` bytes from a source of `cells.len() * size_of::<A>()` bytes. The two only match when `COORDINATES == 8 * size_of::<Self>()`. The safety section implies that but never says it, and the trait isn't sealed. So a downstream `unsafe impl` with more bytes than coordinates makes `pack::<BinaryField8, Wide>` read past the source. Could the contract say it outright? `pack`, `unpack` and `coordinate_bytes` could then carry the same `const { assert!(..) }` block `PackedGf2::as_bytes` uses, including its `cfg!(target_endian = "little")` check. On a big-endian target this reinterpretation would silently disagree with the `BasedVectorSpace` basis the tests pin it against. Elsewhere, `p3-binary-field` either asserts little-endian or falls back.

binary-pcs/src/fold.rs
```diff
@@ -85,17 +108,204 @@ const _: () = assert!(
...
+        folded = fold_round(&folded, beta);
+    }
+    folded
+}
+
+/// Lift the codeword with the batch's first challenge, then hand the rest to the wide field.
+///
+/// Only the first fold of a batch reads the narrow alphabet.
+/// Every later fold runs whichever route the wide field itself offers.
+fn fold_rounds_lifting<F, EF>(codeword: &[F], challenges: &[EF]) -> Vec<EF>
```
Nashtare, line 199: The lifting round materialises the whole lifted codeword in the tower basis before the packed route runs. It's also the slow arm in the description's table (36.8 ms vs 8.9 ms). The packed driver already copies each block into scratch as `symbol.to_repr()` and converts it with `from_tower_slice`. Embedding a narrower level into `BinaryField128` is a zero-extension of the repr. Could `fold_rounds` take `&[F]` and extend each block with `u128::from(symbol.to_repr())`? Then the alphabets under `BinaryField128` would take the packed route for every round, the first one included, with no lifted copy. The domain points already agree across levels, which `a_narrow_domain_point_embeds_into_the_wide_one` pins.

binary-pcs/src/fold.rs
```diff
@@ -85,17 +108,204 @@ const _: () = assert!(
...
+/// ```
+///
+/// Entry `k` is the sum of the basis vectors those bits select, one place up for the doubling:
+///
+/// ```text
+///     steps[k] = sum_{r <= k} v_{r + 1}
+/// ```
+///
+/// A walk over the symbols therefore evaluates the domain once and adds one entry per step.
+fn pair_steps<F: TowerLevel>(num_pairs: usize) -> Vec<F> {
```
Nashtare: `pair_steps` builds the same table as `domain_point_steps` in `binary-dft/src/domain.rs`, with the same recurrence `steps[k] = sum_{r <= k} v_{r+1}`. That helper is `pub(crate)` there. Could it be made public and reused here, so the walk identity lives in one place? The only difference is that this one derives the count from `num_pairs`.

binary-pcs/tests/end_to_end.rs
```diff
@@ -323,6 +336,96 @@ const fn challenger() -> MyChallenger {
...
+        0,
+    )
+}
+
+const fn narrow_challenger() -> NarrowChallenger {
+    NarrowChallenger::from_hasher(Vec::new(), Keccak256Hash)
+}
+
+#[test]
+fn a_narrow_committed_alphabet_round_trips_with_wide_challenges() {
```
Nashtare: This is the only end-to-end run over a narrow alphabet. It covers `BinaryField64` into `BinaryField128` at folding factor 1 with no grinding. CI never proves and verifies with the `BinaryField8`/`16`/`32` encoders, a `BinaryField64` challenge field, batched folding across the lift, or a narrow grinding witness. I ran those locally at the head commit: 8/16/32/64 into 64 and 128, folding 1 to 4, `pow_bits` up to the alphabet cap. All of them round-trip and reject a tampered base symbol, so this is a coverage gap rather than a bug. Could this test be parametrised over those pairs, even at small arities?

binary-field/src/challenger.rs
```diff
@@ -96,28 +96,28 @@ impl<F: TowerLevel, Inner: CanObserve<u8>> CanObserve<F> for BinaryChallenger<F,
     }
 }
 
-impl<F, const N: usize, Inner: CanObserve<u8>> CanObserve<Hash<F, u8, N>>
+impl<F, G, const N: usize, Inner: CanObserve<u8>> CanObserve<Hash<G, u8, N>>
```
Nashtare, line 99: These `u8` digest and cap impls now accept a digest tagged with any field `G`. The `u64` word-digest impls just below are still tied to the challenger's own `F`. A narrow-alphabet run with a `u64`-word Merkle hasher couldn't observe its round commitments over the challenge field. Nothing in the workspace hits this today. Would you widen those three impls the same way, for consistency?

binary-pcs/src/boolean_trace.rs
```diff
@@ -0,0 +1,767 @@
...
+                    table,
+                    expected: shapes[table].num_variables(),
+                    actual: point.num_variables(),
+                });
+            }
+            let placement = by_table[table].expect("the planner places every supplied shape");
+
+            // Slot address as the leading coordinates, the row point as the trailing ones.
+            for &column in batch.current() {
+                lifted.push(placement.selectors()[column].lift_prefix(point));
```
Nashtare, line 192: Every opened column becomes its own bit ring switch over the whole stacked witness. `BitRingSwitch::prove` builds `eq_high` twice, clones the packing and runs an `n - 7` round sumcheck over all `2^(n-7)` elements, and each proof carries a 128-row tensor (2 KiB). For `W` columns of `2^a` rows that is `W^2 * 2^a / 128` field work, which matches the embedded arm's opening sumcheck at roughly `W = 128`, plus about `W * 2.5` KB of proof (12.6 KB of the 156,902 bytes at 2^18 here). The slot prefix is Boolean, so `eq_high` is zero outside the slot. Could the reduction run on the slot's `2^(a-7)` elements and return `(bits of s, r')`, or batch a table's columns, which share `r`, into one reduction?

binary-pcs/src/boolean_trace.rs
```diff
@@ -0,0 +1,767 @@
...
+
+    fn verify(
+        &self,
+        commitment: &Self::Commitment,
+        proof: &Self::Proof,
+        challenger: &mut Challenger,
+        protocol: Self::OpeningProtocol,
+    ) -> Result<(), Self::Error> {
+        // The same draws the prover made, in the same order, before anything is checked.
+        let points = sample_points(&protocol, challenger);
```
Nashtare, line 496: `BinaryPcs::verify` and the WHIR adapter's `verify` both call `observe_commitment` first, because their `commit` bound the root on the prover side. This one does not, so a generic sampled-path caller that works for those two fails here: an honest proof checked with a fresh challenger, as `BinaryPcs` expects, is rejected. `the_sampled_path_draws_the_same_points_on_both_sides` binds by hand before calling `verify`, which would double-bind on `BinaryPcs`. Could this call `self.observe_commitment(commitment, challenger)` before sampling, leaving `verify_at` as the caller-binds path, as it is on `BinaryPcs`?

binary-pcs/src/boolean_trace.rs
```diff
@@ -0,0 +1,767 @@
...
+                };
+
+                if column_len >= WORD_BITS {
+                    // The slot is word aligned and fills whole words, so none is read back.
+                    let (runs, rest) = cells.as_chunks::<WORD_BITS>();
+                    debug_assert!(rest.is_empty(), "a power-of-two column fills whole words");
+                    for (word, run) in words[offset / WORD_BITS..].iter_mut().zip(runs) {
+                        *word = pack_word(run).ok_or_else(refuse)?;
+                    }
+                } else {
```
Nashtare, line 248: No test reaches this branch: every table here and in the example has at least `2^7` rows. The placement looks right, and a multi-stark batch of heights `[9, 3, 2]` round trips, but a table under 64 rows in `an_opened_column_is_the_column_the_table_holds` would pin the in-place bit placement against the reference column, which is the only check that catches a wrong shift here.

binary-pcs/src/boolean.rs
```diff
@@ -0,0 +1,919 @@
...
+        }
+        Ok(())
+    }
+
+    /// The packed multilinear the commitment holds, read back out of the retained table.
+    fn packing(
+        prover_data: &BinaryPcsProverData<EF, EF, MT>,
+    ) -> Result<BitPacking<EF>, BooleanPcsError<EF, MT::Error>> {
+        let packed = Poly::new(prover_data.table(0).poly(0).as_slice().to_vec());
+        BitPacking::from_packed(packed).map_err(|_| BooleanPcsError::WitnessTooNarrow {
```
Nashtare: `from_packed` fails only for a level narrower than a byte or a table whose length is not a power of two. Neither is a witness of zero variables, which is what `WitnessTooNarrow { needed: 7, actual: 0 }` reports. The first is a property of `EF` and could be refused in `new`; the second cannot happen for a table `commit_bits` built. Could this be a dedicated variant, or an `expect` naming the invariant? Separately, the `.filter(|_| num_variables >= absorbed)` in `packed_variables` is redundant after `checked_sub`.

multi-stark/examples/prove_boolean_trace.rs
```diff
@@ -0,0 +1,709 @@
...
+                &pk,
+                &public,
+            )]),
+            0,
+            SECURITY_BITS,
+            &mut challenger(),
+        )
+        .unwrap();
+
+        // Mutation: flip each pinned input in turn; each one breaks a boundary constraint.
```
Nashtare: Public values are absorbed before the zerocheck, so a tampered one moves every later challenge and the proof fails whatever the AIR says. With the three `when_first_row` constraints deleted this test still passes, failing at `Opening(Boolean(ReductionProof(ClaimMismatch)))`. Proving and verifying against the same tampered publics does reach the boundary check, as `Zerocheck(FinalSumMismatch)`. Could the test do that and match on the variant? `rejects_an_invalid_gate_table` only asserts `is_err()` too; it fails with `Zerocheck(FinalSumMismatch)`, and pinning that stops an opening-side rejection from passing it.

multi-stark/examples/prove_boolean_trace.rs
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    fn num_public_values(&self) -> usize {
+        // The three inputs of the first row, pinned as the public boundary.
+        3
+    }
+
+    fn main_next_row_columns(&self) -> Vec<usize> {
+        // No constraint reads a row ahead, so no column is opened through that view.
+        // The default names every column, which would ask for an opening none of them needs.
```
Nashtare, line 158: Every AIR on the Boolean arm has to return `[]` here, since the default names every column and `BooleanTracePcs` refuses any successor view. Multi-stark zero-fills undeclared next-row columns (`TableOpening::next_row`), so an AIR that reads `main.next_slice()` but declares nothing is proved and verified against zeros instead of being refused. The check in `opening_points` only sees what the AIR declares. Could setup compare this list against the next-row reads in the symbolic constraints, or could the `BooleanTracePcs` docs say that its refusal rests on this declaration being complete?

binary-pcs/examples/peak_heap.rs
```diff
@@ -0,0 +1,218 @@
...
+        let table = Table::new(RowMajorMatrix::new(cells.clone(), cells.len()));
+        let plain_witness = SuffixProver::<EF, EF>::new_witness(vec![table], 0);
+        let (_, plain_data) = plain.commit(plain_witness, &mut challenger()).unwrap();
+        let commit_embedded = disarm();
+
+        row("commit", log_bits, commit_embedded, commit_packed);
+
+        // Open, packed: a ring switch, then one opening of the packed commitment.
+        let points = vec![point.clone()];
+        arm();
```
Nashtare: `arm()` resets the counter while `data`, allocated before it, is still live. So the `open` row leaves out the codeword and tree the opening holds throughout, and any free of `data` while armed pushes `LIVE` down and hides a later allocation of the same size. The commit rows charge the embedded witness clone but not `bits`. The ratios probably hold, but these figures answer the memory question on this PR. Could this arm before building the inputs, as `prove_boolean_trace.rs` does, or print the figure as net new allocation rather than peak heap?

binary-pcs/tests/boolean.rs
```diff
@@ -0,0 +1,252 @@
...
+    //
+    //     - a cap mostly widens the committed words it covers
+    //     - a grouped leaf mostly replaces paths the frontier had already merged
+    //
+    // Neither moves this proof by much, in either direction.
+    for log_folding_factor in [1usize, 3] {
+        let config = config(LOG_TALL_BITS, log_folding_factor);
+        let baseline = tall_opening(config, mmcs(0), mmcs(0), "baseline");
+
+        for cap_height in [0usize, 4] {
```
Nashtare, line 250: With `cap_height = 0`, `plain` is the same configuration as `baseline`, so each arity pays one more 2^17 commit, open and verify that can only reproduce the baseline size. Running the plain cell at cap 4 only, and keeping the grouped cell at both caps, covers the same grid with two fewer round trips in CI.

binary-pcs/src/boolean.rs
tcoratger: Fixed. The Boolean verifier now compares every returned opening directly with its surviving reduction value, so this adds zero soundness error. The claim-pool layer and its feature/dependency were removed; the saving comes from opening all surviving points in one `try_open_at` call.

multi-stark/src/security.rs
tcoratger: Fixed. Per-term validation now rejects only NaN or negative bit counts; `+inf` remains the correct representation of a zero-error term. The aggregate error is still required to be finite before the target comparison.

sumcheck/src/ring_switch/bits/transcript.rs
tcoratger: Added the transcript-ordering tests here: perturbing either a tensor coefficient or point coordinate changes every later draw, and the pinned event trace proves `batching_point` is the first challenge after all tensor rows.

binary-pcs/src/boolean.rs
tcoratger: Added the requested end-to-end adversarial test. It runs A's reductions, opens B at those surviving points under B's bound root, replays B's transcript correctly, and reaches the precise `BooleanPcsError::SurvivingClaim` rejection.

sumcheck/src/claim_pool.rs
tcoratger: Addressed after simplifying the protocol: the unused claim-pool protocol was removed entirely. The bit-ring-switch protocol name is now included in the sumcheck crate's uniqueness list.

multi-stark/src/security.rs
tcoratger: Fixed. Each `SecurityTerm` now retains both the component (`main-pcs` / `preprocessed-pcs`) and the scheme-local label, so repeated scheme labels remain attributable and assessed/unassessed reports use the same structure.

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: Fixed with a column-preserving delta constructed from the equality weights. The test now keeps the incoming claim unchanged and asserts the precise `BitRingSwitchProofError::Sumcheck(_)` failure.

sumcheck/src/claim_pool.rs
tcoratger: Resolved by removing the claim-pool mechanism altogether; the Boolean verifier now performs exact pointwise comparisons.

binary-pcs/src/pcs.rs
tcoratger: Fixed. Construction now rechecks the config against the concrete `(F, EF)` pair, including challenge width, codeword-domain capacity, and grinding capacity. Regression tests cover configs derived for a different pair.

binary-pcs/src/params.rs
tcoratger: Fixed by restricting configuration derivation to `EncodableLevel`, matching the levels a `BinaryPcs` can actually instantiate. The witness-capacity documentation now describes the narrow-field cap accurately, with regression coverage for the former underflow case.

binary-pcs/src/params.rs
tcoratger: Added both missing error cases to the `# Errors` documentation, including the narrow-alphabet domain limit.

binary-pcs/src/packing.rs
tcoratger: Fixed. The unsafe trait contract now states the exact coordinate/byte-size equality, and `pack`, `unpack`, and `coordinate_bytes` enforce compile-time size and little-endian assertions at each reinterpretation boundary.

binary-pcs/src/fold.rs
tcoratger: Implemented. Every alphabet targeting `BinaryField128` now enters the packed route directly; narrow symbols are zero-extended into each scratch block, so the first fold no longer materializes a full lifted codeword.

binary-pcs/src/fold.rs
tcoratger: Fixed. The shared helper is public and the PCS fold walk now reuses `domain_point_steps`, leaving the recurrence in one implementation.

binary-pcs/tests/end_to_end.rs
tcoratger: Expanded the end-to-end matrix across 8/16/32/64-bit committed alphabets, 64/128-bit challenge fields, folding factors 1–4, and capped narrow-field grinding. Each configuration round-trips and checks base-symbol tampering.

binary-field/src/challenger.rs
tcoratger: Fixed all three `u64` digest/cap observation impls to accept an independently tagged field `G`, matching the `u8` behavior and narrow-alphabet round commitments.

binary-pcs/src/boolean_trace.rs
tcoratger: Implemented the slot restriction fully. A Boolean fixed prefix selects one contiguous packed slot; prover and verifier run only the remaining rounds, restore the prefix in the surviving point, and tests cover every address/prefix. Work is now linear in the total trace size rather than quadratic in column count.

binary-pcs/src/boolean_trace.rs
tcoratger: Fixed. The sampled `verify` path now observes the commitment before drawing points; `verify_at` remains the prescribed-point path where the caller owns commitment binding. Tests pin both transcript conventions.

binary-pcs/src/boolean_trace.rs
tcoratger: Added explicit sub-64-row coverage that opens a short table column and compares the in-place packed placement with the reference column.

binary-pcs/src/boolean.rs
tcoratger: Fixed by making this an internal invariant with an explanatory `expect`; the misleading `WitnessTooNarrow` mapping is gone. The redundant filter after `checked_sub` was removed too.

multi-stark/examples/prove_boolean_trace.rs
tcoratger: Both tests now reach and match the intended zerocheck failure. The public-boundary case proves and verifies with the same tampered publics and asserts `Zerocheck(FinalSumMismatch)`; the invalid-gate case pins that same variant.

multi-stark/examples/prove_boolean_trace.rs
tcoratger: Documented the required completeness contract at the Boolean trace PCS boundary: every successor read must be declared, because undeclared columns are zero-filled by the multi-STARK layer. The example declaration now calls this out at the implementation site.

binary-pcs/examples/peak_heap.rs
tcoratger: Fixed the measurement lifetime. The counter is armed before constructing the measured inputs, retained prover data remains charged during opening, and the output/documentation now describes peak live heap rather than a misleading reset-relative figure.

binary-pcs/tests/boolean.rs
tcoratger: Removed the duplicate plain `cap_height = 0` round trips. Plain runs only at cap 4; grouped coverage retains both caps.

sumcheck/src/lib.rs
```diff
@@ -26,6 +26,7 @@
 
 extern crate alloc;
 
+pub mod claim_pool;
```
Nashtare: The replies on the pool threads say it was removed, but it still ships. That is `claim_pool.rs` (+530 lines) here, `claim_pool_error`/`claim_pool_term`/`CLAIM_POOL_LABEL` in `p3-security` (+165 lines), and registrations in `CRATE_PROTOCOL_NAMES` and `transcript_separation.rs` (`NUM_PROTOCOLS` 25 to 26). Nothing outside its own tests calls it, and the module docs describe a path the workspace no longer takes. Could it go, along with the `"claim-pool-batching"` check in `prove_boolean_trace.rs` and the `+inf` comment that cites it in `multi-stark/src/security.rs`? If it is meant for jagged stacking, it could land with that caller.

sumcheck/src/claim_pool.rs
```diff
@@ -0,0 +1,530 @@
...
+//! The challenge must be drawn after every claim value is bound.
+//! Drawing it first lets a prover pick its claims knowing the weights.
+//!
+//! One linear equation in the claim values is then always solvable.
+//!
+//! Three things make that order unforgeable here:
+//!
+//! - the challenge lives only on the sealed pool, which sealing alone produces
+//! - sealing consumes the open pool, so no claim can be added once it exists
+//! - an open pool cannot be duplicated, so sealing a copy cannot reveal the draw early
```
Nashtare: Dropping `Clone` doesn't make this true: `new` + `claims()` + `deposit` rebuilds an open pool, which is what `rebuilt` below does. Sealing that copy on a cloned challenger reveals the draw while the original still takes deposits. Moot if the module goes. Otherwise, could this say the guarantee covers one pool value, not the transcript?

binary-pcs/src/pcs.rs
```diff
@@ -30,51 +46,105 @@ use p3_util::log2_ceil_usize;
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error unless the schedule was derived for these two levels.
+    pub fn new(
+        config: BinaryPcsConfig,
+        mmcs: MT,
+        round_mmcs: MX,
+    ) -> Result<Self, BinaryPcsConfigError> {
+        config.check_alphabets::<F, EF>()?;
```
Nashtare, line 111: This closes the mismatch, but no test reaches it: nothing outside `params.rs` names `CommittedFieldMismatch` or `ChallengeFieldMismatch`, and deleting this line keeps the suite green. Could the round-1 case become a test? That is a `(128, 128)` config at 20 variables and security 100 handed to `BinaryPcs<BinaryField64, BinaryField64, ..>`, and a `(16, 128)` config handed to a `BinaryField8` alphabet. `BooleanPcs::new`'s `Config(..)` path could get one too.

binary-pcs/src/packing.rs
```diff
@@ -0,0 +1,769 @@
...
+/// A value carrying more bytes than that would read past the source of a copy.
+/// On a little-endian target coordinate `j` of a run is bit `j % 8` of its byte `j / 8`.
+pub unsafe trait Coordinates: Copy + Send + Sync + 'static {
+    /// Coordinates one value of this type holds.
+    const COORDINATES: usize;
+}
+
+// A byte index is a memory offset and a coordinate is a bit of a value.
+// The two orders agree only on a little-endian target, which every copy below rests on.
+const _: () = assert!(
```
Nashtare: At module level this makes all of `p3-binary-pcs` refuse to build on a big-endian target, including a BF128 PCS that never packs. `cargo +nightly check -Zbuild-std --target powerpc64-unknown-linux-gnu -p p3-binary-pcs` passes on main and fails here with E0080. `p3-binary-field` and `p3-binary-dft` put this check inside the function that reinterprets bytes. Could it move into the three `const {}` blocks in `pack`, `unpack` and `coordinate_bytes`?

sumcheck/src/ring_switch/bits/transcript.rs
```diff
@@ -0,0 +1,748 @@
...
+            assert_ne!(
+                baseline,
+                replay(&point, &tampered, EF::ONE),
+                "tampering with row {position} left the stream where it was",
+            );
+        }
+    }
+
+    #[test]
+    fn the_batching_draw_is_the_first_challenge_and_answers_to_the_rows() {
```
Nashtare, line 610: This pins the draw after the rows. Nothing pins the evaluation point before the draw, though. If both sides move `evaluation_point` after it, all `bits` tests still pass. `BooleanPcs` documents that callers owe no point binding, and this ordering is what makes that hold. Could this also assert that `evaluation_point` comes before the first challenge?

binary-pcs/src/boolean.rs
```diff
@@ -0,0 +1,968 @@
...
+                &false_values,
+                &proof,
+                &mut replayed(&pcs, &commitment),
+            )
+            .unwrap_err();
+        assert!(matches!(err, BooleanPcsError::ReductionProof(_)), "{err:?}");
+    }
+
+    #[test]
+    fn a_reduction_over_another_witness_leaves_a_claim_the_commitment_does_not_open() {
```
Nashtare: This is the only test that reaches the comparison, and it opens one point. With `.take(1)` in the loop at line 470, the whole `p3-binary-pcs` suite still passes. Could this run with two or three points and forge the claim at a non-first index?

sumcheck/src/ring_switch/bits/transcript.rs
```diff
@@ -0,0 +1,748 @@
...
+}
+
+impl BitRingSwitchShape {
+    /// Describe a reduction of a claim at a point of this width.
+    #[must_use]
+    pub const fn new(num_variables: usize) -> Self {
+        Self { num_variables }
+    }
+
+    /// Rounds the delegated sumcheck runs, panicking on a point narrower than one element.
```
Nashtare: Since the slot restriction this is the maximum. A Boolean prefix of length `p` runs `l' - p` rounds.

binary-pcs/src/fold.rs
```diff
@@ -85,17 +115,200 @@ const _: () = assert!(
...
+/// Nothing else in this crate names a field.
+///
+/// # Panics
+///
+/// Every implementation panics unless there is at least one challenge.
+/// A fold with no round would leave every output slot at zero rather than fail.
+pub trait FoldAlphabet<EF: TowerLevel>: TowerLevel {
+    /// The arithmetic representation used by the residual sumcheck.
+    ///
+    /// This belongs to the `(alphabet, challenge field)` route: 128-bit challenges use the
```
Nashtare: This depends only on `EF`: every BF128 impl picks `Ghash128` and every BF64 impl picks `BinaryField64`. Would it read better keyed on the challenge field alone, instead of repeated nine times? Separately, `Poly64` is BF64's carryless-multiply counterpart, if the identity repr's table copy through the default `from_table` ever shows up.

binary-pcs/src/prover.rs
```diff
@@ -165,7 +187,11 @@ where
     // The rounds multiply in the polynomial basis, which needs no change of basis per product.
```
Nashtare: With `EF = BinaryField64` the rounds run in the tower field itself, so this comment only holds for the BF128 route.

binary-pcs/tests/end_to_end.rs
```diff
@@ -323,6 +349,187 @@ const fn challenger() -> MyChallenger {
...
+}
+
+#[test]
+fn every_committed_alphabet_round_trips_under_every_fold_schedule() {
+    // Invariant: the committed alphabet and the challenge field are separate choices.
+    //
+    // The pairs below are exactly the ones the fold route table admits.
+    // A challenge field must fold into itself, which only the two widest levels do.
+    //
+    //     - 8, 16, 32   ->  64 and 128, each by lifting on the batch's first fold
```
Nashtare: The narrow alphabets now take the packed route into 128, and lift only into 64.

examples/src/binary.rs
```diff
@@ -91,15 +91,16 @@ pub fn binary_config<const N: usize>(
...
+        BinaryPcsConfig::try_new_with_folding::<F, F>(arity, params, folding.min(arity))?;
     let merkle = MerkleMmcs::<N>::new(
         Hash::new(Keccak256Hash),
         Compress::<N>::new(Keccak256Hash),
         0,
     );
     let mmcs = Mmcs::<N>::for_folding(merkle, &pcs_config);
     Ok(BinaryStarkConfig {
-        pcs: BinaryPcs::new(pcs_config, mmcs),
+        pcs: BinaryPcs::new(pcs_config, mmcs.clone(), mmcs).unwrap(),
```
Nashtare: The function already returns `BinaryPcsConfigError`, so `?` fits here.

binary-pcs/src/boolean.rs
```diff
@@ -0,0 +1,968 @@
...
+where
+    EF: EncodableLevel + TranscriptField + TowerLevel + FoldAlphabet<EF> + Coordinates,
+    MT: Mmcs<EF>,
+    MX: Mmcs<EF, Error = MT::Error>,
+{
+    /// Build a Boolean commitment over a bit witness of `num_variables` variables.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error unless the schedule commits exactly the elements the packing holds.
```
Nashtare, line 213: `new` now also refuses a config derived for another `(F, EF)`, through `check_alphabets`. The same line in `BooleanTracePcs::new` could say so too.

security/src/report.rs
```diff
@@ -44,15 +44,38 @@ pub const BATCH_LABEL: &str = "batch-combination";
...
+/// The label names the error source, which the crate charging it chooses.
+///
+/// A protocol composing two instances of one scheme sees that label twice.
+/// Naming the instance is what the component is for.
 #[derive(Copy, Clone, Debug, PartialEq, Serialize)]
 pub struct SecurityTerm {
+    /// Error source this term charges.
     pub label: &'static str,
+    /// Which of the composing protocol's parts charged it, when there is a choice.
+    pub component: Option<&'static str>,
```
Nashtare, line 58: Every FRI, STIR and stark term now serializes with `"component": null`. `#[serde(skip_serializing_if = "Option::is_none")]` would keep their shape unchanged.

Nashtare (approved):
Re-reviewed at b2aade27. All the round-1 threads are addressed, and I found no soundness issue in the new code.

- **Slot-restricted ring switch.** It is exact: a Boolean coordinate contributes an identity factor, and the verifier derives the round count from the public point. A padded or a short round list fails with `RoundCountMismatch`, and the zero-round case still binds through the final check.
- **Config check.** It refuses every mismatch from round 1.
- **Zero-extended fold.** It matches the subfield embedding (exhaustive for BF8/BF16, sampled for BF32/BF64).

fmt, clippy and the touched crates' tests pass locally, serial and `parallel`.

Approving, with two asks before merge:
- The claim pool still ships (`sumcheck/src/claim_pool.rs`, and `claim_pool_term` in `p3-security`) with no caller. Could it go, or land with the PR that uses it?
- The description still presents the pool (§2.7). The peak-heap table also predates the harness fix: at this head on an M4 I get 121-122x for commit and 77-114x for commit and open.

Two of my round-1 comments were wrong; I've replied on those threads.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: My round-1 ask was wrong. A round sends `h(0)` and `h(∞)` and derives `h(1)`, so a wrong starting sum can't fail a round and surfaces at the final check. The test's `FinalCheck` is the right variant. Thanks.

multi-stark/examples/prove_boolean_trace.rs
Nashtare: My premise here was wrong. `validate_successor_columns` already refuses an undeclared successor read at setup, on both sides, so nothing is zero-filled. The new test pins that on the Boolean arm, which is enough.

sumcheck/src/ring_switch/bits/transcript.rs
tcoratger: Added a transcript-description test that locates the evaluation point and asserts it is absorbed before the first challenge.

binary-pcs/src/pcs.rs
tcoratger: Added constructor tests for committed-field and challenge-field mismatches, including the requested schedule derived for (128, 128) and used as (64, 64).

binary-pcs/src/packing.rs
tcoratger: Moved the little-endian assertions into pack, unpack, and coordinate_bytes. The crate now compiles on powerpc64; only instantiating the layout-dependent operations is rejected.

sumcheck/src/lib.rs
tcoratger: Removed the claim-pool implementation and all of its exports, transcript registrations, security terms, examples, and tests.

binary-pcs/tests/end_to_end.rs
tcoratger: Corrected the route documentation: narrow alphabets lift into BF64, while BF128 uses the packed route.

binary-pcs/src/boolean.rs
tcoratger: Expanded the forgery regression to three points, corrupting only the middle surviving claim while asserting the first and last claims agree.

security/src/report.rs
tcoratger: Added skip_serializing_if = Option::is_none to SecurityTerm::component.

binary-pcs/src/boolean.rs
tcoratger: Documented on both BooleanPcs::new and BooleanTracePcs::new that the schedule must be derived for (EF, EF), and added a mismatched-field regression.

binary-pcs/src/fold.rs
tcoratger: Keyed the residual representation on the challenge field alone: BF128 uses Ghash128 and BF64 uses Poly64. Added the required narrow-field algebra implementations and an in-place parallel BF64 tower-to-polynomial table conversion.

binary-pcs/src/prover.rs
tcoratger: Updated the comment to state that each challenge width uses its carryless-multiply representation; BF64 now uses Poly64 rather than native tower arithmetic.

sumcheck/src/ring_switch/bits/transcript.rs
tcoratger: Clarified that sumcheck_rounds returns the maximum l-prime, and that a Boolean prefix of length p leaves l-prime minus p rounds.

examples/src/binary.rs
tcoratger: The rebased example now propagates BinaryPcs::with_ntt with ? and carries the required AdditiveNtt<F> + Sync bound.

sumcheck/src/claim_pool.rs
tcoratger: Resolved by removing the claim pool entirely, so the incorrect uniqueness claim no longer exists.

PR #2173, opened 2026-09-17
perf(binary-dft): exploit subfield structure in the additive transform and the encoder

tcoratger:
The transforms-and-encoding epic, finished. Three items ship code; the other two are closed without any, one on measurement and one because its premise was wrong.

Every number below was measured on this branch on an AMD Ryzen 9 9950X3D (Zen 5, 32 threads, `avx512f`/`avx512bw`/`gfni`/`vpclmulqdq`), with `--features parallel`, in two legs: **portable** is a plain `cargo bench` with no `RUSTFLAGS`, so there is no carryless multiply and no byte-map instruction; **native** is `-C target-cpu=native`, so both exist. The noise floor on unrelated arms is about ±5% on this machine, so gains below that are not claimed.

Two tables below were re-measured during review, after a benchmark fix described in their sections. Those re-runs were taken while the same machine was running other 32-thread benchmark jobs, at a load average of 30 on 32 threads. Per-arm confidence intervals stay tight at 60 samples, but the ambient load moves between arms, so a ratio can shift without any arm looking noisy: two runs of the portable subfield leg gave 1.62x and 1.19x for the same comparison. The native ratios reproduced across runs to within 2% and are quoted; the portable ones did not and are marked as needing a quiet host.

## Type the two- and four-byte subfield twiddles

A butterfly whose twiddle lies in a byte-aligned tower subfield scales each coordinate of an element on its own, which the one-byte case already did. The two- and four-byte cases fell through to a full-width product, because the narrow levels have no product against a subfield as wide as themselves and so cannot carry one generic bound. A chain of three generic scaling functions, one per subfield width, plus one forwarding impl per level, gives each level exactly the typed products it has — no macros.

Forward transform, width 16, over the subspace itself:

| arm | portable before | portable after | gain | native before | native after | gain |
| --- | --- | --- | --- | --- | --- | --- |
| `ntt/128/tower` 2^18 | 97.1 ms | 39.6 ms | 2.45x | 3.90 ms | 3.89 ms | 1.00x |
| `ntt/128/tower` 2^20 | 461.9 ms | 189.0 ms | 2.44x | 31.50 ms | 31.35 ms | 1.00x |
| `ntt/64` 2^18 | 33.4 ms | 15.3 ms | 2.19x | 1.563 ms | 1.603 ms | 0.98x |
| `ntt/32` 2^18 | 11.9 ms | 7.87 ms | 1.51x | 0.801 ms | 0.733 ms | 1.09x |
| `ntt/128/tower/narrow` 2^20 | 30.46 ms | 12.56 ms | 2.43x | 2.11 ms | 2.07 ms | 1.02x |

The kernel alone, 1024 elements per side, by twiddle width:

| arm | portable before | portable after | gain | native before | native after | gain |
| --- | --- | --- | --- | --- | --- | --- |
| `128/tower` two-byte twiddle | 80.18 us | 26.14 us | 3.07x | 285.3 ns | 259.6 ns | 1.10x |
| `128/tower` four-byte twiddle | 80.28 us | 31.49 us | 2.55x | 469.9 ns | 449.0 ns | 1.05x |
| `32/short` two-byte twiddle | 25.05 ns | 14.40 ns | 1.74x | 27.25 ns | 14.09 ns | 1.93x |

**The prior 3-10x estimate does not hold: the measured range is 1.5x to 3.1x**, on the portable leg only. The native leg is flat because the byte-map kernel already covers those twiddle widths there, apart from the four-byte level, which has no byte map to fall back on.

One measured negative inside this item. With a hardware carryless multiply the widest two levels price their own full-width product *below* any multi-coordinate expansion, and an ungated typed path cost up to +89% on the short runs. The dispatch therefore takes the wide typed products only where that instruction is absent — a compile-time constant, so one arm survives compilation.

Review moved the *product* half of that crossover into the field crate, where it belongs: the mixed-product gate in `extension.rs` covered only the sixteen-byte level, so the eight-byte one still expanded into coordinates on a carryless-multiply target. A new `subfield_mul` benchmark measures it, and the gate now covers both.

| pair | coordinate expansion | embedded wide product | gain |
| --- | --- | --- | --- |
| `BinaryField64 * BinaryField32` | 14.97 us | 6.045 us | 2.48x |
| `BinaryField64 * BinaryField16` | 6.430 us | 6.189 us | 1.04x |

The transform keeps its own gate, because it is a different decision: it chooses between a scalar element-at-a-time loop and the packed full-width kernel, not between two products. Forcing the typed path on with the field gate already widened costs `butterfly/64` at a full-width twiddle +77.6% and the whole coset transform 15.7% to 27.8% across the four heights, native.

## Run a subfield message's closed layers at its own width

At commit time a trace is bit- or byte-valued, and the widest layers of an unshifted transform scale by twiddles from the message's own subfield, so they map the subfield into itself. Splitting a row index into its top and bottom halves turns the network into two whole transforms: the closed layers are the entire transform of the buffer read as a short wide matrix at the narrow level, and the rest is the stage range the buffer really is. Both are unshifted and share their block indices, so they agree twiddle for twiddle.

Review found the benchmark's reference widening on one thread while the specialised route widens across the machine above 4 MiB, so the *before* column was charged first-touch page faults on a 64/256 MiB buffer that the *after* column never paid. The reference now widens in parallel and the table is re-measured. On the native leg that alone takes `wide/lch` from 15.18 to 10.09 ms at 2^18 and from 78.80 to 58.23 ms at 2^20, so the gains here are much smaller than first reported.

Byte message to a `BinaryField128` codeword, width 16, against the better of the two unspecialised routes (widen-then-transform through either backend):

| height | portable before | portable after | gain | native before | native after | gain |
| --- | --- | --- | --- | --- | --- | --- |
| 2^14 | not re-measured | not re-measured | 1.19-1.62x | 317.9 us | 269.2 us | 1.18x |
| 2^16 | not re-measured | not re-measured | 0.90-1.25x | 811.6 us | 690.8 us | 1.17x |
| 2^18 | not re-measured | not re-measured | 0.98-1.23x | 10.09 ms | 6.56 ms | 1.54x |
| 2^20 | not re-measured | not re-measured | 0.92-0.97x | 58.23 ms | 38.01 ms | 1.53x |

The native columns are one run, with a second run agreeing on every ratio to within 2%. The portable gains are the spread across two 60-sample runs taken under the contention described at the top, and are not quoted as point figures because they did not reproduce; at 2^20 both runs put the split at or below parity, so the portable leg needs a re-run on a quiet host before any portable gain is claimed for this item.

At width 1 the native win is 1.05x to 1.59x, largest at the smallest height, because a narrow row leaves the split least to save once the domain is big. The threshold below which no layer runs narrow is stated as a constant rather than left implicit: a message inside a private cache saves nothing on those layers and pays a thread handoff per layer for the privilege.

**Which route this is a win over.** Phase 2 runs the layers the subfield is not closed under through this crate's tower transform, which at 2^20 is 12 of the 20 layers. On a target where the polynomial-basis backend is substantially faster at the wide element size — AArch64 with `aes`, where `AdditiveRsEncoder` picks it — widening first through *that* backend can beat the split, and a reviewer's M4 measurements show exactly that. On this x86-64 host the polynomial backend is 1 to 49% *slower* than the tower one at these shapes, so the split beats both. The entry point's doc now says which transform the second phase runs, and routing it through the encoder's own choice is recorded under Deferred.

## Encode a column-major message into an interleaved codeword

A commitment that folds several variables at a time opens whole folding blocks, so its codeword is a matrix of one column per block; a prover holds the opposite layout and today brings the two together in a pass of its own. Zero-padding is what extends the domain, and the layers crossing the padding read only zeros, so they copy the message into each coset and compute nothing. Writing that copy directly in the interleaved order puts the transpose inside a pass that has to happen anyway, and what is left is one shifted transform per coset.

Sixteen columns of `BinaryField128`, against the transpose-then-encode route a commitment takes today:

| shape | portable before | portable after | gain | native before | native after | gain |
| --- | --- | --- | --- | --- | --- | --- |
| 2^16, rate 1/2 | 27.88 ms | 24.61 ms | 1.13x | 7.92 ms | 7.18 ms | 1.10x |
| 2^16, rate 1/4 | 58.97 ms | 48.25 ms | 1.22x | 13.42 ms | 11.74 ms | 1.14x |
| 2^18, rate 1/4 | 277.8 ms | 227.1 ms | 1.22x | 63.80 ms | 61.34 ms | 1.04x |
| 2^20, rate 1/4 | 1285 ms | 1014 ms | 1.27x | 291.5 ms | 282.9 ms | 1.03x |
| 2^20, rate 1/8 | 2681 ms | 2057 ms | 1.30x | 520.2 ms | 501.9 ms | 1.03x |

Two small shapes go the other way on the portable leg: 2^14 at rate 1/2 is 0.97x and at rate 1/4 is 0.93x, where there are too few cosets for the split to pay for itself. This is a new entry point rather than a replacement, so no existing caller regresses; a caller at those shapes should keep the two-pass route.

The benchmark carries a same-backend arm alongside the real baseline, because most of the portable-leg gap against the default encoder is the blocked schedule rather than the fusion. Review pointed out that this arm still ran the padding layers as full butterflies while the fused path skipped them, so the old 0.90x-to-1.41x figure was the fusion *plus* the skipped layers. The padding skip now lives in the tower transform's padded entry point, both sides of the pair use it, and the corrected arm measures the transpose alone: **0.86x to 1.05x over the twelve shapes on the native leg**, and 0.75x to 1.18x at 2^16 on the portable one. The fusion on its own is neutral, turning slightly positive only at 2^20 where the separate transpose pass no longer fits anywhere in cache. The main table above is unaffected, since its baseline already skipped those layers through the polynomial backend.

## Benchmark the univariate-skip extension

The table-driven extension had no benchmark, so nothing said what it costs against the composition and the equality weighting around it. Three arms — one row, a block of rows, and the whole streamed round message — over two levels and three domain shapes. There is no baseline to compare against: the arms are the measurement instrument for the item below.

| arm | portable | native |
| --- | --- | --- |
| one row, byte level, six-bit skip, degree two | 18.17 ns | 17.94 ns |
| streamed round message, 2^16 rows, three operands | 2.674 ms | 2.712 ms |

## Dropped: packed lanes for the skip-round extension

Implemented, measured, and reverted. The kernel was restructured to hold its running sums in registers and to extend a group of rows side by side, which needed the accumulator written back as a whole array value before the compiler would keep the eight-byte exclusive-or intact at all.

| arm | portable before | portable after | gain | native before | native after | gain |
| --- | --- | --- | --- | --- | --- | --- |
| one row (the prover's call) | 18.17 ns | 18.77 ns | 0.97x | 17.94 ns | 18.36 ns | 0.98x |
| a block of rows | 85.69 us | 58.98 us | 1.45x | 91.39 us | 63.74 us | 1.43x |
| streamed round message | 2.674 ms | 2.878 ms | 0.93x | 2.712 ms | 2.799 ms | 0.97x |

The grouping is worth 1.10x to 1.45x on the batched entry point, which nothing in the repository calls. On the two paths that are called it is neutral or worse: the single-row kernel does not improve, and grouping the streamed round message *regressed* it by 3-7%, because a group's scratch is large enough to push the composition loop out of the first-level cache. The benchmark also shows why a perfect kernel could not have paid: at 18.2 ns per row, three operands and 2^16 rows come to 3.57 ms of serial extension against a 2.67 ms wall-clock round on 32 threads, so the extension is about four percent of the round and the equality-weighted composition over the wide alphabet is the rest. The eight-byte-register reading of the kernel was also tried and was 3x *slower* at the byte level, because it moves a runtime multiply into the innermost loop. Only the benchmark is kept.

## Closed without code: an explicit domain context

Item 3 asked for the precompute-versus-recompute choice in the twiddles to be exposed as a documented knob. On reading the code the premise turns out to be false: there is no choice left to expose, and adding an API for one would make the transform slower.

Both twiddle implementations already precompute the cheap part and derive the rest. The stage bases are a squaring chain, one squaring per stage rather than `j` squarings apiece, and the increments between consecutive block twiddles depend only on a block index's trailing-zero count, never on the stage, so a single table serves every stage. A stage pass then computes one twiddle per *task* and carries it from block to block by one addition. The transform's own module header already says it: there is no twiddle table.

What a context would add is the one thing left, a full per-stage per-block table. That is `2^dimension - 1` entries, so 16 MiB at a height of `2^20` — the size of the matrix being transformed — to replace two exclusive-ors per block with a load that at that size misses cache. It is slower and it doubles the footprint, which is the opposite of what the item wanted.

So the item is closed as already satisfied, and nothing here changes. Recording it because a plan item that is silently dropped looks like an oversight, and this one is a conclusion.

## Deferred

- **An explicit domain context** (item 3) — not started. The twiddle state is already an O(dimension) squaring chain plus one increment table, so the memory-versus-time knob a context would expose is a full per-stage twiddle table costing as much memory as the matrix; that wants its own measurement before any API is committed to.
- **Grouping the streamed round message** — the restructuring above is what would carry it, and it measured as a regression. The cost that actually dominates that loop is one wide-by-subfield product per transmitted point per row; splitting it into a per-row nibble table is the lead worth following.
- **AArch64 intrinsics**, tracked separately and not validatable here.
- **The staging-tile gather** in `poly.rs` was left untouched, as an open pull request owns it. The interleaved encoder consequently runs the tower-basis transform rather than the polynomial-basis one; fusing the interleaving into that backend's own coset copy is the follow-up.
- **Phase 2 of the subfield split through the encoder's own backend.** After the narrow phase the buffer is `2^head` contiguous chunks, and chunk `a` needs the bottom layers with its block indices offset by `a` — which, the subspace polynomials being `F_2`-linear, is a shifted transform of that chunk with shift `domain_point(a << low)`. So phase 2 is expressible as `2^head` calls to any backend's shifted transform. Through the public trait that costs a copy in and out per chunk; without the copies it means changing basis once over the whole buffer inside `PolyBasisNtt`, which `poly.rs`'s open pull request owns.
- **The padding skip for the portable encoder.** `PolyBasisNtt::ntt_batch_padded` short-circuits to the full transform without a carryless multiply, so it runs the padding layers even though its tower fallback can now skip them. One line in `poly.rs`, again owned elsewhere.


Nashtare (commented):
I found no correctness bug. The tower subfield `GF(2^b)` is both the low `b` bits and the span of the first `b` Cantor vectors. So the typed products are coordinate-wise, and the top `b` layers of an unshifted transform stay in the subfield. The interleaved coset shifts follow from `W_j(v_i) = v_{i-j}`.

I also ran a differential outside CI, at 4 and 32 Rayon threads, with and without `aes`. It matched `PolyBasisNtt` / `AdditiveRsEncoder` for:
- `subfield_ntt_batch` from `Gf2`/8/16/32 into `BinaryField128`, up to 2^18 × 16;
- `subfield_encode_batch`;
- `interleaved_encode_batch`, up to 2^16 × 16 at rates 1/2 to 1/8.

The 32-thread run sends a partial stage range through `fused_group`, which CI never does (inline).

**The speedups need another look before I'd approve.**
- On AArch64 the subfield split is slower than what it replaces. Phase 2 runs the tower stages, so on an M4 it loses 1.5-2.1x to widen + `PolyBasisNtt` (inline, with numbers).
- On your tables, the `wide/*` baselines widen serially while the split widens in parallel. The ~11 / ~48 ms left in "before" at 2^18 / 2^20 looks like serial page faults on a fresh `Vec`, not transform work (inline).
- `two_pass/lch` doesn't isolate the fusion: `LchNtt` runs the zero-padding layers the fused path skips.

**Design**
- The three new public functions have no caller. The interleaved coset split is exactly the `ntt_batch_padded` that `LchNtt` is missing, which #2166's narrow encoders go through. Moving it there would give it callers through `Encoder`.
- The CLMUL crossover probably belongs in `extension.rs`'s gate, extended to `BinaryField64`, rather than in a second copy in the butterfly.

The rest is docs and small test gaps.

binary-dft/benches/ntt.rs
```diff
@@ -234,6 +238,116 @@ fn bench_encode(c: &mut Criterion) {
...
+
+    for log_height in LOG_HEIGHTS {
+        for width in [1, WIDTH] {
+            let message = RowMajorMatrix::<BinaryField8>::rand(&mut rng, 1 << log_height, width);
+            let parameter = format!("h{log_height}/w{width}");
+
+            // Throughput counts the matrix entries, so every arm compares directly.
+            group.throughput(Throughput::Elements((width << log_height) as u64));
+
+            let widen = |m: &RowMajorMatrix<BinaryField8>| {
```
Nashtare, line 270: This `widen` is a serial `iter().map().collect()`, while `subfield_ntt_batch` widens with `par_iter` above 4 MiB (`subfield.rs:42`), which every width-16 shape here reaches. On an M4 that gap is small: widening 2^20 x 16 takes 4.7 ms serially and 2.9 ms in parallel, and a parallel widen moves `wide/poly` from 39.2 to 38.0 ms. Your tables look different, though. Subtracting the width-16 transform from the "before" column leaves ~11 ms at 2^18 and ~48-50 ms at 2^20 on both legs (native 15.18 - 3.89, portable 50.98 - 39.6). On Linux with 4 KiB pages a fresh 64/256 MiB `Vec` is first-touched through page faults, and a serial collect takes all of them on one thread, which would account for that. Could the wide arms widen in parallel, so the table measures the transform?

binary-dft/benches/ntt.rs
```diff
@@ -234,6 +238,116 @@ fn bench_encode(c: &mut Criterion) {
...
+                    let source = RowMajorMatrixView::new(&columns, 1 << log_message);
+                    let mut target =
+                        RowMajorMatrixViewMut::new(&mut values[..columns.len()], WIDTH);
+                    source.transpose_into(&mut target);
+                    encoder.encode_batch_padded(RowMajorMatrix::new(values, WIDTH), log_inv_rate)
+                });
+            });
+
+            // The same two passes over the transform the fused path itself runs on.
+            // The pair therefore isolates the fusion from the choice of backend.
```
Nashtare: `LchNtt` has no `ntt_batch_padded` override, so `lch.ntt_batch` here runs the `log_inv_rate` zero-padding layers as full butterflies. `interleaved_encode_batch` skips them by copying. So the gap between this arm and `fused` is the fused transpose *plus* the skipped layers, not the fusion alone, which is what this comment and the description's "the fusion alone is 0.90x to 1.41x" say. Either reword it, or give this arm a padding-aware transform so it measures only the transpose.

binary-dft/src/butterfly.rs
```diff
@@ -149,6 +150,149 @@ where
...
+
+/// Whether a typed product beats the full-width product at the two widest levels.
+///
+/// - Scaling by a `D`-byte subfield costs one subfield product per coordinate.
+/// - Recursive multiplication triples in cost per doubling of the width.
+/// - The coordinate count only halves, so the narrower subfield always wins.
+///
+/// A carryless-multiply instruction breaks that at the widest two levels.
+/// Only the one-byte coordinates stay ahead of it, being single table lookups.
+const WIDE_TYPED_PRODUCTS_PAY: bool = !HAS_HARDWARE_CLMUL;
```
Nashtare, line 211: The +89% measurement says that with CLMUL, `BinaryField64 * BinaryField16/32` through the coordinate expansion is slower than the full-width product. `Mul<$lower> for $upper` in `binary-field/src/extension.rs:50` already switches to the full product under `HAS_HARDWARE_CLMUL`, but only when `BITS == 128`. So every other `BinaryField64` × subfield product in the workspace still takes the slow route on CLMUL targets. Would it be simpler to add 64 to that gate and drop this one? Then the crossover lives in one place and the other callers get the fix. Or is there a reason the butterfly needs its own copy?

binary-dft/src/lch.rs
```diff
@@ -484,23 +492,26 @@ fn run<F: ButterflyField, const INVERSE: bool>(
...
-    // The schedule is the caller's, so bring it inside what this matrix can hold: a tile is at
-    // most the whole height, and a staged row is a run of matrix rows that must not straddle a
-    // butterfly of the narrowest stage above the tile, which pairs rows `2^tile` apart.
-    let tile = schedule.log_tile_rows.min(log_n);
+    // The schedule is the caller's, so bring it inside what this call can use.
+    //
+    // - A tile is at most the range of stages asked for.
+    // - A staged row is a run of matrix rows.
+    // - That run must not straddle a butterfly of the narrowest stage above the tile.
+    let tile = schedule.log_tile_rows.min(top);
```
Nashtare, line 519: `run` can now execute a stage range that stops short of `log_n`, and the subfield split's phase 2 is the only caller that does that. The schedule tests still pass `top = log_n` everywhere. CI runners have fewer than `STAGED_WORKERS` threads, so `fused_group` never runs a partial range in CI. I ran a differential with 32 Rayon threads: `Gf2` at 2^13 x 16 sends stages 8..12 through a fused group, and it matches `PolyBasisNtt` on the widened matrix. A regression test would still be cheap: sweep `top` in the `for_workers` test, against the same stages run one pass at a time.

binary-dft/src/lch.rs
```diff
@@ -522,27 +533,51 @@ fn run<F: ButterflyField, const INVERSE: bool>(
...
-/// Transform a matrix in place, in whichever direction the flag selects.
-fn transform<F: ButterflyField, const INVERSE: bool>(
-    mut mat: RowMajorMatrix<F>,
+/// Run the `top` narrowest stages over a row-major buffer of `width` columns, in place.
+///
+/// Stages are numbered by the distance they pair rows across, so stage zero is the narrowest.
+/// The whole transform is every stage the row count allows.
+///
+/// A caller that stops short is left holding the matrix part-way through the network.
+/// What remains of it is the stages from `top` upwards.
```
Nashtare: This is right for the inverse, where the narrow stages come first and the wide ones are still to run. The forward direction is the other way round. There the stages from `top` upwards come *first* in the network, so `transform_stages::<_, false>(.., top, ..)` only completes a transform if those stages have already run. `split_ntt_batch` relies on exactly that. Since the split's correctness depends on it, this is worth spelling out for each direction.

binary-dft/src/interleaved.rs
```diff
@@ -0,0 +1,273 @@
...
+
+use crate::butterfly::ButterflyField;
+use crate::domain::domain_point;
+use crate::lch::transform_stages;
+
+/// Reed–Solomon encode a message held one folding column at a time.
+///
+/// # Overview
+///
+/// A commitment that folds `2^f` variables at a time opens whole folding blocks at once.
```
Nashtare: Folding `f` variables at a time gives `2^f` columns, which is what the next line says. "Folds `2^f` variables" reads as `2^(2^f)` columns. Suggest "folds `f` variables at a time", which also matches `commit_base`, where `width = 1 << folding` for a fold of `folding` variables.

binary-dft/src/interleaved.rs
```diff
@@ -0,0 +1,273 @@
...
+        "domain exceeds field dimension"
+    );
+
+    let width = columns.len() / rows;
+    let len = columns.len();
+    let mut values = F::zero_vec(len << log_inv_rate);
+
+    // The first coset is the message with its columns interleaved, blocked by the transpose.
+    let (first, rest) = values.split_at_mut(len);
+    if width != 0 {
```
Nashtare: This guard suggests empty input is handled. It isn't: with `columns` empty, `len` is 0, and `rest.par_chunks_mut(len)` below panics with `chunk_size must not be zero`. The serial shim's `chunks_mut(0)` panics the same way, and `transform_stages` would then divide by a zero width. Either return an empty matrix early, or assert non-empty input up front with a clear message and list it under `# Panics`.

binary-dft/src/interleaved.rs
```diff
@@ -0,0 +1,273 @@
...
+        let source = RowMajorMatrixView::new(columns, rows);
+        let mut target = RowMajorMatrixViewMut::new(first, width);
+        source.transpose_into(&mut target);
+    }
+
+    // Every other coset is a copy of it, transformed where the copy leaves it in cache.
+    rest.par_chunks_mut(len).enumerate().for_each(|(c, coset)| {
```
Nashtare: This copy-then-shifted-transform loop is the zero-padding skip that `LchNtt` itself lacks. `LchNtt` inherits the default `ntt_batch_padded`, which runs the copy layers as full butterflies. With #2166, `AdditiveRsEncoder<BinaryField8..BinaryField64, LchNtt<_>>` goes through exactly that default. Would it make sense to move this split into `LchNtt::ntt_batch_padded` (row-major input)? This function would then only do the transpose into coset 0 and call a shared helper. The encoders, and later `commit_base`'s prefix path, would then reach it through `Encoder`, rather than through a free function with no caller.

binary-dft/src/subfield.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// Evaluate each column of a subfield message on the subspace of the wider level.
+///
+/// Widening the message first pays the wide element size at every layer.
+/// The widest layers do not need it: their twiddles map the subfield into itself.
+///
+/// # Panics
+///
+/// Panics if the height is not a power of two.
+/// Panics if the domain dimension exceeds the bit width of either level.
```
Nashtare: Only the wide level is bounded here. `split_ntt_batch` asserts `log_n <= 1 << F::LOG_BITS` and caps the narrow phase at `closed_layers(S::LOG_BITS).min(log_n)`, so a byte message on a 2^10 domain is accepted. `a_byte_message_transforms_like_the_widened_one` relies on that. Suggest "exceeds the bit width of `F`".

binary-dft/src/subfield.rs
```diff
@@ -0,0 +1,395 @@
...
+where
+    S: ButterflyField,
+    F: ButterflyField + From<S>,
+{
+    let len = message.values.len();
+    let _ = log2_strict_usize(message.height());
+
+    // A shift amount below the word size still leaves the value itself free to overflow.
+    // Recovering the original length from the shifted one is what proves no bits were lost.
+    let padded_len = u32::try_from(log_inv_rate)
```
Nashtare: This is the padded-length check from `AdditiveRsEncoder::encode_batch`, copied verbatim. #2166 moves that code into `encode_by_padding`, so once both land there are two copies. Could they share one helper in `encoder.rs`? Also, `subfield_encode_batch` has no caller and no bench arm, unlike `subfield_ntt_batch`. Is it meant for the Boolean-trace commit in #2166, or can it wait until that caller exists?

binary-dft/src/subfield.rs
```diff
@@ -0,0 +1,395 @@
...
+            split_ntt_batch::<BinaryField8, BinaryField128>(message, 0)
+        );
+    }
+
+    #[test]
+    fn a_byte_message_matches_the_reference_oracle() {
+        // The oracle evaluates the novel basis straight from its product definition.
+        // So it pins the split to that basis, not to another split of the same network.
+        for width in [1usize, 3] {
+            for log_n in 0..=6 {
```
Nashtare, line 299: With `log_n <= 6` the split depth is `min(8, log_n) = log_n`, so phase 2 has no layers. This test therefore pins only the byte-level transform to the oracle, not the split itself. The split is pinned to `LchNtt`, which covers it transitively. Still, adding `log_n` 9 and 10 at width 1 (the naive transform is documented as fine up to 2^10) would check a result with both phases non-empty directly against the product definition.

binary-dft/src/subfield.rs
```diff
@@ -0,0 +1,395 @@
...
+    );
+    let low = log_n - head;
+
+    // Phase 1: every layer of a shorter, wider reading of the same buffer.
+    transform_stages::<S, false>(&mut mat.values, width << low, head, S::ZERO);
+
+    let mut values = widen::<S, F>(&mat.values);
+
+    // Phase 2: the layers the subfield is not closed under, at the alphabet's own width.
+    transform_stages::<F, false>(&mut values, width, low, F::ZERO);
```
Nashtare, line 94: Phase 2 runs the tower-basis stages, so where `PolyBasisNtt` is the faster backend the split loses to the route it is meant to beat. On an M4 (`aes`, so `AdditiveRsEncoder` picks `PolyBasisNtt`), `--features parallel`, this PR's `subfield` bench gives:

| width 16 | `wide/poly` | `wide/lch` | `subfield` |
|---|---:|---:|---:|
| 2^16 | 2.42 ms | 5.89 ms | 4.15 ms |
| 2^18 | 9.16 ms | 22.96 ms | 17.34 ms |
| 2^20 | 39.19 ms | 105.57 ms | 81.63 ms |

Width 1 is 1.5-1.8x slower than `wide/poly` too. It beats its own backend, but the description compares against the better of the two. I haven't run Graviton. Could phase 2 go through `PolyBasisNtt` where the encoder would pick it, or could the docs and description say the win is for the tower backend only?

binary-dft/src/butterfly.rs
tcoratger: Half right, and the half that is right is now fixed in the field crate.

**Adopted: the gate in `extension.rs` was incomplete.** Widened to `matches!(<$upper>::BITS, 64 | 128)`. I added a `subfield_mul` arm to `binary-field/benches/arithmetic.rs` to measure it rather than assert it (1000 products, fixed narrow operand, XOR accumulator so they overlap). Native, Zen 5, `-C target-cpu=native`:

| pair | coordinate expansion | embedded wide product | change |
| --- | ---: | ---: | ---: |
| `BinaryField64 * BinaryField32` | 14.97 us | 6.045 us | **-59.4%** |
| `BinaryField64 * BinaryField16` | 6.430 us | 6.189 us | -8.6% (p=0.00) |

`BinaryField64 * BinaryField8` is untouched at 2.60 us, correctly: eight table lookups still beat one `mul_64`. Portable is unaffected, the gate is inert with no carryless multiply.

**Not adopted: dropping the butterfly's own gate.** The two are not the same decision. `extension.rs` picks between two *products*; `WIDE_TYPED_PRODUCTS_PAY` picks between `typed_butterfly`, a scalar element-at-a-time loop, and `packed_butterfly`, which runs `F::Packing` over a whole register. Widening the field gate does not give the typed loop its packing back, and the whole scaling chain is inlined into one kernel, so carrying the wide levels through it costs the *full-width* route its codegen even where no twiddle takes the typed path.

Measured with the field gate widened and `WIDE_TYPED_PRODUCTS_PAY` forced to `true`, native:

| arm | change |
| --- | ---: |
| `butterfly/64/t128` | **+77.6%** [+65.8, +89.7] |
| `butterfly/128/tower/t128` | +8.2% |
| `ntt/128/tower/shifted` 2^14 / 2^16 / 2^18 / 2^20 | +15.7% / +27.8% / +22.1% / +24.6% |

The shifted arms are the decisive ones: a coset shift puts every twiddle outside the small subfields, so they are all full-width and they all regress. The narrow-twiddle arms do improve (`64/t16` -18%, `128/tower/t16` -25%), but a real transform is mostly the other kind.

One more datum on "other callers get the fix": there are none today. `grep` for `Mul<BinaryField16` / `Mul<BinaryField32` across the workspace returns only `binary-dft/src/butterfly.rs`. The field-crate fix is for the next caller, not an existing one.

The constant's doc now says why the two gates are separate.

binary-dft/src/lch.rs
tcoratger: Agreed, and added.

`a_partial_stage_range_matches_the_plain_passes` now sweeps `top` over `0..=log_n` and compares `run` against the same stages run one plain pass at a time, in both directions and with a round trip of the range on top. The reference is a new `stages_one_at_a_time` helper, which `stage_by_stage` now delegates to, so there is one definition of "these stages in this order".

The shapes are named rather than read off the machine, for exactly the reason you give:

- `shape(3, 2, 0)` at 2^8 — a range ending inside the tile, at its edge, inside a fused group, and past a clipped one.
- `shape(2, 4, 0)` at 2^7 — groups deeper than the stages above them, so the top group is always clipped.
- `shape(0, 3, 0)` at 2^6 — no tile, so every stage of the range is in a group.
- `shape(4, 3, 2)` at 2^8, `BinaryField32` — staged rows holding a run of matrix rows.
- `Schedule::for_workers::<BinaryField128>(16, 11, workers)` for `workers in [1, 32]`, asserting a group exists above the tile.

Widths 1, 3 and 16. That runs a partial range through `fused_group` regardless of the CI runner's thread count.

binary-dft/src/lch.rs
tcoratger: Right, the sentence only described the inverse. Replaced with both directions and a diagram, since the split's correctness rests on the forward one:

```text
///     forward   stage l-1 ... stage top   then   stage top-1 ... stage 0
///               \___ the caller's own ___/        \___ this call _____/
///
///     inverse   stage 0 ... stage top-1   then   stage top ... stage l-1
///               \___ this call _______/          \___ the caller's own ___/
```

followed by: a forward caller has to have run the wider stages already, an inverse caller still has them ahead of it.

binary-dft/src/interleaved.rs
tcoratger: Fixed — "folds `f` variables at a time", which agrees with `width = 1 << folding` in `commit_base`.

binary-dft/src/interleaved.rs
tcoratger: You are right, and the guard was worse than useless — it made the function look like it handled a case it panicked on three lines later, with `chunk_size must not be zero` rather than anything a caller could act on.

Empty input is now rejected up front with `at least one message column`, the `width != 0` guard is gone, and `# Panics` lists it. Returning an empty matrix is not an option: `RowMajorMatrix::new(vec![], 0)` divides by zero computing its own height, so there is no well-formed empty codeword to return.

While in there, two more silent-wrong-answer paths at the same boundary, both reachable through a `pub` entry point:

- `1usize << log_message` — the level assert admits `log_message` up to 128 for a 16-byte level, and a shift that wide is masked in release. `log_message = 100` with a 16-element slice gave `rows = 1`, passed the divisibility check, and returned nonsense. Now a `checked_shl`, with `message dimension overflows usize`.
- `len << log_inv_rate` — same wrap, and the result is the *allocation size*, so it under-allocates and the coset split then reads past the message. Now goes through the shared `padded_message_len` helper.

Three tests: `a_message_with_no_column_is_refused`, `a_message_dimension_past_the_address_space_is_refused`, and the existing overflow test via the shared helper.

binary-dft/src/interleaved.rs
tcoratger: Agreed, and done. The split now lives in `lch.rs` as `transform_cosets`, and `LchNtt` overrides `ntt_batch_padded` to call it:

```rust
fn ntt_batch_padded(&self, mut mat: RowMajorMatrix<F>, log_inv_rate: usize) -> RowMajorMatrix<F> {
    // asserts elided
    transform_cosets::<F>(&mut mat.values, width, log_n - log_inv_rate);
    mat
}
```

`interleaved_encode_batch` is now the transpose into coset 0 plus that one call, and the `# Algorithm` section moved with the code. So `AdditiveRsEncoder<_, LchNtt<_>>` reaches the skip through `Encoder`, as you wanted, and it is no longer a free function with a single caller.

Pinned three ways, none of which goes through the new path:

- `the_padded_transform_matches_the_full_one` — every width in `WIDTHS`, `log_message` 0..=5, rate 0..=3, over `BinaryField32`, `BinaryField128` and `Ghash128`, against `ntt_batch` on the padded matrix, i.e. against running the skipped layers as real butterflies.
- `the_padded_transform_matches_the_reference_oracle` — against `NaiveAdditiveNtt`, so the coset shifts are pinned to the novel basis from its product definition rather than to another split of the same network.
- a proptest over random shapes, plus the two `should_panic` boundary tests.

One thing I could not finish here. On the portable leg `PolyBasisNtt::ntt_batch_padded` short-circuits with `if log_inv_rate == 0 || !HAS_HARDWARE_CLMUL { return self.ntt_batch(mat) }` (`poly.rs:701`), so it runs the padding layers even though its tower fallback can now skip them. Changing that line to `self.tower.ntt_batch_padded(mat, log_inv_rate)` would hand the portable encoder the skip for free, but `poly.rs` is owned by #2155 and I stayed out of it. Worth a one-liner there or in #2166.

binary-dft/src/subfield.rs
tcoratger: Correct, and your reading of the code is right: `split_ntt_batch` bounds only the wide level, and the narrow phase is capped at `closed_layers(S::LOG_BITS).min(log_n)`, so a byte message on a 2^10 domain is accepted and `a_byte_message_transforms_like_the_widened_one` depends on it.

Changed to "Panics if the domain dimension exceeds the bit width of the wider level." Kept to one line — the repo's doc gate caps this file at 30% comment and it was already at 29.4%.

binary-dft/src/subfield.rs
tcoratger: Both points taken.

**Shared.** `padded_message_len` now lives in `encoder.rs` and is used by `AdditiveRsEncoder::encode_batch`, `subfield_encode_batch` and `interleaved_encode_batch`. It has its own tests, including the case `checked_shl` alone misses (`1 << 63` shifted once more keeps a valid shift amount and loses its only set bit). When #2166 lands, `encode_by_padding` should call it rather than carry a fourth copy.

I deliberately left the two *LDE* copies alone (`traits.rs:84`, `poly.rs:761`). They panic with `extended codeword length overflows usize`, a different message for a different operation, and `poly.rs:1095` asserts on that exact string — on the portable leg it reaches `traits.rs`'s copy through the tower fallback, so unifying the message would break a test in a file #2155 owns. Flagging rather than editing.

**Caller.** It is for the Boolean-trace commit in #2166, so it has no caller yet — but it now has a bench arm, so the claim is measured rather than asserted. `subfield/encode` over the same byte message at rate 1/2, native, width 16: 399 us at 2^14, 2.60 ms at 2^16, 17.08 ms at 2^18, 89.69 ms at 2^20. Against the plain `subfield` transform at 2^20 (38.01 ms) that is 2.36x for 2x the codeword, which is the shape you would expect when the padding itself is paid at the byte width.

binary-dft/src/subfield.rs
tcoratger: Right — with `log_n <= 6` the head is the whole height and phase 2 is empty, so that test never pinned the split itself to the product definition.

Split into two tests sharing one `check_against_the_oracle` helper. The second adds `log_n` 9 and 10 at width 1:

```text
    log_n = 9    head = 8, phase 2 runs one layer at the wide width
    log_n = 10   head = 8, phase 2 runs two of them
```

So the result with both phases non-empty is now checked directly against `NaiveAdditiveNtt`, not only transitively through `LchNtt`. Width 1 keeps the oracle's cost per coefficient per row affordable at 2^10, as you suggest.

binary-dft/benches/ntt.rs
tcoratger: Good catch, and the effect on this machine was larger than on yours.

`bench_subfield`'s `widen` closure now uses `par_iter`, matching what `subfield_ntt_batch` does above 4 MiB. Your reconstruction was right: the residual you subtracted was the first-touch page faults of a 64/256 MiB `Vec` taken on one thread.

Native, width 16, `wide/lch` with the old serial closure against the new parallel one:

| height | serial widen | parallel widen | difference |
| --- | ---: | ---: | ---: |
| 2^18 | 15.18 ms | 10.09 ms | -34% |
| 2^20 | 78.80 ms | 58.23 ms | -26% |

So the wide arms were carrying roughly 5 ms at 2^18 and 20 ms at 2^20 that the specialised arm never paid, and the native gains for that item fall from 2.45x/2.15x to about 1.5x. Two independent native runs agree on the ratio to within 2%.

One caveat I want on the record rather than hidden. This box is currently running other 32-thread benchmark jobs from parallel work, load average 30 on 32 threads. Per-arm confidence intervals stay tight at 60 samples, but the *load itself* moves between arms, so two runs of the portable leg gave 1.62x and 1.19x for the same 2^14 comparison. The native ratios reproduced; the portable ones did not. I have corrected the native column and flagged the portable one in the description as needing a re-run on a quiet host, rather than quote a figure I could not reproduce.

binary-dft/benches/ntt.rs
tcoratger: You are right, and the fix changes the conclusion rather than the wording.

`LchNtt` now overrides `ntt_batch_padded` (see the thread on `interleaved.rs:81`), and this arm calls it. Both sides of the pair now run the same backend and skip the same layers, so what is left between them is the fused transpose and nothing else.

Re-measured on the corrected arm, `two_pass/lch` against `fused`, native:

| shape | `two_pass/lch` | `fused` | fusion alone |
| --- | ---: | ---: | ---: |
| 2^16, rate 1/2 | 7.62 ms | 8.56 ms | 0.89x |
| 2^16, rate 1/4 | 11.37 ms | 12.56 ms | 0.90x |
| 2^18, rate 1/4 | 62.69 ms | 64.04 ms | 0.98x |
| 2^20, rate 1/4 | 325.0 ms | 309.4 ms | 1.05x |
| 2^20, rate 1/8 | 559.8 ms | 542.9 ms | 1.03x |

Native range over all twelve shapes: **0.86x to 1.05x**. Portable at 2^16, 60 samples: 0.87x, 0.75x, 1.18x at the three rates.

So the old "0.90x to 1.41x" was the fusion *plus* the skipped layers, exactly as you say — the 1.41x end was the padding skip, which now sits behind `Encoder` for everyone rather than only inside this function. Fusion on its own is neutral, turning slightly positive only at 2^20 where the separate transpose pass no longer fits anywhere in cache. The description's sentence is corrected.

The main table is unaffected: its baseline is `two_pass` through `AdditiveRsEncoder`, whose `PolyBasisNtt` already skipped those layers on the native leg.

binary-dft/src/subfield.rs
tcoratger: Taking your second option — the docs and the description now scope the claim. Three things for the record first, because the picture turns out to be target-specific rather than universal.

**On x86-64 the split beats both backends, including the one the encoder picks.** Re-measured after fixing the serial widen you spotted in the neighbouring thread. Native, `-C target-cpu=native`, width 16, all three arms in one run:

| width 16 | `wide/lch` | `wide/poly` | `subfield` |
| --- | ---: | ---: | ---: |
| 2^14 | 317.9 us | 331.7 us | 269.2 us |
| 2^16 | 811.6 us | 1.212 ms | 690.8 us |
| 2^18 | 10.09 ms | 10.48 ms | 6.555 ms |
| 2^20 | 58.23 ms | 58.65 ms | 38.01 ms |

Two independent runs agree on those ratios to within 2%. `PolyBasisNtt` is *not* the faster route here — it is 1 to 49% slower than the tower transform at these shapes — so on this host the split beats the better of the two at every height. That is the opposite of your M4 table, and the difference is the backend, not the split: with `aes`/PMULL the polynomial basis wins big, without it the change of basis does not pay for itself.

**So your point stands where the backends diverge.** Phase 2 runs `log_n - 8` layers at the wide width through the tower transform, which at 2^20 is 12 of 20 layers — the majority of the work. Where another backend is 2.4x faster at that width, widening first through it beats the split, and the description comparing against "the better of the two" was hiding that. Both now say so: `subfield_ntt_batch`'s doc gained a line, and the description's section is scoped and carries the caveat.

**On actually routing phase 2 through the other backend.** The algebra permits it and is worth writing down for whoever picks it up. After phase 1 the buffer is `2^head` contiguous chunks of `2^low` rows, and chunk `a` needs exactly the bottom `low` layers with its block indices offset by `a`. Since the subspace polynomials are `F_2`-linear and the index bits are disjoint, that is precisely a *shifted* transform of chunk `a` with shift `domain_point(a << low)` — the same identity the interleaved encoder already relies on for its cosets. So phase 2 could be `2^head` calls to `AdditiveNtt::shifted_ntt_batch` on any backend.

I did not do it here for two reasons. Through the public trait it costs a copy in and out per chunk, since `shifted_ntt_batch` takes an owned matrix, which eats a good part of what the faster backend would buy. Doing it without the copies means reaching into `PolyBasisNtt` to change basis once over the whole buffer and run the coset transforms in that basis — and `poly.rs` is owned by #2155, so I stayed out of it. That is the shape of the follow-up.

Nashtare (approved):

PR #2174, opened 2026-09-17
perf(sumcheck,binary-pcs): run binary PCS sumcheck rounds in Ghash128 and bind long suffix rounds in place

Nashtare:
## Summary

Two prover-side speedups for the residual sumcheck that `BinaryPcs` folds in lockstep with its codeword, found while profiling the Keccak-f and BLAKE3 examples of #2164. Proofs and transcripts are unchanged.

### Sumcheck rounds in an isomorphic field (`p3-sumcheck`, `p3-binary-pcs`)

- **`ReprSumcheckProver<F, EF, R>`** moves a `SumcheckProver<F, EF>`'s tables, running claim and held challenge into a field `R` isomorphic to `EF`, and plays the same rounds there. Round messages cross back into `EF` before the transcript observes them, and challenges cross into `R` before they are held, so the transcript is the one the `EF` prover writes.
- **`BinaryPcs` fold rounds run in `Ghash128`.** A `BinaryField128` product changes basis three times around one carryless multiply; in `Ghash128`, the same field in the polynomial basis, it is the multiply and a reduction. The conversion happens once, after `into_sumcheck`.
- **Whole-table conversion.** The tables cross into `R` through `FromTable`, whose default maps entry by entry into a new buffer. `Ghash128` overrides it with `Ghash128::from_tower_vec` (`p3-binary-field`), which converts in the table's own buffer, in parallel chunks, through the slice basis-change kernel that crosses whole blocks together on `gfni` + `avx512f` + `avx512bw` builds. On the M4 Pro that kernel is the per-element path, so the gain there is the reused buffers: peak RSS of Keccak-f `-l 16` 21.0 → 19.8 GiB, `-l 14` 5.5 → 5.0 GiB, BLAKE3 `-l 12` 10.8 → 9.8 GiB.

### In-place suffix binding (`p3-sumcheck`)

- A fused suffix round wrote its bound half into a fresh table, so every round of a long ladder faulted in a new half and released its source. Once the multiplies got cheap, that page traffic dominated the rounds at 2^25 to 2^27 entries.
- Output block `b` only overwrites input that block `b / 2` has already read. The blocks therefore run in levels that double in size (block 0 as a forward sweep, then block 1, blocks 2–3, 4–7, ...) and write over their own tables, with no second buffer.
- Serial passes and tables of at least 2^23 entries bind in place. A threaded table below that keeps the single dispatch into fresh halves: running every threaded table in place made `sumcheck/babybear/prover/plain_suffix` 50% slower at k16 and k20, because the first levels each cost a dispatch.
- The suffix tables' in-round peak drops from 1.5× to 1× the table being bound. A table bound in place keeps its buffer until a round below the threshold replaces it; the examples' peak RSS stayed within run-to-run noise (Keccak-f `-l 16` 20.4–21.3 GiB across builds, BLAKE3 `-l 12` 11.2 → 10.8 GiB).

## Measurements

Apple M4 Pro, release, `-Ctarget-cpu=native`, `parallel`, the examples of #2164 with #2165 applied, `--security-bits 96`. Medians of 3 runs interleaved across builds on a shared machine, so ranges of a few percent are noise.

| Workload | Prove before | Prove after | Sumcheck rounds before → `Ghash128` → in place |
| --- | ---: | ---: | --- |
| Keccak-f `-l 14` | 1324 ms | 1170 ms | 184 → 64 → 37 ms |
| Keccak-f `-l 16` | 5799 ms | 4839 ms | 1183 → 882 → 147 ms |
| BLAKE3 `-l 10` | 633 ms | 577 ms | 96 → 32 → 21 ms |
| BLAKE3 `-l 12` | 2444 ms | 2174 ms | 383 → 130 → 62 ms |

Per size, one in-place suffix pass against the fresh-half pass (parallel, isolated process): equal up to 2^22 entries (the same path), mixed at 2^23 (BabyBear quartic +2–5%, `Ghash128` −3–8%), −2 to −10% at 2^24, and −10 to −17% at 2^26.

Prime-field check, `cargo bench -p p3-sumcheck --bench sumcheck -- babybear/prover/plain_suffix` before vs after: within ±2% at k12/k16/k20 with and without `parallel`.

## Test plan

- [x] `ReprSumcheckProver` over `Ghash128` plays the `BinaryField128` prover's transcript (round messages, challenges, claims, next challenger sample) for packed prefix and scalar suffix pairs, handed over after 0, 1 or 2 batches (carrying the held challenge) and settled mid-ladder; dropping or perturbing the held challenge fails it. Grinding stays off there: with `parallel`, two provers may grind different valid witnesses
- [x] `Ghash128::from_tower_vec` equals the entry-wise map and keeps the buffer at lengths 0 to 2·2^16 + 67; the transcript test runs through both the `Ghash128` override and the default `FromTable` map
- [x] `BinaryPcs` proof digests of Keccak-f `-l 9`/`-l 12` and BLAKE3 `-l 8`/`-l 10` unchanged across both changes (`-l 12` binds its first round in place)
- [x] The in-place suffix pass matches a fresh binding at every multiple of four up to 18,000 entries (partial levels and blocks) and at 2^17 + 16,876 entries, whose partial last level is threaded and ends mid-block; a serial pass keeps the table's address; a threaded pass below the threshold allocates exactly the bound half; mutations of the head sweep, the level split and the level accumulation fail tests
- [x] `cargo test --release -p p3-sumcheck -p p3-binary-pcs -p p3-whir`
- [x] `cargo nextest run --features parallel` (workspace, debug), and without `parallel` for `p3-sumcheck`, `p3-binary-field`, `p3-binary-pcs`, `p3-whir`
- [x] Debug: `cargo test -p p3-sumcheck --lib`, `cargo test -p p3-binary-pcs`
- [x] `clippy`, `docs`, nightly `fmt`


sumcheck/src/strategy/repr.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    #[test]
+    fn repr_rounds_play_the_challenge_field_transcript() {
+        // Batches mix single rounds, which fuse across calls, with a settle and a longer batch.
+        let batches = [1, 2, 1, 3, 1, 1];
+        let num_variables: usize = batches.iter().sum();
+
+        for order in [VariableOrder::Prefix, VariableOrder::Suffix] {
+            for pow_bits in [0, 3] {
```
tcoratger: This is what turns CI red, and it reproduces every run for me:

```
cargo test --release -p p3-sumcheck --features parallel repr_rounds_play_the_challenge_field_transcript
  -> assertion failed: Prefix pow 3 split 0   (3/3 runs)
cargo test --release -p p3-sumcheck          repr_rounds_play_the_challenge_field_transcript
  -> ok
```

CI failed at `Suffix pow 3 split 1` instead of `Prefix pow 3 split 0`, which is the tell: the failing case moves with the machine.

`BinaryChallenger::grind` searches with `.into_par_iter().find_map_any(..)` (binary-field/src/challenger.rs:236). With 3 bits, roughly one candidate in eight is valid, so there are many witnesses and rayon returns whichever thread wins the race. The test runs `reference` and `handoff` on two independent challengers, so the moment the two grinds disagree the witness goes into the transcript and every later challenge diverges - which is exactly the shape of the failure (first challenge equal, the rest not).

Simplest fix is to drop `3` from this list: grinding occupies transcript positions that `pow 0` already walks, and it tests the challenger, not the isomorphism. If you want a grinding case, assert the witnesses are *valid* rather than *equal*.

Worth noting separately that this makes proofs non-reproducible in general. `find_map_first` would fix that at the source, but that's a bigger conversation than this PR.

sumcheck/src/strategy/repr.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+/// Maps every entry of a table into another field, releasing the source once the image is built.
+fn convert_table<A, B>(table: Poly<A>) -> Poly<B>
+where
+    A: Copy + Send + Sync,
+    B: From<A> + Send,
+{
+    let image = Poly::new(table.as_slice().par_iter().map(|&x| B::from(x)).collect());
```
tcoratger: `B::from(x)` per element takes the scalar basis change - sixteen table lookups each. There's already a blocked kernel for exactly this: `p3_binary_field::poly_basis::from_tower_slice(&mut [u128])`, whose own doc says whole blocks of 64 cross together on a gfni + avx512f + avx512bw build, "several times faster than the sixteen table lookups per element".

It wouldn't show up in your numbers because the blocked path is x86-only (`BLOCK = usize::MAX` elsewhere), so on the M4 Pro both routes are the same code.

It's worth chasing because the conversion isn't small next to the ladder: you convert `2 * 2^n` entries once and the whole ladder is only about `2^(n+1)` products, and each product is now a clmul plus a reduction. Same order of magnitude.

The slice kernel also works in place, which would drop the second full-size allocation this `collect` makes. Both types are `#[repr(transparent)]` over `u128`, so the buffer could be reused. A slice-level hook on the conversion (default = this per-element map, specialized for `BinaryField128 -> Ghash128`) would keep `convert_table` generic.

sumcheck/src/strategy/repr.rs
```diff
@@ -0,0 +1,289 @@
...
+/// What changes is the cost of each multiply inside a binding or measuring pass.
+///
+/// # Contract
+///
+/// `R::from` and `EF::from` must be mutually inverse field isomorphisms.
+/// A map that is not a ring homomorphism measures a different round message.
+///
+/// # Storage
+///
+/// Tables are held as scalars, whatever storage the source prover used.
```
tcoratger, line 41: Worth saying *why* scalar storage is free here rather than just that it happens: `BinaryField128::Packing` is itself, so unpacking costs nothing. Point this at a field where packing is real and the ladder silently loses SIMD for every remaining round, with nothing in the signature to warn you.

One line - "only worth using where `EF::ExtensionPacking` is `EF`" - would save the next person the trace.

tcoratger (commented):
The in-place suffix binding is right - I worked the level schedule through and the aliasing argument holds (level `[start, end)` writes over `[start, 2*start)`, which earlier levels already read, and reads `[2*start, 2*end)`, which nothing has touched; `split_at_mut(2*start)` is what makes the borrow checker agree). Nice piece of work, and the proptest over every multiple of four is the right shape for it.

One thing blocks it though: `repr_rounds_play_the_challenge_field_transcript` fails every single time under `--features parallel` on my machine, and CI agrees (it failed there too, at a different case than mine). It isn't your arithmetic - `grind` picks its witness with `find_map_any`, which rayon documents as nondeterministic, and the test compares two independent challengers. Details inline.

Two smaller notes below. Not approving yet, only because of the test.

On the conversions you might be wondering about: there is no base-to-extension promotion anywhere here. `ProductPolynomial<F, EF>` already stores both tables in `EF` (`F` is only there for the packing width), so the products were `EF x EF` before and after - what changes is which 128-bit representation they use. Three conversions per round plus two table sweeps up front, all of them load-bearing.

binary-field/src/ghash.rs
```diff
@@ -78,10 +82,37 @@ const CANTOR_BASIS: [u128; 128] = {
...
+        // SAFETY: `BinaryField128` is transparent over `u128`, so the initialized entries are
+        // `len` valid `u128`s. The vector is never used again, so this is the only reference.
+        let words = unsafe { slice::from_raw_parts_mut(ptr.cast::<u128>(), len) };
+        words.par_chunks_mut(TABLE_CHUNK).for_each(|chunk| {
+            clmul::tower_to_poly_128_slice(chunk);
+        });
+
+        // SAFETY: `Ghash128` is transparent over `u128` as well, so the allocation has its size
+        // and alignment, and every `u128` is a valid element.
+        unsafe { Vec::from_raw_parts(ptr.cast::<Self>(), len, capacity) }
```
tcoratger, line 122: Worth recording in the safety comment why this is specific to the top level, since it reads like it would work anywhere in the tower: `BinaryField128::MASK` is `u128::MAX`, so every bit pattern is a canonical element. At `BinaryField64` and below the mask clears high bits, so a `Vec` reinterpreted that way could hold non-canonical entries.

Not a bug - `from_tower_vec` only takes `BinaryField128`. Just the one sentence that stops someone generalising it.

tcoratger (approved):
All three things from last round are handled, and the table conversion came out better than what I suggested - reusing the buffer instead of just reaching for the slice kernel is the right call.

Re-checked what changed:

- grinding is out of the transcript test, and it passes under `--features parallel` now (it failed every run before). CI is green on all 22.
- `from_tower_vec`'s `unsafe` holds up. `BinaryField128` and `Ghash128` are both `#[repr(transparent)]` over `u128`, so same size and alignment; `BinaryField128::MASK` is `u128::MAX` at the top level so every bit pattern is canonical on the way in, and every `u128` is a valid `Ghash128` on the way out; capacity and allocator are carried through unchanged. `len == 0` lands on `Vec`'s dangling-but-aligned pointer, which `from_raw_parts_mut(_, 0)` and `Vec::from_raw_parts(_, 0, 0)` both accept - and your test covers it. Scoping it to the 128-bit level matters: the narrower tower levels mask off high bits, so the same trick would be wrong there.
- `TABLE_CHUNK` at 2^16 is 1024 whole blocks of the blocked kernel, and only the last chunk can be partial, where the per-element tail picks it up.

Nothing new from me. Approving.

PR #2177, opened 2026-09-17
perf(sumcheck,binary-pcs): build the residual weights in the representation field

Nashtare:
Stacked on #2174 (`robin/sumcheck_repr_rounds`); only the last commit is new.

## Summary

`BinaryPcs` runs its residual sumcheck in `Ghash128` (#2174), but the residual weights it hands over are still built in the tower basis: `SuffixProver::combine_weights` accumulates `scale * weight` per column slot in `BinaryField128`, where every product is three changes of basis around one carryless multiply, and `ReprSumcheckProver::new` then converts both of the prover's 2^25-entry tables.

This builds the weights in the representation field directly and converts only the evaluations table:

- `SuffixProver::preprocess` holds the rounds that `into_sumcheck` and the new `into_sumcheck_in::<R>` share (batching challenge through `transcript.finish()`), so the transcript is written by one body.
- `combine_weights` splits into `weight_plan` (the per-claim tables and the alpha-power assignment, unchanged walk order) and `combine_weights_from_plan::<B, R>` (slot accumulation and the virtual-claim equality tables, generic over the accumulation field). `combine_weights` instantiates it as `<F, EF>` and keeps its packed `SplitEq<F, EF>` for virtual claims, so WHIR's path is unchanged; `combine_weights_in::<R>` converts the plan (tiny per-claim tables and alpha powers) element-wise and accumulates in `R`. A virtual claim's point crosses into `R` before its equality table is built, which commutes with the isomorphism because the table is a polynomial in the point.
- `ReprSumcheckProver::from_tables` takes evaluations in `EF` and weights already in `R`; `SumcheckProver::new`'s debug dot-product check is the cross-check in `R`.
- `binary-pcs` calls `into_sumcheck_in::<Ghash128, _>`.

Proofs are byte-identical: a digest of the postcard bytes of `prove_keccak_binary -l 14 --security-bits 96` is the same before and after (`cf0cef8d95253ece`, 466065 bytes).

## Measurements

Apple M4 Pro, `--profile optimized`, `-C target-cpu=native`, `--features parallel`, Keccak-f `-l 14 --security-bits 96`, medians of 3 (ranges in brackets), shared machine.

| span | #2174 head (`c175d7d9`) | this branch |
| --- | ---: | ---: |
| `combine_weights` → `combine_weights_in` | 49.7 ms [49.0, 51.0] | 32.8 ms [31.4, 39.1] |
| repr construction (`new` → `from_tables`) | 25.2 ms [24.3, 27.4] | 13.7 ms [13.2, 14.0] |
| `into_sumcheck` + repr construction | 83.0 ms | 55.2 ms (−33%) |

Whole-proof ranges overlap (1.88–2.03 s vs 1.89–1.96 s with the generic zerocheck backend), so the ~28 ms is best read from the spans. The weight accumulation is now bound by its 512 MiB output (first touch plus one read-modify-write pass; `compress_stacked` writes the same buffer in 8 ms), not by products, which is why it does not drop further.

## Test plan

- [x] `cargo test --release -p p3-binary-pcs -p p3-sumcheck -p p3-examples`
- [x] New: `combine_weights_in_the_field_itself_matches_the_challenge_field` (`R = EF` over `BabyBearExt4` and `BinaryField128`) and `combine_weights_in_the_polynomial_basis_matches_the_tower` (`BinaryField128 → Ghash128`), on a fixture with claims carrying 3, 2, 1 and 0 contributions per column, successor openings and one virtual claim
- [x] `cargo nextest run -p p3-sumcheck -p p3-binary-pcs` (debug), `cargo test -p p3-whir --release`
- [x] `cargo clippy -p p3-sumcheck -p p3-binary-pcs --all-targets -- -D warnings`, nightly `fmt`


sumcheck/src/layout/prover/suffix.rs
```diff
@@ -666,11 +817,15 @@ impl<F: Field, EF: ExtensionField<F>> SuffixProver<F, EF> {
...
+                .as_slice()
                 .split_at(claim.point.num_variables() - rs.num_variables());
             // Scalar weight: alpha^i times the equality between svo part and rs.
-            let scale = alpha_i * Point::eval_eq(svo.as_slice(), rs.as_slice());
+            let scale = alpha_i * Point::eval_eq(svo, rs.as_slice());
+            // The equality table is a polynomial in the point, so it commutes with `R::from`.
+            let rest = Point::new(rest.iter().copied().map(R::from).collect());
             // Contribute the scaled equality table across the whole output.
-            SplitEq::new_packed(&rest, scale).accumulate_into(out.as_mut_slice(), None);
+            SplitEq::<B, R>::new_packed(&rest, R::from(scale))
```
tcoratger, line 835: This is the first thing to route the PCS's `SplitEq` through `Ghash128`'s own packing, and no CI leg that runs tests covers it.

`Ghash128::Packing` is `PackedGhash128` only under `x86_64 + vpclmulqdq + (avx2 | avx512f)`, width 2 or 4. In ci.yml the only jobs with `vpclmulqdq` are `avx2+vpclmulqdq` and `avx512f+vpclmulqdq`, both **build only**; `linux, pclmulqdq` runs tests but `pclmulqdq` is not `vpclmulqdq`, so it gets width 1. On the M4 Pro you measured on it's width 1 too. So before this PR the binary path was `SplitEq<BinaryField128, BinaryField128>` at width 1 everywhere; now it's width 2/4 on x86, untested.

I ran it on a 9950X3D:

```
RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq"                        cargo test --release -p p3-sumcheck -p p3-binary-pcs --features parallel   -> all pass
RUSTFLAGS="-C target-feature=+avx512f,+vpclmulqdq,+gfni,+avx512bw"     cargo test --release -p p3-binary-field -p p3-sumcheck -p p3-binary-pcs --features parallel   -> all pass
```

So it works today. Flipping one of those two legs from build-only to running tests over `p3-binary-field p3-sumcheck p3-binary-pcs` would keep it that way.

sumcheck/src/layout/prover/suffix.rs
```diff
@@ -619,8 +744,34 @@ impl<F: Field, EF: ExtensionField<F>> SuffixProver<F, EF> {
             }
         }
 
+        (tables, column_weights)
+    }
+
+    /// Accumulates a weight plan into the residual weight polynomial over `R`.
+    ///
+    /// `B` is the base field the virtual claims' factored equality tables are packed over.
```
tcoratger: `B` picks which field the virtual-claim table packs over, and `combine_weights_in` collapses it to `R`. That happens to be a gain for you - `Ghash128::Packing` is width 2/4 on x86 against `BinaryField128::Packing = Self` - but it goes the other way for a prime field: `BinomialExtensionField::Packing = Self`, so `combine_weights_in::<BabyBearExt4>` packs at width 1 where `combine_weights` packs over `BabyBear` at width 8 or 16.

Nothing calls it that way today, and the test that does only checks values. But the signature invites it, so worth saying here that `B` should be the narrowest base `R` extends, not `R` itself.

tcoratger (approved):
The isomorphism argument holds everywhere I pushed on it.

Concrete claims are easy: tables built at unit scale in `EF`, converted, then `sum_j R(scale_j) * R(w_j[i]) = R(sum_j scale_j * w_j[i])` because `R::from` is a ring map. The virtual claims are the part that needed care, and crossing the point *before* building the equality table is right - the table is a polynomial in the coordinates, and `Ghash128::from` fixes 0 and 1, so `eq_R(R(z), x) = R(eq_EF(z, x))`. Sharing `preprocess` makes the transcript identical by construction rather than by inspection, which is the right way to get that guarantee. And `SumcheckProver::new`'s debug dot-product check now runs in `R`, so it validates the crossing itself, not just the claim - nice side effect of `from_tables`.

Also: the per-column `Vec` I flagged on #2165 is gone here, matching on `column_weights[..][..]` directly with `resolve` inside the arms. That was the fix I had in mind.

Two notes below, both about which packing you end up on. Neither is a bug. Approving.

Worth knowing on the conversion side: the bulk crossing is now one full-size table instead of two, and everything still in `EF` is small - the per-claim tables are ~2^11 entries each against a 2^22 output, so the tower arithmetic left in `weight_plan` is noise. `into_evals` is a plain move, so `Poly<EF> -> Vec<EF> -> in-place blocked convert -> Vec<R> -> Poly<R>` copies nothing.

PR #2180, opened 2026-09-17
perf(examples): allocate through jemalloc with transparent huge pages on Linux

Nashtare:
## Summary

The example provers touch several gigabytes of fresh memory per proof and release them again (a 2 GiB codeword, the half-gigabyte witness and weight tables, the halves each sumcheck round binds into). With the system allocator on a 4 KiB-page Linux host, first-touch page faults and the unmapping on release dominate: on a 64-thread c8a.16xlarge running the combined open-PR branch, `prove_keccak_binary -l 14` spends more than half of its time in `do_user_addr_fault`, `lock_vma_under_rcu`, TLB shootdowns and the `munmap` of its buffers, and the same cost shows up as unattributed "self" time in the `binary pcs open`, `binary pcs commit` and fourth-sumcheck-round spans. Setting `/sys/kernel/mm/transparent_hugepage/enabled` to `always` halves the proof time, which is the diagnosis but not a fix a binary can ship.

On Linux the examples crate now installs jemalloc as its global allocator, configured through `_rjem_malloc_conf` with `thp:always` (extents backed by transparent huge pages through `madvise`, so it works under the distribution default of `madvise`), `metadata_thp:auto` and `retain:true` (freed extents stay mapped for the next buffer instead of being unmapped and faulted in again). Other targets keep the system allocator: macOS has no huge pages to offer and measured 3–8% slower through jemalloc, and MSVC is unsupported by jemalloc.

## Measurements

c8a.16xlarge (AMD EPYC 9R45, 64 vCPU, `avx512f`/`avx512bw`/`gfni`/`vpclmulqdq`), Ubuntu with THP `madvise` (default), `--profile optimized`, `-C target-cpu=native`, `--features parallel`, the combined open-PR branch, same build flags on both sides, 3 runs.

| Example | system allocator | jemalloc | 
| --- | ---: | ---: |
| `prove_keccak_binary -l 14 --security-bits 96`, `prove` span | 648–684 ms | 278–290 ms (−57%) |
| — `combine_weights_in` | 78 ms | 29 ms |
| — sumcheck round 4 (first round after the in-place tables are dropped) | 39–45 ms | 7 ms |
| — `binary pcs open` self time | 84 ms | ~2 ms |
| `prove_prime_field_31 --field koala-bear --objective keccak-f-permutations -l 14 --merkle-hash keccak-f`, `prove` span | 79–85 ms | 58–61 ms (−27%) |

Apple M4 Pro (macOS, 16 KiB pages, no THP), jemalloc vs system: Keccak `-l 14` 2.04 → 2.13 s, BLAKE3 `-l 10` 0.80 → 0.84 s (3 runs each, +3–8%), hence the Linux-only gate.

## Test plan

- [x] `cargo check -p p3-examples`, nightly `fmt`
- [x] Linux: the two examples above run and verify through jemalloc on the c8a; `nm` shows `_rjem_malloc_conf` exported
- [x] macOS: builds and runs with the system allocator (the dependency is not compiled there)


tcoratger (approved):
Just the CI to fix :)

PR #2182, opened 2026-09-17
feat(examples): choose the Merkle arity of the binary examples

Nashtare:
## Summary

The binary hash examples commit through a binary Keccak-256 Merkle tree. A four-child node's 128 bytes of children still fit one Keccak-256 block, so a 4-ary tree needs a third of the compressions for the same leaves, at the cost of longer authentication paths (three siblings per level over half as many levels). It is a prover-time versus proof-size trade, so it is an opt-in knob: `BinaryProofOptions::merkle_arity` (2 or 4, default 2) and `--merkle-arity` on both examples, rejected outside `{2, 4}` by the CLI.

The harness types take a `const N: usize` (`Compress<N>`, `MerkleMmcs<N>`, `Mmcs<N>`, `BinaryStarkConfig<N>`, `binary_config::<N>`); `prove_binary_air` dispatches on the option to a generic inner function. `BinaryProofError` projects the PCS errors through `BinaryStarkConfig<2>`, which is sound because neither the commitment nor the error types of the chain depend on `N` (documented at the enum). A test proves the same tiny AIR at both arities and checks the proof sizes differ.

## Measurements

c8a.16xlarge (EPYC 9R45, 64 vCPU), combined open-PR branch with #2180/#2181, `--profile optimized`, `-C target-cpu=native`, `--features parallel`, `--security-bits 96`:

| Example | arity 2 | arity 4 |
| --- | ---: | ---: |
| `prove_keccak_binary -l 14`, prove | 264 ms | 242–246 ms (−7%) |
| — base Merkle tree (`commit_matrix`) | 34.5 ms | 24.4 ms |
| — proof size | 466 065 B | 580 899 B (+25%) |
| — verify | 21 ms | 19–20 ms |
| `prove_blake3_binary -l 10`, prove | 193 ms | 186 ms (−4%) |
| — proof size | 594 481 B | 690 732 B (+16%) |

Apple M4 Pro (on #2175 + #2176, medians of 3): Keccak `-l 14` 1.275 → 1.198 s (−6%), BLAKE3 `-l 10` 0.621 → 0.589 s (−5%), same proof sizes.

## Test plan

- [x] `cargo test --release -p p3-examples` and debug, including `proof_size_differs_between_merkle_arities`
- [x] Both examples at `--merkle-arity 2` and `4` verify; `--merkle-arity 3` is rejected by the CLI
- [x] `cargo clippy --release -p p3-examples --no-deps --all-targets -- -D warnings` (the full `--all-targets` run hits the pre-existing `clippy::extra_unused_lifetimes` in `batch-stark` on `main`), nightly `fmt`


examples/src/binary.rs
```diff
@@ -221,22 +239,45 @@ impl<A> BinaryAir for A where
...
+where
+    A: BinaryAir,
+{
+    match options.merkle_arity {
+        2 => prove_binary_air_with::<A, 2>(air, trace, options),
+        4 => prove_binary_air_with::<A, 4>(air, trace, options),
+        other => panic!(
+            "unsupported Merkle arity {other}: the binary-field harness only builds 2-ary and \
+             4-ary trees"
+        ),
```
tcoratger: Maybe we should have a proper error here in BinaryProofReport instead of panicking no?

tcoratger (approved):

PR #2184, opened 2026-09-18
perf(sha256): hash four messages at once with the ARMv8 SHA-2 extension

### LINKED ISSUE #2168: perf(sha256): hash several messages at once with the ARMv8 SHA-2 extension
`sha256/src/x86_64_sha_ni.rs` hashes four messages at once using the x86-64 SHA-NI extension (#2152). AArch64 has no counterpart, so on Apple silicon the crate falls back to the portable compression.

## The instructions exist on ARM

ARMv8's SHA-2 extension provides the round and schedule instructions, exposed in Rust under `target_feature = "sha2"`:

```
    vsha256hq_u32     vsha256h2q_u32      the round pair
    vsha256su0q_u32   vsha256su1q_u32     the message schedule
```

Every Apple M core has them, and the workspace already uses this feature block elsewhere — `keccak/src/neon_sha3.rs` reaches for `EOR3` from the neighbouring `sha3` block, so the gating pattern is established.

## What is open rather than settled

Whether the four-message interleave that pays on SHA-NI is also the right shape here. The two extensions have different dependency structure and different latency and throughput, so the multi-buffer width that hides SHA-NI's latency may be the wrong choice on an M core. Measure one, two and four messages in flight rather than porting the x86 shape unexamined — the answer is a property of the pipeline, not of the algorithm.

## What done looks like

- Gated on `target_feature = "sha2"`, with the portable path as the fallback and as the reference.
- A proptest pinning the digests byte for byte against the portable implementation, plus the published SHA-256 vectors, **executed on real hardware**.
- Benchmarks from `sha256/benches/sha256.rs` reported on Apple silicon, before and after, at whatever multi-buffer width the measurement selects.

## Why this needs a Mac

This machine is x86-64. There is no `qemu-aarch64` and no cross linker available, and the repository's arm64 CI leg is build-only, so proptests never execute there. A hash kernel that has never run is not a hash kernel.

Counterpart to #2152. Related AArch64 gaps: #2163.

gopikannappan:
Closes #2168.

This brings the batched SHA-256 path that #2152 added for x86 to Apple silicon and other ARM chips. `Sha256::hash_many` now hashes four messages side by side using the ARM SHA-2 instructions, which every Apple M chip has. On an M2 that makes batched hashing about 1.2–1.3× faster.

Nothing needs configuring. Apple targets turn these instructions on by default, and ARM builds without them keep the existing one-message-at-a-time path.

## Results

Apple M2, `cargo bench -p p3-sha256`:

| Benchmark | Before | After | Speedup |
|---|---:|---:|---:|
| `hash_many`, 32-byte messages | 104.6 µs | 87–91 µs | 1.16–1.20× |
| `hash_many`, 64-byte messages | 198.4 µs | 157–159 µs | 1.25–1.26× |
| `hash_many`, 256-byte messages | 510.1 µs | 392–394 µs | 1.29–1.30× |
| `compress_many` | 80.5 µs | 78–80 µs | no change |

The "After" ranges come from two separate runs.

@Nashtare confirmed this on an **M4 Pro**: `hash_many` is 1.15× / 1.12× / 1.29× faster at 32 / 64 / 256 bytes, with `compress` at parity.

**End to end.** Nashtare also measured what this means for a whole proof: with four lanes on Apple silicon, SHA-256 Merkle trees now take the batched path. On `keccak-air`'s `prove_baby_bear_sha256` (M4 Pro, serial), trace commitment goes from 1090–1140 ms to 753–798 ms (**1.45×**), and the whole proof from 1680–1740 ms to 1340–1380 ms (**1.25×**). `prove_baby_bear_sha256_compress` behaves the same (1.51× / 1.27×). Verification is unchanged.

**One regression, now fixed.** This PR made the untouched one-at-a-time path slower for 32-byte messages, because `hash_iter_slices` stopped being inlined into callers outside the crate. Nashtare traced it (76.6 → 93.8 µs on the M4 Pro). I'd seen it on the M2 too (105.3 → 111.4 µs) but mistook it for run-to-run noise. Marking `hash_iter_slices` `#[inline]` brings it back to 105–106 µs on the M2, and the batched numbers don't move.

## Why four at a time, and why `compress_many` doesn't get faster

The issue asked whether four messages in flight, the right choice for x86 SHA-NI, is also right on ARM. It is, but for a different reason.

On x86, a single SHA-NI stream leaves the hardware idle most of the time, so interleaving four streams fills those gaps. On the M2 there's hardly any gap to fill. The `sha2` crate already uses the same ARM instructions, and the CPU overlaps independent blocks by itself. Measured on the raw kernel:

- If every block depends on the one before it, hashing takes about 31 ns per block.
- Independent blocks, hashed one at a time, already run at about 18 ns per block.
- Interleaving 1 to 8 streams by hand doesn't beat that 18 ns.

So `compress_many`, which hashes independent blocks, has nothing to gain. On `hash_many`, most of the speedup for 64- and 256-byte messages comes from Plonky3's shared padding code avoiding the `sha2` crate's per-message overhead. Running four streams together adds a few percent on top, and it provides all of the gain for 32-byte messages. I tried one, two and four in flight. Four was fastest or tied at every length. Two was slightly slower than one on 64- and 256-byte messages.

<details>
<summary>Full sweep: one, two and four messages in flight</summary>

Median µs from criterion, two runs each:

| In flight | `hash_many/32` | `hash_many/64` | `hash_many/256` | `compress` |
|---|---:|---:|---:|---:|
| `main` (no ARM path) | 104.6 | 198.4 | 510.1 | 80.5 |
| 1 | 105.0 / 106.7 | 165.0 / 165.2 | 402.9 / 403.6 | 82.5 / 83.0 |
| 2 | 94.8 / 95.3 | 171.6 / 171.4 | 414.1 / 415.4 | 82.4 / 82.2 |
| 4 (this PR) | 90.5 / 87.1 | 157.0 / 158.5 | 394.2 / 391.8 | 79.8 / 78.4 |

For the sweep, `compress` was temporarily rewritten with index loops so the width could be a constant. The code in this PR was benchmarked separately against the same baseline and landed in line with the four-in-flight row (85.5 / 157.3 / 401.7 µs on `hash_many`).

Raw-kernel probe, 64-byte blocks: dependent chain ~31.3 ns/block, independent blocks via `sha2` ~18.3 ns/block, explicit interleave at 1/2/3/4/6/8 streams 18.3–19.4 ns/block. Every width was bit-exact against `sha2`.
</details>

## Testing

All tests were run on the M2.

The existing tests compare against the `sha2` crate. On ARM, though, that crate uses the same hardware instructions as this code, so a mistake shared by both wouldn't be caught. To cover that, I added a plain-Rust SHA-256 written straight from the FIPS 180-4 standard, with its own copy of the round constants. Both batched functions are checked against it on random inputs, on edge-case blocks, and on the published FIPS test vectors in every lane.

To confirm these tests can actually fail, I planted two bugs, one in the new ARM code and one in the shared padding code. The new tests caught both.

- 17 of 17 tests pass in debug and release builds, and also with the SHA-2 instructions turned off.
- The x86 (default and SHA-NI) and wasm builds still compile, and the x86 default tests pass under Rosetta. The SHA-NI tests can't run on a Mac, because Rosetta doesn't emulate those instructions. This PR only touches shared config lines on that path, not the x86 code.

<details>
<summary>New tests</summary>

- `hash_many_matches_specification_on_random_batches`: random message lengths (0–400 bytes) and batch sizes (1–17), checked against the standard-based reference, including padding
- `compress_many_matches_specification_on_random_batches`: random blocks against the reference
- `compress_many_matches_specification_at_boundary_blocks`: all-`0x00`, all-`0xff`, all-`0x80` and all-`0x7f` blocks
- `hash_many_matches_fips_180_vectors_in_every_lane`: the empty message, `abc`, and the 448- and 896-bit FIPS vectors, four copies each, including padding that spills into a second block

`reports_four_lanes_only_on_vectorized_targets` now expects four lanes on ARM with SHA-2, which confirms the new path is actually the one that runs.
</details>

All numbers come from one machine. An M4 or newer would be a useful second data point.


sha256/src/lib.rs
```diff
@@ -85,6 +104,11 @@ impl CryptographicHasher<u8, [u8; 32]> for Sha256 {
...
             target_feature = "sha",
             target_feature = "sse4.1"
+        ),
+        all(
+            target_arch = "aarch64",
+            target_feature = "neon",
+            target_feature = "sha2"
         )
     ))]
     fn hash_many(&self, input: &[u8], out: &mut [[u8; 32]]) {
```
Nashtare, line 117: This makes the one-at-a-time path slower for short messages. The bench's own `one at a time/32` row, which is a `Sha256.hash_slice` loop this PR doesn't touch, goes from 76.6 µs on main to 93.8 µs here on the M4 Pro (fastest of 9 runs). The 64- and 256-byte rows are unchanged.

The cause is inlining. The scalar tail of `four_lane::hash_many` now instantiates `hash_iter_slices::<Once<&[u8]>>` inside this crate, and after that the bench's loop calls it out of line for every message (`bl <…hash_iter_slices::<Once<&[u8]>>>`). On main it is fully inlined and that symbol doesn't appear in the binary.

Marking `hash_iter_slices` (line 90) `#[inline]` fixes it. The row comes back to 76.8 µs and the batched rows don't move (66.4 vs 66.7 µs). The x86 SHA-NI build likely has the same issue already, since it shares this tail.

sha256/src/lib.rs
```diff
@@ -1,9 +1,10 @@
 //! The SHA2-256 hash function.
 //!
 //! [`Sha256::hash_many`] and [`Sha256Compress::compress_many`] hash four messages at a time on
-//! the targets that have a four-lane backend: wasm32 with `simd128`, and x86-64 with `sha` and
-//! `sse4.1`. The choice is made at compile time, so an x86-64 build needs `-C target-feature=+sha`
-//! or `-C target-cpu=native` to get it; no microarchitecture level turns `sha` on by itself.
+//! the targets that have a four-lane backend: wasm32 with `simd128`, x86-64 with `sha` and
+//! `sse4.1`, and AArch64 with `neon` and `sha2`. The choice is made at compile time, so an x86-64
+//! build needs `-C target-feature=+sha` or `-C target-cpu=native` to get it; no microarchitecture
+//! level turns `sha` on by itself. The Apple silicon targets enable `sha2` by default.
```
Nashtare: Worth covering Linux AArch64 here as well. `rustc --print cfg --target aarch64-unknown-linux-gnu` enables only `neon`, so Graviton, Ampere and Cortex-A builds stay on the one-message path unless they pass `-C target-feature=+sha2` or `-C target-cpu=native`. That's the same situation as x86 with `+sha`. Line 9's "detects SHA-NI at runtime" is x86-specific too, since on AArch64 `sha2` detects the ARM extension. Something like: "…an x86-64 build needs `+sha` and an AArch64 Linux build needs `+sha2` (or `-C target-cpu=native`); the Apple silicon targets enable `sha2` by default."

sha256/src/lib.rs
```diff
@@ -425,6 +551,53 @@ mod tests {
         }
     }
 
+    #[test]
+    fn hash_many_matches_fips_180_vectors_in_every_lane() {
+        // FIPS 180-4 examples: the empty message, one block, and the two messages whose padding
```
Nashtare: Only the 448-bit (56-byte) vector's padding spills. The 896-bit one is 112 bytes, so its last block holds 48 message bytes, then `0x80`, then the 8-byte length, and that fits in 64. It takes two blocks because of its length, not because of padding. Maybe: "…one whose padding spills into a second block, and a two-block message."

sha256/src/lib.rs
```diff
@@ -26,6 +27,19 @@ mod wasm32_simd128;
...
+    target_feature = "sha2"
+))]
+mod aarch64_sha2;
+
+#[cfg(all(
+    target_arch = "aarch64",
+    target_feature = "neon",
+    target_feature = "sha2"
+))]
+use aarch64_sha2::ArmSha2 as Backend;
```
Nashtare, line 48: Nit: this lands above the doc comment that heads the backend-selection `use`s, with no blank line in between. Moving it after the x86 `use` would keep the order wasm, x86, aarch64, the same as the `mod`s above.

Nashtare (approved):
Approving. I checked the kernel against the ACLE round/schedule pattern: `sha256h2` gets the pre-update `abcd`, the `su0`/`su1` operands are right, and the `g % 4` slot rotation is right. I ran everything on an **M4 Pro**, the second data point the description asks for. Clippy `-D warnings` and fmt are clean. 17/17 tests pass in debug, in release, and with `-sha2`, where the fallback correctly reports 1 lane. I planted 9 mutants (round pair, schedule operands, feed-forward, byte order in and out, slot off-by-one, lane crossing, padding boundary, round constant) and the suite caught all 9. The macOS CI leg also runs the kernel on real hardware.

**M4 Pro, `cargo bench -p p3-sha256`** (fastest run per config, main vs this head):

| | main | PR | |
|---|---:|---:|---:|
| `hash_many` 32 / 64 / 256 B | 75.9 / 148.4 / 431.3 µs | 66.1 / 133.0 / 333.6 µs | 1.15× / 1.12× / 1.29× |
| `compress` batched | 66.9 µs | 66.8 µs | parity |

**End to end**, which the description doesn't cover. With `LANES = 4` on Apple, SHA-256 Merkle trees now take the batched drivers. On `keccak-air` `prove_baby_bear_sha256`, serial, 4 alternating runs, trace commitment goes from 1090–1140 ms to 753–798 ms (**1.45×**) and the whole proof from 1680–1740 ms to 1340–1380 ms (**1.25×**). `prove_baby_bear_sha256_compress` behaves the same (1.51× / 1.27×). Verify is unchanged. That seems worth adding to the description.

A few inline comments. Only the `#[inline]` one has a measurable effect. Two more that have no line in the diff:
- No CI job runs clippy on `aarch64_sha2.rs`. The lint job is x86, the macOS leg only tests, and the arm64 build-only leg has `+sve2` but not `+sha2`. It's clean locally, so this is only a possible follow-up.
- The `four_lane` doc still says a padding fix cannot "miss the other" backend, but there are three now.

gopikannappan:
Thanks @Nashtare, especially for the M4 Pro run, the mutants and the end-to-end numbers. I've added those to the description, with credit. All points are addressed in 8c38c2e3:

- **`#[inline]` on `hash_iter_slices`.** You were right, and I had seen it too: on the M2 the one-at-a-time 32-byte row went from 105.3 to 111.4 µs and I'd put it down to noise. With the attribute it's back to 105–106 µs, and the batched rows don't move.
- **Crate docs** now say AArch64 Linux needs `+sha2` (or `target-cpu=native`), and that `sha2` detects SHA-NI or the ARMv8 SHA-2 extension at runtime.
- **`four_lane` doc** now says "the others".
- **FIPS vector comment** now says only the 448-bit message's padding spills, and the 896-bit one is a two-block message.
- **`use` order:** I tried moving it after the x86 `use`, but `group_imports = "StdExternalCrate"` sorts it back to the top even across a blank line. So I moved the backend-selection doc comment onto the first import instead. It now heads the block whatever order rustfmt picks.

I've left the clippy-on-`aarch64_sha2.rs` CI gap for a follow-up, since it's your CI to shape. It's clean locally under `-D warnings`. 17/17 still pass in debug, in release and with `-sha2`.

PR #2185, opened 2026-09-18
examples: merge binary-hash examples, expose NTT/representation choice

Nashtare:
## Summary
- Merges `prove_keccak_binary` and `prove_blake3_binary` into a single `prove_hash_binary` example, selecting the objective via `--objective` (clap-derived CLI, in the style of `prove_prime_field_31`).
- Exposes two previously-hardcoded choices as CLI flags: `--ntt` (`poly-basis`/`lch`/`naive`, the additive NTT encoding the binary-PCS codeword) and `--representation` (`auto`/`subfield`/`poly-basis`, the field representation the zerocheck prover's later rounds run in). Every representation/NTT choice proves and verifies the same statement; this is a performance axis only, backed by the existing `assert_backends_prove_byte_for_byte` tests.
- To make the NTT choice real (not just cosmetic), `BinaryPcs` gains a generic `Ntt` parameter defaulting to `PolyBasisNtt` (so every existing caller — tests, benches, `multi-stark` examples — is unaffected without any changes on their part) plus a new `BinaryPcs::with_ntt` constructor. `AdditiveRsEncoder::new` was added alongside it.
- `examples::binary` gets a new `AdditiveNttChoice` enum (dispatches to `PolyBasisNtt`/`LchNtt`/`NaiveAdditiveNtt`, mirroring the existing `DftChoice` pattern in `examples::dfts`) and makes the existing `Backend` enum (`Subfield`/`PolyBasis`) public so the example can select it explicitly.
- `prove_binary_air` keeps its original signature/behavior; `prove_binary_air_with_ntt` and `prove_binary_air_with_ntt_and_backend` are new, additive entry points.
- README updated to document the merged example and its new flags.

## Test plan
- [x] `cargo test -p p3-binary-pcs -p p3-binary-dft -p p3-examples --lib` and `cargo test -p p3-binary-pcs --tests` pass
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo +nightly fmt --check` clean on all touched crates
- [x] Manually ran `prove_hash_binary` for both objectives, all three `--ntt` choices, both `--representation` choices, and both `--merkle-arity` values; verified proofs, and confirmed `--representation` choices give byte-identical proof size/security as expected
- [x] Confirmed `multi-stark`'s existing `BinaryPcs` consumers (tests, benches, `prove_binary_field`/`bench_binary_grouping` examples) still compile unchanged

https://claude.ai/code/session_01UHxJxc4uKiBiHYD1a1Pg9V


examples/src/parsers.rs
```diff
@@ -43,6 +43,26 @@ pub enum PcsOptions {
     Stir,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum BinaryHashOptions {
+    Blake3Compressions,
+    KeccakFPermutations,
+}
```
tcoratger, line 50: We should maybe open an issue to have SHA as well in a followup?

tcoratger (approved):

examples/src/parsers.rs
Nashtare: Opened in #2186

PR #2187, opened 2026-09-18
ci: split Heavy CI into per-job workflows so path filters actually scope per job

Nashtare:
## Summary

Prompted by PR #2185: an examples-only change fired the WHIR exhaustive tests job, which has nothing to do with `p3-examples`.

`ci-heavy.yml` bundled three unrelated jobs — `whir_exhaustive` (tests only `p3-whir`), `binary_pcs_large_configuration` (tests only `p3-binary-pcs`), and `slow_regressions` (tests `p3-multilinear-util`, `p3-multi-stark`, `p3-whir`, `p3-uni-stark`, `p3-binary-dft`, `p3-stir`, `p3-examples`) — behind one shared `on.pull_request.paths` list. GitHub Actions path filters gate the whole workflow file, not individual jobs, so any path any *one* job needed (e.g. `examples/src/**`, added for `slow_regressions`' example-based tests) fired *every* job in the file, including two that don't exercise `p3-examples` at all — `p3-examples` depends on `p3-whir`/`p3-binary-pcs`, never the reverse.

- Confirmed via git history (`fffd3e72d`, `6ea70c48b`, `90e06b98b`, `8f908fc4b`) that each path was added for a specific job, then folded into the shared list rather than scoped to it.
- Split into three workflow files, one per job, each with its own `paths:` scoped to what that job's `-p <crate>` test target and its real dependencies are:
  - `ci-heavy-whir.yml`: `whir/src/**` only.
  - `ci-heavy-binary-pcs.yml`: `binary-pcs/**` plus `binary-dft/src/**` — the latter wasn't previously job-specific to this test, but `p3-binary-pcs` genuinely depends on `p3-binary-dft` for its codeword encoder (this PR's own `robin/merge_binary_hash_examples` touches both), so it's a deliberate correctness addition, not just a mechanical split.
  - `ci-heavy-slow-regressions.yml`: the remaining paths (`circle`, `fri`, `batch-stark`, `multilinear-util`, `multi-stark`, `uni-stark`, `binary-dft`, `stir`, `examples`), plus `whir/src/**` — two of its tests live in `p3-whir` and previously relied on the shared list to fire on WHIR changes; dropping it without adding it back here would have been a silent coverage loss.
- Every job body (steps, env, matrix) is byte-for-byte unchanged — verified by diffing each extracted job against the original file. Only the trigger scoping and per-workflow `name:` (for clarity in the Actions UI) changed.
- None of these three jobs are required status checks on `main` (confirmed via the branch protection API), so this doesn't touch anything merge-gating.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open(...))"` on all three new files
- [x] Diffed each new file's job body against `git show HEAD:.github/workflows/ci-heavy.yml`'s corresponding job — identical
- [x] Confirmed via `gh api repos/Plonky3/Plonky3/branches/main/protection` that none of `WHIR exhaustive tests`, `Slow regression tests`, `binary-pcs 2^16 round trip` are required checks
- [ ] Will confirm on this PR's own CI run that only the jobs relevant to `.github/workflows/**` fire, and that a follow-up push touching e.g. `examples/src/**` only fires `Heavy CI (slow regressions)`

https://claude.ai/code/session_01UHxJxc4uKiBiHYD1a1Pg9V


tcoratger:
I was wondering if we shouldn't take this a step further and run CI only for the crates affected by the PR (or, to be a bit safer, the affected crates plus those that depend on them).

For instance, if we modify the multilinear crate, there’s no reason to run CI for the field or Poseidon crates, since we know it won't have any impact. I mean running all tests—not just the heavy ones—for those specific crates; what do you think?

In theory, we should be able to run CI only for the impacted crate; if everything is properly tested, the tests within that crate would catch any bugs without needing to run CI for any other crates (even the one on top).

What’s your take on that?

Nashtare:
> I was wondering if we shouldn't take this a step further and run CI only for the crates affected by the PR (or, to be a bit safer, the affected crates plus those that depend on them).
> 
> For instance, if we modify the multilinear crate, there’s no reason to run CI for the field or Poseidon crates, since we know it won't have any impact. I mean running all tests—not just the heavy ones—for those specific crates; what do you think?
> 
> In theory, we should be able to run CI only for the impacted crate; if everything is properly tested, the tests within that crate would catch any bugs without needing to run CI for any other crates (even the one on top).
> 
> What’s your take on that?

That sounds sane in principle, but I'm afraid that some dependencies may be subtle and missed (or introduced by some future PRs) in a way that isn't caught and can cause nasty bugs. Unless we can find a really safe way to do so, I'd rather have the whole test suite run regardless (with isolated known heavy tests running independently and on-demand).

tcoratger (approved):

PR #2188, opened 2026-09-18
perf(sumcheck)!: bind the stacked rows over one aggregated column

Nashtare:
## Summary

The binary PCS runs its residual sumcheck over two stacked-size tables: the witness, and the batched weights expanded across every column. For the binary AIRs those weights factor. Every opened column's weights are one shared row table times a scale of its own:

```text
Keccak:  W(c, x) = alpha^c * (E(x) + alpha^C * N(x))     current and successor openings of every column
BLAKE3:  W(c, x) = alpha^c * E(x)                        current openings only
```

Suffix binding fixes the row variables first. This PR therefore binds them over the aggregate `A(x) = sum_c s_c * P_c(x)` and the shared row weights `T(x)`, both one column long. It builds neither stacked-size table. Once the rows are bound, each column is evaluated at the row point, and the remaining rounds run over one entry per column slot. The round polynomials are identical, so proofs are byte-identical.

## Changes

- `sumcheck/src/layout/prover/residual.rs` (new): `SuffixResidualProver`, which plays either route.
  - The row-first route aggregates the columns, plays the row rounds, and then hands off to the column slots.
  - Every product with a witness cell is taken in the residual field `R` (GHASH for the binary PCS) through `R: Algebra<F>`. This costs one basis change per product, where a tower product costs three.
- `SuffixProver::into_sumcheck_in` checks whether the weight plan splits (`shared_row_weights`) and picks a route. The plan is built once. A layout keeps the dense route when it has a single column, preprocessing rounds, virtual claims, tables of different heights, or columns whose weights are not proportional.
- `ReprSumcheckProver`: the per-round step is extracted as `round()`, with no change to its behaviour. Also adds `weights()` and `from_repr_tables()`.

**Breaking:** `SuffixProver::into_sumcheck_in` now returns `SuffixResidualProver` instead of `ReprSumcheckProver`. The one in-tree caller (`binary-pcs`) compiles unchanged.

## Performance

c8a.4xlarge (EPYC 9R45, AVX-512, native), `--security-bits 96`, 7 interleaved runs per side. Medians, with non-overlapping ranges:

| | main | this PR |
|---|---|---|
| Keccak l=14 prove | 556 ms | **490 ms (−11.9%)** |
| Keccak l=14 sumcheck (setup + 25 rounds) | 79 ms | 15 ms |
| BLAKE3 l=10 prove | 295 ms | **262 ms (−11.2%)** |

On an M4 Pro, Keccak's sumcheck share drops from 77 ms to 19 ms.

**Allocator caveat.** Under glibc's default allocator, a wide-table `open` bench (n = 20, 64 × 2^14 and 8 × 2^17) runs 8–10 ms slower. `perf` puts the extra time in first-touch page faults on the folded-codeword buffers. Main's dense route allocates and frees two stacked-size tables, and that raises glibc's dynamic mmap threshold, so the later folds reuse warm heap pages. With `MALLOC_MMAP_THRESHOLD_` pinned, the same bench is 10–13% faster with this PR. The examples use jemalloc on Linux (#2180) and do not see the effect. The `binary-pcs` benches use the system allocator.

## Testing

- Proofs are byte-identical to main, and verify, on M4 and on Zen 5:
  - Keccak l=14 and l=16
  - BLAKE3 l=10
  - arity-4 Merkle
  - folding 2
  - rate 1/2
- New tests play both routes round by round from equal transcripts and require identical challenges, claims, round messages and final sponge state. They run over `BinaryField128`/`Ghash128` and over `BabyBear`/`Ext4`, and cover these shapes:
  - successor openings
  - two claims
  - unopened columns
  - an unopened table of equal height
  - a `2^13`-row table
  - one-row tables
  - one to three rounds per call, with and without settling
- Mutating the handoff's point order, or the aggregate's non-bit branch, fails these tests in both debug and release.
- `aggregate_columns` is checked against row sums on both of its parallel paths. A separate test checks that every non-splitting layout, and a single column, keeps the dense route.
- `p3-sumcheck` and `p3-binary-pcs` pass serial and parallel, as do the `p3-examples` binary tests. `clippy -D warnings`, `rustdoc -D warnings` and nightly `fmt` are clean.


tcoratger (approved):
The row-first residual sumcheck is sound and preserves the existing transcript. I checked the factorization criterion, suffix challenge ordering, the deferred final row binding, the row-to-column handoff, zero-weight padding slots, and every dense fallback condition.

The targeted `p3-sumcheck` and parallel `p3-binary-pcs` suites pass locally, including exact row-first/dense transcript equivalence over both binary and prime-extension fields and the byte-identical proof test. I found no blocking correctness, soundness, or performance issue.

PR #2192, opened 2026-09-18
perf(field): scale packed extension coefficients in place when multiplying by a packed base

Nashtare:
`Mul<PF>` for `PackedBinomialExtensionField`, `PackedCubicExtensionField` and `PackedQuinticExtensionField` built its result with `self.value.map(|x| x * rhs)`. On x86 with AVX-512, LLVM does not reliably inline that `array::map`: the closure was called out of line for every coefficient, with operands passed through the stack. This PR scales the coefficients in place, so the product stays in registers. The result is identical, coefficient by coefficient.

`Matrix::columnwise_dot_product`'s parallel path does one `EF::ExtensionPacking * F::Packing` per matrix element, so this showed up there first. A build where the width was a compile-time constant, which let the closure inline, ran about 4× faster and pointed at the multiply.

## Measurements

`matrix/benches/columnwise_dot_product.rs`, extended to more fields (that bench change is not in this PR). c8a.4xlarge (AMD EPYC 9R45, `-C target-cpu=native`), criterion medians, change vs `main`:

| field / extension | shape | 1 thread | 4 threads | 16 threads |
| --- | --- | ---: | ---: | ---: |
| BabyBear⁴ | 2^16 × 128 | −64% | −58% | −50% |
| BabyBear⁴ | 2^20 × 256 | −67% | −62% | −41% |
| BabyBear⁴ | 2^20 × 8 | −42% | −37% | −37% |
| KoalaBear⁴ | 2^20 × 256 | −67% | −63% | −40% |
| KoalaBear⁵ | 2^20 × 256 | −67% | −61% | −47% |
| BabyBear⁴, two-point batched | 2^20 × 8 | −71% | −70% | −72% |
| Goldilocks², unbatched | 2^12..2^20 | within noise | within noise | within noise |
| QM31 | 2^12..2^20 | within noise | within noise | within noise |
| Goldilocks², two-point batched | 2^16..2^20 × 4 | +10% | +11..16% | +2..12% |

Across the 2^12–2^20 rows × 8–256 columns grid, every BabyBear⁴ / KoalaBear⁴ / KoalaBear⁵ cell is 6–67% faster. A second `main` run after the branch stayed within about ±3% (up to +8% in a few 16-thread cells).

The one slowdown is the two-point `columnwise_dot_product_batched` over Goldilocks². Its only production caller is circle's CFFT, which uses Mersenne31. On aarch64 (M4 Pro) the change is neutral.

## Test plan

- [x] `cargo test -p p3-field -p p3-baby-bear -p p3-koala-bear -p p3-goldilocks -p p3-mersenne-31` (1208 passed), aarch64. The same crates on x86 `-C target-cpu=native` (AVX-512).
- [x] `cargo test -p p3-matrix` (also with `--features parallel` at 1, 3 and 16 threads), `-p p3-fri -p p3-circle -p p3-stir -p p3-commit`
- [x] `cargo clippy -p p3-field --all-targets -- -D warnings`, nightly `fmt`


tcoratger (approved):

tcoratger:
The small slowdown for Goldilocks^2 is a bit strange but very close to noise and very small compared to the other gains.

PR #2193, opened 2026-09-18
fix(sumcheck): hand off a prefix residual narrower than one packed element unpacked

### LINKED ISSUE #2167: fix(multi-stark): packed prefix compression panics under target-cpu=native
Two security-checked multi-STARK tests panic on `main` when the build enables the host's vector features. They pass on a default build, which is why CI is green.

## Reproduce

On `main` at `7d12cd51`, clean checkout:

```
RUSTFLAGS="-C target-cpu=native" cargo test -p p3-multi-stark --test whir_fibonacci
```

```
test security_checked_whir_roundtrip_and_target_rejection ... FAILED
test security_checked_lookup_accounts_for_every_reduction ... FAILED

thread '...' panicked at multilinear-util/src/split_eq/mod.rs:358:9
```

The other two tests in that file pass. Without `-C target-cpu=native` all four pass.

## What the assertion is

The panic is the second assertion in the packed prefix compression:

```rust
assert!(
    poly.num_variables() >= (self.num_variables() + log2_strict_usize(F::Packing::WIDTH))
);
```

It says the polynomial must keep at least one full packed element's worth of variables after the prefix is fixed, which the routine's own doc comment states as a precondition.

## Why it only shows up with vector features

The guard is a function of the packing width, and on a default build the fallback packing has width one:

```
    default build   ->  WIDTH = 1  ->  log2(WIDTH) = 0  ->  assertion is vacuous
    native build    ->  WIDTH > 1  ->  log2(WIDTH) > 0  ->  assertion has teeth
```

So the caller has always been passing a polynomial too short for the packed route. On a default build the packed route degenerates to the scalar one and the mistake is invisible; with real vector registers it is a panic.

That also means this is a latent correctness question, not only a crash: the precondition exists because the routine writes whole packed elements, so any caller reaching it with too few remaining variables was relying on the width-one degeneracy.

## Where the fix belongs

At the call site, not in the assertion. Something in the security-checked WHIR path selects the packed compression without checking that enough variables remain; it should fall back to the scalar compression below that threshold. The two failing tests differ from the two passing ones in their configured sizes, so the boundary is reachable with ordinary parameters rather than being an extreme case.

Worth checking whether any other caller of the packed route makes the same assumption.

## Suggested guard against a repeat

No CI leg runs the test suite with the host's full feature set, so this class of bug is invisible to CI by construction. The matrix has `avx2`, `pclmulqdq` and `sha` legs that execute tests, and `avx512f`/`gfni` legs that are build-only. A leg that *runs* tests with a wide packing would have caught it, and would catch the next one — any assertion or kernel gated on `Packing::WIDTH` is untested today wherever the executing legs happen to pick width one.

Found while verifying an unrelated pull request: the failure reproduces identically on a pristine `main`, so it is not a regression from anything in flight.

gopikannappan:
Closes #2167.

The sumcheck prover could crash when only a few variables were left to process. This PR makes it fall back to the regular, non-SIMD code path in that case, in both the plain and the ZK prover, and adds tests that reproduce the crash.

## What was going wrong

After its first rounds, `PrefixProver` passes the remaining polynomial to a SIMD routine, `compress_prefix_to_packed`. That routine works on whole SIMD registers, so it needs at least `log2(register width)` variables left over: 2 on ARM NEON, 3 on AVX2 and 4 on AVX-512. The prover never checked that.

Without SIMD, the register width is 1, so the requirement is always met. That's why the default CI never saw the bug. With wider registers, a small enough input trips the routine's assertion and panics.

The issue found this on AVX-512, but it isn't limited to that. The same panic happens on an Apple M2 with a normal build, and on AVX2, just with smaller inputs than `whir_fibonacci` uses. CI couldn't catch it on any platform, including the macOS job that runs the tests on Apple silicon, because no existing test ever reached that case: they always leave at least 4 variables, which is enough for every register width.

## The fix

Before choosing the SIMD routine, the prover now checks that enough variables remain. If not, it uses the regular `compress_prefix` and builds the result with `ProductPolynomial::new_unpacked`, which the codebase already documents as the constructor "for polynomials too small for SIMD packing." Inputs big enough for SIMD take exactly the same path as before.

The issue also asked whether any other caller makes the same assumption. There is one: the ZK version of this handoff, which refused small residuals depending on the machine's SIMD width. Following review, both handoffs now go through one `PrefixProver::residual_product` helper (the ZK one passes its mask scale), so the ZK prover falls back too. That also removed the `prop_assume!(n_vars > k_pack)` workarounds in the ZK prover, verifier and simulator tests, whose prefix folding window is now `1..=n_vars` on every SIMD width. Thanks to @Nashtare for the review and @tcoratger for the follow-up commit.

## Testing

The new test builds the smallest case that triggers the bug: one variable left after the first rounds. It runs the handoff, finishes the sumcheck, and checks that the result equals the original polynomial evaluated at the same point. So it checks the answer is right, not only that nothing crashes.

| Register width | Machine | Before this PR | With this PR |
|---|---|---|---|
| 1 | x86, no SIMD (Rosetta) | not affected | all 521 sumcheck unit tests pass |
| 4 | Apple M2, NEON | crashes at `split_eq/mod.rs:358` | all 521 pass |
| 8 | x86 with AVX2 (Rosetta) | crashes at the same line | all 521 pass |
| 16 | AVX-512 | can't test here | can't test here |

The full `p3-sumcheck` and `p3-multi-stark` suites also pass on the M2, including `whir_fibonacci` with and without `-C target-cpu=native`.

For the ZK path, the widened ZK prefix property tests (prover completeness, simulator, and verifier tampering) fail when the old refusal is put back and pass with this change. There's also a dedicated test for the extreme case where every variable is folded. With the ZK change, `p3-sumcheck` passes 522 of 522 on the M2.

**AVX-512:** I couldn't test this width myself. @Nashtare verified it on a Zen 4 machine with `-C target-cpu=native`: the #2167 failures reproduce on the old code, and pass with this PR.

<details>
<summary>Lint note</summary>

Clippy is clean with `-D warnings` on nightly. On stable 1.95, `p3-binary-field` reports an unknown lint (`clippy::chunks_exact_to_as_chunks`) that only newer clippy versions recognize. That happens on `main` too and is unrelated to this change.
</details>


tcoratger (approved):

sumcheck/src/layout/prover/prefix.rs
```diff
@@ -328,12 +329,26 @@ impl<F: Field, EF: ExtensionField<F>> Layout<F, EF> for PrefixProver<F, EF> {
...
+            let compressed = tracing::info_span!("compress_prefix_to_packed")
+                .in_scope(|| self.poly.compress_prefix_to_packed(&rs, EF::ONE));
+            let weights = self.residual_weights_packed(&rs, alpha);
+            ProductPolynomial::<F, EF>::new_packed(VariableOrder::Prefix, compressed, weights)
+        } else {
+            let compressed = tracing::info_span!("compress_prefix")
+                .in_scope(|| self.poly.compress_prefix(&rs, EF::ONE));
+            let weights = self.combine_weights(&rs, alpha);
+            ProductPolynomial::<F, EF>::new_unpacked(VariableOrder::Prefix, compressed, weights)
+        };
```
Nashtare: This fallback is right, and `zk_residual_handoff` for `PrefixProver` (`zk/prover/layout.rs:54-76`) needs it too. It is the same handoff with `eps` in place of `EF::ONE`, and today it asserts the condition this branch now handles. Suggest one helper for both:

```rust
/// Builds the residual product polynomial left after the folded prefix rounds.
///
/// The compressed polynomial is scaled by `scale`; the weights use the batching challenge `alpha`.
pub(crate) fn residual_product(
    &self,
    rs: &Point<EF>,
    alpha: EF,
    scale: EF,
) -> ProductPolynomial<F, EF> {
    let k_pack = log2_strict_usize(<F as Field>::Packing::WIDTH);
    if self.num_variables() - rs.num_variables() >= k_pack {
        let compressed = tracing::info_span!("compress_prefix_to_packed")
            .in_scope(|| self.poly.compress_prefix_to_packed(rs, scale));
        let weights = self.residual_weights_packed(rs, alpha);
        ProductPolynomial::new_packed(VariableOrder::Prefix, compressed, weights)
    } else {
        let compressed = tracing::info_span!("compress_prefix")
            .in_scope(|| self.poly.compress_prefix(rs, scale));
        let weights = self.combine_weights(rs, alpha);
        ProductPolynomial::new_unpacked(VariableOrder::Prefix, compressed, weights)
    }
}
```

Then:
- `into_sumcheck` calls it with `EF::ONE` and `zk_residual_handoff` with `eps`.
- The assert at `layout.rs:63-67` goes away.
- The `prop_assume!(n_vars > k_pack)` reservations go away too (`zk/prover/zk_prover.rs:545`, `zk/verifier.rs:781`, `zk/simulator.rs:560`). The prefix folding window becomes `1..=n_vars` on every ISA.

I ran that change on NEON. `prop_completeness_prefix` over `folding_factor = 1 + seed % n_vars` panics at `layout.rs:64` without the fallback and passes 256/256 with it.

sumcheck/src/layout/prover/prefix.rs
```diff
@@ -328,12 +329,26 @@ impl<F: Field, EF: ExtensionField<F>> Layout<F, EF> for PrefixProver<F, EF> {
...
 
-        let weights = self.residual_weights_packed(&rs, alpha);
-        let prod_poly =
-            ProductPolynomial::<F, EF>::new_packed(VariableOrder::Prefix, compressed, weights);
+        // The packed compression writes whole packed elements.
+        // The residual must therefore span at least one of them.
+        //
+        //     packed route  <=>  num_variables - folding >= log2(Packing::WIDTH)
+        //
+        // A shorter residual is at most one packed element of scalars, so it hands off unpacked.
```
Nashtare: Two wording points:
- "A shorter residual is at most one packed element of scalars": it has fewer than `W` scalars, so it is *less* than one packed element.
- At equality (`residual == log2(W)`), this branch compresses into exactly one packed element, and `new_packed` → `transition()` unpacks it straight away (`k == 0`). So storage stays packed only *above* the threshold. `>=` is still the right route choice, since it keeps the SIMD compression pass. But the docs here and at lines 31-32 describe storage ("stays packed while it fills a packed element"). Describing the route instead would be exact: packed compression while the residual spans at least one packed element, scalar compression below that.

sumcheck/src/layout/mod.rs
```diff
@@ -57,7 +57,8 @@
 //! Lagrange weights; the modes differ only in the handoff to the residual
 //! product polynomial once folding completes:
 //!
-//! - Prefix-first binding: the handoff is packed (`compress_prefix_to_packed`).
+//! - Prefix-first binding: the handoff stays packed while the residual fills a packed element.
+//! - Narrower prefix residuals use scalar storage.
```
Nashtare: Nit: the second bullet reads like a third mode, between prefix and suffix. Folding it into the prefix bullet keeps the two-mode list intact:

```suggestion
//! - Prefix-first binding: the handoff compresses packed (`compress_prefix_to_packed`), or scalar
//!   (`compress_prefix`) when the residual is narrower than one packed element.
```

Nashtare (changes requested):
Requesting one small change. The fix itself checks out.

**Verified at AVX-512**, the width the PR couldn't test. Run on a c7a.4xlarge (Zen 4) with `RUSTFLAGS="-C target-cpu=native"`:
- On base `5021e1f19`, `cargo test -p p3-multi-stark --test whir_fibonacci` reproduces both #2167 failures at `split_eq/mod.rs:358`.
- On this PR, 35/35 of those tests pass and `p3-sumcheck` passes 521/521. The new test panics at the same assertion when `prefix.rs` is reverted.
- With this PR merged onto current `main` (`e45e11ea1`), `p3-sumcheck`, `p3-whir` and `p3-multi-stark` all pass.

The same checks agree on NEON (width 4). Clippy with `-D warnings` and nightly fmt are clean.

**The change I'm asking for: fall back in the ZK prefix handoff too.** You asked whether it should fall back or refuse. It should fall back.

`ZkLayout::zk_residual_handoff` for `PrefixProver` (`zk/prover/layout.rs:63-67`) refuses depending on the host's SIMD width. So the same ZK statement proves on a width-1 build and panics on NEON, AVX2 or AVX-512, which is #2167 again. The verifier never sees the packing width, so the protocol doesn't need the refusal. It only exists to support the ISA-dependent `prop_assume!(n_vars > k_pack)` reservations in `zk_prover.rs`, `zk/verifier.rs` and `zk/simulator.rs`.

I checked this locally on NEON:
- I widened `prop_completeness_prefix` to `folding_factor = 1 + seed % n_vars` and ran 256 cases.
- Today it panics at `layout.rs:64`.
- With your scalar fallback (`compress_prefix(rs, eps)` + `combine_weights` + `new_unpacked`) it passes, verifier included.

The two handoffs differ only in the compression scale, so one `PrefixProver` helper can fix both (inline comment below). This PR `Closes #2167`, and the issue asked about exactly this caller, so I'd rather the ZK fix land here than go untracked.

The other two comments are wording nits.

tcoratger (approved):
Implemented and verified the requested follow-up at 3eca599b. Plain and ZK prefix handoffs now share one packed-or-scalar residual helper; fully folded and sub-lane ZK residuals no longer depend on the host packing width. The sumcheck suite, downstream WHIR Fibonacci regression, native-SIMD zero-residual test, formatting, and clippy all pass.

gopikannappan:
Thanks @Nashtare for checking this on AVX-512, and for the ZK catch. My earlier probe only went through the WHIR ZK config, which never lets the residual get that small. Calling the sumcheck directly does, as you showed.

@tcoratger's commit 3eca599b covers every point in the review: the shared `residual_product` helper, the ZK fallback, the three removed `prop_assume!` reservations, and the wording, plus a test for the fully folded case. I ran it on an Apple M2 (NEON): `p3-sumcheck` passes 522/522, including that new test. I've updated the PR description to match.

Nashtare (approved):

PR #2195, opened 2026-09-18
perf(sumcheck,multilinear-util): fold suffix tables in place without the parallel feature

Nashtare:
In a build without the `parallel` feature, `p3-maybe-rayon`'s `par_*` iterators are plain sequential iterators. Two suffix folds still picked an allocating "parallel" path by table length alone. In those builds, which include plain WASM, the fresh half-length buffer buys nothing over the in-place fold that already exists:

- `sumcheck::strategy::fold_and_round_coefficients_suffix`, for `PAR_THRESHOLD < len < SUFFIX_IN_PLACE_THRESHOLD`
- `multilinear_util::Poly::fix_suffix_var_mut`, for `num_evals >= PARALLEL_THRESHOLD`

Both allocating branches are now gated on `p3_maybe_rayon::PARALLEL_ENABLED` (`cfg!(feature = "parallel")`), the compile-time idiom `matrix/src/util.rs` already uses for this kind of decision. Builds with `parallel` are unchanged, including a one-thread Rayon pool.

## Equivalence

Results are bit-identical: the in-place algorithms are unchanged, and this PR only routes more lengths to them.

- `fix_suffix_var_mut` sweeps forward: output `i` reads inputs `2i` and `2i + 1`, which are never behind the write position, so it is correct at any length.
- The leveled in-place pass in sumcheck was already the path for every build at `len >= SUFFIX_IN_PLACE_THRESHOLD`. `the_in_place_suffix_pass_matches_a_fresh_binding_with_threaded_levels` already cross-checks it against the fresh-buffer path at about 155k entries, inside the affected window.

Two new tests pin which path each feature configuration takes. They check pointer identity of the backing buffer, which cannot give a false positive: the allocating path builds the new `Vec` while the old one is still alive, so the allocator cannot hand back the same address.

## Memory

`truncate` keeps capacity, so in a build without `parallel` a table keeps its first allocation until it is dropped, instead of being right-sized every round. Every current caller holds one evals/weights pair at a time, so the single-table measurements below are representative. A future caller that keeps many suffix-bound tables alive at once should re-measure.

## Measurements

Apple M4 Pro, release, no `parallel` feature, alternating runs against a baseline binary built from `main`. The machine was shared with other builds, so the noise floor is about 3%, measured as the spread between two identical binaries.

| | size | peak RSS main → PR | time |
| --- | ---: | ---: | ---: |
| sumcheck suffix fold, native | 2^18 | 17.97 → 9.57 MB (−47%) | −3.5..−4% |
| | 2^20 | 51.5 → 34.7 MB (−33%) | −2.5..−3% |
| | 2^22 | 202.6 → 135.4 MB (−33%) | noise |
| `fix_suffix_var_mut`, native | 2^18 | 9.57 → 5.36 MB (−44%) | −3..−8% |
| | 2^20 | 26.33 → 17.94 MB (−32%) | −3..−8% |
| | 2^22 | 101.9 → 68.3 MB (−33%) | −3..−8% |
| sumcheck suffix fold, wasmtime | 2^20 | 57.6 → 40.8 MB (−29%) | noise |
| | 2^22 | 208.8 → 141.5 MB (−32%) | noise |
| `fix_suffix_var_mut`, wasmtime | 2^20 | 32.23 → 23.9 MB (−26%) | noise |
| | 2^22 | 107.9 → 74.2 MB (−31%) | noise |

The main win is peak memory. Speed moves 3–8% in `multilinear-util` and stays within noise for the fused sumcheck pass.

## Test plan

- [x] `cargo test --release -p p3-sumcheck -p p3-multilinear-util`, with and without `parallel`
- [x] `cargo test --release -p p3-whir --lib`, `-p p3-stir --lib` without `parallel`
- [x] `cargo check --target wasm32-wasip1 -p p3-sumcheck -p p3-multilinear-util`
- [x] `cargo clippy --all-targets` with and without `--all-features`, nightly `fmt`


tcoratger (approved):
Reviewed the routing changes in full context. The scalar suffix fold cannot clobber unread inputs, the fused in-place schedule remains equivalent to the fresh-buffer path, and PARALLEL_ENABLED correctly tracks whether the iterator backend can dispatch work. The relevant suites pass locally both with and without the parallel feature. Approving.

PR #2196, opened 2026-09-18
perf(field): interleave independent multiplication chains in Powers::collect_n

Nashtare:
`Powers::collect_n` (`BoundedPowers::collect`) fills a buffer with `shift · base^i`. Each packed fill ran one chain of dependent packed multiplications, so it was limited by multiplier latency rather than throughput. There are two commits.

1. **Interleave eight chains.** A new `fill_packed_shifted_powers` starts chain `k` at packed output `k` and steps every chain by `base^(8·WIDTH)`. Chain `k` writes outputs `k, k+8, k+16, …`, so the eight products in a round are independent. Fills shorter than two rounds (16 packed values) keep the single chain. The output is the same sequence of field elements.
2. **Raise `PARALLEL_THRESHOLD` from 2^10 to 2^16.** With the faster fill, dispatching a request below 2^16 powers to Rayon costs more than the calling thread spends filling it.

## Measurements

c8a.4xlarge (AMD EPYC 9R45, 16 cores, `-C target-cpu=native`), `collect_n` microbench (not in this PR), change vs `main`:

| field | threads | 2^10 | 2^14 | 2^16 | 2^20 | 2^22 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| BabyBear | 1 | −53% | −70% | −73% | −68% | −69% |
| BabyBear | 4 | −89% | −74% | −21% | −16% | −22% |
| BabyBear | 16 | −97% | −82% | −6% | −3% | −3% |
| Goldilocks | 1 | −70% | −76% | −77% | −75% | −36% |
| Goldilocks | 16 | −95% | −62% | −13% | −5% | −1% |
| Mersenne31 | 1 | −53% | −69% | −72% | −67% | −68% |
| Mersenne31 | 16 | −97% | −84% | −8% | −5% | −5% |

KoalaBear tracks BabyBear. Re-running `main` after the branch stayed within ±3.5%, except one cell (Goldilocks, 4 threads, 2^16: −8.9%). Fills under 16 packed values are within ±1.7% of `main`.

- **Number of chains:** eight is the best choice for every field (at 2^20, one thread, eight beat both four and sixteen).
- **Why the threshold changed:** without it, 8- and 16-thread pools filling 2^13–2^15 powers got up to 46% slower. With it, those sizes are 52–91% faster than `main`.
- **At 16 threads and 2^20+:** the gain is small. The zeroed allocation and moving cache lines between cores dominate there.

End to end, `coset_lde_batch` on one column (the DFT builds its twiddles with `collect_n`):

| backend | height | 16 threads, cold / warm DFT | 1 thread, cold / warm |
| --- | ---: | ---: | ---: |
| `RecursiveDft` BabyBear | 4096 | −75% / −62% | −11% / −6% |
| `RecursiveDft` BabyBear | 16384 | −60% / −36% | −12% / −4% |
| `RecursiveDft` BabyBear | 2^20 / 2^22 (cold) | within noise | −5.5% / −2.3% |
| small-batch Goldilocks | 4096 | −49% / −39% | −9% / −6.5% |
| `Radix2DitParallel` BabyBear / Goldilocks | 4096 (cold) | −11.5% / −11% | within noise |

With 256 columns (16384 and 2^20 rows), every backend was within ±2%.

- **Unexplained case:** small-batch BabyBear, 256 × 4096, one thread, cold. Its run time is bimodal (about 1.8 ms or 2.2 ms) on both `main` and this branch. The branch landed in the slow mode more often, a +18% median. `collect_n` accounts for about 1 µs of it, so this looks like code layout, not the change.
- **Arm:** the threshold was tuned on Zen 5. On an M4 Pro (noisy, shared machine), the serial path at 2^14–2^16 was still faster than `main`'s parallel split.

## Test plan

- [x] The shared `test_powers_collect` / `test_powers_collect_char2` lengths cover every packed length up to 34 (±1 scalar), lengths around 2^16, and two lengths that aren't a multiple of the packing width, up to about 2^17. Every field crate runs them through `test_field!`.
- [x] New `field/tests/powers_collect.rs` runs them for BabyBear, KoalaBear, Goldilocks, Mersenne31 and BabyBear⁴ inside Rayon pools of 1, 3, 4 and 7 threads. It fails on a broken tail copy.
- [x] `cargo test` for p3-field, p3-field-testing, baby-bear, koala-bear, goldilocks, mersenne-31, dft (aarch64), and the field crates on x86 with `-C target-cpu=native` (AVX-512) and `+avx2`
- [x] `cargo clippy -p p3-field -p p3-field-testing --all-targets -- -D warnings`, with and without `parallel`, nightly `fmt`


tcoratger (approved):
Reviewed the interleaved power chains, tail handling, and Rayon chunk offsets. Chain k starts at packed output k and advances by base^(8 * WIDTH), while each chunk start remains aligned to its packed offset, so the scalar sequence is preserved for full rounds and tails. Native AVX-512 tests pass with serial execution and Rayon pools of 1, 3, 4, and 7 threads. Approving.

PR #2197, opened 2026-09-18
perf(uni-stark,batch-stark): pack only the repeating period of periodic columns

Nashtare:
The quotient computation in `uni-stark` and `batch-stark` expanded the periodic columns into a `Vec<Vec<Pack>>`, one inner `Vec` per packed row group across the whole quotient domain. That is `quotient_size / pack_width` allocations, each holding one packed value per periodic column, and the values repeat with the AIR's period.

`PeriodicLdeTable` already stores only `max_period * blowup` rows and indexes with `lde_idx & (height - 1)`. So a packed group's values depend only on `i_start % height`. `i_start = g * pack_width`, so the group starts repeat modulo `height` every `groups_in_period = height / gcd(height, pack_width)` groups (`PeriodicLdeTable::packed_group_period`). The packing width need not be a power of two.

Both provers now build one flat `Vec<Pack>` of `groups_in_period * ncols` values and read group `(i_start / pack_width) % groups_in_period`.

## Why it's equivalent

- **Same rows:** `groups_in_period * pack_width = lcm(height, pack_width)` is a multiple of `height`, so groups `g` and `g % groups_in_period` start at the same residue modulo `height`, and every lane reads the same row as before.
- **Power-of-two widths:** `groups_in_period` reduces to `height / pack_width` when `height >= pack_width`, and to one group when `height` divides `pack_width`. The one-group case also covers `quotient_size < pack_width`, since `height <= quotient_size`.
- **Index values:** the `get` indices are unchanged. Only how many groups are materialized changes.
- **Power-of-two heights:** `packed_group_period` takes the gcd from trailing zeros, which relies on this. Every `build_periodic_lde_table` implementation produces one: the default in `commit`, the two-adic and hiding FRI PCSs, and the circle PCS.

Proof bytes were compared against `main` with a fixed seed (postcard): periods `[4, 2]` and `[1, 32]` over 64 rows, and `[1, 2]` over 2 rows, on both provers. All were byte-identical.

## Measurements

2^20-row trace, 4 periodic columns (periods 8/16/32/64), blowup 2, BabyBear, Apple M4 Pro. Medians of 3 alternating runs on a machine shared with other builds, so treat these as directional:

| | quotient span | peak RSS (whole prove) |
| --- | ---: | ---: |
| uni-stark, 1 thread | 59.9 → 54.2 ms | 613.9 → 597.2 MB |
| uni-stark, 14 threads | 33.0 → 29.7 ms | 615.3 → 598.5 MB |
| batch-stark, 1 thread | 173 → 166 ms | 632.9 → 614.1 MB |
| batch-stark, 14 threads | 64.5 → 56.9 ms | 634.0 → 615.6 MB |

The saving is in the quotient stage's allocation count and memory; total proving time barely moves, since DFT and Merkle work dominate. It grows with trace length and shrinks as the period approaches the trace length.

## Test plan

- [x] New `periodic_air_two_adic_wide_period_range_prove_verify` / `test_periodic_air_wide_period_range`: periods `[1, 32]` over 64 rows, so `groups_in_period > 1` (checked at runtime: 4 in uni-stark, whose lockstep evaluation packs 8 BabyBear rows on NEON; 8 in batch-stark, which packs 4)
- [x] New `periodic_air_two_adic_tiny_quotient_prove_verify` / `test_periodic_air_tiny_quotient`: periods `[1, 2]` over a 2-row trace, so `quotient_size < pack_width`
- [x] New `packed_group_period_matches_modular_indexing` (p3-commit): seven (height, width) pairs up to height 8, including widths 3 and 6, checked against `get` over four full cycles. It fails on `height / pack_width` at height 8, width 3
- [x] `cargo test --release -p p3-commit -p p3-uni-stark -p p3-batch-stark`, with and without `parallel`
- [x] `cargo clippy --tests --all-features`, nightly `fmt`


uni-stark/src/prover.rs
```diff
@@ -768,20 +768,24 @@ where
         pcs.build_periodic_lde_table(&periodic_cols, trace_domain, quotient_domain);
 
     let pack_width = Pack::<SC, A, Strat>::WIDTH;
-    let periodic_packed: Vec<Vec<Pack<SC, A, Strat>>> = if periodic_table.is_empty() {
+    // `PeriodicLdeTable::get` indexes modulo `periodic_table.height()`, so a packed row
+    // group at `i_start` only depends on `i_start % periodic_table.height()`. Since
+    // `i_start` is always a multiple of `pack_width` and both are powers of two, the
+    // distinct row groups repeat with period `groups_in_period`; store only those.
+    let ncols = periodic_table.width();
+    let groups_in_period = (periodic_table.height() / pack_width).max(1);
```
tcoratger: [MAJOR] This cycle length is correct only when the packing width is a power of two (more generally, when one size divides the other). Neither PackedValue::WIDTH nor Field::Packing requires that, and the surrounding quotient loop deliberately supports partial and wrapping packed groups. For height = 8 and width = 3, group starts visit residues 0, 3, 6, 1, 4, 7, 2, 5. This sets groups_in_period to 2, so group 2 reuses cached group 0 instead of the values beginning at residue 6, changing the quotient polynomial. The duplicated calculation in batch-stark/src/prover.rs has the same issue. Please compute the actual cycle length as height / gcd(height, width), or establish a power-of-two packing-width invariant at the public type boundary, and add a small index-level regression for height 8 and width 3.

tcoratger (changes requested):
One major generic-correctness issue needs fixing; details are inline. The cache reduction is otherwise algebraically sound for the built-in power-of-two packing widths, and the full native test suites pass.

uni-stark/src/prover.rs
Nashtare: Fixed in 7f39cd391. `PeriodicLdeTable::packed_group_period(pack_width)` returns `height / gcd(height, pack_width)` and both provers use it. `packed_group_period_matches_modular_indexing` covers height 8 / width 3 (and six other pairs) against `get`, and fails on the old `height / pack_width`.

tcoratger (approved):
Re-reviewed commit 7f39cd391 after the requested fix. The cycle length is now correctly computed as height / gcd(height, pack_width), including non-power-of-two and non-dividing widths, and both prover integrations use the shared helper. The new modular-indexing regression covers the original height-8/width-3 counterexample. I also reran the release test suites for p3-commit, p3-uni-stark, and p3-batch-stark; all passed. No remaining findings.

PR #2198, opened 2026-09-18
feat!: add WHIR over binary additive domains

tcoratger:
## Summary

This adds a binary-field WHIR path without changing the existing prime-field protocol geometry or hot loops.

- Generalize `p3-whir` over a statically dispatched evaluation-domain interface.
- Add an additive-domain adapter backed by the LCH transform.
- Commit `Poly64` symbols and use the cubic `Poly192` extension for challenges.
- Evaluate query claims in the additive novel basis through direct multilinear selector coordinates.
- Add stratified query sampling for binary WHIR, following the binary decomposition used by Flock.
- Bind the domain and query schedule into the Fiat–Shamir transcript.
- Add a recommended capped-Merkle configuration, end-to-end tests, reference transform checks, and a focused benchmark.

The adapter is binding-only. It does not claim hiding. Unique-decoding and Johnson-bound configurations retain their proven security classification; capacity-bound configurations remain explicitly conjectural.

## Design

The existing two-adic implementation remains the default `WhirDomain`. Domain methods are monomorphized, so the abstraction adds no dynamic dispatch to either backend.

Binary messages are encoded over `Poly64`. Extension-valued codewords are flattened into three base-field columns before the additive transform and reconstructed afterward. Query positions are mapped to coordinates derived from the Cantor subspace-vanishing polynomials, which makes the existing sumcheck selector operate on the additive novel basis rather than incorrectly treating an additive-domain point as a multiplicative power sequence.

The transcript uses a typed polynomial-field tag and a domain protocol identifier. The legacy two-adic identifier is empty, preserving the existing prime-field transcript byte-for-byte.

## Security review

I reviewed the complete changed surface with the `/audit-crypto` checklist after implementation.

One configuration-integrity issue found during that pass was fixed: domain capacity, domain identity, and query geometry are now encapsulated and checked when the prover is constructed. Regression tests show that changing either the domain identity or the stratified-query schedule changes the Fiat–Shamir seed.

No unresolved critical, high, or medium findings remain.

## Compatibility and performance

The prime-field path keeps identical proof bytes and Merkle-digest counts in the small, medium, and large benchmark cases:

| Case | Proof bytes | Digests |
| --- | ---: | ---: |
| Small | 41,585 | 516 |
| Medium | 72,335 | 1,210 |
| Large | 90,662 | 1,648 |

The focused prime benchmark remains statistically unchanged at roughly 356 ms.

Binary WHIR at a 100-bit target measured locally as follows:

| Message | Commit | Prove | Verify | Proof |
| 2^12 | 372 us | 262 ms | 662 us | 62,168 B |
| 2^16 | 8.05 ms | 2.05 s | 1.10 ms | 108,359 B |

The proving figures include configured grinding.

## Validation

Only targeted checks were run:

- `cargo test -p p3-binary-pcs --test whir`
- `cargo test -p p3-whir --lib` — 195 passed, 3 heavy tests ignored
- `cargo clippy -p p3-whir -p p3-binary-pcs -p p3-binary-dft -p p3-binary-field -p p3-sumcheck --all-targets --features parallel -- -D warnings`
- Complete `p3-binary-pcs` crate tests during development
- Serial and parallel selector tests
- Scalar-reference checks for base-field and extension-field additive encoding
- Transcript domain-separation and tamper-rejection tests

## References

The protocol and implementation choices were checked against WHIR (ePrint 2024/1586), Ligerito (ePrint 2025/1187), Flock (ePrint 2026/1329 and its implementation), leanVM, and Layr Labs' `flock-challenge-multi` implementation.

---

Part of #2232.


binary-pcs/README.md
```diff
@@ -47,3 +47,16 @@ over-budget protocols. Existing callers must handle this new rejection or choose
...
+
+## WHIR over the additive domain
+
+The `whir` module connects WHIR to the Cantor additive code. Commitments use
+the 64-bit polynomial-basis field; challenges use its 192-bit cubic extension.
+
+Queries follow a transcript-bound stratified schedule. Use
+`recommended_cap_height` to stop Merkle paths at its deepest stratum.
+
+Unique decoding and the Johnson regime use proven bounds. The capacity regime
```
Nashtare, line 58: This sentence clashes with lines 10–11 of the same file, which say the capacity bound is refuted over characteristic 2 with `F_2`-subspace domains, and the Cantor domain is one. The code follows this sentence, not line 10: `WhirConfig::new_with_domain` accepts `SecurityAssumption::CapacityBound` for `BinaryWhirDomain`.

Locally, at 2^16, fold 4, rate 1/2, a 100-bit target and 20-bit PoW, the capacity config derives 87/21/12 queries. `prescribed_security` reports 98.9 bits for a 57,163-byte proof. Johnson needs 187/42/24 queries and 107,968 bytes. `prescribed_security` returns a single `WHIR_OPENING_LABEL` number, so nothing downstream learns that the regime is conjectural, let alone refuted here.

Could `new_with_domain` reject the capacity regime for this domain, for example through a `WhirDomain` hook that defaults to `true` and returns a typed error? This paragraph could then say capacity is unsupported here. Small aside: the section sits below the "Part of Plonky3" footer.

whir/src/transcript/mod.rs
```diff
@@ -179,6 +179,29 @@ pub const fn query_draws(folded_domain_size: usize, num_queries: usize) -> usize
     }
 }
 
+/// Canonical power-of-two decomposition of a query count, deepest strata first.
+///
+/// A summand `2^c` samples once in each of `2^c` equal subtrees.
+///
+/// Every query remains marginally uniform. By AM--GM, stratification does not
```
Nashtare: The first sentence isn't quite right. The draw for stratum `j` of a depth-`c` summand is uniform on that stratum only, not on the whole domain.

The second sentence is the property that matters, and it holds: with per-stratum bad densities `δ_j` averaging `δ`, the summand misses with probability `∏(1 − δ_j) ≤ (1 − δ)^{2^c}`. A brute-force check over 5,760 bad sets (domains 2^6 to 2^10, `t` up to 187) never went above `(1 − δ)^t`, and a set concentrated in a few strata does strictly better.

Could the doc state that bound instead of marginal uniformity?

binary-pcs/src/whir.rs
```diff
@@ -0,0 +1,193 @@
...
+        assert_eq!(fast, reference);
+    }
+
+    #[test]
+    fn query_coordinates_evaluate_the_encoded_novel_basis_polynomial() {
+        let coefficients = (0..8)
+            .map(|index| Poly64::new((13 * index + 7) as u64))
+            .collect::<Vec<_>>();
+        let domain: BinaryWhirDomain<p3_binary_dft::LchNtt<Poly64>> = BinaryWhirDomain::default();
+        let encoded = domain.encode_batch(RowMajorMatrix::new(coefficients.clone(), 1), 0);
```
Nashtare: At `log_inv_rate = 0` with 3 variables, every queried index stays inside the message subspace. So this test doesn't pin that LCH row `i` sits at `domain_point(i)` once the domain grows past it, and that is the case WHIR queries. Locally the property holds for m = 0..10 and rates 0..4, on base and extension columns. Would it be worth looping this test over a couple of rates and also asserting that a perturbed value is rejected?

binary-pcs/tests/whir.rs
```diff
@@ -0,0 +1,91 @@
...
+use p3_whir::{FoldingFactor, ProtocolParameters, SecurityAssumption, WhirConfig, WhirProver};
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+type Hash = SerializingHasher<Keccak256Hash>;
+type Compress = CompressionFunctionFromHasher<Keccak256Hash, 2, 32>;
+type Mmcs = MerkleTreeMmcs<Poly64, u8, Hash, Compress, 2, 32>;
+type Challenger = BinaryChallenger<Poly64, HashChallenger<u8, Keccak256Hash, 32>>;
+type Domain = BinaryWhirDomain;
+type Pcs = WhirProver<Poly192, Poly64, Domain, Mmcs, Challenger, PrefixProver<Poly64, Poly192>>;
```
Nashtare, line 24: The only end-to-end test uses `PrefixProver` with one shape (n = 12, fold 3, rate 1). The Suffix layout feeds untransposed rows to the encoder and binds the folding point reversed, so it is the other place a variable-order slip could hide. Locally Suffix passes across folds 1 to 5, mixed schedules and rates 1 to 3. Could this test also run `SuffixProver` and one `ConstantFromSecondRound` schedule?

Nashtare (commented):
I looked at both sides. On the binary side I covered the additive code and its selector coordinates, the stratified schedule, the transcript binding and the field choices. On the two-adic side I checked that nothing moves.

**What holds**
- Locally I encoded random messages for m = 0..10 and rates 0..4, both as base columns and as `Poly192` columns through `encode_extension_batch_padded`.
  - At every checked index, the selector built from `query_point` reproduces the codeword entry (57,372 checks), and every perturbed value is rejected.
  - `combine` and `weights_at` agree with the multilinear extension of the weights.
- End to end, 1,142 honest proofs verify, serial and parallel. They cover Prefix and Suffix, folding 1 to 5 and mixed schedules, rates 1 to 3, and unique decoding and Johnson.
  - 5,341 value tampers are rejected.
  - 26 structural mutations are rejected with an error, never a panic. These include dropped rounds, short rows, swapped opening variants and swapped Merkle proofs.
- The same results hold on a merge with current main, which is clean.

**The prime path is unchanged**
I built one probe against the merge base and against this head. It produces 175 proofs, and all of them are byte-identical and verify on both:
- plain and prescribed-point openings across 9 folding schedules, all three regimes, PoW 0/10/16 and both layouts;
- your bench cases, which give exactly 41,585 / 72,335 / 90,662 bytes;
- 6 HVZK proofs.

The multi-stark WHIR fixture also regenerates byte-identical to the committed postcard. Three more checks are identical too:
- derived configs and `prescribed_security`, over a 270-config grid;
- 434 transcript prefixes (ids, labels, pattern hash and first challenges);
- the univariate `SelectStatement` results.

Commit and prove timings are within noise. Verify came out about 12% faster at this head, and I couldn't attribute that. A 600-flip tamper sweep over 17 proofs matches the base's accept/reject pattern exactly, so no check was dropped.

**Why the stratified schedule is sound**
A depth-`c` summand misses a fixed set with probability `∏(1 − δ_j) ≤ (1 − δ)^{2^c}` by AM–GM. Summands are independent, so the uniform `(1 − δ)^t` accounting still bounds it. A brute-force check over 5,760 bad sets never went above that bound.

**Transcript and fields**
Every challenge is drawn in `Poly192`. The domain id, the stratification flag and the polynomial-field tag all reach the seed before the first draw. The two-adic path adds nothing to its label.

- The adapter accepts `CapacityBound`, but this crate's README says that bound is refuted on the Cantor domain. At 2^16 and a 100-bit target, it reports 98.9 bits for a 57 KB proof, against 108 KB under Johnson (inline, `README.md`).

**Smaller things**
- The public `p3-whir` types change incompatibly. Could the title carry the `!` (inline, `whir.rs`)?
- The other inline comments are about test coverage for the rate and layout cases, one doc sentence on the stratified schedule, and wording.

whir/src/pcs/prover/mod.rs
```diff
@@ -113,7 +113,26 @@ where
     ///
     /// The extension-field MMCS is constructed by wrapping the base-field one,
     /// so callers never have to thread it through manually.
+    ///
+    /// # Panics
+    ///
+    /// Panics when `config` was derived for a different domain.
     pub fn new(config: WhirConfig<EF, F, Challenger>, dft: Dft, mmcs: MT) -> Self {
+        assert_eq!(
```
Nashtare, line 121: These checks only run here. `WhirProver.config` and `.dft` are `pub`, so assigning a config after construction skips them. `WhirVerifier::new` is public too, and takes `config` and `domain` separately with no check. Only one `WhirDomain<Poly64, _>` exists today, so this is minor. Would the same checks be worth adding to `WhirVerifier::new`?

whir/src/parameters/whir.rs
```diff
@@ -141,8 +145,10 @@ pub struct RoundConfig<F> {
     pub log_inv_rate: usize,
     /// Size of the evaluation domain before folding in this round.
     pub domain_size: usize,
-    /// Generator of the folded evaluation domain after this round's fold.
-    pub folded_domain_gen: F,
+    /// Base-two logarithm of the folded evaluation domain size.
+    pub log_folded_domain_size: usize,
+    /// Marker retaining the base-field type in the public configuration.
+    pub _field: PhantomData<F>,
```
Nashtare: This changes the public API. `RoundConfig::folded_domain_gen` is gone, and every downstream literal now needs `_field: PhantomData` (the PR's own tests spell it 8 times). `WhirConfig` gains `pub(crate)` fields, so it can no longer be built with a struct literal outside the crate. `WhirVerifier` gains a `Dft` parameter and a `domain` argument, and `WhirRoundShape` loses `Copy`. Could the title carry a `!`, as #2166 does, so release-plz treats this as breaking? Since the struct breaks anyway, could `RoundConfig` drop the `<F>` parameter rather than keep a public marker?

whir/src/parameters/whir.rs
```diff
@@ -29,12 +30,15 @@ pub enum WhirConfigError {
     #[error(transparent)]
     FoldingFactor(#[from] FoldingFactorError),
 
-    /// The domain after the first fold exceeds the base field two-adicity.
+    /// The domain after the first fold exceeds the encoder's capacity.
+    ///
+    /// The variant name is retained for API compatibility with the original
```
Nashtare: This sentence describes the history of the code rather than the code; the same goes for "legacy two-adic transcript" in `domain.rs`. Could the doc just say this is the encoder-capacity limit, with the `two_adicity` field renamed or documented to match? Separately, the PR drops three rationale comments this change doesn't touch: the `num_claims` note and the `Field::bits` rounding note on `initial_claims_error`, and "Alpha is sampled before the first folding grind, so grinding cannot rescue it" on `validate_initial_claims`. Could those stay?

sumcheck/src/constraints/statement/select.rs
```diff
@@ -279,15 +285,21 @@ impl<F: Field, EF: ExtensionField<F>> SelectStatement<F, EF> {
         // Walk the stored univariate selection points in order.
```
Nashtare: This comment and the doc above it still describe only univariate points, and so do `combine`'s algorithm section and the struct doc. Relatedly, `len()` and `is_empty()` now count direct-point constraints, but `iter()` still zips only `vars` with `evaluations`. On a point statement it yields nothing while `len()` is nonzero. Could the docs cover both representations, and `iter()` say it covers univariate constraints only?

sumcheck/src/constraints/statement/select.rs
```diff
@@ -311,11 +323,30 @@ impl<F: Field, EF: ExtensionField<F>> SelectStatement<F, EF> {
...
             // Evaluate the polynomial at `var` using Horner's method.
             // This computes: p(var) = c_0 + var(c_1 + var(c_2 + ...))
             poly.iter().copied().horner::<EF, _>(var) == expected_eval
-        })
+        });
+        if !univariate {
+            return false;
+        }
+
+        let mut scratch = poly.as_slice().to_vec();
```
Nashtare, line 294: Nit: this copies the whole polynomial even when `points` is empty, which is every two-adic statement. Could it return early in that case?

binary-pcs/README.md
tcoratger: Fixed in d4e2b06a. Domain implementations now declare supported soundness regimes. New domains reject capacity by default, the canonical two-adic domain explicitly retains it, and the Cantor additive domain returns a typed UnsupportedSecurityAssumption error. The README now states that capacity is unsupported and the section sits above the footer.

whir/src/transcript/mod.rs
tcoratger: Fixed in d4e2b06a. The documentation now states the per-stratum density bound prod_j (1 - delta_j) <= (1 - delta)^(2^c) and no longer claims marginal uniformity over the whole domain.

binary-pcs/src/whir.rs
tcoratger: Fixed in d4e2b06a. The selector-coordinate test now covers log-inverse rates 1 through 3, checks every extended-domain row, and verifies that perturbing the claimed codeword value is rejected.

binary-pcs/tests/whir.rs
tcoratger: Fixed in d4e2b06a. End-to-end coverage now includes suffix binding with a ConstantFromSecondRound schedule and a rate-1/4 starting code.

whir/src/pcs/prover/mod.rs
tcoratger: Fixed in d4e2b06a. The verifier constructor now checks encoder capacity, protocol identity, and query stratification against the supplied domain before accepting the replay context.

whir/src/parameters/whir.rs
tcoratger: Fixed in d4e2b06a. The PR title now carries feat!. RoundConfig no longer has a base-field parameter or public marker; it stores only round data.

whir/src/parameters/whir.rs
tcoratger: Fixed in d4e2b06a. The capacity error and field now use encoder-domain terminology. The historical wording is gone, and the initial-claim count, conservative field-width rounding, and pre-grinding batching rationale are restored.

sumcheck/src/constraints/statement/select.rs
tcoratger: Fixed in d4e2b06a. The statement, weighting, verification, and batching documentation now covers scalar-power and direct-coordinate representations. The iterator explicitly documents that it exposes only scalar-point constraints.

sumcheck/src/constraints/statement/select.rs
tcoratger: Fixed in d4e2b06a. Verification returns immediately after scalar-point checks when there are no direct points, avoiding the polynomial copy on every two-adic statement.

binary-pcs/src/whir.rs
```diff
@@ -0,0 +1,211 @@
...
+use p3_matrix::dense::RowMajorMatrix;
+use p3_multilinear_util::point::Point;
+use p3_whir::{SecurityAssumption, WhirDomain, WhirQueryPoint};
+
+/// Returns the Merkle cap height of the deepest query stratum.
+///
+/// One cap node covers each deepest stratum, so authentication paths stop at
+/// a protocol-fixed layer.
+#[must_use]
+pub fn recommended_cap_height<Challenger>(
```
Nashtare, line 21: The doc says this returns "the Merkle cap height of the deepest query stratum", and the README says to use it "to stop Merkle paths at its deepest stratum". Two things do not match that.

1. It reads `round.num_queries`, but the quantity that sets a round's deepest stratum is `query_draws(folded_domain_size, num_queries)`. Those differ exactly when a round saturates: `query_draws` returns 0, `query_indices` opens every position and draws nothing, yet the round still contributes `ilog2(num_queries)` to the max here.
2. It returns one `max` over all rounds, and the caller feeds that single value to one `Mmcs` that is then shared by every round's tree (`ExtensionMmcs::new(mmcs.clone())`). So the height derived from round 0's query count is applied to the last round's tree, whose height is only `2^log_inv_rate * 2^(num_variables - folding)`.

Consequence of (2): for a later round the cap sits strictly deeper than that round's deepest stratum, so its commitment carries more digests than the doc promises — and once a later tree gets short enough, `cap_height >= num_layers` trips `MerkleTree::cap`'s `assert!(cap_height < num_layers)`, i.e. a prover panic from a configuration `WhirConfig::new_with_domain` accepted. I have not run the derivation to pin a concrete panicking parameter set, but nothing in the config path bounds the helper's output against the smallest tree.

Clamping per tree (`min(cap_height, num_layers - 1)`) or deriving from `query_draws` and returning a per-round height would both close it.

binary-pcs/README.md
```diff
@@ -46,4 +46,17 @@ on rejection. The existing infallible `open` / `open_at` traits panic on invalid
...
 
+## WHIR over the additive domain
+
+The `whir` module connects WHIR to the Cantor additive code. Commitments use
+the 64-bit polynomial-basis field. Challenges use its 192-bit cubic extension.
+
+Queries follow a transcript-bound stratified schedule. Use
+`recommended_cap_height` to stop Merkle paths at its deepest stratum.
+
+Unique decoding and the Johnson regime use proven bounds. The capacity regime
```
Nashtare, line 58: This says the Johnson regime uses proven bounds, and line 12 now says the folding step "additionally needs the proven Reed--Solomon mutual-correlated-agreement bound". But `security/src/assumption.rs` — the module that actually prices the regime — documents `JohnsonBound` as "**Requires** mutual correlated agreement up to the Johnson bound", and places it between `UniqueDecoding` ("No conjectures") and `CapacityBound` ("Requires conjecturing ..."). Something in the repo has to give.

It matters here because `supports_security_assumption` rejects only `CapacityBound`, so `JohnsonBound` is accepted on the Cantor domain and both new end-to-end tests plus the bench run under it — the bench reports a 100-bit target. An integrator reading this README will select Johnson believing the folding step is unconditional, on exactly the class of domains (`F_2`-subspace, characteristic 2) where the analogous capacity assumption is refuted two paragraphs above.

The previous wording ("`p3-security` documents it as resting on a correlated-agreement conjecture, and it is excluded here by choice") did not have this problem. Note also that line 13 still says "This BaseFold implementation nevertheless stays in unique decoding by design", which the WHIR adapter does not.

Either restore the caveat and match `p3-security`, or, if `p3-security`'s doc is the one that is wrong, fix it there and say so.

whir/src/pcs/prover/mod.rs
Nashtare, line 121: These three asserts are the fix for my round-1 comment and they are the right ones, but the other half of that comment is still open: `config`, `dft` and `mmcs` are `pub` on `WhirProver` (lines 80-86), so `prover.config = other_config` after construction skips all three. `WhirVerifier`'s fields are `pub(crate)`, so the two types are now inconsistent about this too.

`adapter.rs` rebuilds a `WhirVerifier` at verify time, so this is a footgun rather than a hole. Making the fields private, or `pub(crate)` with accessors, would match the verifier and make the asserts an invariant rather than a one-shot check.

Separately: the regime itself is never re-checked. `supports_security_assumption` is consulted once, inside `new_with_domain`, and `WhirConfig.params` is `pub`, so mutating `params.soundness_type` to `CapacityBound` afterwards passes all three asserts here and seeds the transcript label with a regime the domain declared unsupported. The concrete effect of that particular mutation is conservative (capacity's larger list size makes `validate_initial_claims` stricter and `prescribed_security` fails closed), so I am not claiming a downgrade — but both constructors already have the domain in hand, so re-checking the regime here would make the new invariant hold where it is actually load-bearing.

whir/src/pcs/verifier/mod.rs
```diff
@@ -71,15 +75,39 @@ where
...
+    /// # Panics
+    ///
+    /// Panics when the configuration was derived for another domain.
+    pub fn new(
         config: &'a WhirConfig<EF, F, Challenger>,
+        domain: &'a Dft,
         mmcs: &'a MT,
         variable_order: VariableOrder,
     ) -> Self {
+        // The verifier must reconstruct the same code and query schedule as the prover.
```
Nashtare, line 91: Adding the asserts made `new` non-`const`, but the struct doc at line 37 still says "Construction is `const`; spinning up a fresh verifier per proof is free." Worth updating, since the loss of `const` is part of the `feat!` surface.

sumcheck/src/constraints/statement/select.rs
```diff
@@ -122,76 +123,26 @@ fn packed_batch_pows<F: Field>(points: RowMajorMatrixView<'_, F>) -> RowMajorMat
     mat
 }
 
-/// A batched system of `select`-based evaluation constraints for multilinear polynomials.
+/// Evaluation constraints batched through selector polynomials.
```
Nashtare, line 126: My round-1 ask was to extend this documentation so it covers the direct-coordinate representation as well as the univariate one. The commit instead removed most of it: the definition `select(X, Y) = prod_i (X_i Y_i + (1 - Y_i))`, the derivation of `select(pow(z), b) = z^int(b)`, the statement of what each constraint asserts, the batched weight polynomial and target sum, and `combine`'s three-stage algorithm are now nowhere in the file.

The replacement is accurate but says much less than the original did about a type whose correctness depends on the reader knowing the selector's exact form. Could the removed material come back, extended with the direct-point case, rather than being replaced by it? The four-line summary at the top of the struct reads well as an addition to it.

whir/src/transcript/mod.rs
```diff
@@ -776,16 +869,31 @@ where
...
         // A saturated phase has nothing left to decide, so no draw is described.
         if draws == 0 {
             return (0..1usize << width).collect();
         }
-        self.state
-            .challenge_uniform_bits::<F>(label, width, draws)
-            .into_iter()
-            .map(TranscriptBound::into_inner)
-            .collect()
+        if !stratified {
```
Nashtare: This body and the verifier's at line 1043 are byte-identical — I diffed them, zero differences. They have to stay in lockstep or Fiat-Shamir desynchronises, and nothing in the type system says so. A shared free function taking `(&mut state, label, width, draws, stratified, depths)` would remove the hazard; the two sites already agree on every argument.

binary-field/src/transcript.rs
```diff
@@ -120,6 +120,45 @@ impl_transcript_field!(BinaryField32, u32, 32);
 impl_transcript_field!(BinaryField64, u64, 64);
 impl_transcript_field!(BinaryField128, u128, 128);
 
+impl TranscriptField for Poly64 {
```
Nashtare: `impl_transcript_field!`'s doc says "Every level shares one seeding rule, so every level is generated from here", which this hand-written copy makes false. The body is the macro's with `WIDTH = 8` substituted and `Self::new` in place of `from_repr` (which are the same function for `Poly64`), so the only reason it is not generated is the `TypeTag` variant.

Threading the tag through the macro — an extra argument, or a small `fn algebra_tag` hook — would keep the seeding rule in one place and let the macro's doc stay true.

whir/src/parameters/whir.rs
```diff
@@ -232,11 +249,74 @@ where
...
+    /// When the opening count is known, prefer [`Self::new_with_initial_claims`];
+    /// otherwise the PCS validates the full batch at opening and verification time.
+    /// The target applies per error term. Full PCS security composes their
+    /// probabilities.
+    ///
+    /// # Errors
+    ///
+    /// - The domain does not support the requested soundness regime.
+    /// - The folding factor does not fit the polynomial size.
+    /// - The first fold leaves a domain larger than the base-field two-adicity.
```
Nashtare: Leftover from the rename: "the base-field two-adicity" is now the encoder capacity, and the error it describes is `FoldedDomainExceedsCapacity`. The same sentence appears on `new` just above.

Nashtare (approved):
Round 2, at `d4e2b06a`. I re-checked each of my eight points against the code rather than against the replies, and made a fresh pass over the delta.

**Round-1 follow-ups**

| | |
| --- | --- |
| Capacity on the Cantor domain | **fixed** |
| Stratified-schedule doc | **fixed** |
| Selector-coordinate test coverage | **fixed** |
| Suffix / rate end-to-end coverage | **fixed** |
| Verifier-side domain check | **fixed**, with a residual |
| Breaking API not flagged | **fixed** |
| Change-narrating wording | **fixed**, one line left |
| `SelectStatement` docs and the empty-`points` copy | **fixed**, but see below |

On the capacity gate I traced every other route shut, not just the one you patched: `WhirConfig` has no serde, `max_log_domain_size` / `domain_id` / `stratified_queries` are `pub(crate)` so there is no external struct literal, `new_with_max_domain_log` is `pub(crate)`, the binary fields are not `TwoAdicField` so `WhirConfig::new` is unreachable for them, and the constructor asserts stop a two-adic-derived config being paired with the additive domain. The gate holds.

On the verifier check: the two constructors now run the same three asserts, and `WhirVerifier`'s fields are `pub(crate)`. `WhirProver.config` and `.dft` are still `pub`, so post-construction assignment still skips them — `adapter.rs` rebuilds a `WhirVerifier` at verify time, so this is a footgun rather than a hole, but it is the half of the original comment that is still open (inline).

On the doc rewrite: the ask was to extend `SelectStatement`'s documentation to cover the second representation. The commit instead deleted most of it — the `select` definition, its derivation, the batching formulas and the three-stage algorithm are all gone (inline).

**Fresh pass**

- **The prime path is untouched.** `domain_id` empty means nothing reaches the instance label; `stratified_queries` false keeps `pattern()` and `query_indices` on the original single-step branch; `query_point` returns `two_adic_generator(log_folded_domain_size)^index` and that exponent is exactly the old `folded_domain_gen`'s, in the loop and in both branches of `final_round_config`; `commit_extension`'s `1 << log_inv_rate` equals the old `inv_rate(round)`. Adding `TypeTag::BinaryPolynomial` mid-enum does not move any existing fingerprint, because the pattern hash is over the alternate-`Display` rendering rather than over discriminants.
- **The stratified sampler is sound and bounded.** `query_summand_depths` is the exact deepest-first binary decomposition; `draws < 2^width` whenever `draws != 0`, so `width - depth >= 1` never underflows; `stratified_index` stays in range and hits each stratum once per summand; the draws come from `sample_uniform_bits::<true>`, so no modulo bias; prover and verifier derive identical depths and consume the pattern in the same order.
- **Index bounds.** `shape.index_bits == log_folded_domain_size` for every intermediate round and both branches of `final_round_config`, so a Fiat-Shamir-drawn index always satisfies `BinaryWhirDomain::query_point`'s range assert, and the saturated branch returns `0..1<<width`, also in range.
- **The LCH selector coordinates line up at all four sites.** `query_point` returns `[W_{n-1}(p), ..., W_0(p)]`; `verify` folds `point.iter().rev()`, pairing index bit `j` with `point[k-1-j] = W_j(p)`; `combine`'s `pow_matrix` uses the same mapping; and `weights_at`'s `point[j] <-> row[j]` is exactly what `eval_select` gives the scalar arm. So the direct-point selector is a faithful analogue of the univariate one under both `Prefix` and `Suffix`.
- **Extension flattening is consistent.** Flatten to `3w` base columns, encode per column, reconstitute at width `w` — correct because the `Poly64` twiddles act coordinatewise on any `BasedVectorSpace<Poly64>` basis. The Merkle commitment is taken over the reconstituted `Poly192` matrix through `FlatMatrixView` and the verifier rebuilds `ExtensionMmcs` from the same base mmcs, so there is no prover-choosable split between the two forms.
- **`SelectStatement` cannot be mixed.** `add_constraint` and `add_point_constraint` each assert the other vector is empty, so `combine`'s either/or branch and `verify`'s `points.zip(evaluations)` are both well-formed. `len`/`is_empty` now key off `evaluations`, so direct points count toward the challenge-power shift. The pre-existing `verify` would have accepted a direct-point final statement vacuously; this PR closes that.
- **`Poly64`'s transcript impl round-trips.** `new == from_repr == Self(x)`, `modulus: 0x1b` matches `x^64 + x^4 + x^3 + x + 1`, and `observe_seed` is length-prefixed and injective exactly like the macro.
- `Poly64` correctly gets `plain_butterfly` rather than `coordinate_butterfly`, since its bytes are polynomial-basis coordinates rather than `GF(2^8)` tower coordinates.

**Two things I would fix before merge**

`recommended_cap_height` (inline) is the one that can actually bite: it is a new public helper the README tells integrators to use, and the value it returns is not the quantity its own doc describes. The other is the README's Johnson wording, which now contradicts `p3-security` in the same tree.

Neither is a soundness issue, and the round-1 blocker is resolved, so this is an approval.

binary-pcs/src/whir.rs
tcoratger: Both halves of (1) and (2) were real; the panic in the consequence was not. I pinned it rather than taking it on faith.

**No accepted configuration can panic.** `MerkleTreeMmcs::commit` clamps before it caps:

```rust
let num_layers = tree.num_layers();
let effective_cap_height = self.cap_height.min(num_layers.saturating_sub(1));
let cap = tree.cap(effective_cap_height);
```

`merkle-tree/src/mmcs/batch.rs:60-62`, with the comment "Make cap_height fit this tree (small trees during FRI folding may have fewer layers)". That is the only call to `MerkleTree::cap` in the repo, so the `cap_height < num_layers` assert is unreachable from any WHIR path, and a cap deeper than a tree just commits that tree's leaf digest layer. I checked it end to end rather than only by reading: with `cap_height = 40` forced in, 11 variables / `Constant(4)` / rate 1 (trees of log-height 8 and 7) and 4 variables / `Constant(4)` / rate 1 (a single tree of log-height 1, so the clamp saturates to 0) both prove and verify. The verifier stays in lockstep because it derives its own effective height the same way.

For the record, the parameter sets you did not pin do exist: sweeping variables 4-24, rates 1-4, constant and mixed folding, both accepted regimes, security 16-128, 21158 accepted configurations had the old helper return a value deeper than some tree, including 128-bit ones such as 11 variables / `Constant(4)` / rate 1 / Johnson (cap 8, last tree of log-height 7). They were all absorbed by the clamp.

**What I changed anyway**, because everything except the panic stands: the helper now reads `query_draws(folded_domain_size, num_queries)` instead of the configured count, and clamps the maximum to the shallowest folded domain, so what it returns really is a depth every round's tree can carry and the commitment stops carrying more digests than the doc promises. On the 128-bit set above the returned value drops from 8 to 6, because round 0 saturates (298 queries over 256 positions) and contributes nothing at all.

The per-round strata are safe on their own, which is worth recording: a non-saturated round has `draws < 2^index_bits`, so its deepest stratum is at most `index_bits - 1` and `push_stratified_query_indices`'s `index_bits - depth` can never underflow. Only the cross-round maximum needed the clamp.

`binary-pcs/tests/whir.rs` gets one regression test on a configuration where both halves bite: 11 variables, starting rate 2, explicit round rate 1, `Constant(4)`, Johnson, 32-bit target. The final phase asks for 75 queries over 16 positions, so it draws nothing (the old helper charged it `ilog2(75) = 6`); round 0 draws 35 over 512 positions, depth 5; the last tree has log-height 4, so the answer is 4.

Since the panic is unreachable I added no defensive machinery for it.

binary-pcs/README.md
tcoratger: `p3-security` is the document that is wrong, and I fixed it there.

The regime doc predates the literature. When WHIR was written (ACFY24) mutual correlated agreement for Reed--Solomon was proven only up to `delta/2`, and everything above that, including the Johnson radius, was conjectured. That is no longer the state of the art:

- Haboeck, *A note on mutual correlated agreement for Reed--Solomon codes*, <https://eprint.iacr.org/2025/2110>, generalises the Guruswami--Sudan analysis of [BCI+20] and proves Definition 1 up to the Johnson radius `1 - sqrt(1 - delta)`, for `RS[F_q, D, k]` with an arbitrary evaluation set `D`. No restriction on the characteristic and none on `D` being a subspace, so it covers the Cantor domain.
- Bordage, Chiesa, Guan, Manzur, *All Polynomial Generators Preserve Distance with Mutual Correlated Agreement*, CCC 2026, proves it for all polynomial generators up to the Johnson bound, closing the question ACFY posed.

What was disproved in late 2025 is the up-to-capacity family: correlated agreement at capacity, WHIR's mutual variant at capacity, and DEEP-FRI's list-decodability at capacity. The Johnson regime is untouched, which is exactly the split this crate already encodes elsewhere: `whir/benches/fri_vs_whir.rs:108` lists Johnson as "proven", `stir/src/pcs.rs:45` calls only the capacity variant "the standard WHIR/ACFY conjecture", and `hiding_base_case_security_report` classifies Johnson's mutual-correlated-agreement term as `Approximation`, never `Conjectural`.

So `SecurityAssumption::JohnsonBound` now reads "No conjectures", with one line naming where the property is proven, and the two references are added to the module header. `CapacityBound` keeps its "Requires conjecturing" wording. The README's claim stands as written.

On your last point: the BaseFold sentence now says "; the WHIR adapter below also accepts the Johnson regime", so it no longer reads as a statement about the whole crate.

whir/src/pcs/prover/mod.rs
tcoratger: Both done.

`config`, `dft`, `mmcs` and `extension_mmcs` are `pub(crate)` now, matching the verifier, so the three checks are an invariant rather than a one-shot. Nothing outside the crate loses access to the config: the prover already derefs to it, so `prover.round_parameters` and `prover.check_pow_bits()` keep working, and the three call sites that went through the field (the two benches and the comparison example) just drop the `.config` hop.

One caller was doing exactly what you described: `multi-stark/tests/whir_fibonacci.rs` swapped in a different `WhirConfig` after construction. It now builds the scheme from the config it wants.

The regime is re-checked too. Both constructors ask the domain `supports_security_assumption(config.soundness_type)` alongside the other three checks, so a `soundness_type` mutated to a regime the domain declared unsupported is rejected where the transcript label is seeded, not silently accepted.

whir/src/pcs/verifier/mod.rs
tcoratger: Fixed. The line now says construction only checks the config against the domain, so a fresh verifier per proof is cheap, which is what the asserts left true.

whir/src/transcript/mod.rs
tcoratger: Factored. `assemble_query_indices` in `whir/src/transcript/mod.rs` takes the width, the draw count, the stratification flag, the summand depths and a closure that performs one draw; both drivers now pass their own state and label through that closure and share the body. A free function taking `&mut state` directly would need a trait over the two state types, since `challenge_uniform_bits` is an inherent method on each; the closure gets the same single-definition guarantee without one.

binary-field/src/transcript.rs
tcoratger: Threaded. The macro takes a fourth argument, `|degree, basis| <tag>`, and binds those two names as the parameters of the generated function, so the tag expression can use them despite macro hygiene. The polynomial-basis level is now generated like every other one and the hand-written copy is gone, which also removes the `new`/`from_repr` and error-message divergences. The macro's claim that every level is generated from there is true again.

whir/src/parameters/whir.rs
tcoratger: Fixed, though there is only one site. The bullet now reads "The first fold leaves a domain larger than the encoder's capacity", matching the error it describes. Grepping the crate for the old wording finds no second occurrence; the sentence on the derived-error list is the one that carried it.

sumcheck/src/constraints/statement/select.rs
tcoratger: Restored and extended, not replaced. The struct doc now carries the selector definition, the power map, the derivation that it collapses to a monomial on a Boolean vertex, the claim each stored constraint asserts, and the weighted combination that turns the set into one sumcheck, with the direct-coordinate case folded into the same statement instead of described separately: the two representations differ only in how the `k` coordinates of `X` are obtained, so everything downstream of `select(X, b)` is shared.

The batching method gets its own restored text: both accumulators are added to, the offset exists so several statement kinds can share one challenge without reusing a power, and the three stages are back (coordinate matrix, doubling into the selector matrix, dot with challenge powers), with the scalar and direct sources named on the first one and the flat row-major layout reason kept.

House style applies, so it is one sentence per line with no wrapping, formulas in fenced blocks, and no identifier named in prose. Density on the file goes 31.7% to 33.7%; the longest prose run stays at its previous 6, and the wrapped-sentence gate reports zero hits on any line this PR adds.

PR #2199, opened 2026-09-18
perf(challenger)!: squeeze the transcript before a proof-of-work search

Nashtare:
`SerializingChallenger32/64::grind` tested each candidate on a clone of the whole challenger. A `HashChallenger` inner only hashes when it samples, and then it hashes its entire input buffer. So every candidate rehashed everything observed since the last sample.

At the PCS batching grind that is every claimed opening: the claims are absorbed and the grind runs before anything is sampled. In the STIR Keccak example (KoalaBear, Keccak-f AIR, 2^14 rows) that comes to about 105 KB (2633 columns × 2 points × 5 limbs × 4 bytes), rehashed about 2^16 times at `STIR_BATCH_POW_BITS = 16`. The grind's span is at debug level, so at the default level it showed up as roughly 2 s of self time in `STIR PCS open`.

Both challengers now squeeze first, by sampling and discarding one byte, which hashes the pending input down to the chaining digest. `grind` searches from a squeezed clone. `check_witness` is overridden to squeeze before it absorbs the witness, so prover and verifier both check `H(H(pending) || witness)`. A candidate now costs one hash of the digest and the witness, however long the transcript is. Zero difficulty still accepts any witness without touching the transcript.

## Soundness

The witness is still bound to the whole transcript, because the digest it is hashed with covers everything observed so far. A grinding attempt costs one hash call either way. Before this change, a prover could already absorb the shared prefix once and reuse the sponge state for every candidate. So the old per-candidate cost was never part of the security margin.

## Breaking change

Every proof-of-work witness these challengers produce and accept changes, so proofs made before this PR do not verify after it.

The circle compat fixtures of `batch-stark` and `uni-stark` are regenerated with their `--ignored` generators. The two-adic fixtures use a duplex challenger and are unchanged.

Duplex-based challengers are not touched. Neither is `BinaryChallenger` in `binary-field`, which searches the same way and could get the same change separately.

## Measurements

`prove_prime_field_31 --field koala-bear --objective keccak-f-permutations --log-trace-length 14 --discrete-fourier-transform radix-2-dit-parallel --merkle-hash keccak-f --pcs stir` on an Apple M4 Pro with 14 threads, 6 alternating runs of each build, measured on e45e11ea1. Grinding time is geometric, so the ranges say more than the medians:

| | before: median (range) | after: median (range) |
| --- | ---: | ---: |
| batch grind | 1.26 s (70 ms – 4.01 s) | 3.6 ms (1.8 – 6.6 ms) |
| `STIR PCS open` | 1.43 s (282 ms – 4.19 s) | 187 ms (107 – 259 ms) |
| `prove` | 1.54 s (394 ms – 4.30 s) | 289 ms (209 – 375 ms) |

The FRI example stays at parity over 4 alternating runs. It grinds the same claims, but at `batch_proof_of_work_bits = 10`, where the grind cost about 15–20 ms before this change.

## Test plan

- [x] New `test_serializing_challenger{32,64}_grind_rehashes_pending_input_at_most_twice`. A counting hasher over a 4 KiB pending transcript asserts that one grind hashes the pending input at most twice: once for the search, once in the final check. The tests also check that a fresh verifier accepts the witness and that both transcripts stay in sync. Both failed before the change and pass with and without `parallel`.
- [x] `cargo test --workspace --release --no-fail-fast` before the rebase. The only failures were the two circle compat fixtures, with `InvalidPowWitness`, until they were regenerated.
- [x] After the rebase: `cargo test --release -p p3-challenger -p p3-batch-stark -p p3-uni-stark -p p3-stir -p p3-fri -p p3-circle`
- [x] `cargo clippy -p p3-challenger --all-targets --all-features -- -D warnings`, nightly `fmt`


tcoratger (approved):
Reviewed the transcript transition and grinding paths for both serializing challengers. The squeeze-first construction keeps the witness bound to the full pending transcript through the chaining digest, and prover/verifier advance to the same post-grind state. Zero-difficulty behavior remains unchanged. I also verified both regenerated circle fixtures and ran the p3-challenger tests with and without all features, clippy with all targets/features, and formatting checks locally. No blocking findings.

PR #2200, opened 2026-09-18
perf(challenger)!: batch the byte-hash proof-of-work search

Nashtare:
**Stacked on #2199.** The base is `perf/grind-squeeze-first`, so the diff below is only this PR's change. After #2199 lands, this branch will be rebased onto `main`.

With #2199, `SerializingChallenger32/64::grind` searches from a squeezed copy of the challenger, whose pending input is the 32-byte chaining digest. It still tested each candidate on its own clone of that copy. So every candidate paid two `Vec` allocations, a byte-at-a-time copy of the digest and witness into the hasher, and one single-message hash.

This PR runs the search through a new `ByteGrindingChallenger` trait on the inner challenger:

```rust
pub trait ByteGrindingChallenger: CanObserve<u8> + CanSample<u8> + Clone + Send + Sync {
    fn find_witness<const W: usize, const S: usize>(
        &self,
        num_candidates: u64,
        encode: impl Fn(u64) -> [u8; W] + Sync,
        accepts: impl Fn([u8; S]) -> bool + Sync,
    ) -> Option<u64> { /* clone, observe encode(c), sample S bytes, accepts */ }
}
```

The default method is the clone-per-candidate check, so any byte challenger works with an empty impl.

`HashChallenger<u8, H, OUT_LEN>` overrides it. Observing the non-empty witness bytes discards any buffered output, so candidate `c`'s samples all come from `H(input_buffer || encode(c))`, and sample `k` is `digest[OUT_LEN - 1 - k]`. From the squeezed state, `input_buffer` is the digest, so each candidate is one hash of `digest || witness`. Each worker:

- writes the pending input once into every lane's message slot;
- rewrites only the witness bytes for each batch;
- hashes `H::LANES` candidates per `hash_many` call;
- scans the lanes in candidate order, so a serial build still returns the smallest passing witness.

`W == 0` and `S > OUT_LEN` don't fit that layout (the buffered output survives, or a second flush is needed), so they take the default path.

`grind` runs the search on the squeezed copy's inner challenger (`squeezed.inner.find_witness(..)`). The closures match #2199's check: `encode` is exactly what `observe(F)` serializes, and `accepts` is the test `witness_passes` applies to the word `sample_bits` reads. `grind` still ends with `assert!(self.check_witness(bits, witness))`, #2199's squeeze-first check on the real transcript. `check_witness` and the verifier path are unchanged from #2199, so the set of accepted witnesses cannot change. A hasher whose `hash_many` disagreed with single-message hashing would make the prover panic; it could never produce an invalid proof. The pending input is still hashed at most twice per grind; #2199's rehash-count tests pass.

The second commit asserts `bits < usize::BITS` in `SerializingChallenger64::grind`, as the 32-bit version already does. Without it, a 32-bit target asking for 32+ bits would search about 2^32 candidates before the final check panicked.

## Breaking change

`GrindingChallenger for SerializingChallenger32/64` now requires `Inner: ByteGrindingChallenger`. A custom inner byte challenger needs one line:

```rust
impl ByteGrindingChallenger for MyInner {}
```

Every in-repo configuration uses `HashChallenger<u8, _, 32>`, which is covered. Stable Rust has no specialization, and the fast path needs `HashChallenger`'s private buffers. So the hook has to be a trait on the inner type. A default method on `CanSample`/`CanObserve` would avoid the new bound, but it would put a PoW-specific API on every challenger and still be a minor break. Proof-of-work witnesses are the same as #2199's: the circle compat fixtures regenerate byte-identical.

## Measurements

Apple M4 Pro (Keccak 2 lanes, SHA-256 1 lane, BLAKE3 4 lanes with `p3-blake3/neon`), BabyBear witness, 32-byte pending digest, on a machine shared with other builds.

Candidate checks per second, #2199's per-candidate check against the batched search:

| candidates/s | 1 thread | 14 threads |
| --- | ---: | ---: |
| Keccak | 3.50M → 14.4M (4.1x) | 17.1M → 129M (7.5x) |
| SHA-256 | 6.88M → 45.3M (6.6x) | 23.3M → 354M (15x) |
| BLAKE3 | 6.22M → 19.7M (3.2x) | 19.3M → 175M (9.1x) |

Mean `grind` time over fixed seeds, #2199 (`84c363885`) against this branch, medians of three alternating runs:

| 20 bits | 1 thread (8 seeds) | 14 threads (32 seeds) |
| --- | ---: | ---: |
| Keccak | 246 → 60.3 ms (4.1x) | 91.2 → 11.2 ms (8.1x) |
| SHA-256 | 222 → 40.4 ms (5.5x) | 51.4 → 4.0 ms (13x) |
| BLAKE3 | 149 → 46.1 ms (3.2x) | 64.8 → 8.9 ms (7.3x) |

At 16 bits the gains are 3.2–5.5x at one thread and 7.2–10.6x at 14. At one thread both return the same witness for every seed.

- **Why 14 threads gains more:** the per-candidate path allocates for every candidate and scales only about 5x on 14 threads.
- **BLAKE3:** its lanes don't help here. A 36- or 40-byte message has no full 64-byte block to batch, so its 3.2x comes from dropping the clone and hasher setup.
- **SHA-256:** a 36-byte message is one hardware SHA-2 compression, so it is the fastest here even at one lane.

## Test plan

- [x] `find_witness_matches_clones_*`: for Keccak, SHA-256, BLAKE3 and a Keccak wrapper reporting 1/3/4 lanes, over 24 transcripts, including freshly squeezed ones and long pending inputs. The pending-input lengths put the witness on block and chunk boundaries, with and without a partly consumed digest. For every candidate in the range, the override finds a clone's sampled bytes at exactly that index, including tails of one candidate and one short of a full batch. Shapes (4,4), (8,8), (4,32), and the fallbacks (0,4), (4,33), (4,40). `num_candidates = 0` returns `None`.
- [x] `test_grind_agrees_with_check_witness_{keccak,sha256,blake3,unbatched}`: both serializing challengers, bits {1, 2, 4, 8, 12}. The witness passes the squeeze-first `check_witness`, is the smallest one serially, and the next 12 samples match those after `check_witness` on a pre-grind clone.
- [x] #2199's `*_grind_rehashes_pending_input_at_most_twice` pass
- [x] Review fuzz, on the pre-rebase version (not committed): the override agreed with the clone reference on 8000 random challenger states (random observe/sample/flush histories, `OUT_LEN` ∈ {1,16,32,64}, 1–7 lanes, `W = 0`, `S = 0`, `S = OUT_LEN ± 1`)
- [x] `cargo test --release -p p3-challenger`, with and without `parallel` (207 tests)
- [x] `cargo test --release -p p3-uni-stark -p p3-batch-stark`, including both `verify_circle_compat_fixture` tests; the ignored fixture generators reproduce #2199's fixtures byte for byte
- [x] clippy `-D warnings` with and without `--all-features`, nightly `fmt`


tcoratger (approved):
Reviewed the batched candidate search against the clone/observe/sample semantics. For non-empty encodings whose samples fit one digest, the fast path hashes exactly input_buffer || encode(candidate), reconstructs the reverse-pop sample order correctly, handles partial lane batches, and preserves the serial smallest-witness guarantee. The empty-encoding and multi-flush shapes correctly use the reference fallback; the final check_witness still validates and advances the real transcript. The usize-width guard also prevents an impractical 32-bit search before the eventual sample_bits assertion. Release challenger tests pass with and without parallelism, downstream uni-stark and batch-stark fixtures pass, workspace all-target/all-feature checking and strict clippy are clean. No blocking findings.

PR #2202, opened 2026-09-18
feat(errors)!: make diagnostics actionable

tcoratger:
## Summary

- derive `thiserror::Error` for developer-facing errors that previously exposed only `Debug`, including the shared example runners
- replace raw string construction errors with typed errors for cross-field sponges, the multi-field challenger, and Poseidon2 round selection
- retain nested MMCS, PCS, and transcript diagnostics in wrapper messages, and attach expected/actual context to shape failures
- test exact messages and error-source chains so profiling and CLI failures remain actionable

## Breaking changes

Several public error return types and variant payloads now carry structured diagnostic context. Downstream exhaustive matches may need to accept the new fields or typed errors.

## Validation

- `cargo test -p p3-challenger -p p3-examples -p p3-poseidon2-air -p p3-poseidon2 -p p3-symmetric -p p3-security -p p3-stir -p p3-whir -p p3-merkle-tree -p p3-binary-pcs`
- `cargo clippy -p p3-challenger -p p3-examples -p p3-poseidon2-air -p p3-poseidon2 -p p3-symmetric -p p3-security -p p3-stir -p p3-whir -p p3-merkle-tree -p p3-binary-pcs --all-targets --all-features -- -D warnings`
- `cargo fmt --all`
- `git diff --check`


examples/src/proofs.rs
```diff
@@ -81,12 +79,18 @@ fn example_circle_parameters<EF: Field, M>(mmcs: M) -> FriParameters<M> {
...
-pub enum ProofRunError<P: Debug, V: Debug> {
+#[derive(Debug, thiserror::Error)]
+pub enum ProofRunError<P, V>
+where
+    P: core::error::Error + 'static,
+    V: core::error::Error + 'static,
+{
     /// Proof generation rejected its configuration or opening budget.
-    Prove(P),
+    #[error("proof generation failed: {0}")]
```
Nashtare, line 85: `{0}` puts the inner error in the message, and `#[source]` also returns it from `source()`, so anything that walks the chain prints it twice. I built an enum with the same shape locally, and a chain printer gives `proof verification failed: inner: inner`. anyhow's `{:?}` would repeat it under "Caused by". `BinaryProofError` in `binary.rs` does the same in all four wrapping variants, and its test asserts both halves. `report_result` and `prove_hash_binary` only print `Display`, so could these keep `{0}` and drop `#[source]`/`#[from]`, with plain `From` impls? Keeping the sources and having the runners print the chain would also work.

A separate point: thiserror already adds `P: Error + 'static` to the `Error` impl it generates. The same enum with no where clause compiles and implements `Error`, so the where clause isn't needed. Dropping it would let the type accept Debug-only parameters again, as it did before.

challenger/src/multi_field_challenger.rs
```diff
@@ -85,19 +119,33 @@ where
...
+    /// the transcript's one-byte length tag.
+    pub fn new(permutation: P) -> Result<Self, MultiField32ChallengerError> {
         if F::order() >= PF::order() {
-            return Err(String::from("F::order() must be less than PF::order()"));
+            return Err(MultiField32ChallengerError::FieldOrderNotIncreasing {
+                input_field: type_name::<F>(),
+                sponge_field: type_name::<PF>(),
+            });
         }
         if RATE >= WIDTH {
```
Nashtare: Nothing can reach this branch. `new` ends in `DuplexChallenger::new`, whose `const { assert!(RATE > 0 && RATE < WIDTH && RATE <= 255) }` runs whenever this function is instantiated. Locally, `MultiField32Challenger::<BabyBear, Goldilocks, _, 8, 8>::new` fails to compile with E0080 on that assertion, so it never returns `RateNotSmallerThanWidth`. The test at line 531 says as much and builds the variant by hand. Could the branch and the public variant be dropped, or replaced by a const assert here? As it stands, the API promises an error that can never be returned.

poseidon2/src/round_numbers.rs
```diff
@@ -26,6 +26,40 @@
...
+    )]
+    InvalidSboxExponent {
+        /// Requested S-box exponent.
+        exponent: u64,
+        /// Multiplicative-group order the exponent must be coprime to.
+        field_order_minus_one: u64,
+    },
+    /// No audited round count exists for this width and exponent.
+    #[error(
+        "no audited Poseidon2 round count for a {field_bits}-bit field with width {width} and S-box exponent {exponent}"
```
Nashtare: Nit: according to the module doc, this table was computed with `calc_round_numbers.py` plus the monotonicity argument, so "audited" isn't quite right. "no precomputed round count" would be more accurate, here and in the other two places that say "audited".

security/src/grinding.rs
```diff
@@ -626,66 +642,30 @@ pub enum GrindingMismatch {
...
-}
-
 #[cfg(test)]
 mod tests {
+    use alloc::string::ToString;
+
     use super::*;
 
+    #[test]
+    fn grinding_mismatch_is_an_error_without_changing_its_message() {
```
Nashtare: Nit: `without_changing_its_message` only makes sense next to the old manual impl. A name like `grinding_mismatch_is_an_error_naming_site_and_counts` would describe the test on its own terms.

Nashtare (approved):
I went through every function body the diff touches, looking for behaviour hidden inside the error changes, and found none. `GrindingMismatch` only swaps its manual `Display` for `#[error]` attributes. A probe that renders 300 values (every site and variant, odd strings, `usize::MAX`) through the old impl and the derived one gets identical text. For `poseidon2_round_numbers_128` I compared the merge-base function with the new one for BabyBear, KoalaBear, Mersenne31 and Goldilocks over widths and exponents 0..=256. I also ran copies that take the field order as an argument over every bit length 2..=64, about 50.5M (order, width, exponent) triples. The `Ok` values match exactly, and every old string error maps to the matching typed variant with the right payload. In the hiding MMCS, `check_widths` still runs before salting, and the same pairs are length-checked in the same order. The Merkle, challenger and sponge changes only build different error values, and WHIR's new `InvalidPowWitness` fields come straight from `query_pow`.

`thiserror` comes in through the workspace dependency with `default-features = false`. p3-symmetric, p3-poseidon2, p3-poseidon2-air, p3-security, p3-challenger, p3-merkle-tree, p3-stir, p3-whir and p3-binary-pcs all build for `aarch64-unknown-none-softfloat`. Locally the tests listed in the description pass (1108 passed, 0 failed), and clippy with `--all-features -D warnings` is clean. The branch merges cleanly into current main, and the merged workspace passes `cargo check --all-targets`.

The inline comments are non-blocking follow-ups.

**Smaller things**
- The breaking-changes section could list a few more: `RoundConstantsShapeError::FullRounds` is removed, not just reshaped; `report_result` takes `impl Display` instead of `impl Debug`; `ProofRunError` now requires `Error + 'static` parameters; and `MerkleTreeError::WrongBatchSize` and WHIR's `InvalidPowWitness` change from unit variants to struct variants.
- binary-pcs still maps `TranscriptFailure::PowWitness { .. }` to a unit `InvalidPowWitness`, so it drops the difficulty that WHIR now keeps. This is optional, but it would match the rest of the PR.

PR #2203, opened 2026-09-18
perf(multi-stark)!: run the binary zerocheck's first rounds sixty-four rows at a time

Nashtare:
On a binary trace, the multi-STARK zerocheck costs about six times the prime field's whole constraint stage. At Keccak `-l 14` on an M4 it takes about 112 ms, against 18 ms for KoalaBear's `quotient_values`. Most of that is not field arithmetic:

- Every bit is a 16-byte `BinaryField128` cell.
- `Table` stores each column contiguously, so columns sit exactly 256 KiB apart.
- Round 0 evaluates the AIR one row at a time. Each row gathers about 6,450 cells from as many pages, so the round streams about 850 MB for 3.3 MB of bits.
- The `Ghash128` rounds after it then evaluate the AIR row by row over roughly as many rows again.

This PR runs the first zerocheck rounds of a binary stage on bit planes, sixty-four rows at a time. Every round polynomial and opening is the one `GenericBackend` computes, so proofs are byte-identical.

## Round 0 on bit planes

When a stage fits `GF(4)`, it is repacked once into two coordinate planes per 64 rows (`multi-stark/src/rounds/sliced.rs`). "Fits" means the existing subfield conditions hold, and each half of the stage holds at least 64 rows. The planes cover:

- the cells;
- repeat-last successor planes for the declared next-row columns;
- first-row, last-row and transition selector planes.

The AIR then runs once per 64 rows through a new `SlicedFolder` on `SlicedGf4` values (`multi-stark/src/sliced.rs`):

```text
    lane i  =  (bit i of low) + (bit i of high) * g,      g^2 = g + 1
    a * b   =  three ANDs and four XORs, every lane at once
```

Each asserted constraint leaves the lanes through byte-indexed partial sums of the eq weights of the six lane variables (`LaneSums`: 16 lookups), then meets its alpha power. The word variables' eq weight is applied once per word. An AIR constant outside `GF(4)` poisons the value, as `SubfieldVar` already does, and the round falls back to the scalar subfield kernel, then to the generic one.

## Rounds 1 and 2 on the same planes

A degree-`d <= 3` round polynomial interpolates over `d + 1` nodes, and those nodes are `GF(4)`. So round `j` evaluates the AIR at points whose first `j` coordinates are nodes rather than the challenges already drawn, then interpolates:

```text
    q_j(t) = sum_x eq(x) * C(r_0, .., r_(j-1), t, x)
           = sum_v L_v(r_0, .., r_(j-1)) * sum_x eq(x) * C(v, t, x)        v in {nodes}^j
```

`C` has degree at most `d` in each variable, since every column, successor, periodic column and selector is multilinear. The interpolation is therefore exact.

A sliced stage no longer folds its columns after round 0. `ExtColumns::Sliced` keeps the planes and records each challenge. Once `SLICED_ROUNDS = 3` rounds are spent, `unslice` folds every column from the planes at all bound challenges at once, one byte-indexed subset sum of `eq(r, .)` per plane per row. It reads the repeat-last tails off the successor planes. This replaces round 0's fold of the whole trace into `Ghash128` (213 MB at Keccak `-l 14`) with one of an eighth of the rows.

`SubfieldBackend` accumulates in the challenge field and `ReprBackend` in its representation field. A stage that does not fit `GF(4)`, is too short, declares lookups, or has degree above 3 keeps today's kernels.

## Breaking change

`SubfieldBackend` and `ReprBackend` now also require `A: for<'a> Air<SlicedFolder<'a, F, S, R>>`. Any AIR written against `AirBuilder` meets it. Only code that restates the backend bounds changes, as `examples/src/binary.rs`'s `BinaryAir` does here. `SlicedFolder` implements the lookup builders as no-ops, like `MultilinearFolder`, because a stage declaring lookups never runs it.

## Measurements

The spans are `prove_with_lookup` (the whole zerocheck) and the example's prove time. These are medians of 4 to 7 interleaved runs against `main` (`e45e11ea1`), `--profile optimized`, `-C target-cpu=native`, `--features parallel`, `--security-bits 96`. Proof digests match `main` in every run.

AWS c8a.4xlarge (Zen 5, 16 vCPU):

| | zerocheck | prove |
| --- | ---: | ---: |
| Keccak `-l 14` | 124 → 43 ms (−65%) | 0.573 → 0.494 s (−13.8%) |
| Keccak `-l 16` | 418 → 135 ms (−68%) | 2.145 → 1.850 s (−13.8%) |
| BLAKE3 `-l 10` | 63 → 36 ms (−43%) | 0.303 → 0.275 s (−9.2%) |
| BLAKE3 `-l 12` | 170 → 64 ms (−62%) | 1.055 → 0.949 s (−10%) |

On an M4 Pro (shared, noisy), Keccak `-l 14` goes from 126–133 ms to about 50 ms.

What remains of the 43 ms at Keccak `-l 14` on the c8a:

- 6 ms packing the planes, bound by reading the 426 MB trace;
- about 10 ms for the sliced rounds 1–2;
- 1.5 ms for `unslice`;
- about 16 ms for the `Ghash128` rounds from round 3 on;
- about 6 ms of serial `get_air_profile`, a symbolic pass over the AIR redone on every proof. Caching it belongs in a separate PR.

Tried and not kept:

- Four sliced rounds: zerocheck 48–50 ms, against 41 ms with three.
- Folding the alpha products into `Ghash128::dot_product` batches: no measurable change on either machine.

## Tests

- `SlicedGf4` arithmetic is checked lane-wise against `BinaryField2` (proptest). The `LaneSums` sums, the coordinates, poison, and the ring constants are also checked.
- Both planes of every cell and successor are checked against the table, on random `GF(4)` cells.
- The sliced round 0 matches the generic kernel at 128 to 4096 rows, for gate, pair, gate+pair and periodic fixtures, in both accumulation fields.
- Every round polynomial and the openings match with 1, 2 and 3 sliced rounds.
- Transcript agreement across `GenericBackend`, `SubfieldBackend` and `ReprBackend` covers:
  - tall stages, including one that activates later;
  - grinding;
  - every misfit at 256 rows;
  - a degree-1 fixture whose constant first poisons round 1, the later-round fallback.
- In debug builds, a check that the caller's eq table factors into the word and lane weights built from `tau`.


multi-stark/src/sliced.rs
```diff
@@ -0,0 +1,714 @@
...
+        0
+    }
+
+    fn num_indexed_tables(&self) -> usize {
+        0
+    }
+}
+
+#[cfg(test)]
+mod tests;
```
tcoratger, line 714: Better to have this at the beginning of the file no?

tcoratger (approved):
Reviewed the sliced GF(4) arithmetic, first- and later-round interpolation, fallback paths, successor/tail handling, unslicing, transcript equivalence, and the new differential tests. I found no correctness, soundness, or performance issue requiring a change. I also ran the complete p3-multi-stark all-features test suite locally (276 unit tests plus integration/example suites passed).

PR #2039, opened 2026-09-05
perf(maybe-rayon): size parallel tasks from a cost model

tcoratger:
Replaces fixed element thresholds with a shared cost model for Rayon task sizing.

## Why

A threshold such as 4096 ignores element width and pool size. It represents 16 KiB for a 4-byte field, 256 KiB for a packed extension, and the same dispatch decision on 2 or 96 workers.

The new policy prices the bytes one item reads and writes:

```text
priced loop < platform gate * workers  -> keep the loop whole
otherwise                              -> permit bounded tasks
```

The per-worker gate is 0.625 us outside macOS, calibrated on Linux through 96 workers, and 2.5 us on macOS, calibrated on Apple silicon. A one-worker pool never splits, matching a build without the `parallel` feature.

## API

`p3-maybe-rayon` provides:

- a minimum task length for callers that construct chunks;
- a split predicate for algorithms whose two arms differ;
- iterator terminals that bypass Rayon when the policy keeps a loop whole;
- a process-wide budget override, with `Default` exposing the compiled-in settings.

`P3_MIN_PARALLEL_NS` and `P3_MAX_TASK_NS` remain available for reproducible measurements.

## Scope

The PR migrates existing threshold or chunk-count decisions in `p3-multilinear-util`, `p3-stir`, `p3-field`, `p3-lookup`, and `p3-matrix`. Arithmetic-bound kernels carry explicit charges instead of being priced as memory traffic alone.

## Performance evidence

On a 32-thread Ryzen 9 9950X3D, binding every variable of a degree-four extension polynomial improved by 18.2x at 2^10 evaluations, 7.8x at 2^12, 2.6x at 2^14, and 1.5x at 2^20. Representative gains include 2.91x for a 2^12 suffix fold, 2.00x for split-equality extension evaluation, and 2.69x for batch inversion.

Review measurements exposed two host-specific limits and changed the final policy:

- Cache-sized Horner chunks regressed up to 2.18x on a dedicated 96-core Graviton5. STIR now uses the shared gate but caps the split at one chunk per worker.
- The Linux dispatch rate underpriced Apple silicon by roughly 4x. macOS now has its own measured gate.

The Graviton measurements put fan-out at 0.33–0.47 us per worker through 96 workers. The 0.625 us non-macOS gate retains margin before splitting.

## Safety and regression coverage

- The floor remains in `1..=max(len, 1)` across worker counts, widths, empty inputs, and saturating sizes.
- Structurally different serial and parallel arms are tested against each other.
[...]

PR #2204, opened 2026-09-18
perf(fri): pack the multi-round fold and drop a multiplication from the opening quotient

Nashtare:
Two changes to the two-adic FRI prover's hot loops. Proofs are unchanged.

## 1. Packed multi-round fold

`fold_matrix` folded one extension element at a time. Its arity > 2 path also ran each arity-2 round over the whole vector, with an intermediate buffer per round and a twiddle table as long as the first round's output.

For base fields of at most 32 bits, `fold_matrix_packed` folds `F::Packing::WIDTH` rows at a time, one row per lane of `EF::ExtensionPacking`, and carries each block through all `log_arity` rounds without unpacking. In round `s`, the scalar paths read the halving factor at index `j = t * (arity >> (s+1)) + k` of the bit-reversed `g_inv^i / 2` table. Since `rev_N(j) = 2^s * (rev(k) * H + rev_H(t))` for `j < N / 2^s`, that factor splits into

    g_inv^(2^s * rev_H(t))      per row, squared once per round
    w_inv^(2^s * rev(k)) / 2    per pair, merged with beta^(2^s)

and the per-row factor splits further into a per-block scalar times a constant packed lane vector (`rev_H(b*W + l) = rev_W(l) * (H/W) + rev(b)`). The only table left is `height / WIDTH` long.

- **Gate:** the packed path runs when `WIDTH > 1` is a power of two, `height >= WIDTH` and `F::bits() <= 32`. Everything else uses the unchanged scalar paths. `PackedField` does not bound the width, and the lane bit-reversal and whole-block split both need a power of two.
- **Why 64-bit fields are excluded:** on NEON without SVE2, the Goldilocks packing multiplies lane by lane in scalar registers. There the packed kernel measured 6–48% slower than the scalar paths for Goldilocks + quadratic (2^14..2^20).
- **Not measured:** x86 AVX2/AVX-512 Goldilocks keeps the scalar path as before.

## 2. One fewer extension multiplication in the opening quotient

For every row of every (matrix, point) pair, `open` accumulated `ro += alpha_pow_offset * (Mred(z) - Mred(x)) * inv_denom`, at two extension multiplications per row. The packed alpha powers that compress a matrix are now scaled by the offset of the first point it is opened at. So `mat_compressed` holds `alpha^offset * Mred(x)`, and that point's rows cost one multiplication. Later points of the same matrix rescale by `alpha^(point_index * width)` and keep two. Scaling the powers costs `width / WIDTH` packed multiplications per matrix.

Packing the accumulation loop itself was measured and left out. On an M4 Pro, NEON packed quartic multiplication is barely cheaper than four scalar ones, so the transposes around it cost more than it saves:

| BabyBear + quartic, 2^18 rows, 1 thread | scalar | packed per chunk | `ro`, `inv` kept packed |
| --- | ---: | ---: | ---: |
| two multiplications | 1623 µs | 1682 µs | 1450 µs |
| one multiplication (this PR) | 754 µs | 1059 µs | 810 µs |

## Equivalence

Field arithmetic is exact, so the folded vectors and reduced openings are unchanged.

- **Proof bytes:** identical to `main` (postcard encodings of opened values and proofs, plus a challenge sampled after `open`), at `max_log_arity` 1–4, `log_blowup` 1–2, for BabyBear^4, BabyBear^5 and KoalaBear^4, serial and at 1 and 14 threads. The batch shape covered multiple commitments, several matrices of one height, up to 3 points per matrix, and mixed heights.
- **Fold equivalence, from review:** an independent test ran 4,230 random cases of the old `fold_matrix` against the new one: β ∈ {0, 1, random}, arities 2–64, heights 2^0–2^14, dense and bit-reversed views, BabyBear/KoalaBear with extension degrees 1, 4, 5 and 8, and Goldilocks^2. All were equal, serial and parallel, with debug assertions on.
- **The factor decomposition:** it relies on `two_adic_generator(a + b)^(2^b) == two_adic_generator(a)`, which `fold_row` and the verifier already rely on.

## Measurements

Apple M4 Pro, min over 3–4 alternated runs of a standalone timing loop. The host was shared with other builds, so absolute numbers are noisy.

Fold, BabyBear + quartic, input of 2n elements, n = 2^14..2^22:

| arity | 1 thread | 14 threads |
| ---: | ---: | ---: |
| 2 | −32% .. −37% | −27% .. −63% |
| 4 | −21% .. −30% | −50% .. −64% |
| 8 | −27% .. −34% | −55% .. −68% |

Goldilocks + quadratic stays within ±3%, since it takes the same scalar path. A second run during review, under heavier load, measured −7% to −18% at 1 thread and −20% to −65% at 14 threads: the same direction, but smaller at 1 thread.

`open` (new `pcs_open` bench: a width-64 matrix at two points and two width-4 matrices at one point, BabyBear + quartic, `log_blowup` 1, no grinding), change 2 alone:

| | 1 thread | 14 threads |
| --- | ---: | ---: |
| `max_log_arity` 3, 2^14..2^20 | −4% .. −8% | 0% .. −4% |
| `max_log_arity` 1 | −2% .. −3% | −1% .. −3% |

End to end (`prove_prime_field_31`, KoalaBear, Poseidon2 AIR and Merkle, `-l 16/18`), the FRI commit phase is 10–26% faster and `open` 2–6% faster. Whole-proof time is within noise, since trace commitment is about 73% of `prove`.

## Test plan

- [x] `fold_matrix_matches_fold_row` and `fold_matrix_packed_matches_fold_row`: arities 2–16, heights below, at and above the packing width, for BabyBear + quartic, BabyBear itself and Goldilocks + quadratic
- [x] `packed_fold_gate`: widths 3 and 6, width 1, heights below the width and 64-bit fields keep the scalar path
- [x] New `matrices_opened_at_up_to_three_points` roundtrip. It is the only test that catches a wrong `alpha^(point_index * width)` rescale.
- [x] Mutation checks: dropping the lane bit-reversal, the per-round squaring or the rescale makes these tests fail
- [x] `cargo test --release -p p3-fri`, with and without `parallel`; `-p p3-uni-stark -p p3-batch-stark -p p3-examples`
- [x] `cargo clippy -p p3-fri --all-targets --all-features -- -D warnings`, nightly `fmt`
- [x] `fold_even_odd` gains BabyBear + quartic and Goldilocks + quadratic groups with explicit names. The old names came from `pretty_name::<EF>()` alone, so different base/extension pairs collided into criterion's `#N` ids.


fri/src/two_adic_pcs.rs
```diff
@@ -235,6 +245,103 @@ impl<F: TwoAdicField, InputProof: Sync, InputError: Debug + Sync, EF: ExtensionF
...
+        .collect();
+
+    let fold =
+        |lo: EF::ExtensionPacking, hi: EF::ExtensionPacking, scalar: EF, row_factor: F::Packing| {
+            (lo + hi).halve() + (lo - hi) * scalar * row_factor
+        };
+
+    let mut folded = EF::zero_vec(height);
+    folded
+        .par_chunks_exact_mut(width)
```
tcoratger, line 315: I think that here is a good example where https://github.com/Plonky3/Plonky3/pull/2039 will be useful to automatically discard the parallel path when size becomes too small

fri/src/two_adic_pcs.rs
```diff
@@ -144,6 +144,16 @@ impl<F: TwoAdicField, InputProof: Sync, InputError: Debug + Sync, EF: ExtensionF
 
     #[instrument(skip_all, level = "debug")]
     fn fold_matrix<M: Matrix<EF>>(&self, beta: EF, log_arity: usize, m: M) -> Vec<EF> {
+        // Packed arithmetic pays off where SIMD lanes multiply field elements natively,
+        // which holds for fields of at most 32 bits. Wider fields emulate the lane product
+        // on common targets, so they keep the scalar paths, as do width-1 packings and
+        // heights below one packing width. Heights are powers of two, so any larger height
+        // splits into whole packed blocks.
+        let width = F::Packing::WIDTH;
+        if width > 1 && m.height() >= width && F::bits() <= 32 {
```
tcoratger, line 153: **[MAJOR] Guard the power-of-two packing precondition.** `F::Packing` is only bounded by `PackedField`; unlike `PackedFieldPow2`, that trait does not guarantee `WIDTH` is a power of two. A valid custom field packing with `WIDTH = 3` and a height-4 matrix passes this gate, but `fold_matrix_packed` then calls `log2_strict_usize(3)` (and its whole-block assumption also fails). Please either add `width.is_power_of_two()` to this runtime gate so unsupported packings retain the scalar path, or encode the stronger bound in the API. A width-3 test packing would pin the fallback behavior.

tcoratger (commented):
The optimized identities and opening rescaling look sound. I found one generic packing-width precondition that needs to be made explicit in the dispatch gate.

fri/src/two_adic_pcs.rs
Nashtare: Good catch, fixed in 5fa29b7f3. The gate now also requires `WIDTH.is_power_of_two()`, so any other packing keeps the scalar paths. I moved the gate into `use_packed_fold` and pinned it with `packed_fold_gate` (widths 3 and 6 against heights 4 and 8, plus width 1, height below the width and 64-bit fields). No built-in `PackedField` has a non-power-of-two width, so testing the predicate directly avoids building a fake field for it. The packed kernel's precondition and debug assertion now include the width requirement too.

tcoratger (approved):
Reviewed the current head deeply. The packed fold decomposition matches the scalar FRI fold, the opening rescaling preserves the verifier's per-height alpha ladder, and the non-power-of-two packing fallback is now correctly guarded and tested. Focused release tests pass in serial and parallel modes; CI is green.

PR #2205, opened 2026-09-18
perf(matrix): resolve each lane's inner row once in RowIndexMappedView::vertically_packed_row

Nashtare:
The uni-stark and batch-stark quotient loops pack the trace through `BitReversedMatrixView` (FRI, STIR, hiding FRI) or the CFFT-ordered view (circle). Both are `RowIndexMappedView`s, so the quotient hot path runs `RowIndexMappedView::vertically_packed_row`, not the `DenseMatrix` version. Its column closure called `self.get_unchecked` for every element, which re-mapped the row index and reloaded the inner matrix's pointer, width and permutation parameters through `self`. In both provers the fold that consumes the iterator is outlined and only sees `self` through the closure, so LLVM could not hoist those loads past the stores into the output. Every packed column paid `P::WIDTH` row mappings plus reloads, and even the 4-lane path lost its `st4` transpose.

For packings of at most 8 lanes (`MAX_RESOLVED_LANES`), each lane's inner row slice is now resolved once when the iterator is built and captured by value, so the column loop only reads through those slices. On NEON that is `ldr q` + `st4` for 4 lanes, and 8 loads + 4 `stp` for the 8-lane lockstep packing. The gate is a compile-time `P::WIDTH <= 8`, because a runtime gate made the closure outline per column. Wider packings keep the per-element path. `vertically_packed_row_pair` chains two `vertically_packed_row` calls. The second commit asserts that every resolved lane row spans the view's width, once per lane at construction. The column loop indexes those rows unchecked, so an inner `Matrix` whose rows are too short panics instead of reading out of bounds.

## Equivalence

Output is unchanged. `row_slice_unchecked(lane_row)` maps the row through `map_row_index` exactly as `get_unchecked(lane_row, c)` did, and wrap-around stays modulo the view height. This holds for every `RowIndexMap` in the workspace: `BitReversalPerm`, `CfftPerm`, `VerticallyStridedRowIndexMap` (whose view height differs from the inner height), and the test maps.

Tests compare both methods with a scalar gather at every start row up to `2 * height + WIDTH`, wrapping included. They cover:

- reversed, bit-reversed and strided views, plus a view over `HorizontalPair`, which goes through the default, collecting `row_slice` path;
- widths 1–13 and heights 1–32;
- BabyBear and Goldilocks with their packings, width 1, `Vectorized<_, 2>`, and 3- and 16-lane `FieldArray`.

A `should_panic` test covers an inner matrix with short rows.

## Measurements

Apple M4 Pro (NEON), shared with other builds, so treat these as noisy; medians.

`quotient_values`, `prove_prime_field_31`, 1 thread, 7 alternating runs:

| workload | main | this PR | |
| --- | ---: | ---: | ---: |
| Keccak BabyBear 2^14, FRI (8-lane lockstep) | 174 ms | 160 ms | −8% |
| Keccak Mersenne31 2^14, circle (4-lane) | 212 ms | 174 ms | −18% |
| Poseidon2 BabyBear 2^16 | 497 ms | 435 ms | −12% |
| Blake3 KoalaBear 2^11, STIR | 75.7 ms | 68.9 ms | −9% |

Whole-proof time moves 0.5–2%, since Merkle hashing dominates.

`bitrev_vertically_packed_row_sweep` (new group): BabyBear 4-lane −31% to −66%, `Vectorized<BabyBear, 2>` −5% to −27%, Goldilocks −9% to −44%. On wasm32 + simd128, in an interleaved harness, Goldilocks (2 lanes) is −16% to −76% under wasmtime and −4% to −66% under V8, and unpacked BabyBear is −42% to −78%.

## Known costs and limits

- **Width-1 matrices** with the 8-lane BabyBear or 4-lane Goldilocks lockstep packings are about 10% (≈0.6 ns per call) slower: resolving the lanes costs more than the single column they serve. It did not show up end to end.
- **x86 was not measured.** From the generated assembly, AVX2 BabyBear hoists the row pointers, and AVX2/AVX-512 Goldilocks becomes a register transpose where it was scalar code or gathers. AVX-512 31-bit fields (16 lanes) and `Vectorized<_, 2>` on AVX2 (16 lanes) are above the gate and keep the per-element path.
- **Resolved path assumes a borrowing inner.** It expects the inner matrix's `row_slice_unchecked` to borrow. The trait default, `FlatMatrixView` and `HorizontalPair` collect into a `Vec`, so a view over them would allocate per lane. No in-tree view hits this: every `EvaluationsOnDomain` wraps a `RowMajorMatrixCow` or `RowMajorMatrix`.
- **Height 0 with width > 0** now panics when the iterator is built rather than on its first element. It panicked before as well; `DenseMatrix::vertically_packed_row` already panics eagerly.

A W×W tile transpose through backend shuffles measured another ~4% on the 8-lane path in a throwaway NEON prototype. It would need a new `PackedValue` hook, so it is not part of this PR.

## Test plan

- [x] `cargo test --release -p p3-matrix` (188), plus debug `--lib`
- [x] `cargo test --release -p p3-uni-stark -p p3-batch-stark -p p3-circle -p p3-fri` (405)
- [x] p3-matrix tests on wasm32-wasip1 + simd128 under wasmtime
- [x] `cargo clippy -p p3-matrix --all-targets --all-features -- -D warnings`, nightly `fmt`


tcoratger (approved):
Reviewed the resolved-lane path against the scalar gather and the Matrix safety contracts. Row mapping, wraparound, arbitrary packing widths, short-row validation, and the wider-packing fallback are handled correctly. The full p3-matrix all-features suite passes locally (188 tests).

PR #2206, opened 2026-09-18
perf(multi-stark): cache AIR profiles at setup

tcoratger:
## Summary

Compute each AIR symbolic profile once during setup and retain it in the matched proving and verifying keys. Reused keys no longer walk the AIR symbolically before every zerocheck proof and verification.

This changes neither the transcript nor the proof. The cached values are the constraint count and the ordinary/interaction degrees already fixed by the setup AIR batch.

## E5 audit

- **5.1 — friendly challenges:** already shipped in #2159; no duplicate work.
- **5.2 — elide the C side:** specific to Flock’s batch-R1CS split. Plonky3’s AIR zerocheck has no separately materialized `a · b - c` sides to elide.
- **5.3 — circuit walking:** specific to Flock’s lincheck matrix. The current multi-STARK path evaluates AIR constraints directly and has no substituted sparse matrix.
- **5.4 — packed round kernels:** #2203 covers the material early-round kernel. This PR removes the remaining invariant symbolic pass from repeated proof and verification.
- **5.5 — multipoint ring-switch batching:** the existing bit ring switch already batches its row identities into one degree-two sumcheck and the Boolean PCS discharges the resulting claims in one multi-opening. Flock’s newer capacity-free assist depends on the still-missing jagged/no-pad stacking path.
- **5.6 — AG-code skip:** remains research. Flock’s implementation is tied to its R1CS skip protocol and generated AG-code constants; importing it before the corresponding reduction exists here would add an unauditable, unused protocol.

## Benchmark

`optimized`, Keccak binary AIR, log height 10 (40 permutations), 95-bit target, 20 interleaved runs on the same machine:

| phase | main | this PR | change |
|---|---:|---:|---:|
| prove | 323.30 ms | 321.40 ms | -1.90 ms |
| verify | 30.75 ms | 26.45 ms | -4.30 ms |

This is a fixed per-AIR cost, not a throughput-scaled kernel win. Setup moves the work outside the timed proof path, so the saving is amortized when a key is reused; first-proof setup plus proving is intentionally unchanged.

## Validation

- `cargo clippy -p p3-multi-stark --lib --examples -- -D warnings`
- `cargo test -p p3-multi-stark --lib zerocheck` (55 passed, 1 heavy test ignored)
- `cargo test -p p3-multi-stark --example prove_binary_field` (9 passed)
- `cargo test -p p3-multi-stark --example prove_boolean_trace` (9 passed)


tcoratger:
Superseded by #2215. The replacement uses the same commit on an upstream branch and is layer 2 of native stack #2218.

PR #2207, opened 2026-09-18
feat(sha256-air): add a binary-field SHA-256 AIR and prove it in prove_hash_binary

### LINKED ISSUE #2186: feat(binary): add support for proving SHA in `prove_hash_binary` example
We currently only support Keccak and Blake3.

gopikannappan:
Closes #2186, the follow-up suggested in #2185.

This adds SHA-256 to the binary-field hash example. There's a new AIR, `Sha256BinaryAir`, that proves SHA-256 compressions over a field of characteristic 2, and `prove_hash_binary` gains a third objective, `sha-256-compressions`, alongside Keccak-f and BLAKE3. It proves and verifies end to end with the binary PCS.

## How it's built

It follows `Blake3BinaryAir`: one compression per row, no reads from the next row, every constraint of degree at most 2, and only the input bits (chaining value and message block) constrained to be 0 or 1. Every other column is forced to a bit by the constraint that defines it.

Three choices keep the row narrow:

- **Only the two new words of each round are stored.** A SHA-256 round writes new `a` and `e` words, and the other six just shift down. Like the prime-field AIR, this stores `a` and `e` as chains and reads `b, c, d` and `f, g, h` as earlier chain entries, so they cost no columns. That's also why this uses one row per compression rather than one per round: a per-round layout would have to copy all eight state words and the 16-word message window into every row.
- **Additions store as few carries as possible.** Adding several 32-bit words stores the carry bits of every partial sum except the last one, which reads its carries off the stored result. That's the same trick `Blake3BinaryAir`'s `add2` uses, generalized to any number of words.
- **`Ch` and `Maj` are stored** (32 bits each per round), because they're degree 2 and the additions need linear inputs.

The row comes to **23,712 columns and 23,712 constraints**, about 2.06× `Blake3BinaryAir`'s 11,536. Inputs use the same `[block | chaining value]` layout as the prime-field `generate_trace_rows`, and the output chaining value is stored (`h_out`) so later rows or AIRs can use it.

## Results

`prove_hash_binary --security-bits 96`, Apple M2, re-measured on merged main (two runs each):

| | Prove | Verify | Proof size |
|---|---:|---:|---:|
| 2⁸ SHA-256 compressions | 0.105 / 0.107 s | 0.075 / 0.085 s | 799,326 B |
| 2⁸ BLAKE3 compressions | 0.045 / 0.043 s | 0.032 / 0.034 s | 517,611 B |
| 2¹⁰ SHA-256 compressions | 0.152 / 0.164 s | 0.075 / 0.073 s | 856,139 B |
| 2¹⁰ BLAKE3 compressions | 0.069 / 0.067 s | 0.036 / 0.035 s | 597,211 B |

Proving costs about 2.2–2.4× BLAKE3, in line with the 2.06× width ratio, and the proof is 1.43–1.54× larger.

These are much faster than the numbers I first posted (2.66–3.21 s at 2¹⁰). That is main's doing, not this PR's: #2227, #2228 and #2245 moved these objectives onto the Boolean commitment while this was open. This branch now merges main and takes that path too, through a new `generate_binary_trace_packed` mirroring `Blake3BinaryAir`'s.

## Testing

The tests follow `Blake3BinaryAir`'s:

- **Known answers:** the FIPS 180-4 digests of `""` and `"abc"`.
- **Against the `sha2` crate:** random inputs, plus all-zero, all-ones and walking-bit inputs.
- **Shape:** width and constraint count are exactly 23,712 by symbolic evaluation, the degree is 2, and no next row is read.
- **Tampering is rejected:** flipping a bit in any kind of column (the chains, the schedule, the output, `Ch`, `Maj`, `T1` and every kind of carry) breaks a constraint, on that row only. Non-boolean inputs and non-boolean witness bits are rejected too.
- **`Ch` and `Maj` are pinned by their own constraints.** Flipping one of those bits also breaks the addition that uses it, so a test that only checks "some constraint failed" can't tell whether the defining constraint is there. This test checks that the defining constraint itself fails, by index.
- **Every column is forced to a bit** (added after review): walks all 23,712 symbolic constraints in order and checks that each one past the input booleanity checks introduces exactly one column no earlier constraint fixed, and is linear in it. So each cell equals an expression in cells already known to be bits. Runs in 0.3 s, and fails if a constraint is dropped.
- **The example's prover is exercised** (added after review): `backends_prove_the_sha256_air_byte_for_byte` and `dense_and_packed_sha256_tables_have_identical_boolean_proofs`, beside the Keccak and Blake3 ones. The second also checks the new packed generator against the dense one.

I checked that the tests can fail. With a wrong rotation constant in the AIR's `Σ1`, 8 tests fail. With the `Maj` constraints deleted, the constraint-count test and the `Ch`/`Maj` test both fail.

`p3-sha256-air` passes 27/27 and `p3-examples` 28/28 on the M2. Nightly clippy with `-D warnings` and nightly fmt are clean, and so is the intra-doc link check.


sha256-air/src/lib.rs
```diff
@@ -1,4 +1,7 @@
-//! AIR for the SHA-256 compression function.
+//! AIRs for the SHA-256 compression function.
+//!
+//! [`Sha256Air`] works over prime fields. [`Sha256BinaryAir`] works over fields of
+//! characteristic 2; see its module for how it differs from the prime-field AIR below.
```
Nashtare: `binary` is a private module, so rustdoc never renders its doc. In `cargo doc` there's no page for it, and its text appears nowhere. Could this point to [`Sha256BinaryCols`] instead? It already explains the chains. Or the module paragraph could move onto `Sha256BinaryAir`.

sha256-air/src/binary/tests.rs
```diff
@@ -0,0 +1,286 @@
...
+            failures
+                .iter()
+                .any(|failure| failure.constraint == 32 * word + bit),
+            "bit {bit} of input word {word}"
+        );
+    }
+}
+
+#[test]
+fn rejects_non_boolean_witness_at_the_top_bit() {
```
Nashtare, line 388: These two cells are the only direct check of the claim that every non-input column is forced to a bit, and that claim is what makes the AIR sound. The whole-row version is cheap. Convert `get_symbolic_constraints` to polynomials over GF(2), check that the first 768 are `x^2 + x` on the inputs, then walk the rest in order: each should contain exactly one column not already fixed, and only as the monomial `x`. That takes about 0.3 s in release, and it passes on this branch for all 22,944 non-input columns. Would it be worth adding? Then a later edit that leaves a column free, or pushes one into a product, would fail here instead of depending on which cells the flip tests pick.

examples/examples/prove_hash_binary.rs
```diff
@@ -139,6 +140,18 @@ fn main() {
             );
             prove_binary_air_with_ntt_and_backend(&air, trace, options, ntt, backend)
         }
+        BinaryHashOptions::Sha256Compressions => {
+            println!("Proving {trace_height} SHA-256 compressions");
+
+            let air = Sha256BinaryAir {};
```
Nashtare, line 139: Nothing in CI proves this AIR with the binary prover: the crate tests only check constraints, and CI doesn't run the example. Keccak and BLAKE3 each have a `backends_prove_the_*_air_byte_for_byte` test in `examples/src/binary.rs`. A SHA-256 sibling over `generate_random_trace_rows::<F>(4, 0)` passes locally on this branch merged with main. The cells fit GF(4), all three backends give byte-identical proofs, and a full `prove_binary_air` verifies. Could you add it alongside the other two?

Nashtare (approved):
Thanks for this. It's a clean port of the `Blake3BinaryAir` design, and the write-up made it easy to check.

I went through the soundness argument carefully and it holds. Every sum is a chain of two-operand additions mod 2^32, so each carry is a single bit. Only the last step of a chain reads its carries off the stored sum, and by then its partial sum is already fixed. Nothing relies on an integer identity that would collapse in characteristic 2. The rotation and shift amounts, `Ch`, `Maj`, the K constants, the chain offsets and the `h_out` pairing all match FIPS 180-4.

To check the "every other column is forced to a bit" claim for the whole row, I converted the 23,712 symbolic constraints to polynomials over GF(2) and walked them in order. The first 768 are `x^2 + x` on exactly the input columns. Each of the other 22,944 has exactly one column not fixed by earlier constraints, and that column appears only as a linear term. So each boolean input has exactly one satisfying row, and every cell of it is a bit. I then computed that row straight from the constraints, without the generator, for 1,048,576 inputs (zero, all-ones, the IV, walking bits and random ones). Its `h_out` matched the `sha2` crate every time.

The crate's 26 tests pass, and clippy, nightly fmt and rustdoc with `-D warnings` are clean. The branch merges cleanly into main; #2184 only touches the `sha256` hash crate. On the merge, the example proves and verifies from `-l 8` to `-l 11`, including with 64 Rayon threads. Proof sizes match your table byte for byte (784,202 and 851,838 B). On a loaded 14-core M-series machine, SHA-256 at `-l 10` proves in 1.0-1.2 s against 0.51-0.54 s for BLAKE3, in line with the 2.06x width ratio.

The inline comments are non-blocking follow-ups: two test additions and a doc pointer.

gopikannappan:
Thanks @Nashtare — reconstructing the row from the constraints alone for a million inputs is a lovely check, and the walk over the 22,944 non-input constraints is exactly the argument the AIR rests on.

Pushed in a68fab7c:

- **Merged main and moved to the Boolean commitment.** #2227/#2228/#2245 landed while this was open, so the example's other objectives now generate packed traces. Added `generate_binary_trace_packed` and `generate_random_trace_packed` mirroring `Blake3BinaryAir`, and the example builds its table with `Table::from_packed_bits`. Re-measured on merged main: 2¹⁰ SHA-256 proves in 0.152–0.164 s against BLAKE3's 0.067–0.069 s, still about 2.2–2.4×. Description updated.
- **The whole-row booleanity check is now a test.** `every_column_is_forced_to_a_bit_by_the_constraints` walks the symbolic constraints in order and asserts each one past the input checks introduces exactly one unfixed column and is linear in it. 0.3 s, and I verified it fails if a constraint is dropped: removing the `new e = d + T1` addition leaves that column free and the test catches it.
- **CI now proves this AIR.** Added `backends_prove_the_sha256_air_byte_for_byte` and `dense_and_packed_sha256_tables_have_identical_boolean_proofs` beside the Keccak and Blake3 ones. The second doubles as a check of the new packed generator against the dense one.
- **Doc pointer fixed**: the crate doc now points at `Sha256BinaryCols`, since `binary` is private and rustdoc never rendered it.

27/27 in `p3-sha256-air` and 28/28 in `p3-examples`, with nightly clippy `-D warnings` and fmt clean.

PR #2208, opened 2026-09-18
feat(bus): add direction-aware product GKR core

tcoratger:
## Summary

Add the standalone core of a characteristic-two-safe multiset bus:

- structural `Push` and `Pull` directions, never signed field multiplicities;
- fixed-width tuple fingerprints with Boolean row selection;
- batched radix-4 product-tree GKR over arbitrary extension fields;
- implicit multiplicative-identity suffix padding;
- verifier-derived proof shape and typed Fiat–Shamir transcript;
- a shared terminal point for later PCS authentication.

The existing LogUp path and its characteristic-two guard are unchanged.

## E6 scope

- **6.1 — fingerprint and product check:** core implemented. Callers reserve a tuple slot for their domain separator and sample fingerprint challenges after commitments.
- **6.2 — radix-4 batched product GKR:** implemented for multiple trees with one fresh combiner per layer.
- **6.3 — counts in the exponent:** deferred. It needs a separately audited full-order generator, read bound, and nonzero-count product.
- **6.4 — leaf stacking and point recycling:** identity-padded prefixes and the shared terminal point are implemented; multi-STARK leaf authentication is deliberately deferred.
- **6.5 — table sumcheck:** already provided structurally by the mixed-height global zerocheck; integration comes later.
- **6.6 — ergonomics/debugger:** typed declarations are present. The host-integer imbalance debugger belongs with integration.
- **6.7 — dead cells:** suffix-one padding is implicit at every product level. General sparse live masks remain a measured follow-up.

## Security boundary

The product proof checks its internal reductions only. Its returned leaf evaluations are unauthenticated until a surrounding commitment opening or table sumcheck binds them.

For a width-`2^s` tuple and `N` factors, the fingerprint collision term is at most `max(1, s) N / |F|`. Each radix-4 sumcheck round has degree at most five; tree batching contributes at most `(T - 1) / |F|` per layer, and collapsing four children contributes at most `2 / |F|` per layer. These terms will enter `p3-security` with the integration PR.

All proof-controlled lengths are checked before transcript replay. Roots, round polynomials, and child claims are absorbed before their corresponding challenges.

## Benchmark

Criterion, three unequal `BinaryField128` trees, complete product-tree and sumcheck arithmetic:

| height | radix-2 | radix-4 | change |
|---:|---:|---:|---:|
| 14 | 71.510 ms | 65.169 ms | -8.9% |
| 18 | 1.1496 s | 1.0473 s | -8.9% |

Production uses radix-4, with one binary cap for odd heights. Radix-2 remains only as a benchmark comparator.

## Validation

- `cargo test -p p3-bus --lib` (10 passed)
- `cargo clippy -p p3-bus --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p p3-bus --no-deps`
- `git diff --check`


tcoratger:
Superseded by #2216. The replacement uses the same code on an upstream branch and is the bottom layer of native E6 stack #2219.

PR #2209, opened 2026-09-18
feat(word): add checked word relation IR

tcoratger:
## Summary

Add `p3-word`, a backend-neutral `no_std` relation language for dense binary work:

- sealed 32- and 64-bit word types;
- checked public and witness indices;
- canonical full-width and 32-bit-lane shifts;
- zero-, one-, and irreducible two-shift terms;
- XOR operands;
- zero, bitwise-AND, and full-width unsigned multiplication relations;
- immutable checked systems and an authoritative scalar verifier.

The crate depends only on `alloc` and `thiserror`. It has no AIR, PCS, transcript, multi-STARK, frontend, gadget, chip, or VM dependency.

## E7 scope

- **7.1 — words and shifted indices:** implemented for `Word32` and `Word64`. The four 32-bit-lane operations are intentionally available only in a 64-bit word.
- **7.2 — AND and integer multiplication:** relation IR and scalar semantics implemented. Binary/GHASH multiplication remains a later, consumer-driven addition.
- **7.3–7.4 — proof reductions and ordering:** deferred to the proof-backend PR. The neutral IR does not import protocol code.
- **7.5–7.6 — frontend and gadgets:** remain downstream by design.
- **7.7 — chip composition:** deferred until the E6 bus can bind caller/callee rows. This PR does not reproduce binius64 M4’s currently unbound composition.

## Invariants

- Invalid word/lane combinations and shift distances are rejected before evaluation.
- Zero-distance shifts have one representation.
- Two-slot terms reject maps representable by zero or one shift.
- All addresses are validated before a system is constructed.
- Scalar verification requires exact public and witness lengths.
- Integer multiplication checks both limbs of the complete `2 × word_bits` product.

Every ordered pair of valid shifts at both widths is compared with independently generated bit-linear map signatures. That exhaustive test found and prevented a mixed lane/full-width canonicalization error during development.

## Validation

- `cargo test -p p3-word` (17 passed)
- `cargo clippy -p p3-word --all-targets -- -D warnings`
- `cargo doc -p p3-word --no-deps`
- `cargo fmt -p p3-word`
- `git diff --check`


tcoratger:
Superseded by #2217. The replacement uses the same commit on an upstream branch and is the bottom layer of native E7 stack #2220.

PR #2210, opened 2026-09-18
feat(word): add packed backend layout

tcoratger:
## Summary

Stacked on #2217.

Add the protocol-free packed layout layer for the word relation:

- exact-shape scalar or ownership-taking packed witness ingestion;
- direct 32-/64-bit mapping into Boolean-PCS bit order;
- packed AND-input and integer-multiplication operation columns;
- deterministic shift-key compilation per public or witness word.

There is no transcript, proof claim, PCS padding, frontend, or chip composition in this slice.

## Design

The packed constructor takes existing vectors without repacking or reallocating. Scalar construction remains a convenience path.

Compiled keys are resolved by the segment that owns them. A key cannot be passed into another segment or layout. Each resolved key exposes its semantic operand role, dense shift index, ordered shift pair, and occurrence-preserving references.

Shift IDs are 32-bit. The complete valid shift-sequence space therefore fits without the former 65,536-sequence rejection. Terms remain grouped by their checked spelling because irreducible pairs are not semantically unique.

The reduction order is families, operand roles, constraint positions, then supplied terms. Repeated terms remain repeated so characteristic-two cancellation is preserved.

## Validation

- `cargo test -p p3-word` (21 passed)
- `cargo test -p p3-word-backend` (16 passed)
- `cargo clippy -p p3-word-backend --all-targets -- -D warnings`
- `cargo doc -p p3-word-backend --no-deps`
- `git diff --check`

Coverage includes scalar-versus-packed properties, exact-shape and zero-copy ownership, stable ordering, repeated references, cross-segment isolation, lane distinction, two-shift round trips, and 65,537 distinct shift sequences.


word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+    /// The operand position within its relation.
+    operand: u8,
+    /// The position within its homogeneous relation family.
+    constraint: u32,
+}
+
+impl ConstraintReference {
+    /// Returns the operand position within the operation.
+    #[inline]
+    pub const fn operand(self) -> u8 {
```
Nashtare: This `u8` means different things per family (Zero 0; AND 0/1/2 = left/right/output; MUL 0/1/2/3 = left/right/low/high), so could the doc give that table, or return `OperandRole` instead?

word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+}
+
+/// One word's references under a fixed operation and shift sequence.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CompiledKey {
+    /// The relation family consuming the word.
+    operation: ConstraintKind,
+    /// The compact index of the canonical shift sequence.
+    shift: u16,
```
Nashtare: Interning sequences into a dense table lets a reduction evaluate each shift operator once per distinct sequence and index it by id, as binius64 does with `Key::id`. But this id and the `shifts` table (line 58) are private, and the only route is `shifts(&key) -> Option<[Shift<W>; 2]>`. If the reduction lands in another crate, it would have to re-hash `[Shift; 2]` per key to share that work. Could you expose something like `CompiledKey::shift_index()` and `CompiledSegment::shift_sequences() -> &[[Shift<W>; 2]]`?

word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+
+    /// Returns the keys attached to a word.
+    pub fn keys(&self, word: usize) -> Option<&[CompiledKey]> {
+        let range = self.word_keys.get(word)?;
+        self.keys.get(range.start as usize..range.end as usize)
+    }
+
+    /// Returns a key's canonical inner and outer shifts.
+    pub fn shifts(&self, key: &CompiledKey) -> Option<[Shift<W>; 2]> {
```
Nashtare: `shifts` and `references` accept any `CompiledKey`, including one from the other segment or from another layout, and return whatever sits at that id and range. In a probe with public word 0 under `rotr 3` and witness word 0 under `sll 5`, `layout.witness().shifts(public_key)` returns `Some([sll 5, identity])` instead of `rotr 3`. It only returns `None` when the index happens to be out of range. Could a per-word iterator yield resolved `(ConstraintKind, [Shift<W>; 2], &[ConstraintReference])` so raw keys never leave their segment? Tagging keys with their segment and a `debug_assert` would also do.

word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CompiledKeyLayout<W: Word> {
+    /// The metadata for verifier-known words.
+    public: CompiledSegment<W>,
+    /// The metadata for prover-committed words.
+    witness: CompiledSegment<W>,
+}
+
+impl<W: Word> CompiledKeyLayout<W> {
+    /// Compiles every shifted reference in deterministic reduction order.
```
Nashtare, line 155: This says "deterministic reduction order", but the order a reduction author needs is only spelled out in the PR description. Keys within a word come in first-appearance order of the walk Zero → AND (operand-major) → MUL (operand-major), and references within a key follow operand → constraint → term. Duplicate terms in one operand give duplicate references (as in the test at line 576), which cancel in characteristic 2, so the reduction has to sum them with multiplicity rather than dedupe them. The probe confirms all three. Could the doc here state them?

word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+    constraint: u32,
+    value: &Operand<W>,
+    visit: &mut impl FnMut(Reference) -> Result<(), KeyCompileError>,
+) -> Result<(), KeyCompileError> {
+    for term in value.terms() {
+        let index = term.index();
+        visit(Reference {
+            segment: index.segment(),
+            word: index.position() as usize,
+            key_code: key_code(operation, [term.inner(), term.outer()]),
```
Nashtare: This keys on the pair's spelling, and #2217 accepts several spellings of one map (e.g. `(srl 11, sll 18)`, `(sra 11, sll 18)` and `(rotr 11, sll 18)`). A frontend that writes one mask two ways therefore gets two keys and duplicate reduction work. That's correct, just not minimal; see my comment on `ShiftedValue::pair` in #2217.

word-backend/src/keys.rs
```diff
@@ -0,0 +1,703 @@
...
+) -> Result<CompiledSegment<W>, KeyCompileError> {
+    key_codes.sort_unstable();
+    key_codes.dedup();
+    let mut sequence_codes = key_codes
+        .iter()
+        .map(|code| code & SEQUENCE_MASK)
+        .collect::<Vec<_>>();
+    sequence_codes.sort_unstable();
+    sequence_codes.dedup();
+    if sequence_codes.len() > u16::MAX as usize + 1 {
```
Nashtare: The `u16` shift index caps each segment at 65,536 distinct `[inner, outer]` sequences, but #2217 accepts 108,948 on `Word64` (377 singles plus 108,571 irreducible pairs). So a small valid system hits it: one Zero constraint with 65,537 terms on witness word 0, each a different sequence, passes `ConstraintSystem::new` and then fails here with `LayoutTooLarge { Witness, Shifts, 65537 }`, while 65,536 terms compile and round-trip. Canonicalizing pairs in #2217 wouldn't be enough, since 74,718 distinct maps remain. The error is clean and real circuits use far fewer shifts, but the path is untested. Could the id be a `u32` with the 2-bit operation packed into its top bits, which keeps `CompiledKey` at 12 bytes, or else could `CompiledKeyLayout::new` document the limit and gain a 65,536/65,537 boundary test?

word-backend/src/witness.rs
```diff
@@ -0,0 +1,238 @@
...
+        PackedGf2x64::new(self.get())
+    }
+
+    #[inline]
+    fn unpack(value: Self::Packing) -> Self {
+        Self::new(value.to_bits())
+    }
+}
+
+/// A witness whose words retain the bit order consumed by the Boolean PCS.
```
Nashtare, line 50: For the reduction port: under the big-endian `Point` convention (`Point::hypercube`), bit `i` of word `w` sits at hypercube index `64w + i`. The bit-within-word index therefore occupies the last 6 coordinates (5 for `Word32`) with the LSB last, and the word index the leading ones, which is the reverse of binius64's little-endian order. Would one sentence here or in the crate docs be worth it, so the next PR doesn't trip on it?

Nashtare (approved):
I checked the layout against an independent model, not only by reading. Locally I ran a probe on 3,000 random systems per width (`Word64` and `Word32`), with 0–3 Zero/AND/MUL constraints, 0–4 terms per operand, public and witness terms, and every canonical single and pair shift, biased so that keys collide. For every word of both segments, `CompiledKeyLayout` matched a naive model that walks the terms in reduction order and groups them by (operation, `[inner, outer]`) in first-appearance order: key order, operation, shift pair and the (operand, constraint) reference lists all agreed. On the same systems, `OperationColumns` matched `Operand::evaluate` bit for bit in both AND columns and all four MUL columns. The model is short, so it could become a proptest here if you want it in the suite.

The bit order matches the Boolean PCS `commit_bits` contract: `pack` is `PackedGf2x64::new(word.get())`, so bit `i` of word `w` is committed bit `64w + i` (`32w + i` for `Word32`). The mapping works on values, not bytes, so it doesn't depend on the target's endianness. There is no `unsafe` in the crate, and `from_packed` is a plain move of the two `Vec`s after the length check, so the zero-copy constructor really is zero-copy.

The inline comments are non-blocking follow-ups. The one I'd most like addressed is the `u16` shift-sequence cap, which a small valid `Word64` system can hit.

**Smaller things**
- The description says "Stacked on #2209", but #2209 is closed and the base is #2217.
- As with `p3-word` in #2217, there is no `release-plz.toml` entry for `p3-word-backend`, and the `binary` glob in `.github/labeler.yml` doesn't cover `word-backend/**`.

word-backend/src/keys.rs
tcoratger: Exposed the dense shift index and the segment shift-sequence table for reduction reuse.

word-backend/src/keys.rs
tcoratger: Documented that compilation groups checked spellings rather than quotienting semantic linear maps, matching the parent IR contract.

word-backend/src/keys.rs
tcoratger: Documented exact family, role, constraint, and term order, including duplicate multiplicity in characteristic two.

word-backend/src/keys.rs
tcoratger: Eliminated cross-segment key misuse: callers receive lifetime-bound resolved keys from a segment-specific per-word iterator.

word-backend/src/keys.rs
tcoratger: Removed the u16 cap by using u32 dense shift IDs and added the 65,537-sequence boundary test.

word-backend/src/witness.rs
tcoratger: Documented the word_bits*w+i bit layout and its trailing big-endian multilinear coordinates.

word-backend/src/keys.rs
tcoratger: Replaced the raw operand byte with a typed OperandRole.

PR #2211, opened 2026-09-18
feat(bus): preserve direction in AIR declarations

tcoratger:
## Summary

Stacked on #2216.

Add an opt-in AIR declaration surface for the binary-native bus. It records the bus name, symbolic tuple expressions, row activation, and push or pull direction as structural metadata.

The existing lookup API and prime-field LogUp path are unchanged.

## Soundness

A conditional activation automatically emits s * (s - 1) = 0. The public checked method is supplied by a blanket extension trait and cannot be overridden. Backend recorders require a crate-issued token, so a folder cannot record a conditional declaration while dropping its Booleanity constraint.

Filtered declarations intentionally do not compile. Filtering would constrain activation only on selected rows while recording it globally.

The symbolic builder retains the global emission order of base-field and extension-field constraints. The debug builder executes the same automatic activation check row by row.

## Validation

- cargo test -p p3-bus --lib: 19 passed
- cargo test -p p3-bus --doc: 1 compile-fail test passed
- git diff --check


bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        self.record_bus_interaction_unchecked(bus_name, direction, fields, activation);
+    }
+
+    /// Record a declaration whose activation constraints are already enforced.
+    ///
+    /// This is the implementation hook beneath the checked public declaration path.
+    #[doc(hidden)]
+    fn record_bus_interaction_unchecked<E: Into<Self::Expr>>(
```
Nashtare: A `Boolean(s)` leaf is `1 + s·(f − 1)`, so if `s` is unconstrained on one row the prover can set it to `(target − 1)/(f − 1)` and balance any push/pull product. The `s·(s − 1)` emitted above is the only thing enforcing Booleanity (the check in `materialize_bus_leaves` runs on the prover's side), and two paths lose it:
- `#[doc(hidden)]` only hides this method from the docs. Any `Air::eval` bounded on `AB: BusInteractionBuilder` can call it with `BusActivation::Boolean(s)` and record the activation with no constraint.
- `push_bus_interaction` is a provided method, so an implementer can override it. `MultilinearFolder` (`multi-stark/src/folder.rs:398-418`) and the batch-stark folders (`batch-stark/src/folder.rs:200-237`) already override every `InteractionBuilder` method with empty or drain-only bodies; a bus folder copying that pattern would drop the constraint while the symbolic profile still counts it.

Could this be sealed, e.g. by moving the checked method to an extension trait with a blanket impl (so it can't be overridden) and giving this hook an argument of a type only `p3-bus` can construct, such as `pub struct RecordToken(())`? Alternatively, the future plan/folder could emit `s·(s − 1)` from the recorded `activation`, with `push_bus_interaction` no longer emitting it, so no builder choice can lose it.

bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+    Always,
+    /// One expression selects whether the row contributes its tuple.
+    Boolean(E),
+}
+
+/// Opt-in AIR interface for binary-native bus declarations.
+///
+/// Direction is metadata rather than the sign of a field expression.
+/// It therefore remains meaningful in characteristic two.
+pub trait BusInteractionBuilder: AirBuilder {
```
Nashtare: A few things worth a line in these trait docs:
- `builder.when(c).push_bus_interaction(..)` not compiling is the right behaviour, since a forwarding impl for `FilteredAirBuilder` would constrain only `c·s·(s − 1)` while recording the unfiltered `s`; a `compile_fail` doctest would keep someone from adding that impl later.
- The automatic check adds a degree-`2·deg(s)` constraint to the AIR's maximum, and the builder's row selectors (`is_first_row`, `is_transition`) aren't 0/1 in the univariate backends, so they aren't valid activations.
- If the hook stays, it should be mentioned here, since rustdoc currently shows the trait with no required items.

bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    /// Evaluate one AIR symbolically and retain its bus declarations.
+    ///
+    /// # Panics
+    ///
+    /// Panics when the supplied layout disagrees with the AIR's declared widths.
+    #[must_use]
+    pub fn from_air<A>(air: &A, layout: AirLayout) -> Self
+    where
+        A: BaseAir<F> + Air<Self>,
```
Nashtare: `Air<Self>` already implies `BaseAir<F>`, so this can be `A: Air<Self>`, as in `InteractionSymbolicBuilder::from_air`.

bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    /// Symbolic base-field constraints emitted in the same AIR evaluation.
+    #[must_use]
+    pub fn base_constraints(&self) -> Vec<SymbolicExpression<F>> {
+        self.inner.base_constraints()
+    }
+
+    /// Symbolic extension-field constraints emitted in the same AIR evaluation.
+    #[must_use]
+    pub fn extension_constraints(&self) -> Vec<SymbolicExpressionExt<F, EF>> {
```
Nashtare, line 162: A bus AIR is written as `impl<AB: BusInteractionBuilder> Air<AB>`, so it doesn't satisfy `Air<SymbolicAirBuilder>` or `Air<DebugConstraintBuilder>`. That rules out `get_max_constraint_degree*`, `get_constraint_layout` and `check_constraints` for it, which leaves this builder as the only way in. Without a `constraint_layout()` here (the lookup builder forwards one), a later prover can't rebuild the interleaved base/extension order that `ConstraintLayout::decompose_alpha` needs from these two separate vectors. Could this forward `self.inner.constraint_layout()`, and could `p3-bus` implement `BusInteractionBuilder` for `DebugConstraintBuilder` with a hook that only drains `fields`? The default method would then check `s·(s − 1)` row by row under `check_constraints`.

bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+    }
+
+    impl<AB> Air<AB> for DirectionAir
+    where
+        AB: BusInteractionBuilder<F = BinaryField128>,
+    {
+        fn eval(&self, builder: &mut AB) {
+            // Read one arbitrary AIR expression rather than a materialized column index.
+            let value: AB::Expr = builder.main().current_slice()[0].into();
```
Nashtare, line 318: The fixture's only field is the bare column `main[0]`, so no test records a compound expression (`a·b + c`), a next-row reference, a public value or a periodic column, which is what the description's "arbitrary symbolic expressions" covers. Also, `identical_pushes_do_not_cancel_in_characteristic_two` runs with `conditional = true`, so its second push has a `Boolean` activation and the two pushes aren't identical. Could the fixture add a compound field and an AIR-owned `assert_zero` / `assert_zero_ext`, so the tests also show ordinary, extension and automatic constraints kept in emission order?

bus/src/builder.rs
```diff
@@ -0,0 +1,346 @@
...
+                if variable.entry == BaseEntry::Main { offset: 0 }
+                    && variable.index == 0
+        ));
+    }
+
+    #[test]
+    fn boolean_activation_adds_its_own_constraint() {
+        // One conditional declaration contributes one Booleanity constraint.
+        let conditional = profile(BusDirection::Pull, true);
+        assert_eq!(conditional.base_constraints().len(), 1);
```
Nashtare, line 400: This only counts constraints, so it would still pass if the builder emitted `assert_zero(selector)` (forcing `s = 0`) or `assert_zero(selector − 1)`. And because the fixture uses the same column as tuple field and selector, a Booleanity check on the wrong expression would pass too. Could the fixture take the selector from a second column, and the test evaluate the constraint at `s ∈ {0, 1}` (zero) and at a random `s` (non-zero)?

Nashtare (commented):
I read the builder against `SymbolicAirBuilder` and `p3_lookup::InteractionSymbolicBuilder`. Direction stays structural: `BusDirection` goes into `SymbolicBusInteraction::direction` and is never turned into a field element anywhere in `p3-bus`. The wrapper forwards every `AirBuilder`, `ExtensionBuilder` and `PermutationAirBuilder` method to the inner builder, so an AIR's own constraints and its bus declarations come from one evaluation over one layout, with nothing dropped or counted twice. `from_air` validates the layout before `eval`, as the lookup builder does. `bool_check` is `s·(s − 1)`, whose only roots are 0 and 1 in any field, characteristic 2 included, and it is emitted unconditionally, so it holds on every row. Tuple fields are base-field `SymbolicExpression`s and can't reference permutation columns or challenges, which is the right restriction for tuples that must be fixed before the fingerprint challenges. `FilteredAirBuilder` doesn't implement the trait, so a declaration can't be made under `when(..)` and silently lose its filter. `p3-lookup` is untouched.

Nothing consumes the declarations yet, so there is no exposure today. I'd still like the API shape settled before later layers build on it.

- The Booleanity constraint that keeps a `Boolean(s)` leaf sound is enforced only by convention. `record_bus_interaction_unchecked` is a public required method that any AIR can call, and `push_bus_interaction` can be overridden, as the multi-stark and batch-stark folders already do for `InteractionBuilder` (inline, `builder.rs`).

**Smaller things**
- The description says "Stacked on #2208", but #2208 is closed and the base is now #2216.
- The description's "reserved for trusted builder implementations" and "arbitrary symbolic expressions" go further than the code and tests do (inline).

bus/src/builder.rs
tcoratger: Separated selector and tuple columns and evaluated the Booleanity polynomial at 0, 1, and a non-Boolean value.

bus/src/builder.rs
tcoratger: Removed the redundant bound.

bus/src/builder.rs
tcoratger: Forwarded constraint layout and added a debug builder recorder that preserves row-by-row Boolean checks.

bus/src/builder.rs
tcoratger: Added the filtered-builder compile-fail contract and documented activation degree and selector restrictions.

bus/src/builder.rs
tcoratger: Expanded the fixture to compound current/next/public/periodic expressions and pinned ordinary, extension, and automatic constraint order.

bus/src/builder.rs
tcoratger: Sealed the checked declaration path behind a blanket extension implementation and an unforgeable recording token, so builders cannot bypass Booleanity constraints.

bus/src/builder.rs
```diff
@@ -0,0 +1,516 @@
...
+        self.interactions.push(SymbolicBusInteraction {
+            bus_name: bus_name.to_string(),
+            direction,
+            fields: fields.into_iter().map(Into::into).collect(),
+            activation,
+        });
+    }
+}
+
+impl<F: Field, EF: ExtensionField<F>> BusInteractionRecorder for DebugConstraintBuilder<'_, F, EF> {
```
Nashtare, line 268: Nothing in the tree runs this impl yet. Locally I tried a two-column AIR with `push_bus_interaction(.., Boolean(s))` under `check_constraints`:
- BabyBear: it accepts `s = [0, 1, 1, 0]` and panics on row 2 for `[0, 1, 2, 0]`.
- GF(2^128): it panics on row 3 when `s` is the generator.

Could a small test like that land next to this, so the one row-level check of the automatic constraint is pinned? `BinaryField128::TWO` is zero, so the non-Boolean value needs to be something like `GENERATOR` or the random draw the test above already uses.

bus/src/builder.rs
```diff
@@ -0,0 +1,516 @@
...
+        }
+    }
+
+    #[test]
+    fn compound_fields_preserve_global_constraint_order() {
+        // The AIR emits base, extension, then automatic Booleanity constraints.
+        let air = RichAir;
+        let profile = BusSymbolicBuilder::from_air(&air, AirLayout::from_air(&air));
+        assert_eq!(profile.interactions().len(), 1);
+        assert!(profile.interactions()[0].fields[0].degree_multiple() >= 2);
```
Nashtare: Any product of two columns passes this check, so it doesn't pin the next-row, public or periodic terms. Locally I changed the tuple to `value * selector.clone()` alone, and the test still passed. Could it check the leaves instead? For example, it could collect the `BaseEntry`s of `fields[0]` and assert that `Main { offset: 1 }`, `Public` and `Periodic` are all present, or extend `evaluate` to those entries and compare at a point.

Nashtare (approved):
I re-checked the sealed path from a probe crate outside `p3-bus`. Overriding `push_bus_interaction` fails with E0119 against the blanket impl. Building a `RecordToken` fails every way I tried: struct literal, `Default`, `.clone()`, `From<()>` and destructuring. Implementing the recorder for `SymbolicAirBuilder` or `FilteredAirBuilder` hits the orphan rule. An AIR can no longer reach the hook by accident.

The debug recorder does check `s·(s − 1)` row by row. For a two-column AIR with `Boolean(s)`, `check_constraints` gives:
- BabyBear: it accepts selectors `[0, 1, 1, 0]` and panics on row 2 for `[0, 1, 2, 0]`.
- GF(2^128): it accepts `[1, 0, 1, 0]` and panics on row 3 when the selector is the generator.

The forwarded `constraint_layout()` is the inner builder's global emission order, which is what `decompose_alpha` indexes. The new test pins `[0, 2]` / `[1]` for AIR base, AIR extension, then the automatic constraint. The `compile_fail` doctest fails with E0599 on `FilteredAirBuilder: BusInteractionRecorder`, and the same snippet without `when(..)` compiles, so it fails for the reason it claims. `cargo test -p p3-bus` (20 lib tests + the doctest) and clippy with `-D warnings` pass. The branch merges cleanly into main, and `p3-bus` already has its release-plz entry on main.

The sealed extension trait, the forwarded layout, the debug recorder, the separate selector column and the evaluated Booleanity test close the earlier threads. I've replied in the seal thread about one gap in the token idea I suggested: a token can be obtained through a throwaway recorder, or passed on by a filtering wrapper. Neither happens by accident, so I'm happy for it to be a follow-up. The inline comments are non-blocking test follow-ups.

**Smaller things**
- The description still says "Stacked on #2216", which has merged, and "19 passed" (I get 20).
- "A folder cannot record a conditional declaration while dropping its Booleanity constraint" goes a bit further than the code does (see the reply).

bus/src/builder.rs
Nashtare: The two paths from this thread are closed. From a crate outside `p3-bus`, overriding `push_bus_interaction` is E0119. A `RecordToken` can't be built, defaulted, cloned or destructured. Recorder impls for `SymbolicAirBuilder` / `FilteredAirBuilder` are orphan errors.

The token idea I suggested has one gap: a token proves that *some* builder ran the checked path, not this one. I found two ways to exploit that.
- **A throwaway recorder.** A small recorder outside `p3-bus` that keeps the tokens it receives can mint one with `push_bus_interaction("", Push, [], Always)`. An AIR can then spend it with `builder.record_bus_interaction(token, .., Boolean(s))`. On `BusSymbolicBuilder` that recorded a `Boolean` interaction with 0 base constraints, and `check_constraints` accepted `s = 2`.
- **A forwarding wrapper.** An external `when`-style wrapper whose recorder hands the token to `inner` produces the constraint `is_first_row · s·(s − 1)` next to an unfiltered `Boolean(s)`. The debug check accepted `s = 2` on row 1, which is the case the doctest is meant to rule out.

Neither can happen by accident, so this isn't blocking. The cheap follow-up is to state the implementer contract on `BusInteractionRecorder`: `assert_zero` must reach the same constraint set unconditionally, and the token must not be forwarded past a filter. Could that trait also drop `#[doc(hidden)]`? Every backend has to implement it, but rustdoc currently shows `BusInteractionBuilder: BusInteractionRecorder` with nothing behind the link.

If you want the token to be literal, you could brand it by recorder type: `pub struct RecordToken<B>(PhantomData<fn(B) -> B>)`, with `token: RecordToken<Self>` in the hook. Locally that's a three-line change. All 20 lib tests still pass, and both cases above stop compiling (E0308). The cost is that a forwarding wrapper then has to go through `inner.push_bus_interaction`, so the inner builder emits its own unfiltered check.

PR #2212, opened 2026-09-18
feat(bus): derive mixed-height bus layouts

tcoratger:
## Summary

Stacked on #2211.

Add the verifier-derived layout between symbolic bus declarations and product GKR.

The plan:

- groups named buses in stable lexicographic order;
- assigns nonzero domain identities;
- exposes a typed slot map for payload, identity, and zero-padding positions;
- places mixed-height blocks on aligned Boolean subcubes;
- fixes the exact materialization order;
- stores the product shape checked during construction;
- derives private security geometry from the same plan.

The terminal relation is:

    L(q) - 1 = sum_b eq(prefix_b, q_prefix) * (L_b(q_rows) - 1)

A real plan to materialization to product-GKR test pins the block order and the most-significant-variable-first terminal point.

This PR does not connect the planner to PCS or multi-STARK. That integration must consume the verifier-derived plan.

## Validation

- cargo test -p p3-bus --lib: 25 passed
- cargo test -p p3-bus --doc: 1 passed
- git diff --check


bus/src/plan.rs
```diff
@@ -0,0 +1,876 @@
...
+    pub fn blocks(&self, direction: BusDirection) -> &[BusBlock] {
+        match direction {
+            BusDirection::Push => &self.pushes,
+            BusDirection::Pull => &self.pulls,
+        }
+    }
+
+    /// Terminal shares in the product tree's physical block order.
+    ///
+    /// The leading point coordinates select one aligned block.
```
Nashtare, line 411: This reads the point big-endian: the leading `prefix_variables` coordinates select the block and the trailing ones address its rows (the field docs at lines 85–88 say the same). `prove_product_gkr`/`verify_product_gkr` return `ProductGkrOutput::point` low-variable-first, though: `product.rs:360` builds `[branches[0], branches[1]] ++ round_point`, and `branches[0]` binds the lowest bit of the leaf index, so `point[k]` is bit `k`. Locally I laid out a plan with three AIRs (heights 8, 2 and 1, two buses, three blocks per side, `n = 4`) as leaves, ran the real prover and verifier, and checked this relation on the output point: the split documented here fails for both directions, and it holds on the reversed point. Nothing is wired yet and a fixed permutation of the leaves doesn't change the product, so this is completeness, not soundness: an integration that follows these docs will reject honest proofs. The convention here matches the rest of the repo (`Point::hypercube`, `Selector::lift_prefix`, the logUp* `Block::weight`), so I've left a comment on #2216 asking for the GKR output point to be reversed there; if you'd rather keep the GKR order, could these docs say instead that `q` is `ProductGkrOutput::point` reversed?

Nashtare (commented):
I read the layout end to end and ran a couple of local probes against this branch. The placement is right: blocks are sorted by descending height, so every earlier offset is a sum of powers of two no smaller than `2^h` and each block lands on its own aligned subcube. A run over 2000 random statements (1–6 AIRs, log heights 0–8, five buses of different widths) found every block aligned, disjoint and in range, with per-side totals equal to `non_padding_leaf_counts`. The terminal relation holds as algebra: padding leaves are 1 and `Σ_x eq(x, q) = 1`, so `L(q) − 1` splits block by block. Domain separation is injective across arities, and bit-encoding the identity into 0/1 slots is the right call in characteristic 2, where `from_usize(identity)` would collapse. The overflow paths all return typed errors, and the geometry in the layout test, `(3, [20, 20], 32, 5, 2, 3)`, matches my own count.

- `terminal_shares` reads the point big-endian while `prove_product_gkr`/`verify_product_gkr` return it low-variable-first, so the documented split fails on the real GKR output; I've asked on #2216 for the point to be reversed there, and either way I'd like a test that runs the plan through the real product GKR (inline, `plan.rs`).

**Smaller things**
- `bus/README.md` still only describes materialization and the product GKR; one line on layout planning would help.

The other inline comments are about the order the leaves must be materialized in and the tuple slot map. Neither breaks anything in this PR alone, but both are contracts the integration will have to get exactly right.

bus/src/plan.rs
```diff
@@ -0,0 +1,876 @@
...
+            F::ONE,
+            F::ONE,
+        ];
+        let blocks = [&dense[..4], &dense[4..6]];
+
+        let reconstructed = plan
+            .terminal_shares(BusDirection::Push)
+            .zip(blocks)
+            .map(|(share, block)| {
+                let prefix = &point[..share.prefix_variables];
```
Nashtare, line 866: `evaluate` reads coordinate 0 as the most significant bit, and this split takes the prefix from the front, so the test checks the plan against a reference that makes the same assumption and can't catch the mismatch above. Could this test (or a new one) lay out the leaves by `blocks()`, run `prove_product_gkr`/`verify_product_gkr` with `product_shape()`, and check the relation on `output.point`? Building the leaves with `materialize_bus_leaves` in that test would also pin the order they have to come in (see the comment on `blocks`).

bus/src/plan.rs
```diff
@@ -0,0 +1,876 @@
...
+        self.logical_tuple_width
+    }
+
+    /// Power-of-two tuple width consumed by the fingerprint MLE.
+    #[must_use]
+    pub const fn fingerprint_width(&self) -> usize {
+        self.fingerprint_width
+    }
+
+    /// Physical blocks for one side of the multiset equality.
```
Nashtare, line 377: These offsets follow the sort at line 219 (direction, descending height, bus, AIR, declaration), but `materialize_bus_leaves` appends each direction's rows in the order the declarations are passed. Take two AIRs of heights 2 and 8 with one push each, passed in that order: the leaves put the 2-row block at 0 and the 8-row block at 2..10, while the plan puts them at 8 and 0, so the 8-row block isn't even aligned. Nothing ties the two APIs together yet, so as far as I can tell an integration that materializes in AIR order gets honest proofs rejected. Could `blocks` document that leaves must be materialized in exactly this order, or better, could the plan emit the `BusLeafDeclaration`s in `blocks(direction)` order itself?

bus/src/plan.rs
```diff
@@ -0,0 +1,876 @@
...
+    /// Root-to-leaf product-reduction layers.
+    pub layer_count: usize,
+}
+
+/// A verifier-derived layout for every named binary-native bus.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BusPlan {
+    /// Named groups in lexicographic name order.
+    domains: Vec<BusDomain>,
+    /// Payload positions reserved before the domain identity bits.
```
Nashtare, line 162: The field docs fix payload first, then the identity bits, then zero padding up to `fingerprint_width`, but two things are left implicit: what fills slots `payload_width..payload_slots` for a bus narrower than the widest one, and which domain slot carries which bit of `identity_bit`. Any fixed choice keeps tuples from different buses distinct, since the identity slots always differ, but prover and verifier have to make the same choice, and today each consumer has to re-derive it. Could `BusPlan` document the full slot map (for example zero filler, and identity bit `i` at slot `payload_slots + i`), or expose a small helper that says, for a given bus and slot, whether it holds a payload index, zero or one?

bus/src/plan.rs
```diff
@@ -0,0 +1,876 @@
...
+
+    /// Checked product-reduction shape fixed by this layout.
+    #[must_use]
+    pub fn product_shape(&self) -> ProductGkrShape {
+        ProductGkrShape::new(
+            self.geometry.log_logical_leaf_count,
+            self.geometry.tree_count,
+            ProductGkrRootShape::FirstTwoShared,
+        )
+        .expect("a checked bus plan retains a valid product shape")
```
Nashtare: `ProductGkrShape` is `Copy`, so storing the shape `build` already checked at line 269 would drop this `expect` and the repeated `FirstTwoShared`.

tcoratger: Stored the checked product shape in the plan instead of rebuilding it with an expect.

bus/src/plan.rs
tcoratger: Documented and tested that leaves must follow the plan's exact block order.

bus/src/plan.rs
tcoratger: Added a typed tuple-slot API covering payload, domain bit, and zero filler positions.

bus/src/plan.rs
tcoratger: Added an end-to-end materialize-to-GKR terminal relation test on the real output point.

bus/src/plan.rs
tcoratger: Fixed the point convention in the product core; terminal shares now consume the public point directly.

bus/src/plan.rs
```diff
@@ -0,0 +1,1066 @@
...
+    /// Physical blocks for one side of the multiset equality.
+    #[must_use]
+    pub fn blocks(&self, direction: BusDirection) -> &[BusBlock] {
+        match direction {
+            BusDirection::Push => &self.pushes,
+            BusDirection::Pull => &self.pulls,
+        }
+    }
+
+    /// Meaning of one fingerprint slot for one named bus.
```
Nashtare, line 386: With the product point now most-significant-variable-first, the fingerprint point is the one coordinate vector in the crate that still runs the other way. `BusLeaves::materialize` weights slot `i` by `equality_weights(z)`, where `z[0]` binds bit 0 of the slot index (`leaf.rs:230`, and the `00, 01, 10, 11` leaf test). Whatever later rebuilds a terminal share from openings has to weight these slots in exactly that order; reusing the leading-coordinate `eq` there evaluates the fingerprint at reversed `z` and rejects honest proofs, the same kind of mismatch as the point-order thread. Could this doc (or `materialize`'s) say that coordinate `k` of the fingerprint point binds bit `k` of the slot index? Or would you rather switch `equality_weights` to the leading-coordinate order so the crate has one convention?

Nashtare (approved):
I re-ran the end-to-end check against this head and the merged product core. #2216 now returns the point most-significant-variable-first from both `prove` and `verify`, and the split documented on `terminal_shares` holds on it. Locally I built a plan with six AIRs (log heights 3, 2, 2, 1, 0, 0), three buses of widths 3, 1 and 2, and Boolean activation on two declarations. I built every column from `tuple_slot`, materialized with `BusLeaves::materialize` in `blocks()` order, and ran the real prover and verifier with `plan.product_shape()`. The relation holds for both trees on the output point and fails on the reversed point. All 380 product-preserving swaps inside one tree (which the shared-root reduction accepts) break it, and so does every single-leaf edit at all 32 positions per side, padding included. Materializing in AIR order breaks it on exactly the side whose order differs. Over 2000 random statements every block is aligned, disjoint and in range, and every pair of buses differs in a domain slot; on the 379 I also ran through the real prover (every fifth, up to 12 variables) the relation holds on the output point for both trees. `cargo test -p p3-bus` (26 + 1) and clippy with `-D warnings` pass, and the moved leaf and transcript tests are verbatim.

The stored product shape, the typed slot map, the order contract on `terminal_shares` and the README line close those threads, and #2216 closes the point-order one.

The new integration test doesn't yet pin what it says it pins. Its final check is an identity that holds at any point and for any block order, so it still passes with the reversal removed from the product core and with the blocks written in the wrong order. The point order is still covered by the product reference tests, but nothing covers the block order. Details and a fix are in the two thread replies; once the test compares against `values`, the last sentence of the description is accurate. The inline comment is about the fingerprint point's bit order. All three are non-blocking follow-ups.

bus/src/plan.rs
Nashtare: Thanks for adding it. As written, though, the final `assert_eq!` (line 984) compares the block reconstruction with `evaluate(&dense, &prover_output.point)`, and each block is sliced out of `pushes`, which is the vector `dense` is built from. That is an identity: `MLE(dense, q) - 1` splits block by block for any `q` and any contents, so `prover_output.values` is never checked. Locally the assertion also holds at the reversed point and at an unrelated point `[3, 5, 9]`. With both `point.reverse()` calls deleted from `product/proof.rs`, this test still passes and only the two product reference tests fail. Could it compare `reconstructed` with `prover_output.values[tree] - EF::ONE` for both trees, and rebuild each block by materializing its owner's declaration alone instead of slicing `pushes` at `block.offset`? A version doing both passes at this head and fails in both mutants (reversal removed, blocks written in reverse order). Building the two columns from `tuple_slot` would also exercise the slot map end to end.

bus/src/plan.rs
Nashtare: The sentence on `terminal_shares` covers the docs, thanks. The test can't see a wrong order yet, because it reads each block back from the concatenated leaves at the plan's offset. If I materialize `plan.blocks(direction).iter().rev()` instead, the 4-row block lands at offset 2, unaligned, and the test still passes. Rebuilding each block from its owner, as in the other thread, catches it.

PR #2213, opened 2026-09-18
feat(sumcheck): fill suffix witnesses in place

tcoratger:
## What

Adds the first E8.3 ingestion primitive: a reusable suffix-layout plan that lets heterogeneous witness producers write directly into their final committed column slots.

The API:

- derives placement from verifier-visible `TableShape`s;
- gives each producer all of its disjoint column windows together;
- keeps preprocessing suffixes and stack padding initialized to zero and inaccessible;
- checks source shape and column-level write coverage;
- accepts dense, fragmented, or packed sources without an intermediate scalar table;
- can attach existing dense tables to the filled stack without restacking.

This is intentionally not the complete E8.3 endpoint. Custom packed sources can construct the committed polynomial directly, but current suffix sumcheck rounds still read dense `Table`s. Removing that second representation requires a separate table-access abstraction.

## Measurement

Packed bits were either expanded into four BabyBear columns and then stacked, or decoded directly into the final stack.

| Rows | Materialize + stack | Direct fill | Change |
|---:|---:|---:|---:|
| 2^16 | 236.69 µs | 148.20 µs | −37.4% |
| 2^18 | 940.51 µs | 600.96 µs | −36.1% |
| 2^20 | 8.0231 ms | 2.4482 ms | −69.5% |

At 2^20 rows, direct fill also avoids the extra 16 MiB scalar table allocation. The existing path retains both that table and the 16 MiB committed stack; this primitive retains only the stack.

## Validation

- 34 focused witness tests
- `cargo check -p p3-sumcheck --bench sumcheck`
- focused lib and benchmark clippy with warnings denied
- focused rustdoc with warnings denied
- nightly rustfmt and diff checks

Stacked on #2206. No full workspace tests were run.


sumcheck/src/layout/witness.rs
```diff
@@ -176,59 +177,410 @@ impl<F: Field> Table<F> {
...
+    pub const fn is_empty(&self) -> bool {
+        self.values.is_empty()
+    }
+
+    /// Gives a producer exclusive access to the complete logical column.
+    ///
+    /// Coverage is recorded after the writer returns.
+    pub fn write_with(&mut self, writer: impl FnOnce(&mut [F])) {
+        writer(self.values);
+        self.written = true;
```
Nashtare, line 230: `written` flips as soon as the closure returns, whatever it wrote, so a producer that fills only `destination[..n / 2]` leaves the rest as zeros and `fill` still returns `Ok`. The description's "column-level write coverage" is accurate, but the field doc ("Whether the producer completed the column write") and `UnwrittenColumn` ("returned without completing one of its columns") promise row-level coverage. Could both say "without calling `write_with` for column N" instead? If row-level coverage is wanted, a counted entry point such as `write_from_iter(impl ExactSizeIterator<Item = F>)` that checks the length would give packed and chunked producers that check for free.

sumcheck/src/layout/witness.rs
```diff
@@ -176,59 +177,410 @@ impl<F: Field> Table<F> {
...
+    placements: Vec<TablePlacement>,
+    /// Arity of the final stacked polynomial.
+    num_variables: usize,
+    /// Number of suffix variables reserved for preprocessing.
+    folding: usize,
+}
+
+/// Dimensions that cannot form a suffix-layout witness.
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum SuffixLayoutPlanError {
```
Nashtare, line 264: Could `SuffixLayoutPlanError` and `SuffixFillError` be `#[non_exhaustive]`, like the `ring_switch` errors, since this is the first E8.3 step and row-level coverage would be a natural new variant?

sumcheck/src/layout/witness.rs
```diff
@@ -176,59 +177,410 @@ impl<F: Field> Table<F> {
...
+        self,
+        mut tables: Vec<Table<F>>,
+    ) -> Result<Witness<F>, SuffixFillError> {
+        if tables.len() != self.plan.source_shapes.len() {
+            return Err(SuffixFillError::TableCount {
+                expected: self.plan.source_shapes.len(),
+                actual: tables.len(),
+            });
+        }
+        for (table, (source, expected)) in tables.iter().zip(&self.plan.source_shapes).enumerate() {
```
Nashtare: This checks the table count and each table's shape, but nothing ties the tables' values to the stack that `fill` wrote. The two then go separate ways: `Layout::commit` Merkle-commits `witness.poly`, the filled stack, while `SuffixProver::from_witness` drops the poly, and `record_opening`, the SVO accumulators and `compress_stacked` all read `claims.tables`. So if a single cell differs (say a packed source and its dense materializer disagree on bit order, or a `write_with` closure writes only a prefix), the prover's evaluations come from the tables while its openings come from the codeword of the stack, and the verifier rejects an honest proof far from the cause. That's completeness only, since the verifier never trusts the tables, but the fixtures already write the bit decode twice, once in the source and once in the materializer, which is exactly where this drifts. Could the doc state the precondition (`tables[t]` holds, column for column, exactly what the source for table `t` wrote, and the result is for `SuffixProver` only), and could a debug build compare every stack slot with `tables[t].poly(c)` after `pad_zeros`? A constructor that derives the tables from the stack would make it hold by construction, for one extra copy.

Nashtare (commented):
I read the effective change (`f7e03a7b3` + `f318ff0d6`) against `Witness::new`, and direct fill puts every cell where `Witness::new` does. The plan pads each shape to `max(nv, folding)` as `pad_zeros(folding)` does, the shapes reach the same `plan_layout` through `plan_stacked_layout`, and both paths split the buffer with the same `column_slots_for_shapes`. Each column fills `slot[..2^nv]` and the rest of its slot stays zero, which is exactly where `widen_right` appends its zeros.

There is no `unsafe`: every window comes from `split_at_mut`, and `write_with` takes an `FnOnce`, so a producer can't keep a window past the call. Producers only ever receive `slot[..2^nv]`, so the preprocessing padding and the stack tail are unreachable by type. The count, shape and coverage checks return `Result`s, so they also run in release builds. The change only adds items and re-exports, so no `!` is needed.

The branch still carries the history from before the rebase: the six #2203 commits and the old #2215 commits `11400b7a8`/`a02bbc0e1`, which is why GitHub shows 20 files and +3177 lines. The PR's own change is +746/−22 in 3 files, and it applies cleanly both onto the rebased #2215 head `4a91fc1a7` and onto main. Could you either run `git rebase --onto origin/perf/e5-cache-air-profiles a02bbc0e1` to keep the stack, or rebase onto main and retarget to `main`, since nothing here depends on #2215? Rebasing onto main while keeping #2215 as the base would still show #2203 in the diff, because `4a91fc1a7` sits on `7188aed08`, before #2203 landed.

- `into_witness_with_tables` checks the count and shapes of the attached tables but never their contents, so tables that differ from the filled stack give an honest proof that the verifier rejects (inline, `witness.rs`).

**Smaller things**
- "Stacked on #2206" in the description is stale: #2206 is closed and was reopened as #2215.

sumcheck/src/layout/witness.rs
```diff
@@ -176,59 +177,410 @@ impl<F: Field> Table<F> {
...
         let (slot, tail) = tail.split_at_mut(len);
-        slots.push((slot, table_idx, poly_idx));
+        slots.push((slot, table, column));
         rest = tail;
         consumed = offset + len;
     }
     slots
 }
 
+/// Splits `out` into one disjoint slot per placed column, after `num_folded` suffix variables.
```
Nashtare, line 572: Could the blank `///` after this line come back, so rustdoc doesn't fold the next sentence into the summary?

sumcheck/src/layout/witness.rs
```diff
@@ -463,6 +816,129 @@ mod tests {
...
+                        .selectors()
+                        .iter()
+                        .map(|selector| (selector.num_variables(), selector.index()))
+                        .collect(),
+                )
+            })
+            .collect()
+    }
+
+    fn assert_direct_fill_matches_dense(tables: &[Table<F>], folding: usize) {
```
Nashtare, line 929: Both callers pass `folding = 0` (lines 1400 and 1412), so direct fill is never compared with `Witness::new` under preprocessing padding, where padding changes the committed shapes and can reorder placements, and the `pad_zeros` branch of `into_witness_with_tables` never runs. The only `folding > 0` test uses one single-column table and checks raw positions. Could a case with arities below and above `folding` be added, e.g. shapes `(nv, width)` = `(1, 2), (5, 1), (3, 3), (2, 2)` with `folding = 3`, or a proptest of direct fill against `Witness::new` over `arb_table_shapes()` and `folding in 0..=6`? Two more gaps: no single `fill` mixes dense, chunked and packed sources of different heights, which is the heterogeneous case the PR is for, and nothing runs `fill` → `into_witness_with_tables` → commit → prove → verify. A roundtrip test, plus a negative one with one perturbed table, would also pin the precondition on `into_witness_with_tables`.

sumcheck/benches/sumcheck.rs
```diff
@@ -949,6 +953,85 @@ fn layout(c: &mut Criterion) {
...
+            });
+        }
+    }
+}
+
+fn materialize_bit_table(columns: &[Vec<u64>], k: usize) -> Table<BabyBear> {
+    let rows = 1 << k;
+    let values = columns
+        .iter()
+        .flat_map(|words| {
```
Nashtare: This baseline isn't preallocated: collecting through `flat_map` starts at one column's capacity and regrows as later columns arrive. A probe of the same iterator shape at k = 20 shows one 2^20-element allocation and two regrowths, to 2^21 and 2^22 elements, so up to 12 MiB of extra copying. The description's numbers also scale oddly: from 2^18 to 2^20 the baseline goes up 8.5x (940.51 µs → 8.0231 ms) while direct fill goes up 4.1x. As far as I can tell, part of the −69.5% may come from allocation and regrowth rather than the saved table, though I haven't measured that; would a `Vec::with_capacity(4 << k)` baseline be a fairer stand-in for a production materializer? Separately, the 16 MiB saving holds for `fill` alone, but proving from the result today goes through `into_witness_with_tables`, which needs the dense tables anyway, so could the Measurement section repeat the "not the complete E8.3 endpoint" caveat?

sumcheck/src/layout/witness.rs
tcoratger: Marked both public error enums non-exhaustive.

sumcheck/src/layout/witness.rs
tcoratger: Removed the caller-supplied table attachment path. Dense suffix tables are now derived from the committed stack, so divergent contents are unrepresentable.

sumcheck/src/layout/witness.rs
tcoratger: Added randomized folding differential coverage, a heterogeneous source test, and a direct-fill suffix prove/verify roundtrip.

sumcheck/benches/sumcheck.rs
tcoratger: Preallocated the scalar baseline exactly, re-ran the benchmark, and updated the PR measurement and endpoint caveat.

sumcheck/src/layout/witness.rs
tcoratger: Restored the rustdoc summary separator.

sumcheck/src/layout/witness.rs
tcoratger: Clarified that coverage records invocation of the whole-column writer, without claiming per-row initialization tracking.

sumcheck/src/layout/witness.rs
```diff
@@ -463,6 +826,127 @@ mod tests {
...
+            .collect::<Vec<_>>();
+        let direct = plan.fill(&sources).expect("every dense column is written");
+
+        assert_eq!(direct.poly().as_slice(), legacy.poly().as_slice());
+        assert_eq!(direct.table_shapes(), legacy.table_shapes());
+        assert_eq!(
+            placement_addresses(direct.placements()),
+            placement_addresses(&legacy.placements)
+        );
+        let compatible = direct.into_witness();
```
Nashtare, line 945: After `into_witness` this helper checks `poly()` and `table_shapes()`, and both pass through untouched from the filled stack and the plan. So the new proptest and the mixed-source test would still pass if `into_witness` put a column in the wrong place in its tables. Only the one roundtrip fixture (arities 9 and 10, nothing below `folding`) exercises the derived values. Could the helper also compare `compatible.tables[t].iter_polys()` with `legacy.tables[t].iter_polys()` for each `t`? A local probe over 5000 random batches shows they match today, so this would just pin that down.

Nashtare (approved):
I re-checked at `058e9b2c6`. `into_witness` now builds every dense table from the filled stack. `FilledSuffixWitness` can only come out of `fill` and gives no mutable access to its poly, so the tables the suffix rounds read can no longer differ from the committed stack.

A randomized probe backs this up. It ran 5000 batches of dense, chunked and bit-packed sources: arities 0–9, widths 1–7 (8674 tables with non-power-of-two widths), folding 0–6 (4497 tables below it). Against `Witness::new`, the stack, shapes, arity and every derived table match. In the 4214 cases where the suffix prover can run, the whole transcript is byte-identical, with and without `parallel`. That covers every opening (including successor openings), the virtual claim, the preprocessing rounds, every remaining round and the final value. Perturbing one reference cell makes the transcript check fail, so the comparison is live. `cargo test -p p3-sumcheck` and clippy with `-D warnings` pass, and the branch merges cleanly into current main.

The derivation costs one copy of the table view, about 0.31 ms at 4 × 2^20 cells. With the preallocated baseline, `fill` alone is about 21% faster at every size. `fill` + `into_witness`, the path that reaches `SuffixProver`, is 2–6% faster and allocates the same dense tables as the baseline. That's the expected price and fits the "not the complete E8.3 endpoint" scope, but the description should say it (numbers in the bench thread).

The table derivation, the `#[non_exhaustive]` enums, the coverage wording, the restored doc break, the preallocated baseline and the new proptest, mixed-source and roundtrip tests close the earlier threads.

**Smaller things**
- The description still says "Stacked on #2206" (the base is `main` now), lists attaching existing dense tables without restacking (that method is gone), says the primitive "retains only the stack", and shows the old measurement table.

The inline comment is a non-blocking test follow-up.

sumcheck/benches/sumcheck.rs
Nashtare: The preallocated baseline is in, thanks, but the PR description still shows the old table (−37.4% / −36.1% / −69.5%) and still says the primitive "retains only the stack", so I think the edit didn't save. Locally at `058e9b2c6`, the bench in this PR gives 124.4 → 98.7 µs at 2^16, 495.3 → 392.9 µs at 2^18 and 2.042 → 1.603 ms at 2^20, about −21% at each size. The baseline at 2^20 is now about 2.0 ms rather than 8.0 ms. Now that `into_witness` copies the stack back into dense tables, `fill` + `into_witness`, which is what `SuffixProver` needs, measured 126.8 → 124.8 µs, 499.3 → 472.4 µs and 2.071 → 1.942 ms (−2%, −5%, −6%), and it allocates the same 16 MiB of tables at 2^20. Could the Measurement section show that row too, or a `direct_fill_into_witness` case be added to the bench, so the fill-only number isn't read as the proving-path saving?

PR #2214, opened 2026-09-18
feat(security): account for binary bus soundness

tcoratger:
## Summary

Stacked on #2212.

Add exact soundness accounting derived directly from the verifier-owned bus plan.

| Error source | Bound |
|---|---:|
| tuple fingerprint and offset | max(1, s) * N / |F| |
| radix-four degree-five sumchecks | 5R / |F| |
| per-layer tree batching | L(T - 1) / |F| |
| child-claim collapse | h / |F| |

Here s is the number of tuple MLE variables. N is the larger non-padding side. R, L, and h come from the concrete product-GKR schedule.

The four labelled components are diagnostic. Their probabilities are summed into one binary-bus term before composition with a STARK security report. Passing the four components as separate extras would be unsound because the report takes the minimum across protocol-level terms.

The public adapter accepts a checked bus plan and a nonzero field width. Raw geometry cannot be forged by callers. Schedule counts are derived from the product shape rather than reimplemented in the security crate.

## Security checks

- even, odd, and zero-height product trees;
- every height from zero through forty against the concrete layer schedule;
- union-bound composition through the STARK report path;
- malformed and overflowing numeric profiles;
- the large-integer upward-rounding branch;
- exact 64-bit improvement from 128- to 192-bit challenges.

## Validation

- cargo test -p p3-security --lib bus::tests: 8 passed
- cargo test -p p3-bus --lib: 27 passed
- focused clippy and rustdoc for both crates
- git diff --check

No full workspace tests were run.


bus/src/security.rs
```diff
@@ -0,0 +1,130 @@
...
+    // The redundant capacity fields make accidental geometry drift detectable here.
+    let shift = u32::try_from(geometry.log_logical_leaf_count).ok()?;
+    let logical_leaf_count = 1usize.checked_shl(shift)?;
+    let product_shape = ProductGkrShape::new(
+        geometry.log_logical_leaf_count,
+        geometry.tree_count,
+        ProductGkrRootShape::FirstTwoShared,
+    )
+    .ok()?;
+    if geometry.logical_leaf_count != logical_leaf_count
```
Nashtare: These checks only tie `logical_leaf_count` and `layer_count` to `log_logical_leaf_count`, and `BusSecurityGeometry`'s fields are all `pub`, so the two inputs that drive the fingerprint term, `non_padding_leaf_counts` and `tuple_variables`, reach the model unchecked. Locally, at `field_bits = 128`, `h = 20` and `s = 3`, the honest geometry reports 106.4 bits for the fingerprint term; the same geometry with counts `[1, 1]` is accepted and reports 126.4. `tuple_variables = 0` (108.0 bits) and a consistently lowered height (`h = 10` with counts `2^10`, 116.4 bits) are accepted as well. The verifier builds its own plan, so this is about caller error rather than an attacker, but the description ("rejects inconsistent capacities, layers, heights, or tree counts") and the name `adapter_rejects_forged_redundant_geometry` promise more than this checks. Could this take `&BusPlan` instead, whose fields are private and whose only constructor is `build`? Checking `max(counts).max(1).next_power_of_two() == logical_leaf_count` and renaming the test would be the smaller fix, though it only catches the first case.

security/src/bus.rs
```diff
@@ -0,0 +1,276 @@
...
+        Some(Self {
+            field_bits,
+            tuple_variables,
+            non_padding_leaf_counts,
+            product_log_height,
+            product_tree_count,
+        })
+    }
+
+    /// Return each algebraic error source as a separate report term.
```
Nashtare: These four terms are one sub-protocol's union bound, but passed as `extras` they are combined by `min` (`proven_security_regime` takes `ErrorBits::min` over everything), which charges the largest single error instead of the sum. `binary.rs` sums its fold, sumcheck and batching errors into one numerator for exactly this reason. The gap is small but on the unsafe side: 1.58 bits at `h = 3` (numerators 5, 5, 2, 3), 1.32 bits at `h = 4`, 0.15 bits at `h = 10`. Could `terms()` also return one combined term, or document that callers must `ErrorBits::sum` these four, with a test that composes them through `proven_security_report` like the one in `logup.rs`?

security/src/bus.rs
```diff
@@ -0,0 +1,276 @@
...
+            error_from_numerator(self.field_bits, fingerprint_numerator),
+        )];
+
+        // Radix four contracts two multiplication levels after any leading binary layer.
+        // A layer at point length r runs r degree-five sumcheck rounds.
+        let mut remaining = self.product_log_height;
+        let mut point_len = 0usize;
+        let mut layer_count = 0u128;
+        let mut radix_four_rounds = 0u128;
+        while remaining > 0 {
```
Nashtare: This loop re-implements `ProductGkrShape::layers()`, which `p3-security` can't see. I replayed both for every `h` in `0..=63` and they agree today, but nothing compares the model's `R`, `L` or collapse count with `layers()` (the adapter only checks the geometry's own `layer_count`), so a change to the product schedule, such as a radix-two leaf layer, would leave these numerators silently stale. Could `bus/src/security.rs` get a test that, for each `h` up to 40 or so, checks `5 · Σ rounds` and `Σ log2(arity)` from `layers()` against the model's sumcheck and collapse terms? Passing `R` from `layers()` into the model would also work.

security/src/bus.rs
```diff
@@ -0,0 +1,276 @@
...
+        ));
+    }
+}
+
+/// Convert an exact numerator over the challenge-field order into bits.
+fn error_from_numerator(field_bits: usize, numerator: u128) -> ErrorBits {
+    // Round upward before taking the logarithm.
+    // This keeps the reported error at least as large as the exact rational bound.
+    let mut rounded = numerator as f64;
+    if (rounded as u128) < numerator {
```
Nashtare, line 191: The description lists "upward rounding of large integer error numerators" among the security checks, but no test reaches this branch: the largest fixture numerator is `4 · 2^20`. It is reachable, e.g. `BusSecurityModel::new(128, 3, [(1 << 62) + 1, 1], 63, 2)`, where `3 · 2^62 + 3` rounds to `3 · 2^62` and `next_up` gives `3 · 2^62 + 2048`. Both values then give the same `log2` (64.415 bits), so the comment above holds only up to f64 rounding, and a test here can do little more than exercise the branch. Could the comment say "up to f64 rounding", and either that case become a test or the line come out of the description?

Nashtare (commented):
I re-derived all four numerators from the code #2216 and #2212 actually run, and they're correct:
- `max(1, s) · N` for the fingerprint. Each active factor `γ − fp(t)` has total degree `max(1, s)` in `(α, γ)`, the shared root forces `P = Q` exactly, and `N` counts every declared row, including deselected ones, which over-counts safely.
- `5R` for the sumchecks. Each radix-four round is `eq` times four children, so degree 5, and the binary layer only ever sits at the root with an empty point. Replaying the model's loop and `ProductGkrShape::layers()` for every `h` in `0..=63` gives the same `R` from both: `m(m − 1)` for `h = 2m` and `m²` for `h = 2m + 1`.
- `L(T − 1)` for tree batching. A fresh `ρ` is drawn per layer after the claims it combines, the root layer included.
- `h` for child collapse: 2 per radix-four layer from the bilinear `interpolate_quad`, 1 for the binary root layer.

Rounding goes the safe way (error up, bits down), and no numerator can overflow `u128`.

- `bus_security_terms` trusts a caller-built geometry whose binding inputs are never checked: counts `[1, 1]` at `h = 20` are accepted and put the fingerprint term at 126.4 bits instead of 106.4 (inline, `bus/src/security.rs`).

**Smaller things**
- In the description, `s` is the number of tuple MLE variables (`log2` of the tuple width) rather than the tuple-table arity; the code docs already say this.

The other inline comments cover how the four terms combine with the STARK terms, the duplicated layer schedule, and the rounding test the description lists.

security/src/bus.rs
tcoratger: Exercised the large-integer rounding branch and documented that conservatism is bounded by floating-point rounding.

security/src/bus.rs
tcoratger: Removed the duplicated schedule: security dimensions now derive from the checked product layers, with parity tests through height 40.

security/src/bus.rs
tcoratger: Combined the four diagnostic components into one union-bound security term and tested composition through the public report path.

bus/src/security.rs
tcoratger: The adapter now accepts a BusPlan reference, keeps geometry private, and requires a nonzero field width, eliminating forged public geometry.

bus/src/security.rs
```diff
@@ -0,0 +1,179 @@
...
+        .unwrap()
+        .unwrap()
+    }
+
+    #[test]
+    fn adapter_preserves_components_and_returns_their_union() {
+        // Height four has two radix-four layers with zero and two rounds.
+        let plan = plan(4, 5);
+        let field_bits = NonZeroUsize::new(128).unwrap();
+        let components = plan.security_components(field_bits);
```
Nashtare, line 109: This test checks the labels and that the combined term is the sum of the components. The height sweep below checks the three product-GKR components. But no test checks the fingerprint value that a plan produces, and that is the term the forged geometry moved last round. It also dominates at every realistic size. For `plan(4, 5)` the value is easy to pin. Five payload slots plus one domain bit pad to width 8, so `s = 3`. Each side has 16 rows, so the fingerprint term is `128 − log2(48)`, about 122.415 bits. The combined term is exactly `122.0`, because `48 + 10 + 2 + 4 = 64`. Could this test assert both?

security/src/bus.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// Degree-five sumcheck rounds across every reduction layer.
+    sumcheck_rounds: usize,
+    /// Root-to-leaf reduction layers.
+    layer_count: usize,
+    /// Random coordinates used to collapse child claims.
+    collapse_challenges: usize,
+}
+
+impl ProductGkrSecurityProfile {
+    /// Validate counts derived from one concrete product-reduction schedule.
```
Nashtare: This `new` and `BusSecurityModel::new` check three things: that the dimensions are representable, that `collapse_challenges == log_height`, and that `layer_count` is zero exactly when the height is. They trust everything else: `sumcheck_rounds`, `layer_count`, `tree_count`, `tuple_variables` and the leaf counts. Locally, `ProductGkrSecurityProfile::new(20, 1, 0, 1, 20)` with `BusSecurityModel::new(128, 0, [1, 1], ..)` is accepted and gives 123.6 bits, while the plan for `h = 20`, `s = 3` gives 106.4. That's fine for the numeric layer, since `BusPlan::security_term` is the checked path. But could these docs say that the counts are trusted, point callers to `BusPlan::security_term`, and drop "Validate counts derived from one concrete product-reduction schedule"? Could `BusPlan::security_term` also say what `field_bits` means? It should be a lower bound on `log2 |EF|` for the single field from which `α`, `γ` and every GKR challenge are drawn. Right now only the private field doc says so.

security/src/bus.rs
```diff
@@ -0,0 +1,402 @@
...
+                .map(|component| component.bits)
+                .collect::<Vec<_>>(),
+        );
+        assert_eq!(model.combined_term().bits, expected);
+        assert_eq!(model.combined_term().label, BINARY_BUS_LABEL);
+    }
+
+    #[test]
+    fn combined_term_composes_as_one_protocol_extra() {
+        use crate::fri::FriRegime;
```
Nashtare: Could these four `use` lines move to the top of `mod tests`, next to `use super::*;`?

Nashtare (approved):
I re-derived the four terms against the restacked product GKR (`product/proof.rs`, `shape.rs`, `math.rs` and `transcript.rs`). They are unchanged and correct:
- `max(1, s) · N` for the fingerprint;
- `5R` for the degree-five rounds;
- `L(T − 1)` for tree batching;
- `h` for the collapse coordinates.

The transcript draws only batching, round and branch challenges, all over the challenge field, and every one of them is charged.

To check the numbers end to end, I built ten plans with `BusPlan::build`: heights 0 to 25, one to three named buses, mixed block heights, and both odd and even trees. For each, `security_term(128)` agrees with `128 − log2(Σ numerators)`, computed exactly, to within one ulp. For eight of them I also proved and verified a product GKR on `plan.product_shape()`. Each proof draws exactly `L` batching, `R` round and `h` branch challenges, e.g. 10, 90 and 20 at `h = 20`.

The new commits close the earlier threads:
- the adapter is now a pair of `BusPlan` methods;
- the four terms are summed into one `binary-bus` term;
- the counts come from `layers()`;
- the rounding branch is exercised.

A struct literal of `BusSecurityGeometry` no longer compiles, and `BusPlan` has no constructor besides `build`. The combined term composes with the STARK terms the same way the opening term in `binary.rs` does. Locally, `cargo test` and `clippy -D warnings` are clean on both crates.

The inline comments are non-blocking follow-ups: one test assertion, docs on the raw numeric model, and where some imports sit.

PR #2215, opened 2026-09-18
perf(multi-stark)!: cache AIR profiles at setup

tcoratger:
## Summary

Compute each AIR symbolic profile once during setup and retain it in the matched proving and verifying keys. Reused keys no longer walk the AIR symbolically before every zerocheck proof and verification.

This changes neither the transcript nor the proof. The cached values are the constraint count and the ordinary/interaction degrees already fixed by the setup AIR batch.

## E5 audit

- **5.1 — friendly challenges:** already shipped in #2159; no duplicate work.
- **5.2 — elide the C side:** specific to Flock’s batch-R1CS split. Plonky3’s AIR zerocheck has no separately materialized `a · b - c` sides to elide.
- **5.3 — circuit walking:** specific to Flock’s lincheck matrix. The current multi-STARK path evaluates AIR constraints directly and has no substituted sparse matrix.
- **5.4 — packed round kernels:** #2203 covers the material early-round kernel. This PR removes the remaining invariant symbolic pass from repeated proof and verification.
- **5.5 — multipoint ring-switch batching:** the existing bit ring switch already batches its row identities into one degree-two sumcheck and the Boolean PCS discharges the resulting claims in one multi-opening. Flock’s newer capacity-free assist depends on the still-missing jagged/no-pad stacking path.
- **5.6 — AG-code skip:** remains research. Flock’s implementation is tied to its R1CS skip protocol and generated AG-code constants; importing it before the corresponding reduction exists here would add an unauditable, unused protocol.

## Benchmark

`optimized`, Keccak binary AIR, log height 10 (40 permutations), 95-bit target, 20 interleaved runs on the same machine:

| phase | main | this PR | change |
|---|---:|---:|---:|
| prove | 323.30 ms | 321.40 ms | -1.90 ms |
| verify | 30.75 ms | 26.45 ms | -4.30 ms |

This is a fixed per-AIR cost, not a throughput-scaled kernel win. Setup moves the work outside the timed proof path, so the saving is amortized when a key is reused; first-proof setup plus proving is intentionally unchanged.

## Validation

- `cargo clippy -p p3-multi-stark --lib --examples -- -D warnings`
- `cargo test -p p3-multi-stark --lib zerocheck` (55 passed, 1 heavy test ignored)
- `cargo test -p p3-multi-stark --example prove_binary_field` (9 passed)
- `cargo test -p p3-multi-stark --example prove_boolean_trace` (9 passed)


multi-stark/src/keys.rs
```diff
@@ -70,9 +77,13 @@ pub fn setup<C, A>(
 ) -> Result<(ProvingKey<C>, VerifyingKey<C>), ProvingError<PcsProverError<C>>>
 where
     C: MultiStarkConfig,
-    A: BaseAir<C::Val>,
+    A: BaseAir<C::Val> + Air<InteractionSymbolicBuilder<C::Val, C::Challenge>>,
     Commitment<C>: Clone,
 {
+    let air_profiles = airs
```
Nashtare, line 89: This caches the zerocheck's pass, but two more symbolic evaluations of every AIR still happen on each proof and each verification: `IndexedPlan::build` (via `IndexedLookups::from_air`, `lookup/src/indexed.rs:139-140`) and the lookup plan build (via `Lookups::from_air`, `lookup/src/types.rs:133-139`). Both build the same `InteractionSymbolicBuilder<Val, Challenge>` and call `air.eval`. So one builder per AIR evaluated here could also give the `Lookups` and `IndexedLookups`, and the keys could carry all three. I haven't measured how the 4.3 ms verify saving splits between `air.eval` and the degree/successor analysis that only `get_air_profile` does, so this may or may not be worth it. It would make the description's "no longer walk the AIR symbolically" literally true, though.

multi-stark/src/keys.rs
```diff
@@ -31,12 +34,16 @@ pub(crate) struct PreprocessedProverData<C: MultiStarkConfig> {
...
+    /// Zerocheck metadata fixed by the AIRs at setup.
+    pub(crate) air_profiles: Vec<AirProfile>,
 }
 
 /// The verifier's key for a fixed AIR and trace height.
 pub struct VerifyingKey<C: MultiStarkConfig> {
     /// Batched preprocessed commitment, present only when at least one AIR declares it.
     pub(crate) preprocessed: Option<Commitment<C>>,
+    /// Zerocheck metadata fixed by the AIRs at setup.
+    pub(crate) air_profiles: Vec<AirProfile>,
```
Nashtare, line 52: Until now a `VerifyingKey { preprocessed: None }` was valid for any AIR list without preprocessed columns. With this field it is tied to the exact list, in the same order, that `setup` saw. A mismatch fails closed (`LookupLinkMismatch`, or a shape/degree rejection), so this isn't a soundness concern. But the error a caller gets doesn't say "wrong key". Could the `ProvingKey`/`VerifyingKey` docs say that a key is only valid for the AIRs passed to `setup`, in that order, even when none has preprocessed columns?

multi-stark/src/zerocheck/mod.rs
```diff
@@ -438,7 +440,25 @@ impl<'a, A> AirZerocheck<'a, A> {
...
+        }
+    }
+
+    /// Reuse the symbolic AIR pass performed by multi-STARK setup.
+    pub(crate) fn with_profiles(
+        airs: &'a [&'a A],
+        profiles: &'a [AirProfile],
+        pow_bits: usize,
+    ) -> Self {
+        assert_eq!(airs.len(), profiles.len());
```
Nashtare: Nit: the neighbouring asserts in this impl carry a message ("one public-value group is required for each AIR", …). Could this one say something like "one cached AIR profile is required for each AIR", so a caller who passes the wrong key sees why?

multi-stark/src/keys.rs
```diff
@@ -70,9 +77,13 @@ pub fn setup<C, A>(
 ) -> Result<(ProvingKey<C>, VerifyingKey<C>), ProvingError<PcsProverError<C>>>
 where
     C: MultiStarkConfig,
-    A: BaseAir<C::Val>,
+    A: BaseAir<C::Val> + Air<InteractionSymbolicBuilder<C::Val, C::Challenge>>,
```
Nashtare, line 86: This tightens a public bound. Any AIR that works with `prove`/`verify` already satisfies it, so it's breaking only in principle. Still, the repo usually marks bound changes with `!` in the title (e.g. #2160), and the changelog is built from titles.

Nashtare (approved):
I checked that the cached path is the old computation moved earlier. `setup` computes `get_air_profile::<C::Val, C::Challenge, A>` in `airs` order and hands the same vector to both keys, on the preprocessed and non-preprocessed branches alike. The prover and verifier then run the zerocheck with that same `(Val, Challenge)` pair, so the cached degrees, `num_constraints` and lookup flags are exactly what the per-call pass produced. The transcript and proof are unchanged. The `get_air_profile` panics move from `prove` to `setup`, which is a better place for them.

I also looked at what a key built from a different AIR list does, now that the verifier takes the "declares lookups" bit from the key. It fails closed. `links_by_air` still comes from the real AIRs through `verify_lookup`, so a disagreement in either direction returns `LookupLinkMismatch`. A degree disagreement changes `ZerocheckShape` and the round cap, so the honest proof just stops verifying.

CI ran before #2203 was squash-merged. I type-checked the current merge ref (`p3-multi-stark`, all targets and examples) and it is clean. #2203 adds no other `get_air_profile` call site outside tests.

The inline comments are non-blocking follow-ups.

**Smaller things**
- The description says this "removes the remaining invariant symbolic pass". Two full `air.eval` passes per AIR still run on every proof and verification: `IndexedPlan::build` → `IndexedLookups::from_air`, and the lookup plan → `Lookups::from_air`. Both use the same `InteractionSymbolicBuilder` (see the comment on `setup`).
- No test pins the new invariant. Two cheap ones would be `with_profiles` vs `new` producing byte-identical proofs, and verification with a key from a different AIR list being rejected.

multi-stark/src/keys.rs
tcoratger: Updated the PR title to mark the public-bound change as breaking.

multi-stark/src/keys.rs
tcoratger: Documented that both keys are bound to the exact ordered AIR batch supplied at setup.

multi-stark/src/keys.rs
tcoratger: Resolved by correcting the scope and benchmark description. The lookup passes retain different symbolic data, so combining and caching them would be a separate measured change rather than part of this profile cache.

multi-stark/src/zerocheck/mod.rs
tcoratger: Added the explicit cached-profile count assertion message.

PR #2216, opened 2026-09-18
feat(bus): add direction-aware product GKR core

tcoratger:
## Summary

Add the standalone core of a characteristic-two-safe multiset bus:

- structural push and pull directions;
- fixed-width tuple fingerprints with Boolean row selection;
- batched radix-four product-tree GKR;
- implicit multiplicative-identity suffix padding;
- verifier-derived proof shape and typed Fiat-Shamir transcript;
- a shared terminal point for later PCS authentication.

The existing LogUp path and its characteristic-two guard are unchanged.

## Security boundary

The product proof checks only its internal reduction. The returned leaf evaluations are marked unauthenticated until a commitment opening or table sumcheck binds them.

For a width-2^s tuple and N factors, fingerprint collision is at most max(1, s) N / |F|. Each radix-four sumcheck round contributes at most 5 / |F|. Tree batching contributes at most (T - 1) / |F| per layer. Child collapse contributes one field root per coordinate.

The transcript binds every shape component even for zero-height trees. The public terminal point follows the repository-wide most-significant-variable-first convention.

## Benchmark

The radix comparison is an arithmetic-only model. The checked-in benchmark also times prove_product_gkr directly.

| height | radix two model | radix four model | change |
|---:|---:|---:|---:|
| 14 | 71.510 ms | 65.169 ms | -8.9% |
| 18 | 1.1496 s | 1.0473 s | -8.9% |

## Validation

- cargo test -p p3-bus --lib: 15 passed
- git diff --check


bus/src/product.rs
```diff
@@ -0,0 +1,764 @@
...
+            let children = states
+                .iter()
+                .map(RadixFourState::children)
+                .collect::<Vec<_>>();
+            let branches = transcript.end_radix_four_layer(&children);
+            values = children
+                .iter()
+                .map(|children| interpolate_quad(*children, branches))
+                .collect();
+            point = vec![branches[0], branches[1]];
```
Nashtare: This makes the output point least-significant-bit first: `point = [branches…] ++ round_point`, so `point[k]` binds bit `k` of the leaf index, matching the "low-variable-first" `equality_weights`. The rest of the repo is most-significant-bit first: `multilinear_util::Point::hypercube` puts bit `n − 1 − i` at coordinate `i`, binary-pcs lifts points with `Selector::lift_prefix`, the logUp* `Block::weight` has "the leading coordinates address the block", and multi-stark's `trace_suffix` opens a table's rows on the trailing coordinates. #2212's `terminal_shares` and its test follow that convention and read the leading coordinates as the block prefix. Locally, that split fails on the real GKR output point and holds on the reversed one, so once the plan is wired to this output, honest terminal checks would fail.

Could `ProductGkrOutput::point` be reversed once at the end of both prove and verify (the internal folding can stay LSB-first), with the order stated on the field, so the whole stack uses one convention? A plan → leaves → GKR test that checks the terminal relation on `output.point` would pin it.

bus/src/product.rs
```diff
@@ -0,0 +1,764 @@
...
+        layer: usize,
+    },
+}
+
+/// Prove several product trees in one transcript.
+///
+/// Each slice is an arbitrary leaf prefix.
+/// Every omitted suffix value is the multiplicative identity.
+///
+/// The output values are claims about the input prefixes.
```
Nashtare: The values are evaluations of the ones-padded table, `Σ_{i<2^h} eq(point, i)·ℓ_t[i]` with `ℓ_t[i] = 1` past the prefix, rather than claims about the prefix itself. A PCS opening of a height-`n` column returns the zero-padded MLE, which differs by `Σ_{i≥n} eq(point, i)`, and the verifier's statement doesn't carry the prefix lengths. A probe shows why the distinction matters: a prover that pads with junk instead of ones gets a verifying proof whose leaf claim is the junk table's MLE, and only authenticating the ones-padded MLE catches it. Could this line and the doc on `ProductGkrOutput::values` state the formula? Exporting the O(h) helper for `Σ_{i≥n} eq(point, i)` would also save each caller from rederiving it.

bus/src/product.rs
```diff
@@ -0,0 +1,764 @@
...
+pub struct ProductGkrProof<EF> {
+    /// Root messages encoded according to the statement shape.
+    pub roots: Vec<EF>,
+    /// Layer messages in root-to-leaf order.
+    pub layers: Vec<ProductGkrLayerProof<EF>>,
+}
+
+/// Unauthenticated leaf evaluations produced by a product reduction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ProductGkrOutput<EF> {
```
Nashtare: `verify_product_gkr` returns `Ok` for every internally consistent proof: with `Distinct` roots it checks no relation among them, and the leaf values are unauthenticated. A caller who writes `verify_product_gkr(&proof, shape, &mut ch)?;` and drops the output gets no warning, and has accepted arbitrary roots and leaves. Could this struct get `#[must_use = "leaf evaluations are unauthenticated until bound to committed polynomials"]`, and the `verify_product_gkr` doc say that with `Distinct` it checks no relation among the roots?

bus/src/product.rs
```diff
@@ -0,0 +1,764 @@
...
+    shape: ProductGkrShape,
+    challenger: &mut Challenger,
+) -> Result<ProductGkrOutput<EF>, ProductGkrError>
+where
+    F: TranscriptField,
+    EF: ExtensionField<F>,
+    Challenger: FieldChallenger<F>,
+{
+    // Reject every attacker-controlled length before building a strict transcript driver.
+    if !has_distinct_round_nodes::<EF>() {
```
Nashtare: This is the only guard on the challenge field, and it only asks for six distinct nodes. With `EF = F = BinaryField8`, a probe that grinds over the free child choice produced accepted proofs with wrong roots and leaf claims equal to the honest leaf MLE after 5–284 attempts, including an unbalanced `FirstTwoShared` bus reported as balanced after 5–8 attempts. `EF = BabyBear` (about 2^31) is accepted too. In the other direction, `Gf2` is a `TranscriptField` through `PrimeField64`, and its `interpolation_node` asserts `i < 2`, so this check panics instead of returning `ChallengeFieldTooSmall`. Could the verifier reject challenge fields below a stated size (for example on `EF::bits()`), or at least document the minimum next to the `k/|F|` bounds?

bus/src/product.rs
```diff
@@ -0,0 +1,764 @@
...
+    batching: F,
+) -> [F; ROUND_POLY_LEN] {
+    debug_assert!(logical_len >= 2);
+    debug_assert_eq!(equality.len(), logical_len);
+
+    // Node one is omitted because the running sum reconstructs it.
+    let nodes = [0, 2, 3, 4, 5].map(F::interpolation_node);
+    let mut evaluations = [F::ZERO; ROUND_POLY_LEN];
+
+    for row in 0..logical_len / 2 {
```
Nashtare: This loop visits every row of `logical_len / 2` for every tree, reading `unwrap_or(ONE)` past each prefix, and `equality_weights` is dense, so implicit padding saves memory but not time. A probe measured h = 20 at 422 ms for a 1-leaf prefix against 334 ms for a full one, and since `ProductGkrShape::new` accepts h ≤ 63, the prover runs out of memory from about h = 34 even for a 1-leaf tree. Past a tree's prefix each row contributes `λ^t·eq_node`, which a suffix sum of the eq table covers without the per-child work. Would that, and a `parallel` path for the leaf-level sumcheck, be worth a follow-up?

bus/src/transcript.rs
```diff
@@ -0,0 +1,292 @@
...
+        InteractionPattern::new(steps).expect("one matched bracket per layer is well formed")
+    }
+
+    /// Bind the protocol identity and the complete verifier-derived shape.
+    pub(crate) fn domain_separator<F, EF>(&self) -> DomainSeparator<Alphabet<F>>
+    where
+        F: TranscriptField,
+        EF: ExtensionField<F>,
+    {
+        DomainSeparator::new(VERSION, NAME, self.pattern::<F, EF>())
```
Nashtare: At `log_height = 0` the pattern is only `roots: Fixed(root_message_len)`, so `(Distinct, T)` and `(FirstTwoShared, T + 1)` produce the same seed. A probe confirmed it: a proof for `{h: 0, T: 1, Distinct}` verifies as `{h: 0, T: 2, FirstTwoShared}` (`[r]` becomes `[r, r]`) and leaves the challenger in the same state. These pairs were the only acceptances among 1190 shape mismatches. Soundness per statement holds, because leaf authentication binds each single-leaf tree, but it contradicts the module doc ("enter the transcript seed"), and `transcript_seed_binds_every_shape_dimension` only checks h = 4. Could the seed bind `(log_height, num_trees, root_shape)` explicitly, for example through `DomainSeparator::instance`, with the seed test extended to h = 0?

bus/src/leaf.rs
```diff
@@ -0,0 +1,232 @@
...
+//!
+//! A width-`2^s` tuple is interpreted as a Boolean-cube evaluation table.
+//! Its fingerprint is the table's multilinear extension at a random point in `F^s`.
+//!
+//! Distinct tuples define distinct multilinear polynomials.
+//! Their difference has degree at most one in each of the `s` coordinates.
+//! A random point therefore collides with probability at most `s / |F|`.
+//!
+//! Products of `N` factors are monic polynomials in the random offset.
+//! Unequal multisets collide with probability at most `max(1, s) * N / |F|`.
```
Nashtare: The bound is right, but lines 6–11 don't reach it: composing a per-pair `s/|F|` separation with an `N/|F|` offset step needs a union bound over every pair of tuples, which gives roughly `N²·s/|F| + N/|F|`. The joint argument does give `max(1, s)·N/|F|`. `P(z, γ) = ∏_push(γ − MLE_t(z)) − ∏_pull(γ − MLE_u(z))` is nonzero when the multisets differ, because the factors are distinct monic-in-`γ` irreducibles in `F[z, γ]`. Each factor has total degree `max(1, s)`, so total-degree Schwartz–Zippel over `(z, γ)` finishes it, and it also covers a zero factor on both sides, which is a root of `P`. Could the doc use that argument and define `N = max(#active pushes, #active pulls)`?

bus/src/leaf.rs
```diff
@@ -0,0 +1,232 @@
...
+    Pull,
+}
+
+/// Row activation for one bus declaration.
+#[derive(Clone, Copy, Debug)]
+pub enum BusSelector<'a, F> {
+    /// Every row contributes one factor.
+    Always,
+    /// A Boolean column selects the rows that contribute factors.
+    Boolean(&'a [F]),
```
Nashtare, line 34: `materialize_bus_leaves` rejects a non-Boolean selector, but that check runs only on the prover's side. With an unconstrained committed `s`, the leaf `1 + s·(f − 1)` can take any value on that row, which balances any push/pull product. Could this doc say that the surrounding AIR must constrain the selector to be Boolean?

bus/src/leaf.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub direction: BusDirection,
+    /// Tuple columns in slot order.
+    pub columns: &'a [&'a [F]],
+    /// Row activation mode.
+    pub selector: BusSelector<'a, F>,
+}
+
+/// Materialized product leaves, kept separate by direction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BusLeaves<EF> {
```
Nashtare, line 50: `materialize_bus_leaves` appends every declaration of one direction to a single vector, so declaration `k` starts at the unaligned offset `Σ_{j<k} height_j`. Authenticating its leaves then needs a shifted `eq(point, offset_k + row)` weight rather than a sub-cube eq. Declarations of different buses passed in one call also share `(z, γ)` and one product, so the domain-separator slot at line 113 is then required rather than a recommendation. Could `BusLeaves` document the order and offsets, and line 113 say that separating buses is the caller's job?

bus/src/leaf.rs
```diff
@@ -0,0 +1,232 @@
...
+    EF: ExtensionField<F>,
+{
+    // The point fixes the only tuple width that the fingerprint polynomial accepts.
+    let shift = u32::try_from(point.len()).map_err(|_| BusLeafError::TupleWidthOverflow)?;
+    let width = 1usize
+        .checked_shl(shift)
+        .ok_or(BusLeafError::TupleWidthOverflow)?;
+
+    // One equality table supplies the linear coefficient of every tuple slot.
+    let weights = equality_weights(point);
```
Nashtare: This allocates `2^point.len()` weights before the width check below. Locally, `point.len() = 24` with a width-1 declaration used 272 MB and 0.49 s before returning `TupleWidthMismatch`, and since `TupleWidthOverflow` only fires at 64, lengths from about 36 to 63 abort on allocation instead of returning the documented error. Could the `columns.len() == width` check run over every declaration before this allocation?

bus/benches/product_gkr.rs
```diff
@@ -0,0 +1,180 @@
...
+    for row in 0..output_len {
+        let zero = values.get(2 * row).copied().unwrap_or(F::ONE);
+        let one = values.get(2 * row + 1).copied().unwrap_or(F::ONE);
+        values[row] = zero + challenge * (one - zero);
+    }
+    values.truncate(output_len.min(logical_len / 2));
+}
+
+/// Exercise all product-tree and sumcheck arithmetic for one schedule.
+fn arity_workload(inputs: &[Vec<F>; 3], log_height: usize, radix_four: bool) -> F {
```
Nashtare, line 69: `arity_workload` is a separate model, so the "−8.9% vs radix-2" figure in the description isn't measured on `prove_product_gkr`, which has no benchmark. The model also drifts from the prover. `message` (line 97) accumulates across rows, so `digest += eq * message` (line 113) sums prefix sums rather than a round polynomial. For odd `log_height`, `build_layers` puts the binary step at the root while the descent at line 75 puts radix-4 there, so it reads `layers[h − 2]`, `layers[h − 4]`, …, which were never built; only `[14, 18]` runs, so that one is latent. Could the bench time `prove_product_gkr` directly, and keep the radix-2 model only as a comparator, with `message` reset per row and an even-height assert?

bus/src/tests.rs
```diff
@@ -0,0 +1,360 @@
...
+    // Depth zero exercises the root-only transcript.
+    // Depths one through ten exercise odd and even radix-four schedules.
+    let mut rng = Xoroshiro128Plus::seed_from_u64(0xB1A4_0001);
+    for log_height in 0..=10 {
+        let capacity = 1usize << log_height;
+        let inputs = (0..3)
+            .map(|tree| {
+                let explicit = capacity.saturating_sub(tree.min(capacity));
+                (0..explicit)
+                    .map(|_| Binary::from_u64(rng.random::<u64>()))
```
Nashtare: `BinaryField128::from_u64` maps through the prime subfield, so this is `x mod 2`: `from_u64(0..8)` is `[0, 1, 0, 1, 0, 1, 0, 1]`. Every binary leaf in the suite is therefore 0 or 1. The tamper test's inputs are `[1, 0, 1, …]` with both roots 0, `shared_roots_are_structural` has `left = [0, 1, 1]` and `count = [1, 1]` (empty after trimming), and the bench's `index·0x9E37_79B9 + 1` alternates 1, 0. A probe with random GF(2^128) leaves passed 200/200, so this is about coverage, not a bug. Could these use `rng.random::<BinaryField128>()` instead?

bus/src/tests.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    let mut verifier_challenger = binary_challenger();
+    assert_eq!(
+        verify_product_gkr::<Binary, Binary, _>(&proof, shape, &mut verifier_challenger).unwrap(),
+        output,
+    );
+}
+
+#[test]
+fn tampering_each_message_family_is_rejected() {
```
Nashtare, line 194: This tampers only layer-0 children at h = 4 with `Distinct` roots, and asserts `is_err()`, so it would also pass if the rejection came from the wrong check. Nothing tampers binary-cap children (odd h) or a `FirstTwoShared` proof, for example changing `proof.roots[0]`, or feeding a shared-root verifier a proof built with distinct roots. `RootCountMismatch`, a Binary/RadixFour variant swap and a wrong `round_polys.len()` aren't exercised either. A probe also showed that a leaf-layer child edit that keeps the product (a swap, or `c0·k, c3/k`) is accepted by the core with wrong leaf claims, as documented, and a test pinning that would make the contract explicit. Could the negative tests cover these and match on `LayerConsistency { layer }`?

Nashtare (approved):
I read the whole crate and ran a probe against it locally, and the product reduction holds up.

By reading: the child selector is the low two bits of the leaf index, and the prover, the verifier, the output point and `naive_eval` all agree on that. The degree-5 round polynomial (sent at nodes 0, 2, 3, 4, 5, with p(1) rebuilt from the running sum as `RoundPolyInterpolator` expects), the `eq(r, r')·Σ λ^t ∏ children` closing check, the 4→1 and 2→1 collapses, the odd-height binary cap and heights 0 and 1 are all right. Roots, round polynomials and children are each absorbed before the challenge that depends on them, and λ is redrawn per layer. `validate_proof_shape` checks every proof-controlled length before the transcript exists, and all later indexing is shape-derived. The `max(1, s)·N/|F|` fingerprint bound is correct, though the doc's argument for it has a gap (inline).

The probe:
- An independent re-implementation of the transcript and prover reproduced the real prover's proof, point and values in 400/400 cases (GF(2^128) and BabyBear^4, h = 0..9, T = 1..3, both root shapes).
- 5656 per-field mutations and 327 malformed-length proofs gave no forgery with correct leaf claims and no verifier panic. The mutations the core accepts (product-preserving edits at the leaf layer, root edits at h = 0) all return wrong leaf claims, so leaf authentication rejects them.
- A prover lying about a root (96 cases per field, plus 16 unbalanced `FirstTwoShared` statements) always ended on wrong leaf claims.
- 1190 prover/verifier shape mismatches were rejected apart from the h = 0 alias (inline, `transcript.rs`). Even h = 63 and T = usize::MAX/4 return errors without allocating from the shape.

The inline comments are non-blocking follow-ups. The one I'd settle before the bus plan lands is the bit order of `ProductGkrOutput::point` (inline, `product.rs`), since #2212 already reads it the other way.

**Smaller things**
- `p3-bus` has no entry in `release-plz.toml`. Every other publishable crate has a `[[package]] name = "p3-…" version_group = "plonky3"` entry, and #2021 had to add the same one for `p3-stir` afterwards.

bus/src/product.rs
tcoratger: Marked the output must-use and documented that distinct roots have no relation until the caller authenticates roots and terminal leaf evaluations.

bus/src/leaf.rs
tcoratger: Replaced the informal argument with the joint monic-polynomial Schwartz-Zippel proof and defined the active-factor bound.

bus/src/tests.rs
tcoratger: Replaced prime-subfield-only binary fixtures with random full GF(2^128) values.

bus/src/product.rs
tcoratger: Added a non-panicking field-size precheck and documented the required challenge-field security margin.

bus/src/leaf.rs
tcoratger: Documented physical declaration order and offsets, and made cross-bus domain separation a caller obligation rather than a recommendation.

bus/src/product.rs
tcoratger: Kept dense equality evaluation as an explicit measured follow-up. The current API now documents logical-size work and makes no sparse-time claim; changing this hot algorithm needs its own benchmarked PR.

bus/src/product.rs
tcoratger: Public product points are now returned in the repository's big-endian convention, with a real plan-to-GKR terminal relation test.

bus/src/leaf.rs
tcoratger: Documented that the surrounding AIR must constrain every conditional selector to be Boolean.

bus/src/transcript.rs
tcoratger: The transcript now binds explicit shape bytes, including height, tree count, and root encoding, so zero-height aliases are separated.

bus/src/leaf.rs
tcoratger: Moved every tuple-width validation before the exponential equality-weight allocation.

bus/src/product.rs
tcoratger: Documented the exact ones-padded MLE formula and exported the logarithmic correction helper needed to authenticate a zero-padded committed prefix.

bus/benches/product_gkr.rs
tcoratger: The benchmark now times the production prover directly; the repaired radix-two comparator is restricted to even heights.

bus/src/tests.rs
tcoratger: Added precise malformed-shape, odd-cap, shared-root, and product-preserving unauthenticated-edit tests with error matching.

PR #2217, opened 2026-09-18
feat(word): add checked word relation IR

tcoratger:
## Summary

Add `p3-word`, a backend-neutral `no_std` relation language for dense binary work:

- sealed 32- and 64-bit word types;
- checked public and witness indices;
- checked full-width and 32-bit-lane shifts;
- zero-, one-, and irreducible two-shift terms;
- unnormalized XOR operands;
- zero, bitwise-AND, and full-width unsigned multiplication relations;
- immutable checked systems and an authoritative scalar verifier.

The crate depends only on `alloc` and `thiserror`. It has no AIR, PCS, transcript, frontend, chip, or VM dependency.

## E7 scope

- **7.1:** implemented for 32- and 64-bit words. Lane operations require a 64-bit container.
- **7.2:** relation IR and scalar semantics implemented.
- **7.3–7.4:** proof reductions remain in the stacked backend work.
- **7.5–7.6:** frontend and gadgets remain downstream.
- **7.7:** chip composition remains deferred until the E6 bus binds caller and callee rows.

## Invariants

- Invalid word, lane, shift, and address values are rejected at construction.
- Zero-distance shifts have one representation.
- Two-slot terms reject maps representable by zero or one shift.
- Irreducible pairs may have more than one spelling.
- Exact public and witness lengths are required.
- Integer multiplication checks both limbs of the complete product.

Every ordered shift pair at both widths is classified against an independent bit-linear reference model. Tests live beside the implementation boundary they verify.

## Validation

- `cargo test -p p3-word` (21 passed)
- `cargo clippy -p p3-word --all-targets -- -D warnings`
- `cargo doc -p p3-word --no-deps`
- `git diff --check`


word/src/shift.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Creates a word term with one checked shift.
+    #[inline]
+    pub const fn single(index: ValueIndex, shift: Shift<W>) -> Self {
+        Self {
+            index,
+            shifts: [shift, Shift::identity()],
+        }
+    }
+
+    /// Creates a word term whose two shifts cannot be represented more simply.
```
Nashtare, line 237: `pair` rejects compositions that are zero or a single shift, but an accepted pair's spelling isn't unique. Over every ordered pair, 14,820 of the 74,341 distinct maps accepted on `Word64` have more than one spelling, up to 94 for one map (on `Word32`, 2,448 of 7,068, up to 31). For example, `(srl 11, sll 18)`, `(sra 11, sll 18)` and `(rotr 11, sll 18)` are the same map on both widths, because the three inner shifts differ only in the top 11 bits and `sll 18` discards them. This is sound, but #2210 keys columns by spelling (`key_code(operation, [term.inner(), term.outer()])`), so one mask written two ways gets two keys and duplicate reduction work. Could `pair` canonicalize (e.g. an inner right move whose top bits the outer shift discards becomes `LogicalRight`, plus the symmetric and lane cases), with the exhaustive test extended to assert one spelling per map? If not, could the doc say that accepted pairs are irreducible but not unique?

word/src/shift.rs
```diff
@@ -0,0 +1,357 @@
...
+        self.shifts[0]
+    }
+
+    /// Returns the second shift.
+    #[inline]
+    pub const fn outer(self) -> Shift<W> {
+        self.shifts[1]
+    }
+
+    pub(crate) fn evaluate(self, word: W) -> W {
```
Nashtare: This is the authoritative `outer(inner(x))` order, but it's `pub(crate)`, so #2210 re-spells it as `term.outer().apply(term.inner().apply(word))` in `columns.rs`. That's correct today, but it's exactly the ordering a backend has to agree with the scalar verifier on. Could this be `pub` so backends call the same function?

word/src/shift.rs
```diff
@@ -0,0 +1,357 @@
...
+    Lane32ArithmeticRight,
+    /// Rotates both 32-bit lanes right independently.
+    Lane32RotateRight,
+}
+
+impl ShiftKind {
+    /// Returns whether the operation acts on two independent 32-bit lanes.
+    #[inline]
+    pub const fn is_lane32(self) -> bool {
+        // The lane family is deliberately unavailable to 32-bit words.
```
Nashtare: This comment describes the `Word32` check in `Shift::new` rather than this predicate, so it might read better there.

word/src/constraint.rs
```diff
@@ -0,0 +1,234 @@
...
+pub struct EvaluationError {
+    /// The segment selected by the term.
+    pub segment: Segment,
+    /// The missing position.
+    pub position: u32,
+    /// The available number of words.
+    pub len: usize,
+}
+
+/// An XOR sum of shifted words.
```
Nashtare: `Operand` is an unnormalized XOR list: terms can repeat and cancel (the tests rely on `x ⊕ x = 0`), come in any order, or be absent (so an empty `ZeroConstraint` holds trivially). Could the doc say so, so backends don't assume a canonical term list?

word/tests/relation.rs
```diff
@@ -0,0 +1,486 @@
...
+        for amount in 1..width {
+            shifts.push(Shift::new(kind, amount).expect("enumerated shift is valid"));
+        }
+    }
+    shifts
+}
+
+fn signature32(inner: Shift<Word32>, outer: Shift<Word32>) -> [u32; 32] {
+    // Shifts are F_2-linear maps and are determined by the images of basis bits.
+    core::array::from_fn(|bit| outer.apply(inner.apply(Word32::new(1 << bit))).get())
```
Nashtare: These signatures go through `Shift::apply`, so they are independent of `compose` but not of the evaluator, and `apply` is only compared with `reference_shift*` on randomly sampled (kind, amount) pairs. Could they use the file's own reference helpers instead, e.g. `reference_shift64(outer.kind(), reference_shift64(inner.kind(), 1 << bit, inner.amount().into()), outer.amount().into())`, so the "independently generated" signatures in the description are literally independent? `apply` is correct for every kind and distance, so nothing is wrong today.

word/tests/relation.rs
```diff
@@ -0,0 +1,486 @@
...
+        .expect("the witness term is in range");
+    assert_eq!(
+        zero_system.verify(&[], &[Word32::new(1)]),
+        Err(VerificationError::Unsatisfied {
+            kind: ConstraintKind::Zero,
+            constraint: 0,
+        })
+    );
+
+    // Mutation: flip one bit in the high limb of the product 2^31 * 2.
```
Nashtare: This mutation keeps the low limb correct (`2^31 · 2` has low limb 0) and breaks only the high limb, and every other MUL case is satisfied. So a verifier that compared only `high` would pass the whole suite, even though the code does check both limbs. Could you add the mirror case, with the high limb right and the low limb wrong, e.g. `3 · 5` with `low = 14, high = 0`?

word/tests/relation.rs
```diff
@@ -0,0 +1,486 @@
...
+        .expect_err("position exceeds compact representation");
+
+    assert_eq!(
+        error,
+        p3_word::IndexError::PositionTooLarge {
+            position: u32::MAX as usize + 1,
+        }
+    );
+}
+use std::collections::HashSet;
```
Nashtare: This `use` could move up to the imports at the top of the file.

Nashtare (approved):
I checked the relation semantics against an independent bit-level model, in which each output bit of a shift reads one input bit or zero. Locally I ran it against this head, and everything holds:

- Every shift kind and distance matches the model bit for bit on both widths: the four full-width kinds on `Word32` and `Word64`, and the four lane kinds on `Word64`, at every valid distance, on all basis vectors plus random words. That includes `sra` sign extension per lane and lane rotations. The constructor bounds hold too: all four lane kinds are rejected on `Word32`, and every zero distance normalizes to `Shift::identity()`.
- The two-shift classification is exact over every ordered pair, 142,129 on `Word64` and 15,625 on `Word32`. A pair is `AlwaysZero` exactly when the composed map is zero, `Reducible` exactly when it equals one shift, and otherwise it is accepted and evaluates to the composed map, inner first.
- `verify` agrees with the model on 40,000 random Zero/AND/MUL instances (20,000 per width, about 40% satisfying), including unsigned MUL on both limbs and the length errors.
- None of the checked types can be forged or made to panic. Every field is private, there are no unchecked constructors, `From` impls or serde derives, and `Word` is sealed. `Shift<Word32>` can't hold a lane kind, so the `.expect`s in `Shift::apply` are unreachable.

The inline comments are non-blocking follow-ups. The main one is that accepted pairs are irreducible but their spellings aren't unique, which matters for how #2210 keys its columns.

**Smaller things**
- `release-plz.toml` has no `[[package]]` entry for `p3-word` with `version_group = "plonky3"`. #2003 and #2032 added theirs alongside their crates.
- The `binary` glob in `.github/labeler.yml` doesn't cover `word/**`.
- A few error paths are never exercised: `VerificationError::PublicLength`, `SystemError::SegmentTooLong` (free to test on 64-bit with `u32::MAX as usize + 1`, which allocates nothing), and `IndexOutOfBounds` for a public term or an AND/MUL operand.

word/src/constraint.rs
tcoratger: Documented operands as ordered, unnormalized, multiplicity-preserving XOR lists whose empty form evaluates to zero.

word/src/shift.rs
tcoratger: Made the authoritative ordered evaluation public and replaced separate inner/outer accessors with one ordered shifts value; the backend now calls the shared implementation.

word/tests/relation.rs
tcoratger: Removed the monolithic integration test and colocated its cases with the five implementation modules.

word/src/shift.rs
tcoratger: Documented the deliberate boundary: accepted pairs are irreducible spellings, not canonical representatives of semantic linear maps. Global map normalization would be a different and substantially more expensive IR.

word/tests/relation.rs
tcoratger: The exhaustive signatures now use independent primitive reference helpers rather than the implementation evaluator.

word/tests/relation.rs
tcoratger: Added the mirror low-limb corruption case; both product limbs are now independently pinned.

word/src/shift.rs
tcoratger: Moved the 32-bit restriction explanation to construction and kept the predicate documentation limited to classification.

PR #2221, opened 2026-09-19
ci: test only affected workspace packages

tcoratger:
## Summary

- derive affected crates from Cargo metadata instead of a handwritten dependency map
- test changed crates, their transitive production dependents, and direct dev-test consumers
- scope host, architecture, embedded, wasm, Clippy, rustdoc, and formatting jobs to that set
- skip Rust jobs for known documentation-only changes

Pushes and shared build inputs still run the full workspace. Unknown shared files also fail closed to a full run.

## Selection

| Example change | Before | After |
| --- | ---: | ---: |
| `README.md` | 50 crates | 0 crates |
| `examples/src/lib.rs` | 50 crates | 1 crate |
| `multilinear-util/src/lib.rs` | 50 crates | 8 crates |
| `field/src/field.rs` | 50 crates | 48 crates |

The broad `field` result is intentional: most of the workspace depends on it. Package READMEs remain covered because several crates compile them with `include_str!`.

## Safety

- Cargo remains the source of truth for normal, build, target-specific, optional, and dev dependencies.
- Rename detection considers both the old and new paths.
- Dev edges stop after the direct test consumer, preventing test-helper cycles from selecting unrelated crates.
- Required jobs use job-level conditions, so documentation-only changes report successful skipped checks instead of leaving pending workflow checks.

This follows the graph-query pattern used by large Rust monorepos while keeping the implementation local to the existing CI command script.

## Validation

- `python3 scripts/check.py lint --check scripts`
- `python3 scripts/check.py lint --check fmt`
- scoped nextest and doctest runs for `p3-util,p3-matrix`
- scoped Clippy and rustdoc runs for `p3-util,p3-matrix`
- workflow YAML parse and `git diff --check`


tcoratger:
@Nashtare Let me know what you think about this, related to CI planning, as discuss in https://github.com/Plonky3/Plonky3/pull/2187, I've tried something here :)

.github/workflows/ci.yml
```diff
@@ -126,7 +164,9 @@ jobs:
             target: aarch64-unknown-linux-gnu
             modes: serial parallel
     runs-on: ${{ matrix.os }}
-    if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
+    if: >-
+      needs.plan.outputs.rust == 'true' &&
```
Nashtare: When a job-level `if` skips a matrix job, GitHub skips it before expanding the matrix. It then reports one check under the unexpanded job name rather than one per leg (actions/runner#952, now community discussion #9141, still open).

Branch protection on `main` requires eight of these legs by exact name: `Build and Test (linux)`, `(macos, parallel)`, `(windows, parallel)`, `(linux, avx2)`, `(linux, avx2, parallel)`, `(linux, pclmulqdq)`, `(linux, avx512f, build only)` and `(linux, avx512f+vpclmulqdq, build only)`. For a `README.md`-only change the plan gives `rust=false`. Those eight then stay at "Expected — Waiting for status to be reported", and the PR can't merge without an admin override. The non-matrix jobs are fine, because a skipped job counts as passing.

Could the `rust` gate move onto the steps instead, and keep only `[skip-ci]` here? That includes checkout, the toolchain, the cache and the nextest install. The legs would then expand and pass in seconds. Every step needs the gate, because an empty `--packages ""` falls back to the whole workspace. The other option is an `always()` aggregator job as the single required check. That job would also need to check `needs.plan.result`.

.github/workflows/ci.yml
```diff
@@ -302,52 +378,75 @@ jobs:
...
         run: python3 scripts/check.py lint --check sort
 
       - name: Check TOML formatting
+        if: ${{ needs.plan.outputs.toml == 'true' }}
         run: python3 scripts/check.py lint --check toml
 
       # `--with-metadata` resolves each manifest through cargo.
       # Only that resolution makes dev-dependencies visible to the scan.
       - name: Unused dependencies
+        if: ${{ needs.plan.outputs.manifests == 'true' }}
```
Nashtare: cargo-machete decides whether a dependency is used by scanning the `.rs` sources. So its result can change without any manifest changing.

Locally I deleted the only `use tracing::instrument;` in `sha256-air/src/generation.rs`. The plan selected `p3-sha256-air` with `manifests=false`, so this step would be skipped, but `cargo machete --with-metadata` exits 1. A PR like that goes green here, and then the push run on `main` fails.

Could this step run on `rust || manifests`? It scans the whole workspace in a few seconds.

scripts/check.py
```diff
@@ -53,27 +59,182 @@ def workspace_packages(metadata: dict[str, Any]) -> list[dict[str, Any]]:
...
+        "full": full,
+        "rust": bool(selected),
+        "parallel": any(
+            "parallel" in package.get("features", {}) for package in selected_metadata
+        ),
+        "embedded": bool(embedded),
+        "wasm": "p3-goldilocks" in affected,
+        "keccak": "p3-keccak" in affected,
+        "sha_ni": "p3-sha256" in affected,
+        "gfni": "p3-binary-field" in affected,
```
Nashtare, line 185: `p3-binary-dft` has its own `all(target_feature = "gfni", target_feature = "avx512bw", …)` kernels (`binary-dft/src/lib.rs:15`, `butterfly.rs:307`, `lanes.rs:99`). It depends on `p3-binary-field`, not the other way round. A change to `binary-dft/src/lib.rs` selects 4 packages with `gfni=false`. The `avx512f+avx512bw+gfni` leg then skips "Build and lint", and nothing compiles that code. No required leg enables `avx512bw` or `gfni`.

The other three flags are right today. But all four are a hand-written crate list, which is what the description says this PR replaces.

Would it be simpler to leave "Build and lint" ungated on these legs, since `--packages` already limits it to the affected crates? Only the steps that run `--package X` would keep a gate. If the gate stays, an entry in `package.metadata.plonky3` next to `embedded = false` would let a new gated crate add itself.

scripts/check.py
```diff
@@ -53,27 +59,182 @@ def workspace_packages(metadata: dict[str, Any]) -> list[dict[str, Any]]:
...
+        "toml": force_full or any_toml,
+        "manifests": force_full or any_manifest,
```
Nashtare: These use `force_full`, as `scripts` does on line 171, rather than `full`. `force_full` is only the `--full` flag used on pushes. So a PR that falls back to full CI because it touches `.github/workflows/`, `.cargo/`, `Cargo.lock` or an unknown path still skips `Sort Cargo.toml`, `Check TOML formatting` and `Unused dependencies`.

This PR shows it. Its lint job skipped both `install-action` steps and those three checks, yet the diff rewrites exactly those install steps. They will first run on `main` after merge.

Could `toml`, `manifests`, `scripts` and `lint` use `full` here?

.github/workflows/ci.yml
```diff
@@ -152,47 +192,75 @@ jobs:
...
+        run: >-
+          ${{ runner.os == 'Windows' && 'python' || 'python3' }} scripts/check.py doctest
+          --packages "${{ needs.plan.outputs.packages }}"
 
       - name: Test with parallel
-        if: ${{ !matrix.compile_only && contains(matrix.modes, 'parallel') }}
-        run: ${{ runner.os == 'Windows' && 'python' || 'python3' }} scripts/check.py test --parallel
+        if: >-
+          !matrix.compile_only &&
+          contains(matrix.modes, 'parallel')
```
Nashtare: `feature_args` emits no features when none of the selected crates has a `parallel` feature, for example a PR touching only `p3-bus`, `p3-word`, `p3-word-backend` or `p3-examples`. This step then reruns the serial suite on the three parallel legs. Before, the workspace-wide `--features parallel` unified `p3-maybe-rayon/parallel` into those crates' dependencies. None of the four calls rayon directly, so little coverage is lost. But `Build and lint with parallel` already checks `needs.plan.outputs.parallel`, and this step doesn't. Should it, for consistency?

scripts/check.py
```diff
@@ -30,6 +30,12 @@
     "scripts": [sys.executable, "-m", "unittest", "scripts/test_check.py", "-v"],
 }
 
+DOC_ONLY_NAMES = {"CHANGELOG.md", "LICENSE-APACHE", "LICENSE-MIT"}
+DOC_ONLY_PATHS = {"CONTRIBUTING.md", "README.md", "RELEASING.md"}
+DOC_ONLY_PREFIXES = ("docs/", ".github/ISSUE_TEMPLATE/")
```
Nashtare: `audits/` only holds the audit PDFs, so it could go here. Today, adding a report triggers a full run.

Nashtare (changes requested):
I ran the planner against the real workspace metadata, and the selection logic holds up. The `-z` rename parsing, the deepest-root owner lookup, the transitive production closure with a single dev hop, and the fail-closed fallback for unknown paths all behave as described. The four rows of the table reproduce (0 / 1 / 8 / 48). The repo has no symlinks, and no `include!`, `include_str!` or `#[path]` crosses a package root. Nothing includes the root README, `docs/architecture.md` or any CHANGELOG, so the doc-only list is safe today. `scripts/test_check.py` passes (32 tests).

One thing to keep in mind about the green checks: this PR touches `.github/workflows/` and `scripts/`, which are full-CI prefixes. The plan job printed `"full": true` with all 50 packages, so every leg ran the whole workspace, and the timings are in line with recent pushes to `main`. That is the right behaviour, but it means neither the scoped path nor the doc-only path has run in CI yet.

- A doc-only PR can't go green. Skipping the matrix job at job level means the 8 required `Build and Test (…)` checks are never reported (inline, `ci.yml`).
- `Unused dependencies` only runs when a `Cargo.toml` changes, but machete scans sources. A PR that removes the last use of a dependency passes here and then fails the next push to `main` (inline, `ci.yml`).
- A full run triggered by the changed paths still skips sort, taplo and machete, and the `gfni` flag misses `p3-binary-dft` (inline, `check.py`).

**Smaller things**
- If `Plan CI` fails, `Formatting and Clippy`, `Build embedded` and `Test wasm32 simd128` are skipped, and skipped counts as passing. Only the stuck matrix legs keep the PR blocked. Adding `Plan CI` to the required checks would make that explicit.
- `CiPlanTests` covers package selection only. The job flags could be pinned with a few cases against the real workspace metadata, as `test_repository_metadata_explicitly_excludes_host_only_packages` does. For example: `binary-dft/src/lib.rs` should set `gfni`, and `.github/workflows/ci.yml` should set `manifests` and `toml`.
- Pushes to `main` still run the full suite, but a periodic full run would also catch toolchain drift, runner-image changes and flaky tests between merges. A manual trigger would let a maintainer run the full suite on a PR branch before merging. The `plan` job already falls back to `--full` for every event other than `pull_request`, so adding `schedule:` (nightly) and `workflow_dispatch:` to `on:` would give both with no other change. The concurrency group would need `github.event_name` in it, so that a push to `main` doesn't cancel the nightly run.

Nashtare (approved):
The fix commit closes the earlier threads:
- The matrix job now gates every step instead of the job. I parsed the workflow at this head, and each of the 16 steps checks `rust`, so a doc-only PR still reports all 14 legs. A failing `Plan CI` still skips the job before expansion, so the PR stays blocked.
- "Build and lint" is no longer tied to a crate list, so `p3-binary-dft`'s GFNI kernels are compiled again.
- Machete runs on any Rust change.
- A full run triggered by the changed paths now covers sort, taplo and machete. The lint job on this head ran all three.
- `audits/` is doc-only. The script tests pass.

I also checked what happens when a crate is added. Adding it to `members` edits the root `Cargo.toml`, so that PR runs the full suite. After that, the planner picks the crate up from `cargo metadata`, with no list to maintain. Locally I added a crate that depends on `p3-util` and is used by `p3-examples`:
- A later change to its sources alone selected it and `p3-examples`.
- A change to `p3-util` selected it along with the other 48 packages that depend on `p3-util`.

**Smaller things**
- One gap comes from my own earlier suggestion on the parallel steps (inline).
- The Safety section of the description still says required jobs use job-level conditions.
- `Cargo.lock` is gitignored, so its entry in `FULL_CI_PATHS` never matches, and external dependencies resolve fresh on every run. A PR-scoped run won't see an upstream release that breaks a crate it doesn't select; only the next push to `main` will. A nightly `schedule:` plus `workflow_dispatch:` run, as suggested earlier, would catch that between merges.

.github/workflows/ci.yml
Nashtare: The new condition is consistent, but my suggestion went too far. `macos, parallel`, `windows, parallel`, `linux, parallel` and `linux, avx2, parallel` run in `parallel` mode only. So when the selection has no `parallel` feature, those legs now run no tests at all.

The planner gives `parallel=false` for a PR touching only `p3-bus`, `p3-examples`, `p3-word` or `p3-word-backend`. Such a PR gets no macOS or Windows coverage before merge.

Could both parallel test steps use `(needs.plan.outputs.parallel == 'true' || !contains(matrix.modes, 'serial'))`? The parallel-only legs would then fall back to the plain run, and `linux, pclmulqdq`, which already runs the serial pass, would still skip the duplicate.

Issue #2169, opened 2026-09-17
perf(binary-dft): subfield-twiddle butterfly on AArch64


tcoratger:
`binary-dft`'s subfield-twiddle butterfly (#2160) is gated on `gfni`, `avx512f` and `avx512bw`. AArch64 gets none of it and falls back to the packed product, so the 10.7x measured on x86 is unavailable on Apple silicon.

## What the x86 kernel does

Every stage of an unshifted additive transform up to height `2^32` has a twiddle four bytes wide or narrower. Scaling by an element of a byte-aligned tower subfield acts identically on each group of `D` bytes, which is a matrix of 8x8 `GF(2)` blocks — and `vgf2p8affineqb` applies one such block to 64 bytes in a single instruction. See `binary-dft/src/affine.rs` and `binary-dft/src/lanes.rs`.

## Why ARM cannot copy it

There is no byte-wise affine instruction on AArch64. The structural insight — that the twiddle lives in a small subfield, and that scaling by it is one fixed `GF(2)`-linear map per byte group — is architecture-neutral and still worth exploiting. Only the instruction that applies the map has to change. Two routes worth measuring:

- **Table lookup.** `TBL`/`TBX` over a 16-byte index vector can evaluate a fixed byte-to-byte map on 16 bytes at a time. A one-byte subfield scaling is exactly such a map, so the `D = 1` case is a direct fit; wider groups need the map decomposed into nibble tables.
- **`vmull_p8`.** An 8-bit polynomial multiply plus a fixed reduction gives the subfield product directly, without going through a matrix at all.

Which wins is a measurement, and the two may split by group width.

## A cheaper first step

`binary-dft` already has a typed one-byte subfield product on the portable path. The follow-up noted in #2160 is to add the typed **two- and four-byte** products there as well: roughly twenty lines across two macros, worth an estimated 3-10x on every non-GFNI target including all of AArch64, and blocked today only by the byte level having no multiplication against the sixteen-bit level. That lands value on ARM without any intrinsics at all, and is the sensible thing to do first.

## What done looks like

- Every specialised path keeps the portable kernel as its reference, with a proptest pinning them equal across twiddle widths, both directions, and lengths that are not a whole number of registers — **executed on real hardware**.
- `binary-dft/benches/ntt.rs` reported on Apple silicon, before and after, including the `narrow` and `short` arms added in #2160 which is where a kernel that covers nothing shows up.

## Why this needs a Mac

[...]

PR #16, opened 2023-05-29
Fix `cargo test`

matthiasgoergens:
`cargo test` runs some tests that `cargo test --all-targets` does not run.  See https://github.com/rust-lang/cargo/issues/6669 tracks this issues.  Specifically, doc tests.

This PR adds running the latter to our CI/CD, to ensure that a simple `cargo test` will work.

This PR also marks parts of the documentation that look like tests to cargo, but actually aren't meant to be tests.

PR #2222, opened 2026-09-19
perf(binary-dft): one-byte subfield butterfly on NEON

oksanaphmn:
Refs #2169. This adds a NEON kernel for the butterfly when the twiddle lies in the one-byte subfield GF(2^8), the case where hardware carryless multiply leaves AArch64 with no fast path. Scaling by such a twiddle is GF(2)-linear on every coordinate byte, so it splits over the two nibbles of a byte and each half becomes a single `vqtbl1q_u8` lookup; the two 16-entry tables are built once per call from eight products computed by the field itself, so no field law is restated. Runs under 256 bytes and wider twiddles (t16, t32, t128) keep the existing path, and the kernel is checked against the element-at-a-time reference, including new tests for lengths around the register boundary that the old sweeps never reached. On an Apple M1 the t8 butterfly is about 11x faster at every level (e.g. 3.39 µs to 0.31 µs at 64 bits), and end-to-end `ntt` at 2^16 x 16 improves by 20% (32-bit), 29% (64-bit) and 33% (128/tower), with the `narrow` arms gaining 18-29% and the `short` arms, t16/t32/t128 and the hybrid/ghash routes unchanged within noise. For context, 128/tower is still slower than 128/hybrid on this machine, so the gain matters most where data has to stay in the tower basis.

Benchmarks on Apple M1 (MacBook Air), main vs this branch, `cargo bench -p p3-binary-dft --bench ntt`.

`butterfly`, one-byte twiddle (`t8`), 1024 elements per side:

| level | main | branch | speedup |
|---|---|---|---|
| 16 | 1.016 µs | 89.0 ns | 11.4x |
| 32 | 1.675 µs | 157.8 ns | 10.6x |
| 64 | 3.394 µs | 305.1 ns | 11.1x |
| 128/tower | 6.412 µs | 598.9 ns | 10.7x |

`ntt`, 2^16 rows x 16 columns:

| arm | main | branch | change |
|---|---|---|---|
| 32 | 26.3 ms | 20.9 ms | -20% |
| 64 | 36.7 ms | 26.1 ms | -29% |
| 128/tower | 63.9 ms | 43.3 ms | -33% |
| 32/narrow | 1.79 ms | 1.47 ms | -18% |
| 64/narrow | 2.46 ms | 1.82 ms | -26% |
| 128/tower/narrow | 4.36 ms | 3.14 ms | -29% |
| 128/hybrid | 20.0 ms | 19.7 ms | within noise |
| 128/ghash | 13.1 ms | 13.1 ms | within noise |
| 128/tower/shifted | 89.6 ms | 89.7 ms | within noise |

`t16`, `t32`, `t128` and the `short` arms of `butterfly` are unchanged within noise.
Measured on a fanless laptop with 10 samples for `ntt`, so expect some run-to-run variation.

---

Part of #2243.


binary-dft/src/lib.rs
```diff
@@ -34,6 +34,9 @@ mod interleaved;
 mod lanes;
 mod lch;
 mod naive;
+// One-byte butterfly kernel for AArch64.
+#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
+mod neon;
```
tcoratger: **[MINOR] `neon` is the only kernel module not compiled on every target, which costs it the crate's own scalar stand-in**

`affine` (`lib.rs:11-20`) and `lanes` (`lib.rs:25-34`) are gated `#[cfg(any(test, all(target_arch = "x86_64", ...)))]`, so they build and run on every host against `test_util::Model`, the scalar `ByteLanes` implementation (`binary-dft/src/test_util.rs:16,34`). `neon` gets no `test` arm, so on x86_64 it is not even type-checked: a future refactor of `butterfly.rs` that breaks `neon.rs` is invisible to every non-Apple developer and to every Linux CI leg.

The asymmetry is not forced. `nibble_tables` contains no intrinsics at all — it is pure `[u8; 16]` arithmetic — and is the only piece of *new field math* in the PR. Only `byte_butterfly` needs the `aarch64` gate.

**Fix:** either gate the module `#[cfg(any(test, all(target_arch = "aarch64", target_feature = "neon")))]` and put `byte_butterfly` alone behind the narrow `cfg` (matching `affine`/`lanes`), or move `nibble_tables` into `butterfly.rs` next to the `columns` construction it pairs with.
**Verify:** the free test this unlocks is exhaustive and instant — I ran exactly it offline and it passes:
```rust
#[test]
fn the_nibble_tables_are_the_field_product() {
    for t in 0..=u8::MAX {
        let t = BinaryField8::from_repr(t);
        let columns = core::array::from_fn(|b| (t * BinaryField8::from_repr(1 << b)).to_repr());
        let (low, high) = nibble_tables(&columns);
        for x in 0..=u8::MAX {
            let got = low[(x & 0x0f) as usize] ^ high[(x >> 4) as usize];
            assert_eq!(got, (t * BinaryField8::from_repr(x)).to_repr(), "t={t:?} x={x:#04x}");
        }
    }
}
```

binary-dft/src/neon.rs
```diff
@@ -0,0 +1,76 @@
...
+                let b = veorq_u8(b, a);
+                vst1q_u8(hi.as_mut_ptr(), b);
+                vst1q_u8(lo.as_mut_ptr(), veorq_u8(a, map(b)));
+            } else {
+                let a = veorq_u8(a, map(b));
+                vst1q_u8(lo.as_mut_ptr(), a);
+                vst1q_u8(hi.as_mut_ptr(), veorq_u8(b, a));
+            }
+        }
+    }
```
tcoratger: **[MINOR] This loop is a verbatim second copy of `lanes::butterfly_run`, and the two must stay bit-identical forever**

Lines 60-73 are line-for-line the body of `lanes::butterfly_run` (`binary-dft/src/lanes.rs:72-90`), down to the comment-free ordering of the `INVERSE` arms:

```text
forward:  a' = a ^ map(b);  store lo = a';  store hi = b ^ a'
inverse:  b' = b ^ a;       store hi = b';  store lo = a ^ map(b')
```

Those four lines are the entire correctness of the butterfly. Having two copies means a future fix to one silently does not reach the other, and the copy that is only compiled on `aarch64` is the one nobody will notice. It also forfeits `Model`-backed testing of the arms.

`butterfly_run` is already generic over the map and only uses `BYTES`, `load`, `store`, `xor` from `ByteLanes` — the `rotate_group`/`affine`/`affine_merge` members it does *not* use are the GFNI-specific ones `uint8x16_t` genuinely cannot provide.

**Fix:** split `ByteLanes` into a minimal supertrait (`BYTES` + `load` + `store` + `xor`) that `butterfly_run` is generic over, plus the affine half that stays x86-only; then implement the minimal half for `uint8x16_t` and call `butterfly_run::<uint8x16_t, INVERSE>(lo, hi, map)`. `neon.rs` then holds only `nibble_tables`, the three-intrinsic `map` closure, and the `ByteLanes` impl.
**Verify:** after the refactor `lanes.rs`'s existing `Model` tests cover the AArch64 arms too, on every host.

binary-dft/src/neon.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// # Panics
+/// Panics if the runs differ in length.
+#[inline]
+pub(crate) fn byte_butterfly<const INVERSE: bool>(
+    lo: &mut [u8],
+    hi: &mut [u8],
+    columns: &[u8; 8],
+) -> usize {
+    assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
```
tcoratger: **[MINOR] This assert can never fire, and the real precondition lives elsewhere**

The only caller builds both slices with the *same* length, taken from `lo` alone (`binary-dft/src/butterfly.rs:390,398-399`):

```rust
let bytes = size_of_val(lo);
...
core::slice::from_raw_parts_mut(hi.as_mut_ptr().cast::<u8>(), bytes),
```

So by the time control reaches line 40, `lo.len() == hi.len()` holds by construction. If the two `[F]` runs had actually differed in length, `from_raw_parts_mut` would already have fabricated an out-of-bounds `&mut [u8]` over `hi` — undefined behaviour one frame *before* this check. The assert therefore reads as a safety net that is not one.

The genuine guarantee is `assert_eq!(lo.len(), hi.len())` in `coordinate_butterfly` (`binary-dft/src/butterfly.rs:447`), which runs before `subfield_prefix` is called. That is sound today; it is just undocumented at the point where it becomes a memory-safety obligation.

**Fix:** keep the assert (it is free and documents intent) but extend the `SAFETY` comment at `butterfly.rs:395` to name the precondition, e.g. *"`coordinate_butterfly` has already asserted `lo.len() == hi.len()`, so the shared `bytes` length is in bounds for both."* Same wording would improve the x86 twin at `butterfly.rs:340`.
**Verify:** `a_length_mismatch_is_refused` (`butterfly.rs:722-729`) already pins the upstream assert; nothing new to add.

binary-dft/src/neon.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use core::arch::aarch64::{
+    uint8x16_t, vandq_u8, vdupq_n_u8, veorq_u8, vld1q_u8, vqtbl1q_u8, vshrq_n_u8, vst1q_u8,
+};
+
+/// Bytes in one register.
+pub(crate) const REGISTER_BYTES: usize = 16;
+
+/// Shortest run, in bytes, worth building the tables for. To be tuned with the benchmarks.
+pub(crate) const MIN_BYTES: usize = 256;
```
tcoratger: **[MINOR] A shipped `TODO`, and a threshold 16x wider than the x86 one**

The doc comment says *"To be tuned with the benchmarks"*, but the constant is being merged untuned. More concretely: the x86 kernel's cutoff is exactly one register — `bytes < <__m512i as ByteLanes>::BYTES`, i.e. 64 (`binary-dft/src/butterfly.rs:334-336`) — while this one is 16 registers.

That gap has a real consequence for the narrow stages. In `stage_pass` a butterfly side is `half = (1 << j) * width` elements (`binary-dft/src/lch.rs:167`), so the kernel only fires once `(1 << j) * width * size_of::<F>() >= 256`. For a width-1 `BinaryField32` transform that is `j >= 6`; stages `j = 4, 5` still have 4 and 8 whole 16-byte registers of work each and fall back to the scalar chain. On x86 those same stages vectorize.

The setup being amortized is 8 `GF(2^8)` products plus ~30 XORs plus two `vld1q_u8` — plausibly ~20-40 cycles, which pays for itself well before 16 registers.

**Fix:** either drop the threshold toward `4 * REGISTER_BYTES` (or `REGISTER_BYTES` outright) after measuring, or, if 256 really is the measured crossover on M1, say so in the comment and delete "to be tuned" — the number then needs a reason, because the next reader will otherwise assume it is a placeholder.
**Verify:** the `ntt` `narrow` arms in `binary-dft/benches/ntt.rs` are exactly the shape that exposes this; sweep `MIN_BYTES` over `{16, 64, 128, 256}` on `32/narrow` and `64/narrow` and report the crossover.

binary-dft/src/neon.rs
```diff
@@ -0,0 +1,76 @@
...
+    lo: &mut [u8],
+    hi: &mut [u8],
+    columns: &[u8; 8],
+) -> usize {
+    assert_eq!(lo.len(), hi.len(), "butterfly lengths differ");
+
+    let (low, high) = nibble_tables(columns);
+    let (lo, _) = lo.as_chunks_mut::<REGISTER_BYTES>();
+    let (hi, _) = hi.as_chunks_mut::<REGISTER_BYTES>();
+    let covered = lo.len() * REGISTER_BYTES;
```
tcoratger: **[MINOR, perf] One register per iteration leaves loop overhead on the table**

I cross-compiled a forced monomorphization of `BinaryField128::butterfly::<false>` for `aarch64-unknown-linux-gnu` at `-C opt-level=3`. LLVM does not unroll this; the steady-state body is exactly 11 instructions for 16 bytes:

```asm
.LBB3_19:
	ldr  q3, [x21]                 // b = hi
	ldr  q6, [x20]                 // a = lo
	subs x19, x19, #1
	ushr v4.16b, v3.16b, #4
	and  v5.16b, v3.16b, v2.16b
	tbl  v4.16b, { v1.16b }, v4.16b
	tbl  v5.16b, { v0.16b }, v5.16b
	eor  v4.16b, v6.16b, v4.16b    // a' = a ^ map(b)
	eor  v4.16b, v5.16b, v4.16b
	eor  v3.16b, v4.16b, v3.16b    // hi' = b ^ a'
	str  q4, [x20], #16
	str  q3, [x21], #16
	b.ne .LBB3_19
```

Good news: `low`/`high`/`nibble` are correctly hoisted into `v0`/`v1`/`v2` outside the loop, so the table build is not repeated. But `subs`/`b.ne` is 2 of every 13 slots, and the two `tbl`s sit on one dependency chain with nothing to interleave against.

**Fix:** `as_chunks_mut::<32>()` (or `::<64>`) and process 2-4 registers per iteration with `vld1q_u8_x2` / `vst1q_u8_x2`, keeping the current 16-byte loop as the sub-chunk remainder. Independent chains then fill M1's NEON pipes. Note `vld1q_u8_x2` has the same 1-byte alignment requirement as `vld1q_u8`, so nothing changes in the safety argument.
**Verify:** `cargo bench -p p3-binary-dft --bench ntt -- butterfly/.*t8` on the M1 — the `t8` arms should move again if this is load/issue bound rather than TBL-throughput bound.

binary-dft/src/neon.rs
```diff
@@ -0,0 +1,76 @@
...
+    let mut low = [0u8; 16];
+    let mut high = [0u8; 16];
+    for b in 0..4 {
+        for n in 0..(1usize << b) {
+            low[n | (1 << b)] = low[n] ^ columns[b];
+            high[n | (1 << b)] = high[n] ^ columns[4 + b];
+        }
+    }
+    (low, high)
+}
```
tcoratger: **[NIT] The table construction is correct — recording the proof so the next reader does not have to redo it**

`columns[b] = t * 2^b` are the images of the `GF(2)` basis, `low[n]` accumulates `XOR_{b in n} columns[b] = t * n` for `n < 16`, and `high[n] = XOR_{b in n} columns[4 + b] = t * (n << 4)`. Since `x = (x & 0xf) + ((x >> 4) << 4)` and multiplication by `t` is `GF(2)`-linear, `low[x & 0xf] ^ high[x >> 4] = t * x`. This is the standard `SPLIT_TABLE(8, 4)` decomposition.

I checked it exhaustively offline against `tower_mul` from `binary-dft/src/affine.rs:47` (itself proptested against `BinaryField8` at `affine.rs:241-257`): 256 x 256 table lookups, zero mismatches. Both nibble indices max out at 15, so `vqtbl1q_u8`'s out-of-range-returns-zero behaviour is never relied on — worth one line of comment, since a reader auditing `vqtbl1q_u8` will ask.

Edge cases also hold: `t = 0` gives all-zero tables (though `coordinate_butterfly` short-circuits zero twiddles at `binary-dft/src/butterfly.rs:450-452` and never reaches here), and `t = 1` gives `low[n] = n`, `high[n] = n << 4`, i.e. the identity map.

**Fix:** add `// Both indices are nibbles, so TBL's out-of-range zeroing is never reached.` next to the `vandq_u8`/`vshrq_n_u8` pair on lines 55-56.
**Verify:** the exhaustive test sketched in my comment on `lib.rs`.

binary-dft/src/butterfly.rs
```diff
@@ -366,12 +366,57 @@ fn subfield_prefix<F: ByteCoordinates, const INVERSE: bool>(
     covered / size_of::<F>()
 }
 
+/// The one-byte subfield butterfly on NEON. Returns the elements covered.
+#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
```
tcoratger: **[NIT] `aarch64_be` now fails to compile rather than falling back**

`target_arch = "aarch64"` is also true for `aarch64_be-*`, where `assert!(cfg!(target_endian = "little"))` in the `const` block below (line 381) turns into a monomorphization-time error. Before this PR, `p3-binary-dft` built there via the `const fn` stub.

This matches existing repo practice (`binary-field/src/clmul/aarch64.rs:161,191` does the same under `all(target_arch = "aarch64", target_feature = "aes")`), and big-endian AArch64 is tier 3, so this is genuinely a nit — but a hard build break is a worse failure mode than a fallback, and the fix is two words.

**Fix:** add `target_endian = "little"` to this `cfg`, to the `mod neon` gate in `lib.rs:38`, **and** to the matching arm of the `#[cfg(not(any(...)))]` fallback at line 419 — all three, or `aarch64_be` ends up with no `subfield_prefix` at all.
**Verify:** `cargo check -p p3-binary-dft --target aarch64_be-unknown-linux-gnu -Zbuild-std` (nightly), or just reason it through — nothing in CI covers it.

binary-dft/src/butterfly.rs
```diff
@@ -366,12 +366,57 @@ fn subfield_prefix<F: ByteCoordinates, const INVERSE: bool>(
...
+        return 0;
+    }
+
+    // SAFETY: `ByteCoordinates` guarantees each element is exactly its bytes.
+    let (lo, hi) = unsafe {
+        (
+            core::slice::from_raw_parts_mut(lo.as_mut_ptr().cast::<u8>(), bytes),
+            core::slice::from_raw_parts_mut(hi.as_mut_ptr().cast::<u8>(), bytes),
+        )
+    };
```
tcoratger: **[NIT] The `unsafe` reasoning is sound; the comment states only half of it**

For the record, the four obligations I checked and that do hold:

1. *Provenance/aliasing* — `lo` and `hi` reach `F::butterfly` from `block.split_at_mut(half)` (`binary-dft/src/lch.rs:180,236`), so the two `&mut [F]` are disjoint; the derived raw pointers inherit that, and the shadowed originals are never touched again.
2. *No padding* — `size_of::<F>() == 1 << (F::LOG_BITS - 3)` (line 382) rules it out, so `size_of_val(lo) == len * size_of::<F>()` covers exactly the initialized bytes.
3. *Write validity* — the `ByteCoordinates` contract (`butterfly.rs:30-39`) says every bit pattern of the backing integer is a field element, so arbitrary byte writes cannot forge an invalid `F`. `Ghash128` is correctly excluded (different basis) and routes to `plain_butterfly`.
4. *Alignment* — `vld1q_u8`/`vst1q_u8` lower to `LD1`/`ST1 {v.16b}`, which require 1-byte alignment only, so `u8` chunk pointers are fine.

What the comment omits is the one obligation that is *not* local to this function: `hi` is given `lo`'s length (line 399), so `lo.len() == hi.len()` is a memory-safety precondition here, discharged upstream at line 447. See my note on `neon.rs:40`.

**Fix:** extend the `SAFETY` comment to cover (1) disjointness of the two runs and (2) the length equality asserted by `coordinate_butterfly`.
**Verify:** nothing to run — this is documentation of an invariant the code already upholds.

binary-dft/src/butterfly.rs
```diff
@@ -366,12 +366,57 @@ fn subfield_prefix<F: ByteCoordinates, const INVERSE: bool>(
...
+
+    // The map is linear, so its products with the basis elements determine it.
+    let t = BinaryField8::from_repr(t);
+    let columns: [u8; 8] =
+        core::array::from_fn(|b| (t * BinaryField8::from_repr(1 << b)).to_repr());
+
+    neon::byte_butterfly::<INVERSE>(lo, hi, &columns) / size_of::<F>()
```
tcoratger, line 429: **[NIT] Code size grows ~28% per butterfly monomorphization — worth knowing, not worth blocking**

I cross-compiled the same forced monomorphizations against `main` and against this branch at `-C opt-level=3` for `aarch64-unknown-linux-gnu` and counted emitted instructions per symbol:

| symbol | main | branch |
|---|---|---|
| `BinaryField8::butterfly::<false>` | 239 | 320 |
| `BinaryField16::butterfly::<false>` | 1203 | 1557 |
| `BinaryField32::butterfly::<false>` | 940 | 1184 |
| `BinaryField64::butterfly::<false>` | 670 | 830 |
| `BinaryField128::butterfly::<false>` | 1769 | 2269 |

`main` emits **zero** `tbl` instructions for any of these, which settles the "does it autovectorize anyway" question: it does not, so the intrinsics are buying something real rather than restating what LLVM would find.

The whole chain is `#[inline]` into `stage_pass`/`tile_stages`, so this lands in the hot loop's I-cache footprint. The gating is cheap and correctly placed — the emitted `cmp x19, #16 / b.hs` (the `MIN_BYTES` test for a 16-byte level) sits before any table work, so short runs pay one compare, exactly as `MIN_BYTES`'s doc intends.

I flag it only because the `narrow` arms execute a great many *short* butterflies where the added code is pure footprint. The benchmark numbers in the description say it nets out positive, so this is informational.

**Fix:** none required.
**Verify:** already measured above.

tcoratger (approved):
Adds a `vqtbl1q_u8` nibble-table kernel for the one-byte-subfield butterfly on AArch64, where the GFNI `vgf2p8affineqb` route has no equivalent. I verified the new field math is **bit-exact**: I re-implemented `nibble_tables` and both butterfly arms as scalar code and checked them exhaustively against `tower_mul` (the same tower `GF(2^8)` product `affine.rs` is proptested against) over all 256 twiddles x 256 bytes for the tables, and all `256^3` `(t, u, v)` triples for both directions — zero mismatches, and both TBL indices provably stay in `0..=15` so `vqtbl1q_u8`'s out-of-range zeroing is never reached. Tail handling, the `MIN_BYTES` gate, the `covered / size_of::<F>()` conversion and the `cfg` gating are all correct, and I confirmed on cross-compiled AArch64 codegen that the intended 11-instruction TBL loop is emitted while `main` emits zero `tbl` (so this does not autovectorize on its own). No BLOCKER and no MAJOR: everything below is MINOR or NIT.

**[NIT] CI has not actually run on this head.** `gh pr checks 2222` reports only `label: success` for `02166093`; no `Build and Test` leg has executed. Worth a maintainer re-trigger before merge.

**[NIT] The differential coverage is real, and does execute.** Unlike GFNI (`.github/workflows/ci.yml:99-113`, all `compile_only: true`), NEON is baseline on `macos-latest`, which is arm64 in this workflow (see the `Test keccak (aarch64, no SHA3)` step at `.github/workflows/ci.yml:190-195`) and runs `scripts/check.py test --parallel` over the whole workspace. So `one_byte_twiddles_agree_around_every_register_boundary` (`binary-dft/src/butterfly.rs:683-696`) and `byte_twiddles_agree_on_runs_past_the_register_threshold` (`binary-dft/src/butterfly.rs:752-763`) really do pin this kernel against the element-at-a-time reference on hardware. That is the right test and it is present — please keep the `macos-latest` leg a required check, because it is now the *only* thing executing this code.

**[NIT] Not stale.** Merge base `fed938d` is one day old, `git merge-tree main refs/pr/2222` is clean, and no commit on `main` has touched `binary-dft/` since. Nothing to rebase around.

**What I could not do on this host:** x86_64 only, no `qemu-aarch64` and no `aarch64-linux-gnu-gcc`, so I could not *run* the NEON path or reproduce the M1 timings. `cargo check`/`cargo build --target aarch64-unknown-linux-gnu -p p3-binary-dft` both succeed (the `build` forces codegen, so the `const {}` asserts at `butterfly.rs:380-384` are genuinely evaluated for every reachable `F`); `--all-targets` cannot be cross-built here because a proptest dependency (`alloca`) needs a cross C compiler, so the new tests were only compiled and run on x86_64, where they exercise the scalar path.

**Bottom line:** merge — the kernel is correct and the `unsafe` is sound; the notes below are polish and future-proofing, not blockers.

binary-dft/src/neon.rs
tcoratger: Skipped, and left unresolved.

I am on x86_64 with no Apple silicon or AArch64 host to run on, so I cannot show that unrolling helps — and the review itself frames this as load/issue bound *or* TBL-throughput bound, which only a measurement on the hardware can settle. Widening the loop without that measurement would be a guess at a kernel that is currently verified bit-exact.

Worth noting the cost is now lower than when you wrote this: the loop body has moved into the shared `butterfly_run`, so unrolling it would benefit the GFNI kernel at the same time. That makes it a better standalone follow-up than a change to fold in here.

binary-dft/src/butterfly.rs
tcoratger: No action needed, as you said.

For the record, I re-ran your measurement after deduplicating the loop (all seven forced monomorphizations, `-C opt-level=3`, `aarch64-unknown-linux-gnu`): the emitted assembly is **byte-identical** to this branch before my commit, steady-state loop included. So the code-size figures in your table still stand exactly as measured.

binary-dft/src/neon.rs
tcoratger: Done, exactly as you suggested — and the codegen is free.

`ByteLanes` is split into `ByteRegister` (`BYTES`, `load`, `store`, `xor`) and `ByteLanes: ByteRegister` (`rotate_group`, `affine`, `affine_merge`, still x86-gated). `butterfly_run` is now generic over `ByteRegister`, `uint8x16_t` implements it next to the `__m512i` backend in `lanes.rs`, and `neon.rs` holds only the table construction plus the three-intrinsic map closure. The second copy of the four arms is gone.

Verified by disassembly, as asked. Forced monomorphizations of `BinaryField{8,16,32,64,128}::butterfly::<false>` and `BinaryField{8,128}::butterfly::<true>`, cross-compiled to `aarch64-unknown-linux-gnu` at `-C opt-level=3`, before and after:

| symbol | before | after |
|---|---|---|
| `f8::butterfly::<false>`   | 257 | 257 |
| `f16::butterfly::<false>`  | 330 | 330 |
| `f32::butterfly::<false>`  | 311 | 311 |
| `f64::butterfly::<false>`  | 456 | 456 |
| `f128::butterfly::<false>` | 305 | 305 |
| `f8::butterfly::<true>`    | 170 | 170 |
| `f128::butterfly::<true>`  | 301 | 301 |

Not just equal counts — `diff` reports the extracted assembly byte-identical for all seven, steady-state loop included. `chunks_exact_mut(L::BYTES)` const-folds to the same thing `as_chunks_mut::<16>()` did once `BYTES` is known.

`butterfly_run` also carries the `lo.len() == hi.len()` assert, so that is shared now too.

binary-dft/src/lib.rs
tcoratger: Done.

- `mod neon` is now gated `any(test, all(target_arch = "aarch64", target_feature = "neon", target_endian = "little"))`, matching `affine` and `lanes`. Only `byte_butterfly` and the length threshold keep the narrow arch gate.
- `mod lanes` gained the AArch64 arm, since the shared butterfly loop now backs both kernels. `group_mask` and the `ByteLanes` half stay on the x86 gate so nothing is dead on AArch64.
- Your exhaustive test is in, essentially verbatim, as `neon::tests::the_nibble_tables_are_the_field_product`. It runs on x86_64 and passes.

I checked it is not vacuous: flipping `columns[4 + b]` to `columns[b]` in the high-table accumulation makes it fail immediately at `t=1 x=0x10`.

Bonus from the trait split in the sibling thread: the four butterfly arms are now the single `butterfly_run`, which the existing `Model`-backed tests in `affine.rs` already cover — so the AArch64 arms are exercised on every host too.

binary-dft/src/neon.rs
tcoratger: Done, with one structural change.

The assert now lives in the shared `butterfly_run` (same message), since the loop was deduplicated in the sibling thread — so it still runs, and it is no longer duplicated either.

The `SAFETY` comment on the NEON `subfield_prefix` is extended to state both non-local obligations: that the two runs are halves of one split slice and so never overlap, and that their lengths were asserted equal before that point, which is what makes one shared byte count in bounds for both. The x86 twin got the same two lines.

binary-dft/src/neon.rs
tcoratger: Left at 256, with the comment rewritten to say plainly that it is unmeasured.

I am on x86_64 with no AArch64 host, so I could not run your sweep, and I was not going to invent a number I cannot justify. `to be tuned` is gone; the doc now records what the value actually is (sixteen registers, picked as a round number), that no benchmark stands behind it, what it amortises (eight `GF(2^8)` products, ~30 XORs, two loads), that this should be repaid far below sixteen registers so the number is probably high, and that replacing it means sweeping 16/64/128/256 on AArch64 against the narrow transforms.

So the next reader gets the reason and the recipe rather than a placeholder. If you can run the sweep on the M1, the constant is a one-line change.

binary-dft/src/neon.rs
tcoratger: Done, and thank you for the proof — I folded it into the docs rather than leaving it in the thread.

The table-construction doc now carries the derivation as a `# Overview` with the split laid out as a diagram, states that both indices are nibbles so neither table is ever read out of range, and repeats that line next to the two lookups themselves. The exhaustive check you sketched is now a test that runs on every host.

binary-dft/src/butterfly.rs
tcoratger: Done, all three sites.

`target_endian = "little"` added to the NEON `subfield_prefix` gate, to the module gate, and to the matching arm of the fallback — so big-endian AArch64 goes back to the stub instead of hitting a monomorphization-time error.

The now-redundant `assert!(cfg!(target_endian = "little"))` in that const block is dropped, since the gate enforces it; the comment above says so.

binary-dft/src/butterfly.rs
tcoratger: Done — thank you for writing out the four obligations, it made the comment easy to complete.

Both `SAFETY` blocks (NEON and the GFNI twin) now add the two non-local points: the runs are halves of one split slice so no byte belongs to both, and their lengths were asserted equal upstream, which is what puts one shared byte count in bounds for both.

tcoratger:
Hi @oksanaphmn — I pushed one commit to this branch (`856f64d6`) applying the review findings, so you are not surprised by it. Your commit is untouched; nothing was rebased, amended or force-pushed, and the kernel arithmetic, the tables, the TBL index computation and the tail handling are all exactly as you wrote them.

What the commit does:

- **Gating.** `mod neon` is now `#[cfg(any(test, all(aarch64, neon, little-endian)))]` like `affine` and `lanes`, with only `byte_butterfly` and the length threshold behind the narrow arch gate. The table construction is therefore type-checked and tested on x86 too, with your exhaustive check against the field product as a test.
- **Deduplication.** `ByteLanes` is split into `ByteRegister` (`BYTES`/`load`/`store`/`xor`) and `ByteLanes: ByteRegister` (the GFNI-only byte map). `butterfly_run` is generic over the former, `uint8x16_t` implements it beside the `__m512i` backend, and the second copy of the butterfly arms is gone. Verified free: the assembly of all seven forced monomorphizations, cross-compiled to `aarch64-unknown-linux-gnu` at `-C opt-level=3`, is byte-identical before and after.
- **Big-endian AArch64.** `target_endian = "little"` added to all three gates, so `aarch64_be` falls back to the stub instead of failing to compile.
- **Threshold.** Left at 256 — I have no AArch64 host to sweep on and would not invent a number. The `to be tuned` TODO is replaced by a comment saying plainly that the value is unmeasured, what it amortises, and what a proper sweep would look like. Happy for you to lower it if you can measure on the M1.
- **Safety comments.** Both `SAFETY` blocks now state the two non-local obligations: the runs are disjoint halves of one split slice, and their lengths were asserted equal upstream.

Not done: the 2-4 register unrolling. That needs a measurement on real hardware to justify, and that thread is left unresolved. It is also cheaper as a follow-up now, since the loop is shared and unrolling it would speed up the GFNI kernel at the same time.

Verified on x86_64: `cargo check` and `cargo clippy --all-targets` warning-free, `cargo build --target aarch64-unknown-linux-gnu` clean (which forces codegen and so evaluates the `const {}` asserts), `cargo test -p p3-binary-dft --lib` 127 passed. I cannot execute the NEON path here — only compile and disassemble it.

tcoratger (approved):

The kernel arithmetic, tables, index computation and tail handling are untouched — the changes are module gating, loop sharing, safety documentation and one exhaustive table test.

The shared loop was checked for codegen impact: forced monomorphizations were cross-compiled at `-C opt-level=3` and the emitted assembly diffed before and after, byte-identical in every case including the steady-state loop.

All review threads are resolved except the unrolling suggestion, which is left open as a follow-up since it cannot be demonstrated without AArch64 hardware.

PR #2223, opened 2026-09-19
feat(bus): report unmatched tuples with source rows

tcoratger:
Adds an out-of-circuit debugger for the binary-native multiset bus.

The debugger replays the symbolic declarations already produced by an AIR. It groups exact tuples by named bus, keeps push and pull multiplicities as integers, and reports the tuples that do not balance.

The concrete trace can only be paired through a symbolic bus profile. This avoids a raw interaction-slice API that could silently inspect the wrong declaration set.

Direction remains structural metadata. Push and pull therefore stay distinct in characteristic two.

The two sides of an unmatched tuple are paired in traversal order, and the reported rows are the ones past the pairing rather than rows that cancel. The sample is bounded, so it covers the whole excess only when it holds as many occurrences as the two counts differ by; only then does removing those occurrences restore balance. When both sides are non-empty the excess is a property of the multiset and no single occurrence is the unique culprit. The reported count is per declaration occurrence rather than per row, so one row that activates two declarations is counted twice.

The report is bounded on both axes, and it carries an exact unmatched total per bus that the bound cannot shrink. A truncated bus header therefore states the true total and how much of it the body holds, and a bus whose tuples all fall past the bound is still named with its total instead of vanishing from the output. It renders one line per retained tuple, grouped by bus. Both bounds can be raised for one run.

Declaration columns are lowered through the representation-independent column view, so a packed Boolean trace is read in place rather than panicking on the dense accessor. Declaration expressions are flattened once per instance rather than walked per row. Shape faults are therefore reported once per declaration instead of once per row, and the four affected error variants no longer carry a row.

This is diagnostic infrastructure only. It does not alter the proof, verifier, transcript, or current warning that product-GKR terminal claims are not yet authenticated against committed columns.

One prover-path change rides along: the symbolic-expression walk in the bus planner now skips nodes it has already visited, which turns a path-wise walk that is exponential in expression depth into a node-wise one. It is semantically neutral, but when an expression contains two distinct unsupported accesses it can change which of them is reported first, so a caller matching on the error variant may see a different one.

Checks:

- cargo test -p p3-bus
- cargo +nightly fmt --all -- --check
- python3 scripts/check.py lint

A separate read-only cryptographic audit found no blocking correctness or soundness issue. It specifically checked characteristic-two direction handling, named-domain separation, selector semantics, symbolic replay parity, exact multiplicities, malformed inputs, and reachable panic paths.

---

Part of #2236.


bus/src/debug.rs
```diff
@@ -0,0 +1,711 @@
...
+    /// Symbolic tuple declarations in AIR emission order.
+    interactions: &'a [SymbolicBusInteraction<F>],
+}
+
+impl<'a, F: Field> BusDebugInstance<'a, F> {
+    /// Pairs concrete AIR data with declarations from its symbolic profile.
+    ///
+    /// The profile must come from the same AIR and layout as the concrete data.
+    #[must_use]
+    pub fn new<EF: ExtensionField<F>>(
```
Nashtare, line 62: `Table` is polynomial-major: `Table::num_polys()` is the backing matrix's *height*, and `Table::new` only asserts that the width is a power of two and the height is nonzero. A caller who hands over the usual row-major AIR trace — say 1024 rows by 2 columns — passes both assertions, and the debugger then reads `num_polys() == 1024`, `num_variables() == 1`, scans **two** rows, and interprets trace *rows* 0 and 1 as columns. `MainColumnOutOfRange` never fires because 1024 exceeds every column index, so `check` returns `Ok` with a fabricated imbalance list and no error.

The orientation is documented only on the private `main` field, which rustdoc does not render; `new`'s own doc says nothing about it.

`new` already receives the `BusSymbolicBuilder`, which holds the `AirLayout` (`main_width`, `preprocessed_width`, `num_public_values` are public fields). Could it add an accessor and reject `main.num_polys() != layout.main_width` (and the same for preprocessed and public values)? That turns the worst failure mode into a typed error, and it costs one comparison.

bus/src/debug.rs
```diff
@@ -0,0 +1,711 @@
...
+        public_values: &'a [F],
+        profile: &'a BusSymbolicBuilder<F, EF>,
+    ) -> Self {
+        // The profile remains the sole source of symbolic declarations.
+        // This prevents a caller from passing an unrelated raw interaction slice.
+        Self {
+            main,
+            preprocessed,
+            public_values,
+            interactions: profile.interactions(),
```
Nashtare, line 97: `BusDebugInstance::new` has no callers anywhere in the repo — all four tests build the struct through the private literal. So the documented safety mechanism here ("the profile remains the sole source of symbolic declarations") is never exercised, and neither is the `SymbolicVariable.index -> Table::poly(index)` mapping that the orientation comment above is about. Replacing this line with `interactions: &[]` leaves the suite green.

One end-to-end test that runs a real `Air` through `BusSymbolicBuilder::from_air` into `check` would cover both.

bus/src/debug.rs
```diff
@@ -0,0 +1,711 @@
...
+    /// Committed trace columns.
+    main: &'a Table<F>,
+    /// Optional fixed trace columns.
+    preprocessed: Option<&'a Table<F>>,
+    /// Public inputs supplied to the AIR.
+    public_values: &'a [F],
+}
+
+impl<F: Field> RowEvaluator<'_, F> {
+    fn evaluate(&self, expression: &SymbolicExpression<F>) -> Result<F, BusDebugError> {
```
Nashtare: A design question rather than a defect. The PR chooses symbolic replay so that the concrete trace "can only be paired through a symbolic bus profile", which avoids a raw interaction-slice API. But `air/src/check_constraints.rs` already has `DebugConstraintBuilder`, a concrete per-row `AirBuilder`; implementing `BusInteractionRecorder` for it and calling `air.eval(&mut builder)` per row would give the same guarantee more strongly — the declarations come from the AIR itself, not from a profile the caller supplies — and would drop this second hand-rolled expression evaluator along with the orientation question above.

If there is a reason the profile route is preferable (callers that hold a profile but not the `Air`, say), it would be worth a sentence in the module docs, since the current rationale reads like an argument for the builder route.

Separately: this recurses over an `Arc`-shared DAG once per *path*, not per node. `double()` is `self.dup() + self.dup()`, so a 32-bit recomposition loop is ~64 nodes but ~2^32 paths, and `check` would then hang with no diagnostic. `plan.rs::validate_expression` deliberately uses an explicit worklist for the same shape of traversal. The `# Performance` block bounds memory and output but says nothing about time.

bus/src/debug.rs
```diff
@@ -0,0 +1,711 @@
...
+    use alloc::vec;
+
+    use p3_air::symbolic::{BaseEntry, SymbolicVariable};
+    use p3_baby_bear::BabyBear;
+    use p3_field::PrimeCharacteristicRing;
+    use p3_matrix::dense::RowMajorMatrix;
+
+    use super::*;
+
+    type F = BabyBear;
```
Nashtare, line 1003: All four tests use `BabyBear`. The module's headline claim, and the test named `named_buses_and_directions_never_cancel_in_the_field`, are about characteristic two; `builder.rs`'s own tests use `BinaryField128`, so the pattern already exists in the crate. Could at least the direction/name-separation test run over a binary field?

Also unpinned by any test: `LOCATION_LIMIT` and its "the count remains exact after this sample is full" claim (no test produces more than four occurrences of one tuple on one side); every preprocessed path, including `PreprocessedHeightMismatch`; `PublicValueOutOfRange`; `MainColumnOutOfRange`; and the `IsFirstRow`/`IsLastRow`/`IsTransition` leaves, which are the only place row/height arithmetic happens. Changing `IsLastRow` to `self.row == self.height` (always false) breaks nothing today.

bus/src/debug.rs
```diff
@@ -0,0 +1,711 @@
...
+                    }
+
+                    // Resolve the exact payload tuple recorded by the symbolic AIR pass.
+                    let tuple = interaction
+                        .fields
+                        .iter()
+                        .map(|field| evaluator.evaluate(field))
+                        .collect::<Result<Vec<_>, _>>()?;
+                    let bus = domain_indices[interaction.bus_name.as_str()];
+                    let key = (bus, tuple.clone());
```
Nashtare: Nit: this clones the tuple into a throwaway key on every active occurrence, and on the common path the probe hits and the clone is dropped immediately. A 2^20-row AIR with four declarations does ~4M redundant allocations.

Keying per bus — `Vec<HashMap<Vec<F>, usize>>` indexed by `bus` — lets the probe borrow `&tuple` and removes the clone without changing the deterministic ordering.

Nashtare (commented):
I read the whole module against the rest of `p3-bus` to check that the replay matches what the argument actually constrains. The semantics hold up; my comments are about what the module is paired with and what the tests pin.

**What holds**

- **The integer accounting is the right model, and it errs safely.** The in-circuit bus has no field multiplicity: an active row contributes one factor `offset - fingerprint` and an inactive row contributes `ONE` (`leaf.rs`), so multiplicity *is* cardinality. Equal integer counts therefore imply identical multisets, which implies the products agree for every challenge. The debugger can never report "balanced" on something the argument rejects; the only false positive is a fingerprint collision, which is the safe direction. The characteristic-two worry I went in with does not apply, because this is a product argument rather than a sum.
- **Selector semantics match.** `activation_is_set` accepts only exact `ZERO`/`ONE` and skips on zero, which is what `leaf.rs` materializes and what the Booleanity constraint `push_bus_interaction` emits enforces.
- **Row selectors match the repo's own concrete builder** byte for byte: `IsFirstRow -> row == 0`, `IsLastRow -> row + 1 == height`, `IsTransition -> row + 1 < height` is identical to `air/src/check_constraints.rs`. No off-by-one, no wrap at the last row.
- **No next-row or periodic access can reach the replay.** `plan.rs::validate_expression` rejects `Main{offset != 0}`, `Preprocessed{offset != 0}` and `Periodic` for payload *and* activation expressions, and `BusPlan::build` runs over exactly the interaction slices the row loop later evaluates, so the `unreachable!` in `evaluate_leaf` is genuinely unreachable and every reported `row` is the row that produced the tuple.
- **Bus-name separation is sound.** The grouping key is `(plan-assigned index, tuple)`, the plan rejects two declarations on one name with different payload widths, and distinct buses get distinct nonzero domain identities in reserved slots, so `bus "x"` tuple `(7)` and `bus "y"` tuple `(7,0,0)` cannot merge.
- **The report is deterministic.** Neither hash map is ever iterated; `entries` is a `Vec` in first-occurrence order, `locations` is appended in traversal order and truncated, so the retained four really are the first four. No hash-order dependence in the output path.
- **No reachable panic from caller data.** Every column and public index is bounds-checked into a typed error, the preprocessed height is checked before the row loop, and `Entry::record` uses `checked_add`.
- The diff really is diagnostic-only: nothing outside `debug.rs`, the two `lib.rs` lines, one `Cargo.toml` line and one README sentence.

**What I'd like looked at**

The one that would actually bite a user is the trace orientation (inline, `debug.rs:40`). The rest is test coverage: the public constructor has no callers, and four `BabyBear` tests leave the preprocessed paths, the location cap and the row-selector leaves unpinned.

Nothing here is blocking.

bus/src/debug.rs
tcoratger: Confirmed and fixed. I reproduced the failure mode first: a four-row, two-column trace handed over row-major becomes a table with `num_polys() == 4` and `num_variables() == 1`, and `check` returns `Ok` with one imbalance `{tuple: [1], count: 2}` instead of the four single-occurrence tuples the real trace holds. No error anywhere.

`BusSymbolicBuilder` now retains the `AirLayout` it was built with and exposes it, and `BusDebugInstance::new` returns `Result`, rejecting `main.num_polys() != layout.main_width`, `preprocessed.map_or(0, num_polys) != layout.preprocessed_width`, and `public_values.len() != layout.num_public_values`. Three new typed variants carry the expected and actual counts.

On preprocessed I used the uniform rule `map_or(0, ..)` rather than checking only the `Some` case, because `AirLayout`'s own docs say the committed preprocessed trace is authoritative for that width — so a caller committing none should be passing a layout with width zero.

The orientation is now stated on `new` itself rather than only on the private field. `concrete_data_must_match_the_shape_the_profile_was_built_for` feeds the transposed trace and asserts the new error, plus the two shape errors.

bus/src/debug.rs
tcoratger: Verified the mutation and closed the gap. `a_real_air_reaches_the_checker_through_its_symbolic_profile` builds a two-column AIR, runs it through `BusSymbolicBuilder::from_air`, and feeds the result to `new` and then `check`. The AIR pushes column zero unconditionally and pulls it back under column one as a selector, so on payloads `1,2,3,4` with selectors `1,1,1,0` the only imbalance is tuple `4` pushed once at row three. That pins the variable-index-to-polynomial mapping: any other orientation or column assignment changes the answer.

Replacing the constructor's declarations with `interactions: &[]` now fails that test (expected one imbalance, got zero).

bus/src/debug.rs
tcoratger: The traversal point is right, and it went further than the comment says. I wrote the test first — forty doublings of a main column, eighty nodes and 2^40 paths — and it hung. But it hung in `plan.rs::validate_expression`, before the evaluator ran at all: that worklist bounds the recursion depth, yet it still pushes both children unconditionally, so it also costs one visit per path. The explicit stack was not the part that mattered; memoization was.

Both are fixed now. The evaluator uses an explicit stack plus a completed-value index keyed on node identity, and the planner's validator skips nodes it has already seen, the same way `SymbolicExpr::degree_with` in `p3-air` already does. `shared_operands_are_replayed_once_per_node_rather_than_once_per_path` runs that forty-deep chain and asserts the payload equals one doubled forty times; it completes in milliseconds and did not terminate before the fix. The `# Performance` block now states the time bound.

On the design question: I kept the profile route and added the sentence you asked for to the module docs. The reason is the one you guessed — replay only needs a profile and the committed traces, not the `Air` value itself, so a caller that has serialized or otherwise outlived its AIR can still be diagnosed. I agree the builder route sources declarations more strongly, and it would drop this evaluator entirely; that is a larger change than this PR should absorb, so I have not done it here and am happy to see it filed separately.

bus/src/debug.rs
tcoratger: All of these are now covered.

`named_buses_and_directions_never_cancel_in_the_field` runs over `BinaryField128` and was extended: besides the two identically-payloaded buses, a third bus takes one push and one pull of the same tuple and must not appear in the report, so the test now pins direction as well as name in characteristic two.

New tests: `the_source_sample_is_bounded_while_the_count_stays_exact` pushes one tuple on eight rows and asserts count eight, sample length four, rows `[0,1,2,3]`. `fixed_columns_share_the_committed_row_domain` covers a successful fixed-column read, the height mismatch, the missing trace, and the out-of-range fixed column. `reads_past_the_supplied_data_are_named_rather_than_wrapped` covers the main-column and public-value range errors. `row_selectors_follow_the_position_of_the_row_in_its_trace` drives each of the three leaves as an activation over a four-row trace and asserts exactly which payloads survive.

I checked your mutation claim before writing that last one: `self.row == self.height` did leave the old suite green, and it now fails. So does breaking the transition leaf to `self.row < self.height`.

bus/src/debug.rs
tcoratger: Done as suggested. The index is now `Vec<HashMap<Vec<F>, usize>>` indexed by bus, so the probe borrows `&tuple`; the only remaining clone is the one that seeds a new entry, once per distinct tuple rather than once per active occurrence.

I did not assume the ordering survived. `distinct_tuples_are_reported_in_first_occurrence_order` uses payloads `9,4,9,4` on one bus and asserts the report lists tuple `9` before tuple `4` — scan order, not hash or sort order — with the correct multiplicity on each.

tcoratger (commented):
Second pass over `bus/src/debug.rs` at `710f3464`, focused on the two hazards a debug facility actually carries: does it reach the proving path, and what does it cost. Nashtare's round already closed the orientation bug, the path-vs-node traversal, the clone in the probe, and the `BabyBear`-only coverage; I verified each of those fixes rather than re-raising them.

## Leakage into prove / verify: clean, with one caveat the PR body understates

Checked and confirmed:

- Nothing outside `bus/src/debug.rs` references `BusDebugReport`, `BusDebugInstance`, `BusDebugError` or `BusImbalance`. `grep -rn` over the whole tree returns only `bus/src/lib.rs`'s re-export.
- Stronger still: **no crate in the workspace depends on `p3-bus`.** The only `p3-bus` entries in any `Cargo.toml` are the root workspace declaration and the package itself. So there is no prove or verify path to leak into on this base at all.
- `BusSymbolicBuilder` grew one field, `layout: AirLayout` (7 `usize`, `Copy`, 56 bytes). The struct derives only `Debug` — no `Serialize`, no transcript absorption, and it is a setup-time object. No commitment, no hash, no serialized wire format changed.
- The module is **not** `cfg`-gated. It compiles into release builds. That is correct here precisely because nothing calls it: there is no "collect unconditionally, report conditionally" path, no `Vec` that is always built and dropped. The disabled cost is genuinely zero because there is no disabled state to speak of — the entry point is simply never reached. If a future PR adds an `assert_bus_balanced!` hook inside `prove`, that property has to be re-checked; it is worth one sentence in the module docs saying the module must stay caller-driven.

The caveat: the body says *"It does not alter the proof, verifier, transcript."* `bus/src/plan.rs::validate_expression` is proof-setup code, not diagnostics, and it changed. I convinced myself the change is semantically neutral (details inline at `plan.rs:577`), but "diagnostic infrastructure only" is not quite the right description of a diff that touches the planner.

## Cost, measured

All numbers from the PR head in this worktree, `--release`, one `Push` declaration, `BabyBear`.

| what | measurement |
|---|---|
| throughput, 4-wide tuple, 2 of 4 fields non-leaf | 175 ms / 2^18 rows (1.5 M rows/s) |
| same, all four fields plain leaves | 108 ms / 2^18 rows (2.4 M rows/s) |
| retained memory, 2^20 distinct width-1 tuples | 342 MB RSS, **334 bytes per distinct tuple** |
| `format!("{:#?}", report)`, 2^10 unmatched tuples | 540 KB |

The 175 vs 108 ms gap is the per-row `HashMap` scratch in `evaluate` (`debug.rs:491`). The 334 bytes for 4 bytes of payload is an ~80x blowup (`debug.rs:320`). The 540 KB extrapolates to ~540 MB of text at 2^20 (`debug.rs:120`).

## Verified clean

- `1usize << instance.main.num_variables()` at `debug.rs:327` cannot shift-overflow: `BusPlan::build` rejects `log_height >= usize::BITS` at `plan.rs:205`, and it runs first at `debug.rs:293`.
- `domain_indices[...]` and `entry_indices[bus]` cannot panic: `plan.domains()` is built from a `BTreeMap` keyed on every `bus_name` seen in the same slices, so the index map is total and its range is exactly `0..domains.len()`.
- `poly(i).as_slice()[self.row]` is in range: `row < 1 << num_variables()` is exactly the polynomial length.
- The pointer-keyed memo tables in both `debug.rs:491` and `plan.rs:577` are sound. Every reference is reborrowed from a DAG rooted at a parameter that outlives the walk, so no node can be dropped mid-traversal and no address can be reused for a different node.
- The `Enter`/`Leave` operand-stack discipline is correct, including the case where a shared node is both a deep descendant of the left operand and the right operand itself. Each `Enter` nets exactly one value onto `operands`, each `Leave` pops its arity and pushes one, and `Add`/`Sub`/`Mul` push `y` before `x` so the stack yields left before right.
- `Entry::record` uses `checked_add`; every column and public index becomes a typed error.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+    pub row: usize,
+}
+
+/// Exact occurrence count with a bounded source sample.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BusDebugOccurrence {
+    /// Number of active rows carrying the tuple.
+    pub count: usize,
+    /// First source locations in deterministic traversal order.
+    pub locations: Vec<BusDebugLocation>,
```
tcoratger: **[MAJOR] The retained locations are the *matched* occurrences, not the unmatched ones**

I ran this on the head commit: eight pushes and seven pulls of tuple `9`, one instance each.

```
bus=d tuple=[9] push=8 pull=7
  push_locs = rows 0, 1, 2, 3
  pull_locs = rows 0, 1, 2, 3
```

Every retained row is provably part of a matched pair. The one excess push is at neither end of the sample, and the four rows the report names are the four the developer least needs to look at.

This is not a bug in the code — the field is honestly named `locations` and documented "first source locations". It is a limitation in what the sample is worth, and it splits cleanly by bus shape:

- **Sparse buses** (memory: `(addr, value, timestamp)`) have `count` 1 vs 0 per tuple. The sample is exactly the culprit. Perfect.
- **Dense buses** (range checks, byte tables: one tuple, `2^20` occurrences) have `count = n` vs `n-1`. The sample is `{0,1,2,3}` every time, regardless of where the fault is. Zero information.

The second shape is a first-class use of a multiset bus, so the tool degrades to useless on half its intended workload, silently.

Two things would fix most of it:

1. Retain head *and* tail: `LOCATION_LIMIT/2` from the front, `LOCATION_LIMIT/2` in a two-slot ring from the back. Costs one extra branch in `record` and no extra memory, and at least brackets the range the fault lives in.
2. Say in the doc that for `|push - pull| < count` no individual row *is* the unmatched one — the excess is a property of the multiset, not of a row. Right now the field doc invites the opposite reading, and the PR title ("report unmatched tuples with source rows") reinforces it.

A third, more ambitious option: when the imbalance is `k`, report the `k` locations that are *not* covered by a greedy pairing against the other side. For declaration-level faults (one column wrong in one declaration) that collapses to the right answer immediately, because the two sides differ by declaration index rather than by row.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+    /// Consumed occurrences and their first source locations.
+    pub pulls: BusDebugOccurrence,
+}
+
+/// Deterministic list of unmatched bus tuples.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BusDebugReport<F> {
+    /// Unmatched tuples in first-occurrence order.
+    pub imbalances: Vec<BusImbalance<F>>,
+}
```
tcoratger, line 199: **[MAJOR] The report is unbounded and has no `Display`**

`LOCATION_LIMIT` bounds the sample *within* one tuple. Nothing bounds the number of tuples, and that is the axis that actually blows up.

Measured on the head commit: a 2^10-row trace pushing 1024 distinct values with no pull side at all yields **1024 `BusImbalance` entries**, and `format!("{:#?}", report)` is **540 043 bytes**. At 2^20 that is ~540 MB of text — from one `println!`.

And that case is not exotic. It is the first thing a developer hits: the pull side is not implemented yet, or a whole declaration was omitted, so *every* tuple is unmatched. The tool's very first output on its most common input is a half-gigabyte dump.

There is also no `Display` impl anywhere in the module, so `{:?}` / `{:#?}` is the only way to render it — one line per struct field, field elements at `Debug` fidelity, nothing elided. For a module whose entire product is developer-facing text, that is the missing half.

Suggestion, keeping `imbalances` exhaustive for programmatic use:

```
impl<F: Display> Display for BusDebugReport<F> { ... }
```

printing something like

```
bus "memory": 3 unmatched tuples
  (0x0007, 0x002a)  push 3  pull 2   push@ air0 decl1 row17, row91, ...
  (0x0009, 0x0000)  push 0  pull 1   pull@ air1 decl0 row 3
  ... and 1041 more (see `imbalances`)
```

with a cap of ~32 rendered imbalances. One line per tuple instead of eleven, grouped by bus, and it terminates.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+            .iter()
+            .enumerate()
+            .map(|(index, domain)| (domain.name.as_str(), index))
+            .collect::<HashMap<_, _>>();
+
+        // One hash index per named multiset provides constant-time grouping.
+        // Keying per bus lets the probe borrow the tuple instead of cloning it on every occurrence.
+        // The parallel vector preserves deterministic first-occurrence order for diagnostics.
+        let mut entry_indices = alloc::vec![HashMap::<Vec<F>, usize>::new(); plan.domains().len()];
+        let mut entries = Vec::<Entry<F>>::new();
```
tcoratger: **[MINOR] 334 bytes retained per distinct tuple, for 4 bytes of payload**

Measured: 2^20 rows, one `Push`, width-1 `BabyBear` payload, all values distinct. RSS delta across `check` is **342 200 kB**, i.e. **334 bytes per distinct tuple** — an ~80x blowup over the 4 bytes of payload. The `# Performance` bullet "memory is linear in the number of distinct active tuples" is true, but the constant is the whole story and the doc does not give it.

Breakdown, and two of the three terms are avoidable:

| term | bytes | avoidable |
|---|---|---|
| `Entry<F>` in `entries` (bus + 3 `Vec` headers + 2 counts) | ~96 | partly |
| `tuple` heap allocation | ~32 | — |
| `entry_indices` key: a **second** `Vec<F>` header + heap + table slot | ~70 | yes |
| `locations: Vec` first push (min non-zero cap is 4 for 24-byte elements, so `4 * 24`) | 96 | yes |

1. `locations` is capped at `LOCATION_LIMIT = 4` *at compile time*, yet it is a heap `Vec`. An inline `[MaybeUninit<BusDebugLocation>; LOCATION_LIMIT]` + `len: u8`, or just `[Option<BusDebugLocation>; 4]`, removes 96 bytes and one allocation per entry — ~29% on its own. It is a breaking change to a `pub` field, so it is cheaper to decide now than after release.
2. The tuple is stored twice: once as the `entry_indices` key and once in `Entry.tuple`. Keying on a hash of the tuple and confirming against `entries[i].tuple` (or `hashbrown`'s `RawTable` with a custom eq closure) keeps one copy. Halves the `Vec<F>` header + heap cost.

Neither is urgent for a `2^16` development trace; both matter at `2^22`, which is where you reach for this tool.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+        // Arithmetic nodes share their operands, so the expression is a graph rather than a tree.
+        // Walking it per path costs time exponential in the depth, which a bit recomposition reaches immediately.
+        // An explicit stack keyed on node identity evaluates each distinct node once and bounds the recursion depth.
+        if let SymbolicExpr::Leaf(leaf) = expression {
+            return self.evaluate_leaf(leaf);
+        }
+
+        let mut done = HashMap::<*const SymbolicExpression<F>, F>::new();
+        let mut steps = alloc::vec![Step::Enter(expression)];
+        let mut operands = Vec::<F>::new();
```
tcoratger: **[MINOR] The scratch tables are rebuilt per field, per declaration, per row**

A/B on the head commit, 2^18 rows, one declaration, 4-wide tuple:

- two of four fields non-leaf: **175 ms**
- all four fields plain leaves (leaf fast path at line 487, no `HashMap`): **108 ms**

So ~67 ms, **62% overhead**, for four `HashMap` + `Vec` + `Vec` constructions per row over expressions of 2-4 nodes. That is allocator and hashing cost, not arithmetic.

The key observation: **the DAG shape is row-invariant.** Only the leaf values change. So the pointer hashing is recomputed `height` times for an answer that was already known before the row loop started.

The cheap fix is to hoist `done` / `steps` / `operands` into the row loop and `.clear()` them — keeps the allocations, kills the `malloc`s.

The right fix is to flatten once per declaration, before the row loop:

```
enum Op { Leaf(BaseLeaf<F>), Neg(u32), Add(u32, u32), Sub(u32, u32), Mul(u32, u32) }
// topological order, operands are indices into a reusable Vec<F> of the same length
```

Then each row is a linear scan with no hashing and no allocation, and `slots[i]` replaces the `HashMap` probe. It also moves `MainColumnOutOfRange` / `PublicValueOutOfRange` to the compile step, where they are reported once instead of `height` times — which is what a user wants anyway, since today a bad column index means the error names row 0 and says nothing about the other 2^20 rows that would have failed identically.

bus/src/plan.rs
```diff
@@ -573,8 +574,14 @@ fn validate_expression<F: Field>(
     location: BusExpressionLocation,
     expression: &SymbolicExpression<F>,
 ) -> Result<(), BusPlanError> {
+    // Arithmetic nodes share their operands, so the expression is a graph rather than a tree.
+    // Walking it per path costs time exponential in the depth, which a bit recomposition reaches immediately.
+    let mut seen = HashSet::<*const SymbolicExpression<F>>::new();
```
tcoratger, line 579: **[MINOR] This is a prover-path change riding in a diagnostics PR, and its only test lives in `debug.rs`**

I verified the change itself is safe:

- **Semantically neutral.** A node reaches `seen` only after it failed to match any rejecting arm, so re-visiting it can never produce an error. Skipping repeats cannot turn a rejected plan into an accepted one, or vice versa.
- **Pointer keys are sound.** Every `&SymbolicExpression<F>` in `pending` is reborrowed through an `Arc` rooted at the `expression` parameter, which outlives the loop. No node is dropped mid-walk, so no address is reused for a different node. (Same argument covers `debug.rs:491`.)
- **The exponential is real.** `E::double()` is `dup() + dup()`, so a 32-bit recomposition is ~64 nodes and ~2^32 root-to-leaf paths. The previous worklist bounded depth but still pushed both children unconditionally, so it visited once per path.

Two things I would still flag:

1. **It can change which error is reported.** If two distinct nodes both violate (say a `Main { offset: 1 }` and a `Periodic` in the same expression), skipping duplicates perturbs the stack and therefore which one is popped first. Nothing asserts a specific first `BusPlanError`, so no test breaks — but `BusPlanError` is public, and a caller matching on the variant sees a different one. Worth one line in the PR body.

2. **The regression test is in the wrong module.** `shared_operands_are_replayed_once_per_node_rather_than_once_per_path` lives in `debug.rs::tests`. If `debug` is ever gated or removed, this fix silently loses its coverage — and `plan.rs` already has its own `mod tests` with a dozen `BusPlan::build` cases. A 40-deep `dup() + dup()` chain asserting `BusPlan::build` returns rather than hangs belongs next to them. Reverting just these three lines should fail a test in `plan.rs`, not in `debug.rs`.

Cross-PR: nothing in the workspace depends on `p3-bus` today, so this is latent. #2244 wires buses into multi-stark, at which point this becomes a live fix for a prover-setup hang on any AIR that recomposes bits into a bus payload — which is most of them. That is a reason to land it, not a reason to hide it inside a tooling PR.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+                        air: self.air,
+                        declaration: self.declaration,
+                        row: self.row,
+                        index: variable.index,
+                        len: self.public_values.len(),
+                    },
+                ),
+                BaseEntry::Main { .. } | BaseEntry::Preprocessed { .. } | BaseEntry::Periodic => {
+                    unreachable!("bus planning rejected an unsupported expression access")
+                }
```
tcoratger: **[MINOR] This `unreachable!` is a panic in the one routine that runs when the system is known-broken**

Nashtare established it is unreachable today, and I agree with the argument: `BusPlan::build` runs first at line 293, `validate_expression` rejects `Main { offset != 0 }`, `Preprocessed { offset != 0 }` and `Periodic` for both payload and activation, and it runs over exactly the slices this loop later evaluates.

The depth I would add is that it should not be an invariant at all. It is held across a module boundary by call ordering inside `check`, with nothing in the type system tying the two together — and this function is the last thing that runs before a developer learns what is wrong with their trace. A panic here replaces the diagnosis with a backtrace.

The typed alternative costs nothing, because every piece already exists and is public:

```
BusDebugError::Plan(BusPlanError::UnsupportedAccess {
    location: BusExpressionLocation::...,
    access: UnsupportedBusAccess::MainOffset(offset),
})
```

`BusDebugError` already has `#[from] BusPlanError`, `UnsupportedBusAccess` is already exported from `lib.rs:25`, and `evaluate_leaf` already returns `Result`. Same for the two `expect`s at lines ~535/545 — those are genuine internal invariants of the stack machine and I would leave them, but the leaf arm is caller-reachable data being trusted on the strength of a sibling function.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+                for (declaration, interaction) in instance.interactions.iter().enumerate() {
+                    let evaluator = RowEvaluator {
+                        declaration,
+                        ..evaluator
+                    };
+
+                    // Inactive rows contribute the product identity and no multiset occurrence.
+                    if !activation_is_set(&interaction.activation, &evaluator)? {
+                        continue;
+                    }
```
tcoratger: **[MINOR] Padding rows are reported as first-class imbalances, undocumented and untested**

A real trace is padded to a power of two, usually with zeros, and `check` scans `0..1 << num_variables()` unconditionally. So a declaration with `BusActivation::Always` records the all-zeros tuple on every padding row.

I confirmed the mechanism: a zero-padded pull table produced a separate `tuple=[0] push=0 pull=1` entry alongside the real imbalance.

The replay is *faithful* here — the in-circuit argument does exactly the same thing, so the AIR genuinely has to gate declarations off the padding. That is the point worth documenting: the all-zeros imbalance at the top of the report is not noise, it is a real unbalanced contribution, and the fix is a selector rather than a filter in the debugger.

But nothing in the module says so, and no test uses a padded trace. Concretely, the first real report a user sees will lead with a giant `[0, 0, ...]` entry whose `count` equals the padding height, and they will reasonably assume the tool is broken. One sentence in the `check` docs plus one test (height-4 trace, two real rows, two zero rows, `Always` declaration, assert the zero tuple appears with count 2) closes it.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+        SymbolicBusInteraction {
+            bus_name: name.to_string(),
+            direction,
+            fields,
+            activation,
+        }
+    }
+
+    #[test]
+    fn balanced_mixed_height_instances_have_no_diagnostics() {
```
tcoratger, line 1047: **[MINOR] Missing: a mutation property test — the one test that proves the report points at the right row**

The suite now covers the shapes well (first-occurrence ordering, the location cap, the selector leaves, shape errors, the real-AIR path). What it does not have is the test that the feature is *for*: take something balanced, break exactly one row, assert the report names that row.

It fits `proptest` almost too neatly, because uniqueness makes the answer exactly determined:

```
proptest! {
  #[test]
  fn a_single_mutated_row_is_named_by_the_report(
      log_height in 1usize..8,
      row in 0usize..256,
      delta in 1u64..1000,
  ) {
      let height = 1 << log_height;
      let row = row % height;

      // Distinct payloads => every tuple has multiplicity exactly one,
      // so the excess is attributable to a single row by construction.
      let values: Vec<u64> = (0..height as u64).map(|i| i * 4096 + 1).collect();
      let push = table::<F>(&[&values]);

      let mut pulled = values.clone();
      pulled[row] += delta;            // one row, one column, one fault
      let pull = table::<F>(&[&pulled]);

      let report = BusDebugReport::check(&instances(&push, &pull)).unwrap();

      // Exactly two entries: the original value lost its pull, the new value gained one.
      prop_assert_eq!(report.imbalances.len(), 2);

      let lost = &report.imbalances[0];
      prop_assert_eq!(lost.tuple, vec![F::from_u64(values[row])]);
      prop_assert_eq!((lost.pushes.count, lost.pulls.count), (1, 0));
      prop_assert_eq!(lost.pushes.locations[0].row, row);   // <- names the mutated row

      let gained = &report.imbalances[1];
      prop_assert_eq!((gained.pushes.count, gained.pulls.count), (0, 1));
      prop_assert_eq!(gained.pulls.locations[0].row, row);  // <- and from the other side
  }
}
```

Two assertions in there are load-bearing and nothing in the current suite makes them:

- `locations[0].row == row` on **both** sides — the diagnostic is correct for a sending-side *and* a receiving-side fault, over randomized heights rather than the one hand-built fixture.
- `imbalances.len() == 2` — no spurious entries from padding or from the other rows.

`proptest` is not currently a dev-dependency of `p3-bus`; it is already in the workspace (`p3-field`, `p3-matrix` use it), so this is one line in `Cargo.toml`.

Two further gaps a deterministic test would close more cheaply than a property:

- **Imbalance across instances of different heights.** `balanced_mixed_height_instances_have_no_diagnostics` covers mixed heights when balanced; `mismatch_reports_exact_counts_and_first_sources` covers imbalance at heights 2 and 1. Nothing covers a 3-instance statement where the fault is in the *middle* instance, which is what pins `air` in `BusDebugLocation` against an off-by-one.
- **Two declarations on the same row of the same AIR producing the same tuple.** That pins the `declaration` field, which today is only ever 0 or trivially distinguished.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+        // One hash index per named multiset provides constant-time grouping.
+        // Keying per bus lets the probe borrow the tuple instead of cloning it on every occurrence.
+        // The parallel vector preserves deterministic first-occurrence order for diagnostics.
+        let mut entry_indices = alloc::vec![HashMap::<Vec<F>, usize>::new(); plan.domains().len()];
+        let mut entries = Vec::<Entry<F>>::new();
+
+        for (air, instance) in instances.iter().enumerate() {
+            let height = 1usize << instance.main.num_variables();
+
+            for row in 0..height {
```
tcoratger: **[NIT] An instance with no declarations still costs a full row scan**

If `instance.interactions` is empty, this loop runs `2^num_variables()` times constructing a `RowEvaluator` and immediately falling out of an empty inner loop. On a 2^24 AIR with no bus declarations — entirely normal in a mixed statement where only some tables use the bus — that is 16M iterations of nothing.

```
if instance.interactions.is_empty() {
    continue;
}
```

`BusPlan::build` already established the statement has *some* declaration, so this cannot short-circuit the whole check.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+pub struct BusDebugOccurrence {
+    /// Number of active rows carrying the tuple.
+    pub count: usize,
+    /// First source locations in deterministic traversal order.
+    pub locations: Vec<BusDebugLocation>,
+}
+
+/// One tuple whose push and pull multiplicities differ.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BusImbalance<F> {
```
tcoratger, line 166: **[NIT] `#[non_exhaustive]` on the report types**

`BusDebugError` has it. `BusImbalance`, `BusDebugOccurrence`, `BusDebugLocation` and `BusDebugReport` do not, and every field is `pub`.

These are exactly the types that will grow: a `truncated: bool` once the report is capped, a tail sample alongside `locations`, a `BusDomain` identity next to `bus_name`, a declaration span. Each of those is a breaking change today and a patch release with `#[non_exhaustive]`.

The cost is that callers construct them via `..Default::default()` or not at all — which is fine here, since these are produced by `check` and consumed by the user, never built by hand outside this module's own tests.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+
+use crate::{
+    BusActivation, BusDirection, BusPlan, BusPlanError, BusPlanInput, BusSymbolicBuilder,
+    SymbolicBusInteraction,
+};
+
+/// Maximum source locations retained for one tuple and direction.
+///
+/// The count remains exact after this diagnostic sample is full.
+const LOCATION_LIMIT: usize = 4;
```
tcoratger: **[NIT] `LOCATION_LIMIT` is private but is part of the documented contract**

The PR body and `check`'s `# Performance` block both promise "at most four source locations per direction". The constant that makes it four is module-private, so a reader of the rustdoc cannot confirm the number and a caller cannot branch on it (e.g. "if `locations.len() == LOCATION_LIMIT` the sample is truncated, so do not present it as exhaustive").

Either `pub const LOCATION_LIMIT: usize = 4;` re-exported from `lib.rs`, or reference it from the `check` docs with an intra-doc link so the rendered text carries the value.

Related: there is no way to raise it. A developer staring at a 300-occurrence imbalance would reasonably want 64 samples for one run. A `check_with_sample_limit(instances, limit)` alongside `check` is a two-line addition and makes `check` a thin wrapper.

bus/src/debug.rs
```diff
@@ -0,0 +1,1208 @@
...
+
+/// Concrete data required to replay one AIR's bus declarations.
+#[derive(Clone, Copy, Debug)]
+pub struct BusDebugInstance<'a, F: Field> {
```
tcoratger, line 24: **[NIT] The derived `Debug` on `BusDebugInstance` dumps both entire traces**

The struct is `Copy`, so it is easy to hand to `dbg!` or drop into a `{:?}` — and the derive will render `main: Table { .. }` in full, which is the whole committed trace. At 2^20 x 8 that is a multi-hundred-megabyte string from a reflex debugging action, in the module people reach for precisely when they are reflex-debugging.

A hand-written `Debug` printing shapes rather than contents is friendlier and three lines:

```
f.debug_struct("BusDebugInstance")
    .field("main", &format_args!("2^{} x {}", self.main.num_variables(), self.main.num_polys()))
    .field("preprocessed", &self.preprocessed.map(|t| t.num_polys()))
    .field("public_values", &self.public_values.len())
    .field("declarations", &self.interactions.len())
    .finish()
```

bus/src/debug.rs
tcoratger: Reproduced on the head commit before changing anything, and I get your output exactly:

```
tuple=[9] push=8 pull=7
  push_locs = rows 0, 1, 2, 3
  pull_locs = rows 0, 1, 2, 3
```

Fixed with your option 3, not option 1. Head-and-tail brackets the range but still spends half the sample on rows that provably cancel, and the tail is only the culprit when the fault happens to be last.

The two sides are now paired in traversal order, and the sample is drawn from the first occurrence past the pairing. For a tuple with `p` pushes and `q` pulls the excess side reports occurrences `[min(p,q), min(p,q) + limit)` and the smaller side reports nothing at all, because every one of its occurrences is paired.

Same fixture after the change:

```
tuple=[9] push=8 pull=7
  push_locs = [air 0 decl 0 row 7]
  pull_locs = []
```

This needed a structural change rather than a branch in `record`: which occurrence is unmatched is unknown until the totals are in. The scan is now two passes. The first counts and retains no source row at all; the second runs only when something is unmatched, and samples only the unpaired occurrences. That also deleted the per-entry location vector, which was a third of the retained memory — see the memory thread.

On the honesty point, which I think is the real content of your comment: the field documentation now states that the two sides are paired in traversal order, that the smaller side carries no sample, and that removing the sampled rows restores balance for the tuple but no single row is the unique culprit when both sides are non-empty. The sample is a canonical witness set, not a claim about where the fault is.

Two tests pin it, and both fail under the old behaviour:

- `the_sample_skips_the_occurrences_that_cancel_against_the_other_side` is your 8-push / 7-pull fixture, asserting the push sample is exactly row 7 and the pull sample is empty. The old code named rows 0-3 on both sides.
- `mismatch_reports_exact_counts_and_unmatched_sources` (renamed) asserts the same property on the 2-vs-1 case.

bus/src/debug.rs
tcoratger: Reproduced: 2^10 rows, 1024 distinct pushes, no pull side, giving 1024 entries and `format!("{:#?}")` at 540 543 bytes. That is your number to the byte.

Both halves are fixed.

The report is capped at 64 retained tuples by default and carries a count of the unmatched tuples it dropped, so truncation is visible rather than silent. There is now a `Display` that renders one line per tuple, grouped by bus, with the trailing count.

Same input after the change: 64 entries, 960 unreported, `{:#?}` at 33 708 bytes, `{}` at 3 819 bytes. Neither figure grows with the trace any more.

```
bus "b": 64 unmatched tuples
  (0)  push 1  pull 0  excess push at air 0 decl 0 row 0
  (1)  push 1  pull 0  excess push at air 0 decl 0 row 1
  ...
... and 960 more unmatched tuples
```

`imbalances` stays exhaustive for programmatic use within the cap, and `check_with_limits` raises either bound for one run — which also answers the sample-limit half of your last nit.

On your related point: the four report types now carry `#[non_exhaustive]`. I used a dropped-tuple count rather than a `truncated: bool`, because the renderer can then say how many, and the bool is recoverable from it.

bus/src/debug.rs
tcoratger: Reproduced at 333 bytes per distinct tuple on your fixture (2^20 rows, one push, width-1 BabyBear, all values distinct), one byte off your 334.

I also measured peak rather than retained, because retained understates the working set: the old peak was 494 bytes per distinct tuple.

Both of your avoidable terms are gone, by a different route than the two you proposed.

The location vector has left the accumulator entirely. The counting pass retains no source row, so nothing allocates a location vector; only the capped set of reported tuples ever gets one, during the second pass.

The accumulator no longer holds a second copy of the tuple. It is a bus index and two counters. The tuple lives exactly once, as the hash key, and the reported tuples are recovered during the second pass, which recomputes them anyway.

A/B on the same fixture:

| | before | after |
|---|---|---|
| retained after the call | 333 B/tuple | 32 B/tuple |
| peak during the call | 494 B/tuple | 146 B/tuple |

I did not take the inline-array or `RawTable` routes. Both targeted terms that the two-pass restructure removes outright, and each costs API or `unsafe` surface that this buys for free. If the remaining 146 bytes becomes the binding constraint at 2^22 the raw-table idea is still available, but it would now be buying the 24-byte header rather than the 70 you costed.

bus/src/debug.rs
tcoratger: Reproduced, and my A/B is much worse than yours. Same shape — 2^18 rows, one declaration, 4-wide tuple — all four fields plain leaves at 8.09 ms, two of four non-leaf at 48.44 ms. That is 499% overhead, not 62%. I think the difference is tuple distinctness: my fixture produces 1000 distinct tuples, so the accumulator hashing does not dilute the scratch cost the way a denser fixture would.

I took your second route, not the cheap one. Each declaration's expressions are flattened once per instance into a linear op list in evaluation order, with operands as indices into a reusable slot vector. Flattening resolves every leaf to a concrete column slice, so a row is a linear scan with no hashing, no allocation and no bounds check.

After: all-leaf 8.66 ms, two-non-leaf 11.29 ms. Overhead 30% rather than 499%, and the remaining 30% is the eight extra field additions per row, which is arithmetic that has to happen. Both numbers include the second scan the location fix added, so the flattening paid for that too.

As you predicted, this moves the shape errors to the compile step. A bad column index is now reported once for the declaration instead of once at row 0 with nothing said about the other 2^20 rows, and the four affected variants lost their `row` field.

One thing worth telling you: I got this wrong on the first attempt. My first flattening memoized leaves but not interior nodes, so the forty-deep doubling fixture went back to 2^40 paths and took the test process to 43 GB before I killed it. The existing `debug.rs` regression test is what caught it — which bears on the test-location thread.

bus/src/plan.rs
tcoratger: Added `a_shared_operand_graph_is_walked_once_per_node_rather_than_once_per_path` to `plan.rs::tests`, next to the other `BusPlan::build` cases. It builds a forty-deep `dup() + dup()` chain and asserts the build returns; reverting the three lines hangs it.

It also asserts the half nothing covered: the same chain rooted at a `Main { offset: 1 }` still produces `UnsupportedExpression { access: MainOffset(1), .. }`. Skipping repeated nodes must not skip the rejection, and your soundness argument deserves a test rather than only a paragraph.

I kept the `debug.rs` test rather than moving it, and it is no longer a duplicate. `debug.rs` now flattens expressions itself, into a linear op list, and that walk has its own memoization which is not the one in `plan.rs`. It earned its place this round: my first flattening memoized leaves but not interior nodes, and that test was what caught it, by taking the test process to 43 GB.

Your point 1 is in the PR body now, as a paragraph saying which error is reported can change when an expression contains two distinct unsupported accesses.

Agreed on the cross-PR note, and it is an argument for landing it visibly rather than quietly, which is what the body paragraph does.

bus/src/debug.rs
tcoratger: Gone, and by construction rather than by adding an error branch to a row-time function.

Leaf resolution now happens once per declaration at flattening time, before any row is read. The three unsupported entries return `BusDebugError::Plan(BusPlanError::UnsupportedExpression { .. })` carrying the expression location and the access, which is the shape you sketched — the variant is named `UnsupportedExpression` rather than `UnsupportedAccess`, but every piece was already public as you said.

The two stack-machine `expect`s you said to leave are still there, and they are now genuinely internal: the flattened program is built and consumed inside the same function, so neither depends on a sibling function having run first. The cross-module invariant you objected to no longer exists.

`an_unsupported_leaf_is_named_rather_than_panicked_on` covers a next-row access and a periodic access directly at the leaf resolver, since `BusPlan::build` still rejects both before a caller could reach this path through `check`.

bus/src/debug.rs
tcoratger: Documented and tested, deliberately not filtered.

Two sentences in the `check` documentation say that an unconditional declaration records its padding tuple on every padding row, and that this is a real unbalanced contribution rather than an artefact, so the fix is a selector in the AIR.

`padding_rows_contribute_like_any_other_row` is your fixture: a height-4 pull table with two real rows and two zero rows against a two-row push table, asserting the zero tuple appears with pull count 2 and names rows 2 and 3.

On the choice: the replay is faithful, as you say, and a debugger that hid a contribution the in-circuit argument actually makes would be lying about the one thing it exists to show. The remedy for the leading zeros entry is that it is now one line of rendered text rather than eleven, so it costs the reader a glance instead of a screen.

bus/src/debug.rs
tcoratger: Written, essentially as you wrote it. `proptest` is a dev-dependency now, one line.

Both load-bearing assertions are in: `locations[0].row == row` on the pushing side and on the pulling side, and `imbalances.len() == 2`. I added `locations[0].air` on both sides as well, which only became a meaningful assertion once the sample stopped naming matched occurrences.

Both deterministic gaps are closed too.

`a_fault_in_the_middle_instance_is_attributed_to_that_instance` runs three instances with the fault in the middle one, asserting the lost payload names AIR 0 and the surplus payload names AIR 1. That is the off-by-one you wanted pinned.

`two_declarations_producing_one_tuple_are_told_apart` puts two declarations on the same bus producing the same tuple from the same row, asserting the sample carries declaration 0 and declaration 1 at row 0. Without that, `declaration` was only ever 0.

bus/src/debug.rs
tcoratger: Done. The row loop is skipped for an instance with no declarations, and the guard sits before the per-declaration flattening rather than inside the inner loop, so such an instance now costs nothing at all rather than 2^24 evaluator constructions.

`an_instance_without_declarations_contributes_nothing` puts a silent tall AIR in front of a balanced pair and asserts both that the statement balances and that the AIR positions of the others are unchanged.

bus/src/debug.rs
tcoratger: Added to all four.

Your prediction about which field would arrive first was right: the truncation count landed this round, on the report type.

I left the attribute off the new limits type. That one is an input a caller constructs, and `#[non_exhaustive]` blocks the functional-update form from outside the crate, which would make it unusable for the thing it exists for.

bus/src/debug.rs
tcoratger: Both halves.

The two defaults are public associated constants on a limits type, and `check_with_limits` takes that type, so a caller can raise either the per-direction sample or the tuple cap for a single run. `check` is now a one-line wrapper over it. That covers your "64 samples for one run" case and, separately, the report cap from the unbounded-output thread.

I did not reference the constant from the rustdoc prose. The repository's documentation rules forbid naming an item in prose, including intra-doc links, so the `check` documentation states the behaviour and the number is discoverable on the limits type where a caller can also branch on it.

bus/src/debug.rs
tcoratger: Hand-written, printing shapes, essentially your three lines. The struct still derives `Clone` and `Copy`, so a reflex `dbg!` now yields the trace dimensions, the fixed-column count, the public-input count and the declaration count rather than the committed values.

bus/src/debug.rs
```diff
@@ -0,0 +1,1969 @@
...
+    /// Combine the operands a scheduled node's children already produced.
+    Leave(Combine, &'a SymbolicExpression<F>),
+}
+
+fn compile_instance<'a, F: Field>(
+    air: usize,
+    instance: &BusDebugInstance<'a, F>,
+    domains: &HashMap<&str, usize>,
+) -> Result<Vec<CompiledDeclaration<'a, F>>, BusDebugError> {
+    let main = instance.main.iter_polys().collect::<Vec<_>>();
```
Nashtare: On current main `Table` can hold packed Boolean storage (#2227), and `iter_polys()` panics on it (`dense table access is unavailable for packed Boolean storage`). `num_polys()` and `num_variables()` work on packed tables, so `new` accepts a correctly shaped one and `check` then panics here on the first instance with a declaration. The same goes for the preprocessed table a few lines down. This branch predates #2227, so the merge with main is textually clean and still compiles, and no test builds a packed table, so CI won't catch it. Packed tables are what the Boolean-committed binary prover keeps, which is the pipeline this bus targets.

Could you rebase and lower columns through `Table::columns()` instead? `ColumnView` is `Copy` and `value(row)` decodes either storage, so `Op::Column(ColumnView<'a, F>)` would handle both. Alternatively, reject packed storage in `new` with a typed error. A test with `Table::from_packed_bits` would pin it either way.

bus/src/debug.rs
```diff
@@ -0,0 +1,1969 @@
...
+                continue;
+            }
+            printed.push(imbalance.bus_name.as_str());
+            let group = self
+                .imbalances
+                .iter()
+                .filter(|other| other.bus_name == imbalance.bus_name);
+            writeln!(
+                f,
+                "bus {:?}: {} unmatched tuples",
```
Nashtare, line 244: When the output is truncated, this count covers only the retained entries. The capped test renders `bus "order": 3 unmatched tuples` although that bus has 8, followed by `... and 5 more unmatched tuples` without saying where they are. The cap is by first occurrence across the whole statement, so a bus whose unmatched tuples all fall past it doesn't appear in the text at all, even though each bus balances independently. Could `BusDebugReport` keep a dropped count per bus (a `Vec<usize>` indexed by the plan's bus, which `#[non_exhaustive]` already leaves room for)? Then this line could read `8 unmatched tuples (3 shown)` and name the buses that have no retained entry.

bus/src/debug.rs
```diff
@@ -0,0 +1,1969 @@
...
+pub struct BusDebugOccurrence {
+    /// Number of active rows carrying the tuple.
+    pub count: usize,
+    /// Sources of the occurrences this side holds in excess of the other side.
+    ///
+    /// The two sides are paired in traversal order, so this sample starts at the first unpaired occurrence.
+    ///
+    /// The smaller side is fully paired and therefore carries no sample at all.
+    ///
+    /// Removing the sampled rows restores balance for this tuple, but when both sides are non-empty no single row is the unique culprit.
```
Nashtare: This holds only when the sample carries the whole excess. In `the_source_sample_is_bounded_while_the_count_stays_exact` (8 pushes, no pulls, 4 sampled), removing the sampled rows still leaves 4 unmatched pushes. Also, `count` (line 147) counts declaration occurrences rather than rows: `two_declarations_producing_one_tuple_are_told_apart` gets `count == 2` from a single row. Could both say "occurrences", and this one add "when `locations` holds the whole excess"? The PR description has the same sentence.

bus/src/debug.rs
```diff
@@ -0,0 +1,1969 @@
...
+    /// Sum of two earlier nodes.
+    Add(usize, usize),
+    /// Difference of two earlier nodes.
+    Sub(usize, usize),
+    /// Product of two earlier nodes.
+    Mul(usize, usize),
+}
+
+/// An expression flattened into an order where every operand precedes its use.
+struct Program<'a, F>(Vec<Op<'a, F>>);
```
Nashtare: Once #2244 lands, the bus has two evaluators of the same `SymbolicBusInteraction`s: this `Program` and #2244's `BusEvaluation::evaluate`, which the prover runs per row, `check_tables` dry-runs and the verifier runs at the terminal point. I checked that they agree today: the same three row-selector expressions, `1 + (f - 1)s` at Boolean `s` is the same as skip/count, and both are current-row only. But #2244's is a plain recursive walk. Locally I ran the doubling chain from `shared_operands_are_replayed_once_per_node_rather_than_once_per_path` against #2244's head: one row of `BusFactorPlan::evaluate` took 0.30 ms at depth 16, 4.2 ms at 20 and 56 ms at 24, doubling per level. That is roughly 14 s per row at depth 32, and the depth-40 fixture here would not finish.

Would it make sense for the evaluation module to own this compiled per-node form, with the prover, the verifier and this debugger all running it? p3-air already has a private post-order lowering in `symbolic/flatten.rs` that could be exposed for this. Not blocking here; it's for whichever of the two lands second.

bus/src/debug.rs
```diff
@@ -0,0 +1,1969 @@
...
+
+impl<F: Field> Program<'_, F> {
+    fn run(&self, row: usize, height: usize, slots: &mut Vec<F>) -> F {
+        // A linear scan over shared storage replaces a per-row hash table.
```
Nashtare: Nit: nothing in the code has a per-row hash table any more, so this reads as change narration; could it just describe the scan?

Nashtare (changes requested):
I re-read `debug.rs` against `leaf.rs` and `plan.rs` at `30b9b71`, and against the row loop in #2244. The per-node walk holds. `Compiler::compile` lowers every node once: a shared operand finishes in the left subtree and hits `done` on the right, so a program is about two ops per doubling. `validate_expression` gets the same `seen` set, and the new `plan.rs` test pins both acceptance and the buried `MainOffset(1)` rejection. The unmatched sample is exact: the second pass samples ordinals from `min(p, q)` on the larger side, in `(air, row, declaration)` order. Truncation is visible through `unreported`, `is_truncated()` and the trailing line. Replay still matches `leaf.rs`. The three row selectors are also the exact expressions #2244's `materialize_block` uses.

The shape checks in `new` with the transposed-trace test, the `from_air` end-to-end test, the per-node walk, the `BinaryField128`/cap/preprocessed/selector tests and the borrowed probe close all five earlier threads.

The merge with main is textually clean. With #2244 it conflicts only in `lib.rs` and the README, both mechanical.

- After merging with main, `check` panics on a packed Boolean `Table`, which `new` accepts and the binary prover produces (inline, `debug.rs`).

**Smaller things**
- Under truncation the per-bus header undercounts, and buses whose tuples were all dropped are never named (inline, `debug.rs`).
- The `BusDebugOccurrence` docs overclaim once the sample is capped (inline, `debug.rs`).
- A note on sharing this evaluator with #2244's `BusEvaluation`, which is still path-wise (inline, `debug.rs`).

bus/src/debug.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r1.md

bus/src/debug.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2.md

bus/src/debug.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r3.md

bus/src/debug.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r4.md

bus/src/debug.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r5.md

bus/src/debug.rs
tcoratger: Correction to my previous reply: the three "factual corrections" in it are wrong, and you were right.

I compared against `origin/feat/bus-terminal-binding`, which is a stale copy on the fork at `8e4277ff`. The actual head of #2244 is `7b9fbd5f` on this repository. `BusFactorPlan`, `compile_factor` and `check_tables` all exist there — in `bus/src/plan/mod.rs`, `bus/src/evaluation/mod.rs`, `multi-stark/src/bus/context.rs` and elsewhere. My `git grep` came back empty because it searched the wrong tree, not because the names were absent.

The bad remote came from the instructions I was given, but I should have checked the head against the PR rather than trusting a branch name.

What survives from that reply, because it did not depend on the stale tree:

- The exponential-walk timing, measured in a scratch worktree in release on BabyBear: 0.226 / 0.919 / 3.682 / 14.740 / 58.818 ms at depths 16/18/20/22/24, a clean 2x per level. Your figures land within about 25% at every depth. A debug build is 3.3x higher and does not match, so those were release numbers.
- The duplicate-evaluator concern, and that nothing currently pins the two implementations equal.
- `air/src/symbolic/flatten.rs` already produces the dedup'd post-order lowering both sides need, but the module and its items are private, so sharing needs them exposed first.
- The commitment to do the unification in whichever of the two pull requests lands second.

I am re-checking the ownership question against the correct head and will follow up if the answer changes.

Nashtare (approved):
I re-read the new commit `7c8c9e0` in full, and the four rebased commits are unchanged from `30b9b71`. I also type-checked `p3-bus --all-targets` on this branch merged into current main. It builds clean, with no warnings.

The packed-table fix holds. `compile_instance` now takes `columns()` for both the main and the preprocessed table, and `Program::run` reads `ColumnView::value(row)`. That is an O(1) bit decode with no dense copy, and it follows the `from_packed_bits` layout: bit `row % 64` of word `(row / 64) * width + column`. Nothing in `debug.rs` touches `iter_polys` or `poly` any more, and `num_polys()`/`num_variables()` work on either storage. The height check runs before either scan, so the `row < len` assert in `value` can't fire from caller data.

The new test reads the activation from a packed preprocessed column. It would panic if either table went back through the dense accessor, and its `packed == dense` comparison catches a bit-order slip. I decoded 77/150/107 by hand and the asserted tuples match. Heights past one word are covered by the word-boundary test in `sumcheck`.

The per-bus totals are exact: they're counted before the cap and the headers walk them. `8 unmatched tuples (3 shown)` and the `"zulu": 2 unmatched tuples (0 shown)` line are both pinned, and both tests fail if the increment moves after the cap. With the occurrence wording on `BusDebugOccurrence` and in the description, and the rewritten scan comment, that closes all of my previous threads.

On the evaluator note: #2244's `e3607a2` now lowers declarations per node as well, so the time concern is gone on both sides. What's left is de-duplicating the two lowerings, which you've said the second of the two to land will take on.

PR #2224, opened 2026-09-19
feat(bus): add offline read-only memory checking

tcoratger:
## Summary

Add a backend primitive for offline read-only memory checking over the binary-native bus:

- paired read interactions `(a, c, v)` pulled and `(a, g c, v)` pushed;
- verifier-derived seed and finalization boundaries at addresses `g^i`;
- an inverse-constrained read count, so a declaration is sound under any bus reduction;
- an independent nonzero product over every pre-read count on the materialized path;
- one three-tree radix-four product-GKR reduction;
- exact generator-orbit bounds and composed security accounting;
- typed terminal claims that remain explicitly unauthenticated.

Direction stays structural metadata. No field sign is used to distinguish pushes from pulls, so the construction remains valid in characteristic two.

This is independent of the binary-WHIR work.

## Statement

For a table `T[0..m]`, entry `i` is seeded as `(g^i, 1, T[i])`. A read pulls `(a, c, v)` and pushes `(a, g c, v)`. Finalization pulls `(g^i, c_i, T[i])`.

Subject to the assumptions below, equality of the push and pull multisets implies that every read pair `(a, v)` belongs to the seeded table.

The implementation reduces this statement to three product trees:

1. push factors;
2. pull factors, sharing the first root;
3. pre-read counts, whose root must be nonzero.

## Soundness

The deterministic argument requires:

- `g` to generate the field's multiplicative group;
- distinct seeded addresses, enforced by `m <= ord(g)`;
- fewer than `ord(g)` reads, preventing a forged count orbit from wrapping;
- every pre-read count to be nonzero.

The last two are what make a zero count fatal: at count zero, `g * 0 = 0`, so an invalid read cancels itself between the two bus directions, and a full orbit of nonzero counts does the same.

Both are enforced rather than documented, on both paths.

A declared read carries a count inverse and the helper constrains their product to one, so a zero count cannot be declared. A read is declared through a named-array handle that refuses any field whose count orbit a machine-word row count could span. Declarations are therefore sound under any reduction that proves the bus balance, including the two-tree one a whole bus plan derives. The materialized path has no AIR constraints, so its third product tree rejects a zero count in their place.

Fingerprint challenges are drawn inside the reduction, after the statement's dimensions are bound, so the ordering is not a caller obligation.

Seeding and finalizing the array remains the declaring AIR's job: each entry must sit at its own verifier-derived generator power, since an all-equal committed address column would make the array multiset-valued.

The statistical report includes tuple compression and the exact three-tree product-GKR schedule. Commitment binding and authentication of the returned terminal claims are deliberately excluded; integration with the table sumcheck/PCS is still required before this becomes an end-to-end memory argument.

## API

- `ReadOnlyMemoryBus` names one array and checks its field against the forged-cycle bound before any AIR can declare a read on it.
- `ReadOnlyMemoryInteractionBuilder` emits one paired, direction-aware AIR read transition and constrains the count against its inverse.
- `ReadOnlyMemoryPlan` derives tuple layout, orbit bounds, product shape, and security dimensions from a checked `BusPlan`, and owns the whole prove/verify challenge schedule.
- `ReadOnlyMemoryChallenges` carries the fingerprint point and offset the reduction drew.
- `ReadOnlyMemoryClaims` marks the three leaf evaluations still owed to committed columns.

Leaf construction, the product shape and the statement binding are crate-private, so no caller can reach the product prover or verifier while skipping the nonzero-count root check.

## Tests

- honest two-entry, three-read reduction through product GKR;
- invalid returned value breaks bus balance;
- concrete zero-count self-cancellation attack is rejected;
- full-orbit address boundary and strict read-count boundary;
- malformed witness and reduction dimensions;
- paired AIR direction/count transition and the emitted count-inverse constraint;
- a zero-count row refused by the AIR constraint check;
- a field whose count orbit is reachable refused at declaration time;
- exact three-tree security regression.

## Validation

- `cargo test -p p3-bus` (50 unit tests and 1 doctest passed)
- `cargo clippy -q -p p3-bus --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -q -p p3-bus --no-deps`
- `git diff --check`

## References

- leanVM, bus interactions and offline memory checking
- Georg Wiese, *Offline Memory Checking*

---

Part of #2235.


bus/src/memory/mod.rs
```diff
@@ -0,0 +1,576 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// Authenticated claims still owed after the product reduction.
+#[must_use = "leaf claims must be tied to committed columns"]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ReadOnlyMemoryClaims<EF> {
```
Nashtare, line 244: The three trees have **different** non-padding prefixes — `table_len + read_len` for pushes and pulls, `read_len` for counts — so read `j`'s count sits at index `table_len + j` in the pull tree but at index `j` in the count tree. All three `values` are evaluations of identity-padded tables at one shared `point`, so authenticating them needs each tree's prefix and the shift between them, and `ReadOnlyMemoryClaims` carries neither. `table_len` is not constrained to a power of two either, so the shift is not even a subcube selection (`table_len = 2, read_len = 3, h = 3`: pull reads live at `{2,3,4}`, count leaves at `{0,1,2}`).

The only prefix a composer can reach is `plan.table_len() + plan.read_len()`. Applying that to all three claims accepts a `counts` value unrelated to the `read_counts` column, at which point the nonzero-count tree constrains nothing about the counts that actually appear in the pull factors — and the exact attack `zero_count_rejects_a_self_cancelling_invalid_read` demonstrates becomes live again.

Could the plan expose the per-tree prefix lengths (and the count tree's offset) next to the claims, so the obligation is mechanical rather than reconstructed?

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,576 @@
...
+        };
+
+        // Both bus sides contain one boundary row per entry and one row per read.
+        let factor_count = self.table_len + self.read_len;
+        let mut pushes = Vec::with_capacity(factor_count);
+        let mut pulls = Vec::with_capacity(factor_count);
+        let mut counts = Vec::with_capacity(self.read_len);
+
+        // Seed and finalize each valid pair at matching generator-derived addresses.
+        let mut address = F::ONE;
```
Nashtare, line 489: The seed and finalization addresses are generated here, inside the prover, and only ever leave as opaque leaf values. The stated downstream obligation is "tie leaf claims to committed columns", which does not say that the seed block's address component must be a verifier-recomputed `g^i` rather than a committed column.

Address distinctness is the entire reason a read can match only one table entry. A composer that discharges the obligation literally lets the prover commit `table_addresses = [a, a, ..., a]`; every seed pair becomes `(a, T_i)` for one `a`, the memory turns multiset-valued, a read at `a` may return any `T_i`, the bus still balances and every count is nonzero. The `m <= ord(g)` check then bounds addresses that were never used.

Worth stating explicitly in the `claims()` / `ReadOnlyMemoryClaims` docs that the address coordinate of the seed block is verifier-derived and must not be taken from a commitment.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,576 @@
...
+/// A read consumes its current count and produces the next generator-orbit count.
+/// Both interactions retain structural direction metadata in characteristic two.
+pub trait ReadOnlyMemoryInteractionBuilder: BusInteractionBuilder
+where
+    Self::F: Field,
+{
+    /// Declares one paired read transition on a named array bus.
+    ///
+    /// A conditional activation is constrained to zero or one exactly once.
+    fn read_only_memory(
```
Nashtare, line 109: This helper accepts `BusActivation::Boolean`, so a real trace has inactive read rows — but `ReadOnlyMemoryColumns` has no selector field (unlike `BusLeafDeclaration`'s `BusSelector`) and `materialize` emits a pull factor, a push factor and a count leaf for **every** row of `read_counts` unconditionally.

Two consequences. A naturally zero-padded inactive row drives the honest count product to zero, so `check_products` returns `ZeroCountProduct` and no honest proof exists — the caller has to compact the trace outside the proof system, and the count tree has no `sel*c + (1 - sel)` form that would let inactivity be expressed instead. And the leaves a compacted materializer produces cannot match the block layout `BusPlan::terminal_shares` fixes for AIR-declared interactions, so the AIR half and the materializer half of this PR describe two different leaf layouts.

Either drop `BusActivation::Boolean` from this signature for now, or give `ReadOnlyMemoryColumns` a selector and make the count leaf selector-aware.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,576 @@
...
+        }
+        if output.point.len() != self.product_shape.log_height() {
+            return Err(ReadOnlyMemoryError::ReductionPointDimensionMismatch {
+                expected: self.product_shape.log_height(),
+                actual: output.point.len(),
+            });
+        }
+
+        // Root sharing is structural in a verified reduction.
+        if output.roots[0] != output.roots[1] {
```
Nashtare, line 661: Nit: under this plan's own shape this branch is dead. `ReadOnlyMemoryPlan::new` hardcodes `ProductGkrRootShape::FirstTwoShared`, and `FirstTwoShared::decode` always emits `[e0, e0, e1]`, so any output verified against `self.product_shape()` satisfies `roots[0] == roots[1]` by construction. Balance is really enforced by the layer relations plus the single encoded root.

No hole today, but this line reads as the place balance is checked, and no test covers it — `tests.rs` hand-builds `roots: vec![ONE, ONE, ZERO]`, which passes it. A switch to `Distinct` would silently move the real enforcement. Either assert the shape here or say in the comment that the check is defensive against an output produced under a different shape.

bus/src/memory/tests.rs
```diff
@@ -0,0 +1,380 @@
...
+        values: vec![F::ONE; 3],
+    };
+    assert_eq!(
+        plan.claims(zero_count),
+        Err(ReadOnlyMemoryError::ZeroCountProduct)
+    );
+}
+
+#[test]
+fn security_uses_the_three_tree_schedule() {
```
Nashtare, line 702: Despite the name, this pins one of the four terms: the batching numerator (126.0) and the combined term's *label*. It never asserts the combined bit count, nor the fingerprint term — which is the only component that depends on the memory-specific input `[table_len + read_len; 2]`.

Changing that input to `[read_len, read_len]` (dropping the seed and finalization factors from the union bound, understating the fingerprint error by ~1.3 bits here and more at real sizes), or dropping the fingerprint term entirely, leaves this test green. Asserting `combined_term().bits` would close it.

bus/src/memory/tests.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    // The adapter checks roots but leaves commitment authentication to composition.
+    let claims = plan
+        .claims(verifier_output)
+        .expect("balanced roots and nonzero counts satisfy offline memory");
+    assert_eq!(claims.point.len(), plan.product_shape().log_height());
+}
+
+#[test]
+fn air_helper_emits_one_paired_count_transition() {
```
Nashtare, line 261: Nit: this checks direction order, field counts, the single Booleanity constraint and that `interactions[1].fields[1]` is a `Mul`, but never that `fields[0]` and `fields[2]` are the *same* expressions on the pull and push sides. A regression where the push side reads a different column for the address or the value would leave every honest trace unbalanced, and this is the test that exists to pin the helper's shape.

(The `Mul` assertion does rule out a `ZERO` or `ONE` multiplier, since `SymbolicExpression` folds those. It does not pin `GENERATOR`, though in characteristic two any `g^k` with odd `k` is also full-order, so that part is benign.)

Nashtare (commented):
I checked the construction against what the code actually enforces, recomputed the security accounting independently, and looked for the boundary attacks the bounds are meant to exclude.

**The argument is right**

- **Both orbit bounds are exact, no off-by-one.** `table_len > orbit` is rejected (so `m <= ord(g)` and the seeded addresses `g^0..g^(m-1)` are distinct) and `read_len >= orbit` is rejected (strict `<`). I confirmed the attack the strict bound excludes: at `read_len = ord(g)` a prover forges an unseeded `(a*, v*)` with counts `1, g, ..., g^(ord-1)`; the pushes `{g, ..., g^ord = 1}` equal the pulls as a multiset and every count is nonzero. The test at `tests.rs:296` pins exactly that boundary.
- **The multiset reasoning holds.** For a non-seeded pair, balance forces `{g*c_j} = {c_j}`, and with every `c_j` nonzero the support is a union of `<g>`-orbits of size `ord(g)`, so fewer than `ord(g)` reads makes forgery impossible. For a seeded pair, `{1} u g*C = {f_i} u C` forces `C = {1, g, ..., g^(k-1)}` and `f_i = g^k`.
- **The nonzero-count tree is genuinely independent.** Its leaves are raw counts, not fingerprints, so they do not depend on the fingerprint challenge; its root is carried separately and bound to `values[2]` by the per-tree layer relations. `zero_count_rejects_a_self_cancelling_invalid_read` is a real negative test, not a shape-check artifact: it first asserts the push and pull products *do* agree for the forged zero-count read, then asserts rejection. Deleting the nonzero check makes it fail.
- **Root sharing is sound.** `FirstTwoShared` sends one root and duplicates it on decode, but each tree's leaves are still bound to its own children at every layer, so one tree's leaves cannot be substituted for the other's.
- **No field sign anywhere.** Direction is structural; the only subtractions are `offset - fingerprint` and `ONE - coordinate`, and nothing relies on `-1 != 1`.
- **Identity padding is `ONE`**, so the short count tree keeps a nonzero product and `check_products`'s unpadded product agrees with the GKR root.
- **Reachable panics: none found.** Both `usize::try_from(orbit).expect(...)` sites are only reached when `orbit` is already bounded by a `usize`, so the value provably fits; every `expect` on `ProductGkrShape::new` / `ProductGkrSecurityProfile::new` satisfies its precondition; and all attacker-controlled lengths in `claims()` are checked before indexing.
- `RecordToken(pub(crate) ())` stays unconstructible outside the crate, and the memory helper does emit its own `bool_check` before minting, so the widening is used correctly. The token's doc ("Proof that the public declaration path emitted its activation constraint") is now slightly off, since the public path is no longer the only minter.

**Security accounting: the number is right**

I recomputed it for the fixture (`table_len = 2`, `read_len = 3`, `payload_width = 3`, so `s = 2`, `factors = 5`, `h = 3`, `L = 2` layers, `R = 1` sumcheck round, `T = 3` trees):

| term | numerator | bits at 2^128 |
| --- | ---: | ---: |
| tuple fingerprint | `max(1,s)*N = 10` | 124.678 |
| product-GKR sumcheck | `5R = 5` | 125.678 |
| tree batching | `L(T-1) = 4` | 126.000 |
| child collapse | `h = 3` | 126.415 |
| combined | 22 | **123.541** |

That matches what the code reports. No round is missing (`sumcheck_rounds` is the sum of per-layer point lengths, which is exactly the number of degree-5 rounds executed), the tree factor is `L*(T-1) = 4` rather than the two-tree `2` (shared roots save a proof element, not a claim), and the fingerprint numerator `max(1,s)*N` is tight — each factor `X - f_i(r)` has total degree `max(1,s)` in `(X, r)`, so the product has total degree `N*max(1,s)`. The count tree is correctly excluded from the fingerprint term.

**What I would settle before this is composed**

The three inline comments on `mod.rs` are all about the hand-off to the layer that will authenticate the terminal claims. None of them is a live hole in what is here — the primitive is explicitly staged — but the first two are the kind of gap that a composer discharges wrongly by following the stated obligation literally, and the third is a mismatch between the two halves of this PR. The cheapest of the three (exposing the per-tree prefix lengths) is worth doing in this PR.

Not blocking.

bus/src/memory/mod.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_1.md

bus/src/memory/mod.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_2.md

bus/src/memory/mod.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_3.md

bus/src/memory/mod.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_4.md

bus/src/memory/tests.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_5.md

bus/src/memory/tests.rs
tcoratger: @/tmp/claude-1000/-home-utilisateur-Documents-Plonky3/623aaae0-bf3c-42b8-9ad7-7916e54ab304/scratchpad/r2224_6.md

tcoratger (commented):
I re-derived the multiset argument from scratch, ran the degenerate cases in a worktree, and looked for what a composer would get wrong. The six earlier inline comments all appear addressed in `6ce434c` (`prefix_lens` / `read_offset` now travel with the claims, the seed-address obligation is documented, `BusActivation::Boolean` is gone from the helper, the defensive-root comment is there, and the security test now pins the fingerprint term as well). I am not repeating those.

## The argument itself is correct

The construction is:

```
push  P = { fp(g^i, 1,     T_i) : i < m }  u  { fp(a_j, g*c_j, v_j) : j < n }
pull  Q = { fp(g^i, f_i,   T_i) : i < m }  u  { fp(a_j,   c_j, v_j) : j < n }
count C = { c_j : j < n },   prod C != 0
```

Balance is `prod P == prod Q`. Group the tuples by the pair `(a, v)` and write `C_(a,v)` for the counts of reads on that pair. Two cases:

* `(a, v)` is not a seeded pair. Then `g * C_(a,v) = C_(a,v)` as multisets. Every element is nonzero (third tree), so the support is a disjoint union of `<g>`-orbits, each of size `ord(g)`. With `n < ord(g)` the multiset must be empty. No forged read survives.
* `(a, v) = (g^i, T_i)`. Then `{1} u g*C = {f_i} u C`. Writing `d_j = C(g^j)`, the equation is `d_j - d_(j-1) = [j = 0] - [j = t]` where `f_i = g^t`. Summing around the orbit forces `f_i in <g>`; telescoping then gives `|C| = ord(g)*d + t`, and `|C| < ord(g)` forces `d = 0`, so `C = {1, g, ..., g^(t-1)}` exactly and `f_i = g^t`. The chain is rigid.

Both orbit bounds are exactly the ones this needs, with no off-by-one: `table_len <= ord(g)` (distinct seeded addresses) and `read_len < ord(g)` (strict, so no forged orbit can close). `f_i = 0` is rejected for free: `0` cannot appear on the push side once every read count is nonzero, since `g * c != 0`. I confirmed that by running it — `UnbalancedProducts`.

The whole thing rests on `F::GENERATOR` having full order. That is the documented `Field` contract, and it is genuinely tested: `field-testing` checks `g^(r/p) != 1` for every prime factor `p` of `|F| - 1`, and `binary-field/src/tower.rs` has `generators_generate_the_multiplicative_group`. So `orbit = F::order() - 1` is the right quantity. Worth a one-line doc pointer, since the entire soundness of `read_len < orbit` collapses to it.

## Verified clean

* No reachable panic. Both `usize::try_from(orbit).expect(..)` sites run only after `orbit` has been bounded by a `usize`. `weights[2..2 + value_width]` is always in bounds because `fingerprint_width >= payload_slots >= payload_width = 2 + value_width`. `claims()` checks every attacker-controlled length before indexing.
* `ProductGkrShape::new(0, 3, FirstTwoShared)` is accepted, and `ProductGkrSecurityProfile::new(0, 3, 0, 0, 0)` passes its `collapse_challenges == log_height` and `(log_height == 0) == (layer_count == 0)` guards. So `table_len = 1, read_len = 0` does not panic in a `Result`-returning constructor. I ran it end to end: `prefix_lens = [1, 1, 0]`, proof verifies.
* Identity padding is `ONE` in `ProductLayers`, so the short count tree keeps a nonzero root and `check_products`'s unpadded product agrees with the GKR root.
* `value_width = 0` (payload width 2) works. Note that `validate_columns` then checks nothing at all about the table, because both per-column height loops are vacuous.
* Product-GKR Fiat-Shamir is solid: `domain_separator` binds `log_height`, `num_trees` and `root_shape` before anything, and `ROOTS` is observed before the first `BATCHING` challenge.
* No field sign anywhere; direction stays structural, so characteristic two is fine.

## Findings

Two MAJOR, both about the hand-off rather than the algebra; three MINOR. Details inline.

The one I would not merge without at least a doc fix is the first: `BusPlan` is two-tree **by construction** (`plan.rs:320`), so the AIR route this PR advertises cannot carry the third tree at all — and that third tree is the only thing standing between a prover and the self-cancelling read the PR's own test demonstrates.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+        // Seeds and reads occupy one factor each on both multiset sides.
+        let factor_count = table_len
+            .checked_add(read_len)
+            .ok_or(ReadOnlyMemoryError::FactorCountOverflow)?;
+        let capacity = factor_count
+            .checked_next_power_of_two()
+            .ok_or(ReadOnlyMemoryError::FactorCountOverflow)?;
+        let log_height = capacity.trailing_zeros() as usize;
+        let product_shape =
+            ProductGkrShape::new(log_height, 3, ProductGkrRootShape::FirstTwoShared)
```
tcoratger, line 367: **[MAJOR] The AIR route structurally cannot carry the count tree, and nothing says so**

`BusPlan::build` hardcodes its own product shape:

```rust
// bus/src/plan.rs:320
let product_shape = ProductGkrShape::new(log_logical_leaf_count, 2, FirstTwoShared)?;
```

Two trees. Not three. There is no slot for a count tree anywhere on the bus-plan path. I measured both on the fixture in `tests.rs`:

```
BusPlan            -> num_trees = 2, log_height = 2
ReadOnlyMemoryPlan -> num_trees = 3, log_height = 3
```

Different statements, different transcripts, no reconciliation in the crate.

So this PR ships two mutually exclusive halves:

```
  AIR half                            prover half
  read_only_memory(..)                ReadOnlyMemoryPlan::materialize(..)
    -> 2n declarations                  -> [ seeds (m) ; reads (n) ] flat
    -> BusPlan blocks, 2 trees          -> 3 trees, own log_height
    -> NO seeds, NO finalization        -> AIR declarations unused
    -> NO count tree                    -> count tree present
```

A composer who takes the advertised API — declare reads with `ReadOnlyMemoryInteractionBuilder`, prove the bus with `BusPlan`'s own shape — gets **no nonzero-count tree**. At that point `g * 0 = 0` makes an invalid read contribute the identical tuple to both directions, which is precisely the attack `zero_count_rejects_a_self_cancelling_invalid_read` pins. The safeguard the PR calls "necessary" is silently absent on that path. The AIR path is also unbalanced for an honest prover (reads push `g*c` and pull `c`, with nothing seeding or finalizing), so the failure mode is "no honest proof exists" rather than "forgery accepted" — but only until someone adds seed and finalization declarations from a table AIR, at which point it becomes the live hole.

Note also that `ReadOnlyMemoryPlan::new` takes `&BusPlan` but reads only `domains()`, `fingerprint_width()` and `tuple_slot()`. It discards `bus_plan.product_shape()` and never cross-checks `read_len` against the push/pull blocks actually declared on that bus, so a plan with `read_len` unrelated to the AIR is accepted without a word.

Minimum fix: say in the module docs and on `ReadOnlyMemoryInteractionBuilder` that the helper's declarations are **not** what `materialize` reduces, that `BusPlan`'s two-tree reduction must not be used for a memory bus, and that the count tree exists only on the `materialize` path. Better: have `ReadOnlyMemoryPlan::new` reject a bus whose declared block counts disagree with `read_len`, so the two halves cannot be silently combined.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+        // This check is therefore defensive against an output produced under a different root shape.
+        if output.roots[0] != output.roots[1] {
+            return Err(ReadOnlyMemoryError::UnbalancedProducts);
+        }
+        if output.roots[2] == EF::ZERO {
+            return Err(ReadOnlyMemoryError::ZeroCountProduct);
+        }
+
+        // Each tree pads a different prefix, so the obligation travels with the claims.
+        let bus_prefix = self.table_len + self.read_len;
```
tcoratger, line 669: **[MAJOR] `table_len` and `read_len` are outside the Fiat-Shamir transcript, and that obligation is not stated**

`ProductGkrShape::domain_separator` binds exactly three things (`transcript.rs:118-130`):

```
instance(log_height) ; instance(num_trees) ; instance(root_shape)
```

`table_len`, `read_len`, `prefix_lens` and `read_offset` are never absorbed. So two different memory statements produce a bit-identical transcript whenever their factor counts round to the same power of two:

```
(table_len, read_len) = (2, 3)  ->  log_height 3, 3 trees, FirstTwoShared
(table_len, read_len) = (4, 1)  ->  log_height 3, 3 trees, FirstTwoShared   <- same separator
```

One product-GKR proof is valid for both. The split into `prefix_lens = [5, 5, read_len]` and `read_offset = table_len` is reconstructed here, purely locally, from the verifier's own plan.

That is not a live hole as long as the composer builds its `ReadOnlyMemoryPlan` from a public statement it fixed itself. But `read_len < ord(g)` and `table_len <= ord(g)` are the *only* reasons a forged read cannot close an orbit, and neither is bound to anything the prover committed to. The moment a composer derives `read_len` from a prover-supplied length — a trace height, a proof field, a padded table size — the bound becomes prover-chosen and the argument is gone.

The documented caller obligations here are "commitment binding" and "leaf-claim authentication". Please add a third: the memory dimensions `table_len` and `read_len` must be fixed by the verifier and bound into the transcript before the fingerprint challenge. Or absorb them directly, which is cheap and removes the obligation entirely.

Related and smaller: `ReadOnlyMemoryClaims` carries no identity of the plan that produced it, so claims from plan A can be authenticated under plan B's layout with no type error.

bus/src/builder.rs
```diff
@@ -38,7 +38,7 @@ pub trait BusInteractionRecorder: AirBuilder {
 
 /// Proof that the public declaration path emitted its activation constraint.
 #[doc(hidden)]
-pub struct RecordToken(());
+pub struct RecordToken(pub(crate) ());
```
tcoratger: **[MINOR] The `RecordToken` widening is unnecessary — I verified the revert builds green**

This change makes `RecordToken` constructible anywhere in `p3-bus`, which falsifies its own doc ("Proof that the public declaration path emitted its activation constraint"). Any future crate-internal caller can now mint a token with `BusActivation::Boolean(sel)` and skip the `bool_check` that `push_bus_interaction` would have emitted. Nothing catches that at compile time any more.

It is not needed. `push_bus_interaction` emits no constraint at all for `BusActivation::Always`, so `read_only_memory` can just call it:

```rust
self.push_bus_interaction(bus_name, BusDirection::Pull, pull, BusActivation::Always);
...
self.push_bus_interaction(bus_name, BusDirection::Push, push, BusActivation::Always);
```

I applied exactly that in a worktree, reverted this line to `pub struct RecordToken(());`, dropped the now-unused `RecordToken` import from `memory/mod.rs`, and ran `cargo test -p p3-bus`: 38 unit tests + 4 integration + 1 doctest, all green. The `RecordToken` invariant stays narrow and this file's diff disappears.

(If a future conditional-read variant needs the token, widen it then, together with the `bool_check` that justifies it.)

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+            self.tuple_variables,
+            [factors, factors],
+            profile,
+        )
+        .expect("a checked memory plan has valid bus security dimensions")
+    }
+}
+
+/// Evaluates the Boolean-cube equality polynomial at every tuple slot.
+fn equality_weights<F: Field>(point: &[F]) -> Vec<F> {
```
tcoratger: **[MINOR] `equality_weights` is a verbatim copy of `leaf.rs:227`**

Byte for byte identical, comments aside. That is not merely duplication: the two functions define the coordinate-to-slot-bit convention, and the memory fingerprint is only interchangeable with a bus fingerprint while both conventions agree. `leaf.rs` documents it ("Coordinate `k` binds bit `k` of the tuple-slot index. Fingerprint coordinates therefore run from the least significant slot bit to the most"); this copy does not, and no test ties the two together.

If someone ever reverses the bit order in one copy, memory tuples stop matching bus tuples for the same logical payload, and the failure surfaces as an unexplained imbalance far downstream.

Make `leaf::equality_weights` `pub(crate)` and call it from here. One definition, one convention.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// Value components of the seeded array.
+    pub table: &'a [&'a [F]],
+    /// Address returned for each read event.
+    pub read_addresses: &'a [F],
+    /// Count before each read event advances it.
+    pub read_counts: &'a [F],
+    /// Value components returned for each read event.
+    pub read_values: &'a [&'a [F]],
+    /// Count reached by each array entry after all reads.
+    pub final_counts: &'a [F],
```
tcoratger: **[MINOR] Three `&[F]` fields of two different heights make a silent swap representable**

`read_addresses`, `read_counts` and `final_counts` are all `&'a [F]`, with heights `read_len`, `read_len` and `table_len`. When `table_len == read_len` — the common case for a small table read once per entry — `validate_columns` cannot distinguish `read_counts` from `final_counts`, so swapping them at the call site type-checks, validates, and produces a *different but still provable* memory statement. Same for `table` and `read_values`.

This struct is the one place a caller hand-assembles the witness, so it is exactly where that mistake will be made. A `#[non_exhaustive]` builder, or newtypes (`ReadCounts<'a, F>`, `FinalCounts<'a, F>`), or at minimum a doc line saying the two count columns are not interchangeable, would close it. Field order helps too: `final_counts` currently sits after three read-indexed columns even though it is table-indexed.

bus/src/memory/tests.rs
```diff
@@ -0,0 +1,486 @@
...
+        leaves.product_inputs()[0].iter().copied().product::<F>(),
+        leaves.product_inputs()[1].iter().copied().product::<F>(),
+    );
+
+    // The independent count product closes this otherwise reachable attack.
+    assert_eq!(
+        leaves.check_products(),
+        Err(ReadOnlyMemoryError::ZeroCountProduct)
+    );
+}
```
tcoratger, line 382: **[MINOR] The negative tests are real, but five attacks the construction must stop are uncovered**

Credit where due: `zero_count_rejects_a_self_cancelling_invalid_read` is a genuine negative test, not a shape check — it first asserts the push and pull products *do* agree, then asserts rejection, so deleting the nonzero check makes it fail. That is the right shape for a memory-argument test.

Missing, in rough order of importance:

1. **Forged address with a nonzero count.** The headline attack — "read a value never written" — is only tested in its zero-count variant. A read at `g^77` with count `1` and any value must be rejected by the orbit argument alone, with the count tree passing. I ran it: `Err(UnbalancedProducts)`. Free to add, and it is the test that actually exercises the `{g*c_j} = {c_j}` reasoning.
2. **`final_counts[i] = 0`.** Nothing forbids it explicitly; it is rejected only because `0` cannot appear on the push side once reads are nonzero. I ran it: `Err(UnbalancedProducts)`. Worth pinning, since it is a consequence of the argument rather than an explicit check.
3. **A replayed read tuple.** Two rows with identical `(a, c, v)`. This is the case the chain rigidity must forbid — `C` must be `{1, g, ..., g^(t-1)}`, so a repeated count is illegal even at a valid address with a valid value.
4. **`read_len = 0` / `table_len = 1`.** `log_height = 0`, zero reduction layers, an empty count tree. I ran it end to end: no panic, `prefix_lens = [1, 1, 0]`, proof verifies. A regression here would be a panic in a `Result`-returning constructor, which is worth a guard.
5. **`value_width = 0`** (payload width 2). Works, but `validate_columns` then checks *nothing* about the table, because both height loops are vacuous. Pin the degenerate shape so that stays deliberate.

A proptest would subsume most of this:

```
m in 1..16, n in 0..16
table: m random values ; schedule: n random indices into 0..m
derive counts and final_counts honestly from the schedule
  -> check_products().is_ok()

then apply exactly ONE mutation from
  { flip a read value, flip a read address, bump a read count,
    bump a final count, swap two read counts, zero a read count }
  -> for at least one of several fixed challenge pairs, check_products().is_err()
```

The "several challenge pairs" quantifier is what `wrong_values_and_missing_entries_break_bus_balance` already uses, and it is the right way to keep a statistical argument deterministic in CI.

bus/src/memory/mod.rs
tcoratger: You are right, and the honest position is that the AIR route was never meant to carry the reduction at all. I have made the code say so and refuse the invalid combination rather than only documenting it.

The helper's declarations exist for two reasons: they make the enclosing plan reserve the tuple layout for the named bus, and they publish how many reads the statement covers. They are not the leaves the memory reduction proves. Seeds, finalization and the count tree exist only on the materialized path, and that path has its own three-tree shape. A whole-plan two-tree reduction over a bus that carries read-only memory is therefore never valid, because it has no slot for the count tree and no seed or finalization factor. The module docs and the trait doc now state exactly that.

For the enforceable half, `ReadOnlyMemoryPlan::new` now sums the declared block heights on that bus for each direction and rejects the statement when either total differs from `read_len`. The helper emits one pull and one push per row, so the two totals are the read count by construction; a bus that already carries seed or finalization declarations of its own no longer type-checks against a memory statement either. New variant `DeclaredReadCountMismatch`, new test `the_statement_must_cover_exactly_the_declared_reads`.

One consequence worth flagging, because it changes what your item 4 in the test thread asked for: `read_len = 0` is now unreachable. A bus exists in the plan only because some AIR declared on it, so its declared totals are never zero, and `ReadOnlyMemoryPlan::new(bus, name, 1, 0)` returns `DeclaredReadCountMismatch` instead of building a `log_height = 0` statement. That is still a `Result`, not a panic, and the test pins it. The smallest reachable statement is now one entry and one read, covered end to end by `the_smallest_statement_reduces_without_panicking`.

The fixtures moved with this: each test now builds a bus whose declared row count matches the statement it is about.

bus/src/memory/mod.rs
tcoratger: Bound, and the residual obligation is now stated rather than implied.

Your collision is real: the product separator absorbs only the log height, the tree count and the root shape, so `(3, 2)` and `(4, 4)` share a bit-identical seed. `ReadOnlyMemoryPlan::observe_statement` now seeds a transcript from a memory-specific domain separator whose instance label carries the table length, the read length, the value width and the tuple variable count, each as a length-delimited little-endian chunk. Field-element absorption would not have worked here: the statement lives in characteristic two, so `F::from_usize(2)` and `F::from_usize(4)` are both zero. Bytes are the only injective encoding available.

Two new entry points, `ReadOnlyMemoryPlan::prove` and `ReadOnlyMemoryPlan::verify`, call it before touching the product transcript, and `claims` is now private so the reduction cannot be driven around them. `the_memory_dimensions_separate_two_statements_of_one_product_shape` asserts the two plans really do share a product shape, then proves under the small one and shows the proof fails under the large one. I checked it is not vacuous: deleting the `separator.seed(challenger)` line makes it the only failing test in the suite.

What that does *not* buy is the ordering you asked for. The fingerprint challenge is sampled outside this crate — `materialize` receives the point and offset as arguments — so no binding inside the crate can precede it. Absorbing at reduction time separates the reduction but leaves the fingerprint challenge shared between two dimension choices, which is exactly the case where `read_len < ord(g)` stops being prover-independent. So the third caller obligation is now written down on `security_term` and on `observe_statement`: the memory dimensions must be verifier-fixed and bound before the fingerprint challenge is drawn, and the same public method is the way to do it.

On claims identity: making `claims` private means a `ReadOnlyMemoryClaims` can now only come out of the plan that verified it. That does not stop a composer authenticating it against another plan's committed columns, so I have not claimed it is closed — the prefixes and read offset it carries are its own, and the authentication obligation on the type says they must be the ones used.

bus/src/builder.rs
tcoratger: Reverted. `RecordToken` is back to `pub struct RecordToken(())` and `read_only_memory` issues both directions through `push_bus_interaction` with `BusActivation::Always`, which emits no constraint, so nothing is lost. The unused import is gone and this file's diff against main is now empty. Reproduced your build check: `cargo test -p p3-bus` is green, and the symbolic test still sees zero base constraints from an unconditional read.

bus/src/memory/mod.rs
tcoratger: Shared. `leaf::equality_weights` is now `pub(crate)` and the memory module calls it; the copy is deleted. The coordinate convention moved onto that one function, so the definition and the statement of the convention now sit together. The public doc on `BusLeaves::materialize` keeps its own sentence about it, folded into one line to hold the file's comment density where it was.

bus/src/memory/mod.rs
tcoratger: Made unrepresentable. Every column now has its own wrapper: `MemoryTableValues`, `MemoryFinalCounts`, `MemoryReadAddresses`, `MemoryReadCounts`, `MemoryReadValues`. Swapping the two count columns, or the two value-column sets, is a type error regardless of whether the heights coincide, so the check no longer depends on `table_len != read_len`.

I went one wrapper further than you asked for. `read_addresses` and `read_counts` are the same type and the same height in every statement, not just the coincident ones, so they were the one swap that was *always* representable; they are separated too.

Field order is fixed as you suggested: the two table-indexed columns come first, then the three read-indexed ones.

bus/src/memory/tests.rs
tcoratger: All five added, plus the randomized pass. Each negative test asserts its precondition before asserting rejection, and the ones that depend on tuple compression quantify over the same fixed challenge pairs the existing test uses.

1. `a_forged_address_with_a_nonzero_count_breaks_the_orbit_argument` — read at `g^77` with count one, and it asserts the count product is nonzero before showing the two sides do not balance. This is the one that exercises the orbit reasoning rather than the count tree.
2. `a_zero_final_count_cannot_appear_on_the_push_side` — honest fixture with one finalization at zero, count product asserted nonzero first. Pins that it is a consequence of the argument, not an explicit check.
3. `a_replayed_read_tuple_breaks_chain_rigidity` — two rows of one entry replaying count one, count product asserted nonzero first.
4. The degenerate statement changed shape because of the declared-read cross-check from the first thread: `read_len = 0` is now unreachable, so `the_statement_must_cover_exactly_the_declared_reads` pins that the constructor refuses it with an error rather than panicking, and `the_smallest_statement_reduces_without_panicking` takes one entry and one read end to end and checks `prefix_lens == [2, 2, 1]`.
5. `a_valueless_array_keeps_its_address_and_count_columns` — payload width two, `value_width == 0`, proved and verified, and it also shows a metadata column of the wrong height is still rejected when both value loops are vacuous.

For the property test, `proptest` is not a dependency here, so `one_mutation_of_a_random_honest_schedule_is_always_caught` uses the seeded `Xoroshiro128Plus` this crate already uses elsewhere: twenty-four trials over random table sizes and declared read counts, an honestly derived schedule asserted balanced first, then exactly one mutation from your list, rejected at one of six challenge pairs.

One correction to the mutation list. "Swap two read counts" is not always an attack: the two sides are multisets, so exchanging two counts between reads at the *same* address is a row permutation and still balances. Only a swap across two different addresses breaks the chain. `swapping_two_read_counts_across_addresses_is_caught` pins both halves of that — it asserts the same-address permutation still verifies, then shows the cross-address move is rejected — and the randomized test leaves that mutation out for the same reason.

Nashtare (commented):
I re-checked the argument at `c214c080` against what the code enforces. The first three commits are a pure rebase of what I reviewed, so this pass covers the two fix commits plus a fresh adversarial read.

The algebra still holds. The orbit bounds are unchanged and exact: `table_len > orbit` and `read_len >= orbit` are rejected. `orbit` comes from `F::order()` under the `Field::GENERATOR` contract, which the field tests check against the factorization of `|F| - 1`, so it is never taken from the caller. The count tree's leaves are still raw counts, independent of both fingerprint challenges, and its root is checked for zero. Reads still push `g * c`, and the multiset argument goes through unchanged. For the new fixture I recomputed the security numerators as 10 + 5 + 4 + 3 = 22, which gives the 123.54 bits the test now pins. There is no denominator anywhere, so a factor that hits zero falls under the same Schwartz-Zippel event. `observe_statement` encodes the dimensions as bytes rather than `F::from_usize`, which is the right call in characteristic two. I found no verifier panic on the new `verify` path.

The prefix lengths and read offset on the claims, the verifier-derived address doc, dropping `BusActivation::Boolean`, the defensive-root comment, and the pinned fingerprint and combined bits close all six of my earlier threads. The `RecordToken` revert takes care of the doc remark from the body. (Your replies on those six threads came through as bare local file paths rather than text, so I judged each one from the code.)

**What I'd like settled before this merges**
- Once #2244 lands, multi-stark proves the two-tree bus argument over every declared bus. The public read helper then has no sound home there: adding seeds and finalization as ordinary declarations reopens the zero-count self-cancellation (inline, `memory/mod.rs`).

**Smaller things**
- The orbit test no longer pins the strict read-count bound (inline, `memory/tests.rs`).
- The three-tree reduction can still be driven around `prove`/`verify`, and the fingerprint-before-dimensions ordering is only documented (inline, `memory/mod.rs`).
- For whichever of this and #2244 lands second: #2244 deletes `leaf.rs`, including the `equality_weights` imported here. Moving the memory fingerprint to its MSB-first `equality_weights_msb`, rather than keeping a private copy, keeps the "one convention" doc true.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+        // Retain value expressions once so both directions use identical payloads.
+        let values = values.into_iter().collect::<Vec<_>>();
+        let pull = core::iter::once(address.dup())
+            .chain(core::iter::once(count.dup()))
+            .chain(values.iter().map(Dup::dup));
+        self.push_bus_interaction(bus_name, BusDirection::Pull, pull, BusActivation::Always);
+
+        // Multiplying by the full-order generator advances one logical count.
+        let push = core::iter::once(address)
+            .chain(core::iter::once(count * Self::F::GENERATOR))
```
Nashtare, line 129: The module doc says a memory bus must be left out of the whole-plan two-tree reduction, but #2244's `BusContext::build` feeds every interaction the AIRs declare into `BusPlan::build` and proves them with the plan's two-tree shape, with no way to leave a bus out. So once both land, an AIR calling this helper in multi-stark gets an unbalanced bus. The obvious fix, a table AIR declaring the seed push `(g^i, 1, T_i)` and the finalization pull `(g^i, f_i, T_i)` with plain `push_bus_interaction`, balances honest proofs and also accepts any read `(a*, 0, v*)`. Here `g * 0 = 0`, so its pull and push are the same tuple, which is exactly the case `zero_count_rejects_a_self_cancelling_invalid_read` asserts. On that path nothing checks counts for zero, and nothing checks the orbit bounds either (`ReadOnlyMemoryPlan::new` never runs). That's moot at 128 bits, but this helper is generic over any `Field`, and for BabyBear `ord(g)` is about `2^31`.

Could the helper be made safe in any composition, e.g. by taking a `count_inverse` expression and asserting `count * count_inverse = 1`, or else be kept off the public builder surface until the three-tree path is wired into multi-stark? Whichever of the two PRs lands second seems the natural place for it.

bus/src/memory/tests.rs
```diff
@@ -0,0 +1,916 @@
...
+        Err(ReadOnlyMemoryError::AddressOrbitTooShort {
+            table_len: 256,
+            orbit_len: 255,
+        })
+    );
+
+    // A read multiset of that size could contain one complete forged orbit.
+    let wide_bus = bus_plan::<Rijndael8b>(3, 8);
+    assert_eq!(
+        ReadOnlyMemoryPlan::<Rijndael8b>::new(&wide_bus, "memory", 1, 256),
```
Nashtare: Before the declared-read cross-check this asserted `read_len = 255` is refused, which pinned the strict bound. `256` doesn't pin it: changing `read_len >= orbit` to `>` keeps this test green, and at 255 that mutation admits the forged full orbit `1, g, ..., g^254`. A bus that declares 255 rows is easy to build, e.g. one AIR at `log_height = 0` with 255 push/pull pairs. Could the test assert that 255 is refused and 254 accepted?

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+    /// Read counts whose product must not vanish.
+    counts: Vec<EF>,
+}
+
+impl<EF: Field> ReadOnlyMemoryLeaves<EF> {
+    /// Returns the three product inputs in their protocol order.
+    ///
+    /// The first two roots must be equal and the final root must be nonzero.
+    ///
+    /// Callers should check both before invoking a prover that assumes shared roots.
```
Nashtare: This still describes a caller driving the prover itself, and that path is still open: `product_shape()` is public, so `ProductGkrProof::verify(plan.product_shape(), ..)` skips both `observe_statement` and the `roots[2] != 0` check, which is the only thing rejecting a zero-count read. On the prover side, `prove` accepts leaves from any plan. Leaves of lengths `[5, 5, 2]` passed to a one-entry, one-read plan (capacity 2) hit the `input.len() <= capacity` assert in `ProductGkrProof::prove` and panic, even though `prove` returns a `Result`. Could `product_inputs` and `product_shape` be `pub(crate)`, and `prove` check the three leaf lengths against `table_len + read_len` and `read_len` up front?

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+    pub const fn product_shape(&self) -> ProductGkrShape {
+        // The root encoding forces equality of the first two trees.
+        self.product_shape
+    }
+
+    /// Binds every public dimension of this statement into a transcript.
+    ///
+    /// The reduction transcript otherwise separates only on the padded tree height, so two statements whose factor counts round to the same power of two would share it.
+    ///
+    /// Both reduction entry points call this, and a composer must call it once more before sampling the fingerprint challenge that is drawn outside this crate.
```
Nashtare: `prove`/`verify` bind the dimensions before the product reduction, but the fingerprint point and offset still reach `materialize` as plain arguments, so binding before *those* are drawn is only a documented obligation. #2244's `BusPlan::prove` shows a mechanical form: it seeds the plan, samples `BusChallenges` inside the crate, and only then calls the materializer. Would a `sample_fingerprint` on the plan (observe the statement, then draw `tuple_variables` coordinates plus the offset), or a closure-taking `materialize` in the same style, be worth doing so the ordering can't be skipped? Non-blocking.

bus/src/memory/mod.rs
Nashtare: One precision on "a bus that already carries seed or finalization declarations of its own no longer type-checks": the check sums declared rows per direction and never looks at a declaration's shape. A reader AIR with 2 rows plus a table AIR declaring 2 seed and finalization rows is refused for `read_len = 2` but accepted for `read_len = 4`, and the test fixture's own `bus_plan` declares a plain push/pull pair of identical fields and passes. That's harmless on the materialized path, since the read pair is rebuilt from the read columns, but the sentence at `mod.rs:229` claims more than the check does. "The read count must equal the rows declared per direction on that bus" would be accurate. The check also doesn't run on #2244's whole-plan path (inline at `mod.rs:75`).

bus/src/memory/mod.rs
tcoratger: Confirmed on `origin/feat/bus-terminal-binding`: `BusContext::build` evaluates every AIR under `BusSymbolicBuilder` and forwards `profile.interactions()` wholesale into `BusPlan::build`, with no filter and no opt-out. So the exclusion the module doc asked for was not merely unenforced, it was unsatisfiable. That doc is gone.

Rather than add an opt-out, the helper now discharges both obligations itself, so its declarations are sound under any reduction that proves the bus balance — including the two-tree one.

1. `read_only_memory` takes a `count_inverse` expression and emits `assert_one(count * count_inverse)`. A count with an inverse is nonzero, so the `g * 0 = 0` tuple that pulls and pushes identically can no longer be declared at all. `a_zero_count_row_is_refused_by_the_read_helper` runs the AIR through `check_constraints` on a zero-count row and it panics; `the_read_helper_constrains_every_count_against_its_inverse` pins the honest row.

2. The orbit bound cannot be an AIR constraint, because a declaration cannot see the trace heights. A read is therefore declared through `ReadOnlyMemoryBus::<F>::new(name)`, which refuses any field whose count orbit a machine-word row count could span (`orbit <= usize::MAX`). Rijndael8b is rejected with `CountOrbitReachable { orbit_len: 255 }`, BinaryField128 passes. BabyBear is rejected too, and that is deliberate rather than an oversight: its orbit is about `2^31`, which a trace can reach, so I would rather refuse the field than ship the bound as prose.

The argument this rests on, for the record. Fix an address/value pair that is not a seed pair, and let `C` be the multiset of read counts at it. Balance forces `g * C = C`. Multiplication by `g` is a bijection of `F*` and fixes `0`, so if `0` is not in `C` then `C` is a union of `<g>`-cosets; since `g` has full order, `C` is either empty or has at least `ord(g)` elements. Zero counts are exactly what make `C = {0}` a one-element solution. The inverse constraint removes them and the handle makes `|C| >= ord(g)` unrepresentable.

#2244 needs nothing from this. The inverse check is an ordinary base-field constraint that the folder evaluates like any other, so it holds on exactly the committed count column that feeds the bus factor.

What is left outside p3-bus is the seed and finalization side: a table AIR must place each entry at its own verifier-derived generator power, since an all-equal committed address column makes the array multiset-valued. That obligation is stated on the module and on the leaf claims, and it belongs to the AIR author rather than to the reduction.

bus/src/memory/tests.rs
tcoratger: Right, and reproduced: with `256` in place, changing `read_len >= orbit` to `>` leaves the test green, because `256 > 255` still rejects.

The test now declares an exact, non-power-of-two row count. `bus_plan_rows` emits one single-row block per read, so a bus declaring 255 rows per direction is easy to build. It asserts 255 reads are refused with `CountOrbitTooShort { read_len: 255, orbit_len: 255 }` and 254 are accepted, which pins the inequality from both sides.

Verified against the mutation: with `>` the test fails on the 255 case (`Ok(..)` where `Err(CountOrbitTooShort { read_len: 255, orbit_len: 255 })` is expected). I also checked the address side while I was there — mutating `table_len > orbit` to `>=` fails the same test on the existing `255` acceptance, so both bounds are now pinned strictly.

bus/src/memory/mod.rs
tcoratger: Closed rather than documented.

`product_inputs`, `check_products` and `ReadOnlyMemoryLeaves` are now private, `observe_statement` is private, and the product-shape accessor is gone entirely (the tests read the field directly). There is no longer any way from outside the crate to hand a memory plan's shape to `ProductGkrProof::verify`, so the `roots[2] != 0` check cannot be skipped.

The prover side is closed structurally rather than by a length check. `prove` now takes the columns instead of leaves: it validates them, binds the statement, draws the challenges, and materializes the leaves itself from its own `table_len` and `read_len`. Mismatched leaves are unconstructable, so the `input.len() <= capacity` assert in `ProductGkrProof::prove` is unreachable — I left a `debug_assert_eq!` on the three lengths to keep that stated where it is checked. `materialize` is private for the same reason.

Column validation happens before any challenge is drawn, so a malformed witness returns an error without perturbing the shared transcript.

bus/src/memory/mod.rs
tcoratger: Adopted, in the same shape as `BusPlan::prove`.

`prove` and `verify` both call a private `sample_challenges`, which binds the statement and then draws the `tuple_variables` fingerprint coordinates and the offset from that bound transcript. The fingerprint point and offset no longer appear in any public signature, so the ordering cannot be skipped by a caller.

I passed the columns directly rather than a closure, since the memory materializer only needs borrowed slices, which removes the window in which a caller could see the challenges before materializing.

The drawn challenges travel back to the caller in a `challenges` field on the claims, mirroring `BusReductionOutput`, because the verifier needs them to authenticate the leaf claims against committed columns. The obligation sentence on the security term is gone with it.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,766 @@
...
+        // The enclosing bus plan groups declarations by this name.
+        &self.name
+    }
+}
+
+/// AIR interface for one read-only array access.
+///
+/// A read consumes its current count and produces the next generator-orbit count.
+///
+/// These declarations are not the leaves the reduction proves, only the tuple layout and the read count it covers.
```
Nashtare: This sentence describes the design from before the fix. The module doc now says the declarations are sound under the whole-plan two-tree reduction, and under #2244 that reduction proves exactly these declarations. Could it say that instead, and add that `ReadOnlyMemoryPlan`'s materialized path rebuilds seeds and finalization itself and so expects a bus that carries only the reads? That second part is what the declared-row check in `new` enforces.

bus/src/memory/mod.rs
```diff
@@ -0,0 +1,766 @@
...
+        Ok(ReadOnlyMemoryLeaves {
+            pushes,
+            pulls,
+            counts,
+        })
+    }
+
+    /// Binds this statement, reduces the three trees, and returns the claims still owed.
+    ///
+    /// Challenges are drawn here rather than supplied, so no caller can fingerprint the witness before its dimensions are bound.
```
Nashtare, line 524: Drawing the challenges here also removed the one sentence that said the columns must be committed first. `materialize` used to say both challenges are sampled "after every source column is committed", and now only "commitment binding" in `security_term` hints at it. That ordering matters more now: `prove` takes raw columns and draws the fingerprint from whatever `challenger` holds at call time. A composer that calls it before observing the column commitments lets the prover pick the witness after seeing the point and offset, and balance is then free. Could `prove` and `verify` both say that every column must already be committed, with its commitment observed by `challenger`?

bus/src/memory/tests.rs
```diff
@@ -0,0 +1,949 @@
...
+    }
+}
+
+impl<AB> Air<AB> for ReadAir
+where
+    AB: ReadOnlyMemoryInteractionBuilder<F = F>,
+{
+    fn eval(&self, builder: &mut AB) {
+        // Every row issues one unconditional read through a checked array handle.
+        let bus = ReadOnlyMemoryBus::new("memory").expect("a 128-bit count orbit is unreachable");
```
Nashtare: Nit: #2244's folder runs `eval` once per evaluation point, so an AIR that copies this pays a `BigUint` `F::order()` and a `String` allocation each time. Building the handle once and keeping it on `ReadAir` would make the fixture a better template.

Nashtare (approved):
I re-checked `9575f3d7` against my previous review. The first five commits are a pure rebase of what I reviewed, so this pass covers the fix commit. I also type-checked the branch merged with current main (`p3-bus`, all targets), which is clean.

The main point is closed, and I checked it against #2244's head. `read_only_memory` now emits `assert_one(count * count_inverse)` from a trait default behind a blanket impl, so no builder can override it, and a filtered declaration still doesn't compile. In #2244 the multilinear folder's `record_bus_interaction` is a no-op, but `assert_one` goes through its ordinary `assert_zero` accumulator. The constraint therefore lands in the zerocheck, on the same committed count column the bus factor reads.

With a table AIR declaring seeds and finalization, the two-tree argument goes through:
- Every read count is nonzero, so at a non-seed pair `g·C = C` makes `C` a union of full `<g>`-orbits: empty, or at least `ord(g)` reads.
- The handle refuses `ord(g) <= usize::MAX`, and `BusPlan::build` caps each direction's leaf count at a power of two that fits a `usize`, so the reads stay below `ord(g)` on that path too. BabyBear and the AES field are refused outright.
- At a seed pair it is the usual chain, and no push is zero.

The strict orbit test closes the other threads: 255 reads refused and 254 accepted, so turning `>=` into `>` now fails it. So do dropping the public shape and leaves, with `prove` taking columns, and drawing the fingerprint inside `prove`/`verify` after the statement is bound.

The inline comments are non-blocking. Two docs still describe the design from before the fix, and the commit-before-challenge ordering lost the only sentence that stated it. The precision in the declared-read thread is also still open.

**Smaller things**
- The conflict with #2244 is unchanged: `README.md`, `lib.rs`, and a modify/delete on `leaf.rs`, whose `equality_weights` this module still imports.
- Seeding each entry at `g^i` is still prose on the table side. A table-side helper that pins the address with first-row and transition constraints would make that half mechanical too, as a follow-up.

bus/src/memory/mod.rs
Nashtare: The sentence at `mod.rs:287` still says the check rejects "a bus already carrying seed or finalization declarations of its own", while the check itself (`mod.rs:321-337`) still compares row counts only. It now also clashes with the module doc, which tells the declaring AIR to seed and finalize on this bus (`mod.rs:15`). For that layout `new` refuses the honest `read_len` and accepts a `read_len` that also counts the table rows: with 2 read rows and a 2-row table, 2 fails and 4 passes. Could the sentence say the check compares declared rows only?

bus/src/memory/mod.rs
tcoratger: Stale, and rewritten. The helper doc now reads:

> A whole-plan two-tree reduction proves these declarations directly, alongside whatever seeds and finalization the declaring AIRs add.
>
> The materialized path in this module instead rebuilds seeds and finalization itself, so it refuses a statement whose bus declares anything beyond the reads.

The second sentence is the one you asked for, and it is exactly what the declared-row check refuses: it requires each direction of the named bus to total the statement's read count, so a bus already carrying seed or finalization declarations of its own cannot be used for a materialized statement.

bus/src/memory/mod.rs
tcoratger: Agreed, and the regression is mine. Both entry points now state it:

> Every column passed here must already be committed, with its commitment observed by the transcript.
>
> A caller that commits afterwards may choose the witness once the fingerprint point and offset are known, and balance is then free.

and on verification:

> The transcript must already carry the same column commitments the prover observed, in the same order.

I did look for a type first, since that is the route I took for the orbit bound and for the shape accessor, and I do not think one exists inside this crate.

The obligation is not "these columns exist" — proving already takes them by reference before sampling, so their values are fixed before the challenge. It is "the transcript has already absorbed a commitment to these exact columns". Discharging that needs two things p3-bus does not have. It has no PCS dependency, so it cannot check that a digest opens to the slices it was handed; and `FieldChallenger` exposes no way to ask what a transcript has absorbed. A wrapper type whose constructor takes a commitment would not remove the promise, only relocate it — the constructor would still be trusting the caller that the digest matches the slices, which is the pattern #2271 is about avoiding.

Having the reduction observe the commitment itself is worse rather than better: a composer observes every commitment once, in one fixed order, and a second observation here would fork the transcript from that schedule.

Where this does get typed is one level up, and that is #2244's shape: `BusPlan::prove` is sound about ordering because the multi-stark prover calls it after its own commitment phase, so the control flow supplies the guarantee rather than the callee. The memory reduction should end up driven by the same composer, at which point the sentence becomes a property of the schedule instead of a request. That integration is the part of #2231 this PR deliberately leaves out, and the sentence is the right placeholder until then.

bus/src/memory/tests.rs
tcoratger: Fixed. `ReadAir` now owns a `ReadOnlyMemoryBus<F>` built once in its constructor, and the evaluation borrows it, so a per-point evaluation costs no `F::order()` and no allocation.

PR #2225, opened 2026-09-19
feat(sumcheck): add the basic jagged reduction

tcoratger:
Implements the Basic Jagged reduction from [ePrint 2025/917](https://eprint.iacr.org/2025/917), independently of the binary-WHIR work.

## What lands

- `JaggedLayout` validates arbitrary live column heights and derives the smallest dense power-of-two envelope.
- The prover reduces one virtual zero-padded row-by-column evaluation to one dense multilinear claim with the existing quadratic sumcheck engine.
- The verifier evaluates the terminal selector independently with the paper width-four read-once branching program. The implementation is field-generic and tested in both odd and characteristic-two fields.
- A typed Fiat-Shamir driver binds the complete layout, sparse point, claimed value, every round message, and the delegated dense claim.
- Row equality weights are generated only through the largest live height, so a large virtual row bound does not trigger a rectangular allocation.

## Security statement

For `m` dense variables, the reduction contributes at most `2m / |EF|` soundness error, in addition to the binding error of the PCS that authenticates the returned dense claim. Attacker-controlled proof lengths are checked before replay, and malformed public geometry is rejected before transcript use.

## Tests

- direct selector MLE versus the branching program, including randomized layouts and binary fields
- honest end-to-end reduction and dense-claim equality
- statement mutation, missing rounds, forged terminal values, empty traces, overflow and shape errors
- two layouts of the same shape yield different transcript seeds, checked on a sampled challenge
- a forty-variable sparse row bound to guard against rectangular allocation

Focused checks run:

- `cargo test -p p3-sumcheck jagged --lib`
- `cargo test -p p3-sumcheck jagged --lib --features parallel`
- `cargo clippy -p p3-sumcheck --all-targets --all-features -- -D warnings`

No throughput claim is made, so this PR adds no benchmark result.

## Follow-up

This PR deliberately leaves the commitment adapter, the Jagged Assist/Fancy Jagged verifier acceleration, and ring-switch/WHIR composition to follow-ups. Those layers can consume the single dense claim exposed here without changing this reduction.

---

Part of #2239.


tcoratger:
I checked the reduction directly against Flock at `b684b1258e4b1f202bec24afd660ace851b09e5e` (`crates/flock-core/src/pcs/jagged.rs`).

The Basic Jagged construction agrees on the points that matter:

- column-major prefix-sum layout and the paper's Eq. (3)/(4);
- the width-four `(carry, less_than)` branching program and its accepting state;
- one degree-two product sumcheck per dense variable;
- the terminal relation `claim = dense_evaluation * selector_evaluation`;
- explicit translation between Plonky3's most-significant-variable-first points and the low-bit-first automaton.

This implementation also handles `row_variables > dense_variables`, which Flock's workload does not need, and binds the complete public layout, sparse point, claim, round messages, and terminal dense evaluation through the typed transcript.

Focused validation:

- `cargo test -p p3-sumcheck jagged --lib --features parallel` — 12 passed;
- `cargo clippy -p p3-sumcheck --all-targets --all-features -- -D warnings`;
- `RUSTDOCFLAGS='-D warnings' cargo doc -p p3-sumcheck --no-deps`;
- Flock's `f_hat_t` tests — 3 passed;
- Flock's `sumcheck_roundtrip` — passed.

Flock's Jagged Assist/Fancy Jagged batching is deliberately not claimed here; it is a follow-up optimization above this Basic Jagged primitive.

sumcheck/src/jagged/selector.rs
```diff
@@ -0,0 +1,341 @@
...
+        let selector = JaggedSelector::new(&layout).table(&sparse);
+        let expected = row
+            .iter()
+            .fold(F::ONE, |weight, &value| weight * (F::ONE - value));
+
+        assert_eq!(selector, vec![expected]);
+    }
+
+    #[test]
+    fn the_selector_identity_holds_in_characteristic_two() {
```
Nashtare, line 407: This test is vacuous twice over, and it is the only evidence for the PR's claim that the implementation is "tested in both odd and characteristic-two fields".

`BinaryField128` does not override `from_u64`, so it goes through the blanket `from_prime_subfield(Gf2::from_int(n))` and `Gf2` reduces mod 2 — `gf2.rs`'s own test asserts `Gf2::from_u64(7) == ONE`. So `from_u64(5) = from_u64(9) = from_u64(7) = from_u64(11) = from_u64(13) = ONE`: every coordinate here is `ONE`, and the "binary-field point" is the all-ones **Boolean** point. Nothing about non-Boolean behaviour in characteristic two is exercised, which is exactly the property the branching program needs.

Worse, both sides are zero. With `layout = (2, [1, 3])` the row point `(1,1)` is row index 3, no column has height 4, so `equality_prefix` returns `[0,0,0]` and the materialized selector is `[0,0,0,0]`; the automaton also returns 0 at Boolean `d = 3` (column 0 fails `d < end`, column 1 needs `r + start = 4`). The assertion is literally `0 == 0`, and it would pass against an `evaluate` that returned a constant zero.

Building the points with `BinaryField128::from_repr(<u128>)` (as `tower.rs`'s own tests do) and picking a layout and point where the selector is nonzero would fix both halves. The proptest just below uses `field_point`, i.e. `BabyBear`, so it does not cover this either.

sumcheck/src/jagged/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        let mut padded = dense.into_iter().map(EF::from).collect::<Vec<_>>();
+        padded.resize(layout.dense_capacity(), EF::ZERO);
+        let expected = Poly::new(padded).eval_base(verifier_claim.point());
+
+        assert_eq!(prover_claim, verifier_claim);
+        assert_eq!(*verifier_claim.value(), expected);
+    }
+
+    #[test]
+    fn changing_any_public_statement_component_rejects() {
```
Nashtare, line 374: All three mutations here are rejected by mechanisms independent of the transcript: the verifier recomputes `selector.evaluate` from *its own* layout and point, and folds *its own* `claimed_value`, so a perturbation propagates to the terminal check arithmetically. Deleting `encode_statement` and the `separator.instance(...)` call entirely would leave all three assertions passing.

The mutated-layout case also keeps the same shape (`[2,1,5,1]` gives prefixes `[0,2,3,8,9]` and the same `dense_variables = 4`), so it does not exercise the round-count path either.

A test that actually pins the binding would assert that two statements differing only in the layout produce different challenger states — comparing a sampled challenge, say — rather than only that verification fails. Binding `cumulative_heights` is what makes a nested or adaptive use of this reduction safe, and it currently has no regression.

sumcheck/src/jagged/transcript.rs
```diff
@@ -0,0 +1,265 @@
...
+
+/// Sponge alphabet of a challenger that speaks the base field natively.
+type Alphabet<F> = FieldUnit<F>;
+
+/// Type-level name of the delegated quadratic sumcheck.
+struct ProductSumcheck;
+
+/// Public dimensions fixing the transcript shape of one jagged reduction.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) struct JaggedShape {
```
Nashtare: None of these three fields is ever read. `pattern(&self)` builds a fixed three-step sequence and never touches `self`, and `domain_separator` takes the `layout` as a separate argument, so the interaction pattern — and the pattern fingerprint folded into the seed — is byte-identical for every jagged instance.

Not a soundness gap: the geometry, `dense_variables` included, is bound through `encode_statement`, and the inner sumcheck seeds its own pattern carrying the round count. But the doc comment ("Public dimensions fixing the transcript shape of one jagged reduction") is then not accurate. Either drop the struct and pass the layout, or make the pattern actually depend on it.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    /// Proves an evaluation of a virtual jagged table.
+    ///
+    /// The dense witness contains only live cells in column-major order.
+    ///
+    /// The returned dense claim must be opened against the commitment to the same witness.
+    ///
+    /// # Soundness
+    ///
+    /// The reduction contributes at most `2m / |EF|` error for `m` dense variables.
```
Nashtare: The bound reads as absolute, but `impl ExtensionField<F> for F` holds, so `prove::<BabyBear, BabyBear, _>` / `verify::<BabyBear, BabyBear, _>` compile and `2m / |EF|` silently becomes `2m / 2^31`. There is no `debug_assert` on `EF::bits()` or `p3-security` hook at either entry point.

This is a crate-wide idiom rather than something introduced here, but since this doc states the bound, could it name the assumption — that `|EF|` must be large enough for the target — or add a debug assertion?

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,397 @@
...
+        if dense_witness.len() != self.area() {
+            return Err(JaggedError::DenseLengthMismatch {
+                expected: self.area(),
+                actual: dense_witness.len(),
+            });
+        }
+
+        // The committed data occupies the live prefix.
+        // The power-of-two suffix is represented only by zeros inside this reduction.
+        let mut dense = EF::zero_vec(self.dense_capacity());
```
Nashtare: Nit, cost only: this materializes the whole dense witness in `EF` and then uses `ProductPolynomial::new_unpacked`, which pins the run to the scalar variant for all `m` rounds. The witness is base-field and prefix-ordered, so `new_packed` looks applicable and would keep the early rounds on the SIMD path. Correctness is unaffected.

Nashtare (commented):
I re-derived the width-four branching program from the code rather than from the paper, and checked it against the dense definition, the bit-order conventions and the allocation claim. The construction is correct. My one real ask is about the characteristic-two test.

**What holds**

- **The branching program is an exact MLE, not just a Boolean-cube match.** `boundary_evaluation` unrolls to `sum over Boolean (r, d) of prod_k eq(y_k, r_k) eq(z_k, d_k) [accept]`. `cube_weight` uses `v` / `ONE - v` per bit with no branch on a field value, and `transition` branches only on the four public/Boolean *symbol* bits, never on a field element. Two multilinears agreeing on the cube are equal, so the identity holds at arbitrary extension points — which is what the verifier needs.
- **The automaton is right.** LSB-first ripple carry, `sum = row + start + carry`, reject when `dense != sum & 1`; `next_less_than = if dense == end { less_than } else { end }` latches the highest differing bit; `INITIAL = (carry 0, less 0)` and `ACCEPT = (carry 0, less 1)` give strict `<`. `[d = r + start and d < end]` is equivalent to `r < h_c`, and `d >= start` comes for free.
- **The layer count is exactly right.** `top = max(row_vars, dense_vars)` over `0..=top` gives each of the row and dense points exactly one extra zero layer. `end <= area <= 2^dense_vars <= 2^top`, so a trailing `end = 2^dense_vars` bit is still covered; and `r + start < 2^(top+1)`, so requiring carry 0 after the top layer is correct. I hand-checked `end = 2^dense_vars` both with `top = dense_vars` and with `top = row_vars > dense_vars`.
- **Bit order is coherent end to end.** `equality_prefix` makes the last point coordinate the index LSB, which matches `point_coordinate_from_low`'s `num_variables - layer - 1`, `Poly::new_from_point`, and the `Prefix` variable order the sumcheck returns. `equality_prefix_matches_the_full_table` pins it.
- **The truncating `equality_prefix` is correct**, including for non-power-of-two lengths: descending iteration over `0..ceil(next/2)` means a target `2i` is only ever an unread or already-processed slot, and dropped sources have both children out of range. I hand-verified `length = 3` against the full table.
- **Characteristic two is genuinely safe in the code.** No `halve`, no division, no `NEG_ONE`, no `x^2 = x` shortcut, no reliance on `1 - x != x` anywhere in `selector.rs` — only `ONE - v` and `from_bool`.
- **The 40-variable claim holds.** `table()` sizes `equality_prefix` by the largest *live* column height, and the only other prover allocation is `dense_capacity <= 2*area`. The verifier allocates nothing sized by the row bound: `evaluate` allocates `2^column_variables = num_columns` and runs at most 64 layers per column.
- **Fiat-Shamir is complete.** `encode_statement` binds `row_variables`, `dense_variables`, the height count and every cumulative height, then every row/column coordinate and the claimed value through fixed-width `TranscriptField::encode`, so the concatenation is injective given the already-bound widths. `DomainSeparator::seed` absorbs rather than resets, so the inner sumcheck stacks on the outer state, and `dense_evaluation` is bound before the terminal check.
- **Attacker-controlled lengths are checked before they size anything.** `SumcheckData::verify_rounds` re-checks the round count and witness count before seeding the inner transcript and before the `Vec::with_capacity`; the loop is driven by the verifier-derived `dense_variables`. `verify` also reads nothing statement-shaped out of `JaggedProof`, and `sparse_value` lives on the non-serializable prover output, so it cannot be smuggled in.
- **Overflow and panics.** `u32::try_from` plus `checked_shl` caps `row_variables` at 63, the running area and the capacity use checked arithmetic, `integer_bit` guards the shift with `layer < usize::BITS`, and `point_coordinate_from_low` uses `checked_sub`. The abort path runs `end_protocol` before inspecting the inner error, so an inner rejection surfaces as the real error rather than a drop panic.

**Soundness bound**

`2m / |EF|` is right and nothing is omitted. The summand is a product of two multilinears, so each round message has degree 2; `m` rounds give `2m/|EF|` by Schwartz-Zippel, and the challenges really are drawn in `EF`. There is no batching across columns (the column weights are a deterministic function of the caller's sparse point), the sparse point is an input rather than a challenge, and the selector carries no randomness of its own — the verifier recomputes it deterministically. Worth noting in the doc that `m` is the *dense* arity `log2(next_pow2(area))`, which is unrelated to the (possibly 40-variable) virtual row bound.

**What I would fix**

The characteristic-two test does not test characteristic two (inline, `selector.rs:296`). Everything else is minor.

Not blocking.

sumcheck/src/jagged/selector.rs
tcoratger: Confirmed both halves before touching anything. Printing the raw tower representation of `BinaryField128::from_u64(5|9|7|11|13)` gives `[1, 1, 1, 1, 1]`, the materialized table is `[0, 0, 0, 0]`, and both sides of the assertion evaluated to zero, exactly as described.

The test now builds every point through `BinaryField128::from_repr` on a full 128-bit pattern, with a helper that asserts no coordinate collapsed to zero or one. The layout is the same `[3, 0, 5, 1]` fixture as the odd-characteristic test, so the selector is genuinely nonzero, and the test asserts `evaluation != ZERO` before comparing against the materialized table. Mutating `boundary_evaluation` to return a constant zero now fails it.

The PR description claim is therefore true as of this commit rather than before it.

sumcheck/src/jagged/mod.rs
tcoratger: Verified the claim first: replacing the body of `domain_separator` so that `encode_statement` and `separator.instance(...)` are never called leaves all eleven jagged tests green, including this one.

Added `the_column_geometry_separates_two_transcript_seeds`, which seeds a prover transcript from `[3, 0, 5, 1]` and from `[2, 1, 5, 1]`, samples the first challenge through the lent sponge, and asserts the two differ while the same layout twice gives the same challenge. It also asserts the two layouts agree on `dense_variables`, so nothing but the seed can separate them. Under the same deletion this test is the only one in the module that fails.

sumcheck/src/jagged/transcript.rs
tcoratger: Took the first option: the struct is gone and `pattern` and `domain_separator` are free functions taking the layout. It is the smaller diff, and it avoids the alternative's real problem, which is that anything the pattern could carry about the geometry is already bound byte for byte by the statement encoding, so a geometry-dependent pattern would be duplicate binding rather than new binding. Net change is thirty fewer lines and no inaccurate doc comment left to defend.

sumcheck/src/jagged/layout.rs
tcoratger: Named the assumption. Both `# Soundness` sections now carry one line saying the bound is meaningful only when the extension is large enough for the target security level.

Went with the doc rather than a debug assertion because the assertion would be the only one of its kind here: grepping the crate turns up no `debug_assert` on an extension degree or bit count anywhere, and the one security-budget hook that exists goes through `p3-security` in `prescribed_pcs.rs`, not through an entry-point assertion. Adding one only in this module would suggest the rest of the crate enforces something it does not.

sumcheck/src/jagged/layout.rs
tcoratger: Deferring this one as a cost-only item, correctness unaffected.

It is not a drop-in swap: `Poly::pack` asserts `num_variables() >= log2(F::Packing::WIDTH)`, and a jagged layout may legitimately have fewer dense variables than that. The empty-trace test in this module has zero of them. So the change needs a size guard and two construction paths, and it would also want the witness embedded through the packed extension rather than the scalar `EF` loop it replaces, which is a different piece of work from a constructor swap. Rather than land that on a guess, I would prefer to do it with a measurement behind it in a follow-up.

sumcheck/src/jagged/selector.rs
```diff
@@ -0,0 +1,368 @@
...
+        // Multilinear extension of one automaton layer.
+        for (state, next_weight) in prefix.iter_mut().enumerate() {
+            for symbol in 0usize..16 {
+                if let Some(next_state) = transition(symbol, state) {
+                    *next_weight += cube_weight(&coordinates, symbol) * suffix[next_state];
+                }
+            }
+        }
+
+        suffix = prefix;
```
tcoratger, line 194: **[MAJOR] The verifier's selector evaluation does 16x more field multiplications than it needs**

Measured on this branch, `evaluate` against an optimized rewrite asserted to return the identical value:

```
row_vars=6   cols=64    dense_vars=11   current=670 us    optimized=46 us     14.7x
row_vars=24  cols=256   dense_vars=13   current=5.52 ms   optimized=368 us    15.0x
row_vars=40  cols=1024  dense_vars=15   current=35.99 ms  optimized=2.39 ms   15.1x
```

36 ms of verifier time on the exact shape this module advertises (`sparse_row_weights_do_not_allocate_the_row_bound` picks `row_variables = 40` precisely to show the reduction survives a 40-variable row bound). Two independent redundancies, both removable without touching the automaton:

**1. Twelve of the sixteen symbols are identically zero.** `coordinates[2]` and `coordinates[3]` come from `integer_bit`, which returns exactly `F::ZERO` or `F::ONE`. So in `cube_weight`, the `start` factor is `0` unless `symbol_bit_2 == start_bit`, and likewise for `end`. Only the four symbols whose high two bits match the public `(start_bit, end_bit)` can be nonzero -- the other twelve are computed in full, three multiplications each, and then multiplied by `suffix[..]` and added to zero.

**2. The `(row, dense)` factors do not depend on the column.** `boundary_evaluation` is called once per column, and `point_coordinate_from_low(row_point, layer)` / `(dense_point, layer)` return the same two values every time. Their four products are recomputed `num_columns` times per layer.

Together with the existing `transition` filter, the layer body collapses to eight multiplications per column, plus four shared per layer:

```rust
// once per layer, shared by every column
let (r, d) = (coord_low(row_point, layer), coord_low(dense_point, layer));
let w = [ (F::ONE - r) * (F::ONE - d), (F::ONE - r) * d,
          r * (F::ONE - d),            r * d ];

// per column: start_bit / end_bit are usize, the symbol is fully determined
for (state, slot) in prefix.iter_mut().enumerate() {
    let (carry, less) = (state & 1, state >> 1);
    for row_bit in 0..2 {
        let sum = row_bit + start_bit + carry;
        let dense_bit = sum & 1;                       // forced, not searched
        let next_less = if dense_bit == end_bit { less } else { end_bit };
        *slot += w[row_bit * 2 + dense_bit] * suffix[(sum >> 1) | (next_less << 1)];
    }
}
```

That is `128 -> 8` multiplications per (column, layer), which is the 15x above. `transition` stays as the readable spec; the loop just stops enumerating symbols it has already proved are zero.

sumcheck/src/jagged/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            .unwrap();
+
+        // The surviving claim is exactly an evaluation of the zero-padded dense witness.
+        let mut padded = dense.into_iter().map(EF::from).collect::<Vec<_>>();
+        padded.resize(layout.dense_capacity(), EF::ZERO);
+        let expected = Poly::new(padded).eval_base(verifier_claim.point());
+
+        assert_eq!(prover_claim, verifier_claim);
+        assert_eq!(*verifier_claim.value(), expected);
+    }
```
tcoratger: **[MAJOR] Nothing in the suite pins the sparse side of the reduction**

This test checks the *dense* half: the returned claim is the padded witness's MLE at the returned point. That is the half the sumcheck already forces. The half that defines what this module is for -- that `sparse_value` is the evaluation of the **virtual jagged table** at `(y, z)` -- is never asserted against anything independent.

Trace the closed loop. `sparse_value` comes from `table()`. `evaluate()` is checked only against `table()` (`branching_program_matches_the_materialized_selector`, the char-2 test, the proptest). `verify` checks `terminal == dense_evaluation * evaluate(...)`. Every link is internal. Swap the addressing convention -- row-major instead of column-major, or `r = end_c - 1 - (d - start_c)` instead of `r = d - start_c` -- and `table()` and `evaluate()` move together, the sumcheck still closes, the seed still separates, and all eleven tests stay green while the module proves a statement about a different table than the caller believes.

I wrote the missing reference and ran it. The implementation is **correct** -- 200 proptest cases agreeing exactly. The gap is purely that no regression would catch it:

```rust
fn naive_jagged_eval(heights: &[usize], w: &[F], y: &Point<EF>, z: &Point<EF>) -> EF {
    let (ry, rz) = (Poly::new_from_point(y.as_slice(), EF::ONE),
                    Poly::new_from_point(z.as_slice(), EF::ONE));
    let mut acc = EF::ZERO;
    let mut start = 0;
    for (c, &h) in heights.iter().enumerate() {
        for r in 0..h { acc += rz.as_slice()[c] * ry.as_slice()[r] * EF::from(w[start + r]); }
        start += h;
    }
    acc
}

proptest! {
    #[test]
    fn sparse_value_is_the_jagged_mle(
        heights in prop::collection::vec(0usize..=8, 4),
        y in /* 3 coords */, z in /* 2 coords */, w in /* 32 base elements */,
    ) {
        let layout = JaggedLayout::new(3, &heights).unwrap();
        let witness = &w[..layout.area()];
        let out = layout.prove(witness, &point, &mut challenger()).unwrap();
        prop_assert_eq!(*out.sparse_value(), naive_jagged_eval(&heights, witness, &y, &z));
        // then verify and check the dense claim, as this test already does
    }
}
```

`heights in 0..=8` with `row_variables = 3` sweeps the degenerate shapes for free: all-zero (`dense_variables = 0`, no rounds), a single live cell, one full column at exactly the row bound, interior empty columns, and non-power-of-two areas. A failure names the exact `(heights, y, z)` and immediately says which of the two conventions drifted.

A second one is worth 10 lines: `row_variables = 20` with `heights in 0..=3`, so `dense_variables <= 3` and `top = row_variables`. That is the `row_variables > dense_variables` support this PR claims over Flock, and the existing fixtures never reach it -- `branching_program_matches_the_materialized_selector` and the char-2 test both have `top = dense_variables = 4`, and `sparse_row_weights_do_not_allocate_the_row_bound` calls `table()` only, never `evaluate()`. The proptest does stumble into it for small areas, but incidentally, and it never runs `prove`/`verify`. I ran that shape over 60 cases; it round-trips.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,401 @@
...
+        // The branching program computes the correct multilinear extension at this field point.
+        let selector_evaluation = selector.evaluate(point, &dense_point);
+        if terminal != proof.dense_evaluation * selector_evaluation {
+            return Err(JaggedError::TerminalMismatch);
+        }
+
+        Ok(JaggedDenseClaim {
+            point: dense_point,
+            value: proof.dense_evaluation,
+        })
```
tcoratger, line 341: **[MAJOR] Say what happens when the caller drops the returned claim**

The terminal check is one equation in two prover-chosen unknowns, and only one of them is pinned elsewhere. A prover who runs the transcript honestly right up to here, then sets

```
dense_evaluation := terminal * selector_evaluation^{-1}
```

satisfies line 338 with probability 1 for **any** starting `claimed_value`, for any witness, for no witness at all. One inversion. The entire soundness of the reduction is carried out of this function in the return value, and rests on the caller opening the commitment at `dense_point` and getting `dense_evaluation` back.

The docs do say the claim "must be opened against the commitment to the same witness". What they do not say is that *not* doing so is not a weakening -- it is total, silent acceptance. And the type does not help: `Result` is `#[must_use]`, but

```rust
layout.verify(&point, value, &proof, challenger)?;   // compiles, warns nothing, proves nothing
```

discards the claim through `?`. For brand-new public API whose only output is security-critical, I would:

- put a `# Security` section on `verify` stating the failure mode in one line, not just the obligation;
- `#[must_use]` on `JaggedDenseClaim` itself, so the `?`-and-drop form warns;
- state the arity obligation explicitly, since it is currently nowhere: the commitment must be to a polynomial in exactly `dense_variables()` variables whose first `area()` evaluations are the live cells in this layout's column-major order. Note the padding is genuinely unconstrained -- `table()` is identically zero on `area..dense_capacity` (I checked), so a commitment with nonzero padding is still sound. That is a real and non-obvious property and is worth writing down rather than leaving a future integrator to rediscover.

Also worth pinning `selector_evaluation != 0` somewhere in the tests, so the `TerminalMismatch` case in `malformed_proof_shapes_and_terminal_values_reject` cannot go vacuous the way the char-2 test did.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,401 @@
...
+            *destination = EF::from(source);
+        }
+
+        // On Boolean dense indices this selector maps the contiguous representation back to rows.
+        let selector = selector.table(point);
+        let claimed_value = dense
+            .iter()
+            .zip(&selector)
+            .map(|(&value, &weight)| value * weight)
+            .sum();
```
tcoratger: **[MINOR] The claim can be computed over `area` base-field scalings instead of `dense_capacity` extension products**

Separate from @Nashtare's `new_packed` point, and a much smaller diff. Two costs here are avoidable:

1. `dense` is materialized at `dense_capacity`, which is up to `2 * area`, and the zip at 213 walks all of it. Everything past `area` is `EF::ZERO * selector[i]`, and `selector[i]` is zero there too.
2. `value * weight` is a full `EF x EF` product, but `value` is a base-field element that was just widened at line 208. `EF: ExtensionField<F>: Algebra<F>` gives `EF * F` directly.

Zipping the base-field witness instead keeps both:

```rust
let claimed_value = dense_witness
    .iter()
    .zip(&selector)
    .map(|(&value, &weight)| weight * value)   // EF * F
    .sum();
```

Measured at `area = 2^20` over `BabyBear`/`EF4`, current loop including the widen vs. this one: **9.37 ms -> 3.29 ms, 2.8x**. The `EF::zero_vec` + widen is still needed for the `ProductPolynomial`, so the win is on the claim only -- but it is free, and it is on the prover's critical path once per opening.

tcoratger (commented):
Reviewed as a fresh pass over the construction, not a re-read of @Nashtare's. I re-derived the sparse-to-dense map, then **checked it by experiment** rather than by inspection: I added a naive reference evaluator for the virtual jagged table and ran the real `prove`/`verify` against it over random layouts, including the regimes the fixtures do not reach. Everything I ran passed, and I reverted every experiment.

### The mapping, for the record

```
  sparse view                          dense witness (column-major)
  (row r, column c)                    index d
                                       
  c=0  c=1  c=2  c=3                   prefix = [0, 3, 3, 8, 9]
  +---+     +---+                      
r0| a |     | d |                      d:  0 1 2 | (none) | 3 4 5 6 7 | 8    | 9..15
r1| b |     | e |  +---+                   a b c |        | d e f g h | i    | pad
r2| c |     | f |  | i |                   \_____/         \_________/  \___/
  +---+     | g |  +---+                    c=0             c=2          c=3
   h_0=3    | h |   h_3=1               
   h_1=0    +---+                       d in [prefix[c], prefix[c+1])  <=>  column c
             h_2=5                      r = d - prefix[c]
```

`table()` writes `eq(y, r) * eq(z, c)` into slot `prefix[c] + r` and leaves everything from `area` on at zero; `evaluate()` recomputes the same multilinear at a field point via the width-four `(carry, less_than)` automaton for `[d = r + start_c] AND [d < end_c]`. Row-to-row leakage is impossible by construction: the intervals `[prefix[c], prefix[c+1])` partition `0..area`, the `zip` in `table()` is bounded by the shorter of the slice and the weight vector, and an empty column yields `d = r + s >= s = e`, which the `less_than` latch rejects on both the `r = 0` (all bits equal, latch stays 0) and `r > 0` (top differing bit has `end = 0`) branches. I hand-checked both.

### What I verified by running it

- **Sparse semantics.** A naive reference `sum_c eq(z,c) * sum_{r < h_c} eq(y,r) * W[prefix_c + r]` agrees with `prove`'s `sparse_value` on 200 random `(heights, y, z, W)` draws over `BabyBear`/`EF4`, and the returned dense claim equals `Poly(padded).eval(dense_point)` on every one. The reduction is right.
- **`row_variables >> dense_variables`.** 60 draws at `row_variables = 20` with `dense_variables <= 3`, so `top = row_variables` and the row point carries 17 layers past the dense point. All round-trip.
- **Padding is free.** I confirmed `table()` is identically zero on `area..dense_capacity`, so a commitment with nonzero padding cannot move the claim. The reduction correctly does not need to bind the padding.
- **The optimized selector agrees.** My rewrite in the perf finding below was asserted equal to `evaluate` on every benchmark shape before timing.

### Where the transcript stands

The binding is sound. `encode_statement` covers `row_variables`, `dense_variables`, the prefix-table length and every entry, then the row point, the column point and the claim, all at fixed width, and it runs **before** any sampling because `DomainSeparator::seed` absorbs into the sponge at `ProverState::new`/`VerifierState::new`. `validate_point` runs before the transcript exists, so the widths that make the concatenation parseable are already fixed. The round count is bound separately by the inner `SumcheckShape`'s own pattern hash. Nothing statement-shaped is read out of `JaggedProof`, and `sparse_value` lives only on the non-`Serialize` prover output.

### Findings

3 MAJOR, 3 MINOR, 3 NIT, inline. No blocker: I could not construct a forgery against the reduction as written, and the two mechanisms a jagged reduction usually gets wrong -- boundary leakage and unbound lengths -- are both closed here.

### Unrelated

The PR description on GitHub is the literal string `@/tmp/claude-1000/.../body.md`. The file reference never expanded, so every reviewer arriving at this page sees a dangling local path instead of the body @Nashtare's review quotes. Worth pasting the real text in.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,401 @@
...
+        };
+        let proof = JaggedProof {
+            sumcheck,
+            dense_evaluation,
+        };
+        Ok(JaggedProverOutput {
+            proof,
+            sparse_value: claimed_value,
+            dense_claim,
+        })
```
tcoratger: **[MINOR] `prove` invents the statement rather than being handed it**

`claimed_value` is computed at line 213 from the witness and returned as `sparse_value`. The caller never gets to say what it is proving.

In the composition this reduction exists for, the caller *already has* a value: the jagged opening is being invoked to discharge a claim `v` that came from somewhere upstream. If the witness does not actually evaluate to `v`, `prove` does not complain -- it seeds the transcript with its own value, produces a perfectly valid proof of a different statement, and hands back `sparse_value != v`. The mismatch only surfaces at verification time, as a sumcheck round failure with no indication that the *claim* was the thing that disagreed. Debugging that from a `RoundCountMismatch`-shaped error is unpleasant.

Taking `claimed_value: EF` as a parameter and returning a `ClaimMismatch { expected, actual }` variant puts the failure at the place that can explain it, and makes the signature say that this is a reduction of a given claim rather than a computation of one. Not soundness -- completeness and diagnosability.

sumcheck/src/jagged/error.rs
```diff
@@ -0,0 +1,82 @@
...
+        actual: usize,
+    },
+    /// The dense witness does not contain exactly the live cells.
+    #[error("the dense witness has {actual} cells, expected {expected}")]
+    DenseLengthMismatch {
+        /// Number of live cells fixed by the layout.
+        expected: usize,
+        /// Number of cells supplied by the prover.
+        actual: usize,
+    },
```
tcoratger: **[MINOR] Five error variants have no test, including the three on the pre-transcript validation path**

`DenseLengthMismatch`, `RowPointWidthMismatch` and `ColumnPointWidthMismatch` are never constructed by any test. These are exactly the checks the security argument leans on -- `layout.rs:194` says "Reject malformed public geometry before touching the transcript", and `layout.rs:294` says "Reject caller-owned shape mismatches before advancing the sponge". If either `validate_point` call were deleted, or moved below the `JaggedProverTranscript::new` at 220 / 308, nothing in the suite would notice, and the ordering guarantee that makes `encode_statement`'s concatenation parseable would be gone.

Three asserts:

```rust
let layout = JaggedLayout::new(3, &[3, 0, 5, 1]).unwrap();     // rows 3, cols 2
assert_eq!(layout.prove(&[F::ONE; 8], &point, &mut challenger()),
           Err(JaggedError::DenseLengthMismatch { expected: 9, actual: 8 }));
// row point with 4 coords -> RowPointWidthMismatch { expected: 3, actual: 4 }
// column point with 1 coord -> ColumnPointWidthMismatch { expected: 2, actual: 1 }
```

`JaggedLayoutError::RowVariablesOverflow`, `AreaOverflow` and `DenseAreaOverflow` are likewise untested. The first two are one line each (`JaggedLayout::new(64, &[1])`, and two columns of `usize::MAX / 2 + 1`), and they are the guards that keep `1usize << dense_variables` at line 144 and `1usize.checked_shl` at line 64 from ever being reachable with a bad exponent -- worth locking in, since `dense_capacity()` is a `pub` `const fn` that would silently overflow if the constructor's bound ever loosened.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,401 @@
...
+        selector.validate_point(point)?;
+
+        // The proof must not smuggle unused grinding witnesses into a zero-difficulty reduction.
+        if !proof.sumcheck.pow_witnesses.is_empty() {
+            return Err(SumcheckError::PowWitnessCountMismatch {
+                expected: 0,
+                actual: proof.sumcheck.pow_witnesses.len(),
+            }
+            .into());
+        }
```
tcoratger: **[NIT] This check is already inside `verify_rounds`, with the same error**

`SumcheckData::verify_rounds` computes `expected_witnesses = if pow_bits > 0 { expected_rounds } else { 0 }` and raises the identical `PowWitnessCountMismatch { expected: 0, actual }` when `pow_bits == 0`, which is what line 320 passes. So this block can never fire.

Defence in depth is fine and the ordering argument (reject before seeding) is real, but the comment reads as though this is the only place the canonical-shape rule is enforced, which will mislead the next reader into thinking `verify_rounds` does not check it. Either drop the block and say in a comment that `verify_rounds` binds both counts before any transcript work, or keep it and say it is a deliberate duplicate.

sumcheck/src/jagged/selector.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    weights
+}
+
+impl JaggedSelector<'_> {
+    /// Evaluates the selector polynomial at an arbitrary dense point.
+    ///
+    /// The computation is the width-four read-once branching program from ePrint 2025/917.
+    ///
+    /// It checks both `dense_index = row + column_start` and `dense_index < column_end`.
```
tcoratger, line 109: **[NIT] The automaton deserves its derivation in-tree**

"the width-four read-once branching program from ePrint 2025/917" is the only justification in the repo for the single most subtle piece of code in the module. Anyone reviewing `transition` has to fetch the paper and match conventions by hand -- and the conventions are not the paper's, since `point_coordinate_from_low` exists precisely to translate Plonky3's most-significant-first points into the automaton's LSB-first order.

The whole argument fits in a doc comment:

```text
 State = (carry, less_than), read LSB-first over layers 0 ..= top.

   layer:      0      1      2     ...    top
   row  y:    y_0    y_1    y_2          [0 above row_variables]
   dense z:   z_0    z_1    z_2          [0 above dense_variables]
   start  :   s_0    s_1    s_2           public bits
   end    :   e_0    e_1    e_2           public bits

   carry:      z_k must equal (y_k + s_k + carry) & 1,  carry' = sum >> 1
   less:       less' = if z_k == e_k { less } else { e_k }     (latches at the
                                                                top differing bit)

   INITIAL = (carry 0, less 0)        ACCEPT = (carry 0, less 1)
```

Three things a reader then does not have to re-derive, and which I did check by hand:

- accepting on `less_than = 1` gives **strict** `<`, and `d >= start` is free because `d = r + start` with `r >= 0`;
- `top = max(row_variables, dense_variables)` and the inclusive `0..=top` give each point exactly one extra zero layer, which is what covers a trailing `end = 2^dense_variables` bit and forces the final carry out of bit `top` to be zero (`r + start < 2^(top+1)` always, so no legitimate pair is rejected);
- `[d = r + s] AND [d < e]` is equivalent to `r < h_c`, which is why the whole thing is the selector at all.

Worth also recording, next to the `2m / |EF|` bound at `layout.rs:173`, that `m` is `dense_variables` -- the dense arity `log2(next_pow2(area))` -- and is unrelated to the row bound. At `row_variables = 40` with a handful of live cells those two numbers differ by 35, and a reader who guesses wrong reads the soundness bound as far worse than it is.

sumcheck/src/jagged/transcript.rs
```diff
@@ -0,0 +1,231 @@
...
+        self.state.abort();
+    }
+
+    /// Closes the transcript after every described step has been replayed.
+    pub(super) fn finish(self) {
+        // The replay reads values from the typed proof rather than a byte wire.
+        self.state
+            .finalize()
+            .expect("the jagged reduction reads an empty wire");
+    }
```
tcoratger, line 192: **[NIT] Unreachable `expect` on the verifier path -- say why, or make it total**

`VerifierState::finalize` returns `Err` only on `self.cursor != self.narg.len()`, and `JaggedVerifierTranscript::new` constructs the state with `narg = &[]`. The cursor never moves because no step reads from the wire. So this is structurally `Ok(())` and the `expect` cannot fire.

That is fine, but it is an `expect` in a verifier, which is the one place a reader is trained to stop and check. The reason it is safe lives two files away, in `VerifierState::new(challenger, &separator, &[])` at line 156. One line here -- "the driver is built with an empty wire, so `finalize` can only fail the pattern check, which panics on its own" -- saves that trip. Same for the `assert!` in the prover's `finish` at line 125.

sumcheck/src/jagged/selector.rs
tcoratger: Both redundancies are gone. The four `(row, dense)` products are now computed once per layer and shared by every column, and the inner loop enumerates only the four symbols whose two public bits match the endpoint bits at that layer. `transition` is untouched and still decides the next state, so the automaton spec is unchanged -- the loop just stops offering it symbols whose factor is provably zero. That is 128 -> 8 multiplications per (column, layer) as you described.

I reproduced the ratio but not the absolute numbers; my baseline is about 7x faster than yours, so this is probably hardware. 20 iterations each, release, same three shapes, identical accumulator in every case:

```
row_vars=6   cols=64    dense_vars=11   before=96.97 us   after=9.26 us    10.5x
row_vars=24  cols=256   dense_vars=13   before=801.5 us   after=76.03 us   10.5x
row_vars=40  cols=1024  dense_vars=15   before=5.186 ms   after=0.507 ms   10.2x
```

So 10.5x rather than 15x here, on your shapes. I did not chase the gap further; the mechanism is the one you identified and the win is real either way.

Equality is pinned by more than three shapes. The full sixteen-symbol enumeration now lives in the test module as the reference, and a proptest checks the shipped loop against it over random row bounds 3..=9, random four-column height vectors, and random row/column/dense points -- so the cases where the row bound exceeds the dense arity, which is where the hoisting saves the most, are covered by construction rather than by luck.

sumcheck/src/jagged/mod.rs
tcoratger: Written, and it bites.

`sumcheck/src/jagged/mod.rs` now carries `jagged_evaluation`, built only from the column heights and an `equality_weight` helper that expands the multilinear equality basis straight from its definition. It never mentions a selector, a dense index, a branching program or a sumcheck, so it cannot drift with any of them. Two proptests consume it: one at `row_variables = 3` with `heights in 0..=8`, which sweeps the degenerate shapes you list, and one at `row_variables = 20` with `heights in 0..=3`, which is the `row_variables > dense_variables` support and runs the full prove/verify round trip. Both also check the dense claim against the padded witness.

I verified your closed-loop claim rather than taking it: reversing the column convention in *both* `table()` and `evaluate()`, so the two move together, now fails `the_reduction_proves_the_jagged_table_evaluation` and `a_row_bound_far_above_the_dense_arity_round_trips` while every selector-internal test stays green. Before this commit that mutation left the whole suite green.

One structural note: the statement is no longer something the prover computes (see the `prove` signature thread), so the reference value is now an *input* to `prove` and a disagreement is a hard `ClaimMismatch` before the transcript is seeded. That makes the sparse side pinned by the protocol and not only by an assertion.

sumcheck/src/jagged/layout.rs
tcoratger: Agreed on the analysis, and worth being precise about which remedy actually pins anything, because two plausible ones do not.

Absorbing `dense_evaluation` into the sponge does not pin it. It is the last thing bound and no challenge depends on it, so the prover still chooses it after seeing every challenge.

Rejecting the degenerate case -- the remedy on the word-shift PR -- does not apply here. The degenerate case is `selector_evaluation == 0`, and it is not what enables this. The inversion works precisely *because* the selector value is nonzero. Rejecting zero would remove nothing and would cost completeness. (Separately, `selector_evaluation == 0` is harmless: forcing `terminal` to zero at a random challenge forces the whole chain back to `claimed_value == 0`, so that branch only ever proves a true statement, and it already sits inside the `2m / |EF|` bound.)

Nothing inside `verify` can pin it. The only thing that does is the caller opening its commitment at the returned point. So the remedy has to be at the API boundary, and I took all three of yours:

- `#[must_use]` on `JaggedDenseClaim` itself. This is not cosmetic here: under the workspace's `-D warnings`, `layout.verify(&point, value, &proof, challenger)?;` is now a compile **error**, not a warning. I checked by writing exactly that line: `error: unused JaggedDenseClaim that must be used`.
- A `# Security` section on `verify` stating the failure mode rather than the obligation -- that dividing the surviving value by the publicly computable selector weight satisfies the relation for any claim over any witness or none, and that discarding the claim is unconditional acceptance rather than a weakened check.
- The arity and ordering obligation, and the padding property: evaluations past the live area are unconstrained because the selector vanishes there.

`selector_evaluation != 0` is now asserted in `malformed_proof_shapes_and_terminal_values_reject`, before the mutations, so that test cannot go vacuous.

sumcheck/src/jagged/layout.rs
tcoratger: Done -- the claim is now `dense_witness.iter().zip(&selector).map(|(&cell, &weight)| weight * cell).sum()`, an `EF * F` product over `area` rather than an `EF * EF` product over `dense_capacity`.

I measured the zip alone rather than the zip plus the widen, since as you say the widen stays for the `ProductPolynomial` either way. BabyBear / EF4, 10 iterations, release, both sides asserted equal:

```
area = 2^20,     capacity = 2^20     9.17 ms -> 4.58 ms   2.0x
area = 2^20 + 1, capacity = 2^21    20.20 ms -> 3.68 ms   5.5x
```

So 2.0x from the `EF * EF` -> `EF * F` change alone, and 5.5x on the shape where the padding actually doubles the envelope. With the widen folded in I reproduce your 2.4x at `area = 2^20`, close to your 2.8x.

sumcheck/src/jagged/layout.rs
tcoratger: Taken. `prove` now takes `claimed_value: EF` and returns `JaggedError::ClaimMismatch` when the witness disagrees, before the transcript is seeded.

Two consequences beyond the signature. `sparse_value` is gone from `JaggedProverOutput` -- the caller supplied it, so returning it was redundancy that invited exactly the confusion you describe. And the sparse side of the reduction is now pinned by the protocol rather than by a test: a witness whose jagged evaluation differs from the claim is refused outright, which is what closes the gap in the other thread.

One deviation: the variant carries no payload. `JaggedError` has no field-type parameter, and giving it one would push `JaggedError<EF>` onto `validate_point`, which has no extension field in sight. The repo's own precedent for a value-mismatch error is `BooleanPcsError::SurvivingClaim` -- "the opened value is not the claim the reduction left behind" -- which likewise carries nothing, on the reasoning that a pair of random field elements adds nothing to the diagnosis. The win you were after is that the failure is now reported where it can be explained, and it is.

sumcheck/src/jagged/error.rs
tcoratger: All six are now asserted, and the three pre-transcript ones are asserted on both sides.

`malformed_prover_input_is_rejected_before_the_transcript` covers `DenseLengthMismatch` with your exact fixture (`[3, 0, 5, 1]`, eight cells against nine), `RowPointWidthMismatch { expected: 3, actual: 4 }` and `ColumnPointWidthMismatch { expected: 2, actual: 1 }`, and re-runs the two point checks through `verify` as well, so moving either `validate_point` below the transcript construction now fails on both paths.

`unrepresentable_geometry_is_rejected_at_construction` covers `RowVariablesOverflow` (`JaggedLayout::new(usize::BITS, &[1])`), `AreaOverflow` (two columns at `usize::MAX / 2 + 1`) and `DenseAreaOverflow` (`[usize::MAX / 2 + 1, 1]`, whose area has no representable power-of-two envelope).

A seventh while I was here: `PowWitnessCountMismatch` had no test either. It is now asserted in `malformed_proof_shapes_and_terminal_values_reject` by pushing a stray witness onto an honest proof -- which also keeps that path covered after the duplicate block was removed.

sumcheck/src/jagged/layout.rs
tcoratger: Confirmed and removed. `verify_rounds` checks the round count, then computes `expected_witnesses = if pow_bits > 0 { expected_rounds } else { 0 }` and raises the identical error, both before it constructs a transcript, and `verify` passes `pow_bits = 0`. The block could not fire.

The comment that replaces it says where the rule is enforced rather than implying it is enforced here. The only observable change is ordering: a proof that is wrong in both the round count and the witness count now reports `RoundCountMismatch` first. The path is no longer untested either -- `malformed_proof_shapes_and_terminal_values_reject` now pushes a stray witness onto an honest proof and asserts `PowWitnessCountMismatch { expected: 0, actual: 1 }`.

sumcheck/src/jagged/selector.rs
tcoratger: Added, essentially your text, as a doc block on `evaluate` -- the layer table with the carry and comparison recurrences, the initial and accepting states, and the three facts underneath: that accepting on the comparison bit gives a strict inequality and that the lower bound is free, that the inclusive extra layer absorbs an endpoint's overflow bit and forces the final carry to vanish, and that the conjunction is exactly the row bound on the column.

The `m` clarification is in both soundness sections: `m` is now stated as the base-two logarithm of the padded live area, explicitly not the row bound, on `prove` and by reference on `verify`.

sumcheck/src/jagged/transcript.rs
tcoratger: Both have the reason now. On the verifier: "No step can advance the cursor over that empty wire, so this failure is unreachable." On the prover's `finish`: "No step writes to it, so a nonempty wire would mean the description above had drifted."

I left them as `expect` and `assert!` rather than making them total. Both are statements about the interaction pattern two dozen lines up, not about anything the proof carries, so a silent `Ok` would hide a description that had drifted rather than protect a verifier from a malicious input.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,425 @@
...
+    ///
+    /// Acceptance returns the single dense claim the underlying PCS must authenticate.
+    ///
+    /// # Security
+    ///
+    /// The terminal relation is one equation in the surviving sumcheck value and the claimed dense evaluation, and dividing the first by the publicly computable selector weight satisfies it for any claimed sparse value, over any witness or none.
+    ///
+    /// What rules that out is the caller opening its commitment at the returned point and finding exactly the returned value, so discarding the returned claim is not a weakened check but unconditional acceptance.
+    ///
+    /// The commitment must name as many variables as the dense arity and hold this layout's live cells in column-major order, while evaluations past the live area are unconstrained because the selector vanishes there.
```
Nashtare, line 268: True for the verifier, but `prove` always pads with zeros (lines 217-220), and its `dense_claim.value` is the MLE of the zero-padded table at the returned point. An integrator who reads "unconstrained" and commits with nonzero padding (mask values, say) gets an honest claim that does not match the opening, so the proof fails to verify. Could this say that the verifier accepts any padding, but the prover's claim opens only against a commitment that is zero past the live area? The alternative is to let `prove` take the padded witness.

Separately, and as far as I can tell, the layout is bound only into this reduction's seed, while the sparse point arrives as an input. Would one clause here be worth it, saying the heights must be fixed in the transcript (normally with the commitment) before that point is drawn?

sumcheck/src/jagged/selector.rs
```diff
@@ -0,0 +1,461 @@
...
+/// Materializes the first equality weights without allocating the full row cube.
+fn equality_prefix<F: Field>(point: &[F], length: usize) -> Vec<F> {
+    // Each coordinate doubles the lexicographically ordered table.
+    let mut weights = Vec::with_capacity(length);
+    if length == 0 {
+        return weights;
+    }
+    weights.push(F::ONE);
+
+    for &coordinate in point {
```
Nashtare: Once the table saturates at `length`, every remaining coordinate still rewrites all `length` entries. The cost is therefore about `row_variables * max_height` rather than `2 * max_height`.

Take a 40-variable row bound (the shape `sparse_row_weights_do_not_allocate_the_row_bound` advertises) and a tallest column of 2^20 rows. Coordinates 21 to 40 then do about 2.1e7 extension multiplications, where the whole prefix needs about 2^21. The allocation stays bounded; only the time doesn't.

Every row below `length <= 2^k` has its top `R - k` bits zero. Could those coordinates fold into the scalar `prod (1 - y_i)`, so that only the last `k = log2_ceil(length)` coordinates are expanded (e.g. `Poly::new_from_point(&point[R - k..], scale)` truncated to `length`)?

sumcheck/src/jagged/selector.rs
```diff
@@ -0,0 +1,461 @@
...
+        sparse_point: &JaggedPoint<F>,
+        dense_point: &Point<F>,
+    ) -> F {
+        let row_point = sparse_point.row();
+
+        // One extra zero layer checks the final carry and the endpoint's overflow bit.
+        let top = row_point.num_variables().max(dense_point.num_variables());
+
+        // Only the two secret coordinates carry a nonzero equality factor, and both are shared by every column.
+        // Hoisting their four products out of the column loop is the whole difference from a per-column automaton.
```
Nashtare: Nit: "the whole difference from a per-column automaton" describes the change rather than the code. Also, the public factors are nonzero (exactly 1) on the matching symbol; what sets them apart is that they are 0/1. Maybe: "The four (row, dense) products depend only on the layer, so every column shares them." Along the same lines, "gives each point one zero variable" at line 135 holds only for the longer point; the shorter one gets `|R - D| + 1`.

sumcheck/src/jagged/layout.rs
```diff
@@ -0,0 +1,425 @@
...
+        assert_eq!(JaggedLayout::new(0, &[]), Err(JaggedLayoutError::NoColumns));
+
+        // The empty trace is the constant zero polynomial over one dense cell.
+        let empty = JaggedLayout::new(3, &[0, 0]).unwrap();
+        assert_eq!(empty.area(), 0);
+        assert_eq!(empty.dense_capacity(), 1);
+        assert_eq!(empty.dense_variables(), 0);
+
+        // Keep the allocation alive until every assertion has read it.
+        drop(vec![empty]);
```
Nashtare: Nit: this is a no-op, since `empty` lives to the end of the scope anyway; could it go?

sumcheck/src/jagged/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+        let (layout, _, _, point) = fixture();
+        let moved = JaggedLayout::new(3, &[2, 1, 5, 1]).unwrap();
+        let value = EF::from_u64(42);
+        assert_eq!(moved.dense_variables(), layout.dense_variables());
+
+        assert_eq!(
+            seeded_challenge(&layout, &point, value),
+            seeded_challenge(&layout, &point, value)
+        );
+        assert_ne!(
```
Nashtare: `encode_statement` also binds the sparse point and the claimed value (the comment at `transcript.rs:69-70` says so). Dropping either from the encoding would still pass every test, because `changing_any_public_statement_component_rejects` rejects arithmetically. Could this test add two `assert_ne!`s through the same helper, one for `value + EF::ONE` and one for a single row coordinate `+ EF::ONE`?

Nashtare (approved):
I re-checked the two new commits against my earlier review. The rebase itself was clean: `git range-diff` shows the five earlier commits unchanged.

The pruned selector is exact. The start and end factors were `from_bool` bits, so on every symbol they are exactly 0 or 1 in any field, characteristic two included. Keeping only the four symbols whose public bits match therefore drops terms that are identically zero, for arbitrary row and dense coordinates. No automaton state is dropped: all four states are still computed at every layer. The layer count (`0..=top`), the carry-0 accepting state and the `less` latch are unchanged.

To be sure, I ran a probe comparing `evaluate` against `sum_c eq(z,c) sum_{r<h_c} eq(y,r) eq(x, start_c + r)`, and against the materialized table, at random points. It covered 4094 layouts each over BabyBear^4 and `BinaryField128`:
- row bounds 0 to 6, 10, 20, 40, 62 and 63;
- one to eight columns;
- empty columns and a single full column;
- power-of-two areas whose last column ends at `2^m`.

I got 0 mismatches and about 91% nonzero values. A control that moves one row between columns changed the value every time (1378/1378 and 1352/1352).

The new tests bite. `jagged_evaluation` is independent of the selector code, and `ClaimMismatch` makes `prove` refuse a `table()` that drifts from it. An honest EF4 round trip only verifies if `evaluate` agrees with the table's MLE at the challenge point. The "before the transcript" orderings are pinned too: moving a check after seeding would hit the unfinalized-state `Drop` panic.

The binary-point characteristic-two test, the seed-separation test, dropping `JaggedShape`, the `|EF|` sentence and the dense-arity `m` close my earlier threads. Deferring `new_packed` is fine: `Poly::pack` does need a size guard.

On a fresh pass I found no way to move the layout, no verifier panic from proof input, and nothing missing from `2m/|EF|`. The branch still merges cleanly into main.

The inline comments are non-blocking follow-ups. Only the first two have substance: the padding and layout contract in the new `# Security` section, and the prover cost of `equality_prefix` under a large row bound.

sumcheck/src/jagged/layout.rs
tcoratger: You are right, and the gap was real: an integrator committing mask values in the padding got an honest claim that did not open against their commitment.

Rather than document the obligation I took your alternative and removed it. `prove` now takes the committed vector whole, of length `dense_capacity()` rather than `area()`, and lifts it into the sumcheck unchanged. Nothing needs to be zero: the selector vanishes past the live area, so the sparse sum is unaffected, while the terminal dense evaluation is the MLE of exactly the vector that was passed in. The verifier doc's "unconstrained" is now literally true on both sides.

`padding_is_free_but_the_claim_names_the_committed_vector` pins it: two witnesses differing only in one padding cell state the same sparse claim and both verify, but each delegated claim opens against its own vector and not the other's. The shared fixture now carries nonzero padding throughout, so every other test exercises the general case too, and the empty-trace test commits a single nonzero padding cell and gets it back in the claim.

On the second point, added to the verifier's `# Security`: the column heights must already be bound into the transcript, normally by that commitment, before the sparse point is drawn.

sumcheck/src/jagged/selector.rs
tcoratger: Reproduced, and you are right. Timing `equality_prefix` at a 2^20 table on the head of this PR, release build:

```
row variables  20 -> 5.74 ms
row variables  30 -> 22.2 ms
row variables  40 -> 41.9 ms
```

Linear in the row bound past saturation, exactly as you described, so this was a loop defect and not a doc defect.

Fixed the loop, along the lines you suggested. Every index below `length` has its top `R - k` bits zero for `k = log2_ceil(length)`, so those coordinates contribute the single scalar `prod (1 - y_i)` and only the trailing `k` are expanded, through `Poly::new_from_point(&point[R - k..], scale)` truncated to `length`. Cost is now `2^k + (R - k)` rather than `2^k + (R - k) * length`.

Same measurement after:

```
row variables  20 -> 3.43 ms
row variables  30 -> 3.31 ms
row variables  40 -> 2.87 ms
```

Flat in the bound, and 14.6x faster at 40 variables. The 20-variable case also improves because `new_from_point` uses the batched kernel instead of a hand-rolled interleave.

`equality_prefix_matches_the_full_table` now also checks a forty-variable point at lengths 1, 2, 3, 8 and 17 against a weight written straight from the product definition, since there is no full cube to compare against at that width and the lengths straddling a power of two catch an off-by-one in the count of expanded coordinates.

sumcheck/src/jagged/mod.rs
tcoratger: Confirmed your mutation claim first. Replacing the point-and-value loop in `encode_statement` with an empty iterator leaves all 20 jagged tests green, and so does dropping either component alone, because `changing_any_public_statement_component_rejects` rejects arithmetically.

Extended `the_column_geometry_separates_two_transcript_seeds` through `seeded_challenge`, as you suggested, and renamed it `every_public_statement_component_separates_two_transcript_seeds`: it now takes one baseline challenge and three `assert_ne!`s against it, for the moved row, for a single row coordinate plus one, and for the claimed value plus one.

Re-ran the three mutations against the new test: dropping the point fails it, dropping the value fails it, dropping both fails it, and no other test in the module moves.

sumcheck/src/jagged/selector.rs
tcoratger: Both taken. The two lines collapse to yours: "The four products of a row and a dense coordinate depend only on the layer, so every column shares them." You are right that nonzero was the wrong property to name, since the public factors are 1 on the matching symbol; what the pruning actually relies on is that they are 0/1.

The line about the top layer also went, since it only held for the wider point: "Reading one layer above the wider point leaves both points at zero there, which absorbs the overflow bit of an endpoint at the dense capacity and forces the last carry to vanish."

sumcheck/src/jagged/layout.rs
tcoratger: Gone, along with the now-unused `alloc::vec` import in that test module.

PR #2226, opened 2026-09-19
feat(word): reduce shifted operands to one Boolean opening

tcoratger:
## What

Add the E7.3 reduction from shifted word operands to one opening of the committed Boolean word trace.

The reusable key owns the checked `ConstraintSystem`, compiles its sparse wiring once, and supports both 32-bit and 64-bit words.

## Protocol

The transcript first binds the verifier-derived dimensions, public words, the shared constraint and bit points, and all eight operand claims.

It then samples two batching coordinates for the three relation families and four operand roles, subtracts the exact public-word contribution, and runs:

1. one quadratic sumcheck over the within-word bit index;
2. one quadratic sumcheck over the committed-word index; and
3. one terminal check against the verifier-known shift wiring.

The result is one `(point, value)` claim in word-major, bit-minor Boolean-PCS order.

Shift spellings are fixed by the verifier key, so the reduction sums them exactly instead of sampling eighteen shift-code variables. For a composed shift `outer(inner(word))`, the verifier evaluates the adjoint in the required order, `inner^T(outer^T(output))`.

## Soundness

Before PCS binding, the algebraic error is

```text
4 / |EF| + 2 (log2(word bits) + log2(padded witness words)) / |EF|.
```

The first term batches relation families and operand roles. The second accounts for every quadratic sumcheck round. Both terms are exposed through `p3-security`.

The returned claim is deliberately unauthenticated. The caller must bind the Boolean-PCS commitment before this reduction and open this exact point and value afterward.

## Measurements

Criterion, 64-bit fixture on the development host:

| Witness words | Prove |
| ---: | ---: |
| 2^10 | 2.70 ms |
| 2^14 | 40.67 ms |

The retained sparse set-bit scatter was 9–10% faster than a dense bit scan. Factoring the two-slot transpose retained the 2^10 result and improved the 2^14 case by about 1.9%. A one-shot 2^14 run peaked at 10.4 MiB RSS.

## Validation

- scalar-reference property tests for random witnesses;
- every accepted 64-bit shift in both composition slots, including lane-local and sign-extending shifts;
- genuine non-Boolean binary-field points for coordinate and transpose tests;
- mutations of every public operand claim, public words, both sumcheck bridges, the final opening, and constraint order;
- empty and malformed shapes;
- transcript-base/challenge-extension separation;
- focused tests for `p3-word`, `p3-word-backend`, and `p3-security`;
- all-target, all-feature Clippy with warnings denied;
- nightly formatting, rustdoc, diff checks, and Criterion bench compilation.

An independent cryptographic audit found no Critical, High, or Medium issue. Its one Low test-coverage finding was fixed in `f8f43d0e` before this PR was opened.

---

Part of #2237.


word-backend/src/shift/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+                transcript.abort();
+                return Err(error.into());
+            }
+        };
+
+        // The wiring factor is public and is evaluated directly from the compiled relation graph.
+        let shifts = shift_evaluations(&self.layout, claim.bit_point(), bit_point.as_slice());
+        let wiring = wiring_evaluation(&self.layout, &prepared, &shifts, word_point.as_slice());
+        transcript.finish();
+        if final_claim != proof.witness_evaluation * wiring {
```
Nashtare, line 315: `witness_evaluation` is the only proof field never absorbed into the sponge, and it is pinned solely by this equation. When `wiring == 0` the check degenerates to `final_claim == 0` and the prover may put any value in `witness_evaluation`; `verify` then returns that value as the opening claim.

`wiring_evaluation` is identically zero when the system has no witness-segment references, and otherwise vanishes with probability about `word_variables/|EF|` at the sampled `r_word`. Not exploitable when the caller discharges the opening through the PCS — the Boolean PCS catches it — but "the reduction verified" then does not imply the returned value is determined, which is a surprising property for something handed out as a claim.

`witness_evaluation` is a deterministic function of already-absorbed values (`final_claim / wiring`), so absorbing it costs nothing, and rejecting `wiring == 0` outright would also do.

word-backend/src/shift/claim.rs
```diff
@@ -0,0 +1,150 @@
...
+            .chain(&self.bitwise_and)
+            .chain(&self.integer_mul)
+            .copied()
+            .collect()
+    }
+}
+
+/// One evaluation claim on the committed bit trace.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ShiftOpeningClaim<F> {
```
Nashtare, line 97: Soundness rests on two caller obligations that are documented but not enforced: absorbing the Boolean-trace commitment into the challenger *before* `prove`/`verify` (otherwise the sumcheck challenges do not depend on the commitment and the prover can choose the trace after seeing them — a total break), and discharging this claim through the PCS afterwards (otherwise the reduction proves nothing about committed data).

`ShiftOpeningClaim` is an ordinary public struct, so `let _ = key.verify(...)?;` compiles. `p3-bus`'s analogous type in #2224 carries `#[must_use = "leaf claims must be tied to committed columns"]`; the same attribute here would at least make the second obligation visible at the call site.

Related, a smaller one: `word_variables()` is private, so a caller has to re-derive `system.witness_len().max(1).next_power_of_two().ilog2()` to split the returned point into its word and bit halves, and to know the trace must be padded to exactly `2^word_variables` words. That padding convention is load-bearing for the opening and currently appears only in a code comment in `wiring_evaluation`. Exposing it, or returning the split, would help.

word-backend/src/shift/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+    /// Returns the number of variables covering the committed word segment.
+    fn word_variables(&self) -> usize {
+        // An empty segment is represented by one zero padding slot.
+        self.system.witness_len().max(1).next_power_of_two().ilog2() as usize
+    }
+
+    /// Builds the numeric soundness model from the executed schedule.
+    fn security_model(&self, field_bits: NonZeroUsize) -> WordShiftSecurityModel {
+        // Two two-variable batching axes and both quadratic round counts are exact.
+        WordShiftSecurityModel::new(
```
Nashtare, line 374: Two silent-desync hazards in the accounting, neither wrong today.

The `4` is a literal rather than `OPERATION_VARIABLES + OPERAND_VARIABLES`. Adding a fourth relation family would bump `OPERATION_VARIABLES` and leave the reported error at `4/|EF|` while the real batching degree grows.

And `field_bits` is a caller-supplied `NonZeroUsize` with nothing tying it to the `EF` actually used in `prove`/`verify` — the key has no `EF` parameter. A caller who passes the *base*-field width (64 for a `BinaryField64` transcript with `BinaryField128` challenges) gets a bound wrong by 64 bits in the unsafe direction, with no compile-time signal. `p3-bus`'s `BusPlan::security_term` at least documents that `field_bits` is a lower bound on `log2 |EF|` and must cover every challenge; the doc here drops that. Worth restating, and worth noting that this term excludes the error of whatever produced `constraint_point` and `bit_point`.

word-backend/src/shift/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        let mut opening = proof;
+        opening.witness_evaluation += F::ONE;
+        assert_eq!(
+            key.verify(&public, &claim, &opening, &mut challenger()),
+            Err(ShiftReductionError::FinalClaim)
+        );
+    }
+
+    #[test]
+    fn changing_any_public_operand_claim_is_rejected() {
```
Nashtare, line 549: This test and mutation 3 above both change a value that feeds `expected_claim` / `public_contribution` directly, so the rejection comes from `proof.bit_sumcheck.claimed_sum != expected_claim` — arithmetic, not the transcript. Dropping the `CLAIMS` step or the `separator.instance(word)` loop would leave both green, even though the property they advertise ("precedes both batching challenges", "bound before batching") is exactly the Fiat-Shamir one.

The binding is present — I read `transcript.rs` — it is just untested. A test that keeps the claim arithmetic fixed and varies only what the transcript absorbs (asserting the sampled `BatchWeights` change) would pin it directly.

word-backend/src/shift/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        // The reduction must not force commitments and challenges into one field type.
+        let (key, witness, public, claim) = fixture(13, [17, 19, 23]);
+        let mut prover = base_challenger();
+        let (proof, expected) = key.prove(&witness, &claim, &mut prover).unwrap();
+        let mut verifier = base_challenger();
+        let actual = key.verify(&public, &claim, &proof, &mut verifier).unwrap();
+
+        assert_eq!(actual, expected);
+    }
+
```
Nashtare, line 635: The two `Word32` tests use empty constraint systems (shape rejection and the zero-opening edge case), so the 5-round bit phase with real shifts never runs end to end.

And the `Word64` fixture covers `RotateRight`, `ArithmeticRight`, `Lane32RotateRight` and the pair `(LogicalLeft 9, RotateRight 17)`, which leaves `Lane32ArithmeticRight`, `Lane32LogicalLeft/Right` and full-width `LogicalRight` without an end-to-end proof. `Lane32ArithmeticRight` is the second sign-extending operator (per-lane fan-out into bits 31 and 63), and the only thing covering it is `factorized_transpose_matches_every_shift_pair` — which compares two implementations that both call `Shift::apply`, so it validates the factorization and slot order but not that the transpose convention agrees with the forward evaluation used to build the claim. Only the end-to-end tests cross-check those two.

word-backend/src/shift/wiring.rs
```diff
@@ -0,0 +1,422 @@
...
+                (0..64)
+                    .filter(|&bit| (shifted >> bit) & 1 == 1)
+                    .map(|bit| output[bit])
+                    .sum()
+            })
+            .collect()
+    }
+
+    #[test]
+    fn factorized_transpose_matches_every_shift_pair() {
```
Nashtare: Nit, CI time: `shifts()` yields 384 spellings (4 full-width kinds x 64 plus 4 lane kinds x 32, including 8 duplicate identities), so this is 384^2 = 147,456 pairs, each doing three 64x64 basis walks and three `Vec<F>` allocations — on the order of 1.8e9 `BinaryField128` additions. Fine in `--release`, but under the default `cargo test` profile this one test will dominate the crate. Hoisting the duplicate identity spellings, or sampling the pair space, keeps the coverage.

Nashtare (approved):
I went after the two things most likely to be wrong here — the adjoint and the batching bound — and both hold up. Recomputed the security term independently; it matches the stated formula exactly.

**The adjoint is right, including the case I expected to be wrong**

`shift_transpose` does not hand-write an adjoint per shift kind: it derives the matrix structurally by applying `shift.apply` to each basis bit and summing the output weights at the set bits. So for `ArithmeticRight(k)`, `apply(e_{W-1})` has bits `W-1-k ..= W-1` set and column `W-1` of the adjoint correctly *accumulates* `k+1` output weights into the top position — the replicated-sign case falls out of the construction rather than needing a special case. Same for the lane-local variants.

Composition order is also right and genuinely pinned. `shift_weights` computes `M_in^T(M_out^T(output)) = (M_out M_in)^T`, matching the forward `shifts[1].apply(shifts[0].apply(w))` used by `ShiftedValue::apply` and by `public_contribution`. The fixture's pair `(LogicalLeft 9, RotateRight 17)` does not commute, so an inverted composition would break the end-to-end tests, and `factorized_transpose_matches_every_shift_pair` exhausts all ordered pairs on top of that.

Shift boundaries are clean too: `Shift::new` normalizes amount 0 to the single `LogicalLeft(0)` spelling and rejects `amount >= width`, so nothing wraps or aliases, and `ShiftCode` packs kind and amount injectively.

**Security bound: `4/|EF| + 2(log2(word bits) + log2(padded words))/|EF|` is exact**

The batching is not powers-of-a-challenge, which is what I went looking for. `OPERATION_VARIABLES = 2` and `OPERAND_VARIABLES = 2` independent extension coordinates are drawn and each expanded into a full `equality_weights` table, so statement `(family o, role p)` carries `eq_alpha(o) * eq_beta(p)`. `operand_index` and `ShiftClaim::batched` agree exactly on the map — Zero->(0,0); And->(1,{0,1,2}); IntegerMul->(2,{0,1,2,3}) — so all eight used slots get distinct `(o,p)` pairs and the eight products are linearly independent. No two statements share a coefficient, so no cross-family cancellation. A nonzero claim error is then a nonzero multilinear in 4 variables of total degree 4, giving exactly `4/|EF|`. The unused `(3, *)` slot does not weaken it.

Ordering is correct: the seed carries the pattern hash plus `[constraint_variables | bit_variables | public_words | witness_words]` and every public word, then `constraint_point`, `bit_point` and all eight claims are absorbed, and only then are the two batching axes drawn. Every dimension comes from the key-owned `ConstraintSystem`, never from the proof.

The sumcheck term is also exact. `b = W::BITS.ilog2()` is the real bit-phase round count and `w = word_variables()` equals `word_prover`'s `capacity.ilog2()`, so prover and verifier agree and there is no off-by-one when `witness_len` is not a power of two — `ceil(log2 n)` is simultaneously the round count and `log2(padded_words)`. Both phases are degree 2, `security/src/word.rs` charges `2(b + w)`, no round is dropped, and `error_from_numerator` rounds the numerator up via `next_up()`, i.e. conservatively.

The terminal check correctly carries no error: `wiring_evaluation` is computed by the verifier from the key and public data alone. And the bit phase's terminal claim is discharged by the word phase rather than checked directly, which is a genuine algebraic identity (regrouping keys by source word), so no extra term is needed there either.

**Also verified**

- **Proof-shape hardening.** Both outer `claimed_sum` values are compared against verifier-derived quantities before the corresponding sumcheck is replayed; `GenericDegreeProof::verify` rejects a wrong round count and witness count before replaying; the per-round `Length::Fixed(degree)` step rejects a wrong-width round polynomial without a drop-time panic; every verifier error path calls `transcript.abort()`.
- **Base/extension separation.** Public data is absorbed as base coefficients under `Interaction::algebra::<F, F>` and challenges under `::<F, EF>`; the pattern records `Algebra{modulus, degree}` per step and the pattern hash enters the seed, so a swap changes the transcript. Each sumcheck re-seeds under its own separator carrying `(num_rounds, degree, pow_bits)`.
- **No reachable panic from proof data.** Every `expect` and index is keyed off verifier-owned key dimensions or off checks that run before any transcript step; `trailing_zeros()` on `to_u64()` is below `W::BITS` for both widths; `operand_index`'s `unreachable!` is genuinely unreachable, since `ConstraintSystem` validates each term's `(segment, position)` and neither `CompiledKey` nor `ConstraintReference` has a public constructor.
- **Characteristic 2 is enforced**, not assumed: the reduction sums per-term evaluations while the claim is their XOR, and both entry points carry `EF: Algebra<Gf2>`.

**Comments**

Nothing blocking; the inline notes are one degenerate case in the terminal check, the two caller preconditions, and test coverage. Approving.

word-backend/src/shift/mod.rs
tcoratger: Both changes are in, and only one of them fixes the property you describe.

Absorbing `witness_evaluation` does not make it determined. It is absorbed after the last challenge this reduction draws, so nothing inside the reduction depends on it. It is still worth doing for a different reason: the challenger is handed straight back to the caller, who runs the PCS opening on it, and without the absorption those opening challenges do not depend on the value being opened. So the new `trace_evaluation` transcript step is about the composition, not about this equation.

What fixes the equation is rejecting the vanishing wiring factor, and that is now a hard error in both roles (`ShiftReductionError::DegenerateWiring`), checked before the closing comparison in `verify` and against `first_constants()` in `prove`. The cost is that a system with no witness-segment reference can no longer be proven at all, which is the honest answer: at `wiring == 0` the reduction has learned nothing about the trace, so it should not hand out a claim.

`a_system_with_no_committed_reference_pins_no_opening` pins it. It builds a proof by hand for the empty `Word32` system (five all-zero quadratic rounds, no word round) and checks that three different `witness_evaluation` values are all rejected. I verified the test is not vacuous: with the two `is_zero` checks removed, all three forged proofs verify and `verify` returns the forged value verbatim.

word-backend/src/shift/claim.rs
tcoratger: `ShiftOpeningClaim` now carries `#[must_use = "leaf claims must be tied to committed columns"]`, matching the sibling type.

Both obligations are now stated on both entry points rather than one of them: `prove` and `verify` each say the trace commitment must be absorbed into the challenger first, `prove` spells out that skipping it lets a prover choose the trace after seeing the challenges, and `verify` says the returned claim must be discharged through the commitment scheme because the reduction alone proves nothing about committed data.

`word_variables` is public and documents the padding convention: it is the number of leading word coordinates in the returned point, the trace must be padded with zero words to exactly that many slots, and the remaining coordinates are the within-word ones. I did not add a separate split accessor, since a caller with that count can split the point itself.

word-backend/src/shift/mod.rs
tcoratger: The literal is gone; the model now takes `OPERATION_VARIABLES + OPERAND_VARIABLES`, so a fourth relation family moves the reported term with it.

The doc on `security_term` now restates what the sibling type documents: the argument is a lower bound on the base-two logarithm of the challenge-field order, every batching and sumcheck challenge must come from that same extension field, and the term excludes both the error of whatever produced the two statement points and the binding and authentication of the returned opening.

I did not make the key generic over `EF`. It would be checked rather than documented, but the key is built and cached independently of the field the proof later runs in, so adding the parameter is a redesign rather than a fix, and out of scope here.

word-backend/src/shift/mod.rs
tcoratger: Confirmed before fixing. Deleting the `CLAIMS` step (pattern entry plus both add/observe calls) and the `separator.instance` dimension loop leaves all 28 remaining tests green, exactly as you said.

`every_absorbed_statement_input_moves_the_batching_weights` now tests the binding directly. It only samples: it builds the transcript, draws `BatchWeights`, and plays out the pattern, with no claim arithmetic anywhere. It then varies one absorbed input at a time and asserts the sampled weights move -- one operand claim value, the committed-segment length (which no claim value mentions), and one public word.

Each mutation is caught by its own assertion: removing the `CLAIMS` step fails the first, removing the dimension loop fails the second.

word-backend/src/shift/mod.rs
tcoratger: Two end-to-end tests added.

`word32_shifts_close_the_five_round_bit_phase` runs a real 32-bit system through both phases: two committed words, all four full-width movements including a two-slot pair, with a high bit set in each word so the sign-extending one fans out.

`lane_local_and_full_width_logical_right_close_end_to_end` covers the four operators with no end-to-end proof: `Lane32ArithmeticRight`, `Lane32LogicalLeft`, `Lane32LogicalRight` and full-width `LogicalRight`. Both lanes of the first word carry a set sign bit, so the per-lane fan-out into bits 31 and 63 is exercised.

Your point about the cross-check is what makes these worth adding: the claim is built from `Operand::evaluate` (forward semantics) while the prover's first round message is built from the transpose path, and the verifier compares the two, so a disagreeing transpose convention rejects.

word-backend/src/shift/wiring.rs
tcoratger: Both. `shifts()` now drops the seven redundant zero-distance spellings (377 rather than 384), and the default test walks a deterministic sampled grid instead of the full square: three strides coprime to 377, each a bijection on the index set, so every shift still appears in both slots, three pairings each. The exhaustive square is kept as the same test body behind `#[ignore]`.

Crate wall clock under the default profile: 14.86s to 2.96s. The ignored sweep is 13.4s on demand.

PR #2227, opened 2026-09-19
perf: batch Boolean trace openings and keep Blake3 traces packed

Nashtare:
Opening every Blake3 column through `BooleanTracePcs` previously performed a separate ring-switch reduction for each column. This binds all column evaluations before sampling a column point, then opens the padded stacked bit polynomial once per row-point batch. The composed security estimate includes the additional column-batching error. Subset, reordered, mixed-height, and multi-table openings retain the general path.

Blake3 now generates packed words directly, using a reusable `Gf2` row buffer. Packed storage is retained through commitment and the initial zerocheck rounds; only the reduced residual expands into field elements. Dense APIs remain supported, Keccak continues to use `BinaryPcs`, and CLI defaults remain `2/3/2`.

Measured on macOS arm64 with rustc 1.98.0, optimized profile and parallel enabled, no extra RUSTFLAGS. These are medians of three independent processes at 16,384 rows and a 94-bit target, with matching explicit `--log-inv-rate 1 --folding 4 --merkle-arity 4`:

| | BinaryPcs baseline | Packed, batched BooleanTracePcs |
|---|---:|---:|
| Prove time | 4.471 s | 0.337 s |
| Verify time | 0.029 s | 0.019 s |
| Serialized proof | 1,130,372 bytes | 765,216 bytes |
| Peak process RSS | 14,023 MiB | 865 MiB |

That is about 13.3x faster proving, 32% smaller proofs, and 16.2x lower peak RSS. Prove timing preserves the existing convention, including dense transposition and excluding trace generation; RSS includes the entire process. The baseline matches the code before `eaa8dd028`, with its parameter overrides supplied explicitly. The historical 632 ms result was not reproduced in this environment; the table uses fresh matched measurements.

With the unchanged main defaults and just `--security-bits 94`, the new example also verifies at:

| Trace rows | Prove time | Proof bytes | Peak RSS |
|---|---:|---:|---:|
| 32 | 0.018 s | 324,489 | 37 MiB |
| 16,384 | 0.358 s | 515,779 | 991 MiB |

All measured proofs verified above the requested security target. A separate native-CPU build also passed, with a 0.331 s median for the matched 16,384-row case.

Validation:

- Independent task reviews and whole-branch review, including transcript ordering, soundness accounting, packed layout and zerocheck transitions.
- Full affected core suites: 1,228 passed, 3 existing ignored tests, including integration and doc tests.
- Blake3 AIR and example library suites in serial and parallel; CLI-default regression test.
- Dense/packed proof-byte and post-proof challenge equality, backend parity, tamper rejection, storage boundaries, sliced residual sizing, and early packed indexed-source rejection.
- Workspace all-targets compile check, strict clippy on all targets of the six affected packages, and formatting check.
- Actual optimized standard and native executable builds, with the requested height-5 and height-14 runs.

The optimized BooleanTrace opening transcript changes, so proofs produced by the previous complete-column path must be regenerated. Packed indexed lookup sources are rejected before commitment because the indexed witness API requires dense slices; existing dense indexed proving is unchanged.


Nashtare:
@tcoratger on an 5th Gen AMD EPYC instance, I'm getting the `blake3` variant of `prove_hash_binary` to be only twice slower than its prime field counterpart with FRI and KB field (about for 2^14 trace length). For 2^20 trace lengths, binary prover is actually 36% faster.

I'm working on adding next-row support to the ring switch so that Keccak also can benefit from the `BooleanTracePcs` next, should be ready soon.

blake3-air/src/binary/generation.rs
```diff
@@ -68,6 +71,60 @@ pub fn generate_binary_trace_rows<F: Field>(
...
+/// rows `64 * w..64 * w + 63`, with bit zero holding the first logical row. The generic field
+/// parameter controls only the reusable temporary row used while generating the witness; the
+/// resulting bits are independent of that field.
+///
+/// # Panics
+///
+/// Panics if the field does not have characteristic 2, if `inputs` is empty, or if the number of
+/// inputs is not a power of two.
+#[instrument(name = "generate packed Blake3 binary trace", skip_all)]
+#[allow(clippy::needless_pass_by_value)]
```
tcoratger, line 86: We should take a slice here as argument to avoid this allowance no?

examples/src/binary.rs
```diff
@@ -489,9 +639,7 @@ where
 /// The commitment arity is the trace's log-height plus the ceiling of the log of its width:
 /// one extra variable per doubling of the column count, since every column is stacked into a
 /// single committed polynomial.
-///
-/// The statement's security is assessed once against `options.security_bits` before proving,
-/// so the timed phases are the plain prover and verifier.
+#[allow(clippy::needless_pass_by_value)]
```
tcoratger, line 642: Same probably there is something to change to easily rm this allowance

examples/src/binary.rs
```diff
@@ -502,6 +650,110 @@ fn prove_binary_air_with<A, const N: usize, Ntt>(
...
+    trace: Table<F>,
+    options: BinaryProofOptions,
+    backend: Backend,
+) -> Result<BinaryProofReport, BinaryProofError>
+where
+    A: BinaryAir,
+{
+    match options.merkle_arity {
+        2 => prove_boolean_air_with::<A, 2>(air, trace, options, backend),
+        4 => prove_boolean_air_with::<A, 4>(air, trace, options, backend),
```
tcoratger, line 715: We can group the two branches under a single branch here no?

tcoratger (approved):
Reviewed the full packed-trace and batched-opening path, including transcript ordering, column-batching soundness, packed layout boundaries, security accounting, fallback behavior, and tests. The three inline cleanup comments are non-blocking; I found no correctness or soundness issue. All CI checks are green.

PR #2228, opened 2026-09-19
perf: prove Keccak-f through the Boolean commitment with successor openings

Nashtare:
Stacked on #2227 (`robin/batched-packed-boolean-trace`); retarget to `main` once that merges.

## What this does

`KeccakBinaryAir` reads the next row, and `BooleanTracePcs` refused any batch with next-row columns, so Keccak-f could not use the packed-bit commitment. This adds successor (next-row) openings on top of #2227's batched design, then sends Keccak-f through it with a packed trace.

- **`p3-sumcheck` (bit ring switch):** a reduction can also answer the repeat-last successor reading at its point (row `x` reads `x + 1`, the last row reads itself). When rows span more than one packed element it sends two extra tensors (`carry`, `last`). Every tensor is bound before the batching draws `r''` and one scalar `alpha`, and one degree-two sumcheck covers them. Reductions with no successor view keep their exact challenges; the proof gains one byte for the empty successor slot.
- **`p3-security`:** `bit_ring_switch_tensors_{error,term}` charge the extra tensors as degree `K - 1` in `alpha`. The column-batching term is #2227's `column_batch_*`.
- **`p3-binary-pcs`:**
  - `BooleanPcs::{open_readings, verify_readings, readings_security}` take `BitOpening { point, row_variables, current, next }`. The point API is the current-only special case.
  - `BooleanTracePcs` answers successor views on both routes. The fast route generalizes to "one table, every batch reads all columns at the current row and either no column or every column at the next row". Its transcript gains a `NEXT_VALUES` step only when a next view is present, so current-only seeds and proof bytes are unchanged (the full `next == false` seed is now pinned against digests computed from #2227's transcript). Other protocols use the per-column route. Every value is bound before the challenge that combines it, on both sides. `SuccessorView` is removed.
  - The row equality table is built once per batch and point, and the dense evaluation parallelizes over columns and row chunks.
- **`p3-keccak-air`:** `generate_binary_trace_packed` and `KeccakBinaryAir::generate_random_trace_packed` produce the trace packed 64 rows per `u64`, mirroring #2227's BLAKE3 generator. It regenerates each permutation that overlaps a 64-row block in a worker-local buffer, so it agrees with the dense generator by construction, and memory is bounded by the AIR width.
- **`p3-examples`:** Keccak-f proves through the Boolean commitment from a packed trace, like BLAKE3. `--ntt`, `NttOptions` and the now-unused `AdditiveNttChoice` are removed, since the Boolean commitment encodes through the polynomial-basis NTT only. The README describes the Boolean path for both objectives.

## Decision to confirm

- **CLI defaults.** #2227 pins the example defaults at `2/3/2` (a test plus a README sentence). This branch's `perf(example): change default parameters` sets `1/4/4`, so that test and the README sentence now pin `1/4/4`. If the defaults should stay general, drop that commit.

## Verification

- Full tests pass on the final tree for `p3-sumcheck` (587), `p3-security` (133), `p3-binary-pcs` (127) and `p3-keccak-air` (43), the last two also with `--features parallel`, plus `p3-examples` and the multi-stark `prove_boolean_trace` example (11). Clippy `-D warnings` and nightly fmt are clean.
- `prove_hash_binary` verifies for Keccak-f (debug `l=6, 10`; release `l=14, 16`) and BLAKE3 (debug `l=6`). These were smoke runs on a laptop, not benchmarks. Keccak-f in release with the packed trace: `l=14` gives a 495,119-byte proof at 119 MB peak RSS, and `l=16` gives 605,143 bytes at 454 MB. The dense trace gave byte-identical proofs at 968 MB and 3.85 GB.
- Independent reviews of the successor commit, the follow-ups and the packed generator found no Critical issues. Their Important items (a stale doc paragraph, the missing block-boundary coverage) are fixed. For the generator, the reviewer also checked the block and permutation index arithmetic against an independent model for about 600 sizes, with no discrepancies.

## Remaining, none blocking

- `ROW_GRAIN` and the nested parallelism in the dense evaluation are unmeasured; `binary-pcs/benches/boolean.rs` could put a number on them.
- The packed arm of that evaluation parallelizes over columns only, so a narrow packed table at very large height stays serial.
- The README's `cargo run ... --features parallel` lines for `prove_hash_binary` name a feature the examples crate does not define. That text predates this PR and is left as is.


tcoratger (commented):
The successor-opening construction and its transcript/security accounting look sound to me. I have one non-protocol point on the defaults decision.

examples/examples/prove_hash_binary.rs
```diff
@@ -42,7 +32,7 @@ struct Args {
     representation: RepresentationOptions,
 
     /// Log of the inverse code rate for the binary PCS.
-    #[arg(long, default_value_t = 2)]
+    #[arg(long, default_value_t = 1)]
```
tcoratger, line 35: I would keep the existing 2/3/2 defaults in this PR. The verification section says these were smoke runs rather than benchmarks, so there is not yet evidence that 1/4/4 is the right general default across trace sizes and architectures. It also makes this CLI disagree with BinaryProofOptions::default(), which remains 2/3/2. The new tuple is still available explicitly for performance runs; changing the default would be cleaner in a measured, dedicated PR if it wins broadly.

tcoratger (approved):
Reviewed the successor-opening algebra, carry/last tensors, repeat-last boundary cases, Fiat-Shamir ordering, K-tensor security term, current/next batching routes, packed Keccak generation, and tests. I found no correctness or soundness issue. The inline recommendation to retain the 2/3/2 CLI defaults is non-blocking and concerns benchmark/configuration policy only. All CI checks are green.

examples/examples/prove_hash_binary.rs
Nashtare: I'd recommand rather updating the BinaryProofOptions::default() to match 1/4/4, these seem to be always the optimal (prover-oriented) parameters on all runs I did across various instances. WDYT?

tcoratger: Okay sounds good

Issue #2234, opened 2026-09-19
Authenticate binary bus products and unify mixed-height table proving


tcoratger:
## Deliverable

Bind the characteristic-two multiset bus to committed table columns and combine bus shares with local table constraints.

## Acceptance criteria

- Product-GKR terminal claims are authenticated against prescribed PCS openings.
- Nonlinear row factors are proved as formal compositions rather than reconstructed from separate MLE openings.
- Short tables are lifted with a Boolean-cube factor whose sum is one in characteristic two.
- One back-loaded sumcheck handles mixed table heights and bus shares.
- Transcript chronology fixes commitments and public values before every reduction challenge.
- Every batching and sumcheck contribution appears in the security report.
- Binary-field, preprocessed-column, mixed-height, and adversarial round trips pass.

PR #2244, opened 2026-09-19
feat!(multi-stark): authenticate binary bus claims

tcoratger:
Part of #2234

## What changes

- materialize named binary-bus factors from symbolic AIR declarations
- prove their identity-padded push and pull products with batched radix-4 ProductGKR
- reduce ProductGKR terminal claims to nonlinear rowwise bus compositions
- authenticate those compositions through prescribed PCS openings of main and preprocessed traces
- support conditional activations, mixed table heights, and characteristic-two fields
- account for fingerprinting, ProductGKR, direction batching, composition sumcheck, and opening errors

## Soundness boundary

Push and pull remain structural metadata, rather than field signs, so they cannot cancel in characteristic two. All bus geometry is verifier-derived and transcript-bound. The direction challenge is sampled only after ProductGKR fixes both terminal claims. The terminal identity is rebuilt from commitment-authenticated openings; nonlinear payloads are evaluated rowwise before multilinear extension.

The PCS remains binding-only. This PR authenticates a multiset equality; RAM semantics are a separate stacked layer.

## Compatibility

This intentionally changes the unreleased `MultiStarkProof` postcard shape by adding an optional bus section. The changelog records the break and the pinned WHIR fixture is regenerated. Statements without a bus retain their existing transcript path.

## Focused validation

- `cargo test -p p3-bus --lib`
- `cargo test -p p3-multi-stark --lib bus`
- `cargo test -p p3-multi-stark --test whir_fibonacci verify_whir_compat_fixture`
- `cargo test -p p3-multi-stark --example prove_binary_field binary_bus_proof_round_trips`
- all-target/all-feature Clippy for `p3-bus` and `p3-multi-stark` with warnings denied
- nightly formatting and rustdoc

An independent `/audit-crypto` pass found no blocking or non-blocking security findings.


multi-stark/src/verifier.rs
```diff
@@ -257,7 +282,64 @@ where
...
+                        BusCompositionVerifierTranscript::<_, C::Val, C::Challenge>::new(
+                            challenger,
+                            BusCompositionShape {
+                                num_variables,
+                                degree,
+                                pow_bits,
+                            },
+                        );
+                    let direction = composition.direction_challenge();
+                    let expected_claim = output.batched_terminal_claim(direction)?;
```
tcoratger: **[MAJOR] `?` here drops a live `VerifierState` mid-pattern, so the verifier panics instead of rejecting**

`composition` was constructed one line earlier, which seeded the challenger and started the pattern, and `direction_challenge()` already consumed a step. If `batched_terminal_claim` returns `Err`, `?` unwinds past `composition` without `finish()` or `abort()`.

`Drop for VerifierState` (`challenger/src/fs/state/verifier.rs:57`) then does:

```text
    if !self.player.is_finalized() { panic!("Dropped unfinalized VerifierState: ...") }
```

So the failure mode is a panic, not a `VerificationError`. Every other error path in this function is careful about exactly this -- `BusSumcheck` at line 314 calls `composition.abort()` first, and the indexed branch has the comment "The driver refuses to be dropped mid-pattern, so it is released first". This one line is the exception.

Today it is unreachable: `ProductGkrProof::verify` decodes roots via `root_shape.decode` after checking `root_message_len`, so `values.len()` is always `num_trees == 2` and `TerminalValueCount` never fires. But the code is written as if it were reachable -- it uses `?` on a fallible call -- and a verifier reachability argument that lives in another crate's decode function is not one I want holding up a panic-vs-reject boundary.

**Fix:** either

```rust
let expected_claim = match output.batched_terminal_claim(direction) {
    Ok(claim) => claim,
    Err(error) => { composition.abort(); return Err(error.into()); }
};
```

or hoist the arity check above `BusCompositionVerifierTranscript::new`, so the transcript is only created once the claim is known to exist.

**Verify:** a unit test that hands the closure a `BusReductionOutput` with `values.len() == 1` must return `Err`, not unwind. Easiest as a direct test on the sub-function once it is extracted.

bus/src/argument/transcript.rs
```diff
@@ -0,0 +1,244 @@
...
+            separator.instance(&(blocks.len() as u64).to_be_bytes());
+            for block in blocks {
+                separator
+                    .instance(&(block.bus as u64).to_be_bytes())
+                    .instance(&(block.owner.air as u64).to_be_bytes())
+                    .instance(&(block.owner.declaration as u64).to_be_bytes())
+                    .instance(&(block.log_height as u64).to_be_bytes())
+                    .instance(&(block.offset as u64).to_be_bytes());
+            }
+        }
```
tcoratger, line 99: **[MAJOR] The separator that carries the entire cross-bus/cross-table separation property has no tests**

These ~30 lines are where "a claim valid for one bus is rejected for another" actually lives. The encoding looks injective to me -- every variable-length field is length-prefixed, names are bound as `(len, bytes)`, and block records are fixed-arity -- and `ds.seed()` does reach the sponge, so the binding is real.

But nothing tests it. `bus/src/argument/` has tests only for `batched_terminal_claim`. Compare `multi-stark/src/transcript.rs`, which mutates `MultiStarkShape` one field at a time and asserts `!seeds_agree(...)` for each. The bus separator binds strictly more attacker-relevant structure than `MultiStarkShape` does and has strictly less coverage.

Concretely untested: two plans differing only in `domain.name`; only in `domain.identity`; only in a block's `offset`, `log_height`, `owner.air`, or `owner.declaration`. A regression in any one of those turns a cross-table forgery into a valid proof, and the current suite stays green.

**Fix:** mirror the `seeds_agree` mutation test, then add a proptest that is strictly stronger:

```rust
// BusPlan: PartialEq, so the antecedent is free.
proptest!(|(a in plan_strategy(), b in plan_strategy())| {
    prop_assume!(a != b);
    prop_assert_ne!(seed_bytes(&a), seed_bytes(&b));
});
```

`DomainSeparator::seed_bytes` is already `pub(crate)`; a `#[cfg(test)]` accessor keeps this out of the public surface.

**Verify:** the proptest must fail if you delete any single `.instance(...)` call in this function. That is the property you want -- not that the seeds differ on some fixtures, but that no field is decorative.

bus/src/lib.rs
```diff
@@ -3,18 +3,24 @@
...
 
+pub use argument::{BusArgumentError, BusChallenges, BusProof, BusReductionOutput};
 pub use builder::{
     BusActivation, BusInteractionBuilder, BusInteractionRecorder, BusSymbolicBuilder, RecordToken,
     SymbolicBusInteraction,
 };
-pub use leaf::{BusDirection, BusLeafDeclaration, BusLeafError, BusLeaves, BusSelector};
+pub use direction::BusDirection;
+pub use evaluation::{BusEvaluation, BusEvaluationError};
+pub use leaf::{BusLeafDeclaration, BusLeafError, BusLeaves, BusSelector};
```
tcoratger: **[MAJOR] The superseded, unauthenticated `BusLeaves` path is still public -- and uses the opposite fingerprint bit order**

`BusPlan::evaluate_factor` is now the only correct way to materialize a leaf factor. `BusLeaves::materialize` is the path it replaced, and it survives as public API. Three concrete hazards:

1. **Opposite slot-bit convention.** `bus/src/leaf/mod.rs:175` binds `point[k]` to bit `k` (LSB-first); `bus/src/multilinear.rs::equality_weights_msb`, which `BusChallenges::fingerprint_weights` feeds into `evaluate_factor`, binds `point[0]` to the most significant bit. `BusLeaves::materialize(&decls, &challenges.fingerprint, challenges.offset)` compiles, type-checks, and silently computes a *different* fingerprint for any tuple width >= 4.
2. **No domain separation.** Its own doc says so: "This function does not separate distinct named buses. A caller combining buses must reserve tuple slots for an injective domain separator." That is now `BusPlan`'s job, and a caller who does not know that gets a bus system where every named bus is the same bus.
3. **No block alignment.** It concatenates declarations with no power-of-two padding, so its output can never satisfy `terminal_shares`' `prefix_index = offset >> log_height`.

Its only remaining in-tree use is one test at `bus/src/plan/mod.rs:861`, which happens to be safe because it uses a 1-coordinate fingerprint point where MSB and LSB coincide.

This PR is already `feat!`. This is the moment.

**Fix:** delete `BusLeaves`/`BusLeafDeclaration`/`BusSelector`/`BusLeafError` from the public surface, or make the module `pub(crate)` and rewrite `plan/mod.rs:861` against `evaluate_factor`. If it must stay public for a downstream consumer, at minimum rename the parameter and document the LSB convention at the call site, and add a test asserting the two conventions disagree so nobody "fixes" one to match the other by accident.

**Verify:** `cargo check -p p3-bus` after removing the re-export, plus a grep for external users.

multi-stark/src/bus/context.rs
```diff
@@ -0,0 +1,290 @@
...
+
+                // Resolve every row from the exact tables committed by this proof.
+                for row in 0..height {
+                    for (value, column) in main.iter_mut().zip(main_columns) {
+                        *value = column[row].into();
+                    }
+                    for (value, column) in prep.iter_mut().zip(prep_columns) {
+                        *value = column[row].into();
+                    }
+                    let factor = self
```
tcoratger: **[MAJOR] `evaluate_factor` in the innermost row loop costs ~20 ns/row; nearly all of it is loop-invariant work**

Measured on this branch, `--release`, BabyBear, 3-slot payload, `fingerprint_width = 4`, 2^20 iterations:

```text
    evaluate_factor          20.0 ns/row
    equivalent inner product  0.6 ns/row
```

Three loop-invariant costs sit inside the per-row call:

1. `let payload = interaction.fields.iter().map(...).collect::<Result<Vec<_>, _>>()` -- one heap allocation **per row, per declaration**. At 2^20 rows that is a million malloc/free pairs for a 3-element vector.
2. `weights.iter().enumerate()` walks all `fingerprint_width` slots and calls `self.tuple_slot(bus, slot)` for each, which re-does `domains.get(bus)` plus three range comparisons. Only `payload_width` slots are trace-dependent; the `DomainBit(true)` slots contribute a fixed `sum of weight[slot]` and the `Zero` slots contribute nothing. Both are constant for the whole proof.
3. The symbolic expression tree is walked recursively per row, with no compiled form.

On top of that, this loop and `composition.rs` are entirely serial. `p3-bus` has no `parallel` feature at all and `multi-stark/Cargo.toml` adds it as a plain dependency, outside the `parallel = [...]` list -- so the bus is the one single-threaded stage in an otherwise rayon-parallel prover.

Rough budget at 2^20 rows and one declaration: ~21 ms for `materialize`, plus roughly `2 * degree` times that again across `initialize_claims` and `evaluate_air`, all on one core.

**Fix:** hoist a per-`(bus, weights)` plan computed once:

```rust
struct FactorPlan<EF> { live: Vec<(usize /*slot*/, usize /*payload idx*/)>, domain_constant: EF }
```

then per row do `offset - (domain_constant + sum live weight[slot] * payload[idx])` into a reused `payload` buffer. Parallelize `materialize` over `plan.blocks(direction)` (blocks are independent and their leaf offsets are known) and `evaluate_air` over its row range.

**Verify:** add a criterion bench for `materialize` at 2^20 rows and compare before/after; the two numbers above give you the ceiling to aim at.

multi-stark/src/bus/composition.rs
```diff
@@ -0,0 +1,359 @@
...
+                let row_point = &output.product.point[share.prefix_variables..];
+                let block_weight = equality_at_vertex(
+                    &output.product.point[..share.prefix_variables],
+                    share.prefix_index,
+                );
+                let coefficient = direction_weight * block_weight;
+                let state = airs[air].get_or_insert_with(|| {
+                    let main = tables[air]
+                        .iter_polys()
+                        .map(|column| Poly::new(column.iter().copied().map(Into::into).collect()))
```
tcoratger: **[MINOR] Round 0 lifts the entire trace into the extension field**

`Poly::new(column.iter().copied().map(Into::into).collect())` materializes every main and preprocessed column of every participating AIR as `Poly<EF>` before the first sumcheck round. For a base-field trace over a degree-4 extension that is a 4x memory blowup of the full trace, allocated up front and held for the whole reduction.

The main zerocheck deliberately avoids this: `VerifierAir`/`ProverAir` carry `MultilinearFolder<'a, F, F, F>` and `MultilinearFolder<'a, F, PackedExt<F, F>, PackedExt<F, F>>` bounds precisely so round 0 folds packed base-field rows and only later rounds work in `EF`. The bus composition does the naive thing instead.

Not a correctness issue, and it may well be acceptable for the first landing -- but it is worth a comment saying so, because the asymmetry with the zerocheck looks like an oversight rather than a decision.

**Fix:** keep round 0 in `F` (base-field values with an `EF` equality weight), switch to `EF` on the first `fold`. Or record the tradeoff in the module doc.

**Verify:** peak RSS of `prove` on a 2^20-row two-AIR bus statement, with and without the change.

multi-stark/src/bus/math.rs
```diff
@@ -0,0 +1,44 @@
...
+            } else {
+                challenge
+            }
+        })
+        .product()
+}
+
+/// Evaluation of `eq(left, right)` for two equally sized field points.
+pub(super) fn equality_evaluation<F: Field>(left: &[F], right: &[F]) -> F {
+    debug_assert_eq!(left.len(), right.len());
```
tcoratger: **[MINOR] `debug_assert_eq!` + `zip` silently truncates in release on a security-critical kernel**

If `left.len() != right.len()`, `zip` stops at the shorter one and returns a product over a prefix -- a well-formed but wrong equality weight, which feeds straight into the terminal identity the verifier accepts or rejects on. Release builds get no signal.

It is unreachable today: `terminal_composition` checks `point.num_variables() == max_num_variables()` and `output.product.point.len() == product_shape().log_height()`, and `prefix_variables + row_variables == log_logical_leaf_count` holds by construction in `terminal_shares`. But that is three separate invariants in two crates propping up a silent-truncation primitive.

**Fix:** return `Option<F>` and let the caller map `None` to `BusBindingError`, or use a plain `assert_eq!`. This is called `2 * num_shares` times per verification, not per row -- an unconditional check is free.

**Verify:** `cargo test -p p3-multi-stark --lib bus` stays green; then hand it mismatched slices in a unit test and assert it does not return a value.

multi-stark/src/bus/context.rs
```diff
@@ -0,0 +1,290 @@
...
+            let direction_weight = match direction {
+                BusDirection::Push => EF::ONE,
+                BusDirection::Pull => direction_challenge,
+            };
+            for share in self.plan.terminal_shares(direction) {
+                let air = share.owner.air;
+                let row_point = &point.as_slice()[point.num_variables() - share.row_variables..];
+                let unused = &point.as_slice()[..point.num_variables() - share.row_variables];
+                let fixed_all_one_selector = unused.iter().copied().product::<EF>();
+                let block_weight = equality_at_vertex(
```
tcoratger: **[MINOR] Re-implements `BusTerminalShare::prefix_weight`, dropping its bounds check**

`p3-bus` already exposes exactly this computation, checked:

```rust
pub fn prefix_weight<F: Field>(&self, point: &[F]) -> Option<F> {
    let prefix = point.get(..self.prefix_variables)?;
    equality_at_msb_vertex(prefix, self.prefix_index)   // None if prefix_index >= 2^len
}
```

and `bus/src/plan/mod.rs` has a test pinning that a malformed share returns `None` "without shifting by an invalid amount". The local `equality_at_vertex` here has neither guard: an out-of-range `vertex` silently ignores its high bits rather than failing.

Both the slice `output.product.point[..share.prefix_variables]` and `prefix_index = offset >> log_height` are plan-derived so neither can go out of range today. But the checked helper exists, is public, and is tested -- and using it would also delete one of the four copies of the equality kernel noted in the review body.

**Fix:** `let block_weight = share.prefix_weight(&output.product.point).ok_or(BusBindingError::ProductPointDimension { .. })?;` and drop `equality_at_vertex` from `multi-stark/src/bus/math.rs`.

**Verify:** `cargo test -p p3-multi-stark --lib bus` and `cargo test -p p3-bus --lib`.

tcoratger (commented):
I traced the authentication end to end: plan -> domain separator -> fingerprint/offset -> ProductGKR (shared root) -> direction batching -> composition sumcheck -> prescribed-point openings, and I could not break it. The binding is real: the padded-tuple fingerprint, the identity slots, the block geometry and the physical leaf order are all absorbed into the challenger before `fingerprint`/`offset` are drawn (`ds.seed` does reach the sponge), the direction challenge is drawn strictly after ProductGKR has fixed both terminal values, and `L(q)-1 = sum_b eq(prefix_b, q) * (L_b - 1)` is rebuilt only from PCS-authenticated column evaluations. Two design choices deserve explicit credit: `RecordToken` makes it *impossible* to record a bus declaration without the `bool_check` constraint being emitted into the same AIR pass, and `ProductGkrRootShape::FirstTwoShared` makes unequal push/pull roots unrepresentable on the wire rather than merely checked. Those are the right kind of fix.

No BLOCKER. What I do have is one verifier panic path, three structural/API issues, and a measured prover hot-path regression.

On the soundness question the description raises, here is the argument I convinced myself of, written out because it is load-bearing and not in the docs. Both `r` (fingerprint point) and `X` (offset) are drawn after the commitment, so each leaf is a genuine degree-1 polynomial `s_i * (X - fp_i(r)) + (1 - s_i)` with `s_i` fixed by the trace. Unique factorization of `P(r,X) = Q(r,X)` over the rational function field pairs push factors with pull factors up to scalars. Matching leading coefficients forces `prod s_i = prod s'_j`, and matching roots forces `fp_i(r) - fp'_j(r) = 1/s_i - 1/s'_j`, a constant in `r`. Since `sum_slot eq_slot(r) = 1`, a constant difference means the two padded tuples differ by the *same* constant in every slot. The identity-bit slots and the zero-padding slots are trace-independent, so that constant is pinned to zero and the tuples are equal. The whole argument collapses if `s_i` is not Boolean, which is exactly why `push_bus_interaction` emitting `assert_zero(selector.bool_check())` through a token the recorder cannot forge is the crux of this PR, not an incidental convenience. It is worth putting that paragraph in `bus/src/builder.rs` above `RecordToken`.

- `bus/src/leaf/mod.rs:175-192` -- `equality_weights` here binds **LSB-first** ("coordinate k binds bit k of the tuple-slot index"), while `bus/src/multilinear.rs::equality_weights_msb`, which `BusChallenges::fingerprint_weights` and therefore the whole authenticated path uses, binds **MSB-first**. Both are reachable from the crate's public surface with the same `&[EF]` point type. See the anchored comment on `bus/src/lib.rs:23`.
- The repository now carries **four** copies of the equality-weight kernel in **two** opposite conventions: `bus/src/multilinear.rs` (MSB), `multi-stark/src/bus/math.rs` (MSB), `bus/src/leaf/mod.rs` (LSB), `bus/src/product/math.rs` (LSB). Each is correct in isolation and each is security-critical. One `p3-multilinear-util` home with the convention in the type or the name would remove a whole class of future silent-wrong-answer bugs.
- `bus/src/plan/mod.rs:309` -- `debug_assert_eq!(offset % len, 0)`. Block alignment is what makes `prefix_index = offset >> log_height` meaningful, and it holds only because `pending.sort_by_key` puts `Reverse(log_height)` second. That derivation is not written down anywhere and is not checked in release. Either state it as a comment on the sort key ("descending height is what keeps every offset a multiple of its own length") or promote it to a real check; it costs one branch per block.
- `challenger/src/fs/state/verifier.rs:57-77` -- confirms the panic mechanism behind the `verifier.rs:305` finding: `Drop` for `VerifierState` panics with "Dropped unfinalized VerifierState" whenever the pattern was not replayed to the end.
- `examples/tests/transcript_separation.rs:781` only adds `has_bus: false` to an existing literal. No case in that suite toggles `has_bus`, so the one cross-protocol separation harness in the repo does not exercise the new bracket at all.

## Test coverage

The happy paths are well covered: balanced conditional bus, nonlinear payload, mixed heights, preprocessed-sourced payload, characteristic two. `conditional_nonlinear_bus_is_bound_to_committed_openings` is a genuinely good mutation test -- it tampers with the GKR root, `claimed_sum`, a round polynomial, the declared direction, the declared activation, the declared height, and a prescribed opening answer, and requires rejection for each. That is more than most PRs in this area ship.

What is missing is the *separation* half, and it is the half the PR description leans on ("Domain names and identities prevent two named buses from sharing one tuple space"):

1. **Cross-bus tuple collision.** No test proves that AIR X pushing tuple `(v)` on bus `"a"` and AIR Y pulling the identical tuple `(v)` on bus `"b"` fails to balance. This is the single property the domain-identity slots exist for, and nothing exercises `BusTupleSlot::DomainBit` end to end. If `tuple_slot` ever regressed to emit the same bits for every bus, every bus in a multi-bus statement would silently merge into one and the whole suite would stay green.
2. **Domain-separator mutation.** `bus/src/argument/transcript.rs` has zero tests. `multi-stark/src/transcript.rs` already has the right pattern (`seeds_agree` over ~15 single-field mutations of `MultiStarkShape`); the bus plan's separator deserves the same, one mutation per `instance()` call: `domain.name`, `domain.identity`, `domain.payload_width`, `payload_slots`, `domain_slots`, `fingerprint_width`, `blocks.len()`, and per block `bus`/`owner.air`/`owner.declaration`/`log_height`/`offset`. A `proptest` is cheap here and stronger than a fixed list: generate two `BusPlan`s from random `BusPlanInput` vectors and assert `plan_a != plan_b ==> seed(plan_a) != seed(plan_b)`, since `BusPlan: PartialEq` already gives you the antecedent for free.
3. **Cross-statement replay.** No test takes a valid proof for statement A and feeds it to a verifier holding statement B whose bus differs only in a bus *name*. `verify_with(&proof, &push, &wrong_direction, 2)` covers direction and `..., 3)` covers height, but not the name, which is the only field whose binding lives purely in the separator bytes.
4. **Non-Boolean selector.** No test confirms that a trace whose activation column holds a non-Boolean value is rejected. It is rejected by the zerocheck rather than by the bus, which is precisely why it should be pinned: the bus's soundness proof (above) depends on a constraint owned by a different sub-protocol.

**Bottom line:** the authentication itself looks sound and the two typestate choices are the right ones. Fix the `?` at `verifier.rs:305` so the verifier rejects instead of panicking, decide what `BusLeaves` is now that `BusPlan` supersedes it, and add the cross-bus separation tests -- as it stands the suite proves that tampering is caught but never that two different buses stay apart.

multi-stark/src/verifier.rs
```diff
@@ -257,7 +282,64 @@ where
...
+                    });
+                    let (point, terminal) = match verified {
+                        Ok(verified) => verified,
+                        Err(error) => {
+                            composition.abort();
+                            return Err(VerificationError::BusSumcheck(error));
+                        }
+                    };
+                    composition.finish();
+                    if bus_proof.composition.claimed_sum != expected_claim {
```
tcoratger: **[MINOR] Check the claim before running the sumcheck, not after**

`expected_claim` is known at line 305, but the full composition sumcheck runs first and this comparison happens after `composition.finish()`. A proof with a wrong `claimed_sum` pays for `num_variables` rounds of verification before being rejected.

It is also redundant twice over in the happy case: `GenericDegreeProof::verify` seeds its own transcript with `self.claimed_sum` (`sumcheck/src/generic_degree/transcript.rs:294`), so a tampered value already diverges the sponge. Keeping the explicit check is right -- it is what makes the binding legible rather than accidental -- but it should run first.

**Fix:** move the comparison between line 305 and the `composition.sumcheck(...)` call. Note it must stay *inside* the transcript bracket and go through `composition.abort()` on the mismatch path, for the same reason as the line-305 finding.

**Verify:** `conditional_nonlinear_bus_is_bound_to_committed_openings` already mutates `claimed_sum` and expects `Err`; it should still pass, and now via the cheaper path.

multi-stark/src/bus/composition.rs
```diff
@@ -0,0 +1,359 @@
...
+            };
+            for share in context.plan().terminal_shares(direction) {
+                let air = share.owner.air;
+                let row_point = &output.product.point[share.prefix_variables..];
+                let block_weight = equality_at_vertex(
+                    &output.product.point[..share.prefix_variables],
+                    share.prefix_index,
+                );
+                let coefficient = direction_weight * block_weight;
+                let state = airs[air].get_or_insert_with(|| {
```
tcoratger, line 201: **[MINOR] `get_or_insert_with` silently keeps the first share's geometry for the whole AIR**

`row_point`, `share.row_variables` (via `selectors` and `unused_prefix`) and the `equality` polynomial are captured from whichever share happens to reach this AIR first, then reused for every later term of the same AIR.

That is correct only because `BusPlan::build` sets `log_height: input.log_height` for every block an AIR emits, so all of an AIR's shares have the same `row_variables` and hence the same `prefix_variables` and the same `row_point`. Nothing at this call site says so, and nothing would catch it if per-declaration heights ever became a thing (which is the natural next feature for a mixed-height bus).

**Fix:** one line inside the `else` of the insert, or right after:

```rust
debug_assert_eq!(state.unused_prefix, num_variables - share.row_variables,
    "every block of one AIR shares its trace height");
```

plus a sentence on `PendingBlock::log_height` in `bus/src/plan/mod.rs` recording that it is per-AIR, not per-declaration.

**Verify:** `cargo test -p p3-multi-stark --lib bus` in debug.

multi-stark/src/bus/context.rs
```diff
@@ -0,0 +1,290 @@
...
+                                preprocessed: prep,
+                                public: public_values[air],
+                                is_first_row: EF::from_bool(row == 0),
+                                is_last_row: EF::from_bool(row + 1 == height),
+                                is_transition: EF::from_bool(row + 1 < height),
+                            },
+                            &weights,
+                            challenges.offset,
+                        )
+                        .expect("a checked bus plan resolves against its committed table");
```
tcoratger: **[NIT] `.expect` in the prover's innermost loop turns a caller mistake into a panic a million rows deep**

`evaluate_factor` fails with `MainColumn { column }` when the symbolic declaration indexes past the table's width. If a caller ever hands `prove` a `Table` narrower than `air.width()`, this panics from inside the row loop rather than returning `ProvingError::BusBinding`, and the message ("a checked bus plan resolves against its committed table") points at the plan rather than at the actual cause.

Same pattern at `composition.rs:110` and `composition.rs:243`.

**Fix:** either validate table widths against `air.width()` once in `BusContext::build` (cheap, once per AIR, and gives a good error), or hoist the first row's evaluation out of the loop and `?` on it so the remaining rows can keep the `expect` with a justified comment.

multi-stark/CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Changed
+
+- **Breaking:** Multi-STARK proofs carry an optional binary-bus section.
```
tcoratger, line 12: **[NIT] The recorded break is the wire format; the source break is missing**

The entry covers the postcard shape, which is the break the fixture regeneration is about. But this PR also breaks the public Rust API:

- `MultiStarkShape::new` gains a `has_bus: bool` parameter (5 args -> 6).
- `MultiStarkShape` gains a public `has_bus` field, so every struct literal breaks -- `examples/tests/transcript_separation.rs:781` in this very diff is one.
- `p3-multi-stark` now re-exports `BusBindingError` and `VerificationError`/`ProvingError` gain four new variants each, so exhaustive `match` on them breaks.

Anyone pinning `p3-multi-stark` hits these at compile time with no changelog entry to explain them.

**Fix:** add under `### Changed`:

```text
- **Breaking:** `MultiStarkShape::new` takes a `has_bus` flag and `MultiStarkShape` gains a `has_bus` field.
- **Breaking:** `VerificationError` and `ProvingError` gain binary-bus variants.
```

multi-stark/src/verifier.rs
tcoratger: Reproduced before fixing. A unit test handing the closure a reduction with `values.len() == 1` fails on the old code with:

```
thread '...' panicked at challenger/src/fs/state/verifier.rs:71:13:
Dropped unfinalized VerifierState: 2 pattern step(s) were never replayed and 0 NARG byte(s) were never checked.
```

so the failure mode was exactly the one you describe.

The bus block is now an extracted function, `bind_bus_composition`, which releases the driver on every return. The two tests that pin it are `a_noncanonical_terminal_value_count_rejects_without_unwinding` and `a_tampered_initial_claim_rejects_before_the_sumcheck_runs`; both fail with the drop-time panic if the `match` is reverted to `?`.

The same shape was live on the prover side: `prover.rs` drew the direction challenge and then used `?` on `batched_terminal_claim`, which would have dropped an unfinalized `ProverState`. `BusCompositionProverTranscript` had no `abort` at all, so there was nothing to call; it has one now and the prover goes through it. That path is unreachable for the same reason as yours, but it was written the same way.

bus/src/argument/transcript.rs
tcoratger: Four tests added in this module.

`the_instance_label_spells_out_every_public_dimension` builds the expected byte string by hand from the encoding rules, for a plan with two named buses of unequal payload width over two tables of unequal height, and compares it to the produced label. It does not call the encoder to derive its own expectation, so it pins the encoding rather than agreeing with it.

`one_step_from_a_plan_never_shares_its_seed` is the mutation test you asked for: rename, payload resize, declaration reorder, split across two names, height change, and split across two tables each move one dimension and each must change the seed.

`every_distinct_plan_in_a_sweep_lands_on_its_own_seed` enumerates 72 plans over two names, three widths, three heights and one-or-two tables, and asserts equal plans share a label while distinct plans never do. This is the injectivity property; I used a deterministic sweep rather than a proptest because `p3-bus` has no proptest dev-dependency and the reachable space here is small enough to enumerate.

`a_product_proof_for_one_named_bus_is_rejected_under_another` and `a_product_proof_for_one_table_layout_is_rejected_under_another` take a proof produced under one plan and verify it under a geometrically identical plan differing only in the bus name, or only in which table owns the pull.

On your verification criterion: I deleted each of the fourteen `.instance(...)` calls in turn and reran. All fourteen mutants are killed, including `domain.identity`, which no input-derived plan can distinguish on its own.

One caveat worth recording. The two cross-plan rejection tests need eight leaves per side. At two leaves the product reduction degenerates to a challenge-free root check and verifies under any transcript, so a depth-one version of that test passes for the wrong reason.

bus/src/lib.rs
tcoratger: Deleted, not sealed. `bus/src/leaf/` is gone and `BusLeaves`, `BusLeafDeclaration`, `BusSelector` and `BusLeafError` are off the public surface.

I checked your reading of the bit order and it holds: the deleted helper split each new coordinate into a high half at `old_len + index`, so coordinate zero addressed bit zero, while the planned path interleaves and lands coordinate zero on the most significant bit. Two public routes to a leaf factor that disagree for any tuple width of four or more is not something worth keeping for one test.

`plan/mod.rs:861` now materializes through `compile_factor`/`evaluate`. The leaf values are unchanged, because that test uses a one-coordinate fingerprint point and a plan whose single domain identity bit lands in the slot the old test filled with a column of ones.

`grep` finds no other in-tree user, and `cargo check --workspace --all-targets` plus the repo lint recipe are clean. There is no `bus/CHANGELOG.md` on `main`, so the removal has no changelog entry; say the word if you want the file created.

multi-stark/src/bus/context.rs
tcoratger: Measured first, and my numbers disagree with the attribution, though not with the conclusion.

Same shape as yours: BabyBear, degree-four extension, three-slot payload, `fingerprint_width = 4`, 2^20 rows, release.

```
per-row call, as on this branch              47.3 ns/row
+ hoisted placement, values still in EF      38.7 ns/row
+ hoisted placement, Boolean rows in F       17.7 ns/row
hand-written inner product, no expression    12.4 ns/row
```

So the two loop-invariant costs you name are real but account for about 8 ns of 47. The allocation is cheaper than it looks because it is a hot three-element `Vec` on the glibc fastbin path, and the slot walk is four comparisons. The other 21 ns is arithmetic: the row is lifted to the extension field before the fingerprint is summed, so the tuple terms are four `EF x EF` multiplications, each about sixteen base multiplications, where `EF x F` would be four.

I could not reproduce 0.6 ns/row for an equivalent inner product. Three extension-field multiplications and an addition cannot retire in 0.6 ns on this core; my closest hand-written loop, with the payload read straight from base-field columns, is 12.4 ns/row. I suspect the loop under measurement was hoisted.

What landed:

- `BusPlan::compile_factor` settles slot placement, the fixed named-domain contribution and all three width checks once, returning a reusable `BusFactorPlan`. Because the payload index and the tuple slot coincide over the leading payload slots, the per-row path needs no scratch vector at all: it zips the payload expressions against their weights directly. `evaluate_factor` is now one call to it, for the single-point verifier use.
- `BusFactorPlan::evaluate` is generic over the value algebra, so the prover resolves Boolean rows in the base field and pays `EF x F`, while the composition rounds keep `EF x EF`.
- `materialize` builds base-field row buffers instead of extension-field ones, and `initialize_claims`/`evaluate_air` compile their placements once per term rather than once per row.
- The row loops in `materialize_block` and `evaluate_air` go through the rayon prelude via `into_par_iter().map_init(...)`, so the bus is no longer the single-threaded stage. I parallelized over rows rather than blocks because a statement is frequently one tall block. `p3-bus` still has no `parallel` feature and does not need one now that the loops live in `p3-multi-stark`, which already has it.

End to end on the leaf loop: 49.6 ms -> 18.6 ms per 2^20 rows, 2.7x, before threads. Criterion bench added as `bus/benches/factor.rs` with both arms.

Not done: the symbolic tree is still walked per row. The remaining 5 ns/row between 17.7 and 12.4 is that walk, and closing it needs a compiled expression form, which is a larger change than this review.

multi-stark/src/bus/composition.rs
tcoratger: Taking the second option: the tradeoff is recorded in the module doc rather than fixed.

I measured the lift in isolation rather than as a prove-level A/B, because with the change unimplemented there is no second arm to compare against. A 2^20-row eight-column BabyBear trace against the same trace lifted to a degree-four extension, peak RSS from `/proc/self/status`:

```
base trace                34.3 MiB
after the extension lift  162.4 MiB   (+128.0 MiB)
```

Exactly 4x, as the type sizes predict, and held for the whole reduction. Note the prover materializes every column of every participating AIR, not only the ones the declarations read, so the multiplier applies to the full trace width.

Keeping round zero in the base field here needs the same folder-generic plumbing the zerocheck carries, which is a larger change than the rest of this review. The module doc now says the asymmetry is a decision and what it costs.

multi-stark/src/bus/math.rs
tcoratger: Now returns no value on a length mismatch, and the caller maps that to `BusBindingError::ProductPointDimension` carrying the dimension the share expects. No debug assertion is left in the kernel.

Three tests in `bus::math::tests`: mismatched slices in both directions and against an empty point all return nothing; equal points agree with the product written out by hand; and the weight table is pinned against its four hand-written entries. The sibling helper `equality_at_vertex` is gone entirely, see the neighbouring thread.

multi-stark/src/bus/context.rs
tcoratger: Done as suggested. The verifier now calls the checked helper and maps its absent result to `BusBindingError::ProductPointDimension`, and `equality_at_vertex` is deleted from `multi-stark/src/bus/math.rs`.

The prover side in `composition.rs` was the second copy of the same call and now uses the helper too. There the input is plan-derived with no proof data in it, so the absent case is an `expect` rather than an error return.

That removes one of the four copies of the kernel you counted; the remaining ones are the weight table and the two-point evaluation, which compute different things.

multi-stark/src/verifier.rs
tcoratger: Moved. The comparison now runs immediately after the direction challenge and before the sumcheck, inside the transcript bracket, and goes through `abort` on the mismatch path for the same reason as the neighbouring finding.

`conditional_nonlinear_bus_is_bound_to_committed_openings` still passes, now via the cheap path. `a_tampered_initial_claim_rejects_before_the_sumcheck_runs` pins the ordering directly: it hands the function a proof with no round polynomials at all, so if the sumcheck ran first the error would be a shape rejection rather than the initial-claim mismatch.

multi-stark/src/bus/composition.rs
tcoratger: Added, verbatim, immediately after the insert so it runs for the reused branch as well as the fresh one. Both height fields also now say in one line that the value comes from the owning table rather than the declaration, on the private pending record and on the public block it becomes.

Debug tests are green.

multi-stark/src/bus/context.rs
tcoratger: Took your first option, with one change.

`BusContext::check_tables` runs once before the prover materializes anything, and the prover uses `?` on it, so a narrow table returns `ProvingError::BusBinding` instead of panicking a million rows deep. Rather than compare widths, it does one dry evaluation per declaration against zero rows of the committed widths. That is the same code path the row loop takes, so it cannot pass while the loop would fail, and the error is the precise `MainColumn { column }` you quoted rather than a width summary. It also catches missing preprocessed columns and out-of-range public values.

`rejects_a_table_narrower_than_the_air_that_owns_it` covers it: a two-column AIR reading its second column, handed a one-column table, gets `BusBindingError::Evaluation(BusEvaluationError::MainColumn { column: 1 })`, and the declared width passes.

The two `expect`s in `composition.rs` are covered by the same guard, since the composition prover is built after it runs, and the one in `materialize_block` now carries a comment saying so. No new error variants: the enum already had a transparent evaluation arm.

multi-stark/CHANGELOG.md
tcoratger: Both lines added under `### Changed`, with the re-export folded into the second one since it breaks the same consumers.

multi-stark/src/bus/context.rs
```diff
@@ -0,0 +1,390 @@
...
+    ) -> Vec<EF> {
+        let air = block.owner.air;
+        let interaction = &self.profiles[air].interactions()[block.owner.declaration];
+
+        // Slot placement and the named-domain contribution are settled once for the block.
+        let factor = self
+            .plan
+            .compile_factor(block.bus, interaction, weights, offset)
+            .expect("a checked bus plan compiles against its own declarations");
+        let main_columns = tables[air].iter_polys().collect::<Vec<_>>();
```
Nashtare: `iter_polys()` goes through `dense_matrix()`, which panics on packed Boolean storage (`sumcheck/src/layout/witness.rs:449`). `BooleanTracePcs` keeps the caller's tables as its committed tables, and `prove_hash_binary` builds them with `Table::from_packed_bits`. So any bus AIR run through the Boolean harness panics here, and this PR now makes `BinaryAir` implement `Air<BusSymbolicBuilder>`.

I checked with a probe. It uses a three-column AIR (push `a`, pull `b = reverse(a)`, selector `s = 1`) at height 2^10. With a dense table it proves and verifies. With the same bits packed it panics with "dense table access is unavailable for packed Boolean storage". The backtrace runs `materialize_block` <- `BusPlan::prove` <- `bus_argument`, so the panic comes after the fingerprint challenges are drawn. `check_tables` only calls `num_polys()`, so it misses this, and `composition.rs:168` and `173` make the same call. The verifier is unaffected because it only reads opened values.

Could this read through `column()`/`columns()` instead? Or it could reject a packed table up front, with a test, the way the indexed-lookup path does at `prover.rs:294`.

multi-stark/src/verifier.rs
```diff
@@ -412,6 +533,42 @@ where
...
+            .collect::<Vec<_>>();
+        let expected = context.terminal_composition(
+            output,
+            *direction,
+            point,
+            &bus_main,
+            &bus_preprocessed,
+            &public_values,
+        )?;
+        if expected != *terminal {
```
Nashtare, line 586: This comparison is what authenticates the ProductGKR terminal values, but no test reaches it. I replaced it with `if false && expected != *terminal` and ran `cargo test --release -p p3-multi-stark --lib bus`: 16 passed, 0 failed. Every tamper in `conditional_nonlinear_bus_is_bound_to_committed_openings` is caught earlier, by GKR layer consistency, `InitialClaimMismatch`, the PCS opening, or a changed plan. Nothing asserts `TerminalMismatch`. No test has the verifier reject an unbalanced committed witness either.

`prove_forged` already has a `#[cfg(test)]` `Forgery` hook for the indexed reduction. Could it get a bus arm that commits an unbalanced table but materializes and composes from a balanced substitute, with an assert on `BusBinding(TerminalMismatch)`? It would also help to narrow the `is_err()` asserts in that test to the variant each tamper should produce.

multi-stark/src/bus/composition.rs
```diff
@@ -0,0 +1,403 @@
...
+                        term.bus,
+                        context.interaction(term.owner),
+                        fingerprint_weights,
+                        offset,
+                    )
+                    .expect("a checked bus plan compiles against its own declarations")
+            })
+            .collect::<Vec<_>>();
+
+        for row in 0..height {
```
Nashtare: `row_claim` is only read while `round < unused_prefix` (line 328), but `initialize_claims` fills it for every AIR. It does so with a serial walk over every row in extension-field arithmetic. For an AIR at the maximum height the whole pass is thrown away, and in a single-height statement that is every AIR. For shorter AIRs it's the one serial loop left in the reduction, now that `materialize_block` and `evaluate_air` go through rayon. Could it skip AIRs with `unused_prefix == 0` and run the rest in parallel?

multi-stark/src/instance.rs
```diff
@@ -623,6 +638,22 @@ where
             }
         }
 
+        if let Some(context) = bus {
+            for (air, table) in tables.iter_mut().enumerate() {
+                if context.contains_air(air) {
+                    let role = BatchRole::Bus { air };
+                    let width = self.0[air].air.width();
+                    table.1.push((
+                        OpeningBatch::new((0..width).collect(), Vec::new()),
```
Nashtare: The bus batch opens every main column of a participating AIR, and every preprocessed column too (line 729), however many the declarations actually read. The composition prover also lifts and folds all of them each round (`composition.rs:167-175`, `307-315`). So a width-`W` AIR with a one-column payload pays `W` extra extension-field values in the proof, plus `W` column folds per round. The indexed-lookup batches just above open only the columns they need. Could the context collect the main and preprocessed columns each AIR's declarations reference, then open and fold only those?

multi-stark/src/prover.rs
```diff
@@ -675,6 +764,136 @@ mod tests {
...
+        fn eval(&self, builder: &mut AB) {
+            let main = builder.main().current_slice()[0];
+
+            // Both sides expose the same one-coordinate tuple through different commitments.
+            let value: AB::Expr = if self.fixed.is_some() {
+                // The otherwise unused main column keeps an ordinary zerocheck constraint.
+                builder.assert_zero(main);
+                builder.preprocessed().current_slice()[0].into()
+            } else {
+                // Pin one committed cell so this AIR also contributes an ordinary constraint.
```
Nashtare: This dummy constraint, and the one at line 837, is needed because `get_air_profile` asserts that every AIR contributes constraints or lookup interactions (`zerocheck/mod.rs:280`), and bus declarations don't count toward that. So an AIR whose only content is an `Always` declaration panics in `setup`, and `assess_statement` rejects it with "AIR declares no constraints or lookup tuples". Yet a fixed table that pushes every row is the typical bus provider. Could the zerocheck accept an AIR that only declares bus interactions? If not, it could at least return a typed error, with a note on `push_bus_interaction`.

multi-stark/src/verifier.rs
```diff
@@ -257,7 +341,42 @@ where
...
+                Err(error) => {
+                    transcript.abort();
+                    return Err(error);
+                }
+            }
+        }
+        (None, None) => None,
+        _ => {
+            transcript.abort();
+            return Err(VerificationError::UnexpectedBus);
```
Nashtare, line 372: No test reaches this arm in either direction: a proof with a bus section for AIRs that declare none, or a bus AIR whose proof lacks one. Main's #2278 added these presence tests for the other three optional sections. Could the bus section get the same pair?

bus/src/evaluation/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub is_transition: EF,
+}
+
+impl<F, EF> BusEvaluation<'_, F, EF>
+where
+    F: Field,
+    EF: ExtensionField<F>,
+{
+    /// Resolve one symbolic expression against commitment-bound evaluations.
+    fn evaluate(&self, expression: &SymbolicExpression<F>) -> Result<EF, BusEvaluationError> {
```
Nashtare: This walks the expression as a tree, so a shared subexpression is evaluated once per path. A probe on this head timed one row of a doubling-chain payload at 0.30 ms for depth 16, 4.2 ms for depth 20 and 56 ms for depth 24. The prover pays that per row in `materialize_block`, in `initialize_claims` and in each round of `evaluate_air`. The verifier pays it once per share at the terminal point, and `validate_expression` in `BusPlan::build` walks the same way. AIRs choose their own expressions, so this is about robustness, not soundness. Would it be worth compiling each declaration once into a node-indexed form, so shared nodes are evaluated once?

multi-stark/src/verifier.rs
```diff
@@ -63,6 +66,21 @@ where
     /// The proof and the AIRs disagree on whether an indexed reduction exists.
     #[error("indexed reduction present but not expected, or absent but described")]
     UnexpectedIndexedReduction,
+    /// Binary-native bus declarations do not define a supported statement.
+    #[error("binary-bus planning failed: {0}")]
+    BusPlan(#[from] p3_bus::BusPlanError),
```
Nashtare: `BusContext::build` wraps plan errors as `BusBindingError::Plan`, so neither this variant nor `ProvingError::BusPlan` is ever constructed. Could they go, or could plan errors be routed here instead?

bus/src/builder.rs
```diff
@@ -106,6 +109,25 @@ pub struct SymbolicBusInteraction<F: Field> {
     pub activation: BusActivation<SymbolicExpression<F>>,
```
Nashtare, line 120: These fields are public, and `BusPlan::build` and `compile_factor` take a hand-built interaction, which has no Booleanity constraint behind it. `RecordToken` only covers the builder path. Could this field, or `BusActivation::Boolean`, say that soundness needs the selector constrained to `{0, 1}` elsewhere?

multi-stark/src/bus/composition.rs
```diff
@@ -0,0 +1,403 @@
...
+//!
+//! The terminal expression is checked only after its source columns open from the PCS.
+//!
+//! Round zero lifts every source column into the challenge field before any folding.
+//!
+//! A degree-four extension therefore holds four times the trace for the whole reduction.
+//!
+//! The batched zerocheck avoids that by folding packed base-field rows in its first round.
+//!
+//! Matching it here needs the same folder-generic plumbing and is deferred.
```
Nashtare: Nit: "and is deferred" describes a plan rather than the code; the cost sentences above stand on their own.

Nashtare (changes requested):
I traced the bus end to end on both sides and couldn't break soundness. Every value the verifier uses is absorbed before the challenge that depends on it. The plan separator binds names, widths, identities and every block's owner, height and offset, and it seeds `r` and `X` after the main commitment and public values. ProductGKR binds roots, round polynomials and children before each challenge. The direction challenge comes after both terminal values are fixed, and the initial claim is checked before the composition sumcheck runs. Direction is structural. The selector's Booleanity constraint goes through the same builder the zerocheck folders use. The identity slots rule out cross-bus cancellation, and the terminal identity forces the padding to one. I recomputed the security numerators (fingerprint `max(1, s) * N`, `5R`, `L(T - 1)`, `h`, direction 1, composition `n * D`). They match, they're charged over the challenge field, and they get the list-size penalty. The no-bus transcript is unchanged: the fixture differs by exactly one inserted `0x00` byte, the `None` tag, and the other 7169 bytes are identical.

- The bus prover panics on packed Boolean tables. In a probe with a three-column permutation AIR under `BooleanTracePcs`, a dense table proves and verifies. The same bits packed panic at `witness.rs:449` inside `materialize_block` (inline, `context.rs`).
- Nothing tests the terminal composition check. With it replaced by `if false && ...`, all 16 bus tests still pass (inline, `verifier.rs`).

**Smaller things**
- The conflict with main is one import line in `examples/src/binary.rs` (`p3_blake3::Blake3` against `p3_bus::BusSymbolicBuilder`). Keeping both resolves it.
- #2224 builds its memory fingerprint on `crate::leaf::equality_weights`, which this PR deletes, and that helper orders slots least-significant first where the plan orders them most-significant first. Its module doc also says a memory bus has to stay out of the plan's two-tree reduction, but here the plan takes every declaration an AIR records. Rebasing it needs a design decision, not just a conflict fix.
- No end-to-end multi-stark test uses two named buses of different payload widths, so the identity and zero-padding slots are only unit-tested in `p3-bus`.

tcoratger:
Flagging a defect found while re-checking the evaluator-duplication question on #2223. It is on this branch, on the prover entry path, and this PR currently has no open threads — so it would merge with the problem live.

**`BusEvaluation::evaluate` walks the expression once per path, not once per node.** `bus/src/evaluation/mod.rs:81-84` is plain recursion:

```rust
SymbolicExpr::Add { x, y, .. } => Ok(self.evaluate(x)? + self.evaluate(y)?),
```

Arithmetic nodes share their operands, so a symbolic expression is a DAG. `double()` is `self.dup() + self.dup()`, which means a 32-bit recomposition is about 64 nodes but 2^32 paths.

Measured on this head, release, BabyBear with a degree-4 extension, one evaluation point:

| depth | 16 | 18 | 20 | 22 | 24 |
|---|---|---|---|---|---|
| ms/call | 0.240 | 0.948 | 3.841 | 13.80 | 55.42 |

Least-squares exponent 0.978, i.e. 2.00x per level. That extrapolates to roughly 13 s at depth 32 and 50-60 minutes at depth 40, per evaluation point.

`check_tables` runs this on every declaration at `multi-stark/src/prover.rs:391`, before the row loop, so a machine declaring a word recomposition on a bus hangs the prover with no diagnostic.

**Why the benchmark does not show it.** `bus/benches/factor.rs` uses `WIDTH = 3` bare leaf variables with `BusActivation::Always`, so three trivial nodes per row. It measures the per-row constant this PR removes, a million times, and is structurally incapable of reaching the cliff.

**The fix already exists in this crate.** #2223 hit the same shape twice and closed it both times with a pointer-keyed visited set — see `bus/src/plan.rs:571` on that branch, about seven lines. `compile_factor` is not affected: measured at 6.96-7.01 ns flat across depths 16 to 24, so the hoist never touched the walk.

Not a soundness issue. It is a liveness one, and it is reachable from an ordinary declaration.

bus/src/evaluation/mod.rs
tcoratger: Reproduced on the head you measured: one row of a doubling-chain payload takes 218.70 us at depth 16, 3.4936 ms at depth 20 and 55.651 ms at depth 24, which matches your 0.30 / 4.2 / 56 ms.

Compiled now, as you suggest. `compile_factor` flattens the declaration into a node-indexed graph in an order that resolves every operand before its user, and evaluation is one forward pass over a scratch buffer the caller reuses across rows. Arithmetic nodes are keyed by address in a `BTreeMap`, the same memoization `SymbolicExpr::degree_with` already uses. Leaves are copied at each use rather than recorded, which stays linear and leaves the map empty for an ordinary declaration.

The three depths now cost 40.4, 50.2 and 57.2 ns, so linear in depth rather than exponential.

Leaf support is now decided at compile time, so the unsupported-access errors come out of `compile_factor` instead of out of each row. `validate_expression` in `BusPlan::build` gained the pointer-keyed visited set from #2223; a test builds a depth-32 chain through `BusPlan::build` and a second one buries an unsupported leaf under the same depth, so the dedup cannot swallow the rejection.

Effect on the existing arms, same machine: `compiled` 17.543 -> 17.327 ms, `per-row` 45.368 -> 66.533 ms. The one-shot arm regresses because `evaluate_factor` now pays a compile it did not pay before, about 20 ns per call in allocations. It is called once per terminal share in the verifier and nowhere per row, and the arm exists to show that compiling per row is the wrong thing to do, so I left it.

The benchmark also gained the arm you would have needed to see this: `bus-shared-subexpression` times one row of a shared chain at depths 16, 20 and 24. The three bare leaf variables in the old arms measure the per-row constant and cannot reach the cliff, as you say.

multi-stark/src/bus/context.rs
tcoratger: Reproduced, and fixed by reading through `column()` rather than rejecting.

The reproduction took one extra step to build: the ordinary stacked PCS densifies its retained tables inside `Witness::new`, so `config.committed_table` hands the bus phase a dense table even when the caller supplies a packed one. That is why a plain end-to-end test with `Table::from_packed_bits` passes regardless. `BooleanTracePcs` does keep the caller tables, as you say, and the test config already has a `committed_table_override` hook that reproduces exactly that. With it, `a_bus_over_packed_boolean_tables_proves_and_verifies` panics on this head with "dense table access is unavailable for packed Boolean storage" at `witness.rs:450`, and it panics from both call sites: `materialize_block` first, and then `AirState` construction in the composition prover once that one is fixed.

Both now go through `columns()` and `ColumnView::value` / `ColumnView::values`, so a packed table is read bit by bit without expanding it. `check_tables` was already representation-independent.

The test is a two-AIR push/pull pair over a packed two-column table at the padding floor, proving and verifying end to end.

multi-stark/src/verifier.rs
tcoratger: Confirmed your measurement first: with `if false && expected != *terminal` the whole bus suite still passed. It does now too, except for the one new test.

`Forgery` gained a `bus_table` arm: an AIR index and the column values the bus phase materializes and composes in place of the committed table. `a_bus_terminal_claim_the_committed_trace_does_not_support_is_rejected` commits a balanced push table and an unbalanced pull table, and points the bus phase at a balanced substitute for the pull side. The product reduction closes, the composition sumcheck closes, the openings verify, and the proof is refused with `BusBinding(TerminalMismatch)`.

Verified the mutation: with `if false &&` reapplied, that test is the only one in the bus suite that fails.

The `is_err()` asserts in `conditional_nonlinear_bus_is_bound_to_committed_openings` now name a variant each:

    roots[0] tampered        -> BusArgument(Product(LayerConsistency { layer: 0 }))
    claimed_sum tampered     -> BusBinding(InitialClaimMismatch)
    round_polys[0][0]        -> Opening(_)          the challenge, and so the point, moves
    wrong direction          -> BusArgument(Product(LayerCountMismatch { 2, 1 }))
    wrong activation         -> Opening(_)          see below
    opening answer tampered  -> Opening(_)

The activation case used to reach `TerminalMismatch` and now does not, because of the column narrowing in the other thread: dropping the activation stops the declaration reading the selector column, so the rebuilt bus batch answers one column where the proof carries two, and the opening shape rejects first with `OpeningBatchSizeMismatch { table_idx: 1, expected: 1, actual: 2 }`. That is an earlier and sharper rejection, and the dedicated forgery test above is what holds the terminal comparison.

multi-stark/src/verifier.rs
tcoratger: Added, one per direction, following the shape #2278 used for the other three sections.

`verify_rejects_a_bus_section_for_airs_declaring_no_bus` proves a constraint-only AIR, grafts the section from a genuine bus proof onto it, and expects `UnexpectedBus`. `verify_rejects_bus_airs_whose_proof_carries_no_bus_section` takes the conditional bus fixture and drops the section.

Verified the mutation: replacing the arm with `_ => None` fails exactly those two tests and nothing else.

multi-stark/src/bus/composition.rs
tcoratger: Both done.

An AIR with `unused_prefix == 0` never reads its claim, so it no longer computes one. The remaining tables run the pass in parallel through `par_fold_reduce`, with the per-term partial sums combined per split.

Measured on a pair of 64-column AIRs over 4096 rows, which is the single-height case where the whole pass was thrown away: proving goes from 1.031 s to 0.975 s, about 5 percent of the whole prover. The rest of that figure is the column narrowing in the `instance.rs` thread; with the narrowing in place but the skip removed, proving is 1.024 s, so the skip alone is worth about 50 ms here.

multi-stark/src/instance.rs
tcoratger: Done. `SymbolicBusInteraction` gained a query for the current-row main and preprocessed columns a declaration reads, walked once with a pointer-keyed visited set. `BusContext` merges those per AIR at build time, and both schedules open only that set. A set that comes out empty schedules no batch at all, which is what happens to the main side of a provider whose payload lives in the verifying key.

The composition prover lifts and folds only those columns too. The evaluation view stays indexed by column, so the per-row buffer is declared-width with the read columns scattered into it and the rest left at zero, which no planned expression names. The verifier scatters the batch answers back the same way before the terminal comparison.

The security assessment walks the same two schedule builders, so it follows automatically.

Measured on a pair of 64-column AIRs over 4096 rows whose declarations read one column each:

    opened values   384    -> 258
    proof bytes     48930  -> 46627
    prove           1.031 s -> 0.975 s

The 126 values are the 63 unread columns on each side.

multi-stark/src/prover.rs
tcoratger: I tried the first option and it does not work, so the dummies stay, now with a comment saying why and a note on the declaration API.

The assertion is not the obstacle. I removed it, taught `setup` and `assess_statement` to count bus declarations, and stripped both dummies from the fixture. Proving then panics further in, at `rounds.rs:131`:

    assert!(degrees.iter().all(|degrees| degrees.max() > 0));

`Stage::new` requires every AIR in a zerocheck stage to have a positive-degree expression family, because an AIR with none contributes no round polynomial. Accepting a bus-only AIR means excluding it from the stage, and the stage, its claims, its links and its openings are all indexed by AIR on both sides. That is a change to the batched zerocheck, not to an assertion, and it belongs in its own PR rather than in this one.

So the second option, done properly. The panic message now says binary-bus declarations are reduced separately and do not count, `push_bus_interaction` says a provider still owes the batch a constraint of its own, and the two dummies in `PreprocessedBusAir` carry a comment saying that is why they are there. I reverted the speculative `setup` and `assess_statement` changes rather than leaving a half-relaxed check behind.

Worth its own issue, since a fixed table that pushes every row really is the typical provider.

multi-stark/src/verifier.rs
tcoratger: Removed both. `BusContext::build` already surfaces plan failures as `BusBindingError::Plan`, which both `ProvingError::BusBinding` and `VerificationError::BusBinding` carry, so routing them a second way would only give the same failure two spellings.

bus/src/builder.rs
tcoratger: Said, on both. The activation variant now records that soundness needs the selector to be zero or one on every row, that the public declaration path emits that constraint, and that a declaration assembled by hand carries none, so its author owes one.

multi-stark/src/bus/composition.rs
tcoratger: Dropped.

multi-stark/src/bus/context.rs
```diff
@@ -0,0 +1,446 @@
...
+        let air = block.owner.air;
+        let interaction = &self.profiles[air].interactions()[block.owner.declaration];
+
+        // Slot placement and the named-domain contribution are settled once for the block.
+        let factor = self
+            .plan
+            .compile_factor(block.bus, interaction, weights, offset)
+            .expect("a checked bus plan compiles against its own declarations");
+        // Representation-independent views read a packed Boolean table without expanding it.
+        let main_columns = tables[air].columns().collect::<Vec<_>>();
```
Nashtare: The opening and the composition lift are narrowed now, but this loop still reads every column of the table on every row, once per block. The compiled factor only reads `main_columns(air)` and `preprocessed_columns(air)`. So it's `W` `ColumnView::value` calls per row per declaration, and on a packed table each call is an assert plus a bit extract. For a 64-column pair that's small, but a wide Boolean hash AIR with a few declarations would decode its whole trace once per declaration. Could this reuse the `(indices, width)` scatter from `AirState`, so it only reads the referenced columns?

multi-stark/src/prover.rs
```diff
@@ -361,7 +380,102 @@ where
...
 
-    // 4. Materialize the lookup fractions and reduce them, inside the delegation bracket.
+    // Under test the bus phase may read a table other than the one this proof commits.
+    // A proof can then carry terminal claims the committed trace does not support.
+    #[cfg(test)]
+    let bus_substitute =
+        forgery
+            .and_then(|forgery| forgery.bus_table.as_ref())
+            .map(|&(air, ref values)| {
+                use p3_field::PrimeCharacteristicRing;
```
Nashtare: Nit: this `use` sits inside a closure body. A `#[cfg(test)]` import at the top of the module, or `<C::Val as p3_field::PrimeCharacteristicRing>::from_u64`, would keep imports at module level like the rest of the crate.

bus/src/builder.rs
```diff
@@ -70,6 +80,10 @@ pub trait BusInteractionBuilder: BusInteractionRecorder {
     ///
     /// A conditional activation is constrained to zero or one before it is recorded.
     ///
+    /// The declaration is reduced outside the batched zerocheck and leaves it nothing to fold.
```
Nashtare: Nit: this only holds for an `Always` declaration. A conditional one leaves the zerocheck its Booleanity constraint (line 101), so an AIR whose only content is a `Boolean` declaration is accepted. My probe's permutation AIR has no other constraint, and it proves and verifies. Could these two sentences say "an unconditional declaration"?

Nashtare (approved):
I went through the three new commits after the rebase. The seven rebased commits are unchanged apart from the `examples/src/binary.rs` import, which now keeps both lines. The branch merges cleanly into current main, and the merged tree type-checks for `p3-bus`, `p3-multi-stark` and `p3-examples` with all targets.

The packed-table fix holds. Every table read on the bus prover path now goes through `ColumnView`, and `check_tables` only needs `num_polys`. `ColumnView::value` decodes exactly as `into_dense` does, so packed and dense tables give the same cells. I reran last round's probe under `BooleanTracePcs`, a three-column permutation AIR with `s = 1`. Packed and dense tables both prove and verify at 2^6, 2^7, 2^8, 2^10 and 2^12 rows. With one bit of the pull column flipped, both are refused with `UnbalancedProducts`.

The terminal comparison is now pinned. In the new forgery test, product reduction, initial claim, composition sumcheck and openings all close on the substitute or on honest openings. The comparison is the only check that sees the committed pull table, so the test fails once it's disabled. The compiled node list is keyed by address inside a single `compile_factor` call, so it can't mix nodes across declarations or AIRs. The terminal evaluation, `validate_expression` and `referenced_columns` are all linear now. The column narrowing is consistent: the opened, lifted and scattered columns are exactly the leaves the compiled factor reads, and the security assessment walks the same schedules.

The skipped and parallel row claims, the presence tests, the removed variants, the Booleanity doc and the comment close the other threads. A bus-only provider still needs a dummy constraint, but that is documented now, and I agree the fix belongs with the batched zerocheck.

**Smaller things**
- There's still no end-to-end test with two named buses of different payload widths.
- #2224's new head still imports `crate::leaf::equality_weights`, which this PR deletes. That helper orders slots least-significant first, and `fingerprint_weights` here orders them most-significant first. So its rebase needs the switch, not just a moved helper.

The inline comments are non-blocking.

multi-stark/src/bus/context.rs
tcoratger: You are right, I narrowed two of the three paths and left this one. It now takes the same column indices and declared width the composition state uses, and scatters the read columns into a declared-width row buffer; unread columns keep their zero, which no planned expression names.

Measured on the same two 64-column AIRs over 4096 rows, this change alone, three runs each:

    dense    1.071 / 1.038 / 1.068 s  ->  0.990 / 0.976 / 0.990 s
    packed   1.021 / 1.127 / 1.053 s  ->  0.989 / 0.980 / 0.992 s

About seven percent of the whole prover on each. The packed fixture goes through the override hook so the committed tables stay packed, which is the case your comment is about. The dense and packed timings now sit on top of each other, which they did not before: the extra cost of decoding 63 unread bit columns per row was exactly what this removed.

These absolute numbers are a little above the 1.031 -> 0.975 s I quoted last round for the same fixture, because the branch has since been rebased onto main and the machine is not quiet. The A/B above was taken in one sitting, so the ratio is the part to trust.

bus/src/builder.rs
tcoratger: Correct, and my wording was wrong. Confirmed your probe: an AIR whose whole content is one conditional declaration proves and verifies, because the Booleanity check the declaration path emits is a real zerocheck constraint. `an_air_whose_only_content_is_a_conditional_declaration_is_accepted` is a push and pull pair of exactly that shape, with no other constraint on either side.

The three sentences now say that the Booleanity check is the only thing a declaration leaves the zerocheck, and that an AIR is refused when it declares nothing else and no conditional activation either.

This also narrows #2284, which says a bus-only AIR cannot be expressed. Only an unconditional one cannot. A provider that pushes every row still needs a constraint, or a selector it constrains anyway, so the gap is real but smaller than that issue states. I have not edited the issue.

multi-stark/src/prover.rs
tcoratger: Moved to a `#[cfg(test)]` import beside the other field imports at module level.

PR #2245, opened 2026-09-19
perf(binary): speed up the Boolean-committed prover

Nashtare:
Speeds up the Boolean-committed binary prover, prover side only: every proof in this branch is byte-identical to the one `main` emits, at every shape I measured.

The starting point was a profile of `prove_hash_binary --objective blake-3-compressions -l 20 --security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4` on a c8a.16xlarge (64-thread EPYC, `-C target-cpu=native`), where two thirds of `prove` sat outside any span. The first commit spans the Boolean wrapper's phases; the rest close what that showed, plus the zerocheck rounds underneath.

## What was slow, and what each commit does

**The Boolean trace wrapper (binary-pcs).**
- Opening values were one serial walk per column, adding one row weight per set bit. `packed_column_sums` tabulates each 64-row block's eight byte-subset sums once and shares them across all columns, so a word costs eight lookups whatever its bit count, and row blocks split across threads.
- Gathering the bit witness walked every column serially. Whole-word column runs are now carved out of the witness and filled in parallel, packed columns in groups of eight adjacent columns so one source block row is read one cache line per group. A refusal still names the first non-Boolean column in placement order.
- `pack` copied a whole packing on one thread, and `PackedStack::from_columns` copied a lone column's packing a second time.

**The bit ring switch (sumcheck).**
- The supported run's equality table was built three times per opening: once for the tensor, once for the successor elements, once for the weight multilinear. It is built once in `prove`.
- Each weight is a subset sum of the batching table over an element's set coordinates; the 256 subset sums per byte are tabulated, so a weight costs one read per byte.
- The tensor accumulation added the right factor into every row the left factor's coordinates set — half the rows, one unpredictable branch each. Bucketing by whole bytes of the left factor makes it sixteen adds per term, read back once at the end. This one is a 16x on that pass.

**Single-column passes (sumcheck layout).** Stacking a witness, compressing the stacked slots and accumulating residual weights each ran one task per column slot, so the Boolean commitment's single 2^27-element column filled on one thread. Each slot's fill is now chunked.

**Zerocheck rounds (multi-stark).**
- Every fold task sets up a scratch holding every column — 2.9 MB at Blake3's 11536 columns — and Rayon's own splitting allocated and zeroed one per split. Rounds now split into a fixed few tasks per worker.
- The rounds a stage runs in the representation field evaluated one residual row per constraint pass. Where `R::Packing` covers several elements per register (vpclmulqdq + AVX2/AVX-512), one pass covers a lane group: lane `j` reads residual row `s + j`, carries its own row's eq weight, and the lanes of each per-node sum are added once the fold is over. Rows that no longer fill a group, and targets whose packing holds one element, keep the row-at-a-time kernel. `PackedGhash128` gains the algebra over the tower field `Ghash128` already has.
- Also fixes a latent crash: with the AIR evaluation inlined into it, the sliced rounds' fold helper reserved 44-68 KiB per level of Rayon's recursive split, and 64-thread runs overflowed the 2 MiB worker stack (3 of 8 runs at `-l 20`, Blake3 and Keccak alike). Same shape as #2183; only the per-node AIR evaluation is outlined, so the column fold stays inlined and the sliced rounds keep their speed.

## Measurements

c8a.16xlarge, `--profile optimized --features parallel`, `-C target-cpu=native`, `--security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4`, medians of three, prove time only:

| objective | l | main | this branch | |
|---|---|---|---|---|
| blake-3-compressions | 16 | 0.412 s | 0.271 s | 1.52x |
| blake-3-compressions | 18 | 1.345 s | 0.705 s | 1.91x |
| blake-3-compressions | 20 | 4.853 s | 2.266 s | 2.14x |
| keccak-f-permutations | 16 | 0.114 s | 0.084 s | 1.36x |
| keccak-f-permutations | 18 | 0.336 s | 0.216 s | 1.56x |
| keccak-f-permutations | 20 | 1.116 s | 0.670 s | 1.67x |

Proof digests are identical to main's in all twelve cases.

Blake3 at `-l 20`, by span: `prove` 4.85 s -> 2.27 s, of which the zerocheck 1.63 -> 1.29 s, `binary pcs open` 929 -> 356 ms (`combine_weights_in` 474 -> 36 ms, `compress_stacked` 153 -> 19 ms), witness stacking 157 -> 19 ms, and the previously unspanned column evaluation, gather and ring switch now 9 ms, 22 ms and 314 ms.

For scale, on the same box at 2^20 Blake3 compressions: 73,326 hashes/s for the KoalaBear-field FRI prover against 465,619 hashes/s here, an 84.25% reduction in proving time.

## Test plan

- `cargo test -p p3-binary-pcs -p p3-sumcheck -p p3-multi-stark --lib`, with and without `--features parallel`, on aarch64 and on x86_64 with `-C target-cpu=native` (the lane-group path only exists on the latter).
- `cargo clippy --all-targets --features parallel -- -D warnings` on the touched crates.
- New tests: packed column sums against a per-bit reference at narrow, short and multi-task shapes; the gather's placement and its refusal order; `CoordinateSums` against the per-coordinate walk; and the lane-group round against the row round.
- End to end: the twelve runs above, each checked against main's proof digest.

## Not done here

- `record_opening` (122 ms at `-l 20`) re-evaluates the committed polynomial at the surviving point, a value the ring-switch sumcheck already produced and drops. Reusing it needs a `Layout` API that takes a known value.
- The ring-switch sumcheck (149 ms) still runs in the tower basis; the inner PCS runs its own in the polynomial basis.
- `SLICED_ROUNDS = 4` helps Blake3 at `l >= 16` (zerocheck -14% at `l = 20`) but costs Keccak at every size I measured, so the cutoff is left at 3.


Nashtare:
@tcoratger I think what matters the most here is the following result on AMD EPYC (64 vCPUs) 🙂 

> For scale, on the same box at 2^20 Blake3 compressions: 73,326 hashes/s for the KoalaBear-field FRI prover against 465,619 hashes/s here, an 84.25% reduction in proving time.

tcoratger:
> @tcoratger I think what matters the most here is the following result on AMD EPYC (64 vCPUs) 🙂
> 
> > For scale, on the same box at 2^20 Blake3 compressions: 73,326 hashes/s for the KoalaBear-field FRI prover against 465,619 hashes/s here, an 84.25% reduction in proving time.

Sounds really great!

binary-pcs/src/boolean_trace.rs
```diff
@@ -397,56 +403,204 @@ where
...
+///
+/// So a word costs eight table reads, however many of its bits are set.
+///
+/// The tables are shared by every column of the block, whose words are contiguous.
+///
+/// # Arguments
+///
+/// - `words`: one row per block of 64 rows, one entry per column.
+/// - `row_weights`: one weight per row, at least as many as the column's rows.
+fn packed_column_sums<EF: Field>(words: &RowMajorMatrix<u64>, row_weights: &[EF]) -> Vec<EF> {
```
tcoratger, line 519: Maybe that should go in the matrix crate?

tcoratger: ah no this is specific to `RowMajorMatrix<u64>`, my bad

sumcheck/src/ring_switch/bits/tensor.rs
```diff
@@ -222,6 +222,71 @@ impl<EF: TowerLevel> BitTensor<EF> {
...
+    #[must_use]
+    pub fn into_tensor(self) -> BitTensor<EF> {
+        let mut tensor = BitTensor::zero();
+        for (position, bucket) in self.buckets.iter().enumerate() {
+            for (value, &sum) in bucket.iter().enumerate() {
+                // Byte value `value` sets coordinate `8 * position + bit` for each of its bits.
+                let mut bits = value as u8;
+                while bits != 0 {
+                    let bit = bits.trailing_zeros() as usize;
+                    tensor.rows[position * 8 + bit] += sum;
```
tcoratger: The production reduction is safe because BitPacking rejects sub-byte tower levels, but this new accumulator is publicly reachable through bits::tensor and its generic bound accepts every TowerLevel. BitTensorBuckets::<Gf2>::zero().into_tensor() reaches bucket value 2 and indexes tensor row 1 even though a Gf2 tensor has one row, so it panics. Since this is an internal optimization, could we make BitTensorBuckets and its methods pub(crate)? Alternatively, cap reconstruction at BitTensor::<EF>::DIMENSION. This is API hardening rather than a protocol blocker.

binary-pcs/src/boolean_trace.rs
```diff
@@ -397,56 +403,204 @@ where
...
+        "every row block needs at least one weight"
+    );
+    words
+        .values
+        .par_chunks(width * BLOCKS_PER_TASK)
+        .enumerate()
+        .par_fold_reduce(
+            || EF::zero_vec(width),
+            |mut sums, (task, blocks)| {
+                let mut tables = EF::zero_vec(BYTES_PER_WORD << 8);
```
tcoratger, line 532: Nice-to-have optimization: this allocates the 8 x 256 subset-sum table once for every 16-block iterator item. At 2^20 rows that is 1,024 allocations per view. The Rayon fold state could carry (column_sums, subset_tables), reusing the tables across all items assigned to that task; entries 1..256 are overwritten for each block and entry zero stays zero. Worth benchmarking around the evaluate-boolean-columns span; non-blocking.

multi-stark/src/rounds/repr.rs
```diff
@@ -257,6 +310,253 @@ where
...
+        let interaction_degrees = self
+            .interaction_groups
+            .iter()
+            .map(|group| group.degree)
+            .collect::<Vec<_>>();
+
+        // The eq weights cross into `R` one lane group at a time, as the rows they weigh do.
+        let weights = eq_suffix.as_slice();
+        let lift = |group: usize| lane_group(|lane| R::from(weights[group * lanes + lane]));
+        let eq_suffix: Vec<_> = if weights.len() < PARALLEL_FOLD_CELLS {
```
tcoratger: Nice-to-have optimization: these packed equality weights are materialized into a Vec and then consumed exactly once by the following fold. Driving the fold directly from (0..groups).into_par_iter() and constructing lane_group inside its fold closure would remove one large allocation plus a full write/read pass per representation-field round. This should be benchmarked end-to-end because fusing it may change scheduling; non-blocking.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -1085,12 +1122,10 @@ impl<EF: TranscriptField + TowerLevel> BitRingSwitch<EF> {
...
 
-        let tensor = self
-            .tensor(packing)
-            .expect("the packing was just checked against the reduction");
-        let successor = self
-            .successor_tensors(packing)
-            .expect("the packing was just checked against the reduction");
+        // One equality table over the supported run feeds every element and the weights.
+        let (_, offset, equality) = self.support();
+        let tensor = Self::tensor_over(packing, offset, &equality);
```
tcoratger, line 1127: Nice-to-have optimization: successor openings now walk the same packing and equality table once here and again in successor_tensors_over. A successor-specific combined pass could accumulate tensor, carry, and last together, removing one witness/equality traversal and one parallel dispatch. The tradeoff is a larger per-worker bucket accumulator, so I would gate this on the bit-ring-switch span benchmark; non-blocking.

tcoratger (approved):
Reviewed the full prover-only optimization set, including packed Boolean column sums and gathering, parallel packing copies, shared ring-switch equality tables, byte-bucket tensor accumulation, bounded Rayon task splitting, SIMD lane-group zerocheck evaluation, selector boundaries, successor rows, lookup coupling, transcript preservation, and affected tests. I found no soundness or supported-path correctness issue. The inline BitTensorBuckets note is public-API hardening; the other three comments are optional measured follow-up optimizations. Approved.

PR #2246, opened 2026-09-19
perf(binary): reuse the reduction's evaluation, and make the sliced-round count a choice

Nashtare:
Follow-ups to #2245, from a review of that branch's profile. Prover side only: every proof here is byte-identical to the one `main` emits, at every shape I measured.

Same harness as #2245: c8a.16xlarge (64-thread EPYC), `--profile optimized --features parallel`, `-C target-cpu=native`, `--security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4`, medians of three, prove time only.

| objective | l | #2245 | this branch | |
|---|---|---|---|---|
| blake-3-compressions | 16 | 0.273 s | 0.257 s | 1.06x |
| blake-3-compressions | 18 | 0.690 s | 0.661 s | 1.04x |
| blake-3-compressions | 20 | 2.229 s | 2.121 s | 1.05x |
| keccak-f-permutations | 16 | 0.080 s | 0.082 s | noise |
| keccak-f-permutations | 18 | 0.214 s | 0.201 s | 1.06x |
| keccak-f-permutations | 20 | 0.627 s | 0.576 s | 1.09x |

Proof digests are identical to main's in all six cases.

Trace generation sits outside `prove` and is not in that table. It moves much further:

| objective | l | main | this branch | |
|---|---|---|---|---|
| blake-3-compressions | 18 | 179 ms | 17.0 ms | 10.5x |
| blake-3-compressions | 20 | 612 ms | 47.0 ms | 13.0x |
| keccak-f-permutations | 18 | 6.54 ms | 5.63 ms | 1.16x |
| keccak-f-permutations | 20 | 11.9 ms | 7.73 ms | 1.54x |

End to end at `l = 20`, Blake3, that is 2.84 s of trace generation plus proving against
2.17 s: 369,000 hashes/s against 483,000.

## What each commit does

**The reduction's surviving value, reused.** The bit ring switch ends its rounds holding the packing's evaluation at the point it survives with, and returns it. The Boolean wrapper dropped that value and let the commitment evaluate the committed column again, one pass per opening. `Layout::eval_at_known` plays exactly the transcript `eval_at` plays over evaluations the caller supplies; `record_opening_known` records the claim they belong to. Without preprocessing rounds an opening carries no per-round residual, so the claim is the point and the value alone — which is why the route refuses a layout that folds. A supplied value is bound like any other and the verifier recomputes its own, so a wrong one yields a proof that does not verify; a debug assertion still runs the pass it replaces and compares. At `l = 20`, Blake3: `binary pcs open` 355 ms -> 239 ms.

**A lone whole-space claim's weight table.** The residual weights are a sum of per-claim tables, each written into the slot of the column it weighs. When one table of one column fills the stacked space and carries one directly-opened claim at `alpha^0`, that sum has a single term at unit scale: the claim's own table, read into a second one of the same size. The table becomes the output instead. That is the shape the Boolean wrapper always presents. At `l = 20`, Blake3: `combine_weights_in` 35.8 ms -> 12.5 ms, `into_sumcheck_in` 122 ms -> 93.5 ms.

**The sliced-round count, configurable.** A sliced round evaluates the AIR once per sixty-four residual rows instead of once per row, so each extra round doubles the plane work while halving the dense tail behind it. Where that crosses over depends on how much constraint work the AIR does per column, and the two hash AIRs fall on opposite sides of it. At `l = 20`, moving from three rounds to four takes the Blake3 zerocheck from 1.30 s to 1.22 s and the Keccak-f zerocheck from 423 ms to 657 ms — the wider trace (11536 columns) amortizes the AIR evaluation over its sixty-four rows, the narrower one (1625) does not.

`MultiStarkConfig::sliced_rounds` names the count and defaults to the three the kernel already ran, so **no caller changes and no measurement above reflects it**. The kernel caps it at the corner words one round holds on the stack, and a count of zero leaves the stage to another kernel, as a stage that does not fit its subfield is left. Only the plane rounds read it: every count proves the same statement, which the backend agreement test now covers across the whole range.

**Packed booleanity.** `Ghash128` answers `x * (x - 1)` as `x^2 + x`, which holds in characteristic 2 and lets the squaring kernel skip the cross-term carryless multiplies a general product pays for. The packed register kept the default product though it has the same squaring kernel. An AIR takes this on every booleanity constraint it states; it moves neither hash example, whose booleanity is carried by the Boolean commitment rather than by constraints. It is the register matching what the scalar already does.

**The binary traces, packed from their witness words.** Both packed generators built every row as one field element per cell, then read those cells back to decide which bit of the packed word to set. The witness is words to begin with: the Blake-3 G step and the Keccak round compute in `u32` and `u64`, and the columns are the bits of what they compute. The field carried a bit per element and took part in no arithmetic. Each row is now generated as its witness words and transposed into the block off those words — the Blake-3 row drops from 11536 field elements, 184 KiB written and read back per compression, to 364 words, and the Keccak permutation no longer clears 25 rows of field cells before each block that overlaps it. `g_words` and `generate_perm_words` hold the whole arithmetic, so the field rows and the packed words cannot drift, and the existing packed-against-dense tests pin the result.

## Measured and not taken

Four more of the ideas this came from did not survive measurement. Recording them so they are not retried:

- **Tiling `unslice` by column.** A cell is sixteen bytes, so four adjacent columns share a cache line, and folding one column at a time uses a quarter of each line it pulls in. Tiling four columns per task made it *worse*: 210 ms -> 251 ms at `l = 20`, Blake3. The pass is not bound by that traffic.
- **Packed folds in `fold_repr`.** The scalar fold is one multiply-add per element and `R::Packing` covers two or four. Folding over registers left the span unchanged (210.7 ms -> 208.6 ms) and prove within noise: the first fold moves tens of gibibytes and is bound by bandwidth, not by the multiply.
- **Dropping the successor planes** when no AIR reads a next row. The buffer is allocated zeroed and never written in that case, so its pages never become resident and it costs close to nothing already.
- **Walking the set bits** in the transpose instead of sweeping each word's bits in order. It was 2.5x worse on Keccak, whose state is half ones: the loop carries a dependency per bit where the sweep carries none.
- **Borrowing the committed packing** instead of copying it. Now spanned: `read committed packing` is 18.6 ms on Blake3 at `l = 20` and about 2 GiB of allocation. Removing it needs a lifetime on the public `BitPacking`, which I did not think that bought.

Still open, and the largest single item left: the bit ring switch's own sumcheck is 299 ms on Blake3 at `l = 20` and runs in the tower basis, while the inner PCS runs its in the polynomial basis.

## Test plan

- `cargo test -p p3-sumcheck -p p3-binary-pcs -p p3-multi-stark -p p3-multilinear-util -p p3-blake3-air -p p3-keccak-air`, with and without `--features parallel`, on aarch64 and on x86_64 with `-C target-cpu=native` (the packed booleanity change only compiles on the latter).
- `cargo clippy --all-targets --features parallel -- -D warnings` on the touched crates.
- New tests: the supplied-evaluation opening against the evaluating one, byte-identical proof and sponge, plus its two refusals; the lone-claim weight shortcut against the accumulating route; every sliced-round count against the generic backend's transcript; the packed booleanity check against the general product, lane by lane.
- End to end: the runs above, each checked against main's proof digest.

One note on the sliced-round count: the first version of that test failed, because a count of zero built a plane trace with no round to run and then tripped the round assertion. That is the guard now in `sliced_trace`.


tcoratger (commented):
Reviewed the full diff against `main`, with the changed files read in context rather than as a patch.

**No soundness or correctness defect.** I checked each of the five optimisations against the code it replaces and against the verifier that must reject a wrong claim: the ring-switch evaluation reuse plays the same transcript `eval_at` plays, the `sole_whole_space_table` shortcut returns the same `Poly` as the accumulating route, the `bool_check` squaring shortcut is the char-2 identity, and the sliced-round count cannot move the transcript at any reachable value.

The one real problem is that the headline new capability — a fourth sliced round — is **not exercised by the test added for it**, which I proved by mutation.

---

## [MAJOR] The new sliced-round test never reaches the new ceiling — `multi-stark/src/zerocheck/backend_tests.rs:935`

`every_sliced_round_count_agrees_with_the_generic_backend` loops `sliced_rounds` over `0..=MAX_SLICED_ROUNDS + 1`, but its fixtures are `1 << 9` rows:

```text
num_vars        = 9
LANE_VARIABLES  = SLICED_LANES.trailing_zeros() = 6
rounds          = sliced_rounds.min(4).min(num_vars - LANE_VARIABLES)
                = sliced_rounds.min(4).min(3)          <- capped at 3, always
```

So the loop actually tests 0, 1, 2, 3, 3, 3. The count 4 — the entire point of raising `SLICED_ROUNDS` to `MAX_SLICED_ROUNDS` — is never run.

This is not cosmetic, because 4 rounds reaches code that 3 rounds cannot. In `unslice`, `corners = 1 << challenges.len()`:

```rust
// multi-stark/src/rounds/sliced.rs:953
let mut words_of_plane = [0; 1 << MAX_SLICED_ROUNDS];   // 8 -> 16 entries
...
for (group, (low, high)) in low[..corners].chunks(8).zip(high[..corners].chunks(8)).enumerate() {
    *value += low_sums[group][usize::from(low)] + high_sums[group][usize::from(high)];
}
```

At 3 rounds `corners == 8`, so `chunks(8)` yields exactly **one** group and `group` is always `0`. At 4 rounds `corners == 16` and the loop runs **two** groups, indexing `low_sums[1]` / `high_sums[1]` for the first time in this codebase. The multi-group accumulation is brand-new and untested.

Verified three ways in a worktree at the PR head:

1. Inserting `assert_ne!(rounds, MAX_SLICED_ROUNDS)` in `sliced_trace` — the shipped test still **passes**, so 4 is never reached.
2. Injecting `if group > 0 { continue; }` into the `unslice` accumulation, deliberately dropping the second corner group — the shipped test still **passes**.
3. The same mutant is caught immediately by the same test at `1 << 10` rows.

**Fix:** raise both fixtures so `num_vars - LANE_VARIABLES == 4` and the ceiling is reachable.

```rust
Instance::honest(FixtureAir::Gate { scale: Tower::ONE }, 1 << 10, 20),
Instance::honest(FixtureAir::Pair, 1 << 10, 21),
```

**Verify:** with that change the mutant above fails on `"4 sliced rounds"` in ~1 s (debug). Worth also returning the realised round count from `sliced_trace` under `cfg(test)` and asserting it inside the loop, so a future height change cannot silently re-hollow the test.

---

## [MINOR] `try_open_at_known` panics after the sponge has moved, contradicting its own contract — `binary-pcs/src/pcs.rs:283`

The doc says *"Rejection leaves the challenger unchanged"*, and the three `validate_*` calls run before any transcript work. But the fourth failure mode is not validated:

```rust
// sumcheck/src/layout/prover/claims.rs:133
assert_eq!(
    self.folding, 0,
    "a supplied evaluation carries no preprocessing residual for a round to read"
);
```

That assertion fires from `record_opening_known`, which `eval_at_known` calls *after* `OpeningProverTranscript::new(challenger, shape)` — and that constructor folds the shape's domain separator into the sponge. So a layout with SVO preprocessing rounds gets a panic, not an error, and the sponge is already dirty.

Nothing in-tree commits a `BinaryPcs` witness with nonzero folding today (every `new_witness` call site passes `0`), so this is latent — but `try_open_at_known` is `pub`, `Layout` is `pub`, and the WHIR adapter over the same trait does run `FOLDING = 2`.

**Fix:** validate it alongside the others, before the transcript exists. `Layout::folding()` already exists:

```rust
self.validate_opening_protocol(protocol)?;
Self::validate_points(protocol, points)?;
Self::validate_evals(protocol, evals)?;
if prover_data.layout.folding() != 0 {
    return Err(BinaryPcsError::/* new variant */);
}
```

**Verify:** commit a witness with `SuffixProver::new_witness(tables, 1)`, call `try_open_at_known`, assert the error variant plus an unchanged challenger — the existing "a refused opening leaves the transcript alone" assertion is the right template.

---

## [MINOR] The evals-count mismatch reuses the *points* error, and the shape error reports totals — `binary-pcs/src/pcs.rs:328`

```rust
if protocol.num_openings() != evals.len() {
    return Err(BinaryPcsError::OpeningPointShapeMismatch);   // "prescribed opening points do not match..."
}
...
if !batch.has_same_shape(batch_evals) {
    return Err(BinaryPcsError::OpeningBatchSizeMismatch {
        table_idx, expected: batch.len(), actual: batch_evals.len(),
    });
}
```

Two mismatches between predicate and report. A wrong number of *evaluation batches* surfaces as a message about *points*; and `has_same_shape` compares the two sides separately while `len()` is their sum, so a request of `current=[a,b], next=[]` against evals of shape `current=1, next=1` is correctly rejected but reports `expected 2, got 2` — unfalsifiable from the message alone.

**Fix:** add an `OpeningEvalShapeMismatch` variant for the count, and carry the per-side lengths in the shape error.

---

## [MINOR] The documented ceiling is not reachable from the public API — `multi-stark/src/config.rs:18`

`DEFAULT_SLICED_ROUNDS` is `pub`; `MAX_SLICED_ROUNDS` is `pub(crate)`. The `sliced_rounds` doc says *"Counts above the kernel's ceiling are capped by it"* without saying what the ceiling is, so a caller implementing `MultiStarkConfig` cannot tell that `sliced_rounds() = 8` silently means `4`. Silent clamping of a tuning knob turns into a bug report about a benchmark that did not move.

**Fix:** re-export `MAX_SLICED_ROUNDS` from `config` (or mirror it), name it in the trait doc, and add `debug_assert!(self.sliced_rounds() <= MAX_SLICED_ROUNDS)` at the prover entry so an over-large value is loud in tests and free in release.

---

## [MINOR] `unslice`'s hottest closure now zeroes twice the stack it uses at the default count — `multi-stark/src/rounds/sliced.rs:953`

```rust
let mut words_of_plane = [0; 1 << MAX_SLICED_ROUNDS];   // 16 u64 = 128 B
for (corner, value) in words_of_plane[..corners].iter_mut().enumerate() { ... }
```

`corner_words` is called twice per `fold_word`, and `fold_word` runs `width * words` times plus the successor tails. Raising `MAX_SLICED_ROUNDS` from 3 to 4 doubled the array from 64 B to 128 B, and the `[0; N]` initialiser zeroes all of it while the `[..corners]` fill (a dynamic bound, so not unrollable) touches only half at the default count of 3. The PR measures this pass at 210 ms on Blake3 at `l = 20`, so it is worth not regressing for callers who never raise the count.

**Fix:** keep a fixed `[u64; 8]` per group and drive the group loop directly, so the buffer size stops tracking the ceiling.

**Verify:** the PR's own `unslice` span at `l = 20`, Blake3, `sliced_rounds` left at 3 — should be within noise of #2245's 210 ms, not above it.

---

## [NIT]

- **`MAX_SLICED_ROUNDS` doc states the wrong growth law** — `multi-stark/src/rounds/sliced.rs:41`. "One corner word per bound variable is held on the stack" — it is one corner word per *corner*, and there are `2^rounds` corners per plane, two planes. The stack cost is exponential in the count, not linear, which is precisely why a ceiling is needed; the doc undersells its own justification.

- **Nothing pins `ROW_WORD_BITS` to the AIR width at compile time** — `blake3-air/src/binary/generation.rs:143`. `sum(ROW_WORD_BITS)` must equal `NUM_BLAKE3_BINARY_COLS` (both 11536: `28*32 + 56*(4*32 + 2*31)`). If `Blake3BinaryCols` gains or reorders a field, the transpose either panics on `block[column..column + bits]` or silently leaves trailing columns zero. The packed-vs-dense tests catch it, but a const assertion catches it at compile time and points at the right line:

  ```rust
  const _: () = {
      let mut total = 0usize;
      let mut i = 0;
      while i < NUM_ROW_WORDS { total += ROW_WORD_BITS[i] as usize; i += 1; }
      assert!(total == NUM_BLAKE3_BINARY_COLS);
  };
  ```

- **`Layout` gains a required method with no default body** — `sumcheck/src/layout/prover/mod.rs:399`. `record_opening_known` breaks any out-of-tree implementor. Flagging rather than recommending, since the crate is pre-1.0.

---

## Test coverage

**The four-round sliced path.** Covered under [MAJOR] above.

**Packed `bool_check` over random lanes** — `binary-field/src/packed/x86_64.rs:688`. The new test walks `SPECIAL = [0, u128::MAX, 1 << 127, 0x87]` and separately checks the two roots as *uniform* registers. Two gaps: `SPECIAL` does not contain `Ghash128::ONE`, so no test ever places `ONE` in one lane and an extreme in another; and a lane-crossing bug that happens to produce `ZERO` on a uniform-`ONE` register survives. The scalar path already has `bool_check_agrees_with_the_vanishing_polynomial` as a proptest (`binary-field/src/ghash.rs:786`) and this module already imports `proptest::prelude::*`, so mirroring it is nearly free and strictly stronger than the corner table:

```rust
proptest! {
    #[test]
    fn packed_bool_check_agrees_with_the_general_product(lanes in any::<[u128; WIDTH]>()) {
        let scalars: [Ghash128; WIDTH] =
            core::array::from_fn(|i| Ghash128::from_le_bytes(lanes[i].to_le_bytes()));
        let packed: PackedGhash128 = PackedValue::from_fn(|i| scalars[i]);
        let got = packed.bool_check();
        for (lane, &x) in scalars.iter().enumerate() {
            prop_assert_eq!(got.as_slice()[lane], x * (x - Ghash128::ONE));
        }
    }
}
```

Keep the corner-table test alongside it — known-answer corners and random lanes cover different failures. Minor: the loop calls `packed.bool_check()` three times per lane; hoist it.

**`sole_whole_space_table`'s rejection arms** — `sumcheck/src/layout/prover/suffix.rs:928`. The added test covers the accepting shape against the accumulating route, which is the important direction. The five ways the shortcut must *decline* have no test: a virtual claim present, more than one source table, a table narrower than the stacked space, two claims on the one column, and a successor-only opening (`is_next == true`). A bug in any guard silently returns the wrong `Poly` rather than falling through. A table-driven test asserting `combine_weights_in == combine_weights` for each shape would close it, and would also protect the `tables.get_mut(claim_idx)?.current.take()` path from a future `None`.

**Supplied-evaluation opening with a successor batch.** Every in-tree caller passes `next: Vec::new()`, and the new tests use `OpeningBatch::new(vec![0], Vec::new())`, so the `next` half of `record_known` and of `known_evals_agree` (`point.eval_next_suffix(poly, None)`) is dead in tests. A single opening with `current=[0], next=[0]` on a two-variable table, checked against `try_open_at`'s bytes exactly as `supplying_the_evaluation_leaves_the_proof_byte_identical` does, would exercise it.

---

## Could not verify

- The `l = 20` timings are from a 64-thread c8a.16xlarge I do not have. I checked that the *shapes* of the wins are plausible (a skipped `O(2^n)` pass per opening, a skipped zero-init plus pass in `combine_weights_in`, 11536 field cells replaced by 364 words per Blake3 row) but reproduced no number.
- The claim of proof digests identical to `main` at all six measured shapes: I verified the *mechanism* but did not run an end-to-end digest comparison.
- Whether `BinaryPcs::commit` should itself reject a nonzero-folding witness — that may be a deliberate allowance for a future layout, in which case the [MINOR] fix belongs at `try_open_at_known` only, as written.

---

**Bottom line:** merge after raising the `every_sliced_round_count_agrees_with_the_generic_backend` fixtures to `1 << 10`. Everything else here is MINOR or NIT and can follow.

multi-stark/src/zerocheck/backend_tests.rs
```diff
@@ -911,3 +924,27 @@ fn fixture_degrees_are_the_named_ones() {
...
 }
+
+/// Every sliced-round count must leave the proof the generic backend's.
+///
+/// The count only decides how many rounds run on the planes. Each of those rounds computes
+/// the polynomial the generic kernel computes, so no count may move the transcript.
+#[test]
+fn every_sliced_round_count_agrees_with_the_generic_backend() {
+    let instances = [
+        Instance::honest(FixtureAir::Gate { scale: Tower::ONE }, 1 << 9, 20),
```
tcoratger: **[MAJOR] The new sliced-round test never reaches the new ceiling — `multi-stark/src/zerocheck/backend_tests.rs:935`**

`every_sliced_round_count_agrees_with_the_generic_backend` loops `sliced_rounds` over `0..=MAX_SLICED_ROUNDS + 1`, but its fixtures are `1 << 9` rows:

```text
num_vars        = 9
LANE_VARIABLES  = SLICED_LANES.trailing_zeros() = 6
rounds          = sliced_rounds.min(4).min(num_vars - LANE_VARIABLES)
                = sliced_rounds.min(4).min(3)          <- capped at 3, always
```

So the loop actually tests 0, 1, 2, 3, 3, 3. The count 4 — the entire point of raising `SLICED_ROUNDS` to `MAX_SLICED_ROUNDS` — is never run.

This is not cosmetic, because 4 rounds reaches code that 3 rounds cannot. In `unslice`, `corners = 1 << challenges.len()`:

```rust
// multi-stark/src/rounds/sliced.rs:953
let mut words_of_plane = [0; 1 << MAX_SLICED_ROUNDS];   // 8 -> 16 entries
...
for (group, (low, high)) in low[..corners].chunks(8).zip(high[..corners].chunks(8)).enumerate() {
    *value += low_sums[group][usize::from(low)] + high_sums[group][usize::from(high)];
}
```

At 3 rounds `corners == 8`, so `chunks(8)` yields exactly **one** group and `group` is always `0`. At 4 rounds `corners == 16` and the loop runs **two** groups, indexing `low_sums[1]` / `high_sums[1]` for the first time in this codebase. The multi-group accumulation is brand-new and untested.

Verified three ways in a worktree at the PR head:

1. Inserting `assert_ne!(rounds, MAX_SLICED_ROUNDS)` in `sliced_trace` — the shipped test still **passes**, so 4 is never reached.
2. Injecting `if group > 0 { continue; }` into the `unslice` accumulation, deliberately dropping the second corner group — the shipped test still **passes**.
3. The same mutant is caught immediately by the same test at `1 << 10` rows.

**Fix:** raise both fixtures so `num_vars - LANE_VARIABLES == 4` and the ceiling is reachable.

```rust
Instance::honest(FixtureAir::Gate { scale: Tower::ONE }, 1 << 10, 20),
Instance::honest(FixtureAir::Pair, 1 << 10, 21),
```

**Verify:** with that change the mutant above fails on `"4 sliced rounds"` in ~1 s (debug). Worth also returning the realised round count from `sliced_trace` under `cfg(test)` and asserting it inside the loop, so a future height change cannot silently re-hollow the test.

binary-pcs/src/pcs.rs
```diff
@@ -263,6 +263,51 @@ where
...
+    /// Rejection leaves the challenger unchanged.
+    ///
+    /// Prover data must match the committed tables.
+    ///
+    /// # Soundness
+    ///
+    /// A supplied evaluation is bound exactly as a computed one is, and a verifier recomputes
+    /// its own. A wrong one therefore yields a proof that does not verify, never one that does.
+    #[tracing::instrument(name = "binary pcs open", skip_all)]
+    pub fn try_open_at_known<Challenger>(
```
tcoratger, line 303: **[MINOR] `try_open_at_known` panics after the sponge has moved, contradicting its own contract — `binary-pcs/src/pcs.rs:283`**

The doc says *"Rejection leaves the challenger unchanged"*, and the three `validate_*` calls run before any transcript work. But the fourth failure mode is not validated:

```rust
// sumcheck/src/layout/prover/claims.rs:133
assert_eq!(
    self.folding, 0,
    "a supplied evaluation carries no preprocessing residual for a round to read"
);
```

That assertion fires from `record_opening_known`, which `eval_at_known` calls *after* `OpeningProverTranscript::new(challenger, shape)` — and that constructor folds the shape's domain separator into the sponge. So a layout with SVO preprocessing rounds gets a panic, not an error, and the sponge is already dirty.

Nothing in-tree commits a `BinaryPcs` witness with nonzero folding today (every `new_witness` call site passes `0`), so this is latent — but `try_open_at_known` is `pub`, `Layout` is `pub`, and the WHIR adapter over the same trait does run `FOLDING = 2`.

**Fix:** validate it alongside the others, before the transcript exists. `Layout::folding()` already exists:

```rust
self.validate_opening_protocol(protocol)?;
Self::validate_points(protocol, points)?;
Self::validate_evals(protocol, evals)?;
if prover_data.layout.folding() != 0 {
    return Err(BinaryPcsError::/* new variant */);
}
```

**Verify:** commit a witness with `SuffixProver::new_witness(tables, 1)`, call `try_open_at_known`, assert the error variant plus an unchanged challenger — the existing "a refused opening leaves the transcript alone" assertion is the right template.

binary-pcs/src/pcs.rs
```diff
@@ -279,6 +324,26 @@ where
         Ok(())
     }
 
+    /// Check that supplied evaluations cover every batch the protocol names, column for column.
+    fn validate_evals(
```
tcoratger, line 364: **[MINOR] The evals-count mismatch reuses the *points* error, and the shape error reports totals — `binary-pcs/src/pcs.rs:328`**

```rust
if protocol.num_openings() != evals.len() {
    return Err(BinaryPcsError::OpeningPointShapeMismatch);   // "prescribed opening points do not match..."
}
...
if !batch.has_same_shape(batch_evals) {
    return Err(BinaryPcsError::OpeningBatchSizeMismatch {
        table_idx, expected: batch.len(), actual: batch_evals.len(),
    });
}
```

Two mismatches between predicate and report. A wrong number of *evaluation batches* surfaces as a message about *points*; and `has_same_shape` compares the two sides separately while `len()` is their sum, so a request of `current=[a,b], next=[]` against evals of shape `current=1, next=1` is correctly rejected but reports `expected 2, got 2` — unfalsifiable from the message alone.

**Fix:** add an `OpeningEvalShapeMismatch` variant for the count, and carry the per-side lengths in the shape error.

multi-stark/src/config.rs
```diff
@@ -7,6 +7,16 @@ use p3_commit::MultilinearPcs;
...
 
+/// Zerocheck rounds a stage evaluates on its bit-sliced planes by default.
+///
+/// A sliced round evaluates the AIR once per sixty-four residual rows instead of once per row,
+/// so each extra round doubles the plane work while halving the dense tail that follows it.
+/// Where the crossover falls depends on how much constraint work the AIR does per column, not
+/// on the trace height, so no single count suits every AIR.
+///
+/// Three is the count that does not lose on either of the two hash AIRs measured.
+pub const DEFAULT_SLICED_ROUNDS: usize = 3;
```
tcoratger, line 20: **[MINOR] The documented ceiling is not reachable from the public API — `multi-stark/src/config.rs:18`**

`DEFAULT_SLICED_ROUNDS` is `pub`; `MAX_SLICED_ROUNDS` is `pub(crate)`. The `sliced_rounds` doc says *"Counts above the kernel's ceiling are capped by it"* without saying what the ceiling is, so a caller implementing `MultiStarkConfig` cannot tell that `sliced_rounds() = 8` silently means `4`. Silent clamping of a tuning knob turns into a bug report about a benchmark that did not move.

**Fix:** re-export `MAX_SLICED_ROUNDS` from `config` (or mirror it), name it in the trait doc, and add `debug_assert!(self.sliced_rounds() <= MAX_SLICED_ROUNDS)` at the prover entry so an over-large value is loud in tests and free in release.

multi-stark/src/rounds/sliced.rs
```diff
@@ -940,7 +950,7 @@ where
         // The value at every residual row of one word, from the corner words of both planes.
         let fold_word = |planes: &[[u64; 2]], column: usize, word: usize, out: &mut [R]| {
             let corner_words = |plane: usize| {
-                let mut words_of_plane = [0; 1 << SLICED_ROUNDS];
+                let mut words_of_plane = [0; 1 << MAX_SLICED_ROUNDS];
```
tcoratger: **[MINOR] `unslice`'s hottest closure now zeroes twice the stack it uses at the default count — `multi-stark/src/rounds/sliced.rs:953`**

```rust
let mut words_of_plane = [0; 1 << MAX_SLICED_ROUNDS];   // 16 u64 = 128 B
for (corner, value) in words_of_plane[..corners].iter_mut().enumerate() { ... }
```

`corner_words` is called twice per `fold_word`, and `fold_word` runs `width * words` times plus the successor tails. Raising `MAX_SLICED_ROUNDS` from 3 to 4 doubled the array from 64 B to 128 B, and the `[0; N]` initialiser zeroes all of it while the `[..corners]` fill (a dynamic bound, so not unrollable) touches only half at the default count of 3. The PR measures this pass at 210 ms on Blake3 at `l = 20`, so it is worth not regressing for callers who never raise the count.

**Fix:** keep a fixed `[u64; 8]` per group and drive the group loop directly, so the buffer size stops tracking the ceiling.

**Verify:** the PR's own `unslice` span at `l = 20`, Blake3, `sliced_rounds` left at 3 — should be within noise of #2245's 210 ms, not above it.

tcoratger (commented):

---

Reviewed the full diff against `main`, with the changed files read in context rather than as a patch.

**No soundness or correctness defect.** I checked each of the five optimisations against the code it replaces and against the verifier that must reject a wrong claim: the ring-switch evaluation reuse plays the same transcript `eval_at` plays, the `sole_whole_space_table` shortcut returns the same `Poly` as the accumulating route, the `bool_check` squaring shortcut is the char-2 identity, and the sliced-round count cannot move the transcript at any reachable value.

The one real problem is that the headline new capability — a fourth sliced round — is **not exercised by the test added for it**, which I proved by mutation.

---

### [NIT]

- **`MAX_SLICED_ROUNDS` doc states the wrong growth law** — `multi-stark/src/rounds/sliced.rs:41`. "One corner word per bound variable is held on the stack" — it is one corner word per *corner*, and there are `2^rounds` corners per plane, two planes. The stack cost is exponential in the count, not linear, which is precisely why a ceiling is needed; the doc undersells its own justification.

- **Nothing pins `ROW_WORD_BITS` to the AIR width at compile time** — `blake3-air/src/binary/generation.rs:143`. `sum(ROW_WORD_BITS)` must equal `NUM_BLAKE3_BINARY_COLS` (both 11536: `28*32 + 56*(4*32 + 2*31)`). If `Blake3BinaryCols` gains or reorders a field, the transpose either panics on `block[column..column + bits]` or silently leaves trailing columns zero. The packed-vs-dense tests catch it, but a const assertion catches it at compile time and points at the right line:

  ```rust
  const _: () = {
      let mut total = 0usize;
      let mut i = 0;
      while i < NUM_ROW_WORDS { total += ROW_WORD_BITS[i] as usize; i += 1; }
      assert!(total == NUM_BLAKE3_BINARY_COLS);
  };
  ```

- **`Layout` gains a required method with no default body** — `sumcheck/src/layout/prover/mod.rs:399`. `record_opening_known` breaks any out-of-tree implementor. Flagging rather than recommending, since the crate is pre-1.0.

---

## Test coverage

**The four-round sliced path.** Covered under [MAJOR] above.

**Packed `bool_check` over random lanes** — `binary-field/src/packed/x86_64.rs:688`. The new test walks `SPECIAL = [0, u128::MAX, 1 << 127, 0x87]` and separately checks the two roots as *uniform* registers. Two gaps: `SPECIAL` does not contain `Ghash128::ONE`, so no test ever places `ONE` in one lane and an extreme in another; and a lane-crossing bug that happens to produce `ZERO` on a uniform-`ONE` register survives. The scalar path already has `bool_check_agrees_with_the_vanishing_polynomial` as a proptest (`binary-field/src/ghash.rs:786`) and this module already imports `proptest::prelude::*`, so mirroring it is nearly free and strictly stronger than the corner table:

```rust
proptest! {
    #[test]
    fn packed_bool_check_agrees_with_the_general_product(lanes in any::<[u128; WIDTH]>()) {
        let scalars: [Ghash128; WIDTH] =
            core::array::from_fn(|i| Ghash128::from_le_bytes(lanes[i].to_le_bytes()));
        let packed: PackedGhash128 = PackedValue::from_fn(|i| scalars[i]);
        let got = packed.bool_check();
        for (lane, &x) in scalars.iter().enumerate() {
            prop_assert_eq!(got.as_slice()[lane], x * (x - Ghash128::ONE));
        }
    }
}
```

Keep the corner-table test alongside it — known-answer corners and random lanes cover different failures. Minor: the loop calls `packed.bool_check()` three times per lane; hoist it.

**`sole_whole_space_table`'s rejection arms** — `sumcheck/src/layout/prover/suffix.rs:928`. The added test covers the accepting shape against the accumulating route, which is the important direction. The five ways the shortcut must *decline* have no test: a virtual claim present, more than one source table, a table narrower than the stacked space, two claims on the one column, and a successor-only opening (`is_next == true`). A bug in any guard silently returns the wrong `Poly` rather than falling through. A table-driven test asserting `combine_weights_in == combine_weights` for each shape would close it, and would also protect the `tables.get_mut(claim_idx)?.current.take()` path from a future `None`.

**Supplied-evaluation opening with a successor batch.** Every in-tree caller passes `next: Vec::new()`, and the new tests use `OpeningBatch::new(vec![0], Vec::new())`, so the `next` half of `record_known` and of `known_evals_agree` (`point.eval_next_suffix(poly, None)`) is dead in tests. A single opening with `current=[0], next=[0]` on a two-variable table, checked against `try_open_at`'s bytes exactly as `supplying_the_evaluation_leaves_the_proof_byte_identical` does, would exercise it.

---

## Could not verify

- The `l = 20` timings are from a 64-thread c8a.16xlarge I do not have. I checked that the *shapes* of the wins are plausible (a skipped `O(2^n)` pass per opening, a skipped zero-init plus pass in `combine_weights_in`, 11536 field cells replaced by 364 words per Blake3 row) but reproduced no number.
- The claim of proof digests identical to `main` at all six measured shapes: I verified the *mechanism* but did not run an end-to-end digest comparison.
- Whether `BinaryPcs::commit` should itself reject a nonzero-folding witness — that may be a deliberate allowance for a future layout, in which case the [MINOR] fix belongs at `try_open_at_known` only, as written.

---

**Bottom line:** merge after raising the `every_sliced_round_count_agrees_with_the_generic_backend` fixtures to `1 << 10`. Everything else here is MINOR or NIT and can follow.

multi-stark/src/zerocheck/backend_tests.rs
tcoratger: **[MAJOR — still standing] The sliced-round test still never reaches the new ceiling.**

Unchanged since my last review. With `1 << 9` rows, `rounds = sliced_rounds.min(MAX_SLICED_ROUNDS).min(num_vars - LANE_VARIABLES) = min(4).min(3)`, so the loop runs 0, 1, 2, 3, 3, 3 and the count 4 — the capability this PR adds — is never exercised. `unslice`'s two-corner-group path (`low_sums[1]`/`high_sums[1]`, reached only at `corners == 16`) stays dead.

Re-verified by mutation at `16db71d3`: `if group > 0 { continue; }` in the accumulation at `multi-stark/src/rounds/sliced.rs:961` leaves this test green. Raise both fixtures here to `1 << 10` and the same mutant fails on `"4 sliced rounds"`.

tcoratger (commented):
Re-reviewed at `16db71d3`. **Nothing in this PR changed.** `git range-diff 3cbccb3c..73cfa296 e08d9b38..16db71d3` reports all five commits identical; the only delta since my review is the rebase onto `e08d9b38`. So every prior finding stands as written, and the [MAJOR] is still live.

I re-ran the mutation at the current head, both directions:

- Injecting `if group > 0 { continue; }` into `unslice`'s corner-group accumulation (`multi-stark/src/rounds/sliced.rs:961`) — `every_sliced_round_count_agrees_with_the_generic_backend` still **passes**. The second corner group is dropped and no test notices.
- With the two fixtures at that call site raised to `1 << 10` and the same mutant in place, the test **fails** on `"4 sliced rounds"` in 0.02 s (release). Mutation and fixture change both reverted.

So the one-line fix is confirmed to work and is confirmed to be still missing. The four [MINOR]s and three [NIT]s are unchanged too; none of them block.

Happy to approve as soon as the fixtures move.

multi-stark/src/zerocheck/backend_tests.rs
Nashtare: Reproduced both mutations before touching anything. At `1 << 9` rows `rounds = sliced_rounds.min(4).min(9 - 6) = 3`, and `if group > 0 { continue; }` in the accumulation leaves this test green; with both fixtures at `1 << 10` the same mutant fails on `4 sliced rounds`. Your reading of the cap is exactly right.

Fixed in `test(multi-stark): run the backend agreement over the whole sliced-round range`. Rather than writing the literal `1 << 10`, the fixture height is derived from the ceiling:

```rust
const LANE_VARIABLES: usize = SLICED_LANES.trailing_zeros() as usize;
const HEIGHT: usize = 1 << (MAX_SLICED_ROUNDS + LANE_VARIABLES);
```

so the top of the range stays reachable whatever `MAX_SLICED_ROUNDS` becomes, and a future height change cannot re-hollow the loop because there is no longer a height to change independently.

I did not add the `cfg(test)` return of the realised count from `sliced_trace`. Of the three inputs to `rounds`, the loop varies the first, the ceiling is now the height, and the remaining way for the test to go hollow — a fixture drifting out of `GF(4)` so no stage is sliced at all — is already pinned next door by `round_poly_sliced(..).expect("the stage is sliced")` in `rounds::sliced::tests` (`multi-stark/src/rounds/sliced/tests.rs:296`). That seemed better than a production-side test hook.

Nashtare: Answered in full on the follow-up thread (`4057791180`). Short version: reproduced, both ways — the `group > 0` mutant survives at `1 << 9` and fails on `4 sliced rounds` at `1 << 10` — and fixed in `test(multi-stark): run the backend agreement over the whole sliced-round range`, with the fixture height derived from `MAX_SLICED_ROUNDS + SLICED_LANES.trailing_zeros()` rather than written as a literal.

binary-pcs/src/pcs.rs
Nashtare: Verified, and it is wider than `try_open_at_known`. A witness committed with `SuffixProver::new_witness(tables, 1)` commits fine — `prover::commit` hands `PcsLayout::commit` the crate's `FOLDING = 0` while the layout keeps the witness's own depth, so the two disagree from that point on — and then every opening entry point panics mid-transcript:

* `try_open_at_known` at `sumcheck/src/layout/prover/claims.rs:133`, as you describe;
* `try_open_at` (reproduced) and `try_open` (same `finish_open`) at `binary-pcs/src/prover.rs:195`, `assertion left == right failed: the commit phase runs at zero preprocessing depth, so the sumcheck consumes no head rounds`, raised from inside `fold_rounds_with` — i.e. after `BinaryPcsProverTranscript::new` folded the domain separator in.

So I took your fix but hoisted it to all three, since all three document a rejection that leaves the challenger alone: `Self::validate_preprocessing_depth(&prover_data)?` runs alongside the other shape checks, before any transcript exists, and returns a new `BinaryPcsError::OpeningPreprocessingDepth { folding }`. `fix(binary-pcs): refuse a layout with preprocessing rounds before the transcript moves` (`binary-pcs/src/pcs.rs:337`).

Test built on the template you named: commit at depth one, assert the variant from all three entry points, then assert the sponge is where it was.

binary-pcs/src/pcs.rs
Nashtare: Both confirmed. `OpeningBatch::len()` is `current.len() + next.len()` (`sumcheck/src/table.rs:93`) while `has_same_shape` compares the sides (`:108`), so your `current=[a,b], next=[]` against a `current=1, next=1` list is refused with `expected 2, got 2`.

Fixed in `fix(binary-pcs): report a wrong evaluation count and a batch's two sides apart`:

* the batch-count mismatch gets its own variant. I named it `OpeningEvalCountMismatch { expected, actual }` rather than `OpeningEvalShapeMismatch`, so it reads as a count and sits with the existing `OpeningBatchCountMismatch`;
* `OpeningBatchSizeMismatch` now carries `expected_current` / `expected_next` / `actual_current` / `actual_next`, built by one helper (`binary-pcs/src/pcs.rs:65`) shared with the verifier-side check at `pcs.rs:569`, which reported the same unfalsifiable totals.

The refusal test gains a third case — one batch and one value, but on the successor side — which is the shape the old message could not distinguish.

multi-stark/src/config.rs
Nashtare: Fixed in `docs(multi-stark): expose the sliced-round ceiling next to the knob that it caps`: `MAX_SLICED_ROUNDS` is now `pub` and re-exported from `config`, so it sits beside `DEFAULT_SLICED_ROUNDS`, and the trait doc names it. The doc also names the second cap you did not mention and a caller could see even less of — the row variables a stage keeps once a word's lanes are spent, which holds a short stage below the ceiling whatever the configuration asks for.

I did not add `debug_assert!(self.sliced_rounds() <= MAX_SLICED_ROUNDS)`. The clamp is the documented contract, not an accident, and `every_sliced_round_count_agrees_with_the_generic_backend` deliberately runs `0..=MAX_SLICED_ROUNDS + 1` to show that a count above the ceiling proves the same statement — so the assert would fire on an input the doc promises to accept, and on that test. Exporting the number gives the caller what they need to read a benchmark that did not move, without turning a tolerated value into a test-only abort.

One note on framing: this and the MAJOR are two defects, not one seen twice. That one is a fixture height that held the realised count at three; this one is visibility of the constant. They meet only in the constant itself, and neither fix would have closed the other.

multi-stark/src/rounds/sliced.rs
Nashtare: The arithmetic is right: the array is `[0; 1 << MAX_SLICED_ROUNDS]`, this PR took it from 8 to 16 `u64`, `corner_words` runs twice per `fold_word`, and `[..corners]` fills 8 of the 16 at the default count.

Done as you suggest, in `perf(multi-stark): gather one subset-sum table's corners at a time in unslice`. The gather is now driven by the subset-sum tables directly (`multi-stark/src/rounds/sliced.rs:962`), one `[u64; CORNERS_PER_TABLE]` pair per table, so the buffers stop tracking the ceiling; it also reads each corner's two planes in a single pass rather than one pass per plane. `MAX_SLICED_ROUNDS`'s doc no longer claims a stack rationale it no longer has.

What I could not do is your verification. There is no benchmark host on this branch, and the machine I have is a 14-core M4 shared with other builds, where Blake3 at `l = 20` does not fit and a 210 ms span would be noise — so I have not measured the `unslice` span and am not claiming a number. What I can say is that the change is a strict reduction in work at every count (half the zeroing at three, one index computation per corner instead of two, and nothing added at four), and that the backend-agreement test pins the result to the generic kernel byte for byte across the whole round range. The span itself is worth a look on the next c8a run.

tcoratger (approved):
All six findings verified fixed at `130f3361`, by execution in a worktree at the head.

**The MAJOR — the sliced-round test now reaches the ceiling.** Confirmed positively first: a probe recording every realised round count from `sliced_trace` returns the mask `0b11110`, so the loop actually runs `rounds` = 1, 2, 3 and **4** (count 0 correctly takes the generic kernel and never builds a sliced trace). Deriving `HEIGHT` from `MAX_SLICED_ROUNDS + LANE_VARIABLES` is the better fix — the height can no longer drift out from under the range.

Then by mutation, against the rewritten gathering in `unslice`:

| mutation | result |
| --- | --- |
| `if table > 0 { continue; }` (drop the second corner group) | **FAILS** on `4 sliced rounds` |
| `(corner * words + word)` — drop the table offset | **FAILS** on `4 sliced rounds` |
| `spanned.min(7)` — one corner short per table | **FAILS** on `3 sliced rounds` |
| swap `cell[0]` / `cell[1]` | **FAILS** on `1 sliced rounds` |

The exact mutant that survived at the old head now dies on the exact case it was blind to.

**The `unslice` rewrite (`a105c156`)** — read in full and checked for equivalence, since it rewrites the code the MAJOR said was untested. It is equivalent: `corners == weights.len()` is a power of two, so `low_sums` holds `ceil(corners / 8)` tables and `spanned = (corners - first).min(8)` reproduces each `chunks(8)` length exactly, with no underflow (`first < corners` for every table in range). The global corner index `first + corner` matches the old `corner`, and `lane_masks(&low[..spanned])` sees the same slice the old chunk was. It is a strict reduction in work — 8 zeroed words per table instead of 16 per plane, and one index computation per corner instead of two — and the four mutations above show the agreement test pins every part of it.

**The four MINORs**

1. **Stack zeroing** — the buffers are `[u64; CORNERS_PER_TABLE]` per table and no longer track `MAX_SLICED_ROUNDS`; the doc no longer claims the stack rationale. I have not measured the span either, and agree the direction is unambiguous.
2. **Ceiling exposed** — verified reachable from outside: an integration test (a separate crate) compiles against `p3_multi_stark::config::MAX_SLICED_ROUNDS` and reads `4`. Your argument against the `debug_assert!` is right — the clamp is the documented contract and the test deliberately runs `MAX_SLICED_ROUNDS + 1`, so the assert would abort on an input the doc promises to accept. Naming the second cap in the trait doc is a good addition.
3. **Preprocessing depth** — `validate_preprocessing_depth` runs alongside the other shape checks in all three entry points, before `BinaryPcsProverTranscript::new` / `OpeningProverTranscript::new` exists, and returns `OpeningPreprocessingDepth { folding }`. Hoisting it to `try_open` and `try_open_at` is the right call; the test asserts the variant from all three and that the sponge is where it was.
4. **Evals count / batch shape** — distinct `OpeningEvalCountMismatch`, and `OpeningBatchSizeMismatch` carries all four side lengths through a helper shared with the verifier-side check. The new `current=[], next=[sample]` case is exactly the shape the old totals could not distinguish.

**Suites** — `p3-multi-stark` + `p3-binary-pcs` green in debug (so `debug_assert!`s fire) and release; `cargo clippy --all-targets` and `cargo fmt --check` clean. Worktree reverted and clean after every experiment.

Approving.

PR #2247, opened 2026-09-20
perf: run the bit ring-switch reduction in the polynomial basis

Nashtare:
Stacked on #2246.

A tower-basis product in `GF(2^128)` is three changes of basis around one carryless multiply; in
the GHASH polynomial basis it is that multiply alone. The zerocheck's later rounds already exploit
this through `ReprBackend`, and the binary PCS fold crosses a whole block over once and runs every
round there. The bit ring-switch reduction did not: it ran its sumcheck in the tower basis while
the inner PCS it feeds runs in the polynomial one.

### `perf(sumcheck): run the bit ring-switch rounds in the polynomial basis`

`BitRingSwitch::prove` takes a representation parameter and drives `ReprSumcheckProver`. The
packing crosses over through the bulk `from_tower_vec` kernel, riding the slot copy
`restricted_packing` already made. `weights_over` builds the weight multilinear directly in the
representation field by carrying the `d`-entry batching table over first — a subset sum is
additive and a characteristic-two isomorphism carries addition, so that is 128 conversions rather
than one per entry. `CoordinateSums<EF>` becomes `CoordinateSums<EF, A>`: the level whose bytes
index the coordinates stays in the tower basis, which is what the `F_2` decomposition is read in,
while the weights move.

The tensor elements, the equality table and the initial sum stay in the tower basis: `tensor_over`
and `CoordinateSums::sum` read their inputs' tower bytes, and the elements are wire format.

### `perf(binary-pcs): read the committed packing through a view`

`BitPacking<EF, S = Vec<EF>>` takes a backing store as `Poly<F, S>` does, with
`BitPackingView<'a, EF>`. `BooleanPcs::packing` returns a view over the `PolyView` the table
already hands back instead of collecting the committed column.

### Measurements

branch runs interleaved.

| | `bit ring switch` | `read committed packing` | prove |
| --- | --- | --- | --- |
| BLAKE3 `-l 18` | 441 ms -> 253 ms | 19.3 ms -> 0.52 us | 3.01 s -> 2.80 s (-7.0%) |
| Keccak-f `-l 18` | 79.5 ms -> 46.2 ms | 3.05 ms -> 0.53 us | 797 ms -> 772 ms (-3.1%) |

The rounds alone go from roughly 240 ms to 41.5 ms at BLAKE3 `-l 18`. The view is worth about
50 ms, more than the span accounts for, the remainder being allocation and first-touch page
faults. Peak RSS is unchanged at 10.24 GB: the maximum is set during `prove_with_lookup`, well
before the packing is read.

What remains inside the reduction at BLAKE3 `-l 18`: the equality table 82.7 ms, `tensor_over`
52.0 ms, rounds 41.5 ms, `weights_over` 39.6 ms, `restricted_packing` 29.4 ms. The equality table
is `Poly::new_from_point` materialising one tower multiply per entry and belongs to
`multilinear-util` rather than here.

### Correctness

Prover-only; proofs are byte-identical to #2246 for BLAKE3 and Keccak-f at `-l 12` and `-l 14`.

`the_polynomial_basis_rounds_play_the_tower_basis_transcript` runs a reduction through both
representations over three point shapes — random, Boolean-prefixed so rounds are dropped, and one
sending successor elements — and compares the tensor, the successor elements, the polynomial
evaluations element by element, the grinding witnesses, the final evaluation, the surviving point
and value, and the next sample off both sponges. `a_borrowed_packing_plays_the_owned_packing_transcript`
does the same across the owned and borrowed packings, since production now takes the borrowed path
exclusively.

`cargo test --workspace --features parallel` passes, clippy and rustfmt are clean.

### Review note

Rust has no partial turbofish, so the storage parameter forces `prove::<R, _, _>` and
`restricted_packing::<R, _>` at 21 call sites. That churn is most of the second commit's line
count and none of its content.


Nashtare:
Recreating this against its correct stack position: the branch is now rebased onto `perf/encode-merkle-fusion`, and GitHub will not let an existing stacked PR change its base. Superseded by the replacement opened immediately below; no review had been posted here.

PR #2249, opened 2026-09-20
perf(multi-stark): read the boundary zerocheck round and fold off the planes

Nashtare:
Stacked on #2246. Sibling of #2247, which it does not touch.

After the sliced rounds are spent, `unslice` expanded every residual column to full `R` elements
in one shot, and only then did the boundary round read them and the boundary fold halve them.
For BLAKE3's 11,536 live columns at `-l 20` that intermediate is 5.6 GiB, written once and then
swept twice.

The planes are eight times smaller than the elements they expand to, so this reads them instead.
`round_poly_boundary` builds one word pair's tile, runs its rows against it and drops it;
`fold_boundary` reads the planes per column and writes the half-size folded `Poly<R>` directly.
The full-size array never exists.

`PlaneFold` holds the round-invariant byte-subset tables, flattened to one 256-entry run per
corner group and built once for every task. `unslice` becomes a thin caller of it and stays as
the fallback.

The tile is the load-bearing part: a word pair's corner masks are transposed into per-lane
contiguous runs **in mask-byte space**, four bytes per column-row rather than the sixteen an `R`
element takes, so a tile is 3 MB for BLAKE3 rather than 24 MB. A row-major `R` tile would move
the same bytes through DRAM as the eager route and save only the allocation; eight-row tiles
shrink the tile but read the planes eight times over. Neither is worth it.

`SlicedColumns::at_boundary` guards both kernels: the planes serve the boundary only once the
sliced rounds are spent and there are at least two words of residual rows, and the round also
needs the packing width to divide 64. Every other shape takes the eager route unchanged.

`accumulate_row` and `accumulate_lanes` are split so that the node walk carries the
`#[inline(never)]` and the fill half can inline. Ranking `sub sp, sp, #N` across every symbol in
both binaries gives the same 3840-byte maximum frame, so the AIR frame has not grown — it moved.

### Measurements

c7i.4xlarge (Xeon 8488C, 16 vCPU, 30 GB), `-C target-cpu=native`, `--features parallel`,
branch runs interleaved. The packing width is 4 on this host, so the lane-group path runs; on a
target where it is 1 the scalar row path runs, and the mode test pins the two against each other.

| | prove | `prove_with_lookup` | peak RSS |
| --- | --- | --- | --- |
| BLAKE3 `-l 18` | 3.257 s -> 2.974 s (-8.7%) | 2.04 s -> 1.76 s (-13.7%) | 10.24 GB -> 6.47 GB (-36.8%) |
| BLAKE3 `-l 16` | 0.917 s -> 0.859 s (-6.3%) | 567 ms -> 505 ms | 2.587 GB -> 2.111 GB |
| Keccak-f `-l 18` | 0.892 s -> 0.858 s (-3.8%) | 678 ms -> 642 ms | 1.565 GB -> 1.136 GB (-27.4%) |
| Keccak-f `-l 16` | 0.273 s -> 0.261 s (-4.4%) | 195 ms -> 185 ms | 456 MB -> 336 MB |

No objective regresses at any size. Raising the sliced-round count is the other way to shrink this
intermediate, and it is not equivalent: it helps BLAKE3 and costs Keccak-f badly, which is why
`sliced_rounds` is a configuration knob left at three. This helps both.

At BLAKE3 `-l 20` the base is OOM-killed on a 30 GB host (`anon-rss:31383848kB`). The branch
completes at 26.2 GB peak, proves in 11.725 s and verifies.

### Correctness

Prover-only; proofs are byte-identical to #2246, checked on x86 at `-l 18` for both objectives
(the lane-group path) and on aarch64 at `-l 14` and `-l 16` (the scalar path).

`sliced_rounds` in the tests now takes a boundary mode, and a new test runs both modes and asserts
identical round polynomials, openings and residual-row counts — at a height where the planes serve
the boundary and at one where they cannot. A transpose test covers the lane masks.

`cargo test --workspace --features parallel --release` passes on x86; clippy and rustfmt are clean.

### Not in scope

`SubfieldBackend`'s `unslice_packed` route is untouched and still eager; only `ReprBackend` takes
the lazy path. Extending it needs a packed variant of the tile read.


tcoratger (commented):
Reviewed against the stack base (`perf/binary-prover-followups`), with the changed files read in full and the plane layout reconstructed by hand.

**I could not find a correctness, soundness or memory-safety defect.** The row/lane pairing, the successor (`next`) semantics including the repeat-last tail, the `BoundaryEvals` prefix, the transcript order and the `finish_round` bookkeeping are all preserved exactly, and I verified byte-identical proofs end to end. The findings below are a real test-coverage hole and hot-loop polish.

## What I verified, and how

**End-to-end proof identity.** I patched `SlicedColumns::at_boundary` to `false` (forcing the old eager route) and diffed an FNV digest of the postcard-serialised proof against the unpatched build, on the real `prove_hash_binary` AIRs:

| objective, `-l 14` | eager route | plane route |
| --- | --- | --- |
| blake-3-compressions | `581323 / 4b6ebd3120512de7` | `581323 / 4b6ebd3120512de7` |
| keccak-f-permutations | `336663 / 21e58f59a4dd52f5` | `336663 / 21e58f59a4dd52f5` |

Identical, and identical again with `-C target-feature=+avx2,+vpclmulqdq` (which flips `R::Packing::WIDTH` from 1 to 2 and so switches `boundary_evals_rows` -> `boundary_evals_lanes`; I confirmed the branch is taken by panicking inside it).

**The win is real, not noise.** Span breakdown from one `-l 18` blake3 run each, same binary, same host:

```
eager : unslice 2.68s + round_poly_repr 2.27s + fold_repr 451ms = 5.40s
planes: round_poly_boundary 1.74s + fold_boundary 1.14s        = 2.88s
```

The next round's `round_poly_repr_lanes` is 1.04s vs 1.08s — unchanged, so nothing was pushed downstream.

### The invariants I checked by hand

Residual row `x` after `k` bound challenges maps to the plane array as `cells[(corner * words + word) * width + column]`, `word = x / 64`, `lane = x % 64`, so the original trace row is `corner * 2^(nv-k) + x` — corner-major, exactly the "bound variables are the most significant" convention `fix_prefix_var_mut` assumes. The round pairs `x` with `x + half`, `half = words/2 * 64`, so the two rows sit at the same lane in words `p` and `p + words/2`; `RowTile::fill` uses precisely `[pair, pair + fold.words / ROW_HALVES]`. The `SLICED_LANES + 1`-th lane is the load-bearing trick:

```
        column 0     column 1        ...        column w-1
      +------------+------------+-----------+------------+
lane 0| lo hi | lo hi | ... groups pairs ... |            |   rows 64p+0 and 64p+0+half
lane 1|                                                  |   rows 64p+1 ...
 ...  |                                                  |
lane63|                                                  |   rows 64p+63 ...
      +--------------------------------------------------+
lane64| successor planes, top lane only, next_columns only|   the row after lane 63
      +--------------------------------------------------+
        <- column_stride = 2*groups*PLANE_BYTES bytes ->
        <------------- lane_stride = column_stride * width ------------->
```

A row's `next` value is the *following* lane's bytes. For `lane < 63` that is `cells[word][lane+1]`, which equals `column[x+1]` after folding because `successor_cell(b, x) = cell(b, x+1)` whenever `x+1` stays inside the block. For `lane == 63` it is the extra lane, read from `trace.successors`, whose fold at the global last word's top lane is bit-for-bit the `next_tail` that `unslice` would have written — exactly the value `accumulate_row`/`accumulate_lanes` substitute when `s + half + 1 == num_evals`. The index arithmetic tops out at `65 * lane_stride - 1`, one byte short of `bytes.len()`, for both `write_lane_masks` and `write_top_lane_mask`.

Two smaller things that had to line up and do: `fold_boundary`'s tail is `lo + (tail - lo) * r` with `lo = fold_word(cells, c, words/2)[0]` — residual row `half`, the same row `fold_claims_and_tails` reads; and `corners = 2^rounds <= 2^MAX_SLICED_ROUNDS = MAX_CORNERS`, so `corner_words`' fixed `[u64; 16]` can never be indexed out of range.

The dispatch ordering at `backend.rs:311` is sound, and I checked the state machine rather than assuming it: `fold_boundary` is tried before `fold_sliced`, but `at_boundary()` requires `challenges.len() == trace.rounds` while `fold_sliced`'s precondition is `challenges.len() < trace.rounds`, so the two guards are disjoint. Every path that makes `round_poly_boundary` return `None` also runs `unslice`, which leaves `ExtColumns::Scalar` and makes both sliced folds decline. `fold_sliced`'s `debug_assert!` cannot be reached with the planes still in place.

Finally, everything sums in a characteristic-2 field, so `par_fold_reduce` is exactly associative and the result is independent of how Rayon splits — the byte-identity claim is not an "in practice" claim, it is forced.

---

## [MAJOR] The equivalence test only ever exercises a single word pair — `multi-stark/src/rounds/sliced/tests.rs:422`

```rust
for (height, on_planes) in [(8 * SHORTEST, true), (4 * SHORTEST, false)] {
```

`SHORTEST = 128`, so the "planes serve" case is `height = 1024`, `num_vars = 10`, `rounds = 3`, `num_evals = 128`, `words = 2`, and therefore `pairs = fold.words / ROW_HALVES == 1`.

Three distinct behaviours are invisible at that size:

1. **The tile is never reused.** `RowTile::fill` overwrites the main lanes for every column and the extra lane only for `next_columns`; correctness across pairs depends on that covering every byte a later `read_row` touches. With one pair the tile is only ever read right after being built from a zeroed allocation.
2. **The `par_fold_reduce` reduce closure never runs.** With one item Rayon never joins, so the `constraint_evals`/`interaction_evals` merge in `boundary_evals_rows`/`boundary_evals_lanes` is dead code in the test.
3. **The high half's successor word is always the last word.** At `pairs == 1` the high word is `words - 1`, so `write_top_lane_mask(&successors, column, pair + words/2, ..)` and `write_top_lane_mask(&successors, column, words - 1, ..)` are the same call. An off-by-one there — the single most likely bug in this patch — passes the test.

I ran the extended loop locally (`16 * SHORTEST`, `32 * SHORTEST`, `64 * SHORTEST` added, `pairs` = 2, 4, 8) under both `R = Tower` and `R = Ghash128` and both packing widths: all pass. So this is a hole in the net, not a live bug.

**Fix:** extend the array to include at least `(16 * SHORTEST, true)`; `(32 * SHORTEST, true)` costs little more and gives two levels of Rayon join.

**Verify:** with the test widened, mutate `RowTile::fill`'s `let words = [pair, pair + fold.words / ROW_HALVES];` to `[pair, fold.words - 1]`. It must fail at `16 * SHORTEST` and still pass at `8 * SHORTEST`.

---

## [MAJOR] No CI job ever executes the new lane-group kernel

`boundary_evals_lanes` + `RowTile::read_lane_group` (`multi-stark/src/rounds/sliced.rs:1489` and `:1160`) run only when `R::Packing::WIDTH > 1`. The only `R` in use is `Ghash128`, whose `type Packing` is `crate::packed::Packing`, and `binary-field/src/packed/mod.rs` gates the wide packing on `target_feature = "vpclmulqdq"` *and* (`avx2` or `avx512f`). Every CI leg carrying `vpclmulqdq` is declared **build only**:

```
Build and Test (linux, avx2+vpclmulqdq, build only)
Build and Test (linux, avx512f+vpclmulqdq, build only)
```

`(linux, pclmulqdq)` is the scalar carryless multiply, not `vpclmulqdq`, so it does not enable the packing either. I confirmed the split empirically: default build takes `boundary_evals_rows`; with `-C target-feature=+avx2,+vpclmulqdq` the lane branch is taken at `WIDTH = 2`.

This is pre-existing for `round_poly_repr_lanes`, but it matters more here: the lane kernel is the one the PR's own measurements are taken on, and its indexing (`self.cell(lane + step + 1, column)` reaching into the 65th lane) is the subtlest code in the patch. The PR body's "the mode test pins the two against each other" is true only on a host that has the packing — CI is green on this PR, but green CI is not evidence for this kernel.

**Fix:** flip one of the `+vpclmulqdq` legs from build-only to a real test run (the `p3-multi-stark` unit tests alone would do), or note in the PR that the lane path rests on the author's local x86 run.

**Verify:** `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" cargo test -p p3-multi-stark --release --features parallel --lib rounds::sliced`.

---

## [MINOR] The table lookup in the innermost loop keeps a bounds check — `multi-stark/src/rounds/sliced.rs:988` and `:1001`

```rust
*value += self.low_sums[table + usize::from(low)] + self.high_sums[table + usize::from(high)];
```

`low_sums` is a `Vec<R>` of length `groups * GROUP_ENTRIES`, and `table = group * GROUP_ENTRIES` is derived from `self.groups`. LLVM has no relation between the field `self.groups` and `self.low_sums.len()` — they are independent loads from the same struct — so it cannot prove `table + mask < len` and must keep the check. That is two bounds checks per corner group per plane per column-row, in the two functions that my span measurement shows are 18.7% of a `-l 18` blake3 prove (`round_poly_boundary` 1.74s + `fold_boundary` 1.14s of 15.4s).

Making the table a `Vec<[R; GROUP_ENTRIES]>` moves the byte index onto a fixed-size array, where `usize::from(u8) <= 255 < 256` is provable and the check folds away; only the per-group outer index survives, once per group rather than twice per lookup.

```rust
low_sums: Vec<[R; GROUP_ENTRIES]>,   // built with vec![[R::ZERO; GROUP_ENTRIES]; groups]
...
let (lo_table, hi_table) = (&self.low_sums[group], &self.high_sums[group]);
*value += lo_table[usize::from(low)] + hi_table[usize::from(high)];
```

`row_value` then becomes `for (group, masks) in bytes.as_chunks::<PLANE_BYTES>().0.iter().enumerate()`.

**Verify:** I did not measure this — my host was too loaded to resolve a 1-2% delta (`-l 18` wall time ranged 14.9–18.5s across three runs). Worth a criterion bench on `PlaneFold::fold_word` alone rather than an end-to-end timing.

---

## [MINOR] `read_lane_group` slices each tile cell twice per lane — `multi-stark/src/rounds/sliced.rs:1174`

```rust
let lo = lane_group(|step| fold.row_value(&self.cell(lane + step, column)[..half]));
let hi = lane_group(|step| fold.row_value(&self.cell(lane + step, column)[half..]));
```

`lane_group` is `R::Packing::from_fn`, so the closure runs `WIDTH` times; writing it twice evaluates `self.cell(..)` — two bounds-checked subslices, `lane(..)` then the column offset — `2 * WIDTH` times per column instead of `WIDTH`. The scalar twin gets this right: `read_row` calls `cell` once and lets `row_pair` split the halves.

**Fix:** fill both packings in one pass over the lanes, e.g.

```rust
let (mut lo, mut hi) = (R::Packing::ZERO, R::Packing::ZERO);
for (step, (l, h)) in lo.as_slice_mut().iter_mut().zip(hi.as_slice_mut()).enumerate() {
    let (a, b) = fold.row_pair(self.cell(lane + step, column));
    (*l, *h) = (a, b);
}
(*local, *local_delta) = (PackedExt::new(lo), PackedExt::new(hi) - PackedExt::new(lo));
```

**Verify:** the existing equivalence test, run under `+avx2,+vpclmulqdq`, covers this refactor exactly.

---

## [MINOR] `RowTile`'s stride and `PlaneFold`'s group count are coupled but never checked — `multi-stark/src/rounds/sliced.rs:1069`

`RowTile::new(groups, width)` bakes `groups` into `column_stride`, and then `fill`/`read_row`/`read_lane_group` all re-derive the layout from `fold.groups` and `fold.trace.width`. Both call sites happen to pass matching values, but nothing enforces it; a mismatch would not panic, it would silently read the wrong mask bytes and produce a wrong-but-well-formed round polynomial. That is the worst failure mode for a prover: the proof simply stops verifying, with no pointer to the cause.

**Fix:** `debug_assert_eq!(self.column_stride, ROW_HALVES * fold.groups * PLANE_BYTES)` and `debug_assert_eq!(self.lane_stride, self.column_stride * fold.trace.width)` at the top of `fill`. Or store the `&PlaneFold` in the tile and drop the separate `groups`/`width` arguments.

---

## [MINOR] `row_value` infers the group count from the slice length — `multi-stark/src/rounds/sliced.rs:995`

```rust
for masks in bytes.as_chunks::<PLANE_BYTES>().0 { ... }
```

`fold_word` iterates `0..self.groups`; `row_value` iterates however many pairs the caller's slice happens to hold, and `as_chunks` silently drops an odd trailing byte. The two must agree for the round and the fold to produce the same values, which is the entire premise of the patch. A `debug_assert_eq!(bytes.len(), self.groups * PLANE_BYTES)` costs nothing and pins the contract.

---

## [NIT]

- **Four copies of the same degree-vector and reduce boilerplate.** `constraint_degrees` / `interaction_degrees` are rebuilt identically in `round_poly_unpacked`, `round_poly_repr_lanes`, `boundary_evals_rows` and `boundary_evals_lanes`, and the `par_fold_reduce` merge closure now appears four times too (twice with `R::add_slices`, twice with `add_slice`). Two small helpers on `RoundStateExt` would remove ~60 lines and make the next kernel harder to get subtly out of step.
- `top_lane_mask` (`:843`) opens with `lane_masks`' sentence verbatim — "For each lane, the byte whose bit `i` is…" — but returns one byte, for one lane. Lead with the second paragraph instead.
- `group_words` (`:962`) says "one array per plane"; it returns slices.
- `RowTile::new` (`:1068`) says "An empty tile"; it is zero-filled and full length.
- `lane_masks`' `debug_assert!(words.len() <= 8)` (`:829`) should read `GROUP_CORNERS` now that the constant exists.
- `SlicedTrace::cells` (`:58`) is documented as `cells[w * width + c]`, which omits the corner dimension that `corner_words` strides over. Pre-existing, but this PR is the third reader of that layout; spelling it out would pay for itself.

---

## Test coverage

| Gap | Concrete test |
| --- | --- |
| `pairs > 1` (tile reuse, Rayon join, non-final high word) | Add `(16 * SHORTEST, true)` and `(32 * SHORTEST, true)` to the loop at `tests.rs:422`. Verified passing locally. |
| Lane-group kernel unexecuted in CI | Promote one `+vpclmulqdq` CI leg from build-only to a test run, or at minimum run `p3-multi-stark`'s lib tests there. |
| `RowTile` transpose in isolation | A proptest is the right tool here, not vectors: the property is uniform over a large input space and the oracle is already in the tree. Inputs: `groups in 1..=2`, `width in 1..=8`, `words in 2..=8`, `corners in {2,4,8,16}`, random `[u64; 2]` plane arrays and random challenges. Invariant: for every `(column, lane)`, `RowTile::fill(pair).read_row(..)` must reproduce `fold_word(cells, column, pair)[lane]` for the low half and `fold_word(cells, column, pair + words/2)[lane]` for the high half, and the extra lane must reproduce `fold_word(successors, column, word)[63]`. A failure pins the bug to the transpose rather than to the AIR walk, which the current end-to-end test cannot do. |
| `next_row_runs` empty | No fixture AIR without next-row columns is run through the boundary path; `RowTile::fill` then skips the extra lane entirely and `read_row`'s second loop is empty. Cheap to add as a third instance. |

---

## Could not verify

- **The stack-frame claim.** The PR states that ranking `sub sp, sp, #N` across both binaries gives the same 3840-byte maximum after moving `#[inline(never)]` from `accumulate_row`/`accumulate_lanes` onto `walk_row_nodes`/`walk_lane_nodes`. I did not reproduce that measurement. I did run `-l 18` blake3 with 32 Rayon workers without incident, and the split is structurally right — the AIR-heavy node walk keeps the attribute, the column-read half does not touch `next_tail` and so can be replaced wholesale by the tile read. That last point is what makes the whole design work, and it is worth calling out in the code: `walk_row_nodes` is correct for the plane path *because* it reads nothing but `scratch`, `self.boundary` and `self.slots`.
- **The published benchmark numbers** are from a c7i.4xlarge; my host was busy enough that only the intra-run span breakdown was trustworthy. That breakdown supports the direction and rough magnitude (5.40s -> 2.88s at `-l 18`), not the exact percentages.
- **`SubfieldBackend`'s `unslice_packed`** is genuinely untouched — its `R = EF` and its `fold` still goes `fold_sliced` then generic. Nothing in this patch makes extending it harder.

---

**Bottom line:** merge after widening the equivalence test to at least `16 * SHORTEST` (one line), and decide explicitly whether the lane kernel gets a CI leg that runs it. The algorithm and its implementation are correct, and I have byte-identical proofs on two real AIRs and both kernels to back that.

multi-stark/src/rounds/sliced/tests.rs
```diff
@@ -381,6 +416,46 @@ fn every_round_on_and_off_the_planes_matches_the_generic_kernel() {
     }
 }
 
+#[test]
+fn a_boundary_round_and_fold_on_the_planes_match_the_unsliced_kernels() {
+    // A word pair of residual rows is the shortest the planes can serve a boundary round.
+    for (height, on_planes) in [(8 * SHORTEST, true), (4 * SHORTEST, false)] {
```
tcoratger: **[MAJOR] The equivalence test only ever exercises a single word pair — `multi-stark/src/rounds/sliced/tests.rs:422`**

```rust
for (height, on_planes) in [(8 * SHORTEST, true), (4 * SHORTEST, false)] {
```

`SHORTEST = 128`, so the "planes serve" case is `height = 1024`, `num_vars = 10`, `rounds = 3`, `num_evals = 128`, `words = 2`, and therefore `pairs = fold.words / ROW_HALVES == 1`.

Three distinct behaviours are invisible at that size:

1. **The tile is never reused.** `RowTile::fill` overwrites the main lanes for every column and the extra lane only for `next_columns`; correctness across pairs depends on that covering every byte a later `read_row` touches. With one pair the tile is only ever read right after being built from a zeroed allocation.
2. **The `par_fold_reduce` reduce closure never runs.** With one item Rayon never joins, so the `constraint_evals`/`interaction_evals` merge in `boundary_evals_rows`/`boundary_evals_lanes` is dead code in the test.
3. **The high half's successor word is always the last word.** At `pairs == 1` the high word is `words - 1`, so `write_top_lane_mask(&successors, column, pair + words/2, ..)` and `write_top_lane_mask(&successors, column, words - 1, ..)` are the same call. An off-by-one there — the single most likely bug in this patch — passes the test.

I ran the extended loop locally (`16 * SHORTEST`, `32 * SHORTEST`, `64 * SHORTEST` added, `pairs` = 2, 4, 8) under both `R = Tower` and `R = Ghash128` and both packing widths: all pass. So this is a hole in the net, not a live bug.

**Fix:** extend the array to include at least `(16 * SHORTEST, true)`; `(32 * SHORTEST, true)` costs little more and gives two levels of Rayon join.

**Verify:** with the test widened, mutate `RowTile::fill`'s `let words = [pair, pair + fold.words / ROW_HALVES];` to `[pair, fold.words - 1]`. It must fail at `16 * SHORTEST` and still pass at `8 * SHORTEST`.

multi-stark/src/rounds/sliced.rs
```diff
@@ -835,6 +857,351 @@ fn subset_sums<R: Field>(weights: &[R]) -> Vec<R> {
...
+    fn fold_word(&self, planes: &[[u64; 2]], column: usize, word: usize, out: &mut [R]) {
+        let (low, high) = self.corner_words(planes, column, word);
+        out.fill(R::ZERO);
+        for group in 0..self.groups {
+            let (low, high) = self.group_words(&low, &high, group);
+            let (low, high) = (lane_masks(low), lane_masks(high));
+            let table = group * GROUP_ENTRIES;
+            for (value, (&low, &high)) in out.iter_mut().zip(low.iter().zip(&high)) {
+                *value += self.low_sums[table + usize::from(low)]
+                    + self.high_sums[table + usize::from(high)];
```
tcoratger: **[MINOR] The table lookup in the innermost loop keeps a bounds check — `multi-stark/src/rounds/sliced.rs:988` and `:1001`**

```rust
*value += self.low_sums[table + usize::from(low)] + self.high_sums[table + usize::from(high)];
```

`low_sums` is a `Vec<R>` of length `groups * GROUP_ENTRIES`, and `table = group * GROUP_ENTRIES` is derived from `self.groups`. LLVM has no relation between the field `self.groups` and `self.low_sums.len()` — they are independent loads from the same struct — so it cannot prove `table + mask < len` and must keep the check. That is two bounds checks per corner group per plane per column-row, in the two functions that my span measurement shows are 18.7% of a `-l 18` blake3 prove (`round_poly_boundary` 1.74s + `fold_boundary` 1.14s of 15.4s).

Making the table a `Vec<[R; GROUP_ENTRIES]>` moves the byte index onto a fixed-size array, where `usize::from(u8) <= 255 < 256` is provable and the check folds away; only the per-group outer index survives, once per group rather than twice per lookup.

```rust
low_sums: Vec<[R; GROUP_ENTRIES]>,   // built with vec![[R::ZERO; GROUP_ENTRIES]; groups]
...
let (lo_table, hi_table) = (&self.low_sums[group], &self.high_sums[group]);
*value += lo_table[usize::from(low)] + hi_table[usize::from(high)];
```

`row_value` then becomes `for (group, masks) in bytes.as_chunks::<PLANE_BYTES>().0.iter().enumerate()`.

**Verify:** I did not measure this — my host was too loaded to resolve a 1-2% delta (`-l 18` wall time ranged 14.9–18.5s across three runs). Worth a criterion bench on `PlaneFold::fold_word` alone rather than an end-to-end timing.

multi-stark/src/rounds/sliced.rs
```diff
@@ -835,6 +857,351 @@ fn subset_sums<R: Field>(weights: &[R]) -> Vec<R> {
...
+        fold: &PlaneFold<'_, R>,
+        lane: usize,
+        next_columns: &[Range<usize>],
+        scratch: &mut PackedScratch<PackedRepr<F, R>, PackedRepr<F, R>>,
+    ) {
+        let half = fold.groups * PLANE_BYTES;
+        let PackedScratch {
+            local_point,
+            local_diff,
+            next_point,
```
tcoratger, line 1210: **[MINOR] `read_lane_group` slices each tile cell twice per lane — `multi-stark/src/rounds/sliced.rs:1174`**

```rust
let lo = lane_group(|step| fold.row_value(&self.cell(lane + step, column)[..half]));
let hi = lane_group(|step| fold.row_value(&self.cell(lane + step, column)[half..]));
```

`lane_group` is `R::Packing::from_fn`, so the closure runs `WIDTH` times; writing it twice evaluates `self.cell(..)` — two bounds-checked subslices, `lane(..)` then the column offset — `2 * WIDTH` times per column instead of `WIDTH`. The scalar twin gets this right: `read_row` calls `cell` once and lets `row_pair` split the halves.

**Fix:** fill both packings in one pass over the lanes, e.g.

```rust
let (mut lo, mut hi) = (R::Packing::ZERO, R::Packing::ZERO);
for (step, (l, h)) in lo.as_slice_mut().iter_mut().zip(hi.as_slice_mut()).enumerate() {
    let (a, b) = fold.row_pair(self.cell(lane + step, column));
    (*l, *h) = (a, b);
}
(*local, *local_delta) = (PackedExt::new(lo), PackedExt::new(hi) - PackedExt::new(lo));
```

**Verify:** the existing equivalence test, run under `+avx2,+vpclmulqdq`, covers this refactor exactly.

tcoratger (commented):

---

Reviewed against the stack base (`perf/binary-prover-followups`), with the changed files read in full and the plane layout reconstructed by hand.

**I could not find a correctness, soundness or memory-safety defect.** The row/lane pairing, the successor (`next`) semantics including the repeat-last tail, the `BoundaryEvals` prefix, the transcript order and the `finish_round` bookkeeping are all preserved exactly, and I verified byte-identical proofs end to end. The findings below are a real test-coverage hole and hot-loop polish.

---

## What I verified, and how

**End-to-end proof identity.** I patched `SlicedColumns::at_boundary` to `false` (forcing the old eager route) and diffed an FNV digest of the postcard-serialised proof against the unpatched build, on the real `prove_hash_binary` AIRs:

| objective, `-l 14` | eager route | plane route |
| --- | --- | --- |
| blake-3-compressions | `581323 / 4b6ebd3120512de7` | `581323 / 4b6ebd3120512de7` |
| keccak-f-permutations | `336663 / 21e58f59a4dd52f5` | `336663 / 21e58f59a4dd52f5` |

Identical, and identical again with `-C target-feature=+avx2,+vpclmulqdq` (which flips `R::Packing::WIDTH` from 1 to 2 and so switches `boundary_evals_rows` -> `boundary_evals_lanes`; I confirmed the branch is taken by panicking inside it).

**The win is real, not noise.** Span breakdown from one `-l 18` blake3 run each, same binary, same host:

```
eager : unslice 2.68s + round_poly_repr 2.27s + fold_repr 451ms = 5.40s
planes: round_poly_boundary 1.74s + fold_boundary 1.14s        = 2.88s
```

The next round's `round_poly_repr_lanes` is 1.04s vs 1.08s — unchanged, so nothing was pushed downstream.

### The invariants I checked by hand

Residual row `x` after `k` bound challenges maps to the plane array as `cells[(corner * words + word) * width + column]`, `word = x / 64`, `lane = x % 64`, so the original trace row is `corner * 2^(nv-k) + x` — corner-major, exactly the "bound variables are the most significant" convention `fix_prefix_var_mut` assumes. The round pairs `x` with `x + half`, `half = words/2 * 64`, so the two rows sit at the same lane in words `p` and `p + words/2`; `RowTile::fill` uses precisely `[pair, pair + fold.words / ROW_HALVES]`. The `SLICED_LANES + 1`-th lane is the load-bearing trick:

```
        column 0     column 1        ...        column w-1
      +------------+------------+-----------+------------+
lane 0| lo hi | lo hi | ... groups pairs ... |            |   rows 64p+0 and 64p+0+half
lane 1|                                                  |   rows 64p+1 ...
 ...  |                                                  |
lane63|                                                  |   rows 64p+63 ...
      +--------------------------------------------------+
lane64| successor planes, top lane only, next_columns only|   the row after lane 63
      +--------------------------------------------------+
        <- column_stride = 2*groups*PLANE_BYTES bytes ->
        <------------- lane_stride = column_stride * width ------------->
```

A row's `next` value is the *following* lane's bytes. For `lane < 63` that is `cells[word][lane+1]`, which equals `column[x+1]` after folding because `successor_cell(b, x) = cell(b, x+1)` whenever `x+1` stays inside the block. For `lane == 63` it is the extra lane, read from `trace.successors`, whose fold at the global last word's top lane is bit-for-bit the `next_tail` that `unslice` would have written — exactly the value `accumulate_row`/`accumulate_lanes` substitute when `s + half + 1 == num_evals`. The index arithmetic tops out at `65 * lane_stride - 1`, one byte short of `bytes.len()`, for both `write_lane_masks` and `write_top_lane_mask`.

Two smaller things that had to line up and do: `fold_boundary`'s tail is `lo + (tail - lo) * r` with `lo = fold_word(cells, c, words/2)[0]` — residual row `half`, the same row `fold_claims_and_tails` reads; and `corners = 2^rounds <= 2^MAX_SLICED_ROUNDS = MAX_CORNERS`, so `corner_words`' fixed `[u64; 16]` can never be indexed out of range.

The dispatch ordering at `backend.rs:311` is sound, and I checked the state machine rather than assuming it: `fold_boundary` is tried before `fold_sliced`, but `at_boundary()` requires `challenges.len() == trace.rounds` while `fold_sliced`'s precondition is `challenges.len() < trace.rounds`, so the two guards are disjoint. Every path that makes `round_poly_boundary` return `None` also runs `unslice`, which leaves `ExtColumns::Scalar` and makes both sliced folds decline. `fold_sliced`'s `debug_assert!` cannot be reached with the planes still in place.

Finally, everything sums in a characteristic-2 field, so `par_fold_reduce` is exactly associative and the result is independent of how Rayon splits — the byte-identity claim is not an "in practice" claim, it is forced.

---

### [MAJOR] No CI job ever executes the new lane-group kernel

`boundary_evals_lanes` + `RowTile::read_lane_group` (`multi-stark/src/rounds/sliced.rs:1489` and `:1160`) run only when `R::Packing::WIDTH > 1`. The only `R` in use is `Ghash128`, whose `type Packing` is `crate::packed::Packing`, and `binary-field/src/packed/mod.rs` gates the wide packing on `target_feature = "vpclmulqdq"` *and* (`avx2` or `avx512f`). Every CI leg carrying `vpclmulqdq` is declared **build only**:

```
Build and Test (linux, avx2+vpclmulqdq, build only)
Build and Test (linux, avx512f+vpclmulqdq, build only)
```

`(linux, pclmulqdq)` is the scalar carryless multiply, not `vpclmulqdq`, so it does not enable the packing either. I confirmed the split empirically: default build takes `boundary_evals_rows`; with `-C target-feature=+avx2,+vpclmulqdq` the lane branch is taken at `WIDTH = 2`.

This is pre-existing for `round_poly_repr_lanes`, but it matters more here: the lane kernel is the one the PR's own measurements are taken on, and its indexing (`self.cell(lane + step + 1, column)` reaching into the 65th lane) is the subtlest code in the patch. The PR body's "the mode test pins the two against each other" is true only on a host that has the packing — CI is green on this PR, but green CI is not evidence for this kernel.

**Fix:** flip one of the `+vpclmulqdq` legs from build-only to a real test run (the `p3-multi-stark` unit tests alone would do), or note in the PR that the lane path rests on the author's local x86 run.

**Verify:** `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" cargo test -p p3-multi-stark --release --features parallel --lib rounds::sliced`.

### [NIT]

- **Four copies of the same degree-vector and reduce boilerplate.** `constraint_degrees` / `interaction_degrees` are rebuilt identically in `round_poly_unpacked`, `round_poly_repr_lanes`, `boundary_evals_rows` and `boundary_evals_lanes`, and the `par_fold_reduce` merge closure now appears four times too (twice with `R::add_slices`, twice with `add_slice`). Two small helpers on `RoundStateExt` would remove ~60 lines and make the next kernel harder to get subtly out of step.
- `top_lane_mask` (`:843`) opens with `lane_masks`' sentence verbatim — "For each lane, the byte whose bit `i` is…" — but returns one byte, for one lane. Lead with the second paragraph instead.
- `group_words` (`:962`) says "one array per plane"; it returns slices.
- `RowTile::new` (`:1068`) says "An empty tile"; it is zero-filled and full length.
- `lane_masks`' `debug_assert!(words.len() <= 8)` (`:829`) should read `GROUP_CORNERS` now that the constant exists.
- `SlicedTrace::cells` (`:58`) is documented as `cells[w * width + c]`, which omits the corner dimension that `corner_words` strides over. Pre-existing, but this PR is the third reader of that layout; spelling it out would pay for itself.

---

## Test coverage

| Gap | Concrete test |
| --- | --- |
| `pairs > 1` (tile reuse, Rayon join, non-final high word) | Add `(16 * SHORTEST, true)` and `(32 * SHORTEST, true)` to the loop at `tests.rs:422`. Verified passing locally. |
| Lane-group kernel unexecuted in CI | Promote one `+vpclmulqdq` CI leg from build-only to a test run, or at minimum run `p3-multi-stark`'s lib tests there. |
| `RowTile` transpose in isolation | A proptest is the right tool here, not vectors: the property is uniform over a large input space and the oracle is already in the tree. Inputs: `groups in 1..=2`, `width in 1..=8`, `words in 2..=8`, `corners in {2,4,8,16}`, random `[u64; 2]` plane arrays and random challenges. Invariant: for every `(column, lane)`, `RowTile::fill(pair).read_row(..)` must reproduce `fold_word(cells, column, pair)[lane]` for the low half and `fold_word(cells, column, pair + words/2)[lane]` for the high half, and the extra lane must reproduce `fold_word(successors, column, word)[63]`. A failure pins the bug to the transpose rather than to the AIR walk, which the current end-to-end test cannot do. |
| `next_row_runs` empty | No fixture AIR without next-row columns is run through the boundary path; `RowTile::fill` then skips the extra lane entirely and `read_row`'s second loop is empty. Cheap to add as a third instance. |

---

## Could not verify

- **The stack-frame claim.** The PR states that ranking `sub sp, sp, #N` across both binaries gives the same 3840-byte maximum after moving `#[inline(never)]` from `accumulate_row`/`accumulate_lanes` onto `walk_row_nodes`/`walk_lane_nodes`. I did not reproduce that measurement. I did run `-l 18` blake3 with 32 Rayon workers without incident, and the split is structurally right — the AIR-heavy node walk keeps the attribute, the column-read half does not touch `next_tail` and so can be replaced wholesale by the tile read. That last point is what makes the whole design work, and it is worth calling out in the code: `walk_row_nodes` is correct for the plane path *because* it reads nothing but `scratch`, `self.boundary` and `self.slots`.
- **The published benchmark numbers** are from a c7i.4xlarge; my host was busy enough that only the intra-run span breakdown was trustworthy. That breakdown supports the direction and rough magnitude (5.40s -> 2.88s at `-l 18`), not the exact percentages.
- **`SubfieldBackend`'s `unslice_packed`** is genuinely untouched — its `R = EF` and its `fold` still goes `fold_sliced` then generic. Nothing in this patch makes extending it harder.

---

**Bottom line:** merge after widening the equivalence test to at least `16 * SHORTEST` (one line), and decide explicitly whether the lane kernel gets a CI leg that runs it. The algorithm and its implementation are correct, and I have byte-identical proofs on two real AIRs and both kernels to back that.

multi-stark/src/rounds/sliced.rs
```diff
@@ -835,6 +857,351 @@ fn subset_sums<R: Field>(weights: &[R]) -> Vec<R> {
...
+}
+
+impl RowTile {
+    /// An empty tile for a stage of `width` columns whose rows read `groups` corner groups.
+    fn new(groups: usize, width: usize) -> Self {
+        let column_stride = ROW_HALVES * groups * PLANE_BYTES;
+        let lane_stride = column_stride * width;
+        Self {
+            bytes: vec![0; (SLICED_LANES + 1) * lane_stride],
+            column_stride,
```
tcoratger, line 1080: **[MINOR] `RowTile`'s stride and `PlaneFold`'s group count are coupled but never checked — `multi-stark/src/rounds/sliced.rs:1069`**

`RowTile::new(groups, width)` bakes `groups` into `column_stride`, and then `fill`/`read_row`/`read_lane_group` all re-derive the layout from `fold.groups` and `fold.trace.width`. Both call sites happen to pass matching values, but nothing enforces it; a mismatch would not panic, it would silently read the wrong mask bytes and produce a wrong-but-well-formed round polynomial. That is the worst failure mode for a prover: the proof simply stops verifying, with no pointer to the cause.

**Fix:** `debug_assert_eq!(self.column_stride, ROW_HALVES * fold.groups * PLANE_BYTES)` and `debug_assert_eq!(self.lane_stride, self.column_stride * fold.trace.width)` at the top of `fill`. Or store the `&PlaneFold` in the tile and drop the separate `groups`/`width` arguments.

multi-stark/src/rounds/sliced.rs
```diff
@@ -835,6 +857,351 @@ fn subset_sums<R: Field>(weights: &[R]) -> Vec<R> {
...
+        }
+    }
+
+    /// The value one half of one residual row's mask bytes stands for.
+    #[inline]
+    fn row_value(&self, bytes: &[u8]) -> R {
+        let mut value = R::ZERO;
+        let mut table = 0;
+        for masks in bytes.as_chunks::<PLANE_BYTES>().0 {
+            value += self.low_sums[table + usize::from(masks[0])]
```
tcoratger: **[MINOR] `row_value` infers the group count from the slice length — `multi-stark/src/rounds/sliced.rs:995`**

```rust
for masks in bytes.as_chunks::<PLANE_BYTES>().0 { ... }
```

`fold_word` iterates `0..self.groups`; `row_value` iterates however many pairs the caller's slice happens to hold, and `as_chunks` silently drops an odd trailing byte. The two must agree for the round and the fold to produce the same values, which is the entire premise of the patch. A `debug_assert_eq!(bytes.len(), self.groups * PLANE_BYTES)` costs nothing and pins the contract.

multi-stark/src/rounds/sliced/tests.rs
Nashtare: Widened to `(32 * SHORTEST, true), (16 * SHORTEST, true), (8 * SHORTEST, true), (4 * SHORTEST, false)` in 8035b12e5 — four, two and one word pair plus the off-the-planes case.

I ran your mutation. With `RowTile::fill`'s `let words = [pair, pair + fold.words / ROW_HALVES];` changed to `[pair, fold.words - 1]`, the widened test fails at 32 * SHORTEST (`4096 rows, tower`) and at 16 * SHORTEST (`2048 rows, tower`), and still passes at `[(8 * SHORTEST, true), (4 * SHORTEST, false)]` alone. Mutation reverted.

One thing your note did not cover: on aarch64 `Ghash128::Packing = Ghash128`, so `R::Packing::WIDTH == 1` and `boundary_evals_lanes` never runs locally — the widened test only reaches `read_lane_group` on a `vpclmulqdq` target. I checked that leg on a Sapphire Rapids box under `-C target-cpu=native`: changing `lane_pair`'s `self.cell(lane + step, column)` to `self.cell(lane, column)` fails the same test there, so the packed reader is genuinely covered.

multi-stark/src/rounds/sliced.rs
Nashtare: Measured, and it holds — landed as cfb3f4f15. `low_sums`/`high_sums` are now `Vec<[R; GROUP_ENTRIES]>`, `fold_word` hoists `&self.low_sums[group]` out of the lane loop, and `row_value` zips the tables with the mask chunks.

A/B on a quiet c7i.4xlarge (Xeon 8488C, `-C target-cpu=native`, `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`), base = this branch plus the other three review fixes, work = that plus this one, interleaved, 5 reps each, `RUST_LOG=info,p3_multi_stark=debug` so the two spans are visible. Medians, with the full range in brackets:

| | span | base | work | delta |
| --- | --- | ---: | ---: | ---: |
| blake3 `-l 18` | `round_poly_boundary` | 319 ms [314-324] | 301 ms [298-305] | −5.6% |
| | `fold_boundary` | 198 ms [196-199] | 195 ms [194-195] | −1.5% |
| | prove | 2.610 s [2.601-2.620] | 2.584 s [2.579-2.598] | −1.0% |
| keccak `-l 18` | `round_poly_boundary` | 111 ms [109-111] | 104 ms [103-105] | −6.3% |
| | `fold_boundary` | 30.0 ms [29.8-30.1] | 28.8 ms [28.7-29.8] | −4.0% |
| | prove | 0.733 s [0.732-0.742] | 0.727 s [0.725-0.736] | −0.8% |

The two span distributions do not overlap on either objective. Proof size is unchanged (993065 / 661898 bytes) and both still verify.

I skipped the criterion bench: the production shape has `groups == 1` (`sliced_rounds = 3`, so 8 corners), and a microbenchmark would run the tables against a cache-resident working set, which overstates a bounds-check removal in a pass that is bandwidth-bound at scale. The debug spans give the same isolation on the real shape. Your end-to-end wall time could not resolve this because it is ~1% of prove; the spans it sits in are 15-20% and resolve it fine.

multi-stark/src/rounds/sliced.rs
Nashtare: Confirmed and fixed in 113214144. `read_lane_group` now calls a `RowTile::lane_pair` helper that walks the lanes once, taking both halves of each cell from `fold.row_pair`, exactly as `read_row` does; the two `lane_group` calls are gone from both the local and the successor loop.

Verified on a `vpclmulqdq` target as you suggested (`-C target-cpu=native`, Sapphire Rapids): `rounds::sliced` passes, and the widened equivalence test fails there if `lane_pair` reads `self.cell(lane, column)` instead of `self.cell(lane + step, column)`, so the refactored path is the one under test. It is invisible on aarch64, where `Ghash128::Packing = Ghash128` and `boundary_evals_lanes` never runs.

multi-stark/src/rounds/sliced.rs
Nashtare: Added both asserts at the top of `RowTile::fill` in bbb9463ab:

```rust
debug_assert_eq!(self.column_stride, ROW_HALVES * fold.groups * PLANE_BYTES);
debug_assert_eq!(self.lane_stride, self.column_stride * fold.trace.width);
```

I kept the separate `groups`/`width` arguments rather than storing the `&PlaneFold`: the tile outlives the fold reference across `par_fold_reduce`'s identity closure, so holding it would put a borrow of the fold into the accumulator type for no gain over the assert.

Nashtare: Correction to my reason above — the borrow is not the obstacle. `par_fold_reduce`'s identity closure already captures `fold` by reference, so a tile holding `&PlaneFold` would live inside that same borrow region and compile fine. The actual cost is that it makes `RowTile` generic in `R` and carries two lifetimes, where today the type is plain and only `fill`/`read_row`/`read_lane_group` are generic. That is why I took the assert.

multi-stark/src/rounds/sliced.rs
Nashtare: Added in bbb9463ab: `row_value` opens with `debug_assert_eq!(bytes.len(), self.groups * PLANE_BYTES)`, with a `# Panics` line on the doc comment. Both callers (`row_pair`'s two halves) already satisfy it, so it is a contract pin rather than a fix; cfb3f4f15 then replaces the `as_chunks` walk with a zip over the per-group tables, which makes the group count come from `self` rather than from the slice length.

tcoratger (approved):
Re-reviewed the delta `fe3858a3..4c7dbd3e` (four new commits) at head `4c7dbd3e`. Everything I raised is addressed; **approving.**

**[MAJOR] single word pair — fixed, confirmed by mutation both ways.** With the loop widened to `32/16/8/4 * SHORTEST`, `RowTile::fill`'s `[pair, pair + fold.words / ROW_HALVES] -> [pair, fold.words - 1]` **fails** at `4096 rows, tower`. Narrowing the array back to the original two entries with that same mutant in place makes it **pass** again. So the widening is what catches it, exactly as you reported.

**[MAJOR] lane kernel unexecuted in CI — accepted as documented.** No CI leg changed, so `boundary_evals_lanes` / `read_lane_group` still never run in CI. My finding offered documentation as the alternative remedy and you took it, with a local Sapphire Rapids run; I reproduced that here. Under `-C target-feature=+avx2,+vpclmulqdq` (WIDTH = 2) the packed path is live and genuinely covered: `lane_pair` reading `self.cell(lane, column)` instead of `self.cell(lane + step, column)` **fails** `a_boundary_round_and_fold_on_the_planes_match_the_unsliced_kernels` there, and **passes** on the default target. Not blocking this PR, but the gap is now a repo-level one — worth a follow-up issue rather than leaving it in a review thread.

**[MINOR] ×4 — all fixed.** `Vec<[R; GROUP_ENTRIES]>` with the per-group tables hoisted (cfb3f4f15); `RowTile::lane_pair` walking the lanes once (113214144); the two stride `debug_assert_eq!`s in `fill` and the `bytes.len()` assert in `row_value` (bbb9463ab). I read the delta line by line: `lane_pair` is exactly `lane_group`'s `R::Packing::from_fn` unrolled over `as_slice_mut()`, and `PlaneFold::new`'s new triple zip has all three sides at `corners.div_ceil(GROUP_CORNERS)` with the inner zip truncating at `subset_sums`' length just as the old indexed write did — same values, same zero tail. Your correction on the `&PlaneFold` borrow is right, and the assert is the better trade either way.

**Checks I ran on the delta** (nothing modified; experiments reverted):

- `p3-multi-stark` lib tests, all 285, release **and** debug (so the new `debug_assert`s actually fire), under the default target and `+avx2,+vpclmulqdq`. Green everywhere.
- End-to-end proof identity against the stack base `16db71d3`, under `+avx2,+vpclmulqdq` so the lane path is the one running: `prove_hash_binary -l 14 --security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4`, FNV-1a over the postcard bytes.

  | objective | `16db71d3` | `4c7dbd3e` |
  | --- | --- | --- |
  | blake-3-compressions | `767128 / eec91168337a82c9` | `767128 / eec91168337a82c9` |
  | keccak-f-permutations | `474636 / a0e5ef6ae9a2c7b9` | `474636 / a0e5ef6ae9a2c7b9` |

  Identical, and both verify.
- clippy `--all-targets` and rustfmt clean.

No new finding in the delta. The three `PlaneFold::new` call sites are all once per round, so the `vec![[R::ZERO; GROUP_ENTRIES]; groups]` build is off the hot path. The remaining [NIT]s (the four copies of the degree/reduce boilerplate, the `group_words` and `RowTile::new` doc wording) are unaddressed and stay optional.

One caveat that is not a finding: this sits on #2246, where the [MAJOR] about `every_sliced_round_count_agrees_with_the_generic_backend` is still open.

PR #2250, opened 2026-09-20
feat(examples): let the binary harness choose its commitment hash and leaf geometry

Nashtare:
Stacked on #2251. Adds the knobs; changes no default.

The Flock comparison suggests BLAKE3 over Keccak-256 for the Merkle tree and transcript, and
1024-byte leaves over our 256-byte ones. Neither could be measured here, because the harness
hardcoded the hash and sized a leaf from the fold schedule. This makes both choosable and reports
what they are worth.

### What it adds

- `GroupedCodewordMmcs::with_group_size` sizes a Merkle leaf independently of the fold schedule
  while keeping the per-round message-length cap. `for_folding` delegates to it with the fold
  coset, so the default grouping is unchanged. A test pins that a leaf wider than the fold coset
  gives a different root, ships the uncovered coset symbols as `missing_symbols`, and still
  verifies.
- A `HarnessHash` trait carrying the instance and its collision resistance, implemented for
  `Keccak256Hash` and `Blake3`. `BinaryStarkConfig` and `BooleanStarkConfig` become generic over
  it the way they already were over Merkle arity and transform, defaulting to `Keccak256Hash`.
- `--hash` (`keccak-256` | `blake-3`) and `--leaf-elements`, the latter rejecting non-powers-of-two
  at parse time. The report prints both.
- `collision_resistance_bits()` reads `H::COLLISION_RESISTANCE_BITS` instead of a literal `128`
  justified by a comment that Keccak-256 is shared by the transcript and the Merkle tree. Both
  hashes legitimately cap at 128 from their 32-byte digest, so no reported number moves; the
  constant is overridable for a hash weaker than its output width.

### What it measures

`--security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4`, medians of 3. `leaf hash` is
the `commit_matrix` span. On this host BLAKE3 dispatches its 16-wide AVX-512 assembly and
Keccak-256 its 8-wide AVX-512 path; both leaf sizes up to 1024 bytes hit the batched `hash_many`.

BLAKE3 compressions, `-l 18`:

| hash | elements/leaf | prove | leaf hash | verify | proof bytes |
| --- | ---: | ---: | ---: | ---: | ---: |
| keccak-256 | 16 (default) | 2965 ms | 115.0 ms | 36 ms | 993,065 |
| keccak-256 | 64 | 2966 ms | 104.0 ms | 41 ms | 1,644,075 |
| blake-3 | 16 | 2934 ms | 90.4 ms | 30 ms | 987,870 |
| blake-3 | 64 | 2935 ms | 82.4 ms | 32 ms | 1,657,480 |
| blake-3 | 128 | 2978 ms | 118.0 ms | 34 ms | 2,485,162 |

Keccak-f permutations at `-l 18`, and both objectives at `-l 16` and `-l 14`, have the same shape.

Three things follow.

**Merkle hashing is about 4% of prove, not the third it once was.** Every `commit folded codeword`
plus the base leaf hashing is 125.5 ms of 2965 on the BLAKE3 workload and 19.3 of 781 on Keccak-f.
The Boolean commitment shrank the committed codeword by two orders of magnitude, so a hash swap is
capped at roughly 4% before it starts.

**BLAKE3 is a real but small win.** At the default geometry it takes 21% off leaf hashing and 17%
off verification, worth about 1% of prove — inside the run-to-run band. The span deltas are the
trustworthy signal; the total is not.

**Flock's leaf geometry does not transfer.** Sixty-four elements per leaf buys 9% of leaf hashing
and costs 68% more proof bytes. Reaching a 1024-byte leaf cleanly through `--folding 6`, so that
no query authenticates symbols it did not ask for, still costs 61% more bytes and 23% more verify
for 0.7% of prove.

Past 1024 bytes BLAKE3 gets worse, not better: `hash_many` batches at most one chunk, and beyond
that it falls back to hashing each leaf alone and loses the wide batch. That is the cliff, and it
is presumably why Flock stopped where it did.

### Recommendation

Keep both defaults. The knob is worth having for shapes where Merkle work dominates — notably the
dense `prove_binary_air` path, whose codeword is far larger than the Boolean one, where the trade
should flip.

One caveat the accounting does not model, stated rather than fixed: grinding is priced by
hash-invocation count, not hash cost, so a cheaper transcript hash is cheaper for a forger by the
same factor it is for the prover. It affects nothing measured above, since the proof-of-work bits
were zero in every run, but it is a reason to weigh a default change deliberately rather than on
throughput alone.

### Correctness

The default configuration is untouched and its proofs are byte-identical to #2246: 993,065 bytes
and 96.23 bits on every base and branch run at `-l 18`. Every row of the matrix was proved at its
stated composed security, and within each objective and size that figure is identical across all
leaf sizes and both hashes.

`cargo test --workspace --features parallel` passes; clippy and rustfmt are clean.


tcoratger (commented):
Reviewed against the stack base (`perf/opening-share`), treating the new configuration surface as library API rather than "just an example" — example harnesses are what downstream users copy verbatim, and this one exposes the commitment hash and Merkle leaf geometry, which is exactly where concrete security parameters get silently weakened.

**The cryptography is sound.** I checked digest widths, tree-depth binding, leaf/node domain separation and prover/verifier agreement and found no soundness defect; defaults are untouched and both hashes give a full 256-bit digest, so the reported 128-bit cap is correct. **However, the one number the PR exists to measure is misreported**, and that also makes its only wide-leaf Boolean test vacuous.

## The configuration surface

| Knob | Legal values | Consequence | Validated? |
| --- | --- | --- | --- |
| `--hash` | `keccak-256`, `blake-3` (+ prefix aliases) | Both are `CryptographicHasher<u8, [u8;32]>`, so both give 128-bit Merkle collision resistance and a 128-bit transcript birthday bound. Changes only speed and proof bytes. | Yes — `ValueEnum`, exhaustive match, compile-time monomorphization. |
| `--leaf-elements` | any power of two (incl. `1`; `0` rejected since `0.is_power_of_two() == false`) | Leaf width vs tree height. Wider leaf => fewer, longer hash messages, but each query authenticates symbols it did not ask for and those ship in `missing_symbols`. | Power-of-two only. **Not** checked against the codeword: larger values are silently clamped to `message_len` per round and the clamp is never reported. |
| `--merkle-arity` (pre-existing) | 2, 4 | Interacts with the above: `(arity, hash)` is now a 4-arm dispatch in each entry point. | Yes, `UnsupportedMerkleArity`. |

```
committed codeword, height = message_len << log_inv_rate
  +----+----+----+----+----+----+----+----+ ... symbols
  |  0 |  1 |  2 |  3 |  4 |  5 |  6 |  7 |
  +----+----+----+----+----+----+----+----+
  \-------- g=4 -------/\------ g=4 ------/     one Merkle leaf per group
       leaf 0                 leaf 1            tree height = height / g

  g = min(leaf_elements, height >> log_inv_rate)   <-- the clamp, applied PER ROUND
                                                       and never surfaced to the report
```

### What I checked and found safe

- **Tree-depth binding.** The classic second-preimage attack needs the verifier to accept a tree of a different depth. `GroupedCodewordMmcs::verify_multi_batch` (`binary-pcs/src/grouped_mmcs.rs:269`) derives `group_size` from verifier-side `dimensions` and hands `MerkleTreeMmcs` a fixed `Dimensions { width: g, height: dim.height / g }`; `verify_batch_pruned` walks a depth fixed by those dimensions, never by the proof. The prover cannot choose the geometry.
- **Leaf vs node domain separation.** There is none — a leaf is `H(serialized elements)` and a node is `H(d_0 ‖ … ‖ d_{N-1})` (`symmetric/src/compression.rs:109`), both untagged. With `--merkle-arity 2 --leaf-elements 4` the two messages are *literally the same length* (64 bytes) under the same `H`. This is safe only because of the point above, and it was already reachable via `--folding 2`. Worth one sentence in the `with_group_size` doc, since the new knob makes the coincidence a first-class choice rather than an accident.
- **Geometry/hash mismatch fails closed.** `a_leaf_wider_than_the_fold_coset…` pins `assert_ne!(wide_root, root)`, and a verifier with a different `g` recomputes `missing_count = groups.len()*g - requested.len()` from its own `g`, so a cross-geometry proof is rejected on `OpeningShape` or on the root. No silent acceptance.
- **No panic on the new paths.** I swept `(arity ∈ {2,4}) × (log_inv_rate ∈ {1,2}) × (leaf_elements ∈ {1, 32, 2^20, 2^30})` through `prove_binary_air` — no panics, degenerate 2-leaf trees under arity 4 included. `leaf_elements = 1` (narrower than the fold coset) works.
- **BLAKE3 past one chunk.** Leaves over 1024 bytes drop out of `hash_many`'s batched path into the scalar hasher (`blake3/src/lib.rs:226`), which is pinned against `blake3::hash` by that crate's own tests. Correct, just slower — matching the PR body.

---

## [MAJOR] The reported leaf geometry is the request, not the leaf — `examples/src/binary.rs:129`

`leaf_elements_of` echoes `options.leaf_elements` verbatim. The actual leaf is `group_size_at(height) = min(group_size, height >> log_inv_rate)` (`binary-pcs/src/grouped_mmcs.rs:75`), clamped per round. So a benchmark harness built to measure this knob prints a number that can be off by an arbitrary factor.

Measured on the PR head, `prove_hash_binary --objective blake-3-compressions --log-trace-length 5`:

```
le=2048  -> Merkle leaf: 2048  field elements (32768   bytes)  Proof size: 529698 bytes
le=4096  -> Merkle leaf: 4096  field elements (65536   bytes)  Proof size: 529824 bytes
le=8192  -> Merkle leaf: 8192  field elements (131072  bytes)  Proof size: 529824 bytes
le=65536 -> Merkle leaf: 65536 field elements (1048576 bytes)  Proof size: 529824 bytes
```

The last three are the *same run* — the effective leaf is 4096 everywhere — yet one of them claims a 1 MiB leaf. This is not cosmetic: the value is also stored on the config and returned by the public `BinaryStarkConfig::leaf_elements()` / `BooleanStarkConfig::leaf_elements()` (`examples/src/binary.rs:148`, `:235`), both documented as "field elements each Merkle leaf of the base codeword packs". That is false whenever the clamp bites.

It is also why the Boolean path looks flat: the Boolean commitment's codeword is `2^(arity - BitRingSwitch::ABSORBED)` symbols, so at `log_height = 8` the committed message is only 8 symbols and *every* `leaf_elements >= 8` produces a byte-identical proof.

```rust
const fn leaf_elements_of(pcs_config: &BinaryPcsConfig, leaf_elements: Option<usize>) -> usize {
    match leaf_elements {
        Some(elements) => elements,          // <-- never clamped
        None => 1 << pcs_config.log_folding_factor(),
    }
}
```

**Fix:** clamp to the base round's message length, and say so in the report. Best is to make the cap single-sourced — `pub const fn group_size_at(&self, height: usize) -> Option<usize>` on `GroupedCodewordMmcs` — and query it; failing that:

```rust
const fn leaf_elements_of(pcs_config: &BinaryPcsConfig, leaf_elements: Option<usize>) -> usize {
    let requested = match leaf_elements {
        Some(elements) => elements,
        None => 1 << pcs_config.log_folding_factor(),
    };
    // `GroupedCodewordMmcs` caps a leaf at the round's message length.
    let base_message_len = 1usize << pcs_config.num_variables();
    if requested < base_message_len { requested } else { base_message_len }
}
```

Consider also printing the clamp explicitly (`Merkle leaf: 4096 field elements (65536 bytes; requested 65536, capped at the base message length)`), since the geometry still differs per fold round and a single scalar can only describe the base one.

**Verify:** `assert_eq!(report(Some(1 << 20)).leaf_elements, report(Some(1 << pcs_num_vars)).leaf_elements)` — or, end to end, the four CLI runs above must print the same leaf whenever they print the same `Proof size`.

---

## [MAJOR] `a_boolean_trace_proves_under_either_hash` asserts a geometry it never exercises — `examples/src/binary.rs:1231`

The test sets `leaf_elements: Some(64)` at `log_height = 8` and asserts `report.leaf_elements == 64`. The effective leaf there is **8**. The assertion passes *because of* the bug above, so it is a test that cannot fail for the reason it claims to check — and it is the PR's only Boolean × wide-leaf coverage. Fixing the MAJOR above will break this test, which is the correct outcome.

**Fix:** raise `log_height` until the Boolean committed arity clears 64 symbols (`log_height >= 13` gives `arity - 7 >= 6`), or assert against the clamped value.

**Verify:** with the clamp applied, the current test fails with `left: 8, right: 64`.

---

## [MINOR] `Merkle leaf: … bytes` overflows on large `--leaf-elements` — `examples/src/binary.rs:424`

`self.leaf_elements * F::NUM_BYTES` with `NUM_BYTES = 16` overflows at `leaf_elements >= 2^60`. Confirmed on the head:

```
release: Merkle leaf: 1152921504606846976 field elements (0 bytes)
debug:   panicked at examples/src/binary.rs:424: attempt to multiply with overflow
```

Operator-supplied, so not attacker-reachable — but a harness should not crash on its own CLI. Clamping in `leaf_elements_of` makes this unreachable in practice; `saturating_mul` (or rejecting `leaf_elements > 1 << 40` in `parse_leaf_elements`) closes it properly.

---

## [MINOR] Two different behaviours for the same illegal input — `examples/src/binary.rs:381` vs `binary-pcs/src/grouped_mmcs.rs:64`

`BinaryProofOptions::check_leaf_elements` returns `UnsupportedLeafElements` from the `prove_*` dispatchers, but the equally-`pub` `binary_config` / `boolean_config` reach `with_group_size`'s `assert!` and panic. Both are documented (`# Panics` / `# Errors`), so nothing is wrong — but a downstream copier calling the config builder directly gets a panic where the error variant already exists. Push the check down: have `grouped_mmcs` return `Result<_, BinaryProofError>` and keep the `assert!` as the `for_folding` invariant only.

---

## [MINOR] `COLLISION_RESISTANCE_BITS` is a defaulted security constant — `examples/src/binary.rs:71`

```rust
const COLLISION_RESISTANCE_BITS: usize = 128;
```

`collision_resistance_bits()` now reads this instead of a literal, which is the right direction. But a *default* means a new `impl HarnessHash` silently inherits 128 bits. The trait bound `CryptographicHasher<u8, [u8; 32]>` pins the digest width, so the default is currently always correct — the risk is a construction weaker than its output (a reduced-round variant, a truncated-then-padded digest). Dropping the default costs one line per impl and makes every future hash state its own number. This is the single place in the PR where a security parameter could be weakened by omission rather than by choice.

---

## [MINOR] `leaf_elements` is threaded by hand instead of read off the config — `examples/src/binary.rs:~840`

`prove_and_verify` gained a `leaf_elements: usize` parameter and an `#[allow(clippy::too_many_arguments)]`, even though both config types already expose `leaf_elements()`. The reason is that `C: MultiStarkConfig` does not. Adding `fn leaf_elements(&self) -> usize` to the existing private `HarnessErrors` trait removes the parameter, the `allow`, and the two call sites that currently have to remember to pass it.

---

## [MINOR] Monomorphization: the dispatch matrix doubled

Each of `prove_binary_air_with_ntt_and_backend` and `prove_boolean_air_with_backend` now has 4 arms (`{2,4} × {Keccak, Blake3}`), each instantiating a full prover + verifier stack, and the binary path multiplies that by `Ntt` and `Backend`. Dispatch is at setup only — the hash is a zero-sized type resolved statically, nothing dynamic lands in a leaf loop, which is the right call — but debug `cargo test -p p3-examples` compile time is now noticeably worse. Fine for an example; worth a comment if a third hash is ever added, because the growth is multiplicative.

---

## [NIT]

- `HashFamily::fmt` prints `blake3` while the CLI's canonical name is `blake-3` (`examples/src/binary.rs:88`). It round-trips only because `blake3` happens to be a registered alias. Print `blake-3`.
- `HarnessHash: … + Clone + Copy + …` — `Copy: Clone`, so `Clone` is redundant (`examples/src/binary.rs:61`).
- `#[derive(…, Default, …)]` on `HashFamily` is dead: `BinaryProofOptions::default()` sets the field explicitly.
- `BinaryCommitmentHashOptions` variants are undocumented (`examples/src/parsers.rs:61`) where the neighbouring enums document theirs.
- The transcript seed `b"p3-examples-binary-hash-air-v1"` is unchanged across all knob settings. Correct here (the type pins the hash, the config pins the geometry, and both sides share one object), but a copier lifting this into a deployment must pin `--hash`/`--leaf-elements`/`--merkle-arity` out of band — the digest does not bind them. One line in `binary_challenger`'s doc would save someone.

---

## Test coverage

Dispatch arms actually proven-and-verified by the suite:

| entry point | (2, Keccak) | (2, Blake3) | (4, Keccak) | (4, Blake3) |
| --- | :---: | :---: | :---: | :---: |
| `prove_binary_air_with_ntt_and_backend` | yes | yes (`a_blake3_commitment_proves…`) | yes (`proof_size_differs_between_merkle_arities`) | **no** |
| `prove_boolean_air_with_backend` | yes | yes (`a_boolean_trace_proves_under_either_hash`) | **no** | **no** |

Named uncovered combinations:

- **`(merkle_arity = 4, Blake3)` on both paths** and **`(4, Keccak)` on the Boolean path.** The CLI default is `--merkle-arity 4`, so `prove_hash_binary --hash blake-3` — the exact configuration every row of the PR's benchmark table was produced with — runs the one arm with no test. Fix with a parameterized loop, which is cheap because the arms are independent:

  ```rust
  #[test]
  fn every_hash_and_arity_proves_a_boolean_trace() {
      for arity in [2, 4] {
          for hash in [HashFamily::Keccak256, HashFamily::Blake3] {
              let report = prove_boolean_air(&XorAir, Table::new(xor_trace(8).transpose()),
                  BinaryProofOptions { merkle_arity: arity, hash, ..Default::default() })
                  .unwrap_or_else(|e| panic!("{arity}/{hash}: {e}"));
              assert_eq!(report.hash, hash);
          }
      }
  }
  ```

- **Boundary geometries.** `leaf_elements = 1` (narrower than the fold coset, splitting a coset across leaves — the case `with_group_size`'s doc describes) is untested; I ran it and it works, so it is one line to pin. `leaf_elements == message_len` exactly, and `leaf_elements = 2 * message_len` (first clamped value), are the two inputs that would have caught the MAJOR.

- **Cross-configuration rejection.** `a_leaf_wider_than_the_fold_coset…` (`binary-pcs/src/grouped_mmcs.rs:474`) asserts `assert_ne!(wide_root, root)`, which shows the roots differ but not that a mismatched verifier *rejects*. One extra line makes the binding property explicit: `assert!(wide_leaves.verify_multi_batch(&root, &dims, &coset, &wide_rows, &wide_proof).is_err())`.

- **Proptest.** Over `(height, leaf_elements, hash)` it adds little beyond the enumerated matrix here: the interesting structure is the `min(group_size, message_len)` clamp, exercised by three enumerated points (below, equal, above). A proptest *would* pay off one level down, on `GroupedCodewordMmcs` alone, over `(log_height, log_inv_rate, log_group_size, query_set)` asserting round-trip verify and `missing_symbols.len() == groups*g - |queries|` — fast, and it covers the arithmetic the harness only samples.

---

## Could not verify

- The PR body's benchmark table (`-l 18`, and "the same shape" at `-l 16` and `-l 14`) — at `-l 18` the committed message is comfortably above 128 symbols, so those rows are almost certainly genuine. But at `-l 14` the Boolean committed arity is small enough that I could not rule out the clamp flattening the 64- and 128-element rows into the default. Worth re-running those two sizes after the clamp is reported, to confirm the "same shape" claim is a measurement and not an artifact.
- Is there a stated concrete security target for this harness anywhere outside `--security-bits`? I found none — `collision_resistance_bits()` returning `Some(128)` is the only hash-side input to the composed number, and the PR correctly leaves it unmoved.
- The PR body's grinding caveat (a cheaper transcript hash is cheaper for a forger by the same factor) is correct and correctly scoped as "stated rather than fixed" — `pow_bits` was 0 in every run. No objection, but it is the one thing that would need real thought before `--hash blake-3` became a default.

---

**Bottom line:** merge after fixing the leaf-geometry misreport (`examples/src/binary.rs:129`) and the test it invalidates (`:1231`). No soundness issue found; the crypto-facing parts of the configuration surface — digest width, tree-depth binding, prover/verifier agreement — are all correct.

examples/src/binary.rs
```diff
@@ -39,41 +41,132 @@ use p3_multi_stark::{
...
+        0,
+    );
+    match leaf_elements {
+        Some(elements) => Mmcs::with_group_size(merkle, pcs_config, elements),
+        None => Mmcs::for_folding(merkle, pcs_config),
+    }
+}
+
+/// Field elements one Merkle leaf packs under `pcs_config`.
+const fn leaf_elements_of(pcs_config: &BinaryPcsConfig, leaf_elements: Option<usize>) -> usize {
```
tcoratger: **[MAJOR] The reported leaf geometry is the request, not the leaf — `examples/src/binary.rs:129`**

`leaf_elements_of` echoes `options.leaf_elements` verbatim. The actual leaf is `group_size_at(height) = min(group_size, height >> log_inv_rate)` (`binary-pcs/src/grouped_mmcs.rs:75`), clamped per round. So a benchmark harness built to measure this knob prints a number that can be off by an arbitrary factor.

Measured on the PR head, `prove_hash_binary --objective blake-3-compressions --log-trace-length 5`:

```
le=2048  -> Merkle leaf: 2048  field elements (32768   bytes)  Proof size: 529698 bytes
le=4096  -> Merkle leaf: 4096  field elements (65536   bytes)  Proof size: 529824 bytes
le=8192  -> Merkle leaf: 8192  field elements (131072  bytes)  Proof size: 529824 bytes
le=65536 -> Merkle leaf: 65536 field elements (1048576 bytes)  Proof size: 529824 bytes
```

The last three are the *same run* — the effective leaf is 4096 everywhere — yet one of them claims a 1 MiB leaf. This is not cosmetic: the value is also stored on the config and returned by the public `BinaryStarkConfig::leaf_elements()` / `BooleanStarkConfig::leaf_elements()` (`examples/src/binary.rs:148`, `:235`), both documented as "field elements each Merkle leaf of the base codeword packs". That is false whenever the clamp bites.

It is also why the Boolean path looks flat: the Boolean commitment's codeword is `2^(arity - BitRingSwitch::ABSORBED)` symbols, so at `log_height = 8` the committed message is only 8 symbols and *every* `leaf_elements >= 8` produces a byte-identical proof.

```rust
const fn leaf_elements_of(pcs_config: &BinaryPcsConfig, leaf_elements: Option<usize>) -> usize {
    match leaf_elements {
        Some(elements) => elements,          // <-- never clamped
        None => 1 << pcs_config.log_folding_factor(),
    }
}
```

**Fix:** clamp to the base round's message length, and say so in the report. Best is to make the cap single-sourced — `pub const fn group_size_at(&self, height: usize) -> Option<usize>` on `GroupedCodewordMmcs` — and query it; failing that:

```rust
const fn leaf_elements_of(pcs_config: &BinaryPcsConfig, leaf_elements: Option<usize>) -> usize {
    let requested = match leaf_elements {
        Some(elements) => elements,
        None => 1 << pcs_config.log_folding_factor(),
    };
    // `GroupedCodewordMmcs` caps a leaf at the round's message length.
    let base_message_len = 1usize << pcs_config.num_variables();
    if requested < base_message_len { requested } else { base_message_len }
}
```

Consider also printing the clamp explicitly (`Merkle leaf: 4096 field elements (65536 bytes; requested 65536, capped at the base message length)`), since the geometry still differs per fold round and a single scalar can only describe the base one.

**Verify:** `assert_eq!(report(Some(1 << 20)).leaf_elements, report(Some(1 << pcs_num_vars)).leaf_elements)` — or, end to end, the four CLI runs above must print the same leaf whenever they print the same `Proof size`.

examples/src/binary.rs
```diff
@@ -977,6 +1174,97 @@ mod tests {
...
+        assert_eq!(keccak.hash, HashFamily::Keccak256);
+        assert_eq!(blake3.hash, HashFamily::Blake3);
+        // Both digests are 32 bytes, so the composed security is the same bound.
+        assert_eq!(blake3.security_bits, keccak.security_bits);
+        // A different transcript draws different challenges, hence a different proof.
+        assert_ne!(blake3.proof_bytes, keccak.proof_bytes);
+    }
+
+    #[test]
+    fn a_boolean_trace_proves_under_either_hash() {
```
tcoratger, line 1273: **[MAJOR] `a_boolean_trace_proves_under_either_hash` asserts a geometry it never exercises — `examples/src/binary.rs:1231`**

The test sets `leaf_elements: Some(64)` at `log_height = 8` and asserts `report.leaf_elements == 64`. The effective leaf there is **8**. The assertion passes *because of* the bug above, so it is a test that cannot fail for the reason it claims to check — and it is the PR's only Boolean × wide-leaf coverage. Fixing the MAJOR above will break this test, which is the correct outcome.

**Fix:** raise `log_height` until the Boolean committed arity clears 64 symbols (`log_height >= 13` gives `arity - 7 >= 6`), or assert against the clamped value.

**Verify:** with the clamp applied, the current test fails with `left: 8, right: 64`.

examples/src/binary.rs
```diff
@@ -283,6 +416,13 @@ impl fmt::Display for BinaryProofReport {
         writeln!(f, "Rows: {}", self.rows)?;
         writeln!(f, "Width: {}", self.width)?;
         writeln!(f, "Stacked variables: {}", self.stacked_variables)?;
+        writeln!(f, "Hash: {}", self.hash)?;
+        writeln!(
+            f,
+            "Merkle leaf: {} field elements ({} bytes)",
+            self.leaf_elements,
+            self.leaf_elements * F::NUM_BYTES
```
tcoratger: **[MINOR] `Merkle leaf: … bytes` overflows on large `--leaf-elements` — `examples/src/binary.rs:424`**

`self.leaf_elements * F::NUM_BYTES` with `NUM_BYTES = 16` overflows at `leaf_elements >= 2^60`. Confirmed on the head:

```
release: Merkle leaf: 1152921504606846976 field elements (0 bytes)
debug:   panicked at examples/src/binary.rs:424: attempt to multiply with overflow
```

Operator-supplied, so not attacker-reachable — but a harness should not crash on its own CLI. Clamping in `leaf_elements_of` makes this unreachable in practice; `saturating_mul` (or rejecting `leaf_elements > 1 << 40` in `parse_leaf_elements`) closes it properly.

tcoratger (commented):

---

Reviewed against the stack base (`perf/opening-share`), treating the new configuration surface as library API rather than "just an example" — example harnesses are what downstream users copy verbatim, and this one exposes the commitment hash and Merkle leaf geometry, which is exactly where concrete security parameters get silently weakened.

**The cryptography is sound.** I checked digest widths, tree-depth binding, leaf/node domain separation and prover/verifier agreement and found no soundness defect; defaults are untouched and both hashes give a full 256-bit digest, so the reported 128-bit cap is correct. **However, the one number the PR exists to measure is misreported**, and that also makes its only wide-leaf Boolean test vacuous.

---

## The configuration surface

| Knob | Legal values | Consequence | Validated? |
| --- | --- | --- | --- |
| `--hash` | `keccak-256`, `blake-3` (+ prefix aliases) | Both are `CryptographicHasher<u8, [u8;32]>`, so both give 128-bit Merkle collision resistance and a 128-bit transcript birthday bound. Changes only speed and proof bytes. | Yes — `ValueEnum`, exhaustive match, compile-time monomorphization. |
| `--leaf-elements` | any power of two (incl. `1`; `0` rejected since `0.is_power_of_two() == false`) | Leaf width vs tree height. Wider leaf => fewer, longer hash messages, but each query authenticates symbols it did not ask for and those ship in `missing_symbols`. | Power-of-two only. **Not** checked against the codeword: larger values are silently clamped to `message_len` per round and the clamp is never reported. |
| `--merkle-arity` (pre-existing) | 2, 4 | Interacts with the above: `(arity, hash)` is now a 4-arm dispatch in each entry point. | Yes, `UnsupportedMerkleArity`. |

```
committed codeword, height = message_len << log_inv_rate
  +----+----+----+----+----+----+----+----+ ... symbols
  |  0 |  1 |  2 |  3 |  4 |  5 |  6 |  7 |
  +----+----+----+----+----+----+----+----+
  \-------- g=4 -------/\------ g=4 ------/     one Merkle leaf per group
       leaf 0                 leaf 1            tree height = height / g

  g = min(leaf_elements, height >> log_inv_rate)   <-- the clamp, applied PER ROUND
                                                       and never surfaced to the report
```

### What I checked and found safe

- **Tree-depth binding.** The classic second-preimage attack needs the verifier to accept a tree of a different depth. `GroupedCodewordMmcs::verify_multi_batch` (`binary-pcs/src/grouped_mmcs.rs:269`) derives `group_size` from verifier-side `dimensions` and hands `MerkleTreeMmcs` a fixed `Dimensions { width: g, height: dim.height / g }`; `verify_batch_pruned` walks a depth fixed by those dimensions, never by the proof. The prover cannot choose the geometry.
- **Leaf vs node domain separation.** There is none — a leaf is `H(serialized elements)` and a node is `H(d_0 ‖ … ‖ d_{N-1})` (`symmetric/src/compression.rs:109`), both untagged. With `--merkle-arity 2 --leaf-elements 4` the two messages are *literally the same length* (64 bytes) under the same `H`. This is safe only because of the point above, and it was already reachable via `--folding 2`. Worth one sentence in the `with_group_size` doc, since the new knob makes the coincidence a first-class choice rather than an accident.
- **Geometry/hash mismatch fails closed.** `a_leaf_wider_than_the_fold_coset…` pins `assert_ne!(wide_root, root)`, and a verifier with a different `g` recomputes `missing_count = groups.len()*g - requested.len()` from its own `g`, so a cross-geometry proof is rejected on `OpeningShape` or on the root. No silent acceptance.
- **No panic on the new paths.** I swept `(arity ∈ {2,4}) × (log_inv_rate ∈ {1,2}) × (leaf_elements ∈ {1, 32, 2^20, 2^30})` through `prove_binary_air` — no panics, degenerate 2-leaf trees under arity 4 included. `leaf_elements = 1` (narrower than the fold coset) works.
- **BLAKE3 past one chunk.** Leaves over 1024 bytes drop out of `hash_many`'s batched path into the scalar hasher (`blake3/src/lib.rs:226`), which is pinned against `blake3::hash` by that crate's own tests. Correct, just slower — matching the PR body.

---

### [MINOR] Monomorphization: the dispatch matrix doubled

Each of `prove_binary_air_with_ntt_and_backend` and `prove_boolean_air_with_backend` now has 4 arms (`{2,4} × {Keccak, Blake3}`), each instantiating a full prover + verifier stack, and the binary path multiplies that by `Ntt` and `Backend`. Dispatch is at setup only — the hash is a zero-sized type resolved statically, nothing dynamic lands in a leaf loop, which is the right call — but debug `cargo test -p p3-examples` compile time is now noticeably worse. Fine for an example; worth a comment if a third hash is ever added, because the growth is multiplicative.

### [NIT]

- `HashFamily::fmt` prints `blake3` while the CLI's canonical name is `blake-3` (`examples/src/binary.rs:88`). It round-trips only because `blake3` happens to be a registered alias. Print `blake-3`.
- `HarnessHash: … + Clone + Copy + …` — `Copy: Clone`, so `Clone` is redundant (`examples/src/binary.rs:61`).
- `#[derive(…, Default, …)]` on `HashFamily` is dead: `BinaryProofOptions::default()` sets the field explicitly.
- `BinaryCommitmentHashOptions` variants are undocumented (`examples/src/parsers.rs:61`) where the neighbouring enums document theirs.
- The transcript seed `b"p3-examples-binary-hash-air-v1"` is unchanged across all knob settings. Correct here (the type pins the hash, the config pins the geometry, and both sides share one object), but a copier lifting this into a deployment must pin `--hash`/`--leaf-elements`/`--merkle-arity` out of band — the digest does not bind them. One line in `binary_challenger`'s doc would save someone.

---

## Test coverage

Dispatch arms actually proven-and-verified by the suite:

| entry point | (2, Keccak) | (2, Blake3) | (4, Keccak) | (4, Blake3) |
| --- | :---: | :---: | :---: | :---: |
| `prove_binary_air_with_ntt_and_backend` | yes | yes (`a_blake3_commitment_proves…`) | yes (`proof_size_differs_between_merkle_arities`) | **no** |
| `prove_boolean_air_with_backend` | yes | yes (`a_boolean_trace_proves_under_either_hash`) | **no** | **no** |

Named uncovered combinations:

- **`(merkle_arity = 4, Blake3)` on both paths** and **`(4, Keccak)` on the Boolean path.** The CLI default is `--merkle-arity 4`, so `prove_hash_binary --hash blake-3` — the exact configuration every row of the PR's benchmark table was produced with — runs the one arm with no test. Fix with a parameterized loop, which is cheap because the arms are independent:

  ```rust
  #[test]
  fn every_hash_and_arity_proves_a_boolean_trace() {
      for arity in [2, 4] {
          for hash in [HashFamily::Keccak256, HashFamily::Blake3] {
              let report = prove_boolean_air(&XorAir, Table::new(xor_trace(8).transpose()),
                  BinaryProofOptions { merkle_arity: arity, hash, ..Default::default() })
                  .unwrap_or_else(|e| panic!("{arity}/{hash}: {e}"));
              assert_eq!(report.hash, hash);
          }
      }
  }
  ```

- **Boundary geometries.** `leaf_elements = 1` (narrower than the fold coset, splitting a coset across leaves — the case `with_group_size`'s doc describes) is untested; I ran it and it works, so it is one line to pin. `leaf_elements == message_len` exactly, and `leaf_elements = 2 * message_len` (first clamped value), are the two inputs that would have caught the MAJOR.

- **Cross-configuration rejection.** `a_leaf_wider_than_the_fold_coset…` (`binary-pcs/src/grouped_mmcs.rs:474`) asserts `assert_ne!(wide_root, root)`, which shows the roots differ but not that a mismatched verifier *rejects*. One extra line makes the binding property explicit: `assert!(wide_leaves.verify_multi_batch(&root, &dims, &coset, &wide_rows, &wide_proof).is_err())`.

- **Proptest.** Over `(height, leaf_elements, hash)` it adds little beyond the enumerated matrix here: the interesting structure is the `min(group_size, message_len)` clamp, exercised by three enumerated points (below, equal, above). A proptest *would* pay off one level down, on `GroupedCodewordMmcs` alone, over `(log_height, log_inv_rate, log_group_size, query_set)` asserting round-trip verify and `missing_symbols.len() == groups*g - |queries|` — fast, and it covers the arithmetic the harness only samples.

---

## Could not verify

- The PR body's benchmark table (`-l 18`, and "the same shape" at `-l 16` and `-l 14`) — at `-l 18` the committed message is comfortably above 128 symbols, so those rows are almost certainly genuine. But at `-l 14` the Boolean committed arity is small enough that I could not rule out the clamp flattening the 64- and 128-element rows into the default. Worth re-running those two sizes after the clamp is reported, to confirm the "same shape" claim is a measurement and not an artifact.
- Is there a stated concrete security target for this harness anywhere outside `--security-bits`? I found none — `collision_resistance_bits()` returning `Some(128)` is the only hash-side input to the composed number, and the PR correctly leaves it unmoved.
- The PR body's grinding caveat (a cheaper transcript hash is cheaper for a forger by the same factor) is correct and correctly scoped as "stated rather than fixed" — `pow_bits` was 0 in every run. No objection, but it is the one thing that would need real thought before `--hash blake-3` became a default.

---

**Bottom line:** merge after fixing the leaf-geometry misreport (`examples/src/binary.rs:129`) and the test it invalidates (`:1231`). No soundness issue found; the crypto-facing parts of the configuration surface — digest width, tree-depth binding, prover/verifier agreement — are all correct.

examples/src/binary.rs
```diff
@@ -257,6 +376,16 @@ impl BinaryProofOptions {
             security_level: self.security_bits,
         }
     }
+
+    /// Reject a leaf size no Merkle commitment can be grouped to.
+    const fn check_leaf_elements(&self) -> Result<(), BinaryProofError> {
```
tcoratger: **[MINOR] Two different behaviours for the same illegal input — `examples/src/binary.rs:381` vs `binary-pcs/src/grouped_mmcs.rs:64`**

`BinaryProofOptions::check_leaf_elements` returns `UnsupportedLeafElements` from the `prove_*` dispatchers, but the equally-`pub` `binary_config` / `boolean_config` reach `with_group_size`'s `assert!` and panic. Both are documented (`# Panics` / `# Errors`), so nothing is wrong — but a downstream copier calling the config builder directly gets a panic where the error variant already exists. Push the check down: have `grouped_mmcs` return `Result<_, BinaryProofError>` and keep the `assert!` as the `for_folding` invariant only.

examples/src/binary.rs
```diff
@@ -39,41 +41,132 @@ use p3_multi_stark::{
...
+    CryptographicHasher<u8, [u8; 32]> + Clone + Copy + fmt::Debug + Send + Sync + 'static
+{
+    /// The hash itself.
+    const INSTANCE: Self;
+
+    /// Collision resistance of this hash, in bits, as the security report's cap.
+    ///
+    /// The digest is 32 bytes wide, so the birthday bound is half of that. An implementor whose
+    /// construction is weaker than its output width lowers this.
+    const COLLISION_RESISTANCE_BITS: usize = 128;
```
tcoratger: **[MINOR] `COLLISION_RESISTANCE_BITS` is a defaulted security constant — `examples/src/binary.rs:71`**

```rust
const COLLISION_RESISTANCE_BITS: usize = 128;
```

`collision_resistance_bits()` now reads this instead of a literal, which is the right direction. But a *default* means a new `impl HarnessHash` silently inherits 128 bits. The trait bound `CryptographicHasher<u8, [u8; 32]>` pins the digest width, so the default is currently always correct — the risk is a construction weaker than its output (a reduced-round variant, a truncated-then-padded digest). Dropping the default costs one line per impl and makes every future hash state its own number. This is the single place in the PR where a security parameter could be weakened by omission rather than by choice.

examples/src/binary.rs
```diff
@@ -631,47 +795,71 @@ pub fn prove_boolean_air_with_backend<A>(
...
 /// Proves and verifies `air` against `trace` under `config`, reporting size and timing
 /// measurements.
 ///
 /// The statement's security is assessed once against `options.security_bits` before proving,
-/// so the timed phases are the plain prover and verifier.
-fn prove_and_verify<A, C>(
+/// so the timed phases are the plain prover and verifier. `leaf_elements` is what `config`
+/// resolved its Merkle grouping to, reported alongside the measurements.
+#[allow(clippy::too_many_arguments)]
+fn prove_and_verify<A, C, H>(
```
tcoratger: **[MINOR] `leaf_elements` is threaded by hand instead of read off the config — `examples/src/binary.rs:~840`**

`prove_and_verify` gained a `leaf_elements: usize` parameter and an `#[allow(clippy::too_many_arguments)]`, even though both config types already expose `leaf_elements()`. The reason is that `C: MultiStarkConfig` does not. Adding `fn leaf_elements(&self) -> usize` to the existing private `HarnessErrors` trait removes the parameter, the `allow`, and the two call sites that currently have to remember to pass it.

examples/src/binary.rs
Nashtare: Reproduced on the head: `--objective blake-3-compressions --log-trace-length 5` prints `le=4096`, `8192` and `65536` as three different leaves for the same 529,824-byte proof, because the committed arity there is 19 - 7 = 12 and the base message is 4096 symbols.

Fixed in `2b10f740c`, single-sourced the way you suggested: `GroupedCodewordMmcs::group_size_at` is now `pub` (`79c6ef36d`) and `leaf_elements_of` queries it at the base codeword height instead of echoing the request, so the report and both `leaf_elements()` accessors carry what the commitment packed. The report also keeps the request (`BinaryProofReport::requested_leaf_elements`) and names the clamp when the two differ:

```
Merkle leaf: 4096 field elements (65536 bytes; requested 65536, capped at the base message length)
```

Your `assert_eq!` is now a test, `a_leaf_wider_than_the_base_message_reports_the_cap`: at `log_height = 8` a 2^20 request and an exact 8 report the same leaf and the same proof bytes.

On what this does to the PR's table: no row of it was wrong. The cap is `2^(arity - 7)` and grows with the trace, so across the published sizes (`-l 14/16/18`, leaves of 16/64/128) the tightest cap is 2^18, at Keccak-f `-l 14`, and it never bit. I ran the same matrix at `-l 10` for both objectives, both hashes, at 16/64/128 before and after this fix: every proof size is byte-identical and every reported leaf is unchanged, including the default configuration.

examples/src/binary.rs
Nashtare: Confirmed. Three columns over 256 rows stack into ten variables, one committed element absorbs seven, and the base message is eight symbols, so the assertion was reading the echoed request; with the clamp applied it fails `left: 8, right: 64`.

`2b10f740c` takes your first route: `a_boolean_trace_proves_under_either_hash` runs at `log_height = 12`, where the stack is fourteen variables and the base message is 128 symbols, so the 64-element leaf is packed whole and the assertion means what it claims. The clamped case did not just disappear with it - `a_leaf_wider_than_the_base_message_reports_the_cap` pins it at `log_height = 8`, including that every request at or above the message length yields byte-identical proofs.

examples/src/binary.rs
Nashtare: Reproduced both halves on the head: `0 bytes` in release and the multiply panic in debug.

`2b10f740c` closes it twice over. The printed count is now the clamped leaf, which is at most the base message length, and the byte count is a `saturating_mul`. `--leaf-elements 1152921504606846976` prints `4096 field elements (65536 bytes; requested 1152921504606846976, capped at the base message length)` in release, and a debug build of the same run no longer panics.

I left `parse_leaf_elements` alone: a power of two is the only property the grouping needs, and rejecting a magnitude there would be a second, weaker statement of a bound the commitment already enforces.

examples/src/binary.rs
Nashtare: Fixed at the harness level in `64c26f929`: `grouped_mmcs` returns `Result<_, BinaryProofError>` and rejects a non-power-of-two before `with_group_size` sees it, so `binary_config` (whose error type is now `BinaryProofError`) and `boolean_config` return `UnsupportedLeafElements` where they used to panic. `BinaryProofOptions::check_leaf_elements` goes with it, since the dispatchers would otherwise validate the same input a second time and the two checks could drift apart.

I did not change `with_group_size`'s signature. Its assertion is now reachable only as the `for_folding` invariant, which is the shape you asked for, and keeping it an assertion is what lets both constructors stay `const fn`.

examples/src/binary.rs
Nashtare: Dropped the default in `dc8b68dc1`; `Keccak256Hash` and `Blake3` each state 128 for themselves. Your reading of the current bound is right, so no reported number moves - but a future impl now has to name its own figure instead of inheriting one it may not meet.

examples/src/binary.rs
Nashtare: Done in `9fe3e9da1`: `leaf_elements` is on the trait, the parameter and the two call sites that had to remember to pass it are gone, and so is the `#[allow]` - `prove_and_verify` takes six arguments now. I also renamed the trait to `HarnessConfig`, since it no longer only projects errors.

tcoratger (approved):
Re-reviewed at `b10eaac1` against my findings at `e5afb33c`. **All six threads are genuinely fixed; resolved all six. Approving.**

### The two MAJORs

**Leaf geometry misreport** — fixed properly, at the source rather than at the print site. `GroupedCodewordMmcs::group_size_at` is `pub` (`c1d2da02`) and `leaf_elements_of` now queries it at the base codeword height instead of echoing the request, so both `leaf_elements()` accessors and the report carry what the commitment packed. Re-ran the exact invocations from my finding, `--objective blake-3-compressions --log-trace-length 5`:

```
le=2048  -> Merkle leaf: 2048 field elements (32768 bytes)                                            Proof size: 529698
le=4096  -> Merkle leaf: 4096 field elements (65536 bytes)                                            Proof size: 529824
le=8192  -> Merkle leaf: 4096 field elements (65536 bytes; requested 8192, capped ...)                Proof size: 529824
le=65536 -> Merkle leaf: 4096 field elements (65536 bytes; requested 65536, capped ...)               Proof size: 529824
le=2^60  -> Merkle leaf: 4096 field elements (65536 bytes; requested 1152921504606846976, capped ...)  Proof size: 529824
```

Identical proof bytes now print identical leaves, and the clamp is named. That last row also closes the overflow MINOR: `0 bytes` is gone in release and the debug build no longer panics.

I also checked your claim that no row of the published table was wrong. At the tightest case, Keccak-f `-l 14`, leaves of 16/64/128 all report unclamped (495119 / 837777 / 1238012 bytes). Confirmed.

**The vacuous wide-leaf assertion** — fixed the right way, raised rather than loosened. `a_boolean_trace_proves_under_either_hash` runs at `log_height = 12`: 3 columns x 4096 rows stack to 14 variables, 7 absorbed, base message 128 symbols, so the requested 64 (8x the default fold coset of 8) is packed whole and the assertion means what it says. The clamped case did not vanish with it — I mutated `leaf_elements_of` back to an unclamped value and `a_leaf_wider_than_the_base_message_reports_the_cap` is the one test that fails (`left: 1048576, right: 8`), with the other 20 still green. The new test bites.

### The four MINORs

- `grouped_mmcs` returns `Result<_, BinaryProofError>` and rejects a non-power-of-two before `with_group_size` sees it, so `binary_config`/`boolean_config` return `UnsupportedLeafElements` where they used to panic; `check_leaf_elements` is gone, so there is one check and it cannot drift. Keeping `with_group_size`'s assertion as the `for_folding` invariant is the right call.
- `COLLISION_RESISTANCE_BITS` default dropped; both impls state 128 for themselves.
- `HarnessConfig::leaf_elements` removes the threaded parameter, the two call sites and the `#[allow(clippy::too_many_arguments)]`.

`cargo test -p p3-examples --lib binary::` passes 21/21 in both debug and release; `cargo clippy -p p3-examples -p p3-binary-pcs --all-targets` is clean.

### Still open from my earlier body, none blocking

`HashFamily::fmt` still prints `blake3` where the CLI's canonical name is `blake-3`; `Clone` is still redundant next to `Copy` on `HarnessHash`; `#[derive(Default)]` on `HashFamily` is still dead; `BinaryCommitmentHashOptions` variants are still undocumented; `(merkle_arity = 4, Blake3)` — the CLI default paired with the hash every benchmark row used — still has no test on either entry point. All NIT-class, take them or leave them.

PR #2251, opened 2026-09-20
perf(sumcheck): factor the ring-switch equality weights into scaled blocks

Nashtare:
Stacked on #2253, which moves this reduction's *rounds* into the
polynomial basis; this one removes its *construction* cost. The two touch disjoint code.

`support()` built the equality table of the supported run densely: `2^k − 1` tower multiplies and
a `2^k`-element write, 512 MB at BLAKE3 `-l 18`. Every consumer then read it back.

An equality table factors. Held as an outer weight per block times one inner factor every block
shares, `table[i*block + j] == outer[i] * inner[j]`, so the dense table need never exist:
`tensor_over`, `successor_tensors_over` and `weights_over` sweep a block at a time and absorb that
block's outer weight **once** — as a first-leg scaling of the block's partial tensor element, or as
a rescale of the coordinate sums through `scaled_sums`, which rebuilds a `CoordinateSums` such that
`scaled(sums, c).sum(x) == sums.sum(c * x)`.

Also here: `BitTensor::add_scaled_columns` expresses a first-leg scaling as `sum_u (a·beta_u) ⊗ row_u`,
which is `d` multiplies and no bit transposes — `scale_columns` previously did two 128×128
transposes. The `alpha` successor combination now lands in the table it reads instead of
allocating a second one, and `restricted_packing` copies in chunks into a preallocated buffer.

### Multiplications, counted

An atomic counter in `BinaryField128::dispatched_mul`, read around each phase of
`BitRingSwitch::prove` (local instrumentation, not committed):

| config | `k` | before | after | ratio |
| BLAKE3 `-l 14` | 21 | 2,097,151 | 44,542 | 47x |
| BLAKE3 `-l 16` | 23 | 8,388,607 | 128,322 | 65x |
| Keccak-f `-l 16` | 20 | 2,097,151 | 1,088,318 | 1.9x |

The construction goes from `2^k` multiplies to `(2^(k−m) − 1) + (2^m − 1)` on the two factors plus
`blocks × d` on the tensor scalings and the rescaled sums, with `m = 14` and `d = 128`.
Extrapolated to BLAKE3 `-l 18` (`k = 25`) that is 33.5M -> ~464K.

Keccak-f only reaches 1.9x because `weights_over`'s `alpha · base[w−1]` shift still costs one
multiply per entry. A second coordinate-sums table prescaled by `alpha` would remove it, at one
multiply traded for sixteen table lookups per entry — unmeasured either way, so not attempted.

### Measurements

c7i.4xlarge (Xeon 8488C, 16 vCPU), `-C target-cpu=native`, `--features parallel`, default PCS
flags, medians of 3, **`CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`** (see below).

| | prove | `bit ring switch` |
| --- | --- | --- |
| BLAKE3 `-l 18` | 3120 -> 3000 ms (−3.8%) | 447 -> 352 ms (−21.3%) |
| Keccak-f `-l 18` | 847 -> 831 ms (−1.9%) | 82.6 -> 70.7 ms (−14.4%) |
| Keccak-f `-l 14` | 94.4 -> 92.9 ms (−1.6%) | 10.4 -> 9.41 ms (−9.5%) |

`binary pcs open`, `into_sumcheck_in` and `combine_weights_in` are flat, as expected — the diff
does not reach them. A base-against-base noise run puts spans of a few tens of ms at ±6-9% and the
440 ms ring-switch span under 1%.

After the change the 352 ms is: rounds 233, `tensor_over` 57.0, `weights_over` 29.7,
`restricted_packing` 20.6, `support` **0.54**. The construction is now 30% of the reduction and
the rounds are 66% — which is what #2253 addresses.

`LOG_CHUNK` is 14. At 12 the span is 373.5 ms; at 16 it is 348.0 against 348.5, buying nothing
while leaving only four blocks for sixteen threads at Keccak-f `-l 14`.

### A measurement trap worth knowing

At the default sixteen codegen units, three separate A/B runs showed `commit_matrix` moving
71 -> 117 ms in code this diff never touches. `commit_matrix` lives in the same crate, so adding a
module repartitions the codegen units and re-optimises it. With one codegen unit the swing vanishes
(71-74 ms on both sides) and the total flips from +1.0% to −3.8%. **Any diff inside `p3-sumcheck`
measured at default codegen units can carry a ±50 ms phantom there, 1.6% of BLAKE3 prove.**

### Correctness

Prover-only except for `BitTensor::scale_columns`, which the verifier reaches through
`equality_element` -> `closing_weight` -> `verify_readings`; everything else the diff touches runs
on the prover alone. Proofs are byte-identical to #2246 at `-l 14` and `-l 16` for both
objectives, and proof sizes match at `-l 18`. The factored entry and the dense entry are the same per-coordinate factors
grouped differently, and multiplication is associative, so the tests assert structural equality
rather than relying on verification: `the_factors_multiply_to_the_dense_table` compares
`materialize()` against `Poly::new_from_point` element for element over 0-9 variables × every block
size; `the_successor_weights_are_the_dense_table_entry_by_entry` rebuilds the whole weight table,
shift and `alpha²` term included, from the dense table. Under `cfg(test)` `LOG_CHUNK` is 1, so a two-entry block is all the tests would otherwise see — a
block where `weights_over`'s stride collapses onto a single column and `successor_tensors_over` is
all boundary and no interior. `the_sweeps_read_the_same_table_at_every_block_size` covers that
instead: it pins `tensor_over`, `successor_tensors_over` and `weights_over` against the
one-entry-per-block table at every block size from `2^0` to past the run, so the shipped `2^14`
block is the same program the tests check. `the_columns_scale_at` runs the rewritten
`scale_columns` at `BinaryField128` as well as `BinaryField16`.

`cargo test --workspace --features parallel` passes, as does `-p p3-sumcheck` with no features
(the serial shim, where `par_fold_reduce` never calls `reduce_op`). Clippy and rustfmt clean.

### Negative result

Factoring *without* prescaling is a regression: generating `outer[i] * inner[j]` per entry kills
the table but pays one multiply per entry in both `tensor_over` and `weights_over`, i.e. `2^(k+1)`
against the base's `2^k`. BLAKE3 `-l 18` measured 440 -> 515 ms, +17.0%. Tower multiplication is
expensive enough that 512 MB of avoided traffic does not pay for one extra multiply per entry. The
win only exists once the scale leaves the inner loop.

### Largest thing left here, not done

`weight_plan` inside `into_sumcheck_in` is 76-81 ms at BLAKE3 `-l 18`, 2.6% of prove. It
materialises `eq(surviving_point, ·)` over the full stacked space in `EF` (512 MB, one tower
multiply per entry), and `combine_weights_in` then converts every entry into `R` in a second 512 MB
buffer before `sole_whole_space_table` hands it straight through. Building it directly in `R` is
sound by the isomorphism argument `combine_weights_in` already documents, but the materialisation
is `SplitEq::accumulate_into` in `multilinear-util/`, `EqMaybePacked` exposes no scalar accessor,
and `SvoPoint` does not keep the claim's raw point — so it needs `weight_plan`, `ClaimWeightTables`,
`WeightPlan` and `shared_row_weights` made generic in the table field. A separate change.


tcoratger (commented):
Reviewed against the stack base (`perf/ring-switch-poly-basis`). I derived the factoring identity by hand, checked every consumer against it, and stress-tested the block split by sweeping the test-mode block size across seven values — all pass.

**No blockers; the math is right and the change is prover-side in effect.** Two things need attention before merge: the PR body's "prover-only" claim is wrong in a way that matters for review attention, and the block size the code actually ships (`2^14`) is never exercised by any test (the test build uses `2`).

## The factoring identity

The equality polynomial is a product over coordinates, and `Poly::new_from_point` lays the table out with the **first** coordinate most significant. So splitting the point after its leading coordinates splits the table into equal, contiguous blocks:

```
    point   =  ( r_0 ... r_{k-m-1} | r_{k-m} ... r_{k-1} )
                \---- outer ----/   \----- inner -----/
                  block index i       within-block j

    table[i * 2^m + j]  ==  eq(outer, i) * eq(inner, j)
                        ==  outer[i] * inner[j]
```

`equality.rs:47` splits at `point.len() - min(log_block, point.len())`, i.e. the **inner factor takes the trailing coordinates**. That is the correct end: the trailing coordinates are the ones that vary within a contiguous run of `2^m` indices. Getting this backwards is the classic bug and the code does not have it — `the_factors_multiply_to_the_dense_table` pins it over 0..10 variables × every block size.

Note this split is **not** the ring-switch kappa. Kappa here is `ABSORBED = Coefficients::<EF>::LOG_DIMENSION` (7 for `BinaryField128`), which was already consumed before `high()` is formed. `LOG_CHUNK` is purely a cache/parallelism tuning knob over the *remaining* coordinates, so there is no off-by-one-against-kappa risk to check. Good separation.

The three consumers rely on three sub-identities, all of which I verified:

**1. Tensor leg carrying (`reduction.rs:483`).** In `EF ⊗_{F_2} EF`, `(a ⊗ 1) * (b ⊗ c) = (ab) ⊗ c`, and `⊗` is `F_2`-bilinear, so

```
    sum_j (outer_i * inner_j) ⊗ t'(w)  ==  (outer_i ⊗ 1) * sum_j inner_j ⊗ t'(w)
```

`BitTensor` stores `element = sum_u beta_u ⊗ row_u`, so `(a ⊗ 1) * element = sum_u (a * beta_u) ⊗ row_u` — exactly what `add_scaled_columns` (`tensor.rs:164`) computes. Correct, and it is `d` multiplies instead of a pair of `d × d` bit transposes.

**2. Coordinate-sum rescaling (`equality.rs:137`).** `sums.sum(.)` is `F_2`-linear and so is multiplication by a fixed scale, so

```
    scaled(sums, c).sum(x) = sum_{u in x} sums.sum(c * beta_u)
                           = sums.sum(c * sum_{u in x} beta_u)
                           = sums.sum(c * x)
```

which makes `scaled.sum(inner_j) == sums.sum(outer_i * inner_j)` — bit-identical to the old `sums.sum(dense[w])`. Correct.

**3. Block-boundary carry (`reduction.rs:578`).** The successor sweep reads `table[w-1]`. Inside a block that is `outer_i * inner_{j-1}`, which factors; at `j == 0` it is `outer_{i-1} * inner_{2^m - 1}`, which does **not** share the block's scale, so it is added unscaled via `before_block` (`equality.rs:91`) directly into `carry` rather than into the buckets. This is the one place the factoring genuinely leaks, and it is handled correctly — including when `block < column` (a block starting mid-column, `row != 0` at `j == 0`), which happens whenever `kept > LOG_CHUNK`.

**Normalization / boundaries.** `kappa = 0` is not reachable here (`kept_row_variables` filters `rows > ABSORBED`). Zero variables gives a single unit entry (tested, `equality.rs:244`). A run shorter than one block collapses `outer` to `[ONE]` (tested, `equality.rs:187`). `before_block(0) = ZERO` is a sentinel that is never consumed, because `w = 0` always has `row == 0`.

**Verdict on the central question: this is a pure regrouping, not a value change.** Multiplication is associative and the code groups the same per-coordinate factors differently. The verifier needs no matching change because the verify path (`reduction.rs:1344-1368`) only calls `initial_sum` and `closing_weight`, neither of which touches `support()`. I confirmed that by tracing every caller of `support()` and `weights_over`.

**Structural safety of the in-place alpha combination (`reduction.rs:894-906`) — worth spelling out, because it is the subtlest line in the diff.** The combination `w <- A(w) + alpha*A(w-1) + [row==max] alpha^2*A(w)` now writes into the table it reads. It is safe because the loop descends (`for row in (0..column).rev()`): when `entries[row]` is written, `entries[row-1]` has not been touched yet, and `entries[row]`'s own old value is saved in `settled` first. The chunking is also sound, but only via a three-function argument: `stride = CHUNK.max(column).min(num_evals)` is a power of two and `column <= num_evals` holds because `prefix_limit()` (`reduction.rs:315`) caps the Boolean prefix at `num_variables - kept` exactly when a successor view is present, and `alpha.is_some()` implies `sends_successor_tensors()` via `batch_with_successor`. So `column | stride | num_evals`, and `part.chunks_mut(column)` never yields a short tail. If that chain ever breaks the result is an out-of-bounds panic, not a wrong proof — but see the MINOR on debug asserts.

---

## [MAJOR] The rewritten `scale_columns` runs on the verifier path but is only tested at `BinaryField16` — `sumcheck/src/ring_switch/bits/tensor.rs:145`

The PR body says "Prover-only; proofs byte-identical". That is not true of `tensor.rs`. `scale_columns` was reimplemented (transpose/scale/transpose -> `add_scaled_columns`), and `scale_columns` is called from `BitRingSwitch::equality_element` (`reduction.rs:742,745`), which `closing_weight` calls, which `verify` calls at `reduction.rs:1368`. A divergence there is a completeness break at best and, since `closing_weight` is what ties the sumcheck's final claim to the weight multilinear, a soundness-relevant one at worst.

The only semantic test of the new implementation, `scaling_the_columns_scales_every_column_reading` (`tensor.rs:495`), runs at `type EF = BinaryField16` (`tensor.rs:372`) — `d = 16`. Production is `BinaryField128`, `d = 128`. The rewrite changes which basis elements get multiplied and how coordinates are indexed, so the field width is exactly the axis that should be covered.

The body's framing matters beyond accuracy: a reviewer who reads "prover-only" will not think to check `equality_element`.

```rust
pub fn scale_columns(&mut self, a: EF) {
    let mut scaled = Self::zero();
    scaled.add_scaled_columns(self, a);   // no longer the transpose path
    *self = scaled;
}
```

**Fix:** reword the body to "prover-only except for the `scale_columns` reimplementation, which the verifier reaches through `closing_weight`", and make `scaling_the_columns_scales_every_column_reading` generic over the level, instantiating it at both `BinaryField16` and `BinaryField128` (the file already imports `BinaryField128` for `the_matrix_is_square_in_the_fields_dimension`).

**Verify:** I ran the differential myself — the pre-PR body of `scale_columns` against the new one, over 200 random `BinaryField128` elements plus `a = ZERO`, `a = ONE`, the zero element and `one()`. **All 203 cases are bit-identical.** So this is a documentation-and-coverage gap, not a live bug; the test just needs to exist so the next rewrite is caught.

---

## [MAJOR] The shipped block size is never exercised: `cfg(test)` pins it to 2, and that makes one loop structurally dead — `sumcheck/src/ring_switch/bits/reduction.rs:44`

`LOG_CHUNK` is `14` in production and `1` under `cfg(test)`. Every correctness test therefore runs with `block_len == 2`. Two consequences:

- In `weights_over`, `stride = CHUNK.max(column).min(num_evals)`. Under test `CHUNK = 2` and `column = 2^kept >= 2` (since `kept >= 1`), so `stride == column` **always**, and `for entries in part.chunks_mut(column)` has trip count exactly 1 in every test. The multi-column-per-stride-chunk path — the one production takes, where `stride = 2^14` covers `2^14 / column` columns — is never entered.
- The `j == 0` boundary in `successor_tensors_over` fires on *every* block at `block_len == 2`, which is the opposite of the production regime where it fires once per `2^14` entries. The ratio of boundary path to interior path is inverted, so a bug in the interior arithmetic would be masked by the boundary dominating.

This is the classic shape of a refactor that tests a different program than it ships.

**Fix:** the three sweeps already take `&FactoredEquality` as a parameter, so a test can build one at any block size with no production change. Add a block-size-invariance test:

```rust
#[test]
fn the_weights_match_at_every_block_size() {
    // Invariant: the factored sweep does not depend on where the blocks fall.
    for &row_variables in &ROW_VARIABLES {
        for selector in Selector::ALL {
            let r = successor_point(0xB10CD, 9, row_variables, selector);
            let reduction = BitRingSwitch::with_successor(&r, row_variables).unwrap();
            if !reduction.sends_successor_tensors() { continue; }
            let batch = reduction.batch_with_successor(&r_batch, alpha).unwrap();
            let (prefix, _) = reduction.fixed_prefix();
            let run = &reduction.point.as_slice()[prefix..reduction.num_variables()];

            let reference = batch.weights_over::<EF>(&FactoredEquality::new(run, 0));
            for log_block in 0..=run.len() + 2 {
                assert_eq!(
                    batch.weights_over::<EF>(&FactoredEquality::new(run, log_block)).as_slice(),
                    reference.as_slice(),
                    "{selector:?}, rows {row_variables}, log_block {log_block}",
                );
            }
        }
    }
}
```

**Verify:** I wrote exactly this test into a worktree and it compiles and passes (then reverted). I also swept the `cfg(test)` `LOG_CHUNK` over `{0, 2, 3, 4, 5, 8}` and re-ran all 136 `ring_switch` tests at each: **all pass**, including the values where `stride > column` and where `block > column`. So the implementation is correct across the parameterization — the tests just do not say so, and would not notice if it stopped being true.

---

## [MINOR] `weights_over`'s first pass lost most of its parallelism on small runs — `sumcheck/src/ring_switch/bits/reduction.rs:871`

The pass went from `par_iter_mut()` over all `2^k` entries to `par_chunks_mut(2^14)`, which is `2^(k-14)` tasks. For any run with `k <= 14` that is a single task — fully serial where it used to be fully parallel. A claim about one column of a stacked trace has `k = log(rows) - 7` after the Boolean prefix is stripped, so `k <= 14` covers traces up to `2^21` rows. The absolute cost is small at those sizes, which is why the benchmarks do not show it, but it is a real cliff and it is invisible from the call site.

**Fix:** decouple the *scale-amortization* unit from the *parallelism* unit — chunk the rayon split at `block_len.min(num_evals >> log_threads)`, or leave it and add a comment recording that the run is assumed large, since the same concern applies to `tensor_over` and `restricted_packing`.

**Verify:** benchmark `bit ring switch` at Keccak-f `-l 10` / `-l 11`, which the posted table does not cover (it starts at `-l 14`).

---

## [MINOR] The reduce identity allocates a throwaway 64 KB bucket array per join — `sumcheck/src/ring_switch/bits/reduction.rs:487,587`

`par_fold_reduce` passes the *same* closure to both `fold` and `reduce` (`maybe-rayon/src/lib.rs:59`). The accumulator is now `(BitTensor, BitTensorBuckets)`, and `BitTensorBuckets::zero()` is `NUM_BYTES * 256` field elements — 64 KB at `BinaryField128`. Every `reduce` seed allocates one purely to have its buckets dropped on the next line. At `2^11` blocks that is up to a few thousand wasted 64 KB allocations per reduction.

```rust
|(mut total, buckets), (partial, _)| {   // `_` is the discarded 64 KB
    total += partial;
    (total, buckets)
},
```

**Fix:** make the scratch lazy — accumulator `(BitTensor, Option<BitTensorBuckets>)` with `get_or_insert_with(BitTensorBuckets::zero)` in the fold op. The `Option` also makes the "the reduce arm's buckets are meaningless" fact type-visible instead of a bare `_`.

---

## [MINOR] `scaled_sums` rebuilds a fresh 64 KB subset-sum table per block — `sumcheck/src/ring_switch/bits/equality.rs:137`

Each call allocates a `Vec<[A; 256]>` of `NUM_BYTES` entries and refills it with `NUM_BYTES * 255` additions. Correct, and it amortizes fine at `2^14` entries per block, but it is one allocation per block on a pass the body measures at 29.7 ms, and the allocation is trivially reusable within a rayon task.

**Fix:** add `scaled_sums_into(&mut CoordinateSums<EF, A>, &CoordinateSums<EF, A>, EF)` writing through a scratch carried by the fold, mirroring the bucket scratch the tensor sweeps already carry. Keep `scaled_sums` as the allocating wrapper for the test.

---

## [MINOR] The `stride % column == 0` invariant is derived three functions away and is not asserted — `sumcheck/src/ring_switch/bits/reduction.rs:894`

The in-place alpha pass indexes `entries[row]` for `row in 0..column` on a slice produced by `part.chunks_mut(column)`. That never yields a short tail only because `column <= table.num_evals()`, which holds only because `prefix_limit()` caps the Boolean prefix at `num_variables - kept` whenever a successor view is present. Nothing at the indexing site records this, and the failure mode is a panic inside a rayon closure.

**Fix:**
```rust
debug_assert!(column.is_power_of_two() && stride.is_multiple_of(column));
debug_assert!(table.num_evals().is_multiple_of(stride));
```
plus a one-line comment pointing at `prefix_limit` as the reason `column <= num_evals`.

---

## [MINOR] `restricted_packing` now depends on `EF::zero_vec` being calloc-backed — `sumcheck/src/ring_switch/bits/reduction.rs:400`

`EF::zero_vec(len)` + parallel copy replaces `par_iter().copied().collect()`. This is faster only because every `TowerLevel` impl overrides `zero_vec` with the `ManuallyDrop`/`from_raw_parts` trick over `vec![0u128; len]` (`binary-field/src/ghash.rs:216`, `binary-field/src/tower.rs:413`), which hits the `IsZero` specialization and so `alloc_zeroed`. The `PrimeCharacteristicRing` default (`field/src/field.rs:401`) is a serial `vec![Self::ZERO; len]`, which at `-l 18` would be a serial 512 MB write before the parallel copy. Worth a comment, since the generic bound is `EF: TowerLevel` and nothing enforces the override.

---

## [NIT]

- `adding_a_scaled_element_scales_its_column_reading` (`tensor.rs:511`) is circular: it compares `add_scaled_columns` against `scale_columns`, but `scale_columns` is now *implemented as* `add_scaled_columns` into a zero element. The test proves `zero + x == x` and additivity, not the scaling semantics. Assert against `columns()` directly, the way `scaling_the_columns_scales_every_column_reading` does.
- Stale comment: `tensor.rs:498` still says "The stored rows are the other reading, so this is the transpose path" — there is no transpose path any more.
- `add_scaled_columns` is `pub` but used only in-crate (`tensor.rs:164`). `BitTensor` is a wire type with a deliberately small public surface; `pub(crate)` fits better. Also `BitTensorBuckets::tensor` (`tensor.rs:292`) allocates and returns and should carry `#[must_use]`.
- `drop(equality)` is now vestigial (`reduction.rs:1235`): it existed to free the 512 MB dense table before the sumcheck; `FactoredEquality` is a few hundred KB at most. Either delete it or comment why it stays.
- The doc on `add_scaled_columns` (`tensor.rs:162`) says "one multiplication per coordinate", which undersells the cost: it is `d` multiplications **plus** roughly `d^2 / 2` field additions from the `add_exterior_product` fan-out. Still a win over two `d × d` transposes plus four allocations, but the comment reads as if the additions are free.
- The head branch is `perf/opening-share`, which describes nothing in this diff — I checked all four changed files and there is no opening-sharing change. Harmless, but it makes the stack hard to follow.
- There are now two modules named `equality` under `ring_switch` (`ring_switch/equality.rs` and `ring_switch/bits/equality.rs`). Consider `factored_equality.rs` for the new one.

---

## Test coverage

What already exists and is good: `the_factors_multiply_to_the_dense_table` (0..10 variables × every block size) is the right shape for the core identity; `the_successor_elements_are_their_term_by_term_sums` rebuilds both elements from the dense table and loops over `ROW_VARIABLES × Selector::ALL`; `scaled_sums_read_the_scaled_value` pins the linearity argument including `scale = ZERO` and `scale = ONE`.

Gaps, in priority order:

1. **Block size is a compile-time constant, so the shipped value is untested.** Covered under the MAJOR above — add the block-size-invariance test (written and verified passing) so the `stride > column` path stops being dead code in the test build.

2. **`the_successor_weights_are_the_dense_table_entry_by_entry` tests exactly one configuration** — `reduction.rs:2321`, `row_variables = 6`, `Selector::Random`, `num_variables = 9`. Its sibling element test two functions above loops over `ROW_VARIABLES × Selector::ALL`; this one should too. The missing configurations are the interesting ones: a Boolean or Mixed selector gives a nonzero prefix and shrinks the run (possibly to `num_evals == column`, a single column chunk), and `row_variables = 9` gives `kept = 5` with `column == num_evals`.

3. **`scale_columns` at `BinaryField128`** — covered under the MAJOR above.

4. **A proptest is a good fit for the factoring itself**, better than more fixed seeds:

```rust
proptest! {
    #[test]
    fn the_factored_table_is_the_dense_table(
        num_variables in 0usize..10,
        log_block in 0usize..12,
        seed: u64,
    ) {
        let point = Point::<BinaryField128>::rand(&mut SmallRng::seed_from_u64(seed), num_variables);
        let dense = Poly::new_from_point(point.as_slice(), F::ONE);
        prop_assert_eq!(
            FactoredEquality::new(point.as_slice(), log_block).materialize().as_slice(),
            dense.as_slice()
        );
    }
}
```

A failure here names the exact `(num_variables, log_block)` pair, which pins whether the split consumed the leading or trailing coordinates. `tensor.rs` already depends on `proptest`, so no new dependency. Paper known-answer vectors are *not* the right tool here — this is an internal regrouping of an implementation detail, not a spec-level quantity, so a differential against the dense computation is strictly stronger.

5. **Not a gap, but worth recording in the PR body:** the multiplication accounting checks out against the posted counter. At BLAKE3 `-l 14` (`k = 21`, `m = 14`): factors `2^7 - 1 + 2^14 - 1 = 16510`, plus `2^7 * 128 = 16384` each in `tensor_over` and `weights_over`, totalling ~49.3K against the measured 44.5K — the gap is `add_scaled_columns` skipping zero rows. Before is `2^21 - 1 = 2,097,151`, matching the table exactly. The claimed 47x is real.

---

## Could not verify

- **Benchmarks.** I did not reproduce the c7i.4xlarge numbers. The codegen-units caveat in the body is a genuinely useful finding and I have no reason to doubt the medians, but the `-l 10` / `-l 11` regime flagged in the parallelism MINOR is not in the posted table and I could not measure it here.
- **What `R` is after #2253.** I confirmed the basis question is safe *structurally*: the scale multiplication `scale * beta_u` happens entirely in `EF` on the index side, and `R` only ever receives additions and the `alpha` multiplications that were already there. So `scaled_sums` is basis-agnostic by construction and cannot carry a stale-basis weight. I did not independently re-verify that `R: From<EF>` is a field isomorphism — that is #2253's contract, asserted in the `prove` doc comment, and outside this diff.

---

**Bottom line:** merge after fixes — the two MAJORs are a doc correction plus two tests, neither of which changes production code, and I have already verified that both underlying behaviours are correct. The factoring itself is sound and I would merge it as-is if the test debt were being tracked.

sumcheck/src/ring_switch/bits/tensor.rs
```diff
@@ -142,14 +142,33 @@ impl<EF: TowerLevel> BitTensor<EF> {
     /// Scales the column reading: column `v` becomes `a * column v`.
     ///
     /// This is multiplication by `a ⊗ 1`, acting on the first tensor leg.
-    /// The stored rows are the wrong reading for it.
-    /// So the matrix is transposed, scaled, and transposed back.
     pub fn scale_columns(&mut self, a: EF) {
```
tcoratger, line 145: **[MAJOR] The rewritten `scale_columns` runs on the verifier path but is only tested at `BinaryField16` — `sumcheck/src/ring_switch/bits/tensor.rs:145`**

The PR body says "Prover-only; proofs byte-identical". That is not true of `tensor.rs`. `scale_columns` was reimplemented (transpose/scale/transpose -> `add_scaled_columns`), and `scale_columns` is called from `BitRingSwitch::equality_element` (`reduction.rs:742,745`), which `closing_weight` calls, which `verify` calls at `reduction.rs:1368`. A divergence there is a completeness break at best and, since `closing_weight` is what ties the sumcheck's final claim to the weight multilinear, a soundness-relevant one at worst.

The only semantic test of the new implementation, `scaling_the_columns_scales_every_column_reading` (`tensor.rs:495`), runs at `type EF = BinaryField16` (`tensor.rs:372`) — `d = 16`. Production is `BinaryField128`, `d = 128`. The rewrite changes which basis elements get multiplied and how coordinates are indexed, so the field width is exactly the axis that should be covered.

The body's framing matters beyond accuracy: a reviewer who reads "prover-only" will not think to check `equality_element`.

```rust
pub fn scale_columns(&mut self, a: EF) {
    let mut scaled = Self::zero();
    scaled.add_scaled_columns(self, a);   // no longer the transpose path
    *self = scaled;
}
```

**Fix:** reword the body to "prover-only except for the `scale_columns` reimplementation, which the verifier reaches through `closing_weight`", and make `scaling_the_columns_scales_every_column_reading` generic over the level, instantiating it at both `BinaryField16` and `BinaryField128` (the file already imports `BinaryField128` for `the_matrix_is_square_in_the_fields_dimension`).

**Verify:** I ran the differential myself — the pre-PR body of `scale_columns` against the new one, over 200 random `BinaryField128` elements plus `a = ZERO`, `a = ONE`, the zero element and `one()`. **All 203 cases are bit-identical.** So this is a documentation-and-coverage gap, not a live bug; the test just needs to exist so the next rewrite is caught.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -28,18 +29,22 @@ use crate::data::SumcheckData;
...
 #[cfg(not(test))]
-const CHUNK: usize = 1 << 12;
+const LOG_CHUNK: usize = 14;
 
 /// Small enough under test that a case above one variable splits in chunks.
 ///
 /// The partial-combination fold is then on the path the tests take.
 #[cfg(test)]
-const CHUNK: usize = 1 << 1;
+const LOG_CHUNK: usize = 1;
```
tcoratger, line 43: **[MAJOR] The shipped block size is never exercised: `cfg(test)` pins it to 2, and that makes one loop structurally dead — `sumcheck/src/ring_switch/bits/reduction.rs:44`**

`LOG_CHUNK` is `14` in production and `1` under `cfg(test)`. Every correctness test therefore runs with `block_len == 2`. Two consequences:

- In `weights_over`, `stride = CHUNK.max(column).min(num_evals)`. Under test `CHUNK = 2` and `column = 2^kept >= 2` (since `kept >= 1`), so `stride == column` **always**, and `for entries in part.chunks_mut(column)` has trip count exactly 1 in every test. The multi-column-per-stride-chunk path — the one production takes, where `stride = 2^14` covers `2^14 / column` columns — is never entered.
- The `j == 0` boundary in `successor_tensors_over` fires on *every* block at `block_len == 2`, which is the opposite of the production regime where it fires once per `2^14` entries. The ratio of boundary path to interior path is inverted, so a bug in the interior arithmetic would be masked by the boundary dominating.

This is the classic shape of a refactor that tests a different program than it ships.

**Fix:** the three sweeps already take `&FactoredEquality` as a parameter, so a test can build one at any block size with no production change. Add a block-size-invariance test:

```rust
#[test]
fn the_weights_match_at_every_block_size() {
    // Invariant: the factored sweep does not depend on where the blocks fall.
    for &row_variables in &ROW_VARIABLES {
        for selector in Selector::ALL {
            let r = successor_point(0xB10CD, 9, row_variables, selector);
            let reduction = BitRingSwitch::with_successor(&r, row_variables).unwrap();
            if !reduction.sends_successor_tensors() { continue; }
            let batch = reduction.batch_with_successor(&r_batch, alpha).unwrap();
            let (prefix, _) = reduction.fixed_prefix();
            let run = &reduction.point.as_slice()[prefix..reduction.num_variables()];

            let reference = batch.weights_over::<EF>(&FactoredEquality::new(run, 0));
            for log_block in 0..=run.len() + 2 {
                assert_eq!(
                    batch.weights_over::<EF>(&FactoredEquality::new(run, log_block)).as_slice(),
                    reference.as_slice(),
                    "{selector:?}, rows {row_variables}, log_block {log_block}",
                );
            }
        }
    }
}
```

**Verify:** I wrote exactly this test into a worktree and it compiles and passes (then reverted). I also swept the `cfg(test)` `LOG_CHUNK` over `{0, 2, 3, 4, 5, 8}` and re-ran all 136 `ring_switch` tests at each: **all pass**, including the values where `stride > column` and where `block > column`. So the implementation is correct across the parameterization — the tests just do not say so, and would not notice if it stopped being true.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -831,11 +863,19 @@ impl<EF: TowerLevel> BitRingSwitchBatch<'_, EF> {
...
         // No zero run for another slot is allocated or folded.
+        //
+        // A block's entries are its own weights under one outer weight, and the coordinate
+        // sums absorb that scale once per block instead of once per entry.
         table
             .as_mut_slice()
-            .par_iter_mut()
-            .zip(equality.as_slice().par_iter())
-            .for_each(|(slot, &value)| *slot = sums.sum(value));
+            .par_chunks_mut(equality.block_len())
```
tcoratger, line 879: **[MINOR] `weights_over`'s first pass lost most of its parallelism on small runs — `sumcheck/src/ring_switch/bits/reduction.rs:871`**

The pass went from `par_iter_mut()` over all `2^k` entries to `par_chunks_mut(2^14)`, which is `2^(k-14)` tasks. For any run with `k <= 14` that is a single task — fully serial where it used to be fully parallel. A claim about one column of a stacked trace has `k = log(rows) - 7` after the Boolean prefix is stripped, so `k <= 14` covers traces up to `2^21` rows. The absolute cost is small at those sizes, which is why the benchmarks do not show it, but it is a real cliff and it is invisible from the call site.

**Fix:** decouple the *scale-amortization* unit from the *parallelism* unit — chunk the rayon split at `block_len.min(num_evals >> log_threads)`, or leave it and add a comment recording that the run is assumed large, since the same concern applies to `tensor_over` and `restricted_packing`.

**Verify:** benchmark `bit ring switch` at Keccak-f `-l 10` / `-l 11`, which the posted table does not cover (it starts at `-l 14`).

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -452,35 +458,48 @@ impl<EF: TowerLevel> BitRingSwitch<EF> {
...
-            .zip(values.par_chunks(CHUNK))
+        values
+            .par_chunks(equality.block_len())
+            .zip(equality.outer().par_iter())
             .par_fold_reduce(
-                BitTensorBuckets::zero,
-                |mut accumulator, (weights, values)| {
-                    for (&weight, &value) in weights.iter().zip(values) {
-                        accumulator.add_exterior_product(weight, value);
+                || (BitTensor::zero(), BitTensorBuckets::zero()),
```
tcoratger: **[MINOR] The reduce identity allocates a throwaway 64 KB bucket array per join — `sumcheck/src/ring_switch/bits/reduction.rs:487,587`**

`par_fold_reduce` passes the *same* closure to both `fold` and `reduce` (`maybe-rayon/src/lib.rs:59`). The accumulator is now `(BitTensor, BitTensorBuckets)`, and `BitTensorBuckets::zero()` is `NUM_BYTES * 256` field elements — 64 KB at `BinaryField128`. Every `reduce` seed allocates one purely to have its buckets dropped on the next line. At `2^11` blocks that is up to a few thousand wasted 64 KB allocations per reduction.

```rust
|(mut total, buckets), (partial, _)| {   // `_` is the discarded 64 KB
    total += partial;
    (total, buckets)
},
```

**Fix:** make the scratch lazy — accumulator `(BitTensor, Option<BitTensorBuckets>)` with `get_or_insert_with(BitTensorBuckets::zero)` in the fold op. The `Option` also makes the "the reduce arm's buckets are meaningless" fact type-visible instead of a bare `_`.

sumcheck/src/ring_switch/bits/equality.rs
```diff
@@ -0,0 +1,269 @@
...
+///     scaled(sums, scale).sum(x)  ==  sums.sum(scale * x)   for every x
+/// ```
+///
+/// So a sweep whose values all carry the same scale multiplies once per coordinate, however
+/// many values it then reads.
+///
+/// The scale is a level element, and the weights are whatever the sums hold, so a sweep
+/// whose sums already live in another representation stays there.
+#[must_use]
+pub(crate) fn scaled_sums<EF: TowerLevel, A: Field>(
```
tcoratger: **[MINOR] `scaled_sums` rebuilds a fresh 64 KB subset-sum table per block — `sumcheck/src/ring_switch/bits/equality.rs:137`**

Each call allocates a `Vec<[A; 256]>` of `NUM_BYTES` entries and refills it with `NUM_BYTES * 255` additions. Correct, and it amortizes fine at `2^14` entries per block, but it is one allocation per block on a pass the body measures at 29.7 ms, and the allocation is trivially reusable within a rayon task.

**Fix:** add `scaled_sums_into(&mut CoordinateSums<EF, A>, &CoordinateSums<EF, A>, EF)` writing through a scratch carried by the fold, mirroring the bucket scratch the tensor sweeps already carry. Keep `scaled_sums` as the allocating wrapper for the test.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -844,29 +884,33 @@ impl<EF: TowerLevel> BitRingSwitchBatch<'_, EF> {
...
         let alpha = R::from(alpha);
         let alpha_squared = alpha.square();
 
-        // The shift reads the entry before, so the combined table is a second one.
-        let base = table.as_slice();
-        let mut combined = Poly::zero(table.num_variables());
-        combined
+        // The shift reads the entry before, which a column taken from its last row down still
+        // finds unmoved. So the combination lands in the table it reads.
+        let stride = CHUNK.max(column).min(table.num_evals());
```
tcoratger: **[MINOR] The `stride % column == 0` invariant is derived three functions away and is not asserted — `sumcheck/src/ring_switch/bits/reduction.rs:894`**

The in-place alpha pass indexes `entries[row]` for `row in 0..column` on a slice produced by `part.chunks_mut(column)`. That never yields a short tail only because `column <= table.num_evals()`, which holds only because `prefix_limit()` caps the Boolean prefix at `num_variables - kept` whenever a successor view is present. Nothing at the indexing site records this, and the failure mode is a panic inside a rayon closure.

**Fix:**
```rust
debug_assert!(column.is_power_of_two() && stride.is_multiple_of(column));
debug_assert!(table.num_evals().is_multiple_of(stride));
```
plus a one-line comment pointing at `prefix_limit` as the reason `column <= num_evals`.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -391,10 +396,11 @@ impl<EF: TowerLevel> BitRingSwitch<EF> {
 
         // Only the selected slot feeds the sumcheck.
         // This avoids cloning and folding unrelated columns of a stacked trace.
-        let slot: Vec<EF> = packing.poly().as_slice()[offset..offset + len]
-            .par_iter()
-            .copied()
-            .collect();
+        let values = &packing.poly().as_slice()[offset..offset + len];
+        let mut slot = EF::zero_vec(len);
```
tcoratger, line 402: **[MINOR] `restricted_packing` now depends on `EF::zero_vec` being calloc-backed — `sumcheck/src/ring_switch/bits/reduction.rs:400`**

`EF::zero_vec(len)` + parallel copy replaces `par_iter().copied().collect()`. This is faster only because every `TowerLevel` impl overrides `zero_vec` with the `ManuallyDrop`/`from_raw_parts` trick over `vec![0u128; len]` (`binary-field/src/ghash.rs:216`, `binary-field/src/tower.rs:413`), which hits the `IsZero` specialization and so `alloc_zeroed`. The `PrimeCharacteristicRing` default (`field/src/field.rs:401`) is a serial `vec![Self::ZERO; len]`, which at `-l 18` would be a serial 512 MB write before the parallel copy. Worth a comment, since the generic bound is `EF: TowerLevel` and nothing enforces the override.

tcoratger (commented):

---

Reviewed against the stack base (`perf/ring-switch-poly-basis`). I derived the factoring identity by hand, checked every consumer against it, and stress-tested the block split by sweeping the test-mode block size across seven values — all pass.

**No blockers; the math is right and the change is prover-side in effect.** Two things need attention before merge: the PR body's "prover-only" claim is wrong in a way that matters for review attention, and the block size the code actually ships (`2^14`) is never exercised by any test (the test build uses `2`).

---

## The factoring identity

The equality polynomial is a product over coordinates, and `Poly::new_from_point` lays the table out with the **first** coordinate most significant. So splitting the point after its leading coordinates splits the table into equal, contiguous blocks:

```
    point   =  ( r_0 ... r_{k-m-1} | r_{k-m} ... r_{k-1} )
                \---- outer ----/   \----- inner -----/
                  block index i       within-block j

    table[i * 2^m + j]  ==  eq(outer, i) * eq(inner, j)
                        ==  outer[i] * inner[j]
```

`equality.rs:47` splits at `point.len() - min(log_block, point.len())`, i.e. the **inner factor takes the trailing coordinates**. That is the correct end: the trailing coordinates are the ones that vary within a contiguous run of `2^m` indices. Getting this backwards is the classic bug and the code does not have it — `the_factors_multiply_to_the_dense_table` pins it over 0..10 variables × every block size.

Note this split is **not** the ring-switch kappa. Kappa here is `ABSORBED = Coefficients::<EF>::LOG_DIMENSION` (7 for `BinaryField128`), which was already consumed before `high()` is formed. `LOG_CHUNK` is purely a cache/parallelism tuning knob over the *remaining* coordinates, so there is no off-by-one-against-kappa risk to check. Good separation.

The three consumers rely on three sub-identities, all of which I verified:

**1. Tensor leg carrying (`reduction.rs:483`).** In `EF ⊗_{F_2} EF`, `(a ⊗ 1) * (b ⊗ c) = (ab) ⊗ c`, and `⊗` is `F_2`-bilinear, so

```
    sum_j (outer_i * inner_j) ⊗ t'(w)  ==  (outer_i ⊗ 1) * sum_j inner_j ⊗ t'(w)
```

`BitTensor` stores `element = sum_u beta_u ⊗ row_u`, so `(a ⊗ 1) * element = sum_u (a * beta_u) ⊗ row_u` — exactly what `add_scaled_columns` (`tensor.rs:164`) computes. Correct, and it is `d` multiplies instead of a pair of `d × d` bit transposes.

**2. Coordinate-sum rescaling (`equality.rs:137`).** `sums.sum(.)` is `F_2`-linear and so is multiplication by a fixed scale, so

```
    scaled(sums, c).sum(x) = sum_{u in x} sums.sum(c * beta_u)
                           = sums.sum(c * sum_{u in x} beta_u)
                           = sums.sum(c * x)
```

which makes `scaled.sum(inner_j) == sums.sum(outer_i * inner_j)` — bit-identical to the old `sums.sum(dense[w])`. Correct.

**3. Block-boundary carry (`reduction.rs:578`).** The successor sweep reads `table[w-1]`. Inside a block that is `outer_i * inner_{j-1}`, which factors; at `j == 0` it is `outer_{i-1} * inner_{2^m - 1}`, which does **not** share the block's scale, so it is added unscaled via `before_block` (`equality.rs:91`) directly into `carry` rather than into the buckets. This is the one place the factoring genuinely leaks, and it is handled correctly — including when `block < column` (a block starting mid-column, `row != 0` at `j == 0`), which happens whenever `kept > LOG_CHUNK`.

**Normalization / boundaries.** `kappa = 0` is not reachable here (`kept_row_variables` filters `rows > ABSORBED`). Zero variables gives a single unit entry (tested, `equality.rs:244`). A run shorter than one block collapses `outer` to `[ONE]` (tested, `equality.rs:187`). `before_block(0) = ZERO` is a sentinel that is never consumed, because `w = 0` always has `row == 0`.

**Verdict on the central question: this is a pure regrouping, not a value change.** Multiplication is associative and the code groups the same per-coordinate factors differently. The verifier needs no matching change because the verify path (`reduction.rs:1344-1368`) only calls `initial_sum` and `closing_weight`, neither of which touches `support()`. I confirmed that by tracing every caller of `support()` and `weights_over`.

**Structural safety of the in-place alpha combination (`reduction.rs:894-906`) — worth spelling out, because it is the subtlest line in the diff.** The combination `w <- A(w) + alpha*A(w-1) + [row==max] alpha^2*A(w)` now writes into the table it reads. It is safe because the loop descends (`for row in (0..column).rev()`): when `entries[row]` is written, `entries[row-1]` has not been touched yet, and `entries[row]`'s own old value is saved in `settled` first. The chunking is also sound, but only via a three-function argument: `stride = CHUNK.max(column).min(num_evals)` is a power of two and `column <= num_evals` holds because `prefix_limit()` (`reduction.rs:315`) caps the Boolean prefix at `num_variables - kept` exactly when a successor view is present, and `alpha.is_some()` implies `sends_successor_tensors()` via `batch_with_successor`. So `column | stride | num_evals`, and `part.chunks_mut(column)` never yields a short tail. If that chain ever breaks the result is an out-of-bounds panic, not a wrong proof — but see the MINOR on debug asserts.

---

### [NIT]

- `adding_a_scaled_element_scales_its_column_reading` (`tensor.rs:511`) is circular: it compares `add_scaled_columns` against `scale_columns`, but `scale_columns` is now *implemented as* `add_scaled_columns` into a zero element. The test proves `zero + x == x` and additivity, not the scaling semantics. Assert against `columns()` directly, the way `scaling_the_columns_scales_every_column_reading` does.
- Stale comment: `tensor.rs:498` still says "The stored rows are the other reading, so this is the transpose path" — there is no transpose path any more.
- `add_scaled_columns` is `pub` but used only in-crate (`tensor.rs:164`). `BitTensor` is a wire type with a deliberately small public surface; `pub(crate)` fits better. Also `BitTensorBuckets::tensor` (`tensor.rs:292`) allocates and returns and should carry `#[must_use]`.
- `drop(equality)` is now vestigial (`reduction.rs:1235`): it existed to free the 512 MB dense table before the sumcheck; `FactoredEquality` is a few hundred KB at most. Either delete it or comment why it stays.
- The doc on `add_scaled_columns` (`tensor.rs:162`) says "one multiplication per coordinate", which undersells the cost: it is `d` multiplications **plus** roughly `d^2 / 2` field additions from the `add_exterior_product` fan-out. Still a win over two `d × d` transposes plus four allocations, but the comment reads as if the additions are free.
- The head branch is `perf/opening-share`, which describes nothing in this diff — I checked all four changed files and there is no opening-sharing change. Harmless, but it makes the stack hard to follow.
- There are now two modules named `equality` under `ring_switch` (`ring_switch/equality.rs` and `ring_switch/bits/equality.rs`). Consider `factored_equality.rs` for the new one.

---

## Test coverage

What already exists and is good: `the_factors_multiply_to_the_dense_table` (0..10 variables × every block size) is the right shape for the core identity; `the_successor_elements_are_their_term_by_term_sums` rebuilds both elements from the dense table and loops over `ROW_VARIABLES × Selector::ALL`; `scaled_sums_read_the_scaled_value` pins the linearity argument including `scale = ZERO` and `scale = ONE`.

Gaps, in priority order:

1. **Block size is a compile-time constant, so the shipped value is untested.** Covered under the MAJOR above — add the block-size-invariance test (written and verified passing) so the `stride > column` path stops being dead code in the test build.

2. **`the_successor_weights_are_the_dense_table_entry_by_entry` tests exactly one configuration** — `reduction.rs:2321`, `row_variables = 6`, `Selector::Random`, `num_variables = 9`. Its sibling element test two functions above loops over `ROW_VARIABLES × Selector::ALL`; this one should too. The missing configurations are the interesting ones: a Boolean or Mixed selector gives a nonzero prefix and shrinks the run (possibly to `num_evals == column`, a single column chunk), and `row_variables = 9` gives `kept = 5` with `column == num_evals`.

3. **`scale_columns` at `BinaryField128`** — covered under the MAJOR above.

4. **A proptest is a good fit for the factoring itself**, better than more fixed seeds:

```rust
proptest! {
    #[test]
    fn the_factored_table_is_the_dense_table(
        num_variables in 0usize..10,
        log_block in 0usize..12,
        seed: u64,
    ) {
        let point = Point::<BinaryField128>::rand(&mut SmallRng::seed_from_u64(seed), num_variables);
        let dense = Poly::new_from_point(point.as_slice(), F::ONE);
        prop_assert_eq!(
            FactoredEquality::new(point.as_slice(), log_block).materialize().as_slice(),
            dense.as_slice()
        );
    }
}
```

A failure here names the exact `(num_variables, log_block)` pair, which pins whether the split consumed the leading or trailing coordinates. `tensor.rs` already depends on `proptest`, so no new dependency. Paper known-answer vectors are *not* the right tool here — this is an internal regrouping of an implementation detail, not a spec-level quantity, so a differential against the dense computation is strictly stronger.

5. **Not a gap, but worth recording in the PR body:** the multiplication accounting checks out against the posted counter. At BLAKE3 `-l 14` (`k = 21`, `m = 14`): factors `2^7 - 1 + 2^14 - 1 = 16510`, plus `2^7 * 128 = 16384` each in `tensor_over` and `weights_over`, totalling ~49.3K against the measured 44.5K — the gap is `add_scaled_columns` skipping zero rows. Before is `2^21 - 1 = 2,097,151`, matching the table exactly. The claimed 47x is real.

---

## Could not verify

- **Benchmarks.** I did not reproduce the c7i.4xlarge numbers. The codegen-units caveat in the body is a genuinely useful finding and I have no reason to doubt the medians, but the `-l 10` / `-l 11` regime flagged in the parallelism MINOR is not in the posted table and I could not measure it here.
- **What `R` is after #2253.** I confirmed the basis question is safe *structurally*: the scale multiplication `scale * beta_u` happens entirely in `EF` on the index side, and `R` only ever receives additions and the `alpha` multiplications that were already there. So `scaled_sums` is basis-agnostic by construction and cannot carry a stale-basis weight. I did not independently re-verify that `R: From<EF>` is a field isomorphism — that is #2253's contract, asserted in the `prove` doc comment, and outside this diff.

---

**Bottom line:** merge after fixes — the two MAJORs are a doc correction plus two tests, neither of which changes production code, and I have already verified that both underlying behaviours are correct. The factoring itself is sound and I would merge it as-is if the test debt were being tracked.

sumcheck/src/ring_switch/bits/tensor.rs
Nashtare: Confirmed both halves. `scale_columns` is on the verifier path — `equality_element` calls it at `reduction.rs:742,745`, `closing_weight` calls that at `:959`, and `verify` calls `closing_weight` at `:1368` — and the only semantic test ran at `BinaryField16` alone.

Fixed in `609ce8dd8`: the body moved into `the_columns_scale_at::<F>` and the test instantiates it at `BinaryField16` and `BinaryField128`. It builds its element at the level under test rather than through the `BinaryField16`-typed `element` helper, so no other test in the file changes.

The scope line in the body should read: prover-only except `BitTensor::scale_columns`/`add_scaled_columns`, which the verifier reaches through `equality_element` -> `closing_weight`; proof bytes unchanged. I am updating the description.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Confirmed: `par_fold_reduce` hands the same identity to `fold` and to `reduce` (`maybe-rayon/src/lib.rs:59`), so every reduce seed built a `BitTensorBuckets` only to drop it on the next line.

Done in `0f2076c68` for both folds. The accumulator carries `Option<BitTensorBuckets>`, the fold arm takes it with `get_or_insert_with(BitTensorBuckets::zero)`, and the reduce arm passes the left one through.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Added in `8d8e31692`, with the reason at the site rather than only at `prefix_limit`.

The derivation holds. With a successor view `prefix_limit()` is `min(num_variables, point.num_variables() - rows)`, and `point.num_variables() - rows == num_variables - kept`, so `prefix <= num_variables - kept` and `num_evals = 2^(num_variables - prefix) >= 2^kept == column`. Everything in sight is a power of two, so `stride` is a multiple of `column` and `num_evals` a multiple of `stride`.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Comment added in `d7244baaa`, phrased as the requirement rather than as a fact about the trait — the premise does not hold for every impl. `Gf2` (`binary-field/src/gf2.rs:101`) and `Rijndael8b` (`binary-field/src/aes/mod.rs:399`) implement `TowerLevel` and override nothing, so they take the `PrimeCharacteristicRing` default; the seven `binary_tower_level!` levels, `Ghash128` and `Poly64` do override it. That sharpens the point: the bound admits a level that pays the serial fill today, not only in principle.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Both consequences check out, and the proposed test as written would have caught only the second one.

`stride` in the alpha pass came from the `CHUNK` const, not from the `FactoredEquality` the test passes in, so sweeping `log_block` leaves `stride == column` and the multi-column path stays dead however the table is split. In `e07bb3ac9` the stride comes from `equality.block_len()` instead. That is the same value wherever the reduction builds the table: `support()` builds it at `LOG_CHUNK`, so `block_len == CHUNK.min(num_evals)`; for `num_evals >= CHUNK` both expressions read `CHUNK.max(column)`, and below it both clamp to `num_evals`.

The test then sweeps the split over `0..=run.len() + 2` and compares all three sweeps — `tensor_over`, `successor_tensors_over` and `weights_over` — against the one-entry-per-block reference, over `ROW_VARIABLES` x `Selector::ALL`.

Two mutations to show it bites:

- `inner[j - 1]` -> `inner[j]` in the carry interior arm, invisible at `block_len == 2` where `j` is always 0;
- `part.chunks_mut(column).take(1)` in the combination, a no-op while `stride == column`.

Each passes all 88 other `ring_switch` tests and fails only the new one.

sumcheck/src/ring_switch/bits/equality.rs
Nashtare: Done in `fccddc6a8`, with one departure from the sketch.

`CoordinateSums::overwrite` retabulates through the entries a caller already holds, reading one weight per coordinate, `scaled_sums_into` drives it, and `weights_over` carries one table per task through `for_each_init`.

I dropped `scaled_sums` rather than keeping it as the allocating wrapper. With `weights_over` on the in-place form it had no caller outside `cfg(test)`, so it would have been a test-only symbol standing in front of the function that actually runs. `scaled_sums_read_the_scaled_value` now drives `scaled_sums_into` through one table reused across four scales, which additionally pins that a rescale leaves nothing of the one before — the allocating form could not test that.

sumcheck/src/ring_switch/bits/tensor.rs
tcoratger: Half fixed. The coverage gap is closed: `the_columns_scale_at::<F>` now runs at `BinaryField16` and `BinaryField128`, and I re-ran the 137 `ring_switch` tests at both debug and release — all pass.

The other half is not done. The PR description still opens its Correctness section with **"Prover-only; proofs byte-identical to #2246"**, and still says "Under `cfg(test)` the block is two entries, so the block-boundary paths are on the tested path" — which is the sentence the block-size thread invalidated. The scope line you wrote in your reply is the right one; it just did not land in the body. Not blocking, since I already verified there is no divergence, but please edit the description before merge.

sumcheck/src/ring_switch/bits/reduction.rs
tcoratger: Still open — no reply and no change. `weights_over`'s first pass is still `par_chunks_mut(equality.block_len())` with `block_len == CHUNK.min(num_evals)`, so a run with `k <= 14` is one rayon task where it used to be `2^k`. The `for_each_init` rework in `a71b3c4a` kept the same chunking.

Fine to leave as is — it is a MINOR and the regime the PR targets is `k` well above 14 — but a one-line comment at the `par_chunks_mut` recording that the run is assumed large would stop the next reader reading it as a parallelism unit.

tcoratger (approved):
Re-reviewed at `e532a639` against my findings at `29cdbe0a`. **Five of seven threads fixed and resolved; one MAJOR fixed in code but not in the description, one MINOR untouched. Approving** — nothing outstanding is a code defect.

### MAJOR — `scale_columns` on the verifier path: **code fixed, body not**

`the_columns_scale_at::<F>` now runs at `BinaryField16` and `BinaryField128`, building its element at the level under test, so the production field width is covered. That was the substantive half, and I had already shown there is no divergence (203/203 bit-identical).

The description was not updated. Its Correctness section still opens **"Prover-only; proofs byte-identical to #2246"**, and still carries "Under `cfg(test)` the block is two entries, so the block-boundary paths are on the tested path" — the exact claim the next thread invalidated. Please land the scope line from your reply before merge. Left that thread unresolved.

### MAJOR — the shipped block size: **fixed, and I confirmed the test bites**

You were right that my sketch would only have caught half of it. `stride` now reads `equality.block_len()` rather than the `CHUNK` const, so sweeping `log_block` actually moves the stride. I checked the production equivalence by hand: `support()` builds the table at `LOG_CHUNK`, so `block_len == CHUNK.min(num_evals)`, and `block_len.max(column).min(num_evals)` equals the old `CHUNK.max(column).min(num_evals)` in both regimes (`num_evals >= CHUNK` and below it, using `column <= num_evals`). Both production call sites go through `support()`. No behaviour change.

Ran your two mutations:

| mutation | result |
| --- | --- |
| `part.chunks_mut(column).take(1)` | **136 pass, only `the_sweeps_read_the_same_table_at_every_block_size` fails** |
| `inner[j - 1]` -> `inner[j]` in the carry interior | 9 fail, the new test among them |

The first is the one that matters: a no-op while `stride == column`, invisible to all 136 other tests, caught solely by the new one. The multi-column path production takes is now genuinely exercised. The `debug_assert!`s hold in a debug run of the full suite.

### MINORs

- Bucket scratch is `Option<BitTensorBuckets>` in both folds, `get_or_insert_with` in the fold arm, left-hand passed through in the reduce arm. Fixed.
- `CoordinateSums::overwrite` + `scaled_sums_into` + `for_each_init` removes the per-block allocation. I checked `overwrite` against the old `new`: same subset recurrence, `table[0]` explicitly zeroed so no stale entry survives a rescale, closure called exactly `DIMENSION` times, and `out`/`sums` never alias at the call site. Dropping the allocating `scaled_sums` is the right call given it had no non-test caller.
- `stride % column` invariant asserted with the `prefix_limit` reasoning at the site. Your derivation is correct.
- `zero_vec` comment landed, and phrasing it as the requirement rather than a fact about `TowerLevel` is the better call — `Gf2` and `Rijndael8b` do take the serial default.
- **Parallelism cliff in `weights_over`'s first pass: unaddressed and unanswered.** Still `par_chunks_mut(block_len)`, so `k <= 14` is one task. Left unresolved; MINOR, not blocking.

`cargo test -p p3-sumcheck --lib ring_switch` passes 137/137 in debug and release; clippy `--all-targets` clean.

sumcheck/src/ring_switch/bits/tensor.rs
Nashtare: Confirmed the remaining half, and traced it before rewording anything. `BitTensor::scale_columns` (`sumcheck/src/ring_switch/bits/tensor.rs:145`) is reached from `equality_element` (`reduction.rs:767,770`) -> `closing_weight` (`reduction.rs:1002`) -> `verify_readings` (`reduction.rs:1415`), and `verify` delegates to `verify_readings`, so both verifier entry points run it. Your line numbers drifted with the rebase; the chain did not. "Prover-only" is wrong as written.

Rewording the description to, first sentence of Correctness:

> Prover-only except for `BitTensor::scale_columns`, which the verifier reaches through `equality_element` -> `closing_weight` -> `verify_readings`; everything else the diff touches runs on the prover alone. Proofs are byte-identical to #2246 at `-l 14` and `-l 16` for both objectives, and proof sizes match at `-l 18`.

and the last sentence of that paragraph, which the block-size thread invalidated:

> Under `cfg(test)` `LOG_CHUNK` is 1, so a two-entry block is all the tests would otherwise see -- a block where `weights_over`'s stride collapses onto a single column and `successor_tensors_over` is all boundary and no interior. `the_sweeps_read_the_same_table_at_every_block_size` covers that instead: it pins `tensor_over`, `successor_tensors_over` and `weights_over` against the one-entry-per-block table at every block size from `2^0` to past the run, so the shipped `2^14` block is the same program the tests check. `the_columns_scale_at` runs the rewritten `scale_columns` at `BinaryField128` as well as `BinaryField16`.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Measured it. I put a span around the first pass alone and ran Keccak-f on a c7i.4xlarge (16 vCPU), `-C target-cpu=native`, `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, five reps per point, medians with the full range:

| `-l` | table entries | blocks | first pass | of prove |
| ---: | ---: | ---: | ---: | ---: |
| 10 | 2^14 | 1 | **162 µs** (155-181) | 0.6-0.8% |
| 11 | 2^15 | 2 | 416 µs (182-921) | 0.6-2.6% |
| 14 | 2^18 | 16 | 1.11 ms (679 µs-1.33 ms) | 0.8-1.6% |
| 18 | 2^22 | 256 | 8.05 ms | 1.0% |

`-l 10` is exactly the boundary you predicted: the run is `2^14` entries, one block, one task. Your `k = log(rows) - 7` comes out as `k = l + 4` here, so `k <= 14` is `-l <= 10` for Keccak-f, and the single-task case is the whole of it.

Then both decouplings, same instrument:

- **Chunking the rayon split** at `block_len.min(num_evals >> log_threads)`, floored at `2^8`: at `-l 10` the pass goes **162 -> 865 µs** (800-1270), 5.3x worse. Each sub-block task has to rescale the coordinate sums for itself, and one rescale is 128 tower multiplies plus `NUM_BYTES * 255` additions -- worth a couple of hundred entries read, against the 1024 a task then writes.
- **Rescaling once per block and spreading only the entry loop** (`par_iter_mut().zip(inner.par_iter())` under a `blocks < current_num_threads()` arm, the shape `aggregate_columns` already uses at `layout/prover/residual.rs:285`): at `-l 10` **162 -> 212 µs** (160-457). Still not a win. A `sums.sum()` is one chain through a 64 KB table, about 11 ns, so 16 threads get 11 µs of work each and pay more than that pulling the same table and fanning out.

End-to-end neither shows up, and could not: `bit ring switch` has a ±10% spread on this box, calibrated at `-l 14` where both variants are no-ops and a 12-rep interleaved A/B still read +9.7%.

So the cliff is real, and the parallelism is not recoverable at that size -- the pass is too cheap per entry to spread and the scale table too big to duplicate. Taking your fallback: no code change, and a comment at the `par_chunks_mut` recording that the block is the amortisation unit the split follows, plus why splitting below it does not get the tasks back (`reduction.rs:869-876`, commit `docs(sumcheck): record that the weight sweep splits on its amortisation unit`).

The same note covers `tensor_over` and `restricted_packing`, which split on the same unit for the same reason.

PR #2252, opened 2026-09-20
perf(binary-dft): deepen the polynomial-basis tiles past the shared cache

Nashtare:
Stacked on #2249.

The polynomial-basis transform already tiles its traversal; its tile budgets were sized for a
matrix the cache holds. Once the matrix is larger than the last-level cache the budget should be
the private L2 share of a hardware thread instead, because at that point every pass is a trip to
memory and the thing worth minimising is the number of passes rather than the residency of a tile.

`Plan::budgets(row, log_n)` hands both the contiguous tile and the staging tile a deeper budget
(`DEEP_TILE_BYTES`, 256 KiB) only when the matrix is past `SHARED_CACHE_BYTES`. The figure is half
of a hardware thread's share of private cache on a core holding 1 MiB for two SMT siblings; the
other half is the gather and scatter stream, which transits the same cache for as long as the tile
is live. The two tiles never coexist — `local_stages` and `fused_stages` are separate phases — so
the halving is not a second tile. `TILE_BYTES` and `STAGING_BYTES` are untouched and still serve every
matrix the cache holds.

At the shape the binary commitment actually encodes — width 1, 2^25 rows — the cut moves from
`(local 11, run 32 rows, depth 7)`, three passes over the matrix, to `(local 14, run 8 rows,
depth 11)`, two passes. The run is not preserved: a tile of a fixed size trades run length against
depth one for one, and the search spends the run because a traversal saved outranks a doubling of
the run. It moves in both directions across the shape space — width 16 @ 2^20 drops from two rows
to the cache-line floor of one, while width 48 @ 2^18 grows from one row to two.

### Measurements

The `c7i.4xlarge` table below was measured with `DEEP_TILE_BYTES` at 512 KiB and needs a re-run at
256 KiB.

#### Budget sweep on a small-L2 target

c6a.4xlarge (EPYC 7R13, Milan — 512 KiB of L2 per core, two threads per core, 32 MiB of L3, THP
`madvise`), `encode/BinaryField128`, criterion point estimates, base = `67f7ce78`:

| budget | 2^18, below the gate | 2^20, past it |
| --- | --- | --- |
| shallow pair (base) | 67.21 / 66.37 ms | 267.84 / 268.23 ms |
| 256 KiB | 66.12 ms | 264.80 ms |
| 512 KiB | 66.22 / 66.17 ms | 270.85 / 270.36 ms |
| 1 MiB | 66.51 ms | 274.49 ms |

At 512 KiB the deep budget is a 1.0% regression on that core rather than a win; at 256 KiB it is a
1.2% win. The curve is monotone in the budget, repeated arms land within 0.2% of each other, and
the below-gate control — where every arm plans the same cut — is flat, so the span is the budget.
That is what sets the constant.

c7i.4xlarge (Xeon 8488C, 16 vCPU), `-C target-cpu=native`, `--features parallel`, default PCS
flags, medians of 3, base and branch alternated.

| | `encode` | `binary pcs commit` |
| --- | --- | --- |
| BLAKE3 `-l 18` | 206 -> 187 ms (-9.2%) | 303 -> 283 ms (-6.6%) |
| BLAKE3 `-l 16` | 43.2 -> 43.2 ms | — |
| Keccak-f `-l 18` | 19.0 -> 19.5 ms | — |

The two unchanged rows are below the gate and take today's plan, as intended. Total prove moves
about 20 ms, which is 0.7% — that is the honest figure; the prove spread on this host is wider than
the difference, so the commit span is the number to read.

`binary-dft`'s criterion `encode/BinaryField128` at width 16, rate 1/2: 2^14 1.566 -> 1.525 ms,
2^16 16.43 -> 16.00 ms, 2^18 47.17 -> 46.82 ms (all same plan, criterion noise is +-2-3% here), and
2^20 202.97 -> 198.57 ms (-2.2%) — the only width-16 height past the gate.

Peak RSS unchanged at 10.24 GB.

### Why it is gated rather than a flat bump

Every flat budget increase regresses something:

| variant | result |
| --- | --- |
| staging 512 KiB alone | width 16 @ 2^18 +12.8% (the traversal budget collapses the gathered run to one row) |
| contiguous 512 KiB alone | BLAKE3 `-l 18` +14% |
| both 512 KiB | BLAKE3 `-l 18` -11%, but width 16 @ 2^14 +48%, BLAKE3 `-l 16` +3.6%, Keccak-f `-l 18` +4.1% |
| both, plus a four-tiles-per-worker cap | fixes the 2^14 cliff (-2.9%), leaves `-l 16` at +3.6% and Keccak at +4.1% |

Those variants were swept at a 512 KiB budget; the gate is what removes their regressions, and the
budget was cut to 256 KiB afterwards for the reason above.

A 4 MiB matrix in 512 KiB tiles is eight tiles for sixteen workers, which is the 2^14 cliff. Gating
on the matrix outgrowing the shared cache removes all four regressions, keeps both wins, and makes
the worker-share cap unnecessary — a matrix past 128 MiB always has at least 256 tiles.

### Correctness

`a_deep_budget_cut_matches_the_per_stage_schedule` runs the depths, run lengths and tile depths the
deep budget produces — `(48, 1, 7) (16, 0, 10) (1, 3, 11)` and `local = 14, 10, 8` — against the
per-stage oracle in both directions, on matrices short enough for CI. Before it, every shape the
suite transformed sat below the gate and the deep rows of `PLAN_ONLY_CUTS` were asserted as triples
only. `Plan::for_workers(1, 25, 32)` and `(16, 21, 32)` were also run against the oracle offline at
full magnitude (3.0 s in release, roughly ten minutes in debug, hence not in CI): both pass.

The cut is a pure reordering. `every_cut_of_the_stage_sequence_matches_the_per_stage_schedule` and
`every_width_and_height_matches_the_per_stage_schedule` pin the transform bit-for-bit against a
one-pass-per-stage oracle over every width, height, shift and direction. `PRODUCTION_CUTS` is
unchanged, since every shape there is below the gate; `PLAN_ONLY_CUTS` gains a width-1 @ 2^25 row
and updates the two shapes past it. Every A/B run printed identical proof size and composed
security, and both sides verify.

`cargo test --workspace --features parallel` passes; clippy and rustfmt clean.

### Caveat

`SHARED_CACHE_BYTES` and `DEEP_TILE_BYTES` are independent hardware parameters: the first stands in
for an LLC, the second has to fit the private cache under it, and a target can be large in one and
small in the other. The matrix-size sweep that places the gate is flat inside its noise through the
middle, so 128 MiB is a bracket between its two ends (4 MiB loses half again, 512 MiB wins a tenth)
and not a reading of a cache level. The tile figure is inside a hardware thread's private share on
every target measured, which is what the EPYC sweep fixes it at. A machine with a very different
hierarchy should still re-measure both.

### Negative result: fill fusion

`encode_by_padding` resizes the message and zero-fills the tail, and the transform then copies the
message into each coset, so the tail looks like it is written twice. Deleting that copy — having a
coset's first staging gather read the message and scatter into the coset — measures **+2.4% on
encode**, 3 of 3 reps. Two reasons:

- The premise does not hold on the hot path. `commit_base` calls `encode_batch_padded`, not
  `encode_batch`, so the `resize` never runs; the padded buffer comes from `F::zero_vec`, which is
  `alloc_zeroed`, so the zero fill costs no userspace writes at all.
- An in-place gather and scatter writes back the lines the gather just pulled in, so the scatter
  needs no read-for-ownership and the destination's first-touch faults land in a sequential memcpy.
  Sourcing from a separate buffer adds a full read-for-ownership stream and moves those faults into
  the strided scatter.

For the record, `encode` at BLAKE3 `-l 18` is 213 ms of which the coset copy is 21.7 ms: convert
11.2, coset copy 21.7, tail forward 89.3, message forward 89.2. The copy is a tenth of encode and
is not free to delete.

Separately, the identity `Vec` conversions over the whole codeword, twice per encode, are already
free — LLVM elides them entirely (0 ns on a 64 MiB vector against a `ManuallyDrop` transmute).


tcoratger (commented):
Reviewed against the stack base (`perf/zc-lazy-unslice`). Planner-only change: the executor is untouched. I ran the real production deep plan (width 1 @ `2^25`, `(local 15, log_block 5, depth 10)`) against the per-stage oracle in both directions — **bit-exact**. No soundness or safety issue. What is weak is the perf model and the verification, not the correctness.

## Verified positively

- **The gate is exact.** `log_n > log2_floor_usize((SHARED_CACHE_BYTES / row).max(1))` is equivalent to `row * 2^log_n > SHARED_CACHE_BYTES`, both directions, under integer floor division. Comparing in rows rather than computing `row << log_n` correctly dodges a shift overflow at large `log_n`.
- **The deep budget never costs a traversal.** I reimplemented `budgets`/`groups`/`traversals`/the `log_block` search independently, reproduced all 8 `PLAN_ONLY_CUTS` triples exactly, then swept widths 1..1199 × matrices up to 8 GiB: **zero** shapes where the deep plan takes more passes than the shallow one.
- **The parallelism claim holds.** Over every deep shape in that sweep, `min(contiguous tiles, staging tiles) = 512`; the guaranteed bound is `SHARED_CACHE_BYTES / DEEP_TILE_BYTES = 256`. The 4 MiB / 8-tiles-for-16-workers cliff cannot recur behind this gate.
- **The `unsafe` invariants survive.** `Rows::gather`/`scatter` rest on `stride << depth == 1 << (top - log_block)` and `tile_len | len`, both only `debug_assert`ed. Both are algebraic identities given `log_block + depth <= top`, which the `min` at `poly.rs:757` enforces unconditionally. Deeper plans change neither. `Rows::check` remains a hard assert, so even a mis-derived plan bounds-faults rather than writing OOB.
- On the deep path `local < log_n` strictly (in fact `above >= 8`), so `tile_exit`, `leftover` and the group walk keep their shape. No new inner loop, no new `unsafe`, no new allocation site.

---

## [MAJOR] The deep budget is sized by L2 but gated on L3 — two independent axes — `binary-dft/src/poly.rs:250` and `:261`

`SHARED_CACHE_BYTES` decides *whether* the deep budget applies; `DEEP_TILE_BYTES` is sized as a fraction of **private L2**. These are unrelated hardware parameters, and the doc's safety argument only covers one of them:

> A target whose shared cache is smaller keeps the shallower tiles … which is a win forgone rather than a cost added — `poly.rs:258`

That is true for *small-LLC* targets. It says nothing about a **large-LLC, small-L2** target, which is where the gate fires and the tile does not fit:

```text
                      L2/core   threads/core   L2/thread   aggregate L3   gate fires?
Xeon 8488C (measured)  2 MiB          2          1 MiB        ~105 MiB      at >128 MiB
EPYC 7763 (Milan)    512 KiB          2        256 KiB         256 MiB      at >128 MiB
EPYC 7742 (Rome)     512 KiB          2        256 KiB         256 MiB      at >128 MiB
```

On Milan a 256 MiB matrix passes the gate and each worker then streams a 512 KiB tile — **100% of the core's entire L2, 200% under SMT.** The sibling constant's own doc states the requirement this breaks: *"What it must not do is spill out of the private cache level below the shared one"* (`poly.rs:213`). The tile then re-reads from L3 on each of its `depth` sub-stages, so the fusion buys much less than the traversal it spent.

**Fix:** decouple the two. Either drop `DEEP_TILE_BYTES` to 256 KiB (fits a 512 KiB L2 with SMT, still 8x today's staging budget and still >=256 tiles), or keep 512 KiB and narrow the caveat to say explicitly that it assumes >=1 MiB of L2 per hardware thread and that a smaller-L2 target is a *cost added*, not a win forgone.

**Verify:** one `binary-dft` `encode/BinaryField128` run at `2^20` × width 16 (the one bench point past the gate) on any Zen 2/3 EPYC, base vs branch.

---

## [MAJOR] No test ever executes a plan the deep branch produces — `binary-dft/src/poly.rs:1188`

Every shape that is actually transformed in the suite sits below the gate:

- `PRODUCTION_CUTS` — largest is width 1 @ `2^20` = 16 MiB. All shallow.
- `every_width_and_height_matches_the_per_stage_schedule` — sweeps to `log_n = 14`. All shallow.
- `cut_plans()` — synthetic, hand-written triples; max `depth` is 3, max `local` is 2.
- `PLAN_ONLY_CUTS` — the three deep rows are asserted as *triples only*. Nothing runs them.

So the deep branch is covered by an equality against a table that this same PR wrote. `local = 15`, `depth = 10`, `log_block = 5` are magnitudes no executed test reaches. The PR's correctness section cites the two oracle tests, but neither sees a deep plan.

I ran the real thing to close the gap myself — `Plan::for_workers(1, 25, 32)` against `per_stage_schedule`, both directions: **passes**, in 595 s. Too slow for CI, which is presumably why these were kept plan-only. But the *structural* regimes are cheap: the test below passed in **1.15 s** and covers both new regimes at their real magnitudes — the deep group depths and run lengths `(0,9) (2,9) (5,10)`, and the deep contiguous-tile depths `local = 9, 11, 15`:

```rust
#[test]
fn a_deep_budget_cut_matches_the_per_stage_schedule() {
    // (a) the deep group depths and run lengths, over a shallow contiguous tile.
    for (width, log_block, depth) in [(48usize, 0usize, 9usize), (16, 2, 9), (1, 5, 10)] {
        for extra in [0usize, 1, 2] {
            let (local, log_n) = (1usize, 1 + depth + extra);
            let plan = Plan { width, log_n, local, log_block, depth };
            for inverse in [false, true] {
                let mut expected = coefficients(log_n, width);
                let mut actual = expected.clone();
                per_stage_schedule(&mut expected, width, log_n, test_shift(), inverse);
                scheduled(&mut actual, plan, inverse, NONE);
                assert_eq!(actual, expected, "{plan:?} inverse={inverse}");
            }
        }
    }
    // (b) the deep contiguous tile depths, with a shallow group above them.
    for (width, local) in [(1usize, 15usize), (16, 11), (48, 9)] {
        for extra in [1usize, 2, 3] {
            let (log_block, depth) = (0usize, 2usize);
            let log_n = local + extra;
            let plan = Plan { width, log_n, local, log_block, depth };
            for inverse in [false, true] { /* as above */ }
        }
    }
}
```

Optionally also `const _: () = assert!(SHARED_CACHE_BYTES / DEEP_TILE_BYTES >= 256);` so the tile-count invariant the PR's rationale depends on is enforced rather than argued in prose.

---

## [MINOR] `DEEP_TILE_BYTES`'s stated derivation yields 1 MiB, not 512 KiB — `binary-dft/src/poly.rs:232`

The doc says the budget is "one hardware thread's share, which is half of its core's … on a core holding 2 MiB of private cache for its two threads". Half of 2 MiB is 1 MiB. The PR body repeats the same arithmetic. The two tiles are never live simultaneously — `local_stages` and `fused_stages` are different phases — so "half again for the second tile" does not close the gap either. The table at `:240` sweeps the *matrix size* at one fixed budget; it does not distinguish 512 KiB from 1 MiB.

The review bar here is "derived, not magic": as written the constant reads derived and is not. If the real reason is headroom for the streaming gather/scatter traffic that also transits L2, say that — it is a good reason.

---

## [MINOR] The sweep table does not isolate the mechanism it is cited for — `binary-dft/src/poly.rs:240`

```text
matrix      4 MiB   16 MiB   64 MiB   128 MiB   256 MiB   512 MiB
ratio        1.48     0.95     1.01      1.04      0.98      0.89
```

The claim is "past the shared cache the deep tile pays". But the deep tile **wins 5% at 16 MiB**, deep inside a 105 MiB LLC, and **loses 1–4% at 64 and 128 MiB**, also inside it. The curve is non-monotone. Against the stated ±2–3% criterion noise the only two significant points are 4 MiB (+48%) and 512 MiB (−11%), and the 4 MiB point is separately explained by tile starvation, not by cache level.

So the 128 MiB placement is a curve fit bracketed by two points — which the Caveat section admits honestly. The doc comment, however, presents it as a cache-level argument. The confound is that the deep budget moves four things at once (`local`, `log_block`, `depth`, tiles-per-worker) and nothing separates them. Not a merge blocker — the production shapes were measured end-to-end — but the doc should not claim more than the data shows.

---

## [MINOR] `DEEP_TILE_BYTES` already exists in the sibling module with a different meaning — `binary-dft/src/lch.rs:57`

```rust
// lch.rs:57 — "The byte budget of one contiguous row tile."
const DEEP_TILE_BYTES: usize = 128 * 1024;
// poly.rs:250 — "Bytes both tiles may occupy for a matrix the shared cache cannot hold."
const DEEP_TILE_BYTES: usize = 512 * 1024;
```

Same crate, sibling implementations of the same transform, same name, 4x different value, different role — `lch.rs`'s is the analogue of `poly.rs`'s `TILE_BYTES`, and at `lch.rs:587` it doubles as a "the whole matrix fits one tile" test. Worse, `lch.rs:63` says a 32 KiB–256 KiB sweep *flattens* at 128 KiB for that tile, which reads as a direct contradiction until you realise the names mean different things.

**Fix:** rename the new one for the condition rather than the depth — `UNCACHED_TILE_BYTES` or `STREAMING_TILE_BYTES`.

---

## [MINOR] The gate sees one coset, not the codeword — `binary-dft/src/poly.rs:977` and `:1035`

`ntt_batch_padded` builds `Plan::new(width, log_message)` and `shifted_lde_batch` builds `Plan::new(width, log_n)` from the *message* height, then runs `2^log_inv_rate` cosets concurrently, one rayon task each. The LLC pressure is therefore `min(workers, cosets) * coset_bytes`, but the gate only sees `coset_bytes`.

Concrete miss: width 1, message `2^23` = 128 MiB exactly, rate 1/8. The gate is `>`, so the plan stays shallow — while 1 GiB of codeword is being traversed by 16 workers and no coset is anywhere near resident. That is exactly the regime the PR is chasing, one log-step below the shape it optimised.

The error is one-sided (a coset past 128 MiB means the aggregate is past it too), so this is a win forgone rather than a regression. Worth a sentence in the doc at minimum, since the reader of `budgets` cannot tell which quantity `log_n` is.

---

## [MINOR] Deep plans shorten the gathered run and multiply the gather's page footprint — `binary-dft/src/poly.rs:482`

Two second-order effects the doc does not price:

1. In 42 of the shapes I swept, the deep plan picks a **shorter** `log_block` than the shallow one, sometimes collapsing to the `STAGED_LINE_BYTES` floor (width 4 @ `2^26`: `log_block` 1 -> 0; width 48 @ `2^19`: 1 -> 0). The search prices this correctly against `STAGED_RUN_BYTES`/`STAGED_GATHER_DOUBLINGS` — a traversal is worth more than a doubling — so it is consistent. But the PR body's *"with the gathered run length preserved"* is true only of the single shape it measured.

2. A gather touches `2^depth` runs, each `row << (top - depth)` bytes apart, hence each in its own page:

```text
                 runs/tile   stride     distinct 4 KiB pages/tile
shallow (11,5,7)     128     4 MiB                  128
deep    (15,5,10)   1024   512 KiB                 1024
```

1024 entries is roughly half of Sapphire Rapids' ~2048-entry STLB, so the measured win is real but the margin is thinner than the cache argument suggests, and it scales as `2^depth`: width 1 @ `2^28` would want 8192. Since the PR's own table shows the win still *growing* at 512 MiB, no cliff was hit on that host — but the host's THP setting is unstated and materially changes this.

---

## [NIT]

- `.max(1)` in `budgets` (`poly.rs:451`) is a no-op: `log2_floor_usize(0)` returns `0` by documented convention (`util/src/lib.rs:34`), identical to `log2_floor_usize(1)`. It mirrors the neighbouring lines at `:479`/`:482` where it *is* load-bearing, which invites a reader to think it changes the gate's semantics here.
- The staging tile now crosses glibc's default mmap threshold (`poly.rs:826`): `Vec::with_capacity(tile_len)` inside `for_each_init` goes from <=64 KiB to <=512 KiB, past the default `M_MMAP_THRESHOLD` of 128 KiB. Rayon calls the init once per *split*, not once per thread, so this is tens of 512 KiB allocate/free cycles per fused group. glibc's dynamic threshold adapts after the first free so the steady state is heap-served, and it is noise against a 512 MiB matrix — but it is 8x more churn and worth knowing about on a non-glibc allocator.
- A measurement row labelled "unchanged" moved 2.6%: the body's table shows Keccak-f `-l 18` at 19.0 -> 19.5 ms, then says *"The two unchanged rows are below the gate and take today's plan, as intended."* If the plan is bit-identical, +2.6% is noise — but the sentence asserts a null result next to a number that moved. Give the spread, or say "within noise".

---

## Test coverage

Beyond the deep-plan execution test above:

**The gate boundary itself is unpinned.** `PLAN_ONLY_CUTS` pins 8 points but none straddles the threshold, so an off-by-one turning `>` into `>=` (or `log2_floor` into `log2_ceil`) would pass. Cheap and exact, since the gate is pure arithmetic on `(row, log_n)`:

```rust
#[test]
fn the_deep_budget_starts_exactly_past_the_shared_cache() {
    for width in [1usize, 3, 5, 16, 48, 1024] {
        let row = core::mem::size_of::<u128>() * width;
        for log_n in 0..40 {
            let deep = Plan::budgets(row, log_n).0 == DEEP_TILE_BYTES;
            let past = row.checked_shl(log_n as u32).is_none_or(|b| b > SHARED_CACHE_BYTES);
            assert_eq!(deep, past && log_n > 0, "width {width} log_n {log_n}");
        }
    }
}
```

**No benchmark covers the production shape.** `binary-dft/benches/ntt.rs:36` caps at `LOG_HEIGHTS = [14, 16, 18, 20]`, and `NARROW_WIDTH = 1` only appears in the `ntt` group. The single committed bench point past the gate is `encode/BinaryField128/20` at `WIDTH = 16` (256 MiB). The shape the PR actually optimises — width 1 past 128 MiB — has no bench, so the −9.2% cannot be reproduced with `cargo bench`. Adding width 1 @ `2^24` (256 MiB, the first deep width-1 height) to `bench_encode` would make the win reproducible without the 512 MiB cost.

**Proptest:** I would not add one for the transform. The property is `dft(x) == per_stage_schedule(x)` and it already has an exhaustive deterministic sweep; randomising `(width, log_n)` cannot reach the deep branch without allocating >128 MiB per case. The plan-arithmetic invariants (`log_block + depth <= top`, `tile_len | len`, tiles >= 256) *are* proptest-shaped and cheap — a `proptest!` over `(width in 1..2048usize, log_n in 0..40usize)` asserting them on `Plan::for_workers` would be a better use of the tool.

---

## Could not verify

- **Was anything measured on AMD?** All numbers are one Sapphire Rapids host. The Zen 2/3 configuration (512 KiB L2, >=128 MiB aggregate L3) is the one the gate mis-serves, and I have no such machine here.
- **THP setting on the c7i.4xlarge?** The deep plan's gather touches 8x more distinct pages per tile. With `transparent_hugepage=always` the result is robust; with `madvise` and no `madvise` call in the crate, it is 1024 4 KiB STLB entries per tile at the production shape.
- **Is "width 1, 2^25 rows" the message or the codeword height?** `ntt_batch_padded` plans from `log_message`, so I read it as the message — meaning a 512 MiB per-coset traversal and a 4x margin over the gate. Confirming this would also settle whether the per-coset point above ever bites in production.
- **What is `SHARED_CACHE_BYTES` meant to model on a chiplet part?** On Zen the LLC a core can actually reach is its CCX's 16–32 MiB, not the 256 MiB the part advertises. If the intent is "the cache a traversal is served from", 128 MiB is wrong for that whole family in the *conservative* direction — fine, but worth saying out loud next to the caveat.

For the record, I re-ran `cargo test -p p3-binary-dft --release --lib poly::` (19 passed), `cargo clippy -p p3-binary-dft --all-targets --all-features` (clean) and `cargo fmt --check` (clean) at the PR head.

---

**Bottom line:** merge after fixes — add the deep-plan execution test, and either lower `DEEP_TILE_BYTES` to 256 KiB or narrow the "safe direction" caveat to admit the small-L2 / large-L3 case. The math is sound and I verified bit-exactness at the real production magnitude; what is missing is verification, not correctness.

binary-dft/src/poly.rs
```diff
@@ -222,6 +222,44 @@ const TILE_BYTES: usize = 32 * 1024;
...
+///     ratio        1.48     0.95     1.01      1.04      0.98      0.89
+/// ```
+///
+/// The four smallest are width 16 and the two widest a single column, so the curve is one in
+/// the matrix size rather than two in the height.
+///
+/// The 4 MiB entry is a second reason the deep budget is not for a matrix the cache holds: a
+/// tile that large leaves a matrix that small fewer tiles than there are workers to take
+/// them.
+const DEEP_TILE_BYTES: usize = 512 * 1024;
```
tcoratger: **[MAJOR] The deep budget is sized by L2 but gated on L3 — two independent axes — `binary-dft/src/poly.rs:250` and `:261`**

`SHARED_CACHE_BYTES` decides *whether* the deep budget applies; `DEEP_TILE_BYTES` is sized as a fraction of **private L2**. These are unrelated hardware parameters, and the doc's safety argument only covers one of them:

> A target whose shared cache is smaller keeps the shallower tiles … which is a win forgone rather than a cost added — `poly.rs:258`

That is true for *small-LLC* targets. It says nothing about a **large-LLC, small-L2** target, which is where the gate fires and the tile does not fit:

```text
                      L2/core   threads/core   L2/thread   aggregate L3   gate fires?
Xeon 8488C (measured)  2 MiB          2          1 MiB        ~105 MiB      at >128 MiB
EPYC 7763 (Milan)    512 KiB          2        256 KiB         256 MiB      at >128 MiB
EPYC 7742 (Rome)     512 KiB          2        256 KiB         256 MiB      at >128 MiB
```

On Milan a 256 MiB matrix passes the gate and each worker then streams a 512 KiB tile — **100% of the core's entire L2, 200% under SMT.** The sibling constant's own doc states the requirement this breaks: *"What it must not do is spill out of the private cache level below the shared one"* (`poly.rs:213`). The tile then re-reads from L3 on each of its `depth` sub-stages, so the fusion buys much less than the traversal it spent.

**Fix:** decouple the two. Either drop `DEEP_TILE_BYTES` to 256 KiB (fits a 512 KiB L2 with SMT, still 8x today's staging budget and still >=256 tiles), or keep 512 KiB and narrow the caveat to say explicitly that it assumes >=1 MiB of L2 per hardware thread and that a smaller-L2 target is a *cost added*, not a win forgone.

**Verify:** one `binary-dft` `encode/BinaryField128` run at `2^20` × width 16 (the one bench point past the gate) on any Zen 2/3 EPYC, base vs branch.

binary-dft/src/poly.rs
```diff
@@ -1126,19 +1178,22 @@ mod tests {
...
     /// ```
     ///
-    /// The last two offer no leftover to trade, so the traversal count alone picks their run.
-    /// Width 16 at `2^20` rows is `2^24` elements, which is why none of these is transformed.
-    const PLAN_ONLY_CUTS: [(usize, usize, (usize, usize, usize)); 7] = [
-        (48, 18, (5, 0, 6)),
+    /// The first and the last two are past [`SHARED_CACHE_BYTES`], so they are the rows that
+    /// read the deep budget. Width 16 at `2^20` rows is `2^24` elements, which is why none of
+    /// these is transformed.
+    const PLAN_ONLY_CUTS: [(usize, usize, (usize, usize, usize)); 8] = [
```
tcoratger, line 1235: **[MAJOR] No test ever executes a plan the deep branch produces — `binary-dft/src/poly.rs:1188`**

Every shape that is actually transformed in the suite sits below the gate:

- `PRODUCTION_CUTS` — largest is width 1 @ `2^20` = 16 MiB. All shallow.
- `every_width_and_height_matches_the_per_stage_schedule` — sweeps to `log_n = 14`. All shallow.
- `cut_plans()` — synthetic, hand-written triples; max `depth` is 3, max `local` is 2.
- `PLAN_ONLY_CUTS` — the three deep rows are asserted as *triples only*. Nothing runs them.

So the deep branch is covered by an equality against a table that this same PR wrote. `local = 15`, `depth = 10`, `log_block = 5` are magnitudes no executed test reaches. The PR's correctness section cites the two oracle tests, but neither sees a deep plan.

I ran the real thing to close the gap myself — `Plan::for_workers(1, 25, 32)` against `per_stage_schedule`, both directions: **passes**, in 595 s. Too slow for CI, which is presumably why these were kept plan-only. But the *structural* regimes are cheap: the test below passed in **1.15 s** and covers both new regimes at their real magnitudes — the deep group depths and run lengths `(0,9) (2,9) (5,10)`, and the deep contiguous-tile depths `local = 9, 11, 15`:

```rust
#[test]
fn a_deep_budget_cut_matches_the_per_stage_schedule() {
    // (a) the deep group depths and run lengths, over a shallow contiguous tile.
    for (width, log_block, depth) in [(48usize, 0usize, 9usize), (16, 2, 9), (1, 5, 10)] {
        for extra in [0usize, 1, 2] {
            let (local, log_n) = (1usize, 1 + depth + extra);
            let plan = Plan { width, log_n, local, log_block, depth };
            for inverse in [false, true] {
                let mut expected = coefficients(log_n, width);
                let mut actual = expected.clone();
                per_stage_schedule(&mut expected, width, log_n, test_shift(), inverse);
                scheduled(&mut actual, plan, inverse, NONE);
                assert_eq!(actual, expected, "{plan:?} inverse={inverse}");
            }
        }
    }
    // (b) the deep contiguous tile depths, with a shallow group above them.
    for (width, local) in [(1usize, 15usize), (16, 11), (48, 9)] {
        for extra in [1usize, 2, 3] {
            let (log_block, depth) = (0usize, 2usize);
            let log_n = local + extra;
            let plan = Plan { width, log_n, local, log_block, depth };
            for inverse in [false, true] { /* as above */ }
        }
    }
}
```

Optionally also `const _: () = assert!(SHARED_CACHE_BYTES / DEEP_TILE_BYTES >= 256);` so the tile-count invariant the PR's rationale depends on is enforced rather than argued in prose.

binary-dft/src/poly.rs
```diff
@@ -222,6 +222,44 @@ const TILE_BYTES: usize = 32 * 1024;
...
 const STAGING_BYTES: usize = 64 * 1024;
 
+/// Bytes both tiles may occupy for a matrix the shared cache cannot hold.
+///
+/// Every stage a tile fuses is a traversal of the matrix the transform does not take. Inside
+/// the shared cache those traversals are served from it, and a tile deep enough to remove one
+/// buys a level of cache for a traversal that was cheap. Past it every traversal reaches
+/// memory, and the deeper tile is what keeps the transform off it.
+///
+/// A tile this deep runs a level further out than the budgets above, so it is what the
```
tcoratger, line 232: **[MINOR] `DEEP_TILE_BYTES`'s stated derivation yields 1 MiB, not 512 KiB — `binary-dft/src/poly.rs:232`**

The doc says the budget is "one hardware thread's share, which is half of its core's … on a core holding 2 MiB of private cache for its two threads". Half of 2 MiB is 1 MiB. The PR body repeats the same arithmetic. The two tiles are never live simultaneously — `local_stages` and `fused_stages` are different phases — so "half again for the second tile" does not close the gap either. The table at `:240` sweeps the *matrix size* at one fixed budget; it does not distinguish 512 KiB from 1 MiB.

The review bar here is "derived, not magic": as written the constant reads derived and is not. If the real reason is headroom for the streaming gather/scatter traffic that also transits L2, say that — it is a good reason.

binary-dft/src/poly.rs
```diff
@@ -222,6 +222,44 @@ const TILE_BYTES: usize = 32 * 1024;
...
+/// private cache holds rather than what the level under it holds: one hardware thread's
+/// share, which is half of its core's where two threads sit on one core.
+///
+/// A sweep of the budget, as a ratio to the pair above, on a core holding 2 MiB of private
+/// cache for its two threads and 105 MiB of shared cache for its eight:
+///
+/// ```text
+///     matrix      4 MiB   16 MiB   64 MiB   128 MiB   256 MiB   512 MiB
```
tcoratger: **[MINOR] The sweep table does not isolate the mechanism it is cited for — `binary-dft/src/poly.rs:240`**

```text
matrix      4 MiB   16 MiB   64 MiB   128 MiB   256 MiB   512 MiB
ratio        1.48     0.95     1.01      1.04      0.98      0.89
```

The claim is "past the shared cache the deep tile pays". But the deep tile **wins 5% at 16 MiB**, deep inside a 105 MiB LLC, and **loses 1–4% at 64 and 128 MiB**, also inside it. The curve is non-monotone. Against the stated ±2–3% criterion noise the only two significant points are 4 MiB (+48%) and 512 MiB (−11%), and the 4 MiB point is separately explained by tile starvation, not by cache level.

So the 128 MiB placement is a curve fit bracketed by two points — which the Caveat section admits honestly. The doc comment, however, presents it as a cache-level argument. The confound is that the deep budget moves four things at once (`local`, `log_block`, `depth`, tiles-per-worker) and nothing separates them. Not a merge blocker — the production shapes were measured end-to-end — but the doc should not claim more than the data shows.

binary-dft/src/poly.rs
```diff
@@ -423,11 +475,11 @@ impl Plan {
...
     fn for_workers(width: usize, log_n: usize, workers: usize) -> Self {
         let row = core::mem::size_of::<u128>() * width;
-        let local = log2_floor_usize((TILE_BYTES / row).max(1)).min(log_n);
+        let (tile, staging) = Self::budgets(row, log_n);
+        let local = log2_floor_usize((tile / row).max(1)).min(log_n);
         let above = log_n - local;
         // Stages one tile fuses, with a run of `2^log_block` adjacent rows as its row.
-        let depth_at =
-            |log_block: usize| log2_floor_usize((STAGING_BYTES / (row << log_block)).max(1));
+        let depth_at = |log_block: usize| log2_floor_usize((staging / (row << log_block)).max(1));
```
tcoratger, line 529: **[MINOR] Deep plans shorten the gathered run and multiply the gather's page footprint — `binary-dft/src/poly.rs:482`**

Two second-order effects the doc does not price:

1. In 42 of the shapes I swept, the deep plan picks a **shorter** `log_block` than the shallow one, sometimes collapsing to the `STAGED_LINE_BYTES` floor (width 4 @ `2^26`: `log_block` 1 -> 0; width 48 @ `2^19`: 1 -> 0). The search prices this correctly against `STAGED_RUN_BYTES`/`STAGED_GATHER_DOUBLINGS` — a traversal is worth more than a doubling — so it is consistent. But the PR body's *"with the gathered run length preserved"* is true only of the single shape it measured.

2. A gather touches `2^depth` runs, each `row << (top - depth)` bytes apart, hence each in its own page:

```text
                 runs/tile   stride     distinct 4 KiB pages/tile
shallow (11,5,7)     128     4 MiB                  128
deep    (15,5,10)   1024   512 KiB                 1024
```

1024 entries is roughly half of Sapphire Rapids' ~2048-entry STLB, so the measured win is real but the margin is thinner than the cache argument suggests, and it scales as `2^depth`: width 1 @ `2^28` would want 8192. Since the PR's own table shows the win still *growing* at 512 MiB, no cliff was hit on that host — but the host's THP setting is unstated and materially changes this.

tcoratger (commented):

---

Reviewed against the stack base (`perf/zc-lazy-unslice`). Planner-only change: the executor is untouched. I ran the real production deep plan (width 1 @ `2^25`, `(local 15, log_block 5, depth 10)`) against the per-stage oracle in both directions — **bit-exact**. No soundness or safety issue. What is weak is the perf model and the verification, not the correctness.

---

## Verified positively

- **The gate is exact.** `log_n > log2_floor_usize((SHARED_CACHE_BYTES / row).max(1))` is equivalent to `row * 2^log_n > SHARED_CACHE_BYTES`, both directions, under integer floor division. Comparing in rows rather than computing `row << log_n` correctly dodges a shift overflow at large `log_n`.
- **The deep budget never costs a traversal.** I reimplemented `budgets`/`groups`/`traversals`/the `log_block` search independently, reproduced all 8 `PLAN_ONLY_CUTS` triples exactly, then swept widths 1..1199 × matrices up to 8 GiB: **zero** shapes where the deep plan takes more passes than the shallow one.
- **The parallelism claim holds.** Over every deep shape in that sweep, `min(contiguous tiles, staging tiles) = 512`; the guaranteed bound is `SHARED_CACHE_BYTES / DEEP_TILE_BYTES = 256`. The 4 MiB / 8-tiles-for-16-workers cliff cannot recur behind this gate.
- **The `unsafe` invariants survive.** `Rows::gather`/`scatter` rest on `stride << depth == 1 << (top - log_block)` and `tile_len | len`, both only `debug_assert`ed. Both are algebraic identities given `log_block + depth <= top`, which the `min` at `poly.rs:757` enforces unconditionally. Deeper plans change neither. `Rows::check` remains a hard assert, so even a mis-derived plan bounds-faults rather than writing OOB.
- On the deep path `local < log_n` strictly (in fact `above >= 8`), so `tile_exit`, `leftover` and the group walk keep their shape. No new inner loop, no new `unsafe`, no new allocation site.

---

### [MINOR] `DEEP_TILE_BYTES` already exists in the sibling module with a different meaning — `binary-dft/src/lch.rs:57`

```rust
// lch.rs:57 — "The byte budget of one contiguous row tile."
const DEEP_TILE_BYTES: usize = 128 * 1024;
// poly.rs:250 — "Bytes both tiles may occupy for a matrix the shared cache cannot hold."
const DEEP_TILE_BYTES: usize = 512 * 1024;
```

Same crate, sibling implementations of the same transform, same name, 4x different value, different role — `lch.rs`'s is the analogue of `poly.rs`'s `TILE_BYTES`, and at `lch.rs:587` it doubles as a "the whole matrix fits one tile" test. Worse, `lch.rs:63` says a 32 KiB–256 KiB sweep *flattens* at 128 KiB for that tile, which reads as a direct contradiction until you realise the names mean different things.

**Fix:** rename the new one for the condition rather than the depth — `UNCACHED_TILE_BYTES` or `STREAMING_TILE_BYTES`.

### [MINOR] The gate sees one coset, not the codeword — `binary-dft/src/poly.rs:977` and `:1035`

`ntt_batch_padded` builds `Plan::new(width, log_message)` and `shifted_lde_batch` builds `Plan::new(width, log_n)` from the *message* height, then runs `2^log_inv_rate` cosets concurrently, one rayon task each. The LLC pressure is therefore `min(workers, cosets) * coset_bytes`, but the gate only sees `coset_bytes`.

Concrete miss: width 1, message `2^23` = 128 MiB exactly, rate 1/8. The gate is `>`, so the plan stays shallow — while 1 GiB of codeword is being traversed by 16 workers and no coset is anywhere near resident. That is exactly the regime the PR is chasing, one log-step below the shape it optimised.

The error is one-sided (a coset past 128 MiB means the aggregate is past it too), so this is a win forgone rather than a regression. Worth a sentence in the doc at minimum, since the reader of `budgets` cannot tell which quantity `log_n` is.

### [NIT]

- `.max(1)` in `budgets` (`poly.rs:451`) is a no-op: `log2_floor_usize(0)` returns `0` by documented convention (`util/src/lib.rs:34`), identical to `log2_floor_usize(1)`. It mirrors the neighbouring lines at `:479`/`:482` where it *is* load-bearing, which invites a reader to think it changes the gate's semantics here.
- The staging tile now crosses glibc's default mmap threshold (`poly.rs:826`): `Vec::with_capacity(tile_len)` inside `for_each_init` goes from <=64 KiB to <=512 KiB, past the default `M_MMAP_THRESHOLD` of 128 KiB. Rayon calls the init once per *split*, not once per thread, so this is tens of 512 KiB allocate/free cycles per fused group. glibc's dynamic threshold adapts after the first free so the steady state is heap-served, and it is noise against a 512 MiB matrix — but it is 8x more churn and worth knowing about on a non-glibc allocator.
- A measurement row labelled "unchanged" moved 2.6%: the body's table shows Keccak-f `-l 18` at 19.0 -> 19.5 ms, then says *"The two unchanged rows are below the gate and take today's plan, as intended."* If the plan is bit-identical, +2.6% is noise — but the sentence asserts a null result next to a number that moved. Give the spread, or say "within noise".

---

## Test coverage

Beyond the deep-plan execution test above:

**The gate boundary itself is unpinned.** `PLAN_ONLY_CUTS` pins 8 points but none straddles the threshold, so an off-by-one turning `>` into `>=` (or `log2_floor` into `log2_ceil`) would pass. Cheap and exact, since the gate is pure arithmetic on `(row, log_n)`:

```rust
#[test]
fn the_deep_budget_starts_exactly_past_the_shared_cache() {
    for width in [1usize, 3, 5, 16, 48, 1024] {
        let row = core::mem::size_of::<u128>() * width;
        for log_n in 0..40 {
            let deep = Plan::budgets(row, log_n).0 == DEEP_TILE_BYTES;
            let past = row.checked_shl(log_n as u32).is_none_or(|b| b > SHARED_CACHE_BYTES);
            assert_eq!(deep, past && log_n > 0, "width {width} log_n {log_n}");
        }
    }
}
```

**No benchmark covers the production shape.** `binary-dft/benches/ntt.rs:36` caps at `LOG_HEIGHTS = [14, 16, 18, 20]`, and `NARROW_WIDTH = 1` only appears in the `ntt` group. The single committed bench point past the gate is `encode/BinaryField128/20` at `WIDTH = 16` (256 MiB). The shape the PR actually optimises — width 1 past 128 MiB — has no bench, so the −9.2% cannot be reproduced with `cargo bench`. Adding width 1 @ `2^24` (256 MiB, the first deep width-1 height) to `bench_encode` would make the win reproducible without the 512 MiB cost.

**Proptest:** I would not add one for the transform. The property is `dft(x) == per_stage_schedule(x)` and it already has an exhaustive deterministic sweep; randomising `(width, log_n)` cannot reach the deep branch without allocating >128 MiB per case. The plan-arithmetic invariants (`log_block + depth <= top`, `tile_len | len`, tiles >= 256) *are* proptest-shaped and cheap — a `proptest!` over `(width in 1..2048usize, log_n in 0..40usize)` asserting them on `Plan::for_workers` would be a better use of the tool.

---

## Could not verify

- **Was anything measured on AMD?** All numbers are one Sapphire Rapids host. The Zen 2/3 configuration (512 KiB L2, >=128 MiB aggregate L3) is the one the gate mis-serves, and I have no such machine here.
- **THP setting on the c7i.4xlarge?** The deep plan's gather touches 8x more distinct pages per tile. With `transparent_hugepage=always` the result is robust; with `madvise` and no `madvise` call in the crate, it is 1024 4 KiB STLB entries per tile at the production shape.
- **Is "width 1, 2^25 rows" the message or the codeword height?** `ntt_batch_padded` plans from `log_message`, so I read it as the message — meaning a 512 MiB per-coset traversal and a 4x margin over the gate. Confirming this would also settle whether the per-coset point above ever bites in production.
- **What is `SHARED_CACHE_BYTES` meant to model on a chiplet part?** On Zen the LLC a core can actually reach is its CCX's 16–32 MiB, not the 256 MiB the part advertises. If the intent is "the cache a traversal is served from", 128 MiB is wrong for that whole family in the *conservative* direction — fine, but worth saying out loud next to the caveat.

For the record, I re-ran `cargo test -p p3-binary-dft --release --lib poly::` (19 passed), `cargo clippy -p p3-binary-dft --all-targets --all-features` (clean) and `cargo fmt --check` (clean) at the PR head.

---

**Bottom line:** merge after fixes — add the deep-plan execution test, and either lower `DEEP_TILE_BYTES` to 256 KiB or narrow the "safe direction" caveat to admit the small-L2 / large-L3 case. The math is sound and I verified bit-exactness at the real production magnitude; what is missing is verification, not correctness.

binary-dft/src/poly.rs
Nashtare: Confirmed on hardware. Host: `c6a.4xlarge`, AMD EPYC 7R13 (Milan, Zen 3) — `lscpu` reports 512 KiB of L2 per core, 2 threads per core, 32 MiB of L3, THP `madvise`. One caveat on the setup: that slice reports 32 MiB of LLC rather than the 256 MiB of a full 7763, so it tests the half of your claim that decides the question (the tile against a 256 KiB-per-thread private cache) and not the half where a 256 MiB LLC still holds the matrix. The gate's own premise holds there either way.

`encode/BinaryField128`, criterion point estimates, base = `67f7ce78`:

```text
    budget          2^18 (below the gate)   2^20 (past it)
    shallow pair      67.21 / 66.37 ms      267.84 / 268.23 ms
    256 KiB                66.12 ms                264.80 ms
    512 KiB (branch)  66.22 / 66.17 ms      270.85 / 270.36 ms
    1 MiB                  66.51 ms                274.49 ms
```

The branch as posted is a **1.0% regression** at the point past the gate, 1 MiB is 2.4%, and 256 KiB is a 1.2% win. The curve is monotone in the budget, the repeated arms land within 0.2% of each other, and the 2^18 control — below the gate, so every arm plans the same cut — is flat. So the span is the budget, and your reading of the mechanism is what the numbers show: on that core the tile costs more than the traversal its depth removes.

Took the first fix. `DEEP_TILE_BYTES` is 256 KiB in `f337a425`, and the doc now derives it from a hardware thread's share of the private cache, with a table of where that lands on 512 KiB, 1 MiB and 2 MiB per core.

The win the original sweep measured should survive the drop: at the two shapes that carry it the 256 KiB plan keeps the traversal the 512 KiB plan buys — width 1 @ 2^25 goes `(14, 3, 11)` against `(15, 5, 10)`, both one traversal above the tile where the shallow `(11, 5, 7)` takes two, and 2^26 likewise. It gives up run length, not traversals. What I could not measure is the absolute ratio at 256 KiB on the 2 MiB-per-core host the sweep ran on.

binary-dft/src/poly.rs
Nashtare: Added, as `a_deep_budget_cut_matches_the_per_stage_schedule` in `fa5bb934`, on your structure. Two changes to it:

- The heights are `local + depth + extra` with `local = log_block + 1` rather than `1 + depth + extra`. `fused_stages` clamps the run to `plan.log_block.min(top - depth)`, so at `local = 1` every run in (a) was clamped back to 1..3 and the run lengths the table names were never the ones that ran — the depths were real, the runs were not. Lifting the tile one stage above the run leaves them standing. The matrices stay at 4 MiB or under.
- The magnitudes are the ones the budget produces after `f337a425`: groups `(48, 1, 7) (16, 0, 10) (1, 3, 11)`, tiles `local = 14, 10, 8`.

3.76 s under `cargo test`, 0.09 s in release. Mutation-checked rather than assumed: perturbing the tile's twiddle seed (`block << s` in `tile_stages`) fails it.

`const _: () = assert!(SHARED_CACHE_BYTES / DEEP_TILE_BYTES >= 256);` is in as well, beside `SHARED_CACHE_BYTES`. It holds at 512 with the smaller tile.

binary-dft/src/poly.rs
Nashtare: The arithmetic was wrong and the second reading does not rescue it: `local_stages` and `fused_stages` are separate phases, so the two tiles are never live together and there is no second tile to halve for.

Rewritten in `f337a425`. The half is the gather and the scatter, which stream the matrix through the same private cache for as long as the tile is live, so a tile sized to a thread's whole share evicts itself against its own traffic. With the figure now at 256 KiB the derivation is half of a thread's share on a core holding 1 MiB for two threads, and the doc tabulates where that lands on a 512 KiB, 1 MiB and 2 MiB core beside the budget sweep that fixes it.

binary-dft/src/poly.rs
Nashtare: Agreed, and the table was in the wrong place besides: it is evidence about where the gate goes, not about how large the tile is. Moved to `SHARED_CACHE_BYTES` in `e9ca2ef5`, and read there for what it shows — only the two ends carry against a noise floor of a few percent, the middle is flat inside it and not monotone, the figure is a bracket between the last matrix that does not pay for the deeper tile and the first that does rather than a reading of where the shared cache ends, and the 4 MiB end has tile starvation as a cause that is not the cache level at all.

`DEEP_TILE_BYTES` now carries a sweep of its own, over the budget at one shape past the gate, which is the axis it is actually the constant for.

binary-dft/src/poly.rs
Nashtare: Both reproduced.

The run length: width 4 @ 2^26 plans `log_block` 1 shallow and 0 deep, and width 48 @ 2^19 goes 1 -> 0 the same way. The run is what the added depth is bought with, so this is not incidental — it happens wherever the traversal count leaves the search nothing else to spend.

The pages: `Rows::gather` walks `2^depth` runs `stride * run` elements apart, which is `row << (top - depth)` bytes, and at this budget that is far past a page, so every run lands in a page of its own and the count is `2^depth` rather than the tile's own size in pages. The worst case on the 512 KiB branch was width 1 @ 2^28 at depth 13, i.e. 8192; at 256 KiB that shape plans depth 8. The host I measured on reports THP `madvise`, so 4 KiB pages there.

`Plan::budgets` now prices both in `e9ca2ef5`: the run-length trade, with the note that the search ranks a traversal above a doubling rather than ignoring the trade, and the page count as the thing that bounds the depth from above once the private cache no longer does.

The PR body's "with the gathered run length preserved" is being corrected — it was true of the one shape measured and of nothing else.

tcoratger (approved):
Re-reviewed at `0ce5a195` (delta `26f2321b..HEAD` in `binary-dft/src/poly.rs`). **Both MAJORs are fixed.** Approving.

### Prior findings

- **[MAJOR] deep budget sized by L2, gated on L3 — fixed.** `DEEP_TILE_BYTES` is 256 KiB, and you measured the axis on the machine the finding was about. The doc's new table is arithmetically true on all three rows (512 KiB/2 -> 256 KiB, the whole share; 1 MiB/2 -> half; 2 MiB/2 -> a quarter), the halving is now justified by the gather/scatter stream rather than by a second tile, and the budget sweep `0.99 / 1.01 / 1.02` reproduces your EPYC point estimates against the shallow pair. `const _: () = assert!(SHARED_CACHE_BYTES / DEEP_TILE_BYTES >= 256)` holds at 512.
- **[MAJOR] no test executes a deep plan — fixed, and your correction to my test is right.** `top` starts at `log_n`, not at `above`, so with `local = 1` the clamp `plan.log_block.min(top - depth)` cut the runs to `1..3` and my table's run lengths never ran; `local = log_block + 1` makes `top - depth = local + extra > log_block`, so they stand. I checked the magnitudes are deep-only rather than merely large: at `STAGING_BYTES`/`TILE_BYTES` the same shapes give `depth 9 / 8 / 5` and `local 11 / 8 / 6`, so `(48,1,7) (16,0,10) (1,3,11)` and `local = 14, 10, 8` are reachable only through the deep branch, and each is pinned to `Plan::for_workers(..., 32)` by `the_plan_gathers_the_longest_run_its_budgets_allow` via the deep rows of `PLAN_ONLY_CUTS`. Gate + execution are covered between the two tests. `cargo test -p p3-binary-dft --lib poly::tests` — 20 passed; clippy and fmt clean.
- **[MINOR] derivation yielded 1 MiB — fixed.** The stream is the honest reason and the arithmetic now closes.
- **[MINOR] sweep table cited for the wrong mechanism — fixed.** Moved to `SHARED_CACHE_BYTES` and read for what it shows (two ends carry, middle flat and non-monotone, bracket rather than cache level, 4 MiB explained by tile starvation).
- **[MINOR] run length and page footprint — fixed.** `Plan::budgets` prices both, and the PR body no longer claims the run is preserved.

### Still open (all MINOR/NIT, none blocking)

- `DEEP_TILE_BYTES` still collides by name with `lch.rs:57` (128 KiB, the analogue of `poly.rs`'s `TILE_BYTES`). `UNCACHED_TILE_BYTES` would still read better.
- The gate sees one coset, not the codeword (`Plan::new(width, log_message)` at `poly.rs:1024`); one sentence in `budgets` would tell the reader which quantity `log_n` is.
- `.max(1)` in `budgets` is a no-op, and the gate-boundary test (`budgets(row, log_n).0 == DEEP_TILE_BYTES` iff `row << log_n > SHARED_CACHE_BYTES`) is still not there; the 8 rows of `PLAN_ONLY_CUTS` straddle the gate at three widths, which is most of the value.
- New, minor: `DEEP_BUDGET_GROUPS` / `DEEP_BUDGET_TILES` are transcribed from `PLAN_ONLY_CUTS` with nothing linking them. Change `DEEP_TILE_BYTES` and `PLAN_ONLY_CUTS` fails loudly while these two silently stop being deep magnitudes, and the new test keeps passing on shallow shapes. Deriving them from the table, or asserting they match its deep rows, keeps the coverage honest.

For the record: my earlier claim that CI had never run on this PR was wrong — I had looked minutes after it opened. That thread is resolved with this correction.

PR #2253, opened 2026-09-20
perf: run the bit ring-switch reduction in the polynomial basis

Nashtare:
Stacked on #2252.

A tower-basis product in `GF(2^128)` is three changes of basis around one carryless multiply; in
the GHASH polynomial basis it is that multiply alone. The zerocheck's later rounds already exploit
this through `ReprBackend`, and the binary PCS fold crosses a whole block over once and runs every
round there. The bit ring-switch reduction did not: it ran its sumcheck in the tower basis while
the inner PCS it feeds runs in the polynomial one.

### `perf(sumcheck): run the bit ring-switch rounds in the polynomial basis`

`BitRingSwitch::prove` takes a representation parameter and drives `ReprSumcheckProver`. The
packing crosses over through the bulk `from_tower_vec` kernel, riding the slot copy
`restricted_packing` already made. `weights_over` builds the weight multilinear directly in the
representation field by carrying the `d`-entry batching table over first — a subset sum is
additive and a characteristic-two isomorphism carries addition, so that is 128 conversions rather
than one per entry. `CoordinateSums<EF>` becomes `CoordinateSums<EF, A>`: the level whose bytes
index the coordinates stays in the tower basis, which is what the `F_2` decomposition is read in,
while the weights move.

The tensor elements, the equality table and the initial sum stay in the tower basis: `tensor_over`
and `CoordinateSums::sum` read their inputs' tower bytes, and the elements are wire format.

### `perf(binary-pcs): read the committed packing through a view`

`BitPacking<EF, S = Vec<EF>>` takes a backing store as `Poly<F, S>` does, with
`BitPackingView<'a, EF>`. `BooleanPcs::packing` returns a view over the `PolyView` the table
already hands back instead of collecting the committed column.

### Measurements

branch runs interleaved.

| | `bit ring switch` | `read committed packing` | prove |
| --- | --- | --- | --- |
| BLAKE3 `-l 18` | 441 ms -> 253 ms | 19.3 ms -> 0.52 us | 3.01 s -> 2.80 s (-7.0%) |
| Keccak-f `-l 18` | 79.5 ms -> 46.2 ms | 3.05 ms -> 0.53 us | 797 ms -> 772 ms (-3.1%) |

The rounds alone go from roughly 240 ms to 41.5 ms at BLAKE3 `-l 18`. The view is worth about
50 ms, more than the span accounts for, the remainder being allocation and first-touch page
faults. Peak RSS is unchanged at 10.24 GB: the maximum is set during `prove_with_lookup`, well
before the packing is read.

What remains inside the reduction at BLAKE3 `-l 18`: the equality table 82.7 ms, `tensor_over`
52.0 ms, rounds 41.5 ms, `weights_over` 39.6 ms, `restricted_packing` 29.4 ms. The equality table
is `Poly::new_from_point` materialising one tower multiply per entry and belongs to
`multilinear-util` rather than here.

### Correctness

Prover-only; proofs are byte-identical to #2246 for BLAKE3 and Keccak-f at `-l 12` and `-l 14`.

`the_polynomial_basis_rounds_play_the_tower_basis_transcript` runs a reduction through both
representations over three point shapes — random, Boolean-prefixed so rounds are dropped, and one
sending successor elements — and compares the tensor, the successor elements, the polynomial
evaluations element by element, the grinding witnesses, the final evaluation, the surviving point
and value, and the next sample off both sponges. `a_borrowed_packing_plays_the_owned_packing_transcript`
does the same across the owned and borrowed packings, since production now takes the borrowed path
exclusively.

`cargo test --workspace --features parallel` passes, clippy and rustfmt are clean.

### Review note

Rust has no partial turbofish, so the storage parameter forces `prove::<R, _, _>` and
`restricted_packing::<R, _>` at 21 call sites. That churn is most of the second commit's line
count and none of its content.


tcoratger (approved):
Reviewed against the stack base (`perf/encode-merkle-fusion`). I reconstructed the reduction, checked the basis boundary line by line, ran the suite, and verified the two load-bearing lemmas experimentally.

**The change is correct and is genuinely prover-internal: the diff touches zero verifier code, so the worst failure mode of a wrong `R` is a proof the honest verifier rejects, not a soundness hole.** Approving; the findings below are documentation, API-tightening and test coverage, none blocking.

## What I verified, and how

The reduction sends `s_hat = sum_w eq(r_high, w) (x) t'(w)` in the tensor algebra, the verifier draws `r_batch` and `alpha`, and both sides run

```
    sum_w  t'(w) * A(w),      A(w) = sum_u eq(u, r_batch) * coord_u( eq(r_high, w) )
```

`coord_u` is the `F_2`-coordinate extraction. That is the *only* basis-sensitive operation, and it is exactly what the PR leaves alone:

```
                     tower basis (EF)                    |   polynomial basis (R)
  ------------------------------------------------------+------------------------------
  point r, r_batch, alpha, tensor s_hat, equality table  |
  eq(r_high, .) entries  --coord_u-->  bit index set  ---+--> sum of phi(w_u)   = A(w)
  packed witness t'  -------------------- from_table ----+--> phi(t')
                                                         |   rounds: phi(t')*A
  round message  <-- into_transcript ---------------------+--  c_a, c_inf
```

`coord_u` is applied to `EF` elements *before* `phi`; only the weight *values* move. Since `phi` is additive, `sum_{u in S} phi(w_u) = phi(sum_{u in S} w_u)`. I tested that lemma directly rather than trusting it — 200 random `BinaryField128` values against `CoordinateSums::<BinaryField128, Ghash128>` vs `CoordinateSums::<BinaryField128, BinaryField128>`, all equal after mapping back.

I also checked the three classic failure modes for a basis change:

- **Applied exactly once, not twice.** `restricted_packing` converts the freshly-collected slot once; `weights_over` converts the `d`-entry batching table and `alpha` once; `from_repr_tables` converts the initial claim once; `into_transcript` is applied once to each round message and once to `final_eval`. No value crosses twice. (This basis change is *not* involutive, so a double application would not hide — but it does not happen.)
- **No transposition.** There is no matrix here: the change of basis is a `u128 -> u128` bijection, and `coord_u`'s index set never moves.
- **One-sidedness.** `git diff` touches no verifier function. `verify_readings` recomputes `A`, the rows and the equality element entirely in `EF`.

Beyond the PR's own test I confirmed, by temporary instrumentation (reverted):

- All three cases in the new test run 5, 3 and 5 rounds — so the Boolean-prefix path is genuinely exercised, not degenerate.
- `prove::<Ghash128>` proofs **verify absolutely** against the true claim in all three shapes, not merely equal the tower proofs.
- The `rounds == 0` edge case (whole kept prefix Boolean — reachable, since `prefix_limit()` is uncapped for a plain reduction) produces a proof that verifies in the polynomial basis.

---

## [MINOR] The doc table that defines the basis boundary is wrong in two of its three rows — `sumcheck/src/ring_switch/bits/reduction.rs:1129-1133`

This table is the reader's map of where the isomorphism sits, which is the entire safety argument for the change. Two of its entries do not match the code.

```rust
///     once per reduction   the restricted packing, the weight multilinear
///     once per round       the challenge, and the round message on its way back
///     never                the elements, the claims, the batching challenges
```

The claims *do* cross: `from_repr_tables` does `SumcheckProver::new(poly, R::from(sum))` on `batch.initial_sum(...)` (`repr.rs:110`). The batching challenges *do* cross: `weights_over` converts the `eq_batch` table (`reduction.rs:824-830`) and `alpha` itself (`reduction.rs:848`). Only the tensor elements, the `EF` equality table and the evaluation point never move.

**Fix:** make the rows factual, e.g.

```rust
///     once per reduction   the restricted packing, the batching table, alpha,
///                          the initial claim
///     once per round       the challenge, and the round message on its way back
///     never                the tensor elements, the equality table, the point
```

**Verify:** grep the body of `prove`, `weights_over` and `from_repr_tables` for `R::from` / `from_table` and check each hit appears in the table.

Worth the sixty seconds before merge, since that table is what the next reader will reason from.

---

## [MINOR] `weights_over` is the one place the two fields mix, and its bound does not carry the isomorphism contract — `sumcheck/src/ring_switch/bits/reduction.rs:817-820`

```rust
fn weights_over<R>(&self, equality: &Poly<EF>) -> Poly<R>
where
    EF: Send + Sync,
    R: Field + From<EF> + Send + Sync,
```

`From<EF>` says only "there is a conversion". `weights_over` needs it to be a *ring* homomorphism: it computes `alpha * base[w-1]` and `alpha.square()` in `R` and relies on those equalling `phi(alpha * A(w-1))` and `phi(alpha^2)`. The crate already has the right bound — `IntoTranscriptField<EF>`, whose doc states "must be mutually inverse field isomorphisms" — and `prove` uses it. `weights_over` is weaker than its caller.

This matters because `weights_over` is the *only* function in the reduction that reads `EF` bytes and sums `R` values in the same expression; everything else is uniformly one field or the other. It is the natural site for a future mistake.

Note also what the existing runtime check does and does not catch: `debug_assert_eq!(r_repr.into_transcript(), r)` (`repr.rs:190`) proves the two maps are mutually inverse *bijections*. A map that is additive but not multiplicative — an arbitrary `F_2`-linear bijection of `GF(2^128)` — passes that assert on every round and still measures the wrong round message. Only the equivalence test catches that class.

**Fix:** `R: IntoTranscriptField<EF> + Sync`. This does not compile as-is for the `pub fn weights(&self) -> Poly<EF>` caller, because `EF: FromTable<EF>` only holds for the fields with an explicit reflexive impl — which is precisely what the next finding proposes to fix with a blanket impl. Land the two together.

---

## [MINOR] The reflexive `FromTable<Self>` impls are a needless full-witness copy, and one was added to production solely to serve a test — `sumcheck/src/strategy/repr.rs:249-251`

```rust
impl FromTable<Self> for BinaryField64 {}
impl FromTable<Self> for BinaryField128 {}   // added by this PR
```

The default body is `table.par_iter().map(Self::from).collect()` then `drop(table)` (`repr.rs:222-227`) — for the identity map that is a full witness-sized allocation plus copy, discarding a buffer that was already correct. `restricted_packing` therefore does *two* full copies on the identity path (one `collect`, one `from_table`), which is the very copy the PR's second commit removes from `BooleanPcs::packing`.

Nothing in production selects either impl: `ChallengeField` maps `BinaryField128 -> Ghash128` and `BinaryField64 -> Poly64` (`binary-pcs/src/fold.rs:300-312`). The `BinaryField128` impl exists only so `both_representations` can run a tower-basis leg. There are also two `#[cfg(test)]`-scoped reflexive impls — `reduction.rs:1416` (`BinaryField16`, added here) and the pre-existing `suffix.rs:1395` (`BabyBearExt4`). These are *crate-global* impls gated on `cfg(test)`, so the crate's trait-impl set differs between `cargo build` and `cargo test`, and two test modules picking the same type is an instant `E0119` that only `--all-targets` reveals.

**Fix:** one blanket impl replaces all four and makes the identity path zero-copy:

```rust
/// The identity map, which needs no new buffer.
impl<T: Field> FromTable<T> for T {
    fn from_table(table: Vec<Self>) -> Vec<Self> {
        table
    }
}
```

**Verify:** I applied exactly this, deleted `repr.rs:249`, `repr.rs:251`, `reduction.rs:1416` and `suffix.rs:1395`, and ran `cargo check -p p3-sumcheck --all-targets` (clean, no coherence conflict — `Poly64 != BinaryField64` so the specific impls stay disjoint) and `cargo test -p p3-sumcheck --lib` (**593 passed, 0 failed**). Reverted afterwards.

---

## [MINOR] The new cross-basis test proves equivalence but never checks the polynomial-basis proof actually verifies — `sumcheck/src/ring_switch/bits/reduction.rs:2638`

`the_polynomial_basis_rounds_play_the_tower_basis_transcript` is well-built — it compares tensor, successor elements, every round polynomial, grinding witnesses, `final_eval`, surviving point and value, *and* the next sample off both sponges. That is the right property for a change whose tower leg is unmodified.

But it is a pure equivalence test: if `tensor_over` or `initial_sum` were broken, both legs would be broken identically and the test would still pass. Three lines make it an absolute test as well, and I confirmed they pass:

```rust
let mut v = WideChal::from_hasher(Vec::new(), Keccak256Hash);
assert!(reduction.verify(&poly, claim, &mut v).is_ok(), "case {case}");
```

with `claim` the multilinear embedding of `witness` evaluated at the point. I ran exactly this against all three cases plus a zero-round case; all verified.

---

## [NIT] The guard in the new test does not assert what its message says — `sumcheck/src/ring_switch/bits/reduction.rs:2663`

```rust
assert!(reduction.num_variables() > 0, "case {case} runs no round");
```

`num_variables()` is the kept-coordinate count, not the round count; the rounds are `num_variables() - prefix`. I measured: case 1 (`addressed`) has `num_variables() == 5` but only 3 rounds. The assertion can never fire for the intended reason.

**Fix:** assert on the round count after proving, `assert!(poly.sumcheck.num_rounds() > 0, ...)`, or drop the message's claim.

---

## On the closed #2247

Nothing to carry: #2247 was closed by its own author purely to re-anchor the branch onto `perf/encode-merkle-fusion`, and received zero reviews. I diffed the two bodies — they are byte-identical except `Stacked on #2246.` -> `Stacked on #2252.` and one trailing newline. So the richer-looking #2247 describes exactly the behaviour #2253 implements; nothing described there is missing here. The `1196/-223` vs `268/-84` gap is entirely the intervening stack (#2249, #2252) having already landed `ReprSumcheckProver`, `FromTable` and `IntoTranscriptField`.

---

## Test coverage

| Gap | Concrete test |
| --- | --- |
| The cross-field subset-sum lemma — the exact claim `CoordinateSums`' new doc makes — is never tested directly. `basis.rs`'s `coordinate_sums_match_the_walk` only exercises `A == EF`. | In `basis.rs::tests`: random `w: [BinaryField128; 128]`, assert `BinaryField128::from(CoordinateSums::<BinaryField128, Ghash128>::new(&w.map(Ghash128::from)).sum(v)) == CoordinateSums::<_, BinaryField128>::new(&w).sum(v)` over 200 random `v`. I ran this; it passes. A `proptest` over `raw: u128` for `v` is strictly better and just as cheap — a failure would localise the bug to `phi` failing additivity, independently of the whole reduction. |
| `rounds == 0` is reachable (a fully Boolean kept prefix, since `prefix_limit()` is uncapped for a plain reduction) and is not covered in either basis. | Set every `high()` coordinate to `ONE`, assert `proof.sumcheck.num_rounds() == 0` and that both `prove::<BinaryField128>` and `prove::<Ghash128>` verify. I ran it; both pass. Worth pinning, because `from_repr_tables` -> `SumcheckProver::new` on a one-element table plus a zero-round `SumcheckShape` is an easy thing for a later refactor to break. |
| No absolute-correctness assertion on the polynomial-basis leg. | See the MINOR above — three lines. |
| No negative test that a claim reduced in one basis is *rejected* in the other. | Deliberately not applicable here, and I want to be explicit about why rather than recommend a test that cannot exist: the basis never reaches the transcript, so there is no "verify in the other basis" to reject. The correct negative statement is a *completeness* one — a non-multiplicative `R` yields a proof the honest verifier rejects — and it is not worth a test, since no such `R` is constructible from the crate's impls. |

---

## What I checked and found clean

- **Fiat-Shamir ordering** is untouched: elements are bound by `transcript.statement(...)` before `batch_drawn`, and the batch is built from the drawn challenges. The `R` parameter enters strictly after.
- **`unsafe`**: this PR adds none. It does newly route the ring-switch's witness table through the pre-existing `Ghash128::from_tower_vec` transmute, so I checked its invariants: both `Ghash128` (`binary-field/src/ghash.rs:89`) and the tower macro's fields (`binary-field/src/tower.rs:135`) are `#[repr(transparent)]`, and `BinaryField128::MASK == u128::MAX` so every source bit pattern is canonical — the safety comment's own precondition. `Vec::from_raw_parts` reuses the original `capacity`, and `size_of` is equal, so the over-allocation a rayon `collect` may leave is handled correctly.
- **Borrow lifetimes in `open_readings`**: `packing` borrows `prover_data`, which is later moved into `try_open_at_known`. NLL ends the borrow at the loop's last use; no restructuring or explicit `drop` was needed, and none was added.
- **New panic surface in `BooleanPcs::packing`**: none. `Table::poly(0)` goes through `dense_matrix()`, which panics on `TableStorage::Boolean` — but the pre-PR code called the identical `poly(0)`, and `BooleanPcs` commits the packed 128-bit table, not a `Gf2` bit-sliced one.
- **Transcript identity of `ReprSumcheckProver`**: I compared it against `SumcheckProver::compute_sumcheck_polynomials` with `constraint: None` (`strategy.rs:1524-1552`). Same `SumcheckShape::new(folding_factor, pow_bits, Basis::Evaluation)`, same `measure_round -> observe_and_sample -> reduce_claim -> hold` sequence, same outstanding-challenge-on-return convention. The dropped `None` argument is the only difference and it is a no-op.
- **The basis is in the type**, which is the structural defence: `Poly<Ghash128>` and `Poly<BinaryField128>` are distinct types, so a value cannot silently be used in the wrong basis. The single leak is `weights_over`'s loose `From<EF>` bound, flagged above.
- **Perf reasoning is sound.** Addition is XOR in both bases, so `weights_over`'s subset-sum accumulation costs the same in `R` as in `EF` — carrying the `d`-entry batching table over instead of the `2^n`-entry result is a real, free win, and `alpha * base[w-1]` becomes a bare `clmul`. `restricted_packing` correctly rides its existing `collect` into the blocked in-place `from_tower_vec` kernel rather than doing a per-element `From` during the copy; that is the right ordering, since the blocked kernel beats 16 table lookups per element.
- **CI**: all checks green at the head SHA, including `Formatting and Clippy`, the `pclmulqdq` and `avx512f+vpclmulqdq` configurations that actually select the fast `Ghash128` kernels, `Slow regression tests` and `binary-pcs 2^16 round trip`.

---

## Questions

- The "byte-identical proofs to #2246 at `-l 12` and `-l 14`" claim is the strongest correctness evidence in the PR and I could not reproduce it in a reasonable budget. Is that comparison scripted anywhere a reviewer can re-run? It is the assertion that covers `weights_over`'s `alpha` branch at production scale — the in-repo tests reach it only at 5 variables.
- `BitRingSwitchBatch::weights()` is `pub` and, after this PR, has no caller outside `reduction.rs`'s own tests. Is it public for a downstream consumer, or is it now effectively test scaffolding that could become `pub(crate)`?

---

**Bottom line:** merge. The math is right, the boundary is in the right place, the verifier is untouched, and I confirmed the two load-bearing lemmas and the absolute correctness of the polynomial-basis proofs by hand. Fixing the `prove` doc table (`reduction.rs:1133`) before merge is worth the sixty seconds; the blanket `FromTable` impl and the `verify` assertion in the new test are good follow-ups.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -1098,6 +1122,20 @@ pub enum BitRingSwitchProofError {
...
     ///
+    /// # The field the rounds run in
+    ///
+    /// Every round multiplies a packing entry by a weight, and both tables are witness-sized.
+    /// `R` is the representation those two tables and the rounds over them are held in.
+    ///
+    /// ```text
+    ///     once per reduction   the restricted packing, the weight multilinear
+    ///     once per round       the challenge, and the round message on its way back
+    ///     never                the elements, the claims, the batching challenges
```
tcoratger: **[MINOR] The doc table that defines the basis boundary is wrong in two of its three rows — `sumcheck/src/ring_switch/bits/reduction.rs:1129-1133`**

This table is the reader's map of where the isomorphism sits, which is the entire safety argument for the change. Two of its entries do not match the code.

```rust
///     once per reduction   the restricted packing, the weight multilinear
///     once per round       the challenge, and the round message on its way back
///     never                the elements, the claims, the batching challenges
```

The claims *do* cross: `from_repr_tables` does `SumcheckProver::new(poly, R::from(sum))` on `batch.initial_sum(...)` (`repr.rs:110`). The batching challenges *do* cross: `weights_over` converts the `eq_batch` table (`reduction.rs:824-830`) and `alpha` itself (`reduction.rs:848`). Only the tensor elements, the `EF` equality table and the evaluation point never move.

**Fix:** make the rows factual, e.g.

```rust
///     once per reduction   the restricted packing, the batching table, alpha,
///                          the initial claim
///     once per round       the challenge, and the round message on its way back
///     never                the tensor elements, the equality table, the point
```

**Verify:** grep the body of `prove`, `weights_over` and `from_repr_tables` for `R::from` / `from_table` and check each hit appears in the table.

Worth the sixty seconds before merge, since that table is what the next reader will reason from.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -798,13 +810,24 @@ impl<EF: TowerLevel> BitRingSwitchBatch<'_, EF> {
...
-    /// The weight multilinear, read off the equality table of the supported run.
-    fn weights_over(&self, equality: &Poly<EF>) -> Poly<EF>
+    /// The weight multilinear, read off the equality table of the supported run, held in `R`.
+    ///
+    /// Every entry is a subset sum of the batching table, so carrying that table into `R`
+    /// first puts the whole multilinear there for the cost of `d` conversions.
+    fn weights_over<R>(&self, equality: &Poly<EF>) -> Poly<R>
     where
         EF: Send + Sync,
+        R: Field + From<EF> + Send + Sync,
```
tcoratger: **[MINOR] `weights_over` is the one place the two fields mix, and its bound does not carry the isomorphism contract — `sumcheck/src/ring_switch/bits/reduction.rs:817-820`**

```rust
fn weights_over<R>(&self, equality: &Poly<EF>) -> Poly<R>
where
    EF: Send + Sync,
    R: Field + From<EF> + Send + Sync,
```

`From<EF>` says only "there is a conversion". `weights_over` needs it to be a *ring* homomorphism: it computes `alpha * base[w-1]` and `alpha.square()` in `R` and relies on those equalling `phi(alpha * A(w-1))` and `phi(alpha^2)`. The crate already has the right bound — `IntoTranscriptField<EF>`, whose doc states "must be mutually inverse field isomorphisms" — and `prove` uses it. `weights_over` is weaker than its caller.

This matters because `weights_over` is the *only* function in the reduction that reads `EF` bytes and sums `R` values in the same expression; everything else is uniformly one field or the other. It is the natural site for a future mistake.

Note also what the existing runtime check does and does not catch: `debug_assert_eq!(r_repr.into_transcript(), r)` (`repr.rs:190`) proves the two maps are mutually inverse *bijections*. A map that is additive but not multiplicative — an arbitrary `F_2`-linear bijection of `GF(2^128)` — passes that assert on every round and still measures the wrong round message. Only the equivalence test catches that class.

**Fix:** `R: IntoTranscriptField<EF> + Sync`. This does not compile as-is for the `pub fn weights(&self) -> Poly<EF>` caller, because `EF: FromTable<EF>` only holds for the fields with an explicit reflexive impl — which is precisely what the next finding proposes to fix with a blanket impl. Land the two together.

sumcheck/src/strategy/repr.rs
```diff
@@ -243,6 +248,8 @@ where
 
 impl FromTable<Self> for BinaryField64 {}
 
+impl FromTable<Self> for BinaryField128 {}
```
tcoratger: **[MINOR] The reflexive `FromTable<Self>` impls are a needless full-witness copy, and one was added to production solely to serve a test — `sumcheck/src/strategy/repr.rs:249-251`**

```rust
impl FromTable<Self> for BinaryField64 {}
impl FromTable<Self> for BinaryField128 {}   // added by this PR
```

The default body is `table.par_iter().map(Self::from).collect()` then `drop(table)` (`repr.rs:222-227`) — for the identity map that is a full witness-sized allocation plus copy, discarding a buffer that was already correct. `restricted_packing` therefore does *two* full copies on the identity path (one `collect`, one `from_table`), which is the very copy the PR's second commit removes from `BooleanPcs::packing`.

Nothing in production selects either impl: `ChallengeField` maps `BinaryField128 -> Ghash128` and `BinaryField64 -> Poly64` (`binary-pcs/src/fold.rs:300-312`). The `BinaryField128` impl exists only so `both_representations` can run a tower-basis leg. There are also two `#[cfg(test)]`-scoped reflexive impls — `reduction.rs:1416` (`BinaryField16`, added here) and the pre-existing `suffix.rs:1395` (`BabyBearExt4`). These are *crate-global* impls gated on `cfg(test)`, so the crate's trait-impl set differs between `cargo build` and `cargo test`, and two test modules picking the same type is an instant `E0119` that only `--all-targets` reveals.

**Fix:** one blanket impl replaces all four and makes the identity path zero-copy:

```rust
/// The identity map, which needs no new buffer.
impl<T: Field> FromTable<T> for T {
    fn from_table(table: Vec<Self>) -> Vec<Self> {
        table
    }
}
```

**Verify:** I applied exactly this, deleted `repr.rs:249`, `repr.rs:251`, `reduction.rs:1416` and `suffix.rs:1395`, and ran `cargo check -p p3-sumcheck --all-targets` (clean, no coherence conflict — `Poly64 != BinaryField64` so the specific impls stay disjoint) and `cargo test -p p3-sumcheck --lib` (**593 passed, 0 failed**). Reverted afterwards.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -2528,4 +2577,113 @@ mod tests {
...
+            borrowed_proof.sumcheck.polynomial_evaluations,
+            owned_proof.sumcheck.polynomial_evaluations
+        );
+        assert_eq!(borrowed_proof.final_eval, owned_proof.final_eval);
+        assert_eq!(borrowed_point, owned_point);
+        assert_eq!(borrowed_value, owned_value);
+    }
+
+    #[test]
+    fn the_polynomial_basis_rounds_play_the_tower_basis_transcript() {
```
tcoratger, line 2646: **[MINOR] The new cross-basis test proves equivalence but never checks the polynomial-basis proof actually verifies — `sumcheck/src/ring_switch/bits/reduction.rs:2638`**

`the_polynomial_basis_rounds_play_the_tower_basis_transcript` is well-built — it compares tensor, successor elements, every round polynomial, grinding witnesses, `final_eval`, surviving point and value, *and* the next sample off both sponges. That is the right property for a change whose tower leg is unmodified.

But it is a pure equivalence test: if `tensor_over` or `initial_sum` were broken, both legs would be broken identically and the test would still pass. Three lines make it an absolute test as well, and I confirmed they pass:

```rust
let mut v = WideChal::from_hasher(Vec::new(), Keccak256Hash);
assert!(reduction.verify(&poly, claim, &mut v).is_ok(), "case {case}");
```

with `claim` the multilinear embedding of `witness` evaluated at the point. I ran exactly this against all three cases plus a zero-round case; all verified.

sumcheck/src/ring_switch/bits/reduction.rs
```diff
@@ -2528,4 +2577,113 @@ mod tests {
...
+        addressed[1] = BinaryField128::ZERO;
+        let addressed = Point::new(addressed);
+        let reductions = [
+            BitRingSwitch::new(&random).unwrap(),
+            BitRingSwitch::new(&addressed).unwrap(),
+            BitRingSwitch::with_successor(&random, absorbed + 1).unwrap(),
+        ];
+
+        for (case, reduction) in reductions.iter().enumerate() {
+            assert!(reduction.num_variables() > 0, "case {case} runs no round");
```
tcoratger: **[NIT] The guard in the new test does not assert what its message says — `sumcheck/src/ring_switch/bits/reduction.rs:2663`**

```rust
assert!(reduction.num_variables() > 0, "case {case} runs no round");
```

`num_variables()` is the kept-coordinate count, not the round count; the rounds are `num_variables() - prefix`. I measured: case 1 (`addressed`) has `num_variables() == 5` but only 3 rounds. The assertion can never fire for the intended reason.

**Fix:** assert on the round count after proving, `assert!(poly.sumcheck.num_rounds() > 0, ...)`, or drop the message's claim.

tcoratger (commented):
Re-posting with findings anchored to their lines; supersedes my earlier body-only review. The APPROVE from my earlier review stands; this pass only re-homes the findings onto the lines they concern.

---

Reviewed against the stack base (`perf/encode-merkle-fusion`). I reconstructed the reduction, checked the basis boundary line by line, ran the suite, and verified the two load-bearing lemmas experimentally.

**The change is correct and is genuinely prover-internal: the diff touches zero verifier code, so the worst failure mode of a wrong `R` is a proof the honest verifier rejects, not a soundness hole.** Approving; the findings below are documentation, API-tightening and test coverage, none blocking.

---

## What I verified, and how

The reduction sends `s_hat = sum_w eq(r_high, w) (x) t'(w)` in the tensor algebra, the verifier draws `r_batch` and `alpha`, and both sides run

```
    sum_w  t'(w) * A(w),      A(w) = sum_u eq(u, r_batch) * coord_u( eq(r_high, w) )
```

`coord_u` is the `F_2`-coordinate extraction. That is the *only* basis-sensitive operation, and it is exactly what the PR leaves alone:

```
                     tower basis (EF)                    |   polynomial basis (R)
  ------------------------------------------------------+------------------------------
  point r, r_batch, alpha, tensor s_hat, equality table  |
  eq(r_high, .) entries  --coord_u-->  bit index set  ---+--> sum of phi(w_u)   = A(w)
  packed witness t'  -------------------- from_table ----+--> phi(t')
                                                         |   rounds: phi(t')*A
  round message  <-- into_transcript ---------------------+--  c_a, c_inf
```

`coord_u` is applied to `EF` elements *before* `phi`; only the weight *values* move. Since `phi` is additive, `sum_{u in S} phi(w_u) = phi(sum_{u in S} w_u)`. I tested that lemma directly rather than trusting it — 200 random `BinaryField128` values against `CoordinateSums::<BinaryField128, Ghash128>` vs `CoordinateSums::<BinaryField128, BinaryField128>`, all equal after mapping back.

I also checked the three classic failure modes for a basis change:

- **Applied exactly once, not twice.** `restricted_packing` converts the freshly-collected slot once; `weights_over` converts the `d`-entry batching table and `alpha` once; `from_repr_tables` converts the initial claim once; `into_transcript` is applied once to each round message and once to `final_eval`. No value crosses twice. (This basis change is *not* involutive, so a double application would not hide — but it does not happen.)
- **No transposition.** There is no matrix here: the change of basis is a `u128 -> u128` bijection, and `coord_u`'s index set never moves.
- **One-sidedness.** `git diff` touches no verifier function. `verify_readings` recomputes `A`, the rows and the equality element entirely in `EF`.

Beyond the PR's own test I confirmed, by temporary instrumentation (reverted):

- All three cases in the new test run 5, 3 and 5 rounds — so the Boolean-prefix path is genuinely exercised, not degenerate.
- `prove::<Ghash128>` proofs **verify absolutely** against the true claim in all three shapes, not merely equal the tower proofs.
- The `rounds == 0` edge case (whole kept prefix Boolean — reachable, since `prefix_limit()` is uncapped for a plain reduction) produces a proof that verifies in the polynomial basis.

---

## On the closed #2247

Nothing to carry: #2247 was closed by its own author purely to re-anchor the branch onto `perf/encode-merkle-fusion`, and received zero reviews. I diffed the two bodies — they are byte-identical except `Stacked on #2246.` -> `Stacked on #2252.` and one trailing newline. So the richer-looking #2247 describes exactly the behaviour #2253 implements; nothing described there is missing here. The `1196/-223` vs `268/-84` gap is entirely the intervening stack (#2249, #2252) having already landed `ReprSumcheckProver`, `FromTable` and `IntoTranscriptField`.

---

## Test coverage

| Gap | Concrete test |
| --- | --- |
| The cross-field subset-sum lemma — the exact claim `CoordinateSums`' new doc makes — is never tested directly. `basis.rs`'s `coordinate_sums_match_the_walk` only exercises `A == EF`. | In `basis.rs::tests`: random `w: [BinaryField128; 128]`, assert `BinaryField128::from(CoordinateSums::<BinaryField128, Ghash128>::new(&w.map(Ghash128::from)).sum(v)) == CoordinateSums::<_, BinaryField128>::new(&w).sum(v)` over 200 random `v`. I ran this; it passes. A `proptest` over `raw: u128` for `v` is strictly better and just as cheap — a failure would localise the bug to `phi` failing additivity, independently of the whole reduction. |
| `rounds == 0` is reachable (a fully Boolean kept prefix, since `prefix_limit()` is uncapped for a plain reduction) and is not covered in either basis. | Set every `high()` coordinate to `ONE`, assert `proof.sumcheck.num_rounds() == 0` and that both `prove::<BinaryField128>` and `prove::<Ghash128>` verify. I ran it; both pass. Worth pinning, because `from_repr_tables` -> `SumcheckProver::new` on a one-element table plus a zero-round `SumcheckShape` is an easy thing for a later refactor to break. |
| No absolute-correctness assertion on the polynomial-basis leg. | See the MINOR above — three lines. |
| No negative test that a claim reduced in one basis is *rejected* in the other. | Deliberately not applicable here, and I want to be explicit about why rather than recommend a test that cannot exist: the basis never reaches the transcript, so there is no "verify in the other basis" to reject. The correct negative statement is a *completeness* one — a non-multiplicative `R` yields a proof the honest verifier rejects — and it is not worth a test, since no such `R` is constructible from the crate's impls. |

---

## What I checked and found clean

- **Fiat-Shamir ordering** is untouched: elements are bound by `transcript.statement(...)` before `batch_drawn`, and the batch is built from the drawn challenges. The `R` parameter enters strictly after.
- **`unsafe`**: this PR adds none. It does newly route the ring-switch's witness table through the pre-existing `Ghash128::from_tower_vec` transmute, so I checked its invariants: both `Ghash128` (`binary-field/src/ghash.rs:89`) and the tower macro's fields (`binary-field/src/tower.rs:135`) are `#[repr(transparent)]`, and `BinaryField128::MASK == u128::MAX` so every source bit pattern is canonical — the safety comment's own precondition. `Vec::from_raw_parts` reuses the original `capacity`, and `size_of` is equal, so the over-allocation a rayon `collect` may leave is handled correctly.
- **Borrow lifetimes in `open_readings`**: `packing` borrows `prover_data`, which is later moved into `try_open_at_known`. NLL ends the borrow at the loop's last use; no restructuring or explicit `drop` was needed, and none was added.
- **New panic surface in `BooleanPcs::packing`**: none. `Table::poly(0)` goes through `dense_matrix()`, which panics on `TableStorage::Boolean` — but the pre-PR code called the identical `poly(0)`, and `BooleanPcs` commits the packed 128-bit table, not a `Gf2` bit-sliced one.
- **Transcript identity of `ReprSumcheckProver`**: I compared it against `SumcheckProver::compute_sumcheck_polynomials` with `constraint: None` (`strategy.rs:1524-1552`). Same `SumcheckShape::new(folding_factor, pow_bits, Basis::Evaluation)`, same `measure_round -> observe_and_sample -> reduce_claim -> hold` sequence, same outstanding-challenge-on-return convention. The dropped `None` argument is the only difference and it is a no-op.
- **The basis is in the type**, which is the structural defence: `Poly<Ghash128>` and `Poly<BinaryField128>` are distinct types, so a value cannot silently be used in the wrong basis. The single leak is `weights_over`'s loose `From<EF>` bound, flagged above.
- **Perf reasoning is sound.** Addition is XOR in both bases, so `weights_over`'s subset-sum accumulation costs the same in `R` as in `EF` — carrying the `d`-entry batching table over instead of the `2^n`-entry result is a real, free win, and `alpha * base[w-1]` becomes a bare `clmul`. `restricted_packing` correctly rides its existing `collect` into the blocked in-place `from_tower_vec` kernel rather than doing a per-element `From` during the copy; that is the right ordering, since the blocked kernel beats 16 table lookups per element.
- **CI**: all checks green at the head SHA, including `Formatting and Clippy`, the `pclmulqdq` and `avx512f+vpclmulqdq` configurations that actually select the fast `Ghash128` kernels, `Slow regression tests` and `binary-pcs 2^16 round trip`.

---

## Questions

- The "byte-identical proofs to #2246 at `-l 12` and `-l 14`" claim is the strongest correctness evidence in the PR and I could not reproduce it in a reasonable budget. Is that comparison scripted anywhere a reviewer can re-run? It is the assertion that covers `weights_over`'s `alpha` branch at production scale — the in-repo tests reach it only at 5 variables.
- `BitRingSwitchBatch::weights()` is `pub` and, after this PR, has no caller outside `reduction.rs`'s own tests. Is it public for a downstream consumer, or is it now effectively test scaffolding that could become `pub(crate)`?

---

**Bottom line:** merge. The math is right, the boundary is in the right place, the verifier is untouched, and I confirmed the two load-bearing lemmas and the absolute correctness of the polynomial-basis proofs by hand. Fixing the `prove` doc table (`reduction.rs:1133`) before merge is worth the sixty seconds; the blanket `FromTable` impl and the `verify` assertion in the new test are good follow-ups.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Corrected. I ran the grep you describe over `prove`, `weights_over` and `from_repr_tables`: the crossings are the restricted packing (`reduction.rs:397`), the batching table and `alpha` (`reduction.rs:830`, `:850`), the initial claim (`repr.rs:112`), the per-round challenge (`repr.rs:194`), and back out the round message (`repr.rs:189-193`) plus the surviving claim (`reduction.rs:1212`). Your rows are right on the first four; the surviving claim crosses back too, so the table now names it as well. The `never` row is the tensor elements, the equality table and the point, and a line records that the weight multilinear is accumulated in `R` rather than carried across, since that is why it is no longer in the first row. Commit `docs(sumcheck): state which values the bit ring-switch rounds carry across`.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Tightened to `R: IntoTranscriptField<EF> + Sync` (`reduction.rs:822`), landed on top of the blanket impl as you said it has to be. The bound is strictly stronger than the old one — `IntoTranscriptField: Field + FromTable`, and `FromTable: From<EF> + Send` — so nothing else moved. The doc now says the entries are sums and products of the images, so `R` has to carry `EF`'s arithmetic and not just its elements; the isomorphism wording itself stays on the trait where it belongs. Commit `refactor(sumcheck): bound the ring-switch weight table on the isomorphism`.

sumcheck/src/strategy/repr.rs
Nashtare: Applied as proposed: one blanket `impl<T: Field> FromTable<T> for T` returning the table (`repr.rs:250`), with the four reflexive impls deleted, including the two `cfg(test)` ones.

I checked the coherence argument rather than taking it: the only other impls are `FromTable<BinaryField64> for Poly64` and `FromTable<BinaryField128> for Ghash128`, whose `Self` differs from the parameter in both, so they stay disjoint from the blanket impl. `cargo check -p p3-sumcheck --all-targets --features parallel` is clean and `cargo test -p p3-sumcheck --lib` gives 593 passed, matching your run.

One thing worth stating explicitly, since the change touches a production impl: it cannot move the production path either way. `ChallengeField::SumcheckRepr` is `Ghash128` or `Poly64` (`fold.rs:304`, `:311`), never the challenge field itself, so `restricted_packing::<R>` never selects an identity impl in production and its single `from_tower_vec` crossing is untouched. What now exercises the identity impl is the tower leg of the cross-basis test, and it still replays the polynomial leg message for message.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Added (`reduction.rs:2701-2708`). Each case now pairs its reduction with the point it was built from, embeds the witness over the wide level once outside the loop, and replays the polynomial-basis proof against `embedding.eval_base(point)` on a fresh sponge. The successor case verifies through `verify` with no successor reading, which `check_structure` accepts.

I checked the assertion is not vacuous before keeping it: adding `ONE` to the claim fails case 0 at that line. Commit `test(sumcheck): verify the polynomial-basis ring-switch proof`.

sumcheck/src/ring_switch/bits/reduction.rs
Nashtare: Reproduced your measurement — case 1 has `num_variables() == 5` and `sumcheck.num_rounds() == 3`, so the guard could not fire for the reason it gives. It now reads the round count off the proof after both legs have run (`reduction.rs:2681`), with a line noting that a Boolean prefix drops rounds. Commit `test(sumcheck): guard the cross-basis cases on the round count`.

tcoratger (approved):
Re-reviewed at `7e24bc45` (delta `b91934ca..HEAD`). All four MINORs and the NIT are addressed; the approval stands.

- **Doc table — fixed**, and your addition is right: the surviving claim crosses back too (`reduction.rs:1212`), and noting that the weight multilinear is accumulated in `R` rather than carried is the correct reason it left the first row.
- **`weights_over` bound — fixed.** `R: IntoTranscriptField<EF> + Sync` is strictly stronger (`IntoTranscriptField: Field + FromTable`, `FromTable: From<EF> + Send`), and keeping the isomorphism wording on the trait is the right place for it.
- **Reflexive `FromTable` impls — fixed.** The blanket `impl<T: Field> FromTable<T> for T { fn from_table(t) -> t }` is zero-copy, and all four specific impls including the two `cfg(test)` ones are gone. Coherence holds for the reason you give: `Poly64 != BinaryField64` and `Ghash128 != BinaryField128`. Your point that production never selects the identity (`ChallengeField::SumcheckRepr` is `Ghash128` or `Poly64`) is correct, so this cannot move the prover either way.
- **Absolute check in the cross-basis test — fixed**, and the mutation check (claim + `ONE` fails case 0) is the right way to show it is not vacuous.
- **[NIT] round-count guard — fixed**, now reading `poly.sumcheck.num_rounds()` off the proof.

`cargo test -p p3-sumcheck --lib` — 595 passed, matching the count before the change; `cargo clippy -p p3-sumcheck -p p3-binary-pcs --all-targets --all-features` clean. No new finding in the delta.

PR #2255, opened 2026-09-20
perf(multi-stark): read a later zerocheck round's lane group in one load

Nashtare:
Stacked on #2250.

The rows a lane group of the representation field covers are consecutive rows of the same column,
so the group is one load of that column. Reading it lane by lane cost four address computations
and three inserts per group, and asked for the column's cache line with four partial loads instead
of one. `lane_rows` builds the group with a single `PackedValue::from_slice`, and the three
unconditional reads in `accumulate_lanes` go through it. The conditional next-row read keeps the
lane-by-lane path, since past the last residual row the repeat-last tail stands in.

### Measurements

`CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, medians with the base re-run interleaved against the
branch. `pwl` is `prove_with_lookup`.

| | `prove_with_lookup` | prove |
| --- | --- | --- |
| BLAKE3 `-l 18` | 1.49 -> 1.40 s (-6.0%) | 2.554 -> 2.464 s (-3.5%) |
| Keccak-f `-l 18` | 538 -> 527 ms (-2.0%) | 0.721 -> 0.708 s (-1.8%) |

Per-round `round_poly_repr` at BLAKE3 `-l 18`: 179 -> 139, 93.3 -> 68.1, 49.6 -> 38.3,
27.8 -> 22.5, 17.9 -> 13.2, 11.9 -> 9.91 ms, a total of roughly 400 -> 305 ms (**-24%**).
`round_poly_boundary`, `round_poly_sliced` and `fold_repr` are unmoved, as expected.

The first repr round reads 11536 x 16384 x 16 B = 3.02 GB of column data, so it goes from
16.9 GB/s to 21.7 GB/s. The sequential pole on the same array on the same box is `fold_repr` at
118 GB/s, so this pass is at 18% of the pole rather than 14%.

`perf` counters over the whole run say why: instructions fall 220.46 G -> 217.09 G and
memory-level parallelism (`l1d_pend_miss.pending / .pending_cycles`) rises **2.84 -> 3.33**, +17%.
The shorter loop fits more columns in the out-of-order window. Retired L2 misses are unchanged
(+4%, noise) — this does not avoid misses, it overlaps them.

### A software prefetch was tried and deliberately not landed

A `_mm_prefetch` lookahead over the same loop reaches -5.4% on `pwl`, slightly less than this
change, and **the two are not additive**: together they measure 1.396 s against this change's
1.394 s, a 0.15% difference against a +-1.3% base spread. They work by different mechanisms and
hit the same wall — the prefetch cuts retired L2 misses 23% but does not move MLP at all
(2.84 -> 2.82), while this change does not move misses but raises MLP 17%. The ceiling is MLP of
about 3.3 either way.

The distance sweep is flat across D in {2, 4, 8, 16, 32, 64} (all 1.41 +- 0.01 s) and dead at
D = 0, which is the control proving the effect is the prefetch and not the loop restructure.
Prefetching four lines per column instead of two is worse (-4.7% against -5.4%), so lane-group
reads are not straddling an extra line. Since the result is identical from two columns of
lookahead to sixty-four, no honest justification could be written for a `GATHER_LOOKAHEAD`
constant, and the change would cost an `unsafe` block and a per-architecture split for at most
0.3% on top of this one.

### Where the rest of the gap is

Every column iteration also issues two 64-byte write-allocate stores into `local_point` and
`local_diff`. At width 11536 each of those is 738 KB, so a lane group's scratch is 1.48 MB against
2 MB of L2 shared by two SMT siblings — never resident, always a read-for-ownership to memory. Per
column the core has two load misses and two store misses competing for the same twelve fill
buffers, which is a plausible account of the MLP ceiling that neither more lookahead nor fewer uops
moved.

The scratch, not the gather, is now the larger memory consumer in this pass: `walk_lane_nodes`
re-streams those 1.48 MB from DRAM at **every** interpolation node, so the node walk moves several
times the column data. That is where the remaining distance to 118 GB/s lives.

### Correctness

Prover-only and byte-identical: FNV-1a digests of the postcard proof match for both objectives at
`-l 12`, `-l 14` and `-l 16`, with identical proof sizes. `a_lane_group_read_from_a_column_is_the_one_its_lanes_give`
pins `lane_rows` against the lane-by-lane build at every offset of a three-group column, and the
`backends_agree_*` family compares whole serialized proofs between `ReprBackend` and
`GenericBackend`.

`cargo test --workspace --features parallel` passes with 0 failures; clippy (including for
`x86_64-unknown-linux-gnu`) and rustfmt are clean.

Stack frames are unchanged: the 1239 `bridge_producer_consumer::helper` instantiations have
identical sorted frame lists except two that shrink by ~80 B, the maximum helper frame stays at
10600 B, and `walk_lane_nodes` and `walk_row_nodes` are untouched.


tcoratger (approved):
Reviewed against the stack base (`perf/hash-leaf-geometry`), with #2249's diff read alongside for the lane-layout context.

I verified the index sets are bit-identical to the old ones, that the widened load is exactly the right width with no tail overrun, and that the soundness precondition it depends on (`align_of::<R::Packing>() <= align_of::<R>()`) holds repo-wide. Miri is clean on the lane path at `WIDTH = 4`, all 286 `p3-multi-stark` lib tests pass at `WIDTH` 1, 2 and 4, and objdump confirms the claimed codegen win is real and **larger than the PR body states**. Approving.

First, a disambiguation the title invites: "a later zerocheck round" is the `rounds/repr.rs` module's own terminology for rounds computed in the isomorphic representation field (`repr.rs:1`), **not** a lookahead into a future round's data. The change reads the current round's columns at the same indices as before. There is no reordering and no read of unwritten data.

## The access pattern

A repr round stores each column as `Poly<R>`, one `R` per residual row, dense and contiguous. `num_evals = 2 * half` is a power of two. `round_poly_repr_lanes` (`repr.rs:347`) asserts `groups * lanes == half` exactly, and dispatches `s = group * lanes` for `group in 0..groups`.

```
column: &[R], len = num_evals = 2h            R = Ghash128, 16 B each

 row    0          s   s+1 s+2 s+3        h        s+h            2h-1
        +----------+---+---+---+----------+--------+---+---+---+---+
        |<--------------- lower half ---->|<------- upper half ---->|

 one lane group, W = R::Packing::WIDTH (4 under AVX-512 = 64 B, 2 under AVX2 = 32 B)

            s     s+1    s+2    s+3
          +------+------+------+------+
          | l_0  | l_1  | l_2  | l_3  |   contiguous, stride 1 element
          +------+------+------+------+
```

Four reads per column per group. Three are now one load each:

```
  local_lo  = lane_rows(column, s)          -> column[s       .. s      +W]   WIDENED
  local_hi  = lane_rows(column, s + half)   -> column[s+h     .. s+h    +W]   WIDENED
  next_lo   = lane_rows(column, s + 1)      -> column[s+1     .. s+1    +W]   WIDENED
  next_hi   = lane_group(|lane| ...)        -> column[s+h+1   .. s+h+1  +W]   lane-by-lane
```

The tail, the only place an off-by-one could hide. Max `s = half - W`:

```
  index:            ...   2h-W   2h-W+1   2h-W+2   2h-1 | 2h   <- one past the end
  local_hi reads         [  .  ,   .    ,   .    ,  .  ]        exact fit, len = 2h
  next_lo  reads   ends at h+1 <= 2h                            slack
  next_hi  wants          [  .  ,   .    ,   .    ,  .  , T ]   T = next_tail, out of range
```

So `local_hi` on the last group lands **exactly** on the end of the buffer, and `next_hi` is the one read that genuinely runs one element past it. The PR widens the first three and leaves the fourth alone. That is the correct split — there is no off-by-one.

The bound requirements are unchanged by the patch:

| read | old max index | new slice end | requirement |
| --- | --- | --- | --- |
| `local_lo` | `s + W - 1` | `s + W` | `len >= s + W` |
| `local_hi` | `s + h + W - 1` | `s + h + W` | `len >= 2h` (tight) |
| `next_lo` | `s + W` | `s + 1 + W` | `len >= h + 1` |

`&column[a..a+W]` panics under exactly the same condition that `column[a+W-1]` did. No new panic surface, no new OOB.

---

## [MINOR] `next_hi` keeps the slow path for every group, though only the last group needs it — `multi-stark/src/rounds/repr.rs:494`

The PR body justifies leaving `next_hi` lane-by-lane because "past the last residual row, the repeat-last tail stands in". True, but quantify when: lane `l` reads `row = s + half + l + 1`, so the tail is taken iff `s + half + W > num_evals - 1`, i.e. `(g+1) * W > half - 1`. Since `groups * W == half` exactly, that holds **only for `g == groups - 1`**, and there only for `l == W - 1`. One lane of one group per round.

So `groups - 1` of `groups` next-row reads are fully inside the column and could take the same widening. The predicate depends only on `s`, `half`, `num_evals` and `W`, so it hoists out of the column loop entirely — no per-column branch.

```rust
// repr.rs:485, hoisted above `for run in &round.next_columns`
let next_hi_in_column = s + half + 1 + R::Packing::WIDTH <= num_evals;
...
let next_hi = if next_hi_in_column {
    lane_rows(column, s + half + 1)
} else {
    lane_group(|lane| { /* unchanged tail path */ })
};
```

Why it is worth doing: `next_columns` is every column any AIR of the stage reads on the next row, which for BLAKE3 is a large fraction of the 11536 columns. The PR measures `round_poly_repr` at 400 -> 305 ms by widening three of four reads; the fourth is the same order of work per affected column. Not a correctness issue and not a merge blocker, but it is the obvious remainder and the PR's own reasoning does not cover it.

**Verify:** I applied exactly the patch above in a worktree and ran `cargo test --release -p p3-multi-stark --lib` under `-C target-feature=+avx512f,+vpclmulqdq` and under `+avx2,+vpclmulqdq`: 286 passed, 0 failed in both, including the `backends_agree_*` differential family and `a_lane_group_round_is_the_round_the_rows_give`. Reverted afterwards.

---

## [MINOR] `lane_rows` is a third copy of an idiom that already exists twice — `multi-stark/src/rounds/repr.rs:53`

`multi-stark/src/rounds.rs:32` and `multi-stark/src/fractional_gkr/materialize.rs:18` already define an identical private `packed_column_at`:

```rust
|values| *F::Packing::from_slice(&values[row..row + F::Packing::WIDTH]),
```

`lane_rows` differs only in the field parameter (`R` vs `F`) and in wrapping the result in `PackedExt`. Not worth blocking on — the two existing copies are already duplicates of each other — but if a shared `p3-field` helper (`PackedValue::from_row(slice, row)`) ever lands, these three should collapse into it.

---

## [MINOR] `pub(super)` is wider than the uses — `multi-stark/src/rounds/repr.rs:53`

`lane_rows` is referenced only at `repr.rs:480`, `:481`, `:493` and `repr/tests.rs:74`. `repr::tests` is a child module of `repr`, so plain `fn` suffices. `lane_group` and `sum_lanes` are `pub(super)` because `rounds/sliced.rs:33` imports them; `lane_rows` has no such caller. Narrowing it keeps the module's export surface honest.

---

## [NIT]

- Doc sentence has a dangling comparative — `repr.rs:44`: "One lane group of the representation field, read from as many consecutive rows of a column." As many as *what*? The antecedent (`as it has lanes`) is missing. Suggest: "read from the `WIDTH` consecutive rows of a column it covers."
- The branch name `perf/zc-gather-prefetch` contains neither a gather nor a prefetch. The PR body explains this thoroughly and the title is accurate, so it is cosmetic — but a reader grepping the history for `_mm_prefetch` will be misled.

---

## What I checked and confirmed clean

**Memory safety of the widening load.** `PackedValue::from_slice` is a safe fn whose every impl does an unchecked `slice.as_ptr().cast()` (`field/src/packed/packed_traits.rs:248`, `field/src/array.rs:146`, `field/src/vectorized.rs:502`, `field/src/op_assign_macros.rs:328`). That cast is only sound if `align_of::<Packing>() <= align_of::<Value>()` — the classic trap, and the reason the sibling `pack_slice` carries `const { assert!(align_of::<Self>() <= align_of::<Self::Value>()) }`. `from_slice` itself carries no such assert, so `lane_rows` inherits the obligation at *arbitrary* offsets, which is a stronger use than most existing call sites (`uni-stark/src/prover.rs:821`, `batch-stark/src/prover.rs:964`). It holds:

- The only relevant `R` is `Ghash128`, whose `Packing` is `PackedGhash128`, declared `#[repr(transparent)] pub struct PackedGhash128([Ghash128; WIDTH])` (`binary-field/src/packed/x86_64.rs:272`). `Ghash128` is transparent over `u128`, so both sides are align 16. No over-alignment.
- `grep -rn "repr(align" --include=*.rs .` returns **nothing** across the whole workspace. Every SIMD packing in the repo (`monty-31`, `goldilocks`, `mersenne-31`, `binary-field`) is `repr(transparent)` over an array of the scalar, so the invariant is structural, not incidental. The `__m512i` only appears inside `to_vector()`, which transmutes a value already in a register, never a memory load.
- Off-target, `Packing = Ghash128` and `WIDTH = 1`, so `from_slice` is the trivial `&slice[0]`.

**Miri.** `cargo +nightly miri test -p p3-multi-stark --lib` under `-C target-feature=+avx512f,+vpclmulqdq`:
- `a_lane_group_read_from_a_column_is_the_one_its_lanes_give` — ok (0.97 s). This walks `lane_rows` at every offset `0..=2W` of a `3W`-element column, so it covers offsets that are *not* 64-byte aligned, exactly where a bad alignment assumption would fire.
- `a_lane_group_round_is_the_round_the_rows_give` — ok (5.98 s). This runs the real `accumulate_lanes` loop, including `local_hi` on the last group at the exact end of the buffer.

Miri under Stacked Borrows would flag both the OOB and the under-aligned deref. It found neither.

**Codegen — the claim is understated.** I compiled both forms standalone against the worktree's `p3-binary-field` at `opt-level=3`, `codegen-units=1`, `+avx512f,+vpclmulqdq`:

```
new_path (lane_rows):                      old_path (lane_group + index):
  cmp / ja      (overflow check)             4 x (lea, cmp, je)   -> 4 bounds branches
  cmp / ja      (bounds check)               8 x mov  (%rsi,..)   -> 8 scalar 64-bit loads
  shl $0x4                                   8 x mov  ..,(%rdi)   -> 8 scalar stores
  vmovups (%rsi,%rcx,1),%zmm0                push %rbx            -> extra callee-save
  vmovups %zmm0,(%rdi)
```

The old path was never four 128-bit loads — LLVM lowered `from_fn` to **eight GPR loads** plus four independent panic edges, and spilled `%rbx`. The new path is one `vmovups zmm` and two branches. That fully accounts for the reported instruction-count drop (220.46 G -> 217.09 G) and for the MLP rise (2.84 -> 3.33): eight dependent address computations per group per column were occupying the out-of-order window that now holds more concurrent column misses. The `vmovups` (unaligned form) is itself the proof that LLVM is using align-16 semantics, i.e. it never assumes 64-byte alignment.

**Geometry probes.** I instrumented `round_poly_repr_lanes` (then reverted) to confirm the boundary sizes are actually executed, not merely reachable:
- `a_lane_group_round_is_the_round_the_rows_give` runs at `lanes = 4, half = 16, width = 6`, i.e. `groups = 4`, so the last group reads `column[28..32]` with `len = 32` — the exact-fit case.
- The `half == lanes` single-group case (`half = 4, lanes = 4`, the tightest geometry where `local_hi` reads `column[4..8]` of an 8-element column and `next_lo` reaches `half + 1`) **is** reached, by several tests in the `backends_agree_*` family.
- `half < lanes` and `lanes == 1` are gated out at `repr.rs:316` before `lane_rows` is ever called, so `group_width - 1` and `1` cannot reach this code.

**Crypto soundness.** Purely prover-side. The index sets are identical, so `scratch.local_point` / `local_diff` / `next_point` / `next_diff` receive bit-identical values, `walk_lane_nodes` and `sum_lanes` are untouched, and the round polynomial fed to `finish_round` (`repr.rs:406`) is unchanged. No transcript absorption is reordered and no challenge derivation is touched. The `backends_agree_*` family, which compares whole serialized proofs between `ReprBackend<Gf4, PolyBasis>` and `GenericBackend`, passes at both `WIDTH = 2` and `WIDTH = 4` — a stronger check than the digest comparison cited in the PR body, since it pins the full transcript rather than the final bytes.

**Test runs.** `cargo test --release -p p3-multi-stark --lib`: 286 passed / 0 failed at default (`WIDTH = 1`), `+avx2,+vpclmulqdq` (`WIDTH = 2`) and `+avx512f,+vpclmulqdq` (`WIDTH = 4`). `cargo clippy -p p3-multi-stark --all-targets` under `+avx512f,+vpclmulqdq`: no warnings.

**The new test itself.** `repr/tests.rs:64` is a real test, not a tautology: `seed` is odd, so `seed.wrapping_mul(row + 1)` is injective mod `2^128` and the `3W` column values are distinct; it then sweeps **every** offset including the non-`W`-aligned ones, which is where a lane-order or alignment mistake would show. It does degenerate to a single trivial comparison when `WIDTH == 1`, which is the default build — see below.

---

## Test coverage

**Gap 1 — CI never executes any of this.** Every `vpclmulqdq` leg in `.github/workflows/ci.yml:119` and `:125` is `compile_only: true`, with the comment "Executing these paths needs real silicon" and a note that QEMU cannot decode the 256-bit carryless multiply. On the default legs `Ghash128::Packing = Ghash128`, so `WIDTH = 1`, `round_poly_repr` bails at `repr.rs:316` before `round_poly_repr_lanes`, and the new test collapses to one iteration comparing a 1-lane read against a 1-lane build. **The entire patch is a no-op under CI.** This is pre-existing and not the PR's fault, but it means the green CI on this PR is zero evidence here. I ran the real widths locally, above, which is the evidence that matters.

**Gap 2 — Miri does not cover this crate.** `ci.yml:327` runs Miri on exactly one test, in `p3-binary-field`. An OOB or misaligned read in `multi-stark` would not be caught. The two Miri runs I did above take 1 s and 6 s respectively, so adding a Miri leg for `p3-multi-stark --lib rounds::repr` under `+avx512f,+vpclmulqdq` is nearly free and would turn the alignment argument into a machine-checked one. Worth proposing as a separate PR, not as a condition on this one.

**Gap 3 — one geometry for the equivalence test.** `a_lane_group_round_is_the_round_the_rows_give` pins exactly `(half, lanes, width) = (16, 4, 6)`. The `backends_agree_*` family incidentally covers `half == lanes`, but nothing sweeps geometries deliberately. A cheap proptest:

```rust
proptest! {
    #[test]
    fn the_lane_group_round_matches_the_row_round_at_every_height(
        log_rows in 3usize..8, seed: u8,
    ) {
        let instances = [Instance::honest(FixtureAir::Pair, 1 << log_rows, seed)];
        with_state(&instances, no_lookups(), |state, _| {
            let mut a = state.clone().fold_into::<Ghash128>(first_challenge());
            let mut b = state.fold_into::<Ghash128>(first_challenge());
            let tau = a.tau.as_slice().to_vec();
            let eq = Poly::new_from_point(&tau[2..], Tower::ONE);
            let lifted = Poly::new(eq.as_slice().iter().copied().map(Ghash128::from).collect());
            prop_assert_eq!(a.round_poly_repr(&eq), b.round_poly_unpacked(&lifted));
            Ok(())
        })?;
    }
}
```

A failure would print the `log_rows` that broke, which localizes the tail index directly. `log_rows` starting at 3 keeps `half >= 4 = WIDTH` so the lane path is entered under AVX-512.

**Gap 4 — no test pins the panic contract.** `lane_rows` documents "Panics if the column holds fewer rows past `row` than the group has lanes", and nothing checks it. In release the panic is what stands between a caller mistake and an adjacent-heap read that silently corrupts a round polynomial rather than crashing. A four-line `#[should_panic]` on `lane_rows(&column[..W], 1)` makes that guarantee a tested one.

**Gap 5 — no in-repo benchmark isolates this pass.** `multi-stark/benches/zerocheck.rs` is BabyBear + `GenericBackend` only; it contains no `Ghash128`, no `ReprBackend`, no `PolyBasis`. The measurements in the PR body come from the example harness on a c7i.4xlarge and cannot be reproduced by a reviewer without that setup. Adding a `ReprBackend<Gf4, PolyBasis>` leg to the existing criterion bench would make this stack's remaining perf claims checkable in-tree. Again a follow-up, not a condition.

---

## Could not verify

- **The MLP account.** I have no c7i.4xlarge, so I could not reproduce `l1d_pend_miss.pending / .pending_cycles` at 2.84 -> 3.33. The mechanism the author proposes is consistent with the disassembly I did reproduce (eight dependent address computations per group removed from the window), so I have no reason to doubt it, but I am taking the counter numbers on trust.
- **Cache-line straddling.** The claim "lane-group reads are not straddling an extra line" is inferred from a prefetch experiment, not measured. Note `next_lo` is at `s + 1`, i.e. offset by one 16-byte element, so its 64-byte load straddles two lines *unconditionally* whenever the column base is 64-byte aligned. `local_lo` and `local_hi` are at multiples of `W` and do not. Whether a `Vec<Ghash128>` base is 64-byte aligned depends on the allocator — 16 is all `align_of` guarantees, though a 256 KB column will typically be mmap'd and page-aligned. If the widened `next_hi` above lands, it would be at `s + half + 1` and straddle for the same reason, which is worth measuring alongside it.
- **How much of BLAKE3's width `next_columns` covers.** This determines the payoff of the `next_hi` suggestion. `next_row_runs(&self.slots)` computes it per stage and I did not evaluate it for the real AIR.

---

**Bottom line:** merge. The transformation is observationally equivalent, the tail is exact rather than off-by-one, the `unsafe` beneath `from_slice` is discharged by a `repr(transparent)` invariant that holds across the entire workspace and is confirmed by Miri, and the claimed speedup is real and understated in the disassembly. The `next_hi` widening is the natural follow-up and I have verified it passes the suite, but it does not belong in this PR's scope.

multi-stark/src/rounds/repr.rs
```diff
@@ -475,7 +490,7 @@ where
                 .zip(&self.next_tail[run.clone()])
             {
                 let column = column.as_slice();
-                let next_lo = lane_group(|lane| column[s + lane + 1]);
+                let next_lo = lane_rows(column, s + 1);
                 let next_hi = lane_group(|lane| {
```
tcoratger: **[MINOR] `next_hi` keeps the slow path for every group, though only the last group needs it — `multi-stark/src/rounds/repr.rs:494`**

The PR body justifies leaving `next_hi` lane-by-lane because "past the last residual row, the repeat-last tail stands in". True, but quantify when: lane `l` reads `row = s + half + l + 1`, so the tail is taken iff `s + half + W > num_evals - 1`, i.e. `(g+1) * W > half - 1`. Since `groups * W == half` exactly, that holds **only for `g == groups - 1`**, and there only for `l == W - 1`. One lane of one group per round.

So `groups - 1` of `groups` next-row reads are fully inside the column and could take the same widening. The predicate depends only on `s`, `half`, `num_evals` and `W`, so it hoists out of the column loop entirely — no per-column branch.

```rust
// repr.rs:485, hoisted above `for run in &round.next_columns`
let next_hi_in_column = s + half + 1 + R::Packing::WIDTH <= num_evals;
...
let next_hi = if next_hi_in_column {
    lane_rows(column, s + half + 1)
} else {
    lane_group(|lane| { /* unchanged tail path */ })
};
```

Why it is worth doing: `next_columns` is every column any AIR of the stage reads on the next row, which for BLAKE3 is a large fraction of the 11536 columns. The PR measures `round_poly_repr` at 400 -> 305 ms by widening three of four reads; the fourth is the same order of work per affected column. Not a correctness issue and not a merge blocker, but it is the obvious remainder and the PR's own reasoning does not cover it.

**Verify:** I applied exactly the patch above in a worktree and ran `cargo test --release -p p3-multi-stark --lib` under `-C target-feature=+avx512f,+vpclmulqdq` and under `+avx2,+vpclmulqdq`: 286 passed, 0 failed in both, including the `backends_agree_*` differential family and `a_lane_group_round_is_the_round_the_rows_give`. Reverted afterwards.

multi-stark/src/rounds/repr.rs
```diff
@@ -41,6 +41,21 @@ pub(super) fn lane_group<F, R: Field>(value: impl FnMut(usize) -> R) -> PackedRe
...
+/// One lane group of the representation field, read from as many consecutive rows of a column.
+///
+/// The rows a lane group covers sit next to each other, so the whole group is one load of the
+/// column rather than one load per lane.
+///
+/// # Panics
+///
+/// Panics if the column holds fewer rows past `row` than the group has lanes.
+#[inline]
+pub(super) fn lane_rows<F, R: Field>(column: &[R], row: usize) -> PackedRepr<F, R> {
```
tcoratger: **[MINOR] `lane_rows` is a third copy of an idiom that already exists twice — `multi-stark/src/rounds/repr.rs:53`**

`multi-stark/src/rounds.rs:32` and `multi-stark/src/fractional_gkr/materialize.rs:18` already define an identical private `packed_column_at`:

```rust
|values| *F::Packing::from_slice(&values[row..row + F::Packing::WIDTH]),
```

`lane_rows` differs only in the field parameter (`R` vs `F`) and in wrapping the result in `PackedExt`. Not worth blocking on — the two existing copies are already duplicates of each other — but if a shared `p3-field` helper (`PackedValue::from_row(slice, row)`) ever lands, these three should collapse into it.

tcoratger: **[MINOR] `pub(super)` is wider than the uses — `multi-stark/src/rounds/repr.rs:53`**

`lane_rows` is referenced only at `repr.rs:480`, `:481`, `:493` and `repr/tests.rs:74`. `repr::tests` is a child module of `repr`, so plain `fn` suffices. `lane_group` and `sum_lanes` are `pub(super)` because `rounds/sliced.rs:33` imports them; `lane_rows` has no such caller. Narrowing it keeps the module's export surface honest.

tcoratger (commented):
Re-posting with findings anchored to their lines; supersedes my earlier body-only review. The APPROVE from my earlier review stands; this pass only re-homes the findings onto the lines they concern.

---

Reviewed against the stack base (`perf/hash-leaf-geometry`), with #2249's diff read alongside for the lane-layout context.

I verified the index sets are bit-identical to the old ones, that the widened load is exactly the right width with no tail overrun, and that the soundness precondition it depends on (`align_of::<R::Packing>() <= align_of::<R>()`) holds repo-wide. Miri is clean on the lane path at `WIDTH = 4`, all 286 `p3-multi-stark` lib tests pass at `WIDTH` 1, 2 and 4, and objdump confirms the claimed codegen win is real and **larger than the PR body states**. Approving.

First, a disambiguation the title invites: "a later zerocheck round" is the `rounds/repr.rs` module's own terminology for rounds computed in the isomorphic representation field (`repr.rs:1`), **not** a lookahead into a future round's data. The change reads the current round's columns at the same indices as before. There is no reordering and no read of unwritten data.

---

## The access pattern

A repr round stores each column as `Poly<R>`, one `R` per residual row, dense and contiguous. `num_evals = 2 * half` is a power of two. `round_poly_repr_lanes` (`repr.rs:347`) asserts `groups * lanes == half` exactly, and dispatches `s = group * lanes` for `group in 0..groups`.

```
column: &[R], len = num_evals = 2h            R = Ghash128, 16 B each

 row    0          s   s+1 s+2 s+3        h        s+h            2h-1
        +----------+---+---+---+----------+--------+---+---+---+---+
        |<--------------- lower half ---->|<------- upper half ---->|

 one lane group, W = R::Packing::WIDTH (4 under AVX-512 = 64 B, 2 under AVX2 = 32 B)

            s     s+1    s+2    s+3
          +------+------+------+------+
          | l_0  | l_1  | l_2  | l_3  |   contiguous, stride 1 element
          +------+------+------+------+
```

Four reads per column per group. Three are now one load each:

```
  local_lo  = lane_rows(column, s)          -> column[s       .. s      +W]   WIDENED
  local_hi  = lane_rows(column, s + half)   -> column[s+h     .. s+h    +W]   WIDENED
  next_lo   = lane_rows(column, s + 1)      -> column[s+1     .. s+1    +W]   WIDENED
  next_hi   = lane_group(|lane| ...)        -> column[s+h+1   .. s+h+1  +W]   lane-by-lane
```

The tail, the only place an off-by-one could hide. Max `s = half - W`:

```
  index:            ...   2h-W   2h-W+1   2h-W+2   2h-1 | 2h   <- one past the end
  local_hi reads         [  .  ,   .    ,   .    ,  .  ]        exact fit, len = 2h
  next_lo  reads   ends at h+1 <= 2h                            slack
  next_hi  wants          [  .  ,   .    ,   .    ,  .  , T ]   T = next_tail, out of range
```

So `local_hi` on the last group lands **exactly** on the end of the buffer, and `next_hi` is the one read that genuinely runs one element past it. The PR widens the first three and leaves the fourth alone. That is the correct split — there is no off-by-one.

The bound requirements are unchanged by the patch:

| read | old max index | new slice end | requirement |
| --- | --- | --- | --- |
| `local_lo` | `s + W - 1` | `s + W` | `len >= s + W` |
| `local_hi` | `s + h + W - 1` | `s + h + W` | `len >= 2h` (tight) |
| `next_lo` | `s + W` | `s + 1 + W` | `len >= h + 1` |

`&column[a..a+W]` panics under exactly the same condition that `column[a+W-1]` did. No new panic surface, no new OOB.

---

### [NIT]

- Doc sentence has a dangling comparative — `repr.rs:44`: "One lane group of the representation field, read from as many consecutive rows of a column." As many as *what*? The antecedent (`as it has lanes`) is missing. Suggest: "read from the `WIDTH` consecutive rows of a column it covers."
- The branch name `perf/zc-gather-prefetch` contains neither a gather nor a prefetch. The PR body explains this thoroughly and the title is accurate, so it is cosmetic — but a reader grepping the history for `_mm_prefetch` will be misled.

---

## What I checked and confirmed clean

**Memory safety of the widening load.** `PackedValue::from_slice` is a safe fn whose every impl does an unchecked `slice.as_ptr().cast()` (`field/src/packed/packed_traits.rs:248`, `field/src/array.rs:146`, `field/src/vectorized.rs:502`, `field/src/op_assign_macros.rs:328`). That cast is only sound if `align_of::<Packing>() <= align_of::<Value>()` — the classic trap, and the reason the sibling `pack_slice` carries `const { assert!(align_of::<Self>() <= align_of::<Self::Value>()) }`. `from_slice` itself carries no such assert, so `lane_rows` inherits the obligation at *arbitrary* offsets, which is a stronger use than most existing call sites (`uni-stark/src/prover.rs:821`, `batch-stark/src/prover.rs:964`). It holds:

- The only relevant `R` is `Ghash128`, whose `Packing` is `PackedGhash128`, declared `#[repr(transparent)] pub struct PackedGhash128([Ghash128; WIDTH])` (`binary-field/src/packed/x86_64.rs:272`). `Ghash128` is transparent over `u128`, so both sides are align 16. No over-alignment.
- `grep -rn "repr(align" --include=*.rs .` returns **nothing** across the whole workspace. Every SIMD packing in the repo (`monty-31`, `goldilocks`, `mersenne-31`, `binary-field`) is `repr(transparent)` over an array of the scalar, so the invariant is structural, not incidental. The `__m512i` only appears inside `to_vector()`, which transmutes a value already in a register, never a memory load.
- Off-target, `Packing = Ghash128` and `WIDTH = 1`, so `from_slice` is the trivial `&slice[0]`.

**Miri.** `cargo +nightly miri test -p p3-multi-stark --lib` under `-C target-feature=+avx512f,+vpclmulqdq`:
- `a_lane_group_read_from_a_column_is_the_one_its_lanes_give` — ok (0.97 s). This walks `lane_rows` at every offset `0..=2W` of a `3W`-element column, so it covers offsets that are *not* 64-byte aligned, exactly where a bad alignment assumption would fire.
- `a_lane_group_round_is_the_round_the_rows_give` — ok (5.98 s). This runs the real `accumulate_lanes` loop, including `local_hi` on the last group at the exact end of the buffer.

Miri under Stacked Borrows would flag both the OOB and the under-aligned deref. It found neither.

**Codegen — the claim is understated.** I compiled both forms standalone against the worktree's `p3-binary-field` at `opt-level=3`, `codegen-units=1`, `+avx512f,+vpclmulqdq`:

```
new_path (lane_rows):                      old_path (lane_group + index):
  cmp / ja      (overflow check)             4 x (lea, cmp, je)   -> 4 bounds branches
  cmp / ja      (bounds check)               8 x mov  (%rsi,..)   -> 8 scalar 64-bit loads
  shl $0x4                                   8 x mov  ..,(%rdi)   -> 8 scalar stores
  vmovups (%rsi,%rcx,1),%zmm0                push %rbx            -> extra callee-save
  vmovups %zmm0,(%rdi)
```

The old path was never four 128-bit loads — LLVM lowered `from_fn` to **eight GPR loads** plus four independent panic edges, and spilled `%rbx`. The new path is one `vmovups zmm` and two branches. That fully accounts for the reported instruction-count drop (220.46 G -> 217.09 G) and for the MLP rise (2.84 -> 3.33): eight dependent address computations per group per column were occupying the out-of-order window that now holds more concurrent column misses. The `vmovups` (unaligned form) is itself the proof that LLVM is using align-16 semantics, i.e. it never assumes 64-byte alignment.

**Geometry probes.** I instrumented `round_poly_repr_lanes` (then reverted) to confirm the boundary sizes are actually executed, not merely reachable:
- `a_lane_group_round_is_the_round_the_rows_give` runs at `lanes = 4, half = 16, width = 6`, i.e. `groups = 4`, so the last group reads `column[28..32]` with `len = 32` — the exact-fit case.
- The `half == lanes` single-group case (`half = 4, lanes = 4`, the tightest geometry where `local_hi` reads `column[4..8]` of an 8-element column and `next_lo` reaches `half + 1`) **is** reached, by several tests in the `backends_agree_*` family.
- `half < lanes` and `lanes == 1` are gated out at `repr.rs:316` before `lane_rows` is ever called, so `group_width - 1` and `1` cannot reach this code.

**Crypto soundness.** Purely prover-side. The index sets are identical, so `scratch.local_point` / `local_diff` / `next_point` / `next_diff` receive bit-identical values, `walk_lane_nodes` and `sum_lanes` are untouched, and the round polynomial fed to `finish_round` (`repr.rs:406`) is unchanged. No transcript absorption is reordered and no challenge derivation is touched. The `backends_agree_*` family, which compares whole serialized proofs between `ReprBackend<Gf4, PolyBasis>` and `GenericBackend`, passes at both `WIDTH = 2` and `WIDTH = 4` — a stronger check than the digest comparison cited in the PR body, since it pins the full transcript rather than the final bytes.

**Test runs.** `cargo test --release -p p3-multi-stark --lib`: 286 passed / 0 failed at default (`WIDTH = 1`), `+avx2,+vpclmulqdq` (`WIDTH = 2`) and `+avx512f,+vpclmulqdq` (`WIDTH = 4`). `cargo clippy -p p3-multi-stark --all-targets` under `+avx512f,+vpclmulqdq`: no warnings.

**The new test itself.** `repr/tests.rs:64` is a real test, not a tautology: `seed` is odd, so `seed.wrapping_mul(row + 1)` is injective mod `2^128` and the `3W` column values are distinct; it then sweeps **every** offset including the non-`W`-aligned ones, which is where a lane-order or alignment mistake would show. It does degenerate to a single trivial comparison when `WIDTH == 1`, which is the default build — see below.

---

## Test coverage

**Gap 1 — CI never executes any of this.** Every `vpclmulqdq` leg in `.github/workflows/ci.yml:119` and `:125` is `compile_only: true`, with the comment "Executing these paths needs real silicon" and a note that QEMU cannot decode the 256-bit carryless multiply. On the default legs `Ghash128::Packing = Ghash128`, so `WIDTH = 1`, `round_poly_repr` bails at `repr.rs:316` before `round_poly_repr_lanes`, and the new test collapses to one iteration comparing a 1-lane read against a 1-lane build. **The entire patch is a no-op under CI.** This is pre-existing and not the PR's fault, but it means the green CI on this PR is zero evidence here. I ran the real widths locally, above, which is the evidence that matters.

**Gap 2 — Miri does not cover this crate.** `ci.yml:327` runs Miri on exactly one test, in `p3-binary-field`. An OOB or misaligned read in `multi-stark` would not be caught. The two Miri runs I did above take 1 s and 6 s respectively, so adding a Miri leg for `p3-multi-stark --lib rounds::repr` under `+avx512f,+vpclmulqdq` is nearly free and would turn the alignment argument into a machine-checked one. Worth proposing as a separate PR, not as a condition on this one.

**Gap 3 — one geometry for the equivalence test.** `a_lane_group_round_is_the_round_the_rows_give` pins exactly `(half, lanes, width) = (16, 4, 6)`. The `backends_agree_*` family incidentally covers `half == lanes`, but nothing sweeps geometries deliberately. A cheap proptest:

```rust
proptest! {
    #[test]
    fn the_lane_group_round_matches_the_row_round_at_every_height(
        log_rows in 3usize..8, seed: u8,
    ) {
        let instances = [Instance::honest(FixtureAir::Pair, 1 << log_rows, seed)];
        with_state(&instances, no_lookups(), |state, _| {
            let mut a = state.clone().fold_into::<Ghash128>(first_challenge());
            let mut b = state.fold_into::<Ghash128>(first_challenge());
            let tau = a.tau.as_slice().to_vec();
            let eq = Poly::new_from_point(&tau[2..], Tower::ONE);
            let lifted = Poly::new(eq.as_slice().iter().copied().map(Ghash128::from).collect());
            prop_assert_eq!(a.round_poly_repr(&eq), b.round_poly_unpacked(&lifted));
            Ok(())
        })?;
    }
}
```

A failure would print the `log_rows` that broke, which localizes the tail index directly. `log_rows` starting at 3 keeps `half >= 4 = WIDTH` so the lane path is entered under AVX-512.

**Gap 4 — no test pins the panic contract.** `lane_rows` documents "Panics if the column holds fewer rows past `row` than the group has lanes", and nothing checks it. In release the panic is what stands between a caller mistake and an adjacent-heap read that silently corrupts a round polynomial rather than crashing. A four-line `#[should_panic]` on `lane_rows(&column[..W], 1)` makes that guarantee a tested one.

**Gap 5 — no in-repo benchmark isolates this pass.** `multi-stark/benches/zerocheck.rs` is BabyBear + `GenericBackend` only; it contains no `Ghash128`, no `ReprBackend`, no `PolyBasis`. The measurements in the PR body come from the example harness on a c7i.4xlarge and cannot be reproduced by a reviewer without that setup. Adding a `ReprBackend<Gf4, PolyBasis>` leg to the existing criterion bench would make this stack's remaining perf claims checkable in-tree. Again a follow-up, not a condition.

---

## Could not verify

- **The MLP account.** I have no c7i.4xlarge, so I could not reproduce `l1d_pend_miss.pending / .pending_cycles` at 2.84 -> 3.33. The mechanism the author proposes is consistent with the disassembly I did reproduce (eight dependent address computations per group removed from the window), so I have no reason to doubt it, but I am taking the counter numbers on trust.
- **Cache-line straddling.** The claim "lane-group reads are not straddling an extra line" is inferred from a prefetch experiment, not measured. Note `next_lo` is at `s + 1`, i.e. offset by one 16-byte element, so its 64-byte load straddles two lines *unconditionally* whenever the column base is 64-byte aligned. `local_lo` and `local_hi` are at multiples of `W` and do not. Whether a `Vec<Ghash128>` base is 64-byte aligned depends on the allocator — 16 is all `align_of` guarantees, though a 256 KB column will typically be mmap'd and page-aligned. If the widened `next_hi` above lands, it would be at `s + half + 1` and straddle for the same reason, which is worth measuring alongside it.
- **How much of BLAKE3's width `next_columns` covers.** This determines the payoff of the `next_hi` suggestion. `next_row_runs(&self.slots)` computes it per stage and I did not evaluate it for the real AIR.

---

**Bottom line:** merge. The transformation is observationally equivalent, the tail is exact rather than off-by-one, the `unsafe` beneath `from_slice` is discharged by a `repr(transparent)` invariant that holds across the entire workspace and is confirmed by Miri, and the claimed speedup is real and understated in the disassembly. The `next_hi` widening is the natural follow-up and I have verified it passes the suite, but it does not belong in this PR's scope.

multi-stark/src/rounds/repr.rs
Nashtare: Verified the arithmetic against the code before implementing it, and it holds. `round_poly_repr_lanes` asserts `groups * lanes == half` (`repr.rs:359-363`) and `num_evals` is `1 << num_variables`, so `num_evals == 2 * half` exactly. Lane `l` of group `g` reads `s + half + l + 1` with `s = g * lanes`, and `max(s + l) = half - 1`, so the tail is reached only at `g == groups - 1, l == W - 1`. Your predicate is not just sufficient but exact: `s + half + 1 + W <= num_evals` holds iff every lane of the group is in the column.

Implemented in c74b9742, with the predicate named `next_rows_in_column` so its arithmetic is testable (6b2e0065) — the lane packing only exists on x86-64 with `vpclmulqdq` + `avx2`/`avx512f`, so on every other target CI compiles the widened branch but never runs it. `a_lane_group_leaves_the_column_only_in_the_last_lane_of_the_last_group` pins the predicate against both rules on any target, and `a_lane_group_round_over_successor_columns_is_the_round_the_rows_give` runs the `Gate` fixture (main + preprocessed successor columns) through the lane kernel and the row kernel.

One correction to the sizing, though: **BLAKE3 reads no successor row at all.** `Blake3BinaryAir::eval` only takes `main.current_slice()`, so `next_row_runs` is empty for it and this loop never executes — the PR's 400 -> 305 ms came from widening `local_lo`/`local_hi`, the only two reads BLAKE3 has. Keccak is the objective that exercises it, and there `next_columns` is a single run: `KeccakBinaryAir` reads `next.round_flags`, 25 of the 1625 columns.

So the measurement is a wash, as 25/1625 predicts. c7i.4xlarge (Xeon 8488C), `-C target-cpu=native`, `codegen-units=1`, `--security-bits 95 --log-inv-rate 1 --folding 4 --merkle-arity 4`, both sides at 4f8819f4a, medians of 3:

| objective | span | base | with the patch |
| --- | --- | ---: | ---: |
| keccak-f `-l 18` | `round_poly_repr` (sum) | 103.2 ms | 105.6 ms |
| keccak-f `-l 18` | prove | 659 ms | 658 ms |
| blake-3 `-l 18` | `round_poly_repr` (sum) | 321.2 ms | 315.0 ms |
| blake-3 `-l 18` | prove | 2.187 s | 2.189 s |

Both `round_poly_repr` deltas are inside the run-to-run spread (Keccak base 101.3/103.2/107.1 vs 105.1/105.6/107.4; BLAKE3 base 316.4/321.2/332.4 vs 310.4/315.0/326.5), and the BLAKE3 one cannot be the patch since the branch is dead there. Landing it anyway: it is strictly less work and drops a per-lane bounds test from the inner loop.

Proof sizes were identical on both sides in all 12 runs (661898 / 993065 bytes) and every proof verified. I could not get the proof-byte digest comparison or the `+avx512f`/`+avx2` test runs: the spot host was reclaimed mid-run and I could not replace it, so for the lane path itself I am relying on your two runs plus the two new tests.

multi-stark/src/rounds/repr.rs
Nashtare: Not doing this one. The two `packed_column_at` copies are duplicates of each other, but what they share is the `ColumnView` dispatch — `column.as_dense().map_or_else(|| column.packed_at(row), ...)` — and the `from_slice` call is only its dense arm. `lane_rows` takes a dense `&[R]` and has no dispatch to share, so a `PackedValue::from_row(slice, row)` helper would collapse one line out of each of the three and leave the actual duplication (the `ColumnView` fallback, in `rounds.rs:32` and `fractional_gkr/materialize.rs:18`) untouched. `F::Packing::from_slice` is already the named idiom for that line.

Deduplicating the two `packed_column_at` copies is worth doing on its own, but it is a different change from this PR and does not involve `lane_rows`.

Nashtare: Correct — `lane_rows` has no caller outside `repr` and its `tests` child. Narrowed to a plain `fn` in 416f8ff8; `lane_group` and `sum_lanes` keep `pub(super)` for `rounds/sliced.rs:33`.

tcoratger (approved):
Re-reviewed at `80fd960a` (delta `523afe2a..HEAD`). All three MINORs handled; the approval stands.

- **`next_hi` slow path — implemented.** `next_rows_in_column(s, half, num_evals, W) = s + half + 1 + W <= num_evals` is hoisted above the column loop, and the widened read's bounds hold: `lane_rows(column, s + half + 1)` needs `s + half + 1 + W <= column.len()`, which the predicate gives, and `local_hi`'s worst case `s + half + W = num_evals` is the same bound the lane-by-lane path already required. Naming the predicate was the better call — `a_lane_group_leaves_the_column_only_in_the_last_lane_of_the_last_group` pins it on any target against both rules, and I re-derived the second: `(group+1)*lanes + half < 2*half` iff `group + 1 < groups`, which is what the test asserts. `a_lane_group_round_over_successor_columns_is_the_round_the_rows_give` closes the gap that the `Pair` fixture reads no successor row.
- **Your sizing correction is right**, and I should have caught it: `Blake3BinaryAir::eval` only takes `main.current_slice()`, so `next_row_runs` is empty and this loop never runs for BLAKE3. Keccak's 25-of-1625 `round_flags` predicts the wash your medians show. Landing it for the bounds test alone is the right call.
- **[MINOR] `lane_rows` duplication — correctly declined.** I checked: both `packed_column_at` copies take a `ColumnView` and share the `as_dense().map_or_else(...)` dispatch, of which `from_slice` is only the dense arm; `lane_rows` takes `&[R]` and wraps in `PackedExt`, so a `from_row` helper would collapse one line and leave the real duplication in place. Agreed it is a separate change.
- **[MINOR] `pub(super)` — narrowed** to a plain `fn`, still visible to the `tests` child.

`cargo test --release -p p3-multi-stark --lib` — 288 passed under default, `+avx2,+vpclmulqdq` and `+avx512f,+vpclmulqdq`; clippy and fmt clean. No new finding in the delta.

PR #2256, opened 2026-09-20
fix(uni-stark,batch-stark): reject a present-but-empty preprocessed_next opening

kriss39:
## Summary

For an AIR that does not read the next preprocessed row, the prover emits `opened_values.preprocessed_next = None` and the opening argument does not include that field. The verifiers in `p3-uni-stark` and `p3-batch-stark` only compared its *length* against the expected `0`, so a proof carrying `Some(vec![])` passed the shape check, reached the constraint evaluation with a width-0 "next" row paired with a width-`preprocessed_width` local row, and panicked in `VerticalPair::new` (`matrix/src/stack.rs`) instead of returning a `VerificationError`.

Since the field is not bound by the PCS opening, this needs nothing more than flipping one `Option` on an otherwise valid proof.

## Changes

- `uni-stark/src/verifier.rs`: after the existing width check, reject `preprocessed_next.is_some()` when the AIR does not declare next-row preprocessed columns, mirroring the `trace_next` handling a few lines below.
- `batch-stark/src/verifier/mod.rs`: same check per instance (the `pre_w > 0`, no-next-row branch).
- `uni-stark/src/error.rs`: new `InvalidProofShapeError::UnexpectedPreprocessedNext { air: Option<usize> }`.

## Tests

- `uni-stark/tests/preprocessed_no_next_row.rs`: a one-column AIR with a preprocessed column read on the current row only; the honest proof carries `None`, and the tampered proof is rejected with `UnexpectedPreprocessedNext { air: None }`.
- `batch-stark/tests/simple.rs::test_preprocessed_rejects_present_preprocessed_next`: same with the existing `PreprocessedMulAir` fixture, rejected with `UnexpectedPreprocessedNext { air: Some(0) }`.

Both tests panic on `main` at `matrix/src/stack.rs:68` (`assertion left == right failed, left: 1, right: 0`) and pass with this change. `cargo nextest run -p p3-uni-stark -p p3-batch-stark` (204 tests), clippy and rustfmt are clean.


uni-stark/src/verifier.rs
```diff
@@ -255,6 +256,13 @@ where
         }
         .into());
     }
+    // When the AIR doesn't read the next preprocessed row, the honest opening is `None`. A
+    // present-but-empty vector has the expected length but not the expected shape: it is not
+    // bound by the opening argument and would later be paired with the width-`preprocessed_width`
+    // local row, so reject it here like `trace_next`.
+    if !preprocessed_next_used && opened_values.preprocessed_next.is_some() {
```
tcoratger, line 289: **[NIT] Placement and error choice are right; the trace analogue is inconsistent**

Verified: this runs inside `process_preprocessed_trace`, called at `verifier.rs:465` — before `config.initialise_challenger()` at line 583. So a malformed proof is rejected without the sponge ever absorbing attacker-controlled data, which is the property that matters. Good.

One inconsistency worth noting: the comment says "reject it here like `trace_next`", but uni-stark does *not* report the trace case with `UnexpectedTraceNext`. It folds it into `valid_shape` at `verifier.rs:552` and returns the generic `OpenedValuesDimensionMismatch`, which tells the caller nothing about which field was wrong. batch-stark does use `UnexpectedTraceNext { air: i }`. Since you are adding a precise variant for preprocessed here, it would be a small win to also give uni-stark's `trace_next` case its own variant so the two crates report the same two conditions the same way.

**Fix:** optional — split `trace_next_ok` out of `valid_shape` and return a dedicated error.
**Verify:** a tamper test setting `trace_next = Some(vec![])` on an AIR with `main_next_row_columns()` empty currently yields `OpenedValuesDimensionMismatch`.

batch-stark/src/verifier/mod.rs
```diff
@@ -556,8 +556,18 @@ where
...
-        } else if pre_local_len != pre_w || pre_next_len != 0 {
-            return Err(InvalidProofShapeError::PreprocessedWidthMismatch { air: i }.into());
+        } else {
+            if pre_local_len != pre_w {
+                return Err(InvalidProofShapeError::PreprocessedWidthMismatch { air: i }.into());
+            }
+            // The honest opening is `None` here; a present-but-empty vector passes the length
+            // check but is later paired with the width-`pre_w` local row, so reject it like
+            // `trace_next` above.
+            if inst_base_opened_vals.preprocessed_next.is_some() {
```
tcoratger, line 577: **[MINOR] This half of the fix is the more valuable one — say so**

In uni-stark the panic happened at `verifier.rs:684` (post-opening constraint check). In batch-stark the corresponding `VerticalPair::new` is at `batch-stark/src/verifier/data.rs:85`, which is reached only **after** `transcript.finish()` and after `pcs.verify_with_preprocessing` has succeeded (`mod.rs:669-682`). So on `main` a malicious prover gets the verifier to run the entire Fiat-Shamir replay and the full FRI check and *then* abort the process. That is a strictly worse denial-of-service surface than the uni-stark one, and moving the rejection into this shape loop (before the transcript is built at line ~610) is exactly right.

One residual, not a bug: the `pre_w == 0` branch at line 551-554 still accepts `preprocessed_local = Some(vec![])` and `preprocessed_next = Some(vec![])`, since it only compares lengths against `0`. I traced it — both `VerticalPair` sides end up width 0 and no preprocessed commitment is bound, so it is harmless. Tightening it to `.is_some()` there too would make the rule uniform: the honest prover emits `None` in every one of these positions.

uni-stark/src/error.rs
```diff
@@ -34,6 +34,12 @@ pub enum InvalidProofShapeError {
     /// Trace next values present when AIR doesn't use next row.
     #[error("air {air}: unexpected trace next values")]
     UnexpectedTraceNext { air: usize },
+    /// Preprocessed next values present when the AIR doesn't read the next preprocessed row.
+    #[error(
+        "{}unexpected preprocessed next values",
+        air.map_or_else(String::new, |air| format!("air {air}: "))
+    )]
+    UnexpectedPreprocessedNext { air: Option<usize> },
```
tcoratger, line 52: **[NIT] `Option<usize>` vs `usize`, and the enum is not `#[non_exhaustive]`**

Two small things:

1. The neighbouring `UnexpectedTraceNext { air: usize }` (line 36) carries a plain `usize` because only batch-stark raises it, while this one needs `Option<usize>` for uni-stark's `air: None`. That is the right call and matches `DegreeBitsTooSmall`, but it means two adjacent variants describing the same class of defect have different shapes. If you ever unify, `Option<usize>` everywhere is the one to keep.

2. `InvalidProofShapeError` has no `#[non_exhaustive]`, so adding a variant is a semver-breaking change for any downstream crate that matches exhaustively. At `0.7.0` that is presumably accepted practice here, but `#[non_exhaustive]` on an error enum a verifier returns would let future rejections be added freely — which is precisely the kind of change this PR is.

batch-stark/tests/simple.rs
```diff
@@ -1257,6 +1257,48 @@ fn test_preprocessed_tampered_fails() -> Result<(), Box<dyn std::error::Error>>
...
+    let err = verify_batch(&config, &airs, &proof, from_ref(&pis), common)
+        .expect_err("verification should reject a present preprocessed_next");
+    assert!(
+        matches!(
+            err,
+            BatchVerificationError::Verification(VerificationError::InvalidProofShape(
+                InvalidProofShapeError::UnexpectedPreprocessedNext { air: Some(0) },
+            ))
+        ),
+        "unexpected error: {err:?}"
```
tcoratger, line 1351: **[MINOR] Single instance leaves the `air: Some(i)` payload untested for `i > 0`**

The test asserts `UnexpectedPreprocessedNext { air: Some(0) }`, but with one instance in the batch, index `0` is also what a hardcoded constant or an off-by-one would produce. Building the batch with two instances and tampering instance `1` would actually pin that the reported index is the offending AIR.

Also worth adding next to it: the sibling tamper `preprocessed_next = Some(vec![Challenge::ZERO; pre_w])`, which is caught by the pre-existing length check (`PreprocessedWidthMismatch`). Asserting both keeps someone from later collapsing the length check and the presence check into one and losing a case.

**Verify:** `cargo test -p p3-batch-stark --test simple test_preprocessed`.

uni-stark/tests/preprocessed_no_next_row.rs
```diff
@@ -0,0 +1,140 @@
...
+    MyConfig::new(pcs, challenger)
+}
+
+fn prove_copy_air() -> (
+    MyConfig,
+    CopyAir,
+    p3_uni_stark::Proof<MyConfig>,
+    p3_uni_stark::PreprocessedVerifierKey<MyConfig>,
+) {
+    let num_rows = 1 << 3;
```
tcoratger, line 95: **[NIT] ~60 lines of config boilerplate duplicated from the other uni-stark test files**

`Val`/`Perm`/`MyHash`/`ValMmcs`/`Challenge`/`Pcs`/`MyConfig` plus `make_config()` are byte-for-byte the same stack that `uni-stark/tests/fib_air.rs:118-156` already sets up. Not worth blocking, but a shared `uni-stark/tests/common/mod.rs` would stop the next tamper test from being a third copy.

The test itself is exactly the right one: I deleted only the three added lines in `uni-stark/src/verifier.rs` and `test_preprocessed_no_next_row_rejects_present_preprocessed_next` panics at `matrix/src/stack.rs:68` with `left: 1, right: 0`, so it does prove the fix rather than just exercising it.

tcoratger (approved):
Correct and complete fix for a real panic-on-malicious-input bug: `preprocessed_next = Some(vec![])` is the one shape that slips through a pure length check, and it reaches `VerticalPair::new` paired with a width-`preprocessed_width` local row. I reproduced it by deleting only the new `uni-stark/src/verifier.rs` check and re-running the added test — `assertion left == right failed, left: 1, right: 0` at `matrix/src/stack.rs:68`, exactly as the description says. I confirmed the honest prover never emits this shape (`uni-stark/src/prover.rs:465-480` yields `None` when `!pre_next` and `(None, None)` when `preprocessed_width == 0`), so no well-formed proof is newly rejected, and both new checks sit before the challenger is initialised, so nothing attacker-controlled is absorbed first. Both negative tests pass here (`p3-uni-stark --test preprocessed_no_next_row`, `p3-batch-stark --test simple test_preprocessed`).

No BLOCKER and no MAJOR. Findings are MINOR/NIT only.

### [MINOR] `verify_constraints` is `pub` and still panics on the same shape — `uni-stark/src/verifier.rs:174`
The private `verify` path is now safe, but the `pub fn verify_constraints` that sits behind it takes the two `Option<&[..]>` raw and keeps the old match:
```rust
let preprocessed = match (preprocessed_local, preprocessed_next) {
    (Some(local), Some(next)) => VerticalPair::new(RowMajorMatrixView::new_row(local),
                                                   RowMajorMatrixView::new_row(next)),
    _ => VerticalPair::new(RowMajorMatrixView::new(&[], 0), RowMajorMatrixView::new(&[], 0)),
};
```
Two hazards for an external caller (a recursive verifier is the documented use case for the sibling `commitments_with_opening_points`): `(Some(len w), Some(len 0))` still hits the `assert_eq!` in `VerticalPair::new`, and `(Some(len w), None)` silently yields an **empty** preprocessed matrix, so every preprocessed constraint evaluates against nothing instead of erroring. Suggest either `debug_assert`-free explicit width validation returning `Err`, or narrowing the signature to a single `Option<(&[T], &[T])>` so the half-present case is unrepresentable.

### [MINOR] The stronger fix is at the type level — `uni-stark/src/proof.rs:69`
`preprocessed_local: Option<Vec<_>>` and `preprocessed_next: Option<Vec<_>>` make three illegal states representable: `Some(empty)`, `local = None` with `next = Some`, and lengths disagreeing with each other. This PR fixes one of them at runtime, in two places that must now stay in sync. Grouping them (`preprocessed: Option<PreprocessedOpening { local: Vec<_>, next: Option<Vec<_>> }>`) would make the present-but-empty `next` unconstructible and delete the need for the check in both verifiers. Worth a follow-up issue rather than blocking this one.

### [NIT] Residual present-but-empty laxness in the width-0 branches — `uni-stark/src/verifier.rs:250`, `batch-stark/src/verifier/mod.rs:551`
When the preprocessed width is `0`, both verifiers still accept `preprocessed_local = Some(vec![])` and `preprocessed_next = Some(vec![])`, because only lengths are compared. I traced this one: it is harmless (both `VerticalPair` sides are width 0, and no preprocessed commitment is bound), so it is not a bug — just the same laxness this PR is tightening elsewhere.

### [NIT] `PermutationWidthMismatch` is checked late — `batch-stark/src/verifier/mod.rs:723`
This PR correctly moves the preprocessed shape rejection ahead of the transcript. The permutation width check is still in the post-opening constraint loop, i.e. after the whole Fiat-Shamir replay and the PCS argument. It returns `Err` rather than panicking, so it is not the same class of bug, but it is the same argument for moving it up into the shape loop around line 542.

## Test coverage
Both negative tests are the right ones and they are what stops a future refactor silently reverting this. Two cheap gaps:
- `batch-stark/tests/simple.rs` uses a single instance, so the `air: Some(i)` payload is only ever exercised with `i = 0`. A two-instance batch tampering instance `1` would pin the indexing.
- No test for `preprocessed_next = Some(vec![x; pre_w])` with `preprocessed_next_row_columns()` empty. That is rejected by the pre-existing length check (`PreprocessedTraceWidthMismatch` / `PreprocessedWidthMismatch`), and asserting it keeps the two checks from being collapsed into one later.

**Bottom line:** merge as-is.

uni-stark/tests/preprocessed_no_next_row.rs
tcoratger: Leaving this one. Extracting a shared `uni-stark/tests/common/mod.rs` touches every existing uni-stark test file, which is a wider refactor than this PR should carry — better as its own change once there is a third copy to justify it. Thanks for confirming the test fails without the fix.

tcoratger:
Pushed a follow-up commit (d8f7cb7) applying the review, as maintainer. Feel free to amend, squash or drop it.

Applied:
- `verify_constraints` (public, reached directly by recursive verifiers) now rejects preprocessed rows of differing widths instead of panicking or silently folding against an empty matrix.
- uni-stark and batch-stark both reject a present preprocessed opening when the preprocessed width is zero, closing the residual laxness in the width-0 branches.
- uni-stark reports the next main row with its own two rejections (`TraceNextMismatch` / `UnexpectedTraceNext`) rather than the generic `OpenedValuesDimensionMismatch`, matching batch-stark.
- batch-stark also compares the next preprocessed row by width, so a full-width tamper is a width mismatch and a present-but-empty one is a presence rejection — the two rules stay distinct.
- The lookup permutation width check moved into the shape loop, ahead of the transcript.
- `InvalidProofShapeError` is now `#[non_exhaustive]`, and the three variants both crates raise carry `Option<usize>`.
- Tests: the batch tamper now uses two instances and tampers instance 1 (pinning the reported index), plus the full-width sibling tamper; and a uni-stark tamper for the next main row. Both were confirmed to fail without their respective production change.

Not applied:
- The type-level redesign of `OpenedValues` (grouping the preprocessed openings so the illegal states are unrepresentable). That is a larger change than this PR, better as its own follow-up.
- The shared `uni-stark/tests/common/mod.rs`; replied on that thread.

tcoratger (approved):

The added tests were each confirmed to fail with their production change reverted, so they pin the behaviour rather than merely passing alongside it.

Remaining open threads are deliberate scope calls, answered inline.

PR #2257, opened 2026-09-20
fix(uni-stark,circle): reject degree_bits below the PCS minimum trace height

kriss39:
## Summary

`p3-uni-stark`'s `validate_degree_bits` only bounds a proof's `degree_bits` from below by the ZK bit, because the univariate `Pcs` trait exposes no minimum. `CirclePcs` does have one: it asserts `domain.log_n >= 2` on commit, and the circle-domain selectors (`Point::v_n`, `Point::s_p_at_p`) are only defined for `log_n >= 1`. A proof that claims `degree_bits = 0` therefore passes the gate, and the verifier derives the trace domain and its selectors from it before the opening argument runs. In a debug build that is a panic in `circle/src/point.rs` (`v_n requires log_n >= 1`); a release build gets past the `debug_assert!`s and is only rejected later, inside the FRI verifier.

## Changes

- `p3-commit`: new `Pcs::log_min_trace_height()` (default `0`), the counterpart of `log_max_lde_height()`.
- `p3-circle`: implement it as `2`, sharing the constant with the two existing commit-time assertions.
- `p3-uni-stark`: `validate_degree_bits` takes the PCS minimum and rejects `degree_bits < is_zk + min` with the existing `DegreeBitsTooSmall` error (its message no longer mentions ZK only).
- `p3-batch-stark`: pass the minimum at both call sites.

The default keeps every other backend (`TwoAdicFriPcs`, `HidingFriPcs`, `TwoAdicStirPcs`, downstream implementations) unchanged.

## Tests

`uni-stark/tests/fib_air.rs::test_degree_bits_below_circle_pcs_minimum_rejected`: a Circle STARK proof with `degree_bits` set to `0` and `1` is rejected with `DegreeBitsTooSmall { air: None, minimum: 2, got }`. On `main` the same input panics at `circle/src/point.rs:77` in a debug build. `cargo nextest run -p p3-commit -p p3-circle -p p3-uni-stark -p p3-batch-stark`, workspace clippy and rustfmt are clean.


commit/src/pcs/univariate.rs
```diff
@@ -108,6 +108,15 @@ where
...
     fn log_max_lde_height(&self) -> usize;
 
+    /// The base-2 logarithm of the smallest trace domain this PCS can commit to.
+    ///
+    /// Verifiers reject a claimed `degree_bits` below this bound before deriving any domain
+    /// from it, so a backend whose domain arithmetic assumes a minimum size cannot be driven
+    /// below it by a malformed proof.
+    fn log_min_trace_height(&self) -> usize {
+        0
+    }
```
tcoratger: **[MINOR] The doc says "commit to", but the gate is applied to the *base* trace domain**

`validate_degree_bits` turns this into `degree_bits >= is_zk + log_min_trace_height()`, i.e. it constrains `base_degree_bits = degree_bits - is_zk`. But the domain a PCS actually commits to under ZK is the **extended** one, of size `1 << degree_bits`, not the base one — the base domain is only used for selectors and periodic columns (`init_trace_domain` at `uni-stark/src/verifier.rs:502`). So the doc sentence and the use site describe two different domains.

This is not a live bug: `CirclePcs::ZK = false` (`circle/src/pcs.rs:1112`), so `is_zk = 0` and the two coincide, and I confirmed the resulting bound matches the commit-time assertion exactly. It matters for the next backend. A ZK PCS with a commit minimum `m` would be gated at `degree_bits >= is_zk + m`, i.e. `m` on the base domain — strictly stronger than the `m` on the committed domain that its own `commit` asserts, so a legal proof at `degree_bits = m` would be rejected. That is a completeness regression, not a soundness one, which is why it is MINOR.

**Fix:** either say "the smallest **base** trace domain a proof may claim" (matching the use), or keep the commit wording and split the two bounds: `degree_bits >= log_min_trace_height()` for the committed domain and a separate selector bound for the base domain.

**[MINOR] A defaulted-to-permissive security bound**

`log_max_lde_height` has no default — every backend is forced to state it. `log_min_trace_height` defaults to `0`, so a new backend that assumes a minimum domain size silently inherits exactly the bug this PR fixes, with no compile error. For a bound whose entire purpose is verifier input validation, requiring it (and updating `commit/src/testing.rs:177`, `fri/src/two_adic_pcs.rs:909`, `fri/src/hiding_pcs.rs:417`, `stir/src/pcs.rs:2098` to return `0` explicitly) makes the safe answer the deliberate one. I verified the four affected backends: only `p3-circle` has a real minimum today, so the churn is four one-line impls.

uni-stark/src/verifier.rs
```diff
@@ -65,12 +65,16 @@ pub fn validate_degree_bits(
...
     is_zk: usize,
+    min_log_degree: usize,
     max_log_degree: usize,
 ) -> Result<(usize, usize), InvalidProofShapeError> {
-    if degree_bits < is_zk {
+    // The base trace domain is `degree_bits - is_zk` bits tall and must be one the PCS can
+    // commit to; its selectors are derived from it before the opening argument runs.
+    let minimum = is_zk + min_log_degree;
+    if degree_bits < minimum {
         return Err(InvalidProofShapeError::DegreeBitsTooSmall {
```
tcoratger: **[NIT] The comment overstates what the base domain is used for**

> The base trace domain is `degree_bits - is_zk` bits tall and must be one the PCS can commit to

The base trace domain is not committed. Under ZK the prover commits the *extended* domain (`1 << degree_bits`); the base domain only feeds `natural_domain_for_degree(degree >> is_zk)` at line 502 for the selectors and the periodic columns. The second half of your sentence ("its selectors are derived from it before the opening argument runs") is the accurate justification and is on its own enough — it is what makes `Point::v_n`'s `debug_assert!(log_n >= 1)` reachable from a proof field.

The arithmetic is right either way, because `CirclePcs::ZK = false`. Suggest dropping the "must be one the PCS can commit to" clause, or spelling out that with `is_zk = 0` the two domains coincide.

circle/src/pcs.rs
```diff
@@ -228,6 +228,11 @@ where
     })
 }
 
+/// The smallest trace domain `CirclePcs` commits to: the bivariate layer folds one bit and
+/// circle FRI needs one more, and the domain selectors (`v_n`, `s_p`) are only defined for
+/// `log_n >= 1`.
+const LOG_MIN_TRACE_HEIGHT: usize = 2;
```
tcoratger: **[Verified] `2` is the right constant and the boundary is not off by one**

I checked this against the code rather than the comment. `v_n` (`circle/src/point.rs:77`) and `s_p_at_p` (`:121`) `debug_assert!(log_n >= 1)`, and the commit path needs `log_n >= 2` for the bivariate fold plus one FRI bit. Taking the max gives `2`, matching both `assert!`s this constant now feeds.

More importantly I confirmed `2` is *accepted*, not just that `0` and `1` are rejected: an honest `CirclePeriodicProductAir` proof over a 4-row trace has `proof.degree_bits == 2` and verifies cleanly under this branch. So `degree_bits < minimum` (strict `<`) is correct and the legal minimum is not excluded.

**[NIT]** Naming: `log_min_trace_height` vs `log_max_lde_height` — one says *trace*, the other says *lde*, and they are not bounds on the same object (see my comment on `commit/src/pcs/univariate.rs`). `log_min_trace_height` / `log_max_trace_height` would read as a matched pair.

uni-stark/tests/fib_air.rs
```diff
@@ -593,6 +593,37 @@ fn test_degree_bits_too_small_for_zk_rejected() {
...
+            p3_uni_stark::VerificationError::InvalidProofShape(
+                InvalidProofShapeError::DegreeBitsTooSmall { air, minimum, got },
+            ) => {
+                assert_eq!(air, None);
+                assert_eq!(minimum, 2);
+                assert_eq!(got, degree_bits);
+            }
+            _ => panic!("unexpected error: {err:?}"),
+        }
+    }
```
tcoratger, line 638: **[MINOR] No positive test at the boundary this PR introduces**

The loop covers `degree_bits in [0, 1]`, so it proves the rejection. Nothing proves that `degree_bits == 2` — the smallest *legal* value — is still accepted. A `<` accidentally becoming `<=` in `validate_degree_bits`, or `LOG_MIN_TRACE_HEIGHT` drifting to `3`, would break no test in the repo: every existing circle test uses a 16-row trace (`degree_bits = 4`), comfortably above the bound.

I wrote this test in a scratch worktree and it passes, so it is a pure regression guard, not a discovered bug:

```rust
// A 4-row circle trace is the smallest CirclePcs commits to; degree_bits == 2 must verify.
let air = CirclePeriodicProductAir { column: vec![CircleVal::TWO, CircleVal::from_u32(3)], degree: 3 };
let trace = RowMajorMatrix::new_col((0..4).map(|i| air.column[i % 2].exp_u64(3)).collect());
let proof = prove(&config, &air, trace, &[]).unwrap();
assert_eq!(proof.degree_bits, 2);
verify(&config, &air, &proof, &[]).expect("boundary degree_bits must be accepted");
```

**Fix:** fold those five lines into this test before the rejection loop.
**Verify:** `cargo test -p p3-uni-stark --test fib_air degree_bits`.

tcoratger (approved):
Adds a PCS-supplied lower bound on `degree_bits` so a malformed proof cannot drive `CirclePcs`'s domain arithmetic below `log_n = 2`. I checked the boundary by hand rather than trusting the constant: I built an honest 4-row circle proof in a scratch test, confirmed its `degree_bits` is exactly `2`, and confirmed it still verifies under the new gate — so `>=` at `2` is correct and there is no off-by-one at the legal minimum. I also confirmed `CirclePcs::ZK = false` (`circle/src/pcs.rs:1112`), so `minimum = is_zk + min = 0 + 2` reproduces the commit-time assertion exactly, and that the gate runs at `uni-stark/src/verifier.rs:458`, before the challenger is initialised.

No BLOCKER and no MAJOR. Findings are MINOR/NIT only.

### [Info] I measured the other backend, and the default of `0` does not hide a second panic
The obvious worry with a permissive default is that `TwoAdicFriPcs` has its own unstated minimum. I tested it: a debug build, `degree_bits` forced to `0`, `1` and `2` on an honest BabyBear/`TwoAdicFriPcs` proof, all three return
```
Err(InvalidOpeningArgument(CommitPhaseOpeningsCountMismatch { expected: 0, got: 2 }))
```
No panic, no `debug_assert` fired. I traced why: `TwoAdicMultiplicativeCoset::selectors_at_point` (`commit/src/domain.rs:315-324`) is total at `log_size = 0` — `z_h = point - 1`, and `point != 1` is already guaranteed by the `OodPointInDomain` check — so the two-adic path degrades to a late `Err` rather than an abort. The default is therefore safe today. It is still *late*: rejection happens after the full transcript replay and inside FRI. If you want the same early-rejection property for the two-adic backend, its real bound is roughly `log_final_poly_len.saturating_sub(log_blowup)`, which is what makes the commit phase empty above. Follow-up, not this PR.

### [MINOR] `p3-batch-stark`'s two new call sites are untested — `batch-stark/src/verifier/mod.rs:134` and `:433`
`uni-stark/tests/fib_air.rs` covers the uni-stark gate. Nothing exercises either batch-stark call site with a PCS that has a non-zero minimum, so a `0` accidentally passed at one of them (or the argument order silently swapped with `log_max_lde_height`, which is the adjacent `usize` parameter) would not be caught. A circle-backed batch proof with `degree_bits[i] = 1` asserting `DegreeBitsTooSmall { air: Some(i), minimum: 2, .. }` would close both.

## Test coverage
`test_degree_bits_below_circle_pcs_minimum_rejected` is the right negative test and it covers both sub-minimum values. The gap is the positive boundary, covered below in the inline comment on `uni-stark/tests/fib_air.rs`.

**Bottom line:** merge as-is; the two MINORs are follow-ups.

circle/src/pcs.rs
tcoratger: Implemented the rest of this review, but not the rename. `log_max_lde_height` is pre-existing public API on the `Pcs` trait with five impls and several call sites, and renaming it is a separate, purely cosmetic change — out of scope here. Leaving this thread open so it is not lost.

tcoratger:
Pushed a follow-up commit (dbf742b) applying the review, as maintainer. Feel free to amend, squash or drop it.

Applied:
- Reworded the trait doc to describe the *base* trace domain, which is what the gate constrains, and explained the relationship to the committed domain under ZK. Same for the comment in `validate_degree_bits`.
- Dropped the permissive default: `log_min_trace_height` is now required, and the four backends without a minimum return `0` explicitly (`commit/src/testing.rs`, `fri/src/two_adic_pcs.rs`, `fri/src/hiding_pcs.rs` — delegating to the wrapped scheme — and `stir/src/pcs.rs`).
- Positive boundary: `test_degree_bits_below_circle_pcs_minimum_rejected` now proves over a 4-row trace, asserts `degree_bits == 2` and verifies it before tampering. Confirmed it fails if `<` becomes `<=` or if the constant drifts to `3`.
- New `test_batch_degree_bits_below_circle_pcs_minimum_rejected` covers both batch-stark call sites: `verify_batch`, and `commitments_with_opening_points` called directly the way a recursive verifier would. Confirmed each site independently fails the test when its bound is replaced by `0`.

Not applied:
- The `log_max_lde_height` -> `log_max_trace_height` rename; replied on that thread.

tcoratger (approved):

The added tests were each confirmed to fail with their production change reverted, so they pin the behaviour rather than merely passing alongside it.

Remaining open threads are deliberate scope calls, answered inline.

PR #2258, opened 2026-09-20
fix(mersenne-31): print the canonical value in Display and Debug

kriss39:
## Summary

`Mersenne31` allows the redundant `value == P` encoding of zero, and ordinary arithmetic produces it (`-Mersenne31::ZERO` goes through `new_reduced(P - 0)`; the crate's own test fixtures keep `ZEROS = [ZERO, new_reduced(P)]` for this reason). `PartialEq`, `Hash`, `Ord` and `Serialize` canonicalize before comparing or encoding, but `Display` and `Debug` formatted `self.value` directly, so an element that compares, hashes and serializes as `0` printed as `2147483647`. `Goldilocks`, which has the same kind of non-canonical representation, already formats `as_canonical_u64()`.

## Changes

`Display` and `Debug` for `Mersenne31` format `as_canonical_u32()`.

## Tests

`mersenne_31::tests::display_and_debug_are_canonical`: `-ZERO` prints as `0` in both `{}` and `{:?}`, `NEG_ONE` prints as `2147483646`. Fails on `main` with `left: "2147483647", right: "0"`. `cargo nextest run -p p3-mersenne-31` (246 tests) and clippy are clean.


mersenne-31/src/mersenne_31.rs
```diff
@@ -155,13 +155,14 @@ impl PartialOrd for Mersenne31 {
...
+        Display::fmt(&self.as_canonical_u32(), f)
     }
 }
 
 impl Debug for Mersenne31 {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        Debug::fmt(&self.value, f)
+        Debug::fmt(&self.as_canonical_u32(), f)
     }
 }
```
tcoratger, line 167: **[NIT] The packed types still print non-canonically**

`PackedMersenne31AVX2` / `AVX512` / `Neon` get `Debug` from `#[derive]` over the raw SIMD lane type (`x86_64_avx2/packing.rs:31`, `x86_64_avx512/packing.rs:33`, `aarch64_neon/packing.rs:30`), so a lane holding the `P` encoding of zero still prints as `2147483647` — and the derive prints the vector representation, not the field elements. Out of scope for this PR and arguably fine for a packed type, but it means "`Mersenne31` prints canonically" is only true for the scalar.

No change requested.

tcoratger (approved):
Right fix. `Mersenne31` genuinely admits the redundant `value == P` encoding of zero, `PartialEq`/`Hash`/`Ord`/`Serialize` all canonicalize first (`mersenne_31.rs:181-197`, and `serde_json::to_string(&redundant_zero) == "0"` is already asserted at line 728), so `Display`/`Debug` printing the raw limb was the one accessor that disagreed with every other observable. `as_canonical_u32` (line 536) is a single `== ORDER_U32` branch, so there is no cost worth worrying about, and neither impl is on a hot path.

I checked that this is not load-bearing anywhere before calling it cosmetic: there is no `FromStr` for `Mersenne31`, no `to_string`/parse round-trip, no golden output or `should_panic(expected = ...)` string containing `2147483647`, and the only `format!` mentioning the type is a benchmark name in `monolith/benches/monolith.rs:44`. `cargo test -p p3-mersenne-31 display_and_debug` passes here.

**Bottom line:** merge as-is.

mersenne-31/src/mersenne_31.rs
tcoratger: Agreed, and leaving the packed types alone. Their derived `Debug` prints the SIMD vector representation, which is what you want when debugging packed arithmetic; canonicalizing would mean unpacking lanes and would change what the output means. This PR is scoped to the scalar type, where `Display`/`Debug` were the one accessor disagreeing with `PartialEq`/`Hash`/`Ord`/`Serialize`.

tcoratger (approved):
Approving as-is — the one open thread asked for no code change, and it is answered inline.

PR #2259, opened 2026-09-20
fix(stir): use the rigorous field size on the standalone config path

Bruce039:
`StirConfig` has two constructor paths that should be pricing the same thing but weren't.

`try_new_with_pcs_batch` (what `TwoAdicStirPcs` calls) computes the field size with `pcs_budget::field_bits`, i.e. `floor(log2 |E|)` and one bit less under the Johnson bound, and the comment on that helper explains why `Field::bits()`'s ceiling isn't a lower bound. The standalone constructors (`new`, `try_new`, `new_with_options`, ...) kept passing `EF::bits()` into the same proximity-gap formulas. Since the constructor bisects `eta` until the reported bits clear `security_level`, the standalone schedule is priced against a field that is one bit larger than the rigorous bound under `CapacityBound` and two bits larger under `JohnsonBound` (`155` vs `153` on `QuinticTrinomialExtensionField<KoalaBear>`), so re-evaluating it with the rigorous size lands below the level it claims. In the failing fixture case (`CapacityBound`, `log_deg=8`, `log_blowup=1`, `sec=80`) the observed shortfall is `0.69` bits; the two-bit figure is the field-size delta under Johnson, not a measured gap. The PCS path is unaffected; the crate docs and `benches/stir.rs` go through the standalone path.

This makes both paths use `pcs_budget::field_bits` and changes `test_stir_config_every_stage_meets_buffered_target` to recompute each stage with that value instead of `EF::bits()`. With only the test change applied, the test fails on `main` (`CapacityBound intermediate-fold below buffered target ... fold_alg=72.3106 + pow=12 < 85.0000`); with the config change it passes.

The follow-up commit prices the initial quotient batching term (`try_new_with_batching` and the `initial_batching_error` accessor) off the same rigorous field size, so no term in either path is left on `EF::bits()`. `cargo nextest run -p p3-stir` is green on the current head (267 tests), as is `uni-stark`'s STIR integration test.


tcoratger (commented):
**Direction check first, since that is the whole question: this change is strictly MORE conservative, and it is the right fix.** `Field::bits()` is `order().bits()`, i.e. `floor(log2|E|) + 1` — an upper bound on `log2|E|`. `pcs_budget::field_bits` returns `order().bits() - 1 = floor(log2|E|)`, a rigorous lower bound, minus one more under Johnson to pay for `prox_gaps_error_jb_at_m` keeping only BCSS25's dominant term. Every consumer of `field_size_bits` is monotone increasing in it (`security/src/assumption.rs:190,213` compute `field_size_bits - error`; `soundness.rs:105` computes `field_size_bits + log2(1 - ratio)`), and `minimum_eta_for_target` bisects for the smallest `eta` clearing the target — so a smaller `field_size_bits` forces a larger safety gap. No path gets cheaper. No off-by-one: for `|E| = 2^k` (binary fields) `order().bits() = k + 1`, so `field_bits` lands on exactly `k`, and for every non-power-of-two `|E|` it is a strict underestimate. Correct in both regimes.

I reproduced the empirical claim. Reverting only the `config.rs` hunk and keeping the test change, `test_stir_config_every_stage_meets_buffered_target` fails with `CapacityBound intermediate-fold below buffered target (log_deg=8, log_blowup=1, log_fold=2, sec=80, max_pow=20): fold_alg=72.3106 + pow=12 < 85.0000` — so `main`'s standalone schedule was short of its own buffered target once re-priced honestly. I also instrumented the two paths on `BinomialExtensionField<BabyBear, 4>` (`EF::bits() = 124`): at `log_deg=16, log_blowup=1, sec=80`, query counts and PoW bits are **unchanged** in both regimes, and only `final_eta` moves, from `1.98e-4` to `2.60e-4` under `JohnsonBound` (unchanged under `CapacityBound`). So the honest accounting costs essentially nothing in proof size here — good trade.

One thing to note on the PR description: "two bits short" overstates it. For `CapacityBound` the field-size delta is one bit (`124 -> 123`), and the observed shortfall at the failing stage is `0.69` bits, not two. Two bits is the `JohnsonBound` delta only (`EF::bits() - 2`). Worth correcting in the commit message so the changelog is accurate.

**Findings not anchorable to the diff** — see the inline note on `config.rs:798` for the substantive one; it concerns `stir/src/config.rs:974` and `stir/src/config.rs:1183`, which are outside the diff hunks.

Stale prose the change leaves behind, all NIT, none of it load-bearing:
- `stir/src/pcs_budget.rs:194` — "PCS schedules use a rigorous integer lower bound on `log2(|E|)`". After this PR *every* schedule does, not just PCS ones. One word.
- `stir/src/config.rs:483` — `CombineEtaExceedsCeiling`'s message prints `EF::bits() = {field_size_bits}` but the value handed in is `field_bits`, one or two below `EF::bits()`. This was already wrong on `main` (the variant is only reachable on the PCS path, which already used `field_bits`), so the PR does not introduce it — but the PR is the natural place to fix the label, since it is now wrong on both paths. Same for the envelope doc at `stir/src/config.rs:43`.

`cargo test -p p3-stir`: 265 passed, 0 failed.

**Bottom line:** merge after the `EF::bits() - 1` leftovers are either fixed or explicitly ruled out of scope. The change itself is correct and moves in the safe direction; I am withholding approval only because the PR claims to close a two-path divergence and one term in that same path is still on the other side of it.

stir/src/config.rs
```diff
@@ -789,12 +789,13 @@ where
...
             batch.validate(log_starting_degree)?;
-            crate::pcs_budget::field_bits::<EF>(params.soundness_type)
-        } else {
-            EF::bits()
-        };
+        }
+        // Both paths account for the same field: `Field::bits()` is the ceiling of
+        // `log2(|E|)`, one above the integer lower bound the proximity-gap bounds need,
+        // and Johnson reserves one more (see `pcs_budget::field_bits`).
+        let field_size_bits = crate::pcs_budget::field_bits::<EF>(params.soundness_type);
```
tcoratger, line 810: **[MAJOR] Two `EF::bits() - 1` call sites in this same function survive the change, and under `JohnsonBound` they are one bit more optimistic than everything else in the schedule — `stir/src/config.rs:974`, `stir/src/config.rs:1183`**

The comment above says "Both paths account for the same field", but ~175 lines below, inside the very constructor this line feeds, the initial quotient-combination term is still priced off `EF::bits() - 1`:

```rust
// stir/src/config.rs:967-980
        if batching_term != 0 {
            let batching_eta = minimum_eta_for_target(
                params.soundness_type.stir_eta_upper_bound(log_inv_rate),
                buffered_security_level,
                |eta| {
                    initial_batching_error(
                        params.soundness_type,
                        EF::bits() - 1,          // <-- not `field_size_bits`
                        ...
```

and the public reporting accessor does the same:

```rust
// stir/src/config.rs:1179-1187
    pub fn initial_batching_error(&self) -> f64 {
        ...
        initial_batching_error(
            self.soundness_type,
            EF::bits() - 1,                      // <-- same
```

Why this is not cosmetic. `field_bits` = `EF::bits() - 1` under `CapacityBound` but `EF::bits() - 2` under `JohnsonBound`, and `pcs_budget.rs:194-199` states exactly what that extra bit pays for:

> Johnson reserves one further bit: twice BCSS25's dominant term upper-bounds the full expression [...] Applying this reserve to the field size **covers both the new block and the existing STIR proximity-gap calls**.

`initial_batching_error` *is* one of those calls — it bottoms out in `assumption.prox_gaps_error_at_log_eta` (`stir/src/soundness.rs:171`), the same function, which on `JohnsonBound` routes to `prox_gaps_error_jb_at_m` and keeps only the dominant term. So by the helper's own documented rule this term owes the reserve and is not paying it. Under `JohnsonBound` the alpha-batching error is credited one bit it has not earned, while every other proximity-gap term in the same schedule is now charged correctly. `CapacityBound` is unaffected (the two expressions coincide there).

Reachability: `batching_term != 0` requires a non-empty `quotient_batches` with some `count > 1`, which the PCS path never supplies — `try_new_with_pcs_batch` passes `&[]` (`config.rs:711`) and budgets batching through `PcsBatch::algebraic_bits` with the proper `field_bits`. So `config.rs:974` is reached *only* from `pub fn try_new_with_batching` (`config.rs:627`) — i.e. precisely the standalone path this PR is about. It has no in-repo callers, but it is published API. `config.rs:1183` is worse placed: `self.quotient_batches` is non-empty on the PCS path too (`config.rs:1151-1152`), so the accessor reports a one-bit-optimistic number on *both* paths, including the one `main` already had right.

This predates the PR and the PR does not worsen it — the change is a strict improvement in the safe direction. I am raising it as MAJOR because the stated invariant ("both paths account for the same field") is the thing being shipped, and a reader will reasonably conclude the audit is complete when one term in that path is still on the old accounting.

**Fix:** replace both with `crate::pcs_budget::field_bits::<EF>(params.soundness_type)` / `field_bits::<EF>(self.soundness_type)`. If you believe the batching term genuinely does not need the Johnson reserve, say so in a comment at `config.rs:974` and amend `pcs_budget.rs:194-199`, which currently claims the opposite — but do not leave the two documents disagreeing.

**Verify:** extend the `cases` table in `test_stir_config_every_stage_meets_buffered_target` with a `try_new_with_batching` arm carrying a class of multiplicity `> 1` under `jb`, recompute its stage with `field_bits::<EF>(JohnsonBound)` exactly as the new line at `config.rs:1887` does for the fold stages, and assert it clears the buffered target. On `main` and on this branch that arm should fail; with the two sites fixed it should pass. Also tighten `stir/src/pcs.rs:3658`, which currently asserts `initial_batching_error()` against an independently computed value — check which field size that expectation uses, since it may be enshrining the optimistic one.

stir/src/config.rs
```diff
@@ -789,12 +789,13 @@ where
...
+        if let Some(batch) = pcs_batch {
             batch.validate(log_starting_degree)?;
-            crate::pcs_budget::field_bits::<EF>(params.soundness_type)
-        } else {
-            EF::bits()
-        };
+        }
+        // Both paths account for the same field: `Field::bits()` is the ceiling of
+        // `log2(|E|)`, one above the integer lower bound the proximity-gap bounds need,
+        // and Johnson reserves one more (see `pcs_budget::field_bits`).
```
tcoratger: **[NIT] The comment explains the ceiling but not the Johnson bit, which is the part a reader will trip on**

"`Field::bits()` is the ceiling of `log2(|E|)`, one above the integer lower bound" is precise and correct — `order().bits() = floor(log2|E|) + 1`. But "and Johnson reserves one more" reads like a second rounding step, when it is a completely different thing: it is a factor-2 slack payment for `prox_gaps_error_jb_at_m` dropping all but BCSS25's dominant term (`security/src/assumption.rs:221-228`). Someone auditing the parameter derivation will want to know that the two `-1`s have unrelated justifications.

**Fix:** "...one above the integer lower bound the proximity-gap bounds need. Johnson reserves one further bit, not as rounding but to cover the dominant-term-only Theorem 1.5 approximation (see `pcs_budget::field_bits`)."

stir/src/config.rs
```diff
@@ -1883,6 +1883,9 @@ mod tests {
                     },
                 );
 
+                // Recompute with the rigorous field size the schedule is accountable to.
+                let field_size_bits = crate::pcs_budget::field_bits::<EF>(soundness_type);
```
tcoratger, line 1899: **[NIT] Good change — but note what this test does and does not now cover**

Recomputing each stage against `field_bits` instead of `EF::bits()` is exactly right: the test now asserts the schedule against the same field size the constructor priced it with, so it is a real regression guard rather than a tautology against a looser number. I confirmed it bites: with only this hunk applied and the `config.rs:798` hunk reverted, it fails with `fold_alg=72.3106 + pow=12 < 85.0000`.

What it still misses is the batching stage. `cases` only drives `new_with_options`, so `batching_term` is always `0` and the `EF::bits() - 1` at `config.rs:974` is never exercised. See the note on `config.rs:798` for the arm to add.

**Fix:** none required here; just worth knowing the coverage boundary.

tcoratger:
Pushed `4f42701b` to this branch (append-only — nothing rebased or amended), applying the review feedback. Summary so the extra commit is no surprise.

**MAJOR.** The two remaining `EF::bits() - 1` call sites — `config.rs:974` inside `try_new_with_optional_combine`, and the `initial_batching_error()` reporting accessor — now go through the same rigorous field size as the rest of the schedule. The first simply reuses the `field_size_bits` local that your hunk already made unconditional; the second calls `pcs_budget::field_bits::<EF>(self.soundness_type)`.

**Direction check, with numbers**, since a parameter change has to be shown to move the safe way. `BinomialExtensionField<BabyBear, 4>` (`EF::bits() = 124`), `JohnsonBound`, `log_degree = 8`, `log_blowup = 1`, `log_fold = 2`, `security_level = 80`, classes `[(8, 64), (7, 32)]`, buffered target 85 bits:

| | before | after |
|---|---|---|
| field size fed to the batching term | 123 | 122 |
| round-0 eta | 1.767767e-2 | 1.964185e-2 |
| round-0 queries | 140 | 142 |
| bits the stage actually delivers, priced at 122 | **84.3196** (0.68 short of 85) | **85.0599** (clears) |
| number the accessor reports | 85.3196 (over-claim) | 85.0599 |

I swept 30 configurations (3 batch shapes x 5 parameter sets x both assumptions). Under `CapacityBound` every derived number is identical (123 == 123). Under `JohnsonBound` only round-0 eta and the reported batching figure move: eta rises in every case, round-0 query counts rise or stay (140 -> 142, 69 -> 70, 136 -> 136, 67 -> 68, 144 -> 146), and no query count, query-PoW bit or folding-PoW bit anywhere decreases. Configurations that were already infeasible report exactly one bit less headroom and stay infeasible.

Analytically this is forced: the proximity-gap bounds are `field_size_bits - log2(coefficient)`, so lowering the field size lowers the retained bits by exactly one at every eta; `minimum_eta_for_target` bisects for the smallest clearing eta and can therefore only return a larger one; and `final_eta` only ever takes a `max`. No schedule gets cheaper.

**Tests.** Two new ones in `stir/src/config.rs`.

- `the_initial_quotient_batching_stage_meets_the_buffered_target` drives `try_new_with_batching` with multiplicity-greater-than-one classes under both assumptions, recomputes the stage against `field_bits::<EF>`, asserts it clears the buffered target, and asserts the accessor quotes that same number. Confirmed it fails without the fix: `JohnsonBound (log_deg=8, log_blowup=1, log_fold=2, sec=80, batches=[(8, 64), (7, 32)]): batching retains 84.3196 bits < 85.0000`.
- `standalone_and_pcs_paths_derive_the_same_schedule` is the guard for the divergence this PR closes. It builds the same parameters through `new_with_options` and through `try_new_with_pcs_batch` with an empty batch, and asserts every derived number agrees. Confirmed it fails with your `config.rs:798` hunk reverted: `round 0 eta differs`.

**NITs.** Rewrote the `config.rs:798` comment so the two subtractions are separated — rounding `log2(|E|)` down and the Johnson factor-of-two reserve have unrelated justifications. Expanded the `pcs_budget::field_bits` doc with the derivation and an explicit statement that both adjustments only ever shrink the field size, so the result is always the conservative side. Relabelled the `CombineEtaExceedsCeiling` message and the `StirParameters` envelope doc, both of which printed `EF::bits()` while the value handed in is one or two below it. `stir/src/pcs.rs:3658` was already correct — its `123.` is the rigorous size under `CapacityBound`, where the two coincide — so I only added a comment saying why.

`cargo test -p p3-stir`: 267 passed, 0 failed. `cargo clippy -p p3-stir --all-targets` clean; `cargo +nightly fmt --check` clean.

One thing I deliberately did not touch, so it is yours to decide: the PR body says the standalone path was "two bits short". The measured shortfall at the failing stage is 0.69 bits, and the field-size delta is one bit under `CapacityBound` (124 -> 123); two bits is the `JohnsonBound` delta only. Worth correcting in the description so the changelog is accurate — I left your text alone rather than rewriting it.

tcoratger (approved):
Approving — the MAJOR from my earlier review is fixed, and the parameter change is demonstrated conservative rather than asserted.

Both new tests were confirmed to fail without their production change, so they pin the behaviour instead of merely passing alongside it. The cross-path guard in particular fails when the original hunk is reverted, which is what stops this diverging again.

All review threads resolved.

stir/src/config.rs
Bruce039: Agreed on both counts — thanks for pushing the batching-term fix. I re-ran `cargo nextest run -p p3-stir` on `4f42701b` (267 passed) and rewrote the description so it no longer claims a flat two-bit gap: the field-size delta is one bit under CapacityBound and two under Johnson, and the fixture's measured shortfall is 0.69 bits.

PR #2260, opened 2026-09-20
fix(sumcheck): require a cryptographic RNG for HVZK mask sampling

Bruce039:
The HVZK sumcheck in `p3-sumcheck::zk` draws its masks from the `rng` the caller hands to `ZkProver::into_sumcheck` / `SumcheckProver::into_zk_sumcheck`, and those were bounded by plain `Rng`. The hiding layers above it went through this already: `HidingFriPcs`, `MerkleTreeHidingMmcs` and `HidingWhirPcs`'s adapter all require `CryptoRng`. The crate that actually samples the masks didn't, so a downstream user of `p3-sumcheck` (it's published and re-exports `ZkProver`/`ZkPrefixProver`/`ZkSuffixProver`) can build the masked sumcheck on `SmallRng` and the compiler says nothing. I checked what that buys an observer: with the seed known, replaying `encoding.sample_message` reproduces the round-1 mask exactly and the plain round-polynomial coefficients come straight off the wire.

Changes:

- `R: CryptoRng` on `into_sumcheck`, `into_zk_sumcheck`, `sample_masks` and the three simulator entry points that share it.
- WHIR's private `HidingWhirProver::commit`/`prove` get the same bound (the public adapter already had it, so nothing changes for `HidingWhirPcs` users).
- Tests and the two benches that exercise these paths move from `SmallRng` to `StdRng` (`std_rng` dev-feature on `rand`, as `p3-whir` already does).

No behaviour change for callers that already pass a cryptographic generator. `cargo nextest run -p p3-sumcheck -p p3-whir`: 810 passed; clippy and rustfmt clean.


whir/src/pcs/zk/prover/mod.rs
```diff
@@ -96,7 +96,7 @@ where
     /// Nothing is bound here.
     ///
     /// The root is returned instead, and the caller binds it.
-    pub fn commit<R: Rng>(
+    pub fn commit<R: CryptoRng>(
```
tcoratger, line 99: **[MINOR] The WHIR base case still samples its one-time pads under `R: Rng` — `whir/src/pcs/zk/base_case/prover.rs:83`**

You bounded `HidingWhirProver::commit`/`prove` here, but the function they hand the RNG to is still generic over plain `Rng`:

```rust
// whir/src/pcs/zk/base_case/prover.rs:83
        R: Rng,
```

and that is the function that actually draws the Construction 7.2 pads:

```rust
// whir/src/pcs/zk/base_case/prover.rs:105-107
let fresh_message: Vec<EF> = (0..code.message_len).map(|_| rng.random()).collect();
let fresh_randomness: Vec<EF> = (0..code.randomness_len).map(|_| rng.random()).collect();
// ...:129-131  per-group blinds
let message = encoding.sample_message(rng);
let randomness = encoding.sample_randomness(rng);
```

`g~` and `r_g` are one-time pads for the reveals `f*` and `r*`. Predict them and the reveal is the witness in the clear — the exact leak this PR closes one layer up.

Not exploitable today: `base_case` is `pub(crate)` (`whir/src/pcs/zk/mod.rs:59`) and the only caller is `prove` at `whir/src/pcs/zk/prover/mod.rs:490`, which you just tightened. So the invariant holds, but it holds by caller discipline rather than by the type system, in the one file where the pads are physically drawn.

**Fix:** change `R: Rng` to `R: CryptoRng` at `base_case/prover.rs:83` and swap the import at line 12 to `use rand::{CryptoRng, RngExt};`. Zero cost — the sole caller already satisfies it, and `CryptoRng: Rng` keeps `RngExt::random` available through the blanket impl.
**Verify:** `cargo check -p p3-whir --all-targets` after the swap; nothing should need touching except `whir/src/pcs/zk/base_case/tests.rs:158`'s fixture RNG if it is a `SmallRng`.

sumcheck/src/zk/prover/common.rs
```diff
@@ -66,7 +67,7 @@ where
     Enc: ZkEncodingWithRandomness<F>,
     Enc::Codeword: Matrix<F>,
     M: Mmcs<F>,
-    R: Rng,
+    R: CryptoRng,
```
tcoratger, line 71: **[MINOR] The last hop to the RNG is still `R: Rng`, and it is on a published trait**

`sample_masks` is now `CryptoRng`, but it immediately forwards into `ZkEncoding`, whose sampling methods stayed loose:

```rust
// zk-codes/src/encoding.rs:41,55,58,66
fn sample_message<R: Rng>(&self, rng: &mut R) -> Vec<F>;
fn sample_randomness<R: Rng>(&self, rng: &mut R) -> Vec<F>;
fn encode<R: Rng>(&self, msg: &[F], rng: &mut R) -> Self::Codeword;
fn simulate<R: Rng>(&self, query_set: &[usize], rng: &mut R) -> Vec<F>;
```

This compiles fine (`CryptoRng: Rng`), so the protocol path is safe. But `p3-zk-codes` is published and `ReedSolomonZkEncoding` is `pub`, so a downstream user can still write `encoding.sample_message(&mut SmallRng::seed_from_u64(0))` and get a predictable mask message — the same failure mode, one abstraction lower.

I would not call this a bug: `ZkEncoding` is a generic encoder, and `encode`/`simulate` have legitimate deterministic uses (the simulator, test vectors). But `sample_message` and `sample_randomness` exist specifically to produce hiding material, and the doc on `sample_message` says so ("Samples a uniformly random message from this encoding's message space" for masking). Those two are the ones worth tightening.

**Fix:** either tighten `sample_message`/`sample_randomness` to `R: CryptoRng` (`encode`/`simulate` can stay `Rng`), or add one line to the trait doc saying the bound is deliberately loose and the protocol entry points carry the real requirement, so the next reader does not read the missing bound as an oversight.
**Verify:** `cargo check -p p3-zk-codes -p p3-sumcheck -p p3-whir --all-targets`.

sumcheck/src/zk/prover/zk_prover.rs
```diff
@@ -177,7 +183,7 @@ where
         F: TranscriptField,
         EF: TwoAdicField,
         Enc::Codeword: Matrix<EF>,
-        R: Rng,
+        R: CryptoRng,
```
tcoratger, line 220: **[MINOR] No regression test pins the bound**

The fix is entirely type-level, so the compiler enforces it — I am deliberately not calling this a MAJOR, because unlike a runtime check this cannot silently regress: someone has to edit this exact line to undo it.

That said, the repo already has the idiom to pin it (`bus/src/builder.rs:54` uses a `compile_fail` doctest), and it is three lines here. It also documents the intent for the next person who hits a bound error and is tempted to relax it.

**Fix:** add to the `into_sumcheck` doc, under the new `# Randomness` section:

```rust
/// ```compile_fail
/// # use rand::SeedableRng;
/// # use rand::rngs::SmallRng;
/// fn f<F, EF, Enc, M, L>(p: p3_sumcheck::zk::ZkProver<F, EF, Enc, M, L>) {
///     let mut rng = SmallRng::seed_from_u64(0);
///     // `SmallRng` is not a `CryptoRng`: masks must not come from xoshiro.
///     let _ = &mut rng;
/// }
/// ```
```

A tighter version that actually instantiates the call is better if the generics can be pinned down cheaply with the `test-util` feature's concrete types.
**Verify:** `cargo test -p p3-sumcheck --doc` — the doctest must fail to compile, and flipping the bound back to `Rng` must make it start compiling (i.e. the doctest then fails).

sumcheck/src/zk/simulator.rs
```diff
@@ -65,7 +65,7 @@ where
     Enc::Codeword: Matrix<EF>,
     M: Mmcs<EF>,
     Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F> + CanObserve<M::Commitment>,
-    R: Rng,
+    R: CryptoRng,
```
tcoratger, line 76: **[NIT] `CryptoRng` on the simulator is over-tight in principle, harmless in practice**

The simulator is the HVZK proof artifact: it produces a transcript *without* the witness, so its RNG guards nothing. Requiring `CryptoRng` on `simulate_classic_unpacked`, `simulate_classic_unpacked_claim` and the shared helper at line 68 is a bound with no security content, on three `pub` functions (`sumcheck/src/zk/mod.rs:154`).

I would keep it anyway, and I would say why in one line: the coupling tests (`simulator.rs:455`, `:652`, `:1297`) run the simulator on the *same seed* as the real prover to certify that the two transcripts are identically distributed. Once the real prover is `CryptoRng`, the simulator has to accept the same RNG type or the coupling cannot be set up. That is a real reason, and it is not obvious from the diff.

**Fix:** one comment above the bound, e.g. "matched with the real prover's bound so the coupling tests can drive both from one seeded `StdRng`; the simulator itself has no secrecy requirement."

whir/benches/sumcheck.rs
```diff
@@ -152,7 +152,7 @@ fn run_sumcheck<L: Layout<F, EF>>(prover: L, challenger: &mut Challenger, foldin
 fn run_zk_sumcheck<L>(
     prover: ZkProver<F, EF, MaskEnc, MaskMmcs, L>,
     challenger: &mut Challenger,
-    rng: &mut SmallRng,
+    rng: &mut StdRng,
```
tcoratger, line 165: **[NIT] This bench now measures ChaCha12 in the hot loop, so the series breaks with the swap**

Unlike `sumcheck/benches/sumcheck.rs:888`, where the RNG is built in criterion's untimed `setup` closure, here the `&mut StdRng` is threaded into `run_zk_sumcheck` and the mask draw happens inside the timed routine (`whir/benches/sumcheck.rs:161`, `prover.into_sumcheck(&mut data, 0, challenger, rng)`).

xoshiro256++ is roughly an order of magnitude cheaper per `u64` than ChaCha12, so `zk_sumcheck` numbers from this bench are not comparable to anything recorded before this PR. This is the *correct* thing to measure now — production draws masks from a CSPRNG — but a reader diffing bench output across the merge will see a regression that is not one.

**Fix:** a one-line comment at the `rng` parameter noting that mask sampling is inside the measurement and the generator changed, so pre-PR baselines do not transfer.

tcoratger (approved):
The bound is the right one and it lands on the entry points that matter. In `rand` 0.10.1 (workspace pin, `Cargo.toml:92`) the traits were re-layered: `Rng` is now the base object-safe trait (what 0.9 called `RngCore`), `RngExt: Rng` is the extension trait, and `rand_core-0.10.1/src/lib.rs:95` declares `pub trait CryptoRng: Rng + TryCryptoRng<Error = Infallible> {}`. `StdRng` carries `impl TryCryptoRng for StdRng` (`rand-0.10.1/src/rngs/std.rs:104`, ChaCha12), `SmallRng` (xoshiro256++) does not, so `R: CryptoRng` is exactly the bound that rejects the weak generator while still giving the callee every `RngExt` method through the blanket `impl<R: Rng + ?Sized> RngExt for R`.

**Answer to the question that matters: no production mask-sampling path still accepts a non-cryptographic RNG after this change.** I grepped every `Rng`/`CryptoRng` bound and every concrete RNG construction under `sumcheck/`, `whir/`, `zk-codes/`, `merkle-tree/`, `fri/`, `commit/`, `multi-stark/`, `uni-stark/`. Post-PR the full list of mask/blind draw sites is:
- `sumcheck/src/zk/prover/common.rs:70` `sample_masks` — now `CryptoRng`.
- `sumcheck/src/zk/prover/zk_prover.rs:186`, `sumcheck/src/zk/prover/residual.rs:82` — now `CryptoRng`.
- `whir/src/pcs/zk/prover/mod.rs:99,143` — now `CryptoRng`; `whir/src/pcs/zk/adapter.rs:49,89` already was.
- `fri/src/hiding_pcs.rs:118,208,301` and `merkle-tree/src/hiding_mmcs.rs:88,110` — already `CryptoRng + SeedableRng`.
- `whir/src/pcs/zk/base_case/prover.rs:83` — **still `R: Rng`**, and it is the site that draws the Construction 7.2 one-time pads (`fresh_message`, `fresh_randomness`, and every group blind at lines 105-133). It is not reachable with a weak RNG today because `base_case` is `pub(crate)` (`whir/src/pcs/zk/mod.rs:59`) and its only caller is `HidingWhirProver::prove` (`whir/src/pcs/zk/prover/mod.rs:490`), which this PR bounds to `CryptoRng`. See the inline note.

Every concrete `SmallRng` left in the tree that touches a ZK path is inside `#[cfg(test)]` (`whir/src/pcs/zk/committer.rs:195,232,252`, `whir/src/pcs/zk/code_switch.rs:313,353`, the `multi-stark` test modules); the one production-shaped example, `poseidon2-air/examples/prove_poseidon2_baby_bear_keccak_zk.rs:77,103`, already uses `StdRng`.

Two other things I checked and found clean. Mask draws are unbiased: `rng.random::<F>()` for Monty31 fields is rejection sampling (`monty-31/src/monty_31.rs:154-164`, loop until `next_u32() >> 1 < PRIME`), and `new_monty` is a bijection on the field, so the mask is uniform over `F` with no modular-reduction skew. And the determinism cost is zero: `StdRng::seed_from_u64` is still a deterministic ChaCha12 stream, so every test that needs the prover and the simulator on matched seeds keeps working — this PR gets the type-level guarantee without making a single test unreproducible, which is the right way to resolve that tension. Mask count / `ell_zk` is untouched by this PR, so the ZK budget itself is unchanged.

`cargo check -p p3-sumcheck -p p3-whir --all-targets` clean; `cargo test -p p3-sumcheck --lib zk` 57/57.

**Bottom line:** merge. No BLOCKER, no MAJOR; the findings below are defense-in-depth and polish.

tcoratger:
Pushed `be39ea69` to this branch (append-only — nothing rebased or amended), applying the review feedback. What went in:

- **`whir/src/pcs/zk/base_case/prover.rs:83`** — `R: Rng` is now `R: CryptoRng`, so the function that physically draws the Construction 7.2 one-time pads carries the requirement itself rather than inheriting it from caller discipline. The bound propagated cleanly: the only production caller was already `CryptoRng`, and the only thing that needed touching was the fixture generator in `base_case/tests.rs`, which was a `SmallRng`. Two tests there assert the exact spot-check position that trips first; those positions are an artifact of the seeded query-index sampler, so they moved with the generator (6 -> 9 and 9 -> 14) and I updated them along with a comment saying they carry no protocol meaning.

- **`zk-codes/src/encoding.rs`** — tightened `sample_message` and `sample_randomness` to `CryptoRng`, leaving `encode` and `simulate` on the looser bound as you suggested, since those two have legitimate deterministic uses. Downstream this only touched the Reed-Solomon impl, the sentinel mock in `residual.rs`, and the WHIR fixture already mentioned; no test needed a different generator beyond that one.

- **`sumcheck/src/zk/prover/zk_prover.rs`** — added a `compile_fail` doctest under the `# Randomness` section that names `SmallRng` as the generator for `into_sumcheck`. I did not use the snippet from the review verbatim: as written it only declares an unused generic function and so compiles either way. This version takes the path to the generic method with explicit type arguments, which forces the bound to be checked. Verified both directions — it passes as `compile_fail` today, and with the bounds flipped back to `Rng` through `sample_masks` and the encoding, it starts compiling and the doctest fails, which is exactly the signal wanted.

- **`sumcheck/src/zk/simulator.rs`** — recorded why the simulator carries a bound with no secrecy content of its own: it matches the real prover's so a single seeded generator can drive both sides of a coupling test.

- **`whir/benches/sumcheck.rs`** — noted that mask sampling sits inside the timed region, so the generator's own cost is part of the figure and numbers are only comparable across runs using the same one.

Verification: `cargo clippy -p p3-zk-codes -p p3-sumcheck -p p3-whir --all-targets` clean, `cargo +nightly fmt` clean, `cargo test -p p3-zk-codes` 21 passed, `cargo test -p p3-sumcheck --lib zk` 57 passed, `cargo test -p p3-sumcheck --doc` 1 passed, `cargo test -p p3-whir --lib zk` 95 passed.

tcoratger (approved):

The remaining loose RNG bounds at the sites that physically draw hiding material are now tightened, and the compile-fail doctest was verified in both directions — it passes as written and fails once the bounds are loosened again.

All review threads resolved.

PR #2261, opened 2026-09-20
fix(examples): reject a zero rate in pcs_comparison up front

Bruce039:
Running the PCS comparison with `--rate 0`:

```
cargo run -p p3-examples --example pcs_comparison -- --log-message-size 4 --rate 0
thread 'main' panicked at examples/examples/pcs_comparison.rs:460:14:
attempt to divide by zero
```

The FRI section derives `num_queries` as `(security_level - pow_bits).div_ceil(args.rate)`, so a zero rate dies there before STIR or WHIR get a chance to say `RateNotBelowOne`. Since `main` already asserts the other argument relations up front (`pow-bits < security-level`, `log-width <= log-message-size`, `log-message-size >= 3`), this adds `rate >= 1` to that list with a message that says what the flag means. Verified the message replaces the panic for the command above; clippy is clean.


Nashtare:
We don't accept such kind of contributions.

PR #2262, opened 2026-09-20
fix(goldilocks): reduce round constants in the NEON asm Poseidon2 layers

0xShadowX:
### What

On aarch64, `Poseidon2InternalLayerGoldilocksAsm::new_from_constants` and `Poseidon2ExternalLayerGoldilocksAsm::new_from_constants` copy each round constant's raw `value` into their tables. The rounds then add those through `add_canonical_asm`, which drops the canonicalize-`b` pair on the assumption that `b < P` (its doc comment lists the safe call sites: outputs of the other asm primitives, or constants that `Poseidon2GoldilocksFused::new` reduced with `to_canonical_u64`). The non-fused constructors never got that reduction.

### Why it matters

`Goldilocks::new` documents "any `u64` value is accepted, no reduction is performed", and `from_u64` is `Self::new(int)`. So a constant table built from raw 64-bit words, or from `F::from_u64(rng.random())` as `test_asm_matches_generic` does, is a valid input that lands in the non-canonical band with probability 2^-32 per constant. When it does:

* debug builds hit `add_canonical_asm precondition violated: b = ... >= P` inside `permute_mut`;
* release builds are fine through `Poseidon2GoldilocksAsm::permute_mut` (the initial external layer reduces the state before the internal rounds), but the internal layer used on its own adds the constant to the raw `s0`, and for `s0 ≥ P` the carry correction wraps a second time. `add_canonical_asm(u64::MAX, P + 1)` returns `0` where the field says `2^32 − 1`.

### Fix

Run the constants through the existing `to_canonical_u64` in both constructors, matching the fused type. No change for canonical constants.

### Test

`aarch64_neon::poseidon2::tests::test_asm_accepts_non_canonical_round_constants`: builds `Poseidon2GoldilocksAsm<8>` with `P + 1` and `u64::MAX` among the constants, compares 32 permutations of non-canonical states against the generic implementation, then does the same for the internal layer alone with `s0 = u64::MAX`. Without the change the test panics on the precondition in debug and fails the internal-layer comparison in release. `cargo nextest run -p p3-goldilocks` passes in both profiles (376 tests), clippy clean.


goldilocks/src/aarch64_neon/poseidon2.rs
```diff
@@ -35,7 +35,13 @@ pub struct Poseidon2InternalLayerGoldilocksAsm {
 
 impl InternalLayerConstructor<Goldilocks> for Poseidon2InternalLayerGoldilocksAsm {
     fn new_from_constants(internal_constants: Vec<Goldilocks>) -> Self {
-        let constants_raw = internal_constants.iter().map(|c| c.value).collect();
+        // `Goldilocks::new` and `from_u64` accept any `u64`, so a constant may arrive as
+        // `value >= P`. The internal round adds it through `add_canonical_asm`, whose
+        // reduction is only correct for a canonical `b`, so reduce once here.
+        let constants_raw = internal_constants
+            .iter()
+            .map(|c| to_canonical_u64(c.value))
```
tcoratger: **[MINOR] `to_canonical_u64` duplicates `PrimeField64::as_canonical_u64`**

`to_canonical_u64` (poseidon2.rs:317-321) is byte-for-byte the same one-conditional-subtraction as `Goldilocks::as_canonical_u64` (`goldilocks/src/goldilocks.rs:659-668`), down to the "one subtraction suffices because `2 * ORDER` does not fit in `u64`" argument.

Why it matters: two copies of a reduction rule is how the copies drift. The trait method is also the one a reader already trusts, so `c.as_canonical_u64()` documents the intent without a comment. The only thing the local helper buys is `const fn`, which neither of the three call sites (41-44, 109-113, 114-118) needs — they are all runtime iterators.

```rust
.map(|c| to_canonical_u64(c.value))   // -> .map(Goldilocks::as_canonical_u64)
```

**Fix:** drop `to_canonical_u64` and use `p3_field::PrimeField64::as_canonical_u64` at all five sites (including the two in `Poseidon2GoldilocksFused::new`). If you prefer to keep the `const fn`, at least add `// == PrimeField64::as_canonical_u64` above it so a future reduction change touches both.
**Verify:** existing `test_asm_matches_generic_width_{8,12,16}` plus the new test already cover it; a swap is behaviour-preserving by inspection.

goldilocks/src/aarch64_neon/poseidon2.rs
```diff
@@ -98,15 +104,17 @@ impl<const WIDTH: usize> ExternalLayerConstructor<Goldilocks, WIDTH>
     for Poseidon2ExternalLayerGoldilocksAsm<WIDTH>
 {
     fn new_from_constants(external_constants: ExternalLayerConstants<Goldilocks, WIDTH>) -> Self {
+        // Same reduction as the internal layer: the external rounds add these through
+        // `add_canonical_asm` as well.
         let initial_constants_raw = external_constants
             .get_initial_constants()
             .iter()
-            .map(|rc| core::array::from_fn(|i| rc[i].value))
+            .map(|rc| core::array::from_fn(|i| to_canonical_u64(rc[i].value)))
```
tcoratger: **[MINOR] The comment overstates which external rounds are affected**

"the external rounds add these through `add_canonical_asm` as well" is only true for `WIDTH = 8`. The generic `external_round_asm` (poseidon2_asm.rs:823) and `external_round_dual` (:879) use `add_asm`, which canonicalizes `b` itself with the leading `subs/csel` pair. Widths 12, 16 and 20 were never broken.

Why it matters: this constructor is generic over `WIDTH`, so a reader tracing the invariant from here will conclude all widths shared the bug and may go looking for it in the wrong place. Reducing unconditionally here is still the right thing — it is free at construction time and future-proofs a W12/W16 `add_canonical_asm` specialization — but the comment should say that is why, not claim a bug that is not there.

**Fix:** reword to something like: "Only the `WIDTH = 8` fused rounds use `add_canonical_asm` for the RC today (`external_round_fused_w8`); the generic rounds use `add_asm`. Reduce unconditionally so any future specialization inherits the invariant."
**Verify:** `grep -n 'add_canonical_asm\|add_asm(state' goldilocks/src/aarch64_neon/poseidon2_asm.rs` — lines 823 and 879 are `add_asm`, 896-912 and 955-1007 are `add_canonical_asm`.

goldilocks/src/aarch64_neon/poseidon2.rs
```diff
@@ -634,6 +642,71 @@ mod tests {
...
+            &mut rng,
+        );
+        let mut internal_constants: Vec<Goldilocks> = (0..GOLDILOCKS_POSEIDON2_PARTIAL_ROUNDS_8)
+            .map(|_| rng.random())
+            .collect();
+        // Both equal `1`, `u64::MAX` is `P + 2^32 - 2`, i.e. `2^32 - 2` as a field element.
+        internal_constants[0] = Goldilocks::new(P + 1);
+        internal_constants[5] = Goldilocks::new(u64::MAX);
+        let mut initial = external_constants.get_initial_constants().to_vec();
+        initial[0][3] = Goldilocks::new(u64::MAX);
```
tcoratger: **[MINOR] The external-layer non-canonical constant does not actually discriminate**

`initial[0][3] = Goldilocks::new(u64::MAX)` is placed after `mds_light_permutation_asm`, which leaves every state word canonical (`apply_mat4_asm` returns `add_canonical_asm` outputs, so `< P`). And with `a < P`, `add_canonical_asm` is exact for **any** `b < 2^64` — the two cases:

```text
  carry   : S = a + P + d - 2^64 = a + d - EPSILON,  +EPSILON -> a + d      OK
  no carry: S = a + P + d >= P,  subs P            -> a + d                 OK
```

so line 663 passes with or without the fix. What genuinely fails is the pair where **both** operands are non-canonical, which on this PR's code only arises via the standalone `InternalLayer::permute_state` at lines 690-707 — the second half of your test, and the one that earns its keep.

Why it matters: an assertion that cannot fail is worse than no assertion, because it makes the test look like it covers the external layer when it does not. The comment on line 659 ("`u64::MAX` is `P + 2^32 - 2`") is exactly right and worth keeping; it is the conclusion drawn from it that slips.

**Fix:** keep line 663 (it is a cheap smoke test) but note in the comment that it only bites in debug via the `debug_assert!`; then move the real external-layer coverage to a terminal-round constant, where the state entering `external_round_fused_w8` is *also* only canonical because the preceding round made it so. Alternatively state plainly that the release-mode discriminator is the internal-layer block below.
**Verify:** revert `goldilocks/src/aarch64_neon/poseidon2.rs:109-118` to `|rc| rc[i].value` and run `cargo test --release -p p3-goldilocks test_asm_accepts_non_canonical_round_constants` on aarch64 — I expect it still to pass with only the external constant unreduced.

goldilocks/src/aarch64_neon/poseidon2.rs
```diff
@@ -634,6 +642,71 @@ mod tests {
...
+                *v = Goldilocks::new(u64::MAX - seed * 131 - i as u64);
+            }
+            let mut generic_state = state;
+            asm.permute_mut(&mut state);
+            generic.permute_mut(&mut generic_state);
+            assert_eq!(state, generic_state, "state seed {seed}");
+        }
+
+        // The internal layer on its own, fed a non-canonical state: the raw `s0` add is the
+        // one place where an unreduced constant would change the field element.
```
tcoratger: **[NIT] Good comment, one word too narrow**

"the raw `s0` add is the one place where an unreduced constant would change the field element" is right, and it is worth recording *why* it is the one place: every other entry point runs `mds_light_permutation_asm` first, which canonicalizes the whole state, and a canonical `a` makes `add_canonical_asm` exact for any `b`. Spelling that out turns this from a note about one call site into the invariant that keeps the rest of the file safe.

**Fix:** append "...because every other entry point runs `mds_light_permutation_asm` first, which leaves the state canonical, and `add_canonical_asm` is exact whenever `a < P`."
**Verify:** n/a, comment only.

tcoratger (approved):
Canonicalizes the Poseidon2 round constants in the two non-fused NEON asm constructors so they satisfy `add_canonical_asm`'s `b < P` precondition, matching what `Poseidon2GoldilocksFused::new` already did. I verified the arithmetic by hand and the diagnosis holds; the fix is constructor-only, costs nothing on the hot path, and is correct and complete for every reachable `add_canonical_asm` call site. Approving — findings are 3 MINOR + 1 NIT, none blocking.

**What I checked, concretely.** `add_canonical_asm(a, b)` computes `adds; csetm w-reg, cs; add; subs; csel`. The `:w` modifier on `csetm` is the trick that makes it work: a 32-bit `-1` zero-extends to `0x0000_0000_FFFF_FFFF = EPSILON = 2^64 - P`, so the carry branch folds `2^64 -> EPSILON` correctly.

Working the two cases for `b = P + d`, `0 <= d <= 2^32 - 2` (the whole non-canonical band, since `u64::MAX - P = 2^32 - 2`):

```text
  a canonical (a < P):
    carry  : S = a + P + d - 2^64 = a + d - EPSILON,  result = a + d          OK
    no carry (a + d < EPSILON): S = a + P + d >= P,   result = a + d after subs OK
  a non-canonical too:  a = 2^64-1, b = P+1
    a + b = 2^65 - 2^32 + 1  ->  S = P (carry)
    S + EPSILON = P + 2^32 - 1 = 2^64  ->  wraps to 0 (plain `add`, no second carry fold)
    subs 0 - P borrows, so result = 0.   Field answer is (2^32-2) + 1 = 2^32-1.  WRONG
```

So the bug needs **both** operands non-canonical, which sharpens the PR description: `add_canonical_asm` is in fact exact whenever `a < P`, and `mds_light_permutation_asm` canonicalizes the state on entry (every `apply_mat4_asm` output comes out of `add_canonical_asm`, hence `< P`). That is exactly why the release-mode failure only shows up when `InternalLayer::permute_state` is called standalone, as the PR says.

I also confirmed the blast radius is **WIDTH = 8 only**. Only three sites feed an RC into `add_canonical_asm`: `internal_permute_state_asm_w8` (poseidon2_asm.rs:235), `external_round_fused_w8` (:896-912) and `external_round_fused_w8_dual` (:955-1007). Widths 12/16/20 use `add_asm` for the RC (:147, :363, :530, :823, :879), which canonicalizes `b` itself. Picking `WIDTH = 8` for the new test is therefore the right call.

Other things I verified and found clean: `to_canonical_u64`'s single subtraction is sufficient (`u64::MAX < 2P`); `poseidon1.rs:56-84` also stores raw `.value` but `poseidon1_asm.rs` never calls `add_canonical_asm`, so it is unaffected; `apply_mat4_asm` and `mds_light_permutation_asm` keep their `b`-canonical invariants; `div2_asm` is representative-correct on non-canonical input (`x = P + d` odd gives `(x-1)/2 + (P+1)/2 = P + d/2`); `Default` leaves the tables empty, which is harmless. `cargo check --target aarch64-unknown-linux-gnu -p p3-goldilocks --lib` passes here (I am on x86_64, so I could not execute the NEON tests — see the note below).

**[MINOR] Stale safe-call-site list on `add_canonical_asm`** — `goldilocks/src/aarch64_neon/utils.rs:176-178`. The doc still says the only RC call site is "an RC stored on `Poseidon2GoldilocksFused`, canonicalized at construction time by `to_canonical_u64`". After this PR three types do that. That stale list is precisely the artifact that let the bug in — the next person adding an asm layer reads it and concludes their constructor is not on the list. Please extend it to `Poseidon2InternalLayerGoldilocksAsm` and `Poseidon2ExternalLayerGoldilocksAsm`, and consider adding the sharper statement above: the precondition is only load-bearing when `a` is also non-canonical.

**[MINOR] Test gap: the packed W8 path is not exercised with non-canonical constants** — `goldilocks/src/aarch64_neon/poseidon2_asm.rs:955-1007`. `external_round_fused_w8_dual` is the third `add_canonical_asm`-with-RC site and the new test only drives the scalar `Permutation<[Goldilocks; 8]>`. The fix covers it (same table), but nothing pins it. One extra block in `test_asm_accepts_non_canonical_round_constants` closes it:

```rust
let a: [F; WIDTH] = core::array::from_fn(|i| Goldilocks::new(u64::MAX - i as u64));
let b: [F; WIDTH] = core::array::from_fn(|i| Goldilocks::new(P + i as u64));
let mut packed: [PackedGoldilocksNeon; WIDTH] =
    core::array::from_fn(|i| PackedGoldilocksNeon([a[i], b[i]]));
asm.permute_mut(&mut packed);
let (mut ea, mut eb) = (a, b);
generic.permute_mut(&mut ea);
generic.permute_mut(&mut eb);
for i in 0..WIDTH {
    assert_eq!(packed[i].0[0], ea[i]);
    assert_eq!(packed[i].0[1], eb[i]);
}
```

**[NIT] Verification I could not do.** Host is x86_64, so I read and cross-compiled the NEON code but did not run it. The hand analysis above is my evidence for the `add_canonical_asm(u64::MAX, P + 1) == 0` claim; if you have an aarch64 box, a one-line `assert_eq!(unsafe { add_canonical_asm(u64::MAX, P + 1) }, 0)` next to `test_add_canonical_asm_edge_pairs` (utils.rs:351) would turn the prose in the PR body into a regression test for the primitive itself, independent of Poseidon2.

**Bottom line:** merge.

tcoratger:
Pushed a follow-up commit (`f30764d7`) to this branch applying the review findings directly — feel free to amend, squash or drop it.

- [MINOR] dropped the local `to_canonical_u64` helper; all five sites now use `PrimeField64::as_canonical_u64`.
- [MINOR] reworded the external-layer comment: only the `WIDTH = 8` fused rounds use the reduced-operand addition, the generic rounds reduce `b` themselves, so the unconditional reduction is future-proofing rather than a bug fix.
- [MINOR] the `initial[0][3]` constant is now labelled as a debug-only smoke test, and the comment points at the standalone internal-layer block as the release-mode discriminator.
- [NIT] widened the internal-layer comment with the invariant that makes it the only place that matters (`mds_light_permutation_asm` leaves the state canonical, and the addition is exact whenever `a < P`).
- [MINOR, from the review body] refreshed the stale safe-call-site list on `add_canonical_asm`, and added the sharper statement about the precondition only biting for non-canonical `a`.
- [MINOR, from the review body] added the packed dual-lane block to `test_asm_accepts_non_canonical_round_constants`, so `external_round_fused_w8_dual` is now pinned too.

Comments were rewritten to the repo doc conventions (one sentence per line, no intra-doc links).

**Not done, and why:** the NIT suggesting `assert_eq!(unsafe { add_canonical_asm(u64::MAX, P + 1) }, 0)` next to `test_add_canonical_asm_edge_pairs`. `add_canonical_asm` opens with `debug_assert!(b < P)`, so such a test can only run under `--release`, where CI does not run the unit tests — it would be a permanently dormant assertion. I also have no aarch64 host to confirm the value.

**Verification caveat:** I am on x86_64 with no aarch64 runner or emulator, so the NEON tests were type-checked only (`cargo check --target aarch64-unknown-linux-gnu -p p3-goldilocks --lib --tests` is clean, as is clippy on the lib). Someone on an aarch64 box should run `cargo test -p p3-goldilocks` before merge.

tcoratger (approved):

0xShadowX:
Ran the updated head (`f30764d7`) on an Apple M2 (aarch64, `neon` + `aes` on by default), rustc 1.98.1:

```
cargo nextest run -p p3-goldilocks            376 tests run: 376 passed
cargo nextest run --release -p p3-goldilocks  376 tests run: 376 passed
```

so `test_asm_accepts_non_canonical_round_constants`, including the packed dual-lane block you added, passes in both profiles. I also checked the primitive value from the review directly in a release build: `add_canonical_asm(u64::MAX, P + 1)` returns `0` (field answer is `2^32 - 1`), matching your derivation. Happy with the follow-up commit as is.

PR #2263, opened 2026-09-20
fix(matrix): evaluate coset interpolation at zero instead of inverting it

0xShadowX:
### What

`Interpolate::interpolate_coset` (and therefore `interpolate_subgroup`) computes `f(z)` from coset evaluations through the adjusted-weight form `z · (z^N − g^N)/(N g^N) · Σ (1/(z − x_i) − 1/z) f(x_i)`, and `compute_adjusted_weights` starts with `point.inverse()`. For `z = 0` that is `Field::inverse` on zero and the call panics with `Tried to invert zero`, although `f(0)` is perfectly well defined (it is the constant coefficient) and zero is never on a coset with non-zero shift, so the on-domain shortcut does not catch it.

### Change

At `z = 0` every Lagrange basis polynomial of the coset `gH` evaluates to `1/N` (`L_i(0) = (−g^N) x_i / (N g^N (−x_i))`), so `interpolate_coset` now returns the column mean for that point, computed with the same column-wise dot product the main path uses. `compute_adjusted_weights` and `interpolate_coset_with_precomputation` keep the adjusted form and now say that the point must be non-zero; the in-tree callers of that path (`p3-fri`, `p3-stir`, `p3-commit`) only ever pass Fiat–Shamir challenges, so they are unaffected either way.

### Test

`interpolation::tests::test_interpolate_coset_at_zero`: `3 + 2x + 5x² + 7x³` evaluated over the shifted 8-point coset and over the subgroup, interpolated at `0` in the base field and in a degree-4 extension, returns `3`. Panics on `main`. `cargo nextest run -p p3-matrix`: 189 passed, clippy clean.


matrix/src/interpolation.rs
```diff
@@ -113,6 +119,16 @@ pub trait Interpolate<F: TwoAdicField>: Matrix<F> {
...
         }
 
+        // At z = 0 every Lagrange basis polynomial of the coset takes the value 1/N, so the
+        // evaluation is the column mean. The adjusted-weight form below factors 1/z out of
+        // each weight and cannot express this point.
+        if point.is_zero() {
+            let ones = vec![EF::ONE; self.height()];
+            let mut evals = self.columnwise_dot_product(&ones);
+            scale_slice_in_place_single_core(&mut evals, EF::ONE.div_2exp_u64(log_height as u64));
+            return evals;
```
tcoratger: **[NIT] `vec![EF::ONE; height]` buys an O(N) allocation and N multiplications by one**

`columnwise_dot_product` is the right primitive to reuse, but the weight vector here is all-ones, so every one of the `height * width` packed multiplies has a known-`ONE` operand the compiler cannot see through (`EF` is generic). For a `log_height = 20`, degree-4 extension matrix that is a 32 MiB temporary plus `2^20 * width` redundant extension multiplies, to compute what is structurally a column sum.

This path is cold — `z = 0` is a `2^{-|EF|}` event for a Fiat-Shamir challenge — so it is not worth contorting the code for. But `p3-matrix` has no `columnwise_sum`, and this is the second place that would want one (`columnwise_dot_product_batched` at `matrix/src/lib.rs:502` is the general form). Worth a one-line comment saying the allocation is deliberate because the branch is cold, so the next reader does not assume it was an oversight.

```rust
let ones = vec![EF::ONE; self.height()];
let mut evals = self.columnwise_dot_product(&ones);
```

**Fix:** add `// Cold path (z = 0 is a 2^-|EF| event), so reuse the SIMD dot product rather than add a columnwise-sum primitive.` — or, if you want it free, `par_chunks(width)`-fold the rows and skip both the vector and the multiplies.
**Verify:** no behavioural change either way; `test_interpolate_coset_at_zero` covers it.

matrix/src/interpolation.rs
```diff
@@ -64,6 +65,11 @@ use crate::dense::RowMajorMatrix;
...
 ///
 /// One extension-field inversion + N parallel extension-field subtractions.
+///
+/// # Panics
+///
+/// Panics if `point` is zero: the adjusted form factors `1/z` out of every weight, so it
+/// does not exist there. [`Interpolate::interpolate_coset`] handles that point itself.
 pub fn compute_adjusted_weights<EF: Field>(point: EF, diff_invs: &[EF]) -> Vec<EF> {
     // Single inversion of z, amortised over all N weights.
     let point_inv = point.inverse();
```
tcoratger, line 83: **[MINOR] Documented precondition, unchecked — and the sibling fails silently rather than loudly**

The new `# Panics` block is accurate for `compute_adjusted_weights`: `point.inverse()` does panic at zero. But it panics with `Tried to invert zero`, which names the symptom in `p3-field` rather than the precondition in `p3-matrix`, so a caller landing on it has to reverse-engineer the barycentric form to understand what they did wrong.

The worse half is `interpolate_coset_with_precomputation`: with `point = EF::ZERO` its scaling factor is `point * (z_pow_n - g_pow_n) * denom_inv = 0`, so it returns an **all-zero vector** and does not panic at all. A documented-but-unchecked precondition that returns plausible-looking wrong data is the shape of bug that survives a test suite — a verifier would accept `f(z) = 0` for every column rather than reject.

```rust
let point_inv = point.inverse();   // "Tried to invert zero"
...
let scaling_factor = point * (z_pow_n - g_pow_n) * denom_inv;   // = 0 at z = 0, silently
```

**Fix:** `assert!(!point.is_zero(), "compute_adjusted_weights: the adjusted form factors out 1/z; use Interpolate::interpolate_coset for z = 0");` above the inversion, and a matching `debug_assert!(!point.is_zero(), ...)` next to the existing `debug_assert_eq!(adjusted_weights.len(), self.height())` in `interpolate_coset_with_precomputation` (line ~183). The `assert!` is free relative to the N-element parallel map that follows; the `debug_assert!` keeps the zero-allocation hot path untouched in release.
**Verify:** `#[should_panic(expected = "adjusted form")]` on `compute_adjusted_weights(EF4::ZERO, &[EF4::ONE])`.

matrix/src/interpolation.rs
```diff
@@ -578,6 +594,26 @@ mod tests {
...
+        assert_eq!(m.interpolate_coset(shift, F::ZERO), vec![coeffs[0]]);
+        assert_eq!(
+            m.interpolate_coset(shift, EF4::ZERO),
+            vec![EF4::from(coeffs[0])]
+        );
+
+        let on_subgroup: Vec<F> = eval_poly_on_coset(&coeffs, F::ONE, 3);
+        let m = RowMajorMatrix::new(on_subgroup, 1);
+        assert_eq!(m.interpolate_subgroup(F::ZERO), vec![coeffs[0]]);
+    }
```
tcoratger, line 660: **[MINOR] Test gaps: single column, single height, one fixed polynomial**

The fixture pins the identity on exactly one `(coeffs, shift, log_height, width)` point. Three specific cases the new branch can reach and this does not:

1. **`width > 1`.** `columnwise_dot_product` + `scale_slice_in_place_single_core` operate per column; a transposition or stride bug in that pair is invisible at `width = 1`. Every other caller (`commit/src/domain.rs:419` opens `k` columns at once) is wide.
2. **`log_height = 0`.** Height 1 makes `div_2exp_u64(0)` a no-op and the coset a single point. It used to panic and now must return the row verbatim. It is the boundary most likely to be broken by a future refactor of the scale factor.
3. **A non-`GENERATOR` shift.** `L_i(0) = 1/N` is independent of `g` — that independence is the non-obvious part of the claim and nothing tests it.

A proptest is the right tool here: the space is `(log_height, shift, coeffs, width)` and the oracle is a two-line Horner evaluation at 0, i.e. `coeffs[0]`. That is far cheaper than a table of known-answer vectors and it proves the shift-independence directly — a failing case would immediately tell you whether the `1/N` factor or the column layout is wrong.

```rust
proptest! {
    #[test]
    fn prop_interpolate_coset_at_zero_is_constant_term(
        log_h in 0usize..6, width in 1usize..5, seed in any::<u64>(),
    ) {
        let n = 1 << log_h;
        // width independent polys, each of degree < n
        let coeffs: Vec<Vec<F>> = (0..width).map(|_| (0..n).map(|_| rng.random()).collect()).collect();
        let shift: F = loop { let s: F = rng.random(); if !s.is_zero() { break s } };
        // row-major evaluations over shift * H
        let mut rows = Vec::with_capacity(n * width);
        for x in TwoAdicMultiplicativeCoset::new(shift, log_h).unwrap().iter() {
            for c in &coeffs { rows.push(horner(c, x)); }
        }
        let got = RowMajorMatrix::new(rows, width).interpolate_coset(shift, EF4::ZERO);
        for (j, c) in coeffs.iter().enumerate() {
            prop_assert_eq!(got[j], EF4::from(c[0]));
        }
    }
}
```

**Fix:** add the proptest above (it subsumes cases 1-3), or at minimum extend the existing fixture with a `width = 3` matrix and a `log_height = 0` assertion.
**Verify:** on `main` the proptest fails at the first sample with `Tried to invert zero`; with this PR it must pass for every `log_h` including 0.

tcoratger (approved):
Adds a `z = 0` branch to `Interpolate::interpolate_coset` returning the column mean, replacing a `Tried to invert zero` panic from `compute_adjusted_weights`. I verified the identity by hand and it is exactly right, for every shift and every log-size including the degenerate ones. Approving — 2 MINOR + 1 NIT, none blocking.

**The math, verified.** From the module header, `L_i(z) = x_i * (z^N - g^N) / (N * g^N * (z - x_i))` with `x_i = g h^i`. At `z = 0`:

```text
  L_i(0) = x_i * (0 - g^N) / (N * g^N * (0 - x_i))
         = x_i * (-g^N)   / (N * g^N * (-x_i))
         = 1/N                        <- independent of i AND of the shift g
```

Cross-check from the other direction, which is the one that convinced me it holds for *every* input rather than just the test vector. Write `f = sum_k c_k X^k` with `deg f < N` (always true: N evaluations determine a unique such `f`). Then

```text
  sum_i f(x_i) = sum_k c_k g^k sum_i h^{i k}
               = sum_k c_k g^k * (N if k = 0 mod N else 0)
               = N * c_0        since 0 <= k < N
```

so the mean is `c_0 = f(0)`. The geometric sum kills every `k != 0` because `h` is a primitive `N`-th root, and no `k` in range is a nonzero multiple of `N`. That is the whole proof and it needs no case analysis.

**Degenerate cases I walked through:**

- `log_height = 0` (height 1): coset is `{shift}`, `div_2exp_u64(0)` is the identity, the branch returns the single row — correct, `f` is constant. Previously this panicked.
- `shift = 0`: unreachable. `TwoAdicMultiplicativeCoset::new` (`field/src/coset.rs`) returns `None` unless `shift != F::ZERO`, and `interpolate_coset` unwraps it.
- `z` equal to a domain element: the on-coset scan at line 117-120 runs *before* the new branch, so it still wins. And `0` can never be a coset element (`g h^i != 0` for `g != 0`), so the two branches can never both fire.
- degree 0 / constant column: mean of N copies of `c` is `c`.
- characteristic 2: `div_2exp_u64` would divide by zero, but a char-2 field has `TWO_ADICITY = 0`, so `log_height > 0` is already rejected by the coset constructor and `log_height = 0` makes the division a no-op. Safe by construction.

Ordering, allocation lifetime and the `use alloc::vec;` import are all fine; `cargo test -p p3-matrix --lib interpolat` passes here (42 tests) and clippy is clean.

**[MINOR] The fix does not reach the call sites that actually matter.** `p3-fri` (`fri/src/two_adic_pcs.rs:582`) and `p3-stir` (`stir/src/pcs.rs:943`) call `compute_adjusted_weights` directly and then `interpolate_coset_with_precomputation`; neither goes through `interpolate_coset`. So after this PR, `m.interpolate_coset(shift, 0)` returns the right answer while the production path on the identical input still panics inside `Field::inverse`. `p3-commit` (`commit/src/domain.rs:374, :419`) does benefit.

That asymmetry is defensible — `z` is a Fiat-Shamir challenge, so hitting `0` costs `2^{-|EF|}` — but it is worth saying in the doc, because the reason the batched path cannot be fixed the same way is structural: the adjusted weights `1/(z - x_i) - 1/z` are shared across many matrices at one point, and at `z = 0` the shared scalar `z * (z^N - g^N) / (N g^N)` collapses to `0`, so there is no weight vector that represents this point. A caller who wants `z = 0` must take the `interpolate_coset` branch. One sentence pointing at `Interpolate::interpolate_coset` in the `# Correctness requirements` block would save the next reader the derivation.

**Bottom line:** merge.

tcoratger:
Pushed a follow-up commit (`56e194be`) to this branch applying the review findings directly — feel free to amend, squash or drop it.

- [MINOR] `compute_adjusted_weights` now asserts the non-zero precondition itself instead of surfacing `Tried to invert zero` from the field layer, and `interpolate_coset_with_precomputation` gained a matching `debug_assert!` — it previously returned an all-zero vector at `z = 0` without complaining.
- [MINOR] extended the fixture with a `log_height = 0`, `width = 2` case, and added a proptest over `(log_n, width, shift, coeffs)` that checks the result is each column's constant term. It covers width > 1, height 1 and arbitrary (non-`GENERATOR`) shifts. On `main` it fails on the first sample with `Tried to invert zero`.
- [NIT] documented that the all-ones weight vector is deliberate because the branch is cold.
- [MINOR, from the review body] added a sentence to `# Correctness requirements` explaining why the batched path structurally cannot represent `z = 0`.

Verified: the new `should_panic` test and the proptest both fail without the corresponding production change. `cargo test -p p3-matrix --lib interpolat` passes (44 tests), clippy clean, `p3-fri`/`p3-stir`/`p3-commit` still check.

Comments were rewritten to the repo doc conventions (one sentence per line, no intra-doc links).

tcoratger (approved):

PR #2264, opened 2026-09-20
fix(multilinear-util): state and check the real precondition of fix_prefix_var_to_packed

0xShadowX:
### What

`Poly::fix_prefix_var_to_packed` documents a single panic case ("the polynomial is constant") and asserts exactly that. The body then does `A::Packing::pack_slice(evals)` on the *unfolded* table and splits the packed array in half, which only works when the folded half is at least one lane group, i.e. `num_variables() > log2(WIDTH)`. On this NEON host (`WIDTH = 4`) that leaves two undocumented failures for valid-looking inputs:

* `k < log2(WIDTH)`: `pack_slice` panics with `Slice length (got 2) must be a multiple of packed field width (4)`.
* `k == log2(WIDTH)`: `split_at(0)` gives an empty first half, the zip produces nothing and `Poly::new(vec![])` panics with `Evaluation list length must be a power of two.`

The existing tests already skip these sizes ("Need k >= k_pack + 1 so the output holds >= 1 packed entry"), so the requirement was known but never written down. `Poly::pack` a few lines below states and asserts its own lane-group bound, so this brings the two into line.

### Change

Replace the "no free variables" assert with `num_variables() > log2(A::Packing::WIDTH)` and a message naming the bound, and fix the doc comment. In-tree callers (`p3-multi-stark`'s round fold and fractional GKR) already stay above the bound, so no behaviour changes for them.

### Test

`poly::test::test_fix_prefix_var_to_packed_rejects_lane_group_sized_input` (`k = log2(WIDTH)`, expects the new message). `cargo nextest run -p p3-multilinear-util -p p3-multi-stark -p p3-binary-pcs`: 670 passed; clippy clean.


multilinear-util/src/poly/mod.rs
```diff
@@ -687,18 +687,26 @@ where
...
     ///
-    /// The result has one fewer variable (n - 1).
+    /// The result has one fewer variable (n - 1), stored as 2^{n - 1 - log_2(W)} packed
+    /// elements, so it needs `n > log_2(W)`: the folded half must fill at least one lane group.
+    /// Smaller polynomials fold with [`fix_prefix_var`](Self::fix_prefix_var) and stay scalar.
     ///
     /// # Panics
     ///
-    /// Panics if the polynomial is constant (zero free variables).
+    /// Panics if the polynomial has at most log_2(W) variables.
```
tcoratger: **[NIT] `# Panics` drops the information the body just added**

Line 690-692 explains the bound well — including *why* it is `log_2(W)` and what to use instead. Then `# Panics` says "at most log_2(W) variables" with `W` undefined in that sentence, and the escape hatch is gone. Rustdoc renders `# Panics` as its own section, so a reader who jumps there from the sidebar sees the restatement without the explanation.

Why it matters here specifically: the whole point of the PR is that an undocumented precondition cost someone a confusing `pack_slice` panic. The panic section is the first place that person looks.

```rust
/// Panics if the polynomial has at most log_2(W) variables.
```

**Fix:** `/// Panics if the polynomial has at most `log_2(A::Packing::WIDTH)` variables, i.e. if the folded half would not fill one SIMD lane group. Use [`fix_prefix_var`](Self::fix_prefix_var) below that size.`
**Verify:** `cargo doc -p p3-multilinear-util` and read the rendered section.

multilinear-util/src/poly/mod.rs
```diff
@@ -687,18 +687,26 @@ where
...
     {
         let evals = self.as_slice();
-        assert!(evals.len() > 1, "no free variables");
+        // Below this the packed halves are empty or `pack_slice` sees a partial lane group.
+        assert!(
+            self.num_variables() > log2_strict_usize(A::Packing::WIDTH),
+            "fix_prefix_var_to_packed needs more than log2(WIDTH) = {} variables, got {}",
+            log2_strict_usize(A::Packing::WIDTH),
+            self.num_variables()
+        );
```
tcoratger, line 718: **[MINOR] `evals` is now bound before the assert that justifies it, and the bound is computed twice**

Two small things in one place:

1. `let evals = self.as_slice();` on line 702 runs before the check, so the ordering reads as "take the slice, then find out whether we were allowed to". Moving the assert above line 702 puts the precondition first, which is what the rest of the file does (`fix_suffix_var` at line ~735 asserts before touching `as_slice`). Free, and it makes the invariant the first thing a reader meets.

2. `log2_strict_usize(A::Packing::WIDTH)` appears twice (705 and 707). The format arguments are lazily evaluated so the second call costs nothing on the happy path — this is purely about not making the reader check that the two expressions are the same one. A `let` also lets the message name the actual width:

```rust
let log_w = log2_strict_usize(A::Packing::WIDTH);
assert!(
    self.num_variables() > log_w,
    "fix_prefix_var_to_packed needs more than log2(WIDTH) = {log_w} variables, got {}",
    self.num_variables()
);
let evals = self.as_slice();
```

**Fix:** hoist the assert above line 702 and bind `log_w` once, as above.
**Verify:** `test_fix_prefix_var_to_packed_rejects_lane_group_sized_input` still matches on the `expected` prefix, which is unchanged.

multilinear-util/src/poly/mod.rs
```diff
@@ -2168,6 +2176,16 @@ pub(crate) mod test {
...
 
+    #[test]
+    #[should_panic(expected = "fix_prefix_var_to_packed needs more than log2(WIDTH)")]
+    fn test_fix_prefix_var_to_packed_rejects_lane_group_sized_input() {
+        // Exactly one lane group: the folded half would hold no packed element at all.
+        let mut rng = SmallRng::seed_from_u64(0);
+        let k_pack = log2_strict_usize(PackedF::WIDTH);
+        let poly = Poly::<F>::rand(&mut rng, k_pack);
+        let _ = poly.fix_prefix_var_to_packed::<EF>(EF::ONE);
+    }
```
tcoratger, line 2198: **[MINOR] The test collapses into an existing one on a scalar host, and misses the `n < log2(W)` mode**

Two coverage observations:

1. On a target without SIMD (`PackedF::WIDTH == 1`), `k_pack = 0` and this builds a constant polynomial — byte-for-byte the same fixture as `test_fix_prefix_var_to_packed_panics_on_constant` at line 2236. Both still pass, but the suite silently stops testing the new bound on exactly the configuration where a future change to the `WIDTH == 1` packing could regress it. Not a bug, just something to know when reading a green CI matrix.

2. The PR body identifies **two** distinct failure modes with **two** distinct panic messages, and this test pins only one. `k < k_pack` panics inside `pack_slice` ("Slice length ... must be a multiple of packed field width") — a different call frame and a different diagnostic, which is precisely the confusing message the PR exists to replace. A test that the new assert fires *first* on that input is what proves the improvement, not just that it fires at the boundary.

```rust
#[test]
#[should_panic(expected = "fix_prefix_var_to_packed needs more than log2(WIDTH)")]
fn test_fix_prefix_var_to_packed_rejects_partial_lane_group() {
    let k_pack = log2_strict_usize(PackedF::WIDTH);
    if k_pack == 0 { panic!("fix_prefix_var_to_packed needs more than log2(WIDTH) = 0") } // scalar host: vacuous
    let mut rng = SmallRng::seed_from_u64(0);
    let poly = Poly::<F>::rand(&mut rng, k_pack - 1);
    let _ = poly.fix_prefix_var_to_packed::<EF>(EF::ONE);
}
```

(The `k_pack == 0` guard is ugly; `#[cfg(any(target_feature = "avx2", target_feature = "neon"))]` or an early `return` from a non-`should_panic` test that asserts via `catch_unwind` are cleaner if you prefer.)

**Fix:** add the `k = k_pack - 1` case above, so both documented failure modes are pinned to the new message.
**Verify:** on `main` that input panics with "Slice length (got 2) must be a multiple of packed field width (4)"; with this PR it must panic with the new message instead.

tcoratger (approved):
Replaces the `evals.len() > 1` assert in `fix_prefix_var_to_packed` with `num_variables() > log2(W)`, the bound the body has always needed, and fixes the doc to match. I derived the precondition independently and it is exactly right — not conservative, not loose. Approving — 1 MINOR + 1 NIT, none blocking.

**Derivation.** Let `n = num_variables()`, `W = A::Packing::WIDTH`, so `evals.len() = 2^n`. Two constraints stack:

```text
  pack_slice(evals)     needs  2^n mod W == 0        ->  n >= log2(W)
  split_at(len/2) then
  Poly::new(collected)  needs  (2^n / W) / 2 >= 1    ->  n >= log2(W) + 1
```

The second dominates, so the exact precondition is `n > log2(W)` — matching the assert character for character. The two failure modes in the PR body are the two constraints: `n < log2(W)` trips `pack_slice`, `n == log2(W)` gives `split_at(0)`, an empty `collect`, and `Poly::new(vec![])` panicking on "Evaluation list length must be a power of two".

One thing the description does not say but that I checked, because it is the part that could have made the bound wrong: **splitting the packed array in half is the same map as splitting the scalar array in half.** Packed slot `j` covers scalar indices `[jW, (j+1)W)`, so the packed midpoint `(2^n/W)/2 = 2^{n-1}/W` lands on scalar index `2^{n-1}` exactly — but only if `2^{n-1}` is a multiple of `W`, i.e. `n - 1 >= log2(W)`. Same condition again. If the bound had been off by one the fold would have silently mixed the `x_0 = 0` and `x_0 = 1` halves inside a lane group rather than panicking, which is the failure worth ruling out.

```text
  n = log2(W) + 1, W = 4:   evals = [e0 e1 e2 e3 | e4 e5 e6 e7]
  packed:                   [ P0=(e0..e3) , P1=(e4..e7) ]
  split_at(1):              p0 = [P0]          p1 = [P1]
                             ^ scalar [0,4)      ^ scalar [4,8)   <- halves align
```

**Callers verified, both safe with margin:**

- `multi-stark/src/rounds.rs:1889` is guarded by `want_packed = (half / 2) >= F::Packing::WIDTH` at `rounds.rs:1942`, i.e. `2^n >= 4W`, i.e. `n >= log2(W) + 2`. One step stricter than the assert.
- `multi-stark/src/fractional_gkr/prover.rs:392-393` folds `n0`/`n1`, each half of a `2^m` table, so it needs `m >= log2(W) + 2`. The guard at `prover.rs:851` unpacks to scalar storage whenever `denom.num_variables() < 3`, and the packed denominator counts slots, so the packed path survives only for `m >= log2(W) + 3`. Two steps of margin.

So no in-tree caller violates the new bound, and none is close enough that this assert turns a working prove into a panic.

Also checked: `test_fix_prefix_var_to_packed_panics_on_constant` (line 2236) uses a bare `#[should_panic]` with no `expected`, so the message change does not break it. `cargo test -p p3-multilinear-util --lib fix_prefix_var_to_packed` passes here, 6 tests.

**[MINOR] The sibling branch in the same caller has the identical unguarded assumption** — `multi-stark/src/rounds.rs:1892-1901`. When `column.as_dense()` returns `None`, `fold` hand-rolls the packed fold:

```rust
let half = column.len() / 2;
Poly::new((0..half).step_by(F::Packing::WIDTH)
    .map(|start| EF::ExtensionPacking::from_ext_fn(|lane| {
        let lo = column.value(start + lane);
        let hi = column.value(start + half + lane);
```

This needs the same `half >= WIDTH` (or it reads `start + lane >= half`, crossing into the other half's data) and the same non-empty result. Today `want_packed` at `rounds.rs:1942` covers both branches, so it is correct — but the dense branch now documents and asserts its bound while the sparse one states nothing. Since this PR's whole thesis is "write the requirement down where it is relied on", a one-line `debug_assert!(half >= F::Packing::WIDTH)` there would finish the job. Out of scope for this diff; worth a follow-up.

**Bottom line:** merge.

tcoratger:
Pushed a follow-up commit (`20bbbbe5`) to this branch applying the review findings directly — feel free to amend, squash or drop it.

- [MINOR] hoisted the assert above `as_slice` and bound `log2_strict_usize(A::Packing::WIDTH)` once, so the precondition is the first thing the body does and the bound appears once.
- [MINOR] added `test_fix_prefix_var_to_packed_rejects_partial_lane_group` for the `n < log2(W)` mode, gated on a SIMD target since a scalar host has no smaller size to test. Without the new assert it panics with `Slice length (got 8) must be a multiple of packed field width (16)`, which is exactly the diagnostic this PR replaces.
- [NIT] rewrote the `# Panics` section so it carries the reason and the alternative, not just a restatement.

Verified: both `should_panic` tests fail against the old `evals.len() > 1` assert. `cargo test -p p3-multilinear-util --lib fix_prefix_var_to_packed` passes (7 tests with `-Ctarget-cpu=native`, 6 on a scalar build), clippy clean.

Comments were rewritten to the repo doc conventions (one sentence per line, no intra-doc links), so the `# Panics` wording differs slightly from the suggested text.

**Not done:** the `debug_assert!(half >= F::Packing::WIDTH)` in the sparse branch of `multi-stark/src/rounds.rs` — the review flagged it as out of scope for this diff, so I left it for a follow-up.

tcoratger (approved):

PR #2266, opened 2026-09-20
perf(multi-stark): share one zeroed successor buffer across a stage's workers

Nashtare:
Stacked on #2255.

`PackedScratch::new` allocated and zeroed a `next_point` and a `next_diff` for every Rayon worker.
A stage whose AIR reads no successor row never touches either of them — for BLAKE3 that is 1.48 MB
of zeroing per worker, roughly thirty-two times a round, and `__memset_avx512_unaligned_erms` shows
up in the profile as a result.

The read sites index `next_point[slot.main_offset .. +slot.main_width]` unconditionally and hand
the slice to the AIR folder, so a width-long buffer has to exist even when nothing reads it — a
zero-length `Vec` panics on the slice. A private `NextRows<P>` is therefore either `Filled(Vec<P>)`
or `Shared(Arc<Vec<P>>)`, with `Deref<Target = [P]>` for reads so every read site is unchanged, and
`fill() -> &mut [P]` for writes. Each fold entry point builds one `Arc` of width-long zeros when
`next_row_runs(&self.slots)` is empty, and every worker of that stage clones it for both buffers.

A stage that does read a successor row takes the old per-worker path, bit for bit. `fill()`'s
`unreachable!()` is rigorously unreachable: `fill_next` is only ever called with
`slot.main_next_columns` / `slot.preprocessed_next_columns`, whose concatenation is exactly the
`next_row_runs(&self.slots)` that selects `Shared`.

### Measurements

`CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, default PCS flags, medians of 3 with base and branch
interleaved per rep. Two independent hosts for BLAKE3, since the first was reclaimed mid-run;
absolute speeds differ between them, the deltas do not.

| | prove | `prove_with_lookup` |
| --- | --- | --- |
| BLAKE3 `-l 18`, host A | 2.111 -> 2.080 s (-1.5%) | 1.37 -> 1.35 s (-1.5%) |
| BLAKE3 `-l 18`, host B | 2.402 -> 2.359 s (-1.8%) | 1.56 -> 1.51 s (-3.2%) |
| Keccak-f `-l 18` | 0.752 -> 0.745 s | 592 -> 593 ms |

Every branch rep beat every base rep on both hosts.

Keccak is flat, and it is flat **by construction** rather than empirically: its AIR declares
successor columns, so `next_row_runs` is non-empty, no `Arc` is built, and every worker takes the
`Filled` path. The successor path is untouched.

### Correctness

Prover-only and byte-identical: FNV digests of the postcard proof match at `-l 12` and `-l 14` for
both objectives, with identical proof sizes. `cargo test -p p3-multi-stark --features parallel`
passes 358 tests including all fifteen `backends_agree_*`, which compare whole serialized proofs
between `ReprBackend` and `GenericBackend`; the full workspace suite passes. Clippy and rustfmt
clean.

Stack frames: 1330 `bridge_producer_consumer::helper` instantiations in both binaries, of which
eleven differ, all increases, at most +48 B and +288 B program-wide. It is not a size effect: `NextRows<P>`
measures 24 B, the same as the `Vec<P>` it replaces, because `Vec`'s capacity niche carries the
enum tag whichever arm the other variant holds. `#[inline(never)]` is retained on both node-walk kernels and
they remain out-of-line.

### Adjacent, measured but not done

`fold_boundary` in `rounds/sliced.rs` does `R::zero_vec(half)` per column inside a `par_iter` and
then fully overwrites it through `as_chunks_mut::<SLICED_LANES>()`, so the zeroing is waste
whenever `half % SLICED_LANES == 0`. That is the 0.304%-of-cycles, 100%-memset entry in the
profile, and unlike this change it would fire on Keccak too. It needs its own byte-identity check
and A/B. `RowTile::new`'s extra lane is only dead when `next_columns` is empty and bounds out
around 0.006% of cycles — not worth a diff.


multi-stark/src/rounds.rs
```diff
@@ -556,6 +557,60 @@ pub(crate) fn rows_per_task(rows: usize) -> usize {
...
+    /// row holds those, and every fill runs inside a walk of that stage's successor runs, which
+    /// is then empty.
+    #[inline]
+    fn fill(&mut self) -> &mut [P] {
+        match self {
+            Self::Filled(rows) => rows,
+            Self::Shared(_) => unreachable!("a stage reading no successor row fills no row"),
+        }
+    }
+}
```
tcoratger: **[MINOR] A six-call-site invariant guarded by a runtime panic, where copy-on-write would make it structurally impossible**

The invariant holds today and I verified it both by hand and by instrumentation (see the review body). But it is *derived*, not enforced: it lives entirely in the agreement between `next_row_runs(&self.slots)` at six fold entry points across three files and the run lists the `fill_next` closures are separately handed. The doc comment above is the only place that agreement is written down.

The failure mode if it ever breaks is a `panic!` inside a rayon worker, in release, under whatever AIR shape the tests happen not to cover. `unreachable!()` over `unreachable_unchecked()` is the right call -- a panic, never UB -- but there is an option that removes the failure mode entirely at no hot-path cost:

```rust
fn fill(&mut self) -> &mut [P] {
    if let Self::Shared(zeros) = self {
        // The shared buffer is all zeros, so this rebuilds exactly the
        // per-worker `P::zero_vec(width)` the pre-PR code allocated.
        let owned = zeros.as_slice().to_vec();   // borrow of `zeros` ends here
        *self = Self::Filled(owned);
    }
    let Self::Filled(rows) = self else { unreachable!("replaced one line above") };
    rows
}
```

The `Filled` arm is untouched, so the predicted-taken branch costs the same. The difference is where the `unreachable!()` gets its justification: today from a cross-module argument about six call sites, after the change from the statement immediately above it. And the fallback is not a fallback into wrongness -- it reconstructs bit for bit the buffer every worker used before this PR, so a future refactor that breaks the coupling silently loses the optimisation instead of panicking. Needs `P: Clone` on the impl block, which every `P` instantiated here already has.

**Fix:** the copy-on-write above; or keep `unreachable!()` and add, at the top of each `fill_next` closure, `debug_assert!(runs.is_empty() || matches!(scratch.next_point, NextRows::Filled(_)))`, so a debug run localises a future breakage to the call site rather than to `fill()`.
**Verify:** the invariant is already test-enforced -- 166,872 `Shared` constructions across `cargo test -p p3-multi-stark --features parallel --release` with the `unreachable!()` live, none fired. A regression test would have to construct an `AirSlot` whose `main_next_columns` disagrees with `next_row_runs`, which is currently unconstructible; that is the point.

multi-stark/src/rounds.rs
```diff
@@ -556,6 +557,60 @@ pub(crate) fn rows_per_task(rows: usize) -> usize {
...
+///
+/// A stage whose AIRs read no successor row never writes them, so they hold zeros for the whole
+/// fold and one allocation serves every worker. A stage that does read a successor row gives
+/// each worker buffers of its own to fill.
+enum NextRows<P> {
+    /// Buffers this worker fills, one entry per column of the stage.
+    Filled(Vec<P>),
+    /// Zeros every worker of the stage reads.
+    Shared(Arc<Vec<P>>),
+}
```
tcoratger, line 570: **[MINOR] `Arc<[P]>` is the same 32 bytes as `Arc<Vec<P>>` here and saves a pointer chase on every read**

Reads of a successor buffer now go through `Deref`, and the two arms are not symmetric:

```text
  Filled:  NextRows --> Vec<P> --> [P]                     2 hops
  Shared:  NextRows --> Arc<Vec<P>> --> Vec<P> --> [P]     3 hops
```

`Arc<[P]>` collapses the `Shared` arm to two hops, because a fat `Arc` addresses the elements directly instead of a `Vec` header that addresses them in turn. And it is free in space: `Vec<P>` is 24 bytes and already sets the enum payload, so `Arc<[P]>`'s 16-byte fat pointer fits inside the layout `Arc<Vec<P>>` forces anyway -- 32 bytes either way, which is exactly the `NextRows` (32 B) vs `Vec` (24 B) delta the PR body already accounts for in the frame-size analysis.

This is worth doing because the `Shared` arm is the hot one for this change. The read sites index `next_point[slot.main_offset .. +slot.main_width]` once per slot, per interpolation node, per row, inside `walk_row_nodes` / `walk_lane_nodes` -- the same kernels the PR is trying to speed up. The extra load sits on the dependency chain into the AIR folder.

**Fix:** `Shared(Arc<[P]>)`, built as `Arc::from(<F::Packing>::zero_vec(width))` (and the corresponding `R` / `PackedRepr` / `PackedExt` forms) at the six entry points. `NextRows::new` and the `Deref` arm are unchanged; `zeros.clone()` still just bumps the refcount.
**Verify:** re-run the BLAKE3 `-l 18` A/B. If the pointer chase is on the critical path the delta widens; if it is not, the change is still free in space and one indirection shorter. `Keccak-f` must stay flat, since it never builds an `Arc` at all.

tcoratger (approved):
Replaces the per-worker `next_point`/`next_diff` allocations with one `Arc<Vec<P>>` of zeros shared across a stage's workers, taken only when `next_row_runs(&self.slots)` is empty. I went after the thing that makes this shape dangerous, and it is not dangerous here: the shared buffer is never mutable, it is reachable only through `Deref<Target = [P]>`, every write goes through `fill()`, and `fill()` refuses the `Shared` arm. No `unsafe`, no interior mutability, no byte two workers can both touch. There is also nothing to re-zero, because a `Shared` buffer is allocated fresh per fold entry point and is never written for its whole lifetime. Safe to merge as-is; the two MINORs below are hardening and a read-path micro-optimisation.

## What I verified

**Ownership and the handoff.** The buffer never becomes shared *and* mutable:

```text
  round_poly_*()                         one Arc per fold call
        |
        | next_columns.is_empty() ? Some(Arc::new(zero_vec(width))) : None
        v
  par_fold_reduce(init = || Scratch::new(.., next_zeros.as_ref()))
        |
   +----+----+----+   ... one init call per rayon split
   v         v    v
 worker    worker  worker
   |         |      |
   +---------+------+---> same Arc, refcount only
                          reads:  Deref -> &[P]        (shared, fine)
                          writes: fill() -> &mut [P]   (refused on Shared)
```

`NextRows` has no `DerefMut`, no `get_mut`, no raw pointer. `Arc<Vec<P>>` hands out `&Vec<P>` and nothing more, so the only way to reach `&mut [P]` is `fill()`, which is the `Filled` arm only. This is why the "shared zeroed buffer" framing is safe: it is a shared *immutable* buffer that happens to be the neutral element every reader expects.

**Re-zeroing is a non-question, and I checked why.** A `Shared` buffer is created inside the fold entry point, lives for that one `par_fold_reduce`, and is never written by anybody — so it cannot carry anything from a previous stage or a previous round. A `Filled` buffer is per-worker, `zero_vec(width)` at construction, and is overwritten on every row for exactly the columns inside `next_columns`; columns outside those runs stay zero for the buffer's lifetime, which is the pre-PR behaviour verbatim. There is no reset path because no state survives to need one.

**The `unreachable!()` is unreachable, structurally and empirically.** `next_row_runs` (`multi-stark/src/rounds.rs:1096`) is exactly a `flat_map` over every slot's `main_next_columns` chained with `preprocessed_next_columns`, so it is empty iff every slot's two vectors are empty. Every `fill()` call sits inside a `for run in <runs>` loop whose `<runs>` is one of those vectors or the concatenation itself. All six entry points agree on the source:

| entry point | `next_zeros` computed from | consumers walk |
| --- | --- | --- |
| `round_poly_packed` (`rounds.rs:1595`) | `next_row_runs(&self.slots)` | `slot.main_next_columns`, `slot.preprocessed_next_columns`, `&next_columns` |
| `round_poly_scalar` (`rounds.rs:1828`) | same | same |
| `round_poly_generic` (`rounds.rs:2226`) | same | `&next_columns` |
| `round_poly_ext_packed` (`rounds.rs:2530`) | same | `&next_columns` |
| `round_poly_repr_lanes` (`repr.rs:379`) | `round.next_columns`, itself `next_row_runs(&self.slots)` at `repr.rs:437` | `&round.next_columns` |
| `boundary_evals_rows` / `_lanes` (`sliced.rs:1496`, `sliced.rs:1577`) | same | same |

`rounds/subfield.rs` keeps its own plain-`Vec` `Scratch` and is untouched.

Empirically: I instrumented both arms of `NextRows::new` and ran `cargo test -p p3-multi-stark --features parallel --release`. The `Shared` arm is constructed **166,872** times and the `Filled` arm **264,428** times, and the `unreachable!()` never fires. Both new paths are genuinely reached by the existing suite, and the suite doubles as a runtime proof of the invariant. (358 tests pass, 0 fail.)

**Sharing cannot make the round polynomial depend on rayon's split.** The `par_fold_reduce` reduce closures merge only `constraint_evals` / `interaction_evals`, with `add_slices` / `add_slice`; `next_point` and `next_diff` never enter the reduction, and the surviving `lhs` keeps its own. Field addition is exactly associative and commutative, so the fold is split-independent. Unchanged by this PR, but worth pinning since the diff rewrites the init closure at six sites.

**No prime-field measurement, so I took one.** The PR measures BLAKE3 and Keccak. Poseidon2 over BabyBear is the interesting third case: it declares no successor columns, so it takes the new `Shared` arm, but with a far narrower `width` than BLAKE3 — meaning the `Arc` refcount traffic per `par_fold_reduce` init is amortised over much less work, which is where an atomic-contention regression would show. `multi-stark/benches/poseidon2.rs`, `ZEROCHECK_POSEIDON2_NUM_VARS=13`, `1_airs`, `--features parallel`, `-C target-cpu=native`, base `4f8819f4` vs branch, interleaved:

| | base | branch |
| --- | ---: | ---: |
| rep 1 | 10.581 ms | 10.310 ms |
| rep 2 | 11.832 ms | 11.369 ms |
| rep 3 | 11.481 ms | 10.546 ms |

Noisy host, but the branch is never slower. No contention effect visible. Worth one line in the PR body, since "a stage that reads no successor row" now covers every prime-field AIR of that shape too.

**Bottom line:** merge.

multi-stark/src/rounds.rs
tcoratger: Still open at `f97c9195`. The delta since my approval is a rebase, and it did not touch `fill()`: at the new head all 18 call sites still sit inside a `for run in <runs>` loop, and the seven `next_zeros` predicates still derive from the same `next_row_runs(&self.slots)` / `round.next_columns` those loops walk. So the invariant is intact and the finding is unchanged -- copy-on-write is still what would retire it.

multi-stark/src/rounds.rs
tcoratger: Still open at `f97c9195` -- `Shared(Arc<Vec<P>>)` is unchanged (`rounds.rs:569`), so the `Shared` arm still costs the extra hop.

tcoratger (approved):
Re-review against the new head `f97c9195`. **The approval stands.**

## What the delta contains

Nothing but a rebase. My approval was at `757699775`; the branch was force-pushed three times and now sits on `perf/zc-gather-prefetch` at `80fd960a` instead of `4f8819f4`, i.e. 12 new base commits, several of them in the very files this PR touches (`5b75134f`, `523afe2a`, `cf5c23ec`, `80fd960a` in `rounds/sliced.rs` and `rounds/repr.rs`).

Comparing this PR's own patch (`base..head`) at the two heads, the added/removed line set is identical except for one line, and that one is a rebase artifact rather than a change: the PR no longer has to add `PrimeCharacteristicRing` to `sliced.rs`'s `p3_field` import, because the new base already imports it. Everything else that moved is hunk offsets and context.

The one thing worth checking was the interaction with the base's successor-row rewrite, since `523afe2a` ("read a lane group's upper next row in one load") rewrites `RowTile::read_lane_group` -- the same function this PR converts to `fill()`. It is benign: the base changed only how `lo`/`hi` are computed (a new `lane_pair` helper replacing two `lane_group(...)` closures), not the `for run in next_columns` loop or its guard.

- `NextRows` is still `Filled(Vec<P>)` / `Shared(Arc<Vec<P>>)`, with `Deref<Target = [P]>` and no `DerefMut`, no `get_mut`, no raw pointer, no `unsafe`. The `Shared` buffer is still never mutable.
- Every write still funnels through `fill()`, and all 18 `fill()` call sites still sit inside a `for run in <runs>` loop.
- The seven `next_zeros` predicates (`rounds.rs:1597/1830/2228/2532`, `sliced.rs:1525/1605`, `repr.rs:396`) all still derive from the same `next_row_runs(&self.slots)` / `round.next_columns` value the consumers walk, so the `unreachable!()` is unreachable for the same structural reason as before.
- `rounds/subfield.rs` and the `SlicedGf4` path still keep plain `Vec`s and are untouched.

`cargo check -p p3-multi-stark --features parallel --all-targets` and `cargo test -p p3-multi-stark --features parallel --release` both pass on the new head.

## Still open

Neither MINOR was addressed; both threads stay open (replies posted on each). Both are hardening rather than defects, so they do not hold up the merge:

- the copy-on-write `fill()` that would make the six-call-site invariant structural instead of a runtime panic;
- `Shared(Arc<[P]>)` instead of `Arc<Vec<P>>`, one pointer chase shorter at the same 32 bytes.

**Bottom line:** merge.

multi-stark/src/rounds.rs
Nashtare: Done — `fill()` now copies the shared zeros into buffers this worker owns instead of treating them as an unreachable case (`multi-stark/src/rounds.rs:585-601`, commit `refactor(multi-stark): copy the shared successor zeros before a fill`; the SHA will move when the stack is restacked).

The `Filled` arm is untouched. The `unreachable!()` that remains is discharged by the assignment one line above rather than by the agreement between six call sites and `next_row_runs`, and both paths through the `if let` store the discriminant the second match reads, so LLVM folds it away instead of emitting the panic. It needed `P: Clone` on the impl block, which every `P` instantiated here already satisfies.

I also added `filling_shared_successor_rows_copies_them_first`: it fills one `NextRows` built from a shared `Arc` and asserts that a second holder and the `Arc`'s own buffer are still all zeros. That puts the copy-on-write on the tested path even though, as you note, no production call site reaches it.

multi-stark/src/rounds.rs
Nashtare: Done — `Shared(Arc<[P]>)`, built with `Arc::from(<..>::zero_vec(width))` at all seven entry points (`rounds.rs:1602,1835,2233,2537`, `rounds/sliced.rs:1527,1608`, `rounds/repr.rs:399`). `NextRows::new` and the `Deref` arm are unchanged, and `zeros.clone()` still only bumps the refcount.

One correction to the space argument, which I measured rather than assumed. `NextRows<P>` is **24 bytes both before and after**, not 32. `Vec`'s capacity field carries a validity niche, so rustc puts the tag there and the 24-byte `Vec` payload sets the size whether the other arm is an 8-byte `Arc<Vec<P>>` or a 16-byte `Arc<[P]>`. Checked with `const _: [(); 32] = [(); core::mem::size_of::<NextRows<u64>>()]` against both enum shapes, and against `NextRows<[u64; 8]>` so the payload alignment is covered too — all four report 24.

That makes your conclusion stronger, not weaker: the change is free in space, and `NextRows` is in fact the same size as the bare `Vec` it replaced. It does mean the description's frame-size line explains the measured +48 B / +288 B with a `NextRows` (32 B) against `Vec` (24 B) delta that does not exist. The measurement stands; the explanation does not, and I will correct that line.

PR #2267, opened 2026-09-20
perf(multi-stark): weight a batch of AIR constraints with one dot product

Nashtare:
Stacked on #2266.

### What the AIR evaluation actually costs

Per-instruction sample attribution inside `walk_lane_nodes` (the BLAKE3 instantiation, 7.72% of all
program cycles at `-l 18`) splits it as 47.3% XOR/`vpternlog`, 20.1% `vpclmulqdq`, 12.1% shuffle,
6.5% vector load, 6.2% vector store, 4.3% scalar.

Cross-referenced against the AIR's static structure — 11,536 constraints per row, 56 G steps of 190
constraints each — one `PackedGhash128` multiply is 5 `vpclmulqdq` + 2 `vpshufd` + 2 `vpslldq` +
8 xor, so 382 multiplies per G step account for 6,494 of its ~7,770 vector instructions. **The
evaluation is about 84% field multiplication, and half of those multiplies are the alpha weighting
rather than the constraints**: 192 constraint multiplies against 190 alpha-power multiplies, one
per `assert_zero`. The XOR share is mostly the xors *inside* Karatsuba, not builder overhead.

### The change

`MultilinearFolder` accumulates a batch of `ALPHA_BATCH = 8` asserted constraints and weights the
whole batch with one `Algebra::mixed_dot_product`, which a ring with a linear reduction answers by
accumulating products unreduced and reducing once — three carryless multiplies per term instead of
five. The partial batch an evaluation ends inside is drained a term at a time.

`PackedExt` forwards `dot_product` to the packing it wraps through a `repr(transparent)` array
cast. **This is not optional**: without it the wrapper's default sum-of-products hides
`PackedGhash128`'s deferred-reduction kernel and the batching becomes a regression (+1.23% BLAKE3,
+2.41% Keccak). The two files have to move together.

Separately, `blake3-air`'s G schedule permutes the sixteen message-word *positions* instead of the
words, replacing a copy of the whole block — 32 KB per AIR evaluation at `PackedRepr` width, plus
six more inside `permute` — with a copy of sixteen indices. It is its own commit and is trivially
droppable.

The prime-field paths are unaffected: `mixed_dot_product`'s default is exactly the old per-term
product, so only rings that override `dot_product` change behaviour.

### Measurements

`CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, `-l 18`, medians of 3, base and branch alternated.

| | `round_poly_repr` | prove |
| --- | --- | --- |
| BLAKE3 | 355.4 -> 333.0 ms (-6.3%) | **-2.98%** |
| Keccak-f | 127.5 -> 118.5 ms (-7.0%) | **-2.10%** |

The blake3-air commit cannot touch Keccak, so Keccak's -7.0% is the batching alone.

`perf` counters, BLAKE3: instructions 177.96 G -> 178.18 G (**+0.13%**), cycles 124.95 G -> 121.29 G
(**-2.93%**), IPC 1.424 -> 1.469, elapsed 2.757 -> 2.664 s. **Instructions go up while cycles come
down** — the change removes two `vpclmulqdq` and two shuffles per constraint, which are port-5-only
work, and pays for it with a stack-slot round trip and index arithmetic on the load/store ports.
The win is port pressure and dependency chains, not uop count. Memory-level parallelism is flat
(3.354 -> 3.317), so the neighbouring gather did not widen.

### Correctness

Byte-identical: FNV digests of the postcard proof match for both objectives at `-l 12`, `-l 14` and
`-l 16`, with identical proof sizes. Both `mixed_dot_product` and `PackedGhash128::dot_product` are
exact — reduction is `F2`-linear — so reassociating the sum cannot change a value.
`cargo test --workspace --features parallel --release` passes 144 suites including the fourteen
`backends_agree_*` whole-proof comparisons; clippy and rustfmt clean. New tests cover the batching
against Horner across several batches and a tail, and the `PackedExt` forwarding against the sum of
products.

### Stack frames — they get smaller

The previously quoted frame figures for these kernels were wrong: `frames.sh` used `strtonum`, a
gawk extension that mawk silently evaluates to `0`. Measured properly (prologue reservation
including the stack-probe target), BLAKE3's `walk_lane_nodes` reserves **125,896 B** and Keccak's
**284,136 B**, so `#[inline(never)]` is considerably more load-bearing than the record suggested.

Every frame Rayon's recursive splitter can reserve gets smaller here: the maximum
`bridge_producer_consumer::helper` frame drops **35,144 -> 19,160 B** across 1,432 instantiations.
The deepest frame in the program is unchanged in size (+384 B) but moves one call deeper — for
Keccak, LLVM now outlines `evaluate_air_families` out of `walk_lane_nodes` rather than inlining it,
because the folder struct grew by the 512-byte pending buffer. That is one call per node per slot,
transient stack rather than per-split-level, so the overflow hazard is strictly reduced.

### A measured finding deliberately not landed

A schoolbook packed multiply instead of Karatsuba in `PackedGhash128` is a genuine ~2% win on
Sapphire Rapids — 6 clmul + 2 shuffle + 5 xor against 5 clmul + 4 shuffle + 8 xor. It is **not
landable**: the code comment it would replace records a Zen 5 measurement of 0.47 ns against
0.55 ns favouring Karatsuba by 15%, and no `cfg` distinguishes the two microarchitectures. Trading
2% on Intel for 15% on AMD is not a trade worth making. Recorded here rather than applied.

Useful by-product of that probe: `vpclmulqdq` on zmm costs roughly 3.5x a `vpxorq` in throughput on
Sapphire Rapids, which is why carryless-multiply *count* rather than instruction count is the metric
for this kernel. By that metric eight per constraint — five constraint, three alpha — is the floor
for this structure, and this change reaches it from ten.


multi-stark/src/folder.rs
```diff
@@ -514,16 +546,27 @@ where
...
+                    if slot + 1 == ALPHA_BATCH {
+                        let window = powers[index + 1 - ALPHA_BATCH..]
+                            .first_chunk::<ALPHA_BATCH>()
+                            .expect("the powers reach the end of a batch that holds a constraint");
+                        self.accumulator += <Acc as Algebra<Var>>::mixed_dot_product::<ALPHA_BATCH>(
+                            window,
+                            &self.pending,
+                        );
+                    }
                 }
```
tcoratger, line 590: **[VERIFIED] The batch kernel, checked term by term -- no change requested, recording the argument**

Anchoring the derivation here because this is the span where a one-off would become a silent soundness break rather than a failed build.

```text
  assert_zero call #i          pending slot        power
  ---------------------------  ------------------  -------------------
  i = qB + 0                   pending[0]     <-   (held)
  i = qB + 1                   pending[1]     <-   (held)
   ...                          ...
  i = qB + B-1   -- FLUSH -->  pending[0..B]  x    powers[qB .. qB+B]
                               acc += dot(window, pending)
```

- `slot = i % B` and `i` increments by exactly one per assert, so slots fill `0..B-1` in order and at the flush all `B` slots belong to batch `q`. Higher-numbered slots left over from batch `q-1` are never read, because the flush reads all `B` and every one was just rewritten.
- `powers[i+1-B..].first_chunk::<B>()` selects `powers[qB ..= qB+B-1]`, so slot `s` meets `powers[qB+s]`, which is the weight `C_{qB+s}` had under the old `powers.get(constraint_index)`. Same constraint, same power, same declaration order.
- The `expect` is provably dead: `i % B == B-1` gives `i >= B-1` so `i+1-B >= 0`, and the enclosing `i < powers.len()` gives exactly `powers.len() - (i+1-B) >= B` remaining elements. Never `None`.
- Moving `constraint_index = index + 1` above the bounds check is equivalent to the old increment-after, since nothing between them reads it.

The value is unchanged because the only thing that moved is the association of the sum, and `p3-field`'s addition is exactly associative in every ring reachable here. That is what makes this safe to do at all: in characteristic 2 with an `F2`-linear reduction, `fold(sum_i p_i) = sum_i fold(p_i)` is an identity, not an approximation.

multi-stark/src/folder.rs
```diff
@@ -626,6 +670,7 @@ where
         // Disabled constraints never reach the inner folder, so they consume no power.
         if self.constraints_enabled {
             self.inner.assert_alpha_power_count();
+            self.inner.drain_pending();
         }
         // Both families come out of the one pass, batched independently.
         FolderEvaluations {
```
tcoratger: **[MINOR] Two exit paths must both remember to drain, and forgetting is silent**

`MultilinearFolder::into_accumulator` does `assert_alpha_power_count(); drain_pending();` and then reads `accumulator`. This site re-implements that sequence by hand and then reads `self.inner.accumulator` directly. The sequence is now a convention rather than a type-level guarantee.

The failure mode if a third consumer appears and skips the drain: up to `ALPHA_BATCH - 1` constraints of the final partial batch are dropped from the accumulator. `assert_alpha_power_count` still passes, because `constraint_index` was incremented for every one of them -- the count check is exactly the wrong detector here, since the constraints *were* counted, just never added. Nothing downstream notices; the prover emits a round polynomial that is short by a few weighted constraint values, and the proof simply fails to verify, far from the cause.

**Fix:** collapse the two into one private method and make it the only way out.

```rust
/// Weight the trailing partial batch and return the finished accumulator.
#[inline]
fn finish(&mut self) -> Acc {
    self.assert_alpha_power_count();
    self.drain_pending();
    self.accumulator
}
```

`into_accumulator` becomes `self.finish()`, and this site becomes `constraints: if self.constraints_enabled { self.inner.finish() } else { self.inner.accumulator }`. `drain_pending` and `assert_alpha_power_count` can then both stop being callable from outside `finish`.

**Verify:** the existing `alpha_batching_matches_horner_across_several_batches` covers the drain through `into_accumulator`; add the same comparison through `InteractionMultilinearFolder::eval_air` with `constraints_enabled = true` and a constraint count that is not a multiple of `ALPHA_BATCH`, so this second path is pinned too. Today nothing distinguishes the two exits.

multi-stark/src/folder.rs
```diff
@@ -253,11 +268,28 @@ where
...
+        let Some(powers) = self.alpha_powers else {
+            return;
+        };
+        // Constraints past the last power were never held, so they leave nothing to drain.
+        let filled = self.constraint_index.min(powers.len());
+        let start = filled - filled % ALPHA_BATCH;
+        for (slot, &power) in powers[start..filled].iter().enumerate() {
+            self.accumulator += power * core::mem::replace(&mut self.pending[slot], Var::ZERO);
+        }
+    }
```
tcoratger, line 312: **[MINOR] The `min` and the comment above it describe a state that cannot reach this function**

`assert_alpha_power_count()` is a hard `assert!` (not `debug_assert!`), and it runs immediately before `drain_pending` at both call sites. By the time control arrives here, `self.constraint_index == powers.len()` unconditionally -- a surplus or a shortfall has already panicked one line up. So `filled` is always `powers.len()`, the `min` never clamps, and the comment "Constraints past the last power were never held, so they leave nothing to drain" describes a path that is unreachable rather than handled.

That matters for a reader, not for the machine: the comment makes it look as though surplus constraints are tolerated and absorbed here, when in fact they are rejected upstream. Someone hardening this later could reasonably conclude the assert is redundant and relax it, at which point this function silently becomes load-bearing for a case nobody has tested.

**Fix:** either state the dependency,

```rust
// `assert_alpha_power_count` ran first, so the counts already agree.
debug_assert_eq!(self.constraint_index, powers.len());
let start = self.constraint_index - self.constraint_index % ALPHA_BATCH;
```

or keep the `min` and change the comment to say it only matters if that assert is ever downgraded. The `debug_assert_eq!` is the better of the two: it pins a cross-function invariant that is currently invisible from inside `drain_pending`.

**Verify:** no behavioural test needed; the `debug_assert_eq!` fires immediately in the existing suite if the ordering is ever inverted.

multi-stark/src/folder.rs
```diff
@@ -192,6 +206,7 @@ where
             // No precomputed powers until attached; batching folds by Horner.
             alpha_powers: None,
             constraint_index: 0,
+            pending: [Var::ZERO; ALPHA_BATCH],
```
tcoratger: **[MINOR] This zero-init is provably dead, and LLVM cannot see it**

`drain_pending` only reads slots `0..(filled - start)`, and every one of those was written by an `assert_zero` before the drain; the flush path reads all `B` slots and every one was just written too. So no read of `pending` ever observes this initial value. But the read range depends on the runtime `constraint_index`, so LLVM has no way to prove that and the stores stay.

The size is not negligible on the packed instantiations. For `MultilinearFolder<'a, F, F::Packing, EF::ExtensionPacking>` (the prime-field packed path, `multi-stark/src/rounds.rs:1757`) `Var = F::Packing` is 32 B under AVX2 and 64 B under AVX-512, so this is 256 or 512 bytes of zero stores per `MultilinearFolder::new` -- and `new` runs once per (row block, interpolation node, AIR slot). For BLAKE3 with 11,536 constraints per row it amortises to nothing. For a narrow AIR that asserts a handful of constraints per evaluation it does not: the folder pays a fixed 512-byte memset to hold at most five values.

I measured the case I expected to be worst (BabyBear Poseidon2 zerocheck) and found no regression, so this is an observation rather than a defect -- but it is the cost the change imposes on every path that cannot benefit from it, and it is worth a line in the doc comment on `pending` so the next person sizing `ALPHA_BATCH` knows the width is paid unconditionally.

**Fix:** none required. If it ever shows up, `MaybeUninit<[Var; ALPHA_BATCH]>` plus a written-count is the safe-ish way out, but it trades a provable-by-hand invariant for an `unsafe` one and is not worth it at the measured cost.
**Verify:** `multi-stark/benches/poseidon2.rs` with `ZEROCHECK_POSEIDON2_NUM_VARS=13`, base vs branch -- I get 10.310 / 10.236 / 10.270 ms against 10.052 / 10.070 / 10.249 ms, i.e. flat to slightly better.

multi-stark/src/folder.rs
```diff
@@ -1024,6 +1069,59 @@ mod tests {
...
+
+        let alpha_powers = (0..WIDE_COLS)
+            .map(|i| alpha.exp_u64((WIDE_COLS - 1 - i) as u64))
+            .collect::<Vec<_>>();
+        let batched = TestFolder::new(&local, &next, boundary, &pis, alpha)
+            .with_alpha_powers(&alpha_powers)
+            .eval_air(&WideAir);
+
+        assert_eq!(batched, horner);
+    }
```
tcoratger: **[MINOR] The equivalence test reaches the new path but not the ring the change exists for**

`WIDE_COLS = 2 * ALPHA_BATCH + 3 = 19` is well chosen: it spans two full flushes plus a three-wide tail, so this test does exercise both the `mixed_dot_product` flush and `drain_pending`. I confirmed it reaches them. That is the bookkeeping, which is the part most likely to be wrong, so this is the right test to have written.

What it does not reach is the deferred reduction. `TestFolder = MultilinearFolder<'a, F, EF, EF>` with `EF = BinomialExtensionField<BabyBear, 4>`, which does not override `dot_product`, so `Algebra<R> for R` routes to the generic `sum_array` tree. The whole point of the change -- `PackedGhash128` accumulating unreduced and folding once -- is covered separately by `packed_ext::tests::the_dot_product_matches_the_sum_of_products`, but nothing composes the two. The composition is only covered indirectly, by the `backends_agree_*` whole-proof comparisons, which fail as "the proofs differ" rather than as "batch three used the wrong powers".

**Fix:** a second instantiation of this same test over a ring that does override `dot_product`, which costs almost nothing given `WideAir` already exists:

```rust
#[test]
fn alpha_batching_matches_horner_for_a_deferred_reduction_ring() {
    type R = Ghash128;                       // or PackedRepr<BinaryField128, Ghash128>
    // ... same body, with TestFolder replaced by
    // MultilinearFolder<'_, BinaryField128, R, R>
    assert_eq!(batched, horner);
}
```

**Verify:** run it twice -- once on the default target, where `<Ghash128 as Field>::Packing` is `Ghash128` itself (`WIDTH = 1`), and once under `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq"`, where it is `PackedGhash128` (`WIDTH = 2`) and the `binary-field/src/packed/x86_64.rs:421` lane kernel is compiled in at all. Both pass today; without the second instantiation, neither run tells you the folder's slot arithmetic and that kernel agree.

multi-stark/src/folder.rs
```diff
@@ -156,7 +169,8 @@ pub struct MultilinearFolder<'a, F, Var, Acc> {
 
 impl<'a, F, Var, Acc> MultilinearFolder<'a, F, Var, Acc>
 where
-    Acc: PrimeCharacteristicRing,
+    Var: PrimeCharacteristicRing + Copy,
+    Acc: Algebra<Var> + Copy,
```
tcoratger, line 176: **[NIT] The public builder's bounds tightened, and one half of that is avoidable**

`MultilinearFolder` and its `new` / `with_alpha_powers` / `eval_air` / `into_accumulator` are all `pub`, and this impl block moved from `Acc: PrimeCharacteristicRing` to `Var: PrimeCharacteristicRing + Copy, Acc: Algebra<Var> + Copy`. A downstream folder parameterised by a non-`Copy` `Var` -- a symbolic expression type, say, which is the obvious thing someone would plug in for constraint introspection -- stops compiling.

`Acc: Algebra<Var>` is genuinely new capability and has to be there. `Var: Copy` is not: it is required only by the array-repeat `[Var::ZERO; ALPHA_BATCH]`, which `core::array::from_fn(|_| Var::ZERO)` expresses without it.

**Fix:** drop `Copy` from `Var` and build `pending` with `from_fn`, or keep it and note the bound change in the changelog so the break is not discovered by a downstream build.
**Verify:** nothing in-tree exercises a non-`Copy` `Var`, so the compiler will not tell you; this is a deliberate-choice question rather than a bug.

tcoratger (approved):
Holds back `ALPHA_BATCH = 8` asserted constraints and weights the batch with one `mixed_dot_product`, so a ring with an `F2`-linear reduction reduces once per batch instead of once per constraint; plus the `PackedExt::dot_product` forwarding without which the wrapper hides that kernel; plus an independent `blake3-air` change that permutes message *positions* instead of message *words*. I re-derived the power-to-constraint pairing term by term and it is exact -- same constraint, same power, same order the verifier's own folder produces -- and every reassociation happens in a ring where addition is exactly associative, so no value can move. Safe to merge as-is; three MINORs are invariant-hardening and test reach, one NIT is a public-bounds note.

## What I verified

**The pairing is exact.** Write `i` for `constraint_index` at the assert, `n = powers.len()`, `B = ALPHA_BATCH`.

- `pending[i % B] = C_i`, so within a batch the slots fill `0, 1, ..., B-1` in order.
- The flush fires at `i % B == B-1`, i.e. `i = qB + B - 1`, and pairs `powers[i+1-B ..= i]` with `pending[0..B]`. Slot `s` holds `C_{qB+s}` and receives `powers[qB+s]`. Correct.
- `powers[i+1-B..].first_chunk::<B>()` cannot fail: `i >= B-1` at the flush so `i+1-B >= 0`, and `i < n` gives exactly `n - (i+1-B) >= B` remaining. The `expect` is dead, provably.
- `drain_pending`: `filled = min(i, n)`, `start = filled - filled % B` is a multiple of `B`, and slot `s in 0..(filled-start)` holds `C_{start+s}` because `(start+s) % B == s`. It receives `powers[start+s]`. Correct, including the `filled % B == 0` case where the loop is empty because the last flush already ran.

**Reassociation cannot change a value.** Three kernels can be reached:
- `Ghash128::dot_product` (`binary-field/src/ghash.rs:261`) and `PackedGhash128::dot_product` (`binary-field/src/packed/x86_64.rs:421`) accumulate the Karatsuba limbs unreduced and fold the modulus once. The `GF(2^128)` reduction is `F2`-linear, so `fold(sum_i p_i) = sum_i fold(p_i)`. Exact.
- Everything else falls to the default `mixed_dot_product` -- `products[i] = a[i] * f[i]` then `sum_array::<N>` -- which is a tree sum. Exactly associative in any ring; no floating point anywhere. Exact.

So the prover's batched accumulator equals the old term-by-term one bit for bit, and the verifier, which runs the same `MultilinearFolder`, is changed identically and consistently.

**`PackedExt::as_packings` is sound.** `PackedExt<F, P>` is `#[repr(transparent)]` over `P` beside a ZST `PhantomData`, so it has `P`'s size and alignment; `[PackedExt<F,P>; N]` therefore has `[P; N]`'s layout element for element, and the `&_ as *const _ as *const [P; N]` reborrow is well defined and correctly aligned. The `# SAFETY` comment states exactly the right invariant.

**The `blake3-air` index permutation is exactly equivalent, and the proof is one line.** `permute` (`blake3-air/src/constants.rs:30`) is `m[i] <- m[MSG_PERMUTATION[i]]`, a pure rearrangement of *positions* that never reads, compares or combines element values. So with `A_0 = block`, `A_k = permute(A_{k-1})` and `I_0 = id`, `I_k = permute(I_{k-1})`, induction gives `A_k[i] = A_{k-1}[P[i]] = block[I_{k-1}[P[i]]] = block[I_k[i]]`. Every `eval_g` therefore sees the same two `&[AB::Var; 32]` as before, so not one constraint changes -- which is why this commit can be dropped independently without touching the algebra. The borrow also works out because `local` borrows the `main` window (a `Copy` value moved out of `builder`), not `builder` itself, so `&local.block[..]` coexists with `&mut builder`.

**No prime-field measurement, so I took one.** The PR body says the prime-field paths are "unaffected", which is true of the *values* but not obviously of the *cost*: `MultilinearFolder<'a, F, F::Packing, EF::ExtensionPacking>` at `multi-stark/src/rounds.rs:1757` does attach alpha powers and so does take the new batched branch, where `EF::ExtensionPacking: Algebra<F::Packing>` has no `mixed_dot_product` override -- it gets the default, i.e. the same eight multiplies, plus a stack round trip through `pending` and a 256/512-byte zero-init per folder. That is a plausible mechanism for a small regression with no compensating win.

Measured: `multi-stark/benches/poseidon2.rs` (BabyBear zerocheck, `ZEROCHECK_POSEIDON2_NUM_VARS=13`, `1_airs`, `--features parallel`, `-C target-cpu=native`), base `757699775` vs branch, interleaved.

| | base | branch |
| --- | ---: | ---: |
| rep 1 | 10.310 ms | 10.052 ms |
| rep 2 | 10.236 ms | 10.070 ms |
| rep 3 | 10.270 ms | 10.249 ms |

Flat to slightly better. No prime-field regression on this host, and worth one line in the PR body since it closes the obvious question.

**The wide `Ghash128` packing needs the target features to exist at all.** `binary-field/src/packed/mod.rs:9` gates the whole `x86_64` module on `vpclmulqdq + (avx2 | avx512f)`; without them `<Ghash128 as Field>::Packing` *is* `Ghash128`, `WIDTH = 1`, and the lane kernel at `binary-field/src/packed/x86_64.rs:421` is never compiled. So a default-target CI run exercises `Ghash128::dot_product` but not `PackedGhash128::dot_product`. I ran the suite both ways -- default, and `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq"` (WIDTH 2) -- and both new tests plus all 288 `p3-multi-stark` tests and 492 `p3-binary-field` tests pass in each. Worth saying so explicitly in the PR body, since the kernel the whole change exists to reach is `cfg`-invisible on the default target.

**Bottom line:** merge.

multi-stark/src/folder.rs
tcoratger: Still open at `ca88be55`. The delta is a pure rebase (the added/removed line set is byte-identical to what I reviewed), so the sequence is still spelled out twice: `folder.rs:272-273` and `folder.rs:672-673`.

multi-stark/src/folder.rs
tcoratger: Still open at `ca88be55` -- `drain_pending` is unchanged, `min` and comment included.

multi-stark/src/folder.rs
tcoratger: Still open at `ca88be55`, and still the weakest of the four: I asked for a doc line on `pending`, not a code change, and `Fix: none required` stands. Flagging it only so the thread is not mistaken for something that got addressed.

multi-stark/src/folder.rs
tcoratger: Still open at `ca88be55` -- no second instantiation over a `dot_product`-overriding ring was added, so the folder's slot arithmetic and the `PackedGhash128` kernel are still only composed indirectly.

multi-stark/src/folder.rs
tcoratger: Still open at `ca88be55` -- `folder.rs:172` still carries `Var: PrimeCharacteristicRing + Copy` and `folder.rs:209` still builds `pending` with `[Var::ZERO; ALPHA_BATCH]`.

tcoratger (approved):
Re-review against the new head `ca88be55`. **The approval stands.**

## What the delta contains

A pure rebase, and I mean that literally: comparing this PR's own patch (`base..head`) at my approval head `f62b525b` and at `ca88be55`, the added/removed line set is **byte-identical**. Both commits are still there, unchanged. The branch moved only because its base did -- `perf/zc-alloc-hunt` went from `757699775` to `f97c9195` when #2266 was rebased under it.

Since the patch text did not move, the only question is whether the base move reaches this change. It does not: the base delta touches `multi-stark/src/rounds/{sliced,repr}.rs`, and this PR touches `multi-stark/src/folder.rs`, `multi-stark/src/packed_ext.rs` and `blake3-air/src/binary/air.rs`, none of which the base modified.

One interaction is worth naming because it is new and benign: the base's `RowTile::lane_pair` (`sliced.rs`) is a fresh user of `PackedExt::new`. This PR's `packed_ext.rs` change is purely additive -- a `dot_product` forward and a private `as_packings` -- so it changes nothing for that caller. The slot/power pairing, the exactness of the reassociation, the `repr(transparent)` soundness argument and the blake3-air index permutation are all unchanged.

Tested under the target features the change actually exists for, since the lane kernel is `cfg`-invisible otherwise: `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq" cargo test -p p3-multi-stark -p p3-binary-field -p p3-blake3-air --features parallel --release` -- 1101 passed, 0 failed.

## Still open

Four MINORs and the NIT were not addressed, and their threads stay open (replies posted on each):

- the single `finish()` exit that would stop the drain from being a convention across two call sites;
- the `debug_assert_eq!` pinning `drain_pending`'s dependency on `assert_alpha_power_count`;
- the doc line on `pending` recording that the zero-init is paid unconditionally (this one asked for no code change);
- the second instantiation of the equivalence test over a `dot_product`-overriding ring;
- the avoidable `Var: Copy` bound.

I have resolved the `[VERIFIED]` thread, which requested no change.

**Bottom line:** merge.

multi-stark/src/folder.rs
Nashtare: Left as is, no change.

I re-derived the three steps that carry the value rather than taking them on the page: `slot = i % B` with `i` incrementing by one, so slots fill in order and a flush sees only batch `q`; `powers[i+1-B..].first_chunk::<B>()` selecting `powers[qB ..= qB+B-1]` so slot `s` meets the weight `C_{qB+s}` had under `powers.get(constraint_index)`; and the `expect` unreachable because `i % B == B-1` gives `i + 1 - B >= 0` while `i < powers.len()` leaves at least `B` elements. All three hold as written.

One empirical datum for the thread, since the argument's prediction is testable: temporarily narrowing `ALPHA_BATCH` from 8 to 2 turns a 5-constraint AIR from "never flushes, drains 5" into "flushes twice, drains 1", and `prove_binary_field` at `log_height = 10` serializes to the same 94739 bytes with the same digest `0x0ca6e1c3d6f33e9b`. The batch width is not observable in the transcript, which is what associativity of the sum predicts.

multi-stark/src/folder.rs
Nashtare: Done, as proposed. `MultilinearFolder::finish` (`multi-stark/src/folder.rs:288`) now holds the count check and the drain, `into_accumulator` is a call to it, and the lookup-aware exit reads the accumulator through the same method (`folder.rs:696`). Commit `7b24d93d2`, `refactor(multi-stark): read the constraint accumulator through one exit`.

One exit does stay outside `finish`, and it cannot go through it: the `constraints_enabled == false` arm still reads `self.inner.accumulator` (`folder.rs:698`). Calling `finish` there would panic, because the powers are attached before the wrapper learns the family is off — `rounds.rs:1745` only skips a slot when *both* families are disabled, and `rounds.rs:1757` then calls `with_alpha_powers` unconditionally. So a disabled folder carries `Some(powers)` with `powers.len() > 0` and `constraint_index == 0`, and `assert_alpha_power_count` rejects it. That arm has nothing to drain by construction, since `InteractionMultilinearFolder::assert_zero` gates the forward, and it is now the only remaining read of the field.

The test you asked for is in: `check_wide_air_batching` (`folder.rs:1132`) ends by running the same batched-vs-Horner comparison through `InteractionMultilinearFolder::eval_air` with `constraints_enabled = true` and `WIDE_COLS` not a multiple of `ALPHA_BATCH`, so both exits are pinned rather than only the first.

multi-stark/src/folder.rs
Nashtare: Taken the `debug_assert_eq!` option. `drain_pending` (`multi-stark/src/folder.rs:302`) now opens with `debug_assert_eq!(self.constraint_index, powers.len())`, indexes `powers[start..]` off `powers.len()`, and the comment names `Self::finish` as what ran the check. The `min` and the old comment are gone. Commit `7b24d93d2`.

Checked the premise before editing: `assert_alpha_power_count` is a hard `assert!`, and after the collapse in the neighbouring thread it is the only statement between the caller and the drain on either path, so `filled` was `powers.len()` on every reachable call. Dropping the clamp is therefore a no-op in release and a tripwire in debug, which is the asymmetry you wanted.

multi-stark/src/folder.rs
Nashtare: Doc line added, nothing else. `pending` now records that every slot is zeroed when the folder is built, so an AIR asserting fewer than `ALPHA_BATCH` constraints still pays for the whole width — `multi-stark/src/folder.rs:158`, commit `ff121dda9`. No `MaybeUninit`, and no measurement figures in the comment.

For accuracy: the construction line did change, from `[Var::ZERO; ALPHA_BATCH]` to `core::array::from_fn(|_| Var::ZERO)`, but only to drop the `Var: Copy` bound your `folder.rs:173` thread asks about. It stores the same zeros, so the cost this thread describes is unchanged.

multi-stark/src/folder.rs
Nashtare: Added. The body moved into `check_wide_air_batching<BF, R>` (`multi-stark/src/folder.rs:1132`), and there is now a second instantiation over `MultilinearFolder<'_, BinaryField128, Ghash128, Ghash128>` in `alpha_batching_matches_horner_over_a_deferred_reduction_ring` (`folder.rs:1180`), where `Algebra<R> for R` routes `mixed_dot_product` into `Ghash128::dot_product` and its `poly_dot_128` unreduced accumulation. Both instantiations also run the comparison through `InteractionMultilinearFolder::eval_air`. Commit `5d48d686e`.

Checked the new test is not vacuous by mutating the kernel: with `Ghash128::dot_product` returning `poly_dot_128(..) * Self(3)`, the new test fails and the `BabyBear` one stays green, so the new instantiation is what composes the folder's slot arithmetic with that kernel and the old one genuinely never reached it.

That mutation also turned up a hole in the fixture, which is now fixed. An error added as a *constant* once per flush was invisible at `WIDE_COLS = 2 * ALPHA_BATCH + 3`: two flushes, and in characteristic two the two copies cancel — `+ Self(0xDEADBEEF)` passed. `WIDE_COLS` is now `3 * ALPHA_BATCH + 3`, an odd number of whole batches with the same three-wide tail, and the additive mutation fails as it should.

Not covered: this box is aarch64, so `binary-field/src/packed/x86_64.rs:421` cannot be compiled here at all. The new test covers the `WIDTH = 1` leg only; the `PackedGhash128` leg still needs an x86 run under `RUSTFLAGS="-C target-feature=+avx2,+vpclmulqdq"`, as you described.

multi-stark/src/folder.rs
Nashtare: Dropped. `folder.rs:175` is back to `Var: PrimeCharacteristicRing`, and `pending` is built with `core::array::from_fn(|_| Var::ZERO)` (`folder.rs:212`). Same for the `InteractionMultilinearFolder` inherent block (`folder.rs:645`), which carried the bound only to name the method. Commit `ff121dda9`.

Your non-`Copy` case is real in-tree — `air/src/symbolic/mod.rs:282` implements `Dup` for `SymbolicExpr<A>` by hand, which is exactly the shape the blanket `impl<T: Copy> Dup for T` does not cover.

Two corrections to what this buys, though. First, a non-`Copy` `Var` still cannot fold an AIR: `impl AirBuilder for MultilinearFolder` asks for `Var: Algebra<F> + Copy + Send + Sync`, and that bound is untouched by this PR. So what the revert restores is `new` / `with_*` / `into_accumulator`, not `eval_air`. Second, `Acc: Copy` also arrived with this PR and is still there; it is what the `&power` destructure in `drain_pending` reads, and `Dup::dup` would replace it if you want the pre-PR surface back in full. I left it because `Acc` is the accumulator rather than the trace-cell type, so it is the less likely of the two to be instantiated symbolically — say if you'd rather have it gone too.

PR #2268, opened 2026-09-20
perf(sumcheck): commit the suffix layout straight from the source tables

Nashtare:
Stacked on #2267.

`SuffixProver::from_witness` already says it:

> The stacked polynomial is intentionally discarded: every suffix-mode primitive walks the
> per-table data instead.

So in suffix mode `Witness::new` built a full stacked copy of the tables — 512 MiB at BLAKE3
`-l 18` — `commit_base` read it once into the codeword buffer, and it was dropped. Two copies of
the same data where one would do.

`Layout::commit` no longer hands `commit_base` a `&Poly`. It hands it a writer, and each layout
supplies the writer for its own variable order:

- `commit_base(encoder, mmcs, num_variables, folding, rate, write_message: impl FnOnce(&mut [F]))`
  keeps the part that is its job — allocating the zeroed codeword-height buffer and calling the
  encoder. The order-specific knowledge moves out into `write_stacked_message(order, poly, folding,
  message)`, which is what a caller that genuinely has a stacked polynomial uses.
- `PrefixProver::write_message` goes through `write_stacked_message` with its retained poly.
- `SuffixProver::write_message` calls `column_slots(&placements, &tables, 0, message)` straight
  into the codeword buffer.
- `Witness::poly` becomes `Option<Poly<F>>`, retained only by layouts whose prover reads it back.
  The suffix constructor now only plans the layout; `new_interleaved` still materialises its stack,
  because the prefix prover folds it directly — that is the residual folding polynomial, not merely
  a commit input. Nothing became dead, so nothing was deleted. This also makes one combination
  loud that used to be silent: a prefix-mode prover handed a contiguously stacked witness now
  panics on that `Option` instead of folding a suffix layout in prefix order, which the old
  `poly: Poly<F>` field could not tell apart.

`binary-pcs` is untouched: the Boolean path enters through `SuffixProver::new_witness` and
`MultilinearPcs::commit`, whose signatures are unchanged.

### Measurements

c7i.4xlarge, `-l 18`, `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, medians of 3, interleaved.

| | base | branch |
| --- | ---: | ---: |
| BLAKE3 prove | 2.437 s | 2.424 s (-0.5%) |
| BLAKE3 `new` span | 20.5 ms | 2.5 us |
| BLAKE3 `binary pcs commit` | 303 ms | 303 ms |
| Keccak-f prove | 0.776 s | 0.772 s |
| Keccak-f `new` span | 3.37 ms | 2.1 us |

The stack copy moved *inside* `commit`, where it replaced the message copy one for one — 23 ms of
copying became 20 ms of stacking, a wash — and the win is that `new`'s separate pass over 512 MiB
disappeared entirely. About 20 ms of a 2.44 s prove.

Re-running the `mem*` attribution confirms the copy is gone rather than relocated: total `mem*`
falls from 7.29% to 6.71% of program cycles, and the `commit_base` message-copy site is **absent**
from the profile rather than moved. The remaining stack-copy site also gets slightly cheaper, since
it now writes into a freshly mapped codeword buffer instead of into a separate poly that was then
read back.

### Peak RSS does not move, and it is worth saying why

`/usr/bin/time -v` maximum resident set size is identical on both sides (6,471,316 kB against
6,471,580 kB). Sampling `VmRSS` every 20 ms through a whole prove explains it: during the commit
phase the branch sits at 2.32 GiB against the base's 2.83 GiB, so the 512 MiB really is gone — but
the program's high-water mark is set around 1.7-1.8 s in, during the zerocheck and ring switch, at
roughly 6 GiB, and the commit phase never approaches it.

So this does not change which sizes are provable. Anyone tempted by the remaining 512 MiB in
`packing::pack` should know the same applies to it: the value there would be the ~10 ms and the
allocation, not the high-water mark.

### Correctness

Byte-identical: FNV digests of the postcard proof match at `-l 12`, `-l 14` and `-l 16` for both
objectives, with identical proof sizes. `cargo test --workspace --features parallel` passes 5,757
tests across 144 suites; clippy and rustfmt clean. The non-suffix callers keep working —
`PrefixProver` as used by whir, multi-stark and binary-dft, `SuffixLayoutPlan::fill` /
`FilledSuffixWitness`, and the `commit_base` consumers in `binary-dft`'s tests and benches.


sumcheck/src/layout/witness.rs
```diff
@@ -1086,10 +1071,64 @@ impl<F: Field> Witness<F> {
...
+    ///
+    /// # Panics
+    ///
+    /// - `out` must hold one cell per evaluation of the stacked polynomial.
+    pub(super) fn write_stacked_slots(&self, out: &mut [F]) {
+        assert_eq!(
+            out.len(),
+            1 << self.num_variables,
+            "stacked destination must cover the whole hypercube"
+        );
```
tcoratger, line 1095: **[MINOR] This function commits whatever it does not write, and only the length of that buffer is checked**

`write_stacked_slots` fills the column slots and leaves every other cell alone, so the unused tail of the hypercube is *the caller's bytes*, hashed verbatim into the Merkle root. The `assert_eq!` checks `out.len()`; nothing checks `out`'s contents. Before this PR the tail was guaranteed by construction, because the stack was always built on a fresh `Poly::<F>::zero(num_variables)` inside this same file.

Now the guarantee has moved out of the crate. It is `p3-sumcheck::commit::commit_base`'s `F::zero_vec(codeword_height * width)` that makes this correct, two modules away, through a `FnOnce` boundary that erases which buffer is being written. Both current callers are fine -- `commit_base` and `stacked_poly` -- and I verified the resulting roots match the old path exactly. But the contract holding it together is prose in the doc comment above.

The failure mode is the bad kind: a caller that reuses a buffer, or that swaps `zero_vec` for an uninitialised allocation to shave the memset (which is exactly the optimisation this stack has been hunting), gets a different root with no panic anywhere. The prover produces a well-formed proof of the wrong polynomial, and the only symptom is that the verifier rejects.

**Fix:**

```rust
debug_assert!(
    out.iter().all(|value| *value == F::ZERO),
    "write_stacked_slots leaves non-slot cells untouched, so the destination must arrive zeroed"
);
```

`O(2^num_variables)` but debug-only, and it pins the one cross-crate invariant this function depends on at the exact point it is relied upon. Alternatively, have the function zero the complement itself -- the planner emits placements largest-first with slot offsets equal to the running cursor, so the complement is the contiguous tail past `used`, as the test at the bottom of this file already notes. That would make the function total, at the cost of re-introducing part of the memset the PR removed.

**Verify:** the `debug_assert!` passes today; hand it a `vec![F::ONE; 1 << nv]` and it fires, which is the scenario no existing test can distinguish from the correct one.

sumcheck/src/layout/prover/prefix.rs
```diff
@@ -51,14 +52,25 @@ impl<F: Field, EF: ExtensionField<F>> Layout<F, EF> for PrefixProver<F, EF> {
...
     }
 
+    fn write_message(witness: &Witness<F>, folding: usize, message: &mut [F]) {
+        write_stacked_message(
+            VariableOrder::Prefix,
+            witness.retained_poly(),
+            folding,
+            message,
+        );
+    }
```
tcoratger, line 72: **[MINOR] The layout's variable order now lives in two independent places**

`Self::variable_order()` is a provided method on this very trait, derived from `Self::strategy()`, and it is what the *verifier* replays. This method hardcodes `VariableOrder::Prefix` instead. Before this PR there was one source: `commit_base(Self::variable_order(), ...)` took the order from the strategy, so prover and verifier could not disagree by construction.

Now they can. If `PrefixProver::strategy().variable_order` is ever changed -- a plausible thing to happen when a new strategy variant is added -- the prover keeps committing a prefix-transposed message while the verifier opens it as something else. The symptom is a rejected proof with nothing in the code pointing at the cause, because both halves look locally correct.

**Fix:**

```rust
fn write_message(witness: &Witness<F>, folding: usize, message: &mut [F]) {
    write_stacked_message(Self::variable_order(), witness.retained_poly(), folding, message);
}
```

and, in `suffix.rs`, the corresponding pin for the layout that does *not* go through `write_stacked_message` and so cannot express its order as an argument:

```rust
fn write_message(witness: &Witness<F>, _folding: usize, message: &mut [F]) {
    debug_assert_eq!(Self::variable_order(), VariableOrder::Suffix);
    witness.write_stacked_slots(message);
}
```

That second one is the more valuable of the pair, because "the contiguous slot layout *is* the suffix layout" is the fact the whole PR rests on and is currently written nowhere in executable form.

**Verify:** flip `PrefixProver::strategy()`'s order in a scratch build -- today the code still compiles and commits the old layout; after the fix the commit follows the strategy, and the `debug_assert_eq!` catches the suffix side.

binary-dft/tests/commit.rs
```diff
@@ -150,13 +150,14 @@ fn polynomial_commit_matches_naive_for_both_orders() {
...
                     &AdditiveRsEncoder::<F>::default(),
                     &mmcs,
-                    &Poly::new(values.clone()),
+                    poly.num_variables(),
                     folding,
                     rate,
+                    |message| write_stacked_message(order, &poly, folding, message),
                 );
                 assert_eq!(root, expected_root);
             }
```
tcoratger, line 226: **[MINOR] No non-ignored test pins the new suffix layout against an independent root**

This file has the only root-level oracles for `commit_base`, and after this PR both of them go through `write_stacked_message` -- the helper that takes an already-stacked `Poly`. `commit_base_matches_hand_encoding` covers `VariableOrder::Prefix` only. This test covers both orders, but it is `#[ignore]`d for heavy CI, and it still feeds `write_stacked_message`, not `write_stacked_slots`.

So the path the PR actually introduces -- `SuffixProver::write_message` calling `Witness::write_stacked_slots` straight into the codeword buffer -- has no root-level oracle at all. Its only coverage is end-to-end prove-and-verify in `binary-pcs` and `multi-stark`, which does catch a mismatch, but reports it as "the proof failed" with the cause several thousand lines away. For a change whose entire claim is "the committed bytes are identical", that is the wrong distance between defect and signal.

The cheapest test is the exact equivalence the PR asserts, and it needs no encoder and no MMCS:

```rust
#[test]
fn the_suffix_slot_writer_matches_the_stacked_message() {
    for folding in [0, 1, 2, 3] {
        let witness = SuffixProver::<F, F>::new_witness(tables(), folding);
        let nv = witness.num_variables();

        let mut direct = F::zero_vec(1 << nv);
        SuffixProver::<F, F>::write_message(&witness, folding, &mut direct);

        let mut staged = F::zero_vec(1 << nv);
        write_stacked_message(VariableOrder::Suffix, &witness.stacked_poly(), folding, &mut staged);

        assert_eq!(direct, staged, "folding={folding}");
    }
}
```

Use tables of mixed arity and width so the assertion actually sees selector padding and a non-empty zero tail -- a single square table would pass even if slot ordering were wrong.

**Verify:** I ran the root-level version of this across the branch boundary -- three tables of arities 6/5/4 and widths 3/2/1 (272 occupied cells in a 512-cell hypercube), `SuffixProver::commit` and `PrefixProver::commit` through `AdditiveRsEncoder` and the Keccak `MerkleTreeMmcs`, `folding` in `{0,1,2,3}`, rate 2 -- and all eight roots are byte-identical between `f62b525b` and this head. The gap is coverage, not a defect.

sumcheck/src/commit.rs
```diff
@@ -13,65 +13,81 @@ use crate::strategy::VariableOrder;
...
-        // Folding blocks are already contiguous, so the row width alone selects them.
-        VariableOrder::Suffix => {
-            let poly_values = poly.as_slice();
-            values[..poly_values.len()]
-                .par_chunks_mut(COPY_CHUNK)
-                .zip(poly_values.par_chunks(COPY_CHUNK))
-                .for_each(|(dst, src)| dst.copy_from_slice(src));
-        }
-    };
+    write_message(&mut values[..message_height * width]);
```
tcoratger, line 94: **[NIT] The callback can legally write nothing, and the function will commit the zero polynomial without complaint**

Before this PR `commit_base` took `poly: &Poly<F>` and could not be called wrong. It now takes `write_message: impl FnOnce(&mut [F])`, and if that closure returns without touching the buffer, the function Merkle-commits an all-zero codeword and returns a perfectly ordinary-looking root. The doc says the buffer "arrives zeroed" but does not say the closure is obliged to fill it, and there is no `# Panics` or contract section stating the obligation.

`Layout::write_message` is a required trait method, so no in-tree layout can skip it entirely -- but a layout that writes only part of the message has no backstop either, and `commit_base` is `pub` and reached directly from `binary-dft`'s tests and benches. Checking "not all zero" would be wrong (a zero polynomial is legitimate), so the realistic fix is documentation rather than a guard.

**Fix:** add to the doc above, next to the existing "receives the leading `2^num_variables` cells ... zeroed" sentence, that the callback must write every cell it does not intend to commit as zero -- and that anything it leaves untouched becomes part of the commitment. That is exactly the obligation `Witness::write_stacked_slots` relies on from the other side.
**Verify:** none; this is a contract that can only be stated, not asserted.

sumcheck/src/layout/witness.rs
```diff
@@ -1800,17 +1847,20 @@ mod tests {
...
         // Invariant:
         //     Handing the witness to the prefix prover preserves the
         //     stacked polynomial and the per-table shapes.
-        let w = fixture_witness();
-        let stacked_copy = w.poly.clone();
+        let w = fixture_interleaved_witness();
+        let stacked_copy = w.stacked_poly();
         let num_variables = w.num_variables();
 
         // Build a prefix-mode prover from the witness.
```
tcoratger, line 1885: **[NIT] This fixture change is a bug fix in disguise -- say so, and pin the new contract**

Switching from `fixture_witness()` (built by `Witness::new`, the contiguous suffix stack) to `fixture_interleaved_witness()` is not cosmetic. The old version of this test handed a *suffix-layout* witness to `PrefixProver::from_witness` and passed, because `from_witness` simply moved `parts.poly` across without any notion of which constructor had produced it. The test was asserting a property of a witness the prefix prover should never have accepted.

`parts.poly.expect("this layout retains its stacked polynomial")` in `prefix.rs` is what makes that combination loud instead of silent, and that is a real correctness improvement riding along with a perf change. Worth a line in the PR body, since a reader scanning the diff will read this hunk as a fixture tidy-up.

**Fix:** add the one-line test that documents the new contract, so the `expect` is not load-bearing-but-untested:

```rust
#[test]
#[should_panic = "this layout retains its stacked polynomial"]
fn prefix_prover_refuses_a_contiguously_stacked_witness() {
    let _ = PrefixProver::<F, EF>::from_witness(fixture_witness());
}
```

**Verify:** it passes on this head and would have failed (by *not* panicking) before the PR -- which is precisely the point.

tcoratger (approved):
Stops materialising a full stacked copy of the tables in `Witness::new` and instead writes the suffix layout straight into the codeword buffer, by turning `commit_base`'s `&Poly` argument into a `write_message` callback each layout supplies. The commitment binds these bytes, so the only question that matters is whether the new path produces the same message as the old one -- and it does: I committed the same tables on both sides of the branch and got byte-identical Merkle roots for both layouts, at four folding depths, with a non-trivial zero tail and mixed-arity selector padding. Safe to merge as-is; the MINORs below are about the invariants that now hold only by convention, and one real test-coverage gap.

## What I verified

**Byte identity, measured rather than argued.** I built three tables of arities 6, 5 and 4 with widths 3, 2 and 1 -- so 272 occupied cells in a 512-cell hypercube, leaving a 240-cell zero tail, and selectors of three different widths -- and ran `SuffixProver::commit` and `PrefixProver::commit` through the real `AdditiveRsEncoder` and Keccak `MerkleTreeMmcs` at `folding` in `{0, 1, 2, 3}` and rate 2, on base `f62b525b` and on this head. All eight roots match exactly. That covers the three things a layout change can break -- column order, partial-row padding, and the tail -- against the actual commitment rather than against an intermediate buffer.

**The message region is exactly the old one.** `commit_base` hands `&mut values[..message_height * width]`, and `message_height * width = 2^(nv-folding) * 2^folding = 2^nv`, which is what `write_stacked_slots`' `assert_eq!` demands and what the old `values[..poly_values.len()]` sliced. The prefix helper is also unchanged in substance: `RowMajorMatrixView::new(values, values.len() / width)` has second argument `2^nv / 2^folding = message_height`, the old literal.

**`FilledSuffixWitness::into_witness` dropping its poly does not duplicate work.** It still reconstructs the per-table dense views from `self.poly`, and the commit then re-stacks those into the codeword buffer -- the same two copies as before, with the intermediate `Poly` freed earlier instead of held until commit. Its round trip is covered: `direct.poly() == legacy.stacked_poly()` and `compatible.stacked_poly() == legacy.stacked_poly()` give it transitively.

**`binary-pcs` really is untouched and really is exercised.** `PcsLayout = SuffixProver`, entered at `binary-pcs/src/prover.rs:113`, with unchanged signatures -- and `binary-pcs` has end-to-end commit-open-verify tests over the new path, so a layout mismatch would surface as a failed verification. `cargo test -p p3-sumcheck -p p3-binary-pcs -p p3-binary-dft -p p3-whir --features parallel --release`: all pass.

**A nice side effect worth keeping.** Switching `witness_into_prefix_prover_preserves_stacked_polynomial` to a new `fixture_interleaved_witness` is not cosmetic -- it is the tell that the old test was handing a `Witness::new` (contiguous, suffix layout) to `PrefixProver::from_witness` and passing, because `from_witness` just moved `parts.poly` across without checking it was the interleaved stack. `parts.poly.expect("this layout retains its stacked polynomial")` turns that class of mismatch from silent into loud. That is a correctness improvement smuggled in with a perf change; it deserves a sentence in the PR body.

None -- every finding below sits on a changed line.

**Bottom line:** merge.

sumcheck/src/layout/witness.rs
tcoratger: Still open at `e1166b34`. The delta is a pure rebase, and the base move did not touch `sumcheck/src/commit.rs` or `sumcheck/src/layout/witness.rs`, so the cross-crate zeroed-buffer contract is still prose only.

sumcheck/src/layout/prover/prefix.rs
tcoratger: Still open at `e1166b34` -- `prefix.rs` still hardcodes `VariableOrder::Prefix` and `suffix.rs` still has no `debug_assert_eq!` pinning the suffix order.

binary-dft/tests/commit.rs
tcoratger: Still open at `e1166b34` -- no non-ignored root-level oracle for `write_stacked_slots` was added.

sumcheck/src/commit.rs
tcoratger: Still open at `e1166b34` -- the doc on `commit_base` still does not state the callback's obligation to write every cell it does not intend to commit as zero.

sumcheck/src/layout/witness.rs
tcoratger: Still open at `e1166b34` -- the `#[should_panic]` test pinning the new contract was not added, so the `expect` in `prefix.rs` stays load-bearing but untested.

tcoratger (approved):
Re-review against the new head `e1166b34`. **The approval stands.**

## What the delta contains

A pure rebase. Comparing this PR's own patch (`base..head`) at my approval head `47fe818b` and at `e1166b34`, the added/removed line set is **byte-identical**; the only difference anywhere in the two patches is one blob hash on an `index` line. The branch moved because its base did -- `perf/zc-air-eval` went from `f62b525b` to `ca88be55` as #2266 and #2267 were rebased under it.

The byte-identity result I reported (three mixed-arity tables, 272 occupied cells in a 512-cell hypercube, `SuffixProver::commit` and `PrefixProver::commit` through `AdditiveRsEncoder` and the Keccak `MerkleTreeMmcs` at `folding` in `{0,1,2,3}`, all eight roots matching) carries over, because the base move does not reach any of the surface it was measured on. Over the whole base delta, the only file this PR touches that the base also touched is `sumcheck/src/layout/prover/suffix.rs`, and that change is three deleted lines inside `mod tests` -- a duplicate `impl FromTable<Self> for BabyBearExt4` -- nowhere near the `Layout` impl where `write_message` lands. `sumcheck/src/commit.rs`, `sumcheck/src/layout/witness.rs` and `sumcheck/src/layout/prover/prefix.rs` were not touched by the base at all.

`cargo test -p p3-sumcheck -p p3-binary-pcs -p p3-binary-dft --features parallel --release` passes on the new head: 939 passed, 0 failed.

## Still open

None of the three MINORs or the two NITs were addressed; all five threads stay open (replies posted on each). All are about invariants that now hold by convention, plus the one coverage gap, so none of them blocks the merge:

- the `debug_assert!` pinning the zeroed-buffer contract `write_stacked_slots` depends on across a crate boundary;
- the hardcoded `VariableOrder::Prefix` in `prefix.rs`, and the matching `debug_assert_eq!` on the suffix side;
- the missing non-ignored root-level oracle for `write_stacked_slots`;
- the `commit_base` doc stating the callback's obligation;
- the `#[should_panic]` test pinning the contract that `prefix.rs`'s `expect` now enforces.

**Bottom line:** merge.

sumcheck/src/layout/witness.rs
Nashtare: Fixed in `1b932a3d`. `write_stacked_slots` now opens with the `debug_assert!` you wrote, and its `# Panics` section states that a debug build requires the destination to arrive zeroed.

Verified both directions: it passes on both callers today, and `stacking_rejects_a_destination_that_is_not_zeroed` in `sumcheck/src/layout/witness.rs` hands it `vec![F::ONE; 1 << nv]` and it fires. That test is `#[cfg(debug_assertions)]`-gated so a `--release` test run does not fail on a compiled-out assertion.

I did not take the alternative of zeroing the complement. `commit_base` already zeroes the whole `codeword_height * width` buffer with `F::zero_vec`, so for both in-tree callers a second memset would be duplicated work over a tail that can be up to half the message — and it would not catch the case you are actually worried about, a caller that hands over a dirty buffer, it would silently paper over it.

This is one of three comments on the same theme, and the assertion only closes the half of it that is reachable from inside this function. The other half — nothing independently checks the bytes that land — is `5cb7f1d6`, the root-level oracle in `binary-dft/tests/commit.rs`, which is what catches a *partial* write rather than a dirty destination. Details in that thread.

sumcheck/src/layout/prover/prefix.rs
Nashtare: Fixed in `4a327866`, both halves. `PrefixProver::write_message` passes `Self::variable_order()`, and `SuffixProver::write_message` carries `debug_assert_eq!(Self::variable_order(), VariableOrder::Suffix)` above the `write_stacked_slots` call.

Confirmed the premise rather than taking it on trust: `git show 00e3a9a1^:sumcheck/src/layout/prover/mod.rs` has `commit_base(Self::variable_order(), encoder, mmcs, &witness.poly, folding, starting_log_inv_rate)`. So the single source existed before this PR and hardcoding the order at the two new sites regressed it; this is not a pre-existing gap.

Agreed that the suffix `debug_assert_eq!` is the more valuable of the two — it is the only executable statement of the fact the PR rests on.

binary-dft/tests/commit.rs
Nashtare: Added in `5cb7f1d6` as `suffix_prover_commits_the_hand_stacked_message`, non-ignored, ~0.1 s. Two deliberate departures from your snippet.

**The snippet as written is circular.** `SuffixProver::new_witness` is `Witness::new`, which sets `poly: None`, and `Witness::stacked_poly` falls back to `write_stacked_slots` on a fresh `Poly::zero` (`sumcheck/src/layout/witness.rs:1131`). So `staged` and `direct` are produced by the same function and the assertion holds even if slot ordering is wrong. The oracle has to be built without touching the witness at all, so the test concatenates by hand: every column of the arity-6 table, then the arity-5 one, then the arity-4 one, zeros to the end of the 512-cell hypercube — 272 occupied, as in your run — and compares the root of `SuffixProver::commit` against the root of `commit_base` handed that buffer. The tables go in out of placement order, so largest-first is pinned too, and `folding` sweeps `{0, 2, 4}`.

**The transform is shared by both sides rather than routed through `NaiveAdditiveNtt`.** With the naive reference the test costs 12.5 s in a dev-profile run, 11.2 s of it at `folding = 0` where the width-1 message makes the `O(h^2)` reference worst — too much for a test that has to stay non-ignored. The transform is already pinned by `commit_base_matches_hand_encoding` and by the sweep below it; what had no oracle was the layout, and that is what this pins. The doc comment says so.

Mutation-checked rather than assumed: dropping the last slot from `column_slots` inside `write_stacked_slots` makes it fail at `folding=0` with two different roots.

sumcheck/src/commit.rs
Nashtare: Fixed in `82540dfb`. `commit_base`'s doc now states that every cell the callback leaves untouched is committed as zero, that there is no guard against a partial write, and that a callback which writes nothing commits the zero polynomial without complaint. `Layout::write_message` carries the same clause, since that is where an in-tree implementer reads the contract rather than at the `pub fn`.

Agreed that a not-all-zero guard would be wrong: the binary examples commit tables with genuinely zero columns, and the padded tail is zero by construction.

Documentation is as far as this one goes on its own, so it is paired with the two executable pins in the other threads — `1b932a3d` for the destination arriving zeroed, `5cb7f1d6` for the bytes that actually land.

sumcheck/src/layout/witness.rs
Nashtare: Added in `fb602749` as `prefix_prover_refuses_a_contiguously_stacked_witness`, with the `#[should_panic]` string you gave.

Confirmed the diagnosis instead of taking it on trust: `git show 00e3a9a1^:sumcheck/src/layout/prover/prefix.rs` line 54 is `poly: parts.poly` against a non-`Option` field, so the pre-PR code accepted the combination outright and the test would have failed by not panicking. The `expect` in `prefix.rs` is now covered.

On the PR body: I could not edit it from here, so the line is not in yet. The text I would add, at the end of the `Witness::poly` bullet:

> This also makes one combination loud that used to be silent: a prefix-mode prover handed a contiguously stacked witness now panics on that `Option` instead of folding a suffix layout in prefix order, which the old `poly: Poly<F>` field could not tell apart.

tcoratger:
@Nashtare I think we just need the CI to be green and then this big stack can be merged :)

PR #2269, opened 2026-09-20
test(poseidon2-air): check the constraints against the reference permutation and under corruption

kriss39:
## Summary

`p3-poseidon2-air`'s only tests validate the shape of `RoundConstants`. Nothing checks that the trace generator reproduces `Poseidon2BabyBear<16>`, that an honest trace satisfies `eval`, or that a corrupted trace fails it — the other AIR crates (`poseidon1-air`, `keccak-air`, `blake3-air`, `sha256-air`, `monolith-air`) all carry that kind of coverage. A missing or wrong constraint in this AIR would currently only surface through a downstream STARK test, if at all.

## Tests added (`air.rs`, both `SBOX_REGISTERS = 0` and `= 1`, canonical BabyBear width-16 constants)

- **Known answer**: the final `post` row equals `default_babybear_poseidon2_16().permute(input)`.
- **Constraint satisfaction**: `check_constraints` accepts an honest 8-permutation trace.
- **Targeted corruptions**: an input, a beginning/ending full-round `post`, a partial-round `post_sbox`, an S-box register (full and partial round) and the final output each produce a failing `check_all_constraints` report.
- **Every column feeds a constraint**: bumping any single cell of an honest row violates at least one constraint (298 columns for `REGISTERS = 1`, 157 for `REGISTERS = 0`; both widths are asserted), so no cell in the layout is free. This alone does not show that a given constraint in `eval` is present, which is what the forged rows below establish.
- **Forged cells with a recomputed tail**: a lone corruption is caught by the *next* constraint in the chain, so the sweep above cannot tell whether a cell's own constraint exists. These tests forge one committed cell (an S-box register, a full-round `post`, a partial-round `post_sbox`) and recompute every later cell from the forged value, so every other constraint holds by construction and the row passes if and only if the forged cell's own equality is missing. Sweeps drive this over every chain cell at both layouts; deleting any single `assert_eq` in `eval` now turns exactly one test red.

`cargo nextest run -p p3-poseidon2-air` passes; clippy and rustfmt clean.


poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+        });
+        assert_mutation_detected::<0>("ending full-round post", |cols| {
+            cols.ending_full_rounds[2].post[0] += F::ONE;
+        });
+    }
+
+    #[test]
+    fn test_corrupted_partial_round_post_detected() {
+        assert_mutation_detected::<1>("partial-round post_sbox", |cols| {
+            cols.partial_rounds[6].post_sbox += F::ONE;
```
tcoratger: **[MAJOR] Both `post` corruption tests survive deletion of the constraint they name**

The docstring on `test_forged_sbox_register_detected` states the problem precisely: a lone corruption is caught by the *next* constraint in the chain, so it says nothing about whether the corrupted cell's own constraint exists.
That is true of `post_sbox` and of full-round `post` as well, and here it is not addressed.

I deleted `builder.assert_eq(state[0].dup(), partial_round.post_sbox)` for `round == 6` only, the exact cell this line corrupts.
All 13 tests pass.
The corruption is still detected, but by the *following* round's constraints, which consume `post_sbox` through `internal_linear_layer`.

Same for line 471 and line 468: deleting the `post` assert for `beginning_full_rounds[1]`, or for the single cell `beginning_full_rounds[0].post[5]`, leaves all 13 tests green.
Deleting *all* 13 `post_sbox` asserts also leaves all 13 tests green.

Only two deletions are caught, and for a reason unrelated to the tests that name them.
Dropping every full-round `post` assert frees the final output, so the sweep and `test_corrupted_output_detected` catch it.
Dropping every beginning-round `post` assert frees `inputs`, which nothing else consumes, so `test_corrupted_input_detected` catches it.
Net: of the 282 assertions `eval` emits per row at `REGISTERS = 1`, 265 can be deleted one at a time without turning this suite red.
Since `post_sbox` and `post` are free variables once their equality is gone, each of those deletions is a real soundness break, not a cosmetic one.

The fix is the technique already in this PR, applied to the chain variable instead of the register.
The cheapest placement is the *last* partial round, where the downstream tail is short: bump `partial_rounds[PARTIAL_ROUNDS - 1].post_sbox`, write the forged value into `state[0]`, apply `internal_linear_layer`, then recompute the four ending full rounds' `sbox` and `post` exactly as `test_forged_sbox_register_detected` already recomputes one round.
The full-round case is even shorter: forge `ending_full_rounds[HALF_FULL_ROUNDS - 2].post[j]` and recompute the single remaining round.
Both are near-copies of the code at lines 511-544 and would move those 265 assertions into the pinned set for the round they target.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+        );
+        let report = check_all_constraints(&air, &trace, &[], Some(1));
+        assert!(
+            !report.is_ok(),
+            "a forged S-box register with a consistent post row passed every constraint"
+        );
+    }
+
+    /// Every column of the layout is load-bearing: bumping any single cell of an honest row
+    /// violates at least one constraint, so no cell is left unconstrained by `eval`.
```
tcoratger: **[MINOR] The sweep's conclusion is narrower than this sentence reads**

"no cell is left unconstrained by `eval`" is literally true and the sweep does deliver it: every column feeds at least one assertion.
But a reader will take it as "every assertion in `eval` is exercised", and M5/M6/M7 show that is not what it proves.
Worth rewording to something like "every column feeds at least one constraint", and pointing at the forged tests for the stronger property.

The PR description's framing has the same issue: "A missing or wrong constraint in this AIR would currently only surface through a downstream STARK test, if at all" implies the suite now catches missing constraints, and for 265 of the 282 it still does not.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+    fn every_column_is_constrained<const REGISTERS: usize>() {
+        let air = canonical_air::<REGISTERS>();
+        let honest = trace(&air);
+        let width = honest.width();
```
tcoratger, line 732: **[MINOR] Assert the width, and the description's 170 is wrong**

`num_cols` is `WIDTH + 2 * HALF_FULL_ROUNDS * (WIDTH * REGISTERS + WIDTH) + PARTIAL_ROUNDS * (REGISTERS + 1)`.
At `WIDTH = 16`, `HALF_FULL_ROUNDS = 4`, `PARTIAL_ROUNDS = 13` that is `16 + 256 + 26 = 298` for `REGISTERS = 1` and `16 + 128 + 13 = 157` for `REGISTERS = 0`.
I printed `BaseAir::width` for both: 298 and 157.
The PR description says 170 for `REGISTERS = 0`, which is 13 too many.

Since neither number is asserted anywhere, nothing would have caught that.
An `assert_eq!(width, if REGISTERS == 1 { 298 } else { 157 })` here makes the sweep notice a layout change instead of silently sweeping a different number of columns.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+
+    use super::*;
+    use crate::RoundConstants;
+
+    type F = BabyBear;
+    const WIDTH: usize = 16;
+    const SBOX_DEGREE: u64 = BABYBEAR_S_BOX_DEGREE;
+    const HALF_FULL_ROUNDS: usize = BABYBEAR_POSEIDON2_HALF_FULL_ROUNDS;
+    const PARTIAL_ROUNDS: usize = BABYBEAR_POSEIDON2_PARTIAL_ROUNDS_16;
+    const NUM_HASHES: usize = 8;
```
tcoratger: **[MINOR] `NUM_HASHES = 8` exercises only one of the two trace generators**

`generate_perms` dispatches to `generate_trace_rows_for_perm_batch` when `F::Packing::WIDTH > 1 && inputs.len() % F::Packing::WIDTH == 0`, and to the scalar `generate_trace_rows_for_perm` otherwise.
I instrumented that branch and ran the KAT under three flag sets:

- default (CI's running legs set no `target-feature`, so `Packing = Self`, width 1) -> scalar
- `-C target-feature=+avx2` (width 8) -> packed
- `-C target-feature=+avx512f` (width 16, `8 % 16 != 0`) -> scalar

So the packed generator, which is the one a production prover actually runs, is validated against `default_babybear_poseidon2_16()` on AVX2 hosts only, and never in CI.
`NUM_HASHES = 16` is divisible by 1, 2, 4, 8 and 16, so it reaches the packed path on every `Packing` width the repo builds.
The cost is one extra row per test; runtime is 0.02 s today.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+    /// [`default_babybear_poseidon2_16`] rather than only against itself.
+    fn canonical_air<const REGISTERS: usize>() -> Air<REGISTERS> {
+        Poseidon2Air::new(RoundConstants::new(
+            BABYBEAR_POSEIDON2_RC_16_EXTERNAL_INITIAL,
+            BABYBEAR_POSEIDON2_RC_16_INTERNAL,
+            BABYBEAR_POSEIDON2_RC_16_EXTERNAL_FINAL,
+        ))
+    }
+
+    fn inputs() -> Vec<[F; WIDTH]> {
```
tcoratger, line 407: **[MINOR] The KAT inputs are a single arithmetic progression**

`i * WIDTH + j` gives exactly `0..128` across the eight rows, so the whole known-answer test rides one structured input family.
A KAT should be deterministic, so I would not reach for `proptest` here, but a `SmallRng::seed_from_u64` draw of eight random `[F; 16]` costs nothing and covers full-range limbs instead of small ones.

On `proptest` more broadly: it buys nothing for the corruption tests.
`+= F::ONE` in a prime field always changes the value, and these constraints are polynomial identities, so a random delta cannot fail where `1` succeeds — the sweep over all 298 columns with a fixed delta is already the stronger statement.
The place randomization would pay is exactly this input set, and a seeded RNG gets it without the `proptest` dependency or the runtime.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+        assert_mutation_detected::<1>("full-round sbox register", |cols| {
+            cols.beginning_full_rounds[0].sbox[2].0[0] += F::ONE;
+        });
+        assert_mutation_detected::<1>("partial-round sbox register", |cols| {
+            cols.partial_rounds[12].sbox.0[0] += F::ONE;
+        });
+    }
+
+    #[test]
+    fn test_corrupted_output_detected() {
```
tcoratger, line 517: **[MINOR] `REGISTERS = 0` is missing from the output test**

`test_corrupted_output_detected` runs `REGISTERS = 1` only, but the final `post` row exists in both layouts and is the single most important cell in the AIR — it is the one the S-box-register count does not change.
`assert_mutation_detected::<0>` here costs one more line and one more millisecond.

`test_forged_sbox_register_detected` being `REGISTERS = 1` only is fine: `REGISTERS = 0` has no registers to forge.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+
+    #[test]
+    fn test_corrupted_sbox_register_detected() {
+        // With one register the S-box is split as x^3 · x^3 · x, so a wrong intermediate must
+        // be caught by its own constraint rather than absorbed downstream.
+        assert_mutation_detected::<1>("full-round sbox register", |cols| {
+            cols.beginning_full_rounds[0].sbox[2].0[0] += F::ONE;
+        });
+        assert_mutation_detected::<1>("partial-round sbox register", |cols| {
+            cols.partial_rounds[12].sbox.0[0] += F::ONE;
```
tcoratger: **[NIT] Hard-coded round indices read as arbitrary**

`partial_rounds[12]` is the *last* partial round at `PARTIAL_ROUNDS = 13`, and `ending_full_rounds[2]` on line 471 is the second-to-last.
Neither reads that way, and if the BabyBear constants ever change round counts these become a different position or an out-of-bounds panic.
`PARTIAL_ROUNDS - 1` and `HALF_FULL_ROUNDS - 2` say what is meant, and the file already uses `HALF_FULL_ROUNDS - 1` on lines 423 and 500.

poseidon2-air/src/air.rs
```diff
@@ -338,3 +338,233 @@ fn eval_sbox<AB, const DEGREE: u64, const REGISTERS: usize>(
...
+    #[test]
+    fn test_corrupted_output_detected() {
+        assert_mutation_detected::<1>("final output", |cols| {
+            cols.ending_full_rounds[HALF_FULL_ROUNDS - 1].post[15] += F::ONE;
+        });
+    }
+
+    /// A forged S-box register with every later cell recomputed from it.
+    ///
+    /// Single-cell corruption is always caught by the *next* constraint in the chain, so it
```
tcoratger: **[VERIFIED] This docstring's claim is exactly right**

Confirmed by mutation: removing `builder.assert_eq(committed_x3.dup(), x.cube())` from the `(7, 1)` arm of `eval_sbox` makes `test_forged_sbox_register_detected` the only failing test, with the column sweep and all six corruption tests still green.
The premise that a lone corruption is absorbed by the neighbouring constraint, and that recomputing the round's `post` from the forged value is what isolates the register's own constraint, holds under execution.

This is the right technique and the most valuable test in the PR.
The MAJOR above is just that it is applied once, to the register, and not to the `post` / `post_sbox` chain variables that have the identical problem.

tcoratger (commented):
Reviewed by running it, not by reading it.
Everything below comes from mutating `eval` in a detached worktree at `f56b5c7` and re-running `cargo test -p p3-poseidon2-air --lib`.

The PR's own headline claim checks out exactly as written.
Removing `builder.assert_eq(committed_x3.dup(), x.cube())` from the `(7, 1)` arm of `eval_sbox` makes **only** `test_forged_sbox_register_detected` fail; the other twelve stay green.
The forged-register test is genuinely load-bearing and its stated justification is correct.

But the same reasoning it is built on exposes a gap the rest of the suite does not close.
Full mutation table:

| # | mutation in `eval` | suite result |
|---|---|---|
| M1 | all full-round `post` asserts removed | 4 fail |
| M2 | all 13 partial-round `post_sbox` asserts removed | **13 pass** |
| M3 | `(7,1)` `committed_x3 == x^3` removed | only `test_forged_sbox_register_detected` fails |
| M4 | `post` asserts removed in all beginning full rounds | 2 fail |
| M5 | `post` assert removed only for `beginning_full_rounds[1]` | **13 pass** |
| M6 | `post_sbox` assert removed only for `partial_rounds[6]` | **13 pass** |
| M7 | `post` assert removed only for `beginning_full_rounds[0].post[5]` | **13 pass** |
| M8 | `post` assert removed only for `ending_full_rounds[3].post[15]` | 2 fail |

`check_all_constraints` reports 282 assertions per row at `REGISTERS = 1` and 141 at `REGISTERS = 0`.
Of those, the suite individually pins 17 and 16 respectively: the 16 `post` cells of the final ending round, plus the one S-box register the forged test targets.
Details and the concrete fix are in the inline comment on the partial-round test.

Other things I checked that are clean:

- The KAT is independent. `default_babybear_poseidon2_16()` runs `Poseidon2ExternalLayerBabyBear` / `Poseidon2InternalLayerBabyBear`, while the trace comes from `generate_trace_rows` over `GenericPoseidon2LinearLayersMonty31`. Different code, same canonical constants. Not a self-comparison.
- Both `SBOX_REGISTERS` values really diverge: width 298 vs 157, and `eval_sbox` takes the `(7,1)` arm vs the `(7,0)` arm. Not a copy-paste with a changed constant.
- The sweep covers every column: it iterates `0..width` with no skips, and `+= F::ONE` in a prime field always changes the value, so no cell can be silently missed and no bump can wrap to an equivalent value. The AIR is strictly row-local (`main_next_row_columns()` is empty, `eval` only reads `current_slice`), so corrupting row 1 cannot leak into another row's check, and `max_failures = Some(1)` is a pure early exit rather than a narrowing of what is checked.
- Runtime is a non-issue: 0.02 s for all 13 tests in debug, 0.16 s wall for `test_every_column_is_constrained` alone including process startup. No reduction needed.
- `cargo clippy -p p3-poseidon2-air --all-targets` is warning-free. Debug and release both pass.

One event note: `COMMENT` rather than `APPROVE` only because of the MAJOR.
This is a well-scoped, genuinely useful addition and the forged-register idea is the right one; it just needs to be applied once more.

poseidon2-air/src/air.rs
tcoratger: Fixed in 51271d1b, and carried further than a single placement.

`forge_chain_cell` replays the whole row from `inputs`, forges one committed chain cell, and recomputes every later register and chain cell from the forged value — so the row is honest everywhere except at the target, and only the target's own constraint can reject it. Two sweeps then walk it over the whole chain: `test_forged_full_round_post_detected` over all `2 * HALF_FULL_ROUNDS` full rounds x `WIDTH` post cells, and `test_forged_partial_round_post_sbox_detected` over all `PARTIAL_ROUNDS` `post_sbox` cells, each at `REGISTERS = 0` and `1`. Each forgery also asserts the permutation output moved, since every round is a bijection.

Re-ran your mutation table against the new suite:

| # | mutation in `eval` | before | after |
|---|---|---|---|
| M1 | all full-round `post` asserts removed | 4 fail | 5 fail |
| M2 | all 13 partial-round `post_sbox` asserts removed | **13 pass** | only `test_forged_partial_round_post_sbox_detected` fails |
| M3 | `(7,1)` `committed_x3 == x^3` removed | only forged-register fails | unchanged |
| M5 | `post` assert removed only for `beginning_full_rounds[1]` | **13 pass** | only `test_forged_full_round_post_detected` fails |
| M6 | `post_sbox` assert removed only for `partial_rounds[6]` | **13 pass** | only `test_forged_partial_round_post_sbox_detected` fails |
| M7 | `post` assert removed only for `beginning_full_rounds[0].post[5]` | **13 pass** | only `test_forged_full_round_post_detected` fails |
| M8 | `post` assert removed only for `ending_full_rounds[3].post[15]` | 2 fail | 3 fail |

Five more single-constraint deletions sampled across the chain (`ending_full_rounds[0].post[7]`, `beginning_full_rounds[3].post[0]`, `partial_rounds[0]`, `partial_rounds[12]`, all of `ending_full_rounds[1].post`) each turn exactly one sweep red. Runtime went from 0.02 s to 0.03 s in debug.

poseidon2-air/src/air.rs
tcoratger: Raised to 16. Instrumented the dispatch and confirmed the change, one nuance:

```
NUM_HASHES=8   default   packing_width=1  -> SCALAR
NUM_HASHES=8   +avx2     packing_width=8  -> PACKED
NUM_HASHES=8   +avx512f  packing_width=16 -> SCALAR
NUM_HASHES=16  default   packing_width=1  -> SCALAR
NUM_HASHES=16  +avx2     packing_width=8  -> PACKED
NUM_HASHES=16  +avx512f  packing_width=16 -> PACKED
```

So 16 buys the AVX-512 leg. The default leg stays scalar whatever the count, because the guard is `packing_width > 1` and `Packing = Self` there — at width 1 there is no packed generator to reach. Cost is one extra row; the suite is 0.03 s in debug.

tcoratger:
@kriss39 I pushed the review fixes straight to this branch (51271d1b) rather than leave you a list — force-pushed with a lease pinned to `f56b5c72`, so nothing of yours was rewritten. Amend, squash, or drop it as you like.

What landed:

- **The MAJOR.** `forge_chain_cell` generalises your forged-register idea to the chain variables: it replays the row from `inputs`, forges one committed cell, and recomputes every later register and chain cell from the forged value. Two sweeps drive it over every full-round `post` cell and every partial-round `post_sbox`, at both `SBOX_REGISTERS = 0` and `1`. Deleting any single one of those asserts now turns exactly one test red, where 265 of the 282 assertions could previously be deleted silently. The full before/after mutation table is in the resolved thread.
- Layout width asserted as 298 / 157 in the column sweep.
- `NUM_HASHES` 8 -> 16, so the packed trace generator is reached on AVX-512 too (dispatch trace in the thread).
- Test inputs drawn from a seeded `SmallRng` instead of `i * WIDTH + j`.
- `REGISTERS = 0` added to the output test; hard-coded round indices replaced by `PARTIAL_ROUNDS - 1` / `HALF_FULL_ROUNDS - 2`; column-sweep wording narrowed to what it actually proves.

One thing for you, since I did not touch your description: it says **170 columns** for `SBOX_REGISTERS = 0`. The real width is **157** (`16 + 128 + 13`); the 298 for `SBOX_REGISTERS = 1` is right. Worth correcting in the description — the code now asserts both numbers.

`cargo test -p p3-poseidon2-air` green in debug and release (15 tests, 0.03 s debug), `cargo clippy --all-targets` warning-free.

tcoratger (approved):
Approving on the mutation evidence, at 51271d1b.

The MAJOR is closed. The forged-and-recomputed row is now applied to every committed chain cell, not just the one S-box register, so the corruption tests bite the constraint they name instead of a neighbour.

| mutation in `eval` | before | after |
|---|---|---|
| all 13 partial-round `post_sbox` asserts removed | **13 pass** | 1 fail |
| `post_sbox` assert removed only for `partial_rounds[6]` | **13 pass** | 1 fail |
| `post` assert removed only for `beginning_full_rounds[1]` | **13 pass** | 1 fail |
| `post` assert removed only for `beginning_full_rounds[0].post[5]` | **13 pass** | 1 fail |
| `(7,1)` `committed_x3 == x^3` removed | 1 fail | 1 fail |
| all full-round `post` asserts removed | 4 fail | 5 fail |
| `post` assert removed only for `ending_full_rounds[3].post[15]` | 2 fail | 3 fail |

Each of the four that used to survive now fails exactly one test — the sweep that targets that cell. Five further single-constraint deletions sampled across the chain (`ending_full_rounds[0].post[7]`, `beginning_full_rounds[3].post[0]`, `partial_rounds[0]`, `partial_rounds[12]`, all of `ending_full_rounds[1].post`) behave the same way.

Also in: the 298 / 157 width assertions, `NUM_HASHES = 16` reaching the packed generator on AVX-512, seeded random KAT inputs, `REGISTERS = 0` on the output test, and round indices named by their round counts.

Debug and release green, clippy warning-free, suite runs in 0.03 s. Only open item is the description, which still says 170 columns for `SBOX_REGISTERS = 0` where the asserted width is 157.

kriss39:
Thanks for landing the chain-cell sweeps — that is the generalisation the register test was asking for. Description corrected: 157 columns at `SBOX_REGISTERS = 0`, and the sweep paragraph now says only what it proves. Re-ran the crate on `51271d1b` locally: 15 passed.

Issue #2271, opened 2026-09-20
Make caller soundness obligations enforceable, not documented


tcoratger:
## Deliverable

Make the backend's soundness conditions checkable by the compiler or the code, rather than stated in prose that a caller can read and ignore.

## The pattern

Review of the binary track keeps finding the same shape.

A reduction is sound only if the caller does something, the doc says so, and nothing enforces it.

Four instances surfaced in one review round:

- A claim could be discarded at the call site and the surrounding verification still accepted, because the type carried no use obligation.
- An opening value was determined only by an equation that degenerates, so a prover could choose it freely when one factor vanished.
- A security parameter took a caller-supplied field width with nothing tying it to the field actually in use, so passing the base width understated the bound by 64 bits in the unsafe direction.
- Statement dimensions that decide how a verifier authenticates a leaf sat outside the transcript.

Each was fixed where it was found.

None of them was caught by a mechanism, and the same shape is reachable anywhere a reduction hands a value back for a caller to discharge.

## A related class

Small parameter regimes are not bounded, and at least one of them silently weakens a property.

At two leaves per side a product reduction contracts to a root check with no sampled challenge, which made a cross-instance separation test pass against every transcript until the leaf count was raised.

There is no minimum on the layer count and no test that a reduction is non-degenerate at the smallest shape a caller can construct.

## Acceptance criteria

- Every returned claim that a caller must discharge carries a use obligation the compiler enforces.
- Every security-parameter input is derived from the field in use, or checked against it, rather than supplied unchecked.
- Every statement dimension that changes how a verifier authenticates anything is absorbed before the first challenge that depends on it.
- Ordering obligations that genuinely cannot be enforced in-crate are listed in one place rather than discovered per module.
- Each reduction states its minimum non-degenerate shape and rejects below it, or documents why the degenerate case is harmless.
- A test exercises each reduction at its smallest legal shape and fails if a challenge is never sampled.
- An audit pass sweeps the binary crates for the remaining instances of the pattern and records what it found.

## Scope

Does not add protocol features.

Converts existing documented obligations into checked ones and closes the degenerate-shape gap.

gopikannappan:
Is the last criterion here open for outside help?

> A test exercises each reduction at its smallest legal shape and fails if a challenge is never sampled.

[...]

PR #2272, opened 2026-09-20
feat(multi-stark): publish the machine-facing backend contract and proof envelope

### LINKED ISSUE #2240: Publish the machine-facing backend contract and proof envelope
## Deliverable

Expose a stable backend surface a separate RISC-V machine can use without reaching into protocol internals.

## Acceptance criteria

- A declarative table interface covers columns, local constraints, bus flushes, and runtime log heights.
- No VM-specific concept appears in Plonky3 public types.
- All proof lengths and allocations are derived from public parameters.
- The proof envelope is bounded, versioned, and rejects unknown or malformed shapes before transcript replay.
- Binding-only versus hiding commitments are explicit in types and documentation.
- Serialization changes have compatibility fixtures and an intentional version policy.
- Two repositories can evolve without silently decoding one proof shape as another.

tcoratger:
A separate repository will hold the RISC-V machine: the ISA, the ELF loader, the emulator, the instruction and precompile chips, the guest toolchain. This PR is the seam between it and Plonky3. It has two halves, and only one of them carries soundness.

Closes #2240

## 0. Module layout

`multi-stark/src/contract/`, one concern per file:

| File | Holds |
| --- | --- |
| `table.rs` | the table declaration and its column, constraint, flush and height pieces, with their ceilings |
| `machine.rs` | the machine declaration, its ceilings, and the seal / open / verify methods |
| `run.rs` | the per-proof choices |
| `envelope.rs` | the wire layout, the header, the sealed and accepted proofs |
| `constraints.rs` | the canonical byte string naming everything one table fixes |
| `digest.rs` | length-prefixed absorption into a canonical byte string |
| `error.rs` | every error the contract returns |
| `mod.rs` | wiring and the public re-exports |

Sub-modules are private and the surface is re-exported from `contract`, matching how `bus/src/product` and `multi-stark/src/logup_star` are wired.

## 1. The declarative table interface

`multi-stark/src/contract/table.rs`.

A `TableDeclaration` fixes, before any proof exists:

- how many columns of each kind the table has (committed, preprocessed, public);
- how many local constraints it asserts and their largest per-variable degree;
- the families of tuples it flushes onto named channels, with each tuple's width and its per-row multiplicity bound;
- **every constraint and interaction expression itself**, as a canonical byte string, together with everything the table declares outside its evaluation;
- the **range** of heights it may run at.

`TableDeclaration::from_constraints` reads all of that off the constraint system in one symbolic pass, and it is the **only** constructor: there is no way to state a table by hand, so there is no way to state one the constraint system disagrees with. `MachineDeclaration::verify` additionally reads each system back from the verifier's own AIRs and refuses a table that disagrees on any part, naming the part. The counts alone were not enough — two AIRs with different transitions agree on every count — so `constraints.rs` walks the symbolic expression DAG, the bus interactions, the local lookups and the indexed declarations into one injective byte string, prefixed by the field order and the extension's algebra identity. Sharing is collapsed the way the crate's existing flattened wire form collapses it, so the string is linear in the DAG rather than in its expansion.

A symbolic pass runs `Air::eval` alone, so a table's declaration is not only what its evaluation writes. The string therefore also covers the cells the backend pins to public values, the periodic tables, both next-row column lists and the degree hint; and the declared constraint count and degree are scored the way the reduction scores them, pins included, matching `get_air_profile`.

Bus declarations need a second pass of their own. `InteractionSymbolicBuilder::record_bus_interaction` keeps nothing but the Booleanity check of a conditional activation, so the name, the direction, the tuple and the activation would all be invisible. `from_constraints` runs `BusSymbolicBuilder` as well and writes every declaration it collected. The declaration count is one of the parts a verification compares, and a statement that moves a tuple across a bus refuses a proof carrying no bus section.

What the string deliberately leaves out, and why:

- the contents of a fixed trace: the verifying key commits to them and an opening against that commitment is what binds them, so the fingerprint would be shadowing a stronger binding;
- the PCS parameters: they belong to the verifier's configuration rather than to the statement, and their effect is covered by the declared security target, which is computed from that configuration.

A `MachineDeclaration` is every table of one statement, a proof-size budget, a security target and a hash. The hash is a type parameter bounded by the workspace's existing `p3_symmetric::CryptographicHasher<u8, [u8; 32]>`, supplied on the declaration, so the downstream repository picks its own and the choice is part of the statement rather than an argument a caller can vary per call. `Keccak256Hash` is what the tests use. A `Run` is the per-proof choice: one height per table and the grinding difficulty. A `Run` can only be built by `MachineDeclaration::run`, which rejects a height outside the declared range, so an out-of-range height never reaches the prover or the verifier — that is an obligation in a type rather than in prose (#2271).

The declaration also carries the security level every verification of the statement must reach. It is absorbed into the fingerprint and `MachineDeclaration::verify` routes through `verify_with_security`, so a downstream whose only entry point is this contract cannot verify below a target it published.

A height range starting at a single row is refused at declaration time. It used to be accepted and then panic inside the zerocheck, which asserts at least two rows.

**Vocabulary.** The words are table, column, constraint, channel, flush, height, run, statement. No opcode, register, program counter, or memory word appears anywhere in the surface. A flush carries no direction, because the sign of a multiplicity is a trace value and no constraint system fixes a side.

## 2. The proof envelope

`envelope.rs` owns the wire layout and decides only whether a byte string is framed at all. `machine.rs` decides whether that framing suits a statement. This is the parsing-soundness surface, so here is the threat model line by line.

### The framing

A fixed 48-byte header, then the body:

```
0..8    opening bytes
8..10   framing revision   (u16 LE)
10..12  body revision      (u16 LE)
12..44  fingerprint of the statement and the run
44..48  body length        (u32 LE)
48..    body
```

### What an attacker controls, and where each one is bounded

| Attacker-controlled | Bounded by | Error |
| --- | --- | --- |
| total input length | compared against the header, exactly | `HeaderTooShort` / `Truncated` / `TrailingBytes` |
| the opening bytes | a compiled-in constant | `BadMagic` |
| the framing revision | equality with a compiled-in constant | `EnvelopeVersion` |
| the body revision | equality with a compiled-in constant | `BodyRevision` |
| the fingerprint | recomputed locally from the declaration and the run | `RunMismatch` |
| the body length field | the declared budget, **before** it is used to slice | `BodyAboveBudget` |
| every length inside the body | the body slice, which the budget already bounds | `Malformed` |
| bytes the decoder does not read | required to be zero | `UnreadBodyBytes` |
| presence of the optional parts | the declaration, not the proof | `SectionMismatch` |
| the indexed reader count | the declaration's total indexed reads | `CountMismatch` |
| the grinding difficulty | taken from the run; the decoded proof never leaves the crate | n/a |
| the instance heights | compared against the run | `RunDisagreement` |

Every one of those runs before the statement verifier is called, so before the transcript is seeded. The whole chain is one method, `MachineDeclaration::verify`, sitting beside `seal` and `open`.

**Unbounded length fields.** The only length the reader takes from the input is the header's body length. It is compared against the declared budget *before* it is used to index anything, and then the remaining input must equal it exactly. Inside the body, postcard sequence lengths are still attacker-chosen, but the decoder reads from a slice the budget already bounds; each element costs at least one input byte, so total allocation is linear in the budget plus one capped pre-allocation for the sequence that runs out of input. The budget is a public parameter of the statement, absorbed into the fingerprint, and capped at 1 GiB regardless.

**Counts that disagree with the declared shape.** The optional lookup part, the optional indexed part, and the optional preprocessed opening must be present exactly when the declaration says so, and the indexed reader count must equal the declared total. The sumcheck round count and grinding-witness count were already checked up front by `GenericDegreeProof::verify` before it absorbs anything, so they are not re-checked here.

**Trailing bytes.** Two separate checks, because two separate things can trail. The framing requires `input.len() - 48 == declared`, and the decoder must leave no remainder. The second one is not redundant: `postcard::from_bytes` decodes and silently ignores whatever follows the value, so the envelope uses `take_from_bytes` and rejects a non-empty remainder. `a_body_the_decoder_does_not_finish_is_refused` demonstrates exactly that — the input is a valid proof followed by one padding byte, with a length field that covers it, and only the remainder check catches it.

**Version confusion.** Three independent mechanisms, with distinct errors so a failure is diagnosable:

- the framing revision guards the header layout;
- the body revision guards the serialized proof layout — change any field of `MultiStarkProof` and this must be bumped, or old proofs decode as the new shape;
- the fingerprint guards everything that changes the *meaning* of the bytes without changing their layout: the tables and the constraints they assert, the channel names, the heights, the grinding difficulty, the security target, and the hash itself.

Two repositories can therefore evolve without one decoding the other's proof as a different shape: a mismatch at any level is a distinct, loud rejection.

## 3. No secrecy claim

An earlier revision carried `BindingOnly` / `Hiding` marker types. Nothing tied either marker to a commitment scheme, so a caller could name `Hiding` over a binding-only configuration and the sealed proof would report itself hiding. A promise that can be asserted falsely is worse than no promise, and there is no honest version of it today: no multilinear scheme this backend accepts hides, and the sumcheck and GKR messages are unmasked, so no proof framed here is zero-knowledge whatever the PCS. The markers are gone. The module documentation says so in one line, and the claim can come back when there is a zero-knowledge path to attach it to.

## 4. Fixtures and version policy

`multi-stark/tests/fixtures/backend_contract_v2.envelope` is a sealed proof of a fixed statement, following the existing postcard fixture pattern in this crate. `verify_the_compatibility_fixture` verifies it on every run; `generate_the_compatibility_fixture` is `#[ignore]`d and regenerates it.

The test pins three things: the digest of the fixture bytes, the body revision the fixture was written under, and the equality of that revision with `BODY_REVISION`. Regenerating the file therefore fails the digest assertion, and the failure message says to bump the revision and pin the new digest; the generator prints the digest to pin. `a_fixture_from_an_older_revision_is_refused` covers the path an out-of-date fixture takes.

The honest limit: a test cannot force a human to bump a constant, only make the regeneration visible in the diff and give the refusal path a test. The digest constant sits next to the revision it belongs to, so updating one without the other is a visible claim rather than a silent one.

The first bump happened in this PR and the policy worked unprompted. A bus field landed in `MultiStarkProof` between two existing ones, so the revision-one body stopped decoding; the digest and revision assertions passed and `check` failed with `Malformed`, which is exactly the signal. `BODY_REVISION` is 2, the revision-one file stays on disk with its own digest pinned, and `a_fixture_from_an_older_revision_is_refused` now reads that real file instead of editing a revision byte.

## 5. What one verification costs

The declaration-to-AIR check re-reads every constraint system on every verification, so its constant matters. Measured on `Blake3Air` (9168 columns, 9632 constraints), release, median of seven:

| | before | after |
| --- | --- | --- |
| read a declaration off a constraint system | 31.6 ms | 17.6 ms |
| the encoded string | 2 979 564 B | 527 105 B |
| one statement fingerprint | 4.8 ms, once per `run`, `seal` and `open` | 0 ms, computed once in `new` |

Three changes, and the string got *more* complete rather than less:

- the statement fingerprint is computed in `MachineDeclaration::new` and stored, instead of re-absorbing every constraint system in each of `run`, `seal` and `open`;
- the encoder names a repeated subtree by how far back it sits instead of writing an absolute position for every child, and writes its numbers seven bits at a time, which is where the 5.7x size drop comes from;
- the node table is a `hashbrown::HashMap`, since it is never iterated.

For scale: the verifier already runs four symbolic passes per AIR of its own (the indexed plan, the lookup plan, the zerocheck profile and the security report), about 15 ms for this table on the same machine. A `run` plus a `verify` went from roughly 41 ms per table of that size to 17.6 ms.

## 6. Ceilings

`MAX_POW_BITS` was 64, which admits difficulties the BabyBear challenger asserts against; it is now 30, the largest a thirty-one-bit transcript can sample. `MAX_PROOF_BYTES` was 1 GiB; the decoder can hold about twenty-four bytes per input byte, so that ceiling admitted tens of GiB of allocation. It is now 64 MiB and the expansion factor is written down.

A local interaction carrying no tuple is no longer counted. `LookupPlan::build` drops one as inert, so a declaration that counted it refused an honest proof with `SectionMismatch`. `a_lookup_that_carries_no_tuple_is_not_a_lookup` proves the round trip now verifies.

## Tests

`multi-stark/tests/backend_contract.rs` — 23 integration tests, 11 of which feed the reader bad input: short headers, wrong opening bytes, both wrong revisions, a flipped fingerprint bit, a `u32::MAX` length field, a length field one over and one under the real body, an appended byte, a body of `0xff` varint padding, an instance list at the wrong height, and a statement declaring a lookup the proof does not carry. Five more cover the new obligations: a statement whose table is three columns wide and one whose table agrees on every count but asserts a different transition are both refused by name; a target the configuration cannot reach is refused; a single-row height range is refused at declaration time; and a tuple-less lookup verifies end to end. Two more pin the injected-declaration cases: `a_cell_the_backend_pins_reaches_the_declaration` and `the_values_of_a_periodic_table_reach_the_declaration` both fail on the revision of this branch before them, with the two declarations printing the same `system_bytes: 374` and `count: 3`. Three more cover the bus route: `which_public_value_a_pinned_cell_names_reaches_the_declaration` pins two tables that list the *same* two cells against the other public value each, so it fails when the per-cell loop is deleted rather than passing on a count; `every_part_of_a_bus_route_reaches_the_declaration` separates push from pull, one channel from another, and one tuple expression from another, and shows a balancing batch no longer shares a fingerprint with a non-balancing one; `a_statement_declaring_a_bus_refuses_a_proof_without_one` covers the new section check.

Every rejection is asserted as a whole value, payload included, with the expected numbers written as literals derived by hand. A variant matched with `{ .. }` passes when the error reports the wrong numbers and when the two numbers are swapped, so none is left. The two assertions whose outer error wraps the verifier's own non-comparable error match the wrapper and compare the payload inside it. Checked by swapping the two fields of one version rejection: the unit test and the integration test both fail under the swap.

24 unit tests, each in the `#[cfg(test)] mod tests` of the file it exercises: header framing in `envelope.rs`, declaration and run construction in `machine.rs`, table validation and flush absorption in `table.rs`, and the length-prefixing injectivity in `digest.rs` — the property the whole fingerprint construction rests on, which nothing checked before. Two AIRs that agree on every count and differ only in what they assert now produce different fingerprints, which is asserted in `machine.rs` rather than assumed.

## Verification

`cargo test -p p3-multi-stark` 440 passed across every target. `cargo test -p p3-multi-stark --test backend_contract` 26 passed, 1 ignored. `cargo clippy -p p3-multi-stark --all-targets -- -D warnings` clean. `cargo +nightly fmt --all -- --check` clean. `cargo sort --check`, `taplo fmt --check`, `cargo machete` clean.

## Not done here, and why

`error.rs` sits at 35.7% comment share, above the 30% the doc gate asks for. That is inherent to a file that is nothing but enum variants and their field documentation; the four error files already in the tree measure 39.2%, 36.2%, 43.1% and 45.8%, so this one is the leanest of the five. Every other file in the module is between 5% and 27%.

The proof-size budget is a declared public parameter rather than a formula derived from the table shapes. A closed-form bound would have to model the commitment scheme's opening size, which is scheme-specific and not something this layer can derive honestly. The budget is measured, declared, and fingerprinted; it is not invented.

The header does not carry the per-table heights. A verifier already holds them, because it cannot verify without the instances and each instance carries its own height; `Run` is where it states them and `verify` cross-checks. Putting them in the header would let the prover pick the heights, and a prover-chosen height is a prover-chosen statement.


multi-stark/src/contract/machine.rs
```diff
@@ -0,0 +1,521 @@
...
+            + CanSampleUniformBits<C::Val>
+            + CanObserve<Commitment<C>>,
+        Commitment<C>: Clone,
+        A: VerifierAir<C::Val, C::Challenge>,
+    {
+        let accepted = self
+            .open::<C>(run, bytes)
+            .map_err(SealedVerificationError::Envelope)?;
+
+        if instances.len() != run.log_heights().len() {
```
Nashtare, line 343: This count and the per-instance heights are the only things tying the declaration to the instances. Column counts, public-value counts, constraint count and degree, flush channels and the lookup/indexed counts are never compared with the AIRs. Locally I declared a table with 7 committed columns, 0 public values and one degree-9 constraint (heights 2..=20), sealed the Fibonacci proof from the tests under it, and `declaration.verify` with the Fibonacci instances returned `Ok(())`. So "the two repositories cannot disagree about it" doesn't hold: they can, and nothing notices. `VerifierAir` already requires `Air<InteractionSymbolicBuilder<F, EF>>`. Could `verify` rebuild each table with `TableDeclaration::from_constraints` and refuse a mismatch in the parts the AIR determines?

multi-stark/src/contract/table.rs
```diff
@@ -0,0 +1,427 @@
...
+            DeclarationError::EmptyHeightRange {
+                table,
+                min: self.heights.min,
+                max: self.heights.max,
+            },
+        )
+    }
+
+    /// Absorb everything this table fixes into a statement fingerprint.
+    pub(super) fn absorb(&self, preimage: &mut Preimage) {
```
Nashtare, line 364: The fingerprint absorbs counts and the degree, not the constraints themselves (nor the field or the PCS parameters). Locally, the Fibonacci AIR and a copy whose transition is `next.right = right - left` give equal `TableDeclaration`s, so equal fingerprints. A proof sealed for one opens under the other and is refused only by the zerocheck (`FinalSumMismatch`). That is still sound because the inner verifier uses the real AIRs, but it isn't what "the fingerprint guards everything that changes the meaning of the bytes … the tables" describes. Could the table absorb a canonical encoding of its symbolic constraints and interaction expressions (`SymbolicExpr` already has serde impls)? Otherwise, could the docs and the description say the fingerprint names a shape?

multi-stark/src/contract/secrecy.rs
```diff
@@ -0,0 +1,106 @@
...
+    const SECRECY: Secrecy;
+}
+
+/// Binds its data without hiding it.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct BindingOnly;
+
+/// Binds its data and hides it.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Hiding;
```
Nashtare: Nothing ties `S` to the commitment scheme: `seal`, `open` and `verify` accept any `C`, and `Hiding` is a public unit struct. Locally, `MachineDeclaration::<Hiding, _>` over the binding-only WHIR config from the tests seals the proof, `secrecy()` reports `Secrecy::Hiding`, and `verify` accepts it. So `envelope.rs:103` ("A proof that only binds its trace cannot stand in for one that also hides it") doesn't hold. The caller can name either marker, and the fingerprint only makes both sides agree on the label. Separately, line 3 overstates it (`HidingWhirPcs` and `HidingFriPcs` exist). What is accurate: no PCS that multi-stark accepts hides, and the sumcheck/GKR messages aren't masked, so no multi-stark proof is zero-knowledge whatever the PCS. Could `Hiding` be removed until there is a ZK path? Or could `S` come from the config through a sealed per-PCS trait, so it can't be named for a binding scheme? The univariate `Pcs` trait already carries `const ZK: bool`.

multi-stark/src/contract/machine.rs
```diff
@@ -0,0 +1,521 @@
...
+        &self,
+        proof: &MultiStarkProof<C>,
+    ) -> Result<(), EnvelopeError> {
+        let section = |section, present: bool, declared: bool| {
+            (present == declared)
+                .then_some(())
+                .ok_or(EnvelopeError::SectionMismatch { section, present })
+        };
+
+        section("lookup", proof.lookup.is_some(), self.has_lookups())?;
```
Nashtare, line 391: The verifier already runs all four of these checks against the AIRs and the key: the preprocessed and indexed arms in `verifier.rs`, the presence match in `verify_lookup`, and `statement_from_claims`. This function takes the rule from the declaration instead, and the two already disagree. `from_constraints` counts every local interaction (`table.rs:226`), while `LookupPlan::build` drops one with no tuple as inert. Locally, the Fibonacci AIR plus `push_local_interaction(iter::empty())` gives `local_lookups: 1`. The prover emits no lookup section and plain `verify` accepts, but `declaration.verify` refuses the honest proof with `SectionMismatch { section: "lookup", present: false }`. Could `from_constraints` count only non-empty local interactions? Could `check_shape` also go, or be derived from the AIRs, so there is one rule?

multi-stark/src/contract/table.rs
```diff
@@ -0,0 +1,427 @@
...
+            "height exponent",
+            self.heights.max as usize,
+            MAX_LOG_HEIGHT as usize,
+        )?;
+
+        for flush in &self.flushes {
+            check("flush tuple width", flush.tuple_width, MAX_TUPLE_WIDTH)?;
+        }
+
+        (self.heights.min <= self.heights.max).then_some(()).ok_or(
```
Nashtare, line 354: There's no lower bound on `min`, so a range starting at 0 is accepted. Locally, `HeightRange::new(0, 20)` with `run(&[0])` passes `open` and the height check, then panics at `zerocheck/mod.rs:1173` ("zerocheck requires each trace height to be at least two"). Height 1 gives a clean error. The verifier picks the heights today, but for a machine whose heights come from the prover this range is the only guard. Could `validate` reject `min == 0`?

multi-stark/src/contract/run.rs
```diff
@@ -0,0 +1,48 @@
+//! The choices one proof makes within a statement it cannot change.
+
+use alloc::vec::Vec;
+
+/// What one proof picks, inside what its statement already allows.
+///
+/// The only way to build one is to pick heights a statement already allows.
+///
+/// A height outside the declared range therefore never reaches the prover or the verifier.
```
Nashtare: The prover never sees a `Run`: `prove` takes instances directly, and `seal` accepts any proof without looking at its heights or shape, so this holds for the verifier only. A related question: the header carries only a hash of the heights, and for a RISC-V machine the heights depend on the execution. How is the verifier expected to learn them? Could the header carry the per-table heights, bounded by the declaration, with `open` building the run from them?

multi-stark/src/contract/machine.rs
```diff
@@ -0,0 +1,521 @@
...
+            .iter()
+            .zip(run.log_heights())
+            .all(|(instance, &declared)| instance.num_variables() as u64 == u64::from(declared));
+        if !heights_agree {
+            return Err(SealedVerificationError::RunDisagreement {
+                what: "a table height",
+            });
+        }
+
+        verify(
```
Nashtare: This calls the plain `verify`, which the crate docs say enforces no minimum security level, and the declaration has no target. A downstream whose only entry point is this contract can't reach `verify_with_security`. Could the declaration carry a target, absorbed into the fingerprint, and route through `verify_with_security`? A second entry point would also work.

multi-stark/tests/backend_contract.rs
```diff
@@ -0,0 +1,537 @@
...
+        }
+    );
+}
+
+#[test]
+fn verify_the_compatibility_fixture() {
+    let bytes = std::fs::read(fixture_path()).expect(
+        "missing fixture; run: cargo test -p p3-multi-stark --test backend_contract -- --ignored",
+    );
+    check(&bytes).unwrap();
```
Nashtare, line 1068: `generate_the_compatibility_fixture` writes a header with whatever `BODY_REVISION` is at the time, and this test only checks that the file verifies. If someone changes the proof layout, regenerates the fixture and leaves `BODY_REVISION = 1`, it passes. That contradicts "Neither can be papered over by regenerating the fixture without also bumping a revision". The fixture also pins only one configuration, so a layout change in the binary PCS or `BooleanTracePcs` proofs doesn't touch it. Could the test pin a digest of the fixture bytes next to the revision it was written under? Old fixtures could then be kept, with a check that a fixture older than `BODY_REVISION` is refused with `BodyRevision`.

multi-stark/src/contract/envelope.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+/// A proof whose framing and shape already agree with a statement.
+///
+/// Holding one is the evidence that every check preceding transcript replay has passed.
+///
+/// It also carries the grinding difficulty the run fixed.
+///
+/// Verification cannot then run at a difficulty the proof was not sealed under.
```
Nashtare: Nothing consumes an `AcceptedProof`, so a caller can pass `accepted.proof()` to the free `verify` at any difficulty. Could it either get a verify method or be `pub(crate)`? The run's difficulty is also the verifier's own choice, so "never from the caller" in the description is stronger than it needs to be. On the same topic, `MAX_POW_BITS = 64` admits values the BabyBear challenger asserts against (`sample_bits` needs `1 << bits < p`).

multi-stark/src/contract/machine.rs
```diff
@@ -0,0 +1,521 @@
...
+use crate::verifier::verify;
+
+/// Largest number of tables one statement may declare.
+pub const MAX_TABLES: usize = 1 << 12;
+
+/// Largest grinding difficulty a run may request.
+pub const MAX_POW_BITS: u32 = 64;
+
+/// Hard ceiling on any declared proof-size budget, in bytes.
+pub const MAX_PROOF_BYTES: usize = 1 << 30;
```
Nashtare: Nit: allocation is linear in the budget, but the factor is about 24. An empty inner vector in any `Vec<Vec<EF>>` field (`round_polys`, `reader_claims`, `pushforwards`, `column_claims`) costs one input byte and 24 bytes of memory, so a body at this ceiling can ask for tens of GiB. Could the doc state the factor, or could the ceiling come down?

Nashtare (changes requested):
I went through the envelope against the threat-model table, and most of it holds. The body length is checked against the budget before anything is sliced, the input length has to match it exactly, and the `take_from_bytes` remainder check and `TrailingBytes` are both reachable and distinct. On allocation: with postcard 1.1.3, `SeqAccess::size_hint` returns `None` when the declared length is longer than the remaining input, and serde 1.0.228 caps the hint at 1 MiB per sequence, so a sequence that runs out of input pre-allocates nothing. I found no panic in `open`. The digest grammar is prefix-free (every list has a count, every name a length), so I don't see a collision there. The round-count and witness-count checks are in `GenericDegreeProof::verify` where the description says, though they run after the commitment, the public values and the lookup GKR have been replayed. Between the first and second commit the fixture changes in bytes 12..44 only.

The underlying verifier is still sound: it decides everything from the caller's AIRs, heights and public values. The trouble is that the contract layer promises more than it checks. The numbers inline come from a local probe on the tests' WHIR/BabyBear setup.

- `verify` never compares the declaration with the AIRs it verifies against, and the fingerprint covers counts, not constraints. A 7-column declaration verifies the Fibonacci proof, and changing a Fibonacci constraint leaves the declaration unchanged (inline, `machine.rs` and `table.rs`).
- `Hiding` can be named for the binding-only WHIR config, and the proof seals and verifies under it (inline, `secrecy.rs`).
- `check_shape` rejects an honest proof whenever the AIR pushes a local interaction with no tuple (inline, `machine.rs`).

**Smaller things**
- The fixture can be regenerated without bumping `BODY_REVISION` and the test still passes, which contradicts section 4 of the description (inline, `backend_contract.rs`).
- Nothing outside the tests uses the module yet, and `multi-stark` becomes the first library crate with a runtime `postcard` dependency, so postcard becomes the library's wire format. That seems fine to me, but the description should say so.

The PR merges cleanly into current main.

multi-stark/src/contract/machine.rs
tcoratger: Reproduced exactly: your declaration (7 committed, 0 public, one degree-9 constraint, heights 2..=20) sealed the Fibonacci proof and `declaration.verify` returned `Ok(())`.

Fixed on both sides.

The hand-built constructors are gone. `TableDeclaration::new`, `with_flushes`, `with_local_lookups` and `with_indexed` are now `pub(crate)` and only the test modules reach them; `from_constraints` is the only public way to obtain a declaration. Your repro no longer compiles, which is the stronger half of the fix.

`MachineDeclaration::verify` now does what you proposed: for each table it rebuilds the declaration with `from_constraints` from the instance's own AIR, at the declared height range, and refuses the first part that differs with `SealedVerificationError::AirDisagreement { table, what }`. The parts compared are the three column counts, the constraint count and degree, the flush list, the local-lookup count, the indexed read and table counts, and the canonical encoding of the constraints themselves (see the reply on `table.rs:323`). The height range is the only part left out, because no constraint system fixes it. `VerifierInstance` gained a `pub const fn air()` so the check can reach the AIR.

`a_statement_that_describes_another_table_is_refused` covers both a table of the wrong width ("the committed columns") and one that agrees on every count but asserts a different transition ("the constraints themselves").

The cost is one extra symbolic pass per table per verification, on top of the several the verifier already runs for the indexed plan, the lookup plan and the security report.

multi-stark/src/contract/table.rs
tcoratger: Reproduced: `from_constraints` on the Fibonacci AIR and on a copy whose transition subtracts gave `TableDeclaration`s that compared equal.

Taken the first option. A table now carries a canonical byte string naming its whole constraint system, built in `contract/constraints.rs`, and the fingerprint absorbs it behind its length.

Not via serde, for two reasons. `SymbolicExpr` does have `Serialize` through `air/src/symbolic/flatten.rs`, but it needs `F: Serialize`, which `Field` does not give and `MultiStarkConfig::Val` does not carry either, so `verify` could not call it generically. The walker uses `RawDataSerializable::into_bytes`, which every `Field` has.

What it absorbs: the field order and the extension's algebra identity first, then the base constraints, the extension constraints, each global interaction (bus name, weight, payload expressions, count expression), each local interaction tuple with its weight, and the indexed reads and tables. Expressions are written as a post-order arena with children named by position, the same shape as the crate's existing flattened wire form, so a shared subtree costs one node instead of expanding. Mutually exclusive interactions are left out because `LookupPlan::build` asserts against them, so they cannot appear in a provable statement.

What stays outside, and why: the PCS parameters. They belong to the verifier's configuration rather than to the statement, and the honest way to bind their effect is the security target the declaration now carries (see the reply on `machine.rs:311`), which is computed from that configuration.

`a_run_of_one_statement_is_refused_by_another` in `machine.rs` now builds two AIRs that agree on every count and differ only in what they assert, asserts that agreement, and then asserts the fingerprints differ.

multi-stark/src/contract/secrecy.rs
tcoratger: Reproduced: `MachineDeclaration::<Hiding, _>` over the binding-only test config sealed the proof, `secrecy()` reported `Hiding`, and `verify` accepted.

Removed. `secrecy.rs` is deleted, `MachineDeclaration`, `SealedProof` and `AcceptedProof` lose the `S` parameter, and the statement digest no longer absorbs a promise byte.

I took removal over the sealed per-PCS trait because the trait would still be a false promise. Your own sentence is the reason: the multilinear PCS trait has no `ZK` flag to route, adding one would reach into whir and binary-pcs, and even a hiding scheme would not make these proofs zero-knowledge while the sumcheck and GKR messages travel unmasked. So there is nothing true for the marker to name yet.

The overstated line in the module doc is gone with the file. What `contract/mod.rs` says now is one line: nothing this backend accepts hides anything, so no proof framed here is zero-knowledge. The claim can come back when there is a path to attach it to.

`the_promise_is_part_of_the_type` is replaced by `the_security_target_is_part_of_the_statement`, which makes the same point about a parameter that is real.

multi-stark/src/contract/table.rs
tcoratger: Reproduced by inspection and by test: `HeightRange::new(0, 20)` passed `validate`, `run(&[0])` returned a run, and `zerocheck/mod.rs:1173` asserts `log_height > 0`.

`validate` now rejects it. `MIN_LOG_HEIGHT = 1` is public next to the other ceilings, and the error is `DeclarationError::HeightBelowFloor { table, min, floor }`, refused at declaration time before any run exists. Covered by a unit test in `table.rs` and by `a_height_range_reaching_a_single_row_is_refused` in the integration test, which asserts the whole variant.

You are right that this is the only guard once the heights come from the prover rather than from the verifier, which is why it belongs here rather than in a debug assertion further down.

multi-stark/src/contract/machine.rs
tcoratger: Correct, and taken as proposed. `MachineDeclaration::new` now takes a fourth argument, the security target in bits, refused outside `1..=MAX_SECURITY_BITS` with `DeclarationError::SecurityOutOfRange`. It is absorbed into the statement digest, and `verify` routes through `verify_with_security` with it.

I made it required rather than optional. A contract whose target can be omitted has the defect you describe by default, and there is no sensible value for "no target".

Two tests. `the_security_target_is_part_of_the_statement` shows a proof sealed under one target does not open under another. `a_target_the_statement_cannot_reach_is_refused` declares 256 bits over the test configuration and asserts `SecurityError::InsufficientSecurity { requested: 256, .. }`, so the routing is exercised rather than assumed.

The test configuration had to grow a `collision_resistance_bits` override, since the default supplies no evidence and every security-checked path fails closed without it. It reports 100, matching the other whir tests in this crate, and the declared target is 20.

multi-stark/src/contract/machine.rs
tcoratger: The counting bug is real and reproduced: `FibAir` plus `push_local_interaction(iter::empty())` gave `local_lookups: 1`, the prover emitted no lookup section, plain `verify` accepted, and `declaration.open` refused the honest proof with `SectionMismatch { section: "lookup", present: false }`.

`from_constraints` now counts only local interactions with at least one tuple, which is exactly `LookupPlan::build`'s own activity test (`Lookups::from_air` gives a tuple-less local interaction an empty `elements`, and the `active` filter drops it). Global interactions never hit this: `from_interactions` gives them `elements: vec![fields]`, which is never empty. `a_lookup_that_carries_no_tuple_is_not_a_lookup` proves the whole round trip now verifies.

On the duplication: it is the point, but only now that it cannot drift. `check_shape` runs inside `open`, which has the declaration and no AIRs, so a downstream that only wants to read a byte string gets the shape check without building a verifier. `verify` now rebuilds every table from the AIRs first (see the reply on `machine.rs:296`), so the declaration is provably the AIRs' own answer by the time `check_shape` uses it, and the two rules cannot disagree the way they did. I kept the function and said that in its documentation rather than deleting a check that is the only one available on the `open` path.

multi-stark/tests/backend_contract.rs
tcoratger: Right, and the claim in the description was wrong. Regenerating the fixture with `BODY_REVISION` unchanged did pass.

`verify_the_compatibility_fixture` now pins three things: the Keccak digest of the fixture bytes, the body revision the fixture was written under, and the equality of that revision with `BODY_REVISION`. A layout change moves the bytes, the digest assertion fails, and its message says to regenerate, bump the revision and pin the new digest; the generator prints the digest to pin. `a_fixture_from_an_older_revision_is_refused` covers the refusal path, asserting the whole `BodyRevision` variant.

The limit I will state rather than paper over: a test cannot force a human to bump a constant. What it can do is make the regeneration visible in the diff and put the digest constant next to the revision it belongs to, so updating one without the other is a visible claim instead of a silent one. Keeping older fixtures becomes worthwhile at the first bump; there is only one revision so far, so there is nothing yet to keep.

The single-configuration point stands and I have not fixed it. The binary PCS and `BooleanTracePcs` proof layouts are untouched by this fixture. That wants its own fixture per configuration and a shared harness, which is more than this PR should carry; I would rather it be a follow-up than a half-built second fixture here.

multi-stark/src/contract/envelope.rs
tcoratger: Both taken.

`AcceptedProof::proof()` is now `pub(super)`. Nothing outside the crate can reach the decoded proof, so the escape hatch to the free `verify` at an arbitrary difficulty is gone. The type keeps `pow_bits()` and its `Debug`, and stays the evidence that every pre-replay check passed.

The description line is corrected: the run's difficulty is the verifier's own choice, and the table row now says the decoded proof never leaves the crate rather than claiming the difficulty could never come from the caller.

`MAX_POW_BITS` is 30, not 64. You are right about `sample_bits`: `DuplexChallenger` asserts `(1 << bits) < F::ORDER_U64`, so BabyBear refuses 31. Every prime field this backend proves over has more than 2^30 elements, so 30 is the largest ceiling that is honest for all of them, and it is already far past any reachable grinding cost.

multi-stark/src/contract/machine.rs
tcoratger: Both, since neither is expensive.

`MAX_PROOF_BYTES` is now 1 << 26 rather than 1 << 30, and the constant's documentation states the factor: about twenty-four bytes of memory per input byte, because an empty inner list costs one input byte and a pointer triple to keep. At the new ceiling the worst case is around 1.5 GiB instead of tens of GiB. The fixture budget is 64 KiB, well under it, and nothing in the crate needed the old headroom.

multi-stark/src/contract/run.rs
tcoratger: Correct on both counts, and the documentation was wrong.

`run.rs` now says the run is what the verifier runs at, and says explicitly that the prover never sees one, so it bounds what is accepted rather than what is produced. `seal` still accepts any proof: it frames bytes against a statement and is not a place where a claim can be checked.

On the heights, I am declining the header change, with a reason rather than a shrug. The verifier already holds them. It cannot verify without the instances, and every `VerifierInstance` carries its own `num_variables`; `run` is where the verifier states them, and `verify` compares the two and refuses on `RunDisagreement`. Moving them into the header would let the prover choose them, and a prover-chosen height is a prover-chosen statement. It would also mean either widening the fixed 48-byte header or making it variable-length, which is the part of the framing I least want to make attacker-sized.

For a RISC-V machine the heights come out of the execution, and the verifier learns them the same way it learns the public values: alongside the proof, as part of the statement it chooses to check. The declaration bounds them, `run` refuses one outside the range, and the fingerprint commits to the ones chosen. What the header carries is a commitment to that choice, not the choice itself.

Nashtare (commented):
I re-checked the fix commit against the earlier threads, and re-ran the earlier cases locally on the same WHIR/BabyBear setup. The Fibonacci proof sealed under the `r - l` declaration now fails with `AirDisagreement { table: 0, what: "the constraints themselves" }`, and a hand-built declaration can no longer be written.

These close the other earlier threads:
- removing `Hiding`;
- the tuple filter (it matches `LookupPlan::build`, and the preprocessed and indexed rules in `check_shape` match `setup` and `IndexedPlan`);
- the height floor;
- the target routed through `verify_with_security`;
- the two ceilings, and `AcceptedProof::proof` going `pub(super)`.

The pinned fixture digest is a real improvement. The re-sealed fixture differs only in its 32 fingerprint bytes, so `BODY_REVISION = 1` is right. The address-keyed dedup in `constraints.rs` looks safe: every visited node stays alive for the whole walk, and the map is never iterated. The PR merges cleanly into current main, and `cargo check -p p3-multi-stark --all-targets` is clean on the merged tree.

- The constraint string covers only what `eval` emits. Two AIRs that differ in `public_boundary_io` get one declaration and one fingerprint, even though they accept different public inputs. Periodic tables and the preprocessed trace are also left out (inline, `constraints.rs`).

**Smaller things**
- The per-verify AIR comparison costs about 26 ms per Blake3-sized table, not one symbolic pass (inline, `machine.rs`).
- The encoding isn't prefix-free by construction:
  - `00 01` is both `IsFirstRow` and the start of an `ExtVariable`;
  - the base and extension operations share tags;
  - a root's node run has no length.

  I found no collision. Distinct extension-leaf tags and a node count per root would make it injective by construction, like `digest.rs`.
- `open` (and so `check_shape`) runs before the `disagreement` loop. So a declaration read off another AIR can still fail with `SectionMismatch` first. That's fine, but it isn't the order the reply on the `check_shape` thread describes.
- Description: "never reaches the prover or the verifier" and "capped at 1 GiB regardless" are stale. The earlier postcard wire-format note isn't there.

multi-stark/src/contract/constraints.rs
```diff
@@ -0,0 +1,254 @@
+//! One byte string naming everything a constraint system asserts.
+//!
+//! Two systems that differ anywhere a prover can observe produce different strings.
```
Nashtare: `encode` walks the `InteractionSymbolicBuilder`, which only runs `Air::eval`. So the public boundary cells, the periodic tables and the preprocessed trace never reach this string, although the multi-STARK proves against all three. `air.rs` says anything that scores constraints from a symbolic pass has to add the injected pins back. `constraints.count` doesn't add them back either.

Locally, I compared a Fibonacci AIR without first-row asserts against a copy that pins `(0, First, 0)` and `(1, First, 1)` through `public_boundary_io`:
- the two give identical `TableDeclaration`s (`system_bytes: 374`, `count: 3`) and one fingerprint;
- a run of one seals under the other, and the unpinned declaration verifies the pinned AIR's honest proof;
- with public values `(5, 7, x)`, the unpinned AIR accepts and the pinned one rejects (`FinalSumMismatch`).

Two periodic tables (`[1, 2]` vs `[3, 4]`) or two preprocessed traces under one `eval` also give equal declarations. Soundness is fine, since the verifier uses the instance's own AIR, but this line and the description's "leaves out: the PCS parameters" overstate it. Could `encode` also absorb `public_boundary_io()` and the periodic tables, and could the pins be counted? Could the docs also say that the verifying key binds the preprocessed trace, not the fingerprint?

multi-stark/src/contract/machine.rs
```diff
@@ -0,0 +1,645 @@
...
+            .all(|(instance, &declared)| instance.num_variables() as u64 == u64::from(declared));
+        if !heights_agree {
+            return Err(SealedVerificationError::RunDisagreement {
+                what: "a table height",
+            });
+        }
+
+        // Nothing above reads a constraint system, so this is where the two sides meet.
+        for (table, (declared, instance)) in self.tables.iter().zip(instances.iter()).enumerate() {
+            if let Some(what) = declared.disagreement::<C::Val, C::Challenge, A>(instance.air()) {
```
Nashtare, line 360: This re-runs `from_constraints` for every table on every verification: the symbolic pass, `poly_degree` over every constraint, and the full encoding with a `BTreeMap` insert per node. `statement_digest` then copies every `system` into a fresh preimage and hashes it, once in each of `run`, `seal` and `open`.

I timed it locally on `Blake3Air` (9168 columns, 9632 constraints):
- `from_constraints`: 26 ms, against 2.0-2.3 ms for the plain symbolic pass;
- the string: 2.98 MB;
- one statement digest: 3.6 ms.

So a `run` plus a `verify` add about 33 ms per table of that size, roughly 11 times the "one extra symbolic pass" in the reply. Could the statement digest be computed once in `new`? Could the comparison stream the encoding into a hasher and compare digests, instead of building and comparing megabytes? A hash map for `written` would also be fine, since it's never iterated.

multi-stark/src/contract/constraints.rs
tcoratger: Reproduced both cases exactly, including your numbers. I kept the repros as tests and checked they fail on the previous revision of this branch:

```
a_cell_the_backend_pins_reaches_the_declaration
  left:  LocalConstraints { count: 3, degree: 2 }
  right: LocalConstraints { count: 5, degree: 2 }

the_values_of_a_periodic_table_reach_the_declaration
  left:  TableDeclaration { ..., system_bytes: 374 }
  right: TableDeclaration { ..., system_bytes: 374 }
```

`system_bytes: 374` and `count: 3` on both sides, as you saw. Both pass now.

`encode` takes the AIR as well as the builder, and writes what a symbolic pass never sees: `public_boundary_io()`, `num_periodic_columns()` and the values in `periodic_columns()`, `main_next_row_columns()` and `preprocessed_next_row_columns()`, and `max_constraint_degree()`. I added the last three because they are the same class as the first two: the backend reads them, `Air::eval` never mentions them, and the first two reach `MultiStarkShape`, so two tables differing there do not even share a transcript.

`constraints.count` now adds `boundary_io_pins(...).count`, and `constraints.degree` is scored the way `get_air_profile` scores it: `max_constraint_degree().unwrap_or(symbolic).max(symbolic)` when the evaluation writes anything, then raised to the pin degree. So the two numbers a declaration reports are the two the reduction will use, not the two a symbolic pass happens to see.

The preprocessed trace stays out, and both docs now say why rather than leaving it unsaid: the verifying key commits to it and an opening against that commitment is what binds it, so putting the contents in the fingerprint would shadow a stronger binding with a weaker one. The description's "leaves out" paragraph is rewritten to list both exclusions with their reasons.

`num_constraints()` is the one hook I left alone: nothing in `multi-stark` reads it.

multi-stark/src/contract/machine.rs
tcoratger: Reproduced. Same shape on my machine, about 1.3x slower across the board, and the string length is identical to yours:

```
Blake3Air, 9168 columns, release, median of seven
  symbolic pass          4.7 ms      (you: 2.0-2.3)
  from_constraints      31.6 ms      (you: 26)
  string             2 979 564 B     (you: 2.98 MB)
  one statement digest   4.8 ms      (you: 3.6)
```

Fixed together with the coverage gap above, since that one pulls the other way. After:

```
  from_constraints      17.6 ms
  string               527 105 B
  one statement digest   0.0 ms
```

Three changes, in order of what they bought:

- the statement fingerprint is computed once in `new` and stored. `run`, `seal` and `open` read it. That is the 4.8 ms x 3 gone outright, and it is why `run` now measures 0.
- the encoder no longer writes an absolute `u64` position for each child. A node emits its children first and then its own opcode, and a repeated subtree emits a reference to how far back its value sits; every number is a varint. That took 17 bytes per node to about 3, which is the 5.7x size drop, and most of the time drop with it.
- `written` is a `hashbrown::HashMap`, as you suggested.

I did not go to comparing digests instead of bytes. Comparing 527 KB is a memcmp and does not show up; what costs is building the string at all, and `verify` has to build the AIR side either way. Storing a digest per table instead of the bytes would need a hash inside `TableDeclaration`, which means either hard-coding one under the swappable one or passing a second hasher that can silently disagree with the machine's. Neither trade looked worth 527 KB. Say the word if you would rather have the digest and I will take the extra parameter.

What is left is one symbolic pass, one `poly_degree` sweep and one encoding pass per table per verification: 3.8 + 3.4 + 10.4 ms on that table. For scale, the verifier already runs four symbolic passes per AIR of its own, about 15 ms here. A `run` plus a `verify` went from roughly 41 ms to 17.6 ms per table of that size, while the string covers strictly more than it did.

multi-stark/src/contract/constraints.rs
```diff
@@ -0,0 +1,289 @@
...
+///
+/// - the cells the backend pins to public values;
+/// - the periodic tables;
+/// - the windows opened at the next row;
+/// - the degree hint.
+///
+/// The contents of a fixed trace stay out, because the verifying key is what binds those.
+///
+/// Mutually exclusive interactions are left out, because this backend refuses them outright.
+pub(super) fn encode<F, EF, A>(air: &A, builder: &InteractionSymbolicBuilder<F, EF>) -> Vec<u8>
```
Nashtare: On current main this misses the bus route from #2244. `InteractionSymbolicBuilder::record_bus_interaction` (`bus/src/builder.rs:408` on main) throws away the bus name, the direction, the fields and the activation. Only the Booleanity check of a conditional activation reaches `base_constraints`.

Locally, on the merged tree, I took a width-2 table that pushes `value²` on `"squares"` behind a Boolean selector, and three variants: one that pulls, one on `"memory"`, and one that pushes `value³`.
- All four give identical `TableDeclaration`s (`count: 1`, `system_bytes: 60`, `flushes: []`).
- `[push, pull]` and `[push, push]` share one statement digest.
- The honest `[push, pull]` proof verifies through the contract.
- The same bytes under `[push, push]` pass `open` and every AIR comparison. They fail only inside the verifier, with `BusArgument(Product(LayerCountMismatch { expected: 2, actual: 1 }))`.

Main's own tests call direction and activation "verifier statement metadata". `VerifierAir` already requires `Air<BusSymbolicBuilder<F, EF>>`. Could `encode` walk that builder too, and write each `SymbolicBusInteraction`? This will matter more with #2294: its `BusActivation::Boundary` leaves the zerocheck untouched, so not even the Booleanity check would show.

multi-stark/src/contract/envelope.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Revision of the framing itself.
+///
+/// Bump it whenever the header layout changes.
+pub const ENVELOPE_VERSION: u16 = 1;
+
+/// Revision of the encoded body.
+///
+/// Bump it whenever the serialized form of a proof changes in any way.
+pub const BODY_REVISION: u16 = 1;
```
Nashtare: This has to become 2 in the rebase. #2244 put `bus: Option<BusProof<..>>` into `MultiStarkProof` between `indexed` and `sumcheck`, so a body written before it no longer decodes. On the merged tree, `verify_the_compatibility_fixture` passes the digest and revision asserts, then fails at `check(&bytes)` with `Envelope(Malformed)`. So the policy did its job. The fixture needs regenerating under revision 2, with the new digest pinned. By your reply on the fixture thread, this first bump is also where the revision-1 file becomes worth keeping, refused with `BodyRevision`.

multi-stark/tests/backend_contract.rs
```diff
@@ -0,0 +1,950 @@
...
+        }
+    );
+    assert_eq!(
+        pinned.constraints(),
+        LocalConstraints {
+            count: 5,
+            degree: 2,
+        }
+    );
+    assert_ne!(unpinned, pinned);
```
Nashtare, line 701: The two tables differ in how many pins they have (count 3 vs 5), so this passes even if the pinned cells never reach the string. Locally I deleted the per-cell loop in `encode` (`constraints.rs:42-49`) and kept the count. All 23 tests here and the 24 contract unit tests still pass, and two tables with swapped public indices then compare equal. Could this add a pair with the same number of pins, such as `(0, First, 1), (1, First, 0)` against `PINS`?

multi-stark/src/contract/constraints.rs
```diff
@@ -0,0 +1,289 @@
...
+/// The two fields come first, so a table read over one pair never matches another.
+///
+/// A symbolic pass runs the evaluation alone, so what a table declares beside it goes in here:
+///
+/// - the cells the backend pins to public values;
+/// - the periodic tables;
+/// - the windows opened at the next row;
+/// - the degree hint.
+///
+/// The contents of a fixed trace stay out, because the verifying key is what binds those.
```
Nashtare, line 28: Nit: a prover does observe the fixed trace, so leaving it out here makes line 3 ("differ anywhere a prover can observe") promise more than `encode` does. Could line 3 say "anywhere the verifying key does not already bind"?

Nashtare (changes requested):
I re-ran last round's cases on db6d33f5c with the tests' WHIR/BabyBear setup.

The round-2 gap is closed. The unpinned and pinned Fibonacci tables now differ (`count` 3 vs 5). The unpinned declaration against the pinned instance gives `AirDisagreement { table: 0, what: "the constraint count" }`, the pinned honest proof verifies, and the `(5, 7, x)` proof is refused. Swapping the public indices of the two pins also changes the table, and so do the periodic pair `[1, 2]` / `[3, 4]` and a change to `main_next_row_columns`. The fixed trace stays out, and the key does bind it: A's proof checked against B's instance and key fails with `Opening(MerkleProofInvalid { .. })`.

The cache holds up. `new` computes the digest from tables that can't change afterwards, and `verify` still reads every AIR back through `disagreement`. On `Blake3Air` I get 15-20 ms per `from_constraints`, a 527,105-byte string and under 1 µs per `run`, as the description says. The three blocking threads from the first round are still closed.

Against current main, the only conflict is the two `pub use` lines in `lib.rs`, and the merged tree type-checks for `p3-multi-stark` with all targets. Two things break, though:
- `verify_the_compatibility_fixture` fails with `Envelope(Malformed)`, because #2244 added a `bus` section to `MultiStarkProof` (inline, `envelope.rs`).
- #2244's bus declarations never reach the declaration or the fingerprint, which is the round-2 gap again on a second route (inline, `constraints.rs`). `check_shape` has no arm for the `bus` section either, though the verifier still checks it.

**Smaller things**
- `a_cell_the_backend_pins_reaches_the_declaration` still passes with the per-cell loop deleted (inline, `backend_contract.rs`).
- The new tags remove last round's leaf overlaps. A root still has no length, though, so the string isn't injective by construction. I haven't found a collision.
- Description: "never reaches the prover or the verifier" and "capped at 1 GiB regardless" are still stale, and the postcard note is still missing.

multi-stark/src/contract/constraints.rs
tcoratger: Reproduced on the rebased tree, with your numbers:

```
push   TableDeclaration { ..., constraints: { count: 1, degree: 2 }, flushes: [], system_bytes: 60 }
pull   TableDeclaration { ..., constraints: { count: 1, degree: 2 }, flushes: [], system_bytes: 60 }
memory TableDeclaration { ..., constraints: { count: 1, degree: 2 }, flushes: [], system_bytes: 60 }
cube   TableDeclaration { ..., constraints: { count: 1, degree: 2 }, flushes: [], system_bytes: 60 }
push == pull   true
push == memory true
push == cube   true
[push, pull] == [push, push] true      (statement digests)
```

Fixed the way you suggested. `from_constraints` now runs `BusSymbolicBuilder` alongside the interaction builder and `encode` writes each `SymbolicBusInteraction`: the channel name, a direction byte, the tuple expressions, and the activation, with `Boolean` carrying its selector expression. After:

```
push   system_bytes: 85
pull   system_bytes: 85
memory system_bytes: 84
cube   system_bytes: 90
push == pull   false
push == memory false
push == cube   false
[push, pull] == [push, push] false
```

`every_part_of_a_bus_route_reaches_the_declaration` keeps all four variants and the batch pair. It fails if I replace the direction byte with a constant, so the direction is genuinely load-bearing rather than incidentally covered by the name.

Two things I did beyond the walker, because the same gap ran through them:

- `TableDeclaration` carries a bus declaration count. It is absorbed, it is one of the parts `verify` compares against the AIR (reported as "the bus declarations"), and it has the flush ceiling.
- `check_shape` gained a `"bus"` section. `proof.bus` was the one optional part the reader did not check, so a statement that moves a tuple across a bus used to accept a proof carrying no bus section. `BusPlan::build` returns `None` exactly when the total declaration count is zero and treats an empty tuple as an error rather than as inert, so there is no repeat of the tuple-less-lookup case here. Covered by `a_statement_declaring_a_bus_refuses_a_proof_without_one`.

The cost is a second symbolic pass per table in `from_constraints`, since the bus declarations exist only in that builder. On the `Blake3Air` measurement from the thread above that is one more pass of about 4 ms.

Your `#2294` point is the reason I wrote the activation out in full rather than leaning on the Booleanity check being there: a `Boundary` activation leaves the zerocheck untouched, and the string does not depend on the zerocheck seeing anything.

multi-stark/src/contract/envelope.rs
tcoratger: Confirmed on the rebase, exactly as you describe: the digest and revision assertions pass, then `check(&bytes)` fails with `Envelope(Malformed)`. The policy fired on its own, which is the first evidence it does anything.

`BODY_REVISION` is 2 and `backend_contract_v2.envelope` is generated and pinned.

And yes, this is where keeping the old file starts paying. `backend_contract_v1.envelope` stays on disk with `RETIRED_REVISION` and `RETIRED_DIGEST` next to it, and `a_fixture_from_an_older_revision_is_refused` reads that real file rather than editing a revision byte into the current one. It asserts the retired digest, the revision on disk, and then the whole `BodyRevision { found: 1, expected: 2 }`. So the refusal is covered by a body that genuinely predates the change, not by a synthetic one.

multi-stark/tests/backend_contract.rs
tcoratger: Confirmed: I deleted the per-cell loop and kept the count, and all 23 integration tests plus the 24 unit tests passed. The test was reading a count, not the cells.

Added the pair you suggested. `SwappedPinFibAir` lists `(0, First, 1)` and `(1, First, 0)` against `PINS`' `(0, First, 0)` and `(1, First, 1)`: two cells either way, same columns, same end, the other public value each. `which_public_value_a_pinned_cell_names_reaches_the_declaration` asserts the two constraint records are equal and the declarations are not.

Re-ran your mutation against it: with `constraints.rs` lines 42-49 deleted, that test is the one that fails, and only that one. So it now bites where the old pair did not.

multi-stark/src/contract/constraints.rs
tcoratger: Taken, with your wording. The module line now reads "Two tables differing anywhere the verifying key does not already bind produce different strings."

You are right that the old line promised more than `encode` delivers. The fixed trace is observable to a prover and is deliberately absent from the string, so "anywhere a prover can observe" was the wrong boundary; the verifying key is the right one, because that is precisely what makes the omission safe rather than a hole.

multi-stark/src/contract/constraints.rs
```diff
@@ -0,0 +1,316 @@
...
+        arena.text(&declaration.bus_name);
+        arena.byte(match declaration.direction {
+            BusDirection::Push => 0,
+            BusDirection::Pull => 1,
+        });
+        arena.count(declaration.fields.len());
+        for field in &declaration.fields {
+            arena.expression(field);
+        }
+        match &declaration.activation {
```
Nashtare, line 137: Heads-up for #2294: merged with its current head (`41b49bcc0`), this match fails with E0004 (`BusActivation::Boundary(_)` not covered), and nothing else breaks. So whichever of the two lands second gets a loud error, which is what we want. The `Boundary` arm will be the first activation that leaves nothing in the zerocheck, so it needs to write its end byte. It will also need a test with two tables that differ only in `First`/`Last`, because right now nothing fails when the activation payload is dropped.

Nashtare (approved):
I re-checked a1570598 against last round's points, and I ran it merged with current main (`d87e86e84`). The merge is clean, `p3-multi-stark` type-checks with all targets, and all 26 `backend_contract` tests pass on the merged tree.

All four points from last round are closed:
- **Revision bump.** `BODY_REVISION` is 2. `verify_the_compatibility_fixture` passes on the merged tree, and the retired v1 file is refused with `BodyRevision { found: 1, expected: 2 }`.
- **Bus route.** I reran my probe. Push, pull, `"memory"` and `value³` now give four distinct declarations. `[push, pull]` and `[push, push]` no longer share a statement digest. The honest `[push, pull]` proof still verifies. The same bytes under `[push, push]` now stop in `open` with `Envelope(RunMismatch)`. The balanced statement checked against `[push, push]` instances gives `AirDisagreement { table: 1, what: "the constraints themselves" }`, so the refusal no longer depends on the inner verifier. `BusPlan::build` returns `None` exactly when there are no declarations, so the new `bus` arm in `check_shape` matches what the verifier expects.
- **Pins test.** With last round's mutation (the per-cell loop deleted), `which_public_value_a_pinned_cell_names_reaches_the_declaration` now fails.
- **Doc line.** The module doc now says "anywhere the verifying key does not already bind".

I also mutated the new code. Collapsing the direction byte, dropping the bus name or dropping the tuple expressions each fails `every_part_of_a_bus_route_reaches_the_declaration`, and deleting the `bus` arm fails `a_statement_declaring_a_bus_refuses_a_proof_without_one`. Dropping the selector expression passes every test. That's fine today, because `RecordToken` means every `Boolean` selector also reaches the zerocheck through its Booleanity check. The second symbolic pass costs little: on `Blake3Air` I get 17-25 ms per `from_constraints` (527,106 bytes), against 15-20 ms last round.

**Smaller things**
- The description still has "never reaches the prover or the verifier", and it still says "capped at 1 GiB regardless", which contradicts the 64 MiB paragraph further down. The "Counts that disagree with the declared shape" paragraph doesn't list the new `bus` section, and "23 integration tests" is now 26.
- The expression encoding still has no root length, as noted last round. I haven't found a collision.

PR #2273, opened 2026-09-20
feat(binary-pcs): open Boolean and small-field traces through additive-domain WHIR

### LINKED ISSUE #2232: Integrate binary WHIR with Boolean and small-field traces
## Deliverable

Connect the additive-domain WHIR implementation to the binary PCS and multi-STARK path without changing the prime-field fast path.

## Acceptance criteria

- Binary WHIR passes its protocol and security review.
- Boolean and small-field commitments remain narrow through encoding and commitment.
- Ring-switched claims are discharged through additive-domain WHIR.
- Supported rate, query, and grinding profiles are derived from the security report.
- Unique-decoding, proven-list, and conjectural regimes remain explicit.
- Prime-field proof bytes and performance distributions do not regress.
- A production proof-size and verification-time budget is enforced.

PR #2198 implements the additive-domain WHIR foundation but not the multi-STARK adapter.

tcoratger:
Closes #2232

WHIR over the additive binary domain existed after #2198, but nothing could feed it a narrow trace. The domain was fixed to one 64-bit polynomial-basis alphabet, and the bit commitment had only the folding-only opening path. This makes WHIR a drop-in opening path for a Boolean or small-field commitment, with the alphabet kept narrow all the way through encoding and commitment.

## The adapter

`BooleanWhirPcs` implements the existing `BooleanMultilinearPcs` trait, so it is interchangeable with the folding-only `BooleanPcs` at the same call sites. It composes the same three pieces:

```
packing    bits  ->  the elements a commitment holds, by reinterpretation
switch     a claim about the bits  ->  a claim about the packing
opening    that claim  ->  discharged by WHIR
```

One bit ring switch runs per claim, each leaving one claim about the same packing at a point of its own, and a single WHIR opening at prescribed points discharges all of them. Each surviving value crosses the wire twice: once as the value its reduction closes with, which the verifier replays and takes as its own, and once among the values the commitment opened. The check is that the two agree, and that comparison is the only thing tying the reductions to the commitment.

The evaluation domain now carries its alphabet as a type parameter. Three alphabets are wired: the original 64-bit polynomial basis (unchanged label, unchanged behaviour), `GF(2^32)` for a small-field trace committed at its own width, and `GF(2^128)` for a bit witness packed 128-to-an-element. Each alphabet carries its own Fiat-Shamir label, because two alphabets of the same width can still be different codes — a 64-bit polynomial basis and a 64-bit tower level are not the same code and must not share an instance label.

## Narrowness, measured

The commit path never widens a bit into an element of its own.

| path | committed bytes at 2^16 bits |
|---|---|
| one element per bit | 1 MiB |
| this adapter | 8 KiB |

`the_commitment_holds_one_byte_per_eight_bits` pins that against the retained committed table, not against a formula: the table holds `2^(n-7)` elements of 16 bytes each, which is exactly `2^n / 8` bytes.

For a small-field trace the narrowness is in the alphabet itself. `a_small_field_trace_commits_at_its_own_width` commits a `GF(2^32)` trace with `GF(2^128)` challenges and checks that the opened base-codeword elements are priced at 4 bytes, not 16.

## Profiles, derived rather than tuned

`BinaryWhirProfile` has one constructor per regime and no `Default`, so selecting a regime is always a deliberate act:

- `unique_decoding` — no conjectures.
- `proven_list_decoding` — the Johnson radius, where Reed-Solomon mutual correlated agreement is a theorem (Haböck, eprint 2025/2110; Bordage-Chiesa-Guan-Manzur, CCC 2026).
- `conjectural` — capacity, which every additive binary domain refuses, because the assumption is refuted for characteristic-two subspace domains.

The caller picks the regime, the rate and the folding width. Query counts, out-of-domain sample counts and grinding all come from `p3-security` through the existing schedule derivation, which does no probability arithmetic of its own. Grinding in particular is not a guess: the profile raises the allowance to exactly the figure the analysis reports it needs, iterating to the fixed point rather than hard-coding a number.

Unit tests check all three constructors, that the capacity regime is refused by the domain rather than served, and that decoding further genuinely buys fewer opened positions rather than only changing a label.

## Budget, enforced twice

`BinaryWhirBudget::PRODUCTION` caps opened positions at 200, proof size at 256 KiB and grinding at 24 bits. `BooleanWhirPcs::proof_shape(num_claims, successor_tensors)` reads the whole shape of every proof an opening can produce, straight off the derived schedule and the claim count, before any witness is touched.

It prices every component that crosses the wire, not only the proximity transcript: opened positions, opened elements split by alphabet, the folded codewords' roots and their authentication paths, the sumcheck coefficients and out-of-domain answers, the grinding witnesses at the alphabet's own width, the opened values, and one bit ring switch per claim — its element by rows, two coefficients per round, and the value it survives with. A shape that read only the WHIR schedule would under-count a multi-claim opening badly, and the budget it feeds would be a ceiling on something other than the proof.

The budget is checked in both places: on the schedule up front, and on the serialized proof afterwards. `the_budget_grades_the_schedule_and_the_proof` bounds the estimate against a real proof, so the figure cannot drift into fiction, and checks that a ceiling below the measured proof refuses it rather than warning. `the_estimate_covers_the_reductions_and_not_the_opening_alone` opens twelve claims, enough that the reductions outweigh what the digest count over-estimates, and holds the estimate above the real proof there too.

Neither test pins the derived schedule. A schedule trades grinding against queries, and where it lands on that trade is the derivation's business: #2282 moved this fixture from 105 queries with 3 grinding bits to 108 with none, and the tests were unaffected. What they assert is the shape's own arithmetic — a position opens eight elements off a tree of one depth, the grinding stays inside the budget — and that the claim count adds exactly its reductions and opened values on top of whatever was derived.

The widths the estimate is given are the ones the encoder writes, not the in-memory ones: postcard varint-codes a `BinaryField128` in at most 19 bytes, mean 18.75.

A schedule the challenger could not run is refused where it is built. Grinding searches a candidate space `min(F::bits(), 64)` bits wide and keeps eight bits of headroom, so the real ceiling is 24 bits over a 32-bit alphabet and 56 over the wide ones. A profile whose analysis demands more returns `ProfileError::Grinding` instead of a config that panics in the prover.

## Soundness accounting

The reductions draw their challenges after the commitment but before WHIR's out-of-domain samples. Under proven list decoding the committed word may still sit inside the decoding radius of several codewords at that moment, so a prover can pick which one it answers for after seeing those challenges. That set is fixed by the commitment and cannot be grown, so the cost is exactly a union bound over it: the reduction's term loses `log2_max_candidates` bits.

`PrescribedOpeningSecurity::charge_reduction` performs that subtraction, so the obligation the type documents is carried out by the type rather than by each caller. Both post-commitment terms use it — the bit ring switch here, and the column-batch challenge on the trace's combining route, which is drawn earlier still. At the `2^16`-bit Johnson fixture the candidate count is `2^5.32`, the ring-switch term reads 117.03 rather than 122.36, and the composed bound stays at 99.08, which the WHIR opening term sets. #2282's stronger Johnson bound moves that opening term and so the composed figure; it leaves the candidate count and the charged reduction term exactly where they were.

## Proof size against the folding-only path

At 2^16 committed bits, a 100-bit per-term target, rate `2^-2`, and unique decoding on both sides, on one host:

```
basefold  79417 bytes
whir      33365 bytes
```

Those two figures are not constants, and no test pins them. The proof-of-work search keeps whichever witness a parallel worker reaches first, so a different host absorbs a different witness, every later challenge follows it, and the encoded length moves. The budget fixture, which decodes to the Johnson radius rather than uniquely, measures 27318 bytes here against an estimate of 51455.

What is enforced is therefore the ordering and the ceiling, both of which hold whatever witness wins: `the_whir_proof_is_smaller_than_the_folding_only_one_at_the_same_target` compares the two paths, and the budget test bounds the measured proof by the schedule's own estimate.

## Prime-field regression gate: passed

No file outside `binary-pcs/` is touched, and `p3-whir` does not depend on `p3-binary-pcs`, so the prime-field path cannot move. That is demonstrated rather than asserted: the compiled `whir_pcs` bench binary is byte-identical before and after.

```
before  c7586126aa71ba8715b47588301a6b2bd3a17b45d8928fc70f4acddb9edd785e
after   c7586126aa71ba8715b47588301a6b2bd3a17b45d8928fc70f4acddb9edd785e
```

Proof bytes from that bench, for the record:

```
case      vars    proof_bytes   queries   merkle_digests
small       14          41585       145              516
medium      18          72335       161             1210
large       20          90662       172             1648
```

An identical binary cannot produce different proof bytes or a different timing distribution, so the gate is passed by construction.

## Security review of the adapter

The regime is a security claim, so it must not be reachable by editing a derived schedule. It is checked when the schedule is derived against a domain, and again when the prover and verifier are built. `a_regime_swapped_after_derivation_is_refused_at_construction` takes a schedule derived at the Johnson bound, edits the regime field to capacity, and confirms the prover refuses it.

Every soundness term the adapter introduces appears in the report. `readings_security` returns the proximity argument's own composed budget with the bit ring switch charged on top, and returns nothing when the proximity argument declines to price the schedule, so a security-checked caller fails closed. `the_report_names_every_error_the_adapter_charges` pins both labels and checks that a successor view, which sends two more elements, costs strictly more.

The commitment is bound inside the commit phase, through the same call both sides use. Every public parameter of the run is bound into the Fiat-Shamir instance description before the first challenge that depends on it: the schedule through the proximity argument's own shape, the opening point and the tensor through the reduction's, and the alphabet's identity through its domain label.

## The trace path

A proving system does not hold a bit witness, it holds trace tables. The batched trace commitment is therefore now generic over the bit commitment underneath rather than over two Merkle schemes, so either opening path serves it:

```rust
pub struct BooleanTraceCommitment<EF, B> { .. }

pub type BooleanTracePcs<EF, MT, MX>      = BooleanTraceCommitment<EF, BooleanPcs<EF, MT, MX>>;
pub type BooleanWhirTracePcs<EF, Dft, MT, Challenger>
    = BooleanTraceCommitment<EF, BooleanWhirPcs<EF, Dft, MT, Challenger>>;
```

The surface a trace needs is short, and both backends already had it, so it became three methods on the existing bit-commitment trait plus a transcript-free supertrait naming the data. The supertrait is what lets a proof type be spelled without a challenger, which is why the split exists.

The folding-only spelling is preserved by the alias above, so `multi-stark/examples/prove_boolean_trace.rs` and `examples/src/binary.rs` compile **unchanged** — no file outside `binary-pcs/` is touched. Both are checked in CI and were checked locally.

Two tests drive a real trace through WHIR end to end, one per route the trace commitment takes:

- two tables of unequal height, one read at both row views — the per-column route, which is what a batched prover hits;
- one table read whole — the combining route, which draws a column challenge and charges a batching term for it.

Both check the composed security report term by term, in order, and the first checks that a tampered column value is refused by the reduction that reads the claim off the element.

## Layout

One concern per module, errors in their own file, unit tests at the bottom of the file they exercise, and integration tests driving the public surface from outside.

```
binary-pcs/src/whir/
  mod.rs       wiring and the public surface
  domain.rs    the code, its alphabets and their labels
  profile.rs   one named profile per regime
  shape.rs     what a schedule fixes about its proofs
  budget.rs    the ceiling, and the grading against it
  boolean.rs   the bit-witness adapter and the trace instantiation
  proof.rs     what one opening carries
  error.rs     every refusal the module can return
```

Every error assertion names its variant with the whole payload, and every expected figure is a literal pinned against the fixture rather than recomputed through the code under test. Where a payload comes from a type without equality, the variant is pinned whole and the rendered message is checked alongside it.

`whir/error.rs` sits at 34% comment share, above the usual 30% guide but below this repo's own error modules, which run 39% to 46% because the convention is one doc line per variant field.

## Verification

```
cargo test -p p3-binary-pcs                                       192 passed, 0 failed
cargo check -p p3-multi-stark --example prove_boolean_trace       clean, unedited
cargo check -p p3-examples --lib                                  clean, unedited
cargo clippy -p p3-binary-pcs --all-targets --all-features        clean
cargo +nightly fmt --all -- --check                               clean
```


binary-pcs/src/whir/boolean.rs
```diff
@@ -0,0 +1,538 @@
...
+            .collect::<Result<Vec<_>, _>>()?;
+        let packing = Self::packing(&prover_data);
+
+        let mut readings = Vec::with_capacity(openings.len());
+        let mut sent = Vec::with_capacity(openings.len());
+        let mut surviving_points = Vec::with_capacity(openings.len());
+
+        for (opening, reduction) in openings.iter().zip(&reductions) {
+            let (proof, surviving_point, _) = tracing::info_span!("bit ring switch")
+                .in_scope(|| reduction.prove(&packing, challenger));
```
Nashtare: With this branch merged into current main, `cargo check -p p3-binary-pcs` stops here with `error[E0283]: type annotations needed ... type parameter R declared on the method prove`. #2253 made `BitRingSwitch::prove` generic over the representation the rounds run in.

That is one of three changes main has made to the loop this file copies from `boolean.rs`:
- `prove::<<EF as ChallengeField<EF>>::SumcheckRepr, _, _>`;
- a borrowed `BitPackingView` in place of the cloned packing;
- reusing the reduction's surviving value instead of evaluating the packing again for each claim, which WHIR's `open_at` still does here.

The rest of the copy is `check_openings`, `reduction`, `current_openings`, `packing`, `open_readings`, `verify_readings`, the point API and a mirrored error enum. Could the rebase pick up the same representation? Could the reduce-then-open logic also live in one place, generic over the inner prescribed-point PCS, so the two paths stop drifting?

binary-pcs/src/whir/shape.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub sent_extension_elements: usize,
+
+    /// Hardest grinding any phase of the schedule demands.
+    pub grinding_bits: usize,
+}
+
+impl ProofShape {
+    /// Read the shape of every proof a schedule can produce.
+    #[must_use]
+    pub fn of<EF, F, Challenger>(config: &WhirConfig<EF, F, Challenger>) -> Self
```
Nashtare: This reads only the WHIR schedule, but a `BooleanWhirProof` also carries one `BitRingSwitchProof` per claim, each with a 128-row tensor. Locally, at the budget test's fixture (2^16 bits, Johnson, estimate 41,792 bytes), a reduction was 2,762 bytes with the current reading and 7,566 with the successor:
- 8 current claims gave a 43,468-byte proof;
- 4 successor claims gave 51,603 bytes;
- the trace fixture in `a_boolean_trace_commits_and_opens_through_whir` has an estimate of 2,032 bytes and a 20,700-byte proof.

The budget test passes because it opens two claims, and `merkle_digests` counts full paths with no cap and no shared nodes. The WHIR opening alone was about 21.3 KB. The widths are also below what postcard writes: a `BinaryField128` came out as 17–19 bytes (mean 18.75), not 16. The claim values, round caps and grinding witnesses aren't counted either.

As it stands, `check_shape` would accept a multi-table trace, which takes one reduction per column, whose proof is far past 256 KiB. Could the shape take the claim count and the successor flag and price the reductions? If not, could the docs and the description say it bounds the WHIR opening only?

binary-pcs/src/whir/boolean.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub fn readings_security(
+        &self,
+        num_claims: usize,
+        successor_tensors: bool,
+    ) -> Option<PrescribedOpeningSecurity> {
+        let protocol = self.protocol(num_claims);
+        let mut security = self.inner.prescribed_security(&protocol)?;
+        // The tensor alone, or the tensor with carry and last.
+        let num_tensors = if successor_tensors { 3 } else { 1 };
+        security.terms.push(bit_ring_switch_tensors_term(
```
Nashtare: The ring switch draws its challenges after the commitment. WHIR's out-of-domain samples only come later, in `open_at`. Under `proven_list_decoding`, the committed word can still be close to `2^log2_max_candidates` codewords while the reductions run, so the prover can pick the candidate after seeing their challenges.

`PrescribedOpeningSecurity` asks callers to union-bound their own reductions over that set. This adapter is such a caller, but the term goes in unscaled. At the test fixture the list is 2^5.32, so the ring-switch term should read about 117.0 bits instead of 122.4. The composed 96.92 bits doesn't move at `GF(2^128)`, so this is about the accounting, not a live gap.

The trace's column-batch term on the combining route is drawn even earlier and has the same issue. Could both be charged with `log2_max_candidates` subtracted?

binary-pcs/tests/whir_boolean.rs
```diff
@@ -0,0 +1,510 @@
...
+        tampered[0] += EF::ONE;
+        let mut verifier_chal = challenger();
+        pcs.observe_commitment(&commitment, &mut verifier_chal);
+        let refused = pcs
+            .verify_at_points(&commitment, &points, &tampered, &proof, &mut verifier_chal)
+            .unwrap_err();
+        assert!(
+            matches!(
+                refused,
+                BooleanWhirError::ReductionProof(BitRingSwitchProofError::ClaimMismatch)
```
Nashtare, line 220: This rejection comes from the ring switch's own column check, before WHIR runs. No test in the PR reaches `BooleanWhirError::SurvivingClaim` or `BooleanWhirError::Opening`. So deleting the closing comparison in `verify_readings` (lines 398–406), the only thing tying the reductions to the commitment, would still pass every test.

`boolean.rs` has `a_non_first_reduction_over_another_witness_leaves_a_claim_the_commitment_does_not_open` for the folding-only path. Could it be mirrored here, along with a tampered WHIR opening and a commitment to another witness?

binary-pcs/src/whir/boolean.rs
```diff
@@ -0,0 +1,538 @@
...
+                .then(|| reduction.successor_claim(&proof.tensor, proof.successor.as_ref()))
+                .transpose()
+                .map_err(BooleanWhirError::Reduction)?;
+
+            readings.push(BitReadings { current, next });
+            surviving_points.push(surviving_point);
+            sent.push(proof);
+        }
+
+        // The surviving values never cross the wire: a verifier recomputes its own.
```
Nashtare: The surviving value does cross the wire, twice. It is sent as `BitRingSwitchProof::final_eval`, which the verifier replays and takes as its value, and again in `PcsProof::evals`, which it's compared against. That's sound, but this comment and the description say the opposite. Could the comment say what is actually checked?

binary-pcs/src/whir/profile.rs
```diff
@@ -0,0 +1,224 @@
...
+use p3_whir::{
+    FoldingFactor, ProtocolParameters, SecurityAssumption, WhirConfig, WhirConfigError, WhirDomain,
+};
+
+use crate::whir::error::ProfileError;
+
+/// Grinding search ceiling.
+///
+/// A witness is one alphabet element, so no schedule can demand more bits than a wide one holds.
+const MAX_GRINDING_BITS: usize = 64;
```
Nashtare: `BinaryChallenger::grind` asserts `bits + 8 <= min(F::bits(), 64)`. So the real ceiling is 24 bits for the `GF(2^32)` alphabet and 56 for the wide ones, and a schedule asking for more comes back from `config` and then panics in the prover. Could the cap follow the witness field, or could `config` refuse such a schedule?

binary-pcs/tests/whir_boolean.rs
```diff
@@ -0,0 +1,510 @@
...
+            actual: 105,
+            budget: 104
+        })
+    );
+    assert_eq!(
+        BinaryWhirBudget {
+            max_proof_bytes: 26_818,
+            ..budget
+        }
+        .check_bytes(26_819),
```
Nashtare: These two literals are one host's proof size and aren't tied to the `bytes` measured above. `check_bytes(bytes)` against `max_proof_bytes: bytes - 1` would test the measured proof.

Nashtare (changes requested):
I read the diff against `2ab6faaf8` and traced the adapter through the prescribed-point WHIR opening and the bit ring switch. Most of the description holds:

- **Poly64 domain unchanged.** It keeps its label, capacity (64), stratification and encoder, so its WHIR instance description doesn't change.
- **Capacity refused.** `new_with_domain` refuses capacity. A regime edited afterwards trips the asserts in both `WhirProver::new` and `WhirVerifier::new`.
- **Fiat–Shamir order.** Each ring switch binds its point and tensor before it draws. Every claim is recorded before WHIR's batching challenge. The verifier closes each `final_eval` against the opened value.
- **Field size.** The WHIR terms are priced at `EF::bits() - 1`, so the `GF(2^32)` case is charged against `GF(2^128)`.
- **Johnson citation.** It holds in characteristic two. BCHKS state list correlated agreement (Thm 4.6) for any field and domain, and defer the inseparable case to BCIKS20 App. C.
- **Folding-only trace path.** It reaches the same inherent methods in the same order.

- The branch no longer compiles against main. The merge is textually clean, but the reduction loop this file copies from `boolean.rs` has since changed there (inline, `whir/boolean.rs`).
- `ProofShape` is documented as bounding every proof a schedule can produce. It leaves out the ring-switch proofs, and the trace fixture's proof is ten times its estimate (inline, `whir/shape.rs`).

**Smaller things**
- This breaks public API, so I think the title should carry `!`. `BinaryWhirDomain`'s type parameters changed, `BooleanMultilinearPcs` has a new supertrait and required methods, and `recommended_cap_height` has new generics.
- With the associated types moved into `BooleanBackend`, `<P as BooleanMultilinearPcs<..>>::Commitment` fails with E0576. #2275 uses that spelling in `record.rs` and `verifier.rs`, so whichever of the two lands second has to spell them through `BooleanBackend<EF>`.
- Nothing outside the tests builds `BooleanWhirPcs` or `BooleanWhirTracePcs`, or checks a `BinaryWhirBudget`. So "enforced twice" only holds in the fixture.

binary-pcs/src/whir/boolean.rs
tcoratger: Reproduced: `cargo check -p p3-binary-pcs` after merging onto current main stopped at E0283 on that exact line. Rebased onto `upstream/main` (now on top of #2281) and picked up two of the three changes.

The reduction runs under `<EF as ChallengeField<EF>>::SumcheckRepr`, and the packing is now a `BitPackingView` borrowed straight from the retained table rather than a `Vec` copied out through a parallel iterator. `cargo check -p p3-binary-pcs --all-targets` is clean.

The third change is not reachable from here. Reusing the surviving value on the folding path goes through `try_open_at_known`, which is a `BinaryPcs` method; `PrescribedPointPcs` exposes only `open_at`, which takes no values the caller already holds. So WHIR re-derives them, and a shared reduce-then-open helper generic over the inner PCS would have to either add a known-value opening to the trait or drop that optimization on the folding path. I left the trait alone rather than widen it inside a review round, but I agree the two copies of the same three-step loop should end up in one place, and the missing trait method is what blocks it.

binary-pcs/src/whir/shape.rs
tcoratger: You are right on every count, and I took the first option: the shape now takes the claim count and the successor flag and prices the whole proof.

I measured the fixture component by component before changing anything, and your model of a reduction is exact. At `2^16` bits, Johnson, one reduction is `128 + 2*9 + 1 = 147` challenge-field elements (tensor rows, two coefficients per sumcheck round, the surviving value), which at 18.75 bytes mean is the 2,758 and 2,768 bytes I measured for the two claims. With carry and last it is `3*128 + 19 = 403`.

What the shape missed, and now counts:

- one reduction per claim, `num_tensors * 128 + 2 * num_variables + 1` elements each;
- the opened values in `PcsProof::evals`, one per claim;
- the folded codewords' roots, one digest per round;
- the grinding witnesses, which are now a `sent_base_elements` field priced at the alphabet's width rather than folded into a figure that assumed every sumcheck round grinds.

`ProofShape::of` now takes the opened-value count so it cannot silently omit it, and `ProofShape::of_bit_readings` adds the reductions; `BooleanWhirPcs::proof_shape(num_claims, successor)` is the one call the adapter offers. The fixture's widths were also wrong: postcard writes a `BinaryField128` as a varint of at most 19 bytes, mean 18.7485 over 2,000 samples, so the tests now pass 19 and keep 16 only for the in-memory assertion about committed bytes.

New fixture numbers at two claims: `stir_queries 105, opened_base 840, opened_ext 0, merkle_digests 840, sent_base 10, sent_ext 379`, estimate `(840+10)*19 + 379*19 + 840*32 = 50,231` against a measured 26,840. The slack is the Merkle count, which still ignores the cap and shared nodes; the "not far too loose" assertion moved from a factor of two to three to stay off that edge.

There is a regression test for exactly your eight-claim case: it opens eight claims, asserts the proof exceeds the 41,792 bytes the old opening-only estimate reported, and checks the new estimate still bounds it. That test fails on the previous code.

One caveat now stated in the doc: the figure counts the values, not the framing the serializer adds around each list.

binary-pcs/src/whir/boolean.rs
tcoratger: Worked it out. The ordering is sound; the accounting was not. Both terms now subtract the candidate count.

**Why the ordering is sound.** Write `L` for the set of polynomials whose codewords lie inside the decoding radius of the committed word `w`.

1. `L` is fixed by the commitment. `w` is Merkle-committed and absorbed before anything is drawn, and the code and the radius are public properties of the schedule. So `L` is determined before the ring switch draws its first challenge, and nothing a prover does later can enlarge it. This is the load-bearing fact: the prover is choosing from a set fixed in advance, not from one it can grow adaptively.

2. The ring switch's bound is a statement about one fixed polynomial: for a fixed `f` and a false reading, the probability over its challenges that it accepts and leaves a true surviving claim about `f` is at most `eps`. The extra power the ordering gives a prover is to decide, after seeing those challenges, which member of `L` it will answer for. So the event to bound is "there exists `g` in `L` such that the reduction accepts and its surviving claim about `g` is true", and a union bound over `L` gives `|L| * eps`. That is exactly `log2|L|` bits off the term. Nothing worse than that is available, because the choice is over a set of known, fixed size.

3. Drawing the out-of-domain samples before the reductions would collapse `L` to one element first and let the term go in unscaled, at the cost of splitting the out-of-domain phase out of `open_at` and threading it through the adapter. It buys back `log2|L|` bits on the reduction term alone. At this fixture that is 117.0 instead of 122.4, against a WHIR opening term of 96.92 — the composed bound does not move by a single bit. Restructuring the transcript to buy bits nothing is short of is not worth the risk, so I fixed the accounting instead.

**The numbers, reproduced exactly.** At `2^16` bits, Johnson, two claims: candidates `2^5.3219`, ring switch `117.0342` (was `122.3561`), composed `96.9213` before and after. With carry and last, `116.9232`.

**What changed.** `PrescribedOpeningSecurity` gained `charge_reduction`, which pushes a term with `log2_max_candidates` subtracted and floors at zero. The obligation the type's own doc states is now carried out by the type rather than by each caller remembering to do it, which is the shape of #2271. Both call sites use it: the bit ring switch term here, and the column-batch term on the trace's combining route, which as you say is drawn even earlier and had the same gap. `multi_stark` already did the same subtraction to its own post-commitment reductions, so this is the established convention, and there is no double charge — each layer union-bounds its own reductions once, and the candidate count is forwarded unchanged.

Tests: a unit test on `charge_reduction` (subtraction, the zero floor, and unique decoding leaving a term alone), plus assertions in both integration fixtures that the charged term equals the unscaled term minus `log2_max_candidates`, with the unscaled value recomputed from the term function rather than read back off the adapter.

binary-pcs/tests/whir_boolean.rs
tcoratger: Verified the claim first. With the closing comparison at lines 398–406 replaced by a no-op, `cargo test -p p3-binary-pcs` was fully green: 150 + 3 + 4 + 30 + 5 + 9 + 1 passed, nothing failed. So the only thing tying the reductions to the commitment was untested, exactly as you said.

Two tests now reach the two paths.

`a_non_first_reduction_over_another_witness_leaves_a_claim_the_commitment_does_not_open` mirrors the folding-path one: commit A on a throwaway sponge, commit B on the real one, run the middle of three reductions over A's packing and the other two over B's on B's sponge, then open B at all three surviving points. It asserts that claims 0 and 2 agree and the middle one does not, so a verifier that checked only the first pair would accept, and that `verify_readings` returns `SurvivingClaim`. It then replays B's honest proof at the same points to show the rejection is the mismatch and not the construction. Under the mutation above this test fails and the other ten still pass, so it is the one holding that line.

`a_tampered_proximity_transcript_is_refused_by_the_opening` leaves the reductions honest, moves `opening.whir.initial_ood_answers[0]`, and asserts `Opening`, with the untouched proof accepted alongside.

Both live in `tests/whir_boolean.rs` using public API only — the schedule and the current-row openings are rebuilt in the test rather than exposing the adapter's private helpers.

binary-pcs/src/whir/boolean.rs
tcoratger: Right, and the comment now says what is checked:

    // Each surviving value crosses the wire twice, and the closing check is that the two agree.
    // Every surviving point came out of a reduction's rounds, so all are bound already.

The identical claim sits at the same place on the folding-only path in `binary-pcs/src/boolean.rs`, so I corrected it there too rather than leave the same wrong sentence one file over. The PR description is updated to match.

binary-pcs/src/whir/profile.rs
tcoratger: Confirmed both ceilings directly against the challenger before changing anything: `grind(25)` on a `BinaryField32` witness and `grind(57)` on a `BinaryField128` one both hit the margin assertion, so the real limits are 24 and 56 as you say.

`config` now refuses such a schedule rather than returning it. The cap follows the witness field — `min(F::bits(), 64) - 8`, the two constants named after what the challenger enumerates and the headroom it keeps — and a requirement above it returns a new `ProfileError::Grinding { required, ceiling }` instead of being silently accepted. The old `MAX_GRINDING_BITS = 64` is gone.

Three tests: a 32-bit alphabet at 20 variables asking 27 bits is refused with `{ required: 27, ceiling: 24 }` and its rendered message pinned; the wide alphabet at 16 variables asking 57 is refused with `{ required: 57, ceiling: 56 }`; and two `should_panic` tests call `grind` one bit above each ceiling, so if the challenger's margin ever moves the constant here is caught rather than quietly wrong.

binary-pcs/tests/whir_boolean.rs
tcoratger: Agreed, your version is better. The assertion is now

    BinaryWhirBudget { max_proof_bytes: bytes - 1, ..budget }.check_bytes(bytes)
        == Err(BudgetError::Bytes { actual: bytes, budget: bytes - 1 })

so it grades the proof that was measured a few lines above and carries no host's number. The query-axis assertion next to it stays literal because 105 is read off the schedule, not off a proof.

binary-pcs/tests/whir_boolean.rs
```diff
@@ -0,0 +1,765 @@
...
+        .map(|term| term.bits.bits())
+        .fold(f64::INFINITY, f64::min);
+    let with_successor = pcs
+        .readings_security(2, true)
+        .unwrap()
+        .terms
+        .iter()
+        .map(|term| term.bits.bits())
+        .fold(f64::INFINITY, f64::min);
+    assert!(with_successor <= plain);
```
Nashtare: The minimum over the terms is the WHIR term in both reports (96.921 bits locally), so this holds with equality. It would still pass if `readings_security` ignored `successor_tensors` altogether. The earlier version compared the composed error and would have caught that. The comment above still says "strictly more". Could this compare the `bit-ring-switch` term by label instead? At this fixture that is 116.923 < 117.034.

Nashtare (approved):
I re-checked this at `6d9cebcc9`. `git range-diff` shows the two original commits carried over unchanged apart from context, so I read the two new commits in full.

- **Builds.** Merged with current main, `cargo check` is clean for `p3-binary-pcs` and `p3-sumcheck` (all targets), the `prove_boolean_trace` example and `p3-examples --lib`. With #2275 on top as well, `p3-word-backend` is clean too, since it now spells the types as `Pcs::Commitment`.
- **Proof shape.** `proof_shape(k, successor)` now bounds the proof. Locally I ran 22 openings: 1–32 current claims and 1–16 successor claims at the budget fixture, plus unique decoding and 2^12, 2^18 and 2^20 bits. Every postcard size came in under `max_bytes(19, 19, 32)`, with at least 2.2 KB to spare. The eight-claim case is 43,400 bytes against 67,103.
- **Candidate charge.** At the fixture the ring-switch term reads 117.034 bits, or 116.923 with the successor view, and the composed bound stays at 96.921. The combining route's column-batch term is charged the same way. The folding-only path forwards zero candidates, so nothing moves there.
- **Grinding ceiling.** It matches the assert in `BinaryChallenger::grind`: 24 bits for the 32-bit alphabet, 56 for the wide ones.

The closing-check tests, the corrected comment and the measured-bytes assertion close the other earlier threads. I agree that the known-value opening missing from `PrescribedPointPcs` is what keeps the two reduce-then-open loops apart, and that's fine as a follow-up.

The inline comment and the reply are non-blocking.

**Smaller things**
- The title still has no `!`, and `BinaryWhirDomain`, `BooleanMultilinearPcs` and `recommended_cap_height` still break.
- The description says no file outside `binary-pcs/` is touched, but `sumcheck/src/prescribed_pcs.rs` now is. "Budget, enforced twice" still only holds in the tests, since no library path calls `check_shape` or `check_bytes`.

binary-pcs/src/whir/shape.rs
Nashtare: Locally, the adapter side holds up. I ran 22 openings: 1 to 32 current claims, 1 to 16 successor claims, and four other schedules. Every proof came in under `proof_shape(k, successor).max_bytes(19, 19, 32)`.

The trace path from the same comment still has no way to get a shape. On the per-column route, `a_boolean_trace_commits_and_opens_through_whir` runs five reductions for its two opening batches:
- `proof_shape(2, true)` gives 17,260 bytes, against a real 20,700;
- `proof_shape(5, true)` plus the eight `values` gives 39,870.

A caller has no public way to know it's five. Could `BooleanTraceCommitment` offer a `proof_shape(protocol)` that forwards its route's claim count and adds `values`? A follow-up is fine.

binary-pcs/tests/whir_boolean.rs
tcoratger: You are right, and it is worse than a weak assertion: it was the one guarding the charge I added last round, so that charge was unguarded. I weakened it myself when I moved from the composed error to a minimum, and I did not check that the minimum could still see the difference. It cannot.

Verified both directions before fixing. With `successor_tensors` ignored in `readings_security`, the minimum stays at 96.921313 on both sides — the assertion passes on code that has dropped the feature entirely. The composed error does move, but only by 1.0e-7 bits, which is true but reads as noise.

Took your suggestion: the test picks the `bit-ring-switch` term by label and compares that.

```rust
assert!(reduction_bits(true) < reduction_bits(false));
```

116.923 against 117.034, a gap of 0.111 bits rather than zero. With the successor contribution removed this test now fails on that line, which I confirmed. The same closure feeds the candidate-charge assertion below it, so there is one way of reaching that term rather than two.

The comment was wrong twice over and now states what holds: carry and last are two more elements batched under the same draw, so the reduction is dearer, and the weakest term is the proximity argument's either way, so a minimum sees no change.

Issue #2242, opened 2026-09-19
Add differential verification and binary-backend release gates


tcoratger:
## Deliverable

Make releases depend on independent evidence that the binary backend is correct, interoperable, secure, and within budget.

## Acceptance criteria

- Consume published Flock vectors for compatible challenger, field, transform, Merkle, sumcheck, and zerocheck layers.
- Publish Plonky3 vectors for its tower basis and additive transform.
- Maintain scalar-reference property tests for every specialized path.
- Add an independent verifier or executable specification for the shipped proof format.
- Run at least one CI leg at the full native vector width.
- Fail closed when any soundness contribution is missing from the security report.
- Gate the supported profiles on proof size, verifier time, and critical throughput regressions.

Issue #2231, opened 2026-09-19
Tracking: complete the non-recursive binary proving backend


tcoratger:
## Goal

Make Plonky3 a complete binary-field proving backend for a separate general-purpose RISC-V zkVM.

Recursion is deliberately out of scope. It will be handled by a future Plonky3-recursion upstreaming effort.

Two parameters are still set by it, though, and choosing them on native cost alone would be a mistake.

The Merkle and Fiat-Shamir hash should stay swappable behind one trait until in-circuit cost can be measured.

The proof-size budget should be set with the eventual recursion step in mind rather than for native verification alone.

## Completion criteria

The backend is complete when one public API can prove and verify a multi-table statement where:

- traces remain at their natural bit or small-field width;
- tables may have different heights without maximum-height padding;
- local constraints and cross-table bus claims are authenticated together;
- mutable read-write memory is enforced by a reusable backend argument;
- the word backend proves its nonlinear relations and discharges one committed-trace opening;
- binary WHIR provides the compact opening path without regressing prime-field WHIR;
- proof shape, parsing, and every soundness term are verifier-derived and bounded;
- downstream witness ingestion avoids an unconditional full-trace copy; and
- reproducible benchmarks report throughput, proof size, verification time, and peak memory; and
- those benchmarks are held against a stated target rather than only against the previous run.

## Targets

Capability alone does not finish this.

Every criterion above can be met while the backend remains an order of magnitude off the systems it is measured against.

The reference point is the published comparison on one machine at one security level with witness generation included.

Plonky3 is measured there on its prime-field path at 0.8k Keccak permutations per second on one thread and 3.35 MiB of proof, against 30.7k and 386 KiB for the fastest binary system.

The binary path has never been measured end to end at scale, so the first deliverable is the number, not the gap.

Once it exists, a throughput and proof-size target belongs here, and a sub-issue that cannot move its share of it should be reconsidered rather than completed.

Proof size carries a second constraint the scope boundary hides: it is what decides whether recursion is affordable later, so it cannot be traded away freely while recursion is out of scope.

## Scope boundary

Plonky3 owns fields, commitments, reductions, table composition, memory arguments, proof metadata, and the machine-facing backend contract.

The downstream repository owns the ISA, ELF loader, emulator, instruction chips, precompile chips, guest toolchain, and aggregation policy.

[...]

PR #2274, opened 2026-09-20
feat: add differential verification and binary-backend release gates

tcoratger:
Independent evidence that the binary backend is right, plus the release gates that refuse a
build when that evidence is missing. Every other sub-issue of #2231 adds a capability; this one
only adds ways for a wrong capability to be caught, so each item below states the failure it
would catch and each gate ships with a check that it does catch it.

Part of #2242

---

## 1. Consume published vectors

The differential vectors live under `binary-field/tests/differential/`, one module per layer, so
a mismatch in one layer is not buried under the others.

Two external sources turned out to be usable, and the conversion each one needs is written into
the fixture header rather than left implicit.

**Flock GF(2^128)** (`differential/ghash_field.rs`) — `differential/vectors/flock-ghash-products.txt`, 128 products dumped
from `succinctlabs/flock` at `b684b12` with its own `dump_ghash_vectors` binary. Flock's `F128`
is the same field as our `Ghash128`: reduction `x^128 + x^7 + x^2 + x + 1`, coefficient of `x^i`
at bit `i`. Flock stores it as two little-endian 64-bit words; joining them needs **no bit
reversal**, which is exactly the opposite of a NIST GCM block and is why the header says so
out loud. All 128 products agree.

**FIPS 197** (`differential/rijndael_field.rs`) — the published AES substitution table, checked against `Rijndael8b` inversion plus
the standard affine map. Same convention note: FIPS 197 already puts the coefficient of `x^i` at
bit `i`, so unlike a GCM block no reversal applies. A companion test shows the reversed reading
disagrees, so the check is not convention-blind.

**What it would catch.** Before this, `Ghash128` was pinned externally by exactly one NIST GHASH
case; everything else compared it against our own tower, so an error shared by both
representations was invisible. `Rijndael8b` had no external pin at all, and it is the field the
GFNI kernels run on.

**Layers where no compatible vector exists, and why.** Flock's challenger is a SHA-256
`FsChallenger` with its own domain separation and grinding convention, and its Merkle, sumcheck
and zerocheck dumps are all keyed to that challenger and to Ligerito's stratified query schedule.
There is no shared layer to convert: consuming them would mean implementing Flock's protocol, not
checking ours. binius64 has moved its 128-bit field to the same GHASH basis we already check, and
its 8-bit field is the same Rijndael field — those two are covered above; its tower no longer
matches ours. That leaves the challenger, Merkle, sumcheck and zerocheck layers of criterion 1
unmet, and I would rather say so than publish a conversion nobody can defend.

## 2. Publish our own vectors

- `binary-field/tests/differential/vectors/tower-basis.txt` — the Wiedemann tower: eight products and four
  inverses at each of the 8/16/32/64/128-bit levels, plus all 128 Cantor basis vectors. The header
  states the recurrence, the bit layout and the record grammar, so an outside implementation can
  meet the file without reading any code here.
- `binary-dft/tests/vectors/additive-transform.txt` — four transform cases (two levels, two
  lengths, with and without a coset shift), coefficients in the novel polynomial basis and the
  resulting evaluations.

Both files were produced by a reference written from the definitions, not dumped from the crate.
The transform reference builds the normalised subspace polynomials from the **product** definition
and divides by the value at `v_j`; the crate instead uses the iterated `x^2 + x` shortcut. The two
agree, which is the first independent confirmation that the normalisation convention is the one
the documentation claims. `NaiveAdditiveNtt` is held to the same file, so the two in-crate
transforms cannot drift onto a shared wrong answer.

**What it would catch.** A change to the Cantor basis, to the tower recurrence, or to the
normalisation point of the subspace polynomials. The shift is pinned separately: a test asserts
the shifted case does *not* also hold at the origin.

## 3. Scalar-reference inventory

The full sweep is below. Two gaps were closed here; the rest are reported rather than fixed,
because closing them means editing crates other agents are in or changing what the wasm leg
covers.

Closed:
- `keccak/tests/vectorized_permutation.rs` — the vectorized permutation against `tiny_keccak`
  lane by lane, over random states and a one-hot lane sweep. Each backend previously checked
  itself against a single fixed state.
- `binary-field/tests/packed_gf2.rs` — the generic characteristic-two ring suite over all seven
  bit-sliced packings, which went through no shared suite before.

Reported, not closed:
| Path | State |
|---|---|
| `binary-field/src/packed/x86_64.rs` (`PackedGhash128`), `poly_slice.rs::wide` | Excellent scalar-reference tests that **never execute**: both `vpclmulqdq` legs are build-only with no conditional run step, and QEMU cannot emulate the 256-bit product. The new native leg runs them on any runner that has the instruction. |
| `binary-dft/src/lanes.rs` GFNI/AVX-512 and NEON register impls | Covered only end to end through `butterfly.rs`. `test_util::Model` exists for exactly this and is wired only into `affine.rs`. The sibling crate does it right (`lanes_conform`). Needs an in-crate test. |
| `goldilocks` SVE2 arms | Four tests are `#[cfg(target_feature = "sve2")]` under a build-only leg, so they have never run. |
| `keccak/src/sse2.rs` SSSE3 `rol_8`/`rol_56` | Needs `+ssse3` without `+avx2`; no leg produces that, so the arms are compiled by nothing. |
| `field/src/packed/interleaves.rs` AVX-512VBMI2 variant | `+avx512f` does not imply `+avx512vbmi2`; no leg compiles it. |
| `p3-keccak` / `p3-sha256` wasm32 simd128 backends | Good tests; the wasm leg covers `p3-goldilocks` only. |
| `multi-stark/src/lookup/test_field.rs` `TinyParameters` | Hand-written packed constants with no packed test at all. |
| `uni-stark` aarch64 `VectorizedEval` | No test asserts it agrees with `PlainEval`; it also changes a public trait bound on aarch64. |
| `monty-31` | Zero tests of its own; all three SIMD packings are pinned only by re-instantiation in `baby-bear` and `koala-bear`. |
| `goldilocks::add_no_canonicalize_trashing_input` | Two arms gated on `target_arch` alone, nothing compares them. |

One correction to the sweep: the bit-sliced packings are **not** a missing `PackedField` suite.
They implement `Algebra<Gf2>` and carry no scalar lanes, so the packed-field suite does not apply
to them at all; the ring suite does, and that is what was added.

## 4. Executable specification

`sumcheck/tests/executable_specification.rs` reimplements the generic-degree sumcheck — the
reduction the shipped zerocheck delegates to — from its format description. It rebuilds the
omitted second evaluation, the interpolation node set and the interpolation itself by hand, with
plain Lagrange rather than the shipped barycentric interpolator, and never calls
`GenericDegreeProof::verify` to decide anything.

Three tests: the specification prover and verifier close the reduction against a hand-evaluated
multilinear product; the shipped verifier returns the **same** challenges and final claim on a
specification-produced proof; and the specification rejects five distinct edits to a finished
proof (a round value, the claimed sum, the round order, a round width, a grinding witness).

**Boundary, stated plainly.** The sponge and the step schedule are shared, because those are the
format rather than the reduction. This covers the sumcheck layer of the shipped proof format and
not the Merkle, PCS or multi-STARK layers. A full independent verifier for `MultiStarkProof`
remains open.

## 5. A CI leg at full native vector width

Issue #2167 asked for this and was closed by #2193, which changed nothing under `.github/`. I
re-checked: there is no `target-cpu` anywhere in the workflows, the widest leg that *runs* tests
is `+avx2`, and every AVX-512, VPCLMULQDQ and SVE2 leg is `compile_only: true`.

New leg `linux, native vector width` builds and runs the suite under `-C target-cpu=native`, in
both serial and parallel modes. It prints the resolved feature set, and **fails** if that set
lands below AVX2 — without that, a narrowing of the runner pool would silently turn the leg into
a duplicate of the baseline and nobody would notice.

Verified locally on a Zen 4 host (avx512f/bw/vl/dq, gfni, vaes, vpclmulqdq, sha): the suites for
`p3-binary-field`, `p3-binary-dft`, `p3-sumcheck` and `p3-multilinear-util` pass at native width,
and the new Keccak test exercises the 8-lane AVX-512 backend that no current leg runs.

**What it would catch.** Exactly #2167: an assertion or kernel gated on packing width that no
executing leg reaches.

## 6. Fail closed on a missing soundness contribution

This is the one with teeth. `security/src/budget/` grades seven rounds, and two of them —
the lookup round and the DEEP-composition round — are *waived* when the caller passes `None`.
A waived round is reported at the transcript ceiling, which is also what a genuinely strong round
reports, so **nothing in the crate could tell the two apart**.

The check is its own module, `security/src/budget/coverage/`, with the error type in its own
`error.rs` alongside the twelve other per-module error files in the tree.

- Each term now records whether its bits bound anything (`Accounting`).
- `ChallengeSchedule` is the list of rounds the protocol's transcript actually samples a challenge
  for. It comes from the transcript, not from the shape fields the budget already read, which is
  what makes the check capable of disagreeing.
- `AuditedReport` is the only thing that reports a security level. `binding_term`, `attained` and
  `security_level` are now `pub(crate)` on the raw report, so a level cannot be read off a budget
  that skipped one of the protocol's rounds.
- The audit is `const`, so a configuration can be refused at compile time.

**What it would catch, demonstrated in the test.** A caller grading a protocol that has a bus
while passing `lookup: None` gets 96 bits today — the same number as the fully declared
configuration. The test asserts that 96 first, then asserts the gate refuses it with
`Unaccounted::Waived("lookup-challenge")`. A round the budget does not model at all is refused
as `Unmodeled`.

## 7. Profile gates on proof size, verifier time and throughput

The mechanism ships; the numbers do not. The reproducible scoreboard work that produces them is
not landing yet, and inventing figures would gate releases on fiction.

- `.github/profile-budgets.toml` lists the three supported binary-path profiles with all nine
  budgets explicitly `unset`. `unset` is not the same as absent — the checker refuses a profile
  that omits a key, so a budget can only ever be missing on purpose.
- `.github/scripts/profile_gate.py` compares measurements against the table. Ceilings for proof
  size and verifier time, a floor for throughput. It fails closed: a measured profile with no
  entry, an entry never measured, and a missing measurement key are all failures.
- Wired into the lint job, which runs the gate's own nine-test suite and validates the table.

**What it catches today, with no numbers set.** A supported profile shipped with no budget entry,
a budget entry naming an entry point that has been renamed or deleted, and a malformed budget.
I verified the last one by renaming a profile in the table and watching the self-test fail.

**What remains.** The three thresholds per profile. They land with the scoreboard work.

---

## Why `Part of` and not `Closes`

Three criteria are not fully met:

1. Criterion 1 covers the field layer only. The challenger, Merkle, sumcheck and zerocheck layers
   have no published vector that is compatible with ours, for the reason given above.
2. Criterion 4 covers the sumcheck layer of the shipped format, with the sponge shared.
3. Criterion 7 ships the gate and the table, with every threshold unset pending the benchmark
   harness.

Criteria 2, 3, 5 and 6 are complete.


tcoratger:
Closing for now — not a priority against the rest of the binary-backend track, and the surface it adds costs more in complexity than it currently returns.

The work is preserved on `feat/binary-differential-gates` and the issue it is part of stays open.

Two findings from it are worth keeping regardless of this branch, since both are live on `main` today:

- Every CI leg carrying `vpclmulqdq`, `gfni`, `avx512f` or `sha` is build only, so the packed GHASH and wide polynomial-slice kernels have tests that no leg has ever executed. The widest leg that runs tests is `avx2`. Issue #2167 was closed by #2193, which touched nothing under `.github/`.
- A waived round in the security report returns the transcript ceiling, numerically identical to a properly accounted round, so nothing distinguishes the two. A bussed protocol graded with its lookup term omitted reports 96 bits.

Issue #2237, opened 2026-09-19
Complete the word-level proof protocol


tcoratger:
## Deliverable

Turn the word relation language into a complete proof that ends at one authenticated Boolean-trace opening.

## Acceptance criteria

- Shifted operand claims reduce to committed words with verifier-fixed wiring.
- Zero, bitwise AND, and full-width unsigned multiplication relations are proved.
- Integer multiplication, binary multiplication where needed, AND, zero, shift, and public checks execute in the sound transcript order.
- Every intermediate claim is transcript-bound before its challenge.
- The final claim is discharged through the Boolean PCS.
- Security accounting includes batching, zerocheck, shift, ring-switch, and PCS terms.
- Word32 and Word64 have adversarial and scalar-reference tests.

PR #2226 implements the shift-reduction layer only.

tcoratger:
Scope note on the full-width unsigned multiplication relation, recorded so it is not treated as settled.

That relation is the most expensive part of the word layer and the least load-bearing for a general-purpose machine's common path.

Binius64 proves it, leanVM currently does not, and a simpler and substantially faster protocol for it is expected.

The work in flight still implements it against the criteria above, but it should be built as its own module so that replacing it later is a module swap rather than surgery across a shared file.

This does not change the other relations or the transcript ordering requirement.

PR #2275, opened 2026-09-20
feat(word): prove the zero and bitwise relations through the Boolean PCS

tcoratger:
## What

The word relation language now produces a complete proof that ends at **one authenticated Boolean-trace opening**, for the zero and bitwise AND families.

Full-width unsigned multiplication is **not** proved here. A statement that declares it is refused rather than silently accepted. See *What remains* below.

## Step order

This is the order the code runs, step for step:

```text
commitment -> statement -> vanishing point -> batching coefficient
           -> batched zerocheck -> operand claims
           -> shift batching, public share subtracted
           -> bit sumcheck -> word sumcheck -> trace evaluation
           -> commitment opening
```

The two proved families are one batched zerocheck rather than two steps. Binius64's `reduce_constraints` is the reference design, but its unsigned-product and binary-product slots do not exist here: the relation language has no binary-field product family, and a statement declaring an unsigned product is refused at key construction.

**Why the order is sound.** Every step binds its claims before the challenge that consumes them.

- The trace commitment is bound before the vanishing point, so the committed words cannot follow it.
- Public words, every dimension, and every relation count are bound before both relation challenges.
- The four operand evaluations are bound before the shift reduction draws its two batching axes. This is the same rule Binius64 states for IntMul and BinMul running before BitAnd: the evaluations must be committed before the challenge that collapses them.
- The reduced trace value is bound before the commitment draws its opening challenges.

**The public check.** Public words enter the sponge as fixed statement data before any challenge, and the shift reduction subtracts their exact contribution instead of opening them. Binius64 instead has the prover state the public evaluation and reduces it through a ring switch, which is a verifier-cost trade, not a soundness one.

## The zero and bitwise relations

Both are local to one constraint row and one bit, so one check covers both:

```text
linear    value(row, bit) = 0
bitwise   left(row, bit) * right(row, bit) = output(row, bit)
```

One coefficient separates the two families, and one degree-three sumcheck over the shared cube binds the constraint coordinates then the within-word coordinates. Its final point is exactly the pair the shift reduction's claim type wants, so no re-randomisation is needed between the two.

Padded constraint rows hold the empty operand, which satisfies both shapes, so padding needs no selector.

## Degenerate parameters

Following the reasoning from #2226 — ask what determines a value, not what merely records it:

- A sampled batching coefficient of zero leaves the bitwise family entirely unchecked, so both sides refuse it rather than absorbing it. The refusal lives inside the transcript, which releases its own driver before returning; a refusal that dropped a transcript still owing pattern steps would panic in the destructor instead of rejecting, which is the defect #2244 fixed in the bus composition.
- A commitment too narrow to hold the padded trace is refused before anything is absorbed. A wider one is allowed: the word axis is padded up to the commitment's arity and the opening point is prefixed with zeros, so a statement with one committed 64-bit word — six variables, below the seven a Boolean commitment needs — is expressible.
- A statement declaring an unproved relation family has no key at all. `WordProofKey::new` rejects it, so an unprovable statement cannot be built, priced, or absorbed.
- A proof claiming a nonzero batched sum is rejected under its own error, distinct from the closing check.
- The vanishing wiring rejection from #2226 still applies underneath.

## Security accounting

`WordProofSecurityModel` composes, each separately labelled:

| Stage | Numerator over the challenge field |
| --- | --- |
| relation batching | families - 1 |
| vanishing point | zerocheck variables |
| vanishing rounds | degree x zerocheck variables |
| shift batching | operation + operand variables |
| shift sumchecks | 2 x (bit + word variables) |
| the commitment's own terms | whatever `opening_security(1)` reports |

Nothing in that table is supplied by a caller. The counts are the key's own, the challenge width is `EF::bits()` for the field the proof is instantiated at, and the ring switch and opening terms are taken from the commitment rather than recomputed from parameters the model cannot check against it.

**Every one of the proof's own five draws is charged over the candidates the commitment leaves open.** They all land after the commitment and before the opening names one, so a prover may pick which candidate it is after seeing them, and each owes a union bound over the whole set. The commitment's own terms are carried through untouched, since it already charged its own reductions over the same set. The union bound is linear, so charging per term costs exactly what charging the sum would. Against a unique-decoding commitment, which is every Boolean commitment on main, the charge is the identity and the reported level is unchanged at 121.54 bits; against one leaving 2^5.32 candidates open it falls from 121.54 to 117.09, which is the 4.45 bits that were previously reported too strong. The only input the model cannot derive is that candidate count itself, so a count naming no real set size yields no number at all.

Every degree comes from a named constant rather than a literal: the per-variable degree and the batched family count live beside the round prover, and the shift counts come from the existing model.

## Validation

- End-to-end proof and verification over **real** 64-bit and 32-bit statements, each with a zero relation and two AND relations across rotations, logical and arithmetic shifts, a lane-local sign-extending shift, a cancelling repeated term, and both public and committed operands. No empty constraint system anywhere on the end-to-end path.
- Every dependent witness word is derived with native integer operations, independently of the relation evaluator being checked.
- The four operand evaluations are compared against a direct evaluation built from the declared words.
- Adversarial: broken zero relation, broken AND relation, per-bit corruption of both a zero-relation word and an AND output at both ends and the middle, each of the four operand evaluations, every vanishing round, a nonzero claimed sum, a substituted public word, a commitment to a different witness, and a foreign trace opening substituted into an otherwise valid proof.
- **No rejection test settles for "an error happened".** The four `is_err()` assertions in the shift reduction are gone. Two were wrong about which check fires: perturbing a bit-round message is not caught by that sumcheck at all, because the omitted value at one is reconstructed from the running claim, so the round stays self-consistent and the phase merely ends elsewhere. The two claim-boundary checks also shared one error, so nothing could tell them apart; the first now has its own variant. Each of the three checks was then removed in turn to confirm the tests covering it fail — three of them previously passed with the check gone, because a later check happened to reject.
- **Every rejection test names the exact error, payload included.** No test settles for "some error of roughly this kind", so rejecting for the right reason is distinguished from rejecting by accident. The one wildcard in the suite is the commitment's own inner reduction error, which is opaque to this crate. Expected dimensions are hand-derived literals, not values recomputed through the code under test.
- **A consistent proof of a false statement.** Bumping a claimed sum is not a test of the zero-sum rejection: it moves what the sumcheck absorbs, so the replay ends elsewhere and the closing check rejects for an unrelated reason. The suite now builds the proof that actually threatens it — a broken witness, its true nonzero sum computed directly from the relation definitions, then an honest sumcheck, shift reduction and opening over that sum — and shows the closing check does close on it, so only the zero-sum rejection stands between it and acceptance. Deleting that rejection makes the verifier return `Ok(())` on a false statement, which the test reports.
- **A statement narrower than one committed element**, proved and verified through a wider commitment, and rejected when its single word is wrong, so the padding cannot be absorbing violations.
- **Transcript release.** One test is the reproduction of the drop-time panic, kept as a `should_panic` guard so the destructor check stays live; another drives both transcripts from a sponge that only returns zero, the only way to reach the degenerate branch, and fails if the release call is removed.
- **Swapping the left and right operand claims** leaves their product unmoved, so the vanishing check still closes; the test pins that the shift wiring is what rejects it. This is the "one relation masking another" case.
- Fiat-Shamir: the transcript tests hold all arithmetic fixed and vary only what is absorbed — a pre-absorbed commitment, each public word, each relation count, the committed segment length, the trace width — asserting the sampled challenges move. They also pin that the vanishing point is drawn before the batching coefficient.

## Known cost, not addressed here

The vanishing check materialises five dense challenge-field tables over the constraint-by-bit cube, so the prover's working set is proportional to `constraints * word bits` field elements. That is fine at the sizes tested and wrong at production sizes.

Collapsing the round polynomial's three passes over those tables into one was measured and does not pay; see the numbers on the review thread. The round is bound by challenge-field multiplication, not by the reads, and the multiplication count is the same either way.

Binius64 avoids it with a univariate skip over the bit axis and by keeping the operand columns bit-packed through the early rounds, which is what the shift reduction in #2226 already does for its own bit phase. Applying the same treatment here is a self-contained follow-up and does not change the protocol.

## What remains

Criterion 2 lists integer multiplication and binary multiplication as sub-reductions, and criterion 1 asks for full-width unsigned multiplication. Neither is in this PR, and this is a reported negative result rather than an omission.

Binius64 proves `a * b = hi * 2^64 + lo` by embedding it in the multiplicative group of the challenge field, so integer multiplication becomes exponent addition:

```text
(g^a)^b = g^c_lo * (g^(2^64))^c_hi
```

with a parity check on the least significant bits to close the `a * b = 0` against `c = 2^128 - 1` wraparound. Realising that needs five phases: a GKR product-tree reduction on the variable-base tree, an inverse-Frobenius twist, a batched selector-and-product sumcheck, three constant-base product trees, and a transparent logup-star lookup over a shared power table. Plonky3 has no grand-product GKR and no transparent logup-star today, and the lookup phase commits a second oracle mid-protocol, which cuts against "one final claim".

The alternatives I costed and rejected:

- Compiling the relation to bit-level AND and XOR gadgets is sound but needs roughly 500 extra committed words per 64-bit product. Binius64's stated gain over bit-level approaches is 64x; this would be several hundred times worse.
- Carryless multiplication alone does not decide the integer relation, and no characteristic-two argument checks an integer identity without either carries or the group embedding.

The slot, its ordering position, and the refusal are all in place, so adding the reduction later is additive.

## Composing with #2273

#2273 moves `Commitment`, `Proof` and `Error` onto a new `BooleanBackend<EF>` supertrait. The two branches merge without a textual conflict, but the fully qualified paths this PR used stopped resolving. They are gone: `prove` writes the shorthand `Pcs::Commitment` in a signature that already carries the bound, which resolves through the trait today and through the supertrait afterwards, and the test aliases name the concrete types the impl assigns. Verified by cherry-picking this branch onto `507b2cba`, where `cargo check -p p3-word-backend --tests` is clean. Neither PR has to land first.

Part of #2237.


Nashtare (commented):
I went through the reduction chain against the description, and the soundness argument holds in the code. The commitment is absorbed before the vanishing point on both sides. The dimensions, the three relation counts and the public words are bound in the domain separator before either draw. The four operand evaluations go in as the shift reduction's `claims` step before its two batching draws, and the trace evaluation is absorbed before the opening. The round message and the closing check are both `eq(τ, x)·(linear + α·(left·right − output))`. α is drawn from the full extension, and the verifier recomputes the eq factor itself. Padded rows are fixed by the key's wiring, not by anything the prover sends. The verifier computes the public share from its own words, and the committed trace only holds witness words, so no value can move between the two. I recomputed every numerator in the accounting table from the code and found nothing missing. The operand batching is the shift reduction's four-variable batch.

- The zero-sum check is the line that makes this a zerocheck, and no test pins it. Locally I deleted it: all 27 proof tests still pass, and a consistent proof of a broken zero relation verifies (inline, `verifier.rs`).
- The `DegenerateBatching` refusal panics on both sides instead of returning the error (inline, `verifier.rs`).

**Smaller things**
- #2273 merges with this branch without a textual conflict, but the combination doesn't compile. It moves `Commitment`, `Proof` and `Error` to a supertrait, and a type-check of the merged tree fails with five E0576 errors (inline, `record.rs`).
- The title probably wants a `!`. In a published crate, `PackedWord::Packing` becomes `Lane` plus `Packed<W>`, and `OperationColumns::bitwise_and` goes from two to three columns.

The other inline comments are non-blocking.

word-backend/src/proof/verifier.rs
```diff
@@ -0,0 +1,704 @@
...
+    fn a_nonzero_claimed_sum_is_rejected() {
+        // Mutation: claim a nonzero sum for a check whose whole point is that it vanishes.
+        let (system, public_words, words) = word64_statement(0xDDDD_EEEE_FFFF_0000);
+        let key = WordProofKey::new(system.clone()).unwrap();
+        let scheme = commitment_scheme(key.trace_variables());
+        let values = PackedWitness::new(&system, &public_words, &words).unwrap();
+        let (commitment, proof) = run(&key, &scheme, &values);
+
+        let mut tampered = proof;
+        tampered.zerocheck.claimed_sum += EF::ONE;
```
Nashtare, line 536: This test doesn't pin the check it is named after. Bumping the claimed sum of an honest proof changes what the sumcheck absorbs, so the replay ends on a different point and the closing check at line 86 rejects it anyway. Locally I deleted the `claimed_sum.is_zero()` check at line 66, and this test and the other 26 proof tests still passed. A proof of a false statement then verifies. With `words[0] ^= 1` in `word64_statement`, I ran `RelationZerocheck::prove` on the true (nonzero) sum, then an honest shift reduction and opening. The verifier returned `Ok(())` without the check and `Err(RelationClaim)` with it. Could this test build that consistent proof instead? Could the zero-sum rejection also get its own variant? At the moment both sites return `RelationClaim`, so a test can't tell which one fired.

word-backend/src/proof/verifier.rs
```diff
@@ -0,0 +1,704 @@
...
+            });
+        }
+
+        // The verifier reaches the prover's sponge state from the same commitment.
+        pcs.observe_commitment(commitment, challenger);
+        let variables = self.zerocheck_variables();
+        let mut transcript =
+            ProofVerifierTranscript::<_, F, EF>::new(challenger, self.transcript_shape(), public);
+        let (vanishing_point, batching) = transcript.challenges(variables);
+        if batching.is_zero() {
```
Nashtare: Returning here drops the relation transcript while it still owes the zerocheck bracket, and `VerifierState`'s `Drop` panics when that happens. I ran a local test that draws the two challenges and then returns `Err(DegenerateBatching)` the same way. It panics with `Dropped unfinalized VerifierState: 2 pattern step(s) were never replayed`. The prover at `prover.rs:104` does the same with `ProverState`. This only fires when the sponge outputs zero, so it can't be exploited. Still, the refusal the description advertises is a panic, and no test reaches it. The refusal isn't needed for soundness either. In characteristic 2, α = 1 is just as degenerate: if a zero-relation bit and a bitwise bit fail at the same (row, bit), the sum is `1 + 1·1 = 0`, and α = 1 isn't refused. Both values already fall inside the 1/|F| batching term. Could the refusal be dropped? If not, could the transcript be finished or aborted before returning, as the shift reduction does for `DegenerateWiring`?

word-backend/src/proof/record.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::ShiftReductionProof;
+
+/// Operand evaluations the vanishing check leaves for the shift reduction.
+///
+/// The order is the vanishing operand, then the left, right, and output operands.
+pub(super) const OPERAND_EVALUATIONS: usize = 4;
+
+/// One successful proof: the trace commitment beside its transcript record.
+pub type ProvedStatement<F, EF, Pcs, Challenger> = (
+    <Pcs as BooleanMultilinearPcs<EF, Challenger>>::Commitment,
```
Nashtare: #2273 moves `Commitment`, `Proof` and `Error` onto a new supertrait, `BooleanBackend<EF>`. The two branches merge without a textual conflict, but the fully qualified `<Pcs as BooleanMultilinearPcs<EF, Challenger>>::Commitment` stops resolving. `cargo check -p p3-word-backend --tests` on the merged tree reports E0576 here, on line 17, and on the three test aliases at `verifier.rs:142-144`. Whichever PR lands second will need `<Pcs as BooleanBackend<EF>>::…`.

word-backend/src/proof/key.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    /// Returns the exact algebraic soundness term for the complete proof.
+    ///
+    /// The width argument is a lower bound on the base-two logarithm of the challenge-field order.
+    ///
+    /// Every challenge of every sub-reduction must be drawn from that same extension field.
+    ///
+    /// The absorbed argument is the base-two logarithm of the bits one committed element holds.
+    #[must_use]
+    pub fn security_term(
```
Nashtare: `absorbed_log` and the regime both come from the caller. Nothing ties them to the commitment the proof runs through, or to each other. A regime built for another arity or field width would report a different number without any error. The model can't check the arity either, because `BinaryPcsRegime` doesn't expose it. `BooleanPcs::opening_security(1)` already returns exactly these two terms, the regime's opening term and the bit ring switch, from its own config. #2273 also adds `readings_security` to the trait. Could this take the commitment's terms from the PCS instead? Failing that, could it reject a regime whose `challenge_field_bits()` differs from `field_bits`? This also prices a key that declares unsigned products, even though `prove` and `verify` always refuse such a key. Refusing those in `WordProofKey::new` would close that gap.

word-backend/src/proof/key.rs
```diff
@@ -0,0 +1,302 @@
...
+        if count != 0 {
+            return Err(WordProofError::UnprovedRelation { count });
+        }
+        Ok(())
+    }
+
+    /// Rejects a commitment that covers a different hypercube from the padded trace.
+    pub(super) fn validate_arity<E>(&self, num_variables: usize) -> Result<(), WordProofError<E>> {
+        let expected = self.trace_variables();
+        if num_variables != expected {
```
Nashtare: The commitment's arity has to equal `word_variables + log2(W::BITS)` exactly, which makes small statements unprovable. `BooleanPcs::new` needs at least `ABSORBED` variables, which is 7 for `BinaryField128`. A Word64 statement with one committed word has 6 variables. A Word32 statement with one or two committed words has 5 or 6. Neither has a commitment it can use: construction fails with `WitnessTooNarrow`. Could the word axis be padded up to the commitment's arity instead? Fixing the extra leading coordinates to zero in the opening point reads only the real words, so the reduced claim carries over unchanged. If not, could the minimum be documented on `trace_variables`?

word-backend/src/proof/prover.rs
```diff
@@ -0,0 +1,217 @@
+//! Proving side of the ordered sub-reductions.
+//!
+//! # Sub-reduction order
+//!
+//! ```text
+//!     commitment -> unsigned product -> binary product -> bitwise -> linear
```
Nashtare: This order doesn't match what the code runs. There are no unsigned-product or binary-product steps. `bitwise` and `linear` are one batched zerocheck, not two steps. The public share is subtracted inside the shift reduction before its first sumcheck (`shift/mod.rs:267-269`), not after it. The next paragraph says the order is load-bearing, so could the diagram list the steps the code actually runs? Those are: commitment, statement, τ, α, zerocheck, operand claims, shift batching with the public share subtracted, the two shift sumchecks, trace evaluation, opening.

word-backend/src/proof/relation.rs
```diff
@@ -0,0 +1,180 @@
...
+    fn fold(&mut self, challenge: F) {
+        // Every column binds the same most-significant active variable.
+        self.equality.fix_prefix_var_mut(challenge);
+        self.linear.fix_prefix_var_mut(challenge);
+        for column in &mut self.bitwise {
+            column.fix_prefix_var_mut(challenge);
+        }
+    }
+
+    fn round_poly(&self) -> Vec<F> {
```
Nashtare, line 107: `round_poly` calls `round_at` three times, and each call is a full serial pass over the five dense tables. A single pass that produces the three node values per row would read each table once per round instead of three times. That pass could also be parallelised the way `fix_prefix_var_mut` is. Non-blocking, since the description already lists the dense layout as a follow-up.

word-backend/src/columns.rs
```diff
@@ -5,17 +5,17 @@ use core::array;
...
 
-/// Packed rows used by the AND and integer-multiplication reductions.
-///
-/// The AND output is absent because that reduction derives it from its two inputs.
+/// Packed rows used by the relation reductions.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct OperationColumns<W: PackedWord> {
-    /// The left and right inputs to each bitwise product.
-    bitwise_and: [Vec<W::Packing>; 2],
+    /// The operand each linear relation requires to vanish.
```
Nashtare, line 13: Nit: the module doc on line 1 still says "nonlinear-relation columns", but this field adds the linear family.

word-backend/src/proof/verifier.rs
tcoratger: Verified your claim first. With the `claimed_sum.is_zero()` check deleted, all 59 tests passed, this one included, so it pinned nothing.

Two changes.

The zero-sum rejection has its own variant now, `RelationSum`, distinct from `RelationClaim`. That alone makes the old test a real mutation detector: deleting the check moves the rejection to the closing check and the assertion fails.

And there is now the consistent proof you describe. `a_consistent_proof_of_a_false_statement_only_fails_the_zero_sum_check` flips a bit of `words[0]`, checks the statement really is unsatisfiable, then sums the batched relation over the cube directly from the two relation definitions rather than from the round prover, asserts that sum is nonzero, and runs `RelationZerocheck::prove` on it, followed by an honest shift reduction and an honest opening. Before asserting the rejection it replays the verifier far enough to show the closing check does close on this proof, so the test states in its own assertions that nothing downstream catches it. With the check present the verifier returns `Err(RelationSum)`; with the check deleted it returns `Ok(())`, which is the test's failure message.

word-backend/src/proof/verifier.rs
tcoratger: Reproduced before fixing. A test that draws the two challenges and returns the same way fails on the old code with

```
thread '...' panicked at challenger/src/fs/state/verifier.rs:71:13:
Dropped unfinalized VerifierState: 2 pattern step(s) were never replayed and 0 NARG byte(s) were never checked.
```

so the failure mode is exactly the one you describe, and the prover shape was live too.

I kept the refusal but moved it inside the transcript, because the call site is what kept getting this wrong. `challenges` now returns nothing when the coefficient vanishes and releases its own driver on the way out, so a caller cannot reach the early return with a live state. The prover transcript gained the `abort` it was missing. Both sides then read `.ok_or(DegenerateBatching)?`.

Two tests pin it. `abandoning_a_drawn_transcript_is_a_hard_failure` is your reproduction turned into a `should_panic` guard, so the drop check stays live. `a_vanishing_coefficient_releases_both_transcripts` drives both transcripts from a sponge that only ever returns zero, which is the only way to reach the branch at all; it fails with `Dropped unfinalized ProverState: 2 pattern step(s) were never played.` if the `abort` call is removed.

On dropping the refusal entirely: your argument is right that it buys no soundness. Batching two families under one coefficient is a degree-one polynomial in alpha, so the whole bad set is priced at 1/|F| whether or not zero is excluded, and alpha = 1 is in that set too. I kept it anyway because it is one comparison and it distinguishes "this coefficient checks nothing at all" from "this coefficient happened to cancel", which is worth a named error when someone is debugging a challenger. It is no longer a panic, which was the actual defect.

I swept the crate for the same shape. The other live one was in `prove`: `OperationColumns::new` was called after the transcript existed and used `?`. It does not depend on any challenge, so it now runs before the commitment, which removes the early return rather than patching it, and it subsumes the duplicate `check_shape` that preceded it. The shift reduction's own `prove` and `verify` already release on every path.

word-backend/src/proof/record.rs
tcoratger: Confirmed. I cherry-picked this branch onto `507b2cba` and `cargo check -p p3-word-backend --tests` reports the five E0576 you list, at `record.rs:16`, `record.rs:17` and `verifier.rs:142-144`.

Rather than pick one of the two spellings and break the other, I removed the need for a qualified path at all.

`ProvedStatement` is gone. `prove` now writes its return type as `(Pcs::Commitment, WordProof<F, EF, Pcs::Proof>)`, in a signature where `Pcs: BooleanMultilinearPcs<EF, Challenger>` is already a bound. The shorthand resolves through the trait today and through the supertrait after #2273, so neither PR has to touch the other. The alias could not do this: a type alias carries no bounds, which is why it needed the qualified form in the first place.

The three test aliases now name the concrete types the impl assigns, `BooleanProof<EF, MyMmcs, MyMmcs>`, `BooleanPcsError<EF, <MyMmcs as Mmcs<EF>>::Error>` and `<MyMmcs as Mmcs<EF>>::Commitment`. Those are the same on both branches, so they are neutral to which trait carries the associated type. `p3-commit` is a new dev-dependency for the `Mmcs` bound.

I re-ran the cherry-pick after the fix and `cargo check -p p3-word-backend --tests` is clean on top of #2273. Nothing here depends on #2273 landing first or second.

word-backend/src/proof/key.rs
tcoratger: Both taken, and the arity is derived rather than checked.

`security_term` and `security_components` no longer take `absorbed_log` or a regime. They take what the commitment charges, `&PrescribedOpeningSecurity`, which the caller gets from `pcs.opening_security(1)` on the same commitment the proof runs through. `WordProofSecurityModel` no longer recomputes the ring switch or the opening term from parameters it cannot check; it appends the commitment's own two terms. That also removes the duplicate formula: the ring-switch numerator was written in two places and only agreed by coincidence of arguments.

The field width is gone from the signature too. It is now `EF::bits()` for the `EF` the proof is instantiated at, so it cannot disagree with the field the challenges are drawn from. Both functions became infallible as a result: the only remaining `None` conditions were a zero field width and a zero composition degree, and neither is reachable once both come from types and constants.

On the unprovable key: `WordProofKey::new` now rejects a system declaring unsigned products, with `KeyCompileError::UnprovedRelation`. `WordProofError::UnprovedRelation` and `validate_statement` are gone, because no such key can exist to reach them. Pricing a statement that cannot be proved is now impossible rather than merely wrong.

word-backend/src/proof/key.rs
tcoratger: Relaxed, the way you suggest. `validate_arity` now rejects only a commitment narrower than the padded trace, the word axis is padded up to the commitment's arity, and the opening point is prefixed with one zero per extra coordinate.

The argument is the one you give: the leading coordinates are the most significant bits of the word index, the padding words sit above the declared ones and are zero, so fixing those coordinates to zero restricts the padded multilinear to exactly the smaller one and the reduced claim carries over. The shift reduction is untouched; the lift happens only where the opening is handed to the commitment.

The commitment's arity now varies independently of the statement, so it is absorbed into the relation transcript in place of the derived trace width, which was redundant with the committed-word count and the word width that are already bound.

`a_statement_narrower_than_one_committed_element_still_proves` is your Word64 case: one committed word, six trace variables, proved and verified through an eight-variable commitment. It also runs the same statement with the word set to a nonzero value and asserts the rejection, so the padding cannot be absorbing the violation.

word-backend/src/proof/prover.rs
tcoratger: Rewritten to the order the code runs, which is the list you give:

```text
commitment -> statement -> vanishing point -> batching coefficient
           -> batched zerocheck -> operand claims
           -> shift batching, public share subtracted
           -> bit sumcheck -> word sumcheck -> trace evaluation
           -> commitment opening
```

The two product slots and the separate bitwise and linear steps are gone, and the public share is where it actually happens. I will fix the same diagram in the PR description.

word-backend/src/proof/relation.rs
tcoratger: Measured, and it does not pay. I am leaving the three passes.

A/B over the same five dense tables, same arithmetic, criterion, one round at three cube sizes. The single pass computes the five lower/delta pairs once per row and evaluates all three nodes from them, so each table is read once per round instead of three times.

With `-C target-feature=+pclmulqdq`, which is what CI builds this stack with:

| cells | three passes | one pass | delta |
| --- | --- | --- | --- |
| 2^14 | 4.8037 ms | 4.8020 ms | -0.04% |
| 2^18 | 76.864 ms | 76.475 ms | -0.51% |
| 2^22 | 1.2743 s | 1.2651 s | -0.72% |

At 2^22 the five tables are 335 MB, well past any cache, and the gain is still under 1%.

Without `pclmulqdq` the single pass is *slower* at the two smaller sizes, +4.7% at 2^14 and +3.0% at 2^18, and 2.4% faster at 2^22. Holding five lower/delta pairs live across three nodes costs more registers than the reads it saves.

The reason is that the round is bound by challenge-field multiplication, not by the reads. Both versions do the same 24 multiplications per row: five interpolations, the operand product, the batching scale and the equality scale, times three nodes. The single pass removes loads and five subtractions, and those are not what the round spends its time on.

What would actually move it is the multiplication count, not the pass count. Node zero needs no interpolation at all, since the lower half is the value, so specialising it drops 5 of those 24. That belongs with the dense-layout follow-up rather than here, and I have not implemented it.

Parallelising the pass is orthogonal and still open; nothing here argues against it.

word-backend/src/columns.rs
tcoratger: Fixed: "Packed relation columns derived from a checked witness."

word-backend/src/proof/key.rs
```diff
@@ -0,0 +1,336 @@
...
+    ) -> WordProofSecurityModel {
+        // The counts are the key's own and the terms are the commitment's own.
+        let field_bits = NonZeroUsize::new(EF::bits()).expect("a field has at least one element");
+        WordProofSecurityModel::new(
+            field_bits.get(),
+            BATCHED_FAMILIES,
+            self.zerocheck_variables(),
+            ZEROCHECK_DEGREE,
+            self.shift.security_model(field_bits),
+            commitment.terms.clone(),
```
Nashtare, line 110: `security_model` keeps `commitment.terms` but drops `log2_max_candidates`. The `PrescribedOpeningSecurity` docs ask callers to union-bound their own reductions over that count, and multi-stark subtracts it from every reduction term. All five of this proof's own draws come after the commitment and before the opening fixes a candidate. Every Boolean commitment on main reports zero candidates, so nothing is mispriced today. But #2273's `BooleanWhir` also implements `BooleanMultilinearPcs`, and its Johnson profile reports `log_inv_rate + log2(10)` bits of candidates (about 5.3 at rate 1/4). Over that backend, each of the five terms here would read that much too strong. Could the model subtract `log2_max_candidates` from its own terms, or refuse a nonzero count? Once #2273 puts `readings_security` on the trait, taking `&Pcs` here would also tie `EF` to the proof. Right now `EF` is chosen at this call and bounded only by `Field`.

Nashtare (approved):
I re-checked both points I'd asked to have fixed, at `37c42e090`. The zero-sum rejection now has its own variant, `RelationSum`, and `verifier.rs:69` is the only place that returns it. Both tests that assert it would therefore fail if the check were deleted. The new one builds the consistent true-sum proof of a broken zero relation, and first shows that the closing check accepts it, so the test states what it pins.

For the transcripts, I went through every early return between creating a transcript and finishing it: the relation transcript on both sides, and the shift reduction's. Each one either aborts first or comes after `finish()`. The degenerate draw now aborts inside `challenges`, and the prover's `OperationColumns::new` runs before the commitment. Keeping the refusal as a named error is fine with me.

I type-checked this branch merged with current main (`p3-word-backend` and `p3-security`, all targets). I also type-checked it merged with #2273 at `6d9cebcc9`, adding `p3-binary-pcs`. Both are clean, so the two PRs now compose in either landing order.

The alias removal, the wider-commitment padding with its one-word test, the rewritten step order and the module doc close the other threads. The `round_poly` measurements answer that one. I'm also withdrawing the `!` point: neither `p3-word` nor `p3-word-backend` has been released yet, so no published API breaks.

One non-blocking follow-up is inline. The security model drops the commitment's candidate count. That's harmless for every Boolean commitment on main, but not once #2273's WHIR backend can be plugged in (inline, `key.rs`).

word-backend/src/proof/key.rs
tcoratger: Confirmed, fixed, and charged to all five.

## Which draws are charged

All five, and the commitment's own two terms are not.

The rule is that a draw made after the commitment and before the opening names one candidate gives a prover `2^log2_max_candidates` tries at it. Every challenge this proof makes is inside that window: `prove` produces the commitment itself, and the opening is the last thing it does, so the batching coefficient, the vanishing point, the vanishing rounds, the shift batching axes and the two shift sumchecks all land strictly between the two. Each of them is a Schwartz-Zippel event against a witness the candidate set has not yet pinned down, so each owes the union bound.

I looked for one that does not need it and there is none. None of the five is an event over public data alone; all five reduce, directly or through the shift wiring, to a claim about the committed trace, which is exactly the object the candidate set leaves ambiguous.

Splitting the charge per term is not an over-charge. The union bound is linear, so `N * (e1 + ... + e5)` is `N*e1 + ... + N*e5`; charging each term and summing gives the same total as charging the sum once. The per-term form is what makes a report say which stage is short.

What is *not* charged is `commitment.terms`. Those come back from `opening_security`, which has already charged the commitment's own reductions over the same set, and charging them again here would double-count. The two are extended into the list after the loop, untouched.

## Measured, before and after

Fixture is the eight-word Word64 key, rate 1/4, 40-bit target.

| term | settled commitment | 2^5.32 candidates open |
| --- | --- | --- |
| relation batching | 128.000 | 122.678 |
| vanishing point | 125.415 | 120.093 |
| vanishing rounds | 123.830 | 118.508 |
| shift batching | 126.000 | 120.678 |
| shift sumchecks | 123.830 | 118.508 |
| binary PCS opening | 123.093 | 123.093 |
| bit ring switch | 124.541 | 124.541 |
| **composed** | **121.541** | **117.092** |

The left column is what the model reported before this change and what it still reports, because every Boolean commitment on main is unique-decoding and the charge is then the identity. The right column is what it would have reported as 121.541 before: 4.448 bits optimistic. The composed figure drops by less than the 5.322 charged because the commitment's two terms hold still and are part of the union.

## Mechanism

I did not call `charge_reduction`, and this does not depend on #2273.

#2273 is still open, and this branch has to compile against main, where that method does not exist. Duplicating it into `p3-sumcheck` from here would collide with #2273 textually and is outside the crates I was asked to touch. So the same arithmetic sits on `SecurityTerm` in `p3-security`, the crate that owns the type: subtract, floor at zero, keep the label. Semantics match yours exactly, including the floor and the unique-decoding identity, and I pinned both with the same two cases your unit tests use. Once #2273 lands, `charge_reduction` is one line over it, and multi-stark's open-coded copy of the same subtraction is a third caller.

## On not consuming the count

Your rule is right for a stack of commitment layers, each charging its own reductions and forwarding the count upward. This is the end of that stack rather than a link in it: the window a candidate set is dangerous in opens when the commitment is produced and closes when the opening names one, and this proof brackets both ends itself. No caller can interleave a draw with it, so there is no layer above that could still exploit *this* set, and nothing is lost by consuming it. Say if you would rather it be reported anyway for a composer that does not exist yet.

## A count that names no set

A caller can hand over evidence with a negative, infinite or NaN count. That is the one input here neither the key nor the commitment derives, so it is checked: both entry points now return nothing rather than a number resting on it. That is the only fallibility left; everything else the model prices is derived.

## The `EF` point

Agreed, and not fixable here. Tying it needs `readings_security` on the trait so this can take `&Pcs`, which is yours. Today `EF` is the caller's word. The failure is one-sided in the safe direction if they pass a narrower field, since the reported level just drops, and unsafe only if they name a wider field than they prove over. I have left it, and it closes when #2273 lands.

## Assertions

`candidates_left_open_are_charged_to_every_draw_that_precedes_them` and `a_commitment_leaving_candidates_open_charges_every_draw_before_it` compare each charged term to its own settled counterpart by label, not the composed error and not a minimum. Removing the charge fails both on the first term, `word-relation-batching`, 128.000 against 122.678. I checked that before believing them.

PR #2276, opened 2026-09-21
refactor(commit): own the periodic column shape rule where every consumer can reach it

Bruce039:
`p3_air::check_constraints` / `check_all_constraints` exist to catch AIR mistakes before the real prover runs, but they don't look at the shape of periodic columns at all. `BaseAir::periodic_values` indexes with `row % period`, which works for any non-zero length, so an AIR that declares a periodic column of length 3 (or one longer than the trace) sails through the debug run with no violation and then dies in `fri/src/periodic.rs` with `periodic column length must be a non-zero power of 2, got 3` — the failure is real, it's just reported by the wrong layer, after the debug pass that was supposed to find it.

Writing that check revealed the wider problem: the same rule was spelled out seven times, in three inequivalent variants.

| site | power of two | second condition | on failure |
| --- | --- | --- | --- |
| `air/src/check_constraints.rs` (this PR) | yes | `height % period == 0` | panic |
| `uni-stark/src/verifier.rs` `check_periodic_column_lengths` | yes | `period <= trace_length` | `Result` |
| `fri/src/periodic.rs` `eval_on_lde_with_dft` | yes | `trace_len % period == 0` | panic |
| `fri/src/periodic.rs` `eval_at_point` | yes | `trace_len % period == 0` | panic |
| `commit/src/pcs/univariate.rs` `build_periodic_lde_table` | yes | `trace_size % period == 0` | panic |
| `circle/src/periodic.rs` `eval_on_lde` | yes | `trace_len % period == 0` | panic |
| `circle/src/periodic.rs` `eval_at_point` | yes | `trace_len % period == 0` | panic |

The verifier helper is the odd one out, and it is latently loose: fitting inside the trace is not the same as tiling it. A period of 8 over 12 rows leaves `[0..7][0,1,2,3|4..7]`, a half repetition the commitment layer refuses to evaluate.

## What this PR does

**One rule, owned by `p3-commit`, carried by a value the evaluator consumes.**

```rust
pub struct PeriodicColumns<'a, F> { /* private */ }

impl<'a, F> PeriodicColumns<'a, F> {
    pub fn new(columns: &'a [Vec<F>], height: usize) -> Result<Self, PeriodicColumnShapeError>;
}
```

`PolynomialSpace::evaluate_periodic_columns_at` takes that value in place of `&[Vec<Val>]`, so evaluating unscreened columns is not expressible. Every former assertion site becomes one construction: `commit/src/pcs/univariate.rs`, `circle/src/periodic.rs` (x2), `fri/src/periodic.rs`, and the two verifiers. Sites that panicked keep panicking, now on the `Err`.

`p3-commit` is the owner because it is the unique lowest common ancestor of every consumer:

```
p3-field, p3-matrix
  |-- p3-air         producer:  BaseAir::periodic_columns
  `-- p3-commit      consumer:  PolynomialSpace::evaluate_periodic_columns_at
        |-- p3-fri
        |     `-- p3-circle
        `-- p3-uni-stark -- p3-batch-stark
```

`p3-air` and `p3-commit` are siblings; neither depends on the other, and `p3-air` must not gain that edge. So `p3-air` keeps its screen as a **diagnostic** — an earlier, friendlier failure for an AIR author, not a precondition. Its public error enum is dropped: the only consumer panicked on it, every other rejection in that file is an `assert!`, and two structurally identical public enums in sibling crates is a trap for downstream matchers. The four pinned panic messages are unchanged.

Because the two rules are no longer the same function, `uni-stark/tests/periodic_shape_agreement.rs` compares their verdicts over a shared table of six heights and ten lengths, and asserts the table is not all-accept or all-reject so it cannot pass vacuously. It lives in `p3-uni-stark`, the lowest crate that sees both, so `p3-air` does not need even a dev-dependency on `p3-commit`.

## The tightened predicate rejects nothing that used to verify

Both verifier call sites pass a domain size — `init_trace_domain.size()` and `trace_domains[i].size()` — and every in-repo `PolynomialSpace` returns `1 << log_n`. On a power-of-two height the two candidate rules are the same predicate:

> Let `p = 2^a` and `n = 2^b`. Then `p | n` iff `a <= b` iff `p <= n`.

So no legal verdict moves. `on_a_power_of_two_height_fitting_and_dividing_agree` pins that equivalence over all `(a, b)` in `[0, 16)^2`, and `length_that_fits_but_does_not_divide_is_rejected` (8 over 12) fails the moment anyone widens it back to a size comparison. `every_period_dividing_the_trace_length_verifies` proves and verifies end to end for every period from 1 to 64 over a 64-row trace.

For a hypothetical backend whose domain size is not a power of two, the tightening is a real change — but only in the direction of a clean error instead of a panic: the loose bound admitted exactly the columns the periodic evaluator then refused.

## Test evidence

New coverage: 7 unit tests on the rule in `p3-commit`, 4 shape tests on the debug checkers in `p3-air`, the cross-crate agreement test, 1 end-to-end round-trip over every legal period in `p3-uni-stark`, and 2 `should_panic` tests in `p3-fri` — one per entry point — because its assertions had no coverage at all before.

Each assertion path was mutation-tested by deleting or widening it and checking which tests go red:

| mutation | tests killed |
| --- | --- |
| drop the power-of-two arm | 5 (`p3-commit`), 2 (`p3-fri`), 3 (`p3-uni-stark`), 1 (`p3-batch-stark`) |
| drop the divisibility arm | 4 (`p3-commit`), 1 (`p3-uni-stark`) |
| neuter the whole rule | the above, plus the agreement test on its vacuity guard |
| widen the commitment rule to `length <= height` | agreement test, at height 3 length 2 |
| widen the debug screen to `length <= height` | agreement test, at height 3 length 2 |
| make the debug checker's screen a no-op | 4 (`p3-air`) |

No mutant survives, and nothing unrelated goes red.

`cargo test` (debug and release) and `cargo clippy --all-targets -- -D warnings` are clean for `p3-commit`, `p3-air`, `p3-fri`, `p3-circle`, `p3-uni-stark`, `p3-batch-stark`, `p3-bus` and `p3-multi-stark --test whir_periodic`; `cargo +nightly fmt --check` and `scripts/check.py embedded` pass.

## Breaking changes

- `p3_uni_stark::PeriodicColumnError` becomes `p3_uni_stark::PeriodicColumnShapeError`, re-exported from `p3_commit`. Variant names and fields are unchanged; it stays `#[non_exhaustive]` and stays the payload of `VerificationError::PeriodicColumn`.
- `PolynomialSpace::evaluate_periodic_columns_at` changed its parameter type, so an out-of-repo override has to be updated. Both in-repo implementors are updated here; `evaluate_periodic_column_at` is untouched.
- The three bespoke panic strings in `commit`, `circle` and `fri` collapse into the one the rule reports, which also names the offending column.

## Deliberately left out

**`UnivariateStarkPcs::build_periodic_lde_table`.** The last prover-side entry point taking a raw slice. Threading the value through it means also changing `PeriodicEvaluator` and both `build_periodic_lde_table_*` helpers — a second public break for a path where all three in-repo backends already screen at their own entry.

**A per-column validated type.** Being a power of two is intrinsic to a column, but dividing the height is relational to one particular trace, so the relational half has to be established against a height either way — which is what this value does.


air/src/check_constraints.rs
```diff
@@ -477,6 +477,26 @@ fn boundary_io_label(cell: &BoundaryPublic) -> String {
 
 /// Evaluate every AIR constraint against a concrete trace and panic on failure.
 ///
+/// Panics when a periodic column cannot repeat over a trace of `height` rows.
+///
+/// A periodic column needs a period that is a non-zero power of two dividing the trace
+/// length. `BaseAir::periodic_values` indexes with `row % period`, which is defined for any
+/// non-zero period, so without this check the debug run accepts a shape the prover and the
+/// verifier reject much later, with the failure attributed to the commitment scheme.
+fn assert_periodic_column_shapes<F: Field, A: BaseAir<F>>(air: &A, height: usize) {
```
tcoratger: **[MAJOR] the new function was inserted inside another function's doc comment**

Line 478 (`/// Evaluate every AIR constraint against a concrete trace and panic on failure.`) was the summary line of `check_constraints`. Inserting the helper at line 486 leaves that line attached to the helper and hands `check_constraints` a doc block that now starts at line 500 with `/// The function walks the trace row by row`.

This is not cosmetic: a public function loses its summary, and a private helper advertises itself as the whole constraint evaluator. Rustdoc renders both wrong.

**Fix:** move the helper and its own doc above line 478, so `check_constraints`'s doc block is contiguous again.
**Verify:** `cargo doc -p p3-air --no-deps --document-private-items`, then read the first `<div class="docblock">` of `target/doc/p3_air/fn.check_constraints.html`. On this head it is "The function walks the trace row by row"; `fn.assert_periodic_column_shapes.html` starts with "Evaluate every AIR constraint against a concrete trace and panic on failure."

tcoratger (commented):
The gap is real and the fix belongs here. I reproduced the premise rather than taking it on trust: with the two `assert_periodic_column_shapes(air, height)` call lines commented out, both rejection tests report `note: test did not panic as expected`, so on `main` the fixtures really are accepted silently by both checkers and each new test genuinely fails without its production change. The period-16 fixture also does satisfy the AIR (`col[row % 16] == col[row] == row` for `row < 8`), so the claim in the description holds exactly as written. `cargo test -p p3-air` passes in debug and release, `cargo clippy -p p3-air --all-targets` is clean.

Three things to fix.

**The doc comment got split.** The new function was spliced between the first line of `check_constraints`'s doc block and the rest of it. `cargo doc -p p3-air --document-private-items` shows the damage: the public function's rendered summary is now *"The function walks the trace row by row"*, and the private helper's summary is *"Evaluate every AIR constraint against a concrete trace and panic on failure"*. That is a straight swap of two unrelated docs.

**The predicate is looser than the prover's, and looser than the helper's own doc says it is.** The doc says "a period that is a non-zero power of two **dividing** the trace length", which is right: `fri/src/periodic.rs` asserts `trace_len.is_multiple_of(period)` in both `eval_on_lde` and `eval_at_point`, and the trait doc on `periodic_columns` states "period p (where p divides n, both powers of 2)". The code only asserts `period <= height`. Those agree only when `height` is itself a power of two. `p3-uni-stark`'s `check_periodic_column_lengths` can use the weaker form because its `trace_length` is a domain size; the debug checker takes `main.height()`, which is whatever the caller passed.

On the boundaries I walked through, the current code is otherwise right:

| period | height | this PR | `fri` | agree |
|---|---|---|---|---|
| 0 | 8 | reject (not a power of two) | reject | yes |
| 1 | 8 | accept | accept | yes |
| 8 | 8 | accept | accept | yes |
| 9 | 8 | reject (both) | reject | yes |
| 8 | 12 | **accept** | **reject** | no |

And "at most the trace height" is genuinely required, not merely tidy: with `period > height`, `row % period == row` for every row, so the debug run only ever reads the first `height` entries of the column and silently evaluates a different AIR than the one the prover would build. Rejecting is correct.

**The test fixture is not a legal AIR.** It overrides `periodic_columns` to return one column but leaves `num_periodic_columns` at its default of `0`. `symbolic::builder` has a test named `validate_against_air_rejects_periodic_mismatch` for exactly that disagreement. A fixture should be well-formed in every respect except the one property under test.

Not from this PR, so not a request: `BaseAir::periodic_values` calls `periodic_columns()` once per row, and for any AIR returning `Cow::Owned` that clones the whole table `height` times inside both checkers' row loops. The one extra clone this PR adds is negligible next to that, but the per-row clone is worth its own issue.

I have pushed a commit to the branch with all of the above. Please amend or drop any part of it you disagree with.

air/src/check_constraints.rs
```diff
@@ -477,6 +477,26 @@ fn boundary_io_label(cell: &BoundaryPublic) -> String {
...
+    for (index, col) in air.periodic_columns().iter().enumerate() {
+        let period = col.len();
+        assert!(
+            period.is_power_of_two(),
+            "debug constraint check requires periodic column {index} to have a power-of-two length, got {period}"
+        );
+        assert!(
+            period <= height,
```
tcoratger: **[MINOR] the bound is weaker than what the prover enforces, and weaker than the doc four lines up**

Line 482 says the period must *divide* the trace length. This asserts only that it does not exceed it. `fri/src/periodic.rs` asserts `trace_len.is_multiple_of(period)` in both `eval_on_lde` and `eval_at_point`, and the trait doc on `periodic_columns` says "p divides n, both powers of 2".

The two predicates coincide only when `height` is a power of two. `p3-uni-stark`'s check can assume that because its `trace_length` is a domain size; here `height` is `main.height()`, whatever the caller handed over. A period-8 column on a 12-row trace passes this assert and fails in the commitment layer, which is the exact failure mode this function exists to pre-empt.

**Fix:** `assert!(height.is_multiple_of(period), ...)`, mirroring the two asserts in `fri/src/periodic.rs`. It still names the column index and both lengths, and it is a strict superset of the current bound for any non-empty trace.
**Verify:** add a case with `period = 8` and `height = 12`; it passes on this head and must be rejected after the change.

air/src/check_constraints.rs
```diff
@@ -477,6 +477,26 @@ fn boundary_io_label(cell: &BoundaryPublic) -> String {
 
 /// Evaluate every AIR constraint against a concrete trace and panic on failure.
 ///
+/// Panics when a periodic column cannot repeat over a trace of `height` rows.
+///
+/// A periodic column needs a period that is a non-zero power of two dividing the trace
+/// length. `BaseAir::periodic_values` indexes with `row % period`, which is defined for any
```
tcoratger: **[NIT] the doc block does not follow the repository's documentation rules**

Four points, all from the house style:

- Lines 482-483 and 483-485 each carry one sentence spread over two and three physical lines. One sentence per physical line is a hard rule; if it does not fit, split it.
- ``BaseAir::periodic_values`` is a symbol reference in doc prose. Describe the behaviour instead, so the doc does not go stale on a rename.
- "without this check the debug run accepts..." narrates the change rather than the code. A reader six months from now sees only the final code.
- The helper body has no inline comments; line-by-line narration is mandatory here, and the new tests carry only one comment between them.

**Fix:** rewrite as short bullets, one sentence per line, describing what the function rejects and why a non-power-of-two period has no evaluation subdomain.
**Verify:** `grep -n '\[`' air/src/check_constraints.rs` returns nothing new, and no `///` line in the block wraps a sentence.

air/src/check_constraints.rs
```diff
@@ -1339,4 +1362,59 @@ mod tests {
...
+    struct PeriodicCopyAir {
+        period: usize,
+    }
+
+    impl<F: Field> BaseAir<F> for PeriodicCopyAir {
+        fn width(&self) -> usize {
+            1
+        }
+
+        fn periodic_columns(&self) -> Cow<'_, [Vec<F>]> {
```
tcoratger, line 1410: **[MINOR] the fixture AIR contradicts itself on how many periodic columns it has**

`periodic_columns` returns one column, but `num_periodic_columns` is left at its default of `0`. Those two are meant to agree; `symbolic::builder::tests::validate_against_air_rejects_periodic_mismatch` exists precisely to reject that disagreement.

The debug checker never reads the count, so the tests pass, but a fixture that is malformed in a second, unrelated way weakens what the rejection tests prove. `uni-stark/tests/periodic_column_shape.rs` overrides both.

**Fix:** add `fn num_periodic_columns(&self) -> usize { 1 }` to the impl.
**Verify:** `cargo test -p p3-air periodic` stays green.

air/src/check_constraints.rs
```diff
@@ -1339,4 +1362,59 @@ mod tests {
...
+        // The trace itself satisfies the constraint, so only the shape check can reject it.
+        let air = PeriodicCopyAir { period: 3 };
+        check_constraints(&air, &periodic_copy_trace(3, 8), &[]);
+    }
+
+    #[test]
+    #[should_panic(expected = "periodic column 0 (length 16) to fit in the trace height (8)")]
+    fn test_periodic_column_longer_than_trace_is_rejected() {
+        let air = PeriodicCopyAir { period: 16 };
+        let _ = check_all_constraints(&air, &periodic_copy_trace(16, 8), &[], None);
```
tcoratger, line 1477: **[MINOR] the boundary cases the checker exists for are not covered**

What the three tests cover: period 2 accepted, period 3 rejected via `check_constraints`, period 16 rejected via `check_all_constraints`. What they miss:

- **A zero-length column.** This is the highest-value case: without the new assert, `periodic_values` evaluates `col[row % 0]` and dies on a division by zero with no mention of periodic columns. It cannot reuse the trace helper, since `i % 0` panics there too; build the trace with `BabyBear::zero_vec(8)`.
- **`period == height`** — the largest accepted period, one step from the rejection boundary.
- **`period == 1`** — a constant column, and the one power of two that is not even.
- The power-of-two rejection never runs through `check_all_constraints`, and the oversize rejection never runs through `check_constraints`. Since the whole point is that both entry points now assert, each should reject at least once.

**Fix:** add the zero-length and `period == height` cases, fold `period == 1` into the acceptance test, and cross the two rejection cases over the two entry points.
**Verify:** each new case must fail when the corresponding `assert!` is deleted; `cargo test -p p3-air periodic` green in debug and release.

tcoratger:
Pushed `53a7bf48` to this branch with all five points from the review. Everything is a follow-up commit on top of `44056010`, so feel free to amend, squash, or drop any part of it.

- The helper and its doc moved above the constraint checker, so the checker's summary line is its own again. `cargo doc` now renders "Evaluate every AIR constraint against a concrete trace and panic on failure." on the public function.
- The second assertion is now `height.is_multiple_of(period)`, mirroring `fri/src/periodic.rs` and the divisibility the `periodic_columns` contract already states. It is a strict superset of `period <= height` on any non-empty trace, so no previously accepted shape becomes rejected on a power-of-two height. The message still names the column index and both lengths.
- `num_periodic_columns` added to the fixture so it agrees with the table it returns.
- Tests: the acceptance case now sweeps periods 1, 2 and 8 over an 8-row trace (constant column, common case, period equal to the height), and two rejection cases were added, an empty column and a period-8 column on a 12-row trace. Each of the two entry points now exercises both assertions.
- Doc rewritten to the repository's style, and the helper body and the tests narrated inline.

Evidence:

- Deleting the divisibility assertion fails exactly the oversize and non-divisor tests; deleting the power-of-two assertion fails exactly the period-3 and empty-column tests. Neither over-fires.
- `cargo test -p p3-air` green in debug and release (197 tests), `cargo clippy -p p3-air --all-targets` clean, `cargo fmt` clean.
- `p3-bus`, `p3-uni-stark`, `p3-batch-stark --test simple` and `p3-multi-stark --test whir_periodic` all green, which covers the callers that declare periodic columns.

tcoratger (approved):
All five review points are addressed in `53a7bf48`, which I pushed to this branch. The debug checker now rejects exactly the periodic shapes the commitment layer rejects, and every assertion is pinned by a test that fails when that assertion is deleted. Nothing outstanding from my side.

tcoratger:
Heads up @Bruce039 — I pushed a commit to your branch and rewrote the PR title and description, so the scope is bigger than what you opened. Happy to unwind any of it if you'd rather keep this PR small.

Your diagnosis was the right one and it's untouched: the debug checkers read periodic columns through a wrapping row lookup, so a malformed one gets scored as if it were fine and only blows up later inside the commitment layer. Everything below is downstream of that.

What made me widen it: while checking that your new assertion matched the one the verifier already had, they turned out not to match. `check_periodic_column_lengths` requires `period <= trace_length`, not `trace_length % period == 0`, and that's a genuinely weaker rule — 8 over 12 rows passes it and then hits a panic in the periodic evaluator. Four copies of one predicate, three of them inequivalent, is the kind of thing that only stays consistent by luck, so I gave the rule an owner instead of adding a fifth copy:

- the validator and `PeriodicColumnError` moved into `p3-air`, next to `BaseAir::periodic_columns` which is where a periodic column is declared in the first place;
- `p3-uni-stark` re-exports both, so `p3_uni_stark::check_periodic_column_lengths` and `p3_uni_stark::PeriodicColumnError` still resolve for anyone outside the repo;
- `p3-batch-stark` takes it from `p3-air` directly;
- `p3-fri` keeps its own, because a commitment scheme depending on an AIR abstraction would be the wrong edge — its two in-file copies just collapse into one private helper, messages unchanged.

Two things worth arguing with if you disagree. The verifier's condition is now divisibility rather than a size bound, which I convinced myself changes no legal verdict (for `p = 2^a` and `n = 2^b`, `p | n` iff `p <= n`, and every domain in the tree reports a power-of-two size) and pinned with a test over all exponent pairs plus an end-to-end round trip for every legal period. And `PeriodicColumnError`'s variants changed shape to carry the column index, which is a break for anyone matching on them exhaustively — the path survives, the variants don't.

I also found three more copies of the same rule in `commit/src/pcs/univariate.rs` and `circle/src/periodic.rs`. Those sit below the AIR abstraction too, so folding them together needs a second owner in `p3-commit`; I left it out of here and wrote it up as follow-up in the description rather than grow this further.

Amend, revert or drop any of it — it's your PR.

tcoratger (approved):
Re-approving on 44eabd36. Nothing BLOCKER or MAJOR outstanding; all five earlier threads are resolved.

Verified on the new head: `cargo test` debug and release for `p3-air`, `p3-uni-stark`, `p3-batch-stark`, `p3-fri`, `p3-bus` and `p3-multi-stark --test whir_periodic`; `cargo clippy --all-targets -- -D warnings` clean on each; `cargo +nightly fmt --check` and `scripts/check.py embedded` pass. Each assertion path was deleted in turn and killed by exactly the tests that should cover it, with no surviving mutants and no unrelated failures.

tcoratger:
Pushed `983b485f`: the debug screen was bounded on `F: Field`, but it only reads column lengths, so it is now bounded on `F: Clone` — what the column accessor itself requires and nothing more.

I also considered adding an AIR-taking entry point so callers would not repeat the accessor call, and decided against it. Both verifiers keep the columns and evaluate them on the next line, and the accessor returns a `Cow`, so an AIR-taking form would rebuild the whole set a second time for every implementor that returns owned columns. It would help only the debug checker, which already has the right adapter as a private one-line wrapper. The slice form stays the primary because the columns are what the rule is about, and most callers already hold them.

tcoratger (approved):
Re-approving on the updated head. 204 tests green, clippy and fmt clean.

tcoratger:
One commit on top, `7b0de879`. It replaces "validate then use" for periodic columns with a value that carries the shape rule, so the check becomes unskippable rather than merely present. It is deliberately a single commit and can be dropped cleanly if you dislike the design.

## What it does

`p3-commit` gains a borrowing view plus its error:

```rust
pub struct PeriodicColumns<'a, F> { /* private */ }

impl<'a, F> PeriodicColumns<'a, F> {
    pub fn new(columns: &'a [Vec<F>], height: usize) -> Result<Self, PeriodicColumnShapeError>;
}
```

`PolynomialSpace::evaluate_periodic_columns_at` takes that value instead of `&[Vec<Val>]`, in both definitions (the default and the two-adic override). The four hand-written copies of the rule each become one construction:

| site | before | after |
| --- | --- | --- |
| `commit/src/pcs/univariate.rs` | inline `assert!` pair | construction, panics on `Err` |
| `circle/src/periodic.rs` `eval_on_lde` | inline `assert!` pair | construction, panics on `Err` |
| `circle/src/periodic.rs` `eval_at_point` | per-column `assert!` pair | construction, then the batched evaluator |
| `fri/src/periodic.rs` | private `assert_periodic_column_shapes` | construction in both entry points |
| `uni-stark`, `batch-stark` verifiers | `check_periodic_column_lengths(...)?` then evaluate | construction `?` then evaluate |

No call path constructs the view more than once per evaluation, and no path calls `periodic_columns()` twice. The two verifiers already held the columns and evaluated on the next line, so this replaces their two-step rather than adding a step.

## What becomes impossible to express

Calling the batched evaluator on columns nobody screened. The value cannot be constructed any other way, and its fields are private.

Two honest limits on that claim:

- `evaluate_periodic_column_at`, the per-column primitive, still takes a raw slice. It has to, because the batched default and the two-adic override are written in terms of it. It is now documented as the primitive behind the checked door, and after this commit its only callers are that door and the padded inner loop of the default LDE-table builder, which runs after a construction.
- The rule relates lengths to *one* height, so a view screened against the wrong height is still typed correctly. `evaluate_periodic_columns_at` carries a `debug_assert_eq!` that the recorded height matches the domain size. I considered moving the constructor onto `PolynomialSpace` so the height could never be supplied by the caller, and decided against it: every call site already has the right domain in hand, and a second constructor for one guard is more surface than the guard is worth.

## Why `p3-commit` owns it

`p3-air` and `p3-commit` are siblings; neither depends on the other. Confirmed from the manifests: `p3-air` pulls only `p3-field` and `p3-matrix`.

```
p3-field, p3-matrix
  |-- p3-air         producer:  BaseAir::periodic_columns
  `-- p3-commit      consumer:  PolynomialSpace::evaluate_periodic_columns_at
        |-- p3-fri
        |     `-- p3-circle
        `-- p3-uni-stark -- p3-batch-stark
```

`p3-commit` is the unique lowest common ancestor of every real consumer, which is exactly why copies of the rule survived in `commit`, `circle` and `fri` after the previous commit put the owner in `p3-air`. `p3-air` is the producer and had one consumer of the rule: its own debug checker.

## What happens to `p3-air`

Its screen stays, rescoped honestly as a diagnostic: an earlier, friendlier failure for an AIR author, not a precondition anyone relies on. It does not gain a dependency on `p3-commit`.

`air/src/periodic.rs` is deleted and the checker asserts directly. The public error enum does not earn its place any more:

- its only consumer panicked on the `Err`, so a `Result` API bought nothing;
- every other rejection in `check_constraints.rs` is an `assert!`, so panicking is the file's own idiom;
- two structurally identical public error enums in sibling crates is a trap for a downstream matcher, who would have to guess which one a given call path yields.

The four pinned panic messages are unchanged, so the existing `should_panic` tests still hold.

Replacing the shared function means the two rules can now drift, so `uni-stark/tests/periodic_shape_agreement.rs` compares them over a shared table of six heights and ten lengths. It lives in `p3-uni-stark` because that is the lowest crate that sees both, and putting it there avoids giving `p3-air` even a dev-dependency on `p3-commit`. It also asserts the table is not all-accept or all-reject, so it cannot pass vacuously.

## What breaks downstream

- `p3_air::PeriodicColumnError` and `p3_air::check_periodic_column_lengths` are gone. Neither exists on `main`; both were introduced by this PR.
- `p3_uni_stark::PeriodicColumnError` becomes `p3_uni_stark::PeriodicColumnShapeError`, re-exported from `p3_commit`. Variant names and fields are unchanged, so a matcher needs the new type name and nothing else. It stays `#[non_exhaustive]` and stays the payload of `VerificationError::PeriodicColumn`.
- **Implementors of `PolynomialSpace` outside this repo** are the real break: `evaluate_periodic_columns_at` changed its parameter type, so any override has to be updated. Both in-repo implementors are updated here. Overrides of `evaluate_periodic_column_at` are untouched.
- The three bespoke panic strings in `commit`, `circle` and `fri` collapse into the one the rule reports, which also names the offending column — `periodic column 0 has length 3, which is not a power of two` in place of `periodic column length must be a non-zero power of 2, got 3`. Two `should_panic` expectations in `p3-fri` follow it. This is the one place I went against the brief's "preserve the messages exactly": keeping three phrasings for one rule would preserve precisely the drift the commit removes, and the new message carries the column index the old ones lacked.

## Mutation evidence

| mutation | result |
| --- | --- |
| both arms of `PeriodicColumns::new` neutered | 5 red in `p3-commit`, 2 in `p3-fri`, 3 in `p3-uni-stark`, 1 in `p3-batch-stark`, plus the agreement test on its vacuity guard |
| commitment rule widened to `length <= height` | agreement test red: `height 3, length 2`, left `true`, right `false` |
| debug screen widened to `length <= height` | agreement test red: `height 3, length 2`, left `false`, right `true` |

## Verification

`cargo test` debug and release, green, for `p3-commit`, `p3-air`, `p3-fri`, `p3-circle`, `p3-uni-stark`, `p3-batch-stark`; plus `p3-multi-stark --test whir_periodic` and `p3-bus`. Release counts: 15 / 199 / 147 / 70 / 117 / 88, zero failures. `cargo clippy --all-targets -- -D warnings` clean on each, `cargo +nightly fmt --check` clean, `scripts/check.py embedded` passes.

## Not done, and why

`UnivariateStarkPcs::build_periodic_lde_table` is the remaining prover-side entry point that still takes a raw slice. Threading the view through it would also mean changing `PeriodicEvaluator`'s two methods and both `build_periodic_lde_table_*` helpers — a second public break, on the prover side, for a path where all three in-repo backends already screen at their own entry after this commit. It is the natural follow-up, not part of a droppable commit.

`fri`'s `eval_at_point` still hand-rolls Lagrange interpolation in `eval_periodic_poly` when the two-adic domain's batched evaluator would do it with one interpolation per period. That has no production caller and deleting it is a behaviour question, not a layering one.

tcoratger (approved):
Re-approving on `7b0de879`.

Verified on the new head: `cargo test` debug and release for `p3-commit`, `p3-air`, `p3-fri`, `p3-circle`, `p3-uni-stark`, `p3-batch-stark`, plus `p3-multi-stark --test whir_periodic` and `p3-bus`; `cargo clippy --all-targets -- -D warnings` clean on each; `cargo +nightly fmt --check` and `scripts/check.py embedded` pass.

The layering claim checks out against the manifests: `p3-air` pulls only `p3-field` and `p3-matrix`, so it cannot see `p3-commit`, while `p3-fri`, `p3-circle`, `p3-uni-stark` and `p3-batch-stark` all can. That is why three copies of the rule survived the previous commit and none survive this one.

Mutation-tested the new guarantee rather than taking it on trust. Neutering both arms of the constructor turns 11 tests red across four crates and trips the agreement test's vacuity guard. Widening either side of the rule to a size comparison fails the agreement test at height 3, length 2, in the direction matching whichever side was perturbed — so the two rules really are pinned to each other and not merely both present.

Confirmed no call path builds the value twice per evaluation, and no path calls `periodic_columns()` twice. The remaining raw-slice entry point is `build_periodic_lde_table`, which every in-repo backend now screens at its own entry; leaving it for a follow-up is the right call for a commit that has to stay droppable.

Bruce039:
Thanks for taking it this far — no need to unwind anything. I read the four commits: giving the rule a single owner in `p3-commit` and making the evaluator consume a `PeriodicColumns` value is a better fix than the extra assertion I started with, and the agreement test pins the diagnostic in `p3-air` to it, which was the drift I was worried about. The `PolynomialSpace` signature change is the right trade for that guarantee. Fine to squash on merge from my side.

PR #2277, opened 2026-09-21
fix(symmetric): reject a non-power-of-two root count when deserializing a MerkleCap

Bruce039:
`MerkleCap` keeps one invariant: the number of roots is a power of two. `MerkleCap::new` asserts it, `From<Vec<_>>` goes through `new`, `MerkleTree::cap` slices `1 << cap_height` roots, and `height()` computes `log2_strict_usize(num_roots())` on the strength of it. The one way to build a cap that skipped the check was `#[derive(Deserialize)]`, which wrote the `Vec` straight into the struct. Since `MerkleCap` is the `Commitment` type of `MerkleTreeMmcs` and `MerkleTreeHidingMmcs`, every FRI/STIR/WHIR proof carries caps that arrive through exactly that path, so a proof could hand the verifier a cap with three roots (or zero) that no constructor would have produced. Nothing on the verification path calls `height()` today (`verify_batch` bounds `cap_index` by `num_roots()`), so this is closing a gap rather than fixing a live panic — but the invariant is documented and asserted everywhere else, and a `Deserialize` that bypasses it is the same class of hole as a field element deserializing out of range.

Change: `Deserialize` is now a hand-written impl that deserializes the same wire shape (same field names, same struct name, so postcard/bincode/JSON bytes are unchanged) and returns `invalid_length` for a root count that is not a power of two. `Serialize` stays derived.

Tests: a four-root cap round-trips through `serde_json` and reports `height() == 2`; three roots and zero roots are rejected with an error naming the power-of-two requirement. `cargo nextest run -p p3-symmetric -p p3-merkle-tree -p p3-fri -p p3-uni-stark` passes, including uni-stark's stored compatibility fixtures, which confirms the encoding did not move. `serde_json` is added as a dev-dependency of `p3-symmetric` for the tests.


symmetric/src/hash.rs
```diff
@@ -175,6 +201,29 @@ mod tests {
         assert_eq!(cap.height(), 3);
     }
 
+    #[test]
+    fn test_merkle_cap_deserialize_round_trips() {
```
tcoratger: **MINOR — this round trip cannot detect the regression it is described as guarding against.**

It serializes and deserializes with the same revision of the code, so it is self-consistent by construction: change both halves of the shape together and it still passes. Any test of the form `decode(encode(x)) == x` is blind to the encoding moving.

What pins an encoding is a fixed byte string. Since the format that actually carries proofs is postcard and not JSON, the known-answer test should be in postcard:

```
    cap   = [[1;4],[2;4],[3;4],[4;4]]
    bytes = 04 01010101 02020202 03030303 04040404
            ^^ varint root count, then the roots, no framing for the marker
```

I verified those bytes are identical on `main` and on this head, so they are a valid anchor. `p3-uni-stark`'s stored fixtures do cover this end to end, but they live in another crate, so someone editing this impl can get a green `p3-symmetric` while breaking every stored proof.

I will push this.

symmetric/src/hash.rs
```diff
@@ -175,6 +201,29 @@ mod tests {
...
+        let json = serde_json::to_string(&cap).unwrap();
+        let back: MerkleCap<F, Digest> = serde_json::from_str(&json).unwrap();
+        assert_eq!(back, cap);
+        assert_eq!(back.height(), 2);
+    }
+
+    #[test]
+    fn test_merkle_cap_deserialize_rejects_non_power_of_two_root_count() {
+        // The wire shape `new` would refuse: three roots, then zero.
+        for roots in ["[[1,1,1,1],[2,2,2,2],[3,3,3,3]]", "[]"] {
```
tcoratger: **MINOR — the rejection test only exercises the self-describing map path.**

JSON hands the derive a map and the derive takes the map branch of its visitor. postcard hands it a sequence and takes the other branch. The sequence branch is the one every real proof goes through, and it is untested here.

```
    json     -> visit_map  -> tested
    postcard -> visit_seq  -> untested
```

Both branches feed the same length check, so this is coverage rather than a live bug, but the format that matters is the one left out.

Two smaller points on the same test:

- Asserting on `err.to_string().contains("power-of-two")` couples the test to the wording of the message. Checking that decoding fails, plus one assertion elsewhere that the message names the requirement, separates the two concerns.
- `alloc::format!` is inline-qualified; the file's convention is to import at the top of `mod tests`, as the neighbouring `alloc::vec` import does.

I will push a postcard rejection case covering three roots and zero roots.

symmetric/src/hash.rs
```diff
@@ -21,14 +22,38 @@ pub struct Hash<F, W, const DIGEST_ELEMS: usize> {
 ///
 /// A cap of height 0 contains a single element (the root), while a cap of height `h` contains
 /// `2^h` elements. The `Digest` type is the full digest (e.g. `[W; DIGEST_ELEMS]`).
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+///
+/// The number of roots is always a power of two: [`MerkleCap::new`] asserts it and the
```
tcoratger: **MINOR — this violates the repo's documentation rules on three counts.**

- `[`MerkleCap::new`]` is an intra-doc link, and doc text here may not reference type or function names, links included. Rename the target and the doc goes stale silently.
- Naming the `Deserialize` impl describes the enforcement machinery rather than the rule, and a reader who has not seen this diff has no reason to care which impl does the checking.
- Two sentences share a physical line, where the house rule is one sentence per line.

The deeper miss is that it documents *that* the invariant holds without saying *why* it has to. The reason is concrete: the cap is the layer of a binary tree at a fixed depth, its size is therefore two to the power of that depth, and the height is recovered by taking the base-two logarithm of the root count. A count that is not a power of two has no height, which is what makes it meaningless rather than merely unusual.

Suggested shape:

```
/// The root count is always a power of two.
///
/// A cap is one full layer of a binary tree, so it holds 2^h digests for a layer at depth h.
///
/// The height is recovered as log_2 of the root count, which only exists for a power of two.
///
/// Every path that builds a cap enforces this, including the one that reads it off the wire.
```

I will push this.

symmetric/src/hash.rs
```diff
@@ -21,14 +22,38 @@ pub struct Hash<F, W, const DIGEST_ELEMS: usize> {
...
+/// `Deserialize` impl rejects anything else, so a cap read from the wire upholds it too.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
 #[serde(bound(serialize = "Digest: Serialize"))]
-#[serde(bound(deserialize = "Digest: Deserialize<'de>"))]
 pub struct MerkleCap<F, Digest> {
     cap: Vec<Digest>,
     _marker: PhantomData<F>,
 }
 
+/// The derived wire shape of [`MerkleCap`], deserialized before the root count is checked.
```
tcoratger: **MINOR — same doc-link rule, and the sentence narrates the change rather than the code.**

`[`MerkleCap`]` is a name reference. "deserialized before the root count is checked" is sequencing that only makes sense to someone reading this diff; six months from now the reader wants to know what this type is for, which is that it mirrors the public type's fields so the derive can produce the wire shape without the public type owning an unchecked constructor.

Two other things worth doing while here:

- Give the two fields header docs. Every other struct in this file documents its fields, and these two are exactly where the wire shape is defined, so they are worth naming.
- Add a short comment recording that the field names, their order, and the container rename are what keep the bytes identical to the derive. That is the load-bearing fact about this type and nothing in the code states it.

I will push both.

symmetric/src/hash.rs
```diff
@@ -21,14 +22,38 @@ pub struct Hash<F, W, const DIGEST_ELEMS: usize> {
...
+#[derive(Deserialize)]
+#[serde(rename = "MerkleCap")]
+#[serde(bound(deserialize = "Digest: Deserialize<'de>"))]
+struct MerkleCapRepr<F, Digest> {
+    cap: Vec<Digest>,
+    _marker: PhantomData<F>,
+}
+
+impl<'de, F, Digest: Deserialize<'de>> Deserialize<'de> for MerkleCap<F, Digest> {
```
tcoratger, line 57: **MINOR — the impl has no `# Errors` section.**

This is a fallible boundary on untrusted input and the one place in the crate that turns a malformed proof into an error rather than a panic. It should say what it rejects and why, in the standard rustdoc section the repo uses elsewhere.

One line of body comment is also missing: the check is the whole point of the impl existing, and the code around it reads as a plain delegation.

I will push this.

symmetric/src/hash.rs
```diff
@@ -21,14 +22,38 @@ pub struct Hash<F, W, const DIGEST_ELEMS: usize> {
...
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let MerkleCapRepr { cap, _marker } = MerkleCapRepr::<F, Digest>::deserialize(deserializer)?;
+        if !cap.len().is_power_of_two() {
+            return Err(D::Error::invalid_length(
+                cap.len(),
+                &"a power-of-two number of Merkle cap roots",
```
tcoratger, line 77: **NIT — the enforcement sites now disagree about how loudly they fail.**

Deserializing three roots yields `invalid length 3, expected a power-of-two number of Merkle cap roots`. Passing three roots to the constructor yields `assertion failed: cap.len().is_power_of_two()`, and the constructor's doc comment does not mention that it panics at all.

Same invariant, same violation, two very different experiences, and the documented one is the path that is harder to hit. Since this PR is about making the invariant explicit, giving the constructor a `# Panics` section and an assertion message that states the requirement finishes the job.

I will push this.

symmetric/Cargo.toml
```diff
@@ -22,6 +22,7 @@ p3-goldilocks = { path = "../goldilocks" }
 p3-koala-bear = { path = "../koala-bear" }
 
 proptest.workspace = true
+serde_json.workspace = true
```
tcoratger, line 26: Confirmed dev-only and safe: the workspace sets `resolver = "2"`, so the `std` feature this pulls into `serde` stays on the dev side and cannot reach the `no_std` build of the crate. `Build embedded` on thumbv7em is green on this head.

I am adding `postcard` alongside it, for the known-answer and sequence-path tests noted above. It is already a workspace dependency and it is the format the proofs are actually written in, so the crate that defines the wire shape is the right place to pin it.

tcoratger (commented):
Reviewed as a wire-format change on the commitment type that every FRI/STIR/WHIR proof carries, so the bar was byte-level evidence rather than reasoning. The change is correct and the compatibility claim holds. Findings are all MINOR: the tests do not actually pin the encoding, and the new docs break the repo's documentation rules.

No reviewer comments existed on this PR before this one.

## Wire compatibility — verified by execution, both directions

Two worktrees, `main` at `00fe0a98` (derived `Deserialize`) and this head at `5ac86bcf` (hand-written), each with an integration test that writes the encoding of a four-root cap to disk and reads back the other side's bytes.

Encodings of `MerkleCap<Goldilocks, [u8; 4]>` over `[[1;4],[2;4],[3;4],[4;4]]`, byte-identical across the two revisions:

```
postcard  04 01010101 02020202 03030303 04040404
bincode   0400000000000000 01010101 02020202 03030303 04040404
json      {"cap":[[1,1,1,1],[2,2,2,2],[3,3,3,3],[4,4,4,4]],"_marker":null}
```

Cross-decode matrix, all four cells pass:

```
    bytes from main  -> decoded on this head : postcard, bincode, json  OK
    bytes from head  -> decoded on main      : postcard, bincode, json  OK
```

That is the expected result and the mechanism is sound: the shadow struct carries the same two fields in the same order, and `#[serde(rename = "MerkleCap")]` keeps the name handed to `deserialize_struct`, which matters for the formats that actually write the struct name (RON and friends) and for the `expecting` text in errors. Dropping that rename would have been the silent break, and it is there.

`p3-uni-stark`'s stored fixtures are the real end-to-end guard and they run and pass: `verify_two_adic_compat_fixture` and `verify_circle_compat_fixture` decode `tests/fixtures/uni_stark_*_v0_8_0.postcard`, which contain caps.

Suites run on this head: `cargo test -p p3-symmetric -p p3-merkle-tree -p p3-fri -p p3-uni-stark` (debug), `cargo test --release -p p3-symmetric -p p3-merkle-tree`, and `cargo clippy -p p3-symmetric -p p3-merkle-tree --all-targets`. All clean.

## The other claims in the description, checked

**No remaining construction bypass.** Both fields are private, and `hash.rs` is the only module that can name them. Struct-literal construction appears exactly twice in the crate: in `new` (which asserts) and in the new `Deserialize` (which now checks). Everything else funnels through `new`: `From<Vec<_>>`, `From<Hash<_, _, _>>`, and `MerkleTree::cap`. There is no `Default`, no `DerefMut`, no mutable root accessor, and no `MerkleCap { .. }` literal anywhere else in the workspace. `Clone` cannot manufacture a bad value from a good one. The hole this closes was the only one.

**Zero roots.** `0.is_power_of_two()` is false, so zero is rejected — and that is right, not a regression. `log2_strict_usize(0)` asserts and panics (`0.wrapping_shr(64) == 0 != 1`), `new` already rejects the empty vector, and `MerkleTree::cap` always produces at least one root. An empty cap is illegal everywhere else in the tree, so rejecting it here is the consistent choice.

**No live panic.** Confirmed. Every `height()` call in the workspace is inside a test. The two verification sites that touch a cap's length both bound the index instead of taking a logarithm: `verify_batch` guards with `cap_index < commit.num_roots()`, and `verify_batch_pruned` guards with `cap_idx >= commit.num_roots()`. Nothing in FRI, STIR, WHIR or `p3-binary-pcs` calls `height()` or `log2_strict_usize` on a commitment. STIR's transcript shape takes a root count but only to emit that many absorption steps, which is total for any count. So this is a gap being closed, as the description says, and the severity is correctly stated.

Worth adding for the record: an oversized cap would not have been exploitable either. The cap is absorbed before the query indices are drawn, and the index into it is derived from verifier-known dimensions, so extra roots are unreachable and a short cap just fails to match.

**Dev-dependency.** `serde_json` is dev-only, the workspace is on `resolver = "2"` so its `std` feature cannot unify into the normal dependency graph, and `Build embedded` (thumbv7em) is green on this head. No leak.

**Error choice.** `invalid_length` is apt: the wrong thing genuinely is the length of a sequence, and it renders as `invalid length 3, expected a power-of-two number of Merkle cap roots`. The two existing validating impls in this workspace (`Goldilocks`, `SymbolicExpr`) both reach for `custom`, so this is a slight departure, but the more structured variant is the better call here and the expected-value string does name the requirement.

## Findings

Six MINOR, no BLOCKER or MAJOR. The two that matter are the test ones: nothing in this crate would catch the encoding moving, which is precisely the risk the description sets out to retire.

tcoratger:
Pushed `44198ecf` on top of your commit, addressing all six findings. Your commit is untouched.

**Tests.** The round trip became a known-answer test. It now asserts the exact postcard bytes `04 01010101 02020202 03030303 04040404` and the exact JSON text, then decodes them back — which is what catches the encoding moving, whereas encode-then-decode never can. I confirmed those bytes are byte-identical on `main`, so they are a valid anchor rather than a snapshot of this branch.

The rejection test now drives both decoding paths, since JSON hands the derive a map and postcard hands it a sequence, and the sequence is the one every proof takes. The assertion on the error wording moved into its own test so a message tweak does not look like a validation failure.

**Docs.** Removed the intra-doc links, split the multi-sentence lines, and made the doc say *why* the invariant exists — a cap is one full tree layer, so a layer at depth `h` has `2^h` digests, and the height is the base-two logarithm of that count, which only exists for a power of two. The shadow struct now records that its field names, field order and container name are what keep the bytes identical, gave its fields header docs, and the decoder gained an `# Errors` section.

**Constructor.** Gave it a `# Panics` section and an assertion message naming the requirement, so both enforcement sites now say the same thing when they fire.

**Dependency.** Added `postcard` as a dev-dependency for the known-answer and sequence-path tests. It is already a workspace dependency and it is the format the proofs are written in, so this crate is the right place to pin it.

Each new assertion was mutation-tested. Deleting the power-of-two check fails both rejection tests; renaming the field on the encoder or on the decoder fails the known-answer test.

Green locally: `cargo test -p p3-symmetric -p p3-merkle-tree -p p3-fri -p p3-uni-stark` in debug and release (including the stored uni-stark fixtures), `cargo clippy --all-targets`, `cargo +nightly fmt --check`, and `scripts/check.py embedded --target thumbv7em-none-eabi` to confirm the dev-dependency stays off the `no_std` build.

tcoratger (approved):
Approving. The compatibility claim is verified byte-for-byte in both directions across postcard, bincode and JSON, no other path can build a cap that skips the invariant, and the absence of a live panic is confirmed by the fact that every height call in the workspace is in a test while both verification sites bound the index instead. With the follow-up commit the encoding is now pinned by known bytes inside the crate that defines it, and both decoding paths are covered.

tcoratger:
Pushed `89fd0df1` — the substring assertion was too weak, as it passes on any error that merely contains the phrase.

The self-describing path now pins the whole message, which keeps the offending count in it:

    invalid length 3, expected a power-of-two number of Merkle cap roots
    invalid length 0, expected a power-of-two number of Merkle cap roots

The compact path needed a different treatment. Postcard discards the custom text and reports `Serde Deserialization Error` for everything, so pinning that string would assert a catch-all that a truncated buffer produces just as readily. Instead each rejected count is paired with the next legal one — three roots refused beside four accepted, zero refused beside one accepted — so the count is demonstrably what the refusal turns on.

Both tests were mutation-checked: disabling the invariant check fails exactly these two and nothing else.

Amend or drop as you like.

tcoratger (approved):
Re-approving on the updated head. The rejection tests now pin the full message on the self-describing path and prove causation on the compact path, and both were confirmed to fail with the invariant check disabled.

PR #2278, opened 2026-09-21
test(multi-stark): cover a proof and an AIR set that disagree on an optional section

Bruce039:
A `MultiStarkProof` carries three optional sections — `preprocessed_opening`, `lookup`, `indexed` — and each is present exactly when the AIR set declares the corresponding feature. `verify` checks that agreement before touching the section (`MissingPreprocessedOpening` / `UnexpectedPreprocessedOpening`, `Lookup(MissingProof)` / `Lookup(UnexpectedProof)`, `UnexpectedIndexedReduction`), but nothing in the test suite exercised those branches: the existing negative tests corrupt the *contents* of a section on a proof that legitimately has one, never its presence.

Six tests, one per section and direction, in `whir_fibonacci.rs` and `whir_preprocessed.rs`:

- an honest Fibonacci proof gets `preprocessed_opening = Some(opening.clone())` → `UnexpectedPreprocessedOpening`; an honest preprocessed proof gets `preprocessed_opening = None` → `MissingPreprocessedOpening`;
- an honest lookup section grafted onto the Fibonacci proof → `Lookup(UnexpectedProof)`; the lookup AIR's proof with `lookup = None` → `Lookup(MissingProof)`;
- an honest indexed section grafted onto the Fibonacci proof, and the squares batch's proof with `indexed = None` → `UnexpectedIndexedReduction` in both directions, which is how the verifier's `match` reports either disagreement.

I checked they are load-bearing by deleting the presence checks in `verifier.rs` and `lookup/mod.rs`: the corresponding pair of tests goes red each time (the missing-preprocessed case as a panic further down, which is the failure mode the check exists to prevent). The whole `whir_fibonacci` / `whir_preprocessed` / `whir_periodic` set passes (58 tests); clippy clean.


multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1109,6 +1110,233 @@ fn verify_rejects_tampered_main_commitment() {
 const WHIR_FIXTURE: &str = "tests/fixtures/multi_stark_whir_v0_8_0.postcard";
 
 /// A fixed Fibonacci instance shared by the WHIR compat-fixture generator and checker.
+/// An honest Fibonacci proof over 256 rows, with everything `verify` needs to check it.
```
tcoratger: **MINOR — this steals the doc comment of the helper below.**

Line 1112 is `/// A fixed Fibonacci instance shared by the WHIR compat-fixture generator and checker.`, and it belongs to `whir_compat_case` at line 1340. The new helper is inserted between the two, so that sentence now reads as the first line of *this* helper's doc — where it is simply untrue, this builds a proof and has nothing to do with the fixture generator — and `whir_compat_case` is left undocumented.

The block as a whole is also wedged between `const WHIR_FIXTURE` and the compat-fixture helpers it belongs to, splitting that section in half. Move the whole new block (helpers and tests) to just above `const WHIR_FIXTURE` at line 1110, which puts it after the existing negative tests where it belongs and hands the orphaned doc line back.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1109,6 +1110,233 @@ fn verify_rejects_tampered_main_commitment() {
...
+            &pis,
+        )]),
+        0,
+        &mut challenger(),
+    )
+    .unwrap();
+    (config, vk, log_height, pis, proof)
+}
+
+/// An honest proof of the 64-row `LocalPermutationLookupAir`, which carries a lookup section.
```
tcoratger: **MINOR — `/doc`: no function or type names in doc text.**

`` `LocalPermutationLookupAir` `` here and `` `verify` `` on line 1113 are identifiers in prose. The convention in this repo is to describe what the thing *is*, not to name the item: "the local-permutation lookup AIR, whose interactions force a lookup section into the proof". Same for line 1246 (`indexed-lookup section` is fine, that one is already clean).

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1109,6 +1110,233 @@ fn verify_rejects_tampered_main_commitment() {
...
+    verify(
+        config,
+        VerifierInstances::new(vec![VerifierInstance::new(&FibAir, vk, log_height, pis)]),
+        proof,
+        0,
+        &mut challenger(),
+    )
+}
+
+// The three optional sections of a proof each answer a declaration made by the AIR set:
```
tcoratger: **MINOR — `/doc`: a four-line prose block with sentences wrapped across lines.**

```
// The three optional sections of a proof each answer a declaration made by the AIR set:
// preprocessed columns, interactions, indexed reads. A proof and an AIR set that disagree on
// whether a section is present must be rejected with the error that names the section, in
// both directions, before anything in the section is read.
```

Two sentences spread over four lines, each wrapping mid-clause. House style is one sentence per physical line, and no run of three or more prose lines — a labelled line plus an indented map reads better here:

```
// Invariant: a section is present exactly when the AIR set declares the feature.
//
//     preprocessed columns -> the preprocessed opening
//     interactions         -> the lookup section
//     indexed reads        -> the indexed reduction
//
// A disagreement is rejected before the section is read, in both directions.
```

Note also that this intro sits in this file only, while the sixth case of the six it describes lives in `whir_preprocessed.rs`.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1109,6 +1110,233 @@ fn verify_rejects_tampered_main_commitment() {
...
+        &mut challenger(),
+    )
+    .unwrap_err();
+    assert!(
+        matches!(err, VerificationError::Lookup(LookupError::MissingProof)),
+        "expected Lookup(MissingProof), got {err:?}"
+    );
+}
+
+/// An honest proof of the squares batch, which carries an indexed-lookup section.
```
tcoratger: **MINOR — helper declared in the middle of the test group.**

`honest_fibonacci`, `honest_lookup` and `verify_fibonacci` are grouped at the top of the block; this one lands between the lookup tests and the indexed tests. Move it up with the other three so the block reads helpers-then-tests.

multi-stark/tests/whir_fibonacci.rs
```diff
@@ -1109,6 +1110,233 @@ fn verify_rejects_tampered_main_commitment() {
...
+// preprocessed columns, interactions, indexed reads. A proof and an AIR set that disagree on
+// whether a section is present must be rejected with the error that names the section, in
+// both directions, before anything in the section is read.
+
+#[test]
+fn verify_rejects_an_unexpected_preprocessed_opening() {
+    let (config, vk, log_height, pis, mut proof) = honest_fibonacci();
+    assert!(proof.preprocessed_opening.is_none());
+
+    // Mutation: attach a well-formed opening as the preprocessed one on an AIR set that
```
tcoratger: **MINOR — the test bodies are missing the narration every other negative test in these two files carries.**

Compare `verify_rejects_tampered_main_commitment` just above, or `verify_rejects_tampered_preprocessed_opening` in `whir_preprocessed.rs`. Both open with `// Fixture state:` over the setup and precede the `verify` call with `// Expected rejection:` and a `// Why:` line. The six new tests carry only the `// Mutation:` line, so a reader arriving cold gets the tamper but neither what state is being tampered nor why the rejection is the right one.

The `// Why:` line is the one that earns its place here, because it is where the interesting fact goes: the check matches on presence alone and never reads the section, which is why grafting an opening built for a different commitment is still a legitimate test of presence.

multi-stark/tests/whir_preprocessed.rs
```diff
@@ -639,6 +639,49 @@ fn verify_rejects_violated_main_constraint() {
...
+            &pk,
+            &[],
+        )]),
+        0,
+        &mut challenger(),
+    )
+    .unwrap();
+    assert!(proof.preprocessed_opening.is_some());
+
+    // Mutation: drop the opening the verifying key's preprocessed commitment requires.
```
tcoratger: **MINOR — same narration gap, and this is the case worth narrating most.**

This one reddens under its mutation through a panic deeper in the verifier rather than through the assertion below, which is the point of the check: without it the verifier walks into an `expect` at `verifier.rs:385`. A `// Why:` line saying that the rejection has to come *before* the opening is consumed would record why the check sits where it does, and would distinguish this test from `verify_rejects_tampered_preprocessed_opening` fifty lines down, which tampers with contents rather than presence.

The setup above also repeats `verify_rejects_tampered_preprocessed_opening` verbatim down to the last line. That matches the existing duplication in this file so I would not block on it, but a small `honest_preprocessed()` helper would pay for itself the next time a third test needs the same fixture.

tcoratger (approved):
I reproduced the mutation claim rather than taking it, deleting each of the five presence checks in turn and running the whole `whir_fibonacci` / `whir_preprocessed` / `whir_periodic` set each time. The result is a clean 1:1 map — every mutation reddens exactly the test(s) named for it and nothing else.

| check deleted | replaced by | tests red | how they fail |
|---|---|---|---|
| `verifier.rs:199` `(Some(_), None) -> MissingPreprocessedOpening` | `(Some(_), None) => {}` | `verify_rejects_a_proof_missing_its_expected_preprocessed_opening` | panic at `verifier.rs:385`, `"missing preprocessed opening rejected before verification"` |
| `verifier.rs:200` `(None, Some(_)) -> UnexpectedPreprocessedOpening` | `(None, Some(_)) => {}` | `verify_rejects_an_unexpected_preprocessed_opening` | `verify` returns **`Ok(())`** |
| `verifier.rs:331-337` catch-all `-> UnexpectedIndexedReduction` | `_ => None` | `verify_rejects_an_air_declaring_an_indexed_read_with_no_indexed_proof` **and** `verify_rejects_an_indexed_proof_for_an_air_declaring_no_indexed_read` | first: panic at `instance.rs:65`, `"an indexed batch needs the reduction that produced it"`; second: `verify` returns **`Ok(())`** |
| `lookup/mod.rs:769` `(None, Some(_)) -> UnexpectedProof` | `=> return Ok(None)` | `verify_rejects_a_lookup_proof_for_an_air_declaring_no_interaction` | `verify` returns **`Ok(())`** |
| `lookup/mod.rs:770` `(Some(_), None) -> MissingProof` | `=> return Ok(None)` | `verify_rejects_an_air_declaring_an_interaction_with_no_lookup_proof` | `verify` returns `Err(Zerocheck(LookupLinkMismatch { air: 0 }))` — **the wrong variant, not `Ok`** |

Three of the five make `verify` accept outright, which is the strongest evidence one can get that the rejection is about *presence* and nothing else: with the check gone, the grafted section is not merely tolerated, it is invisible. That also settles the "is the graft well-formed?" question in the only way that matters here — the check matches on `Some`/`None` alone and never reads the payload, and the mutation confirms no downstream stage objects to the payload either.

The last row is the interesting one. Under that mutation the verifier still returns an error, so a bare `is_err()` assertion would have stayed green and the test would have been decorative. It goes red only because the test pins `Lookup(MissingProof)` with `matches!`. All six tests assert a specific variant with a `matches!` plus a message carrying `{err:?}` — none of them settles for `is_err()` — so the suite survives the one case where it had to.

On the missing-preprocessed case the description is accurate: it reddens through an internal `expect` deeper in the verifier rather than through the assertion. That is not a weakness of the test, which does pin `MissingPreprocessedOpening`; it is the defence-in-depth `expect` firing first, which is exactly the outcome the check exists to prevent.

All six combinations are distinct — the 1:1 mapping above is the proof. The two indexed tests share a variant but not a match arm direction, so neither is redundant.

Cost is nil: on this machine `whir_fibonacci` runs in 2.24s debug with the five new tests and 2.24s with them skipped, i.e. below measurement noise; release is 0.17s for the same 40 tests. `cargo clippy -p p3-multi-stark --all-targets` and `cargo fmt --check` are clean.

No correctness objection, no BLOCKER, no MAJOR. What is left is placement and narration, and one of those does real damage to an existing doc comment.

**MINOR — `multi-stark/src/verifier.rs:333`, a comment this PR's own test falsifies.** The catch-all arm is introduced by:

```rust
// The shape describes the bracket exactly when the AIRs declare a read.
//
// The two disagree only if the proof carries a section nobody asked for.
```

The second line is false. That arm also catches `(Some(plan), None)` — an AIR set that declares an indexed read against a proof with no indexed section — and `verify_rejects_an_air_declaring_an_indexed_read_with_no_indexed_proof` is precisely a test of that direction. Deleting the arm sends that case into a panic in `instance.rs`, so the arm is load-bearing in both directions while the comment claims only one. The variant's own `#[error]` string already says "present but not expected, or absent but described"; the comment contradicts it.

**MINOR — one variant for two different disagreements.** `UnexpectedIndexedReduction` is returned for both "the proof carries a section nobody asked for" and "the AIRs declare a read the proof does not answer". A caller cannot tell a malformed proof from a mismatched AIR set, and the variant name only describes one of the two. This is pre-existing and not this PR's to fix, but the PR is the first thing to make the ambiguity visible, and the two neighbouring pairs (`Missing`/`Unexpected` for preprocessed, `MissingProof`/`UnexpectedProof` for lookup) both split it. Splitting it would cost one variant and make the two indexed tests assert different things.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Fixed in 5759de07. The whole block now sits above the fixture constant, and that doc line is back on the compat-fixture helper.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Fixed in 5759de07. Both helper docs are rewritten without naming an item, and a grep for a backtick inside a doc line over the added region returns nothing.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Fixed in 5759de07, as the invariant line plus the arrow map suggested.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Fixed in 5759de07. All four helpers now precede the five tests.

multi-stark/tests/whir_fibonacci.rs
tcoratger: Fixed in 5759de07. Each of the five carries the fixture state and the mutation as a two-line presence table, plus the reason line. The lookup-missing one records what the mutation run showed: with the check gone the argument is skipped and the zerocheck reports a link mismatch instead, which is why that test has to pin the variant rather than settle for an error.

multi-stark/tests/whir_preprocessed.rs
tcoratger: Narration added in 5759de07, including why the rejection has to land before the opening schedule is described, and a line separating this test from the contents-tampering one further down. I left the duplicated fixture in place rather than extracting a helper, since every test in this file is written that way and a lone helper would be the odd one out.

tcoratger:
Pushed `5759de07` on top of your commit — no test was added, removed, or weakened, and the mutation matrix is byte-for-byte the same after as before.

Three things:

- The block was sitting between `/// A fixed Fibonacci instance shared by the WHIR compat-fixture generator and checker.` and the helper that line documents, so the line had become the first line of `honest_fibonacci` and `whir_compat_case` was left bare. Moved the block above the fixture constant, which also stops it splitting the compat-fixture section in two, and grouped `honest_indexed` with the other three helpers.
- Gave the six tests the `// Fixture state:` / `// Mutation:` / `// Expected rejection:` / `// Why:` narration the other negative tests in these two files carry, with the presence disagreement shown as a two-line table. The helper docs no longer name functions or types.
- One line of `verifier.rs`, comment only: the note above the indexed catch-all arm said "The two disagree only if the proof carries a section nobody asked for", which `verify_rejects_an_air_declaring_an_indexed_read_with_no_indexed_proof` disproves. Deleting that arm sends the declared-read-with-no-section case into a panic at `instance.rs:65`, so the arm is load-bearing in both directions. Now it lists both.

Worth recording from the mutation run, since it is the reason the variant assertions are not decorative: with `lookup/mod.rs:770` deleted, `verify` still returns an error — `Zerocheck(LookupLinkMismatch { air: 0 })` — so a bare `is_err()` would have stayed green there. Pinning `Lookup(MissingProof)` is what makes that test bite.

`whir_fibonacci` / `whir_preprocessed` / `whir_periodic`: 58 pass, debug and release. Clippy `--all-targets` and `fmt --check` clean. Runtime is unchanged at 2.24s debug for `whir_fibonacci` whether the new tests run or are skipped.

tcoratger (approved):
Re-approving on `5759de07`, which carries my narration and placement fixes on top of the original six tests.

Re-ran the five deletion mutations against the pushed head: the mapping is unchanged — one test red per check, two for the shared indexed arm, and `Lookup(MissingProof)` still the only one that survives on `is_err()` alone. 58 tests pass debug and release, clippy `--all-targets` and `fmt --check` clean, CI green (19 success, 1 skipped).

PR #2279, opened 2026-09-21
fix(multi-stark): state the lane-group bound on the sparse packed fold

0xShadowX:
### What

`RoundStateExt::fold` folds each column with one of two packed kernels. The dense kernel goes through `Poly::fix_prefix_var_to_packed`, which (since #2264) asserts that the low half of the column holds at least one lane group. The sparse kernel in the `else` branch builds the same packed layout by hand — `(0..half).step_by(WIDTH)` with `column.value(start + lane)` — and needs exactly the same `half >= WIDTH`, but stated nothing: it is correct today only because `fold_columns` selects the packed pair strictly above that size (`want_packed = half / 2 >= WIDTH`).

This adds the matching `debug_assert!` to the sparse branch with a message naming the bound, so the requirement is written at the second place that relies on it. No behaviour change on any current path; the assertion cannot fire from `fold_columns`.

### Test

Nothing new to pin — the guard upstream makes the assertion unreachable through the public API, which is the point of documenting it as a `debug_assert!` rather than a runtime check. `cargo nextest run -p p3-multi-stark --lib` (292) and the `whir_*` integration suites pass; clippy clean.


multi-stark/src/rounds.rs
```diff
@@ -1990,6 +1990,14 @@ where
                     PolyView::new(values).fix_prefix_var_to_packed(r)
                 } else {
                     let half = column.len() / 2;
+                    // Each lane group reads `WIDTH` rows of the low half, so the half must hold
+                    // at least one group; `fix_prefix_var_to_packed` asserts the same bound on
+                    // the dense side, and `fold_columns` only picks this kernel above it.
```
tcoratger: MINOR, style only.

The three lines pack one sentence across two physical lines, run to three contiguous prose lines, and name two functions in comment text, all of which `/doc` rules out.
The cross-reference also narrates where the requirement comes from rather than what holds here.

```suggestion
                    // Invariant: the low half holds at least one full lane group.
                    // Halves are powers of two, so one group then divides the half exactly.
```

tcoratger (approved):
Approving.

**The stated bound is the true one.** The loop `(0..half).step_by(WIDTH)` emits a final partial group whenever `WIDTH` does not divide `half`, and `ColumnView::value` hard-asserts `row < len`, so a partial group would read the high half for `lo` and run off the end for `hi`. So the requirement is really "`half` is a whole number of lane groups", and `half >= WIDTH` is only equivalent to that because `half` and `WIDTH` are both powers of two: column length is `1 << num_variables`, and a packing width is a power of two by construction. That is the same reasoning the dense side spells out, where the two stacked constraints collapse to `n > log_2(W)`. So the assertion is exact here, not too weak — though it is the power-of-two structure, not the inequality alone, that rules out the partial tail.

**It is unreachable.** The sparse packed kernel is a closure with a single consumer: it reaches `fold_each_column` only through the packed arm of `fold_columns`, guarded by `want_packed = (half / 2) >= WIDTH`. That is strictly stronger than the asserted `half >= WIDTH` by a factor of two, and `fold_columns` has exactly one other caller, the subfield fold, which shares the same guard because the guard lives in the callee. Nothing else constructs this layout by hand.

**`debug_assert!` is the right call, and the asymmetry with #2264 is justified.** The dense bound sits on a public method of another crate that any caller can reach with a too-short polynomial, so it earns a hard assert. This one sits inside a private closure whose only selector already enforces twice the bound, so a release-mode check would buy nothing on a path that cannot be taken.

**No test needed** — agreed, and for the reason given: the only way to exercise the assertion is to call the kernel outside `fold_columns`, which the private closure makes impossible, so a test would have to duplicate the kernel rather than test it.

One `/doc` nit on the comment, inline. Local `cargo test -p p3-multi-stark --lib` in debug (292 passed) and `cargo clippy -p p3-multi-stark --all-targets` are clean.

tcoratger:
Pushed `9c0ec5cc` applying the `/doc` nit from the review: the three-line comment above the packed-fold assertion is now two lines, one sentence each, stating the invariant the low half satisfies rather than naming the callers that establish it. The assertion, its condition and its message are untouched — this is a comment-only change. Feel free to amend or drop that commit if you prefer different wording.

tcoratger (approved):
Re-approving on `9c0ec5cc`. The only change since the previous approval is the two-line comment above the `debug_assert!`; the bound `half >= F::Packing::WIDTH` and the rest of the patch are unchanged. `cargo check`, `cargo clippy --all-targets` and `cargo test` on `p3-multi-stark` are clean, and `cargo +nightly fmt` leaves no diff.

Issue #2233, opened 2026-09-19
Publish reproducible binary-backend scoreboards


tcoratger:
## Deliverable

Publish comparable end-to-end measurements for the binary backend and freeze the configurations those measurements use.

## Acceptance criteria

- Measure witness generation, proving, serialization, verification, proof size, and peak memory.
- Report single-threaded and configured multi-threaded results.
- Cover x86-64 and Apple silicon where executable hardware is available.
- Use fixed workloads comparable to Flock and the Flock challenge.
- Record the complete proven security report beside every result.
- Add narrow regression gates for proof size and critical performance budgets.

gopikannappan:
I'd like to take this on, if it's open. I have an Apple M2 (and x86-64 through Rosetta for cross-checks), and I've worked with Flock and the Flock challenge, which the issue names as the comparison workload.

**What's there today.** `BinaryProofReport` reports rows, width, stacked variables, proof size, prove time, verify time and composed security. Missing against the acceptance criteria are witness generation, serialization, peak memory, single- versus multi-threaded runs, a machine-readable format, frozen workloads, and the regression gates.

**A plan in three steps**, so each is reviewable on its own:

1. **Measure what's missing.** Add witness-generation and serialization timings and a thread count to the report, plus `--format json` for machine-readable output. Peak memory I'd capture in the harness by wrapping the run (`/usr/bin/time -l` on macOS, `-v` on Linux), so no new dependency goes into the crate.
2. **Freeze the workloads and publish a scoreboard.** A fixed set of shapes across the `prove_hash_binary` objectives, run single-threaded and multi-threaded, with the full security report beside every row, on Apple silicon and x86-64.
3. **Regression gates.** Narrow budget checks for proof size and the critical timings, in the style of the existing bench smoke job.

Two questions before I start:

- **Which workloads do you want frozen?** I'd propose the `prove_hash_binary` objectives at a couple of heights each, but you may already have shapes you compare against Flock.
- **Where should the scoreboard live:** a markdown file in the repo, CI output, or somewhere else?

I'll start on step 1 meanwhile, since it doesn't depend on either answer.

PR #2280, opened 2026-09-21
feat(examples): measure witness, serialization and threads in the binary report

gopikannappan:
Part of #2233.

The scoreboard needs numbers the report doesn't collect yet. This adds the missing measurements and a machine-readable form to collect runs with. It changes no proving or verifying behaviour.

## What the report gains

- **Witness generation.** The trace is built before the prover is called, so the caller records it: `BinaryProofReport::with_witness_seconds`. `prove_hash_binary` now does this for all three objectives.
- **Serialization and deserialization**, timed around the postcard round trip the run already does.
- **Thread count.** The Rayon pool's size, or one without `parallel`. `p3-examples` gains the `parallel` feature the other crates already have, so it can ask.
- **`--format json`**: one object per run, carrying the same measurements.

Reading one run:

```
Rows: 512                  Hash: keccak-256
Width: 23712               Merkle leaf: 16 field elements (256 bytes)
Threads: 8                 Witness time: 0.019s
Proof size: 825258 bytes   Prove time: 0.108s
Serialize time: 0.001s     Deserialize time: 0.001s
Verify time: 0.070s        Composed security: 97.12 bits
```

```json
{"rows":512,"width":23712,"stacked_variables":24,"hash":"keccak-256","leaf_elements":16,
 "requested_leaf_elements":null,"proof_bytes":825258,"prove_seconds":0.108271,
 "verify_seconds":0.07345975,"security_bits":97.11837503647102,"serialize_seconds":0.000594792,
 "deserialize_seconds":0.000787125,"threads":8,"witness_seconds":0.018778625}
```

## Two choices worth flagging

**Peak memory is not measured in-process.** The acceptance criteria ask for it, but reading peak RSS portably needs a new dependency, and the number is available for free by wrapping the run (`/usr/bin/time -l` on macOS, `-v` on Linux). I'd rather the harness in the next step capture it than add a dependency here. Happy to do it in-process instead if you prefer.

**The thread count is the Rayon pool, not the machine's core count**, so it reflects `RAYON_NUM_THREADS` and is the number a single- versus multi-threaded row should report. Checked: with `RAYON_NUM_THREADS=1` it reports one thread and proving slows from 0.047 s to 0.070 s at 2⁹ BLAKE3 compressions.

## Testing

- `the_report_measures_every_stage` asserts every stage is filled in, that `with_witness_seconds` survives, and that every field appears in the JSON, so a later field addition can't silently miss the scoreboard.
- It also pins the JSON name of the hash to the one the human output prints, since a scoreboard collects both forms.
- `p3-examples` passes 34/34 on an Apple M2, with nightly clippy `-D warnings` and fmt clean. Both feature sets build.

Still to come on #2233: the frozen workloads and the published scoreboard, then the regression gates. I asked in the issue which workloads you want frozen and where the scoreboard should live.


tcoratger:
@gopikannappan CI is failing, could you check?

gopikannappan:
@tcoratger CI is fixed now. sorry about that.

tcoratger (commented):
Adds witness, serialization and thread measurements to the binary report and a `--format json` output for #2233. It changes no proving or verifying code. I checked the timers bracket the right regions and the JSON names match the human output.

The main issue is that JSON mode also writes the tracing tree and progress lines to stdout, so the JSON cannot be parsed as is.

- **[MAJOR] Conflicts with `main`.** Since this branch was cut, `main` added `setup_seconds`, `pcs: PcsIdentity` and `whir: Option<WhirSummary>` to `BinaryProofReport` (`examples/src/binary.rs`), and moved witness generation into `preflight_then_maybe_prove` closures (`examples/examples/prove_hash_binary.rs`). A rebase has to derive `Serialize` on `PcsIdentity`, `WhirSummary` and `WhirRegime`. It should also time the witness once, around `generate()`, not three times.

**Bottom line:** merge after fixes. I'm pushing them to the branch now, rebased on `main`.

examples/examples/prove_hash_binary.rs
```diff
@@ -180,15 +195,23 @@ fn main() {
...
-        }
+        Ok(report) => match args.format {
+            OutputFormat::Human => {
+                println!("{report}");
+                println!("Proof Verified Successfully");
+            }
+            OutputFormat::Json => println!(
+                "{}",
+                serde_json::to_string(&report).expect("the report serializes")
+            ),
```
tcoratger, line 436: **[MAJOR] `--format json` does not produce parseable stdout**

Standard output carries more than the JSON object:

- the `tracing_forest` span tree, since `ForestLayer::default()` uses `PrettyPrinter`, which writes to stdout (line 114);
- the `Proving N ...` line printed before the witness is built (lines 151, 169, 186).

A scoreboard harness that pipes the run into a JSON parser fails on the first line.

```
$ cargo run --release --example prove_hash_binary -- -o blake-3-compressions -l 6 --format json 2>/dev/null | head -2
Proving 64 Blake-3 compressions
INFO     generate packed Blake3 binary trace [ 224µs | 100.00% ]
```

**Fix:** in JSON mode, send progress lines and the span tree to stderr (`ForestLayer::from(PrettyPrinter::new().writer(io::stderr))`), keeping human mode unchanged.

**Verify:** `... --format json 2>/dev/null | python3 -c 'import json,sys; [json.loads(l) for l in sys.stdin]'` succeeds.

examples/src/binary.rs
```diff
@@ -387,8 +390,23 @@ impl BinaryProofOptions {
...
+fn worker_threads() -> usize {
+    p3_maybe_rayon::prelude::current_num_threads()
+}
+
+/// Worker threads the prover can use: without `parallel`, everything runs on the calling thread.
+#[cfg(not(feature = "parallel"))]
+const fn worker_threads() -> usize {
+    1
+}
+
```
tcoratger: **[MINOR] Thread count keyed on the wrong crate's feature**

The `cfg` reads `p3-examples/parallel`, but the prover runs on whichever backend `p3-maybe-rayon` was compiled with.

Cargo unifies features, so `p3-maybe-rayon/parallel` can be on while `p3-examples/parallel` is off (e.g. `--features p3-multi-stark/parallel`).

The report then prints `threads: 1` for a run that used the full Rayon pool, which mislabels exactly the single- versus multi-threaded rows this field exists for.

**Fix:** drop both definitions and call `p3_maybe_rayon::prelude::current_num_threads()` unconditionally.

The serial fallback already exports it as a `const fn` returning 1.

**Verify:** assert `report.threads == current_num_threads()` in the report test.

examples/src/binary.rs
```diff
@@ -1216,6 +1266,51 @@ mod tests {
...
+            "proof_bytes",
+            "prove_seconds",
+            "verify_seconds",
+            "security_bits",
+            "serialize_seconds",
+            "deserialize_seconds",
+            "threads",
+            "witness_seconds",
+        ] {
+            assert!(json.contains(field), "{field} missing from {json}");
```
tcoratger: **[MINOR] The report test cannot fail on the things it names**

- `serialize_seconds >= 0.0` and `deserialize_seconds >= 0.0` always hold for a `Duration`, so they assert nothing.
- `json.contains(field)` is a substring search over the whole string, and the list is not exhaustive (`hash`, `leaf_elements`, `requested_leaf_elements` are missing), so a new field can still skip the scoreboard silently.
- The hash-name check only covers the default Keccak-256; dropping the `blake3` rename still passes.

**Fix:** parse into `serde_json::Value`, sort the object's keys and compare them to an exact list; check the JSON name of every `HashFamily` variant against its `Display`.

**Verify:** delete `#[serde(rename = "blake3")]` and the test must fail.

examples/src/parsers.rs
```diff
@@ -43,6 +43,14 @@ pub enum PcsOptions {
     Stir,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum OutputFormat {
```
tcoratger, line 48: **[NIT] Missing header doc**

A one-line header on the public enum helps: `/// How a prover example prints its measurements.`

Similarly, the report doc at `examples/src/binary.rs:408` links to `with_witness_seconds` by name; plain prose survives renames better.

tcoratger:
@gopikannappan I rebased this on `main` and pushed the review fixes to your branch as `a9e730c7`. I kept your authorship. Feel free to amend or drop anything.

- **Rebase.** `main` added `setup_seconds`, `pcs` and `whir` to the report. `PcsIdentity`, `WhirSummary` and `WhirRegime` now derive `Serialize`, spelled the same way as on the command line (`folding`, `unique-decoding`). The witness is timed once, around the trace generator in `preflight_then_maybe_prove`.
- **Clean JSON stdout.** With `--format json`, progress lines, preflight lines and the tracing span tree go to stderr. Human mode is unchanged.
- **Thread count.** Reads `current_num_threads()` directly, so it matches the backend the prover actually ran on.
- **Tests.** The report test now pins the exact set of JSON keys. A new test checks every hash's JSON name against its `Display`, and there's a CLI test for `--format`. I checked that the tests fail when either serde rename is removed.

Checked: `cargo test -p p3-examples` (lib and `prove_hash_binary`), stable clippy `-D warnings` with and without `parallel`, nightly fmt, sort, taplo, machete. I also ran `--format json` end to end and piped its stdout through a JSON parser.

tcoratger (approved):
Rebased on `main`, and every finding is addressed in `a9e730c7`. JSON stdout now parses cleanly, the thread count follows the real backend, and the tests pin the JSON keys and names. Good to merge once CI is green.

examples/examples/prove_hash_binary.rs
```diff
@@ -292,94 +114,110 @@ fn run(args: &Args) -> Result<(), String> {
...
-fn main() {
-    let args = Args::parse();
-    if let Err(error) = run(&args) {
-        panic!("{error}");
+        Ok(report) => match args.format {
+            OutputFormat::Human => {
+                println!("{report}");
+                println!("Proof Verified Successfully");
+            }
+            OutputFormat::Json => println!(
```
Nashtare, line 433: [P2] Keep JSON-mode stdout parseable

This prints a JSON object, but the unconditional `Proving …` lines and `ForestLayer::default()` also write to stdout. Running `cargo run -q -p p3-examples --example prove_hash_binary -- --objective blake-3-compressions --log-trace-length 2 --format json` produced 32 stdout lines: the progress line, 30 INFO trace lines, then the object; stderr was empty. Parsing the complete output with `jq -e .` fails (exit 5), so a scoreboard cannot consume the promised one-object output directly.

Please send progress/tracing to stderr or suppress it in JSON mode, and add a process-level test that parses the executable's entire stdout as one JSON object. Testing `serde_json::to_string(&report)` alone cannot catch the surrounding output.

Nashtare (changes requested):
Reviewed `30e926e82`. The timing boundaries and serial/parallel thread reporting look correct, and the focused report and CLI tests pass. Real serial and `RAYON_NUM_THREADS=2` runs both prove and verify and report the expected thread counts.

One issue needs fixing before approval: `--format json` does not emit parseable JSON on stdout (inline). Testing the complete executable output catches this; the current test only serializes the report struct.

Separately, rebasing onto current `main` (`6d673ed12`) requires resolving conflicts in `examples/examples/prove_hash_binary.rs` and `examples/src/binary.rs`.

examples/examples/prove_hash_binary.rs
tcoratger: Fixed. The branch is rebased on `main` (`6d673ed1`), so the conflicts are gone too.

- `a9e730c7`: in JSON mode, the `Proving …` line, preflight lines and the `tracing_forest` span tree go to stderr. Human mode is unchanged.
- `7be28348`: `examples/tests/json_output.rs` runs the `prove_hash_binary` executable with `--format json` (same profile and features as the test) and parses its entire stdout as one JSON object.

I checked that the test fails when either the progress line or the span tree goes back to stdout. It passes with nextest in dev and release, with and without `parallel`.

gopikannappan:
Thanks for the rebase, and for pushing the fixes rather than handing them back. That was generous of you.

I checked it on an M2: the example tests and the library tests all pass, `--format json` parses straight from stdout, and clippy is clean both with and without `parallel`.

Both of your code changes are better than what I had. Reading `current_num_threads()` reports the pool the prover really ran on, where mine reported what the build enabled, and those two differ in exactly the case I got wrong. Pinning the JSON key set is what stops the schema drifting quietly, where I had only been checking the values.

@Nashtare, your review landed on the commit just before the fix, so the stdout problem is already sorted. Your second point stood on its own, though, and it was a fair one: `serde_json::to_string(&report)` can never see what else the process prints. I had written the process-level test for it, and then found `7be28348` already there, matching the profile and features so it reuses the build. I dropped mine rather than push a duplicate.

One small thing it does not check: that the progress lines still arrive somewhere. A run that printed nothing at all would pass it. `assert!(stderr.contains("Proving"))` covers that, if either of you thinks it is worth the line. Happy either way, and happy to change anything else you would rather see differently.

Nashtare (approved):

PR #2281, opened 2026-09-21
refactor!(uni-stark,batch-stark,commit): group the preprocessed openings and name the trace-height bounds as a pair

0xShadowX:
Two API changes that fell out of the recent shape-check work, bundled because both are breaking and both touch the same verifier entry points.

### 1. `OpenedValues.preprocessed` — one grouped opening instead of two independent options

`preprocessed_local: Option<Vec<_>>` and `preprocessed_next: Option<Vec<_>>` made three states representable that no prover emits: a next row without a current row, a present-but-empty current row, and a next row whose width disagrees with the current row. Each had to be rejected at runtime in `p3-uni-stark` and again in `p3-batch-stark`, and the two lists had to stay in sync by hand.

`OpenedValues` now carries `preprocessed: Option<PreprocessedOpenedValues { local: Vec<_>, next: Option<Vec<_>> }>`. The first two shapes cannot be built; the width mismatch is refused by the `Deserialize` impl before a proof reaches the verifier. What remains in the verifiers is only what needs the AIR: the opening is present exactly when the AIR has preprocessed columns, and `next` is present exactly when the AIR reads the next row. Both verifiers, the batch-stark opening-claim builder and `verify_constraints` callers now go through the grouped value or the `preprocessed_local()` / `preprocessed_next()` accessors.

**Wire format is unchanged.** `Serialize`/`Deserialize` go through a private repr with the previous six fields, so postcard/JSON/bincode bytes are byte-identical to `0.8.0`. `proof::tests::preprocessed_opening_keeps_the_flat_wire_layout` asserts that against a hand-built flat struct for all three legal shapes, `preprocessed_opening_rejects_shapes_the_prover_never_emits` covers the three illegal ones, and the five stored `*_v0_8_0.postcard` compatibility fixtures in `uni-stark` and `batch-stark` still verify without regeneration.

### 2. `Pcs::log_max_lde_height` → `log_max_trace_height`

`log_min_trace_height` and `log_max_lde_height` bound the same quantity — the trace height a proof may claim, gated by `validate_degree_bits` before any domain is derived — but read as bounds on two different objects. Every backend already returns a trace-height bound (`TwoAdicStirPcs` subtracts its blowup for exactly that reason), so the rename makes the pair match; the trait doc now says what each bounds and points at the other. Five impls, the two STARK verifiers and one STIR test are updated; no values change.

### Verification

`cargo nextest run -p p3-uni-stark -p p3-batch-stark -p p3-commit -p p3-fri -p p3-stir -p p3-circle -p p3-examples`: 754 passed, including the compatibility fixtures. Workspace clippy and rustfmt clean.


tcoratger:
Please rm the ds store files :)

0xShadowX:
Done in f00143c — all six .DS_Store squatters have been evicted. In my defence, Finder felt very strongly about this refactor and wanted co-author credit. Added it to my global gitignore so it doesn't return for the sequel; only the Rust files remain in the diff now.

uni-stark/src/proof.rs
```diff
@@ -58,15 +58,211 @@ pub struct Commitments<Com> {
...
+    preprocessed_next: Option<Vec<Challenge>>,
+    quotient_chunks: Vec<Vec<Challenge>>,
+    random: Option<Vec<Challenge>>,
+}
+
+/// Borrowing twin of [`OpenedValuesRepr`] for serialization; slices encode like vectors.
+#[derive(Serialize)]
+#[serde(rename = "OpenedValues")]
+#[serde(bound(serialize = "Challenge: Serialize"))]
+struct OpenedValuesReprRef<'a, Challenge> {
```
tcoratger: **MAJOR — one struct, not two, and every field documented.**

Two definitions of the same six fields have to be edited in lockstep forever, and neither carries a single field doc. The duplication is avoidable: leave the row container generic and derive both directions from one definition.

```rust
#[derive(Serialize, Deserialize)]
#[serde(rename = "OpenedValues")]
struct OpenedValuesRepr<Row, Chunks> {
    /// Main trace evaluated at `zeta`.
    trace_local: Row,
    ...
}
```

Serializing instantiates it at `Row = &[Challenge]`, deserializing at `Row = Vec<Challenge>`. Serde infers `Row: Serialize` and `Row: Deserialize<'de>` on its own, so both `#[serde(bound(...))]` attributes disappear as well.

I tried the `Cow` version first, since it is the more obvious collapse. It does not survive contact: `Cow<'a, [Challenge]>` is only well-formed for `Challenge: Clone`, that bound lands on `Serialize for OpenedValues`, and from there it propagates into `OpenedValuesWithLookups` and `BatchOpenedValues` in `p3-batch-stark`, which fail to compile until they carry explicit `Clone` bounds of their own. Making a private wire detail widen the public serde bounds of a second crate is too high a price. The generic-container form costs nothing: same bounds as today, no lifetime, no extra crate touched.

Byte-identity is proven, not assumed. Across 12 shapes — three preprocessed layouts crossed with present/absent `trace_next` and present/absent `random` — the borrowed instantiation, the owned instantiation and a hand-written flat struct all produce identical bytes in postcard, bincode and `serde_json`, and the owned instantiation decodes the flat bytes in all three.

uni-stark/src/verifier.rs
```diff
@@ -408,9 +418,11 @@ where
 
     // Add the preprocessed commitment when the AIR declares preprocessed columns.
     if let Some(preprocessed_commit) = preprocessed_commit {
-        let mut pre_points = vec![(zeta, opened_values.preprocessed_local.clone().unwrap())];
-        if !air.preprocessed_next_row_columns().is_empty() {
-            pre_points.push((zeta_next, opened_values.preprocessed_next.clone().unwrap()));
+        // The shape check above guarantees the opening is present with the declared rows.
+        let preprocessed = opened_values.preprocessed.as_ref().unwrap();
```
tcoratger: **MINOR — `unwrap` held up by a comment across two functions.**

This is sound today: the shape check rejects an absent opening whenever the width in force is positive, and the commitment reaching here is filtered on that same width. So no proof can trip it. But the guarantee lives in a different function from the `unwrap`, and the only thing tying them together is the comment above.

The function that establishes the pairing can hand it back instead, which puts the invariant in the type:

```rust
Option<(Com<SC>, &'a PreprocessedOpenedValues<SC::Challenge>)>
```

The commitment and the opening then arrive together or not at all, the `unwrap` disappears, and the `.filter(|_| preprocessed_width > 0)` further down becomes redundant because the zero-width case already leaves the pair empty. Behaviour is unchanged on every path.

commit/src/pcs/univariate.rs
```diff
@@ -105,8 +105,12 @@ where
     /// or an arbitrary use of the underlying opening protocol.
     const ZK: bool;
 
-    /// The base-2 logarithm of the largest evaluation domain this PCS can construct.
-    fn log_max_lde_height(&self) -> usize;
+    /// The base-2 logarithm of the largest trace domain a proof may claim.
+    ///
+    /// A verifier rejects a proof-supplied height above this before deriving any domain from
+    /// it, so backends whose evaluation domain must fit a two-adic subgroup subtract their
+    /// blowup here. The counterpart of [`log_min_trace_height`](Self::log_min_trace_height).
```
tcoratger: **MINOR — two house-style violations in three lines.**

`[`log_min_trace_height`](Self::log_min_trace_height)` is an intra-doc link, which is a name reference that goes stale on rename; the neighbouring doc for the lower bound deliberately has none. And the sentence starting "A verifier rejects" runs across three physical lines.

Bullets, one sentence per line, and the pairing said in plain English:

```rust
/// The base-2 logarithm of the largest trace domain a proof may claim.
///
/// - A verifier rejects a proof-supplied height above this bound.
/// - The rejection happens before any domain is derived from that height.
/// - A backend whose evaluation domain must fit a two-adic subgroup subtracts its blowup.
///
/// This is the upper end of the pair; the smallest claimable height is the lower end.
```

tcoratger (commented):
## Review

Verdict: **COMMENT** — 1 MAJOR, 4 MINOR, 1 NIT. The two breaking changes are both right, and the wire-compatibility claim holds; the objections are to the shape of the serde bridge, one `unwrap`, and how far the evidence for byte-identity actually reaches.

Everything below was checked by running it, not by reading.

### What I confirmed

**The private repr is genuinely necessary.** The obvious alternative is `#[serde(flatten)]` on the grouped field, and flatten forces the serializer into a map of unknown length. Both non-self-describing formats refuse it outright:

```
postcard::to_allocvec(&flattened)  -> Err(SerializeSeqLengthUnknown)
bincode::serialize(&flattened)     -> Err(SequenceMustHaveLength)
serde_json::to_string(&flattened)  -> Ok
```

Postcard is the proof format, so flatten is off the table. A repr is the right answer.

**The three `Deserialize` rejections refuse exactly the illegal shapes and nothing legal.**
- `Some(vec![])`: `p3-batch-stark` maps a zero-width preprocessed matrix to `instances_meta = None` before it ever becomes an opening, and `p3-uni-stark` gates the whole opening on `preprocessed_width > 0`. No prover emits it.
- Unequal widths: both rows come from the same committed matrix, so they are the same width by construction.
- A next row with no current row is unrepresentable on the prover side.

**No panic is reachable there.** Both rejections are ordinary `serde::de::Error` constructions, and `len()` on a decoded `Vec` cannot overflow. A hostile proof is rejected, never aborts.

**The rename is complete and value-preserving.** Five impls plus the trait declaration, two verifiers, one STIR test; every impl body is untouched, and `log_max_lde_height` no longer appears anywhere in the tree.

**Nothing became dead.** Worth recording, though: `UnexpectedPreprocessedNext` is now unreachable *from the wire*, because a present-but-empty next row can never match a non-empty current row at decode time. It stays reachable for an in-memory proof, and both crates' tests construct exactly that, so it is defence in depth rather than dead code.

**The accessors do not leak the invariant.** They flatten two levels of `Option` into one for reading, which is all either verifier needs, and the grouped field stays public for anything that needs the precise shape. Construction still goes through the grouped type.

### One correction to the description

> The first two shapes cannot be built

Only the first. `PreprocessedOpenedValues { local: vec![], next: None }` is still constructible in memory — the empty current row is refused at decode and by both verifiers, not by the type. Worth saying plainly, because the difference matters to anyone building openings by hand.

### Where the evidence stops short

The compatibility fixtures decode a stored `0.8.0` proof and verify it. They never check that re-encoding it reproduces the stored bytes, which is the actual claim being made. One assertion per fixture closes that, and it passes today:

```rust
assert_eq!(postcard::to_allocvec(&proof)?, proof_bytes);
```

Also worth knowing for the record: of the five fixtures, only the two `batch-stark` ones carry a preprocessed opening at all, and both carry the *both-rows* shape. No stored fixture exercises the current-row-only shape, so the unit test is the only thing pinning it.

### Findings

- **MAJOR** — the serde bridge is two structs with the same six fields and no per-field documentation. See the inline comment.
- **MINOR** — `unwrap` on the verifier path. Inline.
- **MINOR** — the new trait doc carries an intra-doc link and spreads one sentence over three lines. Inline.
- **MINOR** — the accessors are missing `#[must_use]`.
- **MINOR** — the layout test only exercises postcard, which does not encode field names. I verified the gap is real: renaming one wire field with `#[serde(rename)]` leaves the postcard assertion green and is caught only once JSON is compared.
- **NIT** — `flat_encoding` writes the four unrelated field values out by hand a second time, so the test compares two hand-written literals instead of one value spread two ways. Deriving the flat struct from the grouped value removes the chance of them drifting apart.

I have pushed all of the above.

tcoratger:
Pushed `61514342` on top of your two commits — nothing of yours was rewritten, so amend or drop it freely.

### The abstraction change

The two repr structs are now one, generic over the row container:

```rust
#[derive(Serialize, Deserialize)]
#[serde(rename = "OpenedValues")]
struct OpenedValuesRepr<Row, Chunks> {
    /// Main trace evaluated at `zeta`.
    trace_local: Row,
    ...
}
```

Serializing instantiates it at `Row = &[Challenge]`, deserializing at `Row = Vec<Challenge>`. The six fields are declared once, serde infers both bounds on its own, and the two `#[serde(bound(...))]` attributes are gone.

I tried `Cow` first, because it is the more obvious collapse and it does produce identical bytes. It fails on the bound: `Cow<'a, [Challenge]>` is well-formed only for `Challenge: Clone`, that lands on `Serialize for OpenedValues`, and from there `OpenedValuesWithLookups` and `BatchOpenedValues` in `p3-batch-stark` stop compiling until they carry `Clone` bounds of their own. A private wire detail should not widen the public serde bounds of a second crate. The generic-container form has exactly today's bounds, no lifetime, and touches no extra file.

Worth recording for anyone who wonders why a repr exists at all: `#[serde(flatten)]` is not an option here. It forces a map of unknown length, and both non-self-describing formats refuse outright — postcard with `SerializeSeqLengthUnknown`, bincode with `SequenceMustHaveLength`. Postcard is the proof format.

### Wire identity

Checked by running it, in both directions.

- Across 12 shapes — three preprocessed layouts crossed with present/absent `trace_next` and present/absent `random` — the borrowed instantiation, the owned instantiation and a hand-written flat struct produce **identical bytes in postcard, bincode and `serde_json`**, and the owned instantiation decodes the flat bytes in all three.
- The unit test now compares against JSON as well as postcard. That matters: postcard does not encode field names, so renaming one wire field with `#[serde(rename)]` leaves the postcard assertion green and is caught only by the JSON comparison.
- All five stored `*_v0_8_0.postcard` fixtures verify without regeneration, and each now also asserts `postcard::to_allocvec(&proof)? == proof_bytes`. Decoding a `0.8.0` proof and re-encoding it reproduces the stored bytes exactly, which is the claim the fixtures were only half-testing before.

For the record: of the five fixtures only the two `batch-stark` ones carry a preprocessed opening, and both carry the both-rows shape. The current-row-only shape is pinned by the unit test alone.

### Mutation evidence

Five mutants, all killed:

| mutant | killed by |
| --- | --- |
| accept a next row with no current row | the matching rejection test, plus the message test |
| accept an empty current row | same |
| accept rows of unequal width | same |
| swap the two preprocessed fields on the wire | the layout test |
| rename one wire field | the layout test, **only** once JSON is compared |

### Also in the commit

- The opened-values impls are grouped under the type they belong to: the struct, its inherent accessors, the private bridge, then the two serde impls, and the preprocessed type after all of it. That is why the file looks reordered — it is a move, not new logic. The accessor bodies are unchanged.
- `process_preprocessed_trace` returns the commitment paired with the opening it binds, so the `unwrap` in the opening-claim builder is gone and the `.filter(|_| preprocessed_width > 0)` further down is no longer needed. Behaviour is identical on every path.
- Every field of the openings and of the wire layout has a doc comment; `#[must_use]` on both accessors.
- The largest-claimable-height doc drops the intra-doc link and splits into one sentence per line, matching the doc for the lower bound next to it.
- `serde_json` added as a `p3-uni-stark` dev-dependency, as it already is in nine other crates.

### Verification

`cargo test`, debug and release, for `p3-uni-stark`, `p3-batch-stark`, `p3-commit`, `p3-fri`, `p3-stir`, `p3-circle`, `p3-examples`, `p3-multi-stark`: 1127 passed debug, 1119 release, zero failures. `cargo clippy --all-targets --all-features -- -D warnings` clean on each. `cargo +nightly fmt --check` clean.

One small correction to the description while you are in there: "The first two shapes cannot be built" holds only for the first. `PreprocessedOpenedValues { local: vec![], next: None }` is still constructible in memory — the empty current row is refused at decode and by both verifiers, not by the type.

tcoratger (approved):
Approving on `61514342`. Nothing at BLOCKER or MAJOR remains.

Both breaking changes are right. Grouping the preprocessed rows removes a class of shape that the verifiers had to reason about twice, and the rename makes the two height bounds read as the pair they are. The wire format really is unchanged, and the fixtures now prove it in both directions rather than only on decode.

The one thing a future reader should keep in mind is that the type makes a next-row-without-a-current-row unrepresentable, but an empty current row and disagreeing widths are still constructible in memory — those are held by the decode boundary and by both verifiers. That is a reasonable split; it just is not the same as "unrepresentable".

tcoratger (approved):
Re-approving on `bdf42774`. Amended `61514342` with a documentation-only pass: a handful of comments were still packing two clauses onto one line with a semicolon or an `and`/`so`, which the house style splits. No code changed, and the crate tests are unchanged at 123 / 91 / 16 passing for `p3-uni-stark`, `p3-batch-stark` and `p3-commit`.

tcoratger:
Amended to `bdf42774`: documentation only, on top of the same commit. A handful of comments were still packing two clauses onto one line with a semicolon or an `and`/`so`, which the house style splits. No code moved, and the crate tests are unchanged at 123 / 91 / 16 passing for `p3-uni-stark`, `p3-batch-stark` and `p3-commit`, with clippy and rustfmt clean.

PR #2282, opened 2026-09-21
feat(security): use the DKT26 Johnson MCA bound

Nashtare:
## Summary

Johnson-regime estimates currently use the BCHKS25 proximity-gap bound: the shared batching helper keeps only its dominant term, while FRI evaluates the older full expression. Replace both with the complete conservative DKT26 line bound and its polynomial-curve extension derived here. Besides increasing reported security, this can reduce the proof-of-work and query counts that WHIR and STIR derive for real proofs.

For code dimension `k = 2^log_degree`, evaluation length `n`, `rho_minus = (k - 1) / n`, and `m >= 3`, the shared helper computes the exceptional line count in log space:

```text
C = 8 * n * (m + 1/2)^3 / (3 * rho_minus)
error <= ell * C / |F|
```

Here `ell` is the degree of the received polynomial curve: `num_functions - 1` for powers batching and `folding_factor - 1` for FRI. The helper includes the finite-degree correction and derives the scaled interpolation support, proving that variables outnumber equations after both ceilings. This cannot be obtained by multiplying an already-rounded finite line count. Constant codes use a separate, multiplicity-independent bound. Unsupported multiplicities/rates fail closed; the public batching APIs retain their existing `num_functions >= 2` precondition, now explicitly documented.

The source is [Dao–Kominers–Thaler, *Reed–Solomon Codes Beyond Johnson*](https://eprint.iacr.org/2026/2056), Theorem 5.12 and Appendix B.1–B.2 for the line. The curve extension uses **Lemma 5.3 and the order-zero case of §7.2, Equation (88)**, which are characteristic-free; it does not invoke the section's positive-order corollary. This is the improvement **within the classical Johnson regime**.

## Behavior and scope

- Share the bound between `SecurityAssumption` and FRI, preserving the existing agreement thresholds, list bounds, independent FRI reconstruction term, and challenge-specific grinding accounting.
- WHIR's `folding_pow_bits` can require approximately **5–6 fewer grinding bits** when the Johnson MCA term controls the budget. STIR's eta bisection can select a smaller safety gap and fewer queries. Its raw 100-bit Combine fixture, previously rejected, now fits: the actual eta/log-eta round trip selects `m = 11` and gives **100.4293 bits**; nominal `m = 10` would give 100.8230 bits.
- Classify the WHIR hiding report's Johnson MCA term as `Proven`. That call uses `num_functions = 2`, hence curve degree one: it relies directly on the literal line theorem, without the curve extension or an extra reserve. Preserve `Approximation` for API compatibility and downstream constructions; this crate's reports no longer produce it.
- Retain the existing conservative one-bit reserves in STIR and the WHIR non-hiding report. The WHIR reserve applies to reporting, not parameter derivation, and now has a regression that detects its removal.
- In the FRI-only fixture, the **commit term increases from 117.462 to 120.578 bits**. The unchanged query term is **120.486 bits**, so the improvement switches the bottleneck from commit to queries and raises the combined round-by-round LDT estimate to 120.486 bits. Parameters: conservative 127-bit field-size input, `k = 2^20`, blowup 8, arity 4, `m = 3`, 81 queries, 28 commit-grinding bits, and 17 query-grinding bits.
- Two `uni-stark` regressions now use 140 queries so that batching remains the limiting term and their checks still detect missing batching-grinding accounting. The examples' runtime query presets are unchanged.

The FRI fixture is not a 120-bit certificate for Miden or a complete STARK. Lookup, AIR composition, OOD/DEEP, batching, hashing, and transcript assumptions still need their own accounting.

## Validation

- Independent mathematical and code review of the finite bound and polynomial-curve extension; exact rounded-support fixtures and a 1,470-case grid independently evaluate interpolation feasibility and Equation (88), including large curve degrees.
- Mutation checks reject a reduced closed constant, an omitted curve factor, altered multiplicity rounding, a removed batch-size assertion, and removal of WHIR's reserve.
- Review follow-up: **609 tests** passed across `p3-security`, `p3-stir`, and `p3-whir` (four tests are marked ignored in the normal suite). Clippy, strict docs, nightly formatting, and whitespace checks passed.
- All **21 applicable CI checks passed** on review follow-up commit `1cfc95f4`: [platform builds, tests, formatting, and Clippy](https://github.com/Plonky3/Plonky3/actions/runs/35696140421), [WHIR exhaustive tests (baseline and AVX2)](https://github.com/Plonky3/Plonky3/actions/runs/35696140552), and [slow regression tests](https://github.com/Plonky3/Plonky3/actions/runs/35696140329). The unaffected Wasm job and release job were skipped by workflow conditions.


security/src/proximity.rs
```diff
@@ -10,13 +10,63 @@
...
+/// The chosen agreement parameter
+/// `alpha = (1 + 1/(2m))·sqrt(k/n)` gives `eta0 >= sqrt(rho_minus)/(2m)`;
+/// therefore an explicit `m >= 3` can be larger than the theorem's minimum
+/// and still satisfies the DKT26 Appendix B.1 interpolant. For a degree-`ell`
+/// curve, write `D = k - 1` and `t = m + 1/2`. Its weighted monomials use
+/// `i + D·j < t·sqrt(Dn)` and
+/// `ell·j + h < ell·t²/(3·rho_minus)`, hence `B_Z < ell·t²/(3·rho_minus)`.
+/// Section 7.2 Equation (88), with Appendix B.2's closed estimate, is linear
+/// in `ell` and yields the displayed `ell·C` factor without multiplying a
+/// rounded finite line count.
```
tcoratger: **[MINOR] The closed constant is this PR's derivation, not a paper theorem, and its decisive step is asserted rather than shown**

The comment states that weighted monomials satisfy `ell*j + h < ell*t^2/(3*rho_minus)`, hence `B_Z < ell*t^2/(3*rho_minus)`, and that the equation and appendix therefore give `ell*C`.

That chain is correct, but only two of its three links are in the paper.

- Equation (88) is jointly homogeneous of degree 1 in the curve degree and the support bound, so linear scaling of the support scales the count. In the paper.
- The appendix bounds each term separately, and each is linear in one or the other, so the substitution scales the whole right side. In the paper.
- `B_Z < ell*C_Z` needs the interpolant of Proposition B.1 to exist over the scaled support. **Not in the paper.** The existence proof needs a variable count above an equation count, and the paper cites the line case only. For a curve both counts carry ceilings that do not factor out of the degree, leaving a residual rounding term that is not multiplied by it.

The step holds. It was reconstructed and checked numerically over degree 1..30, inverse rate 1..12, multiplicity 3..3000 and curve degree 1..65535: the variable count exceeds the equation count everywhere, worst relative margin 1.1e-7 at an unreachable corner and at least 2% at every realistic shape. Evaluating equation (88) directly gives a ratio to `ell*C` peaking at 0.853, and 0.47 to 0.67 at the three shipped configurations.

So the bound is sound with comfortable margin. What is missing is that a security constant ships with its load-bearing step as an unproved sentence in a comment.

**Fix:** state that link as this PR's own derivation, including the rounding point and the fact that the residual is absorbed by the constant's slack.
**Verify:** a test evaluating equation (88) directly against `ell*C` over a parameter grid would pin it. The current tests re-assert the closed form against itself, so they cannot catch a failure of this step.

security/src/fri.rs
```diff
@@ -3,8 +3,8 @@
 //! Conjectured regime: random-words bound, [2025/2010] §1.5.
 //! Legacy regime: historical pre-random-words ethSTARK query estimate, [2021/582].
 //! Proven regime: round-by-round, [2024/1553] Theorems 2 & 3, with the
-//! BCHKS25 LDR commit bound ([2025/2055] Theorem 4.2). Cross-checked
-//! against Ethereum's `soundcalc`.
+//! DKT26 Johnson line-MCA commit bound ([2026/2056] Theorem 5.12 and
```
tcoratger, line 6: **[MINOR] The section citation leads a reviewer to the wrong statement, which carries a hypothesis this route does not**

Pointing at the section alone leads to the named powers-batching corollary. That corollary requires characteristic zero or greater than a bound involving the jet parameter, and its certified support yields a uniform constant far weaker than the one shipped here.

The route actually taken is the order-zero one: equation (88), and the sentence immediately after it stating that the ordinary case has neither the characteristic condition nor a restriction on the received curve degree. The supporting lemma likewise needs no restriction on characteristic or degree.

That route is correct and strictly stronger. The risk is purely in the citation: a reviewer who checks the corollary will conclude a characteristic restriction applies to the small binary-friendly primes, and will either reject a correct change or add a guard that is not needed.

**Fix:** cite the equation and the order-zero case explicitly, and note that characteristic-freeness comes from the lemma rather than the corollary.

security/src/assumption.rs
```diff
@@ -250,7 +237,8 @@ impl SecurityAssumption {
             num_functions >= 2,
             "num_functions must be >= 2 to compute proximity gaps error",
         );
-        let error = jb_prox_gaps_dominant_term_bits(log_degree, log_inv_rate, m);
+        let error =
+            crate::proximity::johnson_exceptional_line_count_log2(log_degree, log_inv_rate, m);
         let num_functions_1_log = libm::log2(num_functions as f64 - 1.);
         field_size_bits as f64 - (error + num_functions_1_log)
     }
```
tcoratger, line 267: **[MINOR] The multiplicity is a free argument, while the theorem defines it from the proximity parameter**

Theorem 5.12 does not permit an arbitrary multiplicity. Its hypothesis defines it as the maximum of three and a quantity fixed by the agreement threshold. This entry point takes it as a parameter and checks only the floor of three, with nothing tying it to the radius the surrounding regime decodes at.

The contract is honoured by every current caller, and here is why, which the comment omits. With the agreement radius set at one plus a half-reciprocal of the multiplicity times the square root of the rate, and the nominal rate strictly above the reduced rate, the implied proximity parameter exceeds the reduced rate's square root over twice the multiplicity. Rearranging gives the theorem's defined value at most the supplied one, and the appendix estimate decreases in that argument, so evaluating at a larger value is conservative. The two coincide exactly at all three shipped configurations.

The residual risk is that this holds only because every caller pairs the same multiplicity with the matching radius helpers. A future caller supplying a value below its own regime's would receive a number with no guarantee behind it, and nothing would complain.

**Fix:** take the radius or its logarithm alongside the multiplicity and assert the coupling in debug, or record the three-line argument above at the function so the contract is auditable.

security/src/proximity.rs
```diff
@@ -10,13 +10,63 @@
...
+) -> f64 {
+    if m < 3 || log_inv_rate == 0 {
+        return f64::INFINITY;
+    }
+
+    if log_degree == 0 {
+        // log₂(n(n − 1)/2), written without constructing n.
+        let log_n_minus_one = log_inv_rate as f64 + log2(1.0 - pow(2.0, -(log_inv_rate as f64)));
+        return log_inv_rate as f64 + log_n_minus_one - 1.0;
+    }
```
tcoratger, line 100: **[INFORMATIONAL] The constant-code branch is the largest loosening in the change, and nothing reaches it**

For dimension one this returns the logarithm of a binomial coefficient. At inverse rate three and multiplicity ten that is 4.81 bits where the previous dominant-term expression gave 23.88 — a 19-bit jump, the single largest loosening here, and the only branch that ignores the multiplicity entirely.

It is also unreachable. The branch was instrumented and all five suites run, including a 36-thousand-case property test sweeping trace lengths from zero: it is never entered. The three plausible entries are each blocked upstream — the folding error returns zero bits at trace length zero, the hiding report needs a randomized dimension of one, and the batched path needs a degree-one class. Even with an attacker-supplied degree of zero through the public reporting API, the composite minimum is pinned to zero by another term.

The constant is sound: the relevant lemma covers polynomial curves of any degree over any field, and the value used upper-bounds both of its cases. So there is no impact today.

Worth a note because it is a 19-bit step sitting behind a single unit test, and a future caller that legitimately reaches dimension one would take it silently.

whir/src/pcs/security.rs
```diff
@@ -96,8 +96,8 @@ where
     };
     let add_folds = |errors: &mut Vec<ErrorBits>, variables, rate, folds, pow| {
         // Use the largest degree/rate bound for every binary fold in a phase.
-        // The Johnson helper retains the dominant term; one additional bit
-        // covers its positive lower-order terms at the fixed m = 10.
+        // Retain the legacy one-bit Johnson reserve. The shared helper now
+        // includes a complete theorem bound, so this is extra conservatism.
         let gap = assumption.prox_gaps_error(variables, rate, field_bits, 2)
```
tcoratger, line 101: **[MINOR] The one-bit reserve is still applied and is enforced by nothing**

The reserve survives the refactor, which is the important part. But deleting it entirely leaves every test in all five crates passing. Its sibling in the other protocol is guarded — deleting that one fails an integration test.

This matters more after this change than before it. The comment previously justified the bit as covering positive lower-order terms; it now describes it as extra conservatism. A reserve documented as unnecessary and tested by nothing is one cleanup away from disappearing.

Note also that the derivation path does not apply this reserve, so the derived proof-of-work bits and the reported bits already differ by exactly this bit. Only the report is conservative.

**Fix:** assert the reported gap equals the unreserved term minus one under this assumption.
**Verify:** that test fails when the subtraction is removed; today nothing does.

security/src/fri.rs
```diff
@@ -203,18 +203,16 @@ pub fn commit_phase_error_ldr_m(
...
 
-    let num = (2.0 * pow(m_shifted, 5.0) + 3.0 * m_shifted * pp * rho) * n;
-    let den = 3.0 * rho * sqrt_rho;
-    let eps_linear = num / den + m_shifted / sqrt_rho;
-    let eps_powers = eps_linear * folding_minus_one;
-    let bits_linear =
-        shape.modulus_bits as f64 - log2(eps_powers.max(1.0)) + regime.commit_pow_bits as f64;
+    let log_n_plus_one = lde_log as f64 + log2(1.0 + pow(2.0, -(lde_log as f64)));
 
     let bits_n_over_q = shape.modulus_bits as f64
```
tcoratger, line 218: **[NIT] The retained cross-check cannot bind for any supported degree**

The comment frames this as a safety net beside the linear bound. Working out the difference, it is at most `-3.03 - r/2` for every multiplicity at or above the floor, so the linear term is always the smaller of the two and always wins the minimum. Confirmed empirically: removing the minimum breaks no test, and in the new fixture the two sit at 120.578 against 125.693.

It is not useless — it is exactly what caps the degenerate dimension-one case — but it is not a guard against a too-optimistic constant, which is what the comment implies.

**Fix:** one sentence saying what it actually bounds.

security/src/whir/hiding.rs
```diff
@@ -46,8 +46,8 @@ pub enum HidingBoundClassification {
     Conditional,
     /// The term is a numerical approximation, not a conservative theorem bound.
     ///
-    /// This is used for the current Johnson MCA helper, which retains only the
-    /// dominant finite-parameter term.
+    /// Retained for compatibility with analyses that use an explicitly
+    /// approximate bound; the current Johnson MCA helper is theorem-backed.
     Approximation,
```
tcoratger, line 51: **[NIT] The variant now has no producers, and one arm of the combinator is unreachable**

Both classification helpers are the only producers, and neither can return this variant after the change. The sole remaining construction is in a test. The enum is marked non-exhaustive so retaining it costs nothing, but the comment describes it as retained for analyses using an explicitly approximate bound — a choice the crate no longer offers.

**Fix:** say what it is retained for, or note that it exists for downstream constructions rather than for anything produced here.

stir/src/soundness.rs
```diff
@@ -710,18 +710,25 @@ mod tests {
...
+        );
+        let eta = jb
+            .stir_combine_eta(155, log_inv_rate, log_d_star, ell, 100)
+            .expect("the improved Johnson bound fits the raw target");
+        assert!(jb.stir_eta_is_valid(log_inv_rate, eta));
+        assert!(
+            combine_error_at_log_eta(jb, log_d_star, log_inv_rate, 155, ell, libm::log2(eta))
+                >= 100.0
         );
     }
```
tcoratger, line 736: **[NIT] The newly admitted configuration clears its target by 0.82 bits, with a multiplicity the codebase itself flags as fragile**

This is the one configuration the previous bound rejected and this one admits, so it is worth recording how much room it has. At degree 2^20, inverse rate one and a 155-bit field the old expression gives 95.539 and the new one 100.823.

The crossing rests entirely on the curve-degree extension: the curve degree here is about 1.57 million, far from the line the theorem literally covers. Checked at exactly these parameters — the variable count exceeds the equation count with 4.6% margin, and the direct evaluation gives a ratio of 0.664. Justified.

The nit is the margin next to a known fragility. The multiplicity here comes from a ceiling landing one unit in the last place below ten, which this crate already documents; reassociating that expression flips it to eleven and moves every Johnson proximity term by 0.39 bits. One slip still clears at 100.43, and this test's window would catch it, but a configuration derived through the production path rather than this test would not.

security/src/assumption.rs
```diff
@@ -250,7 +237,8 @@ impl SecurityAssumption {
             num_functions >= 2,
             "num_functions must be >= 2 to compute proximity gaps error",
```
tcoratger, line 261: **[NIT] One degenerate input panics where the stated contract is to fail closed**

The documented discipline is that unsupported inputs return infinity so callers fail closed. A single function reaches an assertion instead.

It is unreachable today — the caller above returns early below two functions, and every other call site clamps or skips — so there is no impact. It is an inconsistency in a public entry point on exactly the degenerate input the fail-closed discipline would otherwise cover.

tcoratger (approved):
Two independent audits: the mathematics against eprint 2026/2056, and the code integration and blast radius. **No BLOCKER, no MAJOR.** The transcription is exact, every hypothesis of Theorem 5.12 is discharged by a check, the `ell`-linearity extension holds, both pinned vectors and the whole reference table reproduce by hand, every invalid input fails closed, both one-bit reserves survive, and grinding is not double-counted. The findings below are documentation-integrity and test-coverage items around a load-bearing step; none of them changes a number.

Approving. Two things worth reading before merge, both in the description rather than the code.

## The headline figure is not the one this bound produces

`120.4862138717` is the **query-phase** term: `81 * (1.5 - log2(7/6)) + 17`. This PR does not change it. The DKT-dependent number is the **commit** term, `120.57793385796532`, and the two are only 0.09 bits apart. Under BCHKS25 the commit term was `117.46`, so the change moves it **+3.02 bits and flips which term binds**. That flip is why two `uni-stark` tests had to raise `num_queries` to 140: the improved commit bound stopped being binding, and the tests watching the batching site went blind. A legitimate repair, but it is the visible symptom of the whole change and belongs in the description.

## The loosening reaches real parameters, not only reports

The description says the change "improves the reported security". It does more than that. Closed form for the shift, confirmed against the reference table to 1e-14:

```text
delta_bits = 2*log2(m + 1/2) + r/2 - 2 + log2(1 - 2^-d)
```

Every in-domain call gets more permissive, minimum `+1.11` bits at `(m=3, r=1, d=1)`. Two consumers feed real proofs rather than reports:

- WHIR `folding_pow_bits` — up to **5 to 6 fewer proof-of-work bits demanded of the prover**.
- STIR's eta bisection — smaller derived eta, larger delta, **fewer queries**; a 100-bit Combine target that was rejected now fits.

Both are intended and both are justified. They deserve to be named, because a reader of the current description would not expect prover-side work to drop.

**[MINOR] The reclassified label sits on the one path that spends no margin.** `security/src/whir/hiding.rs:500-506`. The Johnson MCA term is `prox_gaps_error(..) - log2(width)` with no one-bit reserve, while the sibling non-hiding report at `whir/src/pcs/security.rs:101` keeps one. So the path that newly emits a certified number is the path with zero margin, and the path that keeps margin emits no certificate. Internally inert today — nothing gates on the label — but both the enum and the accessor are `pub` and re-exported, so a downstream verifier asking "is this proven?" sees the change.

**The label itself is honest**, and for a reason the description does not give: that call site passes `num_functions = 2`, so the curve degree is 1, the literal line case of the theorem. The hiding report never touches the curve extension. One sentence saying so would stop a future caller attaching the same label to a curve-degree term, where it would be overstated.

## What was checked and found clean

Transcription constant by constant, including `(k-1)` rather than `k` in the reduced rate, which the paper requires rather than permits. Monotonicity in `n`, `m` and degree. Rate-1 and sub-floor multiplicity both fail closed. No `0.0` error path exists, so the maximally dangerous "reads as perfect security" case is excluded, and the PR adds the test that catches it. No NaN or infinity is producible through any sink. Unifying the helper with FRI is lossless in the conservative direction for both, since FRI previously carried strictly more terms. Grinding appears once per term, and the fixture pins `commit - ungrounded_commit == 28.0` exactly. Eleven mutations run: a 2x looser bound is caught by six tests across three crates, including two STIR integration tests this PR does not touch — which is the strongest available evidence the numbers were not tuned until green.

**Bottom line:** merge. The findings are documentation and test-coverage debt around a correct change.

security/src/proximity.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

The helper now labels the polynomial-curve extension as derived here and gives the rounded variable/equation counts and a positive-surplus proof for both ceilings. Added exact rounding fixtures (including H=97 for degree three versus 3*32 for the line) and a 1,470-case integer-support grid that evaluates Equation (88) independently before comparing it with the public bound. Reducing the constant from 8 to 6 or omitting the curve factor makes the new grid fail.

security/src/fri.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

The FRI, STIR, and uni-stark citations now identify Equation (88), its order-zero case, and Lemma 5.3 explicitly. The shared derivation also states that this characteristic-free route does not invoke the positive-order powers-batching corollary.

security/src/assumption.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

Added a decoding-radius contract directly to prox_gaps_error_jb_at_m: the required alpha, eta0 >= sqrt(rho_minus)/(2*m), and therefore the theorem minimum <= supplied m. It also states that the bound grows with m and warns that an m below the surrounding radius requirement is unsound.

security/src/proximity.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

Documented the dimension-one branch as a separate, multiplicity-independent Lemma 3.2 bound, including its potentially substantial tightening. Expanded the direct checks to n=2,4,8 and m=3,10,1000 so the separate behavior is explicit and covered.

whir/src/pcs/security.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

Added a regression through the public prescribed_security report using a canonical single-phase Johnson configuration. It reconstructs the complete probability sum, explicitly charging the unreserved gap minus one for each of four binary folds. Removing the production reserve fails the test; restoring it passes. The full WHIR unit suite passes (197 tests, three ignored). The PR description now distinguishes reporting reserve from parameter derivation.

security/src/fri.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

The function documentation now states the ordering explicitly: for positive-degree codes with m >= 3, MCA gives fewer security bits and determines the minimum; the reconstruction term can instead cap the separate constant-code bound.

security/src/whir/hiding.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

The variant documentation now says it is retained for API compatibility and downstream constructions, and that this crate's reports do not currently produce it. The hiding call site and PR description also explain why its Johnson Proven label uses the literal degree-one line theorem.

stir/src/soundness.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

Checked the actual eta -> log_eta production path: it currently selects m=11 and gives 100.42931092148666 bits; 100.82304452132141 is the nominal m=10 value. The former 100..101 assertion accepted both. The regression now pins the actual m=11 value, and changing the multiplicity ceiling to a floor makes it fail. The comment and PR description record both margins.

security/src/assumption.rs
Nashtare: Addressed in [1cfc95f4e](https://github.com/Plonky3/Plonky3/commit/1cfc95f4e12862b31b5f6ab60b262b7a3bc36fa7).

Kept the existing nontrivial-batch API precondition: the other public proximity-gap entry points already assert num_functions >= 2, with existing panic tests. Added explicit Panics sections to all three entry points and direct empty/singleton tests for the explicit-m method; removing its assertion fails both tests. The docs now distinguish this batch-size precondition from the helper's fail-closed behavior for unsupported m or inverse rate, preserving the established API behavior.

